Issue #1021, opened 2016-02-26
signal.clean t_r default value


fliem:
This came up in a conversation with @GaelVaroquaux at Brainhack Paris, 2016.

It seems that in the example http://nilearn.github.io/auto_examples/03_connectivity/plot_adhd_spheres.html#sphx-glr-auto-examples-03-connectivity-plot-adhd-spheres-py

the t_r is set to 2.5 seconds at masking. The time series’ header says tr = 2.0 s. Is the header wrong, or the example?

More generally, signal.clean has a t_r default value of 2.5. It would be more intuitive for the default to be None, and if no t_r is provided, it is fetched from the header. 
@GaelVaroquaux’s point on this: since the header info is often wrong, it would be best to set the default to None, and in the  case of bandpass filtering without explicitly providing a t_r raise an error and print the header t_r with a word of caution that the header info might be wrong.

AlexandreAbraham:
+1, I've already raised this problem in the past.

sahmed95:
If no t_r is provided by user, the script crashes rather than setting of default value of 2.5
```python
from nilearn import datasets
adhd_dataset = datasets.fetch_adhd(n_subjects=1)

from nilearn import input_data
masker = input_data.NiftiSpheresMasker(
    dmn_coords, radius=8,
    detrend=True, standardize=True,
    low_pass=0.1, high_pass=0.01,
    memory='nilearn_cache', memory_level=1, verbose=2)

func_filename = adhd_dataset.func[0]
confound_filename = adhd_dataset.confounds[0]
print (confound_filename)
time_series = masker.fit_transform(func_filename,
                                   confounds=[confound_filename])
Error
```dev/nilearn/nilearn/signal.py in clean(signals, sessions, detrend, standardize, confounds, low_pass, high_pass, t_r)
    501 
    502     if low_pass is not None or high_pass is not None:
--> 503         signals = butterworth(signals, sampling_rate=1. / t_r,
    504                               low_pass=low_pass, high_pass=high_pass)
    505 

TypeError: unsupported operand type(s) for /: 'float' and 'NoneType'
```
Also, how do I find the t_r value from the header? New to this so I tried
```python 
import nibabel as nib
filename = '/nilearn_data/filtered_func_data.nii.gz'
img = nib.load(filename)
hdr = img.header
print (hdr) 
```
But there is no time series parameter in the header. Is it anything to do with confounds?

fliem:
[...]

PR #1164, opened 2016-07-05
[MRG+1] raise explicit TR error

salma1601:
fix #1163

PR #1830, opened 2018-10-20
[MRG+1] Sets default value of t_r in singal.clean and image.clean_img to None

miykael:
This PR closes https://github.com/nilearn/nilearn/issues/1021.

As recapped by @GaelVaroquaux (https://github.com/nilearn/nilearn/issues/1021#issuecomment-190209612), the following changes were requested:
 1. `signals.clean` should have a default of `None` (actually, `t_r` should always default to `None`) https://github.com/nilearn/nilearn/blob/master/nilearn/signal.py#L347
 2. When `t_r` is `None` and we are asking for filtering, we should raise an error (probably in `filter_and_extract` https://github.com/nilearn/nilearn/blob/master/nilearn/input_data/base_masker.py#L24) and inspect the nifti image header to suggest a value.

https://github.com/nilearn/nilearn/pull/1164 added the error message for `clean`, if `t_r` is set to `None`, but the rest was still missing.

This PR...
 - sets the default value of `t_r` in `signal.clean` and `image.clean_img` to `None`
 - adds an error message to `image.clean_img` if `t_r` is `None`, that suggests the TR from the NIfTI header
 - tests for this error.

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -852,7 +852,8 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
         Respectively low and high cutoff frequencies, in Hertz.
 
     t_r: float, optional
-        Repetition time, in second (sampling period).
+        Repetition time, in second (sampling period). Set to None if not
+        specified. Mandatory if used together with low_pass or high_pass.
 
     ensure_finite: bool, optional
         If True, the non-finite values (NaNs and infs) found in the images
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -880,6 +880,15 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
     from .image import new_img_like
 
     imgs_ = check_niimg_4d(imgs)
+
+    # Check if t_r is set, otherwise propse t_r from imgs header
+    if low_pass is not None or high_pass is not None:
+        if t_r is None:
+            t_r = imgs.header.get_zooms()[3]
+            raise ValueError(
+                "Repetition time (t_r) must be specified for filtering. You "
+                "specified None. imgs header suggest it to be {0}".format(t_r))
+
     data = signal.clean(
         imgs_.get_data().reshape(-1, imgs_.shape[-1]).T, sessions=sessions,
         detrend=detrend, standardize=standardize, confounds=confounds,
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -800,7 +800,7 @@ def math_img(formula, **imgs):
 
 
 def clean_img(imgs, sessions=None, detrend=True, standardize=True,
-              confounds=None, low_pass=None, high_pass=None, t_r=2.5,
+              confounds=None, low_pass=None, high_pass=None, t_r=None,
               ensure_finite=False):
     """Improve SNR on masked fMRI signals.
 
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -881,10 +881,13 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
 
     imgs_ = check_niimg_4d(imgs)
 
-    # Check if t_r is set, otherwise propse t_r from imgs header
+    # Check if t_r is set, otherwise propose t_r from imgs header
     if low_pass is not None or high_pass is not None:
         if t_r is None:
             t_r = imgs.header.get_zooms()[3]
+
+            # We raise an error, instead of using the header's t_r as this
+            # value is considered to be non-reliable
             raise ValueError(
                 "Repetition time (t_r) must be specified for filtering. You "
                 "specified None. imgs header suggest it to be {0}".format(t_r))
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -884,13 +884,13 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
     # Check if t_r is set, otherwise propose t_r from imgs header
     if low_pass is not None or high_pass is not None:
         if t_r is None:
-            t_r = imgs.header.get_zooms()[3]
 
             # We raise an error, instead of using the header's t_r as this
             # value is considered to be non-reliable
             raise ValueError(
                 "Repetition time (t_r) must be specified for filtering. You "
-                "specified None. imgs header suggest it to be {0}".format(t_r))
+                "specified None. imgs header suggest it to be {0}".format(
+                    imgs.header.get_zooms()[3]))
 
     data = signal.clean(
         imgs_.get_data().reshape(-1, imgs_.shape[-1]).T, sessions=sessions,
--- a/nilearn/tests/test_signal.py
+++ b/nilearn/tests/test_signal.py
@@ -337,12 +337,13 @@ def test_clean_frequencies():
     sx1 = np.sin(np.linspace(0, 100, 2000))
     sx2 = np.sin(np.linspace(0, 100, 2000))
     sx = np.vstack((sx1, sx2)).T
-    assert_true(clean(sx, standardize=False, high_pass=0.002, low_pass=None)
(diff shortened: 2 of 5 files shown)


nilearn/image/image.py
```diff
@@ -880,6 +880,15 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
     from .image import new_img_like
 
     imgs_ = check_niimg_4d(imgs)
+
+    # Check if t_r is set, otherwise propse t_r from imgs header
+    if low_pass is not None or high_pass is not None:
+        if t_r is None:
+            t_r = imgs.header.get_zooms()[3]
+            raise ValueError(
```
bthirion: To be sure; Should we raise an error here or used the returned t_r ? Are we assuming that this t_r value is not reliable ?
(I think that BIDS conventions indeed require to provide the t_r value in an associated json file).

nilearn/tests/test_signal.py
```diff
@@ -337,12 +337,13 @@ def test_clean_frequencies():
...
-                .max() < 0.01)
-    assert_true(clean(sx, standardize=False, low_pass=0.01).max() > 0.9)
-    assert_raises(ValueError, clean, sx, low_pass=0.4, high_pass=0.5)
+    assert_true(clean(sx, standardize=False, high_pass=0.002, low_pass=None,
+                      t_r=2.5).max() > 0.1)
+    assert_true(clean(sx, standardize=False, high_pass=0.2, low_pass=None,
+                      t_r=2.5) .max() < 0.01)
+    assert_true(
+        clean(sx, standardize=False, low_pass=0.01, t_r=2.5).max() > 0.9)
+    assert_raises(ValueError, clean, sx, low_pass=0.4, high_pass=0.5, t_r=2.5)
```
bthirion: If we agree on raising an error, then you should include an assert_raises for cases where the t_r is not provided.

nilearn/image/image.py
miykael: I implemented this according to Gaël's comment [here](https://github.com/nilearn/nilearn/issues/1021#issuecomment-190209612), but I think the discussion is valid.
In my opinion, the `t_r` value is not reliable. For example, SPM12's slice timing correction resets the TR in the header to 0. The same can happen if users create new niftis, without setting the header correctly. Unfortunately, users rarely check that kind of things.
Therefore, a warning (only if temporal filters are applied), can point to bugs in a users code. Also, the BIDS convention wouldn't help with the automatic setting of `t_r`, but with the corresponding warning, users can verify that everything is correct.

bthirion: Makes sense. Maybe you want to add a short comment in the code stating this logic: namely that this value is considered as non-reliable, so that we chose to raise an error anyway.

miykael: Good idea! Just did this with https://github.com/nilearn/nilearn/pull/1830/commits/ffa2167d681e8632a06d74c5f5af1678f15235ce.

bthirion:
It is actually 'time of repetition'
The BIDS specification calls it "TR" so it is safe to use it.

kchawla-pi:
It is a bad variable\parameter name from a software engineering perspective. We should call it `repetition_time`, which is unambiguous.

nilearn/image/image.py
```diff
@@ -880,6 +880,18 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
     from .image import new_img_like
 
     imgs_ = check_niimg_4d(imgs)
+
+    # Check if t_r is set, otherwise propose t_r from imgs header
+    if low_pass is not None or high_pass is not None:
```
KamalakerDadi: ```clean_img``` calls ```signal.clean```. May be we can improve the error message setting in ```signal.clean``` ?

This made me comment because I tried to run the script and signal.clean reports error message simply as 
"ValueError: Repetition time (t_r) must be specified for filtering". Not consistent with clean_img.

nilearn/image/image.py
```diff
@@ -880,6 +880,18 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
...
+        if t_r is None:
+            t_r = imgs.header.get_zooms()[3]
```
KamalakerDadi: overwriting t_r which is None to a value of imgs.header.get_zooms()[3] looks dangerous to me.

nilearn/image/image.py
```diff
@@ -880,6 +880,18 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
...
+
+            # We raise an error, instead of using the header's t_r as this
+            # value is considered to be non-reliable
+            raise ValueError(
+                "Repetition time (t_r) must be specified for filtering. You "
+                "specified None. imgs header suggest it to be {0}".format(t_r))
```
KamalakerDadi: Can we not directly call imgs.header.get_zooms()[3] ?

KamalakerDadi:
These API changes in signal.clean and clean_img needs to be added in ```doc/whats_new.rst```. 

There are two things changed: default ```t_r=2.5``` is changed from to ```t_r=None```. Second, if ```low_pass``` or ```high_pass``` is specified, then t_r should be specified otherwise it will raise an error. Or something like this will help users understand by looking at the change log.

```t_r``` documentation needs to be improved as well at clean and clean_img. We should say there that t_r is specified if low pass or high pass is specified and defaults to None.

KamalakerDadi:
```t_r``` is used in many places like input_data, decomposition, decoding modules. Do we need update ```t_r``` documentation there in those modules as well ?

Right now, we simply say. Please see related documentation signal.clean for details. 
@GaelVaroquaux @bthirion

nilearn/image/image.py
miykael: You're right. I will feed it directly to the `format` function.

nilearn/image/image.py
miykael: Yes, that's a good idea.

nilearn/image/image.py
miykael: I rewrote the `signal.clean` message with https://github.com/nilearn/nilearn/pull/1830/commits/d864d82cee67f669986839b2c781d80ea8c2e8ca but I don't see how the error message in `signal.clean` could know about a `t_r` value from a header. What kind of error message do you have in mind?

KamalakerDadi: Good point.

>What kind of error message do you have in mind?

What we rewrote is good to me.

GaelVaroquaux:
LGTM. Thanks for the excellent work. As far as I am concerned, this is good for merge.

Unless there are other important comments, I will merge this tomorrow morning.

KamalakerDadi:
>Unless there are other important comments. I will merge this tomorrow morning.

We need to atleast wait for CI's to come back.

GaelVaroquaux:
@KamalakerDadi : indeed, I had not seen.

The problem is that they probably won't on this PR: it's probably due to the outage of github this morning.

@miykael : would you mind issuing another PR to catter for this?

KamalakerDadi:
atleast t_r documentation in clean and clean_img needs to be updated.

miykael:
@KamalakerDadi, I've reacted to all your comments and updated `doc/whats_new.rst` with https://github.com/nilearn/nilearn/pull/1830/commits/383239c43b3d7fbe07936c5fc7a5a7948d1d0650, under 0.5.0 beta, correct?

miykael:
@GaelVaroquaux - To make sure that I understand you correctly. I have to just create a new branch and push a PR from there to get CI running again?

KamalakerDadi:
>under 0.5.0 beta, correct?

If I got it right, next release is 0.5. Then it should be in there

GaelVaroquaux:
@miykael : maybe just pushing anew here will start the tests, as they are running in the other PRs. You can do a git --amend, and push here to trigger the tests.

GaelVaroquaux:
The tests have passed, actually. Hence, I shall merge!

PR #1766, opened 2018-09-28
switch from papaya to brainsprite in plotting.view_stat_map

pbellec:
I really love the new 3D interactive viewer (`plotting.view_stat_map`), but the notebooks it is producing are huge. In this PR, I am proposing to switch from papaya to brainsprite, which is a js library I developed for the exact purpose of embedding lightweight 3D viewers in html pages (http://github.com/simexp/brainsprite.js).

The first difference with papaya is that it is using a jpg or a png containing all sagital slices of a volume as well as json metadata to store the brain images. That tend to be quite smaller than a nifti (depending on the numerical precision of the nifti). That also means that brainsprite can render brains with core html5 features, and no dependencies. So the brainsprite library weighs 15kb (500 lines...), as opposed to 2Mb for the current papaya html template. I have attached two brain viewers embedded in jupyter notebooks. The [Papaya-based notebook](http://nbviewer.jupyter.org/github/pbellec/misc/blob/master/example_view_stat_map_papaya.ipynb) is 12Mb, while the [brainsprite-based notebook](http://nbviewer.jupyter.org/github/pbellec/misc/blob/master/example_view_stat_map_brainsprite.ipynb) is 500kb. Again, this reflects a core difference in design: papaya is a full brain viewer app, featuring nifti reading as well as colorbar etc. Brainsprite is a minimal, fast brain viewer working from a pre-generated sprite.

Which makes a transition with the second point: all the action for the generation of the brain volume happens in python. There is a new function called save_sprite that generates the brain sprite as well as the json meta data. It relies on matplotlib, as well as nilearn's own functions. In particular, thresholding and colormap generation are all done with nilearn's code. Resampling as well. This means that it will be easier to maintain and evolve for nilearn's developpers. The current version replicates all the arguments of plot_stat_map, including draw_cross, annotate, cut_coords and a few other (with a few as bonus, such as opacity).

This PR is far from polished, there are a few oustanding issues, [here](https://github.com/brainsprite/brainsprite.py/issues). I also need to look into the doc and testing. Finally, I dumped some functions in `html_stat_map.py` which should probably live elsewhere. But I think it is time to get feedback, and in particular I'd like to know if there is an interest in merging this PR at all...

PR #1831, opened 2018-10-21
cmap starts from vmin when symmetric_cmap=False

jeromedockes:
when a non-negative image is given and the kwarg symmetric_cmap=False is given, plotting functions that rely on plotting.js_plotting_utils.colorscale, such as view_surf and view_stat_map, don't use a symmetric colormap.

as pointed out by @pbellec  , should the colormap then start from 0 or the min of the image?

this uses the min of the image by default and allows the user to specify a value in view_stat_map
when symmetric_cmap is True, vmin is -vmax
when the image is thresholded and symmetric_cmap is false, vmin is 0


bthirion:
LGTM.

nilearn/plotting/js_plotting_utils.py
```diff
@@ -196,13 +196,11 @@ def colorscale(cmap, values, threshold=None, symmetric_cmap=True, vmax=None):
                       'setting symmetric_cmap to True')
         symmetric_cmap = True
     if vmax is None:
-        if symmetric_cmap:
```
KamalakerDadi: But, aren't these routines compatible with other plotting functions ?

nilearn/plotting/js_plotting_utils.py
```diff
@@ -196,13 +196,11 @@ def colorscale(cmap, values, threshold=None, symmetric_cmap=True, vmax=None):
...
-            vmax = abs_values.max()
-            vmin = - vmax
-        else:
-            vmin, vmax = values.min(), values.max()
+        vmax = abs_values.max()
+    if symmetric_cmap:
+        vmin = - vmax
     else:
-        vmin = -vmax if symmetric_cmap else 0
+        vmin = values.min()
```
GaelVaroquaux: In general, I prefer the option of passing an explicit vmin argument, as this is how it's done in matplotlib.

I am not sure what is the state of this in nilearn plotting: I suspect that we have uneven support for vmin, but that might be a usability bug (I think that I was fighting with it a couple days ago, actually). We should fix it and provide a good support for vmin.

I vote for at least testing for the presence of vmin in the kwargs of the calling function, honoring it if symetric_cmap is false, and not honoring it is symetric_cmap is true, but warning about it.

nilearn/plotting/js_plotting_utils.py
GaelVaroquaux: I think that @KamalakerDadi and myself are getting to the same point there: we need to make this as much consistent as possible as possible across plotting functions, maybe by sharing this outside of the js plotting file.

jeromedockes: agreed

nilearn/plotting/js_plotting_utils.py
jeromedockes: https://github.com/nilearn/nilearn/pull/1766#issuecomment-431857095

nilearn/plotting/js_plotting_utils.py
```diff
@@ -201,6 +201,11 @@ def colorscale(cmap, values, threshold=None, symmetric_cmap=True,
         symmetric_cmap = True
     if symmetric_cmap and vmin is not None:
         warnings.warn('vmin cannot be chosen when cmap is symmetric')
+        vmin = None
+    if threshold is not None:
+        if vmin is not None:
+            warnings.warn('setting both vmin and a threshold is not allowed')
```
kchawla-pi: In the warning add `setting vmin to zero`.
You can use `and` instead of 2 `if` statements.

kchawla-pi:
I have been getting a lot of file coudn't be read for *.png files. We should investigate and fix this uptick. Ideas?

nilearn/plotting/js_plotting_utils.py
jeromedockes: >  You can use `and` instead of 2 `if` statements.

no because vmin is cleared whatever the second if evaluates to

jeromedockes:
> I have been getting a lot of file coudn't be read for *.png files. We should investigate and fix this uptick. Ideas?

is it related to this PR? where does the problem occur? can you provide a snippet to reproduce? thanks!

kchawla-pi:
In addition to the legitimate fails, I encountered a bunch of these in the release+brainsprite-cleanup PR, and my original [WIP] PR.
I don't think they are related to your PR changes (or mine) they tend to disappear on the next rebuild and I usually are unrelated to the changes made.

nilearn/plotting/js_plotting_utils.py
kchawla-pi: Tthat's why I am suggesting this If the user has not set `vmin` to anything they will not expect it have a value. If they have set both `threshold` and `vmin` they'll expect `vmin` to retain that value. When we are warning them that they shouldn't set both values, then we should also make it clear to them that the `vmin` value is being set to zero so there is no confusion to users old and new.

kchawla-pi:
Now appveyor has disappeared. @jeromedockes You have something else to push? it may reappear.

jeromedockes:
>  @jeromedockes You have something else to push?

I don't think so

kchawla-pi:
@GaelVaroquaux Only the organization owner can trigger new appveyor builds, so I can't do it. Would you do it?
@jeromedockes Maybe push a spurious commit, add a blank line or something.

nilearn/plotting/js_plotting_utils.py
```diff
@@ -189,23 +189,30 @@ def remove_temp_file(self):
         self._temp_file = None
 
 
-def colorscale(cmap, values, threshold=None, symmetric_cmap=True, vmax=None):
+def colorscale(cmap, values, threshold=None, symmetric_cmap=True,
+               vmax=None, vmin=None):
```
kchawla-pi: We should complete the docstring here with parameters and return data.
Also gives the new push we need for triggering Appveyor.

GaelVaroquaux:
> @GaelVaroquaux Only the organization owner can trigger new appveyor builds, so I can't do it. Would you do it?

I don't know how to handle appveyor. I have never managed to control its
builds.

> @jeromedockes Maybe push a spurious commit, add a blank line or something.

One option is to use "git amend", changing something trivial.

kchawla-pi:
Merging once green.

Issue #1792, opened 2018-10-12
Ship compressed versions of the Allen atlases


GaelVaroquaux:
The Allen atlases that we ship are not compressed. As a result, they use a total of 75M on a download (eg circleCI). We should reship them using OSF, after a "gzip -9", this drops the disk usage to 31Mb.

GaelVaroquaux:
@kchawla-pi : do you think that you can tackle this in the near future? It should make our circleCI builds a bit faster and a bit more robust, hence streamlining the release.

GaelVaroquaux:
Thanks!

kchawla-pi:
What do you think, how far along are we towards the release?

GaelVaroquaux:
> What do you think, how far along are we towards the release?

Almost there.

kchawla-pi:
Oh wait, the files inside are .nii files, they are not compressed at all. Nevermind

GaelVaroquaux:
The goal is to ship .nii.gz so that less disk space is used.

⁣Sent from my phone. Please forgive typos and briefness.​

kchawla-pi:
@GaelVaroquaux @bthirion 
I have created a project on OSF to store the data in a manner accessible to the fetchers, on the basis of other osf datasets we are using in the fetcher.
Before I proceed, take a look at it and make sure everything is in the clear, legally & scientific etiquette-wise.

kchawla-pi:
Can't seem to find anything about legal rights regarding the data, the paper itself seems open access. Certainly nothing about rehosting the data.

kchawla-pi:
Should I write them an email and clarify?

GaelVaroquaux:
> Should I write them an email and clarify?

I would just go ahead.

kchawla-pi:
I have uploaded the data as a project and made it public, now figuring out the url.

kchawla-pi:
Figured it out. Will require more work than I thought, no worries.

kchawla-pi:
@GaelVaroquaux Review & Discussion nescessary for #1840

Issue #1769, opened 2018-10-02
 Code refactoring necessary for future ease of maintenance and development


kchawla-pi:
Also for Nistats: https://github.com/nistats/nistats/issues/239
There are many functions in Nilearn that are super long, some as much as ~100 lines and do multiple things within the same function body.
There are many difficulties with such an approach. It makes it impossible to do Test-Driven Development, the recommended way of programming. It also means new functions have to work around existing functionality, which can not be segregated from existing functions due to this reason.

In the long-term, this code will become increasingly fragile and difficult to maintain. It also sends the message to new contributors and for contributions that this is an acceptable way to write new code.

As a rule of software engineering, each function should do one, and ONLY one thing. In python, this usually means 5-7 lines in a function. So instead of a 100 line function, we should (for example), have 20 functions around 5 lines each. All called by an encompassing function.

This also ensures we can write more atomic unit tests for each of those small functions.

I intend to start refactoring this code. This is a massive undertaking, so I will do this slowly and piecemeal, while adding more atomic unit tests.

This is in line with industry standard clean code principles for improving code quality and something I intend to do simultaneously where ever I see code that is difficult to comprehend.

tsalo:
Since this is a duplicate of #2383 (now that nistats has been merged in), can we close it?

jeromedockes:
sure! Also, while I agree that very long functions are often harder to maintain and more likely to contain bugs, I think we should be wary of making big changes to code that is working well for the sake of style. there is a very real risk of introducing bugs, and also of spending a lot of time to end up with something that is not that much better than the original version. This is especially true for older parts of the codebase. They sometimes look a bit more messy, but they have been in the hands of users who file bug reports for a long time, and that is something very valuable that unit tests, while very useful, cannot replace. Rather than doing it systematically, I am more in favor of doing the refactoring when necessary, for code that needs to be changed anyway when introducing new features (or fixing bugs), for example as @NicolasGensollen  did when adding the plotly backend to surface plotting functions.

tsalo:
That's a really good point. Perhaps we can add something to the contributing documentation about writing new code in small functions, but avoiding large refactors for the sake of enforcing this style?

jeromedockes:
> As a rule of software engineering, each function should do one, and ONLY one thing. In python, this usually means 5-7 lines in a function. So instead of a 100 line function, we should (for example), have 20 functions around 5 lines each. All called by an encompassing function.

[...]

PR #1835, opened 2018-10-22
[WIP] Refactor _fetch_file(s)

kchawla-pi:
While working on #1792 I found it painful & slow to read the _fetch_files() & _fetch_file_().
Additionally, the new URLs on OSF may necessitate changing how the passed in URLs are handled. 

As per issue #1769, I have refactored the function into several smaller ones, which have made it considerably easier for me to comprehend it, and modify it if necessary.

The logic has not been altered to maintain Python2 compatibility & make the refactoring quick, safe, and easy.
This is in line with industry standard clean code principles for improving code quality and something I intend to do simultaneously where ever I see code that is difficult to comprehend.

### DIFF
--- a/nilearn/datasets/utils.py
+++ b/nilearn/datasets/utils.py
@@ -711,8 +711,10 @@ def _fetch_files(data_dir, files, resume=True, mock=False, verbose=1):
         temp_target_file = os.path.join(temp_dir, file_)
         # Whether to keep existing files
         overwrite = opts.get('overwrite', False)
-        if (abort is None and (overwrite or (not os.path.exists(target_file) and not
-                os.path.exists(temp_target_file)))):
+        file_does_not_exist = _file_does_not_exist(target_file,
+                                                   temp_target_file,
+                                                   )
+        if not abort and (file_does_not_exist or overwrite):
 
             # We may be in a global read-only repository. If so, we cannot
             # download files.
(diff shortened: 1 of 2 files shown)


nilearn/datasets/utils.py
```diff
@@ -774,6 +765,49 @@ def _fetch_files(data_dir, files, resume=True, mock=False, verbose=1):
     return files_
 
 
+def _download_or_mock_file(abort, dl_file, file_, mock, target_file,
+                           temp_target_file, url):
+    if mock:
```
bthirion: Could you add a one-line function description to explain in a nutshell what the function does ?

nilearn/datasets/utils.py
```diff
@@ -774,6 +765,49 @@ def _fetch_files(data_dir, files, resume=True, mock=False, verbose=1):
...
+        if not mock or os.path.getsize(dl_file) != 0:
+            _uncompress_file(dl_file, verbose=verbose)
+        else:
+            os.remove(dl_file)
+    except Exception as e:
+        abort = str(e)
+    return abort
+
+
+def _move_dataset_files(dl_file, opts, temp_dir):
```
bthirion: (same for other private functions)

nilearn/datasets/utils.py
```diff
@@ -774,6 +766,85 @@ def _fetch_files(data_dir, files, resume=True, mock=False, verbose=1):
...
+    return abort
+
+
+def _check_files_existence(file_1, file_2):
+    """ Returns True if both the provided files exist.
+    """
+    target_file_does_not_exist = not os.path.exists(file_1)
+    temp_target_file_does_not_exist = not os.path.exists(file_2)
+    file_does_not_exist = target_file_does_not_exist and temp_target_file_does_not_exist
+    return file_does_not_exist
```
GaelVaroquaux: 3-liner functions are in general bad practice. The reason being that when doing a careful read of the codebase, one needs to read them to know what they actually do (the name is not a contract, and bugs hide this).

I am not very enthousiastic about merging changes that turn 3 lines of code used twice into a function. It makes the code harder to read and maintain.

nilearn/datasets/utils.py
```diff
@@ -774,6 +766,85 @@ def _fetch_files(data_dir, files, resume=True, mock=False, verbose=1):
...
+    None or String(exception)
+    """
+    try:
+        if not mock or os.path.getsize(file_) != 0:
+            _uncompress_file(file_, verbose=verbose)
+        else:
+            os.remove(file_)
+    except Exception as e:
+        abort = str(e)
+    return abort
```
GaelVaroquaux: This function is used only once, and understanding what it does is as difficult from the code as from the function name and description. It should not be factored out.

nilearn/datasets/utils.py
```diff
@@ -774,6 +766,85 @@ def _fetch_files(data_dir, files, resume=True, mock=False, verbose=1):
...
+        path to the destination root directory
+        
+    dst_sub_dir: string
+        Name or path for the subdirectory within the destination root directory
+        
+    Returns
+    -------
+    dst_filepath: string
+        Path to the moved file.
+    """
```
GaelVaroquaux: Another function that is very simple and used once.

GaelVaroquaux:
@kchawla-pi : I am really sorry, but these changes do not make the codebase easier to maintain, rather harder. I've worked on many codebases, and codebases that are structured like this (with many tiny functions used once or twice) are very hard to follow: they require jumping from one function to another all the time, and keeping in one's working memory what the functions do.

I am not in favor of merging these changes. Let's try to discuss this IRL.

Issue #1828, opened 2018-10-20
`nilearn.signal.clean` order might not be as described


miykael:
The [description](https://github.com/nilearn/nilearn/blob/master/nilearn/signal.py#L362) of `nilearn.signal.clean` states that the signal is cleaned in the following order:
    - detrend
    - standardize
    - remove confounds
    - low- and high-pass filter

But if I understand the code correctly then this order is wrongly documented. `standardize` should be described as the last step, as shows the code [here](https://github.com/nilearn/nilearn/blob/master/nilearn/signal.py#L537).

I'm happy to write a quick PR about this, but first wanted to make sure that this needs correction (in the documentation?).

GaelVaroquaux:
I think that the issue is deeper than that as pointed out by the [neurostar](https://neurostars.org/t/nilearn-signal-clean/2690) question that (I believe) triggered your comment: indeed, the documentation is wrong, but also the way things are done is not correct: as recommended by the [Linquist paper](https://www.biorxiv.org/content/biorxiv/early/2018/09/04/407676.full.pdf), we should make sure that at each step, the filter applied to the data is orthogonal to the previous filters (where confounds are understood as filters).

I think that the right solution is to do the following (as recommended by the Linquist paper):

1. Band-pass filter the data (if filtering is specified)
2. Band-pass filter the confounds
3. Remove the confounds.

@bthirion : do you agree with my analysis?

miykael:
Considering the Linquist paper and this issue https://github.com/nilearn/nilearn/issues/1011, would it make sense to replace band-pass filters with cosine filters (DCT bases)? By putting confounds, cosine filters and linear trend into one matrix, we could filter all at once, as suggested by the Linquist paper, correct?

GaelVaroquaux:
> Considering the Linquist paper and this issue #1011, would it make sense to
> replace band-pass filters with cosine filters (DCT bases)?

It would be something interesting to add cosine filter, but I would like
to keep to option of having a Butterworth filter, which has good
signal-processing properties.

> By putting confounds, cosine filters and linear trend into one matrix,
> we could filter all at once, as suggested by the Linquist paper,
> correct?

Linquist suggests several options. His most important message is that one
step should not undo what the previous did.

bthirion:
DCT basis is definitely a good option for high-pass filtering.

imawla:
[...]

PR #1837, opened 2018-10-22
[MRG+1] ENH: Corrects order and behavior of filters in `nilearn.signal.clean`

miykael:
This PR changes the order of the cleaning function, as discussed in https://github.com/nilearn/nilearn/issues/1828.

As a new feature, confound removal will be done orthogonally to temporal filters (according to [Lindquist et al. (2018)](https://www.biorxiv.org/content/biorxiv/early/2018/09/04/407676.full.pdf)), if both are specified.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -8,11 +8,11 @@ Changes
 - Default value of t_r in signal.clean and clean_img is changed from 2.5 to
   None. If low_pass or high_pass is specified, then t_r needs to be specified
   as well otherwise it will raise an error.
-- Order of filters in signal.clean and clean_img is changed to detrend, low- and
-  high-pass filter, remove confounds and standardize. To ensure orthogonality
-  between temporal filter and confound removal, an additional temporal filter
-  will be applied on the confounds before removing them. This is according to
-  `Lindquist et al. (2018)<http://dx.doi.org/10.1101/407676>`_.
+- Order of filters in signal.clean and clean_img has changed to detrend, low-
+  and high-pass filter, remove confounds and standardize. To ensure
+  orthogonality between temporal filter and confound removal, an additional
+  temporal filter will be applied on the confounds before removing them. This
+  is according to Lindquist et al. (2018).
 
 
 Contributors
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -1,3 +1,28 @@
+0.5.0
+=====
+
+
+Changes
+-------
+
+- Default value of t_r in signal.clean and clean_img is changed from 2.5 to
+  None. If low_pass or high_pass is specified, then t_r needs to be specified
+  as well otherwise it will raise an error.
+- Order of filters in signal.clean and clean_img is changed to detrend, low- and
+  high-pass filter, remove confounds and standardize. To ensure orthogonality
+  between temporal filter and confound removal, an additional temporal filter
+  will be applied on the confounds before removing them. This is according to
+  `Lindquist et al. (2018)<http://dx.doi.org/10.1101/407676>`_.
+
+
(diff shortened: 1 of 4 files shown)


bthirion:
Also, I don't see travis/circleci output. Am I missing something ?

miykael:
I think circleci and travis are coming back.

About the test. How should I best implement a test that checks if confound and band-pass filters work correctly?
My inital approach would be: Clean signal with confound and high pass specified and then check the power spectrum for the filtered frequency. If the filters are not orthogonal, then some filtered frequency should be reintroduced into the data. But I'm not sure if this is the smartest approach.

GaelVaroquaux:
I think that it is important that this change is mentioned in the whats_new.

nilearn/signal.py
```diff
@@ -508,11 +517,28 @@ def clean(signals, sessions=None, detrend=True, standardize=True,
...
         confounds = _ensure_float(confounds)
         confounds = _standardize(confounds, normalize=standardize,
                                  detrend=detrend)
+
+        # Apply low- and high-pass filters to keep filters orthogonal
+        # (according to Lindquist et al. (2018))
+        if low_pass is not None or high_pass is not None:
+
+            confounds = butterworth(confounds, sampling_rate=1. / t_r,
+                                    low_pass=low_pass, high_pass=high_pass)
```
GaelVaroquaux: We should filter before standardizing the confounds: the standardizing of confounds is useful for numerical stability.

GaelVaroquaux:
Overall looks great aside two small comments.

miykael:
Moved filtering before standardizing and added changes to whats_new.

GaelVaroquaux:
Some of your jobs had failed for no good reasons (Internet flukes on the CI). I have restarted them.

GaelVaroquaux:
> About the test. How should I best implement a test that checks if confound and band-pass filters work correctly?
> My inital approach would be: Clean signal with confound and high pass specified and then check the power spectrum for the filtered frequency. If the filters are not orthogonal, then some filtered frequency should be reintroduced into the data. But I'm not sure if this is the smartest approach.

I think that this is a good idea: it should be a reasonably simple test that checks that the code does what it should.

GaelVaroquaux:
Also, as you add the test, can you merge with master, to resolve the conflicts that I added when merging your other branch.

Thanks a lot!

miykael:
I've resolved the conflicts and added the tests (https://github.com/nilearn/nilearn/pull/1837/commits/7b445badebb0577ddbd420439020922e2bd02150). A detailed explanation/description of the tests can be found [here](https://gist.github.com/miykael/4bb887029cba3926902feb2e4afe190f).

doc/whats_new.rst
```diff
@@ -5,14 +5,23 @@
...
+  is according to Lindquist et al. (2018).
+
 
 Contributors
 -------------
 
 The following people contributed to this release::
 
-  1  Michael Notter
+  2  Michael Notter
```
GaelVaroquaux: I wouldn't worry about keeping this up to date: it's something that we do when we do the release.

GaelVaroquaux:
LGTM. +1 for merge.

@bthirion : do you want to have another look?

KamalakerDadi:
Is it such a normal procedure to do that we don't worry about documenting the details ?
For instance, in doc/manipulating_images/manipulating_images.rst

GaelVaroquaux:
I don't think that we should document the details in another place than the function docstring, because the danger is that as the function evolves (for instance with addition of the DCT based filtering) the other documentation is not kept up to date.

GaelVaroquaux:
Thanks heaps. Merging!

PR #1840, opened 2018-10-24
[WIP] fetch_atlas_allen_2011() downloads compressed dataset from OSF server

kchawla-pi:
For issue #1792 
Functionality lost: `resume`
I found it too onerous to adapt the new URL as per _fetch_files().
The download is ~31MB so I figure this is not a big deal.

**The tests will break**, I believe as they must be re-adapted according to the new code.
I am pushing this code here for review & discussion so I do not modify the tests in error.

Additionally, why do we have a `url` parameter? Do we expect users to give their own URL here and expect it to work seamlessly with the function? Should we remove it?

### DIFF
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -8,8 +8,6 @@ from tempfile import mkdtemp
 import json
 import shutil
 
-from collections import OrderedDict
-
 import nibabel as nb
 import numpy as np
 from sklearn.datasets.base import Bunch
@@ -862,18 +860,14 @@ def fetch_atlas_allen_2011(data_dir=None, url=None, resume=True,
     url = url if url else 'https://osf.io/hrcku/download'
     dataset_name = "allen_rsn_2011"
     data_dir = get_data_dirs(data_dir=data_dir)[0]
-    dataset_components_filenames = OrderedDict(
-            {'maps': 'ALL_HC_unthresholded_tmaps.nii.gz',
-             'rsn28': 'RSN_HC_unthresholded_tmaps.nii.gz',
-             'comps': 'rest_hcp_agg__component_ica_.nii.gz',
-             }
-            )
-    expected_files = OrderedDict(
-            {component: os.path.join(data_dir, dataset_name, file_)
-             for component, file_
-             in dataset_components_filenames.items()
-             }
-            )
+    dataset_components_filenames = {'maps': 'ALL_HC_unthresholded_tmaps.nii.gz',
+                                    'rsn28': 'RSN_HC_unthresholded_tmaps.nii.gz',
+                                    'comps': 'rest_hcp_agg__component_ica_.nii.gz',
+                                    }
+    expected_files = {component: os.path.join(data_dir, dataset_name, file_)
+                      for component, file_
+                      in dataset_components_filenames.items()
+                      }
     expected_files_present = all([os.path.isfile(filepath)
                                   for filepath in expected_files.values()
                                   ])
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -7,6 +7,8 @@ import xml.etree.ElementTree
 from tempfile import mkdtemp
 import json
 import shutil
+import urllib2
+
 
 import nibabel as nb
 import numpy as np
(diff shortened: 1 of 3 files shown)


bthirion:
The changes LGTM.
Regarding the `url` parameter, dunno, I haven't worked  a lot on this part of the code base.

kchawla-pi:
@GaelVaroquaux I might switch to rewriting a new/additional tests in Pytest, and then add that to the CI infrastructure.

nilearn/datasets/atlas.py
```diff
@@ -808,7 +821,8 @@ def fetch_coords_dosenbach_2010(ordered_regions=True):
     return Bunch(**params)
 
 
-def fetch_atlas_allen_2011(data_dir=None, url=None, resume=True, verbose=1):
+def fetch_atlas_allen_2011(data_dir=None,
+                               verbose=1):
```
GaelVaroquaux: Good practice is to not do chances in a PR that are cosmetic outside of the code that we modify. The reason is that it creates a cognitive overhead reviewing the PR and that it is a recipe for merge conflicts.

nilearn/datasets/atlas.py
```diff
@@ -850,44 +860,83 @@ def fetch_atlas_allen_2011(data_dir=None, url=None, resume=True, verbose=1):
...
+            raise url_excep
+        except (OSError, IOError) as os_excep:
+            if os_excep.strerror == 'Permission denied':
+                error_msg = ('\nERROR: You do not have permission '
+                             'to create and/or modify files in "{}".\n'
+                             'Choose a different download directory '
+                             'or talk to your system administrator.\n'
+                             'Download aborted.'.format(data_dir)
+                             )
+                print(error_msg)
```
GaelVaroquaux: Printing an error message and raising the error afterward is not very efficient in terms of user experience, because a long traceback appears below the interesting message, and the user tends to miss it.

nilearn/datasets/atlas.py
```diff
@@ -850,44 +860,83 @@ def fetch_atlas_allen_2011(data_dir=None, url=None, resume=True, verbose=1):
...
+                  )
+        try:
+            downloaded_filepath = _fetch_file(url, data_dir, verbose=verbose)
+        except (URLError, HTTPError) as url_excep:
+            err_msg = ('\nERROR: There was a problem downloading the file '
+                       '"allen_rsn_2011.zip". \nPlease verify you have '
+                       'a working internet connection and '
+                       'the file is still present at {}'.format(url))
+            print(err_msg)
+            raise url_excep
```
GaelVaroquaux: Printing an error message and raising the error afterward is not very efficient in terms of user experience, because a long traceback appears below the interesting message, and the user tends to miss it.

kchawla-pi:
I believe the earlier failures in Travis and appveyor (circleci was green) were because the tests expect the URL to have the filename in the end, and that is no longer the case. I will work on this tomorrow.

GaelVaroquaux:
OK, glad to hear that you have a lead: because with a quick look I didn't
have one. Good luck with that one!

One precision on the test: it probably "mocks" the download, aka it
avoids actually doing an Internet request (using monkey patching),
because this is too fragile.

kchawla-pi:
@Gael, my code is too fragile or the monkey-patching is too fragile?

GaelVaroquaux:
> @Gael, my code is too fragile or the monkey-patching is too fragile?

I don't know, honestly :$

nilearn/datasets/atlas.py
kchawla-pi: I added new imports here, so figured this would be a decent point to do so.

nilearn/datasets/atlas.py
kchawla-pi: I wanted the traceback in case of errors. I will replace the raise with sys.exit() or quit() once the feature is ready.

kchawla-pi: or maybe return?

GaelVaroquaux:
> I will replace the raise with sys.exit() or quit() once the feature is ready.

Never do that: it breaks the usage of the library as error capture is no
longer possible.

The right solution is to raise a new traceback with an informative
message.

kchawla-pi:
The problem seems to be in the nosetest fixtures. I have removed them and using standard python for the tempdir setup and things are fine.

kchawla-pi:
I will convert the`url` parameter to `_url`, remove it from the docstring, and use it to test URL & HTTP errors by passing in wrong server URLs for testing.

kchawla-pi:
@GaelVaroquaux
I fixed the error in my code, however
 @KamalakerDadi Made it work with the existing code, only a URL change, so I will close this one.

Issue #1839, opened 2018-10-24
Could `image.clean_img` reduce the memory load by using a mask?


miykael:
I have a rather big functional image of dimensions 90 x 120 x 85 x 1055, with `float64` datatype. About 70% of this image is background (i.e. `zeros`). When I want to run `image.clean_img` on this image, my memory load goes up to 15GB.

This is due to the usage of `get_data()`, [here](https://github.com/nilearn/nilearn/blob/master/nilearn/image/image.py#L906) and that we feed all voxels into `signal.clean`. Wouldn't it make sense to introduce additional mask and unmasking steps to compute only on nonempty voxels?

What would the quickest/best way be to create such a mask on the fly? Or should it just be possible to provide a mask to restrict the cleaning?

PS: Such an approach might help other functions as well.

bthirion:
+1 on my side.

GaelVaroquaux:
There are two options possible:

* explicit masking, given as an optional argument
* implicit masking that needs to be guessed.

I think that we should do one, or the other, but not both.

For the second option, I think that we need to have a very simple and non
ambitious strategy, for fear of introducing "magic", ie, behavior that is
not well understood by the user. I think that strategy could be an
argument that list values to be ignored, by default [0, np.nan]. That
should hopefully cover 95% of the use cases.

bthirion:
I think I prefer explicit masking.

miykael:
I think I also prefer explicit masking. I realized that most functional images coming directly from the scanner always have some background noise, therefore `[0, np.nan]` will catch almost nothing in this case.

And as it is just to reduce memory load, an explicit mask could point users to this option, but this is not essential for low-memory-load datasets (and therefore not required).

From the top of your head, are there any other functions where I could implement explicit masking as well?

GaelVaroquaux:
> From the top of your head, are there any other functions where I could
> implement explicit masking as well?

I don't know. Maybe math_img. The reason that I don't know is that I
typically do these operations after after a NiftiMasker: when chaining
operations it is much more efficient to mask only once, and unmask
after.

miykael:
I've created now a new PR for `clean_img`. And yes, you're approach makes sense in a normal environment. I'm currently working in a Nipype environment where everything is modular, so I try to profit from additional input parameters.

PR #1842, opened 2018-10-25
[MRG] ENH: Adds mask parameter to image.clean_img

miykael:
Closes https://github.com/nilearn/nilearn/issues/1839#issuecomment-432988036.

As discussed in https://github.com/nilearn/nilearn/issues/1839#issuecomment-432988036, this PR adds a `mask_img` parameter to `clean.img` to restrict the computation to voxels inside the mask.

I struggled a bit with the header information in the mask condition, [here](https://github.com/nilearn/nilearn/compare/master...miykael:add_mask_to_clean_img?expand=1#diff-b61cedb53334db86de8881757209d3feR921). I solved it by unmasking the image and then extracting the data. Because if we would just save the unmasked image, we wouldn't have the TR (temporal resolution) and probably other header information.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -5,16 +5,17 @@
 Changes
 -------
 
-- Default value of t_r in signal.clean and clean_img is changed from 2.5 to
-  None. If low_pass or high_pass is specified, then t_r needs to be specified
-  as well otherwise it will raise an error.
-- Order of filters in signal.clean and clean_img has changed to detrend, low-
-  and high-pass filter, remove confounds and standardize. To ensure
-  orthogonality between temporal filter and confound removal, an additional
-  temporal filter will be applied on the confounds before removing them. This
-  is according to Lindquist et al. (2018).
-- image.clean_img now accepts a mask to restrict the cleaning of the image.
-  This approach can help to reduce the memory load.
+- Default value of `t_r` in :func:nilearn.signal.clean: and
+  :func:nilearn.image.clean_img: is changed from 2.5 to None. If `low_pass` or
+  `high_pass` is specified, then `t_r` needs to be specified as well otherwise
+  it will raise an error.
+- Order of filters in :func:nilearn.signal.clean: and :func:nilearn.image.clean_img:
+  has changed to detrend, low- and high-pass filter, remove confounds and
+  standardize. To ensure orthogonality between temporal filter and confound
+  removal, an additional temporal filter will be applied on the confounds before
+  removing them. This is according to Lindquist et al. (2018).
+- :func:nilearn.image.clean_img: now accepts a mask to restrict the cleaning of
+  the image. This approach can help to reduce the memory load and computation time.
 
 
 Contributors
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -13,6 +13,8 @@ Changes
   orthogonality between temporal filter and confound removal, an additional
   temporal filter will be applied on the confounds before removing them. This
   is according to Lindquist et al. (2018).
+- image.clean_img now accepts a mask to restrict the cleaning of the image.
+  This approach can help to reduce the memory load.
 
 
 Contributors
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -5,16 +5,16 @@
 Changes
 -------
 
-- Default value of `t_r` in :func:nilearn.signal.clean: and
-  :func:nilearn.image.clean_img: is changed from 2.5 to None. If `low_pass` or
+- Default value of `t_r` in :func:`nilearn.signal.clean` and
+  :func:`nilearn.image.clean_img` is changed from 2.5 to None. If `low_pass` or
   `high_pass` is specified, then `t_r` needs to be specified as well otherwise
   it will raise an error.
-- Order of filters in :func:nilearn.signal.clean: and :func:nilearn.image.clean_img:
+- Order of filters in :func:`nilearn.signal.clean` and :func:`nilearn.image.clean_img`
   has changed to detrend, low- and high-pass filter, remove confounds and
   standardize. To ensure orthogonality between temporal filter and confound
   removal, an additional temporal filter will be applied on the confounds before
   removing them. This is according to Lindquist et al. (2018).
-- :func:nilearn.image.clean_img: now accepts a mask to restrict the cleaning of
+- :func:`nilearn.image.clean_img` now accepts a mask to restrict the cleaning of
   the image. This approach can help to reduce the memory load and computation time.
 
 
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -915,10 +915,9 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
         't_r': t_r, 'ensure_finite': ensure_finite}
 
     if mask_img is not None:
-        mask_img_ = check_niimg_3d(mask_img)
-        sigs = masking.apply_mask(imgs_, mask_img_)
+        sigs = masking.apply_mask(imgs_, mask_img)
         sigs_clean = signal.clean(sigs, **clean_paramters)
-        data = masking.unmask(sigs_clean, mask_img_).get_data()
+        data = masking.unmask(sigs_clean, mask_img).get_data()
     else:
         sigs = imgs_.get_data().reshape(-1, imgs_.shape[-1]).T
         data = signal.clean(sigs, **clean_paramters).T.reshape(imgs_.shape)
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -865,7 +865,8 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
 
     mask_img: Niimg-like object, optional
         See http://nilearn.github.io/manipulating_images/input_output.html
-        If provided, signal is only cleaned from voxels inside the mask.
+        If provided, signal is only cleaned from voxels inside the mask. If
+        mask is provided, it should have same shape and affine as imgs.
         If not provided, all voxels are used.
 
     Returns
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -863,7 +863,7 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
         If True, the non-finite values (NaNs and infs) found in the images
         will be replaced by zeros.
 
-    mask_img: Niimg-like object
+    mask_img: Niimg-like object, optional
         See http://nilearn.github.io/manipulating_images/input_output.html
         If provided, signal is only cleaned from voxels inside the mask.
         If not provided, all voxels are used.
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -801,7 +801,7 @@ def math_img(formula, **imgs):
 
 def clean_img(imgs, sessions=None, detrend=True, standardize=True,
               confounds=None, low_pass=None, high_pass=None, t_r=None,
-              ensure_finite=False):
+              ensure_finite=False, mask_img=None):
     """Improve SNR on masked fMRI signals.
 
     This function can do several things on the input signals, in
@@ -863,6 +863,11 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
         If True, the non-finite values (NaNs and infs) found in the images
         will be replaced by zeros.
 
+    mask_img: Niimg-like object
+        See http://nilearn.github.io/manipulating_images/input_output.html
+        If provided, signal is only cleaned from voxels inside the mask.
+        If not provided, all voxels are used.
+
     Returns
     -------
     cleaned_img: Niimg-like object
@@ -888,6 +893,7 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
     """
     # Avoid circular import
     from .image import new_img_like
+    from .. import masking
 
     imgs_ = check_niimg_4d(imgs)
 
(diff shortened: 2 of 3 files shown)


nilearn/image/image.py
```diff
@@ -902,12 +908,22 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
                 "specified None. imgs header suggest it to be {0}".format(
                     imgs.header.get_zooms()[3]))
 
-    data = signal.clean(
-        imgs_.get_data().reshape(-1, imgs_.shape[-1]).T, sessions=sessions,
-        detrend=detrend, standardize=standardize, confounds=confounds,
-        low_pass=low_pass, high_pass=high_pass, t_r=t_r,
-        ensure_finite=ensure_finite).T.reshape(imgs_.shape)
-    return new_img_like(imgs, data, copy_header=True)
+    # Store function parameters in dictionary for easy access
```
KamalakerDadi: I think it would make easier if we would stick to mask_img implementation rather than cleaning the API for easy access.

nilearn/image/image.py
```diff
@@ -902,12 +908,22 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
...
+    clean_paramters = {
+        'sessions': sessions, 'detrend': detrend, 'standardize': standardize,
+        'confounds': confounds, 'low_pass': low_pass, 'high_pass': high_pass,
+        't_r': t_r, 'ensure_finite': ensure_finite}
+
+    if mask_img is not None:
+        mask_img_ = check_niimg_3d(mask_img)
```
KamalakerDadi: We append this name "_" to return them as an attribute, for instance in class/object. I would rather keep it to mask_img.

miykael: Yes, I was copying from `imgs_`.

nilearn/image/tests/test_image.py
```diff
@@ -582,6 +582,10 @@ def test_clean_img():
     data_img_nifti2_ = image.clean_img(
         data_img_nifti2, detrend=True, standardize=False, low_pass=0.1, t_r=1.0)
 
+    # if mask_img
+    img, mask_img = data_gen.generate_fake_fmri(shape=(10, 10, 10), length=10)
+    data_img_mask_ = image.clean_img(img, mask_img=mask_img)
```
bthirion: You may add some checks on the result ?

nilearn/image/image.py
miykael: Not sure what you mean by that. Having a mask as default? One could imagine `mask=np.ones(imgs_.shape[:3])`. But the missing header issue from `unmask` would persist. But could be solved with a `get/set_zooms`.

KamalakerDadi:
With providing mask_img, you have a reduction in memory consumption ? If yes, how much ?

nilearn/image/image.py
```diff
@@ -902,12 +908,22 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
...
                     imgs.header.get_zooms()[3]))
 
-    data = signal.clean(
-        imgs_.get_data().reshape(-1, imgs_.shape[-1]).T, sessions=sessions,
-        detrend=detrend, standardize=standardize, confounds=confounds,
-        low_pass=low_pass, high_pass=high_pass, t_r=t_r,
-        ensure_finite=ensure_finite).T.reshape(imgs_.shape)
-    return new_img_like(imgs, data, copy_header=True)
+    # Store function parameters in dictionary for easy access
+    clean_paramters = {
```
bthirion: I think I'd clean the parameters explicit in the two calls of the function.

nilearn/image/image.py
KamalakerDadi: Nope keeping code as it was and just modifying the first positional argument. For me, only thing which varies is first argument to signal.clean

nilearn/image/image.py
```diff
@@ -902,12 +908,22 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
...
+        'sessions': sessions, 'detrend': detrend, 'standardize': standardize,
+        'confounds': confounds, 'low_pass': low_pass, 'high_pass': high_pass,
+        't_r': t_r, 'ensure_finite': ensure_finite}
+
+    if mask_img is not None:
+        mask_img_ = check_niimg_3d(mask_img)
+        sigs = masking.apply_mask(imgs_, mask_img_)
+        sigs_clean = signal.clean(sigs, **clean_paramters)
+        data = masking.unmask(sigs_clean, mask_img_).get_data()
```
KamalakerDadi: When you use unmask, it is already nibabel image, no ?

nilearn/image/image.py
miykael: I thought about that. So we would have two possible `signals` as first argument? But this would mean to have two `if`s, which I wanted to prevent.
```python
if mask_img is not None:
    signals = masking.apply_mask(imgs_, mask_img_)
else:
    signals = imgs_.get_data().reshape(-1, imgs_.shape[-1]).T
data = signal.clean(signals,...)
if mask_img is not None:
    data = masking.unmask(sigs_clean, mask_img_).get_data()
return new_img_like(imgs_, data, copy_header=True)
```

nilearn/image/image.py
```diff
@@ -863,6 +863,11 @@ def clean_img(imgs, sessions=None, detrend=True, standardize=True,
         If True, the non-finite values (NaNs and infs) found in the images
         will be replaced by zeros.
 
+    mask_img: Niimg-like object
```
KamalakerDadi: Can you write that it is optional ?

Niimg-like object, optional

nilearn/image/tests/test_image.py
miykael: I didn't add any checks on the result, as the other tests in this section also didn't check for results. But I could verify that a `np.ones(...)` mask and no mask would get the same result?

miykael:
@KamalakerDadi - The reduction of memory consumption is mean for images where the bounding-box/FOV is much bigger than the actual data. For example, something like this:
example
Cropping could help but you would still compute on a lot of empty voxels. So in my particular case, I reduced computation time by 50%... Actually, you're right, memory might not decrease as much.

nilearn/image/image.py
miykael: Yes, unmask is nibabel image. I just used `get_data` to have it available for the `return` command. To make sure that the header is copied.

nilearn/image/image.py
miykael: Not sure if I understand you correctly. You mean, I should have them twice written out? As they are exactly the same, I thought to prevent redundancy.

nilearn/image/image.py
miykael: Done :+1:

nilearn/image/image.py
KamalakerDadi: apply_mask already has check_niimg_3d. May be it is not required.

miykael: Good to know. I removed this check.

KamalakerDadi:
We need make sure that provided mask_img and imgs have same affine, shape, etc. Do you think is that something missing or does not matter ?

KamalakerDadi:
May be when user provides mask_img explicitly, using copy_header=True may not be required ?
I am not exactly sure on this point. You can wait for others to comment on this point.

If it not required, then code can be simplified I think

KamalakerDadi:
May be its worth checking or add tests while not keeping same affine and shape, fov for mask_img and imgs.
Thanks

doc/whats_new.rst
```diff
@@ -13,6 +13,8 @@ Changes
   orthogonality between temporal filter and confound removal, an additional
   temporal filter will be applied on the confounds before removing them. This
   is according to Lindquist et al. (2018).
+- image.clean_img now accepts a mask to restrict the cleaning of the image.
```
KamalakerDadi: small tip, if you could do something like this ":func:`nilearn.image.clean_img`", then it is useful for user to directly click on this and go directly to documentation . You can for instance, look below how a function or class is referred.

nilearn/image/image.py
bthirion: > Not sure if I understand you correctly. You mean, I should have them twice written out? As they are exactly the same, I thought to prevent redundancy.

Sorry I'd just pass all the arguments explictily in the calls of the function. makes it easier to inspect tracebacks etc.

miykael:
To summarize what was discussed so far and what I did since last time:

I've added `:func:` to `whats_new.rst` (https://github.com/nilearn/nilearn/pull/1842/commits/a71b0611d4c0aa64b4761b8546b7f8b8956288b9) and added a test that checks that `clean_img` produces the same result, with or without a mask (mask of non-zero voxels) (https://github.com/nilearn/nilearn/pull/1842/commits/d7c5aa110726e8f18e69a32e6fc89e7b58e0beb1).

@bthirion - Should I still add tests that check the accuracy of the results? I think testing that with and without mask leads to equal results, should be enough for this PR.

@KamalakerDadi - I didn't implement a check for affine and shape between image and mask, as it also wasn't done in [`image.high_variance_confounds`](https://github.com/nilearn/nilearn/blob/master/nilearn/image/image.py#L28). Do you want me to add this (in both places)? Is there a `nilearn` function for this or do I just quickyl check?

Combining @bthirion's comment about explicit arguments for detailed traceback and @KamalakerDadi's about keeping the original implementation but only changing the first positional argument, I've changed the code (https://github.com/nilearn/nilearn/pull/1842/commits/fa46d145b5db93e7a922135ffda087abab2760de).

**Open Issue**: I haven't solved yet how to reset the `header` after unmasking. I'm concernt about TR, which could be handled with `get/set_zooms`, but I guess we should also make sure that the other header information are not lost.

KamalakerDadi:
>@KamalakerDadi - I didn't implement a check for affine and shape between image and mask, as it also wasn't done in image.high_variance_confounds. Do you want me to add this (in both places)?

I was asking if you consider it as an issue or does not matter ?

miykael:
I think the check is not needed here, but should definitely be done in `masking.apply_mask`. If the check fails there, it will also fail in `clean_img`.

kchawla-pi:
The TravisCI test failed for one particular config. I am restarting it to verify the failure is not spurious. Will take 6 min.

miykael:
Not sure what to think of travis' error `Arrays are not equal (mismatch 0.02%)` in python 2.7 and numpy 1.13. I don't think that restarting will help, happened already twice.
But to be sure, I changed `assert_equal` to `assert_almost_equal` (https://github.com/nilearn/nilearn/pull/1842/commits/33517ff2b032b796a5cb9264d6149826138969c3).

KamalakerDadi:
> I am restarting it to verify the failure is not spurious.

It is not spurious. To check whether it is spurious or not. You should look at the function where it is failing and changes made in the function from this PR.

KamalakerDadi:
>I didn't implement a check for affine and shape between image and mask,

@bthirion @GaelVaroquaux Do you think we should add a function which checks this ?

GaelVaroquaux:
> @bthirion @GaelVaroquaux Do you think we should add a function which checks this ?

Is what you are looking for:
nilearn/_utils/niimg_conversions.py:_check_fov
?

kchawla-pi:
> The TravisCI test failed for one particular config. I am restarting it to verify the failure is not spurious. Will take 6 min.

It Passed on the rerun.

KamalakerDadi:
> @bthirion @GaelVaroquaux Do you think we should add a function which checks this ?
Is what you are looking for:
nilearn/_utils/niimg_conversions.py:_check_fov
?

_check_same_fov , if not _check_same_fov then I would use resample_img to resample.

miykael:
> _check_same_fov , if not _check_same_fov then I would use resample_img to resample.

But this should be implemented in `masking.apply_mask`, no?

KamalakerDadi:
>    _check_same_fov , if not _check_same_fov then I would use resample_img to resample.

>But this should be implemented in masking.apply_mask, no?

Either we can do this in ```clean_img``` or leave it to ```apply_mask``` such that it raises an error when both are not in same shape and affine which it already has this implementation

I think we can conclude discussion on this by just adding a small sentence at mask_img documentation, saying that if provided it should be of same shape and affine.

doc/whats_new.rst
```diff
@@ -5,14 +5,17 @@
...
 
-- Default value of t_r in signal.clean and clean_img is changed from 2.5 to
-  None. If low_pass or high_pass is specified, then t_r needs to be specified
-  as well otherwise it will raise an error.
-- Order of filters in signal.clean and clean_img has changed to detrend, low-
-  and high-pass filter, remove confounds and standardize. To ensure
-  orthogonality between temporal filter and confound removal, an additional
-  temporal filter will be applied on the confounds before removing them. This
-  is according to Lindquist et al. (2018).
+- Default value of `t_r` in :func:nilearn.signal.clean: and
```
KamalakerDadi: I think nilearn.signal.clean should be in back quote and : after clean can re removed.

There is a reference if you would wish at 0.5.0 alpha in Enhancements section.

miykael:
@KamalakerDadi - I've updated `whats_new.rst` (thanks, I finally got it), and added a short description, as you suggested.

kchawla-pi:
@miykael Do you think this is ready for merge?

nilearn/image/tests/test_image.py
bthirion: Indeed, it would amount to testing that apply_mask is doing its job, which is not what we need here.

kchawla-pi:
@bthirion @KamalakerDadi @jeromedockes
One of you needs to review & greenlight this before merge.

kchawla-pi:
@GaelVaroquaux It would seem Bertrand already approves of it, based on his last comment.

kchawla-pi:
DOn't delete the branch yet.

PR #1845, opened 2018-10-26
Allen atlases

### LINKED ISSUE #1792: Ship compressed versions of the Allen atlases

KamalakerDadi:
Closes #1792 . Shipping Allen atlases from OSF to reduce the disk space load on CircleCI.


kchawla-pi:
@GaelVaroquaux This also seems ready for merge. It is a simple change, and Bertrand gave his blessing.

kchawla-pi:
Don't close this PR yet D Don't delete this branch yet.

Issue #1822, opened 2018-10-18
2x2 volumetric plot Arrangement


TheChymera:
Sorry for crossposting, but I was unable to get an answer [elsewhere](https://neurostars.org/t/2x2-field-plot-arrangement-from-nilearn/2602).

Is there any way to get the more “traditional” 2x2 plot arrangement (e.g. as seen in SPM, FSL, MRIcroGL, etc.) where the crosshairs of the frontal slice extend right and down into the horizontal slice view?

I appreciate the 3x1 matrix default of nilearn, as it is better packed, but I would also like to be able to use the traditional 2x2 arangement.

This approach has the major drawback that it leaves one cell of the 2x2 plot arrangement empty, but given modern technology that actually turns out to be very useful for a 3D visualizarion. Not least of all the consistent crosshair alignment is also very helpful to better grasp the placement in the brain.

GaelVaroquaux:
No there is not, unfortunately. 

A PR would be most appreciated. The way to do it would be to do a new display subclass. 

⁣Sent from my phone. Please forgive typos and briefness.​

TheChymera:
@GaelVaroquaux I haven't yet figured out where exactly the slices axes are arranged, but you mean make a variation of this class? https://github.com/nilearn/nilearn/blob/0e03cf6644662bd0ee0c420eff59926a22e29289/nilearn/plotting/displays.py#L1037

GaelVaroquaux:
Indeed, that's exactly (sorry, I was on my phone, and in trains)

The positioning of the axes is done by the "locator" method: https://github.com/nilearn/nilearn/blob/0e03cf6644662bd0ee0c420eff59926a22e29289/nilearn/plotting/displays.py#L1109

It's something a bit tricky: it's a callback called by matplotlib. We use it to get optimal ratios depending on the data.

KamalakerDadi:
Closing with #1869

PR #1869, opened 2018-11-21
[MRG] Issue #1822: volumetric plot arrangement

setina42:
Adressing  issue #1822  opened by @TheChymera, we have created a new display class called TiledSlicer, which is a variation of the OrthoSlicer class with changed `_locator `method, changed` _init_axes` method and different `_default_figsize` to arrange plots in a 2x2 fashion.

It can be called with specifying `display_mode='tiled'` in nilearns plotting functions, for example
`plot_anat(display_mode='tiled')`

### DIFF
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -1227,11 +1227,30 @@ class TiledSlicer(BaseSlicer):
     """
     _cut_displayed = 'yxz'
     _axes_class = CutAxes
-    _default_figsize = [2.0,6.0]
+    _default_figsize = [2.0, 6.0]
 
     @classmethod
     def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
-        "Instantiate the slicer and find cut coordinates"
+        """Instantiate the slicer and find cut coordinates.
+
+        Parameters
+        ----------
+        img: 3D Nifti1Image
+            The brain map.
+        threshold: float, optional
+            The lower threshold to the positive activation. If None, the
+            activation threshold is computed using the 80% percentile of
+            the absolute value of the map.
+
+        cut_coords: list of float, optional
+            xyz world coordinates of cuts.
+
+        Returns
+        ----------
+        cut_coords: list of float
+            xyz world coordinates of cuts.
+
+        """
         if cut_coords is None:
             if img is None or img is False:
                 cut_coords = (0, 0, 0)
@@ -1240,33 +1259,52 @@ class TiledSlicer(BaseSlicer):
                     img, activation_threshold=threshold)
             cut_coords = [cut_coords['xyz'.find(c)]
                           for c in sorted(self._cut_displayed)]
+
         return cut_coords
 
-    def _find_inital_axes_coord(self,index):
-        "find coordinates for inital axes placement for xyz cuts"
+    def _find_inital_axes_coord(self, index):
+        """"Find coordinates for initial axes placement for xyz cuts.
+
+        Parameters
+        ----------
+        index: int
+            Index corresponding to current cut 'x', 'y' or 'z'.
+
+        Returns
+        ----------
+        [coord1, coord2, coord3, coord4]: list of int
+            x0, y0, x1, y1 coordinates used by matplotlib to position axes in figure.
+
+        """
 
         rect_x0, rect_y0, rect_x1, rect_y1 = self.rect
 
-        if (index == 0):
+        if index == 0:
                 coord1 = rect_x1 - rect_x0
-                coord2 = 0.5 * (rect_y1-rect_y0) + rect_y0
+                coord2 = 0.5 * (rect_y1 - rect_y0) + rect_y0
                 coord3 = 0.5 * (rect_x1 - rect_x0) + rect_x0
-                coord4 = rect_y1-rect_y0
-
-        if (index == 1):
+                coord4 = rect_y1 - rect_y0
+        elif index == 1:
                 coord1 = 0.5 * (rect_x1 - rect_x0) + rect_x0
-                coord2 = 0.5 * (rect_y1-rect_y0) + rect_y0
+                coord2 = 0.5 * (rect_y1 - rect_y0) + rect_y0
                 coord3 = rect_x1 -rect_x0
-                coord4 = rect_y1-rect_y0
-        if (index == 2):
+                coord4 = rect_y1 - rect_y0
+        elif index == 2:
                 coord1 = rect_x1 - rect_x0
-                coord2 = rect_y1 -rect_y0
+                coord2 = rect_y1 - rect_y0
                 coord3 = 0.5 * (rect_x1 - rect_x0) + rect_x0
-                coord4 = 0.5 * (rect_y1-rect_y0) + rect_y0
- 
-        return([coord1,coord2,coord3,coord4])
+                coord4 = 0.5 * (rect_y1 - rect_y0) + rect_y0
+        return [coord1, coord2, coord3, coord4]
 
     def _init_axes(self, **kwargs):
+        """Initializes and places axes for display of 'xyz' cuts.
+
+        Parameters
+        ----------
+        kwargs:
+            additional arguments to pass to self._axes_class
+
+        """
 
         cut_coords = self.cut_coords
         if len(cut_coords) != len(self._cut_displayed):
@@ -1274,8 +1312,7 @@ class TiledSlicer(BaseSlicer):
                              ' match the display_mode')
 
         facecolor = 'k' if self._black_bg else 'w'
-        
-        # Create our axes:
+
         self.axes = dict()
         for index, direction in enumerate(self._cut_displayed):
             fh = self.frame_axes.get_figure()
@@ -1294,18 +1331,36 @@ class TiledSlicer(BaseSlicer):
             self.axes[direction] = display_ax
             ax.set_axes_locator(self._locator)
 
-    def _adjust_width_height(self,width_dict,height_dict,rect_x0, rect_y0, rect_x1, rect_y1):
-        "adjusts absolute image width and height to ratios" 
-          
+    def _adjust_width_height(self, width_dict, height_dict, rect_x0, rect_y0, rect_x1, rect_y1):
+        """ Adjusts absolute image width and height to ratios.
+
+        Parameters
+        ----------
+        width_dict: dict
+            Width of image cuts displayed in axes.
+        height_dict: dict
+            Height of image cuts displayed in axes.
+        rect_x0, rect_y0, rect_x1, rect_y1: float
+            Matplotlib figure boundaries.
+
+        Returns
+        ----------
+        width_dict: dict
+            Width ratios of image cuts for optimal positioning of axes.
+        height_dict: dict
+            Height ratios of image cuts for optimal positioning of axes.
+
+        """
+
         unique_width = {}
 
-        for key,value in width_dict.items():
+        for key, value in width_dict.items():
             if value not in unique_width.values():
                 unique_width[key] = value
 
         unique_height = {}
 
-        for key,value in height_dict.items():
+        for key, value in height_dict.items():
             if value not in unique_height.values():
                 unique_height[key] = value
 
@@ -1318,10 +1373,26 @@ class TiledSlicer(BaseSlicer):
         for ax, height in height_dict.items():
             height_dict[ax] = height / total_height * (rect_y1 - rect_y0)
 
-        return (width_dict,height_dict)
+        return (width_dict, height_dict)
 
-    def _find_axes_coord(self,rel_width_dict,rel_height_dict,rect_x0, rect_y0, rect_x1, rect_y1):
-        "determines optimal coordinates for positioning slice axes based on image size ratio"
+    def _find_axes_coord(self, rel_width_dict, rel_height_dict, rect_x0, rect_y0, rect_x1, rect_y1):
+        """"find coordinates for inital axes placement for xyz cuts.
+
+        Parameters
+        ----------
+        rel_width_dict: dict
+            Width ratios of image cuts for optimal positioning of axes.
+        rel_height_dict: dict
+            Height ratios of image cuts for optimal positioning of axes.
+        rect_x0, rect_y0, rect_x1, rect_y1: float
+            Matplotlib figure boundaries.
+
+        Returns
+        ----------
+        coord1, coord2, coord3, coord4: dict of {matplotlib.axes._axes.Axes: float}
+            x0, y0, x1, y1 coordinates per axes used by matplotlib to position axes in figure.
+
+        """
 
         coord1 = dict()
         coord2 = dict()
@@ -1337,7 +1408,6 @@ class TiledSlicer(BaseSlicer):
             coord2[ax] = (rect_y1) - rel_height_dict[ax]
             coord3[ax] = rect_x0 + rel_width_dict[ax]
             coord4[ax] = rect_y1
-
         try:
             ax = self.axes['x'].ax
         except KeyError:
@@ -1347,7 +1417,6 @@ class TiledSlicer(BaseSlicer):
             coord2[ax] = (rect_y1) - rel_height_dict[ax]
             coord3[ax] = rect_x1
             coord4[ax] = rect_y1
-
         try:
             ax = self.axes['z'].ax
         except KeyError:
@@ -1358,7 +1427,7 @@ class TiledSlicer(BaseSlicer):
             coord3[ax] = rect_x0 + rel_width_dict[ax]
             coord4[ax] = rect_y0 + rel_height_dict[ax]
 
-        return(coord1,coord2,coord3,coord4)
+        return(coord1, coord2, coord3, coord4)
           
     def _locator(self, axes, renderer):
         """ The locator function used by matplotlib to position axes.
@@ -1397,20 +1466,19 @@ class TiledSlicer(BaseSlicer):
             height_dict[display_ax.ax] = (ymax - ymin)
 
         #relative image height and width
-        rel_width_dict,rel_height_dict = self._adjust_width_height(width_dict,height_dict,rect_x0, rect_y0, rect_x1, rect_y1)
+        rel_width_dict, rel_height_dict = self._adjust_width_height(width_dict, height_dict, rect_x0, rect_y0, rect_x1, rect_y1)
     
         direction_ax = []
         for d in self._cut_displayed:
             direction_ax.append(display_ax_dict.get(d, dummy_ax).ax)
 
-        coord1,coord2,coord3,coord4 = self._find_axes_coord(rel_width_dict,rel_height_dict,rect_x0, rect_y0, rect_x1, rect_y1)
+        coord1, coord2, coord3, coord4 = self._find_axes_coord(rel_width_dict, rel_height_dict, rect_x0, rect_y0, rect_x1, rect_y1)
 
         return transforms.Bbox([[coord1[axes], coord2[axes]],
-                               [coord3[axes],coord4[axes]]])
+                               [coord3[axes], coord4[axes]]])
 
     def draw_cross(self, cut_coords=None, **kwargs):
-        """ Draw a crossbar on the plot to show where the cut is
-        performed.
+        """ Draw a crossbar on the plot to show where the cut is performed.
 
         Parameters
         ----------
@@ -1426,8 +1494,9 @@ class TiledSlicer(BaseSlicer):
         for direction in 'xyz':
             coord_ = None
             if direction in self._cut_displayed:
-                coord_ = cut_coords[
-                    sorted(self._cut_displayed).index(direction)]
+                sorted_cuts = sorted(self._cut_displayed)
+                index = sorted_cuts.index(direction)
+                coord_ = cut_coords[index]
             coords[direction] = coord_
         x, y, z = coords['x'], coords['y'], coords['z']
 
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -1273,7 +1273,8 @@ class TiledSlicer(BaseSlicer):
         Returns
         ----------
         [coord1, coord2, coord3, coord4]: list of int
-            x0, y0, x1, y1 coordinates used by matplotlib to position axes in figure.
+            x0, y0, x1, y1 coordinates used by matplotlib
+            to position axes in figure.
 
         """
 
@@ -1332,7 +1333,7 @@ class TiledSlicer(BaseSlicer):
             ax.set_axes_locator(self._locator)
 
     def _adjust_width_height(self, width_dict, height_dict,
-                            rect_x0, rect_y0, rect_x1, rect_y1):
+                             rect_x0, rect_y0, rect_x1, rect_y1):
         """ Adjusts absolute image width and height to ratios.
 
         Parameters
@@ -1377,13 +1378,14 @@ class TiledSlicer(BaseSlicer):
 
         for ax, width in width_dict.items():
             width_dict[ax] = width / total_width * (rect_x1 - rect_x0)
-        
+
         for ax, height in height_dict.items():
             height_dict[ax] = height / total_height * (rect_y1 - rect_y0)
 
         return (width_dict, height_dict)
 
-    def _find_axes_coord(self, rel_width_dict, rel_height_dict, rect_x0, rect_y0, rect_x1, rect_y1):
+    def _find_axes_coord(self, rel_width_dict, rel_height_dict,
+                         rect_x0, rect_y0, rect_x1, rect_y1):
         """"find coordinates for inital axes placement for xyz cuts.
 
         Parameters
@@ -1397,8 +1399,9 @@ class TiledSlicer(BaseSlicer):
 
         Returns
         ----------
-        coord1, coord2, coord3, coord4: dict of {matplotlib.axes._axes.Axes: float}
-            x0, y0, x1, y1 coordinates per axes used by matplotlib to position axes in figure.
+        coord1, coord2, coord3, coord4: dict
+            x0, y0, x1, y1 coordinates per axes used by matplotlib
+            to position axes in figure.
 
         """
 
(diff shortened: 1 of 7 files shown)


GaelVaroquaux:
I haven't looked at the PR, but it seems very promising. Definitely a
feature that we would like to have.

I do notice that it does not seem that a test and an example has been
added. It would be useful to add this functionality to both on of the
existing tests, and one of the existing examples, for testing and
documentation purposes.

GaelVaroquaux:
PS: thank you. I'm excited!

setina42:
Thank you :) Regarding the tests, I see you have tests for different display_modes in `test_img_plotting.py`, and also in `test_displays.py`. Should we add tests to both, and should we also include tests anywhere else as well?

GaelVaroquaux:
One or two test will suffice.

kchawla-pi:
Coverage seems to indicate display.py is not adequately tested.

setina42:
Sorry for the late reply @kchawla-pi.  I had a look at the codecov report. It seems the class is mostly tested with the exception of a few if-statements,` _black_bg` and the `_locator` function. I will add another test shortly for `black_bg` and the if-statements, but I don't quite understand how to add a test for the` _locator` method since it's a callback by matplotlib.

I found one example in `test_display.py`, would this be sufficient if I adapt it for TiledSlicer? I don't fully understand how it tests the locator code though.

```
def test_stacked_slicer():
    # Test stacked slicers, like the XSlicer
    img = load_mni152_template()
    slicer = XSlicer.init_with_figure(img=img, cut_coords=3)
    slicer.add_overlay(img, cmap=plt.cm.gray)
    # Forcing a layout here, to test the locator code
    with tempfile.TemporaryFile() as fp:
        slicer.savefig(fp)
    slicer.close()
```

kchawla-pi:
Hi @setina42 

1. The methods in the TiledSlicer class are super long. Refactoring them into smaller methods, each doing only 1 thing without side effects, will vastly improve code quality, maintainability and testability.

2. There are no unit tests for each of the methods. Current integration tests are checking if an exception is raised or not, when those methods are called, but not testing whether each method is doing what it is supposed to.

nilearn/plotting/displays.py
```diff
@@ -1203,6 +1203,242 @@ def draw_cross(self, cut_coords=None, **kwargs):
...
+            self.frame_axes.set_zorder(-1000)
+
+    def _locator(self, axes, renderer):
+        """ The locator function used by matplotlib to position axes.
+            Here we put the logic used to adjust the size of the axes.
+        """
+
+        x0, y0, x1, y1 = self.rect
+        width_dict = dict()
+        height_dict = dict()
```
kchawla-pi: `width_dict` & `height_dict` are not good variable names. 

1. It is bad practice to include the datatype (dict) in the variable name.
2. width & height are very generic names. The variable names should be self-documenting, for example: `width_of _<something>`, `height_of _<something>`

nilearn/plotting/displays.py
```diff
@@ -1203,6 +1203,242 @@ def draw_cross(self, cut_coords=None, **kwargs):
...
+            cut_coords = [cut_coords['xyz'.find(c)]
+                          for c in sorted(self._cut_displayed)]
+        return cut_coords
+
+    def _init_axes(self, **kwargs):
+        cut_coords = self.cut_coords
+        if len(cut_coords) != len(self._cut_displayed):
+            raise ValueError('The number cut_coords passed does not'
+                             ' match the display_mode')
+        x0, y0, x1, y1 = self.rect
```
kchawla-pi: Make them `rect_x0` and so on.

kchawla-pi:
Hi @setina42 Thanks for helping us out!
I have added my review, the changes should not be ornery and will improve maintainability down the road.

setina42:
Great, thank you for the review! I'll see that I implement your requested changes within next week.

nilearn/plotting/displays.py
```diff
@@ -1203,6 +1203,242 @@ def draw_cross(self, cut_coords=None, **kwargs):
...
+            else:
+                kwargs['color'] = 'k'
+
+        if 'y' in self.axes:
+            ax = self.axes['y'].ax
+            if x is not None:
+                ax.axvline(x, **kwargs)
+            if z is not None:
+                ax.axhline(z, **kwargs)
+
```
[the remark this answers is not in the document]
kchawla-pi: This is assuming `0` is not a valid value for `x` and `z`. If it is, keep using the `is not None`.

nilearn/plotting/displays.py
setina42: We tried to use the same names or logic for variable names as in the OrthoSlicer Subclass wherever possible for consistency. Both `width_dict` and `height_dict` are used in the OrthoSlicer Subclass similarly. Would you prefer we change them only for the new class TiledSlicer now, or keep them and maybe open another issue to change variable names for OrthoSlicer as well?

setina42:
Thank you for your help! I have adressed your requested changes. However, I am still unsure on how to properly write a test that will also cover the `_locator` method. Could you point me at how you do that for the other classes in display.py? I am unfamiliar with automated testing for callbacks.

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
                           for c in sorted(self._cut_displayed)]
         return cut_coords
 
+    def _find_inital_axes_coord(self,index):
+        "find coordinates for inital axes placement for xyz cuts"
+
+        rect_x0, rect_y0, rect_x1, rect_y1 = self.rect
+
+        if (index == 0):
```
kchawla-pi: You don't need parantheses here.

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
...
+                coord1 = rect_x1 - rect_x0
+                coord2 = 0.5 * (rect_y1-rect_y0) + rect_y0
+                coord3 = 0.5 * (rect_x1 - rect_x0) + rect_x0
+                coord4 = rect_y1-rect_y0
+
+        if (index == 1):
```
kchawla-pi: Make this an `elif` and the remaining conditions need not be checked at all.

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
...
+                coord1 = 0.5 * (rect_x1 - rect_x0) + rect_x0
+                coord2 = 0.5 * (rect_y1-rect_y0) + rect_y0
+                coord3 = rect_x1 -rect_x0
+                coord4 = rect_y1-rect_y0
+        if (index == 2):
```
kchawla-pi: `elif`
Also, is there any possibility that the values of index can be anything other than 0, 1, 2?

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
...
+                coord1 = rect_x1 - rect_x0
+                coord2 = rect_y1 -rect_y0
+                coord3 = 0.5 * (rect_x1 - rect_x0) + rect_x0
+                coord4 = 0.5 * (rect_y1-rect_y0) + rect_y0
+ 
+        return([coord1,coord2,coord3,coord4])
+
     def _init_axes(self, **kwargs):
+
```
kchawla-pi: Docstring.

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
                           for c in sorted(self._cut_displayed)]
         return cut_coords
 
+    def _find_inital_axes_coord(self,index):
+        "find coordinates for inital axes placement for xyz cuts"
```
kchawla-pi: Docstrings in Python should be Triple double quotes, i.e.: 
```
""" Find coordinates...
"""
```

Docstrings should explain what parameters are expected. Here: index, datatype, what is it.
They should also explain what is being returned. Here List(coords). What are cords?

Check out Docstrings in existing functions to see how to format them. That is especially important as we use automated tools to parse & read the docstrings and insert them into our documentation. The tools expect the docstrings to follow very strict formatting rules.

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
                           for c in sorted(self._cut_displayed)]
         return cut_coords
 
+    def _find_inital_axes_coord(self,index):
```
kchawla-pi: There should be a space after `,` when specifying parameters or list elements, or anywhere there's a comma really.
`(self, index)`
Those are the PEP8 rules and they aim to enhance code readability. Google and read up on them, they are pretty simple and sensible.

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
...
+        "find coordinates for inital axes placement for xyz cuts"
+
+        rect_x0, rect_y0, rect_x1, rect_y1 = self.rect
+
+        if (index == 0):
+                coord1 = rect_x1 - rect_x0
+                coord2 = 0.5 * (rect_y1-rect_y0) + rect_y0
+                coord3 = 0.5 * (rect_x1 - rect_x0) + rect_x0
+                coord4 = rect_y1-rect_y0
+
```
kchawla-pi: The empty line here is unnecessary. The code is sparse enough not to need it. The white space here is breaking continuity.

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
...
+                coord1 = 0.5 * (rect_x1 - rect_x0) + rect_x0
+                coord2 = 0.5 * (rect_y1-rect_y0) + rect_y0
+                coord3 = rect_x1 -rect_x0
+                coord4 = rect_y1-rect_y0
+        if (index == 2):
+                coord1 = rect_x1 - rect_x0
+                coord2 = rect_y1 -rect_y0
+                coord3 = 0.5 * (rect_x1 - rect_x0) + rect_x0
+                coord4 = 0.5 * (rect_y1-rect_y0) + rect_y0
+
```
kchawla-pi: Unnecessary line space

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
...
+                coord2 = 0.5 * (rect_y1-rect_y0) + rect_y0
+                coord3 = rect_x1 -rect_x0
+                coord4 = rect_y1-rect_y0
+        if (index == 2):
+                coord1 = rect_x1 - rect_x0
+                coord2 = rect_y1 -rect_y0
+                coord3 = 0.5 * (rect_x1 - rect_x0) + rect_x0
+                coord4 = 0.5 * (rect_y1-rect_y0) + rect_y0
+ 
+        return([coord1,coord2,coord3,coord4])
```
kchawla-pi: Space after commas.
Return value does not need to be, & should not be, in parantheses.
`return [coord1, coord2, ...`

nilearn/plotting/displays.py
```diff
@@ -1242,37 +1242,45 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
...
+
         cut_coords = self.cut_coords
         if len(cut_coords) != len(self._cut_displayed):
             raise ValueError('The number cut_coords passed does not'
                              ' match the display_mode')
-        x0, y0, x1, y1 = self.rect
+
         facecolor = 'k' if self._black_bg else 'w'
+        
         # Create our axes:
```
kchawla-pi: This comment is unnecessary: the function name & code are telling what's ahppenning.

nilearn/plotting/displays.py
```diff
@@ -1285,28 +1293,84 @@ def _init_axes(self, **kwargs):
             display_ax = self._axes_class(ax, direction, coord, **kwargs)
             self.axes[direction] = display_ax
             ax.set_axes_locator(self._locator)
+
+    def _adjust_width_height(self,width_dict,height_dict,rect_x0, rect_y0, rect_x1, rect_y1):
+        "adjusts absolute image width and height to ratios"
```
kchawla-pi: Docstring: parameters, returns. Also, triple double quotes

nilearn/plotting/displays.py
```diff
@@ -1332,61 +1396,17 @@ def _locator(self, axes, renderer):
...
-            if value not in unique_width_dict.values():
-                unique_width_dict[key] = value
-
-        unique_height_dict = {}
-
-        for key,value in height_dict.items():
-            if value not in unique_height_dict.values():
-                unique_height_dict[key] = value
+        #relative image height and width
+        rel_width_dict,rel_height_dict = self._adjust_width_height(width_dict,height_dict,rect_x0, rect_y0, rect_x1, rect_y1)
```
kchawla-pi: Space after commas.

nilearn/plotting/displays.py
```diff
@@ -1332,61 +1396,17 @@ def _locator(self, axes, renderer):
...
-            coord3_dict[ax] = x1
-            coord4_dict[ax] = y1
-
-        if 'z' in self.axes:
-            ax = self.axes['z'].ax
-            coord1_dict[ax] = x0
-            coord2_dict[ax] = y0
-            coord3_dict[ax] = x0 + width_dict[ax]
-            coord4_dict[ax] = y0 + height_dict[ax]
+        coord1,coord2,coord3,coord4 = self._find_axes_coord(rel_width_dict,rel_height_dict,rect_x0, rect_y0, rect_x1, rect_y1)
```
kchawla-pi: Space after commas.

nilearn/plotting/displays.py
```diff
@@ -1404,36 +1424,45 @@ def draw_cross(self, cut_coords=None, **kwargs):
             cut_coords = self.cut_coords
         coords = dict()
         for direction in 'xyz':
-            coord = None
+            coord_ = None
             if direction in self._cut_displayed:
-                coord = cut_coords[
+                coord_ = cut_coords[
                     sorted(self._cut_displayed).index(direction)]
```
kchawla-pi: Break this into one or two variables, the variable name describing what it is.

kchawla-pi:
Hey @setina42 this is great work! 
Thank you for the changes you made! I have posted another round of review which is essentially formatting changes and some work on the docstring. 
I will reply tomorrow or within a few days about testing the _locator() method.

nilearn/plotting/displays.py
kchawla-pi: There should be a space around `-` in line 1252, 1254

setina42:
Thanks again for the review and advice. I have added more detailed docstrings and changed the formatting.

KamalakerDadi:
This option ```display_mode=tiled``` needs to be documented at all plotting functions. ```plot_roi```, ```plot_stat_map```, ```plot_prob_atlas```, ```plot_epi```.

For example:

display_mode : {'ortho', 'x', 'y', 'z', 'xy', 'xz', 'yz'}
            Choose the direction of the cuts: 'x' - sagittal, 'y' - coronal,
            'z' - axial, 'ortho' - three cuts are performed in orthogonal
            directions.

KamalakerDadi:
It would be great if you could add this new option here in user guide.
http://nilearn.github.io/plotting/index.html#different-display-modes

KamalakerDadi:
Can we use ```display_mode='tiled'``` for other plotting function, in ```plot_roi``` ?

KamalakerDadi:
Or only for ```plot_stat_map``` ?

setina42:
You can use `display_mode='tiled'` for all plotting functions in img_plotting.py with the exception of `plot_glass_brain` and `plot_connectome`.

nilearn/plotting/displays.py
```diff
@@ -1203,6 +1203,242 @@ def draw_cross(self, cut_coords=None, **kwargs):
...
+        coord3_dict = dict()
+        coord4_dict = dict()
+
+        if 'y' in self.axes:
+            ax = self.axes['y'].ax
+            coord1_dict[ax] = x0
+            coord2_dict[ax] = (y1) - height_dict[ax]
+            coord3_dict[ax] = x0 + width_dict[ax]
+            coord4_dict[ax] = y1
+
```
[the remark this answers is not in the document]
kchawla-pi: I have been looking into that and I spoke with a few other engineers.
There's a higher level structure in Python StdLib designed for exactly this situation.

```
with contextlib.suppress(KeyError):
    ax = self.axes['y'].ax
    coord1[ax] = x0
    coord2[ax] = (y1) - height[ax]
    coord3[ax] = x0 + width[ax]
    coord4[ax] = y1
```
If the specified error is raised the subsequent `with` block is not executed.

The benefit of this it retains the cleanness of `if` and its not difficult to understand by new contributors - one google search away when someone looks at it the first time and they will understand & remember.
Another major benefit is that it does not add the double processing overhead of doing an `if` and then membership check using `in` on a sequence structure (`dict.keys()`).

The library is littered with such `if-in` lookups and together they add processing overhead that the standard library is well equipped to avoid.

KamalakerDadi:
@setina42 
I just used display_mode='tiled' with plot_roi. It works nice. But, when I use ```bg_img```, it displays only two cuts "x" and "y" but not "z". May be we need to look into that before merging. I posted the code to reproduce.
```python
from nilearn import plotting, datasets

haxby_anat_filename = haxby_dataset.anat[0]
haxby_mask_filename = haxby_dataset.mask_vt[0]

plotting.plot_roi(haxby_mask_filename, bg_img=haxby_anat_filename,
                  title="plot_roi", display_mode='tiled')
plotting.show()
```

nilearn/plotting/displays.py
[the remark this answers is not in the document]
massich: I agree with @jeromedockes. The snippet proposed by @kchawla-pi makes the code less readable (and violates the first 3 rules of the [zen of python](https://en.wikipedia.org/wiki/Zen_of_Python)). The rule of thumb here would use the `try/catch` if it makes sense to write 2 separate tests one just calling whatever needs to be called, and another one calling `with pytest.raises(bla)`. If that's not the case you most probably don't want to use a try catch.

what I dislike about what the author is proposing is that there's no else branch and therefore is incomplete (or degraded) conditional. Still valid python syntax but a degraded conditional case none the less (and therefore can never be cover). Thats not the case for the snipped proposed by @kchawla-pi but this can be solved by doing this:

```py
       if 'y' in self.axes:
            ax = self.axes['y'].ax
            coord1_dict[ax] = x0
            coord2_dict[ax] = (y1) - height_dict[ax]
            coord3_dict[ax] = x0 + width_dict[ax]
            coord4_dict[ax] = y1
       else:
          pass
```
But since this is rearely seen in scientific codebase and degraded conditionals are not offending most scientific users/developers. I would not even bother.



Once said all that, whatever you do for `x`, `y`, and `z` is quite similar. I'm sure you can factor it. And it seems quite a natural thing to be doing several times, so I would check other places in the code doing similar things, factor it nicely and make function out of it.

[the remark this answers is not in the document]
jeromedockes: > There's a higher level structure in Python StdLib designed for exactly this situation.
> 
> ```
> with contextlib.suppress(KeyError):
>     ax = self.axes['y'].ax
>     coord1[ax] = x0
>     coord2[ax] = (y1) - height[ax]
>     coord3[ax] = x0 + width[ax]
>     coord4[ax] = y1
> ```
> 
> If the specified error is raised the subsequent `with` block is not executed.

but then a bug could be hidden if 'y' is in axes but for example ax is not in coord1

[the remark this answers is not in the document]
jeromedockes: > Another major benefit is that it does not add the double processing overhead of doing an `if` and then membership check using `in` on a sequence structure (`dict.keys()`).

dict have hash tables so a lookup is fast. Moreover catching an exception has a  cost too. more importantly this part of the code is definitely not a performance bottleneck so readability is all that matters

nilearn/plotting/displays.py
```diff
@@ -1203,6 +1203,242 @@ def draw_cross(self, cut_coords=None, **kwargs):
...
+            coords[direction] = coord
+        x, y, z = coords['x'], coords['y'], coords['z']
+
+        kwargs = kwargs.copy()
+        if 'color' not in kwargs:
+            if self._black_bg:
+                kwargs['color'] = '.8'
+            else:
+                kwargs['color'] = 'k'
+
```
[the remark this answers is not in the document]
massich: Unless you do through the exception and write a test for it, I would not use try/except.

usually if I've to write 
```py
try:
   bla
except KeyError:
   pass
```
I do:
```py
if _is_safe_to_do_bla():
   bla
else:
   pass
```

nilearn/plotting/displays.py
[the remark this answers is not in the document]
kchawla-pi: >  The snippet proposed by @kchawla-pi makes the code less readabl
Agrees, I looked at it again and didn't really find it clean, so I looked for the contextlib solution.

>  then a bug could be hidden
True, I guess we shouldn't use the contextib.suppress

> if 'y' is in axes but for example ax is not in coord1
That is not being handled anywhere, including in the way it was originally written. We should handle it and give a message that makes sense in this context.

```
      if 'y' in self.axes:
            ax = self.axes['y'].ax 
            coord1_dict[ax] = x0
            coord2_dict[ax] = (y1) - height_dict[ax]
            coord3_dict[ax] = x0 + width_dict[ax]
            coord4_dict[ax] = y1
```

Perhaps switching back to `if` is one way to do it.

>  double processing overhead of doing an if and then membership check using in on a sequence structure (dict.keys()).
I am concerned about this, Nilearn will get more complex & CircleCI build times keep going up. If we can sensibly replace the ton of `if`'s we use with more try-except, it'll be better, esp in new code. If you have ideas about that let me know.

setina42:
Ok, this is not good, thank you for the example. I tried just now with different `bg_img`, and it is working for other images. But with this dataset, also other plotting functions like` plot_img` or `plot_stat_map` are not working properly with the new `display_mode`, this seems like a major  bug. I'll look into it as soon as possible.

KamalakerDadi:
>I'll look into it as soon as possible.

Fantastic. I appreciate it.

setina42:
Ok, I have made a logical error for figuring out the correct image size ratios, that's why the cuts were not properly displayed for all cases. I have fixed the issue and retested the plotting functions with a variety of images. It should now be working properly.

TheChymera:
@KamalakerDadi any updates?

KamalakerDadi:
Everything is great except the flake8 issues which are quite a lot as we can see from here: https://travis-ci.org/nilearn/nilearn/jobs/473595592. 

Also, please add your contribution to doc/whats_new.rst in a new release section 0.5.1

setina42:
Thank you @KamalakerDadi , I went over the job log and fixed the flake8 issues. Please let me know if anything else is missing.

nilearn/plotting/img_plotting.py
```diff
@@ -143,11 +143,13 @@ def _plot_img_with_bg(img, bg_img=None, cut_coords=None,
                    'Tip: Use np.nanmax() instead of np.max().')
         warnings.warn(nan_msg)
 
-    if isinstance(cut_coords, numbers.Number) and display_mode == 'ortho':
+    if (isinstance(cut_coords, numbers.Number) and
```
KamalakerDadi: Please add test related to this in nilearn/plotting/tests/test_img_plotting.py. Sorry for late on this.

There are already tests related to display_mode='ortho' where you could inspire from ```test_invalid_in_display_mode_cut_coords_all_plots```.

nilearn/plotting/tests/test_img_plotting.py
```diff
@@ -958,3 +958,20 @@ def test_add_markers_using_plot_glass_brain():
     coords = [(-34, -39, -9)]
     fig.add_markers(coords)
     fig.close()
+
+
+def test_plotting_functions_with_TiledSlicer():
```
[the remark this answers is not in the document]
KamalakerDadi: Please address this.

nilearn/plotting/tests/test_img_plotting.py
```diff
@@ -958,3 +958,20 @@ def test_add_markers_using_plot_glass_brain():
...
+    img = _generate_img()
+    plot_stat_map(img,display_mode='tiled')
+    plot_anat(display_mode='tiled')
+    plot_img(img,display_mode='tiled')
+    plt.close()
+
+
+def test_display_methods_TiledSlicer():
```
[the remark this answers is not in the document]
KamalakerDadi: Please address this as well.

setina42:
@KamalakerDadi  Great, thank you!

TheChymera:
@KamalakerDadi thanks for your help. What's the policy on getting this merged now? I see some [MRG] issues are still in limbo.

KamalakerDadi:
@TheChymera thank you for orchestrating this.

>I see some [MRG] issues are still in limbo.

ping @GaelVaroquaux @bthirion

GaelVaroquaux:
Overall, we should not wait for @bthirion or my approval to merge: this looks great, and we shouldn't be the bottleneck.

I had a quick look (not in detail). I love it.

If I had a tiny nitpick, it would be: to save space, is it easy to have the colorbar in the empty spot?

jeromedockes:
having a quick look at the diff coverage, it does seem that the tests need to be
completed a bit.
also, I think the conclusion of the discussion with @kchawla-pi  about accessing keys of
dictionaries that might be missing was to use if/else rather than
try/except:pass/else

setina42:
@GaelVaroquaux Thank you. About the colorbar, for our specific use case we would actually prefer to leave the spot open and have the colorbar to the side, as we have cases where we would like to overlay a fourth plot.
Also, to have the colorbar in the open spot would reduce its size, and I think the arrangement might be clearer this way. I would prefer to leave it, what do you think?

jeromedockes:
travis failure seems unrelated

setina42:
@jeromedockes, @ Thanks for your help. I have slightly adapted one test to increase the code coverage.  Please let me know if there is anything still open form my side, or if it will be ok to merge.

kchawla-pi:
Thanks for the great work @setina42 !

GaelVaroquaux:
Yippee Ki Yay!

setina42:
Awesome! Thanks for all your help and patience.

PR #1872, opened 2018-11-28
load_niimg drops header if dtype is changed

wiheto:
Hey, 

This appears to be a bug. Whenever there is a call to `nilearn._utils.load_niimg`, (or `nilearn._utils.check_niimg_4d()`) and dtype='auto' and the data's dtype!=target_dtype, the header gets dropped. This does not seem optimal. Here is an example mimicking the behaviour. 

But this can be fixed by adding copy_header flag and a call to niimg.header.set_data_dtype() 

The code below recreates the behaviour of the bug.

```
from nilearn import datasets
from nilearn._utils import check_niimg_4d
import nibabel as nb 
import numpy as np 
dataset = datasets.fetch_adhd(n_subjects=1)
fname = dataset.func[0]
img1 = nb.load(fname)
img2 = check_niimg_4d(img1,dtype=None)
img3 = check_niimg_4d(img1,dtype=np.int32)
# Answer = TR=2
print(img1.header.get_zooms()[-1])
# Answer = TR=2
print(img2.header.get_zooms()[-1])
# Answer = TR=1,  with my fix this = TR=2
print(img3.header.get_zooms()[-1])
```

### DIFF
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -314,6 +314,7 @@ class _MNI152Template(SpatialImage):
     affine = None
     vmax = None
     _shape = None
+    header = None
 
     def __init__(self, data=None, affine=None, header=None):
         # Comply with spatial image requirements while allowing empty init
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -314,6 +314,7 @@ class _MNI152Template(SpatialImage):
     affine = None
     vmax = None
     _shape = None
+    # Having a header is required by the load_niimg function
     header = None
 
     def __init__(self, data=None, affine=None, header=None):
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -116,8 +116,12 @@ def load_niimg(niimg, dtype=None):
     dtype = _get_target_dtype(niimg.get_data().dtype, dtype)
 
     if dtype is not None:
+        if niimg.header is not None:
+            copyheader = True
+        else:
+            copyheader = False
         niimg = new_img_like(niimg, niimg.get_data().astype(dtype),
-                             niimg.affine,copy_header=True)
+                             niimg.affine, copy_header=copyheader)
         niimg.header.set_data_dtype(dtype)
 
     return niimg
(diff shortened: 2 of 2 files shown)


bthirion:
Thx. I like the idea, but there are side effects (see Travis failures).

wiheto:
Yeah, I noticed it was a problem. I made a workaround on niworkflows where this was causing a problem for us. So I don't mind closing this now and ignoring the solution. 

But regarding the reason why it is failing, it seems to be in cases where there is a header missing from the nifti file. It appears to be a nilearn defined object which is causing the problem where img.header is set to None. Are there any other cases where you would expect where img.header = None? Wouldn't some sort of dummy headers be useful here.

kchawla-pi:
This seems to be something we should fix. At the very minimum the header should not be dropped invisibly for the user, though a fix would be preferred. Any ideas? @GaelVaroquaux

bthirion:
This should be fixed indeed. For this we need to check the failures induces by the current PR.

GaelVaroquaux:
0verall I think that it's just a question of adding "if niimg.header is not None" in this PR.

Also, there should be a space after the coma, for PEP8 :).

wiheto:
Took me a while to get round to this. 

I never got around to looking up if header data can generally be expected to be missing in any nifti data (I don't think it is the case though). Regardless, I added the conditional to check if there was a header present and a MNI152Template.header = None.

nilearn/plotting/img_plotting.py
```diff
@@ -314,6 +314,7 @@ class _MNI152Template(SpatialImage):
     affine = None
     vmax = None
     _shape = None
+    header = None
```
bthirion: Can you give an explanation here ?

wiheto: Without it, the following error is raised (see b9e15b9da31faa5fa9b607f8d541a8c273e32a20): 

AttributeError: '_MNI152Template' object has no attribute '_header'

The check I propose in load_niimg checks to see if the header is None. But MNI152Template has no header at all, this gives it a header.

bthirion: Simply add a short comment: "Having a header is required by the load_niimg function"

wiheto: Done

GaelVaroquaux:
LGTM. Merging. Thank you!

Issue #1877, opened 2018-11-30
Removing Python 3.4 support


kchawla-pi:
Scikit-Learn v0.21 will drop Python 3.4 support. We should make a deprecation plan for it.

kchawla-pi:
I'll add a DeprecationWarning to master for 2 release cycles in the same module as the Python2 warning.

PR #1882, opened 2018-12-06
[MRG] Improve Python Warning code & add a DeprecationWarning for Python 3.4

kchawla-pi:
For Issue #1877 
While addressing https://github.com/nilearn/nilearn/pull/1882#discussion_r241610908 of this PR, I noticed certain deficiencies in the design of the python deprecation warnings code. I have also addressed them, as it made addressing the above linked faux pa cleaner.


nilearn/__init__.py
```diff
@@ -51,6 +51,18 @@ def _py2_deprecation_warning():
                       stacklevel=3,
                       )
 
+def _py34_deprecation_warning():
+    warnings.simplefilter('once')
```
dimrozakis: Hey @kchawla-pi 

This line (as well as the same line in `_py2_deprecation_warning`) permanently change the `warnings` filters for all warning types, from all modules/files/lines everywhere.

In my scenario, I explicitly configure warnings, displaying deprecation warnings by default but masking other warnings, but after `nilearn` is imported, my configuration gets overridden by this statement.

How about nesting this block under a `with warnings.catch_warnings():`? That would allow you to temporarily display deprecation warnings, but after doing so, you'd be reverting the warning filters to their prior settings. Do you think that would be a good idea?

kchawla-pi: Thanks @dimrozakis I will look into this! If this is filtering all warnings, that is definitely unintended & undesirable behavior.
Would you send me a code sample which alerted you to this behavior? This was missed this and would like a more robust test for this.

dimrozakis: Hey @kchawla-pi 

The following snippet:
```python
#!/usr/bin/env python2
import warnings
warnings.warn("Please ignore this warning: 1")  # Will be printed.
warnings.filterwarnings("ignore", message="Please ignore this warning")
warnings.warn("Please ignore this warning: 2")  # Will not be printed.
import nilearn
warnings.warn("Please ignore this warning: 3")  # Should not be printed but is.
```

produces this output:
```
./a.py:3: UserWarning: Please ignore this warning: 1
  warnings.warn("Please ignore this warning: 1")
./a.py:6: DeprecationWarning: Python2 support is deprecated and will be removed in a future release. Consider switching to Python3.
  import nilearn
./a.py:7: UserWarning: Please ignore this warning: 3
  warnings.warn("Please ignore this warning: 3")
```

Warning 1 gets printed, but after adding an ignore filter for warnings matching a particular string, warning 2 isn't printed. Then nilearn gets imported, which prints its deprecation warning. After that, warning 3 gets printed because the filter that ignored warning 2 has been overridden by the filter configured in nilearn. The ignore filter I had configured still exists in `warnings.filters`. But the filter added by nilearn matches first so the warning gets printed.

Hope the above help to illustrate the problem.

dimrozakis: An alternative to using the `warnings.catch_warnings` context manager would be to configure a global warnings filter that is specific enough so that it doesn't mess with other warnings and filters. Filters can be configured by category, message regex, module regex, even line number iirc.

kchawla-pi: Thanks, this is very helpful. I'll look into this.

kchawla-pi:
@dimrozakis I believe the new changes should fix the glitch of changing warning filters.
@GaelVaroquaux This is ready for review.

Issue #1638, opened 2018-05-15
Masker family reloaded: NiftiSpheresMasker silently ignores voxels


banilo:
Using the NiftiSpheresMasker() I found out: If no mask_img is specified and it is thus computed by  a heuristic on the data images fed into transform(), then spheres that touch *any* voxel not inside the automatically computed mask will be set to overall NaN. That is, even if most of the sphere is inside the data space, the overall sphere feature appears to result to NaN if a tiny part is not inside the implicit mask. In my particular case, it took me some time to find this source of unexpected behavior, and it was fixed my explicitly specifying a liberal mask as "search space".

Perhaps it would be good to add an option that controls this behavior more directly and add a warning to the output that informs the user why some values in the sphere space will turn out to be NaN coming out of transform().

AlexandreAbraham:
If I understand well your usecase: you use NiftiSpheresMasker. When you do not put a mask_img, you get NaN for some spheres. When you put a mask, you get values in these spheres.

First, there is no heuristic to compute masks automatically in the NiftiSpheresMasker. So my guess here is that, in your data, you must have NaNs outside of the brain and, if a sphere overlaps without outside brain voxels, the mean ends up being a NaN. When you put the mask, NaNs are removed, and then you get non-NaN values.

Could you check is you have NaNs outside of the brain? This would allow us to rule out this case.

Sidenote: we should probably detect voxels with NaNs and raise an error if they are not masked.

himanshupathak21061998:
I want to work on this Issue . According to this if the value mask_img argument is None than I have to give a warning message and detect those voxels which are turned into NaN . This much is sufficient or anything else is required

himanshupathak21061998:
I want to work on this Issue . In this issue if the mask_img parameter is None in class NiftiSpheresMasker then I have to give a warning message and detect those voxels which are turned into Nan . This much is sufficient or anything else I have to do .

himanshupathak21061998:
Please I need some suggestions help me @KamalakerDadi

KamalakerDadi:
Hi @himanshupathak21061998 Thanks for your interest. Let me get back to this with a reproducible script and then we fix it.
Thanks again

KamalakerDadi:
@banilo Can you provide Nifti files to reproduce ? Thanks a lot!

kimberlylray:
Hi - following on a similar issue I recently ran into with NiftiSpheresMasker. 
[...]

PR #1896, opened 2019-01-11
NiftiSpheresMasker function  _safe_get_data added

himanshupathak21061998:
An attempt to solve issue #1638 @KamalakerDadi @GaelVaroquaux please check this

### DIFF
--- a/nilearn/input_data/nifti_spheres_masker.py
+++ b/nilearn/input_data/nifti_spheres_masker.py
@@ -5,14 +5,14 @@ Transformer for computing seeds signals
 Mask nifti images by spherical volumes for seed-region analyses
 """
 import numpy as np
-import warnings
 import sklearn
-import nibabel
+import warnings
 from sklearn import neighbors
 from sklearn.externals.joblib import Memory
 from distutils.version import LooseVersion
 
 from ..image.resampling import coord_transform
+from .._utils.niimg_conversions import _safe_get_data
 from .._utils import CacheMixin
 from .._utils.niimg_conversions import check_niimg_4d, check_niimg_3d
 from .._utils.class_inspect import get_params
@@ -38,8 +38,17 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
 
         X = masking._apply_mask_fmri(niimg, mask_img)
     else:
-        mask_coords = list(np.ndindex(niimg.shape[:3]))
-        X = niimg.get_data().reshape([-1, niimg.shape[3]]).T
+        #Adding  warning message 
+        if np.isnan(np.sum(_safe_get_data(niimg))):
+            warnings.warn('The imgs you have fedded into fit_transform()' 
+                          'contains NaN values which are converted to zeroes '
+                          'please use mask_imgs to mask out NaNs othewise you '
+                           'should expect some deviation in fit_transform()')
+            mask_coords = list(np.ndindex(niimg.shape[:3]))
+            X = _safe_get_data(niimg,True).reshape([-1, niimg.shape[3]]).T
+        else:
+            mask_coords = list(np.ndindex(niimg.shape[:3]))
+            X = _safe_get_data(niimg).reshape([-1, niimg.shape[3]]).T
 
     # For each seed, get coordinates of nearest voxel
     nearests = []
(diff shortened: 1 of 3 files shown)


himanshupathak21061998:
File "/home/circleci/project/nilearn/input_data/nifti_spheres_masker.py", line 243, in _safe_get_data
    check_imgs = np.array(imgs.dataobj)
AttributeError: 'str' object has no attribute 'dataobj'
I can't able to understand why in circleci test imgs is treated as string please help @bthirion

KamalakerDadi:
As per the issue, we need to implement point 1.

Point 1 requires automatically converting nans to zeros and raise a warning that we are putting nans to zeros. But, does not requires for new argument ```ensure_finite```.

At line:40 https://github.com/nilearn/nilearn/blob/master/nilearn/input_data/nifti_spheres_masker.py#L40
needs to be changed AFAIK. Instead of using niimg.get_data(), we need to use ```_safe_get_data(niimg)```. This _safe_get_data needs to be imported ```from nilearn._utils.niimg_conversions import _safe_get_data```. You don't have to implement this.

If np.isnan(np.sum(data)) then raise a warning and use np.nan_to_num to put them to zeros. This is how I do to implement point 1.

I will make a second pass when this is addressed. Then, we add tests.

himanshupathak21061998:
Sorry for making mistakes I did not knew that utils  already contains  `_safe_get_data()`

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -36,8 +38,17 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
     else:
-        mask_coords = list(np.ndindex(niimg.shape[:3]))
-        X = niimg.get_data().reshape([-1, niimg.shape[3]]).T
+        #Adding  warning message 
+        if np.isnan(np.sum(_safe_get_data(niimg))):
+            warnings.warn('The imgs you have fedded into fit_transform()' 
+                          'contains NaN values which are converted to zeroes '
+                          'please use mask_imgs to mask out NaNs othewise you '
+                           'should expect some deviation in fit_transform()')
+            mask_coords = list(np.ndindex(niimg.shape[:3]))
```
KamalakerDadi: This can be moved to where it was.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -36,8 +38,17 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+            X = _safe_get_data(niimg,True).reshape([-1, niimg.shape[3]]).T
+        else:
+            mask_coords = list(np.ndindex(niimg.shape[:3]))
```
KamalakerDadi: You don't this.

KamalakerDadi: After moving this line of code outside of if or else.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -36,8 +38,17 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
 
         X = masking._apply_mask_fmri(niimg, mask_img)
     else:
-        mask_coords = list(np.ndindex(niimg.shape[:3]))
-        X = niimg.get_data().reshape([-1, niimg.shape[3]]).T
+        #Adding  warning message 
+        if np.isnan(np.sum(_safe_get_data(niimg))):
+            warnings.warn('The imgs you have fedded into fit_transform()' 
+                          'contains NaN values which are converted to zeroes '
+                          'please use mask_imgs to mask out NaNs othewise you '
```
KamalakerDadi: 'please use .....' is not needed. We can end the message at '... converted to zeros.'

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -231,7 +242,8 @@ def __init__(self, seeds, radius=None, mask_img=None, allow_overlap=False,
         self.memory_level = memory_level
 
         self.verbose = verbose
-
+
```
KamalakerDadi: Please avoid touching unrelated lines.

PR #1898, opened 2019-01-14
FIX: Fix abc import

larsoner:
Fixes a DeprecationWarning on Python3.7 about importing from `collections.abc` instead of `collections`.


larsoner:
[Travis error](https://travis-ci.org/nilearn/nilearn/jobs/479479233#L1137) looks like an unrelated `nibabel` problem

KamalakerDadi:
>Travis error looks like an unrelated nibabel problem

Yes it is unrelated.

jeromedockes:
LGTM

PR #1901, opened 2019-01-15
[MRG] use Template.safe_substitute instead of str.replace

jeromedockes:
as was requested somewhere in the discussion of #1766 . this is purely cosmetic

https://github.com/nilearn/nilearn/pull/1766#discussion_r226188070

### DIFF
--- a/nilearn/plotting/data/html/connectome_plot_template.html
+++ b/nilearn/plotting/data/html/connectome_plot_template.html
@@ -3,9 +3,9 @@
 
 <head>
     <title>connectome plot</title>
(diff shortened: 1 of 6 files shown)


bthirion:
LGTM --though I don't know this part of the codebase well.

kchawla-pi:
For future perusers of this PR/codebase, the 4sentences that explain the benefit & pitfalls are worth reading. https://docs.python.org/2/library/string.html#string.Template.safe_substitute

kchawla-pi:
@jeromedockes I will merge this tomorrow; doesn't seem anything more needs to be done here.

PR #1905, opened 2019-01-16
[MRG] Added issue templates, indicating question shoud be posted on Neurostars

kchawla-pi:


kchawla-pi:
@GaelVaroquaux If you wanna amend the text, lemme know, else I'll merge this tomorrow morning.

emdupre:
Thanks for opening this, @kchawla-pi  !

I wonder what you would think of having two issue types: one for feature requests and the other for bug reports. Then, you could have the information about NeuroStars as introductory text in the templates themselves, either "hidden" [in the markdown text](https://raw.githubusercontent.com/ME-ICA/tedana/master/.github/pull_request_template.md) or as plain text (to be sure all users are aware of it). 

Either way, I think having the Neurostars note right at the top of the issue would be nice to make sure that users don't miss it !

kchawla-pi:
Imma merge this so I can see what this looks like in a new issue, and then we can mod it.

kchawla-pi:
@emdupre I'll wait for your feedback before merging.

emdupre:
It looks like this will yield an issue template called "Questions on NeuroStars". I'm not entirely sure I follow the logic, here.

Usually, I'd have template that explicitly list the options I would like users to use. So, for example, I'd have an issue template for "Bug Reports" and another for "Feature Requests".

Then, in the text of each of those, I'd  insert text at the top to explain that general questions about usage should be directed to NeuroStars (just in case users accidentally click on either of those options). Personally I'd do this as hidden text so they don't need to be deleted by the user, but that's a stylistic decision.

Overall, I'd argue the goal is to make users who _don't_ have a Feature Request or Bug Report spend another second thinking if GitHub is the right forum ! But if users _do_ want to file, say, a bug report I want to support that explicitly rather than implicitly (by having them click through to the "Questions on NeuroStars" template and just delete all of the text).

kchawla-pi:
Hey @emdupre That's great info, I was not very clear about how issue templates work. I experimented a bit on a dummy repo and see what you meant.
I am closing this PR and opening #1908 . Check that out .
Thanks for your help in this.

PR #1912, opened 2019-02-01
Lighting surf plot

jeromedockes:
plotly defaults for lighting of the surface plots actually look better than those we use in my opinion; main difference is more ambient light (shadows are not so dark) and no reflection (surface is more rough, less shiny, no bright spots)


jeromedockes:
newplot 1
newplot

GaelVaroquaux:
Looks great!

Maybe a note in the whats_new? I can see users being surprised at the difference between two versions.

GaelVaroquaux:
LGTM. Merging.

kchawla-pi:
Cool.
I was waiting for the CircleCI run from the previous merge to complete before merging this one.

Issue #1886, opened 2018-12-19
Inconsistencies between view_connectome and plot_connectome


GaelVaroquaux:
There are a few inconsistencies between plot_connectome and view_connectome that means that one is not interchangeable with the other. In particular, I note:

* node_coords is called coords in view_connectome
* the colormaps differs.

There might be more. This is just what I saw at a cursory glance.

GaelVaroquaux:
> Which one should we support going forward?

plot_connectome: it has been around for longer.

> We will also add a deprecation warning for the other.

Yes.

> Which other method/functions would be candidates for checking this?

I don't know. I just discover these things as I go: as a user, they smell
weird.

meghalD:
What do the axis range suggest in view_connectome. Is it correlation ? If yes, it should be within (+1,-1) , but as I plot it goes beyond the range.
am I misinterpreting the plot ? Does view_connectome show functional connectivity like plot_connectome ?

bthirion:
As far as I can see, `view_connectome` does not make assumptions on the displayed statsitics nor on its range. If you use inverse covariance or tangent representations, or even simply some z-transform of correlations the values will not be confined in [-1, 1].

meghalD:
> As far as I can see, `view_connectome` does not make assumptions on the displayed statsitics nor on its range. If you use inverse covariance or tangent representations, or even simply some z-transform of correlations the values will not be confined in [-1, 1].

Yes, I got the range [-1,+1] on applying partial correlation. So the colors here represent the level of correlation. right?
Also, why is this not the range for sparse inverse covariance? Is there any way out to transform range for inverse covariance.

bthirion:
> Yes, I got the range [-1,+1] on applying partial correlation. So the colors here represent the level of correlation. right?

Indeed.

bthirion:
> Also, why is this not the range for sparse inverse covariance? Is there any way out to transform range for inverse covariance.

You should not use inverse covariance, which is hard to interpret, especially in high dimension.
Note that partial correlations are inverse covariance estimated reweighted by the diagonal of the inverse covariance matrix (aka partial variance). Partial correlations are thus between -1 and  +1. See e.g. http://nilearn.github.io/connectivity/connectome_extraction.html

jeromedockes:
> * node_coords is called coords in view_connectome
> 
> * the colormaps differs.

these two seem very easy to fix; is someone willing to open a pr?
[...]

PR #1913, opened 2019-02-05
[MRG] Make parameter names in view_connectome() consistent with plot_connectome()

kchawla-pi:
For issue #1886 

Deprecated view_connectome params that differed from params of plot_connectome, docstrings & tests pending.

### DIFF
--- a/nilearn/plotting/html_connectome.py
+++ b/nilearn/plotting/html_connectome.py
@@ -80,6 +80,9 @@ def _make_connectome_html(connectome_info, embed_js=True):
 
 
 def _deprecate_params_view_connectome(func):
+    """ Decorator to deprecate spcific parameters in view_connectome()
+     without modifying view_connectome().
+     """
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         _warn_deprecated_params_view_connectome(kwargs)
@@ -156,6 +159,8 @@ def view_connectome(adjacency_matrix, node_coords, edge_threshold=None,
 
 
 def _warn_deprecated_params_view_connectome(kwargs):
+    """ For view_connectome(), raises warnings about deprecated parameters.
+    """
     all_deprecated_params = {'coords': 'node_coords',
                              'threshold': 'edge_threshold',
                              'cmap': 'edge_cmap',
@@ -171,12 +176,15 @@ def _warn_deprecated_params_view_connectome(kwargs):
                                                             replacement_param,
                                                             )
         )
-        warnings.warn(category=FutureWarning,
+        warnings.warn(category=DeprecationWarning,
                       message=param_deprecation_msg,
                       stacklevel=3)
 
 
 def _transfer_deprecated_param_vals_view_connectome(kwargs):
+    """ For view_connectome(), reassigns new parameters the values passed
+    to their corresponding deprecated parameters.
+    """
     coords = kwargs.setdefault('coords', None)
     threshold = kwargs.setdefault('threshold', None)
     cmap = kwargs.setdefault('cmap', None)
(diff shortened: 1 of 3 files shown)


nilearn/plotting/html_connectome.py
```diff
@@ -134,13 +148,51 @@ def view_connectome(adjacency_matrix, coords, threshold=None,
...
+                             'threshold': 'edge_threshold',
+                             'cmap': 'edge_cmap',
+                             'marker_size': 'node_size',
+                             }
+    used_deprecated_params = set(kwargs.keys()
+                                 ).intersection(all_deprecated_params.keys())
+    for deprecated_param_ in used_deprecated_params:
+        replacement_param = all_deprecated_params[deprecated_param_]
+        param_deprecation_msg = (
+            'The parameter "{}" will be removed in a future version of Nilearn.'
```
bthirion: Shouldn't we state which version ?

kchawla-pi: For Python2 deprecation we hadn't so I didn't here either.
From what I understand we display deprecation warnings for 2 releases and start clarifying the feature etc will be removed in the next release.

I assume we will handle this similarly.
I am happy to add a specific version now, if you tell me which one it should be. Nilearn 0.7.0? Or do we include point releases in next version?

bthirion: I'd go for 0.7.0

jeromedockes: the sooner we make the change the less code it will break

kchawla-pi:
Yeah I'll fix this on Tuesday when I'm back.
@jeromedockes if you have feedback regarding how soon the deprecated parameters should be removed, let us know.

kchawla-pi:
@bthirion
 @KamalakerDadi mirrored what  @jeromedockes has mentioned here.
The function wa introduced in Nilearn 0.5.0 so not a lot of people will be using it.
 believe even less will be using keyword arguments with it.
I don't think it will make much difference whether we remove the old parameter names in 0.6 or 0.7. What do you want?

bthirion:
0.6 is fine  if everybody feels it's better.

jeromedockes:
> I don't think it will make much difference whether we remove the old parameter names in 0.6 or 0.7. What do you want?

I think I would go for 0.6 but it's really up to you!

kchawla-pi:
I am now implementing the change in default colormap. How should I implement that? I am going with always showing the warning for now, but I think there is space for some discussion about more nuance here.

jeromedockes:
> I am now implementing the change in default colormap. How should I implement that? I am going with always showing the warning for now, but I think there is space for some discussion about more nuance here.

I'm not sure you need a warning for that; you can just change it and add a note in the whats_new and the docstring. if someone doesn't like the new colors they will easily figure out that the default colormap has changed.

kchawla-pi:
Once the tests are green I will merge this tomorrow.

jeromedockes:
do you plan to  address the similar issue with view_markers in a separate PR?

jeromedockes:
I see you have kept the same default colormap, is that consistent with plot_connectome?

kchawla-pi:
No, i noticed I hadn't changed it after posting this. I will change the colormap before merging.

kchawla-pi:
> do you plan to address the similar issue with view_markers in a separate PR?

Yeah I am working on that one as we speak.

jeromedockes:
I think tests outside of test_params_deprecation_view_connectome should use the new parameter names, e.g. test_view_connectome uses marker_size

kchawla-pi:
Yeah, that should have changed by the IDE everywhere I'll do that.

kchawla-pi:
This seems ready to be merge. Is it, @jeromedockes ?

jeromedockes:
LGTM

PR #1914, opened 2019-02-06
Remove temporary variable imgs_ from memory

nagaflokhu:
transform_single_imgs loads a (potentially large) dataset into the imgs_ variable, potentially uses it for a couple of things, and then does not use it anymore. filter_and_extract, which gets called later in transform_single_imgs, loads the same 3D/4D scan into a different variable (local to filter_and_extract). This leads to two separate copies of a potentially large dataset being loaded into memory while transform_single_imgs runs. `del imgs_' ensures only a single copy is in memory at a time.


GaelVaroquaux:
Very nice! That's terribly useful.

Maybe you could add a small comment above to say why we do this (to free memory). Have in mind the non Python expert who reads the code.

bthirion:
Great. merging.

kchawla-pi:
There should be a whats new entry for this: `transform_single_imgs` consumes less memory or something to that effect.

PR #1916, opened 2019-02-06
fix: quote iframe attributes

### LINKED ISSUE #1915: Change iframe call for rendering in Jupyter Book
I would like to render interactive brainsprites generated from `view_img` in [Jupyter Book](https://github.com/jupyter/jupyter-book). Jupyter Book is built using Jekyll, and converts executed notebooks to markdown using nbconvert. The nbconvert is then converted to stylized HTML using kramdown.

My difficulty comes that kramdown [cannot currently handle unquoted attributes](https://github.com/jupyter/jupyter-book/issues/72) (even though these are valid HTML!). When we generate the iframe in which the brainsprite viewer is embedded here in nilearn, [we don't quote the `width` and `height` attributes](https://github.com/nilearn/nilearn/blob/ed86f29305a264693156826f0e429b2d6281eeaf/nilearn/plotting/js_plotting_utils.py#L126).

I'd like to open a simple fix to quote these two attributes, and smooth nilearn use with Jupyter Book. Would that be alright?

emdupre:
Closes #1915 

- Quotes iframe attributes for `width` and `height`.

### DIFF
--- a/nilearn/plotting/js_plotting_utils.py
+++ b/nilearn/plotting/js_plotting_utils.py
@@ -123,7 +123,7 @@ class HTMLDocument(object):
         if height is None:
             height = self.height
         escaped = escape(self.html, quote=True)
-        wrapped = ('<iframe srcdoc="{}" width={} height={} '
+        wrapped = ('<iframe srcdoc="{}" width="{}" height="{}" '
                    'frameBorder="0"></iframe>').format(escaped, width, height)
         return wrapped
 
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
LGTM. No issue with this. I don't think that it can cause problems in other web environments. Can it?

emdupre:
I'm still new to web development, but looking at [Mozilla's introduction to HTML](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Getting_started#Omitting_quotes_around_attribute_values), they suggest that quoted are attributes are safer (though both quoted and unquoted can be valid HTML).

bthirion:
LGTM too.

GaelVaroquaux:
Great! Merging.

PR #1918, opened 2019-02-12
[MRG] view_markers() parameters consistent with add_markers()

kchawla-pi:
(As per issue #1886 )

Changed params:  
coords -> marker_coords
colors -> marker_color

### DIFF
--- a/nilearn/plotting/html_connectome.py
+++ b/nilearn/plotting/html_connectome.py
@@ -7,7 +7,7 @@ from scipy import sparse
 from .. import datasets
 from . import cm
 
-from nilearn.plotting.js_plotting_utils import (add_js_lib, HTMLDocument, mesh_to_plotly,
+from .js_plotting_utils import (add_js_lib, HTMLDocument, mesh_to_plotly,
                                 encode, colorscale, get_html_template,
                                 to_color_strings)
 
--- a/nilearn/plotting/html_connectome.py
+++ b/nilearn/plotting/html_connectome.py
@@ -4,10 +4,10 @@ import warnings
 
 import numpy as np
 from scipy import sparse
-from .. import datasets
-from . import cm
+from nilearn import datasets
+from nilearn.plotting import cm
 
-from .js_plotting_utils import (add_js_lib, HTMLDocument, mesh_to_plotly,
+from nilearn.plotting.js_plotting_utils import (add_js_lib, HTMLDocument, mesh_to_plotly,
                                 encode, colorscale, get_html_template,
                                 to_color_strings)
 
@@ -144,7 +144,7 @@ def view_connectome(adjacency_matrix, coords, threshold=None,
 
 
 def _deprecate_params_view_markers(func):
-    """ Decorator to deprecate spcific parameters in view_markers()
+    """ Decorator to deprecate specific parameters in view_markers()
      without modifying view_markers().
      """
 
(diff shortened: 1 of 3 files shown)


kchawla-pi:
Once the CI is green, this seems ready to be merge. Is it, @jeromedockes ?

nilearn/plotting/html_connectome.py
```diff
@@ -1,11 +1,13 @@
+import functools
 import json
+import warnings
 
 import numpy as np
 from scipy import sparse
-from .. import datasets
-from . import cm
+from nilearn import datasets
```
jeromedockes: we prefer relative imports

jeromedockes: why did this change? it seems unrelated

kchawla-pi: We should switch that. PEP8 endorses Absolute imports.
Relative imports misbehave/break more and are less clear.

kchawla-pi: Relative imports tend to break more for reasons that are not readily apparent, especially within IDEs. 
This one broke when I was doing this PR.

jeromedockes: this should be discussed in a separate issue then.

kchawla-pi: Okay, anything else for this PR?

jeromedockes: nope, it looks good otherwise.

nilearn/plotting/html_connectome.py
```diff
@@ -241,8 +255,45 @@ def view_markers(coords, colors=None, marker_size=5.):
...
+    used_dperecated_params = set(kwargs).intersection(all_deprecated_params)
+    for deprecated_param_ in used_dperecated_params:
+        replacement_param = all_deprecated_params[deprecated_param_]
+        param_deprecation_msg = (
+            'The parameter "{}" will be removed in Nilearn version 0.6.0. '
+            'Please use the parameter "{}" instead.'.format(deprecated_param_,
+                                                            replacement_param,
+                                                            )
+        )
+        warnings.filterwarnings('always', message=param_deprecation_msg)
```
jeromedockes: shouldn't you use the warnings context manager here? when does this filter get removed?

nilearn/plotting/html_connectome.py
```diff
@@ -241,8 +255,45 @@ def view_markers(coords, colors=None, marker_size=5.):
...
+def _transfer_deprecated_param_vals_view_markers(kwargs):
+    """ For view_markers(), reassigns new parameters the values passed
+    to their corresponding deprecated parameters.
+    """
+    coords = kwargs.get('coords', None)
+    colors = kwargs.get('colors', None)
+    
+    if coords is not None:
+        kwargs['marker_coords'] = coords
+    if colors:
```
jeromedockes: please be explcit `if colors is None` to avoid problems e.g. if colors can be a numpy array

nilearn/plotting/html_connectome.py
kchawla-pi: Its specific for that message string.

nilearn/plotting/html_connectome.py
kchawla-pi: You mean `if colors is not None:` right?
Okay, I'll open a new PR and fix this in functions for both the PRs.

jeromedockes: yes, sorry. in general we need to be wary of implicit conversions to bool, mostly because of numpy arrays

PR #1924, opened 2019-02-18
Update html_stat_map.py

pbellec:
The viewer does not include the value associated with the cuts, even if `annotate` is `True`. Minor update in the doc to reflect that behavior.


jeromedockes:
thanks! +1 for merge

GaelVaroquaux:
LGTM. Thank you!

PR #1936, opened 2019-03-06
therehold_map_ratio should not change input image

### LINKED ISSUE #1933: _threshold_maps_ratio changes the input map
My maps images keep changing when I use RegionExtractor. I think we need to make a copy [here](https://github.com/nilearn/nilearn/blob/master/nilearn/regions/region_extractor.py#L58)
For instance the following code throws an `AssertionError: Arrays are not equal`
```Python
from nilearn._utils.data_gen import generate_maps
import numpy as np
from nilearn.regions.region_extractor import _threshold_maps_ratio

maps, mask_img = generate_maps((10, 10, 10), 30)
maps.get_data()[:5] = 100
maps_data = maps.get_data().copy()
thresholded_maps = _threshold_maps_ratio(maps, threshold=1)
np.testing.assert_array_equal(maps.get_data(), maps_data)

```

salma1601:
fix #1933

### DIFF
--- a/nilearn/regions/tests/test_region_extractor.py
+++ b/nilearn/regions/tests/test_region_extractor.py
@@ -51,10 +51,15 @@ def test_threshold_maps_ratio():
     # smoke test for function _threshold_maps_ratio with randomly
     # generated maps
 
+    # Check that the input maps are not changed
+    maps, _ = generate_maps((6, 8, 10), n_regions=3)
+    maps.get_data()[:3] = 100
+    maps_data = maps.get_data().copy()
+    thr_maps = _threshold_maps_ratio(maps, threshold=1.0)
+    np.testing.assert_array_equal(maps.get_data(), maps_data)
+
     # make sure that n_regions (4th dimension) are kept same even
     # in thresholded image
-    maps, _ = generate_maps((6, 8, 10), n_regions=3)
-    thr_maps = _threshold_maps_ratio(maps, threshold=1.0)
     assert_true(thr_maps.shape[-1] == maps.shape[-1])
 
     # check that the size should be same for 3D image
--- a/nilearn/regions/tests/test_region_extractor.py
+++ b/nilearn/regions/tests/test_region_extractor.py
@@ -51,8 +51,9 @@ def test_threshold_maps_ratio():
     # smoke test for function _threshold_maps_ratio with randomly
     # generated maps
 
-    # Check that the input maps are not changed
     maps, _ = generate_maps((6, 8, 10), n_regions=3)
+
+    # test that there is no side effect
     maps.get_data()[:3] = 100
     maps_data = maps.get_data().copy()
     thr_maps = _threshold_maps_ratio(maps, threshold=1.0)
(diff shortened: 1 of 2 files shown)


nilearn/regions/tests/test_region_extractor.py
```diff
@@ -51,10 +51,15 @@ def test_threshold_maps_ratio():
...
     # generated maps
 
-    # make sure that n_regions (4th dimension) are kept same even
-    # in thresholded image
+    # Check that the input maps are not changed
     maps, _ = generate_maps((6, 8, 10), n_regions=3)
+    maps.get_data()[:3] = 100
+    maps_data = maps.get_data().copy()
     thr_maps = _threshold_maps_ratio(maps, threshold=1.0)
+    np.testing.assert_array_equal(maps.get_data(), maps_data)
```
bthirion: Have you checked that this test fails with the current codebase ?
Looks good otherwise.

salma1601: yes, and this is why I added the line `maps.get_data()[:3] = 100`. Otherwise actually the thresholded maps are exactly equal to the maps, so the test was passing with the current codebase.

salma1601: what I don't understand is that I was assuming that _safe_get_data is precisely supposed to avoid such side effects but this doesn't work ! I didn't go deep into this but might be interesting to have a look at _safe_get_data behaviour

salma1601: may be I am missing something here

GaelVaroquaux: No, _safe_get_data avoids that the in-cache vs out-of-cache status of an image changes. But it doesn't make a copy.

nilearn/regions/tests/test_region_extractor.py
```diff
@@ -51,10 +51,15 @@ def test_threshold_maps_ratio():
     # smoke test for function _threshold_maps_ratio with randomly
     # generated maps
 
-    # make sure that n_regions (4th dimension) are kept same even
-    # in thresholded image
+    # Check that the input maps are not changed
     maps, _ = generate_maps((6, 8, 10), n_regions=3)
+    maps.get_data()[:3] = 100
+    maps_data = maps.get_data().copy()
     thr_maps = _threshold_maps_ratio(maps, threshold=1.0)
```
GaelVaroquaux: It would be good to have a small comment above: "test that there is no side effect".

nilearn/regions/tests/test_region_extractor.py
salma1601: Ah, Ok ! Thank you for the explanation

nilearn/regions/tests/test_region_extractor.py
salma1601: done

GaelVaroquaux: Sorry, I hadn't seen your comment above :).

GaelVaroquaux:
LGTM. Merging. Thank you

Issue #1940, opened 2019-03-11
Improve docstring entry for NiftiMasker parameters


alpinho:
In nilearn.input_data.NiftiMasker class, many parameters point to documentation details of other functions. A link to the documentation of these functions should be provided.

Example:
low_pass: None or float, optional 
This parameter is passed to signal.clean. Please see the related documentation for details.

In this case, a link to the function "signal.clean' shall be given.

kchawla-pi:
I think the docstring should have some more detail right here, beyond "This parameter is passed to signal.clean."

tsalo:
I believe that this should be closed by #1942, but the keyword close wasn't triggered.

PR #1942, opened 2019-03-11
ENH: added links to signal.clean() in NiftiMasker docstring

alpinho:
solve issue #1940


jeromedockes:
could you try with the trailing underscore for links to footnotes?  e.g. `[1]_`

http://docutils.sourceforge.net/docs/user/rst/quickref.html#footnotes

thanks!

kchawla-pi:
Once #2096 has been merged, pull from master.

kchawla-pi:
@jeromedockes Something weird is happenning. I checked the built page (this one is localy built. The one on CIrcleCI is adding some alphanumeric identifier to the link name), this is what I see:

image

There should be only 4 links not 5. INfact I was not expecting there to be any `(1, 2, 3,...) at all.

jeromedockes:
> this is what I see:

I built the doc locally and see the same thing as you. maybe a sphinx issue?
anyway I think it would be best to replace these footnotes with 
```
:func:`nilearn.signal.clean`
```

which will link directly to the documentation for `clean`. also because the link in the footnote here is the url for the nilearn online documentation, whereas a relative path would be better if someone is reading the documentation elsewhere

nilearn/input_data/nifti_masker.py
```diff
@@ -92,19 +92,19 @@ class NiftiMasker(BaseMasker, CacheMixin):
 
     detrend : boolean, optional
         This parameter is passed to signal.clean. Please see the related
-        documentation for details
+        documentation for details. [1]_
```
jeromedockes:
```suggestion
        documentation for details: :func:`nilearn.signal.clean`.
```

alpinho:
@jeromedockes, I have addressed your comments accordingly. Can you revise it? @kchawla-pi @bthirion

bthirion:
Thx !

PR #1944, opened 2019-03-12
[MRG] add symmetric_cmap option in view_img_on_surf

jeromedockes:
after all I think the symmetric_cmap option is useful for `view_img_on_surf`

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -34,7 +34,8 @@ Changes
  - colors is now marker_color
 
 - :func:`nilearn.plotting.view_img_on_surf` now accepts a `symmetric_cmap`
-  argument to control whether the colormap is centered around 0.
+  argument to control whether the colormap is centered around 0 and a `vmin`
+  argument.
 
 Fixes
 -----
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -33,6 +33,9 @@ Changes
  - coords is now marker_coords
  - colors is now marker_color
 
+- :func:`nilearn.plotting.view_img_on_surf` now accepts a `symmetric_cmap`
+  argument to control whether the colormap is centered around 0.
+
 Fixes
 -----
 
--- a/nilearn/plotting/html_surface.py
+++ b/nilearn/plotting/html_surface.py
@@ -126,7 +126,7 @@ def _fill_html_template(info, embed_js=True):
 
 def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
                      threshold=None, cmap=cm.cold_hot,
-                     black_bg=False, vmax=None):
+                     black_bg=False, symmetric_cmap=True, vmax=None):
     """
     Insert a surface plot of a statistical map into an HTML page.
 
@@ -158,6 +158,10 @@ def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
         If True, image is plotted on a black background. Otherwise on a
         white background.
 
+    symmetric_cmap : bool, optional (default=True)
+        Make colormap symmetric (ranging from -vmax to vmax).
+        You can set it to False if you are plotting only positive values.
+
     vmax : float or None, optional (default=None)
         upper bound for the colorbar. if None, use the absolute max of the
         brain map.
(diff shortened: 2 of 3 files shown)


kchawla-pi:
What if one wants to plot only negative values?

jeromedockes:
> What if one wants to plot only negative values?

it is handled as in other functions, for now they have to flip it I guess

kchawla-pi:
Should we mention that in the doc string?

Kshitij

jeromedockes:
do you think it is a frequent use case?

kchawla-pi:
> do you think it is a frequent use case?

I don't know that, but if it is a 1 line addition to the docstring, I would rather have it documented than not.

jeromedockes:
> I don't know that, but if it is a 1 line addition to the docstring, I would rather have it documented than not.

sure, what would you write in the docstring? I can add it

kchawla-pi:
> it is handled as in other functions, for now they have to flip it I guess

If you tell me exactly which args to pass, I can come up with the language.

jeromedockes:
> If you tell me exactly which args to pass, I can come up with the language.

it's not arguments to pass to this function, it's just that symmetric_cmap=False is only accepted if the map is positive, as stated in the docstring. If they have a negative map, it is not supported but I guess we could add that. If the map is neither positive nor negative, symmetric_cmap should be forced to be True, as is currently the case

kchawla-pi:
I suppose if no one has filed an issue asking for it, we should not feature bloat.

jeromedockes:
so is there anything else?

kchawla-pi:
Shouldn't we add the param at the end? WOn't the current placement break positional arguments?

kchawla-pi:
The code changes LGTM. WOuld you tell me the path to the bult image so I can look at it as well in CircleCI artifacts?

jeromedockes:
sorry I'm not sure I know, what do I need to do?

kchawla-pi:
Never mind that.

We should at this feature in one of the examples.
We should also make an entry in What's New.

jeromedockes:
> We should at this feature in one of the examples.

maybe we need an example dedicated to interactive plots, but that would be a separate PR. I'm not sure showing this in one of the existing examples is a good idea because it would add clutter to an example to explain a feature which is not essential and easy to understand from the docstring

kchawla-pi:
OK. 

The current placement of `vmin` in function signatures breaks positional arguments in certain functions.
The ones I notice are:
 1. view_surf
 2. full_brain_info
 3. view_img_on_surf
 4. full_brain_info
 
In addition, the docstring for `vmin` says:
```
       If `symmetric_cmap` is `True`, `vmin` is always equal to `-vmax` and
        cannot be chosen.
        If `symmetric_cmap` is `False`, `vmin` defaults to the min of the
        image, or 0 when a threshold is used.
``` 
But I don't see the code logic for implementing this. What am I missing?

jeromedockes:
> OK.
> 
> The current placement of `vmin` in function signatures breaks positional arguments in certain functions.
> The ones I notice are:
> 
>     1. view_surf
> 
>     2. full_brain_info
> 
>     3. view_img_on_surf
> 
>     4. full_brain_info

- in view_surf it doesn't, last two arguments (vmin and symmetric_cmap) are both being added in this PR
- full_brain_info isn't exposed to users from the `plotting` package

jeromedockes:
> But I don't see the code logic for implementing this. What am I missing?

it's here:

https://github.com/nilearn/nilearn/blob/4a072a3e0cffa71e83f979de05a2735e8228b89a/nilearn/plotting/js_plotting_utils.py#L205

kchawla-pi:
OK. Should we rename `full_brain_info` to `_full_brain_info`, maybe another PR?

Dunno how I missed the code logic for ` if symmetric_cmap and vmin is not None: `.
Since `vmin` is None or float, we can simply write
```python
 if symmetric_cmap and vmin: 
```
Also, instead of setting vmin to None, shoud we set it to -vmax? 
If not, should we change the warning to:
```python
  warnings.warn('vmin cannot be chosen when cmap is symmetric. Setting vmin to None.')
```

kchawla-pi:
I'll let you decide this, I don't think I have anything to add.

jeromedockes:
> Also, instead of setting vmin to None, shoud we set it to -vmax?

it gets changed to -vmax later, I guess we could just remove this line. but I don't think it matters very much. I would keep the warning as it is because vmin will be set to -vmax.

jeromedockes:
> OK. Should we rename `full_brain_info` to `_full_brain_info`, maybe another PR?

yes that would make sense. in another PR. it's not urgent since this not exposed in plotting, which is what users import

jeromedockes:
ping @kchawla-pi did you have more comments on this?

kchawla-pi:
Nope. Take it home Jerome.

PR #1946, opened 2019-03-15
[MRG] Replace deprecated param 'cachedir' with 'location' in joblib's Memory object

kchawla-pi:


bthirion:
But travis tells us we're no longer compliant with old joblib. should frop support for it I think.

kchawla-pi:
We may have to upsklearn etc for that as well. See the comment for the other similar PR.

KamalakerDadi:
It looks like the way it should be addressed is import ```Memory``` object depending upon the deprecation version.

kchawla-pi:
That is a metric ton of compatibility shims. We will probably have to engineer one for most warnings. That's a ton of code for very little gain.
We either shelve solving the warnings until we drop support for Py2 & 3.4, or we update the minimum requirement for numpy now ( I doubt Joblib version bump will be a user facing problem) However there may be more version bumps in other libraries for other warnings.

kchawla-pi:
@GaelVaroquaux thoughts?

kchawla-pi:
We have been showing the Python2, 3.4 DeprecationWarning for 5 months now.
I think we should quickly do a new release with whatever is in master now, updating the warnings to warn that support will be dropped from next release onwards.

Then we drop support and I fix the warnings before the sprint.

User using it individually can choose not to upgrade, or roll back any upgrade easily, without non-trivial consequences.
Anyone using Nilearn in a serious pipeline, and on servers, it is their responsibility not to deploy new versions of key software without reading up and testing about the changes.
I think the fallout will be minimal, and our responsibility in that is nil.

bthirion:
I agree with your comments, but I think we shouldn't drop support for Python 2 yet.
This probably means that we'll have to deal with a heap of warnings on CircleCI, but I have no better suggestion.

kchawla-pi:
Well we support the deprecations for the next two releases.
We have already made a release 5 months ago.
If we release the new fixes so far landed in Nilearn as v0.5.1 , that will
be the second release supporting Py2.
The release after that will come at least 4-5 months after that, v0.6.0.
This would not support Python 2.
But it will be only a few months before Python 2 is officially end of lifed.
If you don't wanna drop support in the October release, that's fine, but I
think it won't be a bad way to do it.

GaelVaroquaux:
> We either shelve solving the warnings until we drop support for Py2 & 3.4,

Fine with me!

GaelVaroquaux:
> We have been showing the Python2, 3.4 DeprecationWarning for 5 months now.
> I think we should quickly do a new release with whatever is in master now,
> updating the warnings to warn that support will be dropped from next release
> onwards.

> Then we drop support and I fix the warnings before the sprint.

Fine with me.

kchawla-pi:
Joblib==0.11 does not support `location`. We cannot remove this warning while we still support joblib.
We can save this one for Nilearn 0.7.0

emdupre:
Now that #2334 is resolved, want to re-start the testing here, @kchawla-pi ? It should just take an empty commit to trigger ! Or I can trigger it if you "allow edits from maintainers" :smile_cat:

kchawla-pi:
😄 "allow edits from maintainers" is already enabled @emdupre

kchawla-pi:
The first build in Travis is failing because Scikit-learn < 0.21 uses its own vendored version of joblib, which does not support location parameter.

Attending to that now.

kchawla-pi:
This is weird. There was no conflict locally. I pulled from upstream master and git did a fast-forward. I pushed and there is no change here. No CIs were run and  there is still a merge conflict.

I shouldn't have to use the web UI for this.

kchawla-pi:
I will first see if we can use the installed  Joblib for Nilearn while SKLearn is using its vendored Joblib internally. If that doesn't work then we can bump up Sklearn to 0.21

kchawla-pi:
Ready for review @emdupre

nilearn/_utils/compat.py
```diff
@@ -69,10 +69,7 @@ def md5_hash(string):
         return m.hexdigest()
 
 
-if sklearn.__version__ < '0.21':
```
emdupre: This is the only potential sticking point, in my mind -- is this effectively bumping up the sklearn version ?

emdupre:
Thanks, @kchawla-pi ! Only one question on my end. Not sure if you want to re-review, @bthirion ?

nilearn/_utils/compat.py
kchawla-pi: No. What I have done is, irrespective of the version of Scikit-Learn installed, Nilearn always uses the installed Joblib. Scikit-Learn < 0.21 will internally use its own vendored Joblib, independent of Nilearn code.

Unless there is some change in how the vendored Joblib handles its functionality vs the Joblib installed in the library, which I did not notice in the runtimes, this seems to be working.

bthirion:
Thx !

PR #1947, opened 2019-03-15
Fixed VisibleDeprecationWarning in np.genfromtxt by setting encoding==None

kchawla-pi:


kchawla-pi:
That very well may mean bumping the scikit-learn version as well, at least in one build config,  IIRC, all the dependencies are at the lowest supported versions.
@GaelVaroquaux

jeromedockes:
I don't think the required numpy version should be changed. 1.11 is the one required by scikit-learn, and I believe the one installed by apt-get.

GaelVaroquaux:
> That very well may mean bumping the scikit-learn version as well, at least in one build config, IIRC, all the dependencies are at the lowest supported versions.

Let's do this after the release.

GaelVaroquaux:
We are still supporting old version of numpy where "encoding" is not defined. You should do a if statement on the version of numpy.

kchawla-pi:
@GaelVaroquaux SKL is now on Numpy 1.13.3 . I am bumping our minimum numpy to 1.12 in this PR.

kchawla-pi:
The encoding parameter was introduced in numpy 1.14 . Until we bump our Numpy to that, this PR will be frozen.
Python3.5 will be EOL'd in Spetember. After Nilearn 0.7.x, we can drop support for that and bump up Numpy for 0.8.0a.

ymzayek:
Hi, I see that this PR was never followed up as I still get the VisibleDeprecationWarning in the current nilearn version. Since this PR is quite old should I close it and open a new one to make the changes?

kchawla-pi:
It couldn't be followed up at the time as the supported versions included
the old and the new interface.
It was hence tanked until such time as the versions with the old interface
were no longer supported.
If that is the case then this can now be revived, or a new one created.
There's another one similar to this.

kchawla-pi:
The similar one was for a different deprecation warning.

ymzayek:
@kchawla-pi thanks for the feedback. If I understand correctly since you mentioned before that the encoding parameter was introduced in numpy 1.14 and now we support minimum numpy 1.18 we can indeed revive this. Would you be willing to do it?

bthirion:
Hi @ymzayek thank you for reopening this discussion.
@kchawla-pi is most likely no longer available for this task, you may want to proceed. I guess the simplest thing is to open a new PR.
Best,

ymzayek:
Ok I'll go ahead. Will close this and open a new one.

PR #1951, opened 2019-03-22
[MRG] more control on surface javascript plot colorbar

jeromedockes:
here I set the hard-coded font size to a higher value.

another option would be to expose this parameter to the user.

we may also consider adding a parameter to hide the colorbar.

thoughts?


jeromedockes:
new:

newplot

old:

newplot(1)

kchawla-pi:
I vote for exposing it to users. People should be able to set the text size depending on their eye sight, among other things.

kchawla-pi:
About the colorbar, do the figures generated by this function mean anything without the colorbar? If yes, there may be merit in considering that option. If no, then there may be merit in not considering that option.

jeromedockes:
> About the colorbar, do the figures generated by this function mean anything without the colorbar?

I don't know, but plot_stat_map and plot_surf_stat_map have a colorbar kwarg to disable it

kchawla-pi:
> I don't know, but plot_stat_map and plot_surf_stat_map have a colorbar kwarg to disable it

Then we should add it here as well.

jeromedockes:
> Then we should add it here as well.

ok then, I'll do this:
- make colorbar optional
- expose font size

jeromedockes:
brain_colorbar

doc/whats_new.rst
```diff
@@ -107,10 +113,8 @@ Changes
   argument to control whether the colormap is centered around 0 and a `vmin`
   argument.
 
-- :func:`nilearn.plotting.view_img_on_surf`, :func:`nilearn.plotting.view_surf`
-  and :func:`nilearn.plotting.view_connectome` now allow disabling the colorbar,
-  and setting its height and the fontsize of its ticklabels.
-
+- Users can now control the size and fontsize of colorbars in interactive
+  surface and connectome plots, or disable the colorbar alltogether.
```
kchawla-pi: Spelling: altogether.
Disable and altogether are redundant. I would lwrite "disable the colorbar." and be done.

nilearn/plotting/html_surface.py
```diff
@@ -127,8 +127,8 @@ def _fill_html_template(info, embed_js=True):
 
 def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
                      threshold=None, cmap=cm.cold_hot,
-                     black_bg=False, vmax=None, colorbar=True,
-                     cbar_height=.5, cbar_fontsize=25):
+                     black_bg=False, vmax=None, vmin=None, symmetric_cmap=True,
```
kchawla-pi: We might've had this discussion, but view_img_on_surf seems to be a public function, won't this change break it's signature?

nilearn/plotting/js_plotting_utils.py
```diff
@@ -165,7 +165,7 @@ def open_in_browser(self, file_name=None, temp_file_lifetime=30):
 
         """
         if file_name is None:
-            fd, file_name = tempfile.mkstemp('.html', 'nilearn_surface_plot_')
+            fd, file_name = tempfile.mkstemp('.html', 'nilearn_plot_')
```
kchawla-pi: Use a context manager here.

nilearn/plotting/html_surface.py
jeromedockes: I don't think so because we are only appending kwargs at the end of the signature

nilearn/plotting/js_plotting_utils.py
jeromedockes: no because we don't want to close the file when we exit the context, or even when the function returns. here I'm just making the page's title more generic, not changing the logic of `open_in_browser`

kchawla-pi: Ok

nilearn/plotting/html_surface.py
kchawla-pi: Yeah, dunno I confused this.

nilearn/plotting/html_surface.py
```diff
@@ -127,8 +127,8 @@ def _fill_html_template(info, embed_js=True):
 
 def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
                      threshold=None, cmap=cm.cold_hot,
-                     black_bg=False, vmax=None, vmin=None,
-                     symmetric_cmap=True):
+                     black_bg=False, vmax=None, vmin=None, symmetric_cmap=True,
+                     colorbar=True, cbar_height=.5, cbar_fontsize=25):
```
kchawla-pi: We are calling it `colorbar` and also `cbar`.
We should call it `colorbar`, `colorbar_height`, `colorbar_fontsize`.
Saving a few extra characters isn't worth the clarity, especially when most users will rely on autocomplete.

nilearn/plotting/html_surface.py
```diff
@@ -195,12 +204,16 @@ def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
         volume_img=stat_map_img, mesh=surf_mesh, threshold=threshold,
         cmap=cmap, black_bg=black_bg, vmax=vmax, vmin=vmin,
         symmetric_cmap=symmetric_cmap)
+    info['colorbar'] = colorbar
+    info['cbar_height'] = cbar_height
+    info['cbar_fontsize'] = cbar_fontsize
```
kchawla-pi: Same here, make it `colorbar_height` and so on.

kchawla-pi: There are are more instances of cbar_..., this comment applies to all of them.

kchawla-pi:
CircleCI might be failing because localizer server is down. Hopefully they are transitioning it to OSF at this very moment. Build the docs locally before pushing/confirm it is building localy and once we are ready, I will do one more review and merge.

nilearn/plotting/html_surface.py
```diff
@@ -205,15 +205,15 @@ def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
         cmap=cmap, black_bg=black_bg, vmax=vmax, vmin=vmin,
         symmetric_cmap=symmetric_cmap)
     info['colorbar'] = colorbar
-    info['cbar_height'] = cbar_height
-    info['cbar_fontsize'] = cbar_fontsize
+    info['cbar_height'] = colorbar_height
+    info['cbar_fontsize'] = colorbar_fontsize
```
kchawla-pi: The key names should be changed as well.

nilearn/plotting/html_surface.py
jeromedockes: I did it for all python names. the dictionary keys are for javascript functions in the generated pages, never seen by users so I didn't change them to avoid changing the javascript

kchawla-pi:
Ok.

jeromedockes:
ping @kchawla-pi

seralouk:
Hi there just a question. I see in `https://github.com/nilearn/nilearn/blob/master/nilearn/plotting/html_surface.py#L225` that there are some nice attributes available. 

I have installed using pip the latest version `0.5.2` but these input arguments are not available. 

When are you planning to release it?

emdupre:
Hi @seralouk -- we currently have a [beta release on pip](https://pypi.org/project/nilearn/0.6.0b0/), accessible with 

`pip install nilearn==0.6.0b0`

It'd be great to get your feedback if you try out this feature there !

seralouk:
Just installed it. Great thank you. Will keep you posted for any unexcepted issues. Cheers

seralouk:
Could you have al look at this: https://github.com/nilearn/nilearn/issues/2220 ?

kchawla-pi:
@seralouk You can generally install the latest prerelease version using
 `pip install --pre <package>`.
We'll be releasing another beta (b1) soon as well.

PR #1959, opened 2019-04-04
Addition of cortical surface-based searchlight, with example

SylvainTakerkart:


examples/02_decoding/plot_intertva_surface_searchlight.py
```diff
@@ -0,0 +1,122 @@
+"""
+Cortical surface-based searchlight decoding of vocal vs. non-vocal
+stimuli in InterTVA experiment
+https://openneuro.org/datasets/ds001771/
```
GaelVaroquaux: Title too long: this will raise an error.

nilearn/surf_masking.py
```diff
@@ -0,0 +1,109 @@
...
+    """Same as apply_mask().
+
+    The only difference with apply_mask is that some costly checks on mask_img
+    are not performed: mask_img is assumed to contain only two different
+    values (this is checked for in apply_mask, not in this function).
+    """
+
+    return giimgs[surfmask,:].T
+
+
```
GaelVaroquaux: Remove empty lines here

nilearn/decoding/tests/test_surf_searchlight.py
```diff
@@ -13,7 +14,7 @@
 
 
 from nilearn.surface.tests.test_surface import _generate_surf
-from nilearn.decoding.surf_searchlight import SurfSearchLight
+from nilearn.decoding.surf_searchlight import SurfSearchLight, _apply_surfmask_and_get_affinity
```
kchawla-pi: PEP8 standards dictate that each line has no more than 80chars (col 79).
You should rewrite this, and others such as this, as
```python
from nilearn.decoding.surf_searchlight import (SurfSearchLight, 
                                            _apply_surfmask_and_get_affinity,)
```

nilearn/decoding/tests/test_surf_searchlight.py
```diff
@@ -13,7 +14,7 @@
...
 
 def _create_toy_mesh():
```
kchawla-pi: Add a docstring here. That helps other developers quickly understand the code base.

SylvainTakerkart:
Hey guys, I don't really understand why travis and appveyor fail (apparantely for the same reason, pasted below):
```
File "C:\Miniconda35-x64\lib\site-packages\nilearn\decoding\tests\test_surf_searchlight.py", line 16, in <module>
    from ...surface.tests.test_surface import _generate_surf
ImportError: cannot import name '_generate_surf
```

When run with pytest, this test passes without problem...
I also tried with an absolute path for the import of this function, but it failed also...

Any hint?

Thanks,

kchawla-pi:
@SylvainTakerkart The current problem with the fetcher PR is the duplication of the dataset. I am working on that. That should not prevent you from wrapping up this PR. What say?

kchawla-pi:
Hi @SylvainTakerkart , I just ran the `fetch_surf_tva_localizer` function & I don't see a duplication problem. There're only the uncompressed files, not the compressed one.
Would you delete the existing material pertaining to this dataset & run it again?

examples/02_decoding/plot_intertva_surface_searchlight.py
```diff
@@ -0,0 +1,106 @@
...
+radius = 8
+
+# Define the searchlight "estimator"
+searchlight = nilearn.decoding.SurfSearchLight(mesh_data,
+                                               surfmask_tex_data,
+                                               process_surfmask_tex=surfmask_tex_data,
+                                               radius=radius,
+                                               verbose=1,
+                                               n_jobs=n_jobs,
+                                               cv=10)
```
danjgale: If the example takes too long to run, you may want to consider using `cv=3` (as with the Haxby volume searchlight example)

SylvainTakerkart: Thanks! Computation time is in fact very reasonable (because the mesh is fairly coarse)...

kchawla-pi:
Once that PR is green lit, we can also merge that into this one. I have pinged Gael there.

nilearn/decoding/surf_searchlight.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # Get the data ready to go into the SurfSearchLight.fit function
+    X = giimgs_data.T
+
+    return X, A
+
+
+##############################################################################
+### Class for surface searchlight ############################################
+##############################################################################
```
kchawla-pi: Unnecessary. The class name is self-explanatory

nilearn/decoding/tests/test_surf_searchlight.py
```diff
@@ -0,0 +1,175 @@
...
+    # each sphere should contain all vertices of the mesh
+    radius = 2
+    X, A = _apply_surfmask_and_get_affinity(mesh_coords, giimgs_data, radius, surfmask=surfmask)
+    assert_array_equal(X,func_data.T)
+    assert_array_equal(A.toarray(),np.ones([n_vertex, n_vertex])[np.arange(0,n_vertex/2,dtype=int)])
+
+
+
+
+
```
kchawla-pi: This should only be 2 empty lines, not so many

emdupre:
It looks like this git history got a bit muddied. To confirm: Is it still alive, or was it superseded by #2435 ?

bthirion:
AFAIK, it was superseded by #2435.
The code of #2435 may be factorized in the future, but for the moment it offers a great tradeoff between simplicity and added value.

SylvainTakerkart:
Hi!

I'm following @bthirion 's lead here... Yep, the example of #2435 is simple and does the job... However, the present PR might serve in the future to follow @bthirion 's suggestion to encapsulate the example of #2435 in a "light SurfSearchlight object"? Coz the code in the present PR is not far from being already this, no? (or you imagine something even lighter?)

emdupre:
Unfortunately the git history is such that it's currently impossible to quickly find the code directly related to this PR (the diff has almost 27k lines added in 1470 commits).

I'm happy to help clean up this history, but I'm wondering if it wouldn't make more sense to instead just close this PR. We could then continue discussions in #2435 and eventually open a new PR with the specific changes you'd like to make @SylvainTakerkart. WDYT ?

SylvainTakerkart:
sure!

emdupre:
Ok, closing now. Thanks so much for all of your work on this @SylvainTakerkart !

PR #1962, opened 2019-04-09
[MRG+1] Seitzman atlas fetcher

fliem:
This PR adds a fetcher for the [Seitzman ROI coordinates](http://doi.org/10.1101/450452). An example also is appended to the `plot_sphere_based_connectome`.

The original atlas consist of two text files which need some wrangling that I found to be a headache with numpy. I therefore created an extra script (`datasets/data/seitzman_2018/merge_seitzman_files.py`) that uses pandas which takes the original files and writes a nicely formatted csv file (`datasets/data/seitzman_2018/seitzman_2018_rois.csv`). Only this csv file is then used in the fetcher.

In this atlas, not all regions have the same radius (some are 4, some 5 mm). A future PR will make the sphere_masker take a sequence of `len(coords)` as an input for radius (in addition to floats).

Let me know what you think.


fliem:
Hi @kchawla-pi 
The tests seem to fail because of neurovault issues. Do you have any experience with that?
Thanks.

kchawla-pi:
Hi @fliem ! Thanks for doing this PR!
The error log indicates that one test for you PR is failing because it cannot find the file:

`IOError: C:\Miniconda-x64\lib\site-packages\nilearn\datasets\data\seitzman_2018\seitzman_2018_rois.csv not found.`

I don't see a network error. Would you verify that the file exists in the online git repo or is part of the downloaded dataset when it is downloaded afresh?

fliem:
Thanks for getting back. This file is there. I have added 
```
          package_data={'nilearn.datasets.data': ['*.nii.gz', '*.csv'],
                        ...
                        'nilearn.datasets.data.seitzman_2018': ['*.csv'],
```
to `setup.py`.

I found out that if I install my branch with `pip install`, I get the same error, if I install it with `git install -e`, the csv file is found. Any idea what's the issue?

kchawla-pi:
I will look at the code later, off the top of my head to include non .py files in a package, they must be listed in MANIFEST.in.

nilearn/datasets/data/seitzman_2018/merge_seitzman_files.py
```diff
@@ -0,0 +1,30 @@
+"""This script takes original files from Seitzman atlas
+    - ROIs_300inVol_MNI_allInfo.txt
+    - ROIs_anatomicalLabels.txt
+and creates one nicely formatted csv file (seitzman_2018_rois.csv) with
+coordinates, network and region names
+"""
+import pandas as pd
```
GaelVaroquaux: I'm not sure that I like having this code in the nilearn tree, as we cannot test it under CI (because it requires pandas).

GaelVaroquaux: Actually, my point is rather that this file, as well as the two txt files, should be moved out of the nilearn package, and should not be shipped with nilearn. Rather, they should be in the repository for developers to use them.

Maybe we could add a directory "dev_utils" at the base of the source tree, and move them in a subdirectory there, with a clear readme.

fliem: I tried anyway :) It wasn't so tricky, once I found the `recfunctions` module.

The two original input text files now reside in the `data` directory and are read directly by the fetcher. The install issue should now be fixed as well.

gkiar:
@kchawla-pi -- minimum to replicate:

```
from nilearn.datasets import fetch_coords_seitzman_2018
data = fetch_coords_seitzman_2018()
data.regions
```

Currently fails on 3.4 (unix) and 3.5 (Windows)

fliem:
this PR is on hold until support for python 2 and numpy 1 is dropped (making the encoding easier to handle)

kchawla-pi:
You won't have to wait too long.

nilearn/datasets/atlas.py
```diff
@@ -808,6 +809,79 @@ def fetch_coords_dosenbach_2010(ordered_regions=True):
...
+    # get integer regional labels and convert to text labels with mapping
+    # from header line
+    with open(anatomical_file, 'r') as fi:
+        header = fi.readline()
+    region_mapping = {}
+    for r in header.strip().split(","):
+        i, region = r.split("=")
+        region_mapping[int(i)] = region
+
+    anatomical = np.genfromtxt(anatomical_file, skip_header=1,
```
KamalakerDadi: Can we use ```from nilearn._utils.numpy_conversions import csv_to_array``` here ?

nilearn/datasets/atlas.py
```diff
@@ -808,6 +809,79 @@ def fetch_coords_dosenbach_2010(ordered_regions=True):
...
+    """
+    dataset_name = 'seitzman_2018'
+    fdescr = _get_dataset_descr(dataset_name)
+    package_directory = os.path.dirname(os.path.abspath(__file__))
+    roi_file = os.path.join(package_directory, "data",
+                            "seitzman_2018_ROIs_300inVol_MNI_allInfo.txt")
+    anatomical_file = os.path.join(package_directory, "data",
+                                   "seitzman_2018_ROIs_anatomicalLabels.txt")
+
+    rois = np.recfromcsv(roi_file, delimiter=" ", encoding="ascii")
```
KamalakerDadi: Here as well can we use import csv_to_array from _utils.numpy_conversions ?

nilearn/datasets/atlas.py
fliem: `csv_to_array` removes the header, which is not ideal for here

kchawla-pi:
Hey @fliem I m going to cancel all CircleCI builds for this PR. Any reason not to?

fliem:
On hold until support for python<3.5 is dropped

kchawla-pi:
Consider Python<3.5 dropped

examples/03_connectivity/plot_sphere_based_connectome.py
```diff
@@ -15,6 +16,11 @@
...
 using fMRI.", 2010, Science 329, 1358-1361.
 
+[3] `Seitzman, B. A., et al. "A set of functionally-defined brain regions with
+improved representation of the subcortex and cerebellum.", 2018, bioRxiv,
+450452
+<http://doi.org/10.1101/450452>`_
+
 """
 
 ###############################################################################
```
fliem: Hi @KamalakerDadi @kchawla-pi 
Do you have an idea why the section heading 'Load fMRI data and Power atlas' is not rendering correctly? The html looks like this:
```
[...]
bioRxiv, 450452 Load fMRI data and Power atlas ——————————
We [...]
```

Thanks

KamalakerDadi: Giving me a snapshot would help ?

Could you give me a snapshot from your rendered documentation ?

KamalakerDadi: I build the doc on my laptop. It looks ok to me.

pr

fliem: Thanks. Might be an issue with my installation then.
Bild 18 04 19 um 11 42

fliem: The other headings render as expected, btw. It's just this one making problems.

fliem:
The encoding issues are fixed now and the code works for all numpy versions.

One remaining issue: 
Running the `plot_sphere_based_connectome.py` with fwhm=4 gives the following error for the Seitzman atlas (but not for Dosenbach or Power):
```
  File "/anaconda3/envs/nilearn/lib/python3.7/site-packages/sphinx_gallery/gen_rst.py", line 394, in _memory_usage
    out = func()
  File "/anaconda3/envs/nilearn/lib/python3.7/site-packages/sphinx_gallery/gen_rst.py", line 382, in __call__
    exec(self.code, self.globals)
  File "/Users/franzliem/Dropbox/Workspace/nilearn/examples/03_connectivity/plot_sphere_based_connectome.py", line 190, in <module>
    covariance_estimator.fit(timeseries)
  File "/anaconda3/envs/nilearn/lib/python3.7/site-packages/sklearn/covariance/graph_lasso_.py", line 725, in fit
    verbose=inner_verbose, return_n_iter=True)
  File "/anaconda3/envs/nilearn/lib/python3.7/site-packages/sklearn/covariance/graph_lasso_.py", line 269, in graphical_lasso
    raise e
  File "/anaconda3/envs/nilearn/lib/python3.7/site-packages/sklearn/covariance/graph_lasso_.py", line 247, in graphical_lasso
    raise FloatingPointError('The system is too ill-conditioned '
FloatingPointError: The system is too ill-conditioned for this solver. The system is too ill-conditioned for this solver
```


However, this error wasn't raised with fwhm=6.

@GaelVaroquaux @KamalakerDadi any ideas what the problem might be? 
1. @KamalakerDadi Was the change from fwhm=4 to 6 that has been reverted in  #1990 maybe due to similar issues?  
1. Might this be related to the development fmri data having 4mm voxels? 
1. There also are some overlapping ROIs in this atlas (I haven't looked at the extent of this). This might be an issue because there is then more dependency between the signals of overlapping ROIs. However, wouldn't higher smoothing make this problem worse, since the neighboring signals then get even more mixed?

One possibility would be to just set fwhm to 6 for all three atlases in this example. Would be nice to get an idea why this happens, though.

GaelVaroquaux:
Smoothing and nearly overlapping spheres are probably extracting almost the same signal at different spots. 

If you don't smooth at all, what happens? Is there a crash ?  How do the results look like qualitatively ? 

fliem:
> If you don't smooth at all, what happens? Is there a crash ?  How do the results look like qualitatively ?

**fwhm=None, radius=4** and **smoothing_fwhm=None, radius=None** give the same ill-conditioned... error

**fwhm=6, radius=4** looks like this:
fwhm_6


vs the **Power atlas, smoothing_fwhm=4, radius=5**
power

KamalakerDadi:
>    @KamalakerDadi Was the change from fwhm=4 to 6 that has been reverted in #1990 maybe due to similar issues?

I don't remember exactly. If you use 4mm you don't have numerical issues with GraphLasso ?

KamalakerDadi:
>    @KamalakerDadi Was the change from fwhm=4 to 6 that has been reverted in #1990 maybe due to similar issues?

>I don't remember exactly. If you use 4mm you don't have numerical issues with GraphLasso ?

Ok. Using 6mm has no numerical issues. May be I would have done because of that. I can track down the history for confirmation. But I dont' see the reason for not using 6mm. I would use 6mm and add a note about converging problems if example is not cluttered too much.

GaelVaroquaux:
I resolved the merge conflict. Let's wait for CI to run.

GaelVaroquaux:
This is good to go once CI has ran.

examples/03_connectivity/plot_sphere_based_connectome.py
```diff
@@ -282,6 +288,54 @@
...
+                         node_color='darkred', node_size=20)
+
+###############################################################################
+# Now let's calculate connectivity for the Seitzman atlas.
+spheres_masker = input_data.NiftiSpheresMasker(
+    seeds=coords, smoothing_fwhm=6, radius=4,
+    detrend=True, standardize=True, low_pass=0.1, high_pass=0.01, t_r=2,
+    allow_overlap=True)
+
+timeseries = spheres_masker.fit_transform(fmri_filename,
```
GaelVaroquaux:
```suggestion
timeseries = spheres_masker.fit_transform(func_filename,
```

fliem:
Allright. 
I now for all atlases in the example set fwhm to 6.
If the tests pass this should be ready to merge. Thanks

GaelVaroquaux:
Indeed, this PR needs a merge on master, to fix the CI.

fliem:
Circleci gave me `Build timed out after 5h0m0s`
@kchawla-pi is that a known issue?

GaelVaroquaux:
We've seen this before. It depends on the hardware that runs the job (circle uses inhomogeneous hardware). We are working on a fix, but it basically implies changing core algorithms, so it's taking a while.

I will restart your job, and hope that it runs.

kchawla-pi:
That's not right. The build timeout is 7hrs!

fliem:
Thanks @kchawla-pi

kchawla-pi:
@fliem This should have been merged 2 weeks ago. Apologies.

PR #1969, opened 2019-04-12
[WIP] Fix CircleCI cache for built docs so it is inheritable from master to other branches

kchawla-pi:
Follow up of PR #1967 

 - Multicore CPU use option is counter productive. Reverted.
- Inheritable docs cache from master needs to be fixed.
- Implement JUnitXML extension for chase-able tracebacks.


kchawla-pi:
Only way to test this is to merge on master and test. If there's a problem, I'll revert the PR.

PR #1973, opened 2019-04-13
[MRG+2] FIX : new_img_like broken with mgz files

agramfort:
the last release broken the MNE CIs due to plotting from .mgz files. This fixes it.

to replicate the pb:

```
from nilearn.image import new_img_like
import nibabel

niimg = nibabel.load('/Applications/freesurfer/subjects/fsaverage/mri/T1.mgz')
new_img_like(niimg, niimg.get_data().astype(float), niimg.affine, copy_header=True)
```

let me know if I can turn this into a non-regression test

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -636,14 +636,18 @@ def new_img_like(ref_niimg, data, affine=None, copy_header=False):
     header = None
     if copy_header:
         header = copy.deepcopy(ref_niimg.header)
-        header['scl_slope'] = 0.
-        header['scl_inter'] = 0.
+        if 'scl_slope' in header:
+            header['scl_slope'] = 0.
+        if 'scl_inter' in header:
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
It would certainly be good to have a test. However, if the only way to reproduce is to load a file shipped with freesurfer, it will be hard to QA.

agramfort:
Adding à tiny mgz file to repo is not option?

GaelVaroquaux:
> Adding à tiny mgz file to repo is not option?

It would be. But it's even better if you write it during the tests.

agramfort:
ok it's was simple to add a non-regression test

GaelVaroquaux:
Excellent! Thanks, that's useful.

+1 for merge once the tests have run.

GaelVaroquaux:
Well, actually, they have run!

Merging!

PR #1977, opened 2019-04-14
Reduce the computational time for SpaceNet

KamalakerDadi:
Attempts to reduce the computational time in space net examples.

What I am trying to run examples offline. Collect regularization terms and use them manually in SpaceNet object.

### DIFF
--- a/examples/02_decoding/plot_haxby_space_net.py
+++ b/examples/02_decoding/plot_haxby_space_net.py
@@ -42,17 +42,31 @@ from nilearn.image import mean_img
 background_img = mean_img(func_filenames)
 
 ##############################################################################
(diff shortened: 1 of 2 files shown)


KamalakerDadi:
Any thoughts on this ? @GaelVaroquaux @bthirion

GaelVaroquaux:
Nice! The circleCI build succeeded, which hasn't happened in quite a long time.

That said, I don't think that this is the right fix: the parameters added do not make any sense to me. However, I think that you may have a good idea here: maybe it's the convergence parameters of SpaceNet that are killing us. Let me look at this closely.

KamalakerDadi:
Good that atleast it lit some ideas. Closing after your PR made success.

PR #1980, opened 2019-04-14
MAINT: control SpaceNet convergence

GaelVaroquaux:
Avoid too long loops when spacenet does not converge

Inspired by #1977, by @KamalakerDadi 

This PR should be "safe", in the sense that the change should be harmless. Iterating 1000 time is too much anyhow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,7 @@ jobs:
           command: |
             source activate testenv
             pip install -e .
-            set -o pipefail && cd doc && make html-strict 2>&1 | tee log.txt
+            set -o pipefail && cd doc && make html-strict SPHINXOPTS="-v" 2>&1 | tee log.txt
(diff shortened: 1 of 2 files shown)


kchawla-pi:
Azure Pipelines fail has nothing to do with your PR.

GaelVaroquaux:
SpaceNet is still very slow, but at least things converge.

Maybe with a warm cache thing won't be a problem?

I suggest we merge, and keep thinking about this problem.

kchawla-pi:
Running make HTML strict locally should be okay.

kchawla-pi:
Maybe we can ssh and pdb into the example, see what's what, somehow...

kchawla-pi:
> Running make HTML strict locally should be okay.

I mean for this Sprint

GaelVaroquaux:
Merging.

PR #1981, opened 2019-04-14
Fix Azure Pipelines config

kchawla-pi:


azure-pipelines.yml
```diff
@@ -29,10 +29,9 @@ jobs:
     displayName: 'Install dependencies'
 
   - script: |
-      pip install pytest
       pip install .
-      pytest ./nistats/tests --doctest-modules --junitxml=junit/test-results.xml
-    displayName: 'pytest'
+      nosetests ./nilearn -v
```
GaelVaroquaux: Hum, nosetests is deprecated and we are now runing pytest :(.

kchawla-pi:
I tried both. The tests are failing here but passing on appveyor.
Azp is running fine on Nistats.
Will figure it out eventually. For now we will ignore it's results and use
appveyor

PR #1953, opened 2019-03-30
Update ADHD dataset in examples to MAIN

KamalakerDadi:
Follow up to PR #1887 

PR #1887 will be closed for mess up in the commit history.

PR #1990, opened 2019-04-15
[FIX] revert fwhm for dosenbach example back to 4mm

fliem:
In #1953 the smoothing in the `plot_sphere_based_connectome.py` example was changed from 6mm to 4mm, but only for the fist (dosenbach) atlas, not the second (power). Since @emdupre and @KamalakerDadi can't figure out why that happened, we make this consistent by reverting to 4mm for both atlases.

### DIFF
--- a/examples/03_connectivity/plot_sphere_based_connectome.py
+++ b/examples/03_connectivity/plot_sphere_based_connectome.py
@@ -57,7 +57,7 @@ print('Stacked power coordinates in array of shape {0}.'.format(coords.shape))
 from nilearn import input_data
 
 spheres_masker = input_data.NiftiSpheresMasker(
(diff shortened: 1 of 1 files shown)


KamalakerDadi:
Same here. Failures on the travis and appveyor should be fixed after merging this PR #1992 . Please don't worry about the failures.

KamalakerDadi:
Sorry for the noise. You may have to rebase. As we have made some updates to the OSF brain development storage which is causing these failures.

If I am not wrong. Those are fixed now.

Thanks!

fliem:
Done. Thanks.

GaelVaroquaux:
LGTM. Merging. Thanks!

Issue #1697, opened 2018-07-23
plot_surf_roi


jeromedockes:
The interface of plot_surf_roi is buggy, confusing and inconsistent with
plot_roi.

it takes a roi_map, which can be:

- a list -> interpreted as a list of list of node indices, each for an roi.
- an array of the same shape as the mesh -> interpreted as a mask
- an array of a different shape -> interpreted as a list of indices for the roi nodes.
- a string -> should be interpreted as the path to a surface map (according to docstring), but results in an exception.

I suggest we only accept valid surface maps (i.e. arrays of shape (# of nodes,),
or paths to surface map files that contain a map of the same shape), and treat them
as atlases with two regions (0=background, 1=roi). If a user has an ROI in the
form of a list of node indices, they can easily produce a mask themselves with
mask = np.zeros(len(mesh), dtype=int); mask[roi_idx] = 1. plot_roi only accepts full volume maps
so this would be consistent.

at the very least, we must remove the different behaviour for lists and arrays, and
treat paths to surface images correctly.

@KamalakerDadi  @GaelVaroquaux

KamalakerDadi:
@jeromedockes do you think this can be closed in favour of #1991

jeromedockes:
> @jeromedockes do you think this can be closed in favour of #1991

I think so

PR #1991, opened 2019-04-15
Fix/plot surf roi

juhuntenburg:
Fixing #1697 

The input roi_map can now be a) an array of the same length as the mesh, or b) a file that can be loaded with load_surf_data and returns an array of the same length as the mesh.


nilearn/plotting/surf_plotting.py
```diff
@@ -503,35 +503,25 @@ def plot_surf_roi(surf_mesh, roi_map, bg_map=None,
...
-                         'vertices included in that ROI')
-    vmin, vmax = np.min(roi_map), 1 + np.max(roi_map)
-    display = plot_surf(surf_mesh, surf_map=roi_map, bg_map=bg_map,
+    roi = load_surf_data(roi_map)
+    vmin, vmax = np.min(roi), 1 + np.max(roi)
+
+    mesh = load_surf_mesh(surf_mesh)
+
+    if len(roi.shape) is not 1:
+        raise ValueError('roi_map can only have one dimension but has'
```
GaelVaroquaux: I think that you are missing a space at the end of the string

GaelVaroquaux:
Maybe we should update the docstring on the input argument to be more clear about the input format.

nilearn/plotting/surf_plotting.py
juhuntenburg: Done

juhuntenburg:
> Maybe we should update the docstring on the input argument to be more clear about the input format.

As in, adding that it has to have the same length as the number of vertices on the supplied mesh?

doc/connectivity/connectome_extraction.rst
```diff
@@ -105,8 +104,8 @@ of the estimator::
 .. topic:: **Exercise: computing sparse inverse covariance**
    :class: green
 
-   Compute and visualize a connectome on the first subject of the brain
-   development dataset downloaded with :func:`nilearn.datasets.fetch_development_fmri`
+   Compute and visualize a connectome on the first subject of the ADHD
+   dataset downloaded with :func:`nilearn.datasets.fetch_adhd`
```
GaelVaroquaux: Hum, you have a broken merge :(.

juhuntenburg: Ok, will fix it. Just out of interest, how come it doesn't show up as a merge conflict?

juhuntenburg:
> If I look at the current docstring for roi_map, and I am not sure that it is still appropriate.

Can you specify? It says that the input can be an array or a file, which is how it works now. Happy to describe in more detail, just want to make sure I understand what exactly is your concern

GaelVaroquaux:
> Can you specify? It says that the input can be an array or a file, which is how it works now.

Point taken. I read the diff wrong.

Here is what I would write: "a Numpy array. The value at each vertex is
an integer giving the label number."

How about that?

KamalakerDadi:
Failures on the travis and appveyor should be fixed after merging this PR #1992 . Please don't worry about the failures.

KamalakerDadi:
Failures on the travis and appveyor should be fixed after merging this PR #1992 . Please don't worry on the failures.

GaelVaroquaux:
@juhuntenburg : can you merge master into this branch, to fix the failure.

GaelVaroquaux:
OK, merging. Thank you!

PR #1994, opened 2019-04-15
[WIP] PoSCE: population shrinkage of covariance estimator

mrahim:
Population shrinkage of covariance estimator (PoSCE) for connectivity estimation.
PoSCE integrates prior knowledge of covariance distribution leading to a non-isotropic shrinkage.

image

image

TODO :
- Examples
- Tests

### DIFF
--- a/examples/03_connectivity/plot_compare_shrunk_connectome.py
+++ b/examples/03_connectivity/plot_compare_shrunk_connectome.py
@@ -1,0 +1,72 @@
+"""
+Population shrinkage of covariance estimator (PoSCE) for connectome analysis
+============================================================================
+
+This examples applies population-based shrinkage of covariance to build a
+functional connectome.
+
+We use the `MSDL atlas
+<https://team.inria.fr/parietal/18-2/spatial_patterns/spatial-patterns-in-resting-state/>`_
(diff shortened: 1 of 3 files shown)


GaelVaroquaux:
AFAICT, PoSCE now exposed a fit_transform which returns the tangent-space parametrization. The drawback of this is that it confuses users between the covariance estimator and the covariance parametrization.

Here is what I suggest:

* PoSCE becomes an object that only an estimator, and has an API similar to GroupSparseCovariance
* ConnectivityMeasure can take cov_estimator="posce" and kind="tangent", in which case it special cases for PoSCE (note the string above, and not the estimator, to make things simpler).

thomasbazeille:
Hello ! What is the status on this PR ? The discussion is not really explicit here.

bthirion:
It still needs work, but it's worth the effort I think.

mrahim:
I will rebase to see what should to be fixed.

mrahim:
@GaelVaroquaux  : if it is still ok, I will integrate this suggestion

> ConnectivityMeasure can take cov_estimator="posce" and kind="tangent", in which case it special cases for PoSCE (note the string above, and not the estimator, to make things simpler).

PR #1995, opened 2019-04-16
Update brain development fetcher and fix documentation

KamalakerDadi:
Fix underline too short documentation error.


KamalakerDadi:
I would like check CircleCI will be fixed. Merging this one as this is minor detail.

PR #1999, opened 2019-04-16
[ENH] Selected confounds

fliem:
As it is now, `fetch_development_fmri` returns all 28 fmriprep confounds. To make examples more sensible without increasing verbosity, this PR changes the default behavior in that it makes the fetcher return files containing 15 selected confounds. Getting all fmriprep confounds is still possible via setting `reduce_confounds` = False.

 @emdupre @KamalakerDadi


fliem:
On hold until support for python<3.5 is dropped

fliem:
Sorry for the back and forth. I removed the encoding flag. And tested it with 2.7 and 3.7 on mac. Hope that works on windows as well.

nilearn/datasets/func.py
```diff
@@ -2009,6 +2009,14 @@ def fetch_development_fmri(n_subjects=None, data_dir=None, resume=True,
         The number of subjects to load. If None, all the subjects are
         loaded. Total 155 subjects.
 
+    reduce_confounds: bool, optional (default True)
```
KamalakerDadi: Not so strong in my comment but just curious for keeping default to True ?

nilearn/datasets/tests/test_func.py
```diff
@@ -658,3 +658,13 @@ def test_fetch_development_fmri():
...
     assert_equal(data.phenotypic.shape, (2,))
     assert_not_equal(data.description, '')
+
+    # check reduced confounds
+    confounds = np.recfromcsv(data.confounds[0], delimiter='\t')
+    assert_equal(len(confounds[0]), 15)
+
+    # check full confounds
+    data = func.fetch_development_fmri(n_subjects=2, reduce_confounds=False,
+                                  verbose=1)
```
KamalakerDadi: https://travis-ci.org/nilearn/nilearn/jobs/520854805#L437

KamalakerDadi:
Great stuff. LGTM from my side.

kchawla-pi:
Consider Python<3.5 dropped.

nilearn/datasets/func.py
```diff
@@ -2056,6 +2064,11 @@ def fetch_development_fmri(n_subjects=None, data_dir=None, resume=True,
     dataset_name = 'development_fmri'
     data_dir = _get_dataset_dir(dataset_name, data_dir=data_dir,
                                 verbose=1)
+    keep_confounds = ['trans_x', 'trans_y', 'trans_z', 'rot_x', 'rot_y',
```
emdupre: Should we add to the examples that these are the included confounds ?

nilearn/datasets/func.py
```diff
@@ -2098,5 +2111,18 @@ def fetch_development_fmri(n_subjects=None, data_dir=None, resume=True,
                                                            url=None,
                                                            verbose=verbose)
 
+    if reduce_confounds:
+        reduced_regressors = []
+        for in_file in regressors:
+            out_file = os.path.splitext(in_file)[0] + '_reduced.tsv'
```
emdupre: It would be more BIDS-compliant to change the `desc` value to `reducedConfounds`. No strong opinion, but one idea.

nilearn/datasets/func.py
fliem: I think this is there more reasonable option that does not require further interaction with the confounds

nilearn/datasets/tests/test_func.py
fliem: fixed

nilearn/datasets/func.py
fliem: renamed

fliem:
Thanks @emdupre and @KamalakerDadi 
I have changed the filename and fixed the indent. This should be final now.

nilearn/datasets/func.py
fliem: might be too much?

emdupre: I'm not tied to it -- thought of it just since this is a little unusual compared to the other datasets. But that might be a more general point about using confounds than we want to make !

GaelVaroquaux:
Looks great!

PR #2001, opened 2019-04-16
organizing functions for testing surface code

juhuntenburg:
Moving helper functions in separate file

### DIFF
--- a/nilearn/surface/testing.py
+++ b/nilearn/surface/testing.py
@@ -3,11 +3,13 @@
 import numpy as np
 from scipy.spatial import Delaunay
 
+
 def _generate_surf():
     rng = np.random.RandomState(42)
     coords = rng.rand(20, 3)
     faces = rng.randint(coords.shape[0], size=(30, 3))
     return [coords, faces]
+
 
 def _flat_mesh(x_s, y_s, z=0):
     x, y = np.mgrid[:x_s, :y_s]
--- a/nilearn/surface/testing.py
+++ b/nilearn/surface/testing.py
@@ -1,0 +1,24 @@
+# Helper functions for testing surface code
+
+import numpy as np
+from scipy.spatial import Delaunay
+
+def _generate_surf():
+    rng = np.random.RandomState(42)
+    coords = rng.rand(20, 3)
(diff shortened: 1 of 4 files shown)


KamalakerDadi:
Just flake8 issues. https://travis-ci.org/nilearn/nilearn/jobs/520799043#L437

Otherwise, LGTM.

KamalakerDadi:
Could you tell us the motivation to move to separate file ?

jeromedockes:
> Could you tell us the motivation to move to separate file ?

as far as I can tell these functions are only used in test_surface? why create a testing module ouside of the tests package?

juhuntenburg:
Actually, I was asked by @SylvainLan and @GaelVaroquaux to do it this way. I don't have particular strong feelings about it.

nilearn/surface/testing.py
```diff
@@ -0,0 +1,26 @@
+# Helper functions for testing surface code
+
+import numpy as np
+from scipy.spatial import Delaunay
+
+
+def _generate_surf():
```
GaelVaroquaux: These functions should probably not be private (though the module might be). They should also have a tiny docstring.

GaelVaroquaux:
The reason to move it is to foster reuse in the surface code (which is going to grow soon).

Sorry, @juhuntenburg, I'm asking you to add docstring, in other words, you are victim of a classic: you are asked to improve code just because you are moving it.

jeromedockes:
and why move them outside of the tests package? can any of these functions be used for something else than testing? if so, the module should have a different name

GaelVaroquaux:
> and why move them outside of the tests package?

Because only tests go in there, no utilities to test.

> can any of these functions be used for something else than testing? if so, the module should have a different name

The file is called "testing.py", which is a common name for testing
utilities.

jeromedockes:
> Because only tests go in there, no utilities to test.

ok thanks!

kchawla-pi:
> The file is called "testing.py", which is a common name for testing

Should we rename them to `testing_utils.py` for future contributors?

GaelVaroquaux:
Fine with me 

GaelVaroquaux:
Great! Merging!

Issue #1919, opened 2019-02-14
Switch to PEP8 style absolute imports


kchawla-pi:
We should change the relative imports used in NIlearn to PEP8 endorsed absolute imports.
 Relative imports misbehave/break more and are less clear to read.
Relative imports tend to break more for reasons that are not readily apparent, especially within IDEs.

I know relative absolute imports are acceptable alternatives according to PEP8, but the ones in Nilearn and Nistats often break for me when I want to run specialized tools and test runners from within the IDE instead of the terminal.

jeromedockes:
from https://www.python.org/dev/peps/pep-0008/#id23 : 

  Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):
```
  import mypkg.sibling
  from mypkg import sibling
  from mypkg.sibling import example
```

GaelVaroquaux:
Over my 15 years of doing Python, PEP8 has changed its recommendations
quite a few times (I do believe that relative imports where recommended a
few years ago).

I've always considered massively changes to code bases due to changes in
PEP8 recommendations bad practice. They are not only a loss of time, but
also tend to create conflicts in open pull requests.

And if a recommendation is so arbitrary that it changes over time, I
think that it can be seen as a rounding error.

My opinion is not that absolute imports are better, because relative
imports make refactoring easier.

However, if the consensus is that absolute imports are better, I am fine
changing nilearn's policy and encouraging new code to have relative
imports. I do frown on a massive change to the code base.

jeromedockes:
> I've always considered massively changes to code bases due to changes in PEP8 recommendations bad practice. They are not only a loss of time, but also tend to create conflicts in open pull requests.

+1 !

PR #2015, opened 2019-04-17
[MRG] Moving nilearn.decoding.objective_functions._unmask to nilearn.masking._unmask_from_to_3d_array

thomasbazeille:
Renaming stand-alone _unmask function for SpaceNet to _unmask_from_to_3d_array and moving it in nilearn/masking.py with the relevant tests and import changes.

The goal is to use this function later to refactor ReNA for PR#1983

### DIFF
--- a/nilearn/decoding/space_net.py
+++ b/nilearn/decoding/space_net.py
@@ -34,7 +34,7 @@ from sklearn.model_selection import check_cv
 from sklearn.linear_model.base import _preprocess_data as center_data
 from .._utils.compat import _basestring
 from .._utils.cache_mixin import CacheMixin
-from .objective_functions import _unmask
+from nilearn.masking import _unmask_from_to_3d_array
 from .space_net_solvers import (tvl1_solver, _graph_net_logistic,
                                 _graph_net_squared_loss)
 
@@ -103,9 +103,9 @@ def _univariate_feature_screening(
         sX = np.empty(X.shape)
         for sample in range(sX.shape[0]):
             sX[sample] = ndimage.gaussian_filter(
-                _unmask(X[sample].copy(),  # avoid modifying X
-                        mask), (smoothing_fwhm, smoothing_fwhm,
-                                smoothing_fwhm))[mask]
+                _unmask_from_to_3d_array(X[sample].copy(),  # avoid modifying X
+                                         mask), (smoothing_fwhm, smoothing_fwhm,
+                                                 smoothing_fwhm))[mask]
     else:
         sX = X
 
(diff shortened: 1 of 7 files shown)


thomasbazeille:
@KamalakerDadi can you please review this small PR when you've time ? It's a small change necessary to refactor ReNA. Thanks!

nilearn/decoding/space_net.py
```diff
@@ -34,7 +34,7 @@
 from sklearn.linear_model.base import _preprocess_data as center_data
 from .._utils.compat import _basestring
 from .._utils.cache_mixin import CacheMixin
-from .objective_functions import _unmask
+from nilearn.masking import _unmask_from_to_3d_array
```
KamalakerDadi: Could you use relative imports please ? Like the ones above and below.

nilearn/decoding/space_net_solvers.py
```diff
@@ -13,10 +13,11 @@
 
 from math import sqrt
 import numpy as np
+from nilearn.masking import _unmask_from_to_3d_array
```
KamalakerDadi: same here about relative imports.

KamalakerDadi:
If you can address the flake8 issues as much as you can, that would be great. I agree that there are written like that.

Otherwise, LGTM

Great thinking making them into separate PR.

nilearn/decoding/space_net.py
kchawla-pi: @GaelVaroquaux  and I talked about this in #1919 and  while he doesn't want existing imports to be changes, for new imports absolute imports are okay.
PEP8 standard is to use Absolute imports. The only time relative imports are considered okay if  a specific absolute import is too verbose.

KamalakerDadi: +1 for relative imports again.

Either we do this or change everything to absolute imports.

GaelVaroquaux:
> +1 for relative imports again.

> Either we do this or change everything to absolute imports.

I think that we should move to absolute imports, just to be consistent
with the style guide.

However, I am not in favor of doing a major PR to change everything at
once, because I am afraid of the huge number of conflicts that it will
create.

thomasbazeille:
Ok so if there we can keep the new absolute imports as is, can the rest seems ok, can someone merge the PR ?

Thanks a lot for the reviews

nilearn/decoding/space_net.py
GaelVaroquaux: Given the recommendations of PEP8, we should gradually evolve our code to absolute imports.

KamalakerDadi: Are the recommendations changed since last year ? This is what I heard when Loic was there.

GaelVaroquaux:
LGTM. +1 for merge.

PS: next time: turn off your autopep8 :D

Issue #1551, opened 2017-11-14
Avoid resampling when it's just different FOV, and not different affine


GaelVaroquaux:
For two images that are looking at the same data, but with a different FOV, nilearn.image.resample_img currently uses the affine resampling straegy. However, the transformation could easily be done via a simple padding of arrays.

We should probably be able to detect this with the "A" matrix on line 
https://github.com/nilearn/nilearn/blob/e442a34a9eda52b2c2697c5b176ec3d95a99e8d3/nilearn/image/resampling.py#L488
which should pretty much be the identity. Once we've detected this, the challenge will be write code that computes the translation and padding.

The benefit might be a large speedup in some settings, as often the mask and the data are on the same grid (typically the grid of the MNI template at a given resolution), but with a different FOV that was inherited from where they originated.

bthirion:
+1
Actually, when there is only a translation/padding the A matrix should be the identity.

GaelVaroquaux:
Yes, this is what I meant. Sorry, I've corrected.

cancan101:
Is this the same "issue" as calling `nilearn.image.crop_img` and then taking the resulting image affine and passing it to `nilearn.image.resample_img`? e.g I call crop on the image data and then want to apply that same crop to the mask but want to just use array slice-ing rather than a more expensive resample operation?

bthirion:
Indeed.  But the resampling code should detect and handle such situations.

YSanchezAraujo:
anyone know if a PR for this has already been put forth?

GaelVaroquaux:
> anyone know if a PR for this has already been put forth?

I don't think so.

KamalakerDadi:
>anyone know if a PR for this has already been put forth?

Nobody AFAIK. May be @cancan101 ?

gkiar:
Hey everyone! At the Nilearn sprint now and think @GaelVaroquaux and I have a pretty simple plan for taking care of this, in pseudocode below. Let me know if I'm missing a corner case, if I have some flaw in logic, or if you have a better idea? Thanks!

```
if A != I OR (A == I AND not all(b == integers):
  ... normal behaviour (done)
elif A == I AND |b| == 0:
  ... normal behaviour, which happens do be do nothing (done)
elif (A == I OR some combination of permutation(I) and sign-flipped(I)) AND all(b == integers):
  ... special case: can be solved with padding alone
  crop source image and keep track of N voxels offset before/after volume
  flip along sign-flipped dimensions
[...]

PR #2025, opened 2019-04-17
Avoiding resampling for identity transforms with integer translations

gkiar:
- Addresses #1551 
- Handles special case in image resampling when integer translations are accompanied by an identity affine transform in which the resampling can be fully performed by copping and padding of the source image into the target container.
- Moved `pad_img` into `nilearn.image` as it can be a useful utility outside of just testing
- Added test for `crop_img` and `pad_img`.

cc: @GaelVaroquaux


GaelVaroquaux:
You have a bunch of test failures on travis.

gkiar:
Yep - on it :) I wasn't able to run all the tests locally because surface things were being weird and seg-faulting on what I believe is untouched code... (??), but running and resolving the failures now. Will ping you when I believe it's ready

gkiar:
@GaelVaroquaux Ready for your review I believe - while Appveyor and Circle are pending, the tests passed on Travis.

kchawla-pi:
The error on Appveyor is non-trivial.

GaelVaroquaux:
We're getting failures on appveyor that are related to this PR:
https://ci.appveyor.com/project/nilearn-ci/nilearn/builds/23942220

Maybe it's the tolerance of the test that is set too low.

doc/whats_new.rst
```diff
@@ -4,6 +4,11 @@
 NEW
 ---
 
+- Optimization to image resampling
+  :func:`nilearn.image.resample_img` has been optimized to pad rather than
+  resample images in the special case when there is only a translation
+  between two spaces. This is a common case in :class:`nilearn.input_data.NiftiMasker`
+  when using the `mask_strategy="template` option for brains in MNI space.
```
GaelVaroquaux:
```suggestion
  when using the `mask_strategy="template"` option for brains in MNI space.
```

nilearn/image/image.py
```diff
@@ -364,16 +376,53 @@ def crop_img(img, rtol=1e-8, copy=True):
...
+        end = np.minimum(end + 1, data.shape[:3])
+
+    slices = [slice(s, e) for s, e in zip(start, end)][:3]
+    cropped_im = _crop_img_to(img, slices, copy=copy)
+    return cropped_im if not return_offset else (cropped_im, tuple(slices))
 
-    slices = [slice(s, e) for s, e in zip(start, end)]
 
-    return _crop_img_to(img, slices, copy=copy)
+def pad_img(array, *args):
```
GaelVaroquaux: I would rather avoid the *args pattern and use "pad_sizes" as an explicit required argument:
```suggestion
def pad_img(array, pad_sizes):
```

GaelVaroquaux: Also, if the input is an array, the function should be called "pad_array". But in this case, it should also probably be made private: "_pad_array"

gkiar: Agreed - I just inherited this from the previously written utility I copied over. :)

gkiar:
Regarding the issue: after just poking around, I notice that the test which fails is not touching the new code in this PR, but the old/standard resampling function (since the affine transform is not `np.eye(3)`). I verified this with debug statements locally...

I don't have a windows machine to test this on, but am happy to debug if you have suggestions for how to approach it. Thanks for the thorough review!

gkiar:
Update: using the same version of `scipy` that is used in the appveyor build, `0.19.1`, I was able to reproduce the bug locally.

nilearn/image/__init__.py
```diff
@@ -15,5 +15,5 @@
            'smooth_img', 'crop_img', 'mean_img', 'reorder_img',
            'swap_img_hemispheres', 'concat_imgs', 'copy_img',
            'index_img', 'iter_img', 'new_img_like', 'threshold_img',
-           'math_img', 'load_img', 'clean_img',
+           'math_img', 'load_img', 'clean_img', 'pad_array',
```
GaelVaroquaux: Do we want to make pad_array public? I am not sure. Functions in this module are to manipulate images, and not arrays.

kchawla-pi:
Code coverage has dropped. Are we missing some test or is this trivial?

nilearn/image/tests/test_resampling.py
```diff
@@ -116,6 +86,7 @@ def test_downsample():
     affine = np.eye(4)
     rot_img = resample_img(Nifti1Image(data, affine),
                            target_affine=2 * affine, interpolation='nearest')
+    # import pdb; pdb.set_trace()
```
GaelVaroquaux: This line should probably be removed.

nilearn/image/tests/test_resampling.py
```diff
@@ -132,6 +103,7 @@ def test_downsample():
                                target_affine=2 * affine,
                                interpolation='nearest',
                                copy=copy)
+        print(downsampled - rot_img.get_data()[:x, :y, :z, ...])
```
GaelVaroquaux: This line should probably be removed.

nilearn/image/tests/test_resampling.py
```diff
@@ -558,6 +530,44 @@ def test_resample_to_img():
     np.testing.assert_almost_equal(downsampled,
                                    result_img.get_data()[:x, :y, :z, ...])
 
+def test_crop():
```
GaelVaroquaux: Maybe a small comment explaining the logic of this test. It helps future developers.

nilearn/image/tests/test_resampling.py
```diff
@@ -558,6 +530,44 @@ def test_resample_to_img():
...
+    source_affine = np.eye(4)
+    source_affine[:, 3] = np.append(np.random.randint(0, 4, 3), 1)
+    source_data = rand_gen.random_sample(source_shape)
+    source_img = Nifti1Image(source_data, source_affine)
+
+    target_shape = (11, 10, 9)
+    target_data = np.zeros(target_shape)
+    target_affine = source_affine
+    target_affine[:3, 3] -= 3  # make the affine 3 larger in x, y, z
+    target_data[3:9, 3:7, 3:9] = source_data  # put the data at the offset loca
```
GaelVaroquaux:
```suggestion
    target_data[3:9, 3:7, 3:9] = source_data  # put the data at the offset location
```

nilearn/image/tests/test_resampling.py
```diff
@@ -558,6 +530,44 @@ def test_resample_to_img():
...
+    source_shape = (6, 4, 6)
+    source_affine = np.eye(4)
+    source_affine[:, 3] = np.append(np.random.randint(0, 4, 3), 1)
+    source_data = rand_gen.random_sample(source_shape)
+    source_img = Nifti1Image(source_data, source_affine)
+
+    target_shape = (11, 10, 9)
+    target_data = np.zeros(target_shape)
+    target_affine = source_affine
+    target_affine[:3, 3] -= 3  # make the affine 3 larger in x, y, z
```
GaelVaroquaux:
```suggestion
    target_affine[:3, 3] -= 3  # add an offset of 3 in x, y, z
```

GaelVaroquaux:
I believe that the decrease in test coverage is a glitch of codecov (it tends to do this).

nilearn/image/tests/test_resampling.py
gkiar: yep - thanks :)

nilearn/image/__init__.py
gkiar: I'll make private, I think you're right

gkiar:
Great - addressed comments :)

GaelVaroquaux:
pyflakes has a warning that worries me:

https://travis-ci.org/nilearn/nilearn/jobs/522125937#L464

```
nilearn/image/image.py:406:30: F821 undefined name 'args'
                         len(args))
```

gkiar:
Just noticed as well - when renaming a parameter from `args` to `pad_sizes` using the online comment we didn't catch all instances in the function. Just resolved.

GaelVaroquaux:
CircleCI is taking for ever. Merging this in the interest of moving forward.

Hurray!! :rocket:

PR #2032, opened 2019-04-19
Added a fetcher for the fsaverage spherical meshes

SylvainTakerkart:
The new fetcher comes as a new option to the existing fetch_surf_fsaverage function, as discussed with @kchawla-pi and @GaelVaroquaux during the sprint...

These meshes are necessary for the surface-based searchlight example.


jeromedockes:
why not add the sphere mesh to the dictionary returned by fetch_fsaverage when fetching 'fsaverage5' ? this mesh seems to play a similar role to the 'infl' surfaces, ie same triangles but the nodes are moved, inflating the brain for a different display

jeromedockes:
so there could be 'pial_left', 'infl_left', 'spherical_left', 'sulc_left', and *_right

kchawla-pi:
> why not add the sphere mesh to the dictionary returned by fetch_fsaverage when fetching 'fsaverage5'

That was the original plan. We changed it as fsaverage5 is bundled with Nilearn for basic plotting. spherical_meshes are not needed for that. TO bundle it with the vendored fsaverage5 would increase the package size.

Hence the decision was to provide it separately as a download.

jeromedockes:
> Hence the decision was to provide it separately as a download.

thanks!

SylvainTakerkart:
thanks for the reviews!

> I think it is important to state somewhere why this mesh is loaded.

I just added something in the docstring for this!

(I will work on what's leftover to do now... hopefully in the next few days...)

GaelVaroquaux:
Can you merge with master, so that CI runs without failing. Thanks!

GaelVaroquaux:
LGTM. Merging. Thank you!

PR #2034, opened 2019-04-19
Add a pinned Nibabel min version for testing

kchawla-pi:


GaelVaroquaux:
What's the purpose of this?

kchawla-pi:
The minimum version of Nibabel we support is 2.0.2 but we haven't pinned that anywhere, so the CI (even the minimum dependency version I merged today) tests using Nibabel=2.4.0. I am trying to pin the version to 2.0.2 (but so far it is not taking).

Issue #1871, opened 2018-11-27
datasets.fetch_atlas_harvard_oxford()  got wrong number of label


poppintiger:
I want to get the ROI timeseries from rest fmri:
`mask_data = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
mask = mask_data.maps
labels = mask_data.labels`

There 49 labels in this atlas, but when I do the converting, the label changed from different rest fmri files.

`masker = NiftiLabelsMasker(labels_img=mask, standardize=True,memory='nilearn_cache', verbose=10)
time_series = masker.fit_transform(dataset, confounds=compcor)
print(np.shape(time_series))`

The result turn out to be (150, 46) or (150, 47), but I am supposed to get the right shape （150, 49). Is that also likely because my fmri is not in MNI space?

poppintiger:
[https://pan.baidu.com/s/1Itwy3PwgfEzY0icNlYC8zA](url)
This is my frmi data.

kchawla-pi:
Hi @poppintiger , you can write code blocks by encasing them in triple backticks. That will make this much easier to read.

What is `dataset` and `compcor`? Would you add that to this code snippet?

KamalakerDadi:
@poppintiger for some reason I could not access the data from the links you are providing. 

I suppose you used NiftiLabelsMasker to extract timeseries. I would advice you to change ```resampling_target=labels```. Change this resampling strategy and let us know.

poppintiger:
> compcor
@kchawla-pi  thanks for your reminder!
’‘’
data = datasets.fetch_adhd(n_subjects=1)
dataset = data.func[0]
compcor = high_variance_confounds(dataset, n_confounds=5)

mask_data = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
mask = mask_data.maps
labels = mask_data.labels

from nilearn.input_data import NiftiLabelsMasker
masker = NiftiLabelsMasker(labels_img=mask, standardize=True,memory='nilearn_cache', verbose=10)

time_series = masker.fit_transform(dataset, confounds=compcor)
print('timeseries:',np.shape(time_series))
‘’‘
The mask have 49 labels, but the result of my time series only have 48 labels and the number changed from different rest fmri image.

poppintiger:
Hi ，@KamalakerDadi ，I have tried your recommended parameter，but the final time series still lack several labels .

KamalakerDadi:
[...]

PR #2035, opened 2019-04-19
[WIP] Modify fetch_development_fmri to fetch adults or children

illdopejake:
This updates the fetch_development_fmri to do the following:
* Added an argument adult_or_child that lets the user decide if they want to pull only adults or only children
* Made it so the first subject(s) pulled are adults, unless adult_or_child=='child'

@emdupre if you want to look over it. I have not made any project on the n<155 issue
@gkiar I did my first test if you want to have a look.


nilearn/datasets/func.py
```diff
@@ -2078,21 +2085,34 @@ def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
...
+
+    child_adult = participants['Child_Adult'].tolist()
+    child_count = child_adult.count('child') if adults_or_children != 'adults' else 0
+    adult_count = child_adult.count('adult') if adults_or_children != 'children' else 0
+    if adults_or_children == 'both':
+        max_subjects = len(participants)
+    elif adults_or_children == 'children':
+        max_subjects = child_count
+    elif adults_or_children == 'adults':
+        max_subjects = adult_count
```
gkiar: You can remove this `if` block with `max_subjects = child_count + adult_count`.

nilearn/datasets/func.py
```diff
@@ -1995,7 +1995,8 @@ def _fetch_development_fmri_functional(participants, data_dir, url, verbose):
 
 
 def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
-                           data_dir=None, resume=True, verbose=0):
+                           data_dir=None, resume=True, verbose=0,
+                           adults_or_children = 'both'):
```
GaelVaroquaux:
```suggestion
                           adults_or_children='both'):
```

PEP8 style guides says that for keyword arguments there should not be spaces on the side of "="

illdopejake:
might explain #1871

nilearn/datasets/func.py
```diff
@@ -2076,35 +2083,44 @@ def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
...
         n_subjects = max_subjects
 
-    # Download functional and regressors based on participants
-    child_count = participants['Child_Adult'].tolist().count('child')
-    adult_count = participants['Child_Adult'].tolist().count('adult')
-
     # To keep the proportion of children versus adults
-    n_child = np.round(float(n_subjects) / max_subjects * child_count).astype(int)
-    n_adult = np.round(float(n_subjects) / max_subjects * adult_count).astype(int)
+    n_child = (np.round(float(n_subjects) / max_subjects * child_count).astype(int)) if n_subjects>1 else 0
```
bthirion: pep8:  n_subjects > 1

nilearn/datasets/func.py
```diff
@@ -2027,6 +2028,12 @@ def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
     verbose: int, optional (default 0)
         Defines the level of verbosity of the output.
 
+    adults_or_children: str, optional (default 'both')
+        Whether to fetch only adults, only chilren, or both
```
emdupre:
```suggestion
        Whether to fetch only adults, only children, or both
```

nilearn/datasets/func.py
emdupre: This line is also a little too long for PEP8... One (light) suggestion would be:

```python
percent_total = float(n_subjects) / max_subjects
n_child = np.round(percent_total * child_count).astype(int) if n_subjects > 1 else 0
n_adult = np.round(percent_total * adult_count).astype(int) if n_subjects > 1 else 1
```

nilearn/datasets/func.py
```diff
@@ -2076,35 +2083,44 @@ def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
     # Participants data: ids, demographics, etc
     participants = _fetch_development_fmri_participants(data_dir=data_dir,
                                                         url=None,
-                                                        verbose=verbose)
+                                                        verbose=verbose
+                                                        )[::-1] # so adults come first
```
emdupre: Do we need to do this given [L2121](https://github.com/nilearn/nilearn/pull/2035/files#diff-472cc8a2f83316878a7f02c131580611R2121) ?

illdopejake: Yeah -- unfortunately the next line does not actually pay attention to the order of the hstack. When I was doing it locally, it wasn't actually changing the order

emdupre:
@illdopejake I just ran this fetcher and am getting all 155 subjects -- not sure why it wouldn't have worked for you previously... Can you confirm it still isn't working on your machine with the master branch ?

nilearn/datasets/func.py
```diff
@@ -2076,35 +2083,45 @@ def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
...
     # First, restrict the csv files to the adequate number of subjects
     child_ids = participants[participants['Child_Adult'] ==
                              'child']['participant_id'][:n_child]
     adult_ids = participants[participants['Child_Adult'] ==
                              'adult']['participant_id'][:n_adult]
-    ids = np.hstack([child_ids, adult_ids])
+    ids = np.hstack([adult_ids, child_ids])
     participants = participants[np.in1d(participants['participant_id'],
-                                        ids)]
+                                         ids)]
```
emdupre: It looks like this space was accidental ?

nilearn/datasets/func.py
emdupre: That's odd.... maybe I can dig into that.

illdopejake:
> Can you confirm it still isn't working on your machine with the master branch ?

Still getting the 83, even in master. However, I passed a different directory to data_dir, and it downloaded all 155 to that directory...

nilearn/datasets/func.py
emdupre: Looks like the "problem" comes in at `np.in1d`. It returns a sorted array, and since children have the earlier IDs, they're returned first.One idea, then would be:

```python
participants = participants[np.in1d(participants['participant_id'], ids)][ids.argsort()]
```

Since we'd be passing the indices of the unsorted array, this would return the list with the ids in the ordering generated from `np.hstack`. 

WDYT ?

illdopejake: I'm into it! I'll make the adjustment

illdopejake: hmm.. that didn't work for me actually when I ran it locally.

emdupre: Weird -- it works for me locally. What are you getting back ?

illdopejake: Sorry should've been more specific. I did: 
```
ds = datasets.fetch_development_fmri(n_subjects=5)
```
and I got 
```
array([('sub-pixar002',  4.8569473 , '4yo', 'child', 'F', 'R'),
       ('sub-pixar003',  4.15331964, '4yo', 'child', 'F', 'R'),
       ('sub-pixar004',  4.47364819, '4yo', 'child', 'F', 'R'),
       ('sub-pixar123', 27.06      , 'Adult', 'adult', 'F', 'R'),
       ('sub-pixar001',  4.77481177, '4yo', 'child', 'M', 'R')],
      dtype=[('participant_id', '<U12'), ('Age', '<f8'), ('AgeGroup', '<U6'), ('Child_Adult', '<U5'), ('Gender', '<U4'), ('Handedness', '<U4')])
```

emdupre: Hmm. That doesn't match what I'm seeing...

But this is really small so I don't want to hold this PR up over it !

illdopejake: 🤔 
I'm too jetlagged to figure this out. i'll try again in the morning perhaps?

kchawla-pi:
>  I passed a different directory to data_dir, and it downloaded all 155 to that directory...

@illdopejake @emdupre 
Could it be that the fetcher sees some (already downloaded) files in the default directory and decides that the dataset is already downloaded? Could you move/rename the existing data in the default download directory and try re-fetching and check?

illdopejake:
> Could it be that the fetcher sees some (already downloaded) files in the default directory and decides that the dataset is already downloaded?

I don't think so. I tried deleting a few files in the directory. Nilearn found the directory and downloaded the missing files.

> Could you move/rename the existing data in the default download directory and try re-fetching and check?

So to make matters more odd, I went to the directory that nilearn found, and in fact, all 155 images are there. This is really bizarre but I don't know what to tell you to be able to reproduce this. I think we can assume its just idiosyncratic to something I'm doing / have done locally, and may not be worth looking into..

nilearn/datasets/func.py
GaelVaroquaux: Did this question get resolved? (and are y'all over your jetlag? :smile: )

nilearn/datasets/tests/test_func.py
```diff
@@ -668,3 +668,15 @@ def test_fetch_development_fmri():
                                        verbose=1)
     confounds = np.recfromcsv(data.confounds[0], delimiter='\t')
     assert_equal(len(confounds[0]), 28)
+
+    # check first subject is an adult
+    data = func.fetch_development_fmri(n_subjects=1, reduce_confounds=False,
+                                       verbose=1)
+    age_group = data.phenotypic['Child_Adult'][0]
+    assert_equal(age_group,'adult')
```
GaelVaroquaux: That's the right test!!

nilearn/datasets/func.py
illdopejake: @emdupre and I met and solved it! I think this one is all set @GaelVaroquaux !

GaelVaroquaux:
Can you merge master, so that circle runs. 
Thanks!

illdopejake:
> Can you merge master, so that circle runs.
> Thanks!

Hmm.. CircleCI still failing..

GaelVaroquaux:
Circle is passing. Azure pipelines is failing.

kchawla-pi:
I have deactivated Azure Pipelines again.
Never mind that for now.

GaelVaroquaux:
I wanted to merge this PR, but there is something very strange: in the functional-connectivity group-analysis example, all methods give exactly 0.8 classification accuracy. This seems fishy to me. I don't have time to look at it in details right now, but we should check that something didn't go wrong.

illdopejake:
> I wanted to merge this PR, but there is something very strange: in the functional-connectivity group-analysis example, all methods give exactly 0.8 classification accuracy. This seems fishy to me. I don't have time to look at it in details right now, but we should check that something didn't go wrong.

That's fishy indeed. I have been super busy preparing for HBM but I will look into this shortly. I'm wondering if this is related to the fact that the fetcher defaults to grabbing a set ratio of children to adults. If its always grabbing 4/1 ratio, maybe the classifiers are just learning to call everyone children and get 80%.

I'll see if I can verify this when I get a free moment.

emdupre:
I wanted to check in on this -- you can see the plot_age_group_prediction_cross_val example after the most recent commit to master here:

[PR2166 CircleCI plot_age_group_prediction_cross_val artifacts](https://4826-1235740-gh.circle-artifacts.com/0/home/circleci/project/doc/_build/html/auto_examples/05_advanced/plot_age_group_prediction_cross_val.html)

I grabbed @illdopejake 's branch, merged in master locally, and built the auto example (included as a screenshot, since GitHub won't allow me to share the HTML ). 

[Local plot_age_group_prediction_cross_val build](https://user-images.githubusercontent.com/15017191/66683121-b0397e80-ec44-11e9-8200-accfa30175b0.png)

It seems that the two are showing equivalent results, to me. Is there another check we could do for this ?

kchawla-pi:
Is the classification accuracy supposed to be different? Isn't this PR only about letting users choose whether they wish to fetch data for adults or children?

emdupre:
> Is the classification accuracy supposed to be different? Isn't this PR only about letting users choose whether they wish to fetch data for adults or children?

It shouldn't be different, but I was concerned based on @GaelVaroquaux's comment (https://github.com/nilearn/nilearn/pull/2035#issuecomment-488286959) that it would be. But it doesn't seem so :smile_cat:

jeromedockes:
> It shouldn't be different, but I was concerned based on @GaelVaroquaux's comment ([#2035 (comment)](https://github.com/nilearn/nilearn/pull/2035#issuecomment-488286959)) that it would be. But it doesn't seem so

no that was due to too few CV folds in the example, it was fixed in #2059

kchawla-pi:
That would indicate this PR is ready, or ready for another look. CircleCI is failing due to OSF server problem while fetching a dataset, but I will merge master into this so anyone who wishes to review it, or sign off on it can do so.

kchawla-pi:
```python
File "/home/circleci/miniconda3/envs/testenv/lib/python3.6/site-packages/sklearn/covariance/graph_lasso_.py", line 247, in graphical_lasso
    raise FloatingPointError('The system is too ill-conditioned '
FloatingPointError: The system is too ill-conditioned for this solver. The system is too ill-conditioned for this solver
```

kchawla-pi:
Before the OSF problem I was seeing this error. We must not merge this until there is a reliable solution to this problem.

jeromedockes:
IMO the order of returned subjects should stay the same as before. users might be puzzled to see their code using this dataset start giving different results if they were using only the first few subjects

jeromedockes:
but on the other hand I guess this would concern few users since this fetcher was added very recently

nilearn/datasets/func.py
```diff
@@ -2073,33 +2080,42 @@ def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
...
-    # Download functional and regressors based on participants
-    child_count = participants['Child_Adult'].tolist().count('child')
-    adult_count = participants['Child_Adult'].tolist().count('adult')
-
     # To keep the proportion of children versus adults
-    n_child = np.round(float(n_subjects) / max_subjects * child_count).astype(int)
-    n_adult = np.round(float(n_subjects) / max_subjects * adult_count).astype(int)
+    percent_total = float(n_subjects) / max_subjects
+    n_child = np.round(percent_total * child_count).astype(int) if n_subjects > 1 else 0
+    n_adult = np.round(percent_total * adult_count).astype(int) if n_subjects > 1 else 1
```
jeromedockes: there is a bug here: when `n_subjects == 1` it always returns an adult even if `adults_or_children == "adults"`

jeromedockes:
the example is failing because it now fetches the first adult (id 123) instead of the first child (id 001), and for some reason the graphical lasso fails to converge for subject 123. this is not fixed by setting `adults_or_children to "children"` because at the moment there is no way to fetch only the first child, ie the same data as with the fetcher on master.

- we should probably take care to return the same subjects as on master if `adults_or_children` is set to the default, if this broke the nilearn example it may break other people's code too
- we should try it with more subjects and see on how many the graphical lasso fails; if it fails on many subjects it is not suited for this dataset and probably should not be showcased in `plot_sphere_based_connectome.py`

jeromedockes:
>     * we should try it with more subjects and see on how many the graphical lasso fails; if it fails on many subjects it is not suited for this dataset and probably should not be showcased in `plot_sphere_based_connectome.py`

I think there are too many seeds in the Seitzman atlas and too few time points in this dataset for the sparse inverse covariance, so we should not apply graphical lasso to this atlas in the example (and maybe not the previous one either, which has 264 regions)

kchawla-pi:
> but on the other hand I guess this would concern few users since this fetcher was added very recently

IIRC this dataset has not been in any release yet. It was merged during the sprint. We can change whatever we see fit.

jeromedockes:
>> but on the other hand I guess this would concern few users since this fetcher was added very recently
>
> IIRC this dataset has not been in any release yet. It was merged during the sprint. We can change whatever we see fit.
>
I think you're right

bthirion:
So, you want to use an atlas with fewer regions ? (makes sense to me).

jeromedockes:
> So, you want to use an atlas with fewer regions ? (makes sense to me).

the same procedure is already applied to several atlases in the example, so I would simply remove the last one as @kchawla-pi did in #2177

GaelVaroquaux:
> but on the other hand I guess this would concern few users since this fetcher was added very recently

Yes: it has not been in a released version, so I think that we are fine.

emdupre:
> there is a bug here: when n_subjects == 1 it always returns an adult even if adults_or_children == "adults"there is a bug here: when n_subjects == 1 it always returns an adult even if adults_or_children == "adults"

This is now fixed with https://github.com/nilearn/nilearn/pull/2035/commits/18eb9070f902b385f36c1c227fcd90d09ceb2396 ! 

@illdopejake also caught another corner case -- when `n_subjects=2` and `adults_or_children=both`, only two children were returned since the ratio of children to adults is so lopsided (even though a user might reasonably expect one of each when asking for both age groups and fetching only two subjects). He fixed this and added some relevant documentation in https://github.com/nilearn/nilearn/pull/2035/commits/7470f674df8796435aab2ef67ab966957e93f711 :rocket:

Even still, I think this is on hold until #2177 is merged because of the numerical instability in the example @jeromedockes and @kchawla-pi identified above.

But, one question for the group: I really don't love the `adults_or_children` parameter -- I find it's quite long and I often mistype the `children` value as `child` when working with it. WDYT of updating to `age_group` with possible values `['adult', 'child', 'both']` ?

kchawla-pi:
> WDYT of updating to age_group with possible values ['adult', 'child', 'both'] ?

+1 I like this. We are not a stickler for 'adults', I don't see why we have to be for 'children'.

nilearn/datasets/func.py
```diff
@@ -2073,33 +2084,59 @@ def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
                                                         url=None,
                                                         verbose=verbose)
 
-    max_subjects = len(participants)
+    # Download functional and regressors based on participants
+    if age_group not in ['both', 'child', 'adult']:
+        raise ValueError("Wrong value for age_group={0}. "
+                         "Valid arguments are 'adult', 'child' "
+                         "or 'both'".format(age_group))
```
kchawla-pi:
```suggestion
valid_groups = ['both', 'child', 'adult']
if age_group not in valid_groups:
        raise ValueError("Wrong value for age_group={0}. "
                         "Valid arguments are {1}".format(age_group. valid_groups))
```

nilearn/datasets/func.py
```diff
@@ -2073,33 +2084,59 @@ def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
...
         n_subjects = max_subjects
 
     if (isinstance(n_subjects, numbers.Number) and
             ((n_subjects > max_subjects) or (n_subjects < 1))):
         warnings.warn("Wrong value for n_subjects={0}. The maximum "
-                      "value will be used instead n_subjects={1}"
-                      .format(n_subjects, max_subjects))
+                      "value (for age_group={1}) will be used instead: "
+                      "n_subjects={2}"
+                      .format(n_subjects, age_group, max_subjects))
```
kchawla-pi: We need to add a test where this warning is actually raised & verified.

nilearn/datasets/func.py
emdupre: +1 on this !

emdupre:
I think this is blocking the beta release, as it will change the behavior of the development_fmri fetcher. 

... but, it seems that we're just missing tests, here ! Would you be able to look into these some time this week, @illdopejake ?

kchawla-pi:
Thanks for the phenomenal work you did for Niearn @illdopejake !
Thanks @emdupre !

bthirion:
This is awesome work indeed. Thx !

Issue #1635, opened 2018-04-24
Description of Yeo atlas has wierd type


banilo:
Small wrinkle, the description for the yeo atlas in nilearn.datasets seems to be byte string typed, which makes it difficult to print on the fly:

<img width="961" alt="screenshot 2018-04-24 09 22 10" src="https://user-images.githubusercontent.com/3907396/39173589-75ac31a8-47a5-11e8-8db4-b64ef4584a10.png">

emdupre:
I believe the byte string is from python2 , where str and bytes types are equivalent. On my system, looking at `yeo.description` in py27 has no leading `b` and the `print` function does pretty printing, as expected. If you want to render it properly in py3, decoding works: 
```
yeo = nilearn.datasets.fetch_atlas_yeo_2011()
print(yeo.description.decode("utf-8"))
```
Or, if there's interest in updating atlas descriptions in the code base, happy to open a PR !

jeromedockes:
most atlas fetchers return bytes:
```
>>> from nilearn.datasets import atlas
... import warnings
... for f in atlas.__dict__:
...     if 'fetch_atlas' in f and 'harvard' not in f:
...         args = ('ba',) if 'talairach' in f else ()
...         with warnings.catch_warnings():
...             warnings.simplefilter('ignore')
...             print('{:>35} : {}'.format(
...                 f, type(getattr(atlas, f)(*args).description)))
...                 
             fetch_atlas_allen_2011 : <class 'bytes'>
         fetch_atlas_destrieux_2009 : <class 'str'>
              fetch_atlas_talairach : <class 'str'>
               fetch_atlas_yeo_2011 : <class 'bytes'>
         fetch_atlas_surf_destrieux : <class 'bytes'>
             fetch_atlas_smith_2009 : <class 'bytes'>
   fetch_atlas_basc_multiscale_2015 : <class 'bytes'>
                    fetch_atlas_aal : <class 'bytes'>
                   fetch_atlas_msdl : <class 'bytes'>
          fetch_atlas_craddock_2012 : <class 'bytes'>
>>> 

```

most dataset fetchers too, but not all of them.

it would indeed be great if all fetchers had the same behaviour (and description should probably be a unicode string, i.e. it is the fetcher's responsibility to decode it).

jeromedockes:
related to #1489

NicolasGensollen:
Solved by #2655

PR #1608, opened 2018-02-19
Atlas craddock

darya-chyzhyk:
The idea of reorganization of the atlases in order to standardize the parameters and outputs of the atlas fetcher was mentioned in issue #1489 .
I started the modification with fetch_atlas_craddock_2012.

Issue #1489, opened 2017-08-14
ENH: Uniform access to atlas parcellations


raamana:
This is not an issue per se, but more of a discussion for improvement.

I like the dataset fetching part of nilearn, not just the individual functions, but the collection as a whole is helpful in the long run and for the newcomers. However, different methods in `nilearn.datasets` mix a few different things, especially two distinctly different things: atlases and datasets. Even for sufficiently similar things e.g. `fetch_atlas_surf_destrieux` and `fetch_surf_fsaverage5`, the return types are not consistent. My quick searching and skimming of issues and PRs did not show me any discussion and current work towards this (point me to existing/earlier discussion if one exists). Given people regularly compare few atlases in a study, I'll be surprised if one of you haven't already thought about this (or have some code lying around already taking care of this).

In that context, I think we should consider one or a combination or variation of following:
 - add `nilearn.atlas` which contains two methods: `fetch_cortical` and `fetch_wholebrain` [my vote]
 -  add `nilearn.datasets.fetch_atlas_cortical` and `nilearn.datasets.fetch_atlas_wholebrain`

and let the user supply which atlas they want, and these methods return something consistent e.g. *an atlas object* for example.

Regardless of how we provide the functionality, I think arriving at *uniform access to similar types* solution is important IMHO. For example, to make visualizations of ROI-wise mean thicknesses for different atlases, I would like to be able to do the following:
```python
import nilearn

atlas_list = [ 'fsaverage', 'glasser2016', 'destrieux', 'DKT']
for atlas in atlas_list:
    parc = nilearn.atlas.fetch_cortical(atlas)
    mean_thickness = compute_ROI_means(data, parc)
    visualize_ROI(parc, mean_thickness, atlas)

```

So defining a standard data structure for atlases (Freesurfer annotations and label files are a good start, esp. given nibabel already supports them), regardless of where it is returned from would be great.

Looking forward to learn from you guys, and contribute to this, if found helpful to others.

bthirion:
Thx for the suggestion. I like the idea of more uniform access to atlases.
I think that most atlases can be only be accessed through brain or cortex representation *exclusively*. I don't  think that the data access can be factored across these two representations.

raamana:
[...]

Issue #930, opened 2016-01-07
Have a uniform representation of labels for atlases


GaelVaroquaux:
We should do the work of grabbing the labels of the atlases in the downloaders and expose them uniformely. It would make the life of the users easier.

AlexandreAbraham:
What do you mean?

GaelVaroquaux:
Return in-memory labels, loaded from the CSV file. The code to load
labels in Harvard Oxford and MSDL are very different.

AlexandreAbraham:
OK, you mean that labels are not loaded in MSDL. They are loaded in Harvard Oxford.

GaelVaroquaux:
> OK, you mean that labels are not loaded in MSDL. They are loaded in Harvard
> Oxford.

Good point.

banilo:
Perhaps in form of an additional option in ``NiftiLabelsMasker`` and ``NiftiMapsMasker`` ?

GaelVaroquaux:
> Perhaps in form of an additional option in NiftiLabelsMasker and
> NiftiMapsMasker ?

I'd rather not for now.

banilo:
+1 for unifying label meta-data!

PR #2037, opened 2019-04-19
[wip] Harmonize datasets.fetch_atlas output attributes 

illdopejake:
The goal is to ensure that, once fetched, the atlas objects all have consistent attributes. These attributes would have the same name, same data type, and refer to the same or similar objects. At a minimum, that would be:
* description: utf-8 encode string, a description of the atlas
* maps: string, path to the atlas file
* labels: list, string labels corresponding to each atlas label
* labels_index: list, int labels corresponding to each unique value in the atlas (or each index along the 4th dimension)

Other attributes are permissible, as is redundancy (for backwards compatibility), as long as these attributes are present. 

This is related to issues #930 (closed), #1489  and #1635 , and PR #1608

### DIFF
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -99,9 +99,11 @@ def fetch_atlas_destrieux_2009(lateralized=True, data_dir=None, url=None,
     lateralized: boolean, optional
         If True, returns an atlas with distinct regions for right and left
         hemispheres.
+
     data_dir: string, optional
         Path of the data directory. Use to forec data storage in a non-
         standard location. Default: None (meaning: default)
+
     url: string, optional
         Download URL of the dataset. Overwrite the default URL.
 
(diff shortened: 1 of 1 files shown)


illdopejake:
@GaelVaroquaux 
For 3D atlases, atlas.labels refers to integer labels making up a 3D atlas.
For 4D atlases, atlas.labels refers to indexes along the 4th dimension.

Since these would be applied totally differently, would it make sense to give them different names? E.g. "region_labels" and "network_labels" or something like that?

GaelVaroquaux:
I agree with the principle but I am not convinced by the names proposed. "networks" is used in some atlases to denote a group of maps.

illdopejake:
> I agree with the principle but I am not convinced by the names proposed.

Yes good point. What about something like:

3D: atlas.labels, atlas.labels_index
4D: atlas.maps_labels, atlas.maps_index

bthirion:
+1 for these terms.

jeromedockes:
> description: utf-8 encode string, a description of the atlas

shouldn't the description be decoded for the user, ie provided as a unicode string?

illdopejake:
I have no opinions on this one ;-). Was just going with what people seemed to agree on at #1635  

I'm about to push a few commits that I did on the plane, but it would be easy to go back and change this to whatever is agreed upon :-)

bthirion:
Has been stalled for three years. Closing.

PR #2038, opened 2019-04-20
Fix warning generated in `development.rst`

kchawla-pi:
Changed text reference to use __ instead of _ to avoid warnings

 - Fix based on https://github.com/sphinx-doc/sphinx/issues/3921

### DIFF
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -75,7 +75,7 @@ In practice:
 
 Special cases:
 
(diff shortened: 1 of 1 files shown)


kchawla-pi:
I switched CircleCI to my fork, which is why it is not showing here. I will update this space once it has run.

kchawla-pi:
Took 5 hrs but CircleCI is green. Merging.

Issue #2043, opened 2019-04-24
Doc build failure with the latest numpydoc


GaelVaroquaux:
https://circleci.com/gh/nilearn/nilearn/4297

kchawla-pi:
DOnt think this is a numpydoc problem. COuld be a problem due to the caching that I couldn't completely implement.

I will revert CircleCI to pre-sprint configuration and then we will see if master is still failing.

kchawla-pi:
@GaelVaroquaux You are right, the error shows up with v0.9.1 not with 0.8.0 . I will see if I can fix the error, if not I will pin the version in CircleCI for now.
Also, I am reverting to the presprint CircleCI configuration. Don't want unreliable caching/building.

GaelVaroquaux:
> if not I will pin the version in CircleCI for now.

If we need to pin versions, we should go back to vendoring. Pinned
versions are a problem for users / developers.

The best fix, though, would be to fix the problem of the docstring.

kchawla-pi:
> we should go back to vendoring. Pinned versions are a problem for users / developers.

How? I would like to understand that.  Do pinning and vendoring not share the same benefits and drawbacks?

kchawla-pi:
I will pin it for now so the other PRs don't get stalled. I intend to fix the problem. Their error message is cryptic and unhelpful. I will reach out to the devs and see what's it about.

kchawla-pi:
Found this: https://github.com/numpy/numpydoc/issues/206

Seems the See Also sections are causing a problem with the latest version. If I remove the See Also from one docstring, it then fails on the See Also of another docstring. In the issue above, the authors seem to think this is catching tenplating errors that had been ignored all this time, and so won't be changing this. 

I'll explore more.

kchawla-pi:
Okay I think I have figured one of them out, might be applicable to all of them.

The docstring spec is to use See Also for linking, not as a general description.
Currently, our docstring uses See Also like this:

```rst
    See Also
    --------
    nilearn.datasets.load_mni152_template for details about version of the
        MNI152 T1 template and related.
```
However, the spec expects us to write See Also like this:
```rst
    See Also
    --------
    nilearn.datasets.load_mni152_template : for details about version of the
        MNI152 T1 template and related.
```
In the past this was not enforced. Now it is being enforced.
I will change the See Also to be in line with the spec.

GaelVaroquaux:
Good job!!!

kchawla-pi:
Fixed in #2046

PR #2046, opened 2019-04-25
Fix numpydoc parseerror in CircleCI

kchawla-pi:
Fixes issue #2043


GaelVaroquaux:
Circle built. 

I suggest that we merge. Congratulations !

PR #2051, opened 2019-05-02
[MRG] make html templates valid

jeromedockes:
https://validator.w3.org/nu/


jeromedockes:
ping @kchawla-pi these are tiny changes to make generated connectome and slice viewers valid HTML5.

PR #2055, opened 2019-05-14
[MRG] turn print into warning and don't warn when filename is specified in …

jeromedockes:
…open_in_browser


nilearn/plotting/js_plotting_utils.py
```diff
@@ -167,14 +167,19 @@ def open_in_browser(self, file_name=None, temp_file_lifetime=30):
         if file_name is None:
             fd, file_name = tempfile.mkstemp('.html', 'nilearn_surface_plot_')
```
kchawla-pi: Turn this into a context

jeromedockes: no, the file must still exist after the function returns. besides this is outside the scope of this pr

kchawla-pi: Ok

kchawla-pi:
We should add a test to ensure the warnings work as expected. I have learnt they are fickle things.
Also, I think we should put the functionallity in a try-finally block, with the finally calling the cleanup function. That might be separate PR. You should do that.

jeromedockes:
> We should add a test to ensure the warnings work as expected

will do!

jeromedockes:
> Also, I think we should put the functionallity in a try-finally block, with the finally calling the cleanup function. That might be separate PR. You should do that.

which functionality? if the cleanup you mean is deleting the temp file, it has to be done asynchronously to give time to the browser to open it

kchawla-pi:
OK.

nilearn/plotting/tests/test_js_plotting_utils.py
```diff
@@ -310,6 +315,10 @@ def test_temp_file_removing():
             os.remove(html._temp_file)
         except Exception:
             pass
+        try:
+            os.remove(tmpfile)
+        except Exception:
```
kchawla-pi: Whoa, very nicely done @jeromedockes !
You don't need to add `exception` if you are catching them all.

kchawla-pi: Thanks for letting me know that some exceptions such as KeyboardInterruptException don't derive from the `Exception` class, but would be caught by the `except`.

nilearn/plotting/tests/test_js_plotting_utils.py
```diff
@@ -295,11 +295,16 @@ def test_temp_file_removing():
     html = js_plotting_utils.HTMLDocument('hello')
     wb_open = webbrowser.open
     webbrowser.open = _open_mock
+    fd, tmpfile = tempfile.mkstemp()
     try:
+        os.close(fd)
+        assert_no_warnings(html.open_in_browser,
```
kchawla-pi: The trouble with this approach is if in future some component called in this function's stack generates a warning, say a FutureWarning, then this test will fail for no reason.

nilearn/plotting/tests/test_js_plotting_utils.py
```diff
@@ -295,11 +295,16 @@ def test_temp_file_removing():
...
+                           file_name=tmpfile, temp_file_lifetime=None)
         html.open_in_browser(temp_file_lifetime=.5)
         assert os.path.isfile(html._temp_file)
         time.sleep(1.5)
         assert not os.path.isfile(html._temp_file)
+        assert_warns(UserWarning, html.open_in_browser, temp_file_lifetime=None)
```
kchawla-pi: Similarly, if in Future a different warning is generated by any of th components, then we will have 2 warnings and this test may fail.

nilearn/plotting/tests/test_js_plotting_utils.py
jeromedockes: warnings are not meant to be systematically ignored. if this test fails because of a new warning we will fix the cause of the warning or change the test and add a comment explaining why the warning is acceptable.

nilearn/plotting/tests/test_js_plotting_utils.py
jeromedockes: same comment as above https://github.com/nilearn/nilearn/pull/2055#discussion_r285024690

kchawla-pi:
> warnings are not meant to be systematically ignored....

I was not advocating for ignoring the warnings. I was saying that unit tests should specifically only test for the one thing they are supposed to be testing, cause the purpose of a unit test is to quickly pinpoint the exact what & where of problem. A unit test failing due  to reasons unrelated to what it is testing creates confusion, not help narrow down focus with speed.
If more warnings pop up, we should look at them in integration tests.

That might be more labour & the cost may not be worth the benefit, I'll leave that assessment for this PR to you. I am clarifying my previous comment, and if you wish to give this some more thought, now or in future.

jeromedockes:
>  A unit test failing due to reasons unrelated to what it is testing creates confusion, not help narrow down focus with speed.

sure you're right. your comments are adressed in the last commit

kchawla-pi:
You used PyTests! :heart: 
We should add its installation in the CIs then, and the `requirements-dev,txt`.
I think you need to pull from `master`.

jeromedockes:
> You used PyTests! heart
> We should add its installation in the CIs then, and the `requirements-dev,txt`.
> I think you need to pull from `master`.

sorry I got confused and forgot we had not moved to pytest yet. should I revert
those changes? I think we'll start using pytest eventually anyway since nose is
no longer maitained and scikit-learn uses pytest.

kchawla-pi:
Don't revert it yet. I think we can start with Pytest now

kchawla-pi:
I think pytest can successfully run all our nosetests, I will verify this of course but I think it is true. If it is, we can switch now and write subsequent tests using PyTest, while I work on adapting existing nose tests to pytests in a separate branch.

jeromedockes:
@kchawla-pi this is the first PR to benefit from your nose -> pytest switch!

kchawla-pi:
Woot!

jeromedockes:
and it's ready for merge; let's do it before #2019 starts modifying `js_plotting_utils` to avoid conflicts

nilearn/plotting/tests/test_js_plotting_utils.py
```diff
@@ -310,6 +319,10 @@ def test_temp_file_removing():
             os.remove(html._temp_file)
         except Exception:
             pass
+        try:
+            os.remove(tmpfile)
+        except Exception:
+            pass
```
kchawla-pi: How do we know when this test fails? All the exceptions are silently passed.

jeromedockes: it's not a test, it is just cleaning up after the test

kchawla-pi:
Add some comments about why all the exceptions are being passed silently in the tests? It would be useful for understanding this.

kchawla-pi:
OK LGTM. If you wanna add the comment that's fine, else we can merge.

PR #2057, opened 2019-05-15
[MRG] load surf data in view_surf

jeromedockes:
otherwise if it is provided as a file path the function fails e.g.
```
plotting.view_surf(fsaverage['pial_left'], fsaverage['sulc_left'],
                   cmap='Greys', threshold=None, symmetric_cmap=False)

```


kchawla-pi:
Hey, this is cool. Would you document the reason for this PR in the first comment, so there is a record of the problem or even better file an issue and link it here?

kchawla-pi:
I was about to merge it, we should add Fix in whats_new .

nilearn/plotting/html_surface.py
```diff
@@ -297,9 +297,10 @@ def view_surf(surf_mesh, surf_map=None, bg_map=None, threshold=None,
     if surf_map is None:
         surf_map = np.ones(len(surf_mesh[0]))
     if surf_map is not None:
```
kchawla-pi: SHouldn't we just make this `else:`?

jeromedockes: :laughing: why make things simple when you can make them complicated?

kchawla-pi:
WHy did none of the tests catch this earlier? Do we have a test that catches this now?

jeromedockes:
> Do we have a test that catches this now?

yes it is added in test_html_surface.py in this pr

jeromedockes:
@kchawla-pi  do you have more comments or can we merge?

PR #2059, opened 2019-05-16
[MRG+1] 3 fold cv -> 100 fold shuffle split in group-level connectivity example

jeromedockes:


jeromedockes:
ping @GaelVaroquaux

jeromedockes:
also, would it be possible to rename this example so that the name indicates connectivity features are used for a supervised task? plot_group_level_connectivity makes it seem like it is just another example plotting connectivity matrices or connectomes

bthirion:
Sorry, isn't it taking ages with 100 splits ?
B

jeromedockes:
> Sorry, isn't it taking ages with 100 splits ?

no, the connectivity matrices are computed once on the whole dataset (btw doesn't that cause leakage from the test set?)

then the cross-validation of the SVC on this tiny dataset (30 samples) takes very little time even with 100 splits. but actually I can change it to 15 and the scores look about the same

GaelVaroquaux:
> also, would it be possible to rename this example so that the name indicates
> connectivity features are used for a supervised task?

OK, but not using the term "supervised task", as it wouldn't be
understood. Rather "comparing_subjects_with_connectivity.

bthirion:
But, as you say, if you use tangent embedding, then the embedding computation should be cross-validated.

jeromedockes:
> OK, but not using the term "supervised task", as it wouldn't be understood. Rather "comparing_subjects_with_connectivity.

how about "predicting_age_group_with_connectivity"?

GaelVaroquaux:
Great job, Jerome!

+1 for merge on my side.

kchawla-pi:
> how about "predicting_age_group_with_connectivity"?

You gonna rename the file and/or example title?

jeromedockes:
> You gonna rename the file and/or example title?

let's merge this and fix the plot with this tiny change. we can open another pr to improve this example by:

- mentioning age group classification in the file name and title
- adding error bars to the bar plot or replacing it with a box plot
- adding a discussion section that explains that this is a small example to quickly demo nilearn's features, but that we do not recommend drawing conclusions from comparisons with such a tiny sample from a single dataset, and that in the paper the cross-validation is more rigorous (tangent embedding is computed using the train set only).

jeromedockes:
I suggest we merge this quick fix now, and move any further discussion about this example to #2063

bthirion:
+1 for merge

PR #2073, opened 2019-06-10
Adding a fetcher for MPI-Leipzig_Mind-Brain-Body Open dataset

majdabd:
Added a fetcher for the MPI-Leipzig_Mind-Brain-Body (MPI-LMBB) fully preprocessed and open dataset. The fetcher allows the user to download anatomical and functional resting state data ( 4 rs-fmri scans in both MNI2mm space and native space ) for 187 participants along with confounds files and quality control pdf files for each scan/participant. It offers the option of choosing between spaces MNI2mm/native/both , phase encoding directions AP/PA/both, and runs 1/2/both. In addition, it offers the option of downloading only the anatomical MP2RAGE images without downloading the functionals.  I also imported 4 extra packages : glob , urllib, requests, and ssl that are utilized by the fetcher. 

At your disposal for any further adjustments and amelioration of the code. 

Authored-by : Majd Abdallah < majd.abdallah@u-bordeaux.fr // majd-876@hotmail.com >

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -12,6 +12,10 @@ import nibabel
 from sklearn.datasets.base import Bunch
 from sklearn.utils import deprecated
 
+import ssl
+import glob,urllib
+import requests
+
 from .utils import (_get_dataset_dir, _fetch_files, _get_dataset_descr,
                     _read_md5_sum_file, _tree, _filter_columns)
 from .._utils import check_niimg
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
The datasets in nilearn are for teaching and illustration purposes.

We are currently saturating our continuous integration with dataset download. Hence, I believe that we need to be restrictive in adding new datasets.

I think that an inclusion criteria should be: does it enable rewriting several examples to be more didactic?

nicofarr:
Hi Gaël,

More didactic - no. Better SnR (because longer scans = 15 Min), yes. 

We could work on an example (or replacing a current example) with a single subject from the leipzig MPILMBB dataset, if you think this could of interest. Otherwise, this fetcher can live separetely from nilearn.

kchawla-pi:
Another criterion would if we can use a dataset to modify/replace examples for faster computation without reducing their didactic value. We did that recently with  fetch_development_fmri dataset in PR #1953

bthirion:
Closing for now, as we don't have a strong motivation to add it.

PR #2075, opened 2019-06-18
[MRG] allow setting or disabling warning about too many open views

### LINKED ISSUE #2074: "UserWarning: It seems you have created more than 10 nilearn views." But no way to delete nilearn views
When more than 10 _html_view_ objects via `html_view = nilearn.plotting.view_img()` in a loop, you will see the warning:

> UserWarning: It seems you have created more than 10 nilearn views. As each view uses dozens of megabytes of RAM, you might want to delete some of them.

This is a helpful warning, but it doesn't give the user any way to deal with the problem, as there doesn't seem to be any documented way to free the memory associated with these objects. There is also no clean way (besides e.g. https://stackoverflow.com/questions/5644836/in-python-how-does-one-catch-warnings-as-if-they-were-exceptions/30368735) to disable the warning for users that are sure they have sufficient memory, and/or don't wish to have the console spammed.

It is also not clear to me from the [source code](https://github.com/nilearn/nilearn/blob/2df00ffb1826550b9c8e6497c7c67756e2188039/nilearn/plotting/js_plotting_utils.py#L102) how to work around this: I don't seem to be able to call `html_view._all_open_html_repr.clear()` , `html_view._all_open_html_repr.remove()`, `HTMLDocument._all_open_html_repr.clear()` or etc to manually shrink the size of the `WeakRef` object which has the length causing the error message, although this may be due to my lack of familiarity with these objects. 

Any workaround would be appreciated.

jeromedockes:
closes #2074

### DIFF
--- a/nilearn/plotting/js_plotting_utils.py
+++ b/nilearn/plotting/js_plotting_utils.py
@@ -24,6 +24,17 @@ from .._utils.extmath import fast_abs_percentile
 from .._utils.param_validation import check_threshold
 from .. import surface
 
+MAX_IMG_VIEWS_BEFORE_WARNING = 10
+
+
+def set_max_img_views_before_warning(new_value):
+    """Set the number of open views which triggers a warning.
+
+    If `None` or a negative number, disable the memory warning.
+    """
+    global MAX_IMG_VIEWS_BEFORE_WARNING
+    MAX_IMG_VIEWS_BEFORE_WARNING = new_value
+
 
 def add_js_lib(html, embed_js=True):
     """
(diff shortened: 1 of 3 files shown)


nilearn/plotting/js_plotting_utils.py
```diff
@@ -99,7 +110,12 @@ def __init__(self, html, width=600, height=400):
 
     def _check_n_open(self):
         HTMLDocument._all_open_html_repr.add(self)
-        if len(HTMLDocument._all_open_html_repr) > 9:
+        if MAX_IMG_VIEWS_BEFORE_WARNING is None:
+            return
+        if MAX_IMG_VIEWS_BEFORE_WARNING < 0:
+            return
+        if len(HTMLDocument._all_open_html_repr
```
kchawla-pi: Can we do:
```python
         if len(HTMLDocument._all_open_html_repr) ... :
                  # display warning
         else:
                  return
```

jeromedockes: we need to check if `MAX_IMG_VIEWS_BEFORE_WARNING` is not `None` before comparing it to an int. also, this makes it more explicit in which cases the check is not performed

nilearn/plotting/js_plotting_utils.py
```diff
@@ -99,7 +110,12 @@ def __init__(self, html, width=600, height=400):
...
+               ) > MAX_IMG_VIEWS_BEFORE_WARNING - 1:
             warnings.warn('It seems you have created more than 10 '
```
bthirion: Prbably need to changer this number then.

jeromedockes:
@kchawla-pi any other comments?

kchawla-pi:
Nope.

jeromedockes:
> Nope.

i guess it's ready for merge then

kchawla-pi:
We didn't add a whats_new entry for this . Isn't this a user-facing change?

Issue #1752, opened 2018-09-05
Switch from nosetests to pytest


kchawla-pi:
Should be simple, pytest can run most aspects of nosetests. I did a pilot in https://github.com/neurospin/pypreprocess and it worked fine (undid it due to other reasons). We replace the command nosetests with pytest and that's pretty much it.

bthirion:
Nosetests is no longer supported if I'm not mistaken.

kchawla-pi:
The trigger for the change is nosetests is no longer being maintained. 
nose2 is not looking to introduce new features anytime soon. pytests are the current flavor. They are popular, so it will be easier to find resources to help with them. I expect they will get feature improvements and bug fixes faster as well.
pytests can run nose and unittests without any change.

I have written a few simple tests using unittest and more complex ones using pytest Haven't used nose before.
Between unittest and pytest, pytest allows writing a lot more sophisticated tests with a lot less boilerplate.

jeromedockes:
here is the discussion where scikit-learn decided to make the same change:

https://github.com/scikit-learn/scikit-learn/issues/7319

kchawla-pi:
Fixed in #2076

PR #2076, opened 2019-06-20
[MRG] Switch from Nose to Pytest

kchawla-pi:
Issue #1752

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,3 +6,8 @@ omit =
 	plot_.*.py
 	conf\.py
 
+[report]
+omit =
+	plot_.*.py
+	conf\.py
+
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,5 @@
 [run]
 branch = True
-source = nilearn
 parallel = True
 omit =
 	*test*
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,9 +3,11 @@ branch = True
 source = nilearn
 parallel = True
 omit =
+	*test*
 	conf\.py
 
 [report]
 omit =
+	*test*
 	conf\.py
 
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,10 +1,4 @@
 [run]
 branch = True
 parallel = True
-omit =
-	conf\.py
-
-[report]
-omit =
-	conf\.py
 
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,11 +3,9 @@ branch = True
 source = nilearn
 parallel = True
 omit =
-	plot_.*.py
 	conf\.py
 
 [report]
 omit =
-	plot_.*.py
 	conf\.py
 
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,11 +2,9 @@
 branch = True
 parallel = True
 omit =
-	*test*
 	conf\.py
 
 [report]
 omit =
-	*test*
 	conf\.py
 
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,8 @@
+[run]
+branch = True
+source = nilearn
+parallel = True
+omit =
+	plot_.*.py
+	conf\.py
+
--- a/Makefile
+++ b/Makefile
@@ -34,11 +34,9 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-    pwd
     pytest -s -vv --duration=0
 	# $(TEST_RUNNER) -s $(TEST_RUNNER_OPTIONS)
 test-doc:
-    pwd
     pytest -s -vv --duration=0
 	# $(TEST_RUNNER) -s --with-doctest --doctest-tests --doctest-extension=rst \
 	# --doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,6 @@
 
 PYTHON ?= python
 CYTHON ?= cython
-TEST_RUNNER ?= pytest
-TEST_RUNNER_OPTIONS := --durations=0 -vv
 CTAGS ?= ctags
 
 all: clean test doc-noplot
@@ -31,15 +29,15 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	python -m $(TEST_RUNNER) --pyargs nilearn --cov=nilearn
+	python -m pytest --pyargs nilearn --cov=nilearn
 
 test-doc:
-	$(TEST_RUNNER) -s --doctest-glob='*.rst' `find doc/ -name '*.rst'`
+	pytest --doctest-glob='*.rst' `find doc/ -name '*.rst'`
 
 
 test-coverage:
 	rm -rf coverage .coverage
-	$(PYTEST) --pyargs nilearn --showlocals -vv --cov=nilearn --cov-report=html:coverage
+	pytest --pyargs nilearn --showlocals --cov=nilearn --cov-report=html:coverage
 
 test: test-code test-doc
 
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	python -m pytest /tmp/nilearn/ -s -vv --durations=0
+	python -m pytest -s -vv --durations=0
 test-doc:
-	python -m pytest /tmp/nilearn/ -s -vv --durations=0
+	python -m pytest -s -vv --durations=0
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,7 @@ test-code:
 	python -m $(TEST_RUNNER) --pyargs nilearn --cov=nilearn -s -vv --durations=0
 
 test-doc:
-	$(TEST_RUNNER) -s --with-doctest --doctest-tests --doctest-extension=rst \
-	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
+	$(TEST_RUNNER) -s --doctest-glob='*.rst'
 
 
 test-coverage:
--- a/Makefile
+++ b/Makefile
@@ -34,10 +34,14 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	$(TEST_RUNNER) nilearn $(TEST_RUNNER_OPTIONS)
+    pwd
+    pytest -s -vv --duration=0
+	# $(TEST_RUNNER) -s $(TEST_RUNNER_OPTIONS)
 test-doc:
-	$(TEST_RUNNER) -s --with-doctest --doctest-tests --doctest-extension=rst \
-	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
+    pwd
+    pytest -s -vv --duration=0
+	# $(TEST_RUNNER) -s --with-doctest --doctest-tests --doctest-extension=rst \
+	# --doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-    pytest -s -vv --duration=0
+	pytest -s -vv --duration=0
 test-doc:
-    pytest -s -vv --duration=0
+	pytest -s -vv --duration=0
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -35,11 +35,11 @@ inplace:
 
 test-code:
     pytest -s -vv --duration=0
-	# $(TEST_RUNNER) -s $(TEST_RUNNER_OPTIONS)
+#	$(TEST_RUNNER) -s $(TEST_RUNNER_OPTIONS)
 test-doc:
     pytest -s -vv --duration=0
-	# $(TEST_RUNNER) -s --with-doctest --doctest-tests --doctest-extension=rst \
-	# --doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
+#	$(TEST_RUNNER) -s --with-doctest --doctest-tests --doctest-extension=rst \
+#	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	pytest -s -vv --duration=0
+	pytest nilearn -s -vv --duration=0
 test-doc:
-	pytest -s -vv --duration=0
+	pytest nilearn -s -vv --duration=0
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	python -m pytest nilearn -s -vv --durations=0
+	python -m pytest --pyargs nilearn -s -vv --durations=0
 test-doc:
-	python -m pytest nilearn -s -vv --durations=0
+	python -m pytest --pyargs nilearn -s -vv --durations=0
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	python -m pytest --pyargs nilearn -s -vv --durations=0
+	python -m pytest --pyargs nilearn -s -v --durations=0
 test-doc:
-	python -m pytest --pyargs nilearn -s -vv --durations=0
+	python -m pytest --pyargs nilearn -s -v --durations=0
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	pytest nilearn -s -vv --duration=0
+	pytest nilearn/ -s -vv --duration=0
 test-doc:
-	pytest nilearn -s -vv --duration=0
+	pytest nilearn/ -s -vv --duration=0
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,6 @@ PYTHON ?= python
 CYTHON ?= cython
 TEST_RUNNER ?= pytest
 TEST_RUNNER_OPTIONS := --duration=0 -vv
-# TEST_RUNNER ?= nosetests
-# TEST_RUNNER_OPTIONS := $(shell pip list | grep nose-timer > /dev/null && \
-#                       echo '--with-timer --timer-top-n 50')
 CTAGS ?= ctags
 
 all: clean test doc-noplot
@@ -35,11 +32,8 @@ inplace:
 
 test-code:
     pytest -s -vv --duration=0
-#	$(TEST_RUNNER) -s $(TEST_RUNNER_OPTIONS)
 test-doc:
     pytest -s -vv --duration=0
-#	$(TEST_RUNNER) -s --with-doctest --doctest-tests --doctest-extension=rst \
-#	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 PYTHON ?= python
 CYTHON ?= cython
 TEST_RUNNER ?= pytest
-TEST_RUNNER_OPTIONS := --duration=0 -vv
+TEST_RUNNER_OPTIONS := --durations=0 -vv
 CTAGS ?= ctags
 
 all: clean test doc-noplot
@@ -31,16 +31,15 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	python -m $(TEST_RUNNER) --pyargs nilearn --cov=nilearn -s -vv --durations=0
+	python -m $(TEST_RUNNER) --pyargs nilearn --doctest-modules --cov=nilearn -s -vv --durations=0
 
 test-doc:
-	$(TEST_RUNNER) -s --doctest-glob='*.rst' --doctest-modules
+	$(TEST_RUNNER) -s --doctest-glob='*.rst' `find doc/ -name '*.rst'`
 
 
 test-coverage:
 	rm -rf coverage .coverage
-	$(TEST_RUNNER) -s --cov=nilearn --cover-html --cover-html-dir=coverage \
-	--cover-package=nilearn nilearn
+	$(PYTEST) --pyargs nilearn --showlocals -vv --cov=nilearn --cov-report=html:coverage
 
 test: test-code test-doc
 
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	python -m $(TEST_RUNNER) --pyargs nilearn --doctest-modules --cov=nilearn -s -vv --durations=0
+	python -m $(TEST_RUNNER) --pyargs nilearn --cov=nilearn
 
 test-doc:
 	$(TEST_RUNNER) -s --doctest-glob='*.rst' `find doc/ -name '*.rst'`
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,11 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	python -m $(TEST_RUNNER) --pyargs nilearn -s -v --durations=0
+	python -m $(TEST_RUNNER) --pyargs nilearn --cov=nilearn -s -v --durations=0
 test-doc:
-	python -m $(TEST_RUNNER) --pyargs nilearn -s -v --durations=0
+	$(NOSETESTS) -s --with-doctest --doctest-tests --doctest-extension=rst \
+	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
+
 
 test-coverage:
 	rm -rf coverage .coverage
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	pytest nilearn/ -s -vv --duration=0
+	python -m pytest -s -vv --duration=0
 test-doc:
-	pytest nilearn/ -s -vv --duration=0
+	python -m pytest -s -vv --duration=0
 
 test-coverage:
 	rm -rf coverage .coverage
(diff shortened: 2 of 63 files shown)


appveyor.yml
```diff
@@ -24,7 +24,7 @@ install:
   # See similar fix which made for travis and circleci
   # https://github.com/nilearn/nilearn/pull/1525
   # Should be removed after a new matplotlib release 2.1.1
-  - "conda install pip numpy scipy scikit-learn nose wheel matplotlib -y -q"
+  - "conda install pip numpy scipy scikit-learn nose pytest wheel matplotlib -y -q"
```
jeromedockes: why do we still need to install node?

kchawla-pi: The code itself uses modules from nose. Until we rewrite all the tests to use pytest, nose must remain.

kchawla-pi:
I haven't changed the scripts to ru the tests using nose yet, I wll tomorrow. Wanted to get the PR and the momentum started.

emdupre:
Very excited for this !!!!! Will make #2019 much easier for me :raised_hands: 

Chiming in so I'll get notifications and be reminded to review :smile_cat:

kchawla-pi:
Thanks @emdupre Much appreciated!!

appveyor.yml
jeromedockes: ok thanks!

kchawla-pi:
`nilearn/surface/tests/test_surface.py::test_load_surf_data_file_glob` Is failing in Pytest and Nose due to failing assertion that arrays are equal. They are not the same shape even. Will investigate.

Makefile
```diff
@@ -32,14 +31,13 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	$(NOSETESTS) -s nilearn $(NOSETESTS_OPTIONS)
+	python -m pytest nilearn -s -vv --durations=0
```
jeromedockes:
```suggestion
	python -m pytest --pyargs nilearn -s -vv --durations=0
```

Makefile
```diff
@@ -32,14 +31,13 @@ inplace:
...
 test-doc:
-	$(NOSETESTS) -s --with-doctest --doctest-tests --doctest-extension=rst \
-	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
+	python -m pytest nilearn -s -vv --durations=0
```
jeromedockes:
```suggestion
	python -m pytest --pyargs nilearn -s -vv --durations=0
```

kchawla-pi:
Thanks @jeromedockes . I was thinking how unclean  it would be to CD into the sitepackages directory. Applying your feedback now.

jeromedockes:
for the failure due to no .coverage try replacing `--with-coverage` with `--cov=nilearn`

jeromedockes:
for the failure when matplotlib is not installed you need to skip the nose `SkipTest` with pytest.skip

jeromedockes:
also, instead of having the logic to skip plotting tests in nilearn code, I think it would be better to add this:

```python
try:
    import matplotlib
except ImportError:
    collect_ignore = ["nilearn/plotting"]
```

in a `conftest.py` file in the nilearn root directory, and remove the code in nilearn that checks if we are running nose and raises SkipTest if matplotlib is not installed

Makefile
```diff
@@ -32,14 +31,13 @@ inplace:
...
+	python -m $(TEST_RUNNER) --pyargs nilearn -s -v --durations=0
 test-doc:
-	$(NOSETESTS) -s --with-doctest --doctest-tests --doctest-extension=rst \
-	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
+	python -m $(TEST_RUNNER) --pyargs nilearn -s -v --durations=0
 
 test-coverage:
 	rm -rf coverage .coverage
-	$(NOSETESTS) -s --with-coverage --cover-html --cover-html-dir=coverage \
+	$(TEST_RUNNER) -s --cov=nilearn --cover-html --cover-html-dir=coverage \
```
jeromedockes: this rule is actually never used, you need to edit `test-code`

Makefile
```diff
@@ -32,14 +31,13 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	$(NOSETESTS) -s nilearn $(NOSETESTS_OPTIONS)
+	python -m $(TEST_RUNNER) --pyargs nilearn -s -v --durations=0
```
jeromedockes:
```suggestion
	python -m $(TEST_RUNNER) --pyargs nilearn --cov=nilearn -s -v --durations=0
```

Makefile
```diff
@@ -32,14 +31,13 @@ inplace:
...
 test-doc:
-	$(NOSETESTS) -s --with-doctest --doctest-tests --doctest-extension=rst \
-	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
+	python -m $(TEST_RUNNER) --pyargs nilearn -s -v --durations=0
```
jeromedockes: why has this rule become a duplicate of test-code? test-doc should do something different

kchawla-pi:
Adding installation of pytest-cov, required to use switch `pytest --cov=nilearn`

jeromedockes:
@kchawla-pi you are getting there. add the `coftest.py` I suggested above (https://github.com/nilearn/nilearn/pull/2076#issuecomment-513139865) and the ci should be green

jeromedockes:
the problem is that conftest.py doesn't get copied to site-packages

kchawla-pi:
Looking into this more.
FYI
The allowed to fail test that is failing now fails flake8 check, as per the log:
```shell
conftest.py:2:5: F401 'matplotlib' imported but unused
    import matplotlib
```
This would indicate that `conftest.py` is part of site-packages.

Makefile
kchawla-pi: It needs to be rewritten to work with pytest. Will do that once test-code works

jeromedockes:
no, flake8 is run on the source dir.

so that conftest gets installed instead put it in the `/nilearn/` directory (ie
not the root directory, the one that contains an `__init__.py`, and change the
content to:

```python
try:
    import matplotlib
    # avoid flake8 unused import
    matplotlib
except ImportError:
    collect_ignore = ["plotting"]
```

(note that `"plotting"` replaces `"nilearn/plotting"`)

kchawla-pi:
Now gonna fix the make-doc commands. Then the tests should pass.

kchawla-pi:
Gotta replace 
```bash
--with-doctest --doctest-tests --doctest-extension=rst \
	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
```
Are each of these options parsed with OR or AND?

jeromedockes:
for pytest use --doctest-glob to add ".rst" files and --doctest-modules for docstrings:

https://docs.pytest.org/en/latest/doctest.html

kchawla-pi:
Already done.

What's a .inc file? I don't have any locally? SO says its conventionally used with PHP. Why do we have that here?

jeromedockes:
I think you can remove this extension

Makefile
```diff
@@ -32,14 +31,15 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	$(NOSETESTS) -s nilearn $(NOSETESTS_OPTIONS)
+	python -m $(TEST_RUNNER) --pyargs nilearn --cov=nilearn -s -vv --durations=0
+
 test-doc:
-	$(NOSETESTS) -s --with-doctest --doctest-tests --doctest-extension=rst \
-	--doctest-extension=inc --doctest-fixtures=_fixture `find doc/ -name '*.rst'`
+	$(TEST_RUNNER) -s --doctest-glob='*.rst' --doctest-modules
```
jeromedockes: I think this will result in all the other tests (besides doctest) being run twice, because they are already run in test-code and you removed `find doc/ -name '*.rst'`

I don't think pytest has an option to collect only doctests yet

https://github.com/pytest-dev/pytest/issues/4726

kchawla-pi: WOuldn't `--doctest-glob` replace `find doc/ -name '*.rst'`?
I will try `--doctest-glo='doc/*.rst'` .

jeromedockes: if we follow what is done in scikit-learn and nilearn master, doctests in docstrings will be tested in `test-code` and those in rst files will be tested in test-doc

to do this:
- add the `--doctest-modules` option to pytest in test-code, either in the makefile or by adding a `[pytest]` section to the `setup.cfg` 
- select only .rst files as input to pytest in `test-doc` by invoking like `pytest [options] $(find -name "*.rst")`

jeromedockes: > WOuldn't `--doctest-glob` replace `find doc/ -name '*.rst'`?

no, one is telling it in which files to look for doctests while the other is the input (which when not provided defaults to the current directory ie all of nilearn in this case)

Makefile
```diff
@@ -32,14 +31,15 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	$(NOSETESTS) -s nilearn $(NOSETESTS_OPTIONS)
+	python -m $(TEST_RUNNER) --pyargs nilearn --cov=nilearn -s -vv --durations=0
```
jeromedockes:
```suggestion
	python -m $(TEST_RUNNER) --doctest-modules --pyargs nilearn --cov=nilearn -s -vv --durations=0
```

Makefile
jeromedockes:
```suggestion
	$(TEST_RUNNER) $(find -name "*.rst")
```

kchawla-pi: Yeah didn't work when I tried it

kchawla-pi: Whyy are you giving me code samples for everything in this PR? :-)

kchawla-pi:
Module docstring tests failing due to what seems to be spacing issues, so correcting that now.

jeromedockes:
> Module docstring tests failing due to what seems to be spacing issues, so correcting that now.

add this to `setup.cfg`:

https://github.com/scikit-learn/scikit-learn/blob/64a7a9daefc648db25fd79e91ae327ad53459496/setup.cfg#L7

kchawla-pi:
The extra space in the result array in the docstring `permuted_lease_squares.normalize_matrix_on_axis.normalize_matrix_on_axis()`
Seems to be causing failures in first test or the remaining ones. I haven't found something that will keep it fixed for all 5 test setups.

GaelVaroquaux:
Add a NORMALIZE_WHITESPACE test pragma, then

kchawla-pi:
I did in https://github.com/nilearn/nilearn/blob/94e4761ec6a61eaa87ad85ccfc17a4169ee86036/pytest.ini

jeromedockes:
> Add a NORMALIZE_WHITESPACE test pragma, then

@GaelVaroquaux NORMALIZE_WHITESPACE won't do the trick because before numpy 1.14
there was whitespace in some places where there is none now
(NORMALIZE_WHITESPACE treats all whitespace as equal but one space is different
from no space). I think other aspects of numpy array str have also changed in
1.14. scikit-learn's solution seems to be to run doctest only when
numpy >= 1.14; see the `conftest.py`

https://github.com/scikit-learn/scikit-learn/blob/3dc43a5c5faad88120925074c9e925a77c4d5a8b/conftest.py#L70

for discussion see https://github.com/scikit-learn/scikit-learn/pull/10835
I guess nilearn should probably do the same thing as scikit-learn.

another option is to use `set_printoptions`, which was done in scikit-learn
before disabling doctests with numpy < 1.14
( https://github.com/scikit-learn/scikit-learn/pull/10132 )

```python
np.set_printoptions(legacy="1.13")
```

jeromedockes:
@kchawla-pi please copy the part of scikit-learn conftest that skips doctest with numpy < 1.14 as suggested above

https://github.com/scikit-learn/scikit-learn/blob/3dc43a5c5faad88120925074c9e925a77c4d5a8b/conftest.py#L54

this will fix the travis failure

kchawla-pi:
CircleCI failure is due to #2095

kchawla-pi:
Yeah @jeromedockes working on that.

kchawla-pi:
Ready for review @jeromedockes .
I would prefer to move as many pytest config options into `pytest.ini` instead of in-lining it in the command line in `makefile`, so that there is one centralized pace to manage the default options.

Makefile
```diff
@@ -4,9 +4,8 @@
 
 PYTHON ?= python
 CYTHON ?= cython
-NOSETESTS ?= nosetests
-NOSETESTS_OPTIONS := $(shell pip list | grep nose-timer > /dev/null && \
-                       echo '--with-timer --timer-top-n 50')
+TEST_RUNNER ?= pytest
+TEST_RUNNER_OPTIONS := --durations=0 -vv
```
jeromedockes: TEST_RUNNER_OPTIONS doesn't seem to be used anywhere?

jeromedockes: I guess you can just remove it if you have moved those options to pytest.ini

nilearn/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+import pytest
+
+from _pytest.doctest import DoctestItem
+
+try:
+    import matplotlib
+except ImportError:
+    collect_ignore = ['plotting']
+else:
+    matplotlib
```
jeromedockes: could you add a comment stating this is just so flake8 doesn't complain?

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+doctest_optionflags = NORMALIZE_WHITESPACE ELLIPSIS
+junit_family = xunit2
+addopts = --doctest-modules -s -vv --durations=0
```
jeromedockes: doesn't `-s -vv` make the logs too verbose? is `-v` enough?

Makefile
kchawla-pi: Yeah

pytest.ini
kchawla-pi: No I don't think so, Gael specifically wanted to see the output from the tests as they were run and we need -vv for that.

jeromedockes: ok no problem (BTW it is the `-s` option that tells pytest not to capture output)

Makefile
kchawla-pi: I removed the TEST_RUNNER env var as well. Redundant.

pytest.ini
kchawla-pi: -s is included in the pytest.ini

kchawla-pi:
@jeromedockes Is this ready for merge once CIs are green?

jeromedockes:
> @jeromedockes Is this ready for merge once CIs are green?

close but not yet. your `pytest.ini` is not being used because it is not present in the test dir (nor in site-packages)

have a look at the current `test_script.sh`:

https://github.com/nilearn/nilearn/blob/e3082d87c0f5146f7c6942a86f08f38730f8fa80/continuous_integration/test_script.sh#L13

`setup.cfg` is being copied to `$TEST_RUN_FOLDER`. you need to do the same with `pytest.ini`

kchawla-pi:
Why not move them both into the package, the way we did with `conftest.py`?

jeromedockes:
finding a pytest.ini also determines pytest's `rootdir` I suggest we don't change many things at once and copy the ini file to $TEST_RUN_FOLDER. actually I would suggest we revert to having a [tool:pytest] section in setup.cfg, since nilearn had a setup.cfg before and this is what scikit-learn does

kchawla-pi:
Pytest official recommendation is to use `pytest.ini`.

Does this mean `pytest.ini` & `setup,cfg` options are not followed when we run pytest locally?

kchawla-pi:
I would rather use the manifest.in file to include both of them into the package. That seems to me the cleanest & pythonic solution.

kchawla-pi:
> close but not yet. your pytest.ini is not being used because it is not present in the test dir (nor in site-packages)

Could this be the reason our attempts to fix coverage on #2019  were not working? @jeromedockes @emdupre I will try it out in #2092

kchawla-pi:
@jeromedockes What tipped you off that the pytest.ini etc files were not being used? Did you figure that out by looking at the shell script or was there something in the log that told you?

Also, what about now, are the files being read properly?

jeromedockes:
> Could this be the reason our attempts to fix coverage on #2019 were not working? @jeromedockes @emdupre I will try it out in #2092

I haven't had time to look at #2092 yet, nor look into the coverage of externals problem yet, sorry. I'll try to have a look this week

kchawla-pi:
It is also used to package non-python files in a python package. Including HTML Templates & Config files.

kchawla-pi:
> What tipped you off that the pytest.ini etc files were not being used?

This is what would be very helpful to me. Let me know.

jeromedockes:
> It is also used to package non-python files in a python package. Including HTML Templates & Config files.

we use `setup`'s `package_data` kwarg for that in nilearn

jeromedockes:
> What tipped you off that the pytest.ini etc files were not being used?

looking at the changes i thought there was no way it would be copied from the source dir. to confirm you can look at the travis log and see that the `-s -vv` options are not applied, at least at the time I posted this comment. I'll have another look now

jeromedockes:
see in the travis log:

```
rootdir: /tmp, inifile:
```

from the output you can also see `-vv` is not applied

kchawla-pi:
Line 1331 of the log: `rootdir: /tmp`
Line 4547 of the log: `rootdir: /home/travis/build/nilearn/nilearn, inifile: pytest.ini` 

So it is being applied for doctests but not code tests?
I do think `-vv` is not applied

> we use setup's package_data kwarg for that in nilearn

It seems the now recommended way is MANIFEST.in & `include_package_data=True` kwargs in `setup.py`. It is recommended not to use both. However that is beyond the scope of this PR, so I am happy to include that in `package_data`.

I want to include the config files into the package itself. I will include them in `package_data`.

jeromedockes:
>  Line 4547 of the log: `rootdir: /home/travis/build/nilearn/nilearn, inifile: pytest.ini`

you are looking at the log of an environment that has matplotlib, and this corresponds to the `make test-doc`, which is run from the nilearn source dir (/home/travis/build/nilearn/nilearn), hence the pytest.ini is in it. the test-code is run from a separate directory to detect any packaging issues, hence the rootdir=/tmp in the log, so it does not contain the pytest.ini or setup.cfg unless they are copied there

jeromedockes:
scikit-learn also copies the `setup.cfg` to the test dir in their test_script:

https://github.com/scikit-learn/scikit-learn/blob/10998aeffebd8b071f3ece6bf250be408e67abba/build_tools/travis/test_script.sh#L30

so let's keep mimicking scikit-learn's approach in this PR which is what was done before in nilearn,; add a `[tool:pytest]` in the setup.cfg and leave the copy to the test dir in test_script.sh

discussion of any further refactoring of packaging and continuous integration can be deferred to #2093

jeromedockes:
> I want to include the config files into the package itself. I will include them in `package_data`.

also pytest looks for a pytest.ini file from the common ancestor of the arguments (default `pwd`) **upwards**. that means that if you place the pytest.ini or setup.cfg file in the package (ie /nilearn/nilearn/, the folder that contains an __init__.py file), running `pytest` from the root of the repository will result in the ini file not being found

jeromedockes:
> Pytest official recommendation is to use `pytest.ini`.

if you really don't want the `[tool:pytest]` section in `setup.cfg`, you can still use a pytest.ini if it seems very important, but in this case just copy it to the `$TEST_RUN_FOLDER` just as what is done with `setup.cfg` in `test_script.sh`

kchawla-pi:
Its fine, I'll go with your suggestions. Should I do the same with `.coveragerc`, copy it in the `test_script.sh`?

jeromedockes:
you can copy it to the test_dir or give its path as the value of `--cov-config`. AFAIK you need to specify `--cov-config`` anyway.

jeromedockes:
so I would:
- not use the `Makefile` in `test_script.sh` . instead specify the full command `python -m pytest --cov-config="$OLDPWD/.coveragerc" --pyargs nilearn`
- in the `Makefile` test the source directory, not the installed nilearn i.e. the rule to make `test-code` would look like `pytest [...] nilearn`


this is what is done in scikit-learn

kchawla-pi:
WHat about now? DO you think we need to do this?

kchawla-pi:
I don't understand the coverage result....

kchawla-pi:
The coverage upload failed this time.

kchawla-pi:
We have enabled branch coverage, which measures coverage of branches in code, such as if-else statement.
It would seem we were not doing this earlier, as the absolute coverage  dropped by 2% after enabling this, with lots of yellow lines in the coverage report.
This we have not written tests to cover alternate branch executions (for example, test for when if condition is true, but not when it's false  - the else condition).
We either turn-off branch coverage, let abs coverage drop by 2% , or add code in test to check alternate path coverage.

kchawla-pi:
Ready for review.

jeromedockes:
I think we should definitely enable branch coverage. as you say the reported coverage will drop a little (but of course the code that is actually covered by the tests doesn't change)

kchawla-pi:
I concur.

You think this is ready?

jeromedockes:
> You think this is ready?

I will review in 15 mn if that's okay. I think the pytest command is still missing `--cov-config` but I will have a look at the full diff and then comment

kchawla-pi:
I don't think that's necessary. The config files are being correctly applied. I'll await your assessment.

jeromedockes:
> I don't think that's necessary. The config files are being correctly applied. I'll await your assessment.

sorry you're right, .coveragerc is the default

azure-pipelines.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
   - script: |
       pip install .
-      nosetests ./nilearn -v
+      pytest ./nilearn -v
```
jeromedockes: this was the case before so don't change it in this PR but unlike on travis, on azure the tests are run from the source dir

.coveragerc
```diff
@@ -0,0 +1,13 @@
+[run]
+branch = True
+source = nilearn
+parallel = True
+omit =
+	*test*
+	conf\.py
```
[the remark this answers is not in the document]
kchawla-pi: Yup.

.coveragerc
```diff
@@ -0,0 +1,13 @@
+[run]
+branch = True
+source = nilearn
+parallel = True
```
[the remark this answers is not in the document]
kchawla-pi: If its unused then we should remove it. Cruft like this would make it harder to read the existing code, esp for new devs.
OTOH pytest-cov is not providing us any overt benefits I had added that before I realized the site-packages thing. We can remove it and if it doesn't break anything we can leave it out.

Makefile
```diff
@@ -32,15 +29,15 @@ inplace:
 	$(PYTHON) setup.py build_ext -i
 
 test-code:
-	$(NOSETESTS) -s nilearn $(NOSETESTS_OPTIONS)
+	python -m pytest --pyargs nilearn --cov=nilearn
```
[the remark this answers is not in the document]
kchawla-pi: I think that's fine, because they can run pytest directly for that. & Best practices dictate to install Nilearn using pip in editable mode anyway for the paths to work properly during development.

Personally I am not a big fan of so many nested files & scripts (make file, test_script) but that is a a discussion for another PR.

azure-pipelines.yml
kchawla-pi: Yeah, there are deeper issues with Azure (failing tests) which I will work on in a separate PR. I will do that once the current batch of my PRs are merged.

kchawla-pi:
> my main comment is about the `--pyargs` in the makefile. we can remove `--pyargs` in the makefile and not use make in the `test_script.sh` -- as is done in scikit-learn (or add a `test-installed` target)

I think if we change that we should handle it in a separate PR.

.coveragerc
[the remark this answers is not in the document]
jeromedockes: > OTOH pytest-cov is not providing us any overt benefit

sorry I'm not sure what you mean? pytest-code is needed to get coverage reports right?

azure-pipelines.yml
jeromedockes: sounds good

jeromedockes:
@kchawla-pi have you looked into why no coverage data is collected?

.coveragerc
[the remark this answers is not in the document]
jeromedockes: pleas try removing `source = nilearn` since it is both a path and a package name it seems that can interfere with coverage collection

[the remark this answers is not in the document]
kchawla-pi: > sorry I'm not sure what you mean? pytest-code is needed to get coverage reports right?

I'm tallking about `pytest-cov` plugin that this PR is currently installing.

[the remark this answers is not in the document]
jeromedockes: AFAIK it is required to get a .coverage from running pytest isn't that the case?

Makefile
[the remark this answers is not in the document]
jeromedockes: > Personally I am not a big fan of so many nested files & scripts (make file, test_script) but that is a a discussion for another PR.

:+1:

kchawla-pi:
NOt sure, it was collected in 98b357b , then not in the next one, 8f3b7ab . If it does not work this time, I'll revert the omit =*test* & check again.

jeromedockes:
> I'll revert the omit =_test_ & check again.

:+1: 
ah yes I hadn't seen this one

kchawla-pi:
All right! Anything else?

.coveragerc
```diff
@@ -0,0 +1,10 @@
+[run]
+branch = True
+parallel = True
+omit =
+	conf\.py
+
+[report]
+omit =
+	conf\.py
```
jeromedockes: what is `conf\.py`?

kchawla-pi: in the `doc` directory.

kchawla-pi: But yeh  dunno if this will reach it or not, or if we need to bother including this. Looking.

kchawla-pi: Actually I can't remember where I got the idea to include this.

kchawla-pi: I think it is so `conf.py` & `conftest.py` are not covered but shouldn't that need `conf*.py`?

kchawla-pi: I don't think the doc directory is covered at all, do we need to bother with this?

jeromedockes: if you want it to match also `conftest.py` you do need the wildcard. also there is no need to escape the `.`, the syntax being used here is glob's not regular expressions. I don't think you need to omit this file anyway

jeromedockes: > I don't think the doc directory is covered at all, do we need to bother with this?

the `doc/` directory doesn't get installed in site-packages so it is not visible to pytest nor coverage

kchawla-pi: Removed all omit entries for this PR. I am hoping we can merge this today!

jeromedockes:
what could be omitted from coverage are the glass brain scripts in nilearn/plotting/glass_brain_files but it's no big deal.

note that a lot of code that is not covered is code that deals with python2 compatibility, since we don't support and test python 2 anymore.

.coveragerc
jeromedockes: waiting for CI to finish to have a last quick look but I think this PR is ready to be merged

kchawla-pi:
You mean the 2 python scripts `plot_align_svg.py` & `svg_to_json_converter.py`. We can add those. Will
`nilearn/plotting/glass_brain_files/*.py` work?

jeromedockes:
> Will
> `nilearn/plotting/glass_brain_files/*.py` work?

I would start this path with a wildcard otherwise it is interpreted as relative to the current directory. ie `*/nilearn/plotting/glass_brain_files/*.py`

kchawla-pi:
Why are we excluding this from coverge checking? IF we are not testing this, then the report should tell us that we are not testing this.

jeromedockes:
> Why are we excluding this from coverge checking? IF we are not testing this, then the report should tell us that we are not testing this.

fair enough; let's keep the .coveragerc as it is. we'll just add `externals` when we need it

kchawla-pi:
Cool. Anything else? I am excited!

kchawla-pi:
Thanks for all the massive help @jeromedockes !

Issue #2077, opened 2019-07-01
 TravisCI failing due to Conda not supporting older versions


kchawla-pi:
@GaelVaroquaux Similar to https://github.com/nistats/nistats/issues/363 , this problem is surfacing in Nilearn as well.The default conda channel no longer supports `scipy==0.17` and `scikit-learn==0.18`. To keep things simple for users, should we bump up the minimum supported versions. Or only address this in the CIs by switching channels or switch to `pip` instead of `conda`?

GaelVaroquaux:
OK, let's do it!

PR #2078, opened 2019-07-02
Bumped up the minimum versions of scikit-learn, scipy to 0.19, mpl to 2.0

kchawla-pi:
#2077
 - Conda default channel dropped support for these versions.
 - Updated Whats_new.rst

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
   include:
     # without matplotlib
     - env: DISTRIB="conda" PYTHON_VERSION="3.5"
-           NUMPY_VERSION="1.11" SCIPY_VERSION="0.19" PANDAS_VERSION="*"
+           NUMPY_VERSION="1.12" SCIPY_VERSION="0.19" PANDAS_VERSION="*"
            SCIKIT_LEARN_VERSION="0.19" COVERAGE="true"
            LXML_VERSION="*"
     - env: DISTRIB="conda" PYTHON_VERSION="3.5"
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
   include:
     # without matplotlib
     - env: DISTRIB="conda" PYTHON_VERSION="3.5"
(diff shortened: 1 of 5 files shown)


jeromedockes:
why is it necessary to change the minimum supported version of scikit-learn scipy and matplotlib?
users who installed these libraries a while ago might still have those versions, and they can be installed with pip

jeromedockes:
for example scikit-learn still supports scipy 0.17; why does nilearn need 0.19?

kchawla-pi:
We will have to replace conda with pip in all our CIs. Or switch to
conda-forge channel, if it supports the older versions.

I'm fine with either of those scenarios if Gael is.

There's also the consideration that we move to even newer minimum versions
of that helps us address the warnings.

jeromedockes:
how does scikit-learn deal with scipy 0.17?

kchawla-pi:
Does their master branch support Scipy0.17? This change in conda happened
last week for us.

jeromedockes:
ok thanks, I didn't realize this was so recent

kchawla-pi:
@jeromedockes Did you merge this? :smile: 

> how does scikit-learn deal with scipy 0.17?

They pinned conda==4.6, that was also their solution for #2079 

@glemaitre What has SKL been doing for this? Are you still pinned to conda==4.6? Plans going forward?

jeromedockes:
> @jeromedockes Did you merge this? smile

No

kchawla-pi:
Uh, I merged this. I had forgotten. Bah.

kchawla-pi:
@jeromedockes It seems to me scikit-learn uses apt for most of their dependency installation in most of their CI runs, or they use pip. They test conda compatibility in a couple CI runs but not with their minimum supported versions.

jeromedockes:
could we use pip in some runs? that would be good regardless of scipy version

kchawla-pi:
That was one of the options presented to Gael, and he didn't take it.
Personally, I think most people using Python for data analytics gravitate towards using Anaconda across platforms.

Issue #2049, opened 2019-04-29
plotting.plot_matrix requires labels argument to be provided


bthirion:
The following example breaks
```
from nilearn import datasets
atlas = datasets.fetch_atlas_msdl()
atlas_filename = atlas['maps']
labels = atlas['labels']
data = datasets.fetch_adhd(n_subjects=1)
from nilearn.input_data import NiftiMapsMasker
masker = NiftiMapsMasker(maps_img=atlas_filename, standardize=True,
                         memory='nilearn_cache', verbose=5)
time_series = masker.fit_transform(data.func[0],
                                   confounds=data.confounds)
from nilearn.connectome import ConnectivityMeasure
correlation_measure = ConnectivityMeasure(kind='correlation')
correlation_matrix = correlation_measure.fit_transform([time_series])[0]
import numpy as np
from nilearn import plotting
np.fill_diagonal(correlation_matrix, 0)
plotting.plot_matrix(correlation_matrix, labels=[], colorbar=True,
                     vmax=0.8, vmin=-0.8)
```

By contrast replacing the last line by
```
plotting.plot_matrix(correlation_matrix, labels=labels, colorbar=True,
                     vmax=0.8, vmin=-0.8)
```

The documentation of plot_matrix should be revised (labels is currently optional) or this should be fixed.

kchawla-pi:
Is it related to the MatplotlibDeprecationWarning of always using unique axes or axes labels?

```
MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently
reuses the earlier instance. In a future version, a new instance will always be created and returned.
Meanwhile, this warning can be suppressed, and the future behavior ensured, by passing a unique label to
each axes instance. warnings.warn(message, mplDeprecation, stacklevel=1)  
```

bthirion:
AFAIK this is not the same thing.
Note that setting `auto_fit=False` solves the issue.

mjboos:
Hi,
the problem was in the logic of the if statements, since an empty list does evaluate to False, but not if directly queried (i.e. `not []` is unequal to `[] is False`).
I will go ahead and do a pull request.

jeromedockes:
fixed in #2083

PR #2083, opened 2019-07-11
Fix matrix plotting labels

mjboos:
Fixes issues #2049 i.e. `plot_matrix` breaking if an empty list is supplied as the labels argument and `auto_fit=True`.

### DIFF
--- a/nilearn/plotting/matrix_plotting.py
+++ b/nilearn/plotting/matrix_plotting.py
@@ -81,7 +81,7 @@ def plot_matrix(mat, title=None, labels=None, figure=None, axes=None,
             Axes image.
     """
     if reorder:
-        if labels is None or labels is False:
+        if not labels:
             raise ValueError("Labels are needed to show the reordering.")
         try:
             from scipy.cluster.hierarchy import (linkage, optimal_leaf_ordering,
@@ -134,10 +134,10 @@ def plot_matrix(mat, title=None, labels=None, figure=None, axes=None,
                         cmap=cmap, **kwargs)
     axes.set_autoscale_on(False)
     ymin, ymax = axes.get_ylim()
-    if labels is False:
+    if not labels:
         axes.xaxis.set_major_formatter(plt.NullFormatter())
         axes.yaxis.set_major_formatter(plt.NullFormatter())
-    elif labels is not None:
+    else:
         axes.set_xticks(np.arange(len(labels)))
         axes.set_xticklabels(labels, size='x-small')
         for label in axes.get_xticklabels():
@@ -178,7 +178,7 @@ def plot_matrix(mat, title=None, labels=None, figure=None, axes=None,
     axes.set_ylim(ymin, ymax)
 
     if auto_fit:
-        if labels is not None and labels is not False:
+        if labels:
             fit_axes(axes)
         elif own_fig:
             plt.tight_layout(pad=.1,
--- a/nilearn/plotting/matrix_plotting.py
+++ b/nilearn/plotting/matrix_plotting.py
@@ -37,8 +37,10 @@ def plot_matrix(mat, title=None, labels=None, figure=None, axes=None,
             Matrix to be plotted.
         title : string or None, optional
             A text to add in the upper left corner.
-        labels : list or ndarray of strings, optional
-            The label of each row and column
+        labels : list, ndarray of strings, empty list, False, or None, optional
+            The label of each row and column. Needs to be the same
+            length as rows/columns of mat. If False, None, or an
+            empty list, no labels are plotted.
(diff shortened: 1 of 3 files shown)


GaelVaroquaux:
Looks good, but can we have a test: something that used to break and now works.

mjboos:
Fair point, I've added two tests.

nilearn/plotting/matrix_plotting.py
```diff
@@ -80,8 +80,12 @@ def plot_matrix(mat, title=None, labels=None, figure=None, axes=None,
         display : instance of matplotlib
             Axes image.
     """
+    # we need a list so an empty one will be cast to False
+    if isinstance(labels, np.ndarray):
+        labels = labels.tolist()
+
     if reorder:
-        if labels is None or labels is False:
+        if not labels:
```
jeromedockes: I think iit is better to explicitly check if labels is `None` and if not use `len` or a numpy array's `size` if you need to check if it is empty rather than the implict cast to `bool`.

it would also make sense to check that if labels is not None it must be a sequence of the same size as the plotted matrix number of rows

mjboos: The first point might be a matter of style, but [PEP8](https://www.python.org/dev/peps/pep-0008/) recommends exactly that - casting the list to `bool`.

I implemented a check for the length of labels, good idea.

kchawla-pi: If `labels` is a pure Python list/tuple then `not labels` works fine. However if it's a numpy array, then the cast to bool is ambiguous & will raise a warning.
If it's possible this will be a numpy array then `is None` & `is False` should be used.

jeromedockes: the doc doesn't list `False` or an empty list as valid inputs. let's make the docstring more explicit that it should be `None` for no labels, or a sequence of length number of rows in the matrix, raise an error if the input is not one of these two things, build an `ndarray` from the labels if they are not None. and use explicit `is None` when needed.

mjboos: @kchawla-pi it isn't possible that it's a numpy array, and a test is implemented that checks for the correct handling of an ndarray as input
@jeromedockes fair point, the documentation is ambiguous. The motivating example for this PR was an error that arose when `labels=[]`, I think it makes sense to cover that case, keep the implicit casting to bool, then make the docs more explicit. IMO these are all reasonable user inputs and casting to bool makes the code more readable, even if it is somewhat less explicit

jeromedockes: this is true. it just feels a bit awkward having to cast the array to list just evaluate if its empty and then back to array, but it does allow handling None, False and an empty sequence uniformly. let's keep it the way you wrote it!

jeromedockes:
@mjboos the circleci failure should be fixed by merging (or rebasing on) master

mjboos:
> @mjboos the circleci failure should be fixed by merging (or rebasing on) master

I did rebase on master, but it still fails unfortunately

jeromedockes:
> I did rebase on master, but it still fails unfortunately

yest but it's a different failure, this time its an internal error in the server from which we download the development dataset. i restarted it

nilearn/plotting/tests/test_matrix_plotting.py
```diff
@@ -21,6 +21,14 @@ def test_matrix_plotting():
     # test if it returns an AxesImage
     ax.axes.set_title('Title')
     plt.close()
+    # test if an empty list works as an argument for labels
+    ax = plot_matrix(mat, labels=[])
```
jeromedockes: could we also test with `labels=None` (the default)?

kchawla-pi:
Hi @mjboos we are so close to greatness! Would you be able to complete the changes & then we can merge!

kchawla-pi:
Add it to the whats_new under Fixes.

nilearn/plotting/tests/test_matrix_plotting.py
mjboos: Sure, added the test.

kchawla-pi:
@mjboos The failure is related to conda, not changes made by you,  and I will file a fix for that and update you here.

nilearn/plotting/tests/test_matrix_plotting.py
jeromedockes: thanks!

jeromedockes:
+1 for merge

kchawla-pi:
Whats_new fixes entry pending. THen we can merge.

mjboos:
Okay, added it to what's new.

PR #2090, opened 2019-07-17
[MRG,MAINT] Fix joblib import

jeromedockes:
sklearn.externals.joblib is deprecated in 0.21 and will be removed in 0.23

this PR is to import joblib directly rather than from sklearn.externals

this adds joblib as a dependency to nilearn

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
       NUMPY_VERSION: "*"
       SCIPY_VERSION: "*"
       SCIKIT_LEARN_VERSION: "*"
-      JOBLIB_VERSION: "*"_
+      JOBLIB_VERSION: "*"
       MATPLOTLIB_VERSION: "*"
 
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,7 @@ jobs:
       NUMPY_VERSION: "*"
       SCIPY_VERSION: "*"
       SCIKIT_LEARN_VERSION: "*"
+      JOBLIB_VERSION: "*"_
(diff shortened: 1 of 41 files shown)


kchawla-pi:
Isn't Joblib already a dependency for scikit-learn? It wouldnt be a new, explicit dependency for Nilearn.

jeromedockes:
> Isn't Joblib already a dependency for scikit-learn? It wouldnt be a new, explicit dependency for Nilearn.

nilearn also uses joblib directly

kchawla-pi:
> nilearn also uses joblib directly

Yes but scikit-learn is a dependency for Nilearn and joblib is a dependency for scikit-learn. So directly using it or not, joblib doesn't have to be an explicit dependency.

jeromedockes:
ping @kchawla-pi @GaelVaroquaux ready for review

kchawla-pi:
If we have added Joblib as an explicit dependency for Nilearn, that means it will be installed by users irrespective of the version of scikit-learn. Which means we can always do
```python
import joblib  # or
from joblib import Memory
```
and so on. There should be no need for 
```python
 from nilearn._utils.compat import joblib
```
or any such thing.

jeromedockes:
> You know that, so I am missing something here?

in some places scikit-learn 0.19 raises an error if it is given a "memory" parameter that is neither a string or a `sklearn.externals.joblib.Memory`, so for e.g. searchlight to work with scikit-learn 0.19 this is necessary

kchawla-pi:
OKay I don't think there's a problem here. I don't want the CIs to start humming again or to have to do a pull from master until #2076 is merged. If that one is merged today, we can merge this after that.

PR #2091, opened 2019-07-17
Add copy=True parameter to threshold_img

dangom:
The threshold_img was not following the conventions of the other
image functions and was performing in-place thresholding without
offering the possibility to turn it off. This is now fixed.

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -172,8 +172,8 @@ def _smooth_array(arr, affine, fwhm=None, ensure_finite=True, copy=True):
         filtering.
 
     copy: bool
-        if True, input array is not modified. False by default: the filtering
-        is performed in-place.
+        if True, input array is not modified. True by default: the filtering
+        is not performed in-place.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


nilearn/image/image.py
```diff
@@ -172,8 +172,8 @@ def _smooth_array(arr, affine, fwhm=None, ensure_finite=True, copy=True):
         filtering.
 
     copy: bool
-        if True, input array is not modified. False by default: the filtering
-        is performed in-place.
+        if True, input array is not modified. True by default: the filtering
+        is not performed in-place.
```
GaelVaroquaux: Given that it's an internal function, I would leave it not doing copies.

dangom: The default in `_crop_img_to` is `copy=True`, but the documentation is wrong. I just fixed that as I copied the text over to `threshold_img`. Did I understand correctly that you'd like to change that to `copy=False`?

GaelVaroquaux:
> The default in _crop_img_to is copy=True, but the documentation is wrong. I
> just fixed that as I copied the text over to threshold_img. Did I understand
> correctly that you'd like to change that to copy=False?

OK. Sorry, I had not seen. Let's not change the behavior.

Thanks!

dangom:
The only other place using threshold_img was RegionExtractor.fit, which would mutate the image being fit. I think that's also not the intended behaviour - for clarity I added an explicit `copy=True`, even though that's the default now.

The examples have `copy=False` for efficiency, but I'd worry that new users could copy lines from the examples and have `threshold_img` bite them later.

Test added.

bthirion:
LGTM, thx !

GaelVaroquaux:
Great. Thanks! Merging!

PR #2096, opened 2019-07-22
fix imread import in example

### LINKED ISSUE #2095: scipy.misc.imread() replaced by scipy.imageio.imread() in v1.2
`scipy.misc.imread()` was deprecatd in SciPy 1.0 & replaced in SciPy 1.2 by `scipy.imageio.imread()`

https://docs.scipy.org/doc/scipy-1.2.1/reference/generated/scipy.misc.imread.html

This is causing failures in CircleCI. 

I will work on this once PR  #2076 doctest problem has been addressed, since we need this issue to be resolved before it can be merged. Intended today.

jeromedockes:
fixes #2095

### DIFF
--- a/examples/02_decoding/plot_haxby_stimuli.py
+++ b/examples/02_decoding/plot_haxby_stimuli.py
@@ -7,7 +7,6 @@ and Overlapping Representations of Faces and Objects in Ventral Temporal
 Cortex" (Science 2001)
 """
 
-from scipy.misc import imread
 import matplotlib.pyplot as plt
 
 from nilearn import datasets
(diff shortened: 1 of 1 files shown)


kchawla-pi:
I had assigned this one to myself.

jeromedockes:
> I had assigned this one to myself.

sorry didn't see that

kchawla-pi:
Oh don't close it! Now you have done the work.

kchawla-pi:
Should we do a try-except here? The minimum supported scipy is 0.19, which means the examples are expected to work with that as well.

jeromedockes:
> Should we do a try-except here? The minimum supported scipy is 0.19, which means the examples are expected to work with that as well.

I think importing `imread` from pyplot is perfecly fine and works in all cases so let's keep the example simple

examples/02_decoding/plot_haxby_stimuli.py
```diff
@@ -27,8 +26,8 @@
     for i in range(48):
         plt.subplot(6, 8, i + 1)
         try:
-            plt.imshow(imread(file_names[i]), cmap=plt.cm.gray)
-        except:
+            plt.imshow(plt.imread(file_names[i]), cmap=plt.cm.gray)
+        except Exception:
```
kchawla-pi: We should make it a more specific exception, FileNotFoundError  or something only catching missing file condition.

PR #2097, opened 2019-07-22
Brainomics fetcher

AGrigis:
Update Brainomics fetcher functions and associated tests.
Brainomics data are now on OSF.


bthirion:
Thx for taking care of this. CirceCi complains about `tmaps`...

kchawla-pi:
You can go ahead & add pandas as a dependency in the CIs, ReadMe & whats_new.

jeromedockes:
> You can go ahead & add pandas as a dependency in the CIs, ReadMe & whats_new.

@kchawla-pi are you saying that because pandas will be added with nistats anyway? has adding pandas as a dependency before the nistats merged been discussed anywhere?

kchawla-pi:
Yes to the first.
No to the second.

jeromedockes:
> No to the second.

I think we should. we may even consider a scheme similar to what we have with matplotlib where pandas is a soft dependency, required only for the GLM (former nistats) part of nilearn. at any rate extra care has to be taken to include many pandas versions in the CI, since the behaviour and interface of pandas tend to change a lot

kchawla-pi:
Agreed.
@agrigis you mentioned you can implement this with numpy. Please do that.

jeromedockes:
> @AGrigis you mentioned you can implement this with numpy. Please do that.

it it's not too much trouble that would be great indeed, otherwise don't hesitate to let us know

AGrigis:
Hi,
To replace pandas dep in the code I am using the numpy.isin function which is new in version 1.13.0.
Travis complains since you're testing against numpy 1.11.0.
What solution do you prefer:
- test against numpy >=1.13
- 'add' the isin code in nilearn
Thanks for your help.

kchawla-pi:
Hi @AGrigis 
Unfortunately we must write code compatible with numpy version 1.11 .
Whilenp.isin() is a handy function, it is doing element-wise comparison. We can replicate that functionality without resorting to is in.
Are the arrays being compared 1D? If so, we can do something like this:
```python
# np.isin(arr1, arr2) can be written as 
np.array([item in arr2 for item in arr1])
```
If it's not 1D and we only care about membership check, then we can flatten the arrays.
I'm sure there are more elegant ways to do this as well, some googling required.

nilearn/datasets/func.py
```diff
@@ -998,11 +997,11 @@ def _is_valid_path(path, index, verbose):
 
     # Load covariates file
     participants_file = os.path.join(data_dir, participants_file)
-    csv_data = pd.read_csv(participants_file, delimiter="\t")
-    csv_data = csv_data[csv_data["participant_id"].isin(subject_ids)]
+    csv_data = np.recfromcsv(participants_file, delimiter="\t")
```
kchawla-pi: PEP8 recommends using single-quotes for strings everywhere, except when one has to enclose single quotes.

jeromedockes:
> What solution do you prefer:
> 
>     * test against numpy >=1.13
> 
>     * 'add' the isin code in nilearn

Hi, we want to keep supporting numpy 1.11 but if you need you could copy-paste the numpy isin code in nilear._utils.compat, it would look like

```python
try:
    from numpy import isin
except ImportError:
    def isin(args):
        ...
```
and import it from there

AGrigis:
Hi,
The remaining error in circleci is not linked with my changes.
On appveyor I get a 404 error message when trying to fetch the localizer dataset on OSF.
Everything is OK on Travis, and I can't reproduce the error.
How works the tests on appveyor?
Thanks for your help.

kchawla-pi:
Rerunning it Should work. Hopefully the server error is a one off since
Travis passed.

jeromedockes:
>  The remaining error in circleci is not linked with my changes.

indeed, and it has been fixed on master

jeromedockes:
> Rerunning it Should work. Hopefully the server error is a one off since

AFAIK there is no way to trigger a new appveyor build besides a new commit: https://github.com/appveyor/ci/issues/15

kchawla-pi:
There is. You need the login for the appveyor account for Nilearn. I'll
email it to you if you don't have it already

kchawla-pi:
Pulling from mater will fix CircleCI error.

kchawla-pi:
Hi @AGrigis . Before Aug and vacations come into their full swing, do you think you can attend to this? We won't keep you waiting for reviews.

kchawla-pi:
Add it to the whats_new under Changes.

kchawla-pi:
Do Another pull from master to solve the Second build's Failure on Travis.
I have restarted the third TravisCI build and the Appveyor build to check if their failure was a fluke.
The CircleCI failure seems to be related to this PR.

nilearn/datasets/func.py
```diff
@@ -916,107 +924,90 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
         subject_id_max = "S%02d" % np.max(n_subjects)
         n_subjects = len(n_subjects)
     subject_ids = ["S%02d" % s for s in subject_mask]
-    data_types = ["c map"]
+    data_types = ["cmaps"]
```
kchawla-pi: Single quote.

nilearn/datasets/func.py
```diff
@@ -916,107 +924,90 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
...
-                  'S identifier <= "%s", ' % (subject_id_max, ) +
-                  'X type IN(%(types)s), X label "%(label)s"')
-
-    urls = ["%sbrainomics_data_%d.zip?rql=%s&vid=data-zip"
-            % (root_url, i,
-               _urllib.parse.quote(base_query % {"types": rql_types,
-                                          "label": c},
-                            safe=',()'))
-            for c, i in zip(contrasts_wrapped, contrasts_indices)]
+        data_types.append("tmaps")
```
kchawla-pi: Single quote

nilearn/datasets/func.py
```diff
@@ -916,107 +924,90 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
...
             for contrast_id, contrast in enumerate(contrasts_wrapped):
                 name_aux = str.replace(
                     str.join('_', [data_type, contrast]), ' ', '_')
                 file_path = os.path.join(
                     "brainomics_data", subject_id, "%s.nii.gz" % name_aux)
-                file_tarball_url = urls[contrast_id]
-                filenames.append((file_path, file_tarball_url, opts))
+                path = os.path.join(
+                    "/localizer", "derivatives", "spm_1st_level",
+                    "sub-%s" % subject_id,
```
kchawla-pi: INstead of the `%` interpolation, use `str.format()` .
And, single-quotes :-)

nilearn/datasets/func.py
```diff
@@ -916,107 +924,90 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
...
-                                    safe=',()')))
         for subject_id in subject_ids:
             file_path = os.path.join(
                 "brainomics_data", subject_id, "boolean_mask_mask.nii.gz")
-            file_tarball_url = urls[-1]
-            filenames.append((file_path, file_tarball_url, opts))
+            path = os.path.join(
+                "/localizer", "derivatives", "spm_1st_level",
+                "sub-%s" % subject_id, "sub-%s_mask.nii.gz" % subject_id)
+            if _is_valid_path(path, index, verbose=verbose):
```
kchawla-pi: Same as above.

nilearn/datasets/func.py
```diff
@@ -916,107 +924,90 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
...
-                                    "XD", safe=',')
-                       ))
-    else:
-        url_csv = "%s/cubicwebexport.csv" % url
-        url_csv2 = "%s/cubicwebexport2.csv" % url
-    filenames += [("cubicwebexport.csv", url_csv, {}),
-                  ("cubicwebexport2.csv", url_csv2, {})]
+            path = os.path.join(
+                "/localizer", "derivatives", "spm_preprocessing",
+                "sub-%s" % subject_id, "sub-%s_T1w.nii.gz" % subject_id)
```
kchawla-pi: Same as above.

nilearn/datasets/func.py
```diff
@@ -916,107 +924,90 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
...
-            file_tarball_url = urls[-1]
-            filenames.append((file_path, file_tarball_url, opts))
+            path = os.path.join(
+                "/localizer", "derivatives", "spm_1st_level",
+                "sub-%s" % subject_id, "sub-%s_mask.nii.gz" % subject_id)
+            if _is_valid_path(path, index, verbose=verbose):
+                file_url = os.path.join(root_url, index[path][1:])
+                opts = {"move": file_path}
+                filenames.append((file_path, file_url, opts))
+                files.setdefault("masks", []).append(file_path)
```
kchawla-pi: Single quotes.

nilearn/datasets/func.py
```diff
@@ -916,107 +924,90 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
...
+                "sub-%s" % subject_id, "sub-%s_T1w.nii.gz" % subject_id)
+            if _is_valid_path(path, index, verbose=verbose):
+                file_url = os.path.join(root_url, index[path][1:])
+                opts = {"move": file_path}
+                filenames.append((file_path, file_url, opts))
+                files.setdefault("anats", []).append(file_path)
+
+    # Fetch subject characteristics
+    participants_file = os.path.join("brainomics_data", "participants.tsv")
+    path = "/localizer/participants.tsv"
```
kchawla-pi: Single quotes.

nilearn/datasets/func.py
```diff
@@ -916,107 +924,90 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
...
+    if _is_valid_path(path, index, verbose=verbose):
+        file_url = os.path.join(root_url, index[path][1:])
+        opts = {"move": participants_file}
```
kchawla-pi: Single quotes.

nilearn/datasets/tests/test_func.py
```diff
@@ -146,193 +147,89 @@ def test_miyawaki2008():
...
-    assert_true(dataset.masks is None)
+    # 2 subjects
+    dataset = func.fetch_localizer_contrasts(
+        ["checkerboard"],
+        n_subjects=2,
+        data_dir=tst.tmpdir,
+        verbose=1)
+    assert_false(hasattr(dataset, "anats"))
+    assert_false(hasattr(dataset, "tmaps"))
+    assert_false(hasattr(dataset, "masks"))
```
kchawla-pi: Single quotes.

nilearn/datasets/tests/test_func.py
```diff
@@ -146,193 +147,89 @@ def test_miyawaki2008():
...
-    np.testing.assert_array_equal(dataset2.anats,
-                                  np.array(dataset.anats)[[1, 2, 4]])
-    np.testing.assert_array_equal(dataset2.cmaps,
-                                  np.array(dataset.cmaps)[[1, 2, 4]])
-    np.testing.assert_array_equal(dataset2.masks,
-                                  np.array(dataset.masks)[[1, 2, 4]])
-    np.testing.assert_array_equal(dataset2.tmaps,
-                                  np.array(dataset.tmaps)[[1, 2, 4]])
+    assert_equal([row[0] for row in dataset2.ext_vars],
+                 [b"S02", b"S03", b"S05"])
```
kchawla-pi: Single quotes.

jeromedockes:
> PEP8 recommends using single-quotes for strings everywhere, except when one has to enclose single quotes.

are you sure? https://www.python.org/dev/peps/pep-0008/#id25:
"In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. "

either way, I wouldn't worry too much about it

kchawla-pi:
Whoa! 
Okay Never mind the single/double quotes @AGrigis .
But do replace the `%` with `.format()`

kchawla-pi:
The appveyor error is for the localizer related functions. Is this a problem in the PR? Why is Travis not failing?

nilearn/datasets/func.py
DimitriPapadopoulos: On the other hand I have noticed recently that [_Black_ prefers double quotes over single quotes](https://black.readthedocs.io/en/stable/the_black_code_style.html#strings).

DimitriPapadopoulos:
What do we need to get past these failed checks?

kchawla-pi:
First step would be to pull from master and rerun the checks. Some problems here have been fixed there.

AGrigis:
Concerning the examples/05_advanced/plot_localizer_mass_univariate_methods.py the "pseudo" meta are not yet available on OSF.

AGrigis:
I think this PR is now ready to be merged (except the ugly SSL hack).

kchawla-pi:
Hey, remove the SSL hack and see if it passes now. Another PR that was failing due to this is now passing.

examples/05_advanced/plot_localizer_mass_univariate_methods.py
```diff
@@ -35,7 +35,7 @@
 
 tested_var = localizer_dataset.ext_vars['pseudo']
 # Quality check / Remove subjects with bad tested variate
-mask_quality_check = np.where(tested_var != b'None')[0]
+mask_quality_check = np.where(tested_var != b'n/a')[0]
```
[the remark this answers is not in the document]
DimitriPapadopoulos: While converting the dataset to BIDS, the empty cells of CSV files have been converted to `n/a` cells in TSV files. The tabular files have changed, I don't know how feasible it is to insulate end-users against this change.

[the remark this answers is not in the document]
GaelVaroquaux: Ok, let's live with this. Can you add a note about this in the what's new file. 

Thanks!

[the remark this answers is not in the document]
GaelVaroquaux: Ok, let's live with this. Can you add a note about this in the what's new file. 

Thanks!

examples/01_plotting/plot_dim_plotting.py
```diff
@@ -21,9 +21,9 @@
 
 localizer_dataset = datasets.fetch_localizer_button_task()
 # Contrast map of motor task
-localizer_tmap_filename = localizer_dataset.tmap
+localizer_tmap_filename = localizer_dataset.tmaps[0]
```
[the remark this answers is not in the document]
AGrigis: The 'fetch_localizer_button_task function' ships only 2nd subject (S02) specific tmap and its normalized T1 image. Now its working for all subjects so the output is a list. The API of the function also changed: new n_subjects parameter. I can make a little hack if only one subject is requested.

GaelVaroquaux:
> The 'fetch_localizer_button_task function' ships only 2nd subject (S02)
> specific tmap and its normalized T1 image. Now its working for all subjects so
> the output is a list. The API of the function also changed: new n_subjects
> parameter. I can make a little hack if only one subject is requested.

I worry about the change of return argument: this is going to break
people's code (especially code that is used for teaching).

Special-casing n_subjects=1 is a bit ugly (but probably acceptable).

How about n_subjects=None special cased not to return a list, and being
the default value. That way it's more explicit that it's a special case.

jeromedockes:
> I worry about the change of return argument: this is going to break
> people's code (especially code that is used for teaching).
>

+1 . plus, having a function that returns a single image is useful for
demos of plotting functions

> Special-casing n_subjects=1 is a bit ugly (but probably acceptable).
>
> How about n_subjects=None special cased not to return a list, and being
> the default value. That way it's more explicit that it's a special case.

+1

AGrigis:
I will update the code but note that the localizer fetchers API will be inconsitant.
In the 2 other localizer fetchers if None is given,  all 94 subjects are used.

kchawla-pi:
> How about n_subjects=None special cased not to return a list, and being
> the default value. That way it's more explicit that it's a special case.

I don't know what this means. 

If we were previously returning data for only 1 subject then we should `n_subjects=1` as default, so that the code stays backwards compatible and the function signature clearly explains what's happening here. 
`n_subjects=None` returning 1 subject is wildly counter-logic to anyone who reads that function signature, if that is what is being proposed, even if we clarify it in the docstring.

If we are going to make a breaking change, now would be the time to do it. However, I can also raise a warning for cases where there is no n_subjects parameter in user code and give them `n_subjects=1` for 2 release cycles, similar to how I changed the names of certain parameters in a previous PR.

GaelVaroquaux:
> I will update the code but note that the localizer fetchers API will be
> inconsitant.
> In the 2 other localizer fetchers if None is given, all 94 subjects are used.

OK. I really do not like.

Can we just revert to the previous API. The goal of this function was to
be very simple for simple examples and introductory courses.

kchawla-pi:
I think `n_subjects=1` as default and `n_subjects=None` for all, but not default, is the best way here.

bthirion:
But the question is also whether the output is an image for n_subejcys=1 and a list for n_subjects > 1. For this we stick to the previous conventions, right ?
B

kchawla-pi:
To maintain compatibility yeah. Either that or we do this

> However, I can also raise a warning for cases where there is no n_subjects parameter in user
> code and give them n_subjects=1 for 2 release cycles, similar to how I changed the names of 
> certain parameters in a previous PR.

GaelVaroquaux:
My point is: this complexity is not warranted, I am not sure that need
the n_subjects argument. In my eyes, these functions were introduced to
be simpler alternative to the complex function that downloads brainpedia
with all possible bells and whistles. Let's keep the simple functions
simple.

kchawla-pi:
True. So we only return 1 subject result,, following the older API output.

jeromedockes:
But isn't it a big waste to give access to only one subject? and doesn't
nistats need multiple subjects for some examples?

how about:
- n_subjects = None (default): return one image (not a list)
- n_subjects = 1 ... n : return the required number of subjects in a list
- n_subjects = -1 : return all subjects in a list

kchawla-pi:
If we do want to keep the full dataset:

For teaching we could have a function by the current name, returning 1 subject's image. 
Internally this can be a wrapper calling a full dataset function.

GaelVaroquaux:
> But isn't it a big waste to give access to only one subject?

The full dataset can be retrieved with the fetch_localizer_contrasts
function.

https://nilearn.github.io/modules/generated/nilearn.datasets.fetch_localizer_contrasts.html

The ones that we are discussing are simplified functions to
make simpler examples and teaching.

nilearn/datasets/func.py
DimitriPapadopoulos: Also I suggest using `os.path.join()` instead of `"/".join()`.

nilearn/datasets/func.py
DimitriPapadopoulos: Same too.

nilearn/datasets/func.py
```diff
@@ -916,107 +923,104 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
...
-                       _urllib.parse.quote("Any X,XI,XD WHERE X is QuestionnaireRun, "
-                                    "X identifier XI, X datetime "
-                                    "XD", safe=',')
-                       ))
-    else:
-        url_csv = "%s/cubicwebexport.csv" % url
-        url_csv2 = "%s/cubicwebexport2.csv" % url
-    filenames += [("cubicwebexport.csv", url_csv, {}),
-                  ("cubicwebexport2.csv", url_csv2, {})]
+            path = "/".join([
```
DimitriPapadopoulos: Same: `os.path.join()`

AGrigis: Will break on Windows, we have to keep "/".join().

DimitriPapadopoulos: Ah, my wrong. These are URLs, not paths. Sorry.

kchawla-pi: Maybe change the variable name to `url_path` or something better?

jeromedockes:
> The ones that we are discussing are simplified functions to
> make simpler examples and teaching.

sorry I missed that. then yes definitely always return just one image

nilearn/datasets/func.py
DimitriPapadopoulos: Right. I would avoid mixing the two terms _url_ and _path_:
* use _path_ in names of variables that contain a path and use `os.path.join()` to manipulate them,
* use _url_ in names of variables that contain a URL and use `"/".join()` to manipulate them.

Therefore I would avoid _url_path_. Perhaps _file_url_?

kchawla-pi: Sure, like I said, something better, that clearly communicates what it is and what it is not. I'll leave the choice of name to @AGrigis

GaelVaroquaux:
> Also I suggest using os.path.join() instead of "/".join().

Good catch!

nilearn/datasets/func.py
jeromedockes: or urljoin
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

jeromedockes:
the `localizer_index.json` added to test data is 1.4 M large, could we use a subset of it?

DimitriPapadopoulos:
1,4 MB is not that big. It's the size of the index of the Brainomics/Localizer file URLs on OSF. Reconstructing it by querying OSF would take minutes or even hours.

On the other hand this file is also available from OSF, from a helper auxiliary component:
https://osf.io/hwbm2/

kchawla-pi:
For running the test that the fetcher code in Nilearn is working (not that the servers are serving properly) we only need a few URLs, not the entire list

nilearn/datasets/func.py
DimitriPapadopoulos: Ah indeed, but is it worth adding a dependency on _urllib_ just for this? Perhaps we should leave it as is. Antoine won't have much more time to invest in preparing this patch. If needed I'm willing to add minor changes later on, just tell me what needs to be done. Not sure how to take over Antoine commits his changes.

kchawla-pi: `urllib` is a standard python package. Not an external dependency.

kchawla-pi:
The purpose of the test is only to check that the code written in Nilearn is working. Doesn't make sense to include the entire URL list for that.
1.4MB is not large in terms of the file sizes we use day to day but it is a very large size increase for a library, especially when there is no functionality of that increase.

nilearn/datasets/func.py
jeromedockes: keeping it as it is is also fine

jeromedockes:
> 1,4 MB is not that big. 

but it's not negligible either: nilearn/nilearn is 9.4 MB

GaelVaroquaux:
> > 1,4 MB is not that big.

> but it's not negligible either: nilearn/nilearn is 9.4 MB

Yes: if every dataset came with that size, we would be in problem.

kchawla-pi:
> nilearn/nilearn is 9.4 MB

A fresh clone of nilearn/nilearn is 2.8MB ! :-)

kchawla-pi:
@jeromedockes This seems ready. When you concur, I will merge it.

nilearn/datasets/tests/test_func.py
```diff
@@ -146,193 +148,139 @@ def test_miyawaki2008():
...
+    sid = "S{0}".format(idx)
+    LOCALIZER_INDEX.update(
+        dict((key.format(sid), uuid.uuid4().hex)
+        for key in localizer_template))
+LOCALIZER_INDEX["/localizer/phenotype/behavioural.tsv"] = uuid.uuid4().hex
+LOCALIZER_PARTICIPANTS = np.recfromcsv(
+    os.path.join(tst.datadir, "localizer_participants.tsv"), delimiter='\t')
+LOCALIZER_BEHAVIOURAL = np.recfromcsv(
+    os.path.join(tst.datadir, "localizer_behavioural.tsv"), delimiter='\t')
+def mock_localizer_index(*args, **kwargs):
```
bthirion: skip 2 lines before functions

DimitriPapadopoulos:
Oh, @kchawla-pi did spot them the double quotes. I prefer simple quotes myself but since I met [_Black_ the uncompromising Python code formatter](https://github.com/psf/black#strings) I'm not so certain any more.

kchawla-pi:
There are some small Flake8 errors, which should be fixed before merging into `master`.

kchawla-pi:
@AGrigis I will fix the Flake8 errors and merge this. Thank you for this PR and your help to Nilearn!

GaelVaroquaux:
> Merged #2097 into master.

Awesome!!! Thank you all involved. This is a big deal!!!

Issue #2069, opened 2019-06-03
GraphLassoCV renamed GraphicalLassoCV in scikit-learn


jeromedockes:
these files will need to be updated accordingly:

doc/connectivity/connectome_extraction.rst
doc/developers/group_sparse_covariance.rst
examples/03_connectivity/plot_multi_subject_connectome.py
examples/03_connectivity/plot_inverse_covariance_connectome.py
examples/03_connectivity/plot_simulated_connectome.py
examples/03_connectivity/plot_sphere_based_connectome.py
nilearn/connectome/group_sparse_cov.py

jeromedockes:
fixed by #2101

PR #2101, opened 2019-07-23
[MRG] Replace GraphLasso* with GraphicalLasso*

kchawla-pi:
For issue #2069

Since the usage is in examples & documentation, it is improbable that users will run it in the oldest support Scikitlearn version. CIrcleCI uses the latest versions. Hence not accommodating the older version immediately. Will reasses based on feedback.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -287,7 +287,7 @@ intersphinx_mapping = {
     'numpy': ('http://docs.scipy.org/doc/numpy', None),
     'scipy': ('http://docs.scipy.org/doc/scipy/reference', None),
     'matplotlib': ('http://matplotlib.org/', None),
-    'sklearn': ('http://scikit-learn.org/', None),
+    'sklearn': ('http://scikit-learn.org/latest', None),
     'nibabel': ('http://nipy.org/nibabel', None),
     'pandas': ('http://pandas.pydata.org', None),
 }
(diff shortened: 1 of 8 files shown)


jeromedockes:
to fix ci error you may need to configure sphinx to point to a more recent scikit-learn (or don't specify the version to point to the most recent release):

https://github.com/nilearn/nilearn/blob/5b7648cc4d452389f2d9320fafdfd1aa27562c0e/doc/conf.py#L290

kchawla-pi:
The examples will break if SKL < 0.20.0, but I don't think it is worth the hassle to write alternate code for this use. Maybe we can add a note in the examples that they need SKL 0.20.0 to run correctly?

jeromedockes:
the three options are:
1. add `GraphicalLasso` to `nilearn.utils.compat`
2. try/except ImportError in all 5 examples that use it (no nilearn code uses it)
3. note / warn in example stating that the import needs to be changed with older scikit-learn versions

I think I would lean towards option 2 but let's ask for @GaelVaroquaux 's opinion on this one

GaelVaroquaux:
> I think I would lean towards option 2 but let's ask for @GaelVaroquaux 's opinion on this one

+1

kchawla-pi:
:+1:

kchawla-pi:
Done. DO note that I have replace GraphLasso with GraphicalLasso as well, not just the CV version.

kchawla-pi:
Anything else?

PR #2102, opened 2019-07-24
Set verbose=1 to fetchers

### LINKED ISSUE #2100: Some fetchers are not verbose
The following fetchers have `verbose=0` by default :
- `fetch_development_fmri`
- `fetch_mixed_gambles`

It would be nice to be consistent with the other fetchers where `verbose=1`.

mrahim:
fixes #2100

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -1360,7 +1360,7 @@ def _load_mixed_gambles(zmap_imgs):
 
 
 def fetch_mixed_gambles(n_subjects=1, data_dir=None, url=None, resume=True,
-                        return_raw_data=False, verbose=0):
+                        return_raw_data=False, verbose=1):
     """Fetch Jimura "mixed gambles" dataset.
 
     Parameters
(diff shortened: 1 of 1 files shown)


nilearn/datasets/func.py
```diff
@@ -1360,7 +1360,7 @@ def _load_mixed_gambles(zmap_imgs):
 
 
 def fetch_mixed_gambles(n_subjects=1, data_dir=None, url=None, resume=True,
-                        return_raw_data=False, verbose=0):
+                        return_raw_data=False, verbose=1):
```
jeromedockes: could you update the corresponding entry in the docstring "parameters" section?

nilearn/datasets/func.py
```diff
@@ -1995,7 +1995,7 @@ def _fetch_development_fmri_functional(participants, data_dir, url, verbose):
 
 
 def fetch_development_fmri(n_subjects=None, reduce_confounds=True,
-                           data_dir=None, resume=True, verbose=0):
+                           data_dir=None, resume=True, verbose=1):
```
jeromedockes: idem

Issue #2099, opened 2019-07-23
nilearn.image.smooth_img fails if fwhm is an ndarray


pausz:
nilearn.image.smooth_image fails if fwhm is a numpy.ndarray

##### What version of Nilearn are you using?
+ 0.5.0
+ master branch e3082d8

##### What were you trying to do?
Smoothing an image with different widths along each of the spatial axes. 

##### What did you expect will happen?
to work as per the reference documentation.

##### What actually happened?
Fails with a `ValueError` because the [first check in `_smooth_array()`](https://github.com/nilearn/nilearn/blob/master/nilearn/image/image.py#L189) essentially compares a scalar to an ndarray. 

A different behaviour arises if `fwhm` is specified as a numpy ndarray with one element: There is a `FutureWarning` coming from a comparison between 
an ndarray and a string in [line 208 of image.py](https://github.com/nilearn/nilearn/blob/master/nilearn/image/image.py#L208).

##### Steps to rperoduce the bug

The steps necessary to reproduce the `FutureWarning` and `ValueError` are detailed below.

```
~/Work/Code/Python/fmriprep_sandbox/nilearn (master)
└─ 
$ ☻ ▶ ipython
Python 3.7.2 (default, Dec 29 2018, 06:19:36) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.5.0 -- An enhanced Interactive Python. Type '?' for help.

```

```python
In [1]: import nilearn as nl

In [2]: nl.__version__
Out[2]: '0.5.0'

In [3]: import nilearn.image as nl_img

In [4]: imgs = nl_img.load_img('sub-001_task-run1_space-MNI152NLin6Asym_desc-preproc_bold_confounds-removed.nii.gz')

In [5]: smoothed_img = nl_img.smooth_img(imgs, fwhm=5)   

In [6]: import numpy as np

In [7]: smoothed_img = nl_img.smooth_img(imgs, fwhm=np.array([1.]))   
/home/paula/.local/lib/python3.7/site-packages/nilearn/image/image.py:208: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if fwhm == 'fast':

In [8]: smoothed_img = nl_img.smooth_img(imgs, fwhm=np.array([1., 2.5, 4.2]))   
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-8-fb6ccd266709> in <module>()
[...]

PR #2107, opened 2019-07-25
Fix:  nilearn.image.smooth_img fails if fwhm is an ndarray

pausz:
This PR fixes the issue described in issue #2099.


jeromedockes:
Thanks a lot! could you add a regression test that fails on master but passes
with your fix? (in nilearn/image/tests/test_image.py)


also, this part of `test_compute_epi_mask`:

  https://github.com/nilearn/nilearn/blob/43b9990bc7ce063fd5ddd4d73583969c8121ca20/nilearn/tests/test_masking.py#L75

seems to fail in the CI because an extra warning is raised. could you replace it
with:

```python
    with pytest.warns(masking.MaskWarning, match="Computed an empty"):
        compute_epi_mask(mean_image, exclude_zeros=True)
```

nilearn/image/image.py
```diff
@@ -205,18 +208,25 @@ def _smooth_array(arr, affine, fwhm=None, ensure_finite=True, copy=True):
...
         arr[np.logical_not(np.isfinite(arr))] = 0
 
-    if fwhm == 'fast':
+    if isinstance(fwhm, str) and (fwhm == 'fast'):
         arr = _fast_smooth_array(arr)
     elif fwhm is not None:
+        # Assume fwhm is either a nonzero number or a tuple, or a numpy array.
+        # Whatever it is at this point, transform it into an np.array
+        fwhm = np.asarray(fwhm)
+        fwhm = np.where(fwhm == None, 0.0, fwhm)
```
jeromedockes:
```python
fwhm = np.asarray(fwhm, dtype=float)
fwhm = np.nan_to_num(fwhm)
```

this will avoid `fwhm` 's dtype to be "object" if it contained None, (and avoids flake8 complaining about the comparison to None with ==)

jeromedockes:
if you can also have a look at the CI flake8 (https://travis-ci.org/nilearn/nilearn/jobs/563331770), or run it locally, to fix it's complaints (too many blanklines, extra whitespace, line too long). thanks!

nilearn/image/image.py
```diff
@@ -205,18 +208,25 @@ def _smooth_array(arr, affine, fwhm=None, ensure_finite=True, copy=True):
         # SPM tends to put NaNs in the data outside the brain
         arr[np.logical_not(np.isfinite(arr))] = 0
 
-    if fwhm == 'fast':
+    if isinstance(fwhm, str) and (fwhm == 'fast'):
```
kchawla-pi: We don't need `isinstance(fwhm, str)` if it's in `and` with an equivalence test with a string.

nilearn/image/image.py
kchawla-pi: What Jerome said.

For general reference, equivalence with `None` should always be tested with `is`, not `==`

nilearn/image/image.py
```diff
@@ -205,18 +208,25 @@ def _smooth_array(arr, affine, fwhm=None, ensure_finite=True, copy=True):
...
         affine = affine[:3, :3]
 
         # Convert from a FWHM to a sigma:
         fwhm_over_sigma_ratio = np.sqrt(8 * np.log(2))
         vox_size = np.sqrt(np.sum(affine ** 2, axis=0))
         sigma = fwhm / (fwhm_over_sigma_ratio * vox_size)
+
         for n, s in enumerate(sigma):
-            ndimage.gaussian_filter1d(arr, s, output=arr, axis=n)
+            if s > 0.0:
```
kchawla-pi: Generally this is not a good way to compare floats. If `s == 0.00045` this will be `True` which may or may not be the intention. Generally floats are compared by `abs(s-0.0) > epsilon`, where the epsilon value is set according to desired sensitivity. One can also use `math.isclose()` or `numpy.isclose()`.

@jeromedockes WDYT?

nilearn/image/image.py
jeromedockes: we do, in case fwhm is a numpy array

kchawla-pi: if `fwhm` is a numpy array, the if statement will fail the `fwhm=='fast'` comparison regardless.

nilearn/image/image.py
jeromedockes: here fwhm is a numpy array so `==` has a different meaning and performs element-wise comparison. `fwhm is None` would not give what we need. still I would prefer `dtype=float` and `nan_to_num` to get an array of the right dtype

kchawla-pi: > so == has a different meaning and performs element-wise comparison. fwhm is None would not give what we need.

Did not know that that. Thanks @pausz & @jeromedockes

nilearn/image/image.py
kchawla-pi: I'm guessing its the same thing here, `fwhm=='fast'` will do an element-wise comparision, which is why the `isinstance` coparision is necessary?

kchawla-pi: Would `fwmh is 'fast'` work here instead of thse 2 comparisions? 
Regardless, I always forget this one, let's add a comment here explaining this for future code perusers.

nilearn/image/image.py
jeromedockes: this is true, and we could also skip the whole conversion of None to 0 and then skipping smoothing in dimensions where sigma == 0 however smoothing with a very small sigma and not smoothing is the same; the only important thing is to make sure we don't pass a negative number or exactly 0 to gaussian_filter1d as that would cause an error

nilearn/image/image.py
jeromedockes: in some cases strings can be equal but not have the same id:

https://stackoverflow.com/questions/1504717/why-does-comparing-strings-using-either-or-is-sometimes-produce-a-differe

so @pausz 's solution is the correct one here

jeromedockes: > I'm guessing its the same thing here, `fwhm=='fast'` will do an element-wise comparision, which is why the `isinstance` comparison is necessary?

exactly. `fwhm == 'fast'` only gets evaluated if the first condition is true

kchawla-pi:
Hi @pausz . A lot of people over here become unavailable in August. I am hoping we can merge this before August is in full swing. Do you think you can make the necessary changes. We won't keep you waiting for reviews.

kchawla-pi:
Add it to the whats_new under Fixes.

kchawla-pi:
Hi @pausz WOuld you pull from master, and then add a whatsnew entry? We want to release a new Nilearn version by Sept end and would like to include this PR.

kchawla-pi:
Thanks for helping us @pausz !

PR #2118, opened 2019-08-14
fix wrong urls in nki dataset

### LINKED ISSUE #2117: NKI urls for left and right hemispheres point to the same file
it seems that e.g.
https://www.nitrc.org/frs/download.php/8261/A00033747_rh_preprocessed_fsaverage5_fwhm6.gii
and
https://www.nitrc.org/frs/download.php/8261/A00033747_lh_preprocessed_fsaverage5_fwhm6.gii

point to the same file (true for other subjects as well), which corresponds to the left hemisphere.

see 
https://neurostars.org/t/preprocessed-nki-fsaverage5-surface-data/4784

@juhuntenburg do you know if the files uploaded on nitrc have changed recently?

juhuntenburg:
Fixes #2117


kchawla-pi:
What's new fixes entry?

PR #2120, opened 2019-08-22
[NOMRG] Patch & Release v0.5.3 : Don't set aspect of 3d axes: raises starting from matplotlib 3.1

kchawla-pi:
Patching in the fix we landed in Master, releasing it on PyPI.

Users were affected by this problem and a full release at this point would be premature and useless as it would be an alpha release and not installed by default by package installers.


kchawla-pi:
Think I picked the wrong base commit. Will open a new PR.

PR #2122, opened 2019-08-22
[NOMRG] Release 0.5.3

kchawla-pi:
The removal of ax.set_axes() from Matplotlib 3.1 is inconvenience to users and must be patched up. This branch aims to do that. Some other changes need to be made to make sure the CIs pass since the environment since 0.5.2 has changed. 

It is not intended to be merged into `master`. All the changes made here are already present in `master`, excepet the `whats_new`, which I will do separately.

This won't be merged with `master` so no need to resolved conflicts. I intend to push from this branch to PyPI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,6 @@ matrix:
     - env: DISTRIB="conda" PYTHON_VERSION="3.4"
            NUMPY_VERSION="1.11.2" SCIPY_VERSION="0.17"
            SCIKIT_LEARN_VERSION="0.18" MATPLOTLIB_VERSION="1.5.1"
-           SETUPTOOLS="30.3.0"
     # Most recent versions
     - env: DISTRIB="conda" PYTHON_VERSION="3.5"
            NUMPY_VERSION="*" SCIPY_VERSION="*" PANDAS_VERSION="*"
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ matrix:
     - env: DISTRIB="conda" PYTHON_VERSION="3.4"
            NUMPY_VERSION="1.11.2" SCIPY_VERSION="0.17"
            SCIKIT_LEARN_VERSION="0.18" MATPLOTLIB_VERSION="1.5.1"
+           SETUPTOOLS="30.3.0"
     # Most recent versions
     - env: DISTRIB="conda" PYTHON_VERSION="3.5"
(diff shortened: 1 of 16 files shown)


kchawla-pi:
@jeromedockes 
I need your help to quickly get this ready. One of the tests that you had modified (the RAW sprites one) is failing because the sprite's string has `/n` in the end. Any idea why this is happening?

jeromedockes:
> @jeromedockes
> I need your help to quickly get this ready. One of the tests that you had modified (the RAW sprites one) is failing because the sprite's string has `/n` in the end. Any idea why this is happening?

I think this is from #2051

jeromedockes:
> I think this is from #2051

the b64 encoded colormap should not end with a newline, and this was fixed in #2051

kchawla-pi:
Thanks, I'll try this on.

jeromedockes:
and the test was changed in #2071 
when we removed saving the sprites in "raw" format because matplotlib 3.1 doesn't support it

kchawla-pi:
There's a `setuptools>=30.3.0` not found error and a doctest error causing the failures. I will change the doctest to see if it passes on the currently failing examples (then it will fail the currently passing one) and if between the versions of doctest, if everything is green, I will push this to PyPI. 

Ignoring the setuptools problem.

WDYT @jeromedockes

kchawla-pi:
@GaelVaroquaux Yu can keep track here for the 0.5.3 release

kchawla-pi:
The tests are passing locally for python 2.7, 2.5, 3.7. It is that doctest that fails. Gonna skip it. Otherwise, since I will not push new documentation from this release, I might push to PyPI just like this, even if the tests are failing on the cloud CIs.

kchawla-pi:
@jeromedockes Once circleCI passes, I think this is good enough. The one failing test on travis is not due to our code. WDYT? I want your opinion.

jeromedockes:
> @jeromedockes Once circleCI passes, I think this is good enough. The one failing test on travis is not due to our code. WDYT? I want your opinion.

unfortunately I don't think I can look at this in detail before Monday. But I don't think making this release is a huge emergency.

I would recommend not to make the release if some tests are failing

jeromedockes:
> Otherwise, since I will not push new documentation from this release, I might push to PyPI just like this, even if the tests are failing on the cloud CIs.

I would recommend not to push to PyPi with failing tests, whatever the failures are. This can wait until failures are fixed

kchawla-pi:
Cool.

PR #1271, opened 2016-09-11
[MRG] Fix: NaN problem in NiftiMapsMasker

KamalakerDadi:
- Add "ensure_finite" to ilearn.image.clean_img
- Deal automatically with NaNs in NiftiMapsMasker

Attempts to Fix #1249

PR #2125, opened 2019-08-27
Remove inplace modification in signal.clean

AlexandreAbraham:
PR #1271 introduces this regression where ensure_finite=True modifies the data inplace instead of copying it (nan_to_num does an implicit copy). I preferred to get back to nan_to_num instead of just doing a copy because I believe that relying on numpy functions is safer and more future proof.

Note: I have set float defaults as it done by default for the nan value.


GaelVaroquaux:
Good catch. It would need a unit test.

Maybe @kchawla-pi can add one and we merge?

kchawla-pi:
Curious: Why is replacing the original `signals` with a modified copy better than an in-place modification?

kchawla-pi:
We should add the problem faced due to the regression in the original comment, or file an issue elaborating this.

kchawla-pi:
@GaelVaroquaux I'm not sure I an add a unit test that will bring more value than existing test suite. The existing tests are passing, I think that may be sufficient confirmation for the change. Thoughts?

jeromedockes:
> @GaelVaroquaux I'm not sure I an add a unit test that will bring more value than existing test suite. The existing tests are passing, I think that may be sufficient confirmation for the change. Thoughts?

ideally there should be a test that passes on this branch and fails on master

kchawla-pi:
Yeah I have an idea that might work.
Not sure if it adds value to the test suite though.
Kshitij

kchawla-pi:
Okay, I have come up with a test that is currently failing on master and passing on this branch. Should I'll push it here and we can subject it to the usual review. I'll incorporate any feedback and follow up.
@AlexandreAbraham This is your PR, if you wish to take over at any point, let me know.

kchawla-pi:
This PR is based off a 3 year old Nilearn master. Hopefully it won't mess up the git history for this PR.

kchawla-pi:
In the current Nilearn 'master', there are 2 `nilearn.tests.test_signal.test_clean_frequencies tests` . Which is wrong, means the first test is never run. How did that happen. Will fix it here or maybe another PR.

kchawla-pi:
The test is ready for review.

kchawla-pi:
I think this is ready as well. Will do one final check of test failing on master and passing here, then merge tomorrow.

PR #2127, opened 2019-09-02
getting rid of a DeprecationWarning

pbellec:
I get the following warning:
```
/home/pbellec/git/nilearn/nilearn/datasets/func.py:503: VisibleDeprecationWarning: Reading unicode strings without specifying the encoding argument is deprecated. Set the encoding, use None for the system default.
  dtype=None)
```
This is caused by a call to `np.genfromtxt`. By specifying `encoding=None` as suggested the warning goes away. It does change the behaviour of the call, as the default for `encoding` is `bytes`.  I am waiting to see if using `None` breaks any of the tests.


jeromedockes:
thanks!
even better would be to specify the right encoding? e.g. `encoding="utf-8"`

kchawla-pi:
I believe this is already being addressed in PR #1947 .

pbellec:
Sorry I should have researched. This is indeed a duplicate.

PR #2131, opened 2019-09-11
MAINT: Future-compatible cmap reversal

larsoner:
Without this change I get warnings of the form:
```
/home/larsoner/python/nilearn/nilearn/plotting/cm.py:159: MatplotlibDeprecationWarning: 
The revcmap function was deprecated in Matplotlib 3.2 and will be removed two minor releases later. Use Colormap.reversed() instead.
  _cmaps_data[_cmapname_r] = _cm.revcmap(_cmapspec)
```
The suggested fix of `.reversed()` is not immediately usable because that wouldn't give us the data to put in `_cmaps_data`, so just take the (simple) mpl code to do the reversal knowing they're all `dict` and not `callable`.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -66,6 +66,8 @@ Fixes
 - :func:`nilearn.plotting.plot_matrix` providing labels=None, False, or an empty list now correctly disables labels.
 - :func:`nilearn.datasets.fetch_surf_nki_enhanced` is now downloading the correct
   left and right functional surface data for each subject
+- Colormap creation functions have been updated to avoid matplotlib deprecation warnings
+  about colormap reversal
 
 0.5.2
 =====
--- a/nilearn/plotting/cm.py
+++ b/nilearn/plotting/cm.py
@@ -38,16 +38,16 @@ def _pigtailed_cmap(cmap, swap_order=('green', 'red', 'blue')):
     orig_cdict = cmap._segmentdata.copy()
 
     cdict = dict()
-    cdict['green'] = [(0.5*(1-p), c1, c2)
-                        for (p, c1, c2) in reversed(orig_cdict[swap_order[0]])]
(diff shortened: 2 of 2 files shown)


jeromedockes:
thanks!
the travis failure is unrelated

jeromedockes:
how did you trigger the warning? running the tests?
and I guess this is starting from matplotlib 3.2?

larsoner:
Any `nilearn.plotting` import should do it if you are on latest `matplotilb` dev

larsoner:
(I based the 3.2 in the comments on the deprecation warning I got from MPL)

jeromedockes:
sorry for the delay @larsoner 
would you mind rebasing or merging master? that should fix the travis failure

larsoner:
My editor removed trailing whitespace (arguably a good thing), which caused CIs to flag PEP8 errors on those lines (arguably a good thing), so I'm going to fix the PEP8 errors in those vicinities (could be a good or bad thing). There are a ton in the file and I don't want to touch everything, so I'll try to stay localized.

Or if you prefer I can revert the whitespace changes and just keep the relevant ones for MPL deprecation.

jeromedockes:
> My editor removed trailing whitespace (arguably a good thing),

I like to configure mine so that it only does this on lines that I have
touched, to keep diffs clean

> which caused CIs to flag PEP8 errors on those lines (arguably a good thing), so I'm going to fix the PEP8 errors in those vicinities (could be a good or bad thing). There are a ton in the file and I don't want to touch everything, so I'll try to stay localized.

thanks for fixing them!

> Or if you prefer I can revert the whitespace changes and just keep the relevant ones for MPL deprecation.

Now that you have already fixed the PEP8 errors we should keep it that
way

kchawla-pi:
Add this under fixes in what's new.

larsoner:
Done

GaelVaroquaux:
LGTM. Merging.

PR #2141, opened 2019-09-25
[MRG] Replac conda with pip in CI setup

### LINKED ISSUE #2115: Handling Conda's unreliability
Conda, much as I like it as an end user, has been a source of difficulties recently. #2077 #2079 
They suddenly dropped support for older versions of packages, which prompted me to suggest upgrading our minimum supported versions of Scikit-learn, Scipy to 0.19. However, with each minor version bump they seemingly were dropping support for more and more packages (scipy 0.19 is no longer available in the main channel).

I asked them about it here, please read this: https://github.com/ContinuumIO/anaconda-issues/issues/11146#event-2509814992
This response, plus reading the linked blog post, indicates to me that conda might be fine as a package manager for end-users, but we cannot rely on it for running our CIs.

For a short term solution, we can either pin its version to 4.6 (currently we are doing that for Nistats).
Or we add the older channel back, as described in the issue link and the blog post.

For a longer term, I am thinking about switching to pip.
We can let CircleCI keep using conda - it always uses the latest version and is a good way to test our conda pipeline. For Travis etc, I propose we move to pip.

I would like to move quickly on this matter. @GaelVaroquaux

kchawla-pi:
Fixes #2115 

Conda has become unreliable for use with CIs. Replacing it with alternatives such as pip and apt seems a prudent long term solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,39 +16,40 @@ matrix:
   fast_finish: true
   allow_failures:
     # allow_failures keyed to skipping tests.
-    - env: SKIPT_TESTS="true"
+    - env: SKIP_TESTS="true"
   include:
     # without matplotlib
     - name: "Python 3.5 minimum package versions"
       python: "3.5"
       env: DISTRIB="conda" PYTHON_VERSION="3.5"
            NUMPY_VERSION="1.11" SCIPY_VERSION="0.19" PANDAS_VERSION="*"
-           SCIKIT_LEARN_VERSION="0.19" COVERAGE="true" JOBLIB_VERSION="*"
+           SCIKIT_LEARN_VERSION="0.19" COVERAGE="true" JOBLIB_VERSION="0.11"
            LXML_VERSION="*"
     - name: "Python 3.5 latest package versions"
       python: "3.5"
       env: DISTRIB="conda" PYTHON_VERSION="3.5"
            NUMPY_VERSION="*" SCIPY_VERSION="*" PANDAS_VERSION="*"
            SCIKIT_LEARN_VERSION="*" MATPLOTLIB_VERSION="*" COVERAGE="true"
-           JOBLIB_VERSION="0.11"
+           JOBLIB_VERSION=".11"
            LXML_VERSION="*"
     - name: "Python 3.6 latest package versions"
       python: "3.6"
       env: DISTRIB="conda" PYTHON_VERSION="3.6"
            NUMPY_VERSION="*" SCIPY_VERSION="*" PANDAS_VERSION="*"
            SCIKIT_LEARN_VERSION="*" MATPLOTLIB_VERSION="*" COVERAGE="true"
-           JOBLIB_VERSION="*"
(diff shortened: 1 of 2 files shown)


continuous_integration/install.sh
```diff
@@ -72,18 +75,14 @@ create_new_conda_env() {
...
+    conda create -n testenv python==$PYTHON_VERSION --yes
     source activate testenv
-    conda install pytest pytest-cov --yes
+    pip install $REQUIREMENTS
+    pip install pytest pytest-cov
 
     if [[ "$INSTALL_MKL" == "true" ]]; then
         # Make sure that MKL is used
-        conda install --quiet --yes mkl
-    elif [[ -z $CIRCLECI ]]; then
```
jeromedockes: why do we not need to uninstall mkl explicitly anymore?

jeromedockes:
thanks! apart from the small question above +1 for merge

continuous_integration/install.sh
kchawla-pi: The `install.sh` is not used in CircleCI.

kchawla-pi:
Thanks. I will unpin conda==4.6 next and then test these changes in AppveyorCI. Only then can we merge this.

kchawla-pi:
I just remembered conda 4.7 was glitching for python3.5 installation so maybe not unpin it then.

jeromedockes:
> I just remembered conda 4.7 was glitching for python3.5 installation so maybe not unpin it then.

I think you can leave it pinned for now

kchawla-pi:
That would make this PR very easy, but it will still leave conda in play, which kind of defeats the purpose of the PR. If we were gonna rely on a pinned conda, then we don't need pip anyway. I'll use TravisCIs automatic virtualenvs. Then we'll see.

kchawla-pi:
Appveyor manages its own conda installation as part of its images for Python 3.5 so we don't need to change that and CircleCI uses the latest packages so Between the 3 of them Conda is covered.

This is ready for a review.

kchawla-pi:
@jeromedockes Second review?

.travis.yml
```diff
@@ -16,33 +15,42 @@ matrix:
...
            JOBLIB_VERSION="0.11"
            LXML_VERSION="*"
-    - env: DISTRIB="conda" PYTHON_VERSION="3.6"
+    - name: "Python 3.6 latest package versions"
+      python: "3.6"
+      env: DISTRIB="travisci" PYTHON_VERSION="3.6"
            NUMPY_VERSION="*" SCIPY_VERSION="*" PANDAS_VERSION="*"
            SCIKIT_LEARN_VERSION="*" MATPLOTLIB_VERSION="*" COVERAGE="true"
-           JOBLIB_VERSION="*"
+           JOBLIB_VERSION="0.12"
```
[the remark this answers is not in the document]
kchawla-pi: `Memory` switches from keyword `cachedir` to `location` in this version. Making sure we get the deprecation warning.

Issue #1750, opened 2018-09-02
Spatial constraints with Parcellations


dPys:
Parcellations is a great tool!

I'm wondering about how it handles spatial constraints-- when I use method='ward' and include a nifti mask object to constrain the clustering, how can I prevent getting solutions with spatially non-contiguous clusters?

Thanks,
Derek;

GaelVaroquaux:
It should do it automatically. Doesn't it?

dPys:
![ward_clust_nil](https://user-images.githubusercontent.com/16432683/44958551-5e51f100-aea7-11e8-99eb-fa937c8cb8d1.png)

This is an example solution using ward, where I've used a mask to constrain the clustering. Note that the two clusters circled in red have the same label generated by Parcellations, despite their obvious non-contiguity.

Please let me know if I'm just missing something obvious here!
-Derek

GaelVaroquaux:
Did the mask that you used have several connect components? If you do something like this, the notion of connectivity in ward clustering is broken, and we cannot ensure that each cluster spans only one connected component. You need to apply parcellation separately per connected component.

dPys:
It does seem that the mask has several connected components indeed. See attached img. 
[triple_net_ICA_overlap_3_sig_bin.nii.gz](https://github.com/nilearn/nilearn/files/2343610/triple_net_ICA_overlap_3_sig_bin.nii.gz)

What would be the easiest way to isolate connected components in the mask and apply Parcellations to each? I should also mention that when I use spectral NCUT with this mask, it seems to take care of this problem automatically, so I'm guessing the connected components assumption is specific to Ward/hierarchical methods?

bthirion:
You may want to consider:
https://nilearn.github.io/modules/generated/nilearn.regions.connected_label_regions.html
Apply Ward to each resulting image.

And yes, the spatial constraints are specific to Ward.

dPys:
FYI: Working on the latter method. Will get back to you guys in ~ a week or so to let you know how it goes (if you don't mind keeping this issue listed as open)

Thanks for the help :)

KamalakerDadi:
Is there anyway that we could do an example out of this process ?

dPys:
Hi @KamalakerDadi , Sorry for the very delayed reply. I was just going back through my old posts and saw that you had replied to this.  Would be happy to work out an example for this, but would like your feedback first on the best approach.

Here's what I have so far:
```
from nilearn import plotting
[...]

PR #2143, opened 2019-09-29
[ENH] Prototype argument for user-defined connectivity structure in parcellations

dPys:
@GaelVaroquaux @KamalakerDadi 

In the present PR, I am simply making the Parcellations class more flexible to start.

This can then serve as a springboard for better addressing issue #1750
Example use case:
```python
import nibabel as nib
import numpy as np
import time
from nilearn.regions import Parcellations

# Clustering method
clust_type = 'ward'
# Mask file with multiple connected components
mask_file = '/Users/dap3463/Downloads/triple_net_ICA_overlap_3_sig_bin.nii.gz'
#BOLD image file
func_file = '/Users/dap3463/Downloads/002/002.nii.gz'

func_img = nib.load(func_file)

# Get local connectivity structure based on custom function that imposes spatial constraints
local_conn = make_local_connectivity_tcorr(func_file, clust_mask, thresh=0.5)

# Generate ward-based parcellation with spatial constraints according to manually-defined connectivity structure
clust_est = Parcellations(method=clust_type, n_parcels=100,  mask=clust_mask, connectivity=local_conn)

# Fit estimator to time-series
clust_est.fit(func_img)

# Get the connected label regions and viola
region_labels = connected_label_regions(clust_est.labels_img_)
```

### DIFF
--- a/nilearn/regions/parcellations.py
+++ b/nilearn/regions/parcellations.py
@@ -133,6 +133,16 @@ class Parcellations(MultiPCA):
     n_parcels: int, default=50
         Number of parcellations to divide the brain data into.
 
+    connectivity : array-like or callable, optional
+        Connectivity matrix. Defines for each sample the neighboring samples
+        following a given structure of the data. This can be a connectivity
+        matrix itself or a callable that transforms the data into a
+        connectivity matrix. Default is 'auto', i.e, the hierarchical
+        clustering algorithm is determined using `image.grid_to_graph` to
+        capture each voxel's connectivity to every other voxel. Note that this
+        option is only applicable if selected methods are Agglomerative
+        Clustering type, 'ward', 'complete', 'average'.
+
     random_state: int or RandomState
         Pseudo number generator state used for random sampling.
 
@@ -248,10 +258,10 @@ class Parcellations(MultiPCA):
     """
     VALID_METHODS = ['kmeans', 'ward', 'complete', 'average', 'rena']
 
-    def __init__(self, method, n_parcels=50,
+    def __init__(self, method, n_parcels=50, connectivity='auto',
                  random_state=0, mask=None, smoothing_fwhm=4.,
-                 standardize=False, detrend=False,
-                 low_pass=None, high_pass=None, t_r=None,
+                 standardize=False, detrend=False, low_pass=None,
+                 high_pass=None, t_r=None,
                  target_affine=None, target_shape=None,
                  mask_strategy='epi', mask_args=None,
                  scaling=False, n_iter=10,
@@ -262,6 +272,7 @@ class Parcellations(MultiPCA):
         self.n_parcels = n_parcels
         self.scaling = scaling
         self.n_iter = n_iter
+        self.connectivity = connectivity
 
         MultiPCA.__init__(self, n_components=200,
                           random_state=random_state,
@@ -349,9 +360,12 @@ class Parcellations(MultiPCA):
 
         else:
             mask_ = _safe_get_data(mask_img_).astype(np.bool)
-            shape = mask_.shape
-            connectivity = image.grid_to_graph(n_x=shape[0], n_y=shape[1],
-                                               n_z=shape[2], mask=mask_)
+            if self.connectivity is 'auto':
+                shape = mask_.shape
+                connectivity = image.grid_to_graph(n_x=shape[0], n_y=shape[1],
+                                                   n_z=shape[2], mask=mask_)
+            else:
+                connectivity = self.connectivity
 
             from sklearn.cluster import AgglomerativeClustering
 


KamalakerDadi:
Thanks @dPys I come to this soon. I hope I remember the context behind this.

dPys:
Hi @KamalakerDadi ,

Thanks for prioritizing this. The context was to provide a way to spatially constrain the optimization of agglomerative clustering in the case of non-connected components of a mask. When it came down to it, a user-defined connectivity matrix was necessary to accomplish this. 

But the option proposed here (an option to feed a user-defined connectivity matrix to Parcellations that overrides that which is obtained automatically via sklearn.image's grid_to_graph) is more general than this specific case, and could be useful for doing a whole host of applications.

Cheers,
@dPys

bthirion:
I think that the idea is fine, but can you elaborate a bit on the use case ?
I'm wondering whether it's a corner case, or a general pattern ?
Best,
Bertrand

bthirion:
You need to test this functionality.

dPys:
@bthirion -

The capability to manually specify a connectivity matrix for agglomerative-type Parcellations could be useful for a ton of applications. But the specific use-case that I have highlighted as an example above is an incremental upgrade to Parcellations that provides a framework for creating a solution to issue #1750 , where two disconnected components share the same label resulting from agglomerative-type parcellation. 

To elaborate on my example:
The goal is to avoid two clusters circled in red having the same label generated by Parcellations, despite their obvious non-contiguity:
image
As @GaelVaroquaux rightly mentioned, the notion of connectivity in ward clustering is broken when the clustering mask provided has >1 connected components (a _frequent_ scenario as other users have posted about previously). In this scenario, we cannot ensure that each cluster spans only one connected component. However, the Ward criterion can be adapted to work jointly in multiple connect components without merging them, so that we apply it separately to each connected component, getting a tree in each of them, and finding the cut with the specified amount of clusters that minimizes the error in variance. The only way to accomplish this, however, is to manually specify a connectivity matrix for agglomerative clustering, rather than rely on image.grid_to_graph (which is the default). We could accomplish this in at least two ways: 1) We could constructs a spatially constrained connectivity matrix to as an alternative to grid_to_graph such weights w_ij of the connectivity matrix W correspond to the spatial correlation between the mask-restricted FC maps generated from the time series from voxel i and voxel j. Connectivity is only calculated between a voxel and the 27 voxels in its 3D neighborhood (face touching and edge touching); 2) the same as the first approach, except that the weights w_ij of the connectivity matrix W correspond to the temporal correlation across the time-series, as opposed to spatial information.

To flesh out the functions for my use-case in greater detail, it would look like the following:
```
import nibabel as nib
import numpy as np
import time
from nilearn.regions import Parcellations

def indx_1dto3d(idx, sz):
    """
    Translate 1D vector coordinates to 3D matrix coordinates for a 3D matrix of size sz.

    Parameters
    ----------
    idx : array
        A 1D numpy coordinate vector.
    sz : array
        Shape of 3D matrix idx.

    Returns
    -------
    x : int
        x-coordinate of 3D matrix coordinates.
    y : int
        y-coordinate of 3D matrix coordinates.
    z : int
        z-coordinate of 3D matrix coordinates.

    References
    ----------
    .. Adapted from PyClusterROI
    """
    from scipy import divide, prod
    x = divide(idx, prod(sz[1:3]))
    y = divide(idx - x * prod(sz[1:3]), sz[2])
    z = idx - x * prod(sz[1:3]) - y * sz[2]
    return x, y, z


def indx_3dto1d(idx, sz):
    """
    Translate 3D matrix coordinates to 1D vector coordinates for a 3D matrix of size sz.

    Parameters
    ----------
    idx : array
        A 3D numpy array of matrix coordinates.
    sz : array
        Shape of 3D matrix idx.

    Returns
    -------
    idx1 : array
        A 1D numpy coordinate vector.

    References
    ----------
    .. Adapted from PyClusterROI
    """
    from scipy import prod
    if np.linalg.matrix_rank(idx) == 1:
        idx1 = idx[0] * prod(sz[1:3]) + idx[1] * sz[2] + idx[2]
    else:
        idx1 = idx[:, 0] * prod(sz[1:3]) + idx[:, 1] * sz[2] + idx[:, 2]
    return idx1


def make_local_connectivity_scorr(func_file, clust_mask, thresh):
    """
    Constructs a spatially constrained connectivity matrix from a fMRI dataset.
    The weights w_ij of the connectivity matrix W correspond to the
    spatial correlation between the whole brain FC maps generated from the
    time series from voxel i and voxel j. Connectivity is only calculated
    between a voxel and the 27 voxels in its 3D neighborhood
    (face touching and edge touching).

    Parameters
    ----------
    func_file : str
        File path to a 4D Nifti1Image containing fMRI data.
    clust_mask : str
        File path to a 3D NIFTI file containing a mask, which restricts the
        voxels used in the analysis.
    thresh : str
        Threshold value, correlation coefficients lower than this value
        will be removed from the matrix (set to zero).

    Returns
    -------
    W : Compressed Sparse Matrix
        A Scipy sparse matrix, with weights corresponding to the spatial correlation between the time series from voxel i and voxel j

    References
    ----------
    .. Adapted from PyClusterROI
    """
    from scipy.sparse import csc_matrix
    from scipy import prod
    from itertools import product

    neighbors = np.array(sorted(sorted(sorted([list(x) for x in list(set(product({-1, 0, 1}, repeat=3)))], key=lambda k: (k[0])), key=lambda k: (k[1])), key=lambda k: (k[2])))


    # Read in the mask
    msk = nib.load(clust_mask)
    msz = msk.shape

    # Convert the 3D mask array into a 1D vector
    mskdat = np.reshape(msk.get_data(), prod(msz))

    # Determine the 1D coordinates of the non-zero
    # elements of the mask
    iv = np.nonzero(mskdat)[0]

    # Read in the fmri data
    # NOTE the format of x,y,z axes and time dimension after reading
    # nb.load('x.nii.gz').shape -> (x,y,z,t)
    nim = nib.load(func_file)
    sz = nim.shape

    # Reshape fmri data to a num_voxels x num_timepoints array
    imdat = np.reshape(nim.get_data(), (prod(sz[:3]), sz[3]))

    # Mask the datset to only the in-mask voxels
    imdat = imdat[iv, :]
    imdat_sz = imdat.shape

    # Z-score fmri time courses, this makes calculation of the
    # correlation coefficient a simple matrix product
    imdat_s = np.tile(np.std(imdat, 1), (imdat_sz[1], 1)).T

    # Replace 0 with really large number to avoid div by zero
    imdat_s[imdat_s == 0] = 1000000
    imdat_m = np.tile(np.mean(imdat, 1), (imdat_sz[1], 1)).T
    imdat = (imdat - imdat_m) / imdat_s

    # Set values with no variance to zero
    imdat[imdat_s == 0] = 0
    imdat[np.isnan(imdat)] = 0

    # Remove voxels with zero variance, do this here
    # so that the mapping will be consistent across
    # subjects
    vndx = np.nonzero(np.var(imdat, 1) != 0)[0]
    iv = iv[vndx]

    m = len(iv)
    print(m, ' # of non-zero valued or non-zero variance voxels in the mask')

    # Construct a sparse matrix from the mask
    msk = csc_matrix((vndx + 1, (iv, np.zeros(m))), shape=(prod(msz), 1))

    sparse_i = []
    sparse_j = []
    sparse_w = [[]]

    for i in range(0, m):
        if i % 1000 == 0:
            print('voxel #', i)
        # Convert index into 3D and calculate neighbors
        ndx3d = indx_1dto3d(iv[i], sz[:-1]) + neighbors

        # Convert resulting 3D indices into 1D
        ndx1d = indx_3dto1d(ndx3d, sz[:-1])

        # Convert 1D indices into masked versions
        ondx1d = msk[ndx1d].todense()

        # Exclude indices not in the mask
        ndx1d = ndx1d[np.nonzero(ondx1d)[0]]
        ndx1d = ndx1d.flatten()
        ondx1d = np.array(ondx1d[np.nonzero(ondx1d)[0]])
        ondx1d = ondx1d.flatten() - 1

        # Keep track of the index corresponding to the "seed"
        nndx = np.nonzero(ndx1d == iv[i])[0]

        # Extract the time courses corresponding to the "seed"
        # and 3D neighborhood voxels
        tc = np.array(imdat[ondx1d.astype('int'), :])

        # Ensure that the "seed" has variance, if not just skip it
        if np.var(tc[nndx, :]) == 0:
            continue

        # Calculate functional connectivity maps for "seed"
        # and 3D neighborhood voxels
        fc = np.dot(tc, imdat.T) / (sz[3] - 1)

        R = np.corrcoef(fc)

        if np.linalg.matrix_rank(R) == 1:
            R = np.reshape(R, (1, 1))

        # Set nans to 0
        R[np.isnan(R)] = 0

        # Set values below thresh to 0
        R[R < thresh] = 0

        # Calculate the spatial correlation between FC maps
        if np.linalg.matrix_rank(R) == 0:
            R = np.reshape(R, (1, 1))

        # Keep track of the indices and the correlation weights
        # to construct sparse connectivity matrix
        sparse_i = np.append(sparse_i, ondx1d, 0)
        sparse_j = np.append(sparse_j, (ondx1d[nndx]) * np.ones(len(ondx1d)))
        sparse_w = np.append(sparse_w, R[nndx, :], 1)

    # Ensure that the weight vector is the correct shape
    sparse_w = np.reshape(sparse_w, prod(np.shape(sparse_w)))

    # Concatenate the i, j, and w_ij vectors
    outlist = sparse_i
    outlist = np.append(outlist, sparse_j)
    outlist = np.append(outlist, sparse_w)

    # Calculate the number of non-zero weights in the connectivity matrix
    n = len(outlist) / 3

    # Reshape the 1D vector read in from infile in to a 3xN array
    outlist = np.reshape(outlist, (3, int(n)))
    m = max(max(outlist[0, :]), max(outlist[1, :])) + 1

    # Make the sparse matrix, CSC format is supposedly efficient for matrix arithmetic
    W = csc_matrix((outlist[2, :], (outlist[0, :], outlist[1, :])), shape=(int(m), int(m)))

    return W


def make_local_connectivity_tcorr(func_file, clust_mask, thresh):
    """
    Constructs a spatially constrained connectivity matrix from a fMRI dataset.
    The weights w_ij of the connectivity matrix W correspond to the
    temporal correlation between the time series from voxel i and voxel j.
    Connectivity is only calculated between a voxel and the 27 voxels in its 3D
    neighborhood (face touching and edge touching).

    References
    ----------
    .. Adapted from PyClusterROI

    Parameters
    ----------
    func_file : str
        File path to a 4D Nifti1Image containing fMRI data.
    clust_mask : str
        File path to a 3D NIFTI file containing a mask, which restricts the
        voxels used in the analysis.
    thresh : str
        Threshold value, correlation coefficients lower than this value
        will be removed from the matrix (set to zero).

    Returns
    -------
    W : Compressed Sparse Matrix
        A Scipy sparse matrix, with weights corresponding to the temporal correlation between the time series from voxel i and voxel j
    """
    from scipy.sparse import csc_matrix
    from scipy import prod
    from itertools import product

    # Index array used to calculate 3D neigbors
    neighbors = np.array(sorted(sorted(sorted([list(x) for x in list(set(product({-1, 0, 1}, repeat=3)))], key=lambda k: (k[0])), key=lambda k: (k[1])), key=lambda k: (k[2])))

    # Read in the mask
    msk = nib.load(clust_mask)
    msz = np.shape(msk.get_data())
    msk_data = msk.get_data()

    # Convert the 3D mask array into a 1D vector
    mskdat = np.reshape(msk_data, prod(msz))

    # Determine the 1D coordinates of the non-zero elements of the mask
    iv = np.nonzero(mskdat)[0]
    m = len(iv)
    print("%s%s%s" % ('\nTotal non-zero voxels in the mask: ', m, '\n'))

    # Read in the fmri data
    # NOTE the format of x,y,z axes and time dimension after reading
    nim = nib.load(func_file)
    sz = nim.shape

    # Reshape fmri data to a num_voxels x num_timepoints array
    data = nim.get_data()
    imdat = np.reshape(data, (prod(sz[:3]), sz[3]))

    # Construct a sparse matrix from the mask
    msk = csc_matrix((list(range(1, m + 1)), (iv, np.zeros(m))), shape=(prod(sz[:-1]), 1))
    sparse_i = []
    sparse_j = []
    sparse_w = []

    negcount = 0

    # Loop over all of the voxels in the mask
    print('Voxels:')
    for i in range(0, m):
        if i % 1000 == 0:
            print(str(i))
        # Calculate the voxels that are in the 3D neighborhood of the center voxel
        ndx3d = indx_1dto3d(iv[i], sz[:-1]) + neighbors
        ndx1d = indx_3dto1d(ndx3d, sz[:-1])

        # Restrict the neigborhood using the mask
        ondx1d = msk[ndx1d].todense()
        ndx1d = ndx1d[np.nonzero(ondx1d)[0]]
        ndx1d = ndx1d.flatten()
        ondx1d = np.array(ondx1d[np.nonzero(ondx1d)[0]])
        ondx1d = ondx1d.flatten()

        # Determine the index of the seed voxel in the neighborhood
        nndx = np.nonzero(ndx1d == iv[i])[0]

        # Extract the timecourses for all of the voxels in the neighborhood
        tc = np.array(imdat[ndx1d.astype('int'), :])

        # Ensure that the "seed" has variance, if not just skip it
        if np.var(tc[nndx, :]) == 0:
            continue

        # Calculate the correlation between all of the voxel TCs
        R = np.corrcoef(tc)

        if np.linalg.matrix_rank(R) == 1:
            R = np.reshape(R, (1, 1))

        # Set nans to 0
        R[np.isnan(R)] = 0

        # Set values below thresh to 0
        R[R < thresh] = 0

        if np.linalg.matrix_rank(R) == 0:
            R = np.reshape(R, (1, 1))

        # Extract just the correlations with the seed TC
        R = R[nndx, :].flatten()

        # Set NaN values to 0
        negcount = negcount + sum(R < 0)

        # Determine the non-zero correlations (matrix weights) and add their indices and values to the list
        nzndx = np.nonzero(R)[0]
        if len(nzndx) > 0:
            sparse_i = np.append(sparse_i, ondx1d[nzndx] - 1, 0)
            sparse_j = np.append(sparse_j, (ondx1d[nndx] - 1) * np.ones(len(nzndx)))
            sparse_w = np.append(sparse_w, R[nzndx], 0)

    # Concatenate the i, j and w_ij into a single vector
    outlist = sparse_i
    outlist = np.append(outlist, sparse_j)
    outlist = np.append(outlist, sparse_w)

    # Calculate the number of non-zero weights in the connectivity matrix
    n = len(outlist) / 3

    # Reshape the 1D vector read in from infile in to a 3xN array
    outlist = np.reshape(outlist, (3, int(n)))
    m = max(max(outlist[0, :]), max(outlist[1, :])) + 1

    # Make the sparse matrix
    W = csc_matrix((outlist[2, :], (outlist[0, :], outlist[1, :])), shape=(int(m), int(m)))

    return W

# Clustering method
clust_type = 'ward'
# Mask file with multiple connected components
mask_file = '/Users/dap3463/Downloads/triple_net_ICA_overlap_3_sig_bin.nii.gz'
#BOLD image file
func_file = '/Users/dap3463/Downloads/002/002.nii.gz'

func_img = nib.load(func_file)

# Get local connectivity structure based on custom function that imposes spatial constraints
local_conn = make_local_connectivity_tcorr(func_file, clust_mask, thresh=0.5)

# Generate ward-based parcellation with spatial constraints according to manually-defined connectivity structure
clust_est = Parcellations(method=clust_type, n_parcels=100,  mask=clust_mask, connectivity=local_conn)

# Fit estimator to time-series
clust_est.fit(func_img)

# Get the connected label regions and viola
region_labels = connected_label_regions(clust_est.labels_img_)
```

Tests added to nilearn/regions/tests/test_parcellations.py could at least cover the scorr and tcorr cases using a mask with multiple connected components. But again, the addition of a `connectivity` option to the Parcellations class is more general than the use-case outlined here and, after integration, could be tested with a more elaborate suite of custom-connectivity matrices for various use-cases. The bottom line is that grid_to_graph simply does not provide enough flexibility to handle many clustering scenarios on its own.

Let me know what you think
@dPys

bthirion:
My understanding is that ward clustering completes the graph under the hood to connect it.
So I'm not sure that changing the input graph will solve the issue. Did I misunderstand something ?

dPys:
Hi @bthirion ,

Actually AgglomerativeClustering obeys the structure provided by the connectivity argument -- it will not complete the graph to connect non-neighbors if no edges exist in the pre-made connectivity matrix.

Here's the result of the above for comparison:
[triple_net_ICA_overlap_9_sig_bin_reor_RAS_nores2mm_ward_k250.nii.gz](https://github.com/nilearn/nilearn/files/3698661/triple_net_ICA_overlap_9_sig_bin_reor_RAS_nores2mm_ward_k250.nii.gz)
As you can see no disconnected components share the same label, despite this being clustered using Ward and a mask with multiple connected components.

@dPys

dPys:
Update on the status of this PR? This should be a fairly minor one to merge, but if there are any further critiques, I'd be curious to hear them.

emdupre:
> Update on the status of this PR?

I think you still need a test, here. It might also be worth updating an example in the documentation to show the effect of this additional parameter.

dPys:
> > Update on the status of this PR?
> 
> I think you still need a test, here. It might also be worth updating an example in the documentation to show the effect of this additional parameter.

Thanks @emdupre . That makes sense and I'll get to work on a test shortly!

I have to think about how I would do an example though without having to make additional PR'. The trouble is that revealing the power of this parameter involves using various API's for making custom voxel-wise connectivity matrices, since in the majority of cases grid_to_graph will do. In other words, this is somewhat of an advanced parameter, so I'm not sure how Nilearn would prefer to handle the documentation part. My gut says the best thing to do might just be to leave this as a hidden parameter rather than expose it for typical users. But I could do an example using make_connectivity_tcorr/_scorr as in the snippet above. Let me know what you think would be best.

@dPys

kchawla-pi:
> I'll get to work on a test shortly!

Thanks! We are moving away from nose, and to Pytest.

dPys:
Apologies for the delayed update-- I'm still working on writing tests for this and am trying to hammer out the edge cases. A couple important observations already:
1) Given a mask used to create a custom, spatially-constrained connectivity structure for Ward, n_clusters must always be > len(connected_regions(img, extract_type='connected_components', min_region_size=1)[1]). If n_clusters is _less_ then the total number of connected components in the mask, Ward can begin to include non-neighboring voxels within the same cluster (i.e. it ignores the spatial constraints specified by the custom connectivity matrix because it cannot fill every connected component before creating further splits in the tree).
2) Custom, spatially-constrained connectivity structure should not be used for average or complete clustering methods, as they can produce unstable results with a few clusters that grow very quickly. The sparser the custom connectivity matrix fed into Parcellations, the more average and complete linkages begin to resemble the more brittle single linkage, and leads to suboptimal solutions.
3) tcorr and scorr methods of producing a custom connectivity matrix should probably themselves be included as optional settings for the connectivity parameter of Parcellations, just as 'auto' defaults to grid_to_graph.

Curious to hear thoughts.
@dPys

dPys:
We could also add just add what sklearn uses for its Agglomeration methods, where it restricts the input for connectivity to `accept_sparse=['csr', 'coo', 'lil']`. Then also add a check that the array's shape matches that of image?

PR #2144, opened 2019-10-01
[WIP] Rename test to deduplicate name in test_signal

kchawla-pi:
Some how there are two tests in `test_signal.py` with the same name, which means one of the test is overridden by another during collection and never run. This can be fixed by giving the tests unique names.

The purpose of the second test & hence it's name needs to be determined & decided.
@jeromedockes


jeromedockes:
LGTM

PR #2155, opened 2019-10-07
MAINT: Updates for nibabel deprecation

larsoner:
This takes care of some of the deprecation warnings on latest `nibabel` of the form:
```
  File "/home/larsoner/python/nilearn/nilearn/image/resampling.py", line 448, in resample_img
    data = img.get_data()
  File "/home/larsoner/python/nibabel/nibabel/deprecator.py", line 161, in deprecated_func
    warnings.warn(message, warn_class, stacklevel=2)
DeprecationWarning: get_data() is deprecated in favor of get_fdata(), which has a more predictable return type. To obtain get_data() behavior going forward, use numpy.asanyarray(img.dataobj).

* deprecated from version: 3.0
* Will raise <class 'nibabel.deprecator.ExpiredDeprecationError'> as of version: 5.0
```
I only tackled the ones that I needed to get `plot_glass_brain` not to complain anymore. If this seems like the right approach (I just did what `nibabel` recommended) I can try to look for other instances using `git grep`.

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -40,7 +40,7 @@ def _safe_get_data(img, ensure_finite=False):
     # that's why we invoke a forced call to the garbage collector
     gc.collect()
 
-    data = img.get_data()
+    data = np.asanyarray(img.dataobj)
     if ensure_finite:
         non_finite_mask = np.logical_not(np.isfinite(data))
         if non_finite_mask.sum() > 0: # any non_finite_mask values?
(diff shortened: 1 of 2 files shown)


larsoner:
The failures look legitimate but I'm a bit out of depth on fixing them. Someone else feel free to take over if possible!

kchawla-pi:
Will look at this next week.

kchawla-pi:
The miniumum version of Nibabel we currently support is 2.02 . Does that version support `img.get_fdata()` or `img.dataobj` ? If not, then we cannot roll this out without bumping up supported versions.

larsoner:
`dataobj` should be there:

https://github.com/nipy/nibabel/blob/maint/2.0.x/nibabel/spatialimages.py#L389-L391

jeromedockes:
thanks! since getting the image data without casting to float is very useful, we
may want to define a function e.g. in `nilearn.image.get_data` to replace
nibabel images' get_data method. 

the errors we are seeing seem to be due to arrays that were in the nifti images
`_data_cache` rather than `_dataobj`, and to the fact that the `_MNI152Template`
object used as a default background image in many nilearn plotting functions
does not implement the full NiftiImage interface.

jeromedockes:
I can take over this one. I suggest to define `nilearn.image.get_data` by copy-pasting the code from nibabel, and replace all calls to `Nifti1Image.get_data` with it. @GaelVaroquaux does that sound ok?

jeromedockes:
note: here is the discussion for deprecating get_data

https://github.com/nipy/nibabel/wiki/BIAP8

jeromedockes:
since this involves changing a lot of lines I will do it in another pr:

https://github.com/nilearn/nilearn/pull/2172

larsoner:
Closing in favor of #2172

jeromedockes:
thanks @larsoner

PR #2028, opened 2019-04-18
[WIP] ENH: add connectome strength plot

glemaitre:
This PR adds the possibility to plot the strength of a connectome on a glass brain.
The strength is defined as the absolute sum of the edges at a node.

TODO:

- [x] add example for visualization
- [x] add tests

PR #2164, opened 2019-10-09
Release Nilearn 0.6.0 alpha

kchawla-pi:


kchawla-pi:
In light of issue #2170 I had delayed the release of Nilearn 0.6.0a . If this is not fixed by Monday, I'll do a local build and release it. If PR #2028 passes locally, I will merge that into master to include it into the alpha release.

jeromedockes:
I don't understand why is it urgent to do the release? we can wait until
the OSF problem is fixed and CI is green

PR #2167, opened 2019-10-10
do not fail if metadata cannot be updated for an image

jeromedockes:
In some cases the neurovault fetcher tries to update the metadata. at the moment this can fail if the user has write permission for the neurovault data directory, but not for the metadata file of a particular image for some weird reason. this pr skips these updates with a warning

### DIFF
--- a/nilearn/datasets/neurovault.py
+++ b/nilearn/datasets/neurovault.py
@@ -1661,8 +1661,10 @@ def _update_image(image_info, download_params):
             'image_{0}_metadata.json'.format(image_info['id']))
         _write_metadata(image_info, metadata_file_path)
     except OSError:
-        warnings.warn("could not update metadata for image {}".format(
-            image_info["id"]))
+        warnings.warn(
+            "could not update metadata for image {}, "
+            "most likely because you do not have write "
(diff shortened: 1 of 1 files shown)


kchawla-pi:
Add your opening comment into the error message. The current one does not explain what's really going on.

nilearn/datasets/neurovault.py
```diff
@@ -1651,19 +1651,23 @@ def _update_image(image_info, download_params):
...
+            image_info, download_params)
+        image_info, collection = _download_image_terms(
+            image_info, collection, download_params)
+        metadata_file_path = os.path.join(
+            os.path.dirname(image_info['absolute_path']),
+            'image_{0}_metadata.json'.format(image_info['id']))
+        _write_metadata(image_info, metadata_file_path)
+    except OSError:
+        warnings.warn("could not update metadata for image {}".format(
+            image_info["id"]))
```
kchawla-pi: It would be better to only include the error generating step in the try block.

jeromedockes: all statements except the third can fail because of write permissions

PR #2169, opened 2019-10-11
remove is_valid filter

### LINKED ISSUE #2168: remove is_valid from default neurovault filters?
by default the neurovault fetcher ignores images that have their `is_valid` metadata field cleared. I'm not sure what sets this, but some useful collections such as hcp (id 4337) seem to not be "valid".

should we remove this filter?

jeromedockes:
closes #2168


jeromedockes:
> I guess thatthe travis failure is unrelated.

indeed the appveyor failure is a problem with the development dataset fetcher, unrelated

PR #2172, opened 2019-10-14
[MRG] Add get data function

jeromedockes:
cope with nibabel breaking change: deprecating Nifti1Image.get_data


nilearn/_utils/niimg.py
```diff
@@ -14,6 +14,21 @@
 from .compat import _basestring
 
 
+def get_data(img):
+    # copy-pasted from https://github.com/nipy/nibabel/blob/de44a105c1267b07ef9e28f6c35b31f851d5a005/nibabel/dataobj_images.py#L204
+    # get_data is removed from nibabel because:
+    # see https://github.com/nipy/nibabel/wiki/BIAP8
+    caching = "fill"
+    if caching not in ('fill', 'unchanged'):
```
bthirion: ??

bthirion:
We'll have to propagate that into nistats, then.

jeromedockes:
> We'll have to propagate that into nistats, then.

yes unfortunately. I can do it.

jeromedockes:
done for nistats in https://github.com/nistats/nistats/pull/403

jeromedockes:
any more reviews on this?

examples/05_advanced/plot_haxby_mass_univariate.py
```diff
@@ -122,6 +122,8 @@
 
 # Use the fmri mean image as a surrogate of anatomical data
 from nilearn import image
+from nilearn.image import get_data
```
GaelVaroquaux: Maybe better to use image.get_data in this example, as image is already imported

nilearn/_utils/niimg.py
```diff
@@ -14,6 +14,17 @@
 from .compat import _basestring
 
 
+def get_data(img):
```
GaelVaroquaux: This is a public function, use in our examples. Hence it needs a docstring.

As we are exposing such a public function, we should add to it the nilearn goodies: working with filenames and globbing. The easiest way to do this is probably to insert a check_niimg in the beginning of the function

jeromedockes:
> We are adding a public function to nilearn. I do not mind, as it makes nibabel less of a moving target, but we need to go all the way. The following are needed before merging:
>
> - Add a complete docstring
> - Add the function to the function reference in the docs
> - Add check_niimg in the beginning of the function so that it exposes all the goodies of nilearn (filenames and globbing)

Good point. To avoid circular imports between `niimg.py` and
`niimg_conversions.py` I defined in `niimg.py` a private `_get_data` function
with no checks, used everywhere in `nilearn._utils`, and defined the public
`get_data` with input validation and a docstring in `image.py`

PR #2177, opened 2019-10-14
[WIP] Removed sub-example due to unfit for lasso dataset - unstable float values

kchawla-pi:
https://github.com/nilearn/nilearn/pull/2028#issuecomment-541675600

Removing the last example in `plot_sphere_based_connectome`, plot connectome using the Seitzman atlas.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -67,6 +67,10 @@ Changes
 - :func:`nilearn.plotting.view_connectome` now converts NaNs in the adjacency
   matrix to 0.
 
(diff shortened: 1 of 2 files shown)


jeromedockes:
you need to update the example's docstring to remove the mention of Seitzman atlas

jeromedockes:
to give a bit of context:
this example does the same thing 4 times on different atlases, which is redundant. the last atlas has many regions so the procedure is numerically unstable and fails for several subjects in the dataset. hence we are removing this fourth part

PR #2182, opened 2019-10-16
Restore Localizer fetchers interface for backwards compatibility

### LINKED ISSUE #2181: restore localizers interface
it seems @GaelVaroquaux 's comment on #2097 

https://github.com/nilearn/nilearn/pull/2097/files#r328283557

was not addressed and we overlooked this when merging the PR.
`fetch_localizer_button_task` and `fetch_localizer_calculation_task` should be restored to their previous interface before the release

@kchawla-pi

kchawla-pi:
Fixes #2181 

Making `fetch_localizer_button_task` & `fetch_localizer_calculation_task` backwards compatible.

 - TODO: fetch_localizer_calculation_task()
 - TOFIX: Example plot_dim_plotting.py is failing here & in pre PR 2097.

### DIFF
--- a/examples/01_plotting/plot_dim_plotting.py
+++ b/examples/01_plotting/plot_dim_plotting.py
@@ -21,9 +21,9 @@ from nilearn import datasets
 
 localizer_dataset = datasets.fetch_localizer_button_task()
 # Contrast map of motor task
-localizer_tmap_filename = localizer_dataset.tmaps[0]
+localizer_tmap_filename = localizer_dataset.tmap
 # Subject specific anatomical image
-localizer_anat_filename = localizer_dataset.anats[0]
+localizer_anat_filename = localizer_dataset.anat
 ###########################################################################
 # Plotting with enhancement of background image with dim=-.5
 # --------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


kchawla-pi:
@jeromedockes I didn't spot any obvious changes in `fetch_localizer_calculation_task`. Would you point me in what's different?

jeromedockes:
> @jeromedockes I didn't spot any obvious changes in `fetch_localizer_calculation_task`. Would you point me in what's different?

maybe I got confused and only the button task changed. both functions
should have the same interface: no `n_subjects` parameter, and return a
single subject (not a list of subjects).

kchawla-pi:
I am comparing `fetch_localizer_calculation_task` from `e7b118b7` and they both have n_subjects and list as return.

jeromedockes:
> I am comparing `fetch_localizer_calculation_task` from `e7b118b7` and they both have n_subjects and list as return.

ok, it's annoying that these two functions do not have an identical
interface but it was already like that then. the important thing is that
they stay the same as in the previous release.

kchawla-pi:
Yeah I was surprised to see that as well.

jeromedockes:
with the stable release the returned dicts have:

calculation: cmaps: <class 'list'>, ext_vars: <class 'numpy.recarray'>, description: <class 'bytes'>
button: description: <class 'bytes'>, tmap: <class 'str'>, anat: <class 'str'>

kchawla-pi:
> maybe they should both have:

That is a different PR however. I will open a new PR for that.
Anything else here for the backward compatibility?

PR #2186, opened 2019-10-17
FIX: marker size issue in plot_connectome #2185

### LINKED ISSUE #2185: Marker size array is not well displayed in plot_glass_brain
```python
In [36]: nilearn.__version__ 
Out[36]: '0.6.0a'
```
What were you trying to do?
Plot a connectome using ```plot_connectome``` function, giving the a numpy array as node_size.

Here's the snippet to replicate:
```coords``` and ```colors``` are obtained by some functions I've implemented in my libraries.

```python
from nilearn.plotting import plot_connectome
matrix = np.random.rand(90,90)
matrix = (matrix * matrix.T)/2
t_matrix = matrix * np.int_(matrix > 0.45) #Plot only some connections
f = plot_connectome(t_matrix, 
                        coords[:90], 
                        colors[:90], 
                        t_matrix.sum(0)*100, 
                        'magma_r',
                        display_mode='lyr',
                        edge_vmin=0.45,
                        edge_vmax=0.5,
                        figure=pl.figure(figsize=(25,15)),
                        )
```

What did you expect will happen?
fixed

What actually happened?
The node size of left and right images are different from those obtained in coronal view.
buggy

I've found the problem in ```displays.py```and fixed it by adding at line 417
Maybe we also need some control on array/list length.
```python
            if isinstance(marker_size, np.ndarray) or \
                isinstance(marker_size, list):
                marker_size = np.array(marker_size)[relevant_coords]
```
I can do the PR, if you are ok.

robbisg:
fixes #2185

### DIFF
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -416,7 +416,7 @@ class GlassBrainAxes(BaseAxes):
                 marker_color = marker_color[relevant_coords]
 
             if not isinstance(marker_size, numbers.Number):
-              marker_size = np.asarray(marker_size)[relevant_coords]
+                marker_size = np.asarray(marker_size)[relevant_coords]
 
         defaults = {'marker': 'o',
                     'zorder': 1000}
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -414,10 +414,9 @@ class GlassBrainAxes(BaseAxes):
             if not isinstance(marker_color, _utils.compat._basestring) and \
                     len(marker_color) != 1:
                 marker_color = marker_color[relevant_coords]
-            
+
             if not isinstance(marker_size, numbers.Number):
-                marker_size = np.asarray(marker_size)[relevant_coords]
-            
+              marker_size = np.asarray(marker_size)[relevant_coords]
 
         defaults = {'marker': 'o',
                     'zorder': 1000}
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -403,7 +403,7 @@ class GlassBrainAxes(BaseAxes):
                 if self.direction == 'r' and xc >= 0:
                     relevant_coords.append(cidx)
                 elif self.direction == 'l' and xc <= 0:
-                        relevant_coords.append(cidx)
+                    relevant_coords.append(cidx)
             xdata = xdata[relevant_coords]
             ydata = ydata[relevant_coords]
             # if marker_color is string for example 'red' or 'blue', then
@@ -414,6 +414,11 @@ class GlassBrainAxes(BaseAxes):
             if not isinstance(marker_color, _utils.compat._basestring) and \
                     len(marker_color) != 1:
                 marker_color = marker_color[relevant_coords]
+            
+            if isinstance(marker_size, np.ndarray) or \
+                isinstance(marker_size, list):
+                marker_size = marker_size[relevant_coords]
+            
 
         defaults = {'marker': 'o',
                     'zorder': 1000}
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -415,9 +415,8 @@ class GlassBrainAxes(BaseAxes):
                     len(marker_color) != 1:
                 marker_color = marker_color[relevant_coords]
             
-            if isinstance(marker_size, np.ndarray) or \
-                isinstance(marker_size, list):
-                marker_size = marker_size[relevant_coords]
+            if not isinstance(marker_size, numbers.Number):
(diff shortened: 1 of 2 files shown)


nilearn/plotting/displays.py
```diff
@@ -414,6 +414,11 @@ def _add_markers(self, marker_coords, marker_color, marker_size, **kwargs):
             if not isinstance(marker_color, _utils.compat._basestring) and \
                     len(marker_color) != 1:
                 marker_color = marker_color[relevant_coords]
+            
+            if isinstance(marker_size, np.ndarray) or \
+                isinstance(marker_size, list):
```
jeromedockes: you can break the line inside parentheses rather than using a `\`
or maybe `not isinstance(marker_size, numbers.Number)`? this way any sequence will go through

jeromedockes:
> LGTM. I know that it is hard if not impossible to test. You confirm that it solves #2185 ?

not really, the scattered points are a PathCollection in the matplotlib axis `collections`, and the `PathCollection` has a `get_sizes()` method

jeromedockes:
> > LGTM. I know that it is hard if not impossible to test. You confirm that it solves #2185 ?
> 
> not really, the scattered points are a PathCollection in the matplotlib axis `collections`, and the `PathCollection` has a `get_sizes()` method

although I agree it may not be worth the trouble

nilearn/plotting/displays.py
```diff
@@ -415,8 +415,7 @@ def _add_markers(self, marker_coords, marker_color, marker_size, **kwargs):
                     len(marker_color) != 1:
                 marker_color = marker_color[relevant_coords]
             
-            if isinstance(marker_size, np.ndarray) or \
-                isinstance(marker_size, list):
+            if not isinstance(marker_size, numbers.Number):
                 marker_size = marker_size[relevant_coords]
```
jeromedockes: In the fix you suggested in #2185 you made sure `marker_size` got converted to a numpy array when it is a list, don't we still need that?

jeromedockes:
> I edited as suggested!

thanks!

nilearn/plotting/displays.py
robbisg: Yes, I was thinking about that. In the docs of ```plot_connectome``` the ```marker_size``` is a defined as scalar or array_like, so I supposed that it is not a list.
BTW I can edit it quickly.

jeromedockes:
> Yes, I was thinking about that. In the docs of ```plot_connectome``` the ```marker_size``` is a defined as scalar or array_like, so I supposed that it is not a list.
> BTW I can edit it quickly. 

that would be great. a list would be considered acceptable for "array-like"

nilearn/plotting/displays.py
```diff
@@ -414,6 +414,10 @@ def _add_markers(self, marker_coords, marker_color, marker_size, **kwargs):
             if not isinstance(marker_color, _utils.compat._basestring) and \
                     len(marker_color) != 1:
                 marker_color = marker_color[relevant_coords]
+            
+            if not isinstance(marker_size, numbers.Number):
+                marker_size = np.array(marker_size)[relevant_coords]
```
jeromedockes:
```suggestion
                marker_size = np.asarray(marker_size)[relevant_coords]
```

robbisg:
It was a pleasure @jeromedockes !! Thank all of you for developing this package!!

kchawla-pi:
An entry in whats_new under Fixes section for 0.6.0a

kchawla-pi:
THis is almost complete, we need to correct the Flake8 stylistic erro. It is the last entry if you click on details for TravisCI. I am reproducing the log here for convenience:

```
Running flake8 on the diff in the range 510e724..25d969f (6 commit(s)):

--------------------------------------------------------------------------------

nilearn/plotting/displays.py:417:1: W293 blank line contains whitespace

nilearn/plotting/displays.py:420:1: W293 blank line contains whitespace
```
This means that the blank lines indicated here contain spaces or tab characters. These should be removed.

robbisg:
@kchawla-pi I hope now it's ok. If not, can you suggest me the edit? Thank you! ;)

kchawla-pi:
Hey @robbisg That wasn't what I had meant, but it's not a complicated change, though unfamiliar. I'll look at this tomorrow.

kchawla-pi:
Hey yeah, in the next set of change you did most of the formatting fix. There's only one remaining.

In nilearn/plotting/displays.py:419:15 (line 419, position 15), that line is not indented with spaces that are multiple of 4. Fix that to satisfy Flake8.

In your commit message, explain why the change was made, for example "Fixed improper indentation" . Git can already tell us which files were changed and what changes were made to the code.

Issue #2188, opened 2019-10-22
Better documenting `slice()` usage in image.index_img


emdupre:
**What version of Nilearn are you using?**
0.6.0a0

**What were you trying to do?**
Index an image using `nilearn.image.index_img` over a set integer range. Passing a slice with the colon operator (e.g., `index=0:10`) results in a `SyntaxError`. 

The correct usage is `index=slice(0,10)`, but I could not find this documented in `nilearn.image.index_img`.

It'd be great to improve the documentation on using `index_img` with slices of arrays !

bthirion:
Indeed PR welcome ;-)

kchawla-pi:
It occurs to me that we may consider making `index=0:10` a valid way as well, though it would introduce 2 ways of doing the same thing, this is standar Python syntax that I would've expected to work. Thoughts?

bthirion:
If nobody requests it, let's not do it...

GaelVaroquaux:
> It occurs to me that we may consider making index=0:10 a valid way as well, though it would introduce 2 ways of doing the same thing, this is standard Python syntax that I would've expected to work.

I do not believe that this is valid Python syntax. What makes you think
that it is? The ":" to denote slicing can only be used in the bracket
operator, and I believe without keyword argument.

jeromedockes:
indeed `index=0:10` is invalid syntax

emdupre:
Closed with #2189

Issue #2184, opened 2019-10-17
`load_example_image`


jeromedockes:
When I want to show how to plot, resample, or mask an image, I just want one
single functional image and I don't care which, nor about its metadata.

At the moment I have to use one of:

```python
from nilearn import datasets
datasets.fetch_localizer_button_task()["tmap"]
datasets.fetch_localizer_calculation_task()["cmaps"][0]
datasets.fetch_neurovault_motor_task()["images"][0]
datasets.fetch_neurovault_auditory_computation_task()["images"][0]
```

all these functions have long and similar but slightly different names, and
return dictionaries with different keys, and it's difficult to remember which
key is singular or plural or goes with which function or is a list or an image.

Would it be possible to have a function `load_example_functional_img() ->
Nifti1Image` returning not a dictionary but the path to a nifti file (possibly
distributed with nilearn to avoid network failures)? We could use it e.g. for
all the plotting examples, except an example that shows using the subject's
anatomy as `bg_img`.

bthirion:
I like the idea. 
The only issue I can see with this is that the image is not actually random: it corresponds to a given contrast, in a given subject... I'd like the name of the function to acknowledge that fact. 
I'm afraid that some fMRI beginners will be lost other wise. 
Maybe `load_sample_motor_activation_image` or so ?

jeromedockes:
> Maybe `load_sample_motor_activation_image` or so ? 

yes that sounds good

alexisthual:
> should the function be included in `nilearn.plotting` or somewhere else?

I guess this function would belong to `nilearn.datasets` rather than `nilearn.plotting` :slightly_smiling_face: 

> although I am not sure how to make it so that the image is distributed with `nilearn`.

In essence, I think you want to write a new function which wraps one or several existing functions like the ones mentioned by @jeromedockes in the first post of this issue. In the end, you don't have to care about how the image is distributed with `nilearn`, you mostly have to make already accessible images even more accessible :ok_hand:

ymzayek:
> I guess this function would belong to `nilearn.datasets` rather than `nilearn.plotting` slightly_smiling_face

Yes agreed
 
> In essence, I think you want to write a new function which wraps one or several existing functions like the ones mentioned by @jeromedockes in the first post of this issue. In the end, you don't have to care about how the image is distributed with `nilearn`, you mostly have to make already accessible images even more accessible ok_hand

[...]

PR #2189, opened 2019-10-23
[doc] improve documentation re: using index_img with slice()

emdupre:
Addresses #2188 

* Adds a description of using the `slice()` constructor to the `nilearn.image.index_img` docstring *Examples* section
* Adds an example of using the `slice()` constructor to index a 4D image in the `plot_3d_and_4d` examples image

Updated outputs:

* Rendered [example](https://5074-1235740-gh.circle-artifacts.com/0/home/circleci/project/doc/_build/html/auto_examples/plot_3d_and_4d_niimg.html#sphx-glr-auto-examples-plot-3d-and-4d-niimg-py)
* Rendered [docstring](https://5074-1235740-gh.circle-artifacts.com/0/home/circleci/project/doc/_build/html/modules/generated/nilearn.image.index_img.html#nilearn-image-index-img)

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -615,6 +615,10 @@ def index_img(imgs, index):
 
     >>> full_timeseries = datasets.fetch_development_fmri(n_subjects=1,
     ...                                                   verbose=0).func
+    <BLANKLINE>
+    Dataset created in .../development_fmri # doctest:+ELLIPSIS
+    <BLANKLINE>
+
     >>> print(load_img(full_timeseries).shape)
     (50, 59, 50, 168)
 
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -599,7 +599,7 @@ def index_img(imgs, index):
     First we concatenate two mni152 images to create a 4D-image::
 
      >>> from nilearn import datasets
-     >>> from nilearn.image import concat_imgs, index_img
+     >>> from nilearn.image import concat_imgs, index_img, load_img
      >>> joint_mni_image = concat_imgs([datasets.load_mni152_template(),
      ...                                datasets.load_mni152_template()])
      >>> print(joint_mni_image.shape)
@@ -610,6 +610,16 @@ def index_img(imgs, index):
      >>> single_mni_image = index_img(joint_mni_image, 1)
      >>> print(single_mni_image.shape)
      (91, 109, 91)
+
+    We can also select multiple frames using the `slice` constructor::
+
+    >>> full_timeseries = datasets.fetch_development_fmri(n_subjects=1).func
+    >>> print(load_img(full_timeseries).shape)
+    (50, 59, 50, 168)
+
+    >>> first_ten_frames = index_img(func, slice(0, 10))
+    >>> print(first_ten_frames.shape)
+    (50, 59, 50, 10)
     """
     imgs = check_niimg_4d(imgs)
     # duck-type for pandas arrays, and select the 'values' attr
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -599,7 +599,7 @@ def index_img(imgs, index):
     First we concatenate two mni152 images to create a 4D-image::
 
      >>> from nilearn import datasets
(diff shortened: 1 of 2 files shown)


emdupre:
Sorry @bthirion  -- I made a few changes since you reviewed and mucked up the GitHub UI for dismissing a stale review when re-requesting. If you don't mind looking at this again, I'd appreciate it !

GaelVaroquaux:
Total nitpick: in the example, could you use the default parameters for plotting in the lines that you added for the example. The reason is that the example renders like two cuts of the same image.

Also, it might be useful to add a small comment on indexing in Python: the last number is not included, hence slice(3, 5) selects the images number 3 and 4.

On the other hand, I am wondering if we want to make that example longer. It's one of the first example that people hit, and I wonder whether I would not keep it to the essential. Maybe the docstring is enough?

emdupre:
> On the other hand, I am wondering if we want to make that example longer. It's one of the first example that people hit, and I wonder whether I would not keep it to the essential. Maybe the docstring is enough?

I'm not strongly tied to it, but I do like having the example. Partially because it combines index_img and iter_img. But more so for the reason you point out:

> Also, it might be useful to add a small comment on indexing in Python: the last number is not included, hence slice(3, 5) selects the images number 3 and 4.

I think this is something I've seen trip up lots of new users, particularly if they're also new to Python more generally. Similarly, seeing that the component number 3 is the fourth, rather than the third component seems valuable to me.

But, if that's too much of a Python point (as opposed to more specifically a Nilearn point), happy to remove it as well ! If you _do_ want to keep it, I'm happy to add the comment :smile_cat:

GaelVaroquaux:
If you think that you'd be happy to teach those few lines in the first hours of a nilearn course, I'm happy keeping the example (with the suggested modifications).

nilearn/image/image.py
```diff
@@ -610,6 +610,21 @@ def index_img(imgs, index):
      >>> single_mni_image = index_img(joint_mni_image, 1)
      >>> print(single_mni_image.shape)
      (91, 109, 91)
+
+    We can also select multiple frames using the `slice` constructor::
+
+     >>> functional_scan = datasets.fetch_development_fmri(
+     ...    n_subjects=1, verbose=0).func # doctest: +ELLIPSIS
```
jeromedockes: it is better not to download whole datasets in the doctests, especially the less reliable ones. these example snippets are just meant to illustrate the function signature, you can use fake data or data that is distributed with nilearn -- e.g a list of mni template copies as above

kchawla-pi:
Heya @emdupre, any plans for this before the next beta release?

nilearn/image/image.py
emdupre: I found that a list of MNI templates was less clear on read-through. I'd love to illustrate this with a functional dataset, though it looks like we don't ship any. Do you have recommendations of a dataset you'd prefer over development_fmri ?

kchawla-pi:
> On the other hand, I am wondering if we want to make that example longer. It's one of the first example that people hit, and I wonder whether I would not keep it to the essential. Maybe the docstring is enough?

As a user who remembers being new, one of the first things that made me comfortable with Nilearn was being able to look (access) at images the way I expected them to be arranged. I would prefer @emdupre 's long version.

Clarifying Python's indexing at this early stage is a sound idea.

nilearn/image/image.py
jeromedockes: the previous example in the same docstring uses the template so I would stick with it. otherwise any image that does not need to be downloaded is ok. I agree that it would be nice to ship one example functional image for such use cases: https://github.com/nilearn/nilearn/issues/2184

jeromedockes:
> Clarifying Python's indexing at this early stage is a sound idea.

the role of nilearn is not to teach an introduction to python though

kchawla-pi:
> the role of nilearn is not to teach an introduction to python though

No but a one-line reminder at the beginning of our initial tutorial doesn't hurt and may very well help. A lot of audience for our intro to Nilearn workshops are also audience for intro to Python. A tiny helpful reminder to cement that new knowledge is the job of a good teacher/tutorial.

PR #2193, opened 2019-10-25
fix error when colorscale given boolean array

jeromedockes:


nilearn/plotting/tests/test_js_plotting_utils.py
```diff
@@ -62,6 +62,7 @@ def test_colorscale_no_threshold():
     assert colors['cmap'].N == 256
     assert (colors['norm'].vmax, colors['norm'].vmin) == (13, -13)
     assert colors['abs_threshold'] is None
+    colors = js_plotting_utils.colorscale(cmap, values > 0, .5)
```
bthirion: This does not really test the new feature, does it ?

jeromedockes:
> This does not really test the new feature, does it ?

It fails without the fix introduced in this PR, because `vmin = -vmax`
results in a TypeError if vmax is a numpy boolean

nilearn/plotting/js_plotting_utils.py
```diff
@@ -87,6 +87,7 @@ def colorscale(cmap, values, threshold=None, symmetric_cmap=True,
         vmin = 0
     if vmax is None:
         vmax = abs_values.max()
+    vmax = float(vmax)
```
kchawla-pi: A comment here, why we are typecasting this to float, so future maintainers will have an easier time understanding it.

kchawla-pi:
Did this need a whats_new?

PR #2195, opened 2019-10-28
[MRG] Region means are no longer casted to Int type for Int images in NiftLabelsMasker

### LINKED ISSUE #2135: NiftiLabelsMasker misbehaves for integer niftis
For niftis like the example attached to this issue, NiftiLabelsMasker() from the masker family does not seem to carry out expected behavhior. Using this nifti as argument for labels_img leads .transform() to produce only zero outputs. The reason appears to be that the internal mean() operation on the voxels of each atlas region may not work properly if the data nifti is integer typed. In the case of 0/1 integer values, this leads to <1 values in many atlas regions, and thus to only zeros in the transformed variable outputs.

[some_one_zero_nifti.gz](https://github.com/nilearn/nilearn/files/3628442/some_one_zero_nifti.gz)

kchawla-pi:
Fixes #2135


kchawla-pi:
I undid the fix and added the test to demo the test failing without a fix. Will do the fix once the tests here fail.

kchawla-pi:
@GaelVaroquaux  Ready for review.
I went with float64 (only) when data kind is int; float32 was leading to precision errors.

kchawla-pi:
@banilo Does this branch solve you problem?

doc/whats_new.rst
```diff
@@ -11,6 +11,8 @@ NEW
 Fixes
 -----
 
+- Region means are no longer zero or casted to Int type
```
jeromedockes: this is not clear

nilearn/regions/signal_extraction.py
```diff
@@ -101,8 +101,10 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
         labels_data[np.logical_not(mask_data)] = background_label
 
     data = _safe_get_data(imgs)
+    target_datatype = np.float64 if data.dtype.kind == 'i' else data.dtype
```
jeromedockes: this will fail if dtype is bool or unsigned int for example. instead make it float32 if the data is float32 and float64 otherwise

kchawla-pi: If data was not a float64 to begin with, do we need the precision of float64 in target signals?

doc/whats_new.rst
kchawla-pi: The full sentence is not clear or some specific part?

nilearn/regions/signal_extraction.py
jeromedockes: float32 -> float32
anything else -> float64

or we could also consider

float64 -> float64
anything else -> float32

but not the current behaviour which maintains the problem that motivated this PR for anything else than int

doc/whats_new.rst
jeromedockes: I would write something like

labelsmasker no longer truncates region means to their integral part when input images are of integer type

nilearn/regions/signal_extraction.py
kchawla-pi: > the current behaviour which maintains the problem 

If so, then the test I wrote is incorrect. Would you review that please?

kchawla-pi: I believe `dtype.kind='i'` covers unsigned ints.
I should add a test for bool as well.

GaelVaroquaux: +1 on @jeromedockes 's comment

kchawla-pi: Got it. I'll update this and the test.

jeromedockes:
your test is correct but only checks for signed ints. unsigned ints
result in `kind == "u"`. It seems that bools get cast to int when
creating Nifti1Images so an image created with a boolean array would
also result in `kind == "i"`

  ```
  from nilearn import datasets, image

  mni = datasets.load_mni152_brain_mask()
  u_mni = image.new_img_like(mni, mni.get_data().astype('uint'))
  assert u_mni.get_data().dtype.kind == 'u'
  b_mni = image.new_img_like(mni, mni.get_data().astype('bool'))
  assert b_mni.get_data().dtype.kind == 'i'
  ```

kchawla-pi:
Casting it to anything except float64 opens the door for precision errors in some cases.
So no bathwater and no baby.
Ready for final review @jeromedockes

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -410,3 +412,27 @@ def test__trim_maps():
...
+                         )
+def test_img_to_signals_labels_int_type(orig_dtype):
+    fake_fmri_data = np.random.RandomState(0).rand(10, 10, 10, 10, ) > 0.5
+    fake_affine = np.random.rand(4, 4)
+    fake_fmri_img_orig = nibabel.Nifti1Image(
+                                        fake_fmri_data.astype(orig_dtype),
+                                        fake_affine,
+                                        )
+    fake_fmri_img_int = new_img_like(fake_fmri_img_orig,
+                                     fake_fmri_data.astype(np.int))
```
jeromedockes: I think this got mixed up. make `fake_fmri_img_orig` always float, and the other one int, float32, and unsigned int (rename it to something else than `fake_fmri_img_int`

nilearn/regions/signal_extraction.py
```diff
@@ -116,9 +116,11 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
         labels_data[np.logical_not(mask_data)] = background_label
 
     data = _safe_get_data(imgs)
+    target_datatype = np.float64
```
kchawla-pi: Here's something I don't understand.
The tests pass for:
```python
target_datatype = np.float
target_datatype = np.float32
target_datatype = np.float64
```
They fail for:
```python
target_datatype = np.float32 if data.dtype.kind != 'f' else np.float64
```
Someone help me understand why @GaelVaroquaux @jeromedockes

kchawla-pi:
Ready for another @jeromedockes

nilearn/regions/signal_extraction.py
GaelVaroquaux: Can you point us out to the test failure?

GaelVaroquaux: By the way, I think that the actual line that I would like is:
```suggestion
    target_datatype = np.float32 if data.dtype.kind == np.float32 else np.float64
```
In other terms: let float32 pass through, and all the rest is cast to float64

nilearn/regions/signal_extraction.py
```diff
@@ -116,9 +116,10 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
         labels_data[np.logical_not(mask_data)] = background_label
 
     data = _safe_get_data(imgs)
+    target_datatype = np.float
```
GaelVaroquaux:
```suggestion
    target_datatype = np.float32 is data.dtype == np.float32 else np.float64
```
I would prefer the behavior above.

nilearn/regions/signal_extraction.py
kchawla-pi: I have done various combinations of this and when I do, one of the tests fails.
You can try it locally.
Maybe there is something incorrect about the tests?

nilearn/regions/signal_extraction.py
kchawla-pi: Like I said, this fails.

kchawla-pi:
@GaelVaroquaux @emdupre 
I have uploaded both variations of the target_dtype, as documented in the commit message. They both fail & you can examine them here. 
Do the tests need to be modified?
You two can confer and help me figure this out.

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -410,3 +412,27 @@ def test__trim_maps():
...
     mask_data[1, 1, 1] = 1  # reset, just in case.
     np.testing.assert_equal(np.asarray(list(range(4))), maps_i_indices)
+
+
+@pytest.mark.parametrize('target_dtype',
+                         (np.float, np.float32, np.float64, np.int, np.uint),
+                         )
+def test_img_to_signals_labels_int_type(target_dtype):
+    fake_fmri_data = np.random.RandomState(0).rand(10, 10, 10, 10) > 0.5
+    fake_affine = np.random.rand(4, 4).astype(np.float64)
```
jeromedockes: why a random affine? use `np.eye`. also, always use a fixed seed in tests, not `np.random.rand`

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -410,3 +412,27 @@ def test__trim_maps():
...
+                                     fake_fmri_data.astype(target_dtype))
+    fake_mask_data = np.ones((10, 10, 10), dtype=np.uint8)
+    fake_mask = nibabel.Nifti1Image(fake_mask_data, fake_affine)
+
+    masker = NiftiLabelsMasker(fake_mask)
+    masker.fit()
+    timeseries_int = masker.transform(fake_fmri_img_target_dtype)
+    timeseries_float = masker.transform(fake_fmri_img_orig)
+    assert np.sum(timeseries_int) != 0
+    assert np.array_equiv(timeseries_int, timeseries_float)
```
jeromedockes:
```suggestion
    assert np.allclose(timeseries_int, timeseries_float)
```

will fix the failing test. don't use exact comparisons for floating-point types

nilearn/regions/signal_extraction.py
```diff
@@ -116,9 +116,10 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
         labels_data[np.logical_not(mask_data)] = background_label
 
     data = _safe_get_data(imgs)
+    target_datatype = np.float64 if data.dtype == np.float64 else np.float32
```
jeromedockes:
```suggestion
    target_datatype = np.float32 if data.dtype == np.float32 else np.float64
```

please revert to what you had in the previous commit: float64 by default

nilearn/regions/tests/test_signal_extraction.py
kchawla-pi: Isn't that what `RandomState(0)` is doing?

FMI, what's better about np.eye instead of a random affine?

jeromedockes:
> Isn't that what `RandomState(0)` is doing?
>
yes but you are only using it to generate `fake_fmri_data`, you are not
using the RandomState in the second line that generates `fake_affine`

> FMI, what's better about np.eye instead of a random affine?

to avoid confusing the reader about what we are testing here: the affine
is irrelevant

kchawla-pi:
Won't simply setting it to `np.float` let numpy decide the level of precision necessary?

jeromedockes:
> Won't simply setting it to `np.float` let numpy decide the level of precision necessary?

what do you mean? `np.float` is just an alias for the built-in float
(which usually corresponds to a C double).

your last commit is fine, I think we can merge. the CI failure is due to
the development dataset fetcher, not your PR

PR #2196, opened 2019-10-28
[WIP] Updated missing whats_new entry for #2193

kchawla-pi:


kchawla-pi:
I don't think this is a user facing change. Since we don't document changes that affect devs in whats_new, I am closing this one for now.

PR #2198, opened 2019-10-29
Installation should fail on Python < 3.5

### LINKED ISSUE #2197: Pip installing Nilearn 0.6.0a0 in Py2 envs
Not sure why, installing Nilearn 06.0a0 in a Python2environment does not raise an error. This is strange, since the wheels I build this time are not universal and I removed Python 2 tags. Maybe something was overlooked? Will look at this.

kchawla-pi:
Fixes #2197

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -1,6 +1,13 @@
 0.6.0b
 ======
 
+.. warning::
+
+ | **Python2 and 3.4 are no longer supported. Pip will raise an error in these environments.**
+ | **Minimum supported version of Python is now 3.5 .**
+ | **We recommend upgrading to Python 3.6 .**
+
+
 NEW
 ---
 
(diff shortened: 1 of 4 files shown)


jeromedockes:
why do we need this?

kchawla-pi:
Because without this, Pip will not raise an error if someone tries to install Nilearn's next release in a Python2 or Python3.4 env.

kchawla-pi:
Also, building Nilearn will build binaries compatible with both python2 and 3.

jeromedockes:
> Also, building Nilearn will build binaries compatible with both python2 and 3.

Ok thanks

kchawla-pi:
Now that we have the extra CI jobs from Azure pipelines which are not being used, I'll set them up to test stuff like this.

GaelVaroquaux:
LGTM. Merging. Thank you!

+1 also for scheduling a beta soon.

PR #2212, opened 2019-11-14
[MRG] Non-iterable Image headers are not copied anymore

### LINKED ISSUE #2147: Can't plot regions with nilearn 0.5.1 (because new_img_like() errs with non-iterable Headers)
Referencing (private Repo): https://github.com/NeuroLang/NeuroLang/issues/182
I'll investigate & attempt to replicate the problem.

> After update Nilearn to the last version available (nilearn=0.5.2) I can't plot neurolang regions. Is not the case of nilearn=0.5.0 that works well.
> 
> The code in '[plot_neurolang_datalog_example.py'](https://github.com/NeuroLang/NeuroLang/blob/64eac3f0dee3fde9101166685e430d1134b20c08/examples/plot_neurolang_datalog_example.py) is enough to reproduce the problem.
> 
> ```
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> <ipython-input-12-bd2b51b68e5c> in <module>
>      19 )
>      20 for name, region in res.value:
> ---> 21     plotting.plot_roi(region.spatial_image(), title=name)
> 
> ~/miniconda3/envs/neurolang/lib/python3.7/site-packages/nilearn/plotting/img_plotting.py in plot_roi(roi_img, bg_img, cut_coords, output_file, display_mode, figure, axes, title, annotate, draw_cross, black_bg, threshold, alpha, cmap, dim, vmin, vmax, resampling_interpolation, **kwargs)
>     706         threshold=threshold, bg_vmin=bg_vmin, bg_vmax=bg_vmax,
>     707         resampling_interpolation=resampling_interpolation,
> --> 708         alpha=alpha, cmap=cmap, vmin=vmin, vmax=vmax, **kwargs)
>     709     return display
>     710 
> 
> ~/miniconda3/envs/neurolang/lib/python3.7/site-packages/nilearn/plotting/img_plotting.py in _plot_img_with_bg(img, bg_img, cut_coords, output_file, display_mode, colorbar, figure, axes, title, threshold, annotate, draw_cross, black_bg, vmin, vmax, bg_vmin, bg_vmax, interpolation, display_factory, cbar_vmin, cbar_vmax, brain_color, **kwargs)
>     153 
>     154     if img is not False and img is not None:
> --> 155         img = _utils.check_niimg_3d(img, dtype='auto')
>     156         data = _safe_get_data(img, ensure_finite=True)
>     157         affine = img.affine
> 
> ~/miniconda3/envs/neurolang/lib/python3.7/site-packages/nilearn/_utils/niimg_conversions.py in check_niimg_3d(niimg, dtype)
>     320     Its application is idempotent.
>     321     """
> --> 322     return check_niimg(niimg, ensure_ndim=3, dtype=dtype)
>     323 
>     324 
> 
> ~/miniconda3/envs/neurolang/lib/python3.7/site-packages/nilearn/_utils/niimg_conversions.py in check_niimg(niimg, ensure_ndim, atleast_4d, dtype, return_iterator, wildcards)
>     269 
>     270     # Otherwise, it should be a filename or a SpatialImage, we load it
> --> 271     niimg = load_niimg(niimg, dtype=dtype)
>     272 
>     273     if ensure_ndim == 3 and len(niimg.shape) == 4 and niimg.shape[3] == 1:
> 
> ~/miniconda3/envs/neurolang/lib/python3.7/site-packages/nilearn/_utils/niimg.py in load_niimg(niimg, dtype)
>     120         if niimg.header is not None:
>     121             niimg = new_img_like(niimg, niimg.get_data().astype(dtype),
> --> 122                                 niimg.affine, copy_header=True)
>     123             niimg.header.set_data_dtype(dtype)
>     124         else:
> 
> ~/miniconda3/envs/neurolang/lib/python3.7/site-packages/nilearn/image/image.py in new_img_like(ref_niimg, data, affine, copy_header)
>     637     if copy_header:
>     638         header = copy.deepcopy(ref_niimg.header)
> --> 639         if 'scl_slope' in header:
>     640             header['scl_slope'] = 0.
>     641         if 'scl_inter' in header:
> 
> TypeError: argument of type 'SpatialHeader' is not iterable
> ```

kchawla-pi:
Fixes #2147 

Highlighted by the Neurolang project (pvt repo), the headers of at least one of the Nibabel Image type is not iterable (`SpatialHeader` for `SpatialImage`). This is a strange interface inconsistency on Nibabel's part.
This causes the fix included in PR #1872 to fail in such an image type, since `new_niimg_like(copy_header=True)` will fail.
This PR fixes that by handling this error.

Test forthcoming.

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -127,14 +127,16 @@ def load_niimg(niimg, dtype=None):
     dtype = _get_target_dtype(_get_data(niimg).dtype, dtype)
 
     if dtype is not None:
-        # Copyheader and set dtype in header if header exists
-        if niimg.header is not None:
+        # Copy header and set dtype in header if header exists
+        try:  # Is header iterable
+            'something' in niimg.header
+        except TypeError:  # incompatible with new_img_like(copy_header=True).
             niimg = new_img_like(niimg, _get_data(niimg).astype(dtype),
-                                niimg.affine, copy_header=True)
-            niimg.header.set_data_dtype(dtype)        
+                                 niimg.affine)
         else:
             niimg = new_img_like(niimg, _get_data(niimg).astype(dtype),
-                                niimg.affine)
+                                 niimg.affine, copy_header=True)
+            niimg.header.set_data_dtype(dtype)
 
     return niimg
 
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -127,16 +127,14 @@ def load_niimg(niimg, dtype=None):
     dtype = _get_target_dtype(_get_data(niimg).dtype, dtype)
 
     if dtype is not None:
-        # Copy header and set dtype in header if header exists
-        try:  # Is header iterable
-            'something' in niimg.header
-        except TypeError:  # incompatible with new_img_like(copy_header=True).
+        # Copyheader and set dtype in header if header exists
+        if niimg.header is not None:
             niimg = new_img_like(niimg, _get_data(niimg).astype(dtype),
-                                 niimg.affine)
+                                niimg.affine, copy_header=True)
+            niimg.header.set_data_dtype(dtype)        
         else:
             niimg = new_img_like(niimg, _get_data(niimg).astype(dtype),
-                                 niimg.affine, copy_header=True)
-            niimg.header.set_data_dtype(dtype)
+                                niimg.affine)
 
     return niimg
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,15 @@ matrix:
     # allow_failures keyed to python 3.5 & skipping tests.
     - python: "3.5"
       env: DISTRIB="travisci" PYTHON_VERSION="3.5" FLAKE8_VERSION="*" SKIP_TESTS="true"
+    - python: "3.7"
+      env: DISTRIB="travisci" PIP_FLAGS="--pre" COVERAGE="true"
+           NUMPY_VERSION="*" SCIPY_VERSION="*" PANDAS_VERSION="*"
+           SCIKIT_LEARN_VERSION="*" JOBLIB_VERSION="*" LXML_VERSION="*"
   include:
     - name: "Python 3.5 minimum package versions without Matplotlib"
       python: "3.5"
-      env: DISTRIB="travisci" PYTHON_VERSION="3.5"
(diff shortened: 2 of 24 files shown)


nilearn/_utils/niimg.py
```diff
@@ -127,14 +127,16 @@ def load_niimg(niimg, dtype=None):
...
     if dtype is not None:
-        # Copyheader and set dtype in header if header exists
-        if niimg.header is not None:
+        # Copy header and set dtype in header if header exists
+        try:  # Is header iterable
+            'something' in niimg.header
+        except TypeError:  # incompatible with new_img_like(copy_header=True).
             niimg = new_img_like(niimg, _get_data(niimg).astype(dtype),
-                                niimg.affine, copy_header=True)
-            niimg.header.set_data_dtype(dtype)
```
GaelVaroquaux: That's a very hacky solution. From what I read here, it seems that we are no longer keeping the header under the new versions nibabel, aren't we? That would be a problem.

GaelVaroquaux:
I've been trying to suggest a better fix. However, I cannot reproduce the problem: the test in the PR passes on master.

It would be useful to have a test that fails on master.

kchawla-pi:
> the test in the PR passes on master.

Yeah that's weird. Earlier, it was failing on master. Will re-examine this.

nilearn/_utils/niimg.py
kchawla-pi: This comment references the code in the master branch, not the PR. Was that accidental or are you saying the `master` code` is hacky? It was merged in PR #1872

GaelVaroquaux: This was accidental: I just checked, and the code in master does copy the header when it's not none.

kchawla-pi:
The fix works but I need to re-evaluate how to test it from within Nilearn; it might necessitate reworking the fix as well. This requires some thought, after which I will resubmit for review.

kchawla-pi:
@GaelVaroquaux I have decided to redo this PR from scratch, and the fix in a different location.
For now I have added a test that fails on master, which should be visible in the these CI builds.

kchawla-pi:
For the fix, I am moving it to the location where the exception is raised. This was not only easier to test & closer to the source, so seemed the right way, but also was useful across all instances where the function is called & may err, not just the specific use case tat led to the issue's discovery.

kchawla-pi:
@GaelVaroquaux Ready for another review. The Neurolangers will especially appreciate this PR.

PR #2215, opened 2019-11-14
Fix: load_surf_mesh and vol_to_surf offset vertices coords by CRAS

colinlaney:
There is offset between Freesurfer coordinate space and orig T1 space named CRAS. The issue misleads [`vol_to_surf`](https://github.com/nilearn/nilearn/blob/4e1ccac93753ffb2eced2d7cbb20d91688d21fab/nilearn/surface/surface.py#L386) when it try to project voxels from volume (.nii file) in T1 space onto surface (.orig, .pial, .white, .sphere, or .inflated) extracted by Freesurfer from the same volume (or at least from the volume in the same T1 space):
we observe some shift between volume and surface
plot_surf_stat_map-before
instead of uniform distribution of voxels intensity near the cortex
plot_surf_stat_map-after
It is easily solved by CRAS offsetting vertices coords loaded by [`load_surf_mesh`](https://github.com/nilearn/nilearn/blob/4e1ccac93753ffb2eced2d7cbb20d91688d21fab/nilearn/surface/surface.py#L681). It seems this does [not apply to .gii](https://neurostars.org/t/resample-volume-and-surface-to-same-orientation/4069/4) files.

### DIFF
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -713,7 +713,8 @@ def load_surf_mesh(surf_mesh):
                 surf_mesh.endswith('white') or surf_mesh.endswith('sphere') or
                 surf_mesh.endswith('inflated')):
             coords, faces, header = fs.io.read_geometry(surf_mesh, read_metadata=True)
-            coords += header['cras']
+            if 'cras' in header:
+                coords += header['cras']
         elif surf_mesh.endswith('gii'):
             if LooseVersion(nibabel.__version__) >= LooseVersion('2.1.0'):
                 gifti_img = nibabel.load(surf_mesh)
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -712,7 +712,8 @@ def load_surf_mesh(surf_mesh):
         if (surf_mesh.endswith('orig') or surf_mesh.endswith('pial') or
                 surf_mesh.endswith('white') or surf_mesh.endswith('sphere') or
                 surf_mesh.endswith('inflated')):
-            coords, faces = fs.io.read_geometry(surf_mesh)
+            coords, faces, header = fs.io.read_geometry(surf_mesh, read_metadata=True)
+            coords += header['cras']
         elif surf_mesh.endswith('gii'):
             if LooseVersion(nibabel.__version__) >= LooseVersion('2.1.0'):
                 gifti_img = nibabel.load(surf_mesh)


bthirion:
Thx for proposing the fix.
Can this feature be documented, e.g; the freesurfer reference ?
Also, you need to add a test that works with the fixed version and fails with current master version, so that the behavior remains enforced in future changes of the library.

jeromedockes:
thanks @colinlaney this is very useful; please let us know if you have questions regarding the additions outlined by @bthirion above: adding a regression test and updating the documentation

colinlaney:
@bthirion, @jeromedockes, unfortunately FreeSurfer has [almost zero documentation](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MorphAndRecon#A4.0VolumeProcessing.3Atroubleshootingbadoutput) about CRAS or it is extremely hard to find out what the abbreviation means. Me and my colleagues have found [crumbs](https://github.com/freesurfer/freesurfer/blob/a5f6effadbf9319f77679ab9d71e589a9e80717c/utils/mrisutils.cpp#L2286) [of](https://github.com/freesurfer/freesurfer/blob/87374ce4586de33c89ca4963ee4f2a435c8b0c1f/attic/mri_parselabel/mri_parselabel.cpp#L180) [information](https://github.com/freesurfer/freesurfer/blob/8b5272498f73e00781814decff2ca81ae2c81076/mri_coreg/mri_coreg.cpp#L304) about CRAS in FreeSurfer's code comments. So, I can **provide our vision of the purpose of CRAS** only, but not direct references to original documentation.

Also, I have a doubt is the issue a subject of either `nilearn` or `nibabel`. I mean if `nibabel` loaded surface correctly, `nilearn` would not have the problem at all. Perhaps `nibabel` loads _raw_ coordinates of vertices _intensionally_ and leave any transformations up to user, providing components of affine matrix in the header of data (something like here `coords, faces, header = nibabel.freesurfer.io.read_geometry('surf/lh.pial', read_metadata=True)`). Then the transformation really have to be done by `nilearn` as a user of `nibabel`.

At last I would be grateful if you teach me how to refer to data files needed to the test like `surf/lh.pial`. In the [tests](https://github.com/nilearn/nilearn/tree/master/nilearn/tests) I found synthetic data but not empiric one. Should I commit data file into the [tests/data](https://github.com/nilearn/nilearn/tree/master/nilearn/tests/data)? It is not trivial to design sanity checks of such spatial problem at first glance.

jeromedockes:
@colinlaney do you think you could open an issue on the nibabel repo and link to this one?

colinlaney:
> @colinlaney do you think you could open an issue on the nibabel repo and link to this one?

For now, I don't think so,  [since](https://github.com/nipy/nibabel/blob/f84b6654dbc8bf9f0b300365930ea2c378503ce4/doc/source/coordinate_systems.rst#nibabel-always-uses-an-ras-output-space)
> Nibabel always uses RAS+ output coordinates, regardless of the preferred output coordinates of the underlying format.

Where RAS+ means right-anterior-posterior _of a subject and is centered to a scanner_'s magnet isocenter. In my terms
> the transformation really have to be done by `nilearn` as a user of `nibabel`.

However it is not for sure that vertex coordinates being loaded by `nibabel` are truly RAS+. The argument for the truly RAS+ [is](http://www.grahamwideman.com/gw/brain/fs/surfacefileformats.htm)

1. > Updated 2003-05-14: Note that the surface files always seems to use RAS (Right, Anterior, Superior) coordinate system.
2. And `nibabel` does nothing more than loading coordinates as are.

So I have to read documentation of `nibabel` in more detail and search relevant previous issues to decide on which side the bug is. My pull request was done to your repository earlier than to `nibabel`, because it resolved my local problem immediately.

jeromedockes:
@colinlaney any news on this? did you get more info from nibabel?

ymzayek:
I want to pick this up. 
Still unclear if issue should be fixed on our side or nibabel so maybe it is worth to open an issue on nibabel? On the one hand, it is an easy fix on our side but on the other perhaps it also concerns nibabel. 
However before doing that I am wondering where I can find example surface mesh data to test and verify the shift.

jeromedockes:
thanks for reviving this! I think it's definitely worth getting advice from the nibabel folks, so opening an issue there is probably a good idea

jeromedockes:
> @jeromedockes from the answer I got from nibabel it's apparent that other applications compensated for this themselves so I think we can go ahead with this fix on our side. Would you agree? 

Yes, I agree! thanks

ymzayek:
Closing here since picked up in linked PR.

ymzayek:
Ah ok sorry about that! So just to clarify, that goes also for an open PR that is picked up in a new PR like in this case? Like in the new PR I should've put `fixes #prnumber` and that will close it when merged?

jeromedockes:
aaah so sorry I got confused and didn't even realize this was actually a PR and not an issue! what I said does not apply, sorry for being stupid.

However, the problem with opening a new PR from a branch checked out directly from `main` is that we would lose the original commits from @colinlaney (Arseniy Bozhenko in the git log) -- their name would not appear in the git log, which we don't want because they should be credited for their work. To avoid that, I would recommend starting your branch from `colinlaney:fix/loar_surf_mesh_cras` . then you can
- open a new PR on `nilearn/main`, as you did, but this time the original commits will still be there
- or push directly to colinlaney:fix/loar_surf_mesh_cras; I think you should have the permissions to do so, and then work will continue in this PR

to check out @colinlaney 's branch locally you can add their fork (`git@github.com:colinlaney/nilearn.git`) as a remote and fetch it, or you may want to add `  fetch = +refs/pull/*/head:refs/remotes/upstream/pr/*` to the `[upsteam]` section of your `nilearn/.git/config` to make it easy to checkout any branch that opened a PR

jeromedockes:
and then `git checkout upstream/pr/2215 `

PR #2219, opened 2019-11-18
MAINT: sklearn Deprecations

larsoner:
On latest `sklearn` master takes care of warnings of the form:
```
E   FutureWarning: The sklearn.utils.testing module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.utils. Anything that cannot be imported from sklearn.utils is now part of the private API.
```
and
```
E   FutureWarning: The sklearn.datasets.base module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.datasets. Anything that cannot be imported from sklearn.datasets is now part of the private API.
```

### DIFF
--- a/nilearn/_utils/testing.py
+++ b/nilearn/_utils/testing.py
@@ -11,7 +11,7 @@ import warnings
 import gc
 
 import numpy as np
-from sklearn.utils.testing import assert_warns
+from numpy.testing import assert_warns  # noqa: F401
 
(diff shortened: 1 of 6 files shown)


bthirion:
LGTM, thx !

kchawla-pi:
Looks fine to me as well.

PR #2221, opened 2019-11-19
Add parameter "strategy" to NiftiLabelsMasker to choose function to reduce ROIs

### LINKED ISSUE #1926: Generalising img_to_signal_labels to reduce with funcs other than the mean?

I've been working with statistical images using NiftiLabelsMasker. For a personal need, I had to generalise the `img_to_signal_labels` to reduce ROIs to a scalar value using the max and the std, instead of the mean.
In essence, [this part of the function here had to change to](https://github.com/nilearn/nilearn/blob/master/nilearn/regions/signal_extraction.py#L107):

```python
for n, img in enumerate(np.rollaxis(data, -1)):
    signals[n] = np.asarray(measurement_func(img,
                                             labels=labels_data,
                                             index=labels))
```
where `measurement_func` is checked previously as:

```python
# Check func
acceptable_funcs = ('mean', 'minimum', 'maximum', 'standard_deviation',
                    'variance', 'median')
if func in acceptable_funcs:
    measurement_func = getattr(ndimage.measurements, func)
else:
    raise ValueError("func not in %s" % (acceptable_funcs,))
```

I was wondering if this would merit a PR at all, and that's why I opened the issue. And if there is interest, should the user also be allowed to pass his/her own functions, as long as they obey the same signature as `ndimage.measurements.mean(img, labels=labels_data, index=labels)`?

I can think of all sorts of applications. One of them would be using the max and the standard_deviation to look at spikes within ROIs, comparing ROIs in general, or for studying the effects of regressing out confounds, i.e. general Q.A.

dangom:
The keyword allows changing the extraction function to compute
other measures besides the mean.

Closes #1926

### DIFF
--- a/nilearn/input_data/nifti_labels_masker.py
+++ b/nilearn/input_data/nifti_labels_masker.py
@@ -148,8 +148,13 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
         self.memory_level = memory_level
         self.verbose = verbose
 
-        # Reducing function
-        self.strategy = strategy
+        available_reduction_strategies = ('mean', 'median',
+                                          'minimum', 'maximum',
+                                          'standard_deviation', 'variance')
+        if strategy in available_reduction_strategies:
+            self.strategy = strategy
+        else:
+            raise ValueError("strategy not in %s" % (available_reduction_strategies,))
 
         if resampling_target not in ("labels", "data", None):
             raise ValueError("invalid value for 'resampling_target' "
--- a/nilearn/input_data/nifti_labels_masker.py
+++ b/nilearn/input_data/nifti_labels_masker.py
@@ -148,13 +148,18 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
         self.memory_level = memory_level
         self.verbose = verbose
 
-        available_reduction_strategies = ('mean', 'median',
+        available_reduction_strategies = {'mean', 'median',
                                           'minimum', 'maximum',
-                                          'standard_deviation', 'variance')
-        if strategy in available_reduction_strategies:
-            self.strategy = strategy
-        else:
-            raise ValueError("strategy not in %s" % (available_reduction_strategies,))
+                                          'standard_deviation', 'variance'}
+
+        if strategy not in available_reduction_strategies:
+            raise ValueError(str.format(
+                "Invalid strategy '{}'. Valid strategies are {}.",
+                strategy,
+                available_reduction_strategies
+            ))
+
+        self.strategy = strategy
 
         if resampling_target not in ("labels", "data", None):
             raise ValueError("invalid value for 'resampling_target' "
--- a/nilearn/input_data/nifti_labels_masker.py
+++ b/nilearn/input_data/nifti_labels_masker.py
@@ -110,8 +110,8 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
 
     strategy: str
         The name of a valid function to reduce each ROI.
-        Must be one of: mean, median, mininum, maximum, standard_deviation, variance
-
+        Must be one of: sum, mean, median, mininum, maximum, standard_deviation,
+        variance.
 
     See also
     --------
@@ -148,7 +148,7 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
         self.memory_level = memory_level
         self.verbose = verbose
 
-        available_reduction_strategies = {'mean', 'median',
+        available_reduction_strategies = {'mean', 'median', 'sum',
                                           'minimum', 'maximum',
                                           'standard_deviation', 'variance'}
 
--- a/nilearn/input_data/nifti_labels_masker.py
+++ b/nilearn/input_data/nifti_labels_masker.py
@@ -19,16 +19,17 @@ class _ExtractionFunctor(object):
 
     func_name = 'nifti_labels_masker_extractor'
 
-    def __init__(self, _resampled_labels_img_, background_label):
+    def __init__(self, _resampled_labels_img_, background_label, strategy):
         self._resampled_labels_img_ = _resampled_labels_img_
         self.background_label = background_label
+        self.strategy = strategy
 
     def __call__(self, imgs):
         from ..regions import signal_extraction
 
         return signal_extraction.img_to_signals_labels(
             imgs, self._resampled_labels_img_,
-            background_label=self.background_label)
+            background_label=self.background_label, strategy=self.strategy)
 
 
 class NiftiLabelsMasker(BaseMasker, CacheMixin):
@@ -107,6 +108,11 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
     verbose: integer, optional
         Indicate the level of verbosity. By default, nothing is printed
 
+    strategy: str
+        The name of a valid function to reduce each ROI.
+        Must be one of: mean, median, mininum, maximum, standard_deviation, variance
+
+
     See also
     --------
     nilearn.input_data.NiftiMasker
@@ -118,7 +124,7 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
                  low_pass=None, high_pass=None, t_r=None, dtype=None,
                  resampling_target="data",
                  memory=Memory(cachedir=None, verbose=0), memory_level=1,
-                 verbose=0):
+                 verbose=0, strategy="mean"):
         self.labels_img = labels_img
         self.background_label = background_label
         self.mask_img = mask_img
@@ -141,6 +147,9 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
         self.memory = memory
         self.memory_level = memory_level
         self.verbose = verbose
+
+        # Reducing function
+        self.strategy = strategy
 
         if resampling_target not in ("labels", "data", None):
             raise ValueError("invalid value for 'resampling_target' "
@@ -264,7 +273,7 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
                 ignore=['verbose', 'memory', 'memory_level'])(
             # Images
             imgs, _ExtractionFunctor(self._resampled_labels_img_,
-                                     self.background_label),
+                                     self.background_label, self.strategy),
             # Pre-processing
             params,
             confounds=confounds,
--- a/nilearn/input_data/nifti_labels_masker.py
+++ b/nilearn/input_data/nifti_labels_masker.py
@@ -109,9 +109,9 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
         Indicate the level of verbosity. By default, nothing is printed
 
     strategy: str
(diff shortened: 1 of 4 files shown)


dangom:
currently WIP. Suggestions for improvements or modifications are welcome

kchawla-pi:
Since we are reducing ROIs, how about calling the paramwter `reduce_using` or `strategy`, and set the default arg to `mean`

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -19,16 +19,17 @@ class _ExtractionFunctor(object):
 
     func_name = 'nifti_labels_masker_extractor'
 
-    def __init__(self, _resampled_labels_img_, background_label):
+    def __init__(self, _resampled_labels_img_, background_label, func="mean"):
```
kchawla-pi: Do not set default arg in private functions. They should always explicitly receive the args from the calling function.
Only user facing public functions should have default args.

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -107,6 +108,11 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
     verbose: integer, optional
         Indicate the level of verbosity. By default, nothing is printed
 
+    func: str
```
kchawla-pi: I propose calling this `strategy`, `reduce_using`, `aggregate_using`,`summary_function` ... something along such lines. The variable name should immediately indicate what the basic purpose/value is without the need for full context.

nilearn/regions/signal_extraction.py
```diff
@@ -103,10 +116,11 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
     data = _safe_get_data(imgs)
     signals = np.ndarray((data.shape[-1], len(labels)), order=order,
                          dtype=data.dtype)
+
     for n, img in enumerate(np.rollaxis(data, -1)):
-        signals[n] = np.asarray(ndimage.measurements.mean(img,
-                                                          labels=labels_data,
-                                                          index=labels))
+        signals[n] = np.asarray(measurement_func(img,
```
kchawla-pi: `reduction_function` or `aggregation_function` instead of `measurement_func` since we are not directly measuring.
Or `reduction_strategy` if you decide to choose `strategy` s the parameter name.

nilearn/regions/signal_extraction.py
```diff
@@ -76,6 +80,15 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
     target_affine = imgs.affine
     target_shape = imgs.shape[:3]
 
+    # Check aggregating function
+    acceptable_funcs = ('mean', 'median',
```
kchawla-pi: Rename `acceptable_funcs` to `valid_aggregation_functions`
Then remove the superfluous  comment `# Check aggregating function`

nilearn/input_data/nifti_labels_masker.py
dangom: I like `strategy`. Even though it's not immediate what it means, it's easier to remember and to write correctly on the first attempt.

nilearn/regions/signal_extraction.py
dangom: Here I went for reduction_function, as we are now dealing with a callable and not a str.

nilearn/input_data/nifti_labels_masker.py
dangom: Thanks for the advice, I wasn't aware of that but it definitely makes sense. Fixed.

nilearn/regions/signal_extraction.py
dangom: Fixed

dangom:
> Also write a test mocking the calling of other aggregation function types, including invalid function that are expected to raise an error.

Will do.

dangom:
Added tests. Noticed that there is some duplication on checking the available strategies both in `NiftiLabelsMasker` and in `img_to_signals_labels`. Would there be a better way to handle this? I chose to duplicate because otherwise the `ValueError `for `NiftiLabelsMasker` would be delayed until `.fit` is called, whereas I think it makes more sense to raise on `__init__`.

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -125,6 +125,36 @@ def test_nifti_labels_masker_with_nans_and_infs():
...
+                          [-2., -1., 0., 1., 2]]])
+
+    labels_data = np.array([[[0, 0, 0, 0, 0],
+                             [1, 1, 1, 1, 1]]], dtype=np.int8)
+
+    img = nibabel.Nifti1Image(img_data, affine)
+    labels = nibabel.Nifti1Image(labels_data, affine)
+
+    strategies = ("mean", "median", "minimum", "maximum", "standard_deviation", "variance")
+    expected_results = (0, 0, -2, 2, np.sqrt(2), 2)
```
kchawla-pi: Not a criticism, I have started using dicts for such situations.

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -125,6 +125,36 @@ def test_nifti_labels_masker_with_nans_and_infs():
...
+
+    for strategy, expected_result in zip(strategies, expected_results):
```
kchawla-pi: I love your naming conventions!

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -125,6 +125,36 @@ def test_nifti_labels_masker_with_nans_and_infs():
...
+        masker = NiftiLabelsMasker(labels, strategy=strategy)
+        # Here passing [img] within a list because it's a 3D object.
+        result = masker.fit_transform([img]).squeeze()
+        assert result == expected_result
+
+    assert_raises(ValueError, NiftiLabelsMasker, labels,
+                  strategy="A Strategy that does not exist should raise.")
```
kchawla-pi: We are phasing in pytest in lieu of nose. 
Use `pytest.raises`.
I'm including the code here so you don't have to google. :~)
```python
with pytest.raises(ValueError, match='strategy not in'):
        NiftiLabelsMasker(labels, strategy="A Strategy that does not exist should raise.")
```

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -125,6 +125,36 @@ def test_nifti_labels_masker_with_nans_and_infs():
...
+
```
kchawla-pi: Flake8 will complain about more than 2 empty lines between function definitions.

nilearn/regions/signal_extraction.py
```diff
@@ -76,6 +80,15 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
     target_affine = imgs.affine
     target_shape = imgs.shape[:3]
 
+    # Check aggregating function
```
kchawla-pi: Superfluous comment, the code is self-explanatory.

nilearn/regions/signal_extraction.py
```diff
@@ -76,6 +80,15 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
...
+    available_reduction_strategies = ('mean', 'median',
+                                      'minimum', 'maximum',
+                                      'standard_deviation', 'variance')
+    if strategy in available_reduction_strategies:
+        reduction_function = getattr(ndimage.measurements, strategy)
+    else:
+        raise ValueError("strategy not in %s" % (available_reduction_strategies,))
```
kchawla-pi: Usie the`str.format()` style.

kchawla-pi:
Also add a whats_new entry for `0.6.0b1` under NEW

nilearn/input_data/tests/test_nifti_labels_masker.py
dangom: good suggestion. thanks!

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -142,6 +148,18 @@ def __init__(self, labels_img, background_label=0, mask_img=None,
...
+        available_reduction_strategies = {'mean', 'median',
+                                          'minimum', 'maximum',
+                                          'standard_deviation', 'variance'}
+
+        if strategy in available_reduction_strategies:
+            self.strategy = strategy
+        else:
+            raise ValueError(str.format("Strategy '{}' not in {}",
+                                        strategy,
+                                        available_reduction_strategies))
```
kchawla-pi:
```python
"Invalid strategy '{}'. Valid strategies are {}".format(strategy, available_reduction_strategies)
```

kchawla-pi:
The TravisCI failure is unrelated. I am working on fixing it.

nilearn/input_data/nifti_labels_masker.py
dangom: Fixed. Also refactored to remove the else clauses.

kchawla-pi:
Puling from master will fix the TravisCI's first build failure.
Do it after the CircleCI run has completed.

kchawla-pi:
There are some Falke8 failures to be addressed.
You can see them here by scrolling to the end:
https://travis-ci.org/nilearn/nilearn/jobs/615781782?utm_medium=notification&utm_source=github_status

I see at least 2, might be more.

kchawla-pi:
This is pretty much complete. Anyone has any other feedback? I will merge this soon.

jeromedockes:
> Noticed that there is some duplication on checking the available strategies both in `NiftiLabelsMasker` and in `img_to_signals_labels`. Would there be a better way to handle this? I chose to duplicate because otherwise the `ValueError `for `NiftiLabelsMasker` would be delayed until `.fit` is called, whereas I think it makes more sense to raise on `__init__`.

in theory to comply with scikit-learn conventions init should only store the parameters and any processing or checks should be deferred to fit, but in this case I see that the init is already validating the resampling target

jeromedockes:
why not allow passing a user-defined function for the strategy?

kchawla-pi:
> why not allow passing a user-defined function for the strategy?

This is interesting. Any examples what would those be?

dangom:
> why not allow passing a user-defined function for the strategy?

You'd need to roll your own function with a `(input, labels, index)` signature, which is not really straightforward (not like writing a simple throwaway lambda). Things would be easier if we'd abstract the labels and index away, and simply pass an ROI reducing function. Maybe something to consider for the future, if a compelling reason arises.

_Without changing the current implementation_, allowing users to pass their own function would 1. hinder discoverability of all the `scipy.ndimage.measurements` functions, which this PR provides for free and 2. require users to reason around the function signatures.

kchawla-pi:
> why not allow passing a user-defined function for the strategy?

We can make that a new PR, if we decide to do that.

jeromedockes:
>> why not allow passing a user-defined function for the strategy?
>
> We can make that a new PR, if we decide to do that.
>
sure

Issue #2218, opened 2019-11-17
Are there ways to make nilearn/_utils nifti image handling more memory efficient?


dPys:
_What version of Nilearn are you using?_
Version: 0.6.0a0
_What were you trying to do?_
Run Parcellations and NiftiLabelsMasker with multiple simultaneous threads (100's or more in tightly controlled nipype nodes, with more than ample memory allocated e.g. 8GB).
_What did you expect will happen?_
They run successfully
_What actually happened?_
Memory errors. I tried using joblib but that actually just made the errors more frequent. Disabling it altogether appeared to at least minimize issues.

@KamalakerDadi -- this issue might interest you especially

Here are a few examples of the intermittent error messages I get:
```
  File "/opt/conda/lib/python3.6/site-packages/nilearn/input_data/nifti_labels_masker.py", line 238, in transform_single_imgs
    imgs_ = _utils.check_niimg_4d(imgs)
  File "/opt/conda/lib/python3.6/site-packages/nilearn/_utils/niimg_conversions.py", line 351, in check_niimg_4d
    dtype=dtype)
  File "/opt/conda/lib/python3.6/site-packages/nilearn/_utils/niimg_conversions.py", line 261, in check_niimg
    niimg = load_niimg(niimg, dtype=dtype)
  File "/opt/conda/lib/python3.6/site-packages/nilearn/_utils/niimg.py", line 116, in load_niimg
    dtype = _get_target_dtype(niimg.get_data().dtype, dtype)
  File "/opt/conda/lib/python3.6/site-packages/nibabel/dataobj_images.py", line 203, in get_data
    data = np.asanyarray(self._dataobj)
  File "/opt/conda/lib/python3.6/site-packages/numpy/core/_asarray.py", line 138, in asanyarray
    return array(a, dtype, copy=False, order=order, subok=True)
  File "/opt/conda/lib/python3.6/site-packages/nibabel/arrayproxy.py", line 355, in __array__
    raw_data = self.get_unscaled()
  File "/opt/conda/lib/python3.6/site-packages/nibabel/arrayproxy.py", line 350, in get_unscaled
    mmap=self._mmap)
  File "/opt/conda/lib/python3.6/site-packages/nibabel/volumeutils.py", line 524, in array_from_file
    n_read = infile.readinto(data_bytes)
  File "/opt/conda/lib/python3.6/gzip.py", line 276, in read
    return self._buffer.read(size)
MemoryError
```
```
[...]

PR #2223, opened 2019-11-20
[DOC] Add note about decreasing memory usage

dPys:
...by decompressing nii's per issue #2218 at @GaelVaroquaux 's request.

### DIFF
--- a/doc/manipulating_images/input_output.rst
+++ b/doc/manipulating_images/input_output.rst
@@ -234,6 +234,16 @@ data, which we call Niimgs or Niimg-4D. Accepted input arguments are:
    If you provide a sequence of Nifti images, all of them must have the same
    affine.
 
+.. topic:: Decreasing memory used when loading Nifti images
+
+   When Nifti images are stored compressed (.nii.gz), loading them directly
+   consumes more memory. As a result, large 4D images may
+   raise "MemoryError", especially on smaller computers and when using Nilearn
+   routines that require intensive 4D matrix operations. One step to improve
+   the situation may be to gunzip the data onto disk as an initial step.
+   If multiple images are loaded into memory sequentially, another solution may
+   be to `uncache <https://nipy.org/nibabel/images_and_memory.html#using-uncache>`_ one before loading and performing operations on another.
+
 Text files: phenotype or behavior
 ----------------------------------
 
--- a/doc/manipulating_images/input_output.rst
+++ b/doc/manipulating_images/input_output.rst
@@ -240,7 +240,7 @@ data, which we call Niimgs or Niimg-4D. Accepted input arguments are:
    consumes more memory. As a result, large 4D images may
    raise "MemoryError", especially on smaller computers and when using Nilearn
    routines that require intensive 4D matrix operations. One step to improve
-   the situation may be to gunzip the data onto disk as an initial step.
+   the situation may be to decompress the data onto disk as an initial step.
    If multiple images are loaded into memory sequentially, another solution may
    be to `uncache <https://nipy.org/nibabel/images_and_memory.html#using-uncache>`_ one before loading and performing operations on another.
 


GaelVaroquaux:
LGTM. Thank you very much. It's a pity that you skipped CI, as we would have looked at the docs rendered by the CI.

Another review by someone else?

kchawla-pi:
I wouldn't wanna merge this without looking at circleci, maybe I can start
it manually once I have a stable net connection

kchawla-pi:
I can't force build it directly, however there is a workaround.
Squash merging this PR into the `upstream test-circleci` branch and pushing it will build it.

kchawla-pi:
Thanks for the force push @dPys

dPys:
Already one step ahead of you @kchawla-pi !

doc/manipulating_images/input_output.rst
```diff
@@ -234,6 +234,16 @@ data, which we call Niimgs or Niimg-4D. Accepted input arguments are:
    If you provide a sequence of Nifti images, all of them must have the same
    affine.
 
+.. topic:: Decreasing memory used when loading Nifti images
+
+   When Nifti images are stored compressed (.nii.gz), loading them directly
+   consumes more memory. As a result, large 4D images may
+   raise "MemoryError", especially on smaller computers and when using Nilearn
+   routines that require intensive 4D matrix operations. One step to improve
+   the situation may be to gunzip the data onto disk as an initial step.
```
emdupre: Since there are multiple software to achieve this outcome, I might be more general:

```suggestion
   the situation may be to decompress the data onto disk as an initial step.
```

doc/manipulating_images/input_output.rst
```diff
@@ -234,6 +234,16 @@ data, which we call Niimgs or Niimg-4D. Accepted input arguments are:
    If you provide a sequence of Nifti images, all of them must have the same
    affine.
 
+.. topic:: Decreasing memory used when loading Nifti images
+
+   When Nifti images are stored compressed (.nii.gz), loading them directly
+   consumes more memory. As a result, large 4D images may
```
emdupre: Is it worth explaining why this is true ?

doc/manipulating_images/input_output.rst
dPys: I agree

doc/manipulating_images/input_output.rst
dPys: My sense is that it may not be worth getting too much into the weeds about why.  @GaelVaroquaux ?

dPys:
@GaelVaroquaux -- one final question from me on this:

why not run:
```
def make_uncompressed_temp(func_file):
    if func_file.endswith('.nii.gz'):
        import tempfile
        import gzip
        import os
        temp_path = "{}/{}.nii".format(tempfile.mkdtemp(), os.path.basename(func_file))
        decomp_func = gzip.open(path)
        with open(temp_path, 'wb') as outfile:
            outfile.write(decomp_func.read())
        decomp_func.close()
        outfile.close()
    else:
        temp_path = func_file
    return temp_path
```
every time as a default precursor to all Nilearn functions, subsequently deleting the temporary file after operations are performed? Wouldn't that reduce the memory burden universally?

GaelVaroquaux:
Many reasons why this would be fragile:

- It requires a system to clean up the file after, and these are
  challenging to make fool proof (think: crashing in the middle of
  reading)
- It will slow down the loading (though not much)
- It will take disk space, and sometimes there is none available (many
  processes loading in the same time), or mkdtemp might not return the 
  right place (shared systems not well configured)

However, what this gets to is the tradeoff between memory usage and disk
usage, which is a classic one. The right way to do this tradeoff would be
to increase the swap. Unfortunately, in linux the swap cannot be easily
increased dynamically.

doc/manipulating_images/input_output.rst
GaelVaroquaux: I agree. Less is more

GaelVaroquaux:
The docs can be viewed here: https://5013-1235740-gh.circle-artifacts.com/0/home/circleci/project/doc/_build/html/manipulating_images/input_output.html#niimg-like-objects

They look good. I am merging. I just realized that I'd like this title to be bold, but I'll fix it after merge.

Thanks a lot!

dPys:
> Many reasons why this would be fragile: - It requires a system to clean up the file after, and these are challenging to make fool proof (think: crashing in the middle of reading) - It will slow down the loading (though not much) - It will take disk space, and sometimes there is none available (many processes loading in the same time), or mkdtemp might not return the right place (shared systems not well configured) However, what this gets to is the tradeoff between memory usage and disk usage, which is a classic one. The right way to do this tradeoff would be to increase the swap. Unfortunately, in linux the swap cannot be easily increased dynamically.

Yep, well a useful hack anyway even if it doesn't scale :-) I did think about addressing this problem with swap but without being able to modify swappiness nor an interest in getting into the weeds with it, eventually just decided to try tmp. 

Anyway, glad to see this PR merged.

Cheers

PR #2224, opened 2019-11-20
CI: Test pre-release versions of dependencies

### LINKED ISSUE #1728: New CI test env with packages built from their latest master
In light of issue #1724 , should we add an 'Allowed to Fail' test to the CIs, where all the primary packages are built from the current master source? For example, in this issue, SKLearn master had a bug. I used `make` to build it from source, installed that in my local environment and then debugged the issue.

I  may create a local testing script to repeat this formyself.

Would it be worth while to clone masters of SKL, Numpy, SciPy, build from source & install them and then run tests in that environment?
It would catch such glitches earlier and help the package maintainers certainly, it may also add to our CI times.

What do you all think?

effigies:
Fixes #1728 
I noticed that you aren't testing pre-releases (I was checking to see if the nibabel RC was causing problems). I've found it very helpful to see when changes coming down the pipeline (particularly from numpy) are going to break compatibility.

This PR adds an optional `PIP_FLAGS` environment variable that will be added to `pip install` commands for requirements and nilearn.

Feel free to merge if this seems useful. I can also just use the Travis results for this PR to see if we're breaking anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,10 @@ matrix:
     # allow_failures keyed to python 3.5 & skipping tests.
     - python: "3.5"
       env: DISTRIB="travisci" PYTHON_VERSION="3.5" FLAKE8_VERSION="*" SKIP_TESTS="true"
+    - python: "3.7"
+      env: DISTRIB="travisci" PIP_FLAGS="--pre" COVERAGE="true"
+           NUMPY_VERSION="*" SCIPY_VERSION="*" PANDAS_VERSION="*"
+           SCIKIT_LEARN_VERSION="*" JOBLIB_VERSION="*" LXML_VERSION="*"
   include:
     - name: "Python 3.5 minimum package versions without Matplotlib"
       python: "3.5"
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
Absolutely, this is really useful. I wanted to suggest this on the the nibabel thread.

Maybe we should even consider testing nibabel master somewhere? I just wonder if it should live in nibabel or nilearn?

GaelVaroquaux:
+1 for merge once CI has passed.

effigies:
I think it's usually easier for downstream projects to test pre-releases of upstream, since you already have all the tools in place to test nilearn, and can also catch other tools' pre-releases.

Speaking of which, numpy/scipy and a number of other projects also push nightly wheels to http://8b8c47f30575e674b56d-47bd50c35cd79bd838daf386af554a83.r59.cf2.rackcdn.com/, if you want to test against even more bleeding-edge changes.

GaelVaroquaux:
> I think it's usually easier for downstream projects to test pre-releases of upstream, since you already have all the tools in place to test nilearn, and can also catch other tools' pre-releases.

But then changes in nibabel break _our_ CI, which I am not comfortable with, as in my view these changes are bugs in nibabel.

> Speaking of which, numpy/scipy and a number of other projects also push nightly wheels

We haven't really had difficulties with changes in these tools. Maybe scikit-learn, to be fair (which is ironical, as I am also core dev of scikit-learn)

kchawla-pi:
I had opened an issue for this and never implemented it. Thanks Chris!

kchawla-pi:
@effigies Thanks for this! 
> changes in nibabel break _our_ CI, which I am not comfortable with

Should this be "allowed to fail"?

Azure Pipelines(AZP)  has spare containers, moving this there would be better. If you don't wanna mess with AZP settings I can do it after this one is merged.

effigies:
Marked as an allowed failure. I'm definitely not feeling up to working on an Azure test right now.

PR #2240, opened 2019-12-03
[doc] Warning for non-matching n_parcels

### LINKED ISSUE #2238: Add warning when calculated n_parcels does not match requested n_parcels
In the `Parcellations` class, users can provide [a desired n_parcels](https://github.com/nilearn/nilearn/blob/5b7648cc4d452389f2d9320fafdfd1aa27562c0e/nilearn/regions/parcellations.py#L133) that they would like to be learned. I've noticed that sometimes &mdash; particularly, it seems, with k-means &mdash; the number of learned parcels does not exactly match the number of desired parcels. 

This can cause problems with downstream uses, such as in [fmralign](https://github.com/parietal-INRIA/fmralign). I'd like to add a check and to the `Parcellations` class that confirms that the learned number of parcels matches `n_parcels` and raises a warning if it does not.

WDYT ?

emdupre:
Closes #2238.

- Adds a warning when learned n_parcels does not match requested n_parcels


nilearn/regions/parcellations.py
```diff
@@ -366,6 +367,13 @@ def _raw_fit(self, data):
             self.connectivity_ = connectivity
         # Avoid 0 label
         labels = labels + 1
+
+        # Check that appropriate number of labels were created
+        if len(labels) != self.n_parcels:
+            n_parcels_warning = ('The number of generated labels does not '
```
kchawla-pi: ...labels do not match...

kchawla-pi:
Nice! A test & what's new entry & this should be ready.

nilearn/regions/parcellations.py
emdupre: I think "does" is correct, here. It's much more clear-cut if we drop the qualifier "of generated labels"
> The number ... does

kchawla-pi: I don't have an opinion about the "of generated labels", you decide.
With plural objects (number of, labels), do is the correct form. Does is used when the subject is singular. "The generated label does not match"

emdupre: I don't want this to devolve into a grammar conversation :laughing: 

"The number" is a singular noun was my point. But I'm going to resolve this and we can revisit if it's unclear down the line !

kchawla-pi: Number implies plurality. 
Okay,
We can do a grammar cage match next time we meet!

emdupre:
@GaelVaroquaux I was looking at adding another test, to confirm that the warning _isn't_ raised when the `n_parcels` number matches. Here's what I had:

```
@pytest.mark.filterwarnings('ignore::DeprecationWarning')
def test_parcellations_no_warnings():
    data = np.zeros((10, 11, 12, 5))
    img = nibabel.Nifti1Image(data, affine=np.eye(4))
    parcellator_n7 = Parcellations(method='kmeans', n_parcels=1)
    with pytest.warns(None) as record:
        parcellator_n7.fit(img)
    assert record.list is None
```

The problem is that `DeprecationWarning`'s _still_ aren't being filtered. It looks to be related to [this point](https://stackoverflow.com/a/58645998). 

Do you have any suggestions of how to filter appropriately and confirm there's no `UserWarning`, but other (deprecation) warnings are fine ?

GaelVaroquaux:
>     assert record.list is None

I would filter out the list manually here.

kchawla-pi:
Hi @emdupre 
What I suggest is to simply use the `match` parameter in the `pytest.warns(match=)` to specifically match the string from your warning to ensure other UserWarnings are not caught by mistake.

Also instead of doing what you did above, you can write the another test  like the one you have added in the code, but engineered such that the number of parcels are the same, and mark it as expected to fail, using `@pytest.mark.xfail`

GaelVaroquaux:
Once CI is green, I am +1 for merge.

emdupre:
Thanks, @kchawla-pi and @GaelVaroquaux ! 

Hopefully this will pass as-is, but the different methods are good to know moving forward :smile_cat:

kchawla-pi:
Totally. I particularly love `pytest.warn(match=' string')`. It is so much more compact & cleaner than the built in Python one.

kchawla-pi:
In my experience the manual parsing approach is rather fragile, and when implemented in a robust way, looks ugly. Do consider my suggestion, I think that approach is clean and robust.

kchawla-pi:
@emdupre I played around with Pytest to see if we can find ways to improve the testing. It seems to me my suggestion has gaps that pytest does not handle (counter-intuitive, hope they fix it), and your way is the better one.

PR #2246, opened 2019-12-04
More clear CONTRIBUTING.rst

GaelVaroquaux:
A little detail that was confusing  @koudyk


GaelVaroquaux:
Agreed with the why. I think that it's a separate discussion. Will merge once everything is green.

PR #2256, opened 2019-12-05
[MRG+1]: maint: warn upcoming deprecation for fetch_haxby_simple

### LINKED ISSUE #2230: Deprecation cycle for fetch_haxby_simple ?
**What version of Nilearn are you using?** 0.6.0b

The current code suggests that `fetch_haxby_simple` will [be removed in a future release](https://github.com/nilearn/nilearn/blob/39c798e0310999e02341452ada6387958c04f72d/nilearn/datasets/func.py#L23), but it's unclear when that deprecation cycle is expected to occur. 

It looks like this warning was added in 2016 in #1127. Since we seem to be shipping the haxby simple data in our [datasets](https://github.com/nilearn/nilearn/blob/master/nilearn/datasets/data/pymvpa-exampledata.tar.bz2), I'm unsure if this can be used in eg the `index_img` docstring (see #2189) or if it will be removed in the near future.

Any insight is appreciated !

emdupre:
Closes #2230 

- Sets deprecation cycle for `fetch_haxby_simple` to 0.7.x release
- Add  `VisibleDeprecationWarning` on function call

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -9,7 +9,7 @@ import numpy as np
 import numbers
 
 import nibabel
-from sklearn.utils import Bunch, deprecated
+from sklearn.utils import Bunch
 
 from .utils import (_get_dataset_dir, _fetch_files, _get_dataset_descr,
                     _read_md5_sum_file, _tree, _filter_columns, _fetch_file)
@@ -20,10 +20,11 @@ from .._utils.exceptions import VisibleDeprecationWarning
 from nilearn.image import get_data
 
 
-@deprecated("fetch_haxby_simple will be removed in the 0.7.x release. "
-            "Use 'fetch_haxby' instead.")
 def fetch_haxby_simple(data_dir=None, url=None, resume=True, verbose=1):
     """Download and load a simple example haxby dataset.
+
+    NOTE: This function is deprecated and will be removed in
+    0.7.0 release. Use `fetch_haxby` instead.
 
     Parameters
     ----------
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -20,7 +20,7 @@ from .._utils.exceptions import VisibleDeprecationWarning
 from nilearn.image import get_data
 
 
-@deprecated("fetch_haxby_simple will be removed in future releases. "
+@deprecated("fetch_haxby_simple will be removed in the 0.7.x release. "
             "Use 'fetch_haxby' instead.")
 def fetch_haxby_simple(data_dir=None, url=None, resume=True, verbose=1):
     """Download and load a simple example haxby dataset.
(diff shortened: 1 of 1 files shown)


emdupre:
Right now it's failing because there doesn't seem to be a test calling `fetch_haxby_simple` -- do we think it's worth it to add a test that the deprecation warning is raised, given that we're deprecating the function soon ?

GaelVaroquaux:
By "failing", you mean codecov? That's fine, I wouldn't worry about it.

kchawla-pi:
I don't understand, we need both the decorator and the code inside?
The decorator should be sufficient, based on https://scikit-learn.org/stable/modules/generated/sklearn.utils.deprecated.html

emdupre:
> I don't understand, we need both the decorator and the code inside?

The decorator is adding it to the docstring appropriately, but I'm not getting a visible deprecation warning when calling unless I add the `VisibleDeprecationWarning`

kchawla-pi:
I tried it in a standalone script and it seems to be working fine. Is there a warnings filter active?

emdupre:
Another option is I could remove the decorator and just add the deprecation `NOTE` to the docstring of `fetch_haxby_simple` to match what's in `fetch_haxby`

kchawla-pi:
It sucks that this isn't working properly, but if we can't find a solution then the current implementation is more convenient to remove when when we remove the function, so I suppose, keep it. Maybe add a comment why we need both.

kchawla-pi:
@emdupre I disabled the warning code inside the function, built your branch as a wheel, installed it in the environment, and import haxby_simple and I am seeing the decorator's warning.

kchawla-pi:
Well if you wanna look at it again today that's fine, else we can merge it tonight your time.

emdupre:
> @emdupre I disabled the warning code inside the function, built your branch as a wheel, installed it in the environment, and import haxby_simple and I am seeing the decorator's warning.

Thanks for checking that ! 

I might drop the decorator just for consistency within the code, and then we can easily grab all the deprecations from this file at the next release. I'll push that change now :smile_cat:

emdupre:
Actually, looking at that other docstring, it looks like we should also remove the `n_subjects` param from `fetch_haxby` in the next release :laughing: 

Should I update that deprecation warning in this PR, too ?

kchawla-pi:
Nah lets open a new one for that.

GaelVaroquaux:
@kchawla-pi : merge, then?

kchawla-pi:
> I might drop the decorator just for consistency within the code, and then we can easily grab all the deprecations from this file at the next release. I'll push that change now 😸

After this.

kchawla-pi:
oh I had misunderstood your earlier comment, this looks much better @emdupre

Issue #2242, opened 2019-12-04
Tutorial example for decoding 


ltetrel:
*What would you like changed/added and why?*

The idea is to update the [fMRI decoding notebook](https://nilearn.github.io/auto_examples/plot_decoding_tutorial.html#sphx-glr-auto-examples-plot-decoding-tutorial-py) and merge it with [MAIN19 material](https://github.com/main-training/main-training-nilearn-ml/blob/master/03_intro_to_nilearn.ipynb) 

*What would be the benefit? Does the change make something easier to use?*

A new material wlll be genrated (for workshops and training for ex.)

I am planning to work on this as a part of the [brainhack in montreal](https://brainhackmtl.github.io/global2019/projects.html#project8)

jeromedockes:
@ltetrel can we close this issue thanks to your PR #2260 ?

GaelVaroquaux:
Indeed, I forgot to close it. Thanks!!

PR #2260, opened 2019-12-06
typo + changes for plot_decoding_tutorial

ltetrel:
Updating plot_decoding_tutorial tutorial following issue #2242

### DIFF
--- a/examples/plot_decoding_tutorial.py
+++ b/examples/plot_decoding_tutorial.py
@@ -111,11 +111,10 @@ print(fmri_masked.shape)
 # number of non-negative voxels in our binary brain mask.
 #
 # .. seealso::
-# 	There are many other strategies in Nilearn `for masking data and for
-# 	generating masks 
-# 	<https://nilearn.github.io/manipulating_images/manipulating_images.html#computing-and-applying-spatial-masks>`_.
+# 	There are many other strategies in Nilearn :ref:`for masking data and for
+# 	generating masks <computing_and_applying_mask>`
 # 	I'd encourage you to spend some time exploring the documentation for these.
-# 	We can also `display this time series <https://nilearn.github.io/auto_examples/03_connectivity/plot_adhd_spheres.html#display-time-series>`_ to get an intuition of how the 
+# 	We can also `display this time series :ref:`sphx_glr_auto_examples_03_connectivity_plot_sphere_based_connectome.py`_ to get an intuition of how the 
 # 	whole brain signal is changing over time.
 #
 # We'll display the first three voxels by sub-selecting values from the 
--- a/examples/plot_decoding_tutorial.py
+++ b/examples/plot_decoding_tutorial.py
@@ -102,22 +102,22 @@ print(fmri_masked.shape)
 #
 # .. image:: /images/masking.jpg
 #
-# Essentially, we can think about overlaying a 3D grid on an image. Then, 
-# our mask tells us which cubes or "voxels" (like 3D pixels) to sample from. 
-# Since our Nifti images are 4D files, we can't overlay a single grid -- 
-# instead, we use a series of 3D grids (one for each volume in the 4D file), 
-# so we can get a measurement for each voxel at each timepoint. These are 
-# reflected in the shape of the matrix ! You can check this by checking the 
+# Essentially, we can think about overlaying a 3D grid on an image. Then,
+# our mask tells us which cubes or "voxels" (like 3D pixels) to sample from.
+# Since our Nifti images are 4D files, we can't overlay a single grid --
+# instead, we use a series of 3D grids (one for each volume in the 4D file),
+# so we can get a measurement for each voxel at each timepoint. These are
+# reflected in the shape of the matrix ! You can check this by checking the
 # number of non-negative voxels in our binary brain mask.
 #
 # .. seealso::
 # 	There are many other strategies in Nilearn :ref:`for masking data and for
 # 	generating masks <computing_and_applying_mask>`
 # 	I'd encourage you to spend some time exploring the documentation for these.
-# 	We can also `display this time series :ref:`sphx_glr_auto_examples_03_connectivity_plot_sphere_based_connectome.py`_ to get an intuition of how the 
+# 	We can also `display this time series :ref:`sphx_glr_auto_examples_03_connectivity_plot_sphere_based_connectome.py` to get an intuition of how the
 # 	whole brain signal is changing over time.
 #
-# We'll display the first three voxels by sub-selecting values from the 
+# We'll display the first three voxels by sub-selecting values from the
 # matrix. You can also find more information on how to slice arrays `here
 # <https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.indexing.html#basic-slicing-and-indexing>`_.
 import matplotlib.pyplot as plt
--- a/examples/plot_decoding_tutorial.py
+++ b/examples/plot_decoding_tutorial.py
@@ -48,7 +48,7 @@ print('First subject functional nifti images (4D) are at: %s' %
 #
 # Because fmri data is 4D (it consists of many 3D EPI images), we cannot 
 # plot it directly using :func:`nilearn.plotting.plot_epi` (which accepts 
-# just /D input). Here we are using :func:`nilearn.image.mean_img` to 
+# just 3D input). Here we are using :func:`nilearn.image.mean_img` to 
 # extract a single 3D EPI image from the fmri data.
 #
 from nilearn import plotting
@@ -64,7 +64,7 @@ plotting.view_img(mean_img(fmri_filename), threshold=None)
 # matrices, we will use the :class:`nilearn.input_data.NiftiMasker` to 
 # extract the fMRI data on a mask and convert it to data series.
 #
-# A mask of the Ventral Temporal streaming coming from the
+# A mask of the Ventral Temporal (VT) cortex coming from the
 # Haxby study is available:
 mask_filename = haxby_dataset.mask_vt[0]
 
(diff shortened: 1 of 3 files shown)


examples/plot_decoding_tutorial.py
```diff
@@ -65,8 +82,12 @@
 from nilearn.input_data import NiftiMasker
 masker = NiftiMasker(mask_img=mask_filename, standardize=True)
 
+###########################################################################
 # We give the masker a filename and retrieve a 2D array ready
 # for machine learning with scikit-learn
+# .. note::
+#	We can have a look at a report to see the mask computed by the 
+#	NiftiMasker by using :func:`masker.generate_report()``
```
GaelVaroquaux: I think that you should actually run this command in a cell just above.

examples/plot_decoding_tutorial.py
```diff
@@ -246,7 +306,7 @@
 print(coef_)
 
 ###########################################################################
-# It's a numpy array
+# There is one coefficient per voxel :
```
GaelVaroquaux: I would say: "It's a numpy array with only one coefficient per voxel:"

examples/plot_decoding_tutorial.py
ltetrel: I agree, actually I will change the explanation also. Seems not super clear.

examples/plot_decoding_tutorial.py
```diff
@@ -40,36 +40,55 @@
...
-# We will use the :class:`nilearn.input_data.NiftiMasker` to extract the
-# fMRI data on a mask and convert it to data series.
+# One way to visualize a fmri volume is using :func:`nilearn.plotting.plot_epi`.
+# We will visualize the previously fecthed fmri data from Haxby dataset.
 #
-# The mask is a mask of the Ventral Temporal streaming coming from the
-# Haxby study:
+# Because fmri data is 4D (it consists of many 3D EPI images), we cannot 
+# plot it directly using :func:`nilearn.plotting.plot_epi` (which accepts 
+# just /D input). Here we are using :func:`nilearn.image.mean_img` to
```
emdupre:
```suggestion
# just 3D input). Here we are using :func:`nilearn.image.mean_img` to 
```

examples/plot_decoding_tutorial.py
```diff
@@ -40,36 +40,55 @@
...
+###########################################################################
+# Feature extraction: from fMRI volumes to a data matrix
+# .......................................................
+#
+# These are some really lovely images, but for machine learning we need 
+# matrices to work with the actual data. To transform our Nifti images into
+# matrices, we will use the :class:`nilearn.input_data.NiftiMasker` to 
+# extract the fMRI data on a mask and convert it to data series.
+#
+# A mask of the Ventral Temporal streaming coming from the
```
emdupre:
```suggestion
# A mask of the Ventral Temporal (VT) cortex coming from the
```

examples/plot_decoding_tutorial.py
```diff
@@ -40,36 +40,55 @@
...
 from nilearn.input_data import NiftiMasker
 masker = NiftiMasker(mask_img=mask_filename, standardize=True)
-
-# We give the masker a filename and retrieve a 2D array ready
-# for machine learning with scikit-learn
 fmri_masked = masker.fit_transform(fmri_filename)
 
 ###########################################################################
+# .. seealso::
+# 	You can ask the NiftiMasker to fit his own mask given the data. In
```
emdupre:
```suggestion
# 	You can ask the NiftiMasker to derive a mask given the data. In
```

examples/plot_decoding_tutorial.py
```diff
@@ -40,36 +40,55 @@
...
+# 	this case, it is interresting to have a look at a report to see the
```
emdupre:
```suggestion
# 	this case, it is interesting to have a look at a report to see the
```

examples/plot_decoding_tutorial.py
```diff
@@ -40,36 +40,55 @@
...
+# 	computed mask by using :func:`masker.generate_report()`.
```
emdupre:
```suggestion
# 	computed mask by using :func:`masker.generate_report`.
```

examples/plot_decoding_tutorial.py
```diff
@@ -78,41 +97,76 @@
...
+#
+# .. image:: /images/masking.jpg
+#
+# Essentially, we can think about overlaying a 3D grid on an image. Then, 
+# our mask tells us which cubes or "voxels" (like 3D pixels) to sample from. 
+# Since our Nifti images are 4D files, we can't overlay a single grid -- 
+# instead, we use a series of 3D grids (one for each volume in the 4D file), 
+# so we can get a measurement for each voxel at each timepoint. These are 
+# reflected in the shape of the matrix ! You can check this by checking the 
+# number of positive voxels in our brain mask.
```
emdupre: This suggestion is a  really light one -- feel free to ignore: 

```suggestion
# number of non-negative voxels in our binary brain mask.
```

examples/plot_decoding_tutorial.py
```diff
@@ -78,41 +97,76 @@
...
+#
+# .. seealso::
+# 	There are many other strategies in Nilearn `for masking data and for
+# 	generating masks 
+# 	<https://nilearn.github.io/manipulating_images/manipulating_images.html#computing-and-applying-spatial-masks>`_.
+# 	I'd encourage you to spend some time exploring the documentation for these.
+# 	We can also `display this time series <https://nilearn.github.io/auto_examples/03_connectivity/plot_adhd_spheres.html#display-time-series>`_ to get an intuition of how the
```
emdupre: I thought we got rid of the ADHD dataset throughout the examples ?

examples/plot_decoding_tutorial.py
```diff
@@ -78,41 +97,76 @@
...
-behavioral = pd.read_csv(haxby_dataset.session_target[0], sep=" ")
+behavioral = pd.read_csv(haxby_dataset.session_target[0], delimiter=' ')
 print(behavioral)
 
 ###########################################################################
-# Retrieve the experimental conditions, that we are going to use as
-# prediction targets in the decoding
+# The task was a visual-recognition task, and the labels denote the 
+# experimental condition: the type of object that was presented to the 
+# subject. This is what we are going to try to predict
```
emdupre:
```suggestion
# subject. This is what we are going to try to predict.
```

examples/plot_decoding_tutorial.py
```diff
@@ -78,41 +97,76 @@
...
-# classification to the face vs cat discrimination
-fmri_masked = fmri_masked[condition_mask]
+# conditions. As a consequence the data is quite big:
+print(fmri_masked.shape)
 
 ###########################################################################
-# We now have less samples
+# Not all of this data has an interest to us for decoding, so we will keep
+# only fmri signals corresponding to faces or cats. We create a mask of
+# the samples belonging to the condition, this mask is then applied to the
```
emdupre:
```suggestion
# the samples belonging to the condition; this mask is then applied to the
```

examples/plot_decoding_tutorial.py
```diff
@@ -78,41 +97,76 @@
...
+# fmri data to restrict the classification to the face vs cat discrimination.
+# As a consequence, the input data is much less bigger (i.e. fmri signal is shorter):
```
emdupre:
```suggestion
# As a consequence, the input data is much smaller:
```

ltetrel: I thought it usefull to add a comment that the fmri signal is shorter to ensure that the reader is not loosing focus.
What do you think ?

examples/plot_decoding_tutorial.py
ltetrel: Cannot comment on this I don't know :P
@GaelVaroquaux

examples/plot_decoding_tutorial.py
emdupre: I think using "shorter" confuses me a bit in that we're not truncating the fMRI time series, but sub-selecting it. 

Can you think of another word, here ?

examples/plot_decoding_tutorial.py
emdupre: Went back through the `whats_new.rst` -- it was merged into another example and renamed as `plot_sphere_based_connectome.py` !

emdupre: There's still a `display-time-series` heading [here](https://github.com/nilearn/nilearn/blob/ab4db4a0a130061791947434df16b97b607b843d/examples/03_connectivity/plot_sphere_based_connectome.py#L81), so just updating the example name should work :smile_cat:

GaelVaroquaux:
The doc build fails because of the following problem:
/home/circleci/project/doc/building_blocks/manual_pipeline.rst:79:start-after pattern not found: # mask of the samples belonging to the condition.

You need to fix the reference in this rst file

GaelVaroquaux:
@ltetrel : I've just noticed that you are doing your changes on the master branch. This is a dangerous thing to do (it's easy to do mistakes and end up in challenging situations). I would strong advise working on a branch.

GaelVaroquaux:
@ltetrel : I've pushed a fix directly to your repository, with the hope of getting the tests to pass and of merging. If you work again on this, make sure that you pull first.

kchawla-pi:
We have merged several PRs into master. I'll wait for the master to complete and be green on CircleCI and then merge this one.

emdupre:
I think we still need to update the ADHD example reference to `plot_sphere_based_connectome.py` !

GaelVaroquaux:
Good catch, @emdupre . Doing it right now!

GaelVaroquaux:
By the way, this is why we should never use URLs but anchors: anchors would fail in the sphinx build.

My apologies for missing this.

kchawla-pi:
> By the way, this is why we should never use URLs but anchors: anchors would fail in the sphinx build.

If we had to link another example, how would we do that?

GaelVaroquaux:
All the examples define their own anchors.

I am busy doing this (and testing it locally). I'll push soon.

PR #2264, opened 2019-12-08
[DOC] Remove outdated references in plot_ica_resting_state

### LINKED ISSUE #2259: Strange comment in ICA example
This is the comment:
# XXX: must get the code to run for more than 1 subject
And this is the link to the example:
https://nilearn.github.io/auto_examples/05_advanced/plot_ica_resting_state.html

The comment appears to relate to an internal todo rather than comment/instruction to users.

emdupre:
Closes #2259.

- Removes internal note
- Removes outdated references to running on multiple subjects

### DIFF
--- a/examples/05_advanced/plot_ica_resting_state.py
+++ b/examples/05_advanced/plot_ica_resting_state.py
@@ -23,9 +23,7 @@ be given by a call to fit_transform).
 #####################################################################
 # Load movie watching dataset
 from nilearn import datasets
-# Here we use only single subject to get faster-running code. For better
-# results, simply increase this number
-# XXX: must get the code to run for more than 1 subject
+# Here we use only single subject to get faster-running code.
 dataset = datasets.fetch_development_fmri(n_subjects=1)
 func_filename = dataset.func[0]
 
@@ -44,10 +42,6 @@ from nilearn.input_data import NiftiMasker
 masker = NiftiMasker(smoothing_fwhm=8, memory='nilearn_cache', memory_level=1,
                      mask_strategy='epi', standardize=True)
 data_masked = masker.fit_transform(func_filename)
-
-# Concatenate all the subjects
-# fmri_data = np.concatenate(data_masked, axis=1)
-fmri_data = data_masked
 
 
 #####################################################################


kchawla-pi:
> Removes outdated references to running on multiple subjects

How is it outdated?

emdupre:
> How is it outdated?


This commented out section, and assigning to the unused variable `fmri_data` seem to reflect a previous state of the example !

```python
# Concatenate all the subjects
# fmri_data = np.concatenate(data_masked, axis=1)
fmri_data = data_masked
```

examples/05_advanced/plot_ica_resting_state.py
```diff
@@ -45,10 +43,6 @@
                      mask_strategy='epi', standardize=True)
 data_masked = masker.fit_transform(func_filename)
 
-# Concatenate all the subjects
-# fmri_data = np.concatenate(data_masked, axis=1)
```
kchawla-pi: To me it seems this is for when the users decide to try this with more than one subject as indicated in the original text.

examples/05_advanced/plot_ica_resting_state.py
```diff
@@ -45,10 +43,6 @@
...
-fmri_data = data_masked
-
```
kchawla-pi: And this is for maintaining compatibility with that?

examples/05_advanced/plot_ica_resting_state.py
emdupre: It wouldn't work as written. We still only take `func[0]`. We'd have to rework more of the example to allow for multiple subjects.

examples/05_advanced/plot_ica_resting_state.py
emdupre: `fmri_data` is not called after this line AFAICT. Not sure how it would maintain compatibility...

examples/05_advanced/plot_ica_resting_state.py
kchawla-pi: Okay, I misunderstood this. Thanks.

emdupre:
Thanks for the review, @kchawla-pi and @bthirion !

Should I open an issue about reworking this example to handle multiple subjects, as a future improvement ?

kchawla-pi:
I defer to you on this. 
I do think we should show how to do this with multiple subjects, if the code is gonna be different compared to one subject.
2 subjects should suffice.

PR #2269, opened 2019-12-10
Replace nose fixture with Pytest fixture for temp dirs during testing

### LINKED ISSUE #2268: Improve test fixtures for temporary paths
Required for removing nosetests.

A lot of tests in Nilearn tests do not clean up after themselves. They create files and directories which should be temporary but don't delete them afterwards.
This is at least in part because the temporary dirs & files are not programmed to clean up after themselves. This includes the nose fixtures which create temp paths for tests to use.
The nose fixture will be gone once we remove nose calls.

Considerations:
1. There should be one fixture that creates and tears down temporary directories and their contents.
2. It should remove references to file handles in memory before trying to delete the files, or tear down will fail on Windows.
3. It should be pytest supported.

### LINKED ISSUE #2275: [MAINT] Replace nose test temp dir fixture with pytest tmp_path fixture.
Stop the use of `nilearn.datasets.tests.test_utils.setup_tmpdata`, `nilearn.datasets.tests.test_utils.teardown_tmpdata` which relies of Nose's `with_setup` decorator fixture.
Instead, use a Pytest fixture to generate temporary directories for tests.

kchawla-pi:
Fixes #2275 

Step 3 for removing use of nose in testing


kchawla-pi:
Best to review this as it progresses lest it becomes overwhelming.
@GaelVaroquaux

kchawla-pi:
Ready for review. @GaelVaroquaux @jeromedockes 
I will start a new PR to remove the mocking fixture next.

nilearn/conftest.py
```diff
@@ -28,3 +30,25 @@ def pytest_collection_modifyitems(items):
...
+    ------
+
+    temp_path: string
+        Temporary directory path.
+    """
+    temp_path = str(tmp_path)
+    yield temp_path
+    local_objects = locals()
+    for local_object_ in local_objects:
+        del local_object_
```
jeromedockes: could you provide an example of a test that fails (on windows) without this line but passes with it? indeed it is a problem that has occurred a few times  in nilearn and nistats, I think due to memmapped nifti images

jeromedockes:
LGTM

kchawla-pi:
I can't test the fixture directly. 
 1. A pytest fixture cannot be called directly.
 2. The setup & teardown sections only run during a pytest run for a test function. The teardown has the cleanup.
There are ways to do this.
I will try to figure a way out. 
Also, there is a pytest plugin called pytester for this. I can do that.

For now I only checked manually that the files are being removed.

jeromedockes:
I did not mean to add a test for the feature. I just meant if you have
an example of a test that would crash on windows without this line in
the fixture, maybe copy it in this PR discussion for reference.
for example the last line `test_high_level_glm_with_paths` in nistats
could be removed using this fixture

kchawla-pi:
Ah okay. I suppose any test ending with a `del` statement will do.

kchawla-pi:
tmp_path is only available pytest v3.9 onwards. Conda by default installs Pytest v3.8 for Python 3.5 . 
Might switch to tmpdir instead.

jeromedockes:
> tmp_path is only available pytest v3.9 onwards. Conda by default installs Pytest v3.8 for Python 3.5 . 
> Might switch to tmpdir instead.

aren't we installing it with pip?

kchawla-pi:
I was considering it may create a problem for external contributors if they use conda to install everything, but now I am thinking it's a fairly minor case.
tmpdir is deprecated and I will be adding more work by using it for a very edge case.

I'll update Nilearn's README to reflect that for testing, the minimum version of Pytest should be 3.9

jeromedockes:
I think it's ok to say that you need a recent pytest to run the tests;
maybe add it to the developer's documentation.

kchawla-pi:
I just tested the fixture in nistats in `test_high_level_glm_with_data` and the removing open handles functionality doesn't work. 

We'll have to think of a better way to handle that, either using the monkeypath fixture or maybe a custom decorator kinda one I once wrote for Nistats (it wasn't merged).

Either way, it's beyond the scope of this PR. I iwll modify this PR's initial comment to reflect this and simply use pytest's `tmp_path` fixture for now.

jeromedockes:
> I just tested the fixture in nistats in `test_high_level_glm_with_data` and the removing open handles functionality doesn't work. 
>
> We'll have to think of a better way to handle that, either using the monkeypath fixture or maybe a custom decorator kinda one I once wrote for Nistats (it wasn't merged).
>
> Either way, it's beyond the scope of this PR. I iwll modify this PR's initial comment to reflect this and simply use pytest's `tmp_path` fixture for now.

thanks

Issue #2270, opened 2019-12-10
[META] Replace Nose tests completely with Pytests


kchawla-pi:
Replacing nose tests with Pytests:
Reasons: 
https://github.com/nilearn/nilearn/issues/1752#issuecomment-418805890
https://github.com/scikit-learn/scikit-learn/issues/7319

1. Switch to Pytest as test runner **(DONE in #2076  and #2130)** ✔️ 
2. Replace Nose asserts, warns, raises with Pytest equivalents  **(DONE in #2232)** ✔️ 
3. Replace Nose temp path fixtures with Pytest fixtures **(DONE in #2275)** ✔️ 
4. Replace Nose mock fixtures with Pytest mock fixtures **(DONE in #2271 )** ✔️ 
5. Replace references to nose dependent `assert_raises_regex` with `pytest.raises` **(DONE in #2278)** ✔️ 
6. Remove the last vestiges of nosetests code & doc references from the repo, including localizer_mock fixture **(✔️ in #2277)**

PR #2271, opened 2019-12-11
[MRG] Repalce nose with Pytest for dataset url requests mocking fixture

kchawla-pi:
Part of #2270 

I didn't improve the quality of the functions the fixture cals since it won't be used outside the datasets testing, and also it will require more reading to know what it is doing and how it works.

I will attend to it in a later PR, meanwhile this will get nose out of our system.


kchawla-pi:
Ready for review @GaelVaroquaux @jeromedockes

jeromedockes:
> Ready for review @GaelVaroquaux @jeromedockes 

thanks! will have a look this week

nilearn/datasets/tests/test_atlas.py
```diff
@@ -25,12 +25,11 @@
...
-def setup_mock():
-    return tst.setup_mock(utils, atlas)
-
-
-def teardown_mock():
-    return tst.teardown_mock(utils, atlas)
+@pytest.fixture()
+def mock_request():
+    tst.setup_mock(utils, atlas)
+    yield mock_request
```
jeromedockes: this is a bit weird: you are yielding the function itself. it is passed to all the test functions, which of course do not use it. since we don't need to pass anything to the test functions, rather than this fixture use pytest setup and teardown decorators as shown here:

https://docs.pytest.org/en/latest/xunit_setup.html

kchawla-pi: That's one way of doing it, the other way is how I have done it.
https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code
And I like this one better, it offers better control, and it is easier to use multiple fixtures for a test.

kchawla-pi: Essentially nose's 
```python
@with_setup(setup_code, teardown_code)
def test_somthing():
    pass
```
is replaced by Pytest's
```python
@pytest.fixture
def a_fixture():
    setup_code
    yield a_fixture
    teardown_code

def test_something(a_fixture):
    pass
```

jeromedockes: I think you are confusing the fixture with the fixture value, ie the resource it provides to the test function. If you want to use the fixtures it is OK but replace `yield mock_request` with just `yield` or `yield None`.
also here we are still using `nose.with_setup` to get the temp dir. in this case it would make sense to replace it with a fixture yielding the temp dir eg.
```python
@pytest.fixture
def tmpdir():
    with tempfile.TemporaryDirectory() as created_dir:
        yield created_dir


def test_atlas_fetcher(tmpdir):
    print("temp_dir:", tmpdir)
```

but pytest already provides that so use the pytest fixture `tmp_path`:

```python
def test_atlas_fetcher(tmp_path):
    print("temp_dir:", tmp_path)
```

finally we should probably be using `unittest.mock` rather than the custom code for mocking requests but that is probably for another PR

kchawla-pi: I'll look into the first part.
The second part is a separate PR #2269 which also has to be reviewed.
The third part, I didn't write the code for the mocker, I only tweaked it slightly to use pytest instead of nose. It's calling the functions originally written.

kchawla-pi: But I don't like it the way the entire test_utils module is written right now, so I will rewrite the whole thing, starting with the requests mocker soon.

kchawla-pi:
@jeromedockes Ready for another review

PR #2276, opened 2019-12-18
[FIX, DOC] Correct typos in author list

emdupre:
Closes None.

- Corrects typos in author list, as discovered when drafting the Nilearn OHBM abstract


GaelVaroquaux:
Gr8!

@jeromedockes (and any other core devs): do not hesitate to merge such a simple doc PR with only one review when the tests have passed.

PR #2277, opened 2019-12-19
[MRG] Remove explicit calls to nose from library code

kchawla-pi:
This PR removes the final instances of `nose` library being called from Nilearn code.

~We still need to install nose, as components of `numpy.testing` depends on nose. Removing the indirect dependencies, if deemed necessary will be done in another PR.~


kchawla-pi:
@jeromedockes I think there's only one more PR to go and nose will be out of the workflow. I didn't do it here because the change is small but the number of files is large, figured it will be easier to review in a new PR.

nilearn/datasets/tests/test_func.py
```diff
@@ -25,13 +24,70 @@
...
 
 @pytest.fixture()
-def request_mock():
+def request_mocker():
     tst.setup_mock(utils, func)
     yield
     tst.teardown_mock(utils, func)
 
 
-def test_fetch_haxby(tmp_path, request_mock):
```
kchawla-pi: I moved the code block, there was a bunch of stuff, globals, non-test definitions sitting in the middle of all the tests and so I moved it. It has made the PR a little harder to review, but I didn't resist, it was looking terrible.

The transplanted code block is further down, no actual code change.

nilearn/_utils/testing.py
```diff
@@ -245,29 +245,14 @@ def __call__(self, *args, **kwargs):
...
-    # Now check that we have the loader in the call stask
-    stack = inspect.stack()
-    loader_file_name = nose.loader.__file__
-    if loader_file_name.endswith('.pyc'):
-        loader_file_name = loader_file_name[:-1]
-    for _, file_name, _, _, _, _ in stack:
-        if file_name == loader_file_name:
-            return True
-    return False
+    return 'PYTEST_CURRENT_TEST' in os.environ
```
kchawla-pi: Deleting this part and replacing it with a simple one-liner is worth a check.

kchawla-pi:
> I dislike all these functions without a comment/docstring. Could you add this ?

Yeah. I couldn't understand quickly how some of them worked; I plan to rewrite some of these as well.

kchawla-pi:
> I dislike all these functions without a comment/docstring. Could you add this ?

@bthirion You mean the `request_mocker()` and `localizer_mocker()` fixtures?

bthirion:
Indeed.

kchawla-pi:
@bthirion review

nilearn/datasets/tests/test_utils.py
```diff
@@ -26,7 +26,10 @@
 
 
 @pytest.fixture()
-def request_mock():
+def request_mocker():
+    """ Mocks URL calls for data fetchers during testing.
+    Tests the fetcher code without actually downloading he files.
```
bthirion:
```suggestion
    Tests the fetcher code without actually downloading the files.
```

PR #2000, opened 2019-04-16
[ENH] Continuation of  Decoder:Metaestimator

tbng:
Following #698 .

PR #2279, opened 2019-12-20
[maint] Remove fetch_haxby deprecations

emdupre:
With the 0.6.0 release, the `fetch_haxby_simple` function and `n_subjects` parameter in `fetch_haxby` are now at the end of their deprecation cycle. This pull request removes these deprecated features so that they will not appear in a future 0.7.x release !


GaelVaroquaux:
The PR looks great.

However, I hate to say, but my gut feeling would be to hold out a bit from merging it, because we are probably going to have an 0.6.1 release, and maybe and 0.6.2. Not removing deprecated functions now makes it easier to make sure that the following sub releases abide by our promises.

emdupre:
> However, I hate to say, but my gut feeling would be to hold out a bit from merging it, because we are probably going to have an 0.6.1 release, and maybe and 0.6.2. 

Oh that's fine ! Sorry, I wasn't expecting it to get merged right away, but since the last deprecation warning on Haxby-simple had been for the 0.2.x release I thought it was better to have this on our radar -- to make sure it went in, eventually.

Is there a 0.7.x milestone I could add this to ? Or would you rather I close it for now and re-open later ?

GaelVaroquaux:
> Is there a 0.7.x milestone I could add this to ? Or would you rather I close it for now and re-open later ?

That's a good idea!

How about creating a 0.7.x project and adding it there.

emdupre:
Should we go ahead and merge this now that #2000 is merged, @GaelVaroquaux  ?

kchawla-pi:
Once CircleiCI on master has completed successfully, yes!

Shouldn't there be a whats_new entry for this?

kchawla-pi:
CircleCI Passed

GaelVaroquaux:
> Should we go ahead and merge this now that #2000 is merged, @GaelVaroquaux ?

Go ahead! You don't need my approval :)

emdupre:
> Shouldn't there be a whats_new entry for this?

@kchawla-pi looking through WhatsNew, I can't find another example where actually note the _removal_ of a function as opposed to its deprecation. 

For these functions, their deprecation is noted in the release notes for 0.2.6 ! If you think there should be a whats_new entry, though, let me know and I'll add it. Otherwise, let's merge this in :smile_cat:

emdupre:
> Shouldn't there be a whats_new entry for this?

@kchawla-pi looking through WhatsNew, I can't find another example where actually note the _removal_ of a function as opposed to its deprecation. 

For these functions, their deprecation is noted in the release notes for 0.2.6 ! If you think there should be a whats_new entry, though, let me know and I'll add it. Otherwise, let's merge this in :smile_cat:

kchawla-pi:
I think it should be noted. If it hasn't in the past, that puzzles me, but time to set new conventions. I can add it later, if you wish to merge this now.

nilearn/datasets/func.py
```diff
@@ -16,83 +16,10 @@
 from .._utils import check_niimg
 from .._utils.compat import BytesIO, _basestring, _urllib
 from .._utils.numpy_conversions import csv_to_array
-from .._utils.exceptions import VisibleDeprecationWarning
```
GaelVaroquaux: Amusing case of cross-over between PRs: this removal cause https://github.com/nilearn/nilearn/pull/2308 to fail, after merging.

Addressed in: https://github.com/nilearn/nilearn/pull/2321

I think that with the version of numpy that we currently depend upon, we don't need to have a backport of VisibleDeprecationWarning anymore. Let's see what CI says on my other PR, and if it's happy I'll remove our backport.

kchawla-pi: But serious question. Why do we use VisibleDeprecationWarning when Python's built-in FutureWarning does the same thing?

GaelVaroquaux: Let's not get side-tracked and first worry about getting CI to pass on master.

emdupre: Thanks for fixing this !

Issue #2134, opened 2019-09-13
ENH : Adding hierarchical k-means method to Parcellations class (nilearn.regions.parcellation)


thomasbazeille:
Hello, 

While developping [fmralign](https://github.com/thomasbazeille/fmralign), @bthirion included a parcellation based on Hierarchical K-means. From his experience it provides better clustering and works faster than traditional K-means for neuroimaging data.

I can contribute a PR adding this method (hierarchical_kmeans) and corresponding tests to the Parcellation class if it is useful. Just, this method is not available directly in scikit-learn so I'd have to include as well the implementation of the algorithm in a dedicated module.

Would that be helpful ? Should it replace plain kmeans as default if added ?

jeromedockes:
I think that would be useful, if the implementation of hierarchical
K-means is simple.

Note that in the meanwhile you have the Ward method in nilearn for
hierarchical clustering.

ping @KamalakerDadi who knows this part of nilearn well.

GaelVaroquaux:
In my experience too hiearchical kmeans is great of neuroimaging data,
and it would probably be very useful for people.

However: has this been published somewhere? I think that for nilearn we
need to adopt a principled strategy for what we include. Typically,
basing ourselves on publications would be good.

KamalakerDadi:
>While developping fmralign, @bthirion included a parcellation based on Hierarchical K-means. From >his experience it provides better clustering and works faster than traditional K-means for >neuroimaging data.

Just curious, ReNA ? Was it better ? 

>ping @KamalakerDadi who knows this part of nilearn well.

There was ReNA implemented in the same context. So why not, if it is useful.

GaelVaroquaux:
> Just curious, ReNA ? Was it better ?

I would think that ReNA (and Ward) are good for many parcels (10000),
while hierachical kmeans is good in the opposite regime (few hundreds of
parcells).

bthirion:
An interesting property of hierarchical k-means is that you can use it to obtain relatively balanced parcels (e.g. more balanced than Ward).

thomasbazeille:
@GaelVaroquaux after quickly searching through the litterature I didn't find any paper showcasing the use of Hierarchical K-means in neuroimaging. @bthirion do you know such a paper ?

bthirion:
No. The only paper where we were using it was actually not published and never resubmitted ...

I'm not sure I would like to have a dedicated publication on this. We should simply use it in one of our works and then cite that paper.

[...]

PR #2282, opened 2019-12-31
[MRG][ENH] Include Hierarchical KMeans in regions.Parcellations

thomasbazeille:
Here is a pull-request  to include Hierarchical KMeans algorithm as an option in Parcellations object as discussed in #2134 .

The code is behaving as intended, the tests I added are passing. I tried to include this in one example and the user documentation, feedbacks are welcome !

### DIFF
--- a/doc/connectivity/parcellating.rst
+++ b/doc/connectivity/parcellating.rst
@@ -60,9 +60,7 @@ Applying clustering
       spatial smoothing, that will be better balanced than with Kmeans.
 
     All these algorithms are showcased in a full code example :
-    :ref:`here<sphx_glr_auto_examples_03_connectivity_plot_data_driven_parcellations.py>`.
-
-    Below, we focus on explaining the principle of Ward.
+    :ref:`here<sphx_glr_auto_examples_03_connectivity_plot_data_driven_parcellations.py>`. Below, we focus on explaining the principle of Ward.
 
 |
 
--- a/doc/connectivity/parcellating.rst
+++ b/doc/connectivity/parcellating.rst
@@ -53,7 +53,7 @@ Applying clustering
     :class:`nilearn.regions.Parcellations` as well as two algorithms
     tailored to more specific usecases:
 
-    * ReNA is a quicker alternative to Ward with a small loss of precision, it is
+    * :class:`nilearn.regions.ReNA` is a quicker alternative to Ward with a small loss of precision, it is
       ideal to downsize the number of voxels by 10 quickly.
 
     * Hierarchical KMeans is useful to obtain a small number of clusters after
--- a/doc/connectivity/parcellating.rst
+++ b/doc/connectivity/parcellating.rst
@@ -49,12 +49,20 @@ Applying clustering
     * For a small number of clusters, it is preferable to use Kmeans
       clustering after spatially-smoothing the data.
 
-    Both clustering algorithms (as well as many others) are provided by
-    this object :class:`nilearn.regions.Parcellations` and full
-    code example in
+    Both algorithms are provided by this object
+    :class:`nilearn.regions.Parcellations` as well as two algorithms
+    tailored to more specific usecases:
+
+    * ReNA is a quicker alternative to Ward with a small loss of precision, it is
+      ideal to downsize the number of voxels by 10 quickly.
+
+    * Hierarchical KMeans is useful to obtain a small number of clusters after
+      spatial smoothing, that will be better balanced than with Kmeans.
+
+    All these algorithms are showcased in a full code example :
     :ref:`here<sphx_glr_auto_examples_03_connectivity_plot_data_driven_parcellations.py>`.
-    Ward clustering is the easiest to use, as it can be done with the Feature
-    agglomeration object. It is also quite fast. We detail it below.
+
+    Below, we focus on explaining the principle of Ward.
 
 |
 
--- a/examples/03_connectivity/plot_data_driven_parcellations.py
+++ b/examples/03_connectivity/plot_data_driven_parcellations.py
@@ -44,6 +44,14 @@ in the documentation section :ref:`parcellating_brain`.
 #
 # We download one subject of the movie watching dataset from Internet
 
+from matplotlib import patches, ticker
+import matplotlib.pyplot as plt
+from nilearn.image import get_data
+import numpy as np
+from nilearn.image import mean_img, index_img
+from nilearn import plotting
+import time
+from nilearn.regions import Parcellations
 from nilearn import datasets
 dataset = datasets.fetch_development_fmri(n_subjects=1)
 
@@ -59,11 +67,9 @@ print('First subject functional nifti image (4D) is at: %s' %
 # Transforming list of images to data matrix and build brain parcellations,
 # all can be done at once using `Parcellations` object.
 
-from nilearn.regions import Parcellations
 
 # Computing ward for the first time, will be long... This can be seen by
 # measuring using time
-import time
 start = time.time()
 
 # Agglomerative Clustering: ward
@@ -103,8 +109,6 @@ ward_labels_img = ward.labels_img_
 # with the following code:
 ward_labels_img.to_filename('ward_parcellation.nii.gz')
 
-from nilearn import plotting
-from nilearn.image import mean_img, index_img
 
 first_plot = plotting.plot_roi(ward_labels_img, title="Ward parcellation",
                                display_mode='xz')
@@ -121,8 +125,6 @@ cut_coords = first_plot.cut_coords
 # clustering by averaging the signal on each parcel.
 
 # Grab number of voxels from attribute mask image (mask_img_).
-import numpy as np
-from nilearn.image import get_data
 original_voxels = np.sum(get_data(ward.mask_img_))
 
 # Compute mean over time on the functional image to use the mean
@@ -201,11 +203,10 @@ kmeans_labels_img.to_filename('kmeans_parcellation.nii.gz')
 # KMeans clusters to be very unbalanced with a few big clusters and
 # many voxels left as singletons. Hierarchical Kmeans algorithm is
 # tailored to enfore more balanced clusteings. To do this,
-# Hierarchical Kmeans does a first Kmeans clustering in square root
-# of n parcels. In a second step it clusters voxels inside each
+# Hierarchical Kmeans does a first Kmeans clustering in square root of n
+# parcels. In a second step, it clusters voxels inside each
 # of these parcels in m pieces with m adapted to the size of
-# the cluster and  in order to have in the end n clusters
-# as balanced as possible.
+# the cluster in order to have n balanced clusters in the end.
 #
 # This object uses method='hierarchical_kmeans' for Hierarchical KMeans
 # clustering and 10mm smoothing and standardization to compare
@@ -241,7 +242,8 @@ hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
 # distribution. Hierarchical KMeans should give clusters closer to
 # average (600 here) than KMeans.
 #
-# First count how many voxels have each label (except 0 which is the background).
+# First count how many voxels have each label (except 0 which is the
+# background).
 
 kmeans_labels, kmeans_counts = np.unique(
     get_data(kmeans_labels_img), return_counts=True)
@@ -260,8 +262,6 @@ print("... each cluster should contain {} voxels".format(voxel_ratio))
 #
 # You can just skip the plotting code, the important part is the figure
 
-import matplotlib.pyplot as plt
-from matplotlib import patches, ticker
 
 bins = np.concatenate([np.linspace(0, 500, 11), np.linspace(
     600, 2000, 15), np.linspace(3000, 10000, 8)])
@@ -284,8 +284,8 @@ fig.legend(handles, labels, loc=(.5, .8))
 ###########################################################################
 # As we can see, half of the 50 KMeans clusters contain less than
 # 100 voxels whereas three contain several thousands voxels
-# (almost 10000 for the biggest). On the opposite Hierarchical KMeans
-# give clusters that are better balanced.
+# Hierarchical KMeans yield better balanced clusters, with a significant
+# proportion of them containing hundreds to thousands of voxels.
 #
 
 ###########################################################################
--- a/examples/03_connectivity/plot_data_driven_parcellations.py
+++ b/examples/03_connectivity/plot_data_driven_parcellations.py
@@ -202,8 +202,8 @@ kmeans_labels_img.to_filename('kmeans_parcellation.nii.gz')
 # voxels display more and more specific activity patterns causing
 # KMeans clusters to be very unbalanced with a few big clusters and
 # many voxels left as singletons. Hierarchical Kmeans algorithm is
-# tailored to enfore more balanced clusteings. To do this,
-# Hierarchical Kmeans does a first Kmeans clustering in square root of n
+# tailored to enforce more balanced clusteings. To do this,
+# Hierarchical Kmeans does a first Kmeans clustering in square root of n_parcels
 # parcels. In a second step, it clusters voxels inside each
 # of these parcels in m pieces with m adapted to the size of
 # the cluster in order to have n balanced clusters in the end.
--- a/examples/03_connectivity/plot_data_driven_parcellations.py
+++ b/examples/03_connectivity/plot_data_driven_parcellations.py
@@ -203,8 +203,8 @@ kmeans_labels_img.to_filename('kmeans_parcellation.nii.gz')
 # KMeans clusters to be very unbalanced with a few big clusters and
 # many voxels left as singletons. Hierarchical Kmeans algorithm is
 # tailored to enforce more balanced clusterings. To do this,
-# Hierarchical Kmeans does a first Kmeans clustering in square root of n_parcels
-# parcels. In a second step, it clusters voxels inside each
+# Hierarchical Kmeans does a first Kmeans clustering in square root of
+# n_parcels. In a second step, it clusters voxels inside each
 # of these parcels in m pieces with m adapted to the size of
 # the cluster in order to have n balanced clusters in the end.
 #
@@ -230,8 +230,8 @@ plotting.plot_roi(hkmeans_labels_img, mean_func_img,
                   title="Hierarchical KMeans parcellation",
                   display_mode='xz', cut_coords=display.cut_coords)
 
-# kmeans_labels_img is a :class:`nibabel.nifti1.Nifti1Image` object, it can be saved to file with
-# the following code:
+# kmeans_labels_img is a :class:`nibabel.nifti1.Nifti1Image` object, it can be
+# saved to file with the following code:
 hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
 
 ###########################################################################
--- a/examples/03_connectivity/plot_data_driven_parcellations.py
+++ b/examples/03_connectivity/plot_data_driven_parcellations.py
@@ -202,7 +202,7 @@ kmeans_labels_img.to_filename('kmeans_parcellation.nii.gz')
 # voxels display more and more specific activity patterns causing
 # KMeans clusters to be very unbalanced with a few big clusters and
 # many voxels left as singletons. Hierarchical Kmeans algorithm is
-# tailored to enforce more balanced clusteings. To do this,
+# tailored to enforce more balanced clusterings. To do this,
 # Hierarchical Kmeans does a first Kmeans clustering in square root of n_parcels
 # parcels. In a second step, it clusters voxels inside each
 # of these parcels in m pieces with m adapted to the size of
@@ -230,7 +230,7 @@ plotting.plot_roi(hkmeans_labels_img, mean_func_img,
                   title="Hierarchical KMeans parcellation",
                   display_mode='xz', cut_coords=display.cut_coords)
 
-# kmeans_labels_img is a Nifti1Image object, it can be saved to file with
+# kmeans_labels_img is a :class:`nibabel.nifti1.Nifti1Image` object, it can be saved to file with
 # the following code:
 hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
 
@@ -238,7 +238,7 @@ hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
 # Compare Hierarchical Kmeans clusters with those from Kmeans
 # ...........................................................
 # To compare those, we'll first count how many voxels are contained in
-# each of the 50 clusters for both algorithm and compare those sizes
+# each of the 50 clusters for both algorithms and compare those sizes
 # distribution. Hierarchical KMeans should give clusters closer to
 # average (600 here) than KMeans.
 #
@@ -258,7 +258,7 @@ voxel_ratio = np.round(np.sum(kmeans_counts[1:]) / 50)
 print("... each cluster should contain {} voxels".format(voxel_ratio))
 
 ###########################################################################
-# Let's plot clusters sizes distributions for both algorithm
+# Let's plot clusters sizes distributions for both algorithms
 #
 # You can just skip the plotting code, the important part is the figure
 
--- a/examples/03_connectivity/plot_data_driven_parcellations.py
+++ b/examples/03_connectivity/plot_data_driven_parcellations.py
@@ -199,8 +199,8 @@ kmeans_labels_img.to_filename('kmeans_parcellation.nii.gz')
 # As the number of images from which we try to cluster grows,
 # voxels display more and more specific activity patterns causing
 # KMeans clusters to be very unbalanced with a few big clusters and
-# many voxels left almost alone. Hierarchical Kmeans algorithm is
-# tailored to enfore more balanced clusteingsself. To do this,
+# many voxels left as singletons. Hierarchical Kmeans algorithm is
+# tailored to enfore more balanced clusteings. To do this,
 # Hierarchical Kmeans does a first Kmeans clustering in square root
 # of n parcels. In a second step it clusters voxels inside each
 # of these parcels in m pieces with m adapted to the size of
@@ -236,7 +236,7 @@ hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
 ###########################################################################
 # Compare Hierarchical Kmeans clusters with those from Kmeans
 # ...........................................................
-# To compare those, we'll first count how many voxels is contained in
+# To compare those, we'll first count how many voxels are contained in
 # each of the 50 clusters for both algorithm and compare those sizes
 # distribution. Hierarchical KMeans should give clusters closer to
 # average (600 here) than KMeans.
--- a/examples/03_connectivity/plot_data_driven_parcellations.py
+++ b/examples/03_connectivity/plot_data_driven_parcellations.py
@@ -186,7 +186,7 @@ print("KMeans clusters: %.2fs" % (time.time() - start))
 # Grab parcellations of brain image stored in attribute `labels_img_`
 kmeans_labels_img = kmeans.labels_img_
 
-display = plotting.plot_roi(kmeans_labels_img, mean_func_img,
+plotting.plot_roi(kmeans_labels_img, mean_func_img,
                             title="KMeans parcellation",
                             display_mode='xz')
 
--- a/examples/03_connectivity/plot_data_driven_parcellations.py
+++ b/examples/03_connectivity/plot_data_driven_parcellations.py
@@ -232,7 +232,7 @@ plotting.plot_roi(hkmeans_labels_img, mean_func_img,
 
 # kmeans_labels_img is a :class:`nibabel.nifti1.Nifti1Image` object, it can be
 # saved to file with the following code:
-hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
+hkmeans_labels_img.to_filename('hierarchical_kmeans_parcellation.nii.gz')
 
 ###########################################################################
 # Compare Hierarchical Kmeans clusters with those from Kmeans
--- a/examples/03_connectivity/plot_data_driven_parcellations.py
+++ b/examples/03_connectivity/plot_data_driven_parcellations.py
@@ -2,8 +2,8 @@
 Clustering methods to learn a brain parcellation from fMRI
 ==========================================================
 
-We use spatially-constrained Ward-clustering, KMeans, Hierarchical KMeans and Recursive Neighbor
-Agglomeration (ReNA) to create a set of parcels.
+We use spatially-constrained Ward-clustering, KMeans, Hierarchical KMeans
+and Recursive Neighbor Agglomeration (ReNA) to create a set of parcels.
 
 In a high dimensional regime, these methods can be interesting
 to create a 'compressed' representation of the data, replacing the data
@@ -18,7 +18,7 @@ References
 ----------
 
 Which clustering method to use, an empirical comparison can be found in this
-paper
+paper:
 
     * Bertrand Thirion, Gael Varoquaux, Elvis Dohmatob, Jean-Baptiste Poline.
       `Which fMRI clustering gives good brain parcellations ?
@@ -26,7 +26,7 @@ paper
       2014.
 
 This parcellation may be useful in a supervised learning, see for
-instance
+instance:
 
     * Vincent Michel, Alexandre Gramfort, Gael Varoquaux, Evelyn Eger,
       Christine Keribin, Bertrand Thirion. `A supervised clustering approach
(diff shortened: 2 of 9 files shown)


thomasbazeille:
While building the documentation, I noticed that the results of clustering  fmri_development dataset in 50 pieces with Kmeans and Hierarchical Kmeans don't give the same results locally and during the build and I have no clue why. Does someone as an idea why it could be so ?

bthirion:
Well, these are not deterministic methods.
Why is the result difference an issue ?

emdupre:
Looks like there's separately a Sphinx error... ?

> sphinx.errors.SphinxWarning: /home/circleci/project/doc/auto_examples/03_connectivity/plot_data_driven_parcellations.rst:428:Unexpected section title or transition.

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +183,111 @@
...
-##################################################################
+#########################################################################
+# Brain parcellations with Hierarchical KMeans Clustering
+# -------------------------------------------------------
+#
+# As the number of images from which we try to cluster grows,
+# voxels display more and more specific activity patterns causing
+# KMeans clusters to be very unbalanced with a few big clusters and
+# many voxels left almost alone. Hierarchical Kmeans algorithm is
+# tailored to enfore more balanced clusteingsself. To do this,
```
bthirion:
```suggestion
# tailored to enfore more balanced clusteings. To do this,
```

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +183,111 @@
...
 
-##################################################################
+#########################################################################
+# Brain parcellations with Hierarchical KMeans Clustering
+# -------------------------------------------------------
+#
+# As the number of images from which we try to cluster grows,
+# voxels display more and more specific activity patterns causing
+# KMeans clusters to be very unbalanced with a few big clusters and
+# many voxels left almost alone. Hierarchical Kmeans algorithm is
```
bthirion:
```suggestion
# many voxels left as singletons. Hierarchical Kmeans algorithm is
```

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +183,111 @@
...
+                  display_mode='xz', cut_coords=display.cut_coords)
+
+# kmeans_labels_img is a Nifti1Image object, it can be saved to file with
+# the following code:
+hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
+
+###########################################################################
+# Compare Hierarchical Kmeans clusters with those from Kmeans
+# ...........................................................
+# To compare those, we'll first count how many voxels is contained in
```
bthirion:
```suggestion
# To compare those, we'll first count how many voxels are contained in
```

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,239 @@
...
+    ----------
+    X: nd array (n_samples, n_features)
+        Data to cluster
+
+    n_clusters: int,
+        The number of clusters to find.
+
+    init : {'k-means++', 'random' or an ndarray}
+        Method for initialization, defaults to 'k-means++':
+        'k-means++' : selects initial cluster centers for k-mean
```
bthirion:
```suggestion
        'k-means++' : selects initial cluster centers for k-means
```

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,239 @@
...
+    inside these big clusters into smaller ones.
+
+    Parameters
+    ----------
+    n_clusters: int
+        The number of clusters to find.
+
+    init : {'k-means++', 'random' or an ndarray}
+        Method for initialization, defaults to 'k-means++':
+        'k-means++' : selects initial cluster centers for k-mean
```
bthirion:
```suggestion
        'k-means++' : selects initial cluster centers for k-means
```

bthirion:
+ whatsnew entry

thomasbazeille:
I'm not 100% this is needed in Nilearn, but I could definitively push it through if you think so

bthirion:
It has the advantage of returning relatively clusters. In case where Ward or Rena do not work well, its is good to have it.
And the code is relatively simple.

thomasbazeille:
I'm slowly reviving this PR. I noticed there seem to be duplicated test in the [main branch](https://github.com/nilearn/nilearn/blob/main/nilearn/regions/tests/test_parcellations.py) in lines 65 -73

The exception in l65 should just apply to connectivity_ (l66 to l68); so that l69-70 are duplicate of l61 and l73 should apply to every method. Should I proceed with this change in this PR as well @NicolasGensollen.

Thanks a lot !

NicolasGensollen:
@thomasbazeille indeed there are duplicates in this test function.

>Should I proceed with this change in this PR as well @NicolasGensollen.

If you are changing this test for this PR, I think you can refactor here. Otherwise, it would be more clear to open a very small PR targeting this specific change.

thomasbazeille:
> If you are changing this test for this PR, I think you can refactor here. 

As I already touched this test, I proceeded with the refactoring

I also :

- [x] fixed PEP8
- [x] increased coverage, by adding a few test (actually catching a bug !) 
- [x] double checked the example modification (and request a full build)

 I might still have weird test failures only on some macos jobs that I don't understand at all.

Apart from that, this seems ready for review on my side (I didn't add a whatsnew entry yet to not conflict with the release if it is to happen soon)

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +185,109 @@
...
+#########################################################################
+# Brain parcellations with Hierarchical KMeans Clustering
+# -------------------------------------------------------
+#
+# As the number of images from which we try to cluster grows,
+# voxels display more and more specific activity patterns causing
+# KMeans clusters to be very unbalanced with a few big clusters and
+# many voxels left as singletons. Hierarchical Kmeans algorithm is
+# tailored to enfore more balanced clusteings. To do this,
+# Hierarchical Kmeans does a first Kmeans clustering in square root of n
```
bthirion, line 205:
```suggestion
# Hierarchical Kmeans does a first Kmeans clustering in square root of n_parcels
```

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +185,109 @@
...
-##################################################################
+#########################################################################
+# Brain parcellations with Hierarchical KMeans Clustering
+# -------------------------------------------------------
+#
+# As the number of images from which we try to cluster grows,
+# voxels display more and more specific activity patterns causing
+# KMeans clusters to be very unbalanced with a few big clusters and
+# many voxels left as singletons. Hierarchical Kmeans algorithm is
+# tailored to enfore more balanced clusteings. To do this,
```
bthirion, line 204:
```suggestion
# tailored to enforce more balanced clusteings. To do this,
```

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,248 @@
...
+            Data reduced expanded to the original feature space
+        """
+
+        check_is_fitted(self, "labels_")
+
+        _, inverse = np.unique(self.labels_, return_inverse=True)
+
+        if self.scaling:
+            X_red = np.divide(X_red, np.asarray(
+                [np.sqrt(self.sizes_)] * X_red.shape[1]).T)
```
bthirion, line 245: Isn't it better than using `np.ones(X_red.shape[1])  * np.sqrt(self.sizes_)` ?

bthirion (commented):
LGTM, only minor glitches.
Just wondering is the tests could be refactored.

doc/connectivity/parcellating.rst
```diff
@@ -49,12 +49,18 @@ Applying clustering
...
-    this object :class:`nilearn.regions.Parcellations` and full
-    code example in
-    :ref:`here<sphx_glr_auto_examples_03_connectivity_plot_data_driven_parcellations.py>`.
-    Ward clustering is the easiest to use, as it can be done with the Feature
-    agglomeration object. It is also quite fast. We detail it below.
+    Both algorithms are provided by this object
+    :class:`nilearn.regions.Parcellations` as well as two algorithms
+    tailored to more specific usecases:
+
+    * ReNA is a quicker alternative to Ward with a small loss of precision, it is
```
NicolasGensollen, line 56:
```suggestion
    * :class:`nilearn.regions.ReNA` is a quicker alternative to Ward with a small loss of precision, it is
```

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -2,30 +2,31 @@
 Clustering methods to learn a brain parcellation from fMRI
 ==========================================================
 
-We use spatially-constrained Ward-clustering, KMeans, and Recursive Neighbor
-Agglomeration (ReNA) to create a set of parcels.
+We use spatially-constrained Ward-clustering, KMeans, Hierarchical KMeans
+and Recursive Neighbor Agglomeration (ReNA) to create a set of parcels.
```
NicolasGensollen, line 6: Is there a reference we can add for this?

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +185,109 @@
...
-##################################################################
+#########################################################################
+# Brain parcellations with Hierarchical KMeans Clustering
+# -------------------------------------------------------
+#
+# As the number of images from which we try to cluster grows,
+# voxels display more and more specific activity patterns causing
+# KMeans clusters to be very unbalanced with a few big clusters and
+# many voxels left as singletons. Hierarchical Kmeans algorithm is
+# tailored to enforce more balanced clusteings. To do this,
```
NicolasGensollen, line 204:
```suggestion
# tailored to enforce more balanced clusterings. To do this,
```

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +185,109 @@
...
+# ....................................................
+#
+# Grab parcellations of brain image stored in attribute `labels_img_`
+hkmeans_labels_img = hkmeans.labels_img_
+
+plotting.plot_roi(hkmeans_labels_img, mean_func_img,
+                  title="Hierarchical KMeans parcellation",
+                  display_mode='xz', cut_coords=display.cut_coords)
+
+# kmeans_labels_img is a Nifti1Image object, it can be saved to file with
```
NicolasGensollen, line 232:
```suggestion
# kmeans_labels_img is a :class:`nibabel.nifti1.Nifti1Image` object, it can be saved to file with
```

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +185,109 @@
...
+# the following code:
+hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
+
+###########################################################################
+# Compare Hierarchical Kmeans clusters with those from Kmeans
+# ...........................................................
+# To compare those, we'll first count how many voxels are contained in
+# each of the 50 clusters for both algorithm and compare those sizes
```
NicolasGensollen, line 240:
```suggestion
# each of the 50 clusters for both algorithms and compare those sizes
```

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +185,109 @@
...
+    get_data(hkmeans_labels_img), return_counts=True)
+
+voxel_ratio = np.round(np.sum(kmeans_counts[1:]) / 50)
+
+# If all voxels not in background were balanced between clusters ...
+
+print("... each cluster should contain {} voxels".format(voxel_ratio))
+
+###########################################################################
+# Let's plot clusters sizes distributions for both algorithm
```
NicolasGensollen, line 260:
```suggestion
# Let's plot clusters sizes distributions for both algorithms
```

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,248 @@
...
+        Method for initialization, defaults to 'k-means++':
+        'k-means++' : selects initial cluster centers for k-means
+        clustering in a smart way to speed up convergence. See section
+        Notes in k_init for more details.
+        'random': choose k observations (rows) at random from data for
+        the initial centroids.
+        If an ndarray is passed, it should be of shape (n_clusters, n_features)
+        and gives the initial centers.
+
+    batch_size : int, optional, default: 100
```
NicolasGensollen, line 42: Should be 1000?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,248 @@
...
+        Size of the mini batches. (Kmeans performed through MiniBatchKMeans)
+
+    n_init : int, default=3
```
NicolasGensollen, line 45: Should be 10?

nilearn/regions/hierarchical_kmeans_clustering.py
thomasbazeille: it is probably possible to optimize but I didn't find a very compact way. 

Your current expression is false since :
self.sizes_ is of shape (7,)
X_red is of shape (7, 15)

And the previous one was false as well. 

```np.sqrt(self.sizes_) * X_red```

File: nilearn/regions/hierarchical_kmeans_clustering.py [hunk 17]
```diff
@@ -0,0 +1,248 @@
...
+                              max_no_improvement=max_no_improvement,
+                              verbose=verbose,
+                              n_init=n_init,).fit(X[coarse_labels == i])
+        fine_labels[coarse_labels == i] = q + mbk.labels_
+        q += n_small_clusters
+
+    return _remove_empty_labels(fine_labels)
+
+
+class HierarchicalKMeans(BaseEstimator, ClusterMixin, TransformerMixin):
```
NicolasGensollen, line 90: Is this class supposed to be public? If yes, we need to add it to `doc/modules/reference.rst`

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,248 @@
...
+    inside these big clusters into smaller ones.
+
+    Parameters
+    ----------
+    n_clusters: int
+        The number of clusters to find.
+
+    init : {'k-means++', 'random' or an ndarray}
+        Method for initialization, defaults to 'k-means++':
+        'k-means++' : selects initial cluster centers for k-means
```
NicolasGensollen, line 102: Having a bullet list will render better I think

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,248 @@
...
+        clustering in a smart way to speed up convergence. See section
+        Notes in k_init for more details.
+        'random': choose k observations (rows) at random from data for
+        the initial centroids.
+        If an ndarray is passed, it should be of shape (n_clusters, n_features)
+        and gives the initial centers.
+
+    batch_size : int, optional, default: 100
```
NicolasGensollen, line 110: Should be 1000?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,248 @@
...
+        Size of the mini batches. (Kmeans performed through MiniBatchKMeans)
+
+    n_init : int, default=3
```
NicolasGensollen, line 113: Should be 10?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,248 @@
...
+                              verbose=verbose,
+                              n_init=n_init,).fit(X[coarse_labels == i])
+        fine_labels[coarse_labels == i] = q + mbk.labels_
+        q += n_small_clusters
+
+    return _remove_empty_labels(fine_labels)
+
+
+class HierarchicalKMeans(BaseEstimator, ClusterMixin, TransformerMixin):
+    """Hierarchical KMeans:
```
NicolasGensollen, line 91: If it is public we should also expose it in `__init__.py`

NicolasGensollen (commented):
Thanks @thomasbazeille ! :+1: 
I made a quick first pass:

examples/03_connectivity/plot_data_driven_parcellations.py
thomasbazeille: To document the class ?? Unfortunately this has not yet been clearly used in a publication. This has been raised in the related issue but people decided to push forward anyhow

nilearn/regions/hierarchical_kmeans_clustering.py
bthirion: `X_red / sizes[:, np.newaxis]` then ?

examples/03_connectivity/plot_data_driven_parcellations.py
bthirion: indeed.

NicolasGensollen: You're right indeed. 
Closing this conversation...

thomasbazeille:
remaining bugs in the CI (weirdly they happen for macos and Windows only) where the number of clusters outputed through HKmeans are 4 instead of 5 in some examples (which is not really a bug but doesn't correspond to the assertion tested)

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -15,6 +15,25 @@ def _remove_empty_labels(labels):
...
+    elif np.sum(list_round) < n_clusters:
+        while np.sum(list_round) != n_clusters:
+            idx = np.argmax(list - list_round)
+            list_round[idx] += 1
+        return list_round.astype(int)
+    elif np.sum(list_round) > n_clusters:
+        while np.sum(list_round) != n_clusters:
+            idx = np.argmin(list - list_round)
+            list_round[idx] -= 1
+        return list_round.astype(int)
```
bthirion: But you probably need to check that list_round is nonzero ...

thomasbazeille: I agree, and the tests caught that ! Actually, you reviewed an outdated version of the code. The current version also includes `array_round = np.maximum(array_round, np.ones(array.size))`

thomasbazeille:
The CI lights are green and cover a good amount of edge cases. This is ready for review I think.

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +185,109 @@
 # Grab parcellations of brain image stored in attribute `labels_img_`
 kmeans_labels_img = kmeans.labels_img_
 
-plotting.plot_roi(kmeans_labels_img, mean_func_img,
-                  title="KMeans parcellation",
-                  display_mode='xz')
+display = plotting.plot_roi(kmeans_labels_img, mean_func_img,
```
bthirion, line 188:
```suggestion
plotting.plot_roi(kmeans_labels_img, mean_func_img,
```

examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -182,15 +185,109 @@
...
+# Grab parcellations of brain image stored in attribute `labels_img_`
+hkmeans_labels_img = hkmeans.labels_img_
+
+plotting.plot_roi(hkmeans_labels_img, mean_func_img,
+                  title="Hierarchical KMeans parcellation",
+                  display_mode='xz', cut_coords=display.cut_coords)
+
+# kmeans_labels_img is a :class:`nibabel.nifti1.Nifti1Image` object, it can be
+# saved to file with the following code:
+hkmeans_labels_img.to_filename('hkmeans_parcellation.nii.gz')
```
bthirion, line 234:
```suggestion
hkmeans_labels_img.to_filename('hierarchical_kmeans_parcellation.nii.gz')
```

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,273 @@
...
+    if np.sum(array_round) == n_clusters:
+        return array_round.astype(int)
+    elif np.sum(array_round) < n_clusters:
+        while np.sum(array_round) != n_clusters:
+            idx = np.argmax(array - array_round)
+            array_round[idx] += 1
+        return array_round.astype(int)
+    elif np.sum(array_round) > n_clusters:
+        while np.sum(array_round) != n_clusters:
+            mask = array_round != np.ones(array.size)
```
bthirion, line 34: Maybe explain why you mask

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,273 @@
...
+                              n_init=n_init,).fit(X[coarse_labels == i])
+        fine_labels[coarse_labels == i] = q + mbk.labels_
+        q += n_small_clusters
+
+    return _remove_empty_labels(fine_labels)
+
+
+class HierarchicalKMeans(BaseEstimator, ClusterMixin, TransformerMixin):
+    """Hierarchical KMeans:
+    First clusterize the samples into big clusters. Then clusterize the samples
```
bthirion, line 116:
```suggestion
    First cluster the samples into big clusters. Then cluster the samples
```

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,273 @@
...
+        max_no_improvement to None.
+
+    random_state : int, RandomState instance or None (default)
+        Determines random number generation for centroid initialization and
+        random reassignment. Use an int to make the randomness deterministic.
+        See :term:`Glossary <random_state>`.
+
+    scaling: bool, optional (default False)
+        If scaling is True, each cluster is scaled by the square root of its
+        size, preserving the l2-norm of the image.
```
bthirion, line 158: and  the inverse_transform method applied the inverse scaling.

bthirion (commented):
Thx ! This looks good. Simply find minor comments below.

examples/03_connectivity/plot_data_driven_parcellations.py
thomasbazeille: actually this is used to reuse cut_coords in the next figure so this is necessary.

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,277 @@
...
+    max_no_improvement : int, default: 10
+        Control early stopping based on the consecutive number of mini
+        batches that does not yield an improvement on the smoothed inertia.
+        To disable convergence detection based on inertia, set
+        max_no_improvement to None.
+
+    random_state : int, RandomState instance or None (default)
+        Determines random number generation for centroid initialization and
+        random reassignment. Use an int to make the randomness deterministic.
+        See :term:`Glossary <random_state>`.
```
NicolasGensollen, line 156: Same comment.

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,277 @@
...
+        q += n_small_clusters
+
+    return _remove_empty_labels(fine_labels)
+
+
+class HierarchicalKMeans(BaseEstimator, ClusterMixin, TransformerMixin):
+    """Hierarchical KMeans:
+    First clusterize the samples into big clusters. Then clusterize the samples
+    inside these big clusters into smaller ones.
+
```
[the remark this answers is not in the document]
thomasbazeille: Does this make the whatsnew.rst entry useless ?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,277 @@
...
+def _adjust_small_clusters(array, n_clusters):
+    '''Takes a ndarray of floats summing to n_clusters and try to round it while
+    enforcing rounded array still sum to n_clusters and every element is at
+    least 1.
+    '''
+
+    array_round = np.round(array).astype(int)
+    array_round = np.maximum(array_round, np.ones(array.size))
+    if np.sum(array_round) == n_clusters:
```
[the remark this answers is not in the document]
thomasbazeille: I'd rather not for code readability. I reordered the cases to make the most frequent come first but if it had to be done 3 times, it would still only be summing a list of length << 100

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,277 @@
...
+                          batch_size=batch_size, n_init=n_init,
+                          max_no_improvement=max_no_improvement,
+                          verbose=verbose, random_state=random_state).fit(X)
+    coarse_labels = mbk.labels_
+    fine_labels = np.zeros_like(coarse_labels)
+    q = 0
+    exact_clusters = np.asarray([n_clusters * np.sum(coarse_labels == i) * 1.
+                                 / X.shape[0] for i in range(n_big_clusters)])
+
+    adjusted_clusters = _adjust_small_clusters(exact_clusters, n_clusters)
```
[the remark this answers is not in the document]
thomasbazeille: `exact_clusters` can be empty only if `n_big_clusters <= 0` which means  `n_clusters  < 1`(mostly covered by a test in `HierarchicalKMeans` class which checks `n_clusters  > 0`)

I'd rather not add code to raise clean errors in such narrow cases. If one triggers it, he's definitely not trying to make a real use of the function for which the goal is to get meaningful clusters. (but if you think it's better I'll do it anyway)

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,277 @@
...
+    for i, n_small_clusters in enumerate(adjusted_clusters):
+        mbk = MiniBatchKMeans(init=init, n_clusters=n_small_clusters,
+                              batch_size=batch_size, random_state=random_state,
+                              max_no_improvement=max_no_improvement,
+                              verbose=verbose,
+                              n_init=n_init,).fit(X[coarse_labels == i])
+        fine_labels[coarse_labels == i] = q + mbk.labels_
+        q += n_small_clusters
+
+    return _remove_empty_labels(fine_labels)
```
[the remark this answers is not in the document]
thomasbazeille: It can't, it is instantiated as a list of zeros and stays like that at worst.

nilearn/regions/hierarchical_kmeans_clustering.py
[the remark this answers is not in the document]
NicolasGensollen: Fair enough. I was thinking about cases where the user doesn't provide a valid integer for n_clusters, for example 0.5 (which could happen easily in a complex pipeline where you get unexpected inputs...). In this case, n_big_clusters will be zero, but I believe MiniBatchKmeans will fail in this situation.
A one line safety guard could be to cast n_clusters to an integer in the constructor of HierarchicalKmeans. WDYT?

thomasbazeille:
I added the whatsnew a few weeks ago @NicolasGensollen and solved conflicts introduced since then. I think this is ready for merge!

NicolasGensollen (approved):
Thanks @thomasbazeille !
I also think this is ready! :tada:

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,276 @@
...
+            array_round[idx] += 1
+        return array_round.astype(int)
+    elif np.sum(array_round) == n_clusters:
+        return array_round.astype(int)
+    elif np.sum(array_round) > n_clusters:
+        while np.sum(array_round) != n_clusters:
+            # prevent element rounded to 1 to be decreased in edge cases
+            mask = array_round != np.ones(array.size)
+            idx = np.argmin(array[mask] - array_round[mask])
+            parent_idx = np.arange(array_round.shape[0])[mask][idx]
```
jeromedockes, line 39: the `arange` could be outside the loop

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,276 @@
...
+
+    n_big_clusters = int(np.sqrt(n_clusters))
+    mbk = MiniBatchKMeans(init=init, n_clusters=n_big_clusters,
+                          batch_size=batch_size, n_init=n_init,
+                          max_no_improvement=max_no_improvement,
+                          verbose=verbose, random_state=random_state).fit(X)
+    coarse_labels = mbk.labels_
+    fine_labels = np.zeros_like(coarse_labels)
+    q = 0
+    exact_clusters = np.asarray([n_clusters * np.sum(coarse_labels == i) * 1.
```
jeromedockes, line 100: you  could use `np.bincount` or `np.unique(..., return_counts=True)` to get the cluster sizes rather than repeatedly call `np.sum` and `==`

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,276 @@
...
+    for i, n_small_clusters in enumerate(adjusted_clusters):
+        mbk = MiniBatchKMeans(init=init, n_clusters=n_small_clusters,
+                              batch_size=batch_size, random_state=random_state,
+                              max_no_improvement=max_no_improvement,
+                              verbose=verbose,
+                              n_init=n_init,).fit(X[coarse_labels == i])
+        fine_labels[coarse_labels == i] = q + mbk.labels_
+        q += n_small_clusters
+
+    return _remove_empty_labels(fine_labels)
```
jeromedockes, line 113: would it make sense to also return the coarse labels if someone wants to see the hierarchy?

nilearn/regions/hierarchical_kmeans_clustering.py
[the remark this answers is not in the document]
jeromedockes: is it possible for `n_small_clusters` to end up larger than the number of voxels in the large cluster?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,276 @@
...
+            self.n_clusters = n_features
+            warnings.warn("n_clusters should be at most the number of "
+                          "features. Taking n_clusters = %s instead."
+                          % str(n_features))
+        self.labels_ = hierarchical_k_means(X, self.n_clusters, self.init,
+                                            self.batch_size, self.n_init,
+                                            self.max_no_improvement,
+                                            self.verbose, self.random_state)
+        sizes = np.bincount(self.labels_)
+        sizes = sizes[sizes > 0]
```
jeromedockes, line 220: doesn't `_remove_empty_labels` take care of this?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,276 @@
...
+        self.sizes_ = sizes
+        self.n_clusters = len(np.unique(self.labels_))
```
jeromedockes, line 222: isn't that just `len(sizes)`?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,276 @@
...
+            Data to transform with the fitted clustering.
+
+        Returns
+        -------
+        X_red: ndarray, shape = [n_clusters, n_samples]
+            Data reduced with agglomerated signal for each cluster
+        """
+
+        check_is_fitted(self, "labels_")
+        unique_labels = np.unique(self.labels_)
```
jeromedockes, line 240: isn't this 0, ..., n_clusters -1?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,276 @@
...
+
+        mean_cluster = []
```
jeromedockes, line 242: you can use `np.empty` and fill the array if you don't want to build a list and then an array

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -0,0 +1,276 @@
...
+
+        Returns
+        -------
+        X_inv: ndarray, shape = [n_features, n_samples]
+            Data reduced expanded to the original feature space
+        """
+
+        check_is_fitted(self, "labels_")
+
+        _, inverse = np.unique(self.labels_, return_inverse=True)
```
jeromedockes, line 270: iiuc the labels corresponding to reduced dimensions will be 0, ... n_clusters -1 so this might not be necessary? and if it is maybe we can store it after fit

jeromedockes (approved):
only small details, feel free to skip

jeromedockes (approved):
thanks, LGTM! this is a great addition :)

NicolasGensollen:
Thanks everyone! Merging...

PR #2285, opened 2020-01-06
[WIP, ENH] Add Binder links to examples

emdupre:
Closes #2284.

- Building off of https://github.com/scikit-learn/scikit-learn/pull/11221, adds Binder links to Nilearn examples

_To Do:_

- [x] Build off stable release, rather than master branch, so docs remain stable

PR #2290, opened 2020-01-20
[FIX] Patch non-resolving OSF download links

emdupre:
To troubleshoot #2285.

The original resource we're requesting returns a 308 error. Trying with a 302 link, since it _might_ redirect on its own and we can still query it systematically. See:

Screenshot from 2020-01-20 11-33-18

@kchawla-pi did OSF admin get back in touch with you ?

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -920,6 +920,9 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
 
     # Build data URLs that will be fetched
     files = {}
+    # Download from the relevant OSF project, using hashes generated
+    # from the OSF API. Note the trailing slash. For more info, see:
+    # https://gist.github.com/emdupre/3cb4d564511d495ea6bf89c6a577da74
     root_url = "https://osf.io/download/{0}/"
     if isinstance(n_subjects, numbers.Number):
         subject_mask = np.arange(1, n_subjects + 1)
@@ -1957,6 +1960,9 @@ def _fetch_development_fmri_functional(participants, data_dir, url, resume,
                                 verbose=verbose)
 
     if url is None:
+        # Download from the relevant OSF project, using hashes generated
+        # from the OSF API. Note the trailing slash. For more info, see:
+        # https://gist.github.com/emdupre/3cb4d564511d495ea6bf89c6a577da74
         url = 'https://osf.io/download/{}/'
 
     confounds = '{}_task-pixar_desc-confounds_regressors.tsv'
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -920,7 +920,7 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
 
     # Build data URLs that will be fetched
     files = {}
-    root_url = "https://files.de-1.osf.io/v1/resources/vhtf6/providers/osfstorage/{0}"
+    root_url = "https://osf.io/download/{0}/"
     if isinstance(n_subjects, numbers.Number):
         subject_mask = np.arange(1, n_subjects + 1)
         subject_id_max = "S%02d" % n_subjects
@@ -1957,7 +1957,7 @@ def _fetch_development_fmri_functional(participants, data_dir, url, resume,
                                 verbose=verbose)
 
     if url is None:
-        url = 'https://osf.io/download/{}'
+        url = 'https://osf.io/download/{}/'
 
     confounds = '{}_task-pixar_desc-confounds_regressors.tsv'
     func = '{0}_task-pixar_space-MNI152NLin2009cAsym_desc-preproc_bold.nii.gz'
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -920,7 +920,7 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
 
     # Build data URLs that will be fetched
     files = {}
-    root_url = "https://osf.io/download/{0}"
+    root_url = "https://files.de-1.osf.io/v1/resources/vhtf6/providers/osfstorage/{0}"
     if isinstance(n_subjects, numbers.Number):
         subject_mask = np.arange(1, n_subjects + 1)
         subject_id_max = "S%02d" % n_subjects


emdupre:
Ugh I was terrible on this, sorry. I'll fix it, but the circle failure is all I wanted to check on anyway....

kchawla-pi:
Thanks for trying this @emdupre .
What do you mean you were terrible on this?

No they haven't responded.

I tried running
```python

fetch_localizer_contrasts('<all the contrasts in the docstring>')
```
And had these errors so far:

The contrast (listed in the docstring of the fetcher) `"calculation and sentence listening/reading"`  raises a `ValueError`. Omitting it, raised this error:
```python
Downloading data from http://brainomics.cea.fr/localizer/brainomics_data_23.zip ...
Downloaded 9421 of 9421 bytes (100.0%,    0.0s remaining) ...done. (0 seconds, 0 min)
Extracting data from C:\Users\Kshitij/nilearn_data\brainomics_localizer\529d1fc6d37f3f4612b35eda792dabcf\brainomics_data_23.zip...Error uncompressing file: [Uncompress] unknown archive file format: C:\Users\Kshitij/nilearn_data\brainomics_localizer\529d1fc6d37f3f4612b35eda792dabcf\brainomics_data_23.zip
  File "C:/Users/Kshitij/.PyCharm2019.3/config/scratches/scratch_3.py", line 37, in <module>
    fetch_localizer_contrasts(contrasts
  File "C:\Users\Kshitij\Miniconda3\envs\nistats-p27-latest\lib\site-packages\nilearn\datasets\func.py", line 994, in fetch_localizer_contrasts
    files = _fetch_files(data_dir, filenames, verbose=verbose)
  File "C:\Users\Kshitij\Miniconda3\envs\nistats-p27-latest\lib\site-packages\nilearn\datasets\utils.py", line 766, in _fetch_files
    raise IOError('Fetching aborted: ' + abort)
IOError: Fetching aborted: [Uncompress] unknown archive file format: C:\Users\Kshitij/nilearn_data\brainomics_localizer\529d1fc6d37f3f4612b35eda792dabcf\brainomics_data_23.zip
```

emdupre:
> What do you mean you were terrible on this?

I pulled in an unrelated git history (ie, the history of #2285). I'll cherrypick the relevant commit if it looks promising...

> No they haven't responded.

OK, thanks for checking with them !

> I tried running

Hmm. It looks like they're you're not actually downloading from OSF, but from the CEA server:

```Downloading data from http://brainomics.cea.fr/localizer/brainomics_data_23.zip```

The one relevant commit for this PR is https://github.com/nilearn/nilearn/pull/2290/commits/756403339ec224365041f18aba7988904dbd2477, which just updates the `fetch_localizer_contrast` function to the link I saw in the console as its redirect. So, it shouldn't interface with the problem you're seeing at all ! But let's see if I can overcome the 308 error, which might reveal your problem.

**Comment by kchawla-pi** [comment]
> which might reveal your problem.

Let's hope you have better luck than my therapist.

**Comment by kchawla-pi** [comment]
That might be the problem, the CEA server is not redirecting properly?

**Comment by emdupre** [comment]
Woohoo, it worked ! :tada: From the [current Circle job](https://circleci.com/gh/nilearn/nilearn/5226?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link):

``` 
Downloaded 1508563 of 1508563 bytes (100.0%,    0.0s remaining) ...done. (2 seconds, 0 min)
Downloading data from https://files.de-1.osf.io/v1/resources/vhtf6/providers/osfstorage/5d27ddc91c5b4a001b9ef9d0 ...
```

The job should fail as soon as it hits the next OSF dataset (probably development fMRI), but I can spruce up this PR so we can fix all those links in one fell swoop. WDYT ?

**Comment by GaelVaroquaux** [comment]
> but I can spruce up this PR so we can fix all those links in one fell swoop. WDYT ?

Sounds good. Can you document somewhere (eg as a comment in the source
code) how you get the URLs.

**Comment by kchawla-pi** [comment]
@emdupre Yes. Could you describe this in more detail? 
Good enough that a new developer on the project will understand this?
Also, what about a 301 request?

@GaelVaroquaux 

> Hmm. It looks like they're you're not actually downloading from OSF, but from the CEA server:

I was under the impression that all data has been moved to OSF servers. We need to figure this out and change this accordingly ASAP.

**Comment by kchawla-pi** [comment]
What do we do for nistats? Changing this here will not help NIstats builds, unless we build using Nilearn master.
I propose we backport this to a Nilearn 0.6.1 release.

**Comment by GaelVaroquaux** [comment]
> I was under the impression that all data has been moved to OSF servers. We need to figure this out and change this accordingly ASAP.

+1

**Comment by GaelVaroquaux** [comment]
> I propose we backport this to a Nilearn 0.6.1 release.

+1

**Comment by emdupre** [comment]
> I was under the impression that all data has been moved to OSF servers. We need to figure this out and change this accordingly ASAP.

@kchawla-pi can you try re-running your commands with this branch ? I'm now getting the correct downloads. 

Meanwhile, I'll think of a helpful comment to add, here.

**Comment by emdupre** [comment]
Checked in on it and can confirm I'm still getting a `ValueError` with that one contrast when using this branch:

```
ValueError: Contrast 'calculation and sentence listening/reading' is not available
```

Everything else seems to be downloading correctly from OSF !

kchawla-pi:
Think that's a documentation bug?

And woohoo!

emdupre:
> Think that's a documentation bug?

Would have to ask the Brainomics folks to be sure... But maybe we can do that in a separate issue ?

kchawla-pi:
Yeah, separate issue.

The only thing missing was a leading slash?

emdupre:
Yeah, it was just a trailing slash, it looks like.

If you look at the [image at the top of this thread](https://user-images.githubusercontent.com/15017191/72743001-cbd3ef80-3b78-11ea-9176-2d4f536e383f.png), you can see the browser hits:

1. the `308` link, then 
2. (because it has access to `REQUEST` headers, which we don't handle in our data fetcher) correctly follows it to the `302` link -- which is the one with the trailing slash, as I've updated to here -- then 
3. to another `302` link that includes the server identifier and project name, then 
4. finally to the correct signed link which returns `200` for `Found`.

So my best guess is that our previous link(s) used to be handled as a `302` but it was deprecated as a `308`. 

I found [this link quite helpful](https://airbrake.io/blog/http-errors/308-permanent-redirect) for understanding all of the different HTTP response codes.

kchawla-pi:
I can download successfully using your PR as well.
Is this all or do you want to change more URLs?

emdupre:
I think this is all we need for the patch. What other URLs were you thinking to change ?

And let me know if you think I should update the code comments !

kchawla-pi:
We can add `Trailing slash was added to the download URLs to ensure a 302 redirect instead of a 308 error.`.

emdupre:
> We can add Trailing slash was added to the download URLs to ensure a 302 redirect instead of a 308 error.

Do you think that would be meaningful to a new developer ? Genuinely not sure.

kchawla-pi:
Okay then let's leave it out for now.

PR #2300, opened 2020-01-29
[MRG+1] [DOC, ENH] Add binder links to documentation

### LINKED ISSUE #2284: Add Binder links to examples
I just noticed that the sklearn documentation includes Binder links (for example, [here](https://scikit-learn.org/stable/auto_examples/linear_model/plot_lasso_and_elasticnet.html#sphx-glr-auto-examples-linear-model-plot-lasso-and-elasticnet-py)). It'd be great to have these in Nilearn as well, so users could run our examples on Binder !

I can open a PR if that's of interest :smile_cat: I'll plan to build off https://github.com/scikit-learn/scikit-learn/pull/11221 and a few updates that it looks like have been [merged since then](https://github.com/scikit-learn/scikit-learn/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+binder).

emdupre:
Closes #2284.
Supersedes #2285.

- Adds binder links to generated sphinx-gallery documentation
- ~Correctly points to individual examples, [building off PlasmaPy example](https://stanczakdominik.github.io/posts/simple-binder-usage-with-sphinx-gallery-through-jupytext/)~
- Points to examples in nilearn.github.io repository, assuming update there


emdupre:
I think this is working !

Check out the Binder links on some of the generated documentation (for example, on the [introduction to decoding](https://5259-1235740-gh.circle-artifacts.com/0/home/circleci/project/doc/_build/html/auto_examples/plot_decoding_tutorial.html#sphx-glr-auto-examples-plot-decoding-tutorial-py)). It hops straight to the appropriate example and it loads in Binder like a notebook ! :tada: 

The remaining issues I see:

1. The note about "click here to download this example" now mentions Binder and can be a little too big for the CSS rules we have set up, eg:

  Screenshot from 2020-01-29 21-41-08

  It doesn't happen for every example, but it's annoying enough that I can try to fix it in this PR.

2. A more general Binder issue I've been hitting lately is that the matplotlib backend seems to be a bit dicey -- sometimes I have to re-run plots to have them show up correctly. I'm not sure if that's something I can fix in this PR but I can search the Jupyter Discourse to see if anyone has ideas. Or, if you have any suggestions, please let me know !

emdupre:
In response to @GaelVaroquaux 's concern that we shouldn't need Jupytext; I re-read the blog post linked above and it highlighted the issue:

> However, I quickly realized that the system in place is a bit unwieldy. Binder takes a link to an existing GitHub repository and executes .ipynb notebooks located there online. However, with sphinx-gallery, we don't have those notebooks in the repository - we have .py files with comments. The currently recommended way of setting this up with sphinx-gallery is keeping your built documentation in another repository and hosting it via something along the lines of GitHub Pages rather than ReadTheDocs, which we are currently using.

So, we could link to the notebooks in the [built Nilearn documentation](https://github.com/nilearn/nilearn.github.io); however, this does not have the expected structure as the generated notebooks are not in an `examples` folder but in a `downloads` folder. The Jupytext solution generates a user-experience that mimics notebooks without needing to link to a (potentially idiosyncratic, as in our case) pre-built site.

Would be curious to know if you see another, preferred path forward.

GaelVaroquaux:
> In response to @GaelVaroquaux 's concern that we shouldn't need Jupytext; I re-read the blog post linked above and it highlighted the issue:

OK, the point being that with jupytext binder is pointed to the nilearn
repo, instead of docs repo.

I see how, for people who use read-the-docs, this would be a problem.
In principle, this should not be a problem for us.

However, one thing with sphinx that is challenging, is the path that it
generates for downloads (and hence for notebooks) is terribly convoluted:
https://github.com/nilearn/nilearn.github.io/tree/master/_downloads/40e44ba4b1561183bd26fcc6aae0173f
That might be a new evolution of sphinx which somewhat broke the
useability of binder with sphinx-gallery.

I suggest to move on with the current solution that uses jupytext in the
context of nilearn, and open an issue on the tracker of sphinx-gallery to
discuss this.

emdupre:
OK, I was able to build the documentation and host it on my fork of Nilearn as discussed in https://github.com/nilearn/nilearn/issues/2284#issuecomment-580926602. You can interact with the [plot_decoding_tutorial here](https://mybinder.org/v2/gh/emdupre/nilearn/gh-pages?filepath=examples/auto_examples/plot_decoding_tutorial.ipynb), and I'm happy to say it links up appropriately :+1: 

I've updated this PR to assume that the documentation on `nilearn.github.io` has been re-generated with the Binder configuration (i.e., with this PR merged). 

I also spent some time with the CSS issue, and the best solution I was able to come up with puts the download link as a banner along the top of the example, similar to how it's [currently done in the sklearn docs](https://scikit-learn.org/stable/auto_examples/plot_roc_curve_visualization_api.html#sphx-glr-auto-examples-plot-roc-curve-visualization-api-py):

Screenshot from 2020-02-02 20-50-14

This is in contrast to the current docs, where the download link sits to the right side. 

LMK what you think !

nilearn/datasets/func.py
```diff
@@ -920,7 +920,10 @@ def fetch_localizer_contrasts(contrasts, n_subjects=None, get_tmaps=False,
 
     # Build data URLs that will be fetched
     files = {}
-    root_url = "https://osf.io/download/{0}"
+    # Download from the relevant OSF project, using hashes generated
+    # from the OSF API. Note the trailing slash. For more info, see:
+    # https://gist.github.com/emdupre/3cb4d564511d495ea6bf89c6a577da74
+    root_url = "https://osf.io/download/{0}/"
```
GaelVaroquaux: Do we have unrelated changes in this PR?

emdupre: Great catch ! All fixed :smile_cat:

requirements-build-docs.txt
```diff
@@ -1,5 +1,4 @@
 numpy
-pandas
```
GaelVaroquaux: I'm surprised to see a removal of pandas for doc building. We are using it in examples, eg
02_decoding/plot_haxby_anova_svm.py

emdupre: Somehow it was listed twice previously -- this only removes the duplication. It's still included in the requirements !

GaelVaroquaux: Ah, the problem of looking only at the diff.

kchawla-pi:
We should merge this before  #2302 and push the updated docs.

emdupre:
I couldn't leave the CSS alone (:see_no_evil:) so I tried to sync it a bit more closely with the general sphinx admonition, such as you can see in the reference documentation:

Screenshot from 2020-02-03 20-41-18

Here's what the download links now look like after https://github.com/nilearn/nilearn/pull/2300/commits/6fa47f74499ce377f5421e185c8d7fb0f1ed4392:

Screenshot from 2020-02-03 20-41-06

I think this is pretty unobjectionable, but LMK if there are any other suggestions !

emdupre:
> We should merge this before #2302 and push the updated docs.

#2302 would enhance the experience with Binder as the MPL back-end would load correctly -- would it be possible to include it in the updated docs ? Or do you think we should wait until the next release ?

kchawla-pi:
I think you are right, we can release this PR and #2302 in 0.6.2 
BTW, what happens if someone tries to run the Binder examples with Nilearn < 0.6.1 ? Is that even relevant?

kchawla-pi:
You should add this in whats_new, under 0.6.2

emdupre:
> You should add this in whats_new, under 0.6.2

Want to merge #2302 (if you're also +1 on it as-is), I'll pull from master and add to the 'what's new' @GaelVaroquaux started there, and we can cut the docs / release then ?

> BTW, what happens if someone tries to run the Binder examples with Nilearn < 0.6.1 ? Is that even relevant?

It'd be like running the docs with that version of Nilearn locally -- so I imagine for lower versions you'd see some hiccups. We just need to update our doc generating process to include the Binder links, which this PR does.

kchawla-pi:
I think one of the comments is now misplaced in #2302 . I can fix that tomorrow if Gael does not and then merge it.

kchawla-pi:
If you wish, you can add your whats_new entry in 0.6.2 here, and you or I can resolve the merge conflict before merging.

emdupre:
> If you wish, you can add your whats_new entry in 0.6.2 here, and you or I can resolve the merge conflict before merging.

Done in https://github.com/nilearn/nilearn/pull/2300/commits/78f229915b2f74fe20d451d0f5ac7f6f39c75cab :rocket:

PR #2302, opened 2020-01-31
Fix "%matplotlib inline" in notebook

GaelVaroquaux:
Before this, importing nilearn.plotting after running "%matplotlib inline" would reset the backend. Thanks @emdupre for spotting it.

I checked interactively with jupyter that it would fix the problem, but I don't think that we can do a test.

### DIFF
--- a/nilearn/plotting/__init__.py
+++ b/nilearn/plotting/__init__.py
@@ -25,10 +25,13 @@ def _set_mpl_backend():
         # that the version is greater that the minimum required one
         _import_module_with_version_check('matplotlib',
                                           OPTIONAL_MATPLOTLIB_MIN_VERSION)
+        current_backend = matplotlib.get_backend().lower()
         # Set the backend to a non-interactive one for unices without X
+        if 'inline' in current_backend or 'nbagg' in current_backend:
+            return
         if (os.name == 'posix' and 'DISPLAY' not in os.environ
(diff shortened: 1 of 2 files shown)


emdupre:
I also can't think of a great test for this, but I confirmed that it works with [hosting a notebook on GitHub and using your PR](https://github.com/emdupre/nilearn/blob/enh/mpl-test/examples/plot_decoding_tutorial.ipynb) !

So, I'm +1 for merging, thanks !

nilearn/plotting/__init__.py
```diff
@@ -25,10 +25,13 @@ def _set_mpl_backend():
         # that the version is greater that the minimum required one
         _import_module_with_version_check('matplotlib',
                                           OPTIONAL_MATPLOTLIB_MIN_VERSION)
+        current_backend = matplotlib.get_backend().lower()
         # Set the backend to a non-interactive one for unices without X
```
kchawla-pi: We should move this comment to the correct new position.

emdupre:
From Travis, it looks like the majority of errors are transient HTTP errors. One real flake8 error, though:

```
nilearn/plotting/__init__.py:29:1: W293 blank line contains whitespace
```

GaelVaroquaux:
Thank you Kshitij!!

PR #2299, opened 2020-01-29
Nilearn.stats - 1: Incorporate Nistats into Nilearn with former's commit history intact

kchawla-pi:
As per discussion in issue #2298

PR #2297, opened 2020-01-27
Nilearn.stats 1: Import Nistats repo with commit history into Nilearn

kchawla-pi:

PR #2304, opened 2020-02-04
Nilearn.stats - 2: Reorganize functional code from Nistats into Nilearn

kchawla-pi:
Replaces PR #2297 
Branched from #2299 . Must be pulled from this to update for master. Must be merged after this.

Reorganizing Nistats code after importing the Nistats repo in #2299


kchawla-pi:
When I try to move `nilearn.stats.reporting` into `nilearn.reporting`, there are a lot of circular imports.

1. Getting import error with `input_data.NiftiMasker`, because Nistats' `nilearn/stats/glm_reporter.py` imports it, and `nilearn/input_data/nifti_masker.py` imports `reporting.ReportMixin`.

2. Getting import error with `reporting.HTMLDocument`, because `glm_reporter.py` imports from `plotting` and `plotting` imports from `reporting`.

There doesn't seem to be a clean way yet of managing this. I'm working on it. Suggestions welcome.
@emdupre

kchawla-pi:
Think I have a solution.

kchawla-pi:
I fixed the `input_data.NiftiMasker` import error by moving it into the local functions. There are a total of 3 places this has been done.
1 in `stats/contrasts.py`. 2 in `stats/thresholding.py.`
If anyone has suggestions I am happy to receive them to do this better, or we can figure it out later and improve this in a future PR.

bthirion:
No better suggestion. We might rework this later.
Best,

jeromedockes:
> If anyone has suggestions I am happy to receive them to do this better, or we can figure it out later and improve this in a future PR

the problem comes from the fact that a core module like niftimasker is importing
the reporting module (see the niftimasker reporting PR for discussion), which
did not result in errors when the reporting module was almost empty but is a
bigger problem now that it depends on the nistats reporting, which has a
somewhat different approach and does more image manipulation (while the nilearn
reporting only manipulates html templates).

+1 for saving this for a (near) future PR, but a solution would be to remove the
unnecessary ReportMixin which so far only serves the niftimasker reports anyway
and move this code into the niftimasker's `generate_report` function. of course,
importing utilities from the reporting module, hidden inside the
`generate_report` method, is fine.

in short: your solution works but in the long run the module-level import that
really needs to disappear is the niftimasker's
`from nilearn.reporting import ReportMixin`.

kchawla-pi:
Thanks @jeromedockes . i figured it was  `from nilearn.reporting import ReportMixin` which was at the root, but figured it is best dealt with in another PR.
Your explanation indicates this problem can be solved rather easily. I'll see if it is feasible to accommodate it quickly after this one is complete, or perhaps within this one.

jeromedockes:
to clarify a bit:

 - in nilearn, the reporting module is mostly an interface and some utilities
   for manipulating html documents -- no image manipulation involved.
   niftimasker uses this, but the idea is that only objects such as niftimasker
   can know what their reports should look like, so the bulk of the work is done
   by the reported objects themselves.

 - in nistats, the firstlevelmodel is unaware of reporting. the reporting
   function takes it as an argument and uses it as input to perform all the
   necessary computations and produce the report. hence reporting needs to
   import other image processing modules.

 - both approaches make sense but despite having the same name the "reporting"
   packages of nilearn and nistats do very different things and this difference
   explains why you ran into trouble. to mimick nilearn, the bulk of nistats
   reporting would have to go into the glm reporter modules, and the
   firstlevelmodel would need to expose a `generate_report` method. or to mimick
   the nistats GLMReporter, the niftimasker's reporting code would have to go
   into the reporting package. either change requires a discussion and work, but
   in the meanwhile replacing the reportermixin with functions imported inside
   the niftimasker's `generate_report`, and removing the module-level import,
   will work around the `ImportError`.

emdupre:
> We might rework this later.

Strong +1 from me on saving for another PR; I think the solution @jeromedockes proposes above is a fine approach for now until we can revisit this, although it sounds like you had also found a short-term solution, @kchawla-pi ? Would be happy to see either one in this PR.

Should we open an issue to discuss long-term consolidation of the two reporting "styles" ? Obviously I'm lightly partial to the Nilearn-style reporting (:wink:), but would love to discuss which path makes sense before a more substantial remapping.

bthirion:
I think I agree with the plan. And i'm also fine with changing  Nilearn.stats objects toward more Nilearn-style structure.

kchawla-pi:
> i'm also fine with changing Nilearn.stats objects toward more Nilearn-style structure

When we were building the glm reporter, you had objected to the idea of a `FirstLevelModel.report()` style .

bthirion:
Indeed, but I'm not sure why anymore...

kchawla-pi:
This is pretty much ready for a first review.
I'm hoping it will be relatively quick; reviewing it the way we do most PRs will be infeasible due to the number of changes, but since this is essentially Nistats code, ported over, and that one worked fine, so should this. 
Meanwhile, I'll start reorganizing the examples in a new PR branched off this one.
@bthirion @jeromedockes @GaelVaroquaux @emdupre

kchawla-pi:
There are Flake8 errors. Most of them are in the nistats/examples/ files, which won't be fixed.
Some are in the nilearn/stats/ code, which I will.

nistats/.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,30 @@
+---
```
bthirion: This one should be removed I guess

nistats/.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,40 @@
+---
```
bthirion: This one should be removed I guess ?

nistats/LICENSE
```diff
@@ -0,0 +1,28 @@
+Copyright (c) 2015, nistats
```
bthirion: This one should be removed I guess

nistats/azure-pipelines.yml
```diff
@@ -0,0 +1,41 @@
+# Python package
```
bthirion: This one should be removed I guess

nistats/.circleci/README.md
```diff
@@ -0,0 +1,14 @@
+CircleCI is used to build our documentations and tutorial examples.
```
bthirion: All that follows will be removed later I guess ?

nistats/.github/ISSUE_TEMPLATE/feature_request.md
kchawla-pi: The entire nistats directory will be removed in the 3rd PR which reorganizes the examples, so not going to worry about anything in that dir.

kchawla-pi:
@bthirion 
> All that follows will be removed later I guess ?

Yeah, the entire nistats directory will be removed in the 3rd PR which reorganizes the examples, so not going to worry about anything in that dir.

kchawla-pi:
Off Github discussion, I raised the following question:

some submodules have a 2-3 functions only and most of them are not worth making public facing I think. On the other hand, some, like hemodynamic_response, models, regression, have either lots of public facing, useful functions, or do not fit in as standalone functions for direct import.

And should we hide nilearn.stats.utils?

@bthirion suggested:

Let me suggest

    nilearn.stats.hemodynamic_models: Hemodynamic Models
        move to nilearn.first_level_model

    nilearn.stats.design_matrix: Design Matrix Creation
        move to nilearn.first_level_model

    nilearn.stats.experimental_paradigm: Experimental Paradigm
        move to nilearn.first_level_model

    nilearn.stats.model: Statistical models
        move to nilearn.stats

    nilearn.stats.regression: Regression Models
        move to nilearn.stats

    nilearn.stats.first_level_model: First Level Model
        move to nilearn.first_level_model

    nilearn.stats.second_level_model: Second Level Model

    nilearn.stats.contrasts: Contrasts
        move to nilearn.stats

    nilearn.stats.thresholding: Thresholding Maps
        move to nilearn.stats

    nilearn.reporting: Report plotting functions

    nilearn.stats.utils: Utility functions
        move to nilearn._utils

kchawla-pi:
@bthirion I'm putting `make_second_level_design_matrix` in  `second_level_model`.

bthirion:
Many thx !

PR #2306, opened 2020-02-11
[DOC] updated reference Seitzman 2018

fliem:
The preprint referenced in the docs is now published as a paper. I have added the paper reference, but have left the reference to the preprint in so that the year in the function name still makes sense.


jeromedockes:
thanks @fliem ! could you rebase on master? it should fix the CI failure.
LGTM otherwise

kchawla-pi:
Pull from master, Jerome :-P No rebasing here.

kchawla-pi:
Should we deprecate the 2018 name and replace it with the 2020 name for the fetcher function? Most people will in due course become familiar with the paper not the preprint.

PR #2308, opened 2020-02-13
Deprecates fetch_nyu_rest

### LINKED ISSUE #2020: Deprecate fetch_nyu_rest
This dataset is not preprocessed, and it leads to people doing mistakes. We don't use it in our examples. We should deprecate it.

Related to #330

jbteves:
Closes #2020


jbteves:
I attempted to build the docs locally, but here's the tail end of the (very long and unfortunately failed) `make html-strict` call:
```
Downloaded 1720320 of 13947990 bytes (12.3%,    7.5s remaining)
Downloaded 13947990 of 13947990 bytes (100.0%,    0.0s remaining) ...done. (2 seconds, 0 min)
Extracting data from /home/jbteves/nilearn_data/fsaverage/fsaverage.tar.gz..... done.
Output from /home/jbteves/nilearn/examples/01_plotting/plot_haxby_masks.pys.py  
First subject anatomical nifti image (3D) is at: /home/jbteves/nilearn_data/haxby2001/subj2/anat.nii.gz
First subject functional nifti image (4D) is at: /home/jbteves/nilearn_data/haxby2001/subj2/bold.nii.gz
Makefile:155: recipe for target 'check' failed
make[1]: *** [check] Segmentation fault (core dumped)
make[1]: Leaving directory '/home/jbteves/nilearn/doc'
Makefile:66: recipe for target 'html-strict' failed
make: *** [html-strict] Error 2
```

emdupre:
Thanks for this, @jbteves ! :rocket: 

We might not want to delete the code outright, but instead have a deprecation cycle for it. Since we're currently at the 0.6.x release series, we could add a `DeprecationWarning` for the 0.7.x series. Since we usually keep things around for 1 or two releases after the deprecation announcement, we could either remove it in the 0.8.x or 0.9.x release series.

For implementing a `DeprecationWarning`, you could look at the warnings I removed in https://github.com/nilearn/nilearn/pull/2279/

kchawla-pi:
We can use VisibleDeprecationWarning from Numpy or FutureWarning from
standard Python

jbteves:
I'll try to integrate that deprecation warning. I'm confused though, is `codecov/project` failing because the coverage decreases?

kchawla-pi:
CodeCov is misreporting coverage with certain files. You can ignore this failure.

jbteves:
So I just added the deprecation warning from numpy since that's what's in @emdupre's example, but it looks like DeprecationWarning might be nice because then you can check that it's triggered correctly in pytest.

emdupre:
I think what you have is fine ! Especially since it's already used in that file.

You might also want to add a `NOTE` admonition to the function docstring, like [here](https://github.com/nilearn/nilearn/pull/2279/files#diff-472cc8a2f83316878a7f02c131580611L26).

In https://github.com/nilearn/nilearn/pull/2256 we decided it wasn't worth adding a test for an untested function just to check that the deprecation warning is raised. Would be great to get feedback from other devs on whether we're still ok with that approach.

kchawla-pi:
@emdupre 
In #2256 we're removing the function with 0.7.0, which would be around in a few months, so not testing that the warning was raised was acceptable. This one will probably hang around for a year, so we should test that the warning is being raised.

@jbteves 
You should be able to test `VisibleDeprecationWarning` like any other warning using `with pytest.warns(VisibleDeprecationWarning):` and so on. You will just have to import the warning in the test file first.

If you decide to not use that warning then use `FutureWarning`, not `DeprecationWarning`.

jbteves:
Oops, sorry for letting this get away from me! I think the test works correctly now, I tried it with the standard DeprecationWarning to make sure it failed in that case and it does. I hope 8ddbc0a is up to spec.

emdupre:
Thanks, @jbteves !! If you could add the `NOTE:` admonition about [here](https://github.com/nilearn/nilearn/pull/2308/files#diff-472cc8a2f83316878a7f02c131580611L252), that's the only other thing I'd suggest.

Looking forward to having this merged :sparkles:

GaelVaroquaux:
Thank you very much!!

I would be in favor of giving a version number by which we remove. Should we say 0.8? We could also say "0.8 or by 2021", if we fear slipping by in releases.

kchawla-pi:
0.8 seems fins to me, AFAIK we haven't yet slipped in any release. It's always been 1 per year.

After this PR is merged, we can release 0.6.2. WDYT?

kchawla-pi:
We need a whats_new entry.

jbteves:
Do you want me to make it 0.8 then?

kchawla-pi:
Yeah. Make it 0.8 and add a whats_new entry. I think that's all that's needed.

jbteves:
I'm sorry, I don't understand about the whats_new entry, where should I add that?

kchawla-pi:
In the `doc/whats_new.rst` file, under Section 0.6.2, add the sub-section 
```rst
Changes
----------

 - Deprecated atlas func:`nilearn.datasets.func.fetch_nyu_rest` wil be removed in Nilearn 0.8.0 .
```

GaelVaroquaux:
> After this PR is merged, we can release 0.6.2. WDYT?

Yes!!!!

kchawla-pi:
Hey @emdupre I was about to merge this, when I noticed you self-requested a review. Any concerns?

nilearn/datasets/func.py
```diff
@@ -250,6 +250,9 @@ def fetch_nyu_rest(n_subjects=None, sessions=[1], data_dir=None, resume=True,
                    verbose=1):
     """Download and loads the NYU resting-state test-retest dataset.
 
+    NOTE: This function is deprecated and will be removed in a future
```
emdupre:
```suggestion
    NOTE: This function is deprecated and will be removed in the 0.8.x
```

nilearn/datasets/func.py
```diff
@@ -325,6 +328,10 @@ def fetch_nyu_rest(n_subjects=None, sessions=[1], data_dir=None, resume=True,
           F.X. Castellanos, M.P. Milham
 
     """
+    warnings.warn("fetch_nyu_rest has been deprecated and will "
+                  "be removed in a future release.",
```
emdupre:
```suggestion
                  "be removed in the 0.8.x release.",
```

PR #2317, opened 2020-02-26
Import ABC from collections.abc instead of collections for Python 3.9 compatibility.

tirkarthi:


tirkarthi:
This unblocks https://bugzilla.redhat.com/show_bug.cgi?id=1799941

tirkarthi:
Yes, this is compatible from 3.4 and above.

GaelVaroquaux:
@bthirion : we have Python 3.5 in our tests (the good old rule is: "if it ain't test, it's broken")

tirkarthi:
Thank you for review and merge.

PR #2322, opened 2020-02-28
Sphinx should only build changed files during doc building

kchawla-pi:
Added code to find and build changed files only for docs


examples/02_decoding/plot_haxby_anova_svm.py
```diff
@@ -3,7 +3,7 @@
 ===============================================================
 
 This example does a simple <JUNK TET FOR TESTING TO BE REMOVED LATER>
-but efficient decoding on the Haxby dataset:
+JUNK but efficient decoding on the Haxby dataset:
```
emdupre: Is this how you're testing the CI infrastructure ? :rofl: I'm a little nervous this will get left here, so adding a comment just to flag it :smile_cat:

kchawla-pi: Haha Yeah but I will go through the changed files and remove them, you adding a comment absolutely adds another event in the memory, making it unlikely I''ll forget this. 😄

GaelVaroquaux:
If I need to do something this, I add and remove an empty line: two
consecutive empty lines are bad, but not too bad.

kchawla-pi:
Anyone has ideas why I'm getting the error:

```python
Warning, treated as error:
/home/circleci/project/doc/building_blocks/manual_pipeline.rst::image file not readable: building_blocks/../auto_examples/images/sphx_glr_plot_decoding_tutorial_002.png
```

Interestingly, this is the same file that @tbng just changed in his PR #2323 . It hasn't been merged so this could be a coincidence.

tbng:
> Anyone has ideas why I'm getting the error:
> 
> ```python
> Warning, treated as error:
> /home/circleci/project/doc/building_blocks/manual_pipeline.rst::image file not readable: building_blocks/../auto_examples/images/sphx_glr_plot_decoding_tutorial_002.png
> ```
> 
> Interestingly, this is the same file that @tbng just changed in his PR #2323 . It hasn't been merged so this could be a coincidence.

This leaves an error following change in `plot_decoding_tutorial.py` on plotting weight images.
In particular I changed from using `plot_stat_map` to `plotting.view_img`. The first one will output a static `.png` image while the second one an interactive `.html` which makes the example seems more attractive. This however make references to the old `.png` image broken.

kchawla-pi:
Yeah but it should not affect a different branch, right?

kchawla-pi:
I SSHed into the build and there are no files in the `auto_examples/images` directory, only another directory `thumb` with images.

kchawla-pi:
Aah I think I get it. The examples haven't been run so there is nothing to insert maybe. I'll add a feature to inherit mater's cache perhaps.

tbng:
> Aah I think I get it. The examples haven't been run so there is nothing to insert maybe. I'll add a feature to inherit mater's cache perhaps.

Is this a feature to save build time (quick-build..) ? I thought circleCI will have sphinx automatically rebuild examples for every commit pushed?

kchawla-pi:
Yeah but right now it's either a full 1.5 hour long build or a reusing the cached build (which does not reflect changes properly). This PR tries to figure out the files that have changed and rebuild those properly while keeping the rest same.

bthirion:
Hi,
What is the status of that PR ?

kchawla-pi:
Currently stalled.
I just don't have the time to attend to this.

Kshitij

bthirion:
I can understand ! Thx @kchawla-pi

ymzayek:
As discussed with @NicolasGensollen this has already been done. See [build_type.sh](https://github.com/nilearn/nilearn/blob/main/build_tools/circle/build_type.sh). Closing now.

PR #2325, opened 2020-03-02
CI: Fix name

larsoner:


larsoner:
From what I understand, it not having any effect (yet) is expected here in this PR, it will only work/use these changes once merged into `master`.

Also it might be a good idea to enable debugging logging for the action by setting the Actions secret `ACTIONS_STEP_DEBUG=true`, it will produce a more verbose output while running, i.e., [pages like this](https://github.com/nilearn/nilearn/runs/479576981?check_suite_focus=true) will actually have logging messages from `circleci-artifacts-redirector-action`.

GaelVaroquaux:
Thanks @larsoner !!! Merging this to see if it works

PR #2327, opened 2020-03-05
[ENH, MRG] fREM 

thomasbazeille:
Following the merge of #2000, here is the introduction of fREMClassifier and fREMRegressor objects which run pipelines with clustering, feature selection and ensembling of best models across a grid of parameters.

- The files changes are minor.
- The current implementation yields results as expected on an exemple (see attached below on plot_haxby_tutorial)
- But the test of fREM accuracy on small datasets in test_decoder.py keep failing just because accuracy of fREM is not good enough in this setting but after trying various parameters I don't know which tests to do that would be better suited to this usecase. Anybody ? 
- If I remember correctly you wanted to replace Decoder use by fREM in many examples to alleviate their computational cost @GaelVaroquaux . Which one are the main targets ? I will benchmark how many time we could gain but on ROIs it doesn't seem very useful to cluster. (e.g. it slows things down on Haxby ROI example)

Capture d’écran 2020-03-05 à 18 40 24


GaelVaroquaux:
I'm so excited!!

The goal was to replace the SpaceNet, which is slow. Ideally, we'd keep only one example with the SpaceNet.

thomasbazeille:
@GaelVaroquaux @bthirion 

There are three examples using SpaceNet classifier or regressor :

1. Classification : [SpaceNet on Jimura et al “mixed gambles” dataset ](https://nilearn.github.io/auto_examples/02_decoding/plot_mixed_gambles_space_net.html#sphx-glr-auto-examples-02-decoding-plot-mixed-gambles-space-net-py)(7 minutes) focused on segmenting power (no showcasing of performance increase)
2. Classification : [Decoding with SpaceNet: face vs house object recognition](https://nilearn.github.io/auto_examples/02_decoding/plot_haxby_space_net.html#sphx-glr-auto-examples-02-decoding-plot-haxby-space-net-py) (6 minutes) showcasing decoding accuracy that is improved compared to ANOVA-SVM
3. Regression  :  [Voxel-Based Morphometry on Oasis dataset with Space-Net prior](https://nilearn.github.io/auto_examples/02_decoding/plot_oasis_vbm_space_net.html#sphx-glr-auto-examples-02-decoding-plot-oasis-vbm-space-net-py) (3 minutes 30)  predicting age.

It seems that all 3 examples seem to be adaptations of examples that exist otherwise, and are made only to showcase SpaceNet improvement ability compared to classical pipeline:
- in sparsity for TV-l1 for example 1
- in classification accuracy improvement for example 2
- in a regression setting for example 3

All are mentioned in the [userguide](https://nilearn.github.io/decoding/space_net.html#space-net) except the last one, which measures prediction error with different metric from the [original example](https://nilearn.github.io/auto_examples/02_decoding/plot_oasis_vbm.html#sphx-glr-auto-examples-02-decoding-plot-oasis-vbm-py) and thus makes comparison hard.

So I guess the real question is more which example we want to keep to showcase SpaceNet. Maybe the first one is the best but it's also the longest. Then all others are adaptable to showcase fREM, but maybe not the best ones.

I can open a specific issue if it's not clear enough to discuss it here.

GaelVaroquaux:
> So I guess the real question is more which example we want to keep to showcase
> SpaceNet. Maybe the first one is the best but it's also the longest.

The first one is the clearest. The Jimura dataset is a very hard one.

I would add fREM to this example, and discuss the compared runtimes.

> Then all others are adaptable to showcase fREM, but maybe not the best ones.

I'd be quite happy if we could showcase fREM with the others.

tbng:
I can help reviewing this once it's ready

GaelVaroquaux:
@tbng : help in review will be very useful!

kchawla-pi:
DO NOT Pull from master. There is some stuff that is going and until it has been attended to, do NOT pull from master.

kchawla-pi:
Master is good now. All clear. You can go ahead and pull from it.

kchawla-pi:
@thomasbazeille You have a timeframe in mind for this? It would be good to incorporate it into an alpha release very soon.

nilearn/decoding/decoder.py
```diff
@@ -133,24 +138,36 @@ def _check_estimator(estimator):
 
 
 def _parallel_fit(estimator, X, y, train, test, param_grid, is_classification,
-                  scorer, mask_img, class_index, screening_percentile=100):
+                  scorer, mask_img, class_index, screening_percentile=100,
```
bthirion: I think that we do not provide default arguments. in private functions

nilearn/decoding/decoder.py
```diff
@@ -223,6 +243,14 @@ class _BaseDecoder(LinearModel, RegressorMixin, CacheMixin):
         or have no effect. See scikit-learn documentation for more information,
         for example: https://scikit-learn.org/stable/modules/grid_search.html
 
+    do_clustering: boolean. Default False
+        Whether to do a ReNA clustering as first step of fit to agglomerate similar
+        features together and reduce their overall number.
+
+    clustering_percentile: int, float, optional, in the [0, 100] Default: 10.
+        Percentile of features to keep if clustering is performed.
+        ReNA is typically efficient at reducingthe number of features by 10.
```
bthirion:
```suggestion
        ReNA is typically efficient at reducing the number of features by a factor of 10.
```

thomasbazeille:
> @thomasbazeille You have a timeframe in mind for this? It would be good to incorporate it into an alpha release very soon.

No, I'm working regularly on it, I can make it a priority but I have some things unclear and ill-diagnosed so there's a chance that I'll need some time to debug tests and weird cases.

kchawla-pi:
Ah! Push it here, the others can chip in & help out.

thomasbazeille:
@GaelVaroquaux @bthirion After a lot of testing, I think that fREM weird behaviour on Oasis VBM age prediction is not a bug, just a hyperparameter setting problem (where default performs really poorly).

If I set clustering around 20% and screening around 1%, the MSE can decrease (from 10.5 to 9) but with a large standard deviation increase (60 years!). So it indeed does better but predicts very large and negative ages in some cases.

This is a problem to set the right parameter for the example but more importantly for giving guidelines. So I'm trying different set of parameters right now to find one that is suitable for the example but it's a bit worrying.

bthirion:
I'm afradi OASIS is not a very ggod case for FREM: effects are almost everywhere...
What if you disable screening an use a ridge regression, with n_clusters = .2 n_voxels ?

bthirion:
Useless to say, we don't want to put forward a model with high variance.

thomasbazeille:
@bthirion No it seems that the same thing is happening with ridge.

Everything works fine with clustering_percentile =.99 so I think it's just the clustering that is causing that.

So should I drop this example altogether or just leave it with Space Net @GaelVaroquaux ?

thomasbazeille:
also what worries me then is that we have no other example of regression to test it on and no results in the paper either.

bthirion:
Can you quickly explore how the solution degrades when changing `clustering_percentile` from 0.99 to 0.2 ?
Also, it's fine if we use FReM only for classification. Did I understand your point correctly ?

thomasbazeille:
> Can you quickly explore how the solution degrades when changing clustering_percentile from 0.99 to 0.2 ?

Clustering 10 % MSE: 11.43, Age mean(std): 53.292(159), min - max : -217 - 445
Clustering 20 % MSE: 11.38, Age mean(std): 53.274(79), min - max : -84- 260
Clustering 50 % MSE: 13.15, Age mean(std): 53.345(30), min - max : 1.2 - 120.8
Clustering 60 % MSE: 11.93, Age mean(std): 53.260(25), min - max : 10.7 - 111.2
Clustering 80 % MSE: 12.42, Age mean(std): 53.387(19), min - max : 20.9 - 100.4
Clustering 99 % MSE: 11.03 Age mean(std): 53.309(15), min - max : 27.1 - 91.9

Clustering is inefficient compared to screening (MSE :10.5) and both combined worsen the predictions std.

I included fREM in Jimura example and replaced Space Net with it on Haxby example. 

Should I keep both oasis vbm and Jimura with space net ? Time will not decrease much then ( around 5 minutes) I guess.

bthirion:
> Should I keep both oasis vbm and Jimura with space net ? Time will not decrease much then ( around 5 minutes) I guess.

Maybe only oasis then.

We need to understand why FReM fails on Oasis, but no worries, this will be a job for somebody else ;-)

thomasbazeille:
For completeness let me recap all conclusions I've drawn for now on Oasis example.

I tested four versions : 
- nilearn 0.6 version with sklearn pipeline  : Var Thresholding + ANOVA(1%) + SVR(C=1)
- nilearn 0.6 Space Net (with screening 5%)
- nilearn 0.7 with Decoder SVR bagging
- nilearn 0.7 with fREM. 
I observed that with both Decoder and fREM object the cv param choice for SVR mostly chose  C=100 which cause it's high variance results.

0.6 with Var Thresholding + ANOVA(1%) + SVR(C=1)    |       MAE 8.9    Age 43(18)
0.6 without Var Thresholding                                           |       MAE 10.7  Age 46(19)
0.6 SpaceNet (graphnet)                                                  |       MAE 14.5 Age 53.1(16)
0.7 Decoder                                                                       |       MAE 13.1 Age 53(19)
0.7 fREM without screening (clustering  .8)                    |       MAE 12.9 Age 53.3(18)
0.7 fREM without screening (clustering  .2)                    |       MAE 45 Age 53.3(79)
0.7 fREM (screening .05 clustering  .2)                           |       MAE 32 Age 53.3(65)

So for fREM clustering is really destroying information here at least with SVC heavily regularized. More generally the bagging scheme here is playing against us because of the small train set (60 images)

Also variance thresholding seems very efficient on this dataset as well as screening.




Metric is MAE (average of absolute error of preds against ground truth) on same test set for all pipelines

thomasbazeille:
> > Should I keep both oasis vbm and Jimura with space net ? Time will not decrease much then ( around 5 minutes) I guess.
> 
> Maybe only oasis then.
> 
> We need to understand why FReM fails on Oasis, but no worries, this will be a job for somebody else ;-)

OK it seems fine for me to:
- use Oasis to display graphnet as is.
- use fREM on Haxby
- work on Jimura with fREM, mention it is also possible to use SpaceNet (but without doing it) (TO DO)
Any disagreement @GaelVaroquaux ?


Apart from that, I drafted the PR is ready for review, especially the userguide! @tbng @bthirion

thomasbazeille:
@kchawla-pi did I do smth wrong to have codecov failure ? I opened before Nistats was merged into Nilearn ; then merged master in the PR and the diff seems to come from there.

kchawla-pi:
I think CircleCI is failing because you have not imported the new classes in `decoder.py` into `decoding/__init__.py` and added to the `__all__` list as strings.

kchawla-pi:
CodeCov is lower. The log shows me that 2 scenarios are not tested.

The 2 `do_screening` scenarios are not tested.
When `n_final_features < 50`, are you testing that the warning is raised correctly.

nilearn/decoding/decoder.py
```diff
@@ -1,43 +1,48 @@
...
 
 import numpy as np
+from joblib import Parallel, delayed
 from sklearn import clone
 from sklearn.base import RegressorMixin
 from sklearn.linear_model import LogisticRegression
 from sklearn.linear_model.base import LinearModel
 from sklearn.linear_model.ridge import RidgeCV, RidgeClassifierCV, _BaseRidgeCV
-from sklearn.model_selection import ParameterGrid, check_cv, LeaveOneGroupOut
+from sklearn.model_selection import ParameterGrid, check_cv, LeaveOneGroupOut, StratifiedShuffleSplit, ShuffleSplit
```
tbng, line 26: Use parentheses for import line longer than 80 chars 
```suggestion
from sklearn.model_selection import (LeaveOneGroupOut, ParameterGrid,
                                     ShuffleSplit, StratifiedShuffleSplit,
                                     check_cv)
```

nilearn/decoding/decoder.py
```diff
@@ -1,43 +1,48 @@
...
-from .._utils.cache_mixin import _check_memory
-from .._utils.param_validation import (_adjust_screening_percentile,
-                                       check_feature_screening)
-from ..input_data.masker_validation import check_embedded_nifti_masker
+from nilearn._utils import CacheMixin
+from nilearn._utils.cache_mixin import _check_memory
+from nilearn._utils.param_validation import (_adjust_screening_percentile,
+                                             check_feature_screening)
+from nilearn.input_data.masker_validation import check_embedded_nifti_masker
+from nilearn.regions.rena_clustering import ReNA
```
tbng, line 44: These lines should be relatve import instead of absolute 
```suggestion
from .._utils import CacheMixin
from .._utils.cache_mixin import _check_memory
from .._utils.param_validation import (_adjust_screening_percentile,
                                                            check_feature_screening)
from ..input_data.masker_validation import check_embedded_nifti_masker
from ..regions.rena_clustering import ReNA
```

nilearn/decoding/decoder.py
```diff
@@ -175,9 +192,12 @@ def _parallel_fit(estimator, X, y, train, test, param_grid, is_classification,
             best_intercept = estimator.intercept_
             best_param = param
 
-    if (selector is not None) and do_screening:
+    if do_screening:
         best_coef = selector.inverse_transform(best_coef)
```
tbng, line 196: This is the part where I changed in PR #2362 (related to issue #2360). I guess maybe we could change this directly in this PR?

nilearn/decoding/decoder.py
```diff
@@ -223,6 +243,14 @@ class _BaseDecoder(LinearModel, RegressorMixin, CacheMixin):
         or have no effect. See scikit-learn documentation for more information,
         for example: https://scikit-learn.org/stable/modules/grid_search.html
 
+    do_clustering: boolean. Default False
+        Whether to do a ReNA clustering as first step of fit to agglomerate similar
+        features together and reduce their overall number.
+
+    clustering_percentile: int, float, optional, in the [0, 100] Default: 10.
+        Percentile of features to keep if clustering is performed.
+        ReNA is typically efficient at reducing the number of features by 10.
```
tbng, line 252: I guess here you mean 10 folds @thomasbazeille ?

nilearn/decoding/decoder.py
```diff
@@ -445,13 +478,23 @@ def fit(self, X, y, groups=None):
...
             self.screening_percentile, self.mask_img_, verbose=self.verbose)
 
+        if self.do_clustering:
+            n_final_features = X.shape[1] * self.screening_percentile_ * \
+                self.clustering_percentile / 10000
+            if n_final_features < 50:
+                warnings.warn("After clustering and screening, the decoding model will " +
+                              "be trained only on {} features. ".format(n_final_features) +
+                              "You should consider raising clustering_percentile or " +
+                              "screening_percentile parameters", UserWarning)
```
tbng, line 488: Long line
```suggestion
                warnings.warn(
                    "After clustering and screening, the decoding model will "
                    "be trained only on {} features. ".format(n_final_features) +
                    "You should consider raising clustering_percentile or "
                    "screening_percentile parameters", UserWarning)
```

nilearn/decoding/decoder.py
```diff
@@ -876,3 +921,301 @@ def __init__(self, estimator='svr', mask=None, cv=10, param_grid=None,
             mask_strategy=mask_strategy, memory=memory,
             is_classification=False, memory_level=memory_level,
             verbose=verbose, n_jobs=n_jobs)
+
+
+class fREMRegressor(_BaseDecoder):
```
tbng, line 926: Just for the naming convention I thought class should be CapName? (e.g. in https://www.python.org/dev/peps/pep-0008/#id41)

nilearn/decoding/decoder.py
```diff
@@ -1,43 +1,48 @@
 """High-level decoding object that exposes standard classification and
 regression strategies such as SVM, LogisticRegression and Ridge, with optional
-feature selection, integrated hyper-parameter selection
-and aggregation strategy in which the best models
-within a cross validation loop are averaged.
+feature selection, integrated hyper-parameter selection and aggregation strategy
```
tbng, line 3: Line limit should be 79 not 80, see: https://www.python.org/dev/peps/pep-0008/#id19

nilearn/decoding/decoder.py
```diff
@@ -133,24 +138,36 @@ def _check_estimator(estimator):
...
     X_test, y_test = X[test], y[test]
 
-    if (selector is not None) and do_screening:
+    # for fREM Classifier and Regressor : start by doing a quick ReNA clustering
+    # to reduce the number of feature by agglomerating similar ones
+
+    if do_clustering and (clustering_percentile < 100):
+        n_clusters = int(X_train.shape[1] * clustering_percentile / 100.)
+        clustering = ReNA(mask_img, n_clusters=n_clusters, n_iter=20,
+                          threshold=1e-7, scaling=False)
```
tbng, line 159: Is there any specific reason to hardcoded `n_iter`, `threshold` and `scaling` inside `_parallel_fit`? For example users might want to customize these parameters.

nilearn/decoding/tests/test_decoder.py
```diff
@@ -143,7 +144,7 @@ def test_parallel_fit():
                                           is_classification=False,
                                           scorer=scorer,
                                           mask_img=None, class_index=1,
-                                          screening_percentile=None)))
+                                          screening_percentile=None, do_clustering=False, clustering_percentile=100)))
```
tbng, line 147: Long line
```suggestion
        outputs.append(list(_parallel_fit(estimator=estimator, X=X, y=y,
                                          train=train, test=test,
                                          param_grid=param_grid,
                                          is_classification=False,
                                          scorer=scorer, mask_img=None,
                                          class_index=1,
                                          screening_percentile=None,
                                          do_clustering=False,
                                          clustering_percentile=100)))
```

nilearn/decoding/decoder.py
```diff
@@ -876,3 +921,301 @@ def __init__(self, estimator='svr', mask=None, cv=10, param_grid=None,
...
+        in NeuroImage, Elsevier, 2017  pp.1-16, 11 October 2017.
+        https://hal.archives-ouvertes.fr/hal-01615015/
+
+    See Also
+    ------------
+    nilearn.decoding.DecoderRegressor: Regression strategies for Neuroimaging,
+    nilearn.decoding.fREMClassifier: State of the art classification pipeline for Neuroimaging
+    """
+
+    def __init__(self, estimator='svr', mask=None, cv=30, param_grid=None,
```
tbng, line 1053: Is there any reason why cv should be set to a quite relatively large number? This would slow the run time a lot I think.

Edit: Also I see both in the example & test file cv is manually set to cv=10.

nilearn/decoding/decoder.py
kchawla-pi: The revised PEP8 guidelines stimulate absolute imports . Relative imports are warranted only if the absolute import is too verbose.

kchawla-pi:
Nice review @tbng :~)

jeromedockes:
You need to edit nilearn.decoding.__init__.py to import the new classes
```python
"""
Decoding tools and algorithms.
"""


from .searchlight import SearchLight
from .space_net import SpaceNetClassifier, SpaceNetRegressor
from .decoder import Decoder, DecoderRegressor, fREMRegressor, fREMClassifier


__all__ = ['SearchLight', 'SpaceNetClassifier', 'SpaceNetRegressor', 'Decoder',
           'DecoderRegressor', 'fREMClassifier', 'fREMRegressor']

```

doc/decoding/frem.rst
```diff
@@ -0,0 +1,86 @@
+.. for doctests to run, we need to define variables that are define in
+   the literal includes
+    >>> import numpy as np
+    >>> from sklearn import datasets
+    >>> iris = datasets.load_iris()
+    >>> fmri_masked  = iris.data
+    >>> target = iris.target
+    >>> session = np.ones_like(target)
+    >>> n_samples = len(target)
```
bthirion, line 9: Sorry, this looks completely unrelated ?

examples/02_decoding/plot_haxby_frem.py
```diff
@@ -42,53 +46,32 @@
...
 
 #############################################################################
-# Visualization of Graph-net weights
+# Visualization of fREM weights
 # ------------------------------------
 from nilearn.plotting import plot_stat_map, show
-coef_img = decoder.coef_img_
-plot_stat_map(coef_img, background_img,
-              title="graph-net: accuracy %g%%" % accuracy,
+%matplotlib inline
```
bthirion, line 65: This kind of command should go at the beginning of the script ?

examples/02_decoding/plot_haxby_frem.py
```diff
@@ -42,53 +46,32 @@
...
-
-##############################################################################
-# Now Fit SpaceNet with a TV-l1 penalty
-# --------------------------------------
-decoder = SpaceNetClassifier(memory="nilearn_cache", penalty='tv-l1')
-decoder.fit(X_train, y_train)
-y_pred = decoder.predict(X_test)
-accuracy = (y_pred == y_test).mean() * 100.
-print("TV-l1 classification accuracy : %g%%" % accuracy)
+coef_img.to_filename('haxby_fREM_weights.nii')
```
bthirion, line 69:
```suggestion
coef_img.to_filename('haxby_fREM_weights.nii.gz')
```

bthirion: Or, should this go to some output directory ? (I haven't checked what other examples do in that respect)

examples/02_decoding/plot_haxby_frem.py
```diff
@@ -42,53 +46,32 @@
...
-
-###################################
-# We can see that the TV-l1 penalty is 3 times slower to converge and
-# gives the same prediction accuracy. However, it yields much
-# cleaner coefficient maps
+# fREM ensembling procedure yields an important improvement of decoding accuracy (+20%)
+# on this simple example compared to fitting only one model per fold and the
+# clustering mechanism keeps its computational cost reasonable
+# even on heavier examples. Here we ensembled several instances of l2-SVC,
+# a versatile classifier but fREMClassifier also works with ridge or logistic.
```
bthirion, line 76:
```suggestion
#  but fREMClassifier also works with ridge or logistic.
```

examples/02_decoding/plot_mixed_gambles_space_net.py
```diff
@@ -27,30 +27,58 @@
...
+fREM.fit(zmap_filenames, behavioral_target)
+print("fREM was fitted in {} seconds".format(int(time.time() - start_time)))
+
+# Visualize fREM weights
+# ----------------------------
+plot_stat_map(fREM.coef_img_['beta'], title="fREM", display_mode="yz",
+              cut_coords=[20, -2], threshold=.2)
+
+##########################################################################
+# We can see that Space Net model, yields a sparse coefficient map with
```
bthirion, line 79:
```suggestion
# We can see that Space Net model yields a sparse coefficients map with
```

nilearn/decoding/tests/test_decoder.py
bthirion, line 147: I realize that ' do_clustering=False, clustering_percentile=100' are redundant. Can we get rid of `do_clustering` ?

nilearn/decoding/decoder.py
```diff
@@ -1,43 +1,48 @@
 """High-level decoding object that exposes standard classification and
 regression strategies such as SVM, LogisticRegression and Ridge, with optional
-feature selection, integrated hyper-parameter selection
-and aggregation strategy in which the best models
-within a cross validation loop are averaged.
+feature selection, integrated hyper-parameter selection and aggregation strategy
+in which the best models within a cross validation loop are averaged.
+
+Also exposes a high-level method fREM which uses clustering and model
```
bthirion, line 6:
```suggestion
Also exposes a high-level method fREM that uses clustering and model
```

nilearn/decoding/decoder.py
```diff
@@ -133,24 +138,36 @@ def _check_estimator(estimator):
...
 def _parallel_fit(estimator, X, y, train, test, param_grid, is_classification,
-                  scorer, mask_img, class_index, screening_percentile=100):
+                  scorer, mask_img, class_index, screening_percentile,
+                  do_clustering, clustering_percentile):
     """Find the best estimator for a fold within a job.
     This function tries several parameters for the estimator for the train and
     test fold provided and save the ones that performs best. These models are
     used afterwards to build the averaged model.
-
+    This tries may be performed after one step of clustering (ReNA) and one
```
bthirion, line 147: Sorry don't understand what this means.

nilearn/decoding/decoder.py
```diff
@@ -133,24 +138,36 @@ def _check_estimator(estimator):
...
-    do_screening = (n_features > 100) and selector is not None
-
     X_train, y_train = X[train], y[train]
     X_test, y_test = X[test], y[test]
 
-    if (selector is not None) and do_screening:
+    # for fREM Classifier and Regressor : start by doing a quick ReNA clustering
+    # to reduce the number of feature by agglomerating similar ones
+
+    if do_clustering and (clustering_percentile < 100):
```
bthirion, line 156: These 2 conditions are redundant to me.

nilearn/decoding/decoder.py
```diff
@@ -876,3 +921,301 @@ def __init__(self, estimator='svr', mask=None, cv=10, param_grid=None,
...
+        A sequence of dicts signifies a sequence of grids to search, and is
+        useful to avoid exploring parameter combinations that make no sense
+        or have no effect. See scikit-learn documentation for more information,
+        for example: https://scikit-learn.org/stable/modules/grid_search.html
+
+    clustering_percentile: int, float, optional, in the closed interval [0, 100]
+        Used to perform a fast ReNA clustering on input data as a first step
+        of fit. It agglomerates similar features together to reduce their number
+        by this percentile. ReNA is typically efficient at reducing the number
+        of features by 10. Default: 10.
```
bthirion, line 969: I find this formulation a bit  ambiguous. Let me suggest:
"""
ReNA is typically efficient for cluster_percentile equal to 10. Default: 10.
"""

nilearn/decoding/decoder.py
```diff
@@ -876,3 +921,301 @@ def __init__(self, estimator='svr', mask=None, cv=10, param_grid=None,
...
```
bthirion, line 1119: Sam comment here

examples/02_decoding/plot_haxby_frem.py
jeromedockes: we don't need to save it at all, this line can just be removed

nilearn/decoding/decoder.py
thomasbazeille: Yes I thought nilearn was moving to absolute imports everywhere PR after PR ? Is that right ? should I change it for all the code I touch ?

nilearn/decoding/decoder.py
thomasbazeille: no I mean clustering in order to reduce the number of features 
from `n` voxels
to `n * clustering_percentile/100` clusters 

I'm not sure how to make it clearer though if you find it ill-explained. Would you have a suggestion ?

nilearn/decoding/decoder.py
thomasbazeille: You mean adding : 
``
if len(best_coef.shape) == 1:
    best_coef = best_coef.reshape(1, -1)
``
to fix #2360?

nilearn/decoding/decoder.py
thomasbazeille: I wondered about it. So the method is called fREM but the convention would mean using Frem (exact opposite). So I'm not sure which one is more confusing for users.

@emdupre ?

nilearn/decoding/decoder.py
thomasbazeille: When designing the PR with Gael we agreed on those:
- n_iter is enough for cases where clustering_percentile is very low but anyway it will be bypassed when the algorithm reaches the desired number of clusters.
- threshold is ok in general
- scaling I think it's a result of Andres thesis that it's not necessary.

Also you don't want to clutter a complex high-level object with technical parameters when those can be set right directly for all situations.

nilearn/decoding/decoder.py
thomasbazeille: Result from Andres paper in the general case 50 bagged estimators are a good trade-off between performance and computational cost. But for speed in examples and tests I reduced this number since datasets are not full size.

nilearn/decoding/decoder.py
kchawla-pi: Don't change existing imports, it makes the changes noisy and code reviews tiresome.
If you add an import, or change one, for your PR, then switch to absolute imports.

nilearn/decoding/decoder.py
kchawla-pi: While Classes are supposed to LikeThis, for symbolism with a specific meaning & serving a specific purpose, it is fine to change it a bit.
I prefer `fREM` over `Frem`.

jeromedockes:
> I wondered about it. So the method is called fREM but the convention would mean using Frem (exact opposite). So I'm not sure which one is more confusing for users.

keep fREM

nilearn/decoding/decoder.py
tbng: Yes adding this line indeed

thomasbazeille:
Thanks everyone for the comments. I almost solved them all except for one about coverage and those where I asked precisions.

If reviewers want to do one more pass, they are welcomed, if the build looks good for examples and doc it's almost ready to merge.

nilearn/decoding/decoder.py
thomasbazeille: Given the discussions happening on #2362  I guess we can run both in parallel wdyt?

kchawla-pi:
Think you mistyped `plot_stats_map` as `plot_stat_map` in the example.

thomasbazeille:
> Think you mistyped `plot_stats_map` as `plot_stat_map` in the example.

You got me ! I was searching for the problem before realizing plot_stat_map is the legit function.

thomasbazeille:
> Think you mistyped `plot_stats_map` as `plot_stat_map` in the example.
plot_stat_map is the legit function but I forgot to import it...

nilearn/decoding/decoder.py
thomasbazeille: thought about it again as those discussions are not related to this bugfixing. I'm thinking of including it.

examples/02_decoding/plot_mixed_gambles_space_net.py
```diff
@@ -20,37 +28,48 @@
...
-# --------------
-decoder = SpaceNetRegressor(mask=mask_filename, penalty="graph-net",
-                            eps=1e-1,  # prefer large alphas
-                            memory="nilearn_cache")
-decoder.fit(zmap_filenames, behavioral_target)
+#############################################################################
+# We can observe that the coefficients map learnt by fREM is structured,
+# due to the spatial regularity imposed by working on clusters and model
+# ensembling. Although these maps have been thresholded for display,
+# they are not sparse (i.e. almost all voxels have non-zero coefficients).#
```
bthirion, line 51: trailing #

nilearn/decoding/decoder.py
```diff
@@ -876,3 +920,301 @@ def __init__(self, estimator='svr', mask=None, cv=10, param_grid=None,
...
+
+        A sequence of dicts signifies a sequence of grids to search, and is
+        useful to avoid exploring parameter combinations that make no sense
+        or have no effect. See scikit-learn documentation for more information,
+        for example: https://scikit-learn.org/stable/modules/grid_search.html
+
+    clustering_percentile: int, float, optional, in the closed interval [0, 100]
+        Used to perform a fast ReNA clustering on input data as a first step
+        of fit. It agglomerates similar features together to reduce their number
+        by this percentile. ReNA is typically efficient for cluster_percentile
```
bthirion, line 1117:
```suggestion
        down to this percentile. ReNA is typically efficient for cluster_percentile
```

kchawla-pi:
Any other reviews? I think an alpha can be released once this PR is merged.

thomasbazeille:
Sry I was on holiday this week, sphinx still has some trouble building the
examples for stupid reasons but i have a hard time building localy and only
the relevant part of the documentation to check it.

Le sam. 18 avr. 2020 à 15:59, Kshitij Chawla <notifications@github.com> a
écrit :

kchawla-pi:
I think Jerome fixed that glitch. If you pull from master and incorporate BT's review, it'll be fine.

thomasbazeille:
The Travis failure seems spurious

kchawla-pi:
I've rerun the failing test.
There are Flake8 stuff to be fixed.

kchawla-pi:
For the Flake8 error, `W504 line break after binary operator`, there is no winning, so I have set it's sibling error `W504 line break before binary operator` to be ignored. SO place the arithmetic symbol on the next line and it should be fixed.

thomasbazeille:
> For the Flake8 error, `W504 line break after binary operator`, there is no winning, so I have set it's sibling error `W504 line break before binary operator` to be ignored. SO place the arithmetic symbol on the next line and it should be fixed.

Sorry @kchawla-pi I'm not sure to which log / module / line this relates.

Also I have a persisting sphinx error on indentation that make the build fail but I don't manage to locate it precisely. Help would be very useful here

``` Traceback (most recent call last):
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/site-packages/sphinx/cmd/build.py", line 276, in build_main
    app.build(args.force_all, filenames)
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/site-packages/sphinx/application.py", line 349, in build
    self.builder.build_update()
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 299, in build_update
    len(to_build))
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 311, in build
    updated_docnames = set(self.read())
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/site-packages/sphinx/util/logging.py", line 219, in pending_warnings
    memhandler.flushTo(logger)
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/site-packages/sphinx/util/logging.py", line 184, in flushTo
    logger.handle(record)
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/logging/__init__.py", line 1524, in handle
    self.callHandlers(record)
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/logging/__init__.py", line 1586, in callHandlers
    hdlr.handle(record)
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/logging/__init__.py", line 890, in handle
    rv = self.filter(record)
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/logging/__init__.py", line 751, in filter
    result = f.filter(record)
  File "/home/circleci/miniconda3/envs/testenv/lib/python3.7/site-packages/sphinx/util/logging.py", line 404, in filter
    raise SphinxWarning(location + ":" + message)
sphinx.errors.SphinxWarning: /home/circleci/project/doc/auto_examples/02_decoding/plot_mixed_gambles_space_net.rst:201:Unexpected indentation.

Warning, treated as error:
/home/circleci/project/doc/auto_examples/02_decoding/plot_mixed_gambles_space_net.rst:201:Unexpected indentation.
make[1]: *** [Makefile:161: check] Error 2
make[1]: Leaving directory '/home/circleci/project/doc'
make: *** [Makefile:72: html-strict] Error 2  ```

**Comment by kchawla-pi** [comment]
Okay. I'm running it locally and will update here.

Here is something: There shouldn't be a `plot_mixed_gambles_space_net` example, since you replaced it with `plot_mixed_gambles_frem`. I believe You have left a reference to the old file somewhere.

**Comment by kchawla-pi** [comment]
Though I can't find it using git grep. Will wit for the build process to complete.

**Comment by kchawla-pi** [comment]
I replaced that section with this:
```rst
.. |tvl1_haxby| image:: auto_examples/02_decoding/images/sphx_glr_plot_haxby_frem_001.png
   :target: auto_examples/02_decoding/plot_haxby_frem.html
```
Will run it in morning and see if it fixes this. You can try it as well.

kchawla-pi:
All done. 
C'est bon.

thomasbazeille:
Thanks a lot Kshitij! I did not notice that! Evreything is fine imho

kchawla-pi:
Seems that the ### overline and the ---- underline should be of the same length

thomasbazeille:
@kchawla-pi Is it possible to build locally only the doc modified by a PR ? Or to build only part of the doc to make testing quicker ? I couldn't find any doc and with the upcoming sprint I'd be glad to know what is doable.

kchawla-pi:
It's documented in the contributions guide here: 
http://nilearn.github.io/development.html#contributing-to-the-documentation

Specifically, use this pattern when in the docs directory:
```shell
python3 -m sphinx -D sphinx_gallery_conf.filename_pattern=plot_decoding_tutorial.py -b html -d _build/doctrees . _build/html
```

kchawla-pi:
I may have misunderstood the error. For sections, it could be that the symbols have to match as well as the length. You should check what's being done in other unrelated example files.

kchawla-pi:
Ignore my last comment

thomasbazeille:
Soon to be merged ! 

Finally all check pass, sorry for the delay. If some want to give a final look or check build artifacts it is the time ! 

(not much is new, just the fREM section of the documentation, examples were adapted from existing ones)

tbng:
I noticed the sphinx documentation build problem too late but instead of `git grep` I discovered this recently and it's really good alternative solution: https://github.com/vrothberg/vgrep

kchawla-pi:
Any more reviews? I will merge this soon.

nilearn/decoding/decoder.py
```diff
@@ -445,13 +480,22 @@ def fit(self, X, y, groups=None):
         self.screening_percentile_ = _adjust_screening_percentile(
             self.screening_percentile, self.mask_img_, verbose=self.verbose)
 
+        n_final_features = int(X.shape[1] * self.screening_percentile_
+                               * self.clustering_percentile / 10000)
+        if n_final_features < 50:
+            warnings.warn(
+                "After clustering and screening, the decoding model will "
+                "be trained only on {} features. ".format(n_final_features)
+                + "Consider raising clustering_percentile or "
```
jeromedockes, line 489: detail: you don't need to build several strings here. you can remove the `+` and move the `format` call after the last snippet, ie after `parameters"`
```
assert (
    "abc"
    "def"
    ) == "abcdef"
```

jeromedockes:
> Any more reviews? I will merge this soon.

@kchawla-pi I think you can merge this one, I had a quick look and LGTM. if needed further work on this can be done in a separate pr and after  eg #2362 is solved.

once this is done, it would be good to work on removing code duplication between decoder.py and spacenet

kchawla-pi:
@jeromedockes Pycharm has a cool feature to scan for duplicate code, and I am sure there are plugins for other IDEs and Editors. They might be useful.
Maybe we can mark out duplicated regions and address them during the sprint, throughout Nilearn.
I had noticed there is a lot of duplication in Data validation codes within modules.

PR #2340, opened 2020-03-19
[MRG] Replace _basestring compatibility shim with str datatype

kchawla-pi:
Now that Nilearn no longer supports Python2, we can directly use the `str` datatype instead of couching it behind _basestring.
This PR does that.


kchawla-pi:
Ready for review soon as the CI completes green.

PR #2341, opened 2020-03-19
[MRG] Removed Py2 related compatibility code

kchawla-pi:
Since we dont support Python2 aymore, we can start removing the compatibility shims


kchawla-pi:
CircleCI is passing: https://circleci.com/gh/kchawla-pi/nilearn/1063

GaelVaroquaux:
Thanks Kshitij!

Do you know why we are no longer seeing circle in the status?

nilearn/_utils/compat.py
```diff
@@ -1,76 +1,10 @@
...
-
-
-import joblib
-
-Memory = joblib.Memory
-Parallel = joblib.Parallel
-hash = joblib.hash
-delayed = joblib.delayed
-cpu_count = joblib.cpu_count
+def md5_hash(string):
```
GaelVaroquaux: I think that we should move this helper function in the only file where it is used: it is no longer a compatibility layer.

GaelVaroquaux:
We got rid of a file! :). That's great.

Merging!

PR #2343, opened 2020-03-20
[MRG] Fixing FutureWarning due to comparision with None

kchawla-pi:


nilearn/image/image.py
```diff
@@ -227,7 +227,7 @@ def _smooth_array(arr, affine, fwhm=None, ensure_finite=True, copy=True):
         arr = _fast_smooth_array(arr)
     elif fwhm is not None:
         fwhm = np.asarray(fwhm)
```
jeromedockes:
```suggestion
        fwhm = np.asarray([fwhm]).ravel()
```

jeromedockes: in case fwhm is a scalar

Issue #2360, opened 2020-04-01
Decoder with ridge_regressor fails on oasis vbm example


thomasbazeille:
On Nilearn master

To replicate : in Voxel-Based Morphometry on Oasis dataset example, changing
```
decoder = DecoderRegressor(estimator='svr', mask=mask,
                           scoring='neg_mean_absolute_error',
                           screening_percentile=1,
                           n_jobs=1)
``` 
by 
``` 
decoder = DecoderRegressor(estimator='ridge_regressor', mask=mask,
                           scoring='neg_mean_absolute_error',
                           screening_percentile=1,
                           n_jobs=1)
``` 
yields an error, `best_coef` returned is 1D instead of expected 2D array `

Any thoughts @tbng ? It's not related to my PR #2327 but I can include fix if needed

Full traceback: 
```
----> 1 decoder.fit(gm_imgs_train, age_train)

~/nilearn/nilearn/decoding/decoder.py in fit(self, X, y, groups)
    454                 self.mask_img_, c, self.screening_percentile_)
    455             for c, (train, test) in itertools.product(
--> 456                 range(n_problems), self.cv_))
    457 
    458         coefs, intercepts = self._fetch_parallel_fit_outputs(

~/miniconda3/lib/python3.7/site-packages/joblib/parallel.py in __call__(self, iterable)
   1002             # remaining jobs.
   1003             self._iterating = False
-> 1004             if self.dispatch_one_batch(iterator):
   1005                 self._iterating = self._original_iterator is not None
   1006 

~/miniconda3/lib/python3.7/site-packages/joblib/parallel.py in dispatch_one_batch(self, iterator)
    833                 return False
    834             else:
--> 835                 self._dispatch(tasks)
    836                 return True
    837 

~/miniconda3/lib/python3.7/site-packages/joblib/parallel.py in _dispatch(self, batch)
    752         with self._lock:
    753             job_idx = len(self._jobs)
--> 754             job = self._backend.apply_async(batch, callback=cb)
    755             # A job can complete so quickly than its callback is
    756             # called before we get here, causing self._jobs to

~/miniconda3/lib/python3.7/site-packages/joblib/_parallel_backends.py in apply_async(self, func, callback)
    207     def apply_async(self, func, callback=None):
    208         """Schedule a func to be run"""
--> 209         result = ImmediateResult(func)
    210         if callback:
    211             callback(result)

[...]

PR #2362, opened 2020-04-01
WIP Fix behaviour of parallel_fit with ridge based estimator

tbng:
- Condition to check whether output coefficients array is 1D --> change to 2D
- Change all based ridgeCV estimators to non-CV version: avoid redundancy
inside parallel_fit

TODO: test cases to cover scenario where input X is a list of Nibabel images instead of a single image.

Following issue #2360.

### DIFF
--- a/nilearn/decoding/decoder.py
+++ b/nilearn/decoding/decoder.py
@@ -18,7 +18,7 @@ from sklearn import clone
 from sklearn.base import RegressorMixin
 from sklearn.linear_model import LogisticRegression
 from sklearn.linear_model.base import LinearModel
-from sklearn.linear_model.ridge import RidgeCV, RidgeClassifierCV, _BaseRidgeCV
+from sklearn.linear_model.ridge import Ridge, RidgeClassifier, _BaseRidge
 from sklearn.model_selection import ParameterGrid, check_cv, LeaveOneGroupOut
 from sklearn.preprocessing import LabelBinarizer
 from sklearn.svm import SVR, LinearSVC
@@ -46,9 +46,9 @@ SUPPORTED_ESTIMATORS = dict(
     logistic_l1=LogisticRegression(penalty='l1', solver='liblinear'),
     logistic_l2=LogisticRegression(penalty='l2', solver='liblinear'),
     logistic=LogisticRegression(penalty='l2', solver='liblinear'),
-    ridge_classifier=RidgeClassifierCV(),
-    ridge_regressor=RidgeCV(),
-    ridge=RidgeCV(),
+    ridge_classifier=RidgeClassifier(),
+    ridge_regressor=Ridge(),
+    ridge=Ridge(),
     svr=SVR(kernel='linear', max_iter=1e4),
 )
 
@@ -96,7 +96,7 @@ def _check_param_grid(estimator, X, y, param_grid=None):
         # define loss function
         if isinstance(estimator, LogisticRegression):
             loss = 'log'
-        elif isinstance(estimator, (LinearSVC, _BaseRidgeCV, SVR)):
+        elif isinstance(estimator, (LinearSVC, _BaseRidge, SVR)):
             loss = 'squared_hinge'
         else:
             raise ValueError(
@@ -109,10 +109,11 @@ def _check_param_grid(estimator, X, y, param_grid=None):
         else:
             min_c = 0.5
 
-        if not isinstance(estimator, _BaseRidgeCV):
-            param_grid['C'] = np.array([2, 20, 200]) * min_c
+        if isinstance(estimator, _BaseRidge):
+            param_grid['alpha'] = np.array([0.1, 1.0, 10.0])
         else:
-            param_grid = {}
+            param_grid['C'] = np.array([2, 20, 200]) * min_c
+
 
     return param_grid
 
@@ -176,6 +177,10 @@ def _parallel_fit(estimator, X, y, train, test, param_grid, is_classification,
             best_param = param
 
     if (selector is not None) and do_screening:
+        # if reshape best_coef to 2D array to conform with inverse_transform
+        if len(best_coef.shape) == 1:
+            best_coef = best_coef.reshape(1, -1)
+
         best_coef = selector.inverse_transform(best_coef)
 
     return class_index, best_coef, best_intercept, best_param, best_score
--- a/nilearn/decoding/tests/test_decoder.py
+++ b/nilearn/decoding/tests/test_decoder.py
@@ -15,7 +15,7 @@ import numpy as np
 
 from sklearn.datasets import load_iris, make_classification, make_regression
 from sklearn.ensemble import RandomForestClassifier
-from sklearn.linear_model import LogisticRegression, RidgeCV, RidgeClassifierCV
+from sklearn.linear_model import LogisticRegression, Ridge, RidgeClassifier
 from sklearn.metrics import accuracy_score, r2_score
 from sklearn.model_selection import KFold, LeaveOneGroupOut
 from sklearn.svm import SVR, LinearSVC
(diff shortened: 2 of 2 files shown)


tbng:
@kchawla-pi @GaelVaroquaux  is it recommended to include code to download real nilearn images in the test cases? What I want to build is a test dataset where X is a list of images and y is reponse vector, similar to scenario in: https://github.com/nilearn/nilearn/blob/8e972c4d382a74fb126f768a0ae02e01e817515c/examples/02_decoding/plot_oasis_vbm.py#L57-L61

thomasbazeille:
I think we did that in the decoder tests haven't we ? 

You can transform a real dataset such as Iris to Niimgs by unmasking a 2D table to an arbitrary 4d structure with a mask of the right size or just create fake or constant niimgs if it's just to test shapes and all.

tbng:
> I think we did that in the decoder tests haven't we ?
> 
> You can transform a real dataset such as Iris to Niimgs by unmasking a 2D table to an arbitrary 4d structure with a mask of the right size or just create fake or constant niimgs if it's just to test shapes and all.

Indeed it is here I believe: https://github.com/nilearn/nilearn/blob/8e972c4d382a74fb126f768a0ae02e01e817515c/nilearn/decoding/tests/test_decoder.py#L236-L248

 but I think the fact that we did include 'ridge_regressor' within that test and it passed without error like in the `plot_oasis_vbm` example means there should be something that missed?

thomasbazeille:
But here X is a list of Niimgs already no ?  So maybe it's not that that triggers this pb. 
I think as nilearn tests are taking too long already for CI and dev, fetching a real dataset is not a solution.

In fact I think the traceback is "best coef should 2D and is only 1D"  when inverse transform of feature screening is applied. So it's the opposite problem. 

Thanks a lot for looking into this!

tbng:
Yeah I also think fetching real dataset is a poor method. Anyway with this fix the `plot_oasis_vbm` example should work with `ridge_regressor`.

bthirion:
Sorry, what is the status on that PR ?

tbng:
I'm not sure why circleCI build does not run. Otherwise I think I should add a test case to see the behaviour of parallel fit with  estimator type other than svr and the PR will be finished.

kchawla-pi:
If you are following Nilearn in your own CircleCI account, the run it won't show up here. Is that what you are doing? Go to circleci.com/dashboard and check the jobs. Login using GitHub .

kchawla-pi:
So your CircleCI builds are here: https://circleci.com/gh/tbng/nilearn

To see CircleCI builds in the PR, they must run in Nilearn's account on CircleCI, not tbng's account on CircleCI. To make them build in Nilearn's accounnt, go here: https://circleci.com/gh/tbng/nilearn/edit and "unfollow project". Then do a new push.

jeromedockes:
why are you not using ridgecv anymore?

tbng:
> why are you not using ridgecv anymore?

Upon realizing that in `_parallel_fit` of `decoder.py` the CV process are doing manually, therefore if we use another layer of CV estimator object it would be redundant & slower. 

I have think of including a condition to use BaseRidgeCV whenever these type of estimators are called, but these estimators will not return a cv score in which `_parallel_fit` would need to output the best cv score. Do you know a way to do this with RidgeCV @jeromedockes  ?

jeromedockes:
you could have just one dummy param in the grid e.g. the list of alphas passed to RidgeCV if the problem is that the param grid is empty. ie. you would have a parameter grid of length 1

thomasbazeille:
I think switching from RidgeCV to Ridge as @tbng proposes makes sense. Every estimator in Decoder is cross-validated through `_parallel_fit` and not before (SVC, logistics ...).

- To be consistent Ridge should be the same.
- Anyway to use the same function, we cannot take advantage of Ridge CV speed-up for cross-validation or we should make a special case for it.
- Then there is no point using RidgeCV without using its cross-validation, just use Ridge instead.

jeromedockes:
> I think switching from RidgeCV to Ridge as @tbng proposes makes sense. Every estimator in Decoder is cross-validated through `_parallel_fit` and not before (SVC, logistics ...).
>
> - To be consistent Ridge should be the same.
> - Anyway to use the same function, we cannot take advantage of Ridge CV speed-up for cross-validation or we should make a special case for it.
> - Then there is no point using RidgeCV without using its cross-validation, just use Ridge instead.
>
You can just think of RidgeCV as a model that has no hyperparameters.
you can still do one loop of cross-validation to get the score and
ensemble, but you don't need to loop over hyperparameters (alphas).
so you would still get close to 10 x speedup if you are considering 10
values for the hyperparameter

thomasbazeille:
> I think switching from RidgeCV to Ridge as @tbng proposes makes sense. Every estimator in Decoder is cross-validated through `_parallel_fit` and not before (SVC, logistics ...). - To be consistent Ridge should be the same. - Anyway to use the same function, we cannot take advantage of Ridge CV speed-up for cross-validation or we should make a special case for it. - Then there is no point using RidgeCV without using its cross-validation, just use Ridge instead.
> You can just think of RidgeCV as a model that has no hyperparameters. you can still do one loop of cross-validation to get the score and ensemble, but you don't need to loop over hyperparameters (alphas). so you would still get close to 10 x speedup if you are considering 10 values for the hyperparameter

You can indeed but it's not what was coded right now, unless i'm mistaken since the `param_grid` is never passed to the estimator. And you might need to tailor some things in `_parallel_fit` to account for this specific case.

tbng:
@thomasbazeille The change from `Ridge` to `RidgeCV` actually was made after our IRL discussion (me with @jeromedockes), with reasons stated already that `sklearn` native object will be much more efficient in terms of calculation.

@jeromedockes  I have thought of that as well with passing alphas into the RidgeCV inside `_check_param_grid` before passing to `_parallel_fit`. However `best_coef` still has to be output from `_parallel_fit` and inside RidgeCV there is currently no way to output that score. I try to put dummy `best_coef = None` with RidgeCV but in some cases users might need to check this attribute, e.g in `plot_decoding_tutorial.py` example.

kchawla-pi:
> To see CircleCI builds in the PR, they must run in Nilearn's account on CircleCI, not tbng's account on CircleCI. To make them build in Nilearn's accounnt, go here: https://circleci.com/gh/tbng/nilearn/edit and "unfollow project". Then do a new push.

You will lose the ability to restart yur failing build though. Someone with write access will have to do it for you.

jeromedockes:
> @jeromedockes I have thought of that as well with passing alphas into the RidgeCV inside `_check_param_grid` before passing to `_parallel_fit`. However `best_coef` still has to be output from `_parallel_fit` and inside RidgeCV there is currently no way to output that score. I try to put dummy `best_coef = None` with RidgeCV but in some cases users might need to check this attribute, e.g in `plot_decoding_tutorial.py` example.

But if you have just one parameter in the grid, it would still go through the loop in `parallel_fit` (once) and therefore produce best score, coef etc. no?
for example by making a change similar to this:
```diff
diff --git a/nilearn/decoding/decoder.py b/nilearn/decoding/decoder.py
index d924b7b07..17506f3d5 100644
--- a/nilearn/decoding/decoder.py
+++ b/nilearn/decoding/decoder.py
@@ -112,7 +112,7 @@ def _check_param_grid(estimator, X, y, param_grid=None):
         if not isinstance(estimator, _BaseRidgeCV):
             param_grid['C'] = np.array([2, 20, 200]) * min_c
         else:
-            param_grid = {}
+            param_grid = {"alphas": [[.01, .1, 1., 10., 100.]]}
 
     return param_grid
 
diff --git a/nilearn/decoding/tests/test_decoder.py b/nilearn/decoding/tests/test_decoder.py
index 2f94828ed..d5328cc3b 100644
--- a/nilearn/decoding/tests/test_decoder.py
+++ b/nilearn/decoding/tests/test_decoder.py
@@ -43,12 +43,12 @@ logistic_l2 = LogisticRegression(penalty='l2')
 ridge_classifier = RidgeClassifierCV()
 random_forest = RandomForestClassifier()
 
-regressors = {'ridge': (ridge, []),
+regressors = {'ridge': (ridge, ["alphas"]),
               'svr': (svr, 'C')}
 classifiers = {'svc': (svc, 'C'),
                'logistic_l1': (logistic_l1, 'C'),
                'logistic_l2': (logistic_l2, 'C'),
-               'ridge_classifier': (ridge_classifier, [])}
+               'ridge_classifier': (ridge_classifier, ["alphas"])}
 # Create a test dataset
 rand = np.random.RandomState(0)
 X = rand.rand(100, 10)
```

we can replace svc with ridge classifier in the example you mention and it seems to run without any problems

jeromedockes:
>  for example by making a change similar to this:

note: the change is wrt to master, not to your pr branch

tbng:
This PR can be closed with the merge of #2451 since they basically fix the same bug.

bthirion:
thx !

PR #2419, opened 2020-04-13
[MRG] make azure pipelines upload coverage reports

jeromedockes:
at the moment the reported coverage is incorrect because:

- for some reason some lines that are covered e.g. in `rm_file.py`, and
  correctly reported as covered when running tests from source directory, are
  incorrectly reported as missed when testing installed package with `pytest
  --pyargs nilearn`. I will try to look into it; I can reproduce it on my
  machine.

- the windows tests on azure pipelines seems to get coverage right but are not
  uploaded because it seems they do not have the codecov upload token. I don't
  think I have access to the azure account to add it. We can first try the bash
  uploader which apparently does not need the token. also the current yml file
  does not seem to use the correct syntax

- [example failure to upload report](https://dev.azure.com/Parietal/Nilearn/_build/results?buildId=1212&view=logs&jobId=2d020793-cab3-5687-348a-7c5338fd226e&j=2d020793-cab3-5687-348a-7c5338fd226e&t=16f038ee-03a8-5708-8b37-76a78d53707f)


kchawla-pi:
I can give you permissions. Which email address do you want to use to log in? Give it to me here or in the email to my Gmail

kchawla-pi:
I added you. Check your gmail

jeromedockes:
thanks @kchawla-pi !

jeromedockes:
this maps the codecov token so that azure pipelines exports it as an environment variable.
I also [made secret variables available to builds of forks](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#validate-contributions-from-forks). the secret variables in question are just the token that allows uploading reports to codecov.

kchawla-pi:
I considered it but since we are on the free tier, it is better to spread out the jobs so we don't backlog the CT infra when multiple PRs are being run. 
Additionally, the extra containers allow us to add different kind of testing if necessary; for example the dedicated dataset downloader testing that I was supposed to work on.

What was the issue with AZP? Was it because the token was not shared with forks?

jeromedockes:
Kshitij Chawla <notifications@github.com> writes:

> I considered it but since we are on the free tier, it is better to spread out the jobs so we don't backlog the CT infra when multiple PRs are being run. 

ok that's a good reason

> Additionally, the extra containers allow us to add different kind of testing if necessary; for example the dedicated dataset downloader testing that I was supposed to work on.
>
> What was the issue with AZP? Was it because the token was not shared with forks?

- token was not shared with builds of forks (it's a box to tick
  somewhere in the web UI)
- the secret variable was not mapped as an environment variable (they
  are not decrypted unless you request it in the yaml file)
- the cmd script was using bash syntax to refer to the environment
  variable

kchawla-pi:
Thanks man. This was a circuitous one to solve!

Issue #2423, opened 2020-04-21
SurfLabelsMasker


sahahn:
Hi,
I just wanted to gauge interest around potentially adding functionality like NiftiLabelsMasker, but for surface data (just any 1D or 2D - time-series). I know in particular the reverse_transform aspect could be useful, i.e., going from ROI values to a surface that could be easily passed to plot_surf_stat_map. I know I also personally find having pipeline compatible ROI extra extractors helpful. The natural extension to this idea too would be a base SurfMasker class... with surface versions (where possible) of the NiftiMasker functionality. 

My main concern would just be scope, i.e., the majority of nilearn is geared towards working with volumetric data, is, therefore, trying to add more support for surfaces too far outside of the intended project scope? If not, I would be interested in implementing some of it.

Thanks!

jeromedockes:
I think we are indeed planning to add more support for surfaces, see
e.g. https://github.com/nilearn/nilearn/issues/2171

ping @hcherkaoui

GaelVaroquaux:
I agree with Jérome: I think that we would be super excited about such a
contribution!

sahahn:
Great, thanks for the reply. I'll look into putting something together

hcherkaoui:
@jeromedockes I did not start any contribution on this. The road is free :)

thomasbazeille:
Hi everyone we'd love to take advantage of the sprint to have an open discussion about Surface support in Nilearn. It's scheduled today at 6p CET (12p EDT). More informations will be in the [sprint discord](https://discord.gg/bMBhb7w) around this time.

Hope to see you there !

dangom:
Will this be recorded somewhere? Would've loved to attend but unfortunately I have a meeting at that time.

Remi-Gau:
Will close this as we have now an (experimental) surface masker

Issue #2171, opened 2019-10-11
improve support for surface data in nilearn?


jeromedockes:
It seems that there is growing interest for surface data in nilearn and
nistats, for example:

-   <https://github.com/nistats/nistats/issues/401>
-   <https://github.com/nilearn/nilearn/issues/2132>
-   <https://github.com/nilearn/nilearn/issues/2128>
-   <https://github.com/nilearn/nilearn/issues/2423>

and nilearn already has functions for surface plots.

To move forward in this direction, we may want to add a proper way to
represent, mask and unmask, read and write surface data &#x2013; similar to
`Nifti1Image` and `NiftiMasker`. For now surface data is passed around as
separate arrays for the meshes and the data.

I guess nibabel's [`GiftiImage`](https://nipy.org/nibabel/reference/nibabel.gifti.html#nibabel.gifti.gifti.GiftiImage) is meant for that, although I am not sure how
it works and haven't found any examples. Maybe @juhuntenberg has used it
before?

@kchawla-pi @bthirion @GaelVaroquaux @juhuntenburg

dangom:
Is it an idea to also support Freesurfer surfaces?

bthirion:
Yes, because they are supported by nibabel.

sahahn:
I think there is certainly benefit in a more universal Surface class similar to Nifti1Image, but it definitely isn't straightforward. I.e., nibabel has the GiftiImage, for working with Gifti files, but for Freesufer support just has loading functions vs. an actual class ( https://nipy.org/nibabel/reference/nibabel.freesurfer.html). Then there is cifti case... which seems like it would need to be handled differently.

It seems like there are some functions where passing data separate from the associated geometry mesh is okay, i.e., in cases where you only are performing an operation on the data itself, like extracting ROIs or processing time-series w/ signal.clean. On the other hand, there are certain functionalities where knowing meta-data around what 'space' the surface is in would be required, e.g., if you wanted to allow resampling between different spaces (as an analog to the automatic volumetric re-sampling), though I think this would be a rabbit-hole to try and implement.

Likewise, because of the standard in outside software (Freesurfer and HCP) of keeping the data separate from the meshes, maybe it isn't a great idea to try and side-skirt that. 

I'm definitely curious on other's thoughts.

GaelVaroquaux:
The problem of the babel tower of surface file formats is one that we hit
when designing the surface plotting code.

We created a few helper functions to load the data when we did this. 
[...]

PR #2424, opened 2020-04-22
[WIP] Support for SurfLabelsMasker (and a bit of broader Surface support)

sahahn:
This represents my first pass for addressing #2423 and some of #2171.
I implemented a version of a SurfLabelsMasker to start, but along the way did my best to add in some support for more generally for Surfaces as a baseline for adding more Surface support.

I have a few specific points/questions I'd like to raise/get feedback on:
1. A more universal representation of surface data to compatible components of nilearn. 
For this pull, I defined mostly just a conceptual wrapper upon what's already there in regards to the surface plotting functionality. Specifically, if you look at the docs, I've tried to add an analog of 'Niimg-like' types, but adapted to surface data. I call these 'Nisurf-like', w/ 'Nisurf-data' representing just a single surface / a 1D Numpy array, and 'Nisurf-data-2D' or 'Nisurfs-data' representing the analog of Niimgs, so a 2D Numpy array, representing # or vertex by # of scans. Lastly, "Nisurf-mesh", which I don't actually do anything with, but would just be the surface mesh. I'm 100% open to re-working some or all of this, I just needed some more formal way of representing surfaces.

2. Related to 1, and #2171, I am wondering if it would be better to create an actual `Nisurf-data` class similar to NiftiImage. The biggest benefits, besides cleaning up some code, I think would be in scalability & in provided a more similar interface to NiftiImage, where you can easily have data saved on disk, and not be forced to load everything into memory, i.e. the caching functionality. In my current code I did add in the case that if a list of files is passed to SurfLabelsMasker, then they will be loaded and processed 1 by 1, but this feels a bit brittle.

3. Location of where I added functions and files. I tried to mimic how the complementary non-Surface based function were written and stored whenever possible, but there were a number of cases where I wasn't too clear on if I was adding pieces in a good spot. 

Thanks, let me know if you have other feedback too!


jeromedockes:
thanks a lot @sahahn this is super useful. Re. 1 and 2, I think that indeed it would be helpful to have a relatively uniform interface to surface data that mimmicks the nifti images. The diff is somewhat large so a full review will take some time but I will try to give you some feedback by this weekend

sahahn:
No rush!

nilearn/surface/tests/test_nisurf.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+
+
+
+
+
+
+
+
```
kchawla-pi, line 203: This White space should be removed.

nilearn/input_data/tests/test_surf_labels_masker.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+
+    
+
+    
+
+
+    
+
```
kchawla-pi, line 242: This White space should be removed.

nilearn/input_data/tests/test_surf_labels_masker.py
```diff
@@ -0,0 +1,242 @@
...
+from nilearn.input_data.surf_labels_masker import (SurfLabelsMasker,
+                                                   _all_files_check,
+                                                   _get_single_surf,
+                                                   surf_to_signals_labels)
+from numpy.testing import assert_array_equal
+
+
+def test_surf_labels_maskers():
+
```
kchawla-pi, line 19: Instead of writing multiplle tests within a single function, and separating them with comments, it is better to write multiple small tests, with the test function names reflecting what is being tested (instead of the comment).

kchawla-pi: If that means recreating the simple data for testing, that is okay.

bthirion:
Should we keep 'Nisurf-mesh' if we don't use them ? I'd rather not...
@jeromedockes WDYT ?

bthirion:
Suggestion 2 makes sense to me.

sahahn:
> Should we keep 'Nisurf-mesh' if we don't use them ? I'd rather not...
> @jeromedockes WDYT ?

Yeah I just added the Nisurf-mesh option with the idea that if we decided to keep this format, it would make sense to update the parameters for the surface plotting too

jeromedockes:
> Should we keep 'Nisurf-mesh' if we don't use them ? I'd rather not...
> @jeromedockes WDYT ?

meshes are not used in the current diff because it is just masking directly the data, and we don't have resampling for surface data. but the Nisurf-data without a mesh is like a nifti image's data array without a shape and affine.

juhuntenburg:
This is really cool @sahahn

Before moving forward much I think we should make a decision on the internal representation of surfaces as discussed in #2171 . There will be a live discussion during the coding sprint today at 6p CET / 12p EDT, maybe you can join?

Regarding your questions 1. and 2. 
There is a GiftiImage class in Nibabel which I believe would address many of your concerns, and which I am more and more in favour of using here. Whatever it lacks in functionality that we want, we could try to contribute it in Nibabel directly. The advantage is that it is a standardized format. The disadvantage is that it is a bit complex and not so many people actually use it. The big question will be if we want to handle meshes and data together (which is possible in Gifti and avoids confusion) or separately (which many people do and saves memory) or allow for both.
Btw it also allows for 2D data, so that would simplify some of those applications too.

Thanks for this and it would be great to have you at the discussion today!

juhuntenburg:
Turns out I played around with this at some point:
https://github.com/juhuntenburg/gogifti/blob/master/gii.py

juhuntenburg:
> Whatever it lacks in functionality that we want, we could try to contribute it in Nibabel directly.

Turns out I played around with this at some point:
https://github.com/juhuntenburg/gogifti/blob/master/gii.py

effigies:
@juhuntenburg Check out the new [`GiftiImage.agg_data()`](https://nipy.org/nibabel/reference/nibabel.gifti.html#nibabel.gifti.gifti.GiftiImage.agg_data) interface.

juhuntenburg:
Cool! Another point in case for using GiftiImage

sahahn:
> This is really cool @sahahn
> 
> Before moving forward much I think we should make a decision on the internal representation of surfaces as discussed in #2171 . There will be a live discussion during the coding sprint today at 6p CET / 12p EDT, maybe you can join?
> 
> Regarding your questions 1. and 2.
> There is a [GiftiImage object](https://nipy.org/nibabel/reference/nibabel.gifti.html#nibabel.gifti.gifti.GiftiImage) in Nibabel which I believe would address many of your concerns, and which I am more and more in favour of using here. Whatever it lacks in functionality that we want, we could try to contribute it in Nibabel directly. The advantage is that it is a standardized format. The disadvantage is that it is a bit complex and not so many people actually use it. The big question will be if we want to handle meshes and data together (which is possible in Gifti and avoids confusion) or separately (which many people do and saves memory) or allow for both.
> Btw it also allows for 2D data, so that would simplify some of those applications too.
> 
> Thanks for this and it would be great to have you at the discussion today!

Yeah, these definitely get at the heart of some of the issues. I suppose one option would be to just convert all passed surface input to GiftiImages.

I also resonate with "The disadvantage is that it is a bit complex and not so many people actually use it.", and see some advantage to some utility functions that make it easier. Importantly though as you mentioned, we will need to decide on some of the scope of what we want the surface to do, i.e., meshes and textures passed together, first, as that will inform the right class to use for Surfaces + what utility functions will actually be useful

Remi-Gau:

Remi-Gau:
Will close this one as obsolete as we now have an experimental SurfaceLabelMasker

https://github.com/nilearn/nilearn/blob/11a4814492ec5a960f1c27ad9e9649e3a49a56bf/nilearn/experimental/surface/_maskers.py#L251

Issue #2426, opened 2020-04-29
high_variance_confounds fails when images contain NaNs


bthirion:
This is because the function uses  `scipy.stats.scoreatpercentile` (in signal.py)
should be replaced by `numpy.nanpercentile`.

Note that images often contain NaNs (e.g. after SPM preprocessing)

PR #2427, opened 2020-04-29
using robust precentile in variance thresholding for high variance confounds

bthirion:
addresses #2426


bthirion:
Any review ? This is an easy one :-)

PR #2428, opened 2020-04-30
moving brainsprite as an external python dependency

pbellec:
A bit of a weird pull request, as I am asking to basically move a few hundreds lines of code to a different project. 

The core of `html_stat_map.py` is an python interface to the brainsprite library. I have refactored this API to make it more modular, and allow for more flexible use cases, some of which I am hoping will benefit nilearn in the future. But I don't think this type of flexible API fits the scope of nilearn, and I have worked on establishing that code as a separate project (see the [new website](https://brainsprite.github.io) and [code base](https://github.com/brainsprite/brainsprite) ). I am proposing to include this new library under `externals`, just like what was done with `tempita`. I have re-written `view_img` using the new interface. I think it is cleaner than it was, in terms of logic.

The benefits I see for the nilearn community are:
 * less code to maintain. 
 * retain the core functionality identical. 
 * in the future, benefit  of the growth of brainsprite as a separate library. The proposed new interface would allow for example a 4D interactive viewer soon. 

I don't see major drawbacks unless some of the code I have removed could be useful elsewhere (the tests will tell us that). In which case I would be happy to move back some bits of code.

Another somewhat viable alternative I can think of would be to keep all the code inside nilearn, and move brainsprite.py under reporting for example. I could live with that, but I am not sure the PRs on evolving brainsprite's internal structure will be well received. 

Finally, if you have some thoughts on the redesign of `view_img`, feedback would be very useful, or you can also comment here https://github.com/brainsprite/brainsprite/issues/93


jeromedockes:
Thanks Pierre, and congratulations on what can now be done with
brainsprite it's a very impressive tool. I think having the official
python API in externals rather than custom code in html_stat_map makes
sense.
The vendored version wouldn't be updated frequently though, so I would
like to know how stable is the brainsprite python wrapper at the moment?
If I understand correctly it hasn't been released yet?
And nilearn's interface would stay the same (i.e. limited to view_img),
and users interested in more advanced features of brainsprite would
install brainsprite.py, is that correct?
Will you participate in the nilearn sprint on May 20 and 21?

pbellec:
> I would like to know how stable is the brainsprite python wrapper at the moment?
If I understand correctly it hasn't been released yet?

It's not stable at the moment, I just made a pre-release today. I am not in a rush to merge this and, besides a minor bug fix (which I could PR separately), it doesn't change anything in terms of functionality. I just wanted to know what the team thought, and possibly get feedback on the API itself. The API will only solidify when It expands to 4D data and multiple viewers, which will take a few months at least. After that, the other features I can imagine for brainsprite will be mostly on the javascript side, and will not change much for the python code. 

> And nilearn's interface would stay the same (i.e. limited to view_img), and users interested in more advanced features of brainsprite would install brainsprite.py, is that correct?

That's correct: this PR does not introduce any visible change for users of `view_img`. Hopefully there will be other interfaces at some point - I have a few ideas. The idea is for folks who want to build more custom types of html documents (e.g QC dashboards) to do so with brainsprite.py. 

> Will you participate in the nilearn sprint on May 20 and 21?

I am going to try hard to participate, I would like to contribute more to nilearn. But I am part of the organization team of brainhack school that week, so my time will be limited, at best.

bthirion:
Sounds good. There is only one Travis failure
"FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nilearn/externals/brainsprite/data/html/brainsprite_template_base64.html'"

pbellec:
Happy to hear you are positive about the proposed changes. I am going to wait for the next brainsprite release before finalizing this PR, as the API should be more or less stable by then, and it will integrate performance improvement as well. I will get back to you when this is done.

jeromedockes:
Sounds good!

BTajini:
Amazing work!
I had been testing the wrapper, however, I encountered an error..

```
----> 3 template = tempita.Template.from_filename(file_template, encoding="utf-8")
      4 
      5 viewer = bsprite.transform(template, javascript='js', html='html', library='bsprite')
/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/tempita/__init__.py in from_filename(cls, filename, namespace, encoding, default_inherit, get_template)
    139     def from_filename(cls, filename, namespace=None, encoding=None,
    140                       default_inherit=None, get_template=get_file_template):
--> 141         f = open(filename, 'rb')
    142         c = f.read()
    143         f.close()
FileNotFoundError: [Errno 2] No such file or directory: '../docs/source/_static/viewer_template.html'
```

jeromedockes:
> I had been testing the wrapper, however, I encountered an error..
>
> ```
> ----> 3 template = tempita.Template.from_filename(file_template, encoding="utf-8")
>       4 
>       5 viewer = bsprite.transform(template, javascript='js', html='html', library='bsprite')
> /opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/tempita/__init__.py in from_filename(cls, filename, namespace, encoding, default_inherit, get_template)
>     139     def from_filename(cls, filename, namespace=None, encoding=None,
>     140                       default_inherit=None, get_template=get_file_template):
> --> 141         f = open(filename, 'rb')
>     142         c = f.read()
>     143         f.close()
> FileNotFoundError: [Errno 2] No such file or directory: '../docs/source/_static/viewer_template.html'
> ```

Indeed I can reproduce that. in brainsprite, running
tests/plot_stat_map.py only works from the source directory /tests
because of the relative link to the tempita template. also, that
template needs to be included in the package_data in the setup.py

pbellec:
Thanks for catching this @BTajini and @jeromedockes 

> Indeed I can reproduce that. in brainsprite, running
tests/plot_stat_map.py only works from the source directory /tests
because of the relative link to the tempita template

I think this is now [fixed](https://github.com/brainsprite/brainsprite/blob/master/brainsprite/tests/test_brainsprite.py#L233) 

> that template needs to be included in the package_data in the setup.py

This [line](https://github.com/brainsprite/brainsprite/blob/master/setup.py#L23) should take care of that.

also @bthirion 
> "FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nilearn/externals/brainsprite/data/html/brainsprite_template_base64.html'"

I am not sure what happens here, as this file is actually in the library. I will investigate more. 

I did receive excellent feedback re the new internal API, and I think we have something [clean](https://github.com/brainsprite/brainsprite/issues/93#issuecomment-625611821) to move forward (but you are still welcome to comment if you see improvements/shortcomings). It will probably take me a couple of weeks before implementing these changes, and then I'll aim for finalizing this pull request.

jeromedockes:
> I am not sure what happens here, as this file is actually in the library. I will investigate more. 
>
but it is not in package_data in nilearn's setup.py (externals is just a
subpackage, the brainsprite readme in there is not processed by pip so
you need something like "externals.brainsprite": [...] in nilearn's
setup where you list all the non-python files you want copied to the
installation directory)

pbellec:
> but it is not in package_data in nilearn's setup.py 

Got it! Will fix.

BTajini:
Based on what I noticed in my new demo,
the wrapper works well. I don't know why, maybe it's an earlier version?
When I inspect the html page, it looks like we're using brainsprite. 

```
  // On load: build all figures
  $( window ).on('load',function() {
    // Create brain slices
    var brain = brainsprite(
    {"canvas": "3Dviewer", "sprite": "spriteImg", "nbSlice": {"X": 91, "Y": 109, "Z": 91}, "overlay": {"sprite": "overlayImg", "nbSlice": {"X": 91, "Y": 109, "Z": 91}, "opacity": 1}, "colorBackground": "#000000", "colorFont": "#FFFFFF", "crosshair": true, "affine": [[2.0, 0.0, 0.0, -90.0], [0.0, 2.0, 0.0, -126.0], [0.0, 0.0, 2.0, -72.0], [0.0, 0.0, 0.0, 1.0]], "flagCoordinates": true, "title": "Mix mat comp. 0", "flagValue": false, "numSlice": {"X": 62.0, "Y": 48.5, "Z": 65.0}}
    );
  });
 ```
The demo was made directly from plotting.view and not from the root of the source code.

pbellec:
I sadly agree to close this PR.

I could not find the time to complete the work needed on brainsprite itself, and at this stage it is not clear it will happen.

I'll re open a fresh PR if and when I manage to do it.

PR #2431, opened 2020-05-01
Fix colorbar cropping in matplotlib 3.2.0

### LINKED ISSUE #2430: Incorrect colorbar with thresholded statmap
I'm trying to plot a thresholded statistical map, with values ranging between 0.0 and 0.27.

I can't get the colorbar 

Running `plotting.plot_stat_map(svm_loc)` gives me the following:
image

Notice the small colorbar box on the bottom right. I've tried playing around with `threshold`, `symmetric_cbar`, `vmax` without any success.

I'm taking a look at `plot_stat_map`, and specifically this line of `_crop_colorbar`:
https://github.com/nilearn/nilearn/blob/927976dad301ff729d3e695aef0aad7a9be53084/nilearn/plotting/img_plotting.py#L223

From what I can tell, it's being given cbar_vmin and cbar_vmax values of 0 and None, which are adjusted to 0 and 0.27, and _then_ normalized to 0.5 and 1.

If I replace `cbar.ax.set_ylim(cbar.norm(cbar_vmin), cbar.norm(cbar_vmax))` with `    cbar.ax.set_ylim(cbar_vmin, cbar_vmax)`, and disable the outline, I get the following:

image

My guess is due to the following API change in matplotlib 3.x: https://matplotlib.org/3.2.1/api/api_changes.html#matplotlib-colorbar-colorbar-uses-un-normalized-axes-for-all-mappables

nilearn: 0.7.0a
matplotlib: 3.2.1

ariekahn:
Fixes #2430 

Matplotlib 3.2.0 no longer scales colorbars between 0 and 1. `_crop_colorbar` was using `cbar.norm` to scale `cbar_vmin` and `cbar_vmax`, which now leads to broken colorbar values:
image

Removing the norming behavior leads to expected results:
image
I don't see any obvious fix that would work for both new and old behavior, so this simply adds a version check for matplotlib.

### DIFF
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -13,6 +13,7 @@ import collections.abc
 import functools
 import numbers
 import warnings
+from distutils.version import LooseVersion
 
 # Standard scientific libraries imports (more specific imports are
 # delayed, so that the part module can be used without them).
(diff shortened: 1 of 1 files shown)


jeromedockes:
thanks a lot!



> I don't see any obvious fix that would work for both new and old behavior, so this simply adds a version check for matplotlib.

I think that's the best approach

jeromedockes:
example to reproduce:
```python
import numpy as np
import matplotlib
from matplotlib import pyplot as plt

from nilearn.datasets import fetch_neurovault_motor_task
from nilearn.plotting import plot_stat_map
from nilearn import image


img = fetch_neurovault_motor_task()["images"][0]
img = image.new_img_like(img, np.abs(image.get_data(img)) / 20)
plot_stat_map(img, threshold=.15)

plt.gcf().savefig("/tmp/plot.png", bbox_inches="tight")
print("matplotlib:", matplotlib.__version__)
```

using matplotlib 3.2.1:

with master:

plot


with this pr:

plot_fixed

thomasbazeille:
Looks good to me, thanks for the PR

bthirion:
Merging, thx.

PR #2432, opened 2020-05-03
Add high_pass to docs for make_first_level_design_matrix

rob-luke:
This argument is passed to `_make_drift`, so I replaced the text with the same description as is used in `_make_drift `.


bthirion:
Merging.

PR #2434, opened 2020-05-06
add 1% threshold in codecove yml

jeromedockes:
this should reduce noise of checks failing with "+- .01%" diff coverage etc.
1% is the value used by scikit-learn

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,18 @@
+# similar to scikit-learn .codecov.yml
+coverage:
+  status:
+    project:
+      default:
+        # Commits pushed to master should not make the overall
+        # project coverage decrease by more than 1%:
+        target: auto
+        threshold: 1%
+    patch:
+      default:
+        # Be tolerant on slight code coverage diff on PRs to limit
+        # noisy red coverage status on github PRs.
+        target: auto
+        threshold: 1%
+
 ignore:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -1,2 +1,18 @@
...
+    project:
+      default:
+        # Commits pushed to master should not make the overall
+        # project coverage decrease by more than 1%:
+        target: auto
+        threshold: 1%
+    patch:
+      default:
+        # Be tolerant on slight code coverage diff on PRs to limit
+        # noisy red coverage status on github PRs.
```
emdupre, line 13: Is it worth also including the comment from sklearn about using the browser extension ?

```
# Note The coverage stats are still uploaded
# to codecov so that PR reviewers can see uncovered lines
# in the github diff if they install the codecov browser
# extension:
# https://github.com/codecov/browser-extension
```

jeromedockes:
> Is it worth also including the comment from sklearn about using the browser extension ?

I think it may be outdated; I can see coverage reports online and I
haven't installed any extension

emdupre:
You're right; I think it's outdated !

I remembered using the extension at some point (I _think_ I installed it on an older computer), but it looks like it's totally deprecated, now. Maybe worth an issue in sklearn :)

PR #2435, opened 2020-05-12
Addition of cortical surface-based searchlight, with example Surfsearchlight dev

kingjr:
Taking over #1959

- [X] change fetching of novel dataset, and re-use haxby
- [X] fix redundant masking
- [X] pep8, doc etc
- [x] decide whether new object is necessary or whether pure numpy example suffices


kingjr:
- Travis' errors seems unrelated to this PR

- some svm don't converge, shall I explicitly change their number of iterations

- please check how I'm using the masking API, I'm new to this.

kingjr:
On second thought, @SylvainTakerkart's introduction of a new surf searchlight object may be an overkill: an example / tutorial remains relatively straightforward:

```python
from sklearn.linear_model import RidgeClassifier
from sklearn import neighbors
from nilearn.decoding.searchlight import search_light
# load data...

# Bold volume to bold surfaces
fsaverage = nilearn.datasets.fetch_surf_fsaverage()
hemi = 'left'
mesh = fsaverage['pial_%s' % hemi]
radius = 3.
X = surface.vol_to_surf(img, mesh, radius=radius)

# Define each vertex neighbouring for search light
coords, _ = surface.load_surf_mesh(mesh)
nn = neighbors.NearestNeighbors(radius=radius)
A = nn.fit(coords).radius_neighbors_graph(coords).tolil()

# Cross-validated search light
cv = KFold(3)
estimator = RidgeClassifier()
scores = search_light(X.T, y, estimator, A, cv=cv, n_jobs=-1)

# Plot scores on surface
plotting.plot_surf_stat_map(stat_map=scores,
                            surf_mesh=fsaverage['infl_%s' % hemi],
                            view='ventral')
```

LMKWYT

SylvainTakerkart:
> On second thought, @SylvainTakerkart's introduction of a new surf searchlight object may be an overkill: an example / tutorial remains relatively straightforward:
>
> LMKWYT

Hi Jean-Rémi, thanks for your feedback!

I don't think it's an overkill... In your example, you're performing computation in the volume and looking at the result on the surface... Sure it works, but the literature shows that computing directly on the surface can be better:
http://doi.org/10.1016/j.neuroimage.2010.07.035

Btw, ru taking over coding on this one? I've had difficulties finding time to finalize this PR, but I should have time next week during the sprint...

Let me know,

Sylvain

kingjr:
Thanks @SylvainTakerkart 

> I don't think it's an overkill... In your example, you're performing computation in the volume and looking at the result on the surface... Sure it works, but the literature shows that computing directly on the surface can be better:
http://doi.org/10.1016/j.neuroimage.2010.07.035

The above example decodes in surface space

> Btw, ru taking over coding on this one? I've had difficulties finding time to finalize this PR, but I should have time next week during the sprint...

I wanted to try this feature, and saw that you PR hadn't moved in a while, so I used it as practice.  But I'm happy either way, whatever you think is most efficient

SylvainTakerkart:
> Thanks @SylvainTakerkart
> 
> > I don't think it's an overkill... In your example, you're performing computation in the volume and looking at the result on the surface... Sure it works, but the literature shows that computing directly on the surface can be better:
> > http://doi.org/10.1016/j.neuroimage.2010.07.035
> 
> The above example decodes in surface space

ah, yes, sorry I'd looked at your code too quickly!
good call!
then indeed, my original PR might totally be an overkill ;)

> 
> > Btw, ru taking over coding on this one? I've had difficulties finding time to finalize this PR, but I should have time next week during the sprint...
> 
> I wanted to try this feature, and saw that you PR hadn't moved in a while, so I used it as practice. But I'm happy either way, whatever you think is most efficient

I'll check in my original code to make sure it's totally equivalent, but I guess it's not a question anymore... 

@bthirion Shall we just add an example on how to perform surface-based searchlight decoding based on the above (excellent) example of @kingjr ?

thomasbazeille:
This discussion is linked to #2423. I think we need to define clearly which functionalities we want to support natively for surfaces. 

If we feel surface Searchlight is an important usecase, a wrapper is definitely not an overkill to make it available to a wider audience and it should be quite factored to code once SurfaceMasker is available.

We should definitely have the wider discussion bout Surface support we need in nilearn with as many people as possible during the sprint.

SylvainTakerkart:
Yes, a SurfaceMasker is something that was missing when I'd implemented the first version of the surface searchlight... Count me in for the surface discussion during the sprint!

jeromedockes:
In the meanwhile please add any thoughts you may have on the subject to
#2171 !

bthirion:
Thx for the great suggestion. 
I think that it is always good to start with an example: if you can create a simple example that does it, then we should certainly add it.  And I like light design !
This does not obviate the need of dedicated tools to handle surface data as discussed on #2171.

kingjr:
Ok, here is a light example, ready for review.

kingjr:
(I get a bunch of travis / ci errors that seem unrelated to the present code, not sure what to do about it)

SylvainTakerkart:
> Ok, here is a light example, ready for review.

Hi, (dunno if I can review, but took a look ;) )

I'd just spotted that in your initial example quoted above, you were projecting on the inflated surface, which is not correct. But I see that you corrected this and you're now projecting onto the pial surface, which is clearly more correct! However, I dunno what's the correct practice given the projection algorithm implemented in nilearn (using the pial or the white mesh?)...

SylvainTakerkart:
...also, I think there's no reason to use two meshes (one for projecting, one to compute the A matrix); you can compute your A matrix using the mesh you're using for the projection, it will give you the same thing, and the code will be more minimal!

kingjr:
> ...also, I think there's no reason to use two meshes (one for projecting, one to compute the A matrix); you can compute your A matrix using the mesh you're using for the projection, it will give you the same thing, and the code will be more minimal!

My understanding is that
- The first mesh is to average voxels to the nearest surface, and should thus be in pial
- The second mesh is to concatenate nearby surfaces to feed the searchlight, and should thus be in inflated 

the radius could be expliclity different and the documentation improved if you think it helps

kingjr:
Related to this mesh issue, the correct approach would be to project voxels onto pial surface while restricting this to voxels *below* the surface. 

The current projection works by euclidian distance and thus "merges the sides along sulci".

However, I neither know how to retrieve cortical depth nor do I know how to get the cortical midline (e.g. between 'pial' and 'white'). Also, this is a bigger issue to the current searchlight example IMO.

jeromedockes:
> (I get a bunch of travis / ci errors that seem unrelated to the present code, not sure what to do about it)

yes, they are due to a dataset failing to download from OSF and some
doctests whose output was changed by a recent scikit-learn release;
there are open PRs to fix this and these errors should go away soon.

jeromedockes:
> Related to this mesh issue, the correct approach would be to project voxels onto pial surface while restricting this to voxels *below* the surface. 
that (sampling the volume only on one side of the surface) is something
that can be easily added to the surface projection's interface.

what could be nice to add as well is a projection that accepts both a
pial and wm surface with a one-to-one correspondance between the nodes,
and samples on the line segment between each pair of facing nodes. easy
to implement but requires a bit of discussion regarding the interface
because it does not fit the same function signature.

kingjr:
> what could be nice to add as well is a projection that accepts both a
pial and wm surface with a one-to-one correspondance between the nodes,
and samples on the line segment between each pair of facing nodes. easy
to implement but requires a bit of discussion regarding the interface
because it does not fit the same function signature.

That would be a great tool, also useful to do cortical depth analysis. I would recommend to open a separate PR.

SylvainTakerkart:
> > what could be nice to add as well is a projection that accepts both a
> > pial and wm surface with a one-to-one correspondance between the nodes,
> > and samples on the line segment between each pair of facing nodes. easy
> > to implement but requires a bit of discussion regarding the interface
> > because it does not fit the same function signature.
> 
> That would be a great tool, also useful to do cortical depth analysis. I would recommend to open a separate PR.

yep!
overall, "how to project?" is a vast issue with a literature of its own... this definetely should be a separate discussion, based on the existing literature and methods...

I'll get back to you on the one vs. two meshes question later, I need to check a few things...

jeromedockes:
> yep!
> overall, "how to project?" is a vast issue with a literature of its own... this definetely should be a separate discussion, based on the existing literature and methods...

yes. keeping in mind requirements for nilearn:

- very simple interface, not needing a lot of files and in particular
  being able work with the output of nilearn.datasets.fetch_fsaverage
- being fast
- not adding new dependencies
- not adding complex code unless there is a clear improvement
- covering only the most common use cases is fine

bthirion:
> what could be nice to add as well is a projection that accepts both a pial and wm surface with a one-to-one correspondance between the nodes, and samples on the line segment between each pair of facing nodes. easy to implement but requires a bit of discussion regarding the interface because it does not fit the same function signature.

If I were to implement one new sampling method for surface data, I would use that one, as it is regarded as the gold standard. This is what Freesurfer does.

SylvainTakerkart:
> > ...also, I think there's no reason to use two meshes (one for projecting, one to compute the A matrix); you can compute your A matrix using the mesh you're using for the projection, it will give you the same thing, and the code will be more minimal!

Hi! Finally took some time to gather my thoughts...

> 
> My understanding is that
> 
>     * The first mesh is to average voxels to the nearest surface, and should thus be in pial

yes! pial or white mesh are ok here!

> 
>     * The second mesh is to concatenate nearby surfaces to feed the searchlight, and should thus be in inflated

the choice is a bit more tricky here... you want to make sure that you reach your initial objective when you decided to use the surface, i.e not catch in your neighborhood/searchlight a vertex that is "close in the volume" but "far on the surface" (the typical example of the "other bank of a gyrus")... if we use "NearestNeighbor" with a radius argument to construct the neighborhood(s), as done in my initial code and in your example, maybe the safest way to do this is to use the spherical mesh (coz the inflated is still a bit folded)...

WDYT?

the use of "radius" on the inflated or spherical mesh is a bit misleading also (the user might think it's using a radius in native space), but I don't have an idea on how to improve this...

> 
> 
> the radius could be expliclity different and the documentation improved if you think it helps

thomasbazeille:
Hi everyone we'd love to take advantage of the sprint to have an open discussion about Surface support in Nilearn. It's scheduled today at 6p CET (12p EDT). More informations will be in the [sprint discord](https://discord.gg/bMBhb7w) around this time.

SylvainTakerkart:
Hi,

Sooo, we're nilearn-sprinting... I looked at bit more at the @kingjr 's great example... Everything looks good to me (except my question above about sphere vs. inflated)... However, there might be a question of consistency with the volume searchlight example:

- the volume searchlight example uses the Searchlight class
- this surface searchlight example uses the lower-level search_light function...

Sooo, I guess, @bthirion, it's your call to decide what you prefer...

(and to summarize, this is what the SurfSearchlight class was doing in #1959 : providing a class that calls the search_light function in a manner that is specific for surface)

thomasbazeille:
> However, there might be a question of consistency with the volume searchlight example:
> the volume searchlight example uses the Searchlight class
> this surface searchlight example uses the lower-level search_light function...

I think this is a fair point and Bertrand's sound suggestion was to merge the example as is for now, and wait on the SurfMasker to have a light SurfSearchlight object ? 
I'd be +1 on that

bthirion:
I like the simplicity of this example. I believe that in the longer term, we will encapsulate this in a function, and/or create a light SurfSearchlight object.
Regarding the sphere/inflated discussion i see the point. Actually, with few lines of code we could generate a purely topological graph (take all the nearest neighbors on the surface) that would not depend on any reduce parameter. I think that would completely solve the problem.

bthirion:
I mean, something like:
```
def mesh_to_graph(mesh):
    """Convert a mesh into a connectivity matrix"""
     coords, triangles = nib.freesurfer.read_geometry(mesh)
     n_points = coords.shape[0]
     edges = np.hstack((
        np.vstack((triangles[:, 0], triangles[:, 1])),
        np.vstack((triangles[:, 0], triangles[:, 2])),
        np.vstack((triangles[:, 1], triangles[:, 0])),
        np.vstack((triangles[:, 1], triangles[:, 2])),
        np.vstack((triangles[:, 2], triangles[:, 0])),
        np.vstack((triangles[:, 2], triangles[:, 1])),
        ))
    weights = np.ones(edges.shape[1])
    
    connectivity = coo_matrix((weights, edges), (n_points, n_points)).tocsr()

    # Making it symmetrical
    connectivity = (connectivity + connectivity.T) / 2
    return connectivity
```

examples/02_decoding/plot_haxby_searchlight_surface.py
```diff
@@ -0,0 +1,84 @@
...
+# Average voxels around 3 mm of the surface
+radius = 3.
+pial_mesh = fsaverage['pial_' + hemi]
+X = surface.vol_to_surf(fmri_img, pial_mesh, radius=radius).T
+
+# Define the adjacency of the surface vertices
+infl_mesh = fsaverage['infl_' + hemi]
+coords, _ = surface.load_surf_mesh(infl_mesh)
+nn = neighbors.NearestNeighbors(radius=radius)
+A = nn.fit(coords).radius_neighbors_graph(coords).tolil()
```
bthirion, line 54: Could we call this `adjacency` rather than `A` ?

kingjr:
thanks @bthirion. You want this mesh_to_graph in the example, or as a
solution to a subsequent PR?

bthirion:
No, it is not necessary for that PR. it would actually make the example less readable/simple.

Also, I would like to hear from @SylvainTakerkart whether he thinks it's a good way to define searchlights. You know I'm not a big expert in searchlights ;-)

SylvainTakerkart:
> No, it is not necessary for that PR. it would actually make the example less readable/simple.
> 
> Also, I would like to hear from @SylvainTakerkart whether he thinks it's a good way to define searchlights. You know I'm not a big expert in searchlights ;-)

@bthirion using the topology of the mesh is definetely a good way to define a neighborhood on the surface, I'd actually asked myself whether I should do this with my original implementation... but users will still want to change the size of the neighborhood, they will not stick to the 6-closest vertices... if it's not by specifying a radius in mm, it will be by choosing the "maximum order" of the included neighboring vertices when using the topology as you propose (i.e the path length necessary to go from a vertex to the central vertex)...

there's no perfect solution here, but my hunch would be to stay with a radius defined on the freesurfer sphere: i) it has some meaning at the group level, which is a nice property for group analyses (i.e the size of the neighborhood in native space will be a bit different between subjects and between locations, but for a good reason, i.e to maximise the matching between subjects in the freesurfer registration sense); ii) it is invariant with respect to the chosen space (fsaverage vs. fsaverage5; whereas the physical size of the neighborhood would change when working with the same  "topological-order" between fsaverage and fsaverage5); iii) it stays somehow intuitive coz it's in mm (even if somehow misleading coz it's mm on the sphere, as mentioned previously), while the "topological-order" would be more abstract...

WDYT?

thomasbazeille:
This is great work and almost ready to include in the upcoming release, just for the few CircleCI checks not passing ?

Would you have time to give a look at @kingjr ?

kingjr:
@thomasbazeille thanks. I *think* the CI errors. are. not. due to me. I. just rebased, but dont touch `datasets/neurovault.py` which seem to trigger the error. If so, it's ok to merge on my side.

thomasbazeille:
Thanks a lot @kingjr, the rebase did the job tests are passing. Can you merge this if it's ok with you @bthirion ?

GaelVaroquaux:
Hurray!! Thanks!!

PR #2086, opened 2019-07-15
Adding a fetcher for the functional surface data used in the surface searchlight example

SylvainTakerkart:
Hi guys, as planned during the sprint in April, here is a PR to create a fetcher for the functional data that is used in the example of the surface-based searchlight...

The data fetched here is a set of single-trial beta maps from one subject of the InterTVA dataset (https://openneuro.org/datasets/ds001771/), projected on the fsaverage5 surface.

PR #2436, opened 2020-05-12
[MRG] mock 2 osf downloads

jeromedockes:
this fetcher is failing intermittently; mocking the download


emdupre:
@jeromedockes could you share the exact fetcher error ? Wondering if it's like the old OSF one we patched last time.

I do think we should mock the downloads, generally, and have the dataset fetching tested in a separate build step.. This is a great step towards that, thank you !

jeromedockes:
> @jeromedockes could you share the exact fetcher error ? Wondering if it's like the old OSF one we patched last time.
> 
> I do think we should mock the downloads, generally, and have the dataset fetching tested in a separate build step.. This is a great step towards that, thank you !

the error is this one: "[Uncompress] unknown archive file format ..."

I cannot show you the osf response without pushing a change that logs it because
I cannot reproduce locally but I believe OSF is blocking the request from the CI
provider's machine (the problem you are mentioning) and sending a bad response
without setting an error status code, then the fetcher stores that and tries to
decompress something that is not a zip file and fails with this message.

jeromedockes:
note the example that fails in circle ci needs to download the localizer dataset from osf so it will keep failing , and the travis doctest failures are those fixed by #2439

jeromedockes:
Also, the circle ci will keep failing until downloads actually start working because building the documentation needs the datasets. @emdupre and @kchawla-pi what did you do last time, did you contact osf somehow?

kchawla-pi:
To counter CircleCI load, I ran it locally with my the datasets already
downloaded.
For datasets, I pinned the dataset to an earlier cached copy in Circle CI
config so it will not do new downloads.

jeromedockes:
> To counter CircleCI load, I ran it locally with my the datasets already
> downloaded.
> For datasets, I pinned the dataset to an earlier cached copy in Circle CI
> config so it will not do new downloads.

thank you! could you also give this pr a quick review? cheers

emdupre:
> I cannot show you the osf response without pushing a change that logs it because
I cannot reproduce locally but I believe OSF is blocking the request from the CI
provider's machine (the problem you are mentioning) and sending a bad response
without setting an error status code, then the fetcher stores that and tries to
decompress something that is not a zip file and fails with this message.

Yes, this is exactly what I'm wondering. Last time we patched it on our own without OSF (see: https://github.com/nilearn/nilearn/pull/2290), but I could consistently reproduce it locally. I just tried with this dataset and I'm also seeing a 302 link in browser, so no indication that something is amiss....

What's puzzling from the logs is that it seems to download the archive correctly but then fail when decompressing.

GaelVaroquaux:
I guess that we probably don't have a choice, and we need to go this way because we cannot continue working with failing downloads.

So +1 for now.

emdupre:
It looks like it's currently hanging on the download of the language localizer :

```
Dataset created in /home/circleci/nilearn_data/fMRI-language-localizer-demo-dataset

Downloading data from https://osf.io/nh987/download ...

Downloaded 32768 of 749503182 bytes (0.0%, 500.6min remaining)
Downloaded 131072 of 749503182 bytes (0.0%, 229.6min remaining)
...
Downloaded 8929280 of 749503182 bytes (1.2%, 271.5min remaining)
```

Anyone mind if I try restarting the job ?

jeromedockes:
> Anyone mind if I try restarting the job ?

you can try! most likely it will hang up again

kchawla-pi:
This looks okay.
You should pin the dataset cache in CircleCI to an earlier successful job, or I can do it.
Lemme know if you would like that.

jeromedockes:
> This looks okay.
> You should pin the dataset cache in CircleCI to an earlier successful job, or I can do it.
> Lemme know if you would like that.

ah sorry I had done it in #2442 but I can do it all here. I suggest that as a fallback we load any available nilearn_data cache, (at the moment it only matches if it is from the same week). let me know if that's ok with you and then we can merge this one

jeromedockes:
unless there are comments from anybody I will merge this tonight because we need circleci to be running this week

kchawla-pi:
@jeromedockes Can you look at this one also? I believe all thi PR needs is a mocker for the dataset fetcher tester.
https://github.com/nilearn/nilearn/pull/2086#discussion_r357947744

jeromedockes:
merging after @kchawla-pi 's review of #2442

PR #2439, opened 2020-05-13
[MRG] set config to print_changed_only=False for doctests

jeromedockes:
since the scikit-learn release 0.23 (12 may 2020), by default estimator's `__repr__` only show parameters that are different from the default values:

https://scikit-learn.org/stable/whats_new/v0.23.html#miscellaneous
https://github.com/scikit-learn/scikit-learn/pull/17061

this causes some of nilearn's doctests to fail; this PR sets sklearn's config to
the old behaviour for the affected files. 
(note that for the NiftiMasker at least the doctest in question is meant to show
the parameters to the reader so in this particular case the old behaviour is
indeed what we want)

### DIFF
--- a/doc/introduction.rst
+++ b/doc/introduction.rst
@@ -403,6 +403,7 @@ estimator object, for instance an SVC (`support vector classifier
 <http://scikit-learn.org/stable/modules/svm.html>`_).
 It is first created with the relevant parameters::
 
+    >>> import sklearn; sklearn.set_config(print_changed_only=False)
     >>> from sklearn.svm import SVC
     >>> svc = SVC(kernel='linear', C=1.)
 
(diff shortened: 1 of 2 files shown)


bthirion:
Don't understand the azure failure. Seems related to fsaverage5 stuff ?

jeromedockes:

> Don't understand the azure failure. Seems related to fsaverage5 stuff ?

Yes it is the same problem with OSF. it was happening when downloading
fsaverage5 spheres, now apparently with the localizer dataset as well.

jeromedockes:
Also mocked the localizer download in #2436 but quite a few other tests
download data from OSF and those are likely to fail spuriously.
but it is not related to this PR

jeromedockes:
merging with @bthirion 's approval as this is causing CI to fail on all PRs

PR #2449, opened 2020-05-20
Fix: CanICA ensure valid percentile

ChristianGerloff:
# Description
The calculated threshold in `canica` may throw an error.
The reason is that the percentile of the threshold takes on an undefined value range.

# Approach
The range of the percentile used to calculate the threshold in `canica` is limited to [0,100].

### DIFF
--- a/nilearn/decomposition/canica.py
+++ b/nilearn/decomposition/canica.py
@@ -5,7 +5,7 @@ CanICA
 # Author: Alexandre Abraham, Gael Varoquaux,
 # License: BSD 3 clause
 
-import warnings as _warnings 
+import warnings as _warnings
 import numpy as np
 
 from operator import itemgetter
@@ -218,11 +218,13 @@ class CanICA(MultiPCA):
             abs_ica_maps = np.abs(ica_maps)
             percentile = 100. - (100. / len(ica_maps)) * ratio
             if percentile <= 0:
-                _warnings.warn("Critical threshold (= %s percentile). "
-                              "No threshold will be applied. "
-                              "Threshold should be decreased or "
-                              "number of components should be adjusted." %
-                              str(percentile), UserWarning)
+                _warnings.warn("Nilearn's decomposition module "
+                               "obtained a critical threshold "
+                               "(= %s percentile).\n"
+                               "No threshold will be applied. "
+                               "Threshold should be decreased or "
+                               "number of components should be adjusted." %
+                               str(percentile), UserWarning, stacklevel=4)
             else:
                 threshold = scoreatpercentile(abs_ica_maps, percentile)
                 ica_maps[abs_ica_maps < threshold] = 0.
@@ -234,7 +236,8 @@ class CanICA(MultiPCA):
             if component.max() < -component.min():
                 component *= -1
         if hasattr(self, "masker_"):
-            self.components_img_ = self.masker_.inverse_transform(self.components_)
+            self.components_img_ = self.masker_.inverse_transform(
+                self.components_)
 
     # Overriding MultiPCA._raw_fit overrides MultiPCA.fit behavior
     def _raw_fit(self, data):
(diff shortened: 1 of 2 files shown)


nilearn/decomposition/canica.py
```diff
@@ -216,7 +216,7 @@ def _unmix_components(self, components):
             abs_ica_maps = np.abs(ica_maps)
             threshold = scoreatpercentile(
                 abs_ica_maps,
-                100. - (100. / len(ica_maps)) * ratio)
+                np.maximum(0, 100. - (100. / len(ica_maps)) * ratio))
```
GaelVaroquaux, line 219: How about this: we compute the original formula, which can be negative, and if it is negative, issue a warning, and replace it by zero. This seems the most useful thing to do for users.

Does that make sens?

nilearn/decomposition/tests/test_canica.py
```diff
@@ -146,6 +146,14 @@ def test_threshold_bound():
     # is higher than number of components
     pytest.raises(ValueError, CanICA, n_components=4, threshold=5.)
 
+def test_percentile_range():
+    # Smoke test to ensure a valid percentile for determining
+    # the threshold
+    data, mask_img, components, rng = _make_canica_test_data()
+
+    # use large number of components to stress threshold
```
GaelVaroquaux, line 154: We'll have to catch the warning here.

nilearn/decomposition/canica.py
ChristianGerloff: Thank you. alright. Would it perhaps also be more sensible to set the threshold directly to zero instead of the percentile @GaelVaroquaux ?

GaelVaroquaux: Yes, but it's minor.

The warning message feels important in my eyes, because it improves the usability.

nilearn/decomposition/tests/test_canica.py
```diff
@@ -147,13 +148,17 @@ def test_threshold_bound():
...
-    # Smoke test to ensure a valid percentile for determining
-    # the threshold
-    data, mask_img, components, rng = _make_canica_test_data()
-
-    # use large number of components to stress threshold
-    canica = CanICA(n_components=11, threshold=11.)
-    canica.fit(data)
+    # Smoke test to test warning in case ignored thresholds
+    rng = np.random.RandomState(0)                                           
+    edge_case = rng.randint(low = 1, high = 10)
```
GaelVaroquaux, line 153: PEP8:
```suggestion
    edge_case = rng.randint(low=1, high=10) 
```

nilearn/decomposition/canica.py
```diff
@@ -214,10 +216,16 @@ def _unmix_components(self, components):
...
-                abs_ica_maps,
-                100. - (100. / len(ica_maps)) * ratio)
-            ica_maps[abs_ica_maps < threshold] = 0.
+            percentile = 100. - (100. / len(ica_maps)) * ratio
+            if percentile <= 0:
+                _warnings.warn("Critical threshold (= %s percentile). "
+                              "No threshold will be applied. "
+                              "Threshold should be decreased or "
+                              "number of components should be adjusted." %
+                              str(percentile), UserWarning)
```
GaelVaroquaux, line 225: It useful to use the "stacklevel" argument of warnings.warn so that the warning appears closer to user code:
https://docs.python.org/2/library/warnings.html#warnings.warn
From a quick look, it seems that the right value would be here 3, or maybe 2. You'd have to try it out.

nilearn/decomposition/tests/test_canica.py
```diff
@@ -146,6 +147,18 @@ def test_threshold_bound():
     # is higher than number of components
     pytest.raises(ValueError, CanICA, n_components=4, threshold=5.)
 
+def test_percentile_range():
+    # Smoke test to test warning in case ignored thresholds
+    rng = np.random.RandomState(0)                                           
+    edge_case = rng.randint(low = 1, high = 10)
```
GaelVaroquaux, line 153: PEP8
```suggestion
    edge_case = rng.randint(low=1, high=10) 
```

GaelVaroquaux:
CI is currently not running well. I will merge without, and pray that things go well.

PR #1587, opened 2018-01-05
Adding plot_carpet function

chrisgorgo:
Closes #1585.
image

Most of this code has been written by @oesteban for MRIQC.
This is the initial most basic implementation. More features can be added in subsequent PRs.

What is the expected amount of tests/documentation/examples to get this merged?

PR #2450, opened 2020-05-20
Documentation Fixes

ZviBaratz:
Fixing inconsistencies in function/method parameter definitions styling and touching up the docstrings.
@GaelVaroquaux

PR #2456, opened 2020-05-20
ENH: Add plot_carpet function

### LINKED ISSUE #1585: Add function for creating "carpet plots"
image

Potentially ported from MRIQC (see https://github.com/poldracklab/mriqc/issues/674)

tsalo:
Closes #1585. This continues work from #1587.

Changes proposed:
- Add new function `nilearn.plotting.plot_carpet()` (by @oesteban and @chrisgorgo).
- Add test for `nilearn.plotting.plot_carpet()`
- Add example for `nilearn.plotting.plot_carpet()`


nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,128 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+
+def plot_carpet(img, mask_img=None, detrend=True, output_file=None,
+                figure=None, axes=None, title=None):
+    """Plot an image representation of voxel intensities across time.
+
+    This figure is also known as the "carpet plot" or "Power plot".
+
+    Parameters
+    ----------
+    img : Niimg-like object
```
bthirion, line 1476: Could you write the docstrings as in #2450 ?

nilearn/plotting/img_plotting.py
```diff
@@ -22,22 +22,22 @@
 from scipy import sparse
 from nibabel.spatialimages import SpatialImage
 
+from ..input_data import NiftiMasker
```
tsalo, line 25: It looks like this introduces a circular import problem, so importing leads to the following error:
```shell
>>> from nilearn.plotting import plot_carpet

  File "/Users/tsalo/Documents/tsalo/nilearn/nilearn/plotting/__init__.py", line 43, in <module>
    from .img_plotting import plot_img, plot_anat, plot_epi, \
  File "/Users/tsalo/Documents/tsalo/nilearn/nilearn/plotting/img_plotting.py", line 25, in <module>
    from ..input_data import NiftiMasker
  File "/Users/tsalo/Documents/tsalo/nilearn/nilearn/input_data/__init__.py", line 5, in <module>
    from .nifti_masker import NiftiMasker
  File "/Users/tsalo/Documents/tsalo/nilearn/nilearn/input_data/nifti_masker.py", line 16, in <module>
    from nilearn.reporting import ReportMixin
  File "/Users/tsalo/Documents/tsalo/nilearn/nilearn/reporting/__init__.py", line 32, in <module>
  File "/Users/tsalo/Documents/tsalo/nilearn/nilearn/reporting/glm_reporter.py", line 27, in <module>
    from nilearn.plotting import (plot_glass_brain,
ImportError: cannot import name 'plot_glass_brain'
```

tsalo: I think moving the import into the function would fix this, but it's a bit of a hack.

nilearn/plotting/img_plotting.py
tsalo: I think it should be good now, unless I missed something in #2450.

nilearn/plotting/img_plotting.py
bthirion: Or maybe move that to reporting rather than plotting ?

nilearn/plotting/img_plotting.py
bthirion: thx.

nilearn/plotting/tests/test_img_plotting.py
```diff
@@ -245,6 +245,25 @@ def test_plot_glass_brain_threshold_for_uint8():
...
+    affine = np.eye(4)
+    img = nibabel.Nifti1Image(data, affine)
+    mask_data = np.ones((10, 10, 10), int)
+    mask_img = nibabel.Nifti1Image(mask_data, affine)
+    display = plot_carpet(img, mask_img, detrend=False)
+    # Next two lines retrieve the numpy array from the plot
+    ax = display.axes[0]
+    plotted_array = ax.images[0].get_array()
+    # Make sure that the values in the figure match the values in the image
+    assert abs(plotted_array.sum() - data.sum()) < 1e-3
```
bthirion, line 261: Maybe with `assert_almost_equal`

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,130 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+        nifti_masker.fit(img_nii)
+        mask_data = nifti_masker.mask_img_.get_data().astype(bool)
+    else:
+        mask_nii = _utils.check_niimg_3d(mask_img, dtype='auto')
+        mask_data = _safe_get_data(mask_nii, ensure_finite=True)
+
+    data = img_data[mask_data > 0].reshape(-1, ntsteps)
+    # Detrend data
+    if detrend:
+        data = clean(data.T, t_r=tr).T
```
bthirion, line 1532: this also does standardization, right ?
should be indicated somewhere

nilearn/plotting/img_plotting.py
tsalo: I'll move the import for now, but what if I switched from `NiftiMasker` to `compute_epi_mask`?

nilearn/plotting/img_plotting.py
tsalo: In the original PR's review, I saw https://github.com/nilearn/nilearn/pull/1587#discussion_r172327932. Since I didn't write the original code, I'm not sure if standardization is desired (in which case I'll be more explicit), or if it's an unintended effect (in which case I can disable standardization in `clean` call).

Would be good to get @oesteban's opinion on this.

bthirion:
I'd say that standardizing before plotting is good to avoid having the eye caught by spurious details.

nilearn/plotting/img_plotting.py
tsalo: > I'd say that standardizing before plotting is good to avoid having the eye caught by spurious details.

Okay, I be more explicit in the call and also add a comment. It should make the test easier too.

tsalo:
I noticed in the tests that there is the following:

https://github.com/nilearn/nilearn/blob/9f47ddf4a5253245aa12b01acb48dd83df03104c/nilearn/plotting/tests/test_img_plotting.py#L34-L39

This function is called in almost all of the tests for img_plotting. Could it just be converted to a fixture or is there a reason it isn't already?

nilearn/plotting/img_plotting.py
bthirion: Thinking a bit further, the circular import tells us that this would probably better fit  in reporting._plot_matrices than here.

tsalo: I was able to fix it without an import within the function by switching from `NiftiMasker` to `compute_epi_mask` and `apply_mask`. Since none of the extra functionality of `NiftiMasker` was being used in the masking stage (i.e., standardization/detrending was done _after_), it seemed like a good choice. That said, do you still want me to move it to `reporting`?

Also, if I move the function to reporting, should I also move the example from "Visualization of brain images" to another section?

tsalo: It might be best to continue this thread in #2488, given its relevance.

bthirion:
For the time being let's keep that in nilearn.plotting :-)

tsalo:
Awesome! I think the basic function is ready then. The current CI error seems to be unrelated to the changes in the PR. If you have permissions, would you mind re-running the workflow at some point?

There are a number of bells and whistles that were discussed in the associated issue (e.g., voxel decimation, atlas-based masking, and sorting to emphasize patterns). Do you want to have those implemented in this PR or just get the basic method implemented and then expand its capabilities in future PRs?

bthirion:
I think that we're good for that PR. 
Are you suggesting to add an `atlas` argument to the function ?

tsalo:
I think an atlas should be supported, although it could just be done under the hood in cases where the mask img is not binary, right?

I'm not sure if there are any other important features that should be included in this PR though.

GaelVaroquaux:
The build failed, and I believe that this is because plot_carpet needs to be added to docs/modules/reference.rst

examples/01_plotting/plot_carpet.py
```diff
@@ -0,0 +1,43 @@
...
+# -------------------------------------
+import matplotlib.pyplot as plt
+
+from nilearn.plotting import plot_carpet
+
+fig, ax = plt.subplots(figsize=(10, 5))
+
+fig = plot_carpet(adhd_dataset.func[0], mask_img, figure=fig, axes=ax)
+
+fig.tight_layout()
```
GaelVaroquaux, line 42: Ideally, I think that we should try to work on the defaults of plot_carpet so that this example looks fine without creating the figure and calling tight_layout. Maybe a figure size computed as a function of the data (with min and max limits to avoid pathological situations)?

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+
+    Notes
+    -----
+    This figure was originally developed in [1]_.
+
+    References
+    ----------
+    .. [1] Power, J. D. (2017). A simple but useful way to assess fMRI scan
+    qualities. Neuroimage, 154, 150-158. doi:
+    https://doi.org/10.1016/j.neuroimage.2016.08.009
```
GaelVaroquaux, line 1512:
```suggestion
            qualities. Neuroimage, 154, 150-158. doi:
            https://doi.org/10.1016/j.neuroimage.2016.08.009
```
The indentation, seems wrong to me.

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+    """
+    img = _utils.check_niimg_4d(img, dtype='auto')
+
+    # Define TR and number of frames
+    tr = img.header.get_zooms()[-1]
+    ntsteps = img.shape[-1]
+
+    if not mask_img:
```
GaelVaroquaux, line 1520:
```suggestion
    if mask_img is None:
```
It's preferable to use this: suppose someone passes in an array, you will get a strange error message.

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+        mask_img = compute_epi_mask(img)
+    else:
+        mask_img = _utils.check_niimg_3d(mask_img, dtype='auto')
+
+    data = apply_mask(img, mask_img)
+    # Detrend and standardize data
+    if detrend:
+        data = clean(data, t_r=tr, detrend=True, standardize='zscore')
+
+    if not figure:
```
GaelVaroquaux, line 1530:
```suggestion
    if figure is None:
```

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+        if not axes:
+            figure = plt.figure()
+        else:
+            figure = axes.figure
+
+    if not axes:
```
GaelVaroquaux, line 1536:
```suggestion
    if axes is None:
```

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+        mask_img = _utils.check_niimg_3d(mask_img, dtype='auto')
+
+    data = apply_mask(img, mask_img)
+    # Detrend and standardize data
+    if detrend:
+        data = clean(data, t_r=tr, detrend=True, standardize='zscore')
+
+    if not figure:
+        if not axes:
+            figure = plt.figure()
```
GaelVaroquaux, line 1532: We could change the default figure size here.

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+        The title displayed on the figure (default is None).
+
+    Returns
+    -------
+    figure : :class:`matplotlib.figure.Figure`
+        Figure object with carpet plot.
+
+    Notes
+    -----
+    This figure was originally developed in [1]_.
```
GaelVaroquaux, line 1506: Add here a sentence explaining the decimation.

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+    if not axes:
+        axes = figure.add_subplot(1, 1, 1)
+    else:
+        assert axes.figure is figure, ("The axes passed are not "
+                                       "in the figure")
+
+    # Avoid segmentation faults for long acquisitions by decimating the data
+    LONG_CUTOFF = 800
+    if ntsteps > LONG_CUTOFF:
+        data = data[::2, :]
```
GaelVaroquaux, line 1545: Should we do a while loop here, to decimate more if the data are really huge?
```suggestion
    while ntsteps > LONG_CUTOFF:
        data = data[::2, :]
        ntsteps = len(data)
```

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+    ----------
+    .. [1] Power, J. D. (2017). A simple but useful way to assess fMRI scan
+    qualities. Neuroimage, 154, 150-158. doi:
+    https://doi.org/10.1016/j.neuroimage.2016.08.009
+    """
+    img = _utils.check_niimg_4d(img, dtype='auto')
+
+    # Define TR and number of frames
+    tr = img.header.get_zooms()[-1]
+    ntsteps = img.shape[-1]
```
GaelVaroquaux, line 1518: Suggestion of a rename (I find is easier to parse):
```suggestion
    n_tsteps = img.shape[-1]
```

nilearn/plotting/img_plotting.py
```diff
@@ -1463,3 +1463,123 @@ def plot_connectome_strength(adjacency_matrix, node_coords, node_size="auto",
...
+        assert axes.figure is figure, ("The axes passed are not "
+                                       "in the figure")
+
+    # Avoid segmentation faults for long acquisitions by decimating the data
+    LONG_CUTOFF = 800
+    if ntsteps > LONG_CUTOFF:
+        data = data[::2, :]
+
+    axes.imshow(data.T, interpolation='nearest',
+                aspect='auto', cmap='gray', vmin=-2, vmax=2)
```
GaelVaroquaux, line 1548: The vmin and the vmax seem very arbitrary to me. Images could be scaled very differently.

nilearn/plotting/tests/test_img_plotting.py
```diff
@@ -31,12 +31,39 @@
...
+def _generate_img_4d():
+    """A random 4D image for testing figures for multivolume data.
+    """
+    rng = np.random.RandomState(42)
+    data = rng.rand(7, 7, 3, 10)
+    img_4d = nibabel.Nifti1Image(data, mni_affine)
+    pytest.img_4d = img_4d
+
+
+@pytest.fixture(autouse=True)
```
GaelVaroquaux, line 56: I must confess that I find the autouse fixture very implicit and hence hard to read for people who do not know the feature. How easy would it be to use explicit fixtures, rather?

nilearn/plotting/tests/test_img_plotting.py
```diff
@@ -31,12 +31,39 @@
...
+def _generate_img_4d_long():
+    """A random 4D image for testing figures for multivolume data.
+    """
+    rng = np.random.RandomState(42)
+    data = rng.rand(7, 7, 3, 997)
+    img_4d = nibabel.Nifti1Image(data, mni_affine)
+    mask = np.ones((7, 7, 3), int)
+    img_4d_mask = nibabel.Nifti1Image(mask, mni_affine)
+    pytest.img_4d_long = img_4d
+    pytest.img_4d_mask = img_4d_mask
```
GaelVaroquaux, line 66: Monkey patching pytest is also not terribly elegant, in my eye :)

nilearn/plotting/tests/test_img_plotting.py
```diff
@@ -100,22 +127,32 @@ def test_plot_anat():
...
-    # smoke-test for each plotting function with default arguments
+    # smoke-test for 3D plotting functions with default arguments
     for plot_func in [plot_anat, plot_img, plot_stat_map, plot_epi,
                       plot_glass_brain]:
         filename = tempfile.mktemp(suffix='.png')
         try:
-            plot_func(img, output_file=filename)
+            plot_func(img_3d, output_file=filename)
+        finally:
+            os.remove(filename)
```
GaelVaroquaux, line 141: :+1:

nilearn/plotting/img_plotting.py
tsalo: Do you want me to add them as arguments?

GaelVaroquaux:
> Do you want me to add them as arguments?

Yes but with as default "None", where it creates good defaults from the
image values.

nilearn/plotting/img_plotting.py
tsalo: I figured I could do the following to avoid a loop:

```python
n_decimations = int(np.ceil(np.log(np.ceil(n_tsteps / LONG_CUTOFF))))
data = data[::2 ** n_decimations, :]
```

examples/01_plotting/plot_carpet.py
tsalo: The axis labels are shifted out to give it whitespace between the axes and the data. I don't know of a way to account for that at the point where the fig object is created, so I think either the labels have to be up against the data (which doesn't look as good) or tight_layout needs to be used in the function.

nilearn/plotting/img_plotting.py
tsalo: Done!

nilearn/plotting/img_plotting.py
tsalo: I've added them as arguments and switched to calculating M +/- 2SD from the data, since the +/- 2 were good defaults for _z-scored_ data. This should work well when `detrend=False` as well.

nilearn/plotting/tests/test_img_plotting.py
tsalo: I've dropped autouse and just pass the fixtures as arguments to the tests.

nilearn/plotting/tests/test_img_plotting.py
tsalo: Now they're just passed as arguments and don't need to be accessed through pytest.

GaelVaroquaux:
The build of the doc is failing because the AxesSubplot is not documented with the import path `matplotlib.axes._subplots.AxesSubplot` (a quick googling does not tell where / how it is documented so you could just give up on linking to it).

doc/plotting/index.rst
```diff
@@ -119,6 +123,10 @@ different heuristics to find cutting coordinates.
                      |hack|
                      Plotting 4D probabilistic atlas maps
 
+|plot_prob_atlas|    :func:`plot_carpet`
```
GaelVaroquaux, line 126:
```suggestion
|plot_carpet|    :func:`plot_carpet`
```
I think that there is a copy-and-paste error above

examples/01_plotting/plot_carpet.py
```diff
@@ -5,7 +5,7 @@
 A common quality control step for functional MRI data is to visualize the data
 over time in a carpet plot (also known as a Power plot or a grayplot).
 
-The :func:`nilearn.plotting.plot_carpet()` function generates a carpet plot
+The :py:func:`nilearn.plotting.plot_carpet()` function generates a carpet plot
```
GaelVaroquaux, line 8: Give that nilearn is pure Python, I don't think that this is necessary.

examples/01_plotting/plot_carpet.py
```diff
@@ -35,9 +35,6 @@
...
 from nilearn.plotting import plot_carpet
 
-fig, ax = plt.subplots(figsize=(10, 5))
+display = plot_carpet(adhd_dataset.func[0], mask_img)
 
-fig = plot_carpet(adhd_dataset.func[0], mask_img, figure=fig, axes=ax)
-
-fig.tight_layout()
-fig.show()
+display.show()
```
GaelVaroquaux, line 40: I'm super happy to see this example become shorter!

examples/01_plotting/plot_carpet.py
```diff
@@ -0,0 +1,40 @@
+"""
+Visualizing global patterns with a carpet plot
+==============================================
+
+A common quality control step for functional MRI data is to visualize the data
+over time in a carpet plot (also known as a Power plot or a grayplot).
+
+The :py:func:`nilearn.plotting.plot_carpet()` function generates a carpet plot
```
tsalo, line 8:
```suggestion
The :func:`nilearn.plotting.plot_carpet()` function generates a carpet plot
```

doc/whats_new.rst
```diff
@@ -16,6 +16,10 @@ NEW
   The resulting average model is the one used as a classifier or a regressor.
   These two objects also leverage the `NiftiMaskers` to provide a direct
   interface with the Nifti files on disk.
+- New plotting function
+  `nilearn.plotting.plot_carpet` generates a "carpet plot" (also known
```
GaelVaroquaux, line 20: This should have been :func:`nilearn.plotting.plot_carpet` to have a link to the API doc. I still that I'll merge and change this in master, to save everybody some time.

PR #2461, opened 2020-05-20
set a timeout on socket connections (8 s)

jeromedockes:


jeromedockes:
ping @emdupre if you could do a quick review on this one it will prevent an example from hanging waiting for neurosynth for builds that run the examples

jeromedockes:
or anyone else eg @GaelVaroquaux

examples/07_advanced/plot_ica_neurovault.py
```diff
@@ -39,13 +39,17 @@
...
       "this will take several minutes.")
-nv_data = fetch_neurovault(max_images=80, fetch_neurosynth_words=True)
+nv_data = fetch_neurovault(max_images=30, fetch_neurosynth_words=True)
 
 images = nv_data['images']
 term_weights = nv_data['word_frequencies']
 vocabulary = nv_data['vocabulary']
+if term_weights is None:
+    term_weights = np.ones((len(images), 2))
+    vocabulary = np.asarray(["Neurosynth is down", ":("])
```
emdupre, line 49: Love the emoji usage. I was somewhat worried about it as a potential source of confusion, but it looks like it's not returned to the user ?

examples/07_advanced/plot_ica_neurovault.py
```diff
@@ -39,13 +39,17 @@
...
 
 # Clean and report term scores
+print(term_weights)
```
emdupre, line 52: Should this be printed to screen or returned ?

nilearn/datasets/utils.py
```diff
@@ -14,9 +14,12 @@
 import sys
 import tarfile
 import urllib
+import socket
 import warnings
 import zipfile
 
+socket.setdefaulttimeout(8)
+
```
emdupre, line 22: Extra white space ?

examples/07_advanced/plot_ica_neurovault.py
jeromedockes: no, it will appear as the top terms shown on top of the IC plots

examples/07_advanced/plot_ica_neurovault.py
jeromedockes: this is not in a function but in the example. but good point! there is no reason to print it

nilearn/datasets/utils.py
jeromedockes: Do you think so? there should be 2 blank lines before the function that comes below, right? flake8 doesn't seem to complain. anyway, happy to change it if you prefer

emdupre: Sorry about that, you're absolutely right. Ignore this comment :smile_cat:

GaelVaroquaux:
LGTM. However, I think that I'd still like to wait for the build to finish before merging :)

examples/07_advanced/plot_ica_neurovault.py
emdupre: OK, light suggestion to be explicit :point_down: 

```suggestion
    vocabulary = np.asarray(["Neurosynth is down", "Please try again later"])
```

jeromedockes:
> LGTM. However, I think that I'd still like to wait for the build to finish before merging :)

good idea. and I'll make sure it doesn't load the cache. I will cancel the build of your last commit on master because it will hang without this pr

jeromedockes:
> Last comment is an opinion, please feel free to merge when tests pass !

you're right, I changed it to something more explicit

examples/07_advanced/plot_ica_neurovault.py
jeromedockes: sounds good, I'll change that but won't push now to avoid restarting the build

PR #2469, opened 2020-05-21
WIP: Inverse spheres masker

### LINKED ISSUE #1192: extracting voxels timeseries within a sphere and using inverse_transform with NiftiSpheresMasker
Is there an appropriate way to extract voxels time series within spheres ? I am using  `iter_signals_from_spheres`

``` Python
import nibabel
from nilearn import datasets, input_data
from nilearn.input_data.nifti_spheres_masker import _iter_signals_from_spheres
adhd = datasets.fetch_adhd(n_subjects=1)
func_filename = adhd.func[0]
seeds_masker = input_data.NiftiSpheresMasker([(-6, 50, -1)], radius=6.)
seeds_masker.fit()
sphere_timeseries = _iter_signals_from_spheres(
    seeds_masker.seeds_, nibabel.load(func_filename),
    seeds_masker.radius,
    seeds_masker.allow_overlap).next()
```

I also want to go from array to image using `inverse_transform` but I got an error

``` Python
image = seeds_masker.inverse_transform(sphere_timeseries.T)
```

> 'NiftiSpheresMasker' object has no attribute 'mask_img_'

SRSteinkamp:
Closes #1192 .

Implements `inverse_transform` for the NiftiSpheresMasker to visualize data from extracted spheres on the brain volume.


nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,56 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
     return X, A
 
 
+def _apply_mask_get_rows(seeds, affine, target_shape, radius, allow_overlap,
+                                 mask_img=None):
+    seeds = list(seeds)
```
bthirion, line 91: Please provide a concise docstring explaing what this does.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,56 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+        except ValueError:
+            nearests.append(None)
+
+    mask_coords = np.asarray(list(zip(*mask_coords)))
+    mask_coords = coord_transform(mask_coords[0], mask_coords[1],
+                                  mask_coords[2], affine)
+    mask_coords = np.asarray(mask_coords).T
+
+    clf = neighbors.NearestNeighbors(radius=radius)
+    A = clf.fit(mask_coords).radius_neighbors_graph(seeds)
```
bthirion, line 113: Could we call it 'adjacency' rather than 'A' ?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,56 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+    A = A.tolil()
+    for i, nearest in enumerate(nearests):
```
bthirion, line 115: Why are you doing this loop ?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,56 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+        if nearest is None:
+            continue
+        A[i, nearest] = True
+
+    # Include the voxel containing the seed itself if not masked
```
bthirion, line 120: I don't understand how it could be the case ?

nilearn/input_data/nifti_spheres_masker.py
bthirion: Also, for internal functions, you usually don't use kw arguments

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -115,6 +166,37 @@ def _iter_signals_from_spheres(seeds, niimg, radius, allow_overlap,
...
 
 
+def _iter_regions_from_spheres(seeds, affine, target_shape, radius, allow_overlap,
+                               mask_img=None):
+    """Utility function to iterate over spheres.
+    Parameters
+    ----------
+    seeds: List of triplets of coordinates in native space
+        Seed definitions. List of coordinates of the seeds in the same space
+        as the images (typically MNI or TAL).
```
bthirion, line 176: is TAL defined anywhere ?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -115,6 +166,37 @@ def _iter_signals_from_spheres(seeds, niimg, radius, allow_overlap,
...
+    radius: float
+        Indicates, in millimeters, the radius for the sphere around the seed.
+    allow_overlap: boolean
+        If False, an error is raised if the maps overlaps (ie at least two
+        maps have a non-zero value for the same voxel).
+    mask_img: Niimg-like object, optional
+        See http://nilearn.github.io/manipulating_images/input_output.html
+        Mask to apply to regions before extracting signals.
+    """
+    A = _apply_mask_get_rows(seeds, affine, target_shape, radius,
```
bthirion, line 190: i'd also rename `A` to `adjacency` or so.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -138,6 +220,51 @@ def __call__(self, imgs):
         return signals, None
 
 
+class _InversionFunctor(object):
```
bthirion, line 223: Sorry if I misunderstood, but what is the point in making this a class rather than a function ?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -138,6 +220,51 @@ def __call__(self, imgs):
...
+        self.radius = radius
+        self.mask_img = mask_img
+        self.allow_overlap = allow_overlap
+        self.dtype = dtype
+        # Mask is only used to get affine and shape
+        if self.mask_img is not None:
+            mask = check_niimg_3d(self.mask_img)
+            self.target_affine = mask.affine
+            self.target_shape = mask.shape
+        else:
```
bthirion, line 238: Sorry, but this is black magic: we 're figuring out an affine while there is None. I think we should rather spit an error here (with a clear message) ?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -138,6 +220,51 @@ def __call__(self, imgs):
...
+                self.seeds_, self.target_affine, self.target_shape, self.radius,
+                self.allow_overlap, mask_img=self.mask_img)):
+            spheres[sphere, i] = 1
+
+        # Compute overlap scaling for mean signal:
+        if self.allow_overlap:
+            spheres_sum = spheres.sum(1)
+            # signals_to_img_maps uses dot product to create mask over regions,
+            # inverse scaling to create average not sum
+            spheres_scale = 1 / spheres_sum[spheres_sum > 1, np.newaxis]
```
bthirion, line 261: aren't we going to have divisions by 0 here ?

nilearn/input_data/tests/test_nifti_spheres_masker.py
```diff
@@ -178,3 +178,88 @@ def test_standardization():
...
+    mask[:, 1, 1] = True
+    mask[1, :, 1] = True
+    mask[1, 1, :] = True
+
+    inverse_map = masker.inverse_transform(s)
+    # Testing whether mask is applied to inverse transform, and has same length as 
+    # the masked data and data is preserved
+    assert_array_equal(get_data(inverse_map)[get_data(inverse_map) != 0], s[:, 0])
+    # No masking is applied - check dimensions with regard to mni template
+    assert_array_equal(inverse_map.shape[:3], [61, 73, 61])
```
bthirion, line 203: I'd rather not have the case without mask.

nilearn/input_data/tests/test_nifti_spheres_masker.py
```diff
@@ -178,3 +178,88 @@ def test_standardization():
...
+    overlapping_masker.inverse_transform(inv_data)
+
+    overlapping_masker = NiftiSpheresMasker(seeds, radius=2,
+                                            allow_overlap=True,
+                                        mask_img=mask_img).fit()
+
+    overlap = overlapping_masker.inverse_transform(inv_data)
+
+    # Test whether overlapping data is averaged
+    assert np.isclose(get_data(overlap)[1,1,1], np.mean(inv_data))
```
bthirion, line 253:
```suggestion
    assert_array_almost_equal(get_data(overlap)[1, 1, 1], np.mean(inv_data))
```

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: Great for me. This was the big trouble I had with the `.inverse_transform()` in the first place.

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: In many cases I was just following the extraction function - so there might be some redundancy here.

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: I don't think so. `spheres_sum` is only selection overlapping areas. I.e. areas which sums are > 1.  Only for these points the division is done, definitely open to suggestions, but that seemed the easiest way for me to create the average signal of overlapping voxels.

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: Yes, also for some of the points below: I was just repurposing the other code for the extraction of sphere's data. So I basically copied a lot over from there, which was involved in the creation of spheres.

nilearn/input_data/tests/test_nifti_spheres_masker.py
SRSteinkamp: I will implement the error (above) and test for it here then.

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: Will check, whether this is necessary.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,63 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
     return X, A
 
 
+def _get_spheres_rows(seeds, target_affine, target_shape, radius, allow_overlap):
+    '''Utility function to get only the rows which are occupied by sphere at
+    given seed locations and the provided radius. Rows are in affine and
```
bthirion, line 91:
```suggestion
    given seed locations and the provided radius. Rows are in target_affine and 
```

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,63 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+    seeds_: List of triplets of coordinates in native space
+        Seed definitions. List of coordinates of the seeds in the same space
+        as target_affine.
+    target_affine: np.ndarray
+        Affine for the spheres.
+    target_shape: np.array or list
+        Target shape into which the spheres will be projected. 
+    radius: float
+        Indicates, in millimeters, the radius for the sphere around the seed.
+    mask_img: Niimg-like object
```
bthirion, line 105: this argument does not exist...

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -115,6 +173,33 @@ def _iter_signals_from_spheres(seeds, niimg, radius, allow_overlap,
...
+                               allow_overlap):
+    """Utility function to iterate over spheres. Used in NiftiSpheresMasker()
+    .inverse_transform()
+
+    Parameters
+    ----------
+    seeds: List of triplets of coordinates in native space
+        Seed definitions. List of coordinates of the seeds in the same space
+        as target_affine (either of mask or if mask is None MNI space).
+    target_affine: Affine in which the regions/spheres should be projected to
```
bthirion, line 186:
```suggestion
    target_affine: np.ndarray of shape (4,4)
                         Affine in which the regions/spheres should be projected to
```

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -115,6 +173,33 @@ def _iter_signals_from_spheres(seeds, niimg, radius, allow_overlap,
...
+    target_shape: The shape of the output image.
```
bthirion, line 187: Same thing

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -115,6 +173,33 @@ def _iter_signals_from_spheres(seeds, niimg, radius, allow_overlap,
         yield X[:, row]
 
 
+def _iter_regions_from_spheres(seeds, target_affine, target_shape, radius,
```
bthirion, line 176: providing mask_img as an alternative to `target_affine` and `target_shape` would be better I think

SRSteinkamp:
Thank you for your feedback! 
I wasn't sure so I still left the target_affine and target_shape arguments for the _get_spheres_rows function.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -183,15 +180,15 @@ def _iter_regions_from_spheres(seeds, target_affine, target_shape, radius,
...
         If False, an error is raised if the maps overlaps (ie at least two
         maps have a non-zero value for the same voxel).
+    mask_img: Niimg-like object,
+        Mask as a reference space to project the spheres back into brain space.
     """
-    adjacency = _get_spheres_rows(seeds, target_affine, target_shape, radius,
+    adjacency = _get_spheres_rows(seeds, mask_img.affine, mask_img.shape, radius,
                           allow_overlap)
 
     for row in adjacency.rows:
```
SRSteinkamp, line 194: I'm not sure if there are cases where this Error will be raised, and I am not sure how big of an issue this is. 

The inverse tranform might create sphere's which are possibly completely masked in the last-step. Without any warning what-so-ever, whereas an Error will occur in the classic transform.

I am not sure (yet) how to implement masking during _get_spheres_rows, but I can give it a try - If it's necessary

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -115,6 +171,32 @@ def _iter_signals_from_spheres(seeds, niimg, radius, allow_overlap,
...
+    mask_img: Niimg-like object,
+        Mask as a reference space to project the spheres back into brain space.
+    """
+    adjacency = _get_spheres_rows(seeds, mask_img.affine, mask_img.shape, radius,
+                          allow_overlap)
+
+    for row in adjacency.rows:
+        if len(row) == 0:
+            raise ValueError('Sphere around seed #%i is empty' % i)
+        yield row
```
SRSteinkamp, line 197: I'm not sure if there are cases where this Error will be raised, and I am not sure how big of an issue this is.  (i'm referencing lines 195ff (but somehow cannot comment on then)

The inverse tranform might create sphere's which are possibly completely masked in the last-step. Without any warning what-so-ever, whereas an Error will occur in the classic transform.

I am not sure (yet) how to implement masking during _get_spheres_rows, but I can give it a try - If it's necessary

bthirion: > I am not sure (yet) how to implement masking during _get_spheres_rows, but I can give it a try - If it's necessary

Sorry I don't understand  what you want to implement here. maybe I don't understand what you mean witth 'masking'.

SRSteinkamp: In the current implementation, the masker back projeckts spheres into the space of the mask image. But it doesn't take the brain-mask's content into account. So spheres might be outside of the brain mask.
The brain_mask, however, is applied later in the transformation (final step, where I am calling the `input_data.nifti_maps_masker` utils). 

I was wondering whether this might be a  very confusing behavior for the user. The `transform` is checking if `seeds` are inside of `mask_img` (which makes sense, otherwise there'd be no signal), but the `inverse` isn't.
So given the same `mask_img` and `seeds`, you could use the `inverse_transform` to project spheres to the brain (using something like `np.arange` to just visualize the sphere's location), but might be missing certain spheres. The `transform`, however, will fail.

bthirion: OK, thx for the explanations.
Indeed as you say, we should ensure that the behavior of `transform` and `inverse_transform` is consistent. It sounds like you need to factorize some common code between the two methods ?
 Thx again,

SRSteinkamp: I don't think it's as straightforward as that. 

Basically, I tried to repurpose the `transform` code for the `inverse_transform` before (you could see that with the misleading doc-strings and redundant for-loops). 

But I got an idea of how to go on. 

(Sorry I forgot to click on comment..., will tend to it soon)

bthirion:
Thx, simply let us now if you need some more help (I haven't looked in detail so far).
Best,

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -113,13 +113,21 @@ def _get_spheres_rows(seeds, target_affine, target_shape, radius, allow_overlap)
...
 
-    # Compute world coordinates of all voxels. The mask will be applied at 
-    # the last step.
-    brain_coords = list(np.ndindex(*target_shape))
+    # Compute world coordinates of the mask and of all coordinates
+    mask = _safe_get_data(mask_img)
+    
+    mask_coords = list(zip(*np.where(mask == 0)))
+    mask_coords = np.asarray(list(zip(*mask_coords)))
+    if not np.all(mask == 1):
```
SRSteinkamp, line 121: Need to use the data which is NOT in the mask. Basically a reverse of the forward function. 

The `if not np.all(mask==1)`  is unfortunately needed for masks that include the whole brain / image. However, everything else will work with empty lists.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -128,9 +136,18 @@ def _get_spheres_rows(seeds, target_affine, target_shape, radius, allow_overlap)
...
     # Include the voxel containing the seed itself
     brain_coords = brain_coords.astype(int).tolist()
-    for i, seed in enumerate(seeds):
-        adjacency[i, brain_coords.index(seed)] = True
+    # Get the mask indices (in terms of brain coords)
+    mask_idx = [brain_coords.index(i) for i in mask_coords.tolist()]
 
+    for i, seed in enumerate(seeds):
+        try:
+            adjacency[i, brain_coords.index(seed)] = True
```
SRSteinkamp, line 144: As seeds are not integer, but the brain_coords are, the try - catch is needed here. 

Maybe converting `seed` here to `int` would solve this?

nilearn/input_data/tests/test_nifti_spheres_masker.py
```diff
@@ -258,3 +258,35 @@ def test_nifti_spheres_masker_inverse_overlap():
...
+    shape = (3, 3, 3)
+
+    data = np.random.random(shape)
+    mask = np.zeros(shape)
+    mask[1, 1, 1] = 1
+    mask[2, 2, 2] = 1
+    affine = np.eye(4) * 1.2
+    seed = (1.4, 1.4, 1.4)
+
+    masker = NiftiSpheresMasker([seed], radius=0.6,
```
SRSteinkamp, line 274: I'm still having one issue here (I still will need to dig in way more into the code though): 
This code should work with a radius = 0.1 (see the corresponding test for the forward-transform), but I seem to either mask out or not include the seed voxel for a radius <0.4.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -113,13 +113,21 @@ def _get_spheres_rows(seeds, target_affine, target_shape, radius, allow_overlap)
     '''
     seeds = list(seeds)
 
-    # Compute world coordinates of all voxels. The mask will be applied at 
-    # the last step.
-    brain_coords = list(np.ndindex(*target_shape))
+    # Compute world coordinates of the mask and of all coordinates
+    mask = _safe_get_data(mask_img)
```
SRSteinkamp, line 117: I am not sure if I should do more tests on the `mask_img` (as in the transformation)?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,78 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+        as target_affine.
+    target_affine: np.ndarray of shape (4,4)
+        Affine for the spheres.
+    target_shape: np.array or list
+        Target shape into which the spheres will be projected. 
+    radius: float
+        Indicates, in millimeters, the radius for the sphere around the seed.
+    allow_overlap: boolean
+        If False, an error is raised if the maps overlaps (ie at least two
+        maps have a non-zero value for the same voxel).
```
bthirion, line 107: the docstring is not up-to-date.

nilearn/input_data/nifti_spheres_masker.py
bthirion: Logically, you perform any test in the high-level function, not here.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,78 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+    brain_coords = np.asarray(brain_coords).T
+
+    clf = neighbors.NearestNeighbors(radius=radius)
+    adjacency = clf.fit(brain_coords).radius_neighbors_graph(seeds)
+    adjacency = adjacency.tolil()
+
+    # Include the voxel containing the seed itself
+    brain_coords = brain_coords.astype(int).tolist()
+    # Get the mask indices (in terms of brain coords)
+    mask_idx = [brain_coords.index(i) for i in mask_coords.tolist()]
```
bthirion, line 140: this loop looks a bit terrible computationally. Wouldn't it be more efficient to store mask as a binary array and use numpy indexing ?

nilearn/input_data/tests/test_nifti_spheres_masker.py
bthirion: Do you need help on that explicitly ?

SRSteinkamp:
Thank you for alle the feedback :) 
I think I am almost there. The unittests are passing. But with actual data something is still a-miss. 

The `nearests` loop similarly to the forward function is necessary to include the voxels closest to the seed.

I also found a more efficient way to apply the mask-img to the data (yes the list comprehension for loop was indeed terrible performance wise).

As a minimal running example, with the current implementation of the inverse transformation with actual data:
```
from nilearn.input_data import NiftiSpheresMasker
from nilearn.masking import compute_epi_mask
from nilearn.datasets import fetch_development_fmri, fetch_coords_power_2011
test_data = fetch_development_fmri(n_subjects=1)
test_data = nibabel.load(test_data['func'][0])
test_mask = compute_epi_mask(test_data)
coords = fetch_coords_power_2011()
test_coords = list(coords['rois'][['x', 'y', 'z']][0:5])

nsm = NiftiSpheresMasker(test_coords, radius=2, mask_img=test_mask, allow_overlap=True)
nsm.fit()

tmp = nsm.transform(test_data) # This fails as some VOIs are not inside mask

tmp2 = nsm.inverse_transform(np.arange(5)) # Doesn't fail. 
```

I seem to be unable to track down the error. Either I am incorrectly computing the masking indices which I am applying, in the inverse transformation. Or something goes wrong in the masking part (I don't know much about sparse arrays, maybe that's where the issue is...). I am putting comments in the code, where the something unexpected might be happening. 

Thank you again!

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -132,37 +130,39 @@ def _get_spheres_rows(seeds, radius, allow_overlap, mask_img):
             nearests.append(None)
 
     brain_coords = np.asarray(list(zip(*brain_coords)))
-    # Calculate mask
+    # Calculate mask and the corresponding index in flattended array.
     mask_coords = list(zip(*np.where(mask == 1)))
     mask_coords = np.asarray(mask_coords).astype(int).T
     mask_idx = np.vstack([np.in1d(brain_coords[0, :], mask_coords[0, :]),
```
SRSteinkamp, line 136: Possible source of un-expected behavior 1: 
I am calculating the overlapping coordinates between the whole brain and the mask incorrectly. 
Or I am incorrectly extracting the mask coordinates to begin with.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -132,37 +130,39 @@ def _get_spheres_rows(seeds, radius, allow_overlap, mask_img):
...
+            # Include seed itself
             adjacency[i, brain_coords.index(np.array(seed))] = True
         except ValueError:
             # seed is not in the mask
             pass
-        # Removing mask voxels
-        adjacency[i, mask_idx==False] = False
+
+    # Removing voxels not inside of the mask.
+    adjacency[:, mask_idx==False] = False
```
SRSteinkamp, line 165: Possible source 2:
Mask is not applied correctly.

jeromedockes:
> I seem to be unable to track down the error. Either I am incorrectly computing the masking indices which I am applying, in the inverse transformation. Or something goes wrong in the masking part (I don't know much about sparse arrays, maybe that's where the issue is...). I am putting comments in the code, where the something unexpected might be happening. 

What would be the expected behaviour in the example you provided? is the
problem that the `tmp =` line should pass or that the `tmp2 = ` line
should fail?

SRSteinkamp:
`tmp2` should fail. 

Bit more of an explanation:
My goal currently is to make the error messages between `.transform` and `.inverse_transform` consistent.
In the example `nsm.transform(test_data)` is failing, as one or more of the Spheres/Rois/Vois is empty (i.e. voxels of the sphere are not in the mask). 

Therefore, also `tmp2 = nsm.inverse_transform(np.arange(5))` should return an error - Which doesn't happen. 

In this example three VOIs are clearly outside of `mask_img` - even in the inverse they are not displayed (some masking is applied again at a later point, which might be unnecessary).  
So somewhere in `_get_spheres_rows` something goes wrong. 

What is weird for me, however, is that the masking is applied correctly in the unit-test.

jeromedockes:
> In this example three VOIs are clearly outside of `mask_img` - even in the inverse they are not displayed (some masking is applied again at a later point, which might be unnecessary).
> So somewhere in `_get_spheres_rows` something goes wrong.
>
> What is weird for me, however, is that the masking is applied correctly in the unit-test.

Thanks for the explanation. I will try to have a look but a first remark
is that much of `_get_spheres_rows` is copied from the body of
`_apply_mask_and_get_affinity`, do you reckon this could be refactored a
bit to avoid the duplication?

As a side note I am a bit worried about the performance of some parts of
this function and will add comments, but that can be addressed
afterwards

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,94 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
     return X, A
 
 
+def _get_spheres_rows(seeds, radius, allow_overlap, mask_img):
+    '''Utility function to get only the rows which are occupied by sphere at
```
jeromedockes, line 90: IIUC "rows" are boolean arrays that index a masked image, could this be stated a bit more explicitly (e.g. in the "returns" section)?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,94 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+    spheres_map: list of lists 
+        The spheres' location (flattened array), one list for each sphere.
+    '''
+    seeds = list(seeds)
+
+    # Load masking data
+    mask, affine = masking._load_mask_img(mask_img)
+
+    # Get coordinate for alle voxels inside of mask
+    brain_coords = list(np.ndindex(*mask.shape))
```
jeromedockes, line 118: this is expensive because it constructs a very large list

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,94 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+
+    # Get the closest voxels for each seed
+    nearests = []
+    for sx, sy, sz in seeds:
+        nearest = np.round(coord_transform(sx, sy, sz, 
+                           np.linalg.inv(affine)))
+        nearest = nearest.astype(int)
+        nearest = (nearest[0], nearest[1], nearest[2])
+        try:
+            nearests.append(brain_coords.index(nearest))
```
jeromedockes, line 128: looking up `nearest` in a large list several times is expensive

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,94 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+        except ValueError:
+            nearests.append(None)
+
+    brain_coords = np.asarray(list(zip(*brain_coords)))
```
jeromedockes, line 132: also costly because of iterating over `brain_coords` which is a very large list

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,94 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+                          np.in1d(brain_coords[2, :], mask_coords[2, :])]
+                        ).all(0)
+
+    # Bring brain_coords into the masks space.
+    brain_coords = coord_transform(brain_coords[0], brain_coords[1],
+                                  brain_coords[2], affine)
+    brain_coords = np.asarray(brain_coords).T
+    # Calculate the adjacent coordinates in radius.
+    clf = neighbors.NearestNeighbors(radius=radius)
+    adjacency = clf.fit(brain_coords).radius_neighbors_graph(seeds)
```
jeromedockes, line 147: that takes time but we expect it to, it is more or less the bulk of the function's work

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,94 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+    adjacency = adjacency.tolil()
+
+    # Include the voxel containing the seed itself
+    brain_coords = brain_coords.astype(int).tolist()
```
jeromedockes, line 151: converting to list is expensive

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,94 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+
+    Parameters
+    ----------
+    seeds_: List of triplets of coordinates in native space
+        Seed definitions. List of coordinates of the seeds in the same space
+        as target_affine.
+    radius: float
+        Indicates, in millimeters, the radius for the sphere around the seed.
+    allow_overlap: boolean
+        If False, an error is raised if the maps overlaps (ie at least two
```
jeromedockes, line 102: might as well write "a ValueError is raised"

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -85,6 +86,94 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
+            # Include seed itself
+            adjacency[i, brain_coords.index(np.array(seed))] = True
+        except ValueError:
+            # seed is not in the mask
+            pass
+
+    # Removing voxels not inside of the mask.
+    adjacency[:, mask_idx==False] = False
+    
+    if np.any(adjacency.sum(axis=0) >= 2):
```
jeromedockes, line 167:
```suggestion
    adjacency = adjacency.tocsr()
    if np.any(adjacency.sum(axis=0) >= 2):
```
LIL format is good for constructing the matrix but before performing linear operations on it we must convert it to a more efficient format such as CSR, see [the scipy documentation](https://docs.scipy.org/doc/scipy/reference/sparse.html#usage-information)

SRSteinkamp:
Thanks for the feedback :) 
I will need some time to go over it, but maybe I can give you some reasoning at least for the following point:

> is that much of `_get_spheres_rows` is copied from the body of `_apply_mask_and_get_affinity`, do you reckon this could be refactored a bit to avoid the duplication?

I am currently giving it a lot of thought, and it might actually be possible - I was actually going to defend my implementation, but it could be quite simple (doesn't not resolve many of the possible performance issues though, but it's something)

```python
def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
                                 mask_img=None):
    seeds = list(seeds)
    affine = niimg.affine

    # Compute world coordinates of all in-mask voxels.
    # Inserting a new check to allow for refractoring    
    if niimg is None:
        mask, affine = masking._load_mask_img(mask_img)
        # Get coordinate for alle voxels inside of mask
        mask_coords = list(np.ndindex(*mask.shape))
        X = None

    elif mask_img is not None:
        mask_img = check_niimg_3d(mask_img)
        mask_img = image.resample_img(mask_img, target_affine=affine,
                                      target_shape=niimg.shape[:3],
                                      interpolation='nearest')
        mask, _ = masking._load_mask_img(mask_img)
        mask_coords = list(zip(*np.where(mask != 0)))

        X = masking._apply_mask_fmri(niimg, mask_img)
```

This approach, however, would lead to back to the problem I'm currently struggling with. 
It would not be possible to do the `raise ValueError('Sphere around seed #%i is empty' % i)` during 
`_iter_regions_from_spheres`.   As we need all voxels in the brain for back projection (I think - see `spheres_map = image.new_img_like(mask_img, 
                                spheres.reshape([*mask.shape, n_seeds]))` in `_spheres_inversion`.)

To get to a consistend behavior, we would probaly need to check at a later point, whether there are empty spheres in the brain. 

Possibly here in `_spheres_inversion`, doing something like this (not knowing how this will be performance wise):

```python
flat_mask, _ = _masking._load_mask_img(mask_img).reshape(-1)

for i, sphere in enumerate(_iter_regions_from_spheres(seeds_, radius, 
                               allow_overlap, mask)):
    spheres[sphere, i] = 1
    spheres[flat_mask, i] = 0
    if np.all(spheres[:, i] == 0):
         raise ValueError('Sphere around seed #%i is empty' % i)
```

Still this way, we could drop the whole `_get_spheres_row` function, and could possibly resolve my previous issue.

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: Applying `.tocsr()` here breaks the function. I unfortunately have no idea about CSR format etc. but it doesn't seem to have `.rows` Attribute which is later used in the iteration. 

I am not sure if transforming back and forth will help much.

SRSteinkamp:
New Push:

I removed `_get_spheres_row` and refractored `_apply_mask_and_get_affine` to acommodate the inverse transform (I also added a doc-string to the function).

Checking for empty spheres is now performed in `_spheres_inversion`, and `.inverse_transform` should now perform consistently with `.transform`. 

Basically all my previous problems are resolved now. We might need to think about efficiency as mentioned by @jeromedockes. But, this will probably - for better or worse - directly impact `NiftiSpheresMasker.transform()`, as well.

nilearn/input_data/nifti_spheres_masker.py
jeromedockes: we can use `.getrow(i)` and iterate over row indices instead of using the LIL's `.rows` and it will be faster. otherwise, yes transforming back and forth would help. basically the lil (list of lists) format contains for each row, the list of non-zero column indices, and the list of corresponding values. adding rows or changing the sparsity structure is cheap but arithmetic operations are expensive. So we convert to CSR (compressed sparse row), which contains one array of column indices, one array of indices into the first array indicating where each row begins, and one array of values. adding one non-zero can change everything in a CSR matrix so it is not efficient for constructing a sparse matrix by adding elements as we do here, but it is very efficient for computations. this is why a very common scheme is to use a LIL to construct the matrix, then convert to CSR or CSC

jeromedockes:
> New Push:
>
> I removed `_get_spheres_row` and refractored `_apply_mask_and_get_affine` to acommodate the inverse transform (I also added a doc-string to the function).
>
> Checking for empty spheres is now performed in `_spheres_inversion`, and `.inverse_transform` should now perform consistently with `.transform`. 

great! and the diff is significantly smaller which is always a plus.

>
> Basically all my previous problems are resolved now. We might need to
> think about efficiency as mentioned by @jeromedockes. But, this will
> probably - for better or worse - directly impact
> `NiftiSpheresMasker.transform()`, as well.

yes, we can address the biggest inefficiencies but that should be easy
if we have settled on the larger refactoring

> (And all with some unwanted new lines, sigh)

What do you mean?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -138,6 +200,65 @@ def __call__(self, imgs):
...
+    else:
+        raise ValueError('Please provide mask_img at initialization to'
+                            ' provide a reference for the inverse_transform')
+
+    flat_mask, _ = masking._load_mask_img(mask_img)
+    flat_mask = flat_mask.reshape(-1) 
+
+    n_seeds = len(seeds_)
+
+    spheres = np.zeros((np.prod(mask.shape), n_seeds))
```
jeromedockes, line 238: we should use a sparse matrix instead. but do we really need to store all at once, since we are iterating over seeds and treating all separately?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -138,6 +200,65 @@ def __call__(self, imgs):
...
+
+    for i, sphere in enumerate(_iter_regions_from_spheres(seeds_, radius,
```
jeromedockes, line 240: I am wondering if we should store the adjacency matrix in the `transform`, to avoid recomputing it here?

SRSteinkamp: In my mind, there are some (maybe very special) use-cases where using `.inverse_transform` might used without having used `.transform`. 
For example for visualization purposes only. 
But more problematic:
In the `.transform` the brain image will probably be masked, so that the adjacency will have dimensions `shape = (n_seeds, np.sum(mask_img.get_data())` or `shape=(n_seeds, np.prod(niimg.shape))`. Whereas in the `.inverse_transform` we need dimensions `shape = (n_seeds, np.prod(mask.shape))`.
All cases that would have to be checked and transformed accordingly - which I think will just add some more layers of complexities, that might not be wanted.

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: I will give it a try - as I said I am not experienced with sparse matrices.

nilearn/input_data/nifti_spheres_masker.py
bthirion: No, in the `transform` methods, input images are not masked yet.

bthirion:
Let me know whenever this is ready for review.
Best,

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: I meant this as a procedure (as in are going to be masked), which affects in the end the size of the `A` or `adjacency` matrix.

jeromedockes:
> In my mind, there are some (maybe very special) use-cases where using `.inverse_transform` might used without having used `.transform`. 
> For example for visualization purposes only. 
> But more problematic:
> In the `.transform` the brain image will probably be masked, so that the adjacency will have dimensions `shape = (n_seeds, np.sum(mask_img.get_data())` or `shape=(n_seeds, np.prod(niimg.shape))`. Whereas in the `.inverse_transform` we need dimensions `shape = (n_seeds, np.prod(mask.shape))`.
> All cases that would have to be checked and transformed accordingly - which I think will just add some more layers of complexities, that might not be wanted.

Thanks for the explanation. It seems that when niimg is `None` (which is
not the case in transform) `_apply_mask_and_get_affinity` ignores the
mask (only takes its shape into account), but I wonder if that needs to
be the case.

I am surprised to see that when the mask needs to be resampled to the
image, the resampled mask does not seem to be stored anywhere.
the `inverse_transform` of the current implementation uses the
`mask_img` provided to the masker's constructor, but this is not
necessarily the one used to mask images in transform if the images have
a different shape or affine.

jeromedockes:
>
> I will give it a try - as I said I am not experienced with sparse matrices.

No problem, we can help you with that once we have settled the bigger
issue of storing the resampled mask_img_

SRSteinkamp:
Hmm, I think I slightly disagree with your assessment of the `mask_img`. 
In my opinion there are good reasons not to necessarily store the resampled `mask_img`, for example using the sphere's masker on different niimgs / datasets. 
I also don't see too much of  an issue with back-projecting to the mask_img, even if it has different dimensions. As the idea was to allow the `.inverse_transform` even if the Masker has never been used to `transform` data (although it might be nicer).

bthirion:
We can indeed avoid storing the resampled mask.  I think that  skipping that does not break the logic of the object: .fit() does not do anything, and .transform() may not have been called before inverse_transform().

jeromedockes:
> Hmm, I think I slightly disagree with your assessment of the `mask_img`. 
> In my opinion there are good reasons not to necessarily store the resampled `mask_img`, for example using the sphere's masker on different niimgs / datasets. 
> I also don't see too much of  an issue with back-projecting to the mask_img, even if it has different dimensions. As the idea was to allow the `.inverse_transform` even if the Masker has never been used to `transform` data (although it might be nicer).

ok thanks, I understand better why we don't store the resampled mask.
Besides the memory and performance issues, are there still unresolved
problems? Is the failing example you posted earlier now working as
expected?

SRSteinkamp:
I will try to introduce sparse matrices at the suggested locations (tonight).

Apart from that, I think that now everything in the function works as expected. The CI failures seem not to be based on my changes (if I read the test-results correctly).

SRSteinkamp:
I tried some profiling with sparse matrices in lines `236 - 244`, but in profiling it seems to be quite slower than using dense matrices (indexing appears to be very inefficient). Also `csr_matrix().sum()` returns a `np.matrix` so I do not see an advantage to use sparse matrices later. 

In terms of memory consumption (not at all an expert in this), but is it a big boost in terms of performance / memory efficiency if a matrix, that is going to be dense anyways, starts out as being sparse? 

@jeromedockes if you have any ideas to deal with it feel free to go.

Otherwise, I would say it's ready for a review @bthirion?

jeromedockes:
Hi @SRSteinkamp,
to make my suggestion more concrete I opened a PR on your branch. it suggests

- using the voxels inside the mask rather than all voxels
- using the affinity matrix computed by `_apply_mask_and_get_affinity`. At the
  moment we get this matrix, and then iterate over its rows to build a dense
  matrix containing the same data
- then after scaling just taking the dot product of this sparse matrix with the
  signals matrix and unmasking.

I have noticed a 6 x speedup when applying inverse transform with 5 seeds and
the mask provided by `load_mni_mask`, and 12 x with 100 seeds. memory is less
straightforward to profile but this does avoid allocating an extra n_seeds x
n_voxels dense matrix.

That being said, I think the opportunities for larger computation and memory
reduction, both for `transform` and `inverse_transform`, lie in parts of
`_apply_mask_and_get_affinity` that were already there before, so outside the
scope of this PR

In any case AFAICT the current version produces correct results so I am not
opposed to setting performance issues aside and merging as it is.

here is the script I used to measure computation time:

```
import timeit
import numpy as np
from nilearn import datasets
from nilearn.input_data import nifti_spheres_masker

mni = datasets.load_mni152_brain_mask()
seeds = [(0, -52, 18), (-46, -68, 32), (46, -68, 32), (1, 50, -5)]
# add an overlapping seed
seeds.append((46, -68, 28))
signals = np.random.randn(5, len(seeds))
masker = nifti_spheres_masker.NiftiSpheresMasker(
    seeds, radius=5, mask_img=mni, allow_overlap=True
).fit()
res = timeit.timeit(
    "masker.inverse_transform(signals)", globals=globals(), number=3
)
print("timing: {}".format(res))
```

SRSteinkamp:
Oh wow, looks great, thanks :) 
Always good to see how much I can still improve my python-game. 

I really like the PR (there's one line of code I'm sure about yet), even the empty-seed error message seems to be more informative :). 

But I would assume as well that it is better placed in a separate `Improve NiftiSpheresMasker` PR, as it dives deeper in the whole of the function.

bthirion:
> But I would assume as well that it is better placed in a separate `Improve NiftiSpheresMasker` PR, as it dives deeper in the whole of the function.

If you're fine with the PR, I suggest that you merge it: it is better to send out a finalized PR. I swear I'll review it quickly !
Best,

SRSteinkamp:
I merged jerome's PR into my branch. I don't see any issue with it so far :) 
I have to admit that I am not a 100% sure what's happening in certain lines, but I have a rough idea. 

Anyways - there is now less and faster code.

jeromedockes:
thanks! If you want to point out the most obscure lines I am happy to explain them here.

jeromedockes:
the azure failure is a failed download, not related to this PR

SRSteinkamp:
Thank you! :) 

Ah I think that I am not exactly certain what sparse.diags does, but all the sparse matrix operations are still a bit mysterious for me anyways - I don't know it feels just very non-intuitive for me. If I've got the time I will dig into it a bit one my own :) 

@bthirion , I think we are ready :)

jeromedockes:
> Ah I think that I am not exactly certain what sparse.diags does

It creates a diagonal matrix, similarly to `numpy.diag`:
```
>>> sparse.diags([1., 2.]).toarray()
[[1. 0.]
 [0. 2.]]
```
then taking the dot product by this matrix (on the right) is like
multiplying each column by the numbers in the diagonal -- in our case
the scaling by inverse number of spheres a voxel belongs to. 
(with sparse matrices we don't have the numpy broadcasting)

> but all the sparse matrix operations are still a bit mysterious for me anyways - I don't know it feels just very non-intuitive for me. If I've got the time I will dig into it a bit one my own :) 

great! the scipy documentation is very helpful

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -22,12 +23,47 @@
...
+        shape: (number of seeds, number of voxels)
+    '''
     seeds = list(seeds)
-    affine = niimg.affine
 
     # Compute world coordinates of all in-mask voxels.
-
-    if mask_img is not None:
+    if niimg is None:
+        mask, affine = masking._load_mask_img(mask_img)
```
bthirion, line 60: what if `mask_img` is None too ?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -73,11 +110,16 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
...
     for i, seed in enumerate(seeds):
         try:
-            A[i, mask_coords.index(seed)] = True
+            A[i, mask_coords.index(list(map(int, seed)))] = True
         except ValueError:
             # seed is not in the mask
             pass
 
+    sphere_sizes = np.asarray(A.tocsr().sum(axis=1)).ravel()
+    empty_spheres, *_ = np.nonzero(sphere_sizes == 0)
```
bthirion, line 119:
```suggestion
    empty_spheres, _ = np.nonzero(sphere_sizes == 0)
```

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -73,11 +110,16 @@ def _apply_mask_and_get_affinity(seeds, niimg, radius, allow_overlap,
     mask_coords = mask_coords.astype(int).tolist()
     for i, seed in enumerate(seeds):
         try:
-            A[i, mask_coords.index(seed)] = True
+            A[i, mask_coords.index(list(map(int, seed)))] = True
```
bthirion, line 113: Sorry, why do we need that ?

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -110,11 +152,32 @@ def _iter_signals_from_spheres(seeds, niimg, radius, allow_overlap,
...
+    allow_overlap: boolean
+        If False, an error is raised if the maps overlaps (ie at least two
+        maps have a non-zero value for the same voxel).
+    mask_img: Niimg-like object,
+        Mask as a reference space to project the spheres back into brain space.
+    """
+    _, adjacency = _apply_mask_and_get_affinity(seeds, None, radius,
+                                                allow_overlap,
+                                                mask_img=mask_img)
+    yield from adjacency.rows
```
bthirion, line 178: Where is this function used ?

nilearn/input_data/nifti_spheres_masker.py
jeromedockes: I don't think this change would work, `nonzero` returns a tuple with one element per dimension (so 1 here)
It could be also `empty_spheres = np.nonzero(sphere_sizes == 0)[0]`

nilearn/input_data/nifti_spheres_masker.py
jeromedockes: the `mask_coords` are converted to int above so unless the seed coords are too they will (almost) never be found in `mask_coords`

nilearn/input_data/nifti_spheres_masker.py
jeromedockes: it isn't :) forgot to remove it after some refactoring

nilearn/input_data/nifti_spheres_masker.py
jeromedockes: I don't think that happens, because this is called either by `transform` with the input image as `niimg` or by `inverse_transform` which first checks the mask and raises a `ValueError` if it is `None` -- but it is true that this function could check that case (and raise an exception when both are `None`)

jeromedockes:
@SRSteinkamp merging master will fix the circleci failure; azure pipelines failure is unrelated and I restarted it

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: I will use the `empty_spheres = np.nonzero(sphere_sizes == 0)[0]`, feels more intuitive for me.

nilearn/input_data/nifti_spheres_masker.py
SRSteinkamp: Adding an Error here. Although, this should not happen in practice

bthirion:
Merging now. Thank you  @SRSteinkamp

PR #2474, opened 2020-05-21
changing surface mesh from a list to a namedtuple

juhuntenburg:
Updates of surface handling as discussed in #2171

- Changing the representation of a surface mesh from a list to a named tuple

Additional changes towards a surface object that contains both mesh and data are to follow

### DIFF
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -478,11 +478,12 @@ def vol_to_surf(img, surf_mesh,
     img : Niimg-like object, 3d or 4d.
         See http://nilearn.github.io/manipulating_images/input_output.html
 
-    surf_mesh : str or numpy.ndarray
+    surf_mesh : str or numpy.ndarray or Mesh
         Either a file containing surface mesh geometry (valid formats
         are .gii or Freesurfer specific files such as .orig, .pial,
         .sphere, .white, .inflated) or two Numpy arrays organized in a list,
-        tuple or a namedtuple with the fields "coordinates" and "faces".
+        tuple or a namedtuple with the fields "coordinates" and "faces", or
+        a Mesh object with "coordinates" and "faces" attributes.
 
     radius : float, optional
         The size (in mm) of the neighbourhood from which samples are drawn
@@ -814,15 +815,16 @@ def load_surf_mesh(surf_mesh):
 
     Parameters
     ----------
-    surf_mesh : str or numpy.ndarray
+    surf_mesh : str or numpy.ndarray or Mesh
         Either a file containing surface mesh geometry (valid formats
         are .gii .gii.gz or Freesurfer specific files such as .orig, .pial,
         .sphere, .white, .inflated) or two Numpy arrays organized in a list,
-        tuple or a namedtuple with the fields "coordinates" and "faces"
+        tuple or a namedtuple with the fields "coordinates" and "faces", or a
+        Mesh object with "coordinates" and "faces" attributes.
 
     Returns
     --------
-    mesh : namedtuple
+    mesh : Mesh
         With the fields "coordinates" and "faces", each containing a
         numpy.ndarray
 
@@ -877,7 +879,7 @@ def load_surf_mesh(surf_mesh):
                               'array containing  the indices (into coords) of '
                               'the mesh faces. The input was a list with '
                               '%r elements.') % len(surf_mesh))
-    elif (hasattr(surf_mesh, "faces") & hasattr(surf_mesh,"coordinates")):
+    elif (hasattr(surf_mesh, "faces") and hasattr(surf_mesh,"coordinates")):
         coords, faces = surf_mesh.coordinates, surf_mesh.faces
         mesh = Mesh(coordinates=coords, faces=faces)
 
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -6,6 +6,7 @@ import warnings
 import collections
 import gzip
 from distutils.version import LooseVersion
+from collections import namedtuple
 
 
 import numpy as np
@@ -815,18 +816,19 @@ def load_surf_mesh(surf_mesh):
     surf_mesh : str or numpy.ndarray
         Either a file containing surface mesh geometry (valid formats
         are .gii .gii.gz or Freesurfer specific files such as .orig, .pial,
-        .sphere, .white, .inflated) or a list or tuple of two Numpy arrays,
-        the first containing the x-y-z coordinates of the mesh
-        vertices, the second containing the indices (into coords)
-        of the mesh faces.
+        .sphere, .white, .inflated) or two Numpy arrays organized in a list,
+        tuple or a namedtuple with the fields "coordinates" and "faces"
 
     Returns
     --------
-    [coords, faces] : List of two numpy.ndarray
-        The first containing the x-y-z coordinates of the mesh vertices,
-        the second containing the indices (into coords) of the mesh faces.
+    mesh : namedtuple
+        With the fields "coordinates" and "faces", each containing a
+        numpy.ndarray
 
     """
+    # Create a named tuple
+    Mesh = namedtuple("mesh", ["coordinates", "faces"])
+
     # if input is a filename, try to load it
     if isinstance(surf_mesh, str):
         # resolve globbing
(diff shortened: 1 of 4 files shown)


jeromedockes:
azure pipelines failure is unrelated

nilearn/surface/surface.py
```diff
@@ -685,17 +686,18 @@ def load_surf_mesh(surf_mesh):
...
     --------
-    [coords, faces] : List of two numpy.ndarray
-        The first containing the x-y-z coordinates of the mesh vertices,
-        the second containing the indices (into coords) of the mesh faces.
+    mesh : namedtuple
+        With the fields "coordinates" and "faces", each containing a
+        numpy.ndarray
     """
+    # Create a named tuple
+    Mesh = namedtuple("mesh", ["coordinates", "faces"])
```
GaelVaroquaux, line 699: I think that I would do this definition at the module level, and reuse it as much as possible.

GaelVaroquaux: Also, do we want the names to be coordinates & faces? In the document, I believe that we had said "points" and "triangles", but I am not to hung up on those names. What matters to me is that the names make as many people happy as possible.

emdupre: I'm +1 on "coordinates" and "faces" personally, but that's likely since I'm more familiar with the neuroimaging literature, which uses those terms !

jeromedockes:
> I'm +1 on "coordinates" and "faces" personally, but that's likely since I'm more familiar with the neuroimaging literature, which uses those terms !

I also find "coordinates" more explicit hence better than "points". I
guess the rationale behind "points" is that it resembles the "pointset"
in gifti images?

nilearn/surface/surface.py
juhuntenburg: I am happy with either, I used coordinates and faces because that was what we used in the previous function, but since the arguments there were positional and not actually named it also wouldnt be a big issue to change

juhuntenburg:
So, in terms of the discussion in #2171 , if we decide to default to GiftiImages, shouldn't load_surf_mesh and load_surf_data also return GiftiImages (instead of namedtuple of arrays for mesh and simple array for data)?

In that case we would need to replace the functions entirely, as they wouldn't be backwards compatible with GiftiImages as outputs

nilearn/surface/surface.py
GaelVaroquaux: OK, so I am updating the proposal to use "coordinates" and "faces"

GaelVaroquaux:
Indeed, changing the return argument would break backward compatibility.

Here is what I propose (inspired from what we have done for nifti):

* the function modified in this PR (load_surf_mesh), and its companion (load_surf_data) are there to retrieve numpy arrays. In a sens, they can be seen as the mirror of load_data

* we can introduce 2 other functions, "check_surf_mesh" and "check_surf_data", that would be used as check_niimg, and check_niimg3d, to ensure internal consistency of data handling. This could be done in a separate PR, as the need arises.

SurfaceMasker.inverse_transform always returns Gifti.

Does that sound reasonable?

bthirion:
> SurfaceMasker.inverse_transform always returns Gifti.

To be sure: you mean, a mesh and data, or data as a Gifti object ?

GaelVaroquaux:
>     SurfaceMasker.inverse_transform always returns Gifti.

> To be sure: you mean, a mesh and data, or data as a Gifti object ?

Yes indeed, my proposal is: a named-tuple with the mesh and the data as
Gifti objects.

nilearn/surface/surface.py
juhuntenburg: > I think that I would do this definition at the module level, and reuse it as much as possible.

Meaning in surface.py or in surface/\_\_init\_\_.py  ?

juhuntenburg:
I moved the definition of the Mesh namedtuple to the top of surface.py. Let me know if that is what you envisioned @GaelVaroquaux . From the remainder of the discussion I think there are no changes necessary at this point.

bthirion:
I think it's consistent with what was discussed. Probably needs some dedicated tests.

NicolasGensollen:
Hi @juhuntenburg 
I was looking at this PR and it seems pretty close. I think it is a good first step towards the goals of #2171 and it would be nice to finish and merge it. I can help if you don't have the time right now.

nilearn/surface/surface.py
```diff
@@ -685,17 +687,16 @@ def load_surf_mesh(surf_mesh):
...
-        of the mesh faces.
+        .sphere, .white, .inflated) or two Numpy arrays organized in a list,
+        tuple or a namedtuple with the fields "coordinates" and "faces"
 
     Returns
     --------
-    [coords, faces] : List of two numpy.ndarray
-        The first containing the x-y-z coordinates of the mesh vertices,
-        the second containing the indices (into coords) of the mesh faces.
+    mesh : namedtuple
```
jeromedockes, line 695: namedtuple is not a type. mesh is a `nilearn.surface.surface.Mesh` (and also a `tuple`)

nilearn/surface/surface.py
```diff
@@ -740,24 +746,28 @@ def load_surf_mesh(surf_mesh):
                               'array containing  the indices (into coords) of '
                               'the mesh faces. The input was a list with '
                               '%r elements.') % len(surf_mesh))
+    elif (hasattr(surf_mesh, "faces") & hasattr(surf_mesh,"coordinates")):
```
jeromedockes, line 749:
```suggestion
    elif (hasattr(surf_mesh, "faces") and hasattr(surf_mesh,"coordinates")):
```
`&` in python is bitwise and; logical and is `and`

jeromedockes:
maybe the docstrings of the surface plotting functions could be updated as well?

juhuntenburg:
> Hi @juhuntenburg
> I was looking at this PR and it seems pretty close. I think it is a good first step towards the goals of #2171 and it would be nice to finish and merge it. I can help if you don't have the time right now.

Thanks @NicolasGensollen ! If you have space please go for it! I won't be able to look at it before some point in January

NicolasGensollen:
I added the suggestions from @jeromedockes as well as some basic tests.
Let me know if you think this is ready to merge.

NicolasGensollen:
Pinging @alexisthual if you have some time for a review... ;)
Might also be interesting to get your opinion on nilearn's plan for better surface handling (see discussion in #2171) since this PR is basically a first step in this direction.

alexisthual (approved):
Looks good to me :)

alexisthual:
I simply wanted to stress one thing: in Three.js, [`Mesh`](https://threejs.org/docs/#api/en/objects/Mesh) is used to describe an object comprising a [`Geometry`](https://threejs.org/docs/#api/en/core/BufferGeometry) and a `Texture`. Here, what we call a `Mesh` is the equivalent of Three.js's geometry object.
I am not particularly against having our own interpretation / vocabulary, but I still think it would be nice to get inspired by other libraries.

NicolasGensollen:
@alexisthual I believe the plan is to have something like:

```python
Surface = namedtuple('surface', ['mesh', 'data'])
Mesh = namedtuple('mesh', ['coordinates', 'faces'])
```
Such that you have:

```python
mesh = surf.mesh
# Get the data
data = surf.data
# Get the vertices coordinates
coordinates = mesh.coordinates
# Get the triangles
faces = mesh.faces
```

I don't know `Three.js`, but if I understand correctly, nilearn's `Surface` would be the equivalent of their `Mesh` while nilearn's `Mesh` would be equivalent to their `Geometry` (and `data` equivalent to `Material`)?

alexisthual:
Yes, from what I read here and in #2171, I think you got it right :)
Three.js is a javascript API for using WebGL. It is massively used (66K stars on github today), so I believe it is a great idea to get some inspiration from them! I don't have a strong opinion on what is happening here though.

GaelVaroquaux:
The question is whether the terminology relates to what is used in neuroimaging (freesurfer, mostly, I guess).

This is what our users will have in mind.

alexisthual:
Of course! This I am not able to judge though :)

nilearn/surface/surface.py
```diff
@@ -894,10 +907,9 @@ def _check_mesh(mesh):
 def check_mesh_and_data(mesh, data):
```
bthirion, line 907: We never check that the faces are consistent with coordinates, e.g. faces.max() <= len(coords)

nilearn/surface/tests/test_surface.py
```diff
@@ -144,6 +145,16 @@ def test_load_surf_data_file_error():
...
+def test_load_surf_mesh():
+    coords, faces = generate_surf()
+    mesh = Mesh(coords, faces)
+    assert_array_equal(mesh.coordinates, coords)
+    assert_array_equal(mesh.faces, faces)
+    loaded_mesh = load_surf_mesh(mesh)
+    assert isinstance(loaded_mesh, Mesh)
+    assert_array_equal(mesh.coordinates, loaded_mesh.coordinates)
+    assert_array_equal(mesh.faces, loaded_mesh.faces)
+
```
bthirion, line 157: ... and the corresponding test...

bthirion (commented):
Thx ! Still one comment about mesh consistency check.

nilearn/surface/surface.py
NicolasGensollen: Indeed! This should be addressed in 088f2a4

nilearn/surface/surface.py
```diff
@@ -859,15 +867,20 @@ def load_surf_mesh(surf_mesh):
                               'array containing  the indices (into coords) of '
                               'the mesh faces. The input was a list with '
                               '%r elements.') % len(surf_mesh))
+    elif (hasattr(surf_mesh, "faces") and hasattr(surf_mesh,"coordinates")):
```
bthirion, line 870:
```suggestion
    elif (hasattr(surf_mesh, "faces") and hasattr(surf_mesh, "coordinates")):
```

nilearn/surface/surface.py
```diff
@@ -894,10 +907,19 @@ def _check_mesh(mesh):
 def check_mesh_and_data(mesh, data):
     """Load surface mesh and data, check that they have compatible shapes."""
     mesh = load_surf_mesh(mesh)
-    nodes, faces = mesh
     data = load_surf_data(data)
-    if len(data) != len(nodes):
+    # Check that mesh coordinates has a number of nodes
+    # equals to the size of the data.
```
bthirion, line 912:
```suggestion
    # equal to the size of the data.
```

bthirion (commented):
Thx, we're close. Simply minor syntactic issues / typos.

nilearn/surface/surface.py
```diff
@@ -26,6 +27,9 @@ class EfficiencyWarning(UserWarning):
 from .. import _utils
 from nilearn.image import get_data
 
+# Create a namedtuple object for meshes
+Mesh = namedtuple("mesh", ["coordinates", "faces"])
```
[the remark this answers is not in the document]
NicolasGensollen: I believe this has been done. You should be able to import the Mesh like this:

```python
from nilearn.surface import Mesh
```

bthirion (approved):
LGTM, thx !

NicolasGensollen:
I rebased and solved the conflicts introduced by #2649 
I believe this is ready to be merged! :tada: 
Thanks @juhuntenburg for the PR and everybody for the reviews! :+1:

PR #2478, opened 2020-05-21
DOC: Switch to napoleon

GaelVaroquaux:
To have better rendering of attributes

Napoleon is a docstring-formatter maintained by sphinx, which is quite robust.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -48,7 +48,8 @@ sym_links:
 	-cd ../examples/01_plotting/ && ln -sf ../nilearn_cache
 	-cd ../examples/02_decoding/ && ln -sf ../nilearn_cache
 	-cd ../examples/03_connectivity/ && ln -sf ../nilearn_cache
-	-cd ../examples/04_glm/ && ln -sf ../nilearn_cache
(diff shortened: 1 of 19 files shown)


bthirion:
Will merge tomorrow unless I hear otherwise.

GaelVaroquaux:
Yes. This is finally right. It was a bit of a struggle.

I couldn't use napoleon (it emits too many warnings, and I couldn't fix them), but I still was able to fix some rendering problems of the docstrings.

However, some of the rendering problems have been introduced by the recent releases of numpydoc. I added CSS to work around them. A cleaner fix would require us to rewrite all our docstrings to follow the evolution of the standard :(.

bthirion:
> A cleaner fix would require us to rewrite all our docstrings to follow the evolution of the standard :(.

Which we'll have to do eventually... Thx ! 
I'm OK for merging. No afterthought ?

GaelVaroquaux:
> I'm OK for merging. No afterthought ?

Nope, go ahead!

bthirion:
Thx, merged !

Issue #2485, opened 2020-05-22
Thresholded colorbar doesn't look good on a glass brain plot with a black background


kfinc:
Thresholded colorbar doesn't look good on a glass brain plots with a black background because of threshold overlay set to black when `black_bg=True` (`brain_color = (0. if black_bg else 1.,) * 3)`. To keep consistency with `plot_stat_map`, I suggest using gray color overlay for white and black backgrounds.

Here how it look right now:
![image](https://user-images.githubusercontent.com/14907241/82711846-dc12ed00-9c86-11ea-9fe8-d5755ea9bd98.png)

And after setting overlay to gray (`brain_color = (0.5, 0.5, 0.5)`):

![image](https://user-images.githubusercontent.com/14907241/82712228-dcf84e80-9c87-11ea-8c11-3b23b0f7603b.png)
![image](https://user-images.githubusercontent.com/14907241/82712291-00bb9480-9c88-11ea-9afa-c4ca33c8c182.png)

Let me know what you think about the suggestion.

bthirion:
Indeed !
Please issue a PR...
Thx,
Bertrand

GaelVaroquaux:
Fixed by  #2486 thanks!

PR #2486, opened 2020-05-23
Fixed colorbar aesthetics on a black backgroud

kfinc:
This PR deals with the aesthetics of colorbar on black backgrounds (issue #2485). For glass brain, plotting overlay for thresholded colorbar is set up to default gray color in the same way as for `plot_stat_map`. Additionally, this PR changes the outline of the colorbar for a black background as it should be changed to white in the same way as color of `tick_params`. Note, that this will also affect colorbar on other plotting functions when `black_bg=True`.

Plots after fix:
image
image


GaelVaroquaux:
This looks great! Thanks a lot! Merging!

PR #2501, opened 2020-05-28
complete cache key

jeromedockes:
small change suggested in discussion with circleci support. I will probably change a few details later but avoid non-essential changes while discussion with support is ongoing


emdupre:
Thanks for working on this, @jeromedockes !!

Issue #2313, opened 2020-02-22
Add white-matter surfaces (lh.smoothwm and lh.white) to fetch_surf_fsaverage 


braincharter:
Hi!
It would be nice to add the white-matter surfaces from freesurfer (call lh.smoothwm and lh.white) to fetch_surf_fsaverage as well. 

I'm interested in investigating the vessels located on the gm/wm border on a group of subjects, as well as fMRI maps on that location. As for now I am using my own surface computed from one subject, but adding those to fetch_surf_fsaverage would really simplify the process (and make it more consistent across subjects).

jeromedockes:
It could also be useful to select points between the pial and wm
surfaces for the volume -> surface projections

bthirion:
I'm +0
on the negative side, data download is our first cause of ci breakage (but not particularly from nitrc), and I guess that only few users would require both surfaces.
Maybe @jeromedockes has one: being able to perform 'layer-dependent' projection (or mapping).
Any other opinion ?

braincharter:
I think that exactly for layer-dependent projection a lot of users would gain interest in the surf features!
Or it could be another function completely (fetch_surf_fsaverage_wm).

By the way, how were those surfaces computed? By running recon-all on ICBM_MNI_2009?

bthirion:
No, I guess that it's a freesurfer-made average of individual mesh, but I'm not an exepert in freesurfer.

jeromedockes:
that was done in #2517

PR #2517, opened 2020-06-04
[MRG] add vol to surf by sampling between nested surfaces

jeromedockes:
This PR introduces a new method for volume to surface projections: the user
provides two nested surfaces with one-to-one correspondence between nodes, and
values are sampled between each pair of nodes. IIRC @bthirion asked for this a
couple of times. Still WIP but I would appreciate feedback on the suggested
interface (add an `inner_mesh` optional parameter), and to know where I can
download white matter surface meshes for the fsaverage5 and fsaverage surfaces
to add fetchers for these.

~~Also, to be able to test surface-related functions on more realistic data, this
PR adds a couple of utils in `data_gen`: one to generate random low-resolution
images in mni space, and one to extract a (bad) mesh from the MNI mask shipped
with nilearn at a chosen resolution. This enables getting low-resolution
(possibly nested) meshes for testing without needing to add more data to the
package or to require network access, but while still having something that
resembles more a brain surface in size, position and topology than random points
and random triangles. (a few examples below)~~

example_plots


bthirion:
My first thought is not to go into the business of mesh generation. I guess that fsaverage is used 90% of the time and for other cases, people will reuse the freesurfer generated stuff.

nilearn/surface/surface.py
```diff
@@ -442,6 +475,16 @@ def vol_to_surf(img, surf_mesh,
...
+    inner_mesh : str or numpy.ndarray, optional (default=None)
+        Either a file containing a surface mesh or a pair of ndarrays
+        (coordinates, triangles). If provided this is an inner surface that is
+        nested inside the one represented by `surf_mesh` -- e.g. `surf_mesh` is
+        a pial surface and `inner_mesh` a white matter surface. In this case
+        nodes in both meshes must correspond: node i in `surf_mesh` is just
+        across the gray matter thickness from node i in `inner_mesh`. Image
+        values for index i are then sampled along the line joining these two
+        points.
+
```
bthirion, line 487: My understanding is that quite often, people want to select the depth at which they're sampling (layered fMRI). This only makes sense for high resolution data of course. 
So I would define the sampling scheme along the cortical depths by a list of values between 0 and 1: [0]=white, [1]=pial, [.5]=middle surface, [0, 0.1,..,1]= average across 10 intermediate surfaces.

jeromedockes:
> My first thought is not to go into the business of mesh generation. I guess that fsaverage is used 90% of the time and for other cases, people will reuse the freesurfer generated stuff.

note the mesh generation is intended for internal unit tests only, I'm
not suggesting to expose it to users. We can use fsaverage5 for all
tests as it is shipped with nilearn and hence doesn't need downloading,
but it might be a bit large and slow down some tests unnecessarily, and
it may not be ideal to always test on one single mesh. also that would
require also including the fsaverage white matter with the code -- but
we may want to do that anyway.

bthirion:
Also, there is a glitch that remains to be fixed. I could not find the issue, but it was something like a half voxel shift that needs to be addressed for freesurfer meshes.  IIRC this resulted in LR inaccuracies typically. Do you remember that one ?

bthirion:
> also that would require also including the fsaverage white matter with the code -- but we may want to do that anyway.

Yes that's my opinion. It's pretty small.

nilearn/surface/surface.py
jeromedockes: ok! should that be an extra parameter then? or we can deprecate the "ball" sampling strategy and apply this depth selection either to the vertex normal or to the between-mesh segment, depending on whether a second surface is provided or not. In the single mesh /fixed depth case users may also want to select the offsets of sample points. for now they can change the scale but the points are always centered on the vertex

jeromedockes:
related: #2313

jeromedockes:
> Also, there is a glitch that remains to be fixed. I could not find the issue, but it was something like a half voxel shift that needs to be addressed for freesurfer meshes. IIRC this resulted in LR inaccuracies typically. Do you remember that one ?

the neurostars question about reading surface files?

jeromedockes:
> Also, there is a glitch that remains to be fixed. I could not find the issue, but it was something like a half voxel shift that needs to be addressed for freesurfer meshes. IIRC this resulted in LR inaccuracies typically. Do you remember that one ?

do you mean that one: #2215

bthirion:
> also that would require also including the fsaverage white matter with the code -- but we may want to do that anyway.

Yes that's my opinion. It's pretty small.

bthirion:
> do you mean that one: #2215

Yes, exactly

nilearn/surface/surface.py
bthirion: I think that the ball think is a very good idea if you have only one surface. Ideally, I would like to make it possible to use both heuristics. It is also better for backward compatibility.
The problem is how to have both in the same API, and making it clear !

jeromedockes:

>> do you mean that one: #2215
>
> Yes, exactly

Indeed it hasn't been addressed. IIUC it's an issue with nibabel or
load_surf_mesh though, not related to vol_to_surf

jeromedockes:
> The problem is how to have both in the same API, and making it clear !

yes I agree. I seems that the list of depths would compete with the
`n_points` and `radius` parameters ... not sure what is the best way to
handle that, the new parameter would cause these to be ignored?

bthirion:
If a list of depthsis provided, `n_points` and `radius` would be ignored.

emdupre:
> where I can download white matter surface meshes for the fsaverage5 and fsaverage surfaces
to add fetchers for these.

These are included in the [main Freesurfer archive](https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/6.0.1/freesurfer-Linux-centos6_x86_64-stable-pub-v6.0.1.tar.gz) in the standard FS format (e.g. `rh.white`). They're also hosted on [OSF for the templateflow project](https://osf.io/ue5gx/) in the `fsaverage` folder at the 164k (fsaverage) and 10k (fsaverage5) resolutions.

@oesteban can confirm, but I'd imagine we could download directly from OSF !

jeromedockes:
> These are included in the [main Freesurfer archive](https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/6.0.1/freesurfer-Linux-centos6_x86_64-stable-pub-v6.0.1.tar.gz) in the standard FS format (e.g. `rh.white`). They're also hosted on [OSF for the templateflow project](https://osf.io/ue5gx/) in the `fsaverage` folder at the 164k (fsaverage) and 10k (fsaverage5) resolutions.
>
thanks! I saw this a bit late :/ but I did get them from the freesurfer
archive as you had suggested IRL. I converted them to gifti using
freesurfer's mris_convert tool
re. hosting: other fsaverage5 surfaces are shipped with nilearn so I
guess these should be as well, as suggested by @bthirion's comment.
for fsaverage, rather than getting it from a separate place, I was
thinking of adding these two surfaces to the archive containing the rest
currently used by the fetcher, which is on nitrc. we can later also add
a copy on osf in case download from nitrc fails.
however I was planning to wait for 
https://github.com/nilearn/nilearn/pull/2514
to be merged before updating the fetcher

jeromedockes:
I added the white matter surfaces in https://www.nitrc.org/frs/download.php/11807/fsaverage.tar.gz

bthirion:
Thx !

oesteban:
> @oesteban can confirm, but I'd imagine we could download directly from OSF !

Yes, or even better from S3. That would also buy you a couple of nice things:

- Maintenance through templateflow
- GIFTI format.

jeromedockes:
> Yes, or even better from S3. That would also buy you a couple of nice things:
>
> - Maintenance through templateflow
> - GIFTI format.

Thanks. for the moment it seemed easier to convert to gifti and upload
to the collection on nitrc that already contains the other fsaverage
meshes downloaded by the nilearn fetcher, but this is good to keep in
mind and we could add osf or S3 as a fallback solution when nitrc server
is down (or vice-versa)

oesteban:
I would love to see a better integration of templateflow with nilearn. If there's interest in that, I could work out a PR.

I'm aware though, that templateflow might be a bit too experimental for nilearn. Happy to hear other thoughts.

bthirion:
> I'm aware though, that templateflow might be a bit too experimental for nilearn. Happy to hear other thoughts.

You read in our thoughts --- at least mine.  It looks like a rather heavy dependency...

jeromedockes:
@bthirion I added fetchers for fsaverage5 and fsaverage white matter and removed the utility that generated low-resolution testing meshes and use fsaverage instead. I think your other comments had been addressed so if CI passes this will be ready for another review

bthirion (approved):
I don't find any issue. Thx !

bthirion:
Any other review ? @juhuntenburg ?

juhuntenburg:
Yep I can look at it within the next days if that helps

juhuntenburg:
I am running into an issue when running the new tests, which is probably not related to this PR, but maybe someone can help?
```
E       fixture 'tmp_path_factory' not found
>       available fixtures: cache, capfd, capsys, doctest_namespace, monkeypatch, pytestconfig, record_xml_property, recwarn, request_mocker, temp_nilearn_data_dir, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

jeromedockes:
I think you need to update pytest

juhuntenburg:
First thing I notice, the plotting of the nested surfaces doesn't seem to really work, or at least it isn't very intuitive / helpful. Maybe the meshes could be rotated, or the inner one plotted with some transparency so that it becomes clearer?
Figure_1
Figure_4

nilearn/_utils/data_gen.py
```diff
@@ -8,6 +8,8 @@
 import numpy as np
 import pandas as pd
 import scipy.signal
+from scipy.spatial import Delaunay
```
juhuntenburg, line 11: Are you using thing anywhere?

jeromedockes:
Julia Huntenburg <notifications@github.com> writes:

> First thing I notice, the plotting of the nested surfaces doesn't seem to really work, or at least it isn't very intuitive / helpful. Maybe the meshes could be rotated, or the inner one plotted with some transparency so that it becomes clearer?
> Figure_1
> Figure_4

I changed the camera position and made the outer one transparent. it is
still not super easy to see because matplotlib gets the overlays wrong
but I think a bit better than before. to really improve the plot the
solution would be to replace matplotlib with plotly in this example,
which we can do using nilearn's `view_surf` and `view_markers` -- but
this would best be addressed in a different issue and PR

jeromedockes:
>  import numpy as np
>  import pandas as pd
>  import scipy.signal
> +from scipy.spatial import Delaunay
>
> Are you using thing anywhere?

thanks, this was left after I removed some code -- I removed the imports now

jeromedockes:
circleci failure is unrelated and fixed in another PR

nilearn/surface/surface.py
```diff
@@ -424,6 +482,8 @@ def vol_to_surf(img, surf_mesh,
 
     kind : {'line', 'ball'}
         The strategy used to sample image intensities around each vertex.
+        If `inner_mesh` is provided this parameter is ignored and values are
+        sampled along a line
```
juhuntenburg, line 486: Why not make a third kind 'depth' or similar? I find it not obvious otherwise that there is a third way to sample, which is invoked when inner_mesh is provided

juhuntenburg (commented):
Looks good to me, the new way of plotting the surfaces helps clarity! I made some small comments. Most importantly, is there a reason why you chose not to indicate this new sampling method as a third 'kind' value? For me that would make it more obvious that there are three ways of sampling. The way it is implemented now (sampling between surfaces is chosen if inner surface is given) is less explicit. 

Also, I think it would be really great to have a real example for this, i.e. with real brain data, sampling at different depths of the cortex. This is a nice feature and would be worth showcasing. But maybe that is for another PR :)

examples/01_plotting/plot_surface_projection_strategies.py
```diff
@@ -11,10 +10,14 @@
 vertex. Image values are interpolated at each sample location, then these
 samples are averaged to produce a value for the vertex.
 
-Two strategies are available to choose sample locations: they can be spread
-along the normal to the mesh, or inside a ball around the vertex. Don't worry
-too much about choosing one or the other: they take a similar amount of time
-and give almost identical results for most images.
+Three strategies are available to choose sample locations: they can be spread
```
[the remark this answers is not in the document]
jeromedockes: thanks, I updated it with your suggestion

examples/01_plotting/plot_surface_projection_strategies.py
```diff
@@ -58,12 +65,33 @@
...
     ax.scatter(*sample_points.T, color='r')
 
+######################################################################
+# Adjust the sample locations
+######################################################################
+# For "line" and nested surfaces, the depth parameter allows ajusting the
+# position of samples along the line
+
+nested_sample_points = surface._sample_locations_between_surfaces(
+    mesh, inner_mesh, np.eye(4), depth=[-.5, 0., .8, 1.])
```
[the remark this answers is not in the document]
jeromedockes: yes, to show that it is possible to go outside the provided surfaces (ie negative values are allowed)

nilearn/surface/surface.py
jeromedockes: I added the 'depth' strategy, and made the default 'auto', which chooses 'depth' if 'inner_mesh' is provided and 'line' otherwise

examples/01_plotting/plot_surface_projection_strategies.py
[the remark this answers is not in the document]
jeromedockes: or values greater than 1 as well, I added 1.2

juhuntenburg (approved):
All concerns addressed for me. I would just want to check if others agree with referring to the new strategy as "depth" (@bthirion ?). I prefer this because I think it will be intuitive for many people. But I can also see an argument not to get into anatomical definitions. Although I think as long as we use depth and not anything related to layers it is ok

bthirion:
+1 for 'depth'
Merging.
Huge thanks to @jeromedockes

PR #2518, opened 2020-06-04
use pip install --upgrade in install.sh

jeromedockes:
otherwise e.g. numpy which has an old version pre-installed on travis virtualenv
doesn't get upgraded

### DIFF
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -57,8 +57,8 @@ echo_requirements_string() {
 
 create_new_travisci_env() {
     REQUIREMENTS=$(echo_requirements_string)
-    pip install $PIP_FLAGS ${REQUIREMENTS}
-    pip install pytest pytest-cov
+    pip install --upgrade $PIP_FLAGS ${REQUIREMENTS}
+    pip install --upgrade pytest pytest-cov
 
     if [[ "$INSTALL_MKL" == "true" ]]; then
         # Make sure that MKL is used


kchawla-pi:
Isn't the idea to have specific versions? Won't --upgrade break that?
Also, When we don't specify any version, doesn't the latest version get installed anyway?

jeromedockes:
Kshitij Chawla <notifications@github.com> writes:

> Isn't the idea to have specific versions? Won't --upgrade break that?

when we explicitly ask for a specific version we still get it; see for example
https://travis-ci.com/github/nilearn/nilearn/jobs/343892235#L363

> Also, When we don't specify any version, doesn't the latest version get installed anyway?

not if another version is already installed, see e.g.
https://travis-ci.com/github/nilearn/nilearn/jobs/341709675#L327
latest numpy version should be  1.18 but this build is using 1.15

another option would be to create a fresh virtualenv to make sure these
libraries aren't installed when we start

kchawla-pi:
> another option would be to create a fresh virtualenv to make sure these aren't installed when we start

I thought that's what happens everytime but clearly not. It reuses the VM from other builds  in the same job?

kchawla-pi:
Also, wow look at the circleci runtime! 1hr!! SpaceNet example completed in 5min instead of 15! What did you all do? :-)

jeromedockes:
> I thought that's what happens everytime but clearly not. It reuses the VM from other builds  in the same job?

I was also surprised to see this version installed, I can look into it a
bit more. we never install 1.15 AFAIK

GaelVaroquaux:
LGTM. Merging! Thank you!

G

PR #2523, opened 2020-06-14
Fix nilearn under scikit-learn 0.24

GaelVaroquaux:
Do not import from modules that have been made private and moved

Currently, nilearn does not run on sklearn 0.24. This PR fixes it.

### DIFF
--- a/nilearn/_utils/param_validation.py
+++ b/nilearn/_utils/param_validation.py
@@ -191,4 +191,4 @@ def check_feature_screening(screening_percentile, mask_img,
         screening_percentile_ = _adjust_screening_percentile(
             screening_percentile, mask_img, verbose=verbose)
 
-        return SelectPercentile(f_test, int(screening_percentile_))
+        return SelectPercentile(f_test, percentile=int(screening_percentile_))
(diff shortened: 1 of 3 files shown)


jeromedockes:
one travis failure is unrelated and would be fixed by #2514 
[this one](https://api.travis-ci.com/v3/job/348940066/log.txt) seems genuine?
I am also wondering if we should have a ci job running with the master branches of the dependencies (I am thinking a scheduled one with master, rather than one triggered by pull requests) we have one using release candidates ("pre-release checks"), but I am not sure if all dependencies make release candidates and how long before the releases, and I am not sure it makes sense to trigger it for PRs

kchawla-pi:
This was outlined in #1728 . If you wish, you can reopen that discussion.

jeromedockes:
> This was outlined in #1728 . If you wish, you can reopen that discussion.

yes I think we can. With respect to #2224 which closed #1728 I am
wondering if we should:

- test with master branches or nightly builds besides or instead of alpha, beta, rc releases
- make this a scheduled job rather than one triggered by PRs

WDYT?

kchawla-pi:
I'll opine in #1728 to not clutter this PR.

GaelVaroquaux:
I've fixed the genuine failure. This PR is now green.

bthirion:
Will be merged within 24hrs unless I hear otherwise.

GaelVaroquaux:
Based on the feedback above, I am merging this.

PR #2529, opened 2020-06-27
[FIX] Ensure nilearn.signal.clean will not modify original array

### LINKED ISSUE #2281: nilearn.signal.clean modifies input when "session" parameter used.
Hi guys, 

[docs](https://nilearn.github.io/modules/generated/nilearn.signal.clean.html) for nilearn.signal.clean say: 

"signals: numpy.ndarray
Timeseries. Must have shape (instant number, features number). **This array is not modified**."

However, I find that if the session parameter is used the input array **is** modified. 

Code to reproduce:
```
# make random data
np.random.seed(1)
nsamp = 1000
data_source= np.random.rand(nsamp)
confound= np.random.rand(nsamp)
data_observed_orig = data_source + confound
data_observed_input = data_observed_orig.copy()
print(np.corrcoef(data_observed_orig,confound)[0,1]**2) # theoretically 0.5
sessions_desc = np.ones(nsamp)
sessions_desc[0:nsamp//2]=0

data_cleaned_concat = nil.signal.clean(data_observed_input[:,np.newaxis],
                                       confounds=confound[:,np.newaxis],sessions=sessions_desc)
#should be True
np.all(data_observed_input == data_observed_orig)
```
output:
0.5228211029311932
False

Only happens if the session parameter is provided. In fact, the input data seems to change to something like the output, but not exactly. 

nil version 0.6.0

### LINKED ISSUE #2475: clean function modifies the input time series object if performing pass-filtering without detrending
Hi,

I am using version 0.6.0. I've found that function **clean** within the module signal modifies the input time series object if one performs a pass-fitering when _detrend_ option is False. 

For example, if you run the following code (test time series borrowed from [here](https://gist.github.com/miykael/4bb887029cba3926902feb2e4afe190f)) and then plot both the original and clean time series:

```
import numpy as np
import matplotlib.pylab as plt
from nilearn.signal import clean

original_sx = np.array([np.sin(np.linspace(0, 100, 100) * 1.5),
                        np.sin(np.linspace(0, 100, 100) * 3.),
                        np.sin(np.linspace(0, 100, 100) / 8.),
              ]).T

# Clean with detrending and high pass filtering
clean_sx = clean(original_sx, 
                   standardize=False, 
                   t_r=1.0, 
                   detrend=True, 
                   high_pass=0.2)

# Plot both time series
fig, axs = plt.subplots(ncols=2, figsize=(10, 5))  
axs[0].plot(original_sx)
axs[0].set_title("Original Time Series")
axs[1].plot(clean_sx)
axs[1].set_title("Clean Time Series")
plt.show()
```

You should see how "green" signal is totally filtered out (its frequency is ~ 0.02):

example_detrend_filt

However, if you now try to run the same code but without detrending the time series:

```
import numpy as np
import matplotlib.pylab as plt
from nilearn.signal import clean

original_sx = np.array([np.sin(np.linspace(0, 100, 100) * 1.5),
                        np.sin(np.linspace(0, 100, 100) * 3.),
                        np.sin(np.linspace(0, 100, 100) / 8.),
              ]).T

# Clean with only high pass filtering (Note that now detrend is set to False)
clean_sx = clean(original_sx, 
                   standardize=False, 
                   t_r=1.0, 
                   detrend=False, 
                   high_pass=0.2)

# Plot both time series
fig, axs = plt.subplots(ncols=2, figsize=(10, 5))  
axs[0].plot(original_sx)
axs[0].set_title("Original Time Series")
axs[1].plot(clean_sx)
axs[1].set_title("Clean Time Series")
plt.show()
```

The original time series object is modified as well

example_only_filt

When __detrend__ option is set to True, it first calls the function **_standardize**, which creates a copy of the original time series, and then carries on to do the filtering by calling the  function **butterworth**. However, when __detrend__ is False, which skips the first step, since **butterworth** is alway called with its  _copy_ argument set to False (the default value), the original time series are modified too. 

Is this the expected behaviour? I would expect the input time series to remain always the same, as inline modifications can lead to unwanted results if the original time series need to be reused again at some point.

Cheers

tsalo:
Closes #2281 and closes #2475.

- Copy the input array at the top of `nilearn.signal.clean` so that the original array will not be modified.
- Add tests to cover the cases reported in #2281 and #2475.
- Add more tests for the same issue interspersed throughout the testing functions for `nilearn.signal.clean`.

### DIFF
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -482,13 +482,13 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
                          "but you provided ensure_finite={0}"
                          .format(ensure_finite))
 
+    signals = signals.copy()
     if not isinstance(signals, np.ndarray):
         signals = as_ndarray(signals)
 
     if ensure_finite:
         mask = np.logical_not(np.isfinite(signals))
         if mask.any():
-            signals = signals.copy()
             signals[mask] = 0
 
     # Read confounds
--- a/nilearn/tests/test_signal.py
+++ b/nilearn/tests/test_signal.py
@@ -275,14 +275,19 @@ def test_clean_detrending():
     trends = generate_trends(n_features=n_features,
                              length=n_samples)
     x = signals + trends
+    x_orig = x.copy()
 
     # if NANs, data out should be False with ensure_finite=True
     y = signals + trends
     y[20, 150] = np.nan
     y[5, 500] = np.nan
     y[15, 14] = np.inf
-    y = nisignal.clean(y, ensure_finite=True)
-    assert np.any(np.isfinite(y)), True
+    y_orig = y.copy()
+
+    y_clean = nisignal.clean(y, ensure_finite=True)
+    assert np.any(np.isfinite(y_clean)), True
+    # clean should not modify inputs
+    assert np.array_equal(y_orig, y)
 
     # test boolean is not given to signal.clean
     pytest.raises(TypeError, nisignal.clean, x, low_pass=False)
@@ -292,11 +297,15 @@ def test_clean_detrending():
     x_detrended = nisignal.clean(x, standardize=False, detrend=True,
                                  low_pass=None, high_pass=None)
     np.testing.assert_almost_equal(x_detrended, signals, decimal=13)
+    # clean should not modify inputs
+    assert np.array_equal(x_orig, x)
 
     # This should do nothing
     x_undetrended = nisignal.clean(x, standardize=False, detrend=False,
                                    low_pass=None, high_pass=None)
     assert not abs(x_undetrended - signals).max() < 0.06
+    # clean should not modify inputs
+    assert np.array_equal(x_orig, x)
 
 
 def test_clean_t_r():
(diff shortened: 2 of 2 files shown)


tsalo:
`nilearn.image.tests.test_resampling.test_resampling_nan()` is failing because of the following:

> Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) was emitted. The list of emitted warnings is: [DeprecationWarning('Passing `np.nan` to mean no clipping in np.clip has always been unreliable, and is now deprecated. In future, this will always return nan, like it already does when min or max are arrays that contain nan. To skip a bound, pass either None or an np.inf of an appropriate sign.',)].

I didn't touch anything related to resampling, so I'm not sure what the problem is. Also, the test passes locally.

emdupre:
Thanks, @tsalo ! I think the failure was fixed in https://github.com/nilearn/nilearn/pull/2523 if you want to pull from master.

tsalo:
@bthirion Thanks!

@emdupre I think my branch is up to date with master though. I updated right before I branched yesterday. It also looks like #2514 has the same error and updated from master five days ago.

emdupre:
It was only merged yesterday, so I'd expect #2514 to still have it :smile_cat: But yes, you're right, you're up to date ! I know there was at least one failure on that branch with that error, so maybe it's sporadic.

Otherwise, this looks good to me, thanks !!

jeromedockes:
I think it was not related to #2523 but due to the latest numpy release, should be fixed in #2530

tsalo:
@jeromedockes I updated after #2530 was merged and it looks like the problem was fixed. The docs build is now breaking due to some issue with dataset fetching, but it appears to be unrelated, so thank you!

jeromedockes:
> @jeromedockes I updated after #2530 was merged and it looks like the problem was fixed. The docs build is now breaking due to some issue with dataset fetching, but it appears to be unrelated, so thank you!

No problem! indeed the new one was a failing download, I restarted it
and it passed

GaelVaroquaux:
It seems that the tests have actually passed. The changes are good. I am merging.

PR #2532, opened 2020-07-01
WIP: Add higher order AR noise models to run_glm

rob-luke:
Hi nilearn team,

I am interested in using higher order auto regressive noise models with run_glm. Today I started playing around with this, but realised I should have raised an issue first, so... Is there any plans to include AR(N>1) models with `run_glm`? And would you be open to adding this functionality?

Below I include my rough code from today. It's not pretty or optimised, but I thought it would be better to include code if I had it. If you are interested in adding AR(N) models then I can build on what I have in this PR.

These higher order models are used in near infrared spectroscopy analysis. For an example see [figure 1c here](https://www.spiedigitallibrary.org/journals/Neurophotonics/volume-3/issue-1/010401/Commentary-on-the-statistical-properties-of-noise-and-its-implication/10.1117/1.NPh.3.1.010401.full#f1).

Also I have taken the yule walker code from MNE (the lines seem to have last been edited by @larsoner and @agramfort ). I quickly checked nilearn and MNE licenses are both BSD, but is this ok to do?


larsoner:
Yes you should be fine license-wise

nilearn/glm/first_level/first_level.py
```diff
@@ -122,25 +141,39 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
-    if noise_model == 'ar1':
+    if 'ar' in noise_model:
+        ar_order = int(re.split('ar', noise_model)[1])
         # compute and discretize the AR1 coefs
-        ar1 = (
-            (ols_result.residuals[1:]
-             * ols_result.residuals[:-1]).sum(axis=0)
-            / (ols_result.residuals ** 2).sum(axis=0)
-        )
+        ar1 = [_yule_walker(ols_result.residuals[:, res].reshape(-1, 1).T,
```
bthirion, line 147: then we should no longer call that guy `ar1` but simply `ar` or `ar_coef_`

nilearn/glm/first_level/first_level.py
```diff
@@ -122,25 +141,39 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+                            order=ar_order)
+               for res in range(ols_result.residuals.shape[1])]
+        if len(ar1[0]) == 1:
+            ar1 = np.asarray(ar1).ravel()
         del ols_result
-        ar1 = (ar1 * bins).astype(np.int) * 1. / bins
-        # Fit the AR model acccording to current AR(1) estimates
+        for idx in range(len(ar1)):
+            ar1[idx] = (ar1[idx] * bins).astype(np.int) * 1. / bins
```
bthirion, line 154: I'm not sure what this does.
For AR(1) it was simple to cluster voxels according to AR coef, but how should we do here ?

rob-luke:
> Thx for the proposition. You did it well, I'm quite happy with it.

Great thanks. I will continue then.

> Of course, there are lots of details to be discussed. 

Yes I agree. I have lots of questions. How about I review my own code with the questions I have too.

> An important one to me is eprformance.

I may need your help with this one. Lets get the API and results correct first with some tests, then I may need your suggestions on performance.

nilearn/glm/first_level/first_level.py
```diff
@@ -105,8 +124,8 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
         values are RegressionResults instances corresponding to the voxels.
 
     """
-    acceptable_noise_models = ['ar1', 'ols']
-    if noise_model not in acceptable_noise_models:
+    acceptable_noise_models = ['ols']
+    if not (('ar' in noise_model) or (noise_model in acceptable_noise_models)):
```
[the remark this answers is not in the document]
rob-luke: Also this will pass for weird names like 'arturtle'. I will improve.

nilearn/glm/first_level/first_level.py
rob-luke: Yeah I wasn't sure what the clustering was for, I assumed it was something MRI specific that I didn't know about. Can you explain what the clustering is for then we could come up with the best AR(N) approach. We could cluster by first coef. We could not cluster. We could cluster by residual.

nilearn/glm/first_level/first_level.py
```diff
@@ -122,25 +141,39 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+                                                 for lab in labels]])
             for val in vals)
-        for val, result in zip(vals, ar_result):
-            results[val] = result
+
+        if type(ar1[0]) is np.float64:
+            for val, result in zip(vals, ar_result):
+                results[val] = result
+        else:
+            labels = ['_'.join([str(v) for v in val]) for val in vals]
```
rob-luke, line 174: I wasn't sure what the labels should be for AR(N) models. So here I am just joining the values with an underscore. But if I understand better what the clustering is used (comment above) then maybe this will answer itself.

nilearn/glm/first_level/first_level.py
```diff
@@ -71,6 +72,24 @@ def _ar_model_fit(X, val, Y):
     return ARModel(X, val).fit(Y)
 
 
+def _yule_walker(X, order=1):
+    """Compute Yule-Walker (adapted from MNE and statsmodels).
+    Operates in-place.
```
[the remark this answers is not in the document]
rob-luke: Agreed, tests required all around. I will start that now that I know you are open to the PR.

nilearn/glm/first_level/first_level.py
```diff
@@ -71,6 +72,24 @@ def _ar_model_fit(X, val, Y):
...
+    """
+    from scipy import linalg
```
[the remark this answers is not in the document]
rob-luke: Understood

nilearn/glm/first_level/first_level.py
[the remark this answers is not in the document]
rob-luke: I am also considering some model order optimisation approaches, so would naming like `noise_model='arbic"` or `noise_model='ar_aic"` be acceptable? This is thinking slightly ahead, but might save rewriting this check twice.

nilearn/glm/first_level/first_level.py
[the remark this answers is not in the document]
rob-luke: Agreed. I will add tests all around now that I know you are open to this PR.

nilearn/glm/first_level/first_level.py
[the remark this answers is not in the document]
rob-luke: Understood

nilearn/glm/first_level/first_level.py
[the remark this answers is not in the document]
rob-luke: For some reason my old comments have disappeared. 

This would currently pass for `arturtle` so I need to rework this.

nilearn/glm/first_level/first_level.py
rob-luke: I don't understand what the clustering and labelling is for, I assumed it was something MRI specific. Could you please explain the purpose of the clustering, then maybe I can suggest an appropriate approach for AR(N) models. We could cluster by first coef, or by residual, or by sum of coef, or not cluster.

nilearn/glm/first_level/first_level.py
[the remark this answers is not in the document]
bthirion: Please do not consider model selection in this PR.
First, it's not reliable (short time series, unfulfilled assumptions etc.), second, it's expensive, and third, it maybe misinterpreted by users. I think it's fine that the use commits to a certain model order.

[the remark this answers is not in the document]
rob-luke: Great. Lets stick to the original plan then. Thanks for the feedback.

nilearn/glm/first_level/first_level.py
bthirion: Clustering is used to reduce computation: the regression fit, that includes whitening, is done on groups of voxels insead of isolated voxels: looping of a few tens of groups is much more efficient than looping over voxels. Bzw, we need to adapt the whitening scheme implemented in  `model.py` and `regression.py` modules.

rob-luke: Great. Thank you for explaining that, it makes sense to me now. I will also take a look at `model.p` and `regression.py`.

So it seems we want to cluster based on something that will be uniformly distributed then. Could we use the last term in the AR fit then? I assume this term is the most likely to be noise, whereas I have observed the first fit term is often common across channels.

And apologies for differences we might have in terminology, I am not an MRI researcher.

bthirion: > So it seems we want to cluster based on something that will be uniformly distributed then. Could we use the last term in the AR fit then? I assume this term is the most likely to be noise, whereas I have observed the first fit term is often similar across channels.

Not sure I understand your suggestion: you want voxels in a given cluster to be as homogeneous as possible in terms of noise characteristics. So you would typically use all AR coefficients, but the most informative ones are the first and second coefficients.

rob-luke: Sorry I previously misunderstood.

> So you would typically use all AR coefficients

Ok this is what the code is now doing. But please let me know if I am still mistaken.

In regards to your original comment about what the binning now does, I have added the following to the docs to inform the user that the binning is now per AR coefficient. Does this make sense? Or do you suggest we cluster differently?

```
    bins : int, optional
        Maximum number of discrete bins for the AR(1) coef histogram.
        If higher order AR models are specified this specifies the maximum
        number of bins per AR coefficient.
```

rob-luke:
> Bzw, we need to adapt the whitening scheme implemented in model.py and regression.py modules.

I just took a look over the whitening procedure (https://github.com/nilearn/nilearn/blob/master/nilearn/glm/regression.py#L263) and it appears to already account for higher order AR models. 

Sorry to bother you further @bthirion, but could you please point me to the code section that will need to be adapted.

rob-luke:
I also cant figure out why the python 3.6 travis test is failing. The same tests seem to pass on my PC.

jeromedockes:
> edit: Im using python 3.7, so I guess that's why Im not seeing the error locally.

the problem comes from numpy version 1.11, rather than from the python version. 

we can look into it and make it work with numpy 1.11 but scipy, pandas, nibabel and scikit-learn now require numpy 1.13, so I think we can do the same.

bthirion:
In the current code, the AR(k) model is used to separate the voxels into clusters with similar AR coefficients ---beyond AR(1). Yet the model fit is then done using and AR(1) model, see l.169 in
https://github.com/nilearn/nilearn/blob/fb01af20316537f5b058744e428accc50cfbf64c/nilearn/glm/first_level/first_level.py

rob-luke:
I apologise for the delay in getting back to this PR.

Thank you @bthirion for your previous review of the code. Unfortunately I can't yet see where my misunderstanding is. I will try and explain how I am interpreting the code, hopefully this helps to highlight my error so I can ensure the model fit is actually done with the AR(N) model.

As I read the code, line 169 is called with `val` equal to a single float if I use 'ar1', and an array of length 3 if I use 'ar3'

https://github.com/nilearn/nilearn/blob/d847472a86f212bb47fd4e68d8b97eebedb19b9c/nilearn/glm/first_level/first_level.py#L169-L171

I can confirm this is the case as these values are then used to instantiate the ARModel and the `order` is set based on this.

https://github.com/nilearn/nilearn/blob/72d51b27d5faa97122b20abd7aae1b155f45bbc7/nilearn/glm/regression.py#L269

As such, I have added a test to ensure that the models returned do indeed have the correct order assigned. 

https://github.com/nilearn/nilearn/blob/d847472a86f212bb47fd4e68d8b97eebedb19b9c/nilearn/glm/tests/test_first_level.py#L249

And then this is used in the ARModel to whiten the data

https://github.com/nilearn/nilearn/blob/72d51b27d5faa97122b20abd7aae1b155f45bbc7/nilearn/glm/regression.py#L222

And the whitening code does appear to iterate over all values of rho that were passed in
https://github.com/nilearn/nilearn/blob/72d51b27d5faa97122b20abd7aae1b155f45bbc7/nilearn/glm/regression.py#L287-L291

Sorry to bother you with this again, I appreciate any clue that might point me in the right direction on how to actually ensure the AR(N) model is used (or how its not, so I can fix it 😄 ).

bthirion:
Sorry, I probably misread the code and actually forgot that the internal regression classes are indeed powerful enough to support such models.

The main issue at the moment is that the code takes ages to run... I am  running it on the 
'plot_spm_multimodal_faces.py' example. It should not take more than one minute on a laptop.

rob-luke:
Don't apologise, I appreciate the considered feedback.

> The main issue at the moment is that the code takes ages to run... I am running it on the
'plot_spm_multimodal_faces.py' example. It should not take more than one minute on a laptop.

Totally agreed, now we are happy that the implementation is roughly correct I will proceed to add tests and then optimise. I will ping you again once I think its ready for another review. Thanks!

rob-luke:
Thanks for the pointer to the `plot_spm_multimodal_faces.py` example, this helped me to determine where the slow downs were. On my PC the example takes the same time to run with this PR and master. So @bthirion could you please review this PR again?

It may be less accurate (but its more convenient) to compare the execution times for the example using the CI builds. Master takes 1:54 to build (see https://nilearn.github.io/auto_examples/04_glm_first_level/sg_execution_times.html) and this PR takes 1:41 to build (see https://6579-1235740-gh.circle-artifacts.com/0/doc/_build/html/auto_examples/04_glm_first_level/sg_execution_times.html). Of course this could have been done on different cloud platforms with different hardware, but I think it indicates that the PR is now running comparably fast as master.

I have also exposed the `bins` argument in `FirstLevelModel.fit()`. I did this because I assumed that if you have AR(N), this will cause more variability in your coefficient histogram, which will slow computation down, so you may wish to reduce the number of bins if you increase the noise model. For example something like...
```python
fmri_glm = FirstLevelModel(noise_model='ar3')
fmri_glm = fmri_glm.fit(fmri_img, design_matrices=design_matrices, bins=10)
```
But I am not an fMRI scientist, so I can revert this change if you wish.

bthirion:
> fmri_glm = FirstLevelModel(noise_model='ar3')

When I do that the code starts to take ages on my box...
Is the number of actual bins equal to bins=10 or to bins ** order ?
In which case we need to revise the strategy ...

rob-luke:
You are correct that the number of actual bins used was `bins ** order`. 

I just pushed an alternative approach. Here I estimate the number of bins required per order based on the total number of bins requested by the user. Basically...

```python
        if ar_order > 1:
            # If AR(N>1) is requested the number of bins is applied
            # per coefficient, so we compute the bins per coefficient to match
            # the number of total bins requested by the user.
            bins = np.floor(bins ** (1. / ar_order))
```

On my PC the `fmri_glm = FirstLevelModel(noise_model='ar3')` example now runs in about a minute.

@bthirion, is this strategy better? Or do you have something else in mind?

nilearn/glm/first_level/first_level.py
```diff
@@ -72,6 +73,26 @@ def _ar_model_fit(X, val, Y):
...
 
 
+def _yule_walker(X, order):
+    """Compute Yule-Walker (adapted from MNE and statsmodels).
+    Operates in-place.
+    """
+    from scipy import linalg
+    assert X.ndim == 2
+    assert order > 0, "AR order must be positive"
+    assert isinstance(order, int), "AR order must be an integer"
```
NicolasGensollen, line 83: We should test the function `_yule_walker` specifically. Although, I would replace the assertions with `ValueError` and `TypeError` exceptions and test that these are raised with bad arguments.

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +145,45 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
 
-    if noise_model == 'ar1':
-        # compute and discretize the AR1 coefs
-        ar1 = (
-            (ols_result.residuals[1:]
-             * ols_result.residuals[:-1]).sum(axis=0)
-            / (ols_result.residuals ** 2).sum(axis=0)
-        )
+    if 'ar' in noise_model:
+        ar_order = int(re.split('ar', noise_model)[1])
```
NicolasGensollen, line 149: I would add a little bit more checking here. For example, if the (distracted) user provides `arfoo` or `arr3` this would fail with a non informative error.

NicolasGensollen (changes requested):
Thanks for the PR! This looks good! :+1: 
I haven't tested the code extensively, but I left a few comments related to testing and errors handling. 
I also think we should modify or add an example to demonstrate this.

bthirion:
> @bthirion, is this strategy better? Or do you have something else in mind?

Thx ! Indeed, it now runs on my box...
Indeed, there would be a better strategy imho: running a clustering on the vector of AR coefficients directly: it would lead to a much more efficient and accurate representation of the AR model.

bthirion:
Another question I have is whether the benefit of the approach can be demonstrated on a given dataset, e.g. as a gain in sensitivity ? It does not seem obvious to me when looking at the 'plot_spm_multimodal_faces' example ?

rob-luke:
Thank you both for the feedback. I will give the clustering approach a try and let you know how I go.

Regarding the example. This I don't know how to do. I am not an mri researcher and the examples are quite advanced relative to my understanding. I use nilearn to analyse functional near infrared spectroscopy data ( https://mne.tools/mne-nirs/ ). Sure, I could modify an nilearn example to run an AR(N) model, but my commentary won't be insightful, as I don't really understand most of the existing text. But if you point me to which example you want me to modify I can do it. Sorry I'm not more help here.

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +147,54 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
-        ar1 = (
-            (ols_result.residuals[1:]
-             * ols_result.residuals[:-1]).sum(axis=0)
-            / (ols_result.residuals ** 2).sum(axis=0)
-        )
+    if 'ar' in noise_model:
+
+        err_msg = ("AR order must be a positive integer specified as arN, "
+                   "where N is an integer. E.g. ar3. "
+                   f"You provied {noise_model}.")
```
NicolasGensollen, line 154: I think you're getting a lot of errors with Python 3.5 tests because of this fstring. 
Note: We'll drop Python 3.5 support in release 0.8, such that using fstrings will be fine then!

rob-luke: Thanks! Old habits 😄 Should be fixed now.

nilearn/glm/first_level/first_level.py
rob-luke: Thanks for the suggestion. I add the errors and tests.

nilearn/glm/first_level/first_level.py
rob-luke: I tried to catch a few more edge cases. Please let me know if you think of others I missed.

rob-luke:
> running a clustering on the vector of AR coefficients directly

Thanks for the suggestion @bthirion. I have made an attempt at this. Could you please review the approach and see if its along the lines of what you thought.

The current clustering implementations is slightly slower than the previous approach. Using the  `plot_spm_multimodal_faces.py` example with ar3 model. Commit ff770266 (simple binning) ran in 1 minute on my PC. Whereas the clustering approach in  957c05a takes 1.5 min. For AR1 there is no difference in speed. If you are happy with the general approach we could look for optimisations.

bthirion:
I find the approach pretty good, I think this is more or less the way I would implement it (I haven't checked all details yet).
One think that can be easily improved is to have slightly less redundant/costly tests.
Next, we need to be able to motivate/showcase this new features. Among all datasets thata re demoed in nilearn, isn't there any one where say and AR(2) or AR(3) model displays some advanatge over AR(1) ? See eg 
https://nilearn.github.io/auto_examples/04_glm_first_level/plot_first_level_details.html#sphx-glr-auto-examples-04-glm-first-level-plot-first-level-details-py

rob-luke:
Great, I will optimise and add an example.

rob-luke:
Ok the `test_glm_AR_estimates` test time has been reduced from 1.03s to 0.18s and the `test_glm_AR_estimates` times reduced from 0.98s to 0.40s (according to the azure python3.7 tests)

I have added an AR(N) example to the document you suggested. I also removed the display of the design matrix in this subsection, as it seemed irrelevant to the discussion about noise models, but tell me if you want me to put it back. The updated webpage can be viewed at https://6635-1235740-gh.circle-artifacts.com/0/doc/_build/html/auto_examples/04_glm_first_level/plot_first_level_details.html#the-noise-model-ar-1-ols-or-higher-order-ar

NicolasGensollen (commented):
Hi @rob-luke 
This is looking good! Thanks for adding an example in the doc! :+1: 
I left a few small comments/questions.
Also, we should add an entry to `whats_new.rst`.

nilearn/glm/tests/test_first_level.py
```diff
@@ -228,19 +230,70 @@ def test_run_glm():
...
+            for lag in range(ar_order):
+                Y[idx] += ar_vals[lag] * Y[idx - 1 - lag]
+
+        labels, results = run_glm(Y, X, ar_arg, bins=100)
+        assert len(labels) == n
+        for lab in results.keys():
+            ar_estimate = lab.split("_")
+            for lag in range(ar_order):
+                assert_almost_equal(float(ar_estimate[lag]),
+                                    ar_vals[lag], decimal=1)
```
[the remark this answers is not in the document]
rob-luke: > might be a stupid question

Nope, good question.

The accuracy is dependent on the length of the signal you pass in. For the 500 sample we use in this test we get 1 decimal of accuracy. Roughly I have observed that if you have 10x more samples you get 1 more decimal place of accuracy (if you set n=5000 then you can increase decimal=2). However, that's just rough testing on my pc and is probably dependent on the noise properties etc.

I think a faster test is sufficient here as is testing to 1 decimal, particularly as we test out to 4 lags. However, if there is an alternative method to estimate AR coefficients that should be as efficient I'd be happy to try it.

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +148,63 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
-        )
+    if 'ar' in noise_model:
+
+        err_msg = ('AR order must be a positive integer specified as arN, '
+                   'where N is an integer. E.g. ar3. '
+                   'You provied {0}.'.format(noise_model))
+        if 'ar' not in noise_model[0:2]:
+            raise ValueError(err_msg)
+        try:
+            ar_order = int(re.split('ar', noise_model)[1])
```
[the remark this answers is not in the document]
rob-luke: Nice :)

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +148,63 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
-            (ols_result.residuals[1:]
-             * ols_result.residuals[:-1]).sum(axis=0)
-            / (ols_result.residuals ** 2).sum(axis=0)
-        )
+    if 'ar' in noise_model:
+
+        err_msg = ('AR order must be a positive integer specified as arN, '
+                   'where N is an integer. E.g. ar3. '
+                   'You provied {0}.'.format(noise_model))
+        if 'ar' not in noise_model[0:2]:
```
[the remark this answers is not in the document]
rob-luke: Thanks. I also added tests to catch edge cases where people pass in `noise_model = "ar"` or `noise_model = "3ar"`

rob-luke:
Thanks for the input @bthirion and @NicolasGensollen. I have changed the title to MRG as I am now quite happy with the pull request and think it should be close to complete. Are there any other people I should ping to get a review?

nilearn/glm/first_level/first_level.py
```diff
@@ -72,6 +73,28 @@ def _ar_model_fit(X, val, Y):
...
+    """
+    from scipy import linalg
+    if order < 1:
+        raise ValueError("AR order must be positive")
+    if type(order) is not int:
+        raise TypeError("AR order must be an integer")
+
+    denom = X.shape[-1] - np.arange(order + 1)
+    r = np.zeros(order + 1, np.float64)
+    X = X.ravel()
```
agramfort, line 88: to me you should not have to ravel. It's dangerous. You should enforce the ndim is correct ie here ndim=1 on call

nilearn/glm/first_level/first_level.py
```diff
@@ -72,6 +73,28 @@ def _ar_model_fit(X, val, Y):
...
+    X -= X.mean()
```
agramfort, line 89: are you sure we want to have inplace operations here. It seems inplicit from the function call

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +153,61 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+                   'You provied {0}.'.format(noise_model))
+        try:
+            ar_order = int(noise_model[2:])
+        except ValueError:
+            raise ValueError(err_msg)
+
+        # compute the AR coeficents
+        ar_coef_ = [_yule_walker(ols_result.residuals[:, res].reshape(-1, 1).T,
+                                 ar_order)
+                    for res in range(ols_result.residuals.shape[1])]
```
agramfort, line 169: did you consider vectorizing this loope?  computing the cross-correlations in yule walkers should be done easily for all voxels without a for loop. Also np.linalg.solve can operate in parallel to solve all linear systems for all voxels. See:


```
In [1]: import numpy as np

In [2]: A = np.random.randn(100, 2, 2)

In [3]: b = np.random.randn(100, 2)

In [4]: x = np.linalg.solve(A, b)

In [5]: x.shape
Out[5]: (100, 2)
```

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +153,61 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+        ar_coef_ = np.array(ar_coef_)
         del ols_result
-        ar1 = (ar1 * bins).astype(np.int) * 1. / bins
-        # Fit the AR model acccording to current AR(1) estimates
+        if len(ar_coef_[0]) == 1:
+            ar_coef_ = ar_coef_.ravel()
+
+        # Either bin the AR1 coefs or cluster ARN coefs
+        if type(ar_coef_[0]) is np.float64:
```
agramfort, line 176: what else could it be? I am not fan of the type(...) is ... pattern

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +153,61 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
-        # Parallelize by creating a job per ARModel
-        vals = np.unique(ar1)
+
+        # Fit the AR model according to current AR(N) estimates
         ar_result = Parallel(n_jobs=n_jobs, verbose=verbose)(
-            delayed(_ar_model_fit)(X, val, Y[:, labels == val])
-            for val in vals)
-        for val, result in zip(vals, ar_result):
+            delayed(_ar_model_fit)(X,
+                                   ar_coef_[np.where(labels == val)][0],
```
agramfort, line 204:
```suggestion
                                   ar_coef_[labels == val][0],
```
?

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +153,61 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+        if type(ar_coef_[0]) is np.float64:
+            for idx in range(len(ar_coef_)):
+                ar_coef_[idx] = (ar_coef_[idx] * bins).astype(int) \
+                                * 1. / bins
+            labels = np.array([np.str(val) for val in ar_coef_])
+        else:  # AR(N) case
+            n_clusters = np.min([bins, Y.shape[1]])
+            kmeans = KMeans(n_clusters=n_clusters).fit(ar_coef_)
+            ar_coef_ = np.array([kmeans.cluster_centers_[i]
+                                 for i in kmeans.labels_])
```
agramfort, line 185:
```suggestion
            ar_coef_ = kmeans.cluster_centers_[kmeans.labels_]
```
?

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +153,61 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+
+            # Create a set of rounded values for the labels
+            cluster_labels = kmeans.cluster_centers_.copy()
+            for idx in range(len(cluster_labels)):
+                cluster_labels[idx] = (cluster_labels[idx] * 100).\
+                                          astype(int) * 1. / 100
+            cluster_labels = np.array([np.str('_'.join([str(v) for v in val]))
+                 for val in cluster_labels])
```
agramfort, line 193: what are you trying to do here?

nilearn/glm/first_level/first_level.py
```diff
@@ -361,6 +425,10 @@ def fit(self, run_imgs, events=None, confounds=None,
             Design matrices that will be used to fit the GLM. If given it
             takes precedence over events and confounds.
 
+        bins : int, optional
+            Maximum number of discrete bins for the AR coef histogram.
+            Default=100.
```
agramfort, line 430: this is not really an histogram for kmeans. It would call this adaptive quantification.

nilearn/glm/tests/test_first_level.py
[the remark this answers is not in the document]
NicolasGensollen: Thanks for the detailed explanation! :+1:

nilearn/glm/first_level/first_level.py
rob-luke: That makes sense. Thanks

nilearn/glm/first_level/first_level.py
rob-luke: If its ar(N>3) its an array. But I have replaced this with `if ar_order is 1:` which I think is clearer to the reader. Thanks

nilearn/glm/first_level/first_level.py
rob-luke: Im not 100% sure I understood your comment correctly. Did the change I make fix this?

nilearn/glm/first_level/first_level.py
rob-luke: Updated, thanks.

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +155,60 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+                    for res in range(ols_result.residuals.shape[1])]
+        ar_coef_ = np.array(ar_coef_)
         del ols_result
-        ar1 = (ar1 * bins).astype(np.int) * 1. / bins
-        # Fit the AR model acccording to current AR(1) estimates
+        if len(ar_coef_[0]) == 1:
+            ar_coef_ = ar_coef_.ravel()
+
+        # Either bin the AR1 coefs or cluster ARN coefs
+        if ar_order is 1:
```
agramfort, line 178:
```suggestion
        if ar_order == 1:
```

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +155,60 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+        # compute the AR coeficents
+        ar_coef_ = [_yule_walker(ols_result.residuals[:, res],
+                                 ar_order)
+                    for res in range(ols_result.residuals.shape[1])]
+        ar_coef_ = np.array(ar_coef_)
         del ols_result
-        ar1 = (ar1 * bins).astype(np.int) * 1. / bins
-        # Fit the AR model acccording to current AR(1) estimates
+        if len(ar_coef_[0]) == 1:
+            ar_coef_ = ar_coef_.ravel()
```
agramfort, line 175:
```suggestion
            ar_coef_ = ar_coef_[:, 0]
```
less dangerous and more explicit

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +155,60 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+            ar_coef_ = kmeans.cluster_centers_[kmeans.labels_]
+
+            # Create a set of rounded values for the labels
+            cluster_labels = kmeans.cluster_centers_.copy()
+            for idx in range(len(cluster_labels)):
+                cluster_labels[idx] = (cluster_labels[idx] * 100).\
+                                          astype(int) * 1. / 100
+
+            cluster_labels = np.array([np.str('_'.join([str(v) for v in val]))
+                 for val in cluster_labels])
```
agramfort, line 195: can you use soemthing simpler using np.round ? eg.

```
In [13]: order = 3

In [14]: a = np.random.randn(order)

In [15]: '_'.join(map(str, np.round(a, 2)))
Out[15]: '1.12_-1.17_-1.49'
```

nilearn/glm/first_level/first_level.py
agramfort: not here.

agramfort (commented):
besides you did not yet try to vectorize the YW computation

nilearn/glm/first_level/first_level.py
rob-luke: This mess was created when I was trying to optimise the `run_glm()` run time and reduce reused code. I think we can do better

This line is run when AR(N>1) is requested. It converted the label from an array to a string so that the same code could be used in the parallel loop for both the AR(1) and AR(N>1) cases.

I have pushed an alternative approach where I have a specific `ar_model_fit` code for AR(1) vs AR(N). Do you prefer this? Its a bit DRY, but maybe easier to understand for people reading the code.

nilearn/glm/first_level/first_level.py
rob-luke: Thanks, and I did a search for other ravels and didnt find any

nilearn/glm/first_level/first_level.py
rob-luke: OOhhh that is nicer. I think I prefer this approach to the current commit version with too similar code. Let me know what you prefer.

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +155,67 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+            # Create a set of rounded values for the labels
+            cluster_labels = kmeans.cluster_centers_.copy()
+            for idx in range(len(cluster_labels)):
+                cluster_labels[idx] = (cluster_labels[idx] * 100).\
+                                          astype(int) * 1. / 100
+
+            # Create labels and coef per voxel
+            labels = np.array([cluster_labels[i] for i in kmeans.labels_])
+
+        unique_labels = np.unique(labels, axis=0)
```
NicolasGensollen, line 197: This line is failing with minimum dependencies (numpy 1.11) because of the axis argument.

nilearn/glm/first_level/first_level.py
rob-luke: Im reverting to your .join suggestion below.

nilearn/glm/first_level/first_level.py
rob-luke: Thanks, I've reverted to previous approach.

rob-luke:
Its all hands on deck tonight! 
Thanks @agramfort and @NicolasGensollen, an informative review as always. I think I address most comments except the YW, I will look in to this in the coming days.

nilearn/glm/first_level/first_level.py
```diff
@@ -361,6 +426,13 @@ def fit(self, run_imgs, events=None, confounds=None,
             Design matrices that will be used to fit the GLM. If given it
             takes precedence over events and confounds.
 
+        bins : int, optional
+            Maximum number of discrete bins for the AR coef histogram.
+            If an auto regressive model with order greater than one is specified
```
agramfort, line 431:
```suggestion
            If an autoregressive model with order greater than one is specified
```

nilearn/glm/first_level/first_level.py
```diff
@@ -83,11 +107,18 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+        The temporal variance model.
+        To specify the order of an auto regressive model place the
+        order after the characters `ar`, for example to specify a third order
+        model use `ar3`.
+        Default='ar1'.
 
     bins : int, optional
-        Maximum number of discrete bins for the AR(1) coef histogram.
+        Maximum number of discrete bins for the AR coef histogram.
+        If an auto regressive model with order greater than one is specified
```
agramfort, line 119:
```suggestion
        If an autoregressive model with order greater than one is specified
```

nilearn/glm/first_level/first_level.py
```diff
@@ -83,11 +107,18 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
     X : array of shape (n_time_points, n_regressors)
         The design matrix.
 
-    noise_model : {'ar1', 'ols'}, optional
-        The temporal variance model. Default='ar1'.
+    noise_model : {'ar(N)', 'ols'}, optional
+        The temporal variance model.
+        To specify the order of an auto regressive model place the
```
agramfort, line 112:
```suggestion
        To specify the order of an autoregressive model place the
```

agramfort (approved):
nothing more to add on my side but I am expert in GLMs with AR so I am not voicing my opinion on the K-Means approach I had never seen

rob-luke:
Thanks @agramfort, I would still like to try the YW approach you suggested in the coming days, so please dont MRG yet (I changed title back to WIP).

Originally I wrote a histogram approach rather than K-means, but this modification was suggested by @bthirion and as I am not an MRI researcher I defer to his expertise. For my small number of channels I simply set the number of bins = number of channels, but I understand that this isnt feasible for MRI with 10s of 1000s of voxels. If this isnt a standard approach is it worth validating and documenting in a small journal article?

I will ping everyone once I've tackled the YW approach and ready for another review/merge.

agramfort:
I am just worried as ARs coefficients are not really Euclidian vectors. You
can add
2 ARs coefficients from stable filters and get something which is not
stable.
I might not be a problem in practice though

>

bthirion:
But I don't see a better way to group set of AR coefficients than clustering the sequence of value (hence with an L2 norm). The point of clustering is only to decide about a similarity measure.

agramfort:
what comes to mind to to cluster based on the DSP of the filter, find the
voxels
within each cluster. Compute the autocorrelations on all the time series
from these
voxels and solve YW on this. But It might be an overkill...

bthirion:
Yes, this sounds a bit heavy to simply cluster voxels with a similar autocorrelated noise structure.

agramfort:
I agree it seems overly complex... i guess it's now more a matter to
evaluate if it works reasonably well
on real data unless you have some prior experience?

bthirion:
> on real data unless you have some prior experience?

Not really. Actually, simply demoing the added value of such an estimator is a real challenge...

agramfort:
well maybe it needs some tiny validation to demo the added value even minor?
at least that it does not make the pipeline more fragile by adding some
variance?

>

nilearn/glm/first_level/first_level.py
rob-luke: @agramfort I'm embarrassed to admit I have failed to vectorise this loop. Are there any further hints you can provide?

rob-luke:
@NicolasGensollen would you have any tips on how I can vectorise the yule walker loop? I've tried a few times and cant figure it out. Im at a bit of a loss and worried this PR will become too hard to merge the older it gets. Any suggestions appreciated.

NicolasGensollen:
@rob-luke I had a look today at the yule walker loop vectorization. 

Here is what I did so far (the following would go in the body of the function `run_glm` where you compute the coeffs). Although it's not pretty, I've put everything here such that you can copy-paste and test easily. We can re-write this in a cleaner way once we have a stable solution. 

```python
# inside run_glm...
...
# compute the AR coefficients
# imports to be placed somewhere else...
from functools import reduce
from scipy import linalg
 
denom = ols_result.residuals.shape[0] - np.arange(ar_order + 1)
y = ols_result.residuals - np.mean(ols_result.residuals, axis=0)
 
def _get_r_vector(x, i, k):
    """Helper function.
    To be placed somewhere else...
    """
    return reduce(lambda x,y: np.hstack((x,y)),
                   [np.dot(x[:, i], x[:, i])] +
                   [np.dot(x[0:-k, i], x[k:, i])
                    for k in range(1, k+1)])
 
r_vectors = [_get_r_vector(y, i, ar_order) /
              (y.shape[-1] * denom)
              for i in range(y.shape[-1])]
 
R = np.array([linalg.toeplitz(r[:-1]) for r in r_vectors])
r = np.array([_[1:] for _ in r_vectors])

# This is now vectorized
# ar_coef_ should match the ones you were computing before
ar_coef_ = np.linalg.solve(R, r)
...
```

This is building the proper inputs such that the solve step can be vectorized as suggested by @agramfort . 
At this point I'm still not able to vectorize the inputs build steps in an efficient way (although this is less important). If someone has some Numpy black magic to improve this, I'd be happy to learn about it!

There is clearly room for improvement, but I thought this could help the PR move forward.

larsoner:
@NicolasGensollen looks like a good start, I'll look a bit more -- I think the only thing that can't easily be vectorized is the `toeplitz` because it doesn't broadcast, but I think in theory the dot products should be because those can broadcast properly with `@`.

larsoner:
https://github.com/rob-luke/nilearn/pull/1/files

bthirion:
Thx @larsoner !

rob-luke:
Thank you all! I will merge that code and then take some time to study it.

> Actually, simply demoing the added value of such an estimator is a real challenge

> maybe it needs some tiny validation to demo the added value even minor

These seem the final challenge for me to overcome. As I’m not from the mri field I’m unsure how to validate. Could you please point me to an existing relevant tutorial I could update, and perhaps an appropriate metric I should report? Thanks

nilearn/glm/first_level/first_level.py
rob-luke: Tested directly in 81b6d21b8d7e8f6d761a8dbecb15ab1e35a2fe62

NicolasGensollen:
Thanks @larsoner, great solution ! :+1:

NicolasGensollen (approved):
Thanks @rob-luke ! The code looks good to me.
I'm not sure what we can do to demo the added value or show that it doesn't add variance to the pipeline. Maybe @bthirion would have a better idea on this.

nilearn/glm/first_level/first_level.py
```diff
@@ -124,26 +160,54 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
...
+
+        # compute the AR coefficients
+        ar_coef_ = _yule_walker(ols_result.residuals.T, ar_order)
         del ols_result
-        ar1 = (ar1 * bins).astype(np.int) * 1. / bins
-        # Fit the AR model acccording to current AR(1) estimates
+        if len(ar_coef_[0]) == 1:
+            ar_coef_ = ar_coef_[:, 0]
+
+        # Either bin the AR1 coefs or cluster ARN coefs
```
bthirion, line 179: I'm wondering whether we should not simply use clustering for all AR orders ?

nilearn/glm/tests/test_first_level.py
```diff
@@ -298,19 +300,81 @@ def test_run_glm():
     assert results[0.0].theta.shape == (q, n)
     assert_almost_equal(results[0.0].theta.mean(), 0, 1)
     assert_almost_equal(results[0.0].theta.var(), 1. / p, 1)
+    assert type(results[labels[0]].model) == OLSModel
 
     # ar(1) case
-    labels, results = run_glm(Y, X, 'ar1')
+    labels, results = run_glm(Y, X, "ar1")
```
bthirion, line 306: single quote is actually better...

bthirion (commented):
Thx. I'm still amazed that this feature could be added with so little code addition ;-)
I'll take another look at the example. Demoing this type of features is important I think.

rob-luke:
> Thx. I'm still amazed that this feature could be added with so little code addition ;-)

The first version wasn't so small! But I try not to steam roll other peoples repo's on the first PR :)

> I'll take another look at the example. Demoing this type of features is important I think.

Absolutely. Definitely worth getting this PR right. If you think of an appropriate demo of the feature then Ill be pleased to implement it.

> I'm wondering whether we should not simply use clustering for all AR orders ?

As you wish. Originally I just left the ar1 case as is to cause minimal disturbance. But happy to change to your preference. Just let me know if you think we should make the jump.

rob-luke:
Hi @bthirion,

Sorry to bother you about this again. But have you had a chance to think of how we might further demo this feature? Currently we just demonstrate how to use this new option in `examples/04_glm_first_level/plot_first_level_details.py`, but we don't show any benefit to using it. 

Ideally we can find a nice tidy example to demonstrate this new feature, but if nothing is coming to our minds is there another way we can proceed? As I have no insight in to the fMRI world, an alternative option may be to point users to a relevant journal article in the nilearn docs (for example [this fNIRS paper](https://www.spiedigitallibrary.org/journals/Neurophotonics/volume-3/issue-1/010401/Commentary-on-the-statistical-properties-of-noise-and-its-implication/10.1117/1.NPh.3.1.010401.full#f1)). Alternatively, I could do a quick scan of the fMRI literature to find a similar paper if you think that would help?

bthirion (approved):
LGTM. The example is sufficient. Thx !

rob-luke:
Thanks @bthirion, and I will be around, so if you think of further modifications then just ping me and I will open a follow up PR.

Thank you all for the help with this modification. I look forward to contributing more in the future.

NicolasGensollen:
Thanks a lot @rob-luke ! 
Merging... :tada:

rob-luke:
Woohoo!! 🥳 🥳 🥳

PR #2553, opened 2020-08-06
[FIX] Design matrices are wrong if multiple events in the same regressor have the same onset or offset

mwegrzyn:
Dear nilearn team,

using ```glm.first_level.make_first_level_design_matrix()``` I noticed that in some cases the baseline of a regressor can shift and never return to zero, instead being stuck at another value (e.g. -1, +1 etc.). I think the issue is that in the function ```hemodynamic_models._sample_condition()``` each onset (e.g. a stick function of +1) has to be balanced by an offset of the same, but negative value (e.g. -1). However, if multiple events start at the same time (+1) but end at different times (-1 + -1 = -2), the baseline will permanently shift downwards. The opposite effect occurs when different onsets have the same offset (then the baseline will be permanently shifted in the positive direction). For some examples, including a possible solution, please see the following notebook:

[https://gist.github.com/mwegrzyn/f58c1b3996c60ea7a4893dcc2906258c](https://gist.github.com/mwegrzyn/f58c1b3996c60ea7a4893dcc2906258c)

I hope that the suggested pull request is helpful. Thank you very much!

Best,
Martin


bthirion:
I'm happy with the changes, thx !
Need to investigate what's going with circleci...

doc/whats_new.rst
```diff
@@ -38,13 +38,19 @@ NEW
...
 Fixes
 -----
 
 - :class:`nilearn.input_data.NiftiLabelsMasker` no longer ignores its `mask_img`
 - :func:`nilearn.masking.compute_brain_mask` has replaced
   nilearn.masking.compute_gray_matter_mask. Features remained the same but
   some corrections regarding its description were made in the docstring.
+- the default background (MNI template) in plotting functions now has the
+  correct orientation; before left and right were inverted.
+- :func:`nilearn.glm.first_level.hemodynamic_models` can deal with regressors
```
thomasbazeille, line 51:
```suggestion
- first level modelling can deal with regressors
```
Indeed the hemodynamic_models.py is not a function which causes the CI to fail, nor a real module. Anyway I don't think we need to reference this precisely since its a bugfix no changes in API nor behavior

thomasbazeille (commented):
Looks good to me @mwegrzyn , please follow suggestion to make the CI pass and we can merge !

thomasbazeille:
Thanks, just solved the conflict with a recently merged PR also modifying what's new if CI is green we can merge

thomasbazeille:
Can you merge the master branch in your PR ? For some reason the CI still calls for non up-to-date functions. Sorry about that

bthirion:
beware the circleci error: [autosummary] failed to import 'nilearn.decoding.fREMClassifier': no module named nilearn.decoding.fREMClassifier

NicolasGensollen:
AFAICT everything seems to be looking good here. Should we merge?

bthirion:
Indeed, merging.

PR #2555, opened 2020-08-11
[FIX] Transpose array in permuted_ols

### LINKED ISSUE #2554: permuted_ols output is transposed when n_perm=0
I am using the current master branch, and I noticed that `permuted_ols` returns `score_orig_data` with the shape `(n_regressors, n_descriptors)` under most circumstances (e.g., if `n_perms > 0`), but when `n_perms <= 0`, it returns the data with the shape `(n_descriptors, n_regressors)`.

Here is where the t-statistics are typically returned (note the `.T`):

https://github.com/nilearn/nilearn/blob/4c1c7337b38506e6ef60873cfe3dd316ce869cf0/nilearn/mass_univariate/permuted_least_squares.py#L469

Here is where the t-statistics are returned when `n_perms == 0`:

https://github.com/nilearn/nilearn/blob/4c1c7337b38506e6ef60873cfe3dd316ce869cf0/nilearn/mass_univariate/permuted_least_squares.py#L444 

I think the only thing that needs to change is that, in L444, `scores_original_data` should be `scores_original_data.T`.

EDIT: The bug doesn't seem to have to do with intercept testing. Just the number of permutations.

tsalo:
Closes #2554.

- Transpose output array in `nilearn.mass_univariate.tests.test_permuted_least_squares.get_tvalue_with_alternative_library()
    - The docstring said the outputs would be in (n_regressors, n_descriptors) but they were actually in (n_descriptors, n_regressors).
- Transpose t-values output in `nilearn.mass_univariate.permuted_ols()` when `n_perm == 0`.
- Add a number of shape checks to tests in `nilearn.mass_univariate.tests`.
- Add some whitespace to the checks as well, to aid readability.

### DIFF
--- a/nilearn/mass_univariate/tests/test_permuted_least_squares.py
+++ b/nilearn/mass_univariate/tests/test_permuted_least_squares.py
@@ -27,7 +27,7 @@ def get_tvalue_with_alternative_library(tested_vars, target_vars, covars=None):
       Tested variates, the associated coefficient of which are to be tested
       independently with a t-test, resulting in as many t-values.
 
-    target_vars: array-like, shape=(n_samples, n_targets)
+    target_vars: array-like, shape=(n_samples, n_descriptors)
       Target variates, to be approximated with a linear combination of
       the tested variates and the confounding variates.
 
@@ -36,12 +36,12 @@ def get_tvalue_with_alternative_library(tested_vars, target_vars, covars=None):
 
     Returns
     -------
-    t-values: np.ndarray, shape=(n_regressors, n_targets)
+    t-values: np.ndarray, shape=(n_regressors, n_descriptors)
 
     """
     ### set up design
     n_samples, n_regressors = tested_vars.shape
-    n_targets = target_vars.shape[1]
+    n_descriptors = target_vars.shape[1]
     if covars is not None:
         n_covars = covars.shape[1]
         design_matrix = np.hstack((tested_vars, covars))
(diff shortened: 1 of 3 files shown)


nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -36,12 +36,12 @@ def get_tvalue_with_alternative_library(tested_vars, target_vars, covars=None):
...
     Returns
     -------
-    t-values: np.ndarray, shape=(n_regressors, n_targets)
+    t-values: np.ndarray, shape=(n_regressors, n_descriptors)
 
     """
     ### set up design
     n_samples, n_regressors = tested_vars.shape
-    n_targets = target_vars.shape[1]
+    n_descriptors = target_vars.shape[1]
```
bthirion, line 44: Hi, Can you explain the motivation for this change ?

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -356,126 +400,176 @@ def test_permuted_ols_withcovar_multivariate(random_state=0):
...
         tested_var, target_var, confounding_vars, n_perm=0,
         random_state=random_state)
+    assert own_scores.shape == (n_regressors, n_descriptors)
+    assert_array_almost_equal(ref_scores, own_scores, decimal=6)
+
     # same thing but with model_intercept=True to check it has no effect
     _, own_scores_intercept, _ = permuted_ols(
         tested_var, target_var, confounding_vars, model_intercept=True,
         n_perm=0, random_state=random_state)
-    assert_array_almost_equal(ref_scores, own_scores, decimal=6)
```
bthirion, line 431: why do remove that line ?

nilearn/mass_univariate/tests/test_permuted_least_squares.py
tsalo: This is what it's called in the `permuted_ols` docstring. I kept having to refer to the docstring when going through the tests because of the variable terminology, so I figured it would be easier to read if everything was as consistent as possible between the function docstring and the tests.

nilearn/mass_univariate/tests/test_permuted_least_squares.py
tsalo: I just moved it up to L501, so it would be located with the calculation of `own_scores`.

tsalo:
I just updated `whatsnew.rst`. Thanks!

bthirion:
Will merge in the next days. I'd be happy to have  a second review ?

PR #2561, opened 2020-09-01
BUG: Dont set backend on macOS

### LINKED ISSUE #2560: BUG: nilearn.plotting forces matplotlib backend change to Agg on macOS
on macOS:
```
>>> import matplotlib
>>> assert matplotlib.get_backend() == 'Qt5Agg'
>>> import nilearn.plotting
>>> assert matplotlib.get_backend() == 'Qt5Agg', matplotlib.get_backend()
AssertionError: agg
```
I have `backend : Qt5Agg\ninteractive : True` in `~/.matplotlib/matplotlibrc`. I assume it's due to this line:

https://github.com/nilearn/nilearn/blob/master/nilearn/plotting/__init__.py#L37

It seems like unwanted behavior that the backend is changed.

larsoner:
Closes #2560

### DIFF
--- a/nilearn/plotting/__init__.py
+++ b/nilearn/plotting/__init__.py
@@ -30,11 +30,11 @@ def _set_mpl_backend():
         if 'inline' in current_backend or 'nbagg' in current_backend:
             return
         # Set the backend to a non-interactive one for unices without X
-        if (os.name == 'posix' and 'DISPLAY' not in os.environ
-                and not (sys.platform == 'darwin'
-                         and 'macosx' in current_backend)
-        ):
+        # (see gh-2560)
+        if (sys.platform not in ('darwin', 'win32') and
+                'DISPLAY' not in os.environ):
             matplotlib.use('Agg')
+
 
 _set_mpl_backend()
 


nilearn/plotting/__init__.py
```diff
@@ -30,12 +30,12 @@ def _set_mpl_backend():
         if 'inline' in current_backend or 'nbagg' in current_backend:
             return
         # Set the backend to a non-interactive one for unices without X
-        if (os.name == 'posix' and 'DISPLAY' not in os.environ
-                and not (sys.platform == 'darwin'
-                         and 'macosx' in current_backend)
-        ):
+        # (see gh-2560)
+        if (sys.platform not in ('darwin', 'win32') and
+                'DISPLAY' not in os.environ):
```
agramfort, line 35: this changes a bit the logic as you assume that a mac will not be used as a remove headless server but i agree it's unlikely

I can live with that

larsoner: Related discussion is in #2560

FWIW I think/assume Travis and GitHub macOS CIs are headless and they gracefully fall back to software rendering anyway.

agramfort:
this solution is ok for  me but maybe DISPLAY is not defined on travis as
xquartz may not be installed.

larsoner:
> this solution is ok for  me but maybe DISPLAY is not defined on travis as xquartz may not be installed.

I don't see why this is a problem based on the conditional above. If the sys is macOS/darwin, then the presence of the DISPLAY variable is irrelevant (the second part of the conditional isn't evaluated)

jeromedockes:
To me it looks like matplotlib already selects an appropriate backend, starting here:
https://github.com/matplotlib/matplotlib/blob/f7ae15780857c9ddedd09dd614b68069bcb668cb/lib/matplotlib/pyplot.py#L234

also, to find out which interactive framework is available they seem to use `cbook._get_running_interactive_framework()`

btw, I think all the non-interactive backends ie pdf, ps, svg, cairo ... should be left alone and not just nbagg and inline

agramfort (approved):

bthirion:
Have we reached a consensus ? If everybody's happy, I'll merge tomorrow.

thomasbazeille:
Seems ready for merge @bthirion

bthirion:
Thx !

jeromedockes:
Can someone reproduce the problem where matplotlib chooses a wrong
backend on a headless machine, which makes it necessary for nilearn to
set the backend?

Issue #2395, opened 2019-05-28
z-values from compute_contrast are inaccurate for p-values close to 1


mwegrzyn:
Dear nistats team,
using the compute_contrast function, my colleague and I noticed that if one contrast and its exact inversion are plotted, there are some inconsistencies regarding negative z-scores.

> What version of Nistats are you using?

0.0.1b

> What were you trying to do?

use compute_contrast to make z-score maps

> What did you expect will happen?

If i have a contrast [+1 -1] and a contrast [-1 +1], the z-scores should be inverted, respectively

> What actually happened?

z-scores cannot go lower than -8.2, because the associated p-values (e.g. 0.99999999999999999) are represented as 1, while the corresponding p-values of the inverted contrast (e.g. 0.00000000000000001) are not.
See the following notebook which uses the example "2.1.2. Analysis of a single session, single subject fMRI dataset" from the nistats page to illustrate the issue:

[https://gist.github.com/mwegrzyn/9349842ee2abbff05df735a595bb239a](https://gist.github.com/mwegrzyn/9349842ee2abbff05df735a595bb239a)

Thank you.

Best regards,
Martin Wegrzyn

alpinho:
@bthirion @mwegrzyn I'll start working on this problem. Do you have any further suggestions besides what is reported [here](https://gist.github.com/mwegrzyn/9349842ee2abbff05df735a595bb239a)?

mwegrzyn:
Dear @alpinho,

thank you very much for your message and for working on the issue. Right now, I don't have any further ideas. But if you have anything that you need help with, I would be happy to try and contribute something (maybe tests or documentation; I'm not too experienced with working on software, but am looking for opportunities to gain more experience). Thank you!

Best regards,
Martin

alpinho:
[...]

PR #2567, opened 2020-09-19
FIXUP: innacurate z-values for big p-values

alpinho:
This is a first attempt to solve the issue #2395, according to what suggested by @mwegrzyn [here](https://gist.github.com/mwegrzyn/9349842ee2abbff05df735a595bb239a).


nilearn/_utils/glm.py
```diff
@@ -145,7 +145,11 @@ def z_score(pvalue):
     """ Return the z-score corresponding to a given p-value.
     """
     pvalue = np.minimum(np.maximum(pvalue, 1.e-300), 1. - 1.e-16)
-    return norm.isf(pvalue)
+    if np.any(pvalue > .5):
+        zscore = - norm.isf(1 - pvalue)
+    else:
+        zscore = norm.isf(pvalue)
+    return zscore
```
bthirion, line 152: Actually, this function does not does not change anything for me. A way to soleve this is to have a test that fails with the previous versions and passes with the new one.

nilearn/glm/tests/test_utils.py
```diff
@@ -4,8 +4,12 @@
 import numpy as np
 import pandas as pd
 import pytest
+import nibabel as nib
```
jeromedockes, line 7: some of these new imports are not used: nib, shutil, array

nilearn/_utils/glm.py
```diff
@@ -145,6 +145,10 @@ def z_score(pvalue):
     """ Return the z-score corresponding to a given p-value.
     """
     pvalue = np.minimum(np.maximum(pvalue, 1.e-300), 1. - 1.e-16)
+    # if np.any(pvalue > .5):
```
jeromedockes, line 148: why is the change commented out?

nilearn/glm/tests/test_utils.py
```diff
@@ -16,6 +20,9 @@
                                 get_design_from_fslmat, multiple_fast_inverse,
                                 multiple_mahalanobis, parse_bids_filename,
                                 positive_reciprocal, z_score)
+from nilearn.glm.first_level import FirstLevelModel
+from nilearn.datasets import fetch_spm_auditory
```
jeromedockes, line 24: you can't use `fetch_spm_auditory` in tests; they won't be able to download anything (this is by design, to avoid failures due to data servers' failures, and to avoid slowing down tests with long downloads). you have to use data that you generate (see if anything helps in the `data_gen` module) or that is already stored in the package itself (such as the mni template)

nilearn/glm/tests/test_utils.py
```diff
@@ -38,6 +45,40 @@ def test_z_score():
     assert_array_almost_equal(z_score(np.float32(1.e-100)), norm.isf(1.e-300))
 
 
+def test_z_score_opposite_contrast():
+    # put data fetching here
+    subject_data = fetch_spm_auditory()
```
jeromedockes, line 50: this won't work because the internet connection is mocked. functions that test dataset fetchers configure the mock response fetchers receive and fill it with fake data, but here you're better off not using the datasets package at all

nilearn/glm/tests/test_utils.py
```diff
@@ -38,6 +45,40 @@ def test_z_score():
...
+    active_minus_rest = conditions['active'] - conditions['rest']
+    rest_minus_active = -active_minus_rest
+    z_map = fmri_glm.compute_contrast(active_minus_rest,
+                                      output_type='z_score')
+
+    z_map_opposite = fmri_glm.compute_contrast(
+        rest_minus_active, output_type='z_score')
+
+    # assertion
+    assert (z_map.get_data().ravel().min(),
```
jeromedockes, line 76: use `np.allclose` or `pytest.approx` to compare floating-point numbers

jeromedockes (commented):
thanks @alpinho . note you cannot download datasets in tests

nilearn/_utils/glm.py
alpinho: Because, for the moment, my goal is to reproduce the error in a test function. Then, I'll try to fix the problem itself. These lines were just a first attempt to solve the problem.

nilearn/glm/tests/test_utils.py
alpinho: Ah!... I see. It is by design not possible to fetch data in test functions. I was not aware of that.

nilearn/glm/tests/test_utils.py
alpinho: Yes, that's where I get the error. Concretely, I get a `TypeError: 'NoneType' object is not subscriptable` when trying to access `spm_auditory_data['events']` in `fetch_spm_auditory` function (line 2615 of nilearn/datasets/func.py).

nilearn/glm/tests/test_utils.py
jeromedockes: I believe it was always preferred not to download data but it only became systematically enforced recently in #2514

nilearn/_utils/glm.py
alpinho: The problem as reported by the author of the issue is reproducible in the latest version of nilearn, i.e. in vs. "0.7.0a". But in development, it is not possible to reproduce it exactly in the same way since we cannot download spm auditory data as pointed out by @jeromedockes and @KamalakerDadi. We need to produce fake data and be able to get z values > 7 in order to start seeing differences between the max value of a contrast and the min value of its reverse. Shall I still proceed with the development of the test function w/ fake data? Otherwise, what do you suggest?

jeromedockes:
> We need to produce fake data and be able to get z values > 7 in order to start seeing differences 

I imagine this shouldn't be a problem?

alpinho:
Perhaps it is not. I don't really know how to produce fake data and how `data_gen` module works. In "real" data and contrasts, you may see often values > 7 in elementary effects, e.g. when contrasting main_condition vs. rest, but not so frequently in more specific ones.

jeromedockes:
> Perhaps it is not. I don't really know how to produce fake data and how `data_gen` module works. In "real" data and contrasts, you may see often values > 7 in elementary effects, e.g. when contrasting main_condition vs. rest, but not so frequently in more specific ones.

if just pick a voxel and fill its column with a timeseries that is
highly correlated with your contrast you should see small p-values

alpinho:
The test function is working now with fake data. Probably, I added too many prints at the end, in order to make the issue more visible. Let me know what you all think? Plus, I'd like to have some advice on how to solve now the numerical problem. @KamalakerDadi @jeromedockes @bthirion @ja-che

tbng:
> The test function is working now with fake data. Probably, I added too many prints in the end of the function These are the results of the prints:
> 
> `-3.684085191397777 -3.684085191397902`
> `3.925253566151492 3.925253566151854`
> `-3.6627965516762706 -3.662796551676404`

The `AssertionError` is raised because these arrays are not equal -- technically. You can use [`numpy.testing.assert_almost_equal`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html) with up to 10 decimal points and I guess it should be enough.

> `37.0470962993612 8.209536151601387`
> 

But this is not the case. So it is the desired behaviour for the test case to fail here? 

> And this is the AssetionError given by the function:
> 
> `> assert np.all(bool_res)`
> `E assert False`
> `E + where False = <function all at 0x7fe59b4cba60>([True, True, True, False])`
> `E + where <function all at 0x7fe59b4cba60> = np.all`
> 
> `glm/tests/test_utils.py:82: AssertionError`

alpinho:
> > The test function is working now with fake data. Probably, I added too many prints in the end of the function These are the results of the prints:
> > `-3.684085191397777 -3.684085191397902`
> > `3.925253566151492 3.925253566151854`
> > `-3.6627965516762706 -3.662796551676404`
> 
> The `AssertionError` is raised because these arrays are not equal -- technically. You can use [`numpy.testing.assert_almost_equal`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html) with up to 10 decimal points and I guess it should be enough.

OK.
> 
> > `37.0470962993612 8.209536151601387`
> 
> But this is not the case. So it is the desired behaviour for the test case to fail here?
> 
> > And this is the AssetionError given by the function:
> > `> assert np.all(bool_res)`
> > `E assert False`
> > `E + where False = <function all at 0x7fe59b4cba60>([True, True, True, False])`
> > `E + where <function all at 0x7fe59b4cba60> = np.all`
> > `glm/tests/test_utils.py:82: AssertionError`

Exactly. For the moment, this is the desirable outcome. It means that we have reproduced the problem reported by the author of the issue.

tbng:
> > > The test function is working now with fake data. Probably, I added too many prints in the end of the function These are the results of the prints:
> > > `-3.684085191397777 -3.684085191397902`
> > > `3.925253566151492 3.925253566151854`
> > > `-3.6627965516762706 -3.662796551676404`
> > 
> > 
> > The `AssertionError` is raised because these arrays are not equal -- technically. You can use [`numpy.testing.assert_almost_equal`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html) with up to 10 decimal points and I guess it should be enough.
> 
> OK.
> 
> > > `37.0470962993612 8.209536151601387`
> > 
> > 
> > But this is not the case. So it is the desired behaviour for the test case to fail here?
> > > And this is the AssetionError given by the function:
> > > `> assert np.all(bool_res)`
> > > `E assert False`
> > > `E + where False = <function all at 0x7fe59b4cba60>([True, True, True, False])`
> > > `E + where <function all at 0x7fe59b4cba60> = np.all`
> > > `glm/tests/test_utils.py:82: AssertionError`
> 
> Exactly. For the moment, this is the desirable outcome. It means that we have reproduced the problem reported by the author of the issue.

Sory I don't understand this point, so you mean your latest push does not contain a fix to the original problem, which results in a failure of `glm/test_utils.py`?

alpinho:
> Sorry I don't understand this point, so you mean your latest push does not contain a fix to the original problem, which results in a failure of `glm/test_utils.py`?

Yes, my latest push does not contain the fix of the original problem. I built the test function to reproduce the error, as suggested by @bthirion. Now, I'd like to get some recommendations on how to solve the original problem.

alpinho:
I changed the AssertionError according to what recommended by @tbng . This is now the log error:

`E           AssertionError: `
`E           Arrays are not almost equal to 10 decimals`
`E            ACTUAL: 37.0470962993612`
`E            DESIRED: 8.209536151601387`

`glm/tests/test_utils.py:82: AssertionError`

tbng:
> I changed the AssertionError according to what recommended by @tbng . This is now the log error:
> 
> `E AssertionError: `
> `E Arrays are not almost equal to 10 decimals`
> `E ACTUAL: 37.0470962993612`
> `E DESIRED: 8.209536151601387`
> 
> `glm/tests/test_utils.py:82: AssertionError`

Ok again to avoid confusion for other reviewers maybe we should state that this is the desired behaviour on the test case side so now I think we have a good one. 

The remaining part is that you need suggestions for fixing the PR I guess?

alpinho:
> The remaining part is that you need suggestions for fixing the PR I guess?

I received some very good tips today from @ja-che offline. I'll try to move forward with what he said, which relates to not only estimate the survival function but also the cumulative distribution function, probably here:

[p_values = sps.t.sf(self.stat_, np.minimum(self.dof, self.dofmax))](https://github.com/alpinho/nilearn/blob/c07cd45dd3e401d6dfd4e06d26c5c5c0bd70f8f9/nilearn/glm/contrasts.py#L229)

But, please, let me know if you have any hint, too.

alpinho:
I committed my first attempt to solve the real problem, which doesn't work. Changes are commented and they must be run one at the time for each script. For `glm.py`, do not forget to comment line 148; for `contrasts.py`, do not forget to comment line 230. @KamalakerDadi @ja-che @tbng @bthirion

alpinho:
I've made some fixes and it seems it has passed all checks but codecov/patch and Travis CI. Can some someone revise it? Thank you! @bthirion @KamalakerDadi @jeromedockes

NicolasGensollen:
@alpinho It looks like Travis is failing for tests related to the minimum versions of dependencies. It seems like `Nifti1Image` objects don't have the attribute `get_fdata` in nibabel 2.0.2 which causes your test `test_z_score_opposite_contrast` to fail here.

jeromedockes:
> @alpinho It looks like Travis is failing for tests related to the minimum versions of dependencies. It seems like `Nifti1Image` objects don't have the attribute `get_fdata` in nibabel 2.0.2 which causes your test `test_z_score_opposite_contrast` to fail here.

you can use nilearn.image.get_data instead

alpinho:
Thank you! I made the corrections you proposed. But codecov/patch is still failing. What should I do?

nilearn/glm/contrasts.py
```diff
@@ -262,6 +265,37 @@ def p_value(self, baseline=0.0):
         self.p_value_ = p_values
         return p_values
 
+
+    def one_p_value(self, baseline=0.0):
```
bthirion, line 269: I don't understand why it is necessary to introduce a new function.

bthirion: Oh, I understand, this is one_minus_pvalue !
Please rename this !

nilearn/glm/tests/test_utils.py
```diff
@@ -32,11 +37,64 @@ def test_full_rank():
...
 def test_z_score():
     p = np.random.RandomState(42).uniform(size=10)
-    assert_array_almost_equal(norm.sf(z_score(p)), p)
+    one_p = p[::-1]
+    zval_sf = norm.isf(p)
+    zval_cdf = norm.ppf(one_p)
+    zval = np.zeros(p.size)
+    zval[zval_sf < 0] = zval_cdf[zval_sf < 0]
+    zval[zval_sf >= 0] = zval_sf[zval_sf >= 0]
+    assert_array_almost_equal(z_score(p, one_p), zval)
```
bthirion, line 46: Hum, this is a bit disturbing one_p is not 1-p ?

nilearn/glm/tests/test_utils.py
```diff
@@ -32,11 +37,64 @@ def test_full_rank():
...
-    assert_array_almost_equal(z_score(np.float32(1.e-100)), norm.isf(1.e-300))
+        one_p = p
+        zval_sf = norm.isf(p)
+        zval_cdf = norm.ppf(one_p)
+        if p <= .5:
+            zval = zval_sf
+        else:
+            zval = zval_cdf
+        assert_array_almost_equal(z_score(p, one_p), zval)
+    assert_array_almost_equal(z_score(np.float32(1.e-100),
```
bthirion, line 57: What is the numerical accuracy of assert_array_almost_equal ?

bthirion (commented):
Thx.  I left a few correction requests.

Among the test, which one would fail with the current version of the code ?

NicolasGensollen:
@alpinho you get a test failure on Travis with the minimum requirement settings. 
More precisely, with numpy `1.11.0`, 0-d boolean arrays were not interpreted as valid index yet, so the line 

`z_scores[z_scores_sf >= 0] = z_scores_sf[z_scores_sf >=  0]`  in the `z_score` function is failing.

Not sure what the best solution would be, but a quick fix is to replace this line with:

 `z_scores[np.atleast_1d(z_scores_sf >= 0)] = z_scores_sf[z_scores_sf >=  0]`

which should work with all numpy's version.

alpinho:
> @alpinho you get a test failure on Travis with the minimum requirement settings.
> More precisely, with numpy `1.11.0`, 0-d boolean arrays were not interpreted as valid index yet, so the line
> 
> `z_scores[z_scores_sf >= 0] = z_scores_sf[z_scores_sf >= 0]` in the `z_score` function is failing.
> 
> Not sure what the best solution would be, but a quick fix is to replace this line with:
> 
> `z_scores[np.atleast_1d(z_scores_sf >= 0)] = z_scores_sf[z_scores_sf >= 0]`
> 
> which should work with all numpy's versions.

Thank you! Your suggestion has worked!

alpinho:
> Thx. I left a few correction requests.
> 
> Among the test, which one would fail with the current version of the code ?

I tested all test functions from `nilearn/glm/tests/test_utils.py `. All of them are passing. Concretely, `test_z_score_opposite_contrast()` was created to specifically reproduce the original problem and modifications were performed in `test_z_score()` since this one was not passing at the beginning.

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,23 @@ def _check_run_tables(run_imgs, tables_, tables_name):
     return tables_
 
 
-def z_score(pvalue):
+def z_score(sfvalue, cdfvalue):
     """ Return the z-score corresponding to a given p-value.
```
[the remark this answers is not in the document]
alpinho: OK. It is done.

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,23 @@ def _check_run_tables(run_imgs, tables_, tables_name):
     return tables_
 
 
-def z_score(pvalue):
+def z_score(sfvalue, cdfvalue):
```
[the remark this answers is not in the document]
alpinho: I replaced `sfvalue` by `pvalue`, as suggested. Nevertheless, both `pvalue` and `cdfvalue` are required in this function.

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,23 @@ def _check_run_tables(run_imgs, tables_, tables_name):
...
-    return norm.isf(pvalue)
+    sfvalue = np.minimum(np.maximum(sfvalue, 1.e-300), 1. - 1.e-16)
+    cdfvalue = np.minimum(np.maximum(cdfvalue, 1.e-300), 1. - 1.e-16)
+    z_scores_sf = norm.isf(sfvalue)
+    z_scores_cdf = norm.ppf(cdfvalue)
+    if isinstance(sfvalue, (np.ndarray, list)):
+        z_scores = np.zeros(len(sfvalue))
+        z_scores[z_scores_sf < 0] = z_scores_cdf[z_scores_sf < 0]
+        z_scores[z_scores_sf >= 0] = z_scores_sf[z_scores_sf >= 0]
+    elif isinstance(sfvalue, (int, float, np.float32)):
```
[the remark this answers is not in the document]
alpinho: Ups, `pvalue` cannot definitely be an int. I apologize for the mistake.

This snippet was an attempt to accommodate both lists.arrays and floats. But I have re-written it. Please, consider the current version.

nilearn/glm/contrasts.py
alpinho: OK, I have renamed it. It is now called `one_minus_pvalue`.

nilearn/glm/tests/test_utils.py
alpinho: This is definitely another mistake from my side. But I believe 1-p is not the solution, either. If it was, we would not need to solve this problem at the first place. 

In this test function, I have now replaced the random generation of p-values (values within the interval [0.0, 1.0]) by the random sampling of the standard Student’s t distribution. Then, this values were converted into p-value and "cdf-value" using the sf and the cdf, respectively.

I have done the exactly same thing---and written an additional test---for F-values, by randomly sampling the F-distribution.

alpinho: Also,  note that for the first test of the numerical precision, I replaced the original p-values by t-values which will return p-values with the survival function of the same order of magnitude.

nilearn/glm/tests/test_utils.py
alpinho: It is the default, which is 7.

bthirion: But then 1.e-100 = 0, so the test is not useful. Sorry if I isunderstood

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,82 @@ def test_full_rank():
...
+        print(z_map_seed1_vs_seed2.get_data().min())
+        print(-z_map_seed2_vs_seed1.get_data().max())
+        npt.assert_almost_equal(z_map_seed1_vs_seed2.get_data().min(),
+                                -z_map_seed2_vs_seed1.get_data().max(),
+                                decimal=10)
+        print(z_map_seed1_vs_seed2.get_data().max())
+        print(-z_map_seed2_vs_seed1.get_data().min())
+        npt.assert_almost_equal(z_map_seed1_vs_seed2.get_data().max(),
+                                -z_map_seed2_vs_seed1.get_data().min(),
+                                decimal=10)
```
bthirion, line 115: Note sure what is tested here: since the data are pure noise you won't have extreme z_values on one side or the other, will you ?

bthirion (commented):
Thx, this is better. I left a few additional comments/suggestions.

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,31 @@ def _check_run_tables(run_imgs, tables_, tables_name):
     return tables_
 
 
-def z_score(pvalue):
-    """ Return the z-score corresponding to a given p-value.
+def z_score(pvalue, cdfvalue):
```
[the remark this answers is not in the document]
alpinho: OK. Should it be set to "True' or "False" by default? I think it should be "True". Why wouldn't the user wish to have numerical accuracy by default?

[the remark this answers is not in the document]
alpinho: Ok, I set the 2nd arg to None as default in the function. If None, the zscore is computed as before. If not None, then it does the new extra calculations added w/ the float or array assigned. In `nilearn/glm/contrasts.py` and `nilearn/glm/tests/test_utils.py`, I set the arg of this parameter to one_minus_pvalue, when z_score is called.

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,82 @@ def test_full_rank():
...
+    cdf_val = sps.f.cdf(fval, 42, 1e10)
+    p_val = np.array(np.minimum(np.maximum(p_val, 1.e-300), 1. - 1.e-16))
+    cdf_val = np.array(np.minimum(np.maximum(cdf_val, 1.e-300), 1. - 1.e-16))
+    z_val_sf = norm.isf(p_val)
+    z_val_cdf = norm.ppf(cdf_val)
+    z_val = np.zeros(p_val.size)
+    z_val[np.atleast_1d(z_val_sf < 0)] = z_val_cdf[z_val_sf < 0]
+    z_val[np.atleast_1d(z_val_sf >= 0)] = z_val_sf[z_val_sf >= 0]
+    assert_array_almost_equal(z_score(p_val, cdf_val), z_val)
     # check the numerical precision
```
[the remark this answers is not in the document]
alpinho: Ok, it is done.

nilearn/glm/tests/test_utils.py
alpinho: This function is testing the numerical accuracy for a range of z values and it was raising an error for more extreme ones, but not anymore. This problem starts for |z| > 7 and it was visible in the spm_auditory data, as described in the original issue.

bthirion: But here, you may not encounter values of |z| greater than 7 ?

alpinho: Yes, we do. This is the log obtained from the prints in lines 125, 126, 130 and 131:

nilearn/glm/tests/test_utils.py::test_z_score_opposite_contrast 

-3.684085191397902
-3.684085191397902

3.925253566151492
3.925253566151492

-3.662796551676404
-3.662796551676404

37.0470962993612
37.0470962993612

alpinho: Can someone revise this PR? Thanks! :) @bthirion @KamalakerDadi @jeromedockes @NicolasGensollen

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,41 @@ def _check_run_tables(run_imgs, tables_, tables_name):
...
+def z_score(pvalue, one_minus_pvalue=None):
+    """ Return the z-score(s) corresponding to
+    certain p-value(s) and cdfvalues(s) provided as inputs.
+
+    Parameters
+    ----------
+    pvalue: float or 1-d array shape=(n_pvalues,) computed using
+            the survival function
+
+    cdfvalue: float or 1-d array shape=(n_cdfvalues,), optional;
```
jeromedockes, line 155: the name of this parameter has changed to `one_minus_pvalue`?

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,41 @@ def _check_run_tables(run_imgs, tables_, tables_name):
...
+              which computes the p_value using the
+              cumulative distribution function, with n_cdfvalues = n_pvalues
+
+    Returns
+    -------
+    z_scores: 1-d array shape=(n_z_scores,), with n_z_scores = n_pvalues
     """
-    pvalue = np.minimum(np.maximum(pvalue, 1.e-300), 1. - 1.e-16)
-    return norm.isf(pvalue)
+    pvalue = np.array(np.minimum(np.maximum(pvalue, 1.e-300), 1. - 1.e-16))
```
jeromedockes, line 165: you can use `np.clip`

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,41 @@ def _check_run_tables(run_imgs, tables_, tables_name):
...
+
+    if one_minus_pvalue is not None:
+        one_minus_pvalue = np.array(np.minimum(np.maximum(one_minus_pvalue,
+                                                          1.e-300),
+                                               1. - 1.e-16))
+        z_scores_sf = norm.isf(pvalue)
+        z_scores_cdf = norm.ppf(one_minus_pvalue)
+        z_scores = np.zeros(pvalue.size)
```
jeromedockes, line 173: you can use `np.empty` as it will be filled with other values.

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,41 @@ def _check_run_tables(run_imgs, tables_, tables_name):
...
+        z_scores[np.atleast_1d(z_scores_sf < 0)] = \
```
jeromedockes, line 174: could we avoid computing 4 times the comparison to 0? maybe something like `use_cdf = z_scores_sf < 0; z_scores[use_cdf] = ...; z_scores[~use_cdf] = ...`

jeromedockes: detail: when you can, break lines inside of brackets instead of escaping new lines with `\`

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,41 @@ def _check_run_tables(run_imgs, tables_, tables_name):
...
+            z_scores_cdf[z_scores_sf < 0]
+        z_scores[np.atleast_1d(z_scores_sf >= 0)] = \
+            z_scores_sf[z_scores_sf >= 0]
+    else:
+        z_scores = norm.isf(pvalue)
```
jeromedockes, line 179: would it be possible to add a test for the case where `one_minus_pvalue` is `None`?

nilearn/glm/contrasts.py
```diff
@@ -235,9 +236,11 @@ def stat(self, baseline=0.0):
         self.stat_ = stat.ravel()
         return self.stat_
 
+
     def p_value(self, baseline=0.0):
-        """Return a parametric estimate of the p-value associated
-        with the null hypothesis: (H0) 'contrast equals baseline'
+        """Return a parametric estimate using the survival function of the
```
jeromedockes, line 241: maybe "return a parametric estimate of the p-value associated [...], using the survival function" ? otherwise it reads like "... the survival function of the p-value"

nilearn/glm/contrasts.py
```diff
@@ -262,6 +265,35 @@ def p_value(self, baseline=0.0):
         self.p_value_ = p_values
         return p_values
 
+    def one_minus_pvalue(self, baseline=0.0):
+        """Return a parametric estimate using
+        the cumulative distribution function of the p-value associated
```
jeromedockes, line 270: isn't it the cdf of the computed statistic, rather than of the p-value? or maybe say it returns 1 - the pvalue, but computed differently for numerical stability

nilearn/glm/contrasts.py
```diff
@@ -262,6 +265,35 @@ def p_value(self, baseline=0.0):
...
+        baseline : float, optional
+            baseline value for the test statistic
+
+        Returns
+        -------
+        one_minus_pvalues : 1-d array, shape=(n_voxels,)
+            one_minus_pvalues, one per voxel
+        """
+        if self.stat_ is None or not self.baseline == baseline:
+            self.stat_ = self.stat(baseline)
```
jeromedockes, line 284: could you add a test for this case?

nilearn/glm/contrasts.py
```diff
@@ -262,6 +265,35 @@ def p_value(self, baseline=0.0):
...
+        # Valid conjunction as in Nichols et al, Neuroimage 25, 2005.
+        if self.contrast_type == 't':
+            one_minus_pvalues = sps.t.cdf(self.stat_,
+                                          np.minimum(self.dof, self.dofmax))
+        elif self.contrast_type == 'F':
+            one_minus_pvalues = sps.f.cdf(self.stat_, self.dim,
+                                          np.minimum(self.dof, self.dofmax))
+        else:
+            raise ValueError('Unknown statistic type')
```
jeromedockes, line 293: could you add a test for this case?

nilearn/glm/tests/test_utils.py
jeromedockes: > Yes, we do

can you add an assertion to make sure it will always be the case? (eg if the fake data generator is modified)

jeromedockes: so this test fails with the current nilearn but passes after merging this PR, is that correct?

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,101 @@ def test_full_rank():
...
+    z_val_cdf = norm.ppf(cdf_val)
+    # Create the final array of z-scores, ...
+    z_val = np.zeros(p_val.size)
+    # ... in which z-scores < 0 estimated w/ SF are replaced by z-scores < 0
+    # estimated w/ CDF
+    z_val[np.atleast_1d(z_val_sf < 0)] = z_val_cdf[z_val_sf < 0]
+    # ... and z-scores >=0 estimated from SF are kept.
+    z_val[np.atleast_1d(z_val_sf >= 0)] = z_val_sf[z_val_sf >= 0]
+    # Test 'z_score' function in 'nilearn/glm/contrasts.py'
+    assert_array_almost_equal(z_score(p_val, one_minus_pvalue=cdf_val), z_val)
```
jeromedockes, line 85: this doesn't seem to really test if the results are correct, as it compares them with the output of code that is almost identical to that of the function itself?

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,101 @@ def test_full_rank():
...
+    zval_cdf = norm.ppf(cdfval)
+    # Create the final array of z-scores, ...
+    zval = np.zeros(pval.size)
+    # ... in which z-scores < 0 estimated w/ SF are replaced by z-scores < 0
+    # estimated w/ CDF
+    zval[np.atleast_1d(zval_sf < 0)] = zval_cdf[zval_sf < 0]
+    # ... and z-scores >=0 estimated from SF are kept.
+    zval[np.atleast_1d(zval_sf >= 0)] = zval_sf[zval_sf >= 0]
+    # Test 'z_score' function in 'nilearn/glm/contrasts.py'
+    assert_array_almost_equal(z_score(pval, one_minus_pvalue=cdfval), zval)
```
jeromedockes, line 62: this doesn't seem to really test if the results are correct, as it compares them with the output of code that is almost identical to that of the function itself?

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,101 @@ def test_full_rank():
...
+        z_cdf = norm.ppf(cdf)
+        if p <= .5:
+            z = z_sf
+        else:
+            z = z_cdf
+        assert_array_almost_equal(z_score(p, one_minus_pvalue=cdf), z)
+
+
+def test_z_score_opposite_contrast():
+    fmri, mask = generate_fake_fmri(shape=(50, 20, 50), length=96)
```
jeromedockes, line 100: could you create a new random number generator and pass that to `generate_fake_fmri` ? the default generator used by generate_fake_fmri is only initialized once, so the result may depend on the order in which tests are executed. I'll open a separate issue about that

jeromedockes (commented):
almost there!

nilearn/glm/tests/test_utils.py
KamalakerDadi: Yes. With current nilearn the behavior should be as demonstrated [here](https://gist.github.com/mwegrzyn/9349842ee2abbff05df735a595bb239a) qualitatively.

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,101 @@ def test_full_rank():
...
+            contrasts['seed2 - seed1'], output_type='z_score')
+
+        print(z_map_seed1_vs_seed2.get_data().min())
+        print(-z_map_seed2_vs_seed1.get_data().max())
+        npt.assert_almost_equal(z_map_seed1_vs_seed2.get_data().min(),
+                                -z_map_seed2_vs_seed1.get_data().max(),
+                                decimal=10)
+        print(z_map_seed1_vs_seed2.get_data().max())
+        print(-z_map_seed2_vs_seed1.get_data().min())
+        npt.assert_almost_equal(z_map_seed1_vs_seed2.get_data().max(),
```
KamalakerDadi, line 132: ```assert_almost_equal``` is already imported without a need of numpy testing import ```npt```

nilearn/glm/tests/test_utils.py
KamalakerDadi: >can you add an assertion to make sure it will always be the case? 

It is already the case, no? ```assert_almost_equal```

KamalakerDadi:
It is almost close to merge once @jeromedockes comments are addressed.
Please add a whats_new entry.

nilearn/glm/tests/test_utils.py
alpinho: I have the same question.

nilearn/_utils/glm.py
alpinho: Yes, indeed. This has been fixed.

nilearn/_utils/glm.py
alpinho: OK. I have changed both the threshold assessment of pvalue and one_minus_pvalue using np.clip.

nilearn/_utils/glm.py
alpinho: OK. I have changed that, too.

nilearn/glm/contrasts.py
alpinho: OK, fixed!

nilearn/glm/contrasts.py
alpinho: OK, I changed this docstring accordingly, too.

nilearn/glm/tests/test_utils.py
alpinho: Yes, indeed. I have corrected that.

nilearn/glm/tests/test_utils.py
alpinho: What would be an alternative solution?

nilearn/glm/tests/test_utils.py
alpinho: What would be an alternative solution?

nilearn/_utils/glm.py
alpinho: OK. It is done in lines 64 and 90 of `nilearn/glm/tests/test_utils.py`.

nilearn/glm/tests/test_utils.py
alpinho: I assigned the arg `rand_gen=np.random.RandomState(42)` in `generate_fake_fmri` function? It is by default set to 0. Was that what you asked for?

nilearn/glm/contrasts.py
alpinho: This is basically a replication of the previous function `p_value` but using cdf instead of sf. I don't find a concrete test function for `p_value` that I could also replicate for this one. What do you suggest?

nilearn/glm/contrasts.py
alpinho: This is basically a replication of the previous function `p_value` but using cdf instead of sf. I don't find a concrete test function for `p_value` that I could also replicate for this one. What do you suggest?

alpinho:
I added the `one_minus_pvalue` to whats_new. I think I didn't miss any comment but some of them I've asked for further clarification. Thanks a lot and let me know! @jeromedockes @KamalakerDadi

nilearn/glm/contrasts.py
alpinho: I understand that this line has not been tested and that's why codecov/patch is complaining. I think I should add a test where also `p_value` function is tested but I don't find it.

nilearn/glm/contrasts.py
alpinho: I understand that this line has not been tested and that's why codecov/patch is complaining. I think I should add a test where also `p_value` function is tested but I don't find it.

nilearn/_utils/glm.py
bthirion: The code could be simplified (the `else` part reiterates statements of the other part.)

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,108 @@ def test_full_rank():
...
 def test_z_score():
-    p = np.random.RandomState(42).uniform(size=10)
-    assert_array_almost_equal(norm.sf(z_score(p)), p)
-    # check the numerical precision
-    for p in [1.e-250, 1 - 1.e-16]:
-        assert_array_almost_equal(z_score(p), norm.isf(p))
-    assert_array_almost_equal(z_score(np.float32(1.e-100)), norm.isf(1.e-300))
+    # ################# Check z-scores computed from t-values #################
+    # Randomly draw samples from the standard Student’s t distribution
+    tval = np.random.standard_t(1e10, size=10)
```
bthirion, line 42: Why do you use 1e10 dofs ? this amounts to rendering the t distribution identical to a normal distrbution...

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,108 @@ def test_full_rank():
...
+                                   hrf_model='spm', 
+                                   drift_model='cosine')
+        fmri_glm.fit(fmri, design_matrices=design_matrix)
+        z_map_seed1_vs_seed2 = fmri_glm.compute_contrast(
+            contrasts['seed1 - seed2'], output_type='z_score')
+        z_map_seed2_vs_seed1 = fmri_glm.compute_contrast(
+            contrasts['seed2 - seed1'], output_type='z_score')
+
+        print(z_map_seed1_vs_seed2.get_data().min())
+        print(-z_map_seed2_vs_seed1.get_data().max())
```
bthirion, line 133: Pleae remove the 'print' statements.

nilearn/glm/tests/test_utils.py
alpinho: Ups, yes. I changed that to 10. I've also changed the between- and within- group dof's of the F-distribution in line 68.

nilearn/glm/tests/test_utils.py
alpinho: OK.

nilearn/glm/tests/test_utils.py
jeromedockes: sorry I meant to assert @bthirion 's comment: assert there is indeed z-values greater than 7

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,42 @@ def _check_run_tables(run_imgs, tables_, tables_name):
...
+    pvalue = np.clip(np.clip(pvalue, 1.e-300, None), None, 1. - 1.e-16)
+    z_scores_sf = norm.isf(pvalue)
+
+    if one_minus_pvalue is not None:
+        one_minus_pvalue = np.clip(np.clip(one_minus_pvalue, 1.e-300, None),
+                                   None, 1. - 1.e-16)
+        z_scores_cdf = norm.ppf(one_minus_pvalue)
+        z_scores = np.empty(pvalue.size)
+        use_cdf = z_scores_sf < 0
+        use_sf = z_scores_sf >= 0
```
jeromedockes, line 176: maybe `use_sf = np.logical_not(use_cdf)`?

nilearn/glm/contrasts.py
```diff
@@ -235,9 +236,11 @@ def stat(self, baseline=0.0):
         self.stat_ = stat.ravel()
         return self.stat_
 
+
     def p_value(self, baseline=0.0):
-        """Return a parametric estimate of the p-value associated
-        with the null hypothesis: (H0) 'contrast equals baseline'
+        """Return a parametric estimate of the p-value associated with
+        the null hypothesis (H0) where 'contrast equals baseline',
```
jeromedockes, line 242: what does the "where" mean here?

nilearn/glm/contrasts.py
```diff
@@ -262,6 +265,36 @@ def p_value(self, baseline=0.0):
         self.p_value_ = p_values
         return p_values
 
+    def one_minus_pvalue(self, baseline=0.0):
+        """Return a parametric estimate of the 1 - p-value associated with
+        the null hypothesis (H0) where 'contrast equals baseline',
```
jeromedockes, line 270: why "where"?

nilearn/glm/contrasts.py
jeromedockes: maybe just calling `p_value` and `one_minus_pvalue` for models that haven't computed `stat_` yet?

nilearn/glm/tests/test_utils.py
jeromedockes: the comparison with opposite contrast (below) seems to be a more sensible check -- and it is how the issue was initially reported, right? so maybe just remove this test?

nilearn/_utils/glm.py
alpinho: OK.

nilearn/glm/tests/test_utils.py
jeromedockes: yes, that works! by default it is not set to 0, but to a mutable object `RandomState(0)` which is initialized once and then used/ modified by several test functions in unpredictable order, see #2631 .

jeromedockes:
only minor comments on my part, seems ready for merge

nilearn/glm/contrasts.py
alpinho: It is a way to phrase what the null hypothesis means. But perhaps, it is not clear. What do you suggest?

nilearn/glm/contrasts.py
alpinho: It is a way to phrase what the null hypothesis means. But perhaps, it is not clear. What do you suggest?

nilearn/glm/contrasts.py
jeromedockes: For me the previous phrasing
  "the null hypothesis (H0): 'contrast equals baseline'"
was clear, but if you prefer the new phrasing that's ok too!

nilearn/glm/tests/test_utils.py
alpinho: But this test function tests whether the `z_score` function from `nilearn/_utils/glm.py` is working. It was already here before. I had to change it because I changed the `z_score` function and then this test function started producing an error. I had thus to adapt it to the present context.

nilearn/glm/contrasts.py
alpinho: OK, I just changed it.

nilearn/glm/contrasts.py
alpinho: I decided to remove this conditional statement as well as the one from the next comment. The reason these lines were not being tested by codecov/patch---and therefore producing an error---is because the exactly same conditional statements are always tested before in the function `p_value`, which takes place right before this one. Let me know whether you agree with this solution. @jeromedockes @bthirion

jeromedockes: I'm not sure why remove them? it is still possible for a user to specify an
invalid contrast type (different from "t" and "F"), and also to call
`one_minus_pvalue` without calling `stat` first -- so these checks are still
necessary. what was missing was a test covering these 2 situations.

nilearn/glm/contrasts.py
jeromedockes: I would suggest passing an invalid value as `contrast_type` to the model's `__init__`, then calling `one_minus_pvalue`

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,103 @@ def test_full_rank():
...
 def test_z_score():
-    p = np.random.RandomState(42).uniform(size=10)
-    assert_array_almost_equal(norm.sf(z_score(p)), p)
-    # check the numerical precision
-    for p in [1.e-250, 1 - 1.e-16]:
-        assert_array_almost_equal(z_score(p), norm.isf(p))
-    assert_array_almost_equal(z_score(np.float32(1.e-100)), norm.isf(1.e-300))
+    # ################# Check z-scores computed from t-values #################
+    # Randomly draw samples from the standard Student’s t distribution
+    tval = np.random.standard_t(10, size=10)
```
jeromedockes, line 42: please use a random number generator with a fixed seed

nilearn/glm/tests/test_utils.py
```diff
@@ -31,12 +37,103 @@ def test_full_rank():
...
+    # Estimate the p-values using the Survival Function (SF)
+    pval = sps.t.sf(tval, 1e10)
```
jeromedockes, line 44: why 1e10 degrees of freedom?

nilearn/glm/tests/test_utils.py
jeromedockes: also if 10 samples are drawn from the null hypothesis most likely there are no extreme values, so the precision issue doesn't appear? maybe add a few very large values to tval?

nilearn/glm/tests/test_utils.py
alpinho: This number is the maximum accepted when we estimate the p-values in `nilearn/glm/contrasts.py` lines 257, 260, 261, 285, 286, 288, 289.

nilearn/glm/tests/test_utils.py
alpinho: Note that in this function we are not testing the problem of numerical accuracy addressed in this PR. This is a function to check whether the `z_score` function from `nilearn/_utils/glm.py` is working, regardless whether the corresponding p-values are extreme or not. I used size=10 because this was the length of the list of p-values used in the original test. Besides, note that this function had to be changed only because the `z_score` function was changed but its scope is still the same.

nilearn/glm/contrasts.py
jeromedockes: @alpinho  AFAICT this is still to be addressed? ie handle the cases where the contrast type is invalid, and where `stat_` doesn't exist, in `one_minus_pvalue`, and add corresponding tests. I think that's the only one left!

alpinho: Sorry, I didn't see this comment before. I think the reason codecov/patch was producing an error is because these lines I removed were never being tested because they were always tested before in the `p_value` function. So, I am not sure whether that's true that it is still possible for a user to specify an invalid contrast type (different from "t" and "F") and also call `one_minus_pvalue` without calling `p_value` before. But I am happy to solve this problem in a different way; I just don't know exactly how. Concretely, I don't know where to add the tests for this; it should be in the same place where the same lines of `p_value` function are tested, but I don't find them...

jeromedockes: for example:
```python
import numpy as np
from nilearn.glm.contrasts import Contrast

effect = np.ones((1, 3))
variance = effect[0]

contrast = Contrast(effect, variance, contrast_type="t")
# fails because stat() has not been called
print(contrast.one_minus_pvalue())

# fails with cryptic error message because bad stat type not handled
contrast = Contrast(effect, variance, contrast_type="bad")
print(contrast.one_minus_pvalue())
```

alpinho: I would assume it should be in `nilearn/glm/tests/test_contrasts.py` but, if that's the case, where exactly?

jeromedockes: you can create a test `test_one_minus_pvalue`

jeromedockes: maybe something like
```python
def test_one_minus_pvalue():
    effect = np.ones((1, 3))
    variance = effect[0]

    contrast = Contrast(effect, variance, contrast_type="t")
    assert np.allclose(contrast.one_minus_pvalue(), 0.84, 1)
    assert np.allclose(contrast.stat_, 1., 1)

    contrast = Contrast(effect, variance, contrast_type="bad")
    with pytest.raises(ValueError):
        contrast.one_minus_pvalue()

```

alpinho: Thanks! I was working on this when I received your last suggestion. I did the second test in a very similar way but using numpy.testing. I was trying to figure out how to do the first test but you gave me a great a help. I copied/pasted it. I just committed everything. Let me know. Thanks!

alpinho: There's still one segment in red in Codecov Report diagram...

jeromedockes: ah yes, because the exception already gets raised in `stat()` when the stat type is invalid. that's ok, don't worry about it

nilearn/glm/tests/test_contrasts.py
```diff
@@ -164,3 +165,16 @@ def test_low_level_fixed_effects():
                                                precision_weighted=True)
     assert_almost_equal(Xw, 1.2 * X1)
     assert_almost_equal(Vw, .8 * V1)
+
+
+def test_one_minus_pvalue():
+    effect = np.ones((1, 3))
+    variance = effect[0]
+    # fails because stat() has not been called
```
jeromedockes, line 173: I think this comment can be removed?

nilearn/glm/tests/test_contrasts.py
```diff
@@ -164,3 +165,16 @@ def test_low_level_fixed_effects():
...
+    contrast_1 = Contrast(effect, variance, contrast_type="t")
+    assert np.allclose(contrast_1.one_minus_pvalue(), 0.84, 1)
+    assert np.allclose(contrast_1.stat_, 1., 1)
+    # fails with cryptic error message because bad stat type not handled
```
jeromedockes, line 177: same

jeromedockes (commented):
OK I'll do a final pass but I think that's it for me! thanks so much for all this work and your patience @alpinho

nilearn/glm/contrasts.py
alpinho: OK.

nilearn/glm/contrasts.py
alpinho: It seems that the second test of the function `test_one_minus_pvalue` is not working, since line 293 was not evaluated not evaluated. Any suggestion?

jeromedockes: line 293 will never run because an exception would be raised in `stat` before it is reached -- it's ok if it is not covered by the tests.

jeromedockes: if you want to avoid it you can do for example
```
if contrast type == t:
  ...
else:
  assert contrast_type == f
  ...
```
but it's ok the way it is, and it is what is done in p_value so you can just keep it

alpinho: I did what you suggested in my last commit and the problem is solved! But there's still a red segment in Codecov Report...

jeromedockes:
> I did what you suggested in my last commit and the problem is solved! But there's still a red segment in Codecov Report...
>
I don't see it what is the missed line?

alpinho:
> I don't see it what is the missed line?

If you scroll up the conversation tab, you can see a red segment depicted in Codecov Report diagram posted by codecov bot. It seems to be related to `reporting/_get_clusters_table.py`.

jeromedockes:
> If you scroll up the conversation tab, you can see a red segment depicted in Codecov Report diagram posted by codecov bot. It seems to be related to `reporting/_get_clusters_table.py`.
>

I see, that's fine -- no problem

bthirion (approved):
LGTM, thx.

nilearn/_utils/glm.py
```diff
@@ -143,11 +143,42 @@ def _check_run_tables(run_imgs, tables_, tables_name):
...
+    z_scores: 1-d array shape=(n_z_scores,), with n_z_scores = n_pvalues
     """
-    pvalue = np.minimum(np.maximum(pvalue, 1.e-300), 1. - 1.e-16)
-    return norm.isf(pvalue)
+    pvalue = np.clip(np.clip(pvalue, 1.e-300, None), None, 1. - 1.e-16)
+    z_scores_sf = norm.isf(pvalue)
+
+    if one_minus_pvalue is not None:
+        one_minus_pvalue = np.clip(np.clip(one_minus_pvalue, 1.e-300, None),
+                                   None, 1. - 1.e-16)
```
jeromedockes, line 172: same
```suggestion
        one_minus_pvalue = np.clip(one_minus_pvalue, 1.e-300, 1. - 1.e-16)
```

jeromedockes:
@KamalakerDadi  did you have more comments? otherwise I will merge it

KamalakerDadi:
>@KamalakerDadi did you have more comments? otherwise I will merge it

Greatly improved. No other comments

Issue #2481, opened 2020-05-21
ImportError: A sklearn version of at least 0.19 is required to use nilearn. 0.0.0 was found. Please upgrade sklearn


hvannieuwenh:
I recently installed nilearn version 0.6.2, and when I try "import nilearn" I receive the error written in the title. I cannot update sklearn as it is outdated, it has now been switched to scikit-learn, for which I have the latest version, 0.22.2.post1.

Downgrading my scikit-learn version to 0.22 fixed the problem and allowed me to import nilearn with no issues.

thomasbazeille:
Thanks for reaching out !

This is linked to #2288 and on our to-do list for next release as well. It might become a high priority if it makes nilearn fail with latest scikit-learn version.

Issue #2288, opened 2020-01-17
nilearn>=0.6.0 installs empty `sklearn` dependency


rmarkello:
Hi nilearn team!

I was recently creating a new Python environment and noticed the dependency chain for `nilearn` had changed. In particular, it seems that installing version 0.6.0 of `nilearn` into a clean environment automatically installs the (fake) [`sklearn`](https://pypi.org/project/sklearn/) package (in addition to the actual `scikit-learn` package). 

This dependency is not installed when using the previous release (0.6.0b0). Indeed, it looks like this functionality may have been added with #2201, and I believe the relevant code is here:

https://github.com/nilearn/nilearn/blob/9173aaca590df50bbaa6122bc744291617f36b90/setup.py#L33-L44

Specifically, L43 of that snippet adds the `sklearn` dependency to the `install_requires` list.

I may be missing something (the referenced PR has quite a lot happening so it was tough to follow everything!), but given that `sklearn` seems to only exist to prevent malicious actors (the package doesn't actually do anything...), would it be worth removing as a dependency?

GaelVaroquaux:
Right! We should have a dependency on "scikit-learn", not "sklearn". Good
catch!

rmarkello:
You do have a dependency on `scikit-learn` (L41 in the code snippet above makes sure of that)! However, I think there _is_ a reason that `sklearn` is listed in the `nilearn/version.py` file given some of the commits in #2201(e.g., 4cfb901), but I just can't tell post-hoc what that rationale is...

Nonetheless, I can't think of a good reason why it should be required in the final build wheel.

Just in case it helps, the full dependency tree for `nilearn` (from `pip install nilearn`) as of 0.6.0 is:

```
- nilearn
  - joblib [required: >=0.11]
  - nibabel [required: >=2.0.2]
    - numpy [required: >=1.12]
  - numpy [required: >=1.11]
  - scikit-learn [required: >=0.19]
    - joblib [required: >=0.11]
    - numpy [required: >=1.11.0]
    - scipy [required: >=0.17.0]
  - scipy [required: >=0.19]
  - sklearn [required: Any]
```

kchawla-pi:
It was deliberately added because one of our CIs, Appveyor, refused to work without it. 

My assessment was that trying to fix was a lot of cost for little benefit, since it didn't affect the code functionality or quality.
This also opened up the possibility that some systems out in the wild might glitch as well. 

[...]

PR #2504, opened 2020-05-28
Glm docs

swaythe:
Closes #2492 

Corrected some typos in the GLM intro section

Created the first level model section with the following subsections:
- HRF models
- Design matrix
- Fitting a first level model
- Extracting predicted time series and residuals
- Surface-based analysis
The first 3 have some details, the last 2 not yet. 

The second level file is a skeleton, no text added yet.

PR #2571, opened 2020-09-26
Removing fake sklearn import

thomasbazeille:
Tackling #2288, and its consequences as in #2481 (potential bugs with recent versions of scikit)

Now we don't use AppVeyor anymore, this should be useless


jeromedockes:
LGTM, thanks! the circleci failure is genuine but was introduced by #2504 and is unrelated to this PR

bthirion:
Merging, thx.

PR #2573, opened 2020-09-30
[WIP] Confounds numpy pandas

### LINKED ISSUE #2505: Standardizing pd.DataFrame vs np.array usage
Right now, different portions of the code base seem to expect `pd.DataFrame`s or `np.array`s. For example, I hit this issue when generating a set of confounds and then trying to call `NiftiMasker.fit_transform()` or `FirstLevelModel.fit()`.

NiftiMasker expects a `np.array` and rejects a `pd.DataFrame`:

https://github.com/nilearn/nilearn/blob/f03f7250a9cc60b00e28b41801b12e99202f20f0/nilearn/input_data/base_masker.py#L189-L191

FirstLevelModel expects a `pd.DataFrame` and rejects a `np.array`:

https://github.com/nilearn/nilearn/blob/f03f7250a9cc60b00e28b41801b12e99202f20f0/nilearn/stats/first_level_model/first_level_model.py#L359-L366

I can solve this by just providing file paths, or converting between the two array types for each function. But this seems like an easy stumbling block for users, and something that should be standardized !

bthirion:
Closes #2505 .


thomasbazeille (approved):
This PR seems alright to me, except maybe not showcasing the use of dataframe in documentation but maybe we don't need to highlight that. To be sure this completely answer the mentionned issue, we need to do more checks in other modules but this can be completed by a further PR if needed

thomasbazeille:
second_level module displays same problem

bthirion:
It is also important to improve documentation. I'm on it.

bthirion:
One point: I've quickly looked at `make_second_level_design_matrix`: there the confounds table has a necessary subject columns, so i'm not sure it would be a good idea to allow numpy arrays here.
Since this is for second-level modeling, we are not talking about the same confounds as for the other parts (e.g. high-variance regressor in first level-analysis), I'm not not sure we actually need to replace that. What do people think ?

bthirion:
Another point: in the `make_first_level_design_matrix`, we have the 'add_regs' argument that is formally similar to confounds: time-domain regressors, one per time frame. So we could rename that `confounds`. The problem  I see is that in some cases it represents a regressor of interest, supplied directly by the user, hence not a 'confound' stricto sensu: I am afraid that calling it `confounds` would create confusion for users. Any opinion on that @emdupre @jeromedockes ?

thomasbazeille:
same idea Parcellations accept confounds for fit() (inherited from MultiPCA) and transform method() for now, handled by a class specific method _check_parameters_transform(imgs, confounds) and not accepting Dataframes (nor np arrays I believe)

thomasbazeille:
> One point: I've quickly looked at `make_second_level_design_matrix`: there the confounds table has a necessary subject columns, so i'm not sure it would be a good idea to allow numpy arrays here.
> Since this is for second-level modeling, we are not talking about the same confounds as for the other parts (e.g. high-variance regressor in first level-analysis), I'm not not sure we actually need to replace that. What do people think ?

I agree with that, the usual confounds files have already been used in first level analysis, so additional confounds that should be added are quite peculiar and people can afford creating a dataframe for that. A list of np.array would be possible but might induce more errors

thomasbazeille:
> Another point: in the `make_first_level_design_matrix`, we have the 'add_regs' argument that is formally similar to confounds: time-domain regressors, one per time frame. So we could rename that `confounds`. The problem I see is that in some cases it represents a regressor of interest, supplied directly by the user, hence not a 'confound' stricto sensu: I am afraid that calling it `confounds` would create confusion for users. Any opinion on that @emdupre @jeromedockes ?

I think the current name is well suited,  `confounds` would be confusing, however the docstring could mention more clearly this can be used to add both confounding regressor and regressor of interests. Also this does not accept Dataframes for now right ? Which it maybe should ?

bthirion:
If nobody opposes, I'm merge later today.

jeromedockes:
LGTM!

PR #2575, opened 2020-10-01
Small refactoring of reporting and plotting

### LINKED ISSUE #2488: Discussion: Should there be plotting in nilearn.reporting?
There are a lot of plotting functions in nilearn.reporting. I do not really comprehend the logic that drove the choice of putting a function in plotting vs reporting. The whole content of the _plot_matrices file feels like I would expect it to be in plotting, as there is plotting.plot_matrix there.

I would expect to draw the line in the following way:
* Anything that simply does matplotlib plot should be in plotting
* Anything that outputs HTML reports should go in nilearn.reporting

jeromedockes:
closes #2488


bthirion:
The PR LGTM. The non-covered parts were not covered previously I guess ? Maybe we can keep that for a future PR.

jeromedockes:
> The PR LGTM. The non-covered parts were not covered previously I guess ? Maybe we can keep that for a future PR.

I'll see if I can add tests today and if I can use the functions that
were made public in existing examples

jeromedockes:
I added some tests. As @bthirion pointed out no new code has been added,
these are tests for functions that already existed before. there seems to
be some problem with codecov.io; I can't access the online reports ATM

the `plot_img_comparison` doesn't work for 4d images, only lists of 3d,
and when `plot_hist=False` it creates axes but doesn't plot anything
inside -- this could be addressed in another PR if we think it is
worthwhile

I made first level and second level both inherit from a BaseGLM to
share the docstring and (one-line but with many kwargs to forward)
implementation of `generate_report`, but if you would rather not have
this base class I can remove it

At some point we may want to have a `make_nifti_masker_report` as we
have the `make_glm_report`; but we can leave it for another PR -- for
example the PRs that will add reporting to other objects, as these PRs
will likely make further changes in `html_report` and the NiftiMasker's
`generate_report`

bthirion:
@jeromedockes 
I suggest to leave this PR as-is ---it is complex enough. and open issues for the other improvements you mention.
Regarding BaseGLM, it does not hurt, but what benefit is expected from it ?

jeromedockes:

> @jeromedockes 
> I suggest to leave this PR as-is ---it is complex enough. and open issues for the other improvements you mention.

I agree

> Regarding BaseGLM, it does not hurt, but what benefit is expected from it ?

it's to avoid duplicating the `generate_report` method and its docstring
in firstlevel and secondlevel models. But if you prefer to remove it I
don't mind either way

emdupre (approved):
These changes make sense to me, and I think bring us to a sensible UX for reporting v plotting. Thank you, @jeromedockes !

bthirion:
Great, merging.

Issue #1744, opened 2018-08-30
Finding Optimal Number of Components using dict_learning as well as canICA


nilover:
Dear All,
Thanks a lot for helping in Voxel based Morphometry using dictionary learning as well as using canICA.
Regarding selecting the optimal number of components, we can select the log (the number of data points). So the number of data points would be the total number of data points right?

log(x=60 * y=72 * z=60 * n=600) = 18.8 ~19  so
n_components = 19 ??
[here 600 is the number of subjects]

Thanks a lot.

bthirion:
There's no good number. You just need to pick one, reasonable in view of the literature and of the data. People typically use between 20 and 70...  (assuming a much larger number of images).

nilover:
Many thanks @bthirion. Sure we can choose the expected number of components based on the literature review as well as from the number of available images. However is there any way to create "elbow plot" or any way to measure variance of each component in canICA and dict_ learning algorithm?

nilover:
Thanks a lot @bthirion. Yes absolutely, as you advised the selection of number of components should be always based on literature survey and number of available images. 
Thank again for indicating about the score function. I just would like to check about the rough indication of expected number of components.

I have tried the score function for checked the explained variance of each component imgs but the output is 0.0. Now I am really not sure if I am using the correct syntax.

> "estimator.score(components_imgs, confounds=None)" 

Score function mentioned in the attributes and it should be used by DecompositionEstimator derived classes.

> score(imgs, confounds=None)
> Score function based on explained variance on imgs.
> 
> Should only be used by DecompositionEstimator derived classes

So the output of "estimator.score(components_imgs, confounds=None) " is 0.0

Thanks a lot again

nilover:
Thanks a lot @bthirion for the idea of using score function to get a rough idea about number of components. However I could find how the score was calculated using estimated variance per component. Is it cumulative? Because in my case, the variation in the score value is constant (1.0) after 50 components ( k =100), so that means 50 could be rough indication of the expected number of components from the data? I was also looking for score function (source code) to check how it is calculating score of explained variance for each components, but could not find it yet.

Thanks a lot again.

bthirion:
Indeed, the `score()` methods seems brken. Sorry for that. I'll open an issue.

nilover:
Okay thanks a lot @bthirion. I will wait for it.

KamalakerDadi:
[...]

PR #2577, opened 2020-10-06
[MRG] FIX score method with decomposition estimators

### LINKED ISSUE #2342: `explained_variance` yields spurious results when `per_component`=False
Due to this, score functions of canica and dict_learning are useless: they most often return 0.
Can be checked after running 
https://nilearn.github.io/auto_examples/03_connectivity/plot_compare_decomposition.html#sphx-glr-auto-examples-03-connectivity-plot-compare-decomposition-py

KamalakerDadi:
Closes #2342 #1744 .

- The explained variance `score` method now returns a value based on the number of components rather than 0. always. 
- ```per_component```, a new argument introduced in `score` method returns a value per component.

### DIFF
--- a/examples/03_connectivity/plot_compare_decomposition.py
+++ b/examples/03_connectivity/plot_compare_decomposition.py
@@ -129,6 +129,23 @@ for i, cur_img in enumerate(iter_img(dictlearning_components_img)):
     plot_stat_map(cur_img, display_mode="z", title="Comp %d" % i,
                   cut_coords=1, colorbar=False)
 
+###############################################################################
+# Estimate explained variance per component and plot using matplotlib
+#
+# The fitted object `dictlearning` can be used to calculate the score per component
+scores = dict_learning.score(func_filenames, per_component=False)
+
+# Plot the scores
+import numpy as np
+from matplotlib import pyplot as plt
+
+plt.figure(figsize=(6, 4))
+plt.plot(scores, label='Explained variance score per component')
+plt.axis('tight')
+plt.xlabel('Component #')
+plt.xticks(np.arange(20))
+plt.legend(loc='best', frameon=False)
+
 
 show()
 ################################################################################
--- a/examples/03_connectivity/plot_compare_decomposition.py
+++ b/examples/03_connectivity/plot_compare_decomposition.py
@@ -132,20 +132,22 @@ for i, cur_img in enumerate(iter_img(dictlearning_components_img)):
 ###############################################################################
 # Estimate explained variance per component and plot using matplotlib
 #
-# The fitted object `dictlearning` can be used to calculate the score per component
+# The fitted object `dict_learning` can be used to calculate the score per component
 scores = dict_learning.score(func_filenames, per_component=True)
 
 # Plot the scores
 import numpy as np
 from matplotlib import pyplot as plt
+from matplotlib.ticker import FormatStrFormatter
 
-plt.figure(figsize=(6, 4))
-plt.plot(scores, label='Explained variance score per component')
-plt.axis('tight')
-plt.xlabel('Component #')
-plt.xticks(np.arange(20))
-plt.legend(loc='best', frameon=False)
-
+plt.figure(figsize=(4, 4))
+positions = np.arange(len(scores))
+plt.barh(positions, scores)
+plt.ylabel('Component #', size=12)
+plt.xlabel('Explained variance', size=12)
+plt.yticks(np.arange(20))
+plt.gca().xaxis.set_major_formatter(FormatStrFormatter('%.3f'))
+plt.tight_layout()
 
 show()
 ################################################################################
--- a/examples/03_connectivity/plot_compare_decomposition.py
+++ b/examples/03_connectivity/plot_compare_decomposition.py
@@ -133,7 +133,7 @@ for i, cur_img in enumerate(iter_img(dictlearning_components_img)):
 # Estimate explained variance per component and plot using matplotlib
 #
 # The fitted object `dictlearning` can be used to calculate the score per component
-scores = dict_learning.score(func_filenames, per_component=False)
+scores = dict_learning.score(func_filenames, per_component=True)
 
 # Plot the scores
 import numpy as np
@@ -155,4 +155,3 @@ show()
 #     created using Dictionary Learning, see :ref:`example Regions 
 #     extraction using Dictionary Learning and functional connectomes
 #     <sphx_glr_auto_examples_03_connectivity_plot_extract_regions_dictlearning_maps.py>`.
-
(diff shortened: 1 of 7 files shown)


examples/03_connectivity/plot_compare_decomposition.py
```diff
@@ -129,6 +129,23 @@
     plot_stat_map(cur_img, display_mode="z", title="Comp %d" % i,
                   cut_coords=1, colorbar=False)
 
+###############################################################################
+# Estimate explained variance per component and plot using matplotlib
+#
+# The fitted object `dictlearning` can be used to calculate the score per component
+scores = dict_learning.score(func_filenames, per_component=False)
```
bthirion, line 136:
```suggestion
scores = dict_learning.score(func_filenames, per_component=True)
```
(?)

nilearn/tests/test_signal.py
```diff
@@ -266,6 +266,23 @@ def test_mean_of_squares():
...
 
+def test_sum_of_squares():
+    """Test _sum_of_squares."""
+    n_samples = 11
+    n_features = 501  # Higher than 500 required
+    signals, _, _ = generate_signals(n_features=n_features,
+                                     length=n_samples,
+                                     same_variance=True)
+    # Reference computation
+    var1 = np.copy(signals)
```
bthirion, line 277: There's no motivation to do a copy here ?

nilearn/signal.py
```diff
@@ -120,6 +120,41 @@ def _mean_of_squares(signals, n_batches=20):
...
+    # No batching for small arrays
+    if signals.shape[1] < 500:
+        n_batches = 1
+
+    # Fastest for C order
+    var = np.empty(signals.shape[1])
+    for batch in gen_even_slices(signals.shape[1], n_batches):
+        tvar = np.copy(signals[:, batch])
+        tvar **= 2
+        var[batch] = tvar.sum(axis=0)
```
bthirion, line 153: Why not
`var[batch] =  np.sum(signals[:, batch] ** 2, 0)` that avoids a copy ?

bthirion (commented):
Thx for reviving that one !

examples/03_connectivity/plot_compare_decomposition.py
KamalakerDadi: >(?)

Yes, that should be the right one. But, I am checking something. Will get back to ```True```

KamalakerDadi:
@bthirion this is good to go.

examples/03_connectivity/plot_compare_decomposition.py
```diff
@@ -129,6 +129,23 @@
...
+#
+# The fitted object `dictlearning` can be used to calculate the score per component
+scores = dict_learning.score(func_filenames, per_component=True)
+
+# Plot the scores
+import numpy as np
+from matplotlib import pyplot as plt
+
+plt.figure(figsize=(6, 4))
+plt.plot(scores, label='Explained variance score per component')
```
bthirion, line 143: Maybe a bar plot would be better suited here.

examples/03_connectivity/plot_compare_decomposition.py
```diff
@@ -129,6 +129,23 @@
     plot_stat_map(cur_img, display_mode="z", title="Comp %d" % i,
                   cut_coords=1, colorbar=False)
 
+###############################################################################
+# Estimate explained variance per component and plot using matplotlib
+#
+# The fitted object `dictlearning` can be used to calculate the score per component
+scores = dict_learning.score(func_filenames, per_component=True)
```
bthirion, line 136: I find that scores.sum() is differnt from `dict_learning.score(func_filenames, per_component=False)`
Is it expected ?

bthirion (commented):
Thx, we're getting closer !

examples/03_connectivity/plot_compare_decomposition.py
KamalakerDadi: I have no clue. Maybe not. If you think this is not expected then I will try to fix this as well.

bthirion: I see that we use either var(X) or _sum_of_square(X), which may explain the discrepancy. Is there any reason for that ?

KamalakerDadi: No strong motivation for that. Recent push uses ```np.var``` and still there exists discrepancy. Maybe discrepancy in variance is because estimated on reduced data versus masked data?

KamalakerDadi: I just checked with n_components=1, EV scores looked same with some discrepancy in the decimals with ```fit_intercept=False```.

nilearn/decomposition/base.py
```diff
@@ -550,11 +557,12 @@ def explained_variance(X, components, per_component=True):
...
+            del res
         return np.maximum(0., 1. - res_var / full_var)
     else:
         lr = LinearRegression(fit_intercept=True)
         lr.fit(components.T, X.T)
-        res_var = X - lr.coef_.dot(components)
-        res_var **= 2
-        res_var = np.sum(res_var)
+        res = X - lr.coef_.dot(components)
+        res_var = np.var(res)
```
bthirion, line 567: For the sake of consistence, I would replace this by `_sum_of_squares(res).sum()`, and similarly replace full_var by `_sum_of_squares(X).sum()`
I guess this makes no numerical difference, but looks more consistent to me.
Do you agree ?
Best,

KamalakerDadi: Yes I agree and pushed the suggestion. The reason I was testing is because of your comment which is copy paster below. 

>I see that we use either var(X) or _sum_of_square(X), which may explain the discrepancy. Is there any reason for that ?

examples/03_connectivity/plot_compare_decomposition.py
KamalakerDadi: @bthirion  This remains still a question. What shall we do?

bthirion: I think that this is fine: Since components are not forced to be orthogonal, the  two score may indeed differ.
So, there's nothing more to do from my perspective.

nilearn/signal.py
```diff
@@ -120,6 +120,38 @@ def _mean_of_squares(signals, n_batches=20):
     return var
 
 
+def _sum_of_squares(signals, n_batches=20):
```
bthirion, line 123: Maybe a last suggestion: `_row_sum_of _squares` rather than `_sum_of _squares` better describes the function.

bthirion (approved):
Thx !

bthirion:
Merging, thx.

PR #2583, opened 2020-10-20
BUG: Fix for latest matplotlib

larsoner:
On latest matplotlib I get (this is from MNE but we vendor your fix):
```
  File "/home/larsoner/python/mne-python/mne/fixes.py", line 1104, in _crop_colorbar
    xy = cbar._outline(new_X, new_Y)
AttributeError: 'ColorbarBase' object has no attribute '_outline'
```
See inline comments for description.

Not sure if there is a better fix or not...

### DIFF
--- a/nilearn/conftest.py
+++ b/nilearn/conftest.py
@@ -15,6 +15,22 @@ except ImportError:
     collect_ignore = ['plotting',
                       'reporting',
                       ]
+    matplotlib = None
+
+
+def pytest_configure(config):
+    """Use Agg so that no figures pop up."""
+    if matplotlib is not None:
+        matplotlib.use('Agg', force=True)
+
+
+@pytest.fixture(autouse=True)
+def close_all():
+    """Close all matplotlib figures."""
+    yield
+    if matplotlib is not None:
+        import matplotlib.pyplot as plt
+        plt.close('all')  # takes < 1 us so just always do it
 
 
 def pytest_collection_modifyitems(items):
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -231,8 +231,7 @@ def _crop_colorbar(cbar, cbar_vmin, cbar_vmax):
     # _outline was removed in
     # https://github.com/matplotlib/matplotlib/commit/03a542e875eba091a027046d5ec652daa8be6863
     # so we use the code from there
-    if LooseVersion(matplotlib.__version__) >= LooseVersion("3.2.0") and \
-            hasattr(cbar, '_outline'):
+    if LooseVersion(matplotlib.__version__) >= LooseVersion("3.2.0"):
         cbar.ax.set_ylim(cbar_vmin, cbar_vmax)
         X, _ = cbar._mesh()
         X = np.array([X[0], X[-1]])
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -691,7 +691,7 @@ def _plot_roi_contours(display, roi_img, cmap, alpha, linewidths):
         if label == 0:
             continue
         data = (roi_data == label)
-        data = data.astype(np.int)
+        data = data.astype(int)
         img = new_img_like(roi_img, data, affine=roi_img.affine)
         display.add_contours(img, levels=[0.5], colors=[color_list[idx - 1]],
                              alpha=alpha, linewidths=linewidths,
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -228,12 +228,22 @@ def _crop_colorbar(cbar, cbar_vmin, cbar_vmax):
 
     # matplotlib >= 3.2.0 no longer normalizes axes between 0 and 1
     # See https://matplotlib.org/3.2.1/api/prev_api_changes/api_changes_3.2.0.html
-    if LooseVersion(matplotlib.__version__) >= LooseVersion("3.2.0"):
+    # _outline was removed in
+    # https://github.com/matplotlib/matplotlib/commit/03a542e875eba091a027046d5ec652daa8be6863
+    # so we use the code from there
+    if LooseVersion(matplotlib.__version__) >= LooseVersion("3.2.0") and \
+            hasattr(cbar, '_outline'):
         cbar.ax.set_ylim(cbar_vmin, cbar_vmax)
         X, _ = cbar._mesh()
-        new_X = np.array([X[0], X[-1]])
-        new_Y = np.array([[cbar_vmin, cbar_vmin], [cbar_vmax, cbar_vmax]])
-        xy = cbar._outline(new_X, new_Y)
+        X = np.array([X[0], X[-1]])
+        Y = np.array([[cbar_vmin, cbar_vmin], [cbar_vmax, cbar_vmax]])
+        N = X.shape[0]
+        ii = [0, 1, N - 2, N - 1, 2 * N - 1, 2 * N - 2, N + 1, N, 0]
+        x = X.T.reshape(-1)[ii]
+        y = Y.T.reshape(-1)[ii]
+        xy = (np.column_stack([y, x])
+              if cbar.orientation == 'horizontal' else
+              np.column_stack([x, y]))
         cbar.outline.set_xy(xy)
     else:
         cbar.ax.set_ylim(cbar.norm(cbar_vmin), cbar.norm(cbar_vmax))
@@ -1458,7 +1468,7 @@ optional
     -----
     The plotted image should in MNI space for this function to work properly.
 
-    This function is deprecated and will be removed in the 0.9.0 release. Use 
+    This function is deprecated and will be removed in the 0.9.0 release. Use
     plot_markers instead.
     """
     dep_msg = ("This function is deprecated and will be "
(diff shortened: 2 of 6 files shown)


larsoner:
Okay this is clearly not the correct fix:
```
  File "/home/larsoner/python/mne-python/mne/fixes.py", line 1118, in _crop_colorbar
    outline = cbar.outline.get_xy()
AttributeError: '_ColorbarSpine' object has no attribute 'get_xy'
```
I'm wondering if this fix is perhaps not necessary on newer matplotlib. I'll wrap this in a `hasattr` check for now. Any good examples I could run to see if it's working correctly or not?

larsoner:
... nope, just needed to remove the `hasattr` from the conditional I had errantly left in there.

Still interested to know if there is a good example to run to check this behavior -- maybe no adjustments are necessary anymore (or maybe this fix actually does not work!).

nilearn/plotting/img_plotting.py
```diff
@@ -228,12 +228,21 @@ def _crop_colorbar(cbar, cbar_vmin, cbar_vmax):
...
-        xy = cbar._outline(new_X, new_Y)
+        X = np.array([X[0], X[-1]])
+        Y = np.array([[cbar_vmin, cbar_vmin], [cbar_vmax, cbar_vmax]])
+        N = X.shape[0]
+        ii = [0, 1, N - 2, N - 1, 2 * N - 1, 2 * N - 2, N + 1, N, 0]
+        x = X.T.reshape(-1)[ii]
+        y = Y.T.reshape(-1)[ii]
+        xy = (np.column_stack([y, x])
+              if cbar.orientation == 'horizontal' else
+              np.column_stack([x, y]))
```
bthirion, line 245: Thx a lot ! 

All this looks good to me, but would you mind giving a few explanations ? I'm afraid of the maintenance of this code, give that some of the computations are somewhat cryptic ;-)

larsoner: The code is taken from the linked URL / diff above, i.e., it keeps whatever the old computations were

larsoner: (I do not know what it does actually)

bthirion:
LGTM. Any other comment/review ?

KamalakerDadi:
With matplotlib==3.3.2, I could not reproduce the issue. Is this issue anticipated with the upcoming release?

larsoner:
Yes I'm on 3.4-dev

larsoner:
> I wonder if we could add tests. A smoke test for plotting.plot_img with colorbar=True.

It seems as though this is unnecessary -- on `master` I get:
```
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plot_glass_brain - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plot_stat_map - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plot_stat_map_threshold_for_uint8 - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plot_glass_brain_threshold_for_uint8 - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_save_plot - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpdgw_nfrx.png'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plot_stat_map_colorbar_variations - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_singleton_ax_dim - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plot_stat_map_with_nans - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plotting_functions_with_cmaps - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plotting_functions_with_nans_in_bg_img - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plotting_functions_with_display_mode_tiled - AttributeError: 'ColorbarBase' object has no attribute '_outline'
FAILED nilearn/plotting/tests/test_img_plotting.py::test_plot_glass_brain_colorbar_having_nans - AttributeError: 'ColorbarBase' object has no attribute '_outline'
```
On this PR these pass.

I also:

1. Changed some `try/except/os.remove` patterns to use the `tmpdir` pytest fixture. It's both more standard for testing (nowadays) and nicer because it automatically cleans up.
2. Changed some `np.bool`, `np.int`, and `np.float` to their builtin equivalents because of warnings in latest `numpy` like:
    ```
    /home/larsoner/python/nilearn/nilearn/plotting/find_cuts.py:275: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. Use `float` by itself, which is identical in behavior, to silence this warning. If you specifically wanted the numpy scalar type, use `np.float_` here.
    ```
3. Added an `autouse=True` fixture at the `plotting/tests/conftest.py` scope so that plots always close -- at one point with the failures above there were dozens of windows open on my system

larsoner:
... as a side note, I'd also recommend adding this to `plotting/tests/conftest.py`:
```
def pytest_configure(config):
    import matplotlib
    matplotlib.use('Agg', force=True)
```
It makes it so that no plots pop up, which is really nice (avoids focus stealing on window managers that aren't great at it).

KamalakerDadi:
>On this PR these pass.

Alright. Makes sense.

KamalakerDadi:
>Added an autouse=True fixture at the plotting/tests/conftest.py scope so that plots always close -- at one point with the failures above there were dozens of windows open on my system

That's nice. But, we already have ```plt.close()``` in each test function in order to address this issue. 

- Do you think it's not effective?
- If not, may be we remove ```plt.close()``` and use ```close_all()```? WDYT

Slightly we are deviating from the main point.  Fixing ```_outline``` is good for now. Will there be an alternative once 3.4.0 matplotlib is released?

larsoner:
> That's nice. But, we already have plt.close() in each test function in order to address this issue.
>
> - Do you think it's not effective?

Yes -- these only get executed if the test passes (this is point (3) I made above). The `autouse=True` is a better solution to the problem:

1. It's always executed at the end regardless of the test passing or not. It's like wrapping the entire test in a `try/except: plt.close('all')`.
2. In principle we could now delete all of the `plt.close()`s at the end of your tests to make them more DRY, which is a good thing. I can do that if you want more `-` lines in this PR.

> - If not, may be we remove plt.close() and use close_all()? WDYT

Calling `plt.close('all')` when everything is already closed doesn't add any appreciable overhead:
```
In [1]: import matplotlib.pyplot as plt                                                                                                                                             
In [2]: %timeit plt.close('all')                                                                                                                                                    
516 ns ± 2.13 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
So there is no real *need* to do it in this PR. But I can do it if you want, it's easy enough.

> Will there be an alternative once 3.4.0 matplotlib is released?

I'm not sure what you're asking. Alternative to what? The code that is in this PR should work on 3.3 and 3.4. For me locally it works on 3.3, and your CIs at least are happy with it on < 3.4.

jeromedockes:
+1 for setting a non-gui matplotlib backend for tests, and for `plt.close("all")`  in an autoused fixture on my side.

KamalakerDadi:
:thumbsup:

larsoner:
> +1 for setting a non-gui matplotlib backend for tests, and for plt.close("all") in an autoused fixture on my side.

Done, and I moved these to the root `nilearn/conftest.py`. I figure a < 1 us overhead on each test run is acceptable. But if it's not, I can move the `plt.close('all')` autoused figure back into a new `nilearn/plotting/tests/conftest.py` or `nilearn/plotting/conftest.py` file

jeromedockes:
> Done, and I moved these to the root `nilearn/conftest.py`. I figure a < 1 us overhead on each test run is acceptable. But if it's not, I can move the `plt.close('all')` autoused figure back into a new `nilearn/plotting/tests/conftest.py` or `nilearn/plotting/conftest.py` file

Thanks! I think we can keep it in nilearn/conftest.py

nilearn/conftest.py
```diff
@@ -15,6 +15,22 @@
     collect_ignore = ['plotting',
                       'reporting',
                       ]
+    matplotlib = None
+
+
+def pytest_configure(config):
+    """Use Agg so that no figures pop up."""
+    if matplotlib is not None:
+        matplotlib.use('Agg', force=True)
```
KamalakerDadi, line 24: I hope we will not be having issues as in https://github.com/nilearn/nilearn/issues/2560

larsoner: The expectation for running tests should be different from the expectation when importing a library four actual use. In test running it's fine to set the backend. We've done it this way for years in MNE-Python without issue.

bthirion (approved):
Thx!

bthirion:
Merging, thx.

PR #2589, opened 2020-11-05
[MRG] Added glm-level warning

### LINKED ISSUE #2587: Mark nilearn.stats as experimental.

bthirion:
Closes #2587.

### DIFF
--- a/nilearn/glm/__init__.py
+++ b/nilearn/glm/__init__.py
@@ -47,3 +47,8 @@ __all__ = [
     'first_level',
     'second_level',
 ]
+
+from warnings import warn
+
+warn('This module is experimental. ' +
+     'It may change in any future release of Nilearn.', UserWarning)
--- a/nilearn/glm/__init__.py
+++ b/nilearn/glm/__init__.py
@@ -1,6 +1,7 @@
 """
 Analysing fMRI data using GLMs.
 """
+from warnings import warn
 from nilearn.glm.contrasts import (
     compute_contrast,
     compute_fixed_effects,
@@ -48,7 +49,5 @@ __all__ = [
     'second_level',
 ]
 
-from warnings import warn
-
-warn('This module is experimental. ' +
-     'It may change in any future release of Nilearn.', UserWarning)
+warn('The nilearn.glm module is experimental. '
+     'It may change in any future release of Nilearn.', FutureWarning)
--- a/nilearn/glm/__init__.py
+++ b/nilearn/glm/__init__.py
@@ -1,5 +1,8 @@
 """
 Analysing fMRI data using GLMs.
+
+Note that the nilearn.glm module is experimental.
+     It may change in any future (>0.7.0) release of Nilearn.
 """
 from warnings import warn
 from nilearn.glm.contrasts import (
--- a/nilearn/plotting/__init__.py
+++ b/nilearn/plotting/__init__.py
@@ -4,7 +4,7 @@ Plotting code for nilearn
 # Original Authors: Chris Filo Gorgolewski, Gael Varoquaux
 import os
 import sys
-
+import warnings
 
 ###############################################################################
 # Make sure that we don't get DISPLAY problems when running without X on
@@ -46,8 +46,10 @@ from .img_plotting import (
     plot_markers, plot_prob_atlas, plot_carpet, plot_img_comparison, show)
 from .find_cuts import find_xyz_cut_coords, find_cut_slices, \
     find_parcellation_cut_coords, find_probabilistic_atlas_cut_coords
-from .matrix_plotting import (plot_matrix, plot_contrast_matrix,
-                              plot_design_matrix, plot_event)
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    from .matrix_plotting import (plot_matrix, plot_contrast_matrix,
+                                  plot_design_matrix, plot_event)
 from .html_surface import view_surf, view_img_on_surf
 from .html_stat_map import view_img
 from .html_connectome import view_connectome, view_markers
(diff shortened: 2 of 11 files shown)


bthirion:
Questions: is it enough ? Should this be tested ?

nilearn/glm/__init__.py
```diff
@@ -47,3 +47,8 @@
     'first_level',
     'second_level',
 ]
+
+from warnings import warn
+
+warn('This module is experimental. ' +
```
GaelVaroquaux, line 53: It is probably useful to indicate which model you are talking about: in a cell with many module imports, the user will see the warning pop up not knowing which import triggered it.

nilearn/glm/__init__.py
```diff
@@ -47,3 +47,8 @@
...
+     'It may change in any future release of Nilearn.', UserWarning)
```
GaelVaroquaux, line 54: I would use a FutureWarning

GaelVaroquaux (changes requested):
I would think that we don't need additional test, as this is only temporary.

Looks fine to me, aside the two comments that I added.

bthirion:
Thx. I think we should also add something in the docs... Will do asap...

nilearn/glm/__init__.py
jeromedockes, line 53: nitpick: you don't need the `+`

jeromedockes:
I think other nilearn modules that import `glm` (`plotting` and `reporting`) should suppress the warning

nilearn/plotting/__init__.py
```diff
@@ -46,8 +46,10 @@ def _set_mpl_backend():
     plot_markers, plot_prob_atlas, plot_carpet, plot_img_comparison, show)
 from .find_cuts import find_xyz_cut_coords, find_cut_slices, \
     find_parcellation_cut_coords, find_probabilistic_atlas_cut_coords
-from .matrix_plotting import (plot_matrix, plot_contrast_matrix,
-                              plot_design_matrix, plot_event)
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
```
GaelVaroquaux, line 50: Can we be a bit more specific about which warnings we ignore, so as to avoid silencing legit warnings:
- Could we move these lines into the specific file (matrix_plotting)
- Could we can only FutureWarnings

GaelVaroquaux:
@bthirion : I wanted to merging this, but I think that the comment above by @jeromedockes is valid. Thanks!!

bthirion:
As far as I can see, i addressed it. Sorry if I miss something obvious.
For some reason, importing `glm.reporting.glm_reporter` does not raise the warning... not sure why....

jeromedockes:
> For some reason, importing `glm.reporting.glm_reporter` does not raise the warning... not sure why....

because once a module is imported it is cached -- the line raising the
warning is only seen once. if plotting is imported first the warning gets
suppressed then is not raised again. this is what happens in `reporting`
but would also happen if a user imported plotting before glm.
I guess it can be fixed by removing the glm module from the cache so it gets
imported again if a user imports it directly; something like
```
    warnings.simplefilter("ignore", FutureWarning)
    from nilearn.glm.contrasts import expression_to_contrast_vector
    from nilearn.glm.first_level import check_design_matrix
    import sys
    del sys.modules['nilearn']
    for k in list(sys.modules.keys()):
        if 'nilearn.glm' in k:
            del sys.modules[k]
```
but there may be a better solution

jeromedockes:
also note that as reporting imports glm_reporter in its __init__.py, we
cannot get the warning just for gml_reporter and not the rest of
reporting.
a warning could be raised from inside the make_glm_report_function.
but as it is unlikely that someone will use glm_reports without also
using directly glm to fit a model, I think it's ok to suppress the
warning in `reporting` (just as in plotting), and rely on the warning
being seen when glm is imported by user code.

jeromedockes:
also, unfortunately some users ignore warnings so I think a very visible
statement at the beginning of the glm classes' docstrings would also
help a lot

bthirion:
> because once a module is imported it is cached -- the line raising the warning is only seen once. if plotting is imported first the warning gets suppressed then is not raised again. this is what happens in `reporting` but would also happen if a user imported plotting before glm. I guess it can be fixed by removing the glm module from the cache so it gets imported again if a user imports it directly; something like

But, even without prior import of anything else, if simply import `glm_report`, I don't get the warning.

bthirion:
Sorry, I understand the point now.

bthirion:
@jeromedockes can you double check this PR ? We need to validate it as we want to proceed with the release tomorrow.

jeromedockes:
> @jeromedockes can you double check this PR ? We need to validate it as we want to proceed with the release tomorrow.
 sure, looking at it now!

jeromedockes:
I opened a PR on your branch that prevents importing reporting from silencing the warning, too. But messing with sys.modules seems to cause pickling errors I didn't get locally -- I guess it may be a bad idea. maybe we should give up on the warning and rely only on the changes you made to the doc? we could also import the glm functions from a temporary private module, but that may be a lot of work just to show a warning

jeromedockes:
sorry about the wasted time, if you want to remove the warning altogether I can do it.

bthirion:
I simply propose to simplify the behavior: have the warning for the glm, and accept that it is silenced from times to times due to the import history.

jeromedockes:
> I simply propose to simplify the behavior: have the warning for the glm, and accept that it is silenced from times to times due to the import history.

ok that makes sense! so keeping the same warning and simplefilter, just removing modifications of sys.modules

jeromedockes (approved):
thanks!

bthirion:
Merging,  thx to all !

Issue #2508, opened 2020-05-29
restore link to artifacts in github PR pages


jeromedockes:
was removed in #2507 as it was not working and causing trouble for some (at least one :) ) users.
see also https://github.com/nilearn/nilearn/pull/2477#issuecomment-635917872

GaelVaroquaux:
Examples of repo where this work: scikit-learn, MNE-python

agramfort:
I can help if you want

GaelVaroquaux:
@agramfort : thanks, I'll let @NicolasGensollen ping you if he gets to this.

NicolasGensollen:
I reopen this one since this is still not working (at least for me...). 
It worked first shot there: [Add github action artifact link redirector](https://github.com/dirty-cat/dirty_cat/pull/150)

The only difference I see is the path to the artifacts on CircleCI which is `0/doc/_build/html/index.html` here, and `0\doc\index.html` on dirty-cat since its `config.yml` file specifies a destination:

```yml
- store_artifacts: 
     path: doc/_build/html 
     destination: doc 
```

The path is also simplified in the same way in [MNE config.yml](https://github.com/mne-tools/mne-python/blob/master/.github/workflows/circle_artifacts.yml)

NicolasGensollen:
Looks like this is still not working... 
I must have screwed up somewhere in the artifact path. :sweat:

ymzayek:
I would be for removing this workflow since it hasn't worked for a while and doesn't seem worth the effort for me. I think the artifacts are not too difficult to find and we can always add better documentation about it in the CI readmes if we find that it becomes an issue

bthirion:
The artifacts are easy to find for people in the coredev people, I'm less sure for others.
I would suggest to discuss this with people from MNE or dirty-cat ?

Remi-Gau:
To be clear: 

are we talking about the artifacts that are on github?

![Screenshot from 2023-03-15 10-02-38](https://user-images.githubusercontent.com/6961185/225259622-345af301-6ad4-477d-8351-030508d7d6b7.png)


or those on circle ci?

![Screenshot from 2023-03-15 10-02-58](https://user-images.githubusercontent.com/6961185/225259652-18cf8113-aee3-4e53-b5fb-9933826f899c.png)

ymzayek:
[...]

PR #2596, opened 2020-11-12
restore github workflow

NicolasGensollen:
Addresses #2508. Basically only revert deletion in #2507 
Not sure this will work.


NicolasGensollen:
@agramfort you might have a better opinion of whether this is correct or not.

GaelVaroquaux:
Merging to see if it works

Issue #2594, opened 2020-11-12
Show the current doc build


bthirion:
The Built doc from latest master update should be made visible on 
nilearn.github.io/dev
on top of the latest release
See the mechanism used by sklearn/mne.

ymzayek:
This is already in place. Closing

PR #2605, opened 2020-11-13
[MRG] Support multiple doc versions

NicolasGensollen:
Addresses #2594 

This is still WIP but since I'm not 100% sure of what I'm doing, I thought I'd share to see if people have recommendations/concerns on the implementation so far...

**Objective Steps**

- [x] Build doc for `stable` and `dev` versions
- [x] Automate the build and deploy for `dev` doc through CircleCI
- [x] Add red danger banner on `dev` doc with a link to `stable` doc

**To be addressed in future PRs**

- Add redirect in index to the `stable` version at the root of Github pages 
- Maintain older doc versions
- Provide user-friendly way of changing the doc version (dropdown menu or something alike...)

### DIFF
--- a/doc/themes/nilearn/static/nature.css_t
+++ b/doc/themes/nilearn/static/nature.css_t
@@ -65,6 +65,11 @@ div.content {
 }
 
 
+.alert-danger {
+    background-color: #e74c3c;
+    border-color: #e74c3c;
+    color: #ffffff;
+}
 
 div.topic {
     background-color: #eee;
@@ -688,7 +693,7 @@ div.sphinxsidebar ul ul li {
 
 div.sphinxsidebarwrapper {
     width: 227px;
-    padding-top: .5em;
+    padding-top: 1em;
 }
 
 div.sphinxsidebarwrapper ul li p {
--- a/doc/themes/nilearn/static/nature.css_t
+++ b/doc/themes/nilearn/static/nature.css_t
@@ -69,6 +69,7 @@ div.content {
     background-color: #e74c3c;
     border-color: #e74c3c;
     color: #ffffff;
+    padding: 0.2em;
 }
 
 div.topic {
@@ -693,7 +694,7 @@ div.sphinxsidebar ul ul li {
 
 div.sphinxsidebarwrapper {
     width: 227px;
-    padding-top: 1em;
+    padding-top: 1.5em;
 }
 
 div.sphinxsidebarwrapper ul li p {
--- a/doc/themes/nilearn/static/nature.css_t
+++ b/doc/themes/nilearn/static/nature.css_t
@@ -64,6 +64,52 @@ div.content {
     padding-bottom: 10px;
 }
 
+/* Dropdown Button */
+.dropbtn {
+  background-color: #3498DB;
+  color: white;
+  padding: 10px;
+  heigth: max-content;
+  font-size: 16px;
+  border: none;
+  cursor: pointer;
+}
+
+/* Dropdown button on hover & focus */
+.dropbtn:hover, .dropbtn:focus {
+  background-color: #2980B9;
+}
+
+/* The container <div> - needed to position the dropdown content */
+.dropdown {
+  width: max-content;
+  height: 50px;
+  position: relative;
+  display: grid;
+}
+
+/* Dropdown Content (Hidden by Default) */
+.dropdown-content {
+  display: none;
+  background-color: #f1f1f1;
+  min-width: 160px;
+  box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
+  z-index: 1;
+}
+
+/* Links inside the dropdown */
+.dropdown-content a {
+  color: black;
+  padding: 2px 2px;
+  text-decoration: none;
+  display: block;
+}
+
+/* Change color of dropdown links on hover */
+.dropdown-content a:hover {background-color: #ddd}
+
+/* Show the dropdown menu (use JS to add this class to the .dropdown-content container when the user clicks on the dropdown button) */
+.show {display:grid;}
 
 .alert-danger {
     background-color: #e74c3c;
@@ -535,6 +581,7 @@ div.bodywrapper h1 {
 
 div.sphinxsidebar {
     margin: 0;
+    margin-top: 50px;
     padding: 0 15px 15px 0;
     min-width: 226px;
     width: 17%;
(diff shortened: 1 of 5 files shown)


GaelVaroquaux:
> Should we maintain older versions like MNE and Scikit-learn ?)

Let's not do this now.

> Automate the build and deploy for dev doc through CircleCI

This is the hardest step.

> Add redirect in index to the stable version at the root of Github pages

Yes; shouldn't be too hard.

> Provide user-friendly way of changing the doc version (dropdown menu or something alike...)

In a second PR, IMHO.

However, in the first PR, it would be useful to have:

* A red banner pointing out that this the dev version
* A robots.txt so that Google and the like to not index the dev site

bthirion:
Thx for digging into that. Let us know when it's ready for a review (note: you also want to rename it "MRG: XXX" when it becomes close to being merged).

NicolasGensollen:
I think this is getting close. I'm not really sure how to test this in a "safe way" though (especially the modifications to `make install`).
If someone has some advice on the modifications themselves, or on possible ways to test this, that'd be great! :+1: 

Notes:
- The link to stable in the red banner is not pointing to anything yet since there is no `stable` folder in `nilearn.github.io`
- Once the doc versions will be in seperate folders, we'll need to change the index at the root to redirect to `nilearn.github.io/stable`

bthirion:
I see nothing fishy but not 100% sure.
What is needed to build a stable version along the dev version ?

NicolasGensollen:
> I see nothing fishy but not 100% sure.
> What is needed to build a stable version along the dev version ?

The next time we make a release, someone will run the `make install` command, which builds the doc, overwrites `nilearn.github.io/stable` and `nilearn.github.io/dev` with it and push it to github. 
At this point both versions are the same. However, the `stable` folder won't be overwritten by CircleCI while the `dev` folder will be every time it makes a full build.

There is something we'll have to do though: modify the existing `index.html` at the root of `nilearn.github.io` to redirect to `nilearn.github.io/stable`. The code is simple but I'm not 100% how to push it there except by manually pushing the file to `nilearn.github.io` (which is considered bad practice). If someone knows a better way...

NicolasGensollen:
@emdupre do you think you could take a look at this? It would be great to merge this PR before the next release to start building stable and dev doc versions from that point on. 
I rebased it for clarity because it has been sitting here for a little while.
One point I'm little bit unsure about is how to redirect automatically to the stable doc once stable and dev docs will be in separate subfolders. I would be in favor of using a redirect page like MNE is doing [here](https://github.com/mne-tools/mne-tools.github.io/blob/main/index.html) but there might be other/better ways...

emdupre:
> One point I'm little bit unsure about is how to redirect automatically to the stable doc once stable and dev docs will be in separate subfolders. I would be in favor of using a redirect page like MNE is doing here but there might be other/better ways...

Yes, I think a redirection makes sense ! I've used that before [here](https://github.com/ohbm-aperture/ohbm-aperture.github.io) to re-direct a github-pages rendered site to a separate site. You could just have that content in a `gh-pages` branch and direct `nilearn.github.io` to render from there on the GitHub Pages settings of the repository. In the browser the redirect will still show the final url, which I think in our case will be `nilearn.github.io/stable`, but I don't imagine that will be a problem ?

emdupre:
Hm, I hadn't remembered that nilearn was being served through github pages directly, so I don't think you'll want to change the branch that the gh-pages environment is rendering on. Adding a top-level `index.html` as MNE does would make sense, then, since you're rendering the whole branch ! That's probably easiest. 

It should be straight-forward to navigate to the stable version of the docs, then, but do you have a preference for how folks will get to the dev docs ? I checked and `nilearn.dev` is available for a little less than 20 CAD annual :wink: On the MNE site you can choose a stable or dev version from the drop down, but I couldn't figure out how to access the dev docs from the [rendered PR](https://6740-1235740-gh.circle-artifacts.com/0/dev/index.html).

doc/themes/nilearn/static/nature.css_t
```diff
@@ -65,6 +65,11 @@ div.content {
 }
 
 
+.alert-danger {
+    background-color: #e74c3c;
+    border-color: #e74c3c;
+    color: #ffffff;
```
emdupre, line 71: Although the color is visually striking, the lack of padding makes it feel cramped on the page. I added just a bit and it looks better IMHO :

Screenshot_2021-02-23 Nilearn Statistical Analysis for NeuroImaging in Python — Machine learning for NeuroImaging(1)

```suggestion
    color: #ffffff;
    padding: 0.2em;
```

doc/themes/nilearn/static/nature.css_t
```diff
@@ -688,7 +693,7 @@ div.sphinxsidebar ul ul li {
 
 div.sphinxsidebarwrapper {
     width: 227px;
-    padding-top: .5em;
+    padding-top: 1em;
```
emdupre, line 696: To optionally accommodate additional padding on warning:

```suggestion
    padding-top: 1.5em;
```

NicolasGensollen:
Thanks for the reviews @emdupre ! :+1: 

>Adding a top-level index.html as MNE does would make sense, then, since you're rendering the whole branch ! That's probably easiest.

I agree. I think this is the easiest way.

>Although the color is visually striking, the lack of padding makes it feel cramped on the page.

I committed your suggestions, thanks!

>On the MNE site you can choose a stable or dev version from the drop down, but I couldn't figure out how to access the dev docs from the rendered PR.

You're right, for now you can go to stable from dev using the link in the red banner but there is no link to dev from stable. I'll look into this.

NicolasGensollen:
@emdupre I added a drop-down button to enable stable vs. dev version selection as visible [here](https://6772-1235740-gh.circle-artifacts.com/0/dev/index.html). There is clearly room for improvement (especially design-wise...), so let me know if you have ideas/suggestions.

button1

button2

bthirion:
Currently, when you click on the 'select documentation version' button in circleci, you get to a 404 error, as this point to a page that has been built yet. Is it the expected behavior ?

NicolasGensollen:
I think this PR is trying to do too much at once, and there are too many things that are difficult to test before merging and which could go wrong. 
I believe I should break it up in smaller contributions which will be much easier to review and verify once merged.

bthirion:
@NicolasGensollen we can clsoe that one, right ?

NicolasGensollen:
Absolutely !

Issue #2584, opened 2020-10-20
confounds are not effectively regressed with some NiftiMasker parameters


pbellec:
I am working with nilearn 0.6.2
I am specifying confounds to a masker,  like so:
``` python 
masker = NiftiMasker(standardize=True, mask_img=mask).fit()
tseries = masker.transform(img, confounds=conf)
``` 
I expect the confounds to be regressed from tseries, i.e. the covariance between `tseries` and `conf` to be approximately zero. 
I've tested whether that is the case systematically with different values for `standardize`, and with/without demeaning the `conf`.
See the test file appended at the end of this message (written for `pytest`). 

My conclusion is that confounds are not regressed effectively in two situations: (1) with `standardize=False` and no demeaning of the confounds; (2) with `standardize='psc'` and demeaning of the confounds. 

In my persective both are bugs, although if one understands what NiftiMasker does it's logical. 
The problem is that the same standardization is applied to the fMRI time series and the confounds. 
This means that if `standardize=False`, then confounds do not get demeaned, and the "regression" then fits the mean of the provided signals, instead of fitting the fluctuations of the confounds to the fMRI fluctuations. 

If `standardize='psc'` and the confounds are demeaned, then it's impossible to adjust the confounds to a percentage signal change, so nilearn sets the confounds to zero.... While really the weights of the confounds are irrelevant (they will get scaled through the confound regression anyway), so there is no reason to scale them to psc. The regression would be more meaningful if the confounds were demeaned instead.

The solution I think makes the most sense is to systematically demean the confounds prior to regression, instead of standardizing them using the same option as the fMRI time series. At least, there should be a separate parameter for the standardization of the confounds set to `zscore`, but I can't think of a use case where users would not want to demean. 

Happy to work on a PR, and/or incorporate these tests in the code base.
Let me know what you think.

``` python 
import numpy as np
from pandas import read_csv
from sklearn.preprocessing import StandardScaler
from nibabel import Nifti1Image
from nilearn.input_data import NiftiMasker

epsilon = 10e-10


def _simu_img():
    # Random confounds
    conf = 2 + np.random.randn(100, 6)

    # Random 4D volume
    vol = 100 + 10 * np.random.randn(5, 5, 2, 100)
    img = Nifti1Image(vol, np.eye(4))

    # create an nifti image with the data, and corresponding mask
    mask = Nifti1Image(np.ones([5, 5, 2]), np.eye(4))

    return img, mask, conf


[...]

PR #2609, opened 2020-11-16
[MRG] Systematic confounds standardization

NicolasGensollen:
Adresses #2584 .

- Add a parameter `standardize_confounds` to `NiftiMasker` to allow the user to specify whether or not to standardize confounds, and what strategy to use (see #2584 for more details).
- Update broken tests 

With these modifications, the tests proposed by @pbellec in #2584 seem to pass (on my laptop at least...).

Todo:

- [x] Add specific tests (we could maybe add directly the tests of #2584 somewhere?)
- [x] Update other masker objects if needed


bthirion (commented):
Thanks @NicolasGensollen 
Let me suggest that you add a test that fails with previous behavior and now passes.
Best,

NicolasGensollen:
I believe this is ready for review.

Note: Not really sure about where to put the new tests. I can put them somewhere else if needed.

nilearn/decomposition/base.py
```diff
@@ -255,6 +255,9 @@ class BaseDecomposition(BaseEstimator, CacheMixin, TransformerMixin):
         If standardize is True, the time-series are centered and normed:
         their mean is put to 0 and their variance to 1 in the time dimension.
 
+    standardize_confounds: boolean, optional
```
bthirion, line 258: Well, it is a Bool or a string...

nilearn/decomposition/base.py
```diff
@@ -323,8 +326,8 @@ class BaseDecomposition(BaseEstimator, CacheMixin, TransformerMixin):
     def __init__(self, n_components=20,
                  random_state=None,
                  mask=None, smoothing_fwhm=None,
-                 standardize=True, detrend=True,
-                 low_pass=None, high_pass=None, t_r=None,
+                 standardize=True, standardize_confounds=True,
```
bthirion, line 329: We should probably have uniofrm defaults, e.g. True everywhere.

nilearn/input_data/multi_nifti_masker.py
```diff
@@ -49,6 +49,16 @@ class MultiNiftiMasker(NiftiMasker, CacheMixin):
         to zero mean and scaled to unit variance.
         False : Do not standardize the data.
 
+    standardize_confounds: {'zscore', 'psc', True, False}, default is 'zscore'
```
bthirion, line 52: This seems to be the exact docstring for that bit of code.

nilearn/input_data/multi_nifti_masker.py
```diff
@@ -49,6 +49,16 @@ class MultiNiftiMasker(NiftiMasker, CacheMixin):
...
+        Strategy to standardize the confounds.
+        'zscore': the signal is z-scored. Timeseries are shifted
+        to zero mean and scaled to unit variance.
+        'psc':  Timeseries are shifted to zero mean value and scaled
```
bthirion, line 56: I think that 'psc' does not make sense for confounds. I'd rather not have it.

nilearn/tests/test_masking.py
```diff
@@ -30,6 +32,80 @@
...
+    assert(_confounds_regression(standardize_signal='psc',
+                                 standardize_confounds='psc') < 10. * eps)
+
+    # No standardization of signal but psc the confounds
+    assert(_confounds_regression(standardize_signal=False,
+                                 standardize_confounds='psc') < 10. * eps)
+
+    # Z-score signal and default for the confounds
+    assert(_confounds_regression(standardize_signal=True) < 10. * eps)
+
```
bthirion, line 108: But is one of these tests failing with the previous verison of the code ?

bthirion (commented):
Thx ! I left a couple of comments.

NicolasGensollen:
Thanks for the comments @bthirion. 

Since `pcs` does not make much sense for `standardize_confounds`, I think that there is no need for this parameter to behave like `standardize`, and having it as a simple boolean should be enough. I changed it to be an optional boolean parameter defaulting to `True`, meaning that confounds are z-scored by default, as suggested in #2584.

In addition, it seems like the default values for `standardize` are not always consistent between docstrings and code (see #2612). We might want to address this in a separate PR.

For the tests, since I changed the API, they won't run with previous versions of Nilearn as is. Previous to this PR, `standardize` was used for both signals and confounds standardization (which can now be reproduced by setting `standardize_confounds` according to `standardize`). 

The test code from #2584 with previous versions of nilearn displays 2 failures:

- one corresponding to the case where `standardize=False` and no demeaning of confounds 
- one corresponding to the case where `standardize='psc'` and demaning of confounds

My understanding is that, one of the reasons the first test fails, is because `signal.clean` still divides confounds by the max value when standardize is `False` as can be seen here:

https://github.com/nilearn/nilearn/blob/82f4075d8a8ea9aec25e66bd87ebb79a6be6d32f/nilearn/signal.py#L612-L621

while the second test fails because confounds are standardized with `psc` (since the same parameter is used for signal and confounds) while the code from #2584 does a z-score instead.

If you run the same tests with the fix of this PR, they should all pass.

Let me know if this is not clear and if we should have more tests for this.

bthirion (approved):
LGTM, thx !

bthirion:
@pbellec Any feedback on this ?

NicolasGensollen:
Does someone else want to take a look, or should consider this as ready for merging? Thx!

bthirion:
Merging, thx !

PR #2611, opened 2020-11-17
Remove the check URL step from github workflow

### LINKED ISSUE #2508: restore link to artifacts in github PR pages

NicolasGensollen:
Closes #2508.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,3 @@ jobs:
                   repo-token: ${{ secrets.GITHUB_TOKEN  }}
                   artifact-path: 0/doc/_build/html/index.html
                   circleci-jobs: quick-build,full-build
-            - name: Check the URL
-              run: |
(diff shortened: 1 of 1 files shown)


bthirion:
Seems at least inoffensive, but does it have the intended effect ?

emdupre:
I think we'd need to merge this in to have it successfully show as a check in the CI.

It better matches [MNE's configuration](https://github.com/mne-tools/mne-python/blob/master/.github/workflows/main.yml), though, so I'm optimistic !

GaelVaroquaux:
OK, let's try this! Thanks a lot to everyone for the investigation work.

@phaide you may want to check this to see if you can adapt it for dirty_cat

PR #2622, opened 2020-11-30
Dummy option in Decoder and DecoderRegressor

### LINKED ISSUE #2471: Add "dummy" option as an estimator in the Decoder / DecoderRegressor
1 - Accept "dummy" with smart default in the Decoder / DecoderRegressor to have an API consistent way to test chance for the decoding problem at hand

2 - Add in section 2.1 of the userguide an Introduction and example plot_intro_decoding a mention of this and a comparaison of results to chance in the haxby case

KamalakerDadi:
Closes #2471  .

- Implementation of dummy option for classification and regression
- Tests
- Examples

### DIFF
--- a/nilearn/decoding/decoder.py
+++ b/nilearn/decoding/decoder.py
@@ -765,10 +765,6 @@ class _BaseDecoder(LinearRegression, CacheMixin):
                     proba.append(out)
                 scores = np.vstack([proba[k].argmax(axis=1) for
                                     k in range(self.n_outputs_)]).T
-            else:
-                message = ('Implemented only for strategies "most_frequent",'
-                           '"prior", "stratified"')
-                raise NotImplementedError(message)
         elif isinstance(self.estimator, DummyRegressor):
                 scores = np.full((n_samples, self.n_outputs_), self.dummy_output_,
                                   dtype=np.array(self.dummy_output_).dtype)
--- a/nilearn/decoding/decoder.py
+++ b/nilearn/decoding/decoder.py
@@ -95,8 +95,7 @@ def _check_param_grid(estimator, X, y, param_grid=None):
         or have no effect. See scikit-learn documentation for more information.
 
         For Dummy estimators, parameter grid defaults to empty as these
-        estimators are unreal there exists no interesting parameters to grid
-        search.
+        estimators do not have hyperparameters to grid search.
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


nilearn/decoding/decoder.py
```diff
@@ -89,10 +94,14 @@ def _check_param_grid(estimator, X, y, param_grid=None):
         useful to avoid exploring parameter combinations that make no sense
         or have no effect. See scikit-learn documentation for more information.
 
+        For Dummy estimators, parameter grid defaults to empty as these
+        estimators are unreal there exists no interesting parameters to grid
```
bthirion, line 98:
```suggestion
        estimators do not have hyperparameters to grid
```

bthirion (commented):
LGTM, thx ! 
Only a couple of lines are not covered ...

bthirion (commented):
LGTM. Add a whatnew entry ? Thx

thomasbazeille:
Sorry to come in late on this, I had a quick look and it seems to make sense, the only thing i'm worried about is that this makes a lot of code for a "side" feature, which is distributed in many functions and may be hard to maintain. I'll have a look to see if I can propose a few redesign to make it lighter.

nilearn/decoding/decoder.py
```diff
@@ -666,6 +745,35 @@ def _binarize_y(self, y):
...
+                rs = check_random_state(0)
+                proba = []
+                for k in range(self.n_outputs_):
+                    out = rs.multinomial(1, dummy_output[k], size=n_samples)
+                    out = out.astype(np.float64)
+                    proba.append(out)
+                scores = np.vstack([proba[k].argmax(axis=1) for
+                                    k in range(self.n_outputs_)]).T
+            else:
+                message = ('Implemented only for strategies "most_frequent",'
```
thomasbazeille, line 769: One detail  : Codecov reports that this paragraph is not tested, whereas there is a test on this edgecase. But this seems to be a duplicate of the test at new line 119. If you feel both are useful please test this one as well, otherwise I guess the current test 769  would be sufficient since it would be executed independently of the API level used so the other could be removed I guess

thomasbazeille (approved):

nilearn/decoding/decoder.py
KamalakerDadi: Good catch.

bthirion (approved):
LGTM, thx !

Issue #1511, opened 2017-09-14
Add affiliation to author list


GaelVaroquaux:
And update it.

tsalo:
What about linking to folks' `meta` repositories (or other self-maintained location for affiliation info) instead?

bthirion:
Sorry, I'm not familiar with that. Can you describe more precisely what `meta` repository you are thiking of ?

tsalo:
I'm referring to a dedicated personal repository for affiliation and ID information. Here's mine: https://github.com/tsalo/meta

That way, we wouldn't have to manually update affiliations.

bthirion:
OK, I don't think it is easy to systematize this. 

My feeling is that maintaining this information is  a task for the core team :-)

bthirion:
I'm not going to do it. Sorry. If any one wants to jump in, you're welcome.

Remi-Gau:
Will open a draft PR so we can see judge whether we want to go down the CITATION.CFF route

PR #2632, opened 2020-12-08
move list of contributors from readme to authors.rst

### LINKED ISSUE #2601: list of developers in readme is outdated
also it would be easier to have only one such list (eg make sure authors.rst is complete and remove the list from readme)

jeromedockes:
Closes #2601  .

I have moved names listed in the README but not AUTHORS in a section "other contributors"; WDYT?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -33,6 +33,21 @@ The nilearn core developers are:
 * Thomas Bazeille
 * Taylor Salo
 
+Other contributors
+..................
+
+Some other past or present contributors are:
+
+* A. Abraham
+* B. Cipollini.
+* C. F. Gorgolewski
+* D. Bzdok
+* F. Pedregosa
+* L. Esteve
+* M. Eickenberg
+* P. Gervais
+* V. Michel
+
 Funding
 .......
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -38,15 +38,15 @@ Other contributors
 
 Some other past or present contributors are:
 
-* A. Abraham
-* B. Cipollini.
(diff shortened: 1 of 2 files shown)


NicolasGensollen (changes requested):
I think the list of other contributors is not rendered properly in the doc: [link](https://6361-1235740-gh.circle-artifacts.com/0/doc/_build/html/authors.html). 
I suppose that the name's first letter got interpreted as a list item marker.

jeromedockes:
> I think the list of other contributors is not rendered properly in the doc: [link](https://6361-1235740-gh.circle-artifacts.com/0/doc/_build/html/authors.html). 
> I suppose that the name's first letter got interpreted as a list item marker.

Thanks!

NicolasGensollen (approved):
LGTM, thanks!

_Note: If we want to include contributors explicitely in the README, we could use something like [this](https://github.com/all-contributors/all-contributors). Not sure this is what we want but it's worth mentioning it._

bthirion:
See #1511

jeromedockes:
> Note: If we want to include contributors explicitely in the README, we could use something like this. Not sure this is what we want but it's worth mentioning it.

that can be a good option. IMHO all the relevant information re authors is easily available in the git log and github issues and pr discussions so I don't really know what information we want explicitly in the readme and authors.rst ; the goal of this pr was just to have all in one place without changing the content

bthirion (approved):

bthirion:
Indeed, we can merge that one, and see later for better ways to link to individual infrmation

NicolasGensollen:
Merging... Thanks!

Issue #2627, opened 2020-12-04
Move tests from Travis to GitHub Actions and/or Azure Pipelines


NicolasGensollen:
Following discussion started in #2625 

I think that a possible first step in this direction would be to move one travis job to GitHub actions (similar to what was done for MNE a few weeks ago), and move another one to Azure Pipelines.
In the mean time, this could be the opportunity to check whether we need all current jobs, and how we want to distribute them between the available services.

bthirion:
We might discuss that in the upcoming coredev meeting.

Just to be sure: why not simply ask Travis for some resources ?

GaelVaroquaux:
> Just to be sure: why not simply ask Travis for some resources ?

That plan failed for scikit-learn (we were not granted resources), so it
might fail for nilearn.

NicolasGensollen:
Closing since this has been addressed by #2633

Issue #2608, opened 2020-11-16
test_clean_confounds fails on x86_64


ggardet:
What version of Nilearn are you using? 0.7.0
What were you trying to do? Build and test 0.7.0 version
What did you expect will happen? all test pass
What actually happened? `test_clean_confounds ` failed on x86_64

Log:
```
[  882s] =================================== FAILURES ===================================
[  882s] _____________________________ test_clean_confounds _____________________________
[  882s] [gw0] linux -- Python 3.8.6 /usr/bin/python3
[  882s] 
[  882s]     def test_clean_confounds():
[  882s]         signals, noises, confounds = generate_signals(n_features=41,
[  882s]                                                       n_confounds=5, length=45)
[  882s]         # No signal: output must be zero.
[  882s]         eps = np.finfo(np.float).eps
[  882s]         noises1 = noises.copy()
[  882s]         cleaned_signals = nisignal.clean(noises, confounds=confounds,
[  882s]                                          detrend=True, standardize=False)
[  882s]         assert abs(cleaned_signals).max() < 100. * eps
[  882s]         # clean should not modify inputs
[  882s]         assert np.array_equal(noises, noises1)
[  882s]     
[  882s]         # With signal: output must be orthogonal to confounds
[  882s]         cleaned_signals = nisignal.clean(signals + noises, confounds=confounds,
[  882s]                                          detrend=False, standardize=True)
[  882s]         assert abs(np.dot(confounds.T, cleaned_signals)).max() < 1000. * eps
[  882s]     
[  882s]         # Same output when a constant confound is added
[  882s]         confounds1 = np.hstack((np.ones((45, 1)), confounds))
[  882s]         cleaned_signals1 = nisignal.clean(signals + noises, confounds=confounds1,
[  882s]                                           detrend=False, standardize=True)
[  882s]         np.testing.assert_almost_equal(cleaned_signals1, cleaned_signals)
[  882s]     
[  882s]         # Test detrending. No trend should exist in the output.
[  882s]         # Use confounds with a trend.
[  882s]         temp = confounds.T
[  882s]         temp += np.arange(confounds.shape[0])
[  882s]     
[  882s]         cleaned_signals = nisignal.clean(signals + noises, confounds=confounds,
[  882s]                                          detrend=False, standardize=False)
[  882s]         coeffs = np.polyfit(np.arange(cleaned_signals.shape[0]),
[  882s]                             cleaned_signals, 1)
[...]

PR #2633, opened 2020-12-09
[MRG] Move tests to Github Actions

NicolasGensollen:
Move Travis tests to Github actions.
See #2627


KamalakerDadi:
Can you rebase/merge on top of master? Or is it already the case? Because there were some PRs that were merged without Travis.

NicolasGensollen:
>Can you rebase/merge on top of master? Or is it already the case? Because there were some PRs that were merged without Travis.

Done, thx!

I think this is ready for review. The actions test the following:

- Python `3.5` and minimum requirements on ubuntu
- Python `3.6`, `3.7`, and `3.8` on ubuntu, mac OS, and windows with latest package versions

Note: I had to modify (very slightly...) an assertion in a test to make it pass for Mac OS (see 0c8ed03).

jeromedockes:
great, thanks! a couple of questions:
- shoudn't we remove `.travis.yml`?
- are the tests on windows in azure now redundant?

NicolasGensollen:
>shoudn't we remove .travis.yml?

We could. I'm still waiting for the answer from the Travis Support people concerning my additional credit request. 
If they don't grant us enough credits, I was planning to use Travis only for testing on ARM architecture since this is a travis-specific feature for now.

>are the tests on windows in azure now redundant?

Yes (I dont think they run on the same version though, I'll have to double check...). I can remove windows tests from the github actions if this is too redundant.

NicolasGensollen:
I just realized that the assertion that got broken on MacOS (see 0c8ed03) is also responsible for the test failure of #2608 
I'm not sure how the `150. * eps` threshold was chosen, and whether we can increase it up to `200. * eps` like I did here, but this might also solve #2608.

bthirion:
Thx ! The changes look good.
I  think we should avoid redundancies btw azure and  Github-actions.

bthirion:
There seems to be aproblem on some Python versions ?

NicolasGensollen:
> There seems to be aproblem on some Python versions ?

Yes, I saw that yesterday. I think it was related to the very recent version bump of `importlib_metadata` from `3.1.1` to `3.3.0` which caused errors for Python versions `3.6` and `3.7` in the github actions workflows.  

The bug has been fixed, I re-ran all actions this morning and they all pass.

bthirion (approved):

bthirion:
I'd be happy to get another review for that one.

jeromedockes:
> I'd be happy to get another review for that one.
sure!

tools/github_actions_dependencies.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -ef
```
jeromedockes, line 1: why do we need '-f'?

tools/github_actions_install.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -ef
```
jeromedockes, line 1: why '-f' ?

tools/github_actions_test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -ef
```
jeromedockes, line 1: why '-e' and '-f' ?
also we could use '-x' to avoid the need for the 'echo' -- and the printed command might get out of sync with what is actually executed below

tools/github_actions_install.sh
```diff
@@ -0,0 +1,4 @@
...
+
+python setup.py build
+python setup.py install
```
jeromedockes, line 4: I wonder if we should have a job that checks that installation with pip runs smoothly?

tools/github_actions_test.sh
```diff
@@ -0,0 +1,4 @@
...
+
+echo 'python -m pytest --pyargs nilearn --cov=nilearn'
+python -m pytest --pyargs nilearn --cov=nilearn
```
jeromedockes, line 4: I think it would best not to run the tests from the directory in which the repo is checked out, but  from an empty (temporary) directory

jeromedockes:
linking to the issue you opened on setup-python: https://github.com/actions/setup-python/issues/179

NicolasGensollen:
@jeromedockes Thanks for the thorough review! :+1: 
Tests are now passing on MacOS (see the fix proposed in the issue you linked above). The only action that fails now is the `matplotlib-dev` tests which is allowed despite the disappointing red cross...
Feel free to let me know if something doesn't seem good enough for merging.

jeromedockes:
great! indeed at the moment github actions doesn't seem to have an equivalent of `allow_failures`: https://github.com/actions/toolkit/issues/399
so I think for now I would remove the matplotlib-dev job (sorry I didn't realize this before, I'm new to github actions), and in a future PR we can see if we reintroduce it, either on github actions or the azure build, in a way that doesn't cause the overall build to fail.
I think my comment on running the tests in a separate directory is still relevant, but this PR is already a huge improvement over not running the tests at all on linux (which is the current state of the CI), so I would suggest merging this PR now and then addressing this issue in  a separate one.
finally, I think I would remove the unused `.travis.yml`, and any scripts that were used exclusively for the travis build, either in this PR or another one merged just after. Indeed, CI can seem a bit complex and I think having leftover files from builds that don't run anymore could add to confusion of maintainers or contributors

NicolasGensollen:
@jeromedockes sounds good!

jeromedockes:
great, thanks a lot @NicolasGensollen !

Issue #956, opened 2016-01-19
Data returned by _safe_get_data is copied.


AlexandreAbraham:
In the code, the data returned by `_safe_get_data` is sometimes copied. I expected `_safe_get_data` to return a copy of the data. We should check that this copy is necessary.

banilo:
> We should check that this copy is necessary.

+1

NicolasGensollen:
Indeed, `_safe_get_data`  returns a copy of the data only when the `img` argument has a `None` `_data_cache` argument:

https://github.com/nilearn/nilearn/blob/d1a2563cbf10c1470835b04bbc6e7d627d71019a/nilearn/_utils/niimg.py#L44-L52

When this isn't the case, `_safe_get_data` behaves likes `_get_data` (with NaNs handling).

I see several options:

1- do nothing and keep this behavior
2- implement a consistent behavior: it should either always return a copy of the data, or never. 
3- add a `copy` parameter (defaulting to `False`) to either always copy the data or never

Since other parts of the code explicitely assume that `_safe_get_data` does not copy the data, like the `threshold_img` function:

https://github.com/nilearn/nilearn/blob/ddc59b147eecc6b8d7619edd9a1eba7f6487fec5/nilearn/image/image.py#L818-L822

I wouldn't go for a solution 2 with always copying.

WDYT?

bthirion:
My impression is that you made the right choice.

If nobody comes with another unforeseen use case, I'll be happy with #2634.

NicolasGensollen:
Closing since #2634 was merged...

PR #2634, opened 2020-12-11
Make _get_safe_data consistent

NicolasGensollen:
The function `_safe_get_data` makes a copy of the data in some particular case, see #956 
This is a possible solution that makes the data copying behaviour explicit.

Note: If the current behaviour is preferred, I can close both this PR and #956.

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -23,7 +23,7 @@ def _get_data(img):
     return data
 
 
-def _safe_get_data(img, ensure_finite=False):
+def _safe_get_data(img, ensure_finite=False, copy_data=False):
     """ Get the data in the image without having a side effect on the
         Nifti1Image object
 
@@ -36,15 +36,17 @@ def _safe_get_data(img, ensure_finite=False):
         If True, non-finite values such as (NaNs and infs) found in the
         image will be replaced by zeros.
 
+    copy_data: bool, default is False
+        If true, the returned data is a copy of the img data.
+
     Returns
     -------
     data: numpy array
         nilearn.image.get_data return from Nifti image.
     """
-    if hasattr(img, '_data_cache') and img._data_cache is None:
-        # By loading directly dataobj, we prevent caching if the data is
-        # memmaped. Preventing this side-effect can save memory in some cases.
+    if copy_data:
         img = copy.deepcopy(img)
+
     # typically the line below can double memory usage
     # that's why we invoke a forced call to the garbage collector
     gc.collect()
@@ -154,7 +156,7 @@ def copy_img(img):
 
     if not isinstance(img, nibabel.spatialimages.SpatialImage):
         raise ValueError("Input value is not an image")
-    return new_img_like(img, _safe_get_data(img).copy(), img.affine.copy(),
+    return new_img_like(img, _safe_get_data(img, copy_data=True), img.affine.copy(),
                         copy_header=True)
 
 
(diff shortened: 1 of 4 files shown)


nilearn/_utils/niimg.py
```diff
@@ -154,7 +156,7 @@ def copy_img(img):
 
     if not isinstance(img, nibabel.spatialimages.SpatialImage):
         raise ValueError("Input value is not an image")
-    return new_img_like(img, _safe_get_data(img).copy(), img.affine.copy(),
+    return new_img_like(img, _safe_get_data(img, copy_data=True), img.affine.copy(),
```
bthirion, line 159: Is this the only case where the data was copied ?

nilearn/_utils/niimg.py
```diff
@@ -36,15 +36,17 @@ def _safe_get_data(img, ensure_finite=False):
...
     -------
     data: numpy array
         nilearn.image.get_data return from Nifti image.
     """
-    if hasattr(img, '_data_cache') and img._data_cache is None:
-        # By loading directly dataobj, we prevent caching if the data is
-        # memmaped. Preventing this side-effect can save memory in some cases.
+    if copy_data:
         img = copy.deepcopy(img)
+
```
jeromedockes, line 49: if the image did not have a `_data_cache`, the returned array is the (freshly filled in the call to `_get_data`) cache of the image itself, rather than a copy?

NicolasGensollen: @jeromedockes Yes, this was the behavior AFAICT. 

Here is an example script:

```python
import sys
import numpy as np
from nilearn import datasets
from nilearn._utils.niimg import _safe_get_data
from nilearn._utils.niimg_conversions import check_niimg

atlas_data = datasets.fetch_atlas_msdl() # Get some data
data = check_niimg(atlas_data.maps)
print("Ref count on data._data_cache = {0}".format(sys.getrefcount(data._data_cache)))
print("Initial sum = {0}".format(np.sum(data._data_cache)))
data2 = _safe_get_data(data)
data2[0,0,0,0] += 1
print("Ref count on data._data_cache = {0}".format(sys.getrefcount(data._data_cache)))
print("Sum for data._data_cache = {0}".format(np.sum(data._data_cache)))
print("Ref count on data2 = {0}".format(sys.getrefcount(data2)))
print("Sum for data2 = {0}".format(np.sum(data2)))
data._data_cache = None
data3 = _safe_get_data(data)
data3[0,0,0,0] += 2
print("Ref count on data3 = {0}".format(sys.getrefcount(data3)))
print("Sum for data3 = {0}".format(np.sum(data3)))
print("Sum for data2= {0}".format(np.sum(data2)))
```
Output:
```
Ref count on data._data_cache = 2
Initial sum = 3566.948371222825
Ref count on data._data_cache = 3
Sum for data._data_cache = 3567.948371222825
Ref count on data2 = 3
Sum for data2 = 3567.948371222825
Ref count on data3 = 2
Sum for data3 = 3568.948371222825
Sum for data2= 3567.948371222825
```

nilearn/_utils/niimg.py
NicolasGensollen: It was the only call that broke some tests and caught my attention. 
You're right though, there are other calls that make an explicit copy of the data. I updated the code to handle them in c068665.

bthirion (approved):
LGTM, thx !

NicolasGensollen:
Thanks for the reviews! Merging...

PR #2636, opened 2020-12-14
[WIP] Add verbosity and progress to mass-univariate

### LINKED ISSUE #184: Add verbosity and progress in mass_univariate
The mass univariate example doesn't display its progress. It should use a progress indicator like the searchlight (I believe that part of this code should be factored out, @VirgileFritsch  and @AlexandreAbraham : you should discuss this together).

NicolasGensollen:
Fixes #184

### DIFF
--- a/nilearn/mass_univariate/permuted_least_squares.py
+++ b/nilearn/mass_univariate/permuted_least_squares.py
@@ -5,11 +5,12 @@ Massively Univariate Linear Model estimated with OLS and permutation test.
 #         Virgile Fritsch, <virgile.fritsch@inria.fr>, jan. 2014
 import warnings
 
+import sys
+import time
 import joblib
 import numpy as np
 from scipy import linalg
 from sklearn.utils import check_random_state
-
 
 def normalize_matrix_on_axis(m, axis=0):
     """ Normalize a 2D matrix on an axis.
@@ -138,10 +139,10 @@ def _t_score_with_covars_and_normalized_design(tested_vars, target_vars,
     return beta_targetvars_testedvars * np.sqrt((dof - 1.) / rss)
 
 
-def _permuted_ols_on_chunk(scores_original_data, tested_vars, target_vars,
-                           confounding_vars=None, n_perm_chunk=10000,
+def _permuted_ols_on_chunk(scores_original_data, tested_vars, target_vars, thread_id,
+                           confounding_vars=None, n_perm=10000, n_perm_chunk=10000,
                            intercept_test=True, two_sided_test=True,
-                           random_state=None):
+                           random_state=None, verbose=0):
     """Massively univariate group analysis with permuted OLS on a data chunk.
 
     To be used in a parallel computing context.
@@ -157,8 +158,15 @@ def _permuted_ols_on_chunk(scores_original_data, tested_vars, target_vars,
     target_vars : array-like, shape=(n_samples, n_targets)
       fMRI data. F-ordered for efficient computations.
 
+    thread_id : int
+        process id, used for display.
+
     confounding_vars : array-like, shape=(n_samples, n_covars)
       Clinical data (covariates).
+
+    n_perm : int,
+      Tomtal number of permutations to perform, only used for
+      display in this function.
 
     n_perm_chunk : int,
       Number of permutations to be performed.
@@ -176,6 +184,9 @@ def _permuted_ols_on_chunk(scores_original_data, tested_vars, target_vars,
     random_state : int or None,
       Seed for random number generator, to have the same permutations
       in each computing units.
+
+    verbose: int, optional, default is 0.
+      Defines the verbosity level.
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -157,9 +158,16 @@ def _permuted_ols_on_chunk(scores_original_data, tested_vars, target_vars,
...
       fMRI data. F-ordered for efficient computations.
 
+    thread_id : int
+        process id, used for display.
+
     confounding_vars : array-like, shape=(n_samples, n_covars)
       Clinical data (covariates).
 
+    n_perm : int,
+      Tomtal number of permutations to perform, only used for
```
bthirion, line 168:
```suggestion
      Total number of permutations to perform, only used for
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -157,9 +158,16 @@ def _permuted_ols_on_chunk(scores_original_data, tested_vars, target_vars,
     target_vars : array-like, shape=(n_samples, n_targets)
       fMRI data. F-ordered for efficient computations.
 
+    thread_id : int
+        process id, used for display.
```
bthirion, line 162: But if we have multiple threads, the display will become a bit  messy ?

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -446,16 +475,18 @@ def permuted_ols(tested_vars, target_vars, confounding_vars=None,
...
-            n_perm_chunk=n_perm_chunk, intercept_test=intercept_test,
-            two_sided_test=two_sided_test,
-            random_state=rng.randint(1, np.iinfo(np.int32).max - 1))
-        for n_perm_chunk in n_perm_chunks)
+            targetvars_resid_covars.T, thread_id + 1, covars_orthonormalized,
+            n_perm=n_perm, n_perm_chunk=n_perm_chunk,
+            intercept_test=intercept_test, two_sided_test=two_sided_test,
+            random_state=rng.randint(1, np.iinfo(np.int32).max - 1),
+            verbose=verbose)
+        for thread_id,n_perm_chunk in enumerate(n_perm_chunks))
```
bthirion, line 489:
```suggestion
        for thread_id, n_perm_chunk in enumerate(n_perm_chunks))
```

bthirion (commented):
Minor things. I'm realizing that the display may be messy with multiple threads ?

nilearn/mass_univariate/permuted_least_squares.py
NicolasGensollen: With multiple threads, the output looks like (on the console in a Jupyter notebook):

```
...
Job #1, processed 2500/5000 permutations (50.00%, 50 seconds remaining)
Job #2, processed 2530/5000 permutations (50.60%, 49 seconds remaining)
Job #1, processed 2510/5000 permutations (50.20%, 50 seconds remaining)
Job #2, processed 2540/5000 permutations (50.80%, 49 seconds remaining)
Job #1, processed 2520/5000 permutations (50.40%, 50 seconds remaining)
Job #1, processed 2530/5000 permutations (50.60%, 49 seconds remaining)
Job #2, processed 2550/5000 permutations (51.00%, 49 seconds remaining)
Job #1, processed 2540/5000 permutations (50.80%, 49 seconds remaining)
...
```

which is not the best I agree, but I'm not sure we can do much better with simple prints. 

One way to get something cleaner is to use `tqdm`, but I don't think it's worth adding a dependency for this... Here is an example:

```python
from time import sleep
from tqdm import tqdm, trange
from joblib import Parallel, delayed

L = list(range(9))

def func(n, total):
    interval = 0.001 / (n + 2)
    i = 0
    result = 0
    text = "#{}, est. {:<04.2}s".format(n, interval * total)
    for i in trange(total, desc=text, position=n):
        result += i
        sleep(interval)
    return result

Parallel(n_jobs=4)(delayed(func)(l,10000) for l in L)
```

bthirion: Thx ! 
That's good enough I think. No need to add a dependency on  tqdm.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -229,6 +241,23 @@ def _permuted_ols_on_chunk(scores_original_data, tested_vars, target_vars,
         scores_as_ranks_part += (h0_fmax_part[i].reshape((-1, 1))
                                  < scores_original_data.T)
 
+        if verbose > 0:
+            step = 11 - min(verbose, 10)
```
bthirion, line 245: This is almost funny.

bthirion (commented):
Any possibility to test that ... ?

NicolasGensollen:
> Any possibility to test that ... ?

Indeed, we can test that nothing crashes and improve coverage. 
See f09017c. (I also did it for searchlight since these lines weren't executed by any test either).

nilearn/mass_univariate/permuted_least_squares.py
NicolasGensollen: This is how it is implemented in searchlight. :) 
I can change that if needed...

bthirion (approved):
Good for me. Thx !

NicolasGensollen:
Thanks! Merging...

PR #2638, opened 2020-12-15
FIX: Remove missing name

larsoner:
`np.version` no longer has `short_version` in NumPy `master`, but it's also not needed because `nilearn` requires 1.11 nowadays anyway and it was only used in a check for 1.4.


bthirion (approved):

bthirion:
LGTM, thx ! 
Merging.

PR #2640, opened 2020-12-17
[ENH] Harmonize short repr of niimgs

### LINKED ISSUE #747: Harmonise usage of short repr for niimgs
From a side discussion in https://github.com/nilearn/nilearn/pull/738#discussion_r35963547:

> > Side-comment, this _repr_niimgs[:200] is all over the place and I have seen it too often now, not to be annoyed by it. Why not have _repr_niimg which takes an argument like shorten=True ...
> > 
> > Or even have _repr_niimgs always returns something more compact if imgs is a list of string (I seem to remember that's the case for which _repr_niimgs returns something which can be very long)
> 
> I agree with you on the fact that this is annoying. I think that we should add a shorten=True by default. We could show the long version if verbose is higher than 10 for example.
> I don't want to do it here because discussions are needed to solve this problem. If it's OK for you, I'll open an issue and we can solve it in another PR.

I just bumped into the fact that we have already _utils.short_repr which is actually not very useful because it truncates at 18 characters ... See for example this snippet

``` python
from nilearn import _utils

_utils.load_niimg(['/this/is/a/very/long/name/for/a/nifti/file'])
```

The not so great error message you get:

```
TypeError: Data given cannot be loaded because it is not compatible with nibabel format:
[/this/is/a/very/l...
```

Addition by Alex. Steps to solve it:
- [ ] Turn the two `repr_niimg` functions into a single one with `short=True` kwarg.
- [ ] Find a nice way to display the niimgs
- [ ] Replace all occurencees of `_repr_niimgs[:200]` by the shortened version (maybe the long if verboser is high?).

NicolasGensollen:
Fixes #747 

**Description**

function `_repr_niimgs` now accepts a `shorten` boolean parameter to truncate display of niimgs (default is set to True). 
Truncated display was possible before (using `_utils.short_repr`) but not straightforward, and it was truncating on the repr rather than adapting to the type of data provided (see #747 for more details).

**Examples**

See also `test_repr_niimgs`

Previous behavior:

```python
from nilearn import _utils
print(_utils._repr_niimgs('/this/is/a/very/long/name/for/a/nifti/file'))
print("-"*80)
list_of_size_3 = ['/this/is/a/very/long/name/for/a/nifti/file',
                  '/this/is/another/very/long/name/for/a/nifti/file',
                  '/this/is/again/another/very/long/name/for/a/nifti/file']
print(_utils._repr_niimgs(list_of_size_3))
list_of_size_4 = ['/this/is/a/very/long/name/for/a/nifti/file',
                  '/this/is/another/very/long/name/for/a/nifti/file',
                  '/this/is/again/another/very/long/name/for/a/nifti/file',
                  '/this/is/again/another/super/very/long/name/for/a/nifti/file']
print("-"*80)
print(_utils._repr_niimgs(list_of_size_4))
```

Outputs:

```
/this/is/a/very/long/name/for/a/nifti/file
--------------------------------------------------------------------------------
[/this/is/a/very/long/name/for/a/nifti/file, /this/is/another/very/long/name/for/a/nifti/file, /this/is/again/another/very/long/name/for/a/nifti/file]
--------------------------------------------------------------------------------
[/this/is/a/very/long/name/for/a/nifti/file, /this/is/another/very/long/name/for/a/nifti/file, /this/is/again/another/very/long/name/for/a/nifti/file, /this/is/again/another/super/very/long/name/for/a/nifti/file]
```

Example of previous use of shortening:

```python
_utils.load_niimg(list_of_size_4)
```

Raises:

```
TypeError: Data given cannot be loaded because it is not compatible with nibabel format:
[/this/is/a/very/l...
```

New behavior

```python
from nilearn import _utils
print(_utils._repr_niimgs('/this/is/a/very/long/name/for/a/nifti/file'))
print("-"*80)
list_of_size_3 = ['/this/is/a/very/long/name/for/a/nifti/file',
                  '/this/is/another/very/long/name/for/a/nifti/file',
                  '/this/is/again/another/very/long/name/for/a/nifti/file']
print(_utils._repr_niimgs(list_of_size_3, shorten=True))
print("-"*80)
print(_utils._repr_niimgs(list_of_size_3, shorten=False))
list_of_size_4 = ['/this/is/a/very/long/name/for/a/nifti/file',
                  '/this/is/another/very/long/name/for/a/nifti/file',
                  '/this/is/again/another/very/long/name/for/a/nifti/file',
                  '/this/is/again/another/super/very/long/name/for/a/nifti/file']
print("-"*80)
print(_utils._repr_niimgs(list_of_size_4, shorten=True))
print("-"*80)
print(_utils._repr_niimgs(list_of_size_4, shorten=False))
```

Outputs:

```
/this/is/a/very/lo...
--------------------------------------------------------------------------------
[/this/is/a/very/lo..., /this/is/another/v..., /this/is/again/ano...]
--------------------------------------------------------------------------------
[/this/is/a/very/long/name/for/a/nifti/file, /this/is/another/very/long/name/for/a/nifti/file, /this/is/again/another/very/long/name/for/a/nifti/file]
--------------------------------------------------------------------------------
[/this/is/a/very/lo...,
         ...
 /this/is/again/ano...]
--------------------------------------------------------------------------------
[/this/is/a/very/long/name/for/a/nifti/file,
 /this/is/another/very/long/name/for/a/nifti/file,
 /this/is/again/another/very/long/name/for/a/nifti/file,
 /this/is/again/another/super/very/long/name/for/a/nifti/file]
```

```python
_utils.load_niimg(list_of_size_4)
```

```
TypeError: Data given cannot be loaded because it is not compatible with nibabel format:
[/this/is/a/very/lo...,
         ...
 /this/is/again/ano...]
```

Not sure I made the best choices for display, let me know if you have some modifications in mind.

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -229,7 +229,7 @@ def _short_repr(niimg_rep, shorten=True, truncate=20):
                 if len(rep) + len(p) < truncate - 3:
                     rep = str(Path(p, rep))
                 else:
-                    rep = '...' + rep
+                    rep = str(Path("...", rep))
                     break
         return rep
     return niimg_rep
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -228,10 +228,10 @@ def _short_repr(niimg_rep, shorten=True, truncate=20):
             if len(niimg_rep.parts) > 1:
                 for p in niimg_rep.parts[::-1][1:]:
                     if len(rep) + len(p) < truncate - 3:
-                        rep = "{0}/{1}".format(p, rep)
+                        rep = str(pathlib.Path(p, rep))
                     else:
+                        rep = '...' + rep
                         break
-                rep = '...' + rep
             return rep
     if isinstance(niimg_rep, str) and len(niimg_rep) > truncate:
         # Shorten the repr to have a useful error message
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -120,7 +120,7 @@ def load_niimg(niimg, dtype=None):
     elif not isinstance(niimg, nibabel.spatialimages.SpatialImage):
         raise TypeError("Data given cannot be loaded because it is"
                         " not compatible with nibabel format:\n"
-                        + short_repr(niimg))
+                        + _repr_niimgs(niimg, shorten=True))
 
     dtype = _get_target_dtype(_get_data(niimg).dtype, dtype)
 
@@ -158,37 +158,65 @@ def copy_img(img):
                         copy_header=True)
 
 
-def _repr_niimgs(niimgs):
+def _repr_niimgs(niimgs, shorten=True):
     """ Pretty printing of niimg or niimgs.
+
+    Parameters
+    ----------
+    niimgs: image or collection of images
+        nibabel SpatialImage to repr.
+
+    shorten: boolean, optional, default is True
+        If True, filenames with more than 20 characters will be
+        truncated, and lists of more than 3 file names will be
+        printed with only first and last element.
+
+    Returns
+    -------
+    repr: str
+        String representation of the image.
     """
+    # Maximum number of elements to be displayed
+    # Note: should be >= 3 to make sense...
+    list_max_display = 3
+    # Simple string case
     if isinstance(niimgs, str):
-        return niimgs
+        return _short_repr(niimgs, shorten=shorten)
+    # Collection case
     if isinstance(niimgs, collections.abc.Iterable):
-        return '[%s]' % ', '.join(_repr_niimgs(niimg) for niimg in niimgs)
-    # Nibabel objects have a 'get_filename'
+        if shorten and len(niimgs) > list_max_display:
+            return '[%s]' % ',\n         ...\n '.join(_repr_niimgs(niimg, shorten=shorten) for niimg in [niimgs[0], niimgs[-1]])
+        elif len(niimgs) > list_max_display:
+            return '[%s]' % ',\n '.join(_repr_niimgs(niimg, shorten=shorten) for niimg in niimgs)
+        else:
+            return '[%s]' % ', '.join(_repr_niimgs(niimg, shorten=shorten) for niimg in niimgs)
+   # Nibabel objects have a 'get_filename'
     try:
         filename = niimgs.get_filename()
         if filename is not None:
             return "%s('%s')" % (niimgs.__class__.__name__,
-                                 filename)
+                                 _short_repr(filename,
+                                             shorten=shorten))
         else:
+            # No shortening in this case
             return "%s(\nshape=%s,\naffine=%s\n)" % \
                    (niimgs.__class__.__name__,
                     repr(niimgs.shape),
                     repr(niimgs.affine))
     except:
         pass
-    return repr(niimgs)
+    return _short_repr(repr(niimgs), shorten=shorten)
 
 
-def short_repr(niimg):
+def _short_repr(niimg_rep, shorten=True, truncate=20):
     """Gives a shorten version on niimg representation
     """
-    this_repr = _repr_niimgs(niimg)
-    if len(this_repr) > 20:
+    if not shorten:
+        return niimg_rep
+    if len(niimg_rep) > truncate:
         # Shorten the repr to have a useful error message
-        this_repr = this_repr[:18] + '...'
-    return this_repr
+        return niimg_rep[:(truncate-2)] + '...'
+    return niimg_rep
 
 
 def img_data_dtype(niimg):
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -215,7 +215,7 @@ def _short_repr(niimg_rep, shorten=True, truncate=20):
         return niimg_rep
     if len(niimg_rep) > truncate:
         # Shorten the repr to have a useful error message
-        return niimg_rep[:(truncate-2)] + '...'
+        return niimg_rep[: (truncate - 2)] + '...'
     return niimg_rep
 
 
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -10,7 +10,7 @@ import collections.abc
 
 import numpy as np
 import nibabel
-
+import pathlib
 
 def _get_data(img):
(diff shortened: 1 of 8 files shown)


nilearn/_utils/niimg.py
```diff
@@ -158,37 +158,65 @@ def copy_img(img):
...
     """
-    this_repr = _repr_niimgs(niimg)
-    if len(this_repr) > 20:
+    if not shorten:
+        return niimg_rep
+    if len(niimg_rep) > truncate:
         # Shorten the repr to have a useful error message
-        this_repr = this_repr[:18] + '...'
-    return this_repr
+        return niimg_rep[:(truncate-2)] + '...'
```
bthirion, line 218:
```suggestion
        return niimg_rep[: (truncate - 2)] + '...'
```

nilearn/tests/test_niimg_conversions.py
```diff
@@ -389,23 +389,88 @@ def test_iter_check_niimgs_memory():
...
+            '/this/is/a/very/lo...')
+    # Explicit shortening of long names
+    assert (_utils._repr_niimgs('/this/is/a/very/long/name/for/a/nifti/file',
+                                shorten=True) ==
+            '/this/is/a/very/lo...')
+    # Force long display of long names
+    assert (_utils._repr_niimgs('/this/is/a/very/long/name/for/a/nifti/file',
+                                shorten=False) ==
+            '/this/is/a/very/long/name/for/a/nifti/file')
+
```
bthirion, line 406: Could be rewritten more nicely by using 
`long_str='this/is/a/very/long/name/for/a/nifti/file'`
`short_str = '/this/is/a/very/lo...'`

nilearn/tests/test_niimg_conversions.py
```diff
@@ -389,23 +389,88 @@ def test_iter_check_niimgs_memory():
...
+            ("[/this/is/a/very/lo...,\n"
+             "         ...\n"
+             " /this/is/again/ano...]"))
+    # Long names - Force full display in pretty print style for readability
+    assert (_utils._repr_niimgs(list_of_size_4, shorten=False) ==
+            ("[/this/is/a/very/long/name/for/a/nifti/file,\n"
+             " /this/is/another/very/long/name/for/a/nifti/file,\n"
+             " /this/is/again/another/very/long/name/for/a/nifti/file,\n"
+             " /this/is/again/another/super/very/long/name/for/a/nifti/file]"))
+
```
bthirion, line 451: Can be written a bit mroe nicely too.

nilearn/tests/test_niimg_conversions.py
```diff
@@ -389,23 +389,88 @@ def test_iter_check_niimgs_memory():
...
-    assert (
-        _utils._repr_niimgs(img1).replace("10L","10") ==
-        ("%s(\nshape=%s,\naffine=%s\n)" %
-            (img1.__class__.__name__,
-             repr(shape), repr(affine))))
-    _, tmpimg1 = tempfile.mkstemp(suffix='.nii')
+    # Shorten has no effect in this case
+    for shorten in [True, False]:
+        assert (
+            _utils._repr_niimgs(img1, shorten=shorten).replace("10L","10") ==
```
bthirion, line 460:
```suggestion
            _utils._repr_niimgs(img1, shorten=shorten).replace("10L", "10") ==
```

bthirion (commented):
LGTM ! 
I have only superficial comments.

jeromedockes:
looks good! I wonder if this will break doctests in libraries that depend on nilearn but given the places this function is used I'm guessing probably not. 

I think in most cases the end of the string (path) would be more informative than the beginning, ie "...a/subj1.nii.gz" rather than "/home/nilearn/dat..."

jeromedockes:
also should `pathlib.Path`s be converted to strings before further processing?

NicolasGensollen:
Thank you for your quick reviews and comments @bthirion @jeromedockes ! :+1: 

@bthirion : I added your suggestions and cleaned the test function

@jeromedockes : 

>I think in most cases the end of the string (path) would be more informative than the beginning, ie "...a/subj1.nii.gz" rather than "/home/nilearn/dat..."

I agree when it is a path. More precisely, I think we should display "...a/subj1.nii.gz"  when the path to the file is too long, and "sub1-super-long-nam..." when the name of the file itself is too long (or both are too long...). WDYT?
 
I implemented this behavior as well as `pathlib` support as you suggested (see last commits). 

Note: We could also take care of "too-long-path" case in pure string format by splitting on "/" and doing something similar as for `Path` objects. I can add this if you think it's worth it.

jeromedockes:
> I agree when it is a path. More precisely, I think we should display "...a/subj1.nii.gz"  when the path to the file is too long, and "sub1-super-long-nam..." when the name of the file itself is too long (or both are too long...). WDYT?
>  

sounds good! I think for simplicity I would have done in all cases: (i)
convert to string (ii) cut the beginning of the string if necessary; ie
something as basic as

```
# as before
length = max(length, 10)
file_path = str(file_path)
if len(file_path) <= length:
    return file_path
return "... {}".format(file_path[-(length - 4):])
```

but the behavior you implemented makes a lot of sense so we can keep it. 

> I implemented this behavior as well as `pathlib` support as you suggested (see last commits). 
>

great! a couple of details:

- to display paths correctly on windows I would replace `"{0}/{1}".format(p,
  rep)` with `str(Path(p, rep))` (or something equivalent)
- the current version will prepend "..." to a Path even when it is not
  truncated, and will add a "/" at the beginning if the path starts with "/", ie
  _repr_niimgs(Path("/a/b")) -> "...//a/b" (maybe avoiding to think about edge
  cases like this would be an argument for a more basic behavior such as the one
  above)

> Note: We could also take care of "too-long-path" case in pure string format by splitting on "/" and doing something similar as for `Path` objects. I can add this if you think it's worth it.

splitting on "/" will not work on windows, but when input is a string it
will always represent a path so we can convert to Path so that all paths
are treated consistently, whether provided as a string (which will be
the most common case) or Path

NicolasGensollen:
Thanks for the comments @jeromedockes !
I added your suggestions and simplified the code a little bit. Let me know if you think that's good enough to merge.

bthirion (approved):
LGTM, thx !

jeromedockes (approved):
LGTM!

NicolasGensollen:
Merging... Thanks!

Issue #2465, opened 2020-05-21
importing plotting with neither matplotlib nor pytest installed results in pytest import error


jeromedockes:
see:
https://github.com/nilearn/nilearn/blob/b021ba9d02ec719e894b8be2c24204d98ba2cfa4/nilearn/plotting/__init__.py#L17

the import error reported to the user should probably be matplotlib not pytest

NicolasGensollen:
Fixed in #2641

PR #2641, opened 2020-12-22
Raise matplotlib import error instead of pytest

NicolasGensollen:
Basic fix for #2465 
Better solution?


jeromedockes (approved):
LGTM! (btw in `__init__.py` the same problem is avoided by catching the ImportError)

bthirion (approved):
Thx !

NicolasGensollen:
Thanks! Merging...

PR #2653, opened 2021-01-06
[ENH] Warn about cluster centers of mass outside of cluster

### LINKED ISSUE #2647: get_clusters_table COM sometimes outside of irregular shaped objects
Hi,

I was using get_clusters_table to exact the summary statistics of a meta-analysis that I'm doing using NiMARE and I came across some behavior which isn't incorrect but it would be good to make it explicit to the user:

- As I'm using cluster-based inference the entire cluster is associated with a probability which means that any sub peaks are not informative. NiMARE makes this explicit by basically making the clusters binary. 
- When using get_clusters_table I therefore get the Center of Mass (COM) of a cluster instead of the location of a peak value
- As the clusters can be irregular shaped the COM of a cluster can actually fall outside of the body. This is expected behavior when calculating the COM but for a meta-analysis where such summary coordinate is usually linked to an anatomical region this might result in associating a cluster with an anatomical label that doesn't actually overlap with the cluster. 

An extreme example would be a cluster in the shape of a donut but the same holds for clusters shaped like a boomerang  (which is more realistic given the curvature of the cortex). At the moment you can infer that this is happening when the peak stat output of get_clusters_table is equal to 0.0

What I would like to request is to have is a flag / warning to the user when using get_clusters_table on binary masks and the COM is outside the actual body of the cluster. This would ensure that the user is at least explicitly aware that the summary coordinate should not be directly linked to an anatomical region.

Cheers
Max

tsalo:
Closes #2647.

- Warn about centers of mass outside of cluster in `nilearn.reporting._get_clusters_table._identify_subpeaks()` by checking if (sub)peaks fall within the cluster.
- Expand test of `nilearn.reporting._get_clusters_table._local_max()` to check for multiple maxima with the same value and binary cluster with center of mass outside of cluster.

### DIFF
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -47,19 +47,49 @@ def _local_max(data, affine, min_distance):
 
 
 def _identify_subpeaks(data):
+    """Identify cluster peak and subpeaks based on minimum distance.
+
+    Parameters
+    ----------
+    data : `numpy.ndarray`
+        3D array of with masked values for cluster.
+
+    Returns
+    -------
+    ijk : `numpy.ndarray`
+        (n_foci, 3) array of local maxima indices for cluster.
+    vals : `numpy.ndarray`
+        (n_foci,) array of values from data at ijk.
+    """
     # Initial identification of subpeaks with minimal minimum distance
     data_max = ndimage.filters.maximum_filter(data, 3)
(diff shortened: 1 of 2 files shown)


bthirion (approved):
LGTM.

NicolasGensollen (approved):
LGTM thanks!

NicolasGensollen:
Merging, thanks!

PR #2656, opened 2021-01-11
bugfix resample only if type is not contour or contourf

### LINKED ISSUE #2537: Rendering problems with add_contours when binary images have rotations set in the affine matrix
**The problem**
When I plot a binary image that has some rotation to it in the affine header using Nilearn's `add_contours` function, the resulting image includes many additional lines that are not part of the surface of the ROI in the binary image. I don't encounter these issues when I use the same code with the same image that has not rotation set in the affine matrix. 

**Reproduction of the issue**
I have attached example data and code to reproduce this problem:
[contour_issue.zip](https://github.com/nilearn/nilearn/files/4871945/contour_issue.zip)

**My environment**
Python 3.8.3 (default, May 27 2020, 20:53:40)
matplotlib==3.2.2
nilearn==0.6.2

NicolasGensollen:
Fixes #2537 

See explanations in #2537. 
Not sure this is the best fix although it's pretty straightforward.

### DIFF
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -26,7 +26,7 @@ from . import cm, glass_brain
 from .edge_detect import _edge_map
 from .find_cuts import find_xyz_cut_coords, find_cut_slices
 from .. import _utils
-from ..image import new_img_like
+from ..image import new_img_like, load_img
 from ..image.resampling import (get_bounds, reorder_img, coord_transform,
                                 get_mask_bounds)
 from nilearn.image import get_data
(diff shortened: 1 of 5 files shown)


nilearn/plotting/displays.py
```diff
@@ -750,7 +750,10 @@ def add_contours(self, img, threshold=1e-6, filled=False, **kwargs):
     def _map_show(self, img, type='imshow',
                   resampling_interpolation='continuous',
                   threshold=None, **kwargs):
-        img = reorder_img(img, resample=resampling_interpolation)
+        if _utils.niimg._is_binary_niimg(img):
```
bthirion, line 753: Maybe you should explain the logic implemented here.

bthirion (commented):
Thx ! You probably want to add a test.

NicolasGensollen:
@bthirion do you think there is still something to improve here ?

bthirion (approved):
LGTM, thx !

NicolasGensollen:
Merging, thanks !

Issue #1190, opened 2016-08-02
nilearn.plotting.find_xyz_cut_coords() returns the wrong coordinate


banilo:
Applying `nilearn.plotting.find_xyz_cut_coords()`on the nifti attached here, a y coordinate is returned that is positive (i.e., in the anterior part of the brain) but is supposed to be negative (in the back of the brain). This bug is not observed with `nilearn.find_cut_slices()`.

[RTPJ-1.nii.zip](https://github.com/nilearn/nilearn/files/397428/RTPJ-1.nii.zip)

AlexandreAbraham:
The code is a bit weird. In some error cases, the fallback si to return a center coordinate. The problem being that this coordinate is not converted to "real world".

```python
    if mask.max() == 0:
        return .5 * np.array(data.shape)
```

GaelVaroquaux:
>     if mask.max() == 0:
>         return .5 * np.array(data.shape)

That's a pretty bad bug!

AlexandreAbraham:
> That's a pretty bad bug!

Also, the file @banilo sent is a very simple one with one component. Because of the thresholding magic activated by default in the code, it is zeroed out and the function fails. I don't understand why this "second threshold" is here but it is definitely not advertised in the description of the function.

AlexandreAbraham:
I'll do a quick PR to refactor the code without changing the behavior of the code.

NicolasGensollen:
Fixed in #2661

PR #2661, opened 2021-01-15
Fix find xyz cut coords function

### LINKED ISSUE #1189: nilearn.plotting.find_xyz_cut_coords returns inconsistent variable types
I just noticed that `nilearn.plotting.find_xyz_cut_coords` sometimes returns a list() object and sometimes a np.array object. I guess, this should be consistent.

NicolasGensollen:
Fixes #1189 and #1190 

- The function `find_xyz_cut_coords` now always returns a list rather than numpy array or list
- The function `find_xyz_cut_coords` returns the center of mass when all values are masked due to thresholding rather than center of image which was not in real world coordinates.

**Example**

The following code

```python
import numpy as np
import nibabel as nib
from nilearn.plotting import plot_img
from nilearn.plotting import find_xyz_cut_coords
data = np.zeros((20, 20, 20))
data[4:6, 4:6, 4:6] = 1000
img = nib.Nifti1Image(data, np.eye(4))
mask_img = nib.Nifti1Image(np.ones((20, 20, 20), dtype=int), np.eye(4))
cut_coords = find_xyz_cut_coords(img, mask_img=mask_img)
print(cut_coords)
plot_img(img)
```

was producing this output:

find_xyz_before

and now produces:

find_xyz_after

### DIFF
--- a/nilearn/plotting/find_cuts.py
+++ b/nilearn/plotting/find_cuts.py
@@ -114,7 +114,9 @@ def find_xyz_cut_coords(img, mask_img=None, activation_threshold=None):
             "Could not determine cut coords: "
             "All values were masked. "
             "Returning center of mass of unmasked data instead.")
-        cut_coords = ndimage.center_of_mass(np.abs(data)) + offset
+        # Call center of mass on initial data since my_map is zero.
+        # Therefore, do not add offset to cut_coords.
+        cut_coords = ndimage.center_of_mass(np.abs(data))
         x_map, y_map, z_map = cut_coords
         return np.asarray(coord_transform(x_map, y_map, z_map,
                                           img.affine)).tolist()
--- a/nilearn/plotting/find_cuts.py
+++ b/nilearn/plotting/find_cuts.py
@@ -96,7 +96,9 @@ def find_xyz_cut_coords(img, mask_img=None, activation_threshold=None):
         # check against empty mask
         if mask.sum() == 0.:
             warnings.warn(
-                "Provided mask is empty. Returning center of mass instead.")
+                "Could not determine cut coords: "
+                "Provided mask is empty. "
+                "Returning center of mass instead.")
             cut_coords = ndimage.center_of_mass(np.abs(my_map)) + offset
             x_map, y_map, z_map = cut_coords
             return np.asarray(coord_transform(x_map, y_map, z_map,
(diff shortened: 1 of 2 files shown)


nilearn/plotting/find_cuts.py
```diff
@@ -106,10 +108,16 @@ def find_xyz_cut_coords(img, mask_img=None, activation_threshold=None):
...
         offset += [slice_x.start, slice_y.start, slice_z.start]
-
     # Testing min and max is faster than np.all(my_map == 0)
     if (my_map.max() == 0) and (my_map.min() == 0):
-        return .5 * np.array(data.shape)
+        warnings.warn(
+            "Could not determine cut coords: "
+            "All values were masked. "
+            "Returning center of mass of unmasked data instead.")
+        cut_coords = ndimage.center_of_mass(np.abs(data)) + offset
```
bthirion, line 117: Not sure, but when you do that, the affine is not applied: only the translation "offset" part is applied, not the linear transformation. It worked if the affine is eye(4) but not in general ?

nilearn/plotting/tests/test_find_cuts.py
```diff
@@ -40,13 +40,22 @@ def test_find_cut_coords():
...
-    np.testing.assert_array_equal(
-        np.array([x, y, z]),
-        0.5 * np.array(data.shape).astype(np.float))
+    np.testing.assert_array_equal([x, y, z],
+                                  [17.5, 21., 17.5])
+
+    data = np.zeros((20, 20, 20))
+    data[4:6, 4:6, 4:6] = 1000
+    img = nibabel.Nifti1Image(data, np.eye(4))
+    mask_img = nibabel.Nifti1Image(np.ones((20, 20, 20), dtype=int), np.eye(4))
```
bthirion, line 55: I would be curious of what happens with a different affine...

NicolasGensollen:
@bthirion I believe the affine is considered whatever form it might have. The offset addition is just a trick to speed up computations: the function works on partial views of the array and keeps tracks of the offset, such that we need to add it when computing coordinates on `my_map`. The affine is then applied to these coordinates, for example here:

https://github.com/nilearn/nilearn/blob/6c040cb98fe7e70b026522be5d404e5894348a9d/nilearn/plotting/find_cuts.py#L102-L103

However, I realized I made a mistake when adding the offset in the special case where I compute the coordinates on `data` because `my_map` is fully masked (see last commit).

Here is an example with a non diagonal affine:

```python
import numpy as np
import nibabel as nib
from nilearn.plotting import plot_img
from nilearn.plotting import find_xyz_cut_coords
data = np.zeros((20, 20, 20))
data[4:6, 4:6, 4:6] = 1000
affine = np.array([[-1., 0., 0., 123.46980286],
                   [0.,  0., 1., -94.11079407],
                   [0., -1., 0., 160.694],
                   [0.,  0., 0., 1.]])
angle = np.pi / 180 * 15
rotation_matrix = np.array([[np.cos(angle), -np.sin(angle)],
                                 [np.sin(angle), np.cos(angle)]])
affine[:2, :2] = rotation_matrix * 2.0
print(affine)
img = nib.Nifti1Image(data, affine)
mask_img = nib.Nifti1Image(np.ones((20, 20, 20), dtype=int), affine)
cut_coords = find_xyz_cut_coords(img, mask_img=mask_img)
print(cut_coords)
plot_img(img)
```

find_xyz_after_2

bthirion:
Indeed, sorry for the noise.

bthirion (commented):
Some lines are marked as non-covered ?

NicolasGensollen:
>Some lines are marked as non-covered ?

Indeed, some of the added warnings were not covered by tests. 
This should have been fixed in 1290f5c. 
Thanks!

bthirion (approved):
Thx !

NicolasGensollen:
Thanks for the reviews @bthirion !
Merging...

PR #2664, opened 2021-01-20
Enh update badges

NicolasGensollen:
Update build badge since we are relying on Travis since #2633

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,4 +1,4 @@
-name: 'nilearn tests on GitHub Actions'
+name: 'build'
 
 on:
   push:
--- a/README.rst
+++ b/README.rst
@@ -1,14 +1,16 @@
 	.. -*- mode: rst -*-
 
-.. image:: https://travis-ci.org/nilearn/nilearn.svg?branch=master
-   :target: https://travis-ci.org/nilearn/nilearn
-   :alt: Travis Build Status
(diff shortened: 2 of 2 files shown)


emdupre:
The GitHub Actions Build Status doesn't [look quite right](https://github.com/NicolasGensollen/nilearn/blob/enh-update-badges/README.rst), likely because this PR also changes the workflow name.

You could avoid that change -- and get the currently correct badge -- using this URL: https://github.com/nilearn/nilearn/workflows/nilearn%20tests%20on%20GitHub%20Actions/badge.svg?branch=master&event=push

But I'm not sure if you'd prefer to change the workflow name to 'build' more generally !

NicolasGensollen:
Thanks @emdupre ! 

>The GitHub Actions Build Status doesn't look quite right, likely because this PR also changes the workflow name.

Exactly, but the button should be OK once this is merged (assuming I didn't mess up the URL...).
I changed the workflow name mostly for aesthetic reasons (the previous name was a bit too long and not rendering well in a badge in my opinion). But we can leave it as before or split the badge addition and workflow name change in two separate PRs to be sure that there is no rendering issues. I'm OK with all strategies.

emdupre:
> I changed the workflow name mostly for aesthetic reasons (the previous name was a bit too long and not rendering well in a badge in my opinion).

I certainly respect aesthetic reasons 🙂
This LGTM, then ! Thx !

emdupre (approved):

bthirion (approved):
LGTM, Thx

Issue #1632, opened 2018-04-11
Multiple scores in searchlight


robbisg:
Hi,

I am implementing (almost finished) the possibility to estimate multiple scoring functions for the searchlight analysis, using sklearn cross_validate function and returning a dictionary of maps.

I can PR the code, but I don't want to break up tests and so on, so if you are interested I will need some infos to correctly perform a smooth PR! ;)

bthirion:
Well there are 2 things:
* First check that the code is not broken after you change. For that you need to run nostests.
* Second, if you added a new functionality, you should test it in particular.

More indications here:  http://scikit-learn.org/stable/developers/contributing.html#coding-guidelines

robbisg:
Hi,

I've implemented the feature, modified tests and build another little test for the new feature!

NicolasGensollen:
Hi @robbisg 
Any updates on this or should I close the issue?

robbisg:
Hi @NicolasGensollen ,

I did this feature several time ago and it should be tested with the newest version of the code, but I don't know if is ok to include this in nilearn. If yes, I can PR otherwise it is ok to close!

Thank you!

NicolasGensollen:
@robbisg you're welcome to open a PR which will make it easy to assess how your new feature works and how it could be integrated in the newest version of the code. Let me know if I can help!

robbisg:
Great! I hope to open it asap! Thanks @NicolasGensollen !

robbisg:
@NicolasGensollen @bthirion @GaelVaroquaux 
I was modifying this feature and I found that in master version cross-validation results are averaged, but I use to keep these results since sometimes these may be useful (e.g. across-subjects analyses or check the variability of accuracy wrt selected cv-schema). 
Do you think it is useful to keep cv fold values? This may impact also some examples.

bthirion:
In general, I agree with you that having an eye on accuracy variability is instructive !
In the case of searchlight-type analysis, this information is a bit hard to represent as it is varies from voxel to voxel. You cannot do a boxplot etc. How would you convey it to the user ?

robbisg:
I was only wondering if you think that providing also the each cross-validation searchlight map could be useful. since now in ``` _group_iter_search_light``` line 171 is averaging across cv-folds.
``` python
par_scores[i] = np.mean(cross_val_score(estimator, X[:, row],
                                        y, cv=cv, n_jobs=1, **kwargs))
```
[...]

PR #2666, opened 2021-01-21
NF: Multiple scores in searchlight analysis

robbisg:
Solves #1632  .

- It allows to use multiple scoring in searchlight analyses.
- Now ```searchlight.scores_``` is a 4D image (or dict of floats) with the fourth dimension representing the cv-folds.
- Examples changed accordingly (no example has been produced for the multi-score feature)
- Added and modified tests.

Please check for the different outputs of ```search_light``` and ```SearchLight.fit```,  the first outputs always a dictionary (the output of ```cross_validate```) while ```fit``` checks whether we used multiple scores or not, if yes, the output (```sl.scores_```) is a dictionary, otherwise is an array (voxels x folds), as in the previous versions, but it's up to you how you think is the best!

### DIFF
--- a/nilearn/decoding/searchlight.py
+++ b/nilearn/decoding/searchlight.py
@@ -25,7 +25,7 @@ from .. import masking
 from ..image.resampling import coord_transform
 from ..input_data.nifti_spheres_masker import _apply_mask_and_get_affinity
 from .._utils import check_niimg_4d
-from sklearn.model_selection import cross_val_score
+from sklearn.model_selection import cross_validate
 
 ESTIMATOR_CATALOG = dict(svc=svm.LinearSVC, svr=svm.SVR)
 
@@ -53,13 +53,13 @@ def search_light(X, y, estimator, A, groups=None, scoring=None,
         group label for each sample for cross validation. default None
         NOTE: will have no effect for scikit learn < 0.18
 
-    scoring : string or callable, optional
+    scoring : list, string or callable, optional
         The scoring strategy to use. See the scikit-learn documentation
         for possible values.
-        If callable, it taks as arguments the fitted estimator, the
+        If callable, it takes as arguments the fitted estimator, the
         test data (X_test) and the test target (y_test) if y is
         not None.
-
+    
     cv : cross-validation generator, optional
         A cross-validation generator. If None, a 3-fold cross
         validation is used or 3-fold stratified cross-validation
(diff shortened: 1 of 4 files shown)


nilearn/decoding/searchlight.py
```diff
@@ -74,8 +74,9 @@ def search_light(X, y, estimator, A, groups=None, scoring=None,
 
     Returns
     -------
-    scores : array-like of shape (number of rows in A)
-        search_light scores
+    scores : dict of float of shape=(n_voxels, n_fold)
```
bthirion, line 77: I don' understand why you represent a structured set of homogeneous numerical values by a dictionary instead of a Numpy array ?

nilearn/decoding/searchlight.py
```diff
@@ -162,15 +169,27 @@ def _group_iter_search_light(list_rows, estimator, X, y, groups,
...
         kwargs = dict()
         kwargs['scoring'] = scoring
         kwargs['groups'] = groups
-        par_scores[i] = np.mean(cross_val_score(estimator, X[:, row],
-                                                y, cv=cv, n_jobs=1,
-                                                **kwargs))
+        cross_validation_res = cross_validate(estimator, X[:, row],
+                                              y, cv=cv, n_jobs=1,
+                                              **kwargs)
+        if i == 0:
```
bthirion, line 184: I don't understand the logic that is implemented here. It appears that you interpret certain keys differently if the contain 'test' but this feature is not documented...

bthirion (commented):
I understand from the test that you want to be able to generate different scores from a given round of searchlight. 
Here I think the Searchlight object should follow the same structure as the Decoder object, otherwise it is to ad hoc and the features won't be understood / used properly, and will make the user experience actually worse.

nilearn/decoding/searchlight.py
robbisg: The only reason was that I can easily check each measure by looking only at dictionary key. But I can do a numpy array of (n_measures, n_voxels, n_fold).

nilearn/decoding/searchlight.py
robbisg: Since I use the dictionary, I wanted to keep the old ```sl.scores_``` format with ndarray as output for those using only the single scoring and not a vector of scores. But using a ndarray as in the previous comment simplifies this logic.

robbisg:
So your suggestion is to use the ndarray as output with a shape of (n_measures, n_voxels, n_fold).

bthirion:
I think we should pause here and thinks about the behavior of the different objects in the library. When I think of it, the searchlight should remain similar to the Decoder object. Any hint @thomasbazeille  @tbng  ?

bthirion:
Should we close that one ?

ymzayek:
Closing for now...

PR #2669, opened 2021-01-25
Override the decoder score method

### LINKED ISSUE #2667: Decoder.score() behaviour in classification mode
Hello!

I'm not sure if this issue is a documentation or feature request issue: the `.score` seems a bit confusing for me as a new-comer. 

I used the Decoder object in a classification tasks. After fitting the model, I used the  `.score` method, expecting it to compute the mean accuracy (as it is done in scikit-learn for classifiers). I got lucky its value came back negative, it made me look at the documentation and the fact that it computes R^2 coefficient.

I think it should be a bit more explicit to the user that `.score` in a classification fashion does not compute the mean accuracy:

- either by changing the behaviour (which seems tricky as it comes from inherited scikit-learn RegressorMixin)
- or by explicitly warn the user when running the method (which should be easier by using the `self.is_classification` flag) and make a visible warning in the documentation

NicolasGensollen:
Fixes #2667 

This PR proposes to override the `score` method of the base decoder class such as to compute scores using the scoring strategy provided by the user when instantiating the decoder rather than a default method. 

The scoring strategy of the decoder is defined by

- `decoder.scoring` : Callable, or str, or `None`. Either the scoring method itself or its name as defined in the docstring. If set to `None`, `decoder.score(X, y)` will raise an error.
- `decoding._scoring_metric` : Callable. The scoring method.

### DIFF
--- a/nilearn/decoding/decoder.py
+++ b/nilearn/decoding/decoder.py
@@ -387,7 +387,6 @@ class _BaseDecoder(LinearRegression, CacheMixin):
         self.memory_level = memory_level
         self.n_jobs = n_jobs
         self.verbose = verbose
-        self.scorer = None
 
 
     def fit(self, X, y, groups=None):
@@ -468,6 +467,10 @@ class _BaseDecoder(LinearRegression, CacheMixin):
             in the inner cross validation loop. The grid is empty
             when Dummy estimators are provided.
 
+        'scorer_' : function
+            Scorer function used on the held out data to choose the best
+            parameters for the model.
+
         `cv_scores_` : dict, (classes, n_folds)
             Scores (misclassification) for each parameter, and on each fold
 
@@ -494,13 +497,13 @@ class _BaseDecoder(LinearRegression, CacheMixin):
 
         # Setup scorer
         if self.scoring is not None:
-            self.scorer = check_scoring(self.estimator,
-                                        self.scoring)
+            self.scorer_ = check_scoring(self.estimator,
+                                         self.scoring)
         else:
             if self.is_classification:
-                self.scorer = get_scorer("accuracy")
+                self.scorer_ = get_scorer("accuracy")
             else:
-                self.scorer = get_scorer("r2")
+                self.scorer_ = get_scorer("r2")
 
         # Setup cross-validation object. Default is StratifiedKFold when groups
         # is None. If groups is specified but self.cv is not set to custom CV
@@ -560,7 +563,7 @@ class _BaseDecoder(LinearRegression, CacheMixin):
                 X=X, y=y[:, c], train=train, test=test,
                 param_grid=self.param_grid,
                 is_classification=self.is_classification, selector=selector,
-                scorer=self.scorer, mask_img=self.mask_img_, class_index=c,
+                scorer=self.scorer_, mask_img=self.mask_img_, class_index=c,
                 clustering_percentile=self.clustering_percentile)
             for c, (train, test) in itertools.product(
                 range(n_problems), self.cv_))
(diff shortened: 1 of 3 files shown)


nilearn/decoding/decoder.py
```diff
@@ -583,6 +617,33 @@ def fit(self, X, y, groups=None):
...
+        Returns
+        -------
+        score : float
+            Prediction score.
+
+        """
+        if self.scoring is None:
+            raise ValueError("Unable to compute score. "
+                             "No scoring metric was provided "
+                             "to Decoder object.")
```
bthirion, line 644: But in that case, why not relay on a default metric depending on the nature of the classifier ? e.g. accuracy for a classifier, r2 for a regressor.

bthirion: I feel I'm missing something, sorry...

RaphaelMeudec: Suggestions (that might be naive as I'm not used to nilearn/sklearn) regarding the way to get the scorer:

- maybe we could use the `get_scorer` method instead of storing a list of available scorers on nilearn side?
- then we could have a `default_scorer` with value `"accuracy"` or `"r2"` depending on the `self.is_classification` flag

NicolasGensollen: Thanks for the reviews @bthirion @RaphaelMeudec!

You're right, using `get_scorer` is much better :+1:. For some reason I thought that the string values of `decoder.scoring` wouldn't be valid sklearn scorer keys, but they totally are... 

>I feel I'm missing something, sorry...

I don't think so. I hesitated between relying on a default scoring metric and raising an error, and went for the last option to make sure the user doesn't get unexpected results. I agree this is over-restrictive and changed the behaviour to rely on a default metric depending on the nature of the classifier like you both suggested.

bthirion (commented):
LGTM. beware that one line is not tested apparently.
Also, this deserves a whatsnew entry.

nilearn/decoding/decoder.py
```diff
@@ -387,6 +388,15 @@ def __init__(self, estimator='svc', mask=None, cv=10, param_grid=None,
         self.n_jobs = n_jobs
         self.verbose = verbose
 
+        # Determine the scoring strategy
+        if self.scoring is not None:
+            self._scorer = get_scorer(self.scoring)
+        else:
```
jeromedockes, line 394: should `fit` below use `self._scorer` instead of calling `check_scoring`? see:
```
 scorer = check_scoring(self.estimator, self.scoring)
```

although in any case it will not happen often that scoring is None, as users are not supposed to instantiate that class directly, but use Decoder or DecoderRegressor, that set defaults for `scoring`

NicolasGensollen: Good point, thanks! 
I updated the code to setup the scorer in `fit` rather than in the constructor, and made `scorer` an attribute of the decoder.

jeromedockes: looks great, thanks! one last detail is that I believe in scikit-learn's convention any attribute created during `fit` has a name ending with `_`, and the `__init__` does nothing besides storing the parameters passed to it. (I think it should be like this for `sklearn.base.clone` to work properly). So maybe it would stick more closely to sklearn's conventions if we did not initialize `scorer = None` in `__init__`, and added a `scorer_` attribute in `fit`?

jeromedockes (approved):
thanks!

bthirion (approved):
LGTM, thx.

NicolasGensollen:
Thanks for the reviews @bthirion @jeromedockes @RaphaelMeudec ! :+1: 
Merging...

PR #2670, opened 2021-01-25
Check events warns for invalid fields

### LINKED ISSUE #2367: check_paradigm should check for invalid keys in passed dict
Using the old `nipy` user logic, I passed `amplitude=somethx` instead of `modulation=somethx` in the `make_design_matrix`. I didn't crash but the values where ignored (e.g Error: unknown param, etc.). A default value of 1 was forced...

NicolasGensollen:
Fixes #2367 

- Warns the user when an unexpected field is present in the events dataframe
- Add tests for the `check_events` function

### DIFF
--- a/nilearn/glm/first_level/experimental_paradigm.py
+++ b/nilearn/glm/first_level/experimental_paradigm.py
@@ -17,8 +17,7 @@ import warnings
 
 import numpy as np
 
-VALID_FIELDS = set(["name",
-                    "onset",
+VALID_FIELDS = set(["onset",
                     "duration",
                     "trial_type",
                     "modulation",
--- a/nilearn/glm/first_level/experimental_paradigm.py
+++ b/nilearn/glm/first_level/experimental_paradigm.py
@@ -17,6 +17,12 @@ import warnings
 
 import numpy as np
 
+VALID_FIELDS = set(["name",
+                    "onset",
+                    "duration",
+                    "trial_type",
+                    "modulation",
+                    ])
 
 def check_events(events):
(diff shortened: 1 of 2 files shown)


nilearn/glm/first_level/experimental_paradigm.py
```diff
@@ -17,6 +17,12 @@
 
 import numpy as np
 
+VALID_FIELDS = set(["name",
```
bthirion, line 20: is 'name' a valid field ? iIm quite surprised...

bthirion (commented):
Thx for testing the warnings !
Just unsure about the "name" field ?

nilearn/glm/first_level/experimental_paradigm.py
NicolasGensollen: All events creating functions in `test_paradigm.py` return a dataframe with a `name` column, for example:

https://github.com/nilearn/nilearn/blob/a8320c343487d49e0571e3dbcda6baab6d8cdb4c/nilearn/glm/tests/test_paradigm.py#L14-L21

Although `name` is not valid in the sense that it will be used, I didn't want to systematically raise warnings for this specific key. If you think this is a preferable behaviour, I can remove "name" from this set (and/or deal with it as a special case in the function `check_events` itself).

bthirion: This is an error that has been carried by history of the project. We must allow only BIDS compliant fields.

bthirion (approved):
LGTM, thx !

NicolasGensollen:
Merging, thanks!

PR #2672, opened 2021-01-26
[MRG] Add Surface object

NicolasGensollen:
Following #2474 and as part of #2171, this PR

- adds a `Surface` namedtuple object, composed of a `Mesh` and texture data
- adds the related loading functions
- adds tests for these functions

Note: This is still WIP, any advice or suggestions are welcome! :smiley:

### DIFF
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -33,6 +33,8 @@ from nilearn.image import get_data
 # Create a namedtuple object for meshes
 Mesh = namedtuple("mesh", ["coordinates", "faces"])
 
+# Create a namedtuple object for surfaces
+Surface = namedtuple("surface", ["mesh", "data"])
 
 def _uniform_ball_cloud(n_points=20, dim=3, n_monte_carlo=50000):
     """Get points uniformly spaced in the unit ball."""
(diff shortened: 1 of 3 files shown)


nilearn/surface/surface.py
```diff
@@ -895,6 +897,65 @@ def load_surf_mesh(surf_mesh):
     return mesh
 
 
+def load_surface(surface):
+    """Loads a surface.
+
+    Parameters
+    ----------
+    surface : List or numpy.ndarray or Surface
```
bthirion, line 905:
```suggestion
    surface : List of numpy.ndarray or Surface
```

nilearn/surface/surface.py
```diff
@@ -918,7 +979,32 @@ def _check_mesh(mesh):
 
 
 def check_mesh_and_data(mesh, data):
```
bthirion, line 981: Isn't this suposed to be a private function ?

bthirion (commented):
LGTM otherwise. The next step is to use this object wherever it can across the library.

nilearn/surface/surface.py
NicolasGensollen: It seems like `check_mesh_and_data` is not private (although `_check_mesh` is), hence my choice of making `check_surface` public. I don't know if it ever was private at some point.

NicolasGensollen:
>The next step is to use this object wherever it can across the library.

Yes, doing that while preserving backward compatibility is the next step. After second thoughts, I think I will do that in another PR such as to not overload this one.

nilearn/surface/surface.py
```diff
@@ -895,6 +897,65 @@ def load_surf_mesh(surf_mesh):
...
+    ----------
+    surface : List of numpy.ndarray or Surface
+        Surface to be loaded. This can be passed as:
+
+        - a list of two files (valid formats are .gii
+        .gii.gz or Freesurfer specific files such as
+        .orig, .pial, .sphere, .white, .inflated) containing:
+            - a surface mesh geometry
+            - surface data
+        - three Numpy arrays organized in a list with coordinates,
```
jeromedockes, line 913: sorry for commenting a bit late but why do we accept this format (flat sequence of 3 elements)? I think having many options adds confusion rather than usability. I thought it would necessarily be ((coordinates, faces), data) ?

nilearn/surface/surface.py
```diff
@@ -895,6 +897,65 @@ def load_surf_mesh(surf_mesh):
...
+        faces, and data in this specific order
+        - a length 2 tuple or a namedtuple with the fields "mesh" and "data"
```
jeromedockes, line 915: there is no such thing as "a namedtuple", you can say any object with attributes "mesh" and "data", or a Surface

nilearn/surface/surface.py
```diff
@@ -895,6 +897,65 @@ def load_surf_mesh(surf_mesh):
...
+        - a length 3 tuple or a namedtuple with the fileds "coordinates",
```
jeromedockes, line 916: same question about accepting the flat sequence. also the length 3 tuple is redundant with the "three numpy arrays organized in a list" above and gives the impression that tuple and list would be treated differently somehow. (also type: fileds -> fields)

nilearn/surface/surface.py
```diff
@@ -895,6 +897,65 @@ def load_surf_mesh(surf_mesh):
...
+        "faces", and "data"
+        - a Surface object with "mesh" and "data" attributes.
```
jeromedockes, line 918: redundant with "object with the fileds 'mesh' and 'data'" above

nilearn/surface/surface.py
```diff
@@ -895,6 +897,65 @@ def load_surf_mesh(surf_mesh):
...
+
```
jeromedockes, line 919: To avoid having to list all combinations of types for mesh and data I would
suggest something like:
```
surface can be:
  - a nilearn.surface.Surface
  - a sequence (mesh, data) where:
    - mesh can be:
      - a nilearn.surface.Mesh
      - a path to .gii or .gii.gz etc.
      - a sequence of two numpy arrays, the first containing vertex coordinates
        and the second containing triangles.
    - data can be:
      - a path to .gii or .gii.gz etc.
      - a numpy array with shape (n_vertices,) or (n_time_points, n_vertices)
```

also is it possible to pass a single gifti file containing the mesh and the data?

NicolasGensollen: Thanks for the reviews @jeromedockes ! :+1: 
I agree, this is much cleaner and I'm not sure that having the possibility to load a surface with a flat sequence of length 3 would be super useful indeed. I incorporated your suggestions and updated the tests.

>also is it possible to pass a single gifti file containing the mesh and the data?

I don't think you can do that, but I don't know much about the gifti format so I might be wrong here...

jeromedockes:
> I incorporated your suggestions and updated the tests.
cool, thanks!

>> also is it possible to pass a single gifti file containing the mesh and the data?
>
> I don't think you can do that, but I don't know much about the gifti format so I might be wrong here...

Me neither :) but in any case if it is possible it could also be added
in a later PR

bthirion (approved):
LGTM, thx.

bthirion:
The next thing to do will be to include this in the documentation. We should open an issue if this is nit part of this PR.

NicolasGensollen:
>The next thing to do will be to include this in the documentation. We should open an issue if this is nit part of this PR.

I opened #2681 to list the next steps including writing the doc.
Thanks for the reviews @bthirion @jeromedockes ! :+1: 
Merging...

PR #2679, opened 2021-02-02
[MRG] Fix failures due to numpy 1.20

### LINKED ISSUE #2677: Failures in build due to numpy 1.20
Need to fix broken tests with latest package versions.

NicolasGensollen:
Fixes #2677


NicolasGensollen:
This should fix the build failure but I think we will have to clean the code of numpy type aliases to decrease the number of deprecation warnings. I think it'd be better to do that in a separate PR rather than on this one.

jeromedockes (approved):
Thanks @NicolasGensollen  for fixing this. I agree with you: it is best to merge this one now and start rooting out the deprecation warnings after

bthirion (approved):
the changes LGTM. CI seems happy too ;-)

NicolasGensollen:
Merging, thanks!

Issue #2668, opened 2021-01-24
Slightly odd behaviour of make_first_level_design_matrix


rhodricusack:
nilearn 0.70

I discovered a slightly unexpected behaviour of nilearn.glm.first_level.make_first_level_design_matrix which is that if an event is duplicated with the same onset and duration, one is ignored, but if they differ (even very slightly) then both count towards the design. I expected it to add together multiple events. Code to reproduce below - in X0, events with onsets at 0 and 10 s have same amplitude, whereas in X0, first has twice the amplitude.

Thanks so much for all of your work on this package - it is awesome.

```

from nilearn.glm.first_level import make_first_level_design_matrix
import pandas as pd
import numpy as np
from matplotlib import pyplot as plt


data = [[0, 1, 'cond1'], [0, 1, 'cond1'], [10, 1, 'cond1']] # duplicate event at 0 secs not counted
events = pd.DataFrame(data, columns=['onset','duration','trial_type'])
X0 = make_first_level_design_matrix(np.arange(20),events=events)

data = [[0, 1, 'cond1'], [0.001, 1, 'cond1'], [10, 1, 'cond1']] # duplicate event ~ 0 secs is now counted
events = pd.DataFrame(data, columns=['onset','duration','trial_type'])
X1 = make_first_level_design_matrix(np.arange(20),events=events)

fig, ax = plt.subplots(nrows=2)
ax[0].plot(X0)
ax[1].plot(X1)
plt.savefig('show_make_first_level_design_matrix_oddity.jpg')
``` 
-->

bthirion:
Thx for raising the point. I'm not 100% sure what is the right behavior in that case. I tend to think  that we should indeed leave the use the responsibility to provide an event file that does not contain ambiguous and duplicate entries, and thus not silently ignore some of the provided information: this means that in your case, we should basically add the two events.

Any other opinion ? @tsalo

NicolasGensollen:
I think that adding the events makes the most sense. I would also add a user warning, for example here:

https://github.com/nilearn/nilearn/blob/55e3f26dbd9fc6e89516e5f37e8aae23ec6086a9/nilearn/glm/first_level/experimental_paradigm.py#L21-L24

tsalo:
I agree that overlapping, but distinct, events should be added. @bthirion are you saying that the duplicate events should be treated as distinct ones? Personally, I would lean toward logging a warning when duplicate events are detected, but I think that this is rare enough that either approach should be fine.

[...]

PR #2680, opened 2021-02-02
[MRG] Glm duplicate events

NicolasGensollen:
This PR addresses #2668 by:

- proposing a way to handle duplicate events, that is events with the same trial type and onset. These events are handled in the following way: their modulation is summed, the maximum duration is kept (assuming their are different), and a warning is given to the user that something might be wrong in the input data.
- updating the `check_events` function to rely more on pandas functionalities since a `DataFrame` is assumed as input
- update and add tests

Notes
- This is still WIP
- I'm not sure this is the best way to handle this weird case. I'm open to suggestions if someone has other ideas

### DIFF
--- a/nilearn/glm/first_level/experimental_paradigm.py
+++ b/nilearn/glm/first_level/experimental_paradigm.py
@@ -101,11 +101,11 @@ def check_events(events):
     #   - trial type
     #   - onset
     column_defining_event_identity = ['trial_type',
-                                      'onset']
+                                      'onset',
+                                      'duration',]
 
     # Duplicate handling strategy
-    strategy = {'duration': np.max,   # Keep the max duration of duplicate events
-                'modulation': np.sum, # Sum the modulation values of duplicate events
+    strategy = {'modulation': np.sum, # Sum the modulation values of duplicate events
                 }
 
     cleaned_events = events_copy.groupby(
--- a/nilearn/glm/first_level/experimental_paradigm.py
+++ b/nilearn/glm/first_level/experimental_paradigm.py
@@ -16,6 +16,8 @@ Author: Bertrand Thirion, 2015
 import warnings
 
 import numpy as np
+import pandas as pd
+from pandas.api.types import is_numeric_dtype
 
 VALID_FIELDS = set(["onset",
                     "duration",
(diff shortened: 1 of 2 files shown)


tsalo:
:+1: to summing with a warning, although I'm not sure about the duration issue. I think it makes more sense to treat events with the same onset and trial type, but different duration/offset, as distinct.

NicolasGensollen:
Thanks for the comment @tsalo :+1: 

>EDIT: Apologies- I just realized that this is a WIP!

No problem, I opened the PR to see what people would think since I'm not 100% sure what the best way to handle this would be.

>I think it makes more sense to treat events with the same onset and trial type, but different duration/offset, as distinct.

OK, I can update the code to use trial type, onset, and duration do compare events.

bthirion:
I'm afraid this PR contains material from different PRs...

NicolasGensollen:
>I'm afraid this PR contains material from different PRs...

I'm not sure what you mean. Do you think I should split this in separate PRs?

bthirion:
> I'm not sure what you mean. Do you think I should split this in separate PRs?

I thought that the key checking (having 'onset', 'event_type' etc. and no exotic names) was already done in another PR ? Sorry if I' m wrong.

nilearn/glm/first_level/experimental_paradigm.py
```diff
@@ -51,27 +53,73 @@ def check_events(events):
...
+        try:
+            events_copy = events_copy.astype({'duration': float})
+        except:
+            raise ValueError("Could not cast duration to float "
+                             " in events data.")
+
+    # Handle duplicate events
+    # Two events are duplicates if they have the same:
+    #   - trial type
+    #   - onset
```
bthirion, line 102: ... and the same duration !

bthirion (commented):
Just a minro detail. LGTM otherwise.

tsalo (approved):
LGTM!

bthirion (approved):
LGTM, thx.

NicolasGensollen:
Thanks for the reviews @bthirion @tsalo ! :+1: 
Merging...

Issue #2681, opened 2021-02-03
Next steps for Mesh and Surface objects


NicolasGensollen:
Mesh and surface were introduced in #2474 and #2672.
The next steps are:

- [ ] Update related functions with backward compatibility
- [ ] Start deprecating previous usage
- [ ] Update documentation

thomasbazeille:
I think also this means #2424 is not blocked anymore. And in general Surface Maskers implementation can be done.

Once we have surface maskers, most of our object such as Decoder will be able to accept surface.

NicolasGensollen:
Linking [this discussion group](https://nipy.discourse.group/c/surface-api/10) about the surface API in Nibabel.

NicolasGensollen:
Linking this Nibabel PR : https://github.com/nipy/nibabel/pull/1056

Remi-Gau:
I feel that given the ongoing work on the Surface front, this issue is either a duplicate of a more recent one or would need to be made more specific to be useful.

Closing but feel free to reopen if you feel otherwise.

PR #2682, opened 2021-02-04
[MRG] Update surface functions

NicolasGensollen:
This PR attempts to update the surface plotting functions to comply with the new `Surface` object introduced in #2672, in order to obtain a more consistent API. See also #2681.

Assuming we have defined a surface like this:

```python
surf = Surface([mesh, map])
```

Usage:

- `plot_surf(surf_mesh, surf_map=None,...)` ==> `plot_surf(surf,...)` (map is optional)
- `plot_surf_contours(surf_mesh, roi_map,...)` ==> `plot_surf_contours(surf,...)` (map is required)
- `plot_surf_stat_map(surf_mesh, stat_map,...)` ==> `plot_surf_stat_map(surf,...)` (map is required)
- `plot_surf_roi(surf_mesh, roi_map,...)` ==> `plot_surf_roi(surf,...)` (map is required)
- `plot_img_on_surf(stat_map, surf_mesh='fsaverage5',...)` (map and mesh are reversed and special values can be passed to `surf_mesh`). Untouched so far, will probably be handled in a different way.

Passing a mesh and a map as before is still possible but gives a `FutureWarning` saying that this usage is deprecated and will be removed in release 0.9.

Passing a surface object gives a `UserWarning` saying that `surf.data` will be used as the surface map instead of what was potentially provided through `surf_map`, `roi_map`, or `stat_map`.

*Note: Although this is WIP, reviews/suggestions/ideas are welcome since there might be better ways to do this!*

### DIFF
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -120,10 +120,17 @@ pcc_map[pcc_labels] = 1
 pcc_surf = load_surface((fsaverage.pial_left,
                          pcc_map))
 
+# Define a background surface
+background = load_surface((fsaverage.pial_left,
+                           fsaverage.sulc_left))
+
 from nilearn import plotting
 
-plotting.plot_surf_roi(pcc_surf, hemi='left', view='medial',
-                       bg_map=fsaverage['sulc_left'], bg_on_data=True,
+plotting.plot_surf_roi(pcc_surf,
+                       hemi='left',
+                       view='medial',
+                       bg_surf=background,
+                       bg_on_data=True,
                        title='PCC Seed')
 
 ###############################################################################
@@ -132,14 +139,14 @@ surf = load_surface((fsaverage.pial_left,
                      stat_map))
 
 plotting.plot_surf_stat_map(surf, hemi='left', view='medial', colorbar=True,
-                            bg_map=fsaverage['sulc_left'], bg_on_data=True,
+                            bg_surf=background, bg_on_data=True,
                             darkness=.3, title='Correlation map')
 
 ###############################################################################
 # Many different options are available for plotting, for example thresholding,
 # or using custom colormaps
 plotting.plot_surf_stat_map(surf, hemi='left', view='medial', colorbar=True,
-                            bg_map=fsaverage['sulc_left'], bg_on_data=True,
+                            bg_map=background, bg_on_data=True,
                             cmap='Spectral', threshold=.5,
                             title='Threshold and colormap')
 
@@ -156,7 +163,7 @@ plotting.plot_surf_stat_map(surf, hemi='left', view='lateral', colorbar=True,
 ###############################################################################
 # The plots can be saved to file, in which case the display is closed after
 # creating the figure
-plotting.plot_surf_stat_map(surf, hemi='left', bg_map=fsaverage['sulc_left'],
+plotting.plot_surf_stat_map(surf, hemi='left', bg_surf=background,
                             bg_on_data=True, threshold=.5, colorbar=True,
                             output_file='plot_surf_stat_map.png')
 
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -83,7 +83,7 @@ print('Fsaverage5 sulcal depth map of left hemisphere is at: %s' %
 # --------------------------------
 
 # Load resting state time series from nilearn
-from nilearn.surface import load_surf_data, Surface
+from nilearn.surface import load_surf_data, load_surface
 
 timeseries = load_surf_data(nki_dataset['func_left'][0])
 
@@ -117,7 +117,8 @@ stat_map[np.where(np.mean(timeseries, axis=1) == 0)] = 0
 # Transform ROI indices in ROI map
 pcc_map = np.zeros(parcellation.shape[0], dtype=int)
 pcc_map[pcc_labels] = 1
-pcc_surf = Surface(fsaverage.pial_left, pcc_map)
+pcc_surf = load_surface((fsaverage.pial_left,
+                         pcc_map))
 
 from nilearn import plotting
 
@@ -127,7 +128,8 @@ plotting.plot_surf_roi(pcc_surf, hemi='left', view='medial',
 
 ###############################################################################
 # Display unthresholded stat map with a slightly dimmed background
-surf = Surface(fsaverage.pial_left, stat_map)
+surf = load_surface((fsaverage.pial_left,
+                     stat_map))
 
 plotting.plot_surf_stat_map(surf, hemi='left', view='medial', colorbar=True,
                             bg_map=fsaverage['sulc_left'], bg_on_data=True,
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -83,9 +83,9 @@ print('Fsaverage5 sulcal depth map of left hemisphere is at: %s' %
 # --------------------------------
 
 # Load resting state time series from nilearn
-from nilearn import surface
+from nilearn.surface import load_surf_data, Surface
 
-timeseries = surface.load_surf_data(nki_dataset['func_left'][0])
+timeseries = load_surf_data(nki_dataset['func_left'][0])
 
 # Extract seed region via label
 pcc_region = b'G_cingul-Post-dorsal'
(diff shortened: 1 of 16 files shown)


nilearn/plotting/surf_plotting.py
```diff
@@ -26,7 +29,51 @@
 VALID_HEMISPHERES = "left", "right"
 
 
-def plot_surf(surf_mesh, surf_map=None, bg_map=None,
+def _deprecate_seperate_mesh_data(func, argument):
```
tsalo, line 32:
```suggestion
def _deprecate_separate_mesh_data(func, argument):
```

Just a minor typo

bthirion (approved):
LGTM, thx !

examples/01_plotting/plot_surf_stat_map.py
```diff
@@ -147,16 +147,14 @@
...
+plotting.plot_surf_stat_map(surf, hemi='left', view='lateral', colorbar=True,
                             cmap='Spectral', threshold=.5,
                             title='Plotting without background')
 
 ###############################################################################
 # The plots can be saved to file, in which case the display is closed after
 # creating the figure
-plotting.plot_surf_stat_map(fsaverage['infl_left'], stat_map=stat_map,
-                            hemi='left', bg_map=fsaverage['sulc_left'],
+plotting.plot_surf_stat_map(surf, hemi='left', bg_map=fsaverage['sulc_left'],
```
bthirion, line 157: it begs the question whether bg_map should/could also be handled as a Surface ?

bthirion (commented):
This probably deserves a whasnew entry, but LGTM otherwise.
Thx for that one !

bthirion:
This LGTM. I just left a comment on the bg texture.

NicolasGensollen:
> I just left a comment on the bg texture.

Sorry, I missed it. I am not sure what the best option is. 
One solution would be to change the texture `bg_map` parameter of type array to a "standalone" surface `bg_surf` parameter of type Surface, meaning that the background surface would hold both its mesh and background texture (you could basically do `plot_surf(background)` and it would work just fine).
Since the surface parameter is also a Surface, it also has a mesh such that we would need to handle situations where the user provide surface and background on different meshes (raise an error or default to surface mesh rather than background's...). 
The code would then look like this:

```python
import numpy as np
from nilearn.surface import load_surface
from nilearn.plotting import plot_surf_stat_map

fsaverage = fetch_surf_fsaverage()
# Use a surface with random texture as stat map 
texture = np.random.random(10242)
surf = load_surface((fsaverage.infl_left,
                     texture))
# Define the background as a "standalone" surface object
background = load_surface((fsaverage.infl_left,
                           fsaverage.sulc_left))
# plot stat map
plot_surf_stat_map(surf,
                   bg_surf=background,
                   bg_on_data=True)
```

One of the reason we introduced `Surface` and `Mesh` objects is to avoid having untied textures, coordinates, and faces arrays in the code which makes me think that we should indeed provide a surface background rather than a simple texture although some information is redundant. 

Note that a much simpler solution would be to allow Surface objects with multiple textures as we discussed during the sprint. For example, we could provide a dictionary of `texture_label: texture_array` or something alike. 
This could for example look like:

```python
class Surface:
    def __init__(self, mesh, textures):
        self.mesh = mesh
        self._textures = textures
    def texture(self, label=""):
        if label in self._textures:
            return self._textures[label]
        else:
            return None

s = Surface(mesh=fsaverage.infl_left, 
            textures={'background': fsaverage.sulc_left,
                      'stat_map': texture})

# You can then do the following:
s.mesh
s.texture('background')
s.texture('stat_map')
# would be clever enough to plot both map and 
# background is there are matching keys...
plot_surf_stat_map(s)
```

bthirion:
> Since the surface parameter is also a Surface, it also has a mesh such that we would need to handle situations where the user provide surface and background on different meshes (raise an error or default to surface mesh rather than background's...).

Note that there are cases where two meshes are different , yet in one-to-one correspondence, which is quite handy to compare them.  Freesurfer uses is quite often (e.g. pial and white mesh, white and inflated,  fsaverage and spherical, etc). You don't want to raise an error for such cases.
I'm a bit uncertain what to do, as having different textures called through keywords makes the API more complicated (and is reminicent of Gifti's intents).  I think I still prefer to have the texture as a separate surface, but this is not a strong opinion.

nilearn/plotting/html_surface.py
```diff
@@ -227,32 +234,75 @@ def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
...
-    surf_mesh : str or list of two numpy.ndarray
-        Surface mesh geometry, can be a file (valid formats are
-        .gii or Freesurfer specific files such as .orig, .pial,
-        .sphere, .white, .inflated) or
-        a list of two Numpy arrays, the first containing the x-y-z coordinates
-        of the mesh vertices, the second containing the indices
-        (into coords) of the mesh faces.
+    surf_mesh : str or list of two numpy.ndarray or Mesh or Surface
+
+            .. deprecated:: 0.7.2
```
bthirion, line 253: Need to change the version

nilearn/plotting/surf_plotting.py
```diff
@@ -41,25 +101,63 @@ def plot_surf(surf_mesh, surf_map=None, bg_map=None,
...
-        Surface mesh geometry, can be a file (valid formats are
-        .gii or Freesurfer specific files such as .orig, .pial,
-        .sphere, .white, .inflated) or
-        a list of two Numpy arrays, the first containing the x-y-z coordinates
-        of the mesh vertices, the second containing the indices
-        (into coords) of the mesh faces, or a Mesh object with
-        "coordinates" and "faces" attributes.
+    surf_mesh : str or list of two numpy.ndarray or Mesh or Surface
+
+            .. deprecated:: 0.7.2
```
bthirion, line 106: Version to be chnaged here too.

bthirion:
@alexisthual I'd love to have your opinion on that one.
Best,

alexisthual:
Very nice work and ideas!

I haven't spent as much time as you working on this but here are some thoughts.

My opinion as a user is that I want to be free to manipulate `Texture` and `Mesh` objects independently and have them constitute a `Surface` object on the fly when needed. The current PR seems to do a very good job at this! To me, `bg_map` expecting a `Texture` object (and/or eventually a `Surface` whose `Mesh` would simply be ignored) seems natural.

I wouldn't allow a single `Surface` object to have several `Texture` objects available through keys (as a user, I would most likely store these textures in a custom dict I would handle myself). However, I feel like the broader underlying question has to do with the merging of several textures. To me, a user could want to merge several `Texture` objects into a single `Surface` object, and choose how to aggregate them. It could look something like this:

```python
fsaverage = fetch_surf_fsaverage()
texture = np.random.random(10242)
# create a surface whose texture is the point-wise product of two textures
surf = load_surface(
  (
    fsaverage.infl_left,
    [texture, fsaverage.sulc_left]
  ),
  agg="product" # could be "add", "max", etc.
)

plot_surf_stat_map(surf)
```

which, in most cases maybe, could also simply be achieved with `numpy`

```python
surf = load_surface(
  (
    fsaverage.infl_left,
    np.multiply(texture, fsaverage.sulc_left)
  )
)
```

meaning that this solution might be interesting only if there is a need for some complex aggregators.

One final idea would be to combine `Surface`s and/or `Texture`s using operators, allowing for a very concise syntax:

```python
fsaverage = fetch_surf_fsaverage()
texture = np.random.random(10242)
surf = load_surface((fsaverage.infl_left,
                     texture))

# plot point-wise product of surface and sulc texture
plot_surf(surf * fsaverage.sulc_left)
```

If any of this makes sense, it probably calls for a new PR :smile:

jeromedockes:
> Besides, `bg_map` and `bg_on_data` didn't feel clear to me the first time I used them (why set `bg_on_data=True` if I already set a `bg_map`?), so maybe there is room for improvement.

I agree



> However, I feel like the broader underlying question has to do with the merging of several textures.

The main issue is when data from both hemispheres needs to be handled as a single unit, eg to be fed to a decoding pipeline (which I think would correspond to aggregating by concatenation in your description, but there needs to be an inverse transform)



> To me, `bg_map` is just a way to perform a product of textures before plotting it

yes but that's an implementation detail, we don't want the user to be aware of that or take care of the blending of surface map and background



> One final idea would be to combine `Surface`s and/or `Texture`s using operators, allowing for a very concise syntax:

this kind of syntactic sugar often leads to less readable code in my experience. I would prefer to restrict the overloading of mathematical operators to simple objects that truly behave like numbers or matrices

bthirion:
Tx for the discussion.
One aspect is that there are different conventions for bg, which is plotted in grey scale vs the main texture, which gets a more flashy colorbar.

@jeromedockes I don't see how we handle the case with two meshes/textures for two hemispheres ?

jeromedockes:
> @jeromedockes I don't see how we handle the case with two meshes/textures for two hemispheres ?

we don't. what I meant was that the issue discussed during the sprint had to do with Surfaces covering a whole brain, rather than multiple textures for a single mesh. given that this still isn't solved (ongoing discussion in [nibabel issue](https://github.com/nipy/nibabel/issues/936) ) I suggest leaving aside the multiple texture question for now.

for the bg_img, I also think it would be more consistent for it to be a Surface

bthirion:
@NicolasGensollen could we have a quick discussion over that one ? Is it simply waiting for a second review ?

bthirion:
Thx. We should at least discuss it to see how we coordinate with nibabel.

ymzayek:
I think this is worth finishing (and then also working on #2424). Even though work is underway on the Coordinate Image API, I think it will be a while before we can adopt it even if it's merged soon considering waiting for a release and supporting older versions of nibabel. @NicolasGensollen and @jeromedockes WDYT? Do you foresee major issues in the future to convert to the new API after developing surface support a bit using our Surface object?

ymzayek:
Can we close this since now we are working on #3856?

Remi-Gau:
I would say yes

Remi-Gau:

Remi-Gau:
Closing as this will be replaced by other PRs in the devlopment of surface based analysis.

Also the mount of merge conflicts is making this one almost impossible bring back.

jeromedockes:
also this one was for the old surface api which has been abandoned

PR #2684, opened 2021-02-05
[MRG] Mosaic mode of displaying brain images

### LINKED ISSUE #212: mosaic
We need a mosaic mode

KamalakerDadi:
Closes #212 .

- display_mode='mosaic' should plot images of multiple rows and columns
- By default, ```cut_coords``` are automatically found for 7 slices to plot against three direction xyz.

A simple code that works:
```python
from nilearn import datasets

# localizer dataset to have contrast maps

from nilearn import plotting

plotting.plot_stat_map(stat_img, display_mode='mosaic')
plotting.show()
```
plot_stat_map

```python
from nilearn.datasets import MNI152_FILE_PATH
from nilearn import plotting

plotting.plot_img(MNI152_FILE_PATH, display_mode='mosaic')
plotting.show()
```
plot_img

However, there are problems.

- Code breaks if ```cut_coords``` are provided manually by the user. As current code forms a list of 7 values/coordinates per dicretion in a dictionary. e.g., cut_coords={'x': [0], 'y': [0], 'z': [0]}. Shall I raise an error with valuable error message? or change the documentation to accept list of coordinates in a dictionary for mosaic mode?

- Code breaks if same coordinates are found against three directions. E.g., -46.875 -24.375  -5.625. Since I am inspired from the logic implemented in the class ```BaseStackedSlicer```. In a dictionary, each coordinate is a key per axes value that is passed to ```add_overlay```. I don't have any ideas here.
```python
from nilearn import datasets
from nilearn.image.image import mean_img

# Compute the mean EPI: we do the mean along the axis 3, which is time
func_filename = haxby_dataset.func[0]
mean_haxby = mean_img(func_filename)

from nilearn.plotting import find_cut_slices
for direction in ['x', 'y', 'z']:
    print(find_cut_slices(mean_haxby, direction=direction, n_cuts=7))
```
These are the issues so far I came across. Any suggestions on how to overcome these are welcome.

Rest:
-  tests
-  demo
-  documentation

### DIFF
--- a/examples/01_plotting/plot_demo_more_plotting.py
+++ b/examples/01_plotting/plot_demo_more_plotting.py
@@ -134,6 +134,14 @@ plotting.plot_stat_map(stat_img, display_mode='tiled',
                        cut_coords=[36, -27, 60],
                        title="display_mode='tiled'")
 
+########################################
+# Visualizing three views along multiple rows and columns
+# -------------------------------------------------------------------------
+# display_mode='mosaic' for sagittal, coronal and axial view
+
+plotting.plot_stat_map(stat_img, display_mode='mosaic',
+                       title="display_mode='mosaic'")
+
 ###############################################################################
 # Demonstrating various display features
 # ---------------------------------------
--- a/examples/01_plotting/plot_demo_more_plotting.py
+++ b/examples/01_plotting/plot_demo_more_plotting.py
@@ -137,11 +137,31 @@ plotting.plot_stat_map(stat_img, display_mode='tiled',
 ########################################
 # Visualizing three views along multiple rows and columns
 # -------------------------------------------------------------------------
-# display_mode='mosaic' for sagittal, coronal and axial view
+# display_mode='mosaic' for sagittal, coronal and axial view with
+# default option i.e. cut_coords=None
 
 plotting.plot_stat_map(stat_img, display_mode='mosaic',
-                       title="display_mode='mosaic'")
+                       title="display_mode='mosaic' default cut_coords")
 
+########################################
+# Now, changing the number of slices along columns
+# -------------------------------------------------------------------------
+# display_mode='mosaic' for sagittal, coronal and axial view with
+# number of slices specified as integer i.e. cut_coords=3
+
+plotting.plot_stat_map(stat_img, display_mode='mosaic',
+                       cut_coords=3,
+                       title="display_mode='mosaic' with cut_coords=3")
+
+########################################
+# Now, another way of limiting the number of slices along rows and columns
+# -------------------------------------------------------------------------
+# display_mode='mosaic' for sagittal, coronal and axial view with
+# number of slices specified as tuple of length 3
+
+plotting.plot_stat_map(stat_img, display_mode='mosaic',
+                       cut_coords=(5, 4, 10),
+                       title="display_mode='mosaic' with cut_coords as tuple")
 ###############################################################################
 # Demonstrating various display features
 # ---------------------------------------
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -7,6 +7,7 @@ the data with different layout of cuts.
 
 import collections.abc
 import numbers
+import itertools
 from distutils.version import LooseVersion
 
 import matplotlib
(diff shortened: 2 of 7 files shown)


NicolasGensollen:
Hi @KamalakerDadi 

This looks good! :+1: 

>Shall I raise an error with valuable error message? or change the documentation to accept list of coordinates in a dictionary for mosaic mode?

I would expect to be able to specify `cut_coords` as a list/tuple and/or a dict like:

```python
plotting.plot_stat_map(stat_img, display_mode='mosaic', cut_coords=(5,2,3))
```

which would produce (I just changed slightly the `find_cut_coords` method of `MosaicSlicer` to handle iterables for `cut_coords`):

mosaic

So I would vote for updating the documentation to specify the acceptable formats for `cut_coords` in mosaic mode.

>Code breaks if same coordinates are found against three directions.

I looked a little bit into this and I think you can change the keys of `self.axes` to be a tuple `(direction, coord)` instead of just `coord` in `_init_axes`:

```python
self.axes[(direction, coord)] = display_ax
```
This way, it won't be overwritten for each direction. I tried it quickly and it doesn't seem to break anything in the `plot_stat_map` example since the keys values aren't used (I didn't verify this extensively though...).

Hope this helps!

KamalakerDadi:
Hi @NicolasGensollen 

Great suggestions and thanks.

KamalakerDadi:
Hi @NicolasGensollen I looked at thoroughly and seems like the behavior is not consistent with other display modes as far as accpeting inputs for ```cut_coords``` with ```display_mode='mosaic'```.

>I would expect to be able to specify cut_coords as a list/tuple and/or a dict like:

```cut_coords``` already accepts as list/tuple and integers.

If list and display_mode='ortho', then these are positions of the coordinates.

If integer and display_mode='x' or 'y' or 'z', then these are the number of slices where coordinates are found automatically.

Coming to your suggestion:
>plotting.plot_stat_map(stat_img, display_mode='mosaic', cut_coords=(5,2,3))

It founds the number of slices which is not ideal with the following display modes: 'ortho', 'xy', 'yz', 'xz', .... This may confuse users. As list/tuple indicates the position of the coordinates.

**Comment by NicolasGensollen** [comment]
Hi @KamalakerDadi 
I understand your concerns as this behaviour might be confusing even if well documented. Indeed, I am not sure having different numbers of cuts for each directions in mosaic mode is super useful. 
If it is, maybe adding to the docstring something like:

```
If display_mode == 'mosaic', and the number of cuts is the same 
for all directions, cut_coords can be specified as an integer. 
It can also be a length 3 tuple specifying the number of cuts for 
every direction if these are different.
```

could be clear enough?

**Comment by KamalakerDadi** [comment]
>If display_mode == 'mosaic', and the number of cuts is the same 
>for all directions, cut_coords can be specified as an integer. 
>It can also be a length 3 tuple specifying the number of cuts for 
>every direction if these are different.

>could be clear enough?

yes it is. But, still a tuple of length 3 specifies the number of cuts. The one which you were suggesting earlier. Am I right?
Shall I implement this?

**Comment by NicolasGensollen** [comment]
>yes it is. But, still a tuple of length 3 specifies the number of cuts. The one which you were suggesting earlier. Am I right?
Shall I implement this?

Yes, I was indeed referring to the same tuple. The behaviour makes sense to me as long as the docstring is clear enough. So, unless other users feel otherwise, I think we can implement it like this. What do you think?

**Comment by KamalakerDadi** [comment]
> yes it is. But, still a tuple of length 3 specifies the number of cuts. The one which you were suggesting earlier. Am I right?
   > Shall I implement this?

>Yes, I was indeed referring to the same tuple. The behaviour makes sense to me as long as the docstring is clear enough. So, >unless other users feel otherwise, I think we can implement it like this. What do you think?

Ok. Lets see how it goes.

**Comment by KamalakerDadi** [comment]
Once everything passes, my focus will be on documentation and tests.

**Comment by bthirion** [comment]
Tests pass, thx !

File: examples/01_plotting/plot_demo_more_plotting.py [hunk 1]
```diff
@@ -134,6 +134,14 @@
                        cut_coords=[36, -27, 60],
                        title="display_mode='tiled'")
 
+########################################
+# Visualizing three views along multiple rows and columns
+# -------------------------------------------------------------------------
+# display_mode='mosaic' for sagittal, coronal and axial view
+
+plotting.plot_stat_map(stat_img, display_mode='mosaic',
+                       title="display_mode='mosaic'")
```
NicolasGensollen, line 143: Maybe showcase how to use custom cut coordinates here?

nilearn/plotting/displays.py
```diff
@@ -1807,8 +1813,205 @@ class YZSlicer(OrthoSlicer):
     _cut_displayed = 'yz'
 
 
+class MosaicSlicer(BaseSlicer):
+    """
```
NicolasGensollen, line 1817: Could you add a docstring here?

NicolasGensollen (commented):
This is looking great! :+1: 
I'm not sure what is the status of this PR. Sorry if this is still WIP and I'm reviewing it too early. 
I left a few comments. Also:

- The current test error is not related, you should merge/rebase on top of master to get the fix #2706
- I think the docstrings of the plotting functions supporting the mosaic slicer should be updated

examples/01_plotting/plot_demo_more_plotting.py
KamalakerDadi: Done

nilearn/plotting/displays.py
KamalakerDadi: Done

KamalakerDadi:
This PR is ready to merge.

bthirion:
What about the CodeCov complaints ? Is this simply an error ?

KamalakerDadi:
>What about the CodeCov complaints ? Is this simply an error ?

I don't get CodeCov's hungriness for tests. Without being passing those code lines, we don't get an output, right? which means in a way those are covered?

nilearn/plotting/displays.py
```diff
@@ -1807,8 +1812,219 @@ class YZSlicer(OrthoSlicer):
...
+        img : 3D Nifti1Image
+            The brain image.
+
+        threshold : float, optional
+            The lower threshold to the positive activation. If None, the
+            activation threshold is computed using the 80% percentile of
+            the absolute value of the map.
+
+        cut_coords : list/tuple of 3 floats, integer, optional
+            xyz world coordinates of cuts.
```
NicolasGensollen, line 1847: Maybe we can add a line saying what happen when `cut_coords` isn't provided?

nilearn/plotting/tests/test_displays.py
```diff
@@ -53,6 +56,58 @@ def test_tiled_slicer():
...
+    # default cut_coords=None
+    slicer = MosaicSlicer.init_with_figure(img=img)
+    slicer.add_overlay(img, cmap=plt.cm.gray, colorbar=True)
+    # Forcing a layout here, to test the locator code
+    with tempfile.TemporaryFile() as fp:
+        slicer.savefig(fp)
+
+    # cut_coords as an integer
+    slicer = MosaicSlicer.init_with_figure(img=img, cut_coords=4)
+    slicer.add_overlay(img, cmap=plt.cm.gray, colorbar=True)
```
NicolasGensollen, line 70: I think it could be nice to add some additional checks in here. For example, we could check that everything is fine with `cut_coords`:

```python
for d in ['x', 'y', 'z']:
    assert d in slicer.cut_coords
    assert len(slicer.cut_coords[d]) == 4
```
WDYT?

nilearn/plotting/tests/test_displays.py
```diff
@@ -53,6 +56,58 @@ def test_tiled_slicer():
...
+    # cut_coords as a tuple
+    slicer = MosaicSlicer.init_with_figure(img=img, cut_coords=(4, 5, 2))
+    slicer.add_overlay(img, cmap=plt.cm.gray, colorbar=True)
```
NicolasGensollen, line 73: Same here, maybe something like this:

```python
assert len(slicer.cut_coords['x']) == 4
assert len(slicer.cut_coords['y']) == 5
assert len(slicer.cut_coords['z']) == 2
```

nilearn/plotting/tests/test_displays.py
```diff
@@ -53,6 +56,58 @@ def test_tiled_slicer():
...
+    mslicer = MosaicSlicer(cut_coords=(1, 1, 1))
+    img = load_mni152_template()
+    mslicer.add_overlay(img, cmap=plt.cm.gray)
+    # cut_coords as integer
+    mslicer = MosaicSlicer(cut_coords=5)
+    mslicer.add_overlay(img, cmap=plt.cm.gray)
+    # cut_coords as dictionary
+    mslicer = MosaicSlicer(cut_coords={'x': [10, 20],
+                                       'y': [30, 40],
+                                       'z': [15, 16]})
```
NicolasGensollen, line 104: Here we could check than `mslicer` has the provided cut coordinates in the same way.

NicolasGensollen (changes requested):
The code looks good! :+1:  
I have a few very minor comments (see below), mainly about improving the tests a little bit.

KamalakerDadi:
Thanks @NicolasGensollen I addressed your comments

bthirion (approved):
LGTM, thx !

NicolasGensollen (approved):
LGTM, thanks!

NicolasGensollen:
It looks like there is a conflict in `whats_new` due to a recent merge. 
@KamalakerDadi could you solve it?

NicolasGensollen:
Thanks! Merging! :tada:

KamalakerDadi:
>Thanks! Merging! 

Thank you.

PR #2688, opened 2021-02-07
fix for FirstLevelModel compute_contrast for multiple runs with varying matrice size or columns

### LINKED ISSUE #2687: compute_contrast in FLM fails when used on multiple runs with different shapes of design matrices
Context : 
I am trying to obtain activation maps for a contrast between two conditions. I have several runs that I want to input in a FLM. For each run, I have a confounds file generated by [fMRIPrep](https://fmriprep.org/en/stable/). These files include the cosines for drift compensation. Because the length of each run varies a little bit, the number of cosines generated in the confounds files varies too. I first load the confounds file with [load_confounds](https://pypi.org/project/load-confounds/), then I generate the design matrices with [make_first_level_design_matrix](https://nilearn.github.io/modules/generated/nilearn.glm.first_level.make_first_level_design_matrix.html), by passing the confounds in add_regs.

Problem : 
When I use runs with different number of cosines from the confounds files (so, design matrices of the different runs have a different number of columns) the call to compute_contrast fails. It does not fail when I use runs with identical number of cosines.
I've found a [similar problem](https://github.com/nilearn/nilearn/issues/2394) in the issues, but since in my case the two conditions of my contrast are present in all runs, I thought that it might be a related but different issue. 

Or maybe, as some responses to the other issue seemed to suggest, this is a feature and not a bug ? If that's the case, how should I approach this question instead ?

Thanks in advance for your help !

Nilearn version:
0.7.0

### Expected behavior
compute_contrast should (?) be able to handle a list of design matrices with varying number of columns.

### Actual behavior
When two design matrices with a different number of columns are passed to compute_contrast, the function fails and returns the following error : 

`
~/anaconda3/envs/hyruuk_shinobi_behav/lib/python3.8/site-packages/nilearn/glm/contrasts.py in compute_contrast(labels, regression_result, con_val, contrast_type)
     92         for label_ in regression_result:
     93             label_mask = labels == label_
---> 94             resl = regression_result[label_].Tcontrast(con_val)
     95             effect_[:, label_mask] = resl.effect.T
     96             var_[label_mask] = (resl.sd ** 2).T

~/anaconda3/envs/hyruuk_shinobi_behav/lib/python3.8/site-packages/nilearn/glm/model.py in Tcontrast(self, matrix, store, dispersion)
    198             raise ValueError("t contrasts should have only one row")
    199         if matrix.shape[1] != self.theta.shape[0]:
--> 200             raise ValueError("t contrasts should be length P=%d, "
    201                              "but this is length %d" % (self.theta.shape[0],
    202                                                         matrix.shape[1]))

ValueError: t contrasts should be length P=49, but this is length 47`

I joined an pdf file that shows my code and the outputs. First cell are imports. Second cell shows expected behavior, with two runs that share the same number of confounds. Third cell shows the bug, the code fails to generate the contrast because the two runs have different number of confounds (44 and 46). Fourth cell shows the same runs as in the second cell (the one that works), but I manually removed one of the cosines (to make sure that the number of columns of the design_matrix would differ), and that recreates the issue.

[bug_flm_confounds.pdf](https://github.com/nilearn/nilearn/files/5936005/bug_flm_confounds.pdf)

```python

```

bpinsard:
Closes #2687 .

Inverts the order of 2 blocks of codes in `compute_contrast` first
https://github.com/nilearn/nilearn/blob/e372946bfc91419bc1dbbfadebfed6e6a402956d/nilearn/glm/first_level/first_level.py#L578-L583
 then https://github.com/nilearn/nilearn/blob/e372946bfc91419bc1dbbfadebfed6e6a402956d/nilearn/glm/first_level/first_level.py#L572-L578 

This fixes computation of contrast given as text formulas for:
- matrices with varying size (mostly due to cosine drifts in runs with varying length) that for now crashes unless hacked.
- matrices with varying column ordering (for whatever weird reason) for which the contrast would be computed on the wrong columns assuming order from first run matrix.

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -568,6 +568,11 @@ class FirstLevelModel(BaseGLM):
             raise ValueError('contrast_def must be an array or str or list of'
                              ' (array or str)')
 
+        n_runs = len(self.labels_)
+        if len(con_vals) != n_runs:
+            warn('One contrast given, assuming it for all %d runs' % n_runs)
+            con_vals = con_vals * n_runs
+        
         # Translate formulas to vectors
         for cidx, (con, design_mat) in enumerate(zip(con_vals,
                                                      self.design_matrices_)
@@ -576,11 +581,6 @@ class FirstLevelModel(BaseGLM):
             if isinstance(con, str):
                 con_vals[cidx] = expression_to_contrast_vector(
                     con, design_columns)
-
-        n_runs = len(self.labels_)
-        if len(con_vals) != n_runs:
-            warn('One contrast given, assuming it for all %d runs' % n_runs)
-            con_vals = con_vals * n_runs
 
         valid_types = ['z_score', 'stat', 'p_value', 'effect_size',
                        'effect_variance']
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -570,7 +570,7 @@ class FirstLevelModel(BaseGLM):
 
         n_runs = len(self.labels_)
         n_contrasts = len(con_vals)
-        if n_contrasts == 1:
+        if n_contrasts == 1 and n_runs > 1:
             warn('One contrast given, assuming it for all %d runs' % n_runs)
             con_vals = con_vals * n_runs
         elif n_contrasts != n_runs:
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -332,7 +332,7 @@ class FirstLevelModel(BaseGLM):
     def fit(self, run_imgs, events=None, confounds=None,
             design_matrices=None):
         """Fit the GLM
-        
+
         For each run:
         1. create design matrix X
         2. do a masker job: fMRI_data -> Y
@@ -569,10 +569,14 @@ class FirstLevelModel(BaseGLM):
                              ' (array or str)')
 
         n_runs = len(self.labels_)
-        if len(con_vals) != n_runs:
+        n_contrasts = len(con_vals)
+        if n_contrasts == 1:
             warn('One contrast given, assuming it for all %d runs' % n_runs)
             con_vals = con_vals * n_runs
-        
+        elif n_contrasts != n_runs:
+            raise ValueError('%n contrasts given, while there are %n runs' %
+                             (n_contrasts, n_runs))
+
         # Translate formulas to vectors
         for cidx, (con, design_mat) in enumerate(zip(con_vals,
                                                      self.design_matrices_)
--- a/nilearn/glm/tests/test_first_level.py
+++ b/nilearn/glm/tests/test_first_level.py
@@ -7,6 +7,7 @@ import shutil
 import numpy as np
 import pandas as pd
 import pytest
+import warnings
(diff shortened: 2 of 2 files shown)


bpinsard:
Looking at the tests, it seems that coverage could be improved by including tests with contrasts as formulas as most tests for now are with contrast vectors.
From `generate_reports()` specification, it seems that passing a list of vectors for a single contrast is not possible and only the formula can cover the case of differing matrix size.

nilearn/glm/first_level/first_level.py
```diff
@@ -568,6 +568,11 @@ def compute_contrast(self, contrast_def, stat_type=None,
             raise ValueError('contrast_def must be an array or str or list of'
                              ' (array or str)')
 
+        n_runs = len(self.labels_)
```
NicolasGensollen, line 571: This is unrelated to the change of this PR, but it seems like the length of `con_vals` is never checked properly. It will work if it is either 1 or `n_runs` but anything else will give an incorrect warning saying that only one contrast was given and will build a larger list than needed which will fail afterwards.

NicolasGensollen (commented):
Thanks for the PR @bpinsard !
It would be useful if we could replicate easily the bug in #2687 and check that this PR is fixing the issue. I might have missed it but I don't think the input data (or a download link) were provided in #2687. 
Ideally, it would be nice to add a test failing with previous behaviour and passing with the proposed fix.

nilearn/glm/first_level/first_level.py
bpinsard: Nice catch, I will add a validation of that variable.

bpinsard:
I will add tests for that case tomorrow.

bthirion (commented):
The change looks good  thx ! 
To make it irreversible, you need to add a unit test that breaks with the previous version but passes with the new one. Can you do that ?
Best,

bpinsard:
@bthirion @NicolasGensollen 
Just pushed a test to reproduce the error and added a validation that 1 or n_runs contrasts are given.

nilearn/glm/first_level/first_level.py
bthirion: This is great.
Can you add tests to ensure that the warning/error are raised indeed ?
Thx !

nilearn/glm/tests/test_first_level.py
```diff
@@ -216,6 +216,43 @@ def test_high_level_glm_different_design_matrices():
...
+    # Fit a glm with 3 sessions and design matrices
+    multi_session_model = FirstLevelModel(mask_img=mask).fit(
+        fmri_data, design_matrices=design_matrices)
+
+    # raise when n_contrast != n_runs | 1
+    with pytest.raises(ValueError):
+        multi_session_model.compute_contrast([np.eye(rk)[1]]*2)
+
+    multi_session_model.compute_contrast([np.eye(rk)[1]]*3)
+    multi_session_model.compute_contrast([np.eye(rk)[1]])
```
NicolasGensollen, line 254: Could you check that the added warning is correctly given here?

bthirion (approved):
LGTM, thx !

NicolasGensollen (approved):
LGTM, thanks ! :+1:

NicolasGensollen:
Merging!

Issue #1501, opened 2017-08-24
Parallelizing NiftiSpheres/LabelsMasker?


dPys:
Is there any way to speed up NiftiSpheresMasker and NiftiLabelsMasker (e.g. via joblib, iterables, etc.)? Maybe this functionality already exists or is in the works, but I haven't been able to find it.  If it does not yet exist, this would be a very cool feature to incorporate. I would think that these functions would be mpi-parallelizable across coordinates (i.e. the total runtime should theoretically be reducible to the length of time that it takes to extract signal from a single sphere/label assuming that number of coordinates = # cores?)

KamalakerDadi:
- Yes, PR will come soon. All masker families should benefit via joblib caching and parallelization and multi-subject extraction. This is part of issue opened at #1240.

bthirion:
- Addressed in #2294.

bthirion:
'n_jobs' argument could be added to NiftiSpheresMasker though.

dPys:
@bthirion -- were you thinking [here](https://github.com/nilearn/nilearn/blob/55e3f26dbd9fc6e89516e5f37e8aae23ec6086a9/nilearn/regions/signal_extraction.py#L124-L127)

bthirion:
Naively I was thinking about iterating across spheres in the SpheresMasker but this leads to bad memory usage, so indeed, you suggestion might be better.

dPys:
#2693

Remi-Gau:
I wonder if it may be good to revisit this one during the sprint: for images with a lot of images or labels, we may get some speed up from parallelizing over image / label.

The problem is that this will impact MultiMaskers that are parallelizing over subjects: so when user are trying to leverage n_jobs for Multimaskers is the n_jobs supposed to parallelize over subjects or labels? do we hard code that decision or do we let the user choose dependending on where their bottle neck is?

I think this could potentially be a low-hanging fruit where we can get some performance gain.

bthirion:
But NiftiSpheres and NiftiLablesMasker (actually all non-Multi-maskers) are not designed to deal with multi-subjects settings primarily. So, it would make sense to parallelize on regions/spheres/labels instead for those.

Remi-Gau:
> But NiftiSpheres and NiftiLablesMasker (actually all non-Multi-maskers) are not designed to deal with multi-subjects settings primarily. So, it would make sense to parallelize on regions/spheres/labels instead for those.

[...]

PR #2693, opened 2021-02-12
Parallelize nifti labels masker

dPys:
Addresses/closes Issue #1501 

*Revised `img_to_signals_labels` to support n_jobs argument.
*To keep track of node indices and their corresponding extracted time-series vectors with joblib, I modified reduction_function to be nested. To minimize memory footprint further, we could also employ memory mapping if we feel it necessary. 

Cheers,
@dPys

### DIFF
--- a/nilearn/regions/signal_extraction.py
+++ b/nilearn/regions/signal_extraction.py
@@ -132,8 +132,6 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
         reduce_ts = getattr(ndimage.measurements, strategy)
         return n, np.asarray(reduce_ts(img, labels, index))
 
-    img_sig_list = [[n, img] for n, img in enumerate(np.rollaxis(data, -1))]
-
     with Parallel(
             n_jobs=n_jobs,
             verbose=verbose,
@@ -145,7 +143,7 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
                 index=labels,
                 n=n
             )
-            for n, img in img_sig_list
+            for n, img in enumerate(np.rollaxis(data, -1))
         )
     for n, sig in outs:
         signals[n] = sig
--- a/nilearn/regions/signal_extraction.py
+++ b/nilearn/regions/signal_extraction.py
@@ -18,7 +18,8 @@ from ..image import new_img_like
 
 # FIXME: naming scheme is not really satisfying. Any better idea appreciated.
 def img_to_signals_labels(imgs, labels_img, mask_img=None,
-                          background_label=0, order="F", strategy='mean'):
+                          background_label=0, order="F", strategy='mean',
+                          n_jobs=-1, verbose=0):
     """Extract region signals from image.
 
     This function is applicable to regions defined by labels.
@@ -53,6 +54,10 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
         Must be one of: sum, mean, median, mininum, maximum, variance,
         standard_deviation. Default='mean'.
 
+    n_jobs : int, optional
+        The number of CPUs to use to do the computation. -1 means
+        'all CPUs'.
+
     Returns
     -------
     signals : numpy.ndarray
@@ -73,6 +78,8 @@ def img_to_signals_labels(imgs, labels_img, mask_img=None,
         e.g. clusters
 
     """
+    from joblib import Parallel, delayed
+
     labels_img = _utils.check_niimg_3d(labels_img)
 
     # TODO: Make a special case for list of strings (load one image at a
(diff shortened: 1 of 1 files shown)


NicolasGensollen (changes requested):
Thanks for the PR @dPys !
The code looks good, I just have a few comments:

- I think we need to expose `njobs` to the `NiftiLabelsMasker`.
- The default value for `njobs` is usually 1 rather than -1. 
- Also, I would be in favour of adding a couple tests where `njobs` is set to different values and results are checked to be the same.

dPys:
> Thanks for the PR @dPys !
> The code looks good, I just have a few comments:
> 
> * I think we need to expose `njobs` to the `NiftiLabelsMasker`.
Done. I also added some error-handling for it.

> * The default value for `njobs` is usually 1 rather than -1.
This was a typo from when I was tinkering, good catch!
 
> * Also, I would be in favour of adding a couple tests where `njobs` is set to different values and results are checked to be the same.
See commit [a88f21c](https://github.com/nilearn/nilearn/pull/2693/commits/a88f21c2deaf15d1c7d3af253d0805fe09aff2ba) where I've added one. I agree that this is a nice test to have although is there any particular mathematical reason why we should expect the results to diverge with a different number of threads in this context? Also, you raise an interesting new precedent because there are no other tests in nilearn, as far as I can tell, that involve manipulating n_jobs in a systematic fashion as I'm doing here. e.g. in [test_searchlight](https://github.com/nilearn/nilearn/blob/master/nilearn/decoding/tests/test_searchlight.py#L32), n_jobs is held constant at 1. My sense is that if we indeed decide to incorporate this test here, parallelized functionality elsewhere should probably be held to the same standard, no?

NicolasGensollen (changes requested):
Thanks for addressing all the points so quickly! :+1: 
I only have a few small comments (see below).

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -368,3 +368,57 @@ def test_nifti_labels_masker_with_mask():
...
+    masker_1 = NiftiLabelsMasker(labels_img, mask_img=mask_img, n_jobs=1)
+    expected_result = masker_1.fit_transform(fmri_img).squeeze()
+
+    for n in [-1, 2]:
+        masker = NiftiLabelsMasker(labels_img, mask_img=mask_img, n_jobs=n)
+        n_result = masker.fit_transform(fmri_img).squeeze()
+        assert np.allclose(n_result, expected_result)
+
+    with pytest.raises(ValueError,
+                       match=f"Invalid value for n_jobs '{-2}'. Must be an "
```
NicolasGensollen, line 400: I think this f-string is breaking the python 3.5 tests

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -368,3 +368,57 @@ def test_nifti_labels_masker_with_mask():
...
+                             f"integer >= 1 or == to -1."):
+        NiftiLabelsMasker(
+            labels_img,
+            n_jobs=-2
+        )
+
+    with pytest.raises(ValueError,
+                       match=f"Invalid value for n_jobs '{0}'. Must be an "
```
NicolasGensollen, line 408: and this one

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -368,3 +368,57 @@ def test_nifti_labels_masker_with_mask():
...
+                             f"integer >= 1 or == to -1."):
+        NiftiLabelsMasker(
+            labels_img,
+            n_jobs=0
+        )
+
+    with pytest.raises(ValueError,
+                       match=f"Invalid value for n_jobs '{1.5}'. Must be an "
```
NicolasGensollen, line 416: and also this one

NicolasGensollen:
I thought about the `n_jobs` test to make sure the `n_jobs` parameter is handled correctly across the pipeline. I understand your point that there no reason we should get different results (assuming no bugs in joblib). I still think it's nice to have it (especially now that it's written...), what do you think @bthirion ?

dPys:
Requested changes made.

bthirion:
I'm a bit unsure whether using 'n_jobs=1' in test results in a waste of resources.  
testing for 'n_jobs=2' is indeed less susceptible of generating waste of cost/CPU time.
@NicolasGensollen bzw, we should keep an eye on build time: it is good to have everything tested, but we have to be careful about computation resources.

dPys:
> I'm a bit unsure whether using 'n_jobs=1' in test results in a waste of resources.
> testing for 'n_jobs=2' is indeed less susceptible of generating waste of cost/CPU time.
> @NicolasGensollen bzw, we should keep an eye on build time: it is good to have everything tested, but we have to be careful about computation resources.

One thing to note @bthirion -- n_jobs = 1 is actually not really being "tested" here. It is simply used once initially to simulate an `expected_result` to compare with that being derived from multithreaded computation (i.e. n_jobs=2). 

Added build time is ~1.34 seconds.

bthirion:
Thx. LGTM then.

dPys:
> Thx. LGTM then.

Sounds good. Excited to see how this scales to 1000+ labels... 

Thanks @NicolasGensollen and @bthirion for your review.

NicolasGensollen (approved):
I think we can add an entry in whats_new. 
Looks good to me otherwise, thanks! :+1:

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -178,6 +186,14 @@ def __init__(self, labels_img, background_label=0, mask_img=None,
             raise ValueError("invalid value for 'resampling_target' "
                              "parameter: " + str(resampling_target))
 
+        if float(n_jobs) < -1 or not float(n_jobs).is_integer() or \
```
jeromedockes, line 189: these checks look like they should either happen for all maskers or none of them

dPys: @jeromedockes -- I've gone ahead and relocated the checks. Please let me know if the most recent push doesn't address your concern here, in which case let me know how you'd prefer this be handled.

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -286,6 +286,15 @@ def transform_single_imgs(self, imgs, confounds=None):
             shape: (number of scans, number of labels)
 
         """
+
+        if float(self.n_jobs) < -1 or not float(self.n_jobs).is_integer() or \
```
jeromedockes, line 290: IIUC the check is repeated here to avoid doing the resampling if the ExtractionFunctor init is going to fail, is that right?

dPys: Correct!

nilearn/regions/signal_extraction.py
```diff
@@ -18,7 +18,8 @@
 
 # FIXME: naming scheme is not really satisfying. Any better idea appreciated.
 def img_to_signals_labels(imgs, labels_img, mask_img=None,
-                          background_label=0, order="F", strategy='mean'):
+                          background_label=0, order="F", strategy='mean',
+                          n_jobs=1, verbose=0):
```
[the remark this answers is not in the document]
dPys: Added. I also moved the n_jobs correctness error-handling from `_ExtractorFunctor` to live directly in img_to_signals_labels given that the `__call__` to the latter is the only method of the `_ExtractorFunctor` class

PR #2695, opened 2021-02-12
MAINT: Avoid NumPy deprecations

larsoner:
Avoid at least some warnings of the form:
```
  /home/larsoner/python/nilearn/nilearn/glm/regression.py:100: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -132,7 +132,7 @@ def run_glm(Y, X, noise_model='ar1', bins=100, n_jobs=1, verbose=0):
             / (ols_result.residuals ** 2).sum(axis=0)
         )
         del ols_result
-        ar1 = (ar1 * bins).astype(np.int) * 1. / bins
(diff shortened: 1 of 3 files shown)


NicolasGensollen (approved):
LGTM, thanks!

bthirion (commented):
Thx, this is great ! But I see many more of those in the codebase...

larsoner:
The only ambiguity I see is whether to convert `np.int` to `int` (32-bit on Windows) or `np.int64`. If you're okay with me doing `np.int64` everywhere I can probably make the rest of the substitutions

bthirion:
I looked a bit superficially, and my first impression is that `int` would do the job in  most places.

PR #1557, opened 2017-11-16
WIP - adding functionality to resample Neurovault fetched images

nicofarr:
Related to issue #1553 

In order to deal with both resampled and non-resampled version, we need to implement a mechanism that checks the presence or absence of the resampled or original files. 

WIP - currently I'm not happy with the way it's implemented, because it ONLY works if you use the `mode=overwrite` in the fetch_neurovault functions. 

This is because I implemented the caching mechanism at the lowest level (download function), but the check for downloaded images seems to happen earlier.

Issue #1553, opened 2017-11-16
Automatic resampling option for the neurovault fetcher


nicofarr:
It would be good to add an option to the neurovault fetcher so that images are resampled right after downloading, and stored only in resampled format. 

This would be useful to save space when downloading large collections. 

Ran into this issue when considering building an example (see #1506) using a large Neurovault collection (for example collection 503)

@jeromedockes

GaelVaroquaux:
+1 on my side: if we store these via the caching mechanism of the downloaders, it will save a lot of space and compute time for the users.

NicolasGensollen:
Addressed in #2696

PR #2696, opened 2021-02-16
Resampling functionality for NeuroVault

RaphaelMeudec:
This PR updates what have been done in #1557 by @nicofarr and related to issue #1553

What have been done here in addition to prior work:

- rebase on master and resolve conflicts
- modify the neurovault mock to allow a fake image download (discussed with @jeromedockes in #1557)
- get the test up to date with the neurovault mock and split it into multiple ones

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -41,12 +41,16 @@ Enhancements
   colorbar of surface plots. The default format is scientific notation except for :func:`nilearn.plotting.plot_surf_roi`
   for which it is set as integers.
 
-- :class:`nilearn.input_data.NiftiMasker`, :class:`nilearn.input_data.NiftiLabelsMasker`, 
+- :class:`nilearn.input_data.NiftiMasker`, :class:`nilearn.input_data.NiftiLabelsMasker`,
   :class:`nilearn.input_data.MultiNiftiMasker`, :class:`nilearn.input_data.NiftiMapsMasker`,
   and :class:`nilearn.input_data.NiftiSpheresMasker` can now compute high variance confounds
   on the images provided to `transform` and regress them out automatically. This behaviour is 
   controlled through the `high_variance_confounds` boolean parameter of these maskers which
   default to False.
+
+- :func:`nilearn.datasets.neurovault.fetch_neurovault` now implements a `resample` boolean argument to either
+  perform a fixed resampling during download or keep original images. This can be handy to reduce disk usage.
+  By default, the downloaded images are not resampled.
 
 .. _v0.7.0:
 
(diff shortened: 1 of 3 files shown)


RaphaelMeudec:
Codecov seems to fail because it compares test coverage right before the first commit of the PR (which was [last year](https://github.com/nilearn/nilearn/commit/4a3e943)). Though, I don't think there is a coverage problem with this PR as the functionality is tested.

bthirion (commented):
I don't  see any outstanding issue here.
Simply add a whatsnew entry.
Best,

NicolasGensollen (approved):
This looks good to me. Thanks for finishing this PR @RaphaelMeudec ! :+1:

jdkent:
I don't have any formal proof yet, but I believe the default interpolation strategy for resample_img (`continuous`) is sub-optimal for statistical maps. I was working on neurostuff/NiMARE#439, and found when running a meta-analysis using the `resample_img` defaults the results looked very pixelated and not biologically plausible, but when I used `linear` interpolation the results looked more plausible. Also changing `clip=True` may be worth consideration as to not introduce new larger statistical values than were in the original image.

doc/whats_new.rst
```diff
@@ -41,6 +41,10 @@ Enhancements
   colorbar of surface plots. The default format is scientific notation except for :func:`nilearn.plotting.plot_surf_roi`
   for which it is set as integers.
 
+- :func:`nilearn.datasets.neurovault.fetch_neurovault` now implements a `resample` boolean argument to either
```
NicolasGensollen, line 44: I think you need to change the reference to `nilearn.datasets.fetch_neurovault`

NicolasGensollen:
@jdkent do you want to open a separate issue on this? Also, do you have an example we could take a look at?

bthirion:
@jdkent Thx for the comment
I'm just wondering whether we should acknowledge that not all images in Neurovault are statistical images: some of them are atlases, or at least display discrete labels. interpolation='linear' would not given sensible results for such images.
Add another keyword argument then ?

RaphaelMeudec:
@NicolasGensollen I fixed the whatsnew entry. Not sure if I should proceed with the `interpolation` argument?

bthirion:
I think we have to add this as another kw argument.

nilearn/datasets/neurovault.py
```diff
@@ -2056,6 +2117,7 @@ def _read_download_params(
     wanted_image_ids=None, max_images=None,
     max_consecutive_fails=_MAX_CONSECUTIVE_FAILS,
     max_fails_in_collection=_MAX_FAILS_IN_COLLECTION,
+    resample=False, interpolation='continuous',
```
bthirion, line 2120: Following @jdkent comment, it would be better to use `interpolation=linear` as a default.

NicolasGensollen (approved):
LGTM, thanks!

bthirion (approved):
LGTM, thx !

NicolasGensollen:
Thank you all! :+1: 
Merging...

PR #2702, opened 2021-02-17
Support atlases in plot_carpet

### LINKED ISSUE #2515: Support atlases in plot_carpet
Currently, [nilearn.plotting.plot_carpet](https://github.com/nilearn/nilearn/blob/cc0f43ce329bda5af2eb4cb01b7207ed4ef24b9c/nilearn/plotting/img_plotting.py#L1467) only accepts a binary mask, but it would be preferable to accept an atlas to separate tissue types in the plot (see #1585). This better matches the implementation in MRIQC and the description from the original paper.

tsalo:
Closes #2515.

To do:

- [x] Label atlas values in left-side color bar
- [ ] Support probabilistic atlases? Is that even possible?
- [x] Update tests.

- Allow `mask_img` argument to be a discrete atlas.
- Group voxels by values in the discrete atlas.
- Add a color bar, when an atlas is provided, to the left side of the plot.


tsalo:
Here's what it currently looks like with a 10-value atlas:

image

bthirion:
Here is a high-level suggestion: 
If the atlas has labels (provided as another argument), display the labels. The colorbar alone is not very informative regarding region specification.

tsalo:
Agreed. I'm just not sure how to do that part yet.

tsalo:
I think I've got the labels working now. Here's the standard usage (binary mask, no labels):
Screen Shot 2021-02-18 at 3 36 30 PM

Atlas mask, no labels:
Screen Shot 2021-02-18 at 3 36 45 PM

Atlas mask, with labels:
Screen Shot 2021-02-18 at 3 36 39 PM

bthirion:
Looks good !

nilearn/plotting/img_plotting.py
```diff
@@ -1982,10 +2018,60 @@ def plot_carpet(img, mask_img=None, detrend=True, output_file=None,
...
+        if mask_labels:
+            # Add labels to middle of each associated band
+            mask_labels_inv = {v: k for k, v in mask_labels.items()}
+            ytick_locs = [
+                np.mean(np.where(atlas_values == i)[0])
+                for i in np.unique(atlas_values)
+            ]
+            ax0.set_yticks(ytick_locs)
+            ax0.set_yticklabels([
+                mask_labels_inv[int(i)] for i in np.unique(atlas_values)
```
bthirion, line 2050: We should have checked earlier that `atlas_values` and `mask_labels` are consistent.

bthirion (commented):
tests are obviously welcome ;-)

nilearn/plotting/img_plotting.py
tsalo: I'm having some trouble coming up with a clean way to check this and to coerce the `mask_labels` values to the same datatype as the `atlas_values`.

bthirion: But you can coerce `atlas_values` to int ? It would cover all the use cases I know of ?

tsalo: No, that's just a (bad) placeholder. I plan to have something better working before I switch this PR from draft.

tsalo: @bthirion sorry I completely misread your original comment. I thought you were asking why I did it the way I did instead of recommending changing the target of the datatype change! Your idea was definitely the right approach, while my original approach was the "bad placeholder". 🤦

nilearn/plotting/img_plotting.py
```diff
@@ -1996,25 +2089,29 @@ def plot_carpet(img, mask_img=None, detrend=True, output_file=None,
...
     if title:
         axes.set_title(title)
+
     labels = tr * (np.array(xticks))
     labels *= (2 ** n_decimations)
     axes.set_xticklabels(['%.02f' % t for t in labels.tolist()])
 
     # Remove and redefine spines
-    for side in ['top', 'right']:
+    for side in ["top", "right"]:
```
bthirion, line 2102: (Nitpicking) @kchawla-pi had told me that single quote should be preferred to double quotes IIRC.

bthirion (commented):
Looks good !
Could you take care of the two lines not covered by tests (?) and add a whatsnew entry ?
Thx,

nilearn/plotting/img_plotting.py
tsalo: Sorry, I normally use `black` to automate my formatting, and it uses double quotes. I can switch.

bthirion (approved):
LGTM, thx !

NicolasGensollen (commented):
This looks good to me, thanks a lot! :+1: 
I just have a super minor comment (see below).

nilearn/plotting/img_plotting.py
```diff
@@ -1909,8 +1911,10 @@ def plot_markers(node_values, node_coords, node_size='auto',
     return display
 
 
-def plot_carpet(img, mask_img=None, detrend=True, output_file=None,
-                figure=None, axes=None, vmin=None, vmax=None, title=None):
+def plot_carpet(img, mask_img=None, mask_labels=None,
+                detrend=True, output_file=None,
+                figure=None, axes=None, vmin=None, vmax=None, title=None,
+                colorbar=False, cmap=plt.cm.gist_ncar):
```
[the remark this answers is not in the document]
tsalo: Thanks for catching this! `colorbar` is unused, so I removed it. I've added the others to the docstring.

tsalo:
The coverage issues seem to be random and inaccurate. Everything else looks good to me.

NicolasGensollen (approved):
LGTM, thanks! :tada:

PR #2706, opened 2021-02-19
fix sampling in generate_fake_fmri_data_and_design

### LINKED ISSUE #2705: Possible identical design matrix columns with generate_fake_fmri_data_and_design
**Description**

The function `generate_fake_fmri_data_and_design` can return design matrices with identically named columns because these names (letters) are randomly sampled using the numpy `choice`  function which does sampling with replacement by default:

https://github.com/nilearn/nilearn/blob/55e3f26dbd9fc6e89516e5f37e8aae23ec6086a9/nilearn/_utils/data_gen.py#L344

This is causing the recently added test `test_high_level_glm_different_design_matrices_formulas` to fail in a random fashion because it computes contrast of the form "a-a" (see current master failure for instance...).

**Proposed fix**

Do sampling without replacement to avoid this.

NicolasGensollen:
Fixes #2705

### DIFF
--- a/nilearn/_utils/data_gen.py
+++ b/nilearn/_utils/data_gen.py
@@ -341,7 +341,9 @@ def generate_fake_fmri_data_and_design(shapes, rk=3, affine=np.eye(4)):
         data = np.random.randn(*shape)
         data[1:-1, 1:-1, 1:-1] += 100
         fmri_data.append(Nifti1Image(data, affine))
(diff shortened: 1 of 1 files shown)


bthirion (approved):
LGTM, thx.

NicolasGensollen:
Thanks! Merging quick to get rid of these random test failures...

Issue #2689, opened 2021-02-08
Masker reports


NicolasGensollen:
### Improving the NiftiMasker report ?

During our last meeting we talked about NiftiMasker reports and whether they were useful for end users, and if not, how we could improve them. Are the provided information useful and easy to read? Should we add other details or plots?
I think having an issue on this could start the discussion by getting feedbacks.

Below is a basic code to generate a NiftiMasker report:

```python
from nilearn import datasets
from nilearn.input_data import NiftiMasker

dataset = datasets.fetch_development_fmri(n_subjects=1)
nifti_masker = NiftiMasker(standardize=True, mask_strategy='background',
                           smoothing_fwhm=8)
nifti_masker.fit(dataset.func[0])
report = nifti_masker.generate_report()
report
```

![niftimasker_report](https://user-images.githubusercontent.com/2639645/107263092-a181f800-6a41-11eb-9082-1d185225c103.png)

### Generalizing reports to other maskers

We also talked about extending reports to other maskers. I'm not sure yet what they could look like, but if someone has some ideas on what useful information and plots they could include, feel free to make suggestions! We could start with a relatively simple one (like the NiftiLabelsMasker?) for example.

bthirion:
It could display the volume (in cm^3) of the masked region.

NiftiLabelsMasker: the number of regions defined by the masker ?

emdupre:
Thanks, @NicolasGensollen ! Here's an example report that I'm wondering about -- I remember something similar confused @GaelVaroquaux in a tutorial :

```python
from nilearn import datasets
from nilearn.input_data import NiftiMasker

haxby_dataset = datasets.fetch_haxby()
nifti_masker = NiftiMasker(mask_img=haxby_dataset.mask_vt[0])
nifti_masker.fit()
report = nifti_masker.generate_report()
report
```
![Screenshot from 2021-02-08 16-22-19](https://user-images.githubusercontent.com/15017191/107283052-dd10c800-6a29-11eb-92bc-b037d477a499.png)


My question would be : what should we show if no image is supplied to fit ?

NicolasGensollen:
Thanks @bthirion @emdupre for the ideas!
If no image is supplied to fit, the current implementation of the report sets the image to be displayed as the mask:

https://github.com/nilearn/nilearn/blob/e372946bfc91419bc1dbbfadebfed6e6a402956d/nilearn/input_data/nifti_masker.py#L278-L286

[...]

PR #2707, opened 2021-02-19
[MRG][ENH] NiftiLabelsMasker Reporting

NicolasGensollen:
See #2689 

**Description**

This draft PR proposes to add reporting capabilities to the `NiftiLabelsMasker`: 

- It reworks the template logic a little bit to be able to select different templates depending on the estimator
- It proposes a possible report format which is composed of a `plot_roi` of the regions as suggested by @GaelVaroquaux  in #2689, a count of the different regions, and adds a button to display information about each region. For now, only the label number, the name, the size (in number of voxels), and relative size are provided for each region.

*Note: Comments are welcome to improve the report!* 

**Example**

```python
from nilearn import datasets
from nilearn.input_data import NiftiLabelsMasker

atlas = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
data = datasets.fetch_development_fmri(n_subjects=1)
masker = NiftiLabelsMasker(labels_img=atlas.maps,
                           labels=atlas.labels)
masker.fit(data.func[0])
report
```
niftilabelsmasker_report

Clicking on the "Regions summary" button gives the following:

niftilabelsmasker_report2

**Todo**

- [x] Make sure everybody agrees on what information should be provided and update the report accordingly
- [x] Add tests
- [x] Update doc/examples

### DIFF
--- a/examples/06_manipulating_images/plot_nifti_labels_simple.py
+++ b/examples/06_manipulating_images/plot_nifti_labels_simple.py
@@ -1,13 +1,18 @@
 """
-Simple example of NiftiLabelsMasker use
-=======================================
+Extracting signals from brain regions using the NiftiLabelsMasker
+=================================================================
 
-Here is a simple example of automatic signal extraction using the nifti
-labels masker.
+This simple example shows how to extract signals from functional
+fmri data and brain regions defined through an atlas.
+More precisely, this example shows how to use the NiftiLabelsMasker
+object to perform this operation in just a few lines of code.
 """
 
 ###########################################################################
 # Retrieve the brain development functional dataset
+#
+# We start by fetching the brain development functional dataset
+# and we restrict the example to one subject only.
 
 from nilearn import datasets
 dataset = datasets.fetch_development_fmri(n_subjects=1)
@@ -18,6 +23,8 @@ print('First functional nifti image (4D) is at: %s' % func_filename)
 
 ###########################################################################
 # Load an atlas
+#
+# We then load the Harvard-Oxford atlas to define the brain regions
 atlas = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
 
 # The first label correspond to the background
@@ -26,11 +33,13 @@ print('The atlas contains {} non-overlapping regions'.format(
 
 ###########################################################################
 # Instantiate the mask and visualize atlas
+#
 from nilearn.input_data import NiftiLabelsMasker
 
 # Instantiate the masker with label image and label values
 masker = NiftiLabelsMasker(atlas.maps,
-                           labels=atlas.labels)
+                           labels=atlas.labels,
+                           standardize=True)
 
 # Visualize the atlas
 # Note that we need to call fit prior to generating the mask
@@ -54,8 +63,24 @@ report = masker.generate_report()
 report
 
 ###########################################################################
-# Preprocess the data with the NiftiLablesMasker
-fmri_masked = masker.transform(func_filename)
-# fmri_masked is now a 2D matrix, (n_time_points x n_regions)
-fmri_masked.shape
+# Process the data with the NiftiLablesMasker
+#
+# In order to extract the signals, we need to call transform on the
+# functional data
+signals = masker.transform(func_filename)
+# signals is a 2D matrix, (n_time_points x n_regions)
+signals.shape
 
+###########################################################################
+# Plot the signals
+import matplotlib.pyplot as plt
+
+fig = plt.figure(figsize=(15, 5))
+ax = fig.add_subplot(111)
+for label_idx in range(3):
+    ax.plot(signals[:, label_idx],
+            linewidth=2,
+            label=atlas.labels[label_idx+1]) # 0 is background
+ax.legend(loc=2)
+ax.set_title("Signals for first 3 regions")
+plt.show()
--- a/examples/06_manipulating_images/plot_nifti_labels_simple.py
+++ b/examples/06_manipulating_images/plot_nifti_labels_simple.py
@@ -29,7 +29,7 @@ atlas = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
 
 # The first label correspond to the background
 print('The atlas contains {} non-overlapping regions'.format(
-                                len(atlas.labels)-1))
+    len(atlas.labels) - 1))
 
 ###########################################################################
 # Instantiate the mask and visualize atlas
@@ -80,7 +80,7 @@ ax = fig.add_subplot(111)
 for label_idx in range(3):
     ax.plot(signals[:, label_idx],
             linewidth=2,
-            label=atlas.labels[label_idx+1]) # 0 is background
+            label=atlas.labels[label_idx + 1])  # 0 is background
 ax.legend(loc=2)
 ax.set_title("Signals for first 3 regions")
 plt.show()
--- a/examples/06_manipulating_images/plot_nifti_labels_simple.py
+++ b/examples/06_manipulating_images/plot_nifti_labels_simple.py
@@ -1,0 +1,61 @@
+"""
+Simple example of NiftiLabelsMasker use
+=======================================
+
+Here is a simple example of automatic signal extraction using the nifti
+labels masker.
+"""
+
+###########################################################################
+# Retrieve the brain development functional dataset
+
+from nilearn import datasets
+dataset = datasets.fetch_development_fmri(n_subjects=1)
+func_filename = dataset.func[0]
+
+# print basic information on the dataset
+print('First functional nifti image (4D) is at: %s' % func_filename)
+
+###########################################################################
+# Load an atlas
+atlas = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
+
+# The first label correspond to the background
+print('The atlas contains {} non-overlapping regions'.format(
+                                len(atlas.labels)-1))
+
+###########################################################################
+# Instantiate the mask and visualize atlas
+from nilearn.input_data import NiftiLabelsMasker
+
+# Instantiate the masker with label image and label values
+masker = NiftiLabelsMasker(atlas.maps,
+                           labels=atlas.labels)
+
+# Visualize the atlas
+# Note that we need to call fit prior to generating the mask
+masker.fit()
+
+# At this point, no functional image has been provided to the masker.
+# We can still generate a report which can be displayed in a Jupyter
+# Notebook, opened in a browser using the .open_in_browser() method,
+# or saved to a file using the .save_as_html(output_filepath) mathod.
+report = masker.generate_report()
+report
+
+##########################################################################
+# Fitting the mask and generating a report
+masker.fit(func_filename)
+
+# We can again generate a report, but this time, the provided functional
+# image is displayed with the ROI of the atlas.
+# The report also contains a summary table giving the region sizes in mm3
+report = masker.generate_report()
+report
+
+###########################################################################
+# Preprocess the data with the NiftiLablesMasker
+fmri_masked = masker.transform(func_filename)
+# fmri_masked is now a 2D matrix, (n_time_points x n_regions)
+fmri_masked.shape
+
--- a/nilearn/input_data/nifti_labels_masker.py
+++ b/nilearn/input_data/nifti_labels_masker.py
@@ -221,8 +221,8 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
                               message=mpl_unavail_msg)
                 return [None]
 
-        if 'labels_image' in self._reporting_data:
-            labels_image = self._reporting_data['labels_image']
+        labels_image = self._reporting_data['labels_image']
+        if labels_image is not None:
             labels_image_data = image.get_data(labels_image)
             labels_image_affine = image.load_img(labels_image).affine
             # Number of regions excluding the background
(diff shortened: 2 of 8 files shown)


nilearn/input_data/nifti_labels_masker.py
```diff
@@ -178,6 +187,60 @@ def __init__(self, labels_img, background_label=0, mask_img=None,
...
+        if 'labels_image' in self._reporting_data:
+            labels_image = self._reporting_data['labels_image']
+            labels_image_data = image.get_data(labels_image)
+            number_of_regions = np.sum(np.unique(labels_image_data) != 0)
+            self._report_content['description'] = (
+            'This reports shows the regions defined by the labels of the mask.')
+            self._report_content['number_of_regions'] = number_of_regions
+            
+            label_values = np.unique(labels_image_data).nonzero()[0]
+            columns = ['label value', 'region name', 'size (in #voxels)',
```
bthirion, line 222: I'd put the size in mm^3, because the # voxels depends on the arbitrary image resolution.

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -178,6 +187,60 @@ def __init__(self, labels_img, background_label=0, mask_img=None,
...
+                       'relative size (in %)']
+            if self.labels is None:
+                columns.remove('region name')
+            regions_summary = {c:[] for c in columns}
+            for label in label_values:
+                regions_summary['label value'].append(label)
+                if self.labels is not None:
+                    regions_summary['region name'].append(self.labels[label])
```
bthirion, line 230: this assumes that 'self.labels' is consistent with 'label_values'

nilearn/reporting/html_report.py
```diff
@@ -9,6 +10,31 @@
...
+
+def _get_estimator_template(estimator):
+    """Returns the HTML template to use for a given
+    estimator if a specific template was defined in
+    ESTIMATOR_TEMPLATES. Otherwise, return the default
+    template.
+
+    Parameters
+    ----------
+    estimator : ll
```
bthirion, line 25: ??

emdupre:
Thanks, @NicolasGensollen ! I'm running it locally and it looks great. My biggest question is whether we should show the supplied functional niimg under the ROIs rather than an anatomical image (if an image is provided to `fit` as in your example) to check overlap ?

NicolasGensollen:
Thanks for having a look at this @emdupre ! :+1: 
>My biggest question is whether we should show the supplied functional niimg under the ROIs rather than an anatomical image

Good question, I'm still not sure what we should plot for this report. So far I was more focused on extracting and displaying information for regions, and I didn't spend too much time thinking about the plot itself. But I totally agree that we should also show the functional niimg if provided to the masker. 

Maybe something like this :question: 

label-masker

bthirion:
I'm not sure I want this feature: 1) it could lead to non-negligible computational costs simply for reporting 2) the choice of mean_img sounds arbitrary to me ; 3) this mean image is not part of the model (it is not stored in the trained masker), so I'm afraid of it being misleading.

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -49,13 +50,26 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
...
+        Label used in labels_img to represent background.
+        Warning: This value must be consisent with label values and
+        image provided.
+        Default=0.
 
     mask_img : Niimg-like object, optional
         See http://nilearn.github.io/manipulating_images/input_output.html
         Mask to apply to regions before extracting signals.
 
+    resolution : float, optional
```
bthirion, line 69: What is this ?

emdupre:
> I'm not sure I want this feature: 1) it could lead to non-negligible computational costs simply for reporting 2) the choice of mean_img sounds arbitrary to me;

For `NiftiMasker` we arbitrarily select the middle image [here](https://github.com/nilearn/nilearn/blob/13b98cfd057099890a2e7327be9d027f692b8a1b/nilearn/input_data/nifti_masker.py#L285-L289). I imagine we could do the same thing in this case ?

>  3) this mean image is not part of the model (it is not stored in the trained masker), so I'm afraid of it being misleading.

I think the same is true with the `NiftiMasker` report, but the idea of the the report was to show the overlap between the input data and the mask to help users identify common pain points (like a non-aligned image). I'd imagine we would want the same behavior here, as you could have ROIs that are not well-sampled because of misalignment ?

In my understanding, the motivation for the reports is to help users quickly diagnose common pain points. I worry that mis-alignment is one of the largest of these, so I'd like to see some way that it can be addressed, here !

nilearn/input_data/nifti_labels_masker.py
NicolasGensollen: @bthirion In order to compute the volume in mm^3 for each region I need to know what is the resolution of the atlas provided. I'm not sure there is another way to get this information than having the user providing it.

bthirion: No, you get that from the image affine (we have code in Nilearn that already does that).

bthirion:
@emdupre you convinced me.

nilearn/input_data/nifti_labels_masker.py
NicolasGensollen: @bthirion Absolutely! Sorry for my stupid comment.

NicolasGensollen:
I improved the report a little bit by including the functional image in the plot if it was provided to `fit`. In this situation, the ROI are plotted as contours:

```python
from nilearn import datasets
from nilearn.input_data import NiftiLabelsMasker
atlas = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
data = datasets.fetch_development_fmri(n_subjects=1)
masker = NiftiLabelsMasker(labels_img=atlas.maps,
                           labels=atlas.labels)
masker.fit(data.func[0])
report
```
report1

If nothing is provided to `fit`, then the report shows only the label image using `plot_roi` and displays a warning message as in the `NiftiMasker` report:

```python
masker = NiftiLabelsMasker(labels_img=atlas.maps,
                           labels=atlas.labels)
masker.fit()
report
```

report2

There is also the case where the user provided a mask. In this case, I'm not sure whether we should plot it or not. So far, I plot its contour in green on top of whatever image was built by the masker, but this might not be the most aesthetic choice...

```python
masker = NiftiLabelsMasker(labels_img=atlas.maps,
                           labels=atlas.labels,
                           mask_img=haxby_dataset.mask)
masker.fit(data.func[0])
report
```
report3

Finally, unlike the `NiftiMasker`, some of the resampling is done in `transform` rather than in `fit` which can result in inaccurate reports since they rely only on information extracted in `fit`.

https://github.com/nilearn/nilearn/blob/fc519b1919f50971af4ada72d5f997166cfb995e/nilearn/input_data/nifti_labels_masker.py#L209-L212

Is there a reason for that or should we have all the resampling done in `fit`?

bthirion:
> Is there a reason for that or should we have all the resampling done in `fit`?

If I'm not mistaken, doing the resampling at transform time makes it possible to transform images that have a different affine/fov than those used during fit, yet with reasonable computation time.

bthirion:
> So far, I plot its contour in green on top of whatever image was built by the masker

I think that this is a good choice. No better suggestion here.

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -49,8 +50,17 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
         See http://nilearn.github.io/manipulating_images/input_output.html
         Region definitions, as one image of labels.
 
+    labels : list of str, optional
+        Full labels corresponding to the labels image. This is used
+        to improve reporting quality if provided.
+        Warning: The labels must be consistent with the label
```
bthirion, line 56: But we need to check that actively.

bthirion (commented):
The code is a bit cluttered with the Codecov issues . Are these real ?

nilearn/input_data/nifti_labels_masker.py
NicolasGensollen: So far I am checking that the number of provided string labels matches the number of integer labels in the label image, but I don't see a way to check that they weren't shuffled.

bthirion: Sorry I had missed that. LGTM

bthirion (commented):
LGTM, then. Maybe add a whatsnew entry :-)

bthirion (approved):
LGTM, thx !

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -184,7 +204,107 @@ def __init__(self, labels_img, background_label=0, mask_img=None,
...
+                                            cmap='CMRmap_r')
+                display.add_contours(labels_image,
+                                     filled=False,
+                                     linewidths=3)
+
+            # Otherwise, simply plot the ROI of the label image
+            # and give a warning to the user
+            else:
+                msg = ("No image provided to fit in NiftiLabelsMasker. "
+                       "Plotting ROI of label image for reporting.")
```
emdupre, line 290: I think we could be a bit clearer in this warning. Maybe :point_down: 

```suggestion
                msg = ("No image provided to fit in NiftiLabelsMasker. "
                       "Plotting ROIs of label image on the MNI152Template for reporting.")
```

emdupre:
> There is also the case where the user provided a mask. In this case, I'm not sure whether we should plot it or not. So far, I plot its contour in green on top of whatever image was built by the masker, but this might not be the most aesthetic choice...

> I think that this is a good choice. No better suggestion here.

The aesthetics can always be improved :wink: So agreed that this would make sense for now !

> Finally, unlike the NiftiMasker, some of the resampling is done in transform rather than in fit which can result in inaccurate reports since they rely only on information extracted in fit.

I am somewhat concerned, though, about providing potentially inaccurate information in the reports if the mask is not resampled until calling `transform` method. I'd actually lean against showing it, in that case. The reporting won't be consistent with the NiftiMasker, but it at least highlights the key issue for NiftiLabelsMasker (i.e., the location of the labels). WDYT ?

NicolasGensollen:
Thanks for the reviews @emdupre ! :+1: 

>I am somewhat concerned, though, about providing potentially inaccurate information in the reports if the mask is not resampled until calling transform method. I'd actually lean against showing it, in that case. 

I would agree with not showing the mask in this case, but actually, both the mask and the label image can be resampled at `transform` time, here:

https://github.com/nilearn/nilearn/blob/fc519b1919f50971af4ada72d5f997166cfb995e/nilearn/input_data/nifti_labels_masker.py#L286-L295

and here:

https://github.com/nilearn/nilearn/blob/fc519b1919f50971af4ada72d5f997166cfb995e/nilearn/input_data/nifti_labels_masker.py#L296-L304

I'm not sure what the best solution would be, but I see the following options:

- Move all resampling to `fit` 
:heavy_plus_sign: the report is exact and we gain consistency with what is done in the `NiftiMasker`. 
:heavy_minus_sign: we lose support for transforming images with different fov/affines than those used during fitting
- Record/update report data with what is done in `transform` in the same way as for `fit`. 
:heavy_plus_sign: If the user generate the report after calling `fit` and `transform`, the correct information will be provided whatever resampling combinations.
:heavy_minus_sign: I don't think this is the original purpose/philosophy of the report which, as I understand it, provides a way to check that the masker is correctly configured (`fit` time) before using it (`transform` time), but at least you can check afterwards that everything looks good...
- Keep it like this and accept that the report can potentially show inaccurate information if additional resampling was done in `transform`.

I think I would lean toward the second option since it seems like the one which doesn't introduce potential inaccuracies.
WDYT?

nilearn/input_data/nifti_labels_masker.py
NicolasGensollen: Thanks!

NicolasGensollen:
I realize this PR is still lacking an example showing the report. It doesn't look like this would easily fit in current examples using the `NiftiLabelsMasker` because they are usually aiming at showing something else than the masker capabilities themselves.
I am wondering whether we should add a very simple example dedicated to how to use/understand the `NiftiLabelsMasker`, exactly like what is currently done for the `NiftiMasker` [here](http://nilearn.github.io/auto_examples/06_manipulating_images/plot_nifti_simple.html#sphx-glr-auto-examples-06-manipulating-images-plot-nifti-simple-py).

bthirion:
> I realize this PR is still lacking an example showing the report. It doesn't look like this would easily fit in current examples using the `NiftiLabelsMasker` because they are usually aiming at showing something else than the masker capabilities themselves.
> I am wondering whether we should add a very simple example dedicated to how to use/understand the `NiftiLabelsMasker`, exactly like what is currently done for the `NiftiMasker` [here](http://nilearn.github.io/auto_examples/06_manipulating_images/plot_nifti_simple.html#sphx-glr-auto-examples-06-manipulating-images-plot-nifti-simple-py).

+1

bthirion:
Reagrding the discussion. 
* I think that the report should depend only on stuff done in fit()
* I think that resampling at transform time is OK. 
* normally, in terms of image overlay, showing (i1 and I2) or (i1 and resample_to_image(i2, i1)) should amount to the same (up to interpolation inaccuracies), because the viewer takes the affine of I2 into account in any case. 
* Finally, for the time being it is OK to have different reports for NiftiMasker and NiftiLabelsMasker.

examples/06_manipulating_images/plot_nifti_labels_simple.py
```diff
@@ -0,0 +1,61 @@
+"""
+Simple example of NiftiLabelsMasker use
```
bthirion, line 2: This is technically right, but I think that we should strive to label and introduce examples with a user-oriented perspective: what this shows is how to check the regions provided by an atlas, and then the correspondence with a user-supplied image.

NicolasGensollen:
Link to the example in the doc: [link](https://6857-1235740-gh.circle-artifacts.com/0/doc/_build/html/auto_examples/06_manipulating_images/plot_nifti_labels_simple.html#sphx-glr-auto-examples-06-manipulating-images-plot-nifti-labels-simple-py)

NicolasGensollen:
@emdupre I don't think we have found a solution for how to handle the resampling done at `transform` time with respect to the masker report. I'm not really sure what to do here. :thinking:  
I certainly don't want to remove the ability to resample in `transform` as this is apparently a desired feature. 
I also think the report should only display information related to `fit`. 
So maybe we should just add a note in the report saying that the report shows the masker information at `fit` time...
WDYT?

emdupre:
> I also think the report should only display information related to fit.
So maybe we should just add a note in the report saying that the report shows the masker information at fit time...
WDYT?

I think that's a good route forward :) Obviously we can have it in the console, but it might make sense directly on the report, too, since the report is intended to be shared. WDYT ?

emdupre:
I think that LGTM ! Thank you, @NicolasGensollen  !! Excited to see this added :sparkles: :rocket:

NicolasGensollen:
Thanks for the reviews! :+1: 
Merging...

Issue #2637, opened 2020-12-14
NiftiMapsMasker and NiftiLabelsMasker cannot transform 3D images


jeromedockes:
documentation suggests they can (says "imgs: 3D/4D Niimg-like object").
example:
```
from nilearn import datasets
from nilearn.input_data import NiftiMapsMasker, NiftiLabelsMasker


img = datasets.load_mni152_template()

atlas_img = datasets.fetch_atlas_destrieux_2009()["maps"]
labels_masker = NiftiLabelsMasker(atlas_img)

prob_atlas_img = datasets.fetch_atlas_harvard_oxford("cort-prob-2mm")["maps"]
maps_masker = NiftiMapsMasker(prob_atlas_img)

# OK
labels_masker.fit_transform([img])
maps_masker.fit_transform([img])

# both lines below fail
labels_masker.fit_transform(img)
maps_masker.fit_transform(img)
```

jeromedockes:
fit on 3d images is always fine, fit_transform and transform fail. more details:
```
import itertools

from sklearn.base import clone

from nilearn import datasets
from nilearn.input_data import NiftiMapsMasker, NiftiLabelsMasker


img_3d = datasets.load_mni152_template()
images = {"3d": img_3d, "4d": [img_3d]}

atlas_img = datasets.fetch_atlas_destrieux_2009()["maps"]
labels_masker = NiftiLabelsMasker(atlas_img)

prob_atlas_img = datasets.fetch_atlas_harvard_oxford("cort-prob-2mm")["maps"]
maps_masker = NiftiMapsMasker(prob_atlas_img)

for masker, img in itertools.product(
    [labels_masker, maps_masker], images.keys()
):
    masker_name = masker.__class__.__name__

    masker = clone(masker)
    print(f"{masker_name} fit on {img}: ", end="")
    try:
        masker.fit(images[img])
        print("OK")
    except Exception:
        print("Fail")

    masker = clone(masker)
    print(f"{masker_name} fit_transform on {img}: ", end="")
    try:
        masker.fit_transform(images[img])
        print("OK")
    except Exception:
        print("Fail")

    masker = clone(masker)
    masker.fit(images["4d"])
    print(f"{masker_name} transform {img}: ", end="")
    try:
        masker.transform(images[img])
        print("OK")
    except Exception:
        print("Fail")
```

jeromedockes:
[...]

PR #2708, opened 2021-02-20
Solving "NiftiMapsMasker and NiftiLabelsMasker cannot transform 3D images"

achamma723:
In this pull request, I handled the issue of "NiftiMapsMasker and NiftiLabelsMasker cannot transform 3D images" https://github.com/nilearn/nilearn/issues/2637. The problem was when using a 3d image as the transform function for both maskers accepts only 4d images. So, in the case of a 3d image:

* The transform function was always testing using the check for the 4d imag. So the solution I proposed was to include a try-except concept where the function will try first the 4d checker then if it raises an exception it will forward to the 3d checker. Also this problem appeared in the fit function of the maps_masker.
* A test for the 3d iamges is included now in the pytest files.

### DIFF
--- a/nilearn/_utils/niimg_conversions.py
+++ b/nilearn/_utils/niimg_conversions.py
@@ -277,25 +277,21 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
         if return_iterator:
             return _iter_check_niimg(niimg, ensure_ndim=ensure_ndim,
                                      dtype=dtype)
-
         return concat_niimgs(niimg, ensure_ndim=ensure_ndim, dtype=dtype)
 
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
-    
     if ensure_ndim == 3 and len(niimg.shape) == 4 and niimg.shape[3] == 1:
         # "squeeze" the image.
         data = _safe_get_data(niimg)
         affine = niimg.affine
         niimg = new_img_like(niimg, data[:, :, :, 0], affine)
-
     if atleast_4d and len(niimg.shape) == 3:
         data = _get_data(niimg).view()
         data.shape = data.shape + (1, )
         niimg = new_img_like(niimg, data, niimg.affine)
 
     if ensure_ndim is not None and len(niimg.shape) != ensure_ndim:
-
         raise DimensionError(len(niimg.shape), ensure_ndim)
 
     if return_iterator:
--- a/nilearn/_utils/niimg_conversions.py
+++ b/nilearn/_utils/niimg_conversions.py
@@ -277,22 +277,25 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
         if return_iterator:
             return _iter_check_niimg(niimg, ensure_ndim=ensure_ndim,
                                      dtype=dtype)
+
         return concat_niimgs(niimg, ensure_ndim=ensure_ndim, dtype=dtype)
 
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
-
+    
     if ensure_ndim == 3 and len(niimg.shape) == 4 and niimg.shape[3] == 1:
         # "squeeze" the image.
         data = _safe_get_data(niimg)
         affine = niimg.affine
         niimg = new_img_like(niimg, data[:, :, :, 0], affine)
+
     if atleast_4d and len(niimg.shape) == 3:
         data = _get_data(niimg).view()
         data.shape = data.shape + (1, )
         niimg = new_img_like(niimg, data, niimg.affine)
 
     if ensure_ndim is not None and len(niimg.shape) != ensure_ndim:
+
         raise DimensionError(len(niimg.shape), ensure_ndim)
 
     if return_iterator:
--- a/nilearn/_utils/niimg_conversions.py
+++ b/nilearn/_utils/niimg_conversions.py
@@ -281,6 +281,7 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
 
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
+    
     if ensure_ndim == 3 and len(niimg.shape) == 4 and niimg.shape[3] == 1:
         # "squeeze" the image.
         data = _safe_get_data(niimg)
--- a/nilearn/_utils/niimg_conversions.py
+++ b/nilearn/_utils/niimg_conversions.py
@@ -281,7 +281,7 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
 
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
-    
+
     if ensure_ndim == 3 and len(niimg.shape) == 4 and niimg.shape[3] == 1:
         # "squeeze" the image.
         data = _safe_get_data(niimg)
--- a/nilearn/input_data/base_masker.py
+++ b/nilearn/input_data/base_masker.py
@@ -179,7 +179,7 @@ class BaseMasker(BaseEstimator, TransformerMixin, CacheMixin):
 
         """
         self._check_fitted()
-
+        
         return self.transform_single_imgs(imgs, confounds)
 
     def fit_transform(self, X, y=None, confounds=None, **fit_params):
--- a/nilearn/input_data/base_masker.py
+++ b/nilearn/input_data/base_masker.py
@@ -179,7 +179,6 @@ class BaseMasker(BaseEstimator, TransformerMixin, CacheMixin):
 
         """
         self._check_fitted()
-        
         return self.transform_single_imgs(imgs, confounds)
 
     def fit_transform(self, X, y=None, confounds=None, **fit_params):
--- a/nilearn/input_data/base_masker.py
+++ b/nilearn/input_data/base_masker.py
@@ -18,6 +18,7 @@ from .. import signal
 from .. import _utils
 from .._utils.cache_mixin import CacheMixin, cache
 from .._utils.class_inspect import enclosing_scope_name
+from nilearn.image import high_variance_confounds
 
 
 def filter_and_extract(imgs, extraction_function,
@@ -179,8 +180,23 @@ class BaseMasker(BaseEstimator, TransformerMixin, CacheMixin):
 
         """
         self._check_fitted()
-        
-        return self.transform_single_imgs(imgs, confounds)
+
+        if confounds is None and not self.high_variance_confounds:
+            return self.transform_single_imgs(imgs, confounds)
+
+        # Compute high variance confounds if requested
+        all_confounds = []
+        if self.high_variance_confounds:
+            hv_confounds = self._cache(
+                high_variance_confounds)(imgs)
+            all_confounds.append(hv_confounds)
+        if confounds is not None:
+            if isinstance(confounds, list):
+                all_confounds += confounds
+            else:
+                all_confounds.append(confounds)
+
+        return self.transform_single_imgs(imgs, all_confounds)
 
     def fit_transform(self, X, y=None, confounds=None, **fit_params):
         """Fit to data, then transform it
--- a/nilearn/input_data/base_masker.py
+++ b/nilearn/input_data/base_masker.py
@@ -179,6 +179,7 @@ class BaseMasker(BaseEstimator, TransformerMixin, CacheMixin):
 
         """
         self._check_fitted()
+        
         return self.transform_single_imgs(imgs, confounds)
 
     def fit_transform(self, X, y=None, confounds=None, **fit_params):
--- a/nilearn/input_data/nifti_labels_masker.py
+++ b/nilearn/input_data/nifti_labels_masker.py
@@ -275,12 +275,13 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
         # We handle the resampling of labels separately because the affine of
         # the labels image should not impact the extraction of the signal.
 
-        # If a 3D image is given, it will be appended to an empty list
+        # If a 3D image is given, an extra dimension will be added
         is_3d = False
         
         if type(imgs) != list:
-            imgs = [imgs]
-            is_3d = True
+            if len(imgs.shape) < 4:
+                imgs = np.expand_dims(imgs, 0)
+                is_3d = True
 
         if not hasattr(self, '_resampled_labels_img_'):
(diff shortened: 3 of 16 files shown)


nilearn/input_data/nifti_labels_masker.py
```diff
@@ -273,6 +275,14 @@ def transform_single_imgs(self, imgs, confounds=None):
         # We handle the resampling of labels separately because the affine of
         # the labels image should not impact the extraction of the signal.
 
+        # If a 3D image is given, an extra dimension will be added
+        is_3d = False
+        
+        if type(imgs) != list:
+            if len(imgs.shape) < 4:
```
NicolasGensollen, line 282: `imgs` can also be a string in case we pass a filename in. In this case, this will fail because strings have no `shape` argument.

NicolasGensollen (changes requested):
Thanks for the PR @achamma723 ! :+1: 
In addition to my comments below, it'b great if you could add a test failing prior to your fix and passing with it.

nilearn/input_data/nifti_labels_masker.py
achamma723: Hello nicolas,
In fact this was my first PR before discovering the fact of some local tests that i can use before committing the changes. Therefore, i did another PR while adding an extra parameter called is_3d to be indicated. I still have a small issue to check before submitting my last changes

nilearn/_utils/niimg_conversions.py
```diff
@@ -277,22 +277,25 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
         if return_iterator:
             return _iter_check_niimg(niimg, ensure_ndim=ensure_ndim,
                                      dtype=dtype)
+
         return concat_niimgs(niimg, ensure_ndim=ensure_ndim, dtype=dtype)
 
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
-
+
```
bthirion, line 285: Please do not include such lines.

nilearn/_utils/niimg_conversions.py
```diff
@@ -277,22 +277,25 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
...
         affine = niimg.affine
         niimg = new_img_like(niimg, data[:, :, :, 0], affine)
+
     if atleast_4d and len(niimg.shape) == 3:
         data = _get_data(niimg).view()
         data.shape = data.shape + (1, )
         niimg = new_img_like(niimg, data, niimg.affine)
 
     if ensure_ndim is not None and len(niimg.shape) != ensure_ndim:
+
```
bthirion, line 298: This line is useless.

nilearn/input_data/base_masker.py
```diff
@@ -179,7 +179,7 @@ def transform(self, imgs, confounds=None):
 
         """
         self._check_fitted()
-
+
```
bthirion, line 182: This type of lines messes up the identation.

nilearn/input_data/nifti_maps_masker.py
```diff
@@ -173,7 +173,7 @@ def __init__(self, maps_img, mask_img=None,
                 "has been provided.\nSet resampling_target to something else"
                 " or provide a mask.")
 
-    def fit(self, X=None, y=None):
+    def fit(self, X=None, y=None, is_3d=False):
```
bthirion, line 176: we should always avoid adding arguments to fit(). Why do we need that one ?

nilearn/input_data/nifti_maps_masker.py
```diff
@@ -236,9 +240,10 @@ def fit_transform(self, imgs, confounds=None):
         """Prepare and perform signal extraction.
 
         """
+
         return self.fit().transform(imgs, confounds=confounds)
 
-    def transform_single_imgs(self, imgs, confounds=None):
+    def transform_single_imgs(self, imgs, confounds=None, is_3d=False):
```
bthirion, line 246: Same here.

nilearn/input_data/nifti_maps_masker.py
```diff
@@ -346,7 +357,7 @@ def transform_single_imgs(self, imgs, confounds=None):
                 # kwargs
                 verbose=self.verbose)
         self.labels_ = labels_
-
+
```
bthirion, line 360: remove empty spaces.

bthirion:
Thx for addressing this !

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -278,7 +279,10 @@ def transform_single_imgs(self, imgs, confounds=None):
         if not hasattr(self, '_resampled_mask_img'):
             self._resampled_mask_img = self.mask_img_
         if self.resampling_target == "data":
-            imgs_ = _utils.check_niimg_4d(imgs)
+            try:
+                imgs_ = _utils.check_niimg_4d(imgs)
+            except:
```
NicolasGensollen, line 284: I would catch the specific error you expect here (`DimensionError`).
Besides, I think using `check_niimg` would take care of that for you:

```python
imgs_ = _utils.check_niimg(imgs)
```
If you want to enforce 4D, you can use:

```python
imgs_ = _utils.check_niimg(imgs, atleast_4d=True)
```

nilearn/input_data/nifti_maps_masker.py
```diff
@@ -184,12 +184,16 @@ def fit(self, X=None, y=None):
...
                                        shorten=(not self.verbose)),
                    verbose=self.verbose)
-
-        self.maps_img_ = _utils.check_niimg_4d(self.maps_img, dtype=self.dtype)
-        self.maps_img_ = image.clean_img(self.maps_img_, detrend=False,
-                                         standardize=False,
-                                         ensure_finite=True)
-
+        try:
+            self.maps_img_ = _utils.check_niimg_4d(self.maps_img, dtype=self.dtype)
```
NicolasGensollen, line 188: Same comment here

nilearn/input_data/nifti_maps_masker.py
```diff
@@ -236,6 +240,7 @@ def fit_transform(self, imgs, confounds=None):
         """Prepare and perform signal extraction.
 
         """
+
```
NicolasGensollen, line 243: Not sure why you added a line here

NicolasGensollen (changes requested):
Thanks for working on this @achamma723 !
I left a few comments below. I think we can simplify the try catch pattern you implemented using `check_niimg`.

nilearn/input_data/nifti_labels_masker.py
achamma723: Hello Nicolas, 
you mean the check_niimg is able to recognize the dimensionality (3d or 4d) or the try-catch is still needed?

nilearn/input_data/nifti_maps_masker.py
```diff
@@ -184,12 +184,12 @@ def fit(self, X=None, y=None):
                    _utils._repr_niimgs(self.maps_img,
                                        shorten=(not self.verbose)),
                    verbose=self.verbose)
-
-        self.maps_img_ = _utils.check_niimg_4d(self.maps_img, dtype=self.dtype)
+        self.maps_img_ = _utils.check_niimg(self.maps_img, dtype=self.dtype)
+        self.maps_img_ = np.expand_dims(self.maps_img_, 0) if len(
```
NicolasGensollen, line 188: I don't think this will work for several reasons:

- `np.exand_dims` expects an array as [input](https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html) and here it can receive a string. 
- `np.exand_dims` also returns an array, which will cause `clean_img` to raise a TypeError:

```python
import numpy as np
from nilearn._utils import testing, as_ndarray, data_gen
from nilearn.image import clean_img, get_data
affine = np.diag((4, 4, 4, 1))
n_regions = 3
shape3 = (2, 2, 2)

labels33_img = data_gen.generate_labeled_regions(shape3, n_regions,
                                                     affine=affine)
clean_img(np.expand_dims(get_data(labels33_img), 0))
```
```
<ipython-input-6-5f193dc4c2fc> in <module>
      8 labels33_img = data_gen.generate_labeled_regions(shape3, n_regions,
      9                                                      affine=affine)
---> 10 clean_img(np.expand_dims(get_data(labels33_img), 0))
```

- When calling `clean_img`, the signals in the last dimension are filtered, so I think you need to add a dimension "at the end" rather than "at the beginning".

Unless I'm missing something, I think `self.maps_img_ = _utils.check_niimg(self.maps_img, dtype=self.dtype, atleast_4d=True)` should do the trick here.

NicolasGensollen (commented):
@achamma723 it looks like the tests are passing which is a little bit surprising because I think the lines you added with `np.expand_dims` should fail (see my comment below). 
I didn't check extensively, so I might be wrong, or it means that the current tests are not covering this specific case properly. What do you think?

nilearn/input_data/nifti_maps_masker.py
achamma723: Hello @NicolasGensollen , Thank you for your review.
In fact when i was exploring the check_niimg function i discoverd the fact that it is loading the image and returning the loaded image in the case of a filename
    if ensure_ndim == 3 and len(niimg.shape) == 4 and niimg.shape[3] == 1:
        # "squeeze" the image.
        data = _safe_get_data(niimg)
        affine = niimg.affine
        niimg = new_img_like(niimg, data[:, :, :, 0], affine)
    if atleast_4d and len(niimg.shape) == 3:
        data = _get_data(niimg).view()
        data.shape = data.shape + (1, )
        niimg = new_img_like(niimg, data, niimg.affine)

    if ensure_ndim is not None and len(niimg.shape) != ensure_ndim:
        raise DimensionError(len(niimg.shape), ensure_ndim)

    if return_iterator:
        return (_index_img(niimg, i) for i in range(niimg.shape[3]))

    return niimg

Now, for the clean_img function, according to the parameters it can handle only 4d images, this is why in my last commit i added a condiitional operator on the length of the shape of the image (even loaded in the filename case). 
Now, in my new commit, i tested the case with at_least4d=True and it worked with all the tests (This may due to the fact that check_niimg with this parameter is adjusting the dimensionality of the image to 4d). 
Therefore, i'm comitting the changes now

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -368,3 +367,24 @@ def test_nifti_labels_masker_with_mask():
...
+                                                     affine=affine)
+    mask_img = nibabel.Nifti1Image(np.ones(shape3, dtype=np.int8),
+                           affine=affine)
+    epi_img1 = nibabel.Nifti1Image(np.ones(shape3),
+                           affine=affine)
+    epi_img2 = nibabel.Nifti1Image(np.ones(shape3),
+                           affine=affine)
+    masker = NiftiLabelsMasker(labels33_img, mask_img=mask_img)
+
+    epis = masker.fit_transform(epi_img1)
```
NicolasGensollen, line 387: You should check that the shapes of the outputs are correct. For the labels masker you should get a 2D array of shape `n_time_points x n_regions`. So, when you fit transform `epi_img1` you expect `epis` to be `1 x 3`, and when fit transforming `[epi_img1, epi_img2]` you expect `epis` to be `2 x 3`.

nilearn/input_data/tests/test_nifti_maps_masker.py
```diff
@@ -339,3 +339,24 @@ def test_standardization():
...
+    epi_img1 = nibabel.Nifti1Image(np.ones(shape3),
+                           affine=affine)
+    epi_img2 = nibabel.Nifti1Image(np.ones(shape3),
+                           affine=affine)
+    masker = NiftiMapsMasker(labels33_img, mask_img=mask_img)
+
+    epis = masker.fit_transform(epi_img1)
+    epis = masker.fit_transform([epi_img1, epi_img2])
+    # This is mostly a smoke test
+    assert len(epis) == 2
```
NicolasGensollen, line 362: Same here

nilearn/input_data/tests/test_nifti_labels_masker.py
```diff
@@ -368,3 +367,24 @@ def test_nifti_labels_masker_with_mask():
         masked_labels, resampling_target=None, mask_img=mask_img)
     masked_signals = masked_masker.fit().transform(fmri_img)
     assert np.allclose(signals, masked_signals)
+
+def test_3d_images():
+    # Test that the NiftiLabelsMasker works with 3D images
+    affine = np.diag((4, 4, 4, 1))
```
NicolasGensollen, line 373: Is there a reason you used this affine instead of the more classic `np.eye(4)`?

nilearn/_utils/niimg_conversions.py
```diff
@@ -281,7 +281,6 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
 
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
-
```
NicolasGensollen, line 284: Please restore this line. it is the only change you make to this file.

nilearn/input_data/base_masker.py
```diff
@@ -179,7 +179,6 @@ def transform(self, imgs, confounds=None):
 
         """
         self._check_fitted()
-
```
NicolasGensollen, line 182: Same here.

NicolasGensollen (changes requested):
Almost there @achamma723 ! :+1: 
I just have a few small modification requests (see comments below).

NicolasGensollen:
Thanks for addressing my comments @achamma723 !
It looks like you have some conflicts in `base_masker.py`

achamma723:
@NicolasGensollen Maybe someone changed something in base_maker during during my pull request?

achamma723:
@NicolasGensollen Sry, I didn't pull from the repository before pushing the changes. This is why

nilearn/input_data/tests/test_nifti_labels_masker.py
achamma723: @NicolasGensollen I used one of the tests already provided with the same affine then i found out that the affine by default is enough for testing so i changed it

nilearn/_utils/niimg_conversions.py
```diff
@@ -281,7 +281,7 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
...
+
```
NicolasGensollen, line 284: Could you remove the indent here?

NicolasGensollen (changes requested):
Looks good to me except the indented empty line (see comment). We should basically see this file disappear from the diff, which will make things cleaner.

NicolasGensollen (approved):
LGTM, thanks! :+1:

bthirion (approved):
LGTM, thx !

NicolasGensollen:
Merging, thanks!

PR #2709, opened 2021-02-20
Fixes numpy runtime warning in masking.py when img has NaNs.

### LINKED ISSUE #2163: Non-descriptive runtime warning when using NiftiMasker on beta images
0.5.2

What were you trying to do? 
Use NiftiMasker to read fMRI beta images that came from a first level analysis in SPM.

What actually happened?
Encoutered a non-descriptive RuntimeWarning that seems to be passed from numpy:
```
[NiftiMasker.fit] Loading data from E:\beta_0001.nii
[NiftiMasker.fit] Computing the mask
[NiftiMasker.fit] Resampling mask
[NiftiMasker.transform_single_imgs] Loading data from Nifti1Image(
shape=(79, 95, 79, 1),
affine=array([[  -2.,    0.,    0.,   78.],
       [   0.,    2.,    0., -112.],
       [   0.,    0.,    2.,  -70.],
       [   0.,    0.,    0.,    1.]])
)
[NiftiMasker.transform_single_imgs] Extracting region signals
[NiftiMasker.transform_single_imgs] Cleaning extracted signalsC:\Users\Felix\Anaconda3\lib\site-packages\numpy\lib\function_base.py:3405: RuntimeWarning: Invalid value encountered in median
  r = func(a, **kwargs)
``` 

Code to reproduce:
```python

masker = NiftiMasker(verbose=True)

beta_img = 'E:\\beta_0001.nii'
t_img = 'E:\\spmT_0001.nii'

X_beta = masker.fit_transform(beta_img)
X_t = masker.fit_transform(t_img)
``` 

I have attached both the beta- and the t-nifti images. Strangely, the issue doesn't occur for the t-image, but only for the beta-image, although they come from the same participant. Any ideas what the issue could be? Thanks!
[images.zip](https://github.com/nilearn/nilearn/files/3707175/images.zip)

fkalbe:
Fixes #2163 where loading an SPM image containing NaNs would trigger a runtime warning in older versions of numpy.

### DIFF
--- a/nilearn/masking.py
+++ b/nilearn/masking.py
@@ -434,12 +434,12 @@ def compute_background_mask(data_imgs, border_size=2,
                                                 target_shape=target_shape,
                                                 smooth=False)
 
-    background = np.median(get_border_data(data, border_size))
-    if np.isnan(background):
+    if np.isnan(get_border_data(data, border_size)).any():
         # We absolutely need to catter for NaNs as a background:
         # SPM does that by default
         mask = np.logical_not(np.isnan(data))
     else:
+        background = np.median(get_border_data(data, border_size))
         mask = data != background
 
     mask, affine = _post_process_mask(mask, affine, opening=opening,


bthirion:
Thx ! 
I'm simply wondering whether we should not consider the case where tharee are Nans, but also lots of small values, so that you would like the mask to exclude small values, beyond NaNs ? ----which was possible with the previously implemented heuristic.

NicolasGensollen:
>which was possible with the previously implemented heuristic.

I don't understand how the previous implementation would enable this. I think the fix is making sure the median is computed only if there are no Nan value to avoid getting a numpy warning with pre 1.17 versions. But unless I'm missing something, the final behaviour is exactly the same.

Before 1.17:

```
>>> import numpy as np
>>> np.__version__
'1.16.0'
>>> data = np.arange(0,100, dtype=float)
>>> data[80] = np.nan
>>> np.median(data)
/home/nicolas/anaconda3/envs/py35/lib/python3.5/site-packages/numpy/lib/function_base.py:4011: RuntimeWarning: Invalid value encountered in median
  r = func(a, **kwargs)
nan
```

After 1.17:

```
>>> import numpy as np
>>> np.__version__
'1.17.0'
>>> data = np.arange(0,100, dtype=float)
>>> data[80] = np.nan
>>> np.median(data)
nan
```

fkalbe:
I tend to agree with @NicolasGensollen that this change should not lead to any different behavior. In my understanding, `np.median()` will return NaN as soon as any NaNs are present in the data. So whether any small values are additionally present in the data should not make a difference. Could you maybe come up with an example?

bthirion:
OK, sorry for my wrong assessment.

bthirion (approved):

NicolasGensollen (approved):
LGTM, thanks for addressing this! :+1:

NicolasGensollen:
Merging...

Issue #2704, opened 2021-02-18
FirstLevelModel fit method could raise warning when confounds and design_matrices are specified together.


kbonna:
Hi,

I recently used `FirstLevelModel` in my project and discovered a mistake I made that was quite hard to detect. I was using custom design matrix and fitted model instance like this:
```python
first_level_model.fit(imgs, confounds=confounds, design_matrices=dm)
``` 
and didn't realize that in this case confounds would not be used. I noticed that few weeks later when I tried different confounds and see no difference in the images (because in both cases they were not used). I then read the documentation about `design_matrices` arg and found this statement: "If given it takes precedence over events and confounds.". 

I thought it would be nice to add warning if user call fit method just like me to let him know that something is wrong. It's just a few lines of code, but maybe will help someone who didn't read the documentation carefully like me.

Tell me what you think.

NicolasGensollen:
Hi @kbonna 
Thanks for raising the point. I agree that a simple warning could be useful in this situation. 
Do you want to make a PR?

kbonna:
Sure

achamma723:
Hello @NicolasGensollen , I addressed this issue in a new pull request.

PR #2710, opened 2021-02-22
Raise warning for FirstLevelModel fit when confounds and design_matrices are both specified

achamma723:
- Adding a warning in the fit function when both the design matrices and (confounds and events) are provided in #2704

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -362,6 +362,9 @@ class FirstLevelModel(BaseGLM):
             takes precedence over events and confounds.
 
         """
+        # Raise a warning if both design_matrices and confounds are provided
+        if design_matrices is not None and confounds is not None:
+            warn('If the design matrices are given, it takes precedence over events and confounds.')
         # Local import to prevent circular imports
         from nilearn.input_data import NiftiMasker  # noqa
 
@@ -519,8 +522,6 @@ class FirstLevelModel(BaseGLM):
         if self.verbose > 0:
             sys.stderr.write("\nComputation of %d runs done in %i seconds\n\n"
                              % (n_runs, time.time() - t0))
-        if design_matrices is not None and confounds is not None:
-            warn('If the design matrices are given, it takes precedence over events and confounds.')
         return self
 
     def compute_contrast(self, contrast_def, stat_type=None,
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -519,7 +519,7 @@ class FirstLevelModel(BaseGLM):
         if self.verbose > 0:
             sys.stderr.write("\nComputation of %d runs done in %i seconds\n\n"
                              % (n_runs, time.time() - t0))
-
+        warn('If the design matrices are given, it takes precedence over events and confounds.')
         return self
 
     def compute_contrast(self, contrast_def, stat_type=None,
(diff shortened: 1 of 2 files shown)


nilearn/glm/first_level/first_level.py
```diff
@@ -519,7 +519,8 @@ def fit(self, run_imgs, events=None, confounds=None,
         if self.verbose > 0:
             sys.stderr.write("\nComputation of %d runs done in %i seconds\n\n"
                              % (n_runs, time.time() - t0))
-
+        if design_matrices is not None and confounds is not None:
```
NicolasGensollen, line 522: I would expect it to be given at the beginning of the function rather than at the end, but otherwise the warning looks good.

NicolasGensollen (changes requested):
Thanks for addressing this @achamma723 !
Could you add a test checking that the warning is given correctly when providing a design matrix and confounds?

NicolasGensollen:
Linking #2704

achamma723:
@NicolasGensollen Can you tell more about this error?

nilearn/glm/tests/test_first_level.py
```diff
@@ -630,3 +630,22 @@ def test_first_level_predictions_r_square():
 
     r_square_2d = model.masker_.transform(r_square_3d)
     assert_array_less(0., r_square_2d)
+
+
+def test_warning_fit():
```
NicolasGensollen, line 635: I think this is too complicated for a simple warning check. I would be in favor of just checking this in `test_fmri_inputs` for example by adding something like:

```python
with pytest.warns(UserWarning,
                  match="If the design matrices are given"):
        FirstLevelModel(mask_img=mask).fit(fi, 
                                           design_matrices=d,
                                           confounds=conf)
```
WDYT?

NicolasGensollen (changes requested):
I think we can simplify the test a lot (see my comment below). 
I haven't checked the Azure error you're getting yet, but I think we can update the test before debugging.

achamma723:
@NicolasGensollen Everything is good now. Thanks for your remarks!

nilearn/glm/first_level/first_level.py
```diff
@@ -362,6 +362,9 @@ def fit(self, run_imgs, events=None, confounds=None,
             takes precedence over events and confounds.
 
         """
+        # Raise a warning if both design_matrices and confounds are provided
+        if design_matrices is not None and confounds is not None:
+            warn('If the design matrices are given, it takes precedence over events and confounds.')
```
tsalo, line 367:
```suggestion
            warn('If design matrices are given, they take precedence over events and confounds.')
```

I'd lean toward using simpler words though. Perhaps `"If design matrices are supplied, confounds and events will be ignored."`.

nilearn/glm/first_level/first_level.py
```diff
@@ -362,6 +362,9 @@ def fit(self, run_imgs, events=None, confounds=None,
             takes precedence over events and confounds.
 
         """
+        # Raise a warning if both design_matrices and confounds are provided
+        if design_matrices is not None and confounds is not None:
```
tsalo, line 366:
```suggestion
        if design_matrices is not None and (confounds is not None or events is not None):
```
Based on the warning, it seems like both confounds and events should be checked here.

tsalo (changes requested):
Things look great. I just have a couple of minor requests. Additionally, would you mind (1) simplifying the PR title and (2) linking directly to #2704 with a [closing keyword](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (e.g., "Closes ")?

tsalo (approved):
The changes look good to me. Will need to wait for @NicolasGensollen to re-review before merging though.

NicolasGensollen (approved):
This looks good to me too! :tada: 
Thanks for working on this @achamma723 and thanks for the reviews @tsalo ! :+1:

bthirion (approved):
LGTM, thx.

NicolasGensollen:
Thanks everybody! Merging...

PR #2715, opened 2021-03-02
[ENH] Add function to write out BIDS Derivatives-style model results

### LINKED ISSUE #2683: Saving GLM results to BIDS-like files
It would be great to have a function that saves GLM outputs to files, preferably with some BIDS-like formatting. The BIDS Models outputs are not well specified, but we can use [`fitlins` outputs](https://github.com/poldracklab/fitlins/blob/01eeaefa831123913ae1371c0db2e8acc3075149/fitlins/workflows/base.py#L110-L130) as a template.

### Benefits to the change
Adding the ability to save BIDS-like derivatives should make it easier to both share GLM results on NeuroVault or OpenNeuro and to use GLM results in image-based meta-analyses. The associated metadata files should also make the analyses more reproducible.

Perhaps something like the following?

```python
save_glm_results(
    model=model,
    contrasts=["StopSuccess - Go", "StopFailure - Go"],
    out_dir="/path/to/outputs/",
)
```

And this would produce a set of files in the output directory, like:

```
dataset_description.json
contrast-StopSuccessMinusGo_stat-effect_statmap.nii.gz
contrast-StopSuccessMinusGo_stat-variance_statmap.nii.gz
contrast-StopSuccessMinusGo_stat-z_statmap.nii.gz
contrast-StopFailureMinusGo_stat-effect_statmap.nii.gz
contrast-StopFailureMinusGo_stat-variance_statmap.nii.gz
contrast-StopFailureMinusGo_stat-z_statmap.nii.gz
```

tsalo:
Closes #2683.

The goal of the new function(s) is not to implement a BIDS App, but to generate BIDS-compatible outputs that can be collected and integrated automatically within a BIDS App like fitlins. The output files won't use entities from the source files; they'll only have the _new_ entities specific to the model. We also can't define the full derivative directory tree because the function(s) will be tied to an individual model, so we will output any top-level files (like `dataset_description.json`) within the one output directory.

If anyone wants to see what the output looks like, check out [this notebook](https://github.com/tsalo/misc-notebooks/blob/main/nilearn_glm.ipynb), which I try to run and push when I make substantive changes in this PR.

- Create `nilearn.interfaces.bids.save_glm_to_bids()`, which will save first- and second-level model results to BIDS-like files. 

To do:

- [ ] Check outputs with BIDS Derivatives folks
- [x] Write tests
- [x] Finish docstrings
- [x] Generate image-specific json files
- [x] Generate dataset_description file
- [ ] Stop showing figures generated in function
- [x] Convert contrast names to camelCase (thanks @akimbler!)

Open questions:

- [ ] What model- or map-level metadata should be retained?
- [ ] Can any information about the original data files be extracted from models?
    - I know that nibabel images can sometimes include paths to the original files, so if the input images are still accessible then the source files can at least sometimes be inferred.


nilearn/reporting/glm_reporter.py
```diff
@@ -967,3 +968,144 @@ def _dataframe_to_html(df, precision, **kwargs):
...
+            run_name = i_run + 1
+            dm_file = os.path.join(
+                out_dir,
+                "{}run-{}_design.tsv".format(prefix, run_name),
+            )
+            dm.to_csv(dm_file, sep="\t", index=False)
+
+            dm_fig_file = os.path.join(
+                out_dir,
+                "{}run-{}_design.svg".format(prefix, run_name),
```
bthirion, line 1014: Just curious: is the use of svg recommended by BIDS ?

nilearn/reporting/glm_reporter.py
```diff
@@ -967,3 +968,144 @@ def _dataframe_to_html(df, precision, **kwargs):
...
+        "high_pass",
+        "target_shape",
+        "signal_scaling",
+        "drift_order",
+        "scaling_axis",
+        "smoothing_fwhm",
+        "target_affine",
+        "slice_time_ref",
+        "fir_delays",
+    ]
```
bthirion, line 1054: IIRC the use of single quotes is preferred

bthirion (commented):
Thx. A step in the right direction !

nilearn/reporting/glm_reporter.py
tsalo: Since BIDS models (and more importantly their outputs) aren't actually in the specification yet, I'm basing this off of `fitlins`, which does use SVG. Still, I don't think it's necessary to stick with their approach completely. Would you prefer another format?

nilearn/reporting/glm_reporter.py
tsalo: 🤦 I keep forgetting that. Sorry!

nilearn/reporting/glm_reporter.py
bthirion: svg is fine, I was just curious.

nilearn/reporting/glm_reporter.py
```diff
@@ -967,3 +972,221 @@ def _dataframe_to_html(df, precision, **kwargs):
...
+        )
+    return new_name
+
+
+def save_glm_results(model, contrasts, out_dir='.', prefix=None):
+    """Save GLM results to BIDS-like files.
+
+    Parameters
+    ----------
+    model
```
bthirion, line 1009: docstrings to be finished.

nilearn/reporting/glm_reporter.py
```diff
@@ -967,3 +972,221 @@ def _dataframe_to_html(df, precision, **kwargs):
...
+        'target_shape',
+        'signal_scaling',
+        'drift_order',
+        'scaling_axis',
+        'smoothing_fwhm',
+        'target_affine',
+        'slice_time_ref',
+        'fir_delays',
+    ]
+    # ATTRIBUTE_UNITS = {
```
bthirion, line 1111: to be removed ?

bthirion (commented):
Thx. A few comments enclosed.

nilearn/reporting/glm_reporter.py
tsalo: I think that the sidecar JSON files (not yet written out) should include unit information, but the exact metadata to retain,  how the fields should be named, etc. is not yet determined.

nilearn/reporting/glm_reporter.py
```diff
@@ -967,3 +973,289 @@ def _dataframe_to_html(df, precision, **kwargs):
...
+
+    # Write out design matrices to files.
+    if hasattr(model, 'design_matrices_'):
+        design_matrices = model.design_matrices_
+    else:
+        design_matrices = [model.design_matrix_]
+
+    # TODO: Assuming that cases of multiple design matrices correspond to
+    # different runs. Not sure if this is correct. Need to check.
+    for i_run, design_matrix in enumerate(design_matrices):
```
tsalo, line 1061: My assumption here is that multiple design_matrices in a model will _always_ mean applying the same model to multiple runs. Is that accurate?

nilearn/reporting/glm_reporter.py
```diff
@@ -967,3 +973,289 @@ def _dataframe_to_html(df, precision, **kwargs):
...
+    for contrast_name, contrast_maps in statistical_maps.items():
+        contrast_name = _clean_contrast_name(contrast_name)
+
+        # Extract stat_type
+        # TODO: Figure out if this is correct
+        contrast_matrix = contrasts[contrast_name]
+        if contrast_matrix.ndim == 2:
+            stat_type = 'F'
+        else:
+            stat_type = 't'
```
tsalo, line 1208: Is this a good way to determine the statistic type for the contrast? It's how contrast type is inferred when not explicitly set, and I don't think this function allows users to set contrast types explicitly, so it seems reasonable to also infer the statistic type the same way here, but I want to be sure.

nilearn/reporting/glm_reporter.py
tsalo: In retrospect, it's probably not worth the effort to try to incorporate units just yet, so I'll remove.

nilearn/reporting/glm_reporter.py
```diff
@@ -967,3 +973,284 @@ def _dataframe_to_html(df, precision, **kwargs):
...
+        Output directory for files. Default is current working directory.
+    prefix : :obj:`str` or None, optional
+        String to prepend to generated filenames.
+        If a string is provided, "_" will be added to the end.
+        Default is None.
+
+    Warning
+    -------
+    The files generated by this function are a best approximation of
+    appropriate for a GLM-based BIDS derivative.
```
bthirion, line 1027: Is it possible to add a ref to a given spec ?

nilearn/reporting/glm_reporter.py
bthirion: In  theory it is possible to one-dimensional F contrasts (that are equivalent to  two-sided t tests). To me this information has to be sued supplied. See e.g. the `contrast_type` in `compute_contrast` function

nilearn/reporting/glm_reporter.py
tsalo: Outputs aren't written up in a spec yet, but the idea was brought up recently in https://github.com/poldracklab/fitlins/issues/299.

nilearn/reporting/glm_reporter.py
tsalo: Okay, thanks. Since the function accepts contrasts as a dictionary of name-array pairs, I'm not sure where to put the contrast type information. It could maybe be a separate parameter- a dictionary mapping contrast names to contrast types, with `None`s or missing values assumed to mean that automatic type detection (i.e., 1-dim = t, 2-dim = F) should be used. Does that sound reasonable?

bthirion: I have no better suggestion.

tsalo: I'll try the new parameter out then.

nilearn/reporting/glm_reporter.py
tsalo: I've started working on translating the `fitlins` rules into a BEP that we can reference: https://docs.google.com/document/d/1OxKhhEctlQg6nmw1L3vw9Mm0K9-KkTpVKZuNqYnlYEM/edit?usp=sharing

nilearn/interfaces/bids.py
```diff
@@ -158,3 +168,324 @@ def parse_bids_filename(img_path):
         else:
             reference[field] = None
     return reference
+
+
+def save_glm_to_bids(
```
NicolasGensollen, line 173: If this is going to be a public function, could you add it to the API doc?

nilearn/interfaces/bids.py
```diff
@@ -158,3 +168,324 @@ def parse_bids_filename(img_path):
...
+    - Model error (stat-errorts_statmap.nii.gz)
+    - Model r-squared (stat-rSquare_statmap.nii.gz)
+    - Contrast parameter estimates (contrast-[name]_stat-effect_statmap.nii.gz)
+    - Variance of the contrast parameter estimates
+      (contrast-[name]_stat-variance_statmap.nii.gz)
+    - Contrast test statistics (contrast-[name]_stat-[F|t]_statmap.nii.gz)
+    - Contrast p- and z-values (contrast-[name]_stat-[p|z]_statmap.nii.gz)
+    - Contrast weights figure (contrast-[name]_design.svg)
+    """
+    DATA_ATTRIBUTES = [
```
NicolasGensollen, line 232: Also, not a priority as this is still WIP, but I think that this function is way too long, and would gain in clarity from a refactoring in smaller private functions handling the various file outputs.
Also, it looks like having a simple file for all BIDS interfaces might not scale well if we plan to develop our BIDS support.
It could be worth having private utility functions useful across the bids interface in a separate file like `nilearn/interfaces/bids/_utils.py`, such that this would go in something like `nilearn/interfaces/bids/glm.py`.
WDYT?

NicolasGensollen (commented):
Thanks for working on this @tsalo ! 👍 
Just had a first quick look and made a couple general comments.

nilearn/interfaces/bids.py
tsalo: That sounds like a good plan. I'll break it up.

bthirion:
This one needs another round of reviews, right ?

tsalo:
Yes, I believe so.

ymzayek:
I will give it a look

nilearn/interfaces/bids/_utils.py
```diff
@@ -0,0 +1,127 @@
...
+        'signal_scaling',
+        'drift_order',
+        'scaling_axis',
+        'smoothing_fwhm',
+        'target_affine',
+        'slice_time_ref',
+        'fir_delays',
+    ]
+    ATTRIBUTE_RENAMING = {
+        't_r': 'RepetitionTime',
```
bthirion, line 74: Is this a BIDS requirement ? If yes, we should rename the parameter across the codebase.

nilearn/interfaces/bids/_utils.py
```diff
@@ -0,0 +1,127 @@
...
+        json.dump(model_metadata, f_obj, indent=4, sort_keys=True)
+
+
+def _generate_dataset_description(out_file, model_level):
+    """Generate a BIDS dataset_description.json file with relevant metadata."""
+    dataset_description = {
+        'GeneratedBy': {
+            'Name': 'nilearn',
+            'Version': nilearn.__version__,
+            'Description': 'A nilearn {} GLM.'.format(
```
bthirion, line 116:
```suggestion
            'Description': 'A Nilearn {} GLM.'.format(
```

bthirion (commented):
A bunch of small things, but this looks good overall.

nilearn/interfaces/bids/glm.py
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        for map_name, img in renamed_contrast_maps.items():
+            out_file = os.path.join(out_dir, map_name)
+            img.to_filename(out_file)
+
+    # Write out model-level statistical maps
+    model_level_mapping = {
+        'residuals': f'{prefix}stat-errorts_statmap.nii.gz',
+        'r_square': f'{prefix}stat-rSquare_statmap.nii.gz',
```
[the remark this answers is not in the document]
tsalo: The names are primarily based on [the fitlins naming convention](https://github.com/poldracklab/fitlins/blob/c65b7d268c0f1c46c9ef6b9bf623590f107e077a/fitlins/workflows/base.py#L171). Some appear to be camelCase (rSquare, logLikelihood), while most are not.

I don't think the BIDS folks would have a problem with us changing these names, since they're not really codified. I would just want to also propose the change in my [pre-BEP document](https://docs.google.com/document/d/1OxKhhEctlQg6nmw1L3vw9Mm0K9-KkTpVKZuNqYnlYEM/edit?usp=sharing) to make sure that it's openly up for discussion.

nilearn/interfaces/bids/_utils.py
tsalo: I'm not sure if RepetitionTime is a required metadata field for GLM outputs, but, if it is included, it should be called RepetitionTime. I erred on the side of including everything I could infer from the available parameters and data, but we may want to be more selective about what metadata we carry on to the sidecar files, if only because some of it could be extraneous.

nilearn/interfaces/bids.py
tsalo: Done in [d85897e](https://github.com/nilearn/nilearn/pull/2715/commits/d85897ec2e8eebb0aebaa25aaf4f1576b92dff62). Sorry, I forgot to follow up in this thread.

ymzayek:
I think it looks good! I have no additional comments at the moment

bthirion (approved):
LGTM, thx !

ymzayek (commented):
Are you be able to improve the test coverage a bit?

tsalo:
@ymzayek, I took a look at the CodeCov report to see where I could improve coverage, and many of the lines it flags as missed don't seem related to the changes in this PR, so I've updated from `main` and will check the coverage once a new report is generated.

ymzayek (approved):
Ok I see, otherwise LGTM!

NicolasGensollen (approved):
LGTM, thanks @tsalo !
Could you add a what's new entry?

tsalo:
I've just added the whatsnew entry. I still need to review the CodeCov updates to see where coverage can be improved.

doc/changes/latest.rst
```diff
@@ -8,6 +8,7 @@
 NEW
 ---
 
+- :func:`~interfaces.bids.glm.save_glm_to_bids` has been added, which writes model outputs to disk according to BIDS convention.
```
NicolasGensollen, line 11:
```suggestion
- :func:`~interfaces.bids.glm.save_glm_to_bids` has been added, which writes model outputs to disk according to BIDS convention (:gh:`2715` by `Taylor Salo`_).
```

ymzayek:
@tsalo can you resolve the conflicts and merge? Thank you!

tsalo:
@ymzayek that sounds good. I'll merge once CI passes.

PR #2714, opened 2021-03-01
EHN signal.clean standardise option PSC to allow negative mean signal 

htwangtw:
Closes #2713 

Changes proposed in this pull request:

- Allow confound variables with a negative mean signal to pass through PSC option
- Imporve test on PSC

PR #2716, opened 2021-03-03
Fix whatsnew entry with public function reference

NicolasGensollen:
Fix current CircleCI build failure by fixing latest entry of  `whats_new.rst` to reference public function `signal.clean` instead of `signal._standardize` (introduced in #2714).


bthirion (approved):
Thx !

thomasbazeille:
Thanks!

PR #2719, opened 2021-03-04
Add two_sided flag to get_clusters_table

### LINKED ISSUE #2547: nistats: adding two_sided flag to get_clusters_table would result in more consistent API?
Hello,

I've been recently using nistats to report results from fMRI data analysis. I wanted to perform two-sided test for my 2nd level spm, since I was interested in both increased and decreased brain responses. `map_threshold` function provide flag `two_sided` which nicely adjust significance level preserving clusters with both positive and negative test statistic values. 

Next I wanted to use `get_clusters_table` function to pool information about peak coordinates. However, this function can only report peaks above `stat_threshold` (basically neglecting negative test stat values). I made a workaround to make it consistent with `map_threshold` – I multiplied my spm by (-1), reapplied `get_clusters_table` and then merged two tables together. 

Do you think that adding `two_sided` flag and its functionality to `get_clusters_table` is a good idea?

tsalo:
Closes #2547.

- Add a two_sided flag (default=True) to `nilearn.reporting._get_clusters_table()`.
- Update tests.
- Use two_sided thresholding in `glm_reporter`.

### DIFF
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -146,6 +146,10 @@ def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
     cluster_threshold : `int` or `None`, optional
         Cluster size threshold, in voxels.
 
+    two_sided : `bool`, optional
+        Whether to employ two-sided thresholding or to evaluate positive values
+        only. Default=True.
+
     min_distance : `float`, optional
         Minimum distance between subpeaks in mm. Default=8mm.
 
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -131,7 +131,7 @@ def _get_val(row, input_arr):
 
 
 def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
-                       two_sided=True, min_distance=8.):
+                       two_sided=False, min_distance=8.):
     """Creates pandas dataframe with img cluster statistics.
 
     Parameters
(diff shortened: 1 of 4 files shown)


nilearn/reporting/_get_clusters_table.py
```diff
@@ -131,7 +131,7 @@ def _get_val(row, input_arr):
 
 
 def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
-                       min_distance=8.):
+                       two_sided=True, min_distance=8.):
```
bthirion, line 134: I think I'd default to False for the sake of bw compatibility.

bthirion (commented):
Looks good otherwise, thx !

nilearn/reporting/_get_clusters_table.py
tsalo: Good point.

NicolasGensollen (commented):
The changes look good to me, thanks! :+1: 
I think we need to expose `two_sided` arg to `make_glm_report`, right?
Also, this probably needs a whats_new entry.

tsalo:
Thanks to @bthirion and @NicolasGensollen for reviewing. I've added the What's New entry and `two_sided` is now a parameter for `make_glm_report` as well.

NicolasGensollen:
I think you need to change the reference in `whats_new` to be `nilearn.reporting.make_glm_report` and `nilearn.reporting.get_clusters_table`. This should fix the broken CI.

bthirion (commented):
Has CodeCov gone crazy ?
The PR LGTM otherwise.

NicolasGensollen (approved):
LGTM, thanks ! :+1:

PR #2722, opened 2021-03-08
[DOC] Add guidelines in doc for making new releases of Nilearn

NicolasGensollen:
Since we are close to make the `0.7.1` release, I thought it'd be a good idea to have the steps for making a new release of Nilearn in the contributing doc. 
This is a first draft proposal. I am not sure I added the text in the best place, and I don't know if it is too detailed or not enough. Comments and suggestions are welcome, especially if I forgot something!

### DIFF
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -249,7 +249,7 @@ Add these changes and submit a PR:
 
 .. code:: bash
     
-    git add doc/whats_new.rst nilearn.version.py
+    git add doc/whats_new.rst nilearn/version.py
     git commit -m "REL x.y.z"
     git push origin REL-x.y.z
 
@@ -259,9 +259,9 @@ Once the PR has been reviewed and merged, pull from master and tag the merge com
 .. code:: bash
 
     git checkout master
-    git pull origin master
+    git pull upstream master
     git tag x.y.z
-    git push --tags
+    git push upstream --tags
 
 
 Build the distributions and upload them to Pypi
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -176,7 +176,8 @@ Benevolent Dictator, currently Gaël Varoquaux.
 How to make a release?
 ======================
 
-This section describes how to make a new release of Nilearn.
+This section describes how to make a new release of Nilearn. It is targeted to the specific case of Nilearn although it contains generic steps for packaging and distributing projects. More detailed information can be found on `packaging.python.org <https://packaging.python.org/guides/distributing-packages-using-setuptools/#id70>`_.
+
 We assume that we are in a clean state where all the Pull Requests (PR) that we wish to include in the new release have been merged.
 For example, make sure all deprecations that are supposed to be removed with this new version have been adressed. Furthermore, if this new release comes with dependency version bumps (Python, Numpy...), make sure to implement and test these changes beforehand. Ideally, these would have been done before such as to update the code base if necessary. Finally, make sure the documentation can be built correctly.
 
@@ -289,7 +290,7 @@ Build the source and binary distributions:
 This should add two files to the `dist` subfolder:
 
 - one for the source distribution that should look like `PACKAGENAME-VERSION.tar.gz`
-- one for the binary distribution that should look like `PACKAGENAME-PACKAGEVERSION-PYTHONVERSION-PYTHONCVERSION-PLATFORM.whl`
+- one for the built distribution that should look like `PACKAGENAME-PACKAGEVERSION-PYTHONVERSION-PYTHONCVERSION-PLATFORM.whl`
   
 Optionally, we can run some basic checks with `twine`:
 
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -179,7 +179,7 @@ How to make a release?
 This section describes how to make a new release of Nilearn. It is targeted to the specific case of Nilearn although it contains generic steps for packaging and distributing projects. More detailed information can be found on `packaging.python.org <https://packaging.python.org/guides/distributing-packages-using-setuptools/#id70>`_.
 
 We assume that we are in a clean state where all the Pull Requests (PR) that we wish to include in the new release have been merged.
-For example, make sure all deprecations that are supposed to be removed with this new version have been adressed. Furthermore, if this new release comes with dependency version bumps (Python, Numpy...), make sure to implement and test these changes beforehand. Ideally, these would have been done before such as to update the code base if necessary. Finally, make sure the documentation can be built correctly.
+For example, make sure all deprecations that are supposed to be removed with this new version have been addressed. Furthermore, if this new release comes with dependency version bumps (Python, Numpy...), make sure to implement and test these changes beforehand. Ideally, these would have been done before such as to update the code base if necessary. Finally, make sure the documentation can be built correctly.
 
 Prepare the release
 -------------------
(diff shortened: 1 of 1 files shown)


doc/development.rst
```diff
@@ -172,3 +172,186 @@ If there are open questions, final decisions are made by the Temporary
...
+
+We now need to update the documentation:
+
+.. code-block:: bash
+    
+    cd doc
+    make install
+
+
+This will build the documentation (beware, this is time consuming...) and push it to the `GitHub pages repo <https://github.com/nilearn/nilearn.github.io>`_.
```
bthirion, line 324: Maybe you could give here the push command you use (not completely obvious from the context).

bthirion (commented):
Thx, this is great !

doc/development.rst
```diff
@@ -172,3 +172,186 @@ If there are open questions, final decisions are made by the Temporary
...
+.. code:: bash
+    
+    git add doc/whats_new.rst nilearn.version.py
+    git commit -m "REL x.y.z"
+    git tag x.y.z
+    git push origin release-x.y.z
+    git push --tags
+
+
+Build the distributions and upload them to Pypi
```
jeromedockes, line 251: as this matches closely the instructions from https://packaging.python.org/guides/distributing-packages-using-setuptools/#id70, maybe add a link?

doc/development.rst
```diff
@@ -172,3 +172,186 @@ If there are open questions, final decisions are made by the Temporary
...
+
+.. code-block:: bash
+    
+    python setup.py sdist bdist_wheel
+
+
+This should add two files to the `dist` subfolder:
+
+- one for the source distribution that should look like `PACKAGENAME-VERSION.tar.gz`
+- one for the binary distribution that should look like `PACKAGENAME-PACKAGEVERSION-PYTHONVERSION-PYTHONCVERSION-PLATFORM.whl`
```
jeromedockes, line 292: maybe the term "binary distribution" is a bit confusing, as it is just source code + installation instructions in a compressed archive -- no compiled binaries

jeromedockes: see for example https://packaging.python.org/glossary/#glossary

doc/development.rst
NicolasGensollen: Although I never ran this before, I don't think you have to do anything more than that. All the git commands are in the `install` target:

https://github.com/nilearn/nilearn/blob/80572dbf2253bd7a5a00e688082cbd3075e71a12/doc/Makefile#L161-L178

NicolasGensollen:
Link to the page on circleCI: [link](https://6840-1235740-gh.circle-artifacts.com/0/doc/_build/html/development.html#how-to-make-a-release)

thomasbazeille:
I love it ! In terms of website flow. I think we should create some other page dedicated to core maintenance tasks which may be less publicized and anyway not clutter the most visible pages

doc/development.rst
bthirion: OK, I misread it. thx.

bthirion (approved):
Thx !

doc/development.rst
```diff
@@ -172,3 +172,202 @@ If there are open questions, final decisions are made by the Temporary
...
+
+Here, we should remove the last entry and add the new release on top of the list.
+
+In addition, we can have a look at `MANIFEST.in` to check that all additional files that we want to be included or excluded from the release are indicated. Normally we shouldn't have to touch this file.
+
+Add these changes and submit a PR:
+
+.. code:: bash
+    
+    git add doc/whats_new.rst nilearn.version.py
```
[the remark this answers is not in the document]
NicolasGensollen: thanks!

doc/development.rst
```diff
@@ -172,3 +172,202 @@ If there are open questions, final decisions are made by the Temporary
...
 
 .. include:: ../CONTRIBUTING.rst
+
+How to make a release?
+======================
+
+This section describes how to make a new release of Nilearn. It is targeted to the specific case of Nilearn although it contains generic steps for packaging and distributing projects. More detailed information can be found on `packaging.python.org <https://packaging.python.org/guides/distributing-packages-using-setuptools/#id70>`_.
+
+We assume that we are in a clean state where all the Pull Requests (PR) that we wish to include in the new release have been merged.
+For example, make sure all deprecations that are supposed to be removed with this new version have been adressed. Furthermore, if this new release comes with dependency version bumps (Python, Numpy...), make sure to implement and test these changes beforehand. Ideally, these would have been done before such as to update the code base if necessary. Finally, make sure the documentation can be built correctly.
```
tsalo, line 182:
```suggestion
For example, make sure all deprecations that are supposed to be removed with this new version have been addressed. Furthermore, if this new release comes with dependency version bumps (Python, Numpy...), make sure to implement and test these changes beforehand. Ideally, these would have been done before such as to update the code base if necessary. Finally, make sure the documentation can be built correctly.
```

tsalo (commented):
I just noticed one small typo. I also have a question about the approach. I'm obviously not as familiar with nilearn's packaging as I should be, but would it be possible to reduce the number of steps here by using a GitHub Action to deploy to PyPi?

doc/development.rst
NicolasGensollen: Thanks!

NicolasGensollen:
>would it be possible to reduce the number of steps here by using a GitHub Action to deploy to PyPi?

You mean using something like this: [https://github.com/marketplace/actions/pypi-publish](https://github.com/marketplace/actions/pypi-publish)?
I haven't used it before, but can have a look and experiment a little bit to see it could be useful here. 
In any case, having the manual steps in the doc cannot hurt I think.

jeromedockes:
> I just noticed one small typo. I also have a question about the approach. I'm obviously not as familiar with nilearn's packaging as I should be, but would it be possible to reduce the number of steps here by using a GitHub Action to deploy to PyPi?
>

It is possible. however deploying to PyPi is just 2 commands (python
setup.py then twine upload), so it wouldn't make a huge difference. I
believe one drawback is that it would require making the PyPi password
available the the github action while making sure it is never visible by
builds of forks -- and there's always a risk of checking the wrong box
in the CI configuration

tsalo:
> You mean using something like this: https://github.com/marketplace/actions/pypi-publish?
> I haven't used it before, but can have a look and experiment a little bit to see it could be useful here.

@NicolasGensollen Exactly! I use it in `NiMARE` and the `tedana` team (including @emdupre) just started using it as well. It works really well- any time a release is made on GitHub, the Action is triggered to deploy to PyPi.

> In any case, having the manual steps in the doc cannot hurt I think.

I totally agree. I think the documentation changes here are great. I just wasn't sure what the best place to bring up the packaging was. I can open an issue about it though.

> It is possible. however deploying to PyPi is just 2 commands (python setup.py then twine upload), so it wouldn't make a huge difference. I believe one drawback is that it would require making the PyPi password available the the github action while making sure it is never visible by builds of forks -- and there's always a risk of checking the wrong box in the CI configuration

@jeromedockes It does look like many of the steps won't be affected by publishing directly from GitHub, but on the other hand I don't think the drawback you mentioned will be an issue. The password and username will be repository-specific secret keys, so no fork will be able to access them. I'll open an issue about it to not distract from the main purpose of the PR.

NicolasGensollen:
Thank you all for the reviews! :+1: 
Merging...

PR #2723, opened 2021-03-09
Add Juelich Atlas

### LINKED ISSUE #1145: Add Juelich atlas
I just noticed that fsl comes installed with a hell lot of interesting atlases: HarvardOxford, Juelich, JHU, MNI, etc.

Should we have the possibility to fetching this (at least locally when present) ? For example, I find the Juelich atlas really interesting.

inversecovariance

achamma723:
- Closes #1145 
- Fixing the gititerator in the tests_atlas.py (for compatibility with python 3)
- Adding the Juelich atlas (the provided url is from a private website where the atlas is uploaded, can be changed)
- Adding a test for fail_fetch of Juelich Atlas


nilearn/datasets/atlas.py
```diff
@@ -428,6 +428,137 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
     atlas_img = new_img_like(atlas_img, new_atlas, atlas_img.affine)
     return Bunch(maps=atlas_img, labels=new_names)
 
+def fetch_atlas_juelich(atlas_name, data_dir=None,
```
tsalo, line 431: We need two lines before the function definition.

nilearn/datasets/atlas.py
```diff
@@ -428,6 +428,137 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+            new_names.append(name)
+            continue
+        new_atlas[right_atlas == label] = new_label
+        new_names.append(name + ', left part')
+        new_label += 1
+        new_atlas[left_atlas == label] = new_label
+        new_names.append(name + ', right part')
+
+    atlas_img = new_img_like(atlas_img, new_atlas, atlas_img.affine)
+    return Bunch(maps=atlas_img, labels=new_names)
```
tsalo, line 562: We need two lines before the next function definition.

nilearn/datasets/tests/test_atlas.py
```diff
@@ -242,6 +242,97 @@ def test_fail_fetch_atlas_harvard_oxford(tmp_path, request_mocker):
     assert ho.labels[5] == "Left R3"
     assert ho.labels[6] == "Right R3"
 
+def test_fail_fetch_atlas_juelich(tmp_path, request_mocker):
```
tsalo, line 245: Needs two lines.

nilearn/datasets/tests/test_atlas.py
```diff
@@ -242,6 +242,97 @@ def test_fail_fetch_atlas_harvard_oxford(tmp_path, request_mocker):
...
+    assert isinstance(ho.maps, nibabel.Nifti1Image)
+    assert isinstance(ho.labels, list)
+    assert len(ho.labels) == 7
+    assert ho.labels[0] == "Background"
+    assert ho.labels[1] == "R1, left part"
+    assert ho.labels[2] == "R1, right part"
+    assert ho.labels[3] == "R2, left part"
+    assert ho.labels[4] == "R2, right part"
+    assert ho.labels[5] == "R3, left part"
+    assert ho.labels[6] == "R3, right part"
```
tsalo, line 336: Needs two lines.

tsalo (changes requested):
Thanks for opening this! This review is just a first pass. I hope to be able to review in more detail tomorrow.

Could you rename the PR to summarize what is being proposed (e.g., "Add Juelich atlas") and can you use [a keyword](https://docs.github.com/en/enterprise/2.18/user/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to link this PR to the associated issue instead of referencing it?

nilearn/datasets/atlas.py
```diff
@@ -429,6 +429,139 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+    atlas_items = ("maxprob-thr0-1mm", "maxprob-thr0-2mm",
+                   "maxprob-thr25-1mm", "maxprob-thr25-2mm",
+                   "maxprob-thr50-1mm", "maxprob-thr50-2mm",
+                   "prob-1mm", "prob-2mm")
+    if atlas_name not in atlas_items:
+        raise ValueError("Invalid atlas name: {0}. Please chose an atlas "
+                         "among:\n{1}".format(
+                             atlas_name, '\n'.join(atlas_items)))
+
+    url = "https://cdn.filesend.jp/private/1McfSSXL0-f67vqFrnEIbQZbfvS7YHn8ZTHoDL657Goj2FZFMnbsOp21SE1QpdYO/Juelich.tgz"
```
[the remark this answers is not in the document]
achamma723: This url is providing the download part for the dataset. I put it here temporarly to test the work. Now i'm searching to move to orcid

nilearn/datasets/atlas.py
```diff
@@ -429,6 +429,139 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+        left_elements = (left_atlas == label).sum()
+        right_elements = (right_atlas == label).sum()
+        n_elements = float(left_elements + right_elements)
+        if (left_elements / n_elements < 0.05 or
+                right_elements / n_elements < 0.05):
+            new_atlas[atlas == label] = new_label
+            new_names.append(name)
+            continue
+        new_atlas[right_atlas == label] = new_label
+        new_names.append(name + ', left part')
```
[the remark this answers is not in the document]
achamma723: I should check again cause i took this part from the fetch_harvard_oxford function

nilearn/datasets/atlas.py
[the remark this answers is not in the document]
NicolasGensollen: >Now i'm searching to move to orcid

Ok, isn't there a way to add it to osf or nitrc instead?

>@NicolasGensollen What do you mean by expose the function?

By exposing, I mean adding it to `__init__.py` such that users can import it as any other fetcher:

```python
from nilearn.datasets import fetch_atlas_juelich
```

nilearn/datasets/atlas.py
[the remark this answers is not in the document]
NicolasGensollen: OK :+1: 
Again, I didn't look extensively, so this might be a stupid comment...

nilearn/datasets/atlas.py
```diff
@@ -429,6 +429,139 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
     return Bunch(maps=atlas_img, labels=new_names)
 
 
+def fetch_atlas_juelich(atlas_name, data_dir=None,
+                        symmetric_split=False,
+                        resume=True, verbose=1):
+    """Load Juelich parcellations from FSL.
+
+    This function downloads Juelich Histological atlas packaged from FSL 5.0
```
bthirion, line 437: Is it the most up-to-date version of this atlas ?

nilearn/datasets/atlas.py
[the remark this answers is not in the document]
bthirion: But do we actually need all that ?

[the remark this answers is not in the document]
bthirion: Do you mean that the Juelich atlas has the same structure as Harvard-Oxford ?

bthirion (commented):
Thx for addressing that.

NicolasGensollen:
Hi @achamma723 
How is this going? Let me know if I can help!

achamma723:
Hello @NicolasGensollen , Thank you! 
I was occupied by some coding in my thesis. but i was doing some slight modifications locally. Once i finish, i'll push it I think by tomorrow.

nilearn/datasets/tests/test_atlas.py
```diff
@@ -243,6 +243,99 @@ def test_fail_fetch_atlas_harvard_oxford(tmp_path, request_mocker):
...
+    atlas_data[5:, 3:5, :] = 2
+
+    atlas_data[:5, 7:9, :] = 3
+    atlas_data[5:, 7:9, :] = 3
+
+
+    nibabel.Nifti1Image(atlas_data, np.eye(4) * 3).to_filename(
+        target_atlas_nii)
+
+    dummy = open(os.path.join(ho_dir, 'Juelich.xml'), 'w')
```
jeromedockes, line 277: please rely on context managers to release resources such as open file handles:
```
with open("...Juelich.xml") as dummy:
    dummy.write("...")
# dummy is closed automatically
```
or:
```
pathlib.Path("...Juelich.xml").write_text("...")
```

nilearn/datasets/atlas.py
```diff
@@ -429,6 +429,139 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+        data_dir,
+        [(atlas_file, url, opts), (label_file, url, opts)],
+        resume=resume, verbose=verbose)
+
+    names = {}
+    from xml.etree import ElementTree
+    names[0] = 'Background'
+    for label in ElementTree.parse(label_file).findall('.//label'):
+        names[int(label.get('index')) + 1] = label.text
+    names = list(names.values())
```
jeromedockes, line 520: the order is the order of insertion (not, for example, of the sorted keys). why use a dict and take the values rather than a list from the start? are there duplicate indexes?

nilearn/datasets/atlas.py
```diff
@@ -429,6 +429,139 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+
+    atlas_img = check_niimg(atlas_img)
+
+    atlas = get_data(atlas_img)
+
+    labels = np.unique(atlas)
+    # Build a mask of both halves of the brain
+    middle_ind = (atlas.shape[0] - 1) // 2
+    # Put zeros on the median plane
+    atlas[middle_ind, ...] = 0
```
jeromedockes, line 533: why do you need the ellipsis?

jeromedockes: (same comment below)

nilearn/datasets/atlas.py
```diff
@@ -429,6 +429,139 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+    left_atlas[middle_ind:, ...] = 0
+    right_atlas = atlas.copy()
+    right_atlas[:middle_ind, ...] = 0
+
+    new_label = 0
+    new_atlas = atlas.copy()
+    # Assumes that the background label is zero.
+    new_names = [names[0]]
+
+    for label, name in zip(labels[1:], names[1:]):
```
jeromedockes, line 545: doesn't this assume the labels occur in the xml in the right order and without duplicates or missing indexes (see above)?

nilearn/datasets/atlas.py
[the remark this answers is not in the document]
jeromedockes: (note the url returns a 404 error )

nilearn/datasets/tests/test_atlas.py
```diff
@@ -243,6 +243,99 @@ def test_fail_fetch_atlas_harvard_oxford(tmp_path, request_mocker):
...
+                '<label index="2" x="135" y="89" z="109">R3</label>\n'
+                "</data>")
+    dummy.close()
+
+    # Here, with symmetric_split=True, atlas maps are returned as nibabel Nifti
+    # image but not string. Now, with symmetric split number of labels should be
+    # more than without split and contain Left and Right tags in the labels.
+
+    # Create dummy image files too with maxprob specified for symmetric split.
+    split_atlas_fname = 'Juelich-' + 'maxprob-thr0-1mm' + '.nii.gz'
```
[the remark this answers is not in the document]
achamma723: Can you explain please the difference between the two?

nilearn/datasets/atlas.py
```diff
@@ -429,6 +429,139 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+        return Bunch(maps=atlas_img, labels=names)
+
+    atlas_img = check_niimg(atlas_img)
+
+    atlas = get_data(atlas_img)
+
+    labels = np.unique(atlas)
+    # Build a mask of both halves of the brain
+    middle_ind = (atlas.shape[0] - 1) // 2
+    # Put zeros on the median plane
```
jeromedockes, line 532: why do you set this slice to 0?

nilearn/datasets/tests/test_atlas.py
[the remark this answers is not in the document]
jeromedockes: it is more explicit and will work with other types than `str`, so for example we could say say `'Juelich-{}.nii.gz'.format(target_atlas)` in place of `'Juelich-' + target_atlas + '.nii.gz'`; when nilearn drops support for python 3.5 we will also be able to write `f"Juelich-{target_atlas}.nii.gz"` -- but it is really a detail and a question of taste. I just pointed out this example because all parts were literals so there was no need for formatting or concatenation here. 
(the last one, join, is for joining sequences with a given separator eg `", ".join([a,b])`; for large sequences of strings `"".join` should be more efficient than repeatedly using `+` but of course this doesn't apply here)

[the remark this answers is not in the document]
achamma723: Ok, I got it thank you. I made the changes in the new commit and a part of my work with Juelich (before factorizing the two functions for HarvardOxford and Juelich) was inspired by the Harvard_Oxford function. This is why i didn't think of big changes in the structure of the function.

bthirion (commented):
LGTM, thx ! Beware that a couple of lines are not covered.
Best,

nilearn/datasets/atlas.py
```diff
@@ -268,37 +268,47 @@ def fetch_atlas_destrieux_2009(lateralized=True, data_dir=None, url=None,
     return Bunch(**params)
 
 
-def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
-                               symmetric_split=False,
-                               resume=True, verbose=1):
-    """Load Harvard-Oxford parcellations from FSL.
+def fetch_atlases_fsl(atlas_source, atlas_name,
```
NicolasGensollen, line 271: This is not backward compatible. Users relying on `fetch_atlas_harvard_oxford` will have their code broken by this change. 
If we decide to combine harvard-oxford and Juelich fetchers in a new function `fetch_atlases_fsl`, we will need to run a deprecation cycle on `fetch_atlas_harvard_oxford` first. 
One way to do that would be to keep  `fetch_atlas_harvard_oxford` as a wrapper around `fetch_atlases_fsl` and gives a deprecation warning saying that `fetch_atlas_harvard_oxford` is deprecated, will be removed in release 0.9, and that the new usage is to rely on `fetch_atlases_fsl` with `atlas_name` set to `Harvard Oxford`.

NicolasGensollen (changes requested):
Thanks @achamma723 !
I think combining the Harvard-Oxford and Juelich fetchers into a single function is a good idea. However, we need to be careful about backward compatibility (see my comment below). 
Looking at your code, I don't think this will be hard to do since we are just renaming a function but keeping all previous functionalities. 
Note that we will have to maintain tests for `fetch_atlas_harvard_oxford` even if we deprecate it, and we will have to update the examples relying on this fetcher to show the new usage.

jeromedockes:
> If we decide to combine harvard-oxford and Juelich fetchers in a new function `fetch_atlases_fsl`, we will need to run a deprecation cycle on `fetch_atlas_harvard_oxford` first. 

are we sure we want to do this? using the harvard-oxford fetcher is
already not very easy because you have to look up the complicated atlas
names (eg "cort-maxprob-thr0-1mm") in the docstring, and I'm not sure
adding a parameter and removing the atlas name from the function name
will be an improvement. moreover, the new signature is quite confusing,
because the accepted values for one parameter (atlas_name) depend on the
value of another parameter (atlas_source) (none of the atlas_name values
for harvard-oxford are valid for juelich). Finding good names for these
parameters is also hard: I would expect "atlas_name" to be "Juelich" or
"harvard-oxford", but that is actually "atlas_source", because
atlas_name is aready taken. Finally the resulting function is too long
with too deeply nested if/else.

Perhaps more importantly, breaking backwards compatibility (even with a
deprecation cycle) is very bad and should only be a last resort, and I
don't think there is a strong case for it here.

I would suggest keeping 2 separate functions. If they use very similar
code, we can add a private function containing the implementation and
have both interfaces call it.

jeromedockes:
also besides the `symmetric_split` parameter these function's signatures don't have more in common than with any of the other fetchers so I don't really see why they should be merged

GaelVaroquaux:
Agreed with Jerome on the value of keeping separate functions.

NicolasGensollen:
>If they use very similar code, we can add a private function containing the implementation and have both interfaces call it.

This is indeed a good solution if they share very similar code, which seems to be the case AFAICT.

bthirion (commented):
LGTM overall.

nilearn/datasets/atlas.py
```diff
@@ -351,26 +440,31 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
-    atlas_file = os.path.join(root, 'HarvardOxford',
-                              'HarvardOxford-' + atlas_name + '.nii.gz')
+    atlas_file = os.path.join(root, atlas_source,
+                              '{}-{}.nii.gz'.format(atlas_source, atlas_name))
 
     atlas_img, label_file = _fetch_files(
         data_dir,
-        [(atlas_file, url, opts), (label_file, url, opts)],
+        [(atlas_file, urls[source_idx], opts),
+         (label_file, urls[source_idx], opts)],
```
jeromedockes, line 467: I realize now that this was here before but making `names` below a dict is confusing if we never use the keys

nilearn/datasets/atlas.py
```diff
@@ -384,7 +478,8 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
         return Bunch(maps=atlas_img, labels=names)
 
     if atlas_name in ("cort-prob-1mm", "cort-prob-2mm",
-                      "sub-prob-1mm", "sub-prob-2mm"):
+                      "sub-prob-1mm", "sub-prob-2mm",
```
jeromedockes, line 481: we could do this check before downloading the files

nilearn/datasets/atlas.py
```diff
@@ -284,13 +284,17 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
     ----------
     atlas_name : string
         Name of atlas to load. Can be:
-        cort-maxprob-thr0-1mm,  cort-maxprob-thr0-2mm,
+        cort-maxprob-thr0-1mm, cort-maxprob-thr0-2mm,
         cort-maxprob-thr25-1mm, cort-maxprob-thr25-2mm,
         cort-maxprob-thr50-1mm, cort-maxprob-thr50-2mm,
-        sub-maxprob-thr0-1mm,  sub-maxprob-thr0-2mm,
+        cort-prob-1mm, cort-prob-2mm,
+        cortl-maxprob-thr0-1mm, cortl-maxprob-thr0-2mm,
```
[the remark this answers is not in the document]
achamma723: Hello, yes the lateralized versions are added now. The name, as used in the HarvardOxford maps, includes the "```l```" for the distinction. For your last question, I didn't quite understand

nilearn/datasets/atlas.py
```diff
@@ -351,26 +440,31 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
-        else:
-            label_file = 'HarvardOxford-Cortical.xml'
+    if source_idx == 0:
+        if atlas_name[0] == 'c':
             lateralized = False
+            if 'cort-maxprob' in atlas_name or 'cortl-maxprob' in atlas_name and symmetric_split:
+                split_name = atlas_name.split('cort')
+                atlas_name = 'cortl{}'.format(split_name[1])
+                label_file = 'HarvardOxford-Cortical-Lateralized.xml'
+                lateralized = True
```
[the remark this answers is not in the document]
achamma723: Yes. This was an old part of the code. But because HarvardOxford maps now supports the lateralized atlases, so I joined cortl with (cort + symm) in the condtition

bthirion (approved):
LGTM, thx.

jeromedockes:
```
ho = fetch_atlas_harvard_oxford("cortl-maxprob-thr0-2mm")
```

fails on my machine with this error:
```
line 807, in _fetch_files
    raise IOError('Fetching aborted: ' + abort)
OSError: Fetching aborted: Dataset has been downloaded but requested file was not provided:
URL: http://www.nitrc.org/frs/download.php/9902/HarvardOxford.tgz
Target file: nilearn_data/fsl/data/atlases/HarvardOxford/HarvardOxford-cortll-maxprob-thr0-2mm.nii.gz
Downloaded: nilearn_data/fsl/ae2e23157bf21b178b9a5168196ad08c/HarvardOxford.tgz
```

(actualy not just this one but all the `cortl-maxprob` ones)

jeromedockes:
and `symmetric_split` seems to have little effect on `fetch_atlas_juelich`,
`fetch_atlas_juelich("maxprob-thr0-2mm", symmetric_split=False)` has 122 labels, most of which end with 'L' or 'R'
with symmetric_split=True, there are 125 labels
the only difference between the 2 seems to be these regions:

'WM Callosal body', 'WM Fornix', 'GM Mamillary body'

NicolasGensollen (commented):
This is looking good, thanks @achamma723 !
Right now, `fetch_atlas_harvard_oxford` is used in three examples while `fetch_atlas_juelich` isn't used. Could you use the new fetcher in one of these examples? For instance [this one](https://nilearn.github.io/auto_examples/01_plotting/plot_atlas.html#sphx-glr-auto-examples-01-plotting-plot-atlas-py)?
Apart from this, I think this is ready. :tada:

jeromedockes:
> Apart from this, I think this is ready. :tada: 

I don't think so; AFAICT it still produces wrong results for
`symmetric_split=False` (see previous comment). For example

```
from nilearn.datasets import fetch_atlas_juelich

with_split = fetch_atlas_juelich("maxprob-thr0-2mm", symmetric_split=True)
no_split = fetch_atlas_juelich("maxprob-thr0-2mm", symmetric_split=False)
print("n regions with symmetric_split=True:", len(with_split.labels))
print("n regions with symmetric_split=False:", len(no_split.labels))
print("regions that don't end with 'L' or 'R' when symmetric_split=False:")
print(
    [
        lab
        for lab in no_split["labels"]
        if not lab.endswith("L") and not lab.endswith("R")
    ]
)
```

prints

```
n regions with symmetric_split=True: 125
n regions with symmetric_split=False: 122
regions that don't end with 'L' or 'R' when symmetric_split=False:
['Background', 'WM Callosal body', 'WM Fornix', 'GM Mamillary body']
```

unless I am missing something regions are splitted by hemisphere
regardless of the `symmetric_split` argument

achamma723:
Hello @jeromedockes, Thanks for your review. In fact, your comment is true. The Juelich xml file contains names with L and R symbols (like harvard-oxford lateralized where we have right and left in the name) without the 3 you mentioned. So, I kept the symmetric split in order to split the remaining . Do you see anything wrong?

jeromedockes:
> Hello @jeromedockes, Thanks for your review. In fact, your comment is true. The Juelich xml file contains names with L and R symbols (like harvard-oxford lateralized where we have right and left in the name) without the 3 you mentioned. So, I kept the symmetric split in order to split the remaining . Do you see anything wrong?

yes the issue is that with `symmetric_split=False` all the regions
except 3 are still splitted, so we can do one of the two options:

- remove this parameter and always return splitted regions
- keep it and merge the left and right regions when `symmetric_split=False`

NicolasGensollen:
@achamma723 the failures with python 3.5 are not related to python 3.5 itself or the tests (they all pass), but to PEP8 violations (some lines are too long, some contain whipespace at the end, and so on...). They should be pretty easy to fix. You can see the list of violations [here](https://github.com/nilearn/nilearn/pull/2723/checks?check_run_id=2711028357).

This makes me realize that running this check as part of the Python 3.5 job might be confusing. I'm wondering whether having PEP8 checks within their own dedicated workflow wouldn't be better... :thinking:

jeromedockes:
> This makes me realize that running this check as part of the Python 3.5 job might be confusing. I'm wondering whether having PEP8 checks within their own dedicated workflow wouldn't be better... :thinking: 

Yes I think having them in a separate workflow would be nice

achamma723:
Hello @NicolasGensollen, thanks for your response! Indeed, at first i was confused and was reading about the differences between python 3.5 and 3.6 versions (I said maybe something related to data structures) but the error was just related to PEP8. So, I agree to put these tests as seperated in the workflow so later noone will be confused.

NicolasGensollen:
@jeromedockes @achamma723 Sounds good!

NicolasGensollen (commented):
Hi @achamma723 

Maybe I'm missing something, but I would expect to get twice the number of labels (minus 1 because of the background) when setting `symetric_split` to True. It seems to be working with the labels themselves, but it looks like the maps are not affected:

```python
import numpy as np
from nilearn.datasets import fetch_atlas_juelich

jul = fetch_atlas_juelich("maxprob-thr0-1mm")
jul_sym = fetch_atlas_juelich("maxprob-thr0-1mm", 
                              symmetric_split=True)

assert len(jul_sym.labels) == len(jul.labels) * 2 - 1
assert len(np.unique(get_data(jul_sym.maps))) == 2 * (len(np.unique(get_data(jul.maps)))) - 1
```
```
AssertionError                            Traceback (most recent call last)
<ipython-input-1-8b0ce8db0549> in <module>
      8 
      9 assert len(jul_sym.labels) == len(jul.labels) * 2 - 1
---> 10 assert len(np.unique(get_data(jul_sym.maps))) == 2 * (len(np.unique(get_data(jul.maps)))) - 1

AssertionError: 
```

achamma723:
> Hi @achamma723
> 
> Maybe I'm missing something, but I would expect to get twice the number of labels (minus 1 because of the background) when setting `symetric_split` to True. It seems to be working with the labels themselves, but it looks like the maps are not affected:
> 
> ```python
> import numpy as np
> from nilearn.image import get_data
> from nilearn.datasets import fetch_atlas_juelich
> 
> jul = fetch_atlas_juelich("maxprob-thr0-1mm")
> jul_sym = fetch_atlas_juelich("maxprob-thr0-1mm", 
>                               symmetric_split=True)
> 
> assert len(jul_sym.labels) == len(jul.labels) * 2 - 1
> assert len(np.unique(get_data(jul_sym.maps))) == 2 * (len(np.unique(get_data(jul.maps)))) - 1
> ```
> 
> ```
> AssertionError                            Traceback (most recent call last)
> <ipython-input-1-8b0ce8db0549> in <module>
>       8 
>       9 assert len(jul_sym.labels) == len(jul.labels) * 2 - 1
> ---> 10 assert len(np.unique(get_data(jul_sym.maps))) == 2 * (len(np.unique(get_data(jul.maps)))) - 1
> 
> AssertionError: 
> ```

Yes @NicolasGensollen , In fact I merge the names but I still had to merge the maps. In the new commit, I think that everything is fine now.

nilearn/datasets/atlas.py
```diff
@@ -377,18 +487,32 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
 
+    if not symmetric_split:
+        atlas = get_data(atlas_img)
+        labels = np.unique(atlas)
+        new_label = 0
+        new_atlas = atlas.copy()
+        # Assumes that the background label is zero.
+        new_names = [names[0]]
+        for label, name in zip(labels[1:], names[1:]):
+            if name.endswith('R'):
```
NicolasGensollen, line 503: This loop works only if labels are always organized with the left version just before the right version. If this is reversed, or if they aren't consecutive this won't work. 
For Juelich, the labels are *almost* always organized like this, but there are a few exceptions. For example "WM Acoustic radiation" or "WM Inferior occipito-frontal fascicle".

NicolasGensollen (changes requested):
Thanks @achamma723 !

This looks better, but I still have a few concerns:

- The for loop merging left and right regions only works if labels have a specific order (which is not the case, at least in Juelich), see comment below.
- When doing `symmetric_split=True`, the labels that we obtain have either "R/L" or "right part/left part" at the end which is not consistent. We cannot change the addition of "right part/left part" as this would be a breaking change, so I think we need to replace "R/L" with right "part/left part". This could be done as a preprocessing of the Juelich labels such that they are always consistent. WDYT?
- When testing locally, I was expecting to get the same numbers for each region when summing the voxels of the right and left hemispheres and comparing with the non-splitted region: `n_voxel_region_L + n_voxel_region_R == n_voxel_region`. However this is not the case even for background. I think it comes partly from these lines:

https://github.com/nilearn/nilearn/blob/e71e01c8a685e0afbc33303b1cd0380276e495db/nilearn/datasets/atlas.py#L389-L393

where we lose all values in the median plane. 
I don't understand why this is needed... :thinking:

NicolasGensollen:
@achamma723 this is the code I used to test the fetcher:

```python
import numpy as np
from collections import Counter
from nilearn.datasets import fetch_atlas_juelich

jul = fetch_atlas_juelich("maxprob-thr0-1mm")
jul_sym = fetch_atlas_juelich("maxprob-thr0-1mm", 
                              symmetric_split=True)

assert len(jul_sym.labels) == len(jul.labels) * 2 - 1
assert len(np.unique(get_data(jul_sym.maps))) == 2 * (len(np.unique(get_data(jul.maps)))) - 1

jul_sym_labels_distri = Counter(get_data(jul_sym.maps).ravel())
jul_labels_distri = Counter(get_data(jul.maps).ravel())

for l,label in enumerate(jul.labels):
    print(label)
    if label != "Background":
        label_right = label + " R"
        label_left = label + " L"
        right = jul_sym_labels_distri[jul_sym.labels.index(label_right)]
        left = jul_sym_labels_distri[jul_sym.labels.index(label_left)] 
        print(right)
        print(left)
        print(right + left)
        print(jul_labels_distri[l])
    else:
        print(jul_sym_labels_distri[l])
        print(jul_labels_distri[l])
```
```
Background
6132470
6124848
GM Anterior intra-parietal sulcus hIP1
6945
8204
15149
15149
GM Anterior intra-parietal sulcus hIP2
2283
...
```

nilearn/datasets/atlas.py
achamma723: Hello @NicolasGensollen , thanks for the review! In fact, I thought about this yesterday and now It's changed in the new commit along with a small preprocessing for Juelich when is the case of symmetric_split to have the same extenstion as (left part or right part). I still have to check the last fetcher idea you mentioned as it is the logic that existed before my modifications

NicolasGensollen (commented):
Thanks @achamma723 this is looking much better! :+1: 

I still see some differences for some regions (and also background). There are always slightly less voxels when summing left and right:

```
GM Primary motor cortex BA4a (region name)
17763  (number of voxels when summing left and right with symmetric_split)
18311  (number of voxels without symetric_split)
GM Primary motor cortex BA4p
5816
5820
GM Primary somatosensory cortex BA3a
7980
7985
GM Primary somatosensory cortex BA3b
11992
12004
GM Superior parietal lobule 5Ci
7471
7490
GM Superior parietal lobule 5L
12871
12920
GM Superior parietal lobule 5M
11207
11719
GM Superior parietal lobule 7A
42241
42575
GM Superior parietal lobule 7M
5198
5598
GM Superior parietal lobule 7P
20912
21330
GM Visual cortex V1 BA17
39579
40643
GM Visual cortex V2 BA18
46774
47310
GM Visual cortex V3V
14404
14405
GM Premotor cortex BA6
120102
121571
WM Callosal body
104754
106316
WM Fornix
13020
13584
GM Mamillary body
1613
1738
```

I suspect these come from the last point of my previous review. When I replace the related lines by:

```python
# Build a mask of both halves of the brain
middle_ind = (atlas.shape[0]) // 2
# Put zeros on the median plane
#atlas[middle_ind, ...] = 0
# Split every zone crossing the median plane into two parts.
left_atlas = atlas.copy()
```

I don't have these differences anymore.

achamma723:
> Thanks @achamma723 this is looking much better! 
> 
> I still see some differences for some regions (and also background). There are always slightly less voxels when summing left and right:
> 
> ```
> GM Primary motor cortex BA4a (region name)
> 17763  (number of voxels when summing left and right with symmetric_split)
> 18311  (number of voxels without symetric_split)
> GM Primary motor cortex BA4p
> 5816
> 5820
> GM Primary somatosensory cortex BA3a
> 7980
> 7985
> GM Primary somatosensory cortex BA3b
> 11992
> 12004
> GM Superior parietal lobule 5Ci
> 7471
> 7490
> GM Superior parietal lobule 5L
> 12871
> 12920
> GM Superior parietal lobule 5M
> 11207
> 11719
> GM Superior parietal lobule 7A
> 42241
> 42575
> GM Superior parietal lobule 7M
> 5198
> 5598
> GM Superior parietal lobule 7P
> 20912
> 21330
> GM Visual cortex V1 BA17
> 39579
> 40643
> GM Visual cortex V2 BA18
> 46774
> 47310
> GM Visual cortex V3V
> 14404
> 14405
> GM Premotor cortex BA6
> 120102
> 121571
> WM Callosal body
> 104754
> 106316
> WM Fornix
> 13020
> 13584
> GM Mamillary body
> 1613
> 1738
> ```
> 
> I suspect these come from the last point of my previous review. When I replace the related lines by:
> 
> ```python
> # Build a mask of both halves of the brain
> middle_ind = (atlas.shape[0]) // 2
> # Put zeros on the median plane
> #atlas[middle_ind, ...] = 0
> # Split every zone crossing the median plane into two parts.
> left_atlas = atlas.copy()
> ```
> 
> I don't have these differences anymore.

Hello @NicolasGensollen , Indeed for the plane line it's not necessary because later with the left and right parts the mediane plane will be set to zero indirectly. As for the index of the mediane plane, it seems maybe that the number is always even so by subtracting one we are giving one index further to the left. I think now that it's all good.

NicolasGensollen (approved):
LGTM, thanks @achamma723 !

NicolasGensollen:
@jeromedockes I think all the points you raised before were addressed, but maybe you want to have another look before we merge this?

nilearn/datasets/atlas.py
jeromedockes: if the goal is to sort by index please use `names = [item[1] for item in sorted(names.items())]`

nilearn/datasets/atlas.py
```diff
@@ -377,31 +487,46 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+    atlas_niimg = check_niimg(atlas_img)
 
-    atlas_img = check_niimg(atlas_img)
     if lateralized:
-        return Bunch(maps=atlas_img, labels=names)
+        return Bunch(maps=atlas_niimg, labels=names)
 
-    atlas = get_data(atlas_img)
+    if not symmetric_split:
+        atlas = get_data(atlas_niimg)
```
jeromedockes, line 496: this code block which is intended for renaming some labels in discrete atlases gets executed for the probabilistic atlases (both for harvard-oxford and juelich) as well. as a result

- it takes a very long time
- results are incorrect for Juelich (all values are roughly divided by 2) . they are correct for harvard-oxford by coincidence, because none of the labels need to be merged in that case.

Once this is fixed, could you check that the function runs  and returns the correct output for all the atlases it can fetch?


IMO this function is too long and convoluted, with a control flow that is hard to follow, which is why it is difficult to get it to work. I believe simplifying it or breaking it down to smaller functions may make things easier, both for writing it now and for reading and maintaining it later. a few short comments can be useful as well. using more explicit variable names eg `is_already_lateralized` instead of `lateralized` can also help

jeromedockes (commented):
probabilistic atlases are treated as if they were discrete (please see inline comment)

NicolasGensollen:
@jeromedockes you're right, probabilistic atlases with `symmetric_split=False` are not handled correctly at all. :confused: 

For Juelich, since we have left and right labels, I think we need to merge the right and left maps corresponding to the same label in this case, but I might be wrong about that. 
@achamma723 I just pushed a commit to your branch which should fix this without modifying too much the current state of the function. (I think you also need to rebase/merge master to run the CI...). Please have a look and see if this makes sense.

>IMO this function is too long and convoluted, with a control flow that is hard to follow, which is why it is difficult to get it to work.

Yes I agree it is way too complex right now. I think we made a mistake when trying to combine both HarvardOxford and Juelich in the same function. Although they seemed similar, it looks like the control flows are quite different in the end.
Note that if we go back to separate fetchers, we might be able to extract and factorize common operations out.

bthirion:
> Yes I agree it is way too complex right now. I think we made a mistake when trying to combine both HarvardOxford and Juelich in the same function. Although they seemed similar, it looks like the control flows are quite different in the end.
> Note that if we go back to separate fetchers, we might be able to extract and factorize common operations out.

I think you're right. Sorry for the additional work !

NicolasGensollen:
@achamma723 I just pushed a code refactoring proposal to your branch. 
Functions are much smaller and have a more specific purpose, which should make the code more readable and easy to maintain. Let me know what you think (we can always revert my changes if you wanted to tackle this in a different way).

nilearn/datasets/atlas.py
```diff
@@ -326,93 +330,282 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+    new_names = np.unique([name[:-2] if name.endswith(' L')
+                           or name.endswith(' R') else name for name in names])
+    new_atlas_data = np.zeros((*atlas_data.shape[:3],
+                               len(new_names)-1))
+    for i,name in enumerate(new_names):
+        if name != "Background":
+            # If the name has no R or L, then simply copy the map
+            if name in names:
+                idx = names.index(name) - 1 # -1 because of background
+                new_atlas_data[...,(i-1)] = atlas_data[...,idx]
```
bthirion, line 547: pep8

nilearn/datasets/atlas.py
```diff
@@ -326,93 +330,282 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+            # otherwise, merge the maps
+            else:
+                idx_l = names.index(name + ' L') - 1
+                idx_r = names.index(name + ' R') - 1
+                new_atlas_data[...,(i-1)] = atlas_data[...,idx_l] + atlas_data[...,idx_r]
```
bthirion, line 552: and here

jeromedockes:
thanks @NicolasGensollen ! it seems more straightforward now. is it ready for another review?

NicolasGensollen:
>thanks @NicolasGensollen ! it seems more straightforward now. is it ready for another review?

I think it is! It would be easier if we had the CI though. @achamma723 could you merge master or rebase if you agree with the proposed changes?

bthirion (approved):
LGTM

achamma723:
> > thanks @NicolasGensollen ! it seems more straightforward now. is it ready for another review?
> 
> I think it is! It would be easier if we had the CI though. @achamma723 could you merge master or rebase if you agree with the proposed changes?

Hello @NicolasGensollen , thanks for your refactoring! I pushed all the changes and I still have an issue related with the covered lines to solve.

NicolasGensollen:
>Hello @NicolasGensollen , thanks for your refactoring! I pushed all the changes and I still have an issue related with the covered lines to solve.

@achamma723 your branch is behind master, you need to rebase or merge first

examples/01_plotting/plot_atlas.py
```diff
@@ -11,22 +11,39 @@
 
 from nilearn import datasets
 
-dataset = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
-atlas_filename = dataset.maps
+dataset_ho = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
+dataset_ju = datasets.fetch_atlas_juelich('maxprob-thr0-1mm')
 
-print('Atlas ROIs are located at: %s' % atlas_filename)
+atlas_ho_filename = dataset_ho.filename
```
NicolasGensollen, line 17: There is no `filename` key in this case because symmetric_split=False.

```suggestion
atlas_ho_filename = dataset_ho.maps
```

nilearn/datasets/atlas.py
```diff
@@ -320,104 +324,290 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+                        resume=True, verbose=1):
+    """Load Juelich parcellations from FSL.
+
+    This function downloads Juelich atlas packaged from FSL 5.0
+    and stores atlases in NILEARN_DATA folder in home directory.
+
+    This function can also load Juelich atlas from your local directory
+    specified by your FSL installed path given in `data_dir` argument.
+    See documentation for details.
+
```
NicolasGensollen, line 384: We can add `.. versionadded:: 0.8.1` here.

NicolasGensollen (commented):
Thanks @achamma723 !

I think the error you're getting on CircleCI is related to a missing key (see comment below).

nilearn/datasets/atlas.py
```diff
@@ -320,104 +324,290 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
-    left_atlas = atlas.copy()
+    left_atlas = atlas_data.copy()
     left_atlas[middle_ind:, ...] = 0
-    right_atlas = atlas.copy()
+    right_atlas = atlas_data.copy()
     right_atlas[:middle_ind, ...] = 0
 
+    if source == "Juelich":
+        for idx in range(len(names)):
+            if names[idx].endswith('L'):
```
jeromedockes, line 586: `re.sub(r" L$", ", left part", names[idx])`

nilearn/datasets/atlas.py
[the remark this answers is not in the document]
jeromedockes: as @NicolasGensollen said  right and left are swapped. the final result is correct because this turns out to compensate another bug (the orientation of the x axis ie the sign of the first entry of the affine, is ignored, see above)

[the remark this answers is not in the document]
jeromedockes: > But do we actually need all that ?

I agree with @bthirion : can't we do without all this splitting and merging stuff and just provide the standard versions?

jeromedockes (commented):
the fetcher fails for juelich, prob, symmetric_split=False.

nilearn/datasets/atlas.py
```diff
@@ -320,104 +324,290 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+    regions.
+    """
+    if atlas_data.ndim != 4:
+        raise ValueError("Expected a 4D atlas here.")
+    new_names = np.unique([name[:-2] if name.endswith(' L')
+                           or name.endswith(' R') else name for name in names])
+    new_atlas_data = np.zeros((*atlas_data.shape[:3],
+                               len(new_names) - 1))
+    for i, name in enumerate(new_names):
+        if name != "Background":
```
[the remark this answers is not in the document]
achamma723: I think it is the fact that atlas_data has the last dimension 121 but the names has a length of 122. Therefore, we excluded the Background while merging the maps while keeping it in the new names.

NicolasGensollen:
@achamma723 let me know when this is ready for another review (or if there are some blocking points right now).

>>But do we actually need all that ?

>I agree with @bthirion : can't we do without all this splitting and merging stuff and just provide the standard versions?

@jeromedockes You mean remove the `symmetric_split` argument of the Julich fetcher and return the regions and labels as they are?

achamma723:
> @achamma723 let me know when this is ready for another review (or if there are some blocking points right now).
> 
> > > But do we actually need all that ?
> 
> > I agree with @bthirion : can't we do without all this splitting and merging stuff and just provide the standard versions?
> 
> @jeromedockes You mean remove the `symmetric_split` argument of the Julich fetcher and return the regions and labels as they are?

Hello @NicolasGensollen , I think I covered most of the points raised by Jerome. There is still one point related to left and right atlas I didn't get the idea so I need maybe more explanation from his side. But why sometimes it tells me some of the lines are not covered by tests and another time it tells me nothing?

jeromedockes:
>>>But do we actually need all that ?

>>I agree with @bthirion : can't we do without all this splitting and merging stuff and just provide the standard versions?

> @jeromedockes You mean remove the `symmetric_split` argument of the Julich fetcher and return the regions and labels as they are?

Yes that was what I understood from @bthirion's comment and I thought it
may help simplify the code of the fetchers, but after discussion IRL it
seems having both splitted and merged versions for all atlases is
essential so please ignore that comment

nilearn/datasets/atlas.py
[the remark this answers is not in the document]
NicolasGensollen: @achamma723 we need to keep the "-1" in the `new_atlas_data` last dimension because we don't want to build a map for the background (see my last commit). @jeromedockes is this what you were referring to?

NicolasGensollen:
@achamma723 I pushed some changes to your branch to address the remaining comments of @jeromedockes 
Let me know if you agree with the changes and if you see other things.

Note: I'm not sure about the way I deal with the affine in order to swap the hemispheres or not (I just check the sign of the 0,0 element...). If there is a better or more robust way to do that, I'd be happy to update the code.

jeromedockes:
> Note: I'm not sure about the way I deal with the affine in order to swap the hemispheres or not (I just check the sign of the 0,0 element...). If there is a better or more robust way to do that, I'd be happy to update the code.

you can use nilearn.image.reorder_img.
it will reindex the image so that the affine is diagonal and positive

NicolasGensollen:
>you can use nilearn.image.reorder_img.

Great, thanks! :+1:

nilearn/datasets/atlas.py
```diff
@@ -320,104 +325,274 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
-    atlas_img = check_niimg(atlas_img)
-    if lateralized:
-        return Bunch(maps=atlas_img, labels=names)
-
-    atlas = get_data(atlas_img)
-
-    labels = np.unique(atlas)
+        new_idx = int(label.get('index')) + 1
+        if new_idx in names:
+            raise IndexError(
```
jeromedockes, line 512: maybe ValueError?

nilearn/datasets/atlas.py
```diff
@@ -320,104 +325,274 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+                f"Duplicate index {new_idx} for labels "
+                f"'{names[new_idx]}', and '{label.text}'")
+        names[new_idx] = label.text
+    names = [item[1] for item in sorted(names.items())]
```
jeromedockes, line 516: we could add an assert to check the keys() is the range 1, n_labels + 1 ?

nilearn/datasets/atlas.py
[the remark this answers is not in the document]
jeromedockes: yes it was -- why no label and no map for the background. we could maybe add a short comment explaining why it is not included and hence the `-1`?

nilearn/datasets/atlas.py
```diff
@@ -320,104 +325,274 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+        new_atlas_data[atlas_data == label] = new_names_dict[new_name]
+    return new_atlas_data, new_names
+
+
+def _compute_symmetric_split(source, atlas_niimg, names):
+    """Helper function for both fetch_atlas_juelich and
+    fetch_atlas_harvard_oxford.
+    This function handles 3D atlases when symmetric_split=True.
+    """
+    atlas_data = get_data(atlas_niimg)
```
jeromedockes, line 559: maybe assert atlas_niimg.affine[0, 0] > 0 to make it clear to someone reading this function, as the `reorder_img` happens somewhere else

nilearn/datasets/atlas.py
```diff
@@ -320,104 +325,274 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
-    right_atlas = atlas.copy()
-    right_atlas[:middle_ind, ...] = 0
+    left_atlas = atlas_data.copy()
+    left_atlas[middle_ind:] = 0
+    right_atlas = atlas_data.copy()
+    right_atlas[:middle_ind] = 0
+
+    if source == "Juelich":
+        for idx in range(len(names)):
+            if names[idx].endswith('L'):
```
jeromedockes, line 571: this condition is not needed anymore

nilearn/datasets/atlas.py
```diff
@@ -320,104 +325,274 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+                names[idx] = re.sub(r" L$", ", left part", names[idx])
+            if names[idx].endswith('R'):
```
jeromedockes, line 573: same

jeromedockes (commented):
thanks! I left a few small comments but we are getting there.
it would be nice to visually check all the atlases -- a few example regions (eg check that they are splitted iif required, that they are on the right side of the brain), number of regions, etc.
some parts of the functions are not covered by the tests

nilearn/datasets/atlas.py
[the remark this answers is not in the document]
NicolasGensollen: @jeromedockes maybe I'm missing something, but if you do (on master):

```python
from nilearn import datasets
from nilearn import image
d = datasets.fetch_atlas_harvard_oxford("sub-prob-1mm")
print(len(d.labels))
print(image.get_data(d.maps).shape)
```

You get:

```
22
(182, 218, 182, 21)
```

So my conclusion was that we don't return a map for the background (but the label does exist).
Does this make sense?

nilearn/datasets/atlas.py
NicolasGensollen: I think it is needed in order to have the labels converted from "label R" / "label L" to "Right label" / "Left label".

Note: I changed this in ac66400 because labels were not consistent, some were starting with "Right | Left" while others were ending with ",right part |,left part".

NicolasGensollen:
@jeromedockes I pushed a few commits to address most of your comments (and responded to the ones I didn't address yet).

>it would be nice to visually check all the atlases -- a few example regions (eg check that they are splitted iif required, that they are on the right side of the brain), number of regions, etc.

I made a "quick-and-dirty" script to check all atlases (see below). You can set `PLOT` to true to visualize splits. 
One potential issue I noticed is that, sometimes, some labels are not present in the image.

```python
import numpy as np
import copy
from nilearn import datasets, image, plotting
PLOT = False # Careful if set to True: heavy plotting...
fetchers = [datasets.fetch_atlas_harvard_oxford,
            datasets.fetch_atlas_juelich]
atlases = [["cort-maxprob-thr0-1mm", "cort-maxprob-thr0-2mm",
            "cort-maxprob-thr25-1mm", "cort-maxprob-thr25-2mm",
            "cort-maxprob-thr50-1mm", "cort-maxprob-thr50-2mm",
            "cort-prob-1mm", "cort-prob-2mm",
            "cortl-maxprob-thr0-1mm", "cortl-maxprob-thr0-2mm",
            "cortl-maxprob-thr25-1mm", "cortl-maxprob-thr25-2mm",
            "cortl-maxprob-thr50-1mm", "cortl-maxprob-thr50-2mm",
            "cortl-prob-1mm", "cortl-prob-2mm",
            "sub-maxprob-thr0-1mm", "sub-maxprob-thr0-2mm",
            "sub-maxprob-thr25-1mm", "sub-maxprob-thr25-2mm",
            "sub-maxprob-thr50-1mm", "sub-maxprob-thr50-2mm",
            "sub-prob-1mm", "sub-prob-2mm"],
           ["maxprob-thr0-1mm", "maxprob-thr0-2mm",
            "maxprob-thr25-1mm", "maxprob-thr25-2mm",
            "maxprob-thr50-1mm", "maxprob-thr50-2mm",
            "prob-1mm", "prob-2mm"]]
for i,fetcher in enumerate(fetchers):
    print(f"*** {fetcher.__name__} ***")
    for atlas in atlases[i]:
        print(f"--- {atlas} ---")
        if '-prob-' in atlas or atlas.startswith('prob'):
            d = fetcher(atlas, symmetric_split=False)
            print(len(d.labels))
            print(image.get_data(d.maps).shape)
            assert len(d.labels) == image.get_data(d.maps).shape[-1] + 1
        else:
            for sym_split in [True, False]:
                d = fetcher(atlas, symmetric_split=sym_split)
                print(f"symmetric_split = {sym_split}")
                print(len(d.labels))
                print(image.get_data(d.maps).shape)
                # Check that the label values range from 0 to len(labels)
                label_values_in_img = np.unique(image.get_data(d.maps))
                label_values = list(range(len(d.labels)))
                if sorted(label_values_in_img) != label_values:
                    print("Following labels are not in image:")
                    for missing in set(label_values) - set(label_values_in_img):
                        print(d.labels[missing])
                if sym_split:
                    for l_idx, l in enumerate(d.labels):
                        if l != "Background":
                            assert "Right" in l or "Left" in l
                            if PLOT:
                                print(l)
                                img_data = copy.deepcopy(image.get_data(d.maps))
                                img_data[img_data != l_idx] = 0
                                plotting.plot_img(image.new_img_like(d.maps, img_data))
                                plt.show()
```

nilearn/datasets/atlas.py
jeromedockes: then you could simply use `re.sub(r"(.*) L$", r"Left \1", name)`, without a need for the if?

(sorry if this is outdated something weird seems to have happened with the comments in this PR)

jeromedockes:
> I made a "quick-and-dirty" script to check all atlases (see below).

awesome thanks!


>  One potential issue I noticed is that, sometimes, some labels are not present in the image.

I remember this being a source of confusion in the issue tracker or or neurostars for other atlases, because in that case the NiftiLabelsMasker doesn't assign a dimension to that label -- but I guess that is more a problem of documentation of the NiftiLabelsMasker than of the atlas, and I believe it happens with some other atlases as well. So I don't think having some empty regions is a huge problem

bthirion (approved):

NicolasGensollen:
Hi @achamma723 
Did you have a chance to take a look at this? It clearly needs to be rebased since there are a few conflicts, but we were almost done back in early July if I remember correctly.

achamma723:
> Hi @achamma723
> Did you have a chance to take a look at this? It clearly needs to be rebased since there are a few conflicts, but we were almost done back in early July if I remember correctly.

Hello @NicolasGensollen, Sorry I was on vacation so I couldn't take a look. Everything was good back on July, there is still some small modifications to do in the test part.

bthirion:
@achamma723 Can you proceed with the finalization of this PR ? It would be great to merge it before next week to have it in the upcoming release.

achamma723:
Hello @NicolasGensollen, I think it's all good now!

nilearn/datasets/tests/test_atlas.py
```diff
@@ -243,6 +294,130 @@ def test_fail_fetch_atlas_harvard_oxford(tmp_path, request_mocker):
...
+    # This section tests with lateralized version. In other words,
+    # symmetric_split=True
+
+    # We test the fetcher with symmetric_split=True by creating a new
+    # dummy local file and fetch them and test the output variables
+    # accordingly.
+    with open(os.path.join(ho_dir, 'Juelich.xml'), 'w') as dummy2:
+        dummy2.write("<?xml version='1.0' encoding='us-ascii'?>\n"
+                     "<data>\n"
+                     '<label index="0" x="125" y="75" z="107">R1</label>\n'
```
NicolasGensollen, line 392: It looks like there is a lot of repetitions and copy-pasting in this function.
Could we factorize this? Maybe with a pytest fixture?

NicolasGensollen (commented):
Thanks @achamma723 !
See my comment below on DRYing the test functions a little bit.
Could you also add a whatsnew entry ?

nilearn/datasets/tests/test_atlas.py
achamma723: Yes you are right. There was some unnecessary lines so I deleted them. Also, I refactored using ```_write_to_xml``` and ```_test_result_xml```.

bthirion (approved):
Can be merged as far as I can see.

NicolasGensollen:
@achamma723 I refactored the tests further in 76e2ffa
Let me know if this is good from your perspective.

achamma723:
> @achamma723 I refactored the tests further in [76e2ffa](https://github.com/nilearn/nilearn/commit/76e2ffa75a6417ae7c6bebde1377fe7f757d00cf)
> Let me know if this is good from your perspective.

It's very good! I understand now the importance of pytest fixture and parametrize :+1:

NicolasGensollen (approved):
I also think this can be merged. 
It is getting difficult keeping track of discussions on this PR, but maybe @jeromedockes wants to have another look.

NicolasGensollen:
@achamma723 could you fix the conflict in whats_new ?
I will go ahead and merge it once this is done since we have two approvals.

achamma723:
> @achamma723 could you fix the conflict in whats_new ?
> I will go ahead and merge it once this is done since we have two approvals.

@NicolasGensollen It's all set!

NicolasGensollen:
Thanks! Merging then...

bthirion:
Thx @achamma723  !

Issue #460, opened 2015-02-24
Calling mask.invert_transform(transform(img)) on a 3D image does not return a 3D image.


bcipolli:
Calling `mask.invert_transform(transform(img))` on a 3D image appends a 4th dimension.  

To me, this is semantically wrong--if I start with 3D, I should finish with 3D.

The practical problem I hit was while using `nipy`'s plotting functions.  They require 3D data, and this kind of unexpected 4D image made it hard to write generic code there--all other images I was working on were 3D (including others via nilearn functions like `index_img`), but only this one was not.

Here's a script to reproduce the issue.

```python
import nibabel
from nilearn import datasets
from nilearn.image import index_img
from nilearn.input_data import NiftiMasker

haxby_dataset = datasets.fetch_haxby()
mask_filename = haxby_dataset.mask_vt[0]
func_filename = haxby_dataset.func[0]

nifti_masker = NiftiMasker(mask_img=mask_filename, standardize=True)
fmri_img = index_img(nibabel.load(func_filename), 0)  # 3D image
nifti_masker.fit(fmri_img)
print nifti_masker.mask_img_.shape  # 3D mask

fmri_masked = nifti_masker.transform(fmri_img)
fmri_unmasked = nifti_masker.inverse_transform(fmri_masked)
print fmri_img.shape, fmri_unmasked.shape  # 3D start image, 4D end image
```

bcipolli:
I debugged this for a while, and the introduction of the 4th dimenions occurs in `check_niimgs`, which converts a 3D image to 4D (even when `allow_3D=True`).  Once that's done, there's no way to return a 1D vector from `transform`, and so `inverse_transform` also must return 4D.

Not sure if there is any simple fix.

AlexandreAbraham:
We are aware of that and this point is still under discussion. We couldn't come with a satisfying answer yet.

GaelVaroquaux:
Well, the problem is that there is intrinsically a loss of information:
the output of transform must be 2D to obey scikit-learn's convention.
Thus a 3D (n_x, n_y, n_z) image and a 4D (n_x, n_y, n_z, 1) image are
cast to the same 'X' matrix. For one of these two, 
inverse_transform(transform()) will fail to give the same shape as the
original.

We are open to changing the current convention, ie to always cast to 4D,
if you give us good arguments.

AlexandreAbraham:
[...]

PR #2725, opened 2021-03-09
[FIX]  Allow every masker.inverse_transform to accept 1D input.

### LINKED ISSUE #2724: support 1D inputs with inverse_transform
Some maskers support 1D input for `inverse_transform`, while others do not

Nilearn version: 0.7

### Expected behavior

Consistent behavior across maskers (all accept 1D or do not)
(the [base_masker](https://github.com/nilearn/nilearn/blob/c8ecb48ed953c9fb20372e0e110fc42ff3f96145/nilearn/input_data/base_masker.py#L243-L264) accepts [1D input](https://github.com/nilearn/nilearn/blob/5818857e1786985b0d6ce2109929a89147c95090/nilearn/masking.py#L882) so I'm inclined to have `inverse_transform` accept 1D input.
(also biased because [NiMARE](https://github.com/neurostuff/NiMARE) has a lot of cases where we just want to transform 1 statistical map and have the output be 3D)

### Actual behavior

`NiftiMasker` accepts 1D input for `inverse_transform`, but `NiftiLabelsMasker` does not (and presumably other maskers may or may not support 1D inputs)

TBA
```python

```

jdkent:
Closes #2724.

- allow 1D inputs to `signals_to_img_labels`
- test the function with 1D input
- propagate documentation changes

### DIFF
--- a/nilearn/input_data/nifti_labels_masker.py
+++ b/nilearn/input_data/nifti_labels_masker.py
@@ -344,9 +344,10 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
 
         Parameters
         ----------
-        signals : (2D numpy.ndarray)
+        signals : (1D/2D numpy.ndarray)
             Signal for each region.
             shape: (number of scans, number of regions)
+            If signals is one-dimensional, it is assumed that number of scans == 1.
 
         Returns
         -------
--- a/nilearn/regions/parcellations.py
+++ b/nilearn/regions/parcellations.py
@@ -482,8 +482,9 @@ class Parcellations(MultiPCA):
 
         Parameters
         ----------
-        signals : List of 2D numpy.ndarray
+        signals : List of 1/2D numpy.ndarray
             Each 2D array with shape (number of scans, number of regions).
+            If array is one-dimensional, it is assumed that number of scans == 1.
 
         Returns
         -------
--- a/nilearn/regions/signal_extraction.py
+++ b/nilearn/regions/signal_extraction.py
@@ -146,6 +146,7 @@ def signals_to_img_labels(signals, labels_img, mask_img=None,
     ----------
     signals : numpy.ndarray
         2D array with shape: (scan number, number of regions in labels_img).
+        If signals is one-dimensional, it is assumed that scan number == 1.
 
     labels_img : Niimg-like object
         See http://nilearn.github.io/manipulating_images/input_output.html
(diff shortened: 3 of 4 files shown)


NicolasGensollen (changes requested):
Thanks for opening this @jdkent ! :+1: 
I think the changes look good but only ensure that the `NiftiLabelsMasker` can inverse transform 1D inputs. I think that the `NiftiMapsMasker` and the `NiftiSpheresMasker` are also restricted to 2D inputs:

https://github.com/nilearn/nilearn/blob/80572dbf2253bd7a5a00e688082cbd3075e71a12/nilearn/input_data/nifti_maps_masker.py#L356-L365

Also, I think it could be good to add a small test for the inverse transform method of the `NiftiLabelsMasker`, maybe here:
https://github.com/nilearn/nilearn/blob/80572dbf2253bd7a5a00e688082cbd3075e71a12/nilearn/input_data/tests/test_nifti_masker.py#L125-L130

tsalo:
Just to make sure I understand the proposal- between this PR and #2708, the maskers will treat 3D/1D inputs as 4D/2D, respectively, right? I.e., the outputs of both `transform` and `inverse_transform` will have a time element regardless. Is that correct?

jdkent:
@tsalo, great point you brought up, for `inverse_transform`, if a 1D input is passed in then a 3D output will be produced, but looking at https://github.com/nilearn/nilearn/issues/460#issuecomment-100767188, I see that consistency of output was deemed more important, if that decision still holds, `inverse_transform` should always return 4D. Although, the functionality for unmask was [implemented before that conversation](https://github.com/nilearn/nilearn/pull/42).

I do not have a strong opinion on this (but someone should!), for maintainability I believe the 2D/4D output should be enforced, but on the other hand unmask has supported 1D inputs for years suggesting people may depend on that behavior of unmask and the transformers that call unmask.

jdkent:
@bthirion, that sounds good.

Enforcing 4D output implies a bug was introduced ~8 years ago, when `unmask` allowed 1[D inputs and returned 3D outputs](https://github.com/nilearn/nilearn/pull/42/files#diff-ec47025e5fd3f9fb5ad16ef77a0e6aef29507c171ff3720c5c0cd6f9a4e9ba42R398-R401), and the `unmask` function was called in base_masker.py under the `inverse_transform` method.

In proper context, my usage of `inverse_transform` on 1D input and getting 3D output in NiMARE is a long-standing bug (not a feature), that should be fixed.

jdkent:
Alright, I've added tests to all the relevant extractors/maskers, with reference to #2726, The good thing is that now every extractor supports 1D input to `inverse_transform`.

NicolasGensollen (changes requested):
The changes look good to me although I think we should wait to see if we reach a consensus in #2726  before merging this PR. 
Also, I think we should update the docstrings of `inverse_transform` for all maskers in the same way you did with the `NiftiLabelsMasker`.

ymzayek:
Closing this one since #3322 was merged

Issue #2730, opened 2021-03-12
Should not detrend go before low and high-pass filtering too for confounders in signal.clean?


jrasero:
Hi,

I don't know whether this is an issue or not, but I am trying to see if the deconfounding provided by the function _**signal.clean**_ works effectively or not on my data and I think it is not behaving as I expect, in particular when I combine it with a high-pass filtering. I know there's been a lot of issues around all these, with a recent pull-request #2609, which however does not seem to resolve my (apparent) issue.

Since **_signal.clean_** also applies the high-pass filtering to the confounds in order to keep this filter orthogonal, in order to see if the confounds removal works, I need to check that the correlation between the cleaned signal and the detrended and high-pass filtered confounders is indeed zero. 

Here is what I've done and where the issue arises. First, just with detrending and then confounds removal:

```
import numpy as np
from nilearn.signal import clean
import matplotlib.pylab as plt

corrs_x_c = []
for ii in range(1000):
    c = np.random.randn(100)[:,None]
    X = c + np.random.randn(100)[:,None]

    # Detrend + deconfounding
    X_clean = clean(X, detrend = True, high_pass = None, standardize_confounds=True, standardize=False, confounds=c)
    c_clean = clean(c, detrend = True, high_pass = None, standardize=False)
    corrs_x_c.append(np.corrcoef(X_clean.T, c_clean.T)[0,1])
plt.hist(corrs_x_c)
plt.title("Correlation between X_clean and c_clean")
```

![image](https://user-images.githubusercontent.com/25054369/110956473-a96ed980-8318-11eb-8107-84b90ae89fd1.png)

So here, the behaviour is the expected one. 

The problem is when I insert a high-pass filtering (and I guess with a low-pass will happen too). Following the same example:

```
import numpy as np
from nilearn.signal import clean
import matplotlib.pylab as plt

corrs_x_c = []
for ii in range(1000):
    c = np.random.randn(100)[:,None]
    X = c + np.random.randn(100)[:,None]

    # Detrend + High-pass filtering + deconfounding
    X_clean = clean(X, detrend=True, high_pass = 0.01, standardize_confounds=True, standardize=False, confounds=c)
    c_clean = clean(c, detrend=True, high_pass = 0.01, standardize=False)
    corrs_x_c.append(np.corrcoef(X_clean.T, c_clean.T)[0,1])
plt.hist(corrs_x_c)
plt.title("Correlation between X_clean and c_clean")
```

[...]

PR #2732, opened 2021-03-12
Detrending before low and high pass filtering for the confounders in clean function

jrasero:

- Fixes the issue raised in #2730 
- In particular, it changes the order of detrending and low and high pass filtering in `clean` such that it is the same that takes place with the input signal, i.e, first detrend and then frequency filtering. Currently, confounders were first low or high-pass filtered and then detrended, if requested from the beginning.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -7,6 +7,12 @@ NEW
 Fixes
 -----
 
+- Fix detrending and temporal filtering order for confounders
+  in :func:`nilearn.signal.clean`, so that these operations are applied
+  in the same order as for the signals, i.e., first detrending and
+  then temporal filtering (https://github.com/nilearn/nilearn/issues/2730).
+
+
 Enhancements
 ------------
 
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -600,6 +600,13 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
     # Remove confounds
     if confounds is not None:
         confounds = _ensure_float(confounds)
+        
+        # Apply detrend to keep this operation orthogonal
+        # (according to Lindquist et al. (2018))
+        if detrend:
+            confounds = _standardize(confounds, standardize=False,
+                                     detrend=detrend)
+         
         # Apply low- and high-pass filters to keep filters orthogonal
         # (according to Lindquist et al. (2018))
         if low_pass is not None or high_pass is not None:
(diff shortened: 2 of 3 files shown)


bthirion (commented):
This looks good. Thx !
You simply need to add a test checking that the new behavior is enforced. This would be driectly derived from the example in #2730

NicolasGensollen (commented):
There is a test failure for `test_clean_confounds()` on MacOS actions. I think we should increase the threshold on this line:

https://github.com/nilearn/nilearn/blob/e8d5f5dddaede697da1eb821a240ece52172c43f/nilearn/tests/test_signal.py#L436

This would also be a good place to add a test derived from the examples in #2730

jrasero:
Thanks for the feedback @bthirion , @NicolasGensollen. Just to be sure, as for the test checking, is it then OK if I add a few lines to `test_clean_confounds` function where I assert the expected behaviour?

NicolasGensollen:
>is it then OK if I add a few lines to test_clean_confounds function where I assert the expected behaviour?

Absolutely. I think this test would make sense either there, or in 

https://github.com/nilearn/nilearn/blob/e8d5f5dddaede697da1eb821a240ece52172c43f/nilearn/tests/test_signal.py#L288

bthirion (commented):
LGTM, thx.
Maybe simply add a whatsnew entry.

NicolasGensollen (changes requested):
Thanks @jrasero !
This looks good to me except one test currently failing on macOS:

https://github.com/nilearn/nilearn/blob/e747be0b3c39e431a83b0665b6689a67ba63d47b/nilearn/tests/test_signal.py#L432-L437

 As this test is passing on other OS, I think we should try to increase the threshold a little bit.
Also, could you add a whatsnew entry (in the 0.7.2.dev bugfix section of `nilearn/doc/whats_new.rst`)?

jrasero:
Hi @bthirion and @NicolasGensollen ,

As requested, I've added an entry to the whatsnew document.

Let me know if there's anything else to be done regarding this PR.

bthirion:
There is still a numerical error. in test_confounds.py, you need to increase the bound to 1000 * eps instead of 200 * eps.
Best,

jrasero:
> There is still a numerical error. in test_confounds.py, you need to increase the bound to 1000 * eps instead of 200 * eps.
> Best,

I increased the bound as suggested.

 It's strange, because I didn't touch this part of the function. Was this error not being happening before?

NicolasGensollen (approved):
LGTM 
Thanks @jrasero :+1:

bthirion (approved):
LGTM, thx !

Issue #2631, opened 2020-12-08
seeding in `_utils.data_gen`


jeromedockes:
- `utils.data_gen.generate_fake_fmri` creates a `np.random.RandomState` (mutable
  default argument) in its definition. This might actually have been intentional
  so that it returns a different output each time it is called, but it is not
  the behavior of other functions in this module, which use 0 as a fixed seed by
  default, and it means that the data seen eg by unit tests depends on the order
  in which tests are executed, which may cause spurious failures. instead, this
  function should initialize a new random generator by default, and callers can
  call it with different seeds or pass an existing generator.

- most other functions in the same module, eg `generate_timeseries`, have a
  `rand_gen` parameter that has a different behavior than the `random_state`
  used in many nilearn and scikit-learn functions. Indeed, with `rand_gen` (ie
  in `data_gen`):
     - `None` -> `RandomState(0)`
     - anything else must be a random number generator
  whereas with scikit-learn's `check_random_state`:
     - `None` -> `RandomState(None)` ie get a seed from `/dev/urandom` or the clock
     - an int -> use that as a seed
     - a random number generator -> use that as-is
  this is not a big issue but as `_utils.data_gen` is a private module making
  the api more similar to `check_random_state` is doable.

rawwar:
Hi, I want to contribute to this issue. I am working on it. I will update this comment or create another one to notify if I no longer am working on this one.

PR #2741, opened 2021-03-28
[WIP] fix for #2631 by using check_random_state to initialize rand_gen

rawwar:
This PR is for #2631 
I have made the following changes
1. updated `generate_fake_fmri` function's `rand_gen`  parameter to default to `None`
2. checked if `rand_gen` is None, if `None`, used `check_random_state` to get a random generator


nilearn/_utils/data_gen.py
```diff
@@ -274,6 +274,8 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
     width = [s // 2 for s in shape]
     shift = [s // 4 for s in shape]
 
+    if rand_gen is None:
```
NicolasGensollen, line 277: I think you should call `check_random_state` without any condition since the user can pass either None, an int, or a random state as input, and the scikit-learn function is clever enough to handle this.

NicolasGensollen (changes requested):
Hi @rawwar 
Thanks for working on this! :+1: 
I had a quick look even though this is still WIP. You might be planning to do this already, but I think we should make the API more consistent. 

Indeed, `generate_mni_space_img`, `generate_signals_from_precisions`, and  `generate_group_sparse_gaussian_graphs` have a random state argument defaulting at zero, and `check_random_state` is called. I think this is the behavior we should implement in other functions using randomness:

- `generate_timeseries`, `generate_regions_ts`, and `generate_labeled_regions_large` should use `check_random_state` with a default seed set at 0.
- For some reason `generate_fake_fmri_data_and_design`, `write_fake_fmri_data_and_design`, `write_fake_bold_img`, `basic_confounds`, and  `create_fake_bids_dataset` don't rely on a random generator. I think we should change them to behave the same way as the other functions of this module.
- Update the docstrings to reflect these changes.

WDYT?

rawwar:
I agree with you. I did go through the module and noticed the same. I'll Work on updating and standardizing others as well. Thanks!

nilearn/_utils/data_gen.py
rawwar: Sure. I will update it

rawwar: I have fixed this in the latest commit

rawwar:
I have put comments with `#TODO`  text in the file where I needed specific inputs on what to update. I am very new to the domain. So, I am still learning. Please help me with those

nilearn/_utils/data_gen.py
```diff
@@ -330,51 +424,159 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
+    Parameters
+    ----------
+    file_path : str
+        output file path.
+
+    shape : tuple
+        Shape of output array.
+
+    rk : int, optional
+        #TODO what is rk.
```
NicolasGensollen, line 553: It's a good question indeed... 
this parameter isn't used in this function. I suppose it got copy-pasted at some point. 
I suggest we remove it.

NicolasGensollen (changes requested):
Thanks @rawwar ! This is great! :+1: 
The diff is quite large, I made a first pass and tried to answer your questions. I might have missed some of them though.

I also have some general comments:

- It looks like you accidentally committed some unrelated files, could you delete them?
- Despite the fact that `_utils.data_gen` is a private module, you are changing the api which might cause failures in users code. I'm wondering whether we should be a little bit more careful and deprecate `rang_gen` to `random_state`. For example with:

https://github.com/nilearn/nilearn/blob/05ade8989ece76ca8a5cbcb222a64be26e94d461/nilearn/_utils/helpers.py#L5-L11

Pinging @jeromedockes who might have an opinion on this.

- It looks like some functions (like `generate_labeled_regions_large`, `write_fake_bold_img` or ` generate_signals_from_precisions` for example) aren't unit-tested which makes me a little nervous :worried: I'll open an issue on this...

rawwar:
> * Despite the fact that `_utils.data_gen` is a private module, you are changing the api which might cause failures in users code. I'm wondering whether we should be a little bit more careful and deprecate `rang_gen` to `random_state`. For example with:

I agree about the fact that I shouldn't change the API. I was thinking if we should add one more argument `random_state` along with `rand_gen` for functions that are using `rand_gen`. Warn users about the deprecation and remove it in the next major release?

> * It looks like some functions (like generate_labeled_regions_large, write_fake_bold_img or generate_signals_from_precisions for example) aren't unit-tested which makes me a little nervous 😟 I'll open an issue on this...

I am actually working on this. I have been studying what these functions actually do or how they are used.  I am also going through nilearn documentation and reading about neuro-imaging. Hopefully, I will know the basics of this domain soon enough. Also, if you could point me to resources that one must know to be able to contribute to this package, do mention. Thanks 😄

NicolasGensollen:
>I deleted them

Thanks! :+1: 

>I agree about the fact that I shouldn't change the API. I was thinking if we should add one more argument random_state along with rand_gen for functions that are using rand_gen. Warn users about the deprecation and remove it in the next major release?

This module is private, so we could be less careful than with other parts of nilearn. That being said, I don't think it is too costly to deprecate `rand_gen` for `random_state`. Here is an example:

```python
from nilearn._utils import rename_parameters # This would go at the beginning of the file...
@rename_parameters({'rand_gen': 'random_state'}, '0.9')
def generate_timeseries(n_instants, n_features, random_state=0):
        """Generate some random timeseries. """
        rand_gen = check_random_state(random_state)
        return rand_gen.randn(n_instants, n_features) 
```

Which gives you:

```
>>> generate_timeseries(10, 3, rand_gen=0)
<stdin>:1: FutureWarning: The parameter "rand_gen" will be removed in 0.9 release of Nilearn. Please use the parameter "random_state" instead.
array([[ 1.76405235,  0.40015721,  0.97873798],
       [ 2.2408932 ,  1.86755799, -0.97727788],
       [ 0.95008842, -0.15135721, -0.10321885],
...
```

>I am actually working on this.

Awesome! :+1: (FYI, I opened #2750 on this).

> Also, if you could point me to resources that one must know to be able to contribute to this package, do mention

In terms of contributing guidelines, you can refer to [this page](http://nilearn.github.io/development.html). We are planning to add more content there in a near future, but this should give the basics.

localizer_events.tsv
```diff
@@ -0,0 +1,81 @@
...
+vertical checkerboard	264.0	1.0
+horizontal checkerboard	266.7	1.0
+horizontal checkerboard	269.7	1.0
+right button press, visual instructions	275.4	1.0
+vertical checkerboard	278.4	1.0
+auditory sentence	284.4	1.0
+visual sentence	288.0	1.0
+right button press, visual instructions	291.0	1.0
+auditory sentence	293.4	1.0
+auditory sentence	296.7	1.0
```
bthirion: The two above  files should probably not be here

jeromedockes:
> Pinging @jeromedockes who might have an opinion on this.

sorry I completely missed that. I would say that `_utils` is a "private" package so we don't need to worry about backwards compatibility and we can change it without notice.

btw these two examples
```
04_glm_first_level/plot_bids_features.py
03_connectivity/plot_simulated_connectome.py
```
set a bad example for users by importing functions from this "private" package (they are not affected by this pr though)

ymzayek:
I can take this one over.

PR #2729, opened 2021-03-10
Add Pypi badge

NicolasGensollen:
I thought this could be useful. We can close this otherwise.

PR #2749, opened 2021-04-02
[MAINT] Test with Python 3.9

NicolasGensollen:
Add Python 3.9 to build matrix.
Thanks to @emdupre for catching this! :+1: 

Note: Currently failing as expected. We need to merge #2751 before.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -63,7 +63,7 @@ jobs:
         fail-fast: false
         matrix:
             os: ["ubuntu-latest", "macos-latest"]
-            python-version: ["3.6", "3.7", "3.8"]
+            python-version: ["3.6", "3.7", "3.8", "3.9"]
     name: ${{ matrix.os }} with Python ${{ matrix.python-version }} and latest package versions
     defaults:
       run:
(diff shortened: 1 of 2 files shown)


emdupre:
Thanks so much, @NicolasGensollen ! 

It looks like the only failing test is indeed in `fetch_atlas_aal`. I think that we'll want to change [these lines](https://github.com/nilearn/nilearn/blob/05ade8989ece76ca8a5cbcb222a64be26e94d461/nilearn/datasets/atlas.py#L799-L804).

tsalo:
It might also be worth adding 3.8 and 3.9 under the classifiers in setup.py:

https://github.com/nilearn/nilearn/blob/05ade8989ece76ca8a5cbcb222a64be26e94d461/setup.py#L89-L91

Also, what about a badge for the Python versions in the README? 
[PyPI - Python Version](https://pypi.python.org/pypi/nilearn/)

NicolasGensollen:
>I think that we'll want to change these lines.

@emdupre yes indeed, I opened #2751 as a fix.

NicolasGensollen:
@tsalo adding the badge in #2729 seems indeed better. 
I'll add python 3.8 and 3.9 to setup.py in this one, thanks for the suggestion! :+1:

bthirion (approved):
Thx !

NicolasGensollen:
Thanks! Merging...

PR #2731, opened 2021-03-12
[DOC][MRG] Add glossary

NicolasGensollen:
Fixes #75

PR #2758, opened 2021-04-12
[FIX] Fix glossary broken refs

NicolasGensollen:
Fix a couple glossary broken references introduced in #2731


bthirion (approved):
LGTM, thx.

PR #2759, opened 2021-04-13
[DOC][MRG] Configure bibtex references

NicolasGensollen:
## Description

This PR proposes to use [sphinxcontrib-bibtex](https://github.com/mcmtroffaes/sphinxcontrib-bibtex) to manage references in the documentation. This PR configures the extension and provides a couple examples. 
Other references will have to be added to the bib file and the docstrings will have to be updated. I will open small "good first issues" for the sprint targeting the different files which need to be updated. This is what [MNE](https://github.com/mne-tools/mne-python) did for their last sprint and I think it was a very good way to onboard new people with a very simple first PR. 
WDYT?

## Advantages

This will have the following advantages:

- consistent formatting of references across doc
- no more copy-pasting of references, they are all stored in `references.bib`
- easier maintenance
- Auto generated global bibliography page (this PR adds a link in the website navbar, but we can have it somewhere else...)

biblio

## Drawbacks

But the following drawbacks:

- The raw docstrings won't render the reference, you will only see the bib key.

### DIFF
--- a/doc/bibliography.rst
+++ b/doc/bibliography.rst
@@ -1,0 +1,12 @@
+:orphan:
+
+.. _general_bibliography:
+
+General bibliography
+====================
+
+The references below are arranged alphabetically by first author.
+
+.. bibliography:: ./references.bib
+   :all:
+   :list: enumerated
--- a/doc/bibliography.rst
+++ b/doc/bibliography.rst
@@ -5,7 +5,7 @@
 General bibliography
 ====================
 
-The references below are arranged alphabetically by first author.
+The references below are arranged alphabetically by first author. You can download the bib file :download:`here <./references.bib>`.
 
 .. bibliography:: ./references.bib
    :all:
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -51,6 +51,7 @@ extensions = [
               'sphinx.ext.autosummary',
               'sphinx.ext.imgmath',
               'sphinx.ext.intersphinx',
+              'sphinxcontrib.bibtex',
               'numpydoc',
               ]
 
@@ -87,6 +88,11 @@ plot_gallery = 'True'
 
 # The master toctree document.
 master_doc = 'index'
+
+# sphinxcontrib-bibtex
+bibtex_bibfiles = ['./references.bib']
+bibtex_style = 'unsrt'
+bibtex_footbibliography_header = ''
 
 # General information about the project.
 project = u'Nilearn'
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -92,6 +92,7 @@ master_doc = 'index'
 # sphinxcontrib-bibtex
 bibtex_bibfiles = ['./references.bib']
 bibtex_style = 'unsrt'
+bibtex_reference_style = 'author_year'
 bibtex_footbibliography_header = ''
 
 # General information about the project.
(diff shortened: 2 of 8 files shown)


tsalo:
This is awesome! I think the rendering of the docstring links looks a little weird (e.g., `See 1`)- is it possible to use author-year citations instead? It might also be a good idea to add some info to `authors.rst`, since that's where citation info is. Just something about how to incorporate the new reference info into your manuscripts (e.g., download the bib file, identify relevant citations from the docstrings of methods you used, etc.).

I assume that there is no mechanism for automatically inheriting references from called functions/classes like there is with `duecredit`, so the contributing page may require a caveat that contributors should try to add citations for any and all `nilearn` functions that are _used_ by a given function as well as its main citation.

emdupre:
>  I think the rendering of the docstring links looks a little weird (e.g., See 1)- is it possible to use author-year citations instead?

sphinxcontrib-bbibtex supports this -- we'd just need to [set the bibliography style](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#referencing-style) to `author_year`. Jupyter Book has [good docs on this](https://jupyterbook.org/content/citations.html?highlight=citation#basic-citations); it's directly passing all options sphinxcontrib-bibtex under the hood.

doc/conf.py
```diff
@@ -88,6 +89,11 @@
 # The master toctree document.
 master_doc = 'index'
 
+# sphinxcontrib-bibtex
+bibtex_bibfiles = ['./references.bib']
+bibtex_style = 'unsrt'
```
emdupre, line 94:
```suggestion
bibtex_style = 'unsrt'
bibtex_reference_style = 'author_year'
```

doc/bibliography.rst
```diff
@@ -0,0 +1,12 @@
+:orphan:
+
+.. _general_bibliography:
+
+General bibliography
+====================
+
+The references below are arranged alphabetically by first author.
```
tsalo, line 8: Perhaps include a link to the actual bib file to download here as well?

doc/conf.py
NicolasGensollen: Thanks @emdupre ! :+1: 
It looks like `bibtex_style` and `bibtex_reference_style` are not taken into account. I tried all possible values locally and nothing seems to change the rendering... :confused: 
Setting them to invalid values does raise an error though.
I'm probably missing something here...

NicolasGensollen: It looks like `bibtex_reference_style` currently only works with `:cite:t:` and `:cite:p:`, see https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/240 although they are planning to support footcite at some point.

I'll keep an eye on this, but I think we can keep the current formatting while waiting for author-year style to become available. WDYT?

bthirion: I am fine with the plan, if nobody has a solution to author-year display. You may want to contribute upstream to sphinx ;-)

NicolasGensollen:
Should we merge this?

doc/conf.py
emdupre: Sounds good to me ! Thanks, @NicolasGensollen !

bthirion (approved):
Thx !

NicolasGensollen:
Thank you for the reviews! :+1: 
Merging...

PR #2764, opened 2021-04-16
Improve testing/coverage of first level glm

NicolasGensollen:
This PR proposes to improve the testing of `nilearn/glm/first_level/first_level.py` since the coverage of this file was significantly (around 74%) lower than for most others. 
It also adds a few safety checks in `fit` to avoid unexpected `AttributeErrors` to occur in some edge cases.


bthirion (approved):
Thx ! 
I had observed that the coverage could be improved...

NicolasGensollen:
Thanks! Merging...

PR #2773, opened 2021-04-29
Fix artifact redirect action

### LINKED ISSUE #2508: restore link to artifacts in github PR pages

NicolasGensollen:
Fixes #2508

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,5 +11,5 @@ jobs:
               uses: larsoner/circleci-artifacts-redirector-action@master
               with:
                   repo-token: ${{ secrets.GITHUB_TOKEN  }}
-                  artifact-path: 0/doc/_build/html/index.html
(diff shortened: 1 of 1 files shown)


jeromedockes (approved):
thanks!
(note for future reference: doc/_build/html/ is the path to the build output on the machine, 0/dev/ is the stored artifact path).
btw maybe we could call these artifacts 'stable_doc' and 'dev_doc' rather than just 'stable' and 'dev'?

NicolasGensollen:
thanks for the review @jeromedockes !

>maybe we could call these artifacts 'stable_doc' and 'dev_doc' rather than just 'stable' and 'dev'?

I don't have a strong opinion on this. I'm OK with both.

PR #2792, opened 2021-05-05
First Level get_voxelwise_model_attribute should return as many elements as # design matrices

### LINKED ISSUE #2772: `FirstLevelModel._get_voxelwise_model_attribute` only returns first design matrix's attribute
The FirstLevelModel attributes which use `_get_voxelwise_model_attribute()` only return the img for the first design matrix, rather than all of the design matrices' associated imgs.

Nilearn version: ~0.7.1 (`master` at c4839dd)

### Expected behavior

Accessing one of the voxelwise attributes which rely on `FirstLevelModel._get_voxelwise_model_attribute()`, such as `FirstLevelModel.residuals`, `FirstLevelModel.predicted`, or `FirstLevelModel.r_square`, should return a list of Nifti1Image objects with the same length as `FirstLevelModel.design_matrices_`.

### Actual behavior

The attributes are lists with only one item.

### The associated code

https://github.com/nilearn/nilearn/blob/c4839ddfe68ddf15775def1fc0ce9ea23544a527/nilearn/glm/first_level/first_level.py#L668-L686

### The solution

Unindenting line 686 should fix it, I think. There should also be at least one test to make sure that the length of the attribute list is the same as the length of `model.design_matrices`.

RaphaelMeudec:
Closes #2772 

Fixed indentations in `_get_voxelwise_model_attribute` to return as many attributes as design matrices. Added test with either 1 or 2 design matrices to make sure it's working fine.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -11,6 +11,9 @@ Fixes
   in :func:`nilearn.signal.clean`, so that these operations are applied
   in the same order as for the signals, i.e., first detrending and
   then temporal filtering (https://github.com/nilearn/nilearn/issues/2730).
+- Fix number of attributes returned by the
+  :func:`nilearn.glm.first_level.FirstLevelModel._get_voxelwise_model_attribute` method in the first level model.
+  It used to return only the first attribute, and now returns as many attributes as design matrices.
 
 
 Enhancements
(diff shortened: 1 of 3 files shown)


thomasbazeille (approved):
Thanks a lot Raphael !

CI didn't finished yet but it seems well done. 

Pleasse add an entry for this bugfix in doc/whats_new.rst

NicolasGensollen (approved):
LGTM! Thanks @RaphaelMeudec !

bthirion (approved):
LGTM, thx.

PR #2798, opened 2021-05-05
[FIX] Eliminate axis cutoff in Example 9.2.15.9

### LINKED ISSUE #2778: Axes Cutoff in Example 9.2.15.9 (plotting.plot_img_on_surf)
[`Example 9.2.15.9`](https://nilearn.github.io/auto_examples/01_plotting/plot_3d_map_to_surface_projection.html#plot-multiple-views-of-the-3d-volume-on-a-surface) features a quick plot showing multiple views of a volumetric stat map on an average surface. 

sphx_glr_plot_3d_map_to_surface_projection_006

However, the brain is cutoff on both axes (picture shown) both in the online example and when I use it on nilearn 0.7.1 within a jupyter notebook using the following code

```python
fig, ax = plotting.plot_img_on_surf(new_image,
                          views=['lateral','medial'],
                          inflate=True,
                          colorbar=True
                         )
```
Perhaps this is easily fixed post-hoc by adjusting matplotlib parameters, but it is not obvious to me.

tvanasse:
Closes #2778.

This fix simply changes the zoom of the axes (ax.dist) so that surfaces are no longer being cutoff. 

Before:

116889329-b929e080-ac2c-11eb-89a2-085636087f6f

After: 
download

(2x2)
download (2)


tsalo:
Thank you for this PR! Would you mind expanding the PR title to make it more informative, preferably with a PR prefix (e.g., `[FIX] Eliminate axis cutoff in Example 9.2.15.9`)? Also, you'll need to remove the space in the issue reference (#2778 instead of # 2778) in order to link it and automatically close that issue when this PR is merged.

tvanasse:
Thanks, I just made those edits. My first PR :)

bthirion (commented):
Thx ! Is it possible to opitmize a little bit more, as the colorbar seems super long and there is at the end a lot of empty space around the brain figure ?
(Simply try tweaking the parameters you can control here).

tvanasse:
> Thx ! Is it possible to opitmize a little bit more, as the colorbar seems super long and there is at the end a lot of empty space around the brain figure ?
> (Simply try tweaking the parameters you can control here).

Thanks for response. I've already tried tweaking some things to no avail but I will put some more energy in to make that happen.

nilearn/plotting/surf_plotting.py
```diff
@@ -859,7 +859,7 @@ def plot_img_on_surf(stat_map, surf_mesh='fsaverage5', mask_img=None,
         cbar_ax = fig.add_subplot(32, 1, 32)
         fig.colorbar(sm, cax=cbar_ax, orientation='horizontal')
 
-    fig.subplots_adjust(wspace=-0.02, hspace=0.0)
+    fig.subplots_adjust(wspace=-0.02, hspace=0)
```
NicolasGensollen, line 862: Is there a reason you changed this to an integer?

NicolasGensollen (changes requested):
Thanks @tvanasse ! 
In addition to Bertrand's comment, could you add a whatsnew entry?

nilearn/plotting/surf_plotting.py
tvanasse: That was a mistake, I removed this change in my new commit.

tvanasse:
@bthirion I have a new commit so there is now less white space & the color bar is shorter. The result is much better for  & 2x2 plots especially.

2x2 (before):
image

2x2 (after):

download (5)

1x1 (before):
image

1x1 (after)
download (6)

1x2 (after)
download (7)

bthirion:
This looks awesome, thx !

nilearn/plotting/surf_plotting.py
```diff
@@ -849,17 +849,25 @@ def plot_img_on_surf(stat_map, surf_mesh='fsaverage5', mask_img=None,
...
+            cbar_ax = fig.add_axes([0.265,0.17,0.5,0.04]) 
         fig.colorbar(sm, cax=cbar_ax, orientation='horizontal')
-
-    fig.subplots_adjust(wspace=-0.02, hspace=0.0)
+   
+    # if figure has two rows and two axes, adjust subplot for less excess space
+    if (axes.shape[0] == 2) & (axes.shape[1] ==  2):
+        fig.subplots_adjust(wspace=-0.4, hspace=-0.175)
+    else:
+        fig.subplots_adjust(wspace=-0.02, hspace=-0.3)
```
bthirion, line 870: Hm, I had never dared use negative hspace/wspace, but it looks good

nilearn/plotting/surf_plotting.py
```diff
@@ -849,17 +849,25 @@ def plot_img_on_surf(stat_map, surf_mesh='fsaverage5', mask_img=None,
...
                                   vmax, threshold, kwargs,
                                   cmap=get_cmap(cmap))
-
-        cbar_ax = fig.add_subplot(32, 1, 32)
+       
+        # if figure has two rows, change colorbar position for less excess space
+        if (axes.shape[0] == 2):
+            cbar_ax = fig.add_axes([0.265,0.10,0.5,0.04]) 
+        else:
+            cbar_ax = fig.add_axes([0.265,0.17,0.5,0.04])
```
NicolasGensollen, line 863: Are there only these two possibilities? 
I might be missing something but I don't see why you couldn't have more than 2 rows :thinking:

tvanasse: I guess I was only focusing on the two rows, but you are right. My new commit should handle three rows and beyond. 

download (8)

bthirion (commented):
LGTM.

nilearn/plotting/surf_plotting.py
jeromedockes: maybe using a gridspec that takes into account the colorbar instead of not taking it into account in the layout and adding it on top afterwards could help avoid these special cases and fine-tuning? 
maybe not, these things are tricky especially with 3d axes.

jeromedockes: I have the impression that unless it is passes a figure as well as the axes, `plot_surf` messes up the figure size:

https://github.com/nilearn/nilearn/blob/f76dabd6baf91db487cec606a7d0ea46effb50a2/nilearn/plotting/surf_plotting.py#L220

I think the figure size will always be the same whatever we set here 

https://github.com/nilearn/nilearn/blob/f76dabd6baf91db487cec606a7d0ea46effb50a2/nilearn/plotting/surf_plotting.py#L834

which probably doesn't help

jeromedockes: (try replacing figsize in the line above with whatever pair of values it doesn't affect the figure)

NicolasGensollen: I realized that we don't currently check views and hemispheres at all. 
So, you can do crazy things like:

```python
from nilearn.datasets import fetch_neurovault_motor_task
from nilearn.plotting import plot_img_on_surf

motor_images = fetch_neurovault_motor_task()
plot_img_on_surf(stat_img,
                 views=['lateral','lateral','lateral'],
                 hemispheres=['left','left','left'],
                 colorbar=True)
```
which look terrible:

plot_img

Should we constrain these ? (might make the plot setup easier...)

jeromedockes: I think it's ok to allow it -- if someone asks for the same view twice they get it twice which seems reasonable, unsurprising behavior. of course it would be best if they didn't get cut off but (i) I think the problem you see here is related to the bug I pointed out above rather to the fact that the same view is used several times (ii) if the plot doesn't look as good (but still doesn't result in an error or really incorrect results) in a corner case that no one will use it's not a big problem

jeromedockes: for example if you were to apply this patch (bg color, hspace and wspace is just to help see axis boundaries):

```diff
diff --git a/nilearn/plotting/surf_plotting.py b/nilearn/plotting/surf_plotting.py
index 8baa0b9d8..6b8d7822a 100644
--- a/nilearn/plotting/surf_plotting.py
+++ b/nilearn/plotting/surf_plotting.py
@@ -2,9 +2,12 @@
 Functions for surface visualization.
 Only matplotlib is required.
 """
+import itertools
+
 import matplotlib.pyplot as plt
 import numpy as np
 
+from matplotlib import gridspec
 from matplotlib.colorbar import make_axes
 from matplotlib.cm import ScalarMappable, get_cmap
 from matplotlib.colors import Normalize, LinearSegmentedColormap
@@ -227,7 +230,7 @@ def plot_surf(surf_mesh, surf_map=None, bg_map=None,
     else:
         if figure is None:
             figure = axes.get_figure()
-            figure.set_size_inches(*figsize)
+            # figure.set_size_inches(*figsize)
         axes.set_xlim(*limits)
         axes.set_ylim(*limits)
     axes.view_init(elev=elev, azim=azim)
@@ -769,7 +772,7 @@ def plot_img_on_surf(stat_map, surf_mesh='fsaverage5', mask_img=None,
                      views=['lateral', 'medial'],
                      output_file=None, title=None, colorbar=True,
                      vmax=None, threshold=None,
-                     cmap='cold_hot', aspect_ratio=1.4, **kwargs):
+                     cmap='cold_hot', **kwargs):
     """Convenience function to plot multiple views of plot_surf_stat_map
     in a single figure. It projects stat_map into meshes and plots views of
     left and right hemispheres. The *views* argument defines the views
@@ -874,45 +877,41 @@ def plot_img_on_surf(stat_map, surf_mesh='fsaverage5', mask_img=None,
                              mask_img=mask_img)
     }
 
-    figsize = plt.figaspect(len(modes) / (aspect_ratio * len(hemispheres)))
-    fig, axes = plt.subplots(nrows=len(modes),
-                             ncols=len(hemis),
-                             figsize=figsize,
-                             subplot_kw={'projection': '3d'})
-
-    axes = np.atleast_2d(axes)
-
-    if len(hemis) == 1:
-        axes = axes.T
-
-    for index_mode, mode in enumerate(modes):
-        for index_hemi, hemi in enumerate(hemis):
-            bg_map = surf_mesh['sulc_%s' % hemi]
-            plot_surf_stat_map(surf[hemi], texture[hemi],
-                               view=mode, hemi=hemi,
-                               bg_map=bg_map,
-                               axes=axes[index_mode, index_hemi],
-                               colorbar=False,  # Colorbar created externally.
-                               vmax=vmax,
-                               threshold=threshold,
-                               cmap=cmap,
-                               **kwargs)
-
-    for ax in axes.flatten():
+    cbar_h = .1
+    w, h = plt.figaspect((len(modes) + cbar_h) / len(hemispheres))
+    fig = plt.figure(figsize=(w, h))
+    grid = gridspec.GridSpec(
+        len(modes) + 1, len(hemis),
+        height_ratios=[1.] * len(modes) + [cbar_h], hspace=0.05, wspace=0.05)
+    axes = []
+    for i, (mode, hemi) in enumerate(itertools.product(modes, hemis)):
+        bg_map = surf_mesh['sulc_%s' % hemi]
+        ax = fig.add_subplot(grid[i], projection="3d")
+        axes.append(ax)
+        plot_surf_stat_map(surf[hemi], texture[hemi],
+                           view=mode, hemi=hemi,
+                           bg_map=bg_map,
+                           axes=ax,
+                           colorbar=False,  # Colorbar created externally.
+                           vmax=vmax,
+                           threshold=threshold,
+                           cmap=cmap,
+                           **kwargs)
+        ax.set_facecolor("#e0e0e0")
         # We increase this value to better position the camera of the
         # 3D projection plot. The default value makes meshes look too small.
-        ax.dist = 6
+        ax.dist = 7
 
     if colorbar:
         sm = _colorbar_from_array(image.get_data(stat_map),
                                   vmax, threshold, kwargs,
                                   cmap=get_cmap(cmap))
 
-        cbar_ax = fig.add_subplot(32, 1, 32)
+        cbar_grid = gridspec.GridSpecFromSubplotSpec(1, 3, grid[-1, :])
+        cbar_ax = fig.add_subplot(cbar_grid[1])
+        axes.append(cbar_ax)
         fig.colorbar(sm, cax=cbar_ax, orientation='horizontal')
 
-    fig.subplots_adjust(wspace=-0.02, hspace=0.0)
-
     if title is not None:
         fig.suptitle(title)
 
```

the output of the script you posted above would be

fignico

jeromedockes: and a more realistic one:
fig

bthirion:
Thx ! Can you trigger a doc build with images ?

nilearn/plotting/surf_plotting.py
tvanasse: Wow, that looks excellent @jeromedockes ! Should I make those changes, and then commit those to this pull request? Or what's the best way forward?

jeromedockes: maybe the easiest way to apply this patch is for me to open a PR on your `plot_fix` branch. you can then review and merge it into this branch (from the github interface), and keep pushing to the same branch to update this PR with any adjustments you deem necessary, and when we are finally satisfied we merge your branch into upstream.
does that sound reasonable

tvanasse: Sounds good!

jeromedockes: ok great let's do that
I opened a PR, it looks like a large diff but this is just because your branch is a bit behind master. if you merge master into your branch first it will be easier to check!

jeromedockes:
note: the build failure is due to #2837 and not related to your PR

tvanasse:
I merged @jeromedockes 's changes into this PR and everything seems to be working fine. 

The only change I made to that was making the constrained_layout parameter false in figure creation (with the idea to eliminate any margin space). 

fig = plt.figure(figsize=(w, h), constrained_layout=False)

image

Does this all look good?

jeromedockes:
> Does this all look good?

sorry one small detail is I forgot about the title, I just opened another small pr on your branch to address that

jeromedockes:
No strong opinion about the `constrained_layout`, I'm not sure it can work properly with 3d axes anyway so if you prefer disabling it I'm fine with it!

if you want to apply a bit of negative space between subplots as you had done before that is possible too; for now they are set to exactly 0 . when doing so we just have to be careful that different views occupy different amounts of space in x and y directions.
otherwise the brains seem reasonably spaced to me so 0.0 space seems fine as well

jeromedockes:
another thing is that depending on the number of views the figure size changes but the font size is always the same so the relative size of text may vary. (but that is the case for all figures, though)
we can let the user deal with it by using for example 
```python
with plt.rc_context({'font.size': my_font_size}):
    plotting.plot_img_on_surf(img)
```

or to make it slightly easier by adding an optional `font_size` parameter or something like that

bthirion:
Would be good to demo such features so that people get an idea on how to do it.

NicolasGensollen (commented):
This looks good to me, thanks @tvanasse @jeromedockes !
Could you add a whatsnew entry and trigger a full build to make sure examples using `plot_img_on_surf` look good? (just put "[circle full]" in the commit message).

nilearn/plotting/surf_plotting.py
```diff
@@ -876,47 +878,46 @@ def plot_img_on_surf(stat_map, surf_mesh='fsaverage5', mask_img=None,
...
         fig.colorbar(sm, cax=cbar_ax, orientation='horizontal')
 
-    fig.subplots_adjust(wspace=-0.02, hspace=0.0)
-
     if title is not None:
-        fig.suptitle(title)
+        fig.suptitle(title, y=1. - title_h / sum(height_ratios), va="bottom")
 
     if output_file is not None:
         fig.savefig(output_file)
```
jeromedockes, line 923: while we are here

```suggestion
        fig.savefig(output_file, bbox_inches="tight")
```

jeromedockes:
> Please let me know if I didn't do this correctly

you did it correctly! all the examples are now running on circle ci ( [build](https://app.circleci.com/pipelines/github/nilearn/nilearn/2085/workflows/47e3db27-18bd-49e2-901f-9b8d4e19177b/jobs/7263) ), when it finishes in a couple of hours we can check the gallery of examples it will produce and if everything looks good we can merge this PR

bthirion (approved):

jeromedockes (approved):
everything LGTM. thanks a lot @tvanasse !

NicolasGensollen (commented):
LGTM too, thanks @tvanasse !
Could you just add a whatsnew entry (in the fix section in `doc/whats_new.rst`) before we merge this?

tvanasse:
> Could you just add a whatsnew entry (in the fix section in `doc/whats_new.rst`) before we merge this?

Done :)

jeromedockes:
ok everything looks good, merging! thanks a lot for this @tvanasse

Issue #2642, opened 2020-12-23
accept `pathlib.Path` where strings representing filesystem paths are expected


jeromedockes:
for example in `image.load_img`, `_utils.niimg.load_niimg`, `surface.load_surf_mesh` and `surface.load_surf_data`

bakyazi:
hello, i am new in open source. can i work on it?

jeromedockes:
>  hello, i am new in open source. can i work on it?

that would be great, thanks for offering!

emdupre:
This is something that is handled natively with nibabel 3.0 ! Could be a chance to check for those potential updates.

udion:
Hello, 
I am working on implementing this feature.

RaphaelMeudec:
I started looking into it:

- all functions cited by Jerome raise either a `TypeError`, either a `ValueError`
- this is the case because nilearn checks for `isinstance(name, str)`
- as @emdupre said, nibabel underlying functions seem to be working with Path (at least, I checked `nibabel.load` and `nibabel.freesurfer.io`)

I guess we should loosen the `isinstance` check to accept pathlib.Path arguments

RaphaelMeudec:
@udion let me know if you need help on this!

udion:
Thank you. I was also thinking to check for the `Pathlib` object.

effigies:
If it's helpful, you can import or copy [`nibabel.filename_parser._stringify_path()`](https://github.com/nipy/nibabel/blob/5c4f39cef1f04a9afd241242c7f0148d7e5935c9/nibabel/filename_parser.py#L19-L46).

udion:
@RaphaelMeudec @effigies 
I made the following changes, to begin within the `nilearn/_utils/niimg_conversions.py` that has the `check_niimg` function that is invoked when `nilearn.image.load_img` is called. The idea was to extract the string from the `pathlib object` and then process it in the same manner as a regular string (currently handled by the code).

```python
if isinstance(niimg, pathlib.Path):
        # get the string from pathlib object
        # and then process like a normal string
        niimg = niimg.absolute().as_posix()
        if wildcards and ni.EXPAND_PATH_WILDCARDS:
            # Ascending sorting + expand user path
            filenames = sorted(glob.glob(os.path.expanduser(niimg)))
            # processing filenames matching globbing expression
            if len(filenames) >= 1 and glob.has_magic(niimg):
                niimg = filenames  # iterable case
            # niimg is an existing filename
[...]

PR #2803, opened 2021-05-06
[WIP] incorporating `pathlib` in `image.load_img` by extracting string from `pathlib` oject

udion:
WIP for  #2642.

- `pathlib` object supported as an argument in `image.load_image`

### DIFF
--- a/nilearn/_utils/niimg_conversions.py
+++ b/nilearn/_utils/niimg_conversions.py
@@ -10,6 +10,7 @@ import glob
 import nilearn as ni
 import numpy as np
 import itertools
+import pathlib
 
 from joblib import Memory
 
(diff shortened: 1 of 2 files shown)


nilearn/_utils/niimg_conversions.py
```diff
@@ -271,6 +272,34 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
                 raise ValueError("File not found: '%s'" % niimg)
         elif not os.path.exists(niimg):
             raise ValueError("File not found: '%s'" % niimg)
+    
+    if isinstance(niimg, pathlib.Path):
+        # get the string from pathlib object
+        # and then process like a normal string
+        niimg = niimg.absolute().as_posix()
```
NicolasGensollen, line 279: Unless I missed something, the code inside this if statement is exactly the same as the code just above. 
One much simpler way to rewrite this would be to place this if at the beginning of the function (which will cast niimg to a string if it is a Path), and let the previous block of code handle the rest.
Does this make sense?

udion: Yes, will update.

jeromedockes: why take the absolute path and why `as_posix` ?

effigies:
I would like to again push a bit for something like [`nibabel.filename_parser._stringify_path()`](https://github.com/nipy/nibabel/blob/5c4f39cef1f04a9afd241242c7f0148d7e5935c9/nibabel/filename_parser.py#L19-L46).

This is nice because it permits any object that exposes the `os.PathLike` interface to be interpreted, and not exclusively `pathlib.Path`s. One example is the PyBIDS `BIDSFile` object:

```Python
>>> import bids
>>> layout = bids.BIDSLayout("/data/bids/openneuro/ds000114/")
>>> bold_files = layout.get(suffix="bold", extension=".nii.gz")
>>> bold_files[0]
<BIDSImageFile filename='/data/bids/openneuro/ds000114/sub-01/ses-retest/func/sub-01_ses-retest_task-covertverbgeneration_bold.nii.gz'>
>>> nb.load(bold_files[0])
<nibabel.nifti1.Nifti1Image at 0x7fd1aa0ff460>
```

jeromedockes:
> I would like to again push a bit for something like [`nibabel.filename_parser._stringify_path()`](https://github.com/nipy/nibabel/blob/5c4f39cef1f04a9afd241242c7f0148d7e5935c9/nibabel/filename_parser.py#L19-L46).
>
> This is nice because it permits any object that exposes the `os.PathLike` interface to be interpreted, and not exclusively `pathlib.Path`s. One example is the PyBIDS `BIDSFile` object:

I agree; you can copy that function into `nilearn._utils.helpers`, for example.

jeromedockes:
Note now that we dropped support for python 3.5 that function can be slightly simplified by using the [`os.Pathlike` abc](https://docs.python.org/3/library/os.html#os.PathLike)

jeromedockes:
Hi @udion  are you still working on this? let us know if you need some help

PR #2804, opened 2021-05-06
[FIX] Preventing plot title overlapping content

### LINKED ISSUE #2054: Plot title overlays the image
Currently adding a title to a plot results in the title being drawn over the image, especially prominent if the title is long:

image

Ideally, there should be a space between the title and the image below that.

OliverWarrington:
Closes #2054.

- Changed the default y position for the title function in BaseSlicer class. Set to 1.08 rather that 0.99, enough to prevent even really long titles overlapping other elements.
Screenshot 2021-05-06 at 10 00 46


NicolasGensollen (commented):
LGTM! Thanks @OliverWarrington ! :+1: 
Could you just add a whatsnew entry? (in file `doc/whats_new.rst`)

OliverWarrington:
I think it's done. @NicolasGensollen please could you make sure this is fine? I get a little confused in the updating pull requests world!

bthirion:
LGTM, but I think we should trigger a full doc rebuild to check that example figs render well...

doc/whats_new.rst
```diff
@@ -7,6 +7,7 @@ NEW
 Fixes
 -----
 
+- Fix plot title overlapping image content.
```
NicolasGensollen, line 10: Could you add the issue number in the description here? It might be hard to know what was fixed otherwise.

NicolasGensollen (changes requested):
LGTM, thanks @OliverWarrington !
I just have a very small request (see below).
When you commit this, could you put "[circle full]" somewhere in your commit message? This will trigger a full build on circleCI as mentioned by @bthirion.
If the doc builds correctly, we should be good to merge! :tada:

bthirion:
The problem now is that the title gets cut out of the img on plot_stat_map examples (see CircleCI).

OliverWarrington:
> The problem now is that the title gets cut out of the img on plot_stat_map examples (see CircleCI).

Ah yep, I see what you mean. I guess the figure size works differently in Jupyter? I'll have another go!

jeromedockes:
I think part of the difficulty comes from the fact that the nilearn plotting functions make extensive use of `add_axes` to place axes in arbitrary, hard-coded positions in the figure as [here for example](https://github.com/nilearn/nilearn/blob/ae30ef6a4b622eff8a14750535449d4b38feee64/nilearn/plotting/displays.py#L1750) . using this approach rather than positions on grids with GridSpec and add_subplot prevents tight_layout from working and makes the layout more tricky to get right, especially when inputs can vary

jeromedockes:
addressing that would be a bigger change than was intended in this PR though, so maybe here if there really isn't enough space you can slightly increase the figure size (the `_default_figsize` of the `*Slicer` objects that need it)

OliverWarrington:
Thanks @jeromedockes, I'll try increasing the default size. Is there a way to get the same outputs as running CircleCI when I'm checking my changes locally? The outputs of running the plotting examples either in jupyter or just as a script look different to me from the CircleCI outputs.

NicolasGensollen:
@OliverWarrington if you want to check the rendering of the docs locally, you can build it by running `make html-strict` from the `doc` subfolder. Note that this will build the whole documentation which usually takes around 1h30...
If you have already built it once, you can only rebuild the examples that you modified (see [here](http://nilearn.github.io/development.html#contributing-to-the-documentation) for more details).

ymzayek:
@jeromedockes I was wondering regarding your proposed solution of changing the hardcoded `_default_figsize` if this could be done in a way that if a title is given a value can be added to the figsize y axis? This would be kind of like what is done with the colorbar in https://github.com/nilearn/nilearn/blob/1b7994104d08a117b859e784f3de567be87f0e59/nilearn/plotting/displays/_slicers.py#L136-L138

Do you think it could work?

jeromedockes:
> Do you think it could work?

I think so! although it probably isn't a big deal to increase a bit the figure size in all cases

PR #2805, opened 2021-05-06
WIP: Pass through vol_to_surf_kwargs

### LINKED ISSUE #2180: expose `vol_to_surf` parameters in `view_img_on_surf`?
I wonder if we should allow users to specify parameters for the volume to surface projection when using `view_img_on_surf`. This would help explore the effect of these parameters. moreover, for discrete images such as atlases the default parameters of `vol_to_surf` are not appropriate.

Being able to disable interpolation would allow better plots for discrete images.

   ```python 
   cmap = "gist_ncar"

   destrieux = datasets.fetch_atlas_destrieux_2009()
   plotting.view_img_on_surf(destrieux.maps, surf_mesh="fsaverage", cmap=cmap,
                             symmetric_cmap=False).open_in_browser()
   plotting.view_img_on_surf(
       destrieux.maps, surf_mesh="fsaverage",
       vol_to_surf_params={"n_samples": 1, "radius": 0.,
                           "interpolation": "nearest"},
       cmap=cmap, symmetric_cmap=False).open_in_browser()
   ```

newplot(1)

newplot

or simply help getting a notion of voxel and mesh resolutions

newplot(1)

Craskermasker:
Closes #2180  .

- Added an optional parameter `vol_to_surf_kwargs` to `view_img_on_surf` and pass this through to `full_brain_info`.

### DIFF
--- a/nilearn/plotting/html_surface.py
+++ b/nilearn/plotting/html_surface.py
@@ -192,6 +192,11 @@ def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
     title_fontsize : int, optional
         Fontsize of the title. Default=25.
 
+    vol_to_surf_kwargs : dict, optional
+        Dictionary of keyword arguments that are passed on to
+        :func:`nilearn.surface.vol_to_surf` when extracting a surface from
+        the input image. See the function documentation for details.
+
     Returns
     -------
     SurfaceView : plot of the stat map.
--- a/nilearn/plotting/html_surface.py
+++ b/nilearn/plotting/html_surface.py
@@ -129,7 +129,7 @@ def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
                      threshold=None, cmap=cm.cold_hot,
                      black_bg=False, vmax=None, vmin=None, symmetric_cmap=True,
                      colorbar=True, colorbar_height=.5, colorbar_fontsize=25,
-                     title=None, title_fontsize=25):
+                     title=None, title_fontsize=25, vol_to_surf_kwargs={}):
     """Insert a surface plot of a statistical map into an HTML page.
 
     Parameters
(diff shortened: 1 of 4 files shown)


Craskermasker:
IMO `vol_to_surf_kwargs` is better than `vol_to_surf_params` as this matches the parameter on `full_brain_info`. 

Next steps:
- add test
- write demo

NicolasGensollen:
FYI: Here's a link to the rendering of the modified example on CircleCI: [here](https://7092-1235740-gh.circle-artifacts.com/0/dev/auto_examples/01_plotting/plot_3d_map_to_surface_projection.html#sphx-glr-auto-examples-01-plotting-plot-3d-map-to-surface-projection-py)

Craskermasker:
I don't think I understand what you mean by this:
> add a short explanation in the plotting functions' docstrings

Do you mean the `full_brain_info` function?

In my last push I've added an entry to the `view_img_on_surf` docstring and embedded a link to the documentation for `vol_to_surf`. I think this would avoid unnecessary duplication of documentation.

bthirion (approved):
LGTM, thx !

NicolasGensollen (approved):
This looks good to me! 
Thanks for adding the argument description in the docstring!

NicolasGensollen:
Oh actually, I forgot to write that adding a whatsnew entry (under enhancements) would be great! Thx!

nilearn/plotting/html_surface.py
```diff
@@ -192,6 +192,11 @@ def view_img_on_surf(stat_map_img, surf_mesh='fsaverage5',
     title_fontsize : int, optional
         Fontsize of the title. Default=25.
 
+    vol_to_surf_kwargs : dict, optional
+        Dictionary of keyword arguments that are passed on to
+        :func:`nilearn.surface.vol_to_surf` when extracting a surface from
+        the input image. See the function documentation for details.
```
jeromedockes, line 198: > In my last push I've added an entry to the view_img_on_surf docstring and embedded a link to the documentation for vol_to_surf. I think this would avoid unnecessary duplication of documentation.

thanks, I agree with you about avoiding too much duplication but still if someone gets weird results when plotting an atlas and is not used to the vol_to_surf function they might not think to look in there.
So maybe we could repeat here what is the appropriate parameter when plotting an atlas for example:

this parameter is especially useful when plotting an atlas; in that case to avoid averaging or interpolating labels from neighboring voxels pass vol_to_surf_kwargs = {"n_samples": 1, "interpolation": "nearest"}

Craskermasker: Thank you - I've added a bit more info in the docstring and linked to the relevant example too.

jeromedockes (approved):
thanks a lot!

PR #2810, opened 2021-05-06
Generate colors and colorscale for view_connectome

### LINKED ISSUE #2444: Enhancement for plotting.view_connectome: nodes size & color
H!
`plotting.view_connectome` offers 3D plotting of connections (edges with varying weights represented as adjacency matrix) together with spheres (nodes) with a fixed color and specified size (the same for all markers). We propose to provide an enhancement to the `plotting.view_connectome` by 
including varying markers colors/sizes (specified as arrays of nodes' colors and sizes). As `plotting.plot_connectome` already provides such functionality, the enhancement would make the connectome plotting tools in Nilearn more consistent.

Such functionality is also available in `plotting.view_markers` but that plotting option does not enable to visualize edges of the network. 

Do you think it is reasonable and feasible to work on such enhancement (+tutorial demonstration) during the Nilearn Dev Days sprint?

Many researchers in network neuroscience that use [BrainNet Viewer](https://www.nitrc.org/project/list_screenshots.php?group_id=504&screenshot_id=710) for plotting might appreciate such enhancement!

RaphaelMeudec:
Fixes #2444 

The issue is about:

- marker colors
- marker size

While looking into it, it seems that marker size is already supported. Hence, the PR is only about modifying the color (test and whatsnew entry on the way).

I followed the signature of `plot_connectome`, adding a new argument `node_color` which can be:

- a raw matplotlib color such as `cyan` or `red`
- a list of matplotlib colors
- the keyword `auto`

Here are some screenshots (generated with the script below)

connectome

connectome

connectome

```python
import numpy as np

adjacency_matrix = np.array([[1., -2., 0.3, 0.],
                             [-2.002, 1, 1.57, 0.],
                             [0.3, 1.57, 1., 0.],
                             [0., 0., 0., 1.]])
node_coords = np.arange(3 * 4).reshape(4, 3)

args = adjacency_matrix, node_coords
kwargs = dict(edge_threshold=0.38, node_size=10)

view = plotting.view_connectome(node_color="auto", title='auto color', *args, **kwargs)
view.open_in_browser()

view = plotting.view_connectome(node_color="cyan", title='cyan color', *args, **kwargs)
view.open_in_browser()

view = plotting.view_connectome(node_color=['cyan', 'blue', 'green', 'r'], title='array color', *args, **kwargs)
view.open_in_browser()
```


RaphaelMeudec:
@jeromedockes You might be the best person to review it when you have time, thanks!

nilearn/plotting/html_connectome.py
```diff
@@ -27,8 +27,33 @@ def _prepare_line(edges, nodes):
...
+        marker_colorscale = 'Viridis'
+    elif isinstance(marker_color, str):
+        color_as_hex = mpl_colors.to_hex(marker_color)
+        marker_colorscale = [[0, color_as_hex], [1, color_as_hex]]
+        marker_colors = [0] * len(path_nodes)
+    elif isinstance(marker_color, list):
+        if len(marker_color) != np.max(path_nodes) + 1:
+            raise ValueError("Number of colors provided should match the number of nodes")
+        # Colorscale values must be between 0 and 1
+        marker_colorscale = [
```
jeromedockes, line 45: I have the impression it would be more natural to pass a list of colors (one for each node) in this case than to hack a colorscale. WDYT?

nilearn/plotting/html_connectome.py
```diff
@@ -103,6 +133,9 @@ def view_connectome(adjacency_matrix, node_coords, edge_threshold=None,
     node_coords : ndarray, shape=(n_nodes, 3)
         The coordinates of the nodes in MNI space.
 
+    node_color : color or sequence of colors, optional
```
jeromedockes, line 136: ATM the marker size is still the same for all nodes, is that correct?

jeromedockes (commented):
thanks!

jeromedockes:
> @jeromedockes You might be the best person to review it when you have time, thanks!

sure, thanks! I asked a couple of questions but I can spend more time on
it this afternoon

RaphaelMeudec:
Yes, marker_size is fixed in the screenshot. Here is an example with variable size (`node_size=[10, 20, 30, 40]`): 

image

nilearn/plotting/html_connectome.py
RaphaelMeudec: Indeed. For the `auto`, how would you sample the colors? Maybe by using the `Set2` colormap as in `plot_connectome`?

nilearn/plotting/html_connectome.py
```diff
@@ -27,8 +27,33 @@ def _prepare_line(edges, nodes):
     return path_edges, path_nodes
 
 
+def _prepare_colors(marker_color, path_nodes):
```
bthirion, line 30: Can you add a 1-line docstring to the  ?

nilearn/plotting/html_connectome.py
```diff
@@ -27,8 +27,33 @@ def _prepare_line(edges, nodes):
...
+        ]
+        marker_colors = (path_nodes / np.max(path_nodes)).tolist()
+
+    return marker_colors, marker_colorscale
+
+
 def _get_connectome(adjacency_matrix, coords, threshold=None,
-                    marker_size=None, cmap=cm.cold_hot, symmetric_cmap=True):
+                    marker_size=None, marker_color='auto', cmap=cm.cold_hot,
+                    symmetric_cmap=True):
```
bthirion, line 56: A short docstring would be nice here too.

bthirion (commented):
That's awesome, thx !

nilearn/plotting/html_connectome.py
jeromedockes: yes let's do the same as plot_connectome

nilearn/plotting/html_connectome.py
```diff
@@ -27,8 +27,33 @@ def _prepare_line(edges, nodes):
...
+        marker_colors = np.linspace(0, 100, len(path_nodes)).astype("uint8")
+        marker_colors = marker_colors[path_nodes].tolist()
+        marker_colorscale = 'Viridis'
+    elif isinstance(marker_color, str):
+        color_as_hex = mpl_colors.to_hex(marker_color)
+        marker_colorscale = [[0, color_as_hex], [1, color_as_hex]]
+        marker_colors = [0] * len(path_nodes)
+    elif isinstance(marker_color, list):
+        if len(marker_color) != np.max(path_nodes) + 1:
+            raise ValueError("Number of colors provided should match the number of nodes")
```
jeromedockes, line 43: this will be raised if some nodes are disconnected from all the others.

I think that would be another motivation for separating the line and markers

jeromedockes: my suggestion is to treat the line and the markers separately. ie instead of
having only "con_x", "con_y", "con_z", have "con_x", "con_y", "con_z",
"marker_x", "marker_y", "marker_z" (or any better names you can think of -- we
could rename "con_x" -> "line_x" for example)

they won't have the same size because the line goes through all edges, whereas
the list of markers goes through each node exactly once.

note storing the connectome nodes twice is not a problem, what thakes space is
the brain mesh anyway, not the connectome.

then in connectome_plot_template.html, there wouldn't be 2 distinct functions
`addMarkers` and `addConnectome` any more -- there would be only one call to
`Plotly.plot`. but the array of traces would contain 2 traces.
instead of `Plotly.plot("connectome-plot", [{mode: "lines+markers+text", ...}])`
we would have
`Plotly.plot("connectome-plot", [{mode: "lines", ...}, {mode: "markers+text, ...}])`

it will solve the question of showing markers that don't have any connection,
avoid plotting several times the same marker on top of itself, and make it
simpler to pass the colors and sizes for the markers as these lists will now
have the same length as `node_coords`.

does that make sense?

RaphaelMeudec:
@jeromedockes I'm not done yet, but I made significant changes and want to make sure I'm in the right direction.

So far :

- split in pyplot markers and lines as you pointed out
- split the dictionary construction in html_connectome.py between:
  - markers information
  - lines information
- start factorizing elements between `view_connectome` and `view_markers`

What's left to do:

- tackle the colors/colorscale element you also pointed out
- make sure `view_markers` also benefit from the same node management as `view_connectome`
- update tests & docstrings

jeromedockes:
> @jeromedockes I'm not done yet, but I made significant changes and want to make sure I'm in the right direction.
>
> So far :
>
> - split in pyplot markers and lines as you pointed out
> - split the dictionary construction in html_connectome.py between:
>   - markers information
>   - lines information
> - start factorizing elements between `view_connectome` and `view_markers`
>
> What's left to do:
>
> - tackle the colors/colorscale element you also pointed out
> - make sure `view_markers` also benefit from the same node management as `view_connectome`
> - update tests & docstrings
>

Hi, thanks! yes what you describe here is exactly what I had in mind.

Sorry I haven't looked at the new commits yet (actually I didn't realize until
now that you had pushed so many changes), but if you think that's helpful I can
have a look now.

jeromedockes:
just to have a schematic idea of what the new plotting with 2 traces will look like for future reference in this discussion, here is a minimal example without the clutter of dealing with the actual connectome

```python
from nilearn.plotting import js_plotting_utils

TEMPLATE = """
<!DOCTYPE html>
<html lang="en">
<head>
    $INSERT_JS_LIBRARIES_HERE
    <script>
     function addPlot(){
         let x = [0, 0, 0, 20];
         let y = [0, 20, 0, 20];
         let z = [0, 0, 20, 20];

         Plotly.plot('connectome-plot', [{
             type: 'scatter3d',
             mode: 'markers',
             x: x,
             y: y,
             z: z,
             opacity: 1,
             marker: {
                 size: [10, 20, 30, 40],
                 color: ["red", "green", "blue", "orange"],
             }
         }, {type: 'scatter3d',
             mode: 'lines',
             x: x.slice(0, -1),
             y: y.slice(0, -1),
             z: z.slice(0, -1),
         line: {color: ["black", "magenta", "cyan"]}}
         ]);

     }
    </script>
    <script>
        $(document).ready(addPlot);
    </script>
</head>
<body>
    <div id="connectome-plot"></div>
</body>
</html>
"""

html = js_plotting_utils.add_js_lib(TEMPLATE)
js_plotting_utils.HTMLDocument(html).open_in_browser()
```

markers and lines are handled separately and don't have the same length

in the resulting plot we can also see that in the line plot the specified colors actually correspond to points and the edge color is a gradient joining the colors of both points, which is why we have to double each point so that both ends of an edge have the same color and thus we force plotly to paint the edge in the color we want

newplot

nilearn/plotting/html_connectome.py
```diff
@@ -27,49 +65,149 @@ def _prepare_line(edges, nodes):
...
+    ----------
+    marker_color : color or sequence of colors, optional
+        Color(s) of the nodes. Default='auto'.
+
+    number_of_nodes : int
+        Number of nodes in the view
+
+    Returns
+    -------
+    markers_colors: list
```
bthirion, line 82: Describe a little bit the argument (although this is not strictly necessary in a private function)

bthirion (commented):
Looks good. Some lines are not covered apparently.

nilearn/plotting/html_connectome.py
jeromedockes: IIUC the logic of `to_color_strings` is duplicated here to avoid building and
querying a matplotlib colormap twice when `marker_color == "auto"`, is that the
case? If so, I am not sure this optimization is worth the code duplication (the
call to `to_color_strings` takes < 1 ms even with hundreds of nodes and thus is
small compared to `view_markers` execution time which is in the order of dozens
of ms, and of actually displaying the plot which is much longer).

also, WDYT of replacing the gray color of set2 with a different one? it is a bit
hard to see as it is the same color as the transparent brain

nilearn/plotting/html_connectome.py
```diff
@@ -27,49 +65,137 @@ def _prepare_line(edges, nodes):
...
+
+    lines_metadata = {
+        **lines_metadata,
+        **_encode_lines_coordinates(line_coords)
+    }
+
+    return lines_metadata
+
+
+def _prepare_markers_metadata(coords, marker_size, marker_color):
```
jeromedockes, line 161: It seems this preparation of markers is duplicated:
- when plotting a full connectome it appears here (in `_prepare_markers_metadata`)
- when plotting markers only half of it is inlined in `view_markers` and the rest in `_get_markers`.

but the logic in both cases in the same and the only important difference is that "markers_only" needs to be set to `True` in the second case.

could we remove `_get_markers` and use `_prepare_markers_metadata` in both cases?

nilearn/plotting/html_connectome.py
```diff
@@ -1,22 +1,60 @@
...
+    coords = np.asarray(coords, dtype='<f4')
+    marker_x, marker_y, marker_z = coords.T
+    for coord, cname in [(marker_x, "x"), (marker_y, "y"), (marker_z, "z")]:
+        coordinates["{}{}".format(prefix, cname)] = encode(
+            np.asarray(coord, dtype='<f4'))
+
+    return coordinates
+
+
+def _encode_markers_coordinates(coords):
```
jeromedockes, line 48: I am not sure these two functions add a lot of readability.
I would suggest replacing them with the call using the explicit arg name for the second argument, eg 
`_encode_coordinates(coords, prefix="_marker_")`
instead of `_encode_markers_coordinates(coords)` .

I am now thinking this module would maybe be slightly simpler if line and marker attributes were stored in separate (sub-)dictionaries inside the metadata, rather than using prefixes to implement "namespaces", but I don't think it's worth making this change now

jeromedockes (commented):
looks good! here is a first batch of small comments.

nitpick: could you make it so that (physical) lines in the code are shorter than 80 columns?

RaphaelMeudec:
@jeromedockes I fixed all the elements.

- I switched to viridis colormap (instead of Set2) 
- merged _get_markers with _prepare_markers_metadata
- use `to_color_strings` in _prepare_colors: I didn't expect to_color_strings to take numpy array as input, that's why I re-built the logic. But it's working fine!

nilearn/plotting/html_connectome.py
```diff
@@ -17,6 +18,34 @@ class ConnectomeView(HTMLDocument):
...
+    coords = np.asarray(coords, dtype='<f4')
+    marker_x, marker_y, marker_z = coords.T
+    for coord, cname in [(marker_x, "x"), (marker_y, "y"), (marker_z, "z")]:
+        coordinates["{}{}".format(prefix, cname)] = encode(
+            np.asarray(coord, dtype='<f4'))
+
+    return coordinates
+
+
 def _prepare_line(edges, nodes):
```
jeromedockes, line 49:
```suggestion
def _prepare_line(edges, nodes):
    """prepare a plotly scatter3d line plot so that a set of disconnected edges
    can be drawn as a single line.

    `edges` are values associated with each edge (that get mapped to colors
    through a colorscale). `nodes` are pairs of (source, target) node indices
    for each edge.

    the color of a line segment in plotly is a mixture of the colors associated
    with the points it connects. Moreover, segments that begin or end at a
    point whose value is `null` are not drawn.

    given edges = [eab, ecd, eef] and nodes = [(a, b), (c, d), (e, f)], this
    function returns:

        path_edges: eab eab   0 ecd ecd   0 eef eef   0
        path_nodes:   a   b   0   c   d   0   e   f   0

    moreover the javascript code replaces every third element (the '0' in the
    lists above) with `null`, so only the a-b, c-d, and e-f segments will get
    plotted, and their colors are correct because both their start and end
    points are associated with the same value.

    """
```

jeromedockes: I think the role of this function made it harder to get started on this PR, was
that the case? If so, we can add a comment like the one above

jeromedockes (commented):
This is almost ready to merge!
one last thing: could you push a commit with "[circle full]" in the commit message, so that we see the doc built by circle ci when running all the examples?

thanks

jeromedockes:
> I switched to viridis colormap (instead of Set2)

great, thanks. it may be a bit weird to use a sequential colormap for this, but anyway I don't really see the point of using different (but arbitrary) colors by default, so the important thing is that all nodes are very easy to see (as is the case now)

jeromedockes:
if you want to merge master circleci should work now

RaphaelMeudec:
@jeromedockes build looks ok, tell me if there is anything else to add!

jeromedockes:
great, thanks! could you push again with the [circle full] tag? the first time we tried it the build was failing for an unrelated reason so we did not get to see the built documentation.

apart from that everything LGTM, thanks again for all the work!


I also noticed these parameters (nor their equivalents in `plot_connectome`) are not illustrated in the example gallery, if you'd like to modify one of the existing examples such as [this one](https://nilearn.github.io/auto_examples/01_plotting/plot_surf_atlas.html#sphx-glr-auto-examples-01-plotting-plot-surf-atlas-py) to showcase the use of the parameters you just added that would be nice -- but I would recommend doing it in a separate PR.

jeromedockes (approved):

bthirion (approved):
Thx !

NicolasGensollen:
Hi @RaphaelMeudec 
It looks like the doc failed to build this example: `03_connectivity/plot_probabilistic_atlas_extraction.py`
See [this build](https://app.circleci.com/pipelines/github/nilearn/nilearn/2079/workflows/40307544-6682-4a85-a03b-bba2adf7668b/jobs/7257).

jeromedockes:
you just need to make sure the coordinates are stored in a numpy array: `node_coords = np.asarray(node_coords)` in `view_connectome` and `view_markers`

RaphaelMeudec:
@NicolasGensollen @jeromedockes thanks for pointing this out. I fixed it, hopefully it'll be the last commit in this pr!

jeromedockes:
there is now a meaningless legend on the connectome plots; see the top right corner of this example:

newplot


can you remove it by applying this change? thanks

```diff
diff --git a/nilearn/plotting/data/js/surface-plot-utils.js b/nilearn/plotting/data/js/surface-plot-utils.js
index 12556bbc6..11dc7668b 100644
--- a/nilearn/plotting/data/js/surface-plot-utils.js
+++ b/nilearn/plotting/data/js/surface-plot-utils.js
@@ -110,6 +110,7 @@ function getLayout(plotDivId, viewSelectId, blackBg) {
     let width = height * 3 / 2;
 
     let layout = {
+        showlegend: false,
         height: height, width: width,
         margin: {l:0, r:0, b:0, t:0, pad:0},
         hovermode: false,
```

nilearn/plotting/data/js/surface-plot-utils.js
```diff
@@ -110,6 +110,7 @@ function getLayout(plotDivId, viewSelectId, blackBg) {
     let width = height * 3 / 2;
 
     let layout = {
+        showLegend: false,
```
jeromedockes, line 113:
```suggestion
        showlegend: false,
```

RaphaelMeudec: Sorry for not testing it, I've pushed the correction, legend is now off. Should I do a full circle build?

jeromedockes: thanks! no I think it's ok, I also checked it locally and that was the last missing detail

jeromedockes:
Thanks so much for this @RaphaelMeudec ! merging :)

PR #2815, opened 2021-05-06
[MRG] Add support for downloading all resolutions of fsaverage from 3 to 7

### LINKED ISSUE #2807: Add more available resolutions for fsaverage
For now, only fsaverage 5 (10k points per hemisphere) and 7 (150k points) are available.
I think it would make sense to add fsaverage 6 (40k points), and while we are at it do all versions from 1 to 4 as well.
You can visit [this page](https://brainder.org/2016/05/31/downsampling-decimating-a-brain-surface/) from brainder.org for more info about available resolutions :slightly_smiling_face:

Also, I feel like this issue is related to #2313 although not equivalent.

### Benefits to the change

- **fsaverage6**: A lot of people study quadratic problems on the brain surface, and going from 10k points to 150k can be a big step. I think it would be helpful to allow for intermediate resolutions.
- **fsaverage1-4?**: I think it sometimes makes sense to work with a low resolution on the brain surface (for instance when doing a big grid search or just when you want things to go really really fast); I would personally use `fsaverage4` from time to time if it were available!

```python
fsaverage = datasets.fetch_surf_fsaverage(mesh='fsaverage4')
```

### Where to start

I think I would start from https://github.com/nilearn/nilearn/blob/a36c0442fc8eebe401b310dd9799c0f071080e90/nilearn/datasets/struct.py#L452 and try do something with a similar spirit as what has been implemented for fsaverage 5 and 7.

alexisthual:
This work in progress will close #2807.

At the moment, it tries to make use of a dataset we found on osf.io.
A more stable version of this work would imply hosting this dataset on osf.io with an account that we trust.

- adding a fetcher for `fsaverage6` in `nilearn/datasets/struct.py`
- adding some tests for that fetcher in `nilearn/datasets/tests/test_struct.py`
- adding some documentation about `fsaverage6` in `nilearn/datasets/description/fsaverage6.rst`


jeromedockes:
as you plan to also add fsaverage1 and 4 (IIUC), maybe we can avoid creating a new function for each of them? if the contents are the same and the difference is just the url

nilearn/datasets/struct.py
```diff
@@ -562,3 +565,40 @@ def _fetch_surf_fsaverage5_sphere(data_dir=None):
...
+            'surf',
+            '{}.{}'.format(hemi, part)
+        ), url, opts)
+        for hemi in ["lh", "rh"]
+        for part in ["curl", "inflated", "pial", "white"]
+    ]
+
+    _fetch_files(dataset_dir, filenames)
+    result = {
+        name: os.path.join(dataset_dir, '{}.gii'.format(name))
```
jeromedockes, line 600: here could we make sure the keys for equivalent meshes are the same as in fsaverage and fsavrage5?

nilearn/datasets/description/fsaverage6.rst
```diff
@@ -0,0 +1,21 @@
...
+
+Content
+-------
+    :'pial_left': Gifti file, left hemisphere pial surface mesh
+    :'pial_right': Gifti file, right hemisphere pial surface mesh
+    :'infl_left': Gifti file, left hemisphere inflated pial surface mesh
+    :'infl_right': Gifti file, right hemisphere inflated pial
+                   surface mesh
+    :'sulc_left': Gifti file, left hemisphere sulcal depth data
+    :'sulc_right': Gifti file, right hemisphere sulcal depth data
```
jeromedockes, line 16: and also white matter?

nilearn/datasets/struct.py
alexisthual: Yes, definitely, this is what I was planning on doing this next :) Thanks!

nilearn/datasets/description/fsaverage6.rst
alexisthual: Yes, someone mentioned it in #2313. I might integrate the white surface in this PR, or not, we'll see :)

jeromedockes: white matter for fsaverage (aka fsaverage7) and fsaverage5 are aready downloaded by the existing fetchers, so you only need to worry about fsaverage6! #2313 had remained open by mistake AFAICT

alexisthual: However, maybe we could add the curvature to `fsaverage` and `fsaverage5`, what do you think?

alexisthual:
So here is the complete list of the items I wish to have for each resolution of fsaverage available:
- mesh coordinates `inflated`, `pial`, `sphere`, and `white`
- surface maps `area`, `curv`, `sulc` and `thickness` (@jeromedockes was wondering whether `area`, `curv` and `thickness` would really be useful, but as they weight 3 times less than any mesh coordinates item, I think it's worth adding them while we are at it! Also I need the curvature and area in my current projects)

Also, I am not sure I understand what the `{lh, rh}.orig` files are in freesurfer. The meshes they describe look exactly like `{lh, rh}.white` (the faces array are the same and all) but it feels like the coordinates were translated from one to the other. Which one should I keep?

bthirion:
> Also, I am not sure I understand what the `{lh, rh}.orig` files are in freesurfer. The meshes they describe look exactly like `{lh, rh}.white` (the faces array are the same and all) but it feels like the coordinates were translated from one to the other. Which one should I keep?

lh = left hemisphere
rh = right hemisphere
you need both.

bthirion:
Regarding information to include, `curv` and `sulc` are needed to make nice plots. 
`depth` and `area` too
I don't think  that `thickness` is useful for viz, but I don't see any big issue in having it. It's art from freesurfer standard output, so people might expect to get it.

alexisthual:
Great, I'll add all of them then!
Also, do we want to ship these files in the original freesurfer format or in gifti? For now, the fsaverage resolutions available ship in gifti but I feel we might want to stick to the original format... What do you think?

NicolasGensollen:
Hi @alexisthual 
I get an error when I try the following:

```python
fs6 = fetch_surf_fsaverage(mesh='fsaverage6')
```
```
...
OSError: Fetching aborted: Dataset has been downloaded but requested file was not provided:
URL: https://osf.io/3s4r2/download
Target file: /home/nicolas/nilearn_data/fsaverage6/tpl-fsaverage/fsaverage6/surf/lh.curl
Downloaded: /home/nicolas/nilearn_data/fsaverage6/db3b1e8a011da18b0d43a928e5d8688f/download
```

It looks like the temporary file `'/home/nicolas/nilearn_data/fsaverage6/db3b1e8a011da18b0d43a928e5d8688f/tpl-fsaverage/fsaverage6/surf/lh.curl'` isn't downloaded although most files like `lh.curv` or `lh.inflated` are. Note that they all get deleted without being copied to the final folder because of this missing file error. See here:

https://github.com/nilearn/nilearn/blob/ac4967d6a2cbe1f29647fc501a3caf12ad47560d/nilearn/datasets/utils.py#L804-L807

Do you have the same issue on your end?

I'll take a look at the failing tests this afternoon!

nilearn/datasets/struct.py
```diff
@@ -562,3 +565,40 @@ def _fetch_surf_fsaverage5_sphere(data_dir=None):
...
+    names = [
+        "{}.{}".format(hemi, part)
+        for hemi in ["lh", "rh"]
+        for part in ["curl", "inflated", "pial", "white"]
+    ]
+    filenames = [
+        (os.path.join(
+            'tpl-fsaverage',
+            'fsaverage6',
+            'surf',
```
NicolasGensollen, line 591: I'm wondering why you build such a complex folder system here. I think for fsaverage and fsaverage5 the files are directly under `fsaverage` and `fsaverage5` respectively.

nilearn/datasets/struct.py
```diff
@@ -562,3 +565,40 @@ def _fetch_surf_fsaverage5_sphere(data_dir=None):
 
     result['description'] = str(_get_dataset_descr('fsaverage5_sphere'))
     return Bunch(**result)
+
+
+def _fetch_surf_fsaverage6(data_dir=None):
+    """Helper function to ship fsaverage5 spherical meshes.
```
NicolasGensollen, line 571: I think the docstring needs to be updated

nilearn/datasets/struct.py
```diff
@@ -459,6 +459,7 @@ def fetch_surf_fsaverage(mesh='fsaverage5', data_dir=None):
         Which mesh to fetch. Default='fsaverage5'.
 
         - 'fsaverage5': the low-resolution fsaverage5 mesh (10242 nodes)
+        - 'fsaverage6': the medium-resolution fsaverage5 mesh (40962 nodes)
```
NicolasGensollen, line 462: typo here

nilearn/datasets/tests/test_struct.py
```diff
@@ -157,7 +157,7 @@ def test_fetch_icbm152_brain_gm_mask(tmp_path, request_mocker):
     assert isinstance(grey_matter_img, nibabel.Nifti1Image)
 
 
-@pytest.mark.parametrize("mesh", ["fsaverage5", "fsaverage"])
+@pytest.mark.parametrize("mesh", ["fsaverage5", "fsaverage", "fsaverage6"])
```
NicolasGensollen, line 160: We need to update the test a little bit to make it work.
I don't know if the plan is to ship fsaverage6 with nilearn as we are currently doing with fsaverage5 or not.
If we don't ship it, we need to mock the download of fsaverage6 in this test in the same way as for fsaverage. Once the structure of the files will be clear (see previous comments), we should be able to mock the download by adding an entry "*fsaverage6.tar.gz" in the `url_mapping` dict of the mocker. 
Does this make sense? (also pinging @jeromedockes here who probably has an opinion on this).

NicolasGensollen (commented):
Thanks @alexisthual ! 
I left some comments below. Let me know if you want to talk about this during the week.

nilearn/datasets/tests/test_struct.py
jeromedockes: yes exactly.
fsaverage6 would not be distributed with nilearn, but be downloaded and the test needs to mock that as you explain.
I think the plan of @alexisthual  was also to add other fsaverage meshes such as fsaverage4; if there is a very low-resolution one (significantly smaller than fsaverage5), then we could ship that one (in addition to fsaverage5) with nilearn and use it in unit tests instead of fsaverage5

bthirion: IMHO fsaverage6 is actually not that big. Wouldn't it make our life easire to ship it too ?

jeromedockes:
> IMHO fsaverage6 is actually not that big. Wouldn't it make our life easire to ship it too ? 

the archive downloaded in this pr ( https://osf.io/3s4r2/download ) is
18M compressed, and 32M uncompressed. currently the source distribution
of nilearn is 2.8M (compressed), so adding this archive would increase
nilearn's size a lot. (out of these 2.8M, about half is actually
occupied by fsaverage5)

I don't think having a downloader for fsaverage6 would make our life
very hard; we already have one for fsaverage7 and will probably add more
if we also want other resolutions of fsaverage. we included fsaverage5
and removed its downloader so that we could use it in unit tests.

bthirion:
Sorry, my comment was silly. I have to say that this is suprising, given that the mesh file is not that large, but I guess that there are many other files that go with it.

jeromedockes:
> Sorry, my comment was silly. I have to say that this is suprising, given that the mesh file is not that large, but I guess that there are many other files that go with it. 
that's true; the meshes only have 4 times as many nodes but the whole thing is
over 10 times bigger than fsaverage5. as you said there are more files,
like smoothness and curvature that we don't have for fsaverage5. also,
the files we have for fsaverage5 are in .gii format whereas the
fsaverage aren't; I don't know if this could make a difference

alexisthual:
I was still WIP when I pushed one week ago, so I resumed discussions which are no longer relevant :slightly_smiling_face: I'll try to address all on-going comments here!

Here is a quick summary of what I added:
- some logic to deal with `.area` files from freesurfer
- some logic to handle `mesh = "fsaverage{3,4,6}"` in `fetch_surf_fsaverage`
- some tests for both previous points

As of now, `infl`, `pial`, `sulc` and `white` would be available for both hemispheres for versions of fsaverage from 3 to 7, given that we upload (to osf.io for instance) archives of fsaverage 3, 4 and 6.

**There are still at least 2 issues we need to address and on which I need some input:**
1. the archives I am about to upload contain freesurfer files, whereas the fsaverage5 archive currently shipped with nilearn contains gifti files. I think it'd be much better to be consistent across versions, yet I feel like shipping original freesurfer files might be a better fit gifti files (because a. it's easier for me to do b. users might expect files they usually find in fs rather than gifti files). What do you all think?
1. I want to add new attributes to these datasets (thickness, curvature, sphere, etc), which are not currently present in the built-in fsaverage5 archive. Is it safe to add these new files in this PR? (I feel like it should be ok, but I think it's worth asking)

When we are clear on these, I'll be able to:
1. merge the logic for loading fsaverage7 with the new util function I built for fetching fsaverage 3, 4 and 6
1. all new attributes to all datasets

jeromedockes:
> **There are still at least 2 issues we need to address and on which I need some input:**
> 1. the archives I am about to upload contain freesurfer files, whereas the fsaverage5 archive currently shipped with nilearn contains gifti files. I think it'd be much better to be consistent across versions, yet I feel like shipping original freesurfer files might be a better fit gifti files (because a. it's easier for me to do b. users might expect files they usually find in fs rather than gifti files). What do you all think?

I don't think it is super important; nibabel can read both

> 1. I want to add new attributes to these datasets (thickness, curvature, sphere, etc), which are not currently present in the built-in fsaverage5 archive. Is it safe to add these new files in this PR? (I feel like it should be ok, but I think it's worth asking)
>

if they are small I guess it's ok. what is the new size of the
fsaverage5 archive after adding them? Although if we don't use them
anywhere why include them? the fetchers are only there to support
nilearn functionality.

> When we are clear on these, I'll be able to:
> 1. merge the logic for loading fsaverage7 with the new util function I built for fetching fsaverage 3, 4 and 6

great! note fetching fsaverage6 still fails for me with the same error
as for @NicolasGensollen

alexisthual:
> I don't think it is super important; nibabel can read both

I agree it might not seem very important, but I think some users will expect these files to be consistently named and formatted across resolutions.
Maybe we could bundle the freesurfer files in an archive in nilearn and uncompress this archive during the install? The current fsaverage5 folder shipped with nilearn is 1.3M. Adding area, curv, sphere and thickness in `.gii.gz` format would make it 1.8M. A zip file with all these files in the freesurfer format is 1.8M, 3.2M uncompressed. All in all, in seems equivalent to use one solution or the other.

> great! note fetching fsaverage6 still fails for me with the same error
as for NicolasGensollen

This is normal, since I haven't uploaded archives for fsaverage 3, 4 and 6 (the current url does not point to our final dataset) :slightly_smiling_face: I want to discuss issues 1 and 2 before uploading anything.

bthirion:
Maybe rather .gii then (indeed, most users won't even notice). Does it add some issues for you ?

alexisthual:
> Maybe rather .gii then

It's not particularly complicated I guess, I'll check the documentation and upload the script as a gist somewhere!

alexisthual:
However, it'd be very helpful if someone could point me to the scripts which helped parse the freesurfer files into gifti :slightly_smiling_face:

bthirion:
I'd simply read the files with nibabel and  save them under .gii format... I don't know if anyone has kept the script used for fsaverage5.

jeromedockes:
> Maybe we could bundle the freesurfer files in an archive in nilearn and uncompress this archive during the install?

if compressing each file individually (ie saving to `.gii.gz`) results in almost the same size as creating one archive containing them all I think I'd rather do the former.

If you go for a single archive, I'd suggest extracting files in the loader function rather than the install script to keep setup.py simple, and storing them in nilearn_data (ie not in the installation directory which contains the archive)

alexisthual:
I created and ran a script (available as [a gist](https://gist.github.com/alexisthual/919a1131c37c6bacb4bd5a0b68321cfb)) which uses `mris_convert` and `gzip` to build our `.gii.gz` files.

I quickly compared some of the computed files for fsaverage5 (pial right and sulc left) to verify that their data match those of files already present in nilearn:
* the file sizes are the same
* the data arrays are the same according to `np.array_equal`
* the outputs of `nibabel.gifti.gifti.GiftiImage.print_summary()` are the same too, except for the author of these files (hence the files have different md5 sums)

Overall, it feels like the files used in nilearn where computed this way in the first place.
Finally, I think the next reasonable step is to upload the files I just computed on osf.io and try to download them for real :slightly_smiling_face: 

In the meantime, how do you all feel about renaming the files shipped in nilearn for fsaverage5? I don't like doing this kind of things, but I also think it would be nice if their names matched the dataset's arguments. For instance, attribute `pial_left` points to a file named `pial.left.gii.gz` whereas `infl_left` points to `pial_inflated.left.gii.gz`. To me, it feels like attribute `{type}_{side}` should point to a file named `{type}_{side}.gii.gz`.

jeromedockes:
> In the meantime, how do you all feel about renaming the files shipped in nilearn for fsaverage5? I don't like doing this kind of things, but I also think it would be nice if their names matched the dataset's arguments. For instance, attribute `pial_left` points to a file named `pial.left.gii.gz` whereas `infl_left` points to `pial_inflated.left.gii.gz`. To me, it feels like attribute `{type}_{side}` should point to a file named `{type}_{side}.gii.gz`.

I don't have a strong opinion; I guess it would be nice to have the same
names as the original freesurfer files so we can easily see the
correspondance; having a direct mapping from attribute names to file
names as you mention would also be convenient. IIUC at the moment we
have neither. If you want to change them while keeping the old ones
available you could always add some symbolic links.

More generally, it is not entirely clear whether the layout and files in
the nilearn data directories are considered part of the public API or
not. I would say that they are *not*, meaning that names, formats,
directory organization etc. are subject to change without notice and the
only supported way of accessing this data is through the
`nilearn.datasets` functions -- storing the files on disk is like a
caching mechanism.

However I'm not sure that is the correct interpretation, and AFAICT
there isn't an explicit statement regarding this in the documentation,
although the [user guide](https://nilearn.github.io/manipulating_images/input_output.html#fetching-open-datasets-from-internet)
says:

  "Note that you can copy that folder to another user’s computers to avoid the initial dataset download on the first fetching call."

which seems to hint that these files' locations are stable.

In practice the tests don't enforce that these locations remain stable
and I don't think we have considered changing them as breaking backwards
compatibility.

If we consider these details as private, we could say so explicitly in
the documentation and you can rename the files however you see fit
(whatever makes the implementation of the fetchers clear to developers,
rather than users).

bthirion:
> If you want to change them while keeping the old ones available you could always add some symbolic links.

+1

bthirion:
> If we consider these details as private, we could say so explicitly in the documentation

I also think so. Yet for the rpesent case, I think we should maintain current names at least for one deprecation cycle.

alexisthual:
I tried to take everything into account and made the following changes:
1. uploaded [fsaverage3,4,5,6,7 tarballs](https://osf.io/3uvjg/quickfiles?view_only=) to osf.io using a personal account
1. merged fsaverage7 logic with that of fsaverage3,4,6
1. changed shipped fsaverage5 files so as to use the ones from my tarball, and added symlinks so that previously used names point to the new files
1. added `area`, `curv`, `sphere` and `thickness` for all datasets
1. updated the docs so that `fsaverage5_sphere` seems deprecated (I didn't change the code though)
1. updated the docs of `fetch_surf_fsaverage()` to mention that file names are subject to change while attributes will remain consistent

I tested loading and using fsaverage 3, 5 and 7 and plotting various attributes, everything seems to work as expected.
Also, I checked that calling `fetch_surf_fsaverage()` with `mesh` equal to `"fsaverage"` or `"fsaverage7"` actually downloads the new dataset and replaces the old one. Again, everything seems to work as expected.
I might have missed something though, don't hesitate to run this tests too!

In the end, I opted for using file names that match the attributes name, which yields a more concise code (and was easier to do for me).
Eventually, I feel like the logic for fsaverage5 could be merged to that of other versions (the two functions look very similar, the only difference is that we call `_fetch_files` whenever it is not fsaverage5) and that the util function for `fsarevage5_sphere` could be deleted some day, but I am not really sure how to indicate this in the docs.

nilearn/datasets/struct.py
```diff
@@ -531,11 +546,18 @@ def _fetch_surf_fsaverage5(data_dir=None, url=None, resume=True, verbose=1):
     """
     data = {"description": _get_dataset_descr("fsaverage5")}
     data_dir = Path(FSAVERAGE5_PATH)
-    for hemi in ["left", "right"]:
-        for part in ["white", "sulc", "pial", "infl"]:
-            data["{}_{}".format(part, hemi)] = str(
-                data_dir / "{}.{}.gii.gz".format(
-                    {"infl": "pial_inflated"}.get(part, part), hemi))
+
+    data = {
```
NicolasGensollen, line 550: You are redefining data here, and thus missing the description.

nilearn/datasets/struct.py
```diff
@@ -471,55 +479,62 @@ def fetch_surf_fsaverage(mesh='fsaverage5', data_dir=None):
...
+        # rename mesh to "fsaverage" to download it once
+        # regardless of whether mesh equals "fsaverage" or "fsaverage7"
+        if mesh == "fsaverage7":
+            mesh = "fsaverage"
+
+        return _fetch_surf_fsaverage(mesh, data_dir=data_dir)
+    elif mesh == "fsaverage5":
+        return _fetch_surf_fsaverage5()
+    elif mesh == "fsaverage5_sphere":
+        return _fetch_surf_fsaverage5_sphere(data_dir=data_dir)
```
NicolasGensollen, line 528: If we deprecate this option, we should add a warning here and not rely on `_fetch_surf_fsaverage5_sphere` (meaning implement the new way).

nilearn/datasets/struct.py
```diff
@@ -451,15 +451,23 @@ def fetch_oasis_vbm(n_subjects=None, dartel_version=True, data_dir=None,
...
     mesh : str, optional
         Which mesh to fetch. Default='fsaverage5'.
 
+        - 'fsaverage3': the low-resolution fsaverage3 mesh (642 nodes)
+        - 'fsaverage4': the low-resolution fsaverage4 mesh (2562 nodes)
         - 'fsaverage5': the low-resolution fsaverage5 mesh (10242 nodes)
-        - 'fsaverage5_sphere': the low-resolution fsaverage5 spheres (10242 nodes)
+        - 'fsaverage5_sphere': deprecated;
+            since v0.7.2, sphere coordinates are directly accessible from
+            fsaverage5.sphere_{left, right}
```
NicolasGensollen, line 468: I would rather say that this option is deprecated and will raise an error in release 0.9.0 and explain the new way of getting the spheres

NicolasGensollen (changes requested):
Thanks @alexisthual !
I left a few quick comments below. I think you are getting some errors because the description is missing for fsaverage5.
The test failures are difficult to see currently on GitHub actions since #2848. I'm working on fixing this (see #2856 ). Sorry about that...

nilearn/datasets/struct.py
jeromedockes: you can use the `sklearn.utils.deprecated` decorator on `_fetch_surf_fsaverage5_sphere` to issue a warning saying to use `fetch_surf_fsaverage("fsaverage5")` instead of `fetch_surf_fsaverage("fsaverage5_sphere")`

jeromedockes:
> the util function for fsarevage5_sphere could be deleted some day, but I am not really sure how to indicate this in the docs.

you can delete it already, as it is a "private" function (its name starts with an underscore so that users know that by convention they are not supposed to call it directly and it may be removed).

then you can state in `fetch_surf_fsaverage`'s docstring that the "fsaverage5_sphere" option is deprecated, as you did.
when `fetch_surf_fsaverage("fsaverage5_sphere")` is called, issue a warning, then use the output of `fetch_surf_fsaverage("fsaverage5")` to return the spherical mesh (the same output as what was returned before)

alexisthual:
> I left a few quick comments below. I think you are getting some errors because the description is missing for fsaverage5.

Ahh, yes, I think that was it, thanks for the pointers!

Thanks @NicolasGensollen and @jeromedockes for your help. I added a warning for `mesh="fsaverage5_sphere"` and deleted the fetcher.

bthirion (approved):
LGTM. This is great !
The only remaining thing is to subsitute fsaverage datasets to nilearn account no osf... which I don't master...

NicolasGensollen (commented):
Thanks @alexisthual ! This looks great! :+1: 

Could you add a whatsnew entry for this ?

>The only remaining thing is to subsitute fsaverage datasets to nilearn account no osf... which I don't master...

@bthirion do you mean you don't have the credentials? If you have them, I think we can take care of doing the upload with @alexisthual

bthirion:
I don't have them. I think that @jeromedockes  or @KamalakerDadi  has them.

jeromedockes:
> I don't have them. I think that @jeromedockes or @KamalakerDadi has them.

I don't. maybe @kchawla-pi ?

kchawla-pi:
I believe I had emailed the credentials to BT and Jerome. But I will resend them to you @NicolasGensollen .
Its also registered to my email, so you should change that as well. 
There was a common Nilearn account for events, nilearn.events@gmail.com . Perhaps that one or which ever you wish.

I will email you the login info at your gmail account (given on your Github profile).

NicolasGensollen:
Thanks @kchawla-pi I got your email! :+1:

alexisthual:
Great, I am out-of-office for now but I'll finish all this as soon as I come back!

alexisthual:
> Could you add a whatsnew entry for this ?

This is done now! I added two bullet points to the `enhancement` section.

> The only remaining thing is to subsitute fsaverage datasets to nilearn account no osf.

I re uploaded all data using a more official account. You can find it here: https://osf.io/mk53n/

Let me know if there is anything else to add! :sun_with_face:

doc/whats_new.rst
```diff
@@ -63,6 +63,12 @@ Enhancements
   has been removed.
 - Fetcher `nilearn.datasets.fetch_nyu_rest` is deprecated since release 0.6.2 and
   has been removed.
+- Fetcher `nilearn.datasets.fetch_surf_fsaverage` now accepts
+  `fsaverage3`, `fsaverage4` and `fsaverage6` as values for parameter `mesh`, so that
+  all resolutions of fsaverage from 3 to 7 are now available.
+- Fetcher `nilearn.datasets.fetch_surf_fsaverage` now provides
```
NicolasGensollen, line 69:
```suggestion
- Fetcher :func:`nilearn.datasets.fetch_surf_fsaverage` now provides
```

NicolasGensollen (approved):
LGTM, thanks @alexisthual !

kchawla-pi:
We should rename the PR title from "Attempt at adding support for downloading fsaverage6" to "Add support for downloading fsaverage6". it will be nicer in the squash merge history, and comply with typical issue/}R titles.

NicolasGensollen:
Merging, thanks!

PR #2816, opened 2021-05-07
Fix private functions

### LINKED ISSUE #2757: Some public functions from `nilearn.regions` are not exposed
Some public functions from the `regions` module are not exposed. For example:

https://github.com/nilearn/nilearn/blob/00d3873fb6cf23d09f8eba6d0d20e905f26d22dd/nilearn/regions/rena_clustering.py#L148-L152

https://github.com/nilearn/nilearn/blob/00d3873fb6cf23d09f8eba6d0d20e905f26d22dd/nilearn/regions/rena_clustering.py#L349-L355

Or

https://github.com/nilearn/nilearn/blob/00d3873fb6cf23d09f8eba6d0d20e905f26d22dd/nilearn/regions/rena_clustering.py#L286-L291

We should either make them private or add them here:

https://github.com/nilearn/nilearn/blob/00d3873fb6cf23d09f8eba6d0d20e905f26d22dd/nilearn/regions/__init__.py#L15-L20

achamma723:
This PR is fixing the names of some private functions in ```rena_clustering```
Closes #2757

### DIFF
--- a/nilearn/regions/rena_clustering.py
+++ b/nilearn/regions/rena_clustering.py
@@ -145,7 +145,7 @@ def _make_edges_and_weights(X, mask_img):
     return edges, weights
 
 
-def weighted_connectivity_graph(X, mask_img):
+def _weighted_connectivity_graph(X, mask_img):
     """Creating symmetric weighted graph: data and topology are encoded by a
     connectivity matrix.
 
@@ -283,7 +283,7 @@ def _reduce_data_and_connectivity(X, labels, n_components, connectivity,
     return reduced_connectivity, reduced_X
 
 
-def nearest_neighbor_grouping(X, connectivity, n_clusters, threshold=1e-7):
+def _nearest_neighbor_grouping(X, connectivity, n_clusters, threshold=1e-7):
     """Cluster using nearest neighbor agglomeration: merge clusters according
     to their nearest neighbors, then the data and the connectivity are reduced.
 
@@ -346,9 +346,9 @@ def nearest_neighbor_grouping(X, connectivity, n_clusters, threshold=1e-7):
     return reduced_connectivity, reduced_X, labels
 
 
-def recursive_neighbor_agglomeration(X, mask_img, n_clusters,
-                                     n_iter=10, threshold=1e-7,
-                                     verbose=0):
+def _recursive_neighbor_agglomeration(X, mask_img, n_clusters,
+                                      n_iter=10, threshold=1e-7,
+                                      verbose=0):
     """Recursive neighbor agglomeration (ReNA): it performs iteratively
     the nearest neighbor grouping.
     See :footcite:`Hoyos2019`.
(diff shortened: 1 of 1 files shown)


nilearn/regions/rena_clustering.py
```diff
@@ -346,9 +346,9 @@ def nearest_neighbor_grouping(X, connectivity, n_clusters, threshold=1e-7):
     return reduced_connectivity, reduced_X, labels
 
 
-def recursive_neighbor_agglomeration(X, mask_img, n_clusters,
-                                     n_iter=10, threshold=1e-7,
-                                     verbose=0):
+def _recursive_neighbor_agglomeration(X, mask_img, n_clusters,
```
bthirion, line 349: I think that this is a high-level function that can be called/used outside of this module. I would not make it private. But I guess that this is open for discussion

bthirion (commented):
Looks good overall.

NicolasGensollen (commented):
LGTM, thanks @achamma723 !
`recursive_neighbor_agglomeration` is currently not exposed and not documented in the API:

https://github.com/nilearn/nilearn/blob/4f0ef229fbaa1ee32cd2eff352560381dc4ed8e2/nilearn/regions/__init__.py#L15-L20

If we decide to make it public we should change that as part of this PR I think.

NicolasGensollen:
I don't have a strong opinion on making `recursive_neighbor_agglomeration` public. 
I don't think it is used outside of `rena_clustering.py`.

thomasbazeille:
It seems this has been superseeded by #2911 and can be closed

NicolasGensollen:
>It seems this has been superseeded by #2911 and can be closed

I think #2911 only improved the docstrings of the functions/classes in the region module but didn't change the nature of these functions. AFAICT, they are still defined as public functions (without a leading "_"), but not exposed in the public API.

I don't mind making `recursive_neighbor_agglomeration` a proper public function if users want to use it directly.

bthirion:
Can we  bypass the deprecation cycles, given that the functions are not listed ? Or should we have the deprecation cycle ?
+0 for deprecation cycle
and
+1 for making `recursive_neighbor_agglomeration` public.
Best

NicolasGensollen:
>Can we bypass the deprecation cycles, given that the functions are not listed ?

I don't think we need a deprecation cycle. These functions were not exposed and not listed in the public API. They were private functions without the proper leading "_" convention.
Maybe we can still list the name changes in the whats_new file.

@achamma723 do you want to finish this? I think that we still need to do:

- [ ] Fix the confilct
- [ ] Change `_recursive_neighbor_agglomeration` back to `recursive_neighbor_agglomeration`
- [ ] Export `recursive_neighbor_agglomeration` in `nilean/regions/__init__.py`
- [ ] Add `recursive_neighbor_agglomeration` in `doc/modules/reference.rst`
- [ ] Add a whats_new entry

Issue #1011, opened 2016-02-25
signals.clean: Consider filtering based on a cosine base


GaelVaroquaux:
According to @pbellec, this has two benefits compared to our butterworth filter:
- Not having ugly side effects at the beginning and of the acquisition (need to check this)
- Enabling missing data, as eg in a later possible implementation of scrubbing

GaelVaroquaux:
@bthirion : what's your feeling on cosine-based signal cleaning with regards to butterworth?

bthirion:
No strong feeling. I have always used cosine filters (DCT bases).
It would be great to document any bad behavior of butterworth (e.g. border effect).

danlurie:
Just encountered the Butterworth border effects in some of my own analysis. Nasty surprise until I realized what was causing the issue. I'm happy to add a note to the docs warning users about this. Where would be the right place to put this? With signal.clean?

bthirion:
This would be in signal.clean.
Can you show an exmaple of poor outcome of the Butterworth filter ?
Thanks,

danlurie:
I haven't to dug into the issues comprehensively, but one thing I have been able to replicate on multiple datasets (my own data and HCP) is that the filtering with `signal.clean` results in artificially low signal standard deviation at the beginning and end of a time series.

![Butterworth weirdness](https://cloud.githubusercontent.com/assets/1873103/26084332/2d39f2ee-3990-11e7-8bfa-b1bddca16c04.jpg)

Here's the code I used to filter and regress out nuisance signals:
```
masker = input_data.NiftiLabelsMasker(labels_img=bna_img, background_label=0,
                mask_img=bold_T1w_brainmask, standardize=False, detrend=True, 
                low_pass=0.2, high_pass=0.01, t_r=2, resampling_target="data")
```

jerdra:
> I haven't to dug into the issues comprehensively, but one thing I have been able to replicate on multiple datasets (my own data and HCP) is that the filtering with `signal.clean` results in artificially low signal standard deviation at the beginning and end of a time series.
> ...


Hey, sorry if necro-bump, but has anyone attempted to implement DCT-based cosine regressors in lieu of signal.clean's butterworth filter? I've recently attempted this (albeit I'm not a signal processing expert) and have run into the same issue @danlurie has noted for the butterworth implementation *but worse* with a DCT v1.1 implementation.

[...]

PR #2821, opened 2021-05-07
RF extract complex methods out from signal.clean 

### LINKED ISSUE #2820: Refactor `signals.clean`
The current code in `signals.py`, especially `signal.clean` is difficult to follow. To facilitate the integration of `load_confounds` #2777, and easier integration of DCT #1011, we should refactor it to a level where the data flow is easier to follow.

I would suggest, for now, extract some methods out just to clean up `signal.clean`. We can discuss if this function should be written into a module.

htwangtw:
Closes #2820 
The aim of this PR is to simplify the data flow inside `signal.clean` for future maintenance.
I have left a few place holder for DCT #1011 @jerdra

I added an input argument `filter` for specifying the filtering method, preempting DCT to be added. Currently, I set the default to "Butterworth" to prevent breaking the tests. If we want that part to behave more logically, further work will be needed.


nilearn/signal.py
```diff
@@ -17,6 +17,9 @@
 from ._utils.numpy_conversions import csv_to_array, as_ndarray
 
 
+availiable_filters = ["butterworth", ]
```
bthirion, line 20:
```suggestion
available_filters = ['butterworth', ]
```

nilearn/signal.py
```diff
@@ -407,8 +410,8 @@ def _ensure_float(data):
 
 
 def clean(signals, sessions=None, detrend=True, standardize='zscore',
-          confounds=None, standardize_confounds=True, low_pass=None,
-          high_pass=None, t_r=2.5, ensure_finite=False):
+          confounds=None, standardize_confounds=True, filter="butterworth",
```
bthirion, line 413:
```suggestion
          confounds=None, standardize_confounds=True, filter='butterworth',
```

nilearn/signal.py
```diff
@@ -641,3 +567,107 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
                                detrend=False)
 
     return signals
+
+def _process_session(signals, sessions, detrend, standardize, confounds, low_pass, high_pass, t_r):
```
bthirion, line 571: I think that what is called a session here is actually a run in BIDS language.

nilearn/signal.py
```diff
@@ -499,124 +507,42 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
...
-                raise TypeError("confound has an unhandled type: %s"
-                                % confound.__class__)
-            all_confounds.append(confound)
-
-        # Restrict the signal to the orthogonal of the confounds
-        confounds = np.hstack(all_confounds)
-        del all_confounds
+    # Read confounds and signals
+    signals, confounds = _sanitize_inputs(signals, confounds, ensure_finite)
+    # check if filter paramters are satidfied
```
bthirion, line 512:
```suggestion
    # check if filter parameters are satisfied
```

nilearn/signal.py
bthirion: Oh, what you mean is that you are processing a set of runs, so rather `_process_runs` maybe

nilearn/signal.py
```diff
@@ -407,8 +410,8 @@ def _ensure_float(data):
 
 
 def clean(signals, sessions=None, detrend=True, standardize='zscore',
```
bthirion, line 412: maybe we should deprecate `sessions` here, so that we can use  `runs` instead in a future release (0.9).

bthirion (commented):
Thx ! looks quite good already.
There a re a few lines not covered by tests.

nilearn/signal.py
htwangtw: I agree that's a more sensible name

nilearn/signal.py
htwangtw: Sounds good. May I ask if there's an example for implementing deprecation warning in nilearn? I have never handled this kind of things before.

bthirion (commented):
Thx, this is looking pretty good.

htwangtw:
I tried to add the cosine drift method from design_matrix but that introduced an error while running the test:

```
ImportError while loading conftest '/home/hw1012/git_repos/nilearn/nilearn/conftest.py'.
nilearn/conftest.py:9: in <module>
    from nilearn.datasets._testing import request_mocker  # noqa: F401
nilearn/datasets/__init__.py:6: in <module>
    from .struct import (fetch_icbm152_2009, load_mni152_template,
nilearn/datasets/struct.py:16: in <module>
    from ..image import new_img_like, get_data
nilearn/image/__init__.py:5: in <module>
    from .resampling import resample_img, resample_to_img, reorder_img, \
nilearn/image/resampling.py:16: in <module>
    from .image import crop_img
nilearn/image/image.py:19: in <module>
    from .. import signal
nilearn/signal.py:17: in <module>
    from .glm.first_level.design_matrix import _cosine_drift
nilearn/glm/__init__.py:8: in <module>
    from nilearn.glm.contrasts import (
nilearn/glm/contrasts.py:14: in <module>
nilearn/input_data/__init__.py:5: in <module>
    from .nifti_masker import NiftiMasker
nilearn/input_data/nifti_masker.py:12: in <module>
    from .base_masker import BaseMasker, filter_and_extract
nilearn/input_data/base_masker.py:15: in <module>
    from .. import masking
nilearn/masking.py:15: in <module>
    from .image import new_img_like
E   ImportError: cannot import name 'new_img_like' from 'nilearn.image' (/home/hw1012/git_repos/nilearn/nilearn/image/__init__.py)
```

I am certain it is related to `from .glm.first_level.design_matrix import _cosine_drift`, but I am not sure why or how?
I can either hold off adding `filter='cosine'` in the current code or resolve this import error. 
What do you think?

NicolasGensollen:
Hi @htwangtw 
You're getting a circular import issue. I think one simple solution might be to put the import (`from .glm.first_level.design_matrix import _cosine_drift`) inside the function which needs it (`_filter_signal`).
Let me know if this solves the issue!

jerdra:
Two quick questions related to this PR:

- Is there a reason why the `_cosine_drift` function could not be extended to perform low-pass? 
- Would it make sense to also add an option for fourier drift? I know that some people prefer this (although I'm aware that they yield similar results in the end)
 
P.S: Let me know if the issues are a more appropriate place to discuss this!

htwangtw:
I am think perhaps I will finish the refactoring, and implement the existing dct function. Once merged, @jerdra can open a separate issue to improve dct and add other filtering functions.
What do you think?

NicolasGensollen:
@htwangtw @jerdra I agree, I think it would be easier to keep this PR on the pure refactoring of signal.clean and have other discussions within dedicated issues/PRs.

nilearn/signal.py
```diff
@@ -641,3 +562,129 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
...
+        if confounds is not None:
+            # Apply low- and high-pass filters to keep filters orthogonal
+            # (according to Lindquist et al. (2018))
+            confounds = butterworth(confounds, sampling_rate=1. / t_r,
+                                    low_pass=low_pass, high_pass=high_pass)
+    elif filter == "cosine":
+        from .glm.first_level.design_matrix import _cosine_drift
+        frame_times = np.arange(signals.shape[0]) * t_r
+        cosine_drift = _cosine_drift(high_pass, frame_times)
+        confounds = np.hstack((confounds, cosine_drift))
```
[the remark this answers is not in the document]
htwangtw: The last column of output from _cosine_drift will always be a column of ones, but it doesn't hurt to do a check here.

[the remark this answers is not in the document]
jerdra: Would there be an issue with the bias term of `_cosine_drift` being redundant with `standardize=True`? Since the mean will be removed anyway? Or is this a non-problem?

htwangtw:
So the cosine drift filter is currently just added the existing function from the `design_matrix` module. Currently, I only added tests to check the outputs of private function `_filter_signal` with `filter='cosine'` are the expected shape/values. Would it be okay to leave the proper implementation of DCT to another PR, and a proper test on the cleaned signal?

nilearn/signal.py
```diff
@@ -436,8 +442,8 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
         Timeseries. Must have shape (instant number, features number).
         This array is not modified.
 
-    sessions : numpy array, optional
-        Add a session level to the cleaning process. Each session will be
+    runs : numpy array, optional
```
bthirion, line 445: Here we are changing the API. This needs a deprecation cycle (raise a warning, saying that sessions will result in an error after release 0.9)

nilearn/signal.py
```diff
@@ -641,3 +562,136 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
...
+
+
+
+def _process_runs(signals, runs, detrend, standardize, confounds, low_pass, high_pass, t_r):
+    """Process each run independently."""
+    if len(runs) != len(signals):
+        raise ValueError(('The length of the session vector (%i) '
+                            'does not match the length of the signals (%i)')
+                            % (len(runs), len(signals)))
+    for s in np.unique(runs):
```
bthirion, line 595:
```suggestion
    for run in np.unique(runs):
```(nitpick, but this is more readable)

File: nilearn/tests/test_signal.py [hunk 9]
```diff
@@ -486,8 +491,16 @@ def test_clean_confounds():
                   confounds=filename1)
     pytest.raises(TypeError, nisignal.clean, signals,
                   confounds=[None])
+    pytest.raises(ValueError, nisignal.clean, signals, filter='cosine',
```
bthirion, line 494: Maybe indicate that this is supposed t fail due to t_r being None

nilearn/tests/test_signal.py
```diff
@@ -689,3 +702,28 @@ def test_clean_zscore():
...
     np.testing.assert_almost_equal(cleaned_signals.std(0), 1)
+
+
+def test_cosine_filter():
+    '''Testing cosine filter interface and output.'''
+    from nilearn.glm.first_level.design_matrix import _cosine_drift
+
+    t_r, high_pass, low_pass, filter = 2.5, 0.002, None, 'cosine'
+    signals, _, confounds = generate_signals(n_features=41,
+                                                  n_confounds=5, length=45)
```
bthirion, line 713: There may be an indentation problem here.

bthirion (commented):
This is really looking great, thx !
Please also add a whatsnew entry.

nilearn/signal.py
htwangtw: Yes this has been done in L414 with `rename_parameters`! I will add the description to the docstring as well.

nilearn/signal.py
```diff
@@ -641,3 +565,137 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
...
+        from .glm.first_level.design_matrix import _cosine_drift
+        frame_times = np.arange(signals.shape[0]) * t_r
+        cosine_drift = _cosine_drift(high_pass, frame_times)
+        if confounds is None:
+            confounds = cosine_drift.copy()
+        else:
+            confounds = np.hstack((confounds, cosine_drift))
+    return signals, confounds
+
+
```
bthirion, line 590: Leave only 2 blank lines.

nilearn/signal.py
```diff
@@ -641,3 +565,137 @@ def clean(signals, sessions=None, detrend=True, standardize='zscore',
...
+        if confound.shape[0] != n_signal:
+            raise ValueError("Confound signal has an incorrect length")
+    elif isinstance(confound, np.ndarray):
+        if confound.ndim == 1:
+            confound = np.atleast_2d(confound).T
+        elif confound.ndim != 2:
+            raise ValueError("confound array has an incorrect number "
+                                        "of dimensions: %d" % confound.ndim)
+        if confound.shape[0] != n_signal:
+            raise ValueError("Confound signal has an incorrect length")
```
bthirion, line 652: It is good to give the actual and expected value in the error message.

bthirion (commented):
This is great, thx !
Just nitpicking on 2 details, but otherwise it's ready to go.

bthirion (approved):
AFAIK the circleci is unrelated.  
Thx !

NicolasGensollen:
@htwangtw I pushed a temporary fix on master which enables us to build the doc while the issue is resolved on sphinx's side. 
Could you merge master to fix the CI on this branch?
Thanks!

NicolasGensollen (approved):
LGTM, thanks @htwangtw !

bthirion:
Merging, thx !

Issue #2768, opened 2021-04-22
Decoder with dummy classifier estimator always predict same label


NicolasGensollen:
It looks like the decoder with dummy classifier is always predicting the first class no matter the frequencies of the labels.

Here is an example using the Haxby dataset, where I change the target frequencies:

```python
import numpy as np
import pandas as pd
from numpy.random import choice
from collections import Counter
from nilearn.image import index_img
from nilearn.datasets import fetch_haxby
from nilearn.decoding import Decoder

# Load and pre-process data
haxby_dataset = fetch_haxby()
behavioral = pd.read_csv(haxby_dataset.session_target[0], delimiter=' ')
conditions = behavioral['labels']
# We consider 4 classes here
condition_mask = conditions.isin(['face', 'cat', 'bottle', 'shoe'])
data_subset = index_img(haxby_dataset.func[0], 
                        condition_mask)
conditions = conditions[condition_mask].values
print("Classes are initially perfectly balanced:")
print(Counter(conditions))
# Split data in train test for simplicity
train = index_img(data_subset, slice(0, -20)) 
test  = index_img(data_subset, slice(-20, None))
conditions_train = conditions[:-20]
conditions_test = conditions[-20:]
print("Dummy decoder predictions:")
dummy_decoder = Decoder(estimator='dummy_classifier', 
                        mask=haxby_dataset.mask_vt[0])
dummy_decoder.fit(train, 
                  conditions_train)
prediction = dummy_decoder.predict(test)
dummy_decoder.classes_[prediction]
```
```
Classes are initially perfectly balanced:
Counter({'face': 108, 'cat': 108, 'shoe': 108, 'bottle': 108})
Dummy decoder predictions:
array(['bottle', 'bottle', 'bottle', 'bottle', 'bottle', 'bottle',
       'bottle', 'bottle', 'bottle', 'bottle', 'bottle', 'bottle',
       'bottle', 'bottle', 'bottle', 'bottle', 'bottle', 'bottle',
       'bottle', 'bottle'], dtype='<U6')
```

```python
# Resample the labels to make bottle very unfrequent
# and face very frequent, such that we would expect the
# dummy decoder to always predict face rather than bottle
fake_conditions = choice(['face', 'cat', 'bottle', 'shoe'], 
                         len(conditions), 
                         p=[0.6, 0.2, 0.05, 0.15])
print("Resample the labels to make bottle very unlikely:")
print(Counter(fake_conditions))
[...]

Issue #2767, opened 2021-04-22
Inconsistent predict return values for decoders


NicolasGensollen:
It looks like decoders with dummy estimators do not return labels when predicting, although this is what decoders with other estimators do. This can give unexpected results if one is not careful (accuracy will be zero for instance...).

```python
import numpy as np
import pandas as pd
from nilearn.image import index_img
from nilearn.datasets import fetch_haxby
from nilearn.decoding import Decoder

haxby_dataset = fetch_haxby()
behavioral = pd.read_csv(haxby_dataset.session_target[0], delimiter=' ')
conditions = behavioral['labels']
condition_mask = conditions.isin(['face', 'cat', 'bottle'])
data_subset = index_img(haxby_dataset.func[0], 
                        condition_mask)
conditions = conditions[condition_mask].values

train = index_img(data_subset, slice(0, -50)) 
test  = index_img(data_subset, slice(-50, None))
conditions_train = conditions[:-50]
conditions_test = conditions[-50:]

def acc(prediction, y_test=conditions_test):
    print("Prediction Accuracy: {:.3f}".format(
        (prediction == y_test).sum() / float(len(y_test))))

# Train a decoder and predict
decoder = Decoder(estimator='svc',              
                  mask=haxby_dataset.mask_vt[0],
                  standardize=True)    
decoder.fit(train, conditions_train)
prediction = decoder.predict(test)
print(prediction)
acc(prediction)

# Train a dummy decoder and predict
dummy_decoder = Decoder(estimator='dummy_classifier', 
                        mask=haxby_dataset.mask_vt[0])
dummy_decoder.fit(train, 
                  conditions_train)
prediction_dummy = dummy_decoder.predict(test)
print(prediction_dummy)
acc(prediction_dummy)

# Here is what you'd expect:
print(dummy_decoder.classes_[prediction_dummy])
acc(dummy_decoder.classes_[prediction_dummy])
```
```
['cat' 'cat' 'cat' 'cat' 'face' 'bottle' 'bottle' 'bottle' 'bottle'
 'bottle' 'bottle' 'bottle' 'bottle' 'bottle' 'face' 'face' 'face' 'face'
 'face' 'face' 'face' 'cat' 'cat' 'bottle' 'bottle' 'bottle' 'bottle'
 'bottle' 'bottle' 'bottle' 'bottle' 'cat' 'bottle' 'cat' 'face' 'face'
[...]

PR #2769, opened 2021-04-22
[WIP][FIX] Bugfix dummy decoder

NicolasGensollen:
Fixes #2768 #2767 

WIP...

PR #2826, opened 2021-05-07
[MRG] Bugfix Decoder Dummy Classifier

tbng:
* Closes gh-2768 
* Closes gh-2767
* Continuation of PR gh-2769

Multiple changes for Decoder.Dummyclassifier

* Adding better test case for Dummy Classifier: now taken into account of the affect of distribution of the label on predicted value  
* Change default of the Dummy Classifier to `strategy='stratified'`
* Simplify calculation of prediction value


nilearn/decoding/tests/test_decoder.py
```diff
@@ -184,51 +184,91 @@ def test_decoder_binary_classification():
...
+    y = np.zeros(n_samples)
+    y[:int(proportion * n_samples)] = 1.0
+
     model = Decoder(estimator='dummy_classifier', mask=mask)
     model.fit(X, y)
     y_pred = model.predict(X)
-    assert model.scoring == "roc_auc"
-    assert model.score(X, y) == 0.5
-    assert accuracy_score(y, y_pred) == 0.5
+    assert np.sum(y_pred == 1.0) / n_samples - proportion < 0.5
```
bthirion, line 230: Isn't this bound  bit loose ?

tbng: Yes, potentially so, but because the Dummy Classifier use numpy.random.multinomial to generate prediction label based on the proportion of label in training set so there would be a fair amount of randomness.

I think one way to decrease the bound here and still guarantee the test would not fail due to randomness is to increase the sample size (maybe 3x-4x), wdyt?

bthirion: Yes, if it's not too time consuming.

nilearn/decoding/decoder.py
```diff
@@ -57,9 +55,9 @@
     ridge_regressor=RidgeCV(),
     ridge=RidgeCV(),
     svr=SVR(kernel='linear', max_iter=1e4),
-    dummy_classifier = DummyClassifier(strategy='prior',
-                                       random_state=0),
-    dummy_regressor = DummyRegressor(strategy='mean'),
+    dummy_classifier=DummyClassifier(strategy='stratified',
```
NicolasGensollen, line 58: Could you add spaces around "="?

nilearn/decoding/decoder.py
```diff
@@ -57,9 +55,9 @@
...
+                                     random_state=0),
+    dummy_regressor=DummyRegressor(strategy='mean'),
```
NicolasGensollen, line 60: same

NicolasGensollen (commented):
This is looking good, thanks @tbng !
I tried it locally and the dummy estimators now behave as expected.
Let us know when you think this is ready (title still says [WIP]).

nilearn/decoding/decoder.py
tbng: Sorry this is automatically corrected by my editor. Actually it's a bit weird because the above estimator definition does not have spaces around "=".  WDYT?

NicolasGensollen: Sorry, I misunderstood what you were saying @tbng, thanks for clarifying! 
Indeed, this convention is mostly respected across nilearn, but there are still a few exceptions here and there...

NicolasGensollen (commented):
This looks good to me, thanks @tbng !
Could you add a whatsnew entry?

NicolasGensollen:
Hi @tbng 
I think this was ready to merge minus the missing whatsnew entry. 
Could you also merge master such that we get the latest's fixes? 
Thx!

tbng:
> Hi @tbng
> I think this was ready to merge minus the missing whatsnew entry.
> Could you also merge master such that we get the latest's fixes?
> Thx!

Sorry I was neglecting a bit for finishing this PR, but I just added the entry

bthirion:
Can you solve the conflict ?Thx

NicolasGensollen (approved):
LGTM, thanks!

NicolasGensollen:
Merging, thanks!

PR #2663, opened 2021-01-18
Deprecate numpy arrays and bytes in datasets module

NicolasGensollen:
See #2660

PR #2655, opened 2021-01-07
change dataset description from bytes to str

NicolasGensollen:
Only targets the description. 
Related to #1635 and #1489

PR #2829, opened 2021-05-07
[FIX] Replace numpy with pandas in data loaders

### LINKED ISSUE #2621: Use pandas in the data loaders
Since Pandas was added to the dependencies, we can use it in the data loaders to read CSV files instead of Numpy.
See #2619

achamma723:
This PR is fixing the issue of replacing ```Numpy``` with ```Pandas``` in data loaders:
Closes #2621


NicolasGensollen:
Thanks @achamma723 !
FYI: this change to pandas is a breaking change and has been scheduled for release 0.9 (see also #2655)

bthirion (approved):
LGTM, thx !

NicolasGensollen:
Hi @achamma723 
Could you rebase this PR ad we are approaching release 0.9.0 ?

achamma723:
Hello @NicolasGensollen , Is there an error with `test_fetch_atlas_destrieux_2009`?

nilearn/datasets/atlas.py
```diff
@@ -237,7 +238,7 @@ def fetch_atlas_destrieux_2009(lateralized=True, data_dir=None, url=None,
     files_ = _fetch_files(data_dir, files, resume=resume,
                           verbose=verbose)
 
-    params = dict(maps=files_[1], labels=np.recfromcsv(files_[0]))
+    params = dict(maps=files_[1], labels=pd.read_csv(files_[0]))
```
jeromedockes, line 241: This would need to be something like 
```python
    params = dict(maps=files_[1], labels=pd.read_csv(files_[0], index_col=0).to_records())
```

jeromedockes:
> Hello @NicolasGensollen , Is there an error with test_fetch_atlas_destrieux_2009?

there isn't an error with the test; the PR changes the output of the function: it is now a pd DataFrame with 2 columns (one repeating the index) rather than a 1d np recarray

.ipynb_checkpoints/Untitled-checkpoint.ipynb
```diff
@@ -0,0 +1,6 @@
+{
+ "cells": [],
```
jeromedockes, line 2: could you remove the checkpoints and the notebook? thanks!

nilearn/datasets/atlas.py
```diff
@@ -1054,10 +1055,10 @@ def fetch_coords_dosenbach_2010(ordered_regions=True):
     fdescr = _get_dataset_descr(dataset_name)
     package_directory = os.path.dirname(os.path.abspath(__file__))
     csv = os.path.join(package_directory, "data", "dosenbach_2010.csv")
-    out_csv = np.recfromcsv(csv)
+    out_csv = pd.read_csv(csv)
 
     if ordered_regions:
-        out_csv = np.sort(out_csv, order=['network', 'name', 'y'])
+        out_csv = out_csv.sort_values(by=['network', 'name', 'y'])
```
jeromedockes, line 1061: note if the output changes (eg the type of the container or the values) it is a breaking change and needs to go through a deprecation cycle. here for example rois, networks and labels are changed

nilearn/datasets/atlas.py
```diff
@@ -1130,16 +1129,15 @@ def fetch_coords_seitzman_2018(ordered_regions=True):
...
+    rois.columns = list(rois.columns[:-1]) + ["region"]
 
     if ordered_regions:
-        rois = np.sort(rois, order=['network', 'y'])
+        rois = rois.sort_values(by=['network', 'y'])
 
     params = dict(rois=rois[['x', 'y', 'z']],
                   radius=rois['radius'],
-                  networks=rois['network'].astype(str),
+                  networks=rois['network'],
```
jeromedockes, line 1140: here also networks changes from numpy array to Series

nilearn/datasets/func.py
```diff
@@ -740,20 +740,17 @@ def _is_valid_path(path, index, verbose):
...
     subject_names = csv_data["participant_id"].tolist()
     subjects_indices = []
     for name in subject_ids:
-        name = name.encode("utf8")
         if name not in subject_names:
             continue
         subjects_indices.append(subject_names.index(name))
-    csv_data = csv_data[subjects_indices]
-
+    csv_data = csv_data.iloc[subjects_indices]
```
jeromedockes, line 753: ext_vars type changes

nilearn/datasets/struct.py
```diff
@@ -735,14 +736,14 @@ def fetch_oasis_vbm(n_subjects=None, dartel_version=True, data_dir=None,
...
-    csv_data = np.recfromcsv(ext_vars_file)
+    csv_data = pd.read_csv(ext_vars_file)
     # Comparisons to recfromcsv data must be bytes.
     actual_subjects_ids = [("OAS1" +
                             str.split(os.path.basename(x),
                                       "OAS1")[1][:9]).encode()
                            for x in gm_maps]
     subject_mask = np.asarray([subject_id in actual_subjects_ids
-                               for subject_id in csv_data['id']])
+                               for subject_id in csv_data['ID']])
```
jeromedockes, line 746: type and column name change

nilearn/datasets/tests/test_func.py
```diff
@@ -170,20 +170,20 @@ def test_fetch_localizer_contrasts(tmp_path, request_mocker, localizer_mocker):
     assert not hasattr(dataset, 'tmaps')
     assert not hasattr(dataset, 'masks')
     assert isinstance(dataset.cmaps[0], str)
-    assert isinstance(dataset.ext_vars, np.recarray)
+    assert isinstance(dataset.ext_vars, pd.DataFrame)
     assert len(dataset.cmaps) == 2
-    assert dataset.ext_vars.size == 2
+    assert len(dataset['ext_vars']) == 2
```
jeromedockes, line 175: if you need to edit the test it means backwards compatibility is broken. It is possible that making the change is the best solution (eg we could decide that we always return pandas dataframes or series instead of numpy recarrays) but it needs to go through a deprecation cycle, and be discussed because it places a burden on users and so it can be done consistently for all fetchers

jeromedockes:
@NicolasGensollen  I couldn't tell from the issue description: did we just want to start using pandas internally, or do we want to start returning DataFrames or Series instead of recarrays wherever they were used? In the latter case we need a deprecation cycle

NicolasGensollen:
@jeromedockes there is already a deprecation cycle running, initiated with #2663, and supposed to end with release 0.9.

NicolasGensollen:
Also linking related #2655

jeromedockes:
> @jeromedockes there is already a deprecation cycle running, initiated with #2663, and supposed to end with release 0.9.

Ok thanks! in that case we need to track down those that still return recarrays, such as `fetch_atlas_difumo`

NicolasGensollen:
@achamma723 @jeromedockes I might be missing something, but I don't understand why we would sometimes return pandas dataframes and sometimes rec arrays... :thinking: 
My understanding of the goal of this PR was to get rid or rec arrays in all the fetchers, and use (and return) pandas dataframes instead, right?
I see even a few fetchers where we load through Pandas to get a dataframe, and then convert to records.

jeromedockes:
> My understanding of the goal of this PR was to get rid or rec arrays in all the fetchers, and use (and return) pandas dataframes instead, right?

yes, IIUC it will always be dataframes or series

NicolasGensollen (commented):
@achamma723 Could you rebase or merge master to get the doc changes of #3083 ?

NicolasGensollen (commented):
Thanks for merging master @achamma723 
In the last meeting, we decided to extend the deprecation cycle for rec arrays as we realized the current warning might have been a little light for such a change.
The idea would be to add an optional parameter to affected fetchers (`return_dataframe` or something alike) defaulting to the current behavior of returning rec arrays for backward compatibility. 
Let me know if you want to take care of this, otherwise I can commit to your branch.

jeromedockes:
so many fetchers may have a part that roughly looks like:
```python
if legacy_format:
    result["confounds"] = result["confounds"].to_records()
    ...
```

bthirion (approved):
LGTM, thx !

NicolasGensollen:
@jeromedockes I pushed a few commits to implement what we discussed. 
Let me know if this is more or less what you had in mind.

nilearn/datasets/atlas.py
```diff
@@ -259,7 +276,12 @@ def fetch_atlas_destrieux_2009(lateralized=True, data_dir=None, url=None,
     files_ = _fetch_files(data_dir, files, resume=resume,
                           verbose=verbose)
 
-    params = dict(maps=files_[1], labels=np.recfromcsv(files_[0]))
+    params = dict(maps=files_[1],
+                  labels=pd.read_csv(files_[0], index_col=0))
+
+    if legacy_format:
+        warnings.warn(LEGACY_FORMAT_MSG)
+        params['labels'] = params['labels'].to_records()
```
jeromedockes, line 284: this doesn't seem to return the same result as what it was before:
- I think we need to add `to_records(index=False)` or `index_col=0` otherwise the index generated by pandas gets added to the recarray:
```
[(0, 1, 'Superior frontal sulcus', 'DefaultB', 'DefaultA', 0.68999562, 0.18570919, 0.11688445)
 (1, 2, 'Fusiform gyrus', 'No network found', 'No network found', 0.84473252, 0.02102643, 0.13378859)
 (2, 3, 'Calcarine cortex posterior', 'VisCent', 'VisPeri', 0.60176782, 0.30206712, 0.09129732)
...
```
instead of 
```
[(1, 'Superior frontal sulcus', 'DefaultB', 'DefaultA', 0.68999562, 0.18570919, 0.11688445)
 (2, 'Fusiform gyrus', 'No network found', 'No network found', 0.84473252, 0.02102643, 0.13378859)
 (3, 'Calcarine cortex posterior', 'VisCent', 'VisPeri', 0.60176782, 0.30206712, 0.09129732)
```
- the column names are different (np.recfromcsv seems to convert to lower-case)
- the dtypes of columns that contain strings are different (object for numpy and "<U<some int>" for pandas), although that may not be a big issue

I think getting an identical result with `to_records` is harder than I anticipated, (sorry about that! :/ ), so it might actually be easier to have something like 
```
if legacy:
    labels = np.recfromcsv(labels.csv)
else:
    labels = pd.read_csv()
```

I believe this might apply to some other fetchers as well.
WDYT?

nilearn/datasets/atlas.py
```diff
@@ -750,8 +781,10 @@ def fetch_coords_power_2011():
     fdescr = _get_dataset_descr(dataset_name)
     package_directory = os.path.dirname(os.path.abspath(__file__))
     csv = os.path.join(package_directory, "data", "power_2011.csv")
-    params = dict(rois=np.recfromcsv(csv), description=fdescr)
-
+    params = dict(rois=pd.read_csv(csv), description=fdescr)
+    if legacy_format:
+        warnings.warn(LEGACY_FORMAT_MSG)
+        params['rois'] = params['rois'].to_records()
```
jeromedockes, line 787: Here also I think the pandas index gets added WRT to the result returned on the main branch

nilearn/datasets/atlas.py
```diff
@@ -1216,10 +1253,15 @@ def fetch_coords_dosenbach_2010(ordered_regions=True):
                   labels=labels,
                   networks=out_csv['network'], description=fdescr)
 
+    if legacy_format:
+        warnings.warn(LEGACY_FORMAT_MSG)
+        params['rois'] = params['rois'].to_records()
```
jeromedockes, line 1258: I think the pandas index gets added here as well

nilearn/datasets/func.py
```diff
@@ -740,26 +749,26 @@ def _is_valid_path(path, index, verbose):
...
+    if legacy_format:
+        warnings.warn(LEGACY_FORMAT_MSG)
+        csv_data = csv_data.to_records()
```
jeromedockes, line 765: we also have the index issue

nilearn/datasets/atlas.py
jeromedockes: another option is to lower-case column names and use `to_records(index=False)`

jeromedockes (commented):
Thanks @NicolasGensollen ! I think we are very close but it seems `pd.read_csv("file.csv").to_records()` is not equivalent to `np.recfromcsv("file.csv")`.
I think `df = pd.read_csv("file.csv"); df.rename(columns={c: c.lower() for c in df.columns}).to_records(index=False)` is close enough, though.
(I have only checked a few fetchers, not all of them)

if we have time we may want to improve a bit the tests as they did not catch this issue, for example we could check the dtype and shape of the results

nilearn/datasets/atlas.py
```diff
@@ -112,7 +123,10 @@ def fetch_atlas_difumo(dimension=64, resolution_mm=2, data_dir=None, resume=True
 
     # Download the zip file, first
     files_ = _fetch_files(data_dir, files, verbose=verbose)
-    labels = np.recfromcsv(files_[0])
+    labels = pd.read_csv(files_[0])
+    if legacy_format:
+        warnings.warn(_LEGACY_FORMAT_MSG)
+        labels = labels.to_records(index=False)
```
jeromedockes, line 129: great, thanks! we still need to convert column names to lower-case

NicolasGensollen: Thanks @jeromedockes ! I think the other fetchers are OK.
I will take another look to verify I didn't forget something though.

jeromedockes: Great, thanks! apart from that last detail the PR LGTM :)

NicolasGensollen:
@achamma723 Any chance you could rebase?
There is currently a conflict in `doc/whats_new.rst` because this file was recently deleted as part of #3049, the whats_new entry of this PR should be added to `doc/changes/latest.rst` instead.

NicolasGensollen:
Thanks @achamma723 !
@jeromedockes do you want to take another look at it before we merge?

nilearn/datasets/struct.py
```diff
@@ -735,18 +742,27 @@ def fetch_oasis_vbm(n_subjects=None, dartel_version=True, data_dir=None,
...
     csv_data = csv_data[subject_mask]
-
+    csv_data = csv_data.rename(
+        columns={c: c.lower() for c in csv_data.columns}
+    )
+    csv_data = csv_data.rename(
+        columns={c: c.replace("/", "") for c in csv_data.columns}
```
jeromedockes, line 758: doesn't matter but it could be done in one go with `c.lower().replace("/", "")`

doc/whats_new.rst
```diff
@@ -0,0 +1,2376 @@
+0.8.2.dev
```
jeromedockes, line 1: there is a large diff in the whats_new, did something go wrong with the rebase?

jeromedockes (commented):
thanks @achamma723  and @NicolasGensollen 
@NicolasGensollen if you had another look at the fetchers and all the changes in column names have been fixed I think we're good!

my only remaining question is about the whatsnew, not sure if there was an issue with the rebase due to #3049

doc/whats_new.rst
NicolasGensollen: Hmmm, yes the whats_new shouldn't even be there. The file was deleted by #3049 and separated in different files under `doc/changes`.

nilearn/datasets/struct.py
NicolasGensollen: Indeed, thanks! (see 55db108)

requirements-min.txt
```diff
@@ -1,8 +1,8 @@
-numpy==1.16
```
bthirion, line 1: Hm, this should not be here...

bthirion:
besides the requirement thing, this looks OK.

requirements-min.txt
NicolasGensollen: @achamma723 could you rebase to clean this? Thanks!

NicolasGensollen:
Thanks @achamma723 but it looks like there are still unrelated changes. I'm not sure how you rebase, but you  should be able to drop the unrelated commits.
If you cannot clean it, I'll merge the PR anyway since it is not a big deal, it just makes the diff more complicated than it should be...

achamma723:
> Thanks @achamma723 but it looks like there are still unrelated changes. I'm not sure how you rebase, but you should be able to drop the unrelated commits. If you cannot clean it, I'll merge the PR anyway since it is not a big deal, it just makes the diff more complicated than it should be...

Hello @NicolasGensollen , I'm pulling from main, rebasing and then pushing again. Should I drop something unrelated?

NicolasGensollen:
>Hello @NicolasGensollen , I'm pulling from main, rebasing and then pushing again. Should I drop something unrelated?

Hmm, it's weird that you get some commits from main in here. You can try to manually drop them when rebasing. I think the following commits shouldn't be there:

- 26de3fe
- 27e00f0
- 304e782

NicolasGensollen (approved):
LGTM, thx!

NicolasGensollen:
Thanks everyone! Merging...

PR #2831, opened 2021-05-08
BUG: Don't automatically sort `cut_coords` into ascending order

### LINKED ISSUE #1155: cut_coords force plotting of slices in ascending order
When specifying the coordinates of slices in plotting through `cut_coords`, they are automatically sorted in ascending order. This is a problem for people who wants a specific order, e.g. descending.

Craskermasker:
Closes #1155  .

- don't sort `display_ax_dict` by `coord` when positioning axes with `_locator` - within `nilearn.plotting.displays.BaseStackedSlicer`

However, I am not sure what downstream effects this will have, and if this is a desirable behaviour...

image

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -16,7 +16,8 @@ Fixes
   It used to return only the first attribute, and now returns as many attributes as design matrices.
 - Plotting functions that show a stack of slices from a 3D image (e.g. 
   :func:`nilearn.plotting.plot_stat_map`) will now plot the slices in the user 
-  specified order, rather than automatically sorting into ascending order. 
+  specified order, rather than automatically sorting into ascending order
+  (https://github.com/nilearn/nilearn/issues/1155). 
 
 Enhancements
 ------------


bthirion:
Thx !
@jeromedockes  Should we do a deprecation cycle before applying that change ?

jeromedockes:
bthirion ***@***.***> writes:

> Thx !
> @jeromedockes  Should we do a deprecation cycle before applying that change ?
>

I don't have a strong opinion; I think I would lean towards doing the
deprecation cycle because it's very easy in this case: in the line that
would be changed raise a deprecationwarning if the coordinates are not
sorted and in a couple of releases remove the sorting and the warning.

However if you would rather skip the deprecation cycle I think that's
ok too because:

- In the frequent case of specifying only a number of coordinates, they
  will still be sorted (twice actually), as they should:
  https://github.com/nilearn/nilearn/blob/dadc9b1d58dfb73d340e7a0382879d42349eee0c/nilearn/plotting/find_cuts.py#L379
- existing code is unlikely to fail or produce incorrect results after
  the change; the worst thing likely to happen is that slices appear out
  of order in some plots but the coordinates are also printed next to
  the slices so the plots are still understandable

bthirion:
I agree that this is a non-breaking change, and that we can avoid deprecation cycles for that one. Thx.

bthirion:
Maybe simply add a whatsnew entry. Thx.

bthirion (approved):
LGTM, thx !

PR #2836, opened 2021-05-12
Replace mktemp with mkstemp

### LINKED ISSUE #2832: test_with_globbing_patterns_with_single_image sometimes fails on Arm
`test_with_globbing_patterns_with_single_image` **sometimes** fails on Arm (aarch64, armv7 and armv6). It looks like a race condition because the `nii` file is not available.

Nilearn version: 0.7.1

### Expected behavior
The test should be reliable.

### Actual behavior
When the test fails, the log is:
```
[  304s] =================================== FAILURES ===================================
[  304s] ________________ test_with_globbing_patterns_with_single_image _________________
[  304s] [gw10] linux -- Python 3.9.4 /usr/bin/python3.9
[  304s] 
[  304s]     def test_with_globbing_patterns_with_single_image():
[  304s]         # With single image
[  304s]         data_4d = np.zeros((40, 40, 40, 3))
[  304s]         data_4d[20, 20, 20] = 1
[  304s]         img_4d = nibabel.Nifti1Image(data_4d, affine=np.eye(4))
[  304s]         multi_pca = MultiPCA(n_components=3)
[  304s]     
[  304s]         with write_tmp_imgs(img_4d, create_files=True, use_wildcards=True) as img:
[  304s]             input_image = _tmp_dir() + img
[  304s] >           multi_pca.fit(input_image)
[  304s] 
[  304s] nilearn/decomposition/tests/test_multi_pca.py:184: 
[  304s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[  304s] nilearn/decomposition/base.py:418: in fit
[  304s]     data = mask_and_reduce(
[  304s] nilearn/decomposition/base.py:172: in mask_and_reduce
[  304s]     data_list = Parallel(n_jobs=n_jobs)(
[  304s] /usr/lib/python3.9/site-packages/joblib/parallel.py:1044: in __call__
[  304s]     while self.dispatch_one_batch(iterator):
[  304s] /usr/lib/python3.9/site-packages/joblib/parallel.py:859: in dispatch_one_batch
[  304s]     self._dispatch(tasks)
[  304s] /usr/lib/python3.9/site-packages/joblib/parallel.py:777: in _dispatch
[  304s]     job = self._backend.apply_async(batch, callback=cb)
[  304s] /usr/lib/python3.9/site-packages/joblib/_parallel_backends.py:208: in apply_async
[  304s]     result = ImmediateResult(func)
[  304s] /usr/lib/python3.9/site-packages/joblib/_parallel_backends.py:572: in __init__
[  304s]     self.results = batch()
[  304s] /usr/lib/python3.9/site-packages/joblib/parallel.py:262: in __call__
[  304s]     return [func(*args, **kwargs)
[  304s] /usr/lib/python3.9/site-packages/joblib/parallel.py:262: in <listcomp>
[  304s]     return [func(*args, **kwargs)
[  304s] nilearn/decomposition/base.py:211: in _mask_and_reduce_single
[  304s]     this_data = masker.transform(img, confound)
[  304s] nilearn/input_data/multi_nifti_masker.py:344: in transform
[  304s]     return self.transform_single_imgs(imgs)
[  304s] nilearn/input_data/nifti_masker.py:443: in transform_single_imgs
[  304s]     data = self._cache(filter_and_mask,
[  304s] /usr/lib/python3.9/site-packages/joblib/memory.py:352: in __call__
[  304s]     return self.func(*args, **kwargs)
[  304s] nilearn/input_data/nifti_masker.py:56: in filter_and_mask
[  304s]     imgs = _utils.check_niimg(imgs, atleast_4d=True, ensure_ndim=4)
[  304s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[  304s] 
[  304s] niimg = '/tmp/nilearn_h4rbgbk1.nii', ensure_ndim = 4, atleast_4d = True
[  304s] dtype = None, return_iterator = False, wildcards = True
[  304s] 
[  304s]     def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
[  304s]                     return_iterator=False,
[  304s]                     wildcards=True):
[  304s]         """Check that niimg is a proper 3D/4D niimg. Turn filenames into objects.
[  304s]     
[  304s]         Parameters
[  304s]         ----------
[  304s]         niimg : Niimg-like object
[  304s]             See http://nilearn.github.io/manipulating_images/input_output.html
[  304s]             If niimg is a string, consider it as a path to Nifti image and
[  304s]             call nibabel.load on it. The '~' symbol is expanded to the user home
[  304s]             folder.
[  304s]             If it is an object, check if the affine attribute present and that
[  304s]             nilearn.image.get_data returns a result, raise TypeError otherwise.
[  304s]     
[  304s]         ensure_ndim : integer {3, 4}, optional
[  304s]             Indicate the dimensionality of the expected niimg. An
[  304s]             error is raised if the niimg is of another dimensionality.
[  304s]     
[  304s]         atleast_4d : boolean, optional
[  304s]             Indicates if a 3d image should be turned into a single-scan 4d niimg.
[  304s]             Default=False.
[  304s]     
[  304s]         dtype : {None, dtype, "auto"}, optional
[  304s]             Data type toward which the data should be converted. If "auto", the
[  304s]             data will be converted to int32 if dtype is discrete and float32 if it
[  304s]             is continuous. If None, data will not be converted to a new data type.
[  304s]             Default=None.
[  304s]     
[  304s]         return_iterator : boolean, optional
[  304s]             Returns an iterator on the content of the niimg file input.
[  304s]             Default=False.
[  304s]     
[  304s]         wildcards : boolean, optional
[  304s]             Use niimg as a regular expression to get a list of matching input
[  304s]             filenames.
[  304s]             If multiple files match, the returned list is sorted using an ascending
[  304s]             order.
[  304s]             If no file matches the regular expression, a ValueError exception is
[  304s]             raised.
[  304s]             Default=True.
[  304s]     
[  304s]         Returns
[  304s]         -------
[  304s]         result : 3D/4D Niimg-like object
[  304s]             Result can be nibabel.Nifti1Image or the input, as-is. It is guaranteed
[  304s]             that the returned object has an affine attribute and that its data can
[  304s]             be retrieved with nilearn.image.get_data.
[  304s]     
[  304s]         Notes
[  304s]         -----
[  304s]         In nilearn, special care has been taken to make image manipulation easy.
[  304s]         This method is a kind of pre-requisite for any data processing method in
[  304s]         nilearn because it checks if data have a correct format and loads them if
[  304s]         necessary.
[  304s]     
[  304s]         Its application is idempotent.
[  304s]     
[  304s]         See also
[  304s]         --------
[  304s]             _iter_check_niimg, check_niimg_3d, check_niimg_4d
[  304s]     
[  304s]         """
[  304s]         from ..image import new_img_like  # avoid circular imports
[  304s]     
[  304s]         if isinstance(niimg, str):
[  304s]             if wildcards and ni.EXPAND_PATH_WILDCARDS:
[  304s]                 # Ascending sorting + expand user path
[  304s]                 filenames = sorted(glob.glob(os.path.expanduser(niimg)))
[  304s]     
[  304s]                 # processing filenames matching globbing expression
[  304s]                 if len(filenames) >= 1 and glob.has_magic(niimg):
[  304s]                     niimg = filenames  # iterable case
[  304s]                 # niimg is an existing filename
[  304s]                 elif [niimg] == filenames:
[  304s]                     niimg = filenames[0]
[  304s]                 # No files found by glob
[  304s]                 elif glob.has_magic(niimg):
[  304s]                     # No files matching the glob expression, warn the user
[  304s]                     message = ("No files matching the entered niimg expression: "
[  304s]                                 "'%s'.\n You may have left wildcards usage "
[  304s]                                "activated: please set the global constant "
[  304s]                                "'nilearn.EXPAND_PATH_WILDCARDS' to False to "
[  304s]                                "deactivate this behavior.") % niimg
[  304s]                     raise ValueError(message)
[  304s]                 else:
[  304s] >                   raise ValueError("File not found: '%s'" % niimg)
[  304s] E                   ValueError: File not found: '/tmp/nilearn_h4rbgbk1.nii'
[  304s] 
[  304s] nilearn/_utils/niimg_conversions.py:271: ValueError
```

Run tests on aarch64. 
Tested with openSUSE Tumbleweed aarch64 (and armv7, armv6).

NicolasGensollen:
Fixes #2832 

This PR aims at replacing `mktemp` with `mkstemp` as suggested by @jeromedockes in #2832 

I think we also need to test on ARM architectures to detect these kinds of bugs. 
For the moment, this would mean bringing travis back into the CI, and only run ARM tests to save credits (this is what scikit-learn is doing currently for instance). I will open a separate issue on this as this would require some work (especially to avoid systematic building of wheels...).

### DIFF
--- a/nilearn/_utils/testing.py
+++ b/nilearn/_utils/testing.py
@@ -179,10 +179,7 @@ def write_tmp_imgs(*imgs, **kwargs):
         finally:
             # Ensure all created files are removed
             for filename in filenames:
-                try:
-                    os.remove(filename)
-                except OSError:
-                    pass
+                os.remove(filename)
     else:  # No-op
         if len(imgs) == 1:
             yield imgs[0]
--- a/nilearn/_utils/testing.py
+++ b/nilearn/_utils/testing.py
@@ -152,7 +152,7 @@ def write_tmp_imgs(*imgs, **kwargs):
     create_files = kwargs.get("create_files", True)
     use_wildcards = kwargs.get("use_wildcards", False)
 
-    prefix = "nilearn_{}_".format(os.getpid())
+    prefix = "nilearn_"
     suffix = ".nii"
 
     if create_files:
@@ -161,23 +161,25 @@ def write_tmp_imgs(*imgs, **kwargs):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", RuntimeWarning)
                 for img in imgs:
-                    fd, path = tempfile.mkstemp(prefix=prefix,
-                                                suffix=suffix,
-                                                dir=None)
-                    filenames.append(path)
-                    img.to_filename(path)
+                    fd, filename = tempfile.mkstemp(prefix=prefix,
+                                                    suffix=suffix,
+                                                    dir=None)
+                    filenames.append((fd, filename))
+                    img.to_filename(filename)
                     del img
 
                 if use_wildcards:
                     yield prefix + "*" + suffix
                 else:
                     if len(imgs) == 1:
-                        yield filenames[0]
+                        yield filenames[0][1]
                     else:
-                        yield filenames
+                        yield [f[1] for f in filenames]
         finally:
             # Ensure all created files are removed
-            for filename in filenames:
+            for temp in filenames:
+                fd, filename = temp
+                os.close(fd)
                 os.remove(filename)
     else:  # No-op
         if len(imgs) == 1:
--- a/nilearn/_utils/testing.py
+++ b/nilearn/_utils/testing.py
@@ -177,9 +177,20 @@ def write_tmp_imgs(*imgs, **kwargs):
                     else:
                         yield filenames
         finally:
+            failures = []
             # Ensure all created files are removed
             for filename in filenames:
-                os.remove(filename)
+                try:
+                    os.remove(filename)
+                except FileNotFoundError:
+                    # ok, file already removed
+                    pass
+                except OSError as e:
+                    # problem eg permission, or open file descriptor
+                    failures.append(e)
+            if failures:
+                raise ValueError(functools.reduce(lambda x,y: x+'\n'+y,
+                                    [str(e) for e in failures]))
     else:  # No-op
         if len(imgs) == 1:
             yield imgs[0]
--- a/nilearn/_utils/testing.py
+++ b/nilearn/_utils/testing.py
@@ -189,8 +189,11 @@ def write_tmp_imgs(*imgs, **kwargs):
                     # problem eg permission, or open file descriptor
                     failures.append(e)
             if failures:
-                raise ValueError(functools.reduce(lambda x,y: x+'\n'+y,
-                                    [str(e) for e in failures]))
+                raise OSError(
+                    "The following files could not be removed:\n".format(
+                        "\n".join(str(e) for e in failures)
+                    )
+                )
     else:  # No-op
         if len(imgs) == 1:
             yield imgs[0]
--- a/nilearn/_utils/testing.py
+++ b/nilearn/_utils/testing.py
@@ -164,7 +164,8 @@ def write_tmp_imgs(*imgs, **kwargs):
                     fd, filename = tempfile.mkstemp(prefix=prefix,
                                                     suffix=suffix,
                                                     dir=None)
-                    filenames.append((fd, filename))
+                    os.close(fd)
+                    filenames.append(filename)
                     img.to_filename(filename)
                     del img
 
@@ -172,15 +173,16 @@ def write_tmp_imgs(*imgs, **kwargs):
                     yield prefix + "*" + suffix
                 else:
                     if len(imgs) == 1:
-                        yield filenames[0][1]
+                        yield filenames[0]
                     else:
-                        yield [f[1] for f in filenames]
+                        yield filenames
         finally:
             # Ensure all created files are removed
-            for temp in filenames:
-                fd, filename = temp
-                os.close(fd)
-                os.remove(filename)
+            for filename in filenames:
+                try:
+                    os.remove(filename)
+                except OSError:
+                    pass
     else:  # No-op
         if len(imgs) == 1:
             yield imgs[0]
--- a/nilearn/_utils/testing.py
+++ b/nilearn/_utils/testing.py
@@ -152,7 +152,7 @@ def write_tmp_imgs(*imgs, **kwargs):
     create_files = kwargs.get("create_files", True)
     use_wildcards = kwargs.get("use_wildcards", False)
 
-    prefix = "nilearn_"
+    prefix = "nilearn_{}_".format(os.getpid())
     suffix = ".nii"
 
     if create_files:
--- a/nilearn/_utils/testing.py
+++ b/nilearn/_utils/testing.py
@@ -161,11 +161,11 @@ def write_tmp_imgs(*imgs, **kwargs):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", RuntimeWarning)
                 for img in imgs:
-                    filename = tempfile.mktemp(prefix=prefix,
-                                               suffix=suffix,
-                                               dir=None)
-                    filenames.append(filename)
-                    img.to_filename(filename)
+                    fd, path = tempfile.mkstemp(prefix=prefix,
+                                                suffix=suffix,
+                                                dir=None)
+                    filenames.append(path)
+                    img.to_filename(path)
                     del img
 
                 if use_wildcards:
--- a/nilearn/surface/tests/test_surface.py
+++ b/nilearn/surface/tests/test_surface.py
@@ -74,9 +74,11 @@ def test_load_surf_data_array():
     assert_array_equal(load_surf_data(data_squeeze), np.zeros((20, 3)))
 
 
-def test_load_surf_data_file_nii_gii():
+def test_load_surf_data_file_nii_gii(tmp_path):
     # test loading of fake data from gifti file
-    fd_gii, filename_gii = tempfile.mkstemp(suffix='.gii')
+    fd_gii, filename_gii = tempfile.mkstemp(suffix='.gii',
+                                            dir=tmp_path)
+    os.close(fd_gii)
     if LooseVersion(nb.__version__) > LooseVersion('2.0.2'):
         darray = gifti.GiftiDataArray(data=np.zeros((20, )))
     else:
@@ -87,30 +89,32 @@ def test_load_surf_data_file_nii_gii():
     gii = gifti.GiftiImage(darrays=[darray])
     gifti.write(gii, filename_gii)
     assert_array_equal(load_surf_data(filename_gii), np.zeros((20, )))
-    os.close(fd_gii)
     os.remove(filename_gii)
 
     # test loading of data from empty gifti file
-    fd_empty, filename_gii_empty = tempfile.mkstemp(suffix='.gii')
+    fd_empty, filename_gii_empty = tempfile.mkstemp(suffix='.gii',
+                                                    dir=tmp_path)
+    os.close(fd_empty)
     gii_empty = gifti.GiftiImage()
     gifti.write(gii_empty, filename_gii_empty)
     with pytest.raises(ValueError,
                        match='must contain at least one data array'
                        ):
         load_surf_data(filename_gii_empty)
-    os.close(fd_empty)
     os.remove(filename_gii_empty)
 
     # test loading of fake data from nifti file
-    fd_gii2, filename_nii = tempfile.mkstemp(suffix='.nii')
-    fd_niigz, filename_niigz = tempfile.mkstemp(suffix='.nii.gz')
+    fd_gii2, filename_nii = tempfile.mkstemp(suffix='.nii',
+                                             dir=tmp_path)
+    os.close(fd_gii2)
+    fd_niigz, filename_niigz = tempfile.mkstemp(suffix='.nii.gz',
+                                                dir=tmp_path)
+    os.close(fd_niigz)
     nii = nb.Nifti1Image(np.zeros((20, )), affine=None)
     nb.save(nii, filename_nii)
     nb.save(nii, filename_niigz)
     assert_array_equal(load_surf_data(filename_nii), np.zeros((20, )))
     assert_array_equal(load_surf_data(filename_niigz), np.zeros((20, )))
-    os.close(fd_gii2)
-    os.close(fd_niigz)
     os.remove(filename_nii)
     os.remove(filename_niigz)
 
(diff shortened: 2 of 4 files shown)


nilearn/_utils/testing.py
```diff
@@ -161,11 +161,11 @@ def write_tmp_imgs(*imgs, **kwargs):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", RuntimeWarning)
                 for img in imgs:
-                    filename = tempfile.mktemp(prefix=prefix,
-                                               suffix=suffix,
-                                               dir=None)
-                    filenames.append(filename)
-                    img.to_filename(filename)
+                    fd, path = tempfile.mkstemp(prefix=prefix,
```
jeromedockes, line 164: you need to `os.close(fd)`

nilearn/_utils/testing.py
```diff
@@ -152,7 +152,7 @@ def write_tmp_imgs(*imgs, **kwargs):
     create_files = kwargs.get("create_files", True)
     use_wildcards = kwargs.get("use_wildcards", False)
 
-    prefix = "nilearn_"
+    prefix = "nilearn_{}_".format(os.getpid())
```
jeromedockes, line 155: you don't need to add the pid, mkstemp will take care of creating a filename that doesn't exist

nilearn/_utils/testing.py
jeromedockes: btw this function has another issue: if removing one image fails (for example if it has been removed by the caller) the other ones are not removed

nilearn/_utils/testing.py
NicolasGensollen: Thanks! Yes I got confused and that was a stupid attempt... :sweat: :smile:

nilearn/_utils/testing.py
```diff
@@ -161,23 +161,25 @@ def write_tmp_imgs(*imgs, **kwargs):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", RuntimeWarning)
                 for img in imgs:
-                    filename = tempfile.mktemp(prefix=prefix,
-                                               suffix=suffix,
-                                               dir=None)
-                    filenames.append(filename)
+                    fd, filename = tempfile.mkstemp(prefix=prefix,
```
jeromedockes, line 164: sorry if my previous comment was unclear, but you need to close the file right away and return the filename only
otherwise you yield the filename but retain an open file handle and windows won't allow opening the file for writing or removing it.
(the file is not removed when you close it)
you don't need to store the file descriptor once you have closed the file

nilearn/_utils/testing.py
```diff
@@ -161,23 +161,25 @@ def write_tmp_imgs(*imgs, **kwargs):
...
+                                                    suffix=suffix,
+                                                    dir=None)
+                    filenames.append((fd, filename))
```
jeromedockes, line 167:
```suggestion
                    os.close(fd)
                    filenames.append(filename)
```

nilearn/_utils/testing.py
```diff
@@ -161,23 +161,25 @@ def write_tmp_imgs(*imgs, **kwargs):
...
 
                 if use_wildcards:
                     yield prefix + "*" + suffix
                 else:
                     if len(imgs) == 1:
-                        yield filenames[0]
+                        yield filenames[0][1]
                     else:
-                        yield filenames
+                        yield [f[1] for f in filenames]
```
jeromedockes, line 177:
```suggestion
                        yield filenames
```

nilearn/_utils/testing.py
```diff
@@ -161,23 +161,25 @@ def write_tmp_imgs(*imgs, **kwargs):
...
         finally:
             # Ensure all created files are removed
-            for filename in filenames:
+            for temp in filenames:
+                fd, filename = temp
+                os.close(fd)
                 os.remove(filename)
```
jeromedockes, line 183:
```suggestion
            for filename in filenames:
                try:
                    os.remove(filename)
                except OSError:
                    pass
```

nilearn/_utils/testing.py
jeromedockes: not at all! these things can get confusing indeed :smile:

jeromedockes:
> I think we also need to test on ARM architectures to detect these kinds of bugs.
For the moment, this would mean bringing travis back into the CI, and only run ARM tests to save credits (this is what scikit-learn is doing currently for instance). I will open a separate issue on this as this would require some work (especially to avoid systematic building of wheels...).

good idea! in general it's nice to roughly imitate what scikit-learn does so that when a problem pops up we can just copy their solution :smile: 

+1 for doing it in another pr

nilearn/_utils/testing.py
```diff
@@ -178,7 +179,10 @@ def write_tmp_imgs(*imgs, **kwargs):
         finally:
             # Ensure all created files are removed
             for filename in filenames:
-                os.remove(filename)
+                try:
+                    os.remove(filename)
+                except OSError:
```
bthirion, line 184: Why do we need that ?

nilearn/surface/tests/test_surface.py
```diff
@@ -76,7 +76,7 @@ def test_load_surf_data_array():
 
 def test_load_surf_data_file_nii_gii():
     # test loading of fake data from gifti file
-    filename_gii = tempfile.mktemp(suffix='.gii')
+    fd_gii, filename_gii = tempfile.mkstemp(suffix='.gii')
```
jeromedockes, line 79: in all the uses of `mkstemp` below I would suggest closing the file right after creating it.

also, this and most of the nilearn tests we see here leak resources: if the test fails the temporary files are not removed.

it would be better to create these files in pytest's `tmp_path` and do not bother to remove them (it is not removed by default so that it can be inspected when investigating failures, but at least all temp files are in one place in the `/tmp/pytest-of-$USER` and it is easy to remove it or configure it to be removed automatically)


an alternative would be rely on a context manager to remove them (not `tempfile.NamedTemporaryFile`, because it can either remove the file as soon as it is closed or never remove it), but I prefer the pytest tmp_path solution.

I guess it should be done in a separate pr to avoid slowing down this one; if you agree I'll open an issue

nilearn/_utils/testing.py
jeromedockes: so that the other files are still removed if removing this one fails (eg `os.remove` raises an exception if the file doesn't exist, and the test using this function may have removed the file; it also fails if there is an open file handle for that file on windows)

NicolasGensollen:
Thanks for the reviews @jeromedockes !

>I guess it should be done in a separate pr to avoid slowing down this one, except if you think that doing all in one go will actually be easier

I did it on this PR since it is not super long for now.

>it would be better to create these files in pytest's tmp_path and do not bother to remove them

I'm not sure to understand why we wouldn't want to remove them. I'm afraid this could potentially cause some failures. For example if a test creates some of these files and later does something different using wildcards. It would then grab unexpectedly left over garbage files. Does this make sense?

I also removed the try except in `nilearn/_utils/testing.py`. I don't like the fact that this could silently fail when the files cannot be deleted for some reason. Maybe we should aim for an intermediate solution which will raise an error after handling all the files in the list if one or more are missing. Maybe something like:

```python
missing = False
for filename in filenames:
    try:
        os.remove(filename)
    except OSError:
        missing = True
        pass
if missing:
    raise FileNotFoundError("msg...")
```

WDYT?

nilearn/surface/tests/test_surface.py
```diff
@@ -364,17 +391,20 @@ def test_load_surf_mesh_file_glob():
     assert len(load_surf_mesh(fname1)) == 2
     assert_array_almost_equal(load_surf_mesh(fname1)[0], mesh[0])
     assert_array_almost_equal(load_surf_mesh(fname1)[1], mesh[1])
-
     os.remove(fname1)
     os.remove(fname2)
```
jeromedockes, line 395: > I'm not sure to understand why we wouldn't want to remove them. I'm afraid this could potentially cause some failures. For example if a test creates some of these files and later does something different using wildcards. It would then grab unexpectedly left over garbage files. Does this make sense?

the issue is removing them in statements that may never be reached (as opposed
to removing them in `__exit__` methods or `finally` blocks). if we look at this
one for example,

- if removing the temporary file is necessary, there is a bug here. if control
  leaves the function somewhere else due to an exception or a return statement,
  the file is never removed. that was the case before you changed this function
  to use `tmp_path`.
- if removing the file is not necessary, removing it makes it less obvious that
  in any case it is actually being dealt with somewhere else (by the pytest
  fixture, in this case). this is the current situation now that you fixed the
  bug. note that many other tests in the library create files in `tmp_path` and
  never remove them -- dealing with them is pytest's job -- and if we remove all
  the `os.remove` calls in this file the tests still work.

but it's also not an error to remove them so it's not that important; if you
would rather remove them let's keep it like this! and it's not really related with
replacing `mktemp` with `mkstemp` anyway. (btw now that the files are created in
an empty scratch directory the need for `mkstemp` is smaller)

(in a similar way there are some `del` in this file for which the need is not clear)

jeromedockes:
> I also removed the try except in `nilearn/_utils/testing.py`. I don't like the fact that this could silently fail when the files cannot be deleted for some reason. Maybe we should aim for an intermediate solution which will raise an error after handling all the files in the list if one or more are missing. Maybe something like:
> 
> ```python
> missing = False
> for filename in filenames:
>     try:
>         os.remove(filename)
>     except OSError:
>         missing = True
>         pass
> if missing:
>     raise FileNotFoundError("msg...")
> ```
> 
> WDYT?

sure, that makes sense! although I wouldn't consider the file not existing an error: the documentation of this function doesn't say anywhere that the caller is not allowed to remove the file. so maybe the loop body could be roughly similar to 

```python
try:
    os.remove(filename)
except FileNotFoundError:
    # ok, file already removed
    pass
except OSError as e:
    # problem eg permission, or open file descriptor
    failures.append(e)
```

note that raising a new error does have a drawback: if there previously was a live exception it gets buried in the new one's `__context__` and if someone wanted to catch it it is unlikely they will be able to. it would be possible to distinguish this situation by using `except` and `else` rather than `finally` but I think that would be overkill.
also you will only be able to raise one exception even if several files fail to be removed so maybe `msg` could contain the list of files or something like that?

jeromedockes (approved):
this pr already achieves its goal so if you prefer to leave aside the remaining comments it already LGTM!

nilearn/_utils/testing.py
```diff
@@ -177,9 +177,20 @@ def write_tmp_imgs(*imgs, **kwargs):
...
+                    os.remove(filename)
+                except FileNotFoundError:
+                    # ok, file already removed
+                    pass
+                except OSError as e:
+                    # problem eg permission, or open file descriptor
+                    failures.append(e)
+            if failures:
+                raise ValueError(functools.reduce(lambda x,y: x+'\n'+y,
+                                    [str(e) for e in failures]))
```
jeromedockes, line 193: maybe still make it an oserror? eg
```python
if failures:
    raise OSError(
        "The following files could not be removed:\n{}".format(
            "\n".join(str(e) for e in failures)
        )
    )
```

bthirion (approved):

NicolasGensollen:
Thanks for the reviews @jeromedockes @bthirion !
Merging...

Issue #2837, opened 2021-05-12
doc not building with sphinx 4.0.1


jeromedockes:
`make html-noplot` fails since the latest release of sphinx on monday

see for example [this build](https://app.circleci.com/pipelines/github/nilearn/nilearn/2034/workflows/2bd1d45d-b513-41d2-a941-c17feab7e8e2/jobs/7212)

emdupre:
I don't yet see an issue flagging this issue on `sphinx-copybutton` (https://github.com/executablebooks/sphinx-copybutton). It seems like an upstream problem there rather than a nilearn issue. Should we open an issue there directly ?

RaphaelMeudec:
Or maybe you can try switching to `add_js_file` which has replaced `add_javascript` according to the doc : https://www.sphinx-doc.org/en/master/extdev/appapi.html?highlight=add_js_file#sphinx.application.Sphinx.add_js_file

emdupre:
Thanks, @RaphaelMeudec ! That seems like a good solution.

In fact, despite the name, I realized that `sphinx_copybutton` and our `copybutton.js` share almost no overlap. `sphinx_copybutton` adds the ability to copy code blocks to the user's clipboard; for example, [here](https://7214-1235740-gh.circle-artifacts.com/0/dev/auto_examples/01_plotting/plot_visualize_megatrawls_netmats.html#fetching-the-megatrawls-network-matrices).

`copybutton.js`, meanwhile, only acts on code blocks starting with a `>>>` to hide or show the prompt and code output, like [here](http://nilearn.github.io/connectivity/functional_connectomes.html#a-functional-connectome-extracting-coordinates-of-regions). So they're largely non-overlapping in functionality, though I'm not sure what their interaction would be for those cells that are covered by `copybutton.js`.

Either way, I think the best patch to close this issue is just amending the call (exactly as you suggest). I still find `sphinx_copybutton` useful, so I won't close #2838 , but maybe we can continue discussion on its implementation there !

PR #2838, opened 2021-05-12
[INFRA] Adopt sphinx-copybutton

emdupre:
Related #2837 .

- Adds sphinx_copybutton as a dependency
- Removes copybutton.js from the library


jeromedockes:
thanks! weirdly enough `circleci_dependencies.sh` doesn't seem to rely on `requirements-build-doc.txt` -- see [here](https://github.com/nilearn/nilearn/blob/86b12ee8f81198932e11d2f4d6d4c2ee3173696c/tools/circleci_dependencies.sh#L7)

so you would probably need to add `sphinx-copybutton` there as well, then we could open a separate pr to make sure these dependencies are specified in one place only

jeromedockes (approved):
thanks @emdupre ! AFAICT it is ready to be merged?

emdupre:
I don't know that I'd merge this yet .... It looks like `sphinx-copybutton` overrides `copybutton.js`, at least [here](https://7216-1235740-gh.circle-artifacts.com/0/dev/connectivity/functional_connectomes.html). I could try triggering a full build to confirm this behavior.

But I can open a separate PR to just patch the one `app.add_js_file` line ?

tsalo:
Per this week's call, it sounds like `copybutton.js` can just be removed. Is that correct?

emdupre:
Yes, I just need to revert 524a4de50ea101fe42867aa9d4ec1e3ae3fbe78f and look through the artifacts one more time !

emdupre:
I think this is ready for (re)review !

bthirion (approved):
Thx !

emdupre:
@NicolasGensollen are you OK if I merge this ?

NicolasGensollen (approved):
>@NicolasGensollen are you OK if I merge this ?

Yes, looks good to me, thanks a lot @emdupre !

PR #2845, opened 2021-05-19
Add issue assignment action

### LINKED ISSUE #2794: Issue assignment bot?
For the purposes of dev-days, the bandwidth of core devs to assign issues to un-privileged contributors may be stressed.

We should look for a bot that can be installed to allow anybody to assign themselves, e.g., I might comment on an issue:

```
@assign-bot please assign @effigies
```

Something quick and dirty will probably suffice unless there's abuse, which seems unlikely.

NicolasGensollen:
Fixes #2794 

This PR proposes to use a github action to assign issues automatically when commenting "take" on them. It also removes the label "help wanted" in case it was associated to the issue.
This comes from scikit-learn so I can add a comment citing the source at the beginning of the file if we decide to use it.

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -1,3 +1,6 @@
+# Source: Scikit-learn
+# See: https://github.com/scikit-learn/scikit-learn/blob/main/.github/workflows/assign.yml
+
 name: Assign
 on:
   issue_comment:
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -1,0 +1,18 @@
+name: Assign
+on:
+  issue_comment:
+    types: created
+
+jobs:
+  one:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


jeromedockes:
great! +1 for adding a comment pointing to the scikit-learn file it was copied from. you might also want to add a sentence explaining this in the contributing guide?

bthirion:
Sorry, is there any way to check the behavior ?

NicolasGensollen:
@bthirion I don't believe there is a way to check before merging here. 
That being said, it is working on scikit-learn, and I tried it in a test repo and it works as expected. 
Finally, the risk is minimal since it is a self-contained file we can easily delete in case it doesn't work or we are not happy with this behavior.

jeromedockes:
that makes sense, then let's try it out! +1 for merging

bthirion (approved):

PR #2849, opened 2021-05-20
reduce default surf plot width and add the axes to the figure

### LINKED ISSUE #2614: Axes3D no longer adds itself to figure
Upcoming break in Matplotlib `3.4.0`, see: [https://github.com/matplotlib/matplotlib/pull/18564](https://github.com/matplotlib/matplotlib/pull/18564)

This is causing 3 tests to fail on Travis with matplotlib-dev: 

```
FAILED plotting/tests/test_surf_plotting.py::test_plot_surf_stat_map - assert...
FAILED plotting/tests/test_surf_plotting.py::test_plot_surf_contours - IndexE...
FAILED plotting/tests/test_surf_plotting.py::test_plot_surf_contours_error - ...
===== 3 failed, 829 passed, 1 skipped, 1800 warnings in 721.48s
```

`ax = Axes3d(fig)` will need to be explicitly added to the figure with `fig.add_axes(ax)`

### LINKED ISSUE #2847: `plot_surf` creates a figure that is too wide
by default `plot_surf`, `plot_surf_stat_map`, etc create a figure of height, width = 4, 6 (or 4, 6.7 with a colorbar), which is very wide and that results in a lot of white space on the left of the figure, see [here](https://nilearn.github.io/auto_examples/01_plotting/plot_3d_map_to_surface_projection.html#plot-the-result) for example, or with a colored figure and axis to make it easier to see the space occupied by the axis:

fig

AFAICT all that needs to change would be [this default size](https://github.com/nilearn/nilearn/blob/14d23f2a3e8416330a9f4db3110509799d32dc02/nilearn/plotting/surf_plotting.py#L161), but I guess it may have been chosen like this because it is a better choice for some meshes or point of views? it seems to have been set in #2643 @NicolasGensollen  do you remember?

jeromedockes:
closes #2847, closes #2614

### DIFF
--- a/nilearn/plotting/surf_plotting.py
+++ b/nilearn/plotting/surf_plotting.py
@@ -161,7 +161,7 @@ def plot_surf(surf_mesh, surf_map=None, bg_map=None,
     nilearn.surface.vol_to_surf : For info on the generation of surfaces.
 
     """
-    _default_figsize = [6, 4]
+    _default_figsize = [4, 4]
 
     # load mesh and derive axes limits
     mesh = load_surf_mesh(surf_mesh)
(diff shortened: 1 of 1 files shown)


NicolasGensollen (approved):
LGTM! thanks @jeromedockes !

PR #2848, opened 2021-05-20
[MAINT] Clean CI tools

NicolasGensollen:
Closes #2746 

This PR proposes to do some cleaning of the CI utilities:

- Reorganize CI scripts in folder `build_tools` rather than having things in both `tools` and `continuous_integration`. If a script is dedicated to a specific platform, it is then placed in the corresponding sub-folder. For example, all scripts for CircleCI are in `build_tools/circle/`, for Github actions `build_tools/github/`...
- Rename files to make their name shorter and/or clarify their purpose
- Add some documentation for configuring how CircleCI will build the doc through commit messages (we can include that later in the contributing docs once Thomas's PR gets merged)
- Remove unused scripts and files (previously used for caching on CircleCI)
- Update flake8_diff script and make it run through GitHub Actions (currently runs with the min dependency job, but we can change that I needed)
- Run `make test-doc` which was not executed on the CI since we dropped testing on Travis.
- Create a `maint_tools` folder to hold other maintenance scripts

PR #2854, opened 2021-05-27
ARM arch testing

### LINKED ISSUE #2840: Tests on ARM architecture
We should consider adding ARM support to the CI. This would help us catch related issues (see #2832 for example).
For the moment, I think travis is still the best option for this. 
Since we have limited credits, I think we should only run ARM tests (this is what scikit-learn is doing currently for instance).

NicolasGensollen:
Fixes #2840 

This PR proposes to add ARM testing in the CI pipeline. 
Since it relies on travis to run the tests and we have limited credits, the workflow is only triggered whenever a commit message contains the tag "[arm64]". 
Since it is running in about 15-20min, I'm wondering whether we could afford running it systematically on master though... :thinking: 

NOTE: It looks like the only test failing is `test_resample_img_segmentation_fault`:

https://github.com/nilearn/nilearn/blob/7142c5c9199e13c67111e7a44c58c9169259908b/nilearn/image/tests/test_resampling.py#L793-L815

I'm not sure what is happening since the worker just dies at the resampling step. I tried changing the fourth dimension value to very small values (in which case the test passes) and to 1024 (in which case it still fails...). 
For now, I added a skip condition for this test when running under travis with ARM64 architecture, but I'm open to suggestions.


bthirion:
The error does not seem to be related to Python 3.5...

NicolasGensollen:
>The error does not seem to be related to Python 3.5...

No it's not, it is a PEP8 violation but it is running as part of the python 3.5 workflow. 
I realized today that some of the changes I made in #2848 result in confusing outputs in the Github actions board. I'm working on fixing/improving this in #2856

jeromedockes:
@NicolasGensollen can we merge this one?

NicolasGensollen:
@jeromedockes I believe we can merge it. For now, you have to explicitly request ARM tests but putting "[arm64]" in your commit message. 
Not sure whether we want to have them run systematically on master because of credit concerns..

jeromedockes:
> @jeromedockes I believe we can merge it. For now, you have to explicitly request ARM tests but putting "[arm64]" in your commit message. 
> Not sure whether we want to have them run systematically on master because of credit concerns..

that makes sense, and the need for it doesn't arise that often. I say
let's merge it as it is, and if we ever want to start running it with
every push that will be easy to change. does that make sense?

NicolasGensollen:
>does that make sense?

Absolutely!

bthirion:
Thx !

PR #2856, opened 2021-05-28
[MAINT] Run tests, doctests, and flake8 diff in their own steps

NicolasGensollen:
Following PR #2848 this PR makes sure tests, doctests, and flake8 checks are run in separate steps of the workflows instead of running them in a single step and using environment variables to configure which scripts are run.

I realized that if you run, for example, tests and doctests in the same step and tests fail while doctests pass, you will get a :heavy_check_mark:  instead of a :x: next to the github action, which is very confusing... See for example [here](https://github.com/nilearn/nilearn/pull/2815/checks?check_run_id=2689751187)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -33,8 +33,6 @@ jobs:
         shell: bash
     env:
         MIN_REQUIREMENTS: true
-        FLAKE8: true
-        TEST_DOC: false
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/testing.yml
```diff
@@ -48,9 +48,6 @@ jobs:
       - shell: bash -l {0}
         run: ./build_tools/github/install.sh
         name: 'Install nilearn'
-      - shell: bash -l {0}
-        run: ./build_tools/github/test_docs.sh
-        name: 'Run doctests'
```
bthirion: Hi,
For me to understand, what is the reason ? It fails ?

bthirion (approved):

.github/workflows/testing.yml
NicolasGensollen: @bthirion Yes doctest fail with Python 3.5 and minimum requirements because of a few scikit-learn related imports.
For example:

```python
Python 3.5.6 |Anaconda, Inc.| (default, Aug 26 2018, 21:41:56) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sklearn
>>> sklearn.__version__
'0.19.0'
>>> from sklearn.covariance import GraphicalLassoCV
ImportError: cannot import name 'GraphicalLassoCV'
```

NicolasGensollen:
Merging, Thanks!

PR #2858, opened 2021-05-30
Move `sample_mask` to `transform` method in maskers, handle `sample_mask` in `signal.clean`

### LINKED ISSUE #1012: Implement sample mask in signal cleaning and the various maskers
The goal of this issue is partly to implement/support scrubbing of time series.

This issue depends on #1011

htwangtw:
Closes #1012. Replace #2857 due to formatting was messed up. 

- In `base_masker`, move parameter `sample_mask` from initialization to `transform` method, and all the Masker modules.
- Handle `sample_masks` with `signal.clean`
- Move input validation related to `runs` to a separate private function from `_process_runs`  and `_sanitize_inputs` to `_sanitize_runs` 

While implementing the discussed changes, I found how sample_mask should be handled when there are multiple runs more complicated than anticipated.
Here's a summary of the current implementation:  
* accept any input valid for indexing NumPy array as `sample_mask`
* for multiple runs, user must pass a list of sample masks to `sample_mask` (a set of index per run).

I have tried to implement `sample_mask` to match `confounds` for multiple runs, which allows sample_mask to be a 1D array when runs are provided. User needs to correct the index for multiple runs to fit into a 1D array. I think it's too error-prone. This makes me wonder, should we do the same for confounds? Or we want to rethink how multiple runs are handled?

This PR has a lot of difficult things to handle. If there's any redundant checks in signal.clean, or any better idea of handling the `sample_mask`, `runs`, and `confounds` please let me know.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -58,7 +58,7 @@ Enhancements
   deprecates attribute `sessions` in 0.9.0. Match the relevant change in
   :func:`nilearn.signal.clean`.
 - Moves parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
-  to method `transform` in base class :class:`nilearn.input_data.base_masker`.
+  to method `transform` in base class :class:`nilearn.input_data.BaseMasker`.
 
 .. _v0.7.1:
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -5,11 +5,11 @@ NEW
 ---
 
 - :func:`nilearn.signal.clean` accepts new parameter `sample_mask`.
-  Masks the niimgs along time/fourth dimension to perform scrubbing (remove
-  volumes with high motion) and/or non-steady-state volumes. This masking step
-  is applied before signal cleaning.
   shape: (number of scans - number of volumes removed, )
-- All inheritent classes of :class:`nilearn.input_data.base_masker` can use
+  Masks the niimgs along time/fourth dimension to perform scrubbing (remove
+  volumes with high motion) and/or non-steady-state volumes. Masking is applied
+  before signal cleaning.
+- All inherent classes of :class:`nilearn.input_data.base_masker` can use
   parameter `sample_mask` for sub-sample masking.
 
 Fixes
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -8,7 +8,7 @@ NEW
   volumes with high motion) and/or non-steady-state volumes. This masking step
   is applied before signal cleaning.
   shape: (number of scans - number of volumes removed, )
-  - Moving parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
+- Moving parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
   to method `transform` in base class :class:`nilearn.input_data.base_masker`.
   All inheritent classes of :class:`nilearn.input_data.base_masker` can use
   parameter `sample_mask` for sub-sample masking.
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -54,7 +54,7 @@ Enhancements
   'stratified' (https://github.com/nilearn/nilearn/pull/2826/).
 - :class:`nilearn.glm.first_level.run_glm` now allows auto regressive noise
   models of order greater than one.
-- :class:`nilearn.input_data.NiftiMasker` replace `sessions` with `runs` and
+- :class:`nilearn.input_data.NiftiMasker` replaces `sessions` with `runs` and
   deprecates attribute `sessions` in 0.9.0. Match the relevant change in
   :func:`nilearn.signal.clean`.
 - Moves parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -9,7 +9,7 @@ NEW
   Masks the niimgs along time/fourth dimension to perform scrubbing (remove
   volumes with high motion) and/or non-steady-state volumes. Masking is applied
   before signal cleaning.
-- All inherent classes of :class:`nilearn.input_data.BaseMasker` can use
+- All inherent classes of `nilearn.input_data.BaseMasker` can use
   parameter `sample_mask` for sub-sample masking.
 
 Fixes
@@ -58,7 +58,7 @@ Enhancements
   deprecates attribute `sessions` in 0.9.0. Match the relevant change in
   :func:`nilearn.signal.clean`.
 - Moves parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
-  to method `transform` in base class :class:`nilearn.input_data.BaseMasker`.
+  to method `transform` in base class `nilearn.input_data.BaseMasker`.
 
 .. _v0.7.1:
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -52,7 +52,7 @@ Enhancements
   is applied before signal cleaning.
   shape: (number of scans - number of volumes removed, )
 - :class:`nilearn.input_data.NiftiMasker` replace `sessions` with `runs` and
-  depricates attribute `sessions` in 0.9.0. Match the relevant change in
+  deprecates attribute `sessions` in 0.9.0. Match the relevant change in
   :func:`nilearn.signal.clean`.
 - Moving parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
   to method `transform` in base class :class:`nilearn.input_data.base_masker`.
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -3,14 +3,13 @@
 
 NEW
 ---
-- :func:`nilearn.signal.clean` accept new parameter `sample_mask`.
+
+- :func:`nilearn.signal.clean` accepts new parameter `sample_mask`.
   Masks the niimgs along time/fourth dimension to perform scrubbing (remove
   volumes with high motion) and/or non-steady-state volumes. This masking step
   is applied before signal cleaning.
   shape: (number of scans - number of volumes removed, )
-- Moving parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
-  to method `transform` in base class :class:`nilearn.input_data.base_masker`.
-  All inheritent classes of :class:`nilearn.input_data.base_masker` can use
+- All inheritent classes of :class:`nilearn.input_data.base_masker` can use
   parameter `sample_mask` for sub-sample masking.
 
 Fixes
@@ -58,6 +57,8 @@ Enhancements
 - :class:`nilearn.input_data.NiftiMasker` replace `sessions` with `runs` and
   deprecates attribute `sessions` in 0.9.0. Match the relevant change in
   :func:`nilearn.signal.clean`.
+- Moves parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
+  to method `transform` in base class :class:`nilearn.input_data.base_masker`.
 
 .. _v0.7.1:
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -9,7 +9,7 @@ NEW
   Masks the niimgs along time/fourth dimension to perform scrubbing (remove
   volumes with high motion) and/or non-steady-state volumes. Masking is applied
   before signal cleaning.
-- All inherent classes of :class:`nilearn.input_data.base_masker` can use
+- All inherent classes of :class:`nilearn.input_data.BaseMasker` can use
   parameter `sample_mask` for sub-sample masking.
 
 Fixes
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -3,11 +3,20 @@
 
 NEW
 ---
+- :func:`nilearn.signal.clean` accept new parameter `sample_mask`.
+  Masks the niimgs along time/fourth dimension to perform scrubbing (remove
+  volumes with high motion) and/or non-steady-state volumes. This masking step
+  is applied before signal cleaning.
+  shape: (number of scans - number of volumes removed, )
+  - Moving parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
+  to method `transform` in base class :class:`nilearn.input_data.base_masker`.
+  All inheritent classes of :class:`nilearn.input_data.base_masker` can use
+  parameter `sample_mask` for sub-sample masking.
 
 Fixes
 -----
 
-- Convert references in atlas.py, func.py, neurovault.py, and struct.py 
+- Convert references in atlas.py, func.py, neurovault.py, and struct.py
   to use footcite / footbibliography.
 - Fix detrending and temporal filtering order for confounders
   in :func:`nilearn.signal.clean`, so that these operations are applied
@@ -20,7 +29,7 @@ Fixes
   :func:`nilearn.plotting.plot_stat_map`) will now plot the slices in the user
   specified order, rather than automatically sorting into ascending order
   (https://github.com/nilearn/nilearn/issues/1155).
-- Fix the axes zoom on plot_img_on_surf function so brain would not be cutoff, and edited function so less white space surrounds brain views & smaller colorbar using gridspec (https://github.com/nilearn/nilearn/pull/2798). 
+- Fix the axes zoom on plot_img_on_surf function so brain would not be cutoff, and edited function so less white space surrounds brain views & smaller colorbar using gridspec (https://github.com/nilearn/nilearn/pull/2798).
 - Fix inconsistency in prediction values of Dummy Classifier for Decoder
   object (https://github.com/nilearn/nilearn/issues/2767).
 
@@ -46,18 +55,9 @@ Enhancements
   'stratified' (https://github.com/nilearn/nilearn/pull/2826/).
 - :class:`nilearn.glm.first_level.run_glm` now allows auto regressive noise
   models of order greater than one.
-- :func:`nilearn.signal.clean` accept new parameter `sample_mask`.
-  Masks the niimgs along time/fourth dimension to perform scrubbing (remove
-  volumes with high motion) and/or non-steady-state volumes. This masking step
-  is applied before signal cleaning.
-  shape: (number of scans - number of volumes removed, )
 - :class:`nilearn.input_data.NiftiMasker` replace `sessions` with `runs` and
   deprecates attribute `sessions` in 0.9.0. Match the relevant change in
   :func:`nilearn.signal.clean`.
-- Moving parameter `sample_mask` from :class:`nilearn.input_data.NiftiMasker`
-  to method `transform` in base class :class:`nilearn.input_data.base_masker`.
-  All inheritent classes of :class:`nilearn.input_data.base_masker` can use
-  parameter `sample_mask` for sub-sample masking.
 
 .. _v0.7.1:
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -18,7 +18,7 @@ Fixes
   :func:`nilearn.plotting.plot_stat_map`) will now plot the slices in the user
   specified order, rather than automatically sorting into ascending order
   (https://github.com/nilearn/nilearn/issues/1155).
-- Fix the axes zoom on plot_img_on_surf function so brain would not be cutoff, and edited function so less white space surrounds brain views & smaller colorbar using gridspec (https://github.com/nilearn/nilearn/pull/2798). 
+- Fix the axes zoom on plot_img_on_surf function so brain would not be cutoff, and edited function so less white space surrounds brain views & smaller colorbar using gridspec (https://github.com/nilearn/nilearn/pull/2798).
 
 
 Enhancements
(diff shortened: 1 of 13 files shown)


nilearn/input_data/nifti_labels_masker.py
```diff
@@ -242,19 +242,20 @@ def fit(self, X=None, y=None):
...
+                                    sample_mask=sample_mask)
 
     def _check_fitted(self):
         if not hasattr(self, "labels_img_"):
             raise ValueError('It seems that %s has not been fitted. '
                              'You must call fit() before calling transform().'
                              % self.__class__.__name__)
 
-    def transform_single_imgs(self, imgs, confounds=None):
+    def transform_single_imgs(self, imgs, confounds=None, sample_mask=None):
```
bthirion, line 258: Please add a description in docstrings

nilearn/input_data/nifti_maps_masker.py
```diff
@@ -237,13 +237,14 @@ def _check_fitted(self):
...
+    def fit_transform(self, imgs, confounds=None, sample_mask=None):
         """Prepare and perform signal extraction.
 
         """
-        return self.fit().transform(imgs, confounds=confounds)
+        return self.fit().transform(imgs, confounds=confounds,
+                                    sample_mask=sample_mask)
 
-    def transform_single_imgs(self, imgs, confounds=None):
+    def transform_single_imgs(self, imgs, confounds=None, sample_mask=None):
```
bthirion, line 247: Please complete the docstring.

nilearn/input_data/nifti_masker.py
```diff
@@ -408,7 +403,8 @@ def fit(self, imgs=None, y=None):
 
         return self
 
-    def transform_single_imgs(self, imgs, confounds=None, copy=True):
+    def transform_single_imgs(self, imgs, confounds=None, sample_mask=None,
```
bthirion, line 406: Same here.

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -355,17 +355,18 @@ def fit(self, X=None, y=None):
...
+                                    sample_mask=sample_mask)
 
     def _check_fitted(self):
         if not hasattr(self, "seeds_"):
             raise ValueError('It seems that %s has not been fitted. '
                              'You must call fit() before calling transform().'
                              % self.__class__.__name__)
 
-    def transform_single_imgs(self, imgs, confounds=None):
+    def transform_single_imgs(self, imgs, confounds=None, sample_mask=None):
```
bthirion, line 369: +1

nilearn/signal.py
```diff
@@ -459,6 +460,14 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
         If a list is provided, all confounds are removed from the input
         signal, as if all were in the same array.
 
+    sample_mask: None, numpy.ndarray, list, tuple, or list of
+        Default is None.
+        Masks the niimgs along time/fourth dimension. Containing indeice in a
```
bthirion, line 465:
```suggestion
        Masks the niimgs along time/fourth dimension. Containing indices in a
```

bthirion (commented):
Thx ! This is really good.
Best,

bthirion:
I realize that we are breaking  the API. Should we tolerate sample mask at class instantiation and issue a deprecation warning ?

htwangtw:
> I realize that we are breaking  the API. Should we tolerate sample mask at class instantiation and issue a deprecation warning ?

We certainly should. Thanks for pointing this out.

nilearn/_utils/helpers.py
```diff
@@ -107,3 +107,41 @@ def _transfer_deprecated_param_vals(replacement_params, kwargs):
...
+def remove_parameters(removed_params,
+                      reason,
+                      end_version='future'):
+    """Decorator to deprecate but not renamed parameters in the decorated
+    functions and methods.
+
+    Parameters
+    ----------
+    removed_params : list[string]
+        List of old parameters to be rmoved.
```
NicolasGensollen, line 121:
```suggestion
        List of old parameters to be removed.
```

nilearn/input_data/nifti_masker.py
```diff
@@ -227,7 +231,7 @@ def __init__(self, mask_img=None, sessions=None, smoothing_fwhm=None,
         # Mask is provided or computed
         self.mask_img = mask_img
 
-        self.sessions = sessions
+        self.runs = runs
```
NicolasGensollen, line 234: I'm wondering whether we should deprecate this attribute, because this could break user code relying on the "sessions" attribute of the object. WDYT?
Maybe by doing something like this:

```python
@property
def sessions(self):
    warnings.warn(("'sessions' attribute is deprecated and will "
                   "be removed in 0.9.0, use 'runs' instead."))
    return self.runs
```

NicolasGensollen (commented):
Thanks @htwangtw !  
I'll do another pass, but this is looking good to me so far! :+1:

nilearn/input_data/nifti_masker.py
htwangtw: Good point - I haven't test out the impact of my change yet. 
I will try your suggested solution.

NicolasGensollen (commented):
LGTM, thanks @htwangtw !

Could you add a whatsnew entry?

nilearn/input_data/base_masker.py
```diff
@@ -145,6 +147,12 @@ def transform_single_imgs(self, imgs, confounds=None, copy=True):
             documentation for details.
             shape: (number of scans, number of confounds)
 
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+            shape: (number of scans - number of volumes removed, )
```
bthirion, line 154: I think it is mroe common to put the shape before the explanation text in the docstring.

nilearn/input_data/base_masker.py
```diff
@@ -172,6 +180,12 @@ def transform(self, imgs, confounds=None):
...
```
bthirion, line 187: same here (etc.)

nilearn/signal.py
```diff
@@ -459,6 +460,15 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
...
         signal, as if all were in the same array.
 
+    sample_mask: None, numpy.ndarray, list, tuple, or list of
+        Default is None.
+        Masks the niimgs along time/fourth dimension to perform scrubbing
+        (remove volumes with high motion) and/or non-steady-state volumes.
+        This masking step is applied before signal cleaning. When supplying run
+        information, sample_mask must be a list containing sets of indexes for
+        each run.
+        shape: (number of scans - number of volumes removed, )
```
bthirion, line 470: Here agin

bthirion (commented):
This is great. With a small improvement in the docstring, it is almost perfect !

bthirion (approved):
LGTM, thx.

nilearn/input_data/nifti_masker.py
```diff
@@ -169,13 +176,6 @@ class NiftiMasker(BaseMasker, CacheMixin):
         to fine-tune mask computation. Please see the related documentation
         for details.
 
-    sample_mask : Any type compatible with numpy-array indexing, optional
-        Masks the niimgs along time/fourth dimension. This complements
-        3D masking by the mask_img argument. This masking step is applied
-        before data preprocessing at the beginning of NiftiMasker.transform.
-        This is useful to perform data subselection as part of a scikit-learn
-        pipeline.
```
NicolasGensollen, line 177: I would keep the docstring but add a sphinx deprecation directive:

```python
sample_mask : type
    description....

        .. deprecated:: 0.7.2
            `sample_mask` is deprecated in 0.7.2 and will be removed in 0.9.0.

File: nilearn/input_data/nifti_masker.py [hunk 12]
```diff
@@ -423,6 +435,12 @@ def transform_single_imgs(self, imgs, confounds=None, copy=True):
             documentation for details: :func:`nilearn.signal.clean`.
             shape: (number of scans, number of confounds)
 
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            shape: (number of scans - number of volumes removed, )
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
```
NicolasGensollen, line 442: I think we could add a `versionadded` directive (for all maskers) to advertise for this new feature. WDYT?

```python
sample_mask : type
    Description...

        .. versionadded:: 0.7.2

```

nilearn/input_data/nifti_masker.py
```diff
@@ -449,6 +466,7 @@ def transform_single_imgs(self, imgs, confounds=None, copy=True):
             memory=self.memory,
             verbose=self.verbose,
             confounds=confounds,
+            sample_mask=sample_mask,
```
NicolasGensollen, line 469: I realize that we can instantiate a `NiftiMasker` with a `sample_mask` and then call `transform` with another `sample_mask`. Unless I'm missing something, both will be sent to `filter_and_mask` (one through `params` and the other explicitly). The one provided to `transform` will then override the one provided to the constructor which is the desired behavior, but I'm wondering if we should warn the user that this is the case... I don't have a strong opinion about this. I you think this is too much we can leave it like this.

```python
import numpy as np
from nilearn import datasets

sample_mask = np.array([1,3,5,7])
masker = NiftiMasker(sample_mask=sample_mask)
masker.fit(dataset.func[0])
time_series = masker.transform(dataset.func[0],
                               sample_mask=np.array([2,4]))
time_series.shape
```
```
(2, 32504)
```

nilearn/input_data/nifti_masker.py
NicolasGensollen: I think we need to do the same thing for the `sample_mask` attribute which will also disappear. 

Maybe setting:

```python
self._sample_mask = sample_mask
```

in the constructor. And add:

```python
@property
def sample_mask(self):
    warnings.warn('deprecated...will be removed...')
    return self._sample_mask
```

WDYT?

NicolasGensollen (commented):
Thanks @htwangtw !

This looks good to me! :+1:  
I thought about a couple things to improve the way we deprecate `sample_mask` here (see below). Maybe we don't need to implement everything though (especially warning about `sample_mask` override in `transform`...), let me know what you think!

nilearn/input_data/nifti_masker.py
htwangtw: Thanks for finding this out! I would prefer to remove the capacity to remove supplying sample_mask from Niftimasker

nilearn/input_data/nifti_masker.py
htwangtw: Certainly!

nilearn/input_data/nifti_masker.py
htwangtw: After a second thought - 
- `sample_mask` in `NiftiMasker`, nothing from `transform`: use deprecated attribute `NiftiMasker._sample_mask`
- supply both: show warning, use `sample_mask` from transform

nilearn/input_data/nifti_masker.py
```diff
@@ -438,7 +473,18 @@ def transform_single_imgs(self, imgs, confounds=None, copy=True):
...
         # target_shape and target_affine are conveyed implicitly in mask_img
         params = get_params(self.__class__, self,
-                            ignore=['mask_img', 'mask_args', 'mask_strategy'])
+                            ignore=['mask_img', 'mask_args', 'mask_strategy',
+                                    '_sample_mask', 'sample_mask'])
+
+        if hasattr(self, '_sample_mask') and self._sample_mask is not None:
+            if sample_mask is not None:
+                warnings.warn(
+                    UserWarning("Overwriting depricated attribute "
```
NicolasGensollen, line 482:
```suggestion
                    UserWarning("Overwriting deprecated attribute "
```

nilearn/input_data/nifti_masker.py
```diff
@@ -438,7 +473,18 @@ def transform_single_imgs(self, imgs, confounds=None, copy=True):
...
+                                "`NiftiMasker.sample_mask` with parameter "
+                                "`sample_masker` in method `transform`.")
```
NicolasGensollen, line 484:
```suggestion
                                "`sample_mask` in method `transform`.")
```

nilearn/input_data/tests/test_nifti_masker.py
```diff
@@ -160,19 +160,39 @@ def test_mask_4d():
...
     assert_array_equal(data_trans, data_trans_direct)
 
-    masker = NiftiMasker(mask_img=mask_img, sample_mask=sample_mask)
+    masker = NiftiMasker(mask_img=mask_img)
     masker.fit()
-    data_trans2 = masker.transform(data_img_4d)
+    data_trans2 = masker.transform(data_img_4d, sample_mask=sample_mask)
     assert_array_equal(data_trans2, data_trans_direct)
 
+    # check deprication warning, and the old API shoud still work
```
NicolasGensollen, line 176:
```suggestion
    # check deprecation warning, and the old API should still work
```

nilearn/input_data/tests/test_nifti_masker.py
```diff
@@ -160,19 +160,39 @@ def test_mask_4d():
...
+    data_trans_img_diff = index_img(data_img_4d, diff_sample_mask)
+    data_trans_direct_diff = get_data(data_trans_img_diff)[mask_bool, :]
+    data_trans_direct_diff = np.swapaxes(data_trans_direct_diff, 0, 1)
+    masker = NiftiMasker(mask_img=mask_img, sample_mask=sample_mask)
+    masker.fit()
+    with pytest.warns(UserWarning, match=r'^Overwriting') as record:
+        data_trans3 = masker.transform(data_img_4d,
+                                       sample_mask=diff_sample_mask)
+    assert_array_equal(data_trans3, data_trans_direct_diff)
+    assert "Overwriting depricated attribute " in record[0].message.args[0]
```
NicolasGensollen, line 195:
```suggestion
    assert "Overwriting deprecated attribute " in record[0].message.args[0]
```

nilearn/input_data/tests/test_nifti_masker.py
```diff
@@ -233,6 +253,11 @@ def test_sessions():
     masker = NiftiMasker(sessions=np.ones(3, dtype=np.int))
     pytest.raises(ValueError, masker.fit_transform, data_img)
 
+    # check deprication warning of attribute session
```
NicolasGensollen, line 256:
```suggestion
    # check deprecation warning of attribute sessions
```

nilearn/signal.py
```diff
@@ -459,6 +460,15 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
...
 
+    sample_mask: None, numpy.ndarray, list, tuple, or list of
+        Default is None.
+        shape: (number of scans - number of volumes removed, )
+        Masks the niimgs along time/fourth dimension to perform scrubbing
+        (remove volumes with high motion) and/or non-steady-state volumes.
+        This masking step is applied before signal cleaning. When supplying run
+        information, sample_mask must be a list containing sets of indexes for
+        each run.
+
```
NicolasGensollen, line 471: Could you add `versionadded:: 0.7.2` here?

NicolasGensollen (commented):
Thanks @htwangtw !

Just a few typos:

nilearn/input_data/base_masker.py
```diff
@@ -145,6 +144,13 @@ def transform_single_imgs(self, imgs, confounds=None, copy=True):
             documentation for details.
             shape: (number of scans, number of confounds)
 
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            shape: (number of scans - number of volumes removed, )
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+                .. versionadded:: 0.7.2
```
NicolasGensollen, line 152: I think you need a blank line before this line, otherwise Sphinx will give a warning which will be treated as an error when building the doc in strict mode. 

```
...
This parameter is passed to signal.clean.

    .. versionadded:: 0.7.2
```

nilearn/input_data/base_masker.py
```diff
@@ -172,6 +178,13 @@ def transform(self, imgs, confounds=None):
...
```
NicolasGensollen, line 186: same here

nilearn/input_data/base_masker.py
```diff
@@ -213,6 +230,11 @@ def fit_transform(self, X, y=None, confounds=None, **fit_params):
             List of confounds (2D arrays or filenames pointing to CSV
             files). Must be of same length than imgs_list.
 
+        sample_mask : list of sample_mask, optional
+            List of sample_mask (1D arrays) if scrubbing motion outliers.
+            Must be of same length than imgs_list.
+                .. versionadded:: 0.7.2
```
NicolasGensollen, line 236: same

nilearn/input_data/multi_nifti_masker.py
```diff
@@ -261,6 +262,11 @@ def transform_imgs(self, imgs_list, confounds=None, copy=True, n_jobs=1):
             List of confounds (2D arrays or filenames pointing to CSV
             files or pandas DataFrames). Must be of same length than imgs_list.
 
+        sample_mask : list of sample_mask, optional
+            List of sample_mask (1D arrays) if scrubbing motion outliers.
+            Must be of same length than imgs_list.
+                .. versionadded:: 0.7.2
```
NicolasGensollen, line 268: same

nilearn/input_data/multi_nifti_masker.py
```diff
@@ -332,6 +338,11 @@ def transform(self, imgs, confounds=None):
             This parameter is passed to signal.clean. Please see the
             corresponding documentation for details.
 
+        sample_mask : list of sample_mask, optional
+            List of sample_mask (1D arrays) if scrubbing motion outliers.
+            Must be of same length than imgs_list.
+                .. versionadded:: 0.7.2
```
NicolasGensollen, line 344: same

nilearn/input_data/nifti_labels_masker.py
```diff
@@ -269,6 +270,13 @@ def transform_single_imgs(self, imgs, confounds=None):
             documentation for details.
             shape: (number of scans, number of confounds)
 
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            shape: (number of scans - number of volumes removed, )
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+                .. versionadded:: 0.7.2
```
NicolasGensollen, line 278: same

nilearn/input_data/nifti_maps_masker.py
```diff
@@ -258,6 +259,13 @@ def transform_single_imgs(self, imgs, confounds=None):
             documentation for details.
             shape: (number of scans, number of confounds)
 
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            shape: (number of scans - number of volumes removed, )
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+                .. versionadded:: 0.7.2
```
NicolasGensollen, line 267: same

nilearn/input_data/nifti_masker.py
```diff
@@ -175,6 +181,9 @@ class NiftiMasker(BaseMasker, CacheMixin):
         before data preprocessing at the beginning of NiftiMasker.transform.
         This is useful to perform data subselection as part of a scikit-learn
         pipeline.
+            .. deprecated:: 0.7.2
```
NicolasGensollen, line 184: same

nilearn/input_data/nifti_masker.py
NicolasGensollen: same

nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -380,6 +381,13 @@ def transform_single_imgs(self, imgs, confounds=None):
             documentation for details.
             shape: (number of scans, number of confounds)
 
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+            shape: (number of scans - number of volumes removed, )
+                .. versionadded:: 0.7.2
```
NicolasGensollen, line 389: same

nilearn/signal.py
```diff
@@ -459,6 +460,16 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
...
 
+    sample_mask: None, numpy.ndarray, list, tuple, or list of
+        Default is None.
+        shape: (number of scans - number of volumes removed, )
+        Masks the niimgs along time/fourth dimension to perform scrubbing
+        (remove volumes with high motion) and/or non-steady-state volumes.
+        This masking step is applied before signal cleaning. When supplying run
+        information, sample_mask must be a list containing sets of indexes for
+        each run.
+            .. versionadded:: 0.7.2
```
NicolasGensollen, line 471: same

NicolasGensollen (commented):
Thanks @htwangtw !
I think you need to have a blank line before each sphinx directive. Otherwise we will get a warning when building the doc.
Also, could you put the tag "[circle full]" in your commit message in order to trigger a full build in strict mode such that we can check that we don't get any error when building the doc? Thx!

htwangtw:
> Thanks @htwangtw !
> I think you need to have a blank line before each sphinx directive. Otherwise we will get a warning when building the doc.
> Also, could you put the tag "[circle full]" in your commit message in order to trigger a full build in strict mode such that we can check that we don't get any error when building the doc? Thx!

Will certainly do that - only just realised how long the doc build is on my side, even when bandwidth is not a issue...

doc/whats_new.rst
```diff
@@ -4,10 +4,18 @@
 NEW
 ---
 
+- :func:`nilearn.signal.clean` accepts new parameter `sample_mask`.
+  shape: (number of scans - number of volumes removed, )
+  Masks the niimgs along time/fourth dimension to perform scrubbing (remove
+  volumes with high motion) and/or non-steady-state volumes. Masking is applied
+  before signal cleaning.
+- All inherent classes of :class:`nilearn.input_data.BaseMasker` can use
```
NicolasGensollen, line 12: The reference to `BaseMasker` doesn't work since it is not exposed. CircleCI is failing because of this.

NicolasGensollen (commented):
Thanks @htwangtw !

Looks like CircleCI failed because of a reference to `BaseMasker` (see below).

NicolasGensollen (approved):
LGTM, thanks @htwangtw !

NicolasGensollen:
Merging :tada: 
thanks!

PR #2643, opened 2020-12-23
Add cbar_tick_format arg to plot_surf

NicolasGensollen:
Fixes #2220 

Add a `cbar_tick_format` argument to the function `plot_surf` such that users can configure how colorbar tick labels are displayed. The default is in scientific notation.

**Example**

Plot colorbar tick labels as integers:

```python
import matplotlib.pyplot as plt
from matplotlib import ticker
from nilearn import plotting, datasets

destrieux_atlas = datasets.fetch_atlas_surf_destrieux()
parcellation = destrieux_atlas['map_left']
surface = datasets.fetch_surf_fsaverage()

plotting.plot_surf_roi(surface["pial_left"], roi_map = parcellation,
        hemi = 'left', view = 'lateral', bg_map = surface["sulc_left"],
        bg_on_data = True, darkness = 0.6,
        cmap = 'jet', colorbar = True, 
        vmin=1, vmax=210, cbar_tick_format="%i")
plt.show()
```

Issue #2761, opened 2021-04-15
Consistent plotting API


NicolasGensollen:
See #2114 

Most plotting functions in nilearn return a `Figure` instance while some (like `plot_matrix`) return `Axes` (some functions like `plot_img_on_surf` return both figure and axes). This can be confusing for users (see `savefig` errors of #2114 for example).
It would be nice to offer a more consistent API by having all plotting functions return figure-like objects. 

A possible way forward could be to start deprecating these axes-returning functions and plan for changing their return values in release 0.9.
WDYT?

GaelVaroquaux:
> It would be nice to offer a more consistent API by having all plotting functions return figure-like objects.

That would break major features, such as the use of "add_overlay".

tsalo:
Would returning both figure and axes (whether list of Axes or just one Axes) like [`matplotlib.pyplot.subplots`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplots.html) work as a compromise?

bthirion:
This sounds reasonable to me.

mcsitter:
I’d propose starting by making the plotting functions consistently not return `None`, and adding proper return type annotations. Making the API consistent can be left as a follow-up once the return types are clearer.

Before moving forward, just checking if this direction sounds good / is wanted. I’m a bit busy at the moment, so if someone else wants to take this on that would be great. Otherwise I might get to it when I have time.

hndgzkn:
> I’d propose starting by making the plotting functions consistently not return `None`, and adding proper return type annotations. 

There is already a PR on that #6272 :)

mcsitter:
Thanks, didn't notice!

PR #2859, opened 2021-05-31
[ENH] Add `cbar_tick_format` to plotting functions

NicolasGensollen:
This PR adds the possibility to control the number formatting on the colorbars of volume plots through a `cbar_tick_format` argument. This extends what has been done in #2643 for surface plots.

### DIFF
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -296,8 +296,8 @@ def plot_img(img, cut_coords=None, output_file=None, display_mode='ortho',
         Default=False.
     cbar_tick_format: str, optional
         Controls how to format the tick labels of the colorbar.
-        Ex: use "%i" to display as integers.
-        Default is '%.2g' for scientific notation.
+        Ex: use "%%i" to display as integers.
+        Default is '%%.2g' for scientific notation.
     %(resampling_interpolation)s
         Default='continuous'.
     %(bg_img)s
@@ -491,10 +491,13 @@ def plot_anat(anat_img=MNI152TEMPLATE, cut_coords=None,
         Default='auto'.
     %(cmap)s
         Default=`plt.cm.gray`.
+    colorbar : boolean, optional
+        If True, display a colorbar on the right of the plots.
+        Default=False.
     cbar_tick_format: str, optional
         Controls how to format the tick labels of the colorbar.
-        Ex: use "%i" to display as integers.
-        Default is '%.2g' for scientific notation.
+        Ex: use "%%i" to display as integers.
+        Default is '%%.2g' for scientific notation.
     %(vmin)s
     %(vmax)s
 
@@ -548,10 +551,13 @@ def plot_epi(epi_img=None, cut_coords=None, output_file=None,
     %(draw_cross)s
     %(black_bg)s
         Default=True.
+    colorbar : boolean, optional
+        If True, display a colorbar on the right of the plots.
+        Default=False.
     cbar_tick_format: str, optional
         Controls how to format the tick labels of the colorbar.
-        Ex: use "%i" to display as integers.
-        Default is '%.2g' for scientific notation.
+        Ex: use "%%i" to display as integers.
+        Default is '%%.2g' for scientific notation.
     %(cmap)s
         Default=`plt.cm.nipy_spectral`.
     %(vmin)s
@@ -659,10 +665,13 @@ def plot_roi(roi_img, bg_img=MNI152TEMPLATE, cut_coords=None,
         Default=`plt.cm.gist_ncar`.
     %(dim)s
         Default='auto'.
+    colorbar : boolean, optional
+        If True, display a colorbar on the right of the plots.
+        Default=False.
     cbar_tick_format: str, optional
         Controls how to format the tick labels of the colorbar.
-        Ex: use "%i" to display as integers.
-        Default is '%.2g' for scientific notation.
+        Ex: use "%%i" to display as integers.
+        Default is '%%.2g' for scientific notation.
     %(vmin)s
     %(vmax)s
     %(resampling_interpolation)s
@@ -925,8 +934,8 @@ def plot_stat_map(stat_map_img, bg_img=MNI152TEMPLATE, cut_coords=None,
         Default=True.
     cbar_tick_format: str, optional
         Controls how to format the tick labels of the colorbar.
-        Ex: use "%i" to display as integers.
-        Default is '%.2g' for scientific notation.
+        Ex: use "%%i" to display as integers.
+        Default is '%%.2g' for scientific notation.
     %(figure)s
     %(axes)s
     %(title)s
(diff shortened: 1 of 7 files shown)


jeromedockes:
shouldn't `cbar_tick_format` appear in the other functions' signature? `plot_stat_map`, `plot_roi`, etc.

also, for future properties of the figure users may want to configure, rather than having many arguments in all functions that get forwarded through many calls, I think we should recommend using matplotlib directly and address these requests in the documentation. for example the tick format could be changed with one line:

```
plotting.plot_stat_map(img, colorbar=True)
plt.gcf().axes[-1].yaxis.set_major_formatter("z = {x:.1f}")
```

nilearn could make this easier with more examples and by returning better handles to all the matplotlib objects it creates; for example `plot_stat_map` returns handles to the axes containing the brain slices but not to the colorbar, which has to be fetched from the matplotlib figure

NicolasGensollen:
>shouldn't cbar_tick_format appear in the other functions' signature? plot_stat_map, plot_roi, etc.

I think so. I started doing that and quickly realized that, for some of these functions, some parameters are passed "silently"  through kwargs. For example, `plot_roi` doesn't have an explicit `colorbar` parameter:

https://github.com/nilearn/nilearn/blob/685099f4d0e386bcc2eda85375d9620a973b3464/nilearn/plotting/img_plotting.py#L772-L778

I'm not sure whether these are intentional or not. I don't mind fixing them here if they weren't.

>I think we should recommend using matplotlib directly and address these requests in the documentation

I agree. However, if the option is considered useful enough, I think it should be implemented as an argument for convenience (which I think is the case here but this is subjective of course...). 

>nilearn could make this easier with more examples and by returning better handles to all the matplotlib objects it creates

I think this was more or less the general idea behind #2761 but it got stuck because of API breaking change concerns. I'd be happy to bring that up in the next meeting though...

jeromedockes:
> I think so. I started doing that and quickly realized that, for some of these functions, some parameters are passed "silently"  through kwargs. For example, `plot_roi` doesn't have an explicit `colorbar` parameter:
>
> https://github.com/nilearn/nilearn/blob/685099f4d0e386bcc2eda85375d9620a973b3464/nilearn/plotting/img_plotting.py#L772-L778
>
> I'm not sure whether these are intentional or not. I don't mind fixing them here if they weren't.

Maybe it wasn't added because it was assumed that the colorbar would
rarely be used for `plot_roi`?

But having `**kwargs` forwarded through a deep stack of function calls
makes the documentation hard to read and often results in bugs such as
silently ignored arguments, so if you want to fix some of these
inconsistencies I think that would be great.

> I agree. However, if the option is considered useful enough, I think it should be implemented as an argument for convenience (which I think is the case here but this is subjective of course...). 
>

yes that's true, and I agree for the tick formatter as it has already
been added for `plot_surf`

>>nilearn could make this easier with more examples and by returning better handles to all the matplotlib objects it creates
>
> I think this was more or less the general idea behind #2761 but it got stuck because of API breaking change concerns. I'd be happy to bring that up in the next meeting though...
>

I think some small improvements could be done without breaking anything.
for example the `SomethingSlicer` returned by `plot_stat_map` have an
`axes` attribute, which is a dictionary containing the Axes in which
brains are displayed ,eg `{"x": <Axes>, "y": <Axes>, "z": <Axes>}`
a `"colorbar"` key could be added to that dictionary without breaking
backwards compatibility.

but maybe even this isn't necessary, and all that is missing is showing
how to modify these figures in some examples and pointing users to the
matplotlib documentation.

jeromedockes:
another example would be keeping a reference and giving access to the `AxesImage` returned by `imshow` when drawing the brains so that users can more easily create their own colorbars

nilearn/plotting/img_plotting.py
```diff
@@ -767,8 +798,9 @@ def plot_roi(roi_img, bg_img=MNI152TEMPLATE, cut_coords=None,
              output_file=None, display_mode='ortho', figure=None, axes=None,
              title=None, annotate=True, draw_cross=True, black_bg='auto',
              threshold=0.5, alpha=0.7, cmap=plt.cm.gist_ncar, dim='auto',
-             vmin=None, vmax=None, resampling_interpolation='nearest',
-             view_type='continuous', linewidths=2.5, **kwargs):
+             colorbar=False, cbar_tick_format="%.2g", vmin=None, vmax=None,
```
jeromedockes, line 801: for plot_roi don't you think "%i" could be a better default, as images contain integers?

bthirion (commented):
Can we have the equivalent for view_img ?

jeromedockes:
this one might also be good to merge before the 0.9 release?

nilearn/plotting/img_plotting.py
```diff
@@ -292,6 +295,10 @@ def plot_img(img, cut_coords=None, output_file=None, display_mode='ortho',
         Default=False.
     %(colorbar)s
         Default=False.
+    cbar_tick_format: str, optional
+        Controls how to format the tick labels of the colorbar.
+        Ex: use "%%i" to display as integers.
+        Default is '%%.2g' for scientific notation.
```
bthirion, line 301: Just curious: why is the % duplicated ?

jeromedockes: I believe due to the use of the `fill_doc` decorator, which applies `%` string formating to the docstring: https://github.com/nilearn/nilearn/blob/5cba192e3f1183330158f8d7469f36c01ba1f534/nilearn/_utils/docs.py#L790

NicolasGensollen: Absolutely! :)

nilearn/plotting/img_plotting.py
jeromedockes: @NicolasGensollen just pointing out this comment in case it got buried in the discussion and we forgot about it.  I think it should be easy to address by just changing `"%.2g"` to `"%i"` here but if I'm missing something and it isn't we can leave it as it is!

NicolasGensollen: Thanks for the reminder!
I totally agree that integer formatting makes more sense as default behavior in `plot_roi`.
See 3c9afc5

jeromedockes:
> Can we have the equivalent for view_img ?

I think that would be useful! But the formatting in that case is done by brainsprite (not by matplotlib or even python), and seems to be controlled by the "nbDecimals" brainsprite parameter. However when I try to set it to 0 for some reason one of the brains disappears -- maybe we should look into it in another PR to avoid delaying the enhancement of the matplotlib plots

jeromedockes (approved):
LGTM, thanks!

nilearn/plotting/img_plotting.py
bthirion: Thx

bthirion (approved):
Thx !

bthirion (approved):
LGTM, thx.

NicolasGensollen:
Thanks for the reviews!
Merging...

PR #2864, opened 2021-06-04
[MAINT] Add missing versions in some deprecation warnings

NicolasGensollen:
Some attributes in the GLM module are deprecated but the deprecation warnings were missing the version numbers. I assumed these are deprecated since 0.7.0 and will be removed in 0.9.0.
These was also a `@auto_attr` decorator missing for one of the `predicted` method.

### DIFF
--- a/nilearn/glm/regression.py
+++ b/nilearn/glm/regression.py
@@ -43,11 +43,11 @@ def _deprecation_warning(old_param,
                           message=("'{}' has been deprecated in version {} "
                                    "and will be removed in version {}. "
                                    "Please use '{}' instead.".format(
-                               old_param,
-                               start_version,
-                               end_version,
-                               new_param
-                           )))
+                                       old_param,
+                                       start_version,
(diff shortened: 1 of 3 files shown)


nilearn/glm/regression.py
```diff
@@ -312,7 +334,8 @@ class RegressionResults(LikelihoodModelResults):
     """
     @rename_parameters(
         {'wresid': 'whitened_residuals', 'wY': 'whitened_Y'},
-        lib_name='Nistats'
+        lib_name='Nistats',
```
bthirion, line 337: Nilearn

bthirion (commented):
LGTM !

nilearn/glm/regression.py
NicolasGensollen: Thanks!

bthirion (approved):

NicolasGensollen:
Thanks! Merging...

PR #2870, opened 2021-06-10
Fix deprecation of decomposition `components_`

### LINKED ISSUE #2862: Attribute `components_` is deprecated since 0.4.1 but still here
As can be seen in these files:

https://github.com/nilearn/nilearn/blob/dbbd221e3633c97464034c5366faa548728c9999/nilearn/decomposition/canica.py#L130-L134

https://github.com/nilearn/nilearn/blob/dbbd221e3633c97464034c5366faa548728c9999/nilearn/decomposition/dict_learning.py#L152-L155

https://github.com/nilearn/nilearn/blob/a36c0442fc8eebe401b310dd9799c0f071080e90/nilearn/decomposition/multi_pca.py#L121-L125

There is also a note here:

https://github.com/nilearn/nilearn/blob/a36c0442fc8eebe401b310dd9799c0f071080e90/examples/03_connectivity/plot_extract_regions_dictlearning_maps.py#L20-L26

We should remove them in 0.8.0.

NicolasGensollen:
Closes #2862 

This PR:

- Removes the deprecation message for attribute `components_` and replaces it with a usage note explaining that `components_img_` can be used instead. Both attributes are kept to conform with other objects in Nilearn (Decoder for example).
- Uses `components_img_` instead of unmasking `components_`
- Moves the definition and fitting of NiftiMapsMasker from `transform` and `inverse_transform` to `fit`

### DIFF
--- a/examples/03_connectivity/plot_extract_regions_dictlearning_maps.py
+++ b/examples/03_connectivity/plot_extract_regions_dictlearning_maps.py
@@ -17,13 +17,6 @@ ICA decomposition using :class:`nilearn.decomposition.CanICA`
 Please see the related documentation of :class:`nilearn.regions.RegionExtractor`
 for more details.
 
-.. note::
-
-    The use of the attribute `components_img_` from dictionary learning
-    estimator is implemented from version 0.4.1. For older versions,
-    unmask the deprecated attribute `components_` to get the components
-    image using attribute `masker_` embedded in estimator.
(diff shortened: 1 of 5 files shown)


jeromedockes (approved):
LGTM, thanks!

bthirion (approved):
Great, thx !

NicolasGensollen:
Thanks, merging!

PR #2873, opened 2021-06-14
Rewrite the modulation warning as a print

### LINKED ISSUE #2872: modulation column for make_first_level_design_matrix
When having a 'modulation' column in the events.tsv that is passed to ```nilearn.glm.first_level.make_first_level_design_matrix```, a UserWarning occurs saying 
```
UserWarning: ‘modulation’ column found in the given events data.
```
This should be changed into a simple message saying that the 'modulation'-column is being used rather than printing a warning which is confusing.

### Benefits to the change
Clearer information about what is happening 

```python
if modulation:
    print('Modulation column is found in events-data and is used')

```

NicolasGensollen:
Fixes #2872

### DIFF
--- a/nilearn/glm/first_level/experimental_paradigm.py
+++ b/nilearn/glm/first_level/experimental_paradigm.py
@@ -76,8 +76,8 @@ def check_events(events):
 
     # Handle modulation
     if 'modulation' in events_copy.columns:
-        warnings.warn("'modulation' column found in "
-                      "the given events data.")
+        print("A 'modulation' column was found in "
+              "the given events data and is used.")
     else:
         events_copy['modulation'] = 1
 
--- a/nilearn/glm/tests/test_paradigm.py
+++ b/nilearn/glm/tests/test_paradigm.py
@@ -129,9 +129,8 @@ def test_check_events():
 
     # Modulation is provided
     events['modulation'] = np.ones(len(events))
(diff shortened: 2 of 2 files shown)


rob-luke:
Does nilearn have an info or debug logger that would be more appropriate than a print for this purpose? Something like https://mne.tools/stable/generated/mne.set_log_level.html ?

NicolasGensollen:
@rob-luke AFAIK, there is `nilearn._utils.logger.py`, which does a few things but still ends up doing a simple print depending on the verbose level and message level parameters:

https://github.com/nilearn/nilearn/blob/372c19388da752caf724029c3d65c89dd88add0d/nilearn/_utils/logger.py#L79-L81

The function `check_events` has no verbose parameter that we could use here, so I ended up using a simple print rather than logging but I don't have a strong opinion on this.

rob-luke:
Thanks for sharing your thought process. I'm still learning the code base, so I appreciate the description.

bthirion:
We should get inspiration from MNE logging, which is much more consistent than in Nilearn !

bthirion (approved):
LGTM, thx !

NicolasGensollen:
>We should get inspiration from MNE logging, which is much more consistent than in Nilearn !

I'll put that on my todo list then! :)

Merging, thanks!

PR #2875, opened 2021-06-14
[MAINT] Add `fill_doc` decorator

NicolasGensollen:
This PR proposes to add a `fill_doc` decorator to automatically fill parts of docstrings. 

An example on the `datasets`  and `decoding` modules is also provided with the parameters `verbose`, `resume`, `data_dir`, `url`, `smoothing_fwhm`, `standardize`, `target_affine`, `target_shape`, `low_pass`, `high_pass`, `t_r`, `memory`, `memory_level`, and `n_jobs`.

Note: The code for the decorator comes from MNE (link at the top of `nilearn/_utils/docs.py`).

This has the following benefits:

- Standard documentation entries are all in one place
- Simplifies the maintenance of docstrings/documentation
- Reduces the length of files (several hundreds of lines were removed from the `datasets` module with this simple example although the rendering is exactly the same as before).

If we decide to adopt this, there are quite a few arguments that could be handled in this way across Nilearn.

### DIFF
--- a/nilearn/decoding/searchlight.py
+++ b/nilearn/decoding/searchlight.py
@@ -67,10 +67,8 @@ def search_light(X, y, estimator, A, groups=None, scoring=None,
         A cross-validation generator. If None, a 3-fold cross
         validation is used or 3-fold stratified cross-validation
         when y is supplied.
-    %(n_jobs)s
-        Default=-1.
-    %(verbose)s
-        Default=0.
+    %(n_jobs_all)s
+    %(verbose0)s
 
     Returns
     -------
@@ -101,8 +99,6 @@ class GroupIterator(object):
     n_features : int
         Total number of features
     %(n_jobs)s
-    n_jobs : int, optional
-        Defaut=1.
 
     """
     def __init__(self, n_features, n_jobs=1):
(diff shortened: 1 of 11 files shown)


nilearn/decoding/searchlight.py
```diff
@@ -99,10 +100,10 @@ class GroupIterator(object):
     ----------
     n_features : int
         Total number of features
-
+    %(n_jobs)s
     n_jobs : int, optional
```
bthirion, line 104: To be removed too ?

nilearn/decoding/space_net.py
```diff
@@ -1065,22 +1025,12 @@ class SpaceNetClassifier(BaseSpaceNet):
...
-    memory_level: integer, optional (default 1)
-        Rough estimator of the amount of memory used by caching. Higher value
-        means more memory for caching.
-
+    %(verbose)s
+        Default=1.
+    %(n_jobs)s
+        Default=1.
+    %(memory)s
+    %(memory_level)s
```
bthirion, line 1033: Default=1

bthirion (commented):
Looks great overall !

bthirion:
LGTM. When it's finalized, let me know so that I look more carefully at circleCI artifacts.

NicolasGensollen:
Thanks @bthirion ! I think this is already finalized. 
I took care of the most common arguments within the `datasets` and `decoding` modules, but I think we might want to do other files/modules in separate PRs as to not overload this one. WDYT?

bthirion (commented):
LGTM, I did not identify an issue in the doc.
Please add a whatsnew entry.

bthirion (approved):
LGTM, thx !

NicolasGensollen:
Thanks! Merging...

PR #2876, opened 2021-06-14
[FIX] Fix view_img colorbar

### LINKED ISSUE #2874: ticks missing from `view_img` colorbar for some images
```python
from nilearn import plotting, datasets

plotting.view_img(datasets.load_mni152_template(), bg_img=None).open_in_browser()
```

Screenshot from 2021-06-14 10-17-45

NicolasGensollen:
Fixes #2874 

This PR proposes a possible fix for #2874. 
I'm slightly out of my comfort zone here so I'll take any advice... :sweat_smile: 

```python
from nilearn import plotting, datasets

plotting.view_img(datasets.load_mni152_template(), black_bg=False, bg_img=None)
```
ticks2

### DIFF
--- a/nilearn/plotting/html_stat_map.py
+++ b/nilearn/plotting/html_stat_map.py
@@ -3,6 +3,7 @@ Visualizing 3D stat maps in a Brainsprite viewer
 """
 import os
 import json
+import copy
 import warnings
 from io import BytesIO
 from base64 import b64encode
@@ -23,6 +24,7 @@ from .._utils.param_validation import check_threshold
 from .._utils.extmath import fast_abs_percentile
 from .._utils.niimg import _safe_get_data
 from ..datasets import load_mni152_template
+from ..masking import compute_brain_mask
 
 
 def _data_to_sprite(data):
@@ -319,9 +321,17 @@ def _json_view_data(bg_img, stat_map_img, mask_img, bg_min, bg_max, black_bg,
 
     # Create a base64 sprite for the background
     bg_sprite = BytesIO()
-    bg_data = _safe_get_data(bg_img, ensure_finite=True)
-    bg_cmap = 'gray' if black_bg else cmap
-    _save_sprite(bg_data, bg_sprite, bg_max, bg_min, None, bg_cmap, 'png')
+    bg_data = _safe_get_data(bg_img, ensure_finite=True).astype(float)
+    bg_mask = _safe_get_data(compute_brain_mask(bg_img),
+                             ensure_finite=True)
+    bg_mask = np.logical_not(bg_mask).astype(float)
+    bg_mask[bg_mask == 1] = np.nan
+    bg_cmap = copy.copy(cmap)
+    if black_bg:
+        bg_cmap.set_bad('black')
+    else:
+        bg_cmap.set_bad('white')
+    _save_sprite(bg_data, bg_sprite, bg_max, bg_min, bg_mask, bg_cmap, 'png')
     json_view['bg_base64'] = _bytesIO_to_base64(bg_sprite)
 
     # Create a base64 sprite for the stat map
--- a/nilearn/plotting/html_stat_map.py
+++ b/nilearn/plotting/html_stat_map.py
@@ -310,6 +310,20 @@ def _json_view_size(params):
     return width_view, height_view
 
 
+def _get_bg_mask_and_cmap(bg_img, black_bg):
+    """Helper function of _json_view_data."""
+    bg_mask = _safe_get_data(compute_brain_mask(bg_img),
+                             ensure_finite=True)
+    bg_mask = np.logical_not(bg_mask).astype(float)
+    bg_mask[bg_mask == 1] = np.nan
+    bg_cmap = copy.copy(matplotlib.cm.get_cmap('gray'))
+    if black_bg:
+        bg_cmap.set_bad('black')
+    else:
+        bg_cmap.set_bad('white')
+    return bg_mask, bg_cmap
+
+
 def _json_view_data(bg_img, stat_map_img, mask_img, bg_min, bg_max, black_bg,
                     colors, cmap, colorbar):
     """Create a json-like viewer object, and populate with base64 data.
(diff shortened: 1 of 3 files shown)


jeromedockes:
The fix seems to work great! 
OTOH I'm not sure about the approach: rather than modify the copy of brainsprite distributed with nilearn, I believe it would make more sense to fix the bug upstream and distribute a more recent version of brainsprite.
(but we might run into trouble if brainsprite has made breaking changes since the `view_img` was added to nilearn...)

somewhat related: #2428 

WDYT @NicolasGensollen  @pbellec  ?

pbellec:
would be happy to look at and merge changes for a PR here: https://github.com/brainsprite/brainsprite/blob/master/brainsprite.js 
I would prepare a new release to update brainsprite in nilearn. 
The code base has, sadly, not changed much - just some optimizations - so it should be straightforward. 

I have been working (and still have a lot of work to do) on a python companion code, but the js codebase has been stable. 

Because of the js minified code base in this PR, I am not able to see the changes that are being proposed.

NicolasGensollen:
Thanks @pbellec @jeromedockes !

I propose another possible fix in f172b44 which doesn't change our copy of brainsprite.
Here is the rendering:

fix_view

jeromedockes:
Thanks @NicolasGensollen ! I like the idea. there seems to be a few kinks to work out when the background image is not None:
```python
from nilearn import plotting, datasets
img = datasets.fetch_neurovault_motor_task()["images"][0]
plotting.view_img(img, black_bg=False, threshold=2.).open_in_browser()
```

Screenshot from 2021-06-22 10-08-59

NicolasGensollen:
Thanks @jeromedockes ! :+1: 
I proposed a possible better fix in 7b1e150.

There are probably still a few things to improve in the code, but here is what different combinations of the parameters end up doing:

view1

view2

view3

view4

The crazy things:

view5

view6

jeromedockes:
looks good! I think the background's colormap should be restored to being grayscale by default?

nilearn/plotting/tests/test_html_stat_map.py
```diff
@@ -259,8 +259,8 @@ def test_json_view_data():
 
     # Build a sprite
     json_view = html_stat_map._json_view_data(
-        bg_img, stat_map_img, mask_img, bg_min=0, bg_max=1, colors=colors,
-        cmap='cold_hot', colorbar=True)
+        bg_img, stat_map_img, mask_img, bg_min=0, bg_max=1, black_bg=False,
+        colors=colors, cmap='cold_hot', colorbar=True)
```
bthirion, line 263: Does this cover all possible cases ?

bthirion (approved):
Vary hard for me to assess this PR, but it looks good.

bthirion:
Would like to have the opinion of @pbellec 
Thx !

jeromedockes:
so in the end it was necessary to change the brainsprite javascript code? If so, it may be better to fix it upstream than have our copy diverge from it?
practical detail: as the original version was minified the diff displayed by github is not meaningful so checking the changes is not easy

nilearn/plotting/html_stat_map.py
```diff
@@ -319,8 +322,17 @@ def _json_view_data(bg_img, stat_map_img, mask_img, bg_min, bg_max, colors,
 
     # Create a base64 sprite for the background
     bg_sprite = BytesIO()
-    bg_data = _safe_get_data(bg_img, ensure_finite=True)
-    _save_sprite(bg_data, bg_sprite, bg_max, bg_min, None, 'gray', 'png')
+    bg_data = _safe_get_data(bg_img, ensure_finite=True).astype(float)
+    bg_mask = _safe_get_data(compute_brain_mask(bg_img),
```
pbellec, line 326: looks like the block 326-335 could be turned into a helper function to increase code readability.

pbellec:
I went through the github history to check the changes in the javascript code. It does look like no change was applied in brainsprite per say. `brainsprite.min.js` is still marked as updated because it was beautified - transformed from an ultra-compact single-line "minified" state to a more readable, multi-line state. It does not look like this change impacts the size of the file (13K before and after beautification). It is a bit weird to have `min.js` file that's not fully minified, but in effect it still is minified to some degree (minification does simplify the code above and beyond making it one long string). So if the nilearn team would rather have a more readable version of the js code shipping, I think that change is reasonable. 

As of the fix itself. Conceptually I agree the right fix is to deal with brain masking correctly, and very happy to see this pushed. Regarding implementation, I left a comment that `_json_view_data` has become a bit long (35 lines or so) and the fix could be a bit more modular, with a separate helper function to generate the mask and associated colormap. Otherwise the code itself is relatively straightforward and LGTM.

jeromedockes:
> I went through the github history to check the changes in the javascript code. It does look like no change was applied in brainsprite per say. `brainsprite.min.js` is still marked as updated because it was beautified - transformed from an ultra-compact single-line "minified" state to a more readable, multi-line state. It does not look like this change impacts the size of the file (13K before and after beautification). It is a bit weird to have `min.js` file that's not fully minified, but in effect it still is minified to some degree (minification does simplify the code above and beyond making it one long string). So if the nilearn team would rather have a more readable version of the js code shipping, I think that change is reasonable. 

Thanks for looking at this!

If the code hasn't changed, @NicolasGensollen could you restore the
original brainsprite.min.js file?
- it will make it clearer in the git history that it wasn't actually modified
- keeping it as it was makes it clear that it is just a copy of the brainsprite library, not meant to be edited

NicolasGensollen:
>It does look like no change was applied in brainsprite per say. brainsprite.min.js is still marked as updated because it was beautified

Absolutely, I hadn't removed the commit beautifying the js file, sorry for the confusion. I rebased and dropped all commits related to the initial try at fixing this.

> I left a comment that _json_view_data has become a bit long (35 lines or so) and the fix could be a bit more modular, with a separate helper function to generate the mask and associated colormap.

Thanks for the advise! I implemented the proposed change in ae563fb

nilearn/plotting/html_stat_map.py
```diff
@@ -307,8 +310,22 @@ def _json_view_size(params):
...
 
 
-def _json_view_data(bg_img, stat_map_img, mask_img, bg_min, bg_max, colors,
-                    cmap, colorbar):
+def _get_bg_mask_and_cmap(bg_img, black_bg):
+    """Helper function of _json_view_data."""
+    bg_mask = _safe_get_data(compute_brain_mask(bg_img),
+                             ensure_finite=True)
+    bg_mask = np.logical_not(bg_mask).astype(float)
+    bg_mask[bg_mask == 1] = np.nan
```
jeromedockes, line 318: usually it's good to avoid such comparisons involving floating-point numbers

jeromedockes (approved):
LGTM, thanks!

pbellec:
Proposed changes LGTM. 

Very useful PR, I am looking forward to interactive viewers with white background!

bthirion:
merged, thx !

Issue #2879, opened 2021-06-18
The method of FDR threshholding in GLM


gourdchen:
The nilearn/glm/thresholding.py line 101：
` pos = p_vals < alpha * np.linspace(
        .5 / n_samples, 1 - .5 / n_samples, n_samples)`

In my opinion, the code should be 
`pos = p_vals < alpha * np.linspace(1/n_samles, 1,n_samples)`
Why does ’0.5‘ exist there？ It confuses me.

bthirion:
Indeed, this is an error. Can you issue a PR ?

AjithRamachandran:
Is anyone working on this rn?

bthirion:
No. Please go ahead if you wish.

AjithRamachandran:
Thank you

bthirion:
Close as fixed.

PR #2889, opened 2021-06-27
fix method of FDR threshholding in GLM

ajthr:
Fix GH-2879

### DIFF
--- a/nilearn/glm/tests/test_thresholding.py
+++ b/nilearn/glm/tests/test_thresholding.py
@@ -14,7 +14,7 @@ from nilearn.glm import (cluster_level_inference, fdr_threshold,
 def test_fdr():
     rng = np.random.RandomState(42)
     n = 100
-    x = np.linspace(.5 / n, 1. - .5 / n, n)
+    x = np.linspace(1 / n, 1, n)
     x[:10] = .0005
     x = norm.isf(x)
     rng.shuffle(x)
--- a/nilearn/glm/thresholding.py
+++ b/nilearn/glm/thresholding.py
@@ -99,7 +99,6 @@ def fdr_threshold(z_vals, alpha):
     p_vals = norm.sf(z_vals_)
     n_samples = len(p_vals)
     pos = p_vals < alpha * np.linspace(1 / n_samples, 1, n_samples)
-``` We need to enforce pep8 conventions !
     if pos.any():
         return (z_vals_[pos][-1] - 1.e-12)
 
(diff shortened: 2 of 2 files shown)


nilearn/glm/thresholding.py
```diff
@@ -98,8 +98,7 @@ def fdr_threshold(z_vals, alpha):
     z_vals_ = - np.sort(- z_vals)
     p_vals = norm.sf(z_vals_)
     n_samples = len(p_vals)
-    pos = p_vals < alpha * np.linspace(
-        .5 / n_samples, 1 - .5 / n_samples, n_samples)
+    pos = p_vals < alpha * np.linspace(1/n_samples, 1,n_samples)
```
bthirion, line 101:
```suggestion
    pos = p_vals < alpha * np.linspace(1 / n_samples, 1, n_samples)
``` We need to enforce pep8 conventions !

**Review by bthirion — COMMENTED** [review]
Thx ! 
Could you add a unit test to enforce this correction ? This is important to avoid a regression in the future. Thx !

**Comment by AjithRamachandran** [comment]
> Thx !
> Could you add a unit test to enforce this correction ? This is important to avoid a regression in the future. Thx !

Sure :)

File: nilearn/glm/thresholding.py [hunk 2]
```diff
@@ -98,8 +98,8 @@ def fdr_threshold(z_vals, alpha):
     z_vals_ = - np.sort(- z_vals)
     p_vals = norm.sf(z_vals_)
     n_samples = len(p_vals)
-    pos = p_vals < alpha * np.linspace(
-        .5 / n_samples, 1 - .5 / n_samples, n_samples)
+    pos = p_vals < alpha * np.linspace(1 / n_samples, 1, n_samples)
+``` We need to enforce pep8 conventions !
```
NicolasGensollen, line 102: I think you added this line unintentionally. It is not properly indented and causes the test errors you're getting.

NicolasGensollen (commented):
Thanks for opening this @AjithRamachandran !

nilearn/glm/thresholding.py
AjithRamachandran: Yes. I didnt see that there.

nilearn/glm/tests/test_thresholding.py
```diff
@@ -14,7 +14,7 @@
 def test_fdr():
     rng = np.random.RandomState(42)
     n = 100
-    x = np.linspace(.5 / n, 1. - .5 / n, n)
+    x = np.linspace(1 / n, 1, n)
```
bthirion, line 17: Thx, but the idea is to have a test such that it fails with the previous version, and passes after the fix.

bthirion:
I can try and figure out something.

bthirion:
the cranch has been deleted. I'm going to redo the PR from scratch...

bthirion:
superseded by #3137, closing

PR #2895, opened 2021-07-02
Unify argument names in view_markers and view_connectome

### LINKED ISSUE #2852: Discrepancy of argument names in view_markers and view_connectome
Following #2810 (discussion [here](https://github.com/nilearn/nilearn/pull/2810#issuecomment-838597255)), we noticed that arguments for `html_connectome.view_connectome` and `html_connectome.view_markers` have different names even though they reflect the same object: 

- in `view_markers`, arguments relative to the markers are prefixed with `marker_` (e.g `marker_coords`)
- in `view_connectome`, arguments relative to the markers are prefixed with `node_` (e.g `node_coords`)

My opinion is that we should converge to either one of them. What do you think?

Signatures of the two functions:

```python
def view_markers(marker_coords, marker_color=None, marker_size=5., marker_labels=None, title=None, title_fontsize=25)
```

```python
def view_connectome(adjacency_matrix, node_coords, edge_threshold=None, edge_cmap=cm.bwr, symmetric_cmap=True, linewidth=6., node_size=3., colorbar=True, colorbar_height=.5, colorbar_fontsize=25, title=None, title_fontsize=25)
```

RaphaelMeudec:
This PR fixes #2852 by removing any references to `markers` in the connectome plots. This implies changes in:

- html_connectome argument names and methods
- connectome template arguments

### DIFF
--- a/nilearn/plotting/html_connectome.py
+++ b/nilearn/plotting/html_connectome.py
@@ -166,9 +166,9 @@ def _prepare_lines_metadata(adjacency_matrix, coords, threshold,
 
 
 def _prepare_nodes_metadata(coords, node_size, node_color, node_only):
-    nodes_coordinates = _encode_coordinates(coords, prefix="_marker_")
+    nodes_coordinates = _encode_coordinates(coords, prefix="_node_")
     nodes_metadata = {
-        'markers_only': node_only,
+        'nodes_only': node_only,
         **nodes_coordinates
     }
 
@@ -176,8 +176,8 @@ def _prepare_nodes_metadata(coords, node_size, node_color, node_only):
         node_size = np.asarray(node_size)
     if hasattr(node_size, 'tolist'):
         node_size = node_size.tolist()
-    nodes_metadata['marker_size'] = node_size
-    nodes_metadata['marker_color'] = _prepare_colors_for_nodes(
+    nodes_metadata['node_size'] = node_size
+    nodes_metadata['node_color'] = _prepare_colors_for_nodes(
         node_color,
         len(coords),
     )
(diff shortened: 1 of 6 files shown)


nilearn/plotting/html_connectome.py
```diff
@@ -319,25 +319,25 @@ def view_connectome(adjacency_matrix, node_coords, edge_threshold=None,
     return _make_connectome_html(connectome_info)
 
 
-def view_markers(marker_coords, marker_color='auto', marker_size=5.,
-                 marker_labels=None, title=None, title_fontsize=25):
-    """Insert a 3d plot of markers in a brain into an HTML page.
+def view_nodes(node_coords, node_color='auto', node_size=5.,
```
NicolasGensollen, line 322: We need to deprecate `view_markers` (see my comment in the issue).

NicolasGensollen (commented):
Thanks @RaphaelMeudec ! :+1: 
I made a quick first pass and it looks good so far. My main comment is related to the deprecation of `view_markers` (to be removed in v0.10.0). We also need to update `doc/modules/references.rst` and add an entry in whatsnew.rst under "changes".
We also need to update examples using `view_markers` if any.

nilearn/plotting/html_connectome.py
RaphaelMeudec: Yes, I forgot about it sorry. Will take care of this now!

jeromedockes:
sorry but I'm not sure I agree with renaming view_markers -> view_nodes.
it is an important backwards incompatible change for little benefit, and it results in inconsistent naming with add_markers and plot_markers.

NicolasGensollen:
@jeromedockes do you mean that we should keep the fonction name as view_markers but rename it's parameters to use the "node convention"?

jeromedockes:
> @jeromedockes do you mean that we should keep the fonction name as view_markers but rename it's parameters to use the "node convention"?

I think so, otherwise this consistency rabbit hole would lead us to also
renaming plot_markers and add_markers and it becomes too much IMO.

RaphaelMeudec:
@jeromedockes @NicolasGensollen Closing this PR as discussed.

PR #2898, opened 2021-07-19
DOC: fix binder

GaelVaroquaux:
We have changed the name of the master branch to main

This should fix our binder setup


jeromedockes:
circleci failure seems unrelated; for some reason the working directory seems to be /home/circleci whereas it is not specified in the config and thus should be /home/circleci/project by default

jeromedockes:
circleci seems unrelated to the changes but due to the fact that tools has been split into build_tools and maint_tools by #2848 ; I think merging `main` should fix the issue?

GaelVaroquaux:
I rebased on main, let's wait for this to build.

PR #2899, opened 2021-07-19
[FIX] Fix aspect ratio in surface plotting

### LINKED ISSUE #2658: Plotting flat maps + plotting additional data as transparency
I am plotting Cifti files with [hcp-utils](https://rmldj.github.io/hcp-utils/) and nilearn. It works very nicely for the standard inflated surface but the aspect ratio is off for plotting a flatmap (squished to a square):

download (42)

download (43)

I'm so far unable to plot the bilateral brain flat map without this stretching happening. Is there interest in supporting plotting these flatmaps, maybe adding a keyword arg? And/or does anyone have tips on how to preserve the correct aspect ratio? I have been trying to modify the axes and fig.subplots_adjust but no luck.

Another related feature request is to be able to plot multiple stats on one surface map eg one type of data by color and another by transparency/alpha. On the above images for example, I've plotted my surf_data as normal and then blended it into the bg_map based on another map of r2score values. I could help to add this kind of functionality if it would be helpful and not complicate things too much. Thanks!

NicolasGensollen:
Fixes #2658 

WIP...

**Example**

For `plot_surf`:

```python
import hcp_utils as hcp
from nilearn import plotting, surface

surf = surface.load_surf_mesh(hcp.mesh.flat)
plotting.plot_surf(surf)
```

Now gives:

aspect

**TODO:** Fix `view_surf` which, as pointed out by @jeromedockes, suffers from the same problem although it relies on a different machinery.


bthirion (approved):
At least, the fix for `plot_surf` seems good.

NicolasGensollen:
Update: ea0f01e proposes a possible fix for `view_surf`. If you test it with:

```python
import hcp_utils as hcp
from nilearn import plotting, surface

surf = surface.load_surf_mesh(hcp.mesh.flat)
plotting.view_surf(surf)
```

You should get:

view_surf1

With the problem of the initial zoom that still need to be fixed. That is, initially the figure looks like this:

view_surf2


and you have to zoom out.

bthirion:
I still have the problem with the initial zoom.

NicolasGensollen:
>I still have the problem with the initial zoom.

Weird... :thinking:  
Do you still have the issue now? I changed the cameras settings and the examples in the artifacts look good now.

bthirion:
LGTM now.

bthirion (approved):
LGTM. Maybe a small whatsnew entry ?

nilearn/plotting/data/js/surface-plot-utils.js
```diff
@@ -76,22 +76,22 @@ function getCamera(plotDivId, viewSelectId) {
         }
     }
     let cameras = {
-        "left": {eye: {x: -1.7, y: 0, z: 0},
+        "left": {eye: {x: -3.0, y: 0, z: 0},
```
jeromedockes, line 79: the problem is that now the brain is too small and too far from the colorbar when plotting a surface map on a mesh in mni space (or close to it), which is the vast majority of use cases:

before:
newplot

now:

newplot(1)

I don't think deteriorating the plots for the most common use case to address an edge case is the way to go. maybe we should allow users to control the zoom (although they can do it with the mouse once the interactive plot is opened, and maybe further programmatic customization of the plots should be deferred to after #2902 is merged)

otherwise the fix to control the aspect ratio looks good!

ymzayek:
I can take this one over. It is still an issue and concluding from what is done here I would keep the aspect ratio fix for both `plot_surf` and `view_surf` but remove the changes in the camera settings. There will still be the problem of the initial zoom for `view_surf` being too big for this use case but the user can zoom out.

Remi-Gau:
- Closing in favor of #4184.

PR #2861, opened 2021-06-03
[MAINT] Rename master into main

NicolasGensollen:
Fixes #2841 

This PR prepares the renaming of branch `master` into `main`. 
I am not sure when we want to pull the trigger, but the steps will be the following:

- Rename branch `master` into `main` on the repo's main page (this should update the target branch of ongoing PRs including this one)
- Check that everything runs fine here and merge
- All clones will have to run locally:

```bash
$ git branch -m master main
$ git fetch origin
$ git branch -u origin/main main
$ git remote set-head origin -a
```

Note that GitHub will display a note on the main page saying that the main branch has been renamed, and that you should run these commands to update your clone.

PR #2901, opened 2021-07-20
[FIX] Fix binder links

NicolasGensollen:
Binder links seem to be broken due to the recent branch name change from "master" to "main" (see #2861 ).
See also #2898

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -331,7 +331,7 @@ sphinx_gallery_conf = {
         'org': 'nilearn',
         'repo': 'nilearn.github.io',
         'binderhub_url': 'https://mybinder.org',
-        'branch': 'main',
+        'branch': 'master',
         'dependencies': ['../requirements-build-docs.txt',
                          'binder/requirements.txt'],
         'notebooks_dir': 'examples'


emdupre:
The [most recent artifacts](https://7758-1235740-gh.circle-artifacts.com/0/dev/auto_examples/plot_nilearn_101.html#sphx-glr-auto-examples-plot-nilearn-101-py) launch binder successfully (:tada:), though interestingly the dialogue still suggests it's building from a 'master' rather than 'main' branch.

NicolasGensollen:
@emdupre I believe it is because we still have the "master convention" for `nilearn.github.io` (which was causing the bug since I updated it in `doc/conf.py` while I shouldn't had to... :sweat: ). So this PR is simply reverting this update.

NicolasGensollen:
Should I merge this one?

emdupre:
:+1: from me ! I guess we'll update the branch naming for the docs directory at the next release ? I don't think that should hold this up, though !!

emdupre (approved):

NicolasGensollen:
>I guess we'll update the branch naming for the docs directory at the next release ?

This is what I had in mind for some reason, but you're right, we should be able to change the name of the default branch of nilearn.github.io to main now if we want. 
I'll double check but I think we would just need to update the install target of `doc/Makefile`.

NicolasGensollen:
I renamed the branch to main on `nilearn.github.io` and rebuilt the docs.
It looks like everything is working again (including the binder links).
I'm closing this PR as we no longer need it.

emdupre:
Thanks, @NicolasGensollen !!

Issue #2790, opened 2021-05-05
More diverse avg_method for plot_surf to handle sparse matrices


alexisthual:
Currently, the `avg_method` parameter in `plotting.plot_surf` only takes `mean` or `median` as arguments.
Having the possibility `min`, `max` or other functions would be useful to me, in the sense that sometimes I have some sparse surface matrices, therefore using `mean` or `median` kind of kills my plots :smile: 

### Benefits to the change

It would allow for a better representation of sparse matrices on the surface.

### Pseudocode for the new behavior, if applicable

```python
import matplotlib.pyplot as plt
from nilearn import datasets, plotting
import numpy as np

fsaverage = datasets.fetch_surf_fsaverage()
n_voxels = 10242

surface = np.zeros(n_voxels)
n = 100
surface[:n] = np.ones(n)

plotting.plot_surf(
    fsaverage.pial_left,
    surface,
    avg_method="min"
)
plt.show()
```

NicolasGensollen:
Here is the part of the code implementing these:

https://github.com/nilearn/nilearn/blob/b08d1fea1857f988b0b916e6176235b1912bcaf5/nilearn/plotting/surf_plotting.py#L268-L273

I think having the possibility of specifying `min` or `max` would be a good addition indeed :+1:

PR #2902, opened 2021-07-21
[ENH] Add Plotly support for surface plotting

### LINKED ISSUE #2793: Adding Plotly (Python) figures for surface plots
- Matplotlib is not great for 3d (values for faces not vertices, problems with aspect ratios, sometimes slow, plots don't look as good as those produced by other libraries...)
- the `view_*` functions dump plots in an html page so there no way to export to static image formats such as png (except manually clicking on a button after opening the page in a browser)

to easily get nice surface images we can add an **optional** dependency on `plotly` and `kaleido` (if there are suggestions for other libraries **easily installed** with `pip` we could also consider them)

One way to insert these new plots would be to add a keyword argument to the existing `plot_surf*` functions, eg `plotting_engine="matplotlib"` (default), or `"plotly"`.

as a starting point, here is a draft of how a plotly figure for a surface image can be created and saved:

```python
"""requirements: nilearn, plotly, kaleido"""
import plotly.graph_objects as go

from nilearn import datasets, surface
from nilearn.plotting.cm import cold_hot
from nilearn.plotting.js_plotting_utils import colorscale
from nilearn.plotting.html_surface import _get_vertexcolor

AXIS_CONFIG = {
    "showgrid": False,
    "showline": False,
    "ticks": "",
    "title": "",
    "showticklabels": False,
    "zeroline": False,
    "showspikes": False,
    "spikesides": False,
    "showbackground": False,
}

CAMERAS = {
    "left": {
        "eye": {"x": -1.7, "y": 0, "z": 0},
        "up": {"x": 0, "y": 0, "z": 1},
        "center": {"x": 0, "y": 0, "z": 0},
    },
    "right": {
        "eye": {"x": 1.7, "y": 0, "z": 0},
        "up": {"x": 0, "y": 0, "z": 1},
        "center": {"x": 0, "y": 0, "z": 0},
    },
    "top": {
        "eye": {"x": 0, "y": 0, "z": 1.7},
        "up": {"x": 0, "y": 1, "z": 0},
        "center": {"x": 0, "y": 0, "z": 0},
    },
    "bottom": {
        "eye": {"x": 0, "y": 0, "z": -1.7},
        "up": {"x": 0, "y": 1, "z": 0},
        "center": {"x": 0, "y": 0, "z": 0},
    },
    "front": {
        "eye": {"x": 0, "y": 1.7, "z": 0},
        "up": {"x": 0, "y": 0, "z": 1},
        "center": {"x": 0, "y": 0, "z": 0},
    },
    "back": {
        "eye": {"x": 0, "y": -1.7, "z": 0},
        "up": {"x": 0, "y": 0, "z": 1},
        "center": {"x": 0, "y": 0, "z": 0},
    },
}
LAYOUT = {
    "scene": {f"{dim}axis": AXIS_CONFIG for dim in ("x", "y", "z")},
    "paper_bgcolor": "#fff",
    "hovermode": False,
    "margin": {"l": 0, "r": 0, "b": 0, "t": 0, "pad": 0},
}


def plot_surf(mesh, data, view="right", threshold="85%", output_file=None):
    coords, triangles = surface.load_surf_mesh(mesh)
    x, y, z = coords.T
    i, j, k = triangles.T
    colors = colorscale(cold_hot, data, threshold)
    vertexcolor = _get_vertexcolor(
        data,
        colors["cmap"],
        colors["norm"],
        colors["abs_threshold"],
        surface.load_surf_data(fsaverage["sulc_right"]),
    )
    mesh_3d = go.Mesh3d(x=x, y=y, z=z, i=i, j=j, k=k, vertexcolor=vertexcolor)
    fig = go.Figure(data=[mesh_3d])
    fig.update_layout(scene_camera=CAMERAS[view], **LAYOUT)
    if output_file is not None:
        fig.write_image(output_file)
    return fig


if __name__ == "__main__":
    img = datasets.fetch_neurovault_motor_task()["images"][0]
    mesh = fsaverage["pial_right"]
    data = surface.vol_to_surf(img, fsaverage["pial_right"])
    fig = plot_surf(mesh, data, output_file="/tmp/surface_plot.png")
    fig.show()
```

surface_plot

NicolasGensollen:
Fixes #2793 

**Description**

Very WIP...

This PR aims at providing the possibility to plot surfaces with both _matplotlib_ and _plotly_.
The choice for the user is made through the argument `engine` which can be either set to 'matplotlib' (default) or 'plotly' (new).
This PR also does some refactoring of `plot_surf` in order to facilitate the implementation.
The implementation of the plotly part is based on the code provided by @jeromedockes in #2793.

**Example**

```python
from nilearn import plotting, surface, datasets
 
texture = surface.vol_to_surf(stat_img, fsaverage.pial_left)
surf = surface.load_surf_mesh(fsaverage.pial_left)
 
fig = plotting.plot_surf(surf, texture,
                         hemi='left', view='lateral',
                         colorbar=True,
                         engine='plotly')
fig.show()
```

surf_plotly

**Current state**

For now this should work:

- with and without a provided texture
- with all combinations of `view` and `hemi`
- with `colorbar` set to `True` or `False`

**TODO**

- implement other configurations (thresholding, etc...)
- enable saving to files
- add tests

@alexisthual @jeromedockes since you probably have more experience with plotly than I have, let me know if you have comments or advices on how this is going (keeping in mind that it is not finished... :wink: )


jeromedockes:
> @alexisthual @jeromedockes since you probably have more experience with plotly than I have, let me know if you have comments or advices on how this is going (keeping in mind that it is not finished... :wink: )

I think you're on the right track! thanks a lot for refactoring the
`plot_surf` code. one detail: the plotly imports should be hidden inside
`_plot_surf_plotly` to avoid importerrors when it is not installed (we
said it would be an optional dependency)

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+            coords.shape[0], faces.shape[0]
+        )
+        colors = colorscale(cmap, surf_map_faces, threshold)
+        vertexcolor = _get_vertexcolor(
+            surf_map_faces,
+            colors["cmap"],
+            colors["norm"],
+            colors["abs_threshold"],
+            bg_map=bg_map,
+        )
```
alexisthual, line 145: Contrary to `matplotlib`, `plotly` can assign colours to vertices instead of faces (and interpolate the former correctly), so I would probably just stick to @jeromedockes' example implementation and not use `_compute_surf_map_faces` here.
This might yield different images depending on what engine you use, but I think this new setup is way more intuitive (in fact I tried to alleviate this in #2790 by adding more `avg_method`s but this PR would do a much better job).

```suggestion
        colors = colorscale(cmap, surf_map, threshold)
        vertexcolor = _get_vertexcolor(
            surf_map,
            colors["cmap"],
            colors["norm"],
            colors["abs_threshold"],
            bg_map=bg_map,
        )
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+    else:
+        vertexcolor = None
+    if colorbar:
+        mesh_3d = go.Mesh3d(x=x, y=y, z=z, i=i, j=j, k=k,
+                            intensity=surf_map_faces,
```
alexisthual, line 150:
```suggestion
                            intensity=surf_map,
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+        raise ValueError(f"hemi must be one of {VALID_HEMISPHERES}")
+    return elev, azim
+
+
+def _compute_surf_map_faces(surf_map, faces, avg_method, n_vertices,
+                            face_colors_size):
+    """Helper function for plot_surf.
+
+    This function computes the surf map faces using the
+    provided averaging method.
```
alexisthual, line 209:
```suggestion
    provided averaging method.
    This method is called exclusively when using matplotlib,
    since it only supports plotting face-colour maps
    and not vertex-colour maps.
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+    "zeroline": False,
+    "showspikes": False,
+    "spikesides": False,
+    "showbackground": False,
+}
+
+
+CAMERAS = {
+    "left": {
+        "eye": {"x": -1.7, "y": 0, "z": 0},
```
alexisthual, line 51: `1.7` -> `1.5` allows for less padding and still fits for all views
```suggestion
        "eye": {"x": -1.5, "y": 0, "z": 0},
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+        "up": {"x": 0, "y": 0, "z": 1},
+        "center": {"x": 0, "y": 0, "z": 0},
+    },
+    "right": {
+        "eye": {"x": 1.7, "y": 0, "z": 0},
```
alexisthual, line 56:
```suggestion
        "eye": {"x": 1.5, "y": 0, "z": 0},
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+        "up": {"x": 0, "y": 0, "z": 1},
+        "center": {"x": 0, "y": 0, "z": 0},
+    },
+    "top": {
+        "eye": {"x": 0, "y": 0, "z": 1.7},
```
alexisthual, line 61:
```suggestion
        "eye": {"x": 0, "y": 0, "z": 1.5},
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+        "up": {"x": 0, "y": 1, "z": 0},
+        "center": {"x": 0, "y": 0, "z": 0},
+    },
+    "bottom": {
+        "eye": {"x": 0, "y": 0, "z": -1.7},
```
alexisthual, line 66:
```suggestion
        "eye": {"x": 0, "y": 0, "z": -1.5},
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+        "up": {"x": 0, "y": 1, "z": 0},
+        "center": {"x": 0, "y": 0, "z": 0},
+    },
+    "front": {
+        "eye": {"x": 0, "y": 1.7, "z": 0},
```
alexisthual, line 71:
```suggestion
        "eye": {"x": 0, "y": 1.5, "z": 0},
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+        "up": {"x": 0, "y": 0, "z": 1},
+        "center": {"x": 0, "y": 0, "z": 0},
+    },
+    "back": {
+        "eye": {"x": 0, "y": -1.7, "z": 0},
```
alexisthual, line 76:
```suggestion
        "eye": {"x": 0, "y": -1.5, "z": 0},
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +33,416 @@
...
+        elif view == 'posterior':
+            elev, azim = 0, 270
+        else:
+            raise ValueError(f"view must be one of {VALID_VIEWS}")
+    else:
+        raise ValueError(f"hemi must be one of {VALID_HEMISPHERES}")
+    return elev, azim
+
+
+def _compute_surf_map_faces(surf_map, faces, avg_method, n_vertices,
```
alexisthual, line 204: Maybe change the name, indicating that it is a matplotlib-only helper

alexisthual (commented):
Niiice! Next step is probably handling `vmin`, `vmax` and friends :racehorse:

nilearn/plotting/surf_plotting.py
jeromedockes: +1, plot the values of vertices and do not compute values for faces

> This might yield different images depending on what engine you use

yes that's the goal :) -- getting better images than with the matplotlib engine

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +32,489 @@
...
+    face_colors = plt.cm.gray_r(bg_faces)
+    # modify alpha values of background
+    face_colors[:, 3] = alpha * face_colors[:, 3]
+    return face_colors
+
+
+def _threshold_and_rescale(data, threshold, vmin, vmax):
+    """Helper function for plot_surf.
+
+    This function threshold and rescale the provided data.
```
bthirion, line 392:
```suggestion
    This function thresholds and rescales the provided data.
```

nilearn/plotting/surf_plotting.py
```diff
@@ -67,10 +549,23 @@ def plot_surf(surf_mesh, surf_map=None, bg_map=None,
...
+        .. note::
+            To use 'plotly' and save figures to disk you should
+            have both `plotly` and `kaleido` installed.
+
+        Default='matplotlib'.
+
+    cmap : matplotlib/plotly colormap, str or colormap object, optional
+        To use for plotting of the stat_map. Either a string which is a
+        name of a colormap, or a colormap object.
+        If None, matplotlib/plotly default will be chosen.
```
bthirion, line 568:
```suggestion
        If None, matplotlib/plotly default will be chosen.
``` matplotlib for bw compatibility ?

File: nilearn/plotting/surf_plotting.py [hunk 13]
```diff
@@ -29,10 +31,502 @@
...
+def _configure_cmap_matplotlib(cmap, vmin, vmax, threshold,
+                               cbar_tick_format):
+    """Helper function for plot_surf with matplotlib engine.
+
+    This function configure the colormap.
+    """
+    our_cmap = get_cmap(cmap)
+    norm = Normalize(vmin=vmin, vmax=vmax)
+    # Default number of ticks is 5...
+    nb_ticks = 5
```
bthirion, line 352:
```suggestion
    n_ticks = 5
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +31,502 @@
...
+            cmaplist[i] = (0.5, 0.5, 0.5, 1.)
+        our_cmap = LinearSegmentedColormap.from_list(
+            'Custom cmap', cmaplist, our_cmap.N)
+    return our_cmap, norm, ticks, bounds
+
+
+def _compute_facecolors(bg_map, faces, n_vertices, darkness, alpha):
+    """Helper function for plot_surf.
+
+    This function computes the facecolors.
```
bthirion, line 376: Used for matplotlib rendering only, right ?

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +31,502 @@
...
+    if vmax is None:
+        vmax = np.nanmax(data_copy)
+    # treshold if indicated
+    if threshold is None:
+        # If no thresholding and nans, filter them out
+        kept_indices = np.where(
+            np.logical_not(
+                np.isnan(data_copy)))[0]
+    else:
+        kept_indices = np.where(np.abs(data_copy) >= threshold)[0]
```
bthirion, line 419: why is interesting to have `kept_indices` as indices rather than a binary vector ?

bthirion (commented):
Looks well-advanced now !
Beware the few untested lines.
+ whatsnew entry

bthirion:
Question: how do you save the image non-interactively with plotly ?

alexisthual:
> Question: how do you save the image non-interactively with plotly ?

In replacement of `plt.savefig(output_path)` for matplotlib figures, you can do `fig.write_image(output_path)` with plotly.
Users won't be able to use plotly with their existing code by simply adding `engine="plotly"`, they'll also need to refactor saving and subplotting parts.
@NicolasGensollen and I discussed adding a tutorial on the subject, so that people have an example to start from.

jeromedockes:
> In replacement of `plt.savefig(output_path)` for matplotlib figures, you can do `fig.write_image(output_path)` with plotly.
> Users won't be able to use plotly with their existing code by simply adding `engine="plotly"`, they'll also need to refactor saving and subplotting parts.
> @NicolasGensollen and I discussed adding a tutorial on the subject, so that people have an example to start from.

- there would still be the `output_file` parameter that can be used for
  both backends, right?
- also some wrapper class could be used to provide a unified interface
  to the objects returned in both cases. Note that for example all the
  `plot_stat_map`, `plot_img` & co don't return a matplotlib figure but
  an object of a type defined by nilearn
- of course as you point out plt.savefig will only work with matplotlib
  figures. note this won't break any existing code if matplotlib remains
  the default backend

nilearn/plotting/surf_plotting.py
NicolasGensollen: Thanks @bthirion ! When I wrote it I wasn't sure if it would be matplotlib-specific or not.
As for this PR, it is only used when the engine is matplotlib since we are coloring by points rather than faces when using plotly. I will thus update the name and docstring.
That being said, I realized recently that it _could_ still be useful to have the possibility to color by faces with plotly. For example, when trying to extend to `plot_surf_contours`, it seems much easier and faster to draw the contours by coloring faces than by coloring points. I can explain in more details why I think that's the case (I might also be wrong here...).

NicolasGensollen:
@alexisthual @jeromedockes Assuming you have the dependencies, if you provide `output_path` to `plot_surf` the figure is saved whatever the engine. It is handled separately in each sub-function and is not a concern for the user.

But, @jeromedockes is right that the return object of the function isn't the same depending on the value of the plotting engine which can potentially cause confusion. 
I like the idea of a wrapper class providing a unified interface. I'm not sure we want to do that on this PR as it is already quite large, but it could be a good next step.

jeromedockes:
> I'm not sure we want to do that on this PR as it is already quite large, but it could be a good next step.

sure, dealing with it once this PR is merged sounds good!

jeromedockes:
> That being said, I realized recently that it _could_ still be useful to have the possibility to color by faces with plotly. For example, when trying to extend to `plot_surf_contours`, it seems much easier and faster to draw the contours by coloring faces than by coloring points. I can explain in more details why I think that's the case (I might also be wrong here...).

The signals are provided as values at each node coordinates so plotting
them that way is the most sensible.

I agree with you that painting whole triangles that touch the border of
a region sounds easier than computing the contour. It is more of a
"quick and dirty" way though, and borders look weird, we have no control
over line width, etc. so if we could avoid it that would be great. We
should look into how this is implemented elsewhere, eg in mne.

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +31,502 @@
...
+        _check_surf_map(surf_map, coords.shape[0])
+        colors, our_cmap, norm, vmin, vmax = _colorscale_plotly(
+            cmap, surf_map, threshold, vmin, vmax
+        )
+        if not colorbar:
+            vertexcolor = _get_vertexcolor(
+                surf_map, our_cmap, norm, None, bg_map
+            )
+        else:
+            mesh_3d = go.Mesh3d(x=x, y=y, z=z, i=i, j=j, k=k,
```
jeromedockes, line 215: couldn't we use `vertexcolor` in all cases, even when a colorbar is needed?

jeromedockes: the background map seems to be ignored in this case

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +31,502 @@
...
+            raise ValueError('The bg_map does not have the same number '
+                             'of vertices as the mesh.')
+    if surf_map is not None:
+        _check_surf_map(surf_map, coords.shape[0])
+        colors, our_cmap, norm, vmin, vmax = _colorscale_plotly(
+            cmap, surf_map, threshold, vmin, vmax
+        )
+        if not colorbar:
+            vertexcolor = _get_vertexcolor(
+                surf_map, our_cmap, norm, None, bg_map
```
jeromedockes, line 212: here we are always passing `None` as the threshold and as a result the background map is never used

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +31,502 @@
...
+    return fig
+
+
+def _set_view_plot_surf_matplotlib(hemi, view):
+    """Helper function for plot_surf with matplotlib engine.
+
+    This function checks the selected hemisphere and view, and
+    returns elev and azim.
+    """
+    if hemi == 'right':
```
jeromedockes, line 238: maybe we could use dictionaries instead of the chained elif?

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +31,502 @@
...
+    .. note::
+        This method is called exclusively when using matplotlib,
+        since it only supports plotting face-colour maps and not
+        vertex-colour maps.
+
+    """
+    surf_map_data = _check_surf_map(surf_map, n_vertices)
+
+    # create face values from vertex values by selected avg methods
+    if avg_method == 'mean':
```
jeromedockes, line 304: maybe `getattr(np, avg_method)` would be shorter and more general?

nilearn/plotting/surf_plotting.py
jeromedockes: maybe having "matplotlib" in the name or at least the docstring would make it a bit easier to identify when this is used. most likely we will not want to color faces except when forced to do so when using matplotlib. but anyway we can always rename this function later if needed as it is "private"

jeromedockes (commented):
first pass of review, I would stick to using `get_vertexcolor` whether a colorbar is required or not, unless there is a reason not to do so, and make sure to pass the background map and threshold

nilearn/plotting/surf_plotting.py
NicolasGensollen: >couldn't we use vertexcolor in all cases, even when a colorbar is needed?

At this point I am unsuccessful at displaying a colorbar with vertexcolor in `mesh3d`. I might have missed something, but it seems to work only when setting the `intensity` and `colorscale` attributes... This is not a problem in most cases, but when having both a thresholded stat map and a background, we need to combine the colors of both the map and background, which is exactly what the function `_get_vertexcolor` does. I'm not sure how to do the same thing through intensity and colorscale though.
Pinging @alexisthual who might have an idea here

nilearn/plotting/surf_plotting.py
```diff
@@ -29,13 +32,557 @@
...
+                     "color": "RebeccaPurple",
+                     },
+            "y": 0.96,
+            "x": 0.5,
+            "xanchor": "center",
+            "yanchor": "top"}
+
+
+def _get_colors(data, vmin, vmax, threshold, cmap, symmetric_cmap):
+    """Helper function for _plot_surf_plotly.
```
bthirion, line 237: Shouldn't the name indicate the relation to plotly ?

NicolasGensollen:
@jeromedockes @alexisthual here is the [link](https://community.plotly.com/t/display-a-custom-colorbar-with-mesh3d-and-vertexcolor/56202) to the post I made on the plotly support forum about the issue with `vertexcolor` and colorbars that we currently have in this PR.
Feel free to have a look and tell me if I didn't explain the issue well enough.

NicolasGensollen:
There are still a few things to improve (especially on stylizing the colorbar a little bit more...), but I think this is ready for another round of reviews.
I managed to have the background map displayed correctly with the colorbar and the stat map, and I refactored the code a little bit to make the plotly function much shorter and simpler.

You can try it like this:

```python
from nilearn import plotting, surface, datasets
   
texture = surface.vol_to_surf(stat_img, fsaverage.pial_left)
surf = surface.load_surf_mesh(fsaverage.pial_left)
   
fig = plotting.plot_surf(surf, texture,
                         hemi='left', view='lateral',
                         colorbar=True,
                         threshold=2.,
                         bg_map=fsaverage['sulc_left'],
                         title='Threshold + bg_map',
                         engine='plotly')
fig.show()
```
plotly3

nilearn/plotting/surf_plotting.py
```diff
@@ -29,13 +31,571 @@
...
+        raise ValueError(f"view must be one of {VALID_VIEWS}")
+    return MATPLOTLIB_VIEWS[hemi][view]
+
+
+def _check_surf_map(surf_map, n_vertices):
+    """Helper function for plot_surf.
+
+    This function checks the dimensions of provided surf_map.
+    """
+    surf_map_data = load_surf_data(surf_map)
```
bthirion, line 361: I guess this flattens surf_map when shape is (n_vertices, 1), right ?

bthirion (commented):
Looks great thx !
Note that I have a couple of pending comments.

nilearn/plotting/surf_plotting.py
jeromedockes: > At this point I am unsuccessful at displaying a colorbar with vertexcolor in `mesh3d`. I might have missed something, but it seems to work only when setting the `intensity` and `colorscale` attributes...

when we specify a color for each node rather than a colormap and values plotly does not know how to create a colorbar, but we can easily add it by making an invisible plot using that colormap. This is what is done for the current nilearn plots: https://github.com/nilearn/nilearn/blob/cc99fa76df137e9ac6010c311e15fa9ac12ce66d/nilearn/plotting/data/js/surface-plot-utils.js#L144
so the current approach is to use this trick and vertexcolors in all cases, which allows to handle everything in the same way and have a background map and a colorbar easily.

nilearn/plotting/surf_plotting.py
```diff
@@ -29,13 +32,557 @@
...
+    if threshold is not None:
+        if vmin is not None:
+            warnings.warn('choosing both vmin and a threshold is not allowed; '
+                          'setting vmin to 0')
+            vmin = 0
+    if vmax is None:
+        vmax = np.nanmax(abs_values)
+    vmax = float(vmax)
+    vmin = - vmax
+    if vmin is None:
```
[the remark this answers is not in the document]
NicolasGensollen: @bthirion I agree, this function was a non-sense resulting from my attempt to refactor too much code within this PR. I got rid of these complicated bound functions and pushed something simpler. I realized that `plot_surf` is not supposed to play with `vmin`, `vmax`, and `symmetric_cbar`: this is done in `plot_surf_stat_map` (for example) prior to the call to `plot_surf` with the proper values for `vmin` and `vmax`. So we don't have to do much here except computing them from data if they were set to None.
For the moment I don't know how to use `cbar_vmin` and `cbar_vmax` to "crop" the colorbar in the right way. Once we find out how to do that, the behavior should be the same for both engines.
Hope this is clear enough...

nilearn/plotting/surf_plotting.py
NicolasGensollen: @bthirion I agree. In the majority of cases it should be more compact and efficient to use a binary vector here (and the code is easier to read as a bonus...).

nilearn/plotting/surf_plotting.py
```diff
@@ -29,10 +31,502 @@
...
+    """Helper function for plot_surf with plotly engine.
+
+    This function returns the colorscale for a given cmap
+    and thresholding configuration.
+
+    .. note::
+        This function does not enforce symmetric cmap.
+
+    """
+    if vmin is None:
```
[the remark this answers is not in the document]
NicolasGensollen: @jeromedockes I had a go at this but I was not satisfied with the result I got. :sweat: 
You're right though, there is definitely some refactoring to do with this colorscale/colorbar code. I will put that on my todo list, but it'd be easier to do it (and review it...) in a separate PR I think. 
WDYT?

nilearn/plotting/surf_plotting.py
alexisthual: Yes indeed (I just checked what `nilearn.surface.load_surf_data` does, and it eventually calls `np.squeeze()` on the returned array).

nilearn/plotting/surf_plotting.py
```diff
@@ -29,13 +30,535 @@
...
+def _threshold(data, threshold):
+    # If no thresholding and nans, filter them out
+    return (np.logical_not(np.isnan(data)) if threshold is None
+            else np.abs(data) >= threshold)
+
+
+def _rescale(data, vmin, vmax):
+    data_copy = np.copy(data)
+    # if no vmin/vmax are passed figure them out from data
+    vmin, vmax = _get_bounds(data_copy, vmin, vmax)
```
alexisthual, line 444: Just a minor change, but setting `vmin=None, vmax=None` by default could prove more handy in the future, and feels more consistent with other util functions.
```suggestion
def _rescale(data, vmin=None, vmax=None):
    data_copy = np.copy(data)
    # if no vmin/vmax are passed figure them out from data
    vmin, vmax = _get_bounds(data_copy, vmin, vmax)
```

nilearn/plotting/surf_plotting.py
```diff
@@ -29,13 +30,535 @@
...
+    data_copy -= vmin
+    data_copy /= (vmax - vmin)
+    return data_copy, vmin, vmax
+
+
+def _inverse_rescale(data, vmin, vmax):
+    data_copy = np.copy(data)
```
alexisthual, line 451: To be consistent with `_rescale()`:
```suggestion
def _inverse_rescale(data, vmin=None, vmax=None):
    data_copy = np.copy(data)
    # if no vmin/vmax are passed figure them out from data
    vmin, vmax = _get_bounds(data_copy, vmin, vmax)
```

alexisthual (approved):
Looks great to me :champagne:

nilearn/plotting/surf_plotting.py
alexisthual: Yes indeed (I just checked what `nilearn.surface.load_surf_data` does, and it eventually calls `np.squeeze()` on the returned array).

nilearn/plotting/surf_plotting.py
NicolasGensollen: Thanks for the review @alexisthual ! :+1:

jeromedockes:
thanks for addressing these last comments! summarizing here what we should do in follow-up prs:
- return a nilearn-defined wrapper object rather than the plotly figure directly
- make sure the brain never appears cropped in the saved png 
- (somewhat low-priority): harmonize cbar tick formats across both backends (use str.format syntax)
- showcase using the plotly backend in the example gallery

I think this pr is ready to merge, @NicolasGensollen did you have more changes to push or can I merge it?

bthirion:
Great, congrats !

PR #2909, opened 2021-08-03
[DOC] Add intersphinx link for joblib

NicolasGensollen:

PR #2911, opened 2021-08-03
[DOC] Improve docstrings of `regions` module

NicolasGensollen:
This PR proposes to improve the docstrings of the `regions` module by:

- using standard entries when possible
- adding cross-references
- doing some formatting enhancements

Note: This PR requires #2909 to be merged first in order for the Joblib references to work.


nilearn/_utils/docs.py
```diff
@@ -54,28 +54,34 @@
 docdict['smoothing_fwhm'] = """
 smoothing_fwhm : :obj:`float`, optional.
     If ``smoothing_fwhm`` is not ``None``, it gives
-    the size in millimeters of the spatial smoothing
-    to apply to the signal. Default=None."""
+    the full-width half maximum size in millimeters
```
bthirion, line 57:
```suggestion
    the full-width at half maximum in millimeters
```

nilearn/_utils/docs.py
```diff
@@ -87,32 +93,48 @@
...
+memory : instance of :class:`joblib.Memory` or :obj:`str`
     Used to cache the masking process.
-    By default, no caching is done. If a ``str`` is given, it is the
+    By default, no caching is done. If a :obj:`str` is given, it is the
     path to the caching directory."""
 
+# n_parcels
+docdict['n_parcels'] = """
+n_parcels : :obj:`int`, optional
+    Number of parcellations to divide the brain data into.
```
bthirion, line 127:
```suggestion
    Number of parcels to divide the data into.
```

bthirion (commented):
Mostly good, thx !

nilearn/_utils/docs.py
```diff
@@ -163,6 +191,22 @@
     are .png, .pdf, .svg. If ``output_file`` is not None, the plot
     is saved to a file, and the display is closed."""
 
+# extractor / extract_type
+docdict['extractor'] = """
+extractor : :obj:`str` {'local_regions', 'connected_components'}, optional
```
tsalo, line 196: I think numpydoc convention is just to have the options, without the data type. See [here](https://numpydoc.readthedocs.io/en/latest/format.html):

>When a parameter can only assume one of a fixed set of values, those values can be listed in braces, with the default appearing first:
>
> ```
> order : {'C', 'F', 'A'}
>     Description of `order`.
> ```

tsalo (changes requested):
I have a few ideas for improvements, but they're mostly pretty minor, so feel free to take or leave them.

nilearn/regions/parcellations.py
```diff
@@ -194,48 +202,35 @@ class Parcellations(MultiPCA):
...
-        'all CPUs', -2 'all CPUs but one', and so on.
-        Default=1.
-
-    verbose : integer, optional
-        Indicate the level of verbosity. By default, nothing is printed.
-        Default=0.
+    %(memory)s
+    %(memory_level)s
+    %(n_jobs)s
+    %(verbose0)s
```
bthirion, line 222: Sorry, but why `verbose0` instead of  `verbose` ?

nilearn/regions/parcellations.py
```diff
@@ -446,48 +446,56 @@ def transform(self, imgs, confounds=None):
...
+            given as a list.
+
+            .. note::
+                This parameter is passed to :func:`nilearn.signal.clean`.
+                Please see the related documentation for details.
 
-            Note: same confounds will used for cleaning signals before
-            learning parcellations.
+            .. note::
+                Same confounds will used for cleaning signals before
```
bthirion, line 471:
```suggestion
                Confounds will be used for cleaning signals before
```

nilearn/regions/region_extractor.py
```diff
@@ -142,21 +145,13 @@ def connected_regions(maps_img, min_region_size=1350,
         An image of brain activation or atlas maps to be extracted into set of
         separate brain regions.
 
-    min_region_size : int, optional
+    min_region_size : :obj:`int`, optional
```
bthirion, line 148: Should be a float

NicolasGensollen: Indeed, thanks!

nilearn/regions/parcellations.py
NicolasGensollen: Because there are lots of verbose arguments defaulting to 0 (but most often to 1), I created two entries in the standard doc dict:

https://github.com/nilearn/nilearn/blob/65200aa3c81a7e70914ba85c4115c483f7510e8f/nilearn/_utils/docs.py#L26-L31

Instead of having the default added in each docstring.

bthirion (approved):
LGTM, thx !

PR #2912, opened 2021-08-04
[MAINT] Fix CircleCI

NicolasGensollen:
Not sure yet what the issue is, investigating...


KamalakerDadi:
I ran the build locally and it seems fine. Tricky one to detect.

NicolasGensollen:
>I ran the build locally and it seems fine. Tricky one to detect.

Same for me, everything runs fine locally. 
I thought that it could be a memory issue due to some potential changes on CircleCI's side, but when running `plot_haxby_mass_univariate` alone on CircleCI, it runs fine. 

So far I have no idea what's causing this failure. :thinking:

KamalakerDadi:
Do you have the same packaging versions as in CircleCI?

KamalakerDadi:
Can you remove block by block?

Can you test removing from this block?
```python
##############################################################################
# Perform massively univariate analysis with permuted OLS
#
```

NicolasGensollen:
>Do you have the same packaging versions as in CircleCI?

Yes. I just made a brand new env and launch the build locally to be 100% sure.

>Can you remove block by block?
>Can you test removing from this block?

Since we have to wait around 1h30 every time, I did the reverse, I removed everything and will add block by block as long as it succeeds (assuming it does with the almost empty example I'm currently running...). Greedy technique but I don't have a better idea at this point...

NicolasGensollen:
Looks like the issue is happening here, at transform time:

https://github.com/nilearn/nilearn/blob/28bf634ae5133339c9e66bc31c5f2583553d5f12/examples/07_advanced/plot_haxby_mass_univariate.py#L45-L50

KamalakerDadi:
I guess so as it fitted and then bursted at transform.

Can we use ```index_img``` to restrict samples to the conditions mask then fit NiftiMasker?
For example, examples/02_decoding/plot_haxby_anova_svm.py does this.

NicolasGensollen:
>Can we use index_img to restrict samples to the conditions mask then fit NiftiMasker?

@KamalakerDadi this is indeed what I'm trying currently. It worked with the image sliced up to 800 in the last dimension (over 1452). I'll continue to increase and see when it fails...

NicolasGensollen:
Merging....

PR #2918, opened 2021-08-16
[DOC] Typos found by codespell

DimitriPapadopoulos:
Fix typos found by [codespell](https://github.com/codespell-project/codespell).

### DIFF
--- a/examples/01_plotting/plot_overlay.py
+++ b/examples/01_plotting/plot_overlay.py
@@ -1,8 +1,8 @@
 """
-Visualizing a probablistic atlas: the default mode in the MSDL atlas
+Visualizing a probabilistic atlas: the default mode in the MSDL atlas
 =====================================================================
 
-Visualizing a probablistic atlas requires visualizing the different
+Visualizing a probabilistic atlas requires visualizing the different
 maps that compose it.
 
 Here we represent the nodes constituting the default mode network in the
@@ -52,8 +52,8 @@ plotting.show()
 
 
 ###############################################################################
-# Visualizing a probablistic atlas with plot_prob_atlas
-# =====================================================
+# Visualizing a probabilistic atlas with plot_prob_atlas
+# ======================================================
 #
 # Alternatively, we can create a new 4D-image by selecting the 3rd, 4th, 5th and 6th (zero-based) probabilistic map from atlas
 # via :func:`nilearn.image.index_img` and use :func:`nilearn.plotting.plot_prob_atlas` (added in version 0.2)
--- a/examples/04_glm_first_level/plot_hrf.py
+++ b/examples/04_glm_first_level/plot_hrf.py
@@ -1,9 +1,9 @@
 """Example of hemodynamic response functions.
 ==========================================
 
-Within this example we are going to plot the hemodynamic response function (hrf)
-model in SPM together with the hrf shape proposed by G.Glover, as well as their
-time and dispersion derivatives.
+Within this example we are going to plot the hemodynamic response function
+(hrf) model in SPM together with the hrf shape proposed by G.Glover, as well as
+their time and dispersion derivatives.
 
 This example requires matplotlib.
 
--- a/examples/04_glm_first_level/plot_hrf.py
+++ b/examples/04_glm_first_level/plot_hrf.py
@@ -1,7 +1,7 @@
-"""Example of hemodynamic reponse functions.
-=========================================
+"""Example of hemodynamic response functions.
+==========================================
 
-Within this example we are going to plot the hemodynamic reponse function (hrf)
+Within this example we are going to plot the hemodynamic response function (hrf)
 model in SPM together with the hrf shape proposed by G.Glover, as well as their
 time and dispersion derivatives.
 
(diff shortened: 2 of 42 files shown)


DimitriPapadopoulos:
I suggest you disregard the PEP8 error. It's from a previous PR and fixing it will only add noise to this PR.

examples/01_plotting/plot_overlay.py
```diff
@@ -52,7 +52,7 @@
 
 
 ###############################################################################
-# Visualizing a probablistic atlas with plot_prob_atlas
+# Visualizing a probabilistic atlas with plot_prob_atlas
 # =====================================================
```
agramfort, line 56:
```suggestion
# ======================================================
```

examples/04_glm_first_level/plot_hrf.py
```diff
@@ -1,7 +1,7 @@
-"""Example of hemodynamic reponse functions.
+"""Example of hemodynamic response functions.
 =========================================
```
agramfort, line 2:
```suggestion
"""Example of hemodynamic response functions.
==========================================
```

agramfort (commented):
-1 to merge anything that breaks the CI.

@DimitriPapadopoulos can you fix the pep8 errors?

also it would be great to add codespell to the CIs to avoid future typos.

examples/01_plotting/plot_overlay.py
DimitriPapadopoulos: Fixed, but there are such issues all over the place, to the point it feels no such rule is really enforced.

examples/04_glm_first_level/plot_hrf.py
agramfort: yes pep8 does break the CI. It makes the PRs red on merge. There is pep8 action that we cannot ignore.

Regarding the RST fix with the missing `=` it is RestructedText syntax that enforces that section subtitles have the right number of = or - or ^ etc.

DimitriPapadopoulos: I don't see how I break the CI. The line of code hasn't changed, except for a fixed typo **and** less characters per line than before. The issue raised by the CI precedes my patch, so it cannot possibly break the CI more than the previous code. What do you mean by "we cannot ignore", as far as I know it's entirely possible to bypass the CI if needed when merging.

agramfort (commented):
@DimitriPapadopoulos please fix the pep8 error and then +1 for merge.

there reason you need to fix the pep8 here is that pep8 is progressively enforced on diffs (to avoid a big PR with a huge diff just for pep8).

thanks for address my concerns about the .rst issues.

examples/04_glm_first_level/plot_hrf.py
DimitriPapadopoulos: Indeed, the [reStructedText specification](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#sections) states:
> Sections are identified through their titles, which are marked up with adornment: "underlines" below the title text, or underlines and matching "overlines" above the title. An underline/overline is a single repeated punctuation character that begins in column 1 and forms a line extending at least as far as the right edge of the title text.

But then I'm not certain what "_extending at least as far as the right edge of the title text_" means. Exact number of characters or visible length in any editor with a possibly variable-width font? I'm asking because while only the first interpretation makes sense, the rule is (almost) never enforced in nibabel. Actually, the suggested fix doesn't fix anything, the underline was too short before the commit and remains too short after the commit.

DimitriPapadopoulos:
I have fixed the pep8 issue, but I doubt pep8 errors can be fixed by merely enforcing the rule on patches before the 22nd century.:-) Perhaps a huge diff would be better, maybe in a few years time.

examples/04_glm_first_level/plot_hrf.py
agramfort: yes it must be at least the number of characters. If it's too short please fix it. Thanks a lot @DimitriPapadopoulos

agramfort (approved):

examples/04_glm_first_level/plot_hrf.py
jeromedockes: > Actually, the suggested fix doesn't fix anything, the underline was too short before the commit and remains too short after the commit.


the spell fix did add a character to the title, making the underline, which _was_ the right length before, too short. the fix proposed by @agramfort does fix the issue. Note the `"""` that enclose a python literal string are not part of the docstring itself. I understand the frustration, rst syntax can be quite annoying and the error messages are not always helpful.

DimitriPapadopoulos: Indeed, but indentation doesn't help, I would change:
```
"""Example of hemodynamic reponse functions.
=========================================
```
to:
```
"""
Example of hemodynamic reponse functions.
=========================================
```

And reST-like incosistent syntax in non-reST text doesn't help either:
```
############################################################################
# Fetching probabilistic atlas - MSDL atlas
# -----------------------------------------
[...]
#############################################################################
# Visualizing a probabilistic atlas with plot_stat_map and add_overlay object
# ---------------------------------------------------------------------------

[...]
# Visualizing a probablistic atlas with plot_prob_atlas
# =====================================================
```

DimitriPapadopoulos: Anyway, the CI errors have been fixed.

jeromedockes:
thanks!

PR #2920, opened 2021-08-17
[FIX] LGTM warnings

DimitriPapadopoulos:
Fixes some LGTM warnings:
https://lgtm.com/projects/g/nilearn/nilearn/?mode=list&severity=warning

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -241,7 +241,7 @@ def _repr_niimgs(niimgs, shorten=True):
 
 
 def _short_repr(niimg_rep, shorten=True, truncate=20):
-    """Gives a shorten version on niimg representation
+    """Gives a shorter version of niimg representation
     """
     # Make sure truncate has a reasonable value
     truncate = max(truncate, 10)
@@ -263,7 +263,6 @@ def _short_repr(niimg_rep, shorten=True, truncate=20):
                     rep = str(Path("...", rep))
                     break
         return rep
-    return niimg_rep
 
 
 def img_data_dtype(niimg):


agramfort (approved):

nilearn/image/resampling.py
```diff
@@ -445,7 +445,7 @@ def resample_img(img, target_affine=None, target_shape=None,
         if copy and not input_img_is_string:
             img = _utils.copy_img(img)
         return img
-    if target_affine is affine and target_shape is shape:
+    if target_affine is affine and target_shape == shape:
```
jeromedockes, line 448: shouldn't this be `np.allclose(target_affine, affine) and np.array_equal(target_shape, shape)`?

DimitriPapadopoulos: Probably, if you say so :-) But I wouldn't know, I'll have to look much deeper in the code.

jeromedockes: I think so -- the `is` was probably causing the image to be uselessly resampled in some cases (although it is less likely to cause this than the one used to compare affines, due to interning because shapes are often specified as tuples)
but replacing it with `==` will still cause some needless resampling eg if the target_shape is given as a list, or result in a `ValueError` if `target_shape` happens to be a numpy array. comparing the affines with `is` will also result in identical affines being treated as different. so using the numpy functions for comparing them is probably better.

although what I suggested is missing the comparison of the affines' shapes, so it should be:
`np.shape(target_affine) == np.shape(affine) and np.allclose(target_affine, affine) and np.array_equal(target_shape, shape)` , or if we want exact comparison of the affine `np.array_equal(target_affine, affine) and np.array_equal(target_shape, shape)`

DimitriPapadopoulos: Perhaps defining an `__eq__` operator in the first place was an error? The semantics associated to thisclass seem more complex than just "equality"?

jeromedockes: when the user asks to resample to `target_affine` and `target_shape`, and these happen to already be the affine and shape of the input image, then no resampling is necessary. I believe this is what this line is checking. but comparisons are not as straightforward as using `==` because (i) `target_affine` and `target_shape` are user inputs that can be of different types such as list, tuple, numpy array, (ii) `==` for numpy arrays returns the result of elementwise comparison (an array of the same shape as the broadcasted operands, not a single boolean), (iii) the affine contains floating-point values so `==` can return False even for values that are the same up to machine precision. therefore comparing the shapes and then checking that all elements are equal up to machine precision for the floating-point affine, and checking that all elements are equal for the integer shape, is a less error-prone way of comparing the image's affine and shape to what the caller asked for. in the `main` branch the comparison is done with `is`, which probably results in some false negatives (but that is ok because `is` will not result in an exception, and a false negative will cause needlessly resampling the image but will still produce correct results)

jeromedockes, line 448:
```suggestion
    if (
        np.shape(target_affine) == np.shape(affine)
        and np.allclose(target_affine, affine)
        and np.array_equal(target_shape, shape)
    ):
```

jeromedockes: or if you are not convinced you can just revert the change to this line and we can deal with this in another pr

DimitriPapadopoulos: It's not that I'm no convinced. I just need to invest quite some time to learn the details of this module. Also, this does go beyond the initial scope of this PR, so I think I'll follow your advice and move that commit to a new PR.

DimitriPapadopoulos: See new PR #2925.

jeromedockes:
great, thank you! merging

PR #2921, opened 2021-08-17
[FIX] LGTM recommendations

DimitriPapadopoulos:
Fixes some LGTM recommendations:
https://lgtm.com/projects/g/nilearn/nilearn/?mode=list&severity=recommendation

Please review carefully. Not sure about these, especially unused imports. Unused imports might have undocumented  side effects. If that's the case, they should be documented when importing, as is already the case elsewhere in the source code.

### DIFF
--- a/nilearn/_utils/niimg_conversions.py
+++ b/nilearn/_utils/niimg_conversions.py
@@ -476,7 +476,7 @@ def concat_niimgs(niimgs, dtype=np.float32, ensure_ndim=None,
         lengths.append(niimg.shape[-1] if ndim == 4 else 1)
 
     target_shape = first_niimg.shape[:3]
-    if dtype == None:
+    if dtype is None:
         dtype = _get_data(first_niimg).dtype
     data = np.ndarray(target_shape + (sum(lengths), ),
                       order="F", dtype=dtype)
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -28,7 +28,7 @@ from . import cm, glass_brain
 from .edge_detect import _edge_map
 from .find_cuts import find_xyz_cut_coords, find_cut_slices
 from .. import _utils
-from ..image import new_img_like, load_img
+from ..image import new_img_like
 from ..image.resampling import (get_bounds, reorder_img, coord_transform,
                                 get_mask_bounds)
 from nilearn.image import get_data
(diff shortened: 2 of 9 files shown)


nilearn/plotting/surf_plotting.py
```diff
@@ -11,7 +11,6 @@
 from matplotlib.colorbar import make_axes
 from matplotlib.cm import ScalarMappable, get_cmap
 from matplotlib.colors import Normalize, LinearSegmentedColormap
-from mpl_toolkits.mplot3d import Axes3D
```
agramfort, line 14: I think this is required for old matplotlib. Rather put a `#noqa` at the end of the line

agramfort (commented):
besides LGTM

thx @DimitriPapadopoulos

nilearn/plotting/surf_plotting.py
DimitriPapadopoulos: Done.

nilearn/plotting/displays.py
```diff
@@ -1910,6 +1910,7 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
                                                self._cut_displayed)
         return cut_coords
 
+    @classmethod
```
jeromedockes, line 1913: this should be `staticmethod` otherwise the bound method will have the class as its first parameter. btw it is a bit confusing that some class methods in this module call their first parameter `self` rather than the more conventional `cls`, but I guess that is outside the scope of this PR

DimitriPapadopoulos: Ah, right, thank you.

DimitriPapadopoulos: See #2923 for self -> cls.

DimitriPapadopoulos (commented):
Seems to be working now. Strange.

nilearn/_utils/niimg_conversions.py
```diff
@@ -476,7 +476,7 @@ def concat_niimgs(niimgs, dtype=np.float32, ensure_ndim=None,
         lengths.append(niimg.shape[-1] if ndim == 4 else 1)
 
     target_shape = first_niimg.shape[:3]
-    if dtype == None:
+    if dtype is None:
```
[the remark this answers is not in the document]
jeromedockes: what is the error message?

jeromedockes (approved):
thanks!

agramfort (approved):

agramfort:
Thx @DimitriPapadopoulos

bthirion:
Thx !

PR #2923, opened 2021-08-17
[FIX] use "cls" instead of "self" for @classmethod

DimitriPapadopoulos:
This is a convention, not a requirement, but still...

### DIFF
--- a/nilearn/plotting/displays.py
+++ b/nilearn/plotting/displays.py
@@ -1217,7 +1217,7 @@ class OrthoSlicer(BaseSlicer):
     _axes_class = CutAxes
 
     @classmethod
-    def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
+    def find_cut_coords(cls, img=None, threshold=None, cut_coords=None):
         "Instantiate the slicer and find cut coordinates"
         if cut_coords is None:
             if img is None or img is False:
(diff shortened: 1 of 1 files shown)


nilearn/plotting/displays.py
```diff
@@ -1897,17 +1897,17 @@ def find_cut_coords(self, img=None, threshold=None, cut_coords=None):
         if (not isinstance(cut_coords, collections.abc.Sequence) and
                 isinstance(cut_coords, numbers.Number)):
             cut_coords = [cut_coords] * 3
-            cut_coords = self._find_cut_coords(img, cut_coords,
-                                               self._cut_displayed)
+            cut_coords = cls._find_cut_coords(img, cut_coords,
```
jeromedockes, line 1900: thanks! for this to work we still need to make `_find_cut_coords` a static method

DimitriPapadopoulos: Yes, see previous PR #2921.

DimitriPapadopoulos: We do need to merge #2921 first.

DimitriPapadopoulos: Actually, this works (not) the same way before/after the PR. The real fix is in #2921, this here is just a variable name change.

agramfort (approved):

agramfort:
you have CI failures @DimitriPapadopoulos

DimitriPapadopoulos:
Fixed now.

agramfort:
@jeromedockes feel free to MRG

jeromedockes:
great, thanks!

PR #2924, opened 2021-08-18
[FIX] LGTM error: Non-callable called

DimitriPapadopoulos:
Call to a non-callable of class Template.

Looks like an LGTM false positive to me.
Anyway, this alternative sytax should make the error go away.

### DIFF
--- a/nilearn/externals/tempita/__init__.py
+++ b/nilearn/externals/tempita/__init__.py
@@ -147,6 +147,7 @@ class Template(object):
         if default_inherit is not None:
             self.default_inherit = default_inherit
 
+    @classmethod
     def from_filename(cls, filename, namespace=None, encoding=None,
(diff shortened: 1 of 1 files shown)


jeromedockes:
indeed the existing code is fine.
in any case this is not nilearn code but a copy of the tempita library that is distributed with nilearn; so we do not modify this code or anything in `nilearn/externals` (other than by replacing it by a more recent version of the vendored library). I am not very familiar with LGTM but it can probably be configured to ignore everything in `nilearn/externals`?

DimitriPapadopoulos:
@agramfort actually told me to fix this here, there's no upstream repository anymore

DimitriPapadopoulos:
See:
https://github.com/agramfort/tempita/pull/1

jeromedockes:
ok thanks!

PR #2925, opened 2021-08-18
[FIX] LGTM warning: Comparison using is when operands support `__eq__`

DimitriPapadopoulos:
Values compared using 'is' when equivalence is not the same as identity. Use '==' instead.

The above suggestion is incorrect, though. See discussion in https://github.com/nilearn/nilearn/pull/2920#discussion_r690723512. Instead use code suggested in https://github.com/nilearn/nilearn/pull/2920#discussion_r691281123.

### DIFF
--- a/nilearn/image/resampling.py
+++ b/nilearn/image/resampling.py
@@ -445,7 +445,11 @@ def resample_img(img, target_affine=None, target_shape=None,
         if copy and not input_img_is_string:
             img = _utils.copy_img(img)
         return img
-    if target_affine is affine and target_shape is shape:
+    if (
+        np.shape(target_affine) == np.shape(affine)
+        and np.allclose(target_affine, affine)
+        and np.array_equal(target_shape, shape)
+    ):
         return img
     if target_affine is not None:
         target_affine = np.asarray(target_affine)


nilearn/image/resampling.py
```diff
@@ -441,11 +441,15 @@ def resample_img(img, target_affine=None, target_shape=None,
...
+    if (
+        np.shape(target_affine) == np.shape(affine)
+        and np.allclose(target_affine, affine)
+        and np.array_equal(target_shape, shape)
+    ):
         if copy and not input_img_is_string:
             img = _utils.copy_img(img)
         return img
-    if target_affine is affine and target_shape is shape:
+    if target_affine is affine and target_shape == shape:
```
jeromedockes, line 452: thanks! I think it is this line (452) that needs to change and the comparison to None above (l 444) could stay as it was

DimitriPapadopoulos: You're right, of course. Perhaps you should push a commit of your own, right now I'm not comfortable with such details of the code, you actually wrote the patch.

jeromedockes (approved):
thanks @DimitriPapadopoulos ! this LGTM

nilearn/image/resampling.py
jeromedockes: thanks for taking care of it! your last commit seems good!

tsalo (approved):
The change looks good to me too. Thank you.

bthirion:
Thx to all !

PR #2928, opened 2021-08-19
Add pathlib.Path support for nilearn.datasets

RaphaelMeudec:
This PR adds support for `pathlib.Path` to `data_dir` argument in datasets fetchers. All the fetchers from `nilearn.datasets` now accepts `pathlib.Path` or a string (instead of string only).

I have modified the tests from all fetchers which had to cast the `tmp_path` fixture to a string. Now the fixture is used directly as `data_dir` argument. This means there is no more tests for `data_dir` as a string, and I wasn't sure if I needed to keep both case (through a parametrize for each test for example)

The PR can easily be read commit by commit:

1. Changes to `nilearn.datasets.utils` that add `pathlib.Path` support
2. Modifications of the tests to avoid casting `tmp_path` to string everytime
3. Update document variable `data_dir`
4. Whatnew entry

I'm open to your opinions on the "no more `data_dir` as string tests". Maybe I could add both cases (`pathlib.Path` or string) in the tests of the following methods (`nilearn.datasets.utils.get_dataset_dir` and `nilearn.datasets.utils.fetch_files`) and leave the pathlib everywhere else?

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -54,6 +54,8 @@ Enhancements
   parameters or common lists of options for example. The standard parts are defined
   in a single location (`nilearn._utils.docs.py`) which makes them easier to
   maintain and update. (See `#2875 <https://github.com/nilearn/nilearn/pull/2875>`_)
+- The `data_dir` argument can now be either a `pathlib.Path` or a string. This
+  extension affects datasets and atlas fetchers.
 
 Changes
 -------
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -38,7 +38,7 @@ resume : :obj:`bool`, optional
 
 # Data_dir
 docdict['data_dir'] = """
-data_dir : :obj:`str`, optional
+data_dir : :obj:`pathlib.Path` or :obj:`str`, optional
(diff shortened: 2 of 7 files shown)


jeromedockes:
great, thanks! that's very useful

> Maybe I could add both cases (`pathlib.Path` or string) in the tests of the following methods (`nilearn.datasets.utils.get_dataset_dir` and `nilearn.datasets.utils.fetch_files`) and leave the pathlib everywhere else?

yes that sounds good! it's useful to check that these 2 continue working with strings, but probably not worth the trouble to parametrize every fetcher's test

RaphaelMeudec:
@jeromedockes I pushed the changes, let me know if there are stuff to change!

jeromedockes:
> @jeromedockes I pushed the changes, let me know if there are stuff to change!

everything looks great, thanks! On a similar topic if you have time you can have a look at #2803 :)

jeromedockes (approved):

NicolasGensollen (approved):
Looks great ! Thanks @RaphaelMeudec ! :+1:

bthirion (approved):

bthirion:
LGTM. Merging, thx !

PR #2929, opened 2021-08-19
ENH: Add stacklevel

larsoner:
Just adds `stacklevel` to help people more easily find where they need to update their code.


jeromedockes:
thanks a lot!

jeromedockes:
BTW this function keeps track manually of whether the warning has been shown
basically to work around the fact that this shows a warning twice:

```python
import warnings


def f():
    warnings.warn("message")


f()
    pass
f()
```

If you know a better way of dealing with this please share!

PR #2930, opened 2021-08-19
[FIX] LGTM error: Non-iterable used in for loop

### LINKED ISSUE #2926: LGTM error: Non-iterable used in for loop
I don't fully grasp the details of [this LGTM.com alert](https://lgtm.com/projects/g/nilearn/nilearn/snapshot/0cd925e33787e277990b524efeba0c8125d04816/files/nilearn/decoding/space_net.py?sort=name&dir=ASC&mode=heatmap#x4b41318e0293a848:1):
> `[ 3 Values ]` of class '`[ 3 Values ]`' may be used in for-loop.

What are these "3 Values" about?

Nevertheless, it is true that in `l1_ratios` [might be a float, hence not iterable](https://github.com/nilearn/nilearn/blob/bbbc4ba/nilearn/decoding/space_net.py#L685-L687):
```python
            if isinstance(l1_ratios, numbers.Number):
                l1_ratios = [l1_ratios]
            for l1_ratio in l1_ratios:s.Number):
```
Indeed, `l1_ratios` is initialised to be a float, so why check whether it is an instance of `numbers.Number`? Am I missing something here?

DimitriPapadopoulos:
The error message doesn't make much sense:
> `[3 Values]` of class '`[3 Values]`' may be used in for-loop.

Also, this looks like a false positive. The intent of this commit is to work around the false positive. It doesn't fix an actual problem.

Closes #2926


jeromedockes:
I think `alphas` can be `None` so it is a bit different than for `l1_ratios`.
also I think I would prefer using `isinstance(l1_ratios, collections.Sequence)` rather than raising and catching an exception in a situation that is not exceptional at all if that's ok

DimitriPapadopoulos:
Ah, but that's the usual way of detecting iterables in Python: "assume instead of testing". And without additional libraries.

I'll look into `alphas`.

jeromedockes:
`collections` is part of the standard library. and here we have no use for the call to `iter`, the only goal of this is to check if `l1_ratios` is an iterable so calling insinstance with the ABC reflects this goal more directly. also it results in less lines of code

jeromedockes:
btw sorry now it is in `collections.abc`, not `collections`, and actually `Iterable` is sufficient

DimitriPapadopoulos:
I had always read in early texts about Python that you should assume and not test, using exceptions if needed, even in cases that are not exceptional. However, this idiom seems to be obsolete nowadays, and clearly in this case applying that idiom would result in duplicate code of would require an additional function:
```python
    try:
        do_process(l1_ratios, ...)
    except TypeError:
        do_process([l1_ratios], ...)
```
The idiom may aplly to one-liners only, I cannot recall precisely. 

Anyway, the `if` is more readable. I do find the number of ines of code irrelevant as such, it's all about readability and maintanibility, but fewer lines of code without the expense of obscure language tricks (like was often the case with Perl) do align with readability.

In short, I cannot but agree with you. I'll fix that.

DimitriPapadopoulos:
The thing is I'm not certain this will work around the LGTM error, and I'm not sure how to test that (easily) without committing.

DimitriPapadopoulos:
Note that the `l1_ratio`/`l1_ratios` documentation is incorrect.

nilearn/decoding/space_net.py
```diff
@@ -777,11 +778,12 @@ def fit(self, X, y):
         n_samples, _ = X.shape
         y = np.array(y).copy()
         l1_ratios = self.l1_ratios
-        if isinstance(l1_ratios, numbers.Number):
+        if not isinstance(l1_ratios, collections.abc.Iterable):
             l1_ratios = [l1_ratios]
         alphas = self.alphas
-        if isinstance(alphas, numbers.Number):
-            alphas = [alphas]
+        if alphas is not None:
```
bthirion, line 784: Is this case currently tested ?

bthirion (commented):
LGTM otherwise.

nilearn/decoding/space_net.py
DimitriPapadopoulos: It is tested implicitly: 
```
>>> import numbers
>>> isinstance(None, numbers.Number)
False
>>> 
```

bthirion: Sorry, I meant: is it covered by unit tests ?

DimitriPapadopoulos:
@jeromedockes Here is a short note about using exceptions in Python even for non-exceptional stuff: [Is it a good practice to use try-except-else in Python?](https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python#16138864)

nilearn/decoding/space_net.py
DimitriPapadopoulos: Ah, I don't know. It would be nice to add a test if not. That said this is a straightforward change. Just to shut up LGTM.com.

bthirion (approved):

bthirion (approved):
Thx !

NicolasGensollen (approved):
LGTM, thanks!

Issue #2934, opened 2021-08-23
Sphinx-gallery duplicating report images


emdupre:
Recent updates to sphinx-gallery seem to have inserted duplicate images of the report output just above the report object itself, for example [here](https://nilearn.github.io/auto_examples/06_manipulating_images/plot_mask_computation.html#computing-a-mask-from-the-background).

Fixing this may be as simple as not calling the report on its own line, but to be confirmed !

### Affected pages

Any page that includes a NiftiMasker report object.

PR #2935, opened 2021-08-24
[FIX] Duplicate images in reports

NicolasGensollen:
See #2934 

It looks like it also does this when generating reports within Notebooks.

WIP...

### DIFF
--- a/nilearn/input_data/base_masker.py
+++ b/nilearn/input_data/base_masker.py
@@ -6,6 +6,7 @@ Transformer used to apply basic transformations on MRI data.
 
 import warnings
 import abc
+import functools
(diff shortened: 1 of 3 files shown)


bthirion (commented):
Apparently, one line is not tested yet.

emdupre (approved):
The [latest artifacts](https://7962-1235740-gh.circle-artifacts.com/0/dev/auto_examples/06_manipulating_images/plot_mask_computation.html#sphx-glr-auto-examples-06-manipulating-images-plot-mask-computation-py) LGTM ! Thanks, @NicolasGensollen !

PR #2944, opened 2021-08-31
[MAINT] Update jquery version

NicolasGensollen:
This PR proposes to update the version for `jquery.min.js` from 3.3.1 to 3.6.0.
See this [alert](https://github.com/nilearn/nilearn/security/dependabot/nilearn/plotting/data/js/jquery.min.js/jquery/open) where the suggested fix is to update to 3.5.0 or later.


KamalakerDadi:
Great!

I guess we need to do each time whenever we receive this warning?

NicolasGensollen:
>I guess we need to do each time whenever we receive this warning?

Once we update the jquery version, we shouldn't get this exact warning anymore since the vulnerability issues were patched in 3.5.0.

KamalakerDadi:
I guess we need to do each time whenever we receive this warning?

>Once we update the jquery version, we shouldn't get this exact warning anymore since the vulnerability issues were patched in 3.5.0.

Also, I would do simple visualization checks that are linked to this jsscript. Just to be sure. As CircleCI full cannot show this interactively.

NicolasGensollen:
>Also, I would do simple visualization checks that are linked to this jsscript. Just to be sure. As CircleCI full cannot show this interactively.

Yes, I didn't notice any difference. 

I'm gonna go ahead and merge this, thanks for the reviews!

Issue #2777, opened 2021-05-01
Integration of `load_confounds` to `nilearn`


htwangtw:
The `load_confounds` team would like to see `load_confounds` added to `nilearn`! [Link to the repository](https://github.com/SIMEXP/load_confounds)
`load_confounds` is a small tool to load fMRIprep confound timeseries based on methods benchmarked in [Ciric et. al.,  2017](https://doi.org/10.1016/j.neuroimage.2017.03.020). The output can be passed to`NiftiMasker` and related functions directly. 
It's a bit small to be a stand-alone project, and the integration will help more users! 

### Benefits to the change
`load_confounds` is a great addition to nilearn for the following reasons that I can think of:

1. **Bridging `fMRIprep` outputs and nilearn analysis functions**: 
We want to make the confound selection strategy more streamlined and reduce possible errors introduced at this stage. I am very certain there are many people who loaded the full fMRIprep confound output file as the confound regressors. `load_confounds` would be a helpful addition to that and the existing function that handles confound variables.  There's already precedence of the new GLM with fMRIprep/BIDS specific supports.  `load_confounds` can smoothen the process. 

2. **Almost completed project with good code quality**: 
The integration will not be a lot of work. `load_confounds` is almost a finished product with good code quality. There's one more strategy to be added on the `load_confounds` side. Some changes in `NifitMasker` and GLM first-level analysis can make the integration work smoother. The changes will allow volume censoring performed correctly.

    - NifitMasker: move `sample_mask` to the `transform` method rather than at the initialisation stage. This will allow scrubbing outliers per image, along with non-steady-state volumes.
   - GLM: add `sample_mask` to first level GLM for removing non-steady-state volumes and scrubbing motion outliers. 

### Pseudocode for the new behavior, if applicable
This is what I expect to see when `load_confounds` is merged into `nilearn`, using `NiftiMasker` as an example:
```python
from nilearn.load_confounds import Params36
from nilearn.input_data import NiftiMasker 

# load_confounds auto-detects the companion .tsv file (which needs to be in the same directory)
file = "path/to/file/sub-01_ses-001_bold.nii.gz"
confounds, sample_mask = Params36().load(file)

# Use the confounds and sample_mask in a nilearn masker 
# sample mask is defined at initialisation stage in the current implementation
# We would move it to the transform methods
masker = NiftiMasker(smoothing_fwhm=5, standardize=True)
img = masker.fit_transform(file, confounds=confounds, sample_mask=sample_mask)
```

[...]

PR #2946, opened 2021-09-07
[NF] migrate load_confounds base class

htwangtw:
Start migrating `load_confounds` to nilearn (#2777) on behalf of all coauthors. 

* ADD base class `Confounds` and all helper functions
* ADD test data; taken from OpenNeuro ds000003
* ADD all tests associated with class `Confounds`

Few things still needs some discussions and/or changes:

- [x] add more detailed doc on the relationship with fMRIprep
- [x] the version of fMRIprep it supports
- [x] demo
- [ ] Add unit test for some low level functions (compcor and scrubbing especially need this)
- [ ] Remove test data that can be generated on the fly (such as empty files).
- [ ] Raise warning related to fMRIprep version.
- [ ] Refactor `Confounds` into a function?
- [ ] the name of module - how to make clear about the fMRIprep part, do we need to?


htwangtw:
I have done some research on the minimal version of fMRIprep output we can support and added them to the existing docstrings. We can fully support everything from 1.4.X, and 1.2.X except `compcor`.

There are two aspect that I would like @NicolasGensollen to make an executive call on:
1. I can make sure some sanity checks on versions added to this PR, or as a separate one. I prefer to have it as a separate one but would like to see what you think  makes more sense.
2. Name of the module: `load_confounds` only works with fMRIprep outputs. The current name is potentially confusing. It can make user think this is for all kinds of confounds, rather than just for fMRIprep.
My idea is to put `load_confounds` under a directory called `fmriprep`, and import the `Confounds` like this:
```python3
from nilearn.fmriprep import Confounds
```
This can be a place for all potential future fmriprep helper functions. WDYT? cc @pbellec

tsalo:
@htwangtw can the module work generally with BIDS derivatives (especially w.r.t. https://github.com/bids-standard/bids-specification/pull/519, which proposes rules that are probably what dictate fMRIPrep's derivatives)? I think a `nilearn.bids` would be preferable to a `nilearn.fmriprep`.

htwangtw:
@tsalo Thanks for bringing this up! 

I had a quick read through of BEP 012. From what I am aware, `load_confounds` fits well, as there is plan to adapt to new changes in `fMRIprep`. We have some of the very minimal file validity checks for now and it would be great to have it streamlined with BIDS in the future. `nilearn.bids` is a good idea.

pbellec:
I feel like `nilearn.fmriprep` would be the most straightforward at the moment, and something that may need to be revisited if a 
more general set of `bids` helper gets developed for nilearn. 

Currently, the main use case for the confounds features is for fMRIprep users. I am not aware of software using the BIDS fMRI derivatives at the moment, and I am not sure that fMRIprep itself conforms to the standard, which is still a draft as far as I can tell. Also, BIDS is most commonly associated with raw data, and unless `nilearn` expands to offer tools to handle BIDS data (which would then overlap with `pybids`), I am not sure what other modules will live in `nilearn.bids`.

tsalo:
@pbellec one motivation I have for `bids` over `fmriprep` is that I'm working on functions to write out BIDS derivatives for GLMs (also based on a draft instead of the real specification) in #2715. We've also discussed the need for lightweight BIDS querying tools so that nilearn doesn't need pybids as a dependency. Some of those tools may end up in nilearn (or [`pybids-light`](https://github.com/bids-standard/pybids-light), which could become a dependency).

pbellec:
so if there is a push for a set of bids tools in nilearn, `nilearn.bids` sounds good. Then we may need to rename the strategy module something like `fmriprep_denoise.py` and expose the confound strategies directly through `nilearn.bids`. So users would type
``` 
from nilearn.bids import Scrubbing
``` 
Another option would be to have long and explicit class names, like what is done for the masker classes.
``` 
from nilearn.bids import FmriPrepScrubbing
```

NicolasGensollen:
Thanks for opening this @htwangtw ! :+1: 
I'll make a review tomorrow but it looks really good at first glance. 
I think it could be helpful to have some examples that we could use to test things a little bit (they will be useful for the docs at some point anyway). 
Concerning the name of the module, I think it should be explicit enough such that users aren't surprised by what is inside (we get this remark every now and then about `input_data` which could have been `maskers` for example).  
I'm also worried that if we have a `bids` module, then it would make sense to put all bids-related functionalities inside which could quickly get messy (not saying it would necessarily, just trying to think how it could go wrong...).

htwangtw:
I can certainly add an examples. We do have an existing demo jupyter notebook based on [a NiftiMasker example](https://nilearn.github.io/auto_examples/03_connectivity/plot_signal_extraction.html#sphx-glr-auto-examples-03-connectivity-plot-signal-extraction-py). It needs some updates but I can certainly put it somewhere.

nilearn/load_confounds/confounds.py
```diff
@@ -0,0 +1,308 @@
...
+        if key_found:
+            list_confounds.extend(key_found)
+        elif key != "non_steady_state":
+            missing_keys.append(key)
+    if missing_keys:
+        raise MissingConfound(keywords=missing_keys)
+    return list_confounds
+
+
+def _flag_single_gifti(img_files):
```
bthirion, line 56: This function is loosely related to confounds, could go to the  _utils module instead.

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,387 @@
...
+            + f" {missing_confounds}"
+            + f" {missing_keys}"
+            + ". You may want to try a different denoising strategy."
+        )
+        raise ValueError(error_msg)
+
+
+class Confounds:
+    """
+    Confounds from fmriprep
```
bthirion, line 50:
```suggestion
    Use confounds from fmriprep
```

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,387 @@
...
+
+    n_motion : float
+        Number of pca components to keep from head motion estimates.
+        If the parameters is strictly comprised between 0 and 1, a principal
+        component analysis is applied to the motion parameters, and the number
+        of extracted components is set to exceed `n_motion` percent of the
+        parameters variance.
+        If the n_components = 0, then no PCA is performed.
+
+    fd_thresh : float, optional
```
bthirion, line 87:
```suggestion
    fd_threshold : float, optional
```

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,387 @@
...
+        Framewise displacement threshold for scrub (default = 0.2 mm)
+
+    std_dvars_thresh : float, optional
+        Standardized DVARS threshold for scrub (default = 3)
```
bthirion, line 91: Sorry but what is DVARS ? (I don't like these accronyms...)

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,387 @@
...
+        If the parameters is strictly comprised between 0 and 1, a principal
+        component analysis is applied to the motion parameters, and the number
+        of extracted components is set to exceed `n_motion` percent of the
+        parameters variance.
+        If the n_components = 0, then no PCA is performed.
+
+    fd_thresh : float, optional
+        Framewise displacement threshold for scrub (default = 0.2 mm)
+
+    std_dvars_thresh : float, optional
```
bthirion, line 90:
```suggestion
    std_dvars_threshold : float, optional
```

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,387 @@
...
+
+    motion : string, optional
+        Type of confounds extracted from head motion estimates.
+        "basic" translation/rotation (6 parameters)
+        "power2" translation/rotation + quadratic terms (12 parameters)
+        "derivatives" translation/rotation + derivatives (12 parameters)
+        "full" translation/rotation + derivatives + quadratic terms + power2d
+               derivatives (24 parameters)
+
+    n_motion : float
```
bthirion, line 79:
```suggestion
    n_motion_components : float
``` (prefer more explcii names)

File: nilearn/load_confounds/parser.py [hunk 7]
```diff
@@ -0,0 +1,387 @@
...
+        "full" noise components calculated using both temporal and anatomical
+
+    n_compcor : int or "auto", optional
+        The number of noise components to be extracted.
+        For acompcor_combined=False, and/or compcor="full", this is the number
+        of components per mask.
+        Default is "auto": select all components
+        (50% variance explained by fMRIPrep defaults)
+
+    acompcor_combined: boolean, optional
```
bthirion, line 132: The name is not great. Couldn't it be an option in `compcor` above ?

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,387 @@
...
+        error_msg = (
+            "The following keys or parameters are missing: "
+            + f" {missing_confounds}"
+            + f" {missing_keys}"
+            + ". You may want to try a different denoising strategy."
+        )
+        raise ValueError(error_msg)
+
+
+class Confounds:
```
bthirion, line 48: I'm wondering why it is a class not a function ?
Does it have some internal variables / state that justify having a class ?
To me it follows an input/output logic, i.e. that of a function, but I may have missed something.

nilearn/load_confounds/data/missing_desc-confounds_regressors.tsv
```diff
@@ -0,0 +1,2 @@
..._cor_67	a_comp_cor_68	a_comp_cor_69	a_comp_cor_70	a_comp_cor_71	a_comp_cor_72	a_comp_cor_73	a_comp_cor_74	a_comp_cor_75	a_comp_cor_76	a_comp_cor_77	a_comp_cor_78	a_comp_cor_79	a_comp_cor_80	a_comp_cor_81	a_comp_cor_82	a_comp_cor_83	a_comp_cor_84	a_comp_cor_85	a_comp_cor_86	a_comp_cor_87	a_comp_cor_88	a_comp_cor_89	a_comp_cor_90	a_comp_cor_91	a_comp_cor_92	a_comp_cor_93	a_comp_cor_94	a_comp_cor_95	a_comp_cor_96	a_comp_cor_97	a_comp_cor_98	a_comp_cor_99	a_comp_cor_100	a_comp_cor_101	a_comp_cor_102	a_comp_cor_103	a_comp_cor_104	a_comp_cor_105	a_comp_cor_106	a_comp_cor_107	a_comp_cor_108	a_comp_cor_109	a_comp_cor_110	a_comp_cor_111	a_comp_cor_112	a_comp_cor_113	a_comp_cor_114	a_comp_cor_115	a_comp_cor_116	a_comp_cor_117	a_comp_cor_118	a_comp_cor_119	a_comp_cor_120	a_comp_cor_121	a_comp_cor_122	a_comp_cor_123	a_comp_cor_124	a_comp_cor_125	trans_x	trans_x_power2	trans_x_derivative1_power2	trans_y_derivative1	trans_y_derivative1_power2	trans_y_power2	trans_z	trans_z_derivative1	trans_z_power2	trans_z_derivative1_power2	rot_x	rot_x_derivative1	rot_x_derivative1_power2	rot_x_power2	rot_y	rot_y_derivative1	rot_y_derivative1_power2	rot_y_power2	rot_z	rot_z_derivative1	rot_z_derivative1_power2
```
bthirion, line 1: What are the empty files for ?

bthirion (commented):
Thx for this PR !
Now I better understand why the code refers to fMRIprep rather than BIDS. 
What i don't understand atm is why we need a Confound class, not a function ? (functions are earsier to maintain).

nilearn/load_confounds/parser.py
htwangtw: I will keep my personal opinion about this acronym to myself :stuck_out_tongue: . Here's the explanation from [MRIQC docs](https://mriqc.readthedocs.io/en/stable/iqms/bold.html#measures-for-the-temporal-information):

> DVARS - D referring to temporal derivative of timecourses, VARS referring to RMS variance over voxels ([Power2012] dvars_nstd) indexes the rate of change of BOLD signal across the entire brain at each frame of data.

htwangtw: Certainly we can expand this.

nilearn/load_confounds/parser.py
htwangtw: In that case we will have five options:

1. `temp`
2. `anat_separated` (WM compcor and CSF compcor concatenated)
3. `anat_combined` (use compcor calculated on a WM+CSF mask)
4. `temp_anat_separated` (`compcor='full', acompcor_combined=False`)
5. `temp_anat_combined`(`compcor='full', acompcor_combined=True`)

Better idea for the option names?

nilearn/load_confounds/data/missing_desc-confounds_regressors.tsv
htwangtw: For this specific one, it is a place holder for testing the compcor noise component required files. If the associated json is missing, even when the confound timeseries tsv exist, it will raise an error. There's no point to put a real file here so it's only got one line.

The empty imaging files are place holder to test the file grabbing function. When a functional imaging derivative is given, the `load` method look for the associated confound timeseries.

nilearn/load_confounds/parser.py
htwangtw: Here is a piece of important code that's not in this PR:
we have some predefined strategy that inherent the Confounds class: https://github.com/SIMEXP/load_confounds/blob/master/load_confounds/strategies.py
The main motivation is to cut down the parameters exposed and some internal sanity check between different noise components. If we write this as a function, each helper function will have to take more parameters than now. 
Happy to discuss!

nilearn/load_confounds/tests/test_parser.py
```diff
@@ -0,0 +1,476 @@
...
+        lc.Confounds(strategy=["error"])
+
+    with pytest.raises(ValueError):
+        lc.Confounds(strategy=[0])
+
+    conf = lc.Confounds(strategy=["motion"])
+    assert "non_steady_state" in conf.strategy
+
+
+def test_motion():
```
NicolasGensollen, line 245: I'm wondering whether we should unit test low level helper functions (at least some of them...). WDYT?

NicolasGensollen (commented):
Thanks @htwangtw ! :+1: 
I made a first pass (mostly trying to understand how things are structured) and made a few small comments (see below).

examples/03_connectivity/plot_load_confounds.py
```diff
@@ -0,0 +1,97 @@
+"""
+load_confounds
```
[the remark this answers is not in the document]
htwangtw: The demo is a copy of the one you are pointing to. Happy to merge the two!

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,387 @@
...
+    Parameters
+    ----------
+    strategy : list of strings
+        The type of noise components to include.
+        "motion" head motion estimates.
+        "high_pass" discrete cosines covering low frequencies.
+        "wm_csf" confounds derived from white matter and cerebrospinal fluid.
+        "global" confounds derived from the global signal.
+        "ica_aroma" confounds derived from ICA-AROMA.
+        "scrub" regressors for Power 2014 scrubbing approach.
```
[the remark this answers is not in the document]
htwangtw: Nice catch!

nilearn/load_confounds/tests/test_parser.py
htwangtw: Some of the `sample_mask` related functions would be really helpful with some extra tests I feel.

htwangtw:
w.r.t. fmriprep vs BIDS -
I had a think about this today. For now we are happy with what `load_confounds` is doing - it's a one stop shop for users to use fMRIprep outputs in nilearn. With the predefined strategies, it would be even simpler. Thus I have a preference for my original proposal (fmriprep).

Moving load_confounds to part of BIDS integration can be some work down the line. It will be a while before BEP 012 get merged and applied to fMRIPrep. That's a lot of things to anticipate and address for now, and I cannot see the changes of BEP 012 coming in the immediate future. At the moment load_confounds is for fMRIprep rather than BIDS.

nilearn/load_confounds/parser.py
bthirion: yes, replacing `temp` with  `temporal`

nilearn/load_confounds/parser.py
bthirion: I don't fully undestand. You could perfectly have strategies in a function ?

nilearn/load_confounds/data/missing_desc-confounds_regressors.tsv
bthirion: If it is for testing, you maye generate them on-the-fly when runing the tests ?

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+    """
+    Use confounds from fmriprep
+
+    To enable easy confound variables loading from fMRIprep outputs, Confounds
+    provides a interface that groups subsets of confound variables into noise
+    components and their parameters. It is possible to fine-tune a subset of
+    noise components and their parameters through Confounds.
+
+    The implementation will only support fMRIPrep output from the 1.2.x series.
+    The `compcor` noise component requires 1.4.x series or above.
```
NicolasGensollen, line 60: I would emphasize these two lines by putting them under a sphinx warning:

```
.. warning::
    The implementation.....
```

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+
+    Parameters
+    ----------
+    strategy : list of strings
+        The type of noise components to include.
+        "motion" head motion estimates.
```
NicolasGensollen, line 66: Could you turn this into a bullet list? I think it will render better in the docs.

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        "high_pass" discrete cosines covering low frequencies.
+        "wm_csf" confounds derived from white matter and cerebrospinal fluid.
+        "global" confounds derived from the global signal.
+        "compcor" confounds derived from CompCor.
+        "ica_aroma" confounds derived from ICA-AROMA.
+        "scrub" regressors for Power 2014 scrubbing approach.
+
+    motion : string, optional
+        Type of confounds extracted from head motion estimates.
+        "basic" translation/rotation (6 parameters)
```
NicolasGensollen, line 76: Same, could you turn this into a bullet list?

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        "power2" translation/rotation + quadratic terms (12 parameters)
+        "derivatives" translation/rotation + derivatives (12 parameters)
+        "full" translation/rotation + derivatives + quadratic terms + power2d
+               derivatives (24 parameters)
+
+    n_motion_components_components : float
```
NicolasGensollen, line 82:
```suggestion
    n_motion_components : float
```

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        Number of pca components to keep from head motion estimates.
+        If the parameters is strictly comprised between 0 and 1, a principal
+        component analysis is applied to the motion parameters, and the number
+        of extracted components is set to exceed `n_motion_components` percent
+        of the parameters variance.
+        If the n_motion_components_components = 0, then no PCA is performed.
```
NicolasGensollen, line 88:
```suggestion
        If the n_motion_components = 0, then no PCA is performed.
```

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        Standardized DVARS threshold for scrub (default = 3).
+        DVARs is defined as root mean squared intensity difference of volume N
+        to volume N+1 (Power et al. 2012). D referring to temporal derivative
+        of timecourses, VARS referring to root mean squared variance over
+        voxels.
+
+    wm_csf : string, optional
+        Type of confounds extracted from masks of white matter and
+        cerebrospinal fluids.
+        "basic" the averages in each mask (2 parameters)
```
NicolasGensollen, line 103: Same, could you turn this into a bullet list?

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        "power2" averages and quadratic terms (4 parameters)
+        "derivatives" averages and derivatives (4 parameters)
+        "full" averages + derivatives + quadratic terms
+               + power2d derivatives (8 parameters)
+
+    global_signal : string, optional
+        Type of confounds extracted from the global signal.
+        "basic" just the global signal (1 parameter)
```
NicolasGensollen, line 111: Same

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        Framewise displacement threshold for scrub (default = 0.2 mm)
+
+    std_dvars_threshold : float, optional
+        Standardized DVARS threshold for scrub (default = 3).
+        DVARs is defined as root mean squared intensity difference of volume N
+        to volume N+1 (Power et al. 2012). D referring to temporal derivative
+        of timecourses, VARS referring to root mean squared variance over
+        voxels.
+
+    wm_csf : string, optional
```
NicolasGensollen, line 100: For these parameters that can only take a finite number of values, I would use brackets with default value first in the list:

```
wm_csf : {"basic", "power2", "derivatives", ....}, optional
```

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+    strategies for the control of motion artifact in studies of functional
+    connectivity" Neuroimage 154: 174-87
+    https://doi.org/10.1016/j.neuroimage.2017.03.020
+    """
+
+    def __init__(
+        self,
+        strategy=["motion", "high_pass", "wm_csf"],
+        motion="full",
+        n_motion_components=0,
```
NicolasGensollen, line 196: Could you write the missing default values in the docstring?

NicolasGensollen (commented):
Thanks for addressing the reviews so quickly @htwangtw !
Below are a few comments and suggestions (mostly on documentation and formatting).

doc/modules/reference.rst
```diff
@@ -226,6 +226,25 @@ the :ref:`user guide <user_guide>` for more information and usage examples.
    NiftiMapsMasker
    NiftiSpheresMasker
 
+ .. _load_confounds_ref:
+
+ :mod:`nilearn.load_confonds`: Load fMRIPrep generated confounds
```
NicolasGensollen, line 231:
```suggestion
 :mod:`nilearn.load_confounds`: Load fMRIPrep generated confounds
```

doc/modules/reference.rst
```diff
@@ -226,6 +226,25 @@ the :ref:`user guide <user_guide>` for more information and usage examples.
...
+ ===============================================================
```
NicolasGensollen, line 232:
```suggestion
 ================================================================
```

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -8,8 +8,9 @@
 We also show the importance of defining good confounds signals: the
 first correlation matrix is computed after regressing out simple
 confounds signals: movement regressors, white matter and CSF signals, ...
-The second one is without any confounds: all regions are connected to
-each other.
+The second one demonstrated the functionality of `load_confounds` module to
```
NicolasGensollen, line 11:
```suggestion
The second one demonstrated the functionality of the :mod:`nilearn.load_confounds` module to
```

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -73,6 +74,81 @@
 plotting.plot_matrix(correlation_matrix, figure=(10, 8), labels=labels[1:],
                      vmax=0.8, vmin=-0.8, reorder=True)
 
+##############################################################################
+# Load confounds from file using a flexible strategy with load_confounds
+# -----------------------------------------------------
+# load_confounds can be used to create a Confounds class with flexible
```
NicolasGensollen, line 80:
```suggestion
# The :mod:`nilearn.load_confounds` module can be used to create
# a :class:`nilearn.load_confounds.Confounds` class with flexible
```

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -73,6 +74,81 @@
...
+# parameters. We create a Confounds class which specifies strategies and
```
NicolasGensollen, line 81:
```suggestion
# parameters. We create a :class:`nilearn.load_confounds.Confounds` class which specifies strategies and
```

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -73,6 +74,81 @@
...
+# optional parameters. Then the load() method selects the relevant columns from
```
NicolasGensollen, line 82:
```suggestion
# optional parameters. Then the :func:`nilearn.load_confounds.Confounds.load` method selects the relevant columns from
```

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        "ica_aroma" confounds derived from ICA-AROMA.
+        "scrub" regressors for Power 2014 scrubbing approach.
+
+    motion : string, optional
+        Type of confounds extracted from head motion estimates.
+        "basic" translation/rotation (6 parameters)
+        "power2" translation/rotation + quadratic terms (12 parameters)
+        "derivatives" translation/rotation + derivatives (12 parameters)
+        "full" translation/rotation + derivatives + quadratic terms + power2d
+               derivatives (24 parameters)
```
NicolasGensollen, line 80: This line is over-indented and will make Sphinx fail in strict mode.

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        voxels.
+
+    wm_csf : string, optional
+        Type of confounds extracted from masks of white matter and
+        cerebrospinal fluids.
+        "basic" the averages in each mask (2 parameters)
+        "power2" averages and quadratic terms (4 parameters)
+        "derivatives" averages and derivatives (4 parameters)
+        "full" averages + derivatives + quadratic terms
+               + power2d derivatives (8 parameters)
```
NicolasGensollen, line 107: Same, line is over-indented

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+
+    global_signal : string, optional
+        Type of confounds extracted from the global signal.
+        "basic" just the global signal (1 parameter)
+        "power2" global signal and quadratic term (2 parameters)
+        "derivatives" global signal and derivative (2 parameters)
+        "full" global signal + derivatives + quadratic terms
+               + power2d derivatives (4 parameters)
```
NicolasGensollen, line 115: Same

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+    scrub : string, optional
+        Type of scrub of frames with excessive motion (Power et al. 2014)
+        "basic" remove time frames based on excessive FD and DVARS
+        "full" also remove time windows which are too short after scrubbing.
+        one-hot encoding vectors are added as regressors for each scrubbed
+        frame.
+
+    compcor : string, optional
+        .. warning::
+                Require fmriprep >= v:1.4.0.
```
NicolasGensollen, line 126: This needs a blank line after to make sphinx happy

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        Type of confounds extracted from a component based noise correction
```
NicolasGensollen, line 127: Lines 127 to 131 are not indented properly (there is an extra space).

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,397 @@
...
+        The confounds loaded using the specified model. The columns of the
+        dataframe contains the labels.
+
+    `sample_mask_` : None, numpy.ndarray, or list of
+        When no volumns require removal, the value is None.
+        Otherwise, shape: (number of scans - number of volumes removed, )
+        The index of the niimgs along time/fourth dimension for valid volumes
+        for subsequent analysis.
+        This attribute should be passed to parameter `sample_mask` of
+        nilearn.NiftiMasker or nilearn.signal.clean.
```
NicolasGensollen, line 169:
```suggestion
        :class:`nilearn.input_data.NiftiMasker` or :func:`nilearn.signal.clean`.
```

NicolasGensollen (commented):
@htwangtw I spent some time building the docs from your branch in strict mode which always fail because of annoying little things like wrong indents, missing blank lines and so on...
Below are a few things I had to fix to make it work as well as some suggestions for better linking of resources. 
Note: Do not hesitate to request full builds from time to time by adding "[circle full]" in your commit message.

nilearn/load_confounds/confounds.py
```diff
@@ -0,0 +1,247 @@
...
+                    "The CompCor component is only supported for fMRIprep "
+                    "version >= 1.4.0."
+                )
+            )
+    return confounds_json
+
+
+def _get_file_raw(confounds_raw_path):
+    """Load raw confounds as a pandas DataFrame and check if the version
+    of fMRIprep (>=1.2.0) is supported based on header format (to-do)."""
```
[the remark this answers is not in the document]
htwangtw: Certainly doable!

doc/modules/reference.rst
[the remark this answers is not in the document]
htwangtw: Yep I copy-pasted your first suggestion and didn't realise github added a extra space at the beginning of each line.

htwangtw:
Regarding the suggestion on adding tests to private functions, I am only going to do a selective set.
We have a full coverage in the [original `load_confounds` repo](https://app.codecov.io/gh/SIMEXP/load_confounds) through test the high level API that's exposed to the users.
I can think of a few things that can be benefited from dedicated unit test 
1. some edge cases for scrubbing ineracting with non-steady-state outliers (this is done)
2. compcor option parsing. This is anticipating the changes in fMRIprep on the [new anatomical compcor label in 21.0.0rc0](https://github.com/nipreps/fmriprep/pull/2523)
Other than that I think the existing tests are doing good.

NicolasGensollen:
@htwangtw I just pushed a commit to your PR in which I propose to refactor a little bit some tests.
Let me know what you think, and feel free to delete it if you are not satisfied with the proposed changes.

htwangtw:
> @htwangtw I just pushed a commit to your PR in which I propose to refactor a little bit some tests.
> Let me know what you think, and feel free to delete it if you are not satisfied with the proposed changes.

Thanks for the refactoring @NicolasGensollen 
Parametrising tests made the test more concise. I like it.

nilearn/load_confounds/confounds.py
```diff
@@ -0,0 +1,265 @@
...
+        "full": {"derivative1", "power2", "derivative1_power2"},
+    }
+    for par in params:
+        for suff in suffix[model]:
+            params_full.append(f"{par}_{suff}")
+    return params_full
+
+
+def _pca_motion(confounds_motion, n_components):
+    """Reduce the motion paramaters using PCA."""
```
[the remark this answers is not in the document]
pbellec: I agree we should remove this option - it is not standard.

nilearn/_utils/load_confounds.py
```diff
@@ -0,0 +1,24 @@
...
+    flag_single_gifti = []  # gifti in pairs
+    for img in img_files:
+        ext = ".".join(img.split(".")[-2:])
+        flag_single_gifti.append((ext == "func.gii"))
+    return all(flag_single_gifti)
+
+
+def _is_camel_case(s):
+    "Check if the given string is in camel case."
+    return s != s.lower() and s != s.upper() and "_" not in s
```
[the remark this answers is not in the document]
htwangtw: I wondered about that too, but Nope!

nilearn/_utils/load_confounds.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def to_camel_case(snake_str):
```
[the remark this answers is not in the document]
htwangtw: See above

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -73,9 +74,86 @@
...
+# The :mod:`nilearn.load_confounds` module can be used to create
+# a :class:`nilearn.load_confounds.Confounds` class with flexible
+# parameters. We create a :class:`nilearn.load_confounds.Confounds` class
+# which specifies strategies and optional parameters. Then the
+# :func:`nilearn.load_confounds.Confounds.load` method selects the relevant
+# columns from the TSV file. Let's try a strategy similar to Params9 from
+# Ciric et al. 2017, and some of its variations:
+
+from nilearn.load_confounds import Confounds
+params9 = Confounds(strategy=["high_pass", "motion", "wm_csf", "global"],
```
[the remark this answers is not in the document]
htwangtw: It was simply the one a previous demo used. Still I think it's a good one for showcasing: it's a fairly standard motion based strategy.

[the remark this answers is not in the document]
bthirion: I see, but the naming is bad: it is unclear what "9" refers to.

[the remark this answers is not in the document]
pbellec: This name is derived from Ciric et al. 2017 - they call it "9P": 6 motion parameters + average CSF, WM and global signal (9 confounds total + the high pass filter). Maybe it would be better to stick with the terminology we'll use for strategies in Nilearn. This would be `Minimal` and a global signal, and we can mention in passing this fits with the `9P` from Ciric. WDYT @htwangtw ?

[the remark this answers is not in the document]
htwangtw: That sounds like a sensible idea.

NicolasGensollen:
@htwangtw The documentation build error wasn't related to this PR. I pushed a fix earlier today, so if you merge master you should have the CI all green again. :heavy_check_mark: 

Concerning the design choice of having a class vs. functions for Confounds, I also feel like we don't really need the complexity of multiple classes inheriting from the `Confounds` base class as they seem to "only" redefine some of their attributes. I might be wrong, but it seems to me that they mostly provide handy constructors but do not implement additional specific logic. 

Maybe the best way to decide is to compare the two designs (I can have a go at refactoring this if you want). 
WDYT?

htwangtw:
> @htwangtw The documentation build error wasn't related to this PR. I pushed a fix earlier today, so if you merge master you should have the CI all green again. heavy_check_mark
> 
> Concerning the design choice of having a class vs. functions for Confounds, I also feel like we don't really need the complexity of multiple classes inheriting from the `Confounds` base class as they seem to "only" redefine some of their attributes. I might be wrong, but it seems to me that they mostly provide handy constructors but do not implement additional specific logic.
> 
> Maybe the best way to decide is to compare the two designs (I can have a go at refactoring this if you want).
> WDYT?

Looks like there's problem with full build still, but partial build is fine. 

Great timing! I was leaving this comment to the last. I have time and just started a local branch for the refectoring and will see how it goes.

doc/modules/reference.rst
```diff
@@ -235,15 +235,15 @@ the :ref:`user guide <user_guide>` for more information and usage examples.
...
    :no-inherited-members:
 
-**Classes**:
+**Functions**:
 
 .. currentmodule:: nilearn.load_confounds
 
 .. autosummary::
    :toctree: generated/
    :template: class.rst
```
tsalo, line 244: I think you'll need to use the `function` template here too.

bthirion:
> @htwangtw The documentation build error wasn't related to this PR. I pushed a fix earlier today, so if you merge master you should have the CI all green again. heavy_check_mark
> 
> Concerning the design choice of having a class vs. functions for Confounds, I also feel like we don't really need the complexity of multiple classes inheriting from the `Confounds` base class as they seem to "only" redefine some of their attributes. I might be wrong, but it seems to me that they mostly provide handy constructors but do not implement additional specific logic.
> 
> Maybe the best way to decide is to compare the two designs (I can have a go at refactoring this if you want).
> WDYT?

Could not agree more. Thx !

htwangtw:
@NicolasGensollen @bthirion I have refactored the code to function - was less work then I thought and not that much chance in the number of lines. 
`Confounds` is now replaced by`load_confounds`. 
Currently, only the function exposed to the user is `load_confounds`. I tried to get some of the function name and docs terminology more consistent. However I still find the module names a bit chaotic. For now I have:
```
from nilearn.load_confounds import load_confounds
```
Is this okay?
Please let me know if things are making sense!

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -73,9 +74,85 @@
 plotting.plot_matrix(correlation_matrix, figure=(10, 8), labels=labels[1:],
                      vmax=0.8, vmin=-0.8, reorder=True)
 
+##############################################################################
+# Load confounds from file using a flexible strategy with load_confounds
+# ----------------------------------------------------------------------
+# The :mod:`nilearn.load_confounds` module can be used to create
+# a :class:`nilearn.load_confounds.Confounds` class with flexible
```
NicolasGensollen, line 81: The doc build is currently failing because these refs are broken. We should replace them with ":func:`nilearn.load_confounds.load_confounds`", and update the text.

nilearn/load_confounds/parser.py
```diff
@@ -0,0 +1,290 @@
...
+        error_msg = (
+            "The following keywords or parameters are missing: "
+            + f" {missing['confounds']}"
+            + f" {missing['keywords']}"
+            + ". You may want to try a different denoising strategy."
+        )
+        raise ValueError(error_msg)
+
+
+def load_confounds(img_files, **kargs):
```
NicolasGensollen, line 67: I think we should explicitly write the function inputs here to be consistent with the rest of Nilearn's API.

NicolasGensollen (commented):
Thanks for refactoring this so quickly @htwangtw ! :+1: 
I just made a quick pass (see below), I will have another look later today.

NicolasGensollen:
>Is this okay?
Please let me know if things are making sense!

The import will depend on the name we end up choosing but `from nilearn.load_confounds import load_confounds` makes sense to me.

Having only `load_confounds` public for now also makes sense. 
I believe other classes which were inheriting from `Confounds` will then become public functions calling `load_confounds` with some combinations of parameters, right?

htwangtw:
> I believe other classes which were inheriting from `Confounds` will then become public functions calling `load_confounds` with some combinations of parameters, right?

Yes. So it will be like:
```python
from nilearn.load_confounds import minimal
```
Function:
```python
from nilearn.load_confounds import load_confounds

def minimal(img_files, motion="full", wm_csf="basic", demean=True):
    strategy = ["high_pass", "motion", "wm_csf", "non_steady_state"]
    # some sanity checks here 
    return load_confounds(strategy=strategy, motion=motion, wm_csf=wm_csf, demean=demean)
```

GaelVaroquaux:
In terms of import path, I wonder whether it would not make sense put
this together with the maskers. I don't really like the name of the
modules where the maskers are based (I'm to blame).

Maybe an "io" module that groups the confounder loading and the maskers?

My 2 cents from the peanut gallery

bthirion:
You're right, but we need to enter a cycle of deprecation warnings before doing that.

GaelVaroquaux:
> You're right, but we need to enter a cycle of deprecation warnings before doing that.

Yes, and we could make it real slow: there is little harm in having the
import path "input_data" work for the maskers for a long time.

pbellec:
`input_data` would be a reasonable place to put `load_confounds`, as it's used to input confounds into nilearn. That was my first intuition, before realizing `input_data` is really used just for maskers currently. I guess maskers can also be used to write data out, hence the proposed change for `io`? I don't have an opinion on that change, but if you are comfortable putting `load_confounds` together with maskers, then `input_data` sounds like a good place, and `io` in the future would be good too.

htwangtw:
I have now moved `load_counfounds` to `input_data`. Maybe the `input_data` -> `io` change can be a different PR? I have renamed each module to group all confounds stuff and any suggestions are welcome!

NicolasGensollen:
>Maybe the input_data -> io change can be a different PR?

I agree, this should be addressed in a different PR

nilearn/input_data/confounds_parser.py
```diff
@@ -0,0 +1,305 @@
...
+    The implementation will only support fMRIPrep output from the 1.2.x series.
+    The `compcor` noise component requires 1.4.x series or above.
+
+    Parameters
+    ----------
+    img_files : path to processed image files, optionally as a list.
+        Processed nii.gz/dtseries.nii/func.gii file from fmriprep.
+        `nii.gz` or `dtseries.nii`: path to files, optionally as a list.
+        `func.gii`: list of a pair of paths to files, optionally as a list
+        of lists. The companion tsv will be automatically detected.
```
bthirion, line 92: Isn't there any way to document properly how the  tsv file **should** be organized, beyond "being the one output by fmriprep" ? If there is such a strong adherence to fmriprep that it becomes futile to describe the input file, then we should explictcly relate to the fmriprep dependency in the name of the function.

nilearn/input_data/confounds_utils.py
```diff
@@ -0,0 +1,250 @@
+"""Helper functions for the manipulation of confounds.
```
bthirion, line 1: I think that the dependency of all this on fmriprep needs to be made explicit, even possibly in module/function name.
AFAIK not everybody uses fmriprep and this part of the codebase has a lot of implicit assumption on fmriprep outputs.

bthirion (commented):
This looks pretty good, congratulation for the huge work.
I still have a few comments, but we're close !

nilearn/input_data/confounds_utils.py
htwangtw: @pbellec and I talked about this a bit. It should be more explicit from the module or function name. Instead of using `confounds` as prefix, I thought about `fmriprep` and adjust some of the module names accordingly. For the main function, maybe we should use `fmriprep_confounds` or `fmriprep_denoise` instead of `load_confounds`
WDYT?

pbellec: After sleeping on it for a while, I am absolutely in favor of making the tie with `fmriprep` explicit. Maybe in the future this code will support any pipeline compliant with `bids` derivatives, but in the meantime the code is specifically designed to operate on `fmriprep` outputs (even supporting multiple versions of this software). I would be in favor of using `fmriprep_confounds` rather than `fmriprep_denoise`, because `confounds` is the name used throughout the documentation of `fmriprep` and strictly speaking the confounds do not perform denoising by themselves, but are rather fed into a denoising procedure (`signal.clean`).

bthirion: I think we're on the same page (and indeed, I changed my mind over this question). `fmriprep_confounds` is probably the best term.

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -73,9 +74,82 @@
...
+np.fill_diagonal(correlation_matrix, 0)
+
+plotting.plot_matrix(correlation_matrix, figure=(10, 8), labels=labels[1:],
+                     vmax=0.8, vmin=-0.8, title='Minimal', reorder=True)
+
+# Now let's create a set with motion scrubbing
+confounds_minimal_scrub, sample_mask = load_confounds(
+    fmri_filenames,
+    strategy=["high_pass", "motion", "wm_csf", "scrub"],
+    motion="basic", wm_csf="basic", scrub="full")
```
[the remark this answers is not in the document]
htwangtw: I agree. I rewrite the example a bit with more explanation. One other thing is, currently there are some references missing and/or not formatted as how nilearn does it. I can either address the issues with the docs and how references are linked in a different PR or in this one. I don't have a strong opinion on this one.

nilearn/input_data/confounds_parser.py
```diff
@@ -0,0 +1,305 @@
...
+                   compcor="anat_combined", n_compcor="auto",
+                   ica_aroma="full",
+                   demean=True):
+    """
+    Use confounds from fmriprep.
+
+    To enable easy confound variables loading from fMRIprep outputs, Confounds
+    provides a interface that groups subsets of confound variables into noise
+    components and their parameters. It is possible to fine-tune a subset of
+    noise components and their parameters through Confounds.
```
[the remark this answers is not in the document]
htwangtw: This whole block of text needs some rewrite. Thanks for catching it!

nilearn/input_data/confounds_parser.py
htwangtw: This is a good point. We technically didn't write the test on a full fmriprep functional derivative output layout. As long as the image file and the associated confound tsv are in the same directory, things are fine. Definitely should add more details here.

nilearn/input_data/confounds_parser.py
```diff
@@ -0,0 +1,305 @@
...
+    ica_aroma : {'full', 'basic'}
+
+        - "full": use fMRIprep output `~desc-smoothAROMAnonaggr_bold.nii.gz`.
+        - "basic": use noise independent components only.
+
+    demean : boolean, default True
+        If True, the confounds are standardized to a zero mean (over time).
+        This step is critical if the confounds are regressed out of time series
+        using nilearn with no or zscore standardization, but should be turned
+        off with "spc" normalization.
```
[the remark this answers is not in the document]
htwangtw: Firstly there's a type (psc rather than spc) :sweat_smile: 
Thanks for raising this! I found a potential bind spot in how we simulate the signal for testing when looking into this.
Had a closer look at the code of `signal.clean` - the confounds will be standardised or scaled before denoising. Whether the output here is demeaned here doesn't seem to matter following the logic. 
I might need to dig into this further...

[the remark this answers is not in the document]
pbellec: I think this comment is deprecated since this issue has been fixed https://github.com/nilearn/nilearn/issues/2584
I think we should remove this part of the documentation. 
However I would be tempted to keep the tests which check that confounds are properly regressed with different types of normalization - just to prevent an issue with normalization from coming back in the future.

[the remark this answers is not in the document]
pbellec: actually the tests in load_confounds may be redundant with the ones that where added in this PR https://github.com/nilearn/nilearn/pull/2609 I haven't yet had to look in details, but we may want to avoid duplicating tests.

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -0,0 +1,306 @@
...
+        :func:`nilearn.signal.clean`.
+        Volumns are removed if flagged as following:
+
+        - Non-steady-state volumes (if present)
+        - Motion outliers detected by scrubbing
+
+    Notes
+    -----
+    The noise components implemented in this class are
+    adapted from (Ciric et al. 2017). Band-pass filter is replaced
```
NicolasGensollen, line 223: Since there are several scientific references, we could improve the docstring by adding the BibTex refs to `nilearn/doc/references.bib`.

You would then only need to replace this line by:

```
adapted from :footcite:`Ciric2017`. Band-pass filter is replaced
```

(assuming the bibtex entry is `Ciric2017`).

And the References section will simply become:

```
References
-----------
.. footbibliography::
```

(nilearn/datasets/atlas.py is formatted this way if you want an example).

The references will then be much easier to maintain, always formatted properly, and appear in the main Nilearn's [bibliography](https://nilearn.github.io/bibliography.html).
WDYT?

nilearn/input_data/fmriprep_confounds_utils.py
```diff
@@ -0,0 +1,250 @@
...
+    return pd.DataFrame(confounds, columns=confound_cols)
+
+
+class MissingConfound(Exception):
+    """
+    Exception raised when failing to find params in the confounds.
+
+    Parameters
+    ----------
+        params : list of missing params
```
NicolasGensollen, line 241: These lines are over-indented

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,502 @@
...
+
+    # Shift the mean to non-zero
+    vol = vol + 100
+
+    # create an nifti image with the data, and corresponding mask
+    img = Nifti1Image(vol, np.eye(4))
+    mask_conf = Nifti1Image(vol_conf, np.eye(4))
+    mask_rand = Nifti1Image(vol_rand, np.eye(4))
+
+    # generate the assoicated confounds for testing
```
NicolasGensollen, line 63:
```suggestion
    # generate the associated confounds for testing
```

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,502 @@
...
+def _corr_tseries(tseries1, tseries2):
+    """Compute the correlation between two sets of time series."""
+    corr = np.zeros(tseries1.shape[1])
+    for ind in range(tseries1.shape[1]):
+        corr[ind], _ = pearsonr(tseries1[:, ind], tseries2[:, ind])
+    return corr
+
+
+def _regression(confounds, sample_mask, tmp_path):
+    """Simple regression with nilearn."""
```
NicolasGensollen, line 112: This feels weird now, maybe "Simple regression with NiftiMasker"?

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,502 @@
...
+        ("wm_csf", {"wm_csf": "full"}),
+        ("global", {"global_signal": "full"}),
+        ("compcor", {}),
+        ("compcor", {"compcor": "anat_separated"}),
+        ("compcor", {"compcor": "temporal"}),
+        ("ica_aroma", {"ica_aroma": "basic"}),
+    ],
+)
+def test_nilearn_regress(tmp_path, strategy, param):
+    """Try regressing out all motion types in nilearn."""
```
NicolasGensollen, line 140:
```suggestion
    """Try regressing out all motion types."""
```

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,502 @@
...
+    img_nii, _ = create_tmp_filepath(
+        tmp_path, copy_confounds=True, copy_json=True
+    )
+    confounds, _ = fmriprep_confounds(img_nii, strategy=[strategy], **param)
+    _regression(confounds, None, tmp_path)
+
+
+@pytest.mark.filterwarnings("ignore")
+def test_nilearn_standardize_false(tmp_path):
+    """Test removing confounds in nilearn with no standardization."""
```
NicolasGensollen, line 150:
```suggestion
    """Test removing confounds with no standardization."""
```

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,502 @@
...
+
+    # Check that most variance is preserved
+    # in voxels composed of random noise
+    tseries_std = _tseries_std(img, mask_rand, X, None, False)
+    assert np.mean(tseries_std > 0.9)
+
+
+@pytest.mark.filterwarnings("ignore")
+def test_nilearn_standardize_zscore(tmp_path):
+    """Test removing confounds in nilearn with zscore standardization."""
```
NicolasGensollen, line 167:
```suggestion
    """Test removing confounds with zscore standardization."""
```

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,502 @@
...
+    # with vs without confounds in voxels where the signal is uncorrelated
+    # with confounds. The correlation before and after denoising should be very
+    # high as very little of the variance is removed by denoising
+    tseries_raw, tseries_clean = _denoise(img, mask_rand, X, None, "zscore")
+    corr = _corr_tseries(tseries_raw, tseries_clean)
+    assert corr.mean() > 0.8
+
+
+def test_nilearn_standardize_psc(tmp_path):
+    """Test removing confounds in nilearn with psc standardization."""
```
NicolasGensollen, line 190:
```suggestion
    """Test removing confounds with psc standardization."""
```

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,502 @@
...
+    conf, _ = fmriprep_confounds(
+        img_nii, strategy=["compcor"], compcor="anat_combined", n_compcor=2
+    )
+    assert "a_comp_cor_00" in conf.columns
+    assert "a_comp_cor_01" in conf.columns
+    assert "a_comp_cor_02" not in conf.columns
+
+
+def test_not_found_exception(tmp_path):
+    """Check various file or parameter missing scenrio."""
```
NicolasGensollen, line 270:
```suggestion
    """Check various file or parameter missing scenario."""
```

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,502 @@
...
+    np.savetxt(bad_conf, fake_confounds, delimiter="\t")
+    with pytest.raises(ValueError) as error_log:
+        fmriprep_confounds(bad_nii)
+    assert "The confound file contains no header." in str(error_log.value)
+
+    # invalid fmriprep version: old camel case header (<1.2)
+    leagal_confounds, _ = get_leagal_confound()
+    camel_confounds = leagal_confounds.copy()
+    camel_confounds.columns = [
+        to_camel_case(col_name) for col_name in leagal_confounds.columns
```
NicolasGensollen, line 402: See my comment on this function being private (or not?)

NicolasGensollen (commented):
Thanks @htwangtw !
I made a few suggestions below, mainly on documentation and formatting. I haven't looked for duplicated tests (as @pbellec mentioned), but will take a look at that later today.

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -71,21 +72,123 @@
...
 
-# Mask the main diagonal for visualization:
 np.fill_diagonal(correlation_matrix, 0)
 
 plotting.plot_matrix(correlation_matrix, figure=(10, 8), labels=labels[1:],
                      vmax=0.8, vmin=-0.8, title='No confounds', reorder=True)
 
+##############################################################################
+# Load confounds from file using a flexible strategy with fmriprep_confounds
+# ----------------------------------------------------------------------
```
NicolasGensollen, line 98: Needs more '-' to avoid sphinx warnings

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -0,0 +1,306 @@
...
+        Processed nii.gz/dtseries.nii/func.gii file reside in a fmriprep
+        generated functional derivative directory (i.e.The associated confound
+        files should be in the same directory as the image file). As long as
+        the image file, confound related tsv and json are in the same
+        directory with BIDS-complied names, `fmriprep_confounds` can retreive
+        the relevant files correctly.
+
+        - `nii.gz` or `dtseries.nii`: path to files, optionally as a list.
+        - `func.gii`: list of a pair of paths to files, optionally as a list
+        of lists.
```
NicolasGensollen, line 96:
```suggestion
          of lists.
```

NicolasGensollen (commented):
Just two small things currently causing sphinx warnings and doc build failures in strict mode:

nilearn/input_data/fmriprep_confounds.py
htwangtw: Thanks for providing the instructions!

htwangtw:
I have rewrite the example text a bit. I hope this would explain the utility of the function better.
I might need to add some references of the components mentioned, or it would be fine to direct users to the function docstring?

For the simulated data generated in the test - 
Originally we used the `fmriprep_confounds` function to load six motion parameters, demean the confounds based on the parameters passed, create simulated signal with noise, and return the simulated data and the confounds.
I have modified this to make sure the simulated signal are always from the raw confounds output, but the confounds returned would have the demean option applied. Looks like numerically it is not making enough changes to break the test. 
This makes me wonder, should we remove the demean option altogether, keep `fmriprep_confounds` as a simple tool that get the data from fMRIprep, and let `signal.clean` / `NiftiMasker` handle all the manipulation on the data?

pbellec:
@htwangtw sorry for the delay. The reason why I put some demeaning in load_confounds a long time ago place was because otherwise it did not necessarily play well with the nilearn maskers. But since that time, normalization of confounds in the maskers have improved, and I don't think the normalization in load_confounds is useful anymore. It certainly will be less confusing for the users not to be exposed to normalization parameters twice (once in the confound loader, once in the masker). Especially since in 99.9% of the case people will just want to use the default (and still not sure what the 0.1% remaining would look like). Anyway, in short, I am in favor of removing normalization options entirely.

bthirion:
I may be a bit off, but as far as I can see, demeaning never hurts. I'd keep it.

htwangtw:
I have tried to implement a version with the demean option removed looks like, can always roll back.

An alternative would be keeping the option as @bthirion suggested, but change the default to `False` and let masker and signal.clean to handle stuff? As @pbellec suggested, the normalisation issue in nilearn has been resolved and AFAIK numerically it didn't break any of the tests we had.

htwangtw:
I have left the test broken intentionally as a reminder for further investigation. 

I have dived into the demean option and test duplication issue a bit more and found something interesting...
**If demean is set to false and let signal.clean handle everything, `test_nilearn_standardize` breaks.** 

In this test we evaluate the confound removal through correlation between signal containing confounds with and without denoising. The correlation of the two should be small.  
Another check removes confounds from pure random noise, the signal cleaning should achieve nothing. The correlation should be large. If `demean=False`, this check fails no matter how signal and confounds are standardised or not in the `signal.clean` options. 

Would like to know what you think to move this forward @bthirion  @NicolasGensollen @pbellec

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -0,0 +1,300 @@
...
+        of components per mask.
+        "all": select all components (50% variance explained by fMRIPrep
+        defaults)
+
+    ica_aroma : {'full', 'basic'}
+
+        - "full": use fMRIprep output `~desc-smoothAROMAnonaggr_bold.nii.gz`.
+        - "basic": use noise independent components only.
+
+    demean : boolean, default True
```
NicolasGensollen, line 191: I initially thought that this should be renamed to `standardize_confounds` to match nilearn's naming, but looking at the function `_demean_confounds` it seems that only the mean of kept volumes is subtracted. 
But, I believe that when `signal.clean` has `standardize_confounds` set to True, it also standardizes wrt std.
Do you think this difference could be problematic?

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -37,12 +40,13 @@
       atlas_filename)  # 4D data
 
 # One subject of brain development fmri data
-data = datasets.fetch_development_fmri(n_subjects=1)
+data = datasets.fetch_development_fmri(n_subjects=1, reduce_confounds=True)
 fmri_filenames = data.func[0]
+reduced_confounfs = data.confounds[0]  # This is a preselected set of confounds
```
NicolasGensollen, line 45:
```suggestion
reduced_confounds = data.confounds[0]  # This is a preselected set of confounds
```

examples/03_connectivity/plot_signal_extraction.py
```diff
@@ -51,12 +55,11 @@
 # Here we go from nifti files to the signal time series in a numpy
 # array. Note how we give confounds to be regressed out during signal
 # extraction
-time_series = masker.fit_transform(fmri_filenames, confounds=data.confounds)
-
+time_series = masker.fit_transform(fmri_filenames, confounds=reduced_confounfs)
```
NicolasGensollen, line 58:
```suggestion
time_series = masker.fit_transform(fmri_filenames, confounds=reduced_confounds)
```

NicolasGensollen (commented):
Thanks for addressing all the reviews @htwangtw ! :+1: 
I had a look at the broken test and I also think this is linked to `standardize_confounds` being set to False but I'm still not successful at understanding where the issue is.
I'll try to look more closely but please see if my comment below is related or not.

nilearn/input_data/fmriprep_confounds.py
htwangtw: > But, I believe that when `signal.clean` has `standardize_confounds` set to True, it also standardizes wrt std.
> Do you think this difference could be problematic?

I have the same question as well, and that definitely seems weird to me.

I tried to demean the confounds and set `standardize_confounds=False`, and everything got weird....

htwangtw: I think I have found the reason why some combinations of parameters fails. It's related to how `detrend` works in `signal.clean`. I will start with a TL;DR - if we want `fmriprep_confounds` to work with `signal.clean` default settings correctly, the default value of the `demean` option should be `False`. Otherwise the confounds would be demeaned twice. 

For `fmriprep_confounds`, `demean=False`, all combination of the relevant parameters in `signal.clean`:
|standardize_confounds|standardize|detrend|test_fmriprep_confounds|
|--|--|--|--|
|True|zscore|True|PASSED|
|True|zscore|linear|PASSED|
|True|zscore|False|PASSED|
|True|psc|True|FAILED|
|True|psc|linear|FAILED|
|True|psc|False|PASSED|
|True|False|True|PASSED|
|True|False|linear|PASSED|
|True|False|False|PASSED|
|False|zscore|True|PASSED|
|False|zscore|linear|PASSED|
|False|zscore|False|FAILED|
|False|psc|True|FAILED|
|False|psc|linear|FAILED|
|False|psc|False|FAILED|
|False|False|True|PASSED|
|False|False|linear|PASSED|
|False|False|False|FAILED|

AFAIK the `signal.clean` is doing what it's supposed to do, my intuition is some combination of options probably shouldn't be allowed. This issue is a bit out of the scope of `fmriprep_denoise` but something for `signal.clean`.

htwangtw:
I was thinking about the most pragmatic way to move this PR forward. The functions related to getting the right confounds are  fine, and we still have preset strategies to be added after this one. 
The issue we are dealing with is really with `signal.clean`, and probably needs another discussion. 
I am perfectly happy to continue looking into `signal.clean` and make sure we have something sensible to the users. 
Meanwhile, what about we separate the tests related to the performance of `signal.clean` from the tests on the confounds variable parsing?  WDYT?

bthirion:
> what about we separate the tests related to the performance of `signal.clean` from the tests on the confounds variable parsing?

Sorry, I don't understand well your suggestion. Can you elaborate a little bit ?

bthirion:
I'm happy to move forward if at laast default settings lead to a reasonable outcome.
I'm not sure it is the case atm.

And most importantly: thx for digging into that. These are rather complex procedures. The problem is that the parameters of signal.clean  are meant to describe processing applied to the fMRI stuff. But we later considered that these should aply as well to confounding signals. The problem is that there are indeed corner cases where the logic is broken ...

htwangtw:
> > what about we separate the tests related to the performance of `signal.clean` from the tests on the confounds variable parsing?
> 
> Sorry, I don't understand well your suggestion. Can you elaborate a little bit ?

Most of the tests are just checking if the confound variables grabbed from the fMRIprep output is correct and the raw output is in the right type to pass to `signal.clean`. The other tests are checking if `signal.clean` really regress out the confounds from simulated signal. Right now they are all in the same test file. I am thinking of moving the latter kind of tests to a separate file, and include test cases on the default option of `signal.clean`. Does this sound good?

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -0,0 +1,304 @@
...
+        - Non-steady-state volumes (if present)
+        - Motion outliers detected by scrubbing
+
+    Notes
+    -----
+    The noise components implemented in this class are adapted from
+    :footcite:`Ciric2017`. Band-pass filter is replaced by high-pass filter.
+    Low-pass filters can be implemented, e.g., through `NifitMaskers`. Other
+    aspects of the preprocessing listed in :footcite:`Ciric2017` are controlled
+    through fMRIprep, e.g. distortion correction.
```
NicolasGensollen, line 227:
```suggestion
    through :term:`fMRIPrep`, e.g. distortion correction.
```

NicolasGensollen (commented):
Thanks for investigating this @htwangtw ! :+1: 

>Another parameter that might have interfered things would be detrend in signal.clean. This parameter would apply to both confounds and the signal with the current implementation.

Indeed, `detrend` will demean both the signal and confounds with the current implementation  of `signal.clean`.

>the default value of the demean option should be False. Otherwise the confounds would be demeaned twice.

I agree that doing `demean=True` in `load_fmriprep_confounds` and then `detrend=True` in `signal.clean` will lead to confounds being demeaned twice. However, I might be missing something because I don't see why this would be problematic. The second mean subtraction will have almost no effect, right? :thinking: 
Anyway, I don't think it is problematic to have the confounds demeaned only in `signal.clean` by default.

By looking at your table, it seems also that `standardize='psc'` doesn't behave very well with the other parameters. Does your correlation test make sense with this option?

Also thank you for opening the issue on the default values for `signal.clean` and `NiftiMasker`. I must admit that I don't know why they are different and whether this was done on purpose or not. I agree that, if there is no particular reason for these differences, we should have the same defaults.

I made a few small suggestions below, mainly referencing fmriprep since it was recently added to the glossary.

htwangtw:
> I agree that doing `demean=True` in `load_fmriprep_confounds` and then `detrend=True` in `signal.clean` will lead to confounds being demeaned twice. However, I might be missing something because I don't see why this would be problematic. The second mean subtraction will have almost no effect, right? thinking
> Anyway, I don't think it is problematic to have the confounds demeaned only in `signal.clean` by default.

Yes you are right. I admit I was getting confused with things :sweat_smile:  
After sleeping on this for a bit, because any kinds of normalisation/demean should be done on the confounds when performing regression on the signal, the default of demean should be set to `True` just to be on the safe side, and the defaults of `detrend` and `standard_confounds` in `NiftiMasker` and `signal.clean` doesn't really have an affect on the confounds part. 

> By looking at your table, it seems also that `standardize='psc'` doesn't behave very well with the other parameters. Does your correlation test make sense with this option?

I will need to think about this one and have a closer look about why this failed.

htwangtw:
I had a look into how we simulate the signal and found the issue for `psc`: we added 100 to the noise/signal mixture to shift the mean, and that results in extremely small value (in the range of e-14) after standardising signal as percent signal change.  After reducing the constant from 100 to 10, the results became more logical: as long as some kind of standardisation is performed on the confounds, the signal can be cleaned correctly.

|`standardize_confounds`|`detrend`|`standardize`|`test_fmriprep_confounds`|
|--|--|--|--|
|True|True|zscore|PASSED|
|True|True|psc|PASSED|
|True|True|False|PASSED|
|True|False|zscore|PASSED|
|True|False|psc|PASSED|
|True|False|False|PASSED|
|False|True|zscore|PASSED|
|False|True|psc|PASSED|
|False|True|False|PASSED|
|False|False|zscore|FAILED|
|False|False|psc|FAILED|
|False|False|False|FAILED|

I can submit another issue on signal clean to address this - meanwhile I will modify the test here limited to the logical options.

bthirion:
> Most of the tests are just checking if the confound variables grabbed from the fMRIprep output is correct and the raw output is in the right type to pass to `signal.clean`. The other tests are checking if `signal.clean` really regress out the confounds from simulated signal. Right now they are all in the same test file. I am thinking of moving the latter kind of tests to a separate file, and include test cases on the default option of `signal.clean`. Does this sound good?

It makes sense to split the tests if they become too long or heavy, hence for the sake of readability. This may not be a high priority.

NicolasGensollen:
Thanks @htwangtw ! :+1: 

>we added 100 to the noise/signal mixture to shift the mean, and that results in extremely small value (in the range of e-14) after standardising signal as percent signal change.

I agree that adding 100 will result in very small values when doing psc standardization. However, I'm not sure to understand why this was making the correlation test fail (numerical precision maybe...?).

>I can submit another issue on signal clean to address this

We can open an issue if we think this test should pass with non standardized confounds, right?
Looking at your last table, I'm not sure that standardizing the signal but not the confounds makes sense. And the extreme case where nothing is standardized nor detrended doesn't make more sense to me. WDYT?

I'll make another pass today but I think this is almost ready to merge.

htwangtw:
> It makes sense to split the tests if they become too long or heavy, hence for the sake of readability. This may not be a high priority.

@bthirion I think I will keep it as it is for now...

> I agree that adding 100 will result in very small values when doing psc standardization. However, I'm not sure to understand why this was making the correlation test fail (numerical precision maybe...?).

@NicolasGensollen I admit that was a duck tape solution. The correlation of the raw and cleaned signal was nan for some voxels and it was not completely consistent. My guess was a precision issue but no further investigation was done.

> We can open an issue if we think this test should pass with non standardized confounds, right?
Looking at your last table, I'm not sure that standardizing the signal but not the confounds makes sense. And the extreme case where nothing is standardized nor detrended doesn't make more sense to me. 

Non standardised signal with standardised confounds is a mind twister for interpretation, but I won't say it's producing undesired results for the purpose of removing confounds. I didn't include them in the test, just including them in this table for the sake of completion.  
I am thinking about forcing confounds to be detrended or standardised, but I am not sure if it's better to raise error or remove that parameter entirely to deal with confounds internally. (Probably better to move this to an separated issue at this stage)

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -0,0 +1,304 @@
...
+    To enable easy confound variables loading from :term:`fMRIPrep` outputs,
+    `fmriprep_confounds` provides an interface that groups subsets of confound
+    variables into noise components and their parameters. It is possible to
+    fine-tune a subset of noise components and their parameters through this
+    function.
+
+    The implementation will only support :term:`fMRIPrep` functional derivative
+    directory from the 1.2.x series. The `compcor` noise component requires
+    1.4.x series or above.
+
```
NicolasGensollen, line 83:
```suggestion
    .. versionadded:: 0.8.2
    
```

nilearn/input_data/tests/test_fmriprep_confounds.py
```diff
@@ -0,0 +1,524 @@
...
+from nilearn.input_data import NiftiMasker
+from nilearn.input_data import fmriprep_confounds
+from ..fmriprep_confounds import _sanitize_strategy
+
+from nilearn._utils.fmriprep_confounds import _to_camel_case
+
+from .utils import create_tmp_filepath, get_leagal_confound
+
+
+def _simu_img(tmp_path, demean):
```
NicolasGensollen, line 17: I think we should set a random state in these tests to avoid spurious test failures.

NicolasGensollen (commented):
Just a couple small things, but looks good to me otherwise, thanks! :+1: 
Could you add a whats_new entry and trigger a full build?

nilearn/input_data/tests/test_fmriprep_confounds.py
htwangtw: I was looking into that - will do!

htwangtw:
All builds passed :tada: I will implement the four preset strategy next.
cc'd @pbellec

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -0,0 +1,306 @@
...
+        - "basic" remove time frames based on excessive framewise displacement
+          and DVARS.
+        - "full" also remove continuous segments containing fewer than 5
+          volumes.
+
+        One-hot encoding vectors are added as regressors for each scrubbed
+        frame.
+
+    compcor : {'anat_combined', 'anat_separated', 'temporal',
```
NicolasGensollen, line 163:
```suggestion
    compcor : {'anat_combined', 'anat_separated', 'temporal',\
```

NicolasGensollen (commented):
Thanks @htwangtw !
Just a minor formatting suggestion but LGTM otherwise.

NicolasGensollen (approved):
Thanks @htwangtw ! :+1: 
I believe this is ready to merge. I think we can wait a day or two to give a chance for more reviews.
Has anyone planned to make another review? (most likely @bthirion)

bthirion:
Will do asap...

NicolasGensollen:
Thanks everyone for the work on this! :100: 
Merging... :tada:

pbellec:
Congrats @htwangtw for driving this PR home, and big thanks to the `load_confounds` contributors and the nilearn team for working with us on the code base, especially @NicolasGensollen and @bthirion. I hope this feature will be valuable to the community.

PR #2952, opened 2021-09-08
[FIX] LGTM error: Missing call to `__init__` during object initialization

DimitriPapadopoulos:
Class `SimpleRegressionResults` may not be initialized properly as method `LikelihoodModelResults.__init__` is not called from its `__init__` method.

The code is not exactly equivalent, so I need your input here. Method `LikelihoodModelResults.__init__` looks like this:
```python
        self.theta = theta
        self.Y = Y
        self.model = model
        if cov is None:
            self.cov = self.model.information(self.theta,
                                              nuisance=self.nuisance)
        else:
            self.cov = cov
        self.dispersion = dispersion
        self.nuisance = nuisance

        self.df_total = Y.shape[0]
        self.df_model = model.df_model
        # put this as a parameter of LikelihoodModel
        self.df_residuals = self.df_total - self.df_model
```
The old `SimpleRegressionResults.__init__` used to differ on the `cov` test: and `Y` and `model` were not set:
```python
        self.theta = results.theta
        self.cov = results.cov
        self.dispersion = results.dispersion
        self.nuisance = results.nuisance

        self.df_total = results.Y.shape[0]
        self.df_model = results.model.df_model
        # put this as a parameter of LikelihoodModel
        self.df_residuals = self.df_total - self.df_model
```
Was this on purpose.?

### DIFF
--- a/nilearn/glm/regression.py
+++ b/nilearn/glm/regression.py
@@ -469,15 +469,9 @@ class SimpleRegressionResults(LikelihoodModelResults):
         The only difference is that the whitened Y and residual values
         are stored for a regression model.
         """
-        self.theta = results.theta
-        self.cov = results.cov
-        self.dispersion = results.dispersion
-        self.nuisance = results.nuisance
-
-        self.df_total = results.Y.shape[0]
-        self.df_model = results.model.df_model
-        # put this as a parameter of LikelihoodModel
-        self.df_residuals = self.df_total - self.df_model
+        LikelihoodModelResults.__init__(self, results.theta, results.Y,
+                                        results.model, results.cov,
+                                        results.dispersion, results.nuisance)
 
     def logL(self, Y):
         """


bthirion:
> Was this on purpose.?

Yes, the idea it to have a memory-efficient version. There's actually no good reason to store Y as part of the results.

DimitriPapadopoulos:
I see. When the derived class actually does **less** than the base class, there's no easy way to "fix" this LGTM.com alert. We could add an argument to method `LikelihoodModelResults.__init__()`, such as `storeY` boolean, but that would be a bit artificial.

Anyway, I can understand that `SimpleRegressionResults.__init__()` does not store `model` and `Y`, therefore these two lines from `LikelihoodModelResults.__init__()` are removed:
```python
        self.Y = Y
        self.model = model
```
However, what about the difference in initializing `cov`? In `SimpleRegressionResults.__init__()`:
```python
        self.cov = results.cov
```
In `LikelihoodModelResults.__init__()`:
```python
        if cov is None:
            self.cov = self.model.information(self.theta,
                                              nuisance=self.nuisance)
        else:
            self.cov = cov
```
Is that on purpose?

The real problem is that the documentation of `SimpleRegressionResults` seems to have been copied as is from `RegressionResults` and might need to be updated:
```python
        """See LikelihoodModelResults constructor.

        The only difference is that the whitened Y and residual values
        are stored for a regression model.
        """
```
Should I open an issue about the documentation of `SimpleRegressionResults.__init__()`?

bthirion:
Please do. Thx.

DimitriPapadopoulos:
I'll close this PR, I cannot find an easy solution to work around the LGTM.com error.

PR #2953, opened 2021-09-09
[INFRA] Replace individual requirements files with setup.cfg

### LINKED ISSUE #2947: Bundle dependencies in `setup.py` or `info.py` file
I think it would be beneficial to combine the requirements-[*].txt files into a single file, with flags for the different dependency groups.

### Benefits to the change

Installation for developers would be simplified, since `nilearn` and its developer-specific dependencies could be installed in one step (e.g., `pip install -e .[all]` or `pip install -e .[test,doc]`). 

From a maintenance perspective, it may also be easier to manage one file instead of three.

We could put the requirements in the `setup.cfg` file ([as in `pybids`](https://github.com/bids-standard/pybids/blob/00ee1ecd2d986f150b95710072d019414769a35e/setup.cfg#L22-L68)), the `setup.py` file, or a new `info.py` file ([as in `tedana`](https://github.com/ME-ICA/tedana/blob/ab7afd80defafc86a4e4f27bd88dd2b32e499adb/tedana/info.py#L27-L62)).

Current steps for developer installation:

```bash
git clone git@github.com:<your_username>/nilearn.git
conda create -n nilearn
conda activate nilearn
pip install . -e
pip install -r requirements-dev.txt
```

New steps:

```bash
git clone git@github.com:<your_username>/nilearn.git
conda create -n nilearn
conda activate nilearn
pip install .[dev] -e
```

tsalo:
Closes #2947.

- Move requirements from `requirements-min.txt`, `requirements-dev.txt`, and `requirements-build-docs.txt` to `setup.cfg`.
- Remove dependency information from `setup.py` (it will be extracted from `setup.cfg` automatically).
- Replace references to txt requirements files with appropriate references to `setup.cfg` file or standard installation.
- ~~Move `nilearn.version._check_module_dependencies` to `nilearn.__init__._check_dependencies`, which also checks for versions without hardcoded versions in `nilearn.version`. This way, the only place where requirements and their associated versions are written is in `setup.cfg`.~~
    - I dropped this change because there were other parts of `nilearn` that relied on this check and the hardcoded dependency versions in `version.py`.

To do:

- [x] Remove the requirements.txt files.
- [x] `azure-pipelines.yml`
- [x] `build_tools/github/dependencies.sh`
- [x] `build_tools/github/install.sh`?
- [X] `docs/binder/requirements.txt`
- [x] `development.rst` (once contributing docs PR is merged)
- [x] Identify any duplicates/conflicts in the dependency groups within `setup.cfg` and deal with them.
- [x] `nilearn/version.py`

What I wasn't able to address in this PR:
- Hardcoded dependency versions in the code files. Preferably, dependencies and any version restrictions would be documented in one location: `setup.cfg`. However, there are several locations in the code that check those hardcoded versions in `version.py`.
- Installation instructions that de-emphasize installing dependencies ahead of time and focus on using dependency groups to install everything you need for what you want to use nilearn for.
- CircleCI installs both testing- and documentation-related dependencies for its separate testing and documentation-related jobs. This may be somewhat inefficient.
- The conda `environment.yml` file. I don't know if this can reference the `setup.cfg` or not, but I'd love to have it do that instead of include hardcoded packages.
- `maint_tools/show-python-packages-versions.py` includes hardcoded package names. It would be great to just check against the requirements in the packaging, but I don't think it's possible to check anything except the requirement dependencies, which matplotlib isn't one of.


setup.cfg
```diff
@@ -3,6 +3,76 @@
 [bdist_rpm]
 doc-files = doc
 
+[options]
+python_requires = >=3.6
+install_requires =
```
tsalo, line 8: I noticed that there are a lot of duplicates between the three dependency groups (min, dev, and build-docs). Also, some seem to be in the wrong group(s). For example, `pytest`, `pytest-cov`, and `codecov` probably shouldn't be in the minimal dependency group.

thomasbazeille:
The new development.rst contains separates ref to both requirements to help contributors set-up their environment and must be changed accordingly.

tsalo:
In `azure-pipelines.yml`, `build_tools/github/dependencies.sh`, and `build_tools/github/install.sh`, the installation of dependencies is separated from the installation of nilearn. Is that important? Otherwise, it would be easier to install nilearn with its dependencies with `pip install -e .[dev]` or something similar.

tsalo:
@thomasbazeille thanks! I'll add a note to the to-do list. I assume this will have to wait until the contributing docs PR is merged.

setup.cfg
```diff
@@ -3,6 +3,76 @@
...
+	numpy==1.16
+	scipy==1.2.0
+	pandas==0.24.0
+	scikit-learn==0.21.0
```
tsalo, line 12: Some CI jobs are failing due to the following:

>No local packages or working download links found for scikit-learn==0.21.0
>error: Could not find suitable distribution for Requirement.parse('scikit-learn==0.21.0')

Perhaps we need to change the pinned sklearn version?

NicolasGensollen: @tsalo I haven't looked in details but scikit-learn 0.21.0 is only installed under python 3.6:

https://github.com/nilearn/nilearn/blob/cc99fa76df137e9ac6010c311e15fa9ac12ce66d/.github/workflows/testing.yml#L30-L39

setup.cfg
tsalo: An initial proposal:
1. Remove `codecov`, `pytest`, `pytest-cov` from core dependencies.
2. Add `matplotlib`, `requests`, and maybe `plotly` to core dependencies.
3. Remove `coverage` and `flake8` from documentation dependencies.

build_tools/github/install.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash -e
 
 python setup.py build
-python setup.py install
+if [ ! -z "$MIN_REQUIREMENTS" ]; then
+    pip install --progress-bar off --upgrade -e .
```
NicolasGensollen, line 5: @tsalo I'm not sure to follow. I might be wrong, but by looking at the PR it looks like if you do `pip install .` you will install all dependencies in their minimum versions and that you'd have to do `pip install .[dev]` to get the latest versions.
Atm, the minimum dependencies describe the oldest possible configuration that we support (Python 3.6 and minimum deps for all packages).
To me you should be able to pip install with all versions of python above 3.6 without the dev flag (which should only install additional specific packages). Does this make sense?

tsalo: My assumption was that "minimum" dependencies referred to those packages which were necessary to use nilearn, while "dev" was the ones necessary for developers (i.e., additional packages for testing, documentation-building, etc.).

I'm not used to trying to pin different _versions_ of the same packages in different dependency groups. However, I do think it's possible, as long as the minimum group's versions don't conflict with the regular pinned versions.

tsalo: I've added a `min` group with the minimum versions pinned (`==`), while the main dependency group has `>=`. Does that work?

tsalo: Never mind, that doesn't work. Trying to install the min version with `pip install -e .[min]` results in the following:

```
Collecting pandas>=0.24.0
  Using cached pandas-0.24.0.tar.gz (11.8 MB)
ERROR: Cannot install nilearn[min]==0.8.1.dev0 because these package versions have conflicting dependencies.

The conflict is caused by:
    nilearn[min] 0.8.1.dev0 depends on scikit-learn>=0.21.0
    nilearn[min] 0.8.1.dev0 depends on scikit-learn==0.21.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

Even though the versions do not actually clash, having two pins like that apparently is not allowed.

setup.cfg
tsalo: That makes sense. It seems like we should at least change to 0.22 then, right?

build_tools/github/install.sh
tsalo: Actually, it works. The problem was specifically with scikit-learn==0.21.0, as we realized in the previous thread.

CONTRIBUTING.rst
```diff
@@ -68,7 +68,7 @@ Installing the latest code
 
 In order to ensure that any code changes are reflected in your installation, navigate to your cloned Nilearn base directory and install using the following command::
 
-    pip install -e .
+    pip install -e .[dev]
```
bthirion, line 71: I think it's important that you explain here, at least briefly, the logic of the [dev].
At least people should know that they should look at  setup.cfg

azure-pipelines.yml
```diff
@@ -28,12 +28,11 @@ jobs:
 
   - script: |
       python -m pip install --upgrade pip
-      pip install --prefer-binary -r requirements-dev.txt
       pip install pytest-azurepipelines codecov
     displayName: 'Install dependencies'
 
   - script: |
-      pip install .
+      pip install --prefer-binary .[dev]
```
bthirion, line 35: Same here

build_tools/circle/dependencies.sh
```diff
@@ -8,4 +8,4 @@ conda install -n testenv -yq nibabel python-kaleido sphinx-gallery sphinxcontrib
 conda install -c plotly plotly
 source activate testenv
 python -m pip install --user --upgrade --progress-bar off pip setuptools
-python -m pip install .
+python -m pip install .[test]
```
bthirion, line 11: and maybe here

CONTRIBUTING.rst
tsalo: I will add comments to `setup.cfg` and also point to that file in each of the places you commented on here.

azure-pipelines.yml
tsalo: I'm not actually sure what `--prefer-binary` is for. The pip documentation says it's to "[p]refer older binary packages over newer source packages." Does that mean the goal here is to install the oldest versions of dependencies?

setup.cfg
NicolasGensollen: I think we should avoid bumping our dependencies in minor releases, but maybe we can make an exception here... (pinging @jeromedockes to see what he thinks about this).

The current CI is configured in such a way that tests are run on either:

- python 3.6 + packages pinned at minimum dependencies
- python [3.7, 3.8, 3.9] + latest versions of dependencies

which explains why we never saw the issue with python 3.6+ and sklearn==0.21.
_Note: I also realize we should add Python 3.6 to tests with latest versions._

bthirion (approved):
LGTM

setup.cfg
tsalo: That's fair. Tests are currently passing, so what if I just add a comment to the description of the `min` dependency group that it should _only_ work for Python 3.6, and we can push this can down the road?

We may need a new issue for testing Python 3.6 + latest versions.

setup.cfg
tsalo: @NicolasGensollen @bthirion do you have opinions on this proposal? I could be missing the reasons behind the current division of dependencies.

tsalo:
It turns out I completely missed `nilearn/version.py`, so I'm switching this back to a draft.

tsalo:
I think I've dealt with the issues in `version.py`, so I'm opening this back up for review.

tsalo:
Does anyone know why `matplotlib` is treated as an optional dependency?

bthirion:
because we assumed that Nilearn might be run on servers without matplotlib installed. Note sure whether this is still a valid reason.

jeromedockes:
It's only needed for plotting, which is quite independent from all the analysis functionality so it makes sense to only impose it to users who actually need the plotting

tsalo:
>because we assumed that Nilearn might be run on servers without matplotlib installed. Note sure whether this is still a valid reason.

Thanks @bthirion. Working around matplotlib has been difficult, so if we can assume that matplotlib is installable for all users that would be amazing.

>It's only needed for plotting, which is quite independent from all the analysis functionality so it makes sense to only impose it to users who actually need the plotting

@jeromedockes Even with all of the import checks, I'm not sure that matplotlib is as optional as that. For example, in NiMARE I don't use any nilearn plotting functions, but I use `get_clusters_table`, and I had to include matplotlib as a dependency to support that function. I can see why we might need workarounds if matplotlib _can't_ be installed on a machine (if I'm understanding @bthirion's point correctly), but otherwise it seems like a lot of work to avoid installing a very common library.

tsalo:
On the other hand, that might be a conversation for another PR. I can add a new dependency group called `plotting` with matplotlib, kaleido, and plotly for this PR. I just want to make sure the installation instructions are clear about that option, since I think a large portion of users will want to use it.

setup.cfg
```diff
@@ -3,6 +3,60 @@
 [bdist_rpm]
 doc-files = doc
 
+[options]
+python_requires = >=3.6
+install_requires =
+	argparse
```
tsalo, line 9: argparse is only used once, in [`nilearn/plotting/rm_file.py`](https://github.com/nilearn/nilearn/blob/6aaf8f94433dfaf2967d69c64c0f688d7e1c812d/nilearn/plotting/rm_file.py). Could we use some other method for removing files and drop this dependency?

jeromedockes: not sure how it ended up in the setup but argparse is part of the python standard library

jeromedockes:
> For example, in NiMARE I don't use any nilearn plotting functions, but I use get_clusters_table, and I had to include matplotlib as a dependency to support that function

yes but that is part of reporting which depends on plotting. still the numerical analysis / image manipulation / machine learning parts should work independently of plotting/reporting

jeromedockes:
Not sure I understand why it is a lot of work to have matplotlib as an optional dependency? It is already the case

jeromedockes:
> I can add a new dependency group called plotting with matplotlib, kaleido, and plotly for this PR

jeromedockes:
> I can add a new dependency group called plotting with matplotlib, kaleido, and plotly for this PR

that's a good idea, although I would separate matplotlib from the rest -- most users will install matplotlib, as you point out, but only those who need surface plots will install plotly and kaleido ATM

tsalo:
What would be a good name for the plotly/kaleido group?

jeromedockes:
> What would be a good name for the plotly/kaleido group?

maybe "plotly" unless we can think of something better?

btw I just checked scikit-learn also has matplotlib as an optional dependency; even though plotting is a more significant part of nilearn than of scikit-learn I think it makes sense to keep that separation

bthirion (commented):
Thx ! This deserves a whatsnew entry.

tsalo:
@bthirion I added something in c2e483f76c59919a2f99dcebc3b2505bfeb55ab1. Please let me know if you'd prefer something different.

bthirion (approved):
LGTM, thx !

CONTRIBUTING.rst
```diff
@@ -315,13 +315,13 @@ Here are the key steps you need to go through to copy the repo before contributi
 
 3. install the forked version of `nilearn`::
 
-      pip install . -e
+      pip install -e .[dev]
```
jeromedockes, line 318:
```suggestion
      pip install -e '.[dev]'
```
(prevent shell from interpreting the square brackets as a glob pattern)

CONTRIBUTING.rst
```diff
@@ -364,7 +364,7 @@ If you wish to build documentation:
 1. First, ensure that you have installed sphinx and sphinx-gallery. When in your
    fork top folder, you can install the required packages using::
 
-      pip install -r requirements-build-docs.txt
+      pip install .[doc]
```
jeromedockes, line 367:
```suggestion
      pip install '.[doc]'
```

README.rst
```diff
@@ -53,6 +53,8 @@ The required dependencies to use the software are:
 * Nibabel >= 3.0
 * Pandas >= 1.0
 
+For a comprehensive list of dependencies, see the file ``nilearn/setup.cfg``.
```
jeromedockes, line 56: I wonder if  we should remove the min versions from the readme to avoid having to update them in 2 places? if we put a link to the setup.cfg that may be enough

build_tools/travis/install.sh
```diff
@@ -20,9 +20,9 @@ conda update --yes conda
 
 conda init bash
 conda create -n testenv -yq
-conda install -n testenv -yq python=3.8 numpy scipy scikit-learn matplotlib pandas pytest pytest-xdist joblib nibabel cython requests
```
jeromedockes, line 23: I think we don't use travis anymore? there may be some cleanup that we can do in another PR

setup.py
```diff
@@ -24,25 +21,6 @@ def load_version():
     return globals_dict
 
 
-def is_installing():
```
jeromedockes, line 27: that's for another PR but as further clean up we could see if we could put the version in a text file in the package data and avoid having to execute version.py in the setup which is very messy

jeromedockes (approved):
thanks!

README.rst
tsalo: That sounds reasonable. I'll do that.

CONTRIBUTING.rst
tsalo: @jeromedockes based on reports from `tedana` Windows users, using single quotes there seems to cause problems. For context, the relevant issue is https://github.com/ME-ICA/tedana/issues/844.

Is it possible that these instructions will cause problems for Windows users?

jeromedockes: how about `pip install -e ".[dev]"`?

setup.cfg
```diff
@@ -3,6 +3,69 @@
 [bdist_rpm]
 doc-files = doc
 
+[options]
+python_requires = >=3.6
+install_requires =
+	joblib>=0.12
```
NicolasGensollen, line 9: The minimum versions were updated when releasing 0.9 (see readme file). I don't think these updates were taken into account here.

setup.cfg
```diff
@@ -3,6 +3,69 @@
...
+	numpy>=1.16
+	pandas>=0.24
+	requests>=2
+	scikit-learn>=0.22
+	scipy>=1.2
+
+[options.extras_require]
+# Necessary reqs to use nilearn's plotting module
+plotting =
+	matplotlib>=2.0
```
NicolasGensollen, line 21: same for matplotlib (should be >= 3.0)

setup.cfg
```diff
@@ -3,6 +3,69 @@
...
+	sphinx
+	sphinx-copybutton
+	sphinx-gallery
+	sphinxcontrib-bibtex
+	sphinxext-opengraph
+
+# For testing of oldest usable versions of dependencies.
+min =
+	joblib==0.15
+	nibabel==3.0.0
```
NicolasGensollen, line 48: looks like these are fine though.

NicolasGensollen (commented):
Thanks @tsalo ! Looks good! 👍 
I just have a small question on the min versions (see below).

CONTRIBUTING.rst
tsalo: The user only tried single quotes, so I don't know. Do any of the devs use Windows?

setup.cfg
tsalo: Thanks for catching these!

tsalo:
I am only now realizing that there are two approvals. @NicolasGensollen would you like to take another look or should I merge?

NicolasGensollen (approved):
LGTM, thanks @tsalo !

bthirion:
Merging, then. Thanks !

PR #2958, opened 2021-09-12
[FIX] locals() → globals()

### LINKED ISSUE #2957: Modification of dictionary returned by locals()
This code from https://github.com/nilearn/nilearn/blob/ac1a934/nilearn/plotting/cm.py#L221-L222:
```python
# Save colormaps in the scope of the module
locals().update(_cmap_d)
```
raises an [LGTM alert](https://lgtm.com/projects/g/nilearn/nilearn/snapshot/d520eb8468e57b591fe97798f038a56ab134679b/files/nilearn/plotting/cm.py#x63244eecb7d9ea77:1):
> Modification of dictionary returned by locals()
> 
> Modification of the locals() dictionary will have no effect on the local variables.

And indeed, after reading:
* [Update locals from inside a function](https://stackoverflow.com/questions/30295163/update-locals-from-inside-a-function#67367191)
* [Built-in Functions](https://docs.python.org/3/library/functions.html#locals)
  >Update and return a dictionary representing the current local symbol table. Free variables are returned by [`locals()`](https://docs.python.org/3/library/functions.html#locals) when it is called in function blocks, but not in class blocks. Note that at the module level, [`locals()`](https://docs.python.org/3/library/functions.html#locals) and [`globals()`](https://docs.python.org/3/library/functions.html#globals) are the same dictionary.

I suggest changing `locals()` → `globals()` to make the intent clear, since this happens at module level.

### Expected behavior

No LGTM warning.

### Actual behavior

Raises an LGTM warning:
https://lgtm.com/projects/g/nilearn/nilearn/snapshot/d520eb8468e57b591fe97798f038a56ab134679b/files/nilearn/plotting/cm.py#x63244eecb7d9ea77:1

DimitriPapadopoulos:
In general, you cannot modify the local symbol table returned by `locals()`.
However, in the scope of the module, `locals()` is identical to `globals()`.

To make the intent clear, use `globals()` instead of `locals()`.

Closes #2957.

### DIFF
--- a/nilearn/plotting/cm.py
+++ b/nilearn/plotting/cm.py
@@ -219,7 +219,7 @@ _cmap_d['videen_style'] = _colors.LinearSegmentedColormap.from_list(
     'videen_style', videen_style)
 
 # Save colormaps in the scope of the module
-locals().update(_cmap_d)
+globals().update(_cmap_d)
 # Register cmaps in matplotlib too
 for k, v in _cmap_d.items():
     try:  # "bwr" is in latest matplotlib


bthirion (approved):
Sounds right.

NicolasGensollen (approved):
LGTM, thanks!

NicolasGensollen:
Thanks, merging...

PR #2963, opened 2021-09-13
[FIX] Fix `get_data` in `get_clusters_table`

### LINKED ISSUE #2955: AxisError when computing `reporting.get_cluster_table`
This issue concerns what seems to me like a bug in the `reporting.get_cluster_table` method. I have been encoutering a `AxisError` raised by numpy when trying to compute the table with certain `stat_threshold` values.

With a high threshold value (e.g `3.`) the computation works fine, but when using smaller threshold values (1e-2), the error is thrown.

The error happens because the `ijk` value passed to a numpy `apply_along_axis` [here](https://github.com/nilearn/nilearn/blob/ac1a934e3b2b4061f894b518c960412df9ea4f11/nilearn/reporting/_get_clusters_table.py#L83) is an empty list. I don't know why yet. You can reproduce the error with the snippet below.

I guess it makes less sense to extract peaks with a lower threshold, but it still worth investigating why the error occurs.

### Expected behavior

`get_clusters_table` return the clusters table whatever the `stat_threshold` value

### Actual behavior

`get_clusters_table` raises an error when `stat_threshold` values are small.

```python
import tempfile

from nilearn import reporting
from nilearn.datasets import fetch_neurovault_ids

DATA_PATH = tempfile.TemporaryDirectory()

collection_id = 307
neurovault_element = fetch_neurovault_ids(
    collection_ids=[collection_id],
    mode="download_new",
    data_dir=str(DATA_PATH),
)
sample_brain_map = neurovault_element.images[0]

# Working fine with a stat threshold of 3.
neurovault_peaks = reporting.get_clusters_table(
    sample_brain_map,
    stat_threshold=3.,
    min_distance=12.,
)

# Raising error with a stat threshold of 1e-2
neurovault_peaks = reporting.get_clusters_table(
    sample_brain_map,
    stat_threshold=1e-2,
    min_distance=12.,
)
```

NicolasGensollen:
Closes #2955 

Thanks for reporting the bug @RaphaelMeudec ! :+1:

### DIFF
--- a/nilearn/reporting/tests/test_reporting.py
+++ b/nilearn/reporting/tests/test_reporting.py
@@ -117,6 +117,12 @@ def test_get_clusters_table(tmp_path):
     )
     assert len(cluster_table) == 2
 
+    # Test that nans are handled correctly (No numpy axis errors are raised)
+    data[data == 0] = np.nan
+    stat_img_nans = nib.Nifti1Image(data, affine=np.eye(4))
+    cluster_table = get_clusters_table(stat_img_nans, 1e-2, 0, two_sided=False)
+    assert len(cluster_table) == 1
+
 
 def test_get_clusters_table_not_modifying_stat_image():
     shape = (9, 10, 11)
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -12,7 +12,7 @@ import pandas as pd
 import nibabel as nib
 from scipy import ndimage
 
-from nilearn.image import get_data
+from nilearn._utils.niimg import _safe_get_data
 from nilearn.image.resampling import coord_transform
 from nilearn._utils import check_niimg_3d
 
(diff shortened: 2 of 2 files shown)


nilearn/reporting/tests/test_reporting.py
```diff
@@ -117,6 +117,12 @@ def test_get_clusters_table(tmp_path):
     )
     assert len(cluster_table) == 2
 
+    # Test that nans are handled correctly (No numpy axis errors are raised)
+    data[data == 0] = np.nan
+    stat_img_nans = nib.Nifti1Image(data, affine=np.eye(4))
+    cluster_table = get_clusters_table(stat_img_nans, 1e-2, 0, two_sided=False)
+    assert len(cluster_table) == 1
```
bthirion, line 124: This test fails with the current code ?

bthirion (commented):
Thx, LGTM.

nilearn/reporting/tests/test_reporting.py
NicolasGensollen: Yes. See [here](https://github.com/nilearn/nilearn/runs/3587185350)

bthirion (approved):
Great. Just need to wait for green cirecleci then.

tsalo (approved):
The changes look good to me too.

NicolasGensollen:
Thanks for the reviews! 
Merging...

PR #2965, opened 2021-09-13
[ENH] Add `cluster_threshold` and `two_sided` options to `image.threshold_img`

### LINKED ISSUE #2956: Add `cluster_threshold` option to `threshold_img`
I think there's a strong use-case for the ability to apply an arbitrary cluster-size threshold in `nilearn.image.threshold_img`.

This stems from https://neurostars.org/t/plot-stat-map-plot-only-clusters/20134, which I think would be solved by adding a `cluster_threshold` parameter to `threshold_img`.

A `two_sided` parameter would also probably be useful, but I can open a separate issue for that if necessary.

### Benefits to the change
The main benefit would be the ability to apply cluster-extent thresholding outside of a full statistical analysis (as in `threshold_stats_img`).

```python
from nilearn import image

thresh_img = image.threshold_img(img, threshold=3.09, cluster_threshold=10)
```

tsalo:
Closes #2956. Currently in draft form until I have some tests, but I'd love input on how I used `threshold_img` in other functions.

- Add `cluster_threshold` and `two_sided` options to `nilearn.image.threshold_img`.
- Use `threshold_img` in `nilearn.glm.thresholding.threshold_stats_img`. Now all that function has to do on its own is calculate the appropriate statistical threshold based on the correction method requested.
- Use `threshold_img` in `nilearn.reporting.get_clusters_table`. Some of the work in that function can be done with cluster-extent thresholding in `threshold_img`.

To do:
- [x] Tests!
- [x] Update changelog.
- [x] Incorporate new parameters into an example.


nilearn/reporting/_get_clusters_table.py
```diff
@@ -201,26 +206,6 @@ def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
...
-        # this checks for stats higher than threshold after small clusters
-        # were removed from temp_stat_map
-        if np.sum(temp_stat_map > stat_threshold) == 0:
-            warnings.warn(
-                'Attention: No clusters with more than {0} voxels'.format(
-                    cluster_threshold,
-                )
-            )
-            continue
-
```
tsalo, line 223: This initial threshold + label step is just to perform cluster-extent thresholding, which `threshold_img` can now do. The second step, in which the clusters are broken up and investigated individually, is still necessary.

bthirion:
I realize that we are duplicating functionalities from
https://github.com/nilearn/nilearn/blob/main/nilearn/glm/thresholding.py
Deeper refactoring is needed to avoid these duplications.

tsalo:
@bthirion I think I got rid of the duplication in `threshold_stats_img`. Is there anywhere else I should look to clean up?

bthirion:
Sorry, I missed the point. I'll look more in detail tomorrow.

tsalo:
The only thing I'm currently stuck on is which example to use the new parameters in. `threshold_img` is currently only used in two examples:

- [Encoding models for visual stimuli from Miyawaki et al. 2008](https://nilearn.github.io/auto_examples/02_decoding/plot_miyawaki_encoding.html#sphx-glr-auto-examples-02-decoding-plot-miyawaki-encoding-py): A thresholded map is plotted in this example, but it's unclear to me if including a minimum cluster size is appropriate here.
- [Region Extraction using a t-statistical map (3D)](https://nilearn.github.io/auto_examples/06_manipulating_images/plot_extract_rois_statistical_maps.html#sphx-glr-auto-examples-06-manipulating-images-plot-extract-rois-statistical-maps-py): `cluster_threshold` would be useful for showing thresholded maps at the beginning of the example that better match the connected regions identified later on, but it could take away from `connected_regions`, which is the focus of the example.

bthirion:
Maybe add a call in another example, such as 
https://github.com/nilearn/nilearn/blob/main/examples/05_glm_second_level/plot_thresholding.py

nilearn/image/image.py
```diff
@@ -798,6 +808,16 @@ def threshold_img(img, threshold, mask_img=None, copy=True):
         voxels which have intensities greater than this score will be kept.
         The given string should be within the range of "0%" to "100%".
 
+    cluster_threshold : :obj:`float`, optional
+        Cluster size threshold, in voxels. In the returned thresholded map,
+        sets of connected voxels (``clusters``) with size smaller
+        than this number will be removed. Default=0.
+
```
NicolasGensollen, line 815: Can we add a versionadded directive for these two new arguments ?

nilearn/image/image.py
```diff
@@ -842,7 +862,45 @@ def threshold_img(img, threshold, mask_img=None, copy=True):
                                            percentile_func=scoreatpercentile,
                                            name='threshold')
 
-    img_data[np.abs(img_data) < cutoff_threshold] = 0.
+    # Apply threshold
```
NicolasGensollen, line 865: I think the function is a bit long and could be simplified:
First, I think we don't need to check if the threshold is `None` to raise a `ValueError` (see above), as `check_threshold` already does (almost) this (raises a `TypeError`).

nilearn/image/image.py
```diff
@@ -842,7 +862,45 @@ def threshold_img(img, threshold, mask_img=None, copy=True):
...
+    # Perform cluster thresholding, if requested
+    if cluster_threshold > 0:
+        # Define array for 6-connectivity, aka NN1 or "faces"
+        conn_mat = np.zeros((3, 3, 3), int)
+        conn_mat[1, 1, :] = 1
+        conn_mat[1, :, 1] = 1
+        conn_mat[:, 1, 1] = 1
+
+        for i_vol in range(img_data.shape[3]):
+            vol_data = img_data[..., i_vol]
```
NicolasGensollen, line 885: Do you think we could put the code of the inner loop together with the connectivity matrix definition in a helper function?

It seems to me that this could then be parallelized but I might be missing something (or maybe it's fast enough we don't need to bother about this...). :thinking:

NicolasGensollen (commented):
This is looking good, thanks @tsalo ! :+1:

nilearn/image/image.py
```diff
@@ -842,7 +862,45 @@ def threshold_img(img, threshold, mask_img=None, copy=True):
...
+                        vol_data[label_map == c_val] = 0
+
+            img_data[..., i_vol] = vol_data
+
+    if expand_to_4d:
+        # Reduce back to 3D
+        img_data = img_data[:, :, :, 0]
+
+    # Reconstitute img object
     threshold_img = new_img_like(img, img_data, affine)
```
tsalo, line 904: Should copy_header be set to the default (False) here? It seems like it should retain the header info, but I'm not sure if there's an issue with `resample_img` above.

nilearn/image/image.py
tsalo: I moved the inner loop into a new private function, but I didn't try parallelizing or anything.

nilearn/image/image.py
tsalo: Done!

nilearn/image/image.py
tsalo: Makes sense. Fixed!

bthirion (commented):
Looks great. Just need to add whatsnew entry if I'm not mistaken.

bthirion (approved):
Looks great, thx !

NicolasGensollen (approved):
LGTM, thanks @tsalo ! :+1:

NicolasGensollen:
Thanks!
Merging!

PR #2980, opened 2021-09-19
[FIX] LGTM.com recommendations

DimitriPapadopoulos:
Fix some LGTM.com recommendations:
https://lgtm.com/projects/g/nilearn/nilearn/?severity=recommendation

### DIFF
--- a/examples/01_plotting/plot_surface_projection_strategies.py
+++ b/examples/01_plotting/plot_surface_projection_strategies.py
@@ -25,7 +25,6 @@ import numpy as np
 
 import matplotlib
 from matplotlib import pyplot as plt
-from mpl_toolkits.mplot3d import Axes3D
 
 from nilearn.surface import surface
 from nilearn.plotting import show
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -11,7 +11,6 @@ import shutil
 from distutils.version import LooseVersion
 
 import nibabel
-import sklearn
 
 from joblib import Memory
 
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -235,7 +235,7 @@ def _repr_niimgs(niimgs, shorten=True):
                    (niimgs.__class__.__name__,
                     repr(niimgs.shape),
                     repr(niimgs.affine))
-    except:
+    except Exception:
         pass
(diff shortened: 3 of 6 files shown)


bthirion (commented):
I'm not familiar with the `except Exception`
The result does not sound wrong...

DimitriPapadopoulos:
[`Exception`](https://docs.python.org/3/library/exceptions.html#Exception) is the base class of the exceptions we want to catch:
> All built-in, non-system-exiting exceptions are derived from this class. All user-defined exceptions should also be derived from this class.

See rationale in [Except block handles '`BaseException`'](https://lgtm.com/rules/6780080/):
> All exception classes in Python derive from `BaseException`. `BaseException` has three important subclasses, `Exception` from which all errors and normal exceptions derive, `KeyboardInterrupt` which is raised when the user interrupts the program from the keyboard and `SystemExit` which is raised by the `sys.exit()` function to terminate the program.
> 
> Since `KeyboardInterrupt` and `SystemExit` are special they should not be grouped together with other `Exception` classes.
> 
> Catching `BaseException`, rather than its subclasses may prevent proper handling of `KeyboardInterrupt` or `SystemExit`. It is easy to catch `BaseException` accidentally as it is caught implicitly by an empty `except:` statement.

See also [PEP 0352](https://www.python.org/dev/peps/pep-0352/#exception-hierarchy-changes):
>With the exception hierarchy now even more important since it has a basic root, a change to the existing hierarchy is called for. As it stands now, if one wants to catch all exceptions that signal an error and do not mean the interpreter should be allowed to exit, you must specify all but two exceptions specifically in an `except` clause or catch the two exceptions separately and then re-raise them and have all other exceptions fall through to a bare `except` clause:
> ```python
> except (KeyboardInterrupt, SystemExit):
>     raise
> except:
>     ...
> ```
> That is needlessly explicit. This PEP proposes moving KeyboardInterrupt and SystemExit to inherit directly from BaseException.
> ```
> - BaseException
>   |- KeyboardInterrupt
>   |- SystemExit
>   |- Exception
>      |- (all other current built-in exceptions)
> ```

bthirion (approved):
Thx for the explanations.

NicolasGensollen (approved):
LGTM, thanks!

NicolasGensollen:
Merging, thanks!

PR #2984, opened 2021-09-23
"[INFRA][MRG] add permissions to auto-comment workflow"

thomasbazeille:
Closes # .

Changes proposed in this pull request:

-
-

PR #2982, opened 2021-09-23
[IGNORE] Testing auto-comment failure

thomasbazeille:
Closes # .

Changes proposed in this pull request:

-
-

PR #2970, opened 2021-09-15
Automatic commenting on new PR

thomasbazeille:
Closes #2962 .

Changes proposed in this pull request:

- First draft of auto-commenting to support discussion on what its content should be.

Main hurdles : If we want to include a checklist as @tsalo proposed, it is hard to know which elements should absolutely be included.

PR #2986, opened 2021-09-24
[INFRA] Fix auto-comment action

NicolasGensollen:
Try a simple fix for the auto-comment action introduced in #2970 (See also #2982 and #2984).
@thomasbazeille : Looking at this [issue](https://github.com/wow-actions/auto-comment/issues/14), it seems like it is the recommended way to fix the issue. If it doesn't work, we can try #2984


NicolasGensollen:
I'll merge this and try opening a new PR to see if it solves the issue

PR #2990, opened 2021-09-26
[FIX] LGTM.com warning: Conflicting attributes in base classes

DimitriPapadopoulos:
Base classes have conflicting values for attribute '`_more_tags`': Function `_more_tags` and Function `_more_tags`.

Both `BaseEstimator` and `ClassifierMixin` define `_more_tags()`. Override the function in the subclass `ReNA` to explicitly implement the previous behaviour, implicitly chosen by the Python [MRO](https://www.python.org/download/releases/2.3/mro/).

**Please review carefully**, as I'm not certain whether the initial implicit behaviour was the expected behaviour.

### DIFF
--- a/nilearn/regions/rena_clustering.py
+++ b/nilearn/regions/rena_clustering.py
@@ -9,8 +9,7 @@ import warnings
 from scipy.sparse import csgraph, coo_matrix, dia_matrix
 from joblib import Memory
 
-from sklearn.base import TransformerMixin, ClusterMixin
-from sklearn.base import BaseEstimator
+from sklearn.base import BaseEstimator, TransformerMixin, ClusterMixin
 from sklearn.utils.validation import check_is_fitted
 from sklearn.utils import check_array
 from nilearn.masking import _unmask_from_to_3d_array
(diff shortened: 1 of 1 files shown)


bthirion (approved):
LGT, thx.

NicolasGensollen (approved):
LGTM, thanks @DimitriPapadopoulos !
I'm just wondering why you get a warning with the current implementation. The correct method is selected thanks to the MRO and the whole point of class hierarchies is to avoid redefining things.

DimitriPapadopoulos:
It's not an actual bug. Indeed the correct method is selected. The warning is because the method selection is implicit instead of explicit as far as I can see:
[Conflicting attributes in base classes](https://lgtm.com/rules/7860084/)
LGTM.com estimate this is an arcane part of Python that cannot be left to implicit:
> Even if it is the desired behavior it makes the code hard to understand and maintain.

The real fix might be to introduce `super()` in parent classes, but I would need more time for that as I haven't been using complex hierarchies for some years now and have forgotten the details of `super()` usage.

NicolasGensollen:
Thanks, merging...

PR #2991, opened 2021-09-27
[MAINT] Add listing utilities

NicolasGensollen:
This PR proposes to add helper functions to list functions and classes in Nilearn.

**Example**

Listing functions:

```python
>>> from nilearn._utils import all_functions
>>> [_[0] for _ in all_functions(modules_to_consider=["input_data"])]
['filter_and_extract', 'check_embedded_nifti_masker', 'filter_and_mask']
>>> [_[0] for _ in all_functions(modules_to_consider=["input_data"], return_private=True)]
['filter_and_extract', 'check_embedded_nifti_masker', '_get_mask_strategy', '_get_mask_strategy', 'filter_and_mask', '_apply_mask_and_get_affinity', '_iter_signals_from_spheres']
```

Listing classes:

```python
>>> from nilearn._utils import all_classes
>>> [_[0] for _ in all_classes(modules_to_consider=["input_data"])]
['BaseMasker', 'MultiNiftiMasker', 'NiftiLabelsMasker', 'NiftiMapsMasker', 'NiftiMasker', 'NiftiSpheresMasker']
```


bthirion (commented):
Thx. Additionally I would simply showcase this feature in the narrative documentation.
+ whatsnew entry.

NicolasGensollen:
>If this is intended for users why not put it in a public module? client code shouldn't import the "private" _utils module

I was initially not planning to expose them to users as I wrote them with a longer maintenance goal in mind, but I don't have a strong opinion on this. If people think such functions could be useful, I'm OK to put them in a more appropriate location, otherwise you're right that we shouldn't showcase them in the docs.

nilearn/_utils/tests/test_common.py
```diff
@@ -0,0 +1,51 @@
...
+def test_private_vs_public(func):
+    public_only = set(func(return_private=False))
+    private_and_public = set(func(return_private=True))
+    assert public_only.issubset(private_and_public)
+    assert np.all(
+        [elt[0].startswith("_") for elt in private_and_public - public_only]
+    )
+
+
+def test_all_functions():
```
Remi-Gau, line 30: that test would fail every time a public function is added to the API

Remi-Gau:
one possible use case of this: have a test in place that checks the number of public classes and functions we have and make sure that this number does not change by mistake when someone adds new functions or classes.

Remi-Gau:
@ymzayek @bthirion 
could we merge this one in?
I have to code a couple of times some ways to get all the functions in nilearn for maintenance purpose and having this utility would help.

ymzayek (approved):
Agreed. It LGTM!

nilearn/_utils/tests/test_common.py
bthirion: Yes, put this explicitly in the test docstring.

bthirion (commented):
LGTM overall, thx !

nilearn/_utils/tests/test_common.py
Remi-Gau: actually: do you think I should just modify this test to implement what I mention in the comment [below](https://github.com/nilearn/nilearn/pull/2991#issuecomment-1673725380)

bthirion:
> @ymzayek @bthirion could we merge this one in? I have to code a couple of times some ways to get all the functions in nilearn for maintenance purpose and having this utility would help.

I'm basically OK for merging. Can you just update the docstring above to explain the logic ?

bthirion (approved):

nilearn/_utils/tests/test_common.py
ymzayek: @Remi-Gau I think it's a good idea. But feel free to do it in another PR

nilearn/_utils/tests/test_common.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def test_number_public_classes():
+    """Check that number of public classes is stable.
+
+    If it changes, it means that we have added or removed a public function.
+    If this is intentional, then the number should be updated in the test.
+    Otherwise it means that the public API of nilearn has changed by mistake.
+    """
+    assert len({_[0] for _ in all_classes()}) == 61
```
Remi-Gau, line 47: updated the doc strings: those are basically "API change detectors". If they become too annoying to maintain I think we should feel comfortable removing them.

Remi-Gau:
OK the failure is our windows usual flaky suspect from #3925 

merging

PR #2992, opened 2021-09-27
[FIX] LGTM.com warning: Variable defined multiple times

DimitriPapadopoulos:
This assignment to 'display' is unnecessary as it is redefined here before this value is used.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,4 +8,4 @@ omit =
     */nilearn/conftest.py
     */nilearn/version.py
     */nilearn/datasets/data/convert_templates.py
-    */nilearn/plotting/glass_brain_files/svg_to_json_converter.py
+    */nilearn/plotting/glass_brain_files/*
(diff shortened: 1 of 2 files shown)


bthirion (commented):
LGTM, though I'm wondering why we're keeping this __main__ function. Should be replaced by a test, no ?

DimitriPapadopoulos:
I think all the files under `nilearn/plotting/glass_brain_files/` are actual scripts, not modules. Especially `plot_align_svg.py` starts with `#!/usr/bin/env python` and then:
> The goal of this script is to align the glass brain SVGs on top of the anatomy. This is only useful for internal purposes especially when the SVG is modified.

DimitriPapadopoulos:
Just modified **`.coveragerc`** so that it skips the whole `nilearn/plotting/glass_brain_files/` directory, not only `nilearn/plotting/glass_brain_files/svg_to_json_converter.py`.

NicolasGensollen (approved):
LGTM, thanks!

bthirion (approved):
LGTM, thx.

NicolasGensollen:
Merging, thanks!

PR #2994, opened 2021-09-28
[MAINT] LGTM.com recommendation: Unused local variable

DimitriPapadopoulos:
The value assigned to local variable '`Xmean`' is never used.
The value assigned to local variable '`data`' is never used.
The value assigned to local variable '`vertices`' is never used.
The value assigned to local variable '`faces`' is never used.
The value assigned to local variable '`stack_level`' is never used.

### DIFF
--- a/nilearn/decoding/space_net.py
+++ b/nilearn/decoding/space_net.py
@@ -469,8 +469,7 @@ def path_scores(solver, X, y, mask, alphas, l1_ratios, train, test,
         best_w = w_
 
     if len(best_w) == n_features:
-        if Xmean is None:
-            Xmean = np.zeros(n_features)
+        # TODO: do something with Xmean
         best_w = np.append(best_w, 0.)
 
     all_test_scores = np.array(all_test_scores)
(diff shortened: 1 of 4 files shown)


nilearn/decoding/space_net.py
```diff
@@ -468,9 +468,10 @@ def path_scores(solver, X, y, mask, alphas, l1_ratios, train, test,
             w_[support] = best_w
         best_w = w_
 
+    # TODO: do something with Xmean
     if len(best_w) == n_features:
-        if Xmean is None:
-            Xmean = np.zeros(n_features)
+        # if Xmean is None:
```
NicolasGensollen, line 473: I'd be in favor of totally removing Xmean from the function `path_scores` since it is unused and undocumented

NicolasGensollen (commented):
One small comment but LGTM otherwise. Thanks!

nilearn/decoding/space_net.py
DimitriPapadopoulos: Agreed. i've left `# TODO: do something with Xmean`, hope that's OK with you.

bthirion (approved):
LGTM, thx.

NicolasGensollen (approved):
LGTM, thanks!

NicolasGensollen:
Merging...

PR #2988, opened 2021-09-24
[DOC] Add `dictionary learning` to glossary

NicolasGensollen:
This PR proposes to add `dictionary learning` to the glossary.

PR #2995, opened 2021-09-28
[FIX] Fix broken ref

NicolasGensollen:
Fixes broken reference that I introduced in #2988


NicolasGensollen:
Merging fast to fix master...

PR #2998, opened 2021-09-29
[DOC] Add `fwhm` to glossary and standard doc

### LINKED ISSUE #2997: Incorrect unit for `smoothing_fwhm` in `apply_mask`
The docstring of `apply_mask` says that `smoothing_fwhm` is in voxels:

https://github.com/nilearn/nilearn/blob/c1e12c6244182ad1e823a6e97b4dc02b2f82f64d/nilearn/masking.py#L829-L831

but the value is passed as is to `_smooth_array` 

https://github.com/nilearn/nilearn/blob/c1e12c6244182ad1e823a6e97b4dc02b2f82f64d/nilearn/masking.py#L898-L901

which docstring says that `smoothing_fwhm` is in mm:

https://github.com/nilearn/nilearn/blob/c1e12c6244182ad1e823a6e97b4dc02b2f82f64d/nilearn/image/image.py#L180-L182

NicolasGensollen:
This PR proposes to:

- add `FWHM` to the glossary
- add `fwhm` to the standard doc
- use standard entries for both `smoothing_fwhm` and `fwhm` accross Nilearn
 
Also closes #2997


doc/glossary.rst
```diff
@@ -114,6 +114,12 @@ If you wish to add a missing term, please `create a new issue`_ or
         `Family-wise error rate`_ is the probability of making one or more
         false discoveries, or type I errors when performing multiple
         hypotheses tests.
+    
+    FWHM
+        `FWHM`_ stands for "full width at half maximum". In a distribution, it
+        refers to the difference between the two values of the independent
+        variable at which the dependent variable is equal to half of its
+        maximum value.
```
bthirion, line 122: The above  explanation makes sense for 1D variables, while it is often used for 3D data.
More simply, this is the width of a filter, expressed as the diameter of the area on which the filter value is above half it maximal value.

examples/06_manipulating_images/plot_smooth_mean_image.py
```diff
@@ -2,12 +2,12 @@
...
 
-Here we smooth a mean EPI image and plot the result
+Here we smooth a mean :term:`EPI` image and plot the result
 
-As we vary the smoothing FWHM, note how we decrease the amount of noise,
-but also loose spatial details. In general, the best amount of smoothing
-for a given analysis depends on the spatial extent of the effects that
-are expected.
+As we vary the smoothing :term:`FWHM`, note how we decrease the amount
+of noise, but also loose spatial details. In general, the best amount
```
bthirion, line 8:
```suggestion
of noise, but also lose spatial details. In general, the best amount
```

nilearn/regions/region_extractor.py
```diff
@@ -289,12 +289,11 @@ class RegionExtractor(NiftiMapsMasker):
...
-
     %(extractor)s
-    smoothing_fwhm : :obj:`float`, optional
-        To smooth an image to extract most sparser regions. This parameter
-        is passed to `connected_regions` and exists only for extractor
-        'local_regions'. Please set this parameter according to maps
+    %(smoothing_fwhm)s
+        Use this parameter to smooth an image to extract most sparser regions.
+        This parameter is passed to `connected_regions` and exists only for
+        extractor 'local_regions'. Please set this parameter according to maps
```
bthirion, line 296: I don' understand this sentence

bthirion (commented):
Thx. Minor comments appended.

doc/glossary.rst
NicolasGensollen: Thanks! This is indeed much better and general.

nilearn/regions/region_extractor.py
NicolasGensollen: I tried to clarify it in 4a2c0fa. Basically, `smoothing_fwhm` will only be considered if extraction option is set to `local_regions`.

NicolasGensollen:
Thanks, merging!

PR #3001, opened 2021-09-30
[ENH] Refactor `plot_matrix`

NicolasGensollen:
This PR proposes to:

- refactor the code of the function `plot_matrix` which is too long with a hard-to-follow control flow.
- add more tests for `plot_matrix`
- refactor old test function for `plot_matrix` into multiple smaller tests
- improve the docstrings of the functions in the file `matrix_plotting.py`.
- adjust a little bit the space for the colorbar to avoid having the colorbar too close to the matrix (see a few examples like [this one](https://nilearn.github.io/auto_examples/03_connectivity/plot_signal_extraction.html#sphx-glr-auto-examples-03-connectivity-plot-signal-extraction-py))

### DIFF
--- a/nilearn/plotting/matrix_plotting.py
+++ b/nilearn/plotting/matrix_plotting.py
@@ -107,7 +107,7 @@ def _sanitize_reorder(reorder):
 def _reorder_matrix(mat, labels, reorder):
     """Helper function for plot_matrix.
 
-    This function reorder the provided matrix.
+    This function reorders the provided matrix.
     """
     if not labels:
         raise ValueError("Labels are needed to show the reordering.")
--- a/nilearn/plotting/matrix_plotting.py
+++ b/nilearn/plotting/matrix_plotting.py
@@ -143,23 +143,24 @@ def _mask_matrix(mat, tri):
     return np.ma.masked_array(mat, mask)
 
 
-def _configure_axis(axes, labels):
+def _configure_axis(axes, labels, label_size,
+                    x_label_rotation, y_label_rotation):
     """Helper function for plot_matrix."""
     if not labels:
         axes.xaxis.set_major_formatter(plt.NullFormatter())
         axes.yaxis.set_major_formatter(plt.NullFormatter())
     else:
         axes.set_xticks(np.arange(len(labels)))
-        axes.set_xticklabels(labels, size='x-small')
+        axes.set_xticklabels(labels, size=label_size)
         for label in axes.get_xticklabels():
             label.set_ha('right')
-            label.set_rotation(50)
+            label.set_rotation(x_label_rotation)
         axes.set_yticks(np.arange(len(labels)))
-        axes.set_yticklabels(labels, size='x-small')
+        axes.set_yticklabels(labels, size=label_size)
         for label in axes.get_yticklabels():
             label.set_ha('right')
             label.set_va('top')
-            label.set_rotation(10)
+            label.set_rotation(y_label_rotation)
 
 
 def _configure_grid(axes, grid, tri, size):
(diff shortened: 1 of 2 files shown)


nilearn/plotting/matrix_plotting.py
```diff
@@ -37,6 +40,151 @@ def fit_axes(ax):
...
+    if axes is not None and figure is not None:
+        raise ValueError(
+            "Parameters figure and axes cannot be specified "
+            "together. You gave 'figure=%s, axes=%s'" % (figure, axes)
+        )
+    if figure is not None:
+        if isinstance(figure, plt.Figure):
+            fig = figure
+        else:
+            fig = plt.figure(figsize=figure)
```
bthirion, line 66: A bit weird to code figsize in figure...

nilearn/plotting/matrix_plotting.py
```diff
@@ -37,6 +40,151 @@ def fit_axes(ax):
...
+        axes = plt.gca()
+        own_fig = True
+    else:
+        if axes is None:
+            fig, axes = plt.subplots(1, 1, figsize=(7, 5))
```
bthirion, line 71: Why is (7, 5) a good figsize ?

bthirion (commented):
Lgtm overal, thx.

nilearn/plotting/matrix_plotting.py
NicolasGensollen: I agree, but the docstring of `plot_matrix` says you can do it:

```
figure : :class:`matplotlib.figure.Figure`, figsize :obj:`tuple`,\
        or None, optional
        Sets the figure used. This argument can be either an existing
        figure, or a pair (width, height) that gives the size of a
        newly-created figure.
```

If you think we shouldn't allow this, we'd have to deprecate it.

nilearn/plotting/matrix_plotting.py
NicolasGensollen: I don't know why this was chosen as a default. 
To make it a little less hidden in the code, I can add a global constant DEFAULT_FIGSIZE at the beginning of the file and add a note in the docstring saying that if no figure is provided, a new one will be created with this default size.

nilearn/plotting/matrix_plotting.py
bthirion: This is a bit weird, but let's keep that outside of this PR ---given the existing docstrings.

nilearn/plotting/matrix_plotting.py
```diff
@@ -37,6 +40,151 @@ def fit_axes(ax):
...
+        axes.set_xticklabels(labels, size='x-small')
+        for label in axes.get_xticklabels():
+            label.set_ha('right')
+            label.set_rotation(50)
+        axes.set_yticks(np.arange(len(labels)))
+        axes.set_yticklabels(labels, size='x-small')
+        for label in axes.get_yticklabels():
+            label.set_ha('right')
+            label.set_va('top')
+            label.set_rotation(10)
```
bthirion, line 162: I'm a bit worried about all these hardcoded paarmeters. See also l.156.

nilearn/plotting/matrix_plotting.py
```diff
@@ -98,127 +258,38 @@ def plot_matrix(mat, title=None, labels=None, figure=None, axes=None,
...
                              rect=((0, 0, .95, 1) if colorbar
                                    else (0, 0, 1, 1)))
-
     if colorbar:
         cax, kw = make_axes(axes, location='right', fraction=0.05, shrink=0.8,
                             pad=.0)
         fig.colorbar(mappable=display, cax=cax)
         # make some room
-        fig.subplots_adjust(right=0.8)
+        fig.subplots_adjust(right=0.78)
```
bthirion, line 292: Another hard-coded parameter...

bthirion (approved):
LGTM, thx !

bthirion (commented):
LGTM overall, thx !

NicolasGensollen:
Hi @tsalo 
Do you think you can make a review sometime this week such that we can merge this for 0.9?
Thanks a lot!

PR #3002, opened 2021-09-30
[MAINT] LGTM.com recommendations

DimitriPapadopoulos:
-

### DIFF
--- a/nilearn/_utils/data_gen.py
+++ b/nilearn/_utils/data_gen.py
@@ -12,7 +12,6 @@ from scipy import ndimage
 
 from sklearn.utils import check_random_state
 import scipy.linalg
-import nibabel
 
 from nibabel import Nifti1Image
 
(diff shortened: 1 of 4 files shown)


bthirion (approved):
LGTM, thx.

NicolasGensollen (approved):
LGTM, thanks!

nilearn/plotting/data/html/stat_map_template.html
```diff
@@ -28,7 +28,7 @@
   // On load: build all figures
   $( window ).on('load',function() {
     // Create brain slices
-    var brain = brainsprite(
```
NicolasGensollen, line 31: Actually, it looks like there are a few tests checking for `var brain` in the html code, and which are now failing.

NicolasGensollen (changes requested):

nilearn/plotting/data/html/stat_map_template.html
DimitriPapadopoulos: Ah, I couldn't find them - that's probably because this a template, not an actual HTML document. I'll revert that.

DimitriPapadopoulos: Reverted, although I haven't understood what exactly was using the variable.

NicolasGensollen: Some tests are doing some basic checks on the raw html output and look for the presence of the string "var brain" within the html string.

NicolasGensollen (approved):
LGTM, thanks!

NicolasGensollen:
Merging, thanks!

PR #3015, opened 2021-10-12
[ENH] Refactor tests for image plotting

NicolasGensollen:
This PR proposes to refactor the content of file `nilearn/plotting/tests/test_img_plotting.py` which is about 1600 lines of unsorted test functions with some of them having about one hundred lines. 
If someone wants to add a specific test for a specific plotting function, he/she has to basically scroll over the whole file to find the most adequate spot.
Some tests also have a lot of duplicate code which can be simplified a lot by using pytest fixtures.


NicolasGensollen:
I think this is ready for review.

PR #3018, opened 2021-10-14
Bugfix TypeError for FirstLevelModel with custom response function

### LINKED ISSUE #3017: FirstLevelModel with custom response function fails with TypeError
#2942 added the possibility to use custom response functions when regressing signal from images. However, using a custom response function with `FirstLevelModel` fails.

Nilearn version: 0.8.2.dev

### Expected behavior
The GLM should run without error.

### Actual behavior
It currently fails, returning the following stack-trace:

```
/tmp/ipykernel_58955/2957114893.py in <module>
      3     events_primary_categories_filtered,
      4     subject_label=monkey_subject_name,
----> 5     rf_model=mion_rf,
      6 )

/tmp/ipykernel_58955/4150798340.py in fit_glm(functional_img, events, subject_label, rf_model)
     12     )
     13 
---> 14     fmri_glm.fit(functional_img, events)
     15 
     16     return fmri_glm

/storage/store2/work/athual/repo/nilearn/nilearn/glm/first_level/first_level.py in fit(self, run_imgs, events, confounds, design_matrices, bins)
    549                                                         confounds_matrix,
    550                                                         confounds_names,
--> 551                                                         self.min_onset
    552                                                         )
    553             else:

/storage/store2/work/athual/repo/nilearn/nilearn/glm/first_level/design_matrix.py in make_first_level_design_matrix(frame_times, events, hrf_model, drift_model, high_pass, drift_order, fir_delays, add_regs, add_reg_names, min_onset, oversampling)
    360         matrix, names = _convolve_regressors(
    361             events, hrf_model, frame_times, fir_delays, min_onset,
--> 362             oversampling)
    363 
    364     # step 2: additional regressors

/storage/store2/work/athual/repo/nilearn/nilearn/glm/first_level/design_matrix.py in _convolve_regressors(events, hrf_model, frame_times, fir_delays, min_onset, oversampling)
    237             min_onset=min_onset)
    238 
--> 239         regressor_names += names
    240         if regressor_matrix is None:
    241             regressor_matrix = reg

TypeError: 'NoneType' object is not iterable
```

I won't give a fully operational code snippet, but given that one already has functional images, experiment events and a custom `rf_model, the following code will raise the error, `:

```python
fmri_glm = FirstLevelModel(
    t_r=t_r,
    noise_model="ar1",
    standardize=False,
    hrf_model=rf_model,
)

fmri_glm.fit(functional_img, events)
```

### Fix
According to the stack-trace, in `glm/first_level/design_matrix.py`, instruction `regressor_names += names` fails to execute because `names` is `None`.
This variable is computed a few lines above through `compute_regressor()`, which itself calls `_regressor_names()`, which eventually returns a list of names depending on `hrf_model` given in argument. However, it does not explicitly handle the case where `hrf_model` is a custom function (or where `hrf_model` is not a predefined string for that matter), and hence returns nothing.

I think we could:
1. make sure that `_regressor_names()` return an emtpy list by default
2. in case we don't want 1., maybe we could have `_regressor_names()` return an empty list when `hrf_model` is a custom function (or a list of custom functions)
3. modify `compute_regressor()` such that it checks that `names` is iterable before calling the faulty line
4. generally add more tests using custom response functions

I will open a PR soon to fix these things.

alexisthual:
Closes #3017.

Changes implemented in this pull request:

- fix `_regressor_names()` and making it handle new cases (in which given `hrf_model` is a custom function or a list of custom functions) as well as default cases
- add test for `FirstLevelModel`; this new test passes with the fix, and fails without it as described in #3017

### DIFF
--- a/nilearn/glm/tests/test_first_level.py
+++ b/nilearn/glm/tests/test_first_level.py
@@ -834,6 +834,11 @@ def test_first_level_predictions_r_square():
     lambda tr, ov: np.ones(int(tr * ov))
 ])
 def test_first_level_hrf_model(hrf_model):
+    """
+    Test to ensure that FirstLevelModel runs flawlessly
+    with different values of hrf_model. In particular, one checks that it runs
+    without raising errors when given a custom response function.
+    """
     shapes, rk = [(10, 10, 10, 25)], 3
     mask, fmri_data, _ =\
         generate_fake_fmri_data_and_design(shapes, rk)


alexisthual:
I think it should be good now!

PR #3026, opened 2021-10-19
[MAINT] Deprecate Python 3.6

NicolasGensollen:
Python 3.6 has been released in December 2016 and end of support is scheduled for December 2021.
This PR proposes to deprecate usage of Nilearn with Python 3.6 by scheduling end of support for release 0.10.
Users with a Python 3.6 config will be warned at first import and encouraged to upgrade their Python.


bthirion:
The PR LGTM.
Just wondering: should we wait till e.g. sklearn introduces deprecation warnings ?

PR #3036, opened 2021-10-25
[ENH] `Plotly` and `Matplotlib` surface figures

NicolasGensollen:
Follow-up PR for #2902

This is a draft proposal for the implementation of a common interface for surface figures obtained with ``matplotlib`` and ``plotly``. For the moment, they are very basic and only implement `show` and `savefig` so you should be able to run:

```python
from nilearn import plotting, surface, datasets
 
surf = surface.load_surf_mesh(fsaverage.pial_right)

for engine in ['plotly', 'matplotlib']:
    fig = plotting.plot_surf_stat_map(surf, texture, hemi='right',
                                      view='lateral', colorbar=True,
                                      threshold=1.2, bg_map=fsaverage['sulc_right'],
                                      title='Threshold + bg_map',
                                      engine=engine)
    fig.savefig(f"./plotting_with_{engine}.png")
    fig.show()
```

@jeromedockes this is only a first suggestion, we can iterate or close and start over. ;)

WIP...

See also #3028.


nilearn/plotting/displays.py
```diff
@@ -2357,3 +2357,98 @@ def get_slicer(display_mode):
...
+        Output file path.
+
+    figure : Plotly figure instance or ``None``, optional
+        Figure.
+    """
+    def __init__(self, output_file=None, figure=None):
+        super().__init__(output_file=output_file, figure=figure)
+
+    @classmethod
+    def init_with_figure(cls, figure, output_file=None):
```
jeromedockes, line 2413: as we discussed not sure we need this one as it seems to duplicate `__init__`

nilearn/plotting/displays.py
```diff
@@ -2357,3 +2357,98 @@ def get_slicer(display_mode):
...
+    """Implementation of a surface figure obtained with `plotly` engine.
+
+    Parameters
+    ----------
+    output_file : :obj:`str` or ``None``, optional
+        Output file path.
+
+    figure : Plotly figure instance or ``None``, optional
+        Figure.
+    """
```
jeromedockes, line 2408: maybe an `attributes` section where you point out to users that they can use `.figure` to access the underlying plotly figure for further customization and use of plotly functionality?

nilearn/plotting/displays.py
```diff
@@ -2357,3 +2357,98 @@ def get_slicer(display_mode):
...
+            return self.figure
+
+    def savefig(self, output_file=None):
+        """Saves the figure to file.
+
+        Parameters
+        ----------
+        output_file : :obj:`str` or ``None``, optional
+            Path to output file.
+        """
```
jeromedockes, line 2451: do we need to check for the availability of kaleido here as well?

nilearn/plotting/displays.py
NicolasGensollen: Thanks! Done in d2843cf

nilearn/plotting/displays.py
NicolasGensollen: Yes indeed! Thanks! Also done in d2843cf

jeromedockes (approved):
Thanks @NicolasGensollen I think it's good now!

nilearn/plotting/tests/test_surf_plotting.py
```diff
@@ -70,6 +94,91 @@ def test_set_view_plot_surf_matplotlib(hemi, view, expected_view_matplotlib):
...
+                    reason=("Plotly and/or Ipython is not installed; "
+                            "required for this test."))
+@pytest.mark.parametrize("renderer", ['png', 'jpeg', 'svg'])
+def test_plotly_show(renderer):
+    ps = PlotlySurfaceFigure(go.Figure())
+    assert ps.output_file is None
+    assert ps.figure is not None
+    with mock.patch("IPython.display.display") as mock_display:
+        ps.show(renderer=renderer)
+    assert len(mock_display.call_args.args) == 1
```
bthirion, line 156: these checks are slightly cryptic for non-experts. Briefly explain what is done here.

bthirion (commented):
I think that the behavior should be illustrated in an example that does something with the generated figure.

bthirion:
Thx !

NicolasGensollen:
[Link](https://8804-1235740-gh.circle-artifacts.com/0/dev/auto_examples/01_plotting/plot_3d_map_to_surface_projection.html#sphx-glr-auto-examples-01-plotting-plot-3d-map-to-surface-projection-py) to the example.

bthirion (approved):
The example  is great. LGTM, thx !

NicolasGensollen:
Thanks! Merging...

PR #3044, opened 2021-10-27
use ancpbids to read BIDS datasets

erdalkaraca:
Hi guys,
This is from an experimental state of the ancpbids package which handles reading BIDS datasets.
The commit replaces the built-in functions to read files from a BIDS dataset.

The pull request is to discuss this idea and further refine the implementation of ancpbids such that it could be integrated into nilearn.
ancpbids has very few dependencies, in fact, once cleaning up the code base is finished it will only depend on regex and maybe the numpy packages.

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -22,6 +22,7 @@ from nibabel import Nifti1Image
 from sklearn.base import clone
 from sklearn.cluster import KMeans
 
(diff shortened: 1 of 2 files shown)


NicolasGensollen:
Hi @erdalkaraca 

As we discussed yesterday during the office hours, I will close this PR but will keep the related issue (#3045) open such that we can have the discussions over there.

PR #3047, opened 2021-10-30
[MAINT] LGTM.com error: Modification of parameter with default

DimitriPapadopoulos:
The default value of keyword argument `strategy` is mutated.

* The default value of argument `strategy` is set to `["motion", "high_pass", "wm_csf"]`.
  https://github.com/nilearn/nilearn/blob/4280451cf70719953d02860e863747dae70372ca/nilearn/input_data/fmriprep_confounds.py#L64
* Then  `_sanitize_strategy()` is called on `strategy`.
  https://github.com/nilearn/nilearn/blob/4280451cf70719953d02860e863747dae70372ca/nilearn/input_data/fmriprep_confounds.py#L238
* Finally, `_sanitize_strategy()` mutates the default value of argument `strategy` to `["motion", "high_pass", "wm_csf", "non_steady_state"]`.
  https://github.com/nilearn/nilearn/blob/4280451cf70719953d02860e863747dae70372ca/nilearn/input_data/fmriprep_confounds.py#L43-L44

This is currently not a real issue, it is equivalent to setting the default value to `["motion", "high_pass", "wm_csf", "non_steady_state"]` which is what we eventually want. However, using a mutable as a default argument is a Python antipattern that should be avoided.

https://lgtm.com/projects/g/nilearn/nilearn/snapshot/0277ef6bc1aaa54a0419a4102beeb2e410a3a306/files/nilearn/input_data/fmriprep_confounds.py?sort=name&dir=ASC&mode=heatmap#xa4fc6df3a3011767:1


bthirion (approved):
I'm fine with the change, as  the documentation clarifies the behavior of the function.

NicolasGensollen (commented):
Thanks @DimitriPapadopoulos ! :+1: 
This makes me realize a few things:

- ``fmriprep_confounds`` currently mutates `strategy` which might not be what we want (or we should at least document it). I think we should avoid functions which both return something and mutates some of their inputs.
- `non_steady_state` is systematically added to the list of strategies although this is not documented (both the addition and the possible value in the list)
- Same thing for `compcor` which adds `high_pass`, the behavior isn't currently documented

WDYT @htwangtw ?

htwangtw:
Thanks @DimitriPapadopoulos and @NicolasGensollen! This is really helpful 👍

> - ``fmriprep_confounds`` currently mutates `strategy` which might not be what we want (or we should at least document it). I think we should avoid functions which both return something and mutates some of their inputs.

Are you referring to the behaviour of `_sanitize_strategy()`? 

> - `non_steady_state` is systematically added to the list of strategies although this is not documented (both the addition and the possible value in the list)
> - Same thing for `compcor` which adds `high_pass`, the behavior isn't currently documented

We can certainly document the two points here. These are quite important things, as they are some commonly ignored things when using fmriprep outputs. 

I can think about how to better address these to avoid mutations on list. The easiest way is perhaps modify a copy of the original list and return as a new variable?
Let me know what you think!

DimitriPapadopoulos:
There are two issues here:

### Mutable default value of an argument

Using a mutable as the default value of an argument has side effects when you modify the mutable passed to the function and the function is called repeatedly. For that reason, using a mutable as the default value of an argument is considered an antipattern:
```python
def foo(array=[0]):
    array.append(1)
    print(array)

foo()
foo()
```
will print:
```
[0, 1]
[0, 1, 1]
```
Therefore, avoid mutables as default values of arguments. Two solutions:
1. Use `None` instead of a mutable as the default value, then if the value of the variable is `None` assign the default value to the passed variable **inside** the function. In short, as in my patch,  write:
   ```
   def foo(strategy=None):
       if strategy is None:
           strategy = ["motion", "high_pass", "wm_csf"]
       [...]
   ```
   instead of:
   ```
   def foo(strategy=["motion", "high_pass", "wm_csf"]):
       [...]
   ```
2. When it exists, you may use the non-mutable equivalent of a mutable. For example instead of a list you may use a tuple:
   ```
   def foo(strategy=("motion", "high_pass", "wm_csf")):
       [...]
   ```
   That's OK if the passed variable `strategy` is a constant and will not be mutated. However, in this specific case, you actually want to mutate `strategy` so if it is mutable, you'll have to copy it first. Much more complicated than the `None` solution, if you want my opinion, so I'd stick with the `None` solution.

### Documentation

Even if the default `strategy` appears to be `["motion", "high_pass", "wm_csf"]` it will end up to be `["motion", "high_pass", "wm_csf", "non_steady_state"]`. This should be documented. Actually it  should be documented that `"high_pass"` is always added when "`compcor`" is present, and "`non_steady_state`" is **always** added.

But then what do you need "`non_steady_state`" for? Just **always** run whatever "`non_steady_state`" stands for!

htwangtw:
> There are two issues here:
> ### Mutable default value of an argument
> 
> Using a mutable as the default value of an argument has side effects when you modify the mutable passed to the function and the function is called repeatedly. For that reason, using a mutable as the default value of an argument is considered an antipattern:
> 
> ```python
> def foo(array=[0]):
>     array.append(1)
>     print(array)
> 
> foo()
> foo()
> ```
> 
> will print:
> 
> ```
> [0, 1]
> [0, 1, 1]
> ```
> 
> Therefore, avoid mutables as default values of arguments. Two solutions:
> 
>     1. Use `None` instead of a mutable as the default value. Then, when the value of the passed variable is `None`, assign the default value **inside** the function. In short, as in my patch,  write:
>        ```
>        def foo(strategy=None):
>            if strategy is None:
>                strategy = ["motion", "high_pass", "wm_csf"]
>            [...]
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        instead of:
>        ```
>        def foo(strategy=["motion", "high_pass", "wm_csf"]):
>            [...]
>        ```
> 
>     2. When it exists, you may use the non-mutable equivalent of a mutable. For example instead of a list you may use a tuple:
>        ```
>        def foo(strategy=("motion", "high_pass", "wm_csf")):
>            [...]
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        That's OK if the passed variable `strategy` is a constant and will not be mutated. However, in this specific case, you actually want to mutate `strategy` so if it is not mutable, you'll have to copy it first. Much more complicated than the `None` solution, if you want my opinion, so I'd stick with the `None` solution.
> 

For default value, I am with you about the `None` solution. However this doesn't really solve the issue as the users can pass their own set of strategy, and `_sanitize_strategy` will still mutate the input when `compcor` is included in `strategy`. I would keep how you handle the default value, but still make sure `_sanitize_strategy` returns a modified copy. 

> ### Documentation
> 
> Even if the default `strategy` appears to be `["motion", "high_pass", "wm_csf"]` it will end up being `["motion", "high_pass", "wm_csf", "non_steady_state"]`. This should be documented. Actually it should be documented that `"high_pass"` is always added when "`compcor`" is present, and "`non_steady_state`" is **always** added.
> 
> But then what do you need "`non_steady_state`" for? Just **always** run whatever "`non_steady_state`" stands for!

Thinking about it, it's probably worth to refactor the code a bit to make sure the function triggered by `non_steady_state` option is always applied. Either way we will need to add these to the documentation.

I can address the documentation issue and the application of `non_steady_state` option later this week. Shall we open an issue for this?

DimitriPapadopoulos:
> For default value, I am with you about the `None` solution. However this doesn't really solve the issue as the users can pass their own set of strategy, and `_sanitize_strategy` will still mutate the input when `compcor` is included in `strategy`. I would keep how you handle the default value, but still make sure `_sanitize_strategy` returns a modified copy.

Ah, right. Indeed it would be great not to modify the passed `strategy`. Since you will always copy it, you may use tuple `("motion", "high_pass", "wm_csf")` as a default value for the passed argument

htwangtw:
Sounds like we found a good solution! This change will effect a lot of test, do you want me to do it @DimitriPapadopoulos ?

NicolasGensollen:
> But then what do you need "non_steady_state" for? Just always run whatever "non_steady_state" stands for!

I agree, and if we only have the `compcor` case, it might be better for `_sanitize_strategy` to raise an error if `high_pass` is missing from the user input strategy. In this case, we could handle strategies as tuples of strings as suggested by @DimitriPapadopoulos . WDYT?

htwangtw:
@DimitriPapadopoulos  I did a PR to your working branch and addressed the related changes. It was tested locally. See if this is good.

DimitriPapadopoulos:
@htwangtw I have merged your changes into my branch. They look OK to me, but core developers have to look into it now. Any way, thank you for looking into this merge request and fixing it.

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -14,38 +15,42 @@
...
 
-def _sanitize_strategy(strategy):
-    """Define the supported denoising strategies."""
-    if not isinstance(strategy, list):
-        raise ValueError("strategy needs to be a list of strings")
+def _check_strategy(strategy):
+    """Ensure the denoising strategies are valid."""
+    if (not isinstance(strategy, tuple)) and (not isinstance(strategy, list)):
+        raise ValueError("strategy needs to be a tuple or list of strings"
+                         f"{type(strategy)}")
```
NicolasGensollen, line 40:
```suggestion
                         f" A {type(strategy)} was provided instead.")
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,38 @@ def fmriprep_confounds(img_files,
...
         - "ica_aroma" confounds derived from ICA-AROMA :footcite:`Pruim2015`.
+          Associated parameter: `ica_aroma`
         - "scrub" regressors for :footcite:`Power2014` scrubbing approach.
+          Associated parameter: `scrub`, `fd_threshold`, `std_dvars_threshold`
 
-        For each supplied strategy, associated parameters will be applied.
-        Otherwise, any values supplied to the parameters are ignored.
+        For each component above, associated parameters will be applied if
+        present. Otherwise, any values supplied to the parameters are ignored.
+        For example, if `strategy=('motion', 'global_signal')` will allow user
```
NicolasGensollen, line 125:
```suggestion
        For example, `strategy=('motion', 'global_signal')` will allow users
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,38 @@ def fmriprep_confounds(img_files,
...
+        to supply input to associated parameter `motion` and `global_signal`;
+        if user pass `wm_csf` parameter, it will not be applied as it's not
```
NicolasGensollen, line 127:
```suggestion
        if users pass `wm_csf` parameter, it will not be applied as it is not
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,38 @@ def fmriprep_confounds(img_files,
...
+        part of the `strategy`.
+
+        There are two additional noise components with no optional parameters.
+
+        - "non_steady_state" denotes volumes collected before the fMRI scanner
```
NicolasGensollen, line 132:
```suggestion
        - "non_steady_state" denotes volumes collected before the :term:`fMRI` scanner
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,38 @@ def fmriprep_confounds(img_files,
...
+          has reached a stable state.
+        - "high_pass" discrete cosines covering low frequencies.
```
NicolasGensollen, line 134: I think the verb is missing here.

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,38 @@ def fmriprep_confounds(img_files,
...
+
+        Non-steady-state volume will always be checked. There's no need to
```
NicolasGensollen, line 136:
```suggestion
        Non-steady-state volumes will always be checked. There's no need to
```

NicolasGensollen (commented):
Thanks @htwangtw @DimitriPapadopoulos !
LGTM, just a few typos and small suggestions on the wording:

nilearn/input_data/fmriprep_confounds.py
htwangtw:
```suggestion
        - "high_pass" add discrete cosines transformation basis regressors to handle low-frequency signal drifts.
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,38 @@ def fmriprep_confounds(img_files,
...
+        For example, `strategy=('motion', 'global_signal')` will allow users
+        to supply input to associated parameter `motion` and `global_signal`;
+        if users pass `wm_csf` parameter, it will not be applied as it is not
+        part of the `strategy`.
+
+        There are two additional noise components with no optional parameters.
 
-        For each supplied strategy, associated parameters will be applied.
-        Otherwise, any values supplied to the parameters are ignored.
+        - "non_steady_state" denotes volumes collected before the :term:`fMRI` scanner
```
NicolasGensollen, line 132:
```suggestion
        - "non_steady_state" denotes volumes collected before
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,38 @@ def fmriprep_confounds(img_files,
...
+          has reached a stable state.
```
NicolasGensollen, line 133:
```suggestion
          the :term:`fMRI` scanner has reached a stable state.
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,38 @@ def fmriprep_confounds(img_files,
...
+        - "high_pass" add discrete cosines transformation basis regressors to handle low-frequency signal drifts.
```
NicolasGensollen, line 134:
```suggestion
        - "high_pass" add discrete cosines transformation
          basis regressors to handle low-frequency signal drifts.
```

NicolasGensollen (commented):
To make PEP8 happy again:

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,39 @@ def fmriprep_confounds(img_files,
...
+        if users pass `wm_csf` parameter, it will not be applied as it is not
+        part of the `strategy`.
+
+        There are two additional noise components with no optional parameters.
 
-        For each supplied strategy, associated parameters will be applied.
-        Otherwise, any values supplied to the parameters are ignored.
+        - "non_steady_state" denotes volumes collected before
+          the :term:`fMRI` scanner has reached a stable state.
+        - "high_pass" add discrete cosines transformation
```
bthirion, line 134:
```suggestion
        - "high_pass" adds discrete cosines transformation
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,39 @@ def fmriprep_confounds(img_files,
...
+          basis regressors to handle low-frequency signal drifts.
+
+        Non-steady-state volumes will always be checked. There's no need to
+        supply this componenet to the strategy.
```
bthirion, line 138:
```suggestion
        supply this component to the strategy.
```

nilearn/input_data/fmriprep_confounds.py
```diff
@@ -98,19 +103,39 @@ def fmriprep_confounds(img_files,
...
         - "compcor" confounds derived from CompCor :footcite:`BEHZADI200790`.
+          When using this noise component, "high_pass" must also be applied.
+          Associated parameter: `compcor`, `n_compcor`
         - "ica_aroma" confounds derived from ICA-AROMA :footcite:`Pruim2015`.
+          Associated parameter: `ica_aroma`
         - "scrub" regressors for :footcite:`Power2014` scrubbing approach.
+          Associated parameter: `scrub`, `fd_threshold`, `std_dvars_threshold`
+
+        For each component above, associated parameters will be applied if
+        present. Otherwise, any values supplied to the parameters are ignored.
```
bthirion, line 124: Unlcearl what "otherwise" referes to here. Could be "if associated parameters are not present" or could be that the components is not specified...

bthirion (commented):
Looks good, thx.

bthirion (approved):
LGTM, thx.

NicolasGensollen (approved):
LGTM, thanks @DimitriPapadopoulos and @htwangtw !

NicolasGensollen:
Actually, there is a small conflict in `nilearn/input_data/tests/test_fmriprep_confounds.py` to solve before we can merge.

DimitriPapadopoulos:
I've rebased the branch, hopefully it's OK.

NicolasGensollen:
Thanks for the rebase!
Merging now that the conflict has been resolved!

PR #3052, opened 2021-11-06
[FIX] Fix memory issue on CircleCI

NicolasGensollen:


bthirion:
Is this related to an issue ? Best,

NicolasGensollen:
>Is this related to an issue ? 

@bthirion it's related to a sporadic memory failure on CircleCI which seem to occur when running this example (see [yesterday's nightly build](https://app.circleci.com/pipelines/github/nilearn/nilearn/3412/workflows/751ca97b-169d-4f40-ae89-821def33430f/jobs/8641) for example).

bthirion:
All right. I guess that the hard part is to check whether this indeed solves the issue.

bthirion (approved):

NicolasGensollen:
>I guess that the hard part is to check whether this indeed solves the issue.

Yes, but since most builds were failing recently, we should see if it works pretty quickly.
Merging to test it out....
Thanks!

PR #3053, opened 2021-11-08
MAINT: Work around SciPy deprecation

larsoner:
SciPy made the `scipy.io.matlab` namespace private. This is a temporary workaround PR until https://github.com/scipy/scipy/pull/15004 is in and these hopefully land in a public namespace, so don't merge yet.

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -18,9 +18,9 @@ import nibabel
 import pandas as pd
 from scipy.io import loadmat
 try:
+    from scipy.io.matlab import MatReadError
+except ImportError:  # SciPy < 1.8
     from scipy.io.matlab.miobase import MatReadError
-except ImportError:
-    from scipy.io._matlab.miobase import MatReadError
 from sklearn.utils import Bunch, deprecated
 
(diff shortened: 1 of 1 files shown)


larsoner:
BTW at some point you might consider adding a pip-pre CI job to catch these sort of things, if you want to see what we do in MNE it's here:

https://github.com/mne-tools/mne-python/blob/main/tools/github_actions_dependencies.sh#L7-L30

NicolasGensollen:
Thanks @larsoner ! :+1: 
I'll definitely take a look at how you handle this in MNE.
I don't mind merging this fix while your SciPy PR is reviewed if it makes things easier for you. We can always revert back if it gets merged at some point.

larsoner:
This implements the new-namespace solution from scipy/scipy#15004, and tests pass in MNE-Python using this branch still (at least locally, CIs are still running -- but I think it'll work). So this is ready for review/merge from my end now

jeromedockes (approved):
thanks!

agramfort (approved):

agramfort:
🙏 @larsoner

PR #3046, opened 2021-10-29
[FIX] scrubbing parameter names in `nilearn.input_data.fmriprep_confounds` not matching docstring

htwangtw:
In `nilearn.input_data.fmriprep_confounds`, two parameter names are not matching the docstrings.

`fd_thresh` -> `fd_threshold`
`std_dvars_thresh` -> `std_dvars_threshold`

PR #3054, opened 2021-11-09
[FIX] Fix example `fmriprep_confounds`

NicolasGensollen:
Fix example broken in #3046


bthirion (approved):
How comes we did not detect it ?

htwangtw:
> How comes we did not detect it ?

Likely because I didn't trigger a full build so the CI only ran on the code modified

NicolasGensollen:
Thanks!
Merging...

PR #3065, opened 2021-11-18
[ENH] New `maskers` module

### LINKED ISSUE #3014: A proposal to reorganize Maskers and I/O functions
Per discussion in #2946 and last week's developers call, we are thinking about restructuring how certain classes and functions, especially Masker objects, are organized in nilearn. We want to ensure that the community has the ability to provide feedback before we make any changes.

The first issue is that the module name `input_data` is somewhat ambiguous. Therefore, we propose creating a new module, `maskers`, which would contain all Masker objects, including volumetric (i.e., the existing `NiftiMasker`s) and potential future surface maskers. This would remain distinct from the `masking` module, which contains functions for creating and operating on masks.

Secondly, #2946 will introduce a new function for loading fMRIPrep-generated confounds for use with Maskers or in GLMs. We defaulted to placing this in `input_data`, but, should we rename that module to `maskers`, the new function will no longer fit. We are considering a new `interfaces` module, which would contain submodules dedicated to external tools (e.g., fMRIPrep), specifications (e.g., BIDS, NIDM), etc. These submodules would then contain functions and classes used to ingest or write out files associated with each tool. For example, the new function would be located at `nilearn.interfaces.fmriprep.load_confounds`.

Finally, given that some users have expressed an interest in using image-loading functions currently in the private `_utils` module, we are also considering creating a new `io` module for input/output operations. Some functions that would end up in that function are `nilearn._utils.load_niimg`, `nilearn.image.load_img`, and `nilearn.image.get_data`.

Any proposed changes that we actually implement will go through a lengthy deprecation cycle, so please don't expect any breaking changes in the next couple of releases.

NicolasGensollen:
Closes #3014

This PR is the second step at addressing #3014, it:

- creates a new `maskers` module
- moves all NiftiMaskers to the `maskers` module but keeps the old imports functional with a warning for backward compatibility
- updates the code and examples to import maskers from the `maskers` module
- updates the docs

Additional changes:

- `nilearn.input_data.masker_validation.check_embedded_nifti_masker` is now private and renamed to `_check_embedded_nifti_masker` 
- The `BaseMasker` class is now public and documented in the API
- The function `filter_and_extract` is now private and renamed to `_filter_and_extract`
-  The function `filter_and_mask` is now private and renamed to `_filter_and_mask`

Before:

```python
```

After:

```python
```

Deprecated but still works:

```python
```
```
/home/nicolas/GitRepos/nilearn-fork/nilearn/input_data/__init__.py:27: FutureWarning: 
The import path 'nilearn.input_data' is deprecated in version 0.9. 
Importing from 'nilearn.input_data' will be possible at least until release 0.13.0. 
Please import from 'nilearn.maskers' instead.
  warnings.warn(message, FutureWarning)
```


bthirion (approved):
LGTM. I could not find any issue (hard to sport though...)

doc/modules/reference.rst
```diff
@@ -229,20 +229,20 @@ the :ref:`user guide <user_guide>` for more information and usage examples.
    load_confounds
    load_confounds_strategy
 
-.. _io_ref:
+.. _maskers_ref:
 
-:mod:`nilearn.input_data`: Loading and Processing Files Easily
-==============================================================
+:mod:`nilearn.maskers`: Loading and Processing fMRI data Easily
```
tsalo, line 234:
```suggestion
:mod:`nilearn.maskers`: Loading and Processing fMRI Data Easily
```

I don't think I'd commit to fMRI data specifically though. Maybe "Neuroimaging" or "MR"?
However, if not, then data just needs to be capitalized since it's title case.

doc/modules/reference.rst
```diff
@@ -229,20 +229,20 @@ the :ref:`user guide <user_guide>` for more information and usage examples.
    load_confounds
    load_confounds_strategy
 
-.. _io_ref:
+.. _maskers_ref:
```
tsalo, line 232: Did you replace all uses of `io_ref` throughout the documentation with `maskers_ref` too?

nilearn/decoding/searchlight.py
```diff
@@ -23,7 +23,7 @@
 
 from .. import masking
 from ..image.resampling import coord_transform
-from ..input_data.nifti_spheres_masker import _apply_mask_and_get_affinity
+from nilearn.maskers.nifti_spheres_masker import _apply_mask_and_get_affinity
```
tsalo, line 26: Why switch to absolute import here?

nilearn/decoding/space_net.py
```diff
@@ -26,7 +26,7 @@
 from joblib import Memory, Parallel, delayed
 from sklearn.preprocessing import LabelBinarizer
 from sklearn.metrics import accuracy_score
-from ..input_data.masker_validation import check_embedded_nifti_masker
+from nilearn.maskers.masker_validation import check_embedded_nifti_masker
```
tsalo, line 29: Why switch to absolute import here?

nilearn/decomposition/base.py
```diff
@@ -19,8 +19,8 @@
 from .._utils.cache_mixin import CacheMixin, cache
 from .._utils.niimg import _safe_get_data
 from .._utils.niimg_conversions import _resolve_globbing
-from ..input_data import NiftiMapsMasker
-from ..input_data.masker_validation import check_embedded_nifti_masker
+from nilearn.maskers import NiftiMapsMasker
+from nilearn.maskers.masker_validation import check_embedded_nifti_masker
```
tsalo, line 23: Why switch to absolute import here?

nilearn/input_data/__init__.py
```diff
@@ -1,11 +1,31 @@
...
+.. code-blocks::python
+    from nilearn.input_data import NiftiMasker
+
+becomes:
+
+.. code-blocks::python
+    from nilearn.maskers import NiftiMasker
+
+Note that all imports that used to work will continue to do so with
+a simple warning.
```
tsalo, line 17:
```suggestion
a simple warning until 0.11.0.
```

Maybe specify the deprecation cycle length here? I think it was two "major" releases, but I could be wrong.

nilearn/input_data/__init__.py
```diff
@@ -1,11 +1,31 @@
 """
-The :mod:`nilearn.input_data` module includes scikit-learn tranformers.
+The :mod:`nilearn.input_data` module used to include masker objects.
+It is deprecated since release 0.8.2 in favor of the
```
tsalo, line 3: Per today's call, should this be 0.9.0?

tsalo (approved):
There were a few things I noticed that you can take or leave. The main things were:

1. Some relative imports have been changed to absolute imports. They look a little weird, since they're generally in the middle of a bunch of relative imports that haven't been changed. I added comments to the first few I saw, but after realizing that this was a larger pattern I figured I should stop.
2. Should the moved class/module docstrings include `versionchanged` directives?
3. I see 0.8.2 referenced, but maybe it should be 0.9.0 depending on what was decided in today's call.

doc/modules/reference.rst
NicolasGensollen: Thanks for checking! I couldn't find any. Also, the documentation full build runs in strict mode and would have crashed if there were broken references somewhere.

nilearn/input_data/__init__.py
NicolasGensollen: Good question. I think the plan was to keep these imports working (ping @jeromedockes). 
If we go for a deprecation cycle, I'd vote for a longer one.

nilearn/input_data/__init__.py
NicolasGensollen: Thanks! Yes, I need to update the 0.8.2

nilearn/decoding/searchlight.py
NicolasGensollen: I think this is one of our conventions which is not clearly written but that we try to enforce. See #1919 for example, and also the first comment of @emdupre in #3016.
I agree they look a little weird in the middle of all these relative imports though.

NicolasGensollen:
Thanks for the review @tsalo !

>Should the moved class/module docstrings include versionchanged directives?

We can. I didn't do it because there is no actual change to the objects themselves, but I agree this could be useful information for users.

nilearn/decoding/searchlight.py
tsalo: Ah, so this is basically a matter of slowly replacing the relative imports with absolute ones, but you're only changing the imports directly affected by the new module. Thanks for clarifying!

doc/modules/reference.rst
jeromedockes: maybe something like "extracting [data,signals,time series?] from (the relevant regions of) brain images"? this module isn't really about loading (from files) and "processing" is very generic

nilearn/input_data/__init__.py
jeromedockes: yes I think the old imports should keep working forever or at least for a very long time. the size of this PR (87 files) gives a hint as to how annoying it will be for users to change all their imports of the `input_data` module (and the situation is even worse for client code that will have to support both new and old nilearn versions)

jeromedockes: But I agree with @tsalo  that we should make a decision about that so we can be more explicit and tell users until when existing code will keep working

NicolasGensollen: Fair enough. What about doubling the usual deprecation cycle length? This would lead us to release 0.13.0.

nilearn/input_data/__init__.py
NicolasGensollen: Done in d697c4f

doc/modules/reference.rst
NicolasGensollen: I agree. I replaced by "Extracting Signals from Brain Images" in 2b94a62

nilearn/input_data/__init__.py
jeromedockes: how about "at least until release 0.13.0" ?

nilearn/input_data/__init__.py
```diff
@@ -1,11 +1,31 @@
...
 """
-from .nifti_masker import NiftiMasker
-from .multi_nifti_masker import MultiNiftiMasker
-from .nifti_labels_masker import NiftiLabelsMasker
-from .nifti_maps_masker import NiftiMapsMasker
-from .nifti_spheres_masker import NiftiSpheresMasker
+import warnings
+
+
+from nilearn.maskers import (
```
jeromedockes, line 22: backwards compatibility must not be broken without a deprecation cycle for anything that is in the public API currently. For example WDYT of the `input_data` submodules, as their name does not start with an underscore?
```
from nilearn.input_data import nifti_maps_masker
```
from https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces:
> Even with `__all__` set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.

I'm not sure everyone would consider these modules internal just because their name isn't in `__all__`

nilearn/maskers/tests/test_base_masker.py
```diff
@@ -6,7 +6,7 @@
 from numpy.testing import assert_array_almost_equal
 import nibabel
 
-from nilearn.input_data.nifti_masker import filter_and_mask
```
jeromedockes, line 9: where do we test that the old imports still work?

doc/whats_new.rst
```diff
@@ -169,8 +173,8 @@ NEW
 - :func:`nilearn.datasets.load_mni152_wm_mask` loads mask from the white-matter
   MNI152 template.
 - :func:`nilearn.image.binarize_img` binarizes images into 0 and 1.
-- :class:`nilearn.input_data.NiftiMasker`,
-  :class:`nilearn.input_data.MultiNiftiMasker`, and objects relying on such maskers
+- :class:`nilearn.maskers.NiftiMasker`,
```
jeromedockes, line 176: shouldn't we update the name just for the whatsnew of the next release?

nilearn/decoding/searchlight.py
jeromedockes, line 26: for example this import and the one in decoder.py wouldn't work anymore with the current interface

nilearn/input_data/__init__.py
NicolasGensollen: Sounds good to me! Added in dacf94c

doc/whats_new.rst
NicolasGensollen: Yes, I hesitated between de-referencing the old entries and updating the old references. I ended up doing the latter mostly for user convenience and also because entries are usually not related to where the maskers live but to functionalities of the objects. I agree that doing so modify the "history" and conveys the impression that maskers were always in the "maskers" module.
I don't have a strong preference, so we can go with the former if you think this is better.

nilearn/input_data/__init__.py
NicolasGensollen: Thanks @jeromedockes, this is a really good point.
I think the new implementation (see 253f20b) should solve these issues.

nilearn/input_data/__init__.py
```diff
@@ -1,12 +1,31 @@
...
+
+from nilearn import maskers  # noqa:F401
+
+warnings.warn("The module 'input_data' is deprecated since 0.9.0, "
+              "Importing maskers from 'input_data' will be possible "
+              "at least until release 0.13.0. Please import maskers "
+              "from the 'maskers' module instead.",
+              FutureWarning)
 
+sys.modules[__name__] = sys.modules['nilearn.maskers']
```
jeromedockes, line 30: this makes sense and from the documentation on the import system and in particular [this part](https://docs.python.org/3/reference/import.html#loading) seems like it should work, and in practice seems to give the desired result at least on my machine! I'm just wondering if there are traps we're not aware of; by any chance do you have an example of this solution being used in another project, or suggested in some resource?

(note of course the backwards compatibility will not be 100% perfect, eg the modules' `__name__` and objects' `__module__` will change, but I think that's acceptable)

doc/whats_new.rst
jeromedockes: I see, that makes sense! we can keep it like this then

nilearn/input_data/__init__.py
NicolasGensollen: I don't know if this solution is used in other projects. I'll take a look around! :eyes:

NicolasGensollen: @jeromedockes I've asked around on the PARIETAL engineers channel, but got very few answers.
Guillaume pointed me to this PR on scikit-learn: https://github.com/scikit-learn/scikit-learn/pull/14939 where they had to do something similar. It is not exactly the same though, as they don't rename the module (meaning the name of the folder), but the files themselves to prepend them with "_". They basically keep two versions of each file "file.py" and "_file.py" but the former imports from the latter with some deprecation warnings.
Our use case is slightly different since we keep all files, but move them to another folder. We could have the same mechanism, but I don't think it is necessary.
WDYT?

jeromedockes: Yes I agree the situation is a  bit different. we can try it as it is in the PR then. we get some slightly awkward things like the output of the script below but I guess that's unavoidable
```python
from nilearn.input_data import NiftiMapsMasker as masker1
from nilearn.input_data.nifti_maps_masker import NiftiMapsMasker as masker2

print(masker1 is masker2)
print(masker1.__module__)
print(masker2.__module__)
```

```
False
nilearn.maskers.nifti_maps_masker
nilearn.input_data.nifti_maps_masker
```

NicolasGensollen: @jeromedockes I've implemented a possible alternative solution inspired by the PR from scikit-learn above (see commit d2fbbe8). This should solve the issue in your last comment. However, I'm still unsure whether or not this is a better solution. 

Potential issues:

- You cannot import private objects from `input_data`. So all user code doing things like `from nilearn.input_data.nifti_masker import _get_mask_strategy` will break. You can argue that this is a risk you're willing to take if you import private things, but we could also import them manually in each python file in `input_data` to be backward compatible even for private objects.
- When you import `input_data` the first time, you get 5 warnings due to the imports in `__init__.py`. This is not a big issue and there might be ways to avoid this. I can take a closer look if we prefer this solution over the first one. 

Also, I realized a few things:

- `nilearn.input_data.masker_validation.check_embedded_nifti_masker` looks like it is public, but it is not documented in the API. Should we make it private or public?
- Same comment for the content of `base_masker.py` (`BaseMasker` class and `filter_and_extract` function)

jeromedockes: Hi, thanks for looking at this! Re. the potential issues:

- even though nilearn documentation doesn't explicitly state we follow it, the convention that `_` names are for internal/private objects is used everywhere. client code shouldn't use them and we don't need to avoid breaking code that does.
- for the warnings when importing `input_data`, could we just put one warning in `input_data.__init__` and remove those in the submodules?

That being said if you prefer the first solution modifying `sys.modules`, I'm ok with it!

> Also, I realized a few things:

IMO these should be private, they are clearly internals of nilearn's estimators and maskers and it seems unlikely users would use them, especially as they are undocumented and not listed in `__all__`. Except perhaps the `BaseMasker` if we think it's possible for users to inherit from it to create their own maskers

NicolasGensollen:
@jeromedockes I think this should be good now.
I've made the above mentioned functions private, and turned `BaseMasker` to a fully public class such that users can inherit from it when building custom maskers.
The warnings were also simplified (only one warning in `input_data/__init__.py`) as you suggested.

jeromedockes (approved):
Thanks @NicolasGensollen , I think it's good now! this is a big one so it might be useful for someone else to have another look.

NicolasGensollen:
@tsalo Do you want to take another look at it since a few things have changed since your approval?

tsalo (approved):
The new changes look great to me. The only thing I would suggest is updating the PR's summary comment to include something about making several functions private and adding BaseMasker to the public API. Also, the example message in that comment says 0.8.2 instead of 0.9.

NicolasGensollen:
Thanks @tsalo ! I've updated the PR's summary with the latest changes and changed the warning message in the comment.

NicolasGensollen:
Thank you for the reviews!
Time to merge...

Issue #2036, opened 2019-04-19
Harvard-Oxford atlas: some atlases have wrong number of labels


illdopejake:
I'm was working on harmonizing the atlas outputs, but I found that some (5/16) of the Harvard-Oxford maps have the wrong number of labels. In other words, the number of atlas.labels does not match the number of np.unique(atlas.maps.get_data()). 

```python
In [80]: for name in names: 
    ...:     x = datasets.atlas.fetch_atlas_harvard_oxford(name) 
    ...:     img = image.load_img(x.maps).get_data() 
    ...:     try: 
    ...:         assert(all(np.unique(img) == list(range(len(np.unique(img)))))) 
    ...:     except: 
    ...:         print('{}: atlas labels are not a susquential list of' 
    ...:               ' numbers starting at 0'.format(name)) 
    ...:     try: 
    ...:         assert(len(x.labels) == len(np.unique(img))) 
    ...:     except: 
    ...:         print('{}: number of supplied labels do not match' 
    ...:               ' number of atlas labels'.format(name)) 
    ...:                                                                                      
```
produces
```                         
cort-maxprob-thr50-1mm: atlas labels are not a susquential list of numbers starting at 0
cort-maxprob-thr50-1mm: number of supplied labels do not match number of atlas labels
cort-prob-1mm: number of supplied labels do not match number of atlas labels
cort-prob-2mm: number of supplied labels do not match number of atlas labels
sub-prob-1mm: atlas labels are not a susquential list of numbers starting at 0
sub-prob-1mm: number of supplied labels do not match number of atlas labels
sub-prob-2mm: atlas labels are not a susquential list of numbers starting at 0
sub-prob-2mm: number of supplied labels do not match number of atlas labels
```
This does not seem to be an issue of background (i.e. 0) not being counted in atlas.labels.

I'm not an expert on this atlas. Any suggestions as to why this might be?

illdopejake:
might explain #1871

jeromedockes:
in the thresholded versions of harvard-oxford, some regions may be empty so their label doesn't appear in the image. is that a prolem?

illdopejake:
Good point. That indeed seems to be the problem. I'll get rid of the extra labels then

jeromedockes:
[...]

PR #3070, opened 2021-11-22
[FIX] Fix labels in `fetch_atlas_destrieux_2009`

NicolasGensollen:
See #2036 

This PR adds a test failing with extra labels present in the csv file but not in the image, and adds a fix which removes these labels. This PR only targets `fetch_atlas_destrieux_2009`.

### DIFF
--- a/nilearn/datasets/tests/test_atlas.py
+++ b/nilearn/datasets/tests/test_atlas.py
@@ -372,14 +372,14 @@ def test_fetch_atlas_destrieux_2009(tmp_path, request_mocker, n_labels,
 
 def test_fetch_atlas_destrieux_2009_warning_clean_labels(request_mocker):
     """Tests that ``fetch_atlas_destrieux_2009`` gives a FutureWarning
-    when ``clean_labels`` isn't specified explicitely.
+    when ``clean_labels`` isn't specified explicitly.
     """
     request_mocker.url_mapping["*destrieux2009.tgz"] = _destrieux_data()
     with pytest.warns(FutureWarning,
                       match=("Default value for parameter `clean_labels` "
                              "will change from ``False`` to ``True`` "
                              "in Nilearn 0.10.")):
-        bunch = atlas.fetch_atlas_destrieux_2009()
+        atlas.fetch_atlas_destrieux_2009()
     assert request_mocker.url_count == 1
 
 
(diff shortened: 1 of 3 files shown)


nilearn/datasets/tests/test_atlas.py
```diff
@@ -321,32 +321,40 @@ def test_fetch_coords_seitzman_2018(request_mocker):
...
 
 
-def _destrieux_data():
+def _destrieux_data(n_labels=10):
     data = {"destrieux2009.rst": "readme"}
+    atlas = np.random.randint(0, 10, (10, 10, 10))
+    atlas_img = nibabel.Nifti1Image(atlas, np.eye(4))
+    # Create 11 labels. The fetcher should be able to remove
+    # the extra label ('10') not present in the atlas image
+    labels = "\n".join([f"{idx},label {idx}" for idx in range(n_labels)])
```
bthirion, line 330: Sorry, where is the 11th label created ?

bthirion (commented):
I'm also wondering why this would be specific to Destrieux ?

nilearn/datasets/tests/test_atlas.py
NicolasGensollen: Yes, I forgot to remove this comment. Thanks for noticing..
The number of labels in the CSV file is now passed as an argument and used for test parametrization.
I removed the comment and added more informative docstrings for these functions in 16ca0d0.

NicolasGensollen:
>I'm also wondering why this would be specific to Destrieux ?

Based on #2036, it seems like this issue also happens in the Harvard Oxford fetcher for some combinations of the parameters. The cleaning function implemented in this PR should also work for Harvard Oxford but I was planning to fix this in another PR since the fetcher is a little bit more complicated.

I was also debating the need for a way to inform the users that the output has changed (or even implement a deprecation cycle). Technically, it could break user code expecting different labels, although this seems unlikely. We could for example add a parameter `clean_labels=False` to the fetcher, which will switch to True in 2 releases, but it feels a little too much. WDYT? (also pinging @jeromedockes)

bthirion:
Regarding HO: OK, fair enough.
Regarding bw compatibility: you're right, this may break user code. I 'd go for standard deprecation warnings etc.
Thx !

bthirion (commented):
We still have to warn that the default behavior is going to change in v0.10.

bthirion (approved):
LGTM, thx !

NicolasGensollen:
Thanks! Merging....

PR #3073, opened 2021-11-24
[ENH] Add displays to API

NicolasGensollen:
This PR reorganizes the objects from `nilearn/plotting/displays.py` and adds them to the API.
Next steps are to decide what should be public / private and to improve the docstrings.


bthirion:
Please let me know when it gets ready for a review.

nilearn/plotting/displays/axes.py
```diff
@@ -0,0 +1,581 @@
...
+    ----------
+    %(ax)s
+    direction : {'x', 'y', 'z'}
+        The directions of the view.
+
+    coord : :obj:`float`
+        The coordinate along the direction of the cut.
+
+    plot_abs : :obj:`bool`, optional
+        ??.
```
bthirion, line 350: I think that it plots the absolute values of the image.

bthirion (commented):
Apparently there are a few uncovered lines. LGTM otherwise. Thx !

nilearn/_utils/docs.py
```diff
@@ -314,7 +319,7 @@
 'y', 'z', 'yx', 'xz', 'yz'}, optional
     Choose the direction of the cuts:
 
-        - 'x': sagittal
+        - 'x': sagital
```
tsalo, line 322: Shouldn't this be sagittal?

NicolasGensollen: Yes, thanks for spotting this!

NicolasGensollen:
[direct link](https://8820-1235740-gh.circle-artifacts.com/0/dev/auto_examples/01_plotting/plot_demo_more_plotting.html#sphx-glr-auto-examples-01-plotting-plot-demo-more-plotting-py) to the example.

examples/01_plotting/plot_demo_more_plotting.py
```diff
@@ -259,29 +339,39 @@
...
-#########################################
+###############################################################################
 # Saving plots to file
 # --------------------
-# Finally, saving the plots to file with two different ways
+#
+# Finally, we can save a plot to file in two different ways:
+#
+# First, we can save the contrast maps plotted with the function
+# :func:`nilearn.plotting.plot_stat_map` using the inbuilt parameter
```
bthirion, line 357:
```suggestion
# :func:`nilearn.plotting.plot_stat_map` using the built-in parameter
```

bthirion (commented):
The example looks great, thx !

bthirion (commented):
The PR LGTM but there are a few codecov complaints.

nilearn/plotting/tests/test_displays.py
```diff
@@ -2,142 +2,189 @@
...
+                         [XSlicer, YSlicer, ZSlicer, YXSlicer, YZSlicer])
+def test_stacked_slicer(slicer, img, tmp_path):
+    """Tests for saving to file with stacked slicers."""
+    cut_coords = (3, 3) if slicer in [YXSlicer, YZSlicer] else 3
+    slicer = slicer.init_with_figure(img=img, cut_coords=cut_coords)
     slicer.add_overlay(img, cmap=plt.cm.gray)
     # Forcing a layout here, to test the locator code
-    with tempfile.TemporaryFile() as fp:
-        slicer.savefig(fp)
+    slicer.savefig(tmp_path / "out.png")
```
bthirion, line 83: just curious: why do you replace this ?

bthirion (commented):
LGTM otherwise.

nilearn/plotting/tests/test_displays.py
jeromedockes: that is old code, from before we were using pytest and the `tmp_path` fixture was available -- in new tests, we would use `tmp_path` as it is a bit simpler. the older way, using `tempfile`, still works fine though

nilearn/plotting/displays/__init__.py
```diff
@@ -0,0 +1,34 @@
+"""
+Display objects and utilities.
+
+These objects are returned by plotting functions
+from the :mod:`~nilearn.plotting` module.
+"""
+
+from .axes import BaseAxes, CutAxes, GlassBrainAxes
```
jeromedockes, line 8: I don't think it's important but from comments in other PRs I understood the current trend was to prefer absolute imports?

doc/whats_new.rst
```diff
@@ -123,6 +123,14 @@ Changes
   `#2821 <https://github.com/nilearn/nilearn/pull/2821>`_ in order to match BIDS
   semantics. The use of `sessions` in :func:`nilearn.image.clean_img` is deprecated and
   will be removed in 0.10.0.
+- Display objects have been reorganized. For example, Slicers (like the
```
jeromedockes, line 126: maybe we should make these new modules start with `_` at least for now? that gives us the freedom to do some refactoring later if we want. also from the users' point of view this reorganization shouldn't matter because they can still import everything from `displays` as they used to.

jeromedockes (commented):
thanks. I think we should add a more advanced plotting example that shows how to actually use these display objects (beyond adding a title which can be done by passing the title to the function that creates them), to access and manipulate the matplotlib axes, artists etc. but that should probably be done in another PR

jeromedockes (approved):
LGTM, thanks!

doc/whats_new.rst
```diff
@@ -123,6 +123,14 @@ Changes
...
+  :class:`~nilearn.plotting.displays.OrthoSlicer`) are all in file
+  `nilearn/plotting/displays/slicers.py`, and Projectors (like the
```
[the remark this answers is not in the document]
NicolasGensollen: Thanks!

tsalo (approved):
LGTM!

bthirion (approved):
Thx!

bthirion:
Merging, then. This is a big one.

PR #3077, opened 2021-12-03
Matplotlib backend switch warning

### LINKED ISSUE #3063: nilearn.plotting overrides the matplotlib backend, causing troubles on remote SSH development
I am working on a remote settings, accessing a machine (drago/margaret for INRIA context) through a PyCharm Remote Interpreter.
I have been struggling to display nilearn figures in the scientific view of PyCharm, leading to a painful process of saving the image => rsync figures dir => visualize figures once there are on personal laptop.

I did a simple experiment drawing first a simple `plot(list(range(10)))` which is displayed and then a nilearn figure which does not show up (script at the end of the issue).

I found out the reason for this, which is that in `nilearn.plotting.__init__`, the matplotlib backend is set to `agg` if not in `[inline, agg]`, but the backend used initially (which is working for me) is `module://backend_interagg`.

Setting the backend to the initial value after importing `nilearn.datasets` fixes the problem.

I don't know exactly how those matplotlib backends work, if the list of accepted backends in `nilearn.datasets.__init__` could be extended or if we could have a more robust list of valid matplotlib backend. Also, I feel that nilearn shouldn't override matplotlib backend silently? 

Nilearn version: 0.7.1

### Expected behavior

Nilearn does not override matplotlib backend silently.

### Actual behavior

Nilearn sets backend to "agg" if not in ["inline", "agg"], causing trouble with remote SSH development.

Run the following script through SSH remote interpreter

```python
import matplotlib
initial_backend = matplotlib.get_backend().lower()
print(initial_backend)

import numpy as np

arr = np.zeros((100, 100))
plt.figure()
plt.plot(list(range(10)))
plt.show()
# >> The show is displayed in PyCharm

print(matplotlib.get_backend().lower())
# >> Backend has switched to "agg"

sample_brain_map = datasets.fetch_neurovault_motor_task().images[0]
plotting.plot_stat_map(
    sample_brain_map,
    threshold=3,
    title="Before setting back the backend",
)

plt.show()
# >> Does not show up

matplotlib.use(initial_backend)
plotting.plot_stat_map(
    sample_brain_map,
    threshold=3,
    title="After setting back the backend",
)

plt.show()
# >> Shows up

```

RaphaelMeudec:
PR that tackles elements said in #3063 

This PR contains:

- more robust backend switch using `matplotlib.use` instead of testing hard-coded backends
- warning if matplotlib backend is switched
- tests

Closes #3063

### DIFF
--- a/nilearn/plotting/tests/test_matplotlib_backend.py
+++ b/nilearn/plotting/tests/test_matplotlib_backend.py
@@ -1,0 +1,44 @@
+from unittest.mock import patch
+
+SKIP_REASON = "matplotlib missing; necessary for this test"
+try:
+    import matplotlib  # noqa: F401
+except ImportError:
+    MATPLOTLIB_INSTALLED = False
+else:
+    MATPLOTLIB_INSTALLED = True
+import pytest
+
+from nilearn.plotting import _set_mpl_backend
+
+
+@pytest.mark.skipif(not MATPLOTLIB_INSTALLED, reason=SKIP_REASON)
+@patch("matplotlib.use")
+@patch("matplotlib.get_backend", side_effect=["backend_1", "backend_2"])
+def test_should_raise_warning_if_backend_changes(*_):
+    # The backend values returned by matplotlib.get_backend are different.
+    # Warning should be raised to inform user of the backend switch.
+    with pytest.warns(UserWarning, match="Backend changed to backend_2..."):
+        _set_mpl_backend()
+
+
+@pytest.mark.skipif(not MATPLOTLIB_INSTALLED, reason=SKIP_REASON)
+@patch("matplotlib.use")
+@patch("matplotlib.get_backend", side_effect=["backend_1", "backend_1"])
+def test_should_not_raise_warning_if_backend_is_not_changed(*_):
+    # The backend values returned by matplotlib.get_backend are identical.
+    # Warning should not be raised.
+    with pytest.warns(None):
+        _set_mpl_backend()
+
+
+@pytest.mark.skipif(not MATPLOTLIB_INSTALLED, reason=SKIP_REASON)
+@patch("matplotlib.use", side_effect=[Exception("Failed to switch backend"), True])
+def test_should_switch_to_default_agg_backend_if_current_backend_fails(use_mock):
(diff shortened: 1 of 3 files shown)


NicolasGensollen (commented):
Thanks @RaphaelMeudec !
The fix LGTM (just missing a whats_new entry).
Pinging @jeromedockes since he had some reluctance in the linked issue. (We can also talk about this today during the office hour or on Monday).

nilearn/plotting/tests/test_matplotlib_backend.py
```diff
@@ -0,0 +1,46 @@
...
+        _set_mpl_backend()
+
+
+@pytest.mark.skipif(not MATPLOTLIB_INSTALLED, reason=SKIP_REASON)
+@patch("matplotlib.use",
+       side_effect=[Exception("Failed to switch backend"), True])
+def test_should_switch_to_agg_backend_if_current_backend_fails(use_mock):
+    # First call to `matplotlib.use` raises an exception, hence the default Agg
+    # backend should be triggered
+    _set_mpl_backend()
```
jeromedockes, line 42: here also you could catch the warning

jeromedockes (approved):
LGTM, thanks!

nilearn/plotting/tests/test_matplotlib_backend.py
RaphaelMeudec: I don't think so, because if the backend used during the test is already `Agg`, the warning will not be raised. Hence, I'd rather not test the warning here as it could lead to random fails + it is tested on its own

jeromedockes: I meant suppress the warning to avoid having it pollute pytest's warning summary. but actually it doesn't show up because:
- conftest.py specifies that we always use agg in tests
- and even if that wasn't the case, use is mocked but not get_backend so we never have new_backend != current_backend

so we can keep it as it is

NicolasGensollen (approved):
LGTM, thanks!

bthirion (approved):
LGTM, thx !

bthirion:
Merging then.

PR #3080, opened 2021-12-06
[FIX] Fix labels HarvardOxford fetcher

### LINKED ISSUE #2036: Harvard-Oxford atlas: some atlases have wrong number of labels

NicolasGensollen:
Fixes #2036 

For some atlases, `fetch_atlas_harvard_oxford` does not return the same number of labels as different integer values in the maps:

```python

# Do not include probabiltic atlases in the list
atlases = ["cort-maxprob-thr0-1mm", "cort-maxprob-thr0-2mm",
           "cort-maxprob-thr25-1mm", "cort-maxprob-thr25-2mm",
           "cort-maxprob-thr50-1mm", "cort-maxprob-thr50-2mm",
           "cortl-maxprob-thr0-1mm", "cortl-maxprob-thr0-2mm",
           "cortl-maxprob-thr25-1mm", "cortl-maxprob-thr25-2mm",
           "cortl-maxprob-thr50-1mm", "cortl-maxprob-thr50-2mm",
           "sub-maxprob-thr0-1mm", "sub-maxprob-thr0-2mm",
           "sub-maxprob-thr25-1mm", "sub-maxprob-thr25-2mm",
           "sub-maxprob-thr50-1mm", "sub-maxprob-thr50-2mm"]

for name in atlases:
    b = fetch_atlas_harvard_oxford(name, clean_labels=False)
    data_maps = get_data(b.maps)
    try:
        # Same number of labels in list and in image
        assert len(b.labels) == len(np.unique(data_maps))
        # Labels in image are consecutive integers from 0 to Nlabel - 1
        assert np.all(np.unique(data_maps) == np.arange(len(b.labels)))
        # Sets of labels are equal
        assert set(np.unique(data_maps)) == set(range(len(b.labels)))
    except:
        print(name)
        print(set(range(len(b.labels))) - set(np.unique(data_maps)))
```
```
cort-maxprob-thr50-1mm
{10}
cortl-maxprob-thr50-1mm
{19, 20, 93}
cortl-maxprob-thr50-2mm
{93, 94}
```

For now, this PR implements a simple fix which consists in removing labels which corresponding integer value does not appear in the image. For example, the 11th label of `cort-maxprob-thr50-1mm` ('Superior Temporal Gyrus, posterior division') is removed because 10 does not appear in the image:

```python
b = fetch_atlas_harvard_oxford("cort-maxprob-thr50-1mm", clean_labels=True)
data_maps = get_data(b.maps)
assert len(b.labels) == len(np.unique(data_maps))
assert "Superior Temporal Gyrus, posterior division" not in b.labels
```

However, the image will still have non consecutive integer values which means that the following is not verified:

```python
assert set(np.unique(data_maps)) == set(range(len(b.labels)))
```

If this is an issue, I think we will have to re-label the image. WDYT?

### DIFF
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -380,7 +380,7 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
         resume=resume,
         verbose=verbose)
     atlas_niimg = check_niimg(atlas_img)
-    label_idx_to_names = {k:v for k,v in enumerate(names)}
+    label_idx_to_names = {k: v for k, v in enumerate(names)}
     if not symmetric_split or is_lateralized:
         if not is_probabilistic and clean_labels:
             valid_idx = np.unique(get_data(atlas_niimg))
@@ -392,7 +392,7 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
     new_atlas_niimg = new_img_like(atlas_niimg,
                                    new_atlas_data,
                                    atlas_niimg.affine)
-    label_idx_to_names = {k:v for k,v in enumerate(new_names)}
+    label_idx_to_names = {k: v for k, v in enumerate(new_names)}
     if not is_probabilistic and clean_labels:
         valid_idx = np.unique(get_data(new_atlas_niimg))
         new_names = [label_idx_to_names[i] for i in valid_idx]
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -276,7 +276,7 @@ def fetch_atlas_destrieux_2009(lateralized=True, data_dir=None, url=None,
 @fill_doc
 def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
                                symmetric_split=False,
-                               resume=True, verbose=1):
+                               resume=True, verbose=1, clean_labels=None):
     """Load Harvard-Oxford parcellations from FSL.
 
     This function downloads Harvard Oxford atlas packaged from FSL 5.0
@@ -288,7 +288,7 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
 
     Parameters
     ----------
-    atlas_name : string
+    atlas_name : :obj:`str`
         Name of atlas to load. Can be:
         cort-maxprob-thr0-1mm, cort-maxprob-thr0-2mm,
         cort-maxprob-thr25-1mm, cort-maxprob-thr25-2mm,
@@ -303,7 +303,7 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
         sub-maxprob-thr50-1mm, sub-maxprob-thr50-2mm,
         sub-prob-1mm, sub-prob-2mm
 
-    data_dir : string, optional
+    data_dir : :obj:`str`, optional
         Path of data directory where data will be stored. Optionally,
         it can also be a FSL installation directory (which is dependent
         on your installation).
@@ -312,8 +312,8 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
         from your installed directory. Since we mimic same root directory
         as FSL to load it easily from your installation.
 
-    symmetric_split : bool, optional
-        If True, lateralized atlases of cort or sub with maxprob will be
+    symmetric_split : :obj:`bool`, optional
+        If ``True``, lateralized atlases of cort or sub with maxprob will be
         returned. For subcortical types (sub-maxprob), we split every
         symmetric region in left and right parts. Effectively doubles the
         number of regions.
@@ -324,17 +324,23 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
         Default=False.
     %(resume)s
     %(verbose)s
+    clean_labels : :obj:`bool`, optional
+        If set to ``True``, labels which do not appear in the image will
+        be removed from the list of labels.
+        Default=False.
+
+        .. versionadded:: 0.8.2
 
     Returns
     -------
-    data : sklearn.datasets.base.Bunch
+    data : :func:`sklearn.utils.Bunch`
         Dictionary-like object, keys are:
 
-        - "maps": nibabel.Nifti1Image, 4D maps if a probabilistic atlas is
-          requested and 3D labels if a maximum probabilistic atlas was
-          requested.
-
-        - "labels": string list, labels of the regions in the atlas.
+        - "maps": :class:`~nibabel.nifti1.Nifti1Image`, 4D maps if a
+          probabilistic atlas is requested, and 3D labels if a maximum
+          probabilistic atlas was requested.
+        - "labels": :obj:`list` of :obj:`str`, labels of the regions
+          in the atlas.
 
     See also
     --------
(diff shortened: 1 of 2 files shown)


bthirion:
> If this is an issue, I think we will have to re-label the image. WDYT?

I'd rather not. Imagine that you dot it on data from different resolution, then it becomes hard to figure out what is what.
Many atlases do not have consecutive labels, so I don't think it's a big issue.
Best,

jeromedockes:
but the labels are a sequence, and values in the image are supposed to correspond to indices in that sequence, no? so we cannot just delete entries. of course everything would be easier if labels were provided as an {int value: str label name} mapping

bthirion:
Sorry, I didn't check, but shouldn't we handle the labels as a dictionary instead ? You mean that this would require us to do some more work on this atlas ?

jeromedockes:
> Sorry, I didn't check, but shouldn't we handle the labels as a dictionary instead ?
> You mean that this would require us to do some more work on this atlas ?

I mean that at the moment, in all atlases, labels are provided as a list
of label names, and integers in the map correspond to indices in this
list so we cannot change this list without changing the image values
(btw the documentation is quite ambiguous):

```python

ho = fetch_atlas_harvard_oxford("cortl-maxprob-thr0-2mm")
print(type(ho.labels))
print(ho.labels[:4])
```

prints
```
<class 'list'>
['Background', 'Left Frontal Pole', 'Right Frontal Pole', 'Left Insular Cortex']
```

and
```python
view_img(
    image.new_img_like(ho["maps"], image.get_data(ho["maps"]) == 3),
    title=ho.labels[3],
).open_in_browser()
```
shows the left insular cortex.

> but shouldn't we handle the labels as a dictionary instead ? You mean that this would require us to do some more work on this atlas ?

I completely agree it would be nicer if the labels were a dictionary
instead. However that would break backwards compatibility so we have to
keep the list.

What we can do is add new attributes like `region_id_to_label_name` and
`label_name_to_region_id` (or shorter names if we can find some),
containing the explicit mappings from values in the map to string label
names and vice-versa

nilearn/datasets/atlas.py
```diff
@@ -369,14 +380,22 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
         resume=resume,
         verbose=verbose)
     atlas_niimg = check_niimg(atlas_img)
+    label_idx_to_names = {k: v for k, v in enumerate(names)}
     if not symmetric_split or is_lateralized:
+        if not is_probabilistic and clean_labels:
+            valid_idx = np.unique(get_data(atlas_niimg))
+            names = [label_idx_to_names[i] for i in valid_idx]
```
jeromedockes, line 387: if we do this we break the association between image values and labels that the region where image == i corresponds to labels[i], no?

NicolasGensollen: @jeromedockes Yes this is true. The documentation doesn't say clearly that integers from the image are supposed to be interpreted as indices in the list of labels though. Unless I am missing something, this is not the case for all atlases, right? Some have values that are not even consecutive as @bthirion mentioned earlier.

NicolasGensollen: I also agree that using dictionaries for such mappings would be much easier.

jeromedockes:
> @jeromedockes Yes this is true. The documentation doesn't say clearly
> that integers from the image are supposed to be interpreted as indices
> in the list of labels though. Unless I am missing something, this is
> not the case for all atlases, right?

Which ones? and then how do you find the label corresponding to a value
in the image?

> Some have values that are not even consecutive as @bthirion mentioned earlier.

that could be the case while still having image values correspond to
indices in a label list simply if there empty regions

bthirion:
For instance, with aal, indices go from 2001 to 9170 and are non-consecutive. Do you mean that this is an exception and that we should rather see indices as sequences ?

jeromedockes:
> For instance, with aal, indices go from 2001 to 9170 and are non-consecutive. Do you mean that this is an exception and that we should rather see indices as sequences ? 

yes `fetch_atlas_aal` seems to be an exception, and its return value has
an (undocumented) `indices` attribute that is a list of numbers of the
same length as `labels`, giving the numerical (image) values
corresponding to each string label name in `labels`.

Other atlas fetchers I have checked don't have this `indices` attribute,
and it seems that the first item of `labels` (usually "background")
corresponds to voxels with value 0 in the image, the second item of
`labels` corresponds to voxels with value 1 in the image, and so on. --
although this isn't documented clearly.

for probabilistic atlases the first item of labels corresponds to the
first image, the second label to the second image, etc. if we consider
the 4d image as a list of 3d maps

BTW I noticed `fetch_atlas_aal` isn't used in any example so it seems to
be a very good candidate for deprecation

NicolasGensollen:
I'll take a closer look tomorrow but it looks like this PR should be closed as for #2036. I might also have to revert the recently merged #3070 which targeted the same issue, but with `fetch_atlas_destrieux_2009`. 
I think we should improve the documentation w.r.t the labels indexing as it seems to be a source of confusion.

jeromedockes:
> I'll take a closer look tomorrow but it looks like this PR should be closed as for #2036. I might also have to revert the recently merged #3070 which targeted the same issue, but with `fetch_atlas_destrieux_2009`. 

It seems that way, so sorry for missing #3070!

> I think we should improve the documentation w.r.t the labels indexing as it seems to be a source of confusion.

yes I completely agree, and maybe add dictionaries which will be more
helpful for users and keep the lists of labels for backwards
compatibility only

jeromedockes:
> as for #2036.

indeed most problems reported in #2036 seem to be due to the fact that empty regions cause confusion, and the interface and documentation should be improved. 

but https://github.com/nilearn/nilearn/issues/2036#issuecomment-975007816 points to a bug in `generate_report`, so we may want to open a separate issue about that

NicolasGensollen:
>It seems that way, so sorry for missing #3070!

No worries! I merged it too fast without requesting enough reviews. :sweat: 
I'll open a new PR to undo #3070 and will then close this PR as well as #2036 since you opened a more appropriate issue for reports generation.

jeromedockes:
> I'll open a new PR to undo #3070 and will then close this PR as well as #2036 since you opened a more appropriate issue for reports generation.

sounds good, we can discuss all these things during the office hours today!

PR #3083, opened 2021-12-09
[DOC] Improve documentation atlas fetchers

NicolasGensollen:
This PR proposes to improve the documentation for atlas fetchers, especially around their return values.

### DIFF
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -1748,8 +1748,28 @@ def fetch_atlas_schaefer_2018(n_rois=400, yeo_networks=7, resolution_mm=1,
             - 'labels': :class:`numpy.ndarray` of :obj:`str`, array
               containing the ROI labels including Yeo-network annotation.
 
-                .. note::
-                    The list does not have 'Background' by default.
+                .. warning::
+                    The list of labels does not contain 'Background' by
+                    default. To have proper indexing, you should either
+                    manually add 'Background' to the list of labels:
+
+                    .. code-block:: python
+
+                        # Prepend background label
+                        data.labels = np.insert(data.labels, 0, 'Background')
+
+                    Or be careful that the indexing should be offset by one:
+
+                    .. code-block:: python
+
+                        # Get region ID of label '7Networks_LH_Vis_3' when
+                        # 'Background' was not added to the list of labels:
+                        # idx should be equal to 3:
+                        idx = np.where(
+                            data.labels == b'7Networks_LH_Vis_3'
+                        )[0] + 1
+
+
 
             - 'description': :obj:`str`, short description of the atlas
               and some references.
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -437,7 +437,7 @@ def fetch_atlas_juelich(atlas_name, data_dir=None,
         number of regions.
 
         .. note::
-            Not implemented for full probabilistic atlas (*-prob-* atlases).
+            Not implemented for full probabilistic atlas (``*-prob-*`` atlases).
 
         Default=False.
     %(resume)s
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -226,8 +226,14 @@ def fetch_atlas_destrieux_2009(lateralized=True, data_dir=None, url=None,
     data : :func:`sklearn.utils.Bunch`
         Dictionary-like object, contains:
 
-        - Cortical ROIs, lateralized or not (maps)
-        - Labels of the ROIs (labels)
+            - 'maps': :obj:`str`, path to nifti file containing the
+              :class:`~nibabel.nifti1.Nifti1Image` defining the cortical
+              ROIs, lateralized or not. The image has shape ``(76, 93, 76)``,
+              and contains integer values which can be interpreted as the
+              indices in the list of labels.
+            - 'labels': :class:`numpy.recarray`, rec array containing the
+              names of the ROIs.
+            - 'description': :obj:`str`, description of the atlas.
 
     References
     ----------
@@ -274,35 +280,43 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
     specified by your FSL installed path given in `data_dir` argument.
     See documentation for details.
 
+    .. note::
+
+        For atlases 'cort-prob-1mm', 'cort-prob-2mm', 'cortl-prob-1mm',
+        'cortl-prob-2mm', 'sub-prob-1mm', and 'sub-prob-2mm', the returned
+        atlas is probabilistic and the :class:`~nibabel.nifti1.Nifti1Image`
+        returned is 4D, with shape ``(182, 218, 182, 48)``.
+        For deterministic atlases, the :class:`~nibabel.nifti1.Nifti1Image`
+        returned is 3D, with shape ``(182, 218, 182)`` and 48 regions
+        (+ background).
+
     Parameters
     ----------
-    atlas_name : string
+    atlas_name : :obj:`str`
         Name of atlas to load. Can be:
-        cort-maxprob-thr0-1mm, cort-maxprob-thr0-2mm,
-        cort-maxprob-thr25-1mm, cort-maxprob-thr25-2mm,
-        cort-maxprob-thr50-1mm, cort-maxprob-thr50-2mm,
-        cort-prob-1mm, cort-prob-2mm,
-        cortl-maxprob-thr0-1mm, cortl-maxprob-thr0-2mm,
-        cortl-maxprob-thr25-1mm, cortl-maxprob-thr25-2mm,
-        cortl-maxprob-thr50-1mm, cortl-maxprob-thr50-2mm,
-        cortl-prob-1mm, cortl-prob-2mm,
-        sub-maxprob-thr0-1mm, sub-maxprob-thr0-2mm,
-        sub-maxprob-thr25-1mm, sub-maxprob-thr25-2mm,
-        sub-maxprob-thr50-1mm, sub-maxprob-thr50-2mm,
-        sub-prob-1mm, sub-prob-2mm
-
-    data_dir : string, optional
-        Path of data directory where data will be stored. Optionally,
-        it can also be a FSL installation directory (which is dependent
-        on your installation).
-        Example, if FSL is installed in /usr/share/fsl/ then
-        specifying as '/usr/share/' can get you Harvard Oxford atlas
-        from your installed directory. Since we mimic same root directory
+        "cort-maxprob-thr0-1mm", "cort-maxprob-thr0-2mm",
+        "cort-maxprob-thr25-1mm", "cort-maxprob-thr25-2mm",
+        "cort-maxprob-thr50-1mm", "cort-maxprob-thr50-2mm",
+        "cort-prob-1mm", "cort-prob-2mm",
+        "cortl-maxprob-thr0-1mm", "cortl-maxprob-thr0-2mm",
+        "cortl-maxprob-thr25-1mm", "cortl-maxprob-thr25-2mm",
+        "cortl-maxprob-thr50-1mm", "cortl-maxprob-thr50-2mm",
+        "cortl-prob-1mm", "cortl-prob-2mm",
+        "sub-maxprob-thr0-1mm", "sub-maxprob-thr0-2mm",
+        "sub-maxprob-thr25-1mm", "sub-maxprob-thr25-2mm",
+        "sub-maxprob-thr50-1mm", "sub-maxprob-thr50-2mm",
+        "sub-prob-1mm", "sub-prob-2mm".
+    %(data_dir)s
+        Optionally, it can also be a FSL installation directory (which is
+        dependent on your installation).
+        Example, if FSL is installed in ``/usr/share/fsl/`` then
+        specifying as '/usr/share/' can get you the Harvard Oxford atlas
+        from your installed directory. Since we mimic the same root directory
         as FSL to load it easily from your installation.
 
-    symmetric_split : bool, optional
-        If True, lateralized atlases of cort or sub with maxprob will be
-        returned. For subcortical types (sub-maxprob), we split every
+    symmetric_split : :obj:`bool`, optional
+        If ``True``, lateralized atlases of cort or sub with maxprob will be
+        returned. For subcortical types (``sub-maxprob``), we split every
         symmetric region in left and right parts. Effectively doubles the
         number of regions.
 
@@ -315,14 +329,22 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
 
     Returns
     -------
-    data : sklearn.datasets.base.Bunch
+    data : :func:`sklearn.utils.Bunch`
         Dictionary-like object, keys are:
 
-        - "maps": nibabel.Nifti1Image, 4D maps if a probabilistic atlas is
-          requested and 3D labels if a maximum probabilistic atlas was
-          requested.
+            - 'maps': :class:`~nibabel.nifti1.Nifti1Image`, 4D maps if a
+              probabilistic atlas is requested, and 3D labels if a maximum
+              probabilistic atlas was requested.
 
-        - "labels": string list, labels of the regions in the atlas.
+                .. note::
+
+                    For some atlases, it can be the case that some regions
+                    are no longer present. In this case, the integers in the
+                    3D parcellation image might not be consecutive.
+
+            - 'labels': :obj:`list` of :obj:`str`, list of labels for the
+              regions in the atlas.
+            - 'filename': Same as 'maps'.
 
     See also
     --------
@@ -368,6 +390,7 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
     return Bunch(filename=atlas_img, maps=new_atlas_niimg, labels=new_names)
 
 
+@fill_doc
 def fetch_atlas_juelich(atlas_name, data_dir=None,
                         symmetric_split=False,
                         resume=True, verbose=1):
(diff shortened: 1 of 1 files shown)


bthirion (commented):
LGTM so far.

jeromedockes:
thanks for getting started on this! Looks good so far.
- WDYT of adding dict attributes to map image values <-> label names? I think there would be less confusion
- aal is a bit of an outlier and in this case the `indices` attribute provides the mapping
- we should also add documentation for the crucial but undocumented `labels_` attribute of the NiftiLabelsMasker objects, the correct way to find the label corresponding to a masked image dimension `i` is `masker.labels[masker.labels_[i]]`, this is not documented anywhere and is not very convenient -- I'll open a separate issue

nilearn/datasets/atlas.py
```diff
@@ -59,12 +59,18 @@ def fetch_atlas_difumo(dimension=64, resolution_mm=2, data_dir=None, resume=True
...
-        - 'labels': Numpy recarray containing the labels of the regions.
-        - 'description': str, general description of the dataset.
+        - 'maps': :obj:`str`, 4D path to nifti file containing regions
+          definition. The shape of the image is
+          ``(104, 123, 104, dimension)`` where ``dimension`` is the
+          requested dimension of the atlas.
+        - 'labels': :class:`numpy.recarray` containing the labels of
+          the regions. The length of the label array corresponds to the
+          number of dimensions requested. ``data.labels[i]`` is the label
+          corresponding to maps ``i``.
```
bthirion, line 72:
```suggestion
          corresponding to map ``i``.
```

nilearn/datasets/atlas.py
```diff
@@ -327,14 +357,22 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+            - 'maps': :class:`~nibabel.nifti1.Nifti1Image`, 4D maps if a
+              probabilistic atlas is requested, and 3D labels if a maximum
+              probabilistic atlas was requested.
+
+                .. note::
 
-        - "labels": string list, labels of the regions in the atlas.
+                    For some atlases, it can be the case that some regions
+                    are no longer present. In this case, the integers in the
+                    3D parcellation image might not be consecutive.
```
bthirion, line 371: But we're heading toward another heuristic, right ?

bthirion (commented):
LGTM !

nilearn/datasets/atlas.py
NicolasGensollen: I'm not really sure which heuristic you are referring to. We are indeed planning to add more convenient mappings as dictionaries `label -> integer_value` and `integer_value -> label`, but the image can still have non consecutive integer values when some regions are removed by thresholding or resampling for instance. 
I think that the addition of a small note in the documentation saying that this can happen and that it is not a bug would reassure users facing this situation.

nilearn/datasets/atlas.py
```diff
@@ -59,12 +59,18 @@ def fetch_atlas_difumo(dimension=64, resolution_mm=2, data_dir=None, resume=True
 
     Returns
     -------
-    data : sklearn.datasets.base.Bunch
+    data : :func:`sklearn.utils.Bunch`
```
tsalo, line 62:
```suggestion
    data : :class:`sklearn.utils.Bunch`
```

Or I guess `:obj:` would work.

EDIT: Although, I guess it doesn't really matter if the link works...

nilearn/datasets/atlas.py
```diff
@@ -59,12 +59,18 @@ def fetch_atlas_difumo(dimension=64, resolution_mm=2, data_dir=None, resume=True
...
-        - 'labels': Numpy recarray containing the labels of the regions.
-        - 'description': str, general description of the dataset.
+        - 'maps': :obj:`str`, 4D path to nifti file containing regions
+          definition. The shape of the image is
+          ``(104, 123, 104, dimension)`` where ``dimension`` is the
+          requested dimension of the atlas.
+        - 'labels': :class:`numpy.recarray` containing the labels of
+          the regions. The length of the label array corresponds to the
+          number of dimensions requested. ``data.labels[i]`` is the label
+          corresponding to map ``i``.
```
tsalo, line 72:
```suggestion
          corresponding to volume ``i`` in the 'maps' image.
```

Feel free to take or leave this suggestion. I just thought "map `i`" implied that it was a list of images instead of a 4D image.

nilearn/datasets/atlas.py
```diff
@@ -59,12 +59,18 @@ def fetch_atlas_difumo(dimension=64, resolution_mm=2, data_dir=None, resume=True
...
     Returns
     -------
-    data : sklearn.datasets.base.Bunch
+    data : :func:`sklearn.utils.Bunch`
         Dictionary-like object, the interest attributes are :
 
-        - 'maps': str, 4D path to nifti file containing regions definition.
-        - 'labels': Numpy recarray containing the labels of the regions.
-        - 'description': str, general description of the dataset.
+        - 'maps': :obj:`str`, 4D path to nifti file containing regions
```
tsalo, line 65:
```suggestion
        - 'maps': :obj:`str`, path to 4D nifti file containing regions
```

nilearn/datasets/atlas.py
```diff
@@ -327,14 +357,22 @@ def fetch_atlas_harvard_oxford(atlas_name, data_dir=None,
...
+                .. note::
 
-        - "labels": string list, labels of the regions in the atlas.
+                    For some atlases, it can be the case that some regions
+                    are no longer present. In this case, the integers in the
+                    3D parcellation image might not be consecutive.
+
+            - 'labels': :obj:`list` of :obj:`str`, list of labels for the
+              regions in the atlas.
+            - 'filename': Same as 'maps'.
```
tsalo, line 375: If "maps" is a niimg instead of the path to the file, then this shouldn't be the same, right?

nilearn/datasets/atlas.py
```diff
@@ -394,49 +433,62 @@ def fetch_atlas_juelich(atlas_name, data_dir=None,
...
         symmetric region in left and right parts. Effectively doubles the
         number of regions.
-        NOTE Not implemented for full probabilistic atlas (*-prob-* atlases).
-        Default=False.
 
-    resume : bool, optional
-        Whether to resumed download of a partly-downloaded file.
-        Default=True.
+        .. note::
+            Not implemented for full probabilistic atlas (*-prob-* atlases).
```
tsalo, line 468:
```suggestion
            Not implemented for full probabilistic atlas (``*-prob-*`` atlases).
```

nilearn/datasets/atlas.py
```diff
@@ -889,10 +975,25 @@ def fetch_atlas_aal(version='SPM12', data_dir=None, url=None, resume=True,
...
+
+        The maps image (``data.maps``) contains 117 unique integer values
+        defining the parcellation. However, these values are not consecutive
+        integers from 0 to 116 as is usually the case in Nilearn.
+        Therefore, these values shouldn't be interpreted as indices for the
+        list of label names.
+        For example, to get the name of the region corresponding to label
+        5021 in the image, you should do:
+
+        .. code-block::python
```
tsalo, line 988:
```suggestion
        .. code-block:: python
```

The code block doesn't show up (see below).

Screen Shot 2021-12-13 at 11 40 07 AM

nilearn/datasets/atlas.py
```diff
@@ -901,12 +1002,17 @@ def fetch_atlas_aal(version='SPM12', data_dir=None, url=None, resume=True,
...
-        - "maps": str. path to nifti file containing regions.
-
-        - "labels": list of the names of the regions
+            - 'maps': :obj:`str`, path to nifti file containing the
+              regions. The image has shape ``(91, 109, 91)`` and contains
+              117 unique integer values defining the parcellation.
+            - 'labels': :obj:`list` of :obj:`str`, list of the names of the
+              regions. This list has 116 names as 'Background' (label 0) is
+              not included in this list.
+            - 'indices': :obj:`list` of :obj:`str`, ???
```
tsalo, line 1014:
```suggestion
            - 'indices': :obj:`list` of :obj:`str`, indices mapping 'labels' to values in the 'maps' image.
              Since the values in the 'maps' image are not consecutive, the location of a label in the
              'labels' list does not necessary match the associated value in the image.
              Use the 'indices' list to identify the appropriate image value for a given label.
```

Or something like that. I just based this on the warning above.

nilearn/datasets/atlas.py
```diff
@@ -1059,19 +1167,23 @@ def fetch_coords_dosenbach_2010(ordered_regions=True):
...
+    data : :func:`sklearn.utils.Bunch`
         Dictionary-like object, contains:
 
-        - "rois": coordinates of 160 ROIs in MNI space
-        - "labels": ROIs labels
-        - "networks": networks names
+        - 'rois': :class:`numpy.recarray`, rec array with the coordinates
+          of the 160 ROIs in :term:`MNI` space.
+        - 'labels': :class:`numpy.ndarray` of :obj:`str`, list of label
+          names for the 160 ROI.
```
tsalo, line 1183:
```suggestion
          names for the 160 ROIs.
```

nilearn/datasets/atlas.py
```diff
@@ -1059,19 +1167,23 @@ def fetch_coords_dosenbach_2010(ordered_regions=True):
...
-    data : sklearn.datasets.base.Bunch
+    data : :func:`sklearn.utils.Bunch`
         Dictionary-like object, contains:
 
-        - "rois": coordinates of 160 ROIs in MNI space
-        - "labels": ROIs labels
-        - "networks": networks names
+        - 'rois': :class:`numpy.recarray`, rec array with the coordinates
+          of the 160 ROIs in :term:`MNI` space.
+        - 'labels': :class:`numpy.ndarray` of :obj:`str`, list of label
```
tsalo, line 1182: Based on other atlases, it seems like this should be a list. Maybe something worth opening a separate issue for?

nilearn/datasets/atlas.py
```diff
@@ -1545,32 +1696,41 @@ def fetch_atlas_schaefer_2018(n_rois=400, yeo_networks=7, resolution_mm=1,
...
 
-        - maps: 3D Nifti image, values are indices in the list of labels.
-        - labels: ROI labels including Yeo-network annotation,list of strings.
-        - description: A short description of the atlas and some references.
+            - 'maps': :obj:`str`, path to nifti file containing the
+              3D :class:`~nibabel.nifti1.Nifti1Image` (its shape is
+              ``(182, 218, 182)``). The values are consecutive integers
+              between 0 and ``n_rois`` which can be interpreted as indices
+              in the list of labels.
+            - 'labels': :class:`numpy.ndarray`, array containing the ROI
```
tsalo, line 1726:
```suggestion
            - 'labels': :class:`numpy.ndarray` of :obj:`str`, array containing the ROI
```

tsalo (changes requested):
I just have some minor notes, but this looks great overall.

nilearn/datasets/atlas.py
NicolasGensollen: Using `:class:` doesn't seem to work. After looking at https://github.com/scikit-learn/scikit-learn/blob/0d378913be6d7e485b792ea36e9268be31ed52d0/doc/modules/classes.rst, I used `:func:` instead to make it work. I think it might be a small bug in scikit-learn's doc.

nilearn/datasets/atlas.py
NicolasGensollen: Absolutely! I opened #3088 for this specific issue.

nilearn/datasets/atlas.py
NicolasGensollen: Thanks! :+1:

nilearn/datasets/atlas.py
tsalo: Oh... thanks for clarifying. Really glad I didn't comment on every line where it was used!

tsalo (approved):
Thanks. It looks good to me now.

bthirion (approved):
LGTM, thx !

nilearn/datasets/atlas.py
```diff
@@ -821,18 +894,32 @@ def fetch_atlas_yeo_2011(data_dir=None, url=None, resume=True, verbose=1):
...
+            - 'thick_7': :obj:`str`, path to nifti file containing the
+              7 region parcellation fitted to thick template cortex
+              segmentations.
+            - 'thin_17': :obj:`str`, path to nifti file containing the
+              17 region parcellation fitted to thin template cortex
+              segmentations.
+            - 'thick_17': :obj:`str`, path to nifti file containing the
+              17 region parcellation fitted to thick template cortex
+              segmentations.
+            - 'colors_7': :obj:`str`, path to colormaps text file for
```
jeromedockes, line 912: here also could we say values in the image are indices into this list?

nilearn/datasets/atlas.py
```diff
@@ -889,10 +976,25 @@ def fetch_atlas_aal(version='SPM12', data_dir=None, url=None, resume=True,
...
     and :footcite:`TZOURIOMAZOYER2002273`.
 
+    .. warning::
+
+        The maps image (``data.maps``) contains 117 unique integer values
+        defining the parcellation. However, these values are not consecutive
+        integers from 0 to 116 as is usually the case in Nilearn.
+        Therefore, these values shouldn't be interpreted as indices for the
+        list of label names.
+        For example, to get the name of the region corresponding to label
```
jeromedockes, line 986: Here also can we reserve "label" for strings/ region names and use "Region IDs" (or anything else you can think of) for integers in the map?

nilearn/datasets/atlas.py
```diff
@@ -889,10 +976,25 @@ def fetch_atlas_aal(version='SPM12', data_dir=None, url=None, resume=True,
...
+        5021 in the image, you should do:
+
+        .. code-block:: python
+
+            # This should print 'Lingual_L'
+            data.labels[data.indices.index('5021')]
```
jeromedockes, line 992: I would maybe highlight the fact that the region ids are provided as strings in `indices` but are ints in the image. and could we add an example going in the other direction (find the value to look for in the map if we want region `"Precentral_L"`?

nilearn/datasets/atlas.py
```diff
@@ -901,12 +1003,22 @@ def fetch_atlas_aal(version='SPM12', data_dir=None, url=None, resume=True,
...
     Returns
     -------
-    data : sklearn.datasets.base.Bunch
+    data : :func:`sklearn.utils.Bunch`
         Dictionary-like object, keys are:
 
-        - "maps": str. path to nifti file containing regions.
-
-        - "labels": list of the names of the regions
+            - 'maps': :obj:`str`, path to nifti file containing the
```
jeromedockes, line 1009: maybe the warning and examples could come here, or we can point back to them?

nilearn/datasets/atlas.py
```diff
@@ -901,12 +1003,22 @@ def fetch_atlas_aal(version='SPM12', data_dir=None, url=None, resume=True,
...
+              regions. The image has shape ``(91, 109, 91)`` and contains
+              117 unique integer values defining the parcellation.
+            - 'labels': :obj:`list` of :obj:`str`, list of the names of the
+              regions. This list has 116 names as 'Background' (label 0) is
+              not included in this list.
+            - 'indices': :obj:`list` of :obj:`str`, indices mapping 'labels'
```
jeromedockes, line 1015: can you provide the length of `indices` (116) ?

nilearn/datasets/atlas.py
```diff
@@ -901,12 +1003,22 @@ def fetch_atlas_aal(version='SPM12', data_dir=None, url=None, resume=True,
...
+              to values in the 'maps' image.
+              Since the values in the 'maps' image are not consecutive, the
```
jeromedockes, line 1017: here also rather than are not consecutive, "do _not_ correspond to indices in `labels`, but rather to values in `indices`"

nilearn/datasets/atlas.py
```diff
@@ -1191,16 +1315,31 @@ def fetch_atlas_allen_2011(data_dir=None, url=None, resume=True, verbose=1):
...
+        - 'rsn_indices': :obj:`list` of :obj:`tuple`, each tuple is a
+          (:obj:`str`, :obj:`list` of :`int`). This maps the network names
+          to the map indices. For example, the map indices for the 'Visual'
+          network can be obtained:
+
+            .. code-block:: python
+
+                idx = [x[0] for x in data.rsn_indices].index('Visual')
+                # Should return [46, 64, 67, 48, 39, 59]
+                data.rsn_indices[idx][1]
```
jeromedockes, line 1338: maybe simpler:
```python
dict(data.rsn_indices)["Visual"]
```

nilearn/datasets/atlas.py
```diff
@@ -1431,20 +1570,25 @@ def fetch_atlas_talairach(level_name, data_dir=None, verbose=1):
...
 
-        - maps: 3D Nifti image, values are indices in the list of labels.
-        - labels: list of strings. Starts with 'Background'.
-        - description: a short description of the atlas and some references.
+            - 'maps': 3D :class:`~nibabel.nifti1.Nifti1Image`, image has
+              shape ``(141, 172, 110)`` and contains consecutive integer
+              values from 0 to the number of regions, which are indices
+              in the list of labels.
+            - 'labels': :obj:`list` of :obj:`str`. List of region names.
+              The list starts with 'Background' (label 0 in the image).
```
jeromedockes, line 1589: here also, reserve the term "label" to refer to strings

nilearn/datasets/atlas.py
```diff
@@ -1484,12 +1628,25 @@ def fetch_atlas_pauli_2017(version='prob', data_dir=None, verbose=1):
...
+              :class:`~nibabel.nifti1.Nifti1Image`. If ``version='prob'``,
+              the image shape is ``(193, 229, 193, 16)``. If ``version='det'``
+              the image shape is ``(198, 263, 212)``, and values are indices
+              in the list of labels (integers from 0 to 16).
+            - 'labels': :obj:`list` of :obj:`str`. List of region names. The
+              list contains 16 values for both probabilitic and deterministic
+              versions.
+
+                .. note::
+                    For the deterministic version, 'Background' is not
```
jeromedockes, line 1644: isn't it for the probabilistic version? because there is no probabilistic map of the background

jeromedockes: I see, I understand now -- "background" is never included in the list of labels, and that is a mistake only in the deterministic version.

so the image values are actually not indices into the list of labels because they are offset by 1, right?

NicolasGensollen: >so the image values are actually not indices into the list of labels because they are offset by 1, right?

Yes, absolutely. Thanks for raising this. :+1: 
I've updated this part of the docs in 5521855 
Hopefully it is clearer like this.

nilearn/datasets/atlas.py
```diff
@@ -231,8 +247,14 @@ def fetch_atlas_destrieux_2009(lateralized=True, data_dir=None, url=None,
     data : :func:`sklearn.utils.Bunch`
         Dictionary-like object, contains:
 
-        - Cortical ROIs, lateralized or not (maps)
-        - Labels of the ROIs (labels)
+            - 'maps': :obj:`str`, path to nifti file containing the
+              :class:`~nibabel.nifti1.Nifti1Image` defining the cortical
+              ROIs, lateralized or not. The image has shape ``(76, 93, 76)``,
+              and contains integer values which can be interpreted as the
+              indices in the list of labels.
```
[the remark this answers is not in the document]
jeromedockes: btw for this fetcher once you rebase or merge main you may get conflicts or need to update the docstring, as the `clean_labels` parameters has been removed?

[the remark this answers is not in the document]
NicolasGensollen: I'm going to rebase now and see what happens

NicolasGensollen:
@jeromedockes Do you think this can be merged or do you want to take another look at it?

nilearn/datasets/atlas.py
```diff
@@ -1517,32 +1718,41 @@ def fetch_atlas_schaefer_2018(n_rois=400, yeo_networks=7, resolution_mm=1,
...
+            - 'maps': :obj:`str`, path to nifti file containing the
+              3D :class:`~nibabel.nifti1.Nifti1Image` (its shape is
+              ``(182, 218, 182)``). The values are consecutive integers
+              between 0 and ``n_rois`` which can be interpreted as indices
+              in the list of labels.
+            - 'labels': :class:`numpy.ndarray` of :obj:`str`, array
+              containing the ROI labels including Yeo-network annotation.
+
+                .. note::
+                    The list does not have 'Background' by default.
```
jeromedockes, line 1752: so here as well indices into the list of labels are actually offset by 1 ?

jeromedockes (approved):
thanks a lot @NicolasGensollen . LGTM besides one small question

nilearn/datasets/atlas.py
NicolasGensollen: Good catch! Thanks @jeromedockes !
I improved the doc in 4a7560a

NicolasGensollen:
Thanks for the reviews!
Merging....

PR #3090, opened 2021-12-10
[MAINT] Remove `compute_gray_matter_mask`

NicolasGensollen:
Remove function `compute_gray_matter_mask` (`compute_brain_mask` should be used instead) as its deprecation cycle has reached its end.


bthirion (approved):
LGTM, thx.

NicolasGensollen:
Thanks! Merging...

PR #3094, opened 2021-12-14
BUG: Import from collections.abc for 3.10

larsoner:
I recently moved to Python 3.10 and found that `collections.Mapping` no longer exists in that namespace, but rather `collections.abc.Mapping`. This existed all the way back to 3.6 so should be okay.

Closes #3095

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -63,7 +63,7 @@ jobs:
         fail-fast: false
         matrix:
             os: ["ubuntu-latest", "macos-latest"]
-            python-version: ["3.7", "3.8", "3.9"]
+            python-version: ["3.7", "3.8", "3.9", "3.10"]
     name: ${{ matrix.os }} with Python ${{ matrix.python-version }} and latest package versions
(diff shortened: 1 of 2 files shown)


NicolasGensollen (approved):
Thanks @larsoner !
Might be time to consider supporting Python 3.10 officially.

larsoner:
> Might be time to consider supporting Python 3.10 officially.

I couldn't resist the temptation to add 3.10 to the GitHub Actions test matrix. I'll get rid of it if it fails for some non-trivial reason -- or if you want to adjust CIs separately I'm happy to remove it!

NicolasGensollen:
>I couldn't resist the temptation to add 3.10 to the GitHub Actions test matrix.

Yes, adding python versions to the CI is highly addictive ! :smiley: 

Let's see if it works here thanks to your fix, and I'll close #3095

larsoner:
Green!

bthirion (approved):
Thx !

NicolasGensollen (approved):

PR #3102, opened 2021-12-16
[ENH] `NiftiSpheresMasker` reports

NicolasGensollen:
See #2689 

This is still very WIP, but this PR proposes a possible implementation of reports for `NiftiSpheresMasker`.

The reports combine parts of what was done previously  for the `NiftiLabelsMasker` (summary table per regions/spheres), and parts of what was done for the `NiftiMapsMasker` (browsing through selected maps/spheres). The reports also show all spheres on a glass brain template as the first image.

As for most other maskers, `fit()` doesn't do much, and the only information that you can potentially have after fitting is the coordinates of the seeds (which is provided to the constructor anyway...).

When no image is provided to `fit()`, the report still displays all spheres on a glass brain template, but when browsing the selected spheres, they are plotted over the MNI152 template:

```python
import numpy as np
from nilearn.input_data import NiftiSpheresMasker
from nilearn.datasets import fetch_coords_dosenbach_2010, fetch_development_fmri

dosenbach = fetch_coords_dosenbach_2010()
data = fetch_development_fmri(n_subjects=1)
func = data.func[0]

coords = np.vstack((
     dosenbach.rois['x'],
     dosenbach.rois['y'],
     dosenbach.rois['z'],
)).T
 
masker = NiftiSpheresMasker(coords, radius=8)
masker.fit()  # No img provided to fit
# To experiment with images, comment previous line and uncomment next one
# masker.fit(func)
masker.generate_report([2, 10, 19, 26, 100])  # Only browse these specific spheres
```
spheres1

Click Next button:
spheres2

If an image is provided to `fit()`, then we use this image (or the average image if it is 4D):

spheres3


nilearn/input_data/nifti_spheres_masker.py
```diff
@@ -310,8 +315,180 @@ def __init__(self, seeds, radius=None, mask_img=None, allow_overlap=False,
...
+        self.report_id = -1
+        self._report_content = dict()
+        self._report_content['description'] = (
+            'This reports shows the regions '
+            'defined by the spheres of the masker.')
+        self._report_content['warning_message'] = None
+
         self.verbose = verbose
 
+    def generate_report(self, displayed_spheres=10):
```
bthirion, line 329: By contrast, I find that setting displayed _spheres to 10 by default is really arbitrary and surprising. I would default to 'all'

ymzayek:
I'll pick this one up after finishing the #3935 and #4099

bthirion:
Please LMK if you need a review.

nilearn/maskers/nifti_spheres_masker.py
```diff
@@ -318,8 +322,183 @@ def __init__(
         self.memory = memory
         self.memory_level = memory_level
 
+        # Parameters for reporting
+        self.reports = reports
+        self.report_id = -1
```
bthirion, line 327: ??

nilearn/maskers/nifti_spheres_masker.py
```diff
@@ -318,8 +322,183 @@ def __init__(
...
+                return [None]
+
+        if self._reporting_data is not None:
+            seeds = self._reporting_data['seeds']
+        else:
+            self._report_content['summary'] = None
+
+            return [None]
+
+        from nilearn.image import coord_transform
```
bthirion, line 421: Normally we don't have imports in the middle of methods

nilearn/maskers/nifti_spheres_masker.py
```diff
@@ -318,8 +322,183 @@ def __init__(
...
+        embeded_images.append(_embed_img(display))
+        display.close()
+        for idx, seed in enumerate(seeds):
+            regions_summary['seed number'].append(idx)
+            regions_summary['coordinates'].append(str(seed))
+            regions_summary['position'].append(positions[idx])
+            regions_summary['radius'].append(radius)
+            regions_summary['size (in voxels)'].append('not implemented')
+            regions_summary['size (in mm^3)'].append(round(
+                4. / 3. * np.pi * radius, 2))
```
bthirion, line 485:
```suggestion
                4. / 3. * np.pi * radius**3, 2))
```

nilearn/maskers/nifti_spheres_masker.py
```diff
@@ -318,8 +322,183 @@ def __init__(
...
+
+        Returns
+        -------
+        displays : list
+            A list of all displays to be rendered.
+        """
+        from nilearn.reporting.html_report import _embed_img
+        try:
+            from nilearn import plotting
+        except ImportError:
```
bthirion, line 405: A test is needed here

bthirion (commented):
Looks good. Please find a few comments below.

nilearn/maskers/nifti_spheres_masker.py
ymzayek: Now report id is set in the report content and will be incremented every time generate_report is called. This was adapted directly from nifti_maps_masker

ymzayek:
TODO: 
Fix test (then coverage should pass too)
Add whatsnew

bthirion (commented):
The PR LGTM.
Isn't there to showcase this report in an example ?

ymzayek:
Here it is in an example: https://output.circle-artifacts.com/output/job/0dbd134a-c674-47fe-8a56-8cdc3bd4daaf/artifacts/0/dev/auto_examples/03_connectivity/plot_sphere_based_connectome.html#display-spheres-summary-report

But I need to check something as the positions of the spheres when viewed individually seem to be different from when they are plotted all together on the MNI template (first image)

bthirion (approved):

Remi-Gau (changes requested):
I think we should probably add tests for a better coverage, no?

I can take care of it if you want.

ymzayek:
> I think we should probably add tests for a better coverage, no?
> 
> I can take care of it if you want.

Yea I agree, if you want by all means go for it :) Otherwise I can get to it tomorrow maybe

Remi-Gau:
OK will add a few then: should help me get a better understanding of the new code too.

ymzayek:
Note that the patch report in FIles changed tab does not seem to be accurate. I don't know why
I am mostly worried about this https://github.com/nilearn/nilearn/pull/3102#issuecomment-1840723054

Remi-Gau:
Checking from phone so unless the UI
is tricking me but seems the coverage report look happier now

Remi-Gau (approved):
Should be good to go.

bthirion:
Merging then.

PR #3105, opened 2021-12-17
[FIX] Off-by-one error while setting up ticks

### LINKED ISSUE #3104: `_get_ticks_matplotlib()` is incorrect for integers with a small range
https://github.com/nilearn/nilearn/blob/6aaf8f94433dfaf2967d69c64c0f688d7e1c812d/nilearn/plotting/surf_plotting.py#L341-L350

So a "small range" appears to be defined as less than 5.

For larger ranges, we always get 5 ticks. For example:
```python
>>> ticks = _get_ticks_matplotlib(2.01, 7.02, "%i")
>>> 
>>> ticks
array([2.01  , 3.2625, 4.515 , 5.7675, 7.02  ])
>>> 
>>> len(ticks)
5
>>> 
```

For smaller ranges, we may get **5 ticks or more** (instead of **less** as documented):
```python
>>> ticks = _get_ticks_matplotlib(2.01, 6.99, "%i")
>>> 
>>> ticks
array([2.01, 3.01, 4.01, 5.01, 6.01, 7.01])
>>> 
>>> len(ticks)
6
>>> 
>>> 
>>> ticks = _get_ticks_matplotlib(2.01, 5.02, "%i")
>>> 
>>> ticks
array([2.01, 3.01, 4.01, 5.01, 6.01])
>>> 
>>> len(ticks)
5
>>> 
```

DimitriPapadopoulos:
Also remove unused local variable `n_ticks`, reported by LGTM.com.

Closes #3104.

### DIFF
--- a/nilearn/plotting/surf_plotting.py
+++ b/nilearn/plotting/surf_plotting.py
@@ -341,11 +341,11 @@ def _get_ticks_matplotlib(vmin, vmax, cbar_tick_format):
     n_ticks = 5
     # ...unless we are dealing with integers with a small range
     # in this case, we reduce the number of ticks
-    if cbar_tick_format == "%i" and vmax - vmin < n_ticks:
+    if cbar_tick_format == "%i" and vmax - vmin < n_ticks - 1:
         ticks = np.arange(vmin, vmax + 1)
-        n_ticks = len(ticks)
     else:
-        ticks = np.linspace(vmin, vmax, n_ticks)
+        # remove duplicate ticks when vmin == vmax, or almost
+        ticks = np.unique(np.linspace(vmin, vmax, n_ticks))
     return ticks
 
 
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,6 +13,7 @@ NEW
 Fixes
 -----
 
+- Fix off-by-one error when setting ticks in :func:`~plotting.plot_surf` (:gh:`3105` by `Dimitri Papadopoulos Orfanos`_).
 - Regressor names can now be invalid identifiers but will raise an error with :meth:`~glm.first_level.FirstLevelModel.compute_contrast` if combined to make an invalid expression (:gh:`3374` by `Yasmin Mzayek`_).
 - Fix :func:`~plotting.plot_connectome` which was raising a ``ValueError`` when ``vmax < 0`` (:gh:`3390` by `Paul Bogdan`_).
 
--- a/nilearn/plotting/tests/test_surf_plotting.py
+++ b/nilearn/plotting/tests/test_surf_plotting.py
@@ -12,7 +12,8 @@ import unittest.mock as mock
 from nilearn.plotting.img_plotting import MNI152TEMPLATE
 from nilearn.plotting.surf_plotting import (plot_surf, plot_surf_stat_map,
                                             plot_surf_roi, plot_img_on_surf,
-                                            plot_surf_contours)
+                                            plot_surf_contours,
+                                            _get_ticks_matplotlib)
 from nilearn.datasets import fetch_surf_fsaverage
 from nilearn.surface import load_surf_mesh
 from nilearn.surface.testing_utils import generate_surf
@@ -822,3 +823,23 @@ def test_plot_surf_contours_error():
             ValueError,
             match='Levels, labels, and colors argument need to be either the same length or None.'):
         plot_surf_contours(mesh, parcellation, levels=[1, 2], colors=['r'], labels=['1', '2'])
+
+
+@pytest.mark.parametrize("vmin,vmax,cbar_tick_format,expected", [
+    (0, 0, "%i", [0]),
+    (0, 3, "%i", [0, 1, 2, 3]),
+    (0, 4, "%i", [0, 1, 2, 3, 4]),
+    (1, 5, "%i", [1, 2, 3, 4, 5]),
+    (0, 5, "%i", [0, 1.25, 2.5, 3.75, 5]),
+    (0, 10, "%i", [0, 2.5, 5, 7.5, 10]),
+    (0, 0, "%.1f", [0]),
+    (0, 1, "%.1f", [0, 0.25, 0.5, 0.75, 1]),
+    (1, 2, "%.1f", [1, 1.25, 1.5, 1.75, 2]),
+    (1.1, 1.2, "%.1f", [1.1, 1.125, 1.15, 1.175, 1.2]),
+    (0, np.nextafter(0, 1), "%.1f", [0.e+000, 5.e-324]),
+])
+def test_get_ticks_matplotlib(vmin, vmax, cbar_tick_format, expected):
+    ticks = _get_ticks_matplotlib(vmin, vmax, cbar_tick_format)
+    assert 1 <= len(ticks) <= 5
+    assert ticks[0] == vmin and ticks[-1] == vmax
+    assert len(ticks) == len(expected) and (ticks == expected).all()


DimitriPapadopoulos:
I could add a test, but is it worth it for this small private method?

nilearn/plotting/surf_plotting.py
```diff
@@ -342,9 +342,9 @@ def _get_ticks_matplotlib(vmin, vmax, cbar_tick_format):
     n_ticks = 5
     # ...unless we are dealing with integers with a small range
     # in this case, we reduce the number of ticks
-    if cbar_tick_format == "%i" and vmax - vmin < n_ticks:
+    if cbar_tick_format == "%i" and vmax - vmin < n_ticks - 1:
         ticks = np.arange(vmin, vmax + 1)
-        n_ticks = len(ticks)
+        assert len(ticks) <= n_ticks
```
NicolasGensollen, line 347: I think `ticks` should be set to `np.arange(int(vmin), int(vmax) + 1)` since they should be integers, right?

NicolasGensollen (commented):
Thanks for fixing this @DimitriPapadopoulos

nilearn/plotting/surf_plotting.py
jeromedockes: let's add a test rather than an assertion in the function itself. we don't want to crash the program just because there are too many ticks in a colorbar

DimitriPapadopoulos: Should they be integers? I really don't know. That's not the case in the "large range" case either.

In any case, the proper range would be `[floor(vmin), ceil(vmax)]`.

NicolasGensollen: To me they should be integers because the if condition verifies that `cbar_tick_format=='%i'`.

DimitriPapadopoulos: Agreed, but why is this not the case for larger ranges with `np.linspace(vmin, vmax, n_ticks)`?

NicolasGensollen: Forget my initial comment, I was wrong, we don't need to cast anything here.
Formatting happens there:

https://github.com/nilearn/nilearn/blob/6aaf8f94433dfaf2967d69c64c0f688d7e1c812d/nilearn/plotting/surf_plotting.py#L512-L515

DimitriPapadopoulos: I added a new commit to enforce integer ticks if the format is `"%i"`.

DimitriPapadopoulos: Ah, just read the previous comment. So the last commit is useless and can be removed.

bthirion (commented):
Thx, ideally, we would like to have  test.

DimitriPapadopoulos:
Would a unit test of `_get_ticks_matplotlib()` be enough, or do you require a higher level functional test?

NicolasGensollen:
>Would a unit test of _get_ticks_matplotlib() be enough, or do you require a higher level functional test?

I think unit testing `_get_ticks_matplotlib` would be enough. Ideally, some assertions should fail with current implementation (on master) and pass with this fix.

ymzayek:
@DimitriPapadopoulos I was wondering if you are able to pick this back up and finish it? Looks like it's almost there, just missing a unit test `_get_ticks_matplotlib`. If not, let us know and we'll take it over.

DimitriPapadopoulos:
I **think** the patch is OK as it is, but I need to double-check the logic - writing a test is definitely required. I do not have a computer with me, I might look into it at the end of August.

cr3:
Hi @DimitriPapadopoulos, what would you think about applying this patch to add a unit test for `_get_ticks_matplotlib`:
```
diff --git a/nilearn/plotting/tests/test_surf_plotting.py b/nilearn/plotting/tests/test_surf_plotting.py
index 02f165c2..bb88f0ab 100644
--- a/nilearn/plotting/tests/test_surf_plotting.py
+++ b/nilearn/plotting/tests/test_surf_plotting.py
@@ -12,7 +12,7 @@ import unittest.mock as mock
 from nilearn.plotting.img_plotting import MNI152TEMPLATE
 from nilearn.plotting.surf_plotting import (plot_surf, plot_surf_stat_map,
                                             plot_surf_roi, plot_img_on_surf,
-                                            plot_surf_contours)
+                                            plot_surf_contours, _get_ticks_matplotlib)
 from nilearn.surface import load_surf_mesh
 from nilearn.surface.testing_utils import generate_surf
@@ -822,3 +822,22 @@ def test_plot_surf_contours_error():
             ValueError,
             match='Levels, labels, and colors argument need to be either the same length or None.'):
         plot_surf_contours(mesh, parcellation, levels=[1, 2], colors=['r'], labels=['1', '2'])
+
+
+@pytest.mark.parametrize("vmin,vmax,cbar_tick_format,expected", [
+    (0, 0, "%i", [0]),
+    (0, 3, "%i", [0, 1, 2, 3]),
+    (0, 4, "%i", [0, 1, 2, 3, 4]),
+    (1, 5, "%i", [1, 2, 3, 4, 5]),
+    (0, 5, "%i", [0, 1.25, 2.5, 3.75, 5]),
+    (0, 10, "%i", [0, 2.5, 5, 7.5, 10]),
+    (0, 0, "%.1f", [0, 0, 0, 0, 0]),
+    (0, 1, "%.1f", [0, 0.25, 0.5, 0.75, 1]),
+    (1, 2, "%.1f", [1, 1.25, 1.5, 1.75, 2]),
+    (1.1, 1.2, "%.1f", [1.1, 1.125, 1.15, 1.175, 1.2]),
+])
+def test_get_ticks_matplotlib(vmin, vmax, cbar_tick_format, expected):
+    ticks = _get_ticks_matplotlib(vmin, vmax, cbar_tick_format)
+    assert 1 <= len(ticks) <= 5
+    assert ticks[0] == vmin and ticks[-1] == vmax
+    assert len(ticks) == len(expected) and (ticks == expected).all()
```

nilearn/plotting/surf_plotting.py
```diff
@@ -341,9 +341,8 @@ def _get_ticks_matplotlib(vmin, vmax, cbar_tick_format):
     n_ticks = 5
     # ...unless we are dealing with integers with a small range
     # in this case, we reduce the number of ticks
-    if cbar_tick_format == "%i" and vmax - vmin < n_ticks:
+    if cbar_tick_format == "%i" and vmax - vmin < n_ticks - 1:
         ticks = np.arange(vmin, vmax + 1)
-        n_ticks = len(ticks)
     else:
         ticks = np.linspace(vmin, vmax, n_ticks)
```
DimitriPapadopoulos, line 347: What if `vmin == vmax`? Shouldn't we keep just one single tick, like in the `"%i"` case?

DimitriPapadopoulos: More generally, what if the difference between `vmin` and `vmax` is small enough?
```python
>>> vmin = 100
>>> vmax = np.nextafter(vmin, vmin + 1)
>>> n_ticks = 5
>>> 
>>> np.linspace(vmin, vmax, n_ticks)
array([100., 100., 100., 100., 100.])
>>> 
```
```python
>>> vmin = 0
>>> vmax = np.nextafter(vmin, vmin + 1)
>>> n_ticks = 5
>>> 
>>> np.linspace(vmin, vmax, n_ticks)
array([0.e+000, 0.e+000, 0.e+000, 5.e-324, 5.e-324])
>>> 
```
Of course, it's a corner case. Besides, it would only result in multiple ticks drawn at the exact same place. Do we worry about it?

cr3: It makes sense to keep a single tick when `vmin == vmax`. When calling `_get_ticks_matplotlib(0.1 + 0.2, 0.3, "%f")`, however, note that `vmin == vmax` is `False` so you'll have to do something like `cbar_tick_format % vmin == cbar_tick_format % vmax`:
```
>>> fmt = '%f'
>>> vmin = 0.1 + 0.2
>>> vmax = 0.3
>>> vmin == vmax
False
>>> fmt % vmin == fmt % vmax
True
```
This might partially address your other concern about the difference being small enough where I would posit that "small enough" depends on `cbox_tick_format`.

DimitriPapadopoulos: An alternative would be to remove identical ticks, keep only unique keys:
```python
>>> vmin = 100
>>> vmax = 100
>>> n_ticks = 5
>>> 
>>> np.linspace(vmin, vmax, n_ticks)
array([100., 100., 100., 100., 100.])
>>> 
>>> np.array(list(dict.fromkeys(np.linspace(vmin, vmax, n_ticks))))
array([100.])
>>> 
```
```python
>>> vmin = 0
>>> vmax = np.nextafter(vmin, vmin + 1)
>>> n_ticks = 5
>>> 
>>> np.linspace(vmin, vmax, n_ticks)
array([0.e+000, 0.e+000, 0.e+000, 5.e-324, 5.e-324])
>>> 
>>> np.array(list(dict.fromkeys(np.linspace(vmin, vmax, n_ticks))))
array([0.e+000, 5.e-324])
>>> 
```

bthirion: @DimitriPapadopoulos  sounds reasonable. Thx  to both for the discussion & update !

cr3: @DimitriPapadopoulos nice! I would suggest using `np.unique` to make it easier on the reader:
```
>>> np.unique(np.linspace(vmin, vmax, n_ticks))
array([0.e+000, 5.e-324])
```
What is especially cool is that it fixes one of the unit tests - I also added a test for your use case in the patch below:
```
diff --git a/nilearn/plotting/surf_plotting.py b/nilearn/plotting/surf_plotting.py
index 4ab29eee..90697cca 100644
--- a/nilearn/plotting/surf_plotting.py
+++ b/nilearn/plotting/surf_plotting.py
@@ -344,7 +344,7 @@ def _get_ticks_matplotlib(vmin, vmax, cbar_tick_format):
     if cbar_tick_format == "%i" and vmax - vmin < n_ticks - 1:
         ticks = np.arange(vmin, vmax + 1)
     else:
-        ticks = np.linspace(vmin, vmax, n_ticks)
+        ticks = np.unique(np.linspace(vmin, vmax, n_ticks))
     return ticks


diff --git a/nilearn/plotting/tests/test_surf_plotting.py b/nilearn/plotting/tests/test_surf_plotting.py
index bb88f0ab..23a68520 100644
--- a/nilearn/plotting/tests/test_surf_plotting.py
+++ b/nilearn/plotting/tests/test_surf_plotting.py
@@ -831,10 +831,11 @@ def test_plot_surf_contours_error():
     (1, 5, "%i", [1, 2, 3, 4, 5]),
     (0, 5, "%i", [0, 1.25, 2.5, 3.75, 5]),
     (0, 10, "%i", [0, 2.5, 5, 7.5, 10]),
-    (0, 0, "%.1f", [0, 0, 0, 0, 0]),
+    (0, 0, "%.1f", [0]),
     (0, 1, "%.1f", [0, 0.25, 0.5, 0.75, 1]),
     (1, 2, "%.1f", [1, 1.25, 1.5, 1.75, 2]),
     (1.1, 1.2, "%.1f", [1.1, 1.125, 1.15, 1.175, 1.2]),
+    (0, np.nextafter(0, 1), "%.1f", [0.e+000, 5.e-324]),
 ])
 def test_get_ticks_matplotlib(vmin, vmax, cbar_tick_format, expected):
     ticks = _get_ticks_matplotlib(vmin, vmax, cbar_tick_format)
```

nilearn/plotting/surf_plotting.py
```diff
@@ -341,12 +341,12 @@ def _get_ticks_matplotlib(vmin, vmax, cbar_tick_format):
...
-    return ticks
+    # remove duplicate ticks when vmin == vmax
+    return np.array(list(dict.fromkeys(ticks)))
```
cr3, line 349: Duplicates can only ever happen when using `np.linspace` so I'd only use `np.unique` (or `np.array(list(dist.fromkeys(...` if you prefer) with that function.

nilearn/plotting/surf_plotting.py
DimitriPapadopoulos: I don't know how much portable are the results of the last test from platform to platform, but let's give it a try.

And thanks for `np.unique`, it is self-explaining unlike my unreadable code.

nilearn/plotting/surf_plotting.py
cr3: Note that I'm only suggesting this for the reader so that they don't wonder how duplicates might happen when using `np.arange`. I'm not suggesting this for performance reasons because the array is so small that it doesn't really make a difference ;)

NicolasGensollen (approved):
LGTM, thanks !
Could you add an entry in `doc/changes/latest.rst` ?

ymzayek (approved):
Thanks all! LGTM! Just waiting to see if all the tests pass

ymzayek:
Merging

Issue #3108, opened 2021-12-20
distutils.version.LooseVersion is deprecated


drammock:
distutils is deprecated in python 3.10 (the deprecation warning for `LooseVersion` is showing up in our python 3.10 CIs). Over at MNE-Python we use pkg_resources now instead: 

https://github.com/mne-tools/mne-python/blob/433a4e19bb15b061c17a3fee3e9f8e8f6d8ea229/mne/fixes.py#L50-L55

Statsmodels recently addressed it with `packaging.Version` and `packaging.parse`, which seems just as good: https://github.com/statsmodels/statsmodels/pull/7946/files

PR #3109, opened 2021-12-21
Fix hommel value computation

bthirion:
Closes # .

- I just realized that the Hommel values computed by the procedure were slightly wrong.
I fixed the code and added a test to make sure that the computed values are identical to what the theory requires on a simpl, yet non-trivial case, according to the literature


bthirion:
The failure seems unrelated, right ?

jeromedockes:
> The failure seems unrelated, right ?

yes, they are due to #3108

.circleci/config.yml
```diff
@@ -376,3 +376,4 @@ workflows:
                 - main
     jobs:
       - build_docs
+
```
jeromedockes, line 379: just curious: why this change?

nilearn/glm/tests/test_thresholding.py
```diff
@@ -103,6 +104,24 @@ def test_threshold_stats_img():
         threshold_stats_img(None, None, alpha=0.05, height_control='plop')
 
 
+def test_hommel():
+    # Check that the computation of Hommel value
+    # For these, we take the example in  Meijer et al. 2017
+    # 'A shortcut for Hommel's procedure in linearithmic time'
+    # and check that we obtain the same values
+    z = norm.isf([1.e-8, .01, .08, .1, .5, .7, .9])
+    assert(_compute_hommel_value(z, 1.e-9) == 7)
```
jeromedockes, line 113: the parentheses make it look like a function call which is a bit weird, I would remove them or at least add a space between `assert` and the expression

jeromedockes (commented):
LGTM from a quick look! but I don't know anything about the Hommel value

.circleci/config.yml
bthirion: Hm, just me poorly dealing with local changes, sorry.

bthirion:
I'd be grateful for a review ...

NicolasGensollen (approved):
LGTM, I had a look at the reference and the test is consistent with the example of Fig. 1
Maybe we could add a whatsnew entry?

NicolasGensollen:
Thanks! Merging...

PR #3113, opened 2022-01-04
[MAINT] `LooseVersion` deprecations

### LINKED ISSUE #3108: distutils.version.LooseVersion is deprecated

NicolasGensollen:
Closes #3108 

Thanks to MNE folks for the fix. :+1: 
I'll add some unit tests for `_compare_version` once the CI is green again.

### DIFF
--- a/nilearn/version.py
+++ b/nilearn/version.py
@@ -128,11 +128,12 @@ def _compare_version(version_a, operator, version_b):
 
     """
     try:
-        from pkg_resources import parse_version as parse
+        from pkg_resources import parse_version as parse  # noqa:F401
     except ImportError:
-        from distutils.version import LooseVersion as parse
-
-    return eval(f'parse("{version_a}") {operator} parse("{version_b}")')
+        from distutils.version import LooseVersion as parse  # noqa:F401
+    return eval(
+        'parse("{0}") {1} parse("{2}")'.format(version_a, operator, version_b)
+    )
 
 
 def _check_module_dependencies(is_nilearn_installing=False):
--- a/nilearn/version.py
+++ b/nilearn/version.py
@@ -102,6 +102,41 @@ def _import_module_with_version_check(
     return module
 
 
+def _compare_version(version_a, operator, version_b):
+    """Compare two version strings via a user-specified operator.
+
+    ``distutils`` has been deprecated since Python 3.10 and is scheduled
+    for removal from the standard library with the release of Python 3.12.
+    For version comparisons, we use setuptools's `parse_version` if available.
+
+    Note : function borrowed from the MNE team.
+
+    Parameters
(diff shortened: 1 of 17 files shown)


nilearn/version.py
```diff
@@ -102,6 +100,42 @@ def _import_module_with_version_check(
     return module
 
 
+def _compare_version(version_a, operator, version_b):
```
jeromedockes, line 103: if this is copied from mne could we add a link to the place it comes from?

nilearn/version.py
```diff
@@ -102,6 +100,42 @@ def _import_module_with_version_check(
...
+    result : :obj:`bool`
+        The result of the version comparison.
+
+    """
+    try:
+        from pkg_resources import parse_version as parse  # noqa:F401
+    except ImportError:
+        from distutils.version import LooseVersion as parse  # noqa:F401
+    return eval(
+        'parse("{0}") {1} parse("{2}")'.format(version_a, operator, version_b)
```
jeromedockes, line 135: In general using `eval` (esp. on unvalidated input) is not a very good habit, could we use a mapping from strings to functions in the standard library module [`operator`](https://docs.python.org/3/library/operator.html) instead?

NicolasGensollen: Good point indeed, thanks @jeromedockes !
I added a mapping in 3c6e888

nilearn/version.py
```diff
@@ -102,6 +100,42 @@ def _import_module_with_version_check(
...
+        Second version string.
+
+    Returns
+    -------
+    result : :obj:`bool`
+        The result of the version comparison.
+
+    """
+    try:
+        from pkg_resources import parse_version as parse  # noqa:F401
```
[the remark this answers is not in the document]
NicolasGensollen: I can add a comment if you think that it could be useful

jeromedockes (approved):
LGTM, thanks @NicolasGensollen !

nilearn/version.py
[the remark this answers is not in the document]
bthirion: Would be useful indeed.

bthirion (approved):
Looks good, thx !

NicolasGensollen:
Thanks! Merging...

PR #3114, opened 2022-01-04
avoid unnecessary data loading if dtype casting not needed

### LINKED ISSUE #3112: `image.load_img` unbearably slow?
At my end nilearn image's `load_img` is almost 4 orders of magnitude slower than nibabel's `load` on a fresh install.

Nilearn version:

Latest on pip -> 0.8.1

### Expected behavior
`image.load_img` adds no noticeable overhead over nibabel's load.

### Actual behavior
`image.load_img` takes seconds instead of microseconds.

```sh
base ❯ conda --version
conda 4.11.0
base ❯ conda create -n ni python=3
base ❯ conda activate ni
ni ❯ pip install nilearn nibabel
ni ❯ python -m timeit -n 3 -s "from nilearn import image" 'image.load_img("bold.nii.gz")'
3 loops, best of 5: 6.36 sec per loop
ni ❯ python -m timeit -n 3 -s "import nibabel as nib " 'nib.load("bold.nii.gz")'
3 loops, best of 5: 885 usec per loop
```

```python
Python 3.10.1 | packaged by conda-forge | (main, Dec 22 2021, 01:39:03) [Clang 11.1.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from nilearn import image
>>> image.__file__
'/Users/dangom/anaconda3/envs/ni/lib/python3.10/site-packages/nilearn/image/__init__.py'
>>> import nilearn
>>> nilearn.__version__
'0.8.1'
>>> import nibabel as nib
>>> nib.__file__
'/Users/dangom/anaconda3/envs/ni/lib/python3.10/site-packages/nibabel/__init__.py'
>>> nib.__version__
'3.2.1'
```

dangom:
Closes #3112 .

This makes image.load_img behave more like nibabel.load in the case
that no data type conversion is needed.
This improves the speed of image.load_img  and improves usability
used when users only want to access a header or affine from a file.

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -131,7 +131,7 @@ def load_niimg(niimg, dtype=None):
                         " not compatible with nibabel format:\n"
                         + _repr_niimgs(niimg, shorten=True))
 
-    dtype = _get_target_dtype(_get_data(niimg).dtype, dtype)
+    dtype = _get_target_dtype(niimg.dataobj.dtype, dtype)
 
     if dtype is not None:
         # Copyheader and set dtype in header if header exists


bthirion:
Thx. How much does this improve performance ?

dangom:
from 6 seconds to 800 microseconds (7500x faster) in an example dataset with 164x164x16x300.

jeromedockes:
Thanks! I think this is useful.
but it does change the semantics of the function: before, calling `load_img` or `check_niimg` forced loading the data by calling `get_data`, and the documentation does say 
> It is guaranteed that the returned object has an affine attributes and that nilearn.image.get_data returns its data.

(and the term "load" also seems to indicate that)

as the documentation is not super explicit I think it's probably OK to change it though, or maybe we could add an option controlling that behavior

the `fetch_atlas_talairach` failure in the tests is genuine: before this PR it was ok to remove the image file after calling `load_img` because the image was loaded in memory; now it isn't so if we keep the new behavior we should fix the fetcher by explicitly calling `get_data` on the image

jeromedockes:
> from 6 seconds to 800 microseconds (7500x faster) in an example dataset with 164x164x16x300.

note in most cases we will eventually want to do something with the image so these 6 seconds will be paid at some point, but I agree lazy loading can be useful

GaelVaroquaux:
> the fetch_atlas_talairach failure in the tests is genuine: before this PR it was ok to remove the image file after calling load_img because the image was loaded in memory; now it isn't so if we keep the new behavior we should fix the fetcher by explicitly calling get_data on the image

That's somewhat an illustration that:

- The new semantics are more complex for the user (which actually enables the performance)
- Some people will suffer breakage

The tradeoff is here: ease of use vs performance. Historically nilearn has chosen ease of use.

bthirion:
We may discuss this at the newt office hours ?

dangom:
Eagerly loading the data seems to have been happening by accident (for checking a dtype), not intention.
Currently, nibabel and nilearn are not interchangeable: from nilearn it is not possible to access basic header info such as TR or look at the affine without paying a huge upfront cost.

If nilearn were to break on a nibabel input, wouldn't we take that as a nilearn bug? That's what is happening in the PR that is failing.

GaelVaroquaux:
> Eagerly loading the data seems to have been happening by accident (for checking a dtype), not intention.

I wouldn't be so sure that things are completely by chance. We tracked
down really nasty bugs/undesirable behavior that confused user with
regards to eager vs lazy loading of data (sorry, no energy to find the
discussion). Series of problems with the variety of behavior of data
loading depending on whether it with input in memory or given as files
lead us to systematically use our internal handling code.

> Currently, nibabel and nilearn are not interchangeable:

I think that this is fair. Nilearn is certainly not trying to be
interchangeable with nibabel.

> If nilearn were to break on a nibabel input, wouldn't we take that as a nilearn
> bug? That's what is happening in the PR that is failing.

I'm not following you.

Also, I would really like to stress that if you want to squeeze every bit
of performance wrt data I/O, low-level nibabel is recommended. It offers
much more control.

dangom:
Right, I don't need/want to squeeze any bit of performance out of nilearn. I wanted to compare affines and read the TR without loading 1G of data and waiting 10 seconds every time. I'm happy doing that with nibabel.

I was just surprised that nilearn was eagerly loading the data. But the fact that some of nilearn's functionality depends on the data being eagerly loaded is surprising to me, that is what I meant by

"If nilearn were to break on a nibabel input, wouldn't we take that as a nilearn
 bug? That's what is happening in the PR (meant CI) that is failing."

As in that case loading the data with nibabel would've lead to a CI failure.

jeromedockes:
so to summarize -- `nilearn.image.load_img` loads the image, as its name and documentation indicate. Note it can also load a 4d image from a list of filenames (3d images) or a wildcard, in which case "lazy" loading cannot even be considered. to read only the header without reading the data one can use `nibabel.load`.

IIUC we can close this and the related issue?

PR #3119, opened 2022-01-07
[MAINT] Fix `matplotlib` intersphinx URL

NicolasGensollen:
Avoids the redirection.


tsalo (approved):
I just checked the build artifacts and the redirect is indeed no longer present.

NicolasGensollen:
Thanks for the review!
Merging...

PR #3127, opened 2022-01-11
[DOC] Add an example on beta series modeling

tsalo:
Closes None, but follows up on something discussed in this month's devs meeting.

- Add an example that uses nilearn to create beta series models (LSA and LSS).

To do:

- [ ] Add a functional connectivity analysis to the example.

### DIFF
--- a/examples/04_glm_first_level/plot_beta_series.py
+++ b/examples/04_glm_first_level/plot_beta_series.py
@@ -1,0 +1,175 @@
+"""
+Beta-Series Modeling for Task-Based Functional Connectivity and Decoding
+========================================================================
+This example shows how to run :term:`GLM`s for beta series models, which are a
+common modeling approach for a variety of analyses of task-based fMRI
+data with an event-related task design, including functional connectivity,
+decoding, and representational similarity analysis.
+
+Beta series models fit trial-wise conditions, which allow users to create
+"time series" of these trial-wise maps, which can be substituted for the
+typical time series used in resting-state functional connectivity analyses.
+Generally, these models are most useful for event-related task designs,
+while other modeling approaches (e.g., PPI) tend to perform better in block
+designs, depending on the type of analysis.
+See [Cisler, Bush, & Steele (2014)]
+(https://doi.org/10.1016/j.neuroimage.2013.09.018)
+for more information about this, in the context of functional connectivity
+analyses.
+
+Two of the most well-known beta series modeling methods are
+Least Squares- All (LSA) and Least Squares- Separate (LSS).
+In LSA, a single :term:`GLM` is run, in which each trial of each condition of
+interest is separated out into its own condition within the design matrix.
+In LSS, each trial of each condition of interest has its own :term:`GLM`,
+in which the targeted trial receives its own column within the design matrix,
+but everything else remains the ssame as the standard model.
+Trials are then looped across, and many GLMs are fitted,
+with the parameter estimate map extracted from each GLM to build the LSS beta
+series.
+"""
+from nilearn.glm.first_level import FirstLevelModel
+from nilearn import image, masking, plotting
+
+##############################################################################
+# Prepare data and analysis parameters
+# -------------------------------------
+# Prepare the data.
+from nilearn.datasets import fetch_language_localizer_demo_dataset
+
+data_dir, _ = fetch_language_localizer_demo_dataset()
+
+from nilearn.glm.first_level import first_level_from_bids
+
+task_label = 'languagelocalizer'
+models, models_run_imgs, events_dfs, models_confounds = \
+    first_level_from_bids(
+        data_dir,
+        task_label,
+        img_filters=[('desc', 'preproc')],
+    )
+
+standard_glm = models[0]
+fmri_file = models_run_imgs[0][0]
+events_df = events_dfs[0][0]
+confounds_df = models_confounds[0][0]
+
+mask = masking.compute_epi_mask(fmri_file)
+
+# We will use :func:`~nilearn.glm.first_level.first_level_from_bids`'s
+# parameters for the other models.
+glm_parameters = standard_glm.get_params()
+# We need to override one parameter (signal_scaling) with the value of
+# scaling_axis.
+glm_parameters["signal_scaling"] = standard_glm.scaling_axis
+
+##############################################################################
+# Define the standard model
+# -------------------------------------
+# Here, we create a basic GLM for this run, which we can use to highlight
+# differences between the standard modeling approach and beta series models.
+# We will just use the one created by
+# :func:`~nilearn.glm.first_level.first_level_from_bids`.
+standard_glm.fit(fmri_file, events_df)
+
+##############################################################################
+# Define the LSA model
+# -------------------------------------
+# We will now create an LSA model.
+
+# Transform the DataFrame for LSA
+lsa_events_df = events_df.copy()
+conditions = lsa_events_df["trial_type"].unique()
+lsa_events_df["old_trial_type"] = lsa_events_df["trial_type"]
+condition_counter = {c: 0 for c in conditions}
+for i_trial, trial in lsa_events_df.iterrows():
+    trial_condition = trial["old_trial_type"]
+    condition_counter[trial_condition] += 1
+    # We use a unique delimiter here (``__``) that shouldn't be in the
+    # original condition names.
+    trial_name = f"{trial_condition}__{condition_counter[trial_condition]:03d}"
+    lsa_events_df.loc[i_trial, "trial_type"] = trial_name
+
+lsa_glm = FirstLevelModel(**glm_parameters)
+lsa_glm.fit(fmri_file, lsa_events_df)
+
+# Collect the parameter estimate maps
+lsa_beta_maps = {cond: [] for cond in events_df["trial_type"].unique()}
+trialwise_conditions = lsa_events_df["trial_type"].unique()
+for condition in trialwise_conditions:
+    beta_map = lsa_glm.compute_contrast(condition, output_type="effect_size")
+    # Drop the trial number from the condition name to get the original name.
+    condition_name = condition.split("__")[0]
+    lsa_beta_maps[condition_name].append(beta_map)
+
+# We can concatenate the lists of 3D maps into a single 4D beta series for
+# each condition, if we want.
+lsa_beta_maps = {
+    name: image.concat_imgs(maps) for name, maps in lsa_beta_maps.items()
+}
+
+##############################################################################
+# Define the LSS models
+# -------------------------------------
+
+
+def lss_transformer(df, row_number):
+    """Label one trial for one LSS model."""
+    df = df.copy()
+
+    # Technically, all you need is for the requested trial to have a unique
+    # "trial_type" *within* the dataframe, rather than across models.
+    # However, we may want to have meaningful "trial_type"s (e.g., "Left_001")
+    # across models, so that you could track individual trials across models.
+    df["old_trial_type"] = df["trial_type"]
+
+    # Determine which number trial it is *within the condition*.
+    trial_condition = df.loc[row_number, "old_trial_type"]
+    trial_type_series = df["old_trial_type"]
+    trial_type_series = trial_type_series.loc[
+        trial_type_series == trial_condition
+    ]
+    trial_type_list = trial_type_series.index.tolist()
+    trial_number = trial_type_list.index(row_number)
+    # We use a unique delimiter here (``__``) that shouldn't be in the
+    # original condition names.
+    trial_name = f"{trial_condition}__{trial_number:03d}"
+    df.loc[row_number, "trial_type"] = trial_name
+    return df, trial_name
+
+
+# Transform the DataFrame for LSS
+lss_beta_maps = {cond: [] for cond in events_df["trial_type"].unique()}
+lss_design_matrices = []
+
+for i_trial in range(events_df.shape[0]):
+    lss_events_df, trial_condition = lss_transformer(events_df, i_trial)
+
+    # Compute and collect beta maps
+    lss_glm = FirstLevelModel(**glm_parameters)
+    lss_glm.fit(fmri_file, lss_events_df)
+
+    # We will save the design matrices across trials to show them later
+    lss_design_matrices.append(lss_glm.design_matrices_[0])
+
+    beta_map = lss_glm.compute_contrast(
+        trial_condition,
+        output_type="effect_size",
+    )
+
+    # Drop the trial number from the condition name to get the original name.
+    condition_name = trial_condition.split("__")[0]
+    lss_beta_maps[condition_name].append(beta_map)
+
+# We can concatenate the lists of 3D maps into a single 4D beta series for
+# each condition, if we want.
+lss_beta_maps = {
+    name: image.concat_imgs(maps) for name, maps in lss_beta_maps.items()
+}
+
+##############################################################################
+# Compare the three modeling approaches
+# -------------------------------------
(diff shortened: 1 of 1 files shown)


jeromedockes:
cool, thanks! I wonder if this should go in the "advanced" examples section?

tsalo:
It definitely could. Beta series models are generally used for advanced analyses, like functional connectivity, decoding, or RSA. I'll move it to that section.

One random question- as far as I can tell, the examples in nilearn's gallery are sorted using sphinx-gallery's default sorting method, which is based on the number of code lines in the examples. Is that right, or is there a custom sorting method being used that I missed?

bthirion:
Note that the example https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_glm_decoding.html already uses beta series regression to perform decoding. We could cross-references different instance of beta series regression.

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,175 @@
...
+confounds_df = models_confounds[0][0]
+
+mask = masking.compute_epi_mask(fmri_file)
+
+# We will use :func:`~nilearn.glm.first_level.first_level_from_bids`'s
+# parameters for the other models.
+glm_parameters = standard_glm.get_params()
+# We need to override one parameter (signal_scaling) with the value of
+# scaling_axis.
+glm_parameters["signal_scaling"] = standard_glm.scaling_axis
```
bthirion, line 64: The reason for that is not obvious  to me.

tsalo: The parameter `signal_scaling` must be an integer for the axis, but after initialization that integer is stored as `scaling_axis` and the saved parameter for `signal_scaling` is a boolean, so if you grab the saved parameter and feed it directly into a new FirstLevelModel, it will raise an exception when it tries to standardize based on a boolean.

Here's a way to reproduce it:

```python
from nilearn.datasets import fetch_language_localizer_demo_dataset
from nilearn.glm.first_level import FirstLevelModel, first_level_from_bids

data_dir, _ = fetch_language_localizer_demo_dataset()
models, models_run_imgs, events_dfs, models_confounds = \
    first_level_from_bids(
        data_dir,
        'languagelocalizer',
        img_filters=[('desc', 'preproc')],
    )
standard_glm = models[0]
fmri_file = models_run_imgs[0][0]
events_df = events_dfs[0][0]

# Grab the parameters
glm_parameters = standard_glm.get_params()

# We can fit the original GLM just fine
standard_glm.fit(fmri_file, events_df)

# But not a new one
new_glm = FirstLevelModel(**glm_parameters)
new_glm.fit(fmri_file, events_df)
```

Here's the traceback:

```Text
<ipython-input-1-86ffa75b5c7d> in <module>
     21 # But not a new one
     22 new_glm = FirstLevelModel(**glm_parameters)
---> 23 new_glm.fit(fmri_file, events_df)

~/Documents/tsalo/nilearn/nilearn/glm/first_level/first_level.py in fit(self, run_imgs, events, confounds, design_matrices, bins)
    568 
    569             if self.signal_scaling:
--> 570                 Y, _ = mean_scaling(Y, self.scaling_axis)
    571             if self.memory:
    572                 mem_glm = self.memory.cache(run_glm, ignore=['n_jobs'])

~/Documents/tsalo/nilearn/nilearn/glm/first_level/first_level.py in mean_scaling(Y, axis)
     59 
     60     """
---> 61     mean = Y.mean(axis=axis)
     62     if (mean == 0).any():
     63         warn('Mean values of 0 observed.'

/opt/miniconda3/lib/python3.8/site-packages/numpy/core/_methods.py in _mean(a, axis, dtype, out, keepdims, where)
    177             is_float16_result = True
    178 
--> 179     ret = umr_sum(arr, axis, dtype, out, keepdims, where=where)
    180     if isinstance(ret, mu.ndarray):
    181         ret = um.true_divide(

TypeError: an integer is required
```

bthirion: Then this is a bug. We should open an issue for that.

tsalo: I'm going to resolve this thread since I opened #3134. I'll update this branch once #3135 is merged.

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,191 @@
...
+for more information about this, in the context of functional connectivity
+analyses.
+
+Two of the most well-known beta series modeling methods are
+Least Squares- All (LSA) and Least Squares- Separate (LSS).
+In LSA, a single :term:`GLM` is run, in which each trial of each condition of
+interest is separated out into its own condition within the design matrix.
+In LSS, each trial of each condition of interest has its own :term:`GLM`,
+in which the targeted trial receives its own column within the design matrix,
+but everything else remains the ssame as the standard model.
```
bthirion, line 26:
```suggestion
but everything else remains the same as the standard model.
```

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,191 @@
...
+
+from nilearn.glm.first_level import first_level_from_bids
+
+models, models_run_imgs, events_dfs, models_confounds = \
+    first_level_from_bids(
+        data_dir,
+        'languagelocalizer',
+        img_filters=[('desc', 'preproc')],
+    )
+
```
bthirion, line 50: Here you probably want to epxlain that you(re extracting data from a single subject.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -12,7 +12,8 @@
 from scipy import linalg
 from sklearn.utils import check_random_state
```
bthirion, line 13: All what comes next is unrelated, right ?

tsalo: Yeah, that's weird... These changes are coming from #3131 after I updated from `main`. I'm not sure why they're showing up in the diff.

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,354 @@
...
+    figure=fig,
+    axes=axes[1],
+)
+display.add_markers(
+    marker_coords=coords,
+    marker_color='g',
+    marker_size=300,
+)
+
+fig.show()
```
bthirion, line 354: I thought that your point would be to compare LSA vs LSS ?

bthirion (commented):
Looks nice overall.

examples/07_advanced/plot_beta_series.py
tsalo: I had considered it, but I was worried that anything I could show with one run of one specific task would be misleading, given that the tradeoffs between LSA and LSS should only become apparent when you compare across designs (e.g., fast vs. slow event-related tasks).

bthirion: Makes sense, but I'm just worried to leave users with yet another alternative without any information on how to choose one versus the other.

tsalo: I added a topic section on choosing the right model toward the top:

image

There are citations in there too, but they don't show up in the CI build.

bthirion:
LGTM, one should mostly drop the unrelated additions.

tsalo:
Updating my branch from `main` seems to have corrected the diff.

bthirion (commented):
Looks good ! 
Just one minor comment on the final figure.

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,378 @@
...
+string_beta_series = brain_masker.fit_transform(lss_beta_maps['string'])
+string_corrs = np.dot(
+    string_beta_series.T,
+    string_seed_beta_series,
+) / string_seed_beta_series.shape[0]
+string_connectivity_img = brain_masker.inverse_transform(string_corrs.T)
+
+# Show both correlation maps
+fig, axes = plt.subplots(figsize=(10, 8), nrows=2)
+
```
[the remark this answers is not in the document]
tsalo: I expanded the description of the last section in 7f5d780. Is that what you had in mind?

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,382 @@
...
+    t_r=None,
+    memory='nilearn_cache',
+    memory_level=1,
+    verbose=0,
+)
+
+# Perform the seed-to-voxel correlation for the LSS 'language' beta series
+lang_seed_beta_series = seed_masker.fit_transform(lss_beta_maps['language'])
+lang_beta_series = brain_masker.fit_transform(lss_beta_maps['language'])
+lang_corrs = np.dot(
```
bthirion, line 333: Why not use a ConnectivityMeasure of Nilearn ?
It would avoid the np import.

examples/07_advanced/plot_beta_series.py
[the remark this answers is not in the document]
bthirion: Yes, thx.

examples/07_advanced/plot_beta_series.py
bthirion: OK.

bthirion (commented):
Sounds good overall. Maybe add a whatsnew entry.

examples/07_advanced/plot_beta_series.py
tsalo: I just used the seed-to-voxel example's code for this, but I could use ConnectivityMeasure if that will make this example simpler. I haven't used ConnectivityMeasure in the past, though, and all of the examples seem to be for ROI-to-ROI analyses, where every column in an array is compared to every other column. That doesn't seem to fit easily with the seed-to-voxel approach. Am I misunderstanding the class and/or its examples?

tsalo:
I just added the whatsnew entry. Thanks!

examples/07_advanced/plot_beta_series.py
bthirion: Indeed, you're right. I did not realize this when reading the code.

bthirion (approved):
Good on my side. Thx !

tsalo:
@bthirion is this good to merge or should I wait for a second approval?

bthirion:
LGTM ! Ideally, we need a second approval before merging. @NicolasGensollen ?

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,382 @@
...
+    According to :cite:t:`abdulrahman2016effect`,
+    the decision between LSA and LSS should be based on three factors:
+    inter-trial variability, scan noise, and stimulus onset timing.
+    While :cite:t:`mumford2012deconvolving` proposes LSS as a tool primarily
+    for fast event-related designs (i.e., ones with short inter-trial
+    intervals), :cite:t:`abdulrahman2016effect` finds, in simulations,
+    that LSA performs better than LSS when trial variability is greater
+    than scan noise, even in fast designs.
+"""
+# sphinx_gallery_thumbnail_number = -2
```
NicolasGensollen, line 56: I think this doesn't render as expected, right?

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,382 @@
...
+# We will use first_level_from_bids's parameters for the other models
+glm_parameters = standard_glm.get_params()
+# We need to override one parameter (signal_scaling) with the value of
+# scaling_axis
+glm_parameters['signal_scaling'] = standard_glm.scaling_axis
+
+##############################################################################
+# Define the standard model
+# -------------------------
+# Here, we create a basic GLM for this run, which we can use to highlight
```
NicolasGensollen, line 94:
```suggestion
# Here, we create a basic :term:`GLM` for this run, which we can use to highlight
```

NicolasGensollen (commented):
Thanks @tsalo ! This looks great! 👍 
I don't really feel qualified and I don't have a lot of time right now to review the logic of your code, but here are a few points and suggestions on formatting.
Also, it could be nice to reference this specific example somewhere (not sure where...) in the user guide such as to increase the discovery probability for users. WDYT?

examples/07_advanced/plot_beta_series.py
tsalo: It looks like it's using the second-to-last figure as the thumbnail, which is what I was aiming for.

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,382 @@
...
+representational similarity analysis.
+
+Beta series models fit trial-wise conditions, which allow users to create
+"time series" of these trial-wise maps, which can be substituted for the
+typical time series used in resting-state functional connectivity analyses.
+Generally, these models are most useful for event-related task designs,
+while other modeling approaches, such as psychophysiological interactions
+(PPIs), tend to perform better in block designs, depending on the type of
+analysis.
+See :cite:t:`cisler2014comparison` for more information about this,
```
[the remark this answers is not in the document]
tsalo: Thanks! I've added it in [0c197a3](https://github.com/nilearn/nilearn/pull/3127/commits/0c197a380d5d280e8c1487dfc671301ef7b347e0) and I'll see if it renders.

examples/07_advanced/plot_beta_series.py
NicolasGensollen: I meant that it shows up in the example, but I realize that this is the only example in Nilearn where we'd use this.
I think it is possible to avoid having these configuration comments showing up in the rendered example. It _should_ be as simple as [adding a line in the config](https://sphinx-gallery.github.io/stable/configuration.html#removing-config-comments).

examples/07_advanced/plot_beta_series.py
```diff
@@ -0,0 +1,389 @@
...
+    )
+    axes[i_trial].set_title(f'Trial {i_trial + 1}')
+
+fig.show()
+
+##############################################################################
+# Compare the three modeling approaches
+# -------------------------------------
+
+fig, axes = plt.subplots(
```
NicolasGensollen, line 251: Probably less important, and maybe even a choice of yours to increase readability, but this code cell and the next one could be factorized with a for loop. Feel free to ignore this comment if you think that the code is easier to read like this.

NicolasGensollen (commented):
Thanks for addressing my points @tsalo ! :+1: 
The bib references seem to render well now !
Might be worth trying to see if we can make the sphinx config comment disappear in this PR (see comment below). Other than that, this looks good to merge!

examples/07_advanced/plot_beta_series.py
tsalo: Ohhh... I didn't even realize that was an option! That's awesome. I'll try it out.

doc/changes/latest.rst
```diff
@@ -17,6 +17,9 @@ Fixes
 Enhancements
 ------------
 
+- New example in
+  :ref:`sphx_glr_auto_examples_07_advanced_plot_beta_series.py`
+  to demonstrate how to implement common beta series models with nilearn.
```
NicolasGensollen, line 22:
```suggestion
  to demonstrate how to implement common beta series models with nilearn (:gh:`3127` by `Taylor Salo`_).
```

NicolasGensollen (commented):
Thanks @tsalo !
Just one minor thing in the changelog, but LGTM otherwise!

doc/changes/latest.rst
tsalo: Thanks for catching that @NicolasGensollen.

NicolasGensollen (approved):
LGTM, thanks @tsalo ! :+1:

bthirion:
Please go ahead.

tsalo:
Thanks!

PR #3141, opened 2022-01-24
[ENH] Rename objects of `nilearn.decomposition` with leading "_"

### LINKED ISSUE #3138: API of `nilearn.decomposition`
The `decomposition` module only exposes `CanICA` and `DictLearning` currently: https://nilearn.github.io/dev/modules/reference.html#module-nilearn.decomposition

Here are a list of objects that we should either add to the public API or rename to make it clear that they are private (we would also rename `base.py` into `_base.py`):

- `MultiPCA`

https://github.com/nilearn/nilearn/blob/c123b6126ef36dd5a08c20b780b773a80278b907/nilearn/decomposition/multi_pca.py#L13-L17

- `fast_svd`

https://github.com/nilearn/nilearn/blob/c123b6126ef36dd5a08c20b780b773a80278b907/nilearn/decomposition/base.py#L27-L29

- `mask_and_reduce`

https://github.com/nilearn/nilearn/blob/c123b6126ef36dd5a08c20b780b773a80278b907/nilearn/decomposition/base.py#L85-L92

- `BaseDecomposition`

https://github.com/nilearn/nilearn/blob/c123b6126ef36dd5a08c20b780b773a80278b907/nilearn/decomposition/base.py#L233-L237

- `explained_variance`

https://github.com/nilearn/nilearn/blob/c123b6126ef36dd5a08c20b780b773a80278b907/nilearn/decomposition/base.py#L539-L544

NicolasGensollen:
Closes #3138 

This PR proposes to rename some objects and submodules of `nilearn.decomposition` with a leading "_" to make it clear that these objects are private.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -54,3 +54,4 @@ Changes
 - Refactors fmriprep confound loading such that that the parsing of the 
   relevant image file and the loading of the confounds are done in 
   separate steps (:gh:`3274` by `David G Ellis`_).
+- Private submodules, functions, and classes from the :mod:`~nilearn.decomposition` module now start with a "_" character to make it clear that they are not part of the public API (:gh:`3141` by `Nicolas Gensollen`_).
--- a/nilearn/decomposition/__init__.py
+++ b/nilearn/decomposition/__init__.py
@@ -1,6 +1,6 @@
 """
 The :mod:`nilearn.decomposition` module includes a subject level
-variant of the ICA called Canonical ICA.
+variant of the :term:`ICA` called Canonical :term:`ICA`.
 """
 from .canica import CanICA
 from .dict_learning import DictLearning
--- a/nilearn/decomposition/_base.py
+++ b/nilearn/decomposition/_base.py
@@ -24,7 +24,7 @@ from nilearn.maskers._masker_validation import _check_embedded_nifti_masker
 from ..signal import _row_sum_of_squares
 
 
-def fast_svd(X, n_components, random_state=None):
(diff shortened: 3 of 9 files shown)


bthirion (approved):
LGTM, thx !

bthirion:
Sorry if we discussed it already, but we don't have deprecation cycle ?

jeromedockes:
I think if they weren't documented nor available in `nilearn.decompositon`'s namespace it's probably ok to skip the deprecation cycle. I just wonder if BaseDecomposition and MultiPCA should be changed as they seem to be used in other modules

ymzayek:
Hi, I'm happy to give this a review but I am wondering if there's a particular reason it wasn't finished that I should be aware of?

NicolasGensollen:
@ymzayek I don't believe there was a particular reason, it wasn't a priority for releasing `0.9.1` and we just forgot about it afterwards.
I'd be happy to finish it if you have some time for reviews! 🎉

ymzayek (commented):
LGTM! I wonder if it is worth a whatsnew entry. I think it's usually not necessary for private functions but since they weren't explicitly private before might be worth a mention. WDYT?

NicolasGensollen:
Thanks for the review @ymzayek !
I rebased and added a what's new entry as suggested.
I also triggered a full build to make sure docs build without any issue.

ymzayek:
I think this is ok to merge, unless @bthirion you want to have another look since the first approval was done a while ago?

bthirion:
I'm happy with that one. green lights !

ymzayek:
Ok merging

PR #3143, opened 2022-01-25
[MAINT] Bump dependencies for release `0.9.0`

NicolasGensollen:
This PR proposes to bump our dependencies for the 0.9.0 release:

- Joblib: 0.12  -->  0.15
- Numpy: 1.16  -->  1.18
- Scipy: 1.2  -->  1.5
- Scikit-learn: 0.21  -->  0.22
- Pandas:  0.24  --> 1.0

I also realized that the README says that matplotlib >= 1.5.1 is required although we don't test this. The minimum requirements tests are without matplotlib:

https://github.com/nilearn/nilearn/blob/c549dc0a66327b05ab3a25ddff02abe99db02501/.github/workflows/testing.yml#L26-L31

I think we should add a CI job testing minimum requirements WITH matplotlib. In this case, we could also bump the minimum version for matplotlib (>= 3.0 ??). 

WDYT?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,12 +46,12 @@ The required dependencies to use the software are:
 
 * Python >= 3.6
 * setuptools
-* Numpy >= 1.16
-* SciPy >= 1.2
-* Scikit-learn >= 0.21
-* Joblib >= 0.12
+* Numpy >= 1.18
+* SciPy >= 1.5
+* Scikit-learn >= 0.22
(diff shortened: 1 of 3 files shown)


bthirion (approved):
LGTM. You're adding the test with mpl 3.0 in another PR ?

NicolasGensollen:
Thanks! Merging...

PR #3150, opened 2022-01-28
[INFRA] Update docs `install` target

NicolasGensollen:
This PR updates the steps for the `install` target of the documentation Makefile.
This is necessary since we now deploy the development version of the documentation every time a PR is merged in main.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -169,16 +169,17 @@ install:
 	# first clone the nilearn.github.io repo because it may ask
 	# for password and we don't want to delay this long build in
 	# the middle of it
-	# --no-checkout just fetches the root folder without content
 	# --depth 1 is a speed optimization since we don't need the
 	# history prior to the last commit
-	git clone --no-checkout --depth 1 git@github.com:nilearn/nilearn.github.io.git _build/nilearn.github.io
+	git clone --depth 1 git@github.com:nilearn/nilearn.github.io.git _build/nilearn.github.io
 	touch _build/nilearn.github.io/.nojekyll
+	rm -Rf _build/nilearn.github.io/stable;
+	mkdir _build/nilearn.github.io/stable;
 	make html
 	cd _build/ && \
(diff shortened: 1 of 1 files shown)


NicolasGensollen:
FYI, I used this updated version of the Makefile to release 0.9.0, but it'b be great to have some reviews before merging for 0.9.1.

alexisthual (approved):
It looks good to me.
A while ago, we mentioned serving documentation for different releases of nilearn, is it still something we want to do? Should we use something like [`sphinx-multiversion`](https://holzhaus.github.io/sphinx-multiversion/master/index.html)?

NicolasGensollen:
>is it still something we want to do?

I think this is something we should aim for. Having stable/dev versions was a first step toward this goal.

>Should we use something like [sphinx-multiversion](https://holzhaus.github.io/sphinx-multiversion/master/index.html)?

Never tried it, but definitely something to consider! I'll try to experiment a bit on toy repos.

bthirion:
> > is it still something we want to do?
> 
> I think this is something we should aim for. Having stable/dev versions was a first step toward this goal.
> 
> > Should we use something like [sphinx-multiversion](https://holzhaus.github.io/sphinx-multiversion/master/index.html)?
> 
> Never tried it, but definitely something to consider! I'll try to experiment a bit on toy repos.

I can see the advantage, but also the burden in terms of maintenance etc. I'm not 100% convinced we should do it. To be discussed in future meetings?

emdupre:
> I can see the advantage, but also the burden in terms of maintenance etc. I'm not 100% convinced we should do it. To be discussed in future meetings?

+1 to discuss this and try to keep the maintenance / UX relatively light -- maybe we could settle on something like only providing documentation for previous major releases. I think this is what MNE does, from a quick look at their docs !

bthirion (approved):

NicolasGensollen:
Thanks for the reviews! Merging...

PR #3153, opened 2022-02-02
Fix image labels

### LINKED ISSUE #3154: Image labels don't correspond to images
The images for the Haxby 2001 experiment show cats and chairs, but are labeled as faces and cats, respectively. 

This could be fixed by changing the image labels to correspond to what is shown, or by selecting different images to correspond to the labels. I'm submitting a PR that changes the image labels.

### Affected pages
https://nilearn.github.io/stable/decoding/decoding_intro.html

ctw:
Changed image labels to correspond with the images (could alternatively show images of faces and cats to match the original labels).

Closes #3154

### DIFF
--- a/doc/decoding/decoding_intro.rst
+++ b/doc/decoding/decoding_intro.rst
@@ -43,19 +43,19 @@ seeing from the :term:`fMRI` activity recorded in regions of the ventral visual 
 Significant prediction shows that the signal in the region contains
 information on the corresponding category.
 
+.. figure:: ../auto_examples/02_decoding/images/sphx_glr_plot_haxby_stimuli_007.png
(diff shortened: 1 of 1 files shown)


NicolasGensollen (commented):
Thanks @ctw !
I think the right fix would be to change the images rather than the captions such that the rest of the tutorial makes sense. WDYT?

ctw:
Good point @NicolasGensollen -- I've now reverted the labels and updated the images as you suggested!

bthirion (approved):
LGTM, thx !

NicolasGensollen (approved):
LGTM, thanks @ctw !

NicolasGensollen:
Thanks! Merging...

PR #3162, opened 2022-02-09
Adds a minor update to new_img_like docstring

### LINKED ISSUE #3161: Confusing Documentation in new_img_like
I think there is some slightly confusing language in the docstring of `new_img_like`. The Parameters list indicates a copy_header parameter to indicate whether the header is copied, but the Returns section indicates "A loaded image with the same type _(and header)_ as the reference image." (emphasis mine). I think it would be more clear to say "A loaded image with the same _file_ type (and, _optionally_ header) as the reference image."
The reason this is critical is that in nibabel, the header data type will override the data type of the numpy array being held (at least for the common Nifti1Image). Due to a number of changes, this is a little more critical (numpy uses 64-bit by default, whereas previously it was 32-bit, and yet AFNI, FSL, and freesurfer all do not support 64-bit integer data).
Somebody could easily scan the "Returns" field and not the "Parameters" field (certainly not a developer like myself, no, never!) and then mistakenly believe that the header will be copied, and therefore that the output data type of the array will match the original header.

Relevant docstring:
https://github.com/nilearn/nilearn/blob/1607b52458c28953a87bbe6f42448b7b4e30a72f/nilearn/image/image.py#L699-L720

jbteves:
Closes #3161  .

- Slightly modify wording of new_img_like docstring

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -697,7 +697,8 @@ def new_img_like(ref_niimg, data, affine=None, copy_header=False):
     Returns
     -------
     Niimg-like object
-        A loaded image with the same file type (and, optionally, header) as the reference image.
+        A loaded image with the same file type (and, optionally, header)
+    as the reference image.
 
     """
     # Hand-written loading code to avoid too much memory consumption
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -697,7 +697,7 @@ def new_img_like(ref_niimg, data, affine=None, copy_header=False):
     Returns
     -------
     Niimg-like object
-        A loaded image with the same type (and header) as the reference image.
+        A loaded image with the same file type (and, optionally, header) as the reference image.
 
     """
     # Hand-written loading code to avoid too much memory consumption
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -698,7 +698,7 @@ def new_img_like(ref_niimg, data, affine=None, copy_header=False):
     -------
     Niimg-like object
         A loaded image with the same file type (and, optionally, header)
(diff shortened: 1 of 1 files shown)


jbteves:
I'm a bit confused about the PEP8 fail here, it's referencing other commits which should not be in this branch.

bthirion (approved):
LGTM, thx.

bthirion:
> I'm a bit confused about the PEP8 fail here, it's referencing other commits which should not be in this branch.

I'm afraid that the added line is more than 80 characters long ...

jbteves:
It is, but there are many other lines longer than 80 characters. Should I just wrap the text then?

bthirion:
Yes, please ;-)

jbteves:
Should be fixed in 1905526

nilearn/image/image.py
```diff
@@ -697,7 +697,8 @@ def new_img_like(ref_niimg, data, affine=None, copy_header=False):
     Returns
     -------
     Niimg-like object
-        A loaded image with the same type (and header) as the reference image.
+        A loaded image with the same file type (and, optionally, header)
+    as the reference image.
```
NicolasGensollen, line 701: Could you indent this line?

NicolasGensollen (commented):
Thanks @jbteves !

nilearn/image/image.py
jbteves: Should be resolved in 7c2f93c

NicolasGensollen (approved):
LGTM, thanks @jbteves !

NicolasGensollen:
Thanks, merging...

PR #3174, opened 2022-03-09
[FIX] Fix `fetch_abide_pcp`

### LINKED ISSUE #3160: `fetch_abide_pcp` returns empty results
since #2829 `fetch_abide_pcp` returns empty results. This is due to the fact that the filters applied to phenotypic data expect this array to contain `bytes`: https://github.com/nilearn/nilearn/blob/1ebf6b70d79d88e6c3de6d9799c79309c6c99277/nilearn/datasets/func.py#L974  as this is what `np.recfromcsv` returns but it is now `str` since `pd.read_csv` is now used: https://github.com/nilearn/nilearn/blob/1ebf6b70d79d88e6c3de6d9799c79309c6c99277/nilearn/datasets/func.py#L1000

NicolasGensollen:
Fixes #3160

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,6 +13,7 @@ NEW
 Fixes
 -----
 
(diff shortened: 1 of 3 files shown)


bthirion (commented):
Thx for taking care of that !

bthirion (approved):

emdupre (approved):
This LGTM, thank you @NicolasGensollen !

PR #3179, opened 2022-03-17
Expose filename in fetch atlas

### LINKED ISSUE #3088: `fetch_atlas_harvard_oxford` returns same object for `maps` and `filename`
This is also true for `fetch_atlas_juelich` since they both share the same internal logic.
`maps` should indeed be the `Nifti1Image`, but we should either remove `filename` or change it to be the path to the `Nifti1Image`.

Also, this is the only fetcher which returns a `Bunch` object without a `description`.

```python
from nilearn import datasets
b = datasets.fetch_atlas_harvard_oxford("cort-maxprob-thr0-1mm")
assert b.maps == b.filename
b.filename
```
```
<nibabel.nifti1.Nifti1Image at 0x7f1ff37a4280>
```
```python
b.maps
```
```
<nibabel.nifti1.Nifti1Image at 0x7f1ff37a4280>
```

RaphaelMeudec:
Closes #3088 .

- Juelich and Oxford atlases now have the rightful filename in their `filename` attribute instead of the nifti image

### DIFF
--- a/nilearn/datasets/tests/test_atlas.py
+++ b/nilearn/datasets/tests/test_atlas.py
@@ -173,6 +173,7 @@ def _write_sample_atlas_metadata(ho_dir, filename, is_symm):
 
 
 def _test_atlas_instance_should_match_data(atlas, is_symm):
+    assert isinstance(atlas.filename, str) or isinstance(atlas.filename, Path)
     assert isinstance(atlas.maps, nibabel.Nifti1Image)
     assert isinstance(atlas.labels, list)
 
--- a/nilearn/datasets/tests/test_atlas.py
+++ b/nilearn/datasets/tests/test_atlas.py
@@ -150,7 +150,7 @@ def test_fetch_atlas_source(tmp_path, request_mocker):
         atlas._get_atlas_data_and_labels('new_source', 'not_inside')
 
 
-def _write_to_xml(ho_dir, filename, is_symm):
+def _write_sample_atlas_metadata(ho_dir, filename, is_symm):
     with open(os.path.join(ho_dir, filename + '.xml'), 'w') as dm:
         if(not is_symm):
             dm.write("<?xml version='1.0' encoding='us-ascii'?>\n"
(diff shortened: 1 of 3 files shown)


nilearn/datasets/tests/test_atlas.py
```diff
@@ -242,17 +233,29 @@ def atlas_data():
...
         target_atlas_nii)
-    ho_wo = fsl_fetcher(fname, data_dir=str(tmp_path), symmetric_split=split)
-    _test_result_xml(ho_wo, is_symm=is_symm or split)
+    # Check that the fetch lead to consistent results
+    atlas_instance = fsl_fetcher(
+        fname,
+        data_dir=tmp_path,
+        symmetric_split=split,
+    )
+    _test_atlas_instance_should_match_data(
```
NicolasGensollen, line 255: I might have missed something but it seems like you're not testing the returned filename value which is the main change introduced by this PR.

NicolasGensollen (commented):
Thanks for taking care of this @RaphaelMeudec ! 👍 
Looks good to me, I just have a small question below.
Also, could you add an entry to the change log?

nilearn/datasets/tests/test_atlas.py
RaphaelMeudec: Good catch, there was indeed an intermediate `return` I didn't see, so it was useful!

NicolasGensollen:
> Also, I added a line in the whatsnew entry, is there something else to change (when you talk about the change log)?

I was talking about the usual entry in `doc/changes/latest.rst`, but I don't see it in the diff. Are you sure you committed it?

RaphaelMeudec:
@NicolasGensollen The push failed apparently yesterday, probably with the Github trouble. I just pushed it now!

NicolasGensollen (approved):
LGTM, thanks @RaphaelMeudec !

bthirion (approved):
Thx !

NicolasGensollen:
Thanks! Merging...

bthirion:
It looks that this PR has broken the CI. @RaphaelMeudec can you check that ? Thx,

Issue #2366, opened 2015-09-10
Add cluster-level tests and TFCE


bthirion:
This requires a TFCE implementation and non-parametric tests from nilearn/nistats#26.

kingjr:
@bthirion not sure if this is still relevant, but we have a pure python spatio (-temporal) test  implemented in MNE, including tfce which could be directly re-used here for e.g. second-level mass univariate stats and searchlights:

https://mne.tools/0.15/auto_tutorials/plot_stats_cluster_methods.html

bthirion:
Without going too much into detail, I think it is relevant indeed.

tsalo:
Would this be implemented in `permuted_ols` or `SecondLevelModel`?

bthirion:
Maybe `permuted_ols` is the right place.

tsalo:
In that case, I have two questions: first, can `permuted_ols` be extended to work on contrasts? Second, would we just want to calculate and retain maximum cluster mass and/or size, as well as TFCE values, within `_permuted_ols_on_chunk`? If so, that seems fairly straightforward, as long as we calculate TFCE correctly.

bthirion:
I think that the `permuted_ols` does not natively work on contrasts (tbc).
We'll probably have to rewrite a function similar to `_permuted_ols_on_chunk` but taking into account spatial context.
And indeed we need a TFCE implementation (I must have that somewhere).

bthirion:
Sorry to be late on this.
Regarding 3), I would do it optionally, as it probably makes the method less efficient computationally. I'l look at it in #3181

tsalo:
Once #3181 and #3196 are merged, I think all that's left is to support contrasts in `permuted_ols`. Does that sound right?

bthirion:
I've realized recently that contrasts testing was actually more tricky than expected. Since this requires more developments, I would not add it now.

Issue #3180, opened 2022-03-17
permuted_ols only returns h0_fmax for first regressor


tsalo:
When there is more than one regressor in `tested_vars`, `permuted_ols` returns only the null distribution of maximum t-statistics for the first regressor.

Here is the relevant code:

https://github.com/nilearn/nilearn/blob/e7804ebe858349781674c0d2b5c78e27afb52fa1/nilearn/mass_univariate/permuted_least_squares.py#L512

Nilearn version: `main` as of e7804ebe858349781674c0d2b5c78e27afb52fa1.

### Expected behavior

When `n_regressors` is >1, I would expect `h0_fmax` to have that number of rows.

### Actual behavior

`h0_fmax` is a 1D array, with one element per 

### Steps and code to reproduce bug

Here is some code, adapted from one of the permuted_ols tests, that should reproduce the bug.

```python
from sklearn.utils import check_random_state
from nilearn.mass_univariate import permuted_ols

random_state = 0
rng = check_random_state(random_state)

# design parameters
n_samples = 50
n_descriptors = 10
n_regressors = 2
n_covars = 2
n_perm = 10

# create design
target_vars = rng.randn(n_samples, n_descriptors)
tested_vars = rng.randn(n_samples, n_regressors)
confounding_vars = rng.randn(n_samples, n_covars)

neg_log10_pvals, own_scores, h0_fmax = permuted_ols(
    tested_vars,
    target_vars,
    confounding_vars,
    model_intercept=False,
    n_perm=n_perm,
    random_state=random_state,
)
assert own_scores.shape == (n_regressors, n_descriptors)
assert neg_log10_pvals.shape == (n_regressors, n_descriptors)
assert h0_fmax.shape == (n_regressors, n_perm)
# AssertionError, assert (10,) == (2, 10)
```

bthirion:
But the problem is that the permutation scheme we use is invalid if we are to test several covariates (the error rate will not be controlled jointly for the two or more tests if we do them concurrently, because the model will test against a global null).
I think we should document the problem clearly, making this  a documentation issue.

tsalo:
Should it just not support multiple `tested_vars` then? It seems like it should have a single `tested_var` and anything else should go into `confounding_vars`.

bthirion:
Sorry, my comment above was wrong, because test_vars are tested independetly.
But then you're right, the output of the model is wrong, `p_values` shoudl have shape `(n_regressors, n_descriptors)`

PR #3181, opened 2022-03-18
[ENH] Implement cluster-level correction in `permuted_ols`

tsalo:
Stems from #2366, but doesn't close it.

This is currently just a rough draft, with a lot of the new code taken from NiMARE, which implements cluster-level thresholding for coordinate-based meta-analyses.

Adding cluster-level correction will slow `permuted_ols` down, so we may want to add a parameter for vFWE-only correction, which would be what is currently done, for folks who want to run `permuted_ols` as quickly as possible.

- Add `masker` and `threshold` (cluster-forming threshold) parameters to `permuted_ols` to allow cluster-level FWE correction.
- Add `threshold` parameter to `non_parametric_inference` for the same reason. It already has a `mask` parameter.
- Implement both cluster size and cluster mass-based FWE correction within `permuted_ols`, `_permuted_ols_on_chunk`, and `non_parametric_inference`.
    - Three caveats though:
        1. This probably won't work with anything other than a NiftiMasker at the moment.
            - I'm not sure how much of a problem this is, since `non_parametric_inference` only allows NiftiMaskers and MultiNiftiMaskers anyway.
        2. ~~The NiftiMasker resamples the cluster-level corrected maps in an inappropriate manner. It looks like it uses the default "continuous" interpolation, but cluster-level corrected maps should use "nearest", since each _cluster_ has its own statistical value.~~
            - This actually turned out to be because the NiftiMasker in the example I was testing with has smoothing enabled. I got around this by using `apply_mask` instead. I'm not sure if there's a better approach.
        3. I don't have `two_sided_test` working yet.
- Add some functions from NiMARE that are necessary for the cluster-level correction.
    - `_null_to_p` convert arbitrary values to p-values based on a null distribution.
    - `_calculate_cluster_measures` calculates maximum cluster size and maximum cluster mass for 3D numpy arrays (updated to loop through the fourth dim for 4D arrays).
- Add new outputs to `permuted_ols` and `non_parametric_inference`.
- Fix subject labeling in `plot_second_level_one_sample_test.py`.

## To do

- [x] Figure out a better output structure for the two functions. This PR will add ~4 outputs (2 H0s and 2 corrected p-value maps), which overloads the function with outputs, IMHO.
- [x] Determine if positive and negative null distributions need to be built separately.
- [x] Fix the Codespell issue.
- [x] Tests!


nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -265,12 +460,27 @@ def _permuted_ols_on_chunk(scores_original_data, tested_vars, target_vars, threa
...
+    )
+
+
+def permuted_ols(
+    tested_vars,
+    target_vars,
+    confounding_vars=None,
+    model_intercept=True,
+    masker=None,
+    cdt=1.,
```
tsalo, line 477: I think I'd prefer to have the cluster-defining threshold be a p-value instead of a t-statistic. Having a default t-statistic isn't very useful, given that the appropriate value depends on the degrees of freedom.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -291,111 +501,131 @@ def permuted_ols(tested_vars, target_vars, confounding_vars=None,
...
+        target variates.
+        The ranks of the scores into the h0 distribution correspond to the
+        p-values.
 
-    h0_fmax : array-like, shape=(n_perm, )
-      Distribution of the (max) t-statistic under the null hypothesis
-      (obtained from the permutations). Array is sorted.
+    h0_fmax : array-like, shape=(n_regressors, n_perm)
+        Distribution of the (max) t-statistic under the null hypothesis
+        (obtained from the permutations). Array is sorted.
```
tsalo, line 578: This change is related to #3180. I will probably revert it here, but it's worth noting.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -291,111 +501,131 @@ def permuted_ols(tested_vars, target_vars, confounding_vars=None,
...
-        raise ValueError("'n_jobs == 0' is not a valid choice. "
-                         "Please provide a positive number of CPUs, or -1 "
-                         "for all CPUs, or a negative number (-i) for "
-                         "'all but (i-1)' CPUs (joblib conventions).")
+        raise ValueError(
+            "'n_jobs == 0' is not a valid choice. "
+            "Please provide a positive number of CPUs, or -1 for all CPUs, "
+            "or a negative number (-i) for 'all but (i-1)' CPUs "
+            "(joblib conventions)."
+        )
```
tsalo, line 603: This conflicts with the docstring, which says that `n_jobs == 0` will use all CPUs.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -470,43 +715,155 @@ def permuted_ols(tested_vars, target_vars, confounding_vars=None,
...
+        cmfwe_h0_parts,
+    ) = zip(*ret)
+
+    # Voxel-level FWE
+    vfwe_h0 = np.hstack((vfwe_h0_parts))
+    vfwe_scores_as_ranks = np.zeros((n_regressors, n_descriptors))
+    for vfwe_scores_as_ranks_part in vfwe_scores_as_ranks_parts:
+        vfwe_scores_as_ranks += vfwe_scores_as_ranks_part
+
+    vfwe_pvals = (n_perm + 1 - vfwe_scores_as_ranks) / float(1 + n_perm)
```
tsalo, line 770: The generic "pvals" from before are now specifically referred to as "vfwe_pvals" to distinguish them from the other corrected p-value outputs.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -470,43 +715,155 @@ def permuted_ols(tested_vars, target_vars, confounding_vars=None,
...
+        # Convert 3D to image, then to 1D
+        cmfwe_pvals[i_regressor, :] = np.squeeze(
+            masker.transform(
+                nib.Nifti1Image(
+                    p_cmfwe_map,
+                    masker.mask_img_.affine,
+                    masker.mask_img_.header,
+                )
+            )
+        )
```
tsalo, line 836: I am not sure if this is the best way to go from 3D to 1D. I can use something like `new_img_like`, but I worry that suffers from the same limitations (i.e., only works for NiftiMaskers).

nilearn/mass_univariate/permuted_least_squares.py
tsalo: The default p-value threshold would, of course, be 0.001.

nilearn/mass_univariate/permuted_least_squares.py
tsalo: I opened #3182 to address this.

.github/codespell_ignore_words.txt
```diff
@@ -8,3 +8,4 @@ sepulcre
 mapp
 nin
 serie
+FWE
```
tsalo, line 11: This doesn't seem to have worked... 😕

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -8,11 +8,161 @@
...
+
+        labeled_arr3d, _ = ndimage.measurements.label(arr3d > 0, conn)
+
+        if two_sided_test:
+            # Label positive and negative clusters separately
+            n_positive_clusters = np.max(labeled_arr3d)
+            temp_labeled_arr3d, _ = ndimage.measurements.label(arr3d < 0, conn)
+            temp_labeled_arr3d[temp_labeled_arr3d > 0] += n_positive_clusters
+            labeled_arr3d = labeled_arr3d + temp_labeled_arr3d
+            del temp_labeled_arr3d
```
tsalo, line 141: This distinguishes negative and positive values in cluster assignment. However, I have read in a few places that a two sided test should retain the positive and negative maximum values in _separate_ null distributions. For example, in [Albajes-Eizagirre et al. (2019)](https://pubmed.ncbi.nlm.nih.gov/30389629/):

>SDM-PSI can currently save four different maximum statistics from the image of z-values: the largest z-value (i.e., voxel-based statistics), the maximum cluster size or mass after thresholding with a used-defined z- value (i.e., cluster-size or mass statistics) (Bullmore et al., 1999), and the maximum TFCE (Smith and Nichols, 2009). We have implemented TFCE in our software so that users who do not have FSL installed will still be able to use TFCE in SDM-PSI.
>This procedure theoretically only tests hypothesis in one direction (e.g., patients > controls), but not the hypothesis in the other direction (e.g., patients < controls), and thus, the procedure should be conducted twice, one for each direction. However, given that the hypotheses are complementary and a permutation test is computationally consuming, SDM-PSI saves two numbers from each permutation: one for the positive hypothesis (e.g., the highest z-value) and one for the negative hypothesis (e.g., the lowest z-value), in two separate null distributions.

Should the vFWE, csFWE, and cmFWE null distributions be split by sign?

nilearn/mass_univariate/permuted_least_squares.py
tsalo: I replaced `cdt` with a p-value `threshold` parameter. I need to take a second look at the calculation of the minimum t-statistic though.

tsalo:
Here's what the current results look like for [this example](https://nilearn.github.io/stable/auto_examples/05_glm_second_level/plot_second_level_one_sample_test.html#sphx-glr-auto-examples-05-glm-second-level-plot-second-level-one-sample-test-py). I assume the huge colorbars are due to the matplotlib version, per [this NeuroStars post](https://neurostars.org/t/a-strange-bug-with-a-huge-colorbar/21903/4).

image

nilearn/mass_univariate/permuted_least_squares.py
tsalo: Resolving this and reverting the shape documentation change in favor of discussing in #3180.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -297,82 +494,136 @@ def permuted_ols(
...
-    A max-type procedure is used to obtain family-wise corrected p-values.
+    A max-type procedure is used to obtain family-wise corrected p-values
+    based on t-statistics (voxel-level FWE), cluster sizes, and cluster masses.
 
-    The specific permutation scheme implemented here is the one of
-    [3]_. Its has been demonstrated in [1]_ that this
-    scheme conveys more sensitivity than alternative schemes. This holds
-    for neuroimaging applications, as discussed in details in [2]_.
+    The specific permutation scheme implemented here is the one of [3]_.
+    It's has been demonstrated in [1]_ that this scheme conveys more
```
tsalo, line 501:
```suggestion
    It has been demonstrated in [1]_ that this scheme conveys more
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -446,92 +652,234 @@ def permuted_ols(
...
+        cmfwe_h0 = np.hstack((cmfwe_h0_parts))
+
+        csfwe_pvals = np.zeros_like(vfwe_pvals)
+        cmfwe_pvals = np.zeros_like(vfwe_pvals)
+
+        # TODO: Eliminate need to transpose
+        scores_original_data_4d = masker.inverse_transform(
+            scores_original_data.T
+        ).get_fdata()
+        conn = ndimage.generate_binary_structure(3, 1)
```
bthirion, line 787: I think that in other parts of the library , we use a different définition for `conn`.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -446,92 +652,234 @@ def permuted_ols(
...
+    if output_type == 'legacy':
+        outputs = (-np.log10(vfwe_pvals), scores_original_data.T, vfwe_h0)
+    else:
+        outputs = {
+            't': scores_original_data.T,
+            'logp_max_t': -np.log10(vfwe_pvals),
+            'h0_max_t': vfwe_h0,
+        }
+
+        if threshold is not None:
```
bthirion, line 879: Just wondering: shouldn't cluster sizes/mass also not be part of the output, as is `t` for voxel-level statistics ?

bthirion (commented):
Thx. Shouldn't we demo all this in an example ?

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,153 @@
...
+        If 'lower', then lower values for the test_value are more significant.
+        Default is 'two'.
+    symmetric : :obj:`bool`, optional
+        When tail="two", indicates how to compute p-values.
+        When False (default), both one-tailed p-values are computed,
+        and the two-tailed p is double the minimum one-tailed p.
+        When True, it is assumed that the null distribution is zero-centered
+        and symmetric, and the two-tailed p-value is computed as
+        P(abs(test_value) >= abs(null_array)).
+        Default=False.
```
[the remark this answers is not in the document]
tsalo: This is a function I grabbed from NiMARE where there are cases of asymmetric null distributions (e.g., meta-analytic subtraction analyses). I can simplify this function for nilearn.

nilearn/mass_univariate/permuted_least_squares.py
tsalo: Sorry, this question confused me a bit. Could you expand on it?

nilearn/mass_univariate/_utils.py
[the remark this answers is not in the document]
bthirion: Side point: when the null distribution is asymmetric, it is likely that you're only interested in one-sided test (eg for max cluster size).

nilearn/mass_univariate/permuted_least_squares.py
bthirion: Sorry,
For voxel level inference, you're returning the `t` values, together with null distribution and p-values.
For cluster level inference, why not returning the cluster size statistic (as an image), together with he null distribution of max cluster sizes and p-values ?
Or did I miss something ?

nilearn/mass_univariate/permuted_least_squares.py
tsalo: :+1: I'll change it to `bin_struct` (for binary structure).

examples/05_glm_second_level/plot_second_level_one_sample_test.py
```diff
@@ -107,32 +129,60 @@
...
-plotting.show()
+IMAGES = [
+    neg_log_pval,
+    neg_log10_vfwe_pvals_permuted_ols_unmasked,
+    neg_log10_csfwe_pvals_permuted_ols_unmasked,
+    neg_log10_cmfwe_pvals_permuted_ols_unmasked,
+]
+TITLES = [
+    "Parametric Test",
+    "Permutation Test\n(Voxel-Level FWE)",
```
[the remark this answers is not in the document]
tsalo: That works for me. I've changed all three over.

nilearn/mass_univariate/permuted_least_squares.py
tsalo: Ah, okay. I've added that in [f1eebf5](https://github.com/nilearn/nilearn/pull/3181/commits/f1eebf514df388f3ed94a18e3be0d417bfdedc87).

tsalo:
> Thx. Shouldn't we demo all this in an example ?

The new outputs are included in `05_glm_second_level/plot_second_level_one_sample_test.py`. Are there any other examples where you'd like to see the new features?

doc/changes/latest.rst
```diff
@@ -9,6 +9,7 @@
 NEW
 ---
 
+- :func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` now support cluster-level FWE correction (:gh:`3181` by `Taylor Salo`_).
```
bthirion, line 12:
```suggestion
- :func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` now support cluster-level Family-wise error correction (:gh:`3181` by `Taylor Salo`_).
```

examples/05_glm_second_level/plot_second_level_one_sample_test.py
```diff
@@ -107,32 +129,57 @@
...
+    second_level_input,
+    design_matrix=design_matrix,
+    model_intercept=True,
+    n_perm=1000,
+    two_sided_test=False,
+    smoothing_fwhm=8.0,
+    n_jobs=1,
+    threshold=0.001,
+)
+
```
bthirion, line 147: The example now lasts several minutes on my box.  My impression is that it is significantly longer now. I'd do 3 things:
* state explicitly that performing cluster-level inference has a cost in terms of computation time
* (as for tfce) reduce the number of permutations (explaining that this is bad)
* explain that  n_jobs can be chosen to reduce cpu time.

nilearn/mass_univariate/permuted_least_squares.py
bthirion: Sorry, what I meant is that the specification of the structure is different e.g. in 
https://github.com/nilearn/nilearn/blob/0e678341aee2866527d99fb9344cfa5ae56aa7e3/nilearn/image/image.py#L793
I think we should strive to have only one definition across the library and always use it.

tsalo: ohhhhh okay. I'll change the other cases then.

nilearn/mass_univariate/tests/test__utils.py
```diff
@@ -0,0 +1,59 @@
...
+from nilearn.mass_univariate._utils import _null_to_p
+
+
+def test_null_to_p_float():
+    """Test _null_to_p with single float input."""
+    null = [-10, -9, -9, -3, -2, -1, -1, 0, 1, 1, 1, 2, 3, 3, 4, 4, 7, 8, 8, 9]
+
+    # Left/lower-tailed
+    assert math.isclose(_null_to_p(9, null, alternative='smaller'), 0.95)
+    assert math.isclose(_null_to_p(-9, null, alternative='smaller'), 0.15)
```
bthirion, line 14:
```suggestion
    assert math.isclose(_null_to_p(-9, null, alternative='smaller'), 0.05)
```

nilearn/mass_univariate/tests/test__utils.py
```diff
@@ -0,0 +1,59 @@
...
+    assert math.isclose(_null_to_p(0, null, alternative='smaller'), 0.4)
+
+    # Right/upper-tailed
+    assert math.isclose(_null_to_p(9, null, alternative='larger'), 0.05)
+    assert math.isclose(_null_to_p(-9, null, alternative='larger'), 0.95)
+    assert math.isclose(_null_to_p(0, null, alternative='larger'), 0.65)
```
bthirion, line 20:
```suggestion
    assert math.isclose(_null_to_p(0, null, alternative='larger'), 0.5)
```

nilearn/mass_univariate/permuted_least_squares.py
tsalo: @bthirion do you think tracking positive and negative distributions separately is necessary, or is it related to https://github.com/nilearn/nilearn/pull/3181#discussion_r840718578 in that we should assume the null distribution is zero-centered and symmetric, and just the maximum absolute value?

bthirion: No, I would consider only the absolute value. I'm not aware of any case with two-sided tests where asymmetry would matter.

bthirion: Note that in he above paper if the null distribution is obtained by permutation, it has to be symmetric ...

nilearn/mass_univariate/tests/test__utils.py
tsalo: This won't pass though. There are 3/20 (0.15) elements in the null array that are less than or equal to -9.

bthirion: Sorry, I misunderstood the test.

nilearn/mass_univariate/permuted_least_squares.py
tsalo: At least in ALE meta-analyses, the test statistic scales with sample size, so comparing two groups with different numbers of studies will naturally lead to a null distribution that is not zero-centered. Though, I suppose since z-statistics were used in that case, it would end up being symmetric.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -244,38 +287,71 @@ def _permuted_ols_on_chunk(
...
-            perm_scores = np.fabs(perm_scores)
-        h0_fmax_part[i] = np.nanmax(perm_scores, axis=0)
-        # find the rank of the original scores in h0_part
+            _t_score_with_covars_and_normalized_design(
+                tested_vars, target_vars, confounding_vars
+            )
+        )
+
+        if threshold is not None:
+            # TODO: Eliminate need for transpose
```
bthirion, line 296: to be done ?

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -244,38 +287,71 @@ def _permuted_ols_on_chunk(
...
+            arr4d = masker.inverse_transform(perm_scores.T).get_fdata()
+            bin_struct = ndimage.generate_binary_structure(3, 1)
```
bthirion, line 298: can be defined outside of the loop.

bthirion (commented):
Just tiny details pending.
I'm thinking that we should add a bit of narrative doc to explain  the zoology of inference models (cluster size, cluster mess, one-and two-sided, parametric and non-parametric etc.)

nilearn/mass_univariate/permuted_least_squares.py
tsalo: It looks like the array is transposed back and forth in a couple of places. I think the code would be cleaner if it didn't need to be, but I don't want to tackle that in this PR. Should I leave a comment as a note for future refactoring or just remove it?

nilearn/mass_univariate/permuted_least_squares.py
tsalo: Thanks for catching that.

examples/05_glm_second_level/plot_second_level_one_sample_test.py
```diff
@@ -107,32 +129,65 @@
...
+            figure=fig,
+            axes=ax,
+        )
+        ax.set_title(TITLES[img_counter])
+        img_counter += 1
+
+fig.suptitle("Group left-right button press\n(negative log10 p-values)")
 
 # The neg-log p-values obtained with nonparametric testing are capped at 3
 # since the number of permutations is 1e3.
```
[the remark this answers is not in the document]
tsalo: I've added the plt.show(). I also reorganize the comments a little bit, since they no longer matched up well with the new plot. The new vmax is 2.7 to match the new number of permutations.

tsalo:
>I'm thinking that we should add a bit of narrative doc to explain the zoology of inference models (cluster size, cluster mess, one-and two-sided, parametric and non-parametric etc.)

Would the user guide be the place for that?

nilearn/mass_univariate/permuted_least_squares.py
bthirion: I'd just remove it.

tsalo: Done.

bthirion:
> > I'm thinking that we should add a bit of narrative doc to explain the zoology of inference models (cluster size, cluster mess, one-and two-sided, parametric and non-parametric etc.)
> 
> Would the user guide be the place for that?

Yes, I think so. This can definitely go to another PR.

examples/05_glm_second_level/plot_second_level_one_sample_test.py
```diff
@@ -152,7 +152,12 @@
 # This threshold is much more conservative than an uncorrected threshold, but
 # is still more liberal than a typical corrected threshold for this kind of
 # analysis, which tends to be ~0.05.
-threshold = 1  # p<0.1
+#
+# We will also cap the negative log10 p-values at 2.7, because this is the
+# maximum observable value for the nonparametric tests, which were run with
+# only 500 permutations.
+threshold = 1  # p < 0.1
+vmax = 2.7  # ~= -np.log10(1 / 500)
```
bthirion, line 160: Actually should be slightly below (2.69).

bthirion (commented):
LGTM otherwise.

tsalo:
>This can definitely go to another PR.

That's good. I started drafting something and realized I wouldn't be able to throw something together easily. A separate PR sounds great.

examples/05_glm_second_level/plot_second_level_one_sample_test.py
tsalo: Done in [1e7bc40](https://github.com/nilearn/nilearn/pull/3181/commits/1e7bc40628b701735bd58b97082e1ba6493d314a). Thanks!

bthirion (approved):
LGTM, thx !

examples/05_glm_second_level/plot_second_level_one_sample_test.py
```diff
@@ -94,47 +115,90 @@
...
+    '-np.log10(np.minimum(1, img * {}))'.format(str(n_voxels)),
+    img=p_val,
+)
 
-###########################################################################
+###############################################################################
 # Now, we compute the (corrected) p-values with a permutation test.
+#
+# .. important::
+#   In this example, ``threshold`` is set to 0.001, which enables
```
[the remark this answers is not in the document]
emdupre: Ah I see (reading the function docstring further down) ! Is there a reason not to first call `non_parametric_inference` with `threshold` set to `None` and then again as is done here (with a provided `threshold` value) ?

nilearn/glm/second_level/second_level.py
```diff
@@ -650,77 +662,133 @@ def non_parametric_inference(second_level_input, confounds=None,
...
+
+        =============== =======================================================
+        key             description
+        =============== =======================================================
+        t               T-statistics associated with the significance test of
+                        the n_regressors explanatory variates against the
+                        n_descriptors target variates.
+        logp_max_t      Negative log10 family-wise error rate-corrected
+                        p-values corrected based on the distribution of maximum
+                        t-statistics from permutations.
```
emdupre, line 774: These are also only returned if `threshold` is not `None` correctly ? You explicitly include that in the description of the other two keys, so I just wanted to check !

nilearn/glm/second_level/second_level.py
```diff
@@ -773,12 +843,38 @@ def non_parametric_inference(second_level_input, confounds=None,
...
+        output_type='dict',
+    )
+    neg_log10_vfwe_pvals = outputs['logp_max_t']
+    neg_log10_vfwe_pvals_img = masker.inverse_transform(
+        np.ravel(neg_log10_vfwe_pvals),
+    )
+
+    if threshold is not None:
+        t_img = masker.inverse_transform(np.ravel(outputs['t']))
+        neg_log10_csfwe_pvals_img = masker.inverse_transform(
```
emdupre, line 866: what is `csfwe` short for ? I saw `vfwe` defined in the doc, but it would be nice to spell out these abbreviations (at least once) in the comments to help orient folks to the code !

nilearn/glm/second_level/second_level.py
tsalo: This is the standard voxel-level-corrected map, so it will be returned as long as `n_iters > 0`.

examples/05_glm_second_level/plot_second_level_one_sample_test.py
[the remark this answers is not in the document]
tsalo: Do you mean just to show how `non_parametric_inference` would work with and without cluster-level inference? If we ran it without cluster-level inference, it would take some additional time to run the voxel-level-only permutations, and we get those for free when we use `threshold`. Sorry if I'm misunderstanding your question.

nilearn/glm/second_level/second_level.py
tsalo: It stands for cluster size-based FWE-corrected. I can add a comment.

examples/05_glm_second_level/plot_second_level_one_sample_test.py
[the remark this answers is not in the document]
emdupre: No, that's my question :) I think from a UX standpoint this isn't something that's obvious in the example flow. Maybe I'm missing something from the non-rendered files -- I'll try rendering to see if that helps me understand the logical progression a bit more easily.

nilearn/glm/second_level/second_level.py
emdupre: But this object is noted to only be returned when `threshold` is not `None` ? I think that's the source of my confusion !

examples/05_glm_second_level/plot_second_level_one_sample_test.py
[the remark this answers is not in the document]
emdupre: To clarify my question : When we make a general recommendation for how users can correct their p-values _without_ cluster-level inference, will we point them to [this math_img code block](https://github.com/nilearn/nilearn/pull/3181/files/1e7bc40628b701735bd58b97082e1ba6493d314a#diff-ea11adc25ebe65ff700093bf22dc77c490bfce5298bc8d9a26fa9216af068d32R118-R121) ? Or would we point them to the `non_parametric_inference` function and tell them to make sure that `threshold` is set to `None` ?

nilearn/glm/second_level/second_level.py
tsalo: Ah, I see. Yes, in this PR the dictionary output is only returned if threshold is defined. In the TFCE PR (#3196), I add an `output_type` parameter that lets users directly select the new structure.

examples/05_glm_second_level/plot_second_level_one_sample_test.py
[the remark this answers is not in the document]
tsalo: I think the code block you linked to is for the parametric approach instead. I guess I'd point them to the function... but I could add an admonition here about running just voxel-level correction?

nilearn/glm/second_level/second_level.py
tsalo: I added a few comments in [d98d26e](https://github.com/nilearn/nilearn/pull/3181/commits/d98d26e9d443ddda833523e7bb1dae34259e2fb0). WDYT?

examples/05_glm_second_level/plot_second_level_one_sample_test.py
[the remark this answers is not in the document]
tsalo: I added an admonition in c74f809. WDYT?

emdupre (approved):
This LGTM -- thank you, @tsalo !!

tsalo:
@bthirion @emdupre thank you for your reviews. Should I merge?

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -298,87 +376,187 @@ def permuted_ols(
...
+            Performing cluster-level inference will increase the computation
+            time of the permutation procedure.
+
+        .. versionadded:: 0.9.2
+
+    output_type : {'legacy', 'dict'}, optional
+        Determines how outputs should be returned.
+        The two options are:
+
+        -   'legacy': return a pvals, score_orig_data, and h0_fmax
```
NicolasGensollen, line 471: I think we should clarify here that the option `legacy` is deprecated (I would add a `.. deprecated::` directive), and that the parameter `output_type` will default to `dict` from 0.13 (and be removed in 0.15??). WDYT?

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -298,87 +376,187 @@ def permuted_ols(
...
     Returns
     -------
     pvals : array-like, shape=(n_regressors, n_descriptors)
         Negative log10 p-values associated with the significance test of the
         n_regressors explanatory variates against the n_descriptors target
         variates. Family-wise corrected p-values.
 
+        .. note::
+            This is returned if ``output_type`` == 'legacy'.
+
```
NicolasGensollen, line 483: Same comment, I would add a `deprecated` directive for all outputs that we are planning to drop in favor of the dict format.

NicolasGensollen (commented):
Sorry for the delay in coming back to this PR @tsalo 
Taking a look now!

examples/05_glm_second_level/plot_second_level_one_sample_test.py
```diff
@@ -94,47 +115,94 @@
...
+            axes=ax,
+        )
+        ax.set_title(TITLES[img_counter])
+        img_counter += 1
+
+fig.suptitle('Group left-right button press\n(negative log10 p-values)')
+
 # The nonparametric test yields many more discoveries and is more powerful than
 # the usual parametric procedure.
+# Even within the nonparametric test, the different correction metrics produce
```
NicolasGensollen, line 204: These lines render as a comment in the code cell. Maybe we could swap them and the `plt.show()` line above, and make them proper text?

nilearn/glm/second_level/second_level.py
```diff
@@ -650,77 +662,133 @@ def non_parametric_inference(second_level_input, confounds=None,
...
+
+            Performing cluster-level inference will increase the computation
+            time of the permutation procedure.
+
+        .. versionadded:: 0.9.2
+
     Returns
     -------
-    neg_log_corrected_pvals_img : Nifti1Image
+    neg_log10_vfwe_pvals_img : Nifti1Image
```
NicolasGensollen, line 747:
```suggestion
    neg_log10_vfwe_pvals_img : :class:`~nibabel.nifti1.Nifti1Image`
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -176,9 +182,23 @@ def _permuted_ols_on_chunk(
...
+        Cluster-forming threshold in t-scale.
+        This is only used for cluster-level inference.
+        If ``masker`` is set to None, it will be ignored.
+
+        .. versionadded:: 0.9.2
+
     confounding_vars : array-like, shape=(n_samples, n_covars), optional
         Clinical data (covariates).
 
+    masker : None or NiftiMasker or MultiNiftiMasker object, optional
```
NicolasGensollen, line 195:
```suggestion
    masker : None or :class:`~nilearn.maskers.NiftiMasker` or :class:`nilearn.maskers.MultiNiftiMasker`, optional
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -207,10 +227,25 @@ def _permuted_ols_on_chunk(
 
     Returns
     -------
-    h0_fmax_part : array-like, shape=(n_perm_chunk, )
+    scores_as_ranks_part : array-like, shape=(n_regressors, n_descriptors)
+        The ranks of the original scores in h0_fmax_part.
```
NicolasGensollen, line 231:
```suggestion
        The ranks of the original scores in ``h0_fmax_part``.
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -298,87 +376,187 @@ def permuted_ols(
...
         Number of parallel workers.
         If -1 is provided, all CPUs are used.
         A negative number indicates that all the CPUs except (abs(n_jobs) - 1)
         ones will be used. Default=1.
 
-    verbose : int, optional
+    verbose : :obj:`int`, optional
         verbosity level (0 means no message). Default=0.
 
+    masker : None or NiftiMasker or MultiNiftiMasker object, optional
```
NicolasGensollen, line 447:
```suggestion
    masker : None or :class:`~nilearn.maskers.NiftiMasker` or :class: `~nilearn.maskers.MultiNiftiMasker`, optional
```

nilearn/mass_univariate/tests/test__utils.py
```diff
@@ -0,0 +1,155 @@
...
+    p = np.sort(_utils._null_to_p(t, nulldist))
+    assert p.shape == (N,)
+    assert (p < 1).all()
+    assert (p > 0).all()
+    # Resulting distribution should be roughly uniform
+    assert np.abs(p.mean() - 0.5) < 0.02
+    assert np.abs(p.var() - 1 / 12) < 0.02
+
+
+def test__calculate_cluster_measures():
```
NicolasGensollen, line 79:
```suggestion
def test_calculate_cluster_measures():
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -440,92 +650,265 @@ def permuted_ols(
...
+                cmfwe_h0[i_regressor, :],
+                'larger',
+            )
+            p_cmfwe_map = p_cmfwe_vals[np.reshape(idx, labeled_arr3d.shape)]
+            mass_map = cluster_masses[np.reshape(idx, labeled_arr3d.shape)]
+
+            # Convert 3D to image, then to 1D
+            # There is a problem if the masker performs preprocessing,
+            # so we use apply_mask here.
+            cmfwe_pvals[i_regressor, :] = np.squeeze(
```
NicolasGensollen, line 843: There is a lot of similar code here.
I might be wrong, but I think we could factorize a bit with a for loop over something like `zip([cmfwe_pvals, csfwe_pvals], [mass_arr, size_arr])`. WDYT?

examples/05_glm_second_level/plot_second_level_one_sample_test.py
tsalo: Great idea! I've changed it in [f31927e](https://github.com/nilearn/nilearn/pull/3181/commits/f31927e116640f2990301af3f20fb169ea17dce6).

nilearn/mass_univariate/permuted_least_squares.py
tsalo: I've tried to do that in [ce5bc79](https://github.com/nilearn/nilearn/pull/3181/commits/ce5bc79e525f642a319b59d3e80b9c8a534305bb). WDYT?

nilearn/mass_univariate/permuted_least_squares.py
tsalo: Hopefully [ce5bc79](https://github.com/nilearn/nilearn/pull/3181/commits/ce5bc79e525f642a319b59d3e80b9c8a534305bb) handles this.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -298,87 +377,223 @@ def permuted_ols(
...
+        The two options are:
+
+        -   'legacy': return a pvals, score_orig_data, and h0_fmax.
+            This option is the default, but it is deprecated until 0.13,
+            when the default will be changed to 'dict'.
+            It will be removed in 0.15.
+        -   'dict': return a dictionary containing output arrays.
+            This option will be made the default in 0.13.
+
+        .. deprecated:: 0.15
```
NicolasGensollen, line 480:
```suggestion
        .. deprecated:: 0.9.2
```

(otherwise it renders as "deprecated since 0.15".)

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -298,87 +377,223 @@ def permuted_ols(
...
     h0_fmax : array-like, shape=(n_regressors, n_perm)
         Distribution of the (max) t-statistic under the null hypothesis
         (obtained from the permutations). Array is sorted.
 
-        .. versionchanged:: 0.9.1
+        .. note::
+
+            This is returned if ``output_type`` == 'legacy'.
+
+        .. deprecated:: 0.13
```
NicolasGensollen, line 529:
```suggestion
        .. deprecated:: 0.9.2
```

NicolasGensollen (approved):
Thanks for addressing my points @tsalo !
LGTM !

nilearn/mass_univariate/permuted_least_squares.py
tsalo: 🤦 thanks for catching this! I've fixed this and the other instances in 2ab34d0.

bthirion:
Merging now. Thx !

PR #3184, opened 2022-03-22
[FIX] Return `h0_fmax` for all regressors from `permuted_ols`

### LINKED ISSUE #3180: permuted_ols only returns h0_fmax for first regressor

tsalo:
Closes #3180.

- Return `h0_fmax` (the null distribution of maximum t-statistics used for voxel-level FWE correction) for all regressors, instead of just the first regressor.
- Update tests to check for the right shape.
- Add a versionchanged directive to the `permuted_ols` docstring reflecting this change.
- Add a whatsnew entry for the change.

NOTE: It doesn't look like `non_parametric_inference` retains `h0_fmax` from `permuted_ols`, and none of the examples that use `permuted_ols` seem to retain `h0_fmax` either, so I think the scope of these changes is limited to the function and its tests.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,6 +13,7 @@ NEW
 Fixes
 -----
 
+- Fix function :func:`~mass_univariate.permuted_ols`, which was only returning the null distribution (``h0_fmax``) for the first regressor (:gh:`3184` by `Taylor Salo`_).
 - Fix function :func:`~datasets.fetch_abide_pcp` which was returning empty phenotypes and ``func_preproc`` after release ``0.9.0`` due to supporting pandas dataframes in fetchers (:gh:`3174` by `Nicolas Gensollen`_).
 - Fix function :func:`~datasets.fetch_atlas_harvard_oxford` and :func:`~datasets.fetch_atlas_juelich` which were returning the image in the `filename` attribute instead of the path to the image (:gh:`3179` by `Raphael Meudec`_).
 
--- a/nilearn/mass_univariate/permuted_least_squares.py
+++ b/nilearn/mass_univariate/permuted_least_squares.py
@@ -351,7 +351,7 @@ def permuted_ols(tested_vars, target_vars, confounding_vars=None,
       The ranks of the scores into the h0 distribution correspond to the
       p-values.
 
-    h0_fmax : array-like, shape=(n_perm, )
+    h0_fmax : array-like, shape=(n_regressors, n_perm)
       Distribution of the (max) t-statistic under the null hypothesis
       (obtained from the permutations). Array is sorted.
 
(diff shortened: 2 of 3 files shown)


bthirion (commented):
LGTM, thx !

NicolasGensollen (approved):
The changes LGTM! Thanks @tsalo ! 👍 
Without the CircleCI issue, I would have asked for a full doc build to check that we are not breaking an example. Did you check locally that examples relying on `permuted_ols` were running fine?

tsalo:
@NicolasGensollen it looks like the only examples that use `permuted_ols` are:

- examples/02_decoding/plot_oasis_vbm.py
- examples/07_advanced/plot_haxby_mass_univariate.py
- examples/07_advanced/plot_localizer_mass_univariate_methods.py

None of the three actually uses `h0_fmax`. They all just assign that output to `_`. `nilearn.glm.second_level.non_parametric_inference` does the same thing, and I believe that's the only _function_ that uses `permuted_ols`, so I think the impact of the change is minimal.

bthirion (approved):
LGTM, thx.

emdupre (approved):

NicolasGensollen:
Thanks @tsalo ! 👍 
Merging then, since we have 3 approvals.

PR #3186, opened 2022-03-23
Fix documentation build

RaphaelMeudec:
A bug has been introduced by myself in #3179 leading to the documentation not building anymore in the CI. This PR fixes it.

Reason for the bug:

- in #3179, I changed the behaviour of atlas fetching for Juelich and Oxford: before the `.filename` element contained the image element instead of the path, the purpose of my PR was to switch this image element to the appropriate element path
- doing so, I made the method return a relative path instead of the absolute path of the element, returning path like `data/atlases/Juelich/Juelich-maxprob-thr0-1mm.nii.gz`
- the problem is that during `fetch_files`, this relative path is added to the data_dir, which is generally `$HOME/nilearn_data/fsl`
- this lead the documentation build to look for the atlas image in the current directory instead of `$HOME/nilearn_data/fsl`, causing the bug

What's in the fix:

- I fixed it by returning the actual absolute path
- I improved the test to check whether the path did exist + was absolute

Sorry for the trouble!

### DIFF
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -593,13 +593,13 @@ def _get_atlas_data_and_labels(atlas_source, atlas_name, symmetric_split=False,
     atlas_file = os.path.join(root, atlas_source,
                               '{}-{}.nii.gz'.format(atlas_source,
                                                     atlas_name))
-    atlas_img, label_file = _fetch_files(
+    atlas_file, label_file = _fetch_files(
         data_dir,
         [(atlas_file, url, opts),
          (label_file, url, opts)],
         resume=resume, verbose=verbose)
     # Reorder image to have positive affine diagonal
-    atlas_img = reorder_img(atlas_img)
+    atlas_img = reorder_img(atlas_file)
     names = {}
     from xml.etree import ElementTree
     names[0] = 'Background'
--- a/nilearn/datasets/tests/test_atlas.py
+++ b/nilearn/datasets/tests/test_atlas.py
@@ -173,7 +173,7 @@ def _write_sample_atlas_metadata(ho_dir, filename, is_symm):
 
 
 def _test_atlas_instance_should_match_data(atlas, is_symm):
-    assert isinstance(atlas.filename, str) or isinstance(atlas.filename, Path)
+    assert Path(atlas.filename).exists() and Path(atlas.filename).is_absolute()
     assert isinstance(atlas.maps, nibabel.Nifti1Image)
     assert isinstance(atlas.labels, list)
 


NicolasGensollen (commented):
Thanks for fixing this @RaphaelMeudec !
Please make sure to run a full build when ready for reviews.

RaphaelMeudec:
@NicolasGensollen The PR is ready for review, but I'm not sure how I can trigger the full build?

NicolasGensollen:
@RaphaelMeudec you just need to commit and put "[circle full]" in the commit message.

RaphaelMeudec:
@NicolasGensollen @bthirion I feel like the CI does not like me anymore and is not building anymore? Same on #3188

NicolasGensollen:
> I feel like the CI does not like me anymore and is not building anymore?

I tried re-running but it got canceled again. After a quick look, it seems like CircleCI updated their pricing plans recently such that we only get 1h max of free build time: https://support.circleci.com/hc/en-us/articles/4410707277083-Context-deadline-exceeded-after-1-hour-Build-timed-out-Free-tier-only-

bthirion:
Should we merge that one to fix main ? We'll have to discuss what to do with CircleCI.

NicolasGensollen:
>Should we merge that one to fix main ? We'll have to discuss what to do with CircleCI.

We'll still have the CI red since it's not able to finish the builds, but we can merge this fix while we look for a solution for CircleCI...

@RaphaelMeudec could you rebase and drop my last commit (which turned out to be useless at reducing the build time...)?

bthirion:
The azure failure is unrelated I guess, and the circleci did not complete for lack of time. I'd like to merge though. OK ?

RaphaelMeudec:
@bthirion I tested to run the previously failing script locally with my fix, and it's working. But I'm concerned that merging won't be enough, as it's probably the CI job that is deploying the docs? So if the CI job fails, the docs won't be updated.

bthirion:
In any case, we need the fix. Merging now.

jeromedockes:
> @bthirion I tested to run the previously failing script locally with my fix, and it's working. But I'm concerned that merging won't be enough, as it's probably the CI job that is deploying the docs? So if the CI job fails, the docs won't be updated.

do you mean the online documentation here? https://nilearn.github.io/stable/modules/reference.html
AFAIK it is manually uploaded after each release

NicolasGensollen:
The dev version (https://nilearn.github.io/dev/modules/reference.html) won't build until we have a solution for the CircleCI issue. So the deployed dev version dates from the last successful build on master.

PR #3193, opened 2022-03-25
[ENH] add sample_masks to first level GLM

### LINKED ISSUE #2403: Censor time points during rescaling
In [`FirstLevelModel.fit`](https://github.com/nistats/nistats/blob/f7e1056e26bd5b1070fc361a1d4335b14f0de32c/nistats/first_level_model.py#L324-L506), scaling is performed on the whole time series `Y`:

https://github.com/nistats/nistats/blob/f7e1056e26bd5b1070fc361a1d4335b14f0de32c/nistats/first_level_model.py#L475-L476

For dummy scans or other outliers (such as motion-related intensity spikes), it might be desirable to censor these. You can either handle these by modifying these volumes (e.g., removing dummy scans, or mean-imputing spikes) or by adding dummy regressors for them.

These two strategies will clearly have different effects on the calculated means/ranges of the time series when rescaling. Would it be useful to add a censoring parameter to `fit`, such that any rescaling would ignore those time points?

I can think of a couple options:

1) A list of volume indices. This has the advantage of being simple, while the disadvantage that you're depending on the user to pass consistent censored time points and dummy regressors.
2) Given that we'll have access to the design matrix, we could just use a list of columns that identify dummy regressors, and get the volume indices from these.

htwangtw:
Closes #2403.

- Add sample_masks to the `fit` function in first level GLM.
- Move part of the sanity check on sample_masks from
`nilearn.signal.clean` to `nilearn._utils.glm`


bthirion:
You could scrub arbitrarily some volumes in any of the examples ?

nilearn/glm/first_level/first_level.py
```diff
@@ -421,6 +421,13 @@ def fit(self, run_imgs, events=None, confounds=None,
             respective run_img. Ignored in case designs is not None.
             If string, then a path to a csv file is expected.
 
+        sample_masks : Any type compatible with numpy-array indexing, or
+            list of indices, optional
+            shape of array: (number of scans - number of volumes removed, )
+            Indices of retained volumes. Masks the niimgs along time/fourth
+            dimension to perform scrubbing (remove volumes with high motion)
+            and/or non-steady-state volumes.
```
NicolasGensollen, line 429:
```suggestion
            and/or non-steady-state volumes.
            Default=None.
            
            .. versionadded:: 0.9.1
```

NicolasGensollen (commented):
Thanks @htwangtw !
The changes LGTM, can you also add a whats_new entry?
I'm not sure in which example you could showcase this feature 🤔

nilearn/_utils/glm.py
```diff
@@ -330,3 +330,18 @@ def positive_reciprocal(X):
...
+    if not isinstance(sample_masks, (list, tuple, np.ndarray)):
+        raise TypeError(
+            f"sample_mask has an unhandled type: {sample_masks.__class__}"
+        )
+    if not isinstance(sample_masks, (list, tuple)):
+        sample_masks = (sample_masks, )
+    if len(sample_masks) != n_runs:
+        raise ValueError(
+            (f"Number of sample_mask ({len(sample_masks)}) not matching "
+             "number of runs ({n_runs}).")
```
tsalo, line 345:
```suggestion
            f"Number of sample_mask ({len(sample_masks)}) not matching "
            f"number of runs ({n_runs})."
```

nilearn/glm/first_level/first_level.py
```diff
@@ -421,6 +421,13 @@ def fit(self, run_imgs, events=None, confounds=None,
             respective run_img. Ignored in case designs is not None.
             If string, then a path to a csv file is expected.
 
+        sample_masks : Any type compatible with numpy-array indexing, or
+            list of indices, optional
```
tsalo, line 425: This doesn't render correctly. I think you need a backslash, or maybe an extra indent, to keep the second line as part of the parameter type string, instead of the parameter description.

image

tsalo: For generic numpy-compatible arrays, we typically just say `array_like`, right? Would that work here?

nilearn/glm/first_level/first_level.py
tsalo, line 429:
```suggestion
            dimension to perform scrubbing (remove volumes with high motion
            and/or non-steady-state volumes).
```

tsalo (changes requested):
I just noticed a few small things.

For the example, would it be reasonable to directly compare censoring and scrubbing? I feel like that would be helpful to users.

nilearn/glm/first_level/first_level.py
htwangtw: Scrubbing by definition, is removing volumes with high motion. Hence I don't think we should move the bracket.

nilearn/glm/first_level/first_level.py
htwangtw: Nice catch on this one!!! I will check this.

nilearn/glm/first_level/first_level.py
tsalo: That's fair. In that case it still needs to be updated to be grammatically correct.

```suggestion
            dimension to perform scrubbing (remove volumes with high motion)
            and/or remove non-steady-state volumes.
```

bthirion: Just wondering: wouldn't it be more handy to give  a binary vector with length `n_scans` for  `sample_mask` ?

bthirion (commented):
LGTM.

nilearn/glm/first_level/first_level.py
htwangtw: Yeah let's open an issue to keep track of this...

bthirion (approved):

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -437,6 +437,33 @@ def plot_contrast(first_level_model):
...
 # computation) after this addition.
 
+#########################################################################
+# Volume censoring
+# ------------------
+#
+# Volume censoring is a common way to remove non-steady state volumes, or
+# high-motion volumes in scrubbing based noise removal strategies. In this
+# scenario, we can apply a sample mask along the time dimension to exclude
+# unwanted volumes. When using fMRIPrep outputs from 1.4.x series or above, we
```
NicolasGensollen, line 447:
```suggestion
# unwanted volumes. When using :term:`fMRIPrep` outputs from 1.4.x series or above, we
```

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -437,6 +437,33 @@ def plot_contrast(first_level_model):
...
+# can use the `load_confounds`_ function of Nilearn to retrieve sample masks
```
NicolasGensollen, line 448:
```suggestion
# can use the :func:`~nilearn.interfaces.fmriprep.load_confounds` function of Nilearn to retrieve sample masks
```

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -437,6 +437,33 @@ def plot_contrast(first_level_model):
...
+# based on the given scrubbing threshold and the non-steady state columns.
+# For non-fMRIPrep output, we can still define an sample mask. Here we apply a
```
NicolasGensollen, line 450:
```suggestion
# For non-fMRIPrep output, we can still define a sample mask. Here we apply a
```

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -437,6 +437,33 @@ def plot_contrast(first_level_model):
...
+# sample mask that removes the first 50 volumes.
+#
+# .. _load_confounds: https://nilearn.github.io/stable/modules/generated/nilearn.interfaces.fmriprep.load_confounds.html
```
NicolasGensollen, line 453: You can delete this line

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -437,6 +437,33 @@ def plot_contrast(first_level_model):
...
+first_level_model = FirstLevelModel(t_r, hrf_model='spm + derivative',
+                                    slice_time_ref=0.5)
+first_level_model = first_level_model.fit(fmri_img, events=events,
+                                          sample_masks=sample_masks)
+design_matrix = first_level_model.design_matrices_[0]
+plot_design_matrix(design_matrix)
+plt.show()
+
+#########################################################################
+# Note the significantly showrter design matrix comparing to the previous
```
NicolasGensollen, line 465:
```suggestion
# Note the significantly shorter design matrix compared to the previous
```

NicolasGensollen (commented):
Thanks @htwangtw !
Can you also add a whatsnew entry?

nilearn/glm/first_level/first_level.py
```diff
@@ -421,6 +421,15 @@ def fit(self, run_imgs, events=None, confounds=None,
...
             If string, then a path to a csv file is expected.
 
+        sample_masks : array_like, or list of array_like, optional
+            shape of array: (number of scans - number of volumes removed, )
+            Indices of retained volumes. Masks the niimgs along time/fourth
+            dimension to perform scrubbing (remove volumes with high motion)
+            and/or remove non-steady-state volumes.
+            Default=None.
+
+            .. versionadded:: 0.9.1
```
tsalo, line 431:
```suggestion
            .. versionadded:: 0.9.2.dev
```
Per recommendations being added in #3225.

htwangtw: Thanks for catching this!!!!

NicolasGensollen (approved):
LGTM, thanks @htwangtw !

htwangtw:
Thanks all! Merging

PR #3196, opened 2022-03-26
[ENH] Implement TFCE-based correction in `permuted_ols`

### LINKED ISSUE #2366: Add cluster-level tests and TFCE

tsalo:
Closes #2366.

- Add an option for TFCE-based FWE correction to `permuted_ols` and `non_parametric_inference` (`tfce`).
- Create a new module, `nilearn/mass_univariate/_utils.py`, with a `_calculate_tfce` function that computes a TFCE map from a t-statistic map.
    - This new module is also created in #3181, but with different functions.
- Add a new `masker` parameter to `permuted_ols`.
    - This is also done in #3181. 
- Add a whatsnew entry.
- Add an `output_type` parameter to `permuted_ols` and `non_parametric_inference`. The default value ("legacy") returns the original outputs, but raises a DeprecationWarning. The new value ("dict") returns a dictionary containing several outputs. Enabling TFCE or cluster-level inference (#3181) will override this parameter and set it to "dict" automatically.


examples/07_advanced/plot_localizer_mass_univariate_methods.py
```diff
@@ -57,6 +57,9 @@
 fmri_masked = nifti_masker.fit_transform(contrast_map_filenames)
 
 
+""
+contrast_map_filenames
```
bthirion, line 61: WDYM ?

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+    for i_regressor in range(scores_4d.shape[3]):
+        scores_3d = scores_4d[..., i_regressor]
+
+        # Get the step right before the maximum z-statistic in the map
+        max_z = np.floor(np.max(scores_3d) / dh) * dh
+
+        for score_thresh in np.arange(dh, max_z + dh, dh):
+            # Threshold map
+            thresh_scores_3d = scores_3d.copy()
```
bthirion, line 52: Do we need a copy at each iteration ?

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+            # Label each cluster with its extent
+            cluster_map = np.zeros(scores_3d.shape, int)
+            for cluster_val in range(1, n_clusters + 1):
+                bool_map = labeled_arr3d == cluster_val
+                cluster_map[bool_map] = np.sum(bool_map)
+
+            # Calculate each voxel's tfce value based on its cluster extent
+            # and z-value
+            tfce_step_values = (cluster_map**E) * (score_thresh**H)
```
bthirion, line 69:
```suggestion
            tfce_step_values = (cluster_map ** E) * (score_thresh ** H)
```

bthirion (commented):
Thx for this addition !

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+import nibabel as nib
+import numpy as np
+from scipy import ndimage
+from nilearn.masking import apply_mask, unmask
+
+
+def _calculate_tfce(scores_array, masker, E=0.5, H=2, dh=0.1):
+    """Calculate threshold-free cluster enhancement values for scores maps.
+
+    The TFCE calculation is implemented as described in [1]_.
```
NicolasGensollen, line 11: I feel like it could be worth adding `TFCE` to the glossary. I personally had no idea what it meant before reading the docstring of this private function. WDYT?

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -198,6 +203,12 @@ def _permuted_ols_on_chunk(
         hypothesis is that the effect is zero or negative.
         Default=True
 
+    tfce : :obj:`bool`, optional
```
NicolasGensollen, line 206: Is there a reason why you didn't add docstrings for the other parameters you added (`masker` and `tfce_original_data`)?

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -343,6 +380,12 @@ def permuted_ols(
         If False, only positive effects are considered as relevant. The null
         hypothesis is that the effect is zero or negative. Default=True.
 
+    tfce : :obj:`bool`, optional
+        Whether to perform TFCE-based multiple comparisons correction or not.
+        Calculating TFCE values in each permutation can be time-consuming,
+        so this option is disabled by default.
+        Default=False.
```
NicolasGensollen, line 387:
```suggestion
        Default=False.
        
        .. versionadded:: 0.9.1
        
```

NicolasGensollen (commented):
Thanks @tsalo !
Just made a quick first pass.
I would have to read a little bit about TFCE to be able to review the implementation in `_calculate_tfce`.

examples/07_advanced/plot_localizer_mass_univariate_methods.py
tsalo: Ah, I didn't mean to push this change. It's just an artifact from when I was debugging the example. I'll remove it.

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,80 @@
+"""Utility functions for the permuted least squares method."""
+import nibabel as nib
+import numpy as np
+from scipy import ndimage
+from nilearn.masking import apply_mask, unmask
+
+
+def _calculate_tfce(scores_array, masker, E=0.5, H=2, dh=0.1):
```
tsalo, line 8: Here is the relevant description from [Smith & Nichols (2009)](https://pubmed.ncbi.nlm.nih.gov/18501637/):

>Each voxel's TFCE score is given by the sum of the scores of all “supporting sections” underneath it; as the height h is incrementally raised from zero up to the height (signal intensity) _hp_ of a given point _p_, the image is thresholded at _h_, and the single contiguous cluster containing _p_ is used to define the score for that height _h_. This score is simply the height _h_ (raised to some power _H_) multiplied by the cluster extent _e_ (raised to some power _E_). Precisely, the TFCE output at voxel _p_ is
> image
>where _h0_ will typically be zero, and _E_ and _H_, 0.5 and 2 respectively (these choices are discussed in detail below). In practice this integral is estimated as a sum, using finite _dh_ (for example, _dh_ = 0.1 if the input is a raw _t_ or Z image).

nilearn/mass_univariate/_utils.py
tsalo: I've added an entry. Please let me know if you think it needs more info.

nilearn/mass_univariate/permuted_least_squares.py
tsalo: Sorry about that. It was an oversight on my part. I've now added the parameter descriptions though.

nilearn/mass_univariate/_utils.py
tsalo: Good point. Since lower values are the ones being removed, it can just operate on `scores_3d` directly.

tsalo:
Here is what the TFCE results look like. TFCE is on the far right, with permutation-based voxel-level FWE in the middle, and Bonferroni on the left.

image

nilearn/mass_univariate/_utils.py
NicolasGensollen: Thanks for adding it! I think you could even add [your link](https://benediktehinger.de/blog/science/threshold-free-cluster-enhancement-explained/) in the glossary entry.

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,80 @@
...
+    """Calculate threshold-free cluster enhancement values for scores maps.
+
+    The TFCE calculation is implemented as described in [1]_.
```
NicolasGensollen, line 11:
```suggestion
    The :term:`TFCE` calculation is implemented as described in [1]_.
```

doc/changes/latest.rst
```diff
@@ -9,6 +9,7 @@
 NEW
 ---
 
+:func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` now support TFCE-based FWE correction (:gh:`3196` by `Taylor Salo`_).
```
NicolasGensollen, line 12:
```suggestion
:func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` now support :term:`TFCE`-based FWE correction (:gh:`3196` by `Taylor Salo`_).
```

nilearn/mass_univariate/_utils.py
NicolasGensollen: Great! Thanks, this is super useful! 👍

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        Extent weight. Default is 0.5.
+    H : :obj:`float`, optional
+        Height weight. Default is 2.
+    dh : :obj:`float`, optional
+        Step size for TFCE calculation. Default is 0.1.
+
+    Returns
+    -------
+    tfce_arr : :obj:`numpy.ndarray`, shape=(n_regressors, n_descriptors)
+        TFCE values.
```
NicolasGensollen, line 28:
```suggestion
        :term:`TFCE` values.
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -179,6 +184,12 @@ def _permuted_ols_on_chunk(
     confounding_vars : array-like, shape=(n_samples, n_covars), optional
         Clinical data (covariates).
 
+    masker : :obj:`~nilearn.maskers.NiftiMasker` or \
+            :obj:`~nilearn.maskers.MultiNiftiMasker`, optional
+        Mask to be used on data. This is necessary for TFCE-based inference.
```
NicolasGensollen, line 189:
```suggestion
        Mask to be used on data. This is necessary for :term:`TFCE`-based inference.
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -387,7 +458,13 @@ def permuted_ols(
        linear model. Neuroimage.
 
     .. [3] Freedman, D. & Lane, D. (1983). A nonstochastic interpretation of
-       reported significance levels. J. Bus. Econ. Stats., 1(4), 292-298
+       reported significance levels. J. Bus. Econ. Stats., 1(4), 292-298.
```
NicolasGensollen, line 461: Might be a good opportunity to switch to bibtex references for `permuted_ols`, WDYT?

tsalo: Great idea. I've added this in fcf1d2f.

When I propagate the TFCE enhancement to `non_parametric_inference`, I'll do the same there.

nilearn/mass_univariate/_utils.py
tsalo: Done!

nilearn/mass_univariate/_utils.py
tsalo: I can make this change, but I originally didn't have spaces around the power operators because `black` removes them automatically. I've been largely sticking with `black`'s preferred style, except for using single quotes per `nilearn`'s style guidelines. Would you still prefer me to add whitespace?

nilearn/mass_univariate/_utils.py
tsalo: Done!

doc/changes/latest.rst
```diff
@@ -9,6 +9,7 @@
 NEW
 ---
 
+:func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` now support :term:`TFCE`-based FWE correction (:gh:`3196` by `Taylor Salo`_).
```
bthirion, line 12:
```suggestion
:func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` now support :term:`TFCE` statistic (:gh:`3196` by `Taylor Salo`_).
```
(TFCE is a statistic, and the correction scheme is a standard non-parametric, aka pemutations).

doc/glossary.rst
```diff
@@ -267,6 +267,15 @@ If you wish to add a missing term, please `create a new issue`_ or
         `Support vector machines`_ are a set of :term:`supervised learning` methods used
         for :term:`classification`, :term:`regression` and outliers detection.
 
+    TFCE
+        Threshold-free cluster enhancement is a method for thresholding statistical maps
```
bthirion, line 271: I would call it a statistic that combines signal magnitude and cluster extent to enhance the importance of large and high clusters.

examples/07_advanced/plot_localizer_mass_univariate_methods.py
```diff
@@ -71,56 +71,107 @@
...
-                        display_mode='z', cut_coords=[z_slice],
-                        figure=fig, vmax=vmax, black_bg=True)
+n_detections = (
+    get_data(neg_log_pvals_permuted_ols_unmasked) > threshold
+).sum()
+title = (
+    'Negative $\\log_{10}$ p-values\n'
+    '(Non-parametric +\n'
+    'max-type correction)\n'
+    f'{n_detections} detections'
```
bthirion, line 150:
```suggestion
    f'{n_detections} voxels detected'
```

nilearn/mass_univariate/_utils.py
bthirion: I thought it was the PEP rule,  but this is clearly not essential.

doc/glossary.rst
tsalo: Updated in [3bbdcaf](https://github.com/nilearn/nilearn/pull/3196/commits/3bbdcaf841351c108f57407572b0b6705500b29e).

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -534,4 +636,12 @@ def permuted_ols(
...
-    return - np.log10(pvals), scores_original_data.T, h0_fmax
+    if tfce:
+        return (
+            -np.log10(pvals),
+            scores_original_data.T,
+            h0_fmax,
+            -np.log10(tfce_pvals),
+        )
+    else:
+        return - np.log10(pvals), scores_original_data.T, h0_fmax
```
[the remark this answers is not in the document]
tsalo: Then, in #3181, we'd be adding
```python
{
    'neglog10p_max_cluster_size': -np.log10(csfwe_pvals),
    'neglog10p_max_cluster_mass': -np.log10(cfwe_pvals),
    'h0_max_cluster_size': h0_csmax,
    'h0_max_cluster_mass': h0_cmmax,
}
```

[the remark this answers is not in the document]
bthirion: LGTM. But we need to go through deprecation cycles, don't we ?

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,80 @@
...
+            # Label each cluster with its extent
+            cluster_map = np.zeros(scores_3d.shape, int)
+            for cluster_val in range(1, n_clusters + 1):
+                bool_map = labeled_arr3d == cluster_val
+                cluster_map[bool_map] = np.sum(bool_map)
+
+            # Calculate each voxel's tfce value based on its cluster extent
+            # and z-value
+            tfce_step_values = (cluster_map**E) * (score_thresh**H)
+            tfce_4d[..., i_regressor] += tfce_step_values
```
bthirion, line 69: should be multiplied by dh I guess ?

nilearn/mass_univariate/permuted_least_squares.py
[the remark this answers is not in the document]
NicolasGensollen: I agree with @bthirion, if we change the output of `permuted_ols` for `tfce=False` we should have a deprecation cycle.

nilearn/mass_univariate/_utils.py
tsalo: I _think_ `dh` is the differential in the original formula. If so, then I don't think we actually use it in the TFCE calculation. I could be wrong though.

bthirion: I agree it's not a major thing, but if you don't multiply by dh, TFCE will depend on the choice of dh (because the number of terms in the sum changes). Multiplying by dh removes this dependence, which is better I think.

tsalo: Oh, I hadn't considered that. If it improves interpretability of the metric, then I agree that it's worth doing. I'll add it.

nilearn/mass_univariate/permuted_least_squares.py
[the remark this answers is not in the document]
bthirion: +1 for adding the `output_type` kw to handle deprecation.

nilearn/mass_univariate/_utils.py
tsalo: Done in [8a4be9d](https://github.com/nilearn/nilearn/pull/3196/commits/8a4be9d1adc1561daa7186a6eefc8a4ad72d012b).

bthirion:
Question: have you checked whether the TFCE values you obtain approximately match those of FSL ?
Best,

examples/07_advanced/plot_localizer_mass_univariate_methods.py
```diff
@@ -71,56 +71,107 @@
...
+neg_log_pvals_permuted_ols, _, _, neg_log_pvals_tfce = permuted_ols(
+    tested_var,
+    fmri_masked,
     model_intercept=True,
-    n_perm=5000,  # 5,000 for the sake of time. Idealy, this should be 10,000
-    verbose=1, # display progress bar
-    n_jobs=1)  # can be changed to use more CPUs
+    masker=nifti_masker,
+    tfce=True,
+    n_perm=1000,  # 1000 for the sake of time. Ideally, this should be 10000.
```
bthirion, line 83: The example is taking 20min to run on my box. This can be reduced by playing with n_jobs, but this is too much. I'm afraid we should further reduce the number of permutations to 400 or even 200 to make it more reasonable.

tsalo:
>Question: have you checked whether the TFCE values you obtain approximately match those of FSL ?

To be honest, I hadn't realized that FSL let you calculate TFCE outside of randomise. However, I ran it through `fslmaths`, and it passes `numpy.allclose` with the following settings:

FSL: `fslmaths z.nii.gz -tfce 2 0.5 6 tfce.nii.gz`
nilearn: `dh` is set to `np.max(scores_array) / 100`

I tried to figure out how to use a dh of 0.1 for FSL, but the only related option (`-tfce_delta`) seems to be ignored in the standard fslmaths tfce call.

The paper mentions 0.1 for `dh`, but fslmaths definitely uses 100 steps instead. Which would you prefer?

examples/07_advanced/plot_localizer_mass_univariate_methods.py
tsalo: Can do!

tsalo:
Also, it doesn't look like fslmaths supports two-sided tests, but I don't think there's any reason we shouldn't support it.

bthirion:
What matters to me is that there is not a crazy divergence between  the output of our procedure and that of FSL. Regarding the choice of integration step, both choices are actually reasonable. I have no strong opinion. Fixing the number of steps to 100 means will stabilize computation time.

tsalo:
Here is a plot of the nilearn TFCE values against the FSL ones, for when `two_sided` is False vs. True.

image

nilearn/glm/tests/test_second_level.py
```diff
@@ -574,9 +574,39 @@ def test_non_parametric_inference_permutation_computation():
                                                      mask=mask, n_perm=100)
 
         assert get_data(neg_log_pvals_img).shape == shapes[0][:3]
+        raise Exception(np.unique(neg_log_pvals_img.get_fdata()))
         del func_img, FUNCFILE, neg_log_pvals_img, X, Y
 
 
+def test_non_parametric_inference_tfce():
```
tsalo, line 581: This test fails because the t-statistics are all -inf, 0, or inf, which `_calculate_tfce` will fail on. Is there a simple alternative to the current simulated data that won't produce infs?

bthirion: Sorry,  i haven't checked in detail how this test is generated, but there should not be inf values in tests. This is a bug in the test generation imho.

tsalo: If I create 4 images instead of duplicating one image 4 times, the problem goes away. I don't know if it's worth propagating that fix to the other tests, considering it probably slows things down.

nilearn/mass_univariate/_utils.py
```diff
@@ -0,0 +1,111 @@
...
+                # Each voxel's cluster extent at threshold *h* is thus *e(h)*
+                cluster_map = np.zeros(temp_scores_3d.shape, int)
+                for cluster_val in range(1, n_clusters + 1):
+                    bool_map = labeled_arr3d == cluster_val
+                    cluster_map[bool_map] = np.sum(bool_map)
+
+                # Calculate each voxel's tfce value based on its cluster extent
+                # and z-value
+                tfce_step_values = (cluster_map**E) * (score_thresh**H)
+                tfce_4d[..., i_regressor] += sign * tfce_step_values
```
bthirion, line 100: I think that we should add here as a comment that we  ---incorrectly--- do not multiply by dh in order to match the original FSL implementation.

bthirion (commented):
LGTM otherwise !

nilearn/mass_univariate/_utils.py
tsalo: I've added a comment. I also expanded the docstring a bit with similar info.

bthirion (approved):
LGTM, thx !

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -358,31 +417,114 @@ def permuted_ols(
...
+        or not.
+        Calculating TFCE values in each permutation can be time-consuming,
+        so this option is disabled by default.
+        The TFCE calculation is implemented as described in
+        :footcite:t:`smith2009threshold`.
+        Default=False.
+
+        .. versionadded:: 0.9.2
+
+    output_type : {'legacy', 'dict'}, optional
```
NicolasGensollen, line 437: Same comment on deprecated option `legacy` as in my review of #3181

NicolasGensollen (commented):
This LGTM, thanks @tsalo ! 👍 
Since there are some overlap between this PR and #3181, which one would you recommend merging first?

tsalo:
Thanks for the review @NicolasGensollen. Since there's only one review comment in this PR, and the same comment is in the other PR, I think it makes sense to merge this one first, so I can address the reviews and merge conflicts in the other PR in one go.

nilearn/mass_univariate/permuted_least_squares.py
tsalo: I think this should be fixed in #3181 now. Is this good to merge then?

tsalo:
Resolving the merge conflicts from #3181 was a bit tougher than I thought it would be, but I think they're good now.

bthirion:
Oops, it hurts. We should be explicit about that.

tsalo:
I've added a bit more text to the docstrings about TFCE time consumption in [2bb9d99](https://github.com/nilearn/nilearn/pull/3196/commits/2bb9d99b4bd7a579944cddf513b9afa6b1edf9c3). WDYT?

NicolasGensollen (approved):
LGTM, thanks @tsalo !

bthirion:
I confirm it's goos with me. Thx!

bthirion:
All right, merging now.

PR #3201, opened 2022-03-29
MRG: Fixing a bug in `nilearn.image._apply_cluster_size_threshold`

### LINKED ISSUE #3200: get_cluster_table returns inconsistent result depending on `cluster_threshold`
Nilearn version: 0.9.0

### Expected behavior
The example below should return identical information for the large clusters (> 10 voxels)

### Actual behavior
I does not

```
from nilearn import datasets
threshold = 3.0
table1 = get_clusters_table(stat_img, threshold, cluster_threshold=10)
table2 = get_clusters_table(stat_img, threshold, cluster_threshold=0)
print(table1)
print(table2)
```

bthirion:
Closes #3200


NicolasGensollen (commented):
Thanks @bthirion !
Could you add a non regression test and a whatsnew entry?

bthirion:
Sure ;-)

doc/changes/latest.rst
```diff
@@ -15,7 +15,8 @@ Fixes
 
 - Fix function :func:`~datasets.fetch_abide_pcp` which was returning empty phenotypes and ``func_preproc`` after release ``0.9.0`` due to supporting pandas dataframes in fetchers (:gh:`3174` by `Nicolas Gensollen`_).
 - Fix function :func:`~datasets.fetch_atlas_harvard_oxford` and :func:`~datasets.fetch_atlas_juelich` which were returning the image in the `filename` attribute instead of the path to the image (:gh:`3179` by `Raphael Meudec`_).
-
+- Fix function :func:`~image._apply_cluster_size_threshold`, which resulted in wrong clusters extraction when cluster_size was non-zero  (:gh:`3200` by `Bertrand Thirion`_).
```
NicolasGensollen, line 18:
```suggestion
- Fix function ``nilearn.image._apply_cluster_size_threshold``, which resulted in wrong clusters extraction when cluster_size was non-zero  (:gh:`3201` by `Bertrand Thirion`_).
```

NicolasGensollen (commented):
Thanks @bthirion ! 👍 
Just a small suggestion (the ref to `_apply_cluster_size_threshold` won't work since it's a private function), LGTM otherwise!

NicolasGensollen:
And it looks like there is a conflict in `doc/changes/latest.rst` atm.

RaphaelMeudec:
Small comment not on the PR itself but on the function: why is there the iteration over  `np.sign(arr)` in the first place? 

- the function accepts a thresholded array (it is specified in the docstring), so `np.sign(arr)` is also either `[1]` or `[0, 1]`
- next, we binarize the element by `((arr * sign) > 0).astype(int)` before extracting the clusters
- so the `0` value is a useless iteration and we only perform something when `sign == 1`

As the sign of the threshold array can never be `-1`, it seems to me like this whole iteration over `np.sign(arr)` could be removed?

tsalo (approved):
The change looks good to me. Just to clarify though, the new assertion you added to the test was failing before, right?

bthirion:
> The change looks good to me. Just to clarify though, the new assertion you added to the test was failing before, right?

Yes.

bthirion:
Can somebody approve that one ? It's really a bug fix.

NicolasGensollen (approved):
LGTM, thanks !

PR #3202, opened 2022-03-30
More explicit docstring regarding min_distance

### LINKED ISSUE #3199: Unclear documentation regarding `min_distance` in `reporting.get_clusters_table`
The method `nilearn.reporting.get_clusters_table` which extracts peak statistics from an image accepts an argument which is `min_distance` described as "minimal distance between subpeaks in mm".

I find this docstring a bit confusing as the minimal distance is applied at the cluster-level and not at the image level. This results in a dataframe where peaks can actually be closer than `min_distance` if they belong to different clusters. 

Maybe this nuance was implied in the use of "subpeaks" instead of "peaks", but my opinion is that it should be a bit more explicit. WDYT?

Relevant portion of code [here](https://github.com/nilearn/nilearn/blob/1607b52458c28953a87bbe6f42448b7b4e30a72f/nilearn/reporting/_get_clusters_table.py#L234)

RaphaelMeudec:
Closes #3199.

This PR changes the docstring for `min_distance` argument in the `reporting.get_clusters_table` method to explicitly note that the minimal distance applies intra-clusters and not between clusters.

### DIFF
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -152,7 +152,9 @@ def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
         only. Default=False.
 
     min_distance : `float`, optional
-        Minimum distance between subpeaks in mm. Default=8mm.
+        Minimum intra-cluster distance between subpeaks in mm. Default=8mm. If
+        two different clusters are closer than `min_distance`, it can results
+        in peaks closer than `min_distance`.
 
     Returns
     -------
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -155,8 +155,8 @@ def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
         Minimum distance between subpeaks in mm. Default=8mm.
 
         .. note::
-            If two different clusters are closer than `min_distance`, it can
-            result in peaks closer than `min_distance`.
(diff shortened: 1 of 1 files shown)


nilearn/reporting/_get_clusters_table.py
```diff
@@ -152,7 +152,9 @@ def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
         only. Default=False.
 
     min_distance : `float`, optional
-        Minimum distance between subpeaks in mm. Default=8mm.
+        Minimum intra-cluster distance between subpeaks in mm. Default=8mm. If
+        two different clusters are closer than `min_distance`, it can results
```
NicolasGensollen, line 156: It looks like it would fit nicely in a `.. note::` directive

nilearn/reporting/_get_clusters_table.py
```diff
@@ -154,6 +154,10 @@ def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
     min_distance : `float`, optional
         Minimum distance between subpeaks in mm. Default=8mm.
 
+        .. note::
+            If two different clusters are closer than `min_distance`, it can
+            result in peaks closer than `min_distance`.
```
tsalo, line 159:
```suggestion
            If two different clusters are closer than ``min_distance``, it can
            result in peaks closer than ``min_distance``.
```

NicolasGensollen:
Thanks ! Merging...

NicolasGensollen:
Thanks ! Merging...

PR #3212, opened 2022-04-06
[DOC] Fix documentation links

### LINKED ISSUE #3163: Links direct to the old version of documentation
I am not sure how many cases are there, but I encounter some links are pointing to the old doc instead of the stable release doc.

### Affected pages
The one place I can think of is the [glossory](https://nilearn.github.io/stable/glossary.html#term-functional-connectome)
Not sure if there are pages that were systematically populated and share the same problem.

NicolasGensollen:
Closes #3163

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@ It's designed to capture information we've found to be useful in reviewing pull 
 If there is other information that would be helpful to include, please don't hesitate to add it!
 
 Please make sure your pull request also follows the 
-[contribution guidelines](https://nilearn.github.io/development.html#contribution-guidelines) that
+[contribution guidelines](https://nilearn.github.io/stable/development.html#contribution-guidelines) that
 will be enforced during the review process.
 -->
 
--- a/.github/workflows/auto-comment.yml
+++ b/.github/workflows/auto-comment.yml
@@ -16,7 +16,7 @@ jobs:
             in its title, or leave it as a github draft.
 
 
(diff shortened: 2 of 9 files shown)


ymzayek:
@NicolasGensollen I went through the corrections and particularly relevant CircleCI build_docs outputs (e.g. `dev/glossary.html`) to check if all the relevant links were changed and it looks good to me. 

Correct me if I'm wrong, but I presume that the repo `nilearn.github.io` has been cleaned already since #3163 opened because in the release the links are now broken instead of linking to old versions of the documentation. 

I'm also wondering is the workflow already checking for broken links?

NicolasGensollen:
Thanks for the review @ymzayek ! 👍 

>Correct me if I'm wrong, but I presume that the repo nilearn.github.io has been cleaned already since https://github.com/nilearn/nilearn/issues/3163 opened because in the release the links are now broken instead of linking to old versions of the documentation.

No, it hasn't been cleaned yet. The broken links are probably related to this commit:

https://github.com/nilearn/nilearn.github.io/commit/6f34c2924224ceba232f316f9c9f2d5a357f16cb

where I deleted one of the old folders for testing. I haven't had the time to clean the rest since then.

>I'm also wondering is the workflow already checking for broken links?

No, but this would be a good addition indeed !

ymzayek (approved):

bthirion:
Can you take care of the pep8 failures ?
Best,

bthirion (approved):

NicolasGensollen:
Thanks, merging...

PR #3216, opened 2022-04-08
[ENH] Refactor OpenNeuro-fetching functions

### LINKED ISSUE #2853: `fetch_openneuro_dataset_index` should clearly indicate it's hardcoded to retrieve a particular dataset
The docstring for `nilearn.datasets.fetch_openneuro_dataset_index` implies the function can be used to retrieve the file index from any OpenNeuro dataset, but it's actually hardcoded to only work with one dataset (ds000030). This is confusing, as the local directory takes the name of the requested dataset, even though it's always the same file list that's retrieved. A generalized function would be best if feasible, but failing that, it probably makes sense to only take a `version` argument and clarify in the docstring that the scope of the utility is limited to ds000030.

tsalo:
Closes #2853.

- Create a new function, `nilearn.datasets.func.fetch_ds000030_urls`, that re-implements `fetch_openneuro_dataset_index`, but with a clearer scope and name.
- Deprecate `fetch_openneuro_dataset_index`.
- Improve the docstring of `nilearn.datasets.func.patch_openneuro_dataset` and make the code a little clearer.
- Improve the documentation of `nilearn.datasets.func.fetch_openneuro_dataset` and refactor the code slightly.


nilearn/datasets/func.py
```diff
@@ -1894,45 +1957,66 @@ def infer_subjects(urls):
...
+    data_dir=None,
+    dataset_version='ds000030_R1.0.4',
+    verbose=1,
+):
     """Download OpenNeuro :term:`BIDS` dataset.
 
+    This function specifically downloads files from a series of URLs.
+    Unless you use :func:`fetch_ds000030_urls` or the default parameters,
+    it is up to the user to ensure that the URLs are correct,
+    and that they associated with an OpenNeuro dataset.
```
NicolasGensollen, line 2008:
```suggestion
    and that they are associated with an OpenNeuro dataset.
```

nilearn/datasets/func.py
```diff
@@ -1894,45 +1957,66 @@ def infer_subjects(urls):
 
 
 def patch_openneuro_dataset(file_list):
-    """Add symlinks for files not named according to latest :term:`BIDS` conventions.
+    """Add symlinks for files not named according to :term:`BIDS` conventions.
+
+    This function uses a series of hardcoded patterns to generate the corrected
```
NicolasGensollen, line 1962: Maybe we can put this in a warning block?

nilearn/datasets/func.py
```diff
@@ -1779,49 +1779,112 @@ def fetch_bids_langloc_dataset(data_dir=None, verbose=1):
...
+    -------
+    urls_path : :obj:`str`
+        Path to downloaded dataset index.
+    urls : :obj:`list` of :obj:`str`
+        Sorted list of dataset directories.
+    """
+    warnings.warn(
+        (
+            'The "fetch_openneuro_dataset_index" function was deprecated in '
+            'version 0.9.1, and will be removed in 0.13. '
```
NicolasGensollen, line 1816: I thought we were using 2 minor release cycles for deprecation length. I was expecting `0.11` since we deprecate in the 0.9.x series. It is not super costly to keep it until `0.13` though if you think that's better.

nilearn/datasets/func.py
```diff
@@ -1943,22 +2027,65 @@ def fetch_openneuro_dataset(
...
+            data_prefix,
+            data_dir=data_dir,
+            verbose=verbose,
+        )
 
     # The files_spec needed for _fetch_files
     files_spec = []
     files_dir = []
     for url in urls:
+        # NOTE: This seems fragile. What if data_prefix isn't in the URL?
```
NicolasGensollen, line 2088: I agree (esp. when this is provided by the user...), I would at least add an explicit check before trying to split, and raise a comprehensive error if this is not the case.

NicolasGensollen (commented):
Thanks @tsalo ! 👍 
I made a first pass and I think it looks great!
It seems like the coverage should be improved a bit though.
Could you also add a whatsnew entry?

nilearn/datasets/func.py
tsalo: Done!

nilearn/datasets/func.py
tsalo: Ah, right. I think I just blanked on the cycle length. Thanks for catching it.

nilearn/datasets/func.py
tsalo: Done!

tsalo:
I've added a whatsnew entry. I honestly keep forgetting about that. Maybe we could add it to the PR commenter Action's checklist?

I think the coverage should be better now that I've improved the tests, but if it's still missing any chunks of code I'll take another crack at it.

tsalo:
I had a mistake in one of the updated tests. Now that it's passing, coverage looks better.

bthirion (commented):
Coverage can still be improved can't it ?
LGTM otherwise.

tsalo:
It looks like the only adversely-affected lines are:

https://github.com/nilearn/nilearn/blob/dd4b5b2444c5f87e0fedcc90c93ce511fb45efd0/nilearn/image/image.py#L212-L216

I just can't figure out how my changes impact coverage of those lines... 😕

bthirion (approved):
LGTM, thx !

nilearn/datasets/func.py
```diff
@@ -1779,49 +1779,112 @@ def fetch_bids_langloc_dataset(data_dir=None, verbose=1):
...
+        )
+
+    urls_path, urls = fetch_ds000030_urls(data_dir=data_dir, verbose=verbose)
+    return urls_path, urls
+
+
+@fill_doc
+def fetch_ds000030_urls(data_dir=None, verbose=1):
+    """Fetch URLs for files from the ds000030 :term:`BIDS` dataset.
+
```
NicolasGensollen, line 1839:
```suggestion

    .. versionadded:: 0.9.2
    
```

tsalo: Thanks for catching that. Should be good now.

NicolasGensollen (approved):
LGTM, thanks @tsalo !

tsalo:
Since there are two approvals, I'm going to merge this.

PR #3226, opened 2022-04-18
[DOC] Add note about function size/scope to contributing guidelines

### LINKED ISSUE #1769: Code refactoring necessary for future ease of maintenance and development

tsalo:
Closes #1769.

- Recommend writing short, testable functions.
- Recommend against reorganizing existing functions for the sake of style.
- Point to test-driven development in testing section.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -234,7 +234,7 @@ maintenance changes.
 Coding Style
 -------------
 
-Nilearn codebase follow PEP8_ styling.
+The nilearn codebase follows PEP8_ styling.
 The main conventions we enforce are :
 
 - line length < 80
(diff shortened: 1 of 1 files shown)


bthirion (approved):
Very clear to me. Thx !

jeromedockes (approved):
Looks great, thank you!

PR #3227, opened 2022-04-18
avoid creating Niftis containing 64-bit ints

### LINKED ISSUE #3167: 64-bit integer NIfTI images
Today's office hours were dedicated to an issue concerning NiftiImages saved as int64, which is the default type you get in Python/Numpy:

```python
>>> import numpy as np
>>> a = np.array([1, 2, 3])
>>> a.dtype
dtype('int64')
```

This data type isn't currently supported by other tools outside of the Python world (AFNI, FSL...), and NiftiImages saved as int64 are currently responsible for numerous issues on their side.

Nibabel and Nilearn are concerned in the sense that they enable writing these images to disk. As the community of users of these tools grows, so does the amount of Int64 images written on users hardwares, and so does the burden on other tools which do not support this data type.

Since we couldn't think of a reasonable use case where 64 bits integer niftiimages would be indispensable, the decision was taken to deprecate this type.

Here is the plan of action on the Nibabel's side: https://github.com/nipy/nibabel/issues/1089

On the Nilearn's side, we need to assess what part of the codebase would need to be updated, and see whether these changes would break users code. It is not clear to me yet how much work this would require, but we should probably start with places where we write NiftiImages to disk (functions like `new_img_like` for instance).

@jeromedockes @htwangtw feel free to complement this issue. I wanted to open it before the weekend such that we don't forget about it, but I might have missed some important details.

jeromedockes:
Closes #3167 

Several tools such as FSL cannot read int64 Nifti files; Nibabel will refuse to write them in the future.
This PR tries to avoid creating such images in nilearn.
Still WIP, need to
- [x] add more tests
- [x] add comments and a note about int64 Niftis in the contributing guidelines


nilearn/image/tests/test_resampling.py
```diff
@@ -163,8 +163,8 @@ def test_resampling_with_affine():
     """
     rng = np.random.RandomState(42)
 
-    data_4d = rng.randint(4, size=(1, 4, 4, 3))
-    data_3d = rng.randint(4, size=(1, 4, 4))
+    data_4d = rng.randint(4, size=(1, 4, 4, 3), dtype=np.int32)
+    data_3d = rng.randint(4, size=(1, 4, 4), dtype=np.int32)
```
bthirion, line 167: Why not int32 ?

bthirion: (same comments for the lines below)

bthirion (commented):
Thx for taking care of this !

nilearn/image/image.py
```diff
@@ -728,6 +728,12 @@ def new_img_like(ref_niimg, data, affine=None, copy_header=False):
         if isinstance(ref_niimg, nibabel.freesurfer.mghformat.MGHImage):
             default_dtype = np.uint8
         data = as_ndarray(data, dtype=default_dtype)
+    if data.dtype in (np.int64, np.uint64):
```
htwangtw, line 731: Do we want to raise a warning here to tell user that the `dtype` has been changed?

nilearn/image/tests/test_resampling.py
jeromedockes: do you mean why not `dtype="int32"` instead of `dtype=np.int32`? I don't have a preference, I can change to the string version

nilearn/image/image.py
jeromedockes: thanks, I was going to ask for opinions about what should be the behavior here.
I believe that because int64 is the representation chosen by numpy when it is given python ints, this downcasting will happen very frequently. And as there probably very few use cases (if any) for images containing ints that don't fit in 32 bits, for the huge majority of users, it will not make a difference. So I am hesitant to emit a warning. Moreover, I noted that in the case the data contains `bool` (the few lines just above), the casting is done silently (but in that case it is a safe conversion).

At the same time, silently doing an unsafe type conversion is of course not a good idea in principle.

bthirion: I feel we should have a warning.

nilearn/image/tests/test_resampling.py
bthirion: I had the impression that it was recommended to use the string version.

bthirion (commented):
LGTM. Deserves a whatsnew enty.

jeromedockes:
I wonder if functions like `resample_img`, `resample_to_img`, `check_img` should also do conversions?

examples/01_plotting/plot_carpet.py
```diff
@@ -59,11 +58,8 @@
 atlas_data = atlas_img.get_fdata()
 discrete_version = np.argmax(atlas_data, axis=3) + 1
 discrete_version[np.max(atlas_data, axis=3) == 0] = 0
-discrete_atlas_img = nib.Nifti1Image(
-    discrete_version,
-    atlas_img.affine,
-    atlas_img.header,
-)
```
jeromedockes, line 66: not related to this PR but is there an atlas fetcher we could use here instead of building one? seems to complicate the example and it's not related to plotting

nilearn/plotting/tests/test_find_cuts.py
```diff
@@ -52,7 +52,7 @@ def test_find_cut_coords():
     data = np.zeros((20, 20, 20))
     data[4:6, 4:6, 4:6] = 1000
     img = nibabel.Nifti1Image(data, 2 * np.eye(4))
-    mask_data = np.ones((20, 20, 20), dtype=int)
+    mask_data = np.ones((20, 20, 20), dtype="uint8")
```
jeromedockes, line 55: There seems to be a lot of tests throughout the library that build their own example images by creating an array and affine and then creating the `Nifti1Image`. the plotting test utils also provide fixtures for this. It would be better to rely on the utilities in `_utils.data_gen` to avoid all this duplication. I can think of 2 reasons why we keep duplicating this code rather than using the existing functions:
- we forget or don't know they exist -> we can add a note about this in the user guide; maybe rename the file? (`_utils` sounds like utils for the library itself but maybe fake data generation should go in something called `testing_utils`, we also have `_utils/testing.py`); PR reviewers should remember to check if code that already is available is not being duplicated
- the functions in `data_gen` may be a bit too complex; for simple cases re-writing the image generation is faster than reading their docstrings. Maybe we need to add a smaller simple function to cover frequent and simple use cases.

jeromedockes: fixtures could definitely help for tests that only need one default image; for more complex scenarios I bet importing `data_gen` and calling a function would feel simpler than parametrizing a pytest fixture. I completely agree that copy-pasting the data-generation snippets from other tests is very natural and that re-writing those would help; in the meanwhile paying attention to this in code reviews and encouraging the use of `data_gen` would be a start

jeromedockes:
> I'll take a closer look locally at the modifications made to `fetch_atlas_talairach` because the diff was a bit hard to make sense of.

I think you can just look at and try the new version.
The original (downlaoded) file mixes all levels of the atlas in one image.
The previous version of the fetcher kept all levels in one image, but changed the values so that different levels occupy different bytes in the image values.
The version in this PR stores each level of the atlas in a separate image. The fetcher was returning only one level at a time anyway, so there isn't any benefit to having them all in the same image, at least for someone using only the public API.

bthirion:
> I wonder if functions like `resample_img`, `resample_to_img`, `check_img` should also do conversions?

I think so.

jeromedockes:
> > I wonder if functions like `resample_img`, `resample_to_img`, `check_img` should also do conversions?
> 
> I think so.

Ok! WDYT about handling it in a separate PR? this one is already useful as it is, as it removes the creation of many int64 images and forbids them in the unit tests. adding the optional conversions to other functions in a separate PR avoids delaying the main one and keeps it small, and makes it easier for someone else to jump in as I don't know if I will have time in the coming weeks.

sorry I forgot to remove the [WIP] but this is ready for review and potentially merge

bthirion (approved):
LGTM, thx.

NicolasGensollen (approved):
LGTM, thanks @jeromedockes ! 👍

jeromedockes:
merging based on the 2 approvals, thanks for the reviews!

PR #3234, opened 2022-04-29
[FIX] Compensate for `c_ras` offset between FreeSurfer surface and volume

ymzayek:
Picking up #2215.

- Check for 'cras' in header of surface
- Add offset to coordinates


ymzayek:
I will work on this in the original pull request to keep the history of the original contributer.

PR #3235, opened 2022-04-29
[FIX] Compensate for `c_ras` offset between FreeSurfer surface and volume

ymzayek:
Replaces and closes #2215

- Check for 'cras' in header of surface
- Add offset to coordinates


bthirion:
You probably want to add  a test. Is there an example that helps ensuring that the projection is better indeed ?

ymzayek:
I was able to reproduce the shift which can be seen visually before and after the fix with the following code:
```
import numpy as np
from nilearn import plotting, datasets, surface  

mni = datasets.load_mni152_template(resolution=2)
surf_file = "/home/yasmin/freesurfer_test_data/cvs_avg35_inMNI152/surf/lh.pial"

mesh = surface.load_surf_mesh(surf_file)  
mapped = surface.vol_to_surf(mni, mesh)  
fig = plotting.plot_surf_stat_map(mesh, mapped, hemi='left', colorbar=True)
fig.show()
```
The surf_file can be downloaded from [here](https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/tutorial_packages/OSX/freesurfer/subjects/cvs_avg35_inMNI152/surf/lh.pial)

### The results before the fix:
cras_offset

### And after:
cras_offset_fixed

bthirion:
Thx ! I think that the next step is to build  a unit test to ensure that the correction applies properly, by assessing that the variance of the projected T1 image gets below a certain threshold.

ymzayek:
@bthirion I added a skeleton for a test. It of course fails because we do not have the specific surface files that produce the issue nor a way to generate meaningful ones (that I know of). Let me know what you think.

bthirion:
If we have a file that triggers the problem, shipped within the library, then we can rely on it for the test. Otherwise, I'm afraid, we won't have a particular test for that.

ymzayek:
It seems to me we only have surface mesh data in GIFTI format (.gii, .gii.gz) for which this issue does not apply. 
It applies to freesurfer specific file format (.orig, .pial, .white, .sphere, or .inflated).
Would either a warning or not supporting that format be an option?
Perhaps others have another idea...

bthirion:
No, we should not raise a warning when people use these formats. I'm just worried of potential regression in the future.

nilearn/surface/tests/test_surface.py
```diff
@@ -595,23 +595,6 @@ def test_vol_to_surf(kind, n_scans, use_mask):
...
-                                                dir=str(tmp_path))
-        os.close(fd)
-        nb.freesurfer.write_geometry(filename_fs_mesh, mesh[0], mesh[1])
-        surf_file = filename_fs_mesh
-        mesh = surface.load_surf_mesh(surf_file)
-        mapped = surface.vol_to_surf(template, mesh)
-        normalized = (mapped - min(mapped)) / (max(mapped) - min(mapped))
-        assert np.var(normalized) < 0.02
-        os.remove(filename_fs_mesh)
-
```
bthirion: Sorry, I think I misunderstood what you were doing in the discussion.
So, as the mesh is generated locally you only rely on the template, but this is shipped with Nilearn, hence we should not worry about that, right ? Maybe for the template we could simply use
`from nilearn.datasets import MNI152_FILE_PATH` (see e.g. https://nilearn.github.io/auto_examples/plot_nilearn_101.html)
I would only run it on one surface only, since the test may be a bit expensive. Do we agree that the test would fail on the current codebase ?
And I would add count (refs to the issue) to explain the purpose of the test.

ymzayek: Yes this test fails and this was expected, it was just an example of what it should look like. The mesh generated locally is useless because we actually need the freesurfer specific files which we cannot generate and we do not ship with nilearn. 

The issue relates to how the coordinates are read for specific freesurfer files such as .pial and we do not ship those with Nilearn. In other words, the only way I see to test this is if we have a . pial file that we can load with `load_surf_mesh`

ymzayek: Indeed we can use the template you mention but still we need a .pial file (or .white, etc)

bthirion: But can't we geenrate the .pial file with nibabel ?

bthirion: But yes, I understand that we can't generate it with the cras thing as freesrfer does. Forget about my latest comment.

bthirion:
I'd be in favor of merging without tests then. Just add a reference to the issue so that people understand why it has been added.

ymzayek:
At this point is it ready to merge without testing if approved. Note the code coverage test on the diff will fail because of this. 
What do others think?

NicolasGensollen (commented):
LGTM, thanks @ymzayek !
Could you add a what's new entry for this?

ymzayek:
@bthirion @NicolasGensollen 
I think it is ready now; WDYT?

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

nilearn/surface/surface.py
```diff
@@ -813,7 +813,10 @@ def load_surf_mesh(surf_mesh):
         if (surf_mesh.endswith('orig') or surf_mesh.endswith('pial') or
                 surf_mesh.endswith('white') or surf_mesh.endswith('sphere') or
                 surf_mesh.endswith('inflated')):
-            coords, faces = fs.io.read_geometry(surf_mesh)
+            coords, faces, header = fs.io.read_geometry(surf_mesh,
+                                                        read_metadata=True)
+            if 'cras' in header:
```
bthirion, line 818: Would you mind adding here a reference to the issue or PR ?

ymzayek: Is it sufficient to do it like this?

bthirion (approved):
LGTM, thx !

bthirion:
Merging, then !

PR #2294, opened 2020-01-23
MultiNiftiLabelsMasker object for handling 4D + 1D input

Hororohoruru:
One of the objects proposed in #1240 , to adress several issues (e.g., #2247 ).

I used the MultiNiftiMasker object as a rough template for this. This is the very first version and needs a lot of improvements. For now, I added a test to check it accepts 4D + 1D input, and it is handled by parallelizing the function transform_single_imgs of NiftiLabelsMasker.

Several issues need to be addressed, I would be grateful for suggestions on how to do so:
- Since each 4D image needs the labels_img to be resample to its affine, there needs to be a way
of storing the affines/label images.
- For the 'data' strategy, it can be done by parallelizing the single transform method from the NiftiLabelsMasker, but the labels_ attribute gets overwritten after each iteration. This could be fixed by wrapping the transform_single_imgs function in another function that recovers this attribute and allocates it in a list for each iteration of the parallel loop (any ideas?)
- For the other strategies, that are handled at fit time, I don't know what would be the best way to work around it for the 4D + 1D case

Other questions:
- Inverse transform (not implemented) would have a similar problem: If several labels_img are needed, it could be a bit problematic afterwards.
- What else should be tested for this class?

Issue #2247, opened 2019-12-04
Allow NiftiMapsMasker.fit_transform to take a list of images as input


koudyk:
**What would you like changed/added and why?**
Allow NiftiMapsMasker.fit_transform to take a list of images as input, just like how ConnectivityMeasure.fit_transform can take a list of inputs.  

**What would be the benefit? Does the change make something easier to use?
Clarifies something?**
It would make the functions more consistent, and it might be easier for some users to input a list instead of building their own for loop.

jeromedockes:
Thanks! could you provide a small example showing what is the problem? Passing a
list of images to `fit_transform` seems to work for me, as well as passing the
atlas to `__init__` as a list.

for example does this work on your machine?:

```python
from nilearn import datasets, input_data, image

atlas = datasets.fetch_atlas_msdl()
masker = input_data.NiftiMapsMasker(atlas["maps"])
mni = datasets.load_mni152_template()
masked = masker.fit_transform([mni, mni, mni])
print(masker.maps_img_.shape)
print(masked.shape)

atlas_maps = list(image.iter_img(atlas["maps"]))
print(atlas_maps[:3])
masker = input_data.NiftiMapsMasker(atlas_maps)
mni = datasets.load_mni152_template()
masked = masker.fit_transform([mni, mni, mni])
print(masker.maps_img_.shape)
print(masked.shape)
```

koudyk:
Thanks for the quick reply! 
The code you gave does work on my machine, but my code still does not work. The difference that stands out to me is that I give a list of confounds as well as a list of images. Is it possible that the function can't take a list of confounds? I can't tell from the docs or the code (probably because of my lack of experience more than the lack of clarity). 

Here is my code:
```
from nilearn.input_data import NiftiLabelsMasker
from nilearn import datasets

development_dataset = datasets.fetch_development_fmri(n_subjects=90)
multiscale = datasets.fetch_atlas_basc_multiscale_2015(version='sym')
atlas_filename = multiscale.scale064
masker = NiftiLabelsMasker(labels_img=atlas_filename, standardize=True,
                           memory='nilearn_cache', verbose=0)
time_series_all_sub = masker.fit_transform(development_dataset.func, confounds=development_dataset.confounds)
```

KamalakerDadi:
maps masking is done on 4D as input.

[...]

PR #3237, opened 2022-05-04
[ENH] Add multi-subject maskers to handle list of subjects

### LINKED ISSUE #1240: Enhancement suggestion: MultiNiftiMapsMasker / MultiNiftiLabelsMasker
The example http://nilearn.github.io/auto_examples/03_connectivity/plot_multi_subject_connectome.html coud be greatly simplified if the NiftiMapsMasker could take a list of subjects / confounds.

However, it makes very brittle and awkward input verification to accept both 4D and 4+1D data.

Hence, I suggest to create 2 new classes: MultiNiftiMapsMasker and MultiNiftiLabelsMasker that take 4+1D data. These could be chained into a ConnectivityMeasure and a supervised learning to get a full predictive pipeline.

What do people think?

ymzayek:
Replaces #2294
Closes #1240 and #2247

- [x] Add `MultiNiftiLabelsMasker`
- [x] Add `MultiNiftiMapsMasker`
- [x] Add documentation for both new maskers and for `MultiNiftiMasker`
- [x] Add tests for new maskers
- [x] Include in examples
- [x] Add entry in `doc/changes/latest.rst`


bthirion:
Thx ! 
Do you need a review at that point, or do you prefer to proceed with tests etc .?

ymzayek:
@bthirion I think a review already would be helpful. I added one test and would like feedback on that before proceeding with a second test for the new maps masker.

examples/03_connectivity/plot_atlas_comparison.py
```diff
@@ -59,13 +60,11 @@
 from nilearn import plotting
 
 # create masker to extract functional data within atlas parcels
-masker = NiftiLabelsMasker(labels_img=yeo['thick_17'], standardize=True,
-                           memory='nilearn_cache')
+masker = MultiNiftiLabelsMasker(labels_img=yeo['thick_17'], standardize=True,
```
bthirion, line 63: Through a comment, make it explicit that you use `MultiNiftiLabelsMasker`, so that people notice and understand why.

examples/03_connectivity/plot_atlas_comparison.py
```diff
@@ -127,16 +126,14 @@ def lag_correlation(time_series, lag):
 ##########################################################################
 # Iterate over fetched atlases to extract coordinates - probabilistic
 # -------------------------------------------------------------------
-from nilearn.maskers import NiftiMapsMasker
+from nilearn.maskers import MultiNiftiMapsMasker
 
 # create masker to extract functional data within atlas parcels
-masker = NiftiMapsMasker(maps_img=difumo.maps, standardize=True,
-                         memory='nilearn_cache')
+masker = MultiNiftiMapsMasker(maps_img=difumo.maps, standardize=True,
```
bthirion, line 132: Through a comment, make it explicit that you use `MultiNiftiMapsMasker`, so that people notice and understand why.

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,266 @@
...
+from .._utils import CacheMixin
+from .._utils.niimg_conversions import _iter_check_niimg
+from .nifti_labels_masker import NiftiLabelsMasker
+
+
+class MultiNiftiLabelsMasker(NiftiLabelsMasker, CacheMixin):
+    """Class for masking of Niimg-like objects.
+    MultiNiftiLabelsMasker is useful when data from non-overlapping volumes
+    and from different subjects should be extracted (contrary to
+    NiftiLabelsMasker).
```
bthirion, line 18: For lisibility, drop a line here.

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,266 @@
...
+    labels : list of str, optional
+        Full labels corresponding to the labels image. This is used
+        to improve reporting quality if provided.
+        Warning: The labels must be consistent with the label
+        values provided through `labels_img`.
+
+    background_label: number, optional
+        Label used in labels_img to represent background.
+
+    mask_img: Niimg-like object, optional
```
bthirion, line 34: I guess some of the fields of the docstring can be filled in automatically. See the docuemntation of NiftiLabelsMasker.

nilearn/maskers/tests/test_multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,121 @@
...
+import numpy as np
+
+import nibabel
+import pytest
+from nilearn.maskers import NiftiLabelsMasker, MultiNiftiLabelsMasker
+from nilearn._utils import as_ndarray, data_gen
+from nilearn._utils.exceptions import DimensionError
+
+
+def generate_random_img(shape, length=1, affine=np.eye(4),
```
bthirion, line 15: I guess there are already image generators in the test suite ?

nilearn/maskers/tests/test_multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,121 @@
...
+    # NiftiLabelsMasker should not work with 4D + 1D input
+    signals_input = [fmri11_img, fmri11_img]
+    masker11 = NiftiLabelsMasker(labels11_img, resampling_target=None)
+    with pytest.raises(DimensionError, match="incompatible dimensionality"):
+        masker11.fit_transform(signals_input)
+
+    # No exception should be raised either
+    masker11 = MultiNiftiLabelsMasker(labels11_img, resampling_target=None)
+    masker11.fit()
+    assert signals11.shape == (length, n_regions)
```
bthirion, line 79: This tests nothing

bthirion (commented):
Thx. This is in pretty good tests. I think that most of the things come from the fact that the library has evolved since the initial PR.

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,263 @@
...
+        See http://nilearn.github.io/manipulating_images/input_output.html
+        Mask to apply to regions before extracting signals.
+
+    smoothing_fwhm: float, optional
```
bthirion, line 40:
```suggestion
    %(smoothing_fwhm)s
```

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,263 @@
...
+            This parameter is passed to signal.clean. Please see the
+            corresponding documentation for details.
+
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            shape: (number of scans - number of volumes removed, )
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+
+                .. versionadded:: 0.8.0
```
bthirion, line 249: to be updated

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,263 @@
...
+            documentation for details.
+            shape: list of (number of scans, number of confounds)
+
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            shape: (number of scans - number of volumes removed, )
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+
+                .. versionadded:: 0.8.0
```
bthirion, line 201: to be updated

nilearn/maskers/multi_nifti_maps_masker.py
```diff
@@ -0,0 +1,263 @@
...
+            documentation for details.
+            shape: list of (number of scans, number of confounds)
+
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            shape: (number of scans - number of volumes removed, )
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+
+                .. versionadded:: 0.8.0
```
bthirion, line 199: to be updated

nilearn/maskers/multi_nifti_maps_masker.py
```diff
@@ -0,0 +1,263 @@
...
+            This parameter is passed to signal.clean. Please see the
+            corresponding documentation for details.
+
+        sample_mask : Any type compatible with numpy-array indexing, optional
+            shape: (number of scans - number of volumes removed, )
+            Masks the niimgs along time/fourth dimension to perform scrubbing
+            (remove volumes with high motion) and/or non-steady-state volumes.
+            This parameter is passed to signal.clean.
+
+                .. versionadded:: 0.8.0
```
bthirion, line 249: to be updated

bthirion (commented):
Minor things. Thx !

nilearn/maskers/multi_nifti_labels_masker.py
bthirion: the version added should be 0.9.2

ymzayek: I was a bit unsure because to me it seems like the versionadded is referring specifically to when the `sample_mask` argument was added to the function `transform` so I am not sure it makes sense to put the current version. But also maybe it makes sense to remove the versionadded line entirely since it is a 'new' function in a new class?

bthirion: Then remove the thing indeed.

bthirion (approved):
LGTM

bthirion (commented):
Looks mostly good, thx !

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,259 @@
...
+from .._utils.niimg_conversions import _iter_check_niimg
+from .nifti_labels_masker import NiftiLabelsMasker
+
+
+@fill_doc
+class MultiNiftiLabelsMasker(NiftiLabelsMasker, CacheMixin):
+    """Class for masking of Niimg-like objects.
+    MultiNiftiLabelsMasker is useful when data from non-overlapping volumes
+    and from different subjects should be extracted (contrary to
+    NiftiLabelsMasker).
```
NicolasGensollen, line 19:
```suggestion
    :class:`nilearn.maskers.NiftiLabelsMasker`).
```

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,259 @@
...
+    Parameters
+    ----------
+    labels_img: Niimg-like object
+        See http://nilearn.github.io/manipulating_images/input_output.html
+        Region definitions, as one image of labels.
+
+    labels : list of str, optional
+        Full labels corresponding to the labels image. This is used
+        to improve reporting quality if provided.
+        Warning: The labels must be consistent with the label
```
NicolasGensollen, line 30: Maybe use a `.. warning::` block here for readability.

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,259 @@
...
+        See http://nilearn.github.io/manipulating_images/input_output.html
+        Mask to apply to regions before extracting signals.
+
+    %(smoothing_fwhm)s
+        If smoothing_fwhm is not None, it gives the full-width-at-half-maximum
+        in millimeters of the spatial smoothing to apply to the signal.
+
+    standardize: {'zscore', 'psc', True, False}, default is 'zscore'
+        Strategy to standardize the signal.
+        'zscore': the signal is z-scored. Timeseries are shifted
```
NicolasGensollen, line 46: Maybe use a bullet point list here for readability.

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,259 @@
...
+        documentation for details
+
+    dtype: {dtype, "auto"}
+        Data type toward which the data should be converted. If "auto", the
+        data will be converted to int32 if dtype is discrete and float32 if it
+        is continuous.
+
+    resampling_target: {"data", "labels", None}, optional.
+        Gives which image gives the final shape/size. For example, if
+        `resampling_target` is "data", the atlas is resampled to the
```
NicolasGensollen, line 86: I think a bullet point list for the options would also be more readable here. WDYT?

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,259 @@
...
+        shape of the data if needed. If it is "labels" then mask_img
+        and images provided to fit() are resampled to the shape and
+        affine of maps_img. "None" means no resampling: if shapes and
+        affines do not match, a ValueError is raised. Defaults to "data".
+
+    memory: joblib.Memory or str, optional
```
NicolasGensollen, line 92:
```suggestion
    memory: :obj:`joblib.Memory` or :obj:`str`, optional
```

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,259 @@
...
+        Used to cache the region extraction process.
+        By default, no caching is done. If a string is given, it is the
+        path to the caching directory.
+
+    memory_level: int, optional
```
NicolasGensollen, line 97: I think some of these arguments are in the standard docs

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+
+    def __init__(self, labels_img, labels=None, background_label=0,
+                 mask_img=None, smoothing_fwhm=None, standardize=False,
+                 standardize_confounds=True, high_variance_confounds=False,
+                 detrend=False, low_pass=None, high_pass=None, t_r=None,
+                 dtype=None, resampling_target="data",
+                 memory=Memory(cachedir=None, verbose=0), memory_level=1,
+                 n_jobs=1, verbose=0, strategy="mean", reports=True):
+        self.labels_img = labels_img
```
NicolasGensollen, line 132: All of this can be removed as this class is inheriting from `NiftiLabelsMasker`. Simply call the parent constructor here.

nilearn/maskers/multi_nifti_maps_masker.py
```diff
@@ -0,0 +1,259 @@
...
+from .nifti_maps_masker import NiftiMapsMasker
+
+
+@fill_doc
+class MultiNiftiMapsMasker(NiftiMapsMasker, CacheMixin):
+    """Class for masking of Niimg-like objects.
+
+    MultiNiftiMapsMasker is useful when data from overlapping volumes
+    and from different subjects should be extracted (contrary to
+    NiftiMapsMasker).
```
NicolasGensollen, line 20:
```suggestion
    :class:`nilearn.maskers.NiftiMapsMasker`).
```

nilearn/maskers/multi_nifti_maps_masker.py
```diff
@@ -0,0 +1,259 @@
...
+
+    Note that, Inf or NaN present in the given input images are automatically
```
NicolasGensollen, line 22: Maybe use a `.. note::` here

nilearn/maskers/multi_nifti_maps_masker.py
```diff
@@ -0,0 +1,259 @@
...
+        to zero mean and scaled to unit variance.
+        False : Do not standardize the data.
+        Default=False.
+
+    standardize_confounds : boolean, optional
+        If standardize_confounds is True, the confounds are z-scored:
+        their mean is put to 0 and their variance to 1 in the time dimension.
+        Default=True.
+
+    high_variance_confounds : boolean, optional
```
NicolasGensollen, line 56: Same comment, I think some of these are in the standard docs

nilearn/maskers/multi_nifti_maps_masker.py
```diff
@@ -0,0 +1,259 @@
...
+    # memory and memory_level are used by CacheMixin.
+
+    def __init__(self, maps_img, mask_img=None,
+                 allow_overlap=True, smoothing_fwhm=None, standardize=False,
+                 standardize_confounds=True, high_variance_confounds=False,
+                 detrend=False, low_pass=None, high_pass=None, t_r=None,
+                 dtype=None, resampling_target="data",
+                 memory=Memory(location=None, verbose=0), memory_level=0,
+                 n_jobs=1, verbose=0, reports=True):
+        self.maps_img = maps_img
```
NicolasGensollen, line 133: Same comment about the parent constructor

NicolasGensollen (commented):
Thanks for working on this @ymzayek !
I made a quick pass with some suggestions on formatting. 
More important, I think we need to use class inheritance as to not duplicate code. Theoretically these two classes should have very little additional code compared to their parent classes.

ymzayek:
@NicolasGensollen and @bthirion I think this is ready for another round of reviews. I incorporated all the feedback as best as I could and additionally I made `transform_imgs` into a private method since it is only called internally. I think this can also be done to `transform_single_imgs` of the parent classes but this PR is already pretty big so I would handle that in another PR. Also more standardisation can probably be done for the docstrings across all the maskers, but I think that should also be handled in a separate PR.

nilearn/maskers/tests/test_multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,321 @@
...
+
+from nilearn._utils import testing, data_gen
+from nilearn._utils.exceptions import DimensionError
+from nilearn.image import get_data
+from nilearn.maskers import NiftiLabelsMasker, MultiNiftiLabelsMasker
+
+
+def test_multi_nifti_labels_masker():
+    # Check working of shape/affine checks
+    shape1 = (13, 11, 12)
```
NicolasGensollen, line 18: This test data generation is the same as in the tests for the `NiftiLabelsMasker`. 
Do you think we could define it as a (or multiple..??) fixture(s) in a separate file and import it in both test files? `nilearn/_utils/data_gen.py` could also be a good place to put it...

nilearn/maskers/tests/test_multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,321 @@
...
+
+    img_data = np.array([[test_values,
+                          test_values]])
+
+    labels_data = np.array([[[0, 0, 0, 0, 0],
+                             [1, 1, 1, 1, 1]]], dtype=np.int8)
+
+    affine = np.eye(4)
+    img = nibabel.Nifti1Image(img_data, affine)
+    labels = nibabel.Nifti1Image(labels_data, affine)
```
NicolasGensollen, line 136: I think you can refactor this test by:

- using a fixture for the data generation step
- parametrize for the strategy

nilearn/maskers/tests/test_multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,321 @@
...
+    for t in transformed:
+        assert t.shape == (length, n_regions)
+
+        fmri11_img_r = masker.inverse_transform(t)
+        np.testing.assert_almost_equal(fmri11_img_r.affine,
+                                       masker.labels_img_.affine)
+        assert (fmri11_img_r.shape == (
+                masker.labels_img_.shape[:3] + (length,)))
+
+    # Test with clipped labels: mask does not contain all labels.
```
NicolasGensollen, line 220: Looking at this really long function, I feel like it could be split in at least 3 different functions. The fact that new data are generated at multiple places is a good indicator. WDYT?

nilearn/maskers/tests/test_multi_nifti_maps_masker.py
```diff
@@ -0,0 +1,254 @@
...
+    # Check working of shape/affine checks
+    shape1 = (13, 11, 12)
+    affine1 = np.eye(4)
+
+    shape2 = (12, 10, 14)
+    affine2 = np.diag((1, 2, 3, 1))
+
+    n_regions = 9
+    length = 3
+
```
NicolasGensollen, line 25: More or less the same comments as for the previous test file.

NicolasGensollen (commented):
Thanks for addressing my points @ymzayek ! 
I think this is looking good ! 👍 
I made another round with some suggestions (see below).

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,187 @@
...
+    nilearn.maskers.NiftiMasker
+    nilearn.maskers.NiftiLabelsMasker
+
+    """
+
+    def __init__(self, *args, n_jobs=1, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.n_jobs = n_jobs
+
+    def _transform_imgs(self, imgs_list, confounds=None, n_jobs=1,
```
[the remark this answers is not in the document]
ymzayek: What I found a bit odd for all the maskers is that this and the `transform` method are both exposed which may confuse people about the difference or which to use if they only look at the API.
But I do agree that we shouldn't be inconsistent. I was thinking that maybe it should be private for all the maskers. For now though I'm ok for making it public again and discussing this further later on. Also maybe better to handle in another PR. As I mentioned a follow-up to make the docstrings consistent across all the maskers seems necessary.

NicolasGensollen:
@ymzayek Sorry, I completely missed your previous message.
If this is the only thing blocking this PR from getting merged, I think we can leave it like this, and I will try to propose a refactoring of these tests in a following PR when I'll have some time.

ymzayek:
@NicolasGensollen yes then I would say let's do it in a separate PR. It is something I wanted to do anyways on a bigger scale (see https://github.com/nilearn/nilearn/issues/3331) but haven't gotten around to it yet. But it would be good to do it in several PRs, maybe starting with the `maskers` module.

ymzayek:
@NicolasGensollen ok thanks! I'll open an issue to reference the follow-up work that still needs to be done.

Also the first approval is from a while ago, @bthirion do you want to have another look?

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,167 @@
...
+
+from joblib import Parallel, delayed
+
+from .._utils import CacheMixin, fill_doc
+from .._utils.niimg_conversions import _iter_check_niimg
+from .nifti_labels_masker import NiftiLabelsMasker
+
+
+@fill_doc
+class MultiNiftiLabelsMasker(NiftiLabelsMasker, CacheMixin):
```
jeromedockes, line 15: niftilabelsmaskers is already derived from CacheMixin so I don't think we need CacheMixin here

nilearn/maskers/multi_nifti_labels_masker.py
```diff
@@ -0,0 +1,167 @@
...
+         Default=True.
+
+    See also
+    --------
+    nilearn.maskers.NiftiMasker
+    nilearn.maskers.NiftiLabelsMasker
+
+    """
+
+    def __init__(self, *args, n_jobs=1, **kwargs):
```
jeromedockes, line 102: I would prefer to have the full function signature here rather than forwarding *args, **kwargs

bthirion (approved):
All good with me, thx again !

bthirion:
Merging, then.

PR #3248, opened 2022-05-11
[MAINT] Remove extraneous `doc` requirements

### LINKED ISSUE #3244: Setting up nilearn development environment per instructions fails as tbb is not installed
Following [the instructions](https://nilearn.github.io/stable/development.html#contribution-guidelines) in the contribution guidelines fails because tbb is not installed.

Nilearn version: 0.9.1
macOS 12.3.1, Homebrew 3.4.10, Python 3.9.12

### Expected behavior

No errors.

### Actual behavior

```
PackagesNotFoundError: The following packages are missing from the target environment:
  - tbb
```

Which required `conda install tbb`.

tsalo:
Closes #3244.

- Remove `mkl`, `cython`, and `pillow` from the documentation requirement group in `setup.cfg`.
- Remove `cython` from the testing requirement group.
- Reference the `plotly` requirement group in the documentation requirement group instead of duplicating the requirements.

Open questions:

- [x] Do we need `cython`?
    - Cython is mentioned in the Makefile, but only in reference to `pyx` files, and I couldn't find any of those.
- [x] Do we need `pillow`?
    - Neither `pillow` nor `PIL` appears outside `setup.cfg`.
- [x] Is `memory_profiler` necessary for building the documentation, or is it used for tests?
    - Per @NicolasGensollen, it's used to profile the documentation build.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,6 @@ doc =
 	flake8
 	kaleido
  	memory_profiler
-	mkl
 	numpydoc
 	pillow
 	plotly
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ doc =
 	%(plotly)s
 	coverage
 	flake8
- 	memory_profiler
+ 	memory_profiler  # measuring memory during docs building
 	numpydoc
 	pillow
 	sphinx
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,6 @@ doc =
 	flake8
  	memory_profiler  # measuring memory during docs building
 	numpydoc
-	pillow
 	sphinx
 	sphinx-copybutton
 	sphinx-gallery
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,14 +27,13 @@ plotly =
 
 # Requirements necessary for building the documentation
 doc =
+	%(plotly)s
 	cython
 	coverage
 	flake8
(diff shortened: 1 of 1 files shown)


bthirion:
I'm surprised that we need Cython. We should not.

bthirion (approved):
LGTM.

NicolasGensollen (approved):
LGTM, thanks @tsalo !
- I added `memory_profiler` a while ago when I was investigating some memory issues when building the docs in the CI. I don't have a strong opinion on whether we should keep it or not. It doesn't hurt to have some info on how much memory is consumed by each example, so I would be slightly in favor of keeping it.
- I don't think we need `Cython` either.

tsalo:
Thanks! I'll keep `memory_profiler` and drop `cython`. Do you know if `pillow` is being used for anything?

NicolasGensollen:
> Do you know if pillow is being used for anything?

I don't think it is used for anything.

tsalo:
Okay, I've removed that one as well. If tests pass without issue, then I think this will be good to merge.

bthirion:
Regarding MemoryProfiler, there seems to be some complaints from users: https://neurostars.org/t/nilearn-development-environment-fails-with-memory-error/22537/5

bthirion (approved):

bthirion:
Merging, then.

Issue #3253, opened 2022-05-13
`plot_markers` can only plot n>=2 scatter points


caglorithm:
`plot_markers` returns an error if you attempt to plot only one point. 

This works: 
```python
plot_markers(np.ones(2), np.ones((2, 3))) # works
```

### Steps and code to reproduce bug

```python
plot_markers(np.ones(1), np.ones((1, 3))) # does not work
```

Returns 
```bash
ValueError: Dimension mismatch: 'node_values' should be vector of length 1, but current shape is () instead of (1,)
```

NicolasGensollen:
Fixed by #3255

PR #3255, opened 2022-05-13
[Fix] plot_markers: remove squeeze for plotting single scatter point

caglorithm:
This PR fixes issue #3253: `plot_markers`'s ability to plot a single dot. 

Code before had an `np.squeeze()` that deleted the single-value entry in `node_values` when plotting a single point. I removed it entirely and plotting worked fine. 

However, I don't know why it is there in the first place so this PR keeps it in if `ndim > 1`. If there is no specific reason why this `np.squeeze` is in, I would suggest to drop it entirely.


RaphaelMeudec:
Thanks for raising the issue and launching this PR!

Removing the squeeze alltogether is problematic, as it handles the case where user provides `node_values` as an array of shape `(N, 1)` or `(1, N)`. I guess one fix we could do that would solve this issue and at the same time avoid checking the ndim is to replace `squeeze` by a `flatten`. This would handle the `(N, 1)` case while solving the 1 element array case.

Let's wait for an other opinion on this!

caglorithm:
Agreed, thanks for explaining. My tests work with a `flatten()` and I agree it's a more elegant solution.

NicolasGensollen (commented):
Thanks for working on this @caglorithm ! 👍 
Could you add a non-regression test (failing with previous behavior and passing with your fix)?

caglorithm:
Done! Fails on main, works with this PR.

bthirion (approved):
LGTM, thx !

nilearn/plotting/tests/test_img_plotting/test_plot_markers.py
```diff
@@ -150,3 +150,8 @@ def test_plot_markers_threshold_errors(coords):
     with pytest.raises(ValueError,
                        match="Provided 'node_threshold' value"):
         plot_markers([1, 2, 2, 4], coords, node_threshold=5, display_mode='x')
+
```
NicolasGensollen, line 153: Could you add another line to make PEP8 happy?

NicolasGensollen (commented):
LGTM, thanks @caglorithm !
Could you add a what's_new entry in `doc/changes/latest.rst` in the `FIX` section? Since this is your first contribution, you'll need to add your name/profile in `doc/changes/names.rst`.

caglorithm:
Done!

doc/changes/latest.rst
```diff
@@ -15,6 +15,7 @@ Fixes
 
 - Convert references in ``nilearn/mass_univariate/permuted_least_squares.py`` to use bibtex format (:gh:`3222` by `Yasmin Mzayek`_).
 - :func:`~plotting.plot_roi` failed before when used with the "contours" view type and passing a list of cut coordinates in display mode "x", "y" or "z"; this has been corrected (:gh:`3241` by `Jerome Dockes`_).
+- :func:`~plotting.plot_markers` can now plot a single scatter point (:gh:`3253` by `Caglar Cakan`_).
```
NicolasGensollen, line 18:
```suggestion
- :func:`~plotting.plot_markers` can now plot a single scatter point (:gh:`3255` by `Caglar Cakan`_).
```

NicolasGensollen (approved):
Thanks for addressing my points @caglorithm !
Besides the PR number (see below), everything LGTM!

caglorithm:
Thanks, I've mixed up the PR number with the issue number. Now fixed.

NicolasGensollen:
Thanks @caglorithm !
Merging...

PR #3259, opened 2022-05-13
[ENH] Add option for user to specifiy angle directly via view argument

### LINKED ISSUE #3258: Allow user to specify view angle explicitly (for surface plotting)
Allow users to pass (elev, azim) as a tuple or list (or generally, sequence of length 2), instead of only a string. 

### Benefits to the change
This will help for visualizing features not captured entirely in any of the limited number of default views. I generate many images, and it is infeasible to manually (interactively) select the appropriate view for each image individually. Rather, I would prefer to do this programatically. 

```python
view = nilearn.plotting.plot_surf(mesh, map, view=(elev, azim))
```

I will submit a PR to this effect, but would appreciate nilearn maintainer's help in completing it.

sambjohnson:
Closes #3258 .

- I'm proposing to allow users to manually set the view in surfact_plotting.py functions by explicitly supplying the view angle as (elev, azim) as in the "view" argument. This is a relatively simple and self-contained addition to existing functionality. However, I have not updated documentation that would notify users that the "view" argument of surface plotting functions can optionally be a length 2 sequence instead of a string (e.g., 'ventral', 'lateral', etc.). Moreover, because this change relies on the representation of view as a pair (elev, azim), it only functions in the maplotlib engine, not the plotly engine. I could imagine extending this functionality so that view angles can always be specified explicitly across all surface plotting functions, but this would greatly expand the scope of work from the present, small change.


bthirion (commented):
Looks good so far.
Here are some suggestions:
* Add an example (by proposing the view in one of the examples through a numerical specification)
* Add a test in test_surf_plotting.py, e.g.  in the  test_plot_img_on_surf_hemispheres_and_orientations():
to include manually-supplied views.
* regarding the use with plotly, sorry I don't know, I'm nit very familiar with it. may be @jeromedockes would know.

sambjohnson:
Understood -- and thanks for reviewing so quickly. I'm very new to submitting contributions in general; _is there somewhere I should look for a guide on how to contribute examples?_ There are a few examples that I think would be easy, e.g. using fsaverage. I'm just unsure of how to add them to the library of nilearn examples. Sorry if this is answered somewhere in the documentation!

bthirion:
Regarding contributing in general, you have probably seen this page: https://nilearn.github.io/stable/development.html
Developing examples is somehow the easiest part: For instance, you may want to just add a custom view to the https://nilearn.github.io/stable/auto_examples/01_plotting/plot_surf_atlas.html example. For this, you simply need to edit this file: https://github.com/nilearn/nilearn/blob/main/examples/01_plotting/plot_surf_atlas.py.
Is it clear to you ?
Best,

sambjohnson:
Ah, I see. For some reason, I had trouble understanding the relation between the official examples and their location in the code base, but your answer clarifies that. Thanks!

NicolasGensollen (commented):
Thanks for working on this @sambjohnson ! 👍 
In addition to the points of @bthirion, I think that the documentation of the public plotting functions should be updated to reflect the fact that:

- `view` can now be either a string or a length 2 iterable (and the latter option is only available with the Matplotlib engine, at least for now)
- If `view` is a list, then the `hemi` parameter isn't used to set the view

Let us know if it is unclear how to update the docs.

nilearn/plotting/surf_plotting.py
```diff
@@ -446,7 +447,12 @@ def _plot_surf_matplotlib(coords, faces, surf_map=None, bg_map=None,
     limits = [coords.min(), coords.max()]
 
     # set view
-    elev, azim = _set_view_plot_surf_matplotlib(hemi, view)
+    if isinstance(view, str):
+        elev, azim = _set_view_plot_surf_matplotlib(hemi, view)
+    elif isinstance(view, collections.Sequence) and len(view) == 2:
+        elev, azim = view
+    else:
+        raise TypeError(f"view must be a string of a sequence of length 2. Got {view} instead")
```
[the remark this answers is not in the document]
sambjohnson: Thanks, @NicolasGensollen! I had one quick style question: do you have an opinion about whether I should change the documentation manually in surface plotting functions or whether I should attempt to change it through `@filldoc`? I think that the latter would be possible, but I might have to create a new entry the docdict in _utils/docs.py, e.g. `custom_view`. (Otherwise, some functions that use a view but that do not support a custom view might also have their docstrings erroneously changed.)

This may be a small concern; I'm just curious whether you would have a preference.

sambjohnson:
Following up on your recommendations, I have implemented some basic tests, examples, and documentation appropriate to the new option to specify the `view` argument in surface plotting functions as a pair of angles (in degrees) in addition to the previous option, a string. I believe this enhancement is now ready to be integrated into nilearn!

bthirion:
Thx ! 
Please check the pep8 failure. The test error is a bit mysterious to me...

nilearn/plotting/surf_plotting.py
```diff
@@ -446,7 +447,13 @@ def _plot_surf_matplotlib(coords, faces, surf_map=None, bg_map=None,
     limits = [coords.min(), coords.max()]
 
     # set view
-    elev, azim = _set_view_plot_surf_matplotlib(hemi, view)
+    if isinstance(view, str):
+        elev, azim = _set_view_plot_surf_matplotlib(hemi, view)
+    elif isinstance(view, collections.Sequence) and len(view) == 2:
```
NicolasGensollen, line 452: `collections.Sequence` is deprecated since Python 3.3 and was scheduled to be removed in Python 3.10, which is why the tests with Python 3.10 are currently failing. You need to import `Sequence` from `collections.abc`:

```python
from collections.abc import Sequence
```

nilearn/plotting/surf_plotting.py
```diff
@@ -1054,19 +1061,43 @@ def _check_hemispheres(hemispheres):
...
+
 def _check_views(views) -> list:
     """Checks whether the views passed to in plot_img_on_surf are
     correct.
 
     views : list
-        Any combination of "anterior", "posterior", "medial", "lateral",
-        "dorsal", "ventral".
+        Any combination of string in {"anterior", "posterior", "medial", "lateral",
+        "dorsal", "ventral"} and / or pair of floats (elev, azim).
```
ymzayek, line 1087:
```suggestion
        Any combination of string in {"anterior", "posterior", "medial", 
        "lateral", "dorsal", "ventral"} and / or pair of floats (elev, azim).
```

nilearn/_utils/docs.py
```diff
@@ -504,10 +504,17 @@ def custom_function(vertices):
...
+        If a string, must be in {'lateral', 'medial', 'dorsal', 'ventral',\
+        'anterior', 'posterior'}.
+        If a sequence, must be a pair (elev, azim) of float
+        angles in degrees that will manually set a custom view.
+        E.g., view=[270.0, 90.0] or view=(0.0, -180.0). Note that this
+        manual option is only available when the plotting engine is matplotlib.
+        Otherwise, the view can only be specified via one of the pre-set strings.
     View of the surface that is rendered.
     Default='lateral'.
+
```
ymzayek, line 517:
```suggestion

```

nilearn/_utils/docs.py
```diff
@@ -504,10 +504,17 @@ def custom_function(vertices):
...
-view : {'lateral', 'medial', 'dorsal', 'ventral',\
-        'anterior', 'posterior'}, optional
+view : either a string or pair of floats, optional.
+        If a string, must be in {'lateral', 'medial', 'dorsal', 'ventral',\
+        'anterior', 'posterior'}.
+        If a sequence, must be a pair (elev, azim) of float
+        angles in degrees that will manually set a custom view.
+        E.g., view=[270.0, 90.0] or view=(0.0, -180.0). Note that this
+        manual option is only available when the plotting engine is matplotlib.
+        Otherwise, the view can only be specified via one of the pre-set strings.
```
ymzayek, line 514:
```suggestion
        Otherwise, the view can only be specified via one of the pre-set
        strings.
```

ymzayek (commented):
Thx @sambjohnson! I made a few suggestions to address the PEP8 issues. Can you also address the comments from the other reviewers and add an entry to `doc/changes/latest.rst` and as a first time contributor add your name and GitHub profile in `doc/changes/names.rst`? Finally, a couple of additional tests need to be added, see [codecov/patch report](https://github.com/nilearn/nilearn/pull/3259/checks?check_run_id=6842357283)

nilearn/plotting/surf_plotting.py
jeromedockes: I think this work (including raising the exception) should be done in `_set_view_plot_surf_matplotlib`. (btw its name probably shouldn't be `set...` as it doesn't modify any object but returns values it computed

jeromedockes: also this should use the `_check_view_is_valid` helper instead of duplicating its logic

nilearn/plotting/tests/test_surf_plotting.py
```diff
@@ -645,6 +645,11 @@ def test_plot_img_on_surf_hemispheres_and_orientations():
     plot_img_on_surf(nii,
                      hemispheres=['left', 'right'],
                      views=['dorsal', 'medial'])
+    # Check that manually set view angles work.
```
jeromedockes, line 648: this checks it doesn't raise an exception but could we check the view is set correctly?

nilearn/_utils/docs.py
```diff
@@ -504,10 +504,17 @@ def custom_function(vertices):
...
 docdict['view'] = """
-view : {'lateral', 'medial', 'dorsal', 'ventral',\
-        'anterior', 'posterior'}, optional
+view : either a string or pair of floats, optional.
+        If a string, must be in {'lateral', 'medial', 'dorsal', 'ventral',\
+        'anterior', 'posterior'}.
+        If a sequence, must be a pair (elev, azim) of float
+        angles in degrees that will manually set a custom view.
+        E.g., view=[270.0, 90.0] or view=(0.0, -180.0). Note that this
+        manual option is only available when the plotting engine is matplotlib.
```
jeromedockes, line 513: I am concerned about adding options that only work for one of the plotting backends. could we have a conversion from elevation, azimuth to plotly's eye, up, center view specification (or the other way around), so that the view can be specified in the same way(s) for both?

nilearn/_utils/docs.py
sambjohnson: Thanks for the fixes - I will have time this week to incorporate the handful of suggestions!

nilearn/_utils/docs.py
sambjohnson: I was unsure whether it would be worth it, given that the plotly engine is still experimental (if I understand correctly).

If you think it is worth prioritizing this conversion logic so that either view can be used with either engine, then I will look into it (with the caveat that it might take a little while because I have not yet put in the time to understand the (eye, up, center) convention.

jeromedockes: sorry I replied by email but it seems that got lost:

> I was unsure whether it would be worth it, given that the plotly engine is still experimental (if I understand correctly).
> If you think it is worth prioritizing this conversion logic so that either view can be used with either engine, then I will look into it (with the caveat that it might
> take a little while because I have not yet put in the time to understand the (eye, up, center) convention.

That makes sense. Still I think it could be worth comparing conventions
and figuring out if a (approximate) conversion in either or both
directions is possible, and use the one that is easiest to convert to
the other for the new parameter.

If it's hard we can leave the plotly part for another PR, my hesitation
is just that then we may take a long time to get around to it and the
interface of that function will be more complex because of the
discrepency between the plotly and matplotlib engines

bthirion: If it's  not too much of a hassle, it would be great indeed to have the plotly version altogether.
Thx again for your contribution !

alexisthual:
Hi @sambjohnson, I think this is a very interesting feature and I'm willing to spend some time with you if I can be of any help it getting this PR in a state where it can be merged :slightly_smiling_face: 

If I understand correctly, you still need to:
- make the tests pass (always a tricky part haha)
- make your changes available for `plotly`, which [seems straightforward](https://plotly.com/python/3d-camera-controls/) and may fit in this PR or in a different one

nilearn/plotting/surf_plotting.py
```diff
@@ -107,22 +109,19 @@
...
-                else VALID_HEMISPHERES[1])
-    if view not in CAMERAS:
-        raise ValueError(f"view must be one of {VALID_VIEWS}")
+    if _check_views([view]) and _check_hemispheres([hemi]):
+        if view == 'lateral':
+            view = hemi
+        elif view == 'medial':
+            view = (VALID_HEMISPHERES[0]
+                    if hemi == VALID_HEMISPHERES[1]
+                    else VALID_HEMISPHERES[1])
```
bthirion, line 124: Still need to handle the case where view = (elev, azim)

alexisthual:
I took over this PR because I think it is a nice feature :)
It took a bit of trigonometry but I think I found a way to convert matplotlib's `elev, azim` into sensible settings for plotly.

<details>
<summary>Here is a test script I used to check that both engines render the same images when given similar values for `(elev, azim)`:</summary>

```python
# %%
import math

from itertools import product

from nilearn.plotting.surf_plotting import (
    AXIS_CONFIG,
    LAYOUT,
    _get_view_plot_surf_plotly,
)
from plotly.subplots import make_subplots

# %%
surf_map = surface.vol_to_surf(stat_img, fsaverage.pial_left)

# %%
# 1. Test multiple elevation / azimut values

# %%
v = [0, 45, 90, 135, 180, 225, 270, 315]

# %%
fig = plt.figure(figsize=(2 * len(v), 2 * len(v)))
for (i, elev), (j, azim) in product(enumerate(v), repeat=2):
    ax = fig.add_subplot(len(v), len(v), 1 + i * len(v) + j, projection="3d")
    view = (elev, azim)
    plotting.plot_surf_stat_map(
        fsaverage.pial_left,
        surf_map,
        view=view,
        title=f"elev={elev},\nazim={azim}",
        colorbar=False,
        axes=ax,
    )

plt.show()

# %%
fig = make_subplots(
    rows=len(v),
    cols=len(v),
    specs=[[{"type": "surface"}] * len(v)] * len(v),
    subplot_titles=[
        f"elev={elev},\nazim={azim}"
        for (i, elev), (j, azim) in product(enumerate(v), repeat=2)
    ],
)

for (i, elev), (j, azim) in product(enumerate(v), repeat=2):
    view = (elev, azim)
    sub_fig = plotting.plot_surf_stat_map(
        fsaverage.pial_left,
        surf_map,
        view=view,
        title=f"elev={elev},\nazim={azim}",
        engine="plotly",
    ).figure
    fig.add_trace(sub_fig.data[0], row=i + 1, col=j + 1)
    camera_view = _get_view_plot_surf_plotly("left", view)
    fig.update_layout(
        {
            f"scene{len(v) * i + j + 1}": {
                "dragmode": "orbit",
                "camera": camera_view,
                "aspectmode": "data",
                "aspectratio": {"x": 1.5, "y": 1.5, "z": 1.5},
                **{f"{dim}axis": AXIS_CONFIG for dim in ("x", "y", "z")},
            }
        }
    )

fig.update_layout(
    height=200 * len(v),
    width=200 * len(v),
    **{
        "paper_bgcolor": "#fff",
        "hovermode": False,
        "margin": {"l": 0, "r": 0, "b": 0, "t": 0, "pad": 0},
    },
)
fig.show(renderer="png")

# %%
# 2. Test single elevation / azimut

# %%
elev, azim = (225, 90)

# %%
view = (elev, azim)
plotting.plot_surf_stat_map(
    fsaverage.pial_left,
    surf_map,
    view=view,
    title=f"elev={elev}, azim={azim}",
)
plt.show()

# %%
    fsaverage.pial_left,
    surf_map,
    view=view,
    title=f"elev={elev}, azim={azim}",
    engine="plotly",
)
fig.figure.show(renderer="nteract")


# %%
# 3. Test setting plotly camera view from elevation and azimut
def camera_from_view(view):
    r = 1.7
    elev, azim = view
    # The camera position and orientation is set by three 3d vectors,
    # whose coordinates are independent of the plotted data.
    return {
        # Where the camera should look at
        # (it should always be looking at the center of the scene)
        "center": {"x": 0, "y": 0, "z": 0},
        # Where the camera should be located
        "eye": {
            "x": (
                r
                * math.cos(azim / 360 * 2 * math.pi)
                * math.cos(elev / 360 * 2 * math.pi)
            ),
            "y": (
                r
                * math.sin(azim / 360 * 2 * math.pi)
                * math.cos(elev / 360 * 2 * math.pi)
            ),
            "z": r * math.sin(elev / 360 * 2 * math.pi),
        },
        # How the camera should be rotated.
        # It is determined by a 3d vector indicating which direction
        # should look up in the generated plot
        "up": {
            # "x": 0,
            "x": -math.sin(elev / 360 * 2 * math.pi + math.pi) * math.cos(
                azim / 360 * 2 * math.pi + math.pi
            ),
            "y": -math.sin(elev / 360 * 2 * math.pi + math.pi) * math.sin(
                azim / 360 * 2 * math.pi + math.pi
            ),
            # "y": math.sin(elev / 360 * 2 * math.pi),
            "z": math.cos(elev / 360 * 2 * math.pi),
        },
        "projection": {"type": "perspective"},
        # "projection": {"type": "orthographic"},
    }


# %%
view = (elev, azim)
    fsaverage.pial_left,
    surf_map,
    title=f"elev={elev}, azim={azim}",
    engine="plotly",
)

camera_view = camera_from_view(view)
print(camera_view)
fig.figure.update_layout(
    scene_camera=camera_view,
    # scene_camera=CAMERAS["ventral"],
    # scene={f"{dim}axis": AXIS_CONFIG for dim in ("x", "y", "z")},
    # scene_dragmode="orbit",
    **LAYOUT,
)

fig.figure.show(renderer="png")
```
</details>

Matplotlib
view_matplotlib

Plotly
view_plotly

In particular, I had to change the `projection` parameter to `"orthographic"` in plotly so that both engines would look alike.
It was previously set to `"perspective"` by default, which yields the following image:
view_plotly2

I would advocate that `"orthographic"` is better than `"perspective"` as it makes our images similar across engines.
We might want to zoom in a bit though, which is not as easy when you use the `"orthographic"` perspective (I believe one can tweak the `aspectratio` but I'll investigate this later).

bthirion:
Great ? Do you need another review now ?

nilearn/_utils/docs.py
```diff
@@ -567,10 +567,18 @@ def custom_function(vertices):
...
```
ymzayek, line 576:
```suggestion
        E.g., view=[270.0, 90.0] or view=(0.0, -180.0).
```

nilearn/_utils/docs.py
```diff
@@ -567,10 +567,18 @@ def custom_function(vertices):
...
+        Otherwise, the view can only be specified via one of the pre-set
+        strings.
```
ymzayek, line 578:
```suggestion
```

nilearn/plotting/surf_plotting.py
```diff
@@ -1082,6 +1139,11 @@ def plot_surf_stat_map(surf_mesh, stat_map, bg_map=None,
     return display
 
 
+def _check_hemisphere_is_valid(hemi):
+    valid = hemi in VALID_HEMISPHERES
+    return valid
```
ymzayek, line 1144:
```suggestion
def _check_hemisphere_is_valid(hemi):
    return hemi in VALID_HEMISPHERES
```

alexisthual:
It should be ready for review now!

nilearn/plotting/surf_plotting.py
```diff
@@ -1090,26 +1152,59 @@ def _check_hemispheres(hemispheres):
...
+        "dorsal", "ventral" or pair of floats (elev, azim).
+
+    returns: True if view is valid, False otherwise.
+    """
+    valid = False
+    if isinstance(view, str) and (view in VALID_VIEWS):
+        valid = True
+    elif isinstance(view, Sequence) and len(view) == 2:
+        valid = True
+    return valid
```
ymzayek, line 1183:
```suggestion
    if isinstance(view, str) and (view in VALID_VIEWS):
        return True
    elif isinstance(view, Sequence) and len(view) == 2:
        return True
    return False
```

ymzayek (commented):
I reviewed while you pushed some changes, so I'll have another look but it LGTM! Just a few things. Also can you add a whatsnew entry with your name and @sambjohnson?

NicolasGensollen (commented):
Thanks for working on this @sambjohnson !
Couple suggestions below. I think this is looking good ! 👍

alexisthual:
The latest commits should have implemented all suggested improvements @ymzayek and @NicolasGensollen!
Thanks for your feedback.

nilearn/plotting/surf_plotting.py
```diff
@@ -99,31 +92,100 @@
...
-def _set_view_plot_surf_plotly(hemi, view):
-    """Helper function for plot_surf with plotly engine.
+def _get_camera_view_from_string_view(hemi, view):
+    """Return plotly camera parameters from string view."""
+    if view == 'lateral':
+        return CAMERAS[hemi]
+    elif view == 'medial':
+        return CAMERAS[
+            VALID_HEMISPHERES[0]
+            if hemi == VALID_HEMISPHERES[1]
```
bthirion, line 126:
```suggestion
            if hemi == VALID_HEMISPHERES[0]
``` (?)

**Comment by sambjohnson** [comment]
@alexisthual's work to finish this feature looks great to me - very nice to have the plotly engine work too. Thanks very much!

File: nilearn/plotting/surf_plotting.py (hunk 13 above)
Comment (alexisthual) [reply nilearn/plotting/surf_plotting.py:126]: I think the current code is correct: when `view == "medial"`, you want to use the opposite value of the given hemisphere and not the given hemisphere as your code suggests (otherwise you could just call `CAMERAS[hemi]` as is done when `view == "lateral"`).

**Comment by alexisthual** [comment]
It's not clear to me why tests are not passing, I'll relaunch a full build.

**Comment by bthirion** [comment]
@ymzayek any hint about the doc build failure ? I don't find meaningful error messages.

**Review by ymzayek — COMMENTED** [review]
https://github.com/nilearn/nilearn/pull/3259#issuecomment-1492919634 the fix for this was merged yesterday so you'll need to rebase on main. 

And then there are 3 warnings in this doc build that may also be erroring out: https://github.com/nilearn/nilearn/actions/runs/4589876174/jobs/8105061324?pr=3259#step:33:15466

File: nilearn/plotting/surf_plotting.py [hunk 14]
```diff
@@ -1090,26 +1156,64 @@ def _check_hemispheres(hemispheres):
...
 def _check_views(views) -> list:
     """Checks whether the views passed to in plot_img_on_surf are
     correct.
 
+    Parameters
+    ----------
     views : list
-        Any combination of "anterior", "posterior", "medial", "lateral",
-        "dorsal", "ventral".
+        Any combination of string in {"anterior", "posterior", "medial",
```
NicolasGensollen, line 1198:
```suggestion
        Any combination of strings in {"anterior", "posterior", "medial",
```

nilearn/plotting/tests/test_surf_plotting.py
```diff
@@ -71,16 +196,38 @@
...
+    camera_view_elev_azim = _get_camera_view_from_elevation_and_azimut(
+        (elev, azim)
+    )
+    # Check camera view
+    for k in ["center", "eye", "up"]:
+        assert np.allclose(
+            list(camera_view[k].values()),
+            list(expected_camera_view[k].values())
+        )
+    # Check camera view obtained from string view
```
NicolasGensollen, line 219: You can write a single for loop with three assertions here

nilearn/plotting/tests/test_surf_plotting.py
```diff
@@ -181,19 +328,57 @@ def test_instantiation_error_plotly_surface_figure(input_obj):
...
+@pytest.mark.parametrize("hemi,view", [("left", "foo"), ("right", "bar")])
+def test_get_view_plot_surf_view_errors(hemi, view):
+    from nilearn.plotting.surf_plotting import (_get_view_plot_surf_matplotlib,
+                                                _get_view_plot_surf_plotly)
+    with pytest.raises(ValueError,
+                       match="Invalid view definition"):
+        _get_view_plot_surf_matplotlib(hemi, view)
+    with pytest.raises(ValueError,
+                       match="Invalid view definition"):
+        _get_view_plot_surf_plotly(hemi, view)
```
NicolasGensollen, line 381: Seems like you can parametrize on the function also (you'll need to import them at the top then...).

NicolasGensollen (commented):
Thanks @sambjohnson @alexisthual !
This looks good to me !
I just made a few small comments and suggestions. Feel free to take or leave !

doc/changes/latest.rst
```diff
@@ -12,6 +12,7 @@ NEW
 
 - ``fsaverage`` meshes accessed through :func:`~datasets.fetch_surf_fsaverage` now come with flat maps for all resolutions (:gh:`3444` by `Alexis Thual`_).
 
+- surface plotting functions allow setting custom view angles and are no longer limited to predefined views (:gh:`3259` by `Sam Buck Johnson`_ and `Alexis Thual`_).
```
ymzayek, line 15: You'll need to also add @sambjohnson to `doc/changes/names.rst` because of being a first-time contributor or else this link will error out

nilearn/plotting/surf_plotting.py
```diff
@@ -99,31 +92,100 @@
...
-                if hemi == VALID_HEMISPHERES[1]
-                else VALID_HEMISPHERES[1])
-    if view not in CAMERAS:
-        raise ValueError(f"view must be one of {VALID_VIEWS}")
-    return view
+    _check_views([view])
+    _check_hemispheres([hemi])
+    if isinstance(view, str):
+        return _get_camera_view_from_string_view(hemi, view)
+    return _get_camera_view_from_elevation_and_azimut(view)
```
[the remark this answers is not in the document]
alexisthual: Yes it is :) (see `test_surf_plotting.py` line 210)

alexisthual:
The latest commits should implement all suggested improvements @ymzayek and @NicolasGensollen.
Thank you for your feedback!

nilearn/plotting/surf_plotting.py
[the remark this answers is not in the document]
NicolasGensollen: Thanks @alexisthual ! Got confused by the codecov message... 🙄

NicolasGensollen (approved):
LGTM, thanks @sambjohnson and @alexisthual !

ymzayek (approved):
LGTM thanks @alexisthual !

PR #3262, opened 2022-05-19
[ENH] Update docstrings and seeding in `_utils.data_gen`

### LINKED ISSUE #2631: seeding in `_utils.data_gen`

ymzayek:
Closes #2631.
Replaces previous PR (closes #2741).


nilearn/_utils/data_gen.py
```diff
@@ -21,33 +21,84 @@
...
+
+    Parameters
+    ----------
+    n_scans : int, optional
+        Number of scans.
+        Default=1.
+
+    res : int, optional
+        Desired resolution, in mm, of output images.
+        Default=30.
```
bthirion, line 34: Any reason for this default ? This looks really unusual.

nilearn/_utils/data_gen.py
```diff
@@ -21,33 +21,84 @@
...
+        Default=0.
+
+    mask_dilation : int, optional
+        The number of times the binary dilation is repeated on the mask.
+        Default=2.
+
+    Returns
+    -------
+    inverse_img : nibabel.Nifti1Image
+        #TODO review
```
bthirion, line 47: to be cleaned

nilearn/_utils/data_gen.py
```diff
@@ -21,33 +21,84 @@
...
+        4D, Image transformed back to brain space.
+
+    mask_img : nibabel.Nifti1Image
+        #TODO review
+        3D, Mask resampled to mni space.
+
+    """
```
bthirion, line 54: In general, the docstrings need a big refresh to follow our current standards

nilearn/_utils/data_gen.py
```diff
@@ -21,33 +21,84 @@
...
-                image.get_data(mask_img), iterations=mask_dilation))
-    return masker.inverse_transform(data), mask_img
+            mask_img,
+            binary_dilation(image.get_data(mask_img),
+                            iterations=mask_dilation))
+    inverse_img = masker.inverse_transform(data)
+    return inverse_img, mask_img
+
 
+def generate_timeseries(n_instants, n_features, random_state=0):
```
bthirion, line 72: maybe `n_timepoints` rather than `n_instants`

bthirion (commented):
Thx ! Could you take this opportunity to refresh the docstrings of the functions ?

nilearn/_utils/data_gen.py
```diff
@@ -63,30 +114,33 @@ def generate_regions_ts(n_features, n_regions,
...
 
     """
-    if rand_gen is None:
-        rand_gen = np.random.RandomState(0)
+    rand_gen = check_random_state(random_state)
     if window is None:
         window = "boxcar"
 
-    assert(n_features > n_regions)
+    assert (n_features > n_regions)
```
NicolasGensollen, line 136:
```suggestion
    assert n_features > n_regions
```

nilearn/_utils/data_gen.py
```diff
@@ -193,25 +263,52 @@ def generate_labeled_regions(shape, n_regions, rand_gen=None, labels=None,
...
 
     See generate_labeled_regions for details.
+
+    Parameters
+    ----------
+    shape : tuple
+        Shape of returned array.
+
+    n_regions : int
+        Number of regions to generate. By default (if "labels" is None),
```
NicolasGensollen, line 281: Probably copy pasted from `generate_labeled_regions` since there is no `labels` parameter in this function.

nilearn/_utils/data_gen.py
```diff
@@ -193,25 +263,52 @@ def generate_labeled_regions(shape, n_regions, rand_gen=None, labels=None,
...
+                                   random_state=0,
                                    affine=np.eye(4)):
     """Similar to generate_labeled_regions, but suitable for a large number of
     regions.
 
     See generate_labeled_regions for details.
+
+    Parameters
+    ----------
+    shape : tuple
```
NicolasGensollen, line 277:
```suggestion
    shape : tuple of int
```

nilearn/_utils/data_gen.py
```diff
@@ -333,47 +426,153 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
+def generate_fake_fmri_data_and_design(shapes,
+                                       rk=3,
+                                       affine=np.eye(4),
+                                       random_state=0):
+    """
+    Generate random fmri data and design matrices of given shapes.
+
+    Parameters
+    ----------
+    shapes : list of 4D tuples
```
NicolasGensollen, line 438:
```suggestion
    shapes : list of 4D tuples of int
```

nilearn/_utils/data_gen.py
```diff
@@ -333,47 +426,153 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
+def write_fake_fmri_data_and_design(shapes,
+                                    rk=3,
+                                    affine=np.eye(4),
+                                    random_state=0):
+    """
+    Generate random fmri data and design matrices and write them to disk.
+
+    Parameters
+    ----------
+    shapes : List
```
NicolasGensollen, line 491:
```suggestion
    shapes : List of Tuples of int
```

nilearn/_utils/data_gen.py
```diff
@@ -333,47 +426,153 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
+        Random number generator, or seed.
+        Default=0.
+
+    Returns
+    -------
+    mask_file : str
+        path to mask file.
+
+    fmri_files : List
+        list of str.
```
NicolasGensollen, line 512: This would be more appropriate as the type, and the description could be something like "list of paths to the generated fmri files" or something like that. WDYT?

nilearn/_utils/data_gen.py
```diff
@@ -333,47 +426,153 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
+
+    design_files : List
+        list of str.
```
NicolasGensollen, line 515: Same kind of comment here.

nilearn/_utils/data_gen.py
```diff
@@ -333,47 +426,153 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
+                        random_state=0):
+    """
+    Generate a random image of given shape and write it to disk.
+
+    Parameters
+    ----------
+    file_path : str
+        output file path.
+
+    shape : tuple
```
NicolasGensollen, line 549:
```suggestion
    shape : tuple of int
```

NicolasGensollen (commented):
Thanks for working on this @ymzayek !
Below are a few suggestions, mostly on the docstrings.

nilearn/_utils/data_gen.py
ymzayek: Indeed that is strange, maybe it's a typo and supposed to be 3

nilearn/_utils/data_gen.py
ymzayek: Yes I agree that's better, will change it

nilearn/_utils/data_gen.py
ymzayek: idem

nilearn/_utils/data_gen.py
ymzayek: Ok will change

nilearn/_utils/data_gen.py
bthirion: OK. I think we should simply revert that. Sorry.

bthirion (approved):
LGTM, thx.

emdupre:
Thank you, @ymzayek ! It looks like there are a few merge conflicts on the files, currently -- could you update these ? I'll try to look through once that's done :)

nilearn/_utils/data_gen.py
```diff
@@ -102,78 +156,98 @@ def generate_regions_ts(n_features, n_regions,
...
+    border : :obj:`int`, optional
         Number of background voxels on each side of the 3D volumes.
         Default=1.
 
     Returns
     -------
-    maps : nibabel.Nifti1Image
-        4D array, containing maps.
+    maps : Niimg-like object
+        4D array containing maps.
```
tsalo, line 197: The description implies a numpy array to me. What about `4D image object containing the maps.`?

nilearn/_utils/data_gen.py
```diff
@@ -222,47 +326,52 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
         shape: shape + (length,)
 
-    mask : nibabel.Nifti1Image
+    Niimg-like object
         Mask giving non-zero voxels.
 
-    target : numpy.ndarray
+    target : :obj:`numpy.ndarray`
         Classification or regression target. Shape of number of
         time points (length). Returned only if n_blocks is not None.
```
tsalo, line 376:
```suggestion
        Classification or regression target.
        A 1D array with one element for each time point.
        Returned only if ``n_blocks`` is not None.
```

nilearn/_utils/data_gen.py
```diff
@@ -333,136 +438,257 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
+    ----------
+    shapes : :obj:`list` of 4D :obj:`tuple`s of :obj:`int`
+        Shapes of the fmri data to be generated.
+
+    rk : :obj:`int`, optional
+        Number of columns in the design matrix to be generated.
+        Default=3.
+
+    affine : :obj:`numpy.ndarray`, optional
+        Affine of returned images.
```
tsalo, line 457:
```suggestion
        Affine of returned images. Must be a 4x4 array.
```

nilearn/_utils/data_gen.py
```diff
@@ -530,67 +756,97 @@ def basic_paradigm(condition_names_have_spaces=False):
...
-                             n_runs=[1, 3], with_derivatives=True,
+                             n_runs=[1, 3],
+                             with_derivatives=True,
                              with_confounds=True,
                              confounds_tag="desc-confounds_timeseries",
-                             no_session=False):
+                             no_session=False,
+                             random_state=0):
     """Creates a fake :term:`bids<BIDS>` dataset directory with dummy files.
     Returns fake dataset directory name.
```
tsalo, line 796:
```suggestion
    """Creates a fake :term:`bids<BIDS>` dataset directory with dummy files.

    Returns fake dataset directory name.
```

I don't think we use it at the moment, but `flake8-docstrings` requires the summary sentence to use one line, and then the longer description should be after an empty line. It might be good to use that organization generally.

nilearn/_utils/data_gen.py
```diff
@@ -530,67 +756,97 @@ def basic_paradigm(condition_names_have_spaces=False):
...
-    base_dir : string (Absolute path), optional
-        Absolute directory path in which to create the fake :term:`BIDS` dataset dir.
+    base_dir : :obj:`str` (Absolute path), optional
+        Absolute directory path in which to create the fake :term:`BIDS`
+        dataset dir.
         Default: Current directory.
 
-    n_sub : int, optional
+    n_sub : :obj:`int`, optional
         Number of subject to be simulated in the dataset.
```
tsalo, line 806:
```suggestion
        Number of subjects to be simulated in the dataset.
```

nilearn/_utils/data_gen.py
```diff
@@ -530,67 +756,97 @@ def basic_paradigm(condition_names_have_spaces=False):
...
-    n_ses : int, optional
+    n_ses : :obj:`int`, optional
         Number of sessions to be simulated in the dataset.
         Ignored if no_session=True.
         Default=2.
 
-    tasks : List of str, optional
-        List of tasks. Default=["localizer", "main"].
+    tasks : :obj:`list` of :obj:`str`, optional
+        List of tasks.
```
tsalo, line 815:
```suggestion
        List of tasks to be simulated in the dataset.
```

nilearn/_utils/data_gen.py
```diff
@@ -530,67 +756,97 @@ def basic_paradigm(condition_names_have_spaces=False):
...
+        Default=["localizer", "main"].
 
-    n_runs : List of int, optional
+    n_runs : :obj:`list` of :obj:`int`, optional
         Default=[1, 3].
```
tsalo, line 819:
```suggestion
        Number of runs to create, where each element indicates the
        number of runs for the corresponding task.
        The length of this list must match the number of items in ``tasks``.
        Default=[1, 3].
```

nilearn/_utils/data_gen.py
tsalo: Could you also mention somewhere in the docstring that all runs will be created with 100 volumes?

tsalo (commented):
I have a few minor wording notes. Feel free to take or leave my suggestions.

nilearn/_utils/data_gen.py
```diff
@@ -333,136 +438,257 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
-def generate_fake_fmri_data_and_design(shapes, rk=3, affine=np.eye(4)):
+def generate_fake_fmri_data_and_design(shapes,
+                                       rk=3,
+                                       affine=np.eye(4),
+                                       random_state=0):
+    """Generate random fmri data and design matrices of given shapes.
+
+    Parameters
+    ----------
+    shapes : :obj:`list` of 4D :obj:`tuple`s of :obj:`int`
```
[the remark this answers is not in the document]
ymzayek: Yes true, will change

nilearn/_utils/data_gen.py
```diff
@@ -333,136 +438,257 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
             Nifti1Image(mask, affine), target)
 
 
-def generate_fake_fmri_data_and_design(shapes, rk=3, affine=np.eye(4)):
+def generate_fake_fmri_data_and_design(shapes,
+                                       rk=3,
+                                       affine=np.eye(4),
+                                       random_state=0):
+    """Generate random fmri data and design matrices of given shapes.
```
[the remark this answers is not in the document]
ymzayek: Good point. Is it enough to mention "time series"? Or do you think I should also specify in the Returns?

nilearn/_utils/data_gen.py
```diff
@@ -333,136 +438,257 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
+                        random_state=0):
+    """Generate a random image of given shape and write it to disk.
+
+    Parameters
+    ----------
+    file_path : :obj:`str`
+        Output file path.
+
+    shape : :obj:`tuple` of :obj:`int`
+        Shape of output array.
```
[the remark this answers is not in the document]
ymzayek: No indeed it can handle 4+-dimensional data and is used in `create_fake_bids_dataset` for 4D. Should I add it should be at least 3D?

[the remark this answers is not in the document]
tsalo: Absolutely. It would be great if the description had that info.

nilearn/_utils/data_gen.py
ymzayek: Yes, done

nilearn/_utils/data_gen.py
```diff
@@ -21,72 +21,126 @@
...
+        Desired resolution, in mm, of output images.
+        Default=30.
+
+    random_state : :obj:`int` or :obj:`numpy.random.RandomState` instance, \
+                   optional
+        Random number generator, or seed.
+        Default=0.
+
+    mask_dilation : :obj:`int`, optional
+        The number of times the binary dilation is repeated on the mask.
```
NicolasGensollen, line 42:
```suggestion
        The number of times the binary :term:`dilation<Dilation>` is repeated on the mask.
```

nilearn/_utils/data_gen.py
```diff
@@ -102,78 +156,98 @@ def generate_regions_ts(n_features, n_regions,
...
+        Default=np.eye(4).
 
-    affine : numpy.ndarray, optional
-        Affine of returned image. Default=np.eye(4).
-
-    dtype : type, optional
-        Data type of image. Default=np.int32.
+    dtype : :obj:`type`, optional
+        Data type of image.
+        Default=int.
```
NicolasGensollen, line 246:
```suggestion
        Default='int32'.
```

nilearn/_utils/data_gen.py
```diff
@@ -193,78 +269,112 @@ def generate_labeled_regions(shape, n_regions, rand_gen=None, labels=None,
     return Nifti1Image(data, affine)
 
 
-def generate_labeled_regions_large(shape, n_regions, rand_gen=None,
-                                   affine=np.eye(4), dtype="int32"):
+def generate_labeled_regions_large(shape,
+                                   n_regions,
+                                   random_state=0,
+                                   affine=np.eye(4)):
```
NicolasGensollen, line 275: I might have missed something, but why was `dtype` removed?

nilearn/_utils/data_gen.py
```diff
@@ -333,136 +439,257 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
             Nifti1Image(mask, affine), target)
 
 
-def generate_fake_fmri_data_and_design(shapes, rk=3, affine=np.eye(4)):
+def generate_fake_fmri_data_and_design(shapes,
+                                       rk=3,
+                                       affine=np.eye(4),
+                                       random_state=0):
+    """Generate random fMRI time series and design matrices of given shapes.
```
NicolasGensollen, line 446:
```suggestion
    """Generate random :term:`fMRI` time series and design matrices of given shapes.
```

nilearn/_utils/data_gen.py
```diff
@@ -333,136 +439,257 @@ def generate_fake_fmri(shape=(10, 11, 12), length=17, kind="noise",
...
                        affine)
     return mask, fmri_data, design_matrices
 
 
-def write_fake_fmri_data_and_design(shapes, rk=3, affine=np.eye(4)):
+def write_fake_fmri_data_and_design(shapes,
+                                    rk=3,
+                                    affine=np.eye(4),
+                                    random_state=0):
+    """Generate random fmri data and design matrices and write them to disk.
```
NicolasGensollen, line 499:
```suggestion
    """Generate random :term:`fMRI` data and design matrices and write them to disk.
```

NicolasGensollen (approved):
LGTM, thanks for improving these functions @ymzayek !
I made a few small suggestions below, and I think you could add a what's new entry for this.

nilearn/_utils/data_gen.py
ymzayek: That should not have been removed. I will fix it.
I guess it's an issue of resolving a conflict from the original pull request.

NicolasGensollen (commented):
Small point below. Maybe trigger a full build afterwards to make sure we didn't miss anything?

ymzayek:
@NicolasGensollen seems good with the full build. WDYT?

NicolasGensollen:
@ymzayek Yes, everything looks good and there are currently two approvals so I believe this is good for merging.
@emdupre @tsalo do you want to take another look before this gets merged (since you are in the list of reviewers for this PR) ?

tsalo (approved):
Sorry for the delay. It looks good to me too.

ymzayek:
Great thanks for all the reviews! Merging

PR #3263, opened 2022-05-19
[DOC] Fix doc building

NicolasGensollen:
Fix doc build on main.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -8,7 +8,7 @@
 NEW
 ---
 
-- :func:`~interfaces.bids.glm.save_glm_to_bids` has been added, which writes model outputs to disk according to BIDS convention (:gh:`2715` by `Taylor Salo`_).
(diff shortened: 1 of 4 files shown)


NicolasGensollen:
@tsalo This is because the documentation doesn't build in strict mode in the PRs, but does on main.

bthirion (approved):
LGTM. At least the doc builds...

tsalo:
>@tsalo This is because the documentation doesn't build in strict mode in the PRs, but does on main.

Thank you for the clarification! Should we switch to using strict mode on PRs then? It might reduce mistakes in PRs (at least in mine).

bthirion:
The problem is that we're going to explode our CircleCI budget. I think that the problem is not too frequent.

tsalo:
Oh, good point! I forgot about that issue.

NicolasGensollen:
Thanks! Merging...

PR #3267, opened 2022-05-25
[ENH] Smoothing of surface data

### LINKED ISSUE #2949: Smoothing function for surface data
See this [Neurostars post](https://neurostars.org/t/smoothing-surface-data/20159) and #2747 for a related issue.

The idea would be to have a smoothing function for surface data similar to `nilearn.image.smooth_img` for volumes:

https://github.com/nilearn/nilearn/blob/cc99fa76df137e9ac6010c311e15fa9ac12ce66d/nilearn/image/image.py#L244-L248

After discussing this, we agreed that this could be a useful addition, and that it would fit within the scope of Nilearn (most likely within the surface module).

One point that still needs to be discussed is how this function should be implemented as there are several possible ways to do that. 

For example:

For each point:
- Average across mesh direct neighbors
- Average across nodes within a ball of a given radius
- Average across mesh neighborhood (all nodes within some graph distance defined through edge weights)
- Other ideas?

How should we average?

Can we preserve the same API as `smooth_img` (with `fwhm` as input)?

_Note: Some of these implementations are obviously not desirable as they can introduce biases in the smoothing._ 
_Note: The proposed implementation should be fast (avoid for loops...) and not overly complex for maintenance reasons._

jyeatman:
- closes #2949.

Changes proposed in this pull request: This PR implements smoothing of surface data

- add function to smooth surfaces

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -5,11 +5,24 @@
 0.14.0.dev
 ==========
 
+HIGHLIGHTS
+----------
+
+.. warning::
+
+ | **Minimum supported versions of the following packages have been bumped up:**
+ | - numpy -- 1.26.0
+ | - pandas -- 2.3.0
+ | - scikit-learn -- 1.5.0
+ | - scipy -- 1.11.0
+
 NEW
 ---
 
 Fixes
 -----
+
+- :bdg-secondary:`Maint` Allow local installation with ```uv sync`` (:gh:`6024` by `Mathieu Dugré`_)
 
 Enhancements
 ------------
--- a/nilearn/image/tests/test_image.py
+++ b/nilearn/image/tests/test_image.py
@@ -601,6 +601,7 @@ def test_crop_image_empty_image(affine_eye, pad):
 
 
 @pytest.mark.thread_unsafe
+@pytest.mark.slow
 @pytest.mark.parametrize("images_to_mean", _images_to_mean())
 def test_mean_img(images_to_mean, tmp_path):
     affine = np.diag((4, 3, 2, 1))
@@ -2247,6 +2248,7 @@ def test_check_niimg_user_expand(img_in_home_folder, filename):
 
 
 @pytest.mark.thread_unsafe
+@pytest.mark.single_process
 @pytest.mark.parametrize(
     "filename",
     [
--- a/.github/workflows/testing_install.yml
+++ b/.github/workflows/testing_install.yml
@@ -61,6 +61,32 @@ jobs:
                 pip install dist/nilearn-*-py3-none-any.whl
                 python -c "import nilearn"
 
+    test_uv_sync:
+        if: github.repository == 'nilearn/nilearn'
+        runs-on: ${{ matrix.os }}
+        strategy:
+            fail-fast: false
+            matrix:
+                os: [ubuntu-latest, macos-latest, windows-latest]
+        name: check install on ${{ matrix.os }} with latest package versions
+        defaults:
+            run:
+                shell: bash
+        steps:
+        -   name: Install uv
+            uses: astral-sh/setup-uv@v7
+        -   name: Checkout nilearn
+            uses: actions/checkout@v6
+        -   name: Setup python
+            uses: actions/setup-python@v6
+            with:
+                python-version: '3.10'
+        -   name: Test uv sync
+            shell: bash {0}
+            run: |
+                set -e
+                uv sync
+
     check_wheel_size:
         if: github.repository == 'nilearn/nilearn'
         runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,7 @@ package-lock.json
 
 # pydeps output
 nilearn.svg
+
+# uv
+uv.lock
+.python-version
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,7 +82,7 @@ repos:
 
     # Lint and format Python code
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.0
+    rev: v0.15.1
     hooks:
     -   id: ruff-check
         # args: [--statistics]
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -17,7 +17,7 @@ import sys
 import warnings
 from pathlib import Path
 
-from nilearn._version import __version__  # noqa : I001, RUF100
+from nilearn._version import __version__
 from sphinx.domains import changeset
 from sphinx.locale import _
 
--- a/examples/02_decoding/plot_haxby_understand_decoder.py
+++ b/examples/02_decoding/plot_haxby_understand_decoder.py
@@ -232,7 +232,6 @@ classifier = LogisticRegressionCV(
     solver="liblinear",
     Cs=np.geomspace(1e-3, 1e4, 8),
     refit=True,
-    verbose=1,
 )
 
 # %%
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -1205,7 +1205,7 @@ standardize : any of: 'zscore_sample', 'zscore', 'psc', True, False or None; \
 # TODO (nilearn >= 0.15.0) update to ..versionchanged
 deprecation_notice_false_to_none = """
 
-    .. nilearn_deprecated:: 0.15.0dev
+    .. nilearn_deprecated:: 0.13.0
 
         The default will be changed to ``None``
         in version 0.15.0.
@@ -1215,7 +1215,7 @@ deprecation_notice_false_to_none = """
 # TODO (nilearn >= 0.15.0) update to ..versionchanged
 deprecation_notice_true_to_zscore_sample = """
 
-    .. nilearn_deprecated:: 0.15.0dev
+    .. nilearn_deprecated:: 0.13.0
 
         The default will be changed to ``'zscore_sample'``
         in version 0.15.0.
@@ -1225,7 +1225,7 @@ deprecation_notice_true_to_zscore_sample = """
 docdict["standardize_false"] = (
     standardize.format("False") + deprecation_notice_false_to_none
 )
-# TODO (nilearn >= 0.14.0 and 0.15.0)
+# TODO (nilearn >= 0.15.0)
 # adapt the deprecation notices
 docdict["standardize_true"] = (
     standardize.format("True") + deprecation_notice_true_to_zscore_sample
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -1556,8 +1556,10 @@ def check_img_estimator_fit_idempotent(estimator_orig) -> None:
         # TODO
         # some estimator can return some pretty different results
         # investigate why
-        if isinstance(estimator, (Decoder, SearchLight)):
+        if isinstance(estimator, (Decoder)):
             tol = 1e-5
+        elif isinstance(estimator, (SearchLight)):
+            tol = 1e-4
         elif isinstance(estimator, FREMClassifier):
             tol = 0.1
 
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -1,7 +1,6 @@
 """Neuroimaging file input and output."""
 
 import collections.abc
-import gc
 from copy import deepcopy
 from pathlib import Path
 from warnings import warn
@@ -9,7 +8,7 @@ from warnings import warn
 import numpy as np
 from nibabel import Nifti1Image, is_proxy, load, spatialimages
 
-from nilearn._utils.helpers import is_gil_enabled, stringify_path
+from nilearn._utils.helpers import stringify_path
 from nilearn._utils.logger import find_stack_level
 
 
@@ -60,11 +59,6 @@ def safe_get_data(
     """
     if not img.in_memory or copy_data:
         img = deepcopy(img)
-
-    if is_gil_enabled():
-        # typically the line below can double memory usage
-        # that's why we invoke a forced call to the garbage collector
-        gc.collect()
 
     data = _get_data(img)
 
--- a/nilearn/plotting/tests/test_baseline_comparisons.py
+++ b/nilearn/plotting/tests/test_baseline_comparisons.py
@@ -181,9 +181,6 @@ def test_plot_carpet_default_params(img_4d_mni, img_3d_ones_mni):
 @pytest.mark.mpl_image_compare
 def test_plot_prob_atlas_default_params(img_3d_mni, img_4d_mni):
     """Smoke-test for plot_prob_atlas with default arguments."""
-    # TODO (nilearn >= 0.13.0)
-    # using only 2 regions to speed up the test
-    # maps = generate_maps(shape_3d_default, n_regions=2, affine=affine_mni)
     return plot_prob_atlas(img_4d_mni, bg_img=img_3d_mni)
 
 
--- a/nilearn/tests/test_signal.py
+++ b/nilearn/tests/test_signal.py
@@ -789,30 +789,24 @@ def test_clean_confounds():
     assert array_equal(noises, noises1)
 
     # With signal: output must be orthogonal to confounds
-    # TODO (nilearn >= 0.14) remove catch FutureWarning
-    with pytest.warns(FutureWarning):
-        cleaned_signals = clean(
-            signals + noises,
-            confounds=confounds,
-            detrend=False,
-            standardize=True,
-        )
+    cleaned_signals = clean(
+        signals + noises,
+        confounds=confounds,
+        detrend=False,
+        standardize="zscore_sample",
+    )
 
     assert abs(np.dot(confounds.T, cleaned_signals)).max() < 1000.0 * EPS
 
     # Same output when a constant confound is added
     confounds1 = np.hstack((np.ones((45, 1)), confounds))
-    # TODO (nilearn >= 0.15) remove catch_warnings
-    with pytest.warns(
-        FutureWarning,
-        match="boolean values for 'standardize' will be deprecated",
-    ):
-        cleaned_signals1 = clean(
-            signals + noises,
-            confounds=confounds1,
-            detrend=False,
-            standardize=True,
-        )
+
+    cleaned_signals1 = clean(
+        signals + noises,
+        confounds=confounds1,
+        detrend=False,
+        standardize="zscore_sample",
+    )
 
     assert_almost_equal(cleaned_signals1, cleaned_signals)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,11 +68,11 @@ doc = [
 min = [
     "joblib==1.2.0",
     "nibabel==5.2.0",
-    "numpy==1.22.4",
-    "pandas==2.2.0",
+    "numpy==1.26.0",
+    "pandas==2.3.0",
     "requests==2.30.0",
     "scikit-learn==1.5.0",
-    "scipy==1.9.0",
+    "scipy==1.11.0",
     "jinja2==3.1.2"
 ]
 min_plotting = ["nilearn[min]", "matplotlib==3.8.0", "plotly==5.0.0", "kaleido"]
@@ -416,6 +416,7 @@ max-complexity = 23
 "**/{doc}/*" = ["B018", "D103"]
 "**/{examples}/*" = ["B018", "D103", "D400", "E402"]
 "__init__.py" = ["D104", "E402"]
+"doc/conf.py" = ["I001"]  # ignore sorting because this leads to conflict when running pre-commit locally and in CI
 "nilearn/**/tests/test_*" = ["D100"]
 "nilearn/connectome/**/{tests}/*" = ["D103"]  # 35 errors to fix
 "nilearn/datasets/func.py" = ["PTH211"]  # 1 error


nilearn/surface/surface.py
```diff
@@ -1010,3 +1010,114 @@ def check_surface(surface):
...
+    -------
+    >>> from nilearn import datasets, surface, plotting
+    >>> fsaverage = datasets.fetch_surf_fsaverage('fsaverage')
+    >>> white_left = surface.load_surf_mesh(fsaverage.white_left)
+    >>> curv = surface.load_surf_data(fsaverage.curv_left)
+    >>> curv_smooth = surface.smooth_surface_data(surface = white_left, surf_data = curv, smooth_steps=50)
+    >>> plotting.plot_surf(white_left, surf_map = curv_smooth)
+    
+    """
+    if method == 'sparse':
```
bthirion, line 1071: Why wouldn't one use a sparse method ?

alexisthual: I didn't really understand the difference between the two methods (it feels to me that the `sparse` method would be faster since there are less `for` loops).

alexisthual (commented):
I think this would work although I haven't tested it!
IMHO, it would be very useful to have different kernels available. For instance, a gaussian kernel could be very nice I think, as users to then choose a std.

jyeatman:
I changed the title of this pull request. I think it's ready for review. It does need more extensive tests but I might do that later

nilearn/surface/surface.py
```diff
@@ -1014,3 +1014,235 @@ def check_surface(surface):
...
+    # adding the center weight along the diagonal.
+    colsums = matrix.sum(axis=1)
+    colsums = np.asarray(colsums).flatten()
+    matrix = matrix.multiply(surround_weight / colsums[:,None])
+
+    # Add in the diagonal.
+    matrix.setdiag(center_weight)
+    # Run the iteratioons of smooothing.
+    n = len(surf_data)
+    # data = np.hstack([np.reshape(surf_data, (n,-1)), w[:,None]])
```
NicolasGensollen, line 1211: Is the following commented section not useful anymore? In this case could you remove it?

NicolasGensollen (commented):
Thanks a lot for working on this @jyeatman ! 👍 
I just made a first quick pass, mostly commenting on formatting, documentation, and typos. I haven't tested the code yet.
Note that you would need to add `smooth_surface_data` (or whatever name it ends up having) in `doc/modules/reference.rst` such that the function appears in the API documentation.
I'll take a closer look soon!

nilearn/surface/surface.py
```diff
@@ -1014,3 +1014,235 @@ def check_surface(surface):
...
+    Returns
+    -------
+    neighbors : list
+        A list of all the vertices that are connected to each vertex
+    """
+    from scipy.sparse import find
+    matrix = compute_adjacency_matrix(surface)
+    return [find(row)[1] for row in matrix]
+        
+def smooth_surface_data(surface, surf_data,
```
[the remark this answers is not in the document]
jyeatman: I thought about that. But I think that this is where the analogy between surface and volume breaks down. smooth_surf and smooth_surface might also refer to smoothing a surface mesh (otherwise referred to as inflating the mesh). So I went with smooth_surface_data with the analogy to load_surface_data. I separately might write a smooth_surface_mesh to smooth the mesh

nilearn/surface/surface.py
```diff
@@ -1014,3 +1014,235 @@ def check_surface(surface):
     mesh, data = check_mesh_and_data(surface.mesh,
                                      surface.data)
     return Surface(mesh, data)
+
+def compute_adjacency_matrix(surface, values='ones', dtype=None):
```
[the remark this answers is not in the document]
jyeatman: One question about the logic between private and public functions (excuse my ignorance since Python is not my native language): I could see this function being useful for many other things. For example doing clustering on the surface and doing various cluster based thresholding and permutations. In that case, is it better to be public or private? I assume private is harder for users to find so, therefor, someone who would be inclined to write surface based clustering might not realize that most of the heavy lifting is already done here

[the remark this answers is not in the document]
NicolasGensollen: On the contrary to other programming languages, there is no "real" difference between pubic and private functions in Python. Both can be imported and used in the same way.
The key difference is that public functions are part of the project's API that we expose to the users. These functions and classes are properly documented and assumed to be stable and backward compatible. This means that, once an object becomes public it is difficult (and long because of deprecation cycles) to change it, since we are bound to a backward compatibility contract with the users of the library. On the contrary, we are free to change private objects "as we want", which is much easier in terms of maintenance.
I had the feeling that these two functions were helper functions for `smooth_surface_data`, but you are right that they could be directly useful for users, and therefore be public and part of the Nilearn API.
If we agree that this is the case, we should think about the best location for them as well as potential exemples to showcase them. Ping @alexisthual @bthirion @ymzayek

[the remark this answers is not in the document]
alexisthual: In our case, I think these functions should be kept private. They could indeed be useful to some users (I myself have a personal repo for this kind of util functions only haha) but then a whole bunch of similar util functions could also be integrated. I'm afraid this would make the API look fuzzy on the long run

jyeatman:
@ymzayek I just went through and addressed most of the comments. I'll make the changes to private functions next (last outstanding comments).

jyeatman:
Alright made all the suggested fixes

clane9:
I'd be happy to add a review of this if that would help it move forward. Lmk if that would be useful.

nilearn/surface/surface.py
```diff
@@ -1014,3 +1014,229 @@ def check_surface(surface):
...
+                        vertex_weights=None,
+                        return_vertex_weights=False,
+                        center_surround_knob=0,
+                        match='sum'):
+    """Smooth values along the surface.
+    
+    Parameters
+    ----------
+    surface : Surface-like
+        The surface on which the data `surf_data` are to be smoothed.
```
ymzayek, line 1117: Can we be more specific here? And is this the same input as the `surface` parameter under e.g. `load_surface` and `check_surface` ? If so, it might be useful to add a standard docstring entry in `_utils.docs.py`

ymzayek (commented):
Thanks @jyeatman! I noticed that there are a lot of conditions in the functions that are not yet covered by tests as shown by the [codecov/patch report](https://github.com/nilearn/nilearn/pull/3267/checks?check_run_id=7278570664) so I think the testing still needs to be developed. Let me know if you need more specific help on that.

ymzayek:
@clane9 that would be helpful indeed, thanks!

clane9 (commented):
I did a pass over the main additions in `surface.py`. I think it looks good! I left a few small comments. A few are personal preference and can be safely ignored if you like.

My one high-level comment is I think it'd be nice to add a configuration to reproduce [Freesurfer's surface smoothing](https://github.com/freesurfer/freesurfer/blob/fs-7.0/utils/mrisurf_mri.cpp#L5230). They do a similar iterative neighbor averaging. The details are:

- No inverse distance weighting. (Already implemented with `distance_weights=False`.)
- Center vertex and surrounding neighbors have equal weights, i.e. center weight is `1 / n_neighbors`. (Requires some modification of the center/surround weighting.)
- The number of smoothing iterations can be inferred given a target smoothing FWHM. (Requires copying [this function](https://github.com/freesurfer/freesurfer/blob/fs-7.0/utils/mrisutils.cpp#L1100), which uses the average vertex area).

I'd be happy to help with these changes if you like.

alexisthual:
This is a very interesting feature! @jyeatman do you have the bandwidth to get this merged? :blush:

ymzayek:
@alexisthual I think you can go ahead and take this over and we can include @jyeatman as a contributor in a whatsnew entry. I'd suggest first rebasing on main and make sure all the reviewers' comments are addressed. Thanks!

jyeatman:
Sorry for the slow response - I don't think I'll have the capacity to finish and merge this. But I would be thrilled if you wanted to @alexisthual 
I'm very excited to see growth in the surface modules for niLearn

Remi-Gau:

bthirion:
Somebody has to take over that one. Maybe I can.

Remi-Gau:
> Somebody has to take over that one. Maybe I can.

In any case I would only work on this once the SurfaceImage API is finalized.

nilearn/surface/surface.py
```diff
@@ -1975,3 +1976,243 @@ def new_img_like(ref_img, data):
...
+        The dtype that should be used for the returned sparse matrix.
+
+    Returns
+    -------
+    matrix : scipy.sparse.csr_matrix
+        A sparse matrix representing the edge relationships in `surface`.
+
+    """
+    # This is a bit of a hack to quickly find a unique set of all edges.
+    if values not in {"len", "invlen", "ones"}:
```
Remi-Gau, line 2011: from the current implementation in smooth_img "len" is never passed therefore I am not sure it makes sense to include it.

nilearn/surface/surface.py
```diff
@@ -1975,3 +1976,243 @@ def new_img_like(ref_img, data):
...
+        raise ValueError(f"unrecognized values argument: {values}")
```
Remi-Gau, line 2012: given we control what value is passed to this function I am not sure it makes sense to control the inputs here

nilearn/surface/surface.py
```diff
@@ -1975,3 +1976,243 @@ def new_img_like(ref_img, data):
...
+
+    # We can now make a sparse matrix.
+    ee = np.concatenate([edge_lens, edge_lens])
+    uv = np.concatenate([u, v])
+    vu = np.concatenate([v, u])
+
+    return csr_matrix((ee, (uv, vu)), shape=(n, n))
+
+
+def smooth_img(
```
Remi-Gau, line 2062: big picture comment: this function does too many things.

for the sake of inter software reproducibility I would prefer something that sticks closer to what freesurfer does - as suggested in another comment in this PR (or maybe AFNI if they do surface smoothing)

nilearn/surface/surface.py
```diff
@@ -1975,3 +1976,243 @@ def new_img_like(ref_img, data):
...
+        where the weight on each neighbor is the inverse
+        of the distances to it.
+
+    vertex_weights : SurfaceImage or None, default = None
+        A SurfaceImage whose data are vector of weights, one per vertex.
+        These weights are normalized and
+        applied to the smoothing
+        after the application of center-surround weights.
+
+    center_surround_knob : :obj:`float`, default = 0
```
Remi-Gau, line 2096: I would not allow this to be tweaked: I would only average a given vertex with its nearest neighbour

nilearn/surface/surface.py
```diff
@@ -1975,3 +1976,243 @@ def new_img_like(ref_img, data):
...
+        is `2**k` times the weight of the vertex itself (the center).
+        A value of 0 (the default) means that, in each smoothing iteration,
+        each vertex is updated with the average of its value
+        and the average value of its neighbors.
+        A value of `-inf` results in no smoothing because the entire
+        weight is on the center, so each vertex is updated with its own value.
+        A value of `inf` results in each vertex being updated
+        with the average of its neighbors without including its own value.
+
+    match : { 'sum', 'mean', 'var', 'dist', None}, default = "sum"
```
Remi-Gau, line 2111: similarly: is this type of normalization needed?
We do no such thing for volume smoothing: why do it here? Am I missing something?

nilearn/surface/surface.py
```diff
@@ -1975,3 +1976,243 @@ def new_img_like(ref_img, data):
...
+            tmp = (tmp - mu1) * (std0 / std1) + mu0
+
+        new_data[hemi] = tmp
+
+        w /= np.sum(w)
+        weights[hemi] = w
+
+    smoothed_imgs = new_img_like(imgs, new_data)
+
+    return smoothed_imgs, weights
```
Remi-Gau, line 2218: I have changed this to always return the weights but I agree that this may not be needed

nilearn/surface/surface.py
```diff
@@ -1975,3 +1976,243 @@ def new_img_like(ref_img, data):
...
+        Defaults to 1
+
+    distance_weights : :obj:`bool`, default = False
+        Whether to add distance-based weighting to the smoothing.
+        With such weights, the value calculated for each vertex
+        at each iteration is the weighted sum of neighboring vertices
+        where the weight on each neighbor is the inverse
+        of the distances to it.
+
+    vertex_weights : SurfaceImage or None, default = None
```
Remi-Gau, line 2090: unless user ask for it I would not implement a vertex weighting at first

Remi-Gau:
> My one high-level comment is I think it'd be nice to add a configuration to reproduce [Freesurfer's surface smoothing](https://github.com/freesurfer/freesurfer/blob/fs-7.0/utils/mrisurf_mri.cpp#L5230).

I would even go further than @clane9 I would say this should the default (and only implementation for now of smoothing)

Remi-Gau:
@jyeatman if you have the bandwidth to answer a few questions.
I may have missed prior discussions on the implementation for this but what was the reasoning to have a smoothing function with some man options?
is that something you have seen a lot of people request or was it something you needed for yourself?

Remi-Gau:
Possible page to point to
https://www.nmr.mgh.harvard.edu/~greve/surf-smooth/

Remi-Gau:
Also pinging @noahbenson and @arokem who from reading the associated issue were involved in the dev of this PR.

Just asking for clarification on the reasons of having a smoothing function that does so many things.

jyeatman:
I can’t exactly remember all the reasoning but we were trying to implement a variety of principled ways to smooth along the surface. There are probably simpler ways to do something that works fine. @Noah ***@***.***> was the mastermind behind most of this

Jason D. Yeatman, Ph.D.<https://profiles.stanford.edu/jason-yeatman>
Associate Professor
Stanford University, Graduate School of Education and Department of Psychology
Stanford School of Medicine, Division Developmental Behavioral Pediatrics
Brain Development & Education Lab<http://brainandeducation.com/>
________________________________

nilearn/surface/surface.py
```diff
@@ -1975,3 +1976,243 @@ def new_img_like(ref_img, data):
...
+    return csr_matrix((ee, (uv, vu)), shape=(n, n))
+
+
+def smooth_img(
+    imgs,
+    iterations=1,
+    distance_weights=False,
+    vertex_weights=None,
+    center_surround_knob=0,
+    match="sum",
```
Remi-Gau, line 2068: if we decide to keep the rescaling then maybe let's use another word than "match" that is now a [soft keyword for python >=  3.10](https://peps.python.org/pep-0622/)

Remi-Gau:
@noahbenson 

> one should expose the algorithm's knobs but give them smart default values

Makes sense. If we can find a way to have the default give very close result to those of freesurfer then we could have our cake and eat it too. 🍰 

I started modifying one of the nilearn example to show surface smoothing: if you have any suggestion on showing how users can take full advantage of the full functionality of the function, let me know.

nilearn/surface/surface.py
Remi-Gau: @noahbenson

Am still not 100% convinced this is needed though.

Remi-Gau:
> * The number of smoothing iterations can be inferred given a target smoothing FWHM.

@noahbenson

Would it make sense to you to replace n_iteration with FWHM and compute the number of iterations under the hood?

See this comment for more info: https://github.com/nilearn/nilearn/pull/3267#pullrequestreview-1036645716

CITATION.cff
Remi-Gau, line 1: @jyeatman @noahbenson 

I have added info about you in our contributors list: can you double check they are valid?

Remi-Gau:
@noahbenson @jyeatman 
added this PR to the agenda of the next core dev meeting to see how the rest of the team feels about this

Remi-Gau:
By the way this is what the examples with smoothing looks like for now

https://output.circle-artifacts.com/output/job/c5a388bd-1fc0-4d2a-9c25-8c9625262208/artifacts/0/dev/auto_examples/06_manipulating_images/plot_smooth_mean_image.html#sphx-glr-auto-examples-06-manipulating-images-plot-smooth-mean-image-py

CITATION.cff
noahbenson: The only issue I see is that my affiliation should probably say "**Seattle**, Washington, USA". (I can also push this change when I get around to reviewing the code in the next few days.)

noahbenson:
@Remi-Gau This all sounds great—I like the idea of having FreeSurfer's smoothing method be a default, and it looks pretty easy to implement based on the linked comment.
I anticipate diving into this on Friday or so and will probably push some comments and edits then. Just for reference, when is the next core dev meeting?

examples/06_manipulating_images/plot_smooth_mean_image.py
```diff
@@ -24,13 +28,77 @@
...
+curvature = load_fsaverage_data(data_type="curvature")
+
+surface_image = SurfaceImage.from_volume(
+    mesh=fsaverage_meshes["pial"],
+    volume_img=stat_img,
+)
+
+hemi = "right"
+
+for iteration in range(0, 25, 5):
```
bthirion, line 72: We cannot specify smoothing kernels in number of iterations. We should specify them in terms of fwhm.

examples/06_manipulating_images/plot_smooth_mean_image.py
```diff
@@ -24,13 +28,77 @@
...
+        threshold=1.0,
+        bg_map=curvature,
+        cmap="RdBu_r",
+    )
+
+show()
+
+for match in ["sum", "mean", "var", "dist", None]:
+    smoothed_surface_image, _ = smooth_surf_img(
+        surface_image, iterations=25, match=match
```
bthirion, line 91: It is unclear what `match` means here.

nilearn/surface/surface.py
```diff
@@ -2013,3 +2014,243 @@ def new_img_like(ref_img, data):
...
+        # We need to normalize the matrix columns, and we can do this now by
+        # normalizing everything but the diagonal to the surround weight, then
+        # adding the center weight along the diagonal.
+        colsums = matrix.sum(axis=1)
+        colsums = np.asarray(colsums).flatten()
+        matrix = matrix.multiply(surround_weight / colsums[:, None])
+        # Add in the diagonal.
+        matrix.setdiag(center_weight)
+
+        # Run the iteratioons of smooothing.
```
bthirion, line 2220:
```suggestion
        # Run the iterations of smoothing.
```

nilearn/surface/surface.py
bthirion: I think that we can define smoothing intrinsically and leave possible post-processing to the user ?

bthirion (commented):
Thx for taking care of this long waited feature !

CITATION.cff
```diff
@@ -548,6 +552,10 @@ authors:
     family-names: Krish
     website: https://github.com/NIkhilgKrish
     affiliation: University of Texas, Austin
+  - given-names: Noah C.
+    family-names: Benson
+    website: https://nben.net/
+    affiliation: University of Washington, Washington, USA
```
Remi-Gau, line 558:
```suggestion
    affiliation: University of Washington, Seattle, Washington, USA
```

CITATION.cff
Remi-Gau: should be fixed now

bthirion:
Here is the paper I was referring to: https://pages.stat.wisc.edu/~mchung/papers/BMI2004/diffusion_biomed04.pdf

Remi-Gau:
@noahbenson @jyeatman 
I removed the "match" part of this PR.
we can re add it in a follow up PR if we think it should really be in.

nilearn/surface/surface.py
```diff
@@ -1014,3 +1014,229 @@ def check_surface(surface):
...
+    if surround_weight == 0:
+        # There's nothing to do in this case.
+        return np.array(surf_data)
+    # Calculate the adjacency matrix either weighting by inverse distance or not weighting (ones)
+    values = 'invlen' if distance_weights else 'ones'
+    matrix = _compute_adjacency_matrix(surface, values=values)
+    
+    # If there are vertex weights, get them ready.
+    if vertex_weights:
+        w = np.array(vertex_weights)
```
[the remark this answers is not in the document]
Remi-Gau: this is still an issue to fix

nilearn/surface/surface.py
```diff
@@ -2085,3 +2086,212 @@ def new_img_like(ref_img, data):
...
+    surround_weight = 1 - center_weight
+    if surround_weight == 0:
+        # There's nothing to do in this case.
+        return imgs
+
+    # Calculate the adjacency matrix either weighting
+    # by inverse distance or not weighting (ones)
+    values = "invlen" if distance_weights else "ones"
+
+    _ = _sanitize_weights(imgs, vertex_weights=vertex_weights)
```
Remi-Gau, line 2236: I extracted the weight setting into a separate function and made the smoothing function not return the weights anymore to be closer to the smoothing function in volume in terms of API.
Also if users REALLY want to access the weights they can use the private function.

That being said this makes it even more clear that weights are not used in the current implementation.

nilearn/surface/surface.py
```diff
@@ -1592,6 +1593,34 @@ def __getitem__(self, index):
...
+            )
+
+            edge1 = v1 - v0
+            edge2 = v2 - v0
+
+            cross_prod = np.cross(edge1, edge2)
+
+            area = 0.5 * np.linalg.norm(cross_prod)
+
+            total_area += area
```
Remi-Gau, line 1620: this may be very slow
one possible work around for most common cases that involve freesurfer surfaces:
- pre-compute the area of freesurfer surfaces
- check if the mesh is one of the ones known to nilearn
- return the precomputed area if so

nilearn/surface/surface.py
```diff
@@ -2048,3 +2077,76 @@ def extract_data(img, index):
         .reshape(mesh.parts[hemi].n_vertices, last_dim)
         for hemi in data.parts
     }
+
+
+def compute_adjacency_matrix(mesh, values="ones", dtype=None):
```
Remi-Gau, line 2082: technically covered by tests but actually unused in practice

Remi-Gau:
I would say that this is almost ready to merge.

Some of the original code was kept in a function that is private.
But some part of that code is not actually used even if technically covered by tests.

Given that this PR has been a blocker of quite a few things I would suggest:
- extracting the part of the code that are not "used" into a separate PR
- merge this PR if we agree with the implementation

Remi-Gau:
@noahbenson @jyeatman @bthirion @man-shu

nilearn/image/image.py
```diff
@@ -311,6 +328,190 @@ def smooth_img(imgs, fwhm):
...
+        where the weight on each neighbor is the inverse
+        of the distances to it.
+
+    vertex_weights : SurfaceImage or None, default = None
+        A SurfaceImage whose data are vector of weights, one per vertex.
+        These weights are normalized and
+        applied to the smoothing
+        after the application of center-surround weights.
+
+    center_surround_knob : :obj:`float`, default = 0
```
bthirion, line 364: I don't like these distance_weights and vertex_weights things.
To me the combination should be computed using the cotangent formula, and not left to the user's feeling ?

nilearn/image/image.py
```diff
@@ -311,6 +328,190 @@ def smooth_img(imgs, fwhm):
...
+
+        # Compute average vertex area
+        avg_vertex_area = mesh._area / mesh.n_vertices
+
+        # Compute number of iterations using empirical formula
+        niters.append(
+            math.floor(
+                1.14 * (4 * math.pi * G_STD**2) / (7 * avg_vertex_area) + 0.5
+            )
+        )
```
bthirion, line 477: Is there any way to check this formula ? E.g. smoothing a dirac on a sphere, for which we can probably predict the expected outcome?

nilearn/surface/surface.py
bthirion: Have you timed it ? Maybe it's possible to vectorize the whole computation ?

nilearn/surface/surface.py
bthirion: to be removed then...

bthirion (commented):
Thx, this is moving in the right direction !

nilearn/image/tests/test_image.py
```diff
@@ -381,6 +385,68 @@ def test_smooth_img(affine_eye, tmp_path):
...
+        )
+
+    more_smoothed_imgs = _smooth_surface_img(surf_img_1d, iterations=[2, 2])
+    for part in surf_img_1d.data.parts:
+        assert not np.array_equal(
+            more_smoothed_imgs.data.parts[part], smoothed_imgs.data.parts[part]
+        )
+
+
+@pytest.mark.xfail(reason="TODO - does not comply with doc string.")
```
Remi-Gau, line 432: set as xfail for now
this needs fixing

nilearn/image/image.py
noahbenson: In some cases the data itself has weights because there is a confidence attached to it. For example, in visual cortex, we often have population receptive field (pRF) models of the BOLD signal whose parameters tell us what part of the visual field a voxel/vertex is responding to. The models also have an r2 or coefficient of determination (CoD) that gives us some information about how confident we should be in the model parameters. It's not uncommon to weight smoothing of pRF parameters in the visual cortex by the CoD, and in such a case you need a way to explicitly tell the function that you have confidence measures (weights) for each vertex.

It makes more sense to me to replace the `distance_weights` parameter with something else. If the parameter is just removed, then there's no difference between smoothing on the native mesh and smoothing on the inflated spherical mesh, which I think is not desirable. But I'm not sure that I understand what specifically you're proposing to do with the cotangent formula here?

nilearn/image/tests/test_image.py
noahbenson: This fails because in image.py, the line
```python
center_weight = 1 / (1 + np.exp2(-center_surround_knob))
```
should be
```
center_weight = 1 / (1 + np.exp2(center_surround_knob))
```
(I put a comment there too.)

noahbenson (commented):
Apologies for going MIA after promising to work on this earlier in the year (it's been a long year already; I'm sure it's not just me).

Regardless, thanks for continuing to work on it!

nilearn/image/image.py
```diff
@@ -311,6 +328,147 @@ def smooth_img(imgs, fwhm):
...
+
+    """
+    # First, calculate the center and surround weights for the
+    # center-surround knob.
+    center_weight = 1 / (1 + np.exp2(-center_surround_knob))
+    surround_weight = 1 - center_weight
+    if surround_weight == 0:
+        # There's nothing to do in this case.
+        new_img_like(imgs, imgs.data)
+        return imgs
```
Remi-Gau, line 391:
```suggestion
        return new_img_like(imgs, imgs.data)
```

nilearn/image/image.py
```diff
@@ -311,6 +328,146 @@ def smooth_img(imgs, fwhm):
...
+    Examples
+    --------
+    >>> from nilearn import datasets, surface
+    >>> curv = datasets.load_fsaverage_data(data_type="curvature")
+    >>> curv_smooth = surface.smooth_img(curv, iterations=50)
+
+    """
+    # First, calculate the center and surround weights for the
+    # center-surround knob.
+    center_weight = 1 / (1 + np.exp2(-center_surround_knob))
```
Remi-Gau, line 386:
```suggestion
    center_weight = 1 / (1 + np.exp2(center_surround_knob))
```

doc/changes/latest.rst
```diff
@@ -60,8 +62,6 @@ Changes
 
 - :bdg-danger:`Deprecation` The default for ``keep_masked_maps`` and ``keep_masked_labels`` was changed from ``True`` to ``False`` in :class:`~maskers.NiftiMapsMasker`, :class:`~maskers.MultiNiftiMapsMasker`, :class:`~maskers.NiftiLabelsMasker`, :class:`~maskers.MultiNiftiLabelsMasker`, :class:`~regions.RegionExtractor` and :func:`~regions.img_to_signals_labels`. These parameters will be removed in Nilearn>=0.15.0 (:gh:`5632` by `Rémi Gau`_).
 
-- :bdg-danger:`Deprecation`  In the ``scrubbing`` strategy used by :func:`~interfaces.fmriprep.load_confounds`, the default value of ``fd_threshold`` was changed from ``0.2`` to ``0.5`` and that of ``std_dvars_threshold`` was changed from ``3.0`` to ``1.5`` to better match the values used by fmriprep (:gh:`5633` by `Rémi Gau`_).
```
Remi-Gau: revert

nilearn/image/image.py
```diff
@@ -41,13 +42,21 @@
...
     SurfaceImage,
     at_least_2d,
+    compute_adjacency_matrix,
     extract_data,
     find_surface_clusters,
 )
 from nilearn.surface.surface import get_data as get_surface_data
+
+from nilearn.surface.utils import (
+    check_polymesh_equal,
```
```suggestion
    check_polymesh_equal
```

Remi-Gau:
note to self: use smoothing surface in all places that currently throw not implemented errors or warnings

bthirion:
What is the status here ? Does it need a sprint or something like that ?

nilearn/image/tests/test_image.py
```diff
@@ -398,6 +402,74 @@ def test_smooth_img(affine_eye, tmp_path):
...
+
+    assert data.max() > smoothed_data.max()
+    assert data.min() < smoothed_data.min()
+
+
+def test_smooth_list_img_surface(surf_img_1d):
+    """Test smoothing list surface images."""
+    smoothed_img = smooth_img([surf_img_1d, surf_img_1d], fwhm=5)
+    assert isinstance(smoothed_img, list)
+    assert all(isinstance(x, SurfaceImage) for x in smoothed_img)
```
bthirion, line 439: Better check that each image is equal to the `smoothed surf_img_1d`

nilearn/image/tests/test_image.py
```diff
@@ -398,6 +402,74 @@ def test_smooth_img(affine_eye, tmp_path):
...
+    with pytest.raises(
+        ValueError, match="Part 'left' of PolyData instances are not equal"
+    ):
+        assert_surface_image_equal(smoothed_img, surf_img_1d)
+
+    data = get_surface_data(surf_img_1d)
+    smoothed_data = get_surface_data(smoothed_img)
+
+    assert data.max() > smoothed_data.max()
+    assert data.min() < smoothed_data.min()
```
bthirion, line 430: Could also assess that `data.var() > smoothed_data.var()`

nilearn/image/image.py
```diff
@@ -398,39 +401,201 @@ def smooth_img(imgs, fwhm):
...
+
     imgs = stringify_path(imgs)
-    if hasattr(imgs, "__iter__") and not isinstance(imgs, str):
+
+    if isinstance(imgs, SurfaceImage):
+        is_surface = True
+    elif hasattr(imgs, "__iter__") and not isinstance(imgs, str):
+        # Use hasattr() instead of isinstance
+        # to workaround a Python 2.6/2.7 bug
+        # See http://bugs.python.org/issue7624
```
bthirion: Can get rid of that now ;-)

nilearn/image/image.py
```diff
@@ -398,39 +401,201 @@ def smooth_img(imgs, fwhm):
...
+        The relative weighting of the center and the surround
+        in each iteration of the smoothing.
+        If the value of the knob is `k`,
+        then the total weight of vertices that are neighbors
+        of a given vertex (the vertex's surround)
+        is `2**k` times the weight of the vertex itself (the center).
+        A value of 0 (the default) means that, in each smoothing iteration,
+        each vertex is updated with the average of its value
+        and the average value of its neighbors.
+        A value of `-inf` results in no smoothing because the entire
```
bthirion: this should be tested

nilearn/image/image.py
```diff
@@ -398,39 +401,201 @@ def smooth_img(imgs, fwhm):
...
+        SurfaceImage with smoothed data at each vertex.
+
+    """
+    # First, calculate the center and surround weights for the
+    # center-surround knob.
+    center_weight = 1 / (1 + np.exp2(center_surround_knob))
+    surround_weight = 1 - center_weight
+    if surround_weight == 0:
+        # There's nothing to do in this case.
+        return new_img_like(img, img.data)
```
bthirion: to be tested

nilearn/surface/surface.py
```diff
@@ -1607,6 +1607,27 @@ def __getitem__(self, index):
...
+        (v0, v1, v2) = (
+            self.coordinates[idx0],
+            self.coordinates[idx1],
+            self.coordinates[idx2],
+        )
+        edge1 = v1 - v0
+        edge2 = v2 - v0
+        cross_prod = np.cross(edge1, edge2)
+        area = 0.5 * np.sqrt(np.sum(cross_prod**2, axis=1))
+        return np.sum(area)
```
bthirion, line 1629: should be tested

bthirion (commented):
The code looks OK a few more tests are welcome

nilearn/image/image.py
```diff
@@ -398,39 +401,201 @@ def smooth_img(imgs, fwhm):
...
+        if isinstance(mesh, FileMesh):
+            mesh = mesh.loaded()
+
+        # Compute average vertex area
+        avg_vertex_area = mesh._area / mesh.n_vertices
+
+        # Compute number of iterations using empirical formula
+        niters.append(
+            math.floor(
+                1.14 * (4 * math.pi * G_STD**2) / (7 * avg_vertex_area) + 0.5
```
bthirion: feels a but like black magic, but it would be no worse than the freesurfer one ;-)

nilearn/image/image.py
```diff
@@ -398,39 +401,201 @@ def smooth_img(imgs, fwhm):
...
+    imgs : SurfaceImage
+        The surface whose is to be smoothed.
+        In the case of 2D data, each sample is smoothed independently.
+
+    iterations : :obj:`tuple` of :obj:`int` >=0
+        The number of times to repeat the smoothing operation
+        (it must be a positive value).
+        One value per mesh in the image.
+
+    distance_weights : :obj:`bool`, default = False
```
bthirion: `distance_weights` and `center_surround_knob` are not used in the public API, so it would be safe to get rid of them.

bthirion (commented):
2 more comments

nilearn/image/image.py
```diff
@@ -398,39 +401,201 @@ def smooth_img(imgs, fwhm):
...
+        new_data[hemi] = np.reshape(np.asarray(tmp), np.shape(data))
+
+    return new_img_like(img, new_data)
+
+
+def _mris_fwhm_to_niters(fwhm, img) -> list[int]:
+    """Convert a desired FWHM to number of smoothing iterations for surface.
+
+    Adapted from freesurfer
+    https://github.com/freesurfer/freesurfer/tree/main/utils/mrisutils.cpp#L1100
```
bthirion:
```suggestion
    https://github.com/freesurfer/freesurfer/blob/dev/utils/mrisutils.cpp#L1102
```

nilearn/image/image.py
```diff
@@ -398,39 +401,162 @@ def smooth_img(imgs, fwhm):
...
         for a detailed description of the valid input types).
+
     %(fwhm)s
 
     Returns
     -------
-    :class:`nibabel.nifti1.Nifti1Image` or list of
-        Filtered input image. If `imgs` is an iterable,
-        then `filtered_img` is a list.
+    smoothed input image or list of smoothed input image.
```
emdupre: Niimg-like object, :obj:`~nilearn.surface.SurfaceImage`.
        Smoothed input image or surface.

nilearn/image/image.py
```diff
@@ -398,39 +401,162 @@ def smooth_img(imgs, fwhm):
...
+    img : surface image
+
+    Returns
+    -------
+    niters: list of number of smoothing iterations (one per mesh in the image)
+    """
+    if fwhm is None:
+        fwhm = 0
+    # Convert FWHM to standard deviation of Gaussian kernel
+    G_STD = fwhm / math.sqrt(math.log(256.0))
```
emdupre:
```suggestion
    std_kernel = fwhm / math.sqrt(math.log(256.0))
```

nilearn/image/image.py
```diff
@@ -398,39 +401,162 @@ def smooth_img(imgs, fwhm):
...
+        if isinstance(mesh, FileMesh):
+            mesh = mesh.loaded()
+
+        # Compute average vertex area
+        avg_vertex_area = mesh._area / mesh.n_vertices
+
+        # Compute number of iterations using empirical formula
+        niters.append(
+            math.floor(
+                1.14 * (4 * math.pi * G_STD**2) / (7 * avg_vertex_area) + 0.5
```
emdupre:
```suggestion
                1.14 * (4 * math.pi * std_kernel**2) / (7 * avg_vertex_area) + 0.5
```

nilearn/surface/surface.py
```diff
@@ -1014,3 +1014,229 @@ def check_surface(surface):
...
+    surface : Surface-like
+        The surface whose adjacency matrix is to be computed.
+
+    values : { 'len' | 'invlen' | 'ones'}, optional
+        If `values` is `'ones'` (the default), then the returned matrix
+        contains uniform values in the cells representing edges. If the value is
+        `'len'` then the cells contain the edge length of the represented
+        edge. If the value is `'invlen'`, then the the inverse of the distances
+        are returned.
+
```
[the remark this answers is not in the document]
emdupre: Do we test or showcase `invlen` (I did not see if so) ? 
I'm tempted to provide the most minimal implementation possible for now and scale it as we have new motivating use cases.

nilearn/image/tests/test_image.py
Remi-Gau, line 1: paranoid me thinks that we should have a test to check smoothing of 2D surface images

not that AFAICT we do not have tests that check directly running smoothing on 4D niftis

Remi-Gau:
yay let's merge!!!

Remi-Gau:
thanks @jyeatman @noahbenson for this (and your patience for us to get back to it)

noahbenson:
> thanks @jyeatman @noahbenson for this (and your patience for us to get back to it)

And thank you for doing all the substantive work here! 😁

alexisthual:
Super cool, congrats all for merging this! :)

PR #3269, opened 2022-05-31
DOC: Fixes import in surface/data README file

arokem:
This a one-liner change in the README file that describes how the test data for the surface module is generated. The namespace organization of this module might have changed since last time this was updated, causing this regression.


NicolasGensollen (approved):
Indeed, thanks for catching this @arokem !

bthirion (approved):
LGTM. Actually, this naming is not great.

bthirion:
Merging, then.

PR #3271, opened 2022-06-05
avoid creation of int64 nifti in cluster inference + update tests

jeromedockes:
quick fix for #3227 recently breaking the build: ~store cluster sizes in floats not ints. another option is to store in int32 rather than int64, as for realistic resolutions even if a cluster fills the whole image it will still fit in int32. @tsalo  please advise which you prefer~

create the images with `nilearn.image.new_img_like` rather than `nib.Nifti1Image`; `new_img_like` takes care of converting int64 to int32 if it does not cause an overflow

### DIFF
--- a/nilearn/mass_univariate/permuted_least_squares.py
+++ b/nilearn/mass_univariate/permuted_least_squares.py
@@ -11,6 +11,7 @@ import joblib
 import nibabel as nib
 import numpy as np
 from nilearn.masking import apply_mask
+from nilearn import image
 from nilearn.mass_univariate._utils import (
     _calculate_cluster_measures,
     _calculate_tfce,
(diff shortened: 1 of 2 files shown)


jeromedockes:
I think we should add to the developers doc to always use nilearn functions such as `new_img_like`, `resample_to_img` etc rather than raw `nibabel.Nifti1Image` constructor wherever possible.

also the permuted_ols is almost 400 lines long and kind of hard to read so may be due for a bit of refactoring

tsalo (approved):
It looks like you found a better solution than what you initially asked about. The change makes sense to me.

tsalo:
>permuted_ols is almost 400 lines long and kind of hard to read so may be due for a bit of refactoring

I completely agree. It would also be nice to refactor `_permuted_ols_on_chunk` while we're at it.

jeromedockes:
thanks! as this is a trivial change with one approval and needed to fix the build of the `main` branch I will merge now

PR #3277, opened 2022-06-10
CI: Test pre-release versions of dependencies

effigies:
Redo of #2224, which seems to have gotten lost with the transition away from Travis.

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -447,7 +447,7 @@ def _pad_array(array, pad_sizes):
                                           upper_paddings,
                                           new_shape)]
 
-    padded[tuple(target_slices)] = array[source_slices].copy()
+    padded[tuple(target_slices)] = array[tuple(source_slices)].copy()
     return padded
 
 
--- a/nilearn/maskers/nifti_labels_masker.py
+++ b/nilearn/maskers/nifti_labels_masker.py
@@ -243,7 +243,7 @@ class NiftiLabelsMasker(BaseMasker, CacheMixin):
             # previously fitted with no func image and is re-fitted
             if 'warning_message' in self._report_content:
                 self._report_content['warning_message'] = None
-            labels_image = image.load_img(labels_image, dtype=int)
+            labels_image = image.load_img(labels_image, dtype='int32')
             labels_image_data = image.get_data(labels_image)
             labels_image_affine = labels_image.affine
             # Number of regions excluding the background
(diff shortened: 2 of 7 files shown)


nilearn/image/image.py
```diff
@@ -447,7 +447,7 @@ def _pad_array(array, pad_sizes):
                                           upper_paddings,
                                           new_shape)]
 
-    padded[tuple(target_slices)] = array[source_slices].copy()
+    padded[tuple(target_slices)] = array[tuple(source_slices)].copy()
```
effigies, line 450: The `tuple()` fixes address a deprecation that will begin failing in numpy 1.23.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -243,7 +243,7 @@ def _reporting(self):
             # previously fitted with no func image and is re-fitted
             if 'warning_message' in self._report_content:
                 self._report_content['warning_message'] = None
-            labels_image = image.load_img(labels_image, dtype=int)
+            labels_image = image.load_img(labels_image, dtype='int32')
```
effigies, line 246: Nibabel will now require you to specify an integer size when running `set_data_dtype()` on a `Nifti1Image`. I chose `int32` to be conservative, but for a label image `int16` is probably more appropriate. Let me know what you think.

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -338,8 +338,8 @@ def test_signal_extraction_with_maps_and_labels():
 
     # Check that NaNs in regions inside mask are replaced with zeros
     region1 = labels_data == 2
-    indices = [ind[:1] for ind in np.where(region1)]
-    get_data(fmri_img)[indices + [slice(None)]] = np.nan
+    indices = tuple(ind[:1] for ind in np.where(region1))
+    get_data(fmri_img)[indices] = np.nan
```
effigies, line 342: Adding a `slice(None)` is just adding an extra `:`, which has no effect on the result. I just dropped it, but we could restore it by adding `(slice(None), )` instead.

effigies (commented):
The [Python 3.9 error](https://github.com/nilearn/nilearn/runs/6840329017?check_suite_focus=true) looked stochastic to me. Hope re-running with one more pre-release fix will help.

nilearn/maskers/nifti_labels_masker.py
jeromedockes: the `load_img` call shouldn't be necessary as this is always an in-memory Nifti1Image, and rather than the reporting code, the code that creates the `labels_img_` and `_resampled_labels_img_` should be responsible for making sure it has a sensible (ie int16 or int32) dtype. we also discussed adding a check for int images in `load_img`. but I think that is outside the scope of adding the pre-releases CI job. I think the conservative int32 choice is perfectly reasonable so I suggest we keep the choice you made

jeromedockes (approved):
super useful, thanks!!

jeromedockes:
thanks again, merging :)

PR #3280, opened 2022-06-14
[DOC] Add badges to the change log

NicolasGensollen:
I was looking at the really beautiful scikit-learn changelog this morning: https://scikit-learn.org/stable/whats_new/v1.1.html
And thought that the badge system they use could potentially be useful for Nilearn (in addition of being pretty... 😅).
This PR is just a proposal to include badges within the change log. I created a few random instances (`NEW`, `GLM`, `DOC`...),  and modified `doc/changes/latest.rst` for demoing.
I'm totally fine changing them to display more useful information, or to close the PR if people think this is not adding much value.


ymzayek:
@NicolasGensollen thanks for the suggestions, I quite like this visually! I think the only downside I see is it would be an additional thing to remind people to do in order to keep it consistent. But I guess that can be solved by updating the whatsnew just before a release. Also, I know these are just examples but just to note, we should either keep the section headers (e.g. NEW, fixes,...) or add them as badges but not both.

NicolasGensollen:
> But I guess that can be solved by updating the whatsnew just before a release. 

Yes, probably more something to do when cleaning the logs before a release.

> Also, I know these are just examples but just to note, we should either keep the section headers (e.g. NEW, fixes,...) or add them as badges but not both.

Absolutely, this is just an example to show what it looks like, and some labels are redundant with the headers which we definitely don't want. Some of them like `DOC` or `GLM` makes sense though. I also thought about `CI` / `INFRA`, `MAINT`, `TESTS`... but definitely open to suggestions ! :)

bthirion (commented):
Looks great !
Would it be expensive to apply it backwards to previous releases ?

NicolasGensollen:
> Would it be expensive to apply it backwards to previous releases ?

I don't think so, especially if we do it gradually (in different PRs) starting from the latest releases. Once we have agreed on some useful labels, I don't mind doing it.

bthirion:
Gr8
The only way to identify the good labels is to try and label existing items...

ymzayek:
Hi @NicolasGensollen will you be able to pick this back up and finish it in the next week or so before the release planned for the first week of December? If not, let me know so that I can take it over.

NicolasGensollen:
> Hi @NicolasGensollen will you be able to pick this back up and finish it in the next week or so before the release planned for the first week of December? If not, let me know so that I can take it over.

Hi @ymzayek 
I can pick this back up for the `0.10.0` release.
Thanks for the reminder !

ymzayek:
@NicolasGensollen will this be possible to finish this week? Otherwise, maybe better to take it off the milestones for this release. I'm also thinking even if we try to finish it, it could warrant longer discussion for everyone to agree on the labels. WDYT?

ymzayek:
@NicolasGensollen sure I would say go ahead in that case. We can see how far we get on it and whether or not it goes for the release. In any case I think it's a nice addition that we should at some point include so it's not work lost!

In terms of the release, I would like to do it by the end of this week, aiming for Friday morning at the latest.

ymzayek:
Hi @NicolasGensollen, is this feasible by the end of this week?

NicolasGensollen:
> Hi @NicolasGensollen, is this feasible by the end of this week?

Hi @ymzayek, I had a quick look but there was a lot of changes since I initially proposed this PR (especially, the new documentation theme).
I naively tried to copy-paste the CSS in `doc/themes/custom.css` but it doesn't seem to render as expected.
I don't have the time to look into it atm, so I believe we can close this PR (unless you want to work on it of course).

ymzayek:
Hi @NicolasGensollen ok no problem thanks for the update! I think we can keep it open for now and I will try to have a look. But I will remove it from the 0.10 milestones.

ymzayek:
Built in syntax for badges are available with the new theme so it's actually quite simple to add them. See https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html#badges 

See example of badges I added here: https://htmlpreview.github.io/?https://raw.githubusercontent.com/ymzayek/nilearn/add-badges/doc/_build/html/changes/whats_new.html

If there is still enthusiasm for this PR I'll open a new one to supersede this.

emdupre:
> If there is still enthusiasm for this PR I'll open a new one to supersede this.

Pinging in my personal enthusiasm :)

NicolasGensollen:
Good find @ymzayek ! 💯 
I'm also enthusiastic to see this happen, so feel free to close this and open a new PR ! 👍

PR #1500, opened 2017-08-24
Update image.py

Leo-YF-Liu:
dtype issue for anlyzeImage

PR #3286, opened 2022-06-22
[FIX] Default to `np.uint8` for encoding boolean data based on image type

ymzayek:
Replaces and closes #1500 
Related to nipy/nibabel#913

- Defaulting boolean input to `np.uint8` instead of `np.int8` for function `image.new_img_like`

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -61,3 +61,4 @@ Changes
   separate steps (:gh:`3274` by `David G Ellis`_).
 - Private submodules, functions, and classes from the :mod:`~nilearn.decomposition` module now start with a "_" character to make it clear that they are not part of the public API (:gh:`3141` by `Nicolas Gensollen`_).
 - Convert references in ``nilearn/glm/regression.py`` and ``nilearn/glm/thresholding.py`` to use footcite/footbibliography (:gh:`3302` by `Ahmad Chamma`_).
+- Boolean input data in :func:`~image.new_img_like` now defaults to `np.uint8` instead of `np.int8`
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -61,4 +61,4 @@ Changes
   separate steps (:gh:`3274` by `David G Ellis`_).
 - Private submodules, functions, and classes from the :mod:`~nilearn.decomposition` module now start with a "_" character to make it clear that they are not part of the public API (:gh:`3141` by `Nicolas Gensollen`_).
 - Convert references in ``nilearn/glm/regression.py`` and ``nilearn/glm/thresholding.py`` to use footcite/footbibliography (:gh:`3302` by `Ahmad Chamma`_).
-- Boolean input data in :func:`~image.new_img_like` now defaults to `np.uint8` instead of `np.int8`
+- Boolean input data in :func:`~image.new_img_like` now defaults to `np.uint8` instead of `np.int8` (:gh:`3286` by `Yasmin Mzayek`_).
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -755,8 +755,7 @@ def new_img_like(ref_niimg, data, affine=None, copy_header=False):
     if affine is None:
         affine = ref_niimg.affine
     if data.dtype == bool:
-        default_dtype = np.uint8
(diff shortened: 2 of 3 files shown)


nilearn/image/image.py
```diff
@@ -755,9 +755,7 @@ def new_img_like(ref_niimg, data, affine=None, copy_header=False):
     if affine is None:
         affine = ref_niimg.affine
     if data.dtype == bool:
-        default_dtype = np.int8
-        if isinstance(ref_niimg, nibabel.freesurfer.mghformat.MGHImage):
-            default_dtype = np.uint8
+        default_dtype = np.uint8
```
NicolasGensollen, line 758: It doesn't look like you need the intermediate variable `default_dtype`

nilearn/image/tests/test_image.py
```diff
@@ -824,3 +824,21 @@ def test_new_img_like_mgh_image():
...
     new_img_like(niimg, data.astype(float), niimg.affine, copy_header=True)
+
+
+def test_new_img_like_boolean_data_mgh():
+    """Check defaulting boolean input data to np.uint8 dtype is valid for
+    encoding with nibabel image class MGHImage.
+    """
+    data = np.random.randn(5, 5, 5).astype('uint8')
+    in_img = nibabel.freesurfer.MGHImage(dataobj=data, affine=np.eye(4))
+    new_img_like(in_img, data=in_img.get_fdata() > 0.5)
```
NicolasGensollen, line 835: I might be missing something but I don't understand what is tested here. I think you need to assert that the new image is composed of `uint8`, right?

nilearn/image/tests/test_image.py
```diff
@@ -824,3 +824,21 @@ def test_new_img_like_mgh_image():
...
+
+
+def test_new_img_like_boolean_data_analyzeimage():
```
NicolasGensollen, line 838: I think that these two tests can be factorized. Maybe something like this:

```python
from nibabel import MGHImage, AnalyzeImage
@pytest.mark.parametrize("image", [MGHImage, AnalyzeImage])
def test_new_img_like_boolean(image):
    data = np.random.randn(5, 5, 5).astype('uint8')
    in_img = image(dataobj=data, affine=np.eye(4))
    ....
```

doc/changes/latest.rst
```diff
@@ -53,3 +53,4 @@ Changes
 - Refactors fmriprep confound loading such that that the parsing of the 
   relevant image file and the loading of the confounds are done in 
   separate steps (:gh:`3274` by `David G Ellis`_).
+- Boolean input data in :func:`~image.new_img_like` now defaults to `np.uint8` instead of `np.int8`
```
NicolasGensollen, line 56:
```suggestion
- Boolean input data in :func:`~image.new_img_like` now defaults to `np.uint8` instead of `np.int8` (:gh:`3286` by `Yasmin Mzayek`_).
```

NicolasGensollen (commented):
Thanks @ymzayek ! Looking good ! 
I made a few suggestions below.

nilearn/image/tests/test_image.py
ymzayek: Well before `new_img_like` was failing on some image classes when boolean input was defaulting to `np.int8`. I included tests for `MGHImage` and `AnalyzeImage` because I know they would have failed but I am not sure how to test this more generally because there may be other image classes that also don't accept  `np.int8`

ymzayek: I wonder if it is actually sufficient to have one test? It should generalize to all other image classes.
@effigies WDYT?

effigies: Apologies for not getting back. I'm not sure it "generalizes" exactly, since someone could write a new format that doesn't support uint8. It seems like a good practical test, though, as these are less-featured data formats.

ymzayek: Ok thanks for clarifying @effigies ! I think then we can keep it this way then unless someone else has another suggestion.

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

jeromedockes (approved):

bthirion (approved):
Thx !

bthirion:
Merging that one. Thx !

PR #3293, opened 2022-06-28
[FIX] Change setting of critical frequencies to meet `Wn[0] < Wn[1]`

### LINKED ISSUE #3288: Test error in scipy `1.9.0rc1`
The test is currently failing with scipy `1.9.0rc1`: See [here](https://github.com/nilearn/nilearn/runs/7067861378?check_suite_focus=true) for example. The failing test is:

https://github.com/nilearn/nilearn/blob/4242c0d03e3fc790b17a1d88ce5a4140633c6b6f/nilearn/decoding/tests/test_decoder.py#L435-L440

failing with error:

```
if Wn.size > 1 and not Wn[0] < Wn[1]:
>           raise ValueError("Wn[0] must be less than Wn[1]")
E           ValueError: Wn[0] must be less than Wn[1]
/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/scipy/signal/_filter_design.py:2468: ValueError
```

The error is due to this check which was added in this PR: https://github.com/scipy/scipy/pull/15886
See also the related issue: https://github.com/scipy/scipy/issues/15773.

ymzayek:
Closes #3288.
Currently `signal._check_wn` sets critical frequencies to an equal value below 1 if they are above the threshold of 1. A scipy update now enforces that its two-element sequence `Wn` for upper and lower critical frequencies should meet the condition `Wn[0] < Wn[1]`. This causes `test_decoder_apply_mask` to fail. 

- ~~Add negligible value to low-pass filter to satisfy condition~~
- Give warning to user to input appropriate values if critical frequencies are coerced into the same value

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -24,6 +24,8 @@ Fixes
 - Function ``nilearn.glm.second_level.second_level._check_second_level_input`` now raises an error when ``flm_object`` argument is ``False`` and ``second_level_input`` is a list of :class:`~glm.first_level.FirstLevelModel` (:gh:`3283` by `Matthieu Joulot`_).
 - Function :func:`~image.resample_img` now warns the user if the provided image has an ``sform`` code equal to 0 or None  (:gh:`3284` by `Matthieu Joulot`_).
 - Fix usage of ``scipy.stats.gamma.pdf`` in ``_gamma_difference_hrf`` function under ``nilearn/glm/first_level/hemodynamic_models.py``, which resulted in slight distortion of HRF (:gh:`3297` by `Kun CHEN`_).
+- Fix bug introduced due to a fix in the pre-release version of scipy (``1.9.0rc1``) which now enforces that elements of a band-pass filter must meet condition ``Wn[0] < Wn[1]``.
+  Now if band-pass elements are equal :func:`nilearn.signals.butterworth` returns an unfiltered signal with a warning (:gh:`3293` by `Yasmin Mzayek`_).
 
 Enhancements
 ------------
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -25,7 +25,7 @@ Fixes
 - Function :func:`~image.resample_img` now warns the user if the provided image has an ``sform`` code equal to 0 or None  (:gh:`3284` by `Matthieu Joulot`_).
 - Fix usage of ``scipy.stats.gamma.pdf`` in ``_gamma_difference_hrf`` function under ``nilearn/glm/first_level/hemodynamic_models.py``, which resulted in slight distortion of HRF (:gh:`3297` by `Kun CHEN`_).
 - Fix bug introduced due to a fix in the pre-release version of scipy (``1.9.0rc1``) which now enforces that elements of a band-pass filter must meet condition ``Wn[0] < Wn[1]``.
-  Now if band-pass elements are equal :func:`nilearn.signals.butterworth` returns an unfiltered signal with a warning (:gh:`3293` by `Yasmin Mzayek`_).
+  Now if band-pass elements are equal :func:`~nilearn.signal.butterworth` returns an unfiltered signal with a warning (:gh:`3293` by `Yasmin Mzayek`_).
 
 Enhancements
 ------------
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -261,8 +261,6 @@ def _check_wn(btype, freq, nyq):
         # results as pointed in the issue above. Hence, we forced the
         # critical frequencies to be slightly less than 1. but not 1.
         wn = 1 - 10 * np.finfo(1.).eps
-        if btype == 'low':
-            wn += 10e-16
         warnings.warn(
             'The frequency specified for the %s pass filter is '
             'too high to be handled by a digital filter (superior to '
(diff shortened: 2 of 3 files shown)


ymzayek:
On second thought, that would still entail changing the behavior in the function `signal._check_wn`. But I only see this function used in `butterworth`. We could alternatively keep it this way and also handle the other conditions I mentioned in my comment above and add a comment plus extend the warning message that is already being outputted.

ymzayek:
Sorry for all the comments as I am working this out but I think the problem perhaps lies in the fact that we do check for equality in `butterworth` before using `_check_wn` which under certain conditions converts the values to be equal. See here:
https://github.com/nilearn/nilearn/blob/e7bd43ab785a486c1642ffa71a924b61dc0344f0/nilearn/signal.py#L324-L330

bthirion:
But the above lines ensure that `high_pass < low_pass`, then ?

ymzayek:
@bthirion yes this checks the user input directly but then the function `_check_wn` is used after this check and under certain conditions it converts the values to be equal. And putting this function before the check wouldn't be a good idea either because then the user will get an error based on something the function is doing rather than their own input. I suggest we put a second check to make sure user input is between 0 and 1 and get rid of `_check_wn` or alternatively use `_check_wn` to raise this error instead of forcing the conversion.

bthirion:
> and under certain conditions it converts the values to be equal

Can't we simply avoid that ? Sorry, I don't know this code well, I may be a bit off.

ymzayek:
Yes by raising an error to enforce user behavior. I can implement this if no one objects.

bthirion:
I would avoid raising errors when not absolutely necessary. We should handle values that are equal, possibly with a warning.

ymzayek:
Ok then we can add something like this in the function [`butterworth`](https://github.com/nilearn/nilearn/blob/e7bd43ab785a486c1642ffa71a924b61dc0344f0/nilearn/signal.py#L343):
```python
    if len(critical_freq) == 2:
        btype = 'band'
        # Scipy fix enforces that critical frequencies cannot be equal.
        # See https://github.com/scipy/scipy/pull/15886. We add a small value
        # to the critical frequency of the low-pass filter to satisfy this
        # condition.
        if critical_freq[0] == critical_freq[1]:
            critical_freq[1] += 10e-16
    else:
        critical_freq = critical_freq[0]
```

I believe this is what you suggested @NicolasGensollen but I may have misunderstood. And then I would remove the change I originally made under `_check_wn`

bthirion:
Sounds reasonable to me.

NicolasGensollen:
@ymzayek Yes, this is what I had in mind with a possible warning to the user.
To be honest I haven't looked in details into this issue. It could also be worth looking into why we get this particular edge case when running the tests. It might be that we are using poor synthetic data somewhere.
In any case, the safety check plus warning won't hurt.

ymzayek:
Ok here's another go implementing what is described by @emdupre. Let me know what you all think.

NicolasGensollen (commented):
Thanks @ymzayek ! 👍 The proposed solution LGTM.
I'd just add tests and a what's new entry.

doc/changes/latest.rst
```diff
@@ -23,6 +23,8 @@ Fixes
 - :func:`~surface.load_surf_mesh` loaded FreeSurfer specific surface files (e.g. `.pial`) with a shift in the coordinates. This is fixed by adding the c_ras coordinates to the mesh coordinates (:gh:`3235` by `Yasmin Mzayek`_).
 - Function ``nilearn.glm.second_level.second_level._check_second_level_input`` now raises an error when ``flm_object`` argument is ``False`` and ``second_level_input`` is a list of :class:`~glm.first_level.FirstLevelModel` (:gh:`3283` by `Matthieu Joulot`_).
 - Function :func:`~image.resample_img` now warns the user if the provided image has an ``sform`` code equal to 0 or None  (:gh:`3284` by `Matthieu Joulot`_).
+- Fix bug introduced due to a fix in the pre-release version of scipy (``1.9.0rc1``) which now enforces that elements of a band-pass filter must meet condition ``Wn[0] < Wn[1]``.
+  Now if band-pass elements are equal :func:`nilearn.signals.butterworth` returns an unfiltered signal with a warning (:gh:`3293` by `Yasmin Mzayek`_).
```
NicolasGensollen, line 27:
```suggestion
  Now if band-pass elements are equal :func:`~nilearn.signal.butterworth` returns an unfiltered signal with a warning (:gh:`3293` by `Yasmin Mzayek`_).
```

nilearn/signal.py
```diff
@@ -340,6 +340,20 @@ def butterworth(signals, sampling_rate, low_pass=None, high_pass=None,
...
     if len(critical_freq) == 2:
         btype = 'band'
+        # Inappropriate parameter input might lead to coercion of both
+        # elements of critical_freq to a value just below 1.
+        # Scipy fix now enforces that critical frequencies cannot be equal.
+        # See https://github.com/scipy/scipy/pull/15886. If this is the case,
+        # we return the signals unfiltered.
+        if critical_freq[0] == critical_freq[1]:
+            warnings.warn(
+                'Signals are returned unfiltered because band-pass critical '
```
NicolasGensollen, line 350: I think we should add a test checking that this warning is correctly given if we provide equal freqs.

ymzayek:
Thanks @NicolasGensollen for the review. I added the checks as you suggested. Let me know if it can still be improved.

nilearn/tests/test_signal.py
```diff
@@ -178,6 +178,20 @@ def test_butterworth():
     np.testing.assert_almost_equal(out1, out2)
     np.testing.assert_(id(out1) != id(out2))
 
+    # Test check for equal values in critical frequencies
+    sampling = 1
+    low_pass = 2
+    high_pass = 1
```
bthirion, line 184: I'm confused: since high_pass < low_pass, I would consider that the critical frequencies are different here ? Why isn't it the case ?

ymzayek: > I think [the failing test's parameters](https://github.com/nilearn/nilearn/blob/00084e1aafc49681869de6c83aab3b6450dc7fb0/nilearn/decoding/tests/test_decoder.py#L451-L453) are now being correctly flagged. With the requested band-pass filter of 2-1 Hz and the provided TR of 1s, we're only letting through signal occurring at a 0.5-1s timescale, but because 0.5 overlaps with the Nyquist frequency for this sampling rate, it's getting coerced to ~1.

To expand on what Elizabeth describes above,  the passed values for high_pass and low_pass are manipulated inside `butterworth` by a call to `_check_wn`. This check changes each frequency element to freq / float(nyq), but given certain conditions it results in a value just below 1. If this happens for both elements then they are set to an equal value.

bthirion (approved):
OK, LGTM then.
Best,

PR #2644, opened 2020-12-23
Add pathlib.Path to string conversion

bakyazi:
#2642 
Check path parameters that is pathlib.Path instance. If it is, convert it to string.

PR #3300, opened 2022-07-05
[ENH] Add support for path-like objects where string filesystem paths are expected

### LINKED ISSUE #2642: accept `pathlib.Path` where strings representing filesystem paths are expected

ymzayek:
Closes #2642.
Replaces and closes #2803.
Also replaces #2644.

Allow `os.PathLike` and `pathlib.Path` objects to be passed to functions expecting a string.
Use function based on `nibabel.filename_parser._stringify_path` to convert them to string.

- [x] Add conversion to all relevant functions
- [x] Add tests
- [x] Update docstrings
- [x] Add whatsnew entry

### DIFF
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -17,7 +17,7 @@ from joblib import Memory
 MEMORY_CLASSES = (Memory, )
 
 import nilearn
-from .helpers import _stringify_path
+from .helpers import stringify_path
 
 
 __CACHE_CHECKED = dict()
@@ -28,7 +28,7 @@ def _check_memory(memory, verbose=0):
 
     Parameters
     ----------
-    memory : None or instance of joblib.Memory or str or path-like object
+    memory : None or instance of joblib.Memory or str or pathlib.Path
         Used to cache the masking process.
         If a str is given, it is the path to the caching directory.
 
@@ -42,7 +42,7 @@ def _check_memory(memory, verbose=0):
     """
     if memory is None:
         memory = Memory(location=None, verbose=verbose)
-    memory = _stringify_path(memory)
+    memory = stringify_path(memory)
     if isinstance(memory, str):
         cache_dir = memory
         if nilearn.EXPAND_PATH_WILDCARDS:
@@ -181,7 +181,7 @@ def cache(func, memory, func_memory_level=None, memory_level=None,
     func : function
         The function which output is to be cached.
 
-    memory : instance of joblib.Memory or string or path-like object
+    memory : instance of joblib.Memory or string or pathlib.Path
         Used to cache the function call.
 
     func_memory_level : int, optional
@@ -223,7 +223,7 @@ def cache(func, memory, func_memory_level=None, memory_level=None,
 
     if memory is not None and (func_memory_level is None or
                                memory_level >= func_memory_level):
-        memory = _stringify_path(memory)
+        memory = stringify_path(memory)
         if isinstance(memory, str):
             memory = Memory(location=memory, verbose=verbose)
         if not isinstance(memory, MEMORY_CLASSES):
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -17,6 +17,7 @@ from joblib import Memory
 MEMORY_CLASSES = (Memory, )
 
 import nilearn
+from .helpers import _stringify_path
 
 
 __CACHE_CHECKED = dict()
@@ -27,7 +28,7 @@ def _check_memory(memory, verbose=0):
 
     Parameters
     ----------
-    memory : None or instance of joblib.Memory or str
+    memory : None or instance of joblib.Memory or str or path-like object
         Used to cache the masking process.
         If a str is given, it is the path to the caching directory.
 
@@ -41,6 +42,7 @@ def _check_memory(memory, verbose=0):
     """
     if memory is None:
         memory = Memory(location=None, verbose=verbose)
+    memory = _stringify_path(memory)
     if isinstance(memory, str):
         cache_dir = memory
         if nilearn.EXPAND_PATH_WILDCARDS:
@@ -179,7 +181,7 @@ def cache(func, memory, func_memory_level=None, memory_level=None,
     func : function
         The function which output is to be cached.
 
-    memory : instance of joblib.Memory or string
+    memory : instance of joblib.Memory or string or path-like object
         Used to cache the function call.
 
     func_memory_level : int, optional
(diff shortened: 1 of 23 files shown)


jeromedockes:
Great! small detail: we usually avoid using "private" functions of other packages. Especially as it does something trivial (something like 
```python
return path.__fspath__() if isinstance(path, os.PathLike) else path
```
 should do the trick) I would suggest adding a function in nilearn rather than using the nibabel private `_stringify_path`

bthirion (commented):
Looks reasonable so far, if we dependonly on nibabel versions that support it.

ymzayek:
> Looks reasonable so far, if we dependonly on nibabel versions that support it.

I assumed so because the original issue was open 2 years ago but I will double check!

jeromedockes:
looks great, thanks! the failures seem unrelated to this pr, but rather due to #3301 and #3293

nilearn/_utils/niimg_conversions.py
```diff
@@ -191,9 +192,9 @@ def check_niimg(niimg, ensure_ndim=None, atleast_4d=False, dtype=None,
     ----------
     niimg : Niimg-like object
         See https://nilearn.github.io/stable/manipulating_images/input_output.html  # noqa:E501
-        If niimg is a string, consider it as a path to Nifti image and
-        call nibabel.load on it. The '~' symbol is expanded to the user home
-        folder.
+        If niimg is a string or path-like object, consider it as a path to
```
NicolasGensollen, line 195: I'm wondering whether we should create a standard doc entry for `niimg` ? WDYT?

nilearn/_utils/helpers.py
```diff
@@ -145,3 +146,7 @@ def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
         return wrapper
     return _remove_params
+
+
+def _stringify_path(path):
```
NicolasGensollen, line 151: Maybe add a one line docstring?

nilearn/tests/test_niimg.py
```diff
@@ -77,3 +78,11 @@ def test_img_data_dtype():
     # Verify that the distinction is worth making
     assert any(dtype_matches)
     assert not all(dtype_matches)
+
+
+def test_load_niimg():
+    img = Nifti1Image(np.zeros((10, 10, 10)), np.eye(4))
```
NicolasGensollen, line 84: It seems like several tests of this file use this kind of toy data. I think we could start factorizing by writing a fixture.
It will make things easier when further factorizing things across Nilearn in a `data_generation` module as we discussed some time ago with @jeromedockes 
WDYT?

NicolasGensollen (commented):
Looks good so far! 👍 Thanks @ymzayek !
Some small suggestions:

nilearn/_utils/niimg_conversions.py
ymzayek: I was actually thinking the same thing!

nilearn/_utils/helpers.py
ymzayek: Yes can do

nilearn/_utils/niimg_conversions.py
ymzayek: On second thought, is this applicable in the `_utils` module given that it's private? I haven't seen this standardization in any of the other files inside this module (but I didn't look too extensively). So I was wondering if this is by choice or just hasn't been done yet.

NicolasGensollen: The docstrings in the `_utils` module are indeed not rendered since it is private, but for docstrings of public functions, that could still be useful.

ymzayek: Ok I will hold off to doing that in this PR then but I agree it would be useful.

nilearn/_utils/helpers.py
```diff
@@ -145,3 +146,21 @@ def wrapper(*args, **kwargs):
...
```
jeromedockes, line 151: not sure it would need to start with `_` as it is in the internal `_utils` module, and meant to be used outside of it. however function names in nilearn are not consistent in that regard so we can keep it like this for now and maybe consider clarifying this in another pr

nilearn/_utils/niimg.py
```diff
@@ -123,7 +124,7 @@ def load_niimg(niimg, dtype=None):
     """
     from ..image import new_img_like  # avoid circular imports
 
-    if isinstance(niimg, str):
+    if isinstance(niimg, (str, os.PathLike)):
```
jeromedockes, line 127: maybe call `_stringify_path` on it first and then just check `isinstance(niimg, str)` -- this way it will work with any nibabel version

nilearn/image/resampling.py
```diff
@@ -435,7 +435,7 @@ def resample_img(img, target_affine=None, target_shape=None,
                    "or 'nearest' but it was set to '{0}'").format(interpolation)
         raise ValueError(message)
 
-    if isinstance(img, str):
+    if isinstance(img, (str, os.PathLike)):
```
jeromedockes, line 438: here also we could call `_stringify_path` first?

nilearn/surface/surface.py
```diff
@@ -799,7 +799,7 @@ def load_surf_mesh(surf_mesh):
     """
 
     # if input is a filename, try to load it
-    if isinstance(surf_mesh, str):
+    if isinstance(surf_mesh, (str, os.PathLike)):
```
jeromedockes, line 802: use `_stringify_path`?

nilearn/_utils/cache_mixin.py
```diff
@@ -27,7 +28,7 @@ def _check_memory(memory, verbose=0):
 
     Parameters
     ----------
-    memory : None or instance of joblib.Memory or str
+    memory : None or instance of joblib.Memory or str or path-like object
```
jeromedockes, line 31: maybe we could add "path-like" to the glossary? alternatively, we can just call them `pathlib.Path` as this will be the vast majority of "path-like" object actually used by nilearn users

nilearn/_utils/helpers.py
ymzayek: I will change it here since it is quick but I agree on opening an issue/pr to standardize function naming.

bthirion (approved):
I'm happy with this PR, but indeed it could be worthwhile opening issues for the given suggestions.

jeromedockes (approved):
thanks!

ymzayek:
Ok thanks for the reviews! Merging.

PR #3305, opened 2022-07-11
[ENH] Improve test checking for removal of temporary file

### LINKED ISSUE #3301: Seemingly random test failure with `test_temp_file_removing`
There is a test that checks if a temporary file was removed correctly and it seems to fail randomly every once in a while. I am looking into it to see if I can improve the test. 
See [test here](https://github.com/nilearn/nilearn/blob/main/nilearn/plotting/tests/test_html_document.py)
See [failure here](https://dev.azure.com/Parietal/Nilearn/_build/results?buildId=4733&view=logs&j=8d5ac5f6-e89e-50f1-f198-d1a0983ad562&t=73659071-ebed-5a28-0af2-8b5a9c385fe0&l=11372)

ymzayek:
Closes #3301.

- Add ~~`_pid`~~ `_temp_file_removing_proc` attribute to class `HTMLDocument`
- Store suprocess in  `_remove_after_n_seconds` and return it
- ~~Use returned object to get pid of subprocess and store it in `_pid`~~
- Use ~~`os.waitpid`~~ `subprocess.Popen.wait` to pause process until subprocess is finished
    - This replaces `time.sleep(1.5)`

### DIFF
--- a/nilearn/plotting/html_document.py
+++ b/nilearn/plotting/html_document.py
@@ -21,7 +21,9 @@ def set_max_img_views_before_warning(new_value):
 
 def _remove_after_n_seconds(file_name, n_seconds):
     script = os.path.join(os.path.dirname(__file__), 'rm_file.py')
-    subprocess.Popen([sys.executable, script, file_name, str(n_seconds)])
+    proc = subprocess.Popen([sys.executable, script, file_name,
+                             str(n_seconds)])
+    return proc
 
 
 class HTMLDocument(object):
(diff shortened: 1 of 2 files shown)


nilearn/plotting/html_document.py
```diff
@@ -44,6 +46,7 @@ def __init__(self, html, width=600, height=400):
         self.height = height
         self._temp_file = None
         self._check_n_open()
+        self._proc = None
```
jeromedockes, line 49: maybe we could use a more explicit name such as `self._temp_file_removing_proc`?

jeromedockes (approved):
looks great, thanks! LGTM apart from the detail of the attribute name

NicolasGensollen (approved):
LGTM, thanks for fixing this @ymzayek !

PR #3306, opened 2022-07-12
[FIX] Pass alpha on to glass brain when plotting connectome

koenhelwegen:
Minor fix: `alpha` in `plotting.plot_connectome` is currently not passed on to `plot_glass_brain()`.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -26,6 +26,7 @@ Fixes
 - Fix usage of ``scipy.stats.gamma.pdf`` in ``_gamma_difference_hrf`` function under ``nilearn/glm/first_level/hemodynamic_models.py``, which resulted in slight distortion of HRF (:gh:`3297` by `Kun CHEN`_).
 - Fix bug introduced due to a fix in the pre-release version of scipy (``1.9.0rc1``) which now enforces that elements of a band-pass filter must meet condition ``Wn[0] < Wn[1]``.
   Now if band-pass elements are equal :func:`~nilearn.signal.butterworth` returns an unfiltered signal with a warning (:gh:`3293` by `Yasmin Mzayek`_).
+- Fix bug in behavior of ``alpha`` in ``nilearn.plotting.plot_connectome`` (:gh:`3306` by `Koen Helwegen`_).
 
 Enhancements
 ------------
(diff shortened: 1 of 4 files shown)


ymzayek (commented):
@koenhelwegen thanks for catching that! Can you update the PR title to have the prefix [FIX]? See [here](https://nilearn.github.io/stable/development.html#pr-structure). 

I think it might also be good to add a smoke test in `test_plot_connectome.py` with several different values for alpha. 

And then can you add an entry to `doc/changes/latest.rst` and your information to `doc/changes/names.rst`?

koenhelwegen:
Thanks for quickly picking this up @ymzayek!

I added the smoke test, currently only with valid values for `alpha`. Do you think it is useful to test an error gets raised for invalid values as well?

doc/changes/latest.rst
```diff
@@ -26,6 +26,7 @@ Fixes
 - Fix usage of ``scipy.stats.gamma.pdf`` in ``_gamma_difference_hrf`` function under ``nilearn/glm/first_level/hemodynamic_models.py``, which resulted in slight distortion of HRF (:gh:`3297` by `Kun CHEN`_).
 - Fix bug introduced due to a fix in the pre-release version of scipy (``1.9.0rc1``) which now enforces that elements of a band-pass filter must meet condition ``Wn[0] < Wn[1]``.
   Now if band-pass elements are equal :func:`~nilearn.signal.butterworth` returns an unfiltered signal with a warning (:gh:`3293` by `Yasmin Mzayek`_).
+- Fix bug in behavior of ``alpha`` in ``nilearn.plotting.plot_connectome`` (:gh:`3306` by `Koen Helwegen`_).
```
ymzayek, line 29:
```suggestion
- Fix bug in behavior of ``alpha`` in :func:`~plotting.plot_connectome` (:gh:`3306` by `Koen Helwegen`_).
```
The description can also be a bit more explicit, such as 

The parameter \`\`alpha\`\` is now correctly passed to :func:\`~plotting.plot_glass_brain\` in :func:\`~plotting.plot_connectome\`

ymzayek:
> I added the smoke test, currently only with valid values for `alpha`. Do you think it is useful to test an error gets raised for invalid values as well?

Thanks! I don't think it would hurt but I'm not sure it is necessary since checking `alpha` comes from a matplotlib module and is not something implemented in nilearn.
Perhaps @NicolasGensollen or @jeromedockes can chime in on this.

doc/changes/latest.rst
koenhelwegen: Thanks, updated in [e37b9ce](https://github.com/nilearn/nilearn/pull/3306/commits/e37b9cea58ec59ce3b7a36f29d6ffe56b37afd2f)

NicolasGensollen (approved):
LGTM, thanks @koenhelwegen !

NicolasGensollen:
@ymzayek  I don't think we need to test that an error is raised for invalid values of alpha because this is handled in matplotib AFAIK.

ymzayek (approved):
LGTM too!

ymzayek:
Merging

PR #3312, opened 2022-07-25
[FIX] do not mask view_img background with MNI brain mask

### LINKED ISSUE #3120: `view_img` always masks `bg_img` with MNI mask
when plotting with view_img, it appears the bg image is being "masked" by a different brain (perhaps MNI?)
image

problem persists even when using same image as bg_image and plotted img

jeromedockes:
Closes #3120

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -27,6 +27,7 @@ Fixes
 - Fix bug introduced due to a fix in the pre-release version of scipy (``1.9.0rc1``) which now enforces that elements of a band-pass filter must meet condition ``Wn[0] < Wn[1]``.
   Now if band-pass elements are equal :func:`~nilearn.signal.butterworth` returns an unfiltered signal with a warning (:gh:`3293` by `Yasmin Mzayek`_).
 - The parameter ``alpha`` is now correctly passed to :func:`~plotting.plot_glass_brain` in :func:`~plotting.plot_connectome` (:gh:`3306` by `Koen Helwegen`_).
+- Fix plotting of background image in :func:`~nilearn.plotting.view_img` when the background is not the MNI template.
 
 Enhancements
 ------------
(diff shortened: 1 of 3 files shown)


jeromedockes:
current behaviour (main branch).
incorrect results when using the MNI background are expected, as the foreground image (the one linked in the issue) is not in MNI space. However, we also get incorrect results when using background=None or the same image as the background, because the MNI brain mask is always used to crop the background. Also we see some holes in the mni template because it is masked by thresholding its values rather than using the MNI brain mask

Screenshot from 2022-07-25 18-13-15
Screenshot from 2022-07-25 18-13-04
Screenshot from 2022-07-25 18-12-55
Screenshot from 2022-07-25 18-12-31

jeromedockes:
with the proposed fix.
when using the mni template we do not have the holes anymore (note it is still not aligned with this particular image which is not in mni space); when using bg = None the halo of the MNI template does not appear on the empty white background; when using the image itself as background it is not cropped with the contours of the MNI template anymore.
when providing a background image, if its data (output of `nilearn.image.get_data`) is a masked array, masked values are transparent -- as in the 3rd image from the top. otherwise the whole background image is shown and in this case the `black_bg` parameter is not that useful (4th image). this is implemented to support what was decided in the issue and core dev meeting (MNI template would continue to be masked but user-provided images would not at least by default); but users can also take advantage of it to crop their own background images

Screenshot from 2022-07-25 18-12-21
Screenshot from 2022-07-25 18-12-06
Screenshot from 2022-07-25 18-11-55
Screenshot from 2022-07-25 18-11-42

bthirion (approved):
LGTM, thx.

ymzayek (approved):
LGTM!

jeromedockes:
thanks for the reviews!

PR #3315, opened 2022-07-29
[MAINT] Fix broken documentation links

### LINKED ISSUE #3303: Broken links due to cleaning of `nilearn.github.io`
I noticed some broken links due to URLs referring to pages based on the old website structure. After cleaning `nilearn.github.io` we may find more of these.

### Affected pages
One example is in the docstring of the [`SearchLight` class](https://nilearn.github.io/stable/modules/generated/nilearn.decoding.SearchLight.html#nilearn.decoding.SearchLight) where http://nilearn.github.io/manipulating_images/input_output.html is referenced several times. This should be http://nilearn.github.io/stable/manipulating_images/input_output.html 

I haven't spent much time on this but I assume there are other affected pages.

ymzayek:
Closes #3303.

- Add stable directory to fix broken links

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -110,7 +110,7 @@ def load_niimg(niimg, dtype=None):
     -----------
 
     niimg: Niimg-like object
-        See https://nilearn.github.io/stable/manipulating_images/input_output.html  # noqa:E501
+        See https://nilearn.github.io/stable/manipulating_images/input_output.html  # noqa: E501
         Image to load.
 
     dtype: {dtype, "auto"}
@@ -155,7 +155,7 @@ def _is_binary_niimg(niimg):
     Parameters
     ----------
     niimg: Niimg-like object
-        See https://nilearn.github.io/stable/manipulating_images/input_output.html  # noqa:E501
+        See https://nilearn.github.io/stable/manipulating_images/input_output.html  # noqa: E501
         Image to test.
 
     Returns
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -110,7 +110,7 @@ def load_niimg(niimg, dtype=None):
     -----------
 
     niimg: Niimg-like object
-        See https://nilearn.github.io/stable/manipulating_images/input_output.html  # noqa: E501
+        See https://nilearn.github.io/stable/manipulating_images/input_output.html.  # noqa: E501
         Image to load.
 
     dtype: {dtype, "auto"}
@@ -155,7 +155,7 @@ def _is_binary_niimg(niimg):
     Parameters
     ----------
     niimg: Niimg-like object
-        See https://nilearn.github.io/stable/manipulating_images/input_output.html  # noqa: E501
+        See https://nilearn.github.io/stable/manipulating_images/input_output.html.  # noqa: E501
         Image to test.
 
     Returns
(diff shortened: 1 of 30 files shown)


bthirion (approved):
LGTM, I don't see any occurrence missed.
Thx !

htwangtw (approved):
LGTM

ymzayek:
I just realized that the noqa comments are rendered in the docs as sphinx doesn't ignore them... is there a way to avoid this? Or what is the best way to reconcile PEP8 line too long errors and long links in the docstrings?

ymzayek:
I implemented a function in `conf.py` to remove the noqa comments. Seems to work fine. The other solution I see would be to not use noqa in docstrings and accept the PEP8 failures. Any other suggestions welcome!

nilearn/_utils/niimg.py
```diff
@@ -110,7 +110,7 @@ def load_niimg(niimg, dtype=None):
     -----------
 
     niimg: Niimg-like object
-        See https://nilearn.github.io/stable/manipulating_images/input_output.html  # noqa:E501
+        See https://nilearn.github.io/stable/manipulating_images/input_output.html.  # noqa: E501
```
htwangtw, line 113: I am not sure how it would effect rendering if there's a period at the end of URL

nilearn/image/image.py
```diff
@@ -296,7 +296,7 @@ def _crop_img_to(img, slices, copy=True):
     img : Niimg-like object
         Image to be cropped. If slices has less entries than `img` has dimensions,
         the slices will be applied to the first `len(slices)` dimensions (See
-        http://nilearn.github.io/manipulating_images/input_output.html).
+        https://nilearn.github.io/stable/manipulating_images/input_output.html)
```
htwangtw, line 299:
```suggestion
        https://nilearn.github.io/stable/manipulating_images/input_output.html).
```

htwangtw (commented):
I would prefer to see if it's possible to make E501 for URL an exception. That way we can avoid using `# noqa`. Open to other options that makes sense!

nilearn/_utils/niimg.py
ymzayek: I was trying to make things consistent, but yea it doesn't affect rendering.

ymzayek:
> I would prefer to see if it's possible to make E501 for URL an exception. That way we can avoid using `# noqa`. Open to other options that makes sense!

@htwangtw I haven't found a way to do that but I think the problem would be solved anyway by what is proposed in issue https://github.com/nilearn/nilearn/issues/3321 and additionally make the URLs more manageable so maybe I should work on that first. For non-rendered docstrings the noqa comments can stay as they are.

However, I wonder if it's still useful to have the `trim_noqa` function since it will make sure no noqa comments left in docstrings would ever be rendered. Though it might be that this never happens apart from the E501 for links but I'm not sure.

htwangtw:
> However, I wonder if it's still useful to have the `trim_noqa` function since it will make sure no noqa comments left in docstrings would ever be rendered. Though it might be that this never happens apart from the E501 for links but I'm not sure.

I am on the same page as you here. Looked into this briefly before and have a feeling that it's a really fussy issue. 

Resolving #3321 first sounds like a good idea to me.

bthirion:
This one is suprseded by #3332 ?

ymzayek:
@bthirion yes I don't think this is needed anymore. Closing

PR #3317, opened 2022-07-29
[MAINT] [DOC] Add suggestion for running flake8 on diff locally before pushing changes

### LINKED ISSUE #2528: Automate the Formatting stuff
Prompted by https://github.com/nilearn/nilearn/pull/2477#issuecomment-648905550

We have occasionally discussed this, but I think now is the time.

Nilearn contributors, managers, core developers juggle a lot of other responsibilities. We should automate the stuff that we can.

There are automated formatters & linters as well as cloud based code quality platforms that would automate a lot of minutia, making it a non-issue for contributors & reviewers. We can set up our custom for these tools & then we should start incorporating those.

### LINKED ISSUE #3308: Suggestions for contributors to run code quality checks locally
It might be a good idea to add some information about checking code quality locally in the contributing guidelines.
This relates to #2528 but it would just be optional suggestions to use e.g. pre-commit, black, etc. instead of adding dependencies directly to nilearn. At the very least we can suggest running flake8 on the diff locally and fixing PEP8 issues before committing and pushing code since that is already a dependency.

ymzayek:
Closes #3308.
Also close #2528 in my opinion since there is consensus not to automate formatting and linting.

This PR proposes additional information to the contributing guidelines about running flake8 on the diff locally before pushing changes. This is meant to reduce PEP8 check failures.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -264,19 +264,6 @@ This is also useful for writing unit tests.
 Writing small functions is not always possible, and we do not recommend trying to reorganize larger,
 but well-tested, older functions in the codebase, unless there is a strong reason to do so (e.g., when adding a new feature).
 
-.. admonition:: Recommendation
-
-    To lint your code and verify PEP8 compliance, you can run
-    `flake8 <https://flake8.pycqa.org/en/latest/>`_ locally on the
-    changes you have made in your branch compared to the main branch or a
-    previous commit.
-    To do this before committing changes, get the diff between the working
-    directory and the last commit and pipe it to flake8 by running:
-
-    .. code-block:: bash
-
-        git diff HEAD | flake8 --diff
-
 Tests
 ------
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -264,6 +264,19 @@ This is also useful for writing unit tests.
...
+    To lint your code and verify PEP8 compliance, you can run
+    `flake8 <https://flake8.pycqa.org/en/latest/>`_ locally on the
+    changes you have made in your branch compared to the main branch or a
+    previous commit.
+    To do this before committing changes, get the diff between the working
+    directory and the last commit and pipe it to flake8 by running:
+
+    .. code-block:: bash
+
+        git diff HEAD | flake8 --diff
```
jeromedockes, line 278: this will check the diff with HEAD, but we probably want the diff with the latest common ancestor with upstream/main, as is done here: https://github.com/nilearn/nilearn/blob/bd28810b2e6d97ef123d4c935c605ee384f0c40b/build_tools/flake8_diff.sh#L57

ymzayek: Ok I think something like this should work when the branch is checked out:

```bash
COMMIT=$(git merge-base main @) 
git diff $COMMIT | flake8 --diff
```
This way working directory changes as well as all committed changes in the branch are checked. WDYT?

jeromedockes: I think it looks good!

CONTRIBUTING.rst
```diff
@@ -363,21 +376,25 @@ Here are the key steps you need to go through to contribute code to `nilearn`:
 
       git checkout -b your_branch
 
-3. implement and commit your changes on this branch (don't forget to write tests!)
+3. implement changes and (optional but highly recommended) lint::
+
+      git diff HEAD | flake8 --diff
```
htwangtw, line 381: I believe this should also match the suggestion from @jeromedockes

jeromedockes: or even better show the linting command in one place only

jeromedockes (approved):
thanks!

htwangtw (approved):
LGTM :+1:

PR #3322, opened 2022-08-02
[ENH] Add DeprecationWarning for 3D masker.transform inputs

### LINKED ISSUE #2724: support 1D inputs with inverse_transform

### LINKED ISSUE #2726: should inverse_transform always return 4D output?
The controversy as I see it is balancing the philosophical choices of the maintainers versus how the code has actually been behaving over the last near decade (and how many people depend on that behavior).

> @tsalo, great point you brought up, for `inverse_transform`, if a 1D input is passed in then a 3D output will be produced, but looking at https://github.com/nilearn/nilearn/issues/460#issuecomment-100767188, I see that consistency of output was deemed more important, if that decision still holds, `inverse_transform` should always return 4D. Although, the functionality for unmask was [implemented before that conversation](https://github.com/nilearn/nilearn/pull/42).
\
I do not have a strong opinion on this (but someone should!), for maintainability I believe the 2D/4D output should be enforced, but on the other hand unmask has supported 1D inputs for years suggesting people may depend on that behavior of unmask and the transformers that call unmask.
\
I exploited the idiosyncrasy of `inverse_transform` for [NiMARE](https://github.com/neurostuff/NiMARE/blob/713a231527b3d4615a86161bcf174e0a52c96752/nimare/transforms.py#L123), but that behavior may be a bug and not a feature.

_Originally posted by @jdkent in https://github.com/nilearn/nilearn/pull/2725#issuecomment-794386153_

### LINKED ISSUE #3132: Inconsistent behaviour of `inverse_transform()` in masker classes
Under the hood, `nilearn.masker.NiftiMasker` uses `nilearn.masking.unmask` to inverse transform the signal to image, and for `nilearn.masker.NiftiLabelMasker` it uses `nilearn.regions.signal_extraction.signals_to_img_labels`. There are some inconsistency on how to handle signal transformed from one nifti image.

`nilearn.masking.unmask` treats 1-d array as one session / one nifti image (?) automatically:
https://github.com/nilearn/nilearn/blob/1607b52458c28953a87bbe6f42448b7b4e30a72f/nilearn/masking.py#L900-L910

`nilearn.regions.signal_extraction.signals_to_img_labels` forces users to pass a 2-d array, no matter what:
https://github.com/nilearn/nilearn/blob/39d5ee4a1067486b3d0d72c70352bf0b1ac182fa/nilearn/regions/signal_extraction.py#L136-L148

This results in the `inverse_transform()` method behaving slightly differently and is potentially confusing for user who has used both type of maskers. It might not be a problem for data that was all processed with nilearn. We only ran into this issue when trying to plot some output from a DL model. 

Would it be useful to make sure there's some consistency? It might worth looking into other masker objects?

cc'd @anirudhk686

tsalo:
Closes #2726, closes #3132, and closes #2724.

- Fix docstrings of masker classes' `inverse_transform` methods to make it clear that 1D inputs are allowed, and will produce a 3D image.
- Explicitly allow 1D inputs to `NiftiLabelsMasker.inverse_transform`. This is the only masker that did not already accept 1D inputs and produce 3D outputs.
- Raise a DeprecationWarning when providing a 3D image to masker classes' `transform` methods. At the moment, a 3D image will be transformed into a 2D array. Starting in version 0.12, it will be transformed into a 1D array.
- Add tests that cover these input and output shapes, and check for the DeprecationWarning when appropriate.

To do:

- [x] I still need to update `MultiNiftiMasker`.
    - Fortunately, I don't need to. `MultiNiftiMasker` uses the same underlying code for the transform, so it already raises the warning.
- [x] Add whatsnew entry.
- [x] Exhibit expected behavior in examples and documentation.


nilearn/maskers/base_masker.py
```diff
@@ -66,6 +66,19 @@ def _filter_and_extract(
...
+        warnings.warn(
+            'Starting in version 0.12, 3D images will be transformed to '
+            '1D arrays. '
+            'Until then, 3D images will be coerced to 2D arrays, with a '
+            'singleton first dimension representing time.',
+            DeprecationWarning,
+        )
+
     imgs = _utils.check_niimg(imgs, atleast_4d=True, ensure_ndim=4,
                               dtype=dtype)
```
tsalo, line 83: `imgs` may be a string (or list of strings, I guess) at this point. I therefore needed to look at the Nifti1Image object first, _before_ changing its dimensionality with `check_niimg(..., ensure_ndim=4)`. The best solution I could come up with was to call `check_niimg` twice.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -633,11 +637,19 @@ def inverse_transform(self, signals):
 
         Any mask given at initialization is taken into account.
 
+        .. versionchanged:: 0.9.2dev
+
+            This method now supports 1D arrays, which will produce 3D images.
```
tsalo, line 642: This should be the only actual behavior change in this PR.

bthirion (commented):
Maybe we can factor a bit more but overall that sounds right.

nilearn/maskers/tests/test_nifti_maps_masker.py
```diff
@@ -5,23 +5,36 @@
...
 from nilearn.image import get_data
+from nilearn.maskers import NiftiMapsMasker
 
 
-def generate_random_img(shape, length=1, affine=np.eye(4),
-                        rand_gen=np.random.RandomState(0)):
-    data = rand_gen.standard_normal(size=(shape + (length,)))
-    return nibabel.Nifti1Image(data, affine), nibabel.Nifti1Image(
-        as_ndarray(data[..., 0] > 0.2, dtype=np.int8), affine)
+def generate_random_img(
```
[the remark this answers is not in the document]
tsalo: I can move it into `data_gen`.

nilearn/maskers/tests/test_nifti_masker.py
```diff
@@ -404,7 +420,7 @@ def test_filter_and_mask():
     mask_img = nibabel.Nifti1Image(mask, np.eye(4))
 
     masker = NiftiMasker()
-    params = get_params(NiftiMasker, masker)
+    params = class_inspect.get_params(NiftiMasker, masker)
```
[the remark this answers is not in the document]
tsalo: I suppose it isn't, since extra stuff will be imported. I've switched that back.

ymzayek (commented):
@tsalo LGTM! I will take a bit more time to review the tests. Just quickly wondering why you changed the way `Nifti1Image` is imported in the tests. But I guess this is quite inconsistent throughout the library and I believe both ways are equally correct so it's not a big deal.

tsalo:
I changed imports in a few ways, so I'm not sure which specific change you're referring to. The main changes I recall making are (1) reordering imports based on `isort` and (2) consolidating redundant imports. For example, in some cases I've seen something like:

```python
import nibabel
...
from nibabel import Nifti1Image  # redundant with the overall nibabel import earlier
```

in which case I removed the redundant import. It could be that there are cases where I made changes that were less clean-cut (e.g., importing a module instead of many individual functions), and if you'd prefer to revert those please just flag them in your review.

ymzayek (approved):
Nothing else to add, thx @tsalo !

NicolasGensollen (commented):
Thanks @tsalo !
LGTM, just a small documentation suggestion.

nilearn/regions/signal_extraction.py
```diff
@@ -202,10 +202,12 @@ def signals_to_img_labels(signals, labels_img, mask_img=None,
 
     # nditer is not available in numpy 1.3: using multiple loops.
     # Using these loops still gives a much faster code (6x) than this one:
-    ## for n, label in enumerate(labels):
-    ##     data[labels_data == label, :] = signals[:, n]
-    data = np.zeros(target_shape + (signals.shape[0],),
-                    dtype=signals.dtype, order=order)
+    # for n, label in enumerate(labels):
```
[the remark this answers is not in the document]
tsalo: Thanks! Should be addressed in https://github.com/nilearn/nilearn/pull/3322/commits/9522593d97a6aca8b53acdeb21c0bd86f1aef5b4.

NicolasGensollen (approved):
LGTM, thanks @tsalo !

ymzayek:
Can we merge this one? The failure is unrelated and is now fixed

tsalo:
There are two approving reviews, so it should be okay to merge. @bthirion do you want to take another look before we merge?

nilearn/maskers/tests/test_nifti_masker.py
```diff
@@ -219,29 +218,31 @@ def test_4d_single_scan():
 
 
 def test_5d():
-    mask = np.zeros((10, 10, 10))
+    """Test that list of 4D images with last dim=1 are treated as 3d."""
```
bthirion, line 221: I think that this docstring is not accurate.

bthirion (commented):
There seems to be a minor glitch otherwise the PR LGTM.

nilearn/maskers/tests/test_nifti_masker.py
tsalo: Thanks for catching that. It should be fixed now.

bthirion (approved):
All right, thx !

ymzayek:
Thx all! Merging

PR #3327, opened 2022-08-05
[MAINT] Follow-up to pathlib support in #3300

ymzayek:
I realized after merging #3300 that I can improve some things and that I may have missed a couple of things.

- Expose `stringify_path` in `_utils.__init__`
- Missed one case where function should be used
- Docstrings were a bit inconsistent
- Failed to address https://github.com/nilearn/nilearn/pull/3300#discussion_r915643466

### DIFF
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -28,7 +28,7 @@ def _check_memory(memory, verbose=0):
 
     Parameters
     ----------
-    memory : None or instance of joblib.Memory or str or pathlib.Path
+    memory : None,instance of joblib.Memory, str or pathlib.Path
         Used to cache the masking process.
         If a str is given, it is the path to the caching directory.
 
@@ -181,7 +181,7 @@ def cache(func, memory, func_memory_level=None, memory_level=None,
     func : function
         The function which output is to be cached.
 
-    memory : instance of joblib.Memory or string or pathlib.Path
+    memory : instance of joblib.Memory, string or pathlib.Path
         Used to cache the function call.
 
     func_memory_level : int, optional
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -282,7 +282,7 @@ class FirstLevelModel(BaseGLM):
         This parameter is passed to nilearn.image.resample_img.
         Please see the related documentation for details.
     %(smoothing_fwhm)s
-    memory : string or os.PathLike, optional
+    memory : string or pathlib.Path, optional
         Path to the directory used to cache the masking process and the glm
         fit. By default, no caching is done.
         Creates instance of joblib.Memory.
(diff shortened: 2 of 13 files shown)


nilearn/_utils/cache_mixin.py
```diff
@@ -181,7 +181,7 @@ def cache(func, memory, func_memory_level=None, memory_level=None,
     func : function
         The function which output is to be cached.
 
-    memory : instance of joblib.Memory or string or pathlib.Path
+    memory : instance of joblib.Memory, string or pathlib.Path
```
bthirion, line 184:
```suggestion
    memory : instance of :class:`joblib.Memory`, obj:`str` or :class:`pathlib.Path`
```

nilearn/glm/first_level/first_level.py
```diff
@@ -282,7 +282,7 @@ class FirstLevelModel(BaseGLM):
         This parameter is passed to nilearn.image.resample_img.
         Please see the related documentation for details.
     %(smoothing_fwhm)s
-    memory : string or os.PathLike, optional
+    memory : string or pathlib.Path, optional
```
bthirion, line 285: same here etc.

nilearn/tests/test_niimg.py
```diff
@@ -16,21 +16,25 @@
 currdir = os.path.dirname(os.path.abspath(__file__))
 
 
+@pytest.fixture
+def img1():
```
jeromedockes, line 20: maybe we could put it in conftest.py or somewhere else where it would be available to other test modules? I think having a tiny image whose values and affine don't matter is a common requirements and gets redefined in several places. a slightly more explicit name might also help reusing it

ymzayek: Sure, I think that would be good too but would it be better done in another PR? I think it is worth a discussion first how we want to move forward refactoring and modularizing fixtures. For example, having all fixtures in one file and where, creating new fixtures for test files where we see this repetition, better naming for fixtures, etc. WDYT?

nilearn/_utils/cache_mixin.py
ymzayek: I think it may be better to leave as is since the rest of the file does not include these roles. Also, I guess since it's under `_utils` module, the docstrings are not exposed.

jeromedockes: indeed if these docstrings are read directly in the source files rather than in rendered html documentation there is no need for markup

nilearn/tests/test_niimg.py
jeromedockes: sure, that  makes more sense!

jeromedockes (approved):

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

ymzayek:
Thx for the reviews! Merging.

Issue #3266, opened 2022-05-24
Optimize TFCE calculation or replace with probabilistic TFCE


tsalo:
The TFCE method we implemented in #3196 is very slow. The probabilistic TFCE measure implemented in MNE seems to be faster, although I haven't directly compared them. We should look into either speeding up the current implementation or replacing it with a probabilistic TFCE measure.

### Benefits to the change
Faster TFCE-based multiple comparisons correction.

### Pseudocode for the new behavior, if applicable

The `permuted_ols` and `non_parametric_inference` would be the same.

bthirion:
I don't know about it, but I agree that it would be great.

sahahn:
I'm not familiar with the probabilistic implementation of TFCE, but I spent some time today looking at the existing implementation and I think I see some areas which I could help to speed up. I can try a few different things, and if they help then I'll submit a PR ~

ymzayek:
That sounds great, thx @sahahn !

bthirion:
Let's close this one, then. Thx !

PR #3333, opened 2022-08-09
[REF] Performance optimizations for TFCE

sahahn:
This PR is designed as a first pass at solving performance issues as noted in [#3266](https://github.com/nilearn/nilearn/issues/3266).

This PR is rather minimal in scope in that it only changes the internal content of the '_calculate_tfce' method and produces identical results to the previous version of the code. The change is of course that the code is now a bit faster. Running the updated code on the same benchmark as mentioned in https://github.com/nilearn/nilearn/pull/3196 the 'plot_localizer_mass_univariate_methods' now runs in ~7 minutes on a single core, whereas the old version of the code ran in ~30 minutes on my machine with the same settings. 

Notably with the example above and to some extent the current code, the runtime is very dependent on the shape of the original images. For example, the task contrast maps have shape (53, 63, 46) which is a bit borderline on where the existing code starts to get very slow (some of my tests suggest that it gets about 40x slower jumping up to just sizes 75x75x75).

I have a more in depth notebook with a breakdown on each of the individual changes / some thoughts on areas we still might be able to improve which I will post soon, but I just thought I would put up the actual changes here first.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -41,6 +41,8 @@ Enhancements
 - Add ``n_elements_`` attribute to masker classes (:gh:`3311` by `Taylor Salo`_).
 - Functions expecting string filesystem paths now also accept path-like objects (:gh:`3300` by `Yasmin Mzayek`_).
 - Contributing guidelines now include a recommendation to run flake8 locally on the branch diff with main (:gh:`3317` by `Yasmin Mzayek`_).
+- Improvements to :func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` with :term:`TFCE` statistic runtime (:gh:`3333` by `Sage Hahn_`).
+
 
 Changes
 -------
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -41,8 +41,7 @@ Enhancements
 - Add ``n_elements_`` attribute to masker classes (:gh:`3311` by `Taylor Salo`_).
 - Functions expecting string filesystem paths now also accept path-like objects (:gh:`3300` by `Yasmin Mzayek`_).
 - Contributing guidelines now include a recommendation to run flake8 locally on the branch diff with main (:gh:`3317` by `Yasmin Mzayek`_).
-- Improvements to :func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` with :term:`TFCE` statistic runtime (:gh:`3333` by `Sage Hahn_`).
-
+- Improvements to :func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` with :term:`TFCE` statistic runtime (:gh:`3333` by `Sage Hahn`_).
 
 Changes
 -------
--- a/nilearn/mass_univariate/_utils.py
+++ b/nilearn/mass_univariate/_utils.py
@@ -119,7 +119,7 @@ def _calculate_tfce(
                 cluster_counts = np.bincount(labeled_non_zero)
 
                 # Next, we convert each unique cluster count to its TFCE value.
-                # Where each voxel's tfce value is based
+                # Where each cluster's tfce value is based
                 # on both its cluster extent and z-value (via the current score_thresh)
                 # NOTE: We do not multiply by dh, based on fslmaths'
                 # implementation. This differs from the original paper.
--- a/nilearn/mass_univariate/_utils.py
+++ b/nilearn/mass_univariate/_utils.py
@@ -99,8 +99,10 @@ def _calculate_tfce(
                 temp_arr3d[temp_arr3d < score_thresh] = 0
 
                 # Label into clusters - importantly (for the next step)
-                # this returns clusters labelled ordinally from 1 to n_clusters+1,
-                # which allows us to use bincount to count frequencies directly.
+                # this returns clusters labelled ordinally
+                # from 1 to n_clusters+1,
+                # which allows us to use bincount to count
+                # frequencies directly.
                 labeled_arr3d, _ = label(temp_arr3d, bin_struct)
 
                 # Next, we want to replace each label with its cluster
@@ -112,7 +114,7 @@ def _calculate_tfce(
                 labeled_non_zero = labeled_arr3d_flat[non_zero_inds]
 
                 # Count the size of each unique cluster, via its label.
-	        # The reason why we pass only the non-zero labels to bincount
+                # The reason why we pass only the non-zero labels to bincount
                 # is because it includes a bin for zeros, and in our labels
                 # zero represents the background,
                 # which we want to have a TFCE value of 0.
@@ -120,16 +122,19 @@ def _calculate_tfce(
 
                 # Next, we convert each unique cluster count to its TFCE value.
                 # Where each cluster's tfce value is based
-                # on both its cluster extent and z-value (via the current score_thresh)
+                # on both its cluster extent and z-value
+                # (via the current score_thresh)
                 # NOTE: We do not multiply by dh, based on fslmaths'
                 # implementation. This differs from the original paper.
-                cluster_tfces = sign * (cluster_counts ** E) * (score_thresh ** H)
+                cluster_tfces = sign * (cluster_counts ** E) *\
+                    (score_thresh ** H)
 
                 # Before we can add these values to tfce_4d, we need to
                 # map cluster-wise tfce values back to a voxel-wise array,
                 # including any zero / background voxels.
                 tfce_step_values = np.zeros(labeled_arr3d_flat.shape)
-                tfce_step_values[non_zero_inds] = cluster_tfces[labeled_non_zero]
+                tfce_step_values[non_zero_inds] =\
+                    cluster_tfces[labeled_non_zero]
 
                 # Now, we just need to reshape these values back to 3D
                 # and they can be incremented to tfce_4d.
--- a/nilearn/mass_univariate/_utils.py
+++ b/nilearn/mass_univariate/_utils.py
@@ -88,45 +88,53 @@ def _calculate_tfce(
         # If we apply the sign first...
         for sign in signs:
 
-            # Then init a temp array, we can re-use the
-            # same array and save a little bit of time
+            # Init a temp array, which can re-use
+            # by incrementally setting more background
+            # voxel's to background, given that each score
+            # thresh applied is incrementally larger than the one before.
             temp_arr3d = arr3d.copy() * sign
 
             # Prep step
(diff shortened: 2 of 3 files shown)


sahahn:
Notebook with thoughts / comments on the optimization:
https://gist.github.com/sahahn/e44c8778b48221f73b3583ca4583a642

Again noting, that we can probably still do better, and that even the current implementation still scales very poorly as the input size gets large

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,68 @@ def _calculate_tfce(
...
-                temp_arr3d = arr3d * sign
+        # Set based on determined step size
+        score_threshs = np.arange(step, max_score + step, step)
+
+        # If we apply the sign first...
+        for sign in signs:
+
+            # Then init a temp array, we can re-use the
+            # same array and save a little bit of time
+            temp_arr3d = arr3d.copy() * sign
```
ymzayek, line 93: Why do you add the `copy` method here?

ymzayek (commented):
Hi @sahahn, seems like a good improvement already. And thanks for the notebook, I'm not very familiar with the implementation so it really helped and I appreciate the sanity check at the end. Can you also add your name to `doc/changes/names.rst` and a whatsnew entry in `doc/changes/latest.rst`?

nilearn/mass_univariate/_utils.py
sahahn: Just to make it a little more clear that we are working with a copy of arr3d. Though as you point out it is not strictly necessary as "temp_arr3d = arr3d * sign" would also create a copy

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,68 @@ def _calculate_tfce(
...
+                labeled_arr3d, _ = label(temp_arr3d, bin_struct)
+
+                # Get flattened version of the array
+                labeled_arr3d_flat = labeled_arr3d.flatten()
+
+                # And keep track of non-zero inds
+                non_zero_inds = np.where(labeled_arr3d_flat != 0)[0]
+
+                # Get unique values / counts to compute frequencies
+                idx = labeled_arr3d_flat[non_zero_inds]
```
tsalo, line 111: I think these variables could have more informative names. Perhaps `cluster_values` instead of `idx` and `cluster_counts` instead of `freqs_idx`.

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,68 @@ def _calculate_tfce(
...
+                # Easier / slightly faster to compute if we apply
+                # the E / H / sign computations
+                # here on each unique value
+                # Where are calculating each voxel's tfce value based
+                # on its cluster extent and z-value
                 # NOTE: We do not multiply by dh, based on fslmaths'
                 # implementation. This differs from the original paper.
-                tfce_step_values = (cluster_map**E) * (score_thresh**H)
-                tfce_4d[..., i_regressor] += sign * tfce_step_values
+                adj_freqs_idx = sign * (freqs_idx ** E) * (score_thresh ** H)
```
tsalo, line 121: Why not `cluster_tfces` or something similar?

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,68 @@ def _calculate_tfce(
...
+
+                # Gen and fill flat tfce_step_values w/ adjusted freqs directly
+                tfce_step_values = np.zeros(labeled_arr3d_flat.shape)
+                tfce_step_values[non_zero_inds] = adj_freqs_idx[idx]
+
+                # Add the values to the tfce array in the correct spots
+                tfce_4d[..., i_regressor] +=\
+                    tfce_step_values.reshape(temp_arr3d.shape)
```
tsalo, line 129:
```suggestion
                tfce_4d[..., i_regressor] += tfce_step_values.reshape(
                    temp_arr3d.shape,
                )
```

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,68 @@ def _calculate_tfce(
...
+
+                # Get flattened version of the array
+                labeled_arr3d_flat = labeled_arr3d.flatten()
+
+                # And keep track of non-zero inds
+                non_zero_inds = np.where(labeled_arr3d_flat != 0)[0]
+
+                # Get unique values / counts to compute frequencies
+                idx = labeled_arr3d_flat[non_zero_inds]
+                freqs_idx = np.bincount(idx)
```
tsalo, line 112: The thing that initially confused me is that `bincount` also includes a bin for zeros, which works out because `idx` starts with one for the first cluster. I think it would be helpful to have a comment here to that effect.

```suggestion
                # bincount includes a bin for zeros, but this bin is skipped 
                # automatically later
                freqs_idx = np.bincount(idx)
```

tsalo (changes requested):
I love the changes, but I have a few minor requests regarding variable names and comments. Thanks!

nilearn/mass_univariate/_utils.py
tsalo: I think the copy will slow things down very slightly, and since it's not necessary I think it would be good to drop it.

nilearn/mass_univariate/_utils.py
sahahn: Thanks! I've made your suggestions and added some extra clarity to some other comments also~

nilearn/mass_univariate/_utils.py
sahahn: Sure no worries, I just feel like sometimes the behavior isn't super clear, e.g., "temp_arr3d = arr3d" alone just sets temp_arr3d as a reference to the same array, but I'm always down to chase the microsecond gains.

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,76 @@ def _calculate_tfce(
...
-                    cluster_map[bool_map] = np.sum(bool_map)
-
-                # Calculate each voxel's tfce value based on its cluster extent
-                # and z-value
+                # Label into clusters - importantly (for the next step)
+                # this returns clusters labelled ordinally from 1 to n_clusters+1,
+                # which allows us to use bincount to count frequencies directly.
+                labeled_arr3d, _ = label(temp_arr3d, bin_struct)
+
+                # Next, we want to replace each label the it's cluster
```
tsalo, line 106:
```suggestion
                # Next, we want to replace each label with its cluster
```

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,76 @@ def _calculate_tfce(
...
+                # extent, that is, the size of the cluster it is part of
+                # To do this, we will first compute a flattened version of the
```
tsalo, line 108:
```suggestion
                # To do this, we will first compute a flattened version of
```

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,76 @@ def _calculate_tfce(
...
+                # only the non-zero cluster labels.
+                labeled_arr3d_flat = labeled_arr3d.flatten()
+                non_zero_inds = np.where(labeled_arr3d_flat != 0)[0]
+                labeled_non_zero = labeled_arr3d_flat[non_zero_inds]
+
+                # Count the size of each unique cluster, via its label.
+	            # The reason why we pass only the non-zero labels to bincount
```
tsalo, line 115:
```suggestion
	        # The reason why we pass only the non-zero labels to bincount
```

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,76 @@ def _calculate_tfce(
...
+                # is because it includes a bin for zeros, and in our labels
+                # zero represents the background,
+                # which we want to have a TFCE value of 0.
+                cluster_counts = np.bincount(labeled_non_zero)
+
+                # Next, we convert each unique cluster count to its TFCE value.
+                # Where each voxel's tfce value is based
```
tsalo, line 122:
```suggestion
                # Where each cluster's tfce value is based
```

nilearn/mass_univariate/_utils.py
```diff
@@ -65,50 +65,76 @@ def _calculate_tfce(
...
+                # on both its cluster extent and z-value (via the current score_thresh)
                 # NOTE: We do not multiply by dh, based on fslmaths'
                 # implementation. This differs from the original paper.
-                tfce_step_values = (cluster_map**E) * (score_thresh**H)
-                tfce_4d[..., i_regressor] += sign * tfce_step_values
+                cluster_tfces = sign * (cluster_counts ** E) * (score_thresh ** H)
+
+                # Before we can add these values to tfce_4d, we need to
+                # map from each voxel's cluster label to that cluster's tfce value,
+                # but also keep account for adding back in any zero / background labels.
```
tsalo, line 130:
```suggestion
                # Before we can add these values to tfce_4d, we need to
                # map cluster-wise tfce values back to a voxel-wise array,
                # including any zero / background voxels.
```

tsalo (changes requested):
Thanks for this! Just a few small wording requests.

sahahn:
> 
Was having some issues with my vscode environment and flake8, but I think it should be good  now

tsalo (approved):
PEP8 is happy now. Thanks!

tsalo:
I think you can replace the [WIP] prefix with an [REF] (refactor) one now.

doc/changes/latest.rst
```diff
@@ -41,6 +41,8 @@ Enhancements
 - Add ``n_elements_`` attribute to masker classes (:gh:`3311` by `Taylor Salo`_).
 - Functions expecting string filesystem paths now also accept path-like objects (:gh:`3300` by `Yasmin Mzayek`_).
 - Contributing guidelines now include a recommendation to run flake8 locally on the branch diff with main (:gh:`3317` by `Yasmin Mzayek`_).
+- Improvements to :func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` with :term:`TFCE` statistic runtime (:gh:`3333` by `Sage Hahn_`).
+
```
ymzayek, line 45: In order to correctly link to your website the syntax needs to be as follows:

```suggestion
- Improvements to :func:`~mass_univariate.permuted_ols` and :func:`~glm.second_level.non_parametric_inference` with :term:`TFCE` statistic runtime (:gh:`3333` by `Sage Hahn`_).
```

ymzayek (commented):
Thx @sahahn ! Just one small thing (see below) and it should be ready to merge

ymzayek (approved):

ymzayek:
Merging now, thx !

PR #3344, opened 2022-08-22
[MAINT] Fix Matplotlib Pending deprecations

NicolasGensollen:
See https://matplotlib.org/devdocs/api/next_api_changes/deprecations/23668-TC.html

This PR should fix the test with Python 3.10 and pre-release deps.

PR #3345, opened 2022-08-22
WIP: Deal with mpl deprecation of get_cmap

larsoner:
Matplotlib is deprecating get_cmap. Let me know if this is on the right track for you all and I'll:

1. In cases where `cmap` is clearly `str`, use `colormaps[cmap`]
2. In cases where it's not clear (e.g., could be a `Colormap` already) replace `cm.get_cmap`s with a new `_get_cmap` from `plotting/cm.py`

### DIFF
--- a/examples/01_plotting/plot_colormaps.py
+++ b/examples/01_plotting/plot_colormaps.py
@@ -44,7 +44,7 @@ m_cmaps.sort()
 
 for index, cmap in enumerate(m_cmaps):
     plt.subplot(1, len(m_cmaps) + 1, index + 1)
-    plt.imshow(a, cmap=plt.get_cmap(cmap), aspect='auto')
+    plt.imshow(a, cmap=plt.colormaps[cmap], aspect='auto')
     plt.axis('off')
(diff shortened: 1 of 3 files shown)


ymzayek:
Hi @larsoner thanks for this though there is already a pr open to deal with this but we can discuss which way we want to move forward:

@NicolasGensollen I'm linking this pr to yours #3344 because IIUC it is resolving the same thing in a different way.
According to [matplotlib.org/devdocs/api/next_api_changes/deprecations/23668-TC.html](https://matplotlib.org/devdocs/api/next_api_changes/deprecations/23668-TC.html):
- `matplotlib.pyplot.get_cmap` can still be used which is what the linked pr is using
- `matplotlib.colormaps[name]` is replacing `matplotlib.cm.get_cmap` which is what is implemented here

WDYT is best to deal with the deprecation?

larsoner:
Sorry I should have checked to see if there was a PR working on this already -- `matplotlib.pyplot.get_cmap` seems fine to me! And from trying to work around with `_get_cmap` I think I'm implementing a bunch of stuff that already exists in that function anyway.

PR #3371, opened 2022-09-27
Add banner to dev docs

### LINKED ISSUE #3367: Add back stable/dev banners to website
The green `stable` and red `dev` banners were not transferred over to the new documentation theme. 

To do:
- add banners back
- (optional) add a toggle to easily go back and forth between versions

alexisthual:
Closes #3367.

- add banner to dev builds of the documentation (on all pages), with a link to the stable version
- it should display well on small devices too

Screenshot from 2022-09-27 14-01-12


doc/conf.py
```diff
@@ -179,12 +184,28 @@
...
 
+# Add banner in case version is not stable
+if "dev" in release:
+    html_theme_options["announcement"] = (
+        f'<p>This is the development documentation of nilearn ({release}) '
+        '<a class="sd-sphinx-override sd-badge sd-text-wrap '
+        'sd-btn-outline-dark reference external" '
+        'href="https://nilearn.github.io">'
+        f'<span>Switch to stable version ({version})</span></a></p>'
+    )
```
ymzayek, line 207: **A note just so we don't forget (can be done in another PR)**

Since it might be annoying to users to not be able to switch back to dev from stable we can either:

- Add an `else` statement to define the same thing but inverted for the stable docs
- You also mentioned the possibility of a drop down menu to pick from all versions (would require building previous doc versions and deploying to nilearn.github.io)

In any case, changes to stable wouldn't show up until the next release so we have time to think about it.

doc/conf.py
```diff
@@ -101,10 +102,14 @@
...
+
+# Latest release version
+version = re.match(
+    r'v?([0-9]+.[0-9]+.[0-9]+).*',
+    os.popen('git describe --tags').read().strip()
+).groups()[0]
+
+# The full current version, including alpha/beta/rc tags.
 import nilearn
 release = nilearn.__version__
```
ymzayek, line 114: I think during a release, `version` and `release` will be the same. We need to be careful in how we are defining and using these and maybe also use better naming.

ymzayek (commented):
@alexisthual thanks for working on this! I need to think a bit more particularly about the first comment. But both can be handled in another PR if we want to merge this soon. What do others think?

doc/conf.py
alexisthual: Also, users could expect this button to take them to the equivalent page they're currently looking at in the stable docs. In this PR, the button simply takes you to the homepage of the stable docs.

doc/conf.py
ymzayek: Yea I agree with the inversion and also why not just be very explicit. 

It's a problem if we do something like the first point I propose in the previous comment:
> - Add an else statement to define the same thing but inverted for the stable docs

because during the release process the banner for the stable docs would be created with a reference to the same version number for both `dev` and `stable`. But I think we can solve that in an automated way with some additional code.

doc/conf.py
alexisthual: Also, just as a note, maybe this repo could help us deal with multiple versions: https://github.com/Holzhaus/sphinx-multiversion

emdupre: > You also mentioned the possibility of a drop down menu to pick from all versions (would require building previous doc versions and deploying to nilearn.github.io)

Personally I'd lean against this, since I think it will require storing a large number of (largely duplicate) artifacts. But if others feel strongly on this, it'd be nice to unpack the benefits a bit more !

doc/conf.py
```diff
@@ -179,12 +184,29 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
+    "dark_css_variables": {
+        "color-announcement-background": "#935610",
```
emdupre, line 188: This is returning a red for me (and the `light_css_variables` equivalent a green), but in the example screenshot you include it's a light orange. Could you confirm where these colors are shown ?

alexisthual: Humm that's weird haha. I picked the color from [blueprintjs](https://blueprintjs.com/docs/#core/colors) and just double-checked, everything should be alright. 
It seems ok in too ; does yours say otherwise?
Screenshot from 2022-09-27 22-42-06

emdupre: Ah, you're totally right ! Somehow my hex codes were getting converted inappropriately. Sorry !!!

Maybe one idea : Could we replace #935610 with (a shade near) #3499cd to match the nilearn logo ? So "blue" is stable and "orange" is unstable ?

ymzayek (approved):
I think it's ok to merge this now because this is just for the unstable version and users can at least be warned that they are viewing these pages. We can decide on a banner for the stable docs later if we want to have it like we did before the new theme was launched. But I think the option of storing multiple versions of the docs warrants a discussion as well.

doc/conf.py
emdupre: Ah yes I didn't mean exactly that code, but I think something near it would work well -- and it's perhaps a bit clearer than brown that it's stable, since there's more contrast than between brown and orange.

Small details, though, so feel free to overrule.

emdupre (approved):
Adding an official approval -- thank you, @alexisthual !

doc/conf.py
alexisthual: Oh maybe I didn't understand your comment then: I don't think we need to tweak the color for stable version because this banner will only be displayed if you are not looking at the stable version.

bthirion (approved):
Thx ! LGTM.

bthirion:
Merging, then.

doc/conf.py
emdupre: Sorry for being unclear ! I was preempting a bit : My understanding was that we were planning to add back in a stable banner, but this was likely a misunderstanding on my part ! I don't (personally) think it's necessary. 
If in fact we were going to add it back, I thought this CSS was intended for that. Sorry for the noise, though, and thanks again for this change !!

PR #3025, opened 2021-10-18
Bugfix: remove spaces from regressor names to prevent SyntaxError when computing contrasts

alexisthual:
Closes #3024.

Changes implemented in this pull request:

- replace spaces with underscores in `glm.first_level.hemodynamic_models._regressor_names()`
- update related tests

PR #3374, opened 2022-10-03
[FIX] Allow invalid identifiers for design matrix column names

### LINKED ISSUE #3323: design matrix columns name checking is excessive
Currently, a string can only a design matrix column if it fits the  `isidentifier()` condition, which looks excessive.
For instance, strings started with a number are not accepted, while there is no problem to create dataframes with such strings.
Thus I suggest to remove the lines below from the codebase

https://github.com/nilearn/nilearn/blob/2fd96b86cb53304f2f0b0c978d0bcd670f3795d2/nilearn/glm/first_level/hemodynamic_models.py#L443
```
# Check that all names look like proper pandas.DataFrame column names
    if not all([name.isidentifier() for name in names]):
        raise ValueError("At least one regressor name can't be used "
                         f"as a column identifier: {names}")
```
WDYT ?

ymzayek:
Closes #3323.

- Allow design matrix column name to be invalid identifier

### DIFF
--- a/nilearn/glm/contrasts.py
+++ b/nilearn/glm/contrasts.py
@@ -42,7 +42,7 @@ def expression_to_contrast_vector(expression, design_columns):
     df = pd.DataFrame(np.eye(len(design_columns)), columns=design_columns)
     try:
         contrast_vector = df.eval(expression, engine="python").values
-    except AttributeError:
+    except Exception:
         raise ValueError(
             'The expression (%s) is not valid. This could be due to '
             'defining the contrasts using design matrix columns that are '
--- a/nilearn/glm/contrasts.py
+++ b/nilearn/glm/contrasts.py
@@ -43,7 +43,7 @@ def expression_to_contrast_vector(expression, design_columns):
     try:
         contrast_vector = df.eval(expression, engine="python").values
     except AttributeError:
-        raise SyntaxError(
+        raise ValueError(
             'The expression (%s) is not valid. This could be due to '
             'defining the contrasts using design matrix columns that are '
             'invalid python identifiers. '
--- a/nilearn/glm/contrasts.py
+++ b/nilearn/glm/contrasts.py
@@ -40,7 +40,15 @@ def expression_to_contrast_vector(expression, design_columns):
         contrast_vector[list(design_columns).index(expression)] = 1.
         return contrast_vector
     df = pd.DataFrame(np.eye(len(design_columns)), columns=design_columns)
-    contrast_vector = df.eval(expression, engine="python").values
+    try:
+        contrast_vector = df.eval(expression, engine="python").values
+    except AttributeError:
+        raise SyntaxError(
+            'The expression (%s) is not valid. This could be due to '
+            'defining the contrasts using design matrix columns that are '
+            'invalid python identifiers. '
+            % expression
+        )
     return contrast_vector
 
 
--- a/nilearn/glm/first_level/hemodynamic_models.py
+++ b/nilearn/glm/first_level/hemodynamic_models.py
@@ -440,11 +440,6 @@ def _regressor_names(con_name, hrf_model, fir_delays=None):
     # Remove any non-word character
     names = [re.sub("[^a-zA-Z0-9_]+", "", str(name)) for name in names]
 
-    # Check that all names look like proper pandas.DataFrame column names
-    if not all([name.isidentifier() for name in names]):
-        raise ValueError("At least one regressor name can't be used "
-                         f"as a column identifier: {names}")
-
     return names
 
 
--- a/nilearn/glm/first_level/hemodynamic_models.py
+++ b/nilearn/glm/first_level/hemodynamic_models.py
@@ -435,11 +435,6 @@ def _regressor_names(con_name, hrf_model, fir_delays=None):
     if len(np.unique(names)) != len(names):
         raise ValueError(f"Computed regressor names are not unique: {names}")
 
-    # Replace spaces with underscores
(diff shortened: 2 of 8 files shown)


ymzayek:
After looking at the (failing) test previously written for testing for valid identifiers, I am not sure if I understand clearly the impact of allowing condition names to be invalid identifiers. See #3025. Will this cause issues when computing contrasts? @alexisthual

bthirion:
No, I've been using invalid identifiers for a while without hitting issues. Since these are *strings* anyhow, this should be fine.

ymzayek:
@bthirion it's probably fine for most use cases but it would fail in the case shown here:

```python
import numpy as np
import pandas as pd

def basic_paradigm():
    conditions = ['0', '0', '0',
                  '1', '1', '1',
                  '2', '2', '2']
    onsets = [30, 70, 100, 10, 30, 90, 30, 40, 60]
    durations = 1 * np.ones(9)
    events = pd.DataFrame({'trial_type': conditions,
                           'onset': onsets,
                           'duration': durations})
    return events

shapes, rk = [(10, 10, 10, 25)], 3
mask, fmri_data, _ = generate_fake_fmri_data_and_design(shapes, rk)
model = FirstLevelModel(t_r=2.0, mask_img=mask, hrf_model='spm')
events = basic_paradigm()
model.fit(fmri_data, events)
columns = model.design_matrices_[0].columns
model.compute_contrast(f"{columns[0]}-{columns[1]}")
```

```
  File "/home/yasmin/miniconda3/envs/nilearn-dev/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3369, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-4984deb2c750>", line 22, in <cell line: 22>
    model.compute_contrast(f"{columns[0]}-{columns[1]}")
  File "/home/yasmin/nilearn/nilearn/nilearn/glm/first_level/first_level.py", line 708, in compute_contrast
    con_vals[cidx] = expression_to_contrast_vector(
  File "/home/yasmin/nilearn/nilearn/nilearn/glm/contrasts.py", line 43, in expression_to_contrast_vector
    contrast_vector = df.eval(expression, engine="python").values
AttributeError: 'int' object has no attribute 'values'
```

However, I'm not sure if this is a relevant use case. WDYT?

bthirion:
Hmmm. Now I see the reason for this : symbolic contrast specification requires the matrix column names to be valid identifiers... and I don't see any way around this (?) 
Thx for raising the point.

ymzayek:
We could automatically append a prefix as mentiond by @jeromedockes in the linked issue:

> but if we do want to go for transforming it in a valid identifier we could prefix it with some arbitrary string that starts with a letter, eg reg_1_my_name.

bthirion:
Maybe simply prepending an undersore (and spitting a warning, of course).

jeromedockes:
I think the design matrix should still be allowed to have arbitrary
column names, IMO the issue you point out is a problem with string
contrast specification rather than the column names.

The contrast specification in the example above becomes "0-1" after the
f-string interpolation; if the documentation makes it clear that it
should be an expression that is evaluated after binding the design
matrix column names it seems obvious that it is not going to do what we
want.

Note that prepending a "_" to columns that start with a number will not
be enough for contrast specifications to always be valid python
expressions -- for example the involved column names must also not be
python keywords -- so we probably don't want to go through the trouble
of transforming column names to the point that specifying contrasts as a
string will always work (and at the same time surprise the user with
design matrix column names that are not what they asked for).

IMO what the example shows is just that allowing to pass strings and
evaluating them as code is a bit of an anti-feature; in many cases it
will be less error-prone, simpler and safer to just build the contrast
vector with normal pandas series or numpy array assignment.

I would suggest allowing arbitrary column names as was decided in the
issue, and editing the contrast building docstring to say that string
contrasts must be valid expressions, which implies that column names
must be valid python identifiers, and possibly have a better error
message when building the contrast vector fails.

bthirion:
The you would enforce valid identifiers at the contrast building stage. Makes sense to me.

ymzayek:
@jeromedockes I see your point. I will make the changes in the docstring and add an error message.

bthirion (approved):
Thx, this is great.

nilearn/glm/contrasts.py
```diff
@@ -40,7 +40,15 @@ def expression_to_contrast_vector(expression, design_columns):
         contrast_vector[list(design_columns).index(expression)] = 1.
         return contrast_vector
     df = pd.DataFrame(np.eye(len(design_columns)), columns=design_columns)
-    contrast_vector = df.eval(expression, engine="python").values
+    try:
+        contrast_vector = df.eval(expression, engine="python").values
+    except AttributeError:
```
jeromedockes, line 45: depending on the column names and exact contrast the error might be something else -- for example if a column name is a keyword it will be a SyntaxError. you can add SyntaxError to the list or maybe catch a broader type of exception

nilearn/glm/first_level/hemodynamic_models.py
```diff
@@ -440,11 +440,6 @@ def _regressor_names(con_name, hrf_model, fir_delays=None):
     # Remove any non-word character
```
jeromedockes, line 440: shouldn't the transformations above and below here be removed as well?

ymzayek:
@jeromedockes can you have another look here when you get the chance? Thanks!

jeromedockes (approved):
It looks great @ymzayek ! thanks :)

bthirion:
Merging, then. Thx @ymzayek !

PR #3375, opened 2022-10-03
[FIX] Remove `_safe_cache` function that is outdated and error-prone

### LINKED ISSUE #3256: `TypeError: expected string or bytes-like object` when using `nilearn.input_data.NiftiMasker` with `memory` parameter.
I get the following error from a downstream test function which used to work: https://ppb.chymera.eu/cc4f38.log

It gets fixed by changing:

```python
my_memory = Memory(path.abspath(path.expanduser('~/.nilearn_cache')))
masker = NiftiMasker(mask_img=roi, target_affine=ts_img.affine, memory=my_memory, memory_level=1)
```

to:

```python
masker = NiftiMasker(mask_img=roi, target_affine=ts_img.affine, memory_level=1)
```

Given that it seems to fail with a version comparison I suspect this is a fairly obvious issue with the installation? Any idea how to fix it, though?

ymzayek:
Closes #3256.

- `_safe_cache` now handles if versions are different types by continuing to flush the cache as if the versions don't match

### DIFF
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -85,12 +85,6 @@ def _safe_cache(memory, func, **kwargs):
     number of nibabel has changed.
 
     """
-    ''' Workaround for
-     https://github.com/scikit-learn-contrib/imbalanced-learn/issues/482
-    joblib throws a spurious warning with newer scikit-learn.
-    This code uses the recommended method first and the deprecated one
-    if that fails, ensuring th warning is not generated in any case.
-    '''
     try:
         location = os.path.join(memory.location, 'joblib')
     except AttributeError:
@@ -108,9 +102,8 @@ def _safe_cache(memory, func, **kwargs):
         with open(version_file, 'r') as _version_file:
             versions = json.load(_version_file)
 
-    modules = (nibabel, )
     # Keep only the major + minor version numbers
-    my_versions = dict((m.__name__, m.__version__) for m in modules)
+    my_versions = {nibabel.__name__: nibabel.__version__}
     commons = set(versions.keys()).intersection(set(my_versions.keys()))
     try:
         collisions = [
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -141,7 +141,7 @@ def _safe_cache(memory, func, **kwargs):
         else:
             warnings.warn("Incompatible cache in %s: "
                           "old version of nibabel." % location)
-
+            return memory.cache(func, **kwargs)
     # Write json files if configuration is different
     if versions != my_versions:
         with open(version_file, 'w') as _version_file:
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -112,11 +112,14 @@ def _safe_cache(memory, func, **kwargs):
     # Keep only the major + minor version numbers
     my_versions = dict((m.__name__, m.__version__) for m in modules)
     commons = set(versions.keys()).intersection(set(my_versions.keys()))
-    collisions = [
-        m for m in commons if not _compare_version(
-            versions[m], '==', my_versions[m]
-        )
-    ]
+    try:
+        collisions = [
+            m for m in commons if not _compare_version(
+                versions[m], '==', my_versions[m]
+            )
+        ]
+    except TypeError:
+        collisions = ['']
     # Flush cache if version collision
     if len(collisions) > 0:
         if nilearn.CHECK_CACHE_VERSION:
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -80,6 +80,13 @@ def _check_memory(memory, verbose=0):
     return memory
 
 
+def _nibabel_versions_equal(versions, my_versions):
+    try:
+        return _compare_version(versions, '==', my_versions)
+    except TypeError:
+        return False
+
+
 def _safe_cache(memory, func, **kwargs):
     """A wrapper for mem.cache that flushes the cache if the version
     number of nibabel has changed.
@@ -104,17 +111,10 @@ def _safe_cache(memory, func, **kwargs):
 
     # Keep only the major + minor version numbers
     my_versions = {nibabel.__name__: nibabel.__version__}
-    commons = set(versions.keys()).intersection(set(my_versions.keys()))
-    try:
-        collisions = [
-            m for m in commons if not _compare_version(
-                versions[m], '==', my_versions[m]
-            )
-        ]
-    except TypeError:
-        collisions = ['nibabel']
-    # Flush cache if version collision
-    if len(collisions) > 0:
+    if 'nibabel' in versions and not _nibabel_versions_equal(
+            versions['nibabel'], my_versions['nibabel']
+    ):
+        # Flush cache if version collision
         if nilearn.CHECK_CACHE_VERSION:
             warnings.warn("Incompatible cache in %s: "
                           "different version of nibabel. Deleting "
@@ -142,6 +142,7 @@ def _safe_cache(memory, func, **kwargs):
             warnings.warn("Incompatible cache in %s: "
                           "old version of nibabel." % location)
             return memory.cache(func, **kwargs)
+
     # Write json files if configuration is different
     if versions != my_versions:
         with open(version_file, 'w') as _version_file:
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -112,7 +112,7 @@ def _safe_cache(memory, func, **kwargs):
             )
         ]
     except TypeError:
-        collisions = ['']
+        collisions = ['nibabel']
     # Flush cache if version collision
     if len(collisions) > 0:
         if nilearn.CHECK_CACHE_VERSION:
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -4,13 +4,8 @@ Mixin for cache with joblib
 # Author: Gael Varoquaux, Alexandre Abraham, Philippe Gervais
 # License: simplified BSD
 
-import json
 import warnings
 import os
-import shutil
-from nilearn.version import _compare_version
-
-import nibabel
 
 from joblib import Memory
 
@@ -18,9 +13,6 @@ MEMORY_CLASSES = (Memory, )
 
 import nilearn
 from .helpers import stringify_path
-
-
-__CACHE_CHECKED = dict()
 
 
 def _check_memory(memory, verbose=0):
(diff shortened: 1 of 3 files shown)


nilearn/tests/test_cache_mixin.py
```diff
@@ -89,6 +89,14 @@ def test__safe_cache_flush():
...
         assert not os.path.exists(nibabel_dir)
 
+        # Test handling of TypeError if version name is not str
+        cache_mixin.__CACHE_CHECKED = {}
+        with open(version_file, 'w') as f:
+            json.dump({"nibabel": [0.0]}, f)
+        cache_mixin._safe_cache(mem, f)
+        assert os.path.exists(version_file)
+        assert not os.path.exists(nibabel_dir)
+
```
bthirion, line 99: This test simply repeats the previous one, or am I missing something ?

ymzayek: The difference is in line 95 (vs 87) where the nibabel version is a list instead of a string leading to the `TypeError`

NicolasGensollen (approved):
Thanks @ymzayek !
Have you investigated whether this check is still needed?
If it is, I agree with @jeromedockes that we should simplify this function a bit.

ymzayek:
@NicolasGensollen thanks for the review! Indeed this is a quick fix but if there is no rush for this I can investigate a bit more. I haven't reproduced the error yet. And in the least sure I can simplify it.

NicolasGensollen:
> @NicolasGensollen thanks for the review! Indeed this is a quick fix but if there is no rush for this I can investigate a bit more. I haven't reproduced the error yet. And in the least sure I can simplify it.

Sounds good ! I forgot to write that it could be worth adding a what's new entry since this is a bug fix.

ymzayek:
So this check is still needed. The comment in the docstring is outdated and a bit misleading since it only refers to the first few lines defining `location`. I removed this and I tried to simplify some more but it always ended up having to add other lines. For example, removing `commons` and the list comprehension requires adding a conditional statement, etc.

I also think the `AttributeError` exception is only needed to resolve an error if older versions of joblib are used (which we do not support anymore); so it can probably be removed. I kept it there for now because I'm unsure if there are other cases in which it might be needed.

bthirion (approved):
Thx !

nilearn/_utils/cache_mixin.py
```diff
@@ -108,15 +102,17 @@ def _safe_cache(memory, func, **kwargs):
...
-        )
-    ]
+    try:
+        collisions = [
+            m for m in commons if not _compare_version(
+                versions[m], '==', my_versions[m]
+            )
+        ]
+    except TypeError:
+        collisions = ['']
```
jeromedockes, line 115: wouldn't "collisions = ['nibabel']" be a bit clearer?

jeromedockes: but still it is a bit weird to have this set logic with intersections and iterating over sequences if this is just for nibabel we could just have something like `if "nibabel" in versions and versions_differ(versions["nibabel"], nibabel.__version__): flush_cache()`.

also note that if we use nilearn once with CHECK_CACHE_VERSION cleared, we can end up with a cache directory containing a json listing a nibabel version which actually isn't the one used to create the data in the cache (this function does not flush the data but overwrites the versions JSON). and then future checks will be based on false information, or information that is true for part of the results in the cache and false for some other results

ymzayek: The best way to simplify it was to create another function to handle the exception because it made it hard to avoid repeating code.

And I added a return statement to avoid the behavior you describe.

nilearn/_utils/cache_mixin.py
```diff
@@ -80,17 +80,18 @@ def _check_memory(memory, verbose=0):
     return memory
 
 
+def _nibabel_versions_equal(versions, my_versions):
```
jeromedockes, line 83: as this is not necessarily specific to nibabel and applies to a pair of version, not collections, maybe we can change the names a little bit eg `_safe_versions_equal(version_a, version_b)`

nilearn/_utils/cache_mixin.py
```diff
@@ -108,17 +109,12 @@ def _safe_cache(memory, func, **kwargs):
...
-    my_versions = dict((m.__name__, m.__version__) for m in modules)
-    commons = set(versions.keys()).intersection(set(my_versions.keys()))
-    collisions = [
-        m for m in commons if not _compare_version(
-            versions[m], '==', my_versions[m]
-        )
-    ]
-    # Flush cache if version collision
-    if len(collisions) > 0:
+    my_versions = {nibabel.__name__: nibabel.__version__}
```
jeromedockes, line 113: we don't really need to define the `my_versions` collection but could simply compare 
```
    if 'nibabel' in versions and not _nibabel_versions_equal(
            versions['nibabel'], nibabel.__name__
    )
```

nilearn/_utils/cache_mixin.py
```diff
@@ -145,6 +141,7 @@ def _safe_cache(memory, func, **kwargs):
         else:
             warnings.warn("Incompatible cache in %s: "
                           "old version of nibabel." % location)
+            return memory.cache(func, **kwargs)
```
jeromedockes, line 144: adding the return statement here will not prevent the versions JSON being overwritten with false information when `CHECK_CACHE_VERSION` is false. as this is a bit complex and independent of the simple type error in version comparisons I would suggest leaving that aside for now and maybe opening a separate issue.

jeromedockes (commented):
thanks! I still think we should revisit whether flushing the cache whenever the nibabel version changes is still necessary. there is a broad statement in `nilearn.__init__` that "it does not respect the backward compatibility", but that comment dates from 2014 so I wonder if it is still relevant in 2022. @ymzayek  you mentioned it may still be necessary; did you find an example where that is the case?

nilearn/_utils/cache_mixin.py
cr3: I would suggest a few tests for this function with the name suggested by @jeromedockes :
```
diff --git a/nilearn/tests/test_cache_mixin.py b/nilearn/tests/test_cache_mixin.py
index 7687b37e..615223e7 100644
--- a/nilearn/tests/test_cache_mixin.py
+++ b/nilearn/tests/test_cache_mixin.py
@@ -182,3 +182,19 @@ def test_cache_shelving():
         res = cache_mixin.cache(f, mem, shelve=True)(2)
         assert res.get() == 2
         assert len(_get_subdirs(joblib_dir)) == 1
+
+
+@pytest.mark.parametrize('version_a,version_b,expected', [
+    ("1.2", "1.2", True),
+    ("1.2.a.dev", "1.2a0dev", True),
+    ("1.2", "1.3", False),
+    ("1.2.3", "1.2", False),
+    ("1.2", "1.2.3", False),
+    (None, "1.2", False),
+    (None, None, False),
+    ("1.2", None, False),
+    ("1.2", 1, False),
+    ("1.2", {}, False),
+])
+def test_safe_versions_equal(version_a, version_b, expected):
+    assert cache_mixin._safe_versions_equal(version_a, version_b) == expected
```
I also think this is the correct unit where to test the `TypeError` as discussed [here](https://github.com/nilearn/nilearn/pull/3375/files#r986539527).

nilearn/_utils/cache_mixin.py
```diff
@@ -80,17 +80,18 @@ def _check_memory(memory, verbose=0):
...
     """A wrapper for mem.cache that flushes the cache if the version
     number of nibabel has changed.
 
     """
-    ''' Workaround for
-     https://github.com/scikit-learn-contrib/imbalanced-learn/issues/482
-    joblib throws a spurious warning with newer scikit-learn.
-    This code uses the recommended method first and the deprecated one
-    if that fails, ensuring th warning is not generated in any case.
-    '''
```
cr3, line 93: As long as you're removing this comment, I would also suggest removing the `AttributeError` block. The original intent of that block was for falling back to `memory.cachedir` when there was no such `memory.location` attribute. However, the code now falls back to `memory.location` if there is no such `memory.location` attribute which doesn't make sense.

nilearn/_utils/cache_mixin.py
cr3: @jeromedockes we do need `my_versions` near the bottom of `_safe_cache` to overwrite the `version_file` with that nibabel version - also note that you probably meant `nibabel.__version__` instead of `nibabel.__name__` in your snippet. Near the bottom, we see the condition `if versions != my_versions` which doesn't seem to be as thorough as the condition you suggest with `_safe_versions_equal` (using your naming). I think we could safely include overwriting the `version_file` in the same condition such that the patch now looks something like:
```
diff --git a/nilearn/_utils/cache_mixin.py b/nilearn/_utils/cache_mixin.py
index fc8daae3..ff93548a 100644
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -80,6 +80,13 @@ def _check_memory(memory, verbose=0):
     return memory


+def _safe_versions_equal(version_a, version_b):
+    try:
+        return _compare_version(version_a, '==', version_b)
+    except TypeError:
+        return False
+
+
 def _safe_cache(memory, func, **kwargs):
     """A wrapper for mem.cache that flushes the cache if the version
     number of nibabel has changed.
@@ -108,17 +115,8 @@ def _safe_cache(memory, func, **kwargs):
         with open(version_file, 'r') as _version_file:
             versions = json.load(_version_file)

-    modules = (nibabel, )
-    # Keep only the major + minor version numbers
-    my_versions = dict((m.__name__, m.__version__) for m in modules)
-    commons = set(versions.keys()).intersection(set(my_versions.keys()))
-    collisions = [
-        m for m in commons if not _compare_version(
-            versions[m], '==', my_versions[m]
-        )
-    ]
-    # Flush cache if version collision
-    if len(collisions) > 0:
+    my_versions = {nibabel.__name__: nibabel.__version__}
+    if not _safe_versions_equal(versions.get('nibabel'), my_versions.get('nibabel')):
         if nilearn.CHECK_CACHE_VERSION:
             warnings.warn("Incompatible cache in %s: "
                           "different version of nibabel. Deleting "
@@ -146,8 +144,6 @@ def _safe_cache(memory, func, **kwargs):
             warnings.warn("Incompatible cache in %s: "
                           "old version of nibabel." % location)

-    # Write json files if configuration is different
-    if versions != my_versions:
         with open(version_file, 'w') as _version_file:
             json.dump(my_versions, _version_file)

```

cr3: Note that the above changes still work given a few tests like:
```
diff --git a/nilearn/tests/test_cache_mixin.py b/nilearn/tests/test_cache_mixin.py
index 7687b37e..c6b8989f 100644
--- a/nilearn/tests/test_cache_mixin.py
+++ b/nilearn/tests/test_cache_mixin.py
@@ -10,11 +10,15 @@ from pathlib import Path
 import pytest

 from joblib import Memory
+from nibabel._version import get_versions

 import nilearn
 from nilearn._utils import cache_mixin, CacheMixin


+_NIBABEL_VERSION = get_versions()['version']
+
+
 def _get_subdirs(top_dir):
     top_dir = Path(top_dir)
     children = list(top_dir.glob("*"))
@@ -182,3 +186,23 @@ def test_cache_shelving():
         res = cache_mixin.cache(f, mem, shelve=True)(2)
         assert res.get() == 2
         assert len(_get_subdirs(joblib_dir)) == 1
+
+
+@pytest.mark.parametrize('old_versions,new_versions', [
+    ({}, {'nibabel': _NIBABEL_VERSION}),
+    ({'nibabel': '0.0.0'}, {'nibabel': _NIBABEL_VERSION}),
+    ({'nibabel': '1000.0.0'}, {'nibabel': _NIBABEL_VERSION}),
+    ({'nibabel': 0}, {'nibabel': _NIBABEL_VERSION}),
+    ({'test': _NIBABEL_VERSION}, {'nibabel': _NIBABEL_VERSION}),
+])
+def test_safe_cache(old_versions, new_versions):
+    with tempfile.TemporaryDirectory() as temp_dir:
+        version_file = Path(temp_dir) / 'joblib' / 'module_versions.json'
+        version_file.parent.mkdir()
+        version_file.write_text(json.dumps(old_versions))
+
+        memory = Memory(location=temp_dir)
+        cache_mixin._safe_cache(memory, f)
+
+        versions = json.loads(version_file.read_text())
+        assert versions == new_versions
```

ymzayek:
Thanks all for the reviews! I tried to look more into whether this function is actually still needed as this still not clear as pointed out. Originally this was implemented because of [nibabel backwards compatibility issues](https://github.com/nilearn/nilearn/blob/dbda83bd2c95b352ac8d924561b2c6df54ec15d4/nilearn/__init__.py#L54) before nibabel version 2.0.0. Now we support versions >= 3.0.0. I did not see any relevant references about breaking backwards compatibility on the nibabel github issues or pr pages. I haven't done the most thorough search but at the moment I cannot provide an example where this is still the case. 
So the options would be to either to keep it just in case, and in that case I'll make the suggested updates, or to do away with it and see if anything comes up in the future. WDYT?

bthirion:
But people have complained about it recently  ?

bthirion:
OK, sorry. Then let's get rid of it.

ymzayek:
As decided during the core developers meeting, I removed the `_safe_cache` function and related components. This is ready for another round of reviews.

bthirion:
Still good with me, thx !

ymzayek:
Merging this one once all the tests pass. I think we'll know in due time if the function needs to be reinstated

PR #3383, opened 2022-10-13
[INFRA] Migrate documentation build from CircleCI to Actions 

### LINKED ISSUE #3239: CI service discussion
This concerns opening a discussion around the use of CircleCI as our CI platform for documentation building and whether we should consider GitHub Actions. We should think about the pros and cons of each and see whether we can test GitHub Actions. Also just a note, we found out from talking to a circleci associate that there is an open-source version that we qualify for that should give us enough credit to run our build.

ymzayek:
Closes #3239

- Docs are now built using github actions
- CircleCI still used to host artifacts and deploy them

Still to do:
- [x] Data caching
- [x] Add nightly build
- [x] Create CircleCI token
- [x] Update CircleCI README


bthirion:
Thx for launching that one ! Ping me when it is ready.

.circleci/config.yml
```diff
@@ -1,318 +1,43 @@
...
-
-            - save_cache:
-                key: v1-spm_auditory-{{ checksum "week_num" }}
-                paths:
-                    - ~/nilearn_data/spm_auditory          # ~230 MB
-
-            - save_cache:
-                key: v1-spm_multimodal_fmri-{{ checksum "week_num" }}
-                paths:
-                    - ~/nilearn_data/spm_multimodal_fmri   # ~227 MB
```
bthirion, line 206: IIUC, all the pre-caching done here is no longer done with Github actions, right ?

bthirion (commented):
Looks great, thx  @ymzayek !

.circleci/config.yml
ymzayek: That's right, I haven't adapted the pre-caching of the data yet but it is on the todo list. I'm just trying to figure out the best to do it for the setup of the Actions workflow.

ymzayek:
@NicolasGensollen can you take a look at the webbrowser error in this build: https://github.com/nilearn/nilearn/actions/runs/3299083617/jobs/5442033085 ? Any ideas on how to resolve this in the github actions runner configurations?

.circleci/config.yml
```diff
@@ -1,318 +1,43 @@
-# quick-build rebuilds changes using the cached documentation.
-# The cache is emptied everyday, forcing a full build on the day's first push.
-# It doesn't operate on main branch. New branches are always built from scratch.
-# full-build always rebuilds from scratch, without any cache. Only for changes in main branch.
+# Documentation is now built using Github Actions.
+# CircleCI is used to download the artifacts so it can host and deploy them.
```
bthirion, line 2: To be discussed: whether we keep that in the long term.

ymzayek:
I managed to finally get the full build to work. You can find a successful run on my fork here: https://github.com/ymzayek/nilearn/actions/runs/3362057463. At the bottom of the page you can download the artifacts. The data caching is also working now. In order to get it to run in full in this pull request we need to merge #3405. It would be great if that can be reviewed soon!

NicolasGensollen:
> @NicolasGensollen can you take a look at the webbrowser error in this build: https://github.com/nilearn/nilearn/actions/runs/3299083617/jobs/5442033085 ? Any ideas on how to resolve this in the github actions runner configurations?

Sorry for the late reply @ymzayek. I was in holidays and just got your message today...
Looks like you solved the issue in the mean time ! 👍

bthirion (commented):
Looks mostly good to me, thx !

emdupre:
Just to confirm on the doc hosting piece : If these are hosted by CircleCI, are you planning to put a CNAME on nilearn.github.io but otherwise wipe that repository ?

If so (and sorry if I'm totally misunderstanding !), should we update [this section of the documentation](https://nilearn.github.io/stable/maintenance.html#build-and-deploy-the-documentation) in a future PR ?

build_tools/github/download_data.sh
```diff
@@ -0,0 +1,77 @@
...
+datasets.fetch_megatrawls_netmats()
+datasets.fetch_mixed_gambles()
+datasets.fetch_miyawaki2008()
+datasets.fetch_oasis_vbm()
+datasets.fetch_atlas_yeo_2011()
+datasets.fetch_development_fmri(n_subjects=60)
+datasets.fetch_fiac_first_level()
+datasets.fetch_localizer_first_level()
+# TODO: see timeout error that happens sometimes
+#datasets.fetch_spm_multimodal_fmri()
```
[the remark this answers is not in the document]
ymzayek: Sometimes during the data download step, the call to download this dataset results in a connection timeout error. It seems to never happen when the data is downloaded as part of the sphinx build step but I haven't really investigated it. In any case, I can remove it for now, the data will still be downloaded and cached, but just not in this step. This can be fixed in a later PR.

nilearn/datasets/utils.py
```diff
@@ -21,7 +21,7 @@
 
 from .._utils import fill_doc
 
-_REQUESTS_TIMEOUT = (15.1, 61)
+_REQUESTS_TIMEOUT = (20.1, 61)
```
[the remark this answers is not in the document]
ymzayek: I was trying to see if this quick fix resolved the timeout error mentioned in the above comment by giving the server more time to respond. It doesn't. I will revert this change.

ymzayek:
@emdupre sorry I probably wasn't clear about what is hosted. I meant that we can still use CircleCI to host the artifacts, and not the docs. This is already the case and wouldn't change. The only difference is that now the build job happens in github actions. So this doesn't change the hosting of the stable and dev docs using github pages. CircleCI would still push the dev docs to `nilearn.github.io` after a merge.

That being said we can find another way to host the artifacts or maybe this is not even needed. The problem with github actions is that it does not have this feature. What you get after a build is a zip file you can download in order to view the artifacts which is less convenient than what CircleCI offers.

bthirion:
Sorry, where do we find artifacts on this PR?

ymzayek:
@bthirion the artificats can always be found if you click on Details next to the 'Documentation builder' job and then click Summary. Then you can scroll down and click on doc to download them. I link to the last build artifacts here: https://github.com/nilearn/nilearn/actions/runs/3411730145

bthirion:
Got it thx. You mentioned you had still a few details to fix on this PR ?

bthirion:
LMK when it's finalized.

ymzayek:
@bthirion will do! I got everything (including circleci hosting of artifacts) to run on my fork so I think it's just a permissions issue I have to resolve to get the whole workflow to run as expected here.

jeromedockes:
summarizing the office hours discussion:
- main point: unlike in circleci, creating and restoring the cache are done by the same action with the same key, so the old mechanism of using "missing" instead of the week number as the key doesn't work anymore on github actions
- details:
  - 'week_num' instead of '**/week_num*' 
  - 'full doc' or something else instead of 'circle full'
  - echo "false" > restore.txt rather than touch restore.txt

bthirion:
I'm a bit puzzled by the error since the download seems to work ...

ymzayek:
I'm not sure either. I'm going to remove the separate data download step and keep the pipeline as close as possible to what we have on circleci to see if I can isolate the problem to my implementation. In the end if it works like that, we can move forward as this extra step only really saves ~20min once a week when data is freshly downloaded. And it can be implemented at a later stage if we really want it.

bthirion (approved):
I did no identify any conspicuous issue. I guess we should merge and look carefully whether all subsequent docs building and storing work properly.
Thx, this was a tough one !

NicolasGensollen (approved):
Thanks for all this work @ymzayek ! 💯 
I didn't see anything problematic when reviewing but it is difficult to tell with this kind of PR.
I agree with @bthirion that we should go ahead and try it on main.

ymzayek:
Thanks for the reviews! Any ideas about the PEP8 error?

ymzayek:
> Thanks for the reviews! Any ideas about the PEP8 error?

Being handled in https://github.com/nilearn/nilearn/pull/3420

ymzayek:
I think this is ok to merge but I just want to note a few things that probably need to be followed up:
- I use a personal token to access CircleCI API in [.github/workflows/trigger-hosting.yml](https://github.com/nilearn/nilearn/pull/3383/files#diff-ee024873bed83bf6475dbbfa8a7ca21fdd709fa887b5e5edb73014d7d9d74332), it is recommended to use a machine user instead: see https://support.circleci.com/hc/en-us/articles/360060360811-CircleCI-API-v2-Returns-404-Not-Found-
- Use REST API to download artifacts instead of external service: https://github.com/scikit-learn/scikit-learn/pull/21137#issuecomment-1143677419
- Commit tag '[force download]' only works with full build so it could be a bit misleading. I tried to document this better. It would work alone if I manage to successfully implement the separate the data downloading step.

bthirion:
Then you probably want to open issues for the three points and merge ? Unless you ca address these points quickly enough ?

ymzayek:
@bthirion I think they can be done in follow-up prs so I'll move those to new issue.
I can merge this now then

ymzayek:
Failures are not related and will be handled separately.

PR #3385, opened 2022-10-14
[FIX/ENH] signal interpolation for scrubbed volumes when applying butterworth filter in signal.clean

### LINKED ISSUE #3364: Censoring (and interpolation) in `signal.clean`
Hi! I just attended the NiLearn office hours to learn more about how censoring works with `signal.clean`. Through inspecting the [source code](https://github.com/nilearn/nilearn/blob/9aa503f5/nilearn/signal.py#L460), it was determined that the volumes flagged with the `sample_masks` argument are censored, and then the uncensored volumes adjacent to the first and last censored volumes in a given censored block are treated as if they were collected one after another in subsequent processing steps (e.g., for the purposes of bandpass filtering). Conceptually, this seems like it may be problematic, but practically none of us in the NiLearn Office Hours knew if it causes problems/substantially changes results.

[Historically](https://github.com/GrattonLab/GrattonLab-General-Repo/blob/master/FCProcess/FCPROCESS_GrattonLab.m), I have seen the following steps performed: censoring, temporal interpolation (ideally power spectral matched), bandpass filtering, and then censoring again of the interpolated volumes prior to functional connectivity estimation. Power spectral matched temporal interpolation has the benefit of having realistic dummy volumes so that volumes adjacent to the censored volumes are not contaminated by unrealistic data during bandpass filtering, which is especially important in high motion subjects, considering linear interpolation over many volumes will not come close to approximating realistic data.

Has there been any side-by-side comparisons of these, and potentially other, censoring methods? If in fact it is better to utilize temporal interpolation, would it be possible to implement this in NiLearn? Also, if possible, it would be great to have more documentation about exactly what is happening with each of the processing steps after fMRIPrep and before functional connectivity estimation in NiLearn. A timeline of sorts would be awesome, especially for `NiftiLabelsMasker` (selfishly).

htwangtw:
Move volume censoring after signal filtering so the cosine drift term can be correctly generated.

Closes #3364.

I also took the chance to add a small feature - detect if cosine drift terms are already supplied in confounds, fix a void test etc.

- enh: check if confounds supplied by user contains cosine drift terms
- fix/test: change the high-pass filter in the test so the test data actually contains cosine drift term, rather than just a constant (oops)
- fix: remove constant from cosine drift term as this function works on mean centred data. Having the constant doesn't impact the performance (see `test_clean_confounds`)
- add: interpolation function and related documentation
- fix and refactor: processing runs independently, make the code work better for interpolation, and added relevant tests

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -16,7 +16,7 @@ Fixes
 - Regressor names can now be invalid identifiers but will raise an error with :meth:`~glm.first_level.FirstLevelModel.compute_contrast` if combined to make an invalid expression (:gh:`3374` by `Yasmin Mzayek`_).
 - Fix :func:`~plotting.plot_connectome` which was raising a ``ValueError`` when ``vmax < 0`` (:gh:`3390` by `Paul Bogdan`_).
 - Change the order of applying ``sample_masks`` in :func:`~.signal.clean` based on different filtering options (:gh:`3385` by `Hao-Ting Wang`_). 
-- When using cosine filter and ``sample_masks`` is used, :func:`~.signal.clean` generates the cosine descrete regressors using the full time seiries (:gh:`3385` by `Hao-Ting Wang`_). 
+- When using cosine filter and ``sample_masks`` is used, :func:`~.signal.clean` generates the cosine discrete regressors using the full time seiries (:gh:`3385` by `Hao-Ting Wang`_). 
 
 Enhancements
 ------------
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -15,13 +15,13 @@ Fixes
 
 - Regressor names can now be invalid identifiers but will raise an error with :meth:`~glm.first_level.FirstLevelModel.compute_contrast` if combined to make an invalid expression (:gh:`3374` by `Yasmin Mzayek`_).
 - Fix :func:`~plotting.plot_connectome` which was raising a ``ValueError`` when ``vmax < 0`` (:gh:`3390` by `Paul Bogdan`_).
-- Change the order of applying ``sample_masks`` in :func:`~.signal.clean` based on different filtering options (:gh:`3385` by `Hao-Ting Wang`_). 
-- When using cosine filter and ``sample_masks`` is used, :func:`~.signal.clean` generates the cosine discrete regressors using the full time seiries (:gh:`3385` by `Hao-Ting Wang`_). 
+- Change the order of applying ``sample_masks`` in :func:`~signal.clean` based on different filtering options (:gh:`3385` by `Hao-Ting Wang`_). 
+- When using cosine filter and ``sample_masks`` is used, :func:`~signal.clean` generates the cosine discrete regressors using the full time series (:gh:`3385` by `Hao-Ting Wang`_). 
 
 Enhancements
 ------------
 
-- :func:`nilearn.signal.clean` imputes scrubbed volumes (defined through ``sample_masks``) with cubic spline function before applying butterworth filter (:gh:`3385` by `Hao-Ting Wang`_). 
+- :func:`~signal.clean` imputes scrubbed volumes (defined through ``sample_masks``) with cubic spline function before applying butterworth filter (:gh:`3385` by `Hao-Ting Wang`_). 
 
 Changes
 -------
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -11,8 +11,13 @@ NEW
 Fixes
 -----
 
+- Change the order of applying ``sample_masks`` in :func:`~.signal.clean` based on different filtering options (:gh:`3385` by `Hao-Ting Wang`). 
+- When using cosine filter and ``sample_masks`` is used, :func:`~.signal.clean` generates the cosine descrete regressors using the full time seiries (:gh:`3385` by `Hao-Ting Wang`). 
+
 Enhancements
 ------------
+
+- :func:`nilearn.signal.clean` imputes scrubbed volumes (defined through ``sample_masks``) with cubic spline function before applying butterworth filter (:gh:`3385` by `Hao-Ting Wang`). 
 
 Changes
 -------
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -480,7 +480,8 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
     Butterworth filtering is only meaningful on evenly-sampled signals.
 
     When performing scrubbing (censoring high-motion volumes) with butterworth
-    filtering, the signal is processed in the following order:
+    filtering, the signal is processed in the following order, based on the
+    second recommendation in :footcite:`Lindquist2018`:
 
     - interpolate high motion volumes with cubic spline interpolation.
     - detrend
@@ -489,18 +490,20 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
     - remove confounds
     - standardize
 
+    According to :footcite:`Lindquist2018`, removal of confounds will be done
+    orthogonally to temporal filters (low- and/or high-pass filters), if both
+    are specified. The censored volumes should be removed in both signals and
+    confounds before the nuissance regression.
+
     When performing scrubbing with cosine drift term filtering, the signal is
-    processed in the following order:
+    processed in the following order, based on the first recommendation in
+    :footcite:`Lindquist2018`:
 
     - generate cosine drift term
-    - censor high motion volumes
+    - censor high motion volumes in both signal and confounds
     - detrend
     - remove confounds
     - standardize
-
-    According to :footcite:`Lindquist2018`, removal of confounds will be done
-    orthogonally to temporal filters (low- and/or high-pass filters), if both
-    are specified.
 
     Parameters
     ----------
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -483,7 +483,7 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
     filtering, the signal is processed in the following order, based on the
     second recommendation in :footcite:`Lindquist2018`:
 
-    - interpolate high motion volumes with cubic spline interpolation.
+    - interpolate high motion volumes with cubic spline interpolation
     - detrend
     - low- and high-pass butterworth filter
     - censor high motion volumes
@@ -577,7 +577,7 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
     -------
     cleaned_signals : :class:`numpy.ndarray`
         Input signals, cleaned. Same shape as `signals` unless `sample_mask`
-        applied.
+        is applied.
 
     Notes
     -----
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -650,7 +650,7 @@ def _filter_signal(signals, confounds, filter, low_pass,
     elif filter == 'cosine':
         from nilearn.glm.first_level.design_matrix import _cosine_drift
         frame_times = np.arange(signals.shape[0]) * t_r
-        # remove constant, as the signal is mean centred
+        # remove constant, as the signal is mean centered
         cosine_drift = _cosine_drift(high_pass, frame_times)[:, :-1]
         confounds = _check_cosine_by_user(confounds, cosine_drift)
     return signals, confounds
@@ -658,15 +658,15 @@ def _filter_signal(signals, confounds, filter, low_pass,
 
 def _check_cosine_by_user(confounds, cosine_drift):
     """Check if cosine term exists, based on correlation > 0.9. """
-    # stack consine drift terms if there's no cosine drift terms in data
+    # stack consine drift terms if there's no cosine drift term in data
     n_cosines = cosine_drift.shape[1]
     cosine_exists = False
     if confounds is None:
         return cosine_drift.copy()
 
-    # check if consie drift term is supplied by user  
-    # given the threshold and timeseries length, there can be no consine drift
-    # terms
+    # check if cosine drift term is supplied by user  
+    # given the threshold and timeseries length, there can be no cosine drift
+    # term
     if n_cosines > 0:
         corr_cosine = np.corrcoef(cosine_drift.T, confounds.T)
         np.fill_diagonal(corr_cosine, 0)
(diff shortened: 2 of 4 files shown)


nilearn/signal.py
```diff
@@ -641,16 +648,40 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
     elif filter == 'cosine':
-        from .glm.first_level.design_matrix import _cosine_drift
+        from nilearn.glm.first_level.design_matrix import _cosine_drift
         frame_times = np.arange(signals.shape[0]) * t_r
-        cosine_drift = _cosine_drift(high_pass, frame_times)
-        if confounds is None:
-            confounds = cosine_drift.copy()
-        else:
-            confounds = np.hstack((confounds, cosine_drift))
+        # remove constant, as the signal is mean centred
```
NicolasGensollen, line 653:
```suggestion
        # remove constant, as the signal is mean centered
```

nilearn/signal.py
```diff
@@ -641,16 +648,40 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
+        cosine_drift = _cosine_drift(high_pass, frame_times)[:, :-1]
+        confounds = _check_cosine_by_user(confounds, cosine_drift)
     return signals, confounds
 
 
+def _check_cosine_by_user(confounds, cosine_drift):
+    """Check if cosine term exists, based on correlation > 0.9. """
+    # stack consine drift terms if there's no cosine drift terms in data
```
NicolasGensollen, line 661:
```suggestion
    # stack consine drift terms if there's no cosine drift term in data
```

nilearn/signal.py
```diff
@@ -641,16 +648,40 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
+    n_cosines = cosine_drift.shape[1]
+    cosine_exists = False
+    if confounds is None:
+        return cosine_drift.copy()
+
+    # check if consie drift term is supplied by user
```
NicolasGensollen, line 667:
```suggestion
    # check if cosine drift term is supplied by user  
```

nilearn/signal.py
```diff
@@ -641,16 +648,40 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
+def _check_cosine_by_user(confounds, cosine_drift):
+    """Check if cosine term exists, based on correlation > 0.9. """
+    # stack consine drift terms if there's no cosine drift terms in data
+    n_cosines = cosine_drift.shape[1]
+    cosine_exists = False
+    if confounds is None:
+        return cosine_drift.copy()
+
+    # check if consie drift term is supplied by user  
+    # given the threshold and timeseries length, there can be no consine drift
```
NicolasGensollen, line 668:
```suggestion
    # given the threshold and timeseries length, there can be no cosine drift
```

nilearn/signal.py
```diff
@@ -641,16 +648,40 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
+    # terms
```
NicolasGensollen, line 669:
```suggestion
    # term
```

nilearn/signal.py
```diff
@@ -641,16 +648,40 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
+    if n_cosines > 0:
+        corr_cosine = np.corrcoef(cosine_drift.T, confounds.T)
+        np.fill_diagonal(corr_cosine, 0)
+        cosine_exists = sum(corr_cosine[:n_cosines, :].flatten() > 0.9) > 0
+
+    if cosine_exists:
+        warnings.warn(
+            "Consine filter exists in user supplied confounds."
```
NicolasGensollen, line 677:
```suggestion
            "Cosine filter exists in user supplied confounds."
```

NicolasGensollen (commented):
Thanks @htwangtw !
Some small typos I found while making a first pass. Looks good otherwise.

htwangtw:
The tests are failing because volume censoring should happen before detrending. 
I think I should focus on a version that works for the cosine filter option and think about the butterworth filter separately.

nilearn/signal.py
```diff
@@ -641,16 +661,46 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
 
 
+def _check_cosine_by_user(confounds, cosine_drift):
+    """Check if cosine term exists, based on correlation > 0.9. """
+    # stack consine drift terms if there's no cosine drift term in data
+    n_cosines = cosine_drift.shape[1]
+
+    if n_cosines == 0:
+        warnings.warn(
+            "Cosine filter was not create. The time series might be too short "
```
bthirion, line 679:
```suggestion
            "Cosine filter was not created. The time series might be too short "
```

nilearn/signal.py
```diff
@@ -641,16 +661,46 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
+    # given the threshold and timeseries length, there can be no cosine drift
+    # term
+    corr_cosine = np.corrcoef(cosine_drift.T, confounds.T)
+    np.fill_diagonal(corr_cosine, 0)
+    cosine_exists = sum(corr_cosine[:n_cosines, :].flatten() > 0.9) > 0
+
+    if cosine_exists:
+        warnings.warn(
+            "Cosine filter exists in user supplied confounds."
+            "Use user supplied cosine regressors."
```
bthirion, line 697:
```suggestion
            "Use user supplied regressors only."
```

bthirion (commented):
Thx, this looks good.

nilearn/signal.py
```diff
@@ -641,16 +661,46 @@ def _filter_signal(signals, confounds, filter, low_pass, high_pass, t_r):
...
+        from nilearn.glm.first_level.design_matrix import _cosine_drift
         frame_times = np.arange(signals.shape[0]) * t_r
-        cosine_drift = _cosine_drift(high_pass, frame_times)
-        if confounds is None:
-            confounds = cosine_drift.copy()
-        else:
-            confounds = np.hstack((confounds, cosine_drift))
+        # remove constant, as the signal is mean centered
+        cosine_drift = _cosine_drift(high_pass, frame_times)[:, :-1]
+        confounds = _check_cosine_by_user(confounds, cosine_drift)
```
NicolasGensollen, line 668: Also, I might be missing something but I was expecting the signals to be modified somehow by this function since it is called `_filter_signal`. It is the case when `filter == butterworth`, but doesn't seem to be the case with cosine.
I just wanted to check that this is expected as it feels a little surprising when reading the code.

nilearn/signal.py
```diff
@@ -594,11 +608,21 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
...
+    if filter_type == 'butterworth':
+        signals = butterworth(signals, sampling_rate=1. / t_r,
+                              low_pass=low_pass, high_pass=high_pass)
+        if confounds is not None:
+            # Apply low- and high-pass filters to keep filters orthogonal
+            # (according to Lindquist et al. (2018))
+            confounds = butterworth(confounds, sampling_rate=1. / t_r,
+                                    low_pass=low_pass, high_pass=high_pass)
+        # apply sample_mask to remove censored volumes after signal filtering
+        if sample_mask is not None:
```
bthirion, line 622: I'm not sure I understand the logic here: if there is an outlying value, then signals and confounds estimates will be (possibly strongly) biased. I think that the values out of samples mask should be soehow imputed so that they do not contaminate the others ?

htwangtw: Yes you are correct here. The steps should be:
1. censor volumes flagged by `sample_mask`
2. impute volumes flagged by `sample_mask`
3. apply butterworth filter
4. censor volumes flagged by `sample_mask`

I haven't implement any thing related to interpolation and was planning on doing it in a different PR. If you think it makes more sense to address it here, I can convert this PR to draft and do it in one PR. WDYT?

bthirion: I propose to take care of it in this PR. Thx !

bthirion:
Let me know when you're ready for a review. Best, Bertrand

htwangtw:
Just about to!

What's changed:

- interpolation function and related documentation
- fix and refactor processing runs independently, and added relevant tests

nilearn/signal.py
```diff
@@ -630,29 +678,78 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
...
-        from .glm.first_level.design_matrix import _cosine_drift
-        frame_times = np.arange(signals.shape[0]) * t_r
-        cosine_drift = _cosine_drift(high_pass, frame_times)
-        if confounds is None:
-            confounds = cosine_drift.copy()
-        else:
-            confounds = np.hstack((confounds, cosine_drift))
+            confounds = _interpolate_volumes(confounds, sample_mask, t_r)
+    else:  # Or censor when no filtering, or cosine filter
+        signals = signals[sample_mask, :]
```
NicolasGensollen, line 692: I think we can turn these 3 lines into a function `_censor_signals(signals, confounds, sample_mask)` since it is used in multiple places. WDYT?

nilearn/signal.py
```diff
@@ -575,30 +596,57 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
...
+    if filter_type == 'butterworth':
+        signals = butterworth(signals, sampling_rate=1. / t_r,
+                              low_pass=low_pass, high_pass=high_pass)
+        if confounds is not None:
+            # Apply low- and high-pass filters to keep filters orthogonal
+            # (according to Lindquist et al. (2018))
+            confounds = butterworth(confounds, sampling_rate=1. / t_r,
+                                    low_pass=low_pass, high_pass=high_pass)
+        # apply sample_mask to remove censored volumes after signal filtering
+        if sample_mask is not None:
```
NicolasGensollen, line 646: We would use `_censor_signals` here for instance

NicolasGensollen (commented):
Thanks @htwangtw !
I made some small suggestions but LGTM otherwise. 👍

bthirion (approved):
LGTM, thx !

NicolasGensollen (approved):
LGTM, thanks @htwangtw !
Could you also add a what's new entry ?

htwangtw:
@butellyn Do you want to have a look before we merge this PR?

nilearn/signal.py
```diff
@@ -464,19 +465,38 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
...
+
+    When performing scrubbing (censoring high-motion volumes) with butterworth
+    filtering, the signal is processed in the following order:
+
+    - interpolate high motion volumes with cubic spline interpolation.
+    - detrend
+    - low- and high-pass butterworth filter
+    - censor high motion volumes
+    - remove confounds
+    - standardize
```
tsalo, line 490: Sorry for the noise, but does this order of steps stem from Lindquist2018 (cited below), or is there another paper that should be cited for this section?

butellyn:
> @butellyn Do you want to have a look before we merge this PR?

Hey, sorry for the late response. This week is pretty busy for me. I'll try to check out the code another time.

nilearn/signal.py
butellyn: My understanding is that they do temporal bandpass filtering and nuisance regression simultaneously ("One can define an omnibus projection matrix consisting of all nuisance variables one seeks to remove from the data. This entails performing motion regression, CompCor, temporal filtering, and so on in a common joint model."). I think this gets complicated because, as far as I can tell, they didn't perform any form of censoring, so it isn't totally clear where this should be incorporated.

The approach I am taking is the following, which mirrors Caterina Gratton and Jonathan Power's pipeline, but includes nuisance regression twice so as to avoid confounds influencing the interpolation step, while also performing temporal bandpass filtering and nuisance regression simultaneously:
1. Demean and detrend
2. Nuisance regression
3. Censor volumes with fFD > .1
4. Interpolate over censored TRs
5. Temporal bandpass filtering and nuisance regression
6. Censor volumes with fFD > .1

According to the [NiLearn documentation](https://nilearn.github.io/dev/modules/generated/nilearn.image.clean_img.html), when you specify bandpass filtering in the same call as the nuisance regressors, it performs the method from Lindquist (2019).

htwangtw:
@tsalo and @butellyn 
Sorry for the vague replies and I have detailed the references in the docs. 
I will not implement @butellyn 's suggestion as this is deviating from the steps originally in nilearn too much. If you want to implement other signal cleaning recipe, such as matching the recipe in Power 2014, we can discuss in a different issue.

nilearn/signal.py
tsalo: Thanks for adding the description and reference!

doc/changes/latest.rst
```diff
@@ -15,10 +15,14 @@ Fixes
 
 - Regressor names can now be invalid identifiers but will raise an error with :meth:`~glm.first_level.FirstLevelModel.compute_contrast` if combined to make an invalid expression (:gh:`3374` by `Yasmin Mzayek`_).
 - Fix :func:`~plotting.plot_connectome` which was raising a ``ValueError`` when ``vmax < 0`` (:gh:`3390` by `Paul Bogdan`_).
+- Change the order of applying ``sample_masks`` in :func:`~.signal.clean` based on different filtering options (:gh:`3385` by `Hao-Ting Wang`_). 
+- When using cosine filter and ``sample_masks`` is used, :func:`~.signal.clean` generates the cosine discrete regressors using the full time seiries (:gh:`3385` by `Hao-Ting Wang`_). 
 
 Enhancements
 ------------
 
+- :func:`nilearn.signal.clean` imputes scrubbed volumes (defined through ``sample_masks``) with cubic spline function before applying butterworth filter (:gh:`3385` by `Hao-Ting Wang`_).
```
NicolasGensollen, line 24:
```suggestion
- :func:`~signal.clean` imputes scrubbed volumes (defined through ``sample_masks``) with cubic spline function before applying butterworth filter (:gh:`3385` by `Hao-Ting Wang`_). 
```

nilearn/signal.py
```diff
@@ -553,7 +576,8 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
     Returns
     -------
     cleaned_signals : :class:`numpy.ndarray`
-        Input signals, cleaned. Same shape as `signals`.
+        Input signals, cleaned. Same shape as `signals` unless `sample_mask`
+        applied.
```
NicolasGensollen, line 580:
```suggestion
        is applied.
```

NicolasGensollen (approved):
Thanks @htwangtw !
LGTM, just a couple small suggestions

htwangtw:
I am going forward to merge this as there are other issue (#3406) that can be handled more smoothly. 
@butellyn if you have any future comments we can discuss in [issues](https://github.com/nilearn/nilearn/issues).

PR #3392, opened 2022-10-24
[MAINT] CVE-2007-4559 Patch

ymzayek:
Supersedes #3391 

Patching tarfile `extractall` only at vulnerable locations.


nilearn/datasets/utils.py
```diff
@@ -310,6 +310,24 @@ def _get_dataset_dir(dataset_name, data_dir=None, default_paths=None,
...
+def _is_within_directory(directory, target):
+    abs_directory = os.path.abspath(directory)
+    abs_target = os.path.abspath(target)
+
+    prefix = os.path.commonprefix([abs_directory, abs_target])
+
+    return prefix == abs_directory
+
+
+def _safe_extract(tar, path=".", members=None, *, numeric_owner=False):
```
jeromedockes, line 322: could you add a link to the original issue and https://github.com/advisories/GHSA-gw9q-c7gh-j9vm to keep track of where this came from? thanks! LGTM otherwise

ymzayek: Done!

jeromedockes (approved):

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

nilearn/datasets/tests/test_utils.py
```diff
@@ -316,6 +316,22 @@ def test_uncompress():
             shutil.rmtree(dtemp)
 
 
+def test__safe_extract():
```
jeromedockes, line 319: maybe `test_safe_extract`?

nilearn/datasets/tests/test_utils.py
```diff
@@ -316,6 +316,22 @@ def test_uncompress():
...
+    # Test vulnerability patch by mimicking path traversal
+    dtemp = mkdtemp()
```
jeromedockes, line 321: you can use the tmp_path pathlib fixture I think?

nilearn/datasets/tests/test_utils.py
```diff
@@ -316,6 +316,22 @@ def test_uncompress():
...
+    ztemp = os.path.join(dtemp, 'test.tar')
+    fd, temp = mkstemp(dir=dtemp)
```
jeromedockes, line 323: it might be clearer to give it a name and writes somehting in it like `in_archive_file = tmp_path / "something.txt"; in_archive_file.write_text("hello")`

ymzayek:
@jeromedockes I made the suggested changes. If everything passes I'd say it's ready to merge

jeromedockes:
> @jeromedockes I made the suggested changes. If everything passes I'd say it's ready to merge

great, thank you! yes I think we can merge.

(In general in future PRs I would suggest doing all filesystem path manipulations with pathlib rather than os.path)

jeromedockes (approved):

ymzayek:
> (In general in future PRs I would suggest doing all filesystem path manipulations with pathlib rather than os.path)

@jeromedockes would that be like using `ztemp = tmp_path / 'test.tar'` instead of what I wrote and is that for system compatibility reasons? Also I though I needed `os.path.normpath()` because of windows compatibility but the first test I wrote was not very good so I'm not sure if just setting `arcname='../test.tar'` would now be resolved correctly on windows with this version of the test or if there is another recommended way with pathlib.

Happy to make the changes before merging.

bthirion (approved):
LGTM, thx !

PR #3408, opened 2022-11-06
[MAINT] No need to inherit `object` in Python 3

DimitriPapadopoulos:
Classes inherit from `object` implicitly.

### DIFF
--- a/nilearn/_utils/cache_mixin.py
+++ b/nilearn/_utils/cache_mixin.py
@@ -156,7 +156,7 @@ def _safe_cache(memory, func, **kwargs):
     return memory.cache(func, **kwargs)
 
 
-class _ShelvedFunc(object):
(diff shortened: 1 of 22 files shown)


NicolasGensollen (approved):
Was about time to get rid of these...
Thanks @DimitriPapadopoulos ! 👍

ymzayek (commented):
Thanks @DimitriPapadopoulos ! I found one that is not included here https://github.com/nilearn/nilearn/blob/6920ffae3bbc2eafefd0f66ff36af5ca4c9f22ee/nilearn/decoding/space_net.py#L206

Otherwise LGTM!

ymzayek (approved):
Great thanks!

bthirion (approved):
Thx !

PR #3417, opened 2022-11-21
[INFRA][FIX] Actions setup-python for v3.6 does not work with Ubuntu-latest runner

ymzayek:
GitHub Actions now fails with tests that use some python versions such as 3.6 if runner is set to `Ubuntu-latest` (which is now ubuntu-22.04). It should work with ubuntu-20.04.

See failure: https://github.com/nilearn/nilearn/actions/runs/3498548522/jobs/5859058796
More details at similar issue: https://github.com/actions/setup-python/issues/543

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -26,7 +26,7 @@ jobs:
   min_requirements:
     needs: check_skip
     if: ${{ needs.check_skip.outputs.skip == 'false' }}
-    runs-on: "ubuntu-latest"
+    runs-on: "ubuntu-20.04"
     name: Python 3.6, minimal requirements without Matplotlib
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


ymzayek:
@bthirion we can with the next major release: https://github.com/nilearn/nilearn/blob/c123b6126ef36dd5a08c20b780b773a80278b907/nilearn/__init__.py#L53-L56

For now I suggest to merge this and then I'd update the CI accordingly during the release process.

jeromedockes:
> When should we drop support for 3.6 ?

I think it would be nice to delay that a little bit if it's not too
inconvenient, some distributions take a long time to package more recent
python versions, and not all users are running a very recent release of
their OS. For example recently I briefly used one of the virtual
machines made available for the brainhack which was running Red Had 8.3
and the installed python was 3.6.8

bthirion (approved):
LGTM, thx !

NicolasGensollen (approved):
LGTM, thanks @ymzayek !
Even though some systems still use it, Python 3.6 has been EOL since December 2021 and is not maintained anymore (including security fixes). It'd be great to not wait to long to drop it imho.

ymzayek:
I'll put this discussion on the agenda for the coredev meeting. Merging this now.

PR #3420, opened 2022-11-24
[FIX] Flake8 latest release raises error and includes backwards incompatilble change

### LINKED ISSUE #3431: All CI failing because flake8 --diff option has been removed
All CI are currently failing because flake8==6.0.0 (which has been released on Nov 23) does not have the `--diff` flag anymore.

The reason has been explained [in this issue](https://github.com/pycqa/flake8/issues/1389).

A solution which is an adaptation of [this comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3961#issuecomment-1325737480) is replacing the `flake8 --diff` from `build_tools/flake8_diff.sh#L79` with the following command:

```
git diff --name-only $COMMIT | grep '\.py$' | xargs --delimiter='\n' --no-run-if-empty flake8 --show-source
```

Opened PR

ymzayek:
Closes #3431. Flake8 just released a new version that raises a PEP8 failure in our CI check. Also the --diff option is now removed so we have to handle differently how we pipe changes for flake8 to check.
See: https://flake8.pycqa.org/en/latest/release-notes/6.0.0.html#backwards-incompatible-changes

### DIFF
--- a/build_tools/flake8_diff.sh
+++ b/build_tools/flake8_diff.sh
@@ -74,7 +74,8 @@ echo -e '\nRunning flake8 on the diff in the range'\
      "($(git rev-list $COMMIT.. | wc -l) commit(s)):"
 echo '--------------------------------------------------------------------------------'
 
-# Conservative approach: diff without context so that code that was
-# not changed does not create failures
-git diff --unified=0 $COMMIT | flake8 --diff --show-source
+# The --diff argument has been removed from flake8:
+# https://github.com/PyCQA/flake8/issues/1760
+# Now flake8 will be run on entire files but only files that have been changed.
+git diff $COMMIT --name-only -z --diff-filter=d -- '*.py' | xargs -0 flake8
 echo -e "No problem detected by flake8\n"
--- a/build_tools/flake8_diff.sh
+++ b/build_tools/flake8_diff.sh
@@ -77,5 +77,5 @@ echo '--------------------------------------------------------------------------
 # The --diff argument has been removed from flake8:
 # https://github.com/PyCQA/flake8/issues/1760
 # Now flake8 will be run on entire files but only files that have been changed.
-git diff $COMMIT --name-only -z --diff-filter=d -- '*.py' | xargs -0 flake8
+git diff $COMMIT --name-only -z --diff-filter=d -- '*.py' | xargs -0 -r flake8
 echo -e "No problem detected by flake8\n"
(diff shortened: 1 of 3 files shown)


bthirion:
There is still a flake8 error due to --diff option.

ymzayek:
Yes, sorry I haven't had time to look into it. I will change this to a draft until it's ready.

ymzayek:
I changed the `flake8 --diff` command to what is suggested on  this thread: https://github.com/PyCQA/flake8/issues/1760
The problem now is that flake8 will be run on entire changed files instead of just the diff. Do we want this? If we implement automatic formatting discussed in #2528, then it won't be a problem.

jeromedockes:
> I changed the `flake8 --diff` command to what is suggested on  this thread: https://github.com/PyCQA/flake8/issues/1760
> The problem now is that flake8 will be run on entire changed files instead of just the diff. Do we want this? If we implement automatic formatting discussed in #2528, then it won't be a problem.

I would say no, and we can revisit when automatic formatting is in place

ymzayek:
Yes agreed.  I will stall this then

bthirion:
rather -1 on my side. Rather use up-to-date tools.

build_tools/flake8_diff.sh
```diff
@@ -74,7 +74,8 @@ echo -e '\nRunning flake8 on the diff in the range'\
      "($(git rev-list $COMMIT.. | wc -l) commit(s)):"
 echo '--------------------------------------------------------------------------------'
 
-# Conservative approach: diff without context so that code that was
-# not changed does not create failures
-git diff --unified=0 $COMMIT | flake8 --diff --show-source
+# The --diff argument has been removed from flake8:
+# https://github.com/PyCQA/flake8/issues/1760
+# Now flake8 will be run on entire files but only files that have been changed.
+git diff $COMMIT --name-only -z --diff-filter=d -- '*.py' | xargs -0 -r flake8
```
jeromedockes, line 80: does https://github.com/nilearn/nilearn/pull/3420#issuecomment-1327788998 mean we should remove flake8 check entirely for now?

ymzayek: We could since we cannot use it in the way we want. I just stalled it while waiting for a better solution but if the only solution will be implementing automatic formatting then we won't need this I guess

ymzayek:
Closing this since it's superseded by #3484

PR #3422, opened 2022-11-25
[FIX] Use builtin functions instead of numpy's deprecated aliases

ymzayek:
Addresses failures: https://github.com/nilearn/nilearn/actions/runs/3547323012/jobs/5957319235

Using python builtin `float` instead of `np.float` is recommended and does not have any impact on behavior.
Using python builtin `int` instead of `np.int`; idem
See: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations


bthirion (commented):
LGTM so far. Deserves a "whatsnew" entry imho.

bthirion (approved):
LGTM, thx !

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

ymzayek:
Merging

PR #3423, opened 2022-11-28
[MAINT] Extending support for python 3.6 until release 0.11

ymzayek:


jeromedockes:
> I'm not sure what the motivation for extending the support of Python 3.6 is tbh...
> Most projects have dropped 3.6 a while ago (Scikit-learn is >= 3.8, MNE is >= 3.7...)

that's a good point. the motivation is that many machines still only
have python 3.6 installed; but if scikit-learn only supports 3.8, then
as nilearn depends on it it probably makes sense for nilearn to do the
same...

NicolasGensollen:
> that's a good point. the motivation is that many machines still only have python 3.6 installed; but if scikit-learn only supports 3.8, then as nilearn depends on it it probably makes sense for nilearn to do the same...

I could be biased, but I think (and hope) most users rely on virtual environments nowadays.

bthirion:
+1 for being aligned on skl

ymzayek:
@NicolasGensollen thanks for fleshing that out. Yes I think it makes sense to stay aligned with scikit-learn and similar projects. I will close this and open a new PR to drop 3.6 and set a deprecation warning for 3.7

PR #3424, opened 2022-11-28
[MAINT] Remove warning that `nilearn.glm` module is experimental

### LINKED ISSUE #3097: Stop warning that `GLM` module is experimental
Since Nilearn 0.7.0, we have a warning saying that `nilearn.glm` is experimental and might be subject to change, as well as some warnings here and there in the docs.
Since we are about to reach 0.9.0, it could be time to remove these warnings.
There is no explicit end to these warnings though, so we can probably keep them a little longer if needed.
WDYT?

ymzayek:
Closes #3097 .


NicolasGensollen (approved):
The changes LGTM.
I feel like we should add an entry in the change log. WDYT?

bthirion (approved):
LGTM, thx !

ymzayek:
Merging

PR #3427, opened 2022-11-28
[MAINT] Remove deprecated function `compute_multi_gray_matter_mask`

ymzayek:
Remove `masking.compute_multi_gray_matter_mask` which has been deprecated and replaced by `masking.compute_multi_brain_mask` in version 0.8.1

### DIFF
--- a/nilearn/maskers/tests/test_multi_nifti_masker.py
+++ b/nilearn/maskers/tests/test_multi_nifti_masker.py
@@ -193,25 +193,6 @@ def test_mask_strategy_errors():
         mask.fit(imgs)
 
 
-@pytest.mark.parametrize('strategy',
-                         [f'{p}-template' for p in
-                          ['whole-brain', 'gm', 'wm']])
-def test_compute_multi_gray_matter_mask(strategy):
-    imgs = _get_random_imgs((9, 9, 5), 2)
-    masker = MultiNiftiMasker(mask_strategy=strategy,
-                              mask_args={'opening': 1})
-    masker.fit(imgs)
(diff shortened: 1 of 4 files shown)


nilearn/maskers/tests/test_multi_nifti_masker.py
```diff
@@ -193,25 +193,6 @@ def test_mask_strategy_errors():
         mask.fit(imgs)
 
 
-@pytest.mark.parametrize('strategy',
-                         [f'{p}-template' for p in
-                          ['whole-brain', 'gm', 'wm']])
-def test_compute_multi_gray_matter_mask(strategy):
```
NicolasGensollen, line 199: I might be missing something, but I think this test isn't calling the deprecated function :  

https://github.com/nilearn/nilearn/blob/e19b37dcb74cc3819b5b60c49f0917f958c7aacb/nilearn/maskers/multi_nifti_masker.py#L26-L33

ymzayek: Yes you are completely right. I think then it just needs a name change. Would you agree?

NicolasGensollen: > Yes you are completely right. I think then it just needs a name change. Would you agree?

Yup ! The name is misleading !

bthirion (approved):
LGTM, thx.

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

ymzayek:
Merging

PR #3429, opened 2022-11-29
[MAINT] Drop support for python 3.6 and deprecate 3.7

ymzayek:
Even though we initially agreed to extend support for python 3.6 due to many machines still using it, it was then discussed in #3423 that it is more reasonable to drop its support since dependencies such as scikit-learn do not support it anymore and it has been EOL since Dec2021.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -27,7 +27,7 @@ jobs:
     needs: check_skip
     if: ${{ needs.check_skip.outputs.skip == 'false' }}
     runs-on: "ubuntu-20.04"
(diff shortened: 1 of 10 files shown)


NicolasGensollen (commented):
Thanks @ymzayek !
Could you add an entry in the change log saying that support for Python 3.6 has been dropped and that support for Python 3.7 will stop with release `0.12` ?
LGTM otherwise.

bthirion (commented):
LGTM so far.

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

bthirion (approved):
LGTM, thx.

PR #3430, opened 2022-11-29
[FIX] Pass correct type to scikit-learn estimators parameter

ymzayek:
Pre-release of scikit-learn enforces correct behavior of:
 - Passing `int` to the `max_iter` parameter used by some estimators
 - Deletion of deprecated parameter
 - `fit_intercept` only accepts booleans
 
See failing test: https://github.com/nilearn/nilearn/actions/runs/3574331280/jobs/6009470192

### DIFF
--- a/nilearn/decomposition/dict_learning.py
+++ b/nilearn/decomposition/dict_learning.py
@@ -23,7 +23,7 @@ sparse_encode_args = {'check_input': False}
 
 
 def _compute_loadings(components, data):
-    ridge = Ridge(fit_intercept=None, alpha=1e-8)
+    ridge = Ridge(fit_intercept=False, alpha=1e-8)
     ridge.fit(components.T, np.asarray(data.T))
     loadings = ridge.coef_.T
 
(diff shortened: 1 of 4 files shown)


nilearn/decomposition/dict_learning.py
```diff
@@ -23,7 +23,7 @@
 
 
 def _compute_loadings(components, data):
-    ridge = Ridge(fit_intercept=None, alpha=1e-8)
+    ridge = Ridge(fit_intercept=False, alpha=1e-8)
```
bthirion, line 26: Why this change ?

bthirion (commented):
Thx for tackling that. I'm wondering whyw change paramters for Ridge/Lasso and whether the changes are exhaustive.

nilearn/decomposition/dict_learning.py
ymzayek: It seems to me that some intended behavior wasn't enforced. For example see Ridge API of the last stable release: https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.Ridge.html. `fit_intercept` should be a boolean.

And checking the code, it makes sense that `None` could be passed with the same behavior as `False`
https://github.com/scikit-learn/scikit-learn/blob/f3f51f9b6/sklearn/linear_model/_ridge.py#L910

ymzayek:
The failure in https://github.com/nilearn/nilearn/actions/runs/3577698793/jobs/6017001262 seems to be a precision error and I would say is unrelated to this PR. I think it might be because the GitHub Actions runner `ubuntu-latest` was bumped to version `22.04` recently. I cannot reproduce the error on my `ubuntu-20.04` machine with the same prerelease environment. 
I would be for merging this PR and then opening another to take care of that error. WDYT?

bthirion:
In that case, i would address the issue in a separate PR (we need to do it anyhow), and then merge that one after a rebase  ?

NicolasGensollen (approved):
LGTM, thanks for taking care of that @ymzayek !

ymzayek:
I opened an issue to follow-up the other error. Merging this one

Issue #3435, opened 2022-12-01
IndexError in non_parametric_inference when using two_sided_test=True


jrdalenberg:
I am running a non-parametric test for a one-sample second level t-test: 

```python
# Run permutation test
MD_dict = non_parametric_inference(
    MD_imgs,
    design_matrix=MD_design_matrix,
    mask=MD_second_level_mask,
    model_intercept=True,
    second_level_contrast="intercept",
    two_sided_test=True,
    smoothing_fwhm=6,
    n_jobs=8,
    threshold=0.001
)
```
When  `two_sided_test=False`, it runs fine but with `two_sided_test=True`, I get the following error:

```python
File ~/.virtualenvs/brainiak/lib/python3.8/site-packages/nilearn/glm/second_level/second_level.py:862, in non_parametric_inference(second_level_input, confounds, design_matrix, second_level_contrast, first_level_contrast, mask, smoothing_fwhm, model_intercept, n_perm, two_sided_test, random_state, n_jobs, verbose, threshold, tfce)
    859 target_vars = masker.transform(effect_maps)
    861 # Perform massively univariate analysis with permuted OLS
--> 862 outputs = permuted_ols(
    863     tested_var,
    864     target_vars,
    865     model_intercept=model_intercept,
    866     n_perm=n_perm,
    867     two_sided_test=two_sided_test,
    868     random_state=random_state,
    869     n_jobs=n_jobs,
    870     verbose=max(0, verbose - 1),
    871     masker=masker,
    872     threshold=threshold,
    873     tfce=tfce,
    874     output_type='dict',
    875 )
    876 neg_log10_vfwe_pvals_img = masker.inverse_transform(np.ravel(
    877     outputs['logp_max_t']
    878 ))
    880 if (not tfce) and (threshold is None):

File ~/.virtualenvs/brainiak/lib/python3.8/site-packages/nilearn/mass_univariate/permuted_least_squares.py:935, in permuted_ols(tested_vars, target_vars, confounding_vars, model_intercept, n_perm, two_sided_test, random_state, n_jobs, verbose, masker, tfce, threshold, output_type)
    930 for j_val in cluster_labels[1:]:  # skip background
    931     cluster_mass = np.sum(
    932         np.fabs(scores_original_data_3d[labeled_arr3d == j_val])
    933         - threshold_t
    934     )
--> 935     cluster_dict['mass_regressor'][j_val] = cluster_mass
    937 # Calculate p-values from size/mass values and associated h0s
[...]

PR #3436, opened 2022-12-01
[FIX] Combine positive and negative cluster labels for two sided permutation tests 

jrdalenberg:
This addresses the bug report in #3435.

### DIFF
--- a/nilearn/mass_univariate/permuted_least_squares.py
+++ b/nilearn/mass_univariate/permuted_least_squares.py
@@ -908,7 +908,7 @@ def permuted_ols(
                     scores_original_data_3d < -threshold_t,
                     bin_struct,
                 )
-                n_negative_clusters = np.max(temp_labeled_arr3d) 
+                n_negative_clusters = np.max(temp_labeled_arr3d)
                 labeled_arr3d[labeled_arr3d > 0] += n_negative_clusters
                 labeled_arr3d = labeled_arr3d + temp_labeled_arr3d
                 del temp_labeled_arr3d
--- a/nilearn/mass_univariate/permuted_least_squares.py
+++ b/nilearn/mass_univariate/permuted_least_squares.py
@@ -903,15 +903,13 @@ def permuted_ols(
             )
 
             if two_sided_test:
-                # Label positive and negative clusters separately
-                n_positive_clusters = np.max(labeled_arr3d)
+                # Add negative cluster labels
                 temp_labeled_arr3d, _ = label(
                     scores_original_data_3d < -threshold_t,
                     bin_struct,
                 )
-                temp_labeled_arr3d[
-                    temp_labeled_arr3d > threshold_t
-                ] += n_positive_clusters
+                n_negative_clusters = np.max(temp_labeled_arr3d) 
+                labeled_arr3d[labeled_arr3d > 0] += n_negative_clusters
                 labeled_arr3d = labeled_arr3d + temp_labeled_arr3d
                 del temp_labeled_arr3d
 
--- a/nilearn/mass_univariate/tests/test_permuted_least_squares.py
+++ b/nilearn/mass_univariate/tests/test_permuted_least_squares.py
@@ -609,8 +609,10 @@ def test_cluster_level_parameters_smoke(random_state=0):
     ))
 
     columns = np.arange(0, voxel_vars.shape[1])
-    chosen_columns = np.random.choice(columns, size=125, p=[0.1, 0.1, 0.8]) # create 125 voxels
-    target_var = voxel_vars[:, chosen_columns]  # corresponds to 5 x 5 x 5 x 10 niimg
+    # create 125 voxels
+    chosen_columns = rng.choice(columns, size=125, p=[0.1, 0.1, 0.8])
+    # corresponds to 5 x 5 x 5 x 10 niimg
+    target_var = voxel_vars[:, chosen_columns]
     tested_var = np.arange(0, 20, 2)
 
     mask_img = nib.Nifti1Image(np.ones((5, 5, 5)), np.eye(4))
--- a/nilearn/mass_univariate/tests/test_permuted_least_squares.py
+++ b/nilearn/mass_univariate/tests/test_permuted_least_squares.py
@@ -598,7 +598,7 @@ def test_cluster_level_parameters_smoke(random_state=0):
     import nibabel as nib
     from nilearn.maskers import NiftiMasker
 
-    rng = np.random.RandomState(0)
+    rng = np.random.RandomState(random_state)
 
     # create design
     target_var1 = np.arange(0, 10).reshape((-1, 1))  # positive effect
@@ -609,7 +609,7 @@ def test_cluster_level_parameters_smoke(random_state=0):
     ))
 
     columns = np.arange(0, voxel_vars.shape[1])
-    chosen_columns = random.choices(columns, k=125, weights=[1, 1, 5]) # create 125 voxels
+    chosen_columns = np.random.choice(columns, size=125, p=[0.1, 0.1, 0.8]) # create 125 voxels
     target_var = voxel_vars[:, chosen_columns]  # corresponds to 5 x 5 x 5 x 10 niimg
     tested_var = np.arange(0, 20, 2)
 
--- a/nilearn/mass_univariate/tests/test_permuted_least_squares.py
+++ b/nilearn/mass_univariate/tests/test_permuted_least_squares.py
@@ -604,7 +604,7 @@ def test_cluster_level_parameters_smoke(random_state=0):
 
     # create design
     target_var1 = np.arange(0, 10).reshape((-1, 1))  # positive effect
-    voxel_vars = np.hstack(( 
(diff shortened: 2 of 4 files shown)


nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -903,15 +903,13 @@ def permuted_ols(
...
             if two_sided_test:
-                # Label positive and negative clusters separately
-                n_positive_clusters = np.max(labeled_arr3d)
+                # Add negative cluster labels
                 temp_labeled_arr3d, _ = label(
                     scores_original_data_3d < -threshold_t,
                     bin_struct,
                 )
-                temp_labeled_arr3d[
-                    temp_labeled_arr3d > threshold_t
```
tsalo, line 913: Would the bug be fixed by changing `threshold_t` here to `0`?

bthirion (commented):
Thx ! Could you add a test that fails with previous version and passes with the fix ?

nilearn/mass_univariate/permuted_least_squares.py
jrdalenberg: Yes, you are right. I did not fully understand the logic at first and reversed the order it seems: 

`temp_labeled_arr3d > 0` results in `np.concatenate([positive cluster labels], [negative cluster labels])`

And my change results in `np.concatenate([negative clusters labels], [positive cluster labels])`

Which do you prefer?

jrdalenberg:
> Thx ! Could you add a test that fails with previous version and passes with the fix ?

Sure! But I first need to figure out how to build tests 😁.

If you want to have a look already:

```python

from nilearn.glm.second_level import non_parametric_inference
from nilearn.glm.second_level import SecondLevelModel
import pandas as pd


n_subjects = 16
data = fetch_localizer_contrasts(
    ['left vs right button press'],
    n_subjects,
    get_tmaps=True,
    legacy_format=False,
)

second_level_input = data['cmaps']
design_matrix = pd.DataFrame(
    [1] * len(second_level_input),
    columns=['intercept'],
)

second_level_model = SecondLevelModel(smoothing_fwhm=8.0)
second_level_model = second_level_model.fit(
    second_level_input,
    design_matrix=design_matrix,
)

out_dict = non_parametric_inference(
    second_level_input,
    design_matrix=design_matrix,
    model_intercept=True,
    n_perm=500,  # 500 for the sake of time. Ideally, this should be 10,000.
    two_sided_test=True,
    smoothing_fwhm=8.0,
    n_jobs=8,
    threshold=0.001,
)
```

This fails with the old code. 

(I borrowed the example from:
https://nilearn.github.io/dev/auto_examples/05_glm_second_level/plot_second_level_one_sample_test.html)

ymzayek:
@jrdalenberg for some guidance you take a look at already written tests. Perhaps you can just add to one of them or copy one but with specific parameters to reproduce your failure.

For example see: https://github.com/nilearn/nilearn/blob/main/nilearn/mass_univariate/tests/test_permuted_least_squares.py#L408-L470
(Or if it needs to be tested on a higher level: https://github.com/nilearn/nilearn/blob/main/nilearn/glm/tests/test_second_level.py)

For tests we use fake generated data. There are examples in those test files you can copy. Let us know if you need any further help!

nilearn/mass_univariate/permuted_least_squares.py
tsalo: I think the diff would be cleaner if you just changed the threshold, but I don't feel too strongly about it. I don't know if changing the order of clusters in the labeled map (negative before positive, instead of the opposite) is something users will notice, so it's up to you.

jrdalenberg:
I updated the `test_cluster_level_parameters_smoke` function. The test needed a sufficient amount of clusters for it to fail on `two_sided_test=True`.

bthirion:
Can you check the Python 3.9 failure ? Seems related to the commit ...

jrdalenberg:
> Can you check the Python 3.9 failure ? Seems related to the commit ...

The test crashed on:
https://github.com/nilearn/nilearn/blob/4d3208a13062a908450fc5dc71905ae7ce893a64/nilearn/mass_univariate/permuted_least_squares.py#L921

The log file does not show the entire `target_vars` variable:

```python
2022-12-02T15:24:45.4949550Z target_vars = array([[ 0.2658434 ,  0.2658434 ,  0.2658434 , ...,  0.2658434 ,
2022-12-02T15:24:45.4949780Z          0.2658434 ,  0.        ],
2022-12-02T15:24:45.4950130Z        [ 0.0520910...405, -8.        ],
2022-12-02T15:24:45.4950350Z        [ 0.88161576,  0.88161576,  0.88161576, ...,  0.88161576,
2022-12-02T15:24:45.4950650Z          0.88161576, -9.        ]])
```

So it is hard to replicate what it used as input. 

I tried to troubleshoot it locally on a Macbook 2017 running macOS Montery (12.6), but the test runs fine on python 3.9 and 3.10. 

Any help would be very welcome.

ymzayek:
I'll have a look but I think this could be a random failure. @jrdalenberg see https://nilearn.github.io/stable/development.html#tests the section about seeding to see if you can improve the test.

ymzayek:
@jrdalenberg thanks for making an update. I'll do a thorough review of the test as soon as I can. I just want to quickly point out that the failure in the Ubuntu-latest Python 3.10 pre-release environment is not related to this PR so you don't need to worry about it.

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -595,19 +595,27 @@ def test_tfce_smoke(random_state=0):
 
 def test_cluster_level_parameters_smoke(random_state=0):
     """Test combinations of parameters related to cluster-level inference."""
+    import random
```
ymzayek, line 598:
```suggestion
```

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -595,19 +595,27 @@ def test_tfce_smoke(random_state=0):
...
     import nibabel as nib
     from nilearn.maskers import NiftiMasker
 
+    random.seed(random_state)
+    np.random.seed(random_state)
```
ymzayek, line 603:
```suggestion
    rng = np.random.RandomState(0)
```

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -595,19 +595,27 @@ def test_tfce_smoke(random_state=0):
...
+
     # create design
     target_var1 = np.arange(0, 10).reshape((-1, 1))  # positive effect
-    target_var = np.hstack((  # corresponds to 3 x 3 x 3 x 10 niimg
-        target_var1,  # voxel 1 has positive effect
-        - target_var1,  # voxel 2 has negative effect
-        np.random.random((10, 25)),  # 25 remaining voxels
+    voxel_vars = np.hstack((
```
ymzayek, line 607:
```suggestion
    voxel_vars = np.hstack((
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -903,15 +903,13 @@ def permuted_ols(
...
-                ] += n_positive_clusters
+                n_negative_clusters = np.max(temp_labeled_arr3d)
```
ymzayek, line 911:
```suggestion
                n_negative_clusters = np.max(temp_labeled_arr3d)
```

ymzayek (commented):
@jrdalenberg I made some suggestions below. Some are just PEP8 related fixes. For the test I think it's best to stick to one generator so I replaced Python's random and I use a numpy.random.RandomState() instance instead of the numpy.random.seed function

nilearn/mass_univariate/tests/test_permuted_least_squares.py
jrdalenberg: Oh cool! I did not know about this and that the Python stdlib module “random” is aware of this.

Q: Shouldn't this state be inherited from the function input `random_state` instead of hard coding it to 0?

ymzayek: @jrdalenberg yes my mistake,  it should take `random_state` as an argument

ymzayek (commented):
Thanks @jrdalenberg for the timely changes. Something a bit funny happened in my last review and not all my suggestions seem to have been recorded. Below are mainly PEP8 fixes. Otherwise LGTM! 
As a first-time contributor can you add your name to `doc/changes/names.rst` and then add a whatsnew entry to `doc/changes/latest.rst`?

ymzayek:
@jrdalenberg can you rebase on main and resolve the conflict? Then it should be ready to merge. Thanks!

jrdalenberg:
@ymzayek done :)

ymzayek (approved):
@jrdalenberg thanks! I think this one can be merged. The test failures are not related to this PR. They are the same as reported in #3441

ymzayek:
Merging

bthirion:
Thx !

PR #3439, opened 2022-12-01
[EHN] allow binary mask passed to `sample_mask`

### LINKED ISSUE #3204: Pass binary mask to `sample_masks` instead of indices
> Just wondering: wouldn't it be more handy to give  a binary vector with length `n_scans` for  `sample_mask` ?

_Originally posted by @bthirion in https://github.com/nilearn/nilearn/pull/3193#discussion_r838897961_

I did some archaeology and found out why I implemented index, rather than binary vector. I think it was for staying consistent to the OG `sample_mask`:

>I had a look at the relevant code of `sample_mask` application for resolving #2777. Currently, the timeseries censoring is done in the `base_masker` module:
https://github.com/nilearn/nilearn/blob/4cdd8fe8aad085d21e5d8367ad5b94d8efbe5151/nilearn/input_data/base_masker.py#L71-L73

_Originally posted by @htwangtw in https://github.com/nilearn/nilearn/issues/1012#issuecomment-847064092_

And the old doc of `sample_mask`:
https://github.com/nilearn/nilearn/blob/4cdd8fe8aad085d21e5d8367ad5b94d8efbe5151/nilearn/input_data/nifti_masker.py#L172-L177

Here's all the back ground info I can find. We can change it to binary vector as we no longer use the `index_img` function, and is using `signal.clean` to handle things. Let me know how you think we should move forward, given these info. 

cc @pbellec as I know you raised the same question before.

htwangtw:
Closes #3204.

- Allow binary mask for `sample_mask`
- Change the doc string related to sample_mask to reflect the change

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -28,7 +28,7 @@ Enhancements
 - :func:`~signal.clean` imputes scrubbed volumes (defined through ``sample_masks``) with cubic spline function before applying butterworth filter (:gh:`3385` by `Hao-Ting Wang`_). 
 - As part of making the User Guide more user-friendly, the introduction was reworked (:gh:`3380` by `Alexis Thual`_)
 - Added instructions for maintainers to make sure LaTeX dependencies are installed before building and deploying the stable docs (:gh:`3426` by `Yasmin Mzayek`_).
-- Parameter ``sample_masks`` in :func:`~signal.clean` and masker functions accept binary mask (:gh:`3439 by `Hao-Ting Wang`_). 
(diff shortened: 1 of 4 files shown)


bthirion (commented):
LGTM so far, please ass a whatsnew entry.

ymzayek (commented):
Thanks @htwangtw LGTM, just one small missed tick and should be good to go!

htwangtw:
@ymzayek this is ready! Sorry to let you waiting on such a small change...

ymzayek (approved):
No problem thanks!

bthirion (approved):
LGTM, thx !

PR #3453, opened 2023-01-03
[FIX] Resolve CanICA test failures

### LINKED ISSUE #3441: Precision error in CanICA test in pre-release environment
The test `nilearn/decomposition/tests/test_canica.py::test_canica_square_img` fails due to a precision error in ubuntu-latest (`ubuntu-22.04`) pre-release environment: https://github.com/nilearn/nilearn/actions/runs/3577698793/jobs/6017001262

See also: https://github.com/nilearn/nilearn/pull/3430#issuecomment-1331867028

We can look into improving the test.

ymzayek:
Closes #3441 .

- Add and tweak noise parameter for making canica test data

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -22,6 +22,7 @@ Fixes
 - Fix display of colorbar in matrix plots. Colorbar was overlapping in :func:`~matplotlib.pyplot.subplots` due to a hardcoded adjustment value in the subplot (:gh:`3403` by `Raphael Meudec`_).
 - Pass values with correct type to scikit-learn estimator parameters and remove deprecated parameter (:gh:`3430` by `Yasmin Mzayek`_).
 - Fix cluster labeling for two-sided cluster level permutation tests in :func:`~mass_univariate.permuted_ols` and associated function test ``test_cluster_level_parameters_smoke`` (:gh:`3436` by `Jelle Roelof Dalenberg`_).
+- Fix failing CanICA test by adding noise to the data creation step (:gh:`3453` by `Yasmin Mzayek`_).
 
 Enhancements
 ------------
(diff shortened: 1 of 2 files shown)


bthirion (commented):
Is it possible to narrow down to 1 change that fixes the failure ?

bthirion:
& thx for fixing this !

ymzayek:
@bthirion I believe I can remove the warnings fixes but when setting the noise parameters to True without tweaking the noise value, the coefficients that are supposed to be close to 1 end up being less than the .9 threshold and the test fails because of the relevant assertion.

nilearn/decomposition/canica.py
```diff
@@ -206,8 +206,8 @@ def _unmix_components(self, components):
         # Note: fastICA is very unstable, hence we use 64bit on it
         results = Parallel(n_jobs=self.n_jobs, verbose=self.verbose)(
             delayed(self._cache(fastica, func_memory_level=2))
-            (components.astype(np.float64), whiten=True, fun='cube',
-             random_state=seed)
+            (components.astype(np.float64), whiten='arbitrary-variance',
```
bthirion, line 209: The whitening change also maters for getting rid of the error ?

ymzayek: No but using `True` as the value is deprecated and is equivalent to this argument. However, if it's confusing I'll remove the commit handling the warnings since it's the case that these changes are unrelated to the PR and can be handled in another.

ymzayek: Ok I got rid of the 2 changes in the first commit. The tests should still pass.

bthirion (approved):
LGTM. Thx for squeezing the changes

NicolasGensollen (approved):
Thanks @ymzayek !

ymzayek:
Merging

PR #3455, opened 2023-01-05
[MAINT] Update deprecated Ubuntu 18.04 Actions runner

ymzayek:
Noticed because of brownout leading to cancelled tests in #3453 
e.g. https://dev.azure.com/Parietal/Nilearn/_build/results?buildId=5245&view=results

For details see: https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -53,7 +53,7 @@ jobs:
 
 - job: 'test_installing_on_linux'
   pool:
-    vmImage: 'ubuntu-18.04'
+    vmImage: 'ubuntu-20.04'
   strategy:
     matrix:
       Python38:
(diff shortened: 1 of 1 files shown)


NicolasGensollen (approved):
LGTM, thanks @ymzayek !

bthirion (approved):
LGTM, thx !

jeromedockes:
LGTM, thanks! as a side note in the future if we ever want to test on older distributions than those provided we can easily run jobs in a container -- docs for [azure pipelines](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops#single-job) and for [gh actions](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container)

ymzayek:
@jeromedockes good point thanks! Just to double check, for these tests it didn't seem to me like they specifically needed to run on ubuntu 18.04. Correct me if I'm wrong.

jeromedockes:
> @jeromedockes good point thanks! Just to double check, for these tests it didn't seem to me like they specifically needed to run on ubuntu 18.04. Correct me if I'm wrong.

you're right!

ymzayek:
Merging. Failures are unrelated

PR #3458, opened 2023-01-10
[FIX] Nibabel Nifti1Image class now requires `header` or `dtype` arguments for int64 data

ymzayek:
Resolves failures on tests using dtype int64 data arrays. See https://github.com/nilearn/nilearn/actions/runs/3881843135/
From nibabel changelog: Passing an int64 array to [Nifti1Image](https://nipy.org/nibabel/reference/nibabel.nifti1.html#nibabel.nifti1.Nifti1Image) without a header or dtype argument will raise a `ValueError`

### DIFF
--- a/nilearn/tests/test_niimg.py
+++ b/nilearn/tests/test_niimg.py
@@ -52,7 +52,7 @@ def test_get_target_dtype():
 
     hdr = Nifti1Header()
     data = np.ones((2, 2, 2), dtype=np.int64)
-    img2 = Nifti1Image(data, dtype=np.int64, affine=np.eye(4), header=hdr)
+    img2 = Nifti1Image(data, affine=np.eye(4), header=hdr)
     assert get_data(img2).dtype.kind == 'i'
     dtype_kind_int = niimg._get_target_dtype(get_data(img2).dtype,
                                              target_dtype='auto')
--- a/nilearn/tests/test_niimg.py
+++ b/nilearn/tests/test_niimg.py
@@ -55,7 +55,7 @@ def test_get_target_dtype():
     hdr.set_data_dtype(np.int64)
     data = np.ones((2, 2, 2), dtype=np.int64)
     img2 = Nifti1Image(data, affine=np.eye(4), header=hdr)
-    assert get_data(img2).dtype.kind == 'i'
+    assert get_data(img2).dtype.kind == img2.get_data_dtype().kind == 'i'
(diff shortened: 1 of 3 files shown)


jeromedockes:
old nibabel versions don't have a `dtype` arguments for the Nifti1Image init. you can catch the TypeError and retry without `dtype` or provide a header

NicolasGensollen (approved):
LGTM, thanks for fixing the tests @ymzayek !

nilearn/tests/test_niimg.py
```diff
@@ -49,8 +49,11 @@ def test_get_target_dtype():
     dtype_kind_float = niimg._get_target_dtype(get_data(img).dtype,
                                                target_dtype='auto')
     assert dtype_kind_float == np.float32
-
-    img2 = Nifti1Image(np.ones((2, 2, 2), dtype=np.int64), affine=np.eye(4))
+    # Passing dtype or header is required when using int64
+    # https://nipy.org/nibabel/changelog.html#api-changes-and-deprecations
+    hdr = Nifti1Header()
```
effigies, line 54: Do you mean for these to be eventually output as float32?

```Python
>>> import nibabel as nb
>>> nb.Nifti1Header().get_data_dtype()
dtype('<f4')
```

If you want to record the intended dtype, you need to do `hdr.set_data_dtype()`.

effigies: For this sort of test, it might be useful to round-trip images to see what happens when serialized. (You might have a helper function already...)

```Python
def rt_img(img):
    return img.from_bytes(img.to_bytes())
```

ymzayek: Hi @effigies thanks for pointing that out. You are right I should add a line with `hdr.set_data_dtype(np.int64)`. 
I can also add the check with the round trip function. I'm not sure if we actually have such a function but might be useful. Also in this case would it be enough to replace L57 below:

`assert get_data(img2).dtype.kind == 'i'`

with

`assert img2.get_data_dtype().kind == 'i'`

ymzayek:
Merging this after a discussion during the office hours. These tests just test handling of 64-bit integer data arrays without crashing (see https://nilearn.github.io/stable/changes/whats_new.html#id127). It is still not recommended to use these data arrays and will results in a `ValueError` from nibabel when passed without a header or dtype argument to Nifti1Image.

bthirion:
Thx

PR #3460, opened 2023-01-11
[INFRA] Remove deprecated check_skip job

### LINKED ISSUE #3448: Remove ci-skip action from GitHub Actions workflows
GitHub Actions now supports skipping workflows out of the box: https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs and so the Action mstachniuk/ci-skip is deprecated and should be removed from all workflows before it starts failing.

ymzayek:
Closes #3448.

Remove deprecated job that is now automatically supported by GitHub Actions

### DIFF
--- a/.github/workflows/pep8.yml
+++ b/.github/workflows/pep8.yml
@@ -9,23 +9,7 @@ on:
       - '*'
 
 jobs:
-  check_skip:
-    runs-on: ubuntu-latest
-    outputs:
-      skip: ${{ steps.result_step.outputs.ci-skip }}
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - id: result_step
-        uses: mstachniuk/ci-skip@master
-        with:
-          commit-filter: '[skip ci];[ci skip];[skip github]'
-          commit-filter-separator: ';'
-
   pep8:
-    needs: check_skip
-    if: ${{ needs.check_skip.outputs.skip == 'false' }}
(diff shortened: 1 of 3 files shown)


bthirion:
Thx !
We'll probably have to merge #3458 first; but it LGTM.

.github/workflows/pep8.yml
```diff
@@ -9,23 +9,7 @@ on:
...
-    outputs:
-      skip: ${{ steps.result_step.outputs.ci-skip }}
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - id: result_step
-        uses: mstachniuk/ci-skip@master
-        with:
-          commit-filter: '[skip ci];[ci skip];[skip github]'
```
NicolasGensollen, line 23: I think the tag `[skip github]` won't work anymore, right ?
I don't think this is very costly though, considered that we are getting rid of a lot of code as a reward, and I don't think anybody was using it...

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

.github/workflows/pep8.yml
ymzayek: Yes that's true but I agree that it is not worth to keep all this code. And at some point this action will start failing. I think it's enough that we give the example of `[skip ci]` in the readme as well as a link to the other tags that work.

bthirion (approved):
Looks great, thx !

Issue #3464, opened 2023-01-14
Function `non_parametric_inference` ignores any covariates


jrdalenberg:
The function [`non_parametric_inference`](https://github.com/nilearn/nilearn/blob/b556cd1a293e0f8396ea854b9a73214a8bbd1262/nilearn/glm/second_level/second_level.py#L593) extracts [a target contrast from the input design matrix](https://github.com/nilearn/nilearn/blob/b556cd1a293e0f8396ea854b9a73214a8bbd1262/nilearn/glm/second_level/second_level.py#L848) to perform a non-parametric test using the function [`permuted_ols`](https://github.com/nilearn/nilearn/blob/b556cd1a293e0f8396ea854b9a73214a8bbd1262/nilearn/glm/second_level/second_level.py#L854). However, any covariates/confounding variates are ignored here. So, currently, users are unable to perform non-parametric inference tests in the context of covariates.

Also see NeuroStars discussion: https://neurostars.org/t/dof-in-nilearns-non-parametric-inference/24416

bthirion:
Thx !

PR #3465, opened 2023-01-14
[FIX] Add covariates to non-parametric inference

jrdalenberg:
This PR addresses #3464 and adds covariates to non-parametric inference.

Edit: and some other intercept-stuff, lol

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,6 +13,10 @@ Fixes
 
 - Restore :func:`~image.resample_img` compatibility with all :class:`nibabel.spatialimages.SpatialImage` objects (:gh:`3462` by `Mathias Goncalves`_).
 
+- :func:`~nilearn.glm.second_level.non_parametric_inference` now supports confounding variates when they are available in the input design matrix :func:`~nilearn.mass_univariate.permuted_ols` (:gh:`3465` by `Jelle Roelof Dalenberg`_).
+
+- :func:`~nilearn.mass_univariate.permuted_ols` now checks if confounding variates contain an intercept and raises an error when multiple intercepts are defined across target and confounding variates. (:gh:`3465` by `Jelle Roelof Dalenberg`_).
+
 Enhancements
 ------------
 
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -15,7 +15,7 @@ Fixes
 
 - :func:`~nilearn.glm.second_level.non_parametric_inference` now supports confounding variates when they are available in the input design matrix :func:`~nilearn.mass_univariate.permuted_ols` (:gh:`3465` by `Jelle Roelof Dalenberg`_).
 
-- :func:`~nilearn.mass_univariate.permuted_ols` now checks if confounding variates contain an intercept and raises an error when multiple intercepts are defined across target and confounding variates. (:gh:`3465` by `Jelle Roelof Dalenberg`_).
+- :func:`~nilearn.mass_univariate.permuted_ols` now checks if confounding variates contain an intercept and raises an warning when multiple intercepts are defined across target and confounding variates. (:gh:`3465` by `Jelle Roelof Dalenberg`_).
 
 Enhancements
 ------------
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -15,7 +15,7 @@ Fixes
 
 - :func:`~nilearn.glm.second_level.non_parametric_inference` now supports confounding variates when they are available in the input design matrix :func:`~nilearn.mass_univariate.permuted_ols` (:gh:`3465` by `Jelle Roelof Dalenberg`_).
 
-- :func:`~nilearn.mass_univariate.permuted_ols` now checks if confounding variates contain an intercept and raises an warning when multiple intercepts are defined across target and confounding variates. (:gh:`3465` by `Jelle Roelof Dalenberg`_).
+- :func:`~nilearn.mass_univariate.permuted_ols` now checks if confounding variates contain a intercept and raises an warning when multiple intercepts are defined across target and confounding variates (:gh:`3465` by `Jelle Roelof Dalenberg`_).
 
 Enhancements
 ------------
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -847,10 +847,9 @@ def non_parametric_inference(
     var_names = design_matrix.columns.tolist()
 
     # Obtain tested_var
-    if contrast in var_names:
-        tested_var = np.asarray(design_matrix[contrast])
-        # Remove tested var from remaining var names
-        var_names.remove(contrast)
+    tested_var = np.asarray(design_matrix[contrast])
+    # Remove tested var from remaining var names
+    var_names.remove(contrast)
 
     # Obtain confounding vars
     if len(var_names) == 0:
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -858,7 +858,7 @@ def non_parametric_inference(
     else:
         # Exclude the intercept (constant)
         for var_name in var_names:
-            if np.unique(np.asarray(design_matrix[var_name])).size == 1:
+            if np.unique(design_matrix[var_name]).size == 1:
                 var_names.remove(var_name)
         
         confounding_vars = np.asarray(design_matrix[var_names])
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -843,9 +843,25 @@ def non_parametric_inference(
     # Check design matrix and effect maps agree on number of rows
     _check_effect_maps(effect_maps, design_matrix)
 
+    # Obtain design matrix vars
+    var_names = design_matrix.columns.tolist()
+
     # Obtain tested_var
-    if contrast in design_matrix.columns.tolist():
+    if contrast in var_names:
         tested_var = np.asarray(design_matrix[contrast])
+        var_names.remove(contrast)
+
+    # Obtain confounding vars
+    if len(var_names) == 0:
+        # No other vars in design matrix
+        confounding_vars = None
+    else:
+        # Exclude the intercept (constant)
+        for var_name in var_names:
+            if np.unique(np.asarray(design_matrix[var_name])).size == 1:
+                var_names.remove(var_name)
+        
+        confounding_vars = np.asarray(design_matrix[var_names])
 
     # Mask data
     target_vars = masker.transform(effect_maps)
@@ -854,6 +870,7 @@ def non_parametric_inference(
     outputs = permuted_ols(
         tested_var,
         target_vars,
+        confounding_vars=confounding_vars,
         model_intercept=model_intercept,
         n_perm=n_perm,
         two_sided_test=two_sided_test,
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -860,7 +860,6 @@ def non_parametric_inference(
         for var_name in var_names:
             if np.unique(design_matrix[var_name]).size == 1:
                 var_names.remove(var_name)
-        
         confounding_vars = np.asarray(design_matrix[var_names])
 
     # Mask data
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -857,11 +857,6 @@ def non_parametric_inference(
         # No other vars in design matrix
         confounding_vars = None
     else:
-        # Exclude the intercept (constant)
-        for var_name in var_names:
-            if np.unique(design_matrix[var_name]).size == 1:
-                # Remove intercept from remaining var names
-                var_names.remove(var_name)
         # Use remaining vars as confounding vars
         confounding_vars = np.asarray(design_matrix[var_names])
 
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -849,6 +849,7 @@ def non_parametric_inference(
     # Obtain tested_var
     if contrast in var_names:
         tested_var = np.asarray(design_matrix[contrast])
+        # Remove tested var from remaining var names
         var_names.remove(contrast)
 
     # Obtain confounding vars
@@ -859,7 +860,9 @@ def non_parametric_inference(
         # Exclude the intercept (constant)
         for var_name in var_names:
             if np.unique(design_matrix[var_name]).size == 1:
+                # Remove intercept from remaining var names
                 var_names.remove(var_name)
+        # Use remaining vars as confounding vars
         confounding_vars = np.asarray(design_matrix[var_names])
 
     # Mask data
--- a/nilearn/glm/tests/test_second_level.py
+++ b/nilearn/glm/tests/test_second_level.py
@@ -662,7 +662,7 @@ def test_non_parametric_inference_cluster_level_with_covariates(random_state=0):
         # Set up one sample t-test design with two random covariates
         cov1 = rng.random(n_subjects)
         cov2 = rng.random(n_subjects)
-        X = pd.DataFrame({"cov1": cov1, "cov2": cov2, 'intercept': 1})
+        X = pd.DataFrame({"cov1": cov1, "cov2": cov2, "intercept": 1})
 
         # make sure there is variability in the images
         kernels = rng.uniform(low=0, high=5, size=n_subjects)
--- a/nilearn/glm/tests/test_second_level.py
+++ b/nilearn/glm/tests/test_second_level.py
@@ -678,7 +678,7 @@ def test_non_parametric_inference_cluster_level_with_covariates(random_state=0):
         cov2 = rng.random(n_subjects) 
         X = pd.DataFrame({"cov1": cov1, "cov2": cov2, 'intercept': 1})
 
-        # make sure there is variablity in the images
+        # make sure there is variability in the images
         kernels = rng.uniform(low=0, high=5, size=n_subjects)
         Y = [smooth_img(func_img, kernel) for kernel in kernels]
 
--- a/nilearn/glm/tests/test_second_level.py
+++ b/nilearn/glm/tests/test_second_level.py
@@ -639,6 +639,77 @@ def test_non_parametric_inference_cluster_level():
         del func_img, FUNCFILE, out, X, Y
 
 
+def test_non_parametric_inference_cluster_level_with_covariates(random_state=0):
+    """Test non-parametric inference with cluster-level inference in the context of covariates."""
+    from nilearn.glm.second_level import make_second_level_design_matrix
+    from nilearn.image import new_img_like, smooth_img
+    from nilearn.reporting import get_clusters_table
+    from unittest import TestCase
+    from scipy import stats
+
+    class TestClusters(TestCase):
+        """Unit test for cluster sizes"""
+        def setUp(self, expected, result):
+            self.expected = expected
+            self.result = result
+
+        def test_count_eq(self):
+            self.assertCountEqual(self.result, self.expected, msg="Unexpected cluster sizes found. Check covariates and/or degrees of freedom")
+
+    def t_to_neg_log_pval(t_scores, df):
+        """Helper function that calculates negative log p-values from a set of t-values and degrees of freedom."""
+        with np.errstate(divide='ignore'):
+            neg_log_pval = -np.log10(stats.t.sf(t_scores, df=df))
+        return neg_log_pval
+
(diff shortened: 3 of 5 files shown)


jrdalenberg:
I checked this with several second level contrasts / design matrices. Could anyone also confirm that this works for first level contrasts?

nilearn/glm/second_level/second_level.py
```diff
@@ -843,9 +843,25 @@ def non_parametric_inference(
...
+    if contrast in var_names:
         tested_var = np.asarray(design_matrix[contrast])
+        var_names.remove(contrast)
+
+    # Obtain confounding vars
+    if len(var_names) == 0:
+        # No other vars in design matrix
+        confounding_vars = None
+    else:
+        # Exclude the intercept (constant)
```
bthirion, line 859: I think that you need to explicitly need to remove the 'intercept' term ---if it's there, no ? Or test id the corresponding regressor is constant.

bthirion (commented):
Thx for fixing it ! 
There remain 2 things to do for this PR:
* add a test that fails with the current main version of the library, but works with the new one, to ensure that the correct behavior is now enforced.
* add a whatsnew entry

nilearn/glm/second_level/second_level.py
jrdalenberg: Do you mean, "I don't think..."? 

If I do not remove it, and a user leaves the parameter `model_intercept=True`, the function `permuted_ols` will effectively add an intercept and also regress an intercept out of the data. This results in removing an extra degree of freedom. 

`permuted_ols` only checks if there is already an intercept in parameter `tested_vars` and not in `confounding_vars`. 

That's why I explicitly removed the intercept from the `confounding_vars`, if it's in the design matrix.

bthirion: I indeed think that we want to remove it. the point is that you are implicitly relying on the convention that if there is a unique regressor left, this is an intercept. I think that we should test
```if 'intercept' in var_names: ```

**Comment by jrdalenberg** [comment]
Regarding the test. 

I could 
- Add an extra integer as output in `permuted_ols` that indicates how many DF were used in the analysis. A test could than be added that cross-references the output DF with the input DF. 
- Alternatively, I could calculate uncorrected cluster sizes and build a test that makes sure that cluster-sizes from the clusterwise error correction are the same. 

To me neither is ideal. For the first I need to adjust the `permuted_ols` outputs and I cannot oversee if there are other functions that run into issues then. The latter is not great either because cluster sizes do not necessarily have to differ if the results with and without covariates hardly differ. 

Does anyone have any (other) suggestions?

File: nilearn/glm/second_level/second_level.py (hunk 1 above)
Comment (jrdalenberg) [reply nilearn/glm/second_level/second_level.py:859]: Is the intercept always uniquely termed 'intercept'? In other words, users can't build design matrices where they give it an arbitrary name? (I haven't checked.)

Comment (ymzayek) [reply nilearn/glm/second_level/second_level.py:859]: To my knowledge there is no constraint on intercept name. I think this is an okay way to test if the regressor is constant.

File: nilearn/glm/second_level/second_level.py [hunk 2]
```diff
@@ -843,9 +843,25 @@ def non_parametric_inference(
...
+        for var_name in var_names:
+            if np.unique(np.asarray(design_matrix[var_name])).size == 1:
```
ymzayek, line 861:
```suggestion
            if np.unique(design_matrix[var_name]).size == 1:
```
I don't think you need that

ymzayek (commented):
@jrdalenberg thx for this! I just had a quick look. I will get back to you about the tests.

bthirion:
> 

I would consider the second solution. I don't think we should add the output of `permuted_ols`.
Thx again !

jrdalenberg:
> I would consider the second solution. I don't think we should add the output of `permuted_ols`. Thx again !
Took me a while to figure out how to best do that. Let me know what you think! 

This test fails on the old code because of the DOF penalty, while it works with the new code.

nilearn/glm/second_level/second_level.py
```diff
@@ -843,9 +843,24 @@ def non_parametric_inference(
     # Check design matrix and effect maps agree on number of rows
     _check_effect_maps(effect_maps, design_matrix)
 
+    # Obtain design matrix vars
+    var_names = design_matrix.columns.tolist()
```
tsalo, line 847: Is a `design_matrix` guaranteed? I think it's a mutually-exclusive parameter with `confounds`, so users could provide `confounds` and `design_matrix` should be None.

tsalo: Oh sorry, if this _is_ a bug, it's not one you've introduced, so you can ignore my comment.

nilearn/glm/second_level/second_level.py
```diff
@@ -843,9 +843,24 @@ def non_parametric_inference(
...
+    # Obtain confounding vars
+    if len(var_names) == 0:
+        # No other vars in design matrix
+        confounding_vars = None
+    else:
+        # Exclude the intercept (constant)
+        for var_name in var_names:
+            if np.unique(design_matrix[var_name]).size == 1:
+                var_names.remove(var_name)
+        confounding_vars = np.asarray(design_matrix[var_names])
```
bthirion, line 863: I think it is important here to comment what is done.

nilearn/glm/tests/test_second_level.py
```diff
@@ -639,6 +639,77 @@ def test_non_parametric_inference_cluster_level():
...
+    from scipy import stats
+
+    class TestClusters(TestCase):
+        """Unit test for cluster sizes"""
+        def setUp(self, expected, result):
+            self.expected = expected
+            self.result = result
+
+        def test_count_eq(self):
+            self.assertCountEqual(self.result, self.expected, msg="Unexpected cluster sizes found. Check covariates and/or degrees of freedom")
```
bthirion, line 657: Please break the lines to less than 80 characters

bthirion (commented):
Thx !  we 're getting close !

nilearn/glm/second_level/second_level.py
ymzayek: On second thought could it be the case that `model_intercept` is set to False because a constant term is included and that this would remove it without the user knowing?

I'm wondering if it's better to add the check to the `permuted_ols` function here:
https://github.com/nilearn/nilearn/blob/9c29e3376243b643a72f78b8b98a56d12c2e99bb/nilearn/mass_univariate/permuted_least_squares.py#L696-L702

ymzayek (commented):
@jrdalenberg I think the test can be simplified. I made some suggestions below. Also see the comment about removing the intercept from `confounding_vars`

jrdalenberg:
Python 3.7, minimal requirements without Matplotlib is trying to import Matplotlib.

ymzayek:
@jrdalenberg hm ok I see that that is because importing `get_clusters_table` eventually calls to import matplotlib and in this build workflow any tests that require matplotlib should be skipped. To me it's not very clear how this skipping is configured and why it's not working in this case but I see that in previous runs the test is skipped successfully when the import is under the test function. So for now just move that import `from nilearn.reporting import get_clusters_table` back to where you originally had it and we can see if the test is skipped successfully.

jrdalenberg:
@ymzayek yep that fixes it.

jrdalenberg:
Let me know if the the intercept handling needs to change or if there are any other things. 

If not, ill add a whatsnew entry.

ymzayek:
> Let me know if the the intercept handling needs to change or if there are any other things.

https://github.com/nilearn/nilearn/pull/3465#discussion_r1073362883
@jrdalenberg if my logic is correct in this comment then I would move it is as suggested. Also it would be in line with the part of the code checking if `tested_vars` is an intercept. Do you see any reason why it should necessarily be under `non_parametric_inference`? If not, let's move it. In any case, that condition still needs to be tested. That's why the [codecov patch test](https://github.com/nilearn/nilearn/pull/3465/checks?check_run_id=10818089818) is failing. @bthirion @tsalo WDYT?

nilearn/glm/second_level/second_level.py
tsalo: I think adding the check to `permuted_ols` would make the most sense, since the same problem could come up there as well.

jrdalenberg:
@ymzayek @tsalo 

Alright, I moved the check to `permuted_ols`. I formulated the check a little broader than just focusing on an intercept since there is a possibility that users add one or more constant columns by mistake.

jrdalenberg:
Hmm the nilearn.nilearn timeout error seems like an anomaly. 

https://dev.azure.com/Parietal/3cc37faf-0dd1-4dff-b683-062a6cadcc03/_apis/build/builds/5294/logs/64

There are also a lot of errors regarding downloading content from neurovault.

bthirion (commented):
This looks good. 
A couple of lines remain untested: can you make sure that they are covered ?
Thx again !

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -693,6 +693,20 @@ def permuted_ols(
...
+    # check if confounding vars contain constants
+    constants = []
+    if confounding_vars is not None:
+        for column in range(confounding_vars.shape[1]):
+            if np.unique(confounding_vars[:, column]).size == 1:
+                constants.append(column)
+        if len(constants) > 0:
+            warnings.warn(
+            'One or more "confounding_vars" are constant and will be ignored. '
+            'Tip: Use "model_intercept" (default=True) for intercept control.'
```
ymzayek, line 705: Regarding the way this check is implemented, if `confounding_vars` includes an intercept term users would get the warning message whether or not `model_intercept` is True or False. So then the tip doesn't really make sense to me. Maybe it's just a matter of improving the warning.

Alternatively, if it's too difficult to control for all the possibilities we can also raise a warning to the user to check their input if `intercept_test` is a constant and `confounding_vars` has a constant term and not remove anything from `confounding_vars`. This makes it their responsibility to make sure their parameters makes sense. This is just a suggestion. Maybe there are better ways to do this.

bthirion: We should by all means avoid having duplicate regressors.

jrdalenberg:
lol @ `assert logp_unc_cluster_sizes.sort() == logp_max_cluster_sizes.sort()`

This compares the sort-function and not the actual lists -_-

It correctly fails now @ current main branch.

jrdalenberg:
I’m not completely there yet. I think that  there needs to be a check as well in non_parameteic_inference. Since it’s focused around design matrices, and should warn a user if model_intercept is set to True while there is a constant somewhere in the matrix.

nilearn/mass_univariate/permuted_least_squares.py
ymzayek: Thanks @jrdalenberg for updating this check and the warning, it looks good now!

jrdalenberg:
Ok, I made it a bit simpler by expanding the intercept check to the confounding vars, and added tests for it. I removed the warning and added a value error when multiple intercepts are detected. This should take care of mixing `model_intercept` with an intercept in the design matrix.

bthirion (approved):
I think we're good.

nilearn/glm/tests/test_second_level.py
```diff
@@ -576,7 +577,8 @@ def test_non_parametric_inference_permutation_computation():
         Y = [func_img] * 4
         X = pd.DataFrame([[1]] * 4, columns=['intercept'])
 
-        neg_log_pvals_img = non_parametric_inference(Y, design_matrix=X,
+        neg_log_pvals_img = non_parametric_inference(Y, design_matrix=X,
```
ymzayek, line 580:
```suggestion
        neg_log_pvals_img = non_parametric_inference(Y, design_matrix=X,
```

nilearn/glm/tests/test_second_level.py
```diff
@@ -639,6 +643,63 @@ def test_non_parametric_inference_cluster_level():
         del func_img, FUNCFILE, out, X, Y
 
 
+def test_non_parametric_inference_cluster_level_with_covariates(random_state=0):
```
ymzayek, line 646:
```suggestion
def test_non_parametric_inference_cluster_level_with_covariates(
        random_state=0
):
```

nilearn/glm/tests/test_second_level.py
```diff
@@ -639,6 +643,63 @@ def test_non_parametric_inference_cluster_level():
...
+        Y = [smooth_img(func_img, kernel) for kernel in kernels]
+
+        # Set up non-parametric test
+        out = non_parametric_inference(
+            Y,
+            design_matrix=X,
+            mask=mask,
+            model_intercept=False,
+            second_level_contrast="intercept",
+            n_perm=1/unc_pval,
```
ymzayek, line 678:
```suggestion
            n_perm=1 / unc_pval,
```

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -687,12 +687,25 @@ def permuted_ols(
...
+    # check if confounding vars contains an intercept
+    if confounding_vars is not None:
+        constants = []
+        for column in range(confounding_vars.shape[1]):
+            if np.unique(confounding_vars[:, column]).size == 1:
+                constants.append(column)
+        if (intercept_test and len(constants) == 1) or len(constants) > 1:
+            raise ValueError('More than one intercepts '
+                            '(i.e. constant variables) are defined, '
+                            'check inputs')
```
ymzayek, line 705:
```suggestion
            raise ValueError('More than one intercepts '
                             '(i.e. constant variables) are defined, '
                             'check inputs')
```

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -287,6 +287,36 @@ def test_permuted_ols_withcovar(random_state=0):
     assert_array_almost_equal(alt_score_intercept, own_score_intercept,
                               decimal=6)
 
+    ### Intercept in confounding vars
```
ymzayek, line 290:
```suggestion
    # Intercept in confounding vars
```

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -287,6 +287,36 @@ def test_permuted_ols_withcovar(random_state=0):
...
+    assert own_score.shape == (n_regressors, n_descriptors)
+
+    # permuted OLS with constant in confounding_vars, model_intercept=False
+    confounding_vars = np.ones([n_samples, 1])
+    _, own_score, _ = permuted_ols(
+        tested_var, target_var, confounding_vars, model_intercept=False,
+        n_perm=0, random_state=random_state)
+    assert own_score.shape == (n_regressors, n_descriptors)
+
+    ### Multiple intercepts should raise errors
```
ymzayek, line 305:
```suggestion
    # Multiple intercepts should raise errors
```

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -287,6 +287,36 @@ def test_permuted_ols_withcovar(random_state=0):
...
+            tested_var, target_var, confounding_vars,
+            n_perm=0, random_state=random_state)
+
+    # Across tested vars and confounding vars
+    with pytest.raises(ValueError):
+        confounding_vars = np.ones([n_samples, 1])
+        tested_var = np.ones([n_samples, 1])
+        _, own_score, _ = permuted_ols(
+            tested_var, target_var, confounding_vars,
+            n_perm=0, random_state=random_state)
```
ymzayek, line 319:
```suggestion
            n_perm=0, random_state=random_state)
            
```

ymzayek (commented):
Hi @jrdalenberg much better! Thanks for taking the time to make it more compact. There are some small PEP8 issues to fix as suggested below and you'll need to add a whatsnew entry. Other than that LGTM!

jrdalenberg:
Forced push changes:

- rebase to current main branch
- entry to the changes file
- small update to the `model_intercept`  parameter description of `permuted_ols`

ymzayek:
> I noticed this though: [009677f#diff-997d3dad0d51606b78f448aee14e5c2e8ffd65d2befbcd3831463b9b083aeaf9R691](https://github.com/nilearn/nilearn/commit/009677f0a6e9f18c40e787ee8a37e12015f8a37a#diff-997d3dad0d51606b78f448aee14e5c2e8ffd65d2befbcd3831463b9b083aeaf9R691)

Seems to me that `intercept_test` is set to false if there is more than one regressor regardless if any of the regressors are constant. If this can be handled better, it should be done in a separate PR as this PR should only focus on the linked issue and we don't want to get too far from that.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -687,12 +688,25 @@ def permuted_ols(
...
         intercept_test = False
 
+    # check if confounding vars contains an intercept
+    if confounding_vars is not None:
+        constants = []
+        for column in range(confounding_vars.shape[1]):
+            if np.unique(confounding_vars[:, column]).size == 1:
+                constants.append(column)
+        if (intercept_test and len(constants) == 1) or len(constants) > 1:
+            raise ValueError('More than one intercepts '
```
bthirion, line 704: In that  case, should we raise en error ? Why not rather spit a warning and remove an intercept ?

ymzayek: Perhaps even better, the 3rd line should be `if not model_intercept and not intercept_test:`

doc/changes/latest.rst
```diff
@@ -13,6 +13,10 @@ Fixes
 
 - Restore :func:`~image.resample_img` compatibility with all :class:`nibabel.spatialimages.SpatialImage` objects (:gh:`3462` by `Mathias Goncalves`_).
 
+- :func:`~nilearn.glm.second_level.non_parametric_inference` now supports confounding variates when they are available in the input design matrix :func:`~nilearn.mass_univariate.permuted_ols` (:gh:`3465` by `Jelle Roelof Dalenberg`_).
+
+- :func:`~nilearn.mass_univariate.permuted_ols` now checks if confounding variates contain an intercept and raises an error when multiple intercepts are defined across target and confounding variates. (:gh:`3465` by `Jelle Roelof Dalenberg`_).
```
bthirion, line 18:
```suggestion
- :func:`~nilearn.mass_univariate.permuted_ols` now checks if confounding variates contain an intercept and raises an warning when multiple intercepts are defined across target and confounding variates. (:gh:`3465` by `Jelle Roelof Dalenberg`_).
```

jrdalenberg:
Changes: rebase to current main.

ymzayek:
Hi @jrdalenberg something seems to have gone wrong with the rebase and the changes from the last 2 merged PRs now appear in this PR. I suggest to do a hard reset in your branch to the commit [Update doc/changes/latest.rst](https://github.com/nilearn/nilearn/pull/3465/commits/597b9ee40149670c53b5ebfd61b5407e3ef1ffd3) so `git reset --hard 597b9ee4` . Then check out main and pull upstream main if you haven't already. Then checkout your branch again and run `git rebase main`. At this point you can resolve the conflicts if any, add changed files to stage, and continue the rebase. Then force push the changes.

jrdalenberg:
Ah, thanks @ymzayek. I already thought something went wrong. Ill get it fixed.

jrdalenberg:
```Added line #L861 was not covered by tests```
Weirdly enough, that line WAS covered but #L858 was not.

ymzayek:
Indeed @jrdalenberg I don't know why it reported that.
I think the codecov patch failure is coming from partial coverage of L850 in `nilearn/glm/second_level/second_level.py` and L721 of `nilearn/mass_univariate/permuted_least_squares.py`. I think this means that those conditions are tested when True but not when they equal False. Let's add those and see if it resolves it.

jrdalenberg:
> Indeed @jrdalenberg I don't know why it reported that. I think the codecov patch failure is coming from partial coverage of L850 in `nilearn/glm/second_level/second_level.py` and L721 of `nilearn/mass_univariate/permuted_least_squares.py`. I think this means that those conditions are tested when True but not when they equal False. Let's add those and see if it resolves it.

Thank you for pointing this out! Another learning moment :)

ymzayek (approved):
LGTM @jrdalenberg! Thanks for sticking with this and finishing it :)

ymzayek:
Maybe a final review from @bthirion

jrdalenberg:
> LGTM @jrdalenberg! Thanks for sticking with this and finishing it :)

No worries, it's fun and educational, and I am happy to do it when I find some time here and there. Thank you for guiding me through it @ymzayek.

ymzayek:
Can we merge this one? @tsalo or @bthirion maybe a final review?

bthirion (approved):
LGTM, thx.

ymzayek:
Thx merging

PR #3472, opened 2023-01-20
[ENH] Deprecate and replace `plot_epi` function name with `plot_fmri`

### LINKED ISSUE #3388: Misleading name for plot_epi function
There is no specific reason to name the `plot_epi` function as such even though most k-space encoding schemes in fMRI
rely on Echo Planar Imaging. We're working on 3D SPARKLING k-space encoding schemes for fMRI and would like
to use this function too. Of course it's totally doable as this function is independent of the way k-space data has been
collected. Consequently we suggest renaming this function `plot_fmri` to make it usable with other encoding patterns.

### Benefits to the change
Doing so, users would be aware that is function just allows them to plot an fMRI image (for instance the mean of an fMRI sequence)
To preserve backward compatibility I suggest creating an alias to the existing `plot_epi` function

```python
from nilearn.plotting import plot_fmri
plot_fmri(mean_clock_retino, colorbar=True, cbar_tick_format="%i")
```

ymzayek:
Closes #3388 .

- Change name of `plot_epi` function to `plot_fmri`

TODO: 
- [ ] Add deprecation warning
- [ ] Add versionchanged directive
- [ ] Replace name where function is used
- [ ] Change parameter `epi_img` name?
- [ ] Adapt docstring

### DIFF
--- a/nilearn/plotting/__init__.py
+++ b/nilearn/plotting/__init__.py
@@ -46,7 +46,7 @@ _set_mpl_backend()
 ###############################################################################
 from . import cm
 from .img_plotting import (
-    plot_img, plot_anat, plot_epi, plot_roi, plot_stat_map,
+    plot_img, plot_anat, plot_fmri, plot_roi, plot_stat_map,
     plot_glass_brain, plot_connectome, plot_markers, plot_prob_atlas,
     plot_carpet, plot_img_comparison, show)
 from .find_cuts import find_xyz_cut_coords, find_cut_slices, \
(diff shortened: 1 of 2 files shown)


bthirion (commented):
Wait, we need to go through deprecation cycles...

ymzayek:
@bthirion Yes what I've implemented so far allows use of both function names and I plan to raise a warning that will say something like "function plot_epi is deprecated and will be replaced in release 0.13. Use function plot_fmri instead". So you'll have the ability to use both functions and you'll see this warning for 2 release cycles. Let me know if there is a preferred or better way to do this other than this.

bthirion:
You probably want to use a deprecation warning as e.g.  https://github.com/nilearn/nilearn/blob/main/nilearn/glm/first_level/first_level.py#L400

PR #3474, opened 2023-01-27
[FIX] Add corrected standardize strategy to `signal.clean`

### LINKED ISSUE #3406: signal clean should not use population standard deviation
Nilearn version: 0.9.2

In signal.clean, the calculation simply uses numpy.std, this method by default uses N / ddf=0 for calculation, indicating population std. For fMRI studies, especially when std are calculated along time axis, the data is usually a sample of the population. The method should use sample std instead.

### Expected behavior
Given dataframe

1  4  7
2  5  8
3  6  9

z-scored sample should be:

(anp - anp.mean(axis=0))/anp.std(axis=0, ddof=1)
array([[-1., -1., -1.],
       [ 0.,  0.,  0.],
       [ 1.,  1.,  1.]])

### Actual behavior
nilearn.signal.clean(anp)
array([[-1.2247448, -1.2247448, -1.2247448],
       [ 0.       ,  0.       ,  0.       ],
       [ 1.2247448,  1.2247448,  1.2247448]], dtype=float32)

I think it would be beneficial if we change the default behavior.
However, given the scope of this change, I think it's worth adding the ddof as an option instead of changing the default action.

ymzayek:
Closes #3406 .

- Add "zscore_sample" strategy as an option for `standardize` parameter (Suggestions welcome for a better strategy name)
- Add a `FutureWarning` about default behavior changing to this new strategy in release 0.13.
  - If current default should be removed entirely then add this to the warning

### DIFF
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -37,7 +37,7 @@ def _standardize(signals, detrend=False, standardize='zscore'):
         If detrending of timeseries is requested.
         Default=False.
 
-    standardize : {'zscore', 'psc', True, False}, optional
+    standardize : {'zscore_sample', 'zscore', 'psc', True, False}, optional
         Strategy to standardize the signal:
 
             - 'zscore_sample': The signal is z-scored. Timeseries are shifted
@@ -634,7 +634,7 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
 
     %(high_pass)s
     %(detrend)s
-    standardize : {'zscore', 'psc', False}, optional
+    standardize : {'zscore_sample', 'zscore', 'psc', True, False}, optional
         Strategy to standardize the signal:
 
             - 'zscore_sample': The signal is z-scored. Timeseries are shifted
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -40,8 +40,10 @@ def _standardize(signals, detrend=False, standardize='zscore'):
     standardize : {'zscore', 'psc', True, False}, optional
         Strategy to standardize the signal:
 
+            - 'zscore_sample': The signal is z-scored. Timeseries are shifted
+              to zero mean and scaled to unit variance. Uses sample std.
             - 'zscore': The signal is z-scored. Timeseries are shifted
-              to zero mean and scaled to unit variance.
+              to zero mean and scaled to unit variance. Uses population std.
             - 'psc':  Timeseries are shifted to zero mean value and scaled
               to percent signal change (as compared to original mean signal).
             - True: The signal is z-scored (same as option `zscore`).
@@ -69,6 +71,15 @@ def _standardize(signals, detrend=False, standardize='zscore'):
             warnings.warn('Standardization of 3D signal has been requested but '
                           'would lead to zero values. Skipping.')
             return signals
+
+        elif (standardize == 'zscore_sample'):
+            if not detrend:
+                # remove mean if not already detrended
+                signals = signals - signals.mean(axis=0)
+
+            std = signals.std(axis=0, ddof=1)
+            std[std < np.finfo(np.float64).eps] = 1.  # avoid numerical problems
+            signals /= std
 
         elif (standardize == 'zscore') or (standardize is True):
             if not detrend:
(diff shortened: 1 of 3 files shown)


bthirion:
But i think that in that case, we simply don't want to keep the old option. I don't see any good use case for it ?

tsalo:
@bthirion do you think fully replacing it would require a deprecation cycle or anything? The differences will, I assume, generally be tiny, but it still is kind of a breaking change.

ymzayek:
@tsalo the PR doesn't change the default behavior but warns that it will change in release `0.13` and at the same time that the default will be removed entirely at that same time. Shouldn't that suffice or did you mean something else?

ymzayek:
Also for the tests I just copied the tests for the `zscore` strategy but had to lower the precision for the std of the cleaned signal. Is this as expected?

tsalo:
@ymzayek, sorry yes. I agree that the current changes are solid. I interpreted @bthirion's comment as saying that you should just update the `zscore` option without adding a new option or adding a warning, so my question was in response to that, rather than your proposed changes.

ymzayek:
>       * If current default should be removed entirely then add this to the warning

Ok no worries just making sure! I think @bthirion was responding to this note I made in the opening comment

bthirion:
Yes, we have to go through deprecation cycles.

nilearn/tests/test_signal.py
```diff
@@ -831,6 +845,11 @@ def test_clean_zscore():
     np.testing.assert_almost_equal(cleaned_signals.mean(0), 0)
     np.testing.assert_almost_equal(cleaned_signals.std(0), 1)
 
+    # Repeating test above but for new correct strategy
+    cleaned_signals = clean(signals, standardize='zscore_sample')
+    np.testing.assert_almost_equal(cleaned_signals.mean(0), 0)
+    np.testing.assert_almost_equal(cleaned_signals.std(0), 1, decimal=3)
```
bthirion, line 851: the result is thus unchangde with the new standardization (line 851 == line 846)

ymzayek: @bthirion, the results are changed slightly so I has to lower the precision hence why I set the decimal to `3` places instead of the default `7` it was before (comment https://github.com/nilearn/nilearn/pull/3474#issuecomment-1412098490). I'm not sure if this is expected or not

bthirion: Yes, but I was actually noting that the two approaches yield the same results. Do we have a test where the two ways of standardizing yield a distinct outcome ?

ymzayek: Ah ok got it. Will add that

nilearn/signal.py
```diff
@@ -37,11 +37,13 @@ def _standardize(signals, detrend=False, standardize='zscore'):
...
 
-    standardize : {'zscore', 'psc', True, False}, optional
+    standardize : {'zscore_sample', 'zscore', 'psc', True, False}, optional
         Strategy to standardize the signal:
 
+            - 'zscore_sample': The signal is z-scored. Timeseries are shifted
+              to zero mean and scaled to unit variance. Uses sample std.
             - 'zscore': The signal is z-scored. Timeseries are shifted
-              to zero mean and scaled to unit variance.
+              to zero mean and scaled to unit variance. Uses population std.
```
bthirion, line 46: Maybe we need to clarify what 'population std' refers to here.

nilearn/signal.py
```diff
@@ -609,11 +634,13 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
...
     %(detrend)s
-    standardize : {'zscore', 'psc', False}, optional
+    standardize : {'zscore_sample', 'zscore', 'psc', True, False}, optional
         Strategy to standardize the signal:
 
+            - 'zscore_sample': The signal is z-scored. Timeseries are shifted
+              to zero mean and scaled to unit variance. Uses sample std.
             - 'zscore': The signal is z-scored. Timeseries are shifted
-              to zero mean and scaled to unit variance.
+              to zero mean and scaled to unit variance. Uses population std.
```
bthirion, line 643: Same here

nilearn/signal.py
ymzayek: Is it okay like this?

bthirion (approved):
LGTM.

tsalo (approved):
The changes look good to me. Thanks!

ymzayek:
Thanks for the reviews! Merging

Issue #3437, opened 2022-12-01
`filter` option in `signal.clean` is not exposed to `nilearn.maskers.NiftiMasker` and potentially other masker objects


htwangtw:
> However, I don't think a filter option is provided for nilearn.maskers.NiftiMasker so seems it's using default aka butterworth

_Originally posted by @DasDominus in https://github.com/nilearn/nilearn/issues/3434#issuecomment-1333064573_

emdupre:
I think this would need to be modified[here](https://github.com/nilearn/nilearn/blob/ffc0489b19b01b3231862846fc15ef72cfdbd249/nilearn/maskers/base_masker.py#L122-L137) and inherited by the masker objects.

But since `low_pass` isn't supported for `cosine` filtering, we might want to also add a check for sensible args ?

TakshDhabalia:
Hey , Would like to know about this further and know more about this and if it is assigned or not. I am a beginner here so could you explain a bit more on this?

Thanks!

bthirion:
IIUC, you can try and add the parameter, and then check possible inconsistencies that may arise.
For that you need to understand the hierarchy of class and functional calls. Maybe @ymzayek you could have a quick meeting together to review the question ?

ymzayek:
@TakshDhabalia thanks for your interest in working on this! You can start as suggested by @emdupre on where to modify the code and additionally you need to include the argument in the maskers' class `__init__` method. You can start with just one masker for simplicity. Also what could help you get started is to take as an example another parameter of `signal.clean` that is already exposed in e.g. `nilearn.maskers.NiftiMasker` such as `low_pass` or `high_pass` and copy that implementation for `filter`. Does it make sense?

TakshDhabalia:
yep , I agree with that but I need a little more time on reading the documentation as I am struggling a bit on this project as a whole , I will get started on it right away but if anyone wants to help , it would be very much appreciated

ymzayek:
[...]

PR #3478, opened 2023-01-30
[ENH] Make filter options available as kwargs in `signal.clean`

### LINKED ISSUE #3475: Make `butter` and `filtfilt` parameters accessible to users in `clean`
For `scipy.signal.butter`, the `order` parameter is accessible in the nilearn function `signal.butterworth`, but it isn't accessible in `signal.clean`.

For `scipy.signal.filtfilt`, the `padtype`, `padlen`, `method`, and `irlen` parameters aren't accessible through nilearn. In particular, `padtype` and `padlen` would be very nice to set in `signal.clean`.

### Benefits to the change
I don't know if the default `padtype` ("odd"), or the hardcoded `order` (5) are well-suited to BOLD data, and I haven't come across any direct comparisons of padding types in the literature, so it seems like something users should have control over.

I've also done some informal checks with some data I had on hand, and the `odd` option that nilearn uses seems to show some edge effects that the `constant` option, for example, doesn't.

```python
import numpy as np

from nilearn import signal

data = np.random.random((100, 10000))
denoised_data = signal.clean(
    data,
    filter="butterworth",
    low_pass=0.1,
    high_pass=0.01,
    padtype="constant",  # or "odd", or "even"
    filter_order=6,  # fed into scipy.signal.butter as "N"
    t_r=2,
)
```

tsalo:
Closes #3475.

I haven't thought about testing or examples yet, but I can start on that if folks feel like this is a good direction.

I also don't know if I should make these kwargs available in the maskers as well. This kwargs approach could be useful for #3437.

- Add `padtype` and `padlen` as parameters in `nilearn.signal.butterworth`.
- Add `**kwargs` to `nilearn.signal.clean`. Kwargs prefixed with `butterworth__` will be fed into `nilearn.signal.butterworth`. This will allow users to set `padtype`, `padlen`, `order`, and `copy`.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -17,6 +17,7 @@ Enhancements
 ------------
 
 - Addition to docs to note that :meth:`~maskers.BaseMasker.inverse_transform` only performs spatial unmasking (:gh:`3445` by `Robert Williamson`_).
+- Give users control over Butterworth filter (:func:`~signal.butterworth`) parameters in :func:`~signal.clean` as kwargs (:gh:`3478` by `Taylor Salo`_).
 
 Changes
 -------
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -17,7 +17,7 @@ Enhancements
 ------------
 
 - Addition to docs to note that :meth:`~maskers.BaseMasker.inverse_transform` only performs spatial unmasking (:gh:`3445` by `Robert Williamson`_).
-- Give users control over Butterworth filter (:func:`~signal.butterworth`) parameters in :func:`~signal.clean` as kwargs (:gh:`3478` by `Taylor Salo`_).
+- Give users control over Butterworth filter (:func:`~signal.butterworth`) parameters in :func:`~signal.clean` and Masker objects as kwargs (:gh:`3478` by `Taylor Salo`_).
 
 Changes
 -------
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -625,7 +625,7 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
     %(ensure_finite)s
         Default=False.
 
-    kwargs
+    kwargs : dict
         Keyword arguments to be passed to functions called within ``clean``.
         Kwargs prefixed with ``'butterworth__'`` will be passed to
         :func:`~nilearn.signal.butterworth`.
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -377,7 +377,13 @@ def butterworth(
     else:
         critical_freq = critical_freq[0]
 
-    b, a = sp_signal.butter(order, critical_freq, btype=btype, output='ba')
+    b, a = sp_signal.butter(
+        order,
+        critical_freq,
+        btype=btype,
+        output='ba',
+        fs=sampling_rate,
+    )
     if signals.ndim == 1:
         # 1D case
         output = sp_signal.filtfilt(
@@ -499,7 +505,7 @@ def _ensure_float(data):
 def clean(signals, runs=None, detrend=True, standardize='zscore',
           sample_mask=None, confounds=None, standardize_confounds=True,
           filter='butterworth', low_pass=None, high_pass=None, t_r=2.5,
-          ensure_finite=False):
+          ensure_finite=False, **kwargs):
     """Improve :term:`SNR` on masked :term:`fMRI` signals.
 
     This function can do several things on the input signals. With the default
@@ -613,6 +619,12 @@ def clean(signals, runs=None, detrend=True, standardize='zscore',
     %(ensure_finite)s
         Default=False.
 
+    kwargs
+        Keyword arguments to be passed to functions called within ``clean``.
+        Kwargs prefixed with ``'butterworth__'`` will be passed to
+        :func:`~nilearn.signal.butterworth`.
+
+
     Returns
     -------
     cleaned_signals : :class:`numpy.ndarray`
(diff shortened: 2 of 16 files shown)


nilearn/signal.py
```diff
@@ -351,18 +368,31 @@ def butterworth(signals, sampling_rate, low_pass=None, high_pass=None,
...
     else:
         critical_freq = critical_freq[0]
 
-    b, a = sp_signal.butter(order, critical_freq, btype=btype, output='ba')
+    b, a = sp_signal.butter(
+        order,
+        critical_freq,
+        btype=btype,
+        output='ba',
+        fs=sampling_rate,
```
tsalo, line 385: This parameter changes the results, and I don't know why. I'm not sure if including the sampling rate fixes a bug or introduces one.

nilearn/signal.py
```diff
@@ -280,8 +280,16 @@ def _check_wn(btype, freq, nyq):
...
-def butterworth(signals, sampling_rate, low_pass=None, high_pass=None,
-                order=5, copy=False):
+def butterworth(
+    signals,
+    sampling_rate,
+    low_pass=None,
+    high_pass=None,
+    order=5,
+    padtype="odd",
+    padlen=None,
```
tsalo, line 290: These are the default values for `scipy.signal.filtfilt`, so behavior won't change. I'm not sure if any of them work particularly well for BOLD data though.

nilearn/signal.py
tsalo: My guess is that this is why tests are failing as well.

bthirion (commented):
Overall, the changes LGTM, thx !

nilearn/signal.py
tsalo: I think I understand what `fs` does now. Without `fs`, we need to divide the frequencies by the nyquist frequency. With `fs`, we can just feed in the frequencies. I think the code is more interpretable if we include `fs` and _don't_ modify the frequencies, but it's also primarily a stylistic change. @bthirion do you have an opinion on this?

bthirion: I think that people are used to thinking in terms of absolute frequencies.  So I would enforce the use of fs.

tsalo: Sounds good! I'll update the code.

nilearn/signal.py
ymzayek: Ok I suppose we can keep it that way for this PR but can consider optimizing them in a separate one

nilearn/tests/test_signal.py
```diff
@@ -359,6 +359,48 @@ def test_clean_t_r():
...
+        {
+            "butterworth__padtype": "constant",
+            "butterworth__padlen": 20,
+            "butterworth__order": 10,
+        },
+    ]
+    # Base result
+    t_r, high_pass, low_pass = 0.8, 0.01, 0.08
+    base_filtered = nisignal.clean(
+        x_orig, t_r=t_r, low_pass=low_cutoff, high_pass=high_cutoff
```
ymzayek, line 389: Looks like `low_cutoff` and `high_cutoff` are undefined here. Did you mean `low_pass=low_pass, high_pass=high_pass`

tsalo: Ah yes good catch. Thanks!

nilearn/signal.py
```diff
@@ -254,34 +254,48 @@ def _detrend(signals, inplace=False, type="linear", n_batches=10):
...
-            'frequency).' % (btype, wn))
+            f'The frequency specified for the {btype} pass filter is '
+            'too high to be handled by a digital filter '
+            '(superior to Nyquist frequency). '
+            f'It has been lowered to {freq} (Nyquist frequency).'
+        )
 
-    if wn < 0.0: # equal to 0.0 is okay
-        wn = np.finfo(1.).eps
+    elif freq < 0.0:  # equal to 0.0 is okay
```
ymzayek, line 277: I think this condition still needs to be tested. Looks like it was never tested when initially implemented

ymzayek (commented):
@tsalo LGTM! I think the codecov report can be resolved if the condition noted below is tested. The doc builder error is not related.

nilearn/signal.py
tsalo: Good point. I'll test that.

tsalo:
I don't think any of the remaining CI failures are due to my changes, so I think there are two remaining concerns:

1. Should I make these kwargs available to upstream tools, like `clean_img` or Masker objects?
2. AFAICT, `clean` isn't directly used in any examples, so I don't know where to exhibit the new feature.

nilearn/signal.py
tsalo: Agreed. I'm not sure how to actually do that though. If there's no principled way to decide between different padding approaches, then I think we'll have to rely on empirical testing across a range of datasets, which is a whole lot of work.

bthirion: My intuition is that this will have limited impact on final computations (beta maps, correlation, statistical tests). WDYT ?

nilearn/tests/test_signal.py
```diff
@@ -167,16 +167,80 @@ def test_butterworth():
...
+        out = nisignal.butterworth(
+            data,
+            sampling,
+            low_pass=low_pass,
+            high_pass=high_pass,
+            copy=True,
+        )
+    assert not np.array_equal(data, out)
+
+    # Test check for frequency lower than allowed (<0).
```
bthirion, line 224: to be updated

bthirion (commented):
LGTM otherwise.

bthirion:
> Should I make these kwargs available to upstream tools, like `clean_img` or Masker objects?

I'm afraid you have to.

nilearn/signal.py
tsalo: I agree that it's probably not going to have a large, or consistent, effect on results. When I tried to search the literature for mentions of padding, almost all of them mentioned zero-padding. Scipy doesn't implement zero-padding, but I think constant-padding is pretty similar, conceptually speaking. So maybe we could default to that, based on the fact that it's similar to what some previous fMRI papers have done?

bthirion: I have no better suggestion.

tsalo:
Regarding making the kwargs accessible in upstream functions/classes, should I maybe allow an _additional_ prefix (`clean__`) for kwargs that will feed into `clean`? So if I wanted to use "even" padding in a NiftiLabelsMasker, I'd have a kwarg like `clean__butterworth__padtype="even"`.

bthirion:
My impression is that it is clearer and more future-proof to prepend a 'clean_'

ymzayek:
The unit test failures seem to be related to params['clean_kwargs'] not being defined and passed to functions like `_filter_and_extract`.

nilearn/signal.py
tsalo: Do you mind if I change the default in a separate PR? I'd like there to be a separation of concerns (plus the change to the default is kind of a breaking change).

doc/changes/latest.rst
```diff
@@ -17,6 +17,7 @@ Enhancements
 ------------
 
 - Addition to docs to note that :meth:`~maskers.BaseMasker.inverse_transform` only performs spatial unmasking (:gh:`3445` by `Robert Williamson`_).
+- Give users control over Butterworth filter (:func:`~signal.butterworth`) parameters in :func:`~signal.clean` as kwargs (:gh:`3478` by `Taylor Salo`_).
```
bthirion, line 20: mention here that this is added to the masker's signature too.

bthirion (approved):
LGTM otherwise.

doc/changes/latest.rst
tsalo: Done! Thanks for catching that.

nilearn/signal.py
ymzayek: I agree for doing that in a separate PR

ymzayek (approved):
LGTM thanks @tsalo !

ymzayek:
Merging

PR #3483, opened 2023-02-02
[DOC] Add example of plotting confusion matrix from decoder family object

### LINKED ISSUE #2564: [BUG] Decoding docs improvements
Following #2415, two points that were not tackled inside #2476 in the interest of short cycles

- [x]  Add a section in 2.1 on how to determine chance with a Decoder API (waiting for #2147) EDIT : Adressed #2622
- [ ]  Rethink 2.2 section on OVO vs OVA, not sure it is useful as Decoder doesn't provide this flexibility easily. But show how to plot a confusion matrix using Decoder on a simple example.

michellewang:
Closes #2564.

- Compute/plot the confusion matrix of the fitted `FREMClassifier` in the [Decoding with FREM: face vs house object recognition example](https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_frem.html).


bthirion:
Thx for kicking this off !
 I'm afraid of the confusion matrix not being very informative on a 2-class problem.

michellewang:
> I'm afraid of the confusion matrix not being very informative on a 2-class problem.

How about making this into a 3-class problem (e.g., face, house, cat)? This would make it longer to run the example though. Otherwise it looks to me like the other decoding examples don't do a train/test split (and also seem to only use 2 classes). What do you think?

bthirion:
I'd try the 3-class version of this example.

michellewang:
Here's what the 3-class confusion matrix looks like:

sphx_glr_plot_haxby_frem_001

The other parts of the example also changed a bit now that the `'cat'` class was added:
* FREM classification accuracy dropped from 64% to 52%
* FREM weights for `'face'` class changed values (and sign), though the overall regions are similar to [before](https://nilearn.github.io/stable/_images/sphx_glr_plot_haxby_frem_001.png)

  sphx_glr_plot_haxby_frem_002

bthirion:
Not very compelling indeed.
What about using the following example, that already has 8 classes ?
https://nilearn.github.io/dev/auto_examples/02_decoding/plot_haxby_full_analysis.html#sphx-glr-auto-examples-02-decoding-plot-haxby-full-analysis-py

michellewang:
@bthirion it looks like in that example the decoders are trained on binary classification (a different decoder is used for each of the 8 classes). I'm not sure how this could be easily extended to include an 8-way confusion matrix.

If the concern with the 3-class version I shared earlier is the drop in test accuracy, I can try to see if using another class than `'cat'` produces better results.

bthirion:
Sure. I like you suggestion, thx for the additional effort.

bthirion:
Indeed, that seems good enough. It remains a bit disturbing that the 'face' coefs flipped sign, but it is as it is.

ymzayek (commented):
Something to note is that this change adds about 3 minutes to run the example. I guess we can live with this if it's really the only way to showcase plotting the confusion matrix in a simple way. I am wondering if the [multi-class example](https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_multiclass.html) can be improved instead?

Otherwise, the PR looks fine to me!

bthirion:
Indeed, we have to keep an eye on doc generation time. Thx @ymzayek

examples/02_decoding/plot_haxby_frem.py
```diff
@@ -61,16 +61,49 @@
...
+# ------------------------------------
+
+import numpy as np
+from sklearn.metrics import confusion_matrix
+from nilearn import plotting
+
+# Calculate the confusion matrix
+matrix = confusion_matrix(
+    y_test, 
+    y_pred,
```
bthirion, line 75: I hink I would add `normalize=True`

bthirion (commented):
LGTM otherwise

bthirion (approved):
LGTM, thx.

NicolasGensollen (approved):
LGTM thanks @michellewang !
Could you add an entry in `doc/changes/latest.rst` and your name in `doc/changes/names.rst` since this is your first contribution to Nilearn ?

PR #3484, opened 2023-02-03
[INFRA] Initial formatting with black and precommit

Remi-Gau:
Relates to #2528

- Set up pre-commit hooks to work only on the `nilearn/decomposition` module
  - Black 
  - Flake8
  - Codespell
  - Several "classical" hooks from pre-commit (json, yml parsing, merge conflicts...)
  - pyupgrade to update python syntax to use versions of the language >= 3.6
- Apply the hooks on `nilearn/decomposition` using `pre-commit run -a`
- Adapt Flake8 config in setup.cfg to play nice with black
- Add a pyproject.toml to host the config for black.
- Add a flake8 CI workflow to run on entire python files, only for this module
- Remove flake8 workflow that worked just on file diff
- update contributing to mention the install and usage of black and pre-commit


setup.cfg
```diff
@@ -72,6 +72,7 @@ dev =
 # W503: line break before binary operator
 # W504: line break after binary operator
 ignore=E402, W503, W504, W605
+max-line-length = 91
```
Remi-Gau, line 75: I could bring this number down but if I do Flake8 would complain for some files in this PR.

Other possibility I could try is setting the config for black for shorter lines.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+
+files: "nilearn/decomposition"
```
Remi-Gau, line 4: Adding new modules and other files will have to be done by modifying this line.
It should be a valid python regex.

.github/workflows/pep8.yml
```diff
@@ -20,11 +20,14 @@ jobs:
...
-        name: 'Install Flake8'
+          python -m pip install --upgrade pip flake8
+        name: "Install Flake8"
       - shell: bash {0}
         run: ./build_tools/flake8_diff.sh
-        name: 'Run Flake8 diff'
+        name: "Run Flake8 diff"
+      - shell: bash {0}
+        run: flake8 --config=setup.cfg --verbose "nilearn/decomposition"
+        name: "Run Flake8 on whole file"
```
Remi-Gau, line 33: Added it as part of the already existing workflow.

Could have it as a separate one it this is more convenient.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+    rev: v4.3.0
+    hooks:
+      - id: check-added-large-files
+      - id: check-ast
+      - id: check-case-conflict
+      - id: check-json
+      - id: check-merge-conflict
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
```
Remi-Gau, line 17: Not sure about some of those for nilearn.

Happy to hear any thoughts on this.

Remi-Gau:
While the code formatting is transitioning it may be good to:
- mention it in the contributing.md
- also add a mention in there about pre-commit

Remi-Gau:
Regarding mentions in the changelog, it may make sense to have a single entry once most of the code has been blackified (blacked?) rather than one per PR, no?

Remi-Gau:
I could not see any code in the reformated files that contained code that had a "table like" layout that would have been mangled by black. Let me know, if you think I overlooked something.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+
+files: "nilearn/decomposition"
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.3.0
```
NicolasGensollen, line 8: Why not the latest version (`4.4.0`) ?

.pre-commit-config.yaml
NicolasGensollen: Seems reasonable to me. 
Maybe we shouldn't include `check-added-large-files` ?

NicolasGensollen (commented):
Great ! Thanks for taking care of that @Remi-Gau ! 👍 
I went through the changes and they all LGTM.
Concerning the documentation, I also think we should add a few things related to setting up the pre-commit hook, probably here: https://nilearn.github.io/stable/development.html#setting-up-your-environment
Also, we should make sure our coding style guidelines (https://nilearn.github.io/stable/development.html#coding-style) are consistent with the linting we're enforcing.
Concerning the changelog, I also agree it might be cleaner to have only a single entry for the formatting of the whole code base rather than one entry per PR.
That being said, if we keep the one PR per module style, and different contributors end up doing them, I don't think that it would be a huge deal either...

Remi-Gau:
Ok I will update the contributing (but not today).

Also I am realizing that unless we use something like pre-commit ci, it may be good to have something in CI that runs black to at least say which files in the PR should be formatted with black.
Otherwise someone who touches those files later and use pre-commit or black will see some line changes that he or she has not worked on.

bthirion (commented):
The changes so far LGTM.

.pre-commit-config.yaml
Remi-Gau: removed it

.pre-commit-config.yaml
Remi-Gau: because I copy pasted this from another file and forgot to check the latest version 🙈 

fixed now as well as updated the flake8 version

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,39 @@
+name: "flake8"
```
Remi-Gau, line 1: Created a separate workflow as the pep8 one was manually set to not run anymore.

.github/workflows/black.yml
```diff
@@ -0,0 +1,32 @@
+---
+name: Black
```
Remi-Gau, line 2: this workflow should run the black on the diff of a PR

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+line-length = 91
+include = 'nilearn/decomposition'
```
Remi-Gau, line 3: Added in case users decide to run black directly and not via pre-commit.

CONTRIBUTING.rst
```diff
@@ -261,6 +261,22 @@ The main conventions we enforce are :
 - classes in CamelCase
 - 2 empty lines between functions or classes
 
+You can check that any code you may have edited follows these conventions 
+by running `flake8 <https://flake8.pycqa.org/en/latest/user/invocation.html#invoking-flake8>`_.
+
+Additionally, we recommend using `Black <https://black.readthedocs.io/en/stable/getting_started.html#basic-usage>`_ 
+to format your code.
+
+.. warning::
```
Remi-Gau, line 270: Will need to remove this warning once the "fade to black" is completed.

setup.cfg
bthirion: Well, I'm just curious about where these numbers come from...  Ten we simply need to decide collectively what we think is best for readability. 
Actually, I'm a bit surprised to see how these things fluctuate across years ;-)

bthirion (commented):
LGTM otherwise.

setup.cfg
Remi-Gau: For the record there is a quite bit of variability indeed.

Did set things at 90 for now (because 91 is just a weird number).

[flake8_line_length.txt](https://github.com/nilearn/nilearn/files/10661156/flake8_line_length.txt)

Remi-Gau:
OK I think I am done with this PR.

Let me know if I am missing anything or if I should change anything.

setup.cfg
ymzayek: I think the flake8 default is 79 and that's what we've been using before our PEP8 CI broke

CONTRIBUTING.rst
```diff
@@ -247,12 +247,12 @@ Example entry:
...
   
 Coding Style
--------------
+------------
 
 The nilearn codebase follows PEP8_ styling.
 The main conventions we enforce are :
 
-- line length < 80
+- line length < 90
```
ymzayek: Shouldn't we keep this as the flake8 default or is that too short?

setup.cfg
Remi-Gau: Don't feel strongly about any specific value: will change the value and adapt.

ymzayek (commented):
Thanks, this set up looks great!

CONTRIBUTING.rst
Remi-Gau: just reverted to use line length 79
made it explicit in all config files

.github/workflows/pep8.yml
```diff
@@ -1,12 +1,12 @@
-name: 'PEP8'
+name: "PEP8"
```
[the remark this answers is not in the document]
Remi-Gau: just removed it along with the bash script it was calling as I could not see it being used anywhere else

ymzayek:
Can you also remove the recommendation in step 3 under here: https://nilearn.github.io/stable/development.html#id3 and replace with a code block of syntax to flake8 a file?

Remi-Gau:
> Can you also remove the recommendation in step 3 under here: https://nilearn.github.io/stable/development.html#id3 and replace with a code block of syntax to flake8 a file?

done

also added mention of black and pre-commit

ymzayek (approved):
Thanks @Remi-Gau!

NicolasGensollen (approved):
LGTM, thanks @Remi-Gau !

bthirion (approved):
LGTM, thx.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - id: check-ast
+      - id: check-case-conflict
+      - id: check-json
+      - id: check-merge-conflict
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+
+  - repo: https://github.com/psf/black
+    rev: 22.10.0
```
Remi-Gau, line 19: Black had just released a new version, so I should probably bump this before merging.

Remi-Gau: Done.

Once this is merged I will open a separate PR with a workflow to check and update the hooks version weekly so we don't have to remember to regularly update this.

ymzayek: Great thanks!

ymzayek:
Ok merging

PR #3485, opened 2023-02-06
[DOC] update doc threshold image

### LINKED ISSUE #2048: Documentation of threshold_img unclear
1.) From the documentation of `threshold_img` it's unclear whether features below threshold will be kept or just set to zero. When I first saw the documentation I thought that `threshold_img` is a feature reduction procedure and will return a feature reduced subset. I had to find out by myself that features below threshold will be set to zero. The used phrase _"will be kept"_ also leads to the false idea that `threshold_img` is a feature reduction procedure.

2.) If I read the source code right `img_data[np.abs(img_data) < cutoff_threshold] = 0.)`, all features below threshold will be set to zero. This logically implies that all other values greater **or equal** to threshold will be kept. Therefore, the sentence 

> _The voxels which have intensities greater than this score will be kept._ 

should be changed to

>  _The voxels which have intensities greater or equal than this score will be kept._

3.) It's very nice that `threshold_img` can also handle percentiles because otherwise you always have to inspect the data first (or look at descriptive stats) to find a good voxel intensity value as threshold. However it would be nice to give an explanation of how the percentile rank is calculated behind the scenes. Again if I read the source code right the given percentile string (e.g. '80%') is converted to a percentile rank using `scipy.stats.scoreatpercentile`. It would be nice to give a reference to this function and to document which keyword-arguments of `scipy.stats.scoreatpercentile` are used when `check_threshold` is called.

Remi-Gau:
Closes #2048

1. Clarify that voxels with values
  - inferior to threshold are set to 0
  - supeerior or equal to threshold are kept as is

2. Explain how percentiles are computed.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -22,7 +22,4 @@ Enhancements
 Changes
 -------
 
-- The documentation for :func:`~image.threshold_img` has been improved,
-  with more information about which voxels are set to 0
-  and which ones keep their original values
-  (:gh:`3485` by `Rémi Gau`_).
+- The documentation for :func:`~image.threshold_img` has been improved, with more information about which voxels are set to 0 and which ones keep their original values (:gh:`3485` by `Rémi Gau`_).
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -21,3 +21,8 @@ Enhancements
 
 Changes
 -------
+
+- The documentation for :func:`~image.threshold_img` has been improved, 
+  with more information about which voxels are set to 0 
+  and which ones keep their original values
+  (:gh:`3485` by `Rémi Gau`_).
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -22,7 +22,7 @@ Enhancements
 Changes
 -------
 
-- The documentation for :func:`~image.threshold_img` has been improved, 
-  with more information about which voxels are set to 0 
+- The documentation for :func:`~image.threshold_img` has been improved,
+  with more information about which voxels are set to 0
   and which ones keep their original values
   (:gh:`3485` by `Rémi Gau`_).
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -866,13 +866,13 @@ def threshold_img(
         Those with intensities greater or equal than the requested threshold
         will keep their original value.
         If float, we threshold the image based on image intensities.
-        The given value should be within the range of minimum and maximum intensity 
+        The given value should be within the range of minimum and maximum intensity
         of the input image.
-        If string, it should finish with percent sign e.g. "80%" 
-        and we threshold based on the score obtained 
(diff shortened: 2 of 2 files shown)


nilearn/image/image.py
```diff
@@ -861,14 +861,18 @@ def threshold_img(
...
-        If string, it should finish with percent sign e.g. "80%" and we threshold
-        based on the score obtained using this percentile on the image data. The
-        voxels which have intensities greater than this score will be kept.
+        Voxels with intensities less than the requested threshold
+        will be set to zero.
+        Those with intensities greater or equal than the requested threshold
+        will keep their original value.
+        If float, we threshold the image based on image intensities.
+        The given value should be within the range of minimum and maximum intensity
+        of the input image.
```
ymzayek, line 870:
```suggestion
        The given value should be within the range of minimum and maximum
        intensity of the input image.
```

nilearn/image/image.py
```diff
@@ -861,14 +861,18 @@ def threshold_img(
...
+        If string, it should finish with percent sign e.g. "80%"
+        and we threshold based on the score obtained
+        using this percentile on the image data.
         The given string should be within the range of "0%" to "100%".
+        The percentile rank is computed using ``scipy.stats.scoreatpercentile``.
```
ymzayek, line 875:
```suggestion
        The percentile rank is computed using
        :func:`scipy.stats.scoreatpercentile`.
```

ymzayek (commented):
LGTM @Remi-Gau! Just minor things. I'd also add a whatsnew entry.

Remi-Gau:
> I'd also add a whatsnew entry.

added something in latest.rst.

doc/changes/latest.rst
```diff
@@ -21,3 +21,8 @@ Enhancements
 
 Changes
 -------
+
+- The documentation for :func:`~image.threshold_img` has been improved,
```
ymzayek, line 25: In general we have the convention of one line per entry in this file (see [PR Structure](https://nilearn.github.io/stable/development.html#pr-structure)). I know this is broken sometimes and I don't have a strong opinion about it but better to try to keep things consistent.

ymzayek (approved):
LGTM @Remi-Gau

doc/changes/latest.rst
Remi-Gau: woopsy

I am trying to use [semantic linebreaks](https://sembr.org/) more and more, so I completely overlooked this.

To make it harder to miss, would it make sense to add links to the relevant headings in the [PR checklist](https://github.com/nilearn/nilearn/blob/b7e5efdd37a6b4cc276763fc5f2a2cde81f7af73/.github/workflows/auto-comment.yml#L22) ?

ymzayek: Yes I think that's a good idea! Can you open an issue for that?

bthirion (approved):
LGTM, thx.

ymzayek:
Merging

PR #3492, opened 2023-02-08
[FIX] Fix syntax for duplicated target name Sphinx error

ymzayek:
Fixing syntax error causing doc build to fail. Duplicated target names for links need to have 2 underscores instead of one

Failure on main: https://github.com/nilearn/nilearn/actions/runs/4121333245/jobs/7116920487#step:33:9112


NicolasGensollen (approved):
LGTM, thanks @ymzayek !

PR #3494, opened 2023-02-08
[MAINT] Add flag to full doc build to finish build when first warning is encountered 

### LINKED ISSUE #3493: Avoid that Sphinx full build exits with error as soon as first warning is encountered
Sphinx is set to error out when encountering the first warning in a full build. We can set the flag `--keep-going` to let it finish the build and report all the warnings that may come after. It will still exit with an error and the build will fail but then we can quickly see all the warnings that need to be handled. See Sphinx build options: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-keep-going

ymzayek:
Closes #3493.

This will prompt the full sphinx build to keep going and finish when encountering a first warning so that all subsequent warnings are caught. Build will still exit with exit status 1 as expected

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -162,7 +162,7 @@ pdf: latex
 check:
 	rm -rf _build/doctrees
 	rm -rf $(BUILDDIR)/html/_images
-	BUILD_DEV_HTML=1 $(SPHINXBUILD) -W -T -n -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	BUILD_DEV_HTML=1 $(SPHINXBUILD) -W --keep-going -T -n -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 
 install:
 	rm -rf _build/doctrees _build/nilearn.github.io


ymzayek:
This [failure](https://github.com/nilearn/nilearn/actions/runs/4125521718/jobs/7126191608#step:33:8043) is related to nipy server being down causing broken references to nibabel nifti image class. It is also failing on [main](https://github.com/nilearn/nilearn/actions/runs/4125501354/jobs/7126144683#step:33:8111). Once this is resolved on their side, I will test the full build again to make sure it passes with the new flag.

emdupre (approved):
This LGTM, thx !

emdupre:
I don't think this interacts with our dev doc builds, right ?

ymzayek:
> I don't think this interacts with our dev doc builds, right ?

@emdupre well it applies to all of the full builds (dev or stable or those called in PRs). The only thing it does is allows us to see is if there is more than one warning treated as error. But in terms of the build failing or passing the behavior is the same.

emdupre:
Perfect, I just wanted to triple-check that this won't end up changing the end-user experience of the dev docs. Thx !

bthirion (approved):
LGTM, thx.

PR #3495, opened 2023-02-08
[MAINT] Use lower resolution MNI152 templates for tests and examples

### LINKED ISSUE #3482: Azure Pipelines test on Windows with Python 3.9 fails sometimes
The workflow is failing because it's timing out while running pytest. A possible solution is to set a timeout that exceeds 60m if the service allows it for free.

Also need to check if there are certain tests taking longer than usual for any particular reason since the testing runs in 38 minutes on successful runs.

Example failure:
https://dev.azure.com/Parietal/Nilearn/_build/results?buildId=5336&view=logs&j=57060dbb-acbf-556b-a6e6-e748b9693eea

ymzayek:
Closes #3482.

- Lower resolution of loaded templates for tests from 1mm to 2mm
- Change some examples to load lower resolution templates
- Simplify mni space data generator by passing resolution argument directly into `load_mni152_brain_mask`

### DIFF
--- a/examples/06_manipulating_images/plot_resample_to_template.py
+++ b/examples/06_manipulating_images/plot_resample_to_template.py
@@ -13,7 +13,7 @@ Function :func:`nilearn.image.resample_img` could also be used to achieve this.
 from nilearn.datasets import fetch_neurovault_motor_task
 from nilearn.datasets import load_mni152_template
 
-template = load_mni152_template()
+template = load_mni152_template(resolution=2)
 
 motor_images = fetch_neurovault_motor_task()
 stat_img = motor_images.images[0]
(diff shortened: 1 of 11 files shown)


ymzayek:
@bthirion yes locally I saw about a 10min difference. And now I also see that the CI builds are also faster. The time saved varies per OS but it is faster on all of them with the changes in this PR. I just need to check that 1 failure.

ymzayek:
The failure is unrelated to this PR. Seems like another precision error.

bthirion (approved):
Indeed. Looks good.

NicolasGensollen (approved):
LGTM, thanks @ymzayek !

PR #3498, opened 2023-02-10
[ENH] Add utility function to load sample motor contrast image

### LINKED ISSUE #2184: `load_example_image`

michellewang:
Closes #2184.

- Add function `load_sample_motor_activation_image` to `nilearn.datasets.utils`
  - Uses the Neurovault motor task contrast which will now be distributed with `nilearn` (to avoid network issues)
- Refactor 4 plotting examples to use `load_sample_motor_activation_image`:
  - `plot_demo_glass_brain.py`
  - `plot_demo_glass_brain_extensive.py`
  - `plot_demo_plotting.py`
  - `plot_demo_more_plotting.py`


bthirion (commented):
Thx for opening this ! 
The addition of  nilearn/datasets/data/image_10426.nii.gz seems to be a mistake.

nilearn/datasets/utils.py
```diff
@@ -22,7 +22,7 @@
 from .._utils import fill_doc
 
 _REQUESTS_TIMEOUT = (15.1, 61)
-
+_PACKAGE_DIRECTORY = os.path.dirname(os.path.abspath(__file__))
```
bthirion, line 25: Why is this added ?

nilearn/datasets/__init__.py
```diff
@@ -78,6 +80,7 @@
            'load_mni152_wm_mask', 'fetch_icbm152_brain_gm_mask',
            'fetch_atlas_surf_destrieux', 'fetch_atlas_talairach',
            'get_data_dirs',
+           'load_sample_motor_activation_image',
```
NicolasGensollen, line 83: You'll also need to add this to `doc/modules/datasets.rst` such that it gets properly added to the docs.

nilearn/datasets/utils.py
NicolasGensollen: I think it comes from applying the same logic as what was done here:

https://github.com/nilearn/nilearn/blob/3e4652fba970b26b9e66c77a24ebb1655e2838de/nilearn/datasets/struct.py#L22-L32

nilearn/datasets/utils.py
```diff
@@ -950,3 +950,15 @@ def make_fresh_openneuro_dataset_urls_index(
...
+def load_sample_motor_activation_image():
+    """Load a single functional image showing motor activations.
+
+    Returns
+    -------
+    str
+        Path to the sample functional image.
+    """
+
+    return os.path.join(_PACKAGE_DIRECTORY, "data", "image_10426.nii.gz")
```
NicolasGensollen, line 964: Not super important, but another possibility would be to rely on pathlib for this and eventually cast to string when returning:

```python
from pathlib import Path

return str(Path(__file__).parent / "data" / "image_10426.nii.gz")
```

NicolasGensollen (commented):
Thanks for working on this @michellewang !

michellewang:
> The addition of nilearn/datasets/data/image_10426.nii.gz seems to be a mistake.

@bthirion that file was added on purpose because @jeromedockes mentioned distributing the example image with `nilearn` to avoid network failures [in the original issue](https://github.com/nilearn/nilearn/issues/2184#issue-508557494) (also see @ymzayek's [comment](https://github.com/nilearn/nilearn/issues/2184#issuecomment-1420488531)). But if you don't think the image should be shipped with `nilearn` I can just call `nilearn.datasets.fetch_neurovault_motor_task` to download the file if needed.

nilearn/datasets/utils.py
michellewang: Updated to what you suggested, thanks!

nilearn/datasets/__init__.py
michellewang: Ah I missed that, thanks!

bthirion:
> > The addition of nilearn/datasets/data/image_10426.nii.gz seems to be a mistake.
> 
> @bthirion that file was added on purpose because @jeromedockes mentioned distributing the example image with `nilearn` to avoid network failures [in the original issue](https://github.com/nilearn/nilearn/issues/2184#issue-508557494) (also see @ymzayek's [comment](https://github.com/nilearn/nilearn/issues/2184#issuecomment-1420488531)). But if you don't think the image should be shipped with `nilearn` I can just call `nilearn.datasets.fetch_neurovault_motor_task` to download the file if needed.

@jeromedockes, is this what you has in mind ? If yes, we should probably use this image more systematically across the documentation.

ymzayek:
> @jeromedockes, is this what you has in mind ? If yes, we should probably use this image more systematically across the documentation.

It can also be added to:
- `examples/01_plotting/plot_3d_map_to_surface_projection.py`
- `examples/06_manipulating_images/plot_negate_image.py`
- `examples/06_manipulating_images/plot_resample_to_template.py`

@michellewang can you rebase on main so the CI runs with the latest changes?

michellewang:
@ymzayek thanks for pointing out the examples I missed!

ymzayek:
@michellewang looks like something went wrong with the rebase. The changes from other PRs looks like they were made in this one so it makes it hard to see your contributions. It's a bit hard to tell what happened. You might need to git reset or run an interactive rebase and drop the commits from main and redo the rebase. Let me know if you have trouble with this.

ymzayek (approved):
LGTM @michellewang! Can you resolve the conflicts?

jeromedockes:
> @jeromedockes, is this what you has in mind ? If yes, we should probably use this image more systematically across the documentation.

sorry everybody for missing this discussion! yes I suggested that possibility. However that was a smaller point, the main one being to have a very simple function for loading an example image. I'm ok with both options: (i) shipping the image (possibly at a reduced resolution) with the package, or (ii) downloading it with one of the existing fetchers under the hood if we want to avoid increasing the package size.

jeromedockes:
it seems something might have gone wrong with a git manip.
@michelle if your last real commit on this was the one from 2 weeks ago, you can

```
git checkout datasets/load_sample_motor_activation_image
git resset --hard 376442333
git merge upstream/main
```

first make sure that commit 37644 was indeed the last work you have done on this branch, and that you don't have any un-committed work in your working tree, because `git reset --hard` can actually cause work to be lost.
(I have misspelled `reset` to avoid hasty copy/paste gags)
you can also first commit or stash any changes you have in the working tree, then
```
git checkout -b my_pr_backup
git checkout datasets/load_sample_motor_activation_image
```

you will have 2 files with conflicts to fix. that is due to the big reformatting PRs that touch a lot of files.

michellewang:
Thanks a lot @jeromedockes! I knew something got messed up but wasn't sure how to fix it. I think it should be good now.

bthirion (approved):
I think it's OK to add a 180k image to the repo.
Thx !

ymzayek:
Ok merging.
Codespell errors are unrelated to the changes in this PR. I'll fix them separately

PR #3505, opened 2023-02-12
Update html_surface.py

### LINKED ISSUE #3500: Renaming plotting.html_surface.one_mesh_info() and plotting.html_surface.full_brain_info()
This is a follow-up issue from #3173.

`plotting.html_surface.one_mesh_info()` and `plotting.html_surface.full_brain_info()` should be renamed to `plotting.html_surface._one_mesh_info()` and `plotting.html_surface._full_brain_info()` respectively so that it becomes explicit that they're not in the public API.

mehrdadbn9:
closes #3500 
please check it

### DIFF
--- a/nilearn/plotting/html_surface.py
+++ b/nilearn/plotting/html_surface.py
@@ -38,9 +38,10 @@ def _get_vertexcolor(surf_map, cmap, norm,
     return to_color_strings(vertexcolor)
 
 
-def _one_mesh_info(surf_map, surf_mesh, threshold=None, cmap=cm.cold_hot,
+def one_mesh_info(surf_map, surf_mesh, threshold=None, cmap=cm.cold_hot,
                   black_bg=False, bg_map=None, symmetric_cmap=True,
                   vmax=None, vmin=None):
+    
     """Prepare info for plotting one surface map on a single mesh.
 
     This computes the dictionary that gets inserted in the web page,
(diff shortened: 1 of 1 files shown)


alexisthual:
Thanks for opening this @mehrdadbn9!

Here is some kind feedback:
* it's better if you open this PR with a message saying "Closes #3500" or something similar suggested by Github ; this way, when this PR gets merged, the linked issue will automatically be closed.
* as you can see, due to the changes you made in this PR, some tests of the codebase now fail, so we won't merge your PR as is.

You should have a look at the failing tests to see what should be fixed. If you need some help, you can have a peak at the following hint, but I think it's worth trying without it.
<details>
<summary>See hint</summary>
<p>

Several tests fail with `AttributeError: module 'nilearn.plotting.html_surface' has no attribute 'one_mesh_info'`.
This is because the functions you renamed are used elsewhere in the codebase, and therefore you should rename them there as well.

</p>
</details>

mehrdadbn9 (commented):
tried to fix code coverage

mehrdadbn9:
@ymzayek  i think it didn't work

ymzayek:
Hi @mehrdadbn9 sorry if I wasn't clear. You should leave the original function untouched except for changing it's name by adding the underscore. Then under this you will define a new function which is basically what I wrote in code block in the comment above. What I'm doing there is defining a new function called `one_mesh_info` that calls the renamed function `_one_mesh_info`.

mehrdadbn9:
@ymzayek 
oh you are right so as i understood i should firstly define a private def and another definition without underscore that re turn private def when it is called.

ymzayek:
> oh you are right so as i understood i should firstly define a private def and another definition without underscore that re turn private def when it is called.

@mehrdadbn9 yes exactly. The original function can just be renamed to make it private by adding the underscore and nothing else should change under that function. Then a new function named without the underscore will spit a warning and call the private function.

PR #3508, opened 2023-02-13
[FIX] Fix failing test due to use of deprecated Nibabel method

### LINKED ISSUE #3506: Handle failing tests due to Nibabel 5.0.1 deprecation of `get_data()`
See failures on main: https://github.com/nilearn/nilearn/actions/runs/4161555109

ymzayek:
Closes #3506

Replaces deprecated method `get_data()` with `get_fdata()` for one failing test.


bthirion (approved):
LGTM, thx !

emdupre (approved):
Thanks, @ymzayek  !

My only q is whether it's worth explicitly setting the dtype in the `get_fdata` calls to lower memory usage, since the default type is `np.float64`.  Not critical, but we could probably pass `dtype=np.float32` !

PR #3520, opened 2023-02-20
[DOC] Add explanation of pre-releases to maintenance docs

### LINKED ISSUE #3518: Update maintenance docs to include new pre-release schedule
Add a few lines about monthly pre-release schedule to the maintenance docs: https://nilearn.github.io/stable/maintenance.html

ymzayek:
Closes #3518 .

Question: With the pre-releases should we also push a new build of the stable docs?

### DIFF
--- a/doc/maintenance.rst
+++ b/doc/maintenance.rst
@@ -379,3 +379,13 @@ Finally, we need to include this new file in ``doc/changes/whats_new.rst``:
 
    .. _latest:
    .. include:: latest.rst
+
+Pre-Releases
+------------
+
+Every month, in between major and minor releases, we will release a
+pre-release version of Nilearn. This will basically follow the same steps as
+outlined above with the exception that the release with be tagged with
+x.y.zrcN (release candidate). For example, version 0.10.1.dev will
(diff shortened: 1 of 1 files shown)


emdupre:
> Question: With the pre-releases should we also push a new build of the stable docs?

I'd say no, personally ! I don't see this in e.g. the pandas docs, for example.

ymzayek:
@NicolasGensollen thanks for your input. The argument was initially to do more patch releases, for example every time a bug is fixed. However, this is maybe not ideal for nilearn, where it's beneficial for users to not have to deal with frequent changes and potential breaking of the API. This pre-release schedule would allow users who rely on bugfixes to use the rc version, while the majority of users don't have to worry about it. Pinging @tsalo and @jeromedockes 

Also concerning this PR, maybe it's ok to not document it like this and keep it informal for now.

bthirion:
We could get back to the point at a future coredev meeting: what is here represents what was discussed last time.

ymzayek:
Closing this after recent decision in core developer meeting to abandon the prerelease plan

PR #3529, opened 2023-02-21
[MAINT] Update deprecated parameter name for Pandas `DataFrame.to_csv` method

### LINKED ISSUE #3530: `line_terminator` error in more recent versions of pandas
A "new" failure in CI seems related to a recent version of pandas.

for example: https://github.com/nilearn/nilearn/actions/runs/4231491657/jobs/7350086385#step:8:16744

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html

```text
lineterminatorstr, optional

    The newline character or character sequence to use in the output file. Defaults to os.linesep, which depends on the OS in which this method is called (’\n’ for linux, ‘\r\n’ for Windows, i.e.).

    Changed in version 1.5.0: Previously was line_terminator, changed for consistency with read_csv and the standard library ‘csv’ module.
```

ymzayek:
Pandas pre-release 2.0.0rc0 deprecates parameter `line_terminator` for `lineterminator` in method `to_csv` causing some tests to fail on pre-release workflow (e.g. https://github.com/nilearn/nilearn/actions/runs/4233168828/jobs/7353781075#step:8:16674)

Edit: Closes #3530

Edit: Or remove it alltogether if not needed (new parameter name doesn't meet minimum supported requirements )

### DIFF
--- a/nilearn/interfaces/bids/glm.py
+++ b/nilearn/interfaces/bids/glm.py
@@ -143,7 +143,6 @@ def save_glm_to_bids(
         design_matrix.to_csv(
             dm_file,
             sep='\t',
-            lineterminator='\n',
             index=False,
         )
 
--- a/nilearn/interfaces/bids/glm.py
+++ b/nilearn/interfaces/bids/glm.py
@@ -143,7 +143,7 @@ def save_glm_to_bids(
         design_matrix.to_csv(
             dm_file,
             sep='\t',
-            line_terminator='\n',
+            lineterminator='\n',
             index=False,
         )
 


nilearn/interfaces/bids/glm.py
```diff
@@ -143,7 +143,7 @@ def save_glm_to_bids(
         design_matrix.to_csv(
             dm_file,
             sep='\t',
-            line_terminator='\n',
+            lineterminator='\n',
```
Remi-Gau, line 146: I would be tempted to remove it altogether and run the tests.

It is the only occurrence in the entire code base and I am unclear as to why it would be needed specifically here. But maybe I am missing something.

Remi-Gau (approved):
LGTM

nilearn/interfaces/bids/glm.py
ymzayek: I think we are good removing it since it is set to the default value

ymzayek:
Merging to green the CI

Issue #3507, opened 2023-02-13
Wrong underline length in numpy doc strings


Remi-Gau:
Some headings in the doc strings do not have the right number of underlining character and this causes extra warning in the doc build.

This does not fail the build though.

For example: https://github.com/nilearn/nilearn/actions/runs/4122627440/jobs/7119633978#step:33:323

### note

Probably in the category
- low effort
- low impact

Can probably be tackled at the same time as #2528

Remi-Gau:
List of lines to fix:

```bash
$ flake8 --select D409 nilearn

nilearn/_utils/data_gen.py:500:1: D409 Section underline should match the length of its name
nilearn/_utils/ndimage.py:15:1: D409 Section underline should match the length of its name
nilearn/_utils/ndimage.py:15:1: D409 Section underline should match the length of its name
nilearn/_utils/niimg.py:107:1: D409 Section underline should match the length of its name
nilearn/_utils/niimg.py:107:1: D409 Section underline should match the length of its name
nilearn/connectome/group_sparse_cov.py:28:1: D409 Section underline should match the length of its name
nilearn/datasets/func.py:423:1: D409 Section underline should match the length of its name
nilearn/datasets/func.py:771:1: D409 Section underline should match the length of its name
nilearn/datasets/func.py:771:1: D409 Section underline should match the length of its name
nilearn/datasets/func.py:813:1: D409 Section underline should match the length of its name
nilearn/datasets/func.py:813:1: D409 Section underline should match the length of its name
nilearn/datasets/neurovault.py:1151:1: D409 Section underline should match the length of its name
nilearn/datasets/neurovault.py:1151:1: D409 Section underline should match the length of its name
nilearn/datasets/neurovault.py:2616:1: D409 Section underline should match the length of its name
nilearn/datasets/neurovault.py:2616:1: D409 Section underline should match the length of its name
nilearn/datasets/neurovault.py:2656:1: D409 Section underline should match the length of its name
nilearn/datasets/neurovault.py:2656:1: D409 Section underline should match the length of its name
nilearn/datasets/utils.py:820:1: D409 Section underline should match the length of its name
nilearn/image/image.py:58:1: D409 Section underline should match the length of its name
nilearn/image/resampling.py:192:1: D409 Section underline should match the length of its name
nilearn/image/resampling.py:703:1: D409 Section underline should match the length of its name
nilearn/interfaces/fmriprep/load_confounds.py:81:1: D409 Section underline should match the length of its name
[...]

Issue #3331, opened 2022-08-09
Refactor pytest fixtures so that they are reusable


ymzayek:
See https://github.com/nilearn/nilearn/pull/3300#discussion_r915643466 and https://github.com/nilearn/nilearn/pull/3327#discussion_r939007061

Relevant steps:
- Move existing fixtures to one file (`conftest.py`?)
- Create new fixtures where generated data is reused
- Better naming to improve reusability

jeromedockes:
@Remi-Gau  as you are looking at cleaning up some tests see also this one

Remi-Gau:
Yup I had seen it.

My idea was first to start splitting the most complex tests into smaller chunks, create setup functions when needed.
Then once that is done I think we can further refactor the test suite by using more fixture.

The first phase is "easy" because I can use static analysis to tell me where to look. Phase 2 will require me to be less lazy.

Remi-Gau:
Actually I figured "in for a penny, in for a pound" some I am trying to "fixturaze" things in the PR that aim to simplify tests.

ymzayek:
WDYT of adding a fixture to always seed numpy rng such as

```python
@pytest.fixture(autouse=True)
def seed_numpy_random():
    np.random.seed(0)
```

Remi-Gau:
> WDYT of adding a fixture to always seed numpy rng such as
> 
> ```python
> @pytest.fixture(autouse=True)
> def seed_numpy_random():
>     np.random.seed(0)
> ```

My first hunch is that it would make sense. Probably need to update the doc to mention that. Isn't there something about seeding in there?

Note to self: also update Doc in current PR to mention the use of fixtures.

jeromedockes:
from numpy doc:
> This is a convenience, legacy function that exists to support older code that uses the singleton RandomState. Best practice is to use a dedicated Generator instance rather than the random variate generation methods exposed directly in the random module.

instead of using `seed` we should provide a value for nilearn & scikit-learn `random_state` parameter when applicable, or use `np.random.default_rng(0)` to obtain a rng that we use directly

jeromedockes:
seeding the global RandomState in a function is a bad idea in general because it modifies a global object -- other functions that use random numbers will also start getting seeded sequences. it probably wouldn't be a problem for our tests but still it's better to have the habit of never doing it

ymzayek:
Yea sorry for the bad implementation

jeromedockes:
[...]

PR #3534, opened 2023-02-23
[MAINT] update flake8 configuration to cover doc strings

Remi-Gau:
This PR enforces some of the flake8 check on the doc strings.

The default behavior of flake would not report those errors:
http://www.pydocstyle.org/en/6.3.0/error_codes.html

Relates to #2528
Relates to https://github.com/nilearn/nilearn/issues/3507

- apply flake8 to doc strings
- add flake8 doc string plugin
- update flake8 and pre-commit config
- update flake8 ci


setup.cfg
```diff
@@ -67,13 +68,28 @@ dev =
...
+    dist,
+	nilearn/externals/tempita
+--select=D,E,F,W,C90
+docstring-convention=numpy
+max-line-length = 79
+max_complexity = 43
+per-file-ignores =
+    # missing docstring
+	*/tests/*: D103
+	*/tests/__init__.py: D104
```
tsalo, line 85: There's some odd spacing here.

tsalo:
A number of the current flake8 failures come from testing functions. Do you think it's worth linting those, or should they be skipped?

Remi-Gau:
> A number of the current flake8 failures come from testing functions. Do you think it's worth linting those, or should they be skipped?

I should have fixed them but it may be worth deciding how much we want to enforce doc strings in tests.

I have silenced some already but I won't fight back if we decided to just no enforce any of the doc string things in the test files.

setup.cfg
Remi-Gau: what do you mean? the indentation?

Remi-Gau: should be better now

setup.cfg
```diff
@@ -67,13 +68,29 @@ dev =
...
+    .git,
+    __pycache__,
+	env,
+    build,
+    dist,
+	nilearn/externals/tempita
+--select = D,E,F,W,C90
+docstring-convention = numpy
+max-line-length = 79
+max_complexity = 43
```
Remi-Gau, line 81: I have set this to this value as it is currently the maximum of the most complex function / method.
This sets an upper bound to how complex the code can get before flake8 throws an error.
There is a handful of function that would need refactoring to bring this value back towards a more acceptable value (around 20 ish).

Remi-Gau:
> > A number of the current flake8 failures come from testing functions. Do you think it's worth linting those, or should they be skipped?
> 
> I should have fixed them but it may be worth deciding how much we want to enforce doc strings in tests.
> 
> I have silenced some already but I won't fight back if we decided to just no enforce any of the doc string things in the test files.

OK was testing things on some other parts of the code base and it does make sense to relax some of the doc strings formatting rules in the tests suite for now. If we change our mind we start enforcing those rules and clean things up one thing at a time.

nilearn/decoding/decoder.py
```diff
@@ -875,7 +879,7 @@ def _predict_dummy(self, n_samples):
 
 @fill_doc
 class Decoder(_BaseDecoder):
-    """A wrapper for popular classification strategies in neuroimaging.
+    r"""A wrapper for popular classification strategies in neuroimaging.
```
ymzayek, line 882: Why use raw string here?

ymzayek:
> > > A number of the current flake8 failures come from testing functions. Do you think it's worth linting those, or should they be skipped?
> > 
> > 
> > I should have fixed them but it may be worth deciding how much we want to enforce doc strings in tests.
> > I have silenced some already but I won't fight back if we decided to just no enforce any of the doc string things in the test files.
> 
> OK was testing things on some other parts of the code base and it does make sense to relax some of the doc strings formatting rules in the tests suite for now. If we change our mind we start enforcing those rules and clean things up one thing at a time.

Yes I would say to leave them for now. This could be related to pytest refactoring I wanted to do #3331

nilearn/decoding/decoder.py
Remi-Gau: was getting a:
```
nilearn/decoding/decoder.py:882:1: D301 Use r""" if any backslashes in a docstring
```

so I obediently complied. 🙈 

Now pushed this that should be a better fix:
https://github.com/nilearn/nilearn/pull/3534/commits/6b253036a97646bfdbab2d87705956de3d4bf929

ymzayek: Oh I see haha. Ok better now thanks!

Remi-Gau:
BTW this should catch as an error the wrong underline length that the doc build would only throw as warning.

setup.cfg
tsalo: Thanks!

bthirion (approved):
LGTM.

ymzayek (approved):
LGTM thanks @Remi-Gau

PR #3535, opened 2023-02-23
MAINT: Deal with matplotlib deprecation

larsoner:
`matplotlib.cm.register_cmap` is deprecated:
```
>>> matplotlib.cm.register_cmap()
<stdin>:1: MatplotlibDeprecationWarning: The register_cmap function was deprecated in Matplotlib 3.7 and will be removed two minor releases later. Use ``matplotlib.colormaps.register(name)`` instead.
...
```

### DIFF
--- a/nilearn/plotting/cm.py
+++ b/nilearn/plotting/cm.py
@@ -222,8 +222,14 @@ _cmap_d['videen_style'] = _colors.LinearSegmentedColormap.from_list(
 globals().update(_cmap_d)
 # Register cmaps in matplotlib too
 for k, v in _cmap_d.items():
+    try:
+        from matplotlib import colormaps as _colormaps
+    except ImportError:
+        _register_cmap = _cm.register_cmap
+    else:
+        _register_cmap = _colormaps.register  # 3.5+
     try:  # "bwr" is in latest matplotlib
-        _cm.register_cmap(name=k, cmap=v)
+        _register_cmap(name=k, cmap=v)
     except ValueError:
         pass
 


larsoner:
> There is at least one test that would fail under the original bug conditions.

To find stuff like this in MNE we test with warnings-as-errors in pytest, and use a prerelease version of dependencies like `matplotlib`. It helps stay ahead of the curve a bit. But it would be a big change here...

I could instead change your pytest config to at least error on matplotlib deprecations, then you'd at least see them when they cut new releases. Not sure how many other failures this would trigger, it took us a bit to get to 3.7-deprecation-free compat over in MNE...

> Changelog or what's new entry

Not sure if you all do this for trivial changes like this, happy to add one if you want

bthirion:
> To find stuff like this in MNE we test with warnings-as-errors in pytest, and use a prerelease version of dependencies like `matplotlib`. It helps stay ahead of the curve a bit. But it would be a big change here...
> 
> I could instead change your pytest config to at least error on matplotlib deprecations, then you'd at least see them when they cut new releases. Not sure how many other failures this would trigger, it took us a bit to get to 3.7-deprecation-free compat over in MNE...

Indeed we should go into such a direction. Thx.

bthirion (approved):
Thx !

ymzayek:
Hi @larsoner thanks for this. Indeed we could try adding testing with warnings-as-errors.

> I could instead change your pytest config to at least error on matplotlib deprecations, then you'd at least see them when they cut new releases. Not sure how many other failures this would trigger, it took us a bit to get to 3.7-deprecation-free compat over in MNE...

We can start with that in a new PR and see what we get?

ymzayek (approved):

PR #3553, opened 2023-03-01
[FMT] format example/04_glm_first_level

Remi-Gau:

- apply pre-commit hooks to example/04_glm_first_level

### Files in other PRs
- 'examples/04_glm_first_level/plot_fiac_analysis.py',
- 'examples/04_glm_first_level/plot_fixed_effects.py',
- 'examples/04_glm_first_level/plot_localizer_surface_analysis.py',


examples/04_glm_first_level/plot_write_events_file.py
```diff
@@ -24,41 +25,193 @@
...
-    0., 2.4, 8.7, 11.4, 15., 18., 20.7, 23.7, 26.7, 29.7, 33., 35.4, 39.,
-    41.7, 44.7, 48., 56.4, 59.7, 62.4, 69., 71.4, 75., 83.4, 87., 89.7,
-    96., 108., 116.7, 119.4, 122.7, 125.4, 131.4, 135., 137.7, 140.4,
-    143.4, 146.7, 149.4, 153., 156., 159., 162., 164.4, 167.7, 170.4,
-    173.7, 176.7, 188.4, 191.7, 195., 198., 201., 203.7, 207., 210.,
-    212.7, 215.7, 218.7, 221.4, 224.7, 227.7, 230.7, 234., 236.7, 246.,
-    248.4, 251.7, 254.7, 257.4, 260.4, 264., 266.7, 269.7, 275.4, 278.4,
-    284.4, 288., 291., 293.4, 296.7,
-])
+onsets = np.array(
```
ymzayek, line 28: Maybe we can refactor this and the next array or leave as is (before the reformatting)?
Not sure why they have to be hardcoded. We can instead generate them with a numpy function and print them to showcase what they look like

NicolasGensollen: I haven't looked too much into it, but the pattern isn't that obvious to me... 🤔 
A possible alternative, could be to declare it as a string which would be easier to format in a pretty way:

```python
onset_data = "[0.0, 2.4, 8.7, 11.4, 15.0, 18.0, 20.7, 23.7, 26.7]"
onsets = np.array(json.loads(txt))
```

examples/04_glm_first_level/plot_write_events_file.py
```diff
@@ -24,41 +25,193 @@
...
 # Associated trial types: these are numbered between 0 and 9, hence
 # corresponding to 10 different conditions.
-trial_type_idx = np.array([
-    7, 7, 0, 2, 9, 4, 9, 3, 5, 9, 1, 6, 8, 8, 6, 6, 8, 0, 3, 4, 5, 8, 6,
-    2, 9, 1, 6, 5, 9, 1, 7, 8, 6, 6, 1, 2, 9, 0, 7, 1, 8, 2, 7, 8, 3, 6,
-    0, 0, 6, 8, 7, 7, 1, 1, 1, 5, 5, 0, 7, 0, 4, 2, 7, 9, 8, 0, 6, 3, 3,
-    7, 1, 0, 0, 4, 1, 9, 8, 4, 9, 9,
-])
+trial_type_idx = np.array(
+    [
```
NicolasGensollen, line 117: Obviously same here....

examples/04_glm_first_level/plot_write_events_file.py
```diff
@@ -85,17 +240,19 @@
...
 import os
-outdir = 'results'
+
+outdir = "results"
 if not os.path.exists(outdir):
     os.mkdir(outdir)
-tsvfile = outdir + '/localizer_events.tsv'
-events.to_csv(tsvfile, sep='\t', index=False)
-print(f'The event information has been saved to {tsvfile}')
+tsvfile = f"{outdir}/localizer_events.tsv"
```
NicolasGensollen, line 247: I would rely on pathlib for this kind of path manipulations. Or at least use `os.path.join`.

NicolasGensollen (commented):
Thanks @Remi-Gau !
Looks mostly good to me with a few suggestions.

examples/04_glm_first_level/plot_write_events_file.py
Remi-Gau: done

I suspect that in another PR we could try to use Path everywhere possible in the examples.

examples/04_glm_first_level/plot_predictions_residuals.py
```diff
@@ -62,23 +65,23 @@
 # ---------------------------
 from nilearn import plotting
 
-z_map = fmri_glm.compute_contrast('active - rest')
+z_map = fmri_glm.compute_contrast("active - rest")
 
 plotting.plot_stat_map(z_map, bg_img=mean_img, threshold=3.1)
 
+from nilearn.maskers import NiftiSpheresMasker
```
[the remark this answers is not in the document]
Remi-Gau: done

examples/04_glm_first_level/plot_write_events_file.py
Remi-Gau: went with your option

just afraid it may confuse readers

another option make the number of events shorter: say 2 per condition

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -164,20 +189,31 @@ def make_localizer_contrasts(design_matrix):
 #
 from nilearn import plotting
 
+
 def plot_contrast(first_level_model):
-    """ Given a first model, specify, estimate and plot the main contrasts"""
+    """Specify, estimate and plot the main contrasts
```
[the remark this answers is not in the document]
Remi-Gau: done

examples/04_glm_first_level/plot_write_events_file.py
ymzayek: I know we want to avoid this as much possible but this could be a good case for using `# fmt: off` no? I think when it comes to user facing examples we need to worry more about readability and being understandable/simple than when dealing with the codebase.

Remi-Gau: yeah I was thinking about this as well

I think for the examples it does make sense to show pythonic code but make sure it is readable too.

let me test it

ymzayek: Great let's go with then when needed for examples

bthirion (approved):
LGTM, thx.

NicolasGensollen (approved):
LGTM, thanks @Remi-Gau !

PR #3559, opened 2023-03-02
[MAINT] make test_percentile_range for canica less brittle

Remi-Gau:
Closes #3546 

Changes proposed in this pull request:

- simplify test by making sure that only one warning with the expected message is thrown

### Unrelated

tidy up gitignore

PR #3560, opened 2023-03-03
[DOC] Fix Sphinx build error caused by changed pattern

ymzayek:
Fixing build fail on main: https://github.com/nilearn/nilearn/actions/runs/4316077088/jobs/7531359830
In the reformatting process, a pattern for the Sphinx `:start-after:` directive was changed.


ymzayek:
Merging this to fix the dev build. Test failure is unrelated and being fixed in #3559

PR #3563, opened 2023-03-04
[Fix] relabel `label_maps` to match the cluster IDs in the clusters table

### LINKED ISSUE #3561: Possible discrepancies between the `Cluster ID` and its label in the `label_map` of `get_clusters_tables`
Thanks for adding the label maps to the outputs of `get_clusters_tables`.
I was working with the clusters table and the label maps and noticed a mismatch between the Cluster ID and the label in the `label_map`. While cluster IDs are sorted based on their `peak_vals` and relabeled in the for loop, it seems that the `label_maps` are not relabeled since they are appended to the list before sorting the cluster IDs.

Nilearn version: `0.10.1.dev`

### Expected behavior
Cluster `n` corresponds to voxels with value `n` in the label map. Except for `n = 0`, when the peak falls outside of the cluster body.

### Actual behavior
In some cases, cluster `x` corresponds to a cluster with value `y`.

```python
import nibabel as nib
import numpy as np
import pandas as pd

from nilearn import reporting
from nilearn import datasets

# Run get_clusters_table on stat_img
clusters_table, label_maps = reporting.get_clusters_table(
            stat_img, 3.6, two_sided=True, return_label_maps=True
        )
# Select negative tail
neg_clusters_table = clusters_table[clusters_table["Peak Stat"] < 0]
neg_label_maps = label_maps[1]

# Get cluster IDs and its respective coordinates from the clusters table
new_clusters_table = neg_clusters_table.apply(pd.to_numeric, errors='coerce').dropna()
cluster_ids = new_clusters_table["Cluster ID"].to_list()
coords = new_clusters_table[["X", "Y", "Z"]].to_numpy()

# Use the coordinates from the clusters table to find the Cluster ID in the label map 
ijk = nib.affines.apply_affine(np.linalg.inv(neg_label_maps.affine), coords).astype(int)
cluster_label_ids = neg_label_maps.get_fdata()[tuple(ijk.T)]

print(cluster_ids, cluster_label_ids)
# > [1.0, 2.0, 3.0, 4.0, 5.0] [1. 3. 5. 2. 4.]
# In this we get that coordinate of cluster 1 has value of 1 in the label map, 
# but cluster 2 has value of 3, 3 value of 5, 4 value of 2 and 5 value of 4. 
# The array [1. 3. 5. 2. 4.] corresponds to the sorted indices that were not apply 
# to the label_maps.
```

JulioAPeraza:
Closes #3561.

- Relabel the clusters in `label_maps` to match their corresponding cluster IDs in the clusters table.

### DIFF
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -297,6 +297,7 @@ def get_clusters_table(
     # check that stat_img is niimg-like object and 3D
     stat_img = check_niimg_3d(stat_img)
     affine = stat_img.affine
+    shape = stat_img.shape
 
     # Apply threshold(s) to image
     stat_img = threshold_img(
@@ -352,13 +353,18 @@ def get_clusters_table(
         # Sort by descending max value
         clust_ids = [clust_ids[c] for c in (-peak_vals).argsort()]
 
-        re_label_map = np.zeros_like(label_map)
+        if return_label_maps:
+            # Relabel label_map based on sorted ids
+            relabel_idx = np.insert(clust_ids, 0, 0).argsort().astype(np.int32)
+            relabel_map = relabel_idx[label_map.flatten()].reshape(shape)
+            # Save label maps as nifti objects
+            label_maps.append(
+                new_img_like(stat_img, relabel_map, affine=affine)
+            )
+
         for c_id, c_val in enumerate(clust_ids):
             cluster_mask = label_map == c_val
             masked_data = temp_stat_map * cluster_mask
-
-            # Relabel label_map
-            re_label_map[cluster_mask] = c_id + 1
 
             cluster_size_mm = int(np.sum(cluster_mask) * voxel_size)
 
@@ -407,9 +413,6 @@ def get_clusters_table(
                         "",
                     ]
                 rows += [row]
-
-        # Save label maps as nifti objects
-        label_maps.append(new_img_like(stat_img, re_label_map, affine=affine))
 
         # If we reach this point, there are clusters in this sign
         no_clusters_found = False
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -25,6 +25,7 @@ from nilearn._utils.niimg import _safe_get_data
 
 def _local_max(data, affine, min_distance):
     """Find all local maxima of the array, separated by at least min_distance.
+
     Adapted from https://stackoverflow.com/a/22631583/2589328
 
     Parameters
@@ -416,7 +417,4 @@ def get_clusters_table(
     else:
         df = pd.DataFrame(columns=cols, data=rows)
 
-    if return_label_maps:
-        return df, label_maps
-    else:
-        return df
+    return (df, label_maps) if return_label_maps else df
--- a/nilearn/reporting/_get_clusters_table.py
+++ b/nilearn/reporting/_get_clusters_table.py
@@ -311,7 +311,9 @@ def get_clusters_table(
     # If cluster threshold is used, there is chance that stat_map will be
     # modified, therefore copy is needed
     stat_map = _safe_get_data(
-        stat_img, ensure_finite=True, copy_data=(cluster_threshold is not None)
+        stat_img,
+        ensure_finite=True,
+        copy_data=(cluster_threshold is not None),
     )
 
     # Define array for 6-connectivity, aka NN1 or "faces"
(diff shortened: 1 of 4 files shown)


nilearn/reporting/tests/test_reporting.py
```diff
@@ -5,8 +5,9 @@
 # Set backend to avoid DISPLAY problems
 from nilearn.plotting import _set_mpl_backend
 
-from nilearn.reporting import (get_clusters_table,
-                               )
+from nilearn.reporting import (
```
bthirion, line 8:
```suggestion
from nilearn.reporting import get_clusters_table
```

nilearn/reporting/tests/test_reporting.py
```diff
@@ -142,10 +148,7 @@ def test_get_clusters_table(tmp_path):
     data_extra_dim = data[..., np.newaxis]
     stat_img_extra_dim = nib.Nifti1Image(data_extra_dim, np.eye(4))
     cluster_table = get_clusters_table(
-        stat_img_extra_dim,
-        4,
-        0,
-        two_sided=True
+        stat_img_extra_dim, 4, 0, two_sided=True
```
bthirion, line 151: Either everything fits on one line or you have to put one argument per line.

nilearn/reporting/tests/test_reporting.py
```diff
@@ -156,43 +159,55 @@ def test_get_clusters_table(tmp_path):
     assert len(cluster_table) == 1
 
 
+def test_get_clusters_table_relabel_label_maps():
```
bthirion, line 162: Could you add a short description / intent ?

nilearn/reporting/tests/test_reporting.py
```diff
@@ -156,43 +159,55 @@ def test_get_clusters_table(tmp_path):
...
     stat_img = nib.Nifti1Image(data, np.eye(4))
     data_orig = get_data(stat_img).copy()
 
     # test one cluster should be removed
     clusters_table = get_clusters_table(
-        stat_img,
-        4,
-        cluster_threshold=10,
-        two_sided=True
+        stat_img, 4, cluster_threshold=10, two_sided=True
```
bthirion, line 194: Same as above.

nilearn/reporting/tests/test_reporting.py
```diff
@@ -156,43 +159,55 @@ def test_get_clusters_table(tmp_path):
...
     assert len(clusters_table) == 1
 
     # test no clusters should be removed
     stat_img = nib.Nifti1Image(data, np.eye(4))
     clusters_table = get_clusters_table(
-        stat_img,
-        4,
-        cluster_threshold=7,
-        two_sided=False
+        stat_img, 4, cluster_threshold=7, two_sided=False
```
bthirion, line 202: Same as above.

nilearn/reporting/tests/test_reporting.py
```diff
@@ -156,43 +159,55 @@ def test_get_clusters_table(tmp_path):
...
     assert len(clusters_table) == 2
 
     # test cluster threshold is None
     stat_img = nib.Nifti1Image(data, np.eye(4))
     clusters_table = get_clusters_table(
-        stat_img,
-        4,
-        cluster_threshold=None,
-        two_sided=False
+        stat_img, 4, cluster_threshold=None, two_sided=False
```
bthirion, line 210: And again.

bthirion (commented):
Thx ! This lookd good overall.
One question: Which test fails with the current code ?

nilearn/reporting/tests/test_reporting.py
JulioAPeraza: Thanks for catching this. It looks like the lack of a trailing comma caused `black` to reformat that into a single line

JulioAPeraza:
Thanks for the code review!

> One question: Which test fails with the current code ?

This new test `test_get_clusters_table_relabel_label_maps()` fails with the current code (`main` branch).

nilearn/reporting/_get_clusters_table.py
```diff
@@ -327,27 +336,30 @@ def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
...
-        label_maps.append(new_img_like(stat_img, label_map, affine=affine))
         clust_ids = sorted(list(np.unique(label_map)[1:]))
         peak_vals = np.array(
-            [np.max(temp_stat_map * (label_map == c)) for c in clust_ids])
+            [np.max(temp_stat_map * (label_map == c)) for c in clust_ids]
+        )
         # Sort by descending max value
         clust_ids = [clust_ids[c] for c in (-peak_vals).argsort()]
 
+        re_label_map = np.zeros_like(label_map)
```
bthirion, line 355: To make sure I understand: the relabeling is meant to keep consistency after the sorting operation ?
If yes, this could be added explicitly.

bthirion (commented):
The change looks good. 
Last minor thing: apparently one line is not covered by tests. Can you take care of that ?
Best,

bthirion (approved):
LGTM, thx.

NicolasGensollen (approved):
LGTM, thanks @JulioAPeraza !

ymzayek (approved):
LGTM thx @JulioAPeraza ! Merging

PR #3574, opened 2023-03-07
[DOC] Fix Sphinx build error on main

ymzayek:
https://github.com/nilearn/nilearn/actions/runs/4351222925

### DIFF
--- a/doc/connectivity/region_extraction.rst
+++ b/doc/connectivity/region_extraction.rst
@@ -31,7 +31,7 @@ dataset.
 
 .. literalinclude:: ../../examples/03_connectivity/plot_extract_regions_dictlearning_maps.py
     :start-after: # We use nilearn's datasets downloading utilities
-    :end-before: ################################################################################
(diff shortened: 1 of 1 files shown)


doc/connectivity/region_extraction.rst
```diff
@@ -31,7 +31,7 @@ dataset.
 
 .. literalinclude:: ../../examples/03_connectivity/plot_extract_regions_dictlearning_maps.py
     :start-after: # We use nilearn's datasets downloading utilities
-    :end-before: ################################################################################
+    :end-before: ##############################################################################
```
Remi-Gau, line 34: This is totally my doing. Sorry about that.

For my defense, I think that those separators have different length in different examples files. Is that intended?

ymzayek: Hard to say but it's not a big deal I think and an easy fix!

ymzayek:
Failure is unrelated. Merging

PR #3597, opened 2023-03-15
[ENH] Raise a warning if the bids events.tsv file doesn't contain column "trial_type"

### LINKED ISSUE #3588: warn for BIDS events file with no `trial_type` column
Technically `trial_type` columns in BIDS events.tsv files are optional.

https://bids-specification.readthedocs.io/en/latest/modality-specific-files/task-events.html

So it can happen to have valid BIDS dataset with either no `trial_type` column or trial type information stored under a different column.

See from the BIDS examples:
- https://github.com/bids-standard/bids-examples/blob/master/ds000117/sub-01/ses-mri/func/sub-01_ses-mri_task-facerecognition_run-01_events.tsv

The default behavior of `make_first_level_design_mat` to create dummy events makes sense when there is no `trial_type` column but may not be appropriate when trial type information is stored under another column.

https://github.com/nilearn/nilearn/blob/5ed3b2471038c1098412196aa38c5232e166e445/nilearn/glm/first_level/design_matrix.py#L269

It may be nice to warn users by throwing a warning when events are collected by `first_level_from_bids`.

Note that this should happen rarely so the number of users affected should be low and they may probably already know about it if they have explored the dataset they are working with.

vaswin0:
Closes #3588

- Added a  function  "check_trial_type"  to raise a warning if the bids events.tsv file doesn't contain column "trial_type"
- The function is called on the dataframe corresponding to the events.tsv file iteratively from the list of dataframes to check whether the column "trial_type" exists.
-  If there are only two columns, then they must be the required ones. If there are more than two columns, and column "trial_type" doesn't exist, it can be either due to misnaming or the non-existence of that information.


nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,33 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+
+    return
+  
+  
+  
+  
+  
+  
+  
+
```
Remi-Gau, line 249:
```suggestion
```
remove extra lines: only 2 lines between functions defintions

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,33 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+  
+    """ Raises warning if the dataframe corresponding to 
+      to events.tsv doesn't contain column named "trial_type".
+      
+    parameters
+    ----------
+    df : pandas dataframe corresponding to a events.tsv file
+
+    """
+    if len(df.columns) != 2 and 'trial_type' not in df.columns:
```
Remi-Gau, line 234:
```suggestion
    if len(df.columns) > 2 and 'trial_type' not in df.columns:
```

If the value is inferior to 2, then this is not even a BIDS valid dataset because the events file must at least have `onset` and `duration`.

Better if we stick to cases where the value is more than 2 because otherwise we start doing some sort of indirect BIDS validation: let's make our life easier and assume users have a valid BIDS dataset as input.

nilearn/glm/first_level/first_level.py
```diff
@@ -1062,6 +1089,9 @@ def first_level_from_bids(dataset_path, task_label, space_label=None,
                                  (len(events), len(imgs)))
             events = [pd.read_csv(event, sep='\t', index_col=None)
                       for event in events]
+            for event in events:
+                check_trial_type(event)
```
Remi-Gau, line 1093: I was just thinking...

If this is done that way users are going to get one warning for every single events file with a problem in the dataset. Usually if one event file has the problem it is likely all events file may have it, which may flood the user with warnings.

Too many warnings may be a) overwhelming at first b) then users may habituate and not see when new warnings appear.

Maybe a better way to do this is to check the dataframes and then throw a single warning, listing all the "problematic" tsv files.

Remi-Gau (changes requested):
Couple of quick comments and suggestion of a slightly different implementation to make the warning messages both more useful and less noisy for users.

nilearn/glm/first_level/first_level.py
vaswin0: yea, right.  Raising a warning for each file might not be the best idea. We can show something like "files x, y, z(.tsv) lack column "trial_type" " as a single warning.

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,37 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+        df = pd.read_csv(event, sep='\t', index_col=None)
+        if len(df.columns) > 2 and 'trial_type' not in df.columns:
+           file_names = file_names + event.split('/')[-1] + ', '
+            
+    file_names =  file_names[0:-2]
+    
+    if len(file_names) > 0:
+       warn(
+         'No column named "trial_type" found in ' + file_names +
+        '. All rows in those files will be treated'
```
Remi-Gau, line 246:
```suggestion
        f'No column named "trial_type" found in {file_names}.'
        ' All rows in those files will be treated'
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,37 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
         results = {0.0: ols_result}
 
     return labels, results
+  
+def check_trial_type(events):
```
Remi-Gau, line 224:
```suggestion
def _check_trial_type(events):
```

Probably want to keep this function private for now.

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,37 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+  
+    """ Raises warning if the events.tsv file
+    doesn't contain column named "trial_type".
+      
+    parameters
+    ----------
+    events : list of path of events.tsv files
```
Remi-Gau, line 231:
```suggestion
    events : :obj:`list` of path of events.tsv files
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,37 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+
+    """
+    file_names = ''
+    
+    for event in events:
+        df = pd.read_csv(event, sep='\t', index_col=None)
+        if len(df.columns) > 2 and 'trial_type' not in df.columns:
+           file_names = file_names + event.split('/')[-1] + ', '
```
Remi-Gau, line 239:
```suggestion
           file_names += f"{event.split('/')[-1]}, "
```

nilearn/glm/first_level/first_level.py
```diff
@@ -1060,8 +1091,11 @@ def first_level_from_bids(dataset_path, task_label, space_label=None,
                                  'files. Same number of event files as '
                                  'the number of runs is expected' %
                                  (len(events), len(imgs)))
+            check_trial_type(events)
```
Remi-Gau, line 1094:
```suggestion
            _check_trial_type(events)
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,37 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
         results = {0.0: ols_result}
 
     return labels, results
+
```
Remi-Gau, line 223:
```suggestion
  
  
```

Remi-Gau (changes requested):
Looking good.

Do you could can add a test for `check_trial_type`?

bthirion (commented):
Thx ! You need to test all the lines.

bthirion (approved):
LGTM.

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,38 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+    ----------
+    events : :obj:`list` of path of events.tsv files
+
+    """
+    file_names = ''
+    
+    for event in events:
+        df = pd.read_csv(event, sep='\t', index_col=None)
+        if len(df.columns) > 2 and 'trial_type' not in df.columns:
+           file_names += f"{event.split('/')[-1]}, "
```
NicolasGensollen, line 240: I would rely on pathlib for this:

```python
from pathlib import Path

file_names = []
for event in events:
    event = Path(event)
    df = pd.read_csv(...)
    if ....:
        file_names.append(event.name)
if file_names:
    warn(f".... {','.join(file_names)}....")
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,38 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
         results = {0.0: ols_result}
 
     return labels, results
+  
+  
+def _check_trial_type(events):
+
```
NicolasGensollen, line 226:
```suggestion
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,38 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
         results = {0.0: ols_result}
 
     return labels, results
+  
+  
+def _check_trial_type(events):
+  
+    """ Raises warning if the events.tsv file
```
NicolasGensollen, line 227:
```suggestion
    """ Raises a warning if the events.tsv file
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,38 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+    doesn't contain column named "trial_type".
```
NicolasGensollen, line 228:
```suggestion
    doesn't contain a column named "trial_type".
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,38 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+      
+    parameters
+    ----------
+    events : :obj:`list` of path of events.tsv files
+
```
NicolasGensollen, line 233:
```suggestion
    List of paths to the events.tsv files.
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,38 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+  
+  
+def _check_trial_type(events):
+  
+    """ Raises warning if the events.tsv file
+    doesn't contain column named "trial_type".
+      
+    parameters
+    ----------
+    events : :obj:`list` of path of events.tsv files
```
NicolasGensollen, line 232: This is mixing type information and description. Maybe 

```
:obj:`list` of PathLike
```

 ?

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,38 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
...
+    
+    if len(file_names) > 0:
+       warn(
+        f'No column named "trial_type" found in {file_names}.'
+        ' All rows in those files will be treated'
+        ' as if they are instances of same experimental condition.'
+        ' If there is a column in the dataframe corresponding to trial information,'
+        ' consider renaming it to "trial_type".')
+
+    return
```
NicolasGensollen, line 252:
```suggestion
```

nilearn/glm/first_level/first_level.py
```diff
@@ -1060,8 +1092,11 @@ def first_level_from_bids(dataset_path, task_label, space_label=None,
                                  'files. Same number of event files as '
                                  'the number of runs is expected' %
                                  (len(events), len(imgs)))
+            _check_trial_type(events)
             events = [pd.read_csv(event, sep='\t', index_col=None)
                       for event in events]
+
```
NicolasGensollen, line 1098:
```suggestion
```

nilearn/glm/first_level/first_level.py
```diff
@@ -220,6 +220,38 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
         results = {0.0: ols_result}
 
     return labels, results
+  
+  
+def _check_trial_type(events):
```
NicolasGensollen, line 225: Not sure what was decided for type hints. `List[PathLike]` would be better in my opinion, but we can also do:

```suggestion
def _check_trial_type(events: list) -> None:
```

NicolasGensollen (commented):
Thanks @vaswin0 !
I made a few small suggestions below. At the moment, I think this is still lacking a few small things:

- unit tests for the new function `_check_trial_type`
- integration test (basically check that it is doing what it should when used from the FirstLevelModel class)
- add your name to `doc/changes/names.rst`
- add a new entry to `doc/changes/latest.rst`

nilearn/glm/first_level/first_level.py
Remi-Gau: Actually to make sure that flake8 will not complain, we should use the imperative mood.

```
"""Raise a warning if the events.tsv file
```

nilearn/glm/first_level/first_level.py
NicolasGensollen: OK, then I suppose having the simple list annotation as in the suggestion would be better for now.

Remi-Gau:
@vaswin0 let me know if you need help with setting up tests for this PR.

Remi-Gau:
@vaswin0 
you may want to resolve the merge conflict before moving on with this PR.

from what I see it is a matter of now calling `_check_trial_type` from within `_get_events_files`

nilearn/glm/first_level/first_level.py
Remi-Gau: Also maybe easier to go for name the variable `event_` rather than just `event` to a make it easier to tell it apart from `events`.

Remi-Gau (approved):

nilearn/glm/first_level/first_level.py
```diff
@@ -235,6 +235,32 @@ def run_glm(Y, X, noise_model='ar1', bins=100,
     return labels, results
 
 
+def _check_trial_type(events):
+    """Check that the event files contain a "trial_type" column.
+
+    Parameters
+    ----------
+    events : :obj:`list` of path of events.tsv files.
```
ymzayek, line 243: this seems to be mixing type definition and description. Also clarify if the paths are strings or path-like objects (or both) in the type definition and then the description can be: "A list of paths of events.tsv files"

ymzayek (approved):
Minor comment. LGTM otherwise!

nilearn/glm/first_level/first_level.py
Remi-Gau: ah yes.
let me fix that.

Remi-Gau:
ok merging.
thanks @vaswin0 @bthirion @ymzayek @NicolasGensollen

Issue #3570, opened 2023-03-06
simplify "complex" tests


Remi-Gau:
list of "long" tests (function length > 50 lines) that are likely to contain too many asserts and could suffer "test pollution" and should probably be refactored.

```bash
$ flake8 nilearn/**/tests/test_*.py --select CFQ001 --max-function-length 60 
```

> [!TIP]
> # **💻 🍁  For the [hacktoberfest](https://hacktoberfest.com/) contributors 🍁 💻**
>
> Make sure to:
> - read our [contributing guidelines](https://nilearn.github.io/stable/development.html#contribution-guidelines)
>
> If this is your first time contributing to Nilearn, add your details in the [CITATION.cff file](https://github.com/nilearn/nilearn/blob/dcd92aa357ab3a751161aeea2edfa3a775ae644f/CITATION.cff#L42).
>
> For your pull request to be valid, you should refactor some of the long tests mentioned in the list below.
>

- [ ] nilearn/datasets/tests/test_atlas.py:445:1: CFQ001 Function test_fetch_atlas_yeo_2011 has length 63 that exceeds max allowed length 60
- [ ] nilearn/datasets/tests/test_func.py:668:1: CFQ001 Function test_fetch_development_fmri has length 63 that exceeds max allowed length 60
- [ ] nilearn/datasets/tests/test_neurovault.py:792:1: CFQ001 Function test_fetch_neurovault_ids has length 80 that exceeds max allowed length 60
- [ ] nilearn/decoding/tests/test_decoder.py:1442:1: CFQ001 Function test_decoder_vs_sklearn has length 70 that exceeds max allowed length 60
- [ ] nilearn/maskers/tests/test_surface_labels_masker.py:356:1: CFQ001 Function test_surface_label_masker_check_output_2d has length 72 that exceeds max allowed length 60
- [ ] nilearn/plotting/tests/test_find_cuts.py:143:1: CFQ001 Function test_find_cut_slices has length 65 that exceeds max allowed length 60
- [ ] nilearn/plotting/tests/test_find_cuts.py:284:1: CFQ001 Function test_find_parcellation_cut_coords has length 83 that exceeds max allowed length 60
- [ ] nilearn/plotting/tests/test_find_cuts.py:391:1: CFQ001 Function test_find_probabilistic_atlas_cut_coords has length 61 that exceeds max allowed length 60
[...]

PR #3604, opened 2023-03-18
[MAINT] Simplify tests connectome

Remi-Gau:
Relates to https://github.com/nilearn/nilearn/issues/3570

- refactor tests of the regions module to have fewer assertions in each test

### DIFF
--- a/nilearn/connectome/tests/test_connectivity_matrices.py
+++ b/nilearn/connectome/tests/test_connectivity_matrices.py
@@ -275,6 +275,16 @@ def test_geometric_mean_properties():
         assert_array_equal(spd, input_spd)
     assert is_spd(gmean, decimal=7)
 
+
+def test_geometric_mean_properties_check_invariance():
+    n_matrices = 40
+    n_features = 15
+    spds = [
+        random_spd(n_features, eig_min=1.0, cond=10.0, random_state=0)
+        for _ in range(n_matrices)
+    ]
+    gmean = _geometric_mean(spds)
+
     # Invariance under reordering
     spds.reverse()
     spds.insert(0, spds[1])
@@ -295,6 +305,14 @@ def test_geometric_mean_properties():
         _geometric_mean(spds_inv, init=init), linalg.inv(gmean)
     )
 
+
+def test_geometric_mean_properties_check_gradient():
+    n_matrices = 40
+    n_features = 15
+    spds = [
+        random_spd(n_features, eig_min=1.0, cond=10.0, random_state=0)
+        for _ in range(n_matrices)
+    ]
     # Gradient norm is decreasing
     grad_norm = grad_geometric_mean(spds, tol=1e-20)
     difference = np.diff(grad_norm)
@@ -306,13 +324,17 @@ def test_geometric_mean_properties():
     tol = 1e-20
     with warnings.catch_warnings(record=True) as w:
         warnings.simplefilter("always")
-        gmean = _geometric_mean(spds, max_iter=max_iter, tol=tol)
+        _geometric_mean(spds, max_iter=max_iter, tol=tol)
         assert len(w) == 1
     grad_norm = grad_geometric_mean(spds, max_iter=max_iter, tol=tol)
     assert len(grad_norm) == max_iter
     assert grad_norm[-1] > tol
 
-    # Evaluate convergence. A warning is printed if tolerance is not reached
+
+def test_geometric_mean_properties_evaluate_convergence():
+    n_matrices = 40
+    n_features = 15
+    # A warning is printed if tolerance is not reached
     for p in [0.5, 1.0]:  # proportion of badly conditioned matrices
         spds = [
             random_spd(n_features, eig_min=1e-2, cond=1e6, random_state=0)
@@ -323,7 +345,7 @@ def test_geometric_mean_properties():
             for _ in range(int(p * n_matrices), n_matrices)
         )
         max_iter = 30 if p < 1 else 60
-        gmean = _geometric_mean(spds, max_iter=max_iter, tol=1e-5)
+        _geometric_mean(spds, max_iter=max_iter, tol=1e-5)
 
 
 def test_geometric_mean_errors():
--- a/nilearn/connectome/tests/test_connectivity_matrices.py
+++ b/nilearn/connectome/tests/test_connectivity_matrices.py
@@ -482,7 +482,7 @@ def _generate_signals_and_compute_covariances(n_subjects, n_features):
     ledoit_estimator = LedoitWolf()
     for k in range(n_subjects):
         n_samples = 200 + k
-        signal, _, _ = generate_signals(
+        signal, *_ = generate_signals(
             n_features=n_features,
             n_confounds=5,
             length=n_samples,
@@ -497,6 +497,9 @@ def _generate_signals_and_compute_covariances(n_subjects, n_features):
 
 
 @pytest.mark.parametrize(
+    "cov_estimator", [EmpiricalCovariance(), LedoitWolf()]
+)
+@pytest.mark.parametrize(
     "kind",
     [
         "covariance",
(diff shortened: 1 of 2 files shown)


nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -475,82 +499,170 @@ def test_connectivity_measure_outputs():
...
+    signals, emp_covs, ledoit_covs = signals_and_covariances
+    kind = "tangent"
+
+    conn_measure = ConnectivityMeasure(kind=kind, cov_estimator=cov_estimator)
+    connectivities = conn_measure.fit_transform(signals)
+
+    if isinstance(cov_estimator, LedoitWolf):
+        covs = ledoit_covs
+    elif isinstance(cov_estimator, EmpiricalCovariance):
+        covs = emp_covs
```
ymzayek, line 553: See comment above

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -475,82 +499,170 @@ def test_connectivity_measure_outputs():
...
+    signals, emp_covs, ledoit_covs = signals_and_covariances
+    kind = "precision"
+
+    conn_measure = ConnectivityMeasure(kind=kind, cov_estimator=cov_estimator)
+    connectivities = conn_measure.fit_transform(signals)
+
+    if isinstance(cov_estimator, LedoitWolf):
+        covs = ledoit_covs
+    elif isinstance(cov_estimator, EmpiricalCovariance):
+        covs = emp_covs
```
ymzayek, line 587: See comment above

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -475,82 +499,170 @@ def test_connectivity_measure_outputs():
...
+    signals, emp_covs, ledoit_covs = signals_and_covariances
+    kind = "correlation"
+
+    conn_measure = ConnectivityMeasure(kind=kind, cov_estimator=cov_estimator)
+    connectivities = conn_measure.fit_transform(signals)
+
+    if isinstance(cov_estimator, LedoitWolf):
+        covs = ledoit_covs
+    elif isinstance(cov_estimator, EmpiricalCovariance):
+        covs = emp_covs
```
ymzayek, line 609: See comment above

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -475,82 +499,170 @@ def test_connectivity_measure_outputs():
...
+    signals, emp_covs, ledoit_covs = signals_and_covariances
+    kind = "partial correlation"
+
+    conn_measure = ConnectivityMeasure(kind=kind, cov_estimator=cov_estimator)
+    connectivities = conn_measure.fit_transform(signals)
+
+    if isinstance(cov_estimator, LedoitWolf):
+        covs = ledoit_covs
+    elif isinstance(cov_estimator, EmpiricalCovariance):
+        covs = emp_covs
```
ymzayek, line 634: See comment above

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -660,15 +814,13 @@ def test_connectivity_measure_outputs():
...
-    n_features = 49
-
     # Generate signals and compute covariances and apply confounds while
     # computing covariances
     signals = []
     for k in range(n_subjects):
         n_samples = 200 + k
         signal, _, confounds = generate_signals(
-            n_features=n_features,
+            n_features=49,
```
ymzayek, line 823: Use `N_FEATURES`

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -692,6 +844,22 @@ def test_confounds_connectome_measure():
...
+
+def test_confounds_connectome_measure_errors():
+    n_subjects = N_SUBJECTS
+    # Generate signals and compute covariances and apply confounds while
+    # computing covariances
+    signals = []
+    for k in range(n_subjects):
+        n_samples = 200 + k
+        signal, _, confounds = generate_signals(
+            n_features=49,
```
ymzayek, line 856: Use `N_FEATURES`

ymzayek (commented):
Thx @Remi-Gau ! A few minor comments

nilearn/connectome/tests/test_connectivity_matrices.py
Remi-Gau: done

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -453,104 +518,140 @@ def test_connectivity_measure_errors():
...
+    cov_estimator, signals_and_covariances
+):
+    signals, covs = signals_and_covariances
+    kind = "partial correlation"
+
+    conn_measure = ConnectivityMeasure(kind=kind, cov_estimator=cov_estimator)
+    connectivities = conn_measure.fit_transform(signals)
+
+    for k, cov_new in enumerate(connectivities):
+        prec = linalg.inv(covs[k])
```
NicolasGensollen, line 628: I think we could also parametrize these tests on "kind". It seems like only the assertion part is changing, the rest is the same. So maybe something like this (let me know what you think, it's only a suggestion, I haven't tested whether this works or will end up more readable....):

```python
@pytest.mark.parametrize("cov_estimator", [EmpiricalCovariance(), LedoitWolf()])
@pytest.mark.parametrize("kind", ["covariance", "correlation", "tangent", "precision", "partial correlation"])
def test_connectivity_measure(cov_estimator, kind, signals_and_covariance):
    signals, covs = signals_and_covariances
    conn_measure = ConnectivityMeasure(kind=kind, cov_estimator=cov_estimator)
    connectivities = conn_measure.fit_transform(signals)
    connections = (connectivities, conn_measure, covs)
    _assert_connectivities(connections, kind)

def _assert_connectivities(connections: tuple, kind: str) -> bool:
    if kind == "tangent":
        return _assert_tangent_connectivities(connections)
    if kind == "correlation":
        return _assert_correlation_connectivities(connections)
    ...

def _assert_tangent_connectivities(connections: tuple) -> bool:
    """Check positive definiteness if expected."""
    connectivities, covs, conn_measure = connections
    for k, cov_new in enumerate(connectivities):
        assert_array_almost_equal(cov_new, cov_new.T)
        gmean_sqrt = _map_eigenvalues(np.sqrt, conn_measure.mean_)
        assert is_spd(gmean_sqrt, decimal=7)
        assert is_spd(conn_measure.whitening_, decimal=7)
        assert_array_almost_equal(conn_measure.whitening_.dot(gmean_sqrt), np.eye(N_FEATURES))
        assert_array_almost_equal(
             gmean_sqrt.dot(_map_eigenvalues(np.exp, cov_new)).dot(gmean_sqrt), covs[k],
         )
    return True

.... 
```

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -692,6 +821,22 @@ def test_confounds_connectome_measure():
     confounds_df = DataFrame(confounds[:10])
     correlation_measure.fit_transform(signals, confounds=confounds_df)
 
+
+def test_confounds_connectome_measure_errors():
+    n_subjects = N_SUBJECTS
```
NicolasGensollen, line 826:
```suggestion
```

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -692,6 +821,22 @@ def test_confounds_connectome_measure():
...
+    # Generate signals and compute covariances and apply confounds while
+    # computing covariances
+    signals = []
+    for k in range(n_subjects):
```
NicolasGensollen, line 830:
```suggestion
    for k in range(N_SUBJECTS):
```

NicolasGensollen (commented):
Thanks @Remi-Gau !
This is looking good ! 👍 
I made a first pass with few suggestions. Feel free to take or leave.

nilearn/connectome/tests/test_connectivity_matrices.py
Remi-Gau: done

ymzayek (approved):
LGTM @Remi-Gau

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -425,208 +587,196 @@ def test_prec_to_partial():
...
+    connectivities = conn_measure.fit_transform(signals)
+
+    if kind == "tangent":
+        _assert_connectivity_tangent(connectivities, conn_measure, covs)
+    elif kind == "precision":
+        _assert_connectivity_precision(connectivities, covs)
+    elif kind == "correlation":
+        _assert_connectivity_correlation(connectivities, cov_estimator, covs)
+    elif kind == "partial correlation":
+        _assert_connectivity_partial_correlation(connectivities, covs)
```
NicolasGensollen, line 672: It could be nice to put the definitions of these functions just after this test instead of at the beginning of the file. This way, the users won't have to scroll up and down. WDYT ?

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -357,26 +510,31 @@ def test_sym_matrix_to_vec():
...
     sym = vec_to_sym_matrix(vec, diagonal=diagonal)
+
     assert_array_almost_equal(
         sym_matrix_to_vec(sym, discard_diagonal=True), vec
     )
 
     # multiple matrices case when diagonal is included
     vecs = np.asarray([vec, 2.0 * vec, 0.5 * vec])
     syms = vec_to_sym_matrix(vecs)
+
```
NicolasGensollen, line 531: I might be missing something, but I feel like we could split these in multiple tests (or a single test with parametrization maybe...). WDYT ?

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -306,40 +442,57 @@ def test_geometric_mean_properties():
...
 
 def test_geometric_mean_errors():
     n_features = 5
 
     # Non square input matrix
     mat1 = np.ones((n_features, n_features + 1))
-    pytest.raises(ValueError, _geometric_mean, [mat1])
+
+    with pytest.raises(ValueError, match="Expected a square matrix"):
+        _geometric_mean([mat1])
```
NicolasGensollen, line 480: I would do three different tests here, one for non square matrices, one for non spd matrices, and one for different shape matrices.

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -259,6 +187,192 @@ def random_non_singular(p, sing_min=1.0, sing_max=2.0, random_state=0):
...
+    # Test on sqrt map
+    spd_sqrt = np.array(
+        [[2.0, -1.0, 0.0], [-1.0, 2.0, -1.0], [0.0, -1.0, 2.0]]
+    )
+    spd = spd_sqrt.dot(spd_sqrt)
+    assert_array_almost_equal(_map_eigenvalues(np.sqrt, spd), spd_sqrt)
+
+    # Test on log map
+    spd = np.array([[1.25, 0.75], [0.75, 1.25]])
+    spd_log = np.array([[0.0, log(2.0)], [log(2.0), 0.0]])
```
NicolasGensollen, line 320: Same kind of comment, this can be split in three

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -259,6 +187,192 @@ def random_non_singular(p, sing_min=1.0, sing_max=2.0, random_state=0):
...
+            same_variance=False,
+        )
+        signals.append(signal)
+        signal -= signal.mean(axis=0)
+    return signals, confounds
+
+
+@pytest.fixture
+def signals():
+    signals, _ = _signals(N_SUBJECTS)
```
NicolasGensollen, line 258:
```suggestion
    return _signals(N_SUBJECTS)[0]
```

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -259,6 +187,192 @@ def random_non_singular(p, sing_min=1.0, sing_max=2.0, random_state=0):
...
+    return signals
```
NicolasGensollen, line 259:
```suggestion
```

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -259,6 +187,192 @@ def random_non_singular(p, sing_min=1.0, sing_max=2.0, random_state=0):
...
+        assert is_spd(cov_new, decimal=7)
+
+        d = np.sqrt(np.diag(np.diag(covs[k])))
+
+        if cov_estimator == EmpiricalCovariance():
+            assert_array_almost_equal(d.dot(cov_new).dot(d), covs[k])
+        assert_array_almost_equal(np.diag(cov_new), np.ones(N_FEATURES))
+
+
+def _assert_connectivity_partial_correlation(connectivities, covs):
```
NicolasGensollen, line 227: This might be some work, but it could be nice to add a docstring for each of these `_assert_connectivity...` functions explaining how they check the connectivity. It's not super trivial (to me at least...) when looking at the linalg code below.

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -259,6 +187,192 @@ def random_non_singular(p, sing_min=1.0, sing_max=2.0, random_state=0):
...
+    for k, cov_new in enumerate(connectivities):
```
NicolasGensollen, line 228: We don't need the index : 

```suggestion
    for cov, cov_new in zip(covs, connectivities):
```

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -259,6 +187,192 @@ def random_non_singular(p, sing_min=1.0, sing_max=2.0, random_state=0):
...
+        prec = linalg.inv(covs[k])
```
NicolasGensollen, line 229:
```suggestion
        prec = linalg.inv(cov)
```

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -23,6 +23,18 @@
 from sklearn.covariance import EmpiricalCovariance, LedoitWolf
 from sklearn.utils import check_random_state
 
+CONNECTIVITY_KINDS = [
```
NicolasGensollen, line 26: I'd expect a tuple for this (unless there is a reason I didn't see for using a list):

```suggestion
CONNECTIVITY_KINDS = (
```

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -23,6 +23,18 @@
...
+    "covariance",
+    "correlation",
+    "tangent",
+    "precision",
+    "partial correlation",
+]
```
NicolasGensollen, line 32:
```suggestion
)
```

NicolasGensollen (commented):
Thanks @Remi-Gau !
This is looking good ! 
I made some further suggestions as I think some tests can be broken in different pieces.

nilearn/connectome/tests/test_connectivity_matrices.py
Remi-Gau: yeah I went with everything at the top rule, but if a helper function is used by only one test it would help indeed.

still don't have a good rule for organizing this.

will try to make sure things are closer to where they are used if used only once

nilearn/connectome/tests/test_connectivity_matrices.py
Remi-Gau: yes probably.

will slit further

nilearn/connectome/tests/test_connectivity_matrices.py
Remi-Gau: not trivial to me either: but now that things are split it becomes easier to start deciphering each chunk

will try to add some explanations

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -259,6 +187,192 @@ def random_non_singular(p, sing_min=1.0, sing_max=2.0, random_state=0):
...
+
+def test_check_spd():
+    non_sym = np.array([[0, 1], [0, 0]])
+
+    with pytest.raises(
+        ValueError, match="Expected a symmetric positive definite matrix."
+    ):
+        _check_spd(non_sym)
+
+    non_spd = np.ones((3, 3))
```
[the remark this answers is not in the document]
Remi-Gau: nope except that I did not see the opportunity: good catch

nilearn/connectome/tests/test_connectivity_matrices.py
NicolasGensollen: That is what I understand also, hence relying on a tuple rather than a list.

bthirion (approved):
LGTM, thx !

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -306,327 +388,463 @@ def test_geometric_mean_properties():
...
+    ):
+        """Check output value properties for tangent connectivity measure \
+        that they have the expected relationship \
+        to the input covariance matrices.
+
+        - the geometric mean of the eigenvalues
+          of the mean covariance matrix is positive-definite
+        - the whitening matrix (used to transform the data \
+          also produces a positive-definite matrix
+        """
```
Remi-Gau, line 622: Adding some plain english explanation here but I would not mind having someone cross check that I want I am adding is correct: my linear algebra is not what it used to be.

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -306,327 +388,463 @@ def test_geometric_mean_properties():
...
+            ).dot(gmean_sqrt),
+            true_covariance_matrix,
+        )
+
+
+def _assert_connectivity_precision(connectivities, covs):
+    """Estimated precision matrix: \
+    - is positive definite, \
+    - its product with the true covariance matrix \
+      is close to the identity matrix."""
```
Remi-Gau, line 648: same here

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -306,327 +388,463 @@ def test_geometric_mean_properties():
...
+        - values on its diagonal should almost 1
+
+    If the covariance estimator is EmpiricalCovariance,
+    the product of:
+        - the square root of the diagonal of the true covariance matrix,
+        - the estimated covariance matrix,
+        - the square root of the diagonal of the true covariance matrix,
+
+    should be close to the true covariance matrix.
+    """
```
Remi-Gau, line 671: same here

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -306,327 +388,463 @@ def test_geometric_mean_properties():
...
+
+        # expected value
+        partial_corrlelation_matrix = -precision_matrix + 2 * np.diag(
+            np.diag(precision_matrix)
+        )
+
+        assert_array_almost_equal(
+            normalized_partial_correlation_matrix,
+            partial_corrlelation_matrix,
+        )
```
Remi-Gau, line 715: and here

Remi-Gau (commented):
@bthirion @NicolasGensollen 
I have added some comments in plain english and changed some varibles names in places to make things a bit more transparent.
But I would not mind have another pair of eyes to make sure it is correct.

Also let me know if you prefer one big doc string explaining the whole function or more inline comments: right now there is a bit of both.

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -306,327 +388,463 @@ def test_geometric_mean_properties():
...
+        assert_array_equal(input_covs[k], covs[k])
+
+        assert is_spd(covs[k], decimal=7)
+
+
+def _assert_connectivity_tangent(connectivities, conn_measure, covs):
+    for true_covariance_matrix, estimated_covariance_matrix in zip(
+        covs, connectivities
+    ):
+        """Check output value properties for tangent connectivity measure \
```
NicolasGensollen, line 614: It feels a bit weird to have the docstring within the for loop. Either move it after the function declaration, or define another function with this docstring that you will call iteratively.

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -306,327 +388,463 @@ def test_geometric_mean_properties():
...
+        assert_array_almost_equal(
+            estimated_covariance_matrix.dot(true_covariance_matrix),
+            np.eye(N_FEATURES),
+        )
+
+
+def _assert_connectivity_correlation(connectivities, cov_estimator, covs):
+    """Verify that the estimated covariance matrix: \
+        - is symmetric and positive definite
+        - values on its diagonal should almost 1
```
NicolasGensollen, line 662:
```suggestion
        - has values close to 1 on its diagonal
```

NicolasGensollen (approved):
Thanks @Remi-Gau !
I'll let @bthirion validate the correctness of the explanations.
In my opinion, the tests are MUCH more readable now, thanks a lot ! 👍

nilearn/connectome/tests/test_connectivity_matrices.py
Remi-Gau: Note that some of the steps in this functions are not explained in the doc string: mostly because I am not even entirely sure what the test actually does.

nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -295,8 +313,72 @@ def test_geometric_mean_properties():
...
+        vals_gmean, vecs_gmean = linalg.eigh(gmean)
+        gmean_inv_sqrt = _form_symmetric(np.sqrt, 1.0 / vals_gmean, vecs_gmean)
+        whitened_mats = [
+            gmean_inv_sqrt.dot(mat).dot(gmean_inv_sqrt) for mat in mats
+        ]
+        logs = [_map_eigenvalues(np.log, w_mat) for w_mat in whitened_mats]
+        logs_mean = np.mean(logs, axis=0)  # Covariant derivative is
+        # - gmean.dot(logms_mean)
+        norm = np.linalg.norm(logs_mean)  # Norm of the covariant derivative on
+        # the tangent space at point gmean
```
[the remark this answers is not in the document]
Remi-Gau: good point
moving them

Remi-Gau:
It seems this is good to go. Will merge in 24 hours in case anyone wants to have another look before this goes in.

PR #3605, opened 2023-03-18
[FIX] properly infer slice_time_ref from BIDS derivatives

### LINKED ISSUE #3299: first_level_from_bids warnings ignore slice_time_ref parameter
Hello - 

when running _first_level_from_bids_ structure with specified float valued slice_time_ref (e.g: 0.5 for default fmriprep'd data) the following warning:
    if t_r is not None:
        warn('RepetitionTime given in model_init as %d' % t_r)
        warn('slice_time_ref is %d percent of the repetition '
             'time' % slice_time_ref)
%d rounds the values to zero.

In addition 
> UserWarning: SliceTimingRef not found in file C:\Users\saptaf1/nilearn_data\ds000030/ds000030_R1.0.4/uncompressed\sub-10159\func\sub-10159_task-stopsignal_bold.json. It will be assumed that the slice timing reference is 0.0 percent of the repetition time. If it is not the case it will need to be set manually in the generated list of models
ignores specified slice_time_ref argument.

Nilearn version:
0.9.1
### Expected behavior
1.Display slice_time_ref (and the TR) as float in user warning
2.Don't ignore slice_time_ref argument in warning message display logic 

### Actual behavior
TR and slice_time_ref displayed as decimal integers in first warning and slice_time_ref argument ignored in the second warning.

Note this code also triggers the 2nd issue, because the TR is extracted from the json.

```python

from nilearn.datasets import (fetch_openneuro_dataset_index,
                              fetch_openneuro_dataset, select_from_index)

_, urls = fetch_openneuro_dataset_index()

exclusion_patterns = ['*group*', '*phenotype*', '*mriqc*',
                      '*parameter_plots*', '*physio_plots*',
                      '*space-fsaverage*', '*space-T1w*',
                      '*dwi*', '*beh*', '*task-bart*',
                      '*task-rest*', '*task-scap*', '*task-task*']
urls = select_from_index(
    urls, exclusion_filters=exclusion_patterns, n_subjects=1)

data_dir, _ = fetch_openneuro_dataset(urls=urls)

from nilearn.glm.first_level import first_level_from_bids
task_label = 'stopsignal'
space_label = 'MNI152NLin2009cAsym'
derivatives_folder = 'derivatives/fmriprep'
models, models_run_imgs, models_events, models_confounds = \
    first_level_from_bids(data_dir, task_label, space_label,
                          smoothing_fwhm=5.0,
                          derivatives_folder=derivatives_folder, slice_time_ref=0.5)

```

Remi-Gau:
Closes #3299

- decouple fetching `RepetitionTime` from fetching metadata related to slice timing
- start a deprecating cycle to change the default value of slice_time_ref from 0 to None
- add `t_r` and `slice_time_ref` validation for first level models
- extract functions to get metadata from bids sidecar json files
- add appropriate tests including regression test and test for deprecation warning


nilearn/_utils/data_gen.py
```diff
@@ -744,6 +748,21 @@ def basic_confounds(length, random_state=0):
     return confounds
 
 
+def add_metadata_to_bids_derivatives(bids_path: str | Path,
+                                      metadata: dict,
+                                      json_file: str = None) -> Path:
```
tsalo, line 753: The spacing looks off here. Any clue why flake8 didn't catch it?

```suggestion
def add_metadata_to_bids_derivatives(bids_path: str | Path,
                                     metadata: dict,
                                     json_file: str = None) -> Path:
```

tsalo: Also, shouldn't `black` reformat it to something like the following?

```suggestion
def add_metadata_to_bids_derivatives(
    bids_path: str | Path,
    metadata: dict,
    json_file: str = None,
) -> Path:
```

Remi-Gau: that's because flake8 and black are not yet active on those files

trying to keep a list of things that are covered here: https://github.com/nilearn/nilearn/projects/8#card-88537577

Otherwise check the long exception list in here to see what black works or does not work on.

https://github.com/nilearn/nilearn/blob/a744d100c52d6844f643343797eb36d08e04470d/pyproject.toml

so while reformatting is happening, expect some more "liberal" formatting.

nilearn/glm/first_level/first_level.py
```diff
@@ -959,6 +999,8 @@ def first_level_from_bids(dataset_path,
     models_events = []
     models_confounds = []
 
+    print(t_r)
```
Remi-Gau, line 1002: remove

nilearn/glm/first_level/first_level.py
```diff
@@ -1122,6 +1164,7 @@ def _get_processed_imgs(
         supported_filters=_bids_entities()["raw"]
         + _bids_entities()["derivatives"],
         extra_filter=img_filters,
+        verbose=verbose,
```
Remi-Gau, line 1167: I am having to pass this verbose parameter around a bit too much to be able to control whether warnings, messages should be printed...

This should probably be handled better by implementing a logger in another PR.

nilearn/_utils/data_gen.py
```diff
@@ -751,6 +750,25 @@ def basic_confounds(length, random_state=0):
     return confounds
 
 
+def add_metadata_to_bids_derivatives(bids_path: str | Path,
+                                     metadata: dict,
+                                     json_file: str = None) -> Path:
```
bthirion, line 755: I thought we were holding off for now

nilearn/glm/first_level/first_level.py
```diff
@@ -800,13 +807,31 @@ def _get_voxelwise_model_attribute(self, attribute,
...
+
+
 def first_level_from_bids(dataset_path,
                           task_label,
                           space_label=None,
                           sub_labels=None,
                           img_filters=None,
                           t_r=None,
-                          slice_time_ref=0.,
+                          slice_time_ref=None,
```
bthirion, line 834: Amigos, we need a deprecation warning before changing the API !

nilearn/glm/first_level/first_level.py
```diff
@@ -897,58 +922,73 @@ def first_level_from_bids(dataset_path,
...
-    # Get acq specs for models. RepetitionTime and SliceTimingReference.
+    # Get acq specs for models.
+    # RepetitionTime and StartTime for slice timing.
     # Throw warning if no bold.json is found
     if t_r is not None:
-        warn(f'RepetitionTime given in as {t_r}')
-        warn(f'slice_time_ref is {slice_time_ref} percent '
-             'of the repetition time')
+        _check_repetition_time(t_r)
+        warn("'RepetitionTime' given in model_init as {t_r}")
```
bthirion, line 930: I don't think we should have a warning here.

nilearn/glm/first_level/first_level.py
```diff
@@ -897,58 +922,73 @@ def first_level_from_bids(dataset_path,
...
+            t_r = _infer_repetition_time_from_dataset(
+            bids_path=dataset_path,
+            filters=filters,
+            verbose=verbose)
+    if t_r is not None:        
+        _check_repetition_time(t_r)
+    else:
+        warn("'t_r' not provided and cannot be inferred from metadata. " 
+             "It will need to be set manually in the list of models, "
+             "otherwise their fit will throw an exception.")
```
bthirion, line 961: Please test the warning.

nilearn/glm/first_level/first_level.py
```diff
@@ -897,58 +922,73 @@ def first_level_from_bids(dataset_path,
...
+
+    if slice_time_ref is not None:
+        _check_slice_time_ref(slice_time_ref)
+        warn("'slice_time_ref' given in model_init as {slice_time_ref}")
```
bthirion, line 965: We should not warn here.

nilearn/glm/tests/test_first_level.py
```diff
@@ -561,6 +565,195 @@ def test_first_level_glm_computation_with_memory_caching():
...
+                                                          t_r,
+                                                          error_type,
+                                                          error_msg):
+    bids_path = create_fake_bids_dataset(base_dir=tmp_path,
+                                         n_sub=1,
+                                         n_ses=1,
+                                         tasks=['main'],
+                                         n_runs=[1])
+
+    with pytest.raises(error_type, match = error_msg):
```
bthirion, line 609:
```suggestion
    with pytest.raises(error_type, match=error_msg):
```

nilearn/interfaces/bids/query.py
```diff
@@ -1,7 +1,152 @@
...
+        with open(json_files[0], 'r') as f:
+            specs = json.load(f)
+        value = specs.get(field)
+        if value is not None:
+            return value
+        else:
+            warn(f"'{field}' not found in file {json_files[0]}.")
+    else:
+        msg_suffix = f" in {bids_path}" if bids_path else ""
+        warn(f'No bold.json found in BIDS folder{msg_suffix}.')
```
bthirion, line 52: Should be a warning or an error ?

nilearn/interfaces/bids/query.py
```diff
@@ -1,7 +1,152 @@
...
+    
+    img_specs = get_bids_files(bids_path,
+                               modality_folder='func',
+                               file_tag='bold',
+                               file_type='json',
+                               filters=filters)
+    if not img_specs:
+        if verbose:
+            msg_suffix = f" in {bids_path}"
+            warn(f'No bold.json found in BIDS folder{msg_suffix}.')
```
bthirion, line 99: Same question here.

nilearn/interfaces/bids/query.py
```diff
@@ -1,7 +1,152 @@
...
+    img_specs = get_bids_files(main_path=bids_path,
+                               modality_folder='func',
+                               file_tag='bold',
+                               file_type='json',
+                               filters=filters)
+    
+    if not img_specs:
+        if verbose:
+            msg_suffix = f" in {bids_path}"
+            warn(f'No bold.json found in BIDS folder{msg_suffix}.')
```
bthirion, line 143: and here.

nilearn/interfaces/bids/query.py
```diff
@@ -1,7 +1,152 @@
...
+    Returns
+    -------
+    Any
+        Value of the field or None if the field is not found.
+
+    """
+    # slice timing metadata can only be found in derivatives
+
+    if not Path(bids_path).exists():
+        warn(f"derivatives_path {bids_path} does not exist.")
```
bthirion, line 88: Shouldn't this exist anyhow ?

bthirion (commented):
Thx. Please find a few comments.
The main concern is about API change.

nilearn/_utils/data_gen.py
Remi-Gau: you are correct
I had added them before we made a decision: I thought I had removed all of them but I missed a fews

nilearn/glm/first_level/first_level.py
Remi-Gau: I actually need to maybe rethink whether the change of API is even desirable / necessary.

nilearn/glm/first_level/first_level.py
Remi-Gau: Gonna try to make a small table about the different situations (input argument VS what metadata exist in which dataset) to clarify when to throw warnings or errors.

ymzayek: This refactoring isn't very useful if `slice_time_ref` default is not changed to None. If we do this, we need to deprecate it properly as @bthirion pointed out. In any case, given the explanation in the docstrings, this warning becomes redundant with the one under it.

Remi-Gau:
tagging @ymzayek @bthirion @emdupre @tsalo as you are the ones who have interacted with this PR

I tried to write some sort of "behavior requirements" regarding this PR: should probably have started with this rather than getting coding guns blazing, but better later than never.

This is meant to help us decide when to throw warning / errors and also help decide what the default values should be (and hence if we need to change the API).

Can you have a look at the tables below and let me know if you agree with their content?

## `t_r` related behaviors

### `t_r` can be inferred from BIDS

 inputs | behavior when `t_r` inferred BIDS derivatives | behavior when `t_r` inferred from BIDS raw
-- | -- | --
`t_r = None` (default) | t_r value of derivative takes precedence; no error; no warning | use `t_r` value of raw; no error; no warning
t_r provided matches value found in BIDS | no error; no warning | no error; no warning
t_r provided does not match value found in BIDS | throw an error | throw an error

Note that we may want to let users use the "wrong" repetition time and just throw a warning in case of mismatch between the `RepetitionTime` found in the BIDS dataset and the `t_r` value provided.

### `t_r` cannot be inferred from BIDS raw or derivatives

This should be rare as `RepetitionTime` is a required metadata for BIDS raw datasets. This may happen if somehow the relevant JSON file not picked by nilearn `get_bids_file` (possibly because of the BIDS inheritance principle?).

 inputs | behavior
-- | --
t_r = None | throw warning no t_r found: a t_r will need to be provided
t_r provided | no error, no warning

## `slice_timing_ref` related behaviors

### `slice_timing_ref` can be inferred from BIDS

Should be the case for datasets from fmriprep >= 20.2.4

inputs | behavior |
-- | -- |
slice_time_ref = None | use slice_timing_ref of BIDS derivatives |
slice_time_ref provided; matches value inferred from BIDS | no error, no warning |
slice_time_ref provided; does not match value inferred from BIDS | throw an error |

Note that we may want to let users use the "wrong" slice_time_ref and just throw a warning in case of mismatch between the `slice_time_ref` inferred from the BIDS dataset and the `t_r` value provided.

### `slice_timing_ref` cannot be inferred from BIDS

This may happen with (from most likely to more edge case):
- "older" fmriprep datasets (< 20.2.4)
- with processing pipelines that do not apply the suggestions from this [BIDS issue](https://github.com/bids-standard/bids-specification/issues/836) (as this is not yet official BIDS)
- or if the the `StartTime` is in JSON in the BIDS dataset that is not somehow not picked by nilearn `get_bids_file` (possibly because of the BIDS inheritance principle?).

inputs | behavior |
-- | -- |
slice_time_ref = None | set slice_timing_ref to 0
slice_time_ref provided | no error, no warning

### Change API default value

We could decide to NOT change the default value of `slice_timing_ref` to None but in this case users would have to "opt-in" to let nilearn infer the its value by actively passing `slice_timing_ref=None`.

I suspect many users may not be aware of those BIDS / fmriprep subtleties, so having something like this be opt-in rather than a default feels undesirable from my point of view.

Curious to hear others' thoughts.

## TODOs

- [ ] try to infer t_r and slice_time_ref even if they are provided through the arguments
- [ ] implement error when inferred t_r differs from provided t_r
- [ ] implement error when inferred slice_time_ref differs from provided slice_time_ref
- [ ] test ability of get_bids_file to cope with the BIDS inheritance principle (not directly needed for this PR but useful to know what are the "blind spots" of this function)

nilearn/glm/first_level/first_level.py
Remi-Gau: will remove the warning

doc/changes/latest.rst
```diff
@@ -15,11 +15,14 @@ NEW
 Fixes
 -----
 
+- Change calculation of TR in :func:`~.glm.first_level.compute_regressor` to be more precise (:gh:`3362` by `Anne-Sophie Kieslinger`_)
+
```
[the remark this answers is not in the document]
Remi-Gau: yup totally what happened.
thanks for spotting it

nilearn/interfaces/bids/query.py
Remi-Gau: I think I would stick to warning here: see my new comment in the PR.


https://github.com/nilearn/nilearn/pull/3605#issuecomment-1490567765

I think the errors should be thrown by some other part of the code depending on what this function returns and what argument was passed to `first_level_from_bids`.

Throwing may help users understand why a subsequent error happened but missing json may not be a sufficient reason to throw an error.

nilearn/_utils/data_gen.py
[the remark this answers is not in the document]
Remi-Gau: Those are much more related to the generation of dummy BIDS datasets rather than being usable on actual BIDS dataset.

Does that make sense?

nilearn/interfaces/tests/test_bids.py
```diff
@@ -12,13 +13,188 @@
...
+                               filters = [ ('task', 'main')],
+                               sub_folder=False)
+    assert selection != []
+    assert selection[0] == str(json_file)
+
+
+@pytest.mark.xfail(
+        reason=("get_bids_files does not find json files"
+                " that are directly in the subject folder of a dataset.")
+        )
```
Remi-Gau, line 154: I have not seen xfail used in the test suite so not sure if there is any guideline about using them. 

https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

Will make a strict xfail so that an XPASS (“unexpectedly passing”) results from this test will fail the test suite.

bthirion:
I think I agree with your proposition: I'm also hesitating on the case of incompatibility between inferred and provided value, for both TR and slice_time_ref. I think I'd rather spit a warning to let the user know he/she's
doing something strange, but I would still allow it.

nilearn/glm/first_level/first_level.py
ymzayek: @Remi-Gau Yes I agree that if we don't have to change the API, better not. But this implementation will only infer the slice_time_ref if the user passes None, while the old behavior tries to infer it if the default, which is `0.`, is kept. So a user that doesn't change their code that uses the default option will now get `0.` set as their slice_time_ref while they expect it to be inferred. Does that make sense? Or maybe I have the logic wrong. If what I'm saying follows, then we'd need to go through a deprecation cycle as well.

nilearn/_utils/data_gen.py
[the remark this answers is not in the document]
ymzayek: Yes!

nilearn/glm/first_level/first_level.py
Remi-Gau: If I am correct the old behavior was:
- try to infer `t_r` AND `slice_time_ref` **if and only if** `t_r = None` is passed (which was the default)
- if a `slice_time_ref` is found then it overrides ANY value passed as argument.

Making more tables to compare old and new behavior depending on datasets and arguments...

Remi-Gau:
## comparing old and new behavior

See spreadsheet

https://docs.google.com/spreadsheets/d/1Zco4l6T-Bc9t1m7WfujHCC5c4EJwwm_GH8gU-ADitA8/edit?usp=sharing

Using some "actual" values for what could be inferred from a dataset to make this a bit less abstract.

### TL;DR: 

For users who just rely on the defaults this should not lead to a change in behavior. 

Well except for the fact that the old code was not looking for the right metadata for slice_time_ref in BIDS derivatives dataset, so it was unlikely to give the right answer.

Speaks in favor of:

- start deprecating API for slice_time_ref to start using
- use only warning when there is a mistmatch between passed value and inferred value

### old behavior

Note I think that passing None for slice_time_ref may lead to errors when setting up the first level model but not 100% sure.

The Third line is the default.

  |   | t_r (3) and slice_time_ref (0.75) inferred | t_r and slice_time_ref cannot be inferred | t_r (3) only can be inferred | slice_time_ref (0.75) only can be inferred
-- | -- | -- | -- | -- | --
t_r=None | slice_time_ref=None | t_r=3; slice_time_ref=0.75 | t_r=None; slice_time_ref=None + missing t_r warning | t_r=3; slice_time_ref=None + downstream error? | t_r=None; slice_time_ref=0.75 + missing t_r warning
t_r=2 | slice_time_ref=None | t_r=2; slice_time_ref=None + downstream error? | t_r=2; slice_time_ref=None + downstream error? | t_r=2; slice_time_ref=None + downstream error? | t_r=2; slice_time_ref=None + downstream error?
t_r=None | slice_time_ref=0.0 | t_r=3; slice_time_ref=0.75 | t_r=None; slice_time_ref=0.0 + missing t_r warning | t_r=3; slice_time_ref=0.0 | t_r=None; slice_time_ref=0.75 + missing t_r warning
t_r=2 | slice_time_ref=0.0 | t_r=2; slice_time_ref=0.0 | t_r=2; slice_time_ref=0.0 | t_r=2; slice_time_ref=0.0 | t_r=2; slice_time_ref=0.0
t_r=None | slice_time_ref=0.5 | t_r=3; slice_time_ref=0.75 | t_r=None; slice_time_ref=0.5 + missing t_r warning | t_r=3; slice_time_ref=0.5 | t_r=None; slice_time_ref=0.75 + missing t_r warning
t_r=2 | slice_time_ref=0.5 | t_r=2; slice_time_ref=0.5 | t_r=2; slice_time_ref=0.5 | t_r=2; slice_time_ref=0.5 | t_r=2; slice_time_ref=0.5

### new behavior

The first line would be the new default after the deprecation cycle.

  |   | t_r (3) and slice_time_ref (0.75) inferred | t_r and slice_time_ref cannot be inferred | t_r (3) only can be inferred | slice_time_ref (0.75) only can be inferred
-- | -- | -- | -- | -- | --
t_r=None | slice_time_ref=None | t_r=3; slice_time_ref=0.75 | t_r=None; slice_time_ref=0.0 + missing t_r warning | t_r=3; slice_time_ref=0.0 | t_r=None; slice_time_ref=0.75 + missing t_r warning
t_r=2 | slice_time_ref=None | t_r=2; slice_time_ref=0.75 + mismatch warning | t_r=2; slice_time_ref=0 | t_r=2; slice_time_ref=0 + mismatch warning | t_r=2; slice_time_ref=0.75
t_r=None | slice_time_ref=0.0 | t_r=3; slice_time_ref=0.0 + mismatch warning | t_r=None; slice_time_ref=0.0 + missing t_r warning | t_r=3; slice_time_ref=0.0 | t_r=None; slice_time_ref=0.0 + missing t_r warning + mismatch warning
t_r=2 | slice_time_ref=0.0 | t_r=2; slice_time_ref=0.0 + mismatch warning | t_r=2; slice_time_ref=0.0 | t_r=2; slice_time_ref=0.0 + mismatch warning | t_r=2; slice_time_ref=0.0 + mismatch warning
t_r=None | slice_time_ref=0.5 | t_r=3; slice_time_ref=0.5 + mismatch warning | t_r=None; slice_time_ref=0.5 + missing t_r warning | t_r=3; slice_time_ref=0.5 | t_r=None; slice_time_ref=0.5 + missing t_r warning + mismatch warning
t_r=2 | slice_time_ref=0.5 | t_r=2; slice_time_ref=0.5 + mismatch warning | t_r=2; slice_time_ref=0.5 | t_r=2; slice_time_ref=0.5 + mismatch warning | t_r=2; slice_time_ref=0.5 + mismatch warning

doc/changes/latest.rst
```diff
@@ -15,11 +15,12 @@ NEW
 Fixes
 -----
 
+- Improve how :func:`~.glm.first_level.first_level_from_bids` handles fetching slice timing metadata and add additional input validation. In release ``0.XX.XX`` the default `slice_time_ref` will be `None` instead of `0` (:gh:`3605` by `Rémi Gau`_).
```
Remi-Gau, line 18: not sure what number to put for the deprecation version here

nilearn/glm/first_level/first_level.py
```diff
@@ -859,6 +882,17 @@ def first_level_from_bids(dataset_path,
...
+    slice_time_ref : float, optional
+        This parameter indicates the time of the reference slice used in the
+        slice timing preprocessing step of the experimental runs. It is
+        expressed as a percentage of the t_r (time repetition), so it can have
+        values between 0. and 1. Default=0.
+
+        .. deprecated:: 0.10.1
+
+            The default=0 for ``slice_time_ref`` will be deprecated.
+            The default value will change to 'None' in 0.xx.
```
Remi-Gau, line 894: not sure what number to put for the deprecation version here

nilearn/glm/first_level/first_level.py
```diff
@@ -885,72 +919,106 @@ def first_level_from_bids(dataset_path,
         Items for the FirstLevelModel fit function of their respective model.
 
     """
+
+    if slice_time_ref == 0:
+        warn(
+            'Starting in version 0.xx, slice_time_ref will default to None.',
```
Remi-Gau, line 925: not sure what number to put for the deprecation version here

Remi-Gau (commented):
implemented the other changes suggested and a deprecation warning

just not sure how long the deprecation cycles are.

nilearn/glm/first_level/first_level.py
```diff
@@ -800,13 +805,31 @@ def _get_voxelwise_model_attribute(self, attribute,
...
+
+
 def first_level_from_bids(dataset_path,
                           task_label,
                           space_label=None,
                           sub_labels=None,
                           img_filters=None,
                           t_r=None,
-                          slice_time_ref=0.,
+                          slice_time_ref=0,
```
Remi-Gau, line 832:
```suggestion
                          slice_time_ref=0.0,
```

ymzayek:
@Remi-Gau thanks a lot for these tables! The warning/errors table seems fine to me.

Regarding the second table, I do see that the old vs new default behavior is the same but what I meant was that if users who use the default parameter values but explicitly define them such as `slice_time_ref=0.` in their code then in this (perhaps edge) case, the behavior would be different. In any case, this is all fine if we go through deprecation.

doc/changes/latest.rst
Remi-Gau: OK that makes sense to me

Remi-Gau:
> then in this (perhaps edge) case, the behavior would be different

yup your comment just prompted me to actually try to map most cases to get a better idea because I was not even clear about it myself.

Remi-Gau:
Todo

- [ ] add test to specifically catch the deprecation warning.

nilearn/_utils/data_gen.py
```diff
@@ -751,6 +750,25 @@ def basic_confounds(length, random_state=0):
     return confounds
 
 
+def add_metadata_to_bids_dataset(bids_path,
+                                     metadata,
+                                     json_file=None):
+    if json_file is None:
```
bthirion, line 756: if the function is meant to be part of the public API, it should have a docstring.
Otherwise the function name should have a leading underscore

nilearn/_utils/data_gen.py
```diff
@@ -751,6 +750,25 @@ def basic_confounds(length, random_state=0):
...
+        json_file = (Path(bids_path) / 
+                        'derivatives' / 
+                        'sub-01' / 
+                        'ses-01' /
```
bthirion, line 760: On which grounding are this subject and session id chosen ?

bthirion: I guess it's because this is only used for testing. Maybe indicate this in a short docstring.

nilearn/_utils/tests/test_data_gen.py
```diff
@@ -15,6 +14,38 @@
...
 
+from nilearn._utils.data_gen import add_metadata_to_bids_dataset
+
+
+def test_add_metadata_to_bids_derivatives(tmp_path):
+    # bare bone smoke test
+    target_dir = tmp_path / 'derivatives' / 'sub-01' / 'ses-01' / 'func'
+    target_dir.mkdir(parents=True)
+    json_file = add_metadata_to_bids_dataset(bids_path=tmp_path,
+                                                  metadata={"foo": "bar"})
```
bthirion, line 25: I'm suprised by the code indentation here. What magic leads to that ?

nilearn/_utils/tests/test_data_gen.py
```diff
@@ -15,6 +14,38 @@
...
+        assert metadata == {"foo": "bar"}
+
+
+def test_add_metadata_to_bids_derivatives_with_json_path(tmp_path):
+    # bare bone smoke test
+    target_dir = tmp_path / 'derivatives' / 'sub-02' 
+    target_dir.mkdir(parents=True)
+    json_file='derivatives/sub-02/sub-02_task-main_bold.json'
+    json_file = add_metadata_to_bids_dataset(bids_path=tmp_path,
+                                                  metadata={"foo": "bar"},
```
bthirion, line 40: Same thing here.

nilearn/glm/first_level/first_level.py
```diff
@@ -864,6 +891,17 @@ def first_level_from_bids(dataset_path,
         Filter examples would be ('desc', 'preproc'), ('dir', 'pa')
         and ('run', '10').
 
+    slice_time_ref : :obj:`float` between 0.0 and 1.0, optional
+        This parameter indicates the time of the reference slice used in the
+        slice timing preprocessing step of the experimental runs. It is
+        expressed as a percentage of the t_r (time repetition), so it can have
```
bthirion, line 897: Actually, it's not a percentage  it's a fraction

nilearn/glm/tests/test_first_level.py
```diff
@@ -573,6 +575,197 @@ def test_first_level_glm_computation_with_memory_caching():
         del mask, func_img, FUNCFILE, model
 
 
+@pytest.mark.parametrize('t_r, warning_msg',
+                         [(None, "No bold.json .* BIDS")])
+def test_first_level_from_bids_set_repetition_time_warnings(tmp_path,
+                                                            t_r,
+                                                            warning_msg):
+
```
bthirion, line 583: The naming is great, but i'd not be against a short docstring...

nilearn/glm/tests/test_first_level.py
```diff
@@ -573,6 +575,197 @@ def test_first_level_glm_computation_with_memory_caching():
...
+    with pytest.warns(UserWarning,
+                      match="RepetitionTime' not found in file"):
+        models, *_ = first_level_from_bids(
+            dataset_path=str(tmp_path / bids_path),
+            task_label='main',
+            space_label='MNI',
+            img_filters=[('desc', 'preproc')],
+            slice_time_ref=None)
+        assert models[0].t_r == 1.5
```
bthirion, line 765: Sorry, but where does this 1.5 come from ?

nilearn/interfaces/tests/test_bids.py
```diff
@@ -16,9 +18,177 @@
...
+from nilearn.interfaces.bids.query import (
+    _get_metadata_from_bids,
+    _infer_repetition_time_from_dataset,
+    _infer_slice_timing_start_time_from_dataset,
+)
 from nilearn.maskers import NiftiMasker
 
 
+def test_get_metadata_from_bids(tmp_path):
+
```
bthirion, line 30: a one-linear to explain what this does ?

nilearn/_utils/data_gen.py
Remi-Gau: good point

nilearn/_utils/tests/test_data_gen.py
Remi-Gau: Until #3614 is in, the files under the _utils folder will not checked by flake8 so "technically valid albeit interesting" formats are to be expected.

We could wait for #3614 to be in before we merge this PR, or the other way around: I will fix those formatting you are flagging here, but just want to let you know that any "format drift" you see should be fixed soon.

nilearn/glm/first_level/first_level.py
Remi-Gau: indeed

nilearn/glm/tests/test_first_level.py
Remi-Gau: ah yeah that definitely feels like a magic number: this is because the function that generate a BIDS dataset used for testing creates timeseries with a TR of 1.5 seconds: will explain than in the test docstring

nilearn/_utils/data_gen.py
Remi-Gau: done

nilearn/glm/first_level/first_level.py
Remi-Gau: changed it also in the FirstLevelModel doc string

nilearn/glm/tests/test_first_level.py
Remi-Gau: added some explanations: let me know if that's enough

nilearn/glm/tests/test_first_level.py
Remi-Gau: added a comment

nilearn/interfaces/tests/test_bids.py
Remi-Gau: added some explanations: let me know if that's enough

nilearn/glm/first_level/first_level.py
```diff
@@ -805,13 +812,31 @@ def _get_voxelwise_model_attribute(self, attribute,
         return output
 
 
+def _check_repetition_time(t_r):
+    if not isinstance(t_r, (float, int)):
```
bthirion, line 816: One-line docstring welcome

nilearn/glm/first_level/first_level.py
```diff
@@ -805,13 +812,31 @@ def _get_voxelwise_model_attribute(self, attribute,
...
+        raise TypeError("'t_r' must be a float or an integer. "
+                        f"Got {type(t_r)} instead.")
+    if t_r <= 0:
+        raise ValueError("'t_r' must be positive. "
+                         f"Got {t_r} instead.")
+
+
+def _check_slice_time_ref(slice_time_ref):
```
bthirion, line 824: One-line docstring welcome

nilearn/interfaces/tests/test_bids.py
```diff
@@ -16,9 +18,198 @@
...
+    
+    t_r = _infer_repetition_time_from_dataset(
+        bids_path=tmp_path / bids_path / 'derivatives',
+        filters=[('task', 'main'), ('run', '01')])
+    
+    assert t_r == expected_t_r
+
+
+def test_infer_slice_timing_start_time_from_dataset(tmp_path):
+
```
bthirion, line 92: Can you add a docstring here ?

bthirion (commented):
Almost there ! Thx

nilearn/glm/first_level/first_level.py
Remi-Gau: done

nilearn/interfaces/tests/test_bids.py
Remi-Gau: done

bthirion (approved):
LGTM, thx.

ymzayek:
@Remi-Gau can you rebase on main and resolve the conflict? Then this can be merged.

Remi-Gau:
rebased + green CI = merging

Issue #3085, opened 2021-12-10
correspondance between atlas region labels and `NiftiLabelsMasker`'s output dimensions


jeromedockes:
`NiftiLabelsMasker.__init__` takes a discrete atlas (3D map of integers, `labels_img`) and a list of labels (region names, strings, `labels`).
It is assumed that values in the image are region ids between 0 and N, where N is `len(labels)`, and that the region where the image takes the value `j` (0 ≤ j < N) corresponds to the human-readable name `labels[j]`.
Some regions can be empty in the original atlas image or become empty after this atlas image is resampled to the resolution of the data that the masker must transform.
The output of `NiftiLabelsMasker.transform(func_img)` is a vector where each entry contains the average of `func_img` in a given region of the (possibly resampled) atlas image.
Empty regions in the atlas image are discarded: they are not assigned an entry (dimension) in the output of `transform` (we cannot compute the average of 0 voxels).
Moreover, the background (usually the region with value `0` in the atlas image, although that can be controlled) is also discarded.
(For some reason, if a region becomes empty due to computing intersection with the `mask_img` --we can provide a mask in addition to the atlas image, and everything outside the mask is considered as background-- then it still has a dimension in the output but that gets filled with `0` [here](https://github.com/nilearn/nilearn/blob/2fd6665664f27dfbf1f10aeeba660adc9b560361/nilearn/regions/signal_extraction.py#L132), which seems a bit inconsistent.)
Therefore, there is not a direct correspondance between entries of `transform`'s output and entries in `labels` or values in the original atlas image.
This means that `masker.transform(img).T[j]` does not correspond to the region whose id is `j` and whose label is `labels[j]`.

To handle this, the `NiftiLabelsMasker` has a `labels_` attribute that allows to map values in the atlas image (region ids) to dimensions in the output of `transform` (the masked image).
It is a vector of integers of length `M`, where M ≤ N is the number of dimensions in the masked image, and for 0 ≤ j < M, `labels_[j]` gives the corresponding atlas region id. 
That means that the region where the atlas image takes value `masker.labels_[j]` gets averaged to produce `masked_img.T[j]`, and that the corresponding label (region name string) is `labels[labels_[j]]`.

To summarize: the `j`th entry of `masker.transform(img)` corresponds to the region name `masker.labels[masker.labels_[j]]`.
However, this is not documented (and arguably not straightforward, and the name `labels_` is not very explicit and confusingly similar to `labels`).
[...]

PR #3616, opened 2023-03-23
remove empty regions from signals when applying  LabelMasker.transform

mtorabi59:
Addresses #3085.

This PR is to address the issue of empty `signals` dimensions after applying `labelmasker.transform`. So, basically we first apply the additional mask and then find the list of all existing labels, instead of first finding them and then setting them to zero. This way they will be treated as background from the beginning. As a result, no dimension will be dedicated to those empty regions in `signals` output anymore.
I've also added a test function designated to test the NiftiLabelMasker module.

### DIFF
--- a/nilearn/mass_univariate/tests/test__utils.py
+++ b/nilearn/mass_univariate/tests/test__utils.py
@@ -12,7 +12,41 @@ from scipy.ndimage import generate_binary_structure
 from sklearn.utils import check_random_state
 
 
-def test__calculate_tfce():
+@pytest.fixture
+def null():
+    """Return a dummy null distribution that can be reused across tests."""
+    return [-10, -9, -9, -3, -2, -1, -1, 0, 1, 1, 1, 2, 3, 3, 4, 4, 7, 8, 8, 9]
+
+
+@pytest.fixture
+def test_arr4d():
+    test_arr4d = np.zeros((10, 10, 10, 1))
+    test_arr4d[:2, :2, :2, 0] = 5  # 8-voxel cluster, high intensity
+    test_arr4d[7:, 7:, 7:, 0] = 1  # 27-voxel cluster, low intensity
+    test_arr4d[6, 6, 6, 0] = 1  # corner touching second cluster
+    test_arr4d[6, 6, 8, 0] = 1  # edge touching second cluster
+    test_arr4d[3:5, 3:5, 3:5, 0] = -10  # negative cluster, very high intensity
+    test_arr4d[5:6, 3:5, 3:5, 0] = 1  # cluster touching negative one
+    return test_arr4d
+
+
+@pytest.mark.parametrize(
+    "two_sided_test, dh, true_max_tfce",
+    [
+        (
+            False,
+            "auto",
+            5050,
+        ),  # One-sided where positive cluster has highest TFCE
+        (
+            True,
+            "auto",
+            5555,
+        ),  # Two-sided where negative cluster has highest TFCE
+        (False, 1, 550),  # One-sided with preset dh
+    ],
+)
+def test_calculate_tfce(two_sided_test, dh, true_max_tfce):
     """Test _calculate_tfce."""
     test_arr4d = np.zeros((10, 10, 10, 1))
     bin_struct = generate_binary_structure(3, 1)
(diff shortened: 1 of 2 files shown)


Remi-Gau:
@mtorabi59 given how much change there is in some folders because of the reformatting, watch out for merge conflicts theses days !

I will let solve this one, but ping me if you need help.

ymzayek:
Hi @mtorabi59 thanks for working on this! A couple of tips: 

Tests for modules are usually contained within that module under a `tests` directory. Since `signal_extraction.py` is in `nilearn/regions` you will find the relevant test file in `nilearn/regions/tests/test_signal_extraction.py`. So the first thing when changing code in a file is to look to see if any of the already written tests will fail. Indeed this is the case because when I run pytest on that test file I see that `test_signal_extraction_with_maps_and_labels` will fail. So what you could do first is see how you can modify the failing test to test the new behavior and get it to pass.

It may be useful sometimes to add tests at higher levels which I think is why you are testing `NiftiLabelsMasker`. However, this is in the wrong location. Such tests should go in `nilearn/maskers/tests/test_nifti_labels_masker.py`. I know this may be confused with the `masking` module so just to clarify `nilearn/masking.py` and `nilearn/maskers/` are two different modules. So first look at the tests in `test_nifti_labels_masker.py` to see what could be modified or added to test the new behavior (but first run pytest on this test file to check if anything already fails)

Let me know if you have any questions

Remi-Gau:
There is actually other tests in `nilearn/tests/test_masking.py` that probably should go `nilearn/maskers/tests/test_nifti_labels_masker.py`

See for example: 

https://github.com/nilearn/nilearn/blob/32835204227f65204b911fb2fd9ab029968a2182/nilearn/tests/test_masking.py#L552

I am trying to put them back where they belong to avoid confusion.

nilearn/tests/test_masking.py
```diff
@@ -482,6 +482,99 @@ def test_intersect_masks():
...
+
+    # apply atlas_img on img
+    masker = NiftiLabelsMasker(atlas_img,
+                           labels=labels,
+                           standardize=False
+                           )
+    masker.fit()
+    signals = masker.transform(img)
+    expected_signal = np.tile(np.array([[1, 2, 3]]), (10, 1))
+    assert_array_equal(expected_signal, signals)
```
Remi-Gau, line 546:
```suggestion
    assert_array_equal(signals, expected_signal)
```

I have seen both in the nilearn test suite, but I generally find

```python
assert results == expected_result
```

more readable than 

```python
assert expected_result == results
```

For the same reason that we more often write

```python
if length >= 10:
```

than this

```python
if 10 <= length:
```

---

The rule from: ["the art of readable code"](https://mcusoft.files.wordpress.com/2015/04/the-art-of-readable-code.pdf)

### Left-hand side	
The expression “being interrogated,” whose value is more in flux.

### Right-hand side
The expression being compared against, whose value is more constant.

mtorabi59: Thanks @Remi-Gau  for explaining this !

mtorabi59:
So I realized that the `img_to_signals_maps` also has the same behavior of returning empty regions and that is why the test fails when comparing the results of `img_to_signals_maps` and `signals_to_img_labels`. I have to solve the issue for `img_to_signals_maps` too so that they give consistent outputs. Should I do change both in the same PR or I should change  `signals_to_img_labels` and comment that test for now until the next PR that I change `img_to_signals_maps`?

mtorabi59:
Also, should I put a DeprecationWarning for the change? e.g. "Starting in version x, the regions removed by mask_img will be removed from region_signals instead of being filled by zeros."

ymzayek:
@mtorabi59 I think in general it's better to keep PRs open and not overwrite the code changes if there is an issue so that we can better discuss the next steps with context. Sometimes things like that come up and we could either decide to handle them in the same PR or open another one. In any case let's take a closer look regarding the consistency of  behavior of those two functions. And indeed if the output changes in public functions we would need to go through a deprecation cycle.

mtorabi59:
@ymzayek yes you are right. I didn't know that syncing my fork will close the PR automatically. It wasn't intended. Yes I had a closer look. Should we discuss that in [the issue #3085](https://github.com/nilearn/nilearn/issues/3085)? or in office hours?

ymzayek:
@mtorabi59 ok yes makes sense what happened, now worries! Yes discussing it either in the issue or office hours (or both) is fine.

PR #3626, opened 2023-03-27
[ENH] Special handling of ``param_grid``/``cv_params_`` for Decoder objects using models with built-in CV

### LINKED ISSUE #3599: Decoders using `RidgeCV`/`RidgeClassifierCV` may not always be taking advantage of built-in CV capabilities
Some Decoder objects can use underlying `sklearn` models with built-in cross-validation (e.g., [`RidgeCV`](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.RidgeCV.html) and [`RidgeClassifierCV`](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.RidgeClassifierCV.html)). In those cases the `param_grid` input and `cv_params_` fitted attribute are handled in the same way as for the models without built-in CV, which leads to potentially suboptimal/unintuitive behaviour:
1. If a typical dict of sequences of hyperparameter values (e.g., `{'alphas': [0.1, 1, 10]}`) is passed to `param_grid`, then the Decoder will fit one `RidgeCV`/`RidgeClassifierCV` model per hyperparameter value, which means that the model's built-in CV will not be used. Instead, to take advantage of the built-in CV, the user has to wrap the sequence of hyperparameter values in another sequence (see code snippet below), but that is not intuitive and is inconsistent with the models without built-in CV.
2. If the user does pass in a `param_grid` like `{'alphas': [[0.1, 1, 10]]}` as described above, the `cv_params_` attribute of the fitted Decoder will show the full sequence (`[0.1, 1, 10]` here) for each CV iteration (see example below), which is unexpected because there should only be one value per outer CV iteration (per hyperparameter).
3. `cv_params_` will be empty if `param_grid` is not given, even though the model still does CV on the default `alphas` sequence.

Here is some code (adapted from [here](https://github.com/nilearn/nilearn/issues/2886#issuecomment-868760488)) illustrating some of my points above:
```python
from nilearn import datasets
from sklearn.model_selection import train_test_split
from nilearn.decoding import DecoderRegressor

n_subjects = 10 
oasis_dataset = datasets.fetch_oasis_vbm(n_subjects=n_subjects)
gray_matter_map_filenames = oasis_dataset.gray_matter_maps
age = oasis_dataset.ext_vars['age'].astype(float)
gm_imgs_train, gm_imgs_test, age_train, age_test = train_test_split(
    gray_matter_map_filenames, age, train_size=.6, random_state=0)

decoder = DecoderRegressor(
    estimator='ridge', 
    param_grid = {
        'alphas':[[0.1, 1, 10]], # list of list
    },
    cv=2,
)

decoder.fit(gm_imgs_train, age_train)
print(decoder.cv_params_) # {'beta': {'alphas': [[0.1, 1, 10], [0.1, 1, 10]]}}
```

I'd like to make a PR to address these points. My proposed changes for the case where the model is `RidgeCV` or `RidgeClassifierCV` are:
1. If `param_grid` is a sequence of values, wrap it in a list.
2. Within the [inner CV loop](https://github.com/nilearn/nilearn/blob/c33cca7a/nilearn/decoding/decoder.py#L220) (which iterates over hyperparameter values), set `best_param` to the single best value for each hyperparameter instead of the full list. This value should be available from the fitted estimator (e.g., in the `alpha_` attribute for `RidgeCV`/`RidgeClassifierCV`).
3. @jeromedockes also suggested we overwrite the default `alphas` value with a large list of 10-20 log-spaced values, so in most cases the user would not have to specify their own `param_grid`.

### Benefits to the change
1. Users would be able to benefit from the built-in CV capabilities of the models even if they specify `param_grid` as they would for other models.
2. The `cv_params_` attribute of the fitted Decoder would be more informative.

michellewang:
Closes #3599.

- Add default `"alphas"` values for `RidgeCV`/`RidgeClassifierCV` in `_check_param_grid()`
- Add a private function `_wrap_param_grid()` to be called by `_check_param_grid()` for wrapping `param_grid["alphas"]` with an outer list (if needed) for `RidgeCV`/`RidgeClassifierCV` models
- In `_parallel_fit()`, set `best_param["alphas"]` to the single best value instead of the full list

Currently, the only two base models with built-in CV that are used in Decoder objects are `RidgeCV` and `RidgeClassifierCV`, but it should be relatively straightforward to extend this if we add [`LogisticRegressionCV`](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegressionCV.html) and [`LassoCV`](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LassoCV.html) in the future.


nilearn/decoding/decoder.py
```diff
@@ -132,17 +133,69 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
+        param_grids = param_grid
+
+    # update each dict
+    for i_param_grid in range(len(param_grids)):
+        param_grid_ = param_grids[i_param_grid].copy()  # shallow copy
+        try:
+            if not isinstance(param_grid_[param_name][0], Iterable):
+                param_grid_[param_name] = [param_grid_[param_name]]
+        except KeyError:
+            continue
```
bthirion, line 189: I'm not really sure what this means. The expected param_name is not found and we continue ?

bthirion (commented):
Thx ! I just have small suggestions

nilearn/decoding/decoder.py
```diff
@@ -132,17 +133,69 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
             min_c = 0.5
 
-        if not isinstance(
-            estimator,
-            (RidgeCV, RidgeClassifierCV, DummyClassifier, DummyRegressor),
-        ):
+        if isinstance(estimator, (RidgeCV, RidgeClassifierCV)):
+            param_grid["alphas"] = [np.geomspace(1e-4, 100, num=20)]
+        elif not isinstance(estimator, (DummyClassifier, DummyRegressor)):
             param_grid["C"] = np.array([2, 20, 200]) * min_c
```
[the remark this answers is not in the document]
jeromedockes: I think these values were there before this PR

nilearn/decoding/decoder.py
```diff
@@ -132,17 +133,69 @@ def _check_param_grid(estimator, X, y, param_grid=None):
         else:
             min_c = 0.5
 
-        if not isinstance(
-            estimator,
-            (RidgeCV, RidgeClassifierCV, DummyClassifier, DummyRegressor),
-        ):
+        if isinstance(estimator, (RidgeCV, RidgeClassifierCV)):
+            param_grid["alphas"] = [np.geomspace(1e-4, 100, num=20)]
```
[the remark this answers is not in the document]
jeromedockes: or even num=10 should be good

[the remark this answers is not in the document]
jeromedockes: actually I would suggest something like `np.geomspace(1e-3, 1e4, 8)` so the values are round numbers and the grid includes some very strongly regularized models

nilearn/decoding/decoder.py
```diff
@@ -132,17 +133,69 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
+        elif not isinstance(estimator, (DummyClassifier, DummyRegressor)):
             param_grid["C"] = np.array([2, 20, 200]) * min_c
         else:
             param_grid = {}
 
+    else:
```
jeromedockes, line 143: to avoid a big block of code between this and the corresponding `if`, maybe we could have a `_default_param_grid(estimator)` helper function?

nilearn/decoding/decoder.py
jeromedockes: yes, wherever a user has written `"alphas": [1, 2, 3]` this function replaces it with the correct `"alphas": [ [1, 2, 3] ]`. when `"alphas"` is not in the parameter grid (meaning use the estimator's default) there is nothing to do.
this would probably be clearer with a test (using `in`) rather than catching a KeyError.

jeromedockes: ATM this function modifies the provided param grid when it is a list of dicts, but not when it is a single dict, and it also returns the (possibly modified) grid. when it makes sense, we often try to have functions like this one either (i) not modify the input and return a new, updated value (preferred) or (ii) modify the input and return `None`, to make it easier to see if they modify their input or not.

In this case I would suggest building a new list. it will also avoid keeping track of the index and may clarify a bit the function's body, which will look like:
```
new_param_grid = []
for param_grid_item in param_grid:
    if "alphas" in param_grid_item and ...
        param_grid_item = dict(param_grid_item, alphas=[...])
    new_param_grid.append(param_grid_item)
return new_param_grid
```

jeromedockes: - do we want to handle the case where a user writes `alphas=10.0`?
- should we issue a warning when we have to wrap the `alphas` in a list because the user didn't do it, or should not doing it be considered the normal way to use the Decoder?

nilearn/decoding/decoder.py
```diff
@@ -234,6 +287,9 @@ def _parallel_fit(
                     dummy_output = estimator.class_prior_
                 elif isinstance(estimator, DummyRegressor):
                     dummy_output = estimator.constant_
+
+            if isinstance(estimator, (RidgeCV, RidgeClassifierCV)):
+                param["alphas"] = estimator.alpha_
```
jeromedockes, line 292: isn't it a bit surprising to have the name `alphas`, but a single value? esp. since `alphas` corresponds to a list everywhere else. maybe it should be a different name like `best_alpha`, or `alpha`?

nilearn/decoding/decoder.py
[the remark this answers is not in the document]
bthirion: We should probably change that.

nilearn/decoding/decoder.py
bthirion: > do we want to handle the case where a user writes `alphas=10.0`?

No, I don't think we should. (This is open to discussion if people think differently)

nilearn/decoding/decoder.py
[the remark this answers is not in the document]
michellewang: The `[2, 20, 200]` is multiplied by the `min_c` value which is set to 0.5 unless the estimator uses L1 penalty, then `min_c` is computed with [`sklearn.svm.l1_min_c`](https://scikit-learn.org/stable/modules/generated/sklearn.svm.l1_min_c.html). So I think the default is still `[1, 10, 100]`, just obtained in a more convoluted way. I don't know why we would want to multiply `min_c` by `[2, 20, 200]` for estimators with L1 penalty though.

jeromedockes:
@michellewang once you merge `upstream/main` the failures should have been fixed by #3649

michellewang:
@jeromedockes @bthirion I see that the decoder tests have gone through a big refactoring. Let me know if there is a better way to write the tests I added

nilearn/decoding/decoder.py
```diff
@@ -106,41 +107,131 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
+def _default_param_grid(estimator, X, y):
+    """Generate sensible default for param_grid.
+
+    Parameters
+    ----------
+    estimator: str, optional
+        The estimator to choose among:
+        %(classifier_options)s
+        %(regressor_options)s
+        Default 'svc'.
```
bthirion, line 128: There is no default (and there should not be any I guess).

nilearn/decoding/decoder.py
```diff
@@ -106,41 +107,131 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
+    param_grid: dict of str to sequence, or sequence of such. Sensible default
+    dict has size 1 for linear models.
+    """
+    param_grid = {}
+
+    # define loss function
+    if isinstance(estimator, LogisticRegression):
+        loss = "log"
+    elif isinstance(estimator, (LinearSVC, RidgeCV, RidgeClassifierCV, SVR)):
+        loss = "squared_hinge"
```
bthirion, line 151: Why for Ridge ?

nilearn/decoding/decoder.py
```diff
@@ -106,41 +107,131 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
+    else:
+        raise ValueError(
+            "Invalid estimator. The supported estimators are:"
+            f" {list(SUPPORTED_ESTIMATORS.keys())}"
+        )
+    # define sensible default for different types of estimators
+    if hasattr(estimator, "penalty") and (estimator.penalty == "l1"):
+        min_c = l1_min_c(X, y, loss=loss)
+    else:
+        min_c = 0.5
```
bthirion, line 168: What case does this correspond to ? Can you comment on this choice ?

bthirion (commented):
Thx. A bit more pedagogy would help.

nilearn/decoding/decoder.py
michellewang: Good catch, removing the default from the docstring

nilearn/decoding/decoder.py
michellewang: @bthirion this was from the existing code, I just moved it to another function as per @jeromedockes's suggestion. But you're right, it doesn't seem to make much sense: `loss` is only used in the `if` block on Lines 163-164, but `RidgeCV`, `RidgeClassifierCV` and `SVR` all don't have the `penalty` attribute, so it looks like this is useless. I can change it here, or maybe we should have another PR for it? Since this PR is supposed to address wrapping of the `param_grid` for `RidgeCV`/`RidgeClassifierCV`

nilearn/decoding/decoder.py
michellewang: This was also part of the existing code, I don't know what the reasoning was. Should we open a new issue to discuss this?

nilearn/decoding/tests/test_decoder.py
```diff
@@ -187,6 +188,58 @@ def test_check_parameter_grid_is_empty(rand_X_Y):
...
+        ({"alphas": [(1, 10, 100, 1000)]}, False),
+        ({"alphas": ((1, 10, 100, 1000),)}, False),
+        ({"alphas": np.array([1, 10, 100, 1000])}, True),
+        ({"alphas": [np.array([1, 10, 100, 1000])]}, False),
+        ([{"alphas": [1, 10]}, {"alphas": [[100, 1000]]}], True),
+        ([{"alphas": [1, 10]}, {"fit_intercept": [True, False]}], True),
+        (None, False),
+    ],
+)
+def test_wrap_param_grid(param_grid, need_wrap):
```
Remi-Gau, line 208: I think this tests too many things.

I would split this into:
- one test that checks that the warning is thrown (not sure it is worth checking the warning is not thrown)
- one test for  `param_grid_wrapped = _wrap_param_grid(None, param_name)`
- one test for the "is param_name in param_grid_wrapped?"

This should remove of lot of the "logic" you need to go through when putting everything in a single test

nilearn/decoding/tests/test_decoder.py
```diff
@@ -308,6 +361,66 @@ def test_parallel_fit(rand_X_Y):
...
+    )
+
+    # create appropriate scorer and update y for classification
+    if is_classification:
+        scorer = check_scoring(estimator, "accuracy")
+        _, y = rand_X_Y
+    else:
+        scorer = check_scoring(estimator, "r2")
+
+    for params in [param_values, [param_values]]:  # test unwrapped and wrapped
```
Remi-Gau, line 405: this too can probably be parametrized

nilearn/decoding/tests/test_decoder.py
```diff
@@ -308,6 +361,66 @@ def test_parallel_fit(rand_X_Y):
...
+@pytest.mark.parametrize(
+    "estimator, param_name, fitted_param_name, is_classification",
+    [
+        (RidgeCV(), "alphas", "best_alpha", False),
+        (RidgeClassifierCV(), "alphas", "best_alpha", True),
+    ],
+)
+def test_parallel_fit_builtin_cv(
+    rand_X_Y, estimator, param_name, fitted_param_name, is_classification
+):
```
Remi-Gau, line 373:
```suggestion
@pytest.mark.parametrize("param_values", 
    ([0.001, 0.01, 0.1, 1, 10, 100, 1000],
     [[0.001, 0.01, 0.1, 1, 10, 100, 1000]]))
@pytest.mark.parametrize(
    "estimator, param_name, fitted_param_name, is_classification",
    [
        (RidgeCV(), "alphas", "best_alpha", False),
        (RidgeClassifierCV(), "alphas", "best_alpha", True),
    ],
)
def test_parallel_fit_builtin_cv(
    rand_X_Y, estimator, param_name, fitted_param_name, is_classification,
    param_values
):
```

nilearn/decoding/tests/test_decoder.py
```diff
@@ -308,6 +361,66 @@ def test_parallel_fit(rand_X_Y):
...
+    """Check that the `fitted_param_name` output of _parallel_fit is a single
+    value even if param_grid is wrapped in a list for models with built-in CV.
+    """
+    param_values = [0.001, 0.01, 0.1, 1, 10, 100, 1000]
+
+    # y will be replaced if this is a classification
```
Remi-Gau, line 379:
```suggestion
    # y will be replaced if this is a classification
```

nilearn/decoding/tests/test_decoder.py
```diff
@@ -308,6 +361,66 @@ def test_parallel_fit(rand_X_Y):
...
+            test=test,
+            param_grid=param_grid,
+            is_classification=is_classification,
+            scorer=scorer,
+            mask_img=None,
+            class_index=1,
+            selector=selector,
+            clustering_percentile=100,
+        )
+        assert best_param[fitted_param_name] in param_values
```
Remi-Gau, line 421:
```suggestion
      param_grid = {param_name: param_values}
      _, _, _, best_param, _, _ = _parallel_fit(
          estimator=estimator,
          X=X,
          y=y,
          train=train,
          test=test,
          param_grid=param_grid,
          is_classification=is_classification,
          scorer=scorer,
          mask_img=None,
          class_index=1,
          selector=selector,
          clustering_percentile=100,
      )
      assert best_param[fitted_param_name] in param_values
```

nilearn/decoding/tests/test_decoder.py
```diff
@@ -308,6 +361,66 @@ def test_parallel_fit(rand_X_Y):
             assert a == b
 
 
+@pytest.mark.parametrize(
+    "estimator, param_name, fitted_param_name, is_classification",
+    [
+        (RidgeCV(), "alphas", "best_alpha", False),
+        (RidgeClassifierCV(), "alphas", "best_alpha", True),
```
Remi-Gau, line 368: I may be missing something, but what was the reason for adding `param_name`, `fitted_param_name` to the parametrization given that they are the same for both cases?

nilearn/decoding/tests/test_decoder.py
Remi-Gau: not sure the indentation will be quite correct on some of the code I changed, so double checks things locally if you accept the suggestions

nilearn/decoding/tests/test_decoder.py
Remi-Gau: gotcha: 👍🏾 
thanks!

nilearn/decoding/tests/test_decoder.py
```diff
@@ -187,6 +188,67 @@ def test_check_parameter_grid_is_empty(rand_X_Y):
...
+    expected_warning_substring = "should be a sequence of iterables"
+
+    with warnings.catch_warnings(record=True) as raised_warnings:
+        _wrap_param_grid(param_grid, param_name)
+    warning_messages = [str(warning.message) for warning in raised_warnings]
+
+    found_warning = any(
+        expected_warning_substring in warning_message
+        for warning_message in warning_messages
+    )
```
Remi-Gau, line 240:
```suggestion
    found_warning = any(
        expected_warning_substring in x
        for x in warning_messages
    )
```

nilearn/decoding/decoder.py
jeromedockes: as you prefer! as you're improving the handling of the parameter grid and you're touching that code anyway, I think you can do this cleanup in the same PR

nilearn/decoding/decoder.py
```diff
@@ -106,41 +106,130 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
+    -------
+    param_grid_wrapped: dict of str to sequence, or sequence of such
+        The updated parameter grid
+    """
+    if param_grid is None:
+        return param_grid
+
+    # param_grid can be either a dict or a sequence of dicts
+    # we make sure that it is a sequence we can loop over
+    input_is_dict = isinstance(param_grid, dict)
```
jeromedockes, line 204: perhaps `collections.abc.Mapping` rather than `dict`?

nilearn/decoding/decoder.py
```diff
@@ -106,41 +106,130 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
+    if input_is_dict:
+        param_grids = [param_grid]
```
jeromedockes, line 206: I find it helps readability to avoid names that differ by just one letter, like `param_grid` and `param_grids`. here as `param_grid` itself can be a sequence in some cases, you could write `param_grid = [param_grid]`, or if you would rather have a different name it could be for example `all_param_grids` (in which case you might want to make `new_param_grid` plural, too)

nilearn/decoding/decoder.py
michellewang: I am not sure this is a good idea because `param_grid` is eventually (in `_parallel_fit()`) passed to `sklearn.model_selection.ParameterGrid`, which explicitly checks that each grid is a `dict` (see [here](https://github.com/scikit-learn/scikit-learn/blob/364c77e04/sklearn/model_selection/_search.py#L106)). If I pass a `Mapping` object that is not a `dict` I get a `TypeError` when instantiating `ParameterGrid`:

```python
from collections.abc import Mapping
from sklearn.model_selection import ParameterGrid

class NonDictMapping(Mapping):
    def __init__(self, data=None) -> None:
        if data is None:
            data = {}
        self.data = data

    def __getitem__(self, __key):
        return self.data[__key]
    
    def __iter__(self):
        return self.data.__iter__()
    
    def __len__(self) -> int:
        return self.data.__len__()

param_grid = NonDictMapping()
print(ParameterGrid(param_grid))
``` 

I guess we could accept `Mapping` objects and internally cast them to `dict`s, but not sure if it's worth it. What do you think @jeromedockes?

jeromedockes: good point -- sorry I missed that! in this case we can keep the check as it is

nilearn/decoding/tests/test_decoder.py
```diff
@@ -187,6 +188,66 @@ def test_check_parameter_grid_is_empty(rand_X_Y):
...
+        {"alphas": [np.array([1, 10, 100, 1000])]},
+        [{"alphas": [1, 10]}, {"alphas": [[100, 1000]]}],
+        [{"alphas": [1, 10]}, {"fit_intercept": [True, False]}],
+    ],
+)
+def test_wrap_param_grid(param_grid):
+    param_name = "alphas"
+    for param in ParameterGrid(_wrap_param_grid(param_grid, param_name)):
+        try:
+            for param_value in param[param_name]:
```
jeromedockes, line 211: here the goal is to check that `param["alpha"]` is a list of numbers?

jeromedockes (commented):
I noticed none of the examples in the gallery actually pass a `param_grid` to the decoder, and there is only ever 1 parameter used (`C`) in the "grid". maybe [this one](https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_grid_search.html#sphx-glr-auto-examples-02-decoding-plot-haxby-grid-search-py) can be extended to set at least 2 parameters (maybe alpha + l1_ratio or something like that)

nilearn/decoding/tests/test_decoder.py
michellewang: Yes, but another way (if we want to avoid hardcoding the values) would be to use `isinstance(param["alpha"], (float, int))`. Do you think that would be better?

michellewang:
> I noticed none of the examples in the gallery actually pass a param_grid to the decoder, and there is only ever 1 parameter used (C) in the "grid". maybe [this one](https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_grid_search.html#sphx-glr-auto-examples-02-decoding-plot-haxby-grid-search-py) can be extended to set at least 2 parameters (maybe alpha + l1_ratio or something like that)

@jeromedockes sounds good, I will work on extending the example next week. Thanks for all of your feedback!

jeromedockes (approved):
apart from 2 small details LGTM, thanks a lot @michellewang !

examples/02_decoding/plot_haxby_grid_search.py
```diff
@@ -74,6 +74,14 @@
 # on nested cross-validation.
 from nilearn.decoding import Decoder
 
+# We provide a grid of hyperparameter values to the Decoder's internal
+# cross-validation. If no param_grid is provided, the Decoder will use a
+# default grid with sensible values for the chosen estimator
+param_grid = {
+    "C": [1, 10, 100, 1000],
+    "intercept_scaling": [1, 10],
```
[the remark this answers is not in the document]
michellewang: @jeromedockes I agree that it is a more interesting hyperparameter, but "penalty" is supposed to be fixed to L2 for the "svc" estimator (see [here](https://github.com/nilearn/nilearn/blob/983accf6/nilearn/decoding/decoder.py#L54)). I think for L1 penalty the user is expected to use the "svc_l1" estimator instead

nilearn/decoding/decoder.py
```diff
@@ -106,41 +106,133 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
+    if param_grid is None:
+        return param_grid
+
+    # param_grid can be either a dict or a sequence of dicts
+    # we make sure that it is a sequence we can loop over
+    input_is_dict = isinstance(param_grid, dict)
+    if input_is_dict:
+        param_grid = [param_grid]
+    else:
+        param_grid = param_grid
```
ymzayek, line 211:
```suggestion
```

I don't think these 2 lines are needed

ymzayek (approved):
Thx @michellewang ! just a small suggestion below. I'm also wondering if the addition to the example is the most minimal we can do to achieve reasonable results? Just to note that it adds about 40s to run. I think thi is acceptable but just want to make sure we keep the examples as minimal as possible

examples/02_decoding/plot_haxby_grid_search.py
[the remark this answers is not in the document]
jeromedockes: just because it is initialized with that value doesn't mean it's fixed, `set_params` will change it -- note `penalty="l2"` is the default so it is repeated in the line you linked for clarity but if it wasn't there it wouldn't make a difference. otherwise if not the penalty maybe we can add the loss to the grid?

michellewang:
> just a small suggestion below. I'm also wondering if the addition to the example is the most minimal we can do to achieve reasonable results? Just to note that it adds about 40s to run. I think thi is acceptable but just want to make sure we keep the examples as minimal as possible

@ymzayek I updated the sample so that it would use L1 and L2 penalties in `param_grid`, with only 2 values for `C`. Hopefully this reduces the build time a bit. Otherwise I could make it so that only the first Decoder fitting (before the nested CV) uses a custom `param_grid`.

examples/02_decoding/plot_haxby_grid_search.py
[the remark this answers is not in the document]
michellewang: Ok, I changed it to this:
```python
param_grid = [
    {
        "penalty": ["l2"],
        "dual": [True],
        "C": [100, 1000],
    },
    {
        "penalty": ["l1"],
        "dual": [False],
        "C": [100, 1000],
    },
]
```

Note that if `param_grid` is a list of dicts and the dicts have different keys, the fitted Decoder's `cv_params_` output can contain lists that have less elements than the number of CV folds. For example, if `param_grid` is
```python
param_grid = [
    {
        "penalty": ["l2"],
        "C": [100, 1000],
    },
    {
        "penalty": ["l1"],
        "dual": [False],    # "dual" is not in the first dict
        "C": [100, 1000],
    },
]
```

Then it is possible that `decoder.cv_params_["shoe"]` has something like
```python
{
    'C': [1000, 1000, 100, 100, 1000], 
    'dual': [False, False, False],    # less than 5, not aligned with the other hyperparameter values
    'penalty': ['l1', 'l2', 'l1', 'l1', 'l2'],
}
```

I think this should be considered a bug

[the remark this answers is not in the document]
jeromedockes: I think so too!

bthirion (approved):
LGTM, thx.

ymzayek:
Merging, thx @michellewang and all the reviewers!

PR #3631, opened 2023-03-29
add a warning when imgs being upsampled to the mask_img resolution

### LINKED ISSUE #3397: User warning when data is being upsampled when masked
The default beahaviour of maskers is to resample the data to the mask's affine, but when the data has a lower resolution than the mask, it upsamples the data, requiring more computation and memory than necessary. Throwing a warning when it happens could help users to specify a target affine when it is relevant. 

### Benefits to the change
Users can realise when they upsample their data, and that they can save computation time and memory by specifying a target affine (or a resampling strategy).

```python
## in transform_single_imgs
if imgs.resolution < self.mask_img_.resolution:
    raise UserWarning("Images are being upsampled to the mask's resolution, "
        "you might want to provide a target_affine to save memory and computation time.")

```

mtorabi59:
Closes #3397 .


nilearn/maskers/nifti_masker.py
```diff
@@ -100,6 +100,8 @@ def _filter_and_mask(
     # as small as possible in order to speed up the process
 
     if not _utils.niimg_conversions._check_same_fov(imgs, mask_img_):
+        warnings.warn('imgs are being upsampled to the mask_img resolution, '
+            'you might want to provide a target_affine to save memory and computation time.')
```
ymzayek, line 104:
```suggestion
        warnings.warn(
            'imgs are being upsampled to the mask_img resolution, '
            'you might want to provide a target_affine to save memory and computation time.',
            UserWarning
        )
```

We can also probably improve this warning a bit. Maybe adding that this is memory intensive when using a high resolution mask.

ymzayek (commented):
LGTM! We'll also want to add a test that the warning is raised in `nilearn/maskers/tests/test_nifti_masker.py` and then finally add a whatsnew entry and your name in `doc/changes/names.rst`

nilearn/maskers/nifti_masker.py
mtorabi59: @ymzayek  How about "imgs are being upsampled to the mask_img resolution. This process is memory intensive when using a high resolution mask. You might want to provide a target_affine to save memory and computation time." ?

bthirion (approved):
LGTM. Waiting for the CI to complete.

ymzayek:
Ok I think I missed something. The warning will also show if downsampling is happening. So we are checking if the resolution is just different with `_check_same_fov` and then warning that the data is being "upsampled". However, this warning will also show if the data is being downsampled given a mask resolution that is different but lower than the data's. So I think we either need an additional check or a more general warning (which maybe defeats the purpose)

mtorabi59:
@ymzayek you're right! I will try to add additional conditions to make it specific to upsampling.

nilearn/maskers/nifti_masker.py
```diff
@@ -95,6 +95,14 @@ def _filter_and_mask(
     # as small as possible in order to speed up the process
 
     if not _utils.niimg_conversions._check_same_fov(imgs, mask_img_):
+        warnings.warn(
+            'imgs are being resampled to the mask_img resolution. '
+            'This process is memory intensive when using a '
+            'high resolution mask. You might want to provide '
+            'a target_affine or resample the mask beforehand '
+            'to save memory and computation time.',
```
ymzayek, line 103:
```suggestion
            'imgs are being resampled to the mask_img resolution. '
            'This process is memory intensive. You might want to provide '
            'a target_affine that is equal to the affine of the imgs '
            'or resample the mask beforehand '
            'to save memory and computation time.',
```

I think the point of this message is to avoid resampling, whether up or down, which is why I removed the "high resolution mask" part. Indeed using high resolution data in an analysis pipeline is memory intensive in general but it's not the point of this warning. I added that the target affine should be equal to the data affine so that the mask is resampled to this. So that would produce the same behavior as resampling the mask to the data beforehand.

See relevant part of code where the mask would be resampled which then avoids the behavior of resampling the data that comes later in the process: https://github.com/nilearn/nilearn/blob/2e18c91342b65048307a21a45b5a8473414b6d16/nilearn/maskers/nifti_masker.py#L471-L477

nilearn/maskers/tests/test_nifti_masker.py
```diff
@@ -69,6 +69,30 @@ def test_resample():
     assert np.any(X != 0)
 
 
+def test_upsample_warning():
+    """Check that a warning is raised when data is
+    being upsampled to mask's resolution.
```
ymzayek, line 74:
```suggestion
    being resampled to mask's resolution.
```

nilearn/maskers/tests/test_nifti_masker.py
```diff
@@ -69,6 +69,30 @@ def test_resample():
     assert np.any(X != 0)
 
 
+def test_upsample_warning():
```
ymzayek, line 72: Maybe better name is `test_resample_to_mask_warning`

ymzayek (commented):
@mtorabi59 looks good thx! Just need to update the test as suggested and add a whatsnew entry in `doc/changes/latest.rst`

ymzayek:
And also your name to `doc/changes/names.rst`

nilearn/maskers/tests/test_nifti_masker.py
```diff
@@ -69,6 +69,30 @@ def test_resample():
...
 
 
+def test_resample_to_mask_warning():
+    """Check that a warning is raised when data is
+    being resampled to mask's resolution.
+    """
+    data = np.zeros((9, 9, 9))
+    data[3:-3, 3:-3, 3:-3] = 10
+    img = nibabel.Nifti1Image(data, np.eye(4))
+    # defining a mask with a higher resolution than img
```
ymzayek, line 79:
```suggestion
    # defining a mask with different fov than img
```

More accurate

ymzayek (commented):
Thx @mtorabi59! A couple more small details

doc/changes/latest.rst
```diff
@@ -37,6 +37,8 @@ Enhancements
 
 - :func:`~glm.first_level.first_level_from_bids` now takes an optional ``sub_labels`` argument and warns users of given subject labels that are not present in the dataset (:gh:`3351` by `Kevin Sitek`_).
 
+- :class:`NiftiMasker` now warns users if the data is being resampled to mask resolution (:gh:`3631` by `Mohammad Torabi`_).
```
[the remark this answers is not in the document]
mtorabi59: Thanks @ymzayek ! I wasn't sure what's the correct format of referencing things. Is there any instructions about it?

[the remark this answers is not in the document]
ymzayek: With sphinx it's not always so clear so I wouldn't worry too much about it. I just know for linking within nilearn you need to first use the module name. The tilda is so that when it's rendered only the class name is printed. There are some guidelines also here: https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#

ymzayek (approved):
I think this one is good to go. Let's see if all the tests pass

mtorabi59:
@ymzayek  Thank you for walking me through the process!

Issue #3612, opened 2023-03-23
Adding an option to easily build specific examples in a pull request


ymzayek:
I would like to make it easier to build examples in a particular example directory or just one example and have that be built by the documentation builder CI in a pull request and hence hosted by CircleCI for easy access. This can probably be done by adding a commit tag option (like we do for full builds). 

At the moment it is only possible to do a partial build without building any examples (the default), running only changed examples if the example scripts are changed, or doing a full build with the commit tag `[full doc]`. However, if code is changed in nilearn modules that could affect an example or a set of examples, these won't be built by the CI. This will be useful to be able to point reviewers to possible visual changes in e.g. plots in an example.

As an example, something like this triggers building all examples in a particular example directory such as `01_plotting`:

```bash
sphinx-build -D sphinx_gallery_conf.examples_dirs="../examples/01_plotting/" -D sphinx_gallery_conf.run_stale_examples=True -b html -d _build/doctrees -v . _build/html
```

Or to build a specific example or several following a pattern:

```bash
sphinx-build -D sphinx_gallery_conf.filename_pattern="plot_decoding_tutorial.py" -D sphinx_gallery_conf.run_stale_examples=True -b html -d _build/doctrees . _build/html
```

Sphinx gallery configs reference: https://sphinx-gallery.github.io/stable/configuration.html

bthirion:
This perfectly makes sense.

PR #3637, opened 2023-03-31
[FIX] Fix `second_level_input` check function

### LINKED ISSUE #3636: Documentation build failure on main
Some examples failed on the doc build after last merge on main
https://github.com/nilearn/nilearn/actions/runs/4573237098/jobs/8073423194

Looking into it...

ymzayek:
Closes #3636 .

@Remi-Gau is the loop under `_check_input_as_nifti_images` supposed to go in the if statement?

Might also warrant improving testing for new private second level model functions and also add docstrings

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -12,6 +12,7 @@ NEW
 
 - ``fsaverage`` meshes accessed through :func:`~datasets.fetch_surf_fsaverage` now come with flat maps for all resolutions (:gh:`3444` by `Alexis Thual`_).
 
+
 Fixes
 -----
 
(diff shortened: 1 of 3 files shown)


ymzayek:
Or is the input supposed to be 3D?

ymzayek:
Nope and nope, I think you meant to use `atleast_4D`

ymzayek:
@Remi-Gau I will merge this to green the CI and then I agree to do a follow-up PR to add a test.

These things are not caught because of the partial doc builds in PRs. If we called a full build in the PR it would've failed. Most of the time it's not necessary to do a full build unless you change a lot of documentation or examples. So it happens and not a big deal and calling full builds all the time takes up a lot of resources and time. This is why however I would like to do this: #3612. I will get on it asap.

Issue #3655, opened 2023-04-05
[MAINT] refactor complex and long functions


Remi-Gau:
Some functions / methods in the code base seem to be unusually long and / or complex.

That would make them good targets for refactoring.

### Benefits to the change

Ideally this should help with the readability and / or maintenance of those functions in the long run.

This would also allow to set a threshold in terms of code quality (functions should not be longer than X or more complex than Y).

> [!TIP]
> # **💻 🍁  For the [hacktoberfest](https://hacktoberfest.com/) contributors 🍁 💻**
>
> Make sure to:
> - read our [contributing guidelines](https://nilearn.github.io/stable/development.html#contribution-guidelines)
>
> If this is your first time contributing to Nilearn, add your details in the [CITATION.cff file](https://github.com/nilearn/nilearn/blob/dcd92aa357ab3a751161aeea2edfa3a775ae644f/CITATION.cff#L42).
>
> For your pull request to be valid, you should refactor at least one of long / complex functions mentioned below.
>

## McCabe complexity 

```bash
flake8 --select C901 --max-complexity 20 nilearn
```

- nilearn/connectome/group_sparse_cov.py:232:1: C901 '_group_sparse_covariance' is too complex (21)
- nilearn/datasets/func.py:582:1: C901 'fetch_localizer_contrasts' is too complex (21)

## In progress

- #3666 
  - [ ] nilearn/mass_univariate/permuted_least_squares.py:301:1: CFQ001 Function permuted_ols has length 405 that exceeds max allowed length 150
  - [ ] nilearn/mass_univariate/permuted_least_squares.py:301:1: C901 'permuted_ols' is too complex (41)
- #3701 
  - [ ] nilearn/image/resampling.py:349:1: C901 'resample_img' is too complex (31)
  - [ ] nilearn/image/resampling.py:349:1: CFQ001 Function resample_img has length 252 that exceeds max allowed length 150
- #4635 
  - [x] nilearn/plotting/displays/_projectors.py:60:5: CFQ001 Function add_graph has length 154 that exceeds max allowed length 150  
  - [x] nilearn/plotting/displays/_projectors.py:60:5: C901 'OrthoProjector.add_graph' is too complex (21)

Remi-Gau:
To get an idea of the distribution of the complexity of functions in the code base:

the following gives an idea of how many functions have a given complexity: feel like values in the 15-20 range start becoming outliers.

```bash
[...]

PR #3407, opened 2022-11-06
[MAINT] Apply refurb suggestions

DimitriPapadopoulos:
* [FURB108]: Use `x in (y, z)` instead of `x == y or x == z`
* [FURB124]: Use `x == y == z` instead of `x == y and x == z`
* [FURB110]: Use `x or y` instead of `x if x else y`
* [FURB105]: Use `print()` instead of `print("")`
* [FURB125]: Return is redundant here
* [FURB123]: Replace `float(x)` with `x`
* [FURB111]: Use `f` instead of `lambda x: f(x)`
* [FURB121]: Replace `isinstance(x, y) or isinstance(x, z)` with `isinstance(x, (y, z))`

PR #3666, opened 2023-04-07
[MAINT] refactor permuted_ols

Remi-Gau:
- Relates to #3655 

- refactor nilearn.mass_univariate.permuted_least_squares.permuted_ols
- add an extra dictionary for codespell

### TODO

- [x] add doc strings to extracted functions


nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -611,21 +614,148 @@ def permuted_ols(
...
+
+    masker : None or :class:`~nilearn.maskers.NiftiMasker` or \
+            :class:`~nilearn.maskers.MultiNiftiMasker`
+        A mask to be used on the data.
+        This is required for cluster-level inference, so it must be provided
+        if ``threshold`` is not None.
+
+    """
+    if target_vars.ndim != 2:
+        raise ValueError(
```
bthirion, line 747: If the coverage values are not handled in this PR, they should be handled later.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -663,185 +815,220 @@ def permuted_ols(
...
-            threshold_t = stats.t.isf(threshold / 2, df=dof)
-        else:
-            threshold_t = stats.t.isf(threshold, df=dof)
-    else:
-        threshold_t = None
 
-    # Permutations
-    # parallel computing units perform a reduced number of permutations each
+def _get_tfce_original_data(
+    tfce, masker, scores_original_data, two_sided_test
```
bthirion, line 988: Maybe add a one-liner here.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -663,185 +815,220 @@ def permuted_ols(
...
+    masker,
+    tfce,
+    threshold,
+    output_type,
+    n_descriptors,
+    intercept_test,
+    scores_original_data,
+    tfce_original_data,
+):
     if n_perm > n_jobs:
```
bthirion, line 1029: + here

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -912,110 +1110,182 @@ def permuted_ols(
...
+    labeled_arr3d = labeled_arr3d + temp_labeled_arr3d
+    del temp_labeled_arr3d
 
-    return outputs
+    return labeled_arr3d
+
+
+def _compute_cluster_masses(
+    scores_original_data_3d, cluster_labels, labeled_arr3d, threshold_t
+):
```
bthirion, line 1284: Add a short explanation

bthirion (commented):
Thx ! Looks good so far.

nilearn/mass_univariate/permuted_least_squares.py
Remi-Gau: Refactoring without enough coverage is to software development is what free soloing is to climbing.
Will improve coverage.
Ideally and where applicable I will create the test on main and then make sure that they still pass after refactoring to make sure nothing broke.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -636,7 +766,29 @@ def permuted_ols(
             'If "threshold" is not None, masker must be defined as well.'
         )
 
-    # Resolve the output_type as well
+
+def _check_output_type_permuted_ols(output_type, tfce, threshold):
+    """Check that the output type is valid \
+    for the given threhsold argument and depending if we are doing TFCE.
```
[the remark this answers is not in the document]
Remi-Gau: Why wasn't this caught by codespell? 
Will try to see if I can automate this: I don't want you to spend time on my pooor tyipng siklls.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -663,185 +815,220 @@ def permuted_ols(
...
-            model_intercept = True
-
-            # remove confounding vars variable if it is empty
-            if confounding_vars.size == 0:
-                confounding_vars = None
 
-        # intercept is only defined in confounding vars
-        if not intercept_test and len(constants) == 1:
-            intercept_test = True
+def _optionally_add_intercept(
```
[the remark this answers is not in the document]
Remi-Gau: YES!!! 
I had just adapted from the original comment but this make a lot more sense.

setup.cfg
```diff
@@ -77,11 +77,11 @@ exclude =
...
     dist,
 	nilearn/externals/tempita
---select = D,E,F,W,C90
+--select = D,E,F,W,C
 docstring-convention = numpy
 max-line-length = 79
-max_complexity = 43
-max_function_length = 150
+max_complexity = 34
+max_function_length = 255
```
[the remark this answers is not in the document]
Remi-Gau: Note that max_function_length is not (yet?) checked in CI, this value would only trigger a flaek8 error if you ran flake8 locally with flake8-functions installed.

I updated this value so that once this PR is done, there should be no function that are longer than 255 lines (see the list here https://github.com/nilearn/nilearn/issues/3655).

This way if we want to implement function metric checks in CI, this new value will be our upper limit that gives us a green CI.

Remi-Gau:
thanks for the early reviews @NicolasGensollen @bthirion
the good news is that this is refactoring the most complex function of the codebase: so hopefully further refactoring of other part of the code base should be a breeze in comparison (famous last words).

setup.cfg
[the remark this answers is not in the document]
NicolasGensollen: Thanks for the explanations @Remi-Gau ! 🙏 
Great to see these massive functions getting chunked ! 👍

Remi-Gau:
Unsure if this one would actually need an entry in the changelog: there should be no behavioral change, but at the same time this is not a simple reformatting and not "just" a refactoring of parts of the test suite.
So to me this falls in the grey zone of "should be in the changelog": I lean slightly more towards the "add an entry in the changelog" but definitely not a hill I will die on.

nilearn/mass_univariate/permuted_least_squares.py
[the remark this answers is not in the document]
Remi-Gau: done

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -611,21 +614,148 @@ def permuted_ols(
...
+    assert target_vars.shape[0] == tested_vars.shape[0]
+    if confounding_vars is not None:
+        assert target_vars.shape[0] == confounding_vars.shape[0]
+
+    # check explanatory variates' dimensions
+    if tested_vars.ndim == 1:
+        tested_vars = np.atleast_2d(tested_vars).T
 
     # check n_jobs (number of CPUs)
+    if n_jobs < 0:
```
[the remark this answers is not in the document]
Remi-Gau: I tried to have a single call to an input validation function in main function.

Tried to then further split the validation into smaller more focused functions.

I am not happy with how redundant some of the parameters definitions are across doc strings: I find myself doing a lot of copy-pasting-modify which is usually bad sign.

nilearn/mass_univariate/permuted_least_squares.py
Remi-Gau: added a test

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -359,14 +359,46 @@ def test_permuted_ols_no_covar_warning(random_state=RANDOM_STATE):
     assert np.array_equal(output_1["t"][1:], output_2["t"][1:])
 
 
+def test_permuted_ols_tested_var_transpose(design, random_state=RANDOM_STATE):
```
Remi-Gau, line 362: remove

nilearn/mass_univariate/tests/test_permuted_least_squares.py
```diff
@@ -359,14 +359,46 @@ def test_permuted_ols_no_covar_warning(random_state=RANDOM_STATE):
...
+        model_intercept=False,
+        n_perm=0,
+        random_state=random_state,
+        output_type="dict",
+        n_jobs=-1,
+    )
+    compare_to_ref_score(output["t"], tested_var, target_var)
+
+
+def test_permuted_ols_no_covar_negative_n_job(
```
Remi-Gau, line 376: add doc string

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -611,18 +614,290 @@ def permuted_ols(
...
+    """Run permutations and compute p-values.
+
+    If permutation test was requested, this function will run
+    the n_perm permutations and compute the p-values of
+    the voxel-wise scores_original_data
+    and the p-values of the cluster-wise tfce_original_data
+    and return the proper output_type.
+
+    See permuted_ols for a description of the parameters.
+    """
```
Remi-Gau, line 734: added some big overview of what the function does.

Note: if anyone has good tips on how to minimize copy-pasting parameters definition a bazillion times across docstrings within the same module, let me know.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -663,359 +968,464 @@ def permuted_ols(
...
+def _get_tfce_original_data(
+    tfce, masker, scores_original_data, two_sided_test
+):
+    """Calculate threshold-free cluster enhancement values.
 
-    vfwe_pvals = (n_perm + 1 - vfwe_scores_as_ranks) / float(1 + n_perm)
+    This value will be returned as is by permuted_ols
+    if there is no permutation test.
+    Otherwise it will be used as reference after permutations to compute
+    the p-values.
```
Remi-Gau, line 1237: hope this makes sense

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -663,359 +968,464 @@ def permuted_ols(
...
-            "h0_max_t": vfwe_h0,
-        }
+def _label_clusters(scores_original_data_3d, threshold_t, two_sided_test):
+    """Give a label to each voxel depending on the cluster it belongs to.
 
-        if tfce:
-            outputs["tfce"] = tfce_original_data.T
-            outputs["logp_max_tfce"] = neg_log10_tfce_pvals
-            outputs["h0_max_tfce"] = h0_tfcemax
+    This is a preparatory step to estimate cluster mass and size.
```
Remi-Gau, line 1398: not sure what else to add

Remi-Gau (commented):
@bthirion added a couple of doc strings to try to explain things better

let me know what you think

jeromedockes:
> Unsure if this one would actually need an entry in the changelog

I would say yes -- it doesn't hurt and it might be very helpful if the refactoring introduces a bug

nilearn/mass_univariate/permuted_least_squares.py
bthirion: ...Just one line...

bthirion (commented):
Almost there, thx.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -663,359 +968,464 @@ def permuted_ols(
...
+def _check_for_intercept_in_tested_var(tested_vars):
+    """Check if explanatory variates contain an intercept (constant) or not.
+
+    Parameters
+    ----------
+    tested_vars : array-like, shape=(n_samples, n_regressors)
+        Explanatory variates, fitted and tested independently from each others.
+    """
+    n_regressors = tested_vars.shape[1]
+    return n_regressors == np.unique(tested_vars).size == 1
```
jeromedockes, line 1034: this checks that there is only one regressor -- so the tested_vars contain _only_ an intercept?

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -663,359 +968,464 @@ def permuted_ols(
...
+        column
+        for column in range(confounding_vars.shape[1])
+        if np.unique(confounding_vars[:, column]).size == 1
+    ]
+    # check if multiple intercepts are defined across all variates
+    if (tested_var_has_intercept and len(constants) == 1) or len(
+        constants
+    ) > 1:
+        # remove all constant columns
+        confounding_vars = np.delete(confounding_vars, constants, axis=1)
```
jeromedockes, line 1073: doesn't this delete all constant columns (instead of leaving one) in the case where `tested_var_has_intercept is False` but `len(constants) > 1`? also what about the case where there are several regressors but one (or several) is constant?

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -611,18 +614,290 @@ def permuted_ols(
...
+        Cluster-forming threshold in p-scale.
+
+    masker : None or :class:`~nilearn.maskers.NiftiMasker` or \
+            :class:`~nilearn.maskers.MultiNiftiMasker`
+        A mask to be used on the data.
+        This is required for cluster-level inference, so it must be provided
+        if ``threshold`` is not None.
+
+    """
     if n_jobs == 0:  # invalid according to joblib's conventions
```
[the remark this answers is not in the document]
Remi-Gau: will double check and remove if we can

note to self: check if similar checks do not exist in the rest of the codebase

nilearn/mass_univariate/permuted_least_squares.py
jeromedockes: > took me a while to figure out the logic of this whole thing

a sign that this refactoring PR was needed!

>  I can give it a try if you think it is worth it

I think so.

and the behaviour around the intercept can maybe be sanitized a bit: ATM if the tested variates contain several columns, including a constant one, an intercept is never added to the confounds (even when testing the non-constant variates). if we think that such input makes no sense it may be worth a warning or error, or at least the variable names should reflect what is the condition: "is_intercept" rather than "has_intercept"

also, the documentation mentions a `pvals` key in the output which doesn't exist (there may be more)

although this function allows testing several variables, all the examples seem to use only one -- it might be worth seeing if testing several could be illustrated too

nilearn/mass_univariate/permuted_least_squares.py
jeromedockes: and that's really a subjective detail but I see this line and a few others were rewritten based on a linter's suggestions in #3407 and used to be `(a == 1) and (b == 1)`. comparison oparators can be chained in python, but that's not used super frequently and in some other languages it is parsed as 2 binary comparison expressions (so `2 == 2 == 2` is `false`), or is a syntax error -- for example C and C++, Javascript, Octave (and thus I guess Matlab), Java, R, and Rust. so the old version is the "least surprising" one for people relatively new to Python IMO

nilearn/mass_univariate/permuted_least_squares.py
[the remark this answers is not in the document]
Remi-Gau: so it turns out this check is also there to prevent an eventual dividing by zero error that happens when computing the how to chunk the permutations.

```python
    if n_perm > n_jobs:
        n_perm_chunks = np.asarray([n_perm / n_jobs] * n_jobs, dtype=int)
        n_perm_chunks[-1] += n_perm % n_jobs
```

.flake8
```diff
@@ -9,7 +9,9 @@ exclude =
 	nilearn/externals/tempita
 	nilearn/externals/
 	nilearn_cache
---select = D,E,F,W,C
+	venv,
+	doc/_build
+--select = D,E,F,W,C90
 docstring-convention = numpy
 max-line-length = 79
 max_complexity = 34
```
bthirion, line 17: This is unrelated stuff

Remi-Gau:
while trying to refactor this I think  I may have encountered a bug or at least something I do not understand.

this relates to what @jeromedockes pointed above:
https://github.com/nilearn/nilearn/pull/3666#discussion_r1171874822

This relates to this section of code (unrefactored)
https://github.com/nilearn/nilearn/blob/fd75801ae360705e4adea5bd7a98418f4a3a0c60/nilearn/mass_univariate/permuted_least_squares.py#L689-L733

I tried to see what would be the value of `intercept_test` depending on several possible inputs.

See table below:

`model_intercept` | tested_var_has_intercept | nb_cst_in_confounds | `intercept_test`
-- | -- | -- | --
0 | 0 | 0 | 0
0 | 0 | 1 | 1
0 | 0 | > 1 | 0
0 | 1 | 0 | 1
0 | 1 | 1 | 1
0 | 1 | > 1 | 1
1 | 0 | 0 | 0
1 | 0 | 1 | 1
1 | 0 | > 1 | 0
1 | 1 | 0 | 1
1 | 1 | 1 | 1
1 | 1 | > 1 | 1

what I do not understand is why we would not perform an intercept test if the counfounds have several intercepts (constants) but we do if they have one.

am I missing something?

@jeromedockes @ymzayek @bthirion if you have time

Remi-Gau:
Note to self: make sure errors are tested

ymzayek:
Thanks for the table @Remi-Gau ! Seems like suboptimal logic indeed. I think if `nb_cst_in_confounds` > 1, then it doesn't care about `intercept_test` and removes all constant columns in condition in L704 and sets `model_intercept` to True, and then it adds back a constant column in L727. So in the current logic it actually needs `intercept_test` to be False.

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -913,110 +846,607 @@ def permuted_ols(
...
+    Parameters
+    ----------
+    tested_vars : array-like, shape=(n_samples, n_regressors)
+        Explanatory variates, fitted and tested independently from each others.
+
+    target_vars : array-like, shape=(n_samples, n_descriptors)
+        fMRI data to analyze according to the explanatory and confounding
+        variates.
+
+    confounding_vars : array-like, shape=(n_samples, n_covars), optional
```
bthirion, line 896: Here, we're not consistent in the use of 'optional'

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -913,110 +846,607 @@ def permuted_ols(
...
+
+def _check_output_type_permuted_ols(output_type, tfce, threshold):
+    """Check that the output type is valid \
+    for the given threshold argument and depending if we are doing TFCE.
+
+    output_type may be updated when necessary.
+
+    Parameters
+    ----------
+    output_type : {'legacy', 'dict'}, optional
```
bthirion, line 966: not optional

nilearn/mass_univariate/permuted_least_squares.py
```diff
@@ -913,110 +846,607 @@ def permuted_ols(
...
+    -------
+    tested_var_has_intercept : :obj:`bool`
+    """
+    n_regressors = tested_vars.shape[1]
+    return (n_regressors == np.unique(tested_vars).size) and (
+        np.unique(tested_vars).size == 1
+    )
+
+
+def _find_intercept_columns(confounding_vars):
```
bthirion, line 1076: Add a 1-line docstring

Remi-Gau:
will close this one and restart the refactor from scratch across smaller PRs because resolving the merge conflicts on this one is becoming too much

Issue #3658, opened 2023-04-05
Possible test pollution?


Remi-Gau:
Tried to randomize the order tests are run in to make sure that there is no test pollution in the test suite (tests should be independent), and it seems that it leads to some failures:

See this workflow:
https://github.com/Remi-Gau/nilearn/actions/runs/4618308860/jobs/8165599264

On this PR:
https://github.com/Remi-Gau/nilearn/pull/24/files

Will investigate.

Remi-Gau:
Failing tests I can see so far.

- nilearn/image/tests/test_resampling.py::test_resample_identify_affine_int_translation
- nilearn/image/tests/test_resampling.py::test_resampling_nan[core_shape1]
- nilearn/image/tests/test_resampling.py::test_resampling_nan[core_shape0]
- nilearn/image/tests/test_resampling.py::test_reorder_img
- nilearn/decoding/tests/test_decoder.py::test_check_param_grid

Remi-Gau:
trying https://github.com/asottile/detect-test-pollution

```bash 
detect-test-pollution \
    --failing-test nilearn/image/tests/test_resampling.py::test_reorder_img  --tests nilearn/image/tests/test_resampling.py 
discovering all tests...
-> discovered 80 tests!
ensuring test passes by itself...
-> OK!
ensuring test fails with test group...
-> OK!
running step 1:
- 79 tests remaining (about 7 steps)
running step 2:
- 40 tests remaining (about 6 steps)
running step 3:
- 20 tests remaining (about 5 steps)
running step 4:
- 10 tests remaining (about 4 steps)
running step 5:
- 5 tests remaining (about 3 steps)
running step 6:
- 2 tests remaining (about 1 steps)
double checking we found it...
-> the polluting test is: nilearn/image/tests/test_resampling.py::test_reorder_img_error_reorder_axis
```

Remi-Gau:
OK it seems that some tests were modifying some globals.

Remi-Gau:
Wondering if it would be worth having a workflow that runs the suite "regularly" (once a week?) on main but with test in a random order.

I think that some of those failures were introduced during some of my recent tests refactoring, and having something in CI to make sure that tests are really independent of the each other would be a good thing.

bthirion:
Good catch. 

But I guess that this will happen quite rarely. Maybe this is merely something to do before a release ?

Remi-Gau:
> Maybe this is merely something to do before a release ?

[...]

PR #3668, opened 2023-04-11
[MAINT] add monthly workflow to run tests in a random order

Remi-Gau:
Relates to #3658

- add a workflow to run tests in a random order on the 25th of every month
- only runs test on ubuntu with python 3.11

There is no guarantee that if tests are not independent this workflow will systematically fail, but this is better than nothing.


Remi-Gau:
> LGTM, can you just add a section in .github/workflows/README.md?

Done

Remi-Gau:
merging
thanks @NicolasGensollen @ymzayek

PR #3698, opened 2023-04-17
[MAINT] remove comment relative to pandas precision bug

### LINKED ISSUE #3620: check if  `pd.option_context("display.precision", 2)` are still necessary
Let's check if this comment is still relevant or needs to be updated because now we support pandas>=1.1.5

_Originally posted by @ymzayek in https://github.com/nilearn/nilearn/pull/3615#discussion_r1147292448_
       

Remove them if that's not the case

Remi-Gau:
Closes #3620 

- remove workaround to avoid an old pandas bug that existed in a version of pandas that we no longer support

### DIFF
--- a/nilearn/reporting/glm_reporter.py
+++ b/nilearn/reporting/glm_reporter.py
@@ -171,12 +171,6 @@ def make_glm_report(
         Contains the HTML code for the GLM Report.
 
     """
-    """
-    Bug in Pandas 0.18 : https://github.com/pandas-dev/pandas/issues/13257
-    pd.set_option('display.precision', 2)
-    limits number of digits shown instead of precision.
-    Hence pd.option_context('display.precision', 2) has been used.
-    """
     if bg_img == "MNI152TEMPLATE":
         bg_img = MNI152TEMPLATE
     if not display_mode:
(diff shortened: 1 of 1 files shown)


Remi-Gau:
as far as I can tell from the doc the original use of the context manager was correct:
https://pandas.pydata.org/docs/reference/api/pandas.option_context.html#pandas.option_context

So now I am wondering wether we should not just simply remove the comment as it is more confusing than helpful.

@ymzayek any thought?

ymzayek:
@Remi-Gau I agree that the comment is confusing. I think it incorrectly references `pd.set_option` when the linked issue talks about a bug in `pandas.DataFrame.style.set_precision`, which should apply the precision setting to only the object that calls it. So this has been replaced by `pd.option_context` and context manager because the alternative `pd.set_option` works globally. The bug seems to be fixed in https://github.com/pandas-dev/pandas/pull/27934. So we could either leave as is and remove the comment or change it to use `pandas.DataFrame.style.set_precision`, but I don't see any clear benefits to changing it.

Remi-Gau:
> So we could either leave as is and remove the comment or change it to use `pandas.DataFrame.style.set_precision`, but I don't see any clear benefits to changing it.

yeah don't see the benefits of changing it and see potential drawbacks.

will only remove the comment and let's see what the other devs think about it.

bthirion (approved):
LGTM.

ymzayek (approved):

PR #3701, opened 2023-04-17
[MAINT] Refactor image.resampling.resample_img

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/issues/3655

- [x] refactor image.resampling.resample_img


nilearn/image/resampling.py
```diff
@@ -657,42 +545,235 @@ def resample_img(
...
+    # We convert to 'native' order to not have any issues either with
+    # 'little' or 'big' endian data dtypes (non-native endians).
+    A, _ = to_matrix_vector(transform_affine)
+    if len(A.shape) == 1 and not resampled_data_dtype.isnative:
+        resampled_data_dtype = resampled_data_dtype.newbyteorder("N")
+
+    return resampled_data_dtype
+
+
+def _get_bounding_box(target_affine, target_shape, affine, data):
```
jeromedockes, line 610: maybe this could be called something like `_check_target_affine_and_shape` -- functions in nilearn & scikit-learn that validate inputs and transform them or load defaults if necessary are often called `_check` something. and the docstring could state that this is a helper for `resample_img` to infer the missing bits of target affine and shape when they are not fully specified. side note -- it's ok to keep docstrings of private functions short, for example here as the parameters are those of resample_img you could skip it. the `Returns` section would actually be more informative

nilearn/image/resampling.py
```diff
@@ -657,42 +545,235 @@ def resample_img(
...
+        # Embed target_affine in 4x4 shape
+        missing_offset = True
+        target_affine_tmp = np.eye(4)
+        target_affine_tmp[:3, :3] = target_affine
+        target_affine = target_affine_tmp
+    else:
+        missing_offset = False
+        target_affine = target_affine.copy()
+
+    transform_affine = np.linalg.inv(target_affine).dot(affine)
```
jeromedockes, line 646: this seems to be the inverse of the `transform_affine` defined in `resample_img`? only the shape of the `transform_affine` in `resample_img` matters, but it's a bit confusing they have the same name

nilearn/image/resampling.py
```diff
@@ -441,73 +442,26 @@ def resample_img(
...
-                "Please check the provided file. "
-                "Results may not be as expected."
-            )
-
     # noop cases
     if target_affine is None and target_shape is None:
         if copy and not input_img_is_string:
             img = _utils.copy_img(img)
         return img
+
```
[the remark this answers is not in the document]
Remi-Gau: > (I know this part is unchanged from the main branch, but) I'm a bit confused as this seems to check the same thing as the check below (shape and affine are the same)? also, I don't understand why in this case the `copy` parameter is being ignored? maybe we could clarify this a bit

refactoring also helps better understand / explain the code: will try to see if there is something happening that is not obvious from just reading the code.

nilearn/image/resampling.py
```diff
@@ -657,42 +544,235 @@ def resample_img(
...
+    return resampled_data
+
+
+def _interpolation_order(interpolation):
+    if interpolation == "continuous":
+        return 3
+    elif interpolation == "linear":
+        return 1
+    elif interpolation == "nearest":
+        return 0
```
jeromedockes, line 705: maybe `{"continuous": 3, "linear": 1, "nearest": 0}[interpolation]`? to make it clear there is no default case and we don't want to return None

nilearn/image/resampling.py
```diff
@@ -620,8 +509,6 @@ def resample_img(
         dtype=resampled_data_dtype,
     )
 
-    all_img = (slice(None),) * 3
-
     # if (A == I OR some combination of permutation(I) and sign-flipped(I)) AND
     # all(b == integers):
     if (
```
jeromedockes, line 514: here also we might want 2 helper functions for the case where it is just a question of indexing (if) and the case where actual resampling takes place (else)?

Remi-Gau:
will also close this one and refactor across smaller PRs

PR #3702, opened 2023-04-18
[ENH] implement logger for first_level_from_bids

### LINKED ISSUE #3686: add a logger to "first_level_from_bids"
I am having to pass this verbose parameter around a bit too much to be able to control whether warnings, messages should be printed...

This should probably be handled better by implementing a logger in another PR.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/3605#discussion_r1152080917_

Remi-Gau:
Closes #3686

- shows a proof of concept to implement a logger for first_level_from_bids

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -49,6 +49,11 @@ from sklearn.base import clone
 from sklearn.cluster import KMeans
 
 
+import logging
+
+logging.basicConfig(level=logging.ERROR)
+logger = logging.getLogger(name="nilearn")
+
 def mean_scaling(Y, axis=0):
     """Scaling of the data to have percent of baseline change \
     along the specified axis.
@@ -928,6 +933,14 @@ def first_level_from_bids(dataset_path,
         Items for the FirstLevelModel fit function of their respective model.
 
     """
+
+    if verbose == 1:
+        logger.setLevel(logging.WARNING)
+    elif verbose == 2:
+        logger.setLevel(logging.INFO)
+    elif verbose == 3:
+        logger.setLevel(logging.DEBUG)
+
     if slice_time_ref == 0:
         warn(
             'Starting in version 0.12, slice_time_ref will default to None.',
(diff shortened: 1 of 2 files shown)


nilearn/glm/first_level/first_level.py
```diff
@@ -49,6 +49,11 @@
 from sklearn.cluster import KMeans
 
 
+import logging
+
+logging.basicConfig(level=logging.ERROR)
+logger = logging.getLogger(name="nilearn")
```
Remi-Gau, line 55: potentially create a function nilearn/_utils/logger.py ?

nilearn/glm/first_level/first_level.py
```diff
@@ -1527,8 +1511,7 @@ def _make_bids_files_filter(
     if extra_filter and supported_filters:
         for filter_ in extra_filter:
             if filter_[0] not in supported_filters:
-                if verbose:
-                    warn(
+                logger.warning(
```
Remi-Gau, line 1514: should warning be handled by the logger as well?

Remi-Gau:
@jeromedockes this is just a quick dirty example of how things could look like with a logger.

Let me know what you think.

jeromedockes (commented):
sorry I didn't realise implement a logger meant start using the `logging` standard library module. I think it's a good idea, but then we would probably want to use it consistently throughout the library and plan it a bit more... there is some discussion to be had as some people find `logging` confusing, and we should read through related scikit-learn discussions (linked below).

 and maybe wait for the formatting stuff to be over so we don't have too many widespread changes going on at the same time

for the specific case of 1st level from bids, I wonder if we should have a class to have fewer functions with a lot of parameters, and a way to store some info like the discovered files and produce [reports](https://github.com/nilearn/nilearn/tree/main/nilearn/reporting)?

Remi-Gau:
> sorry I didn't realise implement a logger meant start using the `logging` standard library module. I think it's a good idea, but then we would probably want to use it consistently throughout the library and plan it a bit more... 

yeah the "should we use a more systematic and planned approach to this?" was in the back of my mind and hence why I prefer to draft things here to get early feedback.

Do you think I should maybe rename the issue #3686 to reflect a potential bigger scope for this?

> there is some discussion to be had as some people find `logging` confusing, and we should read through related scikit-learn discussions (linked below).

thanks: will have a look


> and maybe wait for the formatting stuff to be over so we don't have too many widespread changes going on at the same time

Agreed

> for the specific case of 1st level from bids, I wonder if we should have a class to have fewer functions with a lot of parameters, and a way to store some info like the discovered files and produce [reports](https://github.com/nilearn/nilearn/tree/main/nilearn/reporting)?

Will look into this as well.

Issue #3660, opened 2023-04-05
[MAINT] switch to a src layout to organize the tests


Remi-Gau:
**Edit**

This issue has been renamed as the topic of having __init__.py in the tests folder fits better within a more global scope of whether we would like to adopt a source layout.

---

Unless I am mistaken, is there a reason why the nilearn tests should be part of the package?

If not, I suspect that there is a bunch of __init__.py files that can be removed.

bthirion:
I see that MNE has a bunch of __init__.py files in the test folders. Dunno why.

ymzayek:
I guess it's because we follow this [test as part of application code](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-as-part-of-application-code) setup. But there is a note below that section with this relevant part: 

> If you use one of the two recommended file system layouts above but leave away the __init__.py files from your directories, it should just work. From “inlined tests”, however, you will need to use absolute imports for getting at your application code.

So I guess they are not necessary if we're using absolute imports for our testing (IIUC).

Whether the tests should be part of the package is another question?

jeromedockes:
as @ymzayek says the `__init__.py` files are there because the tests are part of the package, and get installed along with the rest of nilearn. the reason that is the case is that when scikit-learn and then nilearn (and MNE) were created, that was the normal way of laying out a python package and its tests. It is now common (and IMO desirable) to use the "src" [layout](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code). we can discuss the pros and cons of making that change.

to summarize -- I don't think there is any reason to remove these `__init__.py` files

jeromedockes:
with the old setuptools configuration (0.10.0 release), if the `__init__.py` files are missing the test packages do not get copied to the installation directory when installing nilearn, so for example `pytest --pyargs nilearn` does not run the tests
with the new configuration they do get installed, but as they are missing an `__init__.py` they are [implicit namespace packages](https://peps.python.org/pep-0420/), which is not necessarily a big problem but does not make much sense as we are not in the situation they are intended for:

> Namespace packages are a mechanism for splitting a single Python package across multiple directories on disk

you can check they are namespace packages by importing one of the packages missing the `__init__.py`: it has no `__file__`, its `__path__` is a `_NamespacePath`, its `__loader__` is a `_NamespaceLoader`

[...]

PR #3712, opened 2023-04-20
[MAINT] Remove data extension from .gitignore

ymzayek:
Related to #3660 

The extension .tar.gz in .gitigore should not be ignored. It leads to test data being left out when building the distribution because hatchling uses .gitignore by default to exclude files during build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 *.nt.bz2
 *.swp
 *.swo
-*.tar.gz
 *.tgz
 *.zip
 *#
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,6 @@
 *.nt.bz2
 *.swp
 *.swo
-*.tgz
-*.zip
 *#
 
 .mypy_cache


ymzayek:
I think the first 10 lines of .gitignore could probably be removed but this could be done in another PR if there are any concerns. What do others think?

Remi-Gau:
> I think the first 10 lines of .gitignore could probably be removed but this could be done in another PR if there are any concerns. What do others think?

I did not want to remove those when I last reorganized the gitignore because I assumed they had a good reason to be there. If people with more experience of the codebase agree to nuke those lines, then fine with me.

jeromedockes:
> I think the first 10 lines of .gitignore could probably be removed

most of them are needed: `.pyc` are python bytecode, they will show up if you import nilearn from the source directory (eg if you install it in editable mode), `.so` are shared object files, shouldn't show up because we don't have compiled extensions but having them in gitignore doesn't hurt, `.swp`, `.swo`, `~` and `#` are used by editors for backups of files or buffers.

removing all the compressed archive extensions `.tar.gz`, `.tgz`, `.zip` is fine by me

jeromedockes:
btw I noticed the "created during testing" section of the gitignore; even better would be not to create files outside of pytests' tmp folder in tests. examples should also probably avoid writing files, or clean them up

ymzayek:
> > I think the first 10 lines of .gitignore could probably be removed
> 
> most of them are needed: `.pyc` are python bytecode, they will show up if you import nilearn from the source directory (eg if you install it in editable mode), `.so` are shared object files, shouldn't show up because we don't have compiled extensions but having them in gitignore doesn't hurt, `.swp`, `.swo`, `~` and `#` are used by editors for backups of files or buffers.
> 
> removing all the compressed archive extensions `.tar.gz`, `.tgz`, `.zip` is fine by me

Ok thanks for the clarifications. I will then just remove these 3 archive types

bthirion (approved):
Thx for fixing this !

PR #3713, opened 2023-04-20
[MAINT] Run pytest from different directory than root of repo

ymzayek:
Related to #3660

The testing on azure pipelines should fail until #3712 is merged

This change will help us catch if there are any missing files in a build distribution as detailed in https://github.com/nilearn/nilearn/issues/3660#issuecomment-1513631006


azure-pipelines.yml
```diff
@@ -39,6 +39,8 @@ jobs:
     displayName: 'Install dependencies'
 
   - script: |
+      mkdir temp
```
Remi-Gau, line 42: If we plan to keep this, do you think you can add a comment explaining why this is here?

Because if I did not know the context I would believe this to be "dead code" that was not removed.

ymzayek: Yes you're right!

ymzayek:
As expected the azure testing CI now fails: https://dev.azure.com/Parietal/Nilearn/_build/results?buildId=6357&view=logs&j=e8139889-c603-54b9-e136-6a67238aab48&t=2139de86-2315-50e4-7635-6ab77478eb0d&l=20975

I will rebase on main after merging #3712 and see if it passes

ymzayek:
It passes after merging 3712 so I think it's safe to merge

PR #3723, opened 2023-04-25
[Fix] Fix pathlib.Path not being counted as Niimg-like in new_image_like

mcsitter:
I received `TypeError: The reference image should be a niimg` when putting a `pathlib.Path` as an argument into `nilearn.image.new_image_like`. I fixed it to properly accept Niimg-like objects, which include `pathlib.Path` in its definition.

- Check for `isinstance` of `pathlib.Path` as well as `str`

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -295,7 +295,7 @@ authors:
   - given-names: Matthieu
     family-names: Joulot
     website: https://github.com/MatthieuJoulot
-  - given-names: Maximilian
+  - given-names: Maximilian Cosmo
     family-names: Sitter
     website: https://github.com/mcsitter
   - given-names: Mehdi
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -295,6 +295,9 @@ authors:
   - given-names: Matthieu
     family-names: Joulot
     website: https://github.com/MatthieuJoulot
+  - given-names: Maximilian
+    family-names: Sitter
+    website: https://github.com/mcsitter
   - given-names: Mehdi
     family-names: Rahim
     website: https://github.com/mrahim
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -30,7 +30,7 @@ Fixes
 
 - Relax the :func:`~nilearn.interfaces.fmriprep.load_confounds` confounds selection on `cosine` as not all confound files contained the variables (:gh:`3816` by `Hao-Ting Wang`_).
 
-- Fix pathlib.Path not being counted as Niimg-like object in :func:`~image.image.new_img_like` and :func:`~image.image.mean_img` (:gh:`3723` by `Maximilian Cosmo Sitter`_).
+- Fix pathlib.Path not being counted as Niimg-like object in :func:`~image.new_img_like` (:gh:`3723` by `Maximilian Cosmo Sitter`_).
 
 
 Enhancements
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -30,6 +30,9 @@ Fixes
 
 - Relax the :func:`~nilearn.interfaces.fmriprep.load_confounds` confounds selection on `cosine` as not all confound files contained the variables (:gh:`3816` by `Hao-Ting Wang`_).
 
+- Fix pathlib.Path not being counted as Niimg-like object in :func:`~image.image.new_img_like` and :func:`~image.image.mean_img` (:gh:`3723` by `Maximilian Cosmo Sitter`_).
+
+
 Enhancements
 ------------
 
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -7,6 +7,7 @@ See also nilearn.signal.
 
 import collections.abc
 import copy
+import pathlib
 import warnings
 
 import nibabel
@@ -530,9 +531,9 @@ def mean_img(imgs, target_affine=None, target_shape=None, verbose=0, n_jobs=1):
 
     """
     imgs = stringify_path(imgs)
-    is_str = isinstance(imgs, str)
+    is_path = isinstance(imgs, str) or isinstance(imgs, pathlib.Path)
     is_iterable = isinstance(imgs, collections.abc.Iterable)
-    if is_str or not is_iterable:
+    if is_path or not is_iterable:
         imgs = [
             imgs,
         ]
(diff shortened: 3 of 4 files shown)


nilearn/image/image.py
```diff
@@ -531,9 +532,9 @@ def mean_img(imgs, target_affine=None, target_shape=None, verbose=0, n_jobs=1):
 
     """
     imgs = stringify_path(imgs)
-    is_str = isinstance(imgs, str)
+    is_path = isinstance(imgs, str) or isinstance(imgs, pathlib.Path)
```
NicolasGensollen, line 535: `imgs` should be a string (or an iterable) at this point because `stringify_path` has been called on it:

https://github.com/nilearn/nilearn/blob/1b7994104d08a117b859e784f3de567be87f0e59/nilearn/_utils/helpers.py#L153-L168

NicolasGensollen (commented):
Thanks for contributing @mcsitter 
I might be missing something, but you should be able to provide a `Path` object to these functions. The path object is first casted to a string by `stringify_path`:

```python
>>> stringify_path(Path("./foo.json"))
'foo.json'
```

ymzayek:
@mcsitter could you provide a reproducible example that will fail if we run it? As @NicolasGensollen pointed out, this case is supposed to be handled by `stringify_path`. Also can you check and report the version of nilearn in your environment? This would fail in nilearn versions less than 0.9.2.

nilearn/image/image.py
mcsitter: Sorry, I should not have changed `mean_img`. The error appears when calling `nilearn.image.new_image_like`. I added a new commit reversing the change.

mcsitter:
@ymzayek I am on nilearn version 0.10.0. Here is some could which should allow you to reproduce the TypeError:
```python
import pathlib

import nilearn.image
from nilearn.datasets import MNI152_FILE_PATH

image_path = pathlib.Path(MNI152_FILE_PATH)
image = nilearn.image.load_img(image_path)
nilearn.image.new_img_like(image_path, image.get_fdata())
```

ymzayek:
@mcsitter ok I see, indeed `new_img_like` doesn't take a path-like object for the reference image. Though I suggest to use `stringify_path` function as a solution like the other functions in this module and then leave the rest of the code untouched

mcsitter:
@ymzayek I changed the code to use `stringify_path`.

bthirion (commented):
Thx !  We should add the corresponding unit test to enforce this behavior. Can you do that ?

ymzayek:
@mcsitter can you define a test under `nilearn/image/tests/test_image.py` ?
Also, you need to add your name in `doc/changes/names.rst` and you'll need to rebase on main and resolve the conflicts. Let me know if you need any help.

ymzayek:
@mcsitter are you able to pick this back up and finish it? Otherwise, I can take it over and make the necessary changes.

mcsitter:
I am not sure I did the rebasing correctly... But I added a unit test, hopefully, a useful one, and added my name to the `CITATION.cff`.

Feel free to let me know if there is anything else I can do or how I would fix any mishaps regarding the rebasing.

Remi-Gau:
Given the number of line and file changes I would say that something looks not right.

Am AFK at the moment but I will have a look when on a screen that is more than 20 characters wide.

CITATION.cff
```diff
@@ -295,6 +295,9 @@ authors:
   - given-names: Matthieu
     family-names: Joulot
     website: https://github.com/MatthieuJoulot
+  - given-names: Maximilian
```
ymzayek, line 298:
```suggestion
  - given-names: Maximilian Cosmo
```

doc/changes/latest.rst
```diff
@@ -30,6 +30,9 @@ Fixes
 
 - Relax the :func:`~nilearn.interfaces.fmriprep.load_confounds` confounds selection on `cosine` as not all confound files contained the variables (:gh:`3816` by `Hao-Ting Wang`_).
 
+- Fix pathlib.Path not being counted as Niimg-like object in :func:`~image.image.new_img_like` and :func:`~image.image.mean_img` (:gh:`3723` by `Maximilian Cosmo Sitter`_).
```
ymzayek, line 33:
```suggestion
- Fix pathlib.Path not being counted as Niimg-like object in :func:`~image.new_img_like` (:gh:`3723` by `Maximilian Cosmo Sitter`_).
```

nilearn/image/tests/test_image.py
```diff
@@ -654,6 +654,20 @@ def test_new_img_like():
...
+    nifti_path = tmp_path / "sample.nii"
+    assert isinstance(nifti_path, Path)
+
+    data = np.random.rand(10, 10, 10)
+    img = Nifti1Image(data, np.eye(4))
+    nibabel.save(img, nifti_path)
+
+    new_data = np.random.rand(10, 10, 10)
+    new_img = new_img_like(nifti_path, new_data)
+    assert new_img.shape == (10, 10, 10)
```
ymzayek, line 668: Also to test L762 in image.py

```suggestion
    assert new_img.shape == (10, 10, 10)
    
    # Check that list of pathlib.Path also accepted
    new_img = new_img_like([nifti_path], new_data)
    assert new_img.shape == (10, 10, 10)
```

ymzayek (commented):
@mcsitter I fixed the rebase and made a few comments to get the doc build to pass. This should be good to go now. @Remi-Gau can you also give a final review?

Remi-Gau (approved):
yup it looks good to me

ymzayek:
Oops let me fix the flake8 error

nilearn/image/tests/test_image.py
```diff
@@ -654,6 +654,24 @@ def test_new_img_like():
     assert_array_equal(get_data(img_nifti2), get_data(img2_nifti2))
 
 
+def test_new_img_like_accepts_paths(tmp_path):
+    """Check that new_img_like can accept instances of pathlib.Path."""
+    nifti_path = tmp_path / "sample.nii"
+    assert isinstance(nifti_path, Path)
+
+    data = np.random.rand(10, 10, 10)
+    img = Nifti1Image(data, np.eye(4))
```
Remi-Gau, line 663: not needed for now but I think there may be a lot random img creation in the whole test suite that could be fixturised

ymzayek: Linking #3725 and #3331

PR #3727, opened 2023-04-26
[MAINT] refactor nilearn/datasets (sourcery, f-strings)

Remi-Gau:
Relates to https://github.com/orgs/nilearn/projects/2/views/1?pane=issue&itemId=30677086

-   apply sourcery refactoring suggestions
-   use f strings everywhere


nilearn/datasets/utils.py
```diff
@@ -31,20 +31,18 @@ def md5_hash(string):
...
 
 
 def _md5_sum_file(path):
     """Calculate the MD5 sum of a file."""
     with open(path, "rb") as f:
         m = hashlib.md5()
         while True:
-            data = f.read(8192)
-            if not data:
+            if data := f.read(8192):
```
Remi-Gau, line 42: avoid walrus operator till we deprecate python 3.7

nilearn/datasets/tests/test_neurovault.py
```diff
@@ -319,42 +319,22 @@ def test_scroll_server_results():
...
     is_null = neurovault.IsNull()
     assert is_null != "a"
-    assert not is_null != ""
-    assert "a" != is_null
-    assert not "" != is_null
-    assert not is_null == "a"
     assert is_null == ""
-    assert not "a" == is_null
-    assert "" == is_null
     assert str(is_null) == "IsNull()"
```
Remi-Gau, line 323: I get that this test and the following are aimed at testing the `__eq__` magic method of those different classes but is there any reason to test all of the following:

```python
assert is_null != "a"
assert not is_null == "a"
assert "a" != is_null
assert not "a" == is_null
```

I simplified the tests assuming that those are just unneeded test duplicates.

Let me know if I am missing something.

If there is any reason to keep those, it is then probably to use some sort of test parametrization to make the tests more readable.

bthirion: I think you're right.

bthirion (approved):
LGTM overall

Remi-Gau:
There are quite a few more f stings to implement in nilearn/datasets but I will do that in another PR

PR #3722, opened 2023-04-25
Adding keep_masked_labels parameter to Labels_masker

mtorabi59:
Addresses #3085 .

Changes proposed in this pull request:

- Adding `keep_masked_labels` parameter to `img_to_signals_labels`, and `NiftiLabelsMasker`
- if `keep_masked_labels=True`, which is the default, it will maintain the old behavior. ow, it will remove the labels, or regions, that were masked and became empty by `mask_img`.

PR #3621, opened 2023-03-24
[DOC] move defaults in docs.py

Remi-Gau:
Relates to #3519 

Changes proposed in this pull request:

- [x] move where defaults are located in the templates arguments in docs.py
- [x] add info in contributing doc to show an example of a documented function:
- [x] minor style fixing (semantic line break, markup variables in doc strings)

PR #3732, opened 2023-05-03
add keep_masked_maps to NiftiMapsMasker

mtorabi59:
Addresses #3085  .

- Adding `keep_masked_maps` parameter to `img_to_signals_maps`, and `NiftiMapsMasker`
- If `keep_masked_maps=True`, which is the default, it will maintain the old behavior. ow, it will remove the maps, or regions, that were masked and became empty by `mask_img`.


mtorabi59:
So I found out that interestingly a parameter called `keep_empty` was already implemented in `img_to_signals_maps`. I don't know if it was implemented recently. Anyways, I am using that parameter to keep or remove the maps that are masked by `mask_img`.

mtorabi59:
Is the deprecation warning in the right place?
In general I was thinking should we keep the option to keep or remove empty regions in `img_to_signals_maps` and `img_to_signals_labels` and remove it only from `NiftiLabelsMasker` and `NiftiMapsMasker` or we should delete it from there too (I mean after the Deprecation cycle).  Because for instance the `keep_empty` option has been there in `img_to_signals_maps` from before; should we delete that too?

jeromedockes:
> Is the deprecation warning in the right place?
> In general I was thinking should we keep the option to keep or remove empty regions in `img_to_signals_maps` and `img_to_signals_labels` and remove it only from `NiftiLabelsMasker` and `NiftiMapsMasker` or we should delete it from there too (I mean after the Deprecation cycle). Because for instance the `keep_empty` option has been there in `img_to_signals_maps` from before; should we delete that too?

I would say yes, the warning is in the right place and we should remove it from the function too -- I don't think keeping outputs for empty regions with arbitrary values is a useful feature. but you might want to use the `stacklevel` parameter of `warnings.warn` so the users see where the issue is in their own code rather than in nilearn code

mtorabi59:
@bthirion could you share the code that results in the same behavior? for me it does make a difference. the following code returns signals with shape (17, 3) when `keep_masked_maps=False` and (17, 8) when `keep_masked_maps=True`:

      import warnings
      from nilearn.regions.tests.test_signal_extraction import AFFINE_EYE, get_data
          generate_fake_fmri,
          generate_labeled_regions,
      )
      import numpy as np
      
      warnings.simplefilter('always', DeprecationWarning)
      
      N_REGIONS = 8
      N_TIMEPOINTS = 17
      SHAPE = (8, 9, 10)
      
      def _create_mask_with_3_regions_from_labels_data(labels_data, affine):
          """Create a mask containing only 3 regions."""
          mask_data = (labels_data == 1) + (labels_data == 2) + (labels_data == 5)
          return Nifti1Image(mask_data.astype(np.int8), affine)
      
      def fmri_img():
          return generate_fake_fmri(shape=SHAPE, affine=AFFINE_EYE)[0]
      
      def labeled_regions():
          labels = list(range(N_REGIONS + 1))  # 0 is background
          return generate_labeled_regions(
              shape=SHAPE, n_regions=N_REGIONS, labels=labels
          )
      
      labeled_regions = labeled_regions()
      fmri_img = fmri_img()
      
      labels = list(range(N_REGIONS + 1))
      labels_data = get_data(labeled_regions)
      # Convert to maps
      maps_data = np.zeros(SHAPE + (N_REGIONS,))
      for n, l in enumerate(labels):
          if n == 0:
              continue
          maps_data[labels_data == l, n - 1] = 1
      
      maps_img = Nifti1Image(maps_data, labeled_regions.affine)
      
      # a mask, keeping only 3 regions.
      mask_img = _create_mask_with_3_regions_from_labels_data(
          labels_data, labeled_regions.affine
      )
      
      ## using the NiftiMapsMasker
      from nilearn.maskers import NiftiMapsMasker
      
      masker = NiftiMapsMasker(maps_img=maps_img,
                                 mask_img=mask_img,
                                 standardize=False,
                                 keep_masked_maps=False
                                 )
      
      masker.fit()
      signals = masker.transform(fmri_img)
      
      print(signals.shape)
      print(signals)
      
      signals = masker.transform(fmri_img)
      
      print(signals.shape)

mtorabi59:
@bthirion Do you think it needs to be clarified?

bthirion:
> @bthirion Do you think it needs to be clarified?

I think it should for further reference and fixes. Thx !

nilearn/maskers/nifti_maps_masker.py
```diff
@@ -74,6 +77,18 @@ class NiftiMapsMasker(BaseMasker, _utils.CacheMixin):
     %(memory)s
     %(memory_level)s
     %(verbose0)s
+
+    keep_masked_maps : :obj:`bool`, optional
```
htwangtw, line 81: It's really unclear for now but we can combine some other information from `_trim_maps` to make this clearer. Here's my understanding of the behaviour and we can iterate on this:

When user supply a mask through parameter `mask_img`, the atlas will be masked by `mask_img` before any sign extraction happens. Some maps in the atlas will contain no brain coverage after applying the mask, resulting a map containing only zeroes (i.e. an invalid map for signal extraction). 

If `True`, masked atlas with these invalid maps will be retained in the output, resulting the correspondent time series containing zeros only. If `False`, the invalid maps will be removed from the trimmed altas, hence no empty time series will be present in the output.

Here I copy the docstrings from `_trim_maps` for reference:
```
Crop maps using a mask.

No consistency check is performed (esp. on affine). Every required check
must be performed before calling this function.

Parameters
----------
maps : :class:`numpy.ndarray`
    Set of maps, defining some regions.

mask : :class:`numpy.ndarray`
    Definition of a mask. The shape must match that of a single map.

keep_empty : :obj:`bool`, optional
    If False, maps that lie completely outside the mask are dropped from
    the output. If True, they are kept, meaning that maps that are
    completely zero can occur in the output.
    Default=False.

order : "F" or "C", optional
    Ordering of the output maps array (trimmed_maps).
    Default="F".

Returns
-------
trimmed_maps : :class:``numpy.ndarray`
    New set of maps, computed as intersection of each input map and mask.
    Empty maps are discarded if keep_empty is False, thus the number of
    output maps is not necessarily the same as the number of input maps.
    shape: mask.shape + (output maps number,). Data ordering depends
    on the "order" parameter.

maps_mask : :class:`numpy.ndarray`
    Union of all output maps supports. One non-zero value in this
    array guarantees that there is at least one output map that is
    non-zero at this voxel.
    shape: mask.shape. Order is always C.

indices : :class:`numpy.ndarray`
    Indices of regions that have an non-empty intersection with the
    given mask. len(indices) == trimmed_maps.shape[-1].

```

nilearn/regions/signal_extraction.py
```diff
@@ -430,9 +446,23 @@ def img_to_signals_maps(imgs, maps_img, mask_img=None):
         maps_data, maps_mask, labels = _trim_maps(
             maps_data,
             _safe_get_data(mask_img, ensure_finite=True),
-            keep_empty=True,
+            keep_empty=keep_masked_maps,
         )
         maps_mask = _utils.as_ndarray(maps_mask, dtype=bool)
+        if keep_masked_maps:
+            warnings.warn(
```
htwangtw, line 453: If we update the docstring, we can improve this message as well!

htwangtw (commented):
I had a look at the documentation and here's my attempt to understand it. I hope this makes the behaviour clearer and help us to improve the documentation :tada:

bthirion (approved):
I think that this is clearer now. Thx !

nilearn/regions/signal_extraction.py
```diff
@@ -480,6 +481,21 @@ def img_to_signals_maps(
...
+                labels_after_mask
+            )
+            # Raising a warning if any map is removed due to the mask
+            if len(labels_diff) > 0:
+                warnings.warn(
+                    "After applying mask to the maps image, "
+                    "the following maps were "
+                    f"removed: {labels_diff}. "
+                    "Map image only contains "
+                    f"{len(labels_after_mask)} maps.",
```
htwangtw, line 496:
```suggestion
                    f"Out of {len(labels_before_mask)} maps, the "
                    "masked map image only contains "
                    f"{len(labels_after_mask)} maps.",
```

nilearn/regions/signal_extraction.py
```diff
@@ -427,12 +444,58 @@ def img_to_signals_maps(imgs, maps_img, mask_img=None):
...
+                'Starting from version 0.13, the default behavior will be '
+                'changed to "keep_masked_maps=False". If "keep_masked_maps" '
+                'is set to False, the invalid maps will be removed from the '
+                'trimmed atlas, ensuring no empty time series are present in '
+                'the output (new behavior). '
+                'To explicitly disable the retention of masked '
+                'maps, specify the parameter "keep_masked_maps=False" when '
+                'initializing the "NiftiMapsMasker" object.'
+                '"keep_masked_maps" parameter will be removed '
+                'in version 0.15.',
```
htwangtw, line 480: This is way too long for a deprecation warning.

I would just state three things

1. The output will contain empty time series.
2. Starting from version 0.13, the default behavior will be `keep_masked_maps=False`
3. `keep_masked_maps` will be removed in version 0.15

Just point the users to read the docs.

Probably tighten this up and put it in the change log?

htwangtw (commented):
Thank you! The docs is really clear now. Just have another look at the deprecation warning and perhaps move some of the explanation to tje change log. Otherwise it's almost good!

mtorabi59:
How is this? Does it need to be shorter?

nilearn/regions/signal_extraction.py
```diff
@@ -397,6 +402,18 @@ def img_to_signals_maps(imgs, maps_img, mask_img=None):
         Every point outside the mask is considered
         as background (i.e. outside of any region).
 
+    keep_masked_maps : :obj:`bool`, optional
+        If True, masked atlas with invalid maps (maps with no brain coverage
+        after applying the mask) will be retained in the output, resulting
+        in corresponding time series containing zeros only. If False, the
```
jeromedockes, line 408: if you want you can use the `_utils.fill_doc` to avoid having to update the docstring in several places

jeromedockes (commented):
same comment as for #3722  -- how do we make it easier for users to match masker output dimensions with atlas maps / region names? is this left for another PR?

mtorabi59:
> same comment as for #3722 -- how do we make it easier for users to match masker output dimensions with atlas maps / region names? is this left for another PR?

Yes, it will be handled in another PR.

ymzayek (commented):
Since there is overlap with #3722, let's work on merging that one first and then rebasing this on main. Pretty much the same comments apply but I'll give a more thorough review after that.

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -513,6 +513,55 @@ def test_signal_extraction_with_maps_and_labels(labeled_regions, fmri_img):
...
+        "Out of 8 maps, the "
+        "masked map image only contains "
+        "3 maps.",
+    ):
+        maps_signals, maps_labels = img_to_signals_maps(
+            fmri_img,
+            maps_img,
+            mask_img=mask_img,
+            keep_masked_maps=False
+        )
```
ymzayek, line 533:
```suggestion
        maps_signals, maps_labels = img_to_signals_maps(
            fmri_img, maps_img, mask_img=mask_img, keep_masked_maps=False
        )
```

ymzayek (commented):
LGTM! Just minor formatting things

ymzayek (approved):
LGTM thanks @mtorabi59 !

mtorabi59:
> LGTM thanks @mtorabi59 !

Thank you @ymzayek !!

nilearn/_utils/docs.py
```diff
@@ -847,6 +847,22 @@ def custom_function(vertices):
...
+    in corresponding time series containing zeros only. If False, the
+    invalid maps will be removed from the trimmed atlas, resulting in
+    no empty time series in the output.
+
+    .. deprecated:: 0.10.2.dev
+
+        The 'True' option for ``keep_masked_maps`` is deprecated.
+        The default value will change to 'False' in 0.13,
+        and the ``keep_masked_maps`` parameter will be removed in 0.15.
+"""
```
Remi-Gau, line 864: @mtorabi59 
entries in this docs.py are supposed to be sorted alphabetically so this should probably be moved.

I can do it in #3621 after this is merged, just wanted to let you know.  😉

mtorabi59: @Remi-Gau I'll have that in mind for next time! Thnx!

Remi-Gau: no worries but having things sorted can become helpful for long lists like this (also helps make sure you prevent the blood pressure of maintainers with sub-clinical  OCD from rising too much)

PR #3733, opened 2023-05-03
[FIX] Missing values in `Decoder.cv_params_` when `param_grid` is a list of dicts with different keys

### LINKED ISSUE #3715: [BUG]: missing values in `Decoder.cv_params_` when `param_grid` is a list of dicts with different keys
If `param_grid` is a list of dicts and the dicts have different keys, the fitted Decoder's `cv_params_` output can contain lists that have less elements than the number of CV folds. For example, if `param_grid` is
```python
param_grid = [
    {
        "penalty": ["l2"],
        "C": [100, 1000],
    },
    {
        "penalty": ["l1"],
        "dual": [False],    # "dual" is not in the first dict
        "C": [100, 1000],
    },
]
```

Then it is possible that `decoder.cv_params_["shoe"]` has something like
```python
{
    'C': [1000, 1000, 100, 100, 1000], 
    'dual': [False, False, False],    # less than 5, not aligned with the other hyperparameter values
    'penalty': ['l1', 'l2', 'l1', 'l1', 'l2'],
}
```

_Originally posted by @michellewang in https://github.com/nilearn/nilearn/pull/3626#discussion_r1172699079_

michellewang:
Closes #3715.

- Add test (`nilearn.decoding.decoder.tests.test_decoder.py`)
- In [nilearn.decoding.decoder._parallel_fit](https://github.com/nilearn/nilearn/blob/main/nilearn/decoding/decoder.py#L163), before looping over the hyperparameters, collect all the unique keys in `param_grid`
- Inside the loop, check if `best_param` is missing any keys, and add them if necessary

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -11,6 +11,8 @@ NEW
 Fixes
 -----
 
+- :bdg-dark:`Code` Fix bug where the `cv_params_` attribute of fitter Decoder objects sometimes had missing entries if `grid_param` is a sequence of dicts with different keys (:gh:`3733` by `Michelle Wang`_).
+
 Enhancements
 ------------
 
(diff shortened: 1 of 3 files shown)


bthirion (approved):
LGTM, thx.

Issue #3718, opened 2023-04-21
cancel concurrent CI workflows ?


Remi-Gau:
Some CI workflows (testing, doc build) take a long time to run.

Those are set to keep running even if new commits are pushed on a branch / PR that can render the end result of that workflow obsolete.

This also means that compulsive "git pusher" (I am one of them 🙈) tend to end up with many CI workflows queued for a given PR, which hogs resources and also potentially delays the start of workflows from other PRs / devs.

I'd be tempted to cancel concurrent previous workflows for testing and doc, the way it is done with the black / flake8 / isort workflows: https://github.com/nilearn/nilearn/blob/d50ee5201c1ad7f5d583c5c4d0ec0ea6d8ef6a5d/.github/workflows/isort.yml#L12

Wondering how others feel about this.

ymzayek:
I would be for it though it's the case that sometimes it helps in the debugging process to be able to track which push commits passed and which didn't. But maybe the benefits outweigh the costs especially for these heavier jobs

ymzayek:
Also, we probably would want to restrict this to PRs as applying it on main can make it more difficult to track sources of failures

Remi-Gau:
> Also, we probably would want to restrict this to PRs as applying it on main can make it more difficult to track sources of failures

Agreed.

I think this has some info on how to set it up.
https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value 

> I would be for it though it's the case that sometimes it helps in the debugging process to be able to track which push commits passed and which didn't. But maybe the benefits outweigh the costs especially for these heavier jobs

Yeah that's pretty much the trade-off that I had in mind too: though in most cases once a test fails in CI it is faster to git bisect locally to spot the guilty commit.

bthirion:
I'm always worried to make the behavior of these objects more complicated --- for the sake of long-term maintenance.
That's why I think the current behavior is OK, though suboptimal in terms of resource use.

ymzayek:
> Or we could try to make the test suite faster, but after playing around quite a bit in the test suite, I am not sure that we are going to make considerable gains on that front.

I think we should take another look at this proposition and prioritize it a bit. Even though it varies, the tests used to take 30-40 minutes to run on azure and now it seems closer to 1h which is going to start causing timeout failures like this one https://dev.azure.com/Parietal/Nilearn/_build/results?buildId=6386&view=results in PR #3720

[...]

PR #3747, opened 2023-06-08
[INFRA] Add timeout setting in Azure Pipelines CI 

ymzayek:
Related to #3718 

I'm implementing a solution given to me by Microsoft support which is to set the parameter timeoutInMinutes to 0. This allows for the maximum limit which for public projects is 6 hours. It doesn't really answer the question of why the tests are taking so long to run but for now it's enough to make sure our tests run fully.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,6 +10,7 @@ jobs:
 
 
 - job: 'run_tests_on_windows'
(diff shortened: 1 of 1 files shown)


bthirion:
If you merge with main you should see the errors disappear.

ymzayek:
Yes, that's the hope.

Merging

PR #3749, opened 2023-06-12
[MAINT] Catch warnings as errors on partial doc builds

ymzayek:
I suppressed expected warnings related to partial sphinx builds and changed the calls for partial builds to treat warnings as errors. This will help catch warnings that fail the full build on main while having had passed on a given PR's partial build.


ymzayek:
The failure is known and unrelated. I think we can merge this. It doesn't affect anything for users or contributors except that sphinx errors will now be easily caught during PRs. Let me know if anyone has any questions.

Remi-Gau (approved):
This looks good to me. Should hopefully help us catch issues early.

PR #3751, opened 2023-06-13
[FIX] force ordered output of _sanitize_reorder to match regex in test

### LINKED ISSUE #3750: test error for order of printed argument with python 3.11
See in a few runs of CI

For [example](https://github.com/nilearn/nilearn/actions/runs/5244068680/jobs/9469564671?pr=3742#step:8:14623)

```
 FAILED nilearn/plotting/tests/test_matrix_plotting.py::test_sanitize_reorder_error[None] - AssertionError: Regex pattern did not match.
 Regex: "Parameter reorder needs to be one of {False, True, 'complete', 'single', 'average'}"
 Input: "Parameter reorder needs to be one of {False, True, 'single', 'complete', 'average'}."
FAILED nilearn/plotting/tests/test_matrix_plotting.py::test_sanitize_reorder_error[foo] - AssertionError: Regex pattern did not match.
 Regex: "Parameter reorder needs to be one of {False, True, 'complete', 'single', 'average'}"
 Input: "Parameter reorder needs to be one of {False, True, 'single', 'complete', 'average'}."
FAILED nilearn/plotting/tests/test_matrix_plotting.py::test_sanitize_reorder_error[2] - AssertionError: Regex pattern did not match.
 Regex: "Parameter reorder needs to be one of {False, True, 'complete', 'single', 'average'}"
 Input: "Parameter reorder needs to be one of {False, True, 'single', 'complete', 'average'}."
= 3 failed, 2587 passed, 7 skipped, 2 xfailed, 1311 warnings in 1693.15s (0:28:13) =
```

Remi-Gau:
Closes #3750

- force to print VALID_REORDER_VALUES in an ordered way

### DIFF
--- a/nilearn/plotting/tests/test_matrix_plotting.py
+++ b/nilearn/plotting/tests/test_matrix_plotting.py
@@ -90,9 +90,16 @@ def test_sanitize_reorder_error(reorder):
     from ..matrix_plotting import _sanitize_reorder
     with pytest.raises(ValueError,
                        match=("Parameter reorder needs to be "
-                              f"one of {VALID_REORDER_VALUES}")):
+                              f"one of {list(VALID_REORDER_VALUES)}")):
         _sanitize_reorder(reorder)
 
+@pytest.mark.parametrize("reorder", [None, "foo", 2])
+def test_sanitize_reorder_error_should_fail_with_311(reorder):
+    from ..matrix_plotting import _sanitize_reorder
+    with pytest.raises(ValueError,
+                       match=("Parameter reorder needs to be "
+                              f"one of {VALID_REORDER_VALUES}")):
+        _sanitize_reorder(reorder)
 
 @pytest.fixture
 def mat():
--- a/nilearn/plotting/matrix_plotting.py
+++ b/nilearn/plotting/matrix_plotting.py
@@ -101,7 +101,8 @@ def _sanitize_reorder(reorder):
     VALID_REORDER_ARGS = {True, False, 'single', 'complete', 'average'}
     if reorder not in VALID_REORDER_ARGS:
         raise ValueError(
-            f"Parameter reorder needs to be one of {VALID_REORDER_ARGS}.")
+            f"Parameter reorder needs to be one of {list(VALID_REORDER_ARGS)}."
+        )
     reorder = 'average' if reorder is True else reorder
     return reorder
 


nilearn/plotting/tests/test_matrix_plotting.py
```diff
@@ -89,8 +89,7 @@ def test_sanitize_reorder(reorder):
 def test_sanitize_reorder_error(reorder):
     from ..matrix_plotting import _sanitize_reorder
     with pytest.raises(ValueError,
-                       match=("Parameter reorder needs to be "
-                              f"one of {VALID_REORDER_VALUES}")):
+                       match=("Parameter reorder needs to be ")):
```
Remi-Gau, line 92: My dirty fix is to make the regex less strict.

Remi-Gau:
@ymzayek 
quite unclear what kind of weird interaction led to such a specific failure, so for now I suggest making the the regex for match more general. LMKWYT

nilearn/plotting/tests/test_matrix_plotting.py
ymzayek: Yea I think that's ok. But do VALID_REORDER_ARGS and VALID_REORDER_VALUES need to be an unordered set? Can we just use a list instead?

Remi-Gau: I thought I had tried in the first commit but now realizing that the square brackets would turn add some "grouping" to the regex...

``
FAILED nilearn/plotting/tests/test_matrix_plotting.py::test_sanitize_reorder_error[None] - AssertionError: Regex pattern did not match.
 Regex: "Parameter reorder needs to be one of [False, True, 'average', 'single', 'complete']"
 Input: "Parameter reorder needs to be one of [False, True, 'average', 'single', 'complete']."
```

ymzayek: Ok right. you could maybe use `re.escape("Parameter reorder needs to be one of [False, True, 'average', 'single', 'complete'].")` but if you feel it's not worth the time we just go with the general pattern solution.

Remi-Gau: decided to stringify everything

Remi-Gau:
OK merging now since so we can avoid having this fail in other PRs

PR #3759, opened 2023-06-15
[MAINT] add workflow to check for f strings

### LINKED ISSUE #2528: Automate the Formatting stuff

Remi-Gau:
Closes #2528 
Relates to https://github.com/orgs/nilearn/projects/2/views/1?pane=issue&itemId=30677086

- add workflow to run flynt on main and PRs
- add flynt config to pyproject.toml
- run flynt

### DIFF
--- a/nilearn/_utils/data_gen.py
+++ b/nilearn/_utils/data_gen.py
@@ -677,7 +677,7 @@ def generate_group_sparse_gaussian_graphs(n_subjects=5,
     topology = topology > 0
     assert (np.all(topology == topology.T))
     logger.log(
-        f"Sparsity: {1.0 * topology.sum() / topology.shape[0] ** 2:f}",
+        f"Sparsity: {1.0 * topology.sum() / topology.shape[0] ** 2 :f}",
         verbose=verbose)
 
     # Generate temporal signals
--- a/nilearn/_utils/data_gen.py
+++ b/nilearn/_utils/data_gen.py
@@ -677,7 +677,7 @@ def generate_group_sparse_gaussian_graphs(n_subjects=5,
     topology = topology > 0
     assert (np.all(topology == topology.T))
     logger.log(
-        "Sparsity: {0:f}".format(1. * topology.sum() / (topology.shape[0]**2)),
+        f"Sparsity: {1.0 * topology.sum() / topology.shape[0] ** 2:f}",
         verbose=verbose)
 
     # Generate temporal signals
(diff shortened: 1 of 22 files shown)


Remi-Gau:
oh and it is already one the pre-commit hook that is run, this would just provide an official check in CI

ymzayek (approved):

nilearn/_utils/data_gen.py
```diff
@@ -677,7 +677,7 @@ def generate_group_sparse_gaussian_graphs(n_subjects=5,
     topology = topology > 0
     assert (np.all(topology == topology.T))
     logger.log(
-        "Sparsity: {0:f}".format(1. * topology.sum() / (topology.shape[0]**2)),
+        f"Sparsity: {1.0 * topology.sum() / topology.shape[0] ** 2:f}",
```
bthirion, line 680:
```suggestion
        f"Sparsity: {1.0 * topology.sum() / topology.shape[0] ** 2 :f}",
``` would be more readable IMHO

tsalo:
@Remi-Gau have you looked at `flake8-use-fstring`? I don't know if it's better or worse than `flynt`, but you'd be able to include it in the `flake8` workflow that already exists.

Remi-Gau:
Was not aware of this one. Will investigate.

Remi-Gau:
> @Remi-Gau have you looked at `flake8-use-fstring`? I don't know if it's better or worse than `flynt`, but you'd be able to include it in the `flake8` workflow that already exists.

OK as far I can tell the 2 are not completely mutually exclusive. Flynt will actually try to convert % and Format to f strings but can only do the most simple ones because it tries to be conservative.

The flake8 extension will flag % and format and potentially report them as error.

So the 2 could coexist but one will throw more errors than the other when used for Linting purposes.

I would suggest to start with Flynt and and then tackle the remaining non f strings. Then we can switch to flake8 extensions to ensure we stick to f strings from now on.

tsalo:
👍 sounds good to me. Thanks for taking a look at the flake8 extension.

nilearn/_utils/data_gen.py
Remi-Gau: let's see what black has to say about this

bthirion:
The PR LGTM, but I think lots of red CI.

Remi-Gau:
> The PR LGTM, but I think lots of red CI.

yeah I will rebase to clean some of that

bthirion (commented):
Now it mixed several things but LGTM overall.

Remi-Gau:
yeah some files should not be in that diff: will fix

PR #3761, opened 2023-06-15
[ENH] Add new returns to `NiftiLabelsMasker`'s `transform` output

mtorabi59:
Addresses #3085 .

- add new outputs to `NiftiLabelsMasker` `transform_single_imgs ` besides `region_signals`, including: `region_ids` and `region_img`


mtorabi59:
Hi, 

I want to add new outputs to `transform_single_imgs` as we discussed, but this causes some tests to fail because the function is now returning more than one output. What is the best way to handle this?

mtorabi59:
Also, I can add `region_names` as another output too, which is different from `region_ids`. Or it can now be obtained by `labels[region_ids[i]]` for the signal in `region_signals[:, i]`.

mtorabi59:
> @mtorabi59 it's a breaking change so we would have to go with a deprecation. However, I need to brush up a bit on the issue. I'm wondering if it would make sense to instead add the new returns as attributes to the `NiftiLabelsMasker` class.

I think that makes sense. because they are not depending on `fmri_img` given to `transform_single_imgs`. I can do that👍

bthirion:
Actually, we should strive to have always the same outputs for a given function. Using a dictionary is a bit cheating and will make it harder to find bugs I'm afraid. Isn't there a way to change systematically the output of this function ?

mtorabi59:
> Actually, we should strive to have always the same outputs for a given function. Using a dictionary is a bit cheating and will make it harder to find bugs I'm afraid. Isn't there a way to change systematically the output of this function ?

Yes, I agree. Do you mean the output of `img_to_signals_labels` or  `_filter_and_extract`? could you clarify what you mean by systematically changing it?

bthirion:
I just mean that this function is called from many different places in the codebase. So the change may be quite important.

mtorabi59:
So I added a optional argument called `return_masked_atlas` to `img_to_signals_labels`. If True, it will also return `masked_atlas` or `masked_labels_img`. Then, this will be used to set the new attribute of `NiftiLabelsMasker` called `region_atlas_`.

mtorabi59:
For this error from documentation builder:
ERROR: Unknown target name: "region_ids".
Do I need to add the new attributes somewhere for documentation?

Remi-Gau:
looks like sphinx is trying to build a hyperlink

not sure why it is happening there and not for other places in the code base where we have this

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -112,6 +114,30 @@ class NiftiLabelsMasker(BaseMasker, _utils.CacheMixin):
 
         .. versionadded:: 0.9.2
 
+    region_ids_ : dict
+        A dictionary containing the region ids corresponding to each
+        column in region_signal. The region id corresponding to
+        region_signal[:,i] is region_ids_[i].
+        region_ids_['background'] is the background label.
+
+        .. versionadded:: 0.10.2.dev
```
emdupre, line 123: I'm wondering (though I can't check locally with due to the MKL doc dependency) if this version string is related to the Sphinx error we're seeing. I can only find one other instance of a `.dev` version string in the code base, though this is for the `deprecated` admonition and [in that case](https://github.com/nilearn/nilearn/blob/48aecfacd9f0411c50125658966d5a61f40ec2e3/nilearn/_utils/docs.py#L502) there's an included description afterwards.

Remi-Gau:
> For this error from documentation builder: ERROR: Unknown target name: "region_ids". Do I need to add the new attributes somewhere for documentation?

FIXED by this commit.

https://github.com/nilearn/nilearn/pull/3761/commits/c9110193c3d0d14c69eb4a309fddf51e203160d8

The name of the attribute with a trailing underscore was not the problem.
The issue came from having attribute names in the following description wihtout backticks around it.

```rst
    this_is_ok_ : Niimg-like object
        Something like this will not be ok --> this_is_ok_
        But this is is fine --> ``this_is_ok_``
```

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -604,7 +634,24 @@ def transform_single_imgs(self, imgs, confounds=None, sample_mask=None):
...
+        for i in range(region_signals.shape[1]):
+            # ids does not include background label
+            region_ids[i] = ids[i]
+
+        if self.labels is not None:
+            self.region_names_ = {
+                key: self.labels[region_id]
+                for key, region_id in region_ids.items()
+                if region_id != self.background_label
+            }
```
ymzayek, line 650: can you add a test for this giving some made up label names?

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -604,7 +634,24 @@ def transform_single_imgs(self, imgs, confounds=None, sample_mask=None):
...
         )
 
-        self.labels_ = labels_
+        self.labels_ = ids
+
+        # defining a dictionary containing regions ids
+        region_ids = {'background': self.background_label}
+        for i in range(region_signals.shape[1]):
+            # ids does not include background label
+            region_ids[i] = ids[i]
```
ymzayek, line 643: It's a bit confusing having "background" key map to a 0 label and then a 0 key map to a 1 label. Why have this and region names dictionary? I guess in the end what's useful to the user is one dictionary mapping masked labels to their names (or arbitrary names matching the label if self.labels is None)

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -22,12 +22,14 @@ def __init__(self, _resampled_labels_img_, background_label, strategy,
...
+        from ..regions.signal_extraction import img_to_signals_labels
 
-        return signal_extraction.img_to_signals_labels(
+        signals, labels, masked_labels_img = img_to_signals_labels(
             imgs, self._resampled_labels_img_,
             background_label=self.background_label, strategy=self.strategy,
-            keep_masked_labels=self.keep_masked_labels, mask_img=self.mask_img)
+            keep_masked_labels=self.keep_masked_labels, mask_img=self.mask_img,
+            return_masked_atlas=True)
+        return signals, (labels, masked_labels_img)
```
bthirion, line 32: Thx for handling that !

This is a breaking change of the API... we need to go through deprecation cycles.

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -386,6 +402,22 @@ def test_signals_extraction_with_labels_with_mask(
...
+        imgs=data_img, labels_img=labels_img, mask_img=mask_img,
+        return_masked_atlas=True
+    )
+
+    labels_data_r = get_data(masked_atlas_r)
+
+    # labels should be masked and only contain 3 regions
+    # and the background
+    assert list(np.unique(labels_data_r)) == [0, 1, 2, 5]
+
```
Remi-Gau, line 421: extract into a separate test

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -340,6 +340,22 @@ def test_signals_extraction_with_labels_without_mask(
         assert_almost_equal(signals_r, signals)
         assert labels_r == list(range(1, 9))
 
+    # test return_masked_atlas
```
Remi-Gau, line 343: extract into a separate test

nilearn/maskers/nifti_labels_masker.py
mtorabi59: Is it because of the return? Could you explain a bit more? I thought it should be fine.

bthirion: If I'm not mitsaken, we move from 2 to 3 return arguments.

nilearn/regions/signal_extraction.py
```diff
@@ -334,6 +342,12 @@ def img_to_signals_labels(
         labels_index = {l: n for n, l in enumerate(labels)}
         for this_label in missing_labels:
             signals[:, labels_index[this_label]] = 0
+
+    # finding the new labels image
+    masked_atlas = Nifti1Image(labels_data.astype(np.int8), labels_img.affine)
+
+    if return_masked_atlas:
+        return signals, labels, masked_atlas
```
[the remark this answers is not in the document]
mtorabi59: No this is something that won't break the previous codes. it is more like an additional argument so we can return an additional output (`masked_atlas`) without breaking the previous codes. I tried to make the change in a way that doesn't need deprecation (with help of @jeromedockes). We also tried to do it in a way (using parentheses) that does not require changes to [_filter_and_extract](https://github.com/nilearn/nilearn/blob/65783298e962542a0f0653460803f6e67335547e/nilearn/maskers/base_masker.py#L19)

nilearn/maskers/nifti_labels_masker.py
mtorabi59: I put them in parentheses so I think it will still count as two arguments. It was @jeromedockes idea. and it didn't fail any tests. Maybe I'm missing something?

nilearn/maskers/nifti_labels_masker.py
mtorabi59: The `region_ids_` is indeed needed and it was also mentioned in [the original issue](https://github.com/nilearn/nilearn/issues/3085). The main reason is the empty regions. So in brief, the signal in the ith column does not necessarily correspond to the ith region. We could align `region_names_` and `region_signals` so that the ith signal corresponds to the ith name, but another problem is that the `labels_img` or `region_atlas_` that we are going to return is filled with the original ids which do not correspond to the new order of ids. We could update the ids in `region_atlas_` too, but besides all these @jeromedockes thinks it's a good idea that we have map to the original ids for users reference or some correspondence between the new and original ids, and I agree. What do you think?
But I can remove background key if it is confusing. I just thought it's good to have it.

nilearn/regions/signal_extraction.py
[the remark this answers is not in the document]
bthirion: If we can avoid conditional arguments, that's better in terms of bug tracking and code maintenance.

nilearn/maskers/nifti_labels_masker.py
bthirion: But it can break code from third parties.  
To be honest, I don't have a better suggestion, but I'm wondering whether we should not use explicit deprecation cycles.

nilearn/regions/signal_extraction.py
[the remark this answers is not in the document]
ymzayek: It would be better to deprecate the old behavior. I don't see a good reason we'd want to keep both behaviors. Just like we added `keep_masked_labels` to deprecate old behavior, we can also do this with `return_masked_atlas`  in the same deprecation cycle.

Another option would be to extract the added functionality into a separate new function in `signal_extraction.py` like:

```python
def _get_masked_data(imgs, labels_img, mask_img, background_label=0, keep_masked_labels=False):
    """Get the masked image."""
    labels_img = _utils.check_niimg_3d(labels_img)

    imgs = _utils.check_niimg_4d(imgs)
    _, labels_data = _get_labels_data(
        imgs,
        labels_img,
        mask_img,
        background_label,
        keep_masked_labels=keep_masked_labels,
    )

    return Nifti1Image(labels_data.astype(np.int8), labels_img.affine)
```

and then from `nifti_labels_maskers.NiftiLabelsMasker.transform_single_imgs` maybe under the call for `_filter_andextract` call:

```python
        self.region_atlas_ = _get_masked_data(
            imgs,
            self._resampled_labels_img_,
            self._resampled_mask_img,
            background_label=self.background_label
        )
```

In this case, there would be no modification to labels masker `_ExtractionFunctor` or to `img_to_signals_labels`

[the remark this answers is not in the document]
mtorabi59: @ymzayek yes! another idea is sth like that. Me and @jeromedockes had another idea which was defining a new function like `_img_to_signals_labels_with_masked_atlas` and then call this new function in `img_to_signals_labels` but don't return the `masked_atlas` output. This way we don't need deprecation right?
Which one do you think makes more sense? 1) a new argument and new output or 2) a new function?

              def img_to_signals_labels(
                  imgs,
                  labels_img,
                  mask_img=None,
                  background_label=0,
                  order="F",
                  strategy="mean",
                  keep_masked_labels=True,
              ):
              
                  signals, labels, _ = _img_to_signals_labels_with_masked_atlas(...)
                  return signals, labels

nilearn/maskers/nifti_labels_masker.py
mtorabi59: @bthirion I see. You mean people who use this private function assume that the second output is just `labels` and that will cause errors?

bthirion: Yes, exactly.

nilearn/regions/signal_extraction.py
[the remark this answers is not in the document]
ymzayek: @mtorabi59 Yes that works too, it's along the lines of what I was thinking so let's go with your implementation of adding a new function. Thx !

ymzayek:
After discussion at the drop in hour yesterday with @mtorabi59 and @bthirion we decided it's best to deprecate the function `img_to_signals_labels`'s returns so we can eventually have one function that returns all three outputs (signals, labels, masked_atlas) as there is no reason to have both functions. Also the output of `_ExtractionFunctor` can be changed without deprecation given that it is private. @mtorabi59 feel free to ping us as you conitnue working on this.

mtorabi59:
Is the documentation error from my part? I don't find the error message referring to the line that is causing this:/

doc/changes/latest.rst
```diff
@@ -23,3 +23,5 @@ Enhancements
 
 Changes
 -------
+
+- :bdg-danger:`Deprecation` ``img_to_signals_labels`` in :class:`~regions.signal_extraction` will also return ``masked_atlas`` in release 0.15. Meanwhile, use ``return_masked_atlas`` parameter to enable/disable this behavior. (:gh:`3761` by `Mohammad Torabi`_).
```
ymzayek, line 27:
```suggestion
- :bdg-danger:`Deprecation` :func:`~regions.img_to_signals_labels` will also return ``masked_atlas`` in release 0.15. Meanwhile, use ``return_masked_atlas`` parameter to enable/disable this behavior. (:gh:`3761` by `Mohammad Torabi`_).

Related to the doc build failure. Hard to see in the logs if you're not used to looking at them
```

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -22,12 +22,18 @@ def __init__(self, _resampled_labels_img_, background_label, strategy,
...
+        from ..regions.signal_extraction import (
+            _img_to_signals_labels_with_masked_atlas,
+        )
 
-        return signal_extraction.img_to_signals_labels(
-            imgs, self._resampled_labels_img_,
-            background_label=self.background_label, strategy=self.strategy,
-            keep_masked_labels=self.keep_masked_labels, mask_img=self.mask_img)
+        signals, labels, masked_labels_img =\
+            _img_to_signals_labels_with_masked_atlas(
```
ymzayek, line 30: So now we can go back to using img_to_signal_labels and setting the new parameter to True to return the masked atlas right?

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -22,12 +22,18 @@ def __init__(self, _resampled_labels_img_, background_label, strategy,
...
+                imgs, self._resampled_labels_img_,
+                background_label=self.background_label, strategy=self.strategy,
+                keep_masked_labels=self.keep_masked_labels,
+                mask_img=self.mask_img,
+            )
+        return signals, (labels, masked_labels_img)
```
ymzayek, line 36: And here since this is a private function we can just add the new return without making part of a tuple

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -590,6 +590,40 @@ def test_nifti_labels_masker_with_mask():
...
+    affine = np.eye(4)
+    fmri_img, mask_img = data_gen.generate_random_img(shape, affine=affine)
+    labels_img = data_gen.generate_labeled_regions(
+        shape[:3],
+        affine=affine,
+        n_regions=7,
+    )
+
+    # define region_names
+    region_names = ['background'] + ["region_" + str(i + 1) for i in range(7)]
```
ymzayek, line 611:
```suggestion
    region_names = ["background"] + ["region_" + str(i + 1) for i in range(7)]
```

black

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -590,6 +590,40 @@ def test_nifti_labels_masker_with_mask():
...
+
+    masker = NiftiLabelsMasker(
+        labels_img,
+        labels=region_names,
+        resampling_target=None,
+    )
+    _ = masker.fit().transform(fmri_img)
+
+    region_names_after_fit =\
+        [masker.region_names_[i] for i in masker.region_names_]
```
ymzayek, line 621:
```suggestion
    region_names_after_fit = [
        masker.region_names_[i] for i in masker.region_names_
    ]
```
black

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -590,6 +590,40 @@ def test_nifti_labels_masker_with_mask():
...
+    region_names_after_fit.sort()
+    region_names.sort()
+    region_names.pop(region_names.index('background'))
```
ymzayek, line 624:
```suggestion
    region_names.pop(region_names.index("background"))
```

nilearn/regions/signal_extraction.py
```diff
@@ -334,7 +338,107 @@ def img_to_signals_labels(
         labels_index = {l: n for n, l in enumerate(labels)}
         for this_label in missing_labels:
             signals[:, labels_index[this_label]] = 0
-    return signals, labels
+
+    # finding the new labels image
+    masked_atlas = Nifti1Image(
+        labels_data.astype(np.int8), labels_img.affine
+    )
```
ymzayek, line 345:
```suggestion
    masked_atlas = Nifti1Image(labels_data.astype(np.int8), labels_img.affine)
```

nilearn/regions/signal_extraction.py
```diff
@@ -334,7 +338,107 @@ def img_to_signals_labels(
...
+@_utils.fill_doc
+def img_to_signals_labels(
+    imgs,
+    labels_img,
+    mask_img=None,
+    background_label=0,
+    order="F",
+    strategy="mean",
+    keep_masked_labels=True,
+    return_masked_atlas=False
```
ymzayek, line 359:
```suggestion
    return_masked_atlas=False,
```

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -365,6 +366,31 @@ def test_signals_extraction_with_labels_without_mask(
...
+    """Test if the returned masked_atlas is correct in \
+    conversion between signals and images \
+    using regions defined by labels."""
+    data_img = signals_to_img_labels(signals=signals, labels_img=labels_img)
+
+    # test return_masked_atlas
+    signals_r, labels_r, masked_atlas_r =\
+        _img_to_signals_labels_with_masked_atlas(
+            imgs=data_img, labels_img=labels_img,
+        )
```
ymzayek, line 381:
```suggestion
    (
        signals_r,
        labels_r,
        masked_atlas_r,
    ) = _img_to_signals_labels_with_masked_atlas(
        imgs=data_img,
        labels_img=labels_img,
    )
```

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -411,6 +437,33 @@ def test_signals_extraction_with_labels_with_mask(
...
+
+    # test return_masked_atlas
+    # create a mask_img with only 3 regions
+    mask_img = _create_mask_with_3_regions_from_labels_data(
+        get_data(labels_img), labels_img.affine)
+
+    signals_r, labels_r, masked_atlas_r =\
+        _img_to_signals_labels_with_masked_atlas(
+            imgs=data_img, labels_img=labels_img, mask_img=mask_img,
+        )
```
ymzayek, line 458:
```suggestion
        get_data(labels_img), labels_img.affine
    )

    (
        signals_r,
        labels_r,
        masked_atlas_r,
    ) = _img_to_signals_labels_with_masked_atlas(
        imgs=data_img,
        labels_img=labels_img,
        mask_img=mask_img,
    )
```

nilearn/maskers/nifti_labels_masker.py
mtorabi59: @ymzayek no, the tuple is because `_filter_and_extract` gets only two inputs and we didn't want to change that.

nilearn/maskers/nifti_labels_masker.py
mtorabi59: You mean I should remove `_img_to_signals_labels_with_masked_atlas` function and replace it with `img_to_signals_labels` with `return_masked_atlas=True`?

ymzayek: Well since we decided to go with deprecating the return behavior of `img_to_signals_labels` I'm not sure we even need `_img_to_signals_labels_with_masked_atlas` anymore since it was originally written to avoid deprecation. I would rather get rid of it and revert `img_to_signals_labels` to what it was except for the added deprecation code. And then yes, call `img_to_signals_labels` with `return_masked_atlas=True`

nilearn/maskers/nifti_labels_masker.py
ymzayek: Ok and not wanting to change `_filter_and_extract` is because it is called by other maskers not dealt with in this PR? I think we can leave it like this for this PR but just noting that it seems reasonable to me to change its return behavior if other maskers will be changed to support returning a masked image as well.

mtorabi59: @ymzayek yes we have to change it's behavior in future

nilearn/regions/signal_extraction.py
```diff
@@ -270,18 +272,23 @@ def img_to_signals_labels(
...
+    order : :obj:`str`, optional
+        Ordering of output array ("C" or "F"). Default="F".
 
-    strategy : :obj:`str`, default="mean"
+    strategy : :obj:`str`, optional
         The name of a valid function to reduce the region with.
         Must be one of: sum, mean, median, minimum, maximum, variance,
-        standard_deviation.
+        standard_deviation. Default="mean".
     %(keep_masked_labels)s
```
ymzayek, line 285: Here we shouldn't see a diff. The default values were moved up to the type definition in another PR so maybe this is a result of incorrect merge conflict resolution

nilearn/regions/signal_extraction.py
```diff
@@ -270,18 +272,23 @@ def img_to_signals_labels(
...
 
+    return_masked_atlas : :obj:`bool`, optional
+        If True, the masked atlas is returned. Default=False.
```
ymzayek, line 288:
```suggestion
    return_masked_atlas : :obj:`bool`, default=False
        If True, the masked atlas is returned.
```

ymzayek (commented):
@mtorabi59 thanks for the updates! I want to give it another full review but seems like it's good to go

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -545,6 +609,22 @@ def test_img_to_signals_labels_warnings(labeled_regions, fmri_img):
...
 
+    # test return_masked_atlas deprecation warning
+    with pytest.warns(
+        DeprecationWarning,
+        match='After version 0.13. "img_to_signals_labels" will also return '
+        'the "masked_atlas". Meanwhile "return_masked_atlas" parameter can be '
+        "used to toggle this behavior. In version 0.15, "
+        '"return_masked_atlas" parameter will be removed.',
+    ):
+        labels_signals, labels_labels = img_to_signals_labels(
```
ymzayek, line 620:
```suggestion
        img_to_signals_labels(
```

ymzayek (commented):
Minor stuff

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -590,6 +590,41 @@ def test_nifti_labels_masker_with_mask():
...
+    # masked_labels is a 4D image with shape (13,11,12,1)
+    masked_labels_data = get_data(masked_labels)[:, :, :, 0]
+    assert np.allclose(get_data(masker.region_atlas_), masked_labels_data)
+
+
+def test_region_names():
+    """Test region_names_ attribute in NiftiLabelsMasker"""
+    shape = (13, 11, 12, 3)
+    affine = np.eye(4)
+    fmri_img, mask_img = data_gen.generate_random_img(shape, affine=affine)
```
ymzayek, line 603:
```suggestion
    fmri_img, _ = data_gen.generate_random_img(shape, affine=affine)
```

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -590,6 +590,41 @@ def test_nifti_labels_masker_with_mask():
...
+    _ = masker.fit().transform(fmri_img)
+
+    region_names_after_fit = [
+        masker.region_names_[i] for i in masker.region_names_
+    ]
+    region_names_after_fit.sort()
+    region_names.sort()
+    region_names.pop(region_names.index("background"))
+    assert region_names_after_fit == region_names
+
```
ymzayek, line 627: @mtorabi59 what do you think of also adding this test:

```python
def test_missing_labels_due_to_resampling(affine_eye):
    """Test region_names_ matches signals after resampling drops labels."""
    shape = (13, 11, 12, 3)
    affine_low = np.diag((4, 4, 4, 4))
    fmri_img, _ = data_gen.generate_random_img(shape, affine=affine_low)
    labels_img = data_gen.generate_labeled_regions(
        shape[:3],
        affine=affine_eye,
        n_regions=7,
    )

    # define region_names
    region_names = ["background"] + ["region_" + str(i + 1) for i in range(7)]

    masker = NiftiLabelsMasker(
        labels_img,
        labels=region_names,
        resampling_target="data",
    )
    signals = masker.fit().transform(fmri_img)

    assert len(masker.region_names_) == signals.shape[1]
    assert len(list(masker.region_ids_.items())[1:]) == signals.shape[1]
```

ymzayek (commented):
@mtorabi59 can you make the following changes?

nilearn/maskers/tests/test_nifti_labels_masker.py
bthirion: +1

bthirion (commented):
The PR LGTM. @ymzayek proposed an additional test thatw e could add.

nilearn/maskers/tests/test_nifti_labels_masker.py
mtorabi59: @ymzayek  that's a good idea!

bthirion (approved):
LGTM, thx.

ymzayek:
Merging 🚀

Remi-Gau:
@mtorabi59 
I need to check but I think that some files that were on this PR were not "blackified" yet. 
So before you get back to work on the maskers some more I will probably lint those files.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -111,6 +116,33 @@ class NiftiLabelsMasker(BaseMasker, _utils.CacheMixin):
...
+        .. versionadded:: 0.11.0.dev
+
+    region_atlas_ : Niimg-like object
+        Regions definition as labels.
+        The labels correspond to the indices in ``region_ids_``.
+        The region in ``region_atlas_`` that takes the value ``region_ids_[i]``
+        is used to compute the signal in ``region_signal[:,i]``.
+
+        .. versionadded:: 0.11.0.dev
+
```
Remi-Gau, line 145: @mtorabi59 
Random question on this PR: why was it chosen for those attributes to have a trailing underscore?

Why couldn't we have:
- region_names
- region_atlas
- region_ids

bthirion: I think it means that these quantities are estimated, i.e. fit-/data-dependent.

Remi-Gau: ha. I guess we should probably mention that somewhere in our style guide, because it is clearly not obvious.

Remi-Gau: I am fine just pointing to the scikit learn doc: https://scikit-learn.org/stable/developers/develop.html#estimated-attributes

PR #3762, opened 2023-06-16
[INFRA] Add trigger to build specific examples on a push in pull requests

### LINKED ISSUE #3612: Adding an option to easily build specific examples in a pull request

ymzayek:
Closes #3612 .

TODO:
- [x] Test new option
- [x] Document in github workflows README.md

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -38,6 +38,15 @@ You can request a full build from a Pull Request at any time by including the ta
 $ git commit -m "[full doc] request full build"
 ```
 
+Though partial build will build modified examples, sometimes code changes on the module side could affect the plots in unmodified examples.
+For this, you can request for the CI to build a specific example by using the tag "[example]" and the name of the example. This is useful when wanting to get quick feedback from reviewers.
+
+```bash
+$ git commit -m "[example] plot_nilearn_101.py"
+```
+
+However for quick checks to do yourself you should always opt for local builds following the instructions here: [building-documentation](https://nilearn.github.io/stable/development.html#building-documentation).
+
 #### Dataset caching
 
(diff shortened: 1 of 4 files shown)


bthirion (approved):
Hard to check if this one in doing its job...

Remi-Gau (approved):
LGTM

if we end up making heavy use of this we could think of also adding the possibilty to trigger the workflow from the github GUI
by using dispatch.

https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

ymzayek:
@Remi-Gau yes that would be a good idea. I don't think this should be used much but rather a nice-to-have but that remains to be seen.

PR #3781, opened 2023-06-27
[ENH] Add `LassoCV` to `nilearn.decoding.DecoderRegressor`

### LINKED ISSUE #3769: Add `LassoCV` as an additional estimator for `Decoder` objects
in a future PR it might still be worth adding the `LassoCV` as a possible estimator

_Originally posted by @jeromedockes in https://github.com/nilearn/nilearn/issues/3736#issuecomment-1561496648_

Looking at the [LassoCV documentation](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LassoCV.html), this would be an additional regression estimator (currently we have ridge regression and SVR). So we could add something like `lasso=LassoCV()` to [`SUPPORTED_ESTIMATORS`](https://github.com/nilearn/nilearn/blob/main/nilearn/decoding/decoder.py#L53). 
Maybe also with `lasso_regressor=LassoCV()` to mirror `ridge`/`ridge_regressor`?

michellewang:
Closes #3769.

- Added `'lasso'` and `'lasso_regressor'` as supported estimators (`sklearn.linear_model.LassoCV()`)
  - For the default hyperparameter values, I used `n_alphas=100` instead of directly specifying a list of `alphas`. This is the default for `LassoCV` and makes it so that the alphas depend on the data values, so I think this is more desirable than fixing `alphas` to always be the same, but let me know if you have other thoughts.

I noticed that even though we have `Decoder` for classification and `DecoderRegressor` for regression, there is no check for whether the user provides an appropriate estimator (as far as I can tell). As a result, it is technically possible to pass something like `'svc'` to `DecoderRegressor`, which doesn't make sense since (for example) `DecoderRegressor` uses `r2` scoring.


nilearn/decoding/decoder.py
```diff
@@ -186,6 +198,11 @@ def _default_param_grid(estimator, X, y):
         # so for L2 penalty, param_grid["Cs"] is either 1e-3, ..., 1e4, and
         # for L1 penalty the values are obtained in a more data-driven way
         param_grid["Cs"] = [np.geomspace(2e-3, 2e4, 8) * min_c]
+    elif isinstance(estimator, LassoCV):
+        # the default is to generate 100 alphas based on the data
+        # (alpha values can also be set with the 'alphas' parameter, in which
+        # case 'n_alphas' is ignored)
+        param_grid["n_alphas"] = [100]
```
bthirion, line 205: 100 is a lot, and Lasso fit is expensive. I'd change the default to 30 ---at most.

bthirion (commented):
Thx, looks good so far.
Maybe we could showcase it in an example ?

jeromedockes:
Thanks a lot @michellewang ! LGTM apart from reducing the number of fits as @bthirion requested.

> For the default hyperparameter values, I used n_alphas=100 instead of directly specifying a list of alphas. 

yes I think that's the best thing to do!

> there is no check for whether the user provides an appropriate estimator

that's a good point but let's deal with it in another pr

ymzayek (commented):
@michellewang LGTM, other than addressing @bthirion 's comment is there anything else left to do on this one?

michellewang:
@ymzayek yes I am trying to think if it would make sense to add something to an example. [This one](https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_grid_search.html) is about cross-validation but it does classification (not regression) and there is a section showing a `param_grid` for the CV. Do you think it's worth adding a new example just to show the Decoder with `LassoCV` regressor? It would be similar to other examples with regressors I think

bthirion (commented):
LGTM. Just wondering: is there any example that works better when Lasso is used instead of Ridge ?

ymzayek:
I think we only showcase regression using `svr` in https://nilearn.github.io/stable/auto_examples/02_decoding/plot_oasis_vbm.html#sphx-glr-auto-examples-02-decoding-plot-oasis-vbm-py. Instead of adding an example I suggest to add information about Lasso in https://nilearn.github.io/stable/decoding/estimator_choice.html#different-linear-models

bthirion:
I simply suggest to try once replacing the SVR with a LassoCV to see the difference in accuracy and computation time. If it not enough or too prohibitive, let's keep the example as-is. Otherwise we can switch here.

michellewang:
Ok, I modified the example to show both the SVR and Lasso results. Let me know if this is what you had in mind

ymzayek:
@michellewang thanks, the example looks good! Can you run it with and without lasso and let us know the difference in computation time? If it's big, I think we should stick to showcasing one of the 2 regressors, and if not we can keep it like this.

Also, can you still add lasso to the list here: https://nilearn.github.io/stable/decoding/estimator_choice.html#different-linear-models

michellewang:
@ymzayek on my laptop it takes 1min 8s without lasso and 1min 30s with Lasso, so adding the Lasso added about 22 seconds. Let me know if that's okay or if we should only keep one of them. The Lasso has higher CV error than the SVR (MAE is ~12 instead of ~10).

ymzayek:
I would say to keep the example as the original (without lasso). However, maybe for users this is useful to see so I will defer to what @bthirion and others think.

bthirion:
If it performs worse and takes more time, we should certainly not include it. Sorry for the useless request.

michellewang:
I reverted the example, no worries @bthirion

jeromedockes (approved):
LGTM, thanks a lot @michellewang !

ymzayek:
Thx merging

PR #3795, opened 2023-06-30
[FMT] apply black to maskers - part 1

Remi-Gau:

- update configs to include nilearn/maskers
- apply black

### DIFF
--- a/nilearn/maskers/tests/test_multi_nifti_masker.py
+++ b/nilearn/maskers/tests/test_multi_nifti_masker.py
@@ -25,7 +25,15 @@ def test_auto_mask():
     masker = MultiNiftiMasker(mask_args=dict(opening=0))
     # Check that if we have not fit the masker we get a intelligible
     # error
-    pytest.raises(ValueError, masker.transform, [[img, ]])
+    pytest.raises(
+        ValueError,
+        masker.transform,
+        [
+            [
+                img,
+            ]
+        ],
+    )
     # Check error return due to bad data format
     pytest.raises(ValueError, masker.fit, img)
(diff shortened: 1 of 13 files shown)


nilearn/maskers/tests/test_multi_nifti_masker.py
```diff
@@ -25,7 +25,15 @@ def test_auto_mask():
...
-    pytest.raises(ValueError, masker.transform, [[img, ]])
+    pytest.raises(
+        ValueError,
+        masker.transform,
+        [
+            [
+                img,
+            ]
+        ],
+    )
```
ymzayek, line 36: Anyway to avoid this?

Also note for future refactoring, these should use context manager

nilearn/maskers/tests/test_multi_nifti_masker.py
```diff
@@ -37,16 +45,21 @@ def test_auto_mask():
...
+    assert_array_equal(get_data(masker.mask_img_), np.logical_or(data, data2))
     # Smoke test the transform
-    masker.transform([[img, ]])
+    masker.transform(
+        [
+            [
+                img,
+            ]
+        ]
+    )
```
ymzayek, line 56: And this?

nilearn/maskers/tests/test_multi_nifti_masker.py
Remi-Gau: removing the coma is should tell black to put that on oneline

let me fix this

Remi-Gau: yup that dit it

Issue #3728, opened 2023-04-27
[DOC] Update and improve examples in nilearn documentation


ymzayek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed suggestion in detail.

Unrelated to the changes to this PR: I noticed some suboptimal outputs in some examples so I will open a separate issue to track them

_Originally posted by @ymzayek in https://github.com/nilearn/nilearn/issues/3720#issuecomment-1521902025_

This is a meta issue concerning nilearn examples and their rendering

#### Plotting (@ymzayek)

- [x] overlapping title
  - Known issue that appears in many examples so will not list here. See PR #2804 
- [x] Cut off titles or image
  - https://nilearn.github.io/dev/auto_examples/01_plotting/plot_demo_more_plotting.html 
  - https://nilearn.github.io/dev/auto_examples/07_advanced/plot_localizer_simple_analysis.html
  - https://nilearn.github.io/dev/auto_examples/07_advanced/plot_localizer_mass_univariate_methods.html
  - https://nilearn.github.io/dev/auto_examples/07_advanced/plot_haxby_mass_univariate.html
- [x] plots output shown on top of cell calling them (not a comprehensive list)
  - https://nilearn.github.io/dev/auto_examples/01_plotting/plot_haxby_masks.html  
  - https://nilearn.github.io/dev/auto_examples/01_plotting/plot_prob_atlas.html
  - https://nilearn.github.io/dev/auto_examples/02_decoding/plot_haxby_stimuli.html
  - https://nilearn.github.io/dev/auto_examples/03_connectivity/plot_simulated_connectome.html
  - https://nilearn.github.io/dev/auto_examples/04_glm_first_level/plot_design_matrix.html
- [x] Fix matrix plots titles

#### Output (@ymzayek)

- [x] print output
  - https://nilearn.github.io/dev/auto_examples/00_tutorials/plot_single_subject_single_run.html (first output block)

#### Other (@Remi-Gau)

- [x] These 2 examples are listed separately in the dropdown but are in fact in one file
  - https://nilearn.github.io/dev/auto_examples/01_plotting/plot_overlay.html#
  - https://nilearn.github.io/dev/auto_examples/01_plotting/plot_overlay.html#visualizing-a-probabilistic-atlas-with-plot-prob-atlas
- [x] License or author names render in first code block
  - https://nilearn.github.io/dev/auto_examples/02_decoding/plot_oasis_vbm.html
[...]

PR #3797, opened 2023-06-30
[FIX] Fix long titles overlapping plot contents

### LINKED ISSUE #2054: Plot title overlays the image

ymzayek:
Supersedes and closes #2804 
Relates to #3728 
Closes #2054 .

- Increasing default figsize in different slicers to prevent long titles from overlapping plot contents
-


bthirion:
I'm wondering whether it's a good idea to change default size: Wouldn't it be better to work with fontsize or figsize (and do so in some examples to teach this) ?
(This is a naive question, to be discussed)

ymzayek:
> I'm wondering whether it's a good idea to change default size: Wouldn't it be better to work with fontsize or figsize (and do so in some examples to teach this) ? (This is a naive question, to be discussed)

Yea it would work to just set figsize in a figure object in the examples. But this way at least guarantees that if you add a title it will fit correctly and without it doesn't look worse (but maybe I missed some cases where it does?). It is always possible to override the figsize in any case. TBH I have no strong opinion one way or the other. Any thoughts @jeromedockes ?

ymzayek:
On second thought, I just went through all the examples and there were a lot of them that were affected by this and they look much better. Going through to manually change each one in the examples code is much more of a hassle. At least this way we have a default that looks nice with or without a title

jeromedockes:
I think the proposal in #2804 was to also move the y position of the title, besides increasing the figsize, is this not needed anymore?

nilearn/plotting/displays/_slicers.py
```diff
@@ -1816,7 +1818,7 @@ class MosaicSlicer(BaseSlicer):
 
     _cut_displayed = "yxz"
     _axes_class = CutAxes
-    _default_figsize = [11.1, 7.2]
+    _default_figsize = [11.1, 20.0]
```
jeromedockes, line 1821: this seems like a large increase, why is it necessary?

ymzayek:
> I think the proposal in #2804 was to also move the y position of the title, besides increasing the figsize, is this not needed anymore?

No IIUC that was abandoned for the default figsize approach because increasing the y value will lead to the title getting cut out of the figure, https://github.com/nilearn/nilearn/pull/2804#issuecomment-833815747. I've tested it and it's indeed the case. It seems to me those default x and y position values were chosen to keep the title box "snapped" to the top left corner of the figure.

nilearn/plotting/displays/_slicers.py
ymzayek: > this seems like a large increase, why is it necessary?

It was hard to get the title not to overlap and then to me the rendering in the examples seemed better since the brains are bigger and closer together so less white space. See the mosaic plots in this example ([old](https://nilearn.github.io/stable/auto_examples/01_plotting/plot_demo_more_plotting.html#sphx-glr-auto-examples-01-plotting-plot-demo-more-plotting-py) vs [new](https://output.circle-artifacts.com/output/job/cde255e7-0a62-4aff-8607-6a8301a138ec/artifacts/0/dev/auto_examples/01_plotting/plot_demo_more_plotting.html#sphx-glr-auto-examples-01-plotting-plot-demo-more-plotting-py)).

bthirion (approved):
LGTM, thx.

jeromedockes (approved):
thanks @ymzayek !

PR #3799, opened 2023-07-03
First draft of surface stuff after discussion during meetings & drop-in hours

jeromedockes:
as discussed with @Remi-Gau and @ymzayek 
this adds the surface image and maskers module in a nilearn.experimental module

for now this is just a draft of the basic functionality, some features are missing as well as tests and an example
also it doesn't pass the pre-commit checks (missing docstrings)

the file in tests/example.py is the notebook that was used for discussion and is also in nilearn_sandbox and it should run without issues


nilearn/experimental/surface/_surface_image.py
```diff
@@ -0,0 +1,84 @@
...
+class SurfaceImage:
+    data: PolyData
+    mesh: PolyMesh
+    shape: tuple[int, ...] = dataclasses.field(init=False)
+
+    def __post_init__(self) -> None:
+        data_shapes = {k: v.shape for (k, v) in self.data.items()}
+        vertex_counts = _get_vertex_counts(self.mesh)
+        assert {k: v[-1] for (k, v) in data_shapes.items()} == vertex_counts
+        total_n_vertices = sum(s[-1] for s in data_shapes.values())
```
ymzayek, line 75: Maybe a bit easier to follow:

```suggestion
        total_n_vertices = sum(vertex_counts.values())
```

nilearn/experimental/surface/_surface_image.py
```diff
@@ -0,0 +1,84 @@
...
+    @property
+    def coordinates(self) -> np.ndarray:
+        return _io.read_mesh(self._file)["coordinates"]
+
+    @property
+    def faces(self) -> np.ndarray:
+        return _io.read_mesh(self._file)["faces"]
+
+
+class InMemoryMesh(Mesh):
```
ymzayek, line 38: Why not just expand the type of input for class `Mesh` since `load_surf_mesh` that's called in `read_mesh` takes in loaded arrays

ymzayek (commented):
First run through the core API and it looks good so far. Next I'll go through the masker and plotting APIs.

ymzayek (commented):
Looking good so far!

Remi-Gau:
@jeromedockes 
I merged main into your branch to make those failing tests disappear (hopefully)

nilearn/experimental/surface/_maskers.py
```diff
@@ -0,0 +1,136 @@
...
+class SurfaceMasker:
+    # TODO:
+    # - allow passing an actual roi mask
+    # - don't store a full array of 1 to mean "no masking"
+    # - 1d output for 1d images
+
+    mask_img_: SurfaceImage | None
+    output_dimension_: int | None
+
+    def _fit_mask_img(self, img: SurfaceImage) -> None:
```
bthirion, line 19: I don't understand what to expect from this method. A short docstring could help.

nilearn/experimental/surface/_surface_image.py
```diff
@@ -0,0 +1,84 @@
...
+    n_vertices: int
+
+    _coordinates: np.ndarray
+    _faces: np.ndarray
+
+    def __init__(self, coordinates: np.ndarray, faces: np.ndarray) -> None:
+        self._coordinates = coordinates
+        self._faces = faces
+        self.n_vertices = coordinates.shape[0]
+
```
bthirion, line 48: Do we need some checks that faces and coordinates are consistent ?

bthirion (commented):
The API looks great. In need to take more time to look at the internals of this.

Remi-Gau:
> @jeromedockes I merged main into your branch to make those failing tests disappear (hopefully)

tests are still failing

@ymzayek and I suppose this is because of the monkeypatch in the example

this leads to a more generic question - that we don't have to answer right now but it would be good to get to it at some point: where do we keep examples for experimental features?

quick look at scikit-learn shows that there is doc for the experimental features but not necessarily examples.

https://scikit-learn.org/stable/metadata_routing.html#metadata-routing

nilearn/experimental/surface/_datasets.py
```diff
@@ -0,0 +1,61 @@
...
+from nilearn import datasets
+from nilearn.experimental.surface import _io
+from nilearn.experimental.surface._surface_image import (
+    Mesh,
+    PolyMesh,
+    SurfaceImage,
+)
+
+
+def load_fsaverage(mesh_name: str = "fsaverage5") -> Dict[str, PolyMesh]:
```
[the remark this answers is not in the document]
jeromedockes: right, I remember we thought about that. but I wonder if it will result in user having to call that function several times, as it is quite frequent to need 2 or 3 of them -- eg pial and white matter for volume to surface projection, then inflated for plotting? but at the same time when only 1 is needed adding the argument would make it faster

nilearn/experimental/surface/_surface_image.py
jeromedockes: for now it uses load_surf_mesh just so we can focus on the surface image and maskers first and deal with the io later. but I think it might simplify things to have a distinction between files on disk and in-memory meshes, to avoid (i) functions like `load_surf_mesh` that have parameters that can be of several completely different types and (ii) a surface image that has several redundant/alternative attributes, some of which can be missing, and always having to deal with those multiple possible configurations

nilearn/experimental/surface/_surface_image.py
```diff
@@ -0,0 +1,84 @@
...
+PolyMesh = Dict[str, Mesh]
+
+
+def _get_vertex_counts(mesh: PolyMesh) -> dict[str, int]:
+    return {part_name: part.n_vertices for (part_name, part) in mesh.items()}
+
+
+@dataclasses.dataclass
+class SurfaceImage:
```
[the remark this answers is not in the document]
jeromedockes: I think we will often need all the data, unlike the meshes which we won't use most of the time, so I'd rather keep things a bit simpler for now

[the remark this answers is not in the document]
jeromedockes: btw we can easily have a SurfaceImage that contains only one hemisphere (the polymesh and polydata would have just 1 element each). that can be done when creating the image or if it's useful we could have methods to extract one part of the image something like `full_img.get_part("left_hemisphere")`

Issue #3800, opened 2023-07-04
doc build failures and new warnings on main


Remi-Gau:
Some new warnings seem related to the new sklearn release.

Note: doc warnings are flagged as error during doc build

See
https://github.com/nilearn/nilearn/actions/runs/5441381451/jobs/9895271361#step:33:35688

Attached log
[33_Build docs.txt](https://github.com/nilearn/nilearn/files/11945928/33_Build.docs.txt)

The culprits seems to be:

```
2023-07-03T08:27:09.5655542Z /usr/share/miniconda3/envs/testenv/lib/python3.9/site-packages/nilearn/connectome/connectivity_matrices.py:docstring of sklearn.utils._metadata_requests._MetadataRequester.get_metadata_routing:11: WARNING: py:class reference target not found: utils.metadata_routing.MetadataRequest
```

```
2023-07-03T08:27:09.5704758Z /usr/share/miniconda3/envs/testenv/lib/python3.9/site-packages/nilearn/connectome/connectivity_matrices.py:docstring of sklearn.utils._metadata_requests.RequestMethod.__get__.<locals>.func:26: WARNING: py:class reference target not found: pipeline.Pipeline
```

Seems related to this: https://scikit-learn.org/stable/whats_new/v1.3.html#experimental-under-development

EDIT: [suggestion to rely on intersphinx](https://github.com/scikit-learn/scikit-learn/pull/26747#issuecomment-1619128349) does not seem to apply as this is already the case for nilearn.

---

Listing some of the other most frequent warnings below but are not causing a failure:

### 1

```
2023-07-03T06:42:25.5886498Z /usr/share/miniconda3/envs/testenv/lib/python3.9/site-packages/sklearn/svm/_classes.py:32: FutureWarning: The default value of `dual` will change from `True` to `'auto'` in 1.5. Set the value of `dual` explicitly to suppress the warning.
```

Possibly related to 

https://scikit-learn.org/stable/whats_new/v1.3.html#sklearn-svm

### 2

```
2023-07-03T07:27:55.7437431Z /usr/share/miniconda3/envs/testenv/lib/python3.9/site-packages/sklearn/linear_model/_base.py:121: FutureWarning:
2023-07-03T07:27:55.7442173Z 
[...]

PR #3804, opened 2023-07-04
[DOC] Workaround for sphinx warning reference target not found for 2 scikit-learn (v1.3) classes

ymzayek:
See #3800. See also https://github.com/scikit-learn/scikit-learn/issues/26761

This warning fails the full doc build

- Add nitpicky option to conf.py
- Add reference targets to ignore
-

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -372,6 +372,15 @@ extlinks = {
     "nipy": ("https://nipy.org/%s", None),
 }
 
+# Check intersphinx reference targets exist
+nitpicky = True
+# Temporary solution to nilearn/nilearn#3800
+# See also scikit-learn/scikit-learn#26761
+nitpick_ignore = [
+    ("py:class", "pipeline.Pipeline"),
+    ("py:class", "utils.metadata_routing.MetadataRequest"),
+]
+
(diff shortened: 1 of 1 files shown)


glemaitre:
You can resolve the problem by using intersphinx to make the link work as done there: https://github.com/microsoft/LightGBM/pull/5956/files

You can see the https://github.com/scikit-learn/scikit-learn/pull/26747#issuecomment-1619128349 for more details.

Remi-Gau:
@glemaitre 

I am pretty sure we are already using intersphinx: 🤷🏾‍♂️

Remi-Gau (approved):

ymzayek:
@glemaitre we already use it and it works correctly as can be seen in our stable docs based on sklearn v1.2 e.g. see link to Pipeline class here: https://nilearn.github.io/stable/modules/generated/nilearn.regions.RegionExtractor.html#nilearn.regions.RegionExtractor.set_params. That's why I suspect it comes from the source code particularly from the file referenced in my comment in https://github.com/scikit-learn/scikit-learn/issues/26761. Maybe just a syntax thing?

bthirion (approved):
LGTM, thx.

PR #3811, opened 2023-07-05
[FIX] improve warnings in building glm from BIDS 

### LINKED ISSUE #3735: `'slice_time_ref' not provided` warning, even when provided
https://github.com/nilearn/nilearn/blob/3edbbb88c61477c6115c2b812c4b3f74345a56cf/nilearn/glm/first_level/first_level.py#L1017-L1023

The message on `line 1017` of `nilearn/glm/first_level/first_level.py`, which is triggered when the `slice_time_ref` cannot be inferred, stipulates that `slice_time_ref` is not provided even when it is, in fact. The rest of the code ends up using the value provided by the user, but the warning is confusing as it looks like this:

```
[...]
UserWarning: No bold.json found in BIDS folder in ../bids_dataset/derivatives/fmriprep.
[...]
UserWarning: 'slice_time_ref' not provided and cannot be inferred from metadata.It will be assumed that the slice timing reference is 0.0 percent of the repetition time. If it is not the case it will need to be set manually in the generated list of models.
[...]
UserWarning: 'slice_time_ref' provided (0.5) is different from the value found in the BIDS dataset (0.0). Note this may lead to the wrong model specification.
```

Remi-Gau:
Closes #3735 

- make sure warnings is not raised when slice_time_ref is provided
- improve warnings readibility

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,6 +13,8 @@ Fixes
 
 - Fix bug in :func:`~glm.first_level.first_level_from_bids` that returned no confound files if the corresponding bold files contained derivatives BIDS entities (:gh:`3742` by `Rémi Gau`_).
 
+- Fix bug in :func:`~glm.first_level.first_level_from_bids` that would throw a warning about ``slice_time_ref`` not being provided even when it was (:gh:`3811` by `Rémi Gau`_).
+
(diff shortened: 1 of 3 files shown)


Remi-Gau:
probably needs an entry in the changelog

ymzayek (approved):
LGTM!

PR #3816, opened 2023-07-05
FIX allow confound files with no cosine

### LINKED ISSUE #3537: Make `cosine` regressors an exception for the variable missing error in `load_confounds`
Right now the behaviour is the function will just fail, and a check is implemented in `_load_single_confounds_file`, so we don't need to write things in each function

https://github.com/nilearn/nilearn/blob/ec57e566455b3d929480fc2ce19e151e16673c6e/nilearn/interfaces/fmriprep/load_confounds.py#L315

The only variable that could be missing when fMRIPrep can run correctly is `cosineXX`, when the timeseries is too short. 

This deserves it's own issue

_Originally posted by @htwangtw in https://github.com/nilearn/nilearn/pull/3285#discussion_r1117252941_

This came up in a discussion with @tsalo

htwangtw:
Closes #3537.

Also some code refactoring
- concentrate most of the missing regressor and missing strategy key word catching to `load_confounds_components`
- consistency of internal and user facing strategy key for ica-aroma (only using ica_aroma, instead of ica_aroma and aroma)
- Add unit test for confound components loading and ensure consistent error catching

To-do:

- [x] The new test `test_load_confounds_components.py` is still very ugly


doc/changes/latest.rst
```diff
@@ -17,6 +17,9 @@ Fixes
 
 - Make the :func:`~nilearn.interfaces.fmriprep.load_confounds` confounds file selection more generic (:gh:`3794` by `Taylor Salo`_).
 
+- Relax the :func:`~nilearn.interfaces.fmriprep.load_confounds` confounds selection on `consine` as not all confound files contained the variables (:gh:`` by `Hao-Ting Wang`_).
```
Remi-Gau, line 20:
```suggestion
- Relax the :func:`~nilearn.interfaces.fmriprep.load_confounds` confounds selection on `cosine` as not all confound files contained the variables (:gh:`` by `Hao-Ting Wang`_).
```

bthirion (approved):
LGTM, thx.

ymzayek (approved):
Thx @htwangtw!

PR #3720, opened 2023-04-21
[MAINT] Update `standardize` deprecated strategy in example gallery

ymzayek:
Closes #3719 .

Changes proposed in this pull request:

- Update `standardize` deprecated strategy in example gallery from default "zscore" (or equivalent `True`) to "zscore_sample"

PR #3821, opened 2023-07-06
[MAINT] Update some missed tests and examples to use new standardize strategy to silence warnings

### LINKED ISSUE #3719: Update examples to use "zscore_sample" `standardize` strategy
At the moment the "zscore" `standardize` strategy deprecation warning is saturating the [doc build logs](https://pipelines.actions.githubusercontent.com/serviceHosts/4e916657-41bc-43a2-a3ed-7be699b3544e/_apis/pipelines/1/runs/32526/signedlogcontent/2?urlExpires=2023-04-21T10%3A15%3A43.4242618Z&urlSigningMethod=HMACV1&urlSignature=bz0LusrMyseqRCOWfiKc2KBhuqdUn2CO4IzZrrkgtqs%3D). The examples should be updated to use the correct "zscore_sample" strategy

ymzayek:
Follow-up to #3720 and Closes #3719 .

-
-


ymzayek:
One issue with silencing all instances of this warning in our examples gallery is that the ones that are still throwing the warning come from the functions `plot_carpet` and `ConnectivityMeasure.fit_transform` which call `signal.clean` with the parameter `standardize` equal to `True` or `"zscore"` internally so these are not exposed to the user. Hence the warning doesn't make sense for these cases as the user has no control and cannot call the new behavior. Probably best to open a new issue to see how to resolve this. Might need to add a deprecated parameter to allow the user to access the new behavior with these functions.

The relevant lines are:
https://github.com/nilearn/nilearn/blob/bce754beea677d4d4a51191e21c64d09002f1429/nilearn/connectome/connectivity_matrices.py#L497
and 
https://github.com/nilearn/nilearn/blob/bce754beea677d4d4a51191e21c64d09002f1429/nilearn/plotting/img_plotting.py#L1430

Remi-Gau (approved):

Remi-Gau:
> One issue with silencing all instances of this warning in our examples gallery is that the ones that are still throwing the warning come from the functions `plot_carpet` and `ConnectivityMeasure.fit_transform` which call `signal.clean` with the parameter `standardize` equal to `True` or `"zscore"` internally so these are not exposed to the user. Hence the warning doesn't make sense for these cases as the user has no control and cannot call the new behavior. Probably best to open a new issue to see how to resolve this. Might need to add a deprecated parameter to allow the user to access the new behavior with these functions.

Sounds good to me and thanks for tracking the source of all those warnings!

nilearn/tests/test_signal.py
```diff
@@ -257,12 +257,6 @@ def test_standardize():
 
     # transpose array to fit _standardize input.
     # Without trend removal
-    b = nisignal._standardize(a, standardize='zscore')
-    stds = np.std(b)
-    np.testing.assert_almost_equal(stds, np.ones(n_features))
-    np.testing.assert_almost_equal(b.sum(axis=0), np.zeros(n_features))
-
```
bthirion, line 264: Why are you removing this ?

nilearn/tests/test_signal.py
```diff
@@ -276,16 +270,11 @@ def test_standardize():
     b = nisignal._standardize(a, detrend=True, standardize="zscore_sample")
     np.testing.assert_almost_equal(b, np.zeros(b.shape))
 
-    length_1_signal = np.atleast_2d(np.linspace(0, 2., n_features))
-    np.testing.assert_array_equal(length_1_signal,
-                                  nisignal._standardize(length_1_signal,
-                                                        standardize='zscore'))
-
-    # Repeating test above but for new correct strategy
```
bthirion, line 284: Same question here.

bthirion (commented):
LGTM overall.

nilearn/tests/test_signal.py
ymzayek: @bthirion it's just repeating exactly another test but with a deprecated value for `standardize`. The deprecated value is still tested here https://github.com/nilearn/nilearn/blob/01cfa3d531efb749e53abe0c57c5adf0bb844ee0/nilearn/tests/test_signal.py#L255-L256 so at least we know nothing is breaking but I don't see a reason to keep testing it with different combinations of parameters

nilearn/tests/test_signal.py
ymzayek: Same explanation as above

bthirion (approved):
LGTM, thx.

PR #3831, opened 2023-07-08
[REF] refactor filename building in tests for interfaces/fmriprep

### LINKED ISSUE #3828: Refactor utils to create confounds from utils/data_gen and interfaces/fmriprep
noticing that there is a bit of code duplication (at least in terms of goals) between some of the utils used in testing to generate dummy fmriprep confounds and some of the code in _utils.data_gen to generate dummy fmriprep datasets.

Nothing big but will probably end up opening a PR to refactor.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/3819#issuecomment-1625115096_

Remi-Gau:
Closes #3828 

- rely on utilities from data gen to generate BIDS valid filenames


nilearn/interfaces/fmriprep/tests/utils.py
```diff
@@ -4,24 +4,59 @@
...
-    "cifti": "_space-fsLR_den-91k_bold.dtseries.nii",
-    "den": "_space-fsLR_den-32k_desc-preproc_bold.nii.gz",
-    "part": (
-        "_part-mag_space-MNI152NLin2009cAsym_"
-        "desc-preproc_bold.nii.gz"
-    ),
+        {"suffix": "bold",
+         "entities": {"space": "MNI152NLin2009cAsym",
+                      "desc": "smoothAROMAnonaggr"},
+         "extension": "nii.gz"},
```
bthirion, line 16: I'm thinking that the extension could be ".nii.gz" by default, following BIDS conventions.

nilearn/interfaces/fmriprep/tests/utils.py
```diff
@@ -4,24 +4,59 @@
...
+                      },
+         "extension": "nii.gz"},
     "gifti": (
-        "_space-fsaverage5_hemi-L_bold.func.gii",
-        "_space-fsaverage5_hemi-R_bold.func.gii",
-    ),
+        {"suffix": "bold",
+         "entities": {"hemi": "L",
+                      "space": "fsaverage5"},
+         "extension": "func.gii"},
```
bthirion, line 55: Is this a valid BIDS extension ?

nilearn/interfaces/fmriprep/tests/utils.py
```diff
@@ -51,53 +86,91 @@ def get_testdata_path(non_steady_state=True):
 def create_tmp_filepath(
     base_path,
     image_type="regular",
-    suffix="sub-test01_task-test",
+    bids_fields=None,
     copy_confounds=False,
     copy_json=False,
     old_derivative_suffix=False
 ):
+    entities_to_include = [
```
bthirion, line 94: `entities_to_include` should be a kw argument ?

nilearn/interfaces/fmriprep/tests/utils.py
```diff
@@ -51,53 +86,91 @@ def get_testdata_path(non_steady_state=True):
...
+            entities_to_include=entities_to_include
+        )
         tmp_img = base_path / tmp_img
         tmp_img.touch()
         tmp_img = str(tmp_img)
     else:
         tmp_img = []
-        for root in img_root:
-            tmp_gii = suffix + root
+        for root in img_file_patterns_type:
```
bthirion, line 146: `img_file_patters_type` is a list ?

bthirion (commented):
Sounds good. Just a few comments on the side.

nilearn/interfaces/fmriprep/tests/utils.py
Remi-Gau: currently in the BIDS extension proposal for functional derivatives, but I think we also have it in there to test things against the type of files found in fmriprep outputs.

https://bids-specification--519.org.readthedocs.build/en/519/derivatives/functional-derivatives.html#functional-derivatives-maps

nilearn/interfaces/fmriprep/tests/utils.py
Remi-Gau: either a string or tuple

nilearn/interfaces/fmriprep/tests/utils.py
Remi-Gau: wait do you mean that it should be `.nii.gz` instead of `nii.gz` like written here?
or that we should set a default at tall?

bthirion: No, sorry, I don't make a difference between .nii.gz and nii.gz. I just mean that this is recommended image format in BIDS.

Remi-Gau: OK then I can see if I can simplify this a little bit.

Remi-Gau: OK nii.gz will be set as default extension and bold as default suffix when creating the filename

nilearn/interfaces/fmriprep/tests/utils.py
Remi-Gau: In this case this function will only ever create derivatives filenames so I do not think that it makes sense to add more flexiblity to it.

if they want more flexiblity then it is better to:

```
from nilearn._utils.data_gen import _create_bids_filename
```

as  `_create_bids_filename` has this `entities_to_include` as kwarg

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
```diff
@@ -18,7 +18,7 @@
 )
 from nilearn.interfaces.fmriprep.tests.utils import (
     create_tmp_filepath,
-    get_leagal_confound,
+    get_legal_confound,
```
htwangtw, line 21: Thanks for fixing this typo...

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
```diff
@@ -391,7 +391,8 @@ def test_not_found_exception(tmp_path):
     # Aggressive ICA-AROMA strategy requires
     # default nifti
     aroma_nii, _ = create_tmp_filepath(
-        tmp_path, image_type="ica_aroma", suffix="aroma"
+        tmp_path, image_type="ica_aroma", bids_fields={"entities":
+                                                       {"sub": "aro-ma"}}
```
htwangtw, line 395: This is not a valid bids participant ID pattern. Why are we using this as a test case? Is it a problem that it's not breaking things?

htwangtw (approved):
LGTM just some minor comments

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
Remi-Gau: Ah. Need to double check. I had to make a change there during testing things. Let me see.

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
htwangtw: I checked locally and no merge conflict happened

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
Remi-Gau: OK my bad this change was not required: I must have misunderstood the reason why this test failed at one point during refactoring

bthirion (approved):
LGTM, thx.

Remi-Gau:
thanks for the reviews @bthirion @htwangtw

PR #3834, opened 2023-07-08
[DOC] remove single occurence from docs.py

### LINKED ISSUE #3832: update defaults in _utils/docs.py
I started commenting the defaults that are not defined in docs.py but maybe a separate PR is needed to resolve these. There are 2 cases: defaults are not defined at all and defaults are set in the docstrings where the standard parameter is referenced

_Originally posted by @ymzayek in https://github.com/nilearn/nilearn/pull/3621#pullrequestreview-1395907407_

apply comments from the reviews above

Remi-Gau:
Closes #3832 

- remove parameters definitions that were put in _utils.docs.py even though they are used only once in the whole code base.

### DIFF
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -84,18 +84,6 @@ bg_on_data : :obj:`bool`, default=False
 
 """
 
-# bg_vmax
-docdict['bg_vmax'] = """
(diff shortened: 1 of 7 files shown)


bthirion (approved):
LGTM too.

PR #3836, opened 2023-07-09
[FMT] apply black

Remi-Gau:

- update configs to format more from nilearn/plotting and to glm/tests/tests_second_level
- apply black


Remi-Gau:
BTW this should be the last PR that is black related.

Some files still need to be reformated but they are involved in other PRs so we will have to see how we tackle them.

So I think that when this is merged we can finally close #2528 after updating the contribution guidelines that was saying we were in the process of reformating the code base.

bthirion (approved):
LGTM, thx !

PR #3860, opened 2023-07-20
[FIX] transparent vertices in `surf_plotting` function

### LINKED ISSUE #3470: image generated by `plot_surf_stat_map` has different aspect when saved in png vs pdf
The aspect of a saved 3d image depends on whether it was saved as pdf or png.
Could be a matplotlib bug.

### Expected behavior
The images should be identical

```python
from nilearn import datasets
from nilearn import surface
from nilearn import plotting

output_file = '/tmp/plop.pdf'
    fsaverage.infl_right, texture, hemi='right',
    threshold=1., bg_map=fsaverage.sulc_right,
    output_file=output_file
)

output_file = '/tmp/plop.png'
    fsaverage.infl_right, texture, hemi='right',
    threshold=1., bg_map=fsaverage.sulc_right,
    output_file=output_file
)
```

mathdugre:
Closes #3470 .

Surface from `surf_plotting` with the matplotlib engine has transparent surfaces.

- Add vertices width
- Apply surface color to the vertices

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -10,7 +10,7 @@ NEW
 
 Fixes
 -----
-- Fix bug in :func:`nilearn.plotting.surf_plotting._plot_surf_matplotlib` that would make vertices transparent when saving in PDF or SVG format (:gh:`3860` by `Mathieu Dugré`_).
+- Fix bug in ``nilearn.plotting.surf_plotting._plot_surf_matplotlib`` that would make vertices transparent when saving in PDF or SVG format (:gh:`3860` by `Mathieu Dugré`_).
 
 - Fix bug that would prevent loading the confounds of a gifti file in actual fmriprep datasets (:gh:`3819` by `Rémi Gau`_).
 
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -10,6 +10,7 @@ NEW
 
 Fixes
 -----
+- Fix bug in :func:`nilearn.plotting.surf_plotting._plot_surf_matplotlib` that would make vertices transparent when saving in PDF or SVG format (:gh:`3860` by `Mathieu Dugré`_).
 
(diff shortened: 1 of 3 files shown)


alishakodibagkar:
Was working on this earlier, and I found a different solution. If we set rasterized=True when plotting the mesh, the mesh is converted into pixels, which removes the gaps between the triangles when saving to a .pdf. This would make the image look more similar to the .png but also slightly decreases the quality of the image.

Images are below

[plop_rasterized.pdf](https://github.com/nilearn/nilearn/files/12114354/plop_rasterized.pdf)
[plop_not_rasterized.pdf](https://github.com/nilearn/nilearn/files/12114355/plop_not_rasterized.pdf)

Is there a preferred solution? @jeromedockes

jeromedockes:
thanks for trying it! in that case I would rather go for the `set_edgecolors` solution rather than `rasterized` so that we can get the correct plot while retaining the advantages of vectorized graphics

doc/changes/latest.rst
```diff
@@ -10,6 +10,7 @@ NEW
 
 Fixes
 -----
+- Fix bug in :func:`nilearn.plotting.surf_plotting._plot_surf_matplotlib` that would make vertices transparent when saving in PDF or SVG format (:gh:`3860` by `Mathieu Dugré`_).
```
ymzayek, line 13: The Sphinx cross-reference syntax won't work here because private functions (name starting with an underline) are not rendered in the docs.

```suggestion
- Fix bug in ``nilearn.plotting.surf_plotting._plot_surf_matplotlib`` that would make vertices transparent when saving in PDF or SVG format (:gh:`3860` by `Mathieu Dugré`_).
```

ymzayek (commented):
Thx @mathdugre! As a first time contributor can you also add your name and relevant information to `CITATION.cff`?

bthirion (approved):
LGTM. Ideally, we would like to add a test, but I simply don't see how to do it.

PR #3872, opened 2023-08-01
[FMT] apply toml formatting hook on pyproject.toml

### LINKED ISSUE #3870: [MAINT] Standardize tabs and spaces in pyproject.toml
Currently, the pyproject.toml has a mix of tabs and spaces, leading to `black` placing unevent alignment here:

https://github.com/nilearn/nilearn/blob/a95149b2162eff9bc4c0b6fe503533311a700f02/pyproject.toml#L58-L75

and here: 

https://github.com/nilearn/nilearn/blob/a95149b2162eff9bc4c0b6fe503533311a700f02/pyproject.toml#L92-L98

[Best-practice](https://peps.python.org/pep-0008/) would be to standardize to spaces. I'm happy to do this, but it also makes a good first issue 😸

Remi-Gau:
Closes #3870 

- adds a pre-commit hook for toml formatting to the pre-commit config
- apply the hook

### Advantages

- toml is formatted

### Drawback

- only if people install the hook (could be solved by adding something in CI)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -96,7 +96,7 @@ repos:
     rev: v2.10.0
     hooks:
     - id: pretty-format-toml
-      args: [--autofix, --indent, '2']
+      args: [--autofix, --indent, '4']
 
   - repo: https://github.com/pyCQA/flake8
     rev: 6.1.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -92,6 +92,12 @@ repos:
         args: ["--toml", "pyproject.toml"]
         additional_dependencies: [tomli]
 
+  - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
+    rev: v2.10.0
+    hooks:
+    - id: pretty-format-toml
+      args: [--autofix, --indent, '2']
+
   - repo: https://github.com/pyCQA/flake8
     rev: 6.1.0
     hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,27 +1,9 @@
 [build-system]
-requires = ["hatchling", "hatch-vcs"]
 build-backend = "hatchling.build"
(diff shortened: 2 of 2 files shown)


Remi-Gau:
The default config of this formater sorts the content of the toml alphabetically which has its pros and cons

Remi-Gau:
Can check if there is more config option to avoid this reformatting

.pre-commit-config.yaml
```diff
@@ -92,6 +92,12 @@ repos:
         args: ["--toml", "pyproject.toml"]
         additional_dependencies: [tomli]
 
+  - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
+    rev: v2.10.0
+    hooks:
+    - id: pretty-format-toml
+      args: [--autofix, --indent, '2']
```
emdupre, line 99: Any particular reason why 2 spaces instead of 4 ? I can't find a good, accepted [style guide for TOML](https://github.com/toml-lang/toml/issues/608), so I'd think this is just developer preference.

Remi-Gau: I honestly went with what was in the read of the  pre-commit hook repo? You want 4, let's do 4.you want 8, let's do 8. You want 16... Get me a larger screen.

emdupre: 😂 Just wanted to confirm that there wasn't a style guide I was missing ! I'm fine with 2 or 4.

Remi-Gau: given that our code uses 4, let's go with 4 to avoid the mental gymnastic of 4 for python, 2 for toml (my brain is too old and unflexible for those exercises)

bthirion (commented):
LGTM so far.

ymzayek (commented):
I do prefer the original order for some tables like `project` but I'm not too bothered by it. LGTM otherwise, thx!

Remi-Gau:
OK let's get this in then

Issue #3084, opened 2021-12-09
Adding a vmin option to plot_stat_map plot_glass_brain functions


rvarrier:
Hi,
I see that there have been a couple of related issues in the past requesting a "vmin" parameter for the plot_stat_map and/or plot_glass_brain functions (e.g. [this](https://github.com/nilearn/nilearn/issues/2165) and [this](https://github.com/nilearn/nilearn/issues/1149)), and was wondering if this has been implemented/is being worked on. 

### Benefits to the change
While the current implementation (only one half of the spectrum for positive/negative values alone) is a great way to ensure that positive/negative activations maintain their separate colors, it will be very helpful to get the whole color spectrum (within a desired vmin/vmax) to present other data (e.g. timepoints at which certain events occur) . This is especially important if the data is skewed (i.e., several nodes have very low/high values). Example below:

![image](https://user-images.githubusercontent.com/17010698/145492472-4a94c696-0789-4d48-840d-8f5862f5cae5.png)

I'd like to get the whole range of the colormap 'viridis' on the brainmap - i.e., from dark blue to yellow, such as is possible with seaborn heatmaps (example below): 
![image](https://user-images.githubusercontent.com/17010698/145493301-a2b3e35a-dd49-49be-92a9-fb15f88d427b.png)

The suggested solutions that I found so far - setting symmetric_cbar to False or using the plot_img function instead of the ones  mentioned above won't suffice unfortunately.

Thanks in advance!

bthirion:
Thx for bringing this back.
Note that in plot_stat_map, it is hard to see the lowest values, in a non-sparse setting, because the region with low values are masked visually by regions with high values.  
This being said, I don't remember well why we haven't kept the vmin argument. Maybe @jeromedockes remebers ?
Best,

jeromedockes:
As mentioned by @bthirion  in both linked issues, for the case of `plot_stat_map`, using `plot_img` instead does suffice.
```python
from nilearn import plotting, datasets
plotting.plot_img(datasets.load_mni152_template(), colorbar=True, vmin=.5)
```

![plot_img](https://user-images.githubusercontent.com/9196501/145842570-2fccc231-96a2-4081-a682-aa4e7dd78c47.png)


For the case of `plot_glass_brain`, I agree there is a bug: there is a `vmin` parameter but it is ignored and its documentation is incorrect. We can start to implement its description and possibly add a `symmetric_cmap` parameter as in the `view_img` function

[...]

Issue #1149, opened 2016-06-23
Adding back in vmin to plotting colormap command


KirstieJane:
At the moment the [_get_colorbar_and_data_ranges](https://github.com/nilearn/nilearn/blob/40069cb14b43333a73137797eac895cfb054db29/nilearn/plotting/img_plotting.py#L55) function within [img_plotting.py](https://github.com/nilearn/nilearn/blob/40069cb14b43333a73137797eac895cfb054db29/nilearn/plotting/img_plotting.py) by design forces the stat_map to be shown between `-vmax` and `vmax` or between `0` and `vmax`.

This clearly makes sense for [diverging colormaps](http://matplotlib.org/examples/color/colormaps_reference.html) but it doesn't make sense if you want to show, for example, regions that are significantly positively correlated using the `autumn` map or any other [sequential colormaps](http://matplotlib.org/examples/color/colormaps_reference.html). Ideally I'd like to be able to use the whole range of the colormap for values ranging between `2.3` and `5.0`. As the code currently runs, I'm left with only yellow regions!

![encode trial_gt_baseline allwords allsubs groupmean_glassbrain_ortho](https://cloud.githubusercontent.com/assets/3626306/16312080/f6a2ce66-396a-11e6-8376-f2f2000396fa.png)

What I'd really like to do is to have the colorbar start at `2.3` and finish at `5.0` and for that range to cover the whole colormap.

It looks like it would be easy to change the [_get_colorbar_and_data_ranges](https://github.com/nilearn/nilearn/blob/40069cb14b43333a73137797eac895cfb054db29/nilearn/plotting/img_plotting.py#L55) function to bring back the ability to set lower and upper limits for a colormap? Is there a reason not to?

erramuzpe:
+1 on this

alexlicohen:
This is still an issue

erramuzpe:
Hi again!

So, I encountered myself again with this bug :rofl: 

The key here are in [these](https://github.com/nilearn/nilearn/blob/master/nilearn/plotting/img_plotting.py#L1132-L1137) lines, where it could be that you have passed the function your vmin and `_get_colorbar_and_data_ranges` overwrites it and also assigns a mistaken `cbar_vmin`. I would like to solve this, but in `_get_colorbar_and_data_ranges` says clearly: 

```
if '    vmin_ = vmin' in kwargs:
        raise ValueError('this function does not accept a "vmin" '
                         'argument, as it uses a symmetrical range '
[...]

PR #3873, opened 2023-08-01
[ENH] add `vmin` argument to `plotting.plot_img_on_surf`

### LINKED ISSUE #3540: Adding a vmin option to plot_img_on_surf
I notice that the function "plotting.plot_img_on_surf" still does not support a "vmin" argument (nilearn version 0.10.0). If I enter a vmin parameter, it will raise an error saying "ValueError: this function does not accept a "vmin" argument, as it uses a symmetrical range defined via the vmax argument. To threshold the plotted map, use the "threshold" argument".
But unfortunately, the "threshold" argument still brings out a symetrical colormap. The colorbar always contains symetrical positive and negative sides. How could I set the vmin of colorbar when plotting images on surface, e.g. showing only the positive values?

michellewang:
Closes #3540.

- Added `vmin` and `symmetric_cbar` arguments to `nilearn.plotting.surf_plotting.plot_img_on_surf`
  - If `symmetric_cbar` is `True` (default), `vmin` is expected to be the same as `-vmax` if it is provided.
  - If `symmetric_cbar` is `False`, the colorbar (and the colormap range) will range from `-vmin` to `vmax`. If the colormap is divergent this can give a colorbar that is not centered at 0. The `threshold` argument works as usual (and remains symmetrical to be consistent with all the other plotting functions).

Here are some examples plots with `vmin`:
```python
from nilearn import datasets
from nilearn import plotting

plotting.plot_img_on_surf(
    stat_img,
    views=['lateral', 'medial'],
    colorbar=True,
    vmin=4,
    vmax=10,
    symmetric_cbar=False,
    threshold=4,
    cmap='viridis',
)
```

image

```python
plotting.plot_img_on_surf(
    stat_img,
    views=['lateral', 'medial'],
    colorbar=True,
    vmin=-5,
    vmax=10,
    symmetric_cbar=False,
    threshold=2,
    cmap='viridis',
)
```

image

Note: this PR modifies `_get_colorbar_and_data_ranges` in `nilearn.plotting.img_plotting`, which is also used in `plot_stat_map` and `plot_glass_brain` in the same module. The current PR should not change the behaviour of those functions, though in the future `vmin` should also be added to these two functions (see #3084 and #1149). The reason why I didn't make the changes in this PR is that I'm not sure about the logic of having `vmin`/`vmax` and separate `cbar_vmin`/`cbar_vmax`. As far as I can tell the surface plotting functions do not use `cbar_vmin`/`cbar_vmax`, but `plot_glass_brain` and `plot_stat_map` do, so if those are changed then we would need to decide what `cbar_vmin`/`cbar_vmax` should be for different cases.

While testing some plotting scenarios, I noticed that it was possible to pass `engine='plotly'` to `plot_img_on_surf`, which produces something like this (code runs without errors):

image

I think this is happening because `plot_img_on_surf` creates Matplotlib Axes objects which do not seem to be compatible with Plotly. Do we want to raise an error if users call the function with `engine='plotly'`?


ymzayek:
@michellewang thanks for starting this PR. Concerning the comment about `engine` argument being passed to `plot_img_on_surf` I think we can just add the keyword to this line that looks at the kwargs and raises a ValueError based on the list: https://github.com/nilearn/nilearn/blob/a95149b2162eff9bc4c0b6fe503533311a700f02/nilearn/plotting/surf_plotting.py#L1342

michellewang:
I got some `flake8` errors in the CI checks even though I ran `pre-commit`... So I ran `black` manually (without options) which changed some functions that I hadn't touched, not sure if that was the right thing to do

Remi-Gau:
> I got some `flake8` errors in the CI checks even though I ran `pre-commit`... So I ran `black` (and `flake8`) manually (without options) which changed some functions that I hadn't touched, not sure if that was the right thing to do

ouch... that is one of the consequence of not having 'blacked out' all of our code base.

from what I see only the last commit is problematic and included a lot of black changes.

so just revert that one (or reset to the commit before) and fix the flake8 errors manually (like in the good old days).

running flake8 locally should tell you what the problem is.

Remi-Gau:
looks like you solved it!

michellewang:
Yes, thanks @Remi-Gau!

nilearn/plotting/surf_plotting.py
```diff
@@ -1123,13 +1126,18 @@ def plot_surf_stat_map(surf_mesh, stat_map, bg_map=None,
...
+        symmetric_cbar=symmetric_cbar,
+        symmetric_data_range=False,
+    )
 
     display = plot_surf(
         surf_mesh, surf_map=loaded_stat_map,
         bg_map=bg_map, hemi=hemi,
         view=view, engine=engine, avg_method='mean', threshold=threshold,
-        cmap=cmap, symmetric_cmap=True, colorbar=colorbar,
+        cmap=cmap, symmetric_cmap=symmetric_cbar, colorbar=colorbar,
```
bthirion, line 1140:
```suggestion
        cmap=cmap, symmetric_cmap=symmetric_cmap, colorbar=colorbar,
``` (not sure I understand)

**Review by bthirion — COMMENTED** [review]
Thx for opening that ! This is not simple stuff.
Overall LGTM. I'm just worried about a possible confusion btw symmetric_cbar` and `symmetric_cmap`

File: nilearn/plotting/img_plotting.py [hunk 2]
```diff
@@ -60,42 +60,68 @@ def show():
...
+        if vmin is None and vmax is None:
+            vmax = max(-stat_map_min, stat_map_max)
+            vmin = -vmax
+        elif vmin is None:
+            vmin = -vmax
+        elif vmax is None:
+            vmax = -vmin
+        elif not np.isclose(vmin, -vmax):
+            raise ValueError(
+                "vmin must be equal to -vmax unless symmetric_cbar"
```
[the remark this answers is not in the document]
michellewang: I added it because I wasn't sure how to handle the case where both `vmin` and `vmax` are given (i.e. not `None`) but they are not the same and `symmetric_cbar` is `True`. The options I can think of are to:
1. force the colourbar to be symmetric by changing `vmin` or `vmax`, maybe with a warning that the user input is being overridden
2. ignore `symmetric_cbar` and just use the user-provided `vmin` and `vmax` (also with a warning)
3. not allow this scenario to happen (current behaviour)

What do you think?

To clarify, this is only when the user provides values for `vmin` and `vmax` and `symmetric_cbar=True`. If only `vmin` or `vmax` is provided then that's not a problem (the colourbar can still be symmetric). Maybe the error message isn't clear enough, I can change it if that's the case.

nilearn/plotting/surf_plotting.py
michellewang: I also find this confusing. [`symmetric_cmap`](https://github.com/search?q=repo%3Anilearn%2Fnilearn+symmetric_cmap&type=code) and [`symmetric_cbar`](https://github.com/search?q=repo%3Anilearn%2Fnilearn+symmetric_cbar&type=code) are used in different functions, and it is unclear to me if they are supposed to be the same thing. 

After looking into this a little more, I think I should change this back to what it was before (`symmetric_cmap=True`), since the [docs for `plot_surf`](https://github.com/mathdugre/nilearn/blob/44695d8c9c784d71bc2635650236859155089c0a/nilearn/plotting/surf_plotting.py#L679-L688) say that `symmetric_cmap` is only implemented for the `plotly` engine, and the `plot_img_on_surf` function I changed specifically does not handle plotting with Plotly.

nilearn/plotting/img_plotting.py
```diff
@@ -60,42 +60,68 @@ def show():
...
+                " and symmetric_data_range are False."
+            )
 
-    if symmetric_cbar == 'auto':
-        symmetric_cbar = stat_map_min < 0 and stat_map_max > 0
+    # force vmin to be -vmax
+    if symmetric_data_range:
+        vmin = -vmax
```
ymzayek, line 110: is this necessary? seems like it will be set in the previous if statement

michellewang: I think you're right, I'll remove it

nilearn/plotting/img_plotting.py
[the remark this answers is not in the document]
ymzayek: > I'm wondering: do we require vmin=-vmax is symmetric_cbar is True ?

As I understand it we do, this is what is defined in the docstrings of the plotting functions that use this parameter (e.g. see [plot_surf_stat_map](https://nilearn.github.io/stable/modules/generated/nilearn.plotting.plot_surf_stat_map.html#nilearn-plotting-plot-surf-stat-map))

However, I think symmetric_data_range should not be in the error message as this is used internally and not exposed to the user (in fact in the current implementation the user can pass `symmetric_data_range=True` and `symmetric_cbar=False` with defined unequal vmin/vmax values and the error is not raised because the value of symmetric_data_range is overridden internally)

As I see it symmetric_data_range is only for controlling internal behavior with regards to other functions that do not yet allow vmin like `plot_glass_brain` and this parameter is meant to be temporary until the other functions are dealt with (in other PRs). @michellewang is this correct?

Last thing, the functions that use cbar_vmin/cbar_vmax only use them if symmetric_cbar is not True, otherwise they are set to None and eventually set to the normalized vmin/vmax and they are only used internally

nilearn/plotting/img_plotting.py
```diff
@@ -60,42 +60,68 @@ def show():
...
+    always be set to range from -vmax to vmax. The limits for the colorbar
+    depend on the symmetric_cbar argument, please refer to docstring of
+    plot_stat_map.
     """
-    if 'vmin' in kwargs:
+    if symmetric_data_range and (vmin is not None):
         raise ValueError('this function does not accept a "vmin" '
                          'argument, as it uses a symmetrical range '
                          'defined via the vmax argument. To threshold '
                          'the plotted map, use the "threshold" argument')
```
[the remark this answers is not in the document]
ymzayek: Yes makes sense, we can leave it for now

nilearn/plotting/img_plotting.py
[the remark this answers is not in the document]
michellewang: > However, I think symmetric_data_range should not be in the error message as this is used internally and not exposed to the user (in fact in the current implementation the user can pass symmetric_data_range=True and symmetric_cbar=False with defined unequal vmin/vmax values and the error is not raised because the value of symmetric_data_range is overridden internally)

That makes sense, I'll change the error message

> As I see it symmetric_data_range is only for controlling internal behavior with regards to other functions that do not yet allow vmin like plot_glass_brain and this parameter is meant to be temporary until the other functions are dealt with (in other PRs). @michellewang is this correct?

Yes exactly

nilearn/plotting/surf_plotting.py
```diff
@@ -1262,12 +1277,11 @@ def _colorbar_from_array(array, vmax, threshold, kwargs,
...
-                     hemispheres=['left', 'right'],
-                     bg_on_data=False, inflate=False,
-                     views=['lateral', 'medial'],
+                     hemispheres=['left', 'right'], bg_on_data=False,
+                     inflate=False, views=['lateral', 'medial'],
                      output_file=None, title=None, colorbar=True,
-                     vmax=None, threshold=None,
-                     cmap='cold_hot', **kwargs):
+                     vmin=None, vmax=None, threshold=None,
+                     symmetric_cbar=True, cmap='cold_hot', **kwargs):
```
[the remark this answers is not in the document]
michellewang: No I don't think so, I will change it to `"auto"`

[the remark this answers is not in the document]
michellewang: I realized that, with the way I had the code, if `symmetric_cbar` was `"auto"` and the user provided a `vmin` and a `vmax` that are not opposites of each other, then the function would throw an error telling the user to set `symmetric_cbar` to `False`. So now I changed it so if `symmetric_cbar` is `"auto"`, it's checked and set to `True` or `False` a bit before that.

bthirion (approved):
LGTM, thx.

nilearn/plotting/surf_plotting.py
```diff
@@ -486,14 +486,19 @@ def _threshold_and_rescale(data, threshold, vmin, vmax):
...
     """Thresholds the data."""
     # If no thresholding and nans, filter them out
-    return (np.logical_not(np.isnan(data)) if threshold is None
-            else np.abs(data) >= threshold)
+    if threshold is None:
+        mask = np.logical_not(np.isnan(data))
+    else:
+        mask = np.abs(data) >= threshold
+        if vmin > -threshold:
+            mask = np.logical_and(mask, data >= vmin)
```
ymzayek, line 500: @michellewang I just have one question left about this addition because the resulting behavior of this change is not clear to me. Can you show the behavior difference here given when the condition is met and when it's not met (an example in the PR is fine)? I'm just wondering if it would be clear to the user how vmin value will affect the thresholding.

michellewang: @ymzayek if `vmin <= -threshold` then the old behaviour (`mask = np.abs(data) >= threshold`) works as expected: the image is clipped at `vmin` and shows data between `-threshold` and `vmin`

```python
from nilearn import plotting

plotting.plot_img_on_surf(
    img,
    colorbar=True,
    symmetric_cbar=False,
    vmin=-3,                          # vmin <= -threshold
    threshold=2,
    cmap='viridis',
)
```

image

Now that we have `vmin`, the issue with the previous version of `_threshold()` was that it doesn't account for cases where the data range is asymmetric, for example if the user only wants to plot positive values. Below is what the plot looks like with the old `_threshold()` behaviour if `vmin=0` (i.e., `vmin > -threshold`). We can see that the colourbar is correct but the actual image still shows negative values beyond the threshold, which is unexpected.

image

Updating the mask based on `vmin` if `vmin > -threshold` fixes this:

image

And now I just realized that we would have a similar problem when `vmax < threshold` (for example if the user only wanted to plot negative values, with something like `vmax = 0` and `threshold = 2`), so I will add another part to `_threshold()` so that it also updates the mask in that situation.

Let me know if this makes sense and/or if there are edge cases that I missed and/or we should clarify the new behaviour in the docstring.

ymzayek: Very clear thanks! I think with the addition of handling vmax and a green CI we should be good to go.

ymzayek (approved):
LGTM @michellewang, thanks!

ymzayek:
@michellewang are you willing to work on the other related issues to fix this for `plot_glass_brain` and `plot_stat_map`?

michellewang:
@ymzayek yes I can do it, though it might take a little longer because I will be away next week and part of the one after

PR #3877, opened 2023-08-03
[FMT] apply yml format pre commit hook

Remi-Gau:
Relates to #3870

- set up pre commit hook to format yml and apply it


environment.yml
ymzayek, line 1: Do we need this file? Looks like it was added in #2633 with the intention of using it to set up a conda environment but then never actually used

Remi-Gau: I was convinced it was mentioned in the doc as another way to install nilearn for the devs but it is not.

https://nilearn.github.io/stable/development.html#setting-up-your-environment

Not used in CI anywhere else as far as I can tell.

I would say that it can be safely removed.

PR #3881, opened 2023-08-03
[ENH] check affine and shape of first level models when passed at the second level 

### LINKED ISSUE #3298: SecondLevelModel is late to notice Target and Reference image mismatch
I think that the SecondLevelModel object when operating on a list of FirstLevelModels should first check whether all FirstLevelModels are in the same Affine and Shape before running because in my case it fails after processing multiple models.

In addition it would be nice if there was a feature to force resample everything at the 2nd level as well.

### Benefits to the change
User would know in advanced that data is not fit for 2nd level analysis prior to it starting.

```python

```

Remi-Gau:
- Closes #3298

- add test
- implement checks

### TODO

- [x] check shape

### DIFF
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -119,11 +119,12 @@ def _check_input_as_first_level_model(second_level_input, none_confounds):
 
     - must have been fit
     - must all have a subject label in case confounds are passed
-    - must all have the same affine
-      (checking all against the affine of the first model)
+    - must all have the same affine / shape
+      (checking all against those of the first model)
 
     """
     ref_affine = None
+    ref_shape = None
 
     for model_idx, first_level in enumerate(second_level_input):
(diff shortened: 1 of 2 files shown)


bthirion (commented):
LGTM so far.

nilearn/glm/second_level/second_level.py
```diff
@@ -120,12 +134,42 @@ def _check_input_as_first_level_model(second_level_input, none_confounds):
...
-                "To set it, you can do "
-                "first_level.subject_label = '01'"
+                "To set it, you can do first_level.subject_label = '01'"
+            )
+
+        affine = None
+        shape = None
+        if first_level.mask_img is not None:
+            if isinstance(first_level.mask_img, NiftiMasker):
+                affine = first_level.mask_img.affine_
```
Remi-Gau, line 148: TODO: get shape too in this case

ymzayek (approved):
LGTM thx @Remi-Gau !

PR #3882, opened 2023-08-03
[MAINT] uniformize value of MNI affine used in tests

### LINKED ISSUE #3874: Choose better affine constant for plotting test data
At the time I was "just" refactoring code and started adding fixtures when doing so. So I did not have a close look at the exact content.

Digging into this now I realize we may have some inconsistency.

```python
from nilearn.plotting.img_plotting import MNI152TEMPLATE
mni_affine = MNI152TEMPLATE.get_affine()
print(mni_affine)
```

gives
```
[[   2.    0.    0.  -98.]
 [   0.    2.    0. -134.]
 [   0.    0.    2.  -72.]
 [   0.    0.    0.    1.]]
```

which does not quite match that in conf.py

```python
MNI_AFFINE = np.array(
    [
        [-2.0, 0.0, 0.0, 90.0],
        [0.0, 2.0, 0.0, -126.0],
        [0.0, 0.0, 2.0, -72.0],
        [0.0, 0.0, 0.0, 1.0],
    ]
)
```

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/3172#issuecomment-1662060133_

Remi-Gau:
- Closes #3874


bthirion (commented):
LGTM so far.

nilearn/plotting/tests/test_img_plotting/test_img_plotting.py
```diff
@@ -38,14 +38,14 @@
 PLOTTING_FUNCS_3D = ALL_PLOTTING_FUNCS.difference(PLOTTING_FUNCS_4D)
 
 
-def _test_data_with_nans(img):
+def _test_data_with_nans(img, mni_affine=MNI_AFFINE):
```
ymzayek, line 41: Since a fixture exists for this, can't you just do:

```suggestion
def _test_data_with_nans(img, mni_affine):
```
and remove `from nilearn.conftest import MNI_AFFINE`

This is also relevant for `demo_plot_roi` function in `test_plot_roi.py`

ymzayek (commented):
LGTM otherwise

nilearn/plotting/tests/test_img_plotting/test_img_plotting.py
Remi-Gau: let me check because I think that for this one "it is not that simple" (TM)
but "I may wrong" (story of my life)

Remi-Gau: yup this function actually just adds nans to an image to then pass it to some other tests

so AFAICT it cannot use fixtures

Remi-Gau: am renaming the function to match what it does

bthirion (approved):
LGTM, thx !

PR #3890, opened 2023-08-05
[MAINT] extract dataset ids and file lists

### LINKED ISSUE #3704: extract dataset IDs in separate functions
This isn't related to reformatting but I'm wondering whether we should have helper functions whose only responsibility is to return the list of ids for a given dataset fetcher.

So basically something like this:

```python
# In the fetcher
ids = _get_nki_ids()

# Helper function
def _get_nki_ids():
    return [....]
```

 It's not super important, but these massive lists are breaking the function logic and force the reader to scroll up and down which can be annoying...
WDYT ?

_Originally posted by @NicolasGensollen in https://github.com/nilearn/nilearn/pull/3695#discussion_r1170921752_

Remi-Gau:
- Closes #3704

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -217,46 +217,9 @@ def fetch_haxby(
     )
 
 
-@fill_doc
-def fetch_adhd(n_subjects=30, data_dir=None, url=None, resume=True, verbose=1):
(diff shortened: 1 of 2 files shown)


ymzayek (approved):
LGTM thx!

PR #3897, opened 2023-08-07
[FIX/ENH] Improve generate_report method to work with fit_transform for the maskers

### LINKED ISSUE #3709: [ENH] generate_report method should also work after fit_transform was called

Currently, it seems to be the case that the `.generate_report` method of masker objects can only be called with a previous call of `.fit()` but not when `.fit_transform()` was called (which is contra-intuitive to me because `.fit_transform()` is just `.fit()` + `.transform()`). Most of the time, users will probably use `.fit_transform()` to get their numpy arrays. If they then want to check if the masking worked they currently have to call `.fit()` one more time to be able to call  `.generate_report`.

### Benefits to the change

Avoids calling `masker.fit()` one more time if a report should be generated. 

MRE:

```python
from nilearn import datasets
import numpy as np
from nilearn.image import load_img

# download a single statistical image

# get an atlas
atlas_data = datasets.fetch_atlas_schaefer_2018()

# get the atlas image. This has 401 values, with the 0 included
atlas_img = load_img(atlas_data['maps'])
print(len(np.unique(atlas_img.get_fdata())))

# we have to insert a 0 here so that img and labels have the same number of value
# This is already reported, see #3708
atlas_labels = atlas_data.labels.tolist()
atlas_labels.insert(0,'Background') 

# set a masker
masker = NiftiLabelsMasker(labels_img=atlas_img,labels=atlas_labels)

```
**Would be nice if this would work!**

```python

data = masker.fit_transform(stat_img)
report.open_in_browser()
```

**But we have to call .fit() one more time in order to be able to generate a report**

```
masker.fit(stat_img)
report.open_in_browser()
```

ymzayek:
- Deals with most of #3709

- [x] Pass image data to fit() in fit_transform()
- [x] Use better colormap or instead allow users to pass colormap
- [x] Improve docstrings to explain image data being passed to reporter
- [ ] Store 1 time point for reporting using 4D images (will handle in a separate PR)

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -32,6 +32,7 @@ Fixes
 
 - Fix pathlib.Path not being counted as Niimg-like object in :func:`~image.new_img_like` (:gh:`3723` by `Maximilian Cosmo Sitter`_).
 
+- Fix ``fit_transform`` behavior to match when ``fit`` method is passed image data (:gh:`3897` by `Yasmin Mzayek`_)
 
 Enhancements
 ------------
@@ -47,6 +48,8 @@ Enhancements
 - Add ``LassoCV`` as a new estimator option for Decoder objects (:gh: `3781` by `Michelle Wang`_)
 
 - Add ``vmin`` and ``symmetric_cbar`` arguments to :func:`~nilearn.plotting.plot_img_on_surf` (:gh:`3873` by `Michelle Wang`_).
+
+- Improve ``generate_report`` method of maskers by allowing users to pass a cmap argument for plotting image (:gh:`3897` by `Yasmin Mzayek`_)
 
 Changes
 -------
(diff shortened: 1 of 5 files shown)


ymzayek:
Now 
```python
data = masker.fit_transform(img)
```

works the same as:
```python
masker.fit(img)
```
to generate: 
Untitled


I changed the colormap but maybe there are suggestions for something better??

ymzayek:
Storing one time point for the report seems a bit more involved than I thought given the failures here: https://github.com/nilearn/nilearn/actions/runs/5787425257 so I will handle it in a separate follow-up PR

bthirion:
> 

The red background is not great for the atlas display. I'll try to get a neutral one.

bthirion (commented):
LGTM otherwise

bthirion (approved):
LGTM, thx.

Remi-Gau (approved):
Good with me!

PR #3900, opened 2023-08-10
[INFRA] Prevent Actions workflows from running on people's forks

ymzayek:
- Closes #

I noticed all these workflows will get triggered when people work on PRs from the main branch of their fork or any time they push to their main branch.

### DIFF
--- a/.github/workflows/isort.yml
+++ b/.github/workflows/isort.yml
@@ -15,6 +15,7 @@ concurrency:
 
 jobs:
     isort:
+        if: github.repository == 'nilearn/nilearn'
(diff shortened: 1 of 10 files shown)


ymzayek:
> I would still let the fast CI workflows run on forks ,no?

Ok that's fine

.github/workflows/isort.yml
```diff
@@ -15,6 +15,7 @@ concurrency:
 
 jobs:
     isort:
+        if: github.repository == 'nilearn/nilearn'
```
Remi-Gau, line 18:
```suggestion
```

another short one

PR #3902, opened 2023-08-10
[MAINT] Fix failing tests in prerelease job due to deprecation and bug in matplotlib 3.8.0rc1

ymzayek:
- Closes #

- In matplotlib > 3.7.2 axes have to be explicitly removed using `ax.remove()` instead of being automatically overwritten
- Adjust failing test in `test_displays.py` to get around matplotlib bug https://github.com/matplotlib/matplotlib/issues/26531

### DIFF
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -1513,7 +1513,8 @@ def plot_carpet(img, mask_img=None, mask_labels=None, t_r=None,
             ax0.set_yticks([])
 
         # Carpet plot
-        axes = plt.subplot(gs[1])  # overwrite axes
+        axes.remove()
+        axes = plt.subplot(gs[1])
         axes.imshow(
             data.T,
             interpolation='nearest',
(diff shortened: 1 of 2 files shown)


ymzayek:
I am still unsure about how to fix the second failing test: `nilearn/plotting/tests/test_displays.py::test_contour_fillings_levels_in_add_contours`

ymzayek:
Seems to me a bug on matplotlib's side. IIUC vmin and vmax should map data values to colors on the colormap and data values lower or higher than vmin/vmax should be mapped to the lowest/highest color in the colomap. For some reason, vmax is equal to zero when levels argument is less than vmin and vmin is given. vmax and vmin should not be affected by the levels parameter. Will open an issue there and confirm, then do a hot fix here.

ymzayek:
The documentation failure doesn't seem to be related. I also failed on the [scheduled build on main](https://github.com/nilearn/nilearn/actions/runs/5887398120/job/15966733974). Will look into separately. I think this one can be merged

PR #3904, opened 2023-08-11
typo

arovai:


ymzayek:
Hm isn't this just a case of British vs US English?

arovai:
It is, sorry for that. Never mind.
By the way, thanks for maintaining such an awesome package!

ymzayek:
@arovai no worries, you can have a look at these ["good first issues"](https://github.com/nilearn/nilearn/labels/Good%20first%20issue) if you're interested in starting to contribute to nilearn.
And thanks, happy that you enjoy using it!

Remi-Gau:
Actually a good first issue would be to set up our spell checker (codespell) to only use american english or just british english - but only one of those.

It is a matter of adding one line in this file: https://github.com/nilearn/nilearn/blob/48aecfacd9f0411c50125658966d5a61f40ec2e3/pyproject.toml#L154 

and mention which built-in dictionary to use. See here for example:
https://github.com/bids-standard/bids-specification/blob/3ddcb0d82007ea1a39d14fd1fb563907c7f89807/.codespellrc#L4

@arovai would you be interested in doing this?

Remi-Gau:
to be entirely honest it will likely be more just than one line but you can use the codespell package itself to help you make those other changes fairly quickly.

PR #3905, opened 2023-08-11
[MAINT] Refactor fixtures

Remi-Gau:
- Relates to #3331

In nilearn/conftest.py adds:
- constants for common affines and shapes
- functions to return those constant
- fixtures to use those constants
- functions and fixtures to use 3D or 4D NifTiImages filled with zeros, ones, or random numbers

Apply those changes in the tests that have already been refactored


Remi-Gau:
oops... introduced some test pollution
will see if https://github.com/asottile/detect-test-pollution can help

nilearn/conftest.py
ymzayek, line 1: @Remi-Gau why is it that for mni affine the fixture returns the `AFFINE_MNI` array itself whereas with the eye affine the fixture returns a function that returns `AFFINE_EYE` array? Same question for `shape_3d_default` vs `shape_4d_default`

nilearn/regions/tests/test_region_extractor.py
```diff
@@ -285,7 +273,7 @@ def test_region_extractor_high_resolution_image():
 
 def test_region_extractor_zeros_affine_diagonal():
     n_regions = 9
-    affine = AFFINE_EYE
+    affine = np.eye(4)
```
ymzayek, line 276: why not use fixture here?

nilearn/decomposition/tests/test_base.py
```diff
@@ -4,13 +4,10 @@
...
 from nilearn.decomposition._base import _fast_svd, _mask_and_reduce
 from nilearn.maskers import MultiNiftiMasker
 
-AFFINE_EYE = np.eye(4)
-
-SHAPE = (6, 8, 10)
-
 
 @pytest.fixture
 def data_for_mask_and_reduce():
```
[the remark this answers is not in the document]
ymzayek: Oops sorry ignore this comment since this is a fixture

nilearn/conftest.py
Remi-Gau: General answer.

I went for

```python
# global variable
X = "some value"

# function
def _x()
  return X

# fixture
@pytest.mark.fixture()
def x():
  return _x()
```

When some other fixtures needed to access the value of X and change it.

So to avoid having that fixture change the global (AKA "best way to make things break") I added a function as a go between.

I have not done that systematically because I have not had the need for it for all of those constants. But I think that it would make sense to have the same pattern for all, to avoid this confusion.

nilearn/regions/tests/test_region_extractor.py
Remi-Gau: was changed in this commit

using the fixture lead to some test pollution (non independent tests) as the value of the affine is changed in the following lines

is there a better way to use fixture in this case?

make the fixture return the value and not the constant?

Remi-Gau:
> @Remi-Gau thanks for the explanations. I don't necessarily think you have to change the way you did things if you have a good reason but a comment or docstrings for the helper functions and fixtures might be sufficient. And this could take a bit of extra work to get a good "base" for fixtures that we can then iterate on but I think it will improve testing quite a bit so thanks!

Good point will add some comments and doc strings.

nilearn/conftest.py
```diff
@@ -89,7 +94,114 @@ def close_all():
...
+
+@pytest.fixture()
+def affine_eye():
+    """Return an identity matrix affine."""
+    return _affine_eye()
+
+
+# ------------------------ SHAPES ------------------------#
+
+# Default shatpe for a 3D and 4D images
```
ymzayek, line 140:
```suggestion
# Default shape for a 3D and 4D images
```

nilearn/plotting/tests/test_img_plotting/test_plot_roi.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 from nibabel import Nifti1Image
 
-from nilearn.conftest import MNI_AFFINE
+from nilearn.conftest import AFFINE_MNI
```
ymzayek, line 8: Should you here then use `_affine_mni` instead?

nilearn/plotting/tests/test_img_plotting/test_img_plotting.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 from nibabel import Nifti1Image
 
-from nilearn.conftest import MNI_AFFINE
+from nilearn.conftest import AFFINE_MNI
```
ymzayek, line 8: also use `_affine_mni` function here?

nilearn/conftest.py
```diff
@@ -89,7 +94,114 @@ def close_all():
 )
 
 
-def _mni_3d_img(affine=MNI_AFFINE):
+def _affine_mni():
```
ymzayek, line 97: Does anything prevent from always using this even inside conftest? What is the benefit of still having the global?

Remi-Gau: yeah I was thinking that the global was progressively becoming useless. 

will try to simplify.

bthirion: But isn't that a breaking change ? I'd rather have deprecation cycles...

bthirion (commented):
LGTM.  Simply woried about API breakage in conftest.py

nilearn/conftest.py
Remi-Gau: looking through all the files changed, all those constants only appear in tests so I don't think deprecation is necessary there.

are we actually storing anything in conftest.py that is not just for testing? if so we should probably move it somewhere else

Remi-Gau:
I was thinking of adding a bit of contributing doc regarding the use of fixtures: should I do it in this PR? It feels more appropriate but it may make the PR bigger.

bthirion:
I was just thinking that some extra package may rely on it ...

bthirion:
> I was thinking of adding a bit of contributing doc regarding the use of fixtures: should I do it in this PR? It feels more appropriate but it may make the PR bigger.

Should be in this PR I think.

Remi-Gau:
> I was just thinking that some extra package may rely on it ...

except pytest plugins I think that no other package would expect its config information to go in there (at least I kind of hope so)

ymzayek:
>> I was just thinking that some extra package may rely on it ...

> except pytest plugins I think that no other package would expect its config information to go in there (at least I kind of hope so)

In addition the changes in conftest.py have not been released in a stable version yet

CONTRIBUTING.rst
Remi-Gau, line 1: Updated the doc on the tests to talk about fixtures and improve this section in general with a couple of extra links.

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
       see `test-driven development <https://en.wikipedia.org/wiki/Test-driven_development>`_.
 
+We use `pytest <https://docs.pytest.org/en/6.2.x/contents.html>`_ to run our tests.
+
+If you are not familiar with pytest,
+have a look at this `introductory video <https://www.youtube.com/watch?v=mzlH8lp4ISA>`_
+by one of the pytest core developer.
+
+In general tests for a specific module (say `nilearn/image/image.py`)
+are kept in a `tests` folder in separate module
```
ymzayek, line 398:
```suggestion
are kept in a `tests` folder in a separate module
```

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
+with a name that matches the module being tested
+(so in this case `nilearn/image/tests/test_image.py`).
+
+When you have added a test you can check your changes worked
+and didn't break anything run `pytest nilearn`.
```
ymzayek, line 403:
```suggestion
When you have added a test you can check that your changes worked
and didn't break anything by running `pytest nilearn`.
```

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
+To do quicker checks it's possible to run only a subset of tests:
+
+.. code-block:: bash
+
+      pytest -v nilearn/module/tests/test_module.py
+
+Fixtures
+########
```
ymzayek, line 411: why the hash signs instead of dashes as in the rest of the document?

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
+
+If you need to some special "set up" for your tests
```
ymzayek, line 413:
```suggestion
If you need to do some special "set up" for your tests
```

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
+(for example you need to generate some data, or a NiftiImage object or a file...)
+you can use `pytest fixtures <https://docs.pytest.org/en/6.2.x/fixture.html>`_
+to help you mock this data
+(more information on pytest fixtures in `this video <https://www.youtube.com/watch?v=ScEQRKwUePI>`_).
+
+Fixture are recognizable because they have a `@pytest.fixture` decorator.
+Fixtures that are shared by many tests modules can be found in `nilearn/conftest.py`
+but some fixures specific to certain modules can also be kept in that testing module.
+
+Seending
```
ymzayek, line 423:
```suggestion
Seeding
```

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
+########
```
ymzayek, line 424: same comment

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
+
 Tests must be seeded to avoid random failures.
 For objects using random seeds (e.g. scikit-learn estimators), pass either
-a  `np.random.RandomState` or an `int` as the seed.
-When your test use random numbers,  those must be generated through:
+a `np.random.RandomState` or an `int` as the seed.
+When your test use random numbers,
```
ymzayek, line 429:
```suggestion
When your tests use random numbers,
```

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
+but some fixures specific to certain modules can also be kept in that testing module.
+
+Seending
+########
+
 Tests must be seeded to avoid random failures.
 For objects using random seeds (e.g. scikit-learn estimators), pass either
-a  `np.random.RandomState` or an `int` as the seed.
-When your test use random numbers,  those must be generated through:
+a `np.random.RandomState` or an `int` as the seed.
```
ymzayek, line 428: Maybe for a separate PR but shouldn't we update to use the Generator API as mentioned in https://github.com/nilearn/nilearn/issues/3331#issuecomment-1684047671?

CONTRIBUTING.rst
```diff
@@ -388,26 +388,70 @@ They should run on small mocked data, cover a representative range of parameters
...
+
+If you need to some special "set up" for your tests
+(for example you need to generate some data, or a NiftiImage object or a file...)
+you can use `pytest fixtures <https://docs.pytest.org/en/6.2.x/fixture.html>`_
+to help you mock this data
+(more information on pytest fixtures in `this video <https://www.youtube.com/watch?v=ScEQRKwUePI>`_).
+
+Fixture are recognizable because they have a `@pytest.fixture` decorator.
+Fixtures that are shared by many tests modules can be found in `nilearn/conftest.py`
+but some fixures specific to certain modules can also be kept in that testing module.
```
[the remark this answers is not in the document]
Remi-Gau: added one sentence to make it explicit.

CONTRIBUTING.rst
Remi-Gau: I updated it here already: what do you think?

ymzayek: Yes, good with me! At some point we can update our tests to use it but that really should be in a separate PR

Remi-Gau: agreed

I will take care of it soon but it should be done in a separate or several PRs

PR #3917, opened 2023-08-17
[MAINT] apply black to utils.doc

Remi-Gau:
- Closes none

- as far as I can tell _utils.docs.py is not involved in any PR at the moment so we can let black loose on her.


ymzayek (approved):
LGTM thx!

PR #3934, opened 2023-08-29
[DOC] Plot only a few examples of probabilistic atlases

ymzayek:
- Part of #3728

-
-

### DIFF
--- a/examples/01_plotting/plot_prob_atlas.py
+++ b/examples/01_plotting/plot_prob_atlas.py
@@ -61,57 +61,59 @@ plotting.show()
 #
 # To save build time, the following code is not executed. Try running it
 # locally to get the same plots as above for each of the listed atlases.
-
-# # Harvard Oxford Atlas
-# harvard_oxford = datasets.fetch_atlas_harvard_oxford("cort-prob-2mm")
-# harvard_oxford_sub = datasets.fetch_atlas_harvard_oxford("sub-prob-2mm")
-
-# # Smith ICA Atlas and Brain Maps 2009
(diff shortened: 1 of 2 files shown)


ymzayek:
The built changed example: https://output.circle-artifacts.com/output/job/3da2b1b7-db5e-41a0-8b82-3d61d9d151c2/artifacts/0/dev/auto_examples/01_plotting/plot_prob_atlas.html#sphx-glr-auto-examples-01-plotting-plot-prob-atlas-py now runs in ~25 seconds

examples/01_plotting/plot_prob_atlas.py
```diff
@@ -24,76 +24,94 @@
...
-    title="Smith2009 10 Brainmap (with colorbar)",
+    smith_bm20,
+    title="Smith2009 20 Brainmap (with colorbar)",
     colorbar=True,
 )
-print("ready")
+
 plotting.show()
 
+#########################################################################
```
bthirion, line 58: Isn't there a more elegant way to do this ? Eg use triple quotes for the whole block so that people can easily copy-paste.

ymzayek: Yes I think so. Let me see what renders best. I'll repost an updated link

ymzayek: Updated docs: https://output.circle-artifacts.com/output/job/ff7c070a-3b19-409d-b7ea-6b1323283c17/artifacts/0/dev/auto_examples/01_plotting/plot_prob_atlas.html#sphx-glr-auto-examples-01-plotting-plot-prob-atlas-py

PR #3935, opened 2023-08-29
[MAINT] Store 1 timepoint for 4D generating a report with the maskers

### LINKED ISSUE #3933: Store one time point for reporting that uses 4D images
> the full images are stored for reporting, we should probably only store one time point when the image is 4D

Comes from @jeromedockes comment in https://github.com/nilearn/nilearn/issues/3709#issuecomment-1561608917

ymzayek:
- Closes #3933

-
-

### DIFF
--- a/nilearn/maskers/multi_nifti_masker.py
+++ b/nilearn/maskers/multi_nifti_masker.py
@@ -174,6 +174,7 @@ class MultiNiftiMasker(NiftiMasker, _utils.CacheMixin):
         )
         self.n_jobs = n_jobs
         self._shelving = False
+        self._reporting_data = None
 
     def fit(self, imgs=None, y=None):
         """Compute the mask corresponding to the data.
--- a/nilearn/maskers/multi_nifti_masker.py
+++ b/nilearn/maskers/multi_nifti_masker.py
@@ -241,12 +241,18 @@ class MultiNiftiMasker(NiftiMasker, _utils.CacheMixin):
 
         self._reporting_data = None
         if self.reports:  # save inputs for reporting
-            self._reporting_data = {"mask": self.mask_img_, "images": imgs}
+            self._reporting_data = {
+                "mask": self.mask_img_,
+                "images": imgs,
+                "multi_subject": False,
+            }
             if imgs is not None:
                 imgs = imgs[0] if isinstance(imgs, list) else imgs
                 imgs, dim = compute_middle_image(imgs)
                 self._reporting_data["dim"] = dim
                 self._reporting_data["images"] = imgs
+            if dim == 5:
+                self._reporting_data["multi_subject"] = True
 
         # If resampling is requested, resample the mask as well.
         # Resampling: allows the user to change the affine, the shape or both.
--- a/nilearn/maskers/multi_nifti_masker.py
+++ b/nilearn/maskers/multi_nifti_masker.py
@@ -174,7 +174,6 @@ class MultiNiftiMasker(NiftiMasker, _utils.CacheMixin):
         )
         self.n_jobs = n_jobs
         self._shelving = False
-        self._reporting_data = None
 
     def fit(self, imgs=None, y=None):
         """Compute the mask corresponding to the data.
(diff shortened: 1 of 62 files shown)


ymzayek:
This fails because the multi subject maskers don't support report generation. The same failure can be seen on main if calling `generate_report` on a multi masker object. So the question is do we want to support this for the multi subject maskers? Either way I think the error needs to be improved so I can do that in this PR and can open another issue to see if we would want to support this.

bthirion:
Indeed, we should have report generation for multi subject maskers.

bthirion:
> Ok I tried a few things to make this PR work and not break anything but I think a better strategy might be to better support report generation for the multi subject maskers before doing this. This is because the user could subscript the 5D list and `generate_report` will work as intended and trying to add a warning message about reports in `fit` will probably seem weird to users

Makes sense.

nilearn/maskers/multi_nifti_masker.py
```diff
@@ -240,8 +240,15 @@ def fit(self, imgs=None, y=None):
...
         if self.reports:  # save inputs for reporting
-            imgs = imgs[0] if isinstance(imgs, list) else imgs
-            self._reporting_data = {"images": imgs, "mask": self.mask_img_}
+            self._reporting_data = {"mask": self.mask_img_, "images": imgs}
+            if imgs is not None:
+                imgs = imgs[0] if isinstance(imgs, list) else imgs
+                dim = image.load_img(imgs).shape
+                if len(dim) == 4:
+                    # compute middle image from 4D series for plotting
+                    imgs = image.index_img(imgs, dim[-1] // 2)
```
bthirion, line 249: Can this be factored in a function ? I guess this pattern already exists in the code base ?

nilearn/maskers/tests/test_multi_nifti_labels_masker.py
```diff
@@ -355,6 +355,41 @@ def test_multi_nifti_labels_masker_resampling():
         )
 
 
+def test_multi_nifti_labels_masker_list_of_sample_mask():
```
bthirion: This is something else than the main topic of this PR (?)

nilearn/maskers/tests/test_multi_nifti_maps_masker.py
```diff
@@ -273,6 +273,39 @@ def test_multi_nifti_maps_masker_resampling():
         assert fmri11_img_r.shape == (masker.maps_img_.shape[:3] + (length,))
 
 
+def test_multi_nifti_maps_masker_list_of_sample_mask():
```
bthirion: Same question here.

bthirion (commented):
Technically, the PR looks good.

nilearn/maskers/tests/test_multi_nifti_labels_masker.py
ymzayek: Hm that's weird. It's not in the diff of the latest commit. Maybe you looked at the diff of the merge commit. I think if you check the latest you won't see it.

nilearn/maskers/multi_nifti_masker.py
ymzayek: Yes probably, let me look into it

ymzayek:
I definitely still need to add tests to this but I think it is finally in good shape for what it's intended to do

ymzayek:
@Remi-Gau have you seen similar failures to the ones on windows here? I can't easily see how they are related but they are consistently failing so there is definitely something there

Remi-Gau:
All of those are related to the use of the `write_tmp_imgs` context manager?

Remi-Gau:
I think that we should probably start removing this context manager in the tests and replace it with a just creating the files in the `tmp_path`.

ymzayek:
Seems to also happen with `write_fake_fmri_data_and_design`. Ok I'll see if rewriting the tests without the context manager resolves this

Remi-Gau:
I am struggling to see what in the PR would start triggering those failures though.

ymzayek:
Ok tests related to `write_tmp_imgs` are not failing anymore with the refactoring. `write_fake_fmri_data_and_design` seems to still be causing issues

bthirion (approved):
LGTM.

Remi-Gau (approved):
good with me

PR #3939, opened 2023-08-30
[MAINT] refactor design matrix tests

Remi-Gau:
- Closes https://github.com/nilearn/nilearn/issues/3938

Changes proposed in this pull request:

- refactor tests for design matrix and experimental paradigm

Issue #3937, opened 2023-08-30
[DOC] Replace examples' block splitter symbols with code cell syntax


ymzayek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed suggestion in detail.

Currently the examples in our example gallery use >= 20 `#` symbols to split sections. This can be replaced with either `#%%` or `# %%` with no effect on the rendering of built examples. These alternative syntax have the extra benefit of being recognized as code cells which in some IDEs allow users to run them like jupyter notebooks from the python file.

Reference: https://sphinx-gallery.github.io/stable/syntax.html#embed-rest-in-your-example-python-files

### List any pages that would be impacted.

This would need to be changed in all the examples in https://github.com/nilearn/nilearn/tree/main/examples

Remi-Gau:
that would be nice!

PR #3943, opened 2023-08-30
[ENH] Handle GLM designs with null or nan durations / onsets

### LINKED ISSUE #3884: [BUG] plot_event cannot handle events with 0 duration

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 22.04
- Windows 11

### Python version

- [X] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8
- [ ] 3.7

### nilearn version

0.10.2.dev

### Expected behavior

events with zero duration should be displayed by "delta functions"

Figure_2

compare to when setting event duration to 1 second

Figure_1

```python
from nilearn.plotting import plot_event
import pandas as pd
import numpy as np

onsets = [
    0.0,   2.4,   8.7,   11.4,  15.0,  18.0,  20.7,  23.7,  26.7,  29.7
]
trial_types = ["foo" for _ in onsets]

# with duration = 1 sec
durations = np.ones_like(onsets)
events = pd.DataFrame(
    {
        "trial_type": trial_types,
        "onset": onsets,
        "duration": durations,
    }
)
plot_event(events)
plt.show()

# with duration = 0 sec
durations = np.zeros_like(onsets)
events = pd.DataFrame(
    {
        "trial_type": trial_types,
        "onset": onsets,
        "duration": durations,
    }
)
plot_event(events)
plt.show()
```

Remi-Gau:
- Closes #3884 

- add tests to make sure that events dataframe with nan onsets or durations throw errors and update code to make tests pass
- add tests to throw warning when some events have null duration
- update names of nilearn.glm.tests._utils public functions
- refactor check_events from nilearn.glm.first_level into smaller functions
- update example to show 0 duration events


Remi-Gau:
to be merged after #3939 is merged

Remi-Gau:
```python
from nilearn.plotting.matrix_plotting import (
    plot_event,
)

from nilearn.glm.tests._utils import design_with_null_durations

plot_event(design_with_null_durations())
plt.show()
```

### Before

no_null

### After

with_null

If anyone has suggestion on how to test this, I am all ears.

bthirion (commented):
LGTM so far, thx.

bthirion:
> If anyone has suggestion on how to test this, I am all ears.

Maybe having this occur once in the doc is a good idea (rather than testing).

ymzayek (commented):
LGTM! While reviewing this, I was looking at `plot_event` function example in the docs and in the user guide there is a link to an example that shows its use but the wrong example is linked. It is a small change so maybe it can be done here.

This is referring to this section https://nilearn.github.io/dev/glm/first_level_model.html#event-based in the paragraph right after the bullet points where this example is linked [Decoding of a dataset after GLM fit for signal extraction](https://nilearn.github.io/dev/auto_examples/02_decoding/plot_haxby_glm_decoding.html#sphx-glr-auto-examples-02-decoding-plot-haxby-glm-decoding-py) but it should be this one [Generate an events.tsv file for the NeuroSpin localizer task ](https://nilearn.github.io/dev/auto_examples/04_glm_first_level/plot_write_events_file.html#generate-an-events-tsv-file-for-the-neurospin-localizer-task)

examples/04_glm_first_level/plot_write_events_file.py
```diff
@@ -19,27 +19,25 @@
 
 print(__doc__)
 
-#########################################################################
+# %%
```
Remi-Gau, line 22: implements @ymzayek suggestion: https://github.com/nilearn/nilearn/issues/3937

examples/04_glm_first_level/plot_write_events_file.py
```diff
@@ -71,13 +69,21 @@
...
+# except for button response
+# that are modeled as events with instantaneous duration (0 second).
+
+null_duration_trials = [2, 3, 4, 5]
+durations = []
+for i in trial_type_idx:
+    if i in null_duration_trials:
+        durations.append(0)
+    else:
+        durations.append(1)
```
Remi-Gau, line 84: @bthirion 
I took the liberty to modify this example to show the 0 duration.
Is that OK or would you prefer to stay true to what was done in the paper?

Remi-Gau:
> LGTM! While reviewing this, I was looking at `plot_event` function example in the docs and in the user guide there is a link to an example that shows its use but the wrong example is linked. It is a small change so maybe it can be done here.
> 
> This is referring to this section https://nilearn.github.io/dev/glm/first_level_model.html#event-based in the paragraph right after the bullet points where this example is linked [Decoding of a dataset after GLM fit for signal extraction](https://nilearn.github.io/dev/auto_examples/02_decoding/plot_haxby_glm_decoding.html#sphx-glr-auto-examples-02-decoding-plot-haxby-glm-decoding-py) but it should be this one [Generate an events.tsv file for the NeuroSpin localizer task ](https://nilearn.github.io/dev/auto_examples/04_glm_first_level/plot_write_events_file.html#generate-an-events-tsv-file-for-the-neurospin-localizer-task)

Done. Thanks for catching this one.

examples/04_glm_first_level/plot_write_events_file.py
```diff
@@ -51,9 +49,9 @@
     7, 1, 0, 0, 4, 1, 9, 8, 4, 9, 9
 ]
 # fmt: on
-trial_type_idx = np.array(trial_type_idx)
+trial_type_idx = trial_type_idx
```
ymzayek, line 52: Do we need this line then?

Remi-Gau: fixed

examples/04_glm_first_level/plot_write_events_file.py
bthirion: No pb.

bthirion (approved):
Please add a whatsnew entry, even if this is a limited change.
Thx,

Remi-Gau:
> Please add a whatsnew entry, even if this is a limited change. Thx,

OK I was not 100% sure about this one. Will do.

doc/changes/latest.rst
```diff
@@ -54,6 +54,7 @@ Enhancements
 Changes
 -------
 
+- :func:`~glm.first_level.experimental_paradigm.check_events` will now throw a warning if some events have a 0 second duration and will throw an error if an event has ``NaN`` onset or duration (:gh:`3943` by `Rémi Gau`_).
```
Remi-Gau, line 57: :func:`~glm.first_level.experimental_paradigm.check_events`

and

:func:`~glm.first_level.check_events`

give me the same error.

Am I missing something obvious?

ymzayek: `check_events` is not exposed in the docs so there is nothing to link to. I'm not sure if it is supposed to be part of the public API or not. If so, it would need to be added to `doc/modules/glm.rst` first before you can use that role

bthirion: I wouldn't put it in the public API.

ymzayek: Then we just refer to it with the backticks without the func role

Remi-Gau: Wait... Just to make sure I understand. 

What is the rationale of saying it is not in the public API but check_design_matrix is when both of them are found this init file? 

https://github.com/nilearn/nilearn/blob/main/nilearn/glm/first_level/__init__.py#L24

ymzayek: As the codebase stands I wouldn't take what is in the init file for what is in the public API and supposed to be exposed to users since we know there is heterogeneity in the way "private" functions are written and also, we can have public API of a private module so that would be ok to include in the init. I agree though that given the refactoring we are doing regarding this, check_events should either be in a file named with a leading underscore or, since it is only used in design_matrix.py, it can be moved there and renamed with a leading underscore. I don't see that, following this convention, it would then be a problem if it is imported in a module's init file since it'll be clear from the import statement, which would then have a leading underscore, that it is private. This is what I understand from [PEP8](https://peps.python.org/pep-0008/#public-and-internal-interfaces), though we may also want to make a decision on that aspect if it's confusing.

Remi-Gau: OK thanks for the clarification.

OK so for the moment it really feels the public API is whatever the doc days is in the public API. Hopefully things should get clearer once we clean things up.

For now I will use the backticks without the func role.

PR #3945, opened 2023-09-01
[FIX] Allow both vmin and threshold argument values to be used in plotting functions using `colorscale`

### LINKED ISSUE #3944: [BUG] In plot_to_surf vmin is set to 0 despite not setting a threshold

### Operating system

- [ ] Linux
- [X] Mac
- [ ] Windows

### Operating system version

Mac OS Version 13.4.1 "Ventura"

### Python version

- [ ] 3.11
- [ ] 3.10
- [ ] 3.9
- [X] 3.8
- [ ] 3.7

### nilearn version

Dev version in #3942 

### Expected behavior

The minimum value in the colorbar should be set by `vmin` when the user does not provide a `threshold`.

This is what I got:

```python
UserWarning:

choosing both vmin and a threshold is not allowed; setting vmin to 0
```

I assume `threshold` has a numerical value by default (rather than `None`) and that's why `vmin` is set to 0.

```python
# Paste your code here
cmap = hm_viridis_onesided
# cmap = "viridis"
vmax = 4
# vmin = vmax/data.shape[1] * 60
vmin = 2
event_count = masker.transform(op.join(out_dir, "data_event_count_avg.nii.gz"))
event_count = np.squeeze(event_count)


texture = vol_to_surf(utils.color_rois(out_dir, mask, event_count), fsaverage.pial_left,interpolation='nearest',radius =1, n_samples=1)
surf_plot1=plotting.plot_surf_roi(fsaverage.infl_left, texture, hemi='left', cmap = cmap, colorbar=True, vmin = vmin, vmax = vmax, bg_map=fsaverage.sulc_left, engine="plotly")
```

ymzayek:
- Closes #3944

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -34,6 +34,8 @@ Fixes
 
 - Fix ``fit_transform`` behavior to match when ``fit`` method is passed image data (:gh:`3897` by `Yasmin Mzayek`_)
 
+- Allow using both vmin and threshold with  "plotly" engine to be consistent with "matplotlib" behavior (:gh:`3945` by `Yasmin Mzayek`_)
+
 Enhancements
 ------------
 
--- a/nilearn/plotting/js_plotting_utils.py
+++ b/nilearn/plotting/js_plotting_utils.py
@@ -80,11 +80,6 @@ def colorscale(cmap, values, threshold=None, symmetric_cmap=True,
     if symmetric_cmap and vmin is not None:
         warnings.warn('vmin cannot be chosen when cmap is symmetric')
         vmin = None
-    if threshold is not None:
-        if vmin is not None:
-            warnings.warn('choosing both vmin and a threshold is not allowed; '
-                          'setting vmin to 0')
-        vmin = 0
     if vmax is None:
         vmax = abs_values.max()
     # cast to float to avoid TypeError if vmax is a numpy boolean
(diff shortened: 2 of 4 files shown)


bthirion (commented):
LGTM, but how can we make sure that the behavior is consistent across backends and for the many parameter settings ?

ymzayek:
@bthirion I guess it is tested on a higher level in some cases where `engine` is parameterized in the pytests. I can see if it makes sense to add such a test for this particular case. On a more general level I assume they are consistent by design and, where they diverge, for reasons of available functionality, this is usually documented in the docstrings. I wouldn't be too worried about this but might be something to look into.

ymzayek:
There is an external issue related to the failing jobs that has just been reported to github community forum: https://github.com/orgs/community/discussions/65912

bthirion:
> @bthirion I guess it is tested on a higher level in some cases where `engine` is parameterized in the pytests. I can see if it makes sense to add such a test for this particular case. On a more general level I assume they are consistent by design and, where they diverge, for reasons of available functionality, this is usually documented in the docstrings. I wouldn't be too worried about this but might be something to look into.

Sounds reasonable, but I'm afraid you're a bit optimistic regarding the consistency of the codebase ;-)

Remi-Gau:
> @bthirion I guess it is tested on a higher level in some cases where `engine` is parameterized in the pytests. I can see if it makes sense to add such a test for this particular case. On a more general level I assume they are consistent by design and, where they diverge, for reasons of available functionality, this is usually documented in the docstrings. I wouldn't be too worried about this but might be something to look into.

I guess that I share that optimism and in the worst case scenario, we'll handle inconsistencies when we become aware of them.

Remi-Gau (approved):
good with me

bthirion (approved):
I can't sport any big issue, so this is an OK.

Remi-Gau:
Thanks @ymzayek

Issue #3628, opened 2023-03-29
make sure public / private functions in _utils to start without / with a leading underscore


Remi-Gau:
From what I understood anything in the _utils folder is considered "private" so the leading underscore does not have to be systematic

Yes I'd say that that is the case

> I am completely fine trying to homogenize this to all leading-underscore which would be nice because private / non-private status of the function would be in the filename rather than in the parent folder where its module can be found.

Indeed let's follow up on that in a separate issue as it's not really related to this PR

_Originally posted by @ymzayek in https://github.com/nilearn/nilearn/pull/3525#discussion_r1151533681_

Remi-Gau:
I suspect it'd be worth checking that all functions in any other module called `_utils` also starts with a leading underscore.

jeromedockes:
wouldn't that hint the function is private to the current module?

Remi-Gau:
> wouldn't that hint the function is private to the current module?

That may be because I am still a python noob but I am not sure I understand the nuance you are making.

Can you tell me what you mean by "private to the current module"? Or point to a something t

Remi-Gau:
related to #3319

jeromedockes:
yes I think so!
basically repeating the same thing as what you said with another example:

```
nilearn
├── image.py             # part of public API
├── __init__.py
├── maskers              # part of public API
│   ├── __init__.py
│   ├── nifti_masker.py  # part of public API
│   └── _validation.py   # private to the maskers module
└── _utils.py            # private to the nilearn module
```

inside of `maskers._validation.py`:

```python
import numpy as np # not part of the public API

__all__ = ["check_mask_img", "ValidationError"] # all symbols in the public API


def check_mask_img(mask_img):
    """Public API of _validation module

    can be used in nifti_masker module
    but not the image module (which cannot import maskers._validation),
    unless maskers/__init__.py imports it and lists it in __all__ 
    to make it part of the maskers module's public API
    """
    return _check_mask_shape(mask_img) and _check_mask_values(mask_img)


def _check_mask_shape(mask_img):
    """Private internal of _validation, cannot be used in nifti_masker"""


def _check_mask_values(mask_img):
    """Private internal of _validation, cannot be used in nifti_masker"""


class ValidationError(Exception):
    """Public API of _validation module"""


class _Validator:
    """Private internal of the _validation module"""

    def validate(self, img):
        """Public API of _Validator"""

    def _validate_shape(self, img):
        """Private internal of the _Validator class.

[...]

PR #3946, opened 2023-09-01
[DOC] Include guidelines for private functions in contributing docs

ymzayek:
- Addresses part of #3628

I just copied the example from the issue with @jeromedockes cited as the source at the bottom of the example (but it won't render in the html). IS it ok?

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -290,7 +290,7 @@ The main conventions we enforce are :
 - meaningful variable names
 - function names are underscore separated (e.g., ``a_nice_function``) and as short as possible
 - public functions exposed in their parent module's init file
-- private function names preceded with a "_" and very explicit
+- private function names preceded with a "_" and very explicit, see also :ref:`private_functions`
 - classes in CamelCase
 - 2 empty lines between functions or classes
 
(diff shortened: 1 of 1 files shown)


Remi-Gau (approved):
Good with me

bthirion (approved):
LGTM. Very useful, indeed !

CONTRIBUTING.rst
```diff
@@ -351,6 +351,75 @@ This is also useful for writing unit tests.
 Writing small functions is not always possible, and we do not recommend trying to reorganize larger,
 but well-tested, older functions in the codebase, unless there is a strong reason to do so (e.g., when adding a new feature).
 
+.. _private_functions:
+
+Guidelines for Private Functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In some cases when private functions are in a private module (filename beginning with an underscore),
+but are used outside of that file, we do not name them with a leading underscore. Example:
```
jeromedockes, line 360: thanks! maybe this summary here could read something like:

we start a name with a leading underscore to indicate that it is an internal implementation detail, not to be accessed directly from outside, of the enclosing context: the parent module (for a submodule name), the module (for the name of a top-level function, class or global variable), or the class (for a method or attribute name). Moreover, modules explicitly declare their api through the `__all__` attribute, and any name not listed in `__all__` should not be accessed from outside the module

Remi-Gau:
Sorry @jeromedockes I merged this too quickly: will take your comment in a separate issue, so we do not forget about it

PR #3949, opened 2023-09-04
[ENH] Improve plotting contours for `PlotlySurfaceFigure` objects by adding `add_contours` method

### LINKED ISSUE #3711: [ENH] improving plot_surf_contours

Currently, the plotting.plot_surf_contours() function changes the faces of the outer edge of an ROI to a color. As a consequence, it is impossible to make lines wider or change many other things about this plot. Especially on higher resolution meshes, the contours are very thin. At todays nilearn office hours Yasmin, Jerome, and I discussed some ideas. Below is a messy first pass at moving this whole idea over to plotly. 

### Benefits to the change

This would add the flexibility of using plotly to plotting contours on surfaces. In general, matplotlib isn't the best for plotting surfaces and the current plot_surf_contours() function lacks desirable features, like making the contour more visible. 

```python
from nilearn import datasets, surface, plotting 
import numpy as np 
from scipy.spatial import distance_matrix
import plotly.graph_objects as go

data = datasets.fetch_localizer_button_task(legacy_format=False)
stat_img = data.tmap

destrieux_atlas = datasets.fetch_atlas_surf_destrieux()

parcellation = destrieux_atlas['map_right']

# say this is the region we want to outline
regions_dict = {b'G_postcentral': 'Postcentral gyrus'}

# get indices in atlas for these labels
regions_indices = [
    np.where(np.array(destrieux_atlas['labels']) == region)[0][0]
    for region in regions_dict
]

labels = list(regions_dict.values())

# this function is an adaptation of a function called by the current plot_surf_contours
def get_vertices_on_edge(faces, parc_idx):
    '''
    Function for identifying which vertices lie on the outer
    edge of the parcellation defined by the indices in parc_idx.
    Parameters
    ----------
    faces : numpy.ndarray of shape (n, 3), indices of the mesh faces
    parc_idx : numpy.ndarray, indices of the vertices
        of the region to be plotted
    '''
    # count how many vertices belong to the given parcellation in each face
    verts_per_face = np.isin(faces, parc_idx).sum(axis=1)

    # test if parcellation forms regions
    if np.all(verts_per_face < 2):
        raise ValueError('Vertices in parcellation do not form region.')

    vertices_on_edge = np.intersect1d(np.unique(faces[verts_per_face == 2]),
                                      parc_idx)
    
    return vertices_on_edge


coords, faces = surface.load_surf_mesh(fsaverage.infl_right)

roi_indices = np.where(parcellation == regions_indices)[0]

vertices_on_edge = get_vertices_on_edge(faces, roi_indices)

vertices = coords[vertices_on_edge]
# now that we know where to draw the lines, we need to know in which order. If we pick a vertex to start and 
# move to the closest one, and then to the closest remaining one and so forth, we should get the whole ROI

# Start with the first vertex
current_vertex = 0
visited_vertices = set([current_vertex])

sorted_vertices = [vertices[0]]

# Loop over the remaining vertices in order of distance from the current vertex
while len(visited_vertices) < len(vertices):
    remaining_vertices = np.array([vertex for vertex in range(len(vertices)) if vertex not in visited_vertices])
    remaining_distances = distance_matrix(vertices[current_vertex].reshape(1, -1), vertices[remaining_vertices])
    closest_index = np.argmin(remaining_distances)
    closest_vertex = remaining_vertices[closest_index]
    visited_vertices.add(closest_vertex)
    sorted_vertices.append(vertices[closest_vertex])
    # Move to the closest vertex and repeat the process
    current_vertex = closest_vertex

# at the end we append the first one again to close the outline
sorted_vertices.append(vertices[0])
sorted_vertices = np.asarray(sorted_vertices)

# now plot 
p = plotting.plot_surf_stat_map(fsaverage.infl_right, texture, hemi='right', 
                                title="title", colorbar=True, threshold=0.001, 
                                bg_map=fsaverage.sulc_right, view='lateral', engine='plotly')

p.figure.add_traces(data=go.Scatter3d(x=sorted_vertices[:,0], y=sorted_vertices[:,1], z=sorted_vertices[:,2],
                                      mode='lines', line=dict(width=15, color='white')))


p.show(renderer='jupyterlab')

```

psadil:
- Closes #3711 

- Add `add_contours` method to `PlotlySurfaceFigure`, increasing customizability of contour lines. 
- Added example of the method in [plot_3d_map_to_surface_projection](examples/01_plotting/plot_3d_map_to_surface_projection.py)


ymzayek:
@psadil thanks for getting started on this! Will try to start a review this week

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +114,136 @@ def savefig(self, output_file=None):
         self._check_output_file(output_file=output_file)
         if self.figure is not None:
             self.figure.write_image(self.output_file)
+
+    def add_contours(self, roi_map, levels=None, labels=None, lines=None):
```
Remi-Gau, line 118: probably a good idea to add tests for this method and its errors

psadil:
Tests added! 

Do these seem like they're on the right track? At least coverage doesn't drop in the same way.

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +113,141 @@ def savefig(self, output_file=None):
...
+        ----------
+        roi_map : str or :class:`numpy.ndarray` or list of
+            :class:`numpy.ndarray` ROI map to be displayed on the surface
+            mesh, can be a file (valid formats are .gii, .mgz, .nii,
+            .nii.gz, or Freesurfer specific files such as .annot or .label),
+            or a Numpy array with a value for each vertex of the surf_mesh.
+            The value at each vertex one inside the ROI and zero inside ROI,
+            or an integer giving the label number for atlases.
+
+        levels : list of integers, or None, optional
```
Remi-Gau, line 131:
```suggestion
        levels : list of integers, or None, default=None
```

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +113,141 @@ def savefig(self, output_file=None):
...
+            A list of indices of the regions that are to be outlined.
+            Every index needs to correspond to one index in roi_map.
+            If None, all regions in roi_map are used.
+
+        labels : list of strings or None, or None, optional
```
Remi-Gau, line 136:
```suggestion
        labels : list of strings or None, or None,  default=None
```

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +113,141 @@ def savefig(self, output_file=None):
...
+            A list of labels for the individual regions of interest. Provide
+            None as list entry to skip showing the label of that region. If
+            None, no labels are used.
+
+        lines : list of dict giving the properties of the contours, or None,
+            optional. For valid keys, see
```
Remi-Gau, line 142:
```suggestion
             default=None. For valid keys, see
```

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +113,141 @@ def savefig(self, output_file=None):
...
+            :attr:`plotly.graph_objects.Scatter3d.line`. If length 1, the
+            properties defined in that element will be used to draw all
+            requested contours.
+        """
+        import plotly.graph_objects as go
```
Remi-Gau, line 147: is this import here for a specific reason (avoid circular import for example) otherwise you can definitely put it at the top of the module

Remi-Gau:
> Tests added!

Thanks!!

> Do these seem like they're on the right track?

Had a very quick look: it does help also understand how these methods are used, so definitely useful

nilearn/plotting/displays/_figures.py
psadil: Importing plotly at the top caused this workflow to fail: https://github.com/nilearn/nilearn/blob/1759fa2f719e8eb56627698484f8665cc5336824/.github/workflows/testing.yml#L12-L21. I think that workflow brings in just these packages: https://github.com/nilearn/nilearn/blob/1759fa2f719e8eb56627698484f8665cc5336824/pyproject.toml#L77-L84. Is there a better way to import?

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +113,141 @@ def savefig(self, output_file=None):
...
+            If None, all regions in roi_map are used.
+
+        labels : list of strings or None, or None, optional
+            A list of labels for the individual regions of interest. Provide
+            None as list entry to skip showing the label of that region. If
+            None, no labels are used.
+
+        lines : list of dict giving the properties of the contours, or None,
+            optional. For valid keys, see
+            :attr:`plotly.graph_objects.Scatter3d.line`. If length 1, the
```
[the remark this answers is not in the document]
psadil: Oh, sorry. I left `plotly` out of the intersphinx mapping. I think b681315 will fix that warning.

nilearn/plotting/displays/_figures.py
Remi-Gau: oops my bad.

In this case you want to use a try block for more gracefull failures in case the import fails if plotly is not around

https://github.com/nilearn/nilearn/blob/1759fa2f719e8eb56627698484f8665cc5336824/nilearn/plotting/surf_plotting.py#L252C1-L258C31

Remi-Gau: > n this case you want to use a try block

Actually this may be an overkill given that this is part of the a class where the constructor already has such a check.

psadil: I suppose there could be a try/except at the top of the file, replacing all instances of `import plotly.graph_objects` in the methods? If the import error should still be raised within the constructor, then perhaps the approach used in the tests could be replicated https://github.com/nilearn/nilearn/blob/1759fa2f719e8eb56627698484f8665cc5336824/nilearn/plotting/tests/test_surf_plotting.py#L30-L35

e.g., 
```python
if not PLOTLY_INSTALLED:
    raise ImportError("Plotly is required to use `PlotlySurfaceFigure`.")

```

nilearn/plotting/displays/_figures.py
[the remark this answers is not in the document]
ymzayek: Yes that is needed indeed. If it still doesn't work just reference "the line property of :class:`plotly.graph_objects.Scatter3d`". That should at least work

nilearn/plotting/displays/_figures.py
Remi-Gau: yeah I had seen this one

Wonder if some of this should not be extracted into a separate utility function `is_plotly_installed()`. @ymzayek WDYT?

this should probably be done in a different PR though, so don't overthing it for this PR

ymzayek: Sure it doesn't hurt but yea I would leave it for another PR. Either the current way or what's suggested [here](https://github.com/nilearn/nilearn/pull/3949#discussion_r1317441527) is fine for this PR.

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +113,141 @@ def savefig(self, output_file=None):
...
+
+        Parameters
+        ----------
+        roi_map : str or :class:`numpy.ndarray` or list of
+            :class:`numpy.ndarray` ROI map to be displayed on the surface
+            mesh, can be a file (valid formats are .gii, .mgz, .nii,
+            .nii.gz, or Freesurfer specific files such as .annot or .label),
+            or a Numpy array with a value for each vertex of the surf_mesh.
+            The value at each vertex one inside the ROI and zero inside ROI,
+            or an integer giving the label number for atlases.
```
ymzayek, line 129: parameter type definition and description need to go on separate lines. Also a multi-line type definition needs a `\` at the end of the line. See https://github.com/nilearn/nilearn/blob/1759fa2f719e8eb56627698484f8665cc5336824/nilearn/glm/first_level/first_level.py#L504-L508

nilearn/plotting/tests/test_surf_plotting.py
ymzayek, line 1: I suggest to add one line docstrings to the new tests

psadil: How does c256ed9 look (well, after ignoring typos covered in 84f9885)?

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +113,141 @@ def savefig(self, output_file=None):
...
+
+        levels : list of integers, or None, default=None
```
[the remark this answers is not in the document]
psadil: I think [9d12271](https://github.com/nilearn/nilearn/pull/3949/commits/9d122711281e6a88ae2ed816a6dffbe048039b2e) add all remaining cross references?

nilearn/plotting/displays/_figures.py
psadil: [e3fc2bb](https://github.com/nilearn/nilearn/pull/3949/commits/e3fc2bbc29538febc31e7f9fa71a1b762cae444f) does it

ymzayek:
Thanks @psadil for the updates! Here is the built example: https://output.circle-artifacts.com/output/job/9e7e0100-bdac-4eaf-99d5-b5d2f512e8bd/artifacts/0/dev/auto_examples/01_plotting/plot_3d_map_to_surface_projection.html#sphx-glr-auto-examples-01-plotting-plot-3d-map-to-surface-projection-py

It looks good! There is a small skip that I see in the postcentral gyrus contour. Idk if that can be improved further. And the other contour is hard to see but I guess that is to showcase the line width feature.

Also the API reference looks good: https://output.circle-artifacts.com/output/job/9e7e0100-bdac-4eaf-99d5-b5d2f512e8bd/artifacts/0/dev/modules/generated/nilearn.plotting.displays.PlotlySurfaceFigure.html#nilearn.plotting.displays.PlotlySurfaceFigure

ymzayek:
BTW while checking this I noticed that in the same example where `plot_surf_contours` is used the contours are not rendered anymore. This is not related to this PR as I can also see it in the build artifacts dating back at least a month. I'll open a separate issue

psadil:
> It looks good! There is a small skip that I see in the postcentral gyrus contour. Idk if that can be improved further. And the other contour is hard to see but I guess that is to showcase the line width feature.

Oh dear, thanks. I think this might be a rendering issue. But let me find a way to either confirm or fix. Aiming for the next couple of days.

bthirion (approved):
LGTM besides the last glitch mentioned by  @ymzayek

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +118,141 @@ def savefig(self, output_file=None):
         self._check_output_file(output_file=output_file)
         if self.figure is not None:
             self.figure.write_image(self.output_file)
+
+    def add_contours(self, roi_map, levels=None, labels=None, lines=None):
+        """
+        Draw boundaries around roi.
```
ymzayek, line 124: Just for consistency

```suggestion
        """Draw boundaries around roi.
```

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +118,141 @@ def savefig(self, output_file=None):
...
+        Parameters
+        ----------
+        roi_map : :obj:`str` or :class:`numpy.ndarray` or :obj:`list` of \
+                  :class:`numpy.ndarray`
+            ROI map to be displayed on the surface
+            mesh, can be a file (valid formats are .gii, .mgz, .nii,
+            .nii.gz, or FreeSurfer specific files such as .annot or .label),
+            or a Numpy array with a value for each vertex of the surf_mesh.
+            The value at each vertex one inside the ROI and zero inside ROI,
+            or an :obj:`int` giving the label number for atlases.
```
ymzayek, line 135: ??

```suggestion
            The value at each vertex is one inside the ROI and zero outside
            the ROI, or an :obj:`int` giving the label number for atlases.
```

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +118,141 @@ def savefig(self, output_file=None):
...
+                    mode="lines",
+                    line=line,
+                    name=label,
+                )
+            )
+        self.figure.add_traces(data=traces)
+
+    def _get_vertices_on_edge(self, parc_idx):
+        """
+        Identify which vertices lie on the outer edge of a parcellation.
```
ymzayek, line 189:
```suggestion
        """Identify which vertices lie on the outer edge of a parcellation.
```

nilearn/plotting/displays/_figures.py
```diff
@@ -103,3 +118,141 @@ def savefig(self, output_file=None):
...
+
+        Parameters
+        ----------
+        parc_idx : :class:`numpy.ndarray`
+            Indices of the vertices of the region to be plotted.
+
+        Returns
+        -------
+        data : :class:`numpy.ndarray`
```
ymzayek, line 198:
```suggestion
        sorted_vertices : :class:`numpy.ndarray`
```

ymzayek (commented):
Mostly small doc stuff. I think this is in a good enough state and we can merge after addressing these. A follow-up PR can be done to address the small skip to close the contour. I don't think it should block this

ymzayek:
Oh can you also add a whatsnew entry in `doc/changes/latest.rst`?

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -134,11 +134,15 @@
 ##############################################################################
 # Display outlines of the regions of interest on top of a statistical map
 # -----------------------------------------------------------------------
+#
+# Regions can be outlined using both engines.
```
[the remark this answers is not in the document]
Remi-Gau: yes a more explicit error message would be a good thing for users

psadil:
> A follow-up PR can be done to address the small skip to close the contour. I don't think it should block this

Just to give an update, I'm now suspecting a bug, though I'm not sure of the source. Plotting the `sorted_vertices` with markers shows that one vertex was missing from the postcentral gyrus contour. This vertex that is not in the contour ends up in the second-to-last row of `sorted_vertices` array, and the skip seems to be related to how the lines are drawn when trying to account for this other vertex (one edge of the gap is row in `sorted_vertices` that is one before the one that is missing from the contour.)

newplot

So, I'm hoping to sort this out before the pull request is merged.

ymzayek:
I see in that case we'll hold off until that is resolved. Thanks @psadil

psadil:
Here's a wireframe view of the surface that clarifies the situation:

frame

Vertices in the roi are marked with circles. The circles are orange when they are on the edge of the boundary (defined as a vertex of a triangle inside the roi and on a face with only two vertices in the roi) and otherwise purple. 

Building up the boundary edge by sequentially selecting the next closest vertex ends up skipping the orange vertex that is lowest in the figure (the one with only one orange neighbor). 

Is there an expectation for how to trace this boundary? I see these options
- Color every edge between boundary vertices, and only those edges (would lead to faces whose edges are only partially colored)
- Color every edge between boundary vertices, and only those edges, but exclude vertices that have only one neighbor on the graph (some vertices would be missing from an roi)
- Color the outer edges of every face that has at least two vertices in the roi (could lead to the boundaries of two rois overlapping)

ymzayek:
@psadil first thought would be to go with the third option since it seems most conservative, however the visual consequence of overlapping boundaries is not clear. We'd need to test it.

psadil:
Sorry, but hoping to check another implementation detail. I wonder how to handle drawing handle regions that do not share an edge with another vertex in the roi. For example, this region has 2 isolated vertices. 

newplot

Should I follow the matplotlib approach? I think that would mean ignoring the isolated vertices.

image

psadil:
> @psadil thanks for checking that. I think then it would be best to stay consistent with the matplotlib approach and ignore isolated vertices. That would then be more like the 2nd option you describe in [#3949 (comment)](https://github.com/nilearn/nilearn/pull/3949#issuecomment-1739623696)?

It wouldn't quite be the 2nd option, because it looks like the current matplotlib approach colors the faces around vertices that share only one edge with another vertex in the ROI. For example, here is the post-central gyrus contour in matplotlib

image

In the plotly implementation, I think that a comparable boundary could be the one traced by the purple line in the following (the vertex with only one shared edge is included in the contour)

newplot

bthirion:
I don't know if there are conventions somewhere that explain what is best. If there aren't I would go for the easiest and most maintainable solution (and possibly a warning if some vertices have to be missed).

ymzayek:
I see, thanks @psadil for the clear explanation. There are certainly pros and cons to each approach so if we make it clear what the contours include in the docstring and consider @bthirion 's last comment I think we can move forward with one of these implementations.

psadil:
Thanks for all the input. I'll work on something that tries to match the matplotlib approach (unless someone encounters another convention in the meanwhile). 

FWIW, the behavior of [plot_surf_roi](https://nilearn.github.io/stable/modules/generated/nilearn.plotting.plot_surf_roi.html#nilearn.plotting.plot_surf_roi) includes isolated vertices. But hopefully a clear explanation of how the contours are generated (and warnings about when an isolated vertex is excluded) will avoid too much confusion.

roi27

roi28

ymzayek:
@psadil would you have the bandwidth to finish this in the next month or two? It would be great to include it in the major release we are planning for January. Thanks for all your work on this!

psadil:
Sorry for being _so_ slow on this. Is there still a need for this functionality?

man-shu:
@Remi-Gau @bthirion  I think this could be an interesting futher improvement to Surface functionality. We should add this to #4027

Remi-Gau:
> @Remi-Gau @bthirion I think this could be an interesting futher improvement to Surface functionality. We should add this to #4027

I could have a look. We have not tested the new Surface object with plotly but yeah in general I'd like to be able to use the features in this PR.

Remi-Gau:
@psadil 
let me know if you need help with the merge conflicts

psadil:
> @psadil
> let me know if you need help with the merge conflicts

🙏🏻  It wasn't pretty, but I merged main and it looks like the diff isn't too messy.

examples/01_plotting/plot_3d_map_to_surface_projection.py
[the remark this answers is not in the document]
psadil: For the suggestion to raise an error in `plot_surf_contours` if a object is passed, which argument(s) should be checked? `figure`?

bthirion (approved):
LGTM, thx !

examples/01_plotting/plot_3d_map_to_surface_projection.py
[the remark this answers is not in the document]
man-shu: I think both `figure` and `axes` should be checked somewhere [here](https://github.com/psadil/nilearn/blob/faae8454d1e4badae60479025ce482263d7c3aa2/nilearn/plotting/surf_plotting.py#L972)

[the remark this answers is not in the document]
psadil: Sounds good, thanks! As discussed, now implemented here: https://github.com/nilearn/nilearn/pull/3949/commits/22b1aef0cbdbaa34b8107d594209edce30669927

nilearn/plotting/surf_plotting.py
```diff
@@ -969,6 +970,12 @@ def plot_surf_contours(surf_mesh, roi_map, axes=None, figure=None, levels=None,
     nilearn.surface.vol_to_surf : For info on the generation of surfaces.
 
     """
+    if isinstance(figure, PlotlySurfaceFigure):
+        raise ValueError('figure argument is a PlotlySurfaceFigure'
+                         'but it should be None or a matplotlib figure')
+    if isinstance(axes, PlotlySurfaceFigure):
+        raise ValueError('axes argument is a PlotlySurfaceFigure'
+                         'but it should be None or a matplotlib axes')
```
man-shu, line 978: We should also add tests for these errors.

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -136,18 +136,35 @@
...
+                                legend=True,
+                                colors=['g', 'k'])
+    plotting.show()
+elif engine == "plotly":
+    figure.add_contours(
+        roi_map=parcellation,
+        levels=regions_indices,
+        labels=labels,
+        lines=[{"width": 5}])
+    figure.show()
```
man-shu, line 166: I also noticed that the build_docs workflow was stuck because the plotly figure is rendered in a browser window, which needs to be closed for the workflow to move forward.

We could comment this out for the moment. But @Remi-Gau might have a better solution.
```suggestion
    # figure.show()
```

nilearn/plotting/surf_plotting.py
psadil: Oops! Sorry. Here: c998dc7

man-shu: No worries! Thanks!

examples/01_plotting/plot_3d_map_to_surface_projection.py
Remi-Gau: Don't have a fix for this yet but there is an issue for it.
https://github.com/nilearn/nilearn/issues/4379

Commenting it out has been my way to deal with it for now.
So you are on track!

nilearn/plotting/displays/_figures.py
```diff
@@ -104,3 +130,354 @@ def savefig(self, output_file=None):
...
+            faces_w_one_v = np.flatnonzero(verts_per_face == 1)
+            unique_v, unique_v_counts = np.unique(
+                self._faces[faces_w_one_v], return_counts=True
+            )
+            isolated_v = unique_v[unique_v_counts == 6]
+            if any(isolated_v):
+                warnings.warn(
+                    f"""{label=} contains isolated vertices:
+                    {isolated_v.tolist()}. These will not be included in ROI
+                    boundary line."""
```
Remi-Gau, line 223:
```suggestion
                    boundary line.""", stacklevel=2
```

having a stacklevel higher than the default will help the users know where it's coming from in their code and not in our code:

https://www.youtube.com/watch?v=CtFdXBEwYfk

psadil: Oh! Thanks. I wasn't aware of this.

Remi-Gau: No worries. A lot of warning in the Nilearn code base suffer from that problem so it shows you are not the only one not to know this.

nilearn/plotting/displays/_figures.py
```diff
@@ -104,3 +130,355 @@ def savefig(self, output_file=None):
...
+                self._faces[faces_w_one_v], return_counts=True
+            )
+            isolated_v = unique_v[unique_v_counts == 6]
+            if any(isolated_v):
+                warnings.warn(
+                    f"""{label=} contains isolated vertices:
+                    {isolated_v.tolist()}. These will not be included in ROI
+                    boundary line.""",
+                    stacklevel=2,
+                )
```
man-shu, line 225: Codecov is complaining about testing the outputs in a few places.

nilearn/plotting/displays/_figures.py
```diff
@@ -104,3 +130,355 @@ def savefig(self, output_file=None):
...
+    ) -> bool:
+        a1 = x1 - x2
+        b1 = x3 - x4
+        c1 = x1 - x3
+        a2 = y1 - y2
+        b2 = y3 - y4
+        c2 = y1 - y3
+        d = a1 * b2 - a2 * b1
+        if np.isclose(d, 0):
+            return False
```
man-shu, line 471: here as well...

man-shu (commented):
Thanks for resolving all the comments @psadil ! But just a few more tests :)
And then we can add this to the changelog, see: https://nilearn.github.io/stable/development.html#changelog

nilearn/plotting/displays/_figures.py
psadil: Done! Thanks

nilearn/plotting/displays/_figures.py
psadil: Done! Thanks

psadil:
It looks like those checks for NPY201 are in different parts of the repo?

Remi-Gau:
> It looks like those failures for NPY201 are being triggered by different parts of the repo?

yes that's not you

I will merge the main branch in your PR to solve this: remember to pull first before you push again 😉

nilearn/plotting/displays/_figures.py
```diff
@@ -104,3 +130,355 @@ def savefig(self, output_file=None):
...
+            # that may be slightly farther away -- if the one that is
+            # farther away has fewer vertices within the roi.
+            if len(remaining_vertices) > 5:
+                smallest_5_idx = np.argpartition(
+                    remaining_distances.squeeze(), 5
+                )[:5]
+                xy1 = self._transform_coord_to_plane(
+                    centroids[current_vertex], *vs[current_vertex]
+                )
+                next_index = 9999
```
bthirion, line 352: is this initialization safe ?

nilearn/plotting/displays/_figures.py
```diff
@@ -104,3 +130,355 @@ def savefig(self, output_file=None):
...
+        current_vertex = 0
+        visited_vertices = {current_vertex}
+        last_distance = np.inf
+        prev_first = 0
+
+        sorted_vertices = [centroids[0]]
+
+        # Loop over the remaining vertices in order of distance from the
+        # current vertex
+        while len(visited_vertices) < len(centroids):
```
bthirion, line 327: Can we make sure that the loop ends ?
What algorithm is implemented here ?

nilearn/plotting/displays/_figures.py
```diff
@@ -104,3 +130,355 @@ def savefig(self, output_file=None):
...
+        p_ = point - t0
+
+        p_normal = np.dot(p_, n) * n
+        p_tangent = p_ - p_normal
+
+        closest_point = p_tangent + t0
+        return closest_point + elevation * n
+
+    @staticmethod
+    def _do_segs_intersect(
```
bthirion, line 453: Can you write a short docstring ?

nilearn/plotting/displays/_figures.py
```diff
@@ -104,3 +130,355 @@ def savefig(self, output_file=None):
...
+
+        vertices_on_edge = np.intersect1d(
+            np.unique(self._faces[verts_per_face == 2]), parc_idx
+        )
+        faces_outside_edge = np.isin(self._faces, vertices_on_edge).sum(axis=1)
+
+        return np.logical_and(faces_outside_edge > 0, verts_per_face < 3)
+
+    @staticmethod
+    def _project_above_face(point, t0, t1, t2, elevation=0.1):
```
bthirion, line 438: can you write a 1-line docstring ?

bthirion (commented):
Looks mostly good. A little effort on pedagogy would help.

nilearn/plotting/displays/_figures.py
psadil: Oh, I suppose it would be a problem if there were 9999 regions within a parcellation--changing this to -1

nilearn/plotting/displays/_figures.py
psadil: It's a modification of the approach listed in the original issue: https://github.com/nilearn/nilearn/issues/3711#issue-1675696871, with the modifications aiming to ensure that the contour does not cross itself. I'm not sure how it could fail to end, but if preferred I think I could update it to a for loop.

nilearn/plotting/displays/_figures.py
psadil: [a50d656](https://github.com/nilearn/nilearn/pull/3949/commits/a50d65639b9195f40bee9ee6d2abb46e7536c22b)

nilearn/plotting/displays/_figures.py
psadil: Yes, thanks [bb21ae7](https://github.com/nilearn/nilearn/pull/3949/commits/bb21ae756e6f8fa48133bab7a6d486c33a7170bb)

nilearn/plotting/displays/_figures.py
psadil: Yes, thanks [a86d7ea](https://github.com/nilearn/nilearn/pull/3949/commits/a86d7ea75d3c5abd620c7f0e21352a44e061d7d2)

nilearn/plotting/displays/_figures.py
bthirion: I'd more comfortable with a for loop indeed...

psadil: Sounds good! I cleaned up the sorting procedure a bit and added a few more explanatory comments cb8ad06.

Remi-Gau:
will fix the merge conflicts

psadil:
> will fix the merge conflicts

Great, thanks! I'm still monitoring this, so let me know if changes are needed.

Also, sorry that this ended up being a rather complicated solution. I would not be offended if it were decided that this belonged in some external, unofficial package :)

Remi-Gau:
> let me know if changes are needed.

I need to take the time to do a full review... My bad.

man-shu (approved):
Sorry for the delay @psadil ! But this LGTM. 
Let's just wait for the checks to pass.

Remi-Gau:
will fix the style issues

man-shu:
Merging!

PR #3952, opened 2023-09-06
[DOC] improve section on private functions

### LINKED ISSUE #3947: improve summary in contribution documentation regarding private "functions"
> thanks! maybe this summary here could read something like:
> we start a name with a leading underscore to indicate that it is an internal implementation detail, not to be accessed directly from outside, of the enclosing context: the parent module (for a submodule name), the module (for the name of a top-level function, class or global variable), or the class (for a method or attribute name). Moreover, modules explicitly declare their interface through the `__all__` attribute, and any name not listed in `__all__` should not be accessed from outside the module

_Originally posted by @jeromedockes in https://github.com/nilearn/nilearn/pull/3946#discussion_r1313826407_

Remi-Gau:
- Closes #3947

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -360,7 +360,7 @@ We start a name with a leading underscore to indicate that it is an internal imp
 not to be accessed directly from outside, of the enclosing context:
 
 - the parent module (for a submodule name),
-- the module (for the name of a top-level function, class or global variable),
+- or the module (for the name of a top-level function, class or global variable),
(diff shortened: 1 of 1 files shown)


bthirion (approved):
This makes it more explicit. Thx.

ymzayek (approved):
That's better thanks!

Remi-Gau:
I just copy pasted @jeromedockes words so he should get the credit on this one

PR #3965, opened 2023-09-12
[MAINT] Update handling of matplotlib ax overwrite deprecation

ymzayek:
- Closes none

~~Resets handling of matplotlib deprecation. With the newest matplotlib v3.7.3 release, this change is causing the CI failures. I have to see if I handled it incorrectly or if it's a new bug on their side. For the time being, will merge this PR so the CI passes~~

Edit: see comment below


ymzayek:
Ok maybe it was fine the way it was but I didn't use good arguments for `_compare_version`

PR #3973, opened 2023-09-13
[FIX] Set edgecolors when plotting contours so that they are rendered

ymzayek:
- Closes #3954

PR #3976, opened 2023-09-14
[INFRA] Test minimum supported version of matplotlib for min_requirements testing

ymzayek:
- Closes None

See https://github.com/nilearn/nilearn/pull/3973#issuecomment-1719365341

### DIFF
--- a/build_tools/github/install.sh
+++ b/build_tools/github/install.sh
@@ -6,7 +6,7 @@ if [ ! -z "$MIN_REQUIREMENTS" ]; then
     # See pyproject.toml for dependency group options
     if [ ! -z "$MATPLOTLIB" ]; then
         # Include plotting dependencies too
-        pip install --progress-bar off --upgrade -e .[min,plotting,test]
+        pip install --progress-bar off --upgrade -e .[min,plotting_min,test]
     else
         pip install --progress-bar off --upgrade -e .[min,test]
     fi
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,7 @@ min = [
 plotly = ["kaleido", "plotly"]
 # Necessary req to use nilearn's plotting module
 plotting = ["matplotlib>=3.3.0"]
+plotting_min = ["matplotlib==3.3.0"]
 # For running unit and docstring tests
 test = [
     "coverage",


ymzayek:
Maybe it's overkill but shouldn't we also test the `latest` job in `testing.yml` without matplotlib and plotly (maybe sticking to only latest python version)?

Remi-Gau:
> Maybe it's overkill but shouldn't we also test the `latest` job in `testing.yml` without matplotlib and plotly (maybe sticking to only latest python version)?

maybe I am missing something, but isn't that pretty much what the tests on "nilearn[min]" do?

https://github.com/nilearn/nilearn/blob/0c2782696ee129c9ab0fd2a2b93be0ad64710a64/.github/workflows/testing.yml#L13

ymzayek:
> maybe I am missing something, but isn't that pretty much what the tests on "nilearn[min]" do?
> 
> https://github.com/nilearn/nilearn/blob/0c2782696ee129c9ab0fd2a2b93be0ad64710a64/.github/workflows/testing.yml#L13

So the min_requirements job tests minimum versions of

min = [
    "joblib==1.0.0",
    "nibabel==3.2.0",
    "numpy==1.19.0",
    "pandas==1.1.5",
    "scikit-learn==1.0.0",
    "scipy==1.6.0"
]

The min_requirements_matplotlib job additionally takes plotting = ["matplotlib **>=** 3.3.0"] so it doesn't install the minimum version of matplotlib (it ends up installing 3.5.3).
This PR suggests to install matplotlib **==** 3.3.0 for this job.

The second proposition here https://github.com/nilearn/nilearn/pull/3976#issuecomment-1719530132 is to test latest versions of everything without matplotlib and plotly, which we don't do.

Hope that's clear!

bthirion (approved):
LGTM.

Remi-Gau:
> So the min_requirements job tests minimum versions of
> 
> min = [ "joblib==1.0.0", "nibabel==3.2.0", "numpy==1.19.0", "pandas==1.1.5", "scikit-learn==1.0.0", "scipy==1.6.0" ]
> 
> The min_requirements_matplotlib job additionally takes plotting = ["matplotlib **>=** 3.3.0"] so it doesn't install the minimum version of matplotlib (it ends up installing 3.5.3). This PR suggests to install matplotlib **==** 3.3.0 for this job.
> 
> The second proposition here [#3976 (comment)](https://github.com/nilearn/nilearn/pull/3976#issuecomment-1719530132) is to test latest versions of everything without matplotlib and plotly, which we don't do.
> 
> Hope that's clear!

Very clear thanks.

Though I as thinking that it may be easier to "read" with the proper "matrix" in CI, rather than different set of dependencies in the pyproject.toml.

Not sure how often users (or dev) actually use those different extra requirements. If they are only used in testing in CI may be better to declare them there.

Remi-Gau (approved):

ymzayek:
@Remi-Gau I see what you mean and we can restructure it if there's  a set up that makes more sense but for the developer it's nice to just be able to `pip install -e [min]` in a venv if you need to debug on the minimum versions. I've certainly used it before. So the use case would be to quickly replicate the ci environment to debug

Remi-Gau:
> @Remi-Gau I see what you mean and we can restructure it if there's a set up that makes more sense but for the developer it's nice to just be able to `pip install -e [min]` in a venv if you need to debug on the minimum versions. I've certainly used it before. So the use case would be to quickly replicate the ci environment to debug

OK if you use it once in a while let's keep in the pyproject.toml. I like to use makefile to keep track of my repo specific "dev recipes", so I can really blame any dev who makes themselves more at home in a repo. 🤗

ymzayek:
Ok not a bad approach. I don't mind relying on makefile commands and we can certainly add some recipes to the nilearn Makefile to "replace" the environments in pyproject.toml

Remi-Gau:
my bad: that was not a suggestion. 
more a comment saying: I rarely use those dedicated extra requirements when debugging but if they make your life easier, then that's justification enough for me to have them.

PR #3981, opened 2023-09-15
[INFRA] Use latest Ubuntu machines for CI

ymzayek:
- Closes None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ parameters:
 jobs:
     host_docs:
         machine:
-            image: ubuntu-2004:202111-01
+            image: ubuntu-2004:current
         environment:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -15,7 +15,7 @@ parameters:
 jobs:
     host_docs:
         machine:
-            image: ubuntu-2004:202111-01
+            image: ubuntu-2004:current
```
Remi-Gau, line 18:
```suggestion
            image: ubuntu-2204:current
```

shouldn't it be more something like this?

https://circleci.com/developer/machine/image/ubuntu-2204

ymzayek: Oops yes that should be it

Remi-Gau (approved):
Good to go!

PR #3987, opened 2023-09-19
[FIX] Use `maps` key for API consistency and improve deprecation for new behavior of some atlas fetchers

### LINKED ISSUE #3983: atlas fetchers return inconsistency
Minor API detail but I ran across a user who was confused by this.

Several fetchers seem to return dictionary-like object with a `map` key instead of `maps` like our doc suggest they should: https://nilearn.github.io/stable/development.html#additional-cases

May be a good thing to keep it consistent.

- fetch_atlas_craddock_2012
- fetch_atlas_smith_2009
- fetch_atlas_basc_multiscale_2015

ymzayek:
- Closes #3983

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -44,6 +44,8 @@ Fix bug in method ``transform_imgs`` of :class:`~maskers.MultiNiftiMapsMasker` a
 
 - Set edgecolors to be the same as facecolors in :func:`~plotting.plot_surf_contours` so that ROI contours are rendered clearly (:gh:`3973` by `Yasmin Mzayek`_)
 
+- Refactor deprecation of behavior of :func:`~datasets.fetch_atlas_craddock_2012`, :func:`~datasets.fetch_atlas_smith_2009` and :func:`~datasets.fetch_atlas_basc_multiscale_2015` fetchers including using API consistent ``maps`` key (:gh:`3987` by `Yasmin Mzayek`_)
+
 Enhancements
 ------------
 
--- a/examples/00_tutorials/plot_3d_and_4d_niimg.py
+++ b/examples/00_tutorials/plot_3d_and_4d_niimg.py
@@ -47,7 +47,7 @@ plotting.plot_stat_map(tmap_filename, threshold=3)
 #
 # We can download resting-state networks from the Smith 2009 study on
 # correspondence between rest and task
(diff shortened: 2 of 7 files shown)


Remi-Gau (approved):
Looks great!! Thanks @ymzayek for this.

Remi-Gau:
Do you think we should update the doc (in a separate PR) to say that some functions 'fetch_coords*' return not maps but `rois`?

https://github.com/nilearn/nilearn/blob/ecf7686740f236274be94fe8a0757808e366debc/nilearn/datasets/atlas.py#L1624

Currently our doc mentions only fetch_atlas.

bthirion:
Maybe add a brief whatsnew entry ?

ymzayek:
> Do you think we should update the doc (in a separate PR) to say that some functions 'fetch_coords*' return not maps but `rois`?
> 
> https://github.com/nilearn/nilearn/blob/ecf7686740f236274be94fe8a0757808e366debc/nilearn/datasets/atlas.py#L1624
> 
> Currently our doc mentions only fetch_atlas.

Sure but where exactly? I might be wrong but other than the API reference we only talk about the fetch_atlas function return here: https://nilearn.github.io/stable/development.html#how-to-contribute-an-atlas

If we were to add your suggestion maybe we could think about adding a section on fetchers in https://nilearn.github.io/stable/manipulating_images/input_output.html#input-and-output-neuroimaging-data-representation or another appropriate area in the user guide.

Let me know if you had a different idea.

Remi-Gau:
> Sure but where exactly? I might be wrong but other than the API reference we only talk about the fetch_atlas function return here: https://nilearn.github.io/stable/development.html#how-to-contribute-an-atlas

next to this section was what I had in mind

ymzayek:
Perhaps we should open a separate issue to discuss further

> next to this section was what I had in mind

Ok the idea is to explain about contributing such fetchers? 

> Do you think we should update the doc (in a separate PR) to say that some functions 'fetch_coords*' return not maps but rois?

I agree with this but I understood it as more of a usage thing rather than relating to contributing. Sorry if I'm misunderstanding!

Remi-Gau:
> we should open a separate issue to discuss further

yeah we were getting side tracked.

done.

ymzayek:
Thanks merging

PR #3990, opened 2023-09-19
[INFRA] Configure doc build CI to always host artifacts

ymzayek:
- Closes None

With the current set up, a failure on the doc build prevents circleci from hosting the docs that have been built. This is annoying since the htmls exist but cannot be easily viewed (sometimes for very trivial errors) and thus hinders the review process. This PR changes the configuration so that the docs are always hosted if a build exists. Note that deploy will still only run on main only with a successful doc build.

I've tested this on my fork on [main](https://github.com/ymzayek/nilearn) and in a [PR](https://github.com/ymzayek/nilearn/pulls)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
                     if [[ "$STATUS" == "success" ]]; then
                         echo "Doc build is successful. Deploying docs..."
                     else
-                        echo "Failed to deploy because doc build status is failure"
+                        echo "Failed to deploy because doc build status is 'failure'"
                         exit 1
                     fi
         -   run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,9 @@ parameters:
     GITHUB_RUN_URL:
         type: string
         default: none
+    STATUS:
+        type: string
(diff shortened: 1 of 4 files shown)


bthirion:
This PR is a bit cryptic to me. Any way to check that it works OK ?

ymzayek:
@bthirion I've run several tests on my fork on [main](https://github.com/ymzayek/nilearn) and in a [PR](https://github.com/ymzayek/nilearn/pulls)

ymzayek:
> At some point we are going to need a diagram how the CI for the doc works. I am still not as good as you are juggling with it.

Yes happy to discuss it and explain why we do it this way. We can also think about how to document it better for devs

Remi-Gau:
> > At some point we are going to need a diagram how the CI for the doc works. I am still not as good as you are juggling with it.
> 
> Yes happy to discuss it and explain why we do it this way. We can also think about how to document it better for devs.

Can we use mermaid, pretty please?!

```mermaid
flowchart LR
    build-docs.yml --> build_type.sh
```

# 😃

ymzayek:
I don't know it but sure happy to try it out!

PR #3993, opened 2023-09-19
[ENH] Allow setting `vmin` in `plot_glass_brain` and `plot_stat_map`

### LINKED ISSUE #1149: Adding back in vmin to plotting colormap command

### LINKED ISSUE #3084: Adding a vmin option to plot_stat_map plot_glass_brain functions

michellewang:
- Closes #3084
- Closes #1149

- Made `plot_glass_brain` use its `vmin` argument instead of silently ignoring it
  - This involved changing `_get_colorbar_and_data_ranges`, which changes the behaviour of `plot_stat_map` to also allow `vmin` to be set (instead of throwing an error)
- Updated docstring for `symmetric_cbar`, since the colorbar will no longer be restricted to `(-vmax, vmax)`, `(0, vmax)`, `(-vmax, 0)`, and the colormap range will also change depending on `vmin`/`vmax`
- Removed `symmetric_data_range` argument from `_get_colorbar_and_data_ranges`, which was added in #3873 to preserve the old behaviour in these functions. We can still get the symmetric data range behaviour if:
  - `symmetric_cbar` is `True`, or
  - `symmetric_cbar` is `'auto'`, `vmin` or `vmax` is `None`, and the data has both positive and negative values
    - This means that a call with default arguments will plot a symmetric colorbar, same as before
- Refactored test cases for `_get_colorbar_and_data_ranges` to have clearer expected results
- Added some example glass brain plots in `examples/01_plotting/plot_demo_glass_brain_extensive.py`
  - Other glass brain examples still show the same plots

Example `plot_glass_brain` calls with `vmin`:

```python

plotting.plot_glass_brain(
    stat_img,
    colorbar=True,
    display_mode="lyrz",
    threshold=2,
    cmap='viridis',
    vmin=2,                  # use the whole colormap range even if there is a threshold
    symmetric_cbar=False,
)
```

image

```python
plotting.plot_glass_brain(
    stat_img,
    colorbar=True,
    display_mode="lyrz",
    threshold=2,
    cmap='viridis',
    vmin=-5,
    vmax=8,
    plot_abs=False,
)
```

image

Example `plot_stat_map` calls with `vmin`:

```python

plotting.plot_stat_map(
    stat_img,
    colorbar=True, 
    threshold=3,
    vmin=-9,   # symmetric from -9 to 9
)
```

image

```python
plotting.plot_stat_map(
    stat_img,
    colorbar=True, 
    threshold=3,
    vmin=0,             # non-negative values only
    vmax=10,
    symmetric_cbar=False,
    cmap='viridis',
)
```

image

```python
plotting.plot_stat_map(
    stat_img,
    colorbar=True, 
    threshold=3,
    vmin=3,            # use the whole colormap range even if there is a threshold
    vmax=10,
    symmetric_cbar=False,
    cmap='viridis',
)
```

image


ymzayek:
Thx @michellewang for working on this! The example plots look great! Will give a more thorough review soon.

ymzayek (commented):
Overall all, looks pretty good! Can you explain the refactoring concerning the plotting of absolute values?

nilearn/plotting/displays/_slicers.py
```diff
@@ -461,15 +451,36 @@ def _map_show(
...
+                    vmax=kwargs.get("vmax"),
+                )
+
                 im = display_ax.draw_2d(
                     data_2d, data_bounds, bounding_box, type=type, **kwargs
                 )
                 ims.append(im)
         return ims
 
+    def _threshold(self, data, threshold=None, vmin=None, vmax=None):
```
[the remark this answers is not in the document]
michellewang: Yes it applies the same logic (for the same reason as explained [here](https://github.com/nilearn/nilearn/pull/3873#discussion_r1301785684)), though `surf_plotting._threshold` returns the mask and this returns the masked data array. Do we want to refactor them to use the same function? I wasn't sure how best to do that since the surface plotting functions do not seem to use the Nilearn slicers

michellewang:
> Can you explain the refactoring concerning the plotting of absolute values?

Yes, I had to move the thresholding because it needs to happen after we take the absolute value of the data (which happens in `GlassBrainAxes.transform_to_2d`). Otherwise, when `plot_abs` is `True`, `vmin` is set to `0`, and the new thresholding function will mask all negative values in the data (if a non-zero threshold is defined), so the resulting plot will only show values that were positive in the original data instead of the absolute value of all values.

I'll remove the lines with `data_selection = np.abs(data_selection)` in `nilearn/plotting/displays/_axes.py` because after looking at them again I think they are redundant with what the code already did.

nilearn/plotting/displays/_slicers.py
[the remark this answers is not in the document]
ymzayek: > Do we want to refactor them to use the same function?
At this point I'd say no, your reasoning is good.

ymzayek:
> Also: are there any tests for slicers and `_map_show`? I couldn't find any from a quick search

Slicers are tested in `nilearn/plotting/tests/test_displays.py` but `_map_show` is not tested directly (only through calls of `add_contours` and `add_overlay`). I'd suggest adding a test relevant to the changes you made.

nilearn/_utils/docs.py
```diff
@@ -871,16 +871,13 @@ def custom_function(vertices):
...
-
-        The colormap will always range from `-vmax` to `vmax`.
-
+symmetric_cbar : :obj:`bool`, or "auto", default="auto"
+    Specifies whether the colorbar and colormap should range from `-vmax` to
+    `vmax` (or from `vmin` to `-vmin` if `-vmin` is greater than `vmax`) or
+    from `vmin` to `vmax`.
+    Setting to `"auto"` will select the former if either 1) `vmin` or `vmax` is
+    None and the image has both positive and negative values, or 2) `vmin` is
+    equal to `-vmax`.
```
bthirion, line 880: If  `vmin`  is   equal to `-vmax`, then the two behaviors are equivalent, no ?

nilearn/plotting/img_plotting.py
```diff
@@ -1073,20 +1077,30 @@ def plot_glass_brain(stat_map_img,
...
     if stat_map_img:
         stat_map_img = _utils.check_niimg_3d(stat_map_img, dtype='auto')
         if plot_abs:
-            cbar_vmin, cbar_vmax, vmin, vmax = _get_colorbar_and_data_ranges(
-                _safe_get_data(stat_map_img, ensure_finite=True),
-                vmax=vmax,
-                symmetric_cbar=symmetric_cbar,
-                force_min_stat_map_value=0)
+            if vmin is not None and vmin < 0:
+                raise ValueError(
```
bthirion, line 1090: Wondering if we should really raise en Error here ? A warning could be enough. Think of a loop creating many figures with `plot_abs` set to `True`, but a `vmin` calculated for each figure iteratively.

nilearn/_utils/docs.py
michellewang: Yes, you're right. Updated the docstring

nilearn/plotting/img_plotting.py
michellewang: Ok, I can force `vmin` to be `0` if it's negative and raise a warning instead of the error. Does that sound okay?

bthirion: Why force vmin to 0 ? The user may have his/her reasons to set it to a given value.

michellewang: Ok I see, that makes sense. Updated to allow arbitrary `vmin` value

bthirion (commented):
LGTM. Can you consider codecov complaints ?

nilearn/plotting/tests/test_img_plotting/test_plot_glass_brain.py
```diff
@@ -118,4 +126,12 @@ def test_plot_glass_brain_with_completely_masked_img(display_mode):
 
 def test_plot_glass_brain_vmin_vmax(img_3d_mni):
     """Smoke tests for plot_glass_brain being passed vmin and vmax."""
-    plot_glass_brain(img_3d_mni, vmin=-2, vmax=2)
+    # set plot_abs=False because otherwise negative vmin is not allowed
+    plot_glass_brain(img_3d_mni, vmin=-2, vmax=2, plot_abs=False)
```
ymzayek, line 130: I think with the recent update vmin is allowed so you can revert the change in this test

nilearn/plotting/tests/test_img_plotting/test_plot_glass_brain.py
```diff
@@ -118,4 +126,12 @@ def test_plot_glass_brain_with_completely_masked_img(display_mode):
...
+
+
+def test_plot_glass_brain_negative_vmin_with_plot_abs(img_3d_mni):
+    """Test that error is thrown when plot_abs is True and vmin is negative."""
```
ymzayek, line 134:
```suggestion
    """Test that warning is thrown if plot_abs is True and vmin is negative."""
```

examples/01_plotting/plot_demo_glass_brain_extensive.py
```diff
@@ -87,6 +87,36 @@
     stat_img, threshold=3, colorbar=True, plot_abs=True, display_mode="yx"
 )
 
+###############################################################################
```
ymzayek, line 90:
```suggestion
# %%
```
for style consistency

examples/01_plotting/plot_demo_glass_brain_extensive.py
```diff
@@ -87,6 +87,36 @@
...
+    colorbar=True,
+    plot_abs=False,
+    display_mode="yz",
+    vmin=0,
+    threshold=2,
+    symmetric_cbar=False,
+    cmap="viridis",
+)
+
+###############################################################################
```
ymzayek, line 107:
```suggestion
# %%
```

ymzayek:
> LGTM. Can you consider codecov complaints ?

Not sure why that is happening. I checked the report and the added code is fully covered and the codecov/patch report is 100% so I think it's ok. Seems like there is a small decrease on the project level but this seems to come from indirect changes. I think @michellewang you can just make the changes from the last review and this should be good to go.

michellewang:
Ok, I was confused by the Codecov report. Just addressed the remaining comments, thanks @ymzayek and @bthirion for reviewing!

ymzayek (approved):
Just FYI something went wrong with the merge so I redid it. This PR LGTM, thanks @michellewang!

bthirion (approved):
LGTM, thx.

nilearn/plotting/displays/_slicers.py
```diff
@@ -398,16 +398,6 @@ def _map_show(
...
-        if threshold is not None:
-            data = _safe_get_data(img, ensure_finite=True)
-            if threshold == 0:
-                data = np.ma.masked_equal(data, 0, copy=False)
-            else:
-                data = np.ma.masked_inside(
-                    data, -threshold, threshold, copy=False
-                )
-            img = new_img_like(img, data, img.affine)
-
```
Remi-Gau, line 410: - removing this causes a regression https://github.com/nilearn/nilearn/issues/4255

PR #3994, opened 2023-09-19
[MAINT] check events before plotting them

### LINKED ISSUE #3948: handling of simultaneous events by plot_event
Other "problems" with `plot_event` that I am encountering while working on #3884.

1.  if 2 events have different trial_type but same onset and duration, then only the last one plotted is visible as it is overlayed over the other one

2. design matrix creation will sum the amplitude of 2 "duplicated" events (same trial type, onset, duration), however this won't be apparent when just using the plot_event function.

```python
import numpy as np

from nilearn.glm.first_level import make_first_level_design_matrix
from nilearn.glm.tests._utils import duplicate_events_paradigm
from nilearn.plotting import plot_design_matrix, plot_event
from nilearn.plotting.matrix_plotting import plot_event

events = duplicate_events_paradigm()
events.drop(columns=["modulation"], inplace=True)
print(events)

plot_event(events)
plt.show()

fig, (ax1, ax2) = plt.subplots(figsize=(10, 6), nrows=1, ncols=2)

tr = 1.0
n_scans = 80
frame_times = np.arange(n_scans) * tr

X1 = make_first_level_design_matrix(
    frame_times,
    events,
    drift_model="polynomial",
    drift_order=3,
    hrf_model="spm",
)

plot_design_matrix(X1, ax=ax1)
ax1.set_title("design matrix with duplicated events", fontsize=12)

plt.plot(X1["c0"])
plt.plot(X1["c1"])
ax2.set_title("predicted time course of 'c0' and 'c1'", fontsize=12)

plt.show()
```

```
  trial_type  onset  duration
0         c0     10       1.0
1         c0     30       1.0
2         c0     70       1.0
3         c0     70       1.0  < -- duplicated events
4         c1     10       1.0  < -- "hides" similar events from c0
5         c1     30       1.0  < -- "hides" similar events from c0
```

events

Figure_1

(1) Could be solved but may require plotting events on several "rows" (one per condition).

(2) May be the intended behavior, but if not this could probably be easily solved by running `check_events` on the input of `plot_event` (which may also help with some of the input validation that is missing from this function in anycase: for example, no checking that the required columns are there, do not have NaNs...)

Notes:

- I do find it confusing that a function called `check_events` takes care of summing the modulation of the events duplicates: this behavior should probably extracted in a separate function

- although a warning is sent when events duplicates are detected, it may be more user friendly to have the possibility to drop duplicated events rather than sum their modulation

Remi-Gau:
- Closes #3948

- run check_events on inputs of plot_events to ensure that they get the same treatment as before design matrix generation
- extract summation of modulation of duplicated events in a separate function

### DIFF
--- a/nilearn/glm/first_level/experimental_paradigm.py
+++ b/nilearn/glm/first_level/experimental_paradigm.py
@@ -24,9 +24,6 @@ def check_events(events):
     It is valid if the events data has ``'onset'`` and ``'duration'`` keys
     with numeric non NaN values.
 
-    This function also handles duplicate events
-    by summing their modulation if they have one.
-
     Parameters
     ----------
     events : pandas DataFrame
@@ -34,6 +31,11 @@ def check_events(events):
 
     Returns
     -------
+    events : pandas DataFrame
+        Events data that describes a functional experimental paradigm.
+
+    The dataframe has the following columns:
+
     trial_type : array of shape (n_events,), dtype='s'
         Per-event experimental conditions identifier.
         Defaults to np.repeat('dummy', len(onsets)).
@@ -95,15 +97,7 @@ def check_events(events):
 
     _check_unexpected_columns(events_copy)
 
-    events_copy = _handle_modulation(events_copy)
-
-    cleaned_events = _handle_duplicate_events(events_copy)
-
-    trial_type = cleaned_events["trial_type"].values
-    onset = cleaned_events["onset"].values
-    duration = cleaned_events["duration"].values
-    modulation = cleaned_events["modulation"].values
-    return trial_type, onset, duration, modulation
+    return _handle_modulation(events_copy)
 
 
 def _check_columns(events):
@@ -185,7 +179,8 @@ COLUMN_DEFINING_EVENT_IDENTITY = ["trial_type", "onset", "duration"]
 STRATEGY = {"modulation": "sum"}
 
 
-def _handle_duplicate_events(events):
+def sum_modulation_of_duplicate_events(events):
+    """Sum modulation of duplicate events if they have one."""
     cleaned_events = (
         events.groupby(COLUMN_DEFINING_EVENT_IDENTITY, sort=False)
         .agg(STRATEGY)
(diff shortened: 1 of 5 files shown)


nilearn/glm/first_level/experimental_paradigm.py
```diff
@@ -185,7 +179,8 @@ def _check_unexpected_columns(events):
 STRATEGY = {"modulation": "sum"}
 
 
-def _handle_duplicate_events(events):
+def sum_modulation_of_duplicate_events(events):
```
bthirion, line 182: (Very minor) Slight inconsistency is naming: we leave strategy as a global variable, but call the corresponding function "sum"

bthirion (commented):
A bit on the side of the PR, but I see that some private function in glm/first_level/experimental_paradigm do not even have a one-liner docstring to explain what they do. Maybe we could take this opportunity to add them ?

LGTM otherwise

nilearn/glm/first_level/experimental_paradigm.py
Remi-Gau: to be fair I am not even sure why those globals were used: I guess to allow for other strategies in the future ?

Remi-Gau: Will revert to the older name though I find it less specific regarding what the function actually does.

Remi-Gau:
> do not even have a one-liner docstring

the functions were extracted by turning the comments prefacing them into the function

if the doc string does more than just repeating the function name I will add it one

nilearn/glm/first_level/experimental_paradigm.py
Remi-Gau: done

Remi-Gau:
> if the doc string does more than just repeating the function name I will add it one

done

nilearn/glm/first_level/experimental_paradigm.py
```diff
@@ -185,7 +181,11 @@ def _check_unexpected_columns(events):
 STRATEGY = {"modulation": "sum"}
 
 
-def _handle_duplicate_events(events):
+def handle_modulation_of_duplicate_events(events):
+    """Deal modulation of duplicate events if they have one.
```
ymzayek, line 185:
```suggestion
    """Deal with modulation of duplicate events if they have one.
```

ymzayek (approved):
LGTM!

PR #4012, opened 2023-09-25
[REL] add tags to changelog

Remi-Gau:
- Closes none

- add tags to changelog in view of next release

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -261,12 +261,6 @@ Changelog entries in ``doc/changes/latest.rst`` should adhere to the following c
 - Finish with a link to the PR and the author's profile
 - New contributors to add their details to the ``authors`` section of the ``CITATION.cff`` (see below)
 
-Example entry in ``doc/changes/latest.rst``:
-
-.. code-block:: rst
-
-    - :bdg-dark:`Code` Fix off-by-one error when setting ticks in :func:`~plotting.plot_surf` (:gh:`3105` by `Dimitri Papadopoulos Orfanos`_).
-
 List of badges:
 
 .. codeblock:: rst
@@ -278,6 +272,12 @@ List of badges:
       :bdg-warning:`Test`
       :bdg-danger:`Deprecation`
       :bdg-dark:`Code`
+
+Example entry in ``doc/changes/latest.rst``:
+
+.. code-block:: rst
+
+    - :bdg-dark:`Code` Fix off-by-one error when setting ticks in :func:`~plotting.plot_surf` (:gh:`3105` by `Dimitri Papadopoulos Orfanos`_).
 
 Associated entry in ``CITATION.cff``:
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -263,7 +263,7 @@ Changelog entries in ``doc/changes/latest.rst`` should adhere to the following c
 
 List of badges:
 
-.. codeblock:: rst
+.. code-block:: rst
 
       :bdg-primary:`Doc`
       :bdg-secondary:`Maint`
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -257,7 +257,7 @@ Changelog entries in ``doc/changes/latest.rst`` should adhere to the following c
 
 - Entry in the appropriate category
 - Single line per entry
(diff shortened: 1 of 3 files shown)


doc/changes/0.10.1.rst
```diff
@@ -40,7 +40,7 @@ Enhancements
 
 - :bdg-info:`Plotting` Surface plotting methods no longer automatically rescale background maps, which, among other things, allows to use curvature sign as a background map (:gh:`3173` by `Alexis Thual`_).
 
-- :func:`~glm.first_level.first_level_from_bids` now takes an optional ``sub_labels`` argument and warns users of given subject labels that are not present in the dataset (:gh:`3351` by `Kevin Sitek`_).
+- :bdg-success:`API` :func:`~glm.first_level.first_level_from_bids` now takes an optional ``sub_labels`` argument and warns users of given subject labels that are not present in the dataset (:gh:`3351` by `Kevin Sitek`_).
```
Remi-Gau, line 43: we forgot one

doc/changes/latest.rst
```diff
@@ -2,85 +2,99 @@
...
+HIGHLIGHTS
+----------
+
+- :bdg-info:`Plotting` Volume plotting functions like :func:`~plotting.plot_img` now have an optional ``radiological`` parameter, defaulting to ``False``. If ``True``, this will invert the x-axis and ``L`` and ``R`` annotations to confirm to radiological conventional view. (:gh:`3172` by `Konrad Wagstyl`_ and `Yasmin Mzayek`_).
+
+- :bdg-dark:`Code` Update Decoder objects to use the more efficient ``LogisticRegressionCV`` (:gh:`3736` by `Michelle Wang`_).
+
+- :bdg-success:`API` Add ``LassoCV`` as a new estimator option for Decoder objects (:gh: `3781` by `Michelle Wang`_)
+
+- :bdg-success:`API` Add ``vmin`` and ``symmetric_cbar`` arguments to :func:`~nilearn.plotting.plot_img_on_surf` (:gh:`3873` by `Michelle Wang`_).
```
Remi-Gau, line 19: Can highlights be just repeats of some entries of the changelog or do they have to more generic rephrasings? I think I have seen both precedents.

Also this is my very personal take of what the highlights are. Happy to go with some others.

CONTRIBUTING.rst
```diff
@@ -257,14 +257,27 @@ Changelog entries in ``doc/changes/latest.rst`` should adhere to the following c
...
+
+.. codeblock:: rst
+
+      :bdg-primary:`Doc`
+      :bdg-secondary:`Maint`
+      :bdg-success:`API`
+      :bdg-info:`Plotting`
+      :bdg-warning:`Test`
+      :bdg-danger:`Deprecation`
+      :bdg-dark:`Code`
```
Remi-Gau, line 274: Even if contributors forget them at least we will have them listed somewhere

ymzayek: I originally wanted to put this in https://nilearn.github.io/dev/maintenance.html. The idea was that this would be the responsibility of the maintainer during a release so as to not add an extra "burden" to contributors. But maybe it's ok since people seem to already include it sometimes. WDYT? If we keep it here I still don't see a problem with merging a PR without it and then just checking before a release. This should be easy enough since we have to manipulate the file anyways.

doc/changes/latest.rst
ymzayek: I think we can just use best judgement. Sometimes it makes sense to rephrase or combine related PRs into one highlight. But otherwise just copy and paste is fine IMO

ymzayek: I agree on the highlights

ymzayek (commented):
LGTM!

CONTRIBUTING.rst
ymzayek: Ok let's keep it here then

bthirion (approved):
LGTM so far.

doc/changes/latest.rst
bthirion: So we should focus on #3856 ?

CONTRIBUTING.rst
Remi-Gau: Definitely something we can revisit and change if we see that this approach does not work.

Issue #3812, opened 2023-07-05
Roadmap 0.11.0


Remi-Gau:
To list issues to tackle for the next release.

**See [milestone](https://github.com/nilearn/nilearn/milestone/26)**

## Surface support

- [x] See meta-issue: #4027 

## Dropping python 3.8

- [x] update pyproject.toml
- [x] update CI
- [x] run `pre-commit run -a`

## Support python 3.13

- [x] update pyproject.toml
- [x] update CI

## Follow release protocol

- https://nilearn.github.io/stable/maintenance.html#how-to-make-a-release

ymzayek:
We probably will also need to bump up versions of dependencies and deal with deprecations

Remi-Gau:
> We probably will also need to bump up versions of dependencies and deal with deprecations

added a link to the release protocol that covers this.

ymzayek:
Suggestion TODOs to add:
- [ ] Update python version for building dev docs
- [ ] Update _python_doc_base in conf.py

Remi-Gau:
For reference: 

end of life for 3.8 and release of 3.13: https://devguide.python.org/versions/#status-of-python-versions

Given this and the discussion at the last dev meeting (yesterday), we should aim for a release of Nilearn 0.11.0 towards the end of October.

I have updated the [milestone for it](https://github.com/nilearn/nilearn/milestone/26) to contain only the REQUIRED features or bug fixes. Everything else should be considered nice to have.

PR #4023, opened 2023-10-01
[MAINT] Add python 3.12 to CI

Remi-Gau:
- relates to #3812 

- use python 3.12 in CI

### DIFF
--- a/.github/workflows/detect_test_pollution.yml
+++ b/.github/workflows/detect_test_pollution.yml
@@ -43,7 +43,7 @@ jobs:
         -   name: Setup python
             uses: actions/setup-python@v4
             with:
-                python-version: 3.11
+                python-version: '3.12'
 
         -   name: Display Python version
             run: python -c "import sys; print(sys.version)"
--- a/.github/workflows/detect_test_pollution.yml
+++ b/.github/workflows/detect_test_pollution.yml
@@ -44,6 +44,7 @@ jobs:
             uses: actions/setup-python@v4
             with:
                 python-version: '3.12'
+                allow-prereleases: true
 
         -   name: Display Python version
             run: python -c "import sys; print(sys.version)"
--- a/.github/workflows/detect_test_pollution.yml
+++ b/.github/workflows/detect_test_pollution.yml
@@ -44,7 +44,7 @@ jobs:
             uses: actions/setup-python@v4
             with:
                 python-version: '3.12'
-                allow-prereleases: true
+                allow-prereleases: false
 
         -   name: Display Python version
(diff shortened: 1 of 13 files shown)


.github/workflows/flake8.yml
```diff
@@ -29,7 +29,8 @@ jobs:
         -   name: Setup python
             uses: actions/setup-python@v4
             with:
-                python-version: '3.10'
+                python-version: '3.11'
```
ymzayek, line 32: I haven't looked closely at the flake8 failures related to 3.12 but at some point we'll need to handle these. Perhaps put in a separate issue

ymzayek (commented):
LGTM! Python 3.12 was just released so shouldn't remove the `allow-prereleases` parameter to make sure we're always testing on stable 3.12 releases?

.github/workflows/flake8.yml
Remi-Gau: agreed

they are not massive:

```
./examples/02_decoding/plot_haxby_grid_search.py:127:18: E226 missing whitespace around arithmetic operator
./nilearn/decoding/decoder.py:254:28: E201 whitespace after '{'
./nilearn/decoding/decoder.py:254:40: E231 missing whitespace after ':'
./nilearn/decoding/decoder.py:254:56: E202 whitespace before '}'
./nilearn/interfaces/fmriprep/tests/test_load_confounds.py:583:54: E226 missing whitespace around arithmetic operator
./nilearn/surface/surface.py:961:48: E201 whitespace after '{'
```

strangely though they are not "just" handled by black

ymzayek: Ok maybe just manually handle here and update this workflow to 3.12?

Remi-Gau:
> shouldn't remove the `allow-prereleases` parameter to make sure we're always testing on stable 3.12 releases

yeah why not

.github/workflows/flake8.yml
Remi-Gau: will try

.github/workflows/detect_test_pollution.yml
```diff
@@ -43,7 +43,8 @@ jobs:
         -   name: Setup python
             uses: actions/setup-python@v4
             with:
-                python-version: 3.11
+                python-version: '3.12'
+                allow-prereleases: false
```
Remi-Gau, line 47: set the pre-release to false, eventhough it is the default I feel it can be nice to keep it around for when we want to start checking new python pre-release

bthirion:
This looks good, then ?

Remi-Gau:
> This looks good, then ?

Yup. Merging

PR #4028, opened 2023-10-04
[FIX] Disable extrapolation of out-of-bounds volumes

### LINKED ISSUE #4024: [BUG] Interpolation of high motion volumes at the beginning/end scan

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 22.04

### Python version

- [ ] 3.11
- [ ] 3.10
- [X] 3.9
- [ ] 3.8
- [ ] 3.7

### nilearn version

0.10.2

### Expected behavior

At scrubbing  (censoring high-motion volumes) step, the volumes at the beginning/end of the scan -if labelled as high-motion outliers- shouldn't be interpolated (*extrapolated*) but discarded before filtering.

Currently, the function [_interpolate_volumes()](https://github.com/nilearn/nilearn/blob/b1fa2e0b9161a5055a900b708aed38aaab15866f/nilearn/signal.py#L779) uses a CubicSpline (scipy class) with `extrapolate` feature enabled by default. Asking the interpolation method to predict values outside the range of data available is -in my opinion- not a very good idea. The results of the interpolation, and further scrubbing, are not reliable. 

I am happy to open a PR for this with a fix proposal if there's consensus.

Example of signal interpolation as is and the disabling the `extrapolate` flag,
image
image
image  

Thanks,
Jordi

```python
import numpy as np
from nilearn import signal
import nibabel as nib

data = nib.load("/tmp/xxx_desc-preproc_bold.nii.gz'').get_data()
volumes = np.transpose(data, (3, 0, 1, 2))

# dummy list of censored high-motion volumes
sample_mask = np.zeros(data_reshaped.shape[0], dtype=bool)
sample_mask[0:50] = 1
sample_mask[-50:] = 1
sample_mask = ~sample_mask

volumes = signal._interpolate_volumes(volumes, sample_mask, t_r=1.6)
```

jhuguetn:
- Closes #4024

- include a `extrapolate` parameter in the `signal._interpolate_volumes()` function, exposing the identically named parameter from the Scipy class used underneath for the Splines-based interpolation (see https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.CubicSpline.html#scipy.interpolate.CubicSpline). 
- setting the default `extrapolate` new parameter value to False, disabling the undesired interpolation on out-of-bounds volumes

### DIFF
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -854,9 +854,9 @@ def _interpolate_volumes(volumes, sample_mask, t_r, extrapolate):
         extrapolate_default = (
             "By default the cubic spline interpolator extrapolates "
             "the out-of-bounds censored volumes in the data run. This "
-            "can lead to undesired filtered signal results. In future "
-            "releases, the default strategy will be not to extrapolate "
-            "and discard those volumes at filtering."
+            "can lead to undesired filtered signal results. Starting in "
+            "version 0.13, the default strategy will be not to extrapolate "
+            "but discard those volumes at filtering."
         )
         warnings.warn(
             category=FutureWarning,
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -737,14 +737,9 @@ def clean(
         )
 
     # Interpolation / censoring
-    signals, confounds = _handle_scrubbed_volumes(
+    signals, confounds, sample_mask = _handle_scrubbed_volumes(
         signals, confounds, sample_mask, filter_type, t_r, extrapolate
     )
-    if sample_mask is not None and not extrapolate:
-        # reset the indexing of the sample_mask excluding non-interpolated
-        # volumes at the head of the data
-        sample_mask -= sample_mask[0]
-
     # Detrend
     # Detrend and filtering should apply to confounds, if confound presents
     # keep filters orthogonal (according to Lindquist et al. (2018))
@@ -824,17 +819,25 @@ def _handle_scrubbed_volumes(
 ):
     """Interpolate or censor scrubbed volumes."""
     if sample_mask is None:
-        return signals, confounds
+        return signals, confounds, sample_mask
 
     if filter_type == "butterworth":
         signals = _interpolate_volumes(signals, sample_mask, t_r, extrapolate)
+        # discard non-interpolated out-of-bounds volumes
+        signals = signals[~np.isnan(signals).all(axis=1), :]
         if confounds is not None:
             confounds = _interpolate_volumes(
                 confounds, sample_mask, t_r, extrapolate
             )
+            # discard non-interpolated out-of-bounds volumes
+            confounds = confounds[~np.isnan(confounds).all(axis=1), :]
+        if sample_mask is not None and not extrapolate:
+            # reset the indexing of the sample_mask excluding non-interpolated
+            # volumes at the head of the data
+            sample_mask -= sample_mask[0]
     else:  # Or censor when no filtering, or cosine filter
         signals, confounds = _censor_signals(signals, confounds, sample_mask)
-    return signals, confounds
+    return signals, confounds, sample_mask
 
 
 def _censor_signals(signals, confounds, sample_mask):
@@ -867,8 +870,6 @@ def _interpolate_volumes(volumes, sample_mask, t_r, extrapolate):
     )
     volumes_interpolated = cubic_spline_fitter(frame_times)
     volumes[~sample_mask, :] = volumes_interpolated[~sample_mask, :]
-    # discard non-interpolated out-of-bounds volumes
-    volumes = volumes[~np.isnan(volumes).all(axis=1), :]
     return volumes
 
 
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -547,6 +547,7 @@ def clean(
     high_pass=None,
     t_r=2.5,
     ensure_finite=False,
+    extrapolate=True,
     **kwargs,
 ):
     """Improve :term:`SNR` on masked :term:`fMRI` signals.
@@ -667,6 +668,11 @@ def clean(
         If `True`, the non-finite values (NANs and infs) found in the data
         will be replaced by zeros.
 
+    extrapolate : :obj:`bool`, default=True
+        If `True` and filter='butterworth', censored volumes in both ends of
+        the signal data will be interpolated before filtering. Otherwise, they
+        will be discarded from the band-pass filtering process.
+
     kwargs : dict
         Keyword arguments to be passed to functions called within ``clean``.
         Kwargs prefixed with ``'butterworth__'`` will be passed to
@@ -732,7 +738,7 @@ def clean(
 
     # Interpolation / censoring
     signals, confounds = _handle_scrubbed_volumes(
-        signals, confounds, sample_mask, filter_type, t_r
+        signals, confounds, sample_mask, filter_type, t_r, extrapolate
     )
 
     # Detrend
@@ -810,16 +816,18 @@ def clean(
 
 
 def _handle_scrubbed_volumes(
-    signals, confounds, sample_mask, filter_type, t_r
+    signals, confounds, sample_mask, filter_type, t_r, extrapolate
 ):
     """Interpolate or censor scrubbed volumes."""
     if sample_mask is None:
         return signals, confounds
 
     if filter_type == "butterworth":
-        signals = _interpolate_volumes(signals, sample_mask, t_r)
+        signals = _interpolate_volumes(signals, sample_mask, t_r, extrapolate)
         if confounds is not None:
-            confounds = _interpolate_volumes(confounds, sample_mask, t_r)
+            confounds = _interpolate_volumes(
+                confounds, sample_mask, t_r, extrapolate
+            )
     else:  # Or censor when no filtering, or cosine filter
         signals, confounds = _censor_signals(signals, confounds, sample_mask)
     return signals, confounds
@@ -833,14 +841,29 @@ def _censor_signals(signals, confounds, sample_mask):
     return signals, confounds
 
 
-def _interpolate_volumes(volumes, sample_mask, t_r):
+def _interpolate_volumes(volumes, sample_mask, t_r, extrapolate):
     """Interpolate censored volumes in signals/confounds."""
+    if extrapolate == True:
+        extrapolate_default = (
+            "By default the cubic spline interpolator extrapolates "
+            "the out-of-bounds censored volumes in the data run. This "
+            "can lead to undesired filtered signal results. In future "
+            "releases, the default strategy will be not to extrapolate "
+            "and discard those volumes at filtering."
+        )
+        warnings.warn(
+            category=FutureWarning,
+            message=extrapolate_default,
+        )
     frame_times = np.arange(volumes.shape[0]) * t_r
     remained_vol = frame_times[sample_mask]
     remained_x = volumes[sample_mask, :]
-    cubic_spline_fitter = CubicSpline(remained_vol, remained_x)
+    cubic_spline_fitter = CubicSpline(
+        remained_vol, remained_x, extrapolate=extrapolate
+    )
     volumes_interpolated = cubic_spline_fitter(frame_times)
     volumes[~sample_mask, :] = volumes_interpolated[~sample_mask, :]
+    #volumes = volumes[~np.isnan(volumes).all(axis=1), :]
     return volumes
 
 
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -856,7 +856,7 @@ def _interpolate_volumes(volumes, sample_mask, t_r, extrapolate):
             "the out-of-bounds censored volumes in the data run. This "
             "can lead to undesired filtered signal results. Starting in "
             "version 0.13, the default strategy will be not to extrapolate "
-            "but discard those volumes at filtering."
+            "but to discard those volumes at filtering."
         )
         warnings.warn(
             category=FutureWarning,
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -740,6 +740,10 @@ def clean(
     signals, confounds = _handle_scrubbed_volumes(
         signals, confounds, sample_mask, filter_type, t_r, extrapolate
     )
+    if sample_mask is not None and not extrapolate:
+        # reset the indexing of the sample_mask excluding non-interpolated
+        # volumes at the head of the data
+        sample_mask -= sample_mask[0]
 
     # Detrend
     # Detrend and filtering should apply to confounds, if confound presents
@@ -843,7 +847,7 @@ def _censor_signals(signals, confounds, sample_mask):
 
 def _interpolate_volumes(volumes, sample_mask, t_r, extrapolate):
     """Interpolate censored volumes in signals/confounds."""
-    if extrapolate == True:
+    if extrapolate:
         extrapolate_default = (
             "By default the cubic spline interpolator extrapolates "
             "the out-of-bounds censored volumes in the data run. This "
@@ -863,7 +867,8 @@ def _interpolate_volumes(volumes, sample_mask, t_r, extrapolate):
     )
     volumes_interpolated = cubic_spline_fitter(frame_times)
     volumes[~sample_mask, :] = volumes_interpolated[~sample_mask, :]
-    #volumes = volumes[~np.isnan(volumes).all(axis=1), :]
+    # discard non-interpolated out-of-bounds volumes
+    volumes = volumes[~np.isnan(volumes).all(axis=1), :]
     return volumes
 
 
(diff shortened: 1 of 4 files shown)


ymzayek:
Hi @jhuguetn thanks for opening the PR! Could you also add a test to `nilearn/tests/test_signal.py`?
I'm also wondering if this needs to exposed in the public function `signal.clean` that ends up calling this function? If there is a use case for that then it would be possible to pass it as a butterworth kwarg but then it needs to be added as a parameter to `_handle_scrubbed_volumes`.

nilearn/signal.py
```diff
@@ -731,8 +731,10 @@ def clean(
         )
 
     # Interpolation / censoring
+    extrapolate = kwargs.get('extrapolate', False)
```
htwangtw, line 734: This should be an explicit parameter exposed in `signal.clean` and add documentation on this. 
Since the current behaviour is `extrapolate=True`, we need to warn user about this change in behaviour.
If we want the default behaviour to be `extrapolate=False`, this will need to enter deprecation cycle (@ymzayek please make sure I am correct here)

nilearn/signal.py
```diff
@@ -731,8 +731,10 @@ def clean(
...
     signals, confounds = _handle_scrubbed_volumes(
-        signals, confounds, sample_mask, filter_type, t_r
+        signals, confounds, sample_mask, filter_type, t_r,
+        extrapolate=extrapolate
```
htwangtw, line 737:
```suggestion
        extrapolate
```

nilearn/signal.py
```diff
@@ -810,16 +812,18 @@ def clean(
 
 
 def _handle_scrubbed_volumes(
-    signals, confounds, sample_mask, filter_type, t_r
+    signals, confounds, sample_mask, filter_type, t_r, extrapolate=False
```
htwangtw, line 815:
```suggestion
    signals, confounds, sample_mask, filter_type, t_r, extrapolate
```
Private function should not have default parameter. I don't remember the reason other then my style checker yelled at me before. @Remi-Gau is this is something we should impose?

htwangtw (commented):
This change should be reflected in the parameter of `signal.clean`. 
I raised a few things and would like @ymzayek and @Remi-Gau to have a say. 
Also we need a test for this change

nilearn/signal.py
Remi-Gau: > Private function should not have default parameter. 

Your style checker is complaining about this? Does it point to a "reference" explaining why? That's not a recommendation I am familiar with but then again my python coding style knowledge is that of a trained baboon so...

Also don't show some part of the nilearn code base to your style checker otherwise your screen will light up like its Xmas time because we do have several private functions with default values.

nilearn/signal.py
Remi-Gau: suggestion:
- add the parameter and document it
- do not change default behavior
- in the butterworth case throw a warning when the value of the parameter may lead to the unwanted outcome that @jhuguetn encountered

Remi-Gau:
> Could you also add a test to `nilearn/tests/test_signal.py`?

I agree that we definitely need a test for (preferably not just a smoke test)

nilearn/signal.py
bthirion: Taking this as a commonsense comment: I think that default parameters should not be burried deep in the codebase, but visible in the API so that the behavior is easy to figure out. My 2c,

Remi-Gau: ha yes in that sense I agree: though I am sure we can think of cases where you may not want to have all the parameters of all the private functions exposed in the public API.
I think I was more surprised by a style checker seemingly making some absolute (ish) recommendation on this.

ymzayek (commented):
Initially it seemed that this is a bug fix with limited scope, hence the current solution could be ok. However, it is a behavioral change and to me the impact on user code is hard to tell so I agree with @htwangtw that it is safest to go through a deprecation cycle especially since the faulty implementation did not actually fail any user code IIUC. So I would follow the suggestions as laid out by @htwangtw and @Remi-Gau to expose `extrapolate` in signal clean, keep default behavior, throw a deprecation warning in the appropriate place to tell the user to set `extrapolate=False`. The warning should say that the default behavior will change in release `0.13`. @jhuguetn is it clear?

nilearn/signal.py
jhuguetn: Fine with me. I initially wanted to keep compatibility and impact as less as possible in the current way major/frontend functions are currently defined.

nilearn/signal.py
jhuguetn: Again, the reasoning for doing so was to preserve compatibility with potential usages of the function (not inside the package perhaps, but outside). I am ok with adding the `extrapolate` parameter without a predefined value, specially if we want to make it available to the top-level function `clean()`.

jhuguetn:
Yes, it is @ymzayek. Thanks!

bthirion (commented):
You probably want to add a test to ensure that no extrapolation is performed.

jhuguetn:
> You probably want to add a test to ensure that no extrapolation is performed.  

Yes, I am working on that.   

I have encountered a small bump that perhaps any of you all can help me with. When extrapolation is disabled, scipy `CubicSpline` returns NaNs for those out-of-bounds data points (i.e. volumes to be censored at the ends of the time series). See below example with n=5 first data points being censored:

image

While testing the new `extrapolate` switch parameter inside the `signal.clean()` function -as complex as it is- I've encountered that other steps in the `signal.clean()` workflow do not expect having NaNs in both the signal/confounds discarded (due to avoiding extrapolation) and raise errors:
```
# Pivoting in qr decomposition was added in scipy 0.10
Q, R, _ = linalg.qr(confounds, mode="economic", pivoting=True)
...
ValueError: array must not contain infs or NaNs
```  
To cope with that I've tried dropping the NaNs, but then by design some posterior steps in charge of volume censoring in `signal.clean()` complain as well since censored volumes/samples have already been suppressed beforehand:

```
def _censor_signals(signals, confounds, sample_mask):
      """Apply sample masks to data."""
      signals = signals[sample_mask, :]
...      
IndexError: index 35 is out of bounds for axis 0 with size 35
```
As an alternative, I thought we could assign a fix/hardcoded value temporarily to those NaNs (which would be censored later) or replace them by the original data, but still I don't like the idea since I lose notion of what's been discarded in the interpolation step. In addition, it might have an impact doing so on posterior steps. As an example see below a `signal.clean()` test execution replacing the NaNs by zeroes,

image

I am interested in knowing what you folks think. Thanks!

nilearn/signal.py
htwangtw: No I cannot find PEP8 reference for this. I can see we agree that behaviour changing parameter should be exposed.

htwangtw:
The NaN issue mentioned by @jhuguetn - I pointed out this potential issue in the original discussion. I could have elaborated it more.

After the `_handle_scrubbed_volumes` step, you should get `signals` and `confounds` the same shape, with the heads and tails trimmed if need be. 
You can reuse the `_censor_signals` function to remove NaN after this block:
https://github.com/nilearn/nilearn/blob/db5a97047b5ff6d356664b5e0ee40c46c5ba748a/nilearn/signal.py#L822-L826
What you will need a new `sample_mask` that's for the volumes of the heads and tails you want to discard. 
Let me know if this makes sense. I can implement it if need be.

jhuguetn:
Thanks @htwangtw for the suggestion, I will try that

ymzayek:
Hi @jhuguetn did you accidentally close this or would you like one of the core developers to take over working on it?

jhuguetn:
Sorry, I discarded the first changes and replaced them with new ones based on suggestions made.

htwangtw (commented):
We need a new test for `clean` as there are code modified beyond the private function. 

I suggest to look for all instance of `sample_mask` after `_handle_scrubbed_volumes` and see what's the impact of the NaN values.

Let me know if you would prefer me to take over.

nilearn/signal.py
```diff
@@ -732,8 +738,12 @@ def clean(
 
     # Interpolation / censoring
     signals, confounds = _handle_scrubbed_volumes(
-        signals, confounds, sample_mask, filter_type, t_r
+        signals, confounds, sample_mask, filter_type, t_r, extrapolate
     )
+    if sample_mask is not None and not extrapolate:
+        # reset the indexing of the sample_mask excluding non-interpolated
+        # volumes at the head of the data
+        sample_mask -= sample_mask[0]
```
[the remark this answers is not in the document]
jhuguetn: Sorry @htwangtw but I am afraid I can't follow you fully. If you refer to the data type (list vs numpy.array), then indeed it won't work unless you cast your list of indexes.  But not because of any of the changes introduced, simply because sanitizing `_sanitize_sample_mask()` function will raise a TypeError.  
Otherwise, if you refer to the fact of having a sample_mask which labels your n first volumes as valid/good, the index resetting operation stills valid. Nothing should be modified as is the case when interpolation gets rid of the volumes at the end of the data run, solely when volumes are at the beginning of the run, you need to reset the `sample_mask` indexes (unless I am missing something).

nilearn/signal.py
```diff
@@ -833,14 +845,30 @@ def _censor_signals(signals, confounds, sample_mask):
...
     remained_vol = frame_times[sample_mask]
     remained_x = volumes[sample_mask, :]
-    cubic_spline_fitter = CubicSpline(remained_vol, remained_x)
+    cubic_spline_fitter = CubicSpline(
+        remained_vol, remained_x, extrapolate=extrapolate
+    )
     volumes_interpolated = cubic_spline_fitter(frame_times)
     volumes[~sample_mask, :] = volumes_interpolated[~sample_mask, :]
+    # discard non-interpolated out-of-bounds volumes
+    volumes = volumes[~np.isnan(volumes).all(axis=1), :]
```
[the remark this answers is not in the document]
jhuguetn: > What I recommand is to discard the NaN volumes outside of `_interpolate_volumes`, in `_handle_scrubbed_volumes`. And since sample_mask will be modified, `_handle_scrubbed_volumes` will return signals, confounds, sample_mask. 

Done, see https://github.com/nilearn/nilearn/pull/4028/commits/2ae9efdd0b0d5393566aed23ce430eeb403ad8c0.

nilearn/signal.py
```diff
@@ -833,12 +848,26 @@ def _censor_signals(signals, confounds, sample_mask):
...
 
-def _interpolate_volumes(volumes, sample_mask, t_r):
+def _interpolate_volumes(volumes, sample_mask, t_r, extrapolate):
     """Interpolate censored volumes in signals/confounds."""
+    if extrapolate:
+        extrapolate_default = (
+            "By default the cubic spline interpolator extrapolates "
+            "the out-of-bounds censored volumes in the data run. This "
+            "can lead to undesired filtered signal results. In future "
+            "releases, the default strategy will be not to extrapolate "
```
bthirion, line 858: Starting from release 0.13,...

bthirion (commented):
We're almost there I think ! Thx,

nilearn/signal.py
jhuguetn: Fixed in 9f2c5258faf450b41659a6f5430deac1fa68efc5

bthirion (approved):
LGTM, thx.

ymzayek (commented):
@jhuguetn can you add a test checking that the warning is raised? And can you also add a whatsnew entry to `doc/changes/latest.rst` and your name to `CITATION.cff`? (see https://nilearn.github.io/dev/development.html#changelog for more detail)

nilearn/signal.py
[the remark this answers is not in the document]
htwangtw: Ah thanks for the elaboration! Now this is clearer to me

nilearn/tests/test_signal.py
```diff
@@ -1075,8 +1076,55 @@ def test_handle_scrubbed_volumes():
...
+        sample_mask,
+    ) = nisignal._handle_scrubbed_volumes(
+        signals, confounds, sample_mask, "butterworth", 2.5, False
+    )
+    np.testing.assert_equal(
+        signals.shape[0], interpolated_signals.shape[0] + censored_samples
+    )
+    np.testing.assert_equal(
+        confounds.shape[0], interpolated_confounds.shape[0] + censored_samples
+    )
```
htwangtw, line 1130: I would add a test to check if this modified sample mask can be correctly cast on the `interpolated_signals` and `interpolated_confounds`

nilearn/tests/test_signal.py
```diff
@@ -1075,8 +1076,55 @@ def test_handle_scrubbed_volumes():
...
+    sample_mask = np.arange(signals.shape[0])
+    scrub_index = np.arange(censored_samples)
+    sample_mask = np.delete(sample_mask, scrub_index)
+
+    # Test cubic spline interpolation (enabled extrapolation) in the
+    # very first n=5 samples of generated signal
+    (
+        extrapolated_signals,
+        extrapolated_confounds,
+        sample_mask,
```
htwangtw, line 1106:
```suggestion
        extrapolated_sample_mask,
```

nilearn/tests/test_signal.py
```diff
@@ -1075,8 +1076,55 @@ def test_handle_scrubbed_volumes():
...
+    ) = nisignal._handle_scrubbed_volumes(
+        signals, confounds, sample_mask, "butterworth", 2.5, True
+    )
+    np.testing.assert_equal(signals.shape[0], extrapolated_signals.shape[0])
+    np.testing.assert_equal(
+        confounds.shape[0], extrapolated_confounds.shape[0]
+    )
+
```
htwangtw, line 1114:
```suggestion
    np.testing.assert_equal(sample_mask, extrapolated_sample_mask)

```
I would add a test here to make sure `sample_mask` is not modified here.

nilearn/tests/test_signal.py
```diff
@@ -1075,8 +1076,55 @@ def test_handle_scrubbed_volumes():
...
+def test_handle_scrubbed_volumes_extrapolation():
+    """Check interpolation of signals with and without extrapolation."""
+    signals, _, confounds = generate_signals(
+        n_features=11, n_confounds=5, length=40
+    )
+
+    censored_samples = 5
+    sample_mask = np.arange(signals.shape[0])
+    scrub_index = np.arange(censored_samples)
```
htwangtw, line 1098: I would censor a few more parts in the signal, since there's part of the code manipulating the index of the new sample mask.
```suggestion
    scrub_index = np.arange(censored_samples)
    scrub_index = np.concatenate((censored_samples, np.array([20, 21, 32])))
```

htwangtw (commented):
I would recommand adding one more test in `nilearn/tests/test_signal.py::test_sample_mask` to test all the subsequent processes are performed correctly.

nilearn/tests/test_signal.py
jhuguetn: done, see 05014cc

nilearn/tests/test_signal.py
jhuguetn: done, see 05014cc

nilearn/tests/test_signal.py
jhuguetn: returned `sample_masks` renamed for evaluation, see 05014cc

jhuguetn:
> @jhuguetn can you add a test checking that the warning is raised? And can you also add a whatsnew entry to `doc/changes/latest.rst` and your name to `CITATION.cff`? (see https://nilearn.github.io/dev/development.html#changelog for more detail)

Done, see 41635d4 and 8d4e444.

nilearn/tests/test_signal.py
jhuguetn: done, see 639c483

nilearn/tests/test_signal.py
```diff
@@ -1075,8 +1076,79 @@ def test_handle_scrubbed_volumes():
...
+        scrubbed_confounds,
+        sample_mask,
+    ) = nisignal._handle_scrubbed_volumes(
+        signals, confounds, sample_mask, "cosine", 2.5, True
     )
     np.testing.assert_equal(scrubbed_signals, signals[sample_mask, :])
     np.testing.assert_equal(scrubbed_confounds, confounds[sample_mask, :])
+
+
+def test_handle_scrubbed_volumes_extrapolation():
```
ymzayek, line 1090: @jhuguetn this test has become a bit long. I suggest splitting it up even if it means repeating some data generation lines. Maybe two separate tests one with extrapolation and the other without will do

htwangtw (approved):
LGTM now! Thanks for all the hard work!

nilearn/tests/test_signal.py
jhuguetn: Agree, it makes sense to divide the content into 2 tests for a better code readability. See bd8e627.

ymzayek (approved):
Thx @jhuguetn ! The failure is unrelated to this PR. Merging

PR #4031, opened 2023-10-06
[DOC] Change all block splitters in examples to use short syntax

### LINKED ISSUE #3937: [DOC] Replace examples' block splitter symbols with code cell syntax

ymzayek:
- Closes #3937

### DIFF
--- a/doc/connectivity/region_extraction.rst
+++ b/doc/connectivity/region_extraction.rst
@@ -31,7 +31,7 @@ dataset.
 
 .. literalinclude:: ../../examples/03_connectivity/plot_extract_regions_dictlearning_maps.py
     :start-after: # We use nilearn's datasets downloading utilities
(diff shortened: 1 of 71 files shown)


Remi-Gau (approved):
Good with me

bthirion (approved):
LGTM, thx.

PR #4034, opened 2023-10-07
[MAINT] refactor is plotly installed

### LINKED ISSUE #3974: extract is_plotly_installed function
Wonder if some of this should not be extracted into a separate utility function `is_plotly_installed()`. 

this should probably be done in a different PR though, so don't overthing it for this PR

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/3949#discussion_r1317466053_

Remi-Gau:
- Closes #3974 

- extract functions to check is plotly and kaleido are installed


bthirion (commented):
Looks good. I don't find immediately a way to address the coverage pb.

ymzayek (approved):
LGTM! I think the coverage failure is ok since tests will fail if the functions don't work. You could add a test in test_helpers that actually calls the functions inside

ymzayek:
Can also do this for matplotlib (in a different PR)

bthirion (approved):

PR #4036, opened 2023-10-07
[MAINT] Update issue labels

### LINKED ISSUE #4015: list of labels in maintenance.rst mismatch actual list
Compare: 

https://github.com/nilearn/nilearn/blob/b214dce54c46d2d2117726a1a5192f9bf79e17ab/doc/maintenance.rst?plain=1#L83

and:

https://github.com/nilearn/nilearn/labels

**IF** we want to keep those in synch, some automtation via the github API to update the list of labels may be an option.

Some, I think, should be removed: for example "next-release" should be handled by github milestones.

A more minimalist option could be to only keep in the maintenance doc the most "stable" labels and have some github issue labels that are not listed in the doc.

Remi-Gau:
- Closes #4015

- link to issue label page
- remove obsolete labels


bthirion (approved):
LGTM, thx.

PR #3936, opened 2023-08-30
[MAINT] Implement deprecation warnings-as-errors in prerelease test using pytest fixture

ymzayek:
- Closes #3541 and maybe #1737

Changes proposed in this pull request:

-
-

Issue #3673, opened 2023-04-12
[ENH] remove or repurpose unused deprecation warning function


Remi-Gau:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed enhancement in detail.

As far as I can tell this deprecation warning function is unused in the code base.

https://github.com/nilearn/nilearn/blob/17d4b53d2bdb3a7722530eef9be1d58b305c9a0d/nilearn/glm/regression.py#L31

I would suggest removing it or moving to more general purpose utility module to be used as a standardized way to throw deprecation warnings.

### Benefits to the change

Having a single function to throw typical deprecation warnings may help when the time comes to remove them.

Side note related to this:
- one of the deprecation warning I added recently used [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) instead of [FutureWarning](https://docs.python.org/3/library/exceptions.html#FutureWarning) that is apparently the one to use to warn users and not just developers.

https://docs.python.org/3/library/warnings.html#warning-categories

Note that there are a few other usage of DeprecationWarning in the codebase which from a quick read should probably be FutureWarning.

ymzayek:
Yes, it seems they are used in an inconsistent fashion regarding the guidelines. Maybe even look into how `UserWarning` is used.

Regarding the function, maybe we can consolidate into `_utils.helpers` with the other deprecation related functions.

jeromedockes:
I'm not sure I understand why those fit the `FutureWarning` better than `DeprecationWarning` description:
DeprecationWarning doc says:
> Base category for warnings about deprecated features when those warnings are intended for other Python developers

this sounds adequate, Python developers writing applications that use nilearn should know they are using a nilearn feature that will change

FutureWarning says:

> Base category for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

I don't think it applies to us -- there's not much an end user could do about it anyway, and in any case nilearn is used mostly for scripting 

maybe the [PEP](https://peps.python.org/pep-0565/#additional-use-case-for-futurewarning) helps clarify a bit: 
deprecationwarning:
> [...] The intended audience is Python developers that are at risk of upgrades to their dependencies (including upgrades to Python itself) breaking their software (e.g. developers using Python to script environments where someone else is in control of the timing of dependency upgrades).

FutureWarning:
[...]

Issue #3625, opened 2023-03-27
Use  `pytest.deprecated_call` to test deprecated functions


Remi-Gau:
Replace the use of `with pytest.warns(DeprecationWarning)` with `with pytest.deprecated_call()`

See https://docs.pytest.org/en/latest/how-to/capture-warnings.html#ensuring-code-triggers-a-deprecation-warning

> [!TIP]
> # **💻 🍁  For the [hacktoberfest](https://hacktoberfest.com/) contributors 🍁 💻**
>
> Make sure to:
> - read our [contributing guidelines](https://nilearn.github.io/stable/development.html#contribution-guidelines)
>
> If this is your first time contributing to Nilearn, add your details in the [CITATION.cff file](https://github.com/nilearn/nilearn/blob/dcd92aa357ab3a751161aeea2edfa3a775ae644f/CITATION.cff#L42).
>
> For your pull request to be valid, you should to the following:

- [ ] remove the `check_deprecation` function in [nilearn/nilearn/_utils/testing.py](https://github.com/nilearn/nilearn/blob/1471ba887c1109d8ee8986179e0a229823a25eb6/nilearn/_utils/testing.py#L51)
- [ ] Replace the use of `with pytest.warns(DeprecationWarning)` with `with pytest.deprecated_call()` in the following

```bash
$ grep -nr "with pytest.warns(DeprecationWarning" nilearn/**/*.py

nilearn/connectome/tests/test_connectivity_matrices.py:941:    with pytest.warns(DeprecationWarning, match=match):
nilearn/datasets/tests/test_func.py:860:        with pytest.warns(DeprecationWarning):
nilearn/glm/tests/test_contrasts.py:67:    with pytest.warns(DeprecationWarning, match="0.13.0"):
nilearn/glm/tests/test_contrasts.py:262:    with pytest.warns(DeprecationWarning, match="0.13.0"):
nilearn/glm/tests/test_contrasts.py:265:    with pytest.warns(DeprecationWarning, match="0.13.0"):
nilearn/glm/tests/test_hemodynamic_models.py:60:    with pytest.warns(DeprecationWarning, match='"tr" will be removed in'):
nilearn/maskers/tests/test_nifti_labels_masker.py:220:    with pytest.warns(DeprecationWarning, match="Starting in version 0.12"):
nilearn/maskers/tests/test_nifti_maps_masker.py:186:    with pytest.warns(DeprecationWarning, match="Starting in version 0.12"):
[...]

PR #4043, opened 2023-10-10
[MAINT] Pytest fixture to ignore internal deprecation warnings

### LINKED ISSUE #1737: Too many deprecation warnings cluttering the outputs
Running CircleCI, we get a TON of deprecation warnings. This will be a problem in the future when the bill is suddenly due (and we may not forsee the deadline, since there are so many), and right now it makes the already verbose output of CI systems and makes it worse.

We should consider slowly attending to the deprecations as we go about fixing other stuff.

ymzayek:
- With #3936 closes #1737.

- Add a fixture to suppress known internal deprecation warnings

I'm not sure if this is the best approach, but it avoids having to go through all the tests that raise the warning and adding e.g. a parameter to suppress the warning, which is a bit of a hassle and adds parameters unrelated to the particular unit tests. Since the warnings are known and tested it shouldn't have any negative side effects and could also be a way to track our deprecation in one place.


Remi-Gau:
I would say that until we find a way to uniformize how we throw / test deprecation warnings (see https://github.com/nilearn/nilearn/issues/3673 and https://github.com/nilearn/nilearn/issues/3625) this seems like a good approach for now.

Remi-Gau:
failures are unrelated, no?

bthirion:
I think they are.

Remi-Gau:
Yup. I relaunched the test on main and everything passed so seems unrelated to some dependency change at least.

bthirion (approved):
LGTM, thx !

nilearn/conftest.py
```diff
@@ -112,6 +112,23 @@ def warnings_as_errors():
...
 
 
+@pytest.fixture(autouse=True)
+def suppress_specific_warning():
+    """Ignore internal deprecation warnings."""
+    with warnings.catch_warnings():
+        messages = (
+            "The `darkness` parameter will be deprecated.*|"
+            "`legacy_format` will default to `False`.*|"
+            "In release 0.13, this fetcher will return a dictionary.*|"
```
ymzayek, line 122: Thanks for finishing the PR @Remi-Gau !
Just fyi I did not check that this was a thorough list of internal deprecation messages.

Remi-Gau: will quickly check but your work should allow to make it easy to suppress if there are more of those warnings

PR #4047, opened 2023-10-11
[MAINT] Rename and move testing utilities under tests directory in each module

### LINKED ISSUE #4046: Homogenize location and naming of testing utilities
related to this: there are instances where utility functions used only for tests are in the folder `X` in a module called something like `X/testing.py` instead of in something like `X/tests/_utils.py`

Examples

- nilearn/datasets/_testing.py
- nilearn/surface/testing_utils.py 

Other cases where test utility functions are within the test folder:

- nilearn/mass_univariate/tests/utils.py
- nilearn/decoding/tests/_utils.py
- nilearn/interfaces/fmriprep/tests/utils.py

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/3613#issuecomment-1511578081_

ymzayek:
- Closes #4046

- Testing utilities are now under tests/_testing.py in each module where they exist. Happy to change the name but I didn't use any version of "utils" because there are utils.py files that are relevant to the module functionality as opposed to the testing and where they exist they are usually tested in tests/test_utils.py


Remi-Gau:
What's you take on this one: https://github.com/nilearn/nilearn/blob/main/nilearn/_utils/testing.py ?

I suspect it is in a private module and contains testing utilities used for several other modules so it is fine where it is and named as it is. Just want to make sure we are on the same page.

ymzayek:
> What's you take on this one: https://github.com/nilearn/nilearn/blob/main/nilearn/_utils/testing.py ?
> 
> I suspect it is in a private module and contains testing utilities used for several other modules so it is fine where it is and named as it is. Just want to make sure we are on the same page.

Yes that's what I was thinking too. It doesn't fit the pattern of the others that are private to the testing of a specific module (with one or two functions being an exception like the generate_surf in nilearn/surface/tests/_testing being used in plotting/tests/test_surf_plotting)

Remi-Gau:
OK we agree on this. Good to me then. Let's merge.

Thanks for taking care of this. This one has been itching my inner Marie Kondo for a while.

Remi-Gau (approved):

PR #4017, opened 2023-09-28
[DOC] Set up experimental module and surface API for docs

ymzayek:
- Closes None

PR #4049, opened 2023-10-12
[DOC] Improve docstring rendering for typed experimental surface module

### LINKED ISSUE #4018: [DOC] Fix Sphinx issues related to adding new experimental module and typing

1. Linkcode extension error https://github.com/nilearn/nilearn/pull/4017#issuecomment-1739213117 and related workaround: https://github.com/nilearn/nilearn/blob/9133bd98ddc085e14d545fd0439be621c8679f91/doc/sphinxext/github_link.py#L43-L44

2. See Sphinx warning https://github.com/nilearn/nilearn/pull/4017#issuecomment-1739630179 and related hotfix of remving sphinx role here: https://github.com/nilearn/nilearn/blob/9133bd98ddc085e14d545fd0439be621c8679f91/doc/modules/experimental.rst?plain=1#L10

3. Sphinx autodoc has an issue with resolving external libraries like numpy when using typehints. It seems to be a bug similar to what is reported here https://github.com/sphinx-doc/sphinx/issues/10785. For now this autodoc argument is added to hide the typehints from the integrated typing https://github.com/nilearn/nilearn/blob/9133bd98ddc085e14d545fd0439be621c8679f91/doc/conf.py#L57
And if we have both docstrings and the typehints we have double rendering in the docs of the description of a e.g. class and this is redundant and looks ugly so we should also discuss how we want to approach adding docstrings + typing and how to configure what will be displayed. I'll post an example once the dev docs are built

_No response_

ymzayek:
- Closes #4018

- Add some configs to doc/conf.py for typehints
- I'm using [`nilearn/experimental/surface/_maskers.py`](https://github.com/nilearn/nilearn/compare/main...ymzayek:improve-surface-docstrings?expand=1#diff-9ebbc94c276616a2f50615397567b68efe9c841100ebdc3de6232e33eaf17056) to test some options and after some decisions are made changes will be applied to the rest of this module. I'll add some comments and images to show differences of rendering.

There are two main things to pay attention to. I tried just removing the type from the docstring and let it render in the signature so as to not repeat information. This leaves the parameter names and the description. The other way would be to change to way we write docstrings but it renders better. See docstring of `SurfaceLabelsMasker.fit_transform()`.


nilearn/experimental/surface/_maskers.py
```diff
@@ -175,29 +175,23 @@ class SurfaceLabelsMasker:
...
 
-    labels_data_: type[np.ndarray]
-    labels_: type[np.ndarray]
-    label_names_: type[np.ndarray]
+    #: Labels array concatenated across both hemispheres
+    labels_data_: np.ndarray
+    #: Unique label values
+    labels_: np.ndarray
+    #: Label names if any, otherwise unique label values
+    label_names_: np.ndarray
```
ymzayek, line 194: attributes can just have this `#:` syntax with a description above each (which will render under) and no need for an attributes section in the docstring. See e.g.
image

nilearn/experimental/surface/_maskers.py
```diff
@@ -226,10 +220,10 @@ def fit(
...
         Parameters
         ----------
-        img : SurfaceImage object
+        img
             Mesh and data for both hemispheres.
 
-        y : None
+        y
             This parameter is unused. It is solely included for scikit-learn
             compatibility.
```
ymzayek, line 229: rendering of SurfaceLabelsMasker.fit ( we would render the input parameter types in the signature for this option but I somehow messed up the configs here so use your imagination :) )
image

nilearn/experimental/surface/_maskers.py
```diff
@@ -266,20 +260,12 @@ def transform(self, img: SurfaceImage) -> np.ndarray:
...
+                  scikit-learn compatibility.
 
-        Returns
-        -------
-        numpy.ndarray
-            Signal for each element.
-            shape: (img data shape, total number of vertices)
+        :return: Signal for each element with shape
+                 (img data shape, total number of vertices).
         """
```
ymzayek, line 269: SurfaceLabelsMasker.fit_transform with docstring compatible with sphinx autdoc typehints to get the types and defaults automatically added in the docstring instead of the signature (I think can also be done for the return if we want to keep it out of the signature):
image

doc/modules/experimental.rst
```diff
@@ -7,8 +7,12 @@
...
    :no-inherited-members:
 
-nilearn.experimental.surface
-----------------------------
+:mod:`nilearn.experimental.surface`
+-----------------------------------
+
+.. automodule:: nilearn.experimental.surface
+   :no-members:
+   :no-inherited-members:
```
ymzayek, line 15: this is actually needed to correctly reference the module and wasn't causing the :noindex: error in #4017

ymzayek:
FWIW I did try [Napoleon](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) with the napoleon typehints extension but ran into a bunch of issues. While it has some nice features and can better standardize our docstrings, it seem to be incompatible with our current docstrings as they are so will either require some work to update those in the whole codebase or some more intense configs and workarounds. In contrast, the configs in this PR do not affect the docstrings of untyped code. And at least the decisions/issues resolved in this PR also seem to apply to and be compatible with Napoleon so someone can still switch to it later if desired

Remi-Gau:
Thanks for looking into this @ymzayek !!

Some options would make the docstring writing much easier.

Given this is related to type hints, should we maybe add this to the agenda for the next dev meeting?

This may also help solve how to handle the PR and issues related to where to put "defaults" in the doc string.

nilearn/experimental/surface/_maskers.py
Remi-Gau: The pro is that the documentation is closer to where the attributes are declared in the code but this would mean rewriting a lot of doc, no?

ymzayek:
> Given this is related to type hints, should we maybe add this to the agenda for the next dev meeting?

@Remi-Gau yes definitely, I've put it on the agenda under issues/PRs

What I'm struggling with is trying to merge both approaches, i.e. having numpydoc style docstring in the codebase but substituting types (and defaults) automatically without doubling the parameters. The problem with the use of `#:` and `:param ...` syntax is that it is not as readable in the code. I'm not quite sure yet if the support for typing is lacking or that I haven't figured out the right mix of configs and changes to the docstrings.

ymzayek:
Update: Now this PR deals with only points 1 and 2 of https://github.com/nilearn/nilearn/issues/4018. 
Since we decided not to support typing at the moment, we can deal with its rendering at a later point since how we implement the typing will influence the configuration.

bthirion (approved):
LGTM, thx.

Remi-Gau (approved):
Thanks for looking into all this @ymzayek

PR #4051, opened 2023-10-12
[FIX] image.clean_img() kwargs integration

### LINKED ISSUE #4048: [BUG] clean_img with sample_mask passed to signal.clean fails because output dimensions are not the same as the input

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Red Hat Enterprise Linux 8.8 (Ootpa)

### Python version

- [ ] 3.11
- [X] 3.10
- [ ] 3.9
- [ ] 3.8
- [ ] 3.7

### nilearn version

0.10.1

### Expected behavior

I'm trying to apply a nuisance regression to data preprocessed with fmriprep. I am reading in certain confounds and, according to threshold criteria, creating a `sample_mask` with `nilearn.interfaces.fmriprep.load_confounds`. 

I would like to clean the data according to the algorithm presented in `nilearn.signal.clean`, but I want to input the nifti, so I use `nilearn.image.clean_image`. Since `image.clean_image` passes data to `signal.clean`, and `signal.clean` scrubs the data if you give it `sample_mask`, I expect that giving `image.clean_img` a `sample_mask` as a kwarg should result in cleaned and scrubbed data. 

So the output should be fewer timepoints than the input. This is possible in `signal.clean` but the current help of `image.clean_img` says that the output is the same dimensions as the input.

This is what I got:

```bash
nifti.shape
     (97, 115, 97, 391)

len(sample_mask)
     310

ValueError                                Traceback (most recent call last)
/tmp/ipykernel_2542916/2394484833.py in <module>
      8 nifti = nib.load(f'{path}/{file}')
      9 confounds_out, sample_mask = load_confounds(f'{path}/{file}', strategy=('motion','wm_csf', 'scrub'), motion='full', scrub=5,fd_threshold=0.3, std_dvars_threshold=3, wm_csf='basic', compcor='anat_combined', n_compcor=5, demean=True)
---> 10 clean_nifti = image.clean_img(nifti, detrend=True, standardize='zscore_sample', confounds=confounds_out, standardize_confounds=True, filter='butterworth', low_pass=0.1, high_pass=0.01, t_r=0.8, **{'clean__sample_mask':sample_mask})

~/miniconda3/lib/python3.7/site-packages/nilearn/_utils/helpers.py in wrapper(*args, **kwargs)
     40                                                      kwargs
     41                                                      )
---> 42             return func(*args, **kwargs)
     43 
     44         return wrapper

~/miniconda3/lib/python3.7/site-packages/nilearn/image/image.py in clean_img(imgs, runs, detrend, standardize, confounds, low_pass, high_pass, t_r, ensure_finite, mask_img, **kwargs)
   1287     else:
   1288         imgs_ = new_img_like(
-> 1289             imgs_, data.T.reshape(imgs_.shape), copy_header=True
   1290         )
   1291 

ValueError: cannot reshape array of size 335430850 into shape (97,115,97,391)
```

```python
from nilearn import image
import nibabel as nib

# image should have > 0 timepoints to be scrubbed
file ='preproc_bold.nii.gz'
path = 'path/to/func/'

nifti = nib.load(f'{path}/{file}')
confounds_out, sample_mask = load_confounds(f'{path}/{file}', strategy=('motion','wm_csf', 'scrub'), motion='full', scrub=5,fd_threshold=0.3, std_dvars_threshold=3, wm_csf='basic', compcor='anat_combined', n_compcor=5, demean=True)
clean_nifti = image.clean_img(nifti, detrend=True, standardize='zscore_sample', confounds=confounds_out, standardize_confounds=True, filter='butterworth', low_pass=0.1, high_pass=0.01, t_r=0.8, **{'clean__sample_mask':sample_mask})
```

MIZwally:
Fixed the bug in issue #4048; reformats the kwargs in `clean_kwargs` so they can be properly passed into the `signal.clean` function. Additionally, upon the use of a `sample_mask` kwarg, the size of the nifti is updated to reflect the new number of volumes after cleaning.

- Closes #4048


bthirion (commented):
Can you add a test  to enforce the behavior ?

Remi-Gau:
@MIZwally 

You can use the code from @ymzayek here: https://github.com/nilearn/nilearn/issues/4048#issuecomment-1759680049

to add a test in this file:
https://github.com/nilearn/nilearn/blob/main/nilearn/image/tests/test_image.py

See our contributing doc regarding tests: https://nilearn.github.io/stable/development.html#tests

Remi-Gau:
You should be able to fix code formatting issues with pre-commit:

```bash
pip install pre-commit
pre-commit install
pre-commit run -a
```

nilearn/image/image.py
```diff
@@ -1267,7 +1267,7 @@ def clean_img(
 
     # Clean signal
     clean_kwargs = {
-        k[7:]: v for k, v in kwargs.items() if k.startswith("clean__")
+        k[7:]: v for k, v in kwargs['kwargs'].items() if k.startswith("clean__")
```
ymzayek, line 1270: I don't understand this change. This will lead to a KeyError. I suggest reverting it

nilearn/image/image.py
```diff
@@ -1285,6 +1285,12 @@ def clean_img(
...
         imgs_ = masking.unmask(data, mask_img)
+    elif 'sample_mask' in clean_kwargs :
+        shape_list = list(imgs_.shape[:3])
+        shape_list.append(clean_kwargs['sample_mask'].shape[0])
+        imgs_ = new_img_like(
+            imgs_, data.T.reshape(tuple(shape_list)), copy_header=True
+        ) 
     else:
         imgs_ = new_img_like(
             imgs_, data.T.reshape(imgs_.shape), copy_header=True
```
ymzayek, line 1296: You can simplify this a bit ~~and I think it needs to be an if statement otherwise it won't run if mask_img is not equal to None~~. I suggest replacing L1288-L1297 with:

```python
    elif 'sample_mask' in clean_kwargs:
        sample_shape = imgs_.shape[:3] + clean_kwargs['sample_mask'].shape
    else:
        sample_shape = imgs_.shape

    imgs_ = new_img_like(
        imgs_, data.T.reshape(sample_shape), copy_header=True
    )
```
**edit** elif makes sense I spoke too soon :)

nilearn/image/image.py
```diff
@@ -1285,6 +1285,12 @@ def clean_img(
     # Put results back into Niimg-like object
     if mask_img is not None:
         imgs_ = masking.unmask(data, mask_img)
+    elif 'sample_mask' in clean_kwargs :
+        shape_list = list(imgs_.shape[:3])
+        shape_list.append(clean_kwargs['sample_mask'].shape[0])
+        imgs_ = new_img_like(
+            imgs_, data.T.reshape(tuple(shape_list)), copy_header=True
+        )
```
ymzayek, line 1293:
```suggestion
    elif 'sample_mask' in clean_kwargs:
        sample_shape = imgs_.shape[:3] + clean_kwargs['sample_mask'].shape
        imgs_ = new_img_like(
            imgs_, data.T.reshape(sample_shape), copy_header=True
        )
```

nilearn/image/image.py
MIZwally: I misunderstood how to pass the kwargs into the function. I was using kwargs={}, so I needed to add that change to find the kwarg.

bthirion (commented):
Thx. You need to add the corresponding test.

Remi-Gau (changes requested):
make sure to update the changelog and add your name in the citation.cff file: see instructions here 

Remi-Gau (changes requested):
I can see that you made a [commit](https://github.com/MIZwally/nilearn/commit/1f08dde2440bab228f100779d4c52ffe6aca418b) to update the changelog but it does not appear on the PR.

See all the changes listed in this PR.

Also make sure to run black and flake8 on your code: https://github.com/nilearn/nilearn/pull/4051#issuecomment-1760958133

And make sure to add a test: https://github.com/nilearn/nilearn/pull/4051#issuecomment-1760954070

Remi-Gau:
OK it seems the test is failing because the data fetcher is failing.

@ymzayek I will invetigate but in case you know how your code snippet worked back then and not now let me know.
https://github.com/nilearn/nilearn/issues/4048#issuecomment-1759680049

ymzayek:
@Remi-Gau sorry haven't had time to review this yet. Just in general I wouldn't recommend that snippet as it is for a test. That snippet was just a quick way to reproduce the error in a python environment and not for pytest. We should always use minimal/generated data for our tests. We don't want to rely on fetching data for tests as this could cause unrelated error due to e.g. network issues

Remi-Gau:
yeah I was getting to the same conclusion after 5 minutes I am already in "mocking hell" just to get the test data set up.

Will see if we can generate some dummy fmriprep data + confounds with our _utils.

@MIZwally bare with us, while we get this sorted.

ymzayek:
@MIZwally you can add a test like:

```python
#import at the top
from nilearn._utils.data_gen import _basic_confounds

#replace current test with
def test_clean_img_sample_mask(img_4d_rand_eye):
    """Check sample mask can be passed as a kwarg and used correctly."""
    length = img_4d_rand_eye.shape[3]
    confounds = _basic_confounds(length)
    # exclude last time point
    sample_mask =  np.arange(length - 1)

    img = image.clean_img(
        img_4d_rand_eye,
        detrend=True,
        standardize="zscore_sample",
        confounds=confounds,
        standardize_confounds="zscore_sample",
        **{"clean__sample_mask": sample_mask},
    )
    # original shape is (10, 10, 10, 10)
    assert img.shape == (10, 10, 10, 9)
```

@Remi-Gau I can refactor following this PR to either make _basic_confounds "public" or make a fixture.

nilearn/image/image.py
```diff
@@ -1284,6 +1284,11 @@ def clean_img(
         imgs_ = new_img_like(
             imgs_, data.T.reshape(imgs_.shape), copy_header=True
         )
+    if "sample_mask" in clean_kwargs:
+        sample_shape = imgs_.shape[:3] + clean_kwargs["sample_mask"].shape
+        imgs_ = new_img_like(
+            imgs_, data.T.reshape(sample_shape), copy_header=True
+        )
```
ymzayek, line 1291: I'm not sure at having this at the end like this as it overwrites results from the if else statement above. Don't we need to make sure that whether "sample_mask" is given or not, if mask_img is not None that the cleaned signal is first unmasked?

ymzayek (changes requested):

nilearn/image/image.py
MIZwally: I changed it from an elif statement because they are not mutually exclusive, but this does seem to overwrite the mask_img portion. We could move this if statement to be above the mask_img if statement, or we can somehow incorporate the changes made from the previous statement into this one so that the shape is changed with the sample_mask but does not overwrite the changes from the mask_img.

ymzayek: So what is stored in the data variable does drop the samples already when returned from signal.clean so if both sample_mask and mask_img are given the unmasking function will correctly reshape and put the data into a nifti object. I would put this new code back into an `elif` statement above the last `else` statement.

bthirion (approved):
LGTM, thx !

ymzayek (approved):

ymzayek:
@Remi-Gau did you want to have a final look at this since you've requested changes?

Remi-Gau (approved):

Remi-Gau:
Good with me.
thanks @MIZwally

Issue #3926, opened 2023-08-24
use rng fixture


Remi-Gau:
listing of potential targets to use the `rng` fixture

```bash
grep --include=test\*.py -rnw nilearn/**/tests -e check_random_state

nilearn/connectome/tests/test_connectivity_matrices.py:11:from sklearn.utils import check_random_state
nilearn/connectome/tests/test_connectivity_matrices.py:63:    random_state = check_random_state(random_state)
nilearn/connectome/tests/test_connectivity_matrices.py:94:    random_state = check_random_state(random_state)
nilearn/connectome/tests/test_connectivity_matrices.py:276:    random_state = check_random_state(random_state)
nilearn/datasets/tests/test_func.py:19:from sklearn.utils import check_random_state
nilearn/datasets/tests/test_func.py:376:    rng = check_random_state(42)
nilearn/decoding/tests/test_objective_functions.py:7:from sklearn.utils import check_random_state
nilearn/decoding/tests/test_objective_functions.py:25:    rng = check_random_state(42)
nilearn/decoding/tests/test_graph_net.py:8:from sklearn.utils import check_random_state
nilearn/decoding/tests/test_graph_net.py:66:    rng = check_random_state(42)
nilearn/decoding/tests/test_graph_net.py:80:    rng = check_random_state(42)
nilearn/decoding/tests/test_graph_net.py:94:    rng = check_random_state(42)
nilearn/decoding/tests/test_graph_net.py:118:    rng = check_random_state(42)
nilearn/decoding/tests/test_graph_net.py:196:    rng = check_random_state(42)
nilearn/decoding/tests/test_graph_net.py:222:    rng = check_random_state(42)
nilearn/decoding/tests/test_same_api.py:15:from sklearn.utils import check_random_state
nilearn/decoding/tests/test_same_api.py:43:        rng = check_random_state(42)
nilearn/decoding/tests/test_same_api.py:82:    rng = check_random_state(42)
nilearn/decoding/tests/test_same_api.py:99:    rng = check_random_state(42)
nilearn/decoding/tests/test_same_api.py:113:    rng = check_random_state(42)
nilearn/decoding/tests/test_space_net.py:11:from sklearn.utils import check_random_state
nilearn/decoding/tests/test_space_net.py:47:    rng = check_random_state(42)
nilearn/decoding/tests/test_space_net.py:83:    rng = check_random_state(42)
nilearn/decoding/tests/test_space_net.py:222:    rng = check_random_state(42)
[...]

PR #4052, opened 2023-10-13
[MAINT] Use rng fixture for setting up `np.random.RandomState()` - part 1

ymzayek:
- Part of #3926 (to avoid big diff)

- Use rng fixture in modules _utils, connectome, datasets, decoding, decomposition, glm


nilearn/connectome/tests/test_connectivity_matrices.py
```diff
@@ -463,10 +463,10 @@ def test_sym_matrix_to_vec():
...
 
-def test_sym_matrix_to_vec_is_the_inverse_of_vec_to_sym_matrix():
+def test_sym_matrix_to_vec_is_the_inverse_of_vec_to_sym_matrix(rng):
     n = 5
     p = n * (n + 1) // 2
-    rand_gen = np.random.RandomState(0)
+    rand_gen = rng
 
     # when diagonal is included
     vec = rand_gen.rand(p)
```
Remi-Gau, line 472:
```suggestion
    # when diagonal is included
    vec = rng.rand(p)
```

Remi-Gau: may want to replace `rand_gen` with `rng` in the rest of test

nilearn/glm/tests/test_regression.py
```diff
@@ -1,32 +1,38 @@
...
     model = ARModel(design=X, rho=0.4)
     results = model.fit(Y)
     assert results.df_residuals == 30
     assert results.residuals.shape[0] == 40
     assert results.predicted.shape[0] == 40
 
 
-def test_residuals():
+def test_residuals(X, Y):
     Xintercept = X.copy()
```
Remi-Gau, line 36: I think the copy was there because the tests we using globals, with fixtures this should not be needed, no?

nilearn/glm/tests/test_regression.py
```diff
@@ -53,15 +59,15 @@ def test_predicted_r_square():
     assert_almost_equal(results.r_square, 1.0)
 
 
-def test_OLS_degenerate():
+def test_OLS_degenerate(X, Y):
     Xd = X.copy()
```
Remi-Gau, line 63: same for this copy

nilearn/glm/tests/test_regression.py
```diff
@@ -53,15 +59,15 @@ def test_predicted_r_square():
...
     Xd[:, 0] = Xd[:, 1] + Xd[:, 2]
     model = OLSModel(design=Xd)
     results = model.fit(Y)
     assert results.df_residuals == 31
 
 
-def test_AR_degenerate():
+def test_AR_degenerate(X, Y):
     Xd = X.copy()
```
Remi-Gau, line 71: same for this copy

nilearn/decoding/tests/test_decoder.py
```diff
@@ -80,8 +81,7 @@ def _make_binary_classification_test_data(n_samples=N_SAMPLES):
 
 @pytest.fixture(scope="session")
 def rand_X_Y():
-    rng = np.random.RandomState(0)
-    X = rng.rand(N_SAMPLES, 10)
+    X = _rng().rand(N_SAMPLES, 10)
```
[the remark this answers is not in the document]
ymzayek: yes if they have the same scope. I wasn't sure if this fixture needed to have session scope or not so I went with the quickest and safest option of not changing that but I could check

nilearn/decoding/tests/test_searchlight.py
```diff
@@ -5,14 +5,14 @@
 from nibabel import Nifti1Image
 from sklearn.model_selection import KFold
 
+from nilearn.conftest import _rng
 from nilearn.decoding import searchlight
 
 
 def _make_searchlight_test_data(frames):
```
bthirion, line 12: Why not make rng an argument of the test in that case ?

nilearn/glm/tests/test_regression.py
```diff
@@ -1,45 +1,49 @@
...
-def test_residuals():
-    Xintercept = X.copy()
-
+def test_residuals(X, Y):
     # If design matrix contains an intercept, the
     # mean of the residuals should be 0 (short of
     # some numerical rounding errors)
-    Xintercept[:, 0] = 1
-    model = OLSModel(design=Xintercept)
+    X[:, 0] = 1
```
bthirion, line 39: Isn't X modified ?

bthirion (commented):
LGTM, thx !

nilearn/decoding/tests/test_searchlight.py
Remi-Gau: AFAICT because `rng` is a fixture and fixtures can only be passed as arguments to a test function (or another fixture) but not to a utility function (pretty sure it is more complicated than that but this is the super short version)

nilearn/glm/tests/test_regression.py
Remi-Gau: yeah but it gets regenerated every time the fixture is invoked so eventhough it is modified in this test, the next test will get a "fresh new X".

Remi-Gau (approved):

Issue #3653, opened 2023-04-04
[ENH] add a `.git-blame-ignore-revs` to ignore formatting PR from git blame


Remi-Gau:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed enhancement in detail.

Came across while browsing the black documentation:
https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

One problem with all the reformating PRs is that they make reduce the usefulness (or at least add more friction) to the use of git blame to trace change was introduced when: black may end up touching many if not most lines in the code base and git blame may invariably point to one of those reformatting PRs when we use it, even though it very unlikely that a behavioral change came from them.

Using a `.git-blame-ignore-revs` could allow to "skip" to PR during a git blame call.

### Benefits to the change

Ideally this should facilitate the tracking down of changes in the code base by ignoring changes that pure reformatting and do not entail functional change.

Remi-Gau:
Let me know what you think.

ymzayek:
I already feel that it's a bit slower to track down changes so I would be for this, but I haven't used it before. What would be the downside if any?

NicolasGensollen:
I never used it but I'm +1 on this. 
The risk and cost seem very small and I can clearly see the benefits of having a meaningful blame where @Remi-Gau isn't responsible for all the bugs of Nilearn 😉.

Remi-Gau:
> I can clearly see the benefits of having a meaningful blame where @Remi-Gau isn't responsible for all the bugs of Nilearn

Rémi then proceeds to blame-ignore ALL the PRs he ever authored... 😜

Remi-Gau:
Related to https://github.com/nilearn/nilearn/pull/3701#discussion_r1170076459

```bash
 git blame --ignore-revs-file .git-blame-ignore-revs nilearn/image/resampling.py 
```
copy paste of a comment from #3755 

Still shows content from 3651 event though it is supposed to be ignored:

```
# run isort and flake8 on the whole nilearn code base in CI (#3651)
44933309ecd09add795ce1c3ea5edaab4294da87
```

More annoyingly it is also still the case for:

```bash
 git blame --ignore-rev 44933309  nilearn/image/resampling.py
```

ghisvail:
We use [one of these](https://github.com/aramis-lab/clinica/blob/dev/.git-blame-ignore-rev) in Clinica.

ymzayek:
[...]

PR #4066, opened 2023-10-18
[MAINT] refactor of tests and code in GLM module

Remi-Gau:
- Relates to #3570

- refactor and simplify complex tests in nilearn
- apply some sourcery suggested refactor in nilearn

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -22,7 +22,7 @@ Enhancements
 - :bdg-primary:`Doc` Add backslash to homogenize :class:`~nilearn.regions.Parcellations` documentation (:gh:`4042` by `Nikhil Krish`_).
 - :bdg-success:`API` Allow passing Pandas Series of image filenames to :class:`~nilearn.glm.second_level.SecondLevelModel` (:gh:`4070` by `Rémi Gau`_).
 - Allow setting ``vmin`` in :func:`~nilearn.plotting.plot_glass_brain` and :func:`~nilearn.plotting.plot_stat_map` (:gh:`3993` by `Michelle Wang`_).
-- :bdg-success:`API` Support passing t and F contrasts to :func:`~nilearn.glm.contrasts.compute_contrast` that that have fewer columns than the number of estimated parameters. Remaining columns are padded with zero (:gh:`4067` by `Remi Gau`_).
+- :bdg-success:`API` Support passing t and F contrasts to :func:`~nilearn.glm.contrasts.compute_contrast` that that have fewer columns than the number of estimated parameters. Remaining columns are padded with zero (:gh:`4067` by `Rémi Gau`_).
 
 Changes
 -------
--- a/nilearn/glm/tests/test_thresholding.py
+++ b/nilearn/glm/tests/test_thresholding.py
@@ -24,10 +24,6 @@ def test_fdr(rng):
 
     assert_almost_equal(fdr_threshold(x, 0.1), norm.isf(0.0005))
     assert fdr_threshold(x, 0.001) == np.infty
-    with pytest.raises(ValueError):
-        fdr_threshold(x, -0.1)
-    with pytest.raises(ValueError):
-        fdr_threshold(x, 1.5)
 
     # addresses #2879
     n = 10
@@ -35,6 +31,19 @@ def test_fdr(rng):
     pvals[0] = 0.007
 
     assert np.isfinite(fdr_threshold(norm.isf(pvals), 0.1))
+
+
+def test_fdr_error(rng):
+    n = 100
+    x = np.linspace(0.5 / n, 1.0 - 0.5 / n, n)
+    x[:10] = 0.0005
+    x = norm.isf(x)
+    rng.shuffle(x)
+
+    with pytest.raises(ValueError):
+        fdr_threshold(x, -0.1)
+    with pytest.raises(ValueError):
+        fdr_threshold(x, 1.5)
 
 
 @pytest.fixture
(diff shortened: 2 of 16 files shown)


nilearn/glm/tests/test_thresholding.py
```diff
@@ -37,6 +33,19 @@ def test_fdr(rng):
...
+def test_fdr_error(rng):
+    n = 100
+    x = np.linspace(0.5 / n, 1.0 - 0.5 / n, n)
+    x[:10] = 0.0005
+    x = norm.isf(x)
+    rng.shuffle(x)
+
+    with pytest.raises(ValueError):
+        fdr_threshold(x, -0.1)
+    with pytest.raises(ValueError):
```
bthirion, line 45: Want to make the error explicit ?

bthirion (commented):
LGTM overall

nilearn/glm/tests/test_thresholding.py
Remi-Gau: good point will actually do that for all the "bare" `pytest.raises`

Remi-Gau: This makes me realize that bare pytest.raises were not catching we would expect from the comments in the test or from the logic of the test.

We should probably go through the test suite to ensure that error and warning messages are properly tested.

jeromedockes:
> I am getting a failure on Windows only, on this test. Anyone can spot the issue here?

it seems the default encoding used on windows with a wester/american locale is more lenient than utf-8:

```python
import io

img_data = (
    b"GIF87a\x01\x00\x01\x00\xe7*\x00\x00\x00\x00\x01\x01"
    b"\x01\x02\x02\x07\x08\x08\x08\x0b\x0b\x0b\x0c\x0c\x0c\r;"
)
buf = io.BytesIO(img_data)
stream = io.TextIOWrapper(buf, encoding="cp1252")
stream.readline() # OK (but garbage output)
buf = io.BytesIO(img_data)
stream = io.TextIOWrapper(buf, encoding="utf-8")
stream.readline() # UnicodeDecodeError
```

btw,
- the documentation for `open` says a `ValueError` will be raised if an encoding error occurs (by default). UnicodeDecodeError is a subclass of ValueError so this is consistent with the behavior. but AFAICT there is no guarantee at least in the documentation that a UnicodeDecodeError will be used
- I'm not sure this check is useful, it basically raises the same error as the one that would propagate if it wasn't caught. actually it is less informative because the one raised by open specifies which encoding it tried, and that information is dropped by the nilearn rewording. also it is incorrect, because it states "does not appear to be a valid unicode text file", whereas it may be a perfectly well encoded text file, but encoded with a different encoding than the locale's default, which is used here

jeromedockes:
also looking at the git blame for this file, it seems the black-formatting commits still show up in the git blame

Remi-Gau:
> also looking at the git blame for this file, it seems the black-formatting commits still show up in the git blame

I know. 🙈 

Adding https://github.com/nilearn/nilearn/issues/3653 to the milestone for the next release.

Remi-Gau:
> * I'm not sure this check is useful, it basically raises the same error as the one that would propagate if it wasn't caught. actually it is less informative because the one raised by open specifies which encoding it tried, and that information is dropped by the nilearn rewording. also it is incorrect, because it states "does not appear to be a valid unicode text file", whereas it may be a perfectly well encoded text file, but encoded with a different encoding than the locale's default, which is used here

Thanks for looking this up. I did not even where to look on this one.
So you would suggest nuking this test out of the test suite?

jeromedockes:
> Thanks for looking this up. I did not even where to look on this one.

when you see different encoding errors or absence thereof on windows and unix, it helps to remember they don't use the same default encoding

another difference that comes up often in nilearn tests is windows refuses to unlink files that have an open file handle


> So you would suggest nuking this test out of the test suite?

the check in the code and the test

ymzayek (approved):
LGTM!

bthirion (approved):
LGTM, thx !

Remi-Gau:
Will fix the error in the doc build in this PR: it is just a typo (apparently I cannot spell my own name).

doc/changes/latest.rst
```diff
@@ -22,7 +22,7 @@ Enhancements
 - :bdg-primary:`Doc` Add backslash to homogenize :class:`~nilearn.regions.Parcellations` documentation (:gh:`4042` by `Nikhil Krish`_).
 - :bdg-success:`API` Allow passing Pandas Series of image filenames to :class:`~nilearn.glm.second_level.SecondLevelModel` (:gh:`4070` by `Rémi Gau`_).
 - Allow setting ``vmin`` in :func:`~nilearn.plotting.plot_glass_brain` and :func:`~nilearn.plotting.plot_stat_map` (:gh:`3993` by `Michelle Wang`_).
-- :bdg-success:`API` Support passing t and F contrasts to :func:`~nilearn.glm.contrasts.compute_contrast` that that have fewer columns than the number of estimated parameters. Remaining columns are padded with zero (:gh:`4067` by `Remi Gau`_).
+- :bdg-success:`API` Support passing t and F contrasts to :func:`~nilearn.glm.contrasts.compute_contrast` that that have fewer columns than the number of estimated parameters. Remaining columns are padded with zero (:gh:`4067` by `Rémi Gau`_).
```
Remi-Gau, line 25: @ymzayek 


Do you know why does the following throws a warning and fails in CI ?

```rst
:func:`~nilearn.glm.contrasts.compute_contrast`
```

ymzayek: oh I just opened a pr to fix it. It's how we have it set up in our doc/modules/glm.rst so compute_contrast is refrenced without the actual file name but just as a function of the module glm

PR #4067, opened 2023-10-18
[ENH] pad t and F 'short' contrasts with zeros

### LINKED ISSUE #2396: Length of contrast vectors depends on high cutoff frequency
The length of the contrast vectors in [this example](https://github.com/nistats/nistats/blob/61a86db07429739e0c8e7ee0110dbc7369fe7aa3/examples/01_tutorials/plot_single_subject_single_run.py), lines 156 and 157,

https://github.com/nistats/nistats/blob/61a86db07429739e0c8e7ee0110dbc7369fe7aa3/examples/01_tutorials/plot_single_subject_single_run.py#L156
https://github.com/nistats/nistats/blob/61a86db07429739e0c8e7ee0110dbc7369fe7aa3/examples/01_tutorials/plot_single_subject_single_run.py#L157

depends on the high cutoff frequency set by the user:
https://github.com/nistats/nistats/blob/61a86db07429739e0c8e7ee0110dbc7369fe7aa3/examples/01_tutorials/plot_single_subject_single_run.py#L99

Length of the contrast vectors should only depend on the number of regressors-of-interest. Regressors-of-no-interest should be introduced automatically, according to the parameters provided.

This issue is in reference to the PR nilearn/nistats#351

Remi-Gau:
- Closes #2396

- pad t and F constrast when they are too short
- throw a warning when padding is done
- throw error only when contrast are longer than the number of columns

### DIFF
--- a/nilearn/glm/_utils.py
+++ b/nilearn/glm/_utils.py
@@ -356,11 +356,15 @@ def pad_contrast(con_val, theta, contrast_type):
 
     Parameters
     ----------
-    con_val : numpy.ndarray of shape (p) or (q, p)
-        Where q = number of :term:`contrast` vectors
-        and p = number of regressors.
+    con_val : numpy.ndarray of shape (p) or (n, p)
+        Where p = number of regressors
+        with a value explicitly passed by the user.
+        p must be <= P,
+        where P is the total number of regressors in the design matrix.
 
-    theta theta of RegressionResults instances
+    theta : numpy.ndarray with shape (P,m)
+        theta of RegressionResults instances
+        where P is the total number of regressors in the design matrix.
 
     contrast_type : {'t', 'F'}, optional
         Type of the :term:`contrast`.
--- a/nilearn/glm/_utils.py
+++ b/nilearn/glm/_utils.py
@@ -343,3 +343,59 @@ def positive_reciprocal(X):
     """
     X = np.asarray(X)
     return np.where(X <= 0, 0, 1.0 / X)
+
+
+def pad_contrast(con_val, theta, contrast_type):
+    """Pad contrast with zeros if necessary.
+
+    If the contrast is shorter than the number of parameters,
+    it is padded with zeros.
+
+    If the contrast is longer than the number of parameters,
+    a ValueError is raised.
+
+    Parameters
+    ----------
+    con_val : numpy.ndarray of shape (p) or (q, p)
+        Where q = number of :term:`contrast` vectors
+        and p = number of regressors.
+
+    theta theta of RegressionResults instances
+
+    contrast_type : {'t', 'F'}, optional
+        Type of the :term:`contrast`.
+    """
+    nb_cols = con_val.shape[0] if con_val.ndim == 1 else con_val.shape[1]
+    if nb_cols > theta.shape[0]:
+        if contrast_type == "t":
+            raise ValueError(
+                f"t contrasts should be of length P={theta.shape[0]}, "
+                f"but it has length {nb_cols}."
+            )
+        if contrast_type == "F":
+            raise ValueError(
+                f"F contrasts should have {theta.shape[0]} columns, "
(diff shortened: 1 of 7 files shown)


Remi-Gau:
mentioned in the associated issue, using the contrasts symbolically (eg. 'c1 - c2') in the example mentioned in the issue so avoid having to change contrast length when the high pass filter model is changed.

Remi-Gau:
@bthirion I would not mind an early review on this.

I do not like that the warnings / errors that have to be thrown in different places, feels like there should be a single function checking contrast shape and doing the padding rather than changes in several places.

bthirion:
You mean, one single function for t and F ?

bthirion:
The PR looks good overall.

nilearn/glm/contrasts.py
```diff
@@ -106,6 +107,9 @@ def compute_contrast(labels, regression_result, con_val, contrast_type=None):
         for label_ in regression_result:
             label_mask = labels == label_
             reg = regression_result[label_]
+            con_val = pad_contrast(
+                con_val=con_val, theta=reg.theta, contrast_type=contrast_type
+            )
```
Remi-Gau, line 112: @bthirion 
any reason why for F contrasts we do not rely on the FContrast methods and do something like 

```python
reg = regression_result[label_].Fcontrast(con_val)
```

the way we do it for T contrasts above?

bthirion: IIRC, this is because we had to trick it somehow to support fixed effects on F contrasts.

Remi-Gau: ok this may require a comment if we can remember the why otherwise I could add a TODO to try to refactor this

bthirion: I can add this to your PR.

Remi-Gau:
# TODO

- [ ] update changelog

nilearn/glm/tests/test_second_level.py
```diff
@@ -1083,11 +1078,23 @@ def test_second_level_contrast_computation_errors(tmp_path, rng):
...
+    X = pd.DataFrame([[1]] * 4, columns=["intercept"])
+    model = model.fit(Y, design_matrix=X)
+
     with pytest.raises(
         ValueError,
-        match=("t contrasts should be length P=2, but this is length 1"),
+        match=("t contrasts should be of length P=1, but it has length 2."),
     ):
-        model.compute_contrast([1])
+        model.compute_contrast(second_level_contrast=[1, 2])
```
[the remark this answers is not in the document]
Remi-Gau: yup

nilearn/glm/contrasts.py
Remi-Gau: Added a TODO for now in case we want to merge and tackle this in a separate PR.

nilearn/glm/_utils.py
```diff
@@ -343,3 +343,59 @@ def positive_reciprocal(X):
...
+    If the contrast is longer than the number of parameters,
+    a ValueError is raised.
+
+    Parameters
+    ----------
+    con_val : numpy.ndarray of shape (p) or (q, p)
+        Where q = number of :term:`contrast` vectors
+        and p = number of regressors.
+
+    theta theta of RegressionResults instances
```
ymzayek, line 363: can you add the type for theta here?

ymzayek (commented):
LGTM overall!

nilearn/glm/_utils.py
Remi-Gau: added the type but also tried to improve the doc string: let me know if this reads sufficiently well.

nilearn/glm/contrasts.py
bthirion: Maybe rather in another PR.

bthirion (approved):
LGTM, thx.

PR #4068, opened 2023-10-18
[MAINT] testing minimal version of dependencies

### LINKED ISSUE #4069: nibabel minimal dependency should be bumped up
started investigating after the open office hours.

turns out that for some dependency X if we only allow the minimal version for X and the latest versions for all the others then we get into some situation that prevent installs or lead to failing tests.

Tested here in this dummy PR

https://github.com/nilearn/nilearn/actions/runs/6564645619/job/17831445204

- minimal version of nibabel (3.2.0): fails

```
 + python -m pytest --pyargs nilearn --cov=nilearn
ImportError while loading conftest '/home/runner/work/nilearn/nilearn/nilearn/conftest.py'.
nilearn/conftest.py:5: in <module>
    import nibabel
/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/nibabel/__init__.py:43: in <module>
    from . import nifti1 as ni1
/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/nibabel/nifti1.py:24: in <module>
    from .quaternions import fillpositive, quat2mat, mat2quat
/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/nibabel/quaternions.py:31: in <module>
    MAX_FLOAT = np.maximum_sctype(np.float)
/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/numpy/__init__.py:324: in __getattr__
    raise AttributeError(__former_attrs__[attr])
E   AttributeError: module 'numpy' has no attribute 'float'.
E   `np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
E   The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
E       https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations. Did you mean: 'cfloat'?
```

- minimal version of numpy (1.19.0): passes
- minimal version of nibabel and nibabel: passes

I don't think we need to check all the possible combinations on every PR but I think that it would be good to have a way to test those easily before we make a new release.

Remi-Gau:
- Closes #4069


bthirion (commented):
LGTM so far.

bthirion (approved):
LGTM.

ymzayek (approved):
LGTM!

PR #4071, opened 2023-10-19
[FIX] fix SimpleRegressionResults 

### LINKED ISSUE #2375: Small bugs in SimpleRegressionResults
(1) https://github.com/nistats/nistats/blob/master/nistats/regression.py#L366
`predicted` should be a method call instead of an instance variable reference
```
  File "nistats_model_fitting.py", line 22, in compute_rsquared
    print(model.results_[0][0.0].resid(Y))
  File "/.../nistats/nistats/regression.py", line 366, in resid
    return Y - self.predicted
TypeError: unsupported operand type(s) for -: 'float' and 'method'
```
(2) https://github.com/nistats/nistats/blob/master/nistats/regression.py#L392
`model` is not defined in `SimpleRegressionResults`. It is defined in the parent class `LikelihoodModelResults` but the initializer for the parent is never called.
```
  File "nistats_model_fitting.py", line 22, in compute_rsquared
    print(model.results_[0][0.0].predicted())
  File "/.../nistats/nistats/regression.py", line 392, in predicted
    X = self.model.design
AttributeError: 'SimpleRegressionResults' object has no attribute 'model'
```

### LINKED ISSUE #2971: SimpleRegressionResults.__init__()
The documentation of `SimpleRegressionResults.__init__()` has been copied "as is" from `RegressionResults.__init__()` and is obviously wrong:

#### [SimpleRegressionResults](https://github.com/nilearn/nilearn/blob/e91e02d/nilearn/glm/regression.py#L466-L471)
```python
class SimpleRegressionResults(LikelihoodModelResults):
    [...]
        """See LikelihoodModelResults constructor.

        The only difference is that the whitened Y and residual values
        are stored for a regression model.
        """
```
#### [RegressionResults](https://github.com/nilearn/nilearn/blob/e91e02d/nilearn/glm/regression.py#L340-L347)
```python
class RegressionResults(LikelihoodModelResults):
    [...]
        """See LikelihoodModelResults constructor.

        The only difference is that the whitened Y and residual values
        are stored for a regression model.

        """
```

See discussion in https://github.com/nilearn/nilearn/pull/2952#issuecomment-919751239

### Affected pages

https://nilearn.github.io/modules/generated/nilearn.glm.SimpleRegressionResults.html#nilearn.glm.SimpleRegressionResults.__init__

Remi-Gau:
- Closes #2375
- Closes #2971 

- based on the commits:
  - https://github.com/nilearn/nistats/pull/174/commits/985f4d5b238e244048446cc9877981e2f25a8425
  - https://github.com/nilearn/nistats/pull/174/commits/906b1b6e93e05fbe2ef16f789c8e6c89b7e4e21b
from @Gilles86 PR: https://github.com/nilearn/nistats/pull/174


nilearn/glm/_base.py
```diff
@@ -8,7 +7,7 @@ class BaseGLM(BaseEstimator, TransformerMixin, CacheMixin):
     """Implement a base class \
     for the :term:`General Linear Model<GLM>`."""
 
-    @auto_attr
+    @property
```
bthirion, line 10: I have no issue with the change, but could you explain what it means/implies ?

bthirion (commented):
LGTM overall.

Remi-Gau:
# TODO

- [ ] update changelog

Remi-Gau:
TODO update changelog

nilearn/glm/_base.py
Remi-Gau: As far as I can tell these decorators do the same thing but one comes from the standard library, so feels like it is more appropriate to use it.

Do you want me to mention it in the changelog?

ymzayek: IIUC it's not exactly the same, as property doesn't store the value as an object attribute after initial call (see https://nipy.org/nibabel/reference/nibabel.onetime.html#auto-attr). While using property works as well it seems it could add function call overhead that could affect performance. You could time some tests that we already have to quickly compare one versus the other.

Remi-Gau:
I think that this PR could also address https://github.com/nilearn/nilearn/issues/2971

Remi-Gau:
closed by mistake

bthirion (approved):
Sounds good, thx.

doc/changes/latest.rst
```diff
@@ -15,6 +15,7 @@ Fixes
 -----
 
 - :bdg-success:`API` :class:`~maskers.MultiNiftiMasker` can now call :meth:`~maskers.NiftiMasker.generate_report` which will generate a report for the first subject in the list of subjects (:gh:`4001` by `Yasmin Mzayek`_).
+- :bdg-dark:`Code` Fix :class:`~nilearn.glm.regression.SimpleRegressionResults` to accommodate for the lack of a ``model`` attribute (:gh:`4071` `Remi Gau`_)
```
ymzayek, line 18:
```suggestion
- :bdg-dark:`Code` Fix :class:`~nilearn.glm.regression.SimpleRegressionResults` to accommodate for the lack of a ``model`` attribute (:gh:`4071` `Rémi Gau`_)
```

ymzayek (approved):
Still not quite sure we need to change the decorator but LGTM!

Remi-Gau:
will run some benchmark to check

bthirion:
Can be merged I think ?

nilearn/glm/_base.py
Remi-Gau: > IIUC it's not exactly the same, as property doesn't store the value as an object attribute after initial call (see https://nipy.org/nibabel/reference/nibabel.onetime.html#auto-attr). While using property works as well it seems it could add function call overhead that could affect performance. You could time some tests that we already have to quickly compare one versus the other.

You were right @ymzayek it would make the code quite a bit slower.

Ran the following script on this branch and on main.

```
"""Benchmark for getting GLM residuals

Runs a first level GLM N_ITER time on dummy data
and gets the residuals n times (between 0 and N_ITER_RESIDUALS times).

Gets the time taken to for 1 iteration (fit GLM and get residual n times).

On main I am at about 0.197 seconds per iteration.

On this branch it increases up to 0.943 seconds per iteration if I try to get the residuals 10 times.

So yeah will revert those decorators.

"""python
    generate_fake_fmri_data_and_design,
)
from nilearn.glm.first_level import (
    FirstLevelModel,
)

import time

def compute_residuals(mask, fmri_data, design_matrices):
    model = FirstLevelModel(
        mask_img=mask, minimize_memory=False, noise_model="ols"
    )
    model.fit(fmri_data, design_matrices=design_matrices)
    return model.residuals

N_ITER = 20
N_ITER_RESIDUALS = 11

def main():

    shapes, rk = [(20, 20, 20, 100)], 3
        shapes, rk
    )

    for i in range(len(design_matrices)):
        design_matrices[i][design_matrices[i].columns[0]] = 1

    for n_iter_residuals in range(1, N_ITER_RESIDUALS):

        t0 = time.monotonic_ns()
        for _ in range(N_ITER):
            model = FirstLevelModel(
                mask_img=mask, minimize_memory=False, noise_model="ols"
            )
            model.fit(fmri_data, design_matrices=design_matrices)
            # get the residuals several times to see if this makes a difference
            for __ in range(n_iter_residuals):
                model.residuals
        t1 = time.monotonic_ns()

        print(f"Get residuals {n_iter_residuals} times")
        print(f"{(t1-t0)/10**9/N_ITER:0.3f} seconds per iteration")


if __name__ == "__main__":
    main()

```

PR #4076, opened 2023-10-23
[ENH] Improve error message when missing regressors in FirstlevelGLM for t contrasts

### LINKED ISSUE #2401: Support missing regressors in FirstlevelGLM
Currently, the occurrence of different sets of trial types across sessions in a FirstLevelGLM raises an error, because the model expects the conditions to be identical in all sessions.
This is an issue with randomized designs, where one of the conditions is absent.

Alternatively ---or as a first step ?--- Nistats should provide an example where this is handled with a separate fit of the different sessions and then an explicit fixed effects analysis.

Remi-Gau:
- Closes #2401

- only applies for expression contrasts (e.g. `"c0 - c1"`)
- if the contrast expression cannot be evaluated then an improved message is thrown


nilearn/glm/tests/test_first_level.py
```diff
@@ -1798,3 +1798,29 @@ def test_missing_trial_type_column_warning(tmp_path_factory):
             "No column named 'trial_type' found" in r.message.args[0]
             for r in record
         )
+
+
+def test_first_level_from_bids_with_one_condition_missing(tmp_path_factory):
+    """One condition is missing in one events.tsv file."""
+    bids_dataset = _new_bids_dataset(
+        tmp_path_factory.mktemp("one_coindition_missing")
```
bthirion, line 1806:
```suggestion
        tmp_path_factory.mktemp("one_condition_missing")
```

nilearn/glm/tests/test_first_level.py
```diff
@@ -1798,3 +1798,29 @@ def test_missing_trial_type_column_warning(tmp_path_factory):
             "No column named 'trial_type' found" in r.message.args[0]
             for r in record
         )
+
+
+def test_first_level_from_bids_with_one_condition_missing(tmp_path_factory):
+    """One condition is missing in one events.tsv file."""
```
bthirion, line 1804: This is supposed (not) to trigger something ?

Remi-Gau: sorry I should have mention that this is not ready for review yet: opened a draft PR instead of just stashing

nilearn/glm/contrasts.py
```diff
@@ -23,20 +24,30 @@ def expression_to_contrast_vector(expression, design_columns):
...
         The column names of the design matrix.
 
     """
     if expression in design_columns:
         contrast_vector = np.zeros(len(design_columns))
         contrast_vector[list(design_columns).index(expression)] = 1.0
         return contrast_vector
+
     df = pd.DataFrame(np.eye(len(design_columns)), columns=design_columns)
+    # contrast_vector = df.eval(expression, engine="python").values
```
Remi-Gau, line 40:
```suggestion
```

Remi-Gau:
@bthirion 

the whole implementation relies on catching an error thrown by pandas when evaluating the expression.

Does this seem ok to you?

bthirion (commented):
I'm worried we may arrive to weird situations where the user expects a contrast to have been computed and won't find it. To me contrasts and fixed effects should be defined consistently in all runs of a given fit.

Remi-Gau:
> To me contrasts and fixed effects should be defined consistently in all runs of a given fit.

Can you explain a bit more?

- If by *consistently* you mean that "all runs should have all conditions" then in that case shouldn't we just close #2401.

bthirion:
Yes, that's the idea. 
But we can revise this  strategy if we think it's too conservative. Maybe something to discuss in an upcoming coredev meeting.

Remi-Gau:
that may add quite a bit of friction for users who have designs that are "response" driven (trials where subject has given a certain response): not sure how frequent those are though.

In any case the error that is thrown at the moment is not helping user know what is wrong (it just says something is wrong with the expression), so I can at least recycle this PR to improve this.

bthirion:
We probably need to document current behavior better.

Remi-Gau:
@bthirion 
reworked this PR to only improve error message to try to hint at what conditions are missing

hndgzkn (approved):
LGTM

bthirion (approved):
LGTM, thx.

PR #4079, opened 2023-10-24
[DOC] Fix failing docs on main

ymzayek:
<!---


Remi-Gau:
thanks @ymzayek 

though I admit I am still confused as to why it works like this.

ymzayek:
> though I admit I am still confused as to why it works like this.

@Remi-Gau It's just the way we have it set up in doc/modules for glm. Autodoc will basicaly use these rst files as configuration files (see https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#). I don't know the ins and outs of it as I wasn't involved in setting it up but happy to discuss it in more detail

PR #4086, opened 2023-10-27
[MAINT] remove leading underscore from non private functions

### LINKED ISSUE #3319: Check for inconsistent naming of functions according to PEP8 convention
As pointed out in https://github.com/nilearn/nilearn/pull/3300#discussion_r929071958 I think it is worth checking for inconsistencies in function naming in nilearn. It might also be a good idea to discuss any ambiguities regarding function naming.

Remi-Gau:
- Relates to #3319

- remove leading underscore from functions used outside of their own module in:
  - nilearn/*.py
  - nilearn/decoding
  - nilearn/datasets
  - nilearn/mass_univariate
- rename module `nilearn/decoding.objective_functions.py` to `nilearn/decoding._objective_functions.py`
- keep non user facing functions from `nilearn/datasets/utils.py` into a new file `nilearn/datasets/_utils.py`


ymzayek (commented):
I generally see a bit of an issue with some files that I think should be named with a leading underscore (e.g. having functions in a file not named with a leading underscore that are used outside of that file but only within their respective parent module). So to me naming such functions with a leading underscore was done to signal that they are private to that module. So in short I think it might be worth discussing how to reconcile the way certain modules are structured with the guidelines we defined.

nilearn/decoding/proximal_operators.py
```diff
@@ -7,7 +7,7 @@
 
 import numpy as np
 
-from .objective_functions import _div_id, _gradient_id, _tv_l1_from_gradient
+from .objective_functions import _div_id, _gradient_id, tv_l1_from_gradient
```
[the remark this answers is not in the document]
Remi-Gau: because my detection algo is dumb and has false negative

_div_id and _gradient_id are mentioned more than once in proximal_operators.py so I did not look at them yet

nilearn/datasets/utils.py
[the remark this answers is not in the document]
Remi-Gau: create a _utils.py module where we put the private stuff?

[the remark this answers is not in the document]
ymzayek: Sure if it's that simple I like it

Remi-Gau:
> So in short I think it might be worth discussing how to reconcile the way certain modules are structured with the guidelines we defined.

fine though I think the function level naming rule would still apply even if we renamed the module where it comes from

nilearn/decoding/objective_functions.py
ymzayek, line 1: Regarding my general comment, what would we do about this file for example? It seems all the functions here are only used in the decoding module so we should name it with a leading underscore. This one has no functions exposed in the API in the docs so I guess safe to change the name?

nilearn/datasets/utils.py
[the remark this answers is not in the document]
Remi-Gau: I guess we would have to make it uber explicit in the doc string of the modules why there is a _utils and utils in the same folder...

nilearn/decoding/objective_functions.py
Remi-Gau: rename the file to _objective_functions.py and remove leading underscore from all its functions then?

ymzayek: No only the ones imported outside of it. Again just following the example in the guidelines, `decoding._objective_functions.py` should be considered like the `maskers._validation.py` example. It can have internal private functions with a leading underscore but the ones imported elsewhere within the decoding module need to  be removed. So `_sigmoid` should stay as is for example.

Remi-Gau: oh sorry I misread you previous comment 

agreed

Remi-Gau:
will make the change we agreed on and will probably not do much more in this PR to keep things reviewable

Remi-Gau:
in a separate PR:

rename either `nilearn.datasets.utils._fetch_file` or `nilearn.datasets.utils._fetch_files`

but having both of them differing by a single letter is just not OK

nilearn/datasets/_utils.py
Remi-Gau, line 1: OK the number of line changed blew up because of splitting this file in 2

nilearn/decoding/proximal_operators.py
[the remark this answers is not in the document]
Remi-Gau: actually now I have renamed them too

doc/changes/0.10.1.rst
```diff
@@ -29,7 +29,7 @@ Fixes
 
 - :bdg-success:`API` :func:`~glm.second_level.non_parametric_inference` can accept first level model as input without failing. (:gh:`3600` by `Rémi Gau`_)
 
-- :bdg-danger:`Deprecation` Add correct "zscore_sample" strategy to ``signal._standardize`` which will replace the default "zscore" strategy in release 0.13  (:gh:`3474` by `Yasmin Mzayek`_).
+- :bdg-danger:`Deprecation` Add correct "zscore_sample" strategy to ``signal.standardize`` which will replace the default "zscore" strategy in release 0.13  (:gh:`3474` by `Yasmin Mzayek`_).
```
Remi-Gau, line 32:
```suggestion
- :bdg-danger:`Deprecation` Add correct "zscore_sample" strategy to ``signal.standardize_signal`` which will replace the default "zscore" strategy in release 0.13  (:gh:`3474` by `Yasmin Mzayek`_).
```

nilearn/decoding/_objective_functions.py
```diff
@@ -225,7 +225,7 @@ def _sigmoid(t, copy=True):
     return t
 
 
-def _logistic(X, y, w):
+def logistic_loss(X, y, w):
```
Remi-Gau, line 228: note the function name change

nilearn/decoding/_objective_functions.py
```diff
@@ -267,16 +267,16 @@ def _logistic_loss_grad(X, y, w):
...
 
-def _gradient(w):
+def gradient(w):
     """Pure spatial gradient."""
-    return _gradient_id(w, l1_ratio=0.0)[:-1]  # pure nabla
+    return gradient_id(w, l1_ratio=0.0)[:-1]  # pure nabla
 
 
-def _div(v):
+def divergence(v):
```
Remi-Gau, line 280: note the function name change

nilearn/decoding/_objective_functions.py
```diff
@@ -122,7 +122,7 @@ def _tv_l1_from_gradient(spatial_grad):
     return l1_term + tv_term
 
 
-def _div_id(grad, l1_ratio=0.5):
+def divergence_id(grad, l1_ratio=0.5):
```
Remi-Gau, line 125: note the function name change

Remi-Gau:
Eventhough the functions changed here are private and therefore should not be used by users, I think that those changes should be mentioned in the changelog because I would be surprised if not a single users used some of those utility functions directly in their code.

any thoughts?

nilearn/plotting/html_surface.py
```diff
@@ -121,7 +121,8 @@ def one_mesh_info(
                           vmax=vmax, vmin=vmin)
 
 
-def _check_mesh(mesh):
+def check_mesh(mesh):
+    """Validate type and content of a mesh."""
```
bthirion, line 125: But then we need a full docstring here, right ?

bthirion (commented):
Looks good overall, thx.

nilearn/plotting/html_surface.py
Remi-Gau: Just specify the input type in this case? Sure I can add that though I don't think that we should start adding doc strings just because we remove the underscore.

bthirion: But this is part of the public API, isn't it ?

Remi-Gau: depends public with respect to what
it is public to the `nilearn/plotting/html_surface.py` module but not listed in the `__all__` of `nilearn/plotting/__init__.py` so not exposed to the user

ymzayek: Note that what is defined in inits under `__all__` is not necessarily what is or should be exposed to the user. Take for example everything under `_utils`. We should instead look at what is defined under `doc/modules`. I think the confusion comes from the fact that the leading underscore is a naming convention and it can be done on different levels to imply that an object should be used internally. Additional confusion comes from inconsistencies in nilearn as well as cases where it might be difficult to decide clearly how to name an object based on e.g. its location. Even though we have defined some guidelines, I find that there are quite a few cases where its not so straightforward to follow them

Remi-Gau: I agree the guidelines are neither simple nor convenient.

Well one good thing about https://github.com/nilearn/nilearn/pull/2991 is it may help us at operationalize things a bit.

The thing that irks me is that the private / public nature function in question here has not changed, so I don't see the point of now adding an extensive doc string that was not there before.

The minimal doc string I added was just to pacify flake8.

nilearn/signal.py
```diff
@@ -772,12 +774,12 @@ def clean(
...
     # Remove confounds
     if confounds is not None:
-        confounds = _standardize(
+        confounds = standardize_signal(
             confounds, standardize=standardize_confounds, detrend=False
         )
         if not standardize_confounds:
             # Improve numerical stability by controlling the range of
-            # confounds. We don't rely on _standardize as it removes any
+            # confounds. We don't rely on standardize as it removes any
```
ymzayek, line 782:
```suggestion
            # confounds. We don't rely on standardize_signal as it removes any
```

nilearn/datasets/utils.py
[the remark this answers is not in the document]
Remi-Gau: agreed
we would end doing something similar to what we have in input_data: https://github.com/nilearn/nilearn/blob/main/nilearn/input_data/__init__.py

nilearn/plotting/html_surface.py
bthirion: Sorry, I checked too superficially. You can ignore my comment.

bthirion (approved):
LGTM.

ymzayek (approved):
@Remi-Gau the PR LGTM!

One thing I'm a bit unsure about is that now in some cases as in the change of `_check_mesh` to `check_mesh`, how do we signal to the user and future developers that this function should be used internally (which also implies they can change without deprecation)? You can argue that it is not exposed in the API reference docs but some people don't look at that. One way could be to not include such functions in `__all__` and explicitly use the module.file name when calling the function as is now done with `check_mesh`. But then we'd need to make this consistent across modules (with the exceptions being _utils) making the content of `__all__` consistent with the API references in the docs. I'm not sure if this is the way to go so I'm happy to hear other thoughts

Remi-Gau:
For context worth checking what pep8 says:
https://peps.python.org/pep-0008/#public-and-internal-interfaces

> how do we signal to the user and future developers that this function should be used internally (which also implies they can change without deprecation)? 

I agree that, unless I misunderstand something about PEP8, it feels that users would always need to go and look into the "all" of the __init__ of a module to figure out if a given function is public or private: not convenient.

> You can argue that it is not exposed in the API reference docs but some people don't look at that. 

Well then in that case, it's not on us. We cannot control what people do or do not do. We can control how clear our doc is.

> One way could be to not include such functions in `__all__` 

yes

> and explicitly use the module.file name when calling the function as is now done with `check_mesh`. 

So do not do 

```python
from surface import check_mesh

check_mesh(foo)
```

but do

```python
import surface

surface.check_mesh(foo)
```

If so that still does not signal that check_mesh is private.

> But then we'd need to make this consistent across modules (with the exceptions being _utils) making the content of `__all__` consistent with the API references in the docs

In any case I would be in favor of this.

nilearn/decoding/proximal_operators.py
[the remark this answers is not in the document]
Remi-Gau: will do in a separate PR

nilearn/datasets/utils.py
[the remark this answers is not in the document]
Remi-Gau: will do in a separate PR

Remi-Gau:
FYI: we should expect a few more PRs like this one as this only touches a few modules

Remi-Gau:
Ok the listing utilities test is doing its job as "number of public function change detector" is working as intended as shown by the latest failures.

Remi-Gau:
Unless anyone has second thoughts about this one, I will merge it tomorrow and then open separate issues + PR for the things mentioned in here.

PR #4094, opened 2023-10-30
[MAINT] Rewrite `write_tmp_imgs` function and use it with pytest's `tmp_path` 

ymzayek:
Relates to https://github.com/nilearn/nilearn/pull/3935#issuecomment-1779525153

- The function now looks like other testing helper functions that write data to disk by accepting a file_path argument. Now unit tests can pass it the pytest `tmp_path` fixture letting pytest take care of setting up a tmp directory and cleaning up the files.


nilearn/_utils/testing.py
```diff
@@ -113,12 +113,10 @@ def serialize_niimg(img, gzipped=True):
             return f.read()
 
 
-@contextlib.contextmanager
-def write_tmp_imgs(*imgs, **kwargs):
-    """Context manager for writing Nifti images.
+def write_tmp_imgs(*imgs, file_path=None, **kwargs):
```
ymzayek, line 116: Maybe the name should also be changed to something like `write_fake_imgs`

Remi-Gau: agreed

Remi-Gau (approved):
Except for the name change it looks good to me.

Thanks a bunch for taking care of this one.

Realizing that refactoring this context manager had been on my mind for a while, so defo happy to see it go.

ymzayek:
I would like to rebase #3935 on this so hoping to merge soon. Another review would be great!

bthirion (commented):
I'm mostly happy with the PR, but ---again--- with the naming change from `write_tmp_imgs` to ẁrite_fake_imgs, as there is nothing in this function that implies that images are fake.

Remi-Gau:
I guess that we could just call it `write_imgs` as they are not "fake" but they are not temporary unless when written in a tmp_dir

ymzayek:
Yes makes sense. What about `write_imgs_to_path`?

ymzayek:
> I guess that we could just call it `write_imgs` as they are not "fake" but they are not temporary unless when written in a tmp_dir

Or yea just this one

Remi-Gau:
either are ok with me

PR #3285, opened 2022-06-18
[FIX] Load compcor related confounds generated from fmriprep 21.x series

htwangtw:
Closes #3268. 

- refactor the test file loader to allow adding new major release 
- add test data for 21.x.x and make sure the correct files are loaded
- apply compcor loading suggestions from @askieslinger and iterate on it
- simply `_check_compcor_method` by using all five options as prefix checking keys
- add tests to `test_load_confounds_strategy.py::test_strategy_compcor` and `test_load_confounds.py::test_n_compcor` to make sure the correct components are loaded for 21.x and older

PR #4103, opened 2023-11-09
[ENH] integrate load_confounds into first_level_from_bids

### LINKED ISSUE #4090: [ENH] integrate `load_confounds` into `first_level_from_bids`

- Relates to https://github.com/nilearn/nilearn/issues/2425

Currently one can point `first_level_from_bids` to a bids dataset and get
- model object
- images paths
- events files content
- confounds files content

The whole of the confounds are loaded, it is left to the user to confound wrangling before they can fit their model.

Alternatively they can use the load_confounds function and pass it images paths to load the confounds with one of the strategy (scrubbing, motion...) it accepts.

It may be easier to pass directly some extra argument to `first_level_from_bids` and have it take care of call load_confounds returning only the subset of confounds requested.

### Benefits to the change

make it simpler to load a given set of confounds from a BIDS dataset when running a GLM

currently

```python
from nilearn.glm.first_level import first_level_from_bids

task_label = "stopsignal"
space_label = "MNI152NLin2009cAsym"
data_dir = "/home/bob/data/stop_signal_fmri/data"
derivatives_folder = "derivatives/fmriprep"

(
    models,
    models_run_imgs,
    models_events,
    _,
) = first_level_from_bids(
    data_dir,
    task_label,
    space_label,
    smoothing_fwhm=5.0,
    derivatives_folder=derivatives_folder,
)

# the following would actually have to be done for every subject
models_confounds = load_confounds(
    models_run_imgs[0],
    strategy=("motion", "wm_csf"),
    motion="full",
    scrub=1,
    fd_threshold=0.2,
    wm_csf="basic")
```

what I suggest

```python
from nilearn.glm.first_level import first_level_from_bids

task_label = "stopsignal"
space_label = "MNI152NLin2009cAsym"
data_dir = "/home/bob/data/stop_signal_fmri/data"
derivatives_folder = "derivatives/fmriprep"

(
    models,
    models_run_imgs,
    models_events,
    models_confounds,
) = first_level_from_bids(
    data_dir,
    task_label,
    space_label,
    smoothing_fwhm=5.0,
    derivatives_folder=derivatives_folder,
    strategy=("motion", "wm_csf"),
    motion="full",
    scrub=1,
    fd_threshold=0.2,
    wm_csf="basic")
```

Remi-Gau:
- Closes #4090

- use load_confounds to load specific subset of confounds if extra `confounds_*` arguments are passed
- update fake bids data generation so they contain more realistic confounds

## TODO
- [x] update changelog
- [x] investigate conflicts between confounds strategy and GLM detrenting
- [ ] add in example?


nilearn/_utils/bids.py
Remi-Gau, line 1: had to move the create_bids_filename in a different module to help avoid circular imports

nilearn/interfaces/fmriprep/tests/_testing.py
Remi-Gau, line 1: some of these changes in here are black related

may conflict with https://github.com/nilearn/nilearn/pull/3285

nilearn/glm/tests/test_first_level.py
```diff
@@ -1860,3 +1860,30 @@ def test_missing_trial_type_column_warning(tmp_path_factory):
...
+        space_label="MNI",
+        img_filters=[("desc", "preproc")],
+        slice_time_ref=None,
+        confounds_strategy=("motion", "wm_csf", "scrub"),
+        confounds_motion="full",
+        confounds_wm_csf="basic",
+        confounds_scrub=1,
+        confounds_fd_threshold=0.2,
+        confounds_std_dvars_threshold=3,
+    )
```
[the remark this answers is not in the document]
Remi-Gau: @htwangtw @ymzayek @bthirion 
keeping this as a draft so you can discuss API, implementation...

Will update the doc once the dust settles.

Remi-Gau:
one thing we may want to check and send warning about the compatibility between some confound strategies and first level arguments: especially regarding high pass filters

nilearn/_utils/data_gen.py
```diff
@@ -1390,12 +1357,15 @@ def _write_bids_derivative_func(
...
             fields=fields, entities_to_include=_bids_entities()["raw"]
         )
-        _basic_confounds(length=n_time_points, random_state=rand_gen).to_csv(
-            confounds_path, sep="\t", index=None
+        confounds, metadata = get_legal_confound()
+        confounds.to_csv(
+            confounds_path, sep="\t", index=None, encoding="utf-8"
         )
+        with open(confounds_path.with_suffix(".json"), "w") as f:
+            json.dump(metadata, f)
```
Remi-Gau, line 1368: One issue with this approach is that the "legal_confounds" have a set number of time points so when creating a fake bids dataset, we end up with images that have a number of time points that does not match the number of time points in the confounds.

This does not affect any tests AFIACT but this may lead to confusing errors when testing down the line.

nilearn/_utils/data_gen.py
```diff
@@ -1390,12 +1358,15 @@ def _write_bids_derivative_func(
...
```
Remi-Gau, line 1369: minor change: using legal_confounds allows to add metada files for the confounds in the fake bids derivatives. some tests had to be changed to account for this.

Remi-Gau:
Thought: should this functionality be demoed in the examples?

nilearn/_utils/data_gen.py
bthirion: You mean, because of scrubbing ? Sorry if I miss something obvious.

nilearn/glm/tests/test_first_level.py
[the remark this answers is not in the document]
bthirion: I guess this could be made lighter if we expect to have predefined configurations for confounds, but I don't think that we're at that point.
At least I find the current API quite explicit.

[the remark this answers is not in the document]
bthirion: Actually, we might get rid of `confounds_strategy`, because the next three arguments are redundant with the provided list ?

nilearn/_utils/data_gen.py
Remi-Gau: nope

let me try to rephrase

the way to generate "fake confounds" for the fake fmriprep datasets we use for testing would only create 6 confounds for the realignment parameters filled with random data and for a specified number of time points

to allow testing the load_confounds we need more realistic confounds with more columns with names that match what's in an actual fmriprep dataset

to do this I reuse the strategy used to test the load_confounds functions: use an actual confound file from an fmriprep dataset and copy its content every time it is needed in the fake fmriprep dataset

but this "template" confound file has only a limited number of time points

so we end up with fake fmriprep datasets that have nifti images with 100 volumes but with confounds with only 30 time points

possible solutions:
- easy: set the number of volumes to match the number of time points in the confounds
- hard(er): adapt the content of the confounds to the number of time points

hope this is clearer

for now I will go for the easy solution but we may have to implement the harder option in the future if we want to test more "exotic" stuff

nilearn/glm/tests/test_first_level.py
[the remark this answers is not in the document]
Remi-Gau: Note this test just show a subset of the possible arguments that can be passed to load_confounds

> I guess this could be made lighter if we expect to have predefined configurations for confounds, but I don't think that we're at that point.

> Actually, we might get rid of `confounds_strategy`, because the next three arguments are redundant with the provided list ?

Actually I am just reusing the API from load_confounds, so it can almost be passed as is and we can let load_confounds do the argument validation

https://nilearn.github.io/dev/modules/generated/nilearn.interfaces.fmriprep.load_confounds.html

In short `strategy` defines what type of confounds to include and all the other parameters give more details on how to include them.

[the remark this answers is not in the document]
Remi-Gau: will update the doc string to try to explain this so we can see if it makes sense

nilearn/glm/first_level/first_level.py
```diff
@@ -1015,6 +1017,86 @@ def first_level_from_bids(
...
+
+    Keyword arguments to be passed to functions called within this function.
+
+    Kwargs prefixed with ``confound_``
+    will be passed to :func:`~nilearn.interfaces.fmriprep.load_confounds`.
+    This allows to ``first_level_from_bids`` to return
+    a specific set of confounds by relying confound loading strategies
+    defined in :func:`~nilearn.interfaces.fmriprep.load_confounds`.
+    If no kwargs are passed, ``first_level_from_bids`` will return
+    all the confounds available in the confounds TSV files.
```
Remi-Gau, line 1032: @bthirion let me know if this helps clarify how to use this.

I prefer to add examples here and refer users to the doc of load_confounds for the details to avoid doc duplications.

nilearn/_utils/data_gen.py
bthirion: Let's go for the easy one. The number of volumes should be a parameters of the data simulation function anyhow ?

nilearn/glm/first_level/first_level.py
bthirion: Agreed !

nilearn/_utils/data_gen.py
Remi-Gau: for now it is not: it is hard coded. I would keep it that way for now until we need more flexibility during testing.

but I will change the place where it is hard coded so it is easier to adapt in the future. will also add a comment to explain why this value was chosen.

bthirion:
Is this ready for review ?

nilearn/glm/first_level/first_level.py
```diff
@@ -1015,6 +1017,86 @@ def first_level_from_bids(
...
+
+    Examples
+    --------
+    If you want to only load
+    the rotation and translation motion parameters confounds:
+
+    .. code-block:: python
+
+        models, m_imgs, m_events, m_confounds = first_level_from_bids(
```
bthirion, line 1041:
```suggestion
        models, imgs, events, confounds = first_level_from_bids(
``` (not sure what the `m_` means)

File: nilearn/glm/first_level/first_level.py [hunk 6]
```diff
@@ -1015,6 +1017,86 @@ def first_level_from_bids(
...
+            confounds_strategy=("motion"),
+            confounds_motion="basic",
+        )
+
+    If you want to load the motion parameters confounds
+    with their derivatives:
+
+    .. code-block:: python
+
+        models, m_imgs, m_events, m_confounds = first_level_from_bids(
```
bthirion, line 1055: Same thing for the next examples.

bthirion (commented):
LGTM overall !

nilearn/glm/first_level/first_level.py
Remi-Gau: That was a copy-pasta from the code we have in the tests: probably could change it there too.

Remi-Gau: FYI those were shorter form of the name of the return argument:
- models_run_imgs
- models_events
- models_confounds

nilearn/glm/first_level/first_level.py
```diff
@@ -1015,6 +1017,86 @@ def first_level_from_bids(
...
 
+    kwargs: :obj:`dict`
+
+        .. versionadded:: 0.11.0
+
+    Keyword arguments to be passed to functions called within this function.
+
+    Kwargs prefixed with ``confound_``
+    will be passed to :func:`~nilearn.interfaces.fmriprep.load_confounds`.
+    This allows to ``first_level_from_bids`` to return
```
ymzayek, line 1028:
```suggestion
    This allows ``first_level_from_bids`` to return
```

nilearn/glm/first_level/first_level.py
```diff
@@ -1015,6 +1017,86 @@ def first_level_from_bids(
...
+    a specific set of confounds by relying confound loading strategies
```
ymzayek, line 1029:
```suggestion
    a specific set of confounds by relying on confound loading strategies
```

nilearn/glm/first_level/first_level.py
```diff
@@ -1613,6 +1707,36 @@ def _check_args_first_level_from_bids(
...
+        "compcor": "anat_combined",
+        "n_compcor": "all",
+        "ica_aroma": "full",
+        "demean": True,
+    }
+
+    if "confounds_strategy" not in kwargs:
+        return None
+    if kwargs["confounds_strategy"] is None:
+        return None
```
ymzayek, line 1728:
```suggestion
    if kwargs.get("confounds_strategy") is None:
        return None
```

ymzayek (commented):
Looking good!

ymzayek:
I'm not against demoing in an example. I would always just consider the amount of additional build time, whether or not it is already well documented (which the added docstring does very well), and if parts of an example can be replaced or minimally tweaked to essentially improve them using this new functionality.

Remi-Gau:
> I'm not against demoing in an example. I would always just consider the amount of additional build time, whether or not it is already well documented (which the added docstring does very well), and if parts of an example can be replaced or minimally tweaked to essentially improve them using this new functionality.

I will do a draft in a separate PR but I was considering a minimal tweak to an already existing example.

nilearn/glm/first_level/first_level.py
bthirion: I would prefer  `models, imgs, events, confounds = first_level_from_bids`

Remi-Gau: @bthirion I renamed those variable in the doc strings.

they also appear like this in a few tests, I could do a bit of renaming there too just for internal consistency.

Remi-Gau:
"Conflict" to resolve:

## load_confounds

- load_confounds has the possibility to add a high pass filter to the confounds

```
- "high_pass" adds discrete cosines transformation basis regressors to handle low-frequency signal drifts.
```

and when the compcor strategy is requested "high_pass" must be as well

```
- "compcor" confounds derived from CompCor :footcite:`Behzadi2007`.
  When using this noise component, "high_pass" must also be applied.
  Associated parameter: `compcor`, `n_compcor`
```


## GLM first level

drift_model can be cosine or polynomial or none

```
    drift_model : string, default='cosine'
        This parameter specifies the desired drift model for the design
        matrices. It can be 'polynomial', 'cosine' or None.

    high_pass : float, default=0.01
        This parameter specifies the cut frequency of the high-pass filter in
        Hz for the design matrices. Used only if drift_model is 'cosine'.

    drift_order : int, default=1
        This parameter specifies the order of the drift model (in case it is
        polynomial) for the design matrices.
```

There are some combinations of the above that could lead to "strange" design matrices:
- design matrices with high pass filter defined twice (with possibly different types of high pass filters)

To keep things simple I would say that we let the GLM machinery handle the high pass filtering setting and we ignore in the load_confounds anything that has to do with high pass filtering.

This means that we should also ignore (for now) the compcor strategy.

Remi-Gau:
@bthirion before I start on this, can you tell me if the brief description above makes roughly sense as to what the problem is?

bthirion:
You mean, the problem of duplication of high-pass filtering with the use of compcorr ? My view on this is that this is the user responsibility. You can prevent people from doing wrong things. The point is to help them diagnose it easily using visualization of the design matrix they created. Does that answer your concern ?

Remi-Gau:
You can't prevent people from doing wrong things but you can add more friction to make it harder for them to do so. 😋

I think my views may be a bit more "paternalistic" (more tainted by automated pipelines and avoiding options that would considered wrong in most cases) but you have way more experience than me in the "nilearn philosophy" so I will gladly follow your lead on this. 

I will at least add some explicit warnings to tell them they have chosen options that may be redundant or clash with each other.

bthirion:
Agreed. My impression is that when we impose constraints, we may not foresee all practical consequences.
This is why I think it is important to have good examples or event tutorials that promote good patterns.

D3njo:
Hao-Ting Wang ***@***.***> schrieb am Mo., 11. Dez. 2023,
15:52:

nilearn/glm/tests/test_first_level.py
```diff
@@ -1325,17 +1325,15 @@ def test_first_level_from_bids(
         base_dir=tmp_path, n_sub=n_sub, n_ses=n_ses, tasks=tasks, n_runs=n_runs
     )
 
-    models, m_imgs, m_events, m_confounds = first_level_from_bids(
+    models, models, imgs, events, confounds = first_level_from_bids(
```
Remi-Gau, line 1328: renamed the variables as mentioned in the PR discussion

nilearn/glm/first_level/first_level.py
```diff
@@ -1050,6 +1132,23 @@ def first_level_from_bids(
...
+            if drift_model == "cosine":
+                verb = "duplicate"
+            if drift_model == "polynomial":
+                verb = "conflict with"
+
+            warn(
+                f"""Confounds will contain a high pass filter,
+ that may {verb} the {drift_model} one used in the model.
+ Remember to visualize your design matrix before fitting your model
+ to check that your model is not overspecified.""",
```
Remi-Gau, line 1148: Not sure about the phrasing of the warning.

htwangtw: looks reasonable to me

bthirion: To me toot.

bthirion (approved):
LGTM, thx.

Remi-Gau:
If I don't hear back from anyone I will merge this on Monday.

htwangtw (approved):
LGTM thanks for implementing this!

PR #4112, opened 2023-11-16
[ENH] add support for fitting first levels to numpy arrays directly

### LINKED ISSUE #4110: [ENH] support fitting first levels to numpy arrays directly, in addition to niimgs

We recently fit GLMs to edge timeseries (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC10369861/) in order to look at fluctuations of attention at the network level. We believe it's an exciting approach that offers novel information and which is really easy to implement.

Almost all of the code was based off of nilearn - it mostly involved just changing FirstLevelModel().fit to take numpy arrays, rather than niimgs, so that we could give it the 2D arrays (n_features x n_scans). 

It would be really great to bake this feature directly into nilearn, so that it can be easily kept up to date with other nilearn features and bug-fixes. I believe the changes to the code would be minimal, and add little-to-no maintenance, and I would be happy to submit the pull request myself.

I have some example pieces of potential changes below.

### Benefits to the change

This would add flexibility to the nilearn code for fitting first-level models - letting it fit edges, or even just data that is in a significantly different format (e.g., open data that was already preprocessed and parcellated with some atlas). It would cost very little to maintain. 

```python
# insert your code below

def fit(...):

if (self.mask_img is None) & (type(run_imgs[0])==np.ndarray):
    assert run_imgs[0].shape==2  # nFeatures x nScans
...

if type(run_img)==np.ndarray:
    Y = run_img
else:
    run_img = check_niimg(run_img, ensure_ndim=4)
    Y = self.masker_.transform(run_img, sample_mask=sample_mask)

if design_matrices is None:
     n_scans = Y.shape[-1]
     
     
def compute_contrast(...):

...

if self.masker_ is None:
    output = estimate_
else:
    output = self.masker_.inverse_transform(estimate_)



```

henrymj:
- Closes #4110

- adds function `fit_arrays` to FirstLevelModel; same as `fit`, but expects numpy arrays rather than niimgs
- adds `format_type` argument to FirstLevelModel.compute_contrast. If `array`, it returns contrast results as arrays, rather than images

## TODO
- add tests (probably don't need to replicate all of the high-level calls to .fit ... maybe just simulate fake data and see if the results are the same?)
- add example? (I know there isn't a plan to advertise this functionality)


henrymj:
I was following @bthirion 's suggestion of making a separate function, but that does sound like it could be easier.  Especially if it means it updates self.masker such that compute_contrast and get_voxelwise_attribute would work without modification, and users could just call get_data on the returned object to get back to 2D arrays.

Please pardon my ignorance - how would you cast numpy arrays with an arbitrary length along the first dimension (an arbitrary number of ROIs or edges being fit) into a niimg-like object?

bthirion:
I find it weird to transform arrays into a Nfiti1Image, from which we will extract the data array... It turns out that the GLM does not use any image information than the data array. So I find it better that way.  But I'm open to other opinions !

ymzayek:
> I find it weird to transform arrays into a Nfiti1Image, from which we will extract the data array... It turns out that the GLM does not use any image information than the data array. So I find it better that way. But I'm open to other opinions !

@bthirion yes but the fit method in `FirstLevelModel` has built in functionality of masking the data to extract the data array before fitting the glm and this is the blocker I see since the maskers fit and transform methods expect a Nfiti1Image. Do you suggest a new fitting function that does not include this functionality? If so, we need to make this explicit

bthirion:
> > I find it weird to transform arrays into a Nfiti1Image, from which we will extract the data array... It turns out that the GLM does not use any image information than the data array. So I find it better that way. But I'm open to other opinions !
> 
> @bthirion yes but the fit method in `FirstLevelModel` has built in functionality of masking the data to extract the data array before fitting the glm and this is the blocker I see since the maskers fit and transform methods expect a Nfiti1Image. Do you suggest a new fitting function that does not include this functionality? If so, we need to make this explicit

Yes, that's the idea.

ymzayek:
Ok I see thanks for clarifying. Then let's go ahead with the original plan @henrymj

Remi-Gau:
Will fix the conflict and the styling issues.

Remi-Gau:
This will need a changelog update.

Remi-Gau (changes requested):
This can probably be made shorter by some refactoring to avoid code duplication.

For example the input checks section can probably extracted into a separate function (for all inputs common to fit and fit_arrays).

Remi-Gau:
@henrymj 

Do you think you would have time to do this refactoring?

Otherwise let me know and I can do it.

henrymj:
@Remi-Gau I would prefer to defer to you, as I am not confident I could efficiently refactor it in nilearn's style, and I am unfortunately very focused on an upcoming conference for the next month or so.

Remi-Gau:
No worries. I will take over and hopefully create something that can be sort of clear in terms of refactoring.

doc/changes/latest.rst
```diff
@@ -24,6 +24,7 @@ Enhancements
 
 - :bdg-dark:`Code` Add footer to masker reports (:gh:`4307` by `Rémi Gau`_).
 
+- :bdg-dark:`Code` Add a :meth:`~glm.first_level.FirstLevelModel.fit_array` method to first first level GLMs directly to 2D array (:gh:`4112` by TODO henrymj and `Rémi Gau`_).
```
Remi-Gau, line 27: @henrymj 

I am going to need some info to list you as contributor

See this example

```
authors:

  - given-names: Dimitri Papadopoulos
    family-names: Orfanos
    website: https://github.com/DimitriPapadopoulos
    affiliation: NeuroSpin, C.E.A., Université Paris-Saclay, Gif-sur-Yvette, France
    orcid: https://orcid.org/0000-0002-1242-8990
```

Remi-Gau:
@henrymj @bthirion 

Done with refactoring for this one:

- fit() and fit_arrays() are now MUCH shorter
- extracted the different steps in several smaller private methods (input validation, constructing design matrices, "logging", fit glm for each run) that are reused in fit() and fit_arrays()
- added extra input validation to make sure that the input to fit() and fit_arrays() actually matches what was in the doc or the tests (turns out some test ran with tuple of images not lists)
- added tests
- update doc

Number of line changes looks big but this is mostly some ["extract function" refactoring](https://refactoring.guru/extract-method)

nilearn/glm/first_level/first_level.py
```diff
@@ -490,6 +490,312 @@ def scaling_axis(self):
...
+            msg = "Performing mask computation."
+        elif step == "masking_done":
+            msg = f"Masking took {int(time_in_second)} seconds."
+        elif step == "done":
+            msg = (
+                f"Computation of {n_runs} runs done "
+                f"in {int(time_in_second)} seconds."
+            )
+
+        logger.log(msg, verbose=self.verbose)
```
Remi-Gau, line 569: - actually using nilearn's logger here (see #4443)

doc/changes/latest.rst
henrymj: - given-names: Henry
    family-names: Jones
    website: https://github.com/henrymj
    affiliation: Institute for Mind and Biology, University of Chicago, Chicago, IL, United States of America
    orcid: https://orcid.org/0000-0001-7719-3646

nilearn/glm/first_level/first_level.py
```diff
@@ -490,6 +490,312 @@ def scaling_axis(self):
...
+    ):
+        """Fit the :term:`GLM`.
+
+        For each run:
+        1. create design matrix X
+        2. fit regression to (Y, X)
+
+        Parameters
+        ----------
+        run_imgs : numpy array,\
```
bthirion, line 717: No, if it's an array, don't call to `run_imgs` !
`run_data` would be OK

nilearn/glm/first_level/first_level.py
```diff
@@ -490,6 +490,312 @@ def scaling_axis(self):
...
+            self.drift_order,
+            self.fir_delays,
+            confounds_matrix,
+            confounds_names,
+            self.min_onset,
+        )
+
+        return design
+
+    def fit_arrays(
```
bthirion, line 700: I think  I'd make it a private method.

bthirion (commented):
I'm still wondering whether we're creating more problems tha, those we are solving. People might really get confused with this, do whatever crazy things and then complain that the output is not great. Bzw you need to also implement array-level contrasts...

Remi-Gau:
> I'm still wondering whether we're creating more problems than those we are solving.

Suggestion: let me extract all the refactoring done here in a separate PR because it still improves code modularity and readability.

We can leave this PR open a bit longer to give us time to think about it.

Remi-Gau:
blocked by #4461

Remi-Gau:
OK the refactoring made the PR much easier to read / integrate.

Will leave it as draft. Maybe we can discuss this as our next dev meeting.

bthirion:
I'm hesitating between 2 ways forward:
* If the functions are meant to be sued on arrays, then we should name them differently from those that work on images to avoid confusions.
* Instead, we could provide some information about the spatial configuration (e.g. a masker)  alongside the data arrays; providing  spatial context together withe the data would be an odd way to describe neuroimaging data. But we would then have a unique API.

Remi-Gau:
FYI I am ok with naming it differently and keeping that function private as you suggested in a previous comment.

Remi-Gau:
closing 
see #4110

PR #4115, opened 2023-11-22
[DOC] Clarify docstrings of plotting functions in `img_plotting`

### LINKED ISSUE #911: Inconsistent and undocument return types for plot_stat_map (and probably other such API)
- Stuff returned by  `plot_stat_map` is no documented (it seems a Slicer object is returned by default)
- if output_file is passed, then a None object is returned instead

It would be nice to have the return type documented, and also we should always return the same object type (e.g a Slicer).

### LINKED ISSUE #2062: Add a See Also section to plotting.plot_img
**Original Title:** plotting.plot_img should be a private function

`plotting.plot_img` is not supposed to be directly called by users and should be a private function.

Master, 0.5.2

What were you trying to do?
Used `plotting.plot_img` like I do `plotting.view_img`

What did you expect will happen?
That instead of an interactive view, I'll get a same looking static image.

What actually happened?
The image had no background.

ymzayek:
- Closes #2062, closes #911

- Document returns of functions in img_plotting
- Clarify plot_img is low level function

### DIFF
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -175,6 +175,13 @@ def _plot_img_with_bg(img, bg_img=None, cut_coords=None,
     display_factory : function, default=get_slicer
         Takes a display_mode argument and return a display class.
 
+    Returns
+    -------
+    display : :class:`~plotting.displays.OrthoSlicer` or \
+        :class:`~plotting.displays.OrthoProjector` or None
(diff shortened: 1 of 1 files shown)


bthirion (commented):
LGTM so far ! 
Do you want to add other changes ?

Remi-Gau (approved):
Good with me

bthirion (approved):

PR #4116, opened 2023-11-23
[FIX] Fix PSC normalization when a high-pass filter is applied

### LINKED ISSUE #4087: [BUG] NiftiSpheresMasker - Combination of percent signal change normalization and high-pass filtering only outputs signals in incorrect range.

### Operating system

- [X] Linux
- [X] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 18.04
- Mac OS Ventura 13.6

### Python version

- [ ] 3.12
- [ ] 3.11
- [ ] 3.10
- [X] 3.9
- [ ] 3.8

### nilearn version

0.10.2

### Expected behavior

Percent signal change (PSC) normalization method in NiftiSpheresMasker in combination with high-pass filtering should return reasonable PSC values.

When extracting time courses with NiftiSpheresMasker, the combination of percent signal change normalization with high-pass filtering only outputs signals in an incorrect range.

I am using this example code https://nilearn.github.io/stable/auto_examples/03_connectivity/plot_sphere_based_connectome.html to showcase this.

This is the output image of the code snippet below. Please note the y-axis values range.
image

```python
from nilearn import datasets

dmn_coords = [(0, -52, 18), (-46, -68, 32)]
labels = [
    "Posterior Cingulate Cortex",
    "Left Temporoparietal junction",
]

from nilearn.maskers import NiftiSpheresMasker

masker = NiftiSpheresMasker(
    dmn_coords,
    radius=8,
    #detrend=True,
    standardize="psc",
    #standardize_confounds="zscore_sample",
    #low_pass=0.1,
    high_pass=0.01,
    t_r=2,
    memory="nilearn_cache",
    memory_level=1,
    verbose=2,
    clean__butterworth__padtype="even",  # kwarg to modify Butterworth filter
)

# Additionally, we pass confound information to ensure our extracted
# signal is cleaned from confounds.

func_filename = dataset.func[0]
confounds_filename = dataset.confounds[0]

time_series = masker.fit_transform(
    func_filename, confounds=[confounds_filename]
)

for time_serie, label in zip(time_series.T, labels):
    plt.plot(time_serie, label=label)

plt.title("Default Mode Network Time Series")
plt.xlabel("Scan number")
plt.ylabel("Normalized signal")
plt.legend()
plt.tight_layout()
```

ymzayek:
- Closes #4087


ymzayek:
Is it valid to just add back the original mean to the signal for calculating PSC?

ymzayek:
> Is it valid to just add back the original mean to the signal for calculating PSC?

Or alternatively pass the original mean signal and subtract that from the filtered signal like was suggested in the issue? But that doesn't center around 0

htwangtw:
I will need to do some experiment around this to see what's the most sensible way to move forward.
Indeed the original implementation of `psc` doesn't have enough safe guards. Even without filtering, user can generate things that are unusable if their inputs have mean signals close to 0.

htwangtw (commented):
The default of detrend is `True` so the current changes are 
1. not really correctly tested in `test_clean_psc`
2. doesn't address the bug of the original issue

I tried to add a test with butterworth filter and we ended up with very wonky means (number around -100 or 100). 

I have also discovered the psc output will have really wild numbers if we have means less than 1. This standardization method might need a new review @bthirion

bthirion:
The change looks good, bit we should probably test all relevant configurations. Best,

htwangtw:
It took me a long time to wrap my head around before and after holiday but I have found a satisfying solution. The correct tests broke lots of things but luckily the solution turned out to be quite simple after I figured thins out.

For tests:
- make sure the signal is not detrended in test so psc is tested in
  isolation
- add butterworth filter as a special case to the test

For code:
- streamline the logic of standardisation so we make sure mean signals are added back only when using psc and when butterworth filter or detrend was done.
- remove the mean_signals parameter from standardize_signal but calculate it inside the function. This is due to signal butterworth filter will have a non-zero mean that is really close to zero.
- make sure `filter_type` returns the correct filter that's applied. Before, we relied on the check in the butterworth function to apply things correctly. I have triple-checked and the filters were applied correctly despite `filter_type` was not correctly reflecting the strategy applied in case of butterworth filter... phew

@ymzayek Have a look and see if this is good to go

bthirion (approved):
Thx for digging into this I did not spot any issue.

ymzayek (commented):
@htwangtw thanks a lot! The changes look good and I checked the plots and they are as expected! I just added a whatsnew and will resolve the conflicts, then merge when everything is green

Issue #4027, opened 2023-10-04
[ENH] Next steps for development on experimental surface module


ymzayek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed enhancement in detail.

Following the merge of #3856, here's a non exhaustive list of follow-up tasks for the surface API:

Maskers:
- From (https://github.com/nilearn/nilearn/pull/3856#issuecomment-1723692075):
  - improve a bit the masker for the very frequent use case where we don't want to mask anything by avoiding to store the mask img 
  - Implement a `SurfaceVectorizer` that just checks shapes and concatenates data but doesn't have a mask_img and doesn't mask 
- Implement extraction of signals from ROI and transforming region signals back to surface object

Decoder: 
- Update `_check_embedded_nifti_masker` to work with surface masker

Plotting:
- Adapting surf plotting functions for `SurfaceImage`

Documentation:
- Add information about surface API to user guide
- Add surface terms to glossary

### Benefits to the change

Continue development of new module

### Pseudocode for the new behavior, if applicable

```python
# insert your code below

```

ymzayek:
List of (old) issues and PRs to address or to close once integration is done and stable:
- #2682 
- #2424 
- #3265 
- #3846

Remi-Gau:
Copying some of the notes from our hackathon regarding writing surface data:

Writing surface images

- Separate meshes and textures into different gifti files
- Question about naming: should try and make reasonable defaults for that, yet leave the users the possibility to change names
  - Suggestion to go with with something BIDS like where the hemisphere:
    - https://bids-specification.readthedocs.io/en/latest/derivatives/introduction.html#gifti-surface-data-format 
  - [hemi-{L|R}][_space-<space>][_den-<label>]_suffix.surf.gii
    - Wonder how much of those we can infer without having to keep track of the name of the input file…
    - Suffix is going to be almost impossible to have a sensible default for
    - Hemi specifies the hemisphere of a given file
    - Space for example “fsaverage”
    - Den for density, for example:
      - Den-10k for "10242 vertices per hemisphere (5th order icosahedron)"

ymzayek:
@Remi-Gau we need to note that #4120 needs to be reverted if we release without the surface API.

In addition I'm thinking whether it would be better to develop the integration on a feature branch?

Remi-Gau:
> @Remi-Gau we need to note that #4120 needs to be reverted if we release without the surface API.

Good point

[...]

PR #4117, opened 2023-11-24
[EHN] Implement signal.clean to SurfaceMasker

htwangtw:
- Closes #

This is an initial attempt to implement signal.clean to SurfaceMasker.
Concerns (generalisable to niftimaskers):
 - clean_kwargs - should filter out confounds and sample mask to prevent
user adding those parameters upon creating the masker
 - all the cleaning related parameters - what do we want to choose to
expose?

If we set on the solution I will expand this to `SurfaceLabelsMasker`

### DIFF
--- a/nilearn/experimental/surface/_maskers.py
+++ b/nilearn/experimental/surface/_maskers.py
@@ -4,7 +4,13 @@ from __future__ import annotations
 from typing import Any
 
 import numpy as np
+import pandas as pd
+from joblib import Memory
+from sklearn.base import BaseEstimator, TransformerMixin
 
+from nilearn import signal
+from nilearn._utils.cache_mixin import CacheMixin, cache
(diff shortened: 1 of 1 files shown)


bthirion (commented):
Looks good so far.

bthirion:
You probably want to add tests ?

htwangtw:
> You probably want to add tests ?
When I look at the `NiftiMasker` objects, it seems that they deliberately did not test out all the signal cleaning part and just assumed all the parameters are passed to `signal.clean` correctly. 
I am not sure what's the best thing to do.

ymzayek (approved):
LGTM!

bthirion (approved):
Thx !

htwangtw:
merging...

ymzayek:
linking to #4027 to keep track

PR #4120, opened 2023-11-24
[ENH] Let _check_embedded_nifti_masker work with surface masker

ymzayek:
- Part of https://github.com/nilearn/nilearn/issues/4027

### DIFF
--- a/nilearn/_utils/__init__.py
+++ b/nilearn/_utils/__init__.py
@@ -62,11 +62,7 @@ def all_modules(modules_to_ignore=None, modules_to_consider=None):
             "cannot be both specified."
         )
     if modules_to_ignore is None:
-        modules_to_ignore = {
-            "data",
-            "tests",
-            "externals",
-        }
+        modules_to_ignore = {"data", "tests", "externals", "conftest"}
     all_modules = []
     root = str(Path(__file__).parent.parent)
     with warnings.catch_warnings():
--- a/nilearn/_utils/tests/test_common.py
+++ b/nilearn/_utils/tests/test_common.py
@@ -34,7 +34,7 @@ def test_number_public_functions():
     If this is intentional, then the number should be updated in the test.
     Otherwise it means that the public API of nilearn has changed by mistake.
     """
-    assert len({_[0] for _ in all_functions()}) == 227
+    assert len({_[0] for _ in all_functions()}) == 205
 
 
 def test_number_public_classes():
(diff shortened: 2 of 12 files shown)


nilearn/_utils/__init__.py
```diff
@@ -62,11 +62,7 @@ def all_modules(modules_to_ignore=None, modules_to_consider=None):
             "cannot be both specified."
         )
     if modules_to_ignore is None:
-        modules_to_ignore = {
-            "data",
-            "tests",
-            "externals",
-        }
+        modules_to_ignore = {"data", "tests", "externals", "conftest"}
```
Remi-Gau, line 65: functions in conftest.py should be ignored by default

nilearn/_utils/tests/test_common.py
```diff
@@ -34,7 +34,7 @@ def test_number_public_functions():
     If this is intentional, then the number should be updated in the test.
     Otherwise it means that the public API of nilearn has changed by mistake.
     """
-    assert len({_[0] for _ in all_functions()}) == 227
+    assert len({_[0] for _ in all_functions()}) == 205
```
Remi-Gau, line 37: consequence of ignoring functions in conftest.py

nilearn/conftest.py
```diff
@@ -12,6 +12,7 @@
 # we need to import these fixtures even if not used in this module
 from nilearn.datasets.tests._testing import request_mocker  # noqa: F401
 from nilearn.datasets.tests._testing import temp_nilearn_data_dir  # noqa: F401
+from nilearn.experimental.surface.tests.conftest import *  # noqa: F401, F403
```
Remi-Gau, line 15: I would not be opposed to move the fixture in nilearn/conftest.py rather than doing an "import all".
Any reasons why we should not do it?

Remi-Gau: Not a blocker for a merge.

ymzayek: Well we wanted to isolate the new surface module and related code while it's still being developed. But I don't see a way to do that for this PR. I don't like this solution either I just didn't want to rewrite the fixtures and this is temporary. I can instead just import the needed fixtures

ymzayek: I made the imports explicit and added a note.

bthirion (approved):
LGTM.

ymzayek:
> I agree with that. ATM the masker_type could be "multi_nii", 'surface', 'nifti', but this is fare from clear.

@bthirion it is stated in the docstring of this function and it is a private function in any case. What is supported in terms of the type of makser is also relayed to the functions or classes that use this function.

bthirion:
Can be merged I think.

Issue #3467, opened 2023-01-16
binarize_img seems incorrect used on images with negative values


smeisler:
Hi Nilearn devs,
 I am working with kind of a strange ROI mask from a public dataset, where the background = -1 (black), cortical surface = 0 (dark gray), and ROIs of interest have values >= 1 (light gray and white). See the picture below.
<img width="462" alt="image" src="https://user-images.githubusercontent.com/27028726/212767285-f0c4a0d0-f7f4-4cb0-b28d-a5cdc665d39e.png">
Sure the numbering conventions are a bit odd, but I am just trying to extract the ROI masks, so thresholding at 0 with `binarize_img` should do the trick in theory.

Nilearn version: 0.10.0

### Expected behavior
Should result in an image with the ROI masks at value 1 and 0 everywhere else

### Actual behavior
The binarized image has 0s along the cortical surface (as expected), but 1s **everywhere** else (**background** and ROIs of interest). It appears that perhaps `binarize_img` may be using the magnitude of the values without considering the sign. Otherwise, why would the background (originally valued at -1) now survive 0-thresholding?
<img width="462" alt="image" src="https://user-images.githubusercontent.com/27028726/212768050-7257a425-3576-4e9c-9256-73b0aa879c7d.png">

### Steps and code to reproduce bug
```
roi = '/PATH/TO/lh.floc-bodies.nii.gz'
roi_binary = binarize_img(roi,threshold=0)
roi_binary.to_filename('ROI_check.nii.gz')
```
the ROI can be found here: https://drive.google.com/file/d/1EDNvsLolU496ur5ILe2hXv1CK80BjXs6/view?usp=sharing

Thank you,
Steven

smeisler:
I see what is happening, looking at the [`binarize_img` code](https://github.com/nilearn/nilearn/blob/37f7b0b0/nilearn/image/image.py#L1044).

```
return math_img(
        "img.astype(bool).astype(int)",
        img=threshold_img(img, threshold, mask_img=mask_img)
    )
```
`bool(-1)` is TRUE, so the -1 values are treated the same as positive. Perhaps there's a more robust method than booleans that can be used here.

bthirion:
It means that you want to threshold before binarizing. I'm wondering whether the best thing to do is not to document the behavior through an example (?).

smeisler:
I was able to get around this by just using: `math_img("roi>0", roi=roi)` (where 0 is the threshold)

bthirion:
[...]

PR #4121, opened 2023-11-25
[ENH] Update binarize_img for negative value handling

smeisler:
Addresses #3467.

 This change introduces the `two_sided` argument for `binarize_img` which allows users to control whether the `threshold` is applied to the original value of the image (`two_sided=False`) or absolute value of the image (`two_sided=True`). The behavior does not change from what is was (`two_sided` was always `True`), but I think it should change to default to `False` as it is more intuitive when thresholding images with negative values.

I am also interested in expanding the function to include multiple operators (less than, more than, etc), but figured this is a good place to start that addresses the linked issue.

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -1117,7 +1117,7 @@ def binarize_img(img, threshold=0, two_sided=True, mask_img=None):
     warnings.warn(
         'The current default behavior for the "two_sided argument "'
         'is  "True". This behavior will be changed to "False" in '
-        'version 0.3."',
+        'version 0.13."',
         DeprecationWarning,
         stacklevel=3,
     )
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -1117,7 +1117,7 @@ def binarize_img(img, threshold=0, two_sided=True, mask_img=None):
     warnings.warn(
         'The current default behavior for the "two_sided" argument '
         'is  "True". This behavior will be changed to "False" in '
-        'version 0.13."',
+        'version 0.13.',
         DeprecationWarning,
         stacklevel=3,
     )
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -1114,6 +1114,8 @@ def binarize_img(img, threshold=0, two_sided=True, mask_img=None):
      >>> img = binarize_img(anatomical_image)
 
     """
+
+
 warnings.warn(
     'The current default behavior for the "two_sided argument "'
     'is  "True". This behavior will be changed to "False" in '
@@ -1121,12 +1123,10 @@ warnings.warn(
     DeprecationWarning,
     stacklevel=3,
 )
-        
+
 return math_img(
     "img.astype(bool).astype(int)",
-    img=threshold_img(
-        img, threshold, mask_img=mask_img, two_sided=two_sided
-    ),
+    img=threshold_img(img, threshold, mask_img=mask_img, two_sided=two_sided),
 )
 
 
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -1115,7 +1115,7 @@ def binarize_img(img, threshold=0, two_sided=True, mask_img=None):
 
     """
     warnings.warn(
-        'The current default behavior for the "two_sided argument "'
+        'The current default behavior for the "two_sided" argument '
         'is  "True". This behavior will be changed to "False" in '
         'version 0.13."',
         DeprecationWarning,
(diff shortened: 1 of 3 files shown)


nilearn/image/image.py
```diff
@@ -1110,9 +1114,19 @@ def binarize_img(img, threshold=0, mask_img=None):
      >>> img = binarize_img(anatomical_image)
 
     """
+    warnings.warn(
+        'The current default behavior for the "two_sided argument "'
+        'is  "True". This behavior will be changed to "False" in '
+        'version 0.3."',
```
bthirion, line 1120:
```suggestion
        'version 0.13."',
```

nilearn/image/tests/test_image.py
```diff
@@ -852,12 +853,29 @@ def test_binarize_img(img_4d_rand_eye):
...
+        [True, False], size=img_4d_rand_eye.shape, p=[0.5, 0.5]
+    )
+
+    img4_data[neg_mask] *= -1
+    img4 = new_img_like(img_4d_rand_eye, img4_data)
+    # Binarize using original and absolute values
+    img4_original = binarize_img(img4, threshold=0, two_sided=False)
+    img4_absolute = binarize_img(img4, threshold=0, two_sided=True)
+    # Check that all values are 1 for absolute valued threshold
+    assert_array_equal(np.unique(img4_absolute.dataobj), np.array([1]))
```
bthirion, line 875:
```suggestion
    assert_array_equal(np.unique(img4_absolute.get_fdata()), np.array([1]))
``` (just for the sake of homogeneity)

**Review by bthirion — COMMENTED** [review]
Thx ! LGTM overall.

File: nilearn/image/tests/test_image.py (hunk 2 above)
Comment (smeisler) [reply nilearn/image/tests/test_image.py:874]: Just made this change in a separate commit (had to fix a typo too)

**Review by bthirion — APPROVED** [review]
LGTM, thx !

File: nilearn/image/image.py [hunk 3]
```diff
@@ -1110,9 +1114,19 @@ def binarize_img(img, threshold=0, mask_img=None):
      >>> img = binarize_img(anatomical_image)
 
     """
+    warnings.warn(
+        'The current default behavior for the "two_sided argument "'
```
ymzayek, line 1118:
```suggestion
        'The current default behavior for the "two_sided" argument '
```

nilearn/image/image.py
```diff
@@ -1110,9 +1114,19 @@ def binarize_img(img, threshold=0, mask_img=None):
...
+        'is  "True". This behavior will be changed to "False" in '
+        'version 0.13."',
```
ymzayek, line 1120:
```suggestion
        'version 0.13.',
```

nilearn/image/image.py
```diff
@@ -1063,7 +1063,7 @@ def math_img(formula, **imgs):
     return new_img_like(niimg, result, niimg.affine)
 
 
-def binarize_img(img, threshold=0, mask_img=None):
+def binarize_img(img, threshold=0, two_sided=True, mask_img=None):
```
ymzayek, line 1066: Any reason not to add this new argument to the end? This will break code of people using positional arguments.

nilearn/image/tests/test_image.py
```diff
@@ -852,12 +853,28 @@ def test_binarize_img(img_4d_rand_eye):
...
+    )
+    img4_data[neg_mask] *= -1
+    img4 = new_img_like(img_4d_rand_eye, img4_data)
+    # Binarize using original and absolute values
+    img4_original = binarize_img(img4, threshold=0, two_sided=False)
+    img4_absolute = binarize_img(img4, threshold=0, two_sided=True)
+    # Check that all values are 1 for absolute valued threshold
+    assert_array_equal(np.unique(img4_absolute.dataobj), np.array([1]))
+    # Check that binarized image contains 0 and 1 for original threshold
+    assert_array_equal(np.unique(img4_original.dataobj), np.array([0, 1]))
```
ymzayek, line 876: Can you extract into a separate test?

ymzayek (commented):
Thanks @smeisler ! Minor things

nilearn/image/tests/test_image.py
smeisler: Done!

nilearn/image/image.py
smeisler: Moved it in a new commit

nilearn/image/image.py
```diff
@@ -1084,6 +1084,10 @@ def binarize_img(img, threshold=0, mask_img=None):
         this score will be kept. The given string should be
         within the range of "0%" to "100%".
 
+    two_sided : :obj:`bool`
```
ymzayek, line 1087: This should also move below

ymzayek (commented):
Thanks for the quick changes. LGTM. Last thing then I think it should be good to go

nilearn/image/image.py
smeisler: done!

ymzayek (approved):
Awesome thx! Once the CI finishes and is green, will merge

PR #4123, opened 2023-11-27
[MAINT] Add pytest-mpl unit tests to check ouput changes in plots

ymzayek:
- Closes None

This was referenced in an issue or PR but I cannot find where...

This adds / moves specific tests to make sure plotting function outputs don't change accidentally.

The checks only happens with oldest supported python + matplotlib to avoid reproducibility issues.

## TODO

- [x] update maintenance to do explain what to do when bumping python / matplotlib


Remi-Gau:
Probably worth updating this one: a lot of out plotting test are either smoke tests or 'smoke like' tests as in we may check some 'features' of the figure object or if a figure was created / saved but we never check what the figure looks like.

Given how much people use nilearn as a visualization library, it'd be nice to ensure we have no regressions in behavior on at least some of our main plotting functions.

tox.ini
```diff
@@ -13,6 +13,12 @@ skip_install = true
 deps =
     pre-commit
 
+[min_python]
+description = minimum python version
+skip_install = false
+base_python = 3.9
```
Remi-Gau, line 19: refactor to declare only once what is the mimimum supported python version

Remi-Gau:
@hndgzkn @man-shu
Can you have an early look at this to make sure you get the logic of testing plots this way?

Remi-Gau:
Note that we may want to adopt a "source layout" for those tests and / or at least store the baseline image outside of the package to avoid making the package heavier to install.

nilearn/plotting/tests/test_img_plotting/test_plot_glass_brain.py
```diff
@@ -11,35 +11,13 @@
...
 @pytest.mark.parametrize("plot_abs", [True, False])
-def test_plot_glass_brain_background_absolute(
-    matplotlib_pyplot, img_3d_mni, plot_abs, black_bg
+@pytest.mark.parametrize("resampling_interpolation", ["nearest", "continuous"])
+def test_plot_glass_brain_absolute(
+    matplotlib_pyplot, img_3d_mni, plot_abs, resampling_interpolation
 ):
-    """Somke test for absolute value plotting and background."""
-    plot_glass_brain(img_3d_mni, black_bg=black_bg, plot_abs=plot_abs)
+    """Somke test resampling_interpolation and for absolute value plotting."""
```
bthirion, line 19:
```suggestion
    """Smoke test resampling_interpolation and for absolute value plotting."""
```

nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -0,0 +1,192 @@
+"""
+Test if public plotting functions' output has changed.
+
+Sometimes, the output of a plotting function may unintentionanly change
```
bthirion, line 4:
```suggestion
Sometimes, the output of a plotting function may unintentionally change
```

bthirion (commented):
Just wondering: how much cost is it in terms of  
* repo size (addition of pngs)
* computation time

Remi-Gau:
> Just wondering: how much cost is it in terms of
> 
>     * repo size (addition of pngs)

So far that's how much we have, and this will probably grow.

```bash
du nilearn/plotting/tests/baseline -h
8,0M    nilearn/plotting/tests/baseline
```

As I commented above, we probably want to exclude this from the wheel that people will install.
Either by just excluding the file (I think this can be set in the pyproject.toml) or by moving those tests and their png outside of the nilearn folder (see [source layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/)).

>     * computation time

Need to measure, but keep in mind that those are in most cases not NEW tests. The best way to think about it is that a lot what were 'smoke' tests that could only tell if the code runs without crashing, will now actually have an 'assert' that checks the output.

Remi-Gau:
wheel size

on this PR

```
-rw-r--r--  1 remi remi  20M févr. 19 07:23 nilearn-0.11.2.dev173+g50ba38088.tar.gz
-rw-r--r--  1 remi remi  18M févr. 19 07:23 nilearn-0.11.2.dev173+g50ba38088-py3-none-any.whl
```

on main

```
-rw-r--r--  1 remi remi  13M févr. 19 07:24 nilearn-0.11.2.dev152+gf8a5d0a9d.tar.gz
-rw-r--r--  1 remi remi  11M févr. 19 07:25 nilearn-0.11.2.dev152+gf8a5d0a9d-py3-none-any.whl
```

So we clearly want those out of the wheel.

Remi-Gau:
> * computation time

When checking output (with the `@pytest.mark.mpl_image_compare` decorator on the tests)

- 220 passed, 17 warnings in 75.80s

when just running them as smoke tests

- 220 passed, 225 warnings in 36.69s

So roughly doubles run time for the tests concerned.

Note I only ran it once for each.

Remi-Gau:
> * computation time

Another thing.

In CI only a single job will be affected by this: the one that runs with our oldest supported python (3.9) / matplotlib (3.3.0) (`oldest dependencies - no plotly`): see the `plot_min` in tox.ini.

In the last valid run of this job in this PR it tool around 16 minutes: https://github.com/nilearn/nilearn/actions/runs/13393408513/job/37406339116#logs

For comparison I am adding the run time of the same job in the past month  (and that for all the tests on ubuntu with python 3.13).

newplot

So it does make this job slower, but it is still faster than most other test jobs.

All the other jobs should not be affected.

nilearn/plotting/tests/test_img_plotting/test_img_plotting.py
```diff
@@ -93,20 +85,6 @@ def test_plot_functions_invalid_threshold(plot_func, img_3d_mni, tmp_path):
...
-    plot_func(
-        img_3d_mni,
-        output_file=filename,
-        colorbar=True,
-        cbar_tick_format=cbar_tick_format,
-    )
-    plt.close()
-
-
 def test_plot_carpet_default_params(img_4d_mni, img_3d_ones_mni, tmp_path):
```
man-shu, line 88: I am just wondering why we still need some of the "smoke tests" without the image compare decorator here. Maybe you have yet to remove them?

For example, this one seems like a copy of the one in L136 of `nilearn/plotting/tests/test_baseline_comparisons.py`.

Remi-Gau: I have not been completely thorough in cleaning / moving tests so expect some duplicate.

Just want to make sure we want this before 'making ~perfect~ better'

man-shu: Ok makes sense!

Remi-Gau: Also if we are ok using this for testing, we will have follow up PR to apply it to matrix plotting and other plotting functions...

man-shu: My first thought was that probably just examining/comparing Figure objects could be possible (without saving pngs).

I looked into how matplotlib does/suggests to do the testing, and they seem to have two methods:
- One that compares the "baselines" with PNGs: https://matplotlib.org/devdocs/devel/testing.html#writing-an-image-comparison-test
- But I think the second one seems to just compare Figure objects: https://matplotlib.org/devdocs/devel/testing.html#compare-two-methods-of-creating-an-image

Remi-Gau: But that does not allow to have regression tests to check if a PR is modifying the behavior of a plotting function, no?

man-shu: Yeah looks like you can just give two Figure instances to compare if they are the same or not.

nilearn/plotting/tests/test_baseline_comparisons.py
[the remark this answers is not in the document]
bthirion: Yes, but... I'm afraid of increasing complexity. We would like to keep it simple so that future maintainers do not have too much complexity in the file organization: limit nesting etc.

bthirion (commented):
When is this supposed to be run ?

.github/workflows/testing_install.yml
```diff
@@ -46,6 +46,7 @@ jobs:
             run: |
                 set -e
                 python -m build
+                ls -lrth dist
```
Remi-Gau, line 49: to help view the size of the wheel

Remi-Gau:
see updated maintenance page: # Force resampling on first image

doc/maintenance.rst
```diff
@@ -173,6 +173,34 @@ You can also run any arbitrary command in a given environment with ``tox exec``:
...
+==================================================
+
+We use the ``pytest-mpl`` pytest plugin to run several regression tests on our Matplotlib figures.
+
+Sometimes, the output of a plotting function may unintentionally change
+as a side effect of changing another function or piece of code
+that it depends on.
+These tests ensure that the outputs are not accidentally changed.
+
+Tor those tests and for each figure to test,
```
man-shu, line 186: Maybe this is what you meant?
```suggestion
For each figure to test,
```

nilearn/plotting/tests/baseline/README.md
```diff
@@ -0,0 +1,4 @@
+# Baseline figure for plotting tests
+
+See the maintenance page of our documentation for more information
+https://nilearn.github.io/dev/maintenance.html
```
man-shu, line 4: I think it's better to directly link the subsection?
```suggestion
https://nilearn.github.io/dev/maintenance.html#generating-new-baseline-figures-for-plotting-tests
```

nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -0,0 +1,175 @@
+"""
+Test if public plotting functions' output has changed.
+
+See the  maintenance page of our documentation for more information
+https://nilearn.github.io/dev/maintenance.html
```
man-shu, line 5: here too
```suggestion
https://nilearn.github.io/dev/maintenance.html#generating-new-baseline-figures-for-plotting-tests
```

man-shu (approved):
Otherwise LGTM!

bthirion (approved):
LGTM.

Issue #3846, opened 2023-07-19
[ENH] Surface GLM - Support multiple runs in a GLM


smeisler:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed enhancement in detail.

Related to neurostars post: https://neurostars.org/t/glm-on-surface-data-in-nilearn/22632

It would be good if Nilearn could run GLMs on surface data with multiple BOLD runs. Currently, the only way to run a GLM on surface data (`nilearn.glm.first_level.run_glm`), which only accepts one BOLD run and design matrix per instance.

### Benefits to the change

More flexible processing for multi-run surface analyses.

### Pseudocode for the new behavior, if applicable

I imagine an easiest, but not most stylish, change would be to have `run_glm` accept an iterable object for `[X]` and `[Y]`. If the length of these iterable objects are 1, then run the function as normal. If not, apply whatever fixed effect models are used for the multi-run GLMs on BOLD data.

Thoughts?

bthirion:
Thx for opening  this. What we need is a SurfaceFirstLevelModel object that can has an API similar to the FirstLevelModel. We'll do that as soon as we have a decent SurfaceMasker.

Remi-Gau:
Will close this one as completed by #4126.

See examples:
- https://github.com/nilearn/nilearn/blob/main/examples/08_experimental/plot_localizer_new_surface_analysis_experimental.py
- https://github.com/nilearn/nilearn/blob/main/examples/08_experimental/plot_new_surface_bids_analysis_experimental.py
 
Feedbacks are welcome.

Issue #3265, opened 2022-05-20
FirstLevelModel for volume on surface data


jyeatman:
Currently the GLM pipeline diverges substantially for surface vs volume data. It would be much more straightforward if the same functions operated on surface or volume data. It seems like the desirable functionality would be for FirtLevelModel to operate interchangeably on volume or surface data. I haven't gone under the hood but from the function description it seems like this would be a straightforward update

### Benefits to the change
Fewer errors if the same functions handle volume or surface

### Pseudocode for the new behavior, if applicable

```python

```

bthirion:
Thx @jyeatman 
The main alternative to that is to develop a SurfaceFirstLevelModel that would be dedicated to surface-based analysis.
But it is highly possible that the two objects would share much of the functions, since volume/surface distinctions only make sense when it comes to input/output operations.
Any thought on this ?

jyeatman:
My personal preference/bias is towards one function that flexibility handles different data types. Looking at the routines with FirstLevelModel it seems like very few (beyond reading the data) are specific to volume. So it seams that it would be more straightforward and less error prone for users to be able to, for example, use all the same code and try their analysis in the volume or the surface

jeromedockes:
this is related to discussions in nilearn and nibabel on the representation of surface images; ideally with the right representation for surface data FirstLevelModel will be applicable directly to surface data just by using an appropriate masker

see https://nipy.org/nibabel/devel/biaps/biap_0009.html for a draft on the nibabel side

Remi-Gau:
Will close this one as completed by #4126.

Remi-Gau:
See examples:

    https://github.com/nilearn/nilearn/blob/main/examples/08_experimental/plot_localizer_new_surface_analysis_experimental.py
    https://github.com/nilearn/nilearn/blob/main/examples/08_experimental/plot_new_surface_bids_analysis_experimental.py

Feedbacks are welcome.

PR #4126, opened 2023-11-30
[ENH] Adapt FirstLevelModel to work with surface data objects from new API

ymzayek:
- Part of https://github.com/nilearn/nilearn/issues/4027

TODO:

examples to adapt:

- [x] examples/04_glm_first_level/plot_localizer_surface_analysis.py
- [x] examples/07_advanced/plot_surface_bids_analysis.py
  - before: https://nilearn.github.io/dev/auto_examples/07_advanced/plot_surface_bids_analysis.html#group-study
  - after: https://output.circle-artifacts.com/output/job/628c6b17-5666-4927-9fbd-4a4e5d838c7e/artifacts/0/dev/auto_examples/08_experimental/plot_new_surface_bids_analysis.html#group-study


ymzayek:
@bthirion I started this to work on adapting the glm module. Still very WIP but we can see better what needs to be done. Let me know what you think

ymzayek:
Regarding [L628 `if self.mask_img.mask_img_ is None and self.masker_ is None:`](https://github.com/nilearn/nilearn/pull/4126/files#diff-5163e83595308445660793758eef38161ce350195720189a013a8989d8a3c6d3R628) and what follows it, it seems to me that it is never the case that `self.mask_img.mask_img_ is None` following the `else:` condition and checking that the masker has been fitted. It will only be met if the user explicitly sets `masker.mask_img_ = None` after fitting and this part of the code is tested exactly like that here: https://github.com/nilearn/nilearn/blob/0a08873aa7b7f7c37c1e4fe3febf333f3bce4344/nilearn/glm/tests/test_first_level.py#L62-L72 I find this a bit odd but maybe there is a reason?

Another issue I saw while examining this part of the code that may be worth a separate issue is that the error message when both mask_img and img are None when calling fit on a NiftMasker is `*** TypeError: Data given cannot be loaded because it is not compatible with nibabel format: None`. This is not clear and I think we should rather mimic what was implemented in SurfaceMasker:
https://github.com/nilearn/nilearn/blob/0a08873aa7b7f7c37c1e4fe3febf333f3bce4344/nilearn/experimental/surface/_maskers.py#L79-L82

nilearn/glm/first_level/first_level.py
```diff
@@ -576,11 +578,27 @@ def fit(
         # Learn the mask
         if self.mask_img is False:
             # We create a dummy mask to preserve functionality of api
-            ref_img = check_niimg(run_imgs[0])
-            self.mask_img = Nifti1Image(
-                np.ones(ref_img.shape[:3]), ref_img.affine
-            )
-        if not isinstance(self.mask_img, NiftiMasker):
+            if isinstance(run_imgs[0], SurfaceImage):
+                left, right = run_imgs[0].mesh.keys()
```
bthirion, line 582: there is no reason why the ordering would be left, right ?

nilearn/glm/tests/test_first_level.py
```diff
@@ -1860,3 +1861,86 @@ def test_missing_trial_type_column_warning(tmp_path_factory):
...
+        )
+        mini_img = make_mini_img((shape,))
+        return mini_img, des
+
+    return _make_surface_image
+
+
+def test_flm_fit_surface_image(_make_surface_glm_data):
+    """Test FirstLevelModel with surface image and mask_img set to False."""
+    mini_img, des = _make_surface_glm_data(5)
```
bthirion, line 1885: it takes two arguments, no ?

bthirion (commented):
Thx, this is super important !
Do you plan to add the example in a forthcoming PR ?

Remi-Gau:
Was going through our backlog of issues and wondering if this PR resolves:
- #3265
- #3846

Also wondering more generally if we feel OK closing as "solved" something that is not yet considered "stable" (depends on experimental features).

nilearn/glm/first_level/first_level.py
ymzayek: Yes bad naming. I'll change it. I think we also don't require both hemispheres so maybe a check for that as well

nilearn/glm/first_level/first_level.py
```diff
@@ -576,11 +577,27 @@ def fit(
...
+                self.mask_img = SurfaceImage(
+                    mesh=run_imgs[0].mesh,
+                    data={
+                        "left_hemisphere": np.ones(
+                            run_imgs[0].data[left].shape
+                        ),
+                        "right_hemisphere": np.ones(
+                            run_imgs[0].data[right].shape
```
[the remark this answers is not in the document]
ymzayek: idem

nilearn/glm/tests/test_first_level.py
ymzayek: It works a bit differently within the context of pytest which will automatically pass the 2 arguments (which are actually other fixtures) to the _make_surface_glm_data fixture which then calls the inner function that takes one argument of shape that I'm passing here

Remi-Gau:
> Do you plan to add the example in a forthcoming PR ?

Going through the PR. What I am seeing makes sense to me. Thanks a bunch @ymzayek 

Was trying to think of a follow up example for this.

We mentioned this example during the last hackathon: https://nilearn.github.io/stable/auto_examples/04_glm_first_level/plot_localizer_surface_analysis.html 

However this requires to project the fMRI image to the surface which is something we do not yet support with this new Surface object.

Do we have a dataset available with first level surface data?
If not, it may be possible to grab one of the fmriprep preprocessed dataset from openneuro that should have this type of data.

ymzayek:
@bthirion did you have a chance to test this yet?

bthirion:
Will do tomorrow ! Thx for the reminder.

bthirion:
> @bthirion did you have a chance to test this yet?

See #4185

ymzayek:
TODO:

- Improve handling of mask_img check for setting up masker object:
  - The conditional check is based on `mask_image == False` but default case it equals `None`

bthirion (commented):
LGTM otherwise.

nilearn/glm/first_level/first_level.py
```diff
@@ -576,11 +578,39 @@ def fit(
...
+                            part_1: np.ones(
+                                run_imgs[0].data[part_1].shape[1], dtype=bool
+                            ),
+                            part_2: np.ones(
+                                run_imgs[0].data[part_2].shape[1], dtype=bool
+                            ),
+                        },
+                    )
+                else:
+                    part = list(parts)[0]
```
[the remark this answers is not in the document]
ymzayek: Yes this needs to be changed. It seemed I worked from the assumption of having only 2 hemispheres stored but there is nothing preventing the user from storing more in our surface objects

bthirion:
Thx for moving this fw. What is the status with the GLM report ?

ymzayek:
> Thx for moving this fw. What is the status with the GLM report ?

Having a look today

ymzayek:
@bthirion in order to get the brain plotting in the reports to work we only have to modify `_mask_to_svg` and `_make_stat_maps_contrast_clusters` under `make_glm_report`. The former might be straightforward by calling `plot_surf_roi` instead of `plot_roi` when a SurfaceMasker is passed and then passing the needed parts from the surface image objects but this will probably have to be changed again when surface plotting is adapted to the new surface objects so I'm not sure if it should wait until that. The latter seems a bit more involved as more functions are called that only support nifti. So I think the options are

1) try to adapt it using the code as is (and later some parts need to be changed when brain plotting is adapted)
2) don't support it for surface for now and handle the exception with a custom error message
3) partially support it (no brain plots for now; raise a warning)

Let me know what you think

bthirion:
OK, it does not make sens to hack a solution while plotting function still don't work with the Surface objects. It's OK to through an NotImplementedError for the next release IMHO.

But... What's ging on with plotting functions ?

ymzayek:
> But... What's ging on with plotting functions ?

Nothing in particular, it's just that they don't accept a surface image object yet so it's a bit clunky passing parts of the surface data (like is done now in the example) and to my understanding this will change anyways. In addition, there is thresholding that only works on niimg in `_make_stat_maps_contrast_clusters` and perhaps other details that might need modification, which is ok but I'd rather do that work separately (in a separate PR building on this one). So I'd also vote for NotImplementedError for now

bthirion (approved):
I can't sport any issue. This is great !

ymzayek:
@bthirion I refactored it a bit by taking the mask section out of the `fit` into its own `_apply_mask` function that is called by `fit`. But basically it is exactly the same. Maybe this will help make #4112 more straightforward to implement

bthirion:
> @bthirion I refactored it a bit by taking the mask section out of the `fit` into its own ~`_apply_mask`~ `_prepare_mask` function that is called by `fit`. But basically it is exactly the same. Maybe this will help make #4112 more straightforward to implement

LGTM, thx.

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -64,51 +63,46 @@
...
 
-fsaverage = nilearn.datasets.fetch_surf_fsaverage()
 
 # %%
 # The projection function simply takes the :term:`fMRI` data and the mesh.
 # Note that those correspond spatially, are they are both in :term:`MNI` space.
 from nilearn import surface
-
-texture = surface.vol_to_surf(fmri_img, fsaverage.pial_right)
+from nilearn.experimental.surface import SurfaceImage, load_fsaverage
```
Remi-Gau, line 72: just to make sure I understand:
- if we decide that the surface experimental is "ready" for the next release, we take it out of the `nilearn.experimental` package and adapt the examples and code accordingly
- otherwise we move this example to `examples/08_experimental` and keep along side the old version of this example?

Remi-Gau (dismissed):
Looks good to me too. Thanks @ymzayek

examples/04_glm_first_level/plot_localizer_surface_analysis.py
bthirion: the ideal would be to move everything out of experimental now, but I'm afraid we're too short for the upcoming release ?

nilearn/glm/first_level/first_level.py
```diff
@@ -909,6 +873,90 @@ def _get_voxelwise_model_attribute(self, attribute, result_as_time_series):
...
+        Parameters
+        ----------
+        run_img : Niimg-like object or SurfaceImage object
+            Used for setting up the masker object.
+        """
+        # Local import to prevent circular imports
+        from nilearn.maskers import NiftiMasker
+
+        # Learn the mask
+        if self.mask_img is False:
```
bthirion, line 888: or None ?

bthirion (approved):
LGTM, we just need to handle the end of the experimental module.

ymzayek:
> the ideal would be to move everything out of experimental now, but I'm afraid we're too short for the upcoming release ?

@bthirion this is what I'm doing in #4229 so that needs to be merged before this one and then this needs to be rebased and the imports changed. This will be quite easy to do in this PR so the focus should be on finishing 4229. This goes also for the decoder PR (will need to be rebased on 4229 and imports fixed). I used the labels `blocker` and `blocked` on these related PRs to try to clarify this dependeny.

nilearn/glm/first_level/first_level.py
ymzayek: @bthirion `mask_img=None` is directly passed to initialization of either `SurfaceMasker` in L906 or `NiftiMasker` in L918. Note that in the case of `NiftiMasker` the behavior of `None` and `False` are different because `None` leads to automatic computation of a mask and `False` leads to no masking. Since `SurfaceMasker` doesn't implement automatic mask computation the behavior is the same.

examples/08_experimental/plot_localizer_surface_analysis.py
```diff
@@ -0,0 +1,239 @@
...
+# Here we use an :term:`HRF` model
+# containing the Glover model and its time derivative
+# The drift model is implicitly a cosine basis with a period cutoff at 128s.
+from nilearn.glm.first_level import FirstLevelModel
+
+glm = FirstLevelModel(
+    t_r,
+    slice_time_ref=slice_time_ref,
+    hrf_model="glover + derivative",
+).fit(image, events)
```
Remi-Gau, line 125: this change compare to the original example

examples/08_experimental/plot_localizer_surface_analysis.py
```diff
@@ -0,0 +1,239 @@
...
+image = SurfaceImage(
+    mesh={
+        "lh": fsaverage5["pial"]["left_hemisphere"],
+        "rh": fsaverage5["pial"]["right_hemisphere"],
+    },
+    data={
+        "lh": texture_left.T,
+        "rh": texture_right.T,
+    },
+)
```
Remi-Gau, line 106: this change compare to the original example

examples/08_experimental/plot_new_surface_bids_analysis.py
```diff
@@ -0,0 +1,180 @@
...
+import numpy as np
+from scipy.stats import norm, ttest_1samp
+
+_, pval_left = ttest_1samp(np.array(z_scores_left), 0)
+_, pval_right = ttest_1samp(np.array(z_scores_right), 0)
+
+# %%
+# What we have so far are p-values: we convert them to z-values for plotting.
+z_val_left = norm.isf(pval_left)
+z_val_right = norm.isf(pval_right)
```
Remi-Gau, line 150: Would it make sense to also have second level support for surface images?

Remi-Gau: @bthirion

examples/08_experimental/plot_new_surface_bids_analysis.py
```diff
@@ -0,0 +1,180 @@
...
+    image = SurfaceImage(
+        mesh={
+            "lh": fsaverage5["pial"]["left_hemisphere"],
+            "rh": fsaverage5["pial"]["right_hemisphere"],
+        },
+        data={
+            "lh": texture_left.T,
+            "rh": texture_right.T,
+        },
+    )
```
Remi-Gau, line 114: given that fmriprep datasets can contain surface data, it would make sense for `first_level_from_bids` to return the path to the surface images if the space label corresponds to some fsaverage.

Remi-Gau:
probably for a follow up PR

currently for fitting GLM:
- in volumes users can pass path to nifti files or nifti objects
- for surfaces users must pass a surface object

I think we should have "parity" for those 2 and users should be able to also just pass path to surfaces to fir their GLM.

examples/08_experimental/plot_new_surface_bids_analysis.py
Remi-Gau, line 1: figures before:

image

image


after:

image

image

Remi-Gau: There are small differences but not sure what they are due to.

examples/08_experimental/plot_new_surface_bids_analysis.py
```diff
@@ -0,0 +1,183 @@
...
+    image = SurfaceImage(
+        mesh={
+            "lh": fsaverage5["pial"]["left_hemisphere"],
+            "rh": fsaverage5["pial"]["right_hemisphere"],
+        },
+        data={
+            "lh": texture_left.T,
+            "rh": texture_right.T,
+        },
```
bthirion, line 115: This cries for a function that would do the job.

bthirion (commented):
LGTM overall !

examples/08_experimental/plot_localizer_new_surface_analysis_experimental.py
```diff
@@ -0,0 +1,239 @@
...
+        f"  Contrast {index + 1:1} out of {len(contrasts)}: "
+        f"{contrast_id}, right hemisphere"
+    )
+    # compute contrast-related statistics
+    z_score = glm.compute_contrast(contrast_val, stat_type="t")
+
+    # we plot it on the surface, on the inflated fsaverage mesh,
+    # together with a suitable background to give an impression
+    # of the cortex folding.
+    plotting.plot_surf_stat_map(
```
Remi-Gau, line 220: - further refactoring will be possible once https://github.com/nilearn/nilearn/pull/4431 is merged

examples/08_experimental/plot_localizer_new_surface_analysis_experimental.py
```diff
@@ -0,0 +1,239 @@
...
+        fsaverage.infl_right,
+        z_score.data.parts["right"],
+        hemi="right",
+        title=contrast_id,
+        colorbar=True,
+        threshold=3.0,
+        bg_map=fsaverage.sulc_right,
+    )
+    plotting.plot_surf_stat_map(
```
Remi-Gau, line 229: - further refactoring will be possible once https://github.com/nilearn/nilearn/pull/4431 is merged

examples/08_experimental/plot_new_surface_bids_analysis_experimental.py
```diff
@@ -0,0 +1,178 @@
...
+z_val_right = norm.isf(pval_right)
+
+# %%
+# Plot the resulting maps, at first on the left hemisphere.
+from nilearn import plotting
+from nilearn.datasets import fetch_surf_fsaverage
+
+fsaverage = fetch_surf_fsaverage(mesh="fsaverage5")
+
+plotting.plot_surf_stat_map(
```
Remi-Gau, line 157: - further refactoring will be possible once https://github.com/nilearn/nilearn/pull/4431 is merged

examples/08_experimental/plot_new_surface_bids_analysis_experimental.py
```diff
@@ -0,0 +1,178 @@
...
+    z_val_left,
+    hemi="left",
+    title="language-string, left hemisphere",
+    colorbar=True,
+    threshold=3.0,
+    bg_map=fsaverage.sulc_left,
+)
+# %%
+# Next, on the right hemisphere.
+plotting.plot_surf_stat_map(
```
Remi-Gau, line 168: - further refactoring will be possible once https://github.com/nilearn/nilearn/pull/4431 is merged

examples/08_experimental/plot_new_surface_bids_analysis_experimental.py
```diff
@@ -0,0 +1,178 @@
...
+    texture_right = surface.vol_to_surf(
+        fmri_img[0], fsaverage5["pial"].parts["right"]
+    )
+    image = SurfaceImage(
+        mesh=fsaverage5["pial"],
+        data={
+            "left": texture_left.T,
+            "right": texture_right.T,
+        },
+    )
```
Remi-Gau, line 113:
```suggestion
    texture_left = surface.vol_to_surf(
        fmri_img[0], fsaverage5["pial"].parts["left"]
    )
    texture_right = surface.vol_to_surf(
        fmri_img[0], fsaverage5["pial"].parts["right"]
    )
    image = SurfaceImage(
        mesh=fsaverage5["pial"],
        data={
            "left": texture_left.T,
            "right": texture_right.T,
        },
    )
```

to echo @bthirion 

I am already seeing this pattern a few times (that is: construct a SurfaceImage object from a mesh and a volume image), so I suspect we may have to think about allowing data to be a Nifti object and modify the constructor accordingly)

Remi-Gau: better to do in a later PR though

examples/08_experimental/plot_localizer_new_surface_analysis_experimental.py
Remi-Gau: should not prevent to merge before #4431 

just making a note for future PR

PR #4131, opened 2023-12-05
[ENH] Raise error when negative or non-integer ROI values are passed to  `plot_surf_roi`

### LINKED ISSUE #4104: plot_surf_roi should raise error when negative or floating-point values are passed
plot_surf_roi is for showing atlas region indices, so it is meant to deal with positive integers only.
we should probably raise an error when negative or floating-point values are passed to this function

_Originally posted by @jeromedockes in https://github.com/nilearn/nilearn/issues/4082#issuecomment-1787355353_

michellewang:
- Closes #4104

- Raise `ValueError` in `plot_surf_roi` if `roi_map` has negative or non-integer values
  - For the non-integer case, this is done by casting the `roi` array to integer, then comparing with the original `roi` array. Not sure if there's a better way to do this
- Add tests for both cases above
- Also test checking that another `ValueError` is raised if `roi` is not one-dimensional (there wasn't a test for that)

If we want a deprecation cycle (see https://github.com/nilearn/nilearn/issues/4104#issuecomment-1840323908) I can add that as well.


bthirion:
LGTM so far, but we probably need to add a deprecation cycle.

michellewang:
Ok I replaced the errors by deprecation warnings. Do we usually specify the future version of Nilearn where deprecated features will be removed? If so, which version would we want to name?

bthirion:
Thx. I guess that removal will be in 0.13.

nilearn/plotting/surf_plotting.py
```diff
@@ -1644,6 +1645,24 @@ def plot_surf_roi(surf_mesh,
                          'ROI you can convert them into a ROI map like this:\n'
                          'roi_map = np.zeros(n_vertices)\n'
                          'roi_map[roi_idx] = 1')
+    if (roi < 0).any():
```
ymzayek, line 1648: I think you can delete the comments raising the value error here and in the tests and then just add a note above this if statement with something like `# TODO: raise ValueError in release 0.13`

And then also change the warning message to mention the version

ymzayek (approved):
Thx @michellewang !

bthirion (approved):
LGTM, thx !

PR #4149, opened 2023-12-11
[INFRA] Remove CI workflow to error on DeprecationWarnings of dependencies

ymzayek:
- Closes None

- Remove monthly CI to error on external DeprecationWarnings

This job doesn't work very well. I think there might be a better way to manage this with tox. I suggest to remove for now. Also, we are pretty up-to-date on handling deprecations


bthirion (commented):
I have no personal opinion on this. What is the motivation for removing it ?

ymzayek:
At the moment it's failing on a deprecation warning that would need to be handled by an external library. It's purpose was to notify us about deprecation warnings that we actually need to handle ourselves before they even fail on new releases or even pre releases of our dependencies. In any case removing it is not risky for us and we still have the python 3.12 job with pre releases that will fail if we have deprecated code associated with a pre release of one of our dependencies and that's usually handled quickly enough

Remi-Gau:
Definitely low risk to remove. 

Chekcing the latest runs:
https://github.com/nilearn/nilearn/actions/workflows/check_deprecations.yml

Last one could not even start and the one before that did catch a few things though.

Would it be worth thinking of better ways to replace it (as part of this PR or another?)

ymzayek:
Yes let's do it in another PR I would say. I can reopen the original issue.

> Last one could not even start and the one before that did catch a few things though.

The one before seems to only have caught internal deprecation warnings (the _LEGACY_FORMAT_MSG) which is also not the purpose of this workflow

Remi-Gau:
Ok let's merge this and reopen the old issue then.

bthirion (approved):
Thx for the explanations.

Remi-Gau (approved):

PR #4151, opened 2023-12-12
[MAINT] Improve error for nifti masker when fit is called without imgs or mask_img

ymzayek:
- Related to https://github.com/nilearn/nilearn/pull/4126#issuecomment-1847137828

- Raise a more meaningful error than 
```bash
E TypeError: Data given cannot be loaded because it is not compatible with nibabel format:
E None
```

### DIFF
--- a/nilearn/maskers/nifti_masker.py
+++ b/nilearn/maskers/nifti_masker.py
@@ -428,6 +428,12 @@ class NiftiMasker(BaseMasker, _utils.CacheMixin):
 
         # Compute the mask if not given by the user
         if self.mask_img is None:
+            if imgs is None:
+                raise ValueError(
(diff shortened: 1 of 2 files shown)


bthirion (approved):
LGTM.

Remi-Gau (approved):
Long live meaningful error messages

Remi-Gau:
Just realized that another test is failing. 

Can fix as soon as I am not a tablet.

ymzayek:
Ok I see, so the failure has to due with order of errors basically. If we give an incorrect value to masker_strategy this will fail when `_get_mask_strategy` is called. I added the new error just above that so I can either move it under or change the test to pass an image object to fit.

Remi-Gau:
Thanks @ymzayek

PR #4156, opened 2023-12-12
[ENH]  rename `nilearn.datasets._utils.fetch_file` to `nilearn.datasets._utils.fetch_single_file`

### LINKED ISSUE #4155: rename either `nilearn.datasets._utils._fetch_file` or `nilearn.datasets._utils._fetch_files`
rename either `nilearn.datasets.utils._fetch_file` or `nilearn.datasets.utils._fetch_files`

but having both of them differing by a single letter is just not OK

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/4086#issuecomment-1783156630_

Remi-Gau:
- Closes #4155

- rename `nilearn.datasets._utils.fetch_file` to `nilearn.datasets._utils.fetch_single_file`

### DIFF
--- a/nilearn/datasets/_utils.py
+++ b/nilearn/datasets/_utils.py
@@ -483,7 +483,7 @@ class _NaiveFTPAdapter(requests.adapters.BaseAdapter):
 
 
 @fill_doc
-def fetch_file(
+def fetch_single_file(
     url,
     data_dir,
     resume=True,
@@ -530,7 +530,7 @@ def fetch_file(
     if session is None:
         with requests.Session() as session:
             session.mount("ftp:", _NaiveFTPAdapter())
-            return fetch_file(
+            return fetch_single_file(
                 url,
                 data_dir,
                 resume=resume,
@@ -611,7 +611,7 @@ def fetch_file(
             except Exception:
                 if verbose > 0:
                     print("Resuming failed, try to download the whole file.")
-                return fetch_file(
+                return fetch_single_file(
                     url,
                     data_dir,
                     resume=False,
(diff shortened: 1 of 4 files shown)


bthirion (approved):
LGTM. The codecov complaints are wrong, aren't they ?

Remi-Gau:
> The codecov complaints are wrong, aren't they ?

I would assume but those "false positives" make me unhappy.

ymzayek:
They could be correct. I haven't checked the tests closely but when changing code that wasn't fully tested previously these will be picked up. Same happens in the black PRs. I think ok not to address in this PR though

ymzayek (approved):
LGTM!

Remi-Gau:
> I haven't checked the tests closely but when changing code that wasn't fully tested previously these will be picked up.

OK let me recheck.

If we can increase coverage why not, but given that those are called by the fetchers my assumption was that the tests should hit those lines.

Remi-Gau:
OK had a look at the test and from my understanding the tests that "could" hit those lines, actually create mock dataset and hence the fetching never runs.

If we think this should be changed, I think it would probably be done in a separate PR.

https://github.com/nilearn/nilearn/blob/main/nilearn/datasets/tests/test_func.py#L954

https://github.com/nilearn/nilearn/blob/98ad01944dfb1b3fff0fc4306d852509cc1f225b/nilearn/datasets/tests/test_func.py#L982

https://github.com/nilearn/nilearn/blob/98ad01944dfb1b3fff0fc4306d852509cc1f225b/nilearn/datasets/tests/test_func.py#L1019

Issue #2950, opened 2021-09-08
Streamline the release process


tsalo:
I think we could use integrations and Actions to simplify the release process, with most of the work happening on GitHub instead of on the release maker's local machine.

We've discussed this elsewhere (perhaps in a dev call?), but I can't find any notes on it. Elizabeth pointed out that the main blocker was that the documentation needs to be built locally, so I know the workflow I'm most familiar with won't work on its own, but I'm sure there are some elements that can be adopted. In particular, I think versioneer and a PyPi deployment Action would be most useful.

### Benefits to the change

It would reduce the work necessary by maintainers to make a new release.

### The current release process

1. Prepare a new release on a new branch.
2. Update `whats_new.rst` by cleaning up existing notes and adding a highlights section.
3. ~~Manually update the version string in `nilearn/version.py`.~~ [edit: not needed thanks tho using hatchling]
4. Add new release post to News section of documentation.
5. Submit a PR from the release branch.
6. Merge release branch PR.
7. Tag the merge commit locally.
8. Build the tagged version and push to PyPi, also locally.
9. ~~Build documentation and push it to the GitHub Pages repo.~~ [edit: should now be automatic]
10. ~~Update `nilearn/version.py` and `whats_new.rst`.~~ [edit: not needed thanks tho using hatchling]

### A pie-in-the-sky release process

0. Use integrations to manage version strings and release notes.
    - Use release-drafter to automatically compile the release notes, along with a GitHub Action to update a CHANGES/whatsnew file.
        - This would probably use PR names in the files, so it's probably not sufficiently detailed for `nilearn`'s purposes, but I could be wrong.
    -  Use versioneer to automatically manage the version string.
1. Manually clean up `whats_new.rst` and update the documentation News section in a single PR.
    - Automatically grabbing the three most recent posts should be doable, right?
2. Update the pre-drafted release notes on GitHub and include the appropriate tag and name for the new release.
3. Make the release on GitHub.
    - A GitHub Action will build and deploy to PyPi.
4. Build documentation and push to GitHub Pages repo.
     - I would love to automate this, but my understanding is that it's just not possible with current resources. Is that correct?

bthirion:
"Automatically grabbing the three most recent posts should be doable, right?" -> note that you don't want the most recent, but the most important.

tsalo:
Could that rule be formalized as major and/or minor releases, but not patches, or does it need to be more of judgment call?

NicolasGensollen:
[...]

PR #4160, opened 2023-12-13
[INFRA] Proposal for building stable docs in remote runner for releases

ymzayek:
- Related to  https://github.com/nilearn/nilearn/issues/2950

TODO: 

- Test on fork


Remi-Gau:
Would be great to have this automated!!! 😻 

Let me know if you want me to check something in particular.

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,59 @@
...
+
+on:
+    # Allows you to run this workflow manually from the Actions tab
+    # It should be run right after tagging the release commit on main
+    workflow_dispatch:
+
+jobs:
+    release-docs:
+        # if: github.repository == 'nilearn/nilearn' && contains('["ymzayek","Remi-Gau"]', github.actor)
+        if: github.repository == 'ymzayek/nilearn' && github.actor == 'ymzayek'
```
bthirion, line 12: This will have to evolve ...

ymzayek: It's just for testing on my fork and being extra careful

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,59 @@
+---
+name: Build release docs
+
+on:
+    # Allows you to run this workflow manually from the Actions tab
+    # It should be run right after tagging the release commit on main
+    workflow_dispatch:
```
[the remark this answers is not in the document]
ymzayek: Yes can do

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,59 @@
...
+                    dvipng texlive-latex-base texlive-latex-extra
+
+        -   name: Setup python
+            uses: actions/setup-python@v5
+            with:
+                python-version: '3.12'
+
+        -   name: Install packages
+            run: |
+                python -m pip install --user --upgrade pip setuptools
```
ymzayek, line 40: Any reason setuptools was re-added during the tox implementation? @Remi-Gau

ymzayek: For reference I copied this line from build-docs.yml which was changed during the tox PR but we had removed setuptools when we switched to hatchling backend for builds

Remi-Gau: Going from memory, but pretty sure it got needed when running the doc build on python 3.12.

May be worth rechecking and if that's indeed the case documenting it somewhere.

.github/workflows/release-docs.yml
bthirion: I don't like to have stuff that we will rot quite quickly and repeatedly...

ymzayek: I'm not sure I understand. But before merging this it will only have 
`if: github.repository == 'nilearn/nilearn' `

ymzayek: In fact I'll already make the changing since I'm testing on my fork

.github/workflows/release-docs.yml
ymzayek: Ok sorry I missed that. Yes it seems to be the case. Can add to the readme

ymzayek:
Here is the succesful workflow on my fork

ymzayek:
whoops here it is: https://github.com/ymzayek/nilearn/actions/runs/7222661348/job/19680065387

So this should be good to go

Remi-Gau:
wait I am missing something.

should your workflow run show as a new commit on the doc repo: https://github.com/nilearn/nilearn.github.io/commits/main/ ?

ymzayek:
@Remi-Gau not on main but on this test branch: https://github.com/nilearn/nilearn.github.io/commits/test_build/

In my repo the workflow creates a branch for testing but this is not reflected in this PR

Remi-Gau:
gotcha!!
super cool. 
thanks a bunch for working on this.

Remi-Gau (approved):

Issue #4152, opened 2023-12-12
[DOC] improve surface related documentation 


Remi-Gau:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed suggestion in detail.

Before I forget again...

- add surface related terms to the glossary
  - mesh
  - vertex
  - faces (triangles)
  - coordinates
  - geometry

When looking up during the last mini hackathon I found that the following would good targets: mesh, faces, vertex. They fairly regularly in the code base and the documentation to justify new entries.

- I was also looking for potential sources we could point to to describe how surface data is usually structured. I have not had a lot of success. Anyone knows any? If not it may be useful to have a page describing this in the user guide, no? Also basing this on my memory of the feeling that, when starting with MRI, volumetric data was easy to grok but surface data much less so.

### List any pages that would be impacted.

- glossary
- User guide?

ymzayek:
Probably this section https://nilearn.github.io/stable/manipulating_images/index.html can be reworked a bit to mention SurfaceImage and SurfaceMasker and working with surface data within the context of nilearn

Remi-Gau:
Agreed: I was thinking of adding some things somewhere in this page.

https://nilearn.github.io/stable/manipulating_images/input_output.html#understanding-neuroimaging-data

PR #4161, opened 2023-12-13
[DOC] link surface terms to glossary

Remi-Gau:
- closes https://github.com/nilearn/nilearn/issues/4152

- add surface related terms to glossary (vertex, face, mesh) and links them in doc and codebase


doc/glossary.rst
```diff
@@ -90,6 +90,10 @@ If you wish to add a missing term, please
         Erosion uses a structuring element for probing and reducing the shapes
         contained in the input image.
 
+    faces
+        When referring to surface data, a face corresponds to one of the triangle
```
ymzayek, line 94:
```suggestion
        When referring to surface data, a face corresponds to one of the triangles
```

nilearn/experimental/surface/_surface_image.py
```diff
@@ -14,7 +14,7 @@
 
 
 class Mesh(abc.ABC):
-    """A surface mesh, having vertex coordinates and faces(triangles)."""
+    """A surface mesh, having vertex coordinates and faces (triangles)."""
```
ymzayek, line 17:
```suggestion
    """A surface mesh having vertex, coordinates and faces (triangles)."""
```

ymzayek: why not link the terms here too?

doc/sg_execution_times.rst
ymzayek, line 1: Why add this file? This is usually produced for each example anyways when built and can be viewed with this pattern e.g. `https://nilearn.github.io/stable/auto_examples/08_experimental/sg_execution_times.html` for each example although it's essentially "hidden" on the website. I don't think we need to add it right?

ymzayek (commented):
Thank @Remi-Gau !
I wonder if we should add 'texture'? It's not used a lot in the docs but more so as a variable name when it equals an array projected on the surface

doc/sg_execution_times.rst
Remi-Gau: yup that's a "misfire" from running an example. Sorry about that.

nilearn/experimental/surface/_surface_image.py
Remi-Gau: I missed it. Was working late last night. 🙈

Remi-Gau:
> Thank @Remi-Gau ! I wonder if we should add 'texture'? It's not used a lot in the docs but more so as a variable name when it equals an array projected on the surface

yes definitely

I had listed also:
- coordinates
- geometry

But they feel too generic, no?

Remi-Gau:
> if we should add 'texture'?

Only mentioned once: nilearn/examples/01_plotting/plot_3d_map_to_surface_projection.py:53

So maybe not?

bthirion:
No strong opinion on "textures". We could also use it more frequently...

Remi-Gau:
Would suggest to merge this as is and add texture later

bthirion:
Good with me.

Issue #4137, opened 2023-12-09
making the content of __all__ consistent with the API references in the docs. 


Remi-Gau:
One thing I'm a bit unsure about is that now in some cases as in the change of `_check_mesh` to `check_mesh`, how do we signal to the user and future developers that this function should be used internally (which also implies they can change without deprecation)? You can argue that it is not exposed in the API reference docs but some people don't look at that. One way could be to not include such functions in `__all__` and explicitly use the module.file name when calling the function as is now done with `check_mesh`. But then we'd need to make this consistent across modules (with the exceptions being _utils) making the content of `__all__` consistent with the API references in the docs. I'm not sure if this is the way to go so I'm happy to hear other thoughts

_Originally posted by @ymzayek in https://github.com/nilearn/nilearn/pull/4086#pullrequestreview-1719795351_

bthirion:
Maybe it's a good motivation to maintain the leading underscore.

Remi-Gau:
@ymzayek I tried to go through the inconsistencies between the declared public API of the different modules mentioned in their `__all__` and what we have in the user facing doc.

Did you have other inconsistencies in mind?

ymzayek:
No I think you've captured them well!
Maybe hold off on the ones in the surface module for now until we see about integrating the new one.

Remi-Gau:
yeah that was my thought on those as well

Remi-Gau:
Extracting comments from #4157 

in `nilearn/reporting/__init__.py`

```python
# TODO consider making _get_clusters_table module public?
from nilearn.reporting._get_clusters_table import get_clusters_table
```

in `nilearn/image/__init__.py`

```python
# TODO move copy_img out of nilearn._utils..niimg
from .._utils.niimg import copy_img

# TODO move concat_niimgs out of nilearn._utils.niimg_conversions
from .._utils.niimg_conversions import concat_niimgs as concat_imgs
```

Remi-Gau:
OK opening #4157 for review and will tackle the things I commented on in further PRs to keep PR / reviews smaller.

Remi-Gau:
Can be closed once #4166 and #4167 are done

PR #4166, opened 2023-12-16
[MAINT] move user facing functions out of private modules

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/issues/4137#issuecomment-1853664214
- Relates to #3319

Changes proposed in this pull request:

- make _get_clusters_table module public
- move copy_img out of nilearn._utils.niimg

PR #4167, opened 2023-12-16
[ENH] move user facing functions out of private modules - part 2

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/issues/4137#issuecomment-1853664214
- Relates to #3319

- move concat_imgs out of private module

### DIFF
--- a/nilearn/_utils/niimg_conversions.py
+++ b/nilearn/_utils/niimg_conversions.py
@@ -306,7 +306,9 @@ def check_niimg(
             return iter_check_niimg(
                 niimg, ensure_ndim=ensure_ndim, dtype=dtype
             )
-        return ni.image.concat_imgs(niimg, ensure_ndim=ensure_ndim, dtype=dtype)
+        return ni.image.concat_imgs(
+            niimg, ensure_ndim=ensure_ndim, dtype=dtype
+        )
 
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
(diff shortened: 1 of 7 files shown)


bthirion (commented):
LGTM, thx.

bthirion (approved):
LGTM, thx.

nilearn/_utils/niimg_conversions.py
```diff
@@ -418,124 +420,24 @@ def concat_niimgs(
...
+    warnings.warn(
+        (
+            "Import 'concat_niimgs' from 'nilearn._utils.niimg_conversions'\n"
+            "will be deprecated in version 0.13.\n"
+            "Importing from 'nilearn._utils.niimg_conversions'\n"
+            "will be possible at least until release 0.13.0.\n"
+            "Please import this function from 'nilearn.image' instead."
+        ),
+        DeprecationWarning,
+    )
```
ymzayek, line 433: Is this deprecation needed? `concat_niimgs` is a private function and not exposed in the docs, it is rather aliased as `concat_imgs` in nilearn.image so public API is not actually changing so it shouldn't affect "proper" user code

ymzayek (commented):
Otherwise it LGTM

nilearn/_utils/niimg_conversions.py
Remi-Gau: agreed this is me being extra cautious

am fine removing the warning if you think this is an overkill: it is true that we have a multiplication of new deprecation warnings we'll have to take care of in a couple of versions and it may be better to stick to the ones that are strictly needed.

bthirion: Ok with me to get rid of that one.

Remi-Gau: Wait: just to be clear

@ymzayek you are arguing to remove `concat_niimgs` entirely from `nilearn/_utils/niimg_conversions.py`, not just the warning, right?

- Same for `copy_img` in `nilearn/_utils/niimg.py` in PR #4166?

Because otherwise we are left with private functions that are just "redirects" to public ones which feels a bit futile.

ymzayek: Yes I would remove it

Remi-Gau:
Removed the old function.

Unless I hear otherwise I will merge tomorrow.

nilearn/_utils/tests/test_niimg_conversions.py
```diff
@@ -606,100 +605,9 @@ def test_repr_niimgs_with_niimg(
...
-
-    with pytest.raises(DimensionError, match=_dimension_error_msg):
-        _utils.concat_niimgs([img1, img4d])
-
-    img5d = Nifti1Image(np.ones((2, 2, 2, 2, 2)), affine_eye)
-    with pytest.raises(
-        TypeError,
-        match="Concatenated images must be 3D or 4D. "
-        "You gave a list of 5D images",
+def test_concat_niimgs_warning(img_3d_zeros_eye):
```
ymzayek, line 608: Also remove this test

PR #4180, opened 2024-01-02
[FIX] use a local server rather than temp file to open plots in a browser

jeromedockes:
the reports and html plots (eg `view_img`) have an `open_in_browser` method to open their web page in a web browser.

the user can provide a file name, in which case the page is saved there and the `file:///[...]` url is opened; this path is unchanged by this PR.

If the user does not provide a file name, ATM we write the page in a temporary file, open the file in the browser, then start a process that removes the temporary file after a fixed timeout.
this has several disadvantages:

- the temporary file location may not be visible to the browser. For example on recent ubuntu, by default firefox will be installed with snap and will be in a sandbox where it cannot see `/tmp` which is the default location for temporary files. therefore **opening a nilearn plot will show a "file not found" error in firefox on ubuntu**
- the temporary file may fail to be removed
- we have no way of knowing if opening the plot actually succeeded

this PR proposes an alternative: we start a local server, wait for it to serve one request (from the browser, with a timeout), and shut it down


nilearn/plotting/html_document.py
```diff
@@ -155,41 +190,20 @@ def open_in_browser(self, file_name=None, temp_file_lifetime=30):
...
-            Time, in seconds, after which the temporary file is removed.
-            If None, it is never removed.
 
+            .. deprecated:: 0.11
 
+                The parameter is kept for backward compatibility and will be
+                removed in a future version. It has no effect.
         """
+        if temp_file_lifetime != "deprecated":
+            warnings.warn(
```
ymzayek, line 200: Can you specify the deprecation class as `DeprecationWarning`?

ymzayek (commented):
@jeromedockes thanks this pr LGTM! Can you increase the test coverage a bit?

bthirion:
Thx for opening this ! 
No other comments than Yasmin's.

Remi-Gau (approved):
Good with me.

Side note: Not sure what the Files changed tab is showing so many lines as uncovered when the actual codecov report shows them as covered.

Not the first time I see this and wondering what is triggering this.

doc/changes/latest.rst
```diff
@@ -10,6 +10,7 @@ HIGHLIGHTS
 
 NEW
 ---
+- The ``temp_file_lifetime`` parameter of interactive plots' ``open_in_browser`` method is deprecated and has no effect.
```
ymzayek, line 13:
```suggestion
- :bdg-info:`Plotting` The ``temp_file_lifetime`` parameter of interactive plots' ``open_in_browser`` method is deprecated and has no effect (:gh:`4180` by `Jerome Dockes`_)..
```

ymzayek (approved):
Thx @jeromedockes ! Just modified the whatsnew. Will merge when everything passes

nilearn/plotting/html_document.py
```diff
@@ -155,41 +190,21 @@ def open_in_browser(self, file_name=None, temp_file_lifetime=30):
...
         temp_file_lifetime : :obj:`float`, default=30
-            Time, in seconds, after which the temporary file is removed.
-            If None, it is never removed.
 
+            .. deprecated:: 0.11
 
+                The parameter is kept for backward compatibility and will be
+                removed in a future version. It has no effect.
         """
+        if temp_file_lifetime != "deprecated":
```
Remi-Gau, line 199: Forgot about it but maybe better to use this helper decorator for this:

https://github.com/nilearn/nilearn/blob/1374eee6159f048072efa9c9c7db62cbac28f96d/nilearn/_utils/helpers.py#L109

PR #4185, opened 2024-01-05
[draft] Surface glm example

bthirion:
Follows from  #4126 just as a draft for discussion.

- This adapts the `plot_localizer_surface_analysis.py` example
- Still in draft mode: the example is not yet great because plotting is awkward in particular.


ymzayek:
@jeromedockes and @alexisthual given the planning you've done for a plotting to handle the new surface object as input, would you be able to already draft a PR to do this using this example as a target to apply it to?

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -59,56 +58,46 @@
...
-# representing the geometry of the surface.
-# We could use an individual :term:`mesh`,
-# but we first resort to a standard :term:`mesh`,
-# the so-called fsaverage5 template from the FreeSurfer software.
-import nilearn
+# For this we need to get a mesh representing the geometry of the surface. We
+# could use an individual mesh, but we first resort to a standard mesh, the
+# so-called fsaverage5 template from the FreeSurfer software.
 
-fsaverage = nilearn.datasets.fetch_surf_fsaverage()
```
ymzayek, line 69: would we then want to deprecate this function and update `experimental.surface.load_fsaverage` to include all the mesh types?

bthirion: I'm wondering whether there is still a use case for having all the paths, not the objects.

ymzayek: A path-like object can still be accessed from the `FileMesh` object created by `load_fsaverage`.

```python
from nilearn.experimental import surface
fsaverage = surface.load_fsaverage()
fsaverage['pial']['left_hemisphere'].file_path
```

ymzayek:
I suggest to close this in favor of #4126. @bthirion you can directly push changes there if needed

PR #4188, opened 2024-01-05
[MAINT] Adapt decoder for scikit-learn > 1.3

### LINKED ISSUE #4132: [MAINT] Resolve sklearn 1.4.dev0 related failure of decoder test
The failure still seen here in the numpy2 job: [test_decoder_binary_classification_with_masker_object](https://github.com/nilearn/nilearn/actions/runs/7115012784/job/19370222715?pr=4130)
is actually related to an inconsistency between nilearn and scikit-learn dev version from their nightly releases which we have to use to test against numpy v2.0. All the numpy deprecations so far are dealt with in this PR so I'd rather merge this and deal with the sklearn deprecation separately

_Originally posted by @ymzayek in https://github.com/nilearn/nilearn/issues/4130#issuecomment-1842916077_

ymzayek:
- Closes #4132

- Change inheritance of decoder classes
- Move data extraction to `decision_function`
- Fix some docstrings (already not accurate in stable version)
- Check backward compatibility; may need deprecation


.github/workflows/numpy2_compatibility.yml
```diff
@@ -11,7 +11,6 @@ on:
 
 jobs:
     test_numpy2:
-        if: github.repository == 'nilearn/nilearn'
```
ymzayek, line 14: TODO: Revert before merging PR

nilearn/decoding/decoder.py
```diff
@@ -1201,7 +1201,7 @@ def __init__(
 
 
 @fill_doc
-class DecoderRegressor(_BaseDecoder):
+class DecoderRegressor(_BaseDecoder, MultiOutputMixin, RegressorMixin):
```
glemaitre, line 1204: From the top of the head, the mixins need to be on the most left to not have issue with the MRO (old discussions: https://github.com/scikit-learn/scikit-learn/pull/8022#issuecomment-267091483)

glemaitre:
The changes looks good to me apart from the order of inheritance for the MRO.

ymzayek:
@glemaitre thanks for pointing that out. I fixed the order though then I had to explicitly define `score` under the two decoder classes. Not sure if there is another way to do it. I added:

```python
    def score(self, X, y, *args):
        """Call score method from base class."""
        return _BaseDecoder.score(self, X, y, *args)
```

Does it look ok now?

bthirion (approved):
Thx for taking care of that one !

nilearn/decoding/decoder.py
```diff
@@ -1338,6 +1347,10 @@ def __init__(
             n_jobs=n_jobs,
         )
 
+    def score(self, X, y, *args):
+        """Call score method from base class."""
+        return _BaseDecoder.score(self, X, y, *args)
```
glemaitre, line 1352: OK I see. In terms of MRO, it means that we favor the `score` define in the mixin instead of the base class. Normally, this is usually what we want. But this is not the case, here. However, if we put the mixin on the right, we will break all the tag system.

Since `_BaseDecoder` is defining the score function, it means that you actually don't want to have the functionality defined by the mixin so you don't need to inherate. So from this point, the right fix to me is the following:

```python
class _BaseDecoder(...):
    def score(self, ...):
        ...
    def _more_tags(self):
        return {"require_y": True}

class Decoder(_BaseDecoder):
    _estimator_type = "classifier"

    ...

class DecoderRegressor(MultiOutputMixin, _BaseDecoder):
    _estimator_type = "regressor"

    ...
```

The use of the `RegressorMixin` and `ClassifierMixin` in your case is just required to defined the `_estimator_type` so let's define it directly in the children class. The second thing that we want is the definition of the `"require_y"` tag that can be done in the base class since this is common to both mixins.

I did not foresee the part or redefining the `score` function. Let me know if this is making sense to you @ymzayek.

ymzayek: > I did not foresee the part or redefining the score function. Let me know if this is making sense to you @ymzayek.

I did not either and then I tried to go around it with redefining `score` but you are right, we do not need those mixins and can just define the needed attribute and tag function as you suggest. I was coming to the same conclusion so makes total sense. Thanks!

I'll make the changes and then I think it should be good to go

nilearn/decoding/decoder.py
```diff
@@ -1065,6 +1065,9 @@ def _predict_dummy(self, n_samples):
             )
         return scores.ravel() if scores.shape[1] == 1 else scores
 
+    def _more_tags(self):
+        return {"require_y": True}
```
glemaitre, line 1069: You can easily cover this line by adding a check that `require_y` is set to True and `mulitoutput` is False for the classifier and True for the regressor. It allows to check that the MRO works as expected as well.

Remi-Gau (approved):
Looks good to me

nilearn/decoding/decoder.py
```diff
@@ -870,6 +871,9 @@ def decision_function(self, X):
         y_pred: ndarray, shape (n_samples,)
             Predicted class label per sample.
         """
+        # for backwards compatibility
+        if not isinstance(X, type(np.array)):
+            X = self.masker_.transform(X)
```
ymzayek, line 876: Sorry I though it was ready but then realized that this change to take X as a niimg was not backward compatible so I added this conditional. However, the docstring says X is a niimg like object so it can also be seen as a bug. Should we deprecate behavior here instead? Not sure how often people call this function directly but it is not even tested directly in our test suite

nilearn/decoding/decoder.py
```diff
@@ -870,6 +871,9 @@ def decision_function(self, X):
         y_pred: ndarray, shape (n_samples,)
             Predicted class label per sample.
         """
+        # for backwards compatibility
+        if not isinstance(X, type(np.array)):
```
glemaitre, line 875: This is looking weird to me. Do you mean:

```python
if not isinstance(X, np.ndarray):
    ...
```

Usually, having `type` in an `isinstance` looks weird (sometimes it is because you intend to call `issubclass`).

ymzayek: Yes you are right, thanks!

ymzayek:
If there are no more comments I will merge this when the CI finishes

PR #4190, opened 2024-01-05
[STY] deal with private functions in nilearn.glm that do not coform to PEP8 naming

Remi-Gau:
- Closes none
- Relates to #3319 

- move `_check_events_file_uses_tab_separators` and `_check_run_tables` and their sub functions" from `_utils.py` to `first_level.py` closer to where there are only used.
- move tests accordingly
- rename private functions `_cosine_drift` -> `create_cosine_drift` and `_orthogonalize` -> `orthogaonalize` as they are used outside of the modules where they are defined.


bthirion (approved):
LGTM, thx !

ymzayek (commented):
LGTM. One question: shouldn't the functions renamed to without a leading underscore be listed in the init (such as orthogonalize)?

Remi-Gau:
> LGTM. One question: shouldn't the functions renamed to without a leading underscore be listed in the init (such as orthogonalize)?

At the the moment the `"__all__"` of public modules define whats is part of the user facing API and matches what's in the API documentation.

The leading underscores should just signal if that function is private to the module it is in or not. Given that `_orthogonalize` is used outside of the module it is defined in (for example in `design_matrix.py`), it is not private to that module so we should remove the leading underscore. But I don't think it warrants making it part of the user facing API.

ymzayek:
@Remi-Gau yes that's right thanks for clarifying. That was the old way of doing it and I just had a little mental lapse

ymzayek (approved):
good to go

bthirion:
Merging, then.

PR #4191, opened 2024-01-05
[FIX] rename `contrast_type` to `stat_type`

### LINKED ISSUE #4186: [BUG] inconsistency in the naming

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 22.04
- Windows 11

### Python version

- [ ] 3.12
- [ ] 3.11
- [X] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

dev version

### Expected behavior

The same argument has different names whether called in the `compute_contrast` function of the corresponding method of the *GLM classes.
`contrast_type`  in the function, see  https://github.com/nilearn/nilearn/blob/cfc4c03f/nilearn/glm/contrasts.py#L66
`stat_type` in teh method, see https://github.com/nilearn/nilearn/blob/cfc4c03fe2533dfffd6a6f034f1d6404262e1ec1/nilearn/glm/first_level/first_level.py#L763C9-L763C9
We should  deprecate one of them. 
My imrpession is that `stat_type` is more accurate.

This is what I got:

```bash


```

```python
# Paste your code here


```

Remi-Gau:
- Closes #4186

- rename parameter `contrast_type` by `stat_type` in the function `compute_contrast`, see https://github.com/nilearn/nilearn/blob/cfc4c03f/nilearn/glm/contrasts.py#L66
- rename attribute `contrast_type` by `stat_type` the Contrast class
- apply similar change in downstream private functions


bthirion:
We'll have to go through deprecation cycles for that one...

nilearn/glm/contrasts.py
```diff
@@ -47,7 +48,10 @@ def expression_to_contrast_vector(expression, design_columns):
     return contrast_vector
 
 
-def compute_contrast(labels, regression_result, con_val, contrast_type=None):
+@rename_parameters(
+    replacement_params={"contrast_type": "stat_type"}, end_version="0.13.0"
+)
```
Remi-Gau, line 53: This decorator should take care of the deprecation cycle.

Will add a test to make sure of it.

Remi-Gau:
@bthirion 
for consistency I would also change the attribute contrast_type in the Contrast class.

https://github.com/nilearn/nilearn/blob/cfc4c03fe2533dfffd6a6f034f1d6404262e1ec1/nilearn/glm/contrasts.py#L178

What do you think?

bthirion:
Indeed, the idea is to have a uniform naming across classes and functions.

nilearn/glm/first_level/first_level.py
```diff
@@ -986,6 +986,7 @@ def first_level_from_bids(
     sub_labels : :obj:`list` of :obj:`str`, optional
         Specifies the subset of subject labels to model.
         If 'None', will model all subjects in the dataset.
+
```
Remi-Gau, line 989: This is unrelated but came across a directive that did not render in the doc.

https://nilearn.github.io/stable/modules/generated/nilearn.glm.first_level.first_level_from_bids.html#nilearn.glm.first_level.first_level_from_bids

image

bthirion (approved):
LGTM, thx !

PR #4202, opened 2024-01-10
implement walrus operator

Remi-Gau:
- Closes none

- add hook to implement walrus operator
- run all pre-commit hooks


nilearn/datasets/func.py
```diff
@@ -3040,8 +3037,7 @@ def _glob_fiac_data():
         return Bunch(**_subject_data)
 
     # maybe data_dir already contains the data ?
-    data = _glob_fiac_data()
-    if data is not None:
+    if (data := _glob_fiac_data()) is not None:
```
Remi-Gau, line 3040:
```suggestion
    if data := _glob_fiac_data():
```

Remi-Gau:
walrus operator is a python 3.8 feature so just adding a pre-commit hook to enforce it given that we have dropped python 3.7

Curious to see how others feel about it.

ymzayek:
I like it but some of its effects that may not be clear makes me hesitant about enforcing it. No strong opinion though. Here's this blog post about it: https://martinheinz.dev/blog/79

bthirion:
No strong feeling about this. We should follow what the majority of the community considers best practices.

Remi-Gau:
this is defo not urgent so we can wait to get more feedback

jeromedockes:
weak -1 for me. even though it's been around for a few releases I don't see the walrus operator being used that often; I wouldn't be surprised if some contributors don't know what it is; the updated code in the diff is not really better than the old one so I don't see a reason for the change. 

As 3.7 support is dropped I don't mind new code using that operator when the contributor wants to use it, but I don't see the logic behind systematically enforcing its use wherever it could possibly be applied

jeromedockes:
more generally I would really like to avoid going overboard with the precommit and linting stuff and only enforce coding style rules that actually make the codebase easier to maintain or contribute to (including for new contributors)

Remi-Gau:
Will close this.

If we want to systematically implement it re recreating this should not cost much as it can be automated

PR #4215, opened 2024-01-12
[DOC] update dataset descriptions

### LINKED ISSUE #4210: [DOC] missing description of several datasets

it seems that several datasets (many of them used for GLMs) are missing a description in here:

https://github.com/nilearn/nilearn/tree/main/nilearn/datasets/description

I would suggest:
- adding a test to make sure that each fetcher does return description
- adding the description where it is missing

I also think that the description of those datasets should appear somewhere in the html version documentation, because unless I am mistaken those descriptions need to be "called" by users after fetching a dataset if they want to know more about it.

It feels that having all those .rst files and not render them as html is a bit of waste.

_No response_

Remi-Gau:
- Closes #4210

- add failing tests to show missing descriptions
- add dataset descriptions and make sure they are returned by all fetchers (except the openneuro one as the description could be different for every dataset)

unrelated: update some of the doc string in datasets.func.py


nilearn/datasets/func.py
```diff
@@ -2235,10 +2249,11 @@ def fetch_bids_langloc_dataset(data_dir=None, verbose=1):
     main_path = os.path.join(data_dir, main_folder)
     file_list = [
         os.path.join(path, f)
-        for path, dirs, files in os.walk(main_path)
+        for path, _, files in os.walk(main_path)
         for f in files
     ]
-    return os.path.join(data_dir, main_folder), sorted(file_list)
+    description = get_dataset_descr("bids_langloc")
+    return os.path.join(data_dir, main_folder), sorted(file_list), description
```
Remi-Gau, line 2256: Because this fetcher returns a tuple and not a Bunch, this may break the API of people who use this fetcher in their tutorial.

I'd be tempted to say we should not go through a deprecation because
- not returning a description feels more like a bug that it is being fixed
- dataset fetchers are more used for our examples are rarely used in "production" for research

nilearn/datasets/func.py
```diff
@@ -2195,10 +2205,11 @@ def fetch_language_localizer_demo_dataset(data_dir=None, verbose=1):
 
     file_list = [
         os.path.join(path, f)
-        for path, dirs, files in os.walk(data_dir)
+        for path, _, files in os.walk(data_dir)
         for f in files
     ]
-    return data_dir, sorted(file_list)
+    description = get_dataset_descr("language_localizer_demo")
+    return data_dir, sorted(file_list), description
```
Remi-Gau, line 2212: Break of API: see comment in https://github.com/nilearn/nilearn/pull/4215/files#r1450382982

nilearn/datasets/func.py
Remi-Gau: Ideally I would prefer to have these fetchers also return a bunch rather than a tuple, but I think this should be part of a different PR.

Remi-Gau:
TODOs for other PRs:
- centralize dataset / atlas description contained to the description RST file:
  - in doc string (if there is) of the fetcher 
  - in doc strings of the examples that use the fetcher
- print the description of the dataset in the examples that use the fetcher
- render the description RST files as part of the doc
- make sure that articles mentioned in the descriptions are part of our bibliography

nilearn/datasets/func.py
ymzayek: Agreed. It's also good for consistency with the other fetchers

bthirion: > Because this fetcher returns a tuple and not a Bunch, this may break the API of people who use this fetcher in their tutorial.
> 
> I'd be tempted to say we should not go through a deprecation because
> 
>     * not returning a description feels more like a bug that it is being fixed
> 
>     * dataset fetchers are more used for our examples are rarely used in "production" for research

Not sure about that one, but you're probably right. +0

nilearn/datasets/description/localizer_first_level.rst
```diff
@@ -0,0 +1,34 @@
+localizer first level
```
Remi-Gau, line 1: dataset hosted here: https://osf.io/v7hsw/?view_only=

but cannot find a whole lot more info about it

nilearn/datasets/func.py
Remi-Gau: Will at least mention it in the changelog in any case.

Remi-Gau:
Looking at the data from those 2 datasets, I am not sure I can tell the difference between the 2?

Events files look kind of similar.

- fetch_bids_langloc_dataset
- fetch_language_localizer_demo_dataset

Also the bids_langloc_dataset seems unused in our examples.

doc/changes/latest.rst
```diff
@@ -18,8 +18,11 @@ Fixes
 - :bdg-dark:`Code` Update the ``CompCor`` strategy in :func:`~interfaces.fmriprep.load_confounds` and :func:`~interfaces.fmriprep.load_confounds_strategy` to support ``fmriprep`` 21.x series and above. (:gh:`3285` by `Hao-Ting Wang`_).
 - :bdg-success:`API` :class:`~maskers.MultiNiftiMasker` can now call :meth:`~maskers.NiftiMasker.generate_report` which will generate a report for the first subject in the list of subjects (:gh:`4001` by `Yasmin Mzayek`_).
 - :bdg-dark:`Code` Fix :class:`~nilearn.glm.regression.SimpleRegressionResults` to accommodate for the lack of a ``model`` attribute (:gh:`4071` `Rémi Gau`_)
+- :bdg-dark:`Code` Make sure that all atlases and datasets fetchers return a description. :func:`~datasets.fetch_language_localizer_demo_dataset` and  :func:`~datasets.fetch_bids_langloc_dataset`. (:gh:`4215` `Rémi Gau`_)
```
ymzayek, line 21: why link to those 2 datasets here?

examples/07_advanced/plot_beta_series.py
```diff
@@ -73,7 +73,7 @@
 from nilearn.datasets import fetch_language_localizer_demo_dataset
 from nilearn.glm.first_level import FirstLevelModel, first_level_from_bids
 
-data_dir, _ = fetch_language_localizer_demo_dataset()
+data_dir, _, _ = fetch_language_localizer_demo_dataset()
```
ymzayek, line 76: I'd avoid doing this unless you plan to deprecate the return behavior in this PR

Remi-Gau: yeah I was not 100% sure that this would justify deprecation

see my comment here: https://github.com/nilearn/nilearn/pull/4215#discussion_r1450382982

but this is not a hill I will die on, so if you feel otherwise we can go through a deprecation

in any case, maybe cleaner to do all of this in a separate PR where we ensure that all data fetchers return a Bunch.

let me know what you think

ymzayek: > in any case, maybe cleaner to do all of this in a separate PR where we ensure that all data fetchers return a Bunch.

I would prefer to go with this then

Remi-Gau: okidoki: let me trim this PR down and move the API change to another PR.

doc/changes/latest.rst
Remi-Gau: wanted to mention the API "breakage" but that will now go in a separate PR with deprecation

nilearn/datasets/description/spm_auditory.rst
```diff
@@ -0,0 +1,39 @@
...
+A structural image was also acquired.
+
+This experiment was conducted by Geriant Rees
+under the direction of Karl Friston and the FIL methods group.
+
+Content
+-------
+        - 'func': :obj:`list` of :obj:`str`. Paths to functional images
+        - 'anat': :obj:`list` of :obj:`str`. Path to anat image
+        - 'description': :obj:`str`. Data description
```
ymzayek, line 24: can this indent match the others or is it related to the way it renders?

nilearn/datasets/func.py
```diff
@@ -79,17 +79,21 @@ def fetch_haxby(
...
           Paths to nifti ventral temporal mask file.
-        - 'mask_house_little': string list.
+        - 'mask_face': :obj:`list` of :obj:`str`.
+          Paths to nifti ventral temporal mask file.
+        - 'mask_house': :obj:`list` of :obj:`str`.
+          Paths to nifti ventral temporal mask file.
+        - 'mask_face_little': :obj:`list` of :obj:`str`.
+          Paths to nifti ventral temporal mask file.
+        - 'mask_house_little': :obj:`list` of :obj:`str`.
           Paths to nifti ventral temporal mask file.
```
ymzayek, line 97: Are these descriptions correct?

doc/changes/latest.rst
```diff
@@ -15,12 +15,16 @@ NEW
...
+- :bdg-dark:`Code` Make sure that all atlases and datasets fetchers return a description (:gh:`4215` `Rémi Gau`_).
 - :bdg-dark:`Code` Update the ``CompCor`` strategy in :func:`~interfaces.fmriprep.load_confounds` and :func:`~interfaces.fmriprep.load_confounds_strategy` to support ``fmriprep`` 21.x series and above. (:gh:`3285` by `Hao-Ting Wang`_).
 - :bdg-success:`API` :class:`~maskers.MultiNiftiMasker` can now call :meth:`~maskers.NiftiMasker.generate_report` which will generate a report for the first subject in the list of subjects (:gh:`4001` by `Yasmin Mzayek`_).
 - :bdg-dark:`Code` Fix :class:`~nilearn.glm.regression.SimpleRegressionResults` to accommodate for the lack of a ``model`` attribute (:gh:`4071` `Rémi Gau`_)
 - :bdg-dark:`Code` :func:`~image.clean_img` can now use kwargs ``clean__sample_mask`` argument to correctly reshape the nifti image to the dimensions of the mask in the output (:gh:`4051` by `Mia Zwally`_).
 - :bdg-dark:`Code` Fix plotting of an image with color bar when maximum value is exactly zero (:gh:`4204` by `Rémi Gau`_).
 
+
+
+
```
ymzayek, line 27:
```suggestion
```

nilearn/datasets/func.py
Remi-Gau: actually they were too vague but the dataset descriptions has better ones so I copied them over

nilearn/datasets/description/spm_auditory.rst
Remi-Gau: I had actually not finished cleaning up this one. Thanks for catching it.

Those description don't render anywhere for now, so there may be a lot of clean up to do when start rendering them.

bthirion (commented):
LGTM, thx for all the cleaning.

Issue #4098, opened 2023-11-07
[DOC] experimental-plot-surface-image-and-maskers example does not render well


bthirion:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed suggestion in detail.

The decoding maps proposed at the end of the example are ugly and not informative. We probably need to tweak viz parameters.
https://nilearn.github.io/stable/auto_examples/08_experimental/plot_surface_image_and_maskers.html#sphx-glr-auto-examples-08-experimental-plot-surface-image-and-maskers-py

### List any pages that would be impacted.

https://nilearn.github.io/stable/auto_examples/08_experimental/plot_surface_image_and_maskers.html

Remi-Gau:
not yet fixed

PR #4229, opened 2024-01-17
[WIP][ENH] Move surface image API from experimental to stable API

ymzayek:
- Part of #4027

Changes proposed in this pull request:

- Changes compared to `experimental.surface`
  - Fetchers moved to datasets respectively to `atlas`, `func`, and `struct`
    - `fetch_destrieux` renamed `load_sample_atlas_surf_destrieux`
    - `fetch_nki` renamed `load_sample_surf_nki_enhanced`
    - `load_fsaverage` renamed `load_surf_fsaverage`
  - `mesh_name` parameter of `load_surf_fsaverage` renamed `mesh` for consistency with `fetch_surf_fsaverage`
  - `Mesh` base class renamed `_Mesh` 
    - `Mesh` already exists as namedtuple which we will probably deprecate

TODO:
- [x] Double check public class/method and function names
- [x] Double check/improve docstrings (still need review of course)
- [ ] How to handle I/O and provide some public functions to replace [surface.py](https://nilearn.github.io/stable/modules/surface.html) loaders
  - At the moment just copied over wrappers from `experimental.surface._io` to `surface._io`
- [x] Remove type hints
- [x] Update doc/modules/surface.rst and remove experimental.rst
- [x] Update experimental surface example
  - [ ] Integrate parts into stable examples (to do in follow-up PRs)
- [x] Increase testing
- [ ] What to deprecate from old surface.py API?

The following PRs will depend on merging this one:
#4126 
#4205
#4120 - already merged, relevant imports are updated in this PR

PR #4235, opened 2024-01-23
[ENH] add plotting functions to the experimental surface module

jeromedockes:
added a placeholder basic plotting function

- closes https://github.com/nilearn/nilearn/issues/4098

TODO
- [x] change PolyMesh, PolyData to constrain keys to be "left" and "right" (required for plotting when we want medial or lateral view)
- [ ] split experimental package into: `maskers`, `plotting`, `datasets` to make them easier to later integrate into the rest of the code base
- [ ] add functions to return fsaverage data (sulcal, curvature...) as SurfaceImage instance with specific mesh (pial, flat...)
- [ ] `bg_map` of the new plotting functions can be a `SurfaceImage` instance
- [ ] add method to `SurfaceImage` to check the presence of a given hemisphere in data and mesh
- [ ] more plotting functions

Examples to adapt:

- [ ] examples/01_plotting/plot_3d_map_to_surface_projection.py
- [ ] examples/01_plotting/plot_surf_atlas.py
- [ ] examples/01_plotting/plot_surf_stat_map.py


bthirion (commented):
Thx for kicking that off !

Remi-Gau:
Note to ourselves: most likely there will be some merge conflicts with #4229

jeromedockes:
> Note to ourselves: most likely there will be some merge conflicts with #4229

yes 😰

Remi-Gau:
@jeromedockes 
to maybe minimize conflicts, maybe check #4229 and let me know if some bits should be adapted to make it easier to merge this after.

jeromedockes:
> @jeromedockes 
> to maybe minimize conflicts, maybe check #4229 and let me know if some bits should be adapted to make it easier to merge this after.

sure, sorry I haven't had time to follow that as closely as I would have
wanted. Are you and Yasmin going to the office hours today? Maybe we can
use it to do a quick catch-up on the remaining outstanding points

Remi-Gau:
@jeromedockes 

added the examples to adapt in the top message of the PR.

I am OK to tackle those in a separate PR if you prefer. This is just to let you know.

Remi-Gau:
@jeromedockes 
want me to take over this one?

jeromedockes:
> @jeromedockes 
> want me to take over this one?

Hi Remi sorry for the delay. Please feel free to do so, no problem but
actually I think it might be easier to start a new one as most of the
diff in this one is to add the `.parts` attribute to polymesh and poly
data, which in the end we decided against during the office hours

Remi-Gau:
@jeromedockes 

Will try to adapt this one first. It cannot take THAT long to revert. #FamousLastWords

nilearn/experimental/surface/_datasets.py
```diff
@@ -41,8 +42,8 @@ def fetch_nki(n_subjects=1) -> Sequence[SurfaceImage]:
         img = SurfaceImage(
             mesh=fsaverage["pial"],
             data={
-                "left_hemisphere": left_data,
```
Remi-Gau, line 44: sticking to left and right to match vali hemipsheres names we have in the plotting modules:

https://github.com/nilearn/nilearn/blob/1c90f3fc379d6450b2f4493001b8f9747366276c/nilearn/plotting/surf_plotting.py#L29

Can be changed but would be nice to keep things internally consistent

Remi-Gau:
@jeromedockes so far I have gone quite nulcear and reverted most of the changes you made.

question: should we keep working with data classes for now ?

nilearn/experimental/plotting/_surface_plotting.py
```diff
@@ -0,0 +1,146 @@
...
+        figsize=(4 * len(hemi), 4),
+    )
+    axes = np.atleast_2d(axes)
+
+    title = ""
+    if "title" in kwargs:
+        title = f"{kwargs.pop('title')} - "
+
+    for view, ax_row in zip(views, axes):
+        for ax, mesh_part in zip(ax_row, hemi):
```
Remi-Gau, line 33: I don't we want our plotting functions to be that complex.
Should let users decide how they ant to arrange their subplots.

examples/08_experimental/plot_3d_map_to_surface_projection.py
```diff
@@ -0,0 +1,341 @@
...
+    mesh=fsaverage_meshes["pial"],
+    data={
+        "left": surface.vol_to_surf(
+            stat_img, fsaverage_meshes["pial"]["left"]
+        ),
+        "right": surface.vol_to_surf(
+            stat_img, fsaverage_meshes["pial"]["right"]
+        ),
+    },
+)
```
Remi-Gau, line 70: projecting to surface will probably be a fairly classic use case that it would probably justify having a methods for it.

Remi-Gau:
PR is getting pretty long because of:
- making copies of the examples to adapt into experimental examples to test the new plotting functions.
- splitting the experimental module further into `datasets`, `maskers`...

Probably worth stopping here for now and make further changes in separate PR.

I can also split the examples in different PRs if easier for review.

Remi-Gau:
@jeromedockes if you have time to cross check 😉

nilearn/experimental/plotting/_surface_plotting.py
```diff
@@ -0,0 +1,268 @@
...
+from nilearn.experimental.surface import Mesh, PolyMesh, SurfaceImage
+
+
+@fill_doc
+def plot_surf(
+    surf_map: SurfaceImage | str | numpy.array | None = None,
+    hemi: str = "left",
+    surf_mesh: (
+        str | list[numpy.array, numpy.array] | Mesh | PolyMesh | None
+    ) = None,
```
Remi-Gau, line 18: given that currently surface_mesh can be a str pointing to a file containing a mesh, I think it may be appropriate to also be able to pass a SurfaceImage to surf_mesh

like is done for bg_map

examples/08_experimental/plot_3d_map_to_surface_projection_experimental.py
```diff
@@ -0,0 +1,336 @@
...
+    mesh=fsaverage_meshes["pial"],
+    data={
+        "left": surface.vol_to_surf(
+            stat_img, fsaverage_meshes["pial"]["left"]
+        ),
+        "right": surface.vol_to_surf(
+            stat_img, fsaverage_meshes["pial"]["right"]
+        ),
+    },
+)
```
Remi-Gau, line 68: TODO in another PR: add method to SurfaceImage to handle vol to surf

nilearn/experimental/plotting/_surface_plotting.py
```diff
@@ -0,0 +1,268 @@
...
+            surf_map=surf_map,
+            hemi=hemi,
+            bg_map=bg_map**kwargs,
+        )
+
+    if surf_mesh is None:
+        surf_mesh = surf_map.mesh
+
+    surf_map._check_hemi(hemi)
+    _check_hemi_present(surf_mesh, hemi)
```
Remi-Gau, line 49: this will fail if surf_mesh is not a PolyMesh
fixing this will probably need better define the PolyMesh class

nilearn/experimental/datasets/_datasets.py
```diff
@@ -0,0 +1,147 @@
...
+                "left": left_data,
+                "right": right_data,
+            },
+        )
+        images.append(img)
+
+    return images
+
+
+def fetch_destrieux(
```
Remi-Gau, line 125: testing this will require improving the mocking of the dataset: would probably do this in another PR

nilearn/experimental/datasets/_datasets.py
```diff
@@ -0,0 +1,147 @@
...
+        "flat": "flat",
+    }
+    meshes: dict[str, dict[str, Mesh]] = {
+        mesh_name: {
+            f"{hemisphere}": FileMesh(fsaverage[f"{mesh_type}_{hemisphere}"])
+            for hemisphere in ("left", "right")
+        }
+        for mesh_type, mesh_name in renaming.items()
+    }
+    renaming = {"curv": "curvature", "sulc": "sulcal", "thick": "thickness"}
```
bthirion, line 39: Why are you doing this here ?

nilearn/experimental/plotting/_surface_plotting.py
bthirion: Note sure I follow you here.

nilearn/experimental/plotting/_surface_plotting.py
bthirion: Can you clarify the discussion here ?

bthirion (commented):
Needs more discussion.

nilearn/experimental/plotting/_surface_plotting.py
Remi-Gau: Not 100% sure about this, so to be discussed but it seems that it would convenient if users could directly pass a SurfaceImage instance to this parameter and let nilearn get the Polymesh from the instance rather than forcing users to pass just a Mesh / PolyMesh.

I may be overthinking this but if users can just deal with fewer types of objects (SurfaceImage).

nilearn/experimental/plotting/_surface_plotting.py
Remi-Gau: `_check_hemi_present` is just to make sure that the "PolyMesh" dict has the correct hemisphere

But most of our new code at the moment, does not use the PolyMesh class but simple dictionaries instead, so we cannot do a `isinstance(surf_mesh, PolyMesh)`. So we should probably develop the PolyMesh class, because at the moment it is only `PolyMesh = Dict[str, Mesh]`.

Also at this stage `surf_mesh` can also be a str, list of two numpy.ndarray, Mesh and none of those cases are handled.

nilearn/experimental/plotting/_surface_plotting.py
jeromedockes: it would be weird to pass a different image than the one we are plotting just to grab the mesh from it. either the mesh is `None` in which case we use the mesh that comes with `surf_map`, or we want to provide a different mesh (eg an inflated surface) in which case we pass `surf_mesh`, which should be a mesh. if it happens to be stored as part of another image (which is not very likely if it is an inflated or flat surface used just for plotting) the user can always do `img.mesh` to access it

nilearn/experimental/datasets/_datasets.py
Remi-Gau, line 39:
```suggestion
    renaming = {"curv": "curvature", "sulc": "sulcal", "thick": "thickness"}
```

that line should have been removed

jeromedockes (commented):
ok I'll try to have a look this week. My first comment is I don't see the need for the change from 
```
module.py
```
to
```
module/
   __init__.py (empty besides grabbing names from _module.py)
   _module.py
   (no other submodules)
```

nilearn/experimental/plotting/_surface_plotting.py
Remi-Gau: okidoki
not a hill I wanna die on and if users request it we can think about changing that later

but then do you think that it makes sense to be able to pass a SurfaceImage for the background map the way I have done?

nilearn/experimental/plotting/_surface_plotting.py
Remi-Gau: Not I could start making those changes in this PR but I was getting afraid of having an even bigger PR

Remi-Gau:
> ok I'll try to have a look this week. My first comment is I don't see the need for the change from
> 
> ```
> module.py
> ```
> 
> to
> 
> ```
> module/
>    __init__.py (empty besides grabbing names from _module.py)
>    _module.py
>    (no other submodules)
> ```

For this I was mostly thinking of making a bit easier to integrate all this code into the rest of the code base later.

jeromedockes:
> but then do you think that it makes sense to be able to pass a SurfaceImage for the background map the way I have done?

yes because the background map is an actual image that has both a mesh
and data, right?

Remi-Gau:
@jeromedockes 
I moved things back into a single subpackage.

Let me know if you think that I should start creating "proper" classes for PolyMesh and PolyData in this PR or not

examples/08_experimental/plot_3d_map_to_surface_projection_experimental.py
jeromedockes: surfaceimage is for holding surface data, i don't think it should be responsible for converting volumes

Remi-Gau:
@bthirion @jeromedockes @man-shu 
I have reset the PR but kept my changes in another branch.

I have also reset the list of TODOs in the top of the PR, because I think several of those should still be part of our move to the new API but they can be done in several other PRs, where I can bring things back from my other branch.

Will clean things up a bit to make sure CI is happy and then will ping you again for a final review.

Remi-Gau:
I have also removed all my comments in the discussion threads because most of those were not relevant anymore. If you want you can do the same to yours to make the PR thread less confusing.

Remi-Gau:
@jeromedockes @bthirion @man-shu 

I will start working on a follow up PR to this one that adapt the other examples (the one already in the repo) to work with the new surface object: I had already this in this PR but it will make the review a bit easier if split in several PRs.

nilearn/experimental/plotting/_surface_plotting.py
```diff
@@ -0,0 +1,36 @@
...
+from nilearn import plotting as old_plotting
+from nilearn.experimental.surface import SurfaceImage
+
+
+def plot_surf(
+    img, part: str | None = None, mesh=None, view: str | None = None, **kwargs
+):
+    """Plot a SurfaceImage.
+
+    TODO: docstring.
```
bthirion, line 12: THis one need to be done.

bthirion (commented):
The PR looks good. I am a bit lost wrt the global plan: is this PR just one step in the  plan ?

nilearn/experimental/plotting/_surface_plotting.py
Remi-Gau: Will do that in the follow up PR

Remi-Gau:
> I am a bit lost wrt the global plan: is this PR just one step in the plan ?

in brief yes!

some if the open blocked PR sort of rely on this one being merged before they can move on

will merge this one and adapt other PRs accordingly

PR #3191, opened 2022-03-24
[DOC] Combine similar two-runs GLM examples

tsalo:
Closes #2771
Closes #4211

Changes proposed in this pull request:

- Combine plot_fiac_analysis.py and plot_fixed_effects.py into plot_two_session_model.py.
- Update a link in first_level_model.rst that originally went to plot_fiac_analysis.py.
- Example now relies on the fact that Nilearn will zero pad contrast matrices (uncovered a bug that got fixed)
- Update some of the code to use `run` instead of `session`
- Fix plotting of unpadded contrast matrices and add test for it

PR #4250, opened 2024-01-30
update doc for release

### LINKED ISSUE #4245: update maintenance doc related to doc pushing
Minor detail.

Current version says

> Once the new version tag of a release is pushed upstream following the step above, the Github Actions workflow release-docs.yml will be triggered automatically to build the stable docs and push them to our github pages repository nilearn/nilearn.github.io.

Should probably say to the workflow will be triggered when creating a new **github** release.

Remi-Gau:
- Closes #4245 

Updates to release process and its doc

- add a `git push` to push the doc to be deployed
- move section about github release
- mention zenodo DOI that was not explicitly mentioned anywhere

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -190,5 +190,6 @@ install:
 	git add -A && \
 	git add .nojekyll && \
 	git commit -a -m 'Make install' && \
+	git push && \
 	git tag $(VERSIONTAG) && \
 	git push --tags
--- a/doc/maintenance.rst
+++ b/doc/maintenance.rst
@@ -279,14 +279,6 @@ Once the PR has been reviewed and merged, pull from master and tag the merge com
     When building the distribution as described below, ``hatch-vcs``, defined in ``pyproject.toml``,
     extracts the version number using this tag and writes it to a ``_version.py`` file.
 
-Build of stable docs
---------------------
-
-Once the new version tag of a release is pushed upstream following the step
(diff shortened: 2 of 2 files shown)


bthirion (approved):
LGTM. 
Bzw, I realize we don't have an example showcasing #4067
It would be good to show it otherwise we may surprise users.

Remi-Gau:
> I realize we don't have an example showcasing #4067 It would be good to show it otherwise we may surprise users.

So it was included when we merged the similar GLM exmaples in #3191 : 

Resulting examples has "short" contrasts and there is warning that will always say when the contrasts was padded.

https://nilearn.github.io/stable/auto_examples/04_glm_first_level/plot_two_runs_model.html#compare-run-specific-and-fixed-effects-contrasts

It felt right to do in this case, because the older examples did specifically include a padding function:

https://github.com/nilearn/nilearn/blob/960589c05b6a63ed612f2feb749ee4472e5aa43f/examples/04_glm_first_level/plot_fiac_analysis.py#L85

But you are correct that the padding is now happening "implicitly" (except for the warning) and it should probably be mentioned in the text of the example.

Remi-Gau:
check https://github.com/nilearn/nilearn/issues/4245#issuecomment-1916371137 before merging

bthirion:
OK, i had not realized that. Well, the warning plays its role. Note sure we need to do much more than that.

Remi-Gau:
superseded by #4371

PR #4256, opened 2024-02-02
[FIX] Fix color bar handling with color map with only 1 level

### LINKED ISSUE #4255: [BUG] plot_roi cannot handle cmap with only 1 level

### Operating system

- [ ] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 22.04
- Windows 11

### Python version

- [ ] 3.12
- [ ] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

0.10.3

### Expected behavior

Plot ROI as usual

Given an image with only one level (i.e. 0s and 1s), and a custom cmap, `plot_roi` throws an error constructing the `color_bar`.

```bash
<ipython-input-14-60f0ec2e3bf9> in <module>
----> 1 niplt.plot_roi(img, alpha=0.8, colorbar=True, cmap=cmap)

~/anaconda3/lib/python3.9/site-packages/nilearn/plotting/img_plotting.py in plot_roi(roi_img, bg_img, cut_coords, output_file, display_mode, figure, axes, title, annotate, draw_cross, black_bg, threshold, alpha, cmap, dim, colorbar, cbar_tick_format, vmin, vmax, resampling_interpolation, view_type, linewidths, radiological, **kwargs)
    921     )
    922 
--> 923     display = _plot_img_with_bg(
    924         img=roi_img,
    925         bg_img=bg_img,

~/anaconda3/lib/python3.9/site-packages/nilearn/plotting/img_plotting.py in _plot_img_with_bg(img, bg_img, cut_coords, output_file, display_mode, colorbar, figure, axes, title, threshold, annotate, draw_cross, black_bg, vmin, vmax, bg_vmin, bg_vmax, interpolation, display_factory, cbar_vmin, cbar_vmax, cbar_tick_format, brain_color, decimals, radiological, **kwargs)
    266 
    267     if img is not None and img is not False:
--> 268         display.add_overlay(
    269             new_img_like(img, data, affine),
    270             threshold=threshold,

~/anaconda3/lib/python3.9/site-packages/nilearn/plotting/displays/_slicers.py in add_overlay(self, img, threshold, colorbar, cbar_tick_format, cbar_vmin, cbar_vmax, **kwargs)
    314         # look at test_img_plotting.test_outlier_cut_coords.
    315         if colorbar and ims:
--> 316             self._show_colorbar(
    317                 ims[0].cmap, ims[0].norm, cbar_vmin, cbar_vmax, threshold
    318             )

~/anaconda3/lib/python3.9/site-packages/nilearn/plotting/displays/_slicers.py in _show_colorbar(self, cmap, norm, cbar_vmin, cbar_vmax, threshold)
    551                 "Custom cmap", cmaplist, our_cmap.N
    552             )
--> 553         self._cbar = ColorbarBase(
    554             self._colorbar_ax,
    555             ticks=ticks,

~/anaconda3/lib/python3.9/site-packages/matplotlib/colorbar.py in __init__(self, ax, mappable, cmap, norm, alpha, values, boundaries, orientation, ticklocation, extend, spacing, ticks, format, drawedges, extendfrac, extendrect, label, location)
    409         else:
    410             self._formatter = format  # Assume it is a Formatter or None
--> 411         self._draw_all()
    412 
    413         if isinstance(mappable, contour.ContourSet) and not mappable.filled:

~/anaconda3/lib/python3.9/site-packages/matplotlib/colorbar.py in _draw_all(self)
    535         self.vmin, self.vmax = self._boundaries[self._inside][[0, -1]]
    536         # Compute the X/Y mesh.
--> 537         X, Y = self._mesh()
    538         # draw the extend triangles, and shrink the inner axes to accommodate.
    539         # also adds the outline path to self.outline spine:

~/anaconda3/lib/python3.9/site-packages/matplotlib/colorbar.py in _mesh(self)
   1109         orientation.
   1110         """
-> 1111         y, _ = self._proportional_y()
   1112         # Use the vmin and vmax of the colorbar, which may not be the same
   1113         # as the norm. There are situations where the colormap has a

~/anaconda3/lib/python3.9/site-packages/matplotlib/colorbar.py in _proportional_y(self)
   1249         # make the lower and upper extend lengths proportional to the lengths
   1250         # of the first and last boundary spacing (if extendfrac='auto'):
-> 1251         automin = yscaled[1] - yscaled[0]
   1252         automax = yscaled[-1] - yscaled[-2]
   1253         extendlength = [0, 0]

IndexError: index 1 is out of bounds for axis 0 with size 1
```

```python
# Paste your code here

from nilearn import plotting as niplt
import numpy as np
import nibabel as nib

array_data = np.arange(24, dtype=np.int16).reshape((2, 3, 4))
array_data[:] = 0
array_data[0, 1, 1] = 1
affine = np.diag([1, 2, 3, 1])
img = nib.Nifti1Image(array_data, affine)
clust_ids = list(np.unique(img.get_fdata())[1:])
cmap = plt.cm.get_cmap("tab20", len(clust_ids))
cmap = plt.cm.get_cmap("tab20", len(clust_ids))
niplt.plot_roi(img, alpha=0.8, colorbar=True, cmap=cmap)

```

Remi-Gau:
- Closes #4255

- add regression test
- add code to fix


examples/01_plotting/plot_demo_glass_brain_extensive.py
```diff
@@ -102,6 +102,7 @@
     threshold=2,
     symmetric_cbar=False,
     cmap="viridis",
+    title="only plot positive values",
```
Remi-Gau, line 105: those plots were added in the #3993 
just adding some title so the figures are easier to identify when running the example dumps > 20 figures on you

nilearn/plotting/displays/_slicers.py
```diff
@@ -395,6 +395,16 @@ def _map_show(
...
 
+        if threshold is not None:
+            data = safe_get_data(img, ensure_finite=True)
+            if threshold == 0:
+                data = np.ma.masked_equal(data, 0, copy=False)
+            else:
+                data = np.ma.masked_inside(
+                    data, -threshold, threshold, copy=False
+                )
+            img = new_img_like(img, data, img.affine)
```
Remi-Gau, line 406: @michellewang do you remember why this got removed in #3993 ?

I can see you refactored part of it into a new method used later in the code but I cannot quite see if readding this code now will interfere with something.

bthirion (commented):
some lines are not covered by tests ?

Remi-Gau:
> some lines are not covered by tests ?

Strange because the codecov patch check says that every is fine.

Remi-Gau:
https://github.com/nilearn/nilearn/pull/4256/checks?check_run_id=21154504057

nilearn/plotting/displays/_slicers.py
michellewang: @Remi-Gau I tried plotting some things and I don't see anything obviously wrong. Since the lines you added back only address the threshold part of the masking (not the `vmin`/`vmax` parts of the `_threshold` function I added), I don't think it would break anything. However, it might be confusing/not optimal to have 2 places in a function where thresholding/masking is done? I checked and the reason why I moved the thresholding further down in that function is because, for glass brain plotting with custom `vmin`, the thresholding needed to be done after we take the absolute value of the data if `plot_abs` is `True` (otherwise we incorrectly lose whatever is lower than `vmin`), but for some reason the absolute value is only taken in [`GlassBrainAxes.transform_to_2d`](https://github.com/nilearn/nilearn/blob/0bf5f707d9f2fa10021fef1994acffc03a44e08e/nilearn/plotting/displays/_axes.py#L438), which gives the 2D views to be plotted in each panel. It doesn't look like there is currently an easy way to take the absolute value of the whole 3D data (possibly because `plot_abs` is only for glass brain plots?), but maybe we can add a `plot_abs` parameter to the slicer objects and do the absolute value handling inside `_map_show`?

Anyway if this needs to be fixed as soon as possible I think your fix should be fine, but in the future we might want to consider not having the double thresholding/masking part.

bthirion: I think you're right. Passing the argument below sounds like the best fix.

Remi-Gau: I'd say that we merge this as is and do the refactoring suggested by @michellewang in another PR to keep things easier to review

PR #4257, opened 2024-02-02
[FIX] Replace remaining `contrast_type` by `stat_type`

larsoner:
Looks like `contrast_type` was deprecated but at least one internal use was not updated to `stat_type`. Discovered because I tried to follow the deprecation instructions when using `compute_contrast` but kept getting a DeprecationWarning.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -8,6 +8,8 @@
 Fixes
 -----
 
+- :bdg-dark:`Code` Fix errant warning when using ``stat_type`` in :func:`nilearn.glm.compute_contrast` (:gh:`4257` by `Eric Larson`_)
+
 Enhancements
 ------------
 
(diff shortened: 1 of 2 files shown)


Remi-Gau:
Looks good.

Do you want to update the changelog?

https://github.com/nilearn/nilearn/blob/main/doc/changes/latest.rst?plain=1

Remi-Gau:
I hear you, but if a first time contributor opened their first PR ever to fix a simple typo I think it is worth acknowledging them, so I personally prefer to have the same rule to everyone.

TMK we don't have an official rule at nilearn for what constitute a minimum contribution, personally I prefer to set the bar pretty low. 😉

larsoner:
Pushed, feel free to push a commit to tweak as necessary!

bthirion (approved):
thx !

Remi-Gau (approved):
thansk @larsoner

Issue #2194, opened 2019-10-25
Adding tests for rapid inspection of visual elements


kchawla-pi:
**What would you like changed/added and why?**

Quoting @jeromedockes  here, from https://github.com/nilearn/nilearn/pull/2191#issuecomment-546278888
Since we are adding reports etc. we should add more meaningful inspection of generated plots and reports in the tests

**What would be the benefit? Does the change make something easier to use?**

 I think CircleCI does all this but it takes way too long. Maybe a faster simpler test to quickly generate plots and reports to visually examine will be useful?

jeromedockes:
>
>  I think CircleCI does all this but it takes way too long. Maybe a faster simpler test to quickly generate plots and reports to visually examine will be useful?

Not really, I meant that the tests are at the moment ad-hock and superficial

kchawla-pi:
You mean something like Automated Visual Inspection for example what Selenium does for the Web Frontend?

jeromedockes:
I think you can close this one for now

emdupre:
One short term solution that gets to this idea -- we could create a [reviewing template, such as the one used by fMRIPrep](https://github.com/poldracklab/fmriprep/wiki/Reviewing-a-Pull-Request) that includes specific pages of the documentation that should be visually inspected !

Remi-Gau:
Reading this I realize that our doc has no examples of what the reports look like. Am it right?

Maybe instead of having fully automated visual tests à la selenium, we could add examples on the doc of the different reports that nilearn can generate.

This would at least allow us to inspect them manually like any other part of the doc and also to point users to them.

Ideally we could have the examples generate the reports to integrate in the doc.

ymzayek:
@Remi-Gau can you be more specific? Sorry if I misunderstand but there are calls to maskers' `generate_report` in some examples and in the user guide. It's true that for GLM `generate_report` is never called in an example but instead `make_glm_report` is showcased, which is the function called by `glm.generate_report`. Do you mean to centralize this info and/or be more explicit about the expected outcome?
Btw should we favor calling `glm.generate_report` over `make_glm_report` where possible?

Remi-Gau:
I admit I had more the GLM report in mind. 

I think I had more in mind a specific section in the docs for reports only and where each report is show cased.

At the moment even we as maintainers have to remember in which examples those reports are generated to make sure there is no "regression" when we check things manually.

[...]

PR #4258, opened 2024-02-02
[FIX] threshold images right before ploting in GLM reports

### LINKED ISSUE #4192: [BUG] Glm reports does not apply image threshold at the right place

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 22.04
- Windows 11

### Python version

- [ ] 3.12
- [ ] 3.11
- [X] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

dev version

### Expected behavior

In the example below, the map plotted is all black in report2, because it is reinterpolated after thresholding (compare with report1.html, that displays the expected behavior). 
Interpolation is triggered by the use of a non-diagonal (up to permutation) affine matrix. 
The thresholding should instead occur at  plotting time, so that it would be correctly mask out non-significant area:
for this the `threshold` argument should be frowarded to function `_stat_map_to_svg` in https://github.com/nilearn/nilearn/blob/310f5ca6/nilearn/reporting/glm_reporter.py#L48

see report2.html

```python
from nilearn.datasets import func
import numpy as np
import pandas as pd
from nilearn.reporting import make_glm_report
import nibabel as nib

data = func.fetch_fiac_first_level()
fmri_img = data['func1']
mean_img_ = mean_img(fmri_img)
design_matrices = pd.DataFrame(np.load(data['design_matrix1'])['X'])

fmri_glm = FirstLevelModel(mask_img=data['mask'], minimize_memory=True)
fmri_glm = fmri_glm.fit(fmri_img, design_matrices=design_matrices)
n_columns = design_matrices.shape[1]
contrasts = {'Effects_of_interest': np.eye(n_columns)[:5]}

report = make_glm_report(fmri_glm,
                         contrasts,
                         bg_img=mean_img_,
                         )
report.save_as_html('report1.html')


data_  = nib.load(fmri_img).get_fdata()
affine = np.array([[-2.9, -.1, -.1, 0], [.1, -2.9, -.3, 0], [.1, .3, 3.9, 0], [0, 0, 0, 1]])
fmri_img_ = nib.Nifti1Image(data_, affine)
mean_img_ = mean_img(fmri_img_)
mask_img_ = nib.Nifti1Image(nib.load(data['mask']).get_fdata(), affine=affine)
fmri_glm = FirstLevelModel(mask_img=mask_img_, minimize_memory=True)
fmri_glm = fmri_glm.fit(fmri_img_, design_matrices=design_matrices)
report = make_glm_report(fmri_glm,
                         contrasts,
                         bg_img=mean_img_,
                         )
report.save_as_html('report2.html')
```

Remi-Gau:
- Closes #4192

- pass threshold to `_stat_map_to_svg`


Remi-Gau:
before:

image


after:

image

Also adding all before / after reports from the code snippet from  #4192

[report.zip](https://github.com/nilearn/nilearn/files/14143208/report.zip)

Remi-Gau:
@bthirion 

I feel we should have a "test" for this in the visual inspection suite of the GLM reporter.

https://github.com/nilearn/nilearn/blob/310f5ca659eb135b0c9489199b0c0cc9895780fa/nilearn/reporting/_visual_testing/_glm_reporter_visual_inspection_suite_.py

bthirion:
> @bthirion
> 
> I feel we should have a "test" for this in the visual inspection suite of the GLM reporter.
> 
> https://github.com/nilearn/nilearn/blob/310f5ca659eb135b0c9489199b0c0cc9895780fa/nilearn/reporting/_visual_testing/_glm_reporter_visual_inspection_suite_.py

This means that we just need to look at the results from time to time ? Before releases ? Then we should add that to the dev doc.

Remi-Gau:
> This means that we just need to look at the results from time to time ? Before releases ? Then we should add that to the dev doc.

That would be the "generate once in a while / manual check".

Other alternative "generate automatically / manual check": would be the actually show what the reports look like as part of our doc: see this comment in another thread https://github.com/nilearn/nilearn/issues/2194#issuecomment-1884889353

In any case this would be for another PR.

nilearn/reporting/glm_reporter.py
```diff
@@ -763,28 +763,38 @@ def _make_stat_maps_contrast_clusters(
...
     for contrast_name, stat_map_img in stat_img.items():
         component_text_ = string.Template(components_template_text)
-        thresholded_stat_map, threshold = threshold_stats_img(
+
+        # Only use threshold_stats_img to adjust the threshold
+        # that we will pass to  _clustering_params_to_dataframe
+        # and _stat_map_to_svg
+        # Necessary to avoid :
+        # https://github.com/nilearn/nilearn/issues/4192
+        _, threshold = threshold_stats_img(
```
bthirion, line 772: Actually you cannot do that: you need to get the thresholdedimage because you want to take into account the cluster_threshold that is provided. In the implementation you propose, this gets forgotten.

Remi-Gau:
report 1 after:

image

report 2 after:

image

[report_after.zip](https://github.com/nilearn/nilearn/files/14238208/report_after.zip)

nilearn/reporting/glm_reporter.py
```diff
@@ -224,8 +224,7 @@ def make_glm_report(
             mask_img = model.masker_.mask_img_
 
     mask_plot_html_code = _mask_to_svg(
-        mask_img=mask_img,
-        bg_img=bg_img,
```
Remi-Gau, line 228: unrelated but realized that the cut_coords was not passed to the mask plotting

bthirion (approved):
Seems to do the job properly, thx.

Remi-Gau:
Will update the changelog to log the bug fix

PR #4262, opened 2024-02-05
[DOC] Add versionadded directive to `two_sided` parameter of `binarize_img` function

NicolasGensollen:
Hi,

It looks like a `two_sided` parameter was recently added to `binarize_img` (see #4121). However, the documentation doesn't specify the version in which it was added (0.10.3).
Not sure if people use these, but I actually was looking for this information on the API page of the function, and ended up looking at the changelog to know.


Remi-Gau:
good catch

thanks @NicolasGensollen

PR #4279, opened 2024-02-21
[FIX] use proper grid spec when plotting carpet plots

### LINKED ISSUE #4278: Failure of carpet plot with matplotlib-3.9.0.dev0
See workflow: https://github.com/nilearn/nilearn/actions/runs/7985927180/job/21805307226#step:6:5388

Due to PR: https://github.com/matplotlib/matplotlib/pull/27565

Should be fixable as explained in here: https://stackoverflow.com/questions/77710768/getting-constrained-layout-to-work-with-nested-subplots-in-matplotlib

Remi-Gau:
- Closes #4278 

- fixes #4278
- runs numpy 2 tests on every PR

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -2,12 +2,13 @@
 
 .. include:: names.rst
 
-0.11.0.dev
+0.10.4.dev
 ==========
 
 Fixes
 -----
 
+- :bdg-dark:`Code` Fix plotting of carpet plot due to a change in the coming version of matplolib (3.9.0) (:gh:`4279` by `Rémi Gau`_).
 - :bdg-dark:`Code` Fix errant warning when using ``stat_type`` in :func:`nilearn.glm.compute_contrast` (:gh:`4257` by `Eric Larson`_).
 - :bdg-dark:`Code` FIX when thresholding is applied to images by GLM reports (:gh:`4258` by `Rémi Gau`_).
 


.github/workflows/numpy2_compatibility.yml
```diff
@@ -5,10 +5,9 @@ name: Numpy2 compatibility
 
 on:
 
-    # Comment before merging in main
-    # pull_request:
-    #    branches:
-    #    -   '*'
+    pull_request:
```
bthirion, line 8: oops, this seems to be unrelated.

examples/01_plotting/plot_carpet.py
```diff
@@ -83,8 +84,9 @@
     axes=ax,
     cmap="gray",
     standardize="zscore_sample",
+    title="global patterns over time separated by tissue type",
 )
 
-fig.show()
+plt.show()
```
bthirion, line 90: This is incompatible with latest mpl ?

bthirion (commented):
Thx for tackling this !

.github/workflows/numpy2_compatibility.yml
Remi-Gau: Actually this is intentional: this is the 3rd time that I have had to go and manually check in the action tab to see if something numpy2 related was broken. I'd rather have this check run regularly for better automatic early detection.

By doing this it will be checked with every PR: this would also ensure that every PR does not bring changes that would be numpy2 incompatible.

examples/01_plotting/plot_carpet.py
Remi-Gau: nope this is because if you run the example as a script the figures closes right away and you do not have time for proper visual inspection

PR #4337, opened 2024-03-21
[FIX] Headers not carried over by `math_img`

### LINKED ISSUE #2645: math_img loses header information
In nilearn 0.7.0, using math_img to get the first 180 dynamics of an image. 

```
img = math_img("img[:,:,:,0:180]",img=img)
``` 
Then I save the image and look at the header with fslhd.

new header has pixdim4 as 1:

```
size of header	348
data_type	FLOAT64
dim0		4
dim1		144
dim2		144
dim3		41
dim4		180
dim5		1
dim6		1
dim7		1
vox_units	Unknown
time_units	Unknown
datatype	64
nbyper		8
bitpix		64
pixdim0		-1.000000
pixdim1		1.647059
pixdim2		1.647059
pixdim3		3.000000
pixdim4		1.000000
```

Original header here has pixdim4 as 2:

```
size of header	348
data_type	INT16
dim0		4
dim1		144
dim2		144
dim3		41
dim4		360
dim5		1
dim6		1
dim7		1
vox_units	mm
time_units	s
datatype	4
nbyper		2
bitpix		16
pixdim0		-1.000000
pixdim1		1.647059
pixdim2		1.647059
pixdim3		3.000000
pixdim4		2.000001
```

I'm using nibabel to do this now, but this might need fixing.

man-shu:
- Closes #2645

- ~~[x] copy last image's header in the result image (i.e. if the call is `math_img("img1+img2", img1, img2)`, the result would have img2's header)~~
- [x] copy user-specified image's header in the result image
- ~~[x] check if the images have the same TRs, if not throw the "Input images cannot be compared..." error~~
- [x] check if the number of dimensions is the same between that user-specified image and the result image. Checks for affine and shapes already exist.
- [x] add a short example to demo the usage of the new parameter
- ~~[ ] test the exceptions and equality of TRs~~
- [x] test for the error when the number dimensions are not the same
- [x] test for equality of header (or just the TR) when the correct `copy_header_from` is passed
- [x] test for the error when a non-existent variable is passed in `copy_header_from`
- [x] test for default header (or TR being 1.0) when the `copy_header_from` is `None` 
- [x] test for cases where data values are changed, that changes the `cal_max` and `cal_min` fields in the header, but the rest remains the same.


bthirion:
Maybe also include resampling functions in this PR ?

man-shu:
Seems like a job for `regex` to me..

bthirion:
Oops. It looks very hard to me to handle  properly all cases To me his would justify not to copy the whole header by default...

man-shu:
I agree. The solution I am trying to implement uses:
-  `regex` to identify individual operations (from `numpy`, `math`)
-  doing `eval` on each of them
-  then another `eval` on the remaining basic mathematical operations (* / + -)

Will try to push it soon. But this might be difficult to maintain in long-term, given the complexity of regex.

Same issues would apply to `resample_img` also, which only handles **spatial** resampling as of now. 
So comparing time resolutions doesn't seem useful to me.

Speaking of resampling only handling **spatial resampling** we should probably clarify that in the docs.

bthirion:
Maybe just add an argument to copy the header information, that would be False by default, that people could use at their own risk. Demoing the feature would be relevant.

man-shu:
I think it would also be useful to allow the users to select which `img`'s header to copy.

Remi-Gau:
Still at a conference. Don't have time for an in depth review just yet, but you may want to check why a test is failing with the oldest dependency.

man-shu:
Yeah, it's a failure in doctest. I am downloading a dataset in the short example right under the reference to `math_img` function, that seems to fail for some reason...

nilearn/image/tests/test_image.py
```diff
@@ -837,6 +849,55 @@ def test_math_img(
         assert result.shape == expected_result.shape
 
 
+def test_math_img_copied_header(
+    img_4d_ones_eye_default_header, img_4d_ones_eye_tr2, img_4d_mni_tr2
+):
+    img_eye_default = img_4d_ones_eye_default_header
+    img_eye_tr2 = img_4d_ones_eye_tr2
+    img_mni_tr2 = img_4d_mni_tr2
+
```
bthirion, line 858: Do we need this renaming ?

bthirion (commented):
LGTM overall.
Thx !

man-shu:
Ok, so @Remi-Gau  and I found that the doctest is failing because I am trying to fetch a dataset in the docstring example, and that is currently not possible in the testing runs on the CI. 

So writing a full example to demo the `copy_header_from` functionality would be the solution. Additionally, we can go into full details of what works, and what does not in that example.

I looked into other full examples on `math_img` but describing this feature in those examples looks a bit out of the scope.

So, I am now writing a separate standalone example for this.

Remi-Gau:
> So, I am now writing a separate standalone example for this.

maybe we can have this in a separate PR, so we can already merge the fix?

nilearn/conftest.py
Remi-Gau, line 1: If you are only going to use the fixtures, in in single test module, maybe best to keep them there: closer to what they are being used.

In general you want to put in conftest.py fixtures that are used in very large number of tests in the test suite.

nilearn/image/tests/test_image.py
Remi-Gau: Also maybe try to split this in 3 tests: one each case.

This way an eventual failure of the first assert would not prevent other cases to be run.

Remi-Gau: also shorter test are easier to read, because you don't have to worry if something  written 30 lines higher still matters to keep in mind

nilearn/conftest.py
man-shu: We actually might have to use them in other places. I haven't opened up those issues yet, but similar header copying problems exist in `resample_img` and `mean_img` -- these would need their own separate fixes.

Remi-Gau: ah I see 

then I think you can just move them in a module called `nilearn/image/tests/conftest.py`

nilearn/image/tests/test_image.py
```diff
@@ -837,6 +849,59 @@ def test_math_img(
...
+):
+    # case where data values are not changed and header values are not copied
+    # the result should have default header values
+    formula_no_change = "img * 1"
+    # using img_4d_ones_eye_tr2 with edited header in the formula
+    result = math_img(
+        formula_no_change, img=img_4d_ones_eye_tr2, copy_header_from=None
+    )
+    # header values should instead match default header values
+    assert result.header.__eq__(img_4d_ones_eye_default_header.header)
```
Remi-Gau, line 863:
```suggestion
    assert result.header == img_4d_ones_eye_default_header.header
```

I believe this should work, no?

nilearn/image/tests/test_image.py
```diff
@@ -837,6 +849,59 @@ def test_math_img(
...
+
+def test_math_img_copied_header_from_img(img_4d_mni_tr2):
+    # case where data values are not changed but header values are copied
+    # the result should have the same header values as the given image
+    formula_no_change = "img * 1"
+    result = math_img(
+        formula_no_change, img=img_4d_mni_tr2, copy_header_from="img"
+    )
+    # all header values should be the same
+    assert result.header.__eq__(img_4d_mni_tr2.header)
```
Remi-Gau, line 874:
```suggestion
    assert result.header == img_4d_mni_tr2.header
```

nilearn/image/image.py
```diff
@@ -1033,6 +1042,19 @@ def math_img(formula, **imgs):
...
+    see :gh:`2645`.
+
+    .. versionadded:: 0.10.4.dev
+
+    We can also copy the header from one of the input images using
+    ``copy_header_from``::
+
+     >>> result_img_with_header = math_img("img1 + img2",
+     ...                                   img1=anatomical_image, img2=log_img,
+     ...                                   copy_header_from="img1")
```
Remi-Gau, line 1056: curious to see if doc test will work with this or if it will complain that `anatomical_image` is unknown for example

man-shu: Just checked, works fine. `anatomical_image` is defined just before for the pre-existing doc string example. It doesn't show up in the diff I think.

Remi-Gau: ah!!! nice!!! did not realize this.

nilearn/image/tests/test_image.py
man-shu: Just checked and it does indeed work. 
I think I read somewhere that the `__eq__` method is supposed to be more thorough. 
But let's stick to simpler syntax for now.

nilearn/conftest.py
man-shu: done!

nilearn/image/tests/conftest.py
Remi-Gau, line 1: did a bit of refactoring in this commit

https://github.com/nilearn/nilearn/pull/4337/commits/457c33d64f0105a220b1aecd3aff1920ef8eedbe

more to show how you can use fixture directly when building new fixtures, so no need to "import" them

Remi-Gau (approved):

Remi-Gau (changes requested):
Actually we probably want to update the changelog before merging

man-shu:
> Actually we probably want to update the changelog before merging

Good to know. I wasn't doing that in previous PRs. Done now!

doc/changes/latest.rst
```diff
@@ -49,3 +49,5 @@ Changes
 
 - :bdg-dark:`Code` Use red to blue color map in the GLM reports (:gh:`4266` by `Rémi Gau`_).
 - :bdg-dark:`Code` :class:`nilearn.maskers.NiftiSpheresMasker` will throw warnings if the ``labels`` passed to it is not a :obj:`list` of :obj:`str`, or if the number of items in the list of labels does not match the number of regions in the label image (:gh:`4274` by `Rémi Gau`_).
+
+- :bdg-dark:`Code` Copy headers from user-specified image to the result of :func:`nilearn.image.math_img` (:gh:`4337` by `Himanshu Aggarwal`_).
```
Remi-Gau, line 53: is it just a change or actually a fix?

man-shu: I was a bit confused about that, too.

But here's my thinking behind tagging it as a change:
The default behavior is still what was initially reported. We are just giving the user an option to copy the header if they want.

Remi-Gau: OK makes sense: I got confused because it was closing an issue filed as bug, so I would have seen this as a fix.

that's fine then.

Remi-Gau:
@bthirion had already approved so I will got ahead and merge this.

PR #4383, opened 2024-04-16
[FIX] skip problematic plotting in doc build

Remi-Gau:
skip plotting that makes doc build fails to not hold back other pull requests


man-shu (approved):
LGTM

Remi-Gau:
merging but keeping all the related issues opened

Issue #2744, opened 2021-03-29
Split CircleCI cache into smaller files


jeromedockes:
in circleci nilearn chaches the whole `nilearn_data` folder in one huge cache.
Quite frequently the cache is missed.
When I discussed it with circleci support they could not find exactly what was the problem but they suggested that the huge cache size might be an issue.
We could try splitting the cache into smaller chunks (eg one for each dataset).
As @emdupre pointed out it is what MNE seems to do: https://github.com/mne-tools/mne-python/blob/02bb98b35a34d49d22037e3d494ba1e4f83994de/.circleci/config.yml#L255

NicolasGensollen:
Thanks for opening this @jeromedockes !
I'd be happy to work on this once #2734 is merged.

jeromedockes:
great! are you waiting for another review on #2734 ? If it helps I can look at it today

NicolasGensollen:
>are you waiting for another review on #2734 ?  If it helps I can look at it today

Absolutely! It'd be great to have another review on this PR!

PR #4390, opened 2024-04-17
[MAINT] pre-download data on full doc build

### LINKED ISSUE #4315: Optimize masker and GLM reports in the doc build
1. Generating all the maskers / GLM reports takes at least 4 minutes (more if we include the data download time)
 
- there should be a way to speed this up.

2. Sometimes the documentation build will fail because the maskers and GLM reports cannot be built.

https://github.com/nilearn/nilearn/actions/runs/8194633995/job/22411144135#step:34:122

This seems to because the script that builds them needs to download datasets, atlases... And this may fail (same reason as to why we mock things for the fetchers in the test suite).

To minimize how often that happens we could:
- only build those reports when a full build of the doc is done (on `main` or for a commit with `[full doc]` in the message
- rely on the caching of the data between builds (should be doable given that the data is already cached in the build_doc workflow)

### LINKED ISSUE #4367: make build release doc workflow more robust
see https://github.com/nilearn/nilearn/actions/workflows/release-docs.yml

seems a lot of the failures are related to failures to get the data

https://github.com/nilearn/nilearn/actions/runs/8613300789/job/23609732291

### LINKED ISSUE #4389: doc build fails on data download
I think this is mostly a data availibility issue but the fact that the data download happens during the doc build and not in one step before makes is hard to debug or relaunch the worklow.

Will try to see if I can separate the data download in a previous step.

Remi-Gau:
- Closes #4389
- Closes #4367
- Closes #4315

- getting data and doc builds are run by separate job: this should help debugging CI because we can quickly know if the run failed because of download or an error in the doc build itself
- get_data job
  - the data got by this job is transferred to the build doc by caching it across jobs
  - can access data downloaded on previous attempts of a run of this workflow
  - uses a python script to get the data:
    - for any build this script will get the data required for building the reports page
    - for full build it will get all the data
  - cache will be updated every month instead of every week: I do not think we should expect data to change that often
  - previous rules around caching are unchanged (partial build will not rely on caching from previous runs, full build do, "force download" is still an option that will ignore cache)

Also silences one warning in the doc build:

- See this issue : https://github.com/sphinx-doc/sphinx/issues/12300
- This will be fixed according to https://github.com/sphinx-doc/sphinx/issues/12300#issuecomment-2062238457 by the PR https://github.com/nilearn/nilearn/pull/4390

--- 

**TODO**

- [x] add doc in the new python script
- [x] apply same logic to release doc workflow - test after merge

To test:

Clean cache before testing each of those

- [x] force download --> do not rely on cache.
- [x] partials builds --> do not rely on cache 
- [x] full builds --> rely on cache
- [x] cancelling a run and restarting can succesfully get the cache from the cancelled run


doc/Makefile
```diff
@@ -79,6 +79,10 @@ force:
 view: _build/html/index.html
 	$(BROWSER) _build/html/index.html
 
+data:
+	# downloads all the data for the examples
+	python get_data_examples.py
+
```
Remi-Gau, line 85:
```suggestion
```

doc/conf.py
```diff
@@ -162,7 +162,7 @@
 # modindex_common_prefix = []
 
 # A list of warning types to suppress arbitrary warning messages
-suppress_warnings = ["image.not_readable"]
+suppress_warnings = ["image.not_readable", "config.cache"]
```
Remi-Gau, line 165: unrelated to the main topic of this PR but is suggested workaround for some new sphinx warning

Remi-Gau:
maybe worth checking if reverting the many small cache is an idea we still want: https://github.com/nilearn/nilearn/issues/2744

this was implemented when we used circle ci for the doc build

build_tools/github/determine_restore_data.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Check if gtihub cache can be used
+#
+# If the commit message contains "force download" then cache is ignored
+# and all data will be downloaded.
+#
+# If the build type is a no plot in CI, then cache is not used.
+
+set -e
+set -x
```
Remi-Gau, line 11:
```suggestion
set -e -x
```

Remi-Gau:
@bthirion @man-shu 
This should be ready to go.
Let's see if it all comes green.

Probably a good thing if I explain to one (or both of you) in a video call what's now going on doc build front and we update the doc accordingly to make sure it is clear to everyone.

man-shu:
It does pass, indeed! I can meet today until 6 or after 7pm.

Remi-Gau:
> It does pass, indeed! I can meet today until 6 or after 7pm.

that's a wee late. Let's do tomorrow morning?

bthirion (approved):
LGTM, thx !

Remi-Gau:
OK reviewed with @man-shu 

merging

PR #4397, opened 2024-04-24
Copy input image header to the output by default

### LINKED ISSUE #4393: [BUG] Several functions under the `image` module do not carry over the header information to the output

### Operating system

- [ ] Linux
- [X] Mac
- [ ] Windows

### Operating system version

Mac OS Version 14.4.1 "sonoma"

### Python version

- [X] 3.12
- [ ] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

0.10.4

### Expected behavior

This issue was initially reported in #2645 for `math_img` and was fixed in #4337 with a `copy_header_from` parameter. The usage is explained in detail in #4392.

I have identified other functions with the same issue since then:
- [x] `crop_img`
- [x] `mean_img`
- [x] `threshold_img`
- [x] `binarize_img`
- [x] `resample_img`
- [x] `reorder_img`
- [ ] ~remove the check for number of dimensions in `math_img` when `copy_header_from` is not None, if `new_img_like` can handle it (explained below)~

The fix could be simple - all of these functions use `new_img_like` to create the output nifti object with the `copy_header` parameter set to False, which can just be changed to True.

However, a common thing in all these functions is that they all somehow manipulate the data values/affine/dimensions of the input, which in turn reflects in the header info of the result image. So this makes copying the header a bit more complicated depending on what the function does.

~As shown below for `new_img_like`, the header copying from `new_img_like` does not handle these cases well and returns wrong headers. So I think it would be worthwhile to update `new_img_like` first.~

EDIT: As it turns out, I wasn't passing the affine into `new_img_like` below. I have updated the code now. Once I do this, it does seem to give the correct TR and voxel dimensions.

So `new_img_like` seems to handle well the changes in affine at least for `resample_img`. I will go over each function and use case now.

```python
# header of the input_img
print(input_img.header)
# note the dim and pixdim values
```
```
<class 'nibabel.nifti1.Nifti1Header'> object, endian='<'
sizeof_hdr      : 348
data_type       : b''
db_name         : b''
extents         : 0
session_error   : 0
regular         : b'r'
dim_info        : 48
dim             : [  4  61  73  61 176   1   1   1]
intent_p1       : 0.0
intent_p2       : 0.0
intent_p3       : 0.0
intent_code     : none
datatype        : float32
bitpix          : 32
slice_start     : 0
pixdim          : [-1.  3.  3.  3.  2.  0.  0.  0.]
vox_offset      : 0.0
scl_slope       : nan
scl_inter       : nan
slice_end       : 60
slice_code      : unknown
xyzt_units      : 10
cal_max         : 0.0
cal_min         : 0.0
slice_duration  : 0.0
toffset         : 0.0
glmax           : 0
glmin           : 0
descrip         : b''
aux_file        : b''
qform_code      : scanner
sform_code      : scanner
quatern_b       : -0.0
quatern_c       : 1.0
quatern_d       : 0.0
qoffset_x       : 90.0
qoffset_y       : -126.0
qoffset_z       : -72.0
srow_x          : [-3. -0. -0. 90.]
srow_y          : [  -0.    3.   -0. -126.]
srow_z          : [  0.   0.   3. -72.]
intent_name     : b''
magic           : b'n+1'
```

```python
# %% header of the result_img with default header
print(result_img.header)
# has correct pixdim[1:4] but wrong TR (pixdim[4])
# Several other fields don't match the input either
```
```
<class 'nibabel.nifti1.Nifti1Header'> object, endian='<'
sizeof_hdr      : 348
data_type       : b''
db_name         : b''
extents         : 0
session_error   : 0
regular         : b''
dim_info        : 0
dim             : [  4  31  37  31 176   1   1   1]
intent_p1       : 0.0
intent_p2       : 0.0
intent_p3       : 0.0
intent_code     : none
datatype        : float32
bitpix          : 32
slice_start     : 0
pixdim          : [1. 6. 6. 6. 1. 1. 1. 1.]
vox_offset      : 0.0
scl_slope       : nan
scl_inter       : nan
slice_end       : 0
slice_code      : unknown
xyzt_units      : 0
cal_max         : 0.0
cal_min         : 0.0
slice_duration  : 0.0
toffset         : 0.0
glmax           : 0
glmin           : 0
descrip         : b''
aux_file        : b''
qform_code      : unknown
sform_code      : aligned
quatern_b       : 0.0
quatern_c       : 0.0
quatern_d       : 0.0
qoffset_x       : -90.0
qoffset_y       : -126.0
qoffset_z       : -72.0
srow_x          : [  6.   0.   0. -90.]
srow_y          : [   0.    6.    0. -126.]
srow_z          : [  0.   0.   6. -72.]
intent_name     : b''
magic           : b'n+1'
```

```python
# %% header copied via new_img_like
print(result_img_with_header.header)
# has correct pixdim[1:4] and TR (pixdim[4])
# All other important fields are appropriately updated too
```
```
<class 'nibabel.nifti1.Nifti1Header'> object, endian='<'
sizeof_hdr      : 348
data_type       : b''
db_name         : b''
extents         : 0
session_error   : 0
regular         : b'r'
dim_info        : 48
dim             : [  4  31  37  31 176   1   1   1]
intent_p1       : 0.0
intent_p2       : 0.0
intent_p3       : 0.0
intent_code     : none
datatype        : float32
bitpix          : 32
slice_start     : 0
pixdim          : [1. 6. 6. 6. 2. 1. 1. 1.]
vox_offset      : 0.0
scl_slope       : nan
scl_inter       : nan
slice_end       : 60
slice_code      : unknown
xyzt_units      : 10
cal_max         : 18028.748
cal_min         : 0.0
slice_duration  : 0.0
toffset         : 0.0
glmax           : 0
glmin           : 0
descrip         : b''
aux_file        : b''
qform_code      : unknown
sform_code      : aligned
quatern_b       : 0.0
quatern_c       : 0.0
quatern_d       : 0.0
qoffset_x       : -90.0
qoffset_y       : -126.0
qoffset_z       : -72.0
srow_x          : [  6.   0.   0. -90.]
srow_y          : [   0.    6.    0. -126.]
srow_z          : [  0.   0.   6. -72.]
intent_name     : b''
magic           : b'n+1'
```

```python
# code to see above output

# %% import stuff
from nilearn import datasets
from nilearn.image import load_img, resample_img, new_img_like
import numpy as np

# %% get a sample dataset
dataset = datasets.fetch_adhd(n_subjects=1)
input_img = load_img(dataset.func[0])
# header of the input_img
print(input_img.header)
# note the dim and pixdim values

# %% downsample to 6mm isotropic,
# we expect the pixdim[1:4] to change to 6
# rest should remain the same
result_img = resample_img(input_img, target_affine=np.diag((6, 6, 6)))

# %% header of the result_img with default header
print(result_img.header)
# has correct pixdim[1:4] but wrong TR (pixdim[4])
# Several other fields don't match the input either

# %% explicitly copy the header with new_img_like from input_img to result_img
result_img_with_header = new_img_like(
    ref_niimg=input_img,
    data=result_img.get_fdata(),
    affine=result_img.affine,
    copy_header=True,
)

# %% header copied via new_img_like
print(result_img_with_header.header)
# has correct pixdim[1:4] and TR (pixdim[4])
# All other important fields are appropriately updated too
```

man-shu:
- Closes #4393 and #4426 

The following functions under `nilearn.image` have been updated to copy image header by default
- [x] `crop_img`
- [x] `mean_img`
- [x] `threshold_img`
- [x] `binarize_img`
- [x] `resample_img`
- [x] `reorder_img`
- [x] test for the header equality between input and output
- [ ] ~remove the check for number of dimensions in `math_img` when `copy_header_from` is not None, if `new_img_like` can handle it (explained in #4393)~
- [x] allow users to control this header copying
- [x] warn users of the change that headers will be copied by default from v0.13.0 onwards


man-shu:
The fix was indeed the one proposed in #4393.

So now the default behavior for all functions under `nilearn.image` module (except `math_img`) is to directly copy the input image's header info to the output image in whichever fields possible and update the rest, via `new_img_like`.

As for `math_img`, it was the most peculiar case and I think the choices made in #4337 should stay in place.

nilearn/image/tests/conftest.py
```diff
@@ -53,3 +55,28 @@ def img_4d_mni_tr2(img_4d_mni):
     header = img.header.copy()
     header["pixdim"][4] = 2.0
     return Nifti1Image(img.get_fdata(), img.affine, header=header)
+
+
+def match_headers_keys(source, target, except_keys):
```
man-shu, line 60: Is it ok to have functions defined in `conftest.py` @Remi-Gau ?

I use this in all the tests to check header fields.

bthirion (commented):
The change LGTM but you must at least make it visible. 
I'm wondering whether we should go through deprecation cycles...

nilearn/image/tests/conftest.py
Remi-Gau: Just double checked and the pattern we have followed so far is:
- fixtures or functions used for building fixtures go into `conftest.py` files
- functions used across several tests modules either go into a `_testing.py` modules for example (`nilearn/datasets/tests/_testing.py` or `nilearn/_utils/testing.py`

So maybe just move those into a `nilearn/image/tests/_testing.py` ?

Remi-Gau (changes requested):
See comment: https://github.com/nilearn/nilearn/pull/4397/files#r1579266563

Remi-Gau:
> I'm wondering whether we should go through deprecation cycles...

yeah I am wondering about this as well, because some of this may induce behavior changes in the code.

nilearn/image/tests/conftest.py
man-shu: Ok makes sense. Will do. Thanks!

emdupre:
Personally I would be +1 deprecation cycles ! This is a noticeable enough change (and I don't know that the previous before was _wrong_ ) that I think we should alert users.

Remi-Gau:
> Personally I would be +1 deprecation cycles ! This is a noticeable enough change (and I don't know that the previous before was _wrong_ ) that I think we should alert users.

Good with me to err on the side of caution and let our users know.

bthirion (approved):
LGTM, thx.

man-shu:
It was decided in the core-dev meeting that we'll open another PR that would warn users of this change in release 0.13.0.

So I've created one #4427

nilearn/image/image.py
```diff
@@ -984,7 +986,7 @@ def threshold_img(
         img_data = img_data[:, :, :, 0]
 
     # Reconstitute img object
-    thresholded_img = new_img_like(img, img_data, affine)
+    thresholded_img = new_img_like(img, img_data, affine, copy_header=True)
```
Remi-Gau, line 989: Given that this will go through a deprecation cycle for this you want to leave the current behavior unchanged and let user opt-in until we finally deprecate it.

So you need to change the function signature and the doc string too:

```python
def threshold_img(
    img,
    threshold,
    cluster_threshold=0,
    two_sided=True,
    mask_img=None,
    copy=True,
    copy_header=False # add it here
)
"""
updated doc string with description of the new parameter.

    copy_header: :obj:`bool`, default = False

        explain what this does

        .. versionadded:: 0.8.1

"""

...

   # just pass the value here
   thresholded_img = new_img_like(img, img_data, affine, copy_header=copy_header)
```

Remi-Gau: Only commenting on this function but this should apply to all pubic functions that gets changed

man-shu: Ok, that makes sense. I was thinking of doing that in #4427.

So IIUC, in release 0.13.0, we'll simply set the default to True and remove the warnings, right?

Remi-Gau: > So IIUC, in release 0.13.0, we'll simply set the default to True and remove the warnings, right?

yes exactly 👍🏾

nilearn/image/image.py
```diff
@@ -376,6 +394,22 @@ def crop_img(img, rtol=1e-8, copy=True, pad=True, return_offset=False):
...
+            "Currently, the header is reset to the default Nifti1Header."
+            "To suppress this warning and continue using the current behavior "
+            "set `copy_header=False`. To use the new behavior, set "
+            "`copy_header=True`."
+        )
+        warnings.warn(
+            category=FutureWarning,
+            message=copy_header_default,
+        )
+        copy_header = False
```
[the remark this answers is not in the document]
Remi-Gau: That comment goes for the other functions as well.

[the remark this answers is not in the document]
Remi-Gau: if you find it easier, you can also create a single utility function that checks the value of `copy_header` and throws a warning.

If you do may put it in thre : `nilearn\_utils\helpers.py` where there are a few other deprecation related utilities.

Also thinking that it may make it easier when future us (or future maintainers) have to deal with this deprecation.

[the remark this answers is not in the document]
man-shu: Indeed, I was wondering if we already have a system to track such deprecations.

In addition to the hassle of removing every single warning message and replacing every parameter later on, it is possible that we might just miss/forget about some deprecations when the said release is made.

nilearn/_utils/helpers.py
```diff
@@ -126,19 +130,24 @@ def remove_parameters(removed_params,
...
+                    f"will be removed in version {end_version}; "
+                    f"{reason}"
+                )
+                warnings.warn(
+                    category=DeprecationWarning, message=message, stacklevel=3
+                )
             return func(*args, **kwargs)
+
         return wrapper
+
```
man-shu, line 150: Sorry for the messy diff. Looks like this module wasn't yet blackified.
All changes above this are just blackification.

nilearn/image/tests/conftest.py
```diff
@@ -49,7 +49,9 @@ def img_4d_mni_tr2(img_4d_mni):
     img = image.new_img_like(
         img_4d_mni, data=img_4d_mni.get_fdata(), copy_header=True
     )
-    # Change the TR
+    # Add fake description
     header = img.header.copy()
+    header["descrip"] = b"Fake description"
```
man-shu, line 54: Added this because the `test_mean_img_copied_header` was still passing with `copy_header=False`

nilearn/_utils/helpers.py
Remi-Gau: that's fair: I actually closed the very old PR where this file was also involved and why we were not letting black work on it

can I let you update the black and pre-commit config to let black validate this file in CI:

just remove it from the exclude 
- here: https://github.com/nilearn/nilearn/blob/e384ebc7c5a5a080b476efdb4a9aa8e9af57eff0/.pre-commit-config.yaml#L75
- and here: https://github.com/nilearn/nilearn/blob/e384ebc7c5a5a080b476efdb4a9aa8e9af57eff0/pyproject.toml#L108

nilearn/image/image.py
```diff
@@ -306,6 +310,14 @@ def _crop_img_to(img, slices, copy=True):
     copy : :obj:`bool`, default=True
         Specifies whether cropped data is to be copied or not.
 
+    copy_header : :obj:`bool`, default=None
```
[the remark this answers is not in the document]
Remi-Gau: Once again this may apply to other function definitions.

Remi-Gau (changes requested):
Can you add a test to make sure that the deprecation warning is thrown properly when users pass the value False?

man-shu:
Re: Setting `copy_header=True` to avoid warnings in tests and examples:

There are quite a few tests that throw this warning:
Screenshot 2024-05-31 at 5 26 30 PM
Potentially, several examples, too.

Will update those now.

nilearn/image/tests/test_image.py
```diff
@@ -938,6 +996,40 @@ def test_binarize_negative_img(img_4d_rand_eye):
...
+        img_4d_mni_tr2,
+        except_keys=["cal_min", "cal_max"],
+    )
+    # min value should be 0 in the result
+    assert result.header["cal_min"] == 0
+    # max value should be 1 in the result
+    assert result.header["cal_max"] == 1
+
+
+@pytest.mark.parametrize(
```
man-shu, line 1014: @Remi-Gau, here I test the warning being thrown for `copy_header=False` (for all `image.py` functions)

nilearn/image/tests/test_resampling.py
```diff
@@ -840,6 +869,31 @@ def test_reorder_img_mirror():
...
+    result = reorder_img(img_4d_mni_tr2, copy_header=True)
+    # all header fields should stay the same
+    match_headers_keys(
+        img_4d_mni_tr2,
+        result,
+        except_keys=[],
+    )
+
+
+@pytest.mark.parametrize(
```
man-shu, line 883: and here, the warning for `resampling.py` functions

nilearn/image/tests/test_image.py
man-shu: Just so you dont have to review everything again

examples/00_tutorials/plot_decoding_tutorial.py
```diff
@@ -52,7 +52,7 @@
 from nilearn import plotting
 from nilearn.image import mean_img
 
-plotting.view_img(mean_img(fmri_filename), threshold=None)
+plotting.view_img(mean_img(fmri_filename, copy_header=True), threshold=None)
```
man-shu, line 55: I have added the `copy_header=True` to all examples and also in the codebase (the test warnings were mostly because of that).

But now I am not so sure about this because we are anyway going to make it the default in 0.13.0.

Let me know what you think @Remi-Gau, @bthirion.

man-shu:
Now we are ready for final reviews @Remi-Gau and @bthirion

Remi-Gau (approved):
Let's see what CI says but it looks good to me.

Remi-Gau:
Everything looks green. You can merge at will @man-shu

PR #4406, opened 2024-04-26
[REF] refactor style checks in CI

Remi-Gau:
- Closes #

-
-


.pre-commit-config.yaml
```diff
@@ -24,34 +25,41 @@ repos:
...
     rev: 5.13.2
     hooks:
     -   id: isort
         args: [--profile, black, --settings-path, pyproject.toml]
 
--   repo: https://github.com/ikamensh/flynt/
-    rev: 1.0.1
-    hooks:
-    -   id: flynt
-
```
Remi-Gau, line 37: removed as flake8-fstrings also performs that checks

Remi-Gau:
@bthirion @man-shu

This PR is mostly a suggestion aimed to reduce the number of workflows we have to run and maintain.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -283,9 +283,10 @@ def _sanitize_labels(self, labels):
                     stacklevel=3,
                 )
             if not all(isinstance(x, str) for x in labels):
+                types_labels = {type(x) for x in labels}
```
man-shu, line 286: Is this related?

.github/workflows/README.md
```diff
@@ -78,10 +75,13 @@ To run a full build and download all datasets, you would then combine both tags:
...
 ```
 
-#### Hosting of the docs
+## Building the stable release documentation
 
-Upon a successful build of the doc, it is zipped and uploaded as an artifact.
-A circle-ci workflow is then triggered. See below.
+### release-docs.yml
+
+Should be triggered automatically after merging and tagging a release PR to
```
man-shu, line 82: We should make a note to check that on the next release. Because I think we had to run it manually last time.

man-shu (approved):
Looks good otherwise!

nilearn/maskers/nifti_labels_masker.py
Remi-Gau: Actually yes. Indirectly.

the pyupgrade hook was complaining about this line: https://github.com/nilearn/nilearn/actions/runs/8844898238/job/24287692390#step:5:470

because we can use generator for this

`set([type(x) for x in labels])` --> `{type(x) for x in labels}`

I extracted it out of the f string because a double curly braces is used as an escape character

```python
>>> f"{{type(x) for x in ['foo', 1, 2.0]}}"
"{type(x) for x in ['foo', 1, 2.0]}"

>>> a={type(x) for x in ['foo', 1, 2.0]}
>>> f"{a}"
"{<class 'float'>, <class 'int'>, <class 'str'>}"
```

.github/workflows/README.md
Remi-Gau: Yeah to be fair, we'll know if it works at the next release. :-p
Though as it is a manual workflow we could also test it now.

bthirion (approved):
LGTM, thx.

Remi-Gau:
OK merging this as the other failures are unrelated

let's see in practice if the feed back we get from failing linting are good enough for us and contributors to know what to fix.

PR #4414, opened 2024-05-16
[FIX] Raise error if unknown kwargs are given to `first_level_from_bids`

### LINKED ISSUE #4357: [ENH] Throw error when extra kwargs are given to `glm.first_level.first_level_from_bids`

`nilearn.glm.first_level.first_level_from_bids` takes kwargs, and passes kwargs that have the `confounds_` prefix to another function. However, the remaining kwargs are not checked and nothing happens if for example the confound kwargs are not given correctly (e.g. with typo `confound_` instead of `confounds_`).

I think it would be nice/useful to throw a runtime error or something when that happens (first mentioned in PR #4352). Would be happy to work on this too.

### Benefits to the change

Better debugging when invalid/unknown kwargs are given.

_No response_

michellewang:
- Closes #4357 

- Make `_check_kwargs_load_confounds` return unused `remaining_kwargs` in addition to `kwargs_load_confounds`
- Throw an error in `first_level_from_bids` if there are any remaining kwargs after the confound ones are extracted
  - Not sure if this is too strong, we could also raise a warning instead

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -1292,7 +1292,14 @@ def first_level_from_bids(
 
     dataset_path = Path(dataset_path).absolute()
 
-    kwargs_load_confounds = _check_kwargs_load_confounds(**kwargs)
+    kwargs_load_confounds, remaining_kwargs = _check_kwargs_load_confounds(
+        **kwargs
+    )
+
+    if len(remaining_kwargs) > 0:
+        raise RuntimeError(
+            f"Some keyword arguments were unused: {remaining_kwargs}"
+        )
 
     if drift_model is not None and kwargs_load_confounds is not None:
         if "high_pass" in kwargs_load_confounds.get("strategy"):
@@ -1903,18 +1910,18 @@ def _check_kwargs_load_confounds(**kwargs):
     }
 
     if kwargs.get("confounds_strategy") is None:
-        return None
+        return None, kwargs
 
-    kwargs_load_confounds = {
-        key: (
-            defaults[key]
-            if f"confounds_{key}" not in kwargs
-            else kwargs[f"confounds_{key}"]
-        )
-        for key in defaults
-    }
+    remaining_kwargs = kwargs.copy()
+    kwargs_load_confounds = {}
+    for key in defaults:
+        confounds_key = f"confounds_{key}"
+        if confounds_key in kwargs:
+            kwargs_load_confounds[key] = remaining_kwargs.pop(confounds_key)
+        else:
+            kwargs_load_confounds[key] = defaults[key]
 
-    return kwargs_load_confounds
+    return kwargs_load_confounds, remaining_kwargs
 
 
 def _make_bids_files_filter(
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -1298,7 +1298,8 @@ def first_level_from_bids(
 
     if len(remaining_kwargs) > 0:
         raise RuntimeError(
-            f"Unknown keyword arguments. Keyword arguments should start with `confounds_` prefix: {remaining_kwargs}"
+            "Unknown keyword arguments. Keyword arguments should start with "
+            f"`confounds_` prefix: {remaining_kwargs}"
         )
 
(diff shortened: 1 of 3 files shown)


michellewang:
@man-shu it is now!

nilearn/glm/first_level/first_level.py
```diff
@@ -1292,7 +1292,14 @@ def first_level_from_bids(
...
     dataset_path = Path(dataset_path).absolute()
 
-    kwargs_load_confounds = _check_kwargs_load_confounds(**kwargs)
+    kwargs_load_confounds, remaining_kwargs = _check_kwargs_load_confounds(
+        **kwargs
+    )
+
+    if len(remaining_kwargs) > 0:
+        raise RuntimeError(
+            f"Some keyword arguments were unused: {remaining_kwargs}"
```
man-shu, line 1301: I think the error message could be made more specifc that something is wrong with confounds kwargs. For example:

```suggestion
            f"Unknown keyword arguments. Keyword arguments should start with `confounds_` prefix: {remaining_kwargs}"
```

nilearn/glm/tests/test_first_level.py
```diff
@@ -1958,6 +1958,23 @@ def test_first_level_from_bids_no_subject(tmp_path):
...
 
+def test_first_level_from_bids_unused_kwargs(tmp_path):
+    """Check that unused kwargs are properly handled."""
+    bids_path = create_fake_bids_dataset(
+        base_dir=tmp_path, n_sub=1, n_ses=1, tasks=["main"], n_runs=[2]
+    )
+    with pytest.raises(
+        RuntimeError, match="Some keyword arguments were unused:"
+    ):
+        first_level_from_bids(
```
man-shu, line 1969: It's probably useful to mention the missing "s". Not that obvious at first glance
```suggestion
        # wrong kwarg name `confound_strategy` (wrong)
        # instead of `confounds_strategy` (correct)
        first_level_from_bids(
```

man-shu (commented):
LGTM otherwise. Thanks!

man-shu (approved):

nilearn/glm/tests/test_first_level.py
```diff
@@ -1958,6 +1958,25 @@ def test_first_level_from_bids_no_subject(tmp_path):
         )
 
 
+def test_first_level_from_bids_unused_kwargs(tmp_path):
+    """Check that unused kwargs are properly handled."""
+    bids_path = create_fake_bids_dataset(
+        base_dir=tmp_path, n_sub=1, n_ses=1, tasks=["main"], n_runs=[2]
+    )
+    with pytest.raises(
+        RuntimeError, match="Some keyword arguments were unused:"
```
man-shu, line 1967: Just realised this would have to be updated as well
```suggestion
        RuntimeError, match="Unknown keyword arguments"
```

nilearn/glm/first_level/first_level.py
```diff
@@ -1903,18 +1911,18 @@ def _check_kwargs_load_confounds(**kwargs):
...
-        key: (
-            defaults[key]
-            if f"confounds_{key}" not in kwargs
-            else kwargs[f"confounds_{key}"]
-        )
-        for key in defaults
-    }
+        return None, kwargs
+
+    remaining_kwargs = kwargs.copy()
```
bthirion, line 1916: 'remaining_kwargs' should include only kwargs without the 'confounds' string, right ?

bthirion (commented):
Thx. 
I'm not sure I undertsand the behavior of the function. Please correct me if I'm wrong.

nilearn/glm/first_level/first_level.py
michellewang: Yes:
- If the kwarg is `confounds_`-something, then we effectively strip the `confounds_` prefix and replace the previously defined default value. This gets added to `kwargs_load_confounds`, to be passed to another function
-  Otherwise, the kwarg isn't removed and stays in `remaining_kwargs`. The reason why I am returning `remaining_kwargs` instead of directly raising an error is just in case we want to support other kwargs in the future

bthirion (approved):
LGTM, thx.

PR #4419, opened 2024-05-22
[FIX] update to new colormap api

### LINKED ISSUE #4420: [BUG] get_cmap removed in matplotlib 3.9.0

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

N/A

### Python version

- [ ] 3.12
- [ ] 3.11
- [X] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

N/A

### Expected behavior

N/A

This is what I got:

```bash
`AttributeError: module 'matplotlib.cm' has no attribute 'get_cmap'`

```

```python
# Paste your code here


```

jdkent:
- Closes #4420 

- convert `get_cmap` to `colormaps` (`get_cmap` removed in [matplotlib 3.9.0](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.9.0.html#removals))
- bump base matplotlib version to 3.6.0 (introduced [resampled](https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.6.0.html#colormap-method-for-creating-a-different-lookup-table-size))

### DIFF
--- a/nilearn/plotting/html_stat_map.py
+++ b/nilearn/plotting/html_stat_map.py
@@ -342,7 +342,7 @@ def _json_view_size(params, width_view=600):
 def _get_bg_mask_and_cmap(bg_img, black_bg):
     """Get background data for _json_view_data."""
     bg_mask = np.ma.getmaskarray(get_data(bg_img))
-    bg_cmap = copy.copy(matplotlib.pyplot.get_cmap("gray"))
+    bg_cmap = copy.copy(matplotlib.colormaps["gray"])
     if black_bg:
         bg_cmap.set_bad("black")
     else:
--- a/nilearn/plotting/html_stat_map.py
+++ b/nilearn/plotting/html_stat_map.py
@@ -1,13 +1,12 @@
 """Visualizing 3D stat maps in a Brainsprite viewer."""
 
-import copy
 import json
 import os
 import warnings
 from base64 import b64encode
 from io import BytesIO
 
-import matplotlib
+import matplotlib as mpl
 import numpy as np
 from matplotlib.image import imsave
 from nibabel.affines import apply_affine
@@ -342,7 +341,7 @@ def _json_view_size(params, width_view=600):
 def _get_bg_mask_and_cmap(bg_img, black_bg):
     """Get background data for _json_view_data."""
     bg_mask = np.ma.getmaskarray(get_data(bg_img))
-    bg_cmap = copy.copy(matplotlib.colormaps["gray"])
+    bg_cmap = mpl.colormaps["gray"]
     if black_bg:
         bg_cmap.set_bad("black")
     else:
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -17,7 +17,7 @@ import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib import gridspec as mgs
-from matplotlib.colors import LinearSegmentedColormap, ListedColormap
+from matplotlib.colors import LinearSegmentedColormap
 from nibabel.spatialimages import SpatialImage
 from scipy import stats
 from scipy.ndimage import binary_fill_holes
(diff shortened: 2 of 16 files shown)


bthirion (commented):
The change LGTM. The version bump for mpl is quite large. Is that OK with other people ?

jdkent:
it is a pretty big jump, and I think there are a still a few API idiosyncracies to take care of, another option is to have to `try` `except` statements where relevant to support the old API usage and the new API usage.

Remi-Gau:
> The change LGTM. The version bump for mpl is quite large. Is that OK with other people ?

I think that it makes sense to bump to 3.6: it is the oldest version that is still supported:
https://scientific-python.org/specs/spec-0000/#support-window

It is even recommended to plan to drop it before the end of the year:
https://scientific-python.org/specs/spec-0000/#2024---quarter-3

bthirion:
Thank you. I was not ware of that...

bthirion (approved):
Thx.

Remi-Gau:
Moving to draft until we get all the tests to pass: just to avoid a premature merge.

nilearn/plotting/html_stat_map.py
```diff
@@ -342,7 +342,7 @@ def _json_view_size(params, width_view=600):
 def _get_bg_mask_and_cmap(bg_img, black_bg):
     """Get background data for _json_view_data."""
     bg_mask = np.ma.getmaskarray(get_data(bg_img))
-    bg_cmap = copy.copy(matplotlib.pyplot.get_cmap("gray"))
+    bg_cmap = copy.copy(matplotlib.colormaps["gray"])
```
Remi-Gau, line 345: don't think the copy is needed any more as from the matplotlib this should return a copy in any case

nilearn/plotting/img_plotting.py
```diff
@@ -807,7 +807,7 @@ def _plot_roi_contours(display, roi_img, cmap, alpha, linewidths):
     roi_img = _utils.check_niimg_3d(roi_img)
     roi_data = get_data(roi_img)
     labels = np.unique(roi_data)
-    cmap = plt.get_cmap(cmap)
+    cmap = plt.cm.ColormapRegistry.get_cmap(cmap)
```
Remi-Gau, line 810: FYI: this won't work as you need to instantiate `plt.cm.ColormapRegistry` first.

the matplotlib doc suggests that in general you should not need to do this.

Remi-Gau:
> it is a pretty big jump, and I think there are a still a few API idiosyncracies to take care of, another option is to have to `try` `except` statements where relevant to support the old API usage and the new API usage.

I think that the tests should pass now 🤞🏾 

Let's see if we need `try` blocks.

bthirion (commented):
LGTM. The only pending question is whether we still stick to skelarn's conventions regarding dependencies versions.

Remi-Gau:
> LGTM. The only pending question is whether we still stick to skelarn's conventions regarding dependencies versions.

One possibility if we decide against this is to have some utility function to do

```python
if matplotlib.version < 3.6:
   # get colormap this way
   ...
else:
   # do it that way
   ...
```

this should make a bit easier to refactor when the times come to actually drop support for older version of matplotlib

bthirion:
@emdupre can we conclude on the dependence we would like to have for next release in order to implement the best option.

emdupre:
I think a decision on that should be a discussion topic for the maintainers team. For this PR, I'd suggest sticking with our existing dependency schedule.

Remi-Gau:
@jdkent I am closing this one in favor of the smaller and less intrusive #4458 

See also my comment : https://github.com/nilearn/nilearn/issues/4420#issuecomment-2179183313

AFAICT [matplotlib.pyplot.get_cmap](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.get_cmap.html#matplotlib.pyplot.get_cmap) is not deprecated in matplotlib 3.9 even if matplotlib.cm.get_cmap is.

PR #4437, opened 2024-06-06
[DOC] Explaining how Decoder works 

### LINKED ISSUE #4344: Discrepancy between decoding accuracies from nilearn's `Decoder` and pure sklearn implementation

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Linux Ubuntu 22.04.4

### Python version

- [X] 3.12
- [X] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

0.10.4.dev34+g58f12585

### Expected behavior

As initially reported on neurostars [here](https://neurostars.org/t/decoding-form-beta-maps-decoder-with-leaveonegroupout/28654/10?u=man-shu), the decoding performances should be similar, if not the same, using either nilearn or sklearn. But they aren't -- especially when one does not standardize the input time series / features.

While investigating, I observed the following so far:
- ~~the time series / features (X) we get after masking (to later feed into a sklearn model) vs. after the automatic masking step within the nilearn `Decoder` are different.~~
- with nilearn `Decoder`, hyperparameter search is done by default using the "CV" versions of sklearn models. But in the docs we specify, for example, `LogisticRegression` while we actually use `LogisticRegressionCV`.
- The default parameter grids (given to `LogisticRegressionCV`) are different in nilearn and sklearn.

I am currently looking into the ReNA clustering and ANOVA feature selection steps that are done in nilearn `Decoder`. I have a feeling that they interfere with the decoding even if one switches them off via top-level `Decoder` parameters.

Accuracy scores:

Nilearn, with standardization: 0.6636432350718064
Sklearn, with standardization: 0.6693121693121693

Nilearn, without std.: **0.8784958427815571**
Sklearn, without std.: **0.6243386243386243**

```python
from sklearn.linear_model import LogisticRegressionCV
from sklearn.preprocessing import StandardScaler
from tqdm import tqdm

from nilearn import datasets
from nilearn import plotting
import numpy as np
import pandas as pd
from nilearn.image import index_img
from sklearn.model_selection import LeaveOneGroupOut
from sklearn.model_selection import cross_validate
from sklearn.pipeline import Pipeline

fmri_filename = haxby_dataset.func[0]
mask_filename = haxby_dataset.mask_vt[0]

# Load behavioral information
behavioral = pd.read_csv(haxby_dataset.session_target[0], delimiter=" ")
conditions = behavioral["labels"]
runs = behavioral["chunks"]
condition_mask = conditions.isin(
    ["bottle", "cat", "chair", "face", "house", "scissors", "shoe"]
)

# Apply mask to chunks, runs and func data
fmri_niimgs = index_img(fmri_filename, condition_mask)
conditions = conditions[condition_mask]
runs = runs[condition_mask]

# print len and shape
print(len(conditions))
print(fmri_niimgs.shape[-1])
print(len(runs))

####### NILEARN DECODING ####################################
Xs_nilearn = []
params_nilearn = []
cv = LeaveOneGroupOut()
for standardize_true_false in ["zscore_sample", False]:
    # explicitly avoid screening even though providing a small mask has the same effect
    screening_percentile = 100
    decoder_nilearn = Decoder(
        estimator="logistic_l2",
        mask=mask_filename,
        cv=cv,
        scoring="accuracy",
        standardize=standardize_true_false,
        n_jobs=10,  # parallelizing
        screening_percentile=screening_percentile,
    )
    decoder_nilearn.fit(fmri_niimgs, conditions, groups=runs)
    print(f"Decoder results with standardize={standardize_true_false}")
    print(
        "Mean decoding accuracy:",
        np.mean(list(decoder_nilearn.cv_scores_.values())),
    )
    print("Chance level:", 1.0 / len(np.unique(conditions)))

    Xs_nilearn.append(
        (
            decoder_nilearn.masker_.fit_transform(fmri_niimgs),
            standardize_true_false,
        )
    )

####### NILEARN MASKING WITH/WITHOUT STANDARDIZATION AND SKLEARN DECODING ####################################

Xs_sklearn = []
for standardize_true_false in ["zscore_sample", False]:
    # mask data before decoding
    masker = NiftiMasker(
        mask_img=mask_filename, standardize=standardize_true_false
    )
    X = masker.fit_transform(fmri_niimgs)
    y = np.array(conditions)
    groups = np.array(runs)

    logo = LeaveOneGroupOut()
    clf = LogisticRegressionCV(
        solver="liblinear",
        penalty="l2",
        Cs=np.geomspace(2e-3, 2e4, 8) * 0.5,  # same as nilearn
        random_state=42,
    )

    # Not using StandardScaler because it standaridizes across features
    # nilearn's zscore_sample standardizes across samples
    # scaler = StandardScaler()
    # if standardize_true_false:
    #     pipeline = Pipeline([("scaler", scaler), ("clf", clf)])
    # else:
    #     pipeline = Pipeline([("clf", clf)])

    decoder_sklearn = cross_validate(
        clf,
        X=X,
        y=y,
        groups=groups,
        cv=logo,
        n_jobs=8,
        scoring="accuracy",
    )

    print(
        f"Mean score in mask {mask_filename} with standardize={standardize_true_false}:",
        np.mean(decoder_sklearn["test_score"]),
    )

    X_to_append = X.copy()
    Xs_sklearn.append((X_to_append, standardize_true_false))


# check if the two standardization (or lack of them) gives similar results
for (X_nilearn, _), (X_sklearn, _) in zip(Xs_nilearn, Xs_sklearn):
    print(np.allclose(X_nilearn, X_sklearn))
# returns True
#         True
```

man-shu:
- Closes #4344 

- A new example replicating each step performed inside the Decoder object using sklearn


examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,210 @@
...
+    interpolation="nearest",
+    cmap=cmap,
+)
+ax_multi.set_yticks([])
+ax_multi.set_xlabel("Original trial sequence")
+cbar = fig.colorbar(multi_plt, ax=[ax_binary, ax_multi])
+cbar.set_ticks(np.arange(1 + len(label_multi.classes_)))
+cbar.set_ticklabels([*label_multi.classes_, "all others"])
+
+plt.show()
```
man-shu, line 147: To explain how the One-vs-Other label encoding works
binary_classes

man-shu:
Started drafting the example and thought to get some general feedback on the structure.
@bthirion, please LMK if this looks ok.
Thanks!

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,213 @@
...
+print(f"{len(np.unique(y))} labels to predict (y): {np.unique(y)}")
+print(f"fMRI data shape (X): {fmri_img.shape}")
+print(f"Runs (groups): {np.unique(run)}")
+# %%
+# Preprocess the fMRI data
+# ------------------------
+#
+# As we can see, the fMRI data is a 4D image with shape (40, 64, 64, 864).
+# Here 40x64x64 are the dimensions of the 3D brain image and 864 is the number
+# of trials, each corresponding to one of the 8 labels we selected above.
```
bthirion, line 62: 864 is only the number of time points in the fMRI data after concatenation and removal of rest time intervals. Strictly speaking this is not the number of events

bthirion (commented):
Looks good so far.

man-shu:
In [this comment](https://github.com/nilearn/nilearn/pull/4438#pullrequestreview-2103186479) @bthirion said:

> Thx !
I only have one hesitation: when we use a XXXcv estimator, what cross-validation scheme is used ? The problem is that with neuroimaging, naive cross-validation schemes may lead to bad estimates. The well-known example id LOO, which is broken by the serial dependence in the data.

man-shu:
We use the default cv scheme for all ...CV estimators in sklearn. Turns out they are different for different estimators:
- for LogisticRegressionCV it is StratifiedKFold with K=5: https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegressionCV.html
- for RidgeClassifierCV and RidgeCV it is Leave-One-Out: https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.RidgeClassifierCV.html
- for LassoCV it is KFold with K=5: https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LassoCV.html

bthirion:
OK. I guess that in any case, users are supposed to do the job manually if they want a specific cross val scheme.

man-shu:
> OK. I guess that in any case, users are supposed to do the job manually if they want a specific cross val scheme.

But there's no easy way to select the internal cross-validation scheme for the hyperparameter optimization. The cv parameter in the Decoder object specifies the cross-validation scheme for the final train-test split.

I suppose one could select the internal cross-validation via the estimator parameter in the Decoder object. But users are not even aware that such internal cross-validation exists, so why would they select it.

bthirion:
I think you're right. We also have experts users who want to control details, but precisely, they can make all the necessary steps manually.
Best,

man-shu:
Just noticed that the Decoder does not do ReNA clustering. 
Instead, more specialized FREM classifier and regressor do this. See #4488

man-shu:
> I think you're right. We also have experts users who want to control details, but precisely, they can make all the necessary steps manually. Best,

Do you think we should do something about this? For example, either letting people select the internal cross-validation scheme and/or picking the best practice?

man-shu:
This is ready for a review @bthirion! Please let me know if you have any comments?

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,324 @@
+"""Understanding :class:`nilearn.decoding.Decoder`
+==================================================
+
+Nilearn's :class:`nilearn.decoding.Decoder` object is a composite estimator
+that does several things under the hood and can hence be a bit difficult to
+understand at first.
+
+This example aims to provide a clear understanding of the
+:class:`nilearn.decoding.Decoder` object by demonstrating these steps via a
+Sklearn pipeline.
```
bthirion, line 10: Scikit-Learn

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,324 @@
...
+
+# create a copy of y_binary and manipulate it just for plotting
+y_binary_ = y_binary.copy()
+for col in range(y_binary_.shape[1]):
+    y_binary_[np.where(y_binary_[:, col] == 1), col] = col
+
+fig, (ax_binary, ax_multi) = plt.subplots(
+    2, gridspec_kw={"height_ratios": [10, 1.5]}, figsize=(12, 2)
+)
+cmap = ListedColormap(["white"] + list(plt.cm.tab10.colors)[0:8])
```
bthirion, line 128:
```suggestion
cmap = ListedColormap(["white"] + list(plt.cm.tab10.colors)[0:n_conditions])
``` where n_conditions has been set previously

**Review by bthirion — COMMENTED** [review]
LGTM overall.
Deserves a whatsnew entry

File: examples/02_decoding/plot_haxby_understand_decoder.py [hunk 5]
```diff
@@ -0,0 +1,326 @@
...
+print(f"{n_labels} labels to predict (y): {np.unique(y)}")
+print(f"fMRI data shape (X): {fmri_img.shape}")
+print(f"Runs (groups): {np.unique(run)}")
+# %%
+# Preprocessing
+# -------------
+#
+# As we can see, the fMRI data is a 4D image with shape (40, 64, 64, 864).
+# Here 40x64x64 are the dimensions of the 3D brain image and 864 is the number
+# of selected trials, each corresponding to one of the 8 labels we selected
```
bthirion, line 64: More exactly,  it corresponds to the number of brain volumes acquired while visual stimuli were presented.

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,326 @@
...
+# done via the ``adjust_screening_percentile`` function.
+#
+# So let's define a feature selector for later use in our Scikit-Learn decoding
+# pipeline.
+
+from nilearn._utils.param_validation import adjust_screening_percentile
+from nilearn.image import load_img
+
+mask_vt_loaded = load_img(mask_vt)
+screen_percent = adjust_screening_percentile(20, mask_vt_loaded)
```
bthirion, line 197: Maybe print the value ?

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,326 @@
...
+# %%
+# Feature selection
+# -----------------
+#
+# After preprocessing the provided fMRI data, the
+# :class:`nilearn.decoding.Decoder` performs a univariate feature selection on
+# the voxels of the brain volume. It uses Scikit-Learn's
+# :class:`~sklearn.feature_selection.SelectPercentile` with
+# :func:`~sklearn.feature_selection.f_classif` to calculate ANOVA F-scores for
+# each voxel and to only keep the ones that have highest 20 percentile scores,
```
bthirion, line 180: But IIRC thsi si 20 percent of the MNI brain volume, right ?

bthirion (commented):
Still minor details, but close to merge.

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,330 @@
...
+# by default. This selection threshold can be changed using the
+# ``screening_percentile`` parameter.
+#
+# These 20 percentile voxels are w.r.t. the volume of the standard MNI152 brain
```
Remi-Gau, line 184:
```suggestion
# These 20 percentile voxels are with restpect to the volume of the standard MNI152 brain
```

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,330 @@
...
+
+# %%
+# As we can see, the mean AU-ROC scores from the Scikit-Learn pipeline and
+# Nilearn's :class:`nilearn.decoding.Decoder` are identical.
+#
+# The advantage of using Nilearn's :class:`nilearn.decoding.Decoder` is
+# that it does all these steps under the hood and provides a simple interface
+# to train, cross-validate and predict on new data, while also parallelizing
+# the computations to make the cross-validation faster. It also organises the
+# results in a structured way that can be easily accessed and analysed.
```
Remi-Gau, line 330: maybe link to this new example from this file: examples/00_tutorials/plot_decoding_tutorial.py

This way when people are done with the basic intro they can look deeper with this new example.

Remi-Gau (changes requested):

bthirion (commented):
LGTM otherwise

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,332 @@
...
+# After preprocessing the provided fMRI data, the
+# :class:`nilearn.decoding.Decoder` performs a univariate feature selection on
+# the voxels of the brain volume. It uses Scikit-Learn's
+# :class:`~sklearn.feature_selection.SelectPercentile` with
+# :func:`~sklearn.feature_selection.f_classif` to calculate ANOVA F-scores for
+# each voxel and to only keep the ones that have highest 20 percentile scores,
+# by default. This selection threshold can be changed using the
+# ``screening_percentile`` parameter.
+#
+# These 20 percentile voxels are with restpect to the volume of the standard
```
bthirion, line 186:
```suggestion
# These 20 percentile voxels are with respect to the volume of the standard
```

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -0,0 +1,332 @@
...
+# MNI152 brain template. Furthermore, if the provided mask image has less
+# voxels than the selected percentile, then all voxels in the mask are used.
```
bthirion, line 188:
```suggestion
# voxels than the selected percentile, all voxels in the mask are used.
```

bthirion (commented):
Minor details.

bthirion (approved):
LGTM, thx.

Remi-Gau (approved):
Good with me.
Let's merge when CI finishes.

PR #4438, opened 2024-06-06
[DOC] add ...CV in logistic estimator name

### LINKED ISSUE #4434: [DOC] Update possible `estimator` values and corresponding sklearn objects in `nilearn.decoding.Decoder` API ref

The description for the `estimator` parameter should reflect the following:
- when `estimator="logistic"` or `"logistic_l1"` or `"logistic_l2"`, `LogisticRegressionCV` is used instead of `LogisticRegression`.
- ~`estimator="ridge_regressor"` is not mentioned but is available, uses `RidgeCV`.~
- ~`estimator="svr"` is not mentioned but is available, uses `SVR(kernel="linear", max_iter=10000)`.~
- ~`estimator="lasso"` and `"lasso_regressor"` are not mentioned but are also available, using `LassoCV()`.~

**EDIT**: The regressor objects can be used with `nilearn.decoding.DecoderRegressor`. So only the first point is valid.

https://nilearn.github.io/stable/modules/generated/nilearn.decoding.Decoder.html

man-shu:
- Closes #4434 

- added a "CV"
- this was extending beyond 80 lines, so added `# noqa:E501` to skip black check


Remi-Gau (approved):

bthirion (commented):
Thx ! 
I only have one hesitation: when we use a XXXcv estimator, what cross-validation scheme is used ? The problem is that with neuroimaging, naive cross-validation schemes may lead to bad estimates. The well-known example id LOO, which is broken by the serial dependence in the data.

man-shu:
> Thx ! I only have one hesitation: when we use a XXXcv estimator, what cross-validation scheme is used ? The problem is that with neuroimaging, naive cross-validation schemes may lead to bad estimates. The well-known example id LOO, which is broken by the serial dependence in the data.

This is probably better be mentioned in #4437. Will respond to that there.

bthirion (approved):

PR #4412, opened 2024-05-13
[ENH] Add warning for future change of the default of `force_resample` to True

ajoshiusc:
- closes [2305](https://github.com/nilearn/nilearn/issues/2305).

The force_resample=False which is currently default, the code is not implemented correctly. There are several of #TODO flags in the code for force_resample=False case. Therefore, changed the default option to force_resample=True.

PR #4441, opened 2024-06-07
[MAINT] add force_resample where missing

Remi-Gau:
- Closes none but a follow up on #4412 

- explicitly pass the 'old' default to force_resample when resampling images
- add a TODO on each occurence to clean up when bumping to version > 0.13

### DIFF
--- a/nilearn/datasets/neurovault.py
+++ b/nilearn/datasets/neurovault.py
@@ -1526,11 +1526,14 @@ def _download_image_nii_file(image_info, collection, download_params):
 
         # Resample here
         print("Resampling...")
+        # TODO switch to force_resample=True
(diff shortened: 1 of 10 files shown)


Remi-Gau:
> I guess you're applying this every where you can ?

Yeah that's basically what I did: I was starting to get the new "force_resample" warning in another PR and there was no way I could do anything from the public API I was working with because the warning was thrown by Nilearn internals: kind of annoying as a user.

I think I got them all but I would not be surprised if we had to fix a few more like this.

bthirion (approved):
LGTM, thx.

Remi-Gau:
OK it seemed I missed a few.

PR #4432, opened 2024-06-03
[ENH] add report for SurfaceMasker

Remi-Gau:
- Closes none

Changes proposed in this pull request:

- add reports for SurfaceMasker and SurfaceLabelMasker
- add single HTML template for both type of maskers
- add masker report to the doc page showing all reports

---

- Blocked by https://github.com/nilearn/nilearn/pull/4431
- Currently include some code from #4431 to work: needs to be rebased after #4431 is merged

PR #4442, opened 2024-06-10
[ENH] add report to surface based GLMs

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/4126#issuecomment-1883293775

-
-


Remi-Gau:
example of empty GLM report:

image

bthirion:
Maybe not the most compelling example we could find, but it works ;-)

Remi-Gau:
Yeah the empty report is just mostly for maintenance to.help make sure all the elements are in the right place.

 I will make sure I add a report to the doc too but in the meantime you have the zip file I attached.

Remi-Gau:
I need to double check if the zip file was the correct one. 🙈

Remi-Gau:
This one should be better

image

[report.zip](https://github.com/user-attachments/files/15787244/report.zip)

bthirion:
Oops, the tests look broken.

Remi-Gau:
> Oops, the tests look broken.

- yup: mostly because this is build on top of #4432 and that tests are broken on that PR as well (mostly related to having real world examples of using a surface mask)

Remi-Gau:
This has grown too big.

Will split into several PRs.

Remi-Gau:
Blocked by #4460

examples/08_experimental/plot_localizer_new_surface_analysis_experimental.py
```diff
@@ -205,7 +205,7 @@
 
 # %%
 # Let's estimate the contrasts by iterating over them.
-from nilearn.plotting import show
+# from nilearn.plotting import show
```
Remi-Gau, line 208: revert before merge

examples/08_experimental/plot_localizer_new_surface_analysis_experimental.py
```diff
@@ -231,4 +231,14 @@
...
+# show()
+
+from pathlib import Path
+
+report = glm.generate_report(
+    contrasts,
+    title="surface-based example",
+)
+
+report.save_as_html(Path() / "localizer.html")
```
Remi-Gau, line 244: comment before merge

examples/08_experimental/plot_new_surface_bids_analysis_experimental.py
```diff
@@ -167,3 +167,11 @@
...
 
 show()
+
+from pathlib import Path
+
+report = first_level_glm.generate_report(
+    title="BIDS surface-based example",
+)
+
+report.save_as_html(Path() / "bids.html")
```
Remi-Gau, line 177: comment before merge

examples/08_experimental/plot_two_runs_model_experimental.py
```diff
@@ -0,0 +1,165 @@
+"""
+Simple example of two-runs fMRI surface based model fitting
```
Remi-Gau, line 2: Maybe more interesting as a new example rather than redoing what exist in volume space would be to compare volume and surface based GLM.

Also good for the devs to have some visuals to make sure both approaches do not give wildely different results.

nilearn/experimental/reporting/glm_reporter.py
Remi-Gau, line 1: Note a lot of this is duplicated from what in in nilearn/reporting/glm_reporter.py and will be refactored when adapting the "old" GLM reports.

nilearn/reporting/glm_reporter.py
```diff
@@ -302,41 +312,6 @@ def _check_report_dims(report_size):
     return width, height
 
 
-def _coerce_to_dict(input_arg):
```
Remi-Gau, line 305: extracted into nilearn/reporting/utils to be able to import into the experimental module and avoid circular import

Remi-Gau:
There are examples of the surface GLM reports in the doc.

See for example.

https://output.circle-artifacts.com/output/job/7c0fdce8-1216-48e0-a026-7238f4d863bc/artifacts/0/dev/auto_examples/08_experimental/plot_localizer_new_surface_analysis_experimental.html#generating-a-report

@bthirion @emdupre @man-shu

Can you have a look at the report and give me some UI / UX feedback?

I have already spotted a few issues when viewing on a phone but having more eyes on this may uncover other things.

Also note that once this PR is in, the idea would be to generalize this report format to all GLM and not just the surface GLM.

bthirion:
For some reason, the stat maps don't show up in the report.

bthirion:
Otherwise , the contrast plotting is not great, but this is not for this PR.

Remi-Gau:
on narrower screen the collapsed navbar is unsrollable and some of its items cannot be accessed:

image

Remi-Gau:
change the z index of the navbar as now it appears on top of evrything including the doc navbar

image

Remi-Gau:
make sure the whole nav bar is visible when it is not collapased:

image

Remi-Gau:
do not have toggle for the model details: always show them

image

Remi-Gau:
> For some reason, the stat maps don't show up in the report.

@bthirion 

are you saying they show up also outside of the report?

because they are visible in the report to me:

image

Remi-Gau:
in notebooks: images embedded in the report are also embedded **after** the report

image

Remi-Gau:
reports are too wide for notebooks

image

Remi-Gau:
Clicking links in the navbar in reports generated in a notebook: instead of jumping to the relevant section of the notebook, it opens another notebook in the notebook, so albeit interesting I doubt this inception like behavior is what I or anyone else would expect.

image

Remi-Gau:
reminder: contrasts buttons need some fixing

image

this only happens in the doc rendering

Remi-Gau:
Note I removed the surface GLM reports from the examples and only render them in the API doc with several other reports: https://output.circle-artifacts.com/output/job/59fdb255-e9ae-4b0b-aaa9-19ecdd182d4b/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html#surface-based-reports

Remi-Gau:
> change the z index of the navbar as now it appears on top of evrything including the doc navbar

fixed

Remi-Gau:
> do not have toggle for the model details: always show them

fixed

Remi-Gau:
> make sure the whole nav bar is visible when it is not collapased:

fixed

Remi-Gau:
TODO:

- include the new regressor correlation matrix in the report

bthirion:
Do you want another round of reviews on it ?

Remi-Gau:
> Do you want another round of reviews on it ?

Yes maybe an early one.
But I'd like to make sure that you can also review an actual report rendered in the doc.
The empty reports are not quite enough for now: https://output.circle-artifacts.com/output/job/496bdfd9-eb9b-4dd4-bc93-9ac1663b69ab/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html#surface-glm

Also I need to disable the navbar in notebooks for now to avoid this bug: https://github.com/nilearn/nilearn/pull/4442#issuecomment-2286039996

Remi-Gau:
> an actual report rendered in the doc

- currently blocked by by https://github.com/nilearn/nilearn/pull/5017#issuecomment-2569722063

examples/07_advanced/plot_surface_bids_analysis.py
```diff
@@ -72,8 +68,6 @@
     task_label="languagelocalizer",
     space_label="",
     img_filters=[("desc", "preproc")],
-    sub_labels=["01", "02", "05", "08"],  # comment to run all subjects
-    hrf_model="glover",
```
Remi-Gau, line 76: remove as this is the default

examples/07_advanced/plot_surface_bids_analysis.py
```diff
@@ -72,8 +68,6 @@
     task_label="languagelocalizer",
     space_label="",
     img_filters=[("desc", "preproc")],
-    sub_labels=["01", "02", "05", "08"],  # comment to run all subjects
```
Remi-Gau, line 75: running on all subjects as discussed

TODO: check how much run time is increased

Remi-Gau: [new build time](https://github.com/nilearn/nilearn/actions/runs/12653086833/job/35258014308?pr=4442#step:13:235):

- ../examples/07_advanced/plot_surface_bids_analysis.py:                      112.50 sec   1637.4 MB

[Build time with fewer subjects](https://github.com/nilearn/nilearn/actions/runs/12648293186/job/35248275649#step:13:25265) on main:

- ../examples/07_advanced/plot_surface_bids_analysis.py:                       74.87 sec    742.6 MB

Don't think that shaving 30 seconds on a build time of 90 minutes is worth worrying about for now, so let's run all subjects.

Remi-Gau:
Examples of report for better reviewing:

- in the API reference section under reporting (only empty reports for now)
  - https://output.circle-artifacts.com/output/job/6957f511-6dce-473f-8414-4c2d7d09bb83/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html#second-level-report
- examples with First and Second level reports: https://output.circle-artifacts.com/output/job/6957f511-6dce-473f-8414-4c2d7d09bb83/artifacts/0/dev/auto_examples/07_advanced/plot_surface_bids_analysis.html#mask

bthirion:
two suggestions upon looking at the report:
* the cluster table feels useless. Shouldn't we remove it ?
* the plots with grey opaque meshes are not very readable. Why not use a curvature texture to better render anatomical aspects ?

Remi-Gau:
> * the cluster table feels useless. Shouldn't we remove it ?

Given that we do not do look at the cluster on the surface, you are correct. But I intend to later reuse the same "rendering" for the volume GLM report that actually have a cluster section.

Will find a way to only add this section when relevant.

> * the plots with grey opaque meshes are not very readable. Why not use a curvature texture to better render anatomical aspects ?

Given that we cannot know what mesh users will use, there is no way to guess what the curvature is, but I can add an optional parameter in the generate report to use as a background image.

bthirion:
I'm wondering what information would end up in the table, since we probably don't want to provide surface coordinates for peaks...

Remi-Gau:
> +1 for the background_map argument (analogous to what we have for volumes).

Modified the code and the example to use the freesurfer fsaverage curvature as background image.

image

examples/07_advanced/plot_surface_bids_analysis.py
Remi-Gau, line 1: doesn't this "fetch-language-localizer-demo-dataset" have the same issue as the haxby one: it is in MNI space and we project it as is to a surface that is is fsaverage space.
if so, then we should probably add a warning about this in the example, no?

Remi-Gau:
> * the cluster table feels useless. Shouldn't we remove it ?

How it looks now

image

examples/07_advanced/plot_surface_bids_analysis.py
bthirion: But Haxby is not  in MNI space.

bthirion (commented):
Looks mostly good to me. Can be merged when CI is green.

PR #4446, opened 2024-06-11
[ENH] improve I/O for SurfaceImage

### LINKED ISSUE #4448: Create SurfaceImage from nifti
Relates to https://github.com/nilearn/nilearn/pull/4126#discussion_r1624789396

Already had to write several times a bit code that usually do:

- Get some polymesh
- Create a texture for each hemisphere by project some nifti data onto that PolyMesh
- Create a SurfaceImage instance with the PolyMesh and with the textures as Polydata

https://github.com/nilearn/nilearn/blob/46f00cde4d9198067d9e3997a6cabd9d56cc6d08/examples/08_experimental/plot_new_surface_bids_analysis_experimental.py#L101-L113

https://github.com/nilearn/nilearn/blob/46f00cde4d9198067d9e3997a6cabd9d56cc6d08/examples/08_experimental/plot_localizer_new_surface_analysis_experimental.py#L94-L105

https://github.com/nilearn/nilearn/blob/46f00cde4d9198067d9e3997a6cabd9d56cc6d08/examples/08_experimental/plot_3d_map_to_surface_projection_experimental.py#L58-L68

I expect we will see more of those pop up and it feels like it may be a fairly common use case.

So this starts screaming like need to something to do:

```python
# Can initialize image with a mesh and a NiftilikeObject
# and let the SurfaceImage class project that to the mesh
img = SurfaceImage(mesh: PolyMesh | dict[str, Mesh], data: PolyData | dict[str, Mesh] | NiftilikeObject)
```

Remi-Gau:
- Closes #4448 
- Relates to #4027 
- Relates to #4472

- Improve input/output for `SurfaceImage` by

  - loading meshes from files on disk,
  - loading data from files or Nifti object
  - saving meshes to file


tmp.py
Remi-Gau, line 1: just to test on real data: would probably need to be added to one of several of the examples

bthirion (commented):
Look good so far.

Remi-Gau:
Will most likely split the read and write in different PR.

Read is easier but we may have to make some choices for writing.

nilearn/experimental/surface/_surface_image.py
```diff
@@ -169,13 +190,78 @@ class SurfaceImage:
...
+
+    def _vol_to_surf(self, img: Nifti1Image | str | Path):
+        if isinstance(img, (str, Path)):
+            img = nb.load(img)
+        texture_left = vol_to_surf(img, self.mesh.parts["left"])
+        texture_right = vol_to_surf(img, self.mesh.parts["right"])
+        self.data = PolyData(left=texture_left.T, right=texture_right.T)
+
+    def to_filename(self, filename: str | Path) -> None:
+        """Save mesh to gifti."""
```
Remi-Gau, line 236: only implements saving of mesh which is not the most useful

would probably need to discuss how to handle saving of data (what format? gifti?)

nilearn/experimental/surface/_surface_image.py
```diff
@@ -169,13 +190,78 @@ class SurfaceImage:
...
+                            filename, new_stem=f"{filename.stem}_hemi-{hemi}"
+                        )
+                    )
+
+            return None
+
+        if "hemi-L" in filename.stem:
+            mesh = self.mesh.parts["left"]
+        if "hemi-R" in filename.stem:
+            mesh = self.mesh.parts["right"]
```
Remi-Gau, line 258: made the choice of relying on 'bids like' file naming

nilearn/experimental/surface/_surface_image.py
```diff
@@ -169,13 +190,110 @@ class SurfaceImage:
...
+
         _check_data_and_mesh_compat(self.mesh, self.data)
+
         self.shape = self.data.shape
 
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__} {getattr(self, 'shape', '')}>"
+
+    def _vol_to_surf(self, img: Nifti1Image | str | Path, **kwargs) -> None:
+        """Extract surface data from a Nifti image.
```
Remi-Gau, line 229: keeping as a private function for now

nilearn/experimental/surface/_surface_image.py
```diff
@@ -4,10 +4,15 @@
 
 import abc
 import pathlib
+import sys
+from pathlib import Path
 
+import nibabel as nb
```
bthirion, line 10:
```suggestion
import nibabel as nib
```

nilearn/experimental/surface/_surface_image.py
```diff
@@ -4,10 +4,15 @@
...
 import numpy as np
+from nibabel import Nifti1Image
```
bthirion, line 12: Do we also need the above import of nibabel ?

nilearn/experimental/surface/_surface_image.py
bthirion: Well, it's more than an extraction, it's a projection.

nilearn/experimental/surface/_surface_image.py
bthirion: Shouldn't we document this explicitly ?

bthirion (commented):
Looks good.
There is still one (legitimate) codecov complaint.

nilearn/experimental/surface/_surface_image.py
Remi-Gau: I tried to describe it in the doc string. Let me know if it's not clear enough.

nilearn/experimental/surface/_surface_image.py
Remi-Gau: actually we have a bit of mix of both in the code base: will do a quick clean up

nilearn/experimental/surface/_surface_image.py
bthirion: The explanation is clear. What I'm wondering ios whether we should dedicate a narrative documentation page to the manipulation of surface data, thout would provide this type of information.

bthirion (approved):
LGTM overall.

Remi-Gau:
> we should dedicate a narrative documentation page to the manipulation of surface data, thout would provide this type of information.

like an example or like a page in the user guide?

- in any case, I agree but is it OK with you if we add this as a TODO on the Surface "metaissue" https://github.com/nilearn/nilearn/issues/4027#issue-1925904173

bthirion:
I was thinking of a page on the user guide. 
OK to defer it to another PR.

Remi-Gau:
@man-shu do you think you could cross check this one ?

man-shu:
> @man-shu do you think you could cross check this one ?

yes on it!

Remi-Gau:
> yes on it!


Merci!!! 👍

man-shu (approved):
Just spotted a minor typo otherwise LGTM!

PR #4450, opened 2024-06-11
[REF] add force_resample default to cached calls

Remi-Gau:
- Closes none but follow up on #4441

- add explicit default of `force_resample` in cached calls to `resample_img`


bthirion (commented):
LGTM.

PR #4452, opened 2024-06-11
[FIX] Change default `ratio` value in glover to fit canonical Glover HRF (1999)

### LINKED ISSUE #2981: Glover HRF is not really Glover HRF?
Playing around with the HRFs I noticed that the Glover HRF in the GLM first_level does not actually correspond to the Glover 1999 canonical (auditory) HRF.
As these things go, it may be that I'm the one doing something wrong, so I'm opening this issue to ask whether someone else would be willing to double check my observation.

Here's a minimum working example:

```python
import numpy as np
from nilearn.glm.first_level import glover_hrf

def original_glover(t, n1=6., t1=0.9, a2=0.35, n2=12., t2=0.9):
    """ As seen in Glover 1999, page 420. Using values for the auditory cortex (canonical Glover).
    Variables are named after those in the paper. 
    """
    gamma1 = t ** n1 * np.exp(-t / t1)
    gamma2 = t ** n2 * np.exp(-t / t2)
    
    c1 = gamma1.max()
    c2 = gamma2.max()

    return gamma1 / c1 - a2 * gamma2 / c2

# Let t represent time.
t = np.arange(0, 32, 0.02)  # 32 seconds, with a TR=1 and oversampling=50 (hence 0.02).

# Plot both HRFs normalized so we can compare their shape.
plt.plot(t, original_glover(t) / original_glover(t).max(), label="Glover 1999")
plt.plot(t, glover_hrf(1.) / glover_hrf(1.).max(), label="nilearn 0.8.1")
plt.legend()
```

Results in:
Screen Shot 2021-09-22 at 23 42 41

Compare to the paper:

Screen Shot 2021-09-22 at 23 55 27

At the end of the day this is not terribly worrisome, since both are decent HRF models and neither is "true", but I guess for reporting it'd make sense to fix this if it's a bug, or we could find a way to reproduce the plot in the paper given the currently implemented machinery?

Nilearn version:

### Expected behavior

Both functions would look the same.

### Actual behavior

The Nilearn hrf is slower and has a more subtle undershoot.

const7:
- Closes #2981

- Change the default `ratio` value to `0.48` in `glover_hrf` & `glover_dispersion_derivative` when calling `_gamma_difference_hrf` to fit the canonical Glover HRF (1999) curve.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,6 +13,8 @@ Fixes
 
 - :bdg-dark:`Code` Fix failing test in ``test_nilearn_standardize`` on MacOS 14 by adding trend in simulated data (:gh:`4411` by `Hao-Ting Wang`_).
 
+- :bdg-dark:`Code` Fix previous Glover HRF implementation to fit the original paper (Glover, 1999) (:gh:`4452` by `Kun CHEN`_).
+
 Enhancements
 ------------
 
--- a/nilearn/glm/first_level/hemodynamic_models.py
+++ b/nilearn/glm/first_level/hemodynamic_models.py
@@ -349,7 +349,7 @@ def glover_dispersion_derivative(
         time_length=time_length,
         onset=onset,
         undershoot=12.0,
-        ratio=0.35,
+        ratio=0.48,
         dispersion=0.9,
     )
 
--- a/nilearn/glm/first_level/hemodynamic_models.py
+++ b/nilearn/glm/first_level/hemodynamic_models.py
@@ -141,7 +141,7 @@ def glover_hrf(tr, oversampling=50, time_length=32.0, onset=0.0):
         undershoot=12.0,
         dispersion=0.9,
         u_dispersion=0.9,
-        ratio=0.35,
+        ratio=0.48,
     )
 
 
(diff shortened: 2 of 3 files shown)


Remi-Gau:
Not looked in detail but if we decide to go through with this there should be an entry in the changelog as this may lead to different results for users.

const7:
The modification has been added to the changelog here https://github.com/nilearn/nilearn/pull/4452/commits/07899aaff95acfe085b73c68f94a1ec8df748a8c. Please let me know if there is anything else to add.

bthirion (commented):
Thx for taking care of that. Can you take a look at the broken tests ?
Best,

const7:
The test failed at assertion in `nilearn/glm/tests/test_dmtx.py::test_compare_design_matrix_to_spm[10-arr_1]`:
https://github.com/nilearn/nilearn/blob/47964126e831a9d5f4eae19f24597144fda923ca/nilearn/glm/tests/test_dmtx.py#L432-L446

With `ratio` set to `0.35`, the assertion result is `4.52/187.13=0.024<0.1`;
With `ratio` set to `0.48`, it becomes `20.69/187.13=0.11>0.1`.

Could you suggest how to proceed? Thank you!

Remi-Gau:
> Could you suggest how to proceed?

I think something is strange with this test: it says it compares the generated design matrix to one made with SPM, yet it relies on the default HRF (glover) and not the SPM one.

I would update the test to make it use the SPM hrf.

@bthirion does it sound reasonable?

Remi-Gau:
@const7 I 'fixed' the test.

@bthirion if you are good with this change we can merge this one

bthirion (approved):
LGTM, thx !

Remi-Gau (approved):

Issue #4368, opened 2024-04-09
report CSS and code refactor


Remi-Gau:
- a lot of the CSS should be refactored across reports (glm and maskers) to ensure a common style
- some html sections should be also refactored across reports (for glm and maskers reports the footer section for example)

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/4295#issuecomment-1971863445_

PR #4460, opened 2024-06-20
[REF] refactor report templates

Remi-Gau:
- Closes none
- Related to #4368 
- Extracted from #4442

- use tempita language template inheritance to make html report more modular
    - inherit from a base.html template to contains the common content for all templates
- add "partials" (not a concept with tempita but common in other templating languages) for warning banner, parameter tables...


nilearn/reporting/html_report.py
```diff
@@ -93,7 +104,7 @@ def _update_template(
     title : str
         The title for the report.
 
-    docstring : str
+    docstring : strresource_path
```
Remi-Gau, line 107:
```suggestion
    docstring : str
```

Remi-Gau:
For review

Number of line changes is on the high end, because of all the HTML and CSS refactoring.

Uploading an example of all the reports that should show in the documentation, as some visual reports may speak louder than a 1000 lines of code change.

[reports.zip](https://github.com/user-attachments/files/15915284/reports.zip)

PR #4462, opened 2024-06-21
[FIX] use string as label names in example involving the new surface based destrieux atlas

Remi-Gau:
- Closes none but should fix doc build failure in main

- the exprimental [fetch_destrieux for surface data returns a list of strings](https://github.com/nilearn/nilearn/commit/59c236d74b56d18f0c0c534fa329666cfc7bfb98#r143380304) as label names and not a list of bytes as its volume counterpart: this PR fixes this


Remi-Gau:
will merge just to check that this fixes the doc build on main

PR #4463, opened 2024-06-21
[MAINT] for monitoring of duration of CI only plot doc build on main (full builds)

Remi-Gau:
- Closes none

relates to this workflow that monitor duration of test suite and doc build

- only keep the build of the doc on main branch (push only and not pull request)

previous figures showed a mix of full and partial builds:

doc_runs_timing

new figure only shows full builds:

doc_runs_timing

### DIFF
--- a/maint_tools/check_gha_workflow.py
+++ b/maint_tools/check_gha_workflow.py
@@ -69,7 +69,7 @@ INCLUDE_FAILED_RUNS = False
 
 # Pages of runs to collect
 # 100 per page
-PAGES_TO_COLLECT = range(1, 20)
+PAGES_TO_COLLECT = range(1, 30)
 
 # If False, just plots the content of the TSV
 UPDATE_TSV = True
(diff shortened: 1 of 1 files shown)


bthirion:
> only keep the build of the doc on main branch (push only and not pull request)

But does this mean, that we're going to merge without knowing whether the doc compiles ? I find it a bit worrying. 
I think we should force (partial) doc generation at least once before merging.

Remi-Gau:
> > only keep the build of the doc on main branch (push only and not pull request)
> 
> But does this mean, that we're going to merge without knowing whether the doc compiles ? I find it a bit worrying. I think we should force (partial) doc generation at least once before merging.

Sorry I was not clear.

The modified workflow just checks the github logs and does not build anything. 
It's just that until now the figure was showing the time it took to build things on push events and pull request events. Now it would only show the duration of push events. I could maybe have it plot both as different lines.

To make sure: this does not affect the doc building workflow that will still run on every PR and let us know if it fails. (the deploy part is still broken though 😢 )...

bthirion (approved):
I cannot spot any issue.

Issue #4346, opened 2024-03-28
[ENH] Adding new checks and features to NiftiLabelsMasker reporting


mtorabi59:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed enhancement in detail.

Currently the `NiftiLabelsMasker` reporting is executed only after fit and before the data is transformed. This causes the reporting to not be aware of the changes that are going to happen for the atlas, e.g. due to masking or resampling. It would be great if the reporting was also checking and also previewing how the atlas and its regions will look after applying resampling and masking.

### Benefits to the change

This change will help the user be more aware of the impact of resampling or masking they are applying to the data when applying maskers. For instance, if some regions are going to be dropped due to resampling or masking, the reporting will inform the user.

### Pseudocode for the new behavior, if applicable

```python
# insert your code below

```

bthirion:
I'm unsure about that: I think that the report should represented he fitted object. When the transform method() is used, the internal state of the object no longer changes, it becomes like a function with a simple input/output logic.  So I don't see the point of having report related to the transform.

mtorabi59:
@bthirion It's not about having a report for transform of the data, but about how the atlas will look after resampling and masking. Thing is for now, as far as I understood, resampling and masking are not applied even on the atlas until transform is called. So we won't be aware of their impact on the atlas and its regions before we transform the data. We could add some functions to apply resampling and masking only on the atlas before transforming the data. However, this requires the user to pass the data if the resampling target is the data.
I don't know if overall it worths the hassle.

bthirion:
Not sure either. This is something to discuss in a forthcoming meeting.

PR #4500, opened 2024-08-05
[ENH] add a function to help reporting post-masking atlas in LabelsMaskers

mtorabi59:
- Addresses #4346 and #3085 

- Adding a new private function to `NiftiLabelsMasker` class which applies masking on the atlas after fit and before transform is applied to provide a preview of how the atlas will look like and which regions will be removed and which regions will be kept. Note that some regions despite not being removed, might become very small or thin and the visualizing option can help noticing such cases.


Remi-Gau:
@mtorabi59 
do you think you could include in the PR an image example generated by this function?
probably worth adding a test for it as well.

mtorabi59:
I have made an example which shows how this function can be useful. It starts with "Harvard Oxford atlas": 
image
Next, I create a mask that only keeps 3 of the regions. Also, for one of the regions the mask only keeps 20% of the region. Then I fit `NiftiLabelsMasker` using the created mask. Now, when we call the new function, `_post_masking_atlas()`, it returns the removed regions, the new atlas (masked), and a display of how the new atlas looks like:
image
The user can see and become aware using this visualization which regions are lost and which regions might have been partially lost (like the blue region).

    from nilearn import plotting
    import numpy as np
    from nilearn.regions.tests.test_signal_extraction import get_data
    
    ## load func data
    from nilearn import datasets
    func_filename = dataset.func[0]
    
    ## load the atlas
    atlas = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-2mm')
    # The first label correspond to the background
    print('The atlas contains {} non-overlapping regions'.format(
        len(atlas.labels) - 1))
    # plot the atlas
    plotting.plot_roi(atlas.maps, title="Harvard Oxford atlas")
    plt.show()
    
    ## create the mask
    labels_data = get_data(atlas.maps)
    mask_data = (labels_data == 1) + (labels_data == 2) + (labels_data == 5)
    # randomly remove 80 percent of the data labelled 1
    mask_data[labels_data == 1] = np.random.choice([0, 1], size=mask_data[labels_data == 1].shape, p=[0.8, 0.2])
    mask_img = Nifti1Image(mask_data.astype(np.int8), atlas.maps.affine)
    
    ## create and fit the NiftiLabelsMasker
    # Instantiate the masker with label image and label values
    masker = NiftiLabelsMasker(atlas.maps,
                               labels=atlas.labels,
                               mask_img=mask_img,
                               standardize=True,
                               keep_masked_labels=False
                               )
    _ = masker.fit()
    
    ## use the new function
    masked_atlas, removed_region_ids, removed_region_names, display = masker._post_masking_atlas(visualize=True)
    
    print('Removed region ids:', removed_region_ids)
    print('Removed region names:', removed_region_names)
    print(f"{len(removed_region_ids)} regions were removed")
    plt.show()

bthirion:
SOrry, I'm missing something: removing 80% of the voxels for one of the regions does not change the report, does it ?

mtorabi59:
> SOrry, I'm missing something: removing 80% of the voxels for one of the regions does not change the report, does it ?

No, it just visually shows it to the user. If you think it would be useful, we can set a threshold like 50% and if a region loses more than 50% of its voxels we report it as mostly masked region. We can add another output like `mostly_masked_out_regions` or something like that. WDYT?

Remi-Gau:
discussed with @mtorabi59 
merging this and @mtorabi59 will work on a follow up

PR #4502, opened 2024-08-07
[FIX] Reduce `dtype` of mask from `int64` to `int8` in `image.binarize_img`

### LINKED ISSUE #4498: [BUG] `image.binarize_img` emits a `UserWarning` about 64-bit ints when provided images that have other `dtypes`

### Operating system

- [ ] Linux
- [X] Mac
- [ ] Windows

### Operating system version

  System Version:	macOS 14.5 (23F79)
  Kernel Version:	Darwin 23.5.0

### Python version

- [ ] 3.12
- [X] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

0.10.4

### Expected behavior

When working with `float64` images, I did not expect warnings about `int64`.

This is what I got when calling `image.binarize_img`:

```bash
UserWarning: Data array used to create a new image contains 64-bit ints. This is likely due to creating the array with numpy and passing `int` as the `dtype`. Many tools such as FSL and SPM cannot deal with int64 in Nifti images, so for compatibility the data has been converted to int32.
  return new_img_like(niimg, result, niimg.affine)
```

That seems to make sense, because `binarize_img` is passing `int` as `dtype`: https://github.com/nilearn/nilearn/blob/4412c4210d07c02fe45a17ef0f64131292a7e892/nilearn/image/image.py#L1244-L1254

```python
> import numpy as np
> from nilearn import image, datasets

> img = image.new_img_like(datasets.load_mni152_template(), np.random.normal(size=(3,3,3)))
> img.get_data_dtype()
dtype('<f8')

> image.binarize_img(img)
```

psadil:
- Closes #4498

- Avoids the warning by casting the mask to `np.int8` instead of `int`(64)

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -16,7 +16,7 @@ Fixes
 
 - :bdg-dark:`Code` Fix previous Glover HRF implementation to fit the original paper (Glover, 1999) (:gh:`4452` by `Kun CHEN`_).
 
-- :bdg-dark:`Code` Avoid triggering ``UserWarning`` by :func:`nilearn.image.binarize_img` (:gh:`4498` by `Patrick Sadil`_).
+- :bdg-dark:`Code` :func:`nilearn.image.binarize_img` explicitly cast images to ``int8`` to avoid warnings about ``int64`` when working with ``float64`` images (:gh:`4498` by `Patrick Sadil`_).
 
 Enhancements
 ------------
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -16,6 +16,8 @@ Fixes
 
 - :bdg-dark:`Code` Fix previous Glover HRF implementation to fit the original paper (Glover, 1999) (:gh:`4452` by `Kun CHEN`_).
 
+- :bdg-dark:`Code` Avoid triggering ``UserWarning`` by :func:`nilearn.image.binarize_img` (:gh:`4498` by `Patrick Sadil`_).
+
 Enhancements
 ------------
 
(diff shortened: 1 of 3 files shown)


doc/changes/latest.rst
```diff
@@ -16,6 +16,8 @@ Fixes
 
 - :bdg-dark:`Code` Fix previous Glover HRF implementation to fit the original paper (Glover, 1999) (:gh:`4452` by `Kun CHEN`_).
 
+- :bdg-dark:`Code` Avoid triggering ``UserWarning`` by :func:`nilearn.image.binarize_img` (:gh:`4498` by `Patrick Sadil`_).
```
Remi-Gau, line 19:
```suggestion
- :bdg-dark:`Code` :func:`nilearn.image.binarize_img` explicitly cast images to ``int8`` to avoid warnings about ``int64`` when working with ``float64`` images (:gh:`4498` by `Patrick Sadil`_).
```

Remi-Gau: small suggestion to make the release note a bit clearer as to why the change was made

nilearn/image/image.py
```diff
@@ -1242,7 +1242,7 @@ def binarize_img(
     )
 
     return math_img(
-        "img.astype(bool).astype(int)",
+        "img.astype(bool).astype(np.int8)",
```
bthirion, line 1245:
```suggestion
        "img.astype(bool).astype('int8')",
```

bthirion (commented):
Thx ! Ideally, we should add a test to ensure that the new behavior is strictly enforced

Remi-Gau (approved):

bthirion (approved):
LGTM, thx !

bthirion:
Merging now

PR #4506, opened 2024-08-19
[REF] refactor nibabel import

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/4446#discussion_r1721349617

Changes proposed in this pull request:

- replace `import nibabel` by `import nibabel as nib` or by loading only the required modules, classes or functions
- where possible replace the `nibabel.save` with the relevant `to_filename()` method

PR #4508, opened 2024-08-21
[FIX] revert save analyze for SPM auditory data

Remi-Gau:
- Closes none but addresses https://github.com/nilearn/nilearn/pull/4506#issuecomment-2299636558

- use nibabel.save to to save to 'Analyze' format


bthirion (approved):
Thx !

Issue #3861, opened 2023-07-20
Apply black to the files still involved in other pull requests


Remi-Gau:
- realized that during the hackathon some people would use black and end up reformating files that should not have because they are still involved in other PRs
- some of the PRs with those files have been opened for a certain amount of time and it may still take some time for them to be finished

it would be nice to find a way to be able to apply black to the whole codebase and minimize merge conflicts on those old PRs

possible route: 

- apply black to a file in main
- use the same black config to reformat the file in the PR
- hopefully this should be easy (#FamousLastWords)

bthirion:
No better suggestion. How far are we with black. I thought it was over (did not check carefully, sorry).

Remi-Gau:
We are done for all the files not involved in PR.
The list of excluded files is in the pyproject.toml.
But during the hackathon we realized that those  excluded files can still lead to headaches if people's IDE is set to auto reformat or if they run black manually on some files.
So it is better to find a way to finish the job completely.

bthirion:
We should try your approach in one submodule and see how well it goes.

Remi-Gau:
Either that or do it PR by PR.

ymzayek:
We can start with files in PRs opened by core devs

Remi-Gau:
You read my mind.

Remi-Gau:
started trying "fixing black" on some of the old PRs... it feels that it will prove a bit of headache in any case
will try with a couple more but it seems that there will always some merge conflicts to fix

bthirion:
But "There is no alternative" as they put it...

ymzayek:
I linked these latest PRs here but I misread the title of the issue. The new blacked files are NOT involved in pull requests

Remi-Gau:
I would suggest to format the last few remaining files before the next release:

```
examples/01_plotting/plot_3d_map_to_surface_projection.py
nilearn/plotting/html_surface.py
nilearn/plotting/js_plotting_utils.py
nilearn/plotting/surf_plotting.py
nilearn/plotting/tests/test_surf_plotting.py
nilearn/surface/surface.py
nilearn/surface/tests/test_surface.py
```

many of them were not formatted due to some old PR related to surface handling

PR #4519, opened 2024-09-03
[STY] apply black to surface subpackage

Remi-Gau:
- Closes none but relates to #3861

### DIFF
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -30,25 +30,29 @@ def _uniform_ball_cloud(n_points=20, dim=3, n_monte_carlo=50000):
     """Get points uniformly spaced in the unit ball."""
     rng = np.random.RandomState(0)
     mc_cube = rng.uniform(-1, 1, size=(n_monte_carlo, dim))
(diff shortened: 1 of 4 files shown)


nilearn/surface/surface.py
```diff
@@ -157,26 +172,33 @@ def _ball_sample_locations(
 
     """
     if depth is not None:
-        raise ValueError("The 'ball' sampling strategy does not support "
-                         "the 'depth' parameter")
+        raise ValueError(
+            "The 'ball' sampling strategy does not support "
+            "the 'depth' parameter"
```
bthirion, line 177: The error message is not great. Maybe we could take this opportunity to improve it...

bthirion (commented):
LGTM otherwise.

nilearn/surface/surface.py
Remi-Gau: will try to improve it

Remi-Gau: I at least added a tip on how to avoid the error.

IMO this error should just be a warning.

PR #4527, opened 2024-09-04
[REF] refactor check matplolib and plotly

### LINKED ISSUE #4524: refactor matplolib import check
introducing this function: we have quite a few matplolib import check peppered in the code base: this could help refactor them

```python
def is_matplotlib_installed():
    """Check if matplotlib is installed."""
    try:
        import matplotlib  # noqa: F401
    except ImportError:
        return False
    else:
        return True
```

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/4522#discussion_r1743326043_

Remi-Gau:
- Closes #4524

- use a single function to check that matplotlib is installed

### DIFF
--- a/examples/00_tutorials/plot_python_101.py
+++ b/examples/00_tutorials/plot_python_101.py
@@ -3,16 +3,19 @@ Basic numerics and plotting with Python
 =======================================
 """
 
-try:
-    import matplotlib.pyplot as plt
-except ImportError:
+from nilearn._utils.helpers import is_matplotlib_installed
+
+if not is_matplotlib_installed():
     raise RuntimeError("This script needs the matplotlib library")
+
+import matplotlib.pyplot as plt
+
+# import numpy: the module providing numerical arrays
+import numpy as np
 
 # %%
 # A simple example of basic Python numerics and how to plot it.
 
-# import numpy: the module providing numerical arrays
-import numpy as np
 
 t = np.linspace(1, 10, 2000)
 
--- a/examples/01_plotting/plot_3d_map_to_surface_projection.py
+++ b/examples/01_plotting/plot_3d_map_to_surface_projection.py
@@ -78,9 +78,9 @@ fig.show()
 
 engine = "plotly"
 # If plotly is not installed, use matplotlib
-try:
-    import plotly.graph_objects as go  # noqa: F401
-except ImportError:
+from nilearn._utils.helpers import is_plotly_installed
+
+if not is_plotly_installed():
     engine = "matplotlib"
 
 print(f"Using plotting engine {engine}.")
--- a/nilearn/experimental/surface/maskers.py
+++ b/nilearn/experimental/surface/maskers.py
@@ -13,6 +13,7 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from nilearn import signal
 from nilearn._utils.cache_mixin import CacheMixin, cache
 from nilearn._utils.class_inspect import get_params
+from nilearn._utils.helpers import is_matplotlib_installed
 from nilearn.experimental.surface._surface_image import PolyMesh, SurfaceImage
 
 
@@ -296,9 +297,7 @@ class SurfaceMasker(BaseEstimator, TransformerMixin, CacheMixin):
 
     def generate_report(self):
         """Generate a report."""
-        try:
-            from nilearn.reporting.html_report import generate_report
-        except ImportError:
+        if not is_matplotlib_installed():
             with warnings.catch_warnings():
                 mpl_unavail_msg = (
                     "Matplotlib is not imported! "
@@ -307,6 +306,8 @@ class SurfaceMasker(BaseEstimator, TransformerMixin, CacheMixin):
                 warnings.filterwarnings("always", message=mpl_unavail_msg)
                 warnings.warn(category=ImportWarning, message=mpl_unavail_msg)
                 return [None]
+
+        from nilearn.reporting.html_report import generate_report
 
         return generate_report(self)
 
@@ -608,9 +609,7 @@ class SurfaceLabelsMasker(BaseEstimator):
 
     def generate_report(self):
         """Generate a report."""
-        try:
-            from nilearn.reporting.html_report import generate_report
-        except ImportError:
+        if not is_matplotlib_installed():
             with warnings.catch_warnings():
                 mpl_unavail_msg = (
                     "Matplotlib is not imported! "
@@ -619,6 +618,8 @@ class SurfaceLabelsMasker(BaseEstimator):
                 warnings.filterwarnings("always", message=mpl_unavail_msg)
                 warnings.warn(category=ImportWarning, message=mpl_unavail_msg)
                 return [None]
+
+        from nilearn.reporting.html_report import generate_report
 
         return generate_report(self)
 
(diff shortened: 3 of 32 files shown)


bthirion (approved):
GTM, thx !

Remi-Gau:
this is a bigger PR: will wait for another review

Remi-Gau:
Created a `review required` label for PR that are ready and need at least another review.

nilearn/maskers/tests/test_nifti_masker.py
```diff
@@ -604,7 +604,8 @@ def test_nifti_masker_io_shapes(rng):
 
 
 @pytest.mark.skipif(
-    have_mpl, reason="Test requires matplotlib not to be installed."
+    is_matplotlib_installed(),
+    reason="Test requires matplotlib not to be installed.",
```
hndgzkn, line 608: This marker repeats in many places. One option might be to define it in a module and import it from others.

examples/00_tutorials/plot_python_101.py
```diff
@@ -3,17 +3,20 @@
 =======================================
 """
 
-try:
-    import matplotlib.pyplot as plt
-except ImportError:
+from nilearn._utils.helpers import is_matplotlib_installed
+
+if not is_matplotlib_installed():
     raise RuntimeError("This script needs the matplotlib library")
```
hndgzkn, line 9: This repeats in many places. An option would be to implement it as a function in `helpers`.

nilearn/experimental/surface/maskers.py
```diff
@@ -620,6 +619,8 @@ def generate_report(self):
                 warnings.warn(category=ImportWarning, message=mpl_unavail_msg)
```
hndgzkn, line 619: Could this warning be part of `is_matplotlib_installed`?

nilearn/experimental/surface/maskers.py
```diff
@@ -608,9 +609,7 @@ def inverse_transform(self, masked_img: np.ndarray) -> SurfaceImage:
 
     def generate_report(self):
         """Generate a report."""
-        try:
-            from nilearn.reporting.html_report import generate_report
-        except ImportError:
+        if not is_matplotlib_installed():
             with warnings.catch_warnings():
                 mpl_unavail_msg = (
                     "Matplotlib is not imported! "
```
hndgzkn, line 615: Could this warning be part of `is_matplotlib_installed`?

examples/00_tutorials/plot_python_101.py
Remi-Gau: yeah that's a good point: let me refactor this.

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -78,9 +78,9 @@
 
 engine = "plotly"
 # If plotly is not installed, use matplotlib
-try:
-    import plotly.graph_objects as go  # noqa: F401
-except ImportError:
+from nilearn._utils.helpers import is_plotly_installed
+
+if not is_plotly_installed():
     engine = "matplotlib"
```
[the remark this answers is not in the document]
Remi-Gau: > Isn't it also necessary to check if matplotlib is installed?

yes you are correct: doing that in almost all examples now

also the error message should mention which optional dependency to install to fix the issue

examples/00_tutorials/plot_python_101.py
Remi-Gau: done

nilearn/experimental/surface/maskers.py
Remi-Gau: actually I reverted this change because it would not raise the proper error when matploblib was not installed

nilearn/maskers/tests/test_nifti_masker.py
Remi-Gau: I count 4 repeats in total in the whole codebase: I think this is still manageable at this stage.

bthirion (commented):
I'm wondering whether we want to enforce that in examples.
Obviously, running the examples without mpl installed does not make sense. The drawback is that it adds boilerplate to the examples.

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -10,6 +10,10 @@
 
 """
 
+from nilearn._utils.helpers import check_matplotlib
+
+check_matplotlib()
+
```
Remi-Gau, line 16:
```suggestion
```

examples/07_advanced/plot_ica_neurovault.py
```diff
@@ -70,8 +69,8 @@
 print("\nReshaping and masking images.\n")
 
 with warnings.catch_warnings():
-    warnings.simplefilter("ignore", UserWarning)
-    warnings.simplefilter("ignore", DeprecationWarning)
+    # warnings.simplefilter("ignore", UserWarning)
```
bthirion, line 72: to be removed

Remi-Gau: Quite a few UserWarning so will keep at least one (though I do not like to have examples that show the bad practice of ignoring warnings)

Remi-Gau:
Regarding the "checks for matplotlib" in examples:
- they are run if the script does not import the nilearn.plotting at some point because otherwise nilearn.plotting has its own matlotlib check
- OR if mataplotlib is imported before nilearn.plotting

This makes it hard to maintain: because we need to scan the entire script to know if we must have this check.

I would recommend for all examples to hoist ALL the imports at the top: this would make this kind of check easier.

hndgzkn:
> Regarding the "checks for matplotlib" in examples:
> 
>     * they are run if the script does not import the nilearn.plotting at some point because otherwise nilearn.plotting has its own matlotlib check
> 
>     * OR if mataplotlib is imported before nilearn.plotting
> 

I think it is good to know at the beginning of the notebook whether there are import problems or not. 

> This makes it hard to maintain: because we need to scan the entire script to know if we must have this check.

I agree

> I would recommend for all examples to hoist ALL the imports at the top: this would make this kind of check easier.

However it seems better to me to import packages where they are first used in the examples. 

I would suggest alternatively to put an `examples` option in `pyproject.toml` and document it in `README`.

hndgzkn:
If we document it in the [install section of README](https://github.com/nilearn/nilearn/tree/main?tab=readme-ov-file#install), then do we need to check? I think in that case we can remove all checks in the examples.  

And from users point of view, are we expecting them to run examples using notebooks? If so, we can also add jupyter to examples option.

nilearn/experimental/surface/maskers.py
```diff
@@ -297,9 +298,7 @@ def inverse_transform(self, masked_img: np.ndarray) -> SurfaceImage:
...
```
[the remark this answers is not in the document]
Remi-Gau: It may be possible to refactor those indeed.
I have not looked into it. But that would be for another issue / set of PRs.

Given that I started some work to "uniformize" the reporting across maskers and GLM, I would be in favor of this.

hndgzkn (approved):
Ok with  me!

bthirion (approved):

Remi-Gau:
> If we document it in the [install section of README](https://github.com/nilearn/nilearn/tree/main?tab=readme-ov-file#install), then do we need to check? I think in that case we can remove all checks in the examples.
> 
> And from users point of view, are we expecting them to run examples using notebooks? If so, we can also add jupyter to examples option.

Maybe worth a quick chat at our next dev meeting to decide what we want to do with this ?

Remi-Gau:
thanks for the reviews @hndgzkn @bthirion

bthirion:
I agree that this deserves a coredev discussion.

PR #4529, opened 2024-09-04
[MAINT] Clean up some warnings from the tests

Remi-Gau:
- Closes none 

- change some tested value in space_net to avoid warnings
- update first_level_from bids to reduce amount of meaningless warnings
- update searchlight to no pass groups to KFold CVs as it does not use it anyway (would trigger a warning otherwise)

### DIFF
--- a/nilearn/decoding/tests/test_space_net.py
+++ b/nilearn/decoding/tests/test_space_net.py
@@ -38,7 +38,7 @@ PENALTY = ["graph-net", "tv-l1"]
 
 
 @pytest.mark.parametrize("is_classif", IS_CLASSIF)
-@pytest.mark.parametrize("l1_ratio", [0.5, 1.0])
+@pytest.mark.parametrize("l1_ratio", [0.5, 0.99])
 @pytest.mark.parametrize("n_alphas", range(1, 10))
 def test_space_net_alpha_grid(
     rng, is_classif, l1_ratio, n_alphas, n_samples=4, n_features=3
@@ -101,7 +101,7 @@ def test_early_stopping_callback_object(rng, n_samples=10, n_features=30):
 @pytest.mark.parametrize("penalty", PENALTY)
 @pytest.mark.parametrize("is_classif", IS_CLASSIF)
 @pytest.mark.parametrize("n_alphas", [0.1, 0.01])
-@pytest.mark.parametrize("l1_ratio", [0.5, 1.0])
+@pytest.mark.parametrize("l1_ratio", [0.5, 0.99])
 @pytest.mark.parametrize("n_jobs", [1, -1])
 @pytest.mark.parametrize("cv", [2, 3])
 @pytest.mark.parametrize("perc", [5, 10])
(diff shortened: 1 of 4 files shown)


nilearn/decoding/tests/test_space_net.py
```diff
@@ -38,7 +38,7 @@
 
 
 @pytest.mark.parametrize("is_classif", IS_CLASSIF)
-@pytest.mark.parametrize("l1_ratio", [0.5, 1.0])
+@pytest.mark.parametrize("l1_ratio", [0.5, 0.99])
```
Remi-Gau, line 41: using a value of 0 or 1 triggers a warning

nilearn/glm/first_level/first_level.py
```diff
@@ -1811,6 +1811,10 @@ def _get_events_files(
     events : :obj:`list` of :obj:`str`
         List of fullpath to the events files
     """
+    # pop the 'desc' filter
+    # it would otherwise trigger some meaningless warnings
+    # as desc entity are not supported in BIDS raw datasets
+    img_filters = [x for x in img_filters if x[0] != "desc"]
```
Remi-Gau, line 1817: these should reduce the number of warnings we get when running the tests but also reduce warnings users get when filtering BIDS dataset with the desc entity

Remi-Gau:
Getting about >100 warnings in the searchlight tests:

```
nilearn/decoding/tests/test_searchlight.py: 125 warnings
  /home/remi/miniconda3/lib/python3.12/site-packages/sklearn/model_selection/_split.py:91: UserWarning: The groups parameter is ignored by KFold
    warnings.warn(
```

Seems related to this:

https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_score.html#cross-val-score


image


Probably needs a separate issue and PR

Remi-Gau:
## before

image

## after

image

nilearn/decoding/tests/test_space_net.py
bthirion: Yes, but... isn't it because 1.0 at the limit of the parameter domain ?
It could be that using 1.0 triggers another optimization approach (if not, forget my comment).
We need to check whether having the parameter equal to 0 or 1 triggers a special behavior in the ensuing optimization.

Remi-Gau: Sorry if I was unlclear, there are still some tests that use 0 or 1 that I have not touched that are actual proper check of the behavior for those limit values:

- test_lasso_vs_graph_net
- test_log_reg_vs_graph_net_two_classes_iris
- test_space_net_alpha_grid_pure_spatial

I only changed the values that were used in the test parametrizations.

Remi-Gau:
> Getting about >100 warnings in the searchlight tests:
> 
> ```
> nilearn/decoding/tests/test_searchlight.py: 125 warnings
>   /home/remi/miniconda3/lib/python3.12/site-packages/sklearn/model_selection/_split.py:91: UserWarning: The groups parameter is ignored by KFold
>     warnings.warn(
> ```
> 
> Seems related to this:
> 
> https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_score.html#cross-val-score
> 
> 
> Probably needs a separate issue and PR

my bad this actually relates to this: https://github.com/scikit-learn/scikit-learn/issues/22848

K-Fold used in tests does not support groups

bthirion (commented):
LGTM, thx.

bthirion (approved):

Remi-Gau:
OK this feels smallish enough to be merged with one review

PR #4530, opened 2024-09-05
[MAINT] raise stacklevel of `copy_header` warning

Remi-Gau:
- Closes none

- raise stacklevel of `copy_header` warning to allow user to know which function call triggered the warning so they can act on it.


bthirion (approved):
LGTM, thx.
Your prupose is to un-clutter the tests output, right ?

Remi-Gau:
yes because for maintainers when there are > 1000 warnings it is hard to see new warnings that may come from our dependencies that we should pay attention to

but in this case it's also because the the warning would tell the user "use this default" but it would not tell the user which of their function call triggered it

PR #4535, opened 2024-09-06
[FIX] update scikit-learn decoder tags

Remi-Gau:
- Closes #4533 

Changes proposed in this pull request:

- adapt "_more_tags" methods and tests
- add TODOs to switch to the new `__sklearn_tags__` when dropping sklearn 1.5.1

PR #4546, opened 2024-09-13
[FIX] fix version to check for scikit learn for new decoder tags

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/4535#issuecomment-2348135156


bthirion (approved):
CI is green which is all we need for that one I think.
Best,

PR #4554, opened 2024-09-19
 [ENH] Add last version of the AAL atlas to its fetcher

### LINKED ISSUE #4551: [DOC] Cingulate regions labeled as cingulum in fetched AAL atlas

Hi Nilearn team :wave:
The fetched AAL atlas is version 2 from 2015 (see [datasets/atlas.py](https://github.com/nilearn/nilearn/blob/main/nilearn/datasets/atlas.py#L1300-L1309)), while the current [AAL version 3 dates from 2019 with recent updates](https://www.gin.cnrs.fr/en/tools/aal/). We noticed this because AAL2 labels cingulate regions as 'cingulum'; I'm not sure if terminology has changed, but now '[cingulum](https://en.wikipedia.org/wiki/Cingulum_%28brain%29)' is universally understood as the large white matter track and *not* a gray matter region.
Would it be feasible to update the Nilearn atlas to reflect the latest AAL3 labeling?

JLefortBesnard:
- Closes #4551 

- Added an option in `fetch_atlas_aal(version="3v2")` to fetch the latest AAL version.
- Updated the documentation accordingly.
- Added a _DeprecationWarning_ when the default AAL atlas is fetched, notifying users that version 3v2 will become the default instead of SPM12 starting from version 0.13.

### DIFF
--- a/nilearn/datasets/tests/test_atlas.py
+++ b/nilearn/datasets/tests/test_atlas.py
@@ -17,7 +17,7 @@ from sklearn.utils import Bunch
 from nilearn._utils import data_gen
 from nilearn._utils.testing import serialize_niimg
 from nilearn.conftest import _rng
-from nilearn.datasets import atlas
+from nilearn.datasets import atlas, fetch_atlas_aal
 from nilearn.datasets._utils import fetch_files
 from nilearn.datasets.tests._testing import dict_to_archive
 from nilearn.image import get_data
@@ -824,3 +824,10 @@ def test_fetch_atlas_schaefer_2018(tmp_path, request_mocker):
 
         assert img.header.get_zooms()[0] == resolution_mm
         assert np.array_equal(np.unique(img.dataobj), np.arange(n_rois + 1))
+
+
+def test_all_version_deprecation():
+    with pytest.deprecated_call(
+        match=r"Starting in version 0\.13, the default fetched mask"
+    ):
+        fetch_atlas_aal()
--- a/nilearn/datasets/tests/test_atlas.py
+++ b/nilearn/datasets/tests/test_atlas.py
@@ -826,7 +826,13 @@ def test_fetch_atlas_schaefer_2018(tmp_path, request_mocker):
         assert np.array_equal(np.unique(img.dataobj), np.arange(n_rois + 1))
 
 
-def test_all_version_deprecation():
+def test_aal_version_deprecation(request_mocker):
+
+    img_pattern = re.compile("https://www.gin.cnrs.fr/*")
+    request_mocker.url_mapping[img_pattern] = (
+        data_gen.generate_labeled_regions((7, 6, 5), 16)
+    )
+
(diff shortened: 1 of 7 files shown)


Remi-Gau:
one quick thing to do / check:
- check the content of the atlas desciption to see if it should be updated.
  - I suspect we could actually transfer some of the info from the doc string to the atlas description
  - the name of the atlas description should be renamed to `aal.rst` so it is version independent: may have to also update that in the loading in the of description in the fetcher
  - https://github.com/nilearn/nilearn/blob/main/nilearn/datasets/description/aal_SPM12.rst

Remi-Gau:
to fix the code formatting issue, install precommit: https://nilearn.github.io/stable/development.html#pre-commit

then run 

```bash
pre-commit run -a
```

That should fix most issues that you can then commmit.

Remi-Gau:
2 other things we may want:
- an update to the changelog 

- a test that the warning is indeed thrown when people use the default value


I think adding in this [file](https://github.com/nilearn/nilearn/blob/main/nilearn/datasets/tests/test_atlas.py) something like:

```python
from nilearn._utils.testing import check_deprecation
from nilearn.datasets import fetch_atlas_aal

def test_check_deprecation():
    check_deprecation(fetch_atlas_aal, "the default fetched mask will be AAL 3v2")()
```

Remi-Gau:
The doc build is failing because of this line that needs to be changed:
https://github.com/nilearn/nilearn/blob/1471ba887c1109d8ee8986179e0a229823a25eb6/doc/modules/datasets.rst?plain=1#L82

Sorry I had forgotten about it.

Remi-Gau:
let me check why the test is failing

nilearn/datasets/tests/test_atlas.py
```diff
@@ -824,3 +825,40 @@ def test_fetch_atlas_schaefer_2018(tmp_path, request_mocker):
...
+        shape_3d_default, 15, affine=affine_eye
+    )
+    output_path = tmp_path / "aal_SPM12/aal/atlas/AAL.nii"
+    output_path.parent.mkdir(parents=True)
+    img.to_filename(output_path)
+
+    with open(
+        tmp_path / "aal_SPM12" / "aal" / "atlas" / "AAL.xml", "wb"
+    ) as file:
+        aal_xml.write(file, encoding="ISO-8859-1", xml_declaration=True)
```
Remi-Gau, line 857: had to mock some of the output to make this work

JLefortBesnard: Thank you for taking care of this issue!

Remi-Gau: I asked chatGPT to help me write XML because no one wants to that for sure. :-p

Remi-Gau (approved):

bthirion (approved):
LGTM, thx

PR #4555, opened 2024-09-20
[FIX] avoid warning when creating int64 images

Remi-Gau:
- Closes none

- fix plotting_roi and Parcellations that threw the warning even if users data with a type different from int64:

> Data array used to create a new image contains 64-bit ints. This is likely due to creating the array with numpy and passing `int` as the `dtype`. Many tools such as FSL and SPM cannot deal with int64 in Nifti images, so for compatibility the data has been converted to int32.
 
- Should remove > 80 warnings from the test suite
- warnings in the example:

https://nilearn.github.io/stable/auto_examples/03_connectivity/plot_data_driven_parcellations.html

image

https://nilearn.github.io/stable/auto_examples/01_plotting/plot_atlas.html#visualizing-the-juelich-atlas-with-contours

image

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -59,7 +59,7 @@ Changes
 
 - :bdg-dark:`Code` Improve logging by relying only on the Nilearn logger and adding optional support for rich printing if `rich <https://github.com/Textualize/rich>`_ is installed (:gh:`4469` and :gh:`4544` by `Rémi Gau`_).
 
-- :bdg-dark:`Code` Parcellations returned by :class:`nilearn.regions.Parcellations` will now be of type ``np.int32`` to avoid unnecessary warnings (:gh:`4555` by `Rémi Gau`).
+- :bdg-dark:`Code` Parcellations returned by :class:`nilearn.regions.Parcellations` will now be of type ``np.int32`` to avoid unnecessary warnings (:gh:`4555` by `Rémi Gau`_).
 
 - :bdg-danger:`Deprecation` The parameter ``tr`` for :term:`Repetition time<TR>` will be replaced by ``t_r`` in the "HRF" functions in version 0.13.0. The affected functions are :func:`nilearn.glm.first_level.glover_dispersion_derivative`, :func:`nilearn.glm.first_level.glover_hrf`, :func:`nilearn.glm.first_level.glover_time_derivative`, :func:`nilearn.glm.first_level.spm_dispersion_derivative`, :func:`nilearn.glm.first_level.spm_hrf`, :func:`nilearn.glm.first_level.spm_time_derivative` (:gh:`4470` by `Rémi Gau`_).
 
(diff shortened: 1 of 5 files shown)


bthirion (commented):
LGTM, thx.
Should we add a test to enforce the new behavior + maybe add a whatsnew entry ?

doc/changes/latest.rst
```diff
@@ -59,6 +59,8 @@ Changes
 
 - :bdg-dark:`Code` Improve logging by relying only on the Nilearn logger and adding optional support for rich printing if `rich <https://github.com/Textualize/rich>`_ is installed (:gh:`4469` and :gh:`4544` by `Rémi Gau`_).
 
+- :bdg-dark:`Code` Parcellations returned by :class:`nilearn.regions.Parcellations` will now be of type ``np.int32`` to avoid unnecessary warnings (:gh:`4555` by `Rémi Gau`).
```
Remi-Gau, line 62:
```suggestion
- :bdg-dark:`Code` Parcellations returned by :class:`nilearn.regions.Parcellations` will now be of type ``np.int32`` to avoid unnecessary warnings (:gh:`4555` by `Rémi Gau`_).
```

Remi-Gau:
> LGTM, thx. Should we add a test to enforce the new behavior + maybe add a whatsnew entry ?

Done

nilearn/plotting/tests/test_img_plotting/test_plot_roi.py
```diff
@@ -54,6 +54,12 @@ def test_plot_roi_view_types(
         cut_coords=cut_coords,
         **kwargs,
     )
+
+    for _ in range(len(recwarn)):
```
bthirion, line 58: I'm a bit puzzled. Why do we need this call at this place in particular ? Can you explain ?

Remi-Gau: This relates to the previous comment.

> Should we add a test to enforce the new behavior ?

The new behavior is to NOT throw a that specific warning so in each case, I loop through all the warning and check their type and message to make sure they are not of the one we want to avoid.

Remi-Gau: The recwarn is a pytest fixture to catch warnings.

I don't think we have a consistent way to check for this in the test suite.

bthirion: OK, I see. 
Maybe adding a comment in the code would be useful for people to understand why this was added ?

Remi-Gau: I actually put that in a separate test because there was not point in testing this with the parametrization.

bthirion (approved):
LGTM, thx.

Remi-Gau:
thanks for the reviews @bthirion @man-shu

PR #4559, opened 2024-09-24
[FIX] fix name of parameter in the "extract_type" definition

Remi-Gau:
- Closes none

- make sure the name of the extract_type parameter matches that of the key in the doc dictionary to avoid mismatch in the rendered doc (see example in image below where parameter name 'extract_type' in the function signature does not match the name of the parameter in the doc string 'extractor')

Screenshot_20240924_160042_Firefox


bthirion (commented):
Indeed !

Remi-Gau:
I will merge this one: seems small enough and fairly uncontroversial.

PR #4560, opened 2024-09-25
[ENH] use fsaverage mesh as default for the SurfaceImage

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/issues/4027#issuecomment-2368892520

- uses fsaverage5 pial mesh as default when no mesh is passed to the SurfaceImage constructor
-


Remi-Gau:
If we decided to go through with this, we would / could:
- simplify some examples
- need to set ˋmesh = Noneˋ in the constructor of SurfaceImage

PR #4574, opened 2024-09-29
[WIP][STY] apply ruff - DO NOT REVIEW YET

### LINKED ISSUE #4552: Consider replacing flake8, black, isort, pyupgrade... with ruff
https://docs.astral.sh/ruff/faq/#which-tools-does-ruff-replace

Could reduce the number of linting tools and speed up linting.

Remi-Gau:
- Closes #4552

- replaces black, flake8 and all its plugins, isort, pyugrade with ruff

This is work in progress.


Remi-Gau:
TODO
- reduce the number of line changes.

pyproject.toml
```diff
@@ -113,24 +104,160 @@ packages = ["nilearn"]
...
+docstring-code-line-length = "dynamic"
+indent-style = "space"
+line-ending = "auto"
+quote-style = "double"
+skip-magic-trailing-comma = false
+
+[tool.ruff.lint]
+dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
+fixable = ["ALL"]
+ignore = [
```
Remi-Gau, line 137: ignoring a lot of errors for now: to be fixed in follow up PR

nilearn/signal.py
```diff
@@ -11,7 +11,8 @@
 
 import numpy as np
 import pandas as pd
-from scipy import linalg, signal as sp_signal
+from scipy import linalg
+from scipy import signal as sp_signal
```
Remi-Gau, line 15: ruff treats these import a bit differently from isort so we get a few of those

.github/workflows/nightly_dependencies.yml
```diff
@@ -67,13 +67,6 @@ jobs:
             with:
                 python-version: 3.12
 
-        -   name: Install and run ruff checks for NPY NPY201
-            # see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#ruff-plugin
-            # TODO remove when nilearn only use numpy >= 2.0.0
-            run: |
-                python -m pip install ruff>0.1.8
-                ruff check --preview --select NPY201 .
-
```
Remi-Gau, line 76: not needed as it systematically applied now

man-shu (approved):
I think this qualifies as going through a deprecation cycle?
IIUC the main idea is to reduce the number of tools we use, right? The formatting is the same as with black: https://docs.astral.sh/ruff/faq/#is-the-ruff-linter-compatible-with-black

No comments otherwise.

Remi-Gau:
> I think this qualifies as going through a deprecation cycle?

Don't think it does: there is no change in behavior.

Maybe a mention in the changelog?

man-shu:
It would be a new dependency to be installed, but only for the devs, I think. So I guess only a changelog entry is fine too

Remi-Gau:
> It would be a new dependency to be installed, but only for the devs, I think. So I guess only a changelog entry is fine too

I think that now our pyproject.toml only indicates `tox`as a dependency for 'styling': all the other things should be handled automatically by pre-commit.

so:

tox ---> pre-commit ---> ruff / codespell

Remi-Gau:
> It would be a new dependency to be installed, but only for the devs, I think. So I guess only a changelog entry is fine too

This made me realized I had not update the doc for the dev: it should be fixed now.

bthirion (approved):
LGTM

Remi-Gau:
Thanks for the reviews @bthirion @man-shu 
I will open the follow up PRs right after.
Will try to keep them small and easy to review.

PR #4205, opened 2024-01-10
[ENH] Allow Decoder to work with surface objects

ymzayek:
- Part of #4027

Changes proposed in this pull request:

- [x] Add tests for decoder methods with surface object
- [x] Adapt screening percentile function
- [ ] Adapt FREM and ReNA clustering --> moved to #4205
- [x] raise NotImplemented (or similar) errors if users want to use SurfaceImage with
  - [x] space net
  - [ ] FREM (for now)
- Left untouched: searchlight

Issue #4538, opened 2024-09-09
use scikit-learn check_estimator to ensure compatibiity of Nilearn estimators with scikit-learn


Remi-Gau:
see https://scikit-learn.org/dev/developers/develop.html#rolling-your-own-estimator

quickly checked on a couple of Nilearn estimators and they seem to fail: will add code examples when in front of a real computer

may be worth discussing to see if we want to have 100% compliance or if we should have our own version of check_estimator

Remi-Gau:
Yup me too.
Will list the errors we get in this issue so we can  make a proper decision.

Remi-Gau:
one example that loops through all the checks

```python
from sklearn.utils.estimator_checks import check_estimator
from nilearn.decoding.decoder import (
    Decoder,
)

def test_check_estimator():
    """Check compliance with sklearn estimators."""
    model = Decoder()

    for estimator, check in check_estimator(estimator=model, generate_only=True):
        try:
            check(estimator)
        except Exception:
            print(f"FAIL: {check}")
        else:
            print(f"SUCCESS: {check}")
```

```
SUCCESS: functools.partial(<function check_no_attributes_set_in_init at 0x735a695ef240>, 'Decoder')
FAIL: functools.partial(<function check_estimators_dtypes at 0x735a695eca40>, 'Decoder')
FAIL: functools.partial(<function check_fit_score_takes_y at 0x735a695ec900>, 'Decoder')
FAIL: functools.partial(<function check_estimators_fit_returns_self at 0x735a695ee340>, 'Decoder')
FAIL: functools.partial(<function check_estimators_fit_returns_self at 0x735a695ee340>, 'Decoder', readonly_memmap=True)
FAIL: functools.partial(<function check_complex_data at 0x735a695e3420>, 'Decoder')
FAIL: functools.partial(<function check_dtype_object at 0x735a695e3380>, 'Decoder')
FAIL: functools.partial(<function check_estimators_empty_data_messages at 0x735a695ecc20>, 'Decoder')
FAIL: functools.partial(<function check_pipeline_consistency at 0x735a695ec7c0>, 'Decoder')
FAIL: functools.partial(<function check_estimators_nan_inf at 0x735a695ecd60>, 'Decoder')
FAIL: functools.partial(<function check_estimators_overwrite_params at 0x735a695ef100>, 'Decoder')
FAIL: functools.partial(<function check_estimator_sparse_array at 0x735a695e2b60>, 'Decoder')
[...]

PR #4577, opened 2024-09-30
[ENH] Allow to run clustering on SurfaceImage

Remi-Gau:
- Closes none but relates to #4027 
- Extracted from #4205 

- Add support for SurfaceImage ReNA


bthirion (commented):
Needs a few more tests but LGTM overall.

nilearn/regions/rena_clustering.py
```diff
@@ -153,6 +154,126 @@ def _make_edges_and_weights(X, mask_img):
...
+        edges_ = np.copy(edges_unmasked[:, edges_mask])
+
+        # Reorder the indices of the graph
+        max_index = edges_.max()
+        order = np.searchsorted(
+            np.unique(edges_.ravel()), np.arange(max_index + 1)
+        )
+        # increasing the order by the number of vertices in the previous mask
+        # to avoid overlapping indices
+        order += len_previous_mask
```
man-shu, line 269: @bthirion I made a change here because IIUC this comment https://github.com/nilearn/nilearn/pull/4205#discussion_r1461818983, we should not have any overlapping vertex indices between the left and right parts.

This wasn't the case before, see commit https://github.com/nilearn/nilearn/pull/4577/commits/ce895c53358106f5f490011584a4fd6f6d0a0584.

So just making sure I understand this correctly. Please LMK. Thanks!

nilearn/regions/tests/test_rena_clustering.py
```diff
@@ -104,3 +110,66 @@ def test_rena_clustering():
...
+
+    # edges and weights have two parts, left and right
+    assert len(edges) == 2
+    assert len(weights) == 2
+    for part in ["left", "right"]:
+        assert part in edges
+        assert part in weights
+
+    # make sure there are no overlapping indices between left and right parts
+    assert np.intersect1d(edges["left"], edges["right"]).size == 0
```
man-shu, line 162: testing no overlap between left and right hemisphere here

Remi-Gau:
> Pinging you for a review like this.

thanks for taking over!!!

looks good to me

Remi-Gau:
Actually do you think we should have some highter level tests (using ReNA with surface data) ?

Remi-Gau:
The doc string of recursive_neighbor_agglomeration should need some updating too, no?

bthirion (commented):
You probably want to have a test running Ward's clustering in the surface ?

nilearn/regions/tests/test_rena_clustering.py
```diff
@@ -104,3 +110,66 @@ def test_rena_clustering():
...
+    """
+    # make a new mask for this test
+    # the mask for left part has total 4 vertices out of which 3 are True
+    # and for right part it has total 5 vertices out of which 3 are True
+    data = {
+        "left": np.array([False, True, True, True]),
+        "right": np.array([True, True, False, True, False]),
+    }
+    surf_mask = SurfaceImage(surf_mesh(), data)
+    n_features = (
```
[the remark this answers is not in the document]
man-shu: More precisely, `n_features` is the number of vertices _after masking_.

In the FREM classifier/regressors, we always use ReNA after the masking. 
But in this case, we are creating a random array X, so it should not have n_vertices as in the mesh, but rather only the `True` vertices in the mask.

man-shu:
> Actually do you think we should have some highter level tests (using ReNA with surface data) ?

Yes indeed, will add more tests

nilearn/regions/rena_clustering.py
```diff
@@ -153,6 +155,126 @@ def _make_edges_and_weights(X, mask_img):
...
+
+    return edges.T, edges_masked
+
+
+def _make_edges_and_weights_surface(X, mask_img):
+    """Compute the weights to all edges in the mask.
+
+    Parameters
+    ----------
+    X : ndarray, shape = [n_samples, n_features]
```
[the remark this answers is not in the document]
man-shu: I don't think so.

X is supposed to be the data array where first dim is the timepoints or samples and the second one is the number of features or vertices.

nilearn/regions/rena_clustering.py
```diff
@@ -153,6 +155,126 @@ def _make_edges_and_weights(X, mask_img):
...
+    """Compute the weights for each edge using squared Euclidean distance.
+
+    Parameters
+    ----------
+    X : ndarray, shape = [n_samples, n_features]
+        Training data.
+
+    mask : boolean
+        Object used for masking the data.
+
```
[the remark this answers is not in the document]
man-shu: Samples are just the repetitions of observations or time points.

(Note that X here is a 2D array that we get after masking a SurfaceImage.)

Mask here is the initial boolean mask that was used to get X. So it would be False for vertices that should not be included and True for vertices that should be.

Meaning it would have the shape (1, n_vertices).

nilearn/regions/rena_clustering.py
```diff
@@ -153,6 +155,126 @@ def _make_edges_and_weights(X, mask_img):
...
+    weights : ndarray
+        Weights corresponding to all edges in the mask.
+        shape: (n_edges,).
+
+    """
+    n_samples, _ = X.shape
+    shape = mask.shape
+
+    data = np.empty((shape[0], n_samples))
+    for sample in range(n_samples):
```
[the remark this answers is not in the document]
man-shu: Yes, but `unmask_from_to_3d_array` is written such that it only takes one sample at a time. 

So `unmask_from_to_3d_array` could be improved in a separate PR

nilearn/regions/rena_clustering.py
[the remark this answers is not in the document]
Remi-Gau: - yeah but that will likely be flipped in https://github.com/nilearn/nilearn/pull/4591

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
```diff
@@ -128,3 +129,26 @@ def test_hierarchical_k_means_clustering():
...
+    # fit and transform the data
+    X_transformed = clustering.fit_transform(X)
+    # inverse transform the transformed data
+    X_inverse = clustering.inverse_transform(X_transformed)
+
+    # make sure the n_features in transformed data were reduced to n_clusters
+    assert X_transformed.shape[0] == n_clusters
+
+    # make sure the inverse transformed data has the same shape as the original
+    assert X_inverse.shape == X.shape
```
man-shu, line 154: > You probably want to have a test running Ward's clustering in the surface ?

Did you mean HierarchicalKmeansClustering? If so, now testing it here.

bthirion: Kmeans or the hierarchical version do not use the spatial structure (mesh). Ward does.

man-shu: While doing so, I noticed that `HierarchicalKMeans.fit()` takes transposed X:
https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/regions/hierarchical_kmeans_clustering.py#L233

See in `ReNA.fit()` for comparison:
https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/regions/rena_clustering.py#L518

Is this a bug or a feature?

nilearn/regions/rena_clustering.py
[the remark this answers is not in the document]
man-shu: yes, I remember discussing this with you, and sorry if you have already repeated this a few times, 

but now I am wondering why did we want to do that?

The `NiftiMasker` returns the data as n_samples x n_features:
https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/maskers/nifti_masker.py#L557

if we follow sklearn for such things, they also do the same:
https://github.com/scikit-learn/scikit-learn/blob/6e9039160f0dfc3153643143af4cfdca941d2045/sklearn/svm/_classes.py#L283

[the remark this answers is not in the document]
Remi-Gau: my bad: I was referring on how the data are stored in the SurfaceImage, not what transform returns. ignore my comment

nilearn/regions/tests/test_parcellations.py
```diff
@@ -310,3 +311,26 @@ def test_transform_3d_input_images(affine_eye):
...
+    # fit and transform the data
+    X_transformed = parcellate.fit_transform(X)
+    # inverse transform the transformed data
+    X_inverse = parcellate.inverse_transform(X_transformed)
+
+    # make sure the n_features in transformed data were reduced to n_clusters
+    assert X_transformed.shape[1] == n_parcels
+
+    # make sure the inverse transformed data has the same shape as the original
+    assert X_inverse.shape == X.shape
```
man-shu, line 336: @bthirion @Remi-Gau 

I think we forgot to fix [`Parcellations()`](https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/regions/parcellations.py#L120) class to work with surface data.

This test will fail because of that.

nilearn/regions/rena_clustering.py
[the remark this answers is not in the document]
man-shu: I see, so our maskers would still return the arrays in n_samples x n_features format?

nilearn/regions/tests/test_parcellations.py
Remi-Gau: ouch
do you want to look into it?

man-shu: Yes doing that right now

man-shu: This Parcellations class does masking before running the dimensionality reduction, and that is where we have the issue.

I think we need to update the [`_BaseDecomposition`]( https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/decomposition/_base.py#L247) object for this. This would involve implementing a new `SurfaceMapsMasker` object.

As discussed with @Remi-Gau, we could just merge the current PR because that would allow all Decoders to work with surface data. The current PR fixes the FREM classifier and regressor specifically.

So I could drop a NotImplemented error in case we or some users think that Parcellations class is supported, and then create a separate PR for Parcellations

man-shu: Whether or not we want to wait for the new `SurfaceMapsMasker` object before doing the next release is an open question for discussion...

man-shu: Ok, so I will just start the PR to create `SurfaceMapsMasker` and a separate PR for allowing Parcellations to work with surface API.

man-shu:
Just waiting for the CI to complete, otherwise ready for review

nilearn/regions/tests/test_parcellations.py
Remi-Gau: I would prefer to keep the scope as it was for the coming release:
- get rid of experimental
- do not add new maskers
- therefore let's not support Parcellation on SurfaceImage

I would prefer having another release soonish after that to implement those new features and fix all the bugs that will come with users starting to use 0.11.

man-shu: I don't feel strongly about whether or not to delay the release to wait for `SurfaceMapsMasker`.
In any case, if we are ok with the reviews, let's merge this PR.

As @Remi-Gau suggested previously, I added `NotImplementedError` when trying to use surface data with Parcellations.

nilearn/decomposition/_base.py
```diff
@@ -382,6 +383,12 @@ def __init__(
         self.n_jobs = n_jobs
         self.verbose = verbose
 
+        # Does not support surface-based images yet; See #4756 for updates.
+        if isinstance(self.mask, (SurfaceMasker, SurfaceImage)):
+            raise NotImplementedError(
+                "Surface-based images are not yet supported by this module."
+            )
+
```
man-shu, line 391: Raise an error if surface mask is used with Parcellations

Remi-Gau: can you put this check in the fit method?
scikit learn convention says we should not have input validation in the constructor: https://scikit-learn.org/stable/developers/develop.html#instantiation

it is a rule we have not applied systematically but I intend to start fixing things progressively on this front, so let's at least try to stick to those conventions in the new code we add

Remi-Gau: - see https://github.com/nilearn/nilearn/issues/4538

man-shu: Done!

nilearn/decomposition/_base.py
```diff
@@ -434,6 +435,20 @@ def fit(
...
+        # Does not support surface-based images yet; See #4756 for updates.
+        # check if the mask is a surface-based masker or surface image
+        if isinstance(self.mask, (SurfaceMasker, SurfaceImage)):
+            raise NotImplementedError(
+                "Surface-based images are not yet supported by this module."
+            )
+        # check if the images are surface-based images
+        for img in imgs:
+            if isinstance(img, SurfaceImage):
```
Remi-Gau, line 447:
```suggestion
        if any(isinstance(x, SurfaceImage) for x in imgs):
```

Remi-Gau: Can probably even merge the 2 if block

man-shu: Done!

PR #4604, opened 2024-10-07
Return signals if standardisation was not applied, and change the condition when adding signal mean back for psc method

### LINKED ISSUE #4466: signal.clean adds mean when using Butterworth filter

### Operating system

- [ ] Linux
- [X] Mac
- [ ] Windows

### Operating system version

- Mac OS Version 14.5 "Sonoma"

### Python version

- [ ] 3.12
- [X] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

0.10.4

### Expected behavior

Using signal.clean with only lowpass filtering the mean of the signal should not change.

Signal.clean now adds the mean to the signal if filtered with Butterworth, ending up with a signal twice the mean of the original signal when using lowpass filtering.

```python
from nilearn.signal import clean
import numpy as np

y = np.random.random(100) + 10
y_lowpass = clean(y, tr = 1, low_pass=0.1, detrend=False, standardize=False)
y_highpass = clean(y, tr = 1, high_pass=0.1, detrend=False, standardize=False)

print(y.mean(), y_lowpass.mean(), y_highpass.mean())
```

htwangtw:
This ended up being a pretty chunky PR.

- Closes #4466 

- Return signals if standardization was not applied. This is straight forward.
- Extra tests for PSC - make sure at least the output is correlated with those created by other linear scaling method (z score).
- Tighten the logic for adding back mean to signals.

Instead of checking user supplied strategy, I check if the original signal mean has a factor of 10 difference after filtering / detrending / regressing confounds. This absolutely would be triggered by `detrend` as the mean is usually 1-e16. And for high-pass filtered signal, the mean is usually 1-e2 - 1-e3, so it get triggers too. I added a warning when this is happening, and also inform users that if the original signal mean is too small, the scale would be weird.

Let me know if you think this is a good solution before I add things to the change log.

### DIFF
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -746,7 +746,7 @@ def clean(
     # Detrend and filtering should apply to confounds, if confound presents
     # keep filters orthogonal (according to Lindquist et al. (2018))
     # Restrict the signal to the orthogonal of the confounds
-    mean_signals = signals.mean(axis=0)
+    original_mean_signals = signals.mean(axis=0)
     if detrend:
         signals = standardize_signal(
             signals, standardize=False, detrend=detrend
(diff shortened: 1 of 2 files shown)


bthirion (commented):
You probably want to add a test to enforce this behavior ?

nilearn/signal.py
```diff
@@ -806,12 +806,30 @@ def clean(
...
+        # calculate the psc to avoid weird scaling.
+        warnings.warn(
+            "After cleaning, the new mean of the signal is 1 factor of 10 "
+            "smaller than the original signal. This can obscure the scale of "
+            "the signal with percent signal change standardization. To avoid "
+            "the scale of the signal being obscured drassically, the original "
+            "signal mean has been added back for percent signal change "
+            "standardization. Please note that this might not always solve "
+            "the scale issue if the original signal has a small mean."
+        )
```
bthirion, line 830: I'm a bit puzzled here. under which conditions ca such a shrinkage of the mean signal occur ?

bthirion: If there is no reason to keep it I'd rather remove it.

bthirion (approved):
LGTM.

Remi-Gau (approved):
Could not spot anything. Let's merge.

Issue #4327, opened 2024-03-19
Extend/improve support for pathlib.Path


Remi-Gau:
> [!TIP]
> # **💻 🍁  For the [hacktoberfest](https://hacktoberfest.com/) contributors 🍁 💻**
>
> Make sure to:
> - read our [contributing guidelines](https://nilearn.github.io/stable/development.html#contribution-guidelines)
>
> If this is your first time contributing to Nilearn, add your details in the [CITATION.cff file](https://github.com/nilearn/nilearn/blob/dcd92aa357ab3a751161aeea2edfa3a775ae644f/CITATION.cff#L42).
>
> For your pull request to be valid, you should fix some of [the "path" errors flagged by the ruff checker](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth).
>
> Steps to follow:
> 1. delete one of the ignored `PTH` error from [the ruff configuration in the pyproject.toml](https://github.com/nilearn/nilearn/blob/dcd92aa357ab3a751161aeea2edfa3a775ae644f/pyproject.toml#L136) by deleting it.
> 2. install pre-commit with `pip install pre-commit` and install the pre-commit "hooks" with `pre-commit install`
> 3. let pre-commit run ruff and autofix as many errros as possible with: `pre-commit run -a`
> 4. manually fix the remaining errors

List and number of error instances to fix

- [x] 213    `PTH118`  [ ] os-path-join - assigned to @michellewang 
- [x] 108    `PTH123`  [ ] builtin-open
- [x]  62    `PTH110`  [ ] os-path-exists - assigned to @PrakharJain1509
- [x]  35    `PTH120`  [ ] os-path-dirname - assigned to @hndgzkn 
- [x]  30    `PTH107`  [ ] os-remove - assigned to @PrakharJain1509
- [x]  18    `PTH102`  [ ] os-mkdir - assigned to @PrakharJain1509
- [x]  18    `PTH113`  [ ] os-path-isfile - assigned to @PrakharJain1509
- [x]  17    `PTH100`  [ ] os-path-abspath - assigned to @hndgzkn 
- [x]  12    `PTH119`  [ ] os-path-basename - assigned to @hndgzkn 
- [x]  10    `PTH103`  [ ] os-makedirs - assigned to @hndgzkn 
- [x]   9    `PTH111`  [ ] os-path-expanduser - assigned to @PrakharJain1509
- [ ]   6    `PTH207`  [ ] glob
[...]

PR #4606, opened 2024-10-07
[STY] Fix `PTH118` for all files except in `nilearn.datasets`

michellewang:
- Addresses #4327

- ~Remove `PTH118` from ruff ignore list~ putting it back because not all errors are handled in this PR
- Find-and-replace `os.path.join` to `Path` from `pathlib` module, then manual checking and optimizing things
- Note: `nilearn.plotting.glass_brain.py` has `from matplotlib.path import Path`, so for this one I did `import pathlib` and used `pathlib.Path`


michellewang:
@Remi-Gau if there were functions that returned paths as strings (for example as values in a dictionary), is it right to assume that we would still want to return strings in these functions? Otherwise there would be a lot of tests to change too

Remi-Gau:
Yes indeed.

If they are part of the public API let's not change their behavior for now. Maybe later we can think about it.

Otherwise it may be fine to have them return Path but that may affect a lot of other things: maybe check how often it is called before making the change.

In any case if you can keep track of those functions that return strings where path could go, make a list. In this issue or another one.
This way we can go through it later and decide what to do.

michellewang:
@Remi-Gau I am thinking of splitting this PR into multiple ones because for some modules there are quite a lot of fixes to do to make all tests pass. So maybe one PR for the easy cases and then one for each module that requires more work. That would make things easier to review. What do you think?

michellewang:
Ok so most of the tests were easy to fix but I still think it's worth splitting off at least `nilearn/datasets/func.py` and `nilearn/datasets/tests/test_neurovault.py`. And so far I just tried to make the tests pass but I still need to spend some time and make docstring changes where appropriate

Remi-Gau:
> it's worth splitting off at least `nilearn/datasets/func.py` and `nilearn/datasets/tests/test_neurovault.py`

yeah I would put the changes for the `datasets` subpackage in a different PR.

bthirion (approved):
The PR LGTM.

michellewang:
I have removed all changes to `nilearn.datasets` files and cleaned up the rest, so this should be ready to go as long as the tests pass. The `pre-commit` check is failing now because of the remaining `PTH118` errors in `nilearn.datasets`. So I think I will add back `PTH118` to the `pyproject.toml` ignore list for now.

nilearn/glm/tests/test_first_level.py
```diff
@@ -1831,6 +1832,16 @@ def test_check_trial_type_warning(tmp_path):
...
+def test_list_valid_subjects_with_toplevel_files(tmp_path):
+    """Test that only subject directories are returned, not file names."""
+    (tmp_path / "sub-01").mkdir()
+    (tmp_path / "sub-02").mkdir()
+    (tmp_path / "sub-01.html").touch()
+
+    valid_subjects = _list_valid_subjects(tmp_path, None)
+    assert valid_subjects == ["01", "02"]
+
+
```
michellewang, line 1844: Adding this test because at one point the changes I had made caused the `visual_testing/reporter_visual_inspection_suite.py` script to fail during documentation building, and there wasn't a test to catch that error

michellewang:
Also just wanted to note that I went ahead and fixed some of the other codes too (`open`, `os.path.abspath`, `os.path.dirname`, `glob.glob`) when they made things easier

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -230,9 +230,9 @@ def _get_file_name(nii_file):
     # https://www.geeksforgeeks.org/python-sort-list-of-lists-by-the-size-of-sublists/
     confound_file_candidates = sorted(confound_file_candidates, key=len)[::-1]
     confound_file_candidates = [
-        os.path.join(base_dir, crc) for crc in confound_file_candidates
+        base_dir / crc for crc in confound_file_candidates
     ]
-    found_files = [cr for cr in confound_file_candidates if os.path.isfile(cr)]
+    found_files = [str(cr) for cr in confound_file_candidates if cr.is_file()]
```
michellewang, line 235: Do we want to return a `Path` here since this function is private?

Remi-Gau:
> Also just wanted to note that I went ahead and fixed some of the other codes too (`open`, `os.path.abspath`, `os.path.dirname`, `glob.glob`) when they made things easier

yes make sesne

Remi-Gau:
> So I think I will add back `PTH118` to the `pyproject.toml` ignore list for now.

Otherwise you can 'file ignore' that error for a certain subpackage or module.

nilearn/glm/tests/test_first_level.py
Remi-Gau: nice, thanks

nilearn/externals/tempita/__init__.py
Remi-Gau, line 1: Wondering if we should not leave the `externals` untouched and therefore maybe ignore all errors in that part of the code base?

nilearn/interfaces/fmriprep/load_confounds_utils.py
Remi-Gau: I would say so: though once again, depends how many of our things break.

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -271,7 +271,7 @@ def get_confounds_file(image_file, flag_full_aroma):
 def get_json(confounds_raw_path):
     """Return json data companion file to the confounds tsv file."""
     # Load JSON file
-    return confounds_raw_path.replace("tsv", "json")
+    return str(confounds_raw_path).replace("tsv", "json")
```
Remi-Gau, line 274: get_json is not part of the user facing API so this one as well could return path

Remi-Gau (approved):
This is good with me.
I think I'd rather merge this as is and maybe tackle some of my minor comments in other small follow up PR if we think they are worth it.

PR #4607, opened 2024-10-08
[FIX] fix error PTH106 

hndgzkn:
- Closes  none but relates to #4327 

- fixes error: https://docs.astral.sh/ruff/rules/os-rmdir/

### DIFF
--- a/nilearn/datasets/_utils.py
+++ b/nilearn/datasets/_utils.py
@@ -698,9 +698,14 @@ def get_dataset_descr(ds_name):
 
 
 def movetree(src, dst):
-    """Move an entire tree to another directory.
+    """Move entire tree under `src` inside `dst`.
+
+    Creates `dst` if it does not already exist.
 
     Any existing file is overwritten.
+
+    The difference with `shutil.mv` is that `shutil.mv` moves `src` under `dst`
+    if `dst` already exists.
     """
     src = Path(src)
 
--- a/nilearn/datasets/_utils.py
+++ b/nilearn/datasets/_utils.py
@@ -702,24 +702,24 @@ def movetree(src, dst):
 
     Any existing file is overwritten.
     """
-    names = os.listdir(src)
+    src = Path(src)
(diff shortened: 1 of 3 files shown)


nilearn/datasets/tests/test_utils.py
```diff
@@ -244,43 +244,45 @@ def test_tree():
...
+    dir2 = tmp_path / "dir2"
+    dir212 = dir2 / "dir12"
+
+    dir1.mkdir()
+    dir111.mkdir()
+    dir112.mkdir()
+    dir2.mkdir()
+    dir212.mkdir()
+
+    open(dir1 / "file11", "w").close()
```
Remi-Gau, line 261:
```suggestion
    (dir1 / "file11").touch()
```

maybe double check but I think that Path has a `touch` method that should allow to simplify those open/close even further.

nilearn/datasets/_utils.py
```diff
@@ -702,24 +702,24 @@ def movetree(src, dst):
 
     Any existing file is overwritten.
     """
-    names = os.listdir(src)
```
Remi-Gau, line 705: I am wondering if we even need this movetree function.

Not sure how it differs
https://docs.python.org/3/library/shutil.html#shutil.move

@hndgzkn can you double check and maybe update the doc string to explain why it is different?

Remi-Gau: thanks a bunch: I and future maintainers thanks you for the clarification.

hndgzkn: I hope it is clear enough.

Remi-Gau (approved):
looks good to me: will @michellewang cross check

bthirion (approved):
LGTM, thx !

michellewang (approved):
Looks good to me as well. There will probably be a merge conflict with #4606 in `test_utils.py` but it should be easy to resolve :)

PR #4612, opened 2024-10-08
[FIX] fix several errors (PTH112, PTH114, PTH117)

hndgzkn:
- Closes  none but relates to #4327 

fixes errors:
- PTH 112: https://docs.astral.sh/ruff/rules/os-path-isdir/
- PTH 114: https://docs.astral.sh/ruff/rules/os-path-islink/
- PTH 117: https://docs.astral.sh/ruff/rules/os-path-isabs/

I see that PTH 114 and 117 are in the scope of https://github.com/nilearn/nilearn/pull/4613 . As it overlaps with the functions that I work on, I have refactored them as well.

I've removed completely `readlinkabs` as I saw that it was used only once in the code and it can be replaced by `Path.resolve`.


nilearn/datasets/neurovault.py
```diff
@@ -1407,20 +1419,20 @@ def _download_collection(collection, download_params):
...
-    )
+    collection_dir = Path(download_params["nv_data_dir"]) / collection_name
     collection["relative_path"] = collection_name
-    collection["absolute_path"] = collection_dir
-    if not os.path.isdir(collection_dir):
-        os.makedirs(collection_dir)
-    metadata_file_path = os.path.join(
-        collection_dir, "collection_metadata.json"
-    )
+    collection["absolute_path"] = str(collection_dir.absolute())
```
Remi-Gau, line 1428: Am wondering if we have to cast to string here given this is a private function.
Does a lot of downstream things ion the code break if we stick to Path?

nilearn/datasets/tests/test_utils.py
```diff
@@ -203,45 +203,55 @@ def test_read_md5_sum_file():
     os.remove(f)
 
 
-def test_tree():
-    # Create a dummy directory tree
-    parent = mkdtemp()
+def test_tree(tmp_path):
```
Remi-Gau, line 206: the test is definitely easier to read now!

nilearn/datasets/neurovault.py
hndgzkn: The tests do not pass if I use Path. 

I've checked if it will be straightforward. But I do not think so.  

It might be easier to come back to this after all PTH errors in https://github.com/nilearn/nilearn/issues/4327 are solved. 

I suggest to create an issue for that to come back later.

Remi-Gau: I agree let's flag that in another issue to tackle later

Remi-Gau (commented):
Do you think you can add an entry in the changelog and add yourself as a contributor?

I think we sill most likely have a single line in the changelog for all the path changes and list all the PRs and contributors.

hndgzkn:
> Do you think you can add an entry in the changelog and add yourself as a contributor?
> 
> https://nilearn.github.io/stable/development.html#changelog
> 
> I think we sill most likely have a single line in the changelog for all the path changes and list all the PRs and contributors.

Shall I do it with this PR?

Remi-Gau:
> Shall I do it with this PR?

yup let's do that

bthirion (commented):
There remains one untested line...

hndgzkn:
> There remains one untested line...

Added!

bthirion (approved):
LGTM, thx.

michellewang (approved):
Looks good to me too! I agree with the comment about `nilearn.datasets.neurovault` being tricky to refactor. Will try to work on that once I extract that part from #4606

Remi-Gau:
Thanks @hndgzkn @bthirion @michellewang for the PR and reviews

PR #4630, opened 2024-10-11
Test for screening adjustment

### LINKED ISSUE #4626: Test screening percentile adjustment via `nilearn._utils.param_validation.adjust_screening_percentile`
I wanted to test the screening percentile adjustment somewhat like this. However, I am struggling to create an appropriate sample mask and image.

I saw that there are some fixtures defined in `nilearn/experimental/conftest.py` but I couldn't call them here (even after importing them in `nilearn/conftest.py`).

_Originally posted by @man-shu in https://github.com/nilearn/nilearn/pull/4205#discussion_r1795560569_

man-shu:
- Closes #4626 

- test if screening percentile is adjusted properly during decoding, enforcing following behavior
  - if mask size is smaller than screening percentile wrt to the mesh size, use whole mask
  - if mask size is greater than screening percentile wrt to the mesh size, adjust the screening percentile for that mask size

### DIFF
--- a/nilearn/decoding/tests/test_decoder.py
+++ b/nilearn/decoding/tests/test_decoder.py
@@ -1160,9 +1160,10 @@ def test_decoder_screening_percentile_surface(perc, _make_surface_class_data):
 def test_decoder_screening_percentile_adjustment_surface(
     _make_surface_mask, _make_surface_class_data, screening_percentile
 ):
-    """Test adjustment of screening percentile, such that if mask size is
-    smaller than screening percentile wrt to the mesh size, use whole mask or
-    if mask size is greater than screening percentile wrt to the mesh size,
+    """Test adjustment of screening percentile, such that:
+    - if mask size is smaller than screening percentile wrt to the mesh size,
+    use whole mask.
+    - if mask size is greater than screening percentile wrt to the mesh size,
     adjust the screening percentile for that mask size.
     """
     mask = _make_surface_mask()
--- a/nilearn/decoding/tests/test_decoder.py
+++ b/nilearn/decoding/tests/test_decoder.py
@@ -1167,7 +1167,7 @@ def test_decoder_adjust_screening_lessthan_mask_surface(
     mask_n_vertices = get_mask_volume(mask)
     mesh_n_vertices = img.mesh.n_vertices
     mask_to_mesh_ratio = (mask_n_vertices / mesh_n_vertices) * 100
-    assert mask_to_mesh_ratio <= screening_percentile
+    assert screening_percentile <= mask_to_mesh_ratio
     decoder = Decoder(
         mask=mask,
         param_grid={"C": [0.01, 0.1]},
@@ -1193,7 +1193,7 @@ def test_decoder_adjust_screening_greaterthan_mask_surface(
     mask_n_vertices = get_mask_volume(mask)
     mesh_n_vertices = img.mesh.n_vertices
     mask_to_mesh_ratio = (mask_n_vertices / mesh_n_vertices) * 100
-    assert mask_to_mesh_ratio > screening_percentile
+    assert screening_percentile > mask_to_mesh_ratio
     decoder = Decoder(
         mask=mask,
         param_grid={"C": [0.01, 0.1]},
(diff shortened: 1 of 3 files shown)


nilearn/decoding/tests/test_decoder.py
```diff
@@ -1144,6 +1156,37 @@ def test_decoder_screening_percentile_surface(perc, _make_surface_class_data):
...
+        param_grid={"C": [0.01, 0.1]},
+        cv=3,
+        screening_percentile=screening_percentile,
+    )
+    with pytest.warns(UserWarning, match="Consider raising"):
+        decoder.fit(img, y)
+    adjusted = decoder.screening_percentile_
+    if (
+        screening_percentile == 100
+        or mask_to_mesh_ratio < screening_percentile
```
bthirion, line 1179: I guess that you could simply replace both conditions by  `mask_to_mesh_ratio =< screening_percentile`

bthirion (commented):
LGTM, thx.

nilearn/decoding/tests/test_decoder.py
man-shu: oops yes indeed

bthirion (approved):
LGTM, thx !

nilearn/experimental/conftest.py
```diff
@@ -60,6 +60,24 @@ def f(shape=()):
     return f
 
 
+@pytest.fixture
+def make_mini_mask(mini_mesh) -> Callable:
```
Remi-Gau, line 64: Not sure why you are creating another mask fixture, but I'd say that when 'merging' the experimental module in the stable API we should clean up our test set up because I think we have 3 or 4 different surface mask fixture and I am not sure we need that many.

Fine for this PR though.

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1144,6 +1156,40 @@ def test_decoder_screening_percentile_surface(perc, _make_surface_class_data):
...
+):
+    """Test adjustment of screening percentile, such that:
+    - if mask size is smaller than screening percentile wrt to the mesh size,
+    use whole mask.
+    - if mask size is greater than screening percentile wrt to the mesh size,
+    adjust the screening percentile for that mask size.
+    """
+    mask = _make_surface_mask()
+    img, y = _make_surface_class_data()
+    mask_n_vertices = get_mask_volume(mask)
```
Remi-Gau, line 1171:
```
    mask_n_vertices = get_mask_volume(mask)
```

we will have to rename this private function or split its behavior because something called get_mask_volume that returns a number of vertices is confusing

nilearn/experimental/conftest.py
man-shu: Yes it does look like we need better fixtures for creating sample surface images.

I was super confused as well because I could not import any of the existing fixtures for making sample surface masks. 

Then I noticed the same thing with sample surface images, and there was an existing `make_mini_img` being used repeatedly. So I created a similar one for masks.

nilearn/decoding/tests/test_decoder.py
man-shu: Ok I will create separate issues for this

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1144,6 +1156,40 @@ def test_decoder_screening_percentile_surface(perc, _make_surface_class_data):
...
+@pytest.mark.parametrize("screening_percentile", [30, 80, 100])
+def test_decoder_screening_percentile_adjustment_surface(
+    _make_surface_mask, _make_surface_class_data, screening_percentile
+):
+    """Test adjustment of screening percentile, such that:
+    - if mask size is smaller than screening percentile wrt to the mesh size,
+    use whole mask.
+    - if mask size is greater than screening percentile wrt to the mesh size,
+    adjust the screening percentile for that mask size.
+    """
```
Remi-Gau, line 1168: Am gonna be annoying but do you think you could split that in 2 different tests: one for each condition.

Given that the number of vertices is hard coded in the fixture, it is hard from reading the test to know when `mask_to_mesh_ratio <= screening_percentile` or not.

nilearn/experimental/conftest.py
Remi-Gau: yeah the first thing we will most likely do when merging the new API will be clean up the tests in general: also because some fixture names will need to be clarified to avoid clashing with the volume fixture.

nilearn/decoding/tests/test_decoder.py
man-shu: Done!

Remi-Gau (approved):

man-shu:
Merging then!

Issue #4579, opened 2024-10-02
Rerformating with Ruff


Remi-Gau:
- Started with https://github.com/nilearn/nilearn/pull/4574,there should be a few simple follow things todo with small PRs.

I am listing them here for more transparency and so that things are not just in my head.

- [x] apply ruff to files that were not covered yet (`nilearn/_utils/docs.py` and `nilearn/datasets/neurovault.py`) - see https://github.com/nilearn/nilearn/pull/4578
- [x] apply some doc rules that are ignored for now (notably D209: https://docs.astral.sh/ruff/rules/new-line-after-last-paragraph/) --> https://github.com/nilearn/nilearn/pull/4583
- [x] remove the PTH errors (https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth) from the ignore list and fix them - see also https://github.com/nilearn/nilearn/issues/4327

After this I think, some more rules could be removed from the ignored list and autofixed by running `pre-commit run -a` and see how many files are changed and what the changes are.

I would probably start looking into:
- [x] numpy errors: https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy
- [x] pandas errors: https://docs.astral.sh/ruff/rules/#pandas-vet-pd
- [x] performance errors: https://docs.astral.sh/ruff/rules/#perflint-perf
- [x] simplify errors: https://docs.astral.sh/ruff/rules/#flake8-simplify-sim

### List of errors

| status | number of errors | error code | autofixable | error name | 
| -  | -  |  -   | - | - | 
| - [ ] | 560  |   `N806`    | [ ] | non-lowercase-variable-in-function | 
| - [ ] | 184  |   `ERA001`  | [*] | commented-out-code | 
| - [ ] | 164  |   `PLR2004` | [ ] | magic-value-comparison | 
| - [ ] | 135  |   `B028`    | [ ] | no-explicit-stacklevel | 
| - [ ] | 115  |   `N803`    | [ ] | invalid-argument-name | 
| - [ ] |  17  |   `RUF012`  | [ ] | mutable-class-default | 
| - [ ] |  34  |   `B904`    | [ ] | raise-without-from-inside-except | 
| - [ ] |   8  |   `D105`    | [ ] | undocumented-magic-method | 
| - [ ] |   5  |   `E741`    | [ ] | ambiguous-variable-name | 
| - [ ] |   3  |   `ARG005`  | [ ] | unused-lambda-argument | 
[...]

PR #4644, opened 2024-10-15
[STY] fix PLR1704 (redefined-argument-from-local)

Remi-Gau:
- Closes none but relates #4579

- fixes https://docs.astral.sh/ruff/rules/redefined-argument-from-local/


bthirion (commented):
I find the `session` -> `sess` particularly horrible. What is the reason for that ?

man-shu:
> I find the `session` -> `sess` particularly horrible. What is the reason for that ?

Looking at the doc for this rule: https://docs.astral.sh/ruff/rules/redefined-argument-from-local/
It's because `session` is a parameter in the function, and the for loop was (mistakenly) redefining it.

man-shu (approved):
LGTM!

bthirion (approved):

PR #4653, opened 2024-10-16
[MAINT] always run tests in random order

### LINKED ISSUE #4647: always run tests in random order
Currently we have one workflow that runs once a month to run our test suite in a random order to potentially detect test pollution:

https://github.com/nilearn/nilearn/actions/workflows/detect_test_pollution.yml

test pollution: dependency between tests where test B passes if run after test A because A changes some state that makes B pass. If run in a the order B then A then B would fail.

- When working on the test suite (for example https://github.com/nilearn/nilearn/issues/3570) it may takes us a month to realize that we have added dependency between our tests.

I would suggest getting rid of the monthly workflow and always run tests with pytest-random-order

https://github.com/nilearn/nilearn/blob/2e25896c4d05ab4521293acc74557aa01111482a/tox.ini#L138-L146

Remi-Gau:
- Closes #4647 

- add pytest-random-order as one of the optional test dependency and always use it when running tests
- remove the montly workflow to run tests in a random order


.github/workflows/README.md
```diff
@@ -136,15 +136,9 @@ using this issue [template](../nightly_failure.md).
...
 
-### detect_test_pollution.yml
-
-Runs once a month.
-Use pytest with the pytest-random-order plugin to run all tests in a random order.
-This aims to detect tests that are not properly isolated from each other (test pollution).
-
 ### testing_minimum.yml
 
-Runs once a month.
```
Remi-Gau, line 147: needed updating

.github/workflows/README.md
```diff
@@ -136,15 +136,9 @@ using this issue [template](../nightly_failure.md).
...
+Runs and push to main and once a month.
```
man-shu, line 141:
```suggestion
Runs and pushes to main, once a month.
```

man-shu (approved):
LGTM!

.github/workflows/README.md
Remi-Gau: nope

but let me try to rephrase to see if this is clearer

Remi-Gau, line 141:
```suggestion
This workflow is triggered when a new commit is pushed to the main branch (or when a pull request is merged) and is also automatically run once a month.
```

Remi-Gau: @man-shu is this clearer?

man-shu: Oh I see! 
Yes this is clearer indeed

PR #4658, opened 2024-10-16
[REF] use path for resolve globbing

Remi-Gau:

- make resolve_globbing return list of Paths and adapt some of the downstream code


nilearn/_utils/numpy_conversions.py
```diff
@@ -157,7 +158,7 @@ def csv_to_array(csv_path, delimiters=" \t,;", **kwargs):
         An array containing the data loaded from the CSV file.
     """
     csv_path = stringify_path(csv_path)
-    if not isinstance(csv_path, str):
+    if not isinstance(csv_path, (str, pathlib.Path)):
```
Remi-Gau, line 161: need to think if that makes sense as stringify_path will always return a str if passed a Path...

build.txt
Remi-Gau, line 1: remove

log.txt
Remi-Gau, line 1: remove

nilearn/decomposition/tests/test_canica.py
```diff
@@ -280,6 +280,19 @@ def test_with_globbing_patterns_with_single_subject(mask_img, tmp_path):
     assert components_img.shape, check_shape
 
 
+def test_with_globbing_patterns_with_single_subject_path(mask_img, tmp_path):
```
Remi-Gau, line 283: we may need to add more checks like this that our public API can handle Path objects

Remi-Gau:
@hndgzkn @michellewang 
this is Path related so bringing this one to your attention

nilearn/_utils/tests/test_helpers.py
```diff
@@ -158,3 +159,7 @@ def test_is_plotly_installed():
 
 def test_is_kaleido_installed():
     helpers.is_kaleido_installed()
+
+
+def test_stringify_path():
+    assert isinstance(helpers.stringify_path(Path("foo") / "bar"), str)
```
hndgzkn, line 165:
```suggestion
    assert isinstance(helpers.stringify_path(Path("foo") / "bar"), str)
    assert helpers.stringify_path([]) == []
```

hndgzkn: Shall we add a test with another object to make sure that it does not change the object's type and returns it as it is?

Actually the docstring can also be updated and made more clear for `stringify_path` as it does not always return `str`.

nilearn/_utils/tests/test_path_finding.py
```diff
@@ -0,0 +1,42 @@
+from pathlib import Path
+
+import pytest
+
+from nilearn._utils.path_finding import resolve_globbing
+
+
+def test_resolve_globbing(tmp_path):
+    assert resolve_globbing(tmp_path) == [tmp_path]
```
hndgzkn, line 9:
```suggestion
    assert resolve_globbing(tmp_path) == [tmp_path]
    assert resolve_globbing([]) == []
```

nilearn/_utils/tests/test_helpers.py
hndgzkn: I am not sure if this function is really necessary.

nilearn/_utils/tests/test_path_finding.py
hndgzkn: I would put the same type of test here, to check the case when path is neither `str` nor `path`.

The doctring can be updated as well as in the case of `stringify_path`. Or alternatively this method could accept only `str` or `Path` and raise error otherwise.

nilearn/decomposition/_base.py
```diff
@@ -417,7 +418,7 @@ def fit(
         ):
             imgs = resolve_globbing(imgs)
 
-        if isinstance(imgs, str) or not hasattr(imgs, "__iter__"):
+        if isinstance(imgs, (str, Path)) or not hasattr(imgs, "__iter__"):
```
hndgzkn, line 421: The docstring tells that `imgs` should be "list of Niimg-like objects". If we are expecting `str` or `Path` as well, it might be good to update docstring accordingly.

nilearn/glm/tests/test_first_level.py
```diff
@@ -1995,7 +1995,8 @@ def test_check_run_tables_errors():
     with pytest.raises(ValueError, match="table path .* could not be loaded"):
         _check_run_tables([""] * 2, [".csv", ".csv"], "")
     with pytest.raises(
-        TypeError, match="can only be a pandas DataFrames or a string"
+        TypeError,
+        match="only be a pandas DataFrame, a Path object or a string",
```
hndgzkn, line 1999:
```suggestion
        match="can only be a pandas DataFrame, a Path object or a string",
```

nilearn/surface/surface.py
```diff
@@ -881,16 +881,16 @@ def load_surf_data(surf_data):
 
             if len(data_part.shape) == 1:
                 data_part = data_part[:, np.newaxis]
-            if f == 0:
+            if i == 0:
                 data = data_part
-            elif f > 0:
+            elif i > 0:
```
hndgzkn, line 886:
```suggestion
            else:
```

nilearn/decomposition/_base.py
Remi-Gau: We don't have typing in nilearn but as far as I understand if we did we would have

```python
niimg_like_object: nibabel.nifti1.Nifti1Image | str | pathlib.Path
```

See : https://nilearn.github.io/stable/manipulating_images/input_output.html#file-names-and-objects-3d-and-4d-images

hndgzkn: I meant the below part of the code for `fit` function.

```
        Parameters
        ----------
        imgs : list of Niimg-like objects
            See :ref:`extracting_data`.
```
which displays https://nilearn.github.io/stable/modules/generated/nilearn.decomposition.DictLearning.html#nilearn.decomposition.DictLearning.fit .

It could be

```
Parameters
        ----------
        imgs : list of Niimg-like objects, str or Path
            See :ref:`extracting_data`.
```

However I see that, the ref below sends to the link you mentioned.

nilearn/_utils/path_finding.py
```diff
@@ -10,10 +10,12 @@ def resolve_globbing(path):
     """Resolve globbing patterns in a path."""
     path = stringify_path(path)
```
michellewang, line 11: I think @hndgzkn brought it up too but it's not intuitive to me that `stringify_path` does not necessarily return a string. Wonder if there would be a better name than this

nilearn/_utils/path_finding.py
```diff
@@ -10,10 +10,12 @@ def resolve_globbing(path):
...
     if isinstance(path, str):
-        path_list = sorted(glob.glob(str(Path(path).expanduser())))
+        expanded_path = Path(path).expanduser()
+        path_list = sorted(glob.glob(str(expanded_path)))
```
michellewang, line 14: Will Ruff complain about the use of `glob.glob` instead of `pathlib.Path.glob`/`pathlib.Path.rglob`?

nilearn/_utils/path_finding.py
Remi-Gau: I see what you mean but not sure I can come up with a better name.

michellewang: How about `stringify_if_pathlike`?

Remi-Gau: Sounds good
Peobably for a follow up pr

michellewang (approved):

hndgzkn (approved):

PR #4670, opened 2024-10-18
[STY] fix SIM117 error

Remi-Gau:
- Closes none but relates to #4579

- fix SIM117
- set upper bound for 'complexity' errors flagged by ruff.pylint so that our code does not get more complex than it is.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -160,8 +160,7 @@ ignore = [
     "PTH120",
     "PTH123",
     "PTH207",
-    "RUF012",
-    "SIM117"
+    "RUF012"
 ]
 # List of all the ruff rules (includes why the rule matters)
 # https://docs.astral.sh/ruff/rules/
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -202,3 +197,9 @@ max-complexity = 41
 
 [tool.ruff.lint.pydocstyle]
 convention = "numpy"
+
+[tool.ruff.lint.pylint]
+max-args = 26
+max-branches = 48
+max-returns = 7
+max-statements = 151
```
man-shu, line 205: Is there somewhere I can see what each of these mean?
I get max-args and max-returns, not so sure about branches and statements

man-shu (approved):

bthirion (approved):
LGTM, thx.

pyproject.toml
Remi-Gau: added link to the doc for those

PR #4673, opened 2024-10-21
[MAINT] run tests in parallel

Remi-Gau:
- Closes none

- run tests in parallel
- reduce size of the data for some decoding tests


Remi-Gau:
test that hangs on windows with python 3.12:

nilearn/decoding/tests/test_decoder.py::test_decoder_regression_screening[svr-20]

probably worth checking why this is the case

Remi-Gau:
@bthirion @man-shu 
Would not mind a review on this as it definitely speeds up the test suite: runs now take between 10 and 25 minutes (as opposed to the 25 to 50 minutes currently).

pyproject.toml
```diff
@@ -81,8 +81,9 @@ test = [
     "coverage",
     "pytest>=6.0.0",
     "pytest-cov",
-    "pytest-random-order",
+    "pytest-randomly",
```
man-shu, line 84: Just curious, what does this do differently from `pytest-random-order` that improves the speed?

I read this: https://pytest-dev.github.io/pytest-order/stable/other_plugins.html, but it doesn't seem to have an explanation for the speed improvement.

man-shu (approved):
LGTM!

pyproject.toml
Remi-Gau: ah no xdist is the pytest plugin that parallelize but pytest-random-order is not compatible with it so I had to pick another plugin to randomize the test order

bthirion (approved):
LGTM, thx.

PR #4675, opened 2024-10-21
[FIX] Silence B018 errors

Remi-Gau:
- Closes none but relates to #4579 

- this error does not make sense in examples where we sometimes need to "call" a variable to display its content


nilearn/glm/tests/test_contrasts.py
```diff
@@ -264,7 +264,7 @@ def test_deprecation_contrast_type_attribute():
         contrast = Contrast(effect, variance, contrast_type="t")
 
     with pytest.deprecated_call(match="0.13.0"):
-        contrast.contrast_type
+        contrast.contrast_type  # noqa: B018
```
hndgzkn, line 267: Are we expecting a deprecation warning here?

nilearn/glm/first_level/design_matrix.py
```diff
@@ -510,7 +510,7 @@ def make_second_level_design_matrix(subjects_label, confounds=None):
                 design_matrix.loc[ridx, conf_name] = confounds_value
 
     # check design matrix is not singular
-    sys.float_info.epsilon
+    sys.float_info.epsilon  # noqa: B018
```
hndgzkn, line 513: I couldn't understand why we need this line here.

bthirion: this line is useless, right ?

nilearn/reporting/tests/test_reporting.py
```diff
@@ -13,7 +13,7 @@
 )
 
 # Avoid making pyflakes unhappy
-_set_mpl_backend
+_set_mpl_backend  # noqa: B018
```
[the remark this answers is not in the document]
Remi-Gau: Hmm... Need to investigate this but I think this may just be there to pacify linters that complain about an unused import, but we need to import it to set up the plotting backend.

Probably best to then silence the unused import linting error

nilearn/glm/tests/test_contrasts.py
hndgzkn: It gives the warning if I specify version `0.14.0`, not with `0.13.0`. However the test already does not check if warning is raised.

Remi-Gau: OK had to adapt the stacklevel of the warning but not sure why the test did not catch this.

```python
from nilearn.glm import Contrast
import numpy as np
effect = np.ones((1, 3))
variance = effect[0]
contrast = Contrast(effect, variance, contrast_type="t")
contrast.contrast_type
```

```
/home/remi/github/nilearn/nilearn/tmp/dep.py:9: DeprecationWarning: The parameter "contrast_type" will be removed in 0.13.0 release of Nilearn. Please use the parameter "stat_type" instead.
  contrast = Contrast(effect, variance, contrast_type="t")
/home/remi/github/nilearn/nilearn/tmp/dep.py:11: DeprecationWarning: The attribute "contrast_type" will be removed in 0.13.0 release of Nilearn. Please use the attribute "stat_type" instead.
  contrast.contrast_type
```

nilearn/glm/first_level/design_matrix.py
Remi-Gau: OK this has been there and untouched since nistats was merged into nilearn: no idea why this was put there.

will remove

bthirion (approved):
LGTM, thx

nilearn/reporting/tests/test_reporting.py
```diff
@@ -2,20 +2,17 @@
 import pytest
 from nibabel import Nifti1Image
 
+from nilearn._utils.helpers import is_matplotlib_installed
 from nilearn.image import get_data
 
 # Set backend to avoid DISPLAY problems
-from nilearn.plotting import _set_mpl_backend
+from nilearn.plotting import _set_mpl_backend  # noqa
```
hndgzkn, line 9: @Remi-Gau I think there is still a problem with this line if I am not missing something. I am not sure if this line is necessary.

And a second problem is with the import below:

```
from nilearn.reporting.get_clusters_table import (
    _cluster_nearest_neighbor,
    _local_max,
)
```
These two imports causes in  `nilearn.reporting.__init__` file to run the below line

`from nilearn.reporting.glm_reporter import make_glm_report`  

which causes `matplotlib` to be imported and when it is not installed, it does not skip the tests and gives error.


I think we cannot detect it because  CI tests do not have a configuration where `matplotlib` is not installed. 

I propose to create an issue on that.

Remi-Gau: hmm...

we also run tests with none of the plotting dependencies:

github workflow
https://github.com/nilearn/nilearn/blob/dfe904a95f54ea443e7a9cba60da9d9f3b76bd13/.github/workflows/test_with_tox.yml#L37-L40

calls tox env
https://github.com/nilearn/nilearn/blob/dfe904a95f54ea443e7a9cba60da9d9f3b76bd13/tox.ini#L117-L124

with minimum dependencies and no plotting
https://github.com/nilearn/nilearn/blob/dfe904a95f54ea443e7a9cba60da9d9f3b76bd13/tox.ini#L16-L25

apparently it ran without failure:
https://github.com/nilearn/nilearn/actions/runs/11481708050/job/31953042366?pr=4675

You should be able to check things as they are ran in CI with the following in python 3.9.

```bash
pip install tox
tox run -e test_min -- nilearn/reporting/tests/test_reporting.py
```

More on tox here: https://nilearn.github.io/dev/maintenance.html#using-tox

I may defo have overlooked something, so please investigate !

hndgzkn: > with minimum dependencies and no plotting
> 
> https://github.com/nilearn/nilearn/blob/dfe904a95f54ea443e7a9cba60da9d9f3b76bd13/tox.ini#L16-L25

Ok, I hadn't check this one.

> apparently it ran without failure: https://github.com/nilearn/nilearn/actions/runs/11481708050/job/31953042366?pr=4675
> 

However, for the above run, the tests under `nilearn/reporting` are not run. Is there a condition somewhere that excludes this package when `matplotlib` is not installed?

[For this test](https://github.com/nilearn/nilearn/actions/runs/11481708050/job/31953042849?pr=4675) when matplotlib is installed, tests for `nilearn/reporting` are run.

> You should be able to check things as they are ran in CI with the following in python 3.9.
> 
> ```shell
> pip install tox
> tox run -e test_min -- nilearn/reporting/tests/test_reporting.py
> ```

When I run this, it gives error on my local.

Remi-Gau: > Is there a condition somewhere that excludes this package when `matplotlib` is not installed?

I think this part of the conftest.py is what does this.

https://github.com/nilearn/nilearn/blob/6a7a8078ef4a26e2d49203eb599cb0e395808cfe/nilearn/conftest.py#L27-L38

Remi-Gau: > When I run this, it gives error on my local.

OK
Then maybe we should not have those global collect ignore in the conftest.py otherwise it 'hides' some potential failures. What do you think?

hndgzkn: Yes, I agree.

Remi-Gau: Do you want to give this a go at fixing this?

PR #4676, opened 2024-10-22
[STY] Fix remaining `PTH118` errors in `nilearn.datasets`

michellewang:

~This is not completely done yet (missing the Neurovault fetchers) so I'm leaving this as a draft PR for now.~

- Use `pathlib.Path` instead of `os.path.join` within the `nilearn.datasets` module

Notes:
- In `nilearn.datasets.func`, the function `fetch_localizer_contrasts` has some `'/'.join([...])` statements that are not optimal. I left them as they are for now because they are not that easy to fix, but we might want to address them in a future PR.


Remi-Gau:
@michellewang double check that my merge conflict resolution makes sense: https://github.com/nilearn/nilearn/pull/4676/commits/57b52eae7342acdbc0e195de10b20f873d6aa9e7

Remi-Gau:
> For `nilearn.datasets.neurovault`, I made some breaking changes in private helper functions (and changed the tests accordingly) where I used `Path`s but didn't convert them back to strings. Then made sure all four functions in the public API still returned strings everywhere and added new assertions for that

yeah that's what I would have done too

nilearn/datasets/tests/test_atlas.py
```diff
@@ -109,7 +108,7 @@ def test_fetch_atlas_source():
 
 
 def _write_sample_atlas_metadata(ho_dir, filename, is_symm):
-    with open(os.path.join(ho_dir, f"{filename}.xml"), "w") as dm:
+    with open(Path(ho_dir, f"{filename}.xml"), "w") as dm:
```
Remi-Gau, line 111:
```suggestion
    with Path(ho_dir, f"{filename}.xml").open("w") as dm:
```

Remi-Gau (commented):
one quick thing I spotted that ruff will likely complain about at some point

bthirion (approved):
LGTM, thx !

nilearn/datasets/func.py
```diff
@@ -761,8 +760,8 @@ def fetch_localizer_contrasts(
     # Get the dataset OSF index
     dataset_name = "brainomics_localizer"
     index_url = "https://osf.io/hwbm2/download"
-    data_dir = get_dataset_dir(
-        dataset_name, data_dir=data_dir, verbose=verbose
+    data_dir = Path(
+        get_dataset_dir(dataset_name, data_dir=data_dir, verbose=verbose)
```
Remi-Gau, line 764: definitely the next thing that has to go: get_dataset_dir should return a Path

Issue #4568, opened 2024-09-26
Move experimental Surface related API into stable codebase


Remi-Gau:
- [x] move and rename pytest fixture from experimental/conftest.py - see #4638 
- [x] Update experimental surface example
  - [x] plot_surf_atlas - #4678
  - [x] plot_localizer_surface_analysis - #4681
  - [x] plot_3d_map_to_surface_projection - #4674
  - [x] plot_surface_bids_analysis - #4682
  - [x] plot_surface_image_and_maskers - #4684
  - [x] plot_surf_stat_map - #4683
- [ ] Move modules:
  - [x] IO functions - #4685
  - [x] plotting functions - #4688  
  - [x] maskers - #4692
  - [x] label masker #4714 
  - [ ] datasets - #4693 
  - [x] surface - #4723
- [ ] Check public class/method and function names
- [ ] Update doc/modules/surface.rst and remove experimental.rst
- [ ] Increase testing
- [x] What to deprecate from old surface.py API?
- [ ] Check/improve docstrings (still need review of course)

Remi-Gau:
````
$ find nilearn -name '*.py' -exec grep -H "nilearn.experimental" {} \;

nilearn/experimental/plotting/_surface_plotting.py:from nilearn.experimental.surface import PolyMesh, SurfaceImage
nilearn/experimental/plotting/tests/test_surface_plotting.py:from nilearn.experimental.plotting._surface_plotting import _check_inputs
nilearn/experimental/conftest.py:from nilearn.experimental.surface import (
nilearn/experimental/reporting/tests/test_html_report.py:from nilearn.experimental.conftest import return_mini_binary_mask
nilearn/experimental/reporting/tests/test_html_report.py:from nilearn.experimental.surface import SurfaceLabelsMasker, SurfaceMasker
nilearn/experimental/__init__.py:"""The :mod:`nilearn.experimental` module provides importable modules that \
nilearn/experimental/__init__.py:    All features in the ``nilearn.experimental`` module
nilearn/experimental/surface/maskers.py:from nilearn.experimental.surface._surface_image import PolyMesh, SurfaceImage
nilearn/experimental/surface/maskers.py:        from nilearn.experimental import plotting
nilearn/experimental/surface/maskers.py:        from nilearn.experimental import plotting
nilearn/experimental/surface/tests/test_io.py:from nilearn.experimental.surface import _io
nilearn/experimental/surface/tests/test_datasets.py:from nilearn.experimental.surface import (
[...]

PR #4684, opened 2024-10-22
[DOC] move example for surface API  plot_surface_image_and_maskers

Remi-Gau:
- Closes none but relates to #4568

- part of moving surface API to stable codebase
- move example from experimental
- add error and test in case users pass a SurfaceImage as a mesh to surface plotting functions (otherwise the error message makes debugging really hard)


Remi-Gau:
re-reading the whole example I am wondering if it makes sense to keep it as is or if we should not better try to integrate it into the other examples

it mixes plotting, decoding, scikit learn pipelines... this feels a bit like a grab bag of things related to the new surface API

if we want to keep it as a stand alone then maybe as a release hightlights like scikit learn does it?

https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_5_0.html

bthirion:
> re-reading the whole example I am wondering if it makes sense to keep it as is or if we should not better try to integrate it into the other examples
> 
> it mixes plotting, decoding, scikit learn pipelines... this feels a bit like a grab bag of things related to the new surface API
> 
> if we want to keep it as a stand alone then maybe as a release hightlights like scikit learn does it?
> 
> https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_5_0.html

I'm not sure who reads these release highlights.
But I understand it would be nicer to complement existing examples rather than add yet another one.

Remi-Gau:
Will wait until all the other experimental examples have been moved to see if this one can broken down and integrated into other examples.

Remi-Gau:
decided yesterday to move it into the 'advanced' example section and potentially refactor it after the next release

bthirion (approved):
LGTM, thx.

nilearn/plotting/tests/test_surf_plotting.py
```diff
@@ -272,7 +272,7 @@ def test_check_surface_plotting_inputs_error_mash_and_data_none(fn):
         fn(None, None)
 
 
-def test_check_surface_plotting_inputs_errors():
+def test_check_surface_plotting_inputs_errors(surf_img):
     """Fail is mesh is none and data is not not SurfaceImage."""
```
man-shu, line 276: minor typo
```suggestion
    """Fail if mesh is none and data is not not SurfaceImage."""
```

man-shu (approved):
LGTM!

Issue #4679, opened 2024-10-22
potential refactor in _asarray (old numpy)


Remi-Gau:
```suggestion
    # np.asarray does not take "K" and "A" orders in version 1.3.0
```

unrelated: seeing makes me think that we can probably simplify this function as it seems to handle behavior for old versions of numpy we do not support anymore

probably worth checking in another issue / PR

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/4677#discussion_r1810562085_

Remi-Gau:
@hndgzkn 
can I ask you to look into this one?

PR #4689, opened 2024-10-23
Maint/refactor old numpy

hndgzkn:
Closes https://github.com/nilearn/nilearn/issues/4679

### DIFF
--- a/nilearn/_utils/numpy_conversions.py
+++ b/nilearn/_utils/numpy_conversions.py
@@ -9,15 +9,7 @@ import numpy as np
 
 
 def _asarray(arr, dtype=None, order=None):
-    # np.asarray does not take "K" and "A" orders in version 1.3.0
-    if order in ("K", "A", None):
-        if (arr.itemsize == 1 and dtype in (bool, np.bool_)) or (
-            arr.dtype in (bool, np.bool_) and np.dtype(dtype).itemsize == 1
-        ):
-            ret = arr.view(dtype=dtype)
-        else:
-            ret = np.asarray(arr, dtype=dtype)
-    elif (
+    if (
         (arr.itemsize == 1 and dtype in (bool, np.bool_))
         or (arr.dtype in (bool, np.bool_) and np.dtype(dtype).itemsize == 1)
     ) and (
@@ -25,6 +17,7 @@ def _asarray(arr, dtype=None, order=None):
         and arr.flags["F_CONTIGUOUS"]
         or order == "C"
         and arr.flags["C_CONTIGUOUS"]
+        or order in ("K", "A", None)
     ):
         ret = arr.view(dtype=dtype)
     else:
--- a/nilearn/_utils/numpy_conversions.py
+++ b/nilearn/_utils/numpy_conversions.py
@@ -74,11 +74,13 @@ def as_ndarray(arr, copy=False, dtype=None, order="K"):
         isinstance(arr, np.ndarray) and copy
     ):
         ret = np.array(arr, copy=True, dtype=dtype, order=order)
-    # if the order does not change and bool to/from 1-byte dtype, no need to
-    # create a copy
+    # if the order does not change and dtype does not change or
+    # bool to/from 1-byte dtype,
+    # no need to create a copy
     elif (
         (arr.itemsize == 1 and dtype in (bool, np.bool_))
         or (arr.dtype in (bool, np.bool_) and np.dtype(dtype).itemsize == 1)
+        or arr.dtype == dtype
     ) and (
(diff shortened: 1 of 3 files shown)


nilearn/_utils/numpy_conversions.py
```diff
@@ -88,26 +81,15 @@ def as_ndarray(arr, copy=False, dtype=None, order="K"):
         Numpy array containing the same data as arr, always of class
         numpy.ndarray, and with no link to any underlying file.
     """
-    # This function should work on numpy 1.3
-    # in this version, astype() and copy() have no "order" keyword.
-    # and asarray() does not accept the "K" and "A" values for order.
-
```
hndgzkn, line 94: Probably this part can also be deleted.

Remi-Gau: yup

nilearn/_utils/numpy_conversions.py
```diff
@@ -123,10 +99,7 @@ def as_ndarray(arr, copy=False, dtype=None, order="K"):
             ret = ret.T.copy().T if ret.flags["F_CONTIGUOUS"] else ret.copy()
```
hndgzkn, line 100: I do not know why this line is necessary, couldn't we treat `np.memmap` and `np.ndarray` the same way ?

bthirion: A priori I don't think we need to make a difference. But I have limited experience with mmap manipulation.

hndgzkn: I'll check if it makes difference in terms of efficiency.

Remi-Gau: thanks!

nilearn/_utils/numpy_conversions.py
```diff
@@ -9,22 +9,15 @@
...
-    elif (
+    if (
         (arr.itemsize == 1 and dtype in (bool, np.bool_))
         or (arr.dtype in (bool, np.bool_) and np.dtype(dtype).itemsize == 1)
     ) and (
         order == "F"
         and arr.flags["F_CONTIGUOUS"]
         or order == "C"
         and arr.flags["C_CONTIGUOUS"]
+        or order in ("K", "A", None)
```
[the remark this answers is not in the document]
Remi-Gau: nope
not from the top of my head
would need to look into numpy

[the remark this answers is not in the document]
hndgzkn: I guess, it is for performance concerns. As far as I understand, this is a case where numpy does not need to copy the specified `arr` itself, so if a copy is not required, it sends a view.

nilearn/_utils/numpy_conversions.py
hndgzkn: This is necessary for performance issues.

`ret.T.copy().T` takes half the time compared to `ret.copy()` when ret array is fortran style. 

However `np.array(ret, copy=True)` takes approximately the same time as `ret.T.copy().T` may be less. 

I guess this was necessary for old numpy versions. I tested for 1.22 and last numpy version and there is no need for this line any more.

nilearn/_utils/numpy_conversions.py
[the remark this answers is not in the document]
hndgzkn: As far as I understand, this is also for performance issues. 

`np.asarray` does not create a copy of array unless it is necessary. When `dtype` or `order` changes it creates a copy. 

I think this is to avoid unnecessary copying of array when the the size of the `dtype` speficied is compatible with the `dype` of array and the `order` does not change. This is done only when either `dtype` specified or `dtype` of the array is `bool`. When the condition is satisfied, it returns a view instead of copying the underlying array.

nilearn/image/tests/test_resampling.py
```diff
@@ -1265,3 +1267,20 @@ def test_smoke_resampling_non_nifti(affine_eye, shape, rng, force_resample):
         force_resample=force_resample,
         copy_header=True,
     )
+
+
+@pytest.mark.parametrize("shape", [(1, 4, 4)])
+def test_get_mask_bounds(data, affine_eye):
+    img = Nifti1Image(data, affine_eye)
+    assert_allclose((0.0, 0.0, 0.0, 3.0, 0.0, 3.0), get_mask_bounds(img))
```
hndgzkn, line 1275: @bthirion I've added tests. However I've used the results returned from the function.

bthirion (approved):
LGTM, thx.

Remi-Gau (approved):

PR #4517, opened 2024-09-03
[MAINT] enact deprecations due for 0.11.0

Remi-Gau:
- Closes none but relates to #3812

Changes proposed in this pull request:

- remove deprecated functions
- change default value of deprecated parameters

PR #4694, opened 2024-10-24
[FIX] do not use scaling_axis in example

Remi-Gau:
- Closes none but fixes doc build fail on main

- scaling_axis was removed in https://github.com/nilearn/nilearn/pull/4517 as it was deprecated but not from the example

### DIFF
--- a/examples/07_advanced/plot_beta_series.py
+++ b/examples/07_advanced/plot_beta_series.py
@@ -74,8 +74,9 @@ from nilearn.glm.first_level import FirstLevelModel, first_level_from_bids
 data = fetch_language_localizer_demo_dataset(legacy_output=False)
 
 models, models_run_imgs, events_dfs, models_confounds = first_level_from_bids(
-    data.data_dir,
-    "languagelocalizer",
+    dataset_path=data.data_dir,
+    task_label="languagelocalizer",
+    sub_labels=["01"],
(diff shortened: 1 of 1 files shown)


examples/07_advanced/plot_beta_series.py
```diff
@@ -102,6 +103,8 @@
 # :func:`~nilearn.glm.first_level.first_level_from_bids`.
 import matplotlib.pyplot as plt
 
+print("Fit model")
```
Remi-Gau, line 106: adding some print statement because the example takes a long time to run and this helps know what is happening

examples/07_advanced/plot_beta_series.py
```diff
@@ -74,8 +74,9 @@
 data = fetch_language_localizer_demo_dataset(legacy_output=False)
 
 models, models_run_imgs, events_dfs, models_confounds = first_level_from_bids(
-    data.data_dir,
-    "languagelocalizer",
+    dataset_path=data.data_dir,
+    task_label="languagelocalizer",
+    sub_labels=["01"],
```
Remi-Gau, line 79: mention parameters name and only run on the first subject since this is the only one we need

Remi-Gau:
will merge to make sure our build of the doc on main do not fail

PR #4696, opened 2024-10-24
[STY] Fix `PTH123` in `nilearn.datasets`

michellewang:

- Stop ignoring `nilearn.datasets` files for error `PTH123` in ruff
- Change standard library `open` calls to `pathlib.Path.open`


nilearn/datasets/_utils.py
```diff
@@ -610,7 +610,7 @@ def fetch_single_file(
                     ):
                         raise OSError("Server does not support resuming")
                     initial_size = local_file_size
-                    with open(local_file, "ab") as fh:
+                    with local_file.open("ab") as fh:
```
michellewang, line 613: I think this is a bug. These lines do not seem to be covered by the current tests, and `local_file` is set to `None` at Line ~574 and never changed after, so as far as I can tell the file opening should fail regardless of which function/method is used. I guess it is not that easy to test because it requires a partially downloaded file.

I am not familiar with what the function is supposed to do but if someone confirms that this is a bug then I can create an issue for it.

bthirion (approved):
LGTM, thx.

nilearn/datasets/_utils.py
Remi-Gau: yeah let's open a separate issue: did a bit git-archeology (gitcheology?)

it seems that the relevant changes happened a long time ago

https://github.com/nilearn/nilearn/blame/893b0fcc9b819254efc1ecad659a075b33fd2867/nilearn/datasets/utils.py#L576

nilearn/datasets/tests/test_neurovault.py
```diff
@@ -284,7 +284,7 @@ def test_get_batch(tmp_path):
 
     assert "results" in batch
     assert "count" in batch
-    with open(tmp_path / "test_nv.txt", "w"):
+    with (tmp_path / "test_nv.txt").open("w"):
         pass
```
Remi-Gau, line 288: what does this do?
Is this even needed: does the test fail if you remove it?

if still needed I think we can just have

```suggestion
    (tmp_path / "test_nv.txt").touch()
```

Remi-Gau: Let's check that in another issue / PR

Remi-Gau (approved):

nilearn/datasets/tests/test_neurovault.py
michellewang: @Remi-Gau good catch, indeed it does nothing. Will open an issue/PR to remove it

PR #4701, opened 2024-10-25
[FIX] Fix bug in `fetch_single_file` with `resume=True` and improve tests

### LINKED ISSUE #4697: local_file never set to an actual value
I think this is a bug. These lines do not seem to be covered by the current tests, and `local_file` is set to `None` at Line ~574 and never changed after, so as far as I can tell the file opening should fail regardless of which function/method is used. I guess it is not that easy to test because it requires a partially downloaded file.

I am not familiar with what the function is supposed to do but if someone confirms that this is a bug then I can create an issue for it.

_Originally posted by @michellewang in https://github.com/nilearn/nilearn/pull/4696#discussion_r1815581252_

michellewang:
- Closes #4697

- Use `request_mocker` so that the test actually works
- Deleted `local_file` (always `None`) and use actual filename instead
- Make exception catching more specific
- Some improvements in other tests, especially regarding to the use of context managers and `pathlib.Path`'s `read_text` and `write_text` methods

### DIFF
--- a/nilearn/datasets/_utils.py
+++ b/nilearn/datasets/_utils.py
@@ -571,7 +571,6 @@ def fetch_single_file(
     if temp_full_name.exists() and overwrite:
         temp_full_name.unlink()
     t0 = time.time()
-    local_file = None
     initial_size = 0
 
     try:
@@ -610,7 +609,7 @@ def fetch_single_file(
                     ):
                         raise OSError("Server does not support resuming")
                     initial_size = local_file_size
-                    with local_file.open("ab") as fh:
+                    with temp_full_name.open("ab") as fh:
                         _chunk_read_(
                             resp,
                             fh,
@@ -618,7 +617,7 @@ def fetch_single_file(
                             initial_size=initial_size,
                             verbose=verbose,
                         )
-            except Exception:
+            except OSError:
                 logger.log(
                     "Resuming failed, try to download the whole file.", verbose
                 )
@@ -665,7 +664,7 @@ def fetch_single_file(
         raise
     if md5sum is not None and _md5_sum_file(full_name) != md5sum:
         raise ValueError(
-            f"File {local_file} checksum verification has failed."
+            f"File {full_name} checksum verification has failed."
             " Dataset fetching aborted."
         )
     return full_name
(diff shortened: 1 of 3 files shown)


nilearn/datasets/tests/test_utils.py
```diff
@@ -364,7 +364,7 @@ def test_uncompress_tar(tmp_path, ext, mode):
     # and check if ftemp exists
     ztemp = tmp_path / f"test.{ext}"
     ftemp = "test"
-    with contextlib.closing(tarfile.open(ztemp, mode)) as testtar:
+    with tarfile.open(ztemp, mode) as testtar:
```
michellewang, line 367: Pretty sure the `contextlib.closing` is not needed : https://docs.python.org/3/library/tarfile.html#tarfile-objects

nilearn/datasets/tests/test_utils.py
```diff
@@ -382,7 +382,7 @@ def test_uncompress_zip(tmp_path):
     # and check if ftemp exists
     ztemp = tmp_path / "test.zip"
     ftemp = "test"
-    with contextlib.closing(ZipFile(ztemp, "w")) as testzip:
+    with ZipFile(ztemp, "w") as testzip:
```
michellewang, line 385: `Zipfile` is already a context manager: https://docs.python.org/3/library/zipfile.html#zipfile-objects

nilearn/datasets/tests/test_utils.py
```diff
@@ -422,35 +422,57 @@ def test_safe_extract(tmp_path):
         _utils.uncompress_file(ztemp, verbose=0)
 
 
-def test_fetch_file_part(tmp_path):
+def test_fetch_single_file_part(tmp_path, capsys, request_mocker):
```
michellewang, line 425: Looks like `fetch_single_file` used to be called `fetch_file`, in which case the tests should be renamed too

nilearn/datasets/_utils.py
```diff
@@ -610,15 +609,15 @@ def fetch_single_file(
...
+                    with temp_full_name.open("ab") as fh:
                         _chunk_read_(
                             resp,
                             fh,
                             report_hook=(verbose > 0),
                             initial_size=initial_size,
                             verbose=verbose,
                         )
-            except Exception:
+            except OSError:
```
bthirion, line 620: Can you test this line ?

bthirion (commented):
Thx for the contribution.

nilearn/datasets/_utils.py
michellewang: Added a test for this

bthirion (approved):
LGTM, thx.

Remi-Gau (approved):

PR #4704, opened 2024-10-31
[STY] Apply assorted ruff rules

DimitriPapadopoulos:
- closes none


bthirion (approved):
LGTM ! Thx

bthirion:
Oops, it looks like the test suite is broken.

examples/01_plotting/plot_surf_atlas.py
```diff
@@ -134,7 +134,7 @@
 corr = np.zeros((n_parcels, n_parcels))
 n_parcels_hemi = n_parcels // 2
 corr[np.arange(n_parcels_hemi), np.arange(n_parcels_hemi) + n_parcels_hemi] = 1
-corr = corr + corr.T
+corr += corr.T
```
Remi-Gau, line 137: I have seen python beginners getting confused by some of the typical python shortcut, so I wonder if we should not avoid such shortcuts in the examples?

I also totally acknowledge that I may be responsible for adding quite a few python 'idioms' in the example recently so I am not 100% consistent here.

nilearn/_utils/data_gen.py
```diff
@@ -1112,7 +1112,7 @@ def _mock_bids_dataset(
     ):
         subses_dir = bids_path / f"sub-{subject}"
         if session != "":
-            subses_dir = subses_dir / f"ses-{session}"
+            subses_dir /= f"ses-{session}"
```
Remi-Gau, line 1115: I did not know that one.
😜

examples/01_plotting/plot_surf_atlas.py
bthirion: I take your point.

Remi-Gau (changes requested):
style checks are failing.

run 

```
pip install tox
tox
```

and commit the changes to fix

nilearn/datasets/atlas.py
```diff
@@ -108,8 +108,8 @@ def fetch_atlas_difumo(
         )
     if resolution_mm not in valid_resolution_mm:
         raise ValueError(
-            "Requested resolution_mm={resolution_mm} is not available. "
-            "Valid options: {valid_resolution_mm}"
+            f"Requested resolution_mm={resolution_mm} is not available. "
```
Remi-Gau, line 111: I find it 'worrying' that this 'invalid' f strings (meaning that this should have been an f string and it was just a sting) was not caught by our tests.

is this a ruff autofix thing?

Remi-Gau:
@DimitriPapadopoulos 

how come those errors were not caught / fixed by our ruff set up?

you did not change the ruff config in the pypoject.toml or the ruff version we use in the pre-commit config. 

how can we make sure that those fixes will not be reverted?

nilearn/_utils/niimg_conversions.py
```diff
@@ -178,17 +178,12 @@ def iter_check_niimg(
...
-                        "Reference affine:\n%r\nImage affine:\n%r\n"
-                        "Reference shape:\n%r\nImage shape:\n%r\n"
-                        % (
-                            i,
-                            ref_fov[0],
-                            niimg.affine,
-                            ref_fov[1],
-                            niimg.shape,
-                        )
+                        f"Reference affine:\n{ref_fov[0]:r}\n"
```
Remi-Gau, line 183: seems this line is causing several tests failures

I remember that when I was converting old string formats (% and .format) to f strings, some would lead to some tests failures if I did. I think this may have been one of them.
Sorry for not flagging it in a comment.

If you can find how to fix and keep the f strings, that'd be great, otherwise we should ignore this fix by ruff here.

examples/01_plotting/plot_surf_atlas.py
DimitriPapadopoulos: On the other hand '+=' has its sources in C - not really an Python idiom.

nilearn/datasets/atlas.py
DimitriPapadopoulos: These might be **preview** rules, not yet available in the default set of rules. Let me have a look.

examples/01_plotting/plot_surf_atlas.py
Remi-Gau: yeah yeah I know: but to someone new to programming, they are likely to think of it as "a python thing" (especially as such syntax does not exist in R or matlab, the 2 other languages our users are likely to know a bit about).

as I said: I am not 100% consistent on this, so I'd be happy to leave to those changes in for this PR and open an issue to try to make sure we avoid 'complex syntax' (to be defined) in the examples (and maybe revert those changes later in another PR).

I think this is probably we should discuss witht the core dev to decide.

Remi-Gau:
was checking the release notes of ruff and AFAICT it does not seem to explain why the fixes in this PR were not caught before: https://github.com/astral-sh/ruff/releases/tag/0.7.2

@DimitriPapadopoulos what ruff version / config / command did you use to catch / fix those errors?

nilearn/_utils/niimg_conversions.py
DimitriPapadopoulos: `:r` → `!r`

DimitriPapadopoulos:
While [UP031](https://docs.astral.sh/ruff/rules/printf-string-formatting/) is **not** a preview rule, more issues are reported with `--preview`. In the current main branch:
```console
$ ruff check --select UP031
warning: Selection `DOC` has no effect because preview is not enabled.
All checks passed!
$ 
$ ruff check --preview --select UP031
examples/03_connectivity/plot_extract_regions_dictlearning_maps.py:102:5: UP031 Use format specifiers instead of percent format
    |
100 |   # Visualization of region extraction results
101 |   title = (
102 |       "%d regions are extracted from %d components."
    |  _____^
103 | |     "\nEach separate color of region indicates extracted region"
    | |________________________________________________________________^ UP031
104 |       % (n_regions_extracted, 8)
105 |   )
    |
    = help: Replace with format specifiers
[...]

Found 7 errors.
$

File: examples/01_plotting/plot_surf_atlas.py (hunk 1 above)
Comment (DimitriPapadopoulos) [reply examples/01_plotting/plot_surf_atlas.py:137]: Not a problem to remove this specific commit, if that's what you're asking.

Comment (Remi-Gau) [reply examples/01_plotting/plot_surf_atlas.py:137]: na that's ok, leave them: we will talk with the core dev about this and probably try to simplify examples later

**Comment by Remi-Gau** [comment]
@DimitriPapadopoulos 
We talked about this PR yesterday at the dev meeting.

We are not too keen on adding the "preview" changes from ruff given that they are considered unstable and may never make it into the actual set of rules supported by ruff, so it may be better to not have those in this PR.

I realize this may mean closing this PR.

Some fixes that this PR brought up should still be done for example: see for exmaple https://github.com/nilearn/nilearn/pull/4704#discussion_r1827405848

**Comment by DimitriPapadopoulos** [comment]
Is the idea to keep only the `RUF027` commit?

**Comment by DimitriPapadopoulos** [comment]
Note that `DOC` rules need to be disabled, to avoid this warning:
```
warning: Selection `DOC` has no effect because preview is not enabled.
```

**Comment by Remi-Gau** [comment]
> Is the idea to keep only the `RUF027` commit?
> 
> Would it be it OK to keep the commits that make real sense beyond pleasing the linter?
> 
>     * `E203`
> 
>     * `UP031`
> 
>     * `C419`

Yeah it makes sense to keep those too.

File: nilearn/datasets/_utils.py [hunk 5]
```diff
@@ -159,13 +154,12 @@ def _chunk_read_(
             "Warning: total size could not be determined.",
             verbose=verbose,
             msg_level=2,
-            stack_level=2,
         )
         logger.log(
-            f"Full stack trace: {e}",
+            "Full stack trace: %s",
```
bthirion, line 159: I'm a bit puzzled by this and following changes ?

nilearn/datasets/_utils.py
```diff
@@ -83,9 +83,7 @@ def _chunk_report_(bytes_so_far, total_size, initial_size, t0):
 
     """
     if not total_size:
-        logger.log(
-            f"\rDownloaded {int(bytes_so_far)} of ? bytes.", stack_level=2
-        )
+        logger.log(f"\rDownloaded %d of ? bytes.", bytes_so_far)
```
Remi-Gau, line 86:
```suggestion
        logger.log(f"\rDownloaded {bytes_so_far} of ? bytes.", stack_level=2)
```

Remi-Gau: Shouldn't it be more something like this?

nilearn/datasets/_utils.py
DimitriPapadopoulos: That's all on purpose:
* String interpolation is "usually" deferred to the log function, so that interpolation occurs only when the message will be printed according the log level (lazy interpolation). See [G004](https://docs.astral.sh/ruff/rules/logging-f-string/). The efficiency of this rule is debatable though (and has been debated).
* The `stack_level` argument is already specified by `logger.log` in this case. Repeating it here, with a different value, results in warnings.

DimitriPapadopoulos: With that said, I can defer these changes to a different PR, especially the f-string removal.

Remi-Gau: no problem: your assumption was a sound one !

Remi-Gau:
Will take care of the merge conflict.

Remi-Gau (approved):

Remi-Gau:
thanks @DimitriPapadopoulos for this one

PR #4692, opened 2024-10-23
[ENH] Move expriemental surface maskers to stable part of the code base 

Remi-Gau:
- Closes #

Changes proposed in this pull request:

-
-

PR #4714, opened 2024-11-08
[ENH] move SurfaceLabelMasker to stable API

Remi-Gau:
- Closes none but relates #4568

- move SurfaceLabelMasker from experimental to nilearn/maskers

- [x] add "cleaning" to the transform method (including the relevant parameters in the constructor)


Remi-Gau:
PR stacked on #4692 

blocked by #4692

nilearn/maskers/surface_labels_masker.py
```diff
@@ -0,0 +1,463 @@
...
+        all_labels = set(self._labels_data.ravel())
+        all_labels.discard(self.background_label)
+        self._labels_ = list(all_labels)
+
+        self.n_elements_ = len(self._labels_)
+
+        if self.labels is None:
+            self._label_names_ = [str(label) for label in self._labels_]
+        else:
+            self._label_names_ = [self.labels[x] for x in self._labels_]
```
Remi-Gau, line 176: @mtorabi59 
I am keeping most of the fitted attributes private so we can later improve how to better align how the different label maskers (nifti and surface) work with missing labels and all the other issues we uncovered with label maskers.

Remi-Gau:
failures on no plotting for searchlight estimator feels unrelated...

nilearn/maskers/surface_labels_masker.py
mtorabi59: @Remi-Gau sounds good!

mtorabi59 (approved):
@Remi-Gau Looks good to me! Thanks!

bthirion (commented):
Looks mostly good to me, but there seem to be a couple of codecov complaints in the reporting code.

man-shu (approved):
LGTM!

nilearn/maskers/surface_labels_masker.py
```diff
@@ -0,0 +1,463 @@
...
+            The labels must be consistent with the label values
+            provided through ``labels_img``.
+
+    background_label : :obj:`int` or :obj:`float`, default=0
+        Label used in labels_img to represent background.
+
+        .. warning::
+
+            This value must be consistent with label values
+            and image provided.
```
Remi-Gau, line 45: Not 100% sure it makes sense to have a background label on a **surface** label masker, but in case people project a volume atlas onto a surface, it may happen to have some vertices that are 'unassinged to a region' so probably better to keep this.

May also happen if the atlas definition does not cover all the brain regions (for example only map visual areas).

Remi-Gau:
paranoid me will add one more test to more precisely check that extracted signals are what we should expect.

nilearn/maskers/tests/test_surface_labels_masker.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+def test_surface_label_masker_transform_check_output(surf_mesh, surf_img, rng):
+    """Check actual content of the transform.
+
+    - Use a label mask with more than one label.
+    - Use data with known content and expected mean.
+      and background label data has random value.
+    - Check that output data is properly averaged,
+      even when labels are spread across hemispheres.
```
Remi-Gau, line 119: @man-shu 
can you read this test and tell me if this makes sense to you?
I wanted to have something that does more than just checking the shapes of the output.

man-shu: Yes indeed! Very clear!

nilearn/maskers/surface_labels_masker.py
man-shu, line 1: I am wondering why we don't have a `mask` or `mask_img` in SurfaceLabelsMasker here, as we do in NiftiLabelsMasker.

Remi-Gau: oops 

merged before seeing this one

see [this](https://github.com/nilearn/nilearn/pull/4714#discussion_r1843650150)

some of the features or attributes that we have in the nifti version have either not been implemented or are kept as private attributes: to keep the scope small and be able to rework the internals without breaking the API.

man-shu: Yes, sorry, I started wondering after approving the PR 😅

> some of the features or attributes that we have in the nifti version have either not been implemented or are kept as private attributes: to keep the scope small and be able to rework the internals without breaking the API.

Doing more frequent releases is starting to make more sense to me now

nilearn/maskers/tests/test_surface_labels_masker.py
man-shu: I just realized that in this test, we weren't actually using the data with 2 timepoints and transforming with the masker.

- Will update in #4770

PR #4725, opened 2024-11-14
[MAINT] use pathlib.Path more systematically in nilearn.datasets

Remi-Gau:

- use pathlib.Path more systematically in nilearn.datasets
- still need to do fetch_files: in another PR

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -789,17 +789,15 @@ def fetch_localizer_contrasts(
                 file_path = Path(
                     "brainomics_data", subject_id, f"{name_aux}.nii.gz"
                 )
-                path = "/".join(
-                    [
-                        "/localizer",
-                        "derivatives",
-                        "spm_1st_level",
-                        f"sub-{subject_id}",
-                        (
-                            f"sub-{subject_id}_task-localizer"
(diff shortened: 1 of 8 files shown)


bthirion (approved):
Great, thx !

nilearn/datasets/func.py
```diff
@@ -785,9 +786,8 @@ def fetch_localizer_contrasts(
                 name_aux = str.replace(
                     str.join("_", [data_type, contrast]), " ", "_"
                 )
-                file_path = Path(
-                    "brainomics_data", subject_id, f"{name_aux}.nii.gz"
-                )
+                file_path = f"brainomics_data/{subject_id}/{name_aux}.nii.gz"
```
michellewang, line 789: I'm sure you had a good reason but, for my own understanding, for revert to hardcoded `"/"`?

michellewang (approved):
Minor comment (question) which is really for my own understanding, otherwise looks good!

nilearn/datasets/func.py
Remi-Gau: hmm... I thing I was getting some tests failures and tried to fix it but forgot to revert those.
let me fix this

PR #4727, opened 2024-11-15
[DOC] update internal references to SurfaceImage classes in doc strings

Remi-Gau:
- Closes none

- make sure that we point to the surface image doc internally


man-shu (approved):
LGTM! I should do this too on the clustering PR

Remi-Gau:
> LGTM! I should do this too on the clustering PR

I think we should do a final cross check before the release anyway

PR #4730, opened 2024-11-16
[STY] Apply ruff/flake8-implicit-str-concat rules (ISC)

DimitriPapadopoulos:


Remi-Gau (changes requested):
I like this especially that I have had to fix things that manually before.

Can you update the pyproject.toml to add the rule to the ruff config?

DimitriPapadopoulos:
In theory, `ISC1001` and `ISC002` rules are part of the [conflicting lint rules](https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules) to avoid when using Ruff as a formatter. More details in https://github.com/astral-sh/ruff/issues/8272. In practice, I've never seen conflicts.

That's why I haven't enforced `ISC` rules.

DimitriPapadopoulos:
See #4753.

Remi-Gau (approved):

Remi-Gau:
> In theory, `ISC1001` and `ISC002` rules are part of the [conflicting lint rules](https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules) to avoid when using Ruff as a formatter. More details in [astral-sh/ruff#8272](https://github.com/astral-sh/ruff/issues/8272). In practice, I've never seen conflicts.
> 
> That's why I haven't enforced `ISC` rules.

ah I see
I did not know that: thanks for the explanation!!! 🙏🏽 

will merge as is as the changes are small and uncontroversial

PR #4736, opened 2024-11-17
[STY] Use `callable` to check if an object is callable

DimitriPapadopoulos:
It is unreliable to use `hasattr(object, '__call__')` to check if an object is callable as it can give wrong results if `object` implements custom `__getattr__` or its `__call__` is itself not callable. Use the `callable` built-in instead.

### DIFF
--- a/nilearn/externals/tempita/_looper.py
+++ b/nilearn/externals/tempita/_looper.py
@@ -155,7 +155,7 @@ class loop_pos:
                 return getattr(item, getter)() != getattr(other, getter)()
             else:
                 return getattr(item, getter) != getattr(other, getter)
-        elif hasattr(getter, '__call__'):
+        elif callable(getter):
             return getter(item) != getter(other)
         else:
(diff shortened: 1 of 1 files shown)


bthirion (approved):

Remi-Gau (commented):
This is in our externals subpackage that contains code that is not ours but distribute because we need it.

I'd be tempted to say to leave that one alone.

I just realized that apparently this now has a new release: 
- https://github.com/TurboGears/tempita/releases/tag/0.6

So it may be worth considering, updating this part of the codebase (or adding it as a dependency).

DimitriPapadopoulos:
My wrong. I didn't notice it is a `tempita` file.

I need to find a way to instruct DeepSource to ignore `externals`. This doesn't work as far as I can tell:
https://github.com/DimitriPapadopoulos/nilearn/blob/4e464f7e0a307f862989b73560f58ee185b4209d/.deepsource.toml#L3

PR #4737, opened 2024-11-17
[STY] Comparison with callable

DimitriPapadopoulos:
A comparison with a callable may suggest you have forgotten to actually call the function or method. If you intended to check if both of these callables are the same, it's recommended to use the `is` operator for comparison.


bthirion (approved):

Remi-Gau (approved):

Remi-Gau:
yup makes sense !

PR #4740, opened 2024-11-17
[STY] Appending to dictionary following its definition

DimitriPapadopoulos:


bthirion (approved):
LGTM, thx

nilearn/plotting/cm.py
```diff
@@ -54,40 +54,36 @@ def mix_colormaps(fg, bg):
...
-        )
+    cdict = cmap._segmentdata.copy()
 
-    return cdict
+    return {
+        "green": [
+            (0.5 * (1 - p), c1, c2)
+            for (p, c1, c2) in reversed(cdict[swap_order[0]])
+        ]
+        + [(0.5 * (1 + p), c1, c2) for (p, c1, c2) in cdict["green"]],
```
Remi-Gau, line 75: do you think you could extract this into a small function with explicit name and variable names that are not just one letter because at the moment it is pretty hard to understand what it going on?

DimitriPapadopoulos: I have created a function but am not sure about meaningful names. Indeed I haven't been able to understand enough of the code yet and derive names that would be standard or meaningful 😀

Remi-Gau: haha: that makes me feel so less stupid !!
thanks

Remi-Gau (approved):
Thanks !!!

PR #4743, opened 2024-11-17
[MAINT] Sequence value overwritten unconditionally

DimitriPapadopoulos:
- From #2723.

### DIFF
--- a/nilearn/datasets/atlas.py
+++ b/nilearn/datasets/atlas.py
@@ -804,10 +804,10 @@ def _compute_symmetric_split(source, atlas_niimg, names):
     if source == "Juelich":
         for idx, name in enumerate(names):
             if name.endswith("L"):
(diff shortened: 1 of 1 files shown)


bthirion (commented):
There is no smoke test for these lines ?

Remi-Gau:
> There is no smoke test for these lines ?

I think those lines should be covered: https://github.com/nilearn/nilearn/blob/d73672d41992659395c401fbc08e876708aa78ce/nilearn/datasets/tests/test_atlas.py#L204

Remi-Gau (approved):

Remi-Gau:
I think this is because the data we mock do not have names that match some of the 'idiosyncrasies' of the real data.

DimitriPapadopoulos:
I'm not very familiar with the Jülich atlas names, i'll try to find one that exposes this issue.

bthirion (approved):

PR #4749, opened 2024-11-17
[MAINT] Special method should return `NotImplemented`

DimitriPapadopoulos:
From https://docs.python.org/3/library/constants.html#NotImplemented

> When a binary (or in-place) method returns `NotImplemented` the interpreter will try the reflected operation on the other type (or some other fallback, depending on the operator). If all attempts return `NotImplemented`, the interpreter will raise an appropriate exception.


bthirion (approved):
A test would be great !

Remi-Gau (approved):

Issue #4739, opened 2024-11-17
[STY] Remove `line-length = 79`


DimitriPapadopoulos:
How about discarding this line? 
https://github.com/nilearn/nilearn/blob/b5a19af89677545388a1b365c9d9b0224e99e9e6/pyproject.toml#L124

Stop following that PEP8 recommendation and instead use the black/ruff default value of [`line-length`](https://docs.astral.sh/ruff/settings/) which is 88.

Also, how about disabling  `E501` as recommended in the [`line-length`](https://docs.astral.sh/ruff/settings/#line-length) documentation? See also [Conflicting lint rules](https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules).

bthirion:
No strong opinion on that one. If there is a consensus for extending line length to 88, I'm good with it. We should just avoid getting back and forth on this !

Remi-Gau:
I suggest we circle back to this one after the 0.11 release

Remi-Gau:
I would be tempted to keep the things as they are:
- current line length indirectly forces us to avoid deeper block nesting (an if within an if within a loop...) which helps readability
- silencing E501 would allow to have super long doc strings on a single line, no? I don't see this as a good thing
- more narrow code helps reading it on a narrower screen: that is especially true if one wants to read the examples on a phone. We already have some weird annoying line breaks as it is: wider lines will make it worse.

![Image](https://github.com/user-attachments/assets/d72d355c-4a5f-4287-b35e-4137d17402cf)

DimitriPapadopoulos:
* There are pros and cons. Shorted lines may also force meaningless variable names. Those who suggested changing the line length from 79 to 88 have obviously already thought about these issues and discussed them. I'd rather refer to existing discussions such as:
  * [The _Black_ code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length)
    > You probably noticed the peculiar default line length. _Black_ defaults to 88 characters per line, which happens to be 10% over 80. This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, [90-ish seems like the wise choice](https://youtu.be/wf-BqAjZb8M?t=260).
* I do not suggest disabling `E501`.
[...]

PR #4753, opened 2024-11-17
[MAINT] Disable ruff rules conflicting with the formatter

DimitriPapadopoulos:
Also disable deprecated and problematic rules.

Addresses part of #4739.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -151,7 +151,25 @@ ignore = [
     "PIE790",
     "PLR2004",
     "PTH207",
-    "RUF012"
+    "RUF012",
+    "UP027",  # deprecated
+    "UP038",  # https://github.com/astral-sh/ruff/issues/7871
+    # Avoid linter rules conflicting with the formatter
+    # https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
+    "COM812",
+    "COM819",
+    "D206",
+    "D300",
+    "E111",
+    "E114",
+    "E117",
+    "ISC001",
+    "ISC002",
+    "Q000",
+    "Q001",
+    "Q002",
+    "Q003",
+    "W191"
 ]
 # List of all the ruff rules (includes why the rule matters)
 # https://docs.astral.sh/ruff/rules/


bthirion (commented):
No opinion on this.

pyproject.toml
```diff
@@ -150,7 +150,24 @@ ignore = [
...
+    "D206",
+    "D300",
+    "Q000",
+    "Q001",
+    "Q002",
+    "Q003",
+    "COM812",
+    "COM819",
+    "ISC001",
+    "ISC002"
```
Remi-Gau, line 170: OK with me but can I ask you to sort those alpha-numerically for the sake of our common long term mental heatlh. 🙏🏽

Remi-Gau (changes requested):
one nitpick

pyproject.toml
DimitriPapadopoulos: I kept the order of the documentation. That's something I can change of course. Is it OK to keep the conflicting rules as a separate section?

```
    .
    .
    .
    "UP027",  # deprecated
    "UP038",  # https://github.com/astral-sh/ruff/issues/7871
    # https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
    "COM812",
    "COM819",
    .
    .
    .
    "W191"
```

Remi-Gau: YES I was thinking to actually keep that comment as a separator so we know which ones come from us or not: don't hesitate to make that comment even more clear about this !!!

DimitriPapadopoulos: Done.

DimitriPapadopoulos:
While we're at it, I would suggest changing [`select`](https://docs.astral.sh/ruff/settings/#lint_select) to [`extend-select`](https://docs.astral.sh/ruff/settings/#lint_extend-select):
https://github.com/nilearn/nilearn/blob/d0231a4446a2cad4811987fc8750b93972551041/pyproject.toml#L156-L158

This ensures **default rules** are selected as ruff matures.

Remi-Gau (approved):
thanks !!

Issue #4668, opened 2024-10-18
Excessive number of warning when running searchlight


Remi-Gau:
Just ran `examples/02_decoding/plot_haxby_searchlight.py` and getting MANY non convergence warning:

```
  warnings.warn(
/home/remi/miniconda3/lib/python3.12/site-packages/sklearn/svm/_base.py:1235: ConvergenceWarning: Liblinear failed to converge, increase the number of iterations.
```

The warnings do not show in the doc:
https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_searchlight.html#searchlight-computation

But they show in the doc build: https://github.com/nilearn/nilearn/actions/runs/11370313248/job/31634736528#step:14:1083

This is surprising given those warnings are supposed to be ignored: 

https://github.com/nilearn/nilearn/blob/be11c5de57d1d50db426c5c07373402a70b81679/nilearn/decoding/searchlight.py#L87

Am I missing something?

bthirion:
No emergency IMHO, but indeed !

Issue #3541, opened 2023-02-27
Suggestion to use warnings-as-errors with pytest in prerelease CI


ymzayek:
> There is at least one test that would fail under the original bug conditions.

To find stuff like this in MNE we test with warnings-as-errors in pytest, and use a prerelease version of dependencies like `matplotlib`. It helps stay ahead of the curve a bit. But it would be a big change here...

I could instead change your pytest config to at least error on matplotlib deprecations, then you'd at least see them when they cut new releases. Not sure how many other failures this would trigger, it took us a bit to get to 3.7-deprecation-free compat over in MNE...

_Originally posted by @larsoner in https://github.com/nilearn/nilearn/issues/3535#issuecomment-1442270288_
            

Making sure this doesn't get lost. This would be added to our prerelease job in the testing workflow https://github.com/nilearn/nilearn/blob/c7aa9f74ec0b13f4946921d0b4b921e81a8639f1/.github/workflows/testing.yml#L106-L114

ymzayek:
Reopening (see #4149)

PR #4754, opened 2024-11-17
[MAINT] Apply repo-review rules

DimitriPapadopoulos:
https://learn.scientific-python.org/development/guides/repo-review/?repo=nilearn%2Fnilearn&branch=main


pyproject.toml
```diff
@@ -109,6 +109,7 @@ source = "vcs"
 [tool.pytest.ini_options]
 addopts = "-ra --strict-config --strict-markers --doctest-modules --showlocals -s -vv --durations=0 --template=maint_tools/templates/index.html -n auto"
 doctest_optionflags = "NORMALIZE_WHITESPACE ELLIPSIS"
+filterwarnings = ["error"]
```
Remi-Gau, line 112: I started to make sure that Nilearn follows the scientific python recommendations last year (sorry too lazy to find the PR).

Turning exact warnings into errors triggers >500 to fail so I think we decided to ignore this recommendation.

Vaguely related (as in it relates to warnings):
- https://github.com/nilearn/nilearn/issues/3541
- https://github.com/nilearn/nilearn/issues/4668
- we ingore the flake8 bugbear rule related to warning stack level

I would love to see this implemented but I don't think we have quite the bandwidth at the moment: also we don't have a plan of actions on how to tackle these issues. Probably some "divide & conquer" approach working module by module and possibly having more nilearn specific warning classes.

But for the moment I would remove that pytest option: unless you are volunteering to fix all those tests? 😜

DimitriPapadopoulos: As an example, it is expected to have deprecation warnings in tests that test deprecated functions.

These warnings should be disabled somehow. I probably need to read [How to capture warnings](https://docs.pytest.org/en/stable/how-to/capture-warnings.html) - perhaps in a different PR.

Remi-Gau: OK then I will remove this line and merge the rest

Remi-Gau, line 112:
```suggestion
```

Remi-Gau (approved):

PR #4719, opened 2024-11-12
[ENH] run fit and transform of SurfaceMasker on list of Surface images

Remi-Gau:
- closes #4592
- stacked on top of #4692
- blocked by #4692

PR #4756, opened 2024-11-18
[ENH] Allow `Parcellations` to work with Surface API

man-shu:
- Closes None but relates to #4027 
- Blocked by:
  -  #4830 
  -  #4926
  -  #4937 

- accept SurfaceImage / SurfaceMasker as input to [_BaseDecomposition](https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/decomposition/_base.py#L247) class
- since we are updating [_BaseDecomposition](https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/decomposition/_base.py#L247) class to support surface images, this should also automatically allow DictLearning to work with them
- add corresponding tests


man-shu:
We would need #4719 here as well...

nilearn/regions/parcellations.py
```diff
@@ -409,7 +410,11 @@ def _raw_fit(self, data):
...
         else:
-            mask_ = safe_get_data(mask_img_).astype(bool)
+            if isinstance(mask_img_, SurfaceImage):
+                # concatenate arrays stored in both parts of the mask data
+                mask_ = np.concatenate(mask_img_.data.parts.values())
+            else:
+                mask_ = safe_get_data(mask_img_).astype(bool)
             shape = mask_.shape
             connectivity = image.grid_to_graph(
                 n_x=shape[0], n_y=shape[1], n_z=shape[2], mask=mask_
```
man-shu, line 420: Now [_BaseDecomposition](https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/decomposition/_base.py#L247) class is accepting SurfaceMasker almost everywhere, up to the point where we need `SurfaceMapsMasker`.

But as @bthirion suggested here: https://github.com/nilearn/nilearn/pull/4577#discussion_r1843821348, allowing wards clustering (or agglomerative clustering) would require more work.

man-shu: It's just calculating the edges between the nodes (that are voxels here), i.e., the adjacency matrix in a 3D volume.

IIUC, I think we could reuse what was done in rena clustering to calculate this connectivity/adjacency matrix for surfaces:
https://github.com/nilearn/nilearn/blob/4bcd033d30ca42d32ecdc408b24d3928112cea12/nilearn/regions/rena_clustering.py#L283

nilearn/decomposition/_base.py
```diff
@@ -197,9 +199,17 @@ def _mask_and_reduce(
     subject_n_samples = [subject_data.shape[0] for subject_data in data_list]
 
     n_samples = np.sum(subject_n_samples)
-    n_voxels = int(np.sum(safe_get_data(masker.mask_img_)))
+    # n_features is the number of True vertices in the mask if it is a surface
+    if isinstance(masker, SurfaceMasker):
+        n_features = (
```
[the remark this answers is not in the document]
man-shu: This isn't final yet, so will most probably be removed or refactored to `n_vertices` because we only need it for  surface

nilearn/regions/parcellations.py
man-shu: done!

man-shu:
Now, all tests should fail with the same error, which we can fix by implementing a SurfaceMapsMasker.

Will start another PR for that.

man-shu:
> * since we are updating [_BaseDecomposition](https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/decomposition/_base.py#L247) class to support surface images, this should also automatically allow DictLearning to work with them

Just FYI

nilearn/regions/parcellations.py
```diff
@@ -488,24 +553,40 @@ def transform(self, imgs, confounds=None):
...
-            memory=self.memory,
-            memory_level=self.memory_level,
-            verbose=self.verbose,
-        )
+
+        if isinstance(self.masker_.mask_img_, SurfaceImage):
+            imgs_list = imgs.copy()
+            masker = SurfaceLabelsMasker(
+                self.labels_img_,
+                mask_img=self.masker_.mask_img_,
```
man-shu, line 561: So turns out we need SurfaceLabelsMasker as well to work properly here. As far as this PR is concerned, SurfaceLabelsMasker missing two things as of now:
1. doesn't take mask_img to output region signals only for the vertices within the mask.
2. doesn't take confounds that are later used in signal cleaning

Will make separate PRs for each.

man-shu:
> Blocked by:
> * [[ENH] Implement surface maps masker #4830](https://github.com/nilearn/nilearn/pull/4830)
> * [[ENH] allow using confounds via fit_transform in SurfaceLabelsMasker #4926](https://github.com/nilearn/nilearn/pull/4926)
> * [[ENH[ Accept `mask_img` in `SurfaceLabelsMasker` #4937](https://github.com/nilearn/nilearn/pull/4937)

Update: we need these three to be merged to make progress on this

nilearn/regions/tests/test_parcellations.py
```diff
@@ -313,27 +314,40 @@ def test_transform_3d_input_images(affine_eye):
...
+            size=(mesh["left"].coordinates.shape[0], n_samples)
+        ),
+        "right": rng.standard_normal(
+            size=(mesh["right"].coordinates.shape[0], n_samples)
+        ),
+    }
+    surf_img = SurfaceImage(mesh=mesh, data=data)
+    parcellate = Parcellations(method=method, n_parcels=n_parcels)
+    # fit and transform the data
+    X_transformed = parcellate.fit_transform(surf_img)
```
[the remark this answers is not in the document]
man-shu: The assertion in line 345 fails

nilearn/decomposition/_base.py
```diff
@@ -198,9 +200,17 @@ def _mask_and_reduce(
...
+            masker.mask_img_.data.parts["left"].sum()
+            + masker.mask_img_.data.parts["right"].sum()
+        )
```
Remi-Gau, line 208:
```suggestion
        n_features = self.output_dimension_
```
if the masker is fitted you can do this I think

Remi-Gau: and if it's not fitted, I would just fit it

nilearn/decomposition/_base.py
```diff
@@ -464,15 +477,15 @@ def fit(
...
-            )
+        masker_type = "nii"
+        if isinstance(self.mask, (SurfaceMasker, SurfaceImage)):
+            masker_type = "surface"
 
-        self.masker_ = check_embedded_masker(self)
+        for img in imgs:
+            if isinstance(img, SurfaceImage):
+                masker_type = "surface"
+                break
```
Remi-Gau, line 487:
```suggestion
        if any(isinstance(x, SurfaceImage) for x in imgs):
           masker_type = "surface"

```

Remi-Gau: porbably mergeable with the block above

nilearn/decomposition/_base.py
```diff
@@ -497,15 +510,21 @@ def fit(
...
-        self.nifti_maps_masker_ = NiftiMapsMasker(
-            self.components_img_,
-            self.masker_.mask_img_,
-            resampling_target="maps",
-        )
+        if isinstance(self.masker_, SurfaceMasker):
+            self.surface_maps_masker_ = SurfaceMapsMasker(
+                self.components_img_, self.masker_.mask_img_
+            )
+            self.surface_maps_masker_.fit()
```
Remi-Gau, line 519: maybe better to have a single `maps_masker_` attributes and maybe make `nifti_maps_masker_` just return `self.maps_masker_` in the volume case.

along with a deprecation warning that this public attribute will be removed in the future

nilearn/decomposition/_base.py
```diff
@@ -522,7 +541,8 @@ def transform(self, imgs, confounds=None):
 
         Parameters
         ----------
-        imgs : iterable of Niimg-like objects
+        imgs : iterable of Niimg-like objects or
+        list of :obj:`~nilearn.surface.SurfaceImage`
```
Remi-Gau, line 545:
```suggestion
        imgs : iterable of Niimg-like objects or \
               list of :obj:`~nilearn.surface.SurfaceImage`
```

nilearn/decomposition/_base.py
```diff
@@ -598,7 +618,8 @@ def score(self, imgs, confounds=None, per_component=False):
...
```
Remi-Gau, line 622:
```suggestion
        imgs : iterable of Niimg-like objects or \
               :obj:`list` of :obj:`~nilearn.surface.SurfaceImage`
```

nilearn/regions/parcellations.py
```diff
@@ -116,6 +179,59 @@ def _labels_masker_extraction(img, masker, confound):
     return signals
 
 
+def _apply_surf_mask_on_labels(mask_data, labels_data, background_label=0):
+    """Apply mask to labels data.
+    Copy of nilearn.maskers.surface_labels_masker._apply_mask.
```
Remi-Gau, line 184: may be good to extract this somewhere else instead of copying it, no?

nilearn/regions/parcellations.py
```diff
@@ -116,6 +179,59 @@ def _labels_masker_extraction(img, masker, confound):
...
+    labels = np.unique(labels_data)
+    labels = labels[labels != background_label]
+
+    return labels_data, labels
+
+
+def _signals_to_surf_img_labels(signals, labels_img, mask_img):
+    """Transform signals to surface image labels, used to parallelize.
+    Copy of nilearn.maskers.surface_labels_masker.inverse_transform.
+    """
```
Remi-Gau, line 210: same here

nilearn/regions/parcellations.py
man-shu: I agree! Similar functions for volumes are in nilearn.regions.signal_extraction. So maybe that's the best place

Remi-Gau: I would say to maybe keep it as function in nilearn.maskers.surface_labels_masker and then think later if we can find it a proper home

nilearn/decomposition/_base.py
```diff
@@ -497,15 +505,29 @@ def fit(
...
+            )
+            # TODO: remove in 0.11.4
+            self.nifti_maps_masker_ = self.maps_masker_
+            warnings.warn(
+                FutureWarning,
+                message="The nifti_maps_masker_ attribute is deprecated and"
+                "will be removed in Nilearn 0.11.3. Please use "
+                "maps_masker_ instead.",
+                stacklevel=2,
+            )
```
Remi-Gau, line 529: maybe mode this into a property
```python
            @property
            def nifti_maps_masker_(self):
              warnings.warn(
                FutureWarning,
                message="The nifti_maps_masker_ attribute is deprecated and"
                "will be removed in Nilearn 0.11.3. Please use "
                "maps_masker_ instead.",
                stacklevel=2,
              )
              return self.maps_masker_
```

man-shu: Doing this refactoring made me realize that I didn't update _BaseDecomposition's transform and inverse_transform to work with surface masker. So now I am using `self.maps_masker_` in those too (and hence avoiding doing if else). That wouldn't work with the property like this then

man-shu:
I think this one is big enough, I will open another PR for DictLearning.

nilearn/maskers/surface_labels_masker.py
```diff
@@ -20,13 +20,15 @@
...
-    """Apply mask to labels data."""
-    labels_data[np.logical_not(mask_data.flatten())] = (
-        labels_masker.background_label
-    )
+def _apply_surf_mask_on_labels(mask_data, labels_data, background_label=0):
+    """Apply mask to labels data. Ensures that we only get the data back
+    according to the mask that was applied. So if some labels were removed,
+    we will only get the data for the remaining labels, the vertices that were
+    masked out will be set to the background label.
+    """
```
Remi-Gau, line 28:
```suggestion
    """Apply mask to labels data.
    
    Ensures that we only get the data back
    according to the mask that was applied. So if some labels were removed,
    we will only get the data for the remaining labels, the vertices that were
    masked out will be set to the background label.
    """
```

nilearn/regions/parcellations.py
```diff
@@ -5,17 +5,81 @@
 
 import numpy as np
 from joblib import Memory, Parallel, delayed
+from scipy.sparse import coo_matrix
 from sklearn.base import clone
 from sklearn.feature_extraction import image
 
-from nilearn.maskers import NiftiLabelsMasker
+from nilearn.maskers import NiftiLabelsMasker, SurfaceLabelsMasker
+from nilearn.maskers.surface_labels_masker import _signals_to_surf_img_labels
```
Remi-Gau, line 13:
```suggestion
from nilearn.maskers.surface_labels_masker import signals_to_surf_img_labels
```

if you are going to import in another 'subpackage' then it is not a function only private to the module it is in, so you should remove the leading `_`

Remi-Gau: you will have to rename it in nilearn.maskers.surface_labels_masker

bthirion (commented):
It deserves a whatsnew entry and a dedictaed example.
But thee xample could be in a future PR.

man-shu:
> deserves a whatsnew entry

Done!

Remi-Gau (approved):
Good with me.

Will update the changelog in the release PR I will open today.

Remi-Gau:
ha. I had not seen you post @man-shu

PR #4767, opened 2024-11-19
[ENH] add the option to provide path to labels tsv

### LINKED ISSUE #4710: pass labels for label maskers as path to a look up table mapping a region label and its name
Came to me while moving the surface label masker

Currently the [NiftiLabelMasker](https://nilearn.github.io/stable/modules/generated/nilearn.maskers.NiftiLabelsMasker.html#nilearn-maskers-niftilabelsmasker) (and the soon coming surface label masker) can accept a labels parameters as a list of strings, to give 'proper' names to the regions present in the label mask.

If the user passes labels then it's on them to make sure that the order of the labels is 'consistent' with the order of the regions in the label mask.

On our end we must:
- make run some extra-checks (for example [match between number of ROI and labels](https://github.com/nilearn/nilearn/blob/4f4730163097457cf9ddb5674ffd158ee8fa822e/nilearn/maskers/nifti_labels_masker.py#L295))
- [sanitize the labels](https://github.com/nilearn/nilearn/blob/4f4730163097457cf9ddb5674ffd158ee8fa822e/nilearn/maskers/nifti_labels_masker.py#L271)
- [make some assumptions (name of the 'background' value)](https://github.com/nilearn/nilearn/blob/4f4730163097457cf9ddb5674ffd158ee8fa822e/nilearn/maskers/nifti_labels_masker.py#L768)
- [construct the mapping ROI <--> ROI name](https://github.com/nilearn/nilearn/blob/4f4730163097457cf9ddb5674ffd158ee8fa822e/nilearn/maskers/nifti_labels_masker.py#L774) for internal tracking but alos because we show it in the reports

On the long term I would prefer for labels to be pased ONLY as a lookup table that corresponds to the BIDS dseg.tsv format:
https://bids-specification.readthedocs.io/en/stable/derivatives/imaging.html#common-image-derived-labels

The arguement could be a path to tsv file or its equivalent pandas dataframe.

This would :
- make the mapping explicit on the data side and not reliant on our code making assumptions
- increase BIDS compatibility and also compatibility with TemplateFlow
- give us a standardiwed name for background value

mtorabi59:
- Closes #4710 

- this PR will add the option to provide a path to a tsv file as a lookup table for label names, according to BIDS dseg.tsv format, instead of a list. This will make sure the ids and names correspond to each other, and also the background will be better recognized.


mtorabi59:
You can use the following code to test the new option:

    import csv
        generate_labeled_regions,
        generate_random_img,
    )
    import numpy as np 
    
    labels_as_path = True 
    
    data = [
        {'region id': 0, 'region name': 'Background'},
        {'region id': 1, 'region name': 'Region A'},
        {'region id': 2, 'region name': 'Region B'},
        {'region id': 3, 'region name': 'Region C'},
        {'region id': 4, 'region name': 'Region D'},
        {'region id': 5, 'region name': 'Region E'},
        {'region id': 6, 'region name': 'Region F'},
        {'region id': 7, 'region name': 'Region G'},
        {'region id': 8, 'region name': 'Region H'},
        {'region id': 9, 'region name': 'Region I'}
    ]
    
    with open('./regions.tsv', 'w', newline='') as file:
        writer = csv.DictWriter(file, fieldnames=['region id', 'region name'], delimiter='\t')
        writer.writeheader()
        writer.writerows(data)
    
    if labels_as_path:
        labels = './regions.tsv'
    else:
        labels = [
            'Background', 'Region A', 'Region B', 'Region C', 'Region D',
            'Region E', 'Region F', 'Region G', 'Region H', 'Region I'
        ]
    
    n_regions = 9
    length = 93
    shape_3d_default = (7, 8, 9)
    affine_eye = np.eye(4)
    
    shape1 = (*shape_3d_default, length)
    
    fmri_img, mask11_img = generate_random_img(
        shape1,
        affine=affine_eye,
    )
    
    labels_img = generate_labeled_regions(
        shape1[:3],
        affine=affine_eye,
        n_regions=n_regions,
    )
    
    masker = NiftiLabelsMasker(labels_img, labels=labels, resampling_target=None)
    signals = masker.fit().transform(fmri_img)
    
    print(masker.labels)
    print(masker._region_id_name)

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -553,25 +580,38 @@ def fit(
...
+        # find the region ids existing in the self.labels_img_
+        initial_region_ids = [
+            region_id
+            for region_id in np.unique(
+                _utils.niimg.safe_get_data(self.labels_img_)
+            )
+            if region_id != self.background_label
+        ]
+        if self._region_id_name is not None:
+            # check if initial_region_ids all exist in self._region_id_name
```
Remi-Gau, line 592: here I would do 2 checks:
- see if we have regions in the tsv that do not exist in the image
- see if we have regions in the image that do not exist in the tsv

I wonder if in the long term we could just throw warnings for those an not errors.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -553,25 +580,38 @@ def fit(
...
+            missing_ids = [
                 region_id
-                for region_id in np.unique(
-                    _utils.niimg.safe_get_data(self.labels_img_)
-                )
-                if region_id != self.background_label
+                for region_id in initial_region_ids
+                if region_id not in self._region_id_name
             ]
```
Remi-Gau, line 597:
```python
missing_ids = set(initial_region_ids) - set(self._region_id_name)
```

maybe using set operation is faster and clearer, no?

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -202,6 +203,32 @@ def __init__(
 
         self.background_label = background_label
         self._original_region_ids = self._get_labels_values(self.labels_img)
+        self._region_id_name = None
+
+        # if the labels is a path, it is the path to tsv file, so read it
+        if isinstance(labels, str):
+            try:
```
Remi-Gau, line 210: I think I would not go for a `try / except` here but maybe there is a reason I am missing.
was it to provide more meaningful error message than pandas' "file not found" ?

Remi-Gau (commented):
Did a first pass review.
Let me know if something is unclear.

Also I would suggest starting adding test now because it will help us see what makes sense or not.

Remi-Gau:
Another thing to do that will help us get a feel for this would be to update the fetcher of one of our atlases so it also creates the type of TSV we would expect users to provide.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -202,6 +203,32 @@ def __init__(
...
+                region_id_name = pd.read_csv(labels, sep="\t")
+                # Note that labels should include the background too
+                labels = region_id_name["region name"].tolist()
```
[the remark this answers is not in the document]
mtorabi59: I thought working with `self._region_id_name` as a dict would be easier. But if you have a strong preference we can make it a pandas dataframe instead.

nilearn/maskers/nifti_labels_masker.py
mtorabi59: It's because instead of throwing an error and stopping the code, I thought maybe it's better to throw a warning and set `self.labels` to `None`.

bthirion (commented):
LGTM

nilearn/maskers/nifti_labels_masker.py
mtorabi59: I think the first one is not critical, but it can be a warning.

nilearn/maskers/nifti_labels_masker.py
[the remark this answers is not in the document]
mtorabi59: Done!

PR #4768, opened 2024-11-19
Fix B028 errors

SamiLaayat:
- Closes #

We retrieved ruff via the command line by cloning the Git repository, and we ran ruff on all the projects to identify the B028 error (the one Rémi asked me to work on). The B028 error is triggered when a function (such as warnings.warn(), logging.warning(), etc.) is called without specifying the correct stacklevel. The stacklevel parameter is used to indicate which level of the call stack the warning or log should refer to.

This gave me a list of files containing B028 errors, along with the specific lines involved (at the bottom of the file, click the arrow to expand the errors). There are 125 occurrences detected by ruff.

What we propose with the addition of stacklevel is to make the warnings clearer, but this does not necessarily fix an underlying issue if one exists. In fact, the code can still execute without problems, but we are making the issues more visible. Following this initial improvement to highlight B028 errors more clearly, we will be able to focus on linting the code more easily, as we will have precise error locations.

The stacklevel=2 means that the warning will be associated with the caller of the code (the line above in the call stack).

### DIFF
--- a/nilearn/path/to/RuffB028.py
+++ b/nilearn/path/to/RuffB028.py
@@ -1,34 +1,77 @@
 import warnings
+
 warnings.warn(f"Backend changed to {new_backend}...", stacklevel=2)
-warnings.warn("Affine is different across subjects. Realignment on first subject", stacklevel=2)
-warnings.warn("Random walker only segments unlabeled areas, where labels == 0. No zero valued areas in labels were found.", stacklevel=2)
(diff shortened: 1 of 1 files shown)


Remi-Gau:
I think that in many cases we will have to check 'manually' check what the stacklevel to set so I don't think we want to 'blanket' apply `stacklevel=2`.

In the best case scenario, we want to raise warning at the level where users using can do something about it (like by changing something in one of the public function or methods they used).

Remi-Gau:
@SamiLaayat 
Will reply to you on the issue you opened but let's close the PR for now.

PR #4591, opened 2024-10-04
[ENH] make the number of vertices the first dimension of image data

jeromedockes:
- fixes #4569 
- supersedes #4576 

as discussed with @Remi-Gau

PR #4771, opened 2024-11-20
[FIX] Force the parts of PolyData to be 2D

Remi-Gau:
- Closes none

- make it explicit that left and right for PolyData must be 2D numpy array
  - We used to have SurfaceImage with shape (20484,): this is not allowed anymore and it will be SurfaceImage (1, 20484)
- add check to make that surface images passed for plotting have a single timepoint (shape = (1, x))
- adapt how 1D array are loaded / saved

### DIFF
--- a/doc/visual_testing/reporter_visual_inspection_suite.py
+++ b/doc/visual_testing/reporter_visual_inspection_suite.py
@@ -465,8 +465,6 @@ def report_surface_masker():
         mask.data.parts[part] = mask.data.parts[part] == 34
 
     masker = SurfaceMasker(mask)
-    img = load_nki(mesh_type="inflated")[0]
-
     masker.fit_transform(img)
     surface_masker_with_mask_report = masker.generate_report()
     surface_masker_with_mask_report.save_as_html(
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -59,8 +59,8 @@ destrieux = fetch_atlas_surf_destrieux()
 destrieux_atlas = SurfaceImage(
     mesh=fsaverage["pial"],
     data={
-        "left": np.array([destrieux["map_left"]]),
-        "right": np.array([destrieux["map_right"]]),
+        "left": destrieux["map_left"],
+        "right": destrieux["map_right"],
     },
 )
 labels = [x.decode("utf-8") for x in destrieux.labels]
@@ -94,7 +94,6 @@ print(f"Fsaverage5 sulcal curvature map: {fsaverage_curvature}")
 # %%
 # Extracting the seed time series
 # -------------------------------
-import numpy as np
 
 # Load resting state time series from nilearn
 timeseries = nki_dataset[0].data.parts[hemi].T
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -33,6 +33,8 @@ See :ref:`plotting` for more details on plotting tools.
 # -------------------
 
 # NKI resting state data from nilearn
+import numpy as np
+
 from nilearn.datasets import (
     fetch_atlas_surf_destrieux,
     load_fsaverage,
@@ -57,13 +59,13 @@ destrieux = fetch_atlas_surf_destrieux()
 destrieux_atlas = SurfaceImage(
     mesh=fsaverage["pial"],
     data={
-        "left": destrieux["map_left"],
-        "right": destrieux["map_right"],
+        "left": np.array([destrieux["map_left"]]),
+        "right": np.array([destrieux["map_right"]]),
     },
 )
 labels = [x.decode("utf-8") for x in destrieux.labels]
 
-parcellation = destrieux_atlas.data.parts[hemi]
+parcellation = destrieux_atlas.data.parts[hemi][0]
 
 # Fsaverage5 surface template
 fsaverage_meshes = load_fsaverage()
@@ -97,7 +99,7 @@ import numpy as np
 # Load resting state time series from nilearn
 timeseries = nki_dataset[0].data.parts[hemi].T
 
-# Coercing to float is required to avoid errors withj scipy >= 0.14.0
+# Coercing to float is required to avoid errors with scipy >= 0.14.0
 timeseries = timeseries.astype(float)
 
 # Extract seed region via label
@@ -127,8 +129,8 @@ stat_map[np.where(np.mean(timeseries, axis=1) == 0)] = 0
 from nilearn.plotting import plot_surf_roi, plot_surf_stat_map, show
 
 # Transform ROI indices in ROI map
-pcc_map = np.zeros(parcellation.shape[0], dtype=int)
-pcc_map[pcc_labels] = 1
+pcc_map = np.zeros((1, parcellation.shape[0]), dtype=int)
+pcc_map[0, pcc_labels] = 1
 
 plot_surf_roi(
     surf_mesh=nki_dataset[0].mesh,
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -33,8 +33,6 @@ See :ref:`plotting` for more details on plotting tools.
 # -------------------
 
 # NKI resting state data from nilearn
-import numpy as np
-
 from nilearn.datasets import (
     fetch_atlas_surf_destrieux,
     load_fsaverage,
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -92,6 +92,7 @@ print(f"Fsaverage5 sulcal curvature map: {fsaverage_curvature}")
 # %%
 # Extracting the seed time series
 # -------------------------------
+import numpy as np
 
 # Load resting state time series from nilearn
 timeseries = nki_dataset[0].data.parts[hemi].T
--- a/examples/08_experimental/README.rst
+++ b/examples/08_experimental/README.rst
@@ -1,5 +1,0 @@
-=================================
-Examples for experimental modules
-=================================
-
-These are examples focused on showcasing experimental features and are subject to change without any notice.
--- a/nilearn/datasets/struct.py
+++ b/nilearn/datasets/struct.py
@@ -23,7 +23,6 @@ from nilearn.surface import (
     FileMesh,
     PolyMesh,
     SurfaceImage,
-    load_surf_data,
 )
 
 MNI152_FILE_PATH = (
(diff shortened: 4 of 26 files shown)


nilearn/surface/surface.py
```diff
@@ -1242,15 +1245,15 @@ class PolyData:
...
     ----------
-    left : :obj:`numpy.ndarray` or :obj:`str` or :obj:`pathlib.Path` or None,\
-           default = None
+    left : 2D :obj:`numpy.ndarray` or :obj:`str` or :obj:`pathlib.Path` \
+           or None, default = None
 
-    right : :obj:`numpy.ndarray` or :obj:`str` or :obj:`pathlib.Path` or None,\
-           default = None
+    right : 2D :obj:`numpy.ndarray` or :obj:`str` or :obj:`pathlib.Path` \
+            or None, default = None
```
Remi-Gau, line 1252: Something that could be nicer for users, is that if a 1D numpy array then it is converted to a 2D array.

@man-shu @bthirion WDYT?

man-shu: I see, so this would ensure that the data arrays are 2D, even if there is only one sample/timepoint.

1. Could cause some test failures on #4770. So we should merge this one first.
2. Would the saving/writing be unaffected? For example, would the .gii files saved by nilearn still work with other software?

man-shu: Otherwise I am ok with the changes.

Remi-Gau: > 1. Could cause some test failures on

Yeah I was having a look at what you were doing and thought it may cause issues.

Remi-Gau: > I see, so this would ensure that the data arrays are 2D, even if there is only one sample/timepoint.

Yes but I think it may be easier to use in general if we can still instantiate a Surface image with 1 dim numpy array for one hemisphere but we just force them to be 2 dim.

Remi-Gau: Will wait for the test to all pass before I refactor things.

examples/01_plotting/plot_surf_stat_map.py
```diff
@@ -33,6 +33,8 @@
 # -------------------
 
 # NKI resting state data from nilearn
+import numpy as np
+
```
Remi-Gau, line 37:
```suggestion
```

nilearn/surface/surface.py
```diff
@@ -1744,14 +1759,21 @@ def from_volume(
...
+
         texture_right = vol_to_surf(
             volume_img,
             mesh.parts["right"],
             **vol_to_surf_kwargs,
             **right_kwargs,
         )
+        texture_right = texture_right.T
+        if texture_right.ndim == 1:
+            texture_right = np.array([texture_right])
```
Remi-Gau, line 1774:
```suggestion
```

man-shu (approved):
I am for the change simply because coming from volumetric images, it took a while to wrap my head around why the data was in 1D format.

However, I can't help but notice that it introduces some code that might not initially be obvious to the users. For example, adding a `[0]` to fetch the data even when there's only one sample or adding an extra pair of square brackets.

It would also be inconsistent with NIFTI images where we have 3 dimensions when there's only one sample and the 4th dim only comes into the picture when time gets involved.

So I would be interested in hearing what other people think about this.

doc/visual_testing/reporter_visual_inspection_suite.py
```diff
@@ -465,8 +465,6 @@ def report_surface_masker():
         mask.data.parts[part] = mask.data.parts[part] == 34
 
     masker = SurfaceMasker(mask)
-    img = load_nki(mesh_type="inflated")[0]
-
```
Remi-Gau, line 469: unrelated to the PR but realized this was already called above

examples/08_experimental/README.rst
Remi-Gau, line 1: forgot to remove it before

nilearn/datasets/struct.py
```diff
@@ -1143,12 +1142,8 @@ def load_fsaverage_data(
     img = SurfaceImage(
         mesh=fsaverage[mesh_type],
         data={
-            "left": load_surf_data(
-                fsaverage_data[f"{renaming[data_type]}_left"]
-            ),
-            "right": load_surf_data(
-                fsaverage_data[f"{renaming[data_type]}_right"]
-            ),
```
Remi-Gau, line 1151: `load_surf_data` will be called in `PolyData.__init__`

nilearn/glm/first_level/first_level.py
```diff
@@ -1011,7 +1011,7 @@ def _prepare_mask(self, run_img):
                 surf_data = {}
                 for part in run_img.mesh.parts:
                     surf_data[part] = np.ones(
-                        run_img.data.parts[part].shape[1], dtype=bool
+                        (1, run_img.data.parts[part].shape[1]), dtype=bool
```
Remi-Gau, line 1014:
```suggestion
                        run_img.data.parts[part].shape[1], dtype=bool
```

nilearn/maskers/_utils.py
```diff
@@ -65,9 +65,12 @@ def compute_mean_surface_image(img):
     """
     if img.shape[0] < 2:
         return img
+
+    data = {}
     for part, value in img.data.parts.items():
-        img.data.parts[part] = np.squeeze(value.mean(axis=0)).astype(float)
-    return img
+        data[part] = np.array([np.mean(value, axis=0).astype(float)])
```
Remi-Gau, line 71:
```suggestion
        data[part] = np.mean(value, axis=0).astype(float)
```

nilearn/surface/surface.py
```diff
@@ -1262,33 +1266,51 @@ def __init__(self, left=None, right=None):
...
+            if param is not None:
+                if not isinstance(param, np.ndarray):
+                    param = load_surf_data(param)
+                if param.ndim == 1:
+                    param = np.array([param])
+                    self._squeeze_on_save[hemi] = True
+                parts[hemi] = param
+        self.parts = parts
+
+        self._check_parts()
```
Remi-Gau, line 1284: The issue is that if a user set some parts manually after init and does not run _check_parts they may end up having invalid parts.

We already had this problem before, but this makes it more clear.

nilearn/surface/surface.py
```diff
@@ -1262,33 +1266,51 @@ def __init__(self, left=None, right=None):
                 "Either left or right (or both) must be provided."
             )
 
+        # to ensure that if 1D array are passed as input
+        # they will be saved as 1D array
+        self._squeeze_on_save = {"left": False, "right": False}
```
Remi-Gau, line 1271: Not too happy with this way of 'tracking' what the object was initialized with just to ensure that 1D array are saved properly to a gifti file, but at least it works...

Well unless, once again, some of the parts are manually set with a 1D array after init.

Remi-Gau:
@bthirion @jeromedockes @htwangtw @emdupre 
any opinions on this?

nilearn/surface/surface.py
bthirion: `self._squeeze_on_save` should be a Bool: there should not be a distinction between left and right hemispheres.
And I think it should be configurable at init.

bthirion (approved):
LGTM, thx.

Remi-Gau:
Will wait for CI to finish and merge

jeromedockes:
sorry I took a few days to respond but what's the rationale behind this? when the dimension of the image is one value per vertex -- for example a mask, a non-probabilistic atlas, a statistical image or anything that isn't a time series -- what is the benefit of adding an extra dimension? that would be like making all volume images 4d, it hides the actual dimension of the data and means we have to check that the last length is 1 and squeeze it whenever we have a 1d image (similar comment to that of @man-shu [here](https://github.com/nilearn/nilearn/pull/4771#pullrequestreview-2450964339) )

bthirion:
"Always 4D" makes things simpler in the sense that you don't need to figure out, but indeed we could also strive to have the best-suited data dimension.

Remi-Gau:
An example of the weird things that would happen before this PR.

Ran the following after `git checkout ad79dadf0`

```python
from nilearn.datasets import load_fsaverage
import numpy as np

mesh = load_fsaverage()["pial"]

# one time point image: impmlicit
data = {
    "left": np.ones((mesh.parts["left"].n_vertices)),
    "right": np.ones((mesh.parts["right"].n_vertices)),
}
img = SurfaceImage(mesh=mesh, data=data)

print(f"{img=}")
print(f"{img.shape[0]=}")

# one time point image: explicit
data = {
    "left": np.ones((1, mesh.parts["left"].n_vertices)),
    "right": np.ones((1, mesh.parts["right"].n_vertices)),
}
img = SurfaceImage(mesh=mesh, data=data)

print(f"{img=}")
print(f"{img.shape[0]=}")

# 2 time points image
data = {
    "left": np.ones((2, mesh.parts["left"].n_vertices)),
    "right": np.ones((2, mesh.parts["right"].n_vertices)),
}
img = SurfaceImage(mesh=mesh, data=data)

print(f"{img=}")
print(f"{img.shape[0]=}")
```

Gives me this:

```
img=<SurfaceImage (20484,)>
img.shape[0]=20484
img=<SurfaceImage (1, 20484)>
img.shape[0]=1
img=<SurfaceImage (2, 20484)>
img.shape[0]=2
```

> it hides the actual dimension of the data
To me I would say that this actually makes it explicit.

The analogy to volume does not quite work as the 'time' dimension comes last for volume data (x, y, z, **t**) but first for surfaces (**t**, v).

jeromedockes:
not after #4591 which makes t the last dimension, so the first dimension would always be n vertices.

as a user I expect the dimensions of image to match those of the data I passed or that I have in my Gifti file. I don't really see why they would change or why I am asked if they should be squeezed on save: just save the data and shape of what I passed.

so IMO as the maskers (AFAICT) were working OK in 1, 2 or more dimensions, forcing it to be 2 is a regression. more than 2 may not be useful but 1D certainly is (an atlas is not a time series with a single time point: it does not have a time dimension).

in any case it's not a big deal. pr #4591 has been updated to match the "always 2D" choice and should be ok to merge now

Remi-Gau:
> not after #4591 which makes t the last dimension, so the first dimension would always be n vertices.

yes true

> why I am asked if they should be squeezed on save: just save the data and shape of what I passed.

Agree that the squeeze on save is clunky.

> an atlas is not a time series with a single time point
When working with volume I would still stay that my image image contains a single volume

FYI I admit my matlab upbringing may be tainting my thinking on this:

```matlab
% create 3 array
>> a = ones(3,3,3);
% asking for the 4th dimension of that 3D array
>> size(a, 4)
ans =
     1
>> 
```

Remi-Gau:
I'd be OK to maybe revert force to 2D after #4591 is in.

bthirion:
I prefer to leave users control rather than coercing to 2D...

Remi-Gau:
OK I will revert after #4591 is merged then

Issue #4663, opened 2024-10-17
Seepd up doc build


Remi-Gau:
On top of the script that generates HTML reports to put in the documentation (https://github.com/nilearn/nilearn/blob/main/doc/visual_testing/reporter_visual_inspection_suite.py) that could probably be made faster, here is a list of the run time of all the examples from a recent full build of the doc (https://github.com/nilearn/nilearn/actions/runs/11383393419/job/31669645502?pr=4650#step:13:21362).

Could be useful if we want to optimize things.

Note that almost all the experimental are duplicates of examples that already exist so they will 'disappear' when we merge the new surface API.

computation time summary:
    - ../examples/02_decoding/plot_haxby_different_estimators.py:                        433.27 sec   2107.9 MB
    - ../examples/02_decoding/plot_haxby_understand_decoder.py:                          333.65 sec   1308.5 MB
    - ../examples/05_glm_second_level/plot_second_level_association_test.py:             239.81 sec    306.5 MB
    - ../examples/05_glm_second_level/plot_oasis.py:                                     212.28 sec   2824.8 MB
    - ../examples/02_decoding/plot_haxby_frem.py:                                        161.69 sec   1111.8 MB
    - ../examples/02_decoding/plot_haxby_full_analysis.py:                               156.00 sec   2071.1 MB
    - ../examples/07_advanced/plot_age_group_prediction_cross_val.py:                    152.12 sec   2476.6 MB
    - ../examples/02_decoding/plot_haxby_grid_search.py:                                 148.10 sec   1052.9 MB
    - ../examples/07_advanced/plot_beta_series.py:                                       136.41 sec    779.7 MB
    - ../examples/04_glm_first_level/plot_spm_multimodal_faces.py:                       135.56 sec   1151.9 MB
    - ../examples/02_decoding/plot_haxby_glm_decoding.py:                                130.10 sec   1096.5 MB
    - ../examples/04_glm_first_level/plot_first_level_details.py:                        129.60 sec    455.0 MB
    - ../examples/08_experimental/plot_surface_image_and_maskers.py:                     119.31 sec   9538.1 MB
    - ../examples/03_connectivity/plot_atlas_comparison.py:                              117.16 sec   2010.5 MB
    - ../examples/06_manipulating_images/plot_extract_rois_smith_atlas.py:               106.37 sec    505.4 MB
    - ../examples/07_advanced/plot_surface_bids_analysis.py:                              99.17 sec    730.5 MB
[...]

PR #4781, opened 2024-11-22
[TEST] try run full doc build with later python [full doc]

Remi-Gau:
- Closes #

-
-


Remi-Gau:
buuild took 94 minutes which a bit faster (5 minutes on the optimist end) than the recent full build on main 

that may explain part of the recent "slow down of the build" but not the >20 min jump in buiid time
https://github.com/nilearn/nilearn/issues/4663#issuecomment-2489420992

PR #4785, opened 2024-11-24
[STY] Implicit enumerate calls

DimitriPapadopoulos:
Use `enumerate(...)` instead of `range(len(...))` where possible.


bthirion (approved):
LGTM.

Remi-Gau (approved):

Issue #4686, opened 2024-10-23
Further refactoring of surface fixture


Remi-Gau:
It still feels weird to me to use the last four as fixture and not as functions that we store somewhere in nilearn._utils ? Maybe nilearn._utils.testing?

But I don't think this is urgent. Let's open a side issue for these and move on to merge.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/4657#discussion_r1812079537_
            
--- 

- [x] see if some fixtures should not be some general testing utility functions
- [x] try to refactor with 'old'  surface code testing utilities

PR #4791, opened 2024-11-25
[Enh] Remove old surface

### LINKED ISSUE #4789: [ENH] Move/remove/drop support for old surface-related stuff
Had a quick chat with @jeromedockes, and he mentioned that it is a bit confusing to know which objects are old and which ones are new because we have both old and new surface objects in the same module right now.

Old surface image:
https://github.com/nilearn/nilearn/blob/e897ec4196015a05ea216b3b6e665a0c8ae46149/nilearn/surface/surface.py#L29

New surface image:
https://github.com/nilearn/nilearn/blob/e897ec4196015a05ea216b3b6e665a0c8ae46149/nilearn/surface/surface.py#L1674

Should we move or remove them from here?

man-shu:
- Closes #4789

- remove old `Mesh` and `Surface` objects and related `check_*` functions
- update `load_surf_mesh` function to output new `InMemoryMesh` object
- remove tests using old `Mesh` and `Surface` objects and related `check_*` functions
- update tests to use `mesh.coordinates` and `mesh.faces` attributes instead of `mesh[0]` and `mesh[1]`
- update old test fixtures to output new `InMemoryMesh` object instead of a list
- [TODO] update plotting related tests that use old surface objects
- [TODO] update docstring

### DIFF
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -4,8 +4,6 @@ import abc
 import gzip
 import pathlib
 import warnings
-from collections import namedtuple
-from collections.abc import Mapping
 from pathlib import Path
 
 import numpy as np
@@ -21,12 +19,6 @@ from nilearn._utils import stringify_path
 from nilearn._utils.niimg_conversions import check_niimg
 from nilearn._utils.path_finding import resolve_globbing
 from nilearn.image import get_data, load_img, resampling
-
-# Create a namedtuple object for meshes
-Mesh = namedtuple("mesh", ["coordinates", "faces"])
-
-# Create a namedtuple object for surfaces
-Surface = namedtuple("surface", ["mesh", "data"])
 
 
 def _uniform_ball_cloud(n_points=20, dim=3, n_monte_carlo=50000):
@@ -563,13 +555,12 @@ def vol_to_surf(
     img : Niimg-like object, 3d or 4d.
         See :ref:`extracting_data`.
 
-    surf_mesh : :obj:`str`, :obj:`pathlib.Path`, :obj:`numpy.ndarray`, or Mesh
+    surf_mesh : :obj:`str`, :obj:`pathlib.Path`, :obj:`numpy.ndarray`, or
+                :obj:`SurfaceMesh`
         Either a file containing surface :term:`mesh` geometry
         (valid formats are .gii or Freesurfer specific files
         such as .orig, .pial, .sphere, .white, .inflated)
-        or two Numpy arrays organized in a list,
-        tuple or a namedtuple with the fields "coordinates" and "faces", or
-        a Mesh object with "coordinates" and "faces" attributes.
+        or a SurfaceMesh object with "coordinates" and "faces" attributes.
 
     radius : :obj:`float`, default=3.0
         The size (in mm) of the neighbourhood from which samples are drawn
@@ -975,17 +966,17 @@ def load_surf_mesh(surf_mesh):
     Parameters
     ----------
     surf_mesh : :obj:`str`, :obj:`pathlib.Path`, or \
-        :obj:`numpy.ndarray` or Mesh
+        :obj:`numpy.ndarray` or SurfaceMesh
         Either a file containing surface :term:`mesh` geometry
         (valid formats are .gii .gii.gz or Freesurfer specific files
         such as .orig, .pial, .sphere, .white, .inflated)
         or two Numpy arrays organized in a list,
         tuple or a namedtuple with the fields "coordinates" and "faces",
-        or a Mesh object with "coordinates" and "faces" attributes.
+        or a SurfaceMesh object with "coordinates" and "faces" attributes.
 
     Returns
     -------
-    mesh : Mesh
+    mesh : SurfaceMesh
         With the fields "coordinates" and "faces", each containing a
         :obj:`numpy.ndarray`
 
@@ -1010,14 +1001,14 @@ def load_surf_mesh(surf_mesh):
             # See https://github.com/nilearn/nilearn/pull/3235
             if "cras" in header:
                 coords += header["cras"]
-            mesh = Mesh(coordinates=coords, faces=faces)
+            mesh = SurfaceMesh(coordinates=coords, faces=faces)
         elif surf_mesh.endswith("gii"):
             coords, faces = _gifti_img_to_mesh(load(surf_mesh))
-            mesh = Mesh(coordinates=coords, faces=faces)
+            mesh = SurfaceMesh(coordinates=coords, faces=faces)
         elif surf_mesh.endswith("gii.gz"):
             gifti_img = _load_surf_files_gifti_gzip(surf_mesh)
             coords, faces = _gifti_img_to_mesh(gifti_img)
-            mesh = Mesh(coordinates=coords, faces=faces)
+            mesh = SurfaceMesh(coordinates=coords, faces=faces)
         else:
             raise ValueError(
                 "The input type is not recognized. "
@@ -1033,7 +1024,7 @@ def load_surf_mesh(surf_mesh):
     elif isinstance(surf_mesh, (list, tuple)):
         try:
             coords, faces = surf_mesh
-            mesh = Mesh(coordinates=coords, faces=faces)
+            mesh = SurfaceMesh(coordinates=coords, faces=faces)
         except Exception:
             raise ValueError(
                 "If a list or tuple is given as input, "
@@ -1046,7 +1037,7 @@ def load_surf_mesh(surf_mesh):
             )
     elif hasattr(surf_mesh, "faces") and hasattr(surf_mesh, "coordinates"):
         coords, faces = surf_mesh.coordinates, surf_mesh.faces
-        mesh = Mesh(coordinates=coords, faces=faces)
+        mesh = SurfaceMesh(coordinates=coords, faces=faces)
 
     else:
         raise ValueError(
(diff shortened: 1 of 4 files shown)


nilearn/surface/surface.py
```diff
@@ -1423,6 +1246,39 @@ def __init__(self, coordinates, faces):
         self.faces = faces
         self.n_vertices = coordinates.shape[0]
 
+    def __getitem__(self, index):
```
man-shu, line 1249: We can allow `InMemoryMesh` to be subscriptable with [0] and [1] and throw a future deprecation warning like this

nilearn/surface/tests/test_surface.py
```diff
@@ -1155,3 +1064,13 @@ def test_polydata_error():
 def test_polymesh_error():
     with pytest.raises(ValueError, match="Either left or right"):
         PolyMesh(left=None, right=None)
+
+
+def test_future_warning_to_fetch_mesh_coordinates_and_faces():
```
man-shu, line 1069: Testing it the warning here

nilearn/surface/surface.py
man-shu: Or we could just support this subscriptability forever...

Remi-Gau:
> this whole deprecation should come after the release.

would make things easier

nilearn/surface/surface.py
```diff
@@ -968,160 +964,12 @@ def _gifti_img_to_mesh(gifti_img):
...
-        data = load_surf_data(surface[1])
-    else:
-        raise ValueError(
-            "Wrong parameter `surface` in `load_surface`. "
-            "Please refer to the documentation for more information."
-        )
-    return Surface(mesh, data)
-
-
 def check_mesh(mesh):
```
man-shu, line 967: Ok so this function is only used in `plot_img_on_surf` to ensure that we only plot Nifti stat maps on a FreeSurfer-like mesh.

I think it could be named better. Something like `check_mesh_is_fsaverage`. I can add a deprecation warning saying that it will be renamed. 

Or we skip deprecating this one because as @Remi-Gau pointed out here: https://github.com/nilearn/nilearn/issues/4789#issuecomment-2497914202 it wasn't shown in the public docs.

Remi-Gau: no deprecation needed: not in the doc and not in the [surface/__init__.py](https://github.com/nilearn/nilearn/blob/main/nilearn/surface/__init__.py)

bthirion (approved):
LGTM. Need to check green CI though.

nilearn/plotting/html_surface.py
```diff
@@ -596,9 +570,11 @@ def view_surf(
     if surf_map is None:
         surf_map = np.ones(len(surf_mesh[0]))
     else:
-        surf_mesh, surf_map = surface.check_mesh_and_data(surf_mesh, surf_map)
+        surf_mesh, surf_map = surface.surface.check_mesh_and_data(
+            surf_mesh, surf_map
+        )
     if bg_map is not None:
-        _, bg_map = surface.check_mesh_and_data(surf_mesh, bg_map)
+        _, bg_map = surface.surface.check_mesh_and_data(surf_mesh, bg_map)
```
Remi-Gau, line 577: maybe for another PR but maybe we should just import the functions we need and not the whole module to avoid doing this kind `surface.surface.function_with_an_explicit_name_but_oh_no_now_are_beyond_the_allowed_line_length`

nilearn/plotting/tests/test_html_surface.py
```diff
@@ -96,17 +97,17 @@ def test_get_vertexcolor():
 
 
 def test_check_mesh():
-    mesh = html_surface.check_mesh("fsaverage5")
-    assert mesh is html_surface.check_mesh(mesh)
+    mesh = check_mesh("fsaverage5")
+    assert mesh is check_mesh(mesh)
```
Remi-Gau, line 101: ok yeah definitely we need a better name than check_mesh

man-shu:
Guys, please don't review yet! I have some more changes incoming

nilearn/plotting/html_surface.py
man-shu: 🤣

nilearn/surface/tests/test_surface.py
```diff
@@ -39,42 +50,43 @@
...
-    @property
-    def mesh(self):
-        return self._mesh
 
-    @property
-    def data(self):
-        return self._data
+def test_check_mesh_and_data(rng):
+    coords, faces = generate_surf()
+    mesh = InMemoryMesh(coords, faces)
```
Remi-Gau, line 67:
```suggestion
    mesh = generate_surf()
```

I think you already modified generate_surf to return a InMemoryMesh

nilearn/surface/tests/_testing.py
```diff
@@ -4,6 +4,7 @@
 from scipy.spatial import Delaunay
 
 from nilearn.conftest import _rng
+from nilearn.surface import InMemoryMesh
 
 
 def generate_surf():
```
Remi-Gau, line 10: Maybe for a post realease PR but I think that those functions could maybe be turned into fixtures, no?

Also it will be worth checking if they are not redundant with some of the fixtures we already have

See https://github.com/nilearn/nilearn/issues/4686

nilearn/surface/surface.py
```diff
@@ -1200,39 +1045,99 @@ def check_mesh_and_data(mesh, data):
...
+            raise ValueError(
+                "If a list or tuple is given as input, "
+                "it must have two elements, the first is "
+                "a Numpy array containing the x-y-z coordinates "
+                "of the mesh vertices, the second is a Numpy "
+                "array containing  the indices (into coords) of "
+                "the mesh faces. The input was a list with "
+                f"{len(surf_mesh)} elements."
+            )
+    elif hasattr(surf_mesh, "faces") and hasattr(surf_mesh, "coordinates"):
```
Remi-Gau, line 1124: why not check directly if the type is InMemoryMesh or SurfaceMesh?

nilearn/surface/surface.py
```diff
@@ -1200,39 +1045,99 @@ def check_mesh_and_data(mesh, data):
...
-                    - a numpy array with shape (n_vertices,)
-                    or (n_time_points, n_vertices)
+    surf_mesh : :obj:`str`, :obj:`pathlib.Path`, or \
+        :obj:`numpy.ndarray` or SurfaceMesh
+        Either a file containing surface :term:`mesh` geometry
+        (valid formats are .gii .gii.gz or Freesurfer specific files
+        such as .orig, .pial, .sphere, .white, .inflated)
+        or two Numpy arrays organized in a list,
+        tuple or a namedtuple with the fields "coordinates" and "faces",
+        or a SurfaceMesh object with "coordinates" and "faces" attributes.
```
Remi-Gau, line 1061:
```suggestion
        or a SurfaceMesh or InMemoryMesh object .
```

man-shu:
@Remi-Gau @bthirion 
I think if we remove the old Surface and Mesh without any warning, it might break some users' code. 
So I suggest we deprecate those too and support them until 0.13.0

Please see commit https://github.com/nilearn/nilearn/pull/4791/commits/e7fce1a298c1ae8907d7325be6030b478ef5c8dc for a narrower diff on this.

nilearn/surface/tests/_testing.py
man-shu: Yes, we should do that. 

I actually like these better because they are more flexible -- we can create meshes with however many vertices, and the data can be more random.

The ones we currently use are more static.

nilearn/surface/surface.py
man-shu: As it turns out, you cannot instantiate a SurfaceMesh object with coordinates and faces. 
It only works with InMemoryMesh.

The issue is that `load_surf_mesh` allows users to load a single hemisphere. It seems to me that the whole point of the new surface API was to discourage users from doing that and instead force them to specify which hemisphere they want to load/define.

However, we use this function internally within InMemoryMesh, and it does make sense to use it there. So in the long term, we should probably remove this function from public API and only use it privately.

man-shu: Same applies for `load_surf_data`

man-shu: PolyMesh and PolyData or simply dicts seem to be the objects we want users to be using instead.

Remi-Gau: ok thanks for the clarification

Remi-Gau:
FYI doc build failure is unrelated: will merge main to confirm

man-shu:
Following a discussion with @Remi-Gau, we think it's fine to just remove the old Mesh and Surface objects. 

So don't review this yet, I have some changes to make.

man-shu:
Ok this is ready for review @bthirion @Remi-Gau

nilearn/plotting/_utils.py
```diff
@@ -44,8 +44,8 @@ def check_surface_plotting_inputs(
     if isinstance(surf_mesh, SurfaceImage):
         raise TypeError(
             "'surf_mesh' cannot be a SurfaceImage instance. ",
-            "Accepted types are: str, list of two numpy.ndarray, Mesh, "
-            "PolyMesh, or None.",
+            "Accepted types are: str, list of two numpy.ndarray, "
```
Remi-Gau, line 47: I think that in the long term we will also deprecate the 2 numpy arrays as input and only allow InMemoryMesh and PolyMesh, no?

But not here because that would break the API.

nilearn/surface/surface.py
```diff
@@ -968,25 +965,110 @@ def _gifti_img_to_mesh(gifti_img):
...
+    }.difference(mesh.keys())
+    if missing:
+        raise ValueError(
+            f"{missing} {'are' if len(missing) > 1 else 'is'} "
+            "missing from the provided mesh dictionary"
+        )
+    return mesh
+
+
+def check_mesh_and_data(mesh, data):
```
Remi-Gau, line 1002: some of those checks look very much like some of the ones we do in PolyMesh, PolyData or SurfaceImage: maybe worth refactoring later...

nilearn/plotting/_utils.py
man-shu: Yes, I agree.

Remi-Gau:
Let's see what CI says

nilearn/surface/surface.py
man-shu: Yes, these functions are mostly used in the plotting module (which was using the old surface API and didn't have those checks otherwise)

man-shu:
CI looks happy too

Remi-Gau (approved):

PR #4793, opened 2024-11-26
[ENH] revert coercing surface data to 2D

### LINKED ISSUE #4792: Revert coercing surface data to 2D
I prefer to leave users control rather than coercing to 2D...

_Originally posted by @bthirion in https://github.com/nilearn/nilearn/issues/4771#issuecomment-2497962288_

Remi-Gau:
- Closes #4792

- remove squeeze on data
- add some tests for different shapes
- remove coercing surface data to 2D
- adapt other tests and examples

### DIFF
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -120,7 +120,7 @@ for i in range(timeseries.shape[0]):
     stat_map[i] = stats.pearsonr(seed_timeseries, timeseries[i])[0]
 
 # Re-mask previously masked nodes (medial wall)
-medial_wall_vertices = np.where(np.mean(timeseries, axis=1) == 0)
+medial_wall_vertices = np.mean(timeseries, axis=1) == 0
 stat_map[medial_wall_vertices] = 0
 
 # %%
--- a/examples/01_plotting/plot_surf_stat_map.py
+++ b/examples/01_plotting/plot_surf_stat_map.py
@@ -63,7 +63,7 @@ destrieux_atlas = SurfaceImage(
 )
 labels = [x.decode("utf-8") for x in destrieux.labels]
 
-parcellation = destrieux_atlas.data.parts[hemi][0]
+parcellation = destrieux_atlas.data.parts[hemi]
 
 # Fsaverage5 surface template
 fsaverage_meshes = load_fsaverage()
@@ -95,7 +95,7 @@ print(f"Fsaverage5 sulcal curvature map: {fsaverage_curvature}")
 import numpy as np
 
 # Load resting state time series from nilearn
-timeseries = nki_dataset[0].data.parts[hemi].T
+timeseries = nki_dataset[0].data.parts[hemi]
 
 # Coercing to float is required to avoid errors with scipy >= 0.14.0
 timeseries = timeseries.astype(float)
@@ -120,7 +120,8 @@ for i in range(timeseries.shape[0]):
     stat_map[i] = stats.pearsonr(seed_timeseries, timeseries[i])[0]
 
 # Re-mask previously masked nodes (medial wall)
-stat_map[np.where(np.mean(timeseries, axis=1) == 0)] = 0
+medial_wall_vertices = np.where(np.mean(timeseries, axis=1) == 0)
+stat_map[medial_wall_vertices] = 0
 
 # %%
 # Display ROI on surface
@@ -190,7 +191,7 @@ plot_surf_stat_map(
     colorbar=True,
     bg_map=fsaverage_sulcal,
     bg_on_data=True,
-    cmap="Spectral",
+    cmap="bwr",
     threshold=0.5,
     title="Threshold and colormap",
 )
@@ -209,7 +210,7 @@ plot_surf_stat_map(
     hemi=hemi,
     view="lateral",
     colorbar=True,
-    cmap="Spectral",
+    cmap="bwr",
     threshold=0.5,
     title="Plotting without background",
 )
@@ -234,9 +235,8 @@ plot_surf_stat_map(
     threshold=0.5,
     colorbar=True,
     output_file=output_dir / "plot_surf_stat_map.png",
+    cmap="bwr",
 )
-
-show()
 
 # %%
 # References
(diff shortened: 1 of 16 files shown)


bthirion (commented):
LGTM overall.

Remi-Gau:
@jeromedockes 
do you have time to do a cross check on this one ?

Remi-Gau:
> I'm unsure about the PR because you're reverting more that the "coerce to 2D" part.

well I had not introduced the coerce to 2D just because the data curator in me likes consistency, it was actually solving a couple of issues internally: so I have had to introduced fixes for the tests / examples that started failing when I removed it.

Remi-Gau:
will flag the few things that are "extra"

examples/01_plotting/plot_surf_stat_map.py
```diff
@@ -120,7 +120,8 @@
     stat_map[i] = stats.pearsonr(seed_timeseries, timeseries[i])[0]
 
 # Re-mask previously masked nodes (medial wall)
-stat_map[np.where(np.mean(timeseries, axis=1) == 0)] = 0
+medial_wall_vertices = np.where(np.mean(timeseries, axis=1) == 0)
+stat_map[medial_wall_vertices] = 0
```
Remi-Gau, line 124: unrelated: just to make it easier to read

jeromedockes:
> @jeromedockes 
> do you have time to do a cross check on this one ?

sure, I'll have a look :)

examples/01_plotting/plot_surf_stat_map.py
```diff
@@ -190,7 +191,7 @@
     colorbar=True,
     bg_map=fsaverage_sulcal,
     bg_on_data=True,
-    cmap="Spectral",
+    cmap="bwr",
```
Remi-Gau, line 194: using diverging colormap with only 2 colors in the rest of the example

examples/07_advanced/plot_surface_image_and_maskers.py
Remi-Gau, line 1: This example was failing in the doc build in main:

- main changes is line 149

examples/07_advanced/plot_surface_image_and_maskers.py
```diff
@@ -51,10 +60,13 @@
             cmap="bwr",
             symmetric_cmap=True,
             bg_on_data=True,
+            vmin=vmin,
+            vmax=vmax,
+            bg_map=fsaverage_sulcal,
```
Remi-Gau, line 65: force same vmin and vmax across all subplots to have the same range and avoid changes in baseline due to differences between hemispheres

Remi-Gau: also force to have bg_map for esthethics

examples/07_advanced/plot_surface_image_and_maskers.py
```diff
@@ -51,10 +60,13 @@
...
         )
 fig.set_size_inches(6, 8)
 
-# show()
+show()
```
Remi-Gau, line 69: otherwise the example would not show any figure

examples/07_advanced/plot_surface_image_and_maskers.py
```diff
@@ -142,11 +147,9 @@
 # as we do for volume images.
 from nilearn.decoding import Decoder
 
-img = load_nki()[0]
-
 # create some random labels
 rng = np.random.RandomState(0)
-y = rng.choice([0, 1], replace=True, size=img.shape[0])
+y = rng.choice([0, 1], replace=True, size=img.shape[1])
```
Remi-Gau, line 152: fix that make sure the example can run

nilearn/datasets/func.py
```diff
@@ -1871,13 +1871,11 @@ def load_nki(
     ):
         logger.log(f"Loading subject {i} of {n_subjects}.", verbose=verbose)
 
-        left_data = load_surf_data(left)
-        right_data = load_surf_data(right)
         img = SurfaceImage(
             mesh=fsaverage[mesh_type],
             data={
-                "left": left_data,
-                "right": right_data,
```
Remi-Gau, line 1880: minor refactor: SurfaceImage will take care of calling load_surf_data if filenames are passed

Remi-Gau:
@bthirion I have flagged / explained all the extra changes in this PR: all the others were downstream consequence of not having A 2D array all the time.

nilearn/maskers/tests/test_surface_labels_masker.py
```diff
@@ -150,7 +150,7 @@ def test_surface_label_masker_check_output(surf_mesh, rng):
     surf_img = SurfaceImage(surf_mesh(), data)
     signal = masker.transform(surf_img)
 
-    assert signal.shape == (surf_img.shape[1], masker.n_elements_)
+    assert signal.shape == (1, masker.n_elements_)
```
man-shu, line 153: Wondering whether the output signal being 2D after masking is consistent with what we do with volume images? Only relevant when the input has one time point.

Would always be 2D otherwise.

man-shu (approved):
Otherwise LGTM!

nilearn/maskers/tests/test_surface_labels_masker.py
Remi-Gau: not sure but I would say that we can be flexible in what we accept as input but we should be predictible in what give as output: so I don't think the shape of the output should depend on the shape of the input.

examples/01_plotting/plot_surf_stat_map.py
bthirion: Do we need the np.where here ?

bthirion (commented):
LGTM.

examples/01_plotting/plot_surf_stat_map.py
Remi-Gau: not sure: it's always been there

see stable doc
https://nilearn.github.io/stable/auto_examples/01_plotting/plot_surf_stat_map.html#extracting-the-seed-time-series

Remi-Gau: tested locally: not needed apparently

bthirion (approved):
LGTM

Remi-Gau:
OK will merge now but I think we will release only on Monday so @jeromedockes if you see some horror in here, you still have a bit of time.

PR #4802, opened 2024-11-27
[ENH] allow to ignore certain BIDS entities when using query_bids

### LINKED ISSUE #4801: Localizer BIDS dataset derivative not BIDS compliant
This caused some examples to fail because our the BIDS derivatives files in our example dataset are not proper derivatives: they do not have a space entity.

- examples/07_advanced/plot_beta_series.py
- examples/07_advanced/plot_bids_analysis.py

see for example: https://github.com/nilearn/nilearn/actions/runs/12038962141/job/33569168884#step:13:27052

Will fix in separate PR.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/4507#issuecomment-2503775909_

Remi-Gau:
- relates to #4801 

- allow query_bids to not return filename containing `entity` when passed filter contain `entityname = ""`
- add test
- adapt first level from bids
- adapt example

### DIFF
--- a/examples/07_advanced/plot_bids_analysis.py
+++ b/examples/07_advanced/plot_bids_analysis.py
@@ -17,6 +17,11 @@ More specifically:
 3. Fit a second level model on the fitted first level models.
    Notice that in this case the preprocessed :term:`bold<BOLD>`
    images were already normalized to the same :term:`MNI` space.
+
+.. note::
+
+      We are only using a subset of participants from the dataset
+      to lower the run time of the example.
 """
 
 from nilearn import plotting
@@ -59,7 +64,12 @@ task_label = "languagelocalizer"
     models_events,
     models_confounds,
 ) = first_level_from_bids(
-    data.data_dir, task_label, img_filters=[("desc", "preproc")], n_jobs=2
+    data.data_dir,
+    task_label,
+    img_filters=[("desc", "preproc")],
+    n_jobs=2,
+    space_label="",
+    sub_labels=["01", "02", "03", "04"],  # comment to run all subjects
 )
 
 # %%
@@ -103,9 +113,16 @@ p001_unc = norm.isf(0.001)
 
 # %%
 # Prepare figure for concurrent plot of individual maps.
-import matplotlib.pyplot as plt
+from math import ceil
 
-fig, axes = plt.subplots(nrows=2, ncols=5, figsize=(8, 4.5))
+import matplotlib.pyplot as plt
+import numpy as np
+
+ncols = 3
+nrows = ceil(len(models) / ncols)
+
+fig, axes = plt.subplots(nrows=nrows, ncols=ncols, figsize=(8, 4.5))
+axes = np.atleast_2d(axes)
 model_and_args = zip(models, models_run_imgs, models_events, models_confounds)
 for midx, (model, imgs, events, confounds) in enumerate(model_and_args):
     # fit the GLM
@@ -117,9 +134,10 @@ for midx, (model, imgs, events, confounds) in enumerate(model_and_args):
         colorbar=False,
         threshold=p001_unc,
         title=f"sub-{model.subject_label}",
-        axes=axes[int(midx / 5), int(midx % 5)],
+        axes=axes[int(midx / ncols), int(midx % ncols)],
         plot_abs=False,
         display_mode="x",
+        cmap="bwr",
     )
 fig.suptitle("subjects z_map language network (unc p<0.001)")
 plotting.show()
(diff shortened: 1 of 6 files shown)


Remi-Gau:
@tsalo @htwangtw can you have a look at this one maybe?
I know you rely more on pybids for your BIDS things but you may have opinions about this new behavior of get_bids_files.

For context this is necessary because one of our examples has 'BIDS' derivatives that are not BIDS compliant (no space entity) but I figured that it may be generally useful to have a way to fetch files that DO NOT have specific entity in general.

examples/07_advanced/plot_bids_analysis.py
```diff
@@ -50,6 +51,17 @@
 # in the :term:`BIDS` dataset.
 # To get the first level models we only have to specify the dataset directory
 # and the task_label as specified in the file names.
+#
+# .. warning::
+#
+#   When using :func:`~nilearn.glm.first_level.first_level_from_bids` below
+#   we only fetch files without the space entity in their filename
+#   ``space_label=""```.
+#   This is because our preprocessed data is not fully BIDS comlpliant.
```
bthirion, line 60:
```suggestion
#   This is because our preprocessed data is not fully BIDS compliant.
```

examples/07_advanced/plot_surface_bids_analysis.py
```diff
@@ -53,12 +50,24 @@
 # in the :term:`BIDS` dataset.
 # To get the first level models we only have to specify the dataset directory
 # and the task_label as specified in the file names.
+#
+# .. warning::
+#
+#   When using :func:`~nilearn.glm.first_level.first_level_from_bids` below
+#   we only fetch files without the space entity in their filename
+#   ``space_label=""```.
+#   This is because our preprocessed data is not fully BIDS comlpliant.
```
bthirion, line 59:
```suggestion
#   This is because our preprocessed data is not fully BIDS compliant.
```

bthirion (commented):
The long term solution would be to fix the imported dataset, right ?

Remi-Gau:
> The long term solution would be to fix the imported dataset, right ?

yes that would be good too but the changes from this PR can also be useful outside the narrow "nilearn dev need a hack for poorly named example data" usecase. 😜

man-shu (approved):
LGTM!

nilearn/interfaces/bids/query.py
```diff
@@ -206,10 +206,15 @@ def get_bids_files(
...
-        ('task', 'localizer').
+        Filters are of the form (field, label).
+        Only one filter per field allowed.
+        A file that does not match a filter will be discarded.
+        Filter examples would be
+        ``('ses', '01')``,
+        ``('dir', 'ap')`` or
+        ``('task', 'localizer')``.
+        If an empty string is passed ``('space', '')``,
+        then only files that do NOT contain that entity are returned.
```
emdupre, line 217: Can you give an example use case (here or in the doc itself) for how to use this filtering other than for the `space` entity ? I'm struggling to think of when I'd use this, but that's likely just my lack of imagination !

Remi-Gau: It could apply to any entities in the raw dataset too: 

So pick any from one of those filename templates 
https://bids-website.readthedocs.io/en/latest/getting_started/folders_and_files/files.html#func

and imagine any data horror story you wish:

Data was reconstructed using different algorithm on the scanner side but the 'default' one was not 'flagged' with a `rec` entity in the filename (because "that's obvious that no rec entity means it is the base case!" said the person acquiring the data).

Same scenario when using contrast agent (`ce`), different acquisition parameters (`acq`), direction (`dir`)...

SO now if you want to select your "base case" you need to have a way to say: get me the files without that entity...

I agree that all these may still seem like edge cases but I have seen too many edge cases in the world of BIDS...

emdupre: Thanks for clarifying ! I can definitely see these edge-case scenarios, but I'm struggling to imagine a common use case and (as always my avoid-code-creep-at-all-cost conservatism is coming out when I say) I'm hesitant to continue expanding our BIDS functionality with the goal of covering edge-cases. 

I'm pro a short term fix for the examples, if needed, but I don't want to push that to user facing functionality (which would then have to go through deprecation cycles) if we think it's really only for edge-cases / until we fix the example datasets !

bthirion (approved):
LGTM.

nilearn/interfaces/bids/query.py
Remi-Gau: OK I will keep the change but remove any advertisement of it

emdupre: Perfect, thank you !

Issue #4805, opened 2024-11-27
[ENH] Add Tests to Better Cover the Cleaning Behavior of Maskers


ekhella:
Description:

Following discussions in [PR #3437](https://github.com/nilearn/nilearn/pull/3437), it was identified that the cleaning functionality in maskers is not sufficiently covered by tests. Specifically:

The cleaning functionality relies heavily on nilearn.signal.clean, but there are currently limited tests that validate its behavior when integrated with different masker classes (e.g., NiftiMasker, NiftiSpheresMasker).
Arguments passed to clean should be thoroughly tested across scenarios, especially with edge cases like t_r as an integer.

ekhella:
Proposed Changes:

Add unit tests in nilearn/maskers/tests/test_base_masker.py and other relevant test files to:
Cover different configurations of clean_kwargs passed to clean.
Test various filter parameter values (e.g., "cosine", "butterworth") and their edge cases.
Consider adding a new parameter like clean_args (similar to mask_args) to improve clarity and align the API with scikit-learn's best practices.
Review and expand the tests in nilearn.maskers.tests.test_base_masker to ensure comprehensive coverage of cleaning functionality.

quentden:
Hello, my name is Quentin and I work on Nilearn as part of a subject at my engineering school. That's why I would like to know some more details about the purpose of this issue if possible ?

Thanks in advance.
Quentin

PR #4830, opened 2024-11-29
[ENH] Implement surface maps masker

### LINKED ISSUE #4829: [ENH] Implement SurfaceMapsMasker to allow `Parcellations` to work with surface data
Now, all tests should fail with the same error, which we can fix by implementing a SurfaceMapsMasker.

Will start another PR for that.

_Originally posted by @man-shu in https://github.com/nilearn/nilearn/issues/4756#issuecomment-2507463369_

man-shu:
- Closes #4829
- Blocker for #4756

- A new `SurfaceMapsMasker` object:
  - [x] implement fit_transform
  - [x] implement inverse_transform
    - [x] will return unmasked img as of now, should return masked if mask is provided
  - [x] input validation
    - [x] check dimensions of maps_img, mask_img, img
  - [x] allow memory caching
  - [x] logging
  - [x] tests

Will add reporting stuff in another PR
  - ~[x] implement generate_report and everything else related to reporting~
    - ~[x] test the reporting with an example~
    - ~[x] validate different inputs for displayed_maps param in generate_report~


nilearn/conftest.py
```diff
@@ -484,6 +484,33 @@ def surf_label_img(surf_mesh):
     return SurfaceImage(surf_mesh(), data)
 
 
+@pytest.fixture
+def surf_maps_img(surf_mesh):
+    """Return a sample surface map image using the sample mesh.
+    Has 4 different overlapping regions per hemisphere.
```
bthirion, line 490: Wouldn't it be better to have different numbers of regions per hemisphere ?

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,480 @@
...
+                  shape: (number of scans - number of volumes removed) \
+                  for explicit index, or (number of scans) for binary mask, \
+                  default=None
+            sample_mask to pass to :func:`nilearn.signal.clean`.
+
+        Returns
+        -------
+        output : :obj:`numpy.ndarray`
+            Signal for each element.
+            shape: (img data shape, total number of vertices)
```
bthirion, line 250: does not sound right.

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,480 @@
...
+            Keyword arguments to be passed
+            to :func:`nilearn.signal.clean`
+            called within the masker.
+
+    Attributes
+    ----------
+        maps_img_ : :obj:`~numpy.ndarray`
+            The maps image converted to a numpy array by concatenating the \
+            data of both hemispheres/parts.
+            shape: (n_vertices, n_regions)
```
bthirion, line 92: What is the distinction between n_regions and self.n_elements ?

bthirion (commented):
This looks quit good, but coverage could still be improved.

nilearn/conftest.py
man-shu: Hmm... I agree that's a possibility in maps images, but I don't think we allow having different "n_timepoints" across hemispheres in SurfaceImage as of now. See:

https://github.com/nilearn/nilearn/blob/5271b3234ba450205060349b7cd401044d099220/nilearn/surface/surface.py#L1219

nilearn/maskers/surface_maps_masker.py
man-shu: yep need to update all docstrings

nilearn/conftest.py
man-shu: On that note, I don't think we have a probabilistic atlas for surfaces. I wanted to write an example to test out the reports...

nilearn/maskers/_utils.py
```diff
@@ -125,3 +125,32 @@ def concatenate_surface_images(imgs):
     output = SurfaceImage(mesh=imgs[0].mesh, data=output_data)
 
     return output
+
+
+def check_surface_data_ndims(img, dim, var_name="img"):
```
man-shu, line 130: Added this function to validate the dimensions of input images, and I noticed we are not doing that in SurfaceLabelsMasker.

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,502 @@
...
+                : self.maps_img.data.parts["left"].shape[0], :
+            ],
+            "right": vertex_signals[
+                : self.maps_img.data.parts["right"].shape[0], :
+            ],
+        }
+
+        return SurfaceImage(mesh=self.maps_img.mesh, data=vertex_signals)
+
+    def generate_report(self, displayed_maps=10):
```
Remi-Gau, line 408: @man-shu 

should we keep the report part for another PR?

Would make it easier to review (shorter PR) and we seem to have some report issues to fix anyway?

man-shu: Oops you are right! We are at >800 lines as of now. Will try to figure out how I can move some of the commits to another PR

man-shu: A lot of them are the HTML report btw

nilearn/reporting/data/html/report_body_template_surfacemapsmasker.html
```diff
@@ -0,0 +1,115 @@
...
+  function displayPreviousMap() {
+    document.getElementById("map" + current_map_idx).style["display"] = "none";
+    current_map_idx = current_map_idx - 1;
+    if (current_map_idx < 0) {
+      current_map_idx = number_maps - 1;
+    }
+    document.getElementById("map" + current_map_idx).style["display"] = "block";
+    document.getElementById("comp").innerHTML = displayed_maps[current_map_idx];
+  }
+</script>
```
Remi-Gau, line 114: if we are reusing code from the nifti maps masker it may be useful to create a partial to make the html more modular

man-shu:
> Will add reporting stuff in another PR

Just FYI

man-shu:
Ok now we are 100% ready for review @Remi-Gau @bthirion

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def test_surface_maps_masker_fit_transform_shape(
+    surf_maps_img, surf_img, surf_mask
+):
+    """Test that the fit_transform method returns the expected shape."""
+    # TODO: remove after #4897 is merged
+    surf_mask = surf_mask()
+    surf_mask.data.parts["left"] = surf_mask.data.parts["left"].squeeze()
+    surf_mask.data.parts["right"] = surf_mask.data.parts["right"].squeeze()
```
man-shu, line 15: We forgot to squeeze the extra dimension in the fixtures in an older PR. So I had to do this here and in several other places.
* But #4897 is taking care of that...

nilearn/conftest.py
```diff
@@ -484,6 +484,33 @@ def surf_label_img(surf_mesh):
     return SurfaceImage(surf_mesh(), data)
 
 
+@pytest.fixture
+def surf_maps_img(surf_mesh):
```
Remi-Gau, line 488: maybe put this one in nilearn/maskers/tests/test_surface_maps_masker.py

unless we use the fixture in more than one subpackage it may be easier to keep the fixture 'scope' as small as possible

This may be something to clean more globally as part of the surface fixture refactoring.

nilearn/conftest.py
```diff
@@ -484,6 +484,33 @@ def surf_label_img(surf_mesh):
...
+                [1, 1, 1, 0, 0, 0],
+            ]
+        ),
+        "right": np.asarray(
+            [
+                [1, 0, 0, 0, 1, 1],
+                [1, 1, 0, 0, 1, 1],
+                [0, 1, 1, 0, 1, 1],
+                [1, 1, 1, 0, 0, 1],
+                [0, 0, 1, 0, 0, 1],
```
Remi-Gau, line 507: given that surface maps can also be 'probability' maps should with have value between 0 and 1 here to make this fixture more general?

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,415 @@
...
+            Only relevant for the report figures.
+
+        clean_args : :obj:`dict` or None, default=None
+            Keyword arguments to be passed
+            to :func:`nilearn.signal.clean`
+            called within the masker.
+
+    Attributes
+    ----------
+        maps_img_ : :obj:`~numpy.ndarray`
```
Remi-Gau, line 87: in nifti masker this is a nifti image, so I think it should be a surface image here:

https://nilearn.github.io/dev/modules/generated/nilearn.maskers.NiftiMapsMasker.html#nilearn-maskers-niftimapsmasker

image

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,415 @@
...
+        logger.log(
+            msg=f"loading regions from {self.maps_img.__repr__()}",
+            verbose=self.verbose,
+        )
+        # check maps_img data is 2D
+        check_surface_data_ndims(self.maps_img, 2, "maps_img")
+
+        self.maps_img_ = np.concatenate(
+            list(self.maps_img.data.parts.values()), axis=0
+        )
```
Remi-Gau, line 184: TODO
use the concatenate function from nilearn.maskers._utils

man-shu: But that one concatenates different surface images provided in a list. 

Here I want to concatenate the data from two hemispheres of the _same_ surface image for later computation.

nilearn/maskers/surface_maps_masker.py
man-shu: But then there would be no difference between the fitted `maps_img_` and input `maps_img` and I would have to store the hemisphere-concatenated numpy array separately (leading to more memory usage).

man-shu: With surface maskers, I think the `.fit()` method would not be doing much except for input validation and SurfaceImage -> array conversions.

In nifti maskers, it is doing the resampling between mask and maps in addition.

nilearn/maskers/surface_maps_masker.py
Remi-Gau: my bad

nilearn/maskers/surface_maps_masker.py
Remi-Gau: > But then there would be no difference between the fitted maps_img_ and input maps_img

That does necessarily shock me: I'd prefer to have some consistency across maskers in terms of the type of their attributes. Less cognitive load in general I think for devs and users.

I would do the concatenation across hemisphere at transform time.

nilearn/maskers/surface_maps_masker.py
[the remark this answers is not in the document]
man-shu: this is for **maps**

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,415 @@
...
+        if self.mask_img is not None:
+            logger.log(
+                msg=f"loading regions from {self.mask_img.__repr__()}",
+                verbose=self.verbose,
+            )
+            check_same_n_vertices(self.maps_img.mesh, self.mask_img.mesh)
+            check_surface_data_ndims(self.mask_img, 1, "mask_img")
+            self.mask_img_ = np.concatenate(
+                list(self.mask_img.data.parts.values()), axis=0
+            )
```
man-shu, line 196: this is for **mask**

nilearn/maskers/surface_maps_masker.py
man-shu: I see... So should we still return a fitted `self.maps_img_`?

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def _check_fitted(self):
+        if not self.__sklearn_is_fitted__():
+            raise ValueError(
+                f"It seems that {self.__class__.__name__} "
+                "has not been fitted."
+            )
+
+    @property
+    def _get_concatenated_maps_and_mask(self):
```
man-shu, line 200: @Remi-Gau, added a property to concatenate parts from mask and maps, used in both transform and inverse_transform methods.

Let me know if this is better.

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -0,0 +1,221 @@
+import numpy as np
+import pytest
+
+from nilearn.maskers import SurfaceMapsMasker
+from nilearn.surface import SurfaceImage
+
```
Remi-Gau, line 6: Can you add test to check we are compliant with sklearn ?

see https://github.com/nilearn/nilearn/blob/6a2107674849cb19d6916cba6793c07a6f483b45/nilearn/maskers/tests/test_surface_masker.py#L11-L43

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,426 @@
...
+        parameters["clean_args"] = self.clean_args
+
+        if self.memory is None:
+            self.memory = Memory(location=None)
+
+        # apply mask if provided
+        # and then extract signal via least square regression
+        if mask_data is not None:
+            region_signals = cache(
+                linalg.lstsq,
```
bthirion, line 290: Just wondering what happens if the least squares fit fails, due to redundancy in the Surface maps ?

bthirion (commented):
LGTM overall.

nilearn/maskers/surface_maps_masker.py
man-shu: I think scipy handle that and would throw a LinAlgError:

https://github.com/scipy/scipy/blob/92d2a8592782ee19a1161d0bf3fc2241ba78bb63/scipy/linalg/_basic.py#L1191

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -0,0 +1,263 @@
...
+    }
+    # multiply with random "probability" values
+    data = {part: data[part] * rng.random(data[part].shape) for part in data}
+    return SurfaceImage(surf_mesh(), data)
+
+
+# tests for scikit-learn compatibility
+extra_valid_checks = [
+    "check_no_attributes_set_in_init",
+    "check_parameters_default_constructible",
```
man-shu, line 43: It seems this test enforces that we cannot have any required parameters while instantiating estimators. 

Would be a problem with maskers e.g. Nifti/SurfaceLabelsMasker or Nifti/SurfaceMapsMasker which have `labels_img` and `maps_img` as required parameters.

* Related to #4538

man-shu: This would mean change in behavior and consequently deprecation cycles for all maskers?

Remi-Gau: yup at the moment I think we want to see how to maximize following the sklearn recommendations without breaking any API

when we introduce new things though let's try to comply as much as possible, no?

man-shu: Indeed, that makes sense. 

But just for my understanding, we would rather provide `maps_img` during `.fit()`?

man-shu: And should it be None at instantiation or should it not exist at all in the variables?

nilearn/maskers/surface_maps_masker.py
Remi-Gau: feels weird to have a property return a tuple

one private method and maybe one private property for each ?

nilearn/maskers/tests/test_surface_maps_masker.py
Remi-Gau: > we would rather provide maps_img during .fit()?
ah no I think that there are some cases where we can have our own rules: in this case, I would pass this at instantiation

man-shu: Yeah I think then we need to have at least a default value of `None` for `maps_img` and `labels_img`:

https://scikit-learn.org/stable/developers/develop.html#instantiation

Screenshot 2024-12-09 at 2 22 05 PM

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,426 @@
...
+        self._report_content = {
+            "description": (
+                "This report shows the input surface image "
+                "(if provided via img) overlaid with the regions provided via "
+                "maps_img."
+            ),
+            "n_vertices": {},
+            "number_of_regions": 0,
+            "summary": {},
+        }
```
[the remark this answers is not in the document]
man-shu: I also noticed that for SurfaceLabelsMasker, we are not actually doing the scikit-learn checks on SurfaceLabelsMasker, but SurfaceMasker:

https://github.com/nilearn/nilearn/blob/4808962a12026e0f0723e17073b7ad540b18be3e/nilearn/maskers/tests/test_surface_labels_masker.py#L29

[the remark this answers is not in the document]
Remi-Gau: OK that's my fault: wanna fix that in a small separate PR?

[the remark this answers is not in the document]
man-shu: > Setting those in fit() should also help better comply with the sklearn recommendation regarding instanciation of estimators.

Moved to fit()

nilearn/maskers/surface_maps_masker.py
```diff
@@ -0,0 +1,428 @@
...
+        self.detrend = detrend
+        self.low_pass = low_pass
+        self.high_pass = high_pass
+        self.t_r = t_r
+        self.memory = memory
+        self.memory_level = memory_level
+        self.verbose = verbose
+        self.reports = reports
+        self.cmap = cmap
+        self.clean_args = clean_args
```
Remi-Gau, line 137: do you think you can add a smoke test that uses the cleaning part of the masker: it seems that we are not testing this systematically for our maskers.

man-shu: Done!

But it seems that the way we expect arguments for clean is not consistent between Nifti and Surface maskers as of now:

NiftiMasker:
Screenshot 2024-12-09 at 5 21 34 PM

SurfaceMasker:
Screenshot 2024-12-09 at 5 22 12 PM

man-shu: * I think there is a related issue #4805

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def test_surface_maps_masker_fit_transform_shape(
+    surf_maps_img, surf_img_2d, surf_mask
+):
+    """Test that the fit_transform method returns the expected shape."""
+    # TODO: squeeze surf_mask to 1D
+    surf_mask = surf_mask()
+    surf_mask.data.parts["left"] = surf_mask.data.parts["left"].squeeze()
+    surf_mask.data.parts["right"] = surf_mask.data.parts["right"].squeeze()
```
man-shu, line 88: We forgot to squeeze the surf_mask fixture, will make a PR for that

nilearn/maskers/surface_maps_masker.py
man-shu: I feel like we do this across hemisphere concatenation fairly often, maybe a function in maskers/_utils would be better

man-shu: However, having a method within SurfaceImage like `.get_data()` would be the most ideal I think...
* relates to this comment: https://github.com/nilearn/nilearn/pull/4884#pullrequestreview-2482561037

nilearn/maskers/_utils.py
```diff
@@ -128,6 +128,52 @@ def concatenate_surface_images(imgs):
...
+    n_dim_left = img.data.parts["left"].ndim
+    n_dim_right = img.data.parts["right"].ndim
+    if not all(x == dim for x in [n_dim_left, n_dim_right]):
+        raise ValueError(
+            f"Data for each hemisphere of {var_name} should be {dim}D, "
+            f"but found {n_dim_left}D for left and {n_dim_right}D for right."
+        )
+
+
+def concat_extract_surface_data_parts(img):
```
man-shu, line 160: Using this function to concatenate the parts of mask_img and maps_img now

Remi-Gau:
converting to draft as it is not on the roadmap for 0.11.1

man-shu:
This is ready for review otherwise @Remi-Gau @bthirion

Remi-Gau:
> Oh, so I can't tag an object in changelog if I create that object in the same PR, @Remi-Gau ?
> 
> See the doc build failure: https://github.com/nilearn/nilearn/actions/runs/12297964193/job/34324774761?pr=4830#step:13:1475:
> 
> ```
> /home/runner/work/nilearn/nilearn/doc/changes/latest.rst:20: WARNING: py:obj reference target not found: nilearn.maskers.SurfaceMapsMasker
> ```

you can but I totally forgot that we need to update the list of things that will be in the public API in the doc: 

just need to update this file and the ref should work
https://github.com/nilearn/nilearn/blob/3d9017833e62701280932304aff21e44ca75aeb4/doc/modules/maskers.rst?plain=1#L30

Remi-Gau:
let's update the doc ref and see how the doc page looks

man-shu:
So this one is ok to go I think

Remi-Gau:
HTML page of the doc:
https://output.circle-artifacts.com/output/job/215d700a-c4db-46a2-b542-30a56e2ae0b5/artifacts/0/dev/modules/generated/nilearn.maskers.SurfaceMasker.html#nilearn.maskers.SurfaceMasker.transform

Remi-Gau (approved):

man-shu:
Thanks for the review @Remi-Gau!
@bthirion also need one from you

bthirion (approved):
LGTM, thx !

PR #4865, opened 2024-12-03
[REF] Simple refactor of permuted_ols

Remi-Gau:
- Relates to #3655 

- extract input checks
- return early
- remove some else blocks


bthirion (commented):
The changes LGTM, thx.
The diff just need a bit more coverage.

PR #4881, opened 2024-12-04
[MAINT] optional import of get_cluster_table

### LINKED ISSUE #4877: reporting/get_clusters_table.py has no matplotlib dependency and yet cannot be imported if matplotlib is not installed
there is no maplotlib import in here https://github.com/nilearn/nilearn/blob/main/nilearn/reporting/get_clusters_table.py

yet because this module is in the reporting subpackage we get:

```python
>>> from nilearn.reporting import get_clusters_table
/home/remi/github/nilearn/nilearn/nilearn/_utils/helpers.py:25: UserWarning: Some dependencies of nilearn.plotting package seem to be missing.
They can be installed with:
  warnings.warn(
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 1
----> 1 from nilearn.reporting import get_clusters_table

File ~/github/nilearn/nilearn/nilearn/reporting/__init__.py:10
      6 # Author: Martin Perez-Guevara, Elvis Dohmatob, 2017
      8 from nilearn._utils.helpers import _set_mpl_backend
---> 10 _set_mpl_backend()
     12 ###############################################################################
     13 from nilearn.reporting.get_clusters_table import get_clusters_table

File ~/github/nilearn/nilearn/nilearn/_utils/helpers.py:23, in _set_mpl_backend()
     21 # We are doing local imports here to avoid polluting our namespace
     22 try:
---> 23     import matplotlib
     24 except ImportError:
     25     warnings.warn(
     26         "Some dependencies of nilearn.plotting package seem to be missing."
     27         "\nThey can be installed with:\n"
     28         " pip install 'nilearn[plotting]'"
     29     )

ModuleNotFoundError: No module named 'matplotlib'
```

I am tempted to say that this module should be moved out of reporting of more easily change what nilearn.reporting import if matploblib is present or not.

Remi-Gau:
- Closes #4877 

- Add parameter `message` to _set_mpl_backend`
- Delete file `nilearn/plotting/tests/test_matplotlib_backend.py` and move all the tests in the file to `nilearn/_utils/tests/test_helpers.py` 
- Update imports in `nilearn/_utils/tests/test_helpers.py`
- Update conftest to ignore only `reporting/glm_reporter.py` and `reporting/html_report.py` modules instead of whole reporting package when matplotlib is not installed.
- Make only `get_cluster_table` public from reporting package when matplotlib is not installed


Remi-Gau:
@hndgzkn 
if you have some time I'd like your take on this: one part of reporting has no dependency on matplotlib so it should be importable and testable even when matplotlib is not there

this PR is just quick draft 

it should also change conftest.py to cover tests get_cluster_table when matplotlib is installed

hndgzkn:
> @hndgzkn if you have some time I'd like your take on this: one part of reporting has no dependency on matplotlib so it should be importable and testable even when matplotlib is not there

I have some suggesitions. Can I push to your branch?

nilearn/_utils/helpers.py
```diff
@@ -6,7 +6,7 @@
 OPTIONAL_MATPLOTLIB_MIN_VERSION = "3.3.0"
 
 
-def _set_mpl_backend():
+def _set_mpl_backend(message):
```
hndgzkn, line 9: I've added warning as parameter so that we can modify the message depending on the calling package.

hndgzkn (approved):

nilearn/_utils/tests/test_helpers.py
```diff
@@ -23,6 +30,74 @@ def _mock_args_for_testing_replace_parameter():
...
+    reason="Test requires matplotlib not to be installed.",
+)
+def test_should_raise_warning_if_mpl_not_installed():
+    with (
+            pytest.warns(UserWarning, match="matplotlib is not installed"),
+            pytest.raises(
+                ModuleNotFoundError, match="No module named 'matplotlib'"
+            ),
+    ):
+        _set_mpl_backend(WARNING)
```
hndgzkn: Added to test `_set_mpl_backend` with message.

nilearn/_utils/tests/test_helpers.py
```diff
@@ -23,6 +30,74 @@ def _mock_args_for_testing_replace_parameter():
...
+
+
+@pytest.mark.skipif(
+    not is_matplotlib_installed(),
+    reason="Test requires matplotlib to be installed.",
+)
+@patch("matplotlib.__version__", "0.0.0")
+def test_should_raise_import_error_for_version_check():
+    with pytest.raises(ImportError, match="A matplotlib version of at least"):
+        _set_mpl_backend(WARNING)
```
hndgzkn, line 104: As `_set_mpl_backend` is moved in `helpers.py`, I moved these tests to `test_helpers.py`.

hndgzkn:
@Remi-Gau All tests seem to fail except the one with oldest dependencies. As far as I understand, it is due the new version 1.6.0 release of scikit-learn.

Remi-Gau:
> @Remi-Gau All tests seem to fail except the one with oldest dependencies. As far as I understand, it is due the new version 1.6.0 release of scikit-learn.

yup I just opened a PR to fix

We are starting to run tests to improve our API compliance with sklearn recommendations. But the latest versions of sklearn are also updating those things a bit, so we are playing catch up.

That also means that the number of tests we are running (and the number of expected test failures) has recently exploded: but it will get better.

bthirion (commented):
Some minor stuff. LGTM otherwise

nilearn/conftest.py
```diff
@@ -22,7 +22,9 @@
     SurfaceImage,
 )
 
-collect_ignore = ["datasets/data/convert_templates.py"]
+collect_ignore = [
```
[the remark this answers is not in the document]
hndgzkn: I had put other modules here accidentally, then I moved them to right place. I'll fix formatting.

nilearn/_utils/helpers.py
hndgzkn: I've put another warning in `reporting.__init__` file as below:

```
        "nilearn.reporting.make_glm_report and nilearn.reporting.HTMLReport "
        "requires 'matplotlib' to be installed."
        "\nIt can be installed with:\n"
        " pip install 'matplotlib'"
```

The standard message was:
```
    "Some dependencies of nilearn.plotting package seem to be missing."
    "\nThey can be installed with:\n"
    " pip install 'nilearn[plotting]'"
```

I find it more explicit from the users' perspective. It seems strange to me to see the dependency problem for `nilearn.plotting` when I try to import from `nilearn.reporting`.

However I am not sure if reporting requires other dependencies than `matplotlib`. I'll verify that. In either case, I would suggest to customize the warning depending on the origin of the problem.

Finally, if you think that it's ok to return the same message for all, I can revert changes.

bthirion (commented):
LGTM, but a few lines are not covered.

nilearn/_utils/tests/test_helpers.py
```diff
@@ -23,6 +34,93 @@ def _mock_args_for_testing_replace_parameter():
...
+    # Warning should be raised to inform user of the backend switch.
+    with pytest.warns(UserWarning, match="Backend changed to backend_2..."):
+        _set_mpl_backend()
+
+
+@pytest.mark.skipif(
+    not is_matplotlib_installed(),
+    reason="Test requires matplotlib to be installed.",
+)
+@patch("matplotlib.use")
```
Remi-Gau, line 87: that's new to me!! 😄

hndgzkn:
@Remi-Gau Do you have any idea why tests fail? 

In my local it fails for the first run, for the second run it passes. 
If I run the tests one by one, it passes.  

I thought there might be a problem with pytest capturing warnings.

hndgzkn:
The problem is not there, it is about imports. Not necessary to look at it for the moment.

hndgzkn:
@Remi-Gau The problem with tests is that, once a module is imported from `nilearn.reporting` package, the warning is raised, and if you want to reimport the same module or another module, it is not raised any more, as `nilearn.reporting` package is already there with only `get_clusters_table` available (for the case matplotlib is not installed).  

When I run tests individually, each test passes. However, for two tests importing from `nilearn.reporting`, if I run them consecutively, the second one fails.

As a quick solution, I would suggest to keep these changes temporarily to make it possible to import `nilearn.reporting.get_clusters_table` removing the test that imports for a second time, thus causes failure.

And as a long term solution, I think it might be good to have a look at all packages that import from plotting (I see that maskers, surface, dataset) and refactor code to separate parts that require plotting and make the implementation dependant to an interface rather then the plotting package itself. 

How does it sound to you?

bthirion:
You suggestion sounds tight.

bthirion (approved):
LGTM, thx.

Remi-Gau:
Cannot approve my own pr but good with me.
Thanks for taking over @hndgzkn

PR #4884, opened 2024-12-04
[ENH]  add max and min property to PolyData and SurfaceImage

### LINKED ISSUE #4878: add min and max properties to PolyData and SurfaceImage
Wopuld help to have a sortcut to find the max and min across hemispheres:

Would help avoid this:

https://github.com/nilearn/nilearn/blob/9492ea67cce7f1f62281adef4469f3e7f2b0d0bf/examples/07_advanced/plot_surface_image_and_maskers.py#L77-L80

and do

```python
vmax = np.absolute([mean_img.max, mean_img.min]).max()
```

Remi-Gau:
- Closes #4878 

- add min and max property to get min and max across hemispheres

### DIFF
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -1667,6 +1667,16 @@ class SurfaceImage:
         """Shape of the data."""
         return self.data.shape
 
+    @property
+    def max(self):
+        """Maximum of the data across hemispheres."""
+        return self.data.max
+
+    @property
+    def min(self):
+        """Minimum of the data across hemispheres."""
+        return self.data.min
+
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__} {self.shape}>"
 
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -1239,6 +1239,16 @@ class PolyData:
             else (sum_vertices,)
         )
 
+    @property
+    def max(self):
+        """Maximum of the data across all parts."""
+        return max(x.ravel().max() for x in self.parts.values())
+
+    @property
+    def min(self):
+        """Minimum of the data across all parts."""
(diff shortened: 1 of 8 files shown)


nilearn/surface/surface.py
```diff
@@ -1239,6 +1239,16 @@ def shape(self):
             else (sum_vertices,)
         )
 
+    @property
+    def max(self):
+        """Maximum of the data across all parts."""
+        return max(x.ravel().max() for x in self.parts.values())
```
Remi-Gau, line 1245:
```suggestion
        return max(x.max() for x in self.parts.values())
```

nilearn/surface/surface.py
```diff
@@ -1239,6 +1239,16 @@ def shape(self):
...
+
+    @property
+    def min(self):
+        """Minimum of the data across all parts."""
+        return min(x.ravel().min() for x in self.parts.values())
```
Remi-Gau, line 1250:
```suggestion
        return min(x.min() for x in self.parts.values())
```

nilearn/surface/tests/test_surface.py
```diff
@@ -801,6 +808,11 @@ def test_surface_image_shape(surf_img, shape):
     assert surf_img(shape).shape == (9, shape)
 
 
+def test_surface_image_min_max(surf_img):
```
man-shu, line 811: Would it make sense to also test for 2D data?

bthirion (commented):
The PR LGTM in general, but it opens a ton of questions: why do this for mean/max and not any statistic that could be computed from the data (mean, sum, median, quantiles...).
And why not do it on 3D images too ?
To me it would be better to give an easy access to the data rather than statistics.

Remi-Gau:
Closing for now as per our discussion yesterday.

leaving the issue open for a bit longer in case want to revisit.

Issue #4856, opened 2024-12-02
[DOC] method `set_output` mentioned but not yet implemented


psadil:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe your proposed suggestion in detail.

The documentation currently mentions the [`set_output` methods of transformers](https://scikit-learn.org/stable/auto_examples/miscellaneous/plot_set_output.html#sphx-glr-auto-examples-miscellaneous-plot-set-output-py)

![Image](https://github.com/user-attachments/assets/49937d42-cf82-4c67-85a7-36544b760c00)

Sadly, it looks like those aren't available yet (nilearn 0.10.4).

```python
import tempfile
from nilearn.maskers import NiftiMasker


with tempfile.NamedTemporaryFile(suffix=".nii.gz") as in_file:
    NiftiMasker(in_file).set_output('polars')

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[14], line 5
      2 from nilearn.maskers import NiftiMasker
      4 with tempfile.NamedTemporaryFile(suffix=".nii.gz") as in_file:
----> 5     NiftiMasker(in_file).set_output('polars')

File ~/git/a2cps/biomarkers/.venv/lib/python3.12/site-packages/sklearn/utils/_available_if.py:40, in _AvailableIfDescriptor.__get__(self, obj, owner)
     36 def __get__(self, obj, owner=None):
     37     if obj is not None:
     38         # delegate only on instances, not the classes.
     39         # this is to allow access to the docstrings.
---> 40         self._check(obj, owner=owner)
     41         out = MethodType(self.fn, obj)
     43     else:
     44         # This makes it possible to use the decorated method as an unbound method,
     45         # for instance when monkeypatching.

File ~/git/a2cps/biomarkers/.venv/lib/python3.12/site-packages/sklearn/utils/_available_if.py:34, in _AvailableIfDescriptor._check(self, obj, owner)
     31     raise AttributeError(attr_err_msg) from e
     33 if not check_result:
---> 34     raise AttributeError(attr_err_msg)

AttributeError: This 'NiftiMasker' has no attribute 'set_output'

```

### List any pages that would be impacted.

```shell
❯ rg -ul set_output doc 
doc/_build/html/modules/generated/nilearn.regions.Parcellations.html
doc/_build/html/modules/generated/nilearn.maskers.BaseMasker.html
[...]

PR #4902, opened 2024-12-06
[FIX] no set_output

Remi-Gau:
- Closes #

-
-

### DIFF
--- a/nilearn/connectome/connectivity_matrices.py
+++ b/nilearn/connectome/connectivity_matrices.py
@@ -370,7 +370,9 @@ def prec_to_partial(precision):
 
 
 @fill_doc
-class ConnectivityMeasure(BaseEstimator, TransformerMixin):
+class ConnectivityMeasure(
+    TransformerMixin, BaseEstimator, auto_wrap_output_keys=None
+):
     """A class that computes different kinds of \
        :term:`functional connectivity` matrices.
 
--- a/nilearn/decoding/searchlight.py
+++ b/nilearn/decoding/searchlight.py
@@ -237,7 +237,7 @@ def _group_iter_search_light(
 # Class for search_light #####################################################
 ##############################################################################
 @fill_doc
-class SearchLight(TransformerMixin, BaseEstimator):
+class SearchLight(TransformerMixin, BaseEstimator, auto_wrap_output_keys=None):
     """Implement search_light analysis using an arbitrary type of classifier.
 
     Parameters
--- a/nilearn/decomposition/_base.py
+++ b/nilearn/decomposition/_base.py
@@ -245,7 +245,9 @@ def _mask_and_reduce_single(
 
 
 @fill_doc
-class _BaseDecomposition(CacheMixin, TransformerMixin, BaseEstimator):
+class _BaseDecomposition(
+    CacheMixin, TransformerMixin, BaseEstimator, auto_wrap_output_keys=None
+):
     """Base class for matrix factorization based decomposition estimators.
 
     Handles mask logic, provides transform and inverse_transform methods
--- a/nilearn/glm/_base.py
+++ b/nilearn/glm/_base.py
@@ -4,7 +4,9 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from nilearn._utils import CacheMixin
 
 
-class BaseGLM(TransformerMixin, CacheMixin, BaseEstimator):
+class BaseGLM(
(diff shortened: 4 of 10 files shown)


Remi-Gau:
OK the set_output is still in the doc...

https://output.circle-artifacts.com/output/job/c326af4b-4eb6-43a7-a447-94269f498a8c/artifacts/0/dev/modules/generated/nilearn.maskers.BaseMasker.html#nilearn.maskers.BaseMasker.set_output

Remi-Gau:
@psadil 

- do you have any idea why the doc bug you reported in #4856 is still happening in this PR?

I tried to follow the sklearn doc to opt-out of the set_output and it seems to not be working.

If you have a bit of time, can you quickly look into this see if I am missing something?

If both of us fail I will open an issue on scikit learn to see if someone has had the same problem.

psadil:
> * do you have any idea why the doc bug you reported in [[DOC] method `set_output` mentioned but not yet implemented #4856](https://github.com/nilearn/nilearn/issues/4856) is still happening in this PR?

I'm not sure but can take a look

psadil:
This is an issue with classes in sklearn, too. For example, `LabelEncoder` uses `auto_wrap_output_keys=None`: https://github.com/scikit-learn/scikit-learn/blob/a23aef18586be9a3be763f4b89da0873288c6f55/sklearn/preprocessing/_label.py#L30

but the `set_output` method shows up in the docs

Screenshot 2024-12-06 at 16 50 00

An issue about that was raised, but it hasn't gone anywhere (maybe some restriction about what can be configured with Sphinx https://github.com/scikit-learn/scikit-learn/issues/26711#issuecomment-1800843281 )

doc/templates/class.rst
```diff
@@ -13,10 +13,9 @@
 
 .. autoclass:: {{ objname }}
    :inherited-members:
-
-   {% block methods %}
-   .. automethod:: __init__
-   {% endblock %}
+   :special-members: __init__
+   :exclude-members: set_output
+   :member-order: alphabetical
```
Remi-Gau, line 18: unrelated but it was bugging to see the methods not alphabetically sorted

Remi-Gau:
@psadil 

went nuclear and I am raising a not implemented error when set_ouput is called

also hiding the method from the doc for now

we should probably first discuss internally if its a method we want to support, but at least for now it won't show in the doc and if by chance the method is called (by users who may know about it in sklearn) the error will better match reality

Remi-Gau:
no set_output in doc anymore:

image

https://output.circle-artifacts.com/output/job/f24de9aa-9e80-41ca-884b-b6aec7bc46ba/artifacts/0/dev/search.html?q=set_output&check_keywords=yes&area=default

doc/templates/class.rst
man-shu: But the `:exclude-members:` is what removes it from the docs right?

man-shu (approved):
LGTM!

doc/templates/class.rst
Remi-Gau: yup

Remi-Gau: also no need for a special block to document init, it can just be done via special-members

psadil:
> went nuclear and I am raising a not implemented error when set_ouput is called
> 
> also hiding the method from the doc for now

Sounds good!

> we should probably first discuss internally if its a method we want to support, but at least for now it won't show in the doc and if by chance the method is called (by users who may know about it in sklearn) the error will better match reality

FWIW, the `set_output` api feels a bit like sugar. But as someone who often converts outputs to a `polars` dataframe (sometimes with an unpivot), I also think the api is nice for encouraging thinking about arrays in different ways. Happy to share thoughts if they'd be useful.

Remi-Gau:
> the api is nice for encouraging thinking about arrays in different ways.

I think it'd be nice to have if only for compatibility with vanilla sklearn, but I'd suggest to implement that in a follow up PR

doc/templates/class.rst
bthirion: This is something you want to add in this PR?

bthirion (commented):
LGTM overall.

doc/templates/class.rst
Remi-Gau: it's already done: the previous template made sure that the init of each class was documented, but this can just be done with the `:special-members: __init__` sphinx option

PR #4909, opened 2024-12-06
[REF] fix last-if-guard and lift-return-into-if

Remi-Gau:
- Closes none

apply following refactoring from sourcery

- lift-return-into-if
- last-if-guard

### DIFF
--- a/nilearn/_utils/numpy_conversions.py
+++ b/nilearn/_utils/numpy_conversions.py
@@ -74,6 +74,9 @@ def as_ndarray(arr, copy=False, dtype=None, order="K"):
         isinstance(arr, np.ndarray) and copy
     ):
         return np.array(arr, copy=True, dtype=dtype, order=order)
+    # if the order does not change and dtype does not change or
+    # bool to/from 1-byte dtype,
+    # no need to create a copy        
     elif (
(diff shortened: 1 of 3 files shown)


nilearn/_utils/numpy_conversions.py
```diff
@@ -73,10 +73,7 @@ def as_ndarray(arr, copy=False, dtype=None, order="K"):
     if isinstance(arr, (np.memmap, list, tuple)) or (
         isinstance(arr, np.ndarray) and copy
     ):
-        ret = np.array(arr, copy=True, dtype=dtype, order=order)
-    # if the order does not change and dtype does not change or
-    # bool to/from 1-byte dtype,
-    # no need to create a copy
+        return np.array(arr, copy=True, dtype=dtype, order=order)
```
Remi-Gau, line 76:
```suggestion
        return np.array(arr, copy=True, dtype=dtype, order=order)
    # if the order does not change and dtype does not change or
    # bool to/from 1-byte dtype,
    # no need to create a copy        
```

bthirion (commented):
One line needs testing. LGTM otherwise.

Remi-Gau:
> One line needs testing. LGTM otherwise.

not too worried about that uncovered line: a print statement that would run in case rich is not installed in the python environement

PR #4921, opened 2024-12-09
[ENH] Do sklearn checks on SurfaceLabelsMasker

man-shu:
- Closes None

- sklearn checks were happening on SurfaceMasker instead of SurfaceLabelsMasker

### DIFF
--- a/nilearn/maskers/tests/test_surface_labels_masker.py
+++ b/nilearn/maskers/tests/test_surface_labels_masker.py
@@ -6,7 +6,7 @@ from numpy.testing import assert_array_equal
 
 from nilearn._utils.class_inspect import check_estimator
 from nilearn._utils.helpers import is_matplotlib_installed
-from nilearn.maskers import SurfaceLabelsMasker, SurfaceMasker
+from nilearn.maskers import SurfaceLabelsMasker
 from nilearn.surface import SurfaceImage
 
 extra_valid_checks = [
@@ -26,7 +26,8 @@ extra_valid_checks = [
 @pytest.mark.parametrize(
     "estimator, check, name",
     check_estimator(
-        estimator=[SurfaceMasker()], extra_valid_checks=extra_valid_checks
+        estimator=[SurfaceLabelsMasker()],
+        extra_valid_checks=extra_valid_checks,
     ),
 )
 def test_check_estimator(estimator, check, name):  # noqa: ARG001
@@ -38,7 +39,7 @@ def test_check_estimator(estimator, check, name):  # noqa: ARG001
 @pytest.mark.parametrize(
     "estimator, check, name",
     check_estimator(
-        estimator=[SurfaceMasker()],
+        estimator=[SurfaceLabelsMasker()],
         valid=False,
         extra_valid_checks=extra_valid_checks,
     ),


nilearn/maskers/surface_labels_masker.py
```diff
@@ -95,7 +95,7 @@ class SurfaceLabelsMasker(TransformerMixin, CacheMixin, BaseEstimator):
 
     def __init__(
         self,
-        labels_img,
+        labels_img=None,
```
man-shu, line 98: So this seems to pass the `check_parameters_default_constructible` test, but I am not quite sure what complies the best with sklearn guidelines i.e.

whether to have `labels_img` as a parameter to `.fit()` or maintain our old behavior of letting users pass it during instantiation (but now it won't throw any error if they don't)

man-shu: Guidelines here: https://scikit-learn.org/stable/developers/develop.html#instantiation

Remi-Gau: If we do this we must thrown an error at fit if labels_img is None.

I think it'd be good to do this for all maskers: looking at NiftiMasker (which I believe is the oldest and most canonical) it allows None for the maask image.

nilearn/maskers/tests/test_surface_labels_masker.py
```diff
@@ -392,10 +393,6 @@ def test_surface_label_masker_inverse_transform_list_surf_images(
     assert img.shape == (surf_label_img.mesh.n_vertices, 7)
 
 
-def test_surface_label_masker_list_inverse_transform_output(surf_mesh):
```
man-shu, line 395: unrelated: seems like I started defining a test in a previous branch but then refactored other tests to test the same thing

nilearn/maskers/surface_labels_masker.py
bthirion: It's because the mask image can be computed from the data when using `NiftiMasker`.
To me it's OK to provide `labels_img` at instantiation time rather than fit time.

bthirion (commented):
LGTM.

nilearn/maskers/surface_labels_masker.py
```diff
@@ -168,6 +168,12 @@ def fit(self, img=None, y=None):
         """
         del img, y
 
+        if self.labels_img is None:
+            raise ValueError(
+                "Please provide a labels_img to the masker. For example, "
+                "masker = SurfaceLabelsMasker(labels_img=labels_img)"
+            )
```
Remi-Gau, line 175: Will probably need a test.

man-shu: Done!

nilearn/maskers/tests/test_surface_labels_masker.py
```diff
@@ -6,9 +6,22 @@
 
 from nilearn._utils.class_inspect import check_estimator
 from nilearn._utils.helpers import is_matplotlib_installed
-from nilearn.maskers import SurfaceLabelsMasker, SurfaceMasker
+from nilearn.maskers import SurfaceLabelsMasker
 from nilearn.surface import SurfaceImage
 
+
+@pytest.fixture
+def _sklearn_surf_label_img(surf_mesh):
```
man-shu, line 14: Had to create a new label img fixture locally because an invalid scikit-learn check was passing for unrelated reasons...

bthirion (approved):
LGTM, thx.

nilearn/reporting/data/html/report_body_template.html
```diff
@@ -24,11 +24,6 @@
       {{endif}}
     </div>
   </div>
-  <div class="pure-u-1 pure-u-md-1-3">
-    {{if description}}
-    <p>{{description}}</p>
-    {{endif}}
-  </div>
```
man-shu, line 31: If we don't define `description` during instantiation, we can't expect it here

man-shu: But it should nevertheless show up after fitting because the `report_body_template_surfacemasker.html` template has a field for it.

Actually let me confirm that

man-shu: Yes it does 👍🏽

man-shu:
All tests are passing, so now it's ok to review @Remi-Gau

Remi-Gau (approved):

Remi-Gau:
thanks @man-shu

man-shu:
Thanks for the reviews!

PR #4927, opened 2024-12-10
[MAINT] Fail fast for test matrix

Remi-Gau:
- fail early on tests runs
  - if one of the OS / python version fails then all the other ones are cancelled (whether queued or in progress)

- pros: speed things up by vacating resources for other CI jobs (which we currently need)
- cons: things may fail for only a specific OS or python version and this may hide it

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -43,7 +43,7 @@ jobs:
         name: 'Test with ${{ matrix.py }} on ${{ matrix.os }}: ${{ matrix.description }}'
         runs-on: ${{ matrix.os }}
         strategy:
-            fail-fast: false
+            fail-fast: true
             matrix:
                 description: [latest dependencies]
                 py: ['3.13', '3.12', '3.11', '3.10', '3.9']


Remi-Gau:
only a suggestion to speed things up in CI that has some drawback: we could revert it if we realize this is more annoying than waiting for CI to finish

Remi-Gau:
@man-shu @bthirion any opinions on this?

man-shu:
How about making it configurable with a commit message?

bthirion:
IMHO, this is indeed a reversible decision. Let's try.

bthirion (approved):

Remi-Gau:
> How about making it configurable with a commit message?

Let's try to have it on all the time and see if we feel like we need more flexibility

> I am completely for it btw: if I had a cent for thinking "wish I could cancel the rest of the checks", I'd be very rich...

Technically you can by going in `Actions` and selecting the workflow and then cancelling a specififc run: 

For example the test run.
https://github.com/nilearn/nilearn/actions/workflows/test_with_tox.yml

man-shu:
> Technically you can by going in `Actions` and selecting the workflow and then cancelling a specififc run:

Ooh ok good to know!

PR #4928, opened 2024-12-10
[REF] refactor fetch localizer contrasts

Remi-Gau:
- Closes none but relates to #3655 

- extract input checks


pyproject.toml
```diff
@@ -227,10 +227,10 @@ convention = "numpy"
 
 [tool.ruff.lint.pylint]
 # https://docs.astral.sh/ruff/settings/#lint_pylint_max-args
-max-args = 26
+max-args = 27
```
Remi-Gau, line 230:
```suggestion
max-args = 26
```

nilearn/datasets/func.py
```diff
@@ -902,6 +898,24 @@ def fetch_localizer_contrasts(
     return Bunch(ext_vars=csv_data, description=fdescr, **files)
 
 
+def _check_inputs_fetch_localizer_contrasts(contrasts):
```
bthirion, line 901: I like to have one-line docstrings on private functions.

bthirion (commented):
LGTM otherwise

nilearn/datasets/func.py
```diff
@@ -902,6 +898,24 @@ def fetch_localizer_contrasts(
...
+            "Contrasts should be a list of strings, but "
+            f'a single string was given: "{contrasts}"'
+        )
+    unknown_contrasts = [
+        x
+        for x in contrasts
+        if (x not in ALLOWED_CONTRASTS and x not in CONTRAST_NAME_WRAPPER)
+    ]
+    if unknown_contrasts:
+        raise ValueError(
```
man-shu, line 913: Maybe a couple of tests for the ValueErrors here? Or would that be out-of-scope for the PR?

nilearn/datasets/func.py
```diff
@@ -784,33 +781,32 @@ def fetch_localizer_contrasts(
...
+                "/localizer",
+                "derivatives",
+                "spm_1st_level",
+                f"sub-{subject_id}",
+                (
+                    f"sub-{subject_id}_task-localizer"
+                    f"_acq-{contrast}_{data_type}.nii.gz"
+                ),
+            ]
+        )
```
man-shu, line 803: I am wondering if this can also be refactored to a Path object?

man-shu (approved):
Otherwise LGTM!

nilearn/datasets/func.py
Remi-Gau: I think I tried this in the past and there was a good reason not to this, but I cannot remember.

In any case I would say this is beyond the scope of the PR.

Remi-Gau: but feel free to try it in another PR. 😜

man-shu: > I think I tried this in the past and there was a good reason not to this, but I cannot remember.

Looked at it again and it looks like we are trying to make a URL not a path so Path might give weird URL on windows machine

PR #4929, opened 2024-12-10
[ENH] rename parameter: `version` -> `atlas_type`

### LINKED ISSUE #2239: Improper parameter name & non-optimal args for atlas Pauli 2017
The `version` parameter for :func:`nilearn.datasets.fetch_atlas_pauli_2017` is a weird name. Version makes me think of numbers and iterations, and chronological versions of atlases, not the type of atlas. We should change it to `type`.

Additionally, currently, the valid arguments for this are 'prob' and 'det' which are not at all good. We should deprecate them with "probabilistic" & "deterministic".

I will open a PR for this.

hndgzkn:
- Closes #2239 

- Rename parameter: `version` as `atlas_type` in `nilearn.datasets.fetch_atlas_pauli_2017`
- Update documentation


nilearn/datasets/atlas.py
```diff
@@ -1961,17 +1961,18 @@ def fetch_atlas_talairach(level_name, data_dir=None, verbose=1):
     return Bunch(maps=atlas_img, labels=labels, description=description)
 
 
+@rename_parameters(replacement_params={"version": "atlas_type"})
```
hndgzkn, line 1964: I haven't set `end_version`. Shall I update it with a specific version?

hndgzkn:
> 'prob' and 'det' which are not at all good. We should deprecate them with "probabilistic" & "deterministic"

I haven't changed the possible values for `version`/`atlas_type`, shall I?

nilearn/datasets/atlas.py
```diff
@@ -13,7 +13,7 @@
 from nibabel import freesurfer, load
 from sklearn.utils import Bunch
 
-from .._utils import check_niimg, fill_doc, logger
+from .._utils import check_niimg, fill_doc, logger, rename_parameters
```
Remi-Gau, line 16:
```suggestion
from nilearn._utils import check_niimg, fill_doc, logger, rename_parameters
```

Remi-Gau: maybe this is just me but I find relative import confusing, so taking the opportunity to fix this one

nilearn/datasets/atlas.py
Remi-Gau: yeah we usually have a 2 releases deprecation cycle so it would be 0.13.1

Remi-Gau:
> I haven't changed the possible values for `version`/`atlas_type`, shall I?

yes

can be in this PR or another one if you prefer

hndgzkn:
> can be in this PR or another one if you prefer

Is there something special for deprecation, or just changing the values is ok?

Remi-Gau:
> Is there something special for deprecation, or just changing the values is ok?

we probably need to map old values to new values and throw a deprecation warning if old values are passed

Remi-Gau:
@hndgzkn 
quick FYI: as this PR implies a behavior change it won't be in the next 'patch release' (next week 🤞🏽) that should contain only bug fixes, refactor, doc changes.

nilearn/datasets/tests/test_atlas.py
```diff
@@ -763,6 +763,48 @@ def test_fetch_atlas_pauli_2017(tmp_path, request_mocker):
         atlas.fetch_atlas_pauli_2017("junk for testing", data_dir)
 
 
+def test_fetch_atlas_pauli_2017_deprecated_values(tmp_path, request_mocker):
+    # TODO: remove this test after release 0.13.0
```
bthirion, line 767: Please add a one-liner docstring to state what the test tests.

bthirion (commented):
LGTM otherwise

nilearn/datasets/tests/test_atlas.py
```diff
@@ -763,6 +763,52 @@ def test_fetch_atlas_pauli_2017(tmp_path, request_mocker):
         atlas.fetch_atlas_pauli_2017("junk for testing", data_dir)
 
 
+# TODO: remove this test after release 0.13.0
+def test_fetch_atlas_pauli_2017_deprecated_values(tmp_path, request_mocker):
```
Remi-Gau, line 767: Do you think you could refactor this test and the previous one into a single one parametrized over atlas_type ?

If it makes sense

hndgzkn: I think that would complicate the tests. The assertions are different for each atlas_type. 

The previous test function tests if the function behaves as expected. 

This test tests if the warnings are correct  and also checks the functionality.

Remi-Gau: ok fair enough: let's keep it like that then. 👍🏽

Remi-Gau (approved):

bthirion (approved):
LGTM, thx.

PR #4930, opened 2024-12-10
[ENH] Refactor `surf_mask` fixture to be 1D

man-shu:
- Closes None but related to https://github.com/nilearn/nilearn/pull/4830#discussion_r1877787373

- remove extra dimension in surface mask fixture


Remi-Gau:
should we still have a 2D mask fixture?

man-shu:
> should we still have a 2D mask fixture?

At least with volumes, I think we always expect 3D masks (no time points, meaning you can't pass different masks for different time points). For example, with NiftiMasker we check mask_img to be 3D:

https://github.com/nilearn/nilearn/blob/04be71faa15194e41e9455ca87d3b4427580ff42/nilearn/maskers/nifti_masker.py#L454

man-shu:
Yes we can have it but I can't think of a use case for that

Remi-Gau:
> I can't think of a use case for that

wait until one user provide one of those mask and you will have a use for it. 😜

bthirion (approved):

man-shu:
> wait until one user provide one of those mask and you will have a use for it. 😜

🤣 ok so you mean we should have the fixture to test for the eventuality when a user passes a SurfaceImage.
* I can see a use of it in #4830 at least.
Will make one..

Remi-Gau:
Do you think you could paramatrize some of the test to check that places that accept a 1D mask are also fine with a 2D one.

I think they should all pass, but this way we can prevent some regression.

Remi-Gau (changes requested):
Do you think you could paramatrize some of the test to check that places that accept a 1D mask are also fine with a 2D one.

I think they should all pass, but this way we can prevent some regression.

man-shu:
> Do you think you could paramatrize some of the test to check that places that accept a 1D mask are also fine with a 2D one.

Done!

PR #4939, opened 2024-12-12
[FIX] make HierarchicalKMeans.fit() takes X as array of shape [n_samples, n_features]

### LINKED ISSUE #4755: [BUG] `HierarchicalKMeans.fit()` takes transposed X
.... I noticed that `HierarchicalKMeans.fit()` takes transposed X:
https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/regions/hierarchical_kmeans_clustering.py#L233

See in `ReNA.fit()` for comparison:
https://github.com/nilearn/nilearn/blob/135e5388386c24ca879328040273127f8acb4a6a/nilearn/regions/rena_clustering.py#L518

Is this a bug or a feature?

_Originally posted by @man-shu in https://github.com/nilearn/nilearn/pull/4577#discussion_r1843840330_

Remi-Gau:
- Closes #4755


nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
```diff
@@ -151,16 +160,17 @@ def test_hierarchical_k_means_clustering_surface(
...
+    hkmeans = HierarchicalKMeans(n_clusters=n_clusters)
     # fit and transform the data
-    X_transformed = clustering.fit_transform(X)
+    X_transformed = hkmeans.fit_transform(X)
     # inverse transform the transformed data
-    X_inverse = clustering.inverse_transform(X_transformed)
+    X_inverse = hkmeans.inverse_transform(X_transformed)
 
     # make sure the n_features in transformed data were reduced to n_clusters
     assert X_transformed.shape[0] == n_clusters
```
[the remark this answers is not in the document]
man-shu: I think the transform should reduce the number of features because, in the end, we want to have one feature that represents many similar features.

See the docs:
Screenshot 2024-12-12 at 10 46 20 AM

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
```diff
@@ -112,9 +112,13 @@ def test_hierarchical_k_means():
 
 
 def test_hierarchical_k_means_clustering():
-    data_img, mask_img = generate_fake_fmri(shape=(10, 11, 12), length=15)
+    n_samples = 15
+    n_features = 150
```
man-shu, line 116: I don't understand why is `n_features` 150 here?

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
```diff
@@ -112,9 +112,13 @@ def test_hierarchical_k_means():
...
+    data_img, mask_img = generate_fake_fmri(
+        shape=(10, 11, 12), length=n_samples
+    )
     masker = NiftiMasker(mask_img=mask_img).fit()
-    X = masker.transform(data_img).T
```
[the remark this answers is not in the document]
man-shu: It is indeed worrying, but AFAICT it won't break because in line 135, we only test the first dimension to be equal to n_clusters, which it would always be:
```python
assert X_red.shape[0] == n_clusters
```
The problem would occur when we check the second dimension of `X_red`, and it won't be equal to n_samples we expected

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
Remi-Gau: yup me neither but this is the n_features of the data we get... 

# 🤷🏽 ❓

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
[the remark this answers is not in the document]
Remi-Gau: yeah but I added a check for this and it fails

https://github.com/nilearn/nilearn/actions/runs/12293225132/job/34305525450?pr=4939#step:6:27061

```
FAILED nilearn/regions/tests/test_hierarchical_kmeans_clustering.py::test_hierarchical_k_means_clustering - assert 7 == 8
```

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
```diff
@@ -123,13 +127,20 @@ def test_hierarchical_k_means_clustering():
     ):
         HierarchicalKMeans(n_clusters=-2).fit(X)
 
-    hkmeans = HierarchicalKMeans(n_clusters=8)
+    n_clusters = 8
+    hkmeans = HierarchicalKMeans(n_clusters=n_clusters)
     X_red = hkmeans.fit_transform(X)
+
+    # make sure the n_features in transformed data were reduced to n_clusters
+    assert X_red.shape[0] == n_clusters
```
Remi-Gau, line 135: this one fails here but not for the surface case

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
man-shu: So that's after the inverse_transform, right? I think the shape is all wonky because of the `.T` you just removed

Remi-Gau: not sure before the change I would get (15, 150) and after I just got (150,15)

or the other way around...

man-shu:
Hmmm... very weird indeed! I think we should first start with writing separate tests for each: fit, transform, and inverse_transform

Remi-Gau:
> Hmmm... very weird indeed! I think we should first start with writing separate tests for each: fit, transform, and inverse_transform

agreed

do you want to take care of it ore should I?

man-shu:
> do you want to take care of it ore should I?

On it!

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
man-shu: ok so 150 is the n_features after masking

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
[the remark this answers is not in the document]
Remi-Gau: why does it do that?

sorry (task switching means I can't follow this one)

man-shu:
Ok, so this is what I understood so far:

This method uses sklearn's [KMeans](https://scikit-learn.org/stable/modules/clustering.html#k-means) under the hood (actually, it uses MiniBatchKMeans but they are the essentially the same for us and the difference between them is beside the point).

Sklearn's KMeans is trying to find the clusters in **samples**, and we want to find clusters in **features**, so unless we transpose our array after masking, it will give us weird results. 

That is why the `.transform` expects us to provide X in the `(n_features, n_samples)` form so that later on, our features would be passed into MiniBatchKMeans as samples (remember all sklearn estimators expect X in `(n_samples, n_features)` form), and it would find the clusters in **our features**. So, the onus is on the user to provide the X in "correct" form.

I think we should make the API consistent and do the transpose under the hood and not expect the user to do so.

Please correct me if I am wrong @emdupre @bthirion .

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
```diff
@@ -16,30 +15,25 @@
 
 extra_valid_checks = [
     "check_clusterer_compute_labels_predict",
-    "check_clustering",
```
man-shu, line 19: Had to remove some of the sklearn checks. The valid checks are more similar to ReNA clustering now

man-shu:
@Remi-Gau pinging you for a review on this one

man-shu:
Oh so turns out almost all parcellation methods have this transpose issue because they are based on sklearn methods...

Remi-Gau:
> Oh so turns out almost all parcellation methods have this transpose issue because they are based on sklearn methods...

irony mode: on

great!!!!

irony mode: off

emdupre:
> Oh so turns out almost all parcellation methods have this transpose issue because they are based on sklearn methods...

Sorry what is the "issue" in this case, just to make sure I understand ? I thought the original issue was to streamline the user experience across parcellation methods, so I'm not sure how all parcellation methods could have that issue !

nilearn/regions/parcellations.py
```diff
@@ -40,26 +43,18 @@ def _estimator_fit(data, estimator, method=None):
...
-
+    estimator = clone(estimator)
+    if method in ["rena", "hierarchical_kmeans"]:
+        estimator.fit(data)
+    # transpose data for KMeans, AgglomerativeClustering because
+    # they cluster first dimension of data (samples) but we want to cluster
+    # features (voxels)
     else:
-        estimator = clone(estimator)
         estimator.fit(data.T)
```
man-shu, line 53: By "issue" I am just referring to the transpose of samples and features between nilearn vs. sklearn.

Since we have methods like kmeans and wards that use sklearn implementations we do this explicit transpose of the data for those methods here and no transpose for ReNA and now HierarchicalKMeans too

emdupre: I see, so this is still the same point about aligning the user interface between parcellation methods, it's just that in some parcellation cases we were handling this internally (i.e., imposing the idea that users would want to find clusters in voxels rather than in time points) while in ReNA and HierarchicalKMeans we were allowing the user to pass through directly to the sklearn API, which would have given unexpected results (assuming that they indeed want clusters in voxels, which almost all neuroimagers do with these methods) unless they transposed the data themselves.

Is that an accurate summary ? If so, I'd say that we're indeed just streamlining user experience 😸  In a very important way !

man-shu: Yes indeed! It isn't a problem with our pipeline or method

nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
```diff
@@ -6,40 +6,35 @@
...
     "check_fit1d",
-    "check_fit_score_takes_y",
     "check_no_attributes_set_in_init",
-    "check_pipeline_consistency",
-    "check_readonly_memmap_input",
     "check_transformers_unfitted",
     "check_transformer_n_iter",
+    "check_methods_subset_invariance",
+    "check_methods_sample_order_invariance",
+    "check_dict_unchanged",
```
man-shu, line 37: Hmm... This one passes with old dependencies but fails with new ones. Any pointers @Remi-Gau ? Otherwise I could just do the sklearn-version-check thing

man-shu: `check_dict_unchanged` I meant

Remi-Gau: Sklearn version check sounds appropriate then

man-shu:
Ok, this one is ready for a review @Remi-Gau

Remi-Gau:
probably worth adding a changelog entry, no?

man-shu:
> probably worth adding a changelog entry, no?

Oh yes, sorry, I always forget that

bthirion (approved):
LGTM

Remi-Gau (commented):
here I go approving my "own" pull request taken over by @man-shu

PR #4943, opened 2024-12-12
[STY] Apply a couple preview ruff rules

DimitriPapadopoulos:

- Apply a couple preview ruff rules that make real sense.

### DIFF
--- a/nilearn/glm/tests/test_contrasts.py
+++ b/nilearn/glm/tests/test_contrasts.py
@@ -20,12 +20,12 @@ from nilearn.glm.first_level import run_glm
     [
         (
             "face / 10 + (window - face) * 2 - house",
-            "a face xy_z house window".split(),
+            ["a", "face", "xy_z", "house", "window"],
             [0.0, -1.9, 0.0, -1.0, 2.0],
         ),
         (
             "xy_z",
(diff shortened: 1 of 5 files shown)


nilearn/glm/tests/test_contrasts.py
```diff
@@ -20,12 +20,12 @@
     [
         (
             "face / 10 + (window - face) * 2 - house",
-            "a face xy_z house window".split(),
+            ["a", "face", "xy_z", "house", "window"],
```
Remi-Gau, line 23: am kinda "meh" on this one but it only happens in single test so sure

Remi-Gau (approved):

Remi-Gau:
will also merge those one with a single review

thanks @DimitriPapadopoulos

nilearn/glm/tests/test_contrasts.py
DimitriPapadopoulos: We really want a list of strings here, why call a function to construct it instead of writing the list as a literal? Sure, the result might be less visually appealing, but that's on Python.

Remi-Gau: yeah I agree but it just looks weird: I will live with it. np

PR #4950, opened 2024-12-13
[MAINT] Refactor tags surface masker

Remi-Gau:
- Closes none

- create a _BaseSurfaceMasker from which all surface maskers inherit
- set the tags on the base masker

### DIFF
--- a/nilearn/maskers/base_masker.py
+++ b/nilearn/maskers/base_masker.py
@@ -396,3 +396,31 @@ class BaseMasker(TransformerMixin, CacheMixin, BaseEstimator):
                 "has not been fitted. "
                 "You must call fit() before calling transform()."
             )
+
+
+class _BaseSurfaceMasker(TransformerMixin, CacheMixin, BaseEstimator):
+    def _more_tags(self):
+        """Return estimator tags.
+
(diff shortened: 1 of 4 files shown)


Remi-Gau:
@man-shu does this refactoring make sense to you?

man-shu (approved):
Yes! Very clear indeed

nilearn/maskers/base_masker.py
```diff
@@ -396,3 +396,33 @@ def _check_fitted(self):
                 "has not been fitted. "
                 "You must call fit() before calling transform()."
             )
+
+
+class _BaseSurfaceMasker(TransformerMixin, CacheMixin, BaseEstimator):
+    """Class from which all surface maskers should inherit."""
```
man-shu, line 402: Maybe a "version added" tag? But I guess it doesn't really matter because it wont be public anyway

bthirion: +addition in Whatsnew

Remi-Gau: hmmm not sure given this is a private class
I'd wait before adverstising it

Issue #4896, opened 2024-12-05
new_img_like should be able to handle SurfaceImage


Remi-Gau:

Remi-Gau:
- closing in favor of #4879

Issue #4879, opened 2024-12-04
Add surface 'utility functions'


Remi-Gau:
Integrate the following surface functions into the nilearn.image if possible or make them public somewhere else otherwise

## mean image

https://github.com/nilearn/nilearn/blob/9492ea67cce7f1f62281adef4469f3e7f2b0d0bf/nilearn/maskers/_utils.py#L54

could go into: https://github.com/nilearn/nilearn/blob/9492ea67cce7f1f62281adef4469f3e7f2b0d0bf/nilearn/image/image.py#L529

## concatenate image

https://github.com/nilearn/nilearn/blob/9492ea67cce7f1f62281adef4469f3e7f2b0d0bf/nilearn/maskers/_utils.py#L93

could go into: https://github.com/nilearn/nilearn/blob/9492ea67cce7f1f62281adef4469f3e7f2b0d0bf/nilearn/image/image.py#L1473

## 2D_to_1D

https://github.com/nilearn/nilearn/blob/9492ea67cce7f1f62281adef4469f3e7f2b0d0bf/nilearn/maskers/_utils.py#L129

## new_img_like

to create new surface image from the mesh of an old one

## threshold_img

- [x] see #2747 and https://github.com/nilearn/nilearn/pull/4999

## load_mask_img

- [x] see https://github.com/nilearn/nilearn/pull/4999

bthirion:
One related question if we want to extend stuff in `nilearn.image` to surface data seamlessly.

Issue #4808, opened 2024-11-27
[ENH] `concat_imgs` should be able to concatenate surface images


pbarbarant:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe your proposed enhancement in detail.

As for now `concat_imgs ` expects only niimgs. Since it is a common operation, it would be nice to be able to concatenate surface signals expressed on similar meshes. The following code:
```python
concat_imgs([surf_img(4), surf_img(3)])
```
returns:
```python
Exception has occurred: TypeError
Data given cannot be loaded because it is not compatible with nibabel format:
<SurfaceImage (9, ...
  File "/Users/plbar/Code/nilearn/nilearn/_utils/niimg.py", line 126, in load_niimg
    raise TypeError(
  File "/Users/plbar/Code/nilearn/nilearn/_utils/niimg_conversions.py", line 315, in check_niimg
    niimg = load_niimg(niimg, dtype=dtype)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/plbar/Code/nilearn/nilearn/image/image.py", line 1549, in concat_imgs
    first_niimg = check_niimg(next(literator), ensure_ndim=ndim)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/plbar/Code/fmralign/fmralign/tests/test_preprocessing.py", line 197, in <module>
    concat_imgs([surf_img(4), surf_img(3)])
TypeError: Data given cannot be loaded because it is not compatible with nibabel format:
<SurfaceImage (9, ...
```

### Benefits to the change

Improve the seamless interface between nifti and surface handling.

### Pseudocode for the new behavior, if applicable

```python
# insert your code below

```

Remi-Gau:
yeah we have a private function for this at the moment: https://github.com/nilearn/nilearn/blob/8318065a433186d0553db281fcf634f21c46ca48/nilearn/maskers/_utils.py#L94

I think after the coming release we will do some refactoring and better integrate some of those utility function with the rest of the code base.

pbarbarant:
> yeah we have a private function for this at the moment:
> 
> [nilearn/nilearn/maskers/_utils.py](https://github.com/nilearn/nilearn/blob/8318065a433186d0553db281fcf634f21c46ca48/nilearn/maskers/_utils.py#L94)
> 
[...]

PR #4956, opened 2024-12-16
[REF] move surface utility functions

Remi-Gau:
- closes none but relates to:
  - #4808 
  - #4879 
  - #4896

- move surface function from nilearn.maskers._utils to nilearn.surface.surface
- keep them private
- rename some of them so their name match that of their nilearn.image volume counterpart
- add a new_img_like private function


Remi-Gau:
Also relates to https://github.com/nilearn/nilearn/issues/4808

@pbarbarant this PR will moves the private function for concatenating surface image 

```
from nilearn.surface.surface import concat_imgs
```

Remi-Gau:
linkcheck failure is unrelated: osf links time out once in a while

nilearn/surface/surface.py
```diff
@@ -1753,3 +1779,167 @@ def from_volume(
...
+
+    return {
+        hemi: data.parts[hemi][..., index]
+        .copy()
+        .reshape(mesh.parts[hemi].n_vertices, 1)
+        for hemi in data.parts
+    }
+
+
+def new_img_like(ref_niimg, data):
```
man-shu, line 1939: maybe rename this variable?
```suggestion
def new_img_like(ref_img, data):
```

nilearn/surface/surface.py
```diff
@@ -1753,3 +1779,167 @@ def from_volume(
...
+        data = img.data
+    else:
+        data = {
+            part: np.mean(value, axis=1).astype(float)
+            for part, value in img.data.parts.items()
+        }
+    return new_img_like(img, data=data)
+
+
+def two_dim_to_one_dim(img):
```
man-shu, line 1904: I actually liked the name before, it was something like `deconcatenate...` IIRC?
But as you mentioned in the dev meeting, maybe keeping close to the Nifti image functions is better. I found that we have `iter_img` function that does a similar thing with nifti images.
https://github.com/nilearn/nilearn/blob/04be71faa15194e41e9455ca87d3b4427580ff42/nilearn/image/image.py#L726
Screenshot 2024-12-17 at 12 04 48 PM

nilearn/surface/surface.py
```diff
@@ -1753,3 +1779,167 @@ def from_volume(
...
+    vmin : float
+
+    vmax : float
+    """
+    vmin = min(min(x.ravel()) for x in img.data.parts.values())
+    vmax = max(max(x.ravel()) for x in img.data.parts.values())
+    return vmin, vmax
+
+
+def concat_extract_surface_data_parts(img):
```
man-shu, line 1831: In the same vein, how about renaming this to `get_data`? One caveat might be that we are doing concatenation across hemispheres but that could be mentioned in the docstring.
https://github.com/nilearn/nilearn/blob/04be71faa15194e41e9455ca87d3b4427580ff42/nilearn/image/image.py#L46
Screenshot 2024-12-17 at 12 07 46 PM

nilearn/surface/surface.py
```diff
@@ -1753,3 +1779,167 @@ def from_volume(
...
+
+    Returns
+    -------
+    :obj:`~numpy.ndarray`
+        Concatenated data across hemispheres.
+    """
+    return np.concatenate(list(img.data.parts.values()), axis=0)
+
+
+def concat_imgs(imgs):
```
bthirion, line 1848: you deliberately want to have names more similar btw the surface module and the image module I guess. I find it reasonable, but the names of the different routines are less homogeneous, some with _surface_, some without.

nilearn/surface/surface.py
```diff
@@ -1753,3 +1779,167 @@ def from_volume(
...
+            f"but found {n_dim_left}D for left and {n_dim_right}D for right."
+        )
+
+
+def get_min_max_surface_image(img):
+    """Get min and max across hemisphere for a SurfaceImage.
+
+    Parameters
+    ----------
+    img : SurfaceImage
```
bthirion, line 1818:
```suggestion
    img : :obj:`~nilearn.surface.SurfaceImage`
```
(and ame thing for other functions)

man-shu (commented):
Otherwise looks good

nilearn/surface/surface.py
Remi-Gau: Could try but I was trying to align with nibabel's four_to_three but I felt that `two_to_one` was a bit strange

https://nipy.org/nibabel/reference/nibabel.funcs.html#nibabel.funcs.four_to_three

If we go for iter_img, then we probably want to implement a generator by default, no?
This is what the name implies at least.

nilearn/surface/surface.py
Remi-Gau: sounds fair indeed

nilearn/surface/surface.py
Remi-Gau: > but the names of the different routines are less homogeneous

can you point to an example? I am unsure what you mean by routines.

nilearn/surface/surface.py
man-shu: I checked and this is what it is returns in the end ([code](
https://github.com/nilearn/nilearn/blob/d642bb8a63e19620729fbb52d72dffb933fa707a/nilearn/_utils/niimg_conversions.py#L332))
Screenshot 2024-12-17 at 1 25 44 PM

the `_index_img` function just returns `new_img_like`.

So it seems like it returns a very long tuple? But either way, I think tthe end use case is the same here, no?

man-shu:
Also, I just remembered: didn't we wanna keep them somewhere private? 
Because later, we might decide to refactor the original `image.*` functions to accept surface images, and if they are public now, they'd have to be deprecated...

nilearn/surface/surface.py
Remi-Gau: na pretty sure this is an "iterator" not a tuple: it returns a generator:

```python
In [1]: foo = (a for a in [1, 2])
In [2]: type(foo)
Out[2]: generator
```

nilearn/surface/surface.py
bthirion: Sorry, I mean the names of the functions in this module, that seem to be mostly ancillary, hence the name "routines": for instance, we have `mean_img` but `_extract_surface_image_data`, meaning that surface data are called surface_image or image, depending on the function.

nilearn/surface/surface.py
```diff
@@ -1753,3 +1780,191 @@ def from_volume(
...
+    -------
+    a SurfaceImage object
+    """
+    if not isinstance(img, SurfaceImage):
+        raise TypeError("Input must a be SurfaceImage.")
+    img = at_least_2d(img)
+    return new_img_like(img, data=_extract_data(img, index))
+
+
+def _extract_data(img, index):
```
bthirion, line 1946: Can you add a short docstring ?

nilearn/surface/tests/test_surface.py
```diff
@@ -1087,3 +1096,91 @@ def test_inmemorymesh_index_error(in_memory_mesh):
...
+    img.data.parts["right"][:, 0] = np.zeros(shape=(5))
+    img.data.parts["right"][0][0] = -3.5
+
+    vmin, vmax = get_min_max(img)
+
+    assert vmin == -3.5
+    assert vmax == 10
+
+
+def test_concat_imgs(surf_img_2d):
```
bthirion, line 1135: I like short docstrings

nilearn/surface/tests/test_surface.py
```diff
@@ -1087,3 +1096,91 @@ def test_inmemorymesh_index_error(in_memory_mesh):
...
+    output = iter_img(input, return_iterator=False)
+
+    assert_surface_image_equal(output[0], input)
+
+    output = iter_img(surf_img_1d, return_iterator=False)
+
+    assert_surface_image_equal(output[0], surf_img_1d)
+
+
+def test_iter_img_wrong_input():
```
bthirion, line 1169: docstring please.

nilearn/surface/surface.py
bthirion: Any opinion on this ?

bthirion (commented):
Looks good overall

nilearn/surface/tests/test_surface.py
Remi-Gau: We should probably use ruff to start flagging those so you don't have to keep on reminding us.

At the moment the error for this D103 (I think) to make sure public functions have a doc string is ignored in the tests folders.

Remi-Gau: If we do not ignored it, it flags around 900 functions with missing doc strings. This could be done over several PRs.

Am just a bit worried of adding meaningless doc strings.

```
def test_check_inputs_foo():
   """Test the check_inputs_foo function."""
```

nilearn/surface/surface.py
Remi-Gau: na at the moment I would keep the name similar to those in nilearn.image when the function does the same thing.

In general I would also just call avoid using surface_image for variable names or in function since we are in the surface module.

nilearn/surface/surface.py
Remi-Gau, line 1946: @man-shu 
will make this a private method in SurfaceImage

nilearn/surface/surface.py
```diff
@@ -1753,3 +1780,191 @@ def from_volume(
         data = PolyData(left=texture_left, right=texture_right)
 
         return cls(mesh=mesh, data=data)
+
+
+def check_surface_data_ndims(img, dim, var_name="img"):
```
Remi-Gau, line 1785: will make this a private method in polydata

nilearn/surface/surface.py
```diff
@@ -1753,3 +1780,191 @@ def from_volume(
...
+    n_dim_right = img.data.parts["right"].ndim
+    if not all(x == dim for x in [n_dim_left, n_dim_right]):
+        raise ValueError(
+            f"Data for each hemisphere of {var_name} should be {dim}D, "
+            f"but found {n_dim_left}D for left and {n_dim_right}D for right."
+        )
+
+
+def get_min_max(img):
+    """Get min and max across hemisphere for a SurfaceImage.
```
Remi-Gau, line 1815: should be private method of polydata

nilearn/surface/surface.py
```diff
@@ -1753,3 +1780,191 @@ def from_volume(
...
+    vmin : float
+
+    vmax : float
+    """
+    vmin = min(min(x.ravel()) for x in img.data.parts.values())
+    vmax = max(max(x.ravel()) for x in img.data.parts.values())
+    return vmin, vmax
+
+
+def get_data(img):
```
Remi-Gau, line 1832: should also work with polydata

nilearn/surface/tests/test_surface.py
bthirion: I see, in some cases the test name is transparent, but I find it helpful in general.

bthirion (approved):
LGTM, thx.

Issue #4628, opened 2024-10-11
Speed up test suite


Remi-Gau:
Our test suite is getting slower.

See stats over the past year (same thing for windows, macos has a discontunuity most likely due to an infrastructure change on the github side)

![Image](https://github.com/user-attachments/assets/55d69b1f-a1a1-4112-89bd-074000d82a7b)

I am putting below all the 'slow tests' (> 2 sec) from a [recent run](https://github.com/nilearn/nilearn/actions/runs/11288313394/job/31395844370) to see if we can optimize some tests.

<details>
<summary>slow tests</summary>
<pre>
2024-10-11T08:30:02.3200983Z 35.40s call     nilearn/decoding/tests/test_decoder.py::test_decoder_regression_clustering[svr-99]
2024-10-11T08:30:02.3202555Z 31.38s call     nilearn/decoding/tests/test_decoder.py::test_decoder_regression_clustering[ridge-99]
2024-10-11T08:30:02.3204105Z 22.32s call     nilearn/glm/tests/test_second_level.py::test_non_parametric_inference_cluster_level_with_covariates
2024-10-11T08:30:02.3205650Z 18.95s call     nilearn/reporting/tests/test_html_report.py::test_report_empty_fit[NiftiMapsMasker]
2024-10-11T08:30:02.3207141Z 18.71s call     nilearn/reporting/tests/test_html_report.py::test_report_empty_fit[MultiNiftiMapsMasker]
2024-10-11T08:30:02.3208826Z 18.10s call     nilearn/reporting/tests/test_html_report.py::test_nifti_maps_masker_report_integer_and_all_displayed_maps[all]
2024-10-11T08:30:02.3210645Z 17.94s call     nilearn/reporting/tests/test_html_report.py::test_nifti_maps_masker_report_integer_and_all_displayed_maps[9]
2024-10-11T08:30:02.3212383Z 17.81s call     nilearn/reporting/tests/test_html_report.py::test_warning_in_report_after_empty_fit[NiftiMapsMasker]
2024-10-11T08:30:02.3214136Z 17.61s call     nilearn/reporting/tests/test_html_report.py::test_nifti_maps_masker_report_integer_and_all_displayed_maps[12]
2024-10-11T08:30:02.3215528Z 17.55s call     nilearn/glm/tests/test_first_level.py::test_fmri_inputs
[...]

PR #4958, opened 2024-12-17
[MAINT] use UV to install in CI

Remi-Gau:
- Closes none but relates to #4628 #4663

- use uv to run "pip install" in CI: https://docs.astral.sh/uv/
- allows to shave of precious time in CI
- unrelated: simplify the names of the tox "environments"

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -81,12 +81,14 @@ jobs:
             uses: actions/checkout@v4
             with:
                 ref: ${{ github.event.pull_request.head.sha }}
+        -   name: Install the latest version of uv
+            uses: astral-sh/setup-uv@v4
         -   name: Setup python
             uses: actions/setup-python@v5
             with:
                 python-version: '3.13'
         -   name: Install packages
-            run: python -m pip install tox
+            run: uv tool install tox --with=tox-uv --with=tox-gh-actions
         -   name: check links
             run: tox run -e linkcheck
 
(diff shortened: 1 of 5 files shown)


bthirion:
What is the expected benefit of using uv ?

Remi-Gau:
> What is the expected benefit of using uv ?

SPEEED... and saving a lot of time in CI:

try it locally

## with pip

```bash
$ time pip install --no-cache-dir -e .[dev]

...

41,50s user 5,17s system 65% cpu 1:10,94 total
```

## with uv

```bash
$ pip install uv
$ uv tool install --no-cache-dir -e '.[dev]'

...
7,80s user 5,53s system 62% cpu 21,200 total
```

It's a >30 seconds difference but given we install nilearn for every PR:
- 3 times per doc build
- 18 per test run

this starts to add up.

FYI this does not constrain in terms of build and will not make a difference for users.

man-shu (approved):
I'm all for speeeeeed. It's just because of Rust right?

Remi-Gau:
> It's just because of Rust right?

yup

bthirion:
Why doesn't the doc build ?

bthirion (approved):

Remi-Gau:
> Why doesn't the doc build ?

I think that the PR deploy on circle CI will only happen if the whole Doc workflow passes (including link checks).

Given how brittle this latter one is I am more and more tempted to say that linkcheks should be run less often (just on main?).

bthirion:
But we need to check them *because* they're brittle...

Remi-Gau:
hmm some failures in main seem to happen on the doc build and nightly workflow after merging this one so I will revert this one and rework it

PR #4964, opened 2024-12-18
[MAINT] use UV to install in CI

Remi-Gau:
- Closes none but relates to #4628 #4663

- use uv to run "pip install" in CI: https://docs.astral.sh/uv/
- allows to shave of precious time in CI
- unrelated: simplify the names of the tox "environments"


Remi-Gau:
This is the same as https://github.com/nilearn/nilearn/pull/4958 (minus the bugs) that was approved so I assume this is still good to merge.

PR #4967, opened 2024-12-18
[FIX] allow support for confounds as numpy arrays in FirstLevelModel

### LINKED ISSUE #4893: GLM confounds as numpy array are not supported
ok something does not make sense about allowing numpy array: they are accepted as inputs but never appended to the output list... so they will never be passed on to whatever needs them later...

I think that in a follow up PR I will check if we actually support array as confounds and it is properly implemented or if this is a doc string error.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/4807#discussion_r1869224785_

Remi-Gau:
- Closes #4893

- add test to check that numpy array confounds are used in model
- adapt code to make test pass

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -639,19 +639,38 @@ class FirstLevelModel(BaseGLM):
         return design_matrices
 
     def _create_single_design(self, n_scans, events, confounds, run_idx):
-        """Build experimental design of a single run."""
+        """Build experimental design of a single run.
+
+        Parameters
+        ----------
+        n_scans: int
+
+        events : list of pandas.DataFrame
+
+        confounds : list of pandas.DataFrame or numpy.arrays
+
+        run_idx : int
+        """
+        confounds_matrix = None
+        confounds_names = None
         if confounds is not None:
-            confounds_matrix = confounds[run_idx].to_numpy()
+            confounds_matrix = confounds[run_idx]
+
+            if isinstance(confounds_matrix, pd.DataFrame):
+                confounds_names = confounds[run_idx].columns.tolist()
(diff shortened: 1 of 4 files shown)


nilearn/glm/first_level/first_level.py
```diff
@@ -639,19 +639,38 @@ def _create_all_designs(
...
+        confounds : list of pandas.DataFrame or numpy.arrays
+
+        run_idx : int
+        """
+        confounds_matrix = None
+        confounds_names = None
         if confounds is not None:
-            confounds_matrix = confounds[run_idx].to_numpy()
+            confounds_matrix = confounds[run_idx]
+
```
hndgzkn, line 658: It might be good to verify that type of confounds is list, raise error otherwise.

nilearn/glm/first_level/first_level.py
```diff
@@ -639,19 +639,38 @@ def _create_all_designs(
...
+            if isinstance(confounds_matrix, pd.DataFrame):
+                confounds_names = confounds[run_idx].columns.tolist()
+                confounds_matrix = confounds_matrix.to_numpy()
+            else:
+                # create dummy names when dealing with numpy arrays
```
hndgzkn, line 663: Also here it can be verified that type of list element is numpy array, and raise error otherwise.

hndgzkn (approved):
Minor comments, otherwise ok for me.

nilearn/glm/first_level/first_level.py
Remi-Gau: Pretty sure that by then inputs have been sanitized so we should not get other types.

Remi-Gau:
@bthirion

man-shu (approved):
LGTM!

Issue #4886, opened 2024-12-04
[DOC] [BUG] No plots in reports for surface maskers on Chromium browsers


man-shu:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe your proposed suggestion in detail.

The plots in the reports of SurfaceMasker and SurfaceLabelsMasker are not visible on Chromium browsers:

![Image](https://github.com/user-attachments/assets/5e554420-e495-4525-a943-324020722414)

They do show up on Firefox, apparently.

man-shu:
On Firefox mobile:
![Screenshot_20241204-182126.png](https://github.com/user-attachments/assets/9108696f-e445-4827-9d64-446a4b9d9ae9)

Remi-Gau:
My hunch or something to investigate: how the figure are converted before embedding.

# Surface masker

this is done during the building of the report in the surface masker code

https://github.com/nilearn/nilearn/blob/d1eb079e61bda736ffdcefe0978ad494bfb57ec9/nilearn/maskers/surface_masker.py#L440

# Nifti masker

The masker code creates a list of figures: 

https://github.com/nilearn/nilearn/blob/d1eb079e61bda736ffdcefe0978ad494bfb57ec9/nilearn/maskers/nifti_masker.py#L403

and the conversion is done by the reporting code:
https://github.com/nilearn/nilearn/blob/d1eb079e61bda736ffdcefe0978ad494bfb57ec9/nilearn/reporting/html_report.py#L292

At the moment it is VERY unclear what types are expected by the "embedding" in reporting.

I would suggest trying to have better tests at the reporting level to try to reproduce this bug outside of the masker context.

man-shu:
> At the moment it is VERY unclear what types are expected by the "embedding" in reporting.

Indeed! It's also a bit different when you look at NiftiMapsMasker:
https://github.com/nilearn/nilearn/blob/d1eb079e61bda736ffdcefe0978ad494bfb57ec9/nilearn/maskers/nifti_maps_masker.py#L348

The `_embed_img` function uses `figure_to_png_base64` internally but extracts the figure from the display:
https://github.com/nilearn/nilearn/blob/d1eb079e61bda736ffdcefe0978ad494bfb57ec9/nilearn/reporting/html_report.py#L71

Remi-Gau:
go for it

Issue #4664, opened 2024-10-17
[ENH] nilearn.plotting.view_surf for both hemispheres?


JohannesWiesner:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe your proposed enhancement in detail.

Is there a plan to enhance `nilearn.plotting.view_surf` to show both hemispheres simultaneously?

### Benefits to the change

Instead of having two tabs open in your browser (where you have to switch between tabs) you could see the whole brain in one plot.

Remi-Gau:
I think there we are constrained only by the fact that the surface meshes are only of one hemipshere, no?

bthirion:
It should definitely be doable.

man-shu:
While we work on this, a quick fix could be something like this:

```python
import numpy as np
from nilearn.experimental.surface import InMemoryMesh, PolyMesh, SurfaceImage

# create fake mesh
# for the tetrahedron
left_coords = np.asarray(
    [
        [0.0, 0, 0],  # vertex 0
        [1, 0, 0],  # vertex 1
        [0, 1, 0],  # vertex 2
        [0, 0, 1],  # vertex 3
    ]
)
left_faces = np.asarray(
    [
        [1, 0, 2],  # face created by connecting vertices 1, 0, 2
        [0, 1, 3],  # face created by connecting vertices 0, 1, 3
        [0, 3, 2],  # face created by connecting vertices 0, 3, 2
        [1, 2, 3],  # face created by connecting vertices 1, 2, 3
    ]
)
# for the pyramid
right_coords = (
    np.asarray([[0.0, 0, 0], [1, 0, 0], [1, 1, 0], [0, 1, 0], [0, 0, 1]]) + 0.5
)
right_faces = np.asarray(
    [
        [0, 1, 4],
        [0, 3, 1],
        [1, 3, 2],
        [1, 2, 4],
        [2, 3, 4],
        [0, 4, 3],
    ]
)
# put the two parts together
mesh = PolyMesh(
    left=InMemoryMesh(left_coords, left_faces),
    right=InMemoryMesh(right_coords, right_faces),
)

# create fake data
rng = np.random.default_rng(0)
left_data = rng.random(mesh.parts["left"].n_vertices)
right_data = rng.random(mesh.parts["right"].n_vertices)
# put them together in a dictionary
data = {"left": left_data, "right": right_data}

# combine to make a SurfaceImage
surface_image = SurfaceImage(mesh=mesh, data=data)

[...]

PR #4968, opened 2024-12-18
[ENH] Add reports for SurfaceMapsMasker

man-shu:
- Closes None but relates to #4830


man-shu:
Preview of what the report looks like on 1080x1920 res
Screenshot 2024-12-20 at 3 21 19 PM
You need to scroll down to see the other side of the hemispheres, even though there is space at the bottom. There seems to be a "box" that is smaller than the whole page.

I will either to expand that box or decrease the vertical space between the subplots

man-shu:
Any specific reason why we don't use plotly for the plots in reports? I tried it and it looks better IMO:
Screenshot 2024-12-23 at 3 57 36 PM

- However would be much better to first resolve #4664

man-shu:
* And doing this would (partly) resolve #4886

bthirion:
I have no issue with using plotly in reports.

Remi-Gau:
> I have no issue with using plotly in reports.

good with me as well though it means report can only be generated when plotly is installed where as for other maskers they only require matplotlib

where you thinking of embedding static figures or interactive HTML iframes?

man-shu:
> good with me as well though it means report can only be generated when plotly is installed where as for other maskers they only require matplotlib

How I've implemented it right now, it would fall back to matplotlib if no plotly.

> where you thinking of embedding static figures or interactive HTML iframes?

It's interactive, see: https://drive.google.com/file/d/1QLEFEKnMhYjPR2YbXKMQvSLMVQm1LfZS/view?usp=sharing

I guess one downside might be that the report would be huge -- this one is 14 MB. It's because we are embedding Plotly's CSS into the HTML as of now. They have an option where we can just provide Plotly's CDN link, and it would fetch the CSS from there every time you open it. But that would mean the report won't show the figure when you're offline...

Remi-Gau:
> I guess one downside might be that the report would be huge

that was my other concern: I'd prefer to have smaller report by default

man-shu:
Ok, now I am using `view_surf` instead of `plot_surf` to get the plotly figure's HTML, and that seems to be lighter.

So, if we plot using the following example, these are the options we have:
1. Matplotlib's SVG: static, **57MB**, [report_mpl_svg.html](https://drive.google.com/file/d/17VEOpJYB4hK0xvg6whK3HLRnZhsmmltW/view?usp=sharing) : Right now, we are using this in SurfaceMasker and SurfaceLabelsMasker. On this PR, with SurfaceMapsMasker, this would be used if no plotly.
2. Plotly: interactive, 9.1 MB, [report_plotly.html](https://drive.google.com/file/d/11HHIuO6Xb36CiF9WrHrk_BcyzkGGAz1r/view?usp=sharing) : Proposed default on this PR.
3. Matplotlib's PNG: static, 1.7MB, [report_mpl_png.html](https://drive.google.com/file/d/1XD-YzDZVsuqX_y1lw5xkVd-isOwQUv8s/view?usp=sharing): Does not work on Chrome browsers as of now, but easily fixable.


Code to generate above reports:
```python
# %%
from nilearn.datasets import (
    fetch_atlas_difumo,
    fetch_atlas_msdl,
    load_nki,
    load_fsaverage,
)
from nilearn.maskers import SurfaceMapsMasker
from nilearn.maskers._utils import (
    compute_mean_surface_image,
)
from nilearn.plotting import plot_surf, show, view_surf
import numpy as np
from pathlib import Path

# Fetch the a volumetric probabilistic atlas
atlas = fetch_atlas_msdl()
# or
# atlas = fetch_atlas_difumo()
# Fetch the fsaverage5 mesh
fsaverage5_mesh = load_fsaverage("fsaverage5")["pial"]
# Fetch the NKI dataset
surf_img = load_nki()[0]
mean_surf_img = compute_mean_surface_image(surf_img)
# project atlas to the surface
surf_atlas = SurfaceImage.from_volume(
    volume_img=atlas.maps, mesh=fsaverage5_mesh
)
# Create a masker object
masker = SurfaceMapsMasker(surf_atlas).fit()
# Extract the time series
time_series = masker.transform(surf_img)
# see masker report
report = masker.generate_report(3)
report.save_as_html(Path("report.html"))
```

bthirion:
So, we use plotly if available, otherwise matplotlib if available. Sounds reasonable to me.

man-shu:
So the current failure seems to happen because we require matplotlib to be installed in order to use reporting, via:
https://github.com/nilearn/nilearn/blob/655bb5fb4b224b44c4c6013f57e5c2a074fdf38e/nilearn/reporting/__init__.py#L10

The way I see it, we would prefer Matplotlib for Nifti maskers and Plotly for Surface maskers (only surface maps masker in this PR, but this behavior could be replicated for the others). 

So I suppose instead of requiring only matplotlib at the init of reporting module, it would be better to set the preference at individual masker-level? I will open a separate PR for this.

But LMWYT @Remi-Gau @bthirion

bthirion:
Not sure I follow you: For sufaceMasker, we prefer plotly, but using mpl remains OK ? So what is wrong about requiring only mpl at the module level ?
(my thought is: try to keep it simple...)

man-shu:
> Not sure I follow you: For sufaceMasker, we prefer plotly, but using mpl remains OK ? So what is wrong about requiring only mpl at the module level ? (my thought is: try to keep it simple...)

I wanted to make plotly the 1st priority for surface maskers. This means surface masker reports should work as long as plotly is there (even if matplotlib is not). But as we have it right now, we can't even import the reporting module unless matplotlib is installed.

Remi-Gau:
> This means surface masker reports should work as long as plotly is there (even if matplotlib is not).

I am pretty sure that matplotlib is dependency of plotly, so if you have plotly, then matplotlib will be there.

Remi-Gau:
> So, we use plotly if available, otherwise matplotlib if available. Sounds reasonable to me.

I would actually stick to matplotlib being the default and only have plotly as optional given the size of of the generated report.

man-shu:
> given the size of of the generated report.

I actually found that the report would be heavier if we insert a matplotlib figure as SVG in the report. See https://github.com/nilearn/nilearn/pull/4968#issuecomment-2561122727

> I am pretty sure that matplotlib is dependency of plotly, so if you have plotly, then matplotlib will be there.

OK then let's just stick to matplotlib as 1st priority

nilearn/reporting/utils.py
```diff
@@ -38,6 +38,11 @@ def figure_to_svg_base64(fig):
     return base64.b64encode(figure_to_svg_bytes(fig)).decode()
 
 
+def figure_to_png_base64(fig):
+    """Save figure as png and return it as 64 bytes."""
+    return base64.b64encode(figure_to_png_bytes(fig)).decode()
```
man-shu, line 43: This function was removed in #5038. Seems like we never used it before, but we need it to create png images in the reports.

man-shu:
* Ok looks like now all the fails are related to #5044

bthirion (commented):
Almost there. Where can I see an example of generated report ?

nilearn/maskers/surface_maps_masker.py
```diff
@@ -429,3 +455,231 @@ def inverse_transform(self, region_signals):
...
+        maps_img = self._reporting_data["maps_img"]
+
+        img = self._reporting_data["images"]
+        if img:
+            img = mean_img(img)
+
+        # Handle the edge case where this function is
+        # called with a masker having report capabilities disabled
+        if self._reporting_data is None:
+            return [None]
```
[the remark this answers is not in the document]
man-shu: `test_reports_false` was supposed to cover this but somehow it isn't. Let me have another look

man-shu:
The plotly engine doesn't work anymore (issue with `view_surf`). But it was when I started working on this -- all the screenshots and HTMLs above are the proof. Something seems to have changed since then (between 0.11.0 and 0.11.1) it could either be on our end or maybe plotly

* This is related to #5018 specifically [this comment](https://github.com/nilearn/nilearn/issues/5018#issuecomment-2592402254)

nilearn/plotting/html_surface.py
```diff
@@ -118,8 +118,8 @@ def _one_mesh_info(
         vmax=vmax,
         vmin=vmin,
     )
-    info = {"inflated_left": mesh_to_plotly(surf_mesh)}
-    info["vertexcolor_left"] = get_vertexcolor(
+    info = {"inflated_both": mesh_to_plotly(surf_mesh)}
+    info["vertexcolor_both"] = get_vertexcolor(
```
man-shu, line 122: > The plotly engine doesn't work anymore (issue with `view_surf`).

This fixes `view_surf`. Seems like I introduced it in #4991. But that shouldn't have affected the stable docs. Weird...

man-shu:
> Almost there. Where can I see an example of generated report ?

Ran a full doc build let's wait

man-shu:
The docs are here: https://github.com/nilearn/nilearn/actions/runs/12793251992/artifacts/2436872603

Seems like there are a couple of issues:
1. The plotly figure is a bit too big when displayed on a smaller screen; maybe I can make the size of the figure responsive.
2. The buttons to see the next maps don't work properly. Not just for surface maps masker but all the maps masker (nifti and multi nifti).

On a positive note, the issue with `view_surf` has been fixed (related to https://github.com/nilearn/nilearn/issues/5018#issuecomment-2592402254), even on the `plot_surface_101.py` example (see the screenshot below). This means the bug was indeed introduced in #4991. However, this PR has not been released yet, so the issues on stable docs don't make any sense to me.

Screenshot 2025-01-16 at 10 10 02 AM

man-shu:
Updated doc build: https://github.com/nilearn/nilearn/actions/runs/12808487443/artifacts/2441538351.

The plotly figure can't be any more responsive than it is now. And we will deal with the other two issues in separate PRs.
So good to review now @bthirion @Remi-Gau

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -294,3 +296,206 @@ def test_surface_maps_masker_sample_mask_to_fit_transform(
...
     assert signals.shape == (3, masker.n_elements_)
+
+
+# ------------------ Tests for reporting ------------------
+
+
+@pytest.mark.skipif(
+    not is_matplotlib_installed(),
+    reason="Test requires matplotlib to be installed.",
+)
```
Remi-Gau, line 307: I think it may be useful to maybe use the `pyplot` fixure

https://github.com/nilearn/nilearn/blob/bc67775f6153c533e18edaabdb551b7f4c1f5e4e/nilearn/plotting/tests/conftest.py#L7

Like used here for example: 

https://github.com/nilearn/nilearn/blob/bc67775f6153c533e18edaabdb551b7f4c1f5e4e/nilearn/plotting/tests/test_surf_plotting.py#L275

it will just skip the test if matplotlib is not there

feel free to move the content from `nilearn/plotting/tests/conftest.py` to `nilearn/conftest.py`

nilearn/reporting/data/html/report_body_template_surfacemapsmasker.html
```diff
@@ -0,0 +1,108 @@
...
+        <tbody>
+          {{py: vertices = n_vertices.items()}} {{for hemi, nb in vertices}}
+          <tr>
+            <td data-column="Hemisphere">{{hemi}}</td>
+            <td data-column="Number of vertices">{{nb}}</td>
+          </tr>
+          {{endfor}}
+        </tbody>
+      </table>
+    </div>
```
Remi-Gau, line 76: realise that the glm report code has a way to easily turn pandas dataframe into html code: let's refactor after this PR is in

nilearn/reporting/data/html/report_body_template_surfacemapsmasker.html
```diff
@@ -0,0 +1,108 @@
...
+
+  function displayPreviousMap() {
+    document.getElementById("map" + current_map_idx).style["display"] = "none";
+    current_map_idx = current_map_idx - 1;
+    if (current_map_idx < 0) {
+      current_map_idx = number_maps - 1;
+    }
+    document.getElementById("map" + current_map_idx).style["display"] = "block";
+    document.getElementById("comp").innerHTML = displayed_maps[current_map_idx];
+  }
```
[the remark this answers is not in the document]
emdupre: +1 to extract and also to document in a follow-up PR ; most nilearn devs don't have JS expertise so longterm maintenance gets tricker for these snippets unless they're clearly marked and have some description !

nilearn/maskers/tests/test_surface_maps_masker.py
man-shu: Nice! Is it ok if I rename `pyplot` fixture to `matplotlib`?
I got confused for a second between `plotly` and `pyplot`.

Remi-Gau: > Is it ok if I rename `pyplot` fixture to `matplotlib`?

yup fine: though makes me realize that having a fixture with the same name as a package or a module sounds like recipe for disaster if we end up doing `import matplolib` and no alias it via `import matplotlib as mpl`

man-shu: Yeah fair point. How about matplotlib_pyplot?

man-shu: Also @Remi-Gau how did you get around ARG001 Ruff failures in [`nilearn/nilearn/plotting/tests/test_surf_plotting.py`](https://github.com/nilearn/nilearn/blob/bc67775f6153c533e18edaabdb551b7f4c1f5e4e/nilearn/plotting/tests/test_surf_plotting.py#L275)

Screenshot 2025-01-20 at 11 10 21 AM

Remi-Gau: used the very unsubtle nuclear option:

https://github.com/nilearn/nilearn/blob/b851ecc5a14a6d11aa9f8c2aa59cbede493e699f/nilearn/plotting/tests/test_surf_plotting.py#L3

man-shu: Done: https://github.com/nilearn/nilearn/pull/4968/commits/fe94ca58b1223effbf54a4c625eff3b5300fa760

nilearn/maskers/surface_maps_masker.py
```diff
@@ -429,3 +455,231 @@ def inverse_transform(self, region_signals):
...
+                )
+            fig = view_surf(
+                surf_map=roi,
+                bg_map=bg_img,
+                bg_on_data=True,
+                threshold=threshold,
+                hemi="both",  # TODO: use "both" after #4991 is merged
+            ).get_iframe()
+        elif self._report_content["engine"] == "matplotlib":
+            # TODO: possibly allow to generate a report with other views
```
bthirion, line 659: do you plan to do this in a future PR ?

bthirion (commented):
Almost done. Just one minor comment.

nilearn/maskers/surface_maps_masker.py
man-shu: I think it got copied from another masker. But let's open an issue to track it.

nilearn/maskers/surface_maps_masker.py
```diff
@@ -429,3 +455,231 @@ def inverse_transform(self, region_signals):
...
+        if engine not in ["plotly", "matplotlib"]:
+            raise ValueError(
+                "Parameter ``engine`` should be either 'matplotlib' or "
+                "'plotly'."
+            )
+
+        # switch to matplotlib if plotly is selected but not installed
+        if engine == "plotly" and not is_plotly_installed():
+            engine = "matplotlib"
+            warnings.warn(
```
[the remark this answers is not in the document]
man-shu: Ok no sorry it's actually the opposite -- we run this test when we only have matplotlib and no plotly. We do have the `no_plotly` test env for that. And this test actually passes there: https://github.com/nilearn/nilearn/actions/runs/12871770600/job/35888836502#step:7:7061

So I am not sure why does codecov complain...

nilearn/maskers/surface_maps_masker.py
man-shu: * see #5081

Remi-Gau (changes requested):
changelog needs an update

man-shu:
Done!

man-shu:
Reports are at the bottom of this page: https://output.circle-artifacts.com/output/job/e70086e5-20f0-47e2-848b-0ac4fa5452bf/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html#surface-maskers

bthirion (approved):
LGTM, thx.

nilearn/maskers/surface_maps_masker.py
```diff
@@ -429,3 +455,231 @@ def inverse_transform(self, region_signals):
...
+                roi.data.parts[part] = np.squeeze(
+                    roi.data.parts[part], axis=-1
+                )
+            fig = view_surf(
+                surf_map=roi,
+                bg_map=bg_img,
+                bg_on_data=True,
+                threshold=threshold,
+                hemi="both",
+            ).get_iframe()
```
Remi-Gau, line 657: pass the cmap here:

the report show a divergent colormap and we most likely will be plotting probabilities that work better with a sequential map like inferno

image

Remi-Gau: also the iframe should probably not be so wide: that's a CSS issue we can maybe fix in a follow up PR

man-shu: > also the iframe should probably not be so wide: that's a CSS issue we can maybe fix in a follow up PR

Ok, for now, I hard-coded a smaller width, but I think Plotly allows the plots to be responsive to window size. Let's do that in a later PR.

man-shu: Better now

Screenshot 2025-01-24 at 10 00 49 AM

Remi-Gau (approved):

PR #4969, opened 2024-12-19
[MAINT] only run link checks on full build of the doc

Remi-Gau:
Closes none

Add a check in the doc build workflow and only run the linkcheck if we are doing a html-strict build (so on main or when requested on PR).

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -68,7 +68,7 @@ jobs:
             run: |
                 headCommitMsg=$(git show -s --format=%s)
                 if [[ $headCommitMsg == *"[skip doc]"* ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docs.yml
```diff
@@ -81,13 +82,24 @@ jobs:
             uses: actions/checkout@v4
             with:
                 ref: ${{ github.event.pull_request.head.sha }}
+        -   name: Check if we are doing a full or partial build
```
Remi-Gau, line 85: this 'pattern' appears now more than 3 times in this workflow and should be refactored (in a follow up PR)

bthirion (approved):
LGTM.

PR #4975, opened 2024-12-20
[FIX] ensure list of images can be passed if masked has a mask

### LINKED ISSUE #4974: [BUG] _fit_mask_img does not support list of `SurfaceImage` when mask_img already exists

### Operating system

- [ ] Linux
- [x] Mac
- [ ] Windows

### Operating system version

_No response_

### Python version

- [ ] 3.13
- [x] 3.12
- [ ] 3.11
- [ ] 3.10
- [ ] 3.9

### nilearn version

Dev

### Expected behavior

Support a list of `SurfaceImage`.

Whenever `mask_img` is not None, `_fit_mask_img` expects only one image and crashes when a list of `SurfaceImage` is passed as input.
```python
    def _fit_mask_img(self, img):
        """Get mask passed during init or compute one from input image.

        Parameters
        ----------
        img : SurfaceImage object or :obj:`list` of SurfaceImage or None
        """
        if self.mask_img is not None:
            if img is not None:
                check_same_n_vertices(self.mask_img.mesh, img.mesh) <<<<< supposes img to be SurfaceImage
            self.mask_img_ = self.mask_img
            return
```

_No response_

Remi-Gau:
- Closes #4974

### DIFF
--- a/nilearn/maskers/surface_masker.py
+++ b/nilearn/maskers/surface_masker.py
@@ -150,22 +150,26 @@ class SurfaceMasker(_BaseSurfaceMasker):
         ----------
         img : SurfaceImage object or :obj:`list` of SurfaceImage or None
         """
-        if self.mask_img is not None:
-            if img is not None:
-                check_same_n_vertices(self.mask_img.mesh, img.mesh)
-            self.mask_img_ = self.mask_img
-            return
-
         if img is None:
-            raise ValueError(
-                "Please provide either a mask_img "
-                "when initializing the masker "
-                "or an img when calling fit()."
-            )
+            if self.mask_img is None:
+                raise ValueError(
+                    "Please provide either a mask_img "
+                    "when initializing the masker "
+                    "or an img when calling fit()."
+                )
+
+            if self.mask_img is not None:
+                self.mask_img_ = self.mask_img
+                return
 
         if not isinstance(img, list):
             img = [img]
         img = concatenate_surface_images(img)
+
+        if self.mask_img is not None:
+            check_same_n_vertices(self.mask_img.mesh, img.mesh)
+            self.mask_img_ = self.mask_img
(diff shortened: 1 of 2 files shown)


nilearn/maskers/surface_masker.py
```diff
@@ -150,23 +150,27 @@ def _fit_mask_img(self, img):
...
+            if self.mask_img is None:
+                raise ValueError(
+                    "Please provide either a mask_img "
+                    "when initializing the masker "
+                    "or an img when calling fit()."
+                )
+
+            if self.mask_img is not None:
+                self.mask_img_ = self.mask_img
+                return
```
Remi-Gau, line 163: - deal with the cases where img is None first (move guard clause to the top)
- then coerce img to list

nilearn/maskers/surface_masker.py
```diff
@@ -150,23 +150,27 @@ def _fit_mask_img(self, img):
...
 
         if not isinstance(img, list):
             img = [img]
         img = concatenate_surface_images(img)
 
+        if self.mask_img is not None:
+            check_same_n_vertices(self.mask_img.mesh, img.mesh)
+            self.mask_img_ = self.mask_img
+            return
+
```
man-shu, line 173: I feel like this could be combined with the if block above?

man-shu (approved):
Just a minor point, but feel free to ignore and merge

nilearn/maskers/surface_masker.py
Remi-Gau: how so?

man-shu: ok maybe not just had a _feeling_ 😝

Remi-Gau: # 🤣

PR #4976, opened 2024-12-20
[TEST] do not install latex for doc build

Remi-Gau:
- Closes none

removes latex config as it seems not needed for a proper rendering of :math: directive content and we do not produce PDF of the doc anymore

example of :math: directive left unchanged in this PR.

https://output.circle-artifacts.com/output/job/63793075-cbf6-449b-a580-b4f060126c7e/artifacts/0/dev/glm/glm_intro.html#multiple-comparisons

image

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -149,19 +149,6 @@ qthelp:
 	@echo "To view the help file:"
 	@echo "# assistant -collectionFile $(BUILDDIR)/qthelp/nilearn.qhc"
 
-latex:
-	$(SPHINXBUILD) -b latex $(ALLSPHINXOPTS) $(BUILDDIR)/latex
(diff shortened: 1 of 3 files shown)


man-shu:
I have seen some Sphinx docs that could be downloaded as PDFs. I know we don't provide that function on our docs as of now but could be nice to have. 

Would we still need Latex for the conversion?

bthirion:
I don't see myself downloading a big doc as a pdf. I'm curious whether many people would ?

Remi-Gau:
We have not provided PDF for the doc for a while (or ever?) and no one has asked for.

I would say: let's worry about when people ask for it.

man-shu (approved):
LGTM!

PR #4977, opened 2024-12-21
[STY] Apply assorted ruff preview rules

DimitriPapadopoulos:

- [E203](https://docs.astral.sh/ruff/rules/whitespace-before-punctuation/)
- [C420](https://docs.astral.sh/ruff/rules/unnecessary-dict-comprehension-for-iterable/)
- [FURB118](https://docs.astral.sh/ruff/rules/reimplemented-operator/)
- [FURB156](https://docs.astral.sh/ruff/rules/hardcoded-string-charset/)
- [RUF036](https://docs.astral.sh/ruff/rules/none-not-at-end-of-union/)
- [RUF039](https://docs.astral.sh/ruff/rules/unraw-re-pattern/)

### DIFF
--- a/examples/07_advanced/plot_beta_series.py
+++ b/examples/07_advanced/plot_beta_series.py
@@ -126,7 +126,7 @@ print("Define and fit LSA")
 # Transform the DataFrame for LSA
 lsa_events_df = events_df.copy()
 conditions = lsa_events_df["trial_type"].unique()
-condition_counter = {c: 0 for c in conditions}
+condition_counter = dict.fromkeys(conditions, 0)
 for i_trial, trial in lsa_events_df.iterrows():
     trial_condition = trial["trial_type"]
     condition_counter[trial_condition] += 1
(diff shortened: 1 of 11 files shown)


examples/07_advanced/plot_beta_series.py
```diff
@@ -126,7 +126,7 @@
 # Transform the DataFrame for LSA
 lsa_events_df = events_df.copy()
 conditions = lsa_events_df["trial_type"].unique()
-condition_counter = {c: 0 for c in conditions}
+condition_counter = dict.fromkeys(conditions, 0)
```
bthirion, line 129: I don't really like it more.

bthirion (approved):
Looks good overall. Readability is not always improved.

examples/07_advanced/plot_beta_series.py
Remi-Gau: I would say let's go for it and revisit this one and some others when we discuss if we should avoid 'complex' python code in examples to make sure we are beginner friendly.

Remi-Gau (approved):

PR #4978, opened 2024-12-21
[STY] Appending to dictionary following its definition

DimitriPapadopoulos:

- Instead of adding items to a dictionary just after creation, it should be refactored to add those items into the dict definition itself.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -416,16 +416,15 @@ latex_logo = "logos/nilearn-transparent.png"
 latex_elements = {
     "classoptions": ",oneside",
     "babel": "\\usepackage[english]{babel}",
+    "preamble": r"""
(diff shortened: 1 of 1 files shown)


bthirion (approved):
LGTM, thx.

doc/conf.py
```diff
@@ -416,17 +416,16 @@
 latex_elements = {
     "classoptions": ",oneside",
     "babel": "\\usepackage[english]{babel}",
+    "preamble": r"""
+\usepackage{amsmath}\usepackage{amsfonts}\usepackage{bm}\usepackage{morefloats}
+\let\oldfootnote\footnote
+\def\footnote#1{\oldfootnote{\small #1}}
+""",
```
Remi-Gau, line 423: Am actually trying to figure out  how much of our latex config is required given we do not generate pdf version of our doc.

So this may actually end up being deleted.

Remi-Gau:
- made obsolete by #4976

PR #4979, opened 2024-12-23
[FIX] fix rendering unordered list in doc

Remi-Gau:
- Closes none

- remove extra indent in unordered list that would be rendered as quote blocks

See https://nilearn.github.io/stable/decoding/frem.html#frem-pipeline

image


doc/conf.py
```diff
@@ -216,7 +216,7 @@
 linkcheck_report_timeouts_as_broken = False
 
 # double default rate_limit_timeout
-linkcheck_rate_limit_timeout = 600
```
Remi-Gau, line 219: unrelated but threw a warning during doc build

man-shu (approved):
LGTM!

Remi-Gau:
The number of line change is pretty large but only affects the user guide over rendering issue so I think a single review is enough.

PR #4986, opened 2024-12-23
[REF] refactor test plotting

Remi-Gau:
- Closes none

- refactor several tests for plotting to make them smaller
- use pyplot fixture to clean up figures on teardown

### DIFF
--- a/nilearn/plotting/tests/test_img_plotting/test_plot_carpet.py
+++ b/nilearn/plotting/tests/test_img_plotting/test_plot_carpet.py
@@ -1,5 +1,7 @@
 """Tests for :func:`nilearn.plotting.plot_carpet`."""
 
+# ruff: noqa: ARG001
+
 import warnings
 
 import matplotlib.pyplot as plt
@@ -9,11 +11,12 @@ import pytest
 from nilearn.plotting import plot_carpet
 
 
-def test_plot_carpet(img_4d_mni, img_3d_ones_mni, img_4d_long_mni):
+def test_plot_carpet(pyplot, img_4d_mni, img_3d_ones_mni):
     """Check contents of plot_carpet figure against data in image."""
     display = plot_carpet(
         img_4d_mni, img_3d_ones_mni, detrend=False, title="TEST"
     )
+
     # Next two lines retrieve the numpy array from the plot
     ax = display.axes[0]
     plotted_array = ax.images[0].get_array()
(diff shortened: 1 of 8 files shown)


nilearn/plotting/tests/test_img_plotting/test_plot_carpet.py
```diff
@@ -46,12 +51,10 @@ def test_plot_carpet(img_4d_mni, img_3d_ones_mni, img_4d_long_mni):
         img_4d_long_mni.shape[-1] / 2
     )
     assert plotted_array.size == n_items
-    plt.close(display)
 
 
-def test_plot_carpet_with_atlas(img_4d_mni, img_atlas):
+def test_plot_carpet_with_atlas(pyplot, img_4d_mni, img_atlas):
```
bthirion, line 56: Sorry, what is the point ?

Remi-Gau: This uses the same fixture as in sklearn: 

https://scikit-learn.org/dev/developers/contributing.html#testing-and-improving-test-coverage

image

The fixture is in nilearn/plotting/tests/conftest.py

- checks if matplotlib is there, skips test if not
- setup: close any open figure before the test is run
- teardown: closes any figure open by the test

hence why I am removing all the plt.close('all') calls: the fixture does it for us even if the test fails: this makes sure tests are more independent and do not pollute each other

hndgzkn (approved):
It seems OK to me.

There are a few points, may be to consider for follow-up PRs:

- It might be an option to move it to `nilearn/conftest.py` file as there might be tests in other packages which might reuse `pyplot`.
- In that case will we still need this line? https://github.com/nilearn/nilearn/blob/87ab7395669320e8aeadc512f60093ba7df88aee/nilearn/conftest.py#L90 
- and this line? https://github.com/nilearn/nilearn/blob/87ab7395669320e8aeadc512f60093ba7df88aee/nilearn/conftest.py#L33

nilearn/plotting/tests/test_img_plotting/test_plot_carpet.py
Remi-Gau: probably worth adding something similar in our own doc

Remi-Gau:
> actually the close_all fixture is not used so we can probably just remove it

nope my bad: it is ALWAYS used for EVERY test even if plotting is not involved, so maybe better to have a fixture we explicitly call for tests that need clean up.

bthirion (commented):
Is this ready for merge ? LGTM.

Remi-Gau:
Yup ready for merge.

PR #4988, opened 2024-12-24
[FIX] fix deploy on release

Remi-Gau:
- Closes none

- use same workflow to deploy doc on main for dev and stable
- refactor most of the deploy into a bash script that can be run locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,7 +61,6 @@ coverage.xml
 nosetests_coverage/
 
 # documentation
-doc/.nojekyll
 doc/_build/
 doc/auto_examples/
 doc/building_blocks/generated/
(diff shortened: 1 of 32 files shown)


Remi-Gau:
For this one I would suggest, we will only know if it works after merging.
I would suggest:
- merge and check that the classic deploy dev doc on main is not affected: if it is we revert this PR and go back to the drawing board
- if the dev deploy is not affected then we can go a dummy test of a github release to see if it triggers a stable doc build and deploy

bthirion:
Looks good, but it should be checked whether the deployment work well...

bthirion:
> For this one I would suggest, we will only know if it works after merging. I would suggest:
> 
>     * merge and check that the classic deploy dev doc on main is not affected: if it is we revert this PR and go back to the drawing board
> 
>     * if the dev deploy is not affected then we can go a dummy test of a github release to see if it triggers a stable doc build and deploy

Makes sense.

bthirion (approved):

man-shu (approved):
Yes let's try it

Remi-Gau:
This failed on main.
Reverting.

PR #4989, opened 2024-12-24
[DOC] add page for older doc versions

Remi-Gau:
- Closes none

- add a page to link to older versions of the doc similar to what can be found in sklearn doc: https://scikit-learn.org/dev/versions.html
- I had to to do some git archeology to dig out old versions of the doc and put them in separate folder in the doc repo: https://github.com/nilearn/nilearn.github.io
-

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -315,7 +315,7 @@ Featured examples
    maintenance.rst
    changes/whats_new.rst
    authors.rst
-   versions.rst
+   Versions <versions.rst>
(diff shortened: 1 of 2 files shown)


Remi-Gau:
If we want to keep this:
- dig out older versions of the doc on the doc repo and copy them in separate folders
- minify old versions of the doc to reduce size
- update release protocol to put a copy of the stable doc in a version folder

bthirion (approved):
LGTM, thx !

Issue #2747, opened 2021-04-01
Thresholding surface images


alpinho:
Functions [`nilearn.glm.threshold_stats_img`](https://nilearn.github.io/modules/generated/nilearn.glm.threshold_stats_img.html#nilearn.glm.threshold_stats_img) and [`nilearn.image.threshold_img`](https://nilearn.github.io/modules/generated/nilearn.image.threshold_img.html) do not accept input formats other than NIfTI. This poses a limitation if one wants, for instance, to obtain the threshold of surface images. WDYT?

jeromedockes:
I think it should wait until representation of surface images and masks, and surface maskers, have progressed a bit, but definitely worth keeping in mind for those who will work on surface data in the upcoming sprint as an example of something the new interface should allow to implement easily

NicolasGensollen:
Linking this [Neurostars post](https://neurostars.org/t/smoothing-surface-data/20159) asking for a smoothing function (like `nilearn.image.smooth_img`) for surface data.

bthirion:
Makes sense. If you want to have something strictly equivalent to Gaussian smoothing, it may be a bit tricky, but we can have a simple version that does local averaging. @alexisthual any thought on that ?

alexisthual:
Regarding @alpinho 's comment, I quickly checked the implementation of `nilearn.image.threshold_img` and `nilearn.glm.threshold_stats_img`, and it seems doable to enable support for gifti files (or surface objects).

I agree with you @bthirion that an exact equivalent of Gaussian smoothing might be costly to derive (computing the exact `n x n` matrix of all geodesic distances on the mesh will be slow) and that one might, in most cases, only be interested in an approximation.
We could probably modify `nilearn.image.smooth_img` such that it runs a simple weighted local average when given a gifti file.

It could allow users to input an int `d_neighbours` setting the maximum distance (in number of voxels) between pairs of voxels used to compute this approximation:
- `d_neighbours=1` would mean that, for each voxel, we only take _direct_ neighbours in consideration to compute our mean
- `d_neighbours=2` would mean that we also take the direct neighbours of our direct neighbours
- eventually, `d_neighbours=None` (equivalent to +infinity) would derive the exact result (but not be advised).

Remi-Gau:
implemented in #4999

PR #4999, opened 2024-12-30
[ENH] add function to threshold surface image

Remi-Gau:
- Closes to #2747 
- Relates to #4879 

- Extend nilearn.image.threshold_img and nilearn.masking.load_mask_img to work with nilearn.surface.SurfaceImage

## TODO
- [x] add more specific test to make sure image are properly thresholded (check output)
- [x] extend support to 2D surface image


nilearn/surface/surface.py
```diff
@@ -1829,14 +1830,30 @@ def get_data(img):
...
+
+    if ensure_finite:
+        non_finite_mask = np.logical_not(np.isfinite(data))
+        if non_finite_mask.sum() > 0:  # any non_finite_mask values?
+            warn(
+                "Non-finite values detected. "
+                "These values will be replaced with zeros.",
+                stacklevel=3,
+            )
+            data[non_finite_mask] = 0
```
Remi-Gau, line 1854: this constitutes 90% of the code of nilearn._utils.niimg.get_safe_data

https://github.com/nilearn/nilearn/blob/cd2871895581caacfcbd2bb443fa5d54304d94b5/nilearn/_utils/niimg.py#L59-L67

Remi-Gau: consider refactoring

nilearn/surface/surface.py
```diff
@@ -1975,3 +1992,139 @@ def new_img_like(ref_img, data):
         mesh=copy.deepcopy(mesh),
         data=data,
     )
+
+
+
+def threshold_img(
```
Remi-Gau, line 1998: A lot of this is duplicated from nilearn.image.threshold_img in terms of code and doc string.

Probably another candidate function may be able to handle both niimg and surfimg.

nilearn/surface/surface.py
```diff
@@ -1975,3 +1992,139 @@ def new_img_like(ref_img, data):
...
+            )
+    else:
+        # If there are more than 2 values, the mask is invalid
+        raise ValueError(
+            f"Given mask is not made of 2 values: {values}. "
+            "Cannot interpret as true or false."
+        )
+
+    mask = as_ndarray(mask, dtype=bool)
+    return mask, mask_img.affine
```
Remi-Gau, line 2130:
```suggestion
def load_mask_img(mask_img, allow_empty=False):
    """Check that a mask is valid, ie with two values including 0 and load it.

    Parameters
    ----------
    mask_img : SurfaceImage object
        See :ref:`extracting_data`.
        The mask to check.

    allow_empty : :obj:`bool`, default=False
        Allow loading an empty mask (full of 0 values).

    Returns
    -------
    mask : SurfaceImage
        Boolean version of the mask.
    """
    mask = get_data(mask_img, ensure_finite=True)
    values = np.unique(mask)

    if len(values) == 1:
        # We accept a single value if it is not 0 (full true mask).
        if values[0] == 0 and not allow_empty:
            raise ValueError(
                "The mask is invalid as it is empty: it masks all data."
            )
    elif len(values) == 2:
        # If there are 2 different values, one of them must be 0 (background)
        if 0 not in values:
            raise ValueError(
                "Background of the mask must be represented with 0. "
                f"Given mask contains: {values}."
            )
    else:
        # If there are more than 2 values, the mask is invalid
        raise ValueError(
            f"Given mask is not made of 2 values: {values}. "
            "Cannot interpret as true or false."
        )

    mask = as_ndarray(mask, dtype=bool)
    return mask, mask_img.affine
```


Almost all of this is duplicated from

https://github.com/nilearn/nilearn/blob/cd2871895581caacfcbd2bb443fa5d54304d94b5/nilearn/masking.py#L43-L88

Remi-Gau: consider refactoring

Remi-Gau:
OK for this one I am finding myself just copy-pasting so much code and tests from nilearn.image that I really think it makes no sense to have the surface 'utility' function live in a separate place than their volume counterpart.

Will implement surface thresholding in nilearn.image.

nilearn/image/image.py
```diff
@@ -1097,16 +1144,104 @@ def threshold_img(
...
+        img.data = img_data.data
+        return img
+
+
+def _get_img_data_for_cutoff(img, mask_img, data=None, mask_data=None):
+    """Get image data that thershold will be applied to.
+
+    Parameters
+    ----------
+    ----------*
```
bthirion, line 1167:
```suggestion

```

bthirion (commented):
The changes looks good, thx ! 
A few lines seems to be uncovered.

bthirion (approved):
LGTM, nice addition !

man-shu (approved):
LGTM!

PR #5008, opened 2024-12-31
[REF] extract nifti_maps_masker_ into property

Remi-Gau:
- Closes none

- extract nifti_maps_masker_ into property to avoid warning being thrown that users cannot do anything about


bthirion (commented):
LGTM, thx ! we just need a test.

PR #5012, opened 2025-01-02
[DOC] add ':obj:' link in doc strings

Remi-Gau:
- Closes none

- add a maintenance script to detect parameter type definitions with missing directive to link to common object definitions
- fix problematic doc strings
- refactor other maintenance scripts


maint_tools/missing_obj_in_docstring.py
```diff
@@ -0,0 +1,90 @@
...
+in the file `missing_objefault.md`.
+"""
+
+import ast
+
+from numpydoc.docscrape import NumpyDocString
+from rich import print
+from utils import list_functions, list_modules, root_dir
+
+VALUES = [
```
Remi-Gau, line 14: possibly also check for pandas dataframe and numpy arrays...

maint_tools/check_docstrings.py
```diff
@@ -0,0 +1,152 @@
...
+    "Boolean",
+    "list",
+    "List",
+    "tuple",
+    "Tuple",
+    "dict",
+    "Dict",
+    # "int",
+    "Int",
+    # "float",
```
man-shu, line 35: To be removed?

maint_tools/check_private.py
```diff
@@ -2,33 +2,16 @@
...
-    """Check functions of a module or methods of a class."""
-    function_definitions = [
-        node for node in body if isinstance(node, ast.FunctionDef)
-    ]
-    return [f.name for f in function_definitions if f.name.startswith("_")]
+from utils import list_functions, list_modules
 
 
 def main():
     """List private function that are only mnetioned once in their module."""
```
man-shu, line 11:
```suggestion
    """List private functions that are only mentioned once in their module."""
```

man-shu (approved):
LGTM otherwise

maint_tools/check_docstrings.py
Remi-Gau: na, this list is meant to be edited when one of us is looking for specific object that have not been documented properly: acts a bit as a 'filter'

man-shu: I see. So there would be situations where some of them don't need to tagged with :obj:

Remi-Gau: there are a few (and also because my script catches false positive)

PR #5022, opened 2025-01-06
[MAINT] pacify mypy

Remi-Gau:
- Closes none

- minor changes to make sure that mypy does not raise any error on the code base even if it does not have any type annotation (yet)

### DIFF
--- a/nilearn/__init__.py
+++ b/nilearn/__init__.py
@@ -52,7 +52,7 @@ except ImportError:
 
 # Monkey-patch gzip to have faster reads on large gzip files
 if hasattr(gzip.GzipFile, "max_read_chunk"):
-    gzip.GzipFile.max_read_chunk = 100 * 1024 * 1024  # 100Mb
+    gzip.GzipFile.max_read_chunk = 100 * 1024 * 1024  # type: ignore[attr-defined]
 
 # Boolean controlling the default globbing technique when using check_niimg
 # and the os.path.expanduser usage in CacheMixin.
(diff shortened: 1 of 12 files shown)


nilearn/__init__.py
```diff
@@ -52,7 +52,7 @@
 
 # Monkey-patch gzip to have faster reads on large gzip files
 if hasattr(gzip.GzipFile, "max_read_chunk"):
-    gzip.GzipFile.max_read_chunk = 100 * 1024 * 1024  # 100Mb
+    gzip.GzipFile.max_read_chunk = 100 * 1024 * 1024  # type: ignore[attr-defined]
```
Remi-Gau, line 55: I have not been able to find any documentation where this is used. Git blame was not very helpful in explaining why this was introduced.

Very tempted to remove it.

nilearn/plotting/surf_plotting.py
```diff
@@ -326,7 +326,7 @@ def _plot_surf_plotly(
     i, j, k = faces.T
 
     if cmap is None:
-        cmap = cold_hot
+        cmap = "cold_hot"
```
Remi-Gau, line 329:
```suggestion
        cmap = "RdBu_r"
```

nilearn/__init__.py
bthirion: Try it ?

Remi-Gau: ok but I doubt it will break anything since it may 'only' affect performance, but I could not find where this even exist in the gzip module so I think this won't actually change much.

bthirion (approved):
LGTM, thx.

Remi-Gau:
Reverting: realized I had forgotten to check something

PR #5037, opened 2025-01-09
[MAINT] Make functions used outside of their module public

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/issues/3319

- make several private functions (with a leading underscore) public as they are used outside of the module where they are defined.


nilearn/glm/second_level/second_level.py
```diff
@@ -60,6 +60,8 @@ def _check_second_level_input(
     second_level_input, design_matrix, confounds=None
 ):
     """Check second_level_input type."""
+    _check_design_matrix(design_matrix)
```
Remi-Gau, line 63: This one was actually NOT used anywhere, so I added back where it should be run.

man-shu:
Just for my clarification:

For us, private functions (the ones with a leading `_`) are simply the ones that are only used within a submodule/class and have nothing to do with how users are expected to use them. In other words, the functions made "public" here are still not recommended for users to use directly, as they can change without warnings.

I looked up if there are some conventions for this, but it seems that what's public vs. private can change from project to project. So maybe this is worth mentioning in the contributing doc page?

bthirion:
I guess it deserves some explanations, or a discussion in coredev meeting.

hndgzkn:
Considering @man-shu 's comments and nilearn's guidelines for private functions, I agree that it might be good to discuss. For example `set_mpl_backend`, I think is more implementation detail then a function to be publicly available to nilearn users.

Remi-Gau:
as it is not listed in any `__all__` variable of any module it nos available to users

man-shu (approved):
LGTM!

nilearn/plotting/surf_plotting.py
```diff
@@ -21,7 +21,7 @@
 from nilearn.plotting._utils import check_surface_plotting_inputs
 from nilearn.plotting.cm import cold_hot, mix_colormaps
 from nilearn.plotting.displays._figures import PlotlySurfaceFigure
-from nilearn.plotting.displays._slicers import _get_cbar_ticks
+from nilearn.plotting.displays._slicers import get_cbar_ticks
```
hndgzkn, line 24: As this is accessed from outside `nilearn.plotting.displays` shouldn't it be declared in `__all__` in `nilearn/plotting/displays/__init__.py`  and `nilearn.plotting.displays._slicers`?

hndgzkn (approved):
I've reread the contribution guidelines and if I do not interpret wrong, I've added some comments. 

Otherwise I think it is OK.

nilearn/plotting/surf_plotting.py
Remi-Gau: > shouldn't it be declared in __all__ in nilearn/plotting/displays/__init__.py

nope because in this case that would be making it part of the user facing API

nilearn/_utils/__init__.py
```diff
@@ -7,8 +7,8 @@
 from pathlib import Path
 
 from nilearn._utils.helpers import (  # noqa: F401
-    _constrained_layout_kwargs,
     compare_version,
+    constrained_layout_kwargs,
```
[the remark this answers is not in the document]
Remi-Gau: for things under nilearn._utils a lot of functions will be accessed by other modules, so if we start putting all the functions accessed by other modules in the __all__ of nilearn._utils we will end up with a very long list that we would have to always keep updating.

In the long run, I'd rather go the other way and remove the __all__ from nilearn._utils, so we are forced to import from a specific module in nilearn._utils.

PR #5049, opened 2025-01-15
[MAINT] Fix `check_docstrings` not checking any files and not catching unneeded `fill_doc`

michellewang:
- Closes none

- `maint_tools/utils.list_modules` ignores all modules if a parent directory above the repository root matches a folder to skip. For example if my Nilearn repo is at `/data/nilearn`, then the `check_docstrings.py` script doesn't work. Updated to only check parent directories within the `nilearn` root repo
- I noticed that [`nilearn.masking.load_mask_img`](https://github.com/nilearn/nilearn/blob/d52c31f79a74219aa38477fde267f4fd25600ce8/nilearn/masking.py#L43-L65) is decorated with `fill_doc` but does not have any `%(` substring inside it. Somehow that wasn't caught by `check_fill_doc_decorator`, so I updated it to catch that too
- Removed 2 instances of unneeded `fill_doc`

### DIFF
--- a/maint_tools/check_docstrings.py
+++ b/maint_tools/check_docstrings.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import ast
 import re
+from pathlib import Path
 
 from numpydoc.docscrape import NumpyDocString
 from rich import print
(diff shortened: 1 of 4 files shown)


maint_tools/check_docstrings.py
```diff
@@ -94,7 +95,12 @@ def check_fill_doc_decorator(
                 "- [red]missing @fill_doc decorator."
             )
     elif any(
-        getattr(x, "name", "") == "fill_doc" for x in ast_node.decorator_list
+        (
+            getattr(x, "name", "") == "fill_doc"
+            or getattr(x, "id", "") == "fill_doc"
+            or getattr(x, "attr", "") == "fill_doc"
```
michellewang, line 101: This was needed for the instances of `@fill_doc` and `@_utils.fill_doc` in `nilearn.masking`

Remi-Gau (approved):

Remi-Gau:
thanks 
let me know when this is ready to merge

bthirion (approved):

PR #5061, opened 2025-01-17
[FIX] improve compliance with sklearn checks

Remi-Gau:
- Closes none but relates to #4538

Changes proposed in this pull request:

- blocked by #4917 
- fix last estimators that did not comply with `check_parameters_default_constructible`
- improve compliance with `check_no_attributes_set_in_init`

PR #5062, opened 2025-01-17
[FIX] better compliance with check_no_attributes_set_in_init

Remi-Gau:
- Closes none but relates to #4538


Remi-Gau:
combining with #5061 5061

PR #5069, opened 2025-01-20
[MAINT] add check dtype for all nifti maskers

Remi-Gau:
- Closes none

- extract dtype check from test_nifti_masker into a check to run on all nifti maskers


man-shu (approved):
LGTM!

PR #5077, opened 2025-01-21
[REF] refactor permuted_ols

Remi-Gau:
- relates to #3655


bthirion (commented):
LGTM so far !

man-shu (approved):
I think the function names are self-explanatory but I also think we implicitly decided to have at-least one-liner docstrings? Not a major point for me though

Issue #4335, opened 2024-03-21
Improving brain atlases and standardize them


mtorabi59:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Operating system

- [ ] Linux
- [X] Mac
- [ ] Windows

### Operating system version

Mac OS Version 13.4

### Python version

- [ ] 3.12
- [X] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

0.10.3

### Expected behavior

Me and @Remi-Gau found out that a couple of nilearn atlases may benefit from some standardization and some of them even have inconsistencies. 
Specifically:

- Schaefer atlas: 
  - [x] the region names list is in byte string format, it'd be better if they were in string format
  - [x] `labels_img` is filled with float values instead of integers, i.e. region ids are float numbers.
- AAL atlas:
  - [x] the region ids (the numbers filling the `labels_img`) start from 2001 instead of 1 and are not consecutive numbers
- Destrieux atlas:
  - [x] The number of region names (151) does not match with the number of region ids (148), i.e. there are regions that do not exist in `labels_img` even without masking or resampling
  - [x] region names are a list of tuples like (id, name) instead of being only a string containing the name of the region. 

- [x] it seems the list of region names in all atlases includes "background" except AAL and Schaefer. For some atlases the name considered for background is different, e.g. "bg" or "Background"

@Remi-Gau feel free to add to the list
  

### Current behavior & error messages

NA

### Steps and code to reproduce bug

NA

Remi-Gau:
from #4329 

- ensuring that nilearn atlases have a "standardized" data structure
  (for example that each fetcher returns a list of labels as a list of strings)
- checking the label types passed to the constructor of NiftiLabelMasker
  and throw errors if they do not pass
  instead of warnings like they do here.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/4289#discussion_r1530968406_

---

Even though working on this should fix #4330 and #4331, it will change what it returned by some of the fetchers (some even if not an API change, it would be a behavior change that should probably go through a deprecation cycle - not 100% about this).

For example, the Destrieux fetcher would return 'labels' as list of strings and not a numpy.recarray containing the names of the ROIs.

So this feels like a bit of an "invasive" change to include into 0.10.4 that should be a bug fix release. So I suggest making that change after 0.10.4 is out.

> Note that the list of labels should also include "background" (or "Background")

[...]

PR #5085, opened 2025-01-23
[ENH] standardize yeo atlas fetcher

Remi-Gau:
- Closes none but relates to #4335

- start deprecation of current output of the fetcher
- add parameters to choose which parcellation to return
- update calls to yeo fetcher in examples


nilearn/datasets/atlas.py
```diff
@@ -1467,6 +1500,24 @@ def fetch_atlas_yeo_2011(data_dir=None, url=None, resume=True, verbose=1):
 
         - %(atlas_type)s
 
+        otherwise the keys are:
+
+        - 'anat': :obj:`str`
+            Path to nifti file containing the anatomy image.
+
+        - 'maps': 3D :class:`~nibabel.nifti1.Nifti1Image`.
+          The image contains integer values for each networf.
```
Remi-Gau, line 1509:
```suggestion
          The image contains integer values for each network.
```

bthirion (commented):
LGTM. Cleaning simply needed in whatsnew ?

Remi-Gau:
> Cleaning simply needed in whatsnew ?

yup!!! thanks I had forgotten about those

nilearn/datasets/atlas.py
```diff
@@ -1412,10 +1419,36 @@ def fetch_atlas_yeo_2011(data_dir=None, url=None, resume=True, verbose=1):
     %(resume)s
     %(verbose)s
 
+    n_networks : {7, 17, None}, default = None
```
man-shu, line 1422:
```suggestion
    n_regions : {7, 17, None}, default = None
```
I know "regions" and "networks" are sometimes used interchangeably in atlases but maybe it would be better to keep it consistent here?

Would require changes warnings and errors too

Remi-Gau: for this one I used netwroks to try to match hat is used in this fetcher that mentions yeo's netwroks: https://nilearn.github.io/dev/modules/generated/nilearn.datasets.fetch_atlas_schaefer_2018.html#nilearn.datasets.fetch_atlas_schaefer_2018

also I believe that networks is more appropriate as several disconnected 'regions' do compose a network: see https://nilearn.github.io/dev/auto_examples/06_manipulating_images/plot_extract_regions_labels_image.html#sphx-glr-auto-examples-06-manipulating-images-plot-extract-regions-labels-image-py

what do you think?

man-shu: Agreed! Then I guess regions could changed to networks

nilearn/datasets/atlas.py
```diff
@@ -1412,10 +1419,36 @@ def fetch_atlas_yeo_2011(data_dir=None, url=None, resume=True, verbose=1):
...
+        If not None,
+        then only specific version of the atlas is returned:
+
+        - 7 regions parcellation,
+        - 17 regions parcellation.
```
man-shu, line 1427:
```suggestion
        - 7 networks parcellation,
        - 17 networks parcellation.
```
Like here

nilearn/datasets/atlas.py
Remi-Gau: probably worth checking that the correct terminology is used across fetchers

nilearn/datasets/atlas.py
man-shu: I guess that's all. Can't find any other instances

doc/get_data_examples.py
```diff
@@ -65,7 +65,6 @@ def main(args=sys.argv) -> None:
         datasets.fetch_atlas_juelich("maxprob-thr0-1mm")
         for dimension in [10, 20]:
             datasets.fetch_atlas_smith_2009(resting=False, dimension=dimension)
-        datasets.fetch_atlas_yeo_2011()
```
Remi-Gau, line 68: removing a duplicate

nilearn/regions/region_extractor.py
Remi-Gau, line 1: change in this file are unrelated but should prevent some warning  of the type:

Data array used to create a new image contains 64-bit ints. This is likely due to creating the array with numpy and passing `int` as the `dtype`. Many tools such as FSL and SPM cannot deal with int64 in Nifti images, so for compatibility the data has been converted to int32.

man-shu (approved):
LGTM!

PR #5090, opened 2025-01-24
[ENH] Move js out of HTML report templates

### LINKED ISSUE #5064: [ENH] Move js out of masker report templates
TODO in follow up PR?

extract into a js file or a partial to reuse across maps maskers

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/4968#discussion_r1920425354_

man-shu:
- Closes #5064

- store javascript as separte .js.tpl file
- use tempita to insert variables in this javascript template
- unescape the safe HTML insertions


nilearn/reporting/data/js/maps_buttons.js.tpl
Remi-Gau, line 1: maybe better to have just a .js extension for this, no?

man-shu: > maybe better to have just a .js extension for this, no?

I did that in the beginning, but I think that would be misleading because this isn't exactly a js file.
This is more of a template for tempita to read and then insert all the variables between the curly braces `{{ }}`.
Only after the substitution, it becomes a javascript, which is then inserted into the HTML.

> having a top comment saying what this file contains and does may help.
Indeed! Will add!

Remi-Gau: > This is more of a template for tempita to read and then insert all the variables between the curly braces `{{ }}`.

wonder if there is a way to avoid doing this : so we can still have pure js

maybe something to work on later: for now it is enough to extract that code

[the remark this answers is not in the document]
man-shu: Nice!

man-shu: AFAIK, I think that would require creating a local server for interfacing between python and js using jinja. 
I think niivue does something like that when we open their plots in a browser

Remi-Gau:
@man-shu will let you handle the conflict resolution on this one

nilearn/reporting/data/js/maps_carousel.js.tpl
Remi-Gau, line 1: not sure if possible but maybe we can refactor the 2 js.tpl files to have a single one

I suspect this will require touching some of the maskers code so let's do it in a follow up PR

man-shu: Yeah I think it could be done. I'll open a separate issue.

Remi-Gau:
@man-shu 
can you trigger a full doc build so we can check the template report?

man-shu:
A loooot of links seem to be down. 

But the host_docs workflow did get triggered, however only partial docs have been deployed:

https://output.circle-artifacts.com/output/job/33420c69-543d-4ad7-bb3c-3b91d6060437/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html

https://output.circle-artifacts.com/output/job/33420c69-543d-4ad7-bb3c-3b91d6060437/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html

Remi-Gau:
> A loooot of links seem to be down.

I think we should just stop checking the DOI links: they are too fickle

Remi-Gau:
> But the host_docs workflow did get triggered, however only partial docs have been deployed:

OK that's weird and unexpected.

Remi-Gau (approved):
good with me

let's check why the reports are not rendering in the doc build in a follow up

man-shu:
Ok, so seems like the full doc is deployed when all doc workflows are successful, and the partial one is deployed when some of them aren't.

Here are the masker reports:
https://output.circle-artifacts.com/output/job/ce4ccb5e-6136-4836-97ec-ff26ec234f5d/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html

bthirion (approved):
LGTM, thx.

PR #5099, opened 2025-01-27
[MAINT] check maskers fit return self

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/issues/4538

- make sure nilearn estimators return self when calling fit
- some estimators are not yet covered by this (GLM ones for example)


bthirion (approved):
LGTM, thx.

PR #5104, opened 2025-01-28
[FIX] fix typo that prevents doc build on main

Remi-Gau:
- Closes none

- fix typo that cause crash of doc build on main


man-shu (approved):
Haha can't believe we missed that!

Issue #5095, opened 2025-01-27
[META] Standardizing maskers


Remi-Gau:
This is a meta issue to keep track of potential checks / tests to implement to make sure the behavior / API of our maskers is consistent across maskers.

I am opening it so we can refer to this issue when working on a PR related to this.

--- 

Sometimes it may be just an issue of refactoring similar tests across maskers into a single test run for all of them via `check_estimator` (see nilearn._utils.class_inspect).

Other times it may require adding a new common test for all maskers to check / ensure that all maskers comply with this behavior.

Doing this may reveal differences that across maskers that may or may not be justified and should potentially be corrected. 

---

- [x] see if possible to use a single test for using sample_mask for scrubbing when calling transform()
- [x] test smoothing for maskers: https://github.com/nilearn/nilearn/pull/5094
- [x] name of the first parameter to pass to maskers.fit is sometimes: X, sometimes imgs, sometimes img...
- [x] check transforming list of images with different affines / shapes
- [x] check joblib usage (memory, caching... see https://github.com/nilearn/nilearn/pull/5147#issuecomment-2652965692
- [x] check_generate_report:
  - [x] https://github.com/nilearn/nilearn/pull/5129

--- 

### Possibly related

- https://github.com/nilearn/nilearn/issues/4345
- https://github.com/nilearn/nilearn/issues/3413
- https://github.com/nilearn/nilearn/issues/2720
- https://github.com/nilearn/nilearn/issues/5101

Remi-Gau:
pretty sure this issue is either done or obsolete and replaced by more specific ones

PR #5107, opened 2025-01-28
[REF] refactor test to check 5D input to nifti masker transform

Remi-Gau:
- Closes none but relates to #5095 

- add test to check that nifti maskers cannot transform 5D but that multi niftimasker can


bthirion (approved):
LGTM, thx.

PR #5109, opened 2025-01-29
[FIX] ensure datasets do not contain bytes or numpy arrays 

### LINKED ISSUE #2263: [META] Replace bytestrings across all fetchers with unicode strings
Most atlases & many fetchers currently use byte-strings in their labels and descriptions etc. This is suboptimal and should be changed to regular strings.

3 examples of this are #1635 , #1181  and `fetch_haxby` .

Question, this will be a backwards incompatible change. Does that matter for fetchers used didactic training?
Will it break anything with the atlases, which might be used in the code?

Everyone's input here is appreciated.

### LINKED ISSUE #2660: Deprecate bytes and metadata as numpy arrays in datasets
Most fetchers in `nilearn.datasets` currently return descriptions and label names in `bytes`. See issue #1635 and recent PR #2655 for example.
Moreover, metadata is currently returned as numpy arrays which is not optimal from a usage perspective. 
For example:

```python
from nilearn.datasets import fetch_adhd
dset = fetch_adhd()
print(type(dset['phenotypic']))
print(dset['phenotypic'][0])
```
 outputs

```
<class 'numpy.ndarray'>
(b'"21"', 10042, b'"rest_1"', 0.0559, 0, 0.2365, 0.0922, 0., 2.2915, 1.0089, b'"NYU"', b'NA', b'"data_set"', 10.65, b'"M"', b'"0.91"', b'NA', b'108', b'100', b'115', b'2', 0, 1, b'NA', b'NA', b'NA', b'""', b'""', b'""', b'NA', b'NA', b'NA', b'NA', b'NA', b'NA', b'NA', b'NA', b'NA', b'NA', b'NA', b'59', b'56', b'65', b'NA', b'"pass"', b'""', b'"pass"', b'""', b'""', b'""', b'NA', b'NA', b'NA', b'NA', b'NA', b'NA', b'"pass"', b'NA', b'NA', b'NA', b'NA', b'NA', b'NA', b'""', b'""')
```

Since we were planning to use Pandas instead of Numpy for this (see #2621), I think this could be the opportunity to update all fetchers to return proper python strings and pandas dataframes. 

Since this would be a breaking change, we should add a deprecation message in the datasets module, and aim for releasing these changes in version `0.9`.

Remi-Gau:
- Closes #2660
- Closes #2263

- add checks in datasets tests to restrict allowed variables types returned by fetchers
- adapt fetchers to comply

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -1490,7 +1490,8 @@ def fetch_megatrawls_netmats(
 
         - 'matrices': str, consists of given type of specific matrices.
 
-        - 'correlation_matrices': ndarray, consists of correlation matrices
+        - 'correlation_matrices': pd.DataFrame
+          consists of correlation matrices
           based on given type of matrices. Array size will depend on given
           dimensions (n, n).
 
@@ -1555,8 +1556,8 @@ def fetch_megatrawls_netmats(
     # Fetch all the files
     files = fetch_files(data_dir, filepath, resume=resume, verbose=verbose)
 
-    # Load the files into arrays
-    correlation_matrices = csv_to_array(files[0])
+    # Load the files into dataframe
+    correlation_matrices = pd.read_table(files[0], sep=r"\s+", header=None)
 
     return Bunch(
         dimensions=dimensionality,
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -1752,20 +1752,15 @@ def fetch_surf_nki_enhanced(
     )[0]
 
     # Load the csv file
-    phenotypic = np.genfromtxt(
+    phenotypic = pd.read_csv(
         phenotypic,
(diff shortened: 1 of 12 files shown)


nilearn/datasets/func.py
```diff
@@ -1382,15 +1378,16 @@ def fetch_mixed_gambles(
 
         - 'zmaps': :obj:`list` of :obj:`str`
           Paths to realigned gain betamaps (one nifti per subject).
+        - 'subject_id':  pd.DaaFrame of subjects IDs
```
bthirion, line 1381:
```suggestion
        - 'subject_id':  pd.DataFrame of subjects IDs
```

bthirion (commented):
LGTM overall.

man-shu (approved):
LGTM!

PR #5113, opened 2025-01-30
[DOC] Masking huge fMRI images efficiently

man-shu:
- Closes nilearn#4811

- compare masking with NiftiMasker vs using dataobj and simple np masking vs np masking with shared mem
- use [`memory_profiler`](https://github.com/pythonprofilers/memory_profiler)
- final doc: https://output.circle-artifacts.com/output/job/db3b6d03-d533-4060-babc-4e1fbd4da022/artifacts/0/dev/auto_examples/07_advanced/plot_mask_large_fmri.html


man-shu:
Here's what the comparison of memory usage looks like
memory_profiling_10_images_20_regions_edited_figure_1_edited

Still working on how to properly show this in a jupyter notebook way

bthirion:
Thx ! Just trying to understand: what does the 78GB jump at the end correspond to ?
The output is much smaller, right ?
Best,

examples/07_advanced/plot_efficient_data_loading.py
```diff
@@ -0,0 +1,142 @@
...
+    del ts_nifti_masker
+
+    print("waiting")
+    time.sleep(30)
+    print("start efficient masker")
+
+    ts_efficient = mask_fmri_efficient_parallel(
+        fmri_path, mask_paths, n_jobs=n_regions
+    )
+    del ts_efficient
```
man-shu, line 138: Here we are trying to mask a big-ish fmri image, 20 times in parallel -- first with NiftiMasker (L126) and then with numpy indexing (L135).

So in both these cases a majority of memory usage is because we load this fmri image 20 times simultaneously. We can check this like this:


```python
# load the image once and check memory usage
In [3]: %memit load_img("temp/fmri.nii.gz")
# it's > 3GBs in memory
peak memory: 3714.92 MiB, increment: 3521.98 MiB

# multiply with 20 and divide by 1024 to convert to GBs
In [5]: (3714.92 * 20)/1024
Out[5]: 72.55703125
```

man-shu, line 138: Here we are trying to mask a big-ish fmri image, 20 times in parallel -- first with NiftiMasker (L126) and then with numpy indexing (L135).

In both these cases, the majority of memory usage is due to the fact that we load this fmri image 20 times simultaneously. We can check that like this:

```python
# load the image once and check memory usage
In [3]: %memit load_img("temp/fmri.nii.gz")
# it's > 3GBs in memory
peak memory: 3714.92 MiB, increment: 3521.98 MiB

# multiply with 20 and divide by 1024 to convert to GBs
In [5]: (3714.92 * 20)/1024
# close to what we see with numpy masking, but NiftiMasker has big overhead
Out[5]: 72.55703125
```

man-shu: The first peak corresponds to NiftiMasker and last one to the numpy masking

examples/07_advanced/plot_efficient_data_loading.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+    np.copyto(shared_img, fmri_data)
+    del fmri_data
+
+    print("waiting")
+    time.sleep(30)
+    print("start shared masker")
+    ts_shared = mask_fmri_shared_parallel(
+        shared_img, mask_paths, n_jobs=n_regions
+    )
```
man-shu, line 160: Ok so @MengxingLiu 's solution works even better. 

In L147-152 we load the image in a shared memory, which is then shared in L158 across 20 parallel jobs (avoiding loading the data 20 times simultaneously).
memory_profiling_10_images_20_regions_shared

man-shu:
The example is here: https://output.circle-artifacts.com/output/job/bd9c3fd0-6c85-4088-8feb-592452f84141/artifacts/0/dev/auto_examples/07_advanced/plot_mask_large_fmri.html

man-shu:
On an 8 core M3 MacBook Pro, with N_SUBJECTS=5 and N_REGIONS=6, things look similar as before:

sphx_glr_plot_mask_large_fmri_001

man-shu:
@emdupre, it would be nice if you could have a look at this one. 
Just to make sure I am not missing anything.

emdupre:
> @emdupre, it would be nice if you could have a look at this one.
Just to make sure I am not missing anything.

Thanks ! Will have a look. I'm a little puzzled when skimming through the results in thread, but I'll check the example itself :)

man-shu:
> Thanks ! Will have a look. I'm a little puzzled when skimming through the results in thread, but I'll check the example itself :)

Great! Thanks!

emdupre:
> On an 8 core M3 MacBook Pro, with N_SUBJECTS=5 and N_REGIONS=6, things look similar as before:

Am I right that these plots are all for loading multiple large images in parallel, rather than working on parcels in parallel within one image ?

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,279 @@
...
+# the image.
+#
+# In addition, we will use :func:`nibabel.loadsave.load` function to load the
+# fMRI image as a proxy object. This will allow us to load the data directly
+# from the file as a numpy array without loading the entire image into memory.
+# You can find more information about this in the :mod:`nibabel` documentation,
+# here: https://nipy.org/nibabel/images_and_memory.html
+
+import numpy as np
+from nibabel import load
```
emdupre, line 161: small but this is an unusual import ; can you just load `nibabel as nib` and call `nib.load` ? That's how we do it in [other](https://nilearn.github.io/stable/auto_examples/03_connectivity/plot_atlas_comparison.html#extract-coordinates-on-yeo-atlas-parcellations) [examples](https://nilearn.github.io/stable/auto_examples/04_glm_first_level/plot_bids_features.html#first-level-model-estimation-one-subject)

man-shu:
> Am I right that these plots are all for loading multiple large images in parallel, rather than working on parcels in parallel within one image ?

So we have one large-ish fMRI image and we wanna extract several regions out of it in parallel.

I am trying to show that when we do that with NiftiMasker or numpy indexing, the image is loaded multiple times (which is == N_REGIONS or n_jobs).

However, if we load the image once in shared memory and then pass to joblib's Parallel, it's only loaded once and masked.

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,279 @@
...
+from pathlib import Path
+
+from nilearn.datasets import fetch_development_fmri
+from nilearn.image import concat_imgs
+
+N_SUBJECTS = 5
+N_REGIONS = 6
+
+fmri_data = fetch_development_fmri(n_subjects=N_SUBJECTS)
+fmri_img = concat_imgs(fmri_data.func)
```
emdupre, line 41: Do we want to do this with real data ? It doesn't seem to make the example _too_ heavy, although I remember that was a concern

man-shu: Yeah I wanted to keep it reasonable for the doc so that it doesn't slow it down even further.

But I made sure to check much heavier versions on an HPC side-by-side (like in this [comment](https://github.com/nilearn/nilearn/pull/5113#issuecomment-2627902466)) and the results seem to align -- at least between the HPC and the doc build.

emdupre:
As a first comment : I think the fairer comparison is `joblib` sharedmem semantics vs `SharedMemory` (the latter is only available Python 3.8+, though [there is open discussion about adding it as a joblib backend](https://github.com/joblib/joblib/issues/1094)) :

- https://joblib.readthedocs.io/en/latest/parallel.html#working-with-numerical-data-in-shared-memory-memmapping
- https://joblib.readthedocs.io/en/latest/auto_examples/parallel_memmap.html#speed-up-processing-of-a-large-data-array

I don't know that this is actually what's done here (looking at this [large, late spike](https://github.com/nilearn/nilearn/pull/5113#issuecomment-2624922851) suggest that it might not be), but I'll need to confirm !

man-shu:
> As a first comment : I think the fairer comparison is joblib sharedmem semantics vs SharedMemory

Thanks for the refs! I will look into these further.

> I don't know that this is actually what's done here (looking at this https://github.com/nilearn/nilearn/pull/5113#issuecomment-2624922851 suggest that it might not be), but I'll need to confirm !

Yeah, the older plots (memory usage vs. time) were coming from a different script, and now I've removed that script and switched to a different one that's suitable for our doc examples. All this has made it more complicated to follow than it needs to be. I will do some clean up of the conversation. 

Now I also know the reason behind the discrepancies noted in the bar plots above. So, I will improve the current example too.

man-shu:
Updated example: https://output.circle-artifacts.com/output/job/56aef91b-1f50-48d0-814b-5ed69d32b7f1/artifacts/0/dev/auto_examples/07_advanced/plot_mask_large_fmri.html

bthirion:
Looks good. Just notice that the conclusion "using numpy indexing with shared memory is the most efficient way to mask large fMRI images in parallel." is not very accurate.

man-shu:
> "using numpy indexing with shared memory is the most efficient way to mask large fMRI images in parallel."

Ah yes forgot to update that. Will do.

Also, just wanted to say, I am no rush to merge this one. There are still a few things I wanna look into, one, for example is @emdupre's comment about comparing joblib shared memory vs. multiprocessing one (currently used).

However, I think joblib is somehow sharing the memory already (looking at the last two bars on the right)
Screenshot 2025-02-03 at 11 26 28 AM

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,411 @@
...
+    data[data != 0] = 1
+    resampled_mask = new_img_like(
+        ref_niimg=resampled_mask,
+        data=data,
+        affine=resampled_mask.affine,
+        copy_header=True,
+    )
+    mask_imgs.append(resampled_mask)
+    resampled_mask.to_filename(path)
+    mask_paths.append(path)
```
Remi-Gau, line 100:
```suggestion
    data = resampled_mask.get_fdata()
    data[data != 0] = 1
    resampled_mask = new_img_like(
        ref_niimg=resampled_mask,
        data=data,
        affine=resampled_mask.affine,
        copy_header=True,
    )
    mask_imgs.append(resampled_mask)
    
    path = output_dir / f"mask_{i}.nii.gz"    
    resampled_mask.to_filename(path)
    mask_paths.append(path)
```

Remi-Gau (commented):
don't have much to say except that I learnt quite a few things 🙏🏽

man-shu:
> I learnt quite a few things

Me too!

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,412 @@
...
+        ref_niimg=resampled_mask,
+        data=data,
+        affine=resampled_mask.affine,
+        copy_header=True,
+    )
+    mask_imgs.append(resampled_mask)
+
+    path = output_dir / f"mask_{i}.nii.gz"
+    resampled_mask.to_filename(path)
+    mask_paths.append(path)
```
emdupre, line 101:
```suggestion
masks = datasets.fetch_atlas_basc_multiscale_2015(resolution=7).maps
mask_img = nib.load(masks)
resampled_mask = resample_to_img(
    mask_img,
    fmri_path,
    interpolation="nearest",
    copy_header=True,
    force_resample=True,
)

for idx in range(N_REGIONS):
    mask = (mask_img.get_fdata() == idx)
    resampled_mask = image.new_img_like(
        ref_niimg=fmri_path,
        data=mask,
        affine=mask_img.affine,
        copy_header=True,
    )
    
    mask_imgs.append(resampled_mask)
    path = output_dir / f"mask_{i}.nii.gz"
    resampled_mask.to_filename(path)
    mask_paths.append(path)
```

emdupre: Note that this changes the atlas we use, as I think a deterministic atlas is a little closer to the common use case (and so easier to understand !)

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,412 @@
...
+
+# %%
+# Furthermore, we can input the fmri image and the masks in two different ways:
+#
+# 1. Using the file paths
+# 2. Using the in-memory objects
+#
+# So we will measure the memory usage for both cases.
+#
+# Let's first create a dictionary to store the memory usage for each method
```
emdupre, line 129: I don't think I would introduce this here ; as discussed, it really should only impact the method (2) using array proxy with `nibabel` and `joblib`

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,412 @@
...
+# the image.
+#
+# In addition, we will use :func:`nibabel.loadsave.load` function to load the
+# fMRI image as a proxy object. This will allow us to load the data directly
+# from the file as a numpy array without loading the entire image into memory.
+# You can find more information about this in the :mod:`nibabel` documentation,
+# here: https://nipy.org/nibabel/images_and_memory.html
+#
+# As before we will do this by loading the data from the file paths and from
+# the in-memory objects. This time we will define two functions to do that.
```
emdupre, line 213: I don't know that it makes sense to show the non-memmap example here ; I'm lightly against it to keep things clean

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,412 @@
...
+# Masking using numpy indexing in parallel with shared memory
+# -----------------------------------------------------------
+# Finally, let's see how we can mask the fMRI image using multiple masks in
+# parallel using numpy indexing and shared memory from the
+# :mod:`multiprocessing` module.
+#
+# For this method, we would have to load the fMRI image into shared memory
+# that can be accessed by multiple processes. This way, each process can
+# access the data directly from the shared memory without loading the entire
+# image into memory.
```
emdupre, line 311:
```suggestion
# image into memory again.
```

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,412 @@
...
+import matplotlib.pyplot as plt
+
+plt.figure(figsize=(8, 6))
+plt.bar(
+    [
+        "NiftiMasker,\nwith path",
+        "NiftiMasker,\nwith in-memory\nimage",
+        "Numpy indexing,\nwith path",
+        "Numpy indexing,\nwith in-memory\nimage",
+        "Numpy indexing\nwith shared\nmemory",
```
emdupre, line 377: To be explicit, I think the bars should be
"NiftiMasker", "ArrayProxy", "SharedMemory"

emdupre (commented):
Thank you, @man-shu  ! Wanted to leave a few comments, following up on our conversation from yesterday.

man-shu:
Ok, now the results align with what we see on the memory vs. time plot in #5128. See here: https://output.circle-artifacts.com/output/job/f51f01a1-9e42-40ef-a7d5-fdb039f55c6a/artifacts/0/dev/auto_examples/07_advanced/plot_mask_large_fmri.html

Screenshot 2025-02-07 at 4 55 53 PM

I was making a very stupid mistake by still giving paths as input to the `NiftiMasker, parallel, in_memory` case, which is why the peak usage looked the same for them 🥲

examples/07_advanced/plot_mask_large_fmri.py
man-shu: So now that I've fixed my silly mistake. We do indeed see the difference between the two cases (NiftiMasker path vs. NiftiMasker in-memory). 

I don't know why that is the case, but as long as it exists, I think we should be highlighting that here.

We could probably look into the why question if we decide to improve things with NiftiMasker in future.

examples/07_advanced/plot_mask_large_fmri.py
man-shu: > to keep things clean

I agree that the example does read a bit repetitive. But I would say the difference between path vs. in-memory is pretty significant and since this is an advanced example, it does fall within the scope.

Maybe removing the "single" instances would make it better?

emdupre:
Huh, now I'm wondering why this would be the case !

As far as I know we call `.get_fdata()` inside NiftiMasker, so it should be loading all the data into memory ; I wonder if (when you've loaded the data into memory beforehand), you just don't _see_ this memory spike in the plot, even though it's required in the usage of NiftiMasker that all data is in memory..

I need to understand your memory profiling tools a bit better to actually investigate this, but this is my current guess !

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,414 @@
...
+# Let's add another key to the previous dictionary to store the memory usage
+# for this case.
+
+from joblib import Parallel, delayed
+
+
+def nifti_masker_parallel(fmri_path, mask_paths):
+    return Parallel(n_jobs=N_REGIONS)(
+        delayed(nifti_masker_single)(fmri_path, mask) for mask in mask_paths
+    )
```
man-shu, line 170: > so it should be loading all the data into memory

I think joblib is doing some memory management here because in `Parallel` by default `mmap_mode="r"`. If we change this to `mmap_mode=None` (which turns off this memory mapping), then all the "in_memory" cases fail (for NiftiMasker, numpy indexing and numpy indexing with shared memory), with the error at the bottom:

```suggestion
def nifti_masker_parallel(fmri_path, mask_paths):
    return Parallel(n_jobs=N_REGIONS, mmap_mode=None)(
        delayed(nifti_masker_single)(fmri_path, mask) for mask in mask_paths
    )
```

```bash
  File "/data/parietal/store3/work/haggarwa/miniconda3/envs/nilearnpy/lib/python3.13/site-pa
ckages/joblib/numpy_pickle.py", line 598, in load_temporary_memmap
    JOBLIB_MMAPS.add(obj.filename)
                     ^^^^^^^^^^^^
AttributeError: 'numpy.ndarray' object has no attribute 'filename'
```

man-shu: See joblib Parallel's [docstring](https://arc.net/l/quote/fkjrqazv)

man-shu:
> As far as I know we call .get_fdata() inside NiftiMasker

I looked into NiftiMasker a bit, and we instead use an internal `_get_data()`, which also loads it via array proxy using `.dataobj`:
https://github.com/nilearn/nilearn/blob/4ece106c78f26df9ef2556d68b190634223d9184/nilearn/_utils/niimg.py#L25

Same in `concat_imgs` (which is what we use here to indirectly load the fmri_img):
https://github.com/nilearn/nilearn/blob/4ece106c78f26df9ef2556d68b190634223d9184/nilearn/image/image.py#L1645

So I think when we provide path the data is loaded N_REGIONS times across workers but for an already in_memory image that's not the case. This explains the parallel case.

Note, however, that the difference is still there in the "single" masking case. There, my explanation here is that by loading the image beforehand we are offloading the memory peak outside the masking step but that peak shows up when we load the image within the masking step by providing the path.

man-shu:
I think we are ready for another round of reviews @emdupre @Remi-Gau @bthirion

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,414 @@
...
+from nilearn.image import load_img, new_img_like, resample_to_img
+
+atlas_path = fetch_atlas_basc_multiscale_2015(resolution=64).maps
+
+atlas_img = load_img(atlas_path)
+
+# %%
+# In this case, the atlas and the fMRI image have different resolutions.
+# So we will resample the atlas to the fMRI image. It is important to do that
+# because only :class:`~nilearn.maskers.NiftiMasker` can handle the resampling
```
bthirion, line 78: This is an interesting point (deserves an issue ?) but maybe out of topic here ?

bthirion (commented):
Looks mostly good.

examples/07_advanced/plot_mask_large_fmri.py
man-shu: Sorry I don't understand what you are referring to? 

We do resampling beforehand here because when masking with numpy, the image array and mask array need to have the same dimensions.

And also to make sure that NiftiMasker isn't doing anything more than masking -- so that the comparison is apples-to-apples.

Remi-Gau:
> Seems like we're fetching this dataset too many times

yeah we have had a few crashes in doc build because of things like this: we should improve our data caching

man-shu:
> yeah we have had a few crashes in doc build because of things like this: we should improve our data caching

I wonder if those crashes happen on the days we do our monthly (?) caching?

examples/07_advanced/plot_mask_large_fmri.py
emdupre: Given the current pattern of results, let's try to explain this difference in more detail somewhere and then (at least) point to that explanation (if it's not directly in this text).

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,414 @@
...
+# 1. Using the file paths
+# 2. Using the in-memory objects
+#
+# So we will measure the memory usage for both cases.
+#
+# Let's first create a dictionary to store the memory usage for each method
+
+from memory_profiler import memory_usage
+
+nifti_masker = {"single": {"path": [], "in_memory": []}}
```
emdupre, line 135: Maybe let's define the dictionary all at once rather than iteratively ? I don't think we've explained why we expect `parallel` to be different than `single` in any of the text to date ; this might be a good place to state that directly. 

```suggestion
nifti_masker = {
    "single": {"path": [], "in_memory": []},
    "parallel": {"path": [], "in_memory": []}
}
```

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,414 @@
...
+
+nifti_masker["single"]["path"] = memory_usage(
+    (nifti_masker_single, (fmri_path, mask_paths[0])),
+    max_usage=True,
+)
+print(
+    "Peak memory usage: with NiftiMasker, single mask, with paths:\n"
+    f"{nifti_masker['single']['path']} MiB"
+)
```
emdupre, line 144: Can we consolidate these print statements to the end of the `NiftiMasker` block ?

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,414 @@
...
+print(
+    "Peak memory usage: with NiftiMasker, single mask, with in-memory image:\n"
+    f"{nifti_masker['single']['in_memory']} MiB"
+)
+
+# %%
+# Masking using NiftiMasker in parallel
+# -------------------------------------
+# Now let's see how we would mask the fMRI image using multiple masks in
+# parallel using the :mod:`joblib` package.
```
emdupre, line 159: As noted above, we should explain why this is even another dimension to consider. I would also lightly suggest consolidating into one `NiftiMasker` section.

examples/07_advanced/plot_mask_large_fmri.py
emdupre: Great ! Yes, I thought NiftiMasker was calling `get_fdata()` but I'm glad to know that I was wrong :) This is a sensible handling of this !

emdupre: Maybe this error (and the explanation for it) could feature in the user guide ! But will resolve for this issue.

examples/07_advanced/plot_mask_large_fmri.py
emdupre: Yes, I think I'm leaning towards removing all of the `single` instances ; presumably folks can load a large array into memory at least once ! If not, then we're more likely in the "high resolution" rather than "long time series" case.

emdupre (changes requested):
Thanks, @man-shu ! This is really starting to come together.

I think to focus it a bit more, I'd really like to see all of the `single` instances removed (as you had suggested in your comment). We should probably also have some text here or in the User Guide PR to explain what's happening "under the hood" for "in-memory" vs "path" passing in the NiftiMasker and array proxy cases.

examples/07_advanced/plot_mask_large_fmri.py
bthirion: Sorry, I misread the paragraph.

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,414 @@
...
+Currently, this example focuses on memory-efficient interactions with long
+time series fMRI data. In this case, loading the whole time series into memory
+may represent a significant computational cost. We will therefore explore
+strategies to minimize the amount of data that is loaded into memory,
+and we will compare these strategies against a naive usage of
+:class:`~nilearn.maskers.NiftiMasker`.
+
+To make this more concrete, we will create a large fMRI image with over
+800 time points by concatenating individual subjects in the
+:func:`~nilearn.datasets.fetch_development_fmri` dataset.
```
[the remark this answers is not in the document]
man-shu: Oops I thought I updated that

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,414 @@
...
+2. Using array proxies defined with :mod:`nibabel`
+3. Using :class:`multiprocessing.shared_memory.SharedMemory`
+
+"""
+
+# %%
+# Create a large fMRI image
+# -------------------------
+# Here we will create a "large" fMRI image by fetch 5 subjects'
+# fMRI images via the :func:`~nilearn.datasets.fetch_adhd`
```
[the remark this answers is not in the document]
man-shu: Oh yeah I changed this one but forgot we talk about this in another place

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -0,0 +1,414 @@
...
+
+def nifti_masker_single(fmri_path, mask_path):
+    return NiftiMasker(mask_img=mask_path).fit_transform(fmri_path)
+
+
+# %%
+# Furthermore, we can input the fmri image and the masks in two different ways:
+#
+# 1. Using the file paths
+# 2. Using the in-memory objects
```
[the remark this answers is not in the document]
man-shu: Yeah User Guide makes sense to me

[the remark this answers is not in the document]
man-shu: Or let's first make sure this example is self-contained, merge it and then we can move stuff out of here to the user guide

examples/07_advanced/plot_mask_large_fmri.py
man-shu: I'm not storing the memory usage for single cases anymore. So this isn't needed

man-shu:
I think we should deal with this separately, but I wanted to test our hypothesis that joblib + memory mapped in_memory image is what's causing the path vs. in_memory difference. 

So I compared the image loaded via `load_img` (I think it doesn't use dataobj) vs. `concat_imgs` (defo uses dataobj):

With `load_img` the difference between NiftiMasker in_memory, numpy masker path and numpy masker in_memory disappears: https://output.circle-artifacts.com/output/job/c99698a5-290c-4ee7-9750-1e551c4774fa/artifacts/0/dev/auto_examples/07_advanced/plot_mask_large_fmri.html#conclusion

examples/07_advanced/plot_mask_large_fmri.py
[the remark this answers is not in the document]
man-shu: I've tried to explain (at least what I understand as of now) in the example for now. But we can move it out later.

man-shu:
Updated example: https://output.circle-artifacts.com/output/job/8d52e53d-4b23-4aa9-ac80-11402783d4b2/artifacts/0/dev/auto_examples/07_advanced/plot_mask_large_fmri.html

emdupre:
> So using numpy indexing with shared memory is the most efficient way to mask large fMRI images in parallel.

It seems like this isn't reflecting the current bar plot, where SharedMemory is the second best solution. I won't have time to do a full review until tomorrow, but it seems like the results have shifted a few times now between which result is best ; maybe we should reframe the takeaway on this to just point more generally to how users can improve on naive niftimasker usage ?

bthirion (approved):
LGTM, thx.

emdupre (approved):
Thanks, @man-shu ! I think this is a great addition that we can keep iterating on as we add in the accompanying user guide 🚀

PR #5117, opened 2025-02-03
[MAINT] add type checking for parameters listed in nilearn_utils.docs

Remi-Gau:
- Closes none

- add a way to enforce types of parameters that are aliased in our doc
- run checks for a subset of parameters (others to follow in subsequent PR)


nilearn/decoding/tests/test_objective_functions.py
```diff
@@ -74,8 +73,3 @@ def test_logistic_loss_derivative(rng, n_samples=4, n_features=10, decimal=5):
         0.0,
         decimal=decimal,
     )
-
-
-def test_baseestimator_invalid_l1_ratio():
```
Remi-Gau, line 79: moved to nilearn/decoding/tests/test_space_net.py

nilearn/typing.py
```diff
@@ -0,0 +1,54 @@
...
+
+MemoryLevel = int
+NJobs = int
+Resume = bool
+Standardize = bool
+Verbose = int
+
+# TODO update when dropping python 3.9
+if sys.version_info[1] < 10:
+    DataDir = (str, pathlib.Path)
```
Remi-Gau, line 26: from what I can tell python 3.9 isinstance is not too happy with using a type annotation like notation for its second argument

Remi-Gau: 3.9

```python
>>> isinstance(1, int | None)
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

3.13

```python
>>> isinstance(1, int | None)
True
```

nilearn/typing.py
```diff
@@ -0,0 +1,54 @@
...
+    DataDir: TypeAlias = str | pathlib.Path | None
+    DType: TypeAlias = DTypeLike | str | None
+    HrfModel = str | Callable | list | None
+    HighPass: TypeAlias = float | int | None
+    LowPass: TypeAlias = float | int | None
+    MemoryLike: TypeAlias = Memory | str | pathlib.Path | None
+    Resolution: TypeAlias = int | None
+    SmoothingFwhm = float | int | None
+    TargetAffine: TypeAlias = ndarray | None
+    TargetShape: TypeAlias = tuple | list | None
```
Remi-Gau, line 52: This probably won't be usable as for static type checking, as type checkers will complain about using a generic and would prefer : `list[int]` to `list`

nilearn/typing.py
```diff
@@ -0,0 +1,54 @@
...
+    Tr = (float, int)
+    Url = str
+
+
+else:
+    from typing import TypeAlias
+
+    DataDir: TypeAlias = str | pathlib.Path | None
+    DType: TypeAlias = DTypeLike | str | None
+    HrfModel = str | Callable | list | None
```
Remi-Gau, line 45: note that `str` is too generic here and it should actually be Literal["spm", "glover", ...] if we wanted to use proper type annotation

bthirion (commented):
In general, is there a rule to decide on which functions `check_params` should be called ? There seems to be something implicit here.
A few lines are not covered.

Remi-Gau:
> is there a rule to decide on which functions `check_params` should be called ?

Will update the doc string but it can be used for functions / class / methods that have the @fill_doc decorator

nilearn/_utils/param_validation.py
```diff
@@ -361,3 +363,85 @@ def _cast_to_int32(sample_mask):
...
+    This function is made to check the types of the parameters
+    described in ``nilearn._utils.docs``
+    that are shared by many functions / methods / class
+    and thus ensure a generic way to do input validation
+    in several important points in the code base.
+
+    In most cases this means that this function can be used
+    on functions / classes that have the ``@fill_doc`` decorator,
+    or whose doc string uses parameter templates
+    (for example ``%(data_dir)s``).
```
man-shu, line 383: So IIUC, for this check to work, would it be enough to write the expected type of parameters in the docstring? 

If so, is there some syntax that needs to be followed or simply using the `:obj:` rst tags is enough.

(I might be getting this completely wrong, so please correct me if that's the case)

Remi-Gau: > would it be enough to write the expected type of parameters in the docstring?

not quite: I wished it were that powerful

to expand the functionality of the function you need to:
- describe the expected type for that parameter / attribute in `nilearn.typing` (it must be something that `isinstance` can handle
- expand the `type_map` dictionary of `check_params` top pair the name of the parameter / attribute with its expected type

does that help?

man-shu: I see. So, everywhere we call this, we only check the type of a _subset_ of parameters defined in the function.

And, this subset consists of the parameters we have in `nilearn._utils.docs`/

Remi-Gau: this subset consists of **SOME** of the parameters we have in `nilearn._utils.docs`

ideally I would like to cover all of them but at the moment it is not the case

nilearn/_utils/param_validation.py
```diff
@@ -1,11 +1,13 @@
 """Utilities to check for valid parameters."""
 
 import numbers
+import sys
 import warnings
 
 import numpy as np
 from sklearn.feature_selection import SelectPercentile, f_classif, f_regression
 
+import nilearn.typing as nil_typing
```
bthirion, line 10:
```suggestion
import nilearn.typing as nilearn_typing
```

bthirion (commented):
Besides 1 uncovered line, the PR LGTM. Thx,

nilearn/typing.py
Remi-Gau: Depends how you see it...

python 3.9 EOL is in November :https://endoflife.date/python

so in terms of release that's 1 or 2 releases for us: pretty soon

nilearn/_utils/param_validation.py
man-shu: > to expand the functionality of the function you need to:

I was just thinking that it might be good to document this in some way? Maybe the contributing page? But since it isn't full-fledged typing maybe even top of the `typing.py` would suffice?

Remi-Gau: will do in a follow up PR to keep this one under 500 lines changed

PR #5122, opened 2025-02-06
[MAINT] apply check_estimators_unfitted to all estimators

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/issues/4538

- make sure all estimators pass the check_estimators_unfitted check


nilearn/decoding/decoder.py
```diff
@@ -824,6 +824,7 @@ def decision_function(self, X):
         """
         # for backwards compatibility - apply masker transform if X is
         # niimg-like or a list of strings
+        check_is_fitted(self)
```
man-shu, line 827:
```suggestion
        check_is_fitted(self)
        # for backwards compatibility - apply masker transform if X is
        # niimg-like or a list of strings
```

Just moved this above the comment

nilearn/maskers/tests/conftest.py
```diff
@@ -7,9 +7,7 @@
 
 def check_valid_for_all_maskers():
     """Return list of names of sklearn checks valid for all maskers."""
-    return [
-        "check_estimators_unfitted",
-    ]
+    return []
```
man-shu, line 10: do we still need this one if it's only returning empty list?

man-shu (approved):
Otherwise LGTM!

nilearn/maskers/tests/conftest.py
Remi-Gau: no we don't 
thanks for catching this one

PR #5130, opened 2025-02-06
[MAINT] use macos to build doc

Remi-Gau:
- Closes none

- just a test to see if the full build is faster


man-shu:
> just a test to see if the full build is faster

Was it?

Remi-Gau:
na: the doc build requires to install some things that are more easily done with unix

PR #5135, opened 2025-02-10
[ENH] allow numpy equivalent to int and float for several common parameters

Remi-Gau:
- Closes none but fixes doc build crash on main

- expand accepted numeric types for common parameters to accept numpy.floating and numpy.integer where appropriate

Should avoid error due to 

```python
>>> import numpy as np
>>> a = np.int32(1)
>>> isinstance(a, int)
False
```

### DIFF
--- a/examples/01_plotting/plot_demo_plotting.py
+++ b/examples/01_plotting/plot_demo_plotting.py
@@ -45,12 +45,14 @@ stat_img = datasets.load_sample_motor_activation_image()
 
 from nilearn import plotting
 
+# %%
 # Visualizing t-map image on EPI template with manual
 # positioning of coordinates using cut_coords given as a list
 plotting.plot_stat_map(
     stat_img, threshold=3, title="plot_stat_map", cut_coords=[36, -27, 66]
 )
 
+# %%
 # It's also possible to visualize volumes in a LR-flipped "radiological" view
 # Just set radiological=True
 plotting.plot_stat_map(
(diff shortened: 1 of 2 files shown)


examples/01_plotting/plot_demo_plotting.py
```diff
@@ -45,12 +45,14 @@
 
 from nilearn import plotting
 
+# %%
```
Remi-Gau, line 48: doing some changes to make sure that the examples that crashed on main can be run now

Remi-Gau, line 48: doing some changes to make sure that the examples that crashed on main can be run now

nilearn/typing.py
```diff
@@ -27,46 +27,46 @@
...
 Standardize = bool
 Tfce = bool
 TwoSidedTest = bool
-UpperCutoff = float
-Verbose = int
-
 
 # TODO update when dropping python 3.9
 if sys.version_info[1] < 10:
+    BorderSize = (int, np.integer)
```
Remi-Gau, line 39: I think the types parameter templates in nilearn/_utils.doc.py should be adapted to match, no?

@man-shu @bthirion

man-shu: Screenshot 2025-02-10 at 9 51 58 AM

This is what's there in `nilearn/_utils.doc.py` right now. Do you mean we should add np.integer too here?

man-shu: I feel like the difference between the two is too technical (explained [here](https://stackoverflow.com/a/38156630)) and `int` semantically covers all of them.

Remi-Gau: yeah I was not sure about how 'pedantic" we want our doc string to be

Remi-Gau:
OK let's keep things as they are for now in terms of doc

Issue #5134, opened 2025-02-10
Improvement suggestion for example: Understanding parameters of the first-level model


Remi-Gau:
Example (on dev):
https://nilearn.github.io/dev/auto_examples/04_glm_first_level/plot_first_level_details.html

Couple of things that should be improved:

- warning against p-hacking: eventhough the example mention "exploration", I think there should be a BIG warning on why people should not do the type of 'manual model selection" done in this example on their contrast(s) of interest because that can quickly lead to p-hacking. Proper alternative should be suggested: either test model on independent data or at least on a contrast independent to the contrast(s) of interest.

- plots of design matrix correlation should be added to help with diagnostic : https://nilearn.github.io/dev/modules/generated/nilearn.plotting.plot_design_matrix_correlation.html#nilearn.plotting.plot_design_matrix_correlation
  
  For example, 
  > The effect on the activation maps is complex: auditory/visual effects are killed, probably because they were somewhat colinear to the confounds. 

- avoid the "eyeball test": use whole plot_img_comparison rather just plotting a single cut at a single threshold

- some models that are compared seem to be nested and an F test could probably be used to compare them (if I am correct)

- more cosmetic: the design matrix plots are HUGE and the brain slices plots are tiny in comparison: the opposite would be preferable (or least bigger brain activation maps)

Remi-Gau:
- this type of model overall fit evaluation could be improved by adding things suggested in https://github.com/nilearn/nilearn/issues/4525

bthirion:
Thx !  Your points are indeed valid.
I'm less sure about:
* the use of plot_img_comparison: when only a few voxels are involved, the effect may just be hidden in scatter plots.
* F test to compare nested models. the eveidence may vary across voxels and anyway, the criterion of null rejection remains arbitrary...

emdupre:
Could be a good issue for the sprint ! Pre-emptively tagging, feel free to re-label

PR #5138, opened 2025-02-10
[DOC] improve FLM details example

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/issues/5134

- improve plotting range
- improve F test results display
- add design matrix correlation


examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -352,13 +389,16 @@ def plot_contrast(first_level_model):
 )
 first_level_model = first_level_model.fit(fmri_img, events=events)
 z_map = first_level_model.compute_contrast(contrast_val, output_type="z_score")
-plotting.plot_stat_map(
+plot_stat_map(
     z_map,
     display_mode="z",
-    threshold=3.0,
+    threshold=1.96,
```
bthirion, line 395: Why do you lower the threshold ? I think it's more reasonable to have something around p<.001 unc.

bthirion (commented):
Thx. LGTM overall but we need to see the generated doc.

Remi-Gau:
will get back to this one when we have transparency plot enabled: this would allow us to keep the "strict" threshold at 3 but allow some transparency between 1.96 and 3

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -189,27 +213,34 @@ def make_localizer_contrasts(design_matrix):
...
+        ax = plt.subplot(2, 2, 1 + index)
         z_map = first_level_model.compute_contrast(
             contrast_val, output_type="z_score"
         )
-        plotting.plot_stat_map(
+        plot_stat_map(
             z_map,
             display_mode="z",
-            threshold=3.0,
+            threshold=1.96,
```
bthirion, line 242: Still wondering why you want to change this value.

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -189,38 +215,55 @@ def make_localizer_contrasts(design_matrix):
...
+        results = first_level_model.compute_contrast(
+            contrast_val, output_type="all"
         )
-        plotting.plot_stat_map(
-            z_map,
+        plot_stat_map(
+            results["stat"],
             display_mode="z",
-            threshold=3.0,
+            threshold=1,
```
bthirion, line 252: Why do you change the threshold to 1 ?

examples/04_glm_first_level/plot_first_level_details.py
```diff
@@ -336,10 +379,16 @@ def plot_contrast(first_level_model):
 plot_contrast_matrix(contrast_val, design_matrix)
 
 z_map = first_level_model.compute_contrast(contrast_val, output_type="z_score")
-plotting.plot_stat_map(
-    z_map, display_mode="z", threshold=3.0, title="effect of time derivatives"
+plot_stat_map(
+    z_map,
+    display_mode="z",
+    threshold=1.96,
```
bthirion, line 385: I'd rather not change the threshold.

examples/04_glm_first_level/plot_first_level_details.py
Remi-Gau: because these plots are now on the stats estimates not Z maps, so the rule of thumbs of 1.96 and 3 don't apply

the transparency uses the z maps

examples/04_glm_first_level/plot_first_level_details.py
Remi-Gau: except that doing model selection based on what is significant will most likely lead to overfitting

bthirion: Fair enough, but here we're not doing feature selection. We're simply looking at it.

examples/04_glm_first_level/plot_first_level_details.py
bthirion: OK, but then make it explicit. Otherwise, people will be surprised.

Issue #4218, opened 2024-01-16
Fetcher description improvements


Remi-Gau:
- [ ] make sure all dataset fetchers return a Bunch that include a description
- [ ] centralize dataset / atlas description contained to the description RST file:
  - in doc string (if there is) of the fetcher 
  - in doc strings of the examples that use the fetcher
- [ ] print the description of the dataset in the examples that use the fetcher
- [ ] render the description RST files as part of the doc
- [ ] make sure that articles mentioned in the descriptions are part of our bibliography

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/4215#issuecomment-1889100873_

PR #5142, opened 2025-02-11
[FIX] return design matrices of FIAC as dataframes

### LINKED ISSUE #2400: Add names for FIAC design matrices
Current report has no sense.
The best way is to embed the provided design matrices into a dataframe with proper column names.

Remi-Gau:
- Closes #2400
- Relates to #4218

- improve doc FIAC dataset
- return design matrices as dataframes with proper column names
- update examples that use this dataset

### DIFF
--- a/doc/visual_testing/reporter_visual_inspection_suite.py
+++ b/doc/visual_testing/reporter_visual_inspection_suite.py
@@ -238,8 +238,7 @@ def report_flm_fiac(build_type):
 
     mean_img_ = mean_img(fmri_img[0], copy_header=True)
 
-    design_files = [data["design_matrix1"], data["design_matrix2"]]
-    design_matrices = [pd.DataFrame(np.load(df)["X"]) for df in design_files]
+    design_matrices = [data["design_matrix1"], data["design_matrix2"]]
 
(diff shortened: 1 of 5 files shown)


bthirion (approved):
The changes LGTM (with the build problem pending).
One point is that I'd be happy to drop this FIAC example that adds another dataset for little benefit (I think that we introduced it because it's a two-run dataset but we could get that from another dataset).

Remi-Gau:
@pbarbarant
the dataset that you were planning to add for surface analysis: does it have several runs?

man-shu (approved):
LGTM!

Remi-Gau:
> The changes LGTM (with the build problem pending). One point is that I'd be happy to drop this FIAC example that adds another dataset for little benefit (I think that we introduced it because it's a two-run dataset but we could get that from another dataset).

@bthirion 
note that the spm multimodal dataset has 2 runs

bthirion:
I'd be in favor of dropping FIAC at some point. Maybe there's an issue with SPM miltimodal (e.g. not normalized), we need to check.

pbarbarant:
> @pbarbarant the dataset that you were planning to add for surface analysis: does it have several runs?

Yes it has 6 runs.

PR #5133, opened 2025-02-07
[ENH] extend type checking at run time

Remi-Gau:
- Closes none

Changes proposed in this pull request:

- check type of more parameters defined in _utils.doc

PR #5146, opened 2025-02-11
[FIX] Update parameters to comply with run time checks

Remi-Gau:
- Closes none but should fix some doc build fail on main

- update some parameters to make sure they are the proper type


bthirion:
Can you elaborate on the logic here ? I don't get it.

Remi-Gau:
> Can you elaborate on the logic here ? I don't get it.

This PR https://github.com/nilearn/nilearn/pull/5133 enforced that `standardize_confounds` must be a bool as per our doc string.

In our example we had a string which evaluated as truthy and was fine until we started actually checking its type.

Remi-Gau:
I have requested a full doc build to make sure we have not some other issues in our examples.

man-shu (approved):
LGTM!

PR #5148, opened 2025-02-11
[ENH] Refactor plotting

hndgzkn:
- Closes None

- Refactor `nilearn.plotting` without changing package interface to better separate concerns in subpackages.

> All plot functions can normally be imported from `nilearn.plotting`.
If plotting functions are imported as `nilearn.plotting.img_plotting`, it is no more possible. If this is a concern, it can be reverted.
----

The final objective is to create separate backends for `matplotlib` and `plotly`.


hndgzkn:
@Remi-Gau I've made a first push. I'd like to have your comments on the directory structure created. I'm not sure if we try to keep a certain level of depth in nilearn. What I do creates the structure:

- nilearn.plotting.image
- nilearn.plotting.surface
- nilearn.plotting.matrix
etc.

Is this convenient?

Remi-Gau:
I think in general I'd be ok with this but I'd prefer to talk about this at the next coredev meeting to make sure the rest of the team is onboard.

couple of things

- maybe instead of `image` we could have `volume`, because SurfaceImage are images too

```
nilearn.plotting.volume
nilearn.plotting.surface
```

- we may need to have redirect / aliases functions with deprecation warnings to warn users who do

```
from nilearn.plotting.img_plotting import plot_anat
```

and not

```
from nilearn.plotting import plot_anat
```

otherwise they will get some import errors

Remi-Gau:
@hndgzkn do we want to close this one then?

hndgzkn:
> @hndgzkn do we want to close this one then?

Yes, I'm closing this now. I'll open a new one.

PR #5165, opened 2025-02-18
[MAINT] set license for documentation

Remi-Gau:
- Closes none

- mention that license applies to code and doc (+ examples)
- remove unused files from doc


bthirion (approved):

emdupre:
It looks like this notice lands in the footer ; is there a reason to describe it there ?

Comparing with MNE or sklearn, it doesn't look like either of them take this approach. I'm not innately opposed (!), but just not sure that I immediately see the benefit beyond having it e.g. on our landing page, as sklearn does

Remi-Gau:
sklearn has it on the landing page but...

from a random page: https://scikit-learn.org/stable/auto_examples/cluster/plot_kmeans_digits.html#sphx-glr-auto-examples-cluster-plot-kmeans-digits-py

license mentioned in the footer

image

For MNE only found it in the governance.

I think that it could also be mentioned in the landing page, but people may skip the landing page so the footer ensures it is on every page people visit.

Remi-Gau:
For context, we were discussing where to mention licenses on another repo (see here) and realized that we did not mention the license for the doc and the examples, so this was related to this.

PR #5169, opened 2025-02-20
[FIX] Assign most frequent value in the neighborhood for atlases via `vol_to_surf` with ``method='nearest_most_frequent'``

### LINKED ISSUE #5168: [BUG] _nearest_voxel_sampling does not truly perform nearest voxel interpolation

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 22.04

### Python version

- [ ] 3.13
- [ ] 3.12
- [ ] 3.11
- [x] 3.10
- [ ] 3.9

### nilearn version

dev version

### Expected behavior

Values on the mesh should onl be a subset of values in the image

Interpolation occurs.
Actually, what the code does is the recombination of nearest-voxel values.

```python
import numpy as np
from nilearn.image import load_img
from nilearn import datasets
from nilearn.plotting import view_surf

# Load a volumetric atlas (example: Harvard-Oxford cortical atlas)
atlas = datasets.fetch_atlas_harvard_oxford('cort-maxprob-thr25-1mm')
atlas_img = load_img(atlas.maps)

# sanity check print number of unique labels
atlas_img_data = atlas_img.get_fdata()
uniques,counts = np.unique(atlas_img_data,return_counts=True)
print(f"Volumetric atlas has {len(counts)} unique ids")

# Load the fsaverage mesh(s)
fsaverage_meshes = datasets.load_fsaverage()

# Question: How to convert volumetric to surface with only valid atlas
# region ids? 
atlas_img_surf = SurfaceImage.from_volume(
    mesh=fsaverage_meshes["pial"],
    volume_img=atlas_img,
    interpolation='nearest'
)

# check if it worked
for hemi, data in atlas_img_surf.data.parts.items():
    uniques,counts = np.unique(data,return_counts=True)
    print(f"{hemi} hemi of surface atlas has {len(counts)} unique ids")


```

Remi-Gau:
- Closes #5168

- add regression test
-


Remi-Gau:
@bthirion @man-shu 
any of you has the bandwidth to look into this: I only added the regression test but I'd like to close the other PRs I have first

man-shu:
> any of you has the bandwidth to look into this

Yes will do!

nilearn/surface/tests/test_surface.py
```diff
@@ -594,21 +599,36 @@ def test_vol_to_surf(kind, n_scans, use_mask):
...
+def test_vol_to_surf_labels(img_labels, single_mesh):
+    """Test nearest interpolation does no 'averaging'."""
+    img_labels_data = img_labels.get_fdata()
+    uniques_vol = np.unique(img_labels_data)
+
+    mesh_labels = vol_to_surf(img_labels, single_mesh, interpolation="nearest")
+
+    uniques_surf = np.unique(mesh_labels)
+
+    assert uniques_vol == uniques_surf
```
Remi-Gau, line 631:
```suggestion
    assert uniques_vol =< uniques_surf
```

man-shu: hmm that doesn't seem to work

man-shu:
Ok so I ended up picking the most frequent voxel value when the input volume is an atlas (i.e. has all integer values). Here's a comparison.

The "after, default n_samples" is what it would look like after this change:

Screenshot 2025-02-25 at 10 53 00 AM

man-shu:
This is ready for review @Remi-Gau @bthirion @JohannesWiesner

Remi-Gau:
not checked the rest but you can already update the changelog 😉

nilearn/surface/tests/test_surface.py
```diff
@@ -594,21 +599,38 @@ def test_vol_to_surf(kind, n_scans, use_mask):
...
     other_proj = vol_to_surf(img, center_mesh, kind=kind, mask_img=mask_img)
+
     correlation = pearsonr(proj.ravel(), other_proj.ravel())[0]
+
     assert correlation > 0.99
-    with pytest.raises(ValueError, match=".*interpolation.*"):
-        vol_to_surf(img, mesh, interpolation="bad")
+
+
+def test_vol_to_surf_labels(img_labels, single_mesh):
```
bthirion, line 622: This test fails with current main branch ?

bthirion (commented):
LGTM overall.

nilearn/surface/tests/test_surface.py
man-shu: Will check!

man-shu: Ok, it returns the same output on the main as on this branch. 
The projected surface only has the background label, which makes me think that our fixtures are not suitable for the test.

man-shu: > it returns the same output on the main as on this branch.

This wasn't true because I was not using the implementation on main and the test actually failed when I did.

But the output on this branch was still just the background label. So I am using a better mesh now.

Remi-Gau:
yup makes sense to me : as discussed in the open hours, stat maps of integers should be a corner case for this one.

man-shu:
> as discussed in the open hours, stat maps of integers should be a corner case for this one.

While discussing, we also realized that someone could even give an atlas, which is supposed to have all ints, but those ints could still be encoded as floats (eg. 1 as 1.0). And this one doesn't seem like much of an edge case.

One solution could be to allow the user to specify another new `interpolation` method, say, `nearest_most_frequent`. We could also mention that this one is more suitable for **deterministic** atlases in the docstring. WDYT @bthirion ?

(Also noticed that for the Harvard-Oxford there is at least one _probabilistic_ atlas (cort-prob-2mm) that has all integers. So in that case, the current implementation would use the most-frequent method instead of sticking to the mean.)

nilearn/surface/surface.py
```diff
@@ -479,23 +492,52 @@ def _nearest_voxel_sampling(
...
-    texture = proj.dot(data)
-    # if all samples around a mesh vertex are outside the image,
-    # there is no reasonable value to assign to this vertex.
-    # in this case we return NaN for this vertex.
-    texture[np.asarray(proj.sum(axis=1) == 0).ravel()] = np.nan
+    data = np.asarray(images)
+    # if the data is all integers we know it's an atlas, so the texture
+    # value for each vertex would be the most frequent voxel value in the
+    # neighborhood (out of the n_points samples)
+    if np.issubdtype(data.dtype, np.integer):
```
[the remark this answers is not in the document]
man-shu: So the original issue was that vol_to_surf gave weird results when we projected a volumetric _deterministic_ atlas to surface.

Therefore I want to do this whenever the input volume image is a deterministic atlas. Here my assumption was that it would always have a dtype int. Now that assumption feels flawed.

So I want to introduce another new `interpolation` method, say, `"nearest_most_frequent"`, which is more suitable for deterministic atlases. All other cases (probabilistic atlases and stat maps) are already being handled.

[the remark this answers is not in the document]
man-shu: We already have the nearest option (the one that averages all the nearest voxels). That one is more suitable for stat maps and prob atlases.

Given that there is no other reliable way to differentiate deterministic atlases vs. stat maps/prob atlases, I want to allow users to be able to select the method depending upon the kind of volumetric image they are inputting.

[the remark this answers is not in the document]
bthirion: I could be wrong, but my understanding is that the `nearest` option as it is currently, is not consistent. Is there a motivation to maintain it ?
Maybe for bw compatibility we should keep it, but through a warning, saying that it will be changed in the future.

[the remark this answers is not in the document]
man-shu: I think the current `nearest` method still makes sense for **statistical maps and probabilistic atlases**. The reason is that there won't be exactly the same repeated (decimal) voxel values in the neighborhood of a voxel. So, averaging across those values seems like the correct choice.

The problem only arises with deterministic atlases where, for example, the boundary of regions 2 and 3 can't be 2.5. So picking the most frequent voxel value in the neighborhood, in this case, works well (as in [this figure](https://github.com/nilearn/nilearn/pull/5169#issuecomment-2681390102))

nilearn/surface/surface.py
```diff
@@ -460,6 +460,60 @@ def _projection_matrix(
...
+    masks = np.split(masks, mesh_n_vertices)
+    # mask sample locations and make a list of masked indices because
+    # masked locations are not contiguous
+    masked_sample_locations = [
+        sample_locations[idx][~mask] for idx, mask in enumerate(masks)
+    ]
+    return masked_sample_locations
+
+
+def _nearest_most_frequent(
```
man-shu, line 476: So now I have separated the nearest most frequent as a new interpolation method.

I also noticed that, right now, we only use vol_to_surf to project stat maps in our examples.

So it would be nice to have one where we demonstrate this for atlases (and also compare the different methods for different use cases). In a separate PR though. LMWYT @Remi-Gau @bthirion

nilearn/surface/surface.py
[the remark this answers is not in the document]
bthirion: But for statistical maps or probabilistic atlases it makes more sense to use linear interpolation.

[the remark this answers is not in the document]
man-shu: Ok, yeah, I did not consider the "linear" interpolation. 

The doc string says "nearest" is supposed to be faster than "linear". 

Screenshot 2025-03-07 at 11 28 04 AM

But now that we have clearly different use cases for the two, and if "nearest" can't fulfill the use case it's supposed to, then doing it faster does not matter. So I agree to deprecate it.

However, looking into the speeds, I see that my implementation of "nearest_most_frequent" is pretty slow. Will try to improve it.

nilearn/surface/surface.py
```diff
@@ -460,6 +460,60 @@ def _projection_matrix(
...
+        n_points=n_points,
+        inner_mesh=inner_mesh,
+        depth=depth,
+    )
+    sample_locations = _mask_sample_locations(
+        sample_locations, images[0].shape, mesh.n_vertices, mask
+    )
+    texture = np.zeros((mesh.n_vertices, images.shape[0]))
+    for img in range(images.shape[0]):
+        for loc in range(len(sample_locations)):
```
man-shu, line 507: So I was trying to get around these for loops using numpy indexing to improve speed, but turns out that after the masking step in L502, we won't necessarily have the same number of sampling locations for each vertex. Meaning we need to stick to lists. Any other suggestions are welcome though

man-shu:
updated doc string: https://output.circle-artifacts.com/output/job/e732fc04-5efc-4776-930a-37bd3d329550/artifacts/0/dev/modules/generated/nilearn.surface.vol_to_surf.html

bthirion (approved):
LGTM, thx.

nilearn/surface/surface.py
```diff
@@ -582,14 +636,23 @@ def vol_to_surf(
...
             Use the intensity of the nearest voxel.
 
+        .. versionadded:: 0.11.2.dev
+
+        - 'nearest_most_frequent':
+            Use the most frequent value in the neighborhood (out of the
+            `n_samples` samples) instead of the mean value. This is useful
+            when the image is a
+            :term:`deterministic atlas<Deterministic atlas>`.
+
```
Remi-Gau, line 655:
```suggestion
        - 'nearest_most_frequent':
            Use the most frequent value in the neighborhood (out of the
            `n_samples` samples) instead of the mean value. This is useful
            when the image is a
            :term:`deterministic atlas<Deterministic atlas>`.
            
            .. versionadded:: 0.11.2.dev

```

Remi-Gau: I think you want the directive after the new option.

nilearn/surface/surface.py
```diff
@@ -582,14 +636,23 @@ def vol_to_surf(
...
         For one image, the speed difference is small, 'linear' takes about x1.5
         more time. For many images, 'nearest' scales much better, up to x20
         faster.
```
Remi-Gau, line 658: a bit a above the there is list that renders as "quote" block: 

image

can probably be fixed here or in another PR

nilearn/surface/surface.py
man-shu: Ok yes thanks!

nilearn/surface/surface.py
man-shu: Will do now!

man-shu:
@Remi-Gau, should I approve this one on your behalf so that I can merge it? 🤣

man-shu (approved):

PR #5176, opened 2025-02-24
[REF] apply sourcery refactoring

Remi-Gau:
- Closes none

- apply some automatic refactoring suggested by sourcery


nilearn/decoding/decoder.py
```diff
@@ -119,14 +119,14 @@ def _check_param_grid(estimator, X, y, param_grid=None):
     if param_grid is None:
         param_grid = _default_param_grid(estimator, X, y)
 
-    elif isinstance(estimator, (RidgeCV, RidgeClassifierCV)):
+    elif isinstance(estimator, (RidgeCV, RidgeClassifierCV)) or (
+        not isinstance(estimator, LogisticRegressionCV)
+        and isinstance(estimator, LassoCV)
+    ):
```
Remi-Gau, line 125: if we agree that those different options are mutually exclusive this can even be refactored even further

bthirion (commented):
LGTM so far.

.sourcery.yaml
man-shu, line 1: So would this run be run automatically on each commit, or is it something we can run and get suggestions if we want?

nilearn/decoding/decoder.py
man-shu: I think this would try to set the `param_grid` for "alphas" with estimators that were not in the previous if.. elif cases.

I think this would happen for LinearSVC and SVR. For these estimators this would probably fail because they do not have a parameter alpha.

man-shu: However all the tests are passing, so either I am mistaken or this isn't getting caught in our tests

.sourcery.yaml
Remi-Gau: The latter: this is a slow thing to run so it would be a pain to run on every commit.

There is a service from sourcery that can auto-review your PRs and make refactoring suggestions: see examples here on my fork

- https://github.com/Remi-Gau/nilearn/pull/37
- https://github.com/Remi-Gau/nilearn/pull/71

I like to have them on my fork, but I don't think we want to add them to the upstream repo.

We could think of setting an action to run it once a month or something to clean up refactoring we may have missed.

nilearn/decoding/decoder.py
Remi-Gau: OK I will investigate and see if something is missing in our tests or if we should have an else --> error to catch unknown cases

.sourcery.yaml
Remi-Gau: yeah it uses the fact that you can have mermaid diagram in posts on github.

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams

but the PR reviews of sourcery are AI generated and hence wildly variable in quality

nilearn/decoding/decoder.py
```diff
@@ -119,12 +119,11 @@ def _check_param_grid(estimator, X, y, param_grid=None):
...
 
-    elif isinstance(estimator, (RidgeCV, RidgeClassifierCV)):
-        param_grid = _wrap_param_grid(param_grid, "alphas")
     elif isinstance(estimator, LogisticRegressionCV):
         param_grid = _replace_param_grid_key(param_grid, "C", "Cs")
         param_grid = _wrap_param_grid(param_grid, "Cs")
-    elif isinstance(estimator, LassoCV):
+
+    elif isinstance(estimator, (RidgeCV, RidgeClassifierCV, LassoCV)):
         param_grid = _wrap_param_grid(param_grid, "alphas")
```
Remi-Gau, line 127: unless I am missing something, this can be refactored that way

man-shu: Yes agreed!

bthirion (commented):
LGTM

PR #5179, opened 2025-02-24
[FIX] prevent range overrun

### LINKED ISSUE #5178: [BUG] Range overrun

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Linux Ubuntu 22.04

### Python version

- [ ] 3.13
- [ ] 3.12
- [x] 3.11
- [ ] 3.10
- [ ] 3.9

### nilearn version

0.11.1

### Expected behavior

In `nilearn/mass_univariate/_utils.py`, I expect  `score_threshs = np.arange(step, max_score + step, step)` to produce values up through `max_score` (modulo a small amount of floating-point rounding), which is what it usually does.

For example, `np.arange(1/49, 1 + 1/49, 1/49)` also produces a number very close to 50/49, which is a number that exceeds 1 by significantly more than floating-point round-off error.

```python
step = 1/49
max_score = 1
max(np.arange(step, max_score + step, step))
```

Leengit:
- Closes #5178 

- Changed the floating-point upper bound value of an `np.arange` call so that the range terminates with the correct value.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,6 +13,10 @@ HIGHLIGHTS
 Fixes
 -----
 
+- :bdg-dark:`Code` Prevent range overrun in TFCE (:gh:`5179` by `Lee Newberg`_).
+
+- :bdg-dark:`Code` Bound the number of thresholds in TFCE calculation (:gh:`5179` by `Lee Newberg`_).
+
 - :bdg-dark:`Code` Ensure that all figures are generated with a color bar by default (:gh:`5172` by `Rémi Gau`_).
 
 - :bdg-dark:`Code` Fixes datasets that returned numpy arrays instead of pandas dataframes (:gh:`5109` by `Rémi Gau`_).
--- a/nilearn/mass_univariate/_utils.py
+++ b/nilearn/mass_univariate/_utils.py
@@ -69,6 +69,7 @@ def calculate_tfce(
         arr3d = arr4d[..., i_regressor]
 
         signs = [-1, 1] if two_sided_test else [1]
+        score_threshs = _return_score_threshs(arr3d, dh, two_sided_test)
 
         # If we apply the sign first...
         for sign in signs:
@@ -78,7 +79,6 @@ def calculate_tfce(
             # is incrementally larger
             temp_arr3d = arr3d * sign
 
-            score_threshs = _return_score_threshs(arr3d, dh, two_sided_test)
             # Prep step
             for score_thresh in score_threshs:
                 temp_arr3d[temp_arr3d < score_thresh] = 0
@@ -135,7 +135,9 @@ def _return_score_threshs(arr3d, dh, two_sided_test):
     max_score = (
         np.nanmax(np.abs(arr3d)) if two_sided_test else np.nanmax(arr3d)
     )
-    number_steps = 100 if dh == "auto" else max(1, round(max_score / dh))
+    number_steps = (
+        100 if dh == "auto" else np.clip(round(max_score / dh), 10, 1000)
+    )
     return np.linspace(0, max_score, number_steps + 1)[1:]
 
 
--- a/nilearn/mass_univariate/_utils.py
+++ b/nilearn/mass_univariate/_utils.py
@@ -76,10 +76,9 @@ def calculate_tfce(
             signs = [1]
             max_score = np.max(arr3d)
 
-        step = max_score / 100 if dh == "auto" else dh
-
         # Set based on determined step size
-        score_threshs = np.arange(step, max_score + step, step)
+        number_steps = 100 if dh == "auto" else max(1, round(max_score / dh))
+        score_threshs = np.linspace(0, max_score, number_steps + 1)[1:]
 
         # If we apply the sign first...
         for sign in signs:
--- a/nilearn/mass_univariate/_utils.py
+++ b/nilearn/mass_univariate/_utils.py
@@ -68,17 +68,7 @@ def calculate_tfce(
     for i_regressor in range(arr4d.shape[3]):
         arr3d = arr4d[..., i_regressor]
 
-        # Get signs / threshs
-        if two_sided_test:
-            signs = [-1, 1]
-            max_score = np.max(np.abs(arr3d))
-        else:
-            signs = [1]
-            max_score = np.max(arr3d)
-
-        # Set based on determined step size
-        number_steps = 100 if dh == "auto" else max(1, round(max_score / dh))
-        score_threshs = np.linspace(0, max_score, number_steps + 1)[1:]
+        signs = [-1, 1] if two_sided_test else [1]
 
         # If we apply the sign first...
         for sign in signs:
@@ -88,6 +78,7 @@ def calculate_tfce(
             # is incrementally larger
             temp_arr3d = arr3d * sign
 
+            score_threshs = _return_score_threshs(arr3d, dh, two_sided_test)
             # Prep step
             for score_thresh in score_threshs:
                 temp_arr3d[temp_arr3d < score_thresh] = 0
@@ -137,6 +128,15 @@ def calculate_tfce(
                 )
 
     return tfce_4d
+
+
+def _return_score_threshs(arr3d, dh, two_sided_test):
+    """Compute list of score threshold to use for TFCE."""
+    max_score = (
+        np.nanmax(np.abs(arr3d)) if two_sided_test else np.nanmax(arr3d)
+    )
+    number_steps = 100 if dh == "auto" else max(1, round(max_score / dh))
+    return np.linspace(0, max_score, number_steps + 1)[1:]
 
 
 def null_to_p(test_values, null_array, alternative="two-sided"):
--- a/nilearn/mass_univariate/_utils.py
+++ b/nilearn/mass_univariate/_utils.py
@@ -1,5 +1,7 @@
 """Utility functions for the permuted least squares method."""
 
+from warnings import warn
+
 import numpy as np
 from scipy import linalg
 from scipy.ndimage import label
@@ -135,9 +137,23 @@ def _return_score_threshs(arr3d, dh, two_sided_test):
     max_score = (
         np.nanmax(np.abs(arr3d)) if two_sided_test else np.nanmax(arr3d)
     )
-    number_steps = (
-        100 if dh == "auto" else np.clip(round(max_score / dh), 10, 1000)
-    )
+
+    number_steps = 100 if dh == "auto" else round(max_score / dh)
+    if number_steps < 10:
+        warn(
+            f"Not enough steps for TFCE. Got: {number_steps=}. "
+            "Setting it to 10.",
+            stacklevel=4,
+        )
+        number_steps = 10
+    if number_steps > 1000:
+        warn(
+            f"Too many steps for TFCE. Got: {number_steps=}. "
+            "Setting it to 1000.",
+            stacklevel=4,
+        )
+        number_steps = 1000
+
     return np.linspace(0, max_score, number_steps + 1)[1:]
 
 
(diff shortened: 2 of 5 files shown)


Remi-Gau:
we probably want a test for this to avoid regression in the future

my suggestion would :

- extract this part of the code into a separate function:
https://github.com/nilearn/nilearn/blob/b06150ca01a0b8729f01063b0a2431afbdf91487/nilearn/mass_univariate/_utils.py#L71-L82

- create a test in `nilearn/mass_univariate/tests/test_utils.py` to demonstrate the bug before the fix

regarding implementation, I think I'd rather go with removing any values from score_threshs that is > to max_score rather than tweaking the `stop` of np.arrange: it makes more clear when reading the code what the intention is.

Remi-Gau:
Probably worth mentioning this bug fix in the changelog and also, unless I am mistaken, this is your first contribution to nilearn 🚀, so add yourself to the list of contributors in citation.cff.

More info on this:
- https://nilearn.github.io/dev/development.html#changelog

Leengit:
> regarding implementation, I think I'd rather go with removing any values from score_threshs that is > to max_score rather than tweaking the `stop` of np.arrange: it makes more clear when reading the code what the intention is.

I it my belief that we want to include as a final value in the range the value that would be exactly equal to `max_score` in the absence of floating-point round-off error.  In particular, a value of `max_score * (1 + 1e-12)` should be kept; it should _not_ be discarded simply because it exceeds `max_score`.  I think using the increment of `0.5 * step` achieves this nicely.  Possibly we want a follow up statement to make that last value exact:
```python
score_threshs = np.arange(step, max_score + 0.5 * step, step) 
score_threshs[-1] = max_score
```

bthirion:
I'm afraid that the integration requires the interval between 2 consecutive values to be identical (there is in theory a multiplication by dh, but it is ommited in the implementation as discussed in https://github.com/nilearn/nilearn/blob/main/nilearn/mass_univariate/_utils.py#L122).

Leengit:
> I'm afraid that the integration requires the interval between 2 consecutive values to be identical 

I just pushed an implementation that uses `score_threshs = np.linspace(0, max_score, number_steps + 1)[1:]`.   When `dh == "auto"` then `number_steps = 100`.  When `dh != "auto"` then `number_steps  = max(1, round(max_score / dh))`.

nilearn/mass_univariate/_utils.py
```diff
@@ -76,10 +76,9 @@ def calculate_tfce(
             signs = [1]
             max_score = np.max(arr3d)
 
-        step = max_score / 100 if dh == "auto" else dh
-
         # Set based on determined step size
-        score_threshs = np.arange(step, max_score + step, step)
+        number_steps = 100 if dh == "auto" else max(1, round(max_score / dh))
+        score_threshs = np.linspace(0, max_score, number_steps + 1)[1:]
```
Remi-Gau, line 81: what I am reading makes sense to me
let's wait for CI to finish

do you want me to add the test?

Leengit: > do you want me to add the test?

Yes, please.

nilearn/mass_univariate/tests/test_utils.py
```diff
@@ -268,3 +269,16 @@ def test_t_score_with_covars_and_normalized_design_withcovar(rng):
...
+@pytest.mark.parametrize("dh", [1 / 49, "auto", 0.1])
+@pytest.mark.parametrize("arr3d", [np.ones((10, 11)), _rng().random((10, 11))])
+def test_return_score_threshs(arr3d, two_sided_test, dh):
+    """Check that the range of thresholds to test."""
+    score_threshs = _utils._return_score_threshs(
+        arr3d, dh=dh, two_sided_test=two_sided_test
+    )
+
+    max_score = np.max(np.abs(arr3d)) if two_sided_test else np.max(arr3d)
+    assert np.sum(score_threshs > max_score) == 0
```
Remi-Gau, line 284: I checked and the previous implementation failed this test

Leengit: It looks good to me, thank you.

Leengit:
It could be my user error, but I am running into trouble in the same code (whether with this pull request or not), where the computation of `max_score` (regardless of whether it is `np.max(arr3d)` or `np.max(abs(arr3d))`) is giving a `np.nan` because some of the array entries are `np.nan`.  Would it be appropriate to ignore those bad entries in `arr3d` by changing `np.max` to `np.nanmax`?  (I realize that this is mostly off-topic for the present pull request, but since I have your attention....)

Remi-Gau:
> It could be my user error, but I am running into trouble in the same code (whether with this pull request or not), where the computation of `max_score` (regardless of whether it is `np.max(arr3d)` or `np.max(abs(arr3d))`) is giving a `np.nan` because some of the array entries are `np.nan`. Would it be appropriate to ignore those bad entries in `arr3d` by changing `np.max` to `np.nanmax`? (I realize that this is mostly off-topic for the present pull request, but since I have your attention....)

let me add some tests in here

Remi-Gau:
also realizing that we do not ensure a minimum number of steps to cover in the range

if maxscore is 1 and users give a dh of 0.9, this gives a single 'step' which I doubt is appropriate

nilearn/mass_univariate/_utils.py
```diff
@@ -140,6 +130,15 @@ def calculate_tfce(
     return tfce_4d
 
 
+def _return_score_threshs(arr3d, dh, two_sided_test):
+    """Compute list of score threshold to use for TFCE."""
+    max_score = (
+        np.nanmax(np.abs(arr3d)) if two_sided_test else np.nanmax(arr3d)
+    )
+    number_steps = 100 if dh == "auto" else max(1, round(max_score / dh))
```
Remi-Gau, line 138: should probably set a minimum number of steps, no?

Leengit: I don't know.  If the user has bothered to set `dh` to other than `"auto"`, maybe they actually want that value... as part of plot across multiple values, for testing purposes, ...?

Remi-Gau: I think that in general it is better to make each step more robust without making too many assumoptions about the previous ones.

Maybe nilearn will not give you data with nans in them but some other software may (I know that SPM gives outputs 'beta images' padded with nans for examples), so I'd say it makes sense to do this.

Remi-Gau: @bthirion 
any opinion on this?

bthirion: I think it's a good idea to use nanmax

Remi-Gau: and any suggestion if we should a minimum number of steps for TFCE procedure?

nilearn/mass_univariate/_utils.py
```diff
@@ -89,6 +78,7 @@ def calculate_tfce(
             # is incrementally larger
             temp_arr3d = arr3d * sign
 
+            score_threshs = _return_score_threshs(arr3d, dh, two_sided_test)
```
Leengit, line 81: Let's move the setting of `score_threshs` to just before the start of the `for sign in signs:` loop.  There's no need to call it twice, and potentially a bug could creep in so that the two calls are given different arguments and produce different values.

Remi-Gau: Agreed this can be hoisted out of the loop.

Remi-Gau:
- OK the test failure is unrelated and should be fixed by #5191

will mark as blocked until #5191 is merged

nilearn/mass_univariate/_utils.py
bthirion: I would impose at least 10.

bthirion (commented):
LGTM 
Ideally we would like to have a test hat checks that the computed TFCE is numerically  correct...

Remi-Gau:
> Ideally we would like to have a test hat checks that the computed TFCE is numerically correct...

agreed but I would say this is beyond the scope of this PR

nilearn/mass_univariate/_utils.py
Remi-Gau: @Leengit do you want to add a check that we got at least 10 steps and throw a warning that the number of steps will be set to 10 when it's not the case?

nilearn/mass_univariate/_utils.py
```diff
@@ -140,6 +130,15 @@ def calculate_tfce(
     return tfce_4d
 
 
+def _return_score_threshs(arr3d, dh, two_sided_test):
+    """Compute list of score threshold to use for TFCE."""
+    max_score = (
+        np.nanmax(np.abs(arr3d)) if two_sided_test else np.nanmax(arr3d)
+    )
+    number_steps = 100 if dh == "auto" else max(10, round(max_score / dh))
```
Remi-Gau, line 138:
```suggestion
    number_steps = 100 if dh == "auto" else max(10, round(max_score / dh))
    if number_steps < 10:
       warns(f'Not enough steps for TFCE. Got: {number_steps=}. '
       'Setting it to 10.', stacklevel=3
       )
       number_steps = 10 
```

nilearn/mass_univariate/_utils.py
Remi-Gau: throwing a warning could be done a bit like this: https://github.com/nilearn/nilearn/pull/5179/files#r1983395313

but indeed we would need to add a test that ensure that the warning is triggered: 

example of checking if a warning is thrown when running a test:
https://github.com/nilearn/nilearn/blob/8517db819c43c75f7401e18c60b5108749d12b33/nilearn/tests/test_masking.py#L267-L268

nilearn/mass_univariate/_utils.py
Leengit: Thank you.  Q: we have `max(10, round(max_score / dh))` in there so the warning will never be issued with the code the way it is.  If we change that to simply `round(max_score / dh)` then the warning will be issued when `round(max_score / dh) < 10`.  Despite that these two possible implementations have different behaviors with the warning, with both we proceed with `number_steps = 10` whenever `round(max_score / dh)` is too small, so that's good.  Do we want to trigger the warning when `round(max_score / dh) < 10` by changing `max(10, round(max_score / dh))` to `round(max_score / dh)`?

At the risk of feature creep ... do we also want to warn and do something sensible when `number_steps` is too large?  How about when `np.isnan(number_steps)`?

Leengit: I just pushed a commit that checks that `number_steps` does not exceed 1000.  The code warns when the value has to be coerced to be within the range of 10 to 1000.

bthirion (commented):
LGTM.

nilearn/mass_univariate/_utils.py
Remi-Gau: > CI informs me that `warns` is not defined, and exploration indicates that it is for pytest code rather than somewhere like `_utils.py`. I have just pushed a commit that clips to the range of 10 to 1000, but does not issue a warning.

ah yeah you probably need to to from `warnings import warn` to get it from the standard library

Remi-Gau:
will push a commit to add a warning

doc/changes/latest.rst
```diff
@@ -13,6 +13,8 @@ HIGHLIGHTS
 Fixes
 -----
 
+- :bdg-dark:`Code` Prevent range overrun in TFCE (:gh:`5179` by `Lee Newberg`_).
+
```
Remi-Gau, line 17: I wonder if the changelog should be updated further to mention that the TFCE procedure has now an upper and lower bound in the number of thresholds it will test.

Leengit: I just modified the first commit to update the changelog for that, and forced pushed.

Remi-Gau: yeah but that removed the last commit I added.

Remi-Gau: in general there is no need to rewrite older commits of a pull request because we will do a squash-merge of all the commit of the PR: so just add new changes and keep pushing

will get back my commit from the dead

Remi-Gau: no worries: I already repushed it
so if you do a git pull on your branch you should have it

bthirion (approved):
LGTM, thx.

Remi-Gau (approved):

Remi-Gau:
thanks @Leengit for working on this

PR #5185, opened 2025-02-26
[FIX] Load images once in parallel function instead of single function

### LINKED ISSUE #5150: [ENH] Check whether `load_img` loads images in a memory mapped way
I think we should deal with this separately, but I wanted to test our hypothesis that joblib + memory mapped in_memory image is what's causing the path vs. in_memory difference. 

So I compared the image loaded via `load_img` (I think it doesn't use dataobj) vs. `concat_imgs` (defo uses dataobj):

With `load_img` the difference between NiftiMasker in_memory, numpy masker path and numpy masker in_memory disappears: https://output.circle-artifacts.com/output/job/c99698a5-290c-4ee7-9750-1e551c4774fa/artifacts/0/dev/auto_examples/07_advanced/plot_mask_large_fmri.html#conclusion

_Originally posted by @man-shu in https://github.com/nilearn/nilearn/issues/5113#issuecomment-2654402188_

man-shu:
- Closes #5150

- load the image once in "parallel" functions
- "numpy path" variant is not relavant anymore
- proxy images/memory mapping isn't relavant because they have the same peak memory usage as array images:
  - probably because full array is needed for masking
  - lower memory usage should only be expected for use cases where segments of the array are needed

### DIFF
--- a/examples/07_advanced/plot_mask_large_fmri.py
+++ b/examples/07_advanced/plot_mask_large_fmri.py
@@ -30,23 +30,17 @@ the fMRI image in a different way:
 2. Masking the fMRI image's data array using numpy indexing.
 3. Using :class:`multiprocessing.shared_memory.SharedMemory`.
 
-For the first two methods, there are two ways to input the fMRI image:
+For the first method, there are two ways to input the fMRI image:
 
 1. passing the file path (i.e., the location of the large fMRI image on-disk).
-2. loading image using :mod:`nibabel` and passing this in-memory object.
-
-These two input methods can impact the memory usage of the task, so we will
-also compare these variants of the first two methods.
+2. loading image first via :func: and passing this in-memory object.
 
 When using file paths, the entire image is loaded into memory for each process,
 and that is exactly the problem we described earlier.
 
-However, when the fMRI image is loaded in a way that allows
-`numpy memory mapping <https://numpy.org/doc/stable/reference/generated/numpy.memmap.html>`_,
-as with :class:`nibabel.arrayproxy.ArrayProxy`,
-only the segments of data that are needed are read from the disk. Joblib
-can handle this efficiently across multiple processes and this results in
-lower memory usage. We will see that this can already be a significant
+However, when the fMRI image is loaded once and then passed to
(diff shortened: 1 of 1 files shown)


bthirion (approved):
LGTM, thx.

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -116,61 +110,73 @@
...
+def nifti_masker_single(fmri, mask):
+    return NiftiMasker(mask_img=mask).fit_transform(fmri)
+
+
+def nifti_masker_parallel_path(fmri_path, mask_paths):
+    Parallel(n_jobs=N_REGIONS)(
+        delayed(nifti_masker_single)(fmri_path, mask_path)
+        for mask_path in mask_paths
+    )
+    return 1
```
emdupre, line 153: Just to make sure I understand : Why are we adding these `return` statements ?

examples/07_advanced/plot_mask_large_fmri.py
```diff
@@ -320,27 +289,27 @@ def numpy_masker_shared_parallel(img, masks):
     [
         "NiftiMasker,\nwith path",
         "NiftiMasker,\nwith in-memory\nimage",
-        "Numpy indexing,\nwith path",
-        "Numpy indexing,\nwith in-memory\nimage",
+        "Numpy indexing",
         "Numpy indexing,\nwith shared\nmemory",
```
emdupre, line 293: Light suggestion to be clearer that we're referring to the class `SharedMemory` rather than the idea of shared memory
```suggestion
        "Numpy indexing,\nwith\nSharedMemory",
```

emdupre (approved):
Thanks, @man-shu ! This pattern of results makes a bit more sense to me now. I left two light comments / clarifications, but I don't think we need a full re-review for those.

examples/07_advanced/plot_mask_large_fmri.py
man-shu: Nothing specific, I just wanted to avoid returning the output of Parallel and later storing it in a variable because that would show up as an irrelavant memory peak.

PR #5188, opened 2025-02-27
[MAINT] track and plot tests duration

### LINKED ISSUE #5186: recent increase in runtime of the test suite
noticed this yesterday

Image

I suspect that the time increase may be related to the increase of plotting related tests

will try to find a convenient way of keeping track of test run time

Remi-Gau:
- Closes #5186

- use a pytest plugin to dump info about each test in a CSV
- generate histogram of test run time by: subpackage, module, test, parametrization
- upload those as artifacts in CI
- only run tests with plot comparison for the oldest matplotlib 
- remove duplicate tests, tests with too many parametrizations or that also ran some default values covered by other tests
- skip one test on mac-os that sometimes lasted > 1000 seconds
- set time out for tests:
  - test sessions should not last more than one hour
  - a single test should not last more than 45 seconds


Remi-Gau:
test timings for each run are available in the 'artifact' section : 

See for example:

https://github.com/nilearn/nilearn/actions/runs/13589402176?pr=5188

bthirion (approved):
I can't spot any issue. Thx !

Remi-Gau:
will merge with a single review because this affects only tests with not apparent change in coverage and an increase of testing performance

Remi-Gau:
ok this is actually already turning out to be useful as it is flagging some slow test on main

will open a PR to fix

PR #5198, opened 2025-03-06
[MAINT] up timeout of test

Remi-Gau:
Still some tests that sometimes fail in CI for being too slow, so I am upping the timeout to 60 seconds globally.


man-shu (approved):
LGTM!

PR #5202, opened 2025-03-07
[REF] refactor GLM reports

Remi-Gau:
- Closes https://github.com/nilearn/nilearn/issues/4368

- use same report layout for surface and GLM reports
- also make sure that all our reports (masker or glm) will use the same base layout
- start adding a "method section" to the GLM reports to be ideally copy-pastable into articles
- switch to all GLM report to use png instead of svg


nilearn/reporting/data/html/method_section.html
```diff
@@ -0,0 +1,25 @@
...
+  (full-width at half maximum={{ smoothing_fwhm }} mm).
+</p>
+{{endif}}
+<p>
+  Regressors were entered into run-specific design matrices and onsets were
+  convolved with {{ hrf_model }} canonical hemodynamic response function.
+</p>
+{{if contrasts_dict }}
+<p>Contrast were computed using a fixed-effect approach across runs.</p>
+{{endif}} {{endif}}
```
Remi-Gau, line 25: will need to be refined in a follow up PR

Remi-Gau:
struggling to reproduce the failing test locally...

man-shu:
> also it may be worth considering starting adding the functionality to write the report figures to disk rather than embedding gigangtic bytes into the html

But then the HTML wouldn't be standalone. For example, you wouldn't be able to just share one file with everything in it. You'd have to share a directory with all the appropriate figures, and HTML files.

I think using PNGs would decrease the size A LOT.

Remi-Gau:
> I think using PNGs would decrease the size A LOT.

cool will try that

Remi-Gau:
> But then the HTML wouldn't be standalone. For example, you wouldn't be able to just share one file with everything in it. You'd have to share a directory with all the appropriate figures, and HTML files.

correct but I was thinking of the case when we use `save_glm_to_bids` that dumps all your GLM results to disk in structured manner **including** a GLM report: in that case it would make sense.

Remi-Gau:
Note that when saving glm output to disk, it could maybe be a usecase for using niivue in the glm reports to have interactive viewing and minimize report size.

man-shu:
> it could maybe be a usecase for using niivue in the glm reports

Yeah, I think it would be good to compare the sizes of the plotly HTML vs. the niivue HTML.
We weren't completely sure if niivue ones are lighter in the last dev meeting.

Remi-Gau:
if the data are saved to disk they will be tons smaller because we would just point niivue to read it from there rather than imbed the bytes in the html

examples/04_glm_first_level/plot_bids_features.py
```diff
@@ -251,25 +251,3 @@
 # View the generated files
 files = sorted((output_dir / "derivatives" / "nilearn_glm").glob("**/*"))
 print("\n".join([str(x.relative_to(output_dir)) for x in files]))
-
-# %%
-# Only generating the HTML report
-# -------------------------------
```
Remi-Gau, line 257: report is already generated by save_glm_to_bids

examples/05_glm_second_level/plot_oasis.py
```diff
@@ -155,41 +155,22 @@
 # gray matter density on that dataset.
 
 # %%
-# Generating a report
```
Remi-Gau, line 158: report is already generated by save_glm_to_bids

examples/07_advanced/plot_surface_bids_analysis.py
```diff
@@ -164,33 +175,18 @@
...
-    plot_surf_stat_map(
-        surf_mesh=fsaverage5["inflated"],
-        stat_map=results,
-        hemi=hemi,
-        title=f"(language-string), {hemi} hemisphere\nabs(t) >= 1.96",
-        threshold=1.96,
-        bg_map=fsaverage_data,
-    )
-
-show()
```
Remi-Gau, line 196: use report for visualization instead

examples/04_glm_first_level/plot_localizer_surface_analysis.py
Remi-Gau, line 1: report output example: https://output.circle-artifacts.com/output/job/0b2654c5-4f81-4750-8f50-d580e863ad37/artifacts/0/dev/auto_examples/04_glm_first_level/plot_localizer_surface_analysis.html#estimate-contrasts

bthirion: Thx ! it looks good.
A couple of comments:
* the mask display is not completely clear to me. 
* regarding surface display, I think it would make more sense to have a more systematic default display mode, e.g. left hemisphere  lateral, lh medial, rh medial, rh lateral.

Remi-Gau:
test failing in CI but does not fail when ran alone or locally: checking for another instance of test pollution

nilearn/glm/_base.py
```diff
@@ -67,6 +67,7 @@ def generate_report(
         height_control="fpr",
         min_distance=8.0,
         plot_type="slice",
+        cut_coords=None,
```
Remi-Gau, line 70: I think that this parameter add been missing since forever eventhough the underlying make_glm_report called by this method accepts it: so that's a bug fix for me.

should I add this to the changelog?

examples/02_decoding/plot_haxby_glm_decoding.py
```diff
@@ -121,11 +121,9 @@
 # and have the :term:`contrast`, we can quickly create a summary report.
 
 from nilearn.image import mean_img
-from nilearn.reporting import make_glm_report
 
 mean_img_ = mean_img(func_filename, copy_header=True)
-report = make_glm_report(
-    glm,
+report = glm.generate_report(
```
Remi-Gau, line 126: decided to emphasize more the model.generate_report() method rather than the make_glm_report function to be more consistent with what the maskers do

nilearn/reporting/glm_reporter.py
```diff
@@ -190,234 +188,138 @@ def make_glm_report(
...
-    Returns
-    -------
-    url_plot_svg : String
-        SVG Image Data URL.
-
-    """
-    try:
-        return figure_to_svg_quoted(plot)
-    except AttributeError:
-        return figure_to_svg_quoted(plot.figure)
```
Remi-Gau, line 325: dumped this into a lower level function and made it more explicit because the try / catch here was just to handle save figures when the plot is an Axes and not a Figure

Remi-Gau:
TODO
- reports are not showing in all examples
- take care of the few lines not covered

man-shu:
Some of the contrast names are getting cut in this example [plot_localizer_surface_analysis.py](https://github.com/nilearn/nilearn/pull/5202/files#diff-3af8d11acd6a8db1b21afa458d8b957e6f3c134fb1438d20f66254786d6b3c6b)

So maybe as we discussed (during the drop in hour) switching to SVG for contrast/design matrices and PNG for surface plots would solve this

Screenshot 2025-03-12 at 6 33 36 PM

examples/04_glm_first_level/plot_adhd_dmn.py
```diff
@@ -137,11 +135,13 @@
 # %%
 # We have several ways to access the report:
 
-# report  # This report can be viewed in a notebook
+# This report can be viewed in a notebook
+report
 # report.open_in_browser()
```
man-shu, line 140: I think putting this in a separate cell would print the report

bthirion (approved):
LGTM, thx !

Remi-Gau:
> TODO
> 
>     * reports are not showing in all examples
> 
>     * take care of the few lines not covered

done

Remi-Gau:
> Some of the contrast names are getting cut

would suggest tackling this in a follow up

nilearn/glm/_base.py
man-shu: I think it would be good to mention in the changelog

man-shu (approved):
Otherwise LGTM!

nilearn/glm/_base.py
Remi-Gau: missed this comment:
will update the changelog post-hoc

PR #5203, opened 2025-03-10
[ENH] Refactor plotting to separate backends

hndgzkn:
- Closes None

- Move `plotting.surf_plotting` to `plotting.surface.surf_plotting`
- Break down `plotting.surface.surf_plotting.py` and move:
        - general utility functions to `plotting.surface._utils`
        -  backend specific code to `plotting.surface._matplotlib` and `plotting.surface._plotly` modules
- Update imports in `test_surf_plotting.py` accordingly

This would improve coherence of the modules and we would avoid the problems due to matplotlib and plotly specific imports from interface modules.

This PR is an example. If it seems reasonable, the same can be applied to `img_plotting`, etc. in `plotting` package.


hndgzkn:
@Remi-Gau For the moment, I've separated backends only for `surf_plotting.plot_surf`. If this structure is ok with you, I'll continue to do the same for the rest of the functions in `surf_plotting`.

Moving `plotting.surf_plotting` under `plotting.surface` is not mandatory, but I think it creates a more coherent subpackage.  (I think of moving html_surface.py here as well) As far as I see, `surf_plotting` (`html_surface` as well) module is not mentioned in public API, so changing package structure would not create a problem, I guess. Do you see any inconvenience?

I thought of separating backends for other modules in separate PRs. Would that be OK? Otherwise it seems to grow big and becomes difficult to solve merge conflicts.

Remi-Gau:
trying to follow along as this progresses: ping me when you need a review

also I think you will an __init__.py file in plotting.surface

Remi-Gau:
> Moving `plotting.surf_plotting` under `plotting.surface` is not mandatory, but I think it creates a more coherent subpackage. 

Agree

> As far as I see, `surf_plotting` (`html_surface` as well) module is not mentioned in public API, so changing package structure would not create a problem, I guess. Do you see any inconvenience?

unless I missed it I think you are correct that surf_plotting is not mentioned in the user facing API

> I thought of separating backends for other modules in separate PRs. Would that be OK? Otherwise it seems to grow big and becomes difficult to solve merge conflicts.

agree: this PR is already big

bthirion:
LMK when you want a review.

hndgzkn:
> LMK when you want a review.

Ok, I'm planning to make a few more changes, then it will be ok.

Remi-Gau:
@hndgzkn 
while working on https://github.com/nilearn/nilearn/pull/5221 I am realizing that I do not see any reason why nilearn/plotting/html_document.py should be in the the plotting subpackage?

it does not depend on any of our plotting backends?

does it make sense to move to nilearn._utils instead as it is not part of any of the public API ?

nilearn/plotting/_utils.py
```diff
@@ -6,14 +6,17 @@
...
+    warning = (
+        f"'{engine}' is not installed. To be able to use '{engine}' as "
+        "plotting engine for 'nilearn.plotting' package:\n"
+        " pip install 'nilearn[plotting]'"
+    )
+    warn(warning)
 
 
-def save_figure_if_needed(fig, output_file):
+def save_figure_if_needed(fig, output_file, bbox_inches=None):
```
hndgzkn, line 19: It would be good to update this later to use SurfaceFigure interface. And implementing an equivalent of PlotlySurfaceFigure for matplotlib might be an option.

nilearn/plotting/surface/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+)
+
+__all__ = [
+    "check_surface_plotting_inputs",
+    "plot_img_on_surf",
+    "plot_surf",
+    "plot_surf_contours",
+    "plot_surf_roi",
+    "plot_surf_stat_map",
+    "sanitize_hemi_for_surface_image",
```
hndgzkn, line 20: I want to move `html_surface` under `plotting.surface`. Then we won't need to make `check_surface_plotting_inputs` and `sanitize_hemi_for_surface_image` public. 

Or is it possible that we need those in `nilearn.surface`?

nilearn/plotting/surface/_backend.py
```diff
@@ -0,0 +1,434 @@
...
+            labels=labels,
+            colors=colors,
+            legend=legend,
+            cmap=cmap,
+            title=title,
+            output_file=output_file,
+            axes=axes,
+            figure=figure,
+            **kwargs,
+        )
```
hndgzkn, line 168: Shall I set `hemi='left'` if `hemi=None`? 

It is not done explicitly in the original code, it is done implicitly by not passing `hemi` value to `plot_surf` function and `plot_surf` uses its default which is `hemi=left`. But I think it is better to do it explicitly; ie. either set `hemi='left'` (or other valid value) or handle the error when `hemi=None`.

nilearn/plotting/surface/_matplotlib_backend.py
```diff
@@ -0,0 +1,810 @@
...
+        title=None,
+        title_font_size=None,
+        output_file=None,
+        axes=None,
+        figure=None,
+    ):
+        parameters_not_implemented = {
+            "symmetric_cmap": symmetric_cmap,
+            "title_font_size": title_font_size,
+        }
```
hndgzkn, line 360: This did not exist in original code, let me know if it is not ok.

nilearn/plotting/surface/__init__.py
Remi-Gau: if things need to be shared across different modules and be kept private then in this case they should go somewhere in nilearn._utils

nilearn/plotting/surface/_backend.py
Remi-Gau: yes let's do that but it'd be good to centralize those defaults in a single place otherwise we end up having to manual check they are consistent in an increasing number of places

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -0,0 +1,938 @@
...
+    labels=None,
+    colors=None,
+    legend=False,
+    cmap="tab20",
+    title=None,
+    output_file=None,
+    axes=None,
+    figure=None,
+    **kwargs,
+):
```
hndgzkn, line 270: This is not implemented for plotly. But I've put a method in plotly backend and raise NotImplementedError. I thought of adding `engine` parameter here. Would that be ok?

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -0,0 +1,938 @@
...
+    title=None,
+    colorbar=True,
+    vmin=None,
+    vmax=None,
+    threshold=None,
+    symmetric_cbar="auto",
+    cmap=DEFAULT_DIVERGING_CMAP,
+    cbar_tick_format="%i",
+    **kwargs,
+):
```
hndgzkn, line 618: This is not implemented for plotly. But I've put a method in plotly backend and raise NotImplementedError. I thought of adding `engine` parameter here. Would that be ok?

nilearn/plotting/surface/__init__.py
hndgzkn: They are only used by `plotting.html_surface` outside of `plotting.surface`. If I move it under `plotting.surface`, there will be no other module using it.

nilearn/plotting/surface/_backend.py
hndgzkn: I can set all defaults at the top of `surf_plotting`. However, if the user enters `hemi=None`, then there is an error. Then does that mean that `hemi` should never have value `None`?

hndgzkn:
> > @hndgzkn while working on #5221 I am realizing that I do not see any reason why `nilearn/plotting/html_document.py` should be in the the plotting subpackage?
> > it does not depend on any of our plotting backends?
> > does it make sense to move to nilearn._utils instead as it is not part of any of the public API ?
> 
> I did not have a deep look at it, but yes, it might.
> 
> I'll work on it in a follow up PR.

Sorry, I answered quickly. If you are working on that and want to move, I think it makes sense. It seems to be used by both reporting and plotting packages.

nilearn/plotting/surface/_matplotlib_backend.py
hndgzkn: Gives warning as plotly backend gives warnings for the parameters not implemented for plotly.

nilearn/plotting/displays/tests/test_figures.py
```diff
@@ -0,0 +1,108 @@
...
+
+try:
+    import IPython.display  # noqa:F401
+except ImportError:
+    IPYTHON_INSTALLED = False
+else:
+    IPYTHON_INSTALLED = True
+
+
+def test_surface_figure():
```
bthirion, line 17: Please add one-line docstrings for tests, to explain what they do.

nilearn/plotting/surface/tests/test_backend.py
```diff
@@ -0,0 +1,91 @@
...
+
+def test_get_surface_backend_unknown_error():
+    """Tests to see if get_surface_backend raises error when plotly is not
+    installed.
+    """
+    with pytest.raises(ValueError, match="Unknown plotting"):
+        get_surface_backend("unknown")
+
+
+class BaseTestSurfaceBackend:
```
bthirion, line 39: Why do we need a class here ?

nilearn/plotting/surface/tests/test_matplotlib_backend.py
```diff
@@ -0,0 +1,215 @@
...
+    },
+}
+
+
+@pytest.fixture
+def engine():
+    return ENGINE
+
+
+class TestMatplotlibBackend(BaseTestSurfaceBackend):
```
bthirion, line 58: Same question here

nilearn/plotting/surface/tests/test_matplotlib_backend.py
```diff
@@ -0,0 +1,215 @@
...
+    """Tests MatplotlibBackend class methods."""
+
+    pass
+
+
+@pytest.mark.parametrize("hemi, views", MATPLOTLIB_VIEWS.items())
+def test_get_view_plot_surf_matplotlib(hemi, views):
```
bthirion, line 65: docstring...

nilearn/plotting/surface/tests/test_matplotlib_backend.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+@pytest.mark.parametrize(
+    "data,expected",
+    [
+        (np.linspace(0, 1, 100), (0, 1)),
+        (np.linspace(-0.7, -0.01, 40), (-0.7, -0.01)),
+    ],
+)
+def test_get_bounds(data, expected):
```
bthirion, line 79: idem ...

nilearn/plotting/surface/tests/test_plotly_backend.py
```diff
@@ -0,0 +1,232 @@
...
+    ),
+]
+
+
+@pytest.fixture
+def engine():
+    return ENGINE
+
+
+class TestPlotlyBackend(BaseTestSurfaceBackend):
```
bthirion, line 158: Same question here.

bthirion (commented):
The PR is really hard to review. isn't there any means to split it into smaller chunks ?

hndgzkn:
> The PR is really hard to review. isn't there any means to split it into smaller chunks ?

Yes, you are right. I'll divide it into pieces.

nilearn/plotting/surface/tests/test_backend.py
hndgzkn: I created this base test class, that has all necessary methods to test both backends. 

When I extend this in `test_plotly_backend.py` it will take `engine='plotly'` and when I extend it in `test_matplotlib_backend.py`, it will take `engine='matplotlib'` and run all tests in the base class. I do not need to rewrite  the same tests for both backends, or check with if conditions depending on the engine, etc. 

Tests in the backend test files won't run, unless the backend is installed.

Remi-Gau:
related and encountered in another PR: _MNI152Template() should be accessible even when no plotting backend is present, not sure if you touched on this

Remi-Gau:
@hndgzkn I think that we can close this one

please reopen if you feel otherwise

PR #5216, opened 2025-03-13
[WIP][ENH] plot RGB images (diffusion)

### LINKED ISSUE #4224: [ENH] Nilearn plotting of RGB-type images

Opening this issue as a follow up to [this neurostars thread](https://neurostars.org/t/plotting-rgb-images-in-nilearn/27798/4)
There is also an existing issue that is somewhat similar: https://github.com/nilearn/nilearn/issues/522

## Background
In many dMRI methods we'll derive a direction encoded color (DEC) map that is used to show the principal direction of the WM in a voxel. The X Y and Z values of a unit vector get mapped to the R G and B channels respectively. Sometimes the A channel is weighted by FA or some other derived measure so the coloring is proportional to the certainty of the direction.

It also looks like some very early examples in this repo used RGB to encode how face/house/car-representing a voxel was. Perhaps since then the plotting methods made stronger assumptions about data being scalar? 

### Benefits to the change

Nilearn plotting is hands-down my favorite way to display volumetric imaging data, particularly when displaying images with multiple resolutions as foregrounds and backgrounds. I would love to be able to plot as T1w image as a background and a DEC map as an overlay.

Currently there is no python-based implementation of DEC plotting on a high-res background. The current standard practice is to resample the RGB image into the resolution of the background image and plot with imshow, selecting a slice based on the i,j,k of the high res image. This can introduce lots of orientation/resolution bugs.

It would be awesome to have something like this 

```python
# Easiest version

plotting.view_img(
    color_fa_image,
    bg_img=background_t1w_image,
    title="Color FA",
    cmap="direct_rgba"
)
```

Maybe there's a way to do this using nilearn plotting as is? For example, maybe I could loop over the different color channels and add them as separate overlays?

There is some example data available from ITK Snap [here](http://www.nitrc.org/frs/download.php/6176/diffusion.zip) unfortunatrly `t1.nii.gz` isn't high res, but the `tensor_rgb.nii.gz` is a good example of what the RGB channels would look like and `tensor_fa.nii.gz` would be a good example of the alpha channel. @arokem probably has better examples of these kinds of images

Remi-Gau:
- Closes https://github.com/nilearn/nilearn/issues/4224

- allow to use plot_stat_map to plot RGB images (4D images with 3 volumes)

### DIFF
--- a/nilearn/plotting/displays/_slicers.py
+++ b/nilearn/plotting/displays/_slicers.py
@@ -12,11 +12,11 @@ from matplotlib.colors import LinearSegmentedColormap, ListedColormap
 from matplotlib.transforms import Bbox
 from nibabel import Nifti1Image
 
-from nilearn._utils import check_niimg_3d, fill_doc
+from nilearn._utils import check_niimg_3d, check_niimg_4d, fill_doc
 from nilearn._utils.niimg import is_binary_niimg, safe_get_data
 from nilearn._utils.niimg_conversions import _check_fov
 from nilearn._utils.param_validation import check_params
-from nilearn.image import get_data, new_img_like, reorder_img
+from nilearn.image import get_data, index_img, new_img_like, reorder_img
 from nilearn.image.resampling import get_bounds, get_mask_bounds, resample_img
 from nilearn.plotting._utils import (
     check_threshold_not_negative,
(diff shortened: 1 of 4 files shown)


Remi-Gau:
right now just hacking things to see how annoying it will be

Issue #4625, opened 2024-10-10
refactor generate_report across maskers


Remi-Gau:
I hesitate to propose, but at least to discuss:

This part of code is common in `generate_report`.  `generate_report` exists in many maskers under `nilearn/maskers` and `nilearn/experimental/surface/maskers`. Maskers generally inherit from `BaseMasker` or `BaseEstimator`, and `BaseMasker` inherits from `BaseEstimator`.  

Isn't it possible to move this method to base class, and implement any difference in subclasses to avoid repetitions. In `nifti_maps_masker.py` and `nifti_spheres_masker.py` `generate_report` has params and additional lines though.

_Originally posted by @hndgzkn in https://github.com/nilearn/nilearn/pull/4527#discussion_r1795425381_

hndgzkn:
@Remi-Gau Can I work on this one?

Remi-Gau:
go ahead and start having a look but I am also working on this from a slightly different angle:
1. creating report GLM for surface GLM (https://github.com/nilearn/nilearn/pull/4442)
2. then will apply the same 'style' to the volume based GLM
3. then standardize all reports (GLM and maskers) to have the same style (see also: https://github.com/nilearn/nilearn/issues/4368)

So I think this issues related more to 3, that is partly orthogonal to 1 at the moment.

But we may have to coordinate a bit to make we don't work on the same.

Actually #442 is almost ready for review so feel free to have a look at it.

Remi-Gau:
was closed by #5221

PR #5221, opened 2025-03-13
[REF] refactor generating report warnings when no matplotlib is present

Remi-Gau:
- Relates to #4625 

- move common warning to the generate_report function


Remi-Gau:
to be discussed but we could change the behavior when matplotlib is not installed:

instead of returning None, we could fall back on what the maskers do when they are initialized with `reports = False`, in this case generate_report return an empty report:

a bit like what is done for GLM reports

https://nilearn.github.io/dev/modules/generated_reports/glm_reports_examples.html#surface-glm

image

man-shu (commented):
I am wondering why aren't you removing the warnings from some of the maskers like nifti/surfacemapsmasker or spheres masker?

Remi-Gau:
> I am wondering why aren't you removing the warnings from some of the maskers like nifti/surfacemapsmasker or spheres masker?

not sure why I did not do it at first: just done it now, let's see if CI is happy with it

bthirion:
> to be discussed but we could change the behavior when matplotlib is not installed:
> 
> instead of returning None, we could fall back on what the maskers do when they are initialized with `reports = False` or before fitting, in this case generate_report return an empty report where some parts (like the images) are missing:
> 
> a bit like what is done for GLM reports
> 
> https://nilearn.github.io/dev/modules/generated_reports/glm_reports_examples.html#surface-glm
> 
> image

Unclear to me what it looks like ?

bthirion:
Still (naively) wondering whether this is the right approach. If there is no matplotlib installed, do you expect to have people looking at a report ?

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -378,15 +377,6 @@ def _post_masking_atlas(self, visualize=False):
...
         """Generate a report."""
-        if not is_matplotlib_installed():
-            with warnings.catch_warnings():
-                mpl_unavail_msg = (
-                    "Matplotlib is not imported! No reports will be generated."
-                )
-                warnings.filterwarnings("always", message=mpl_unavail_msg)
-                warnings.warn(category=ImportWarning, message=mpl_unavail_msg)
-                return [None]
-
```
hndgzkn, line 389: Could `generate_report` be implemented in the `BaseMasker` to avoid multiple similar implementations of `generate_report` ?  

```
class BaseMasker:

    def _has_plotting_dependencies(self):
        ....

    def generate_report(self):
         if not self._has_plotting_dependencies():
             return None
          
         from nilearn.reporting.html_report import generate_report

        # this function is to be used by maskers, which change self before generating report
         self._prepare_self()

         return generate_report(self)
```

Remi-Gau: maybe but some maskers do a few extra things before calling nilearn.reporting.html_report.generate_report
so for now I wanted to keep it simple and refactor the easiest bit first.

hndgzkn (approved):
Seems OK.

PR #5226, opened 2025-03-13
[ENH] generate method section in GLM reports

### LINKED ISSUE #4493: GLM method section
Given I was working on methods reporting recently, I thought it would be nice if the GLM report (or maybe even the GLM first and second level class instances) contained a method section that could copy pasted into an article.

Remi-Gau:
- Closes #4493

- generate a method section for GLM reports

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 doc/templates/components/edit-this-page.html
+nilearn/reporting/data/html/method_section.html
--- a/examples/02_decoding/plot_haxby_glm_decoding.py
+++ b/examples/02_decoding/plot_haxby_glm_decoding.py
@@ -25,9 +25,9 @@ More specifically:
 import numpy as np
 import pandas as pd
 
-from nilearn import datasets
+from nilearn.datasets import fetch_haxby
 
-haxby_dataset = datasets.fetch_haxby()
+haxby_dataset = fetch_haxby()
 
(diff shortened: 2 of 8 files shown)


bthirion:
Looks good, but I couldn't find a generated report to see how it looks like.

Remi-Gau:
> Looks good, but I couldn't find a generated report to see how it looks like.

@bthirion 

you can find one here: 

https://output.circle-artifacts.com/output/job/b2f34854-6009-498f-b026-f2311cc59867/artifacts/0/dev/auto_examples/02_decoding/plot_haxby_glm_decoding.html#generating-a-report

image

bthirion:
Looks good, thx ! 
The only that is weird is the values range for the correlation matrix display (you expect values to range from -1 to +1).

Remi-Gau:
> Looks good, thx ! The only that is weird is the values range for the correlation matrix display (you expect values to range from -1 to +1).

that's on purpose:

this is because the diagonal is always 1, so if the rest of the matrix has very low values (like here) they would all appear white in comparison to the diagonal

https://github.com/nilearn/nilearn/blob/845f63bd4bef4e5e015779964cfb7a34f9d2f546/nilearn/plotting/matrix_plotting.py#L723-L727

hndgzkn (approved):
LGTM.

PR #5228, opened 2025-03-14
[MAINT] add check_transformers replacement for instead of using sklearn one

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/issues/4538

- add a check_transformer 
  - fit_transform method should work on non fitted estimator
  - fit_transform should give same result as fit then transform


nilearn/glm/_base.py
```diff
@@ -1,10 +1,10 @@
 from nibabel.onetime import auto_attr
-from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.base import BaseEstimator
 
 from nilearn._utils import CacheMixin
 
 
-class BaseGLM(TransformerMixin, CacheMixin, BaseEstimator):
```
Remi-Gau, line 7: as far I can tell the GLM estimators do not need the transformers functionalities

bthirion (approved):
LGTM, thx.

nilearn/_utils/estimator_checks.py
```diff
@@ -103,6 +106,14 @@
     "check_fit_check_is_fitted": (
         "replaced by check_masker_fitted or check_glm_is_fitted"
     ),
+    "check_fit_check_is_fitted": "replaced by check_masker_fitted",
```
hndgzkn: The same key exists in the above line. Is this line a replacement for that?

nilearn/_utils/estimator_checks.py
```diff
@@ -466,6 +475,25 @@ def check_masker_clean(estimator):
...
+    - fit_transform should give same result as fit then transform
+    """
+    if accept_niimg_input(estimator):
+        input_img = _img_4d_rand_eye_medium()
+    else:
+        input_img = _make_surface_img(100)
+
+    signal_1 = estimator.fit_transform(input_img)
+
+    estimator = clone(estimator)
```
hndgzkn, line 490: The [doc for `clone`](https://scikit-learn.org/stable/modules/generated/sklearn.base.clone.html) says that if the `random_state` is specified and it is not `integer`, the clone might return a different result. Shall we consider that?

nilearn/_utils/estimator_checks.py
Remi-Gau: yeah I fixed that: pre-commit / ruff were complaining about it

nilearn/_utils/estimator_checks.py
Remi-Gau: this merits a double check but I think we do set an integer to all our random state when testing

PR #5231, opened 2025-03-14
[ENH] use saved GLM output to generate report

### LINKED ISSUE #2404: Generate reports post-hoc, using saved information
As of now, make_glm_report() has to be run alongside the analysis since it expects a FirstLevelModel or SecondLevelModel object as input. It might be useful to have a feature that allows for post-hoc report generation using saved design matrix, contrast maps etc.

Remi-Gau:
- Closes #2404

- uses pregenerated GLM output for reports
- refactor make_glm_report and save_glm_to_bids as they both:
  - compute contrasts
  - generate figures for design matrices and contrasts
- add design matrix correlation to output of save_glm_to_bids
- ensure that save_glm_to_bids and make_glm_report can run even if matplotlib is missing

## TODO

in follow up PR

- [ ] save mask to disk and reuse in report
- [ ] save tables of activation locations
- [ ] make outtput name more BIDS like by relying on input filenames
- [ ] support saving surface data to disk


Remi-Gau:
@tsalo 

realizing that saving the GLM to BIDS does not:
- save the mask of where the GLM was run
- does not handle surface data

both of those should probably be included in BEP 41

https://bids.neuroimaging.io/extensions/beps/bep_041.html

nilearn/glm/_base.py
```diff
@@ -70,6 +70,7 @@ def generate_report(
         cut_coords=None,
         display_mode=None,
         report_dims=(1600, 800),
+        input=None,
```
Remi-Gau, line 73: probably need a better name

Remi-Gau:
ah also saving the matrix of the correlation among regressors seems like a valuable output

Remi-Gau:
@tsalo 
another thing was there any reason why we are not trying to get the full name of the input file to prepend to the output files in case of first level GLM?

currently: if no prefix is passed to `save_glm_to_bids` then it will not add anything before the filenames, but if files are passed as input to fitting the GLM, we could keep track of those and use them when saving output to disk to have more informative filenames (could at least include the name of the space the data are in).

nilearn/glm/_base.py
```diff
@@ -65,11 +65,14 @@ def generate_report(
...
         cluster_threshold=0,
         height_control="fpr",
+        two_sided=False,
         min_distance=8.0,
         plot_type="slice",
         cut_coords=None,
         display_mode=None,
         report_dims=(1600, 800),
+        input=None,
+        verbose=1,
```
Remi-Gau, line 75: document in doc string

nilearn/glm/_base.py
Remi-Gau, line 1: quite a few other parameters we missed

nilearn/glm/utils.py
Remi-Gau, line 1: considering turning some of those into private methods into the base GLM class

nilearn/reporting/glm_reporter.py
```diff
@@ -79,6 +81,8 @@ def make_glm_report(
     cut_coords=None,
     display_mode=None,
     report_dims=(1600, 800),
+    input=None,
```
Remi-Gau, line 84: @bthirion @man-shu 

The way I am reusing the output generated when saving the GLM to disk to generate the report is by using a dict to keep track of where we saved the files for each design matrix figures, contrast nifti image...

Then this dict can be passed to generate_report as a parameter.

What I do not like about this is that we now give user access to a new parameter that they can tweak themselves and whose content we may then have to validate.

Given the odds of human beings handling file paths correctly is close to 0, I am not a big fan of this approach.
So I am thinking of making this dict of filenames a private attribute of the GLM model object.

Before I start doing this, does this make sense to you?

A simple 👍🏽 is enough of an answer (if my explanation was clear).

Remi-Gau:
also something to implement: save_glm_to_bids should be able to run without maptlotlib, sure figures won't be generated but saving nifti to disk does not require matplotlib

nilearn/reporting/glm_reporter.py
man-shu: Yes makes sense to me!

Remi-Gau: done

Remi-Gau:
> also something to implement: save_glm_to_bids should be able to run without maptlotlib, sure figures won't be generated but saving nifti to disk does not require matplotlib

DONE

nilearn/_utils/plotting.py
Remi-Gau, line 1: @hndgzkn 

I moved some 'private' plotting functions here as they are used both interfaces/bids and reporting/glm_reporter

they can probably go in a better place if we want to clean up where the code for different backends should reside

hndgzkn: Ok. I note that.

Remi-Gau:
some figures are printed in the doc when they should not be

image

Remi-Gau:
also @tsalo what about extra TSV that would list activation location for a given contrast / threshold?
we need to compute those for the html report so it may make sense to have them as output... though it becomes hard to name things properly to avoid overwriting previous unless we start abusing the desc entity to have things like `desc-thres3pt1AlphaTwoSided`
was it considered to have them in the BEP?

Remi-Gau:
> I _think_ I leaned toward not inferring the filename because we didn't necessarily expect the filenames to follow BIDS convention.

I was only thinking of trying to parse filenames using what nilearn can alreayd do and err on the side of caution

Remi-Gau:
> what about extra TSV that would list activation location for a given contrast / threshold?

otherwise we could just dump those as tsv files and not worry about overwriting because that's anyways already the case for the report.html

tsalo:
I think the cluster tables would be very useful in general (reporting results in manuscript, sharing data for coordinate-based meta-analyses, etc.) but I think they're outside of the scope of my BEP. I do wonder if there would be any benefit to something BEP-like for coordinate-based meta-analyses though.

nilearn/_utils/plotting.py
man-shu, line 1: Looks like none of the functions introduced here are being tested?

nilearn/interfaces/tests/test_bids.py
man-shu, line 1: I dont understand why are we removing the matplotlib fixture from these tests. Is that because of the plotting reorganization?

nilearn/_utils/plotting.py
Remi-Gau: ok that's weird...

Remi-Gau: and not consistent with this:

image

nilearn/interfaces/tests/test_bids.py
man-shu: OK I see, so we only check the existence of plots (.svg files) when `is_matplotlib_installed` is True.

But interestingly, all the lines under `is_matplotlib_installed` are also flagged by codecov. Could this be related to codecov warnings in _utils/plotting.py? Seems like somehow codecov is not detecting the test envs where matplotlib is installed...

man-shu (approved):
But looks good to me overall!

nilearn/glm/_base.py
```diff
@@ -176,6 +240,10 @@ def generate_report(
             For display purposes only.
             Minimum distance between subpeaks in mm.
 
+        two_sided : :obj:`bool`, default=False
+            Whether to employ two-sided thresholding
+            or to evaluate positive values only.
```
hndgzkn, line 245:
```suggestion
            Whether to employ two-sided thresholding
            or to evaluate either positive or negative values only.
```

hndgzkn: Couldn't it be also negative values depending on the threshold?

nilearn/glm/_base.py
```diff
@@ -227,9 +295,178 @@ def generate_report(
...
+                    f"contrast names must be strings, not {type(k)}"
+                )
+
+            if not isinstance(v, (str, np.ndarray, list)):
+                raise ValueError(
+                    "contrast definitions must be strings or array_likes, "
+                    f"not {type(v)}"
+                )
+
+        if self.__str__() == "Second Level Model":
```
hndgzkn, line 353: Could "Second Level Model" be defined as class constant in `nilearn.glm.second_level.second_level.SecondLevelModel` and comparison use that?

nilearn/glm/first_level/first_level.py
```diff
@@ -414,6 +414,9 @@ class FirstLevelModel(BaseGLM):
 
     """
 
+    def __str__(self):
+        return "First Level Model"
```
hndgzkn, line 418: Could "First Level Model" be defined as class constant in `FirstLevelModel` and that constant be returned here?

nilearn/glm/_base.py
hndgzkn: SecondLevelModel can have subdirectory property, and design_matrices property that returns the correct result for itself to avoid if conditions (if these are checked elsewhere as well).

nilearn/glm/first_level/first_level.py
hndgzkn: Or create a `level` property for the model?

nilearn/interfaces/bids/glm.py
```diff
@@ -332,88 +254,41 @@ def save_glm_to_bids(
...
-def _model_level(model):
-    from nilearn.glm.first_level import FirstLevelModel
-
-    return 1 if isinstance(model, FirstLevelModel) else 2
+    return model
 
 
 def _write_model_level_statistical_maps(model, prefix, out_dir):
-    if _model_level(model) == 2:
+    if model.__str__() == "Second Level Model":
```
hndgzkn, line 291: Could some of these functions actually be part of the model class definition and implementation stays in the specific model instead of putting many if checks for the model type?

nilearn/reporting/_utils.py
```diff
@@ -172,67 +131,20 @@ def dataframe_to_html(df, precision, **kwargs):
...
     Returns
     -------
     attributes_df: pandas.DataFrame
         DataFrame with the pertinent attributes of the model.
     """
     if model.__class__.__name__ in ["FirstLevelModel", "SecondLevelModel"]:
-        return _glm_model_attributes_to_dataframe(
-            model, is_volume_glm=is_volume_glm
-        )
+        return _glm_model_attributes_to_dataframe(model)
```
hndgzkn, line 147: Could we check using `isinstance` here (at line 232) instead of string comparison?

hndgzkn (approved):
I've added some comments. Otherwise seems Ok to me.

nilearn/reporting/_utils.py
Remi-Gau: pretty sure I did this to avoid having a nested import to avoid circular import,
so trying to rely in something

Remi-Gau: actually I think that if I move some function we can avoid this

nilearn/interfaces/bids/glm.py
Remi-Gau: yeah I can probably improve the function that generate output filenames that covers this so this one becomes simpler

PR #5234, opened 2025-03-17
[ENH] Refactor nilearn.plotting - create plotting.surface package 

hndgzkn:
- Closes None

- move `plotting.surf_plotting` to `plotting.surface.surf_plotting`
- move `nilearn.plotting.html_surface` under `nilearn.plotting.surface`
- move corresponding tests under `nilearn.plotting.surface.tests`


Remi-Gau (approved):
so far so good

one thing that you may want to do, running a full build of the doc to make sure we don't get any import error on the exmaples

https://nilearn.github.io/dev/ci.html#continuous-integration

Remi-Gau:
it will take a long time to run (>1h30 hour) so run it once all your changes are in

hndgzkn:
> so far so good
> 
> one thing that you may want to do, running a full build of the doc to make sure we don't get any import error on the exmaples
> 
> https://nilearn.github.io/dev/ci.html#continuous-integration

I've checked with `grep` in the working directory

hndgzkn:
> in your case it is indeed currently building the whole doc:
>
Ok, thanks!

nilearn/surface/surface.py
```diff
@@ -1119,8 +1119,8 @@ def check_mesh_is_fsaverage(mesh):
     with sufficient entries. Basically ensures that the mesh data is
     Freesurfer-like fsaverage data.
 
-    Used by plotting.surf_plotting.plot_img_on_surf and
-    plotting.html_surface._full_brain_info.
+    Used by plotting.surface.plot_img_on_surf and
+    plotting.surface.html_surface._full_brain_info.
```
hndgzkn, line 1123: `nilearn.surface.surface.check_mesh_is_fsaverage` is only used by `plotting.surface.plot_img_on_surf` and `plotting.surface.html_surface._full_brain_info`, shall I move it to `plotting.surface` package?

hndgzkn:
I also think that I can rename `nilearn.plotting.surface.surf_plotting` to `nilearn.plotting.surface._surf_plotting` and `nilearn.plotting.surface.html_surface` to `nilearn.plotting.surface._html_surface` as public functions are exposed in `nilearn.surface.__init__`. Would that be ok @Remi-Gau ?

nilearn/surface/surface.py
Remi-Gau: maybe not: we may still want to run this check even if we have not intention of plotting

Remi-Gau:
> I also think that I can rename `nilearn.plotting.surface.surf_plotting` to `nilearn.plotting.surface._surf_plotting` and `nilearn.plotting.surface.html_surface` to `nilearn.plotting.surface._html_surface` as public functions are exposed in `nilearn.surface.__init__`. Would that be ok @Remi-Gau ?

I think this is pattern I have seen in sklearn and a few other places: make most module private and only expose things via the __init__

but it is not a common one in nilearn, so I am not sure we should do this without having a clear repo wide policy

hndgzkn:
> I think this is pattern I have seen in sklearn and a few other places: make most module private and only expose things via the **init**
> 
> but it is not a common one in nilearn, so I am not sure we should do this without having a clear repo wide policy

I thought it would comply with https://nilearn.github.io/stable/development.html#guidelines-for-private-functions . And from the users' point of view, it would be more clear what is expected to be used by the users and what not (and thus subject to change).

nilearn/surface/surface.py
hndgzkn: Shall I completely remove these two lines then? I'm not sure if it is a good idea to put such information in docstring.

Remi-Gau: yes definitely misleading
better to remove them

bthirion:
The PR looks technically OK.
However, this adds a level of nesting which means a bit of complexity. Maybe this is needed now ?

bthirion (approved):

hndgzkn:
> The PR looks technically OK. However, this adds a level of nesting which means a bit of complexity. Maybe this is needed now ?

From users' point of view, it does not change anything actually. All imports are still from `nilearn.plotting`. 

From contributors' point of view, it increases depth (not that much actually, there is already nilearn.plotting.displays package at the same level) however it creates more coherent subpackages. It becomes simpler to identify which part of the code is related to which concept. 

I also think that, these subpackages should keep implementation details inside the package and not expose to outside so that if we change implementation, users will not be affected from that, and we do not need to have deprecation cycle each time.

hndgzkn:
> The PR looks technically OK. However, this adds a level of nesting which means a bit of complexity. Maybe this is needed now ?

Or if you prefer, we can discuss that in team meeting today.

Remi-Gau:
> From contributors' point of view, it increases depth (not that much actually, there is already nilearn.plotting.displays package at the same level) however it creates more coherent subpackages. It becomes simpler to identify which part of the code is related to which concept.

I kind of agree that this helps quickly (literally at first sight) knowing which modules 'go together' rather than a flat listing of all the modules.

PR #5236, opened 2025-03-17
[REF] refactor more extra_valid_checks into expected_failed_checks

Remi-Gau:
- Relates to #4538

- further refactor of 'valid' sklearn checks
- expand list of of valid sklearn checks but specify explicitly why some tests need to be skipped


hndgzkn (approved):
Seems Ok to me.

PR #5224, opened 2025-03-13
[ENH] make surface utility functions public

Remi-Gau:
- Closes #4879

Changes proposed in this pull request:

- move the following private functions from nilearn.surface into their public volume equivalent in nilearn.image:

  - new_img_like
  - index_img
  - mean_img
  - concat_imgs
  - iter_imgs

PR #5254, opened 2025-03-19
[REF] refactor code checking compatibility betwen mask and images

Remi-Gau:
- Closes none

- tries to only use a single function to check that a mask and an image have compatible types as this is becoming a more systematic thing we need to do for functions that can handle both surface and volume images

### DIFF
--- a/nilearn/plotting/img_comparison.py
+++ b/nilearn/plotting/img_comparison.py
@@ -1,12 +1,10 @@
 """Functions to compare volume or surface images."""
 
 import warnings
-from pathlib import Path
 
 import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib import gridspec
-from nibabel import Nifti1Image
 from scipy import stats
 
 from nilearn import DEFAULT_SEQUENTIAL_CMAP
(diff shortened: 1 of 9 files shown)


Remi-Gau:
- should probably also be extended to https://github.com/nilearn/nilearn/pull/5224

nilearn/plotting/img_comparison.py
```diff
@@ -469,32 +469,19 @@ def _sanitize_masker(masker, image_type, ref_img):
...
+        masker = NiftiMasker(
+            mask_img=masker,
+            target_affine=ref_img.affine,
+            target_shape=ref_img.shape,
+        )
+    elif isinstance(masker, SurfaceImage):
+        check_same_n_vertices(ref_img.mesh, masker.mesh)
+        masker = SurfaceMasker(
+            mask_img=masker,
+        )
```
man-shu, line 484: I am wondering if this check can also happen within the `check_compatibility_mask_and_images` function

man-shu (commented):
Also I am not sure if this check is being used in all the Surface maskers? Or were you keeping that for another PR?

Otherwise LGTM!

bthirion (approved):
LGTM.

nilearn/plotting/img_comparison.py
Remi-Gau: maybe but I'd rather not do that just yet

Remi-Gau:
> Also I am not sure if this check is being used in all the Surface maskers? Or were you keeping that for another PR?
> 
> Otherwise LGTM!

had assumed that this was not going to be needed but this could a nice systematic check to our maskers (volume or surface): in a follow up PR?

PR #5262, opened 2025-03-21
[ENH] use `imgs` instead of `img` / `X` for fit, transform, fit_transform

Remi-Gau:
- relates to https://github.com/nilearn/nilearn/issues/5095

- rename parameter ``img`` or ``X`` to ``imgs`` for the ``fit``, ``transform`` and ``fit_transform`` method of all the surface maskers as well as the `fit` method of the `NiftiSphereMasker` and the `RegionExtractor`, and fit_transform of the NiftiMasker
- if users pass the old named parameter a deprecation warning will be thrown till 0.13.2
- add an estimator check to enforce this

also:

- implement extract fit_transform method from children classes to the parent class


Remi-Gau:
@bthirion @man-shu @emdupre @jeromedockes 

asking for an early opinion on this

our API for estimators is more "diverse" than it should be:

as first parameter to `fit`, `transform`, `fit_transform`
some of our estimartors use `img` (mostly the surface maskers), some other uses `imgs` (mostly the niftimaskers except for sphere masker that uses `X`), and the GLM estimator also do their own thing, while sklearn uses X pretty much everywhere (AFAICT)

in this PR I propose to use `X` everywhere (note I'd be happy to use `imgs` or `img` but at least to be consistent, or at least to have some guidelines has to when to use `X` or `imgs` or `img`)

we have a decorator to rename parameters on the fly, so users who use the old named parameter would just get a deprecation warning: I would implement this, only if / when we come to a agreement on this

note that something similar could be done to ensure consistency for `inverse_transform`

man-shu:
> note I'd be happy to use imgs or img but at least to be consistent, or at least to have some guidelines has to when to use X or imgs or img

I am more inclined towards using `img` and deprecating `imgs` or `X`. This would be more consistent with functions under plotting/image modules. 

Also adding ConnectivityMeasure and Decoder to the zoo where we use `X`.

I think for inverse_transform `X` would make more sense?

For me `img` -> NiftiImage or SurfaceImage and `X` -> numpy array

Remi-Gau:
that'd be ok with me 
as I asaid as long as we have something that's internally consistent, it's fine

bthirion:
Aren't we bothering users with our homogeneity constraints ? People will have to follow the change. This can break a lot of code...

Remi-Gau:
> Aren't we bothering users with our homogeneity constraints ? People will have to follow the change. This can break a lot of code...

nope: we have a rename_parameter that renames things on the fly and throws a deprecation warning

```python

stat_img = load_sample_motor_activation_image()

nifti_masker = NiftiMasker()
# this is fine
nifti_masker.fit(stat_img)

nifti_masker = NiftiMasker()
# this only throw a deprecation warning
nifti_masker.fit(imgs=stat_img)
```

output

```
DeprecationWarning: The parameter "imgs" will be removed in 0.13.2 release of Nilearn. Please use the parameter "X" instead.
  nifti_masker.fit(imgs=stat_img)
```

Remi-Gau:
but i'd prefer to make sure we agree:
- we want to do this
- what we name things

before I proceed further

jeromedockes:
> nope: we have a rename_parameter that renames things on the fly and throws a deprecation warning

the deprecation just gives them a heads-up, but this breaks the client code just the same: it needs to be changed eventually (or immediately to avoid the warning), and it is still hard to support multiple nilearn versions. 

that being said, in that case I would _guess_ that many users pass X as a positional argument so would not be affected, but it's hard to be sure. we could collect a bunch of scripts that use nilearn on github and check. Not sure if it's worth the effort though

Remi-Gau:
sure it will break something eventually if people don't act on the warning

FYI: using `imgs` across nilearn estimators would be one way to:
- minimize downstream breakage: I think it's the most common variable name for that parameter at the moment, but not used everywhere
- make semantic distinction between estimators that expect images and those that do not

also the surface maskers that use `img` and not `imgs` have been out for only a few months so I'd be surprised it you found them all over github

emdupre:
I'm not going to be able to think this through until later in the week but at a high-level, to make sure I understand :

The core concern IIUC is that developers have to know that the same parameter across function calls (i.e., `imgs` in `MultiNiftiLabelsMasker` vs `img` in `NiftiLabelsMasker`) maps to the `X` of scikit-learn ? 

From my (admittedly poor) memory of this, I thought we had been internally consistent s.t. `imgs` was when multiple images are expected, `img` with a single image, and `X` with a matrix. Is that not what's happening right now ?

bthirion:
Note that the semantics  of X was to be an array-like thing, which an image/SurfaceImage is not.
Won't it be confusing to call images X ?

Remi-Gau:
> all of the SurfaceMaskers are using `img`

I checked with Himanshu and I think that this is just something we overlooked when working on those maskers: so we could almost label those as a bug and not throw deprecation warning for the variable renaming in the surface maskers

Remi-Gau:
OK I am seeing a soft consensus on avoiding `X` as a variable name and trying to use `imgs`.

will adapt the PR a bit to reflect this

Remi-Gau:
seems some maskers would still prefer to use X 

```
  FAILED nilearn/maskers/tests/test_multi_nifti_labels_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: NiftiLabelsMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_nifti_spheres_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: NiftiSpheresMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_surface_maps_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: SurfaceMapsMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_multi_nifti_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: BaseMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_nifti_maps_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: NiftiMapsMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_nifti_labels_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: NiftiLabelsMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_surface_labels_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: SurfaceLabelsMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_multi_nifti_maps_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: NiftiMapsMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_surface_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: SurfaceMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/regions/tests/test_region_extractor.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: NiftiMapsMasker.fit_transform() missing 1 required positional argument: 'X'
  FAILED nilearn/maskers/tests/test_nifti_masker.py::test_check_estimator[estimator42-check_masker_transformer-check_masker_transformer] - TypeError: BaseMasker.fit_transform() missing 1 required positional argument: 'X'
```

nilearn/maskers/nifti_spheres_masker.py
```diff
@@ -621,6 +625,7 @@ def fit(
 
         return self
 
+    @rename_parameters(replacement_params={"X": "imgs"}, end_version="0.13.2")
```
Remi-Gau, line 628:
```suggestion
```

Remi-Gau:
in a similar way, it'd be nice if we had the same parameter name for inverse_transform

signals, region_signals ? entre les 2 mon coeur balance ?

```
grep -rni  'def inverse_transform' nilearn 
nilearn/maskers/surface_labels_masker.py:515:    def inverse_transform(self, signals):
nilearn/maskers/surface_masker.py:349:    def inverse_transform(self, signals):
nilearn/maskers/nifti_labels_masker.py:904:    def inverse_transform(self, signals):
nilearn/maskers/base_masker.py:379:    def inverse_transform(self, X):
nilearn/maskers/nifti_maps_masker.py:675:    def inverse_transform(self, region_signals):
nilearn/maskers/nifti_spheres_masker.py:741:    def inverse_transform(self, region_signals):
nilearn/maskers/surface_maps_masker.py:395:    def inverse_transform(self, region_signals):
nilearn/decomposition/_base.py:578:    def inverse_transform(self, loadings):
nilearn/regions/hierarchical_kmeans_clustering.py:359:    def inverse_transform(self, X_red):
nilearn/regions/parcellations.py:646:    def inverse_transform(self, signals):
nilearn/regions/rena_clustering.py:809:    def inverse_transform(self, X_red):
nilearn/connectome/connectivity_matrices.py:681:    def inverse_transform(self, connectivities, diagonal=None):
```

@emdupre any reason for this diversity?

emdupre:
I don't remember an explicit discussion of this, but I believe the idea is to make it as clear as possible what is being used as inputs to the inverse_transform, since it can be quite class specific. Just writing out the listed classes (exempting surfaces, for now) and their docstrings :

- `NiftiLabelsMasker` (signals): "Signal for each region. "
- `Parcellations` (signals): "Transform signals extracted from parcellations"
- `NiftiMapsMasker`, `NiftiSpheresMasker`: (region_signals) "Signal for each region."
- `ReNA`, `HierarchicalKMeans` (X_red): "reduced 2D data matrix"
- `ConnectivityMeasure` (connectivities): "Return connectivity matrices from connectivities"

If there's any inconsistency that jumps out to me, it's that I'd argue `NiftiLabelsMasker` should also use `region_signals` rather than `signals`. But I'm not sure the gain is big enough to justify the cost, since this is such a widely used class.

nilearn/maskers/base_masker.py
```diff
@@ -298,15 +301,16 @@ def transform(self, imgs, confounds=None, sample_mask=None):
             imgs, confounds=all_confounds, sample_mask=sample_mask
         )
 
+    @rename_parameters(replacement_params={"X": "imgs"}, end_version="0.13.2")
     @fill_doc
     def fit_transform(
```
man-shu, line 306: Just to make sure I am getting this right, wouldn't this also change the input parameter name from `X` ->` imgs` in `fit_transform` for all classes that inherit BaseMasker (meaning all Nifti*Maskers)?

I did not expect that just from reading the changelog entry. Maybe that should also be reflected there.

man-shu (commented):
This is unrelated but also noticed that docstrings for `confounds` and `sample_mask` are not being rendered properly in `transform_single_imgs` everywhere (all Nifti*Maskers and Surface*Maskers):

https://output.circle-artifacts.com/output/job/fefac0f6-5213-47ba-bee7-dd6e4a7a2a0f/artifacts/0/dev/modules/generated/nilearn.maskers.NiftiMasker.html#nilearn.maskers.NiftiMasker.transform_single_imgs

Screenshot 2025-04-14 at 2 24 54 PM

Remi-Gau:
> This is unrelated but also noticed that docstrings for `confounds` and `sample_mask` are not being rendered properly in `transform_single_imgs` everywhere (all `Nifti*Masker` and `Surface*Masker`):
> 
> https://output.circle-artifacts.com/output/job/fefac0f6-5213-47ba-bee7-dd6e4a7a2a0f/artifacts/0/dev/modules/generated/nilearn.maskers.NiftiMasker.html#nilearn.maskers.NiftiMasker.transform_single_imgs
> 

need to check if this is not because of the order of the decorator: I remember getting things like that in the past

Remi-Gau:
> need to check if this is not because of the order of the decorator: I remember getting things like that in the past

OK it also turned out that some methods were straight missing a fill_doc decorator: and this was not caught by our doc string checking scripts.

Remi-Gau:
> and this was not caught by our doc string checking scripts.

my bad: it's listed here: https://github.com/nilearn/nilearn/actions/runs/14442181169/job/40498574835?pr=5262

nilearn/maskers/base_masker.py
Remi-Gau: ha yeah I forgot that one.

Remi-Gau: though in practice only the NiftiMasker does not have its own fit_transform method and therefore relies on that of the parent class

nilearn/glm/_base.py
```diff
@@ -156,6 +157,7 @@ def r_square(self):
             "r_square", result_as_time_series=False
         )
 
+    @fill_doc
```
Remi-Gau, line 160: unrelated but fixes doc

Remi-Gau:
> docstrings for `confounds` and `sample_mask` are not being rendered properly

should be fixed

bthirion (approved):
LGTM, thx !

PR #5263, opened 2025-03-21
[REF] use check_embedded_masker in GLM fit

### LINKED ISSUE #4605: See if `check_embedded_masker` can be used more systematically across estimators
check_embedded_masker runs some checks on the masker of an estimator but only for those found in the decoding / decomposition subpackages.

https://github.com/nilearn/nilearn/blob/f5114ee0b831530907e554a669427e70837029d2/nilearn/_utils/masker_validation.py#L13-L42

Yet very similar checks are run for GLM and other estimators.

See for example:

https://github.com/nilearn/nilearn/blob/f5114ee0b831530907e554a669427e70837029d2/nilearn/glm/first_level/first_level.py#L982-L989

It may be worth checking whether some refactoring cannot be done to reuse some of those checks across estimators.

Some of those 'repeats' may become more and more obivous as we start extending support for SurfaceImage data.

I may be missing something obvious why it cannot be done, so please let me know before I go down this rabbithole.

Remi-Gau:
- Closes #4605


nilearn/glm/first_level/first_level.py
```diff
@@ -1102,71 +1107,40 @@ def _prepare_mask(self, run_img):
...
-                memory=self.memory,
-                verbose=max(0, self.verbose - 2),
-                target_shape=self.target_shape,
-                memory_level=self.memory_level,
+            self.masker_ = check_embedded_masker(
+                self, masker_type, ignore=["high_pass"]
             )
+
+            if isinstance(self.masker_, NiftiMasker):
+                self.masker_.mask_strategy = "epi"
```
bthirion, line 1137: Why are we imposing that ?

bthirion (commented):
Only one point pending. Best,

nilearn/glm/first_level/first_level.py
Remi-Gau: This is what we do on main at the moment so I was trying to make sure we keep things the same:

https://github.com/nilearn/nilearn/blob/171b73c0683956dc12018cd7f4d86f72935131c6/nilearn/glm/first_level/first_level.py#L1128-L1139

We also used to do that before we started supported surface data:

https://github.com/nilearn/nilearn/blob/4f4730163097457cf9ddb5674ffd158ee8fa822e/nilearn/glm/first_level/first_level.py#L590-L601

Remi-Gau: also because the default for nifti masher is to use a background strategy and not EPI

man-shu (approved):
LGTM!

Issue #4797, opened 2024-11-26
B028 Errors with Stack Level Adjustments


SamiLaayat:
This issue focuses on resolving the B028 errors identified in the Nilearn codebase. The B028 rule (no-explicit-stacklevel) enforces specifying an appropriate stacklevel for warnings or logging calls to ensure that the correct source location is reported in the traceback.

In the previous issue ([#4579](https://github.com/nilearn/nilearn/issues/4579)), B028 was listed among the errors to address but was not tackled. Resolving this issue involves updating all occurrences of warnings or logging calls in the codebase to use an explicit and correct stacklevel.

**Reason for Opening This Issue**:
The initial issue became too large to handle efficiently as it grouped multiple errors together. Addressing these individually allows for more focused contributions and precise discussions. This issue specifically targets B028 errors to improve traceability and maintain adherence to best practices.

Proposed Fix:
For each instance triggering a B028 error, **the following actions were taken**:

Identified the appropriate stacklevel based on the context of the warning or logging call.
Adjusted the stacklevel parameter in the relevant code to ensure the correct source of the warning is reflected in the traceback.
Verified that the modifications resolved the B028 errors and ensured no regression or additional warnings were introduced.
**Impact**:
Addressing B028 errors enhances the developer experience by improving the accuracy of traceback information. This makes debugging and maintenance easier for contributors and users of the Nilearn library.

Remi-Gau:
@hndgzkn 
can I let you wrap up this one?

the only warnings that are not updated are those in the plotting subpackage

given that you are focusing on this at the moment, you would just need to replace (or set) the stacklevel of each warning in plotting with a call to nilearn._utils.find_stack_level()

PR #5265, opened 2025-03-22
[MAINT] adjust stacklevel for several warnings

Remi-Gau:
- Relates to #4797


pyproject.toml
```diff
@@ -277,7 +275,7 @@ max-complexity = 21
 "nilearn/decoding/*" = ["B028"]  # 5 errors to fix
 "nilearn/decoding/**/{tests}/*" = ["D103"]  # 81 errors to fix
 "nilearn/decomposition/**/{tests}/*" = ["D103"]  # 32 errors to fix
-"nilearn/glm/*" = ["B028"]  # 19 errors to fix
+# "nilearn/glm/*" = ["B028"]  # 19 errors to fix
```
Remi-Gau, line 278:
```suggestion
```

bthirion (commented):
Thx.
Are the rules to  define stacklevel described somewhere ?

Remi-Gau:
> Are the rules to define stacklevel described somewhere ?

not in our doc but the algorithm is:
- start at stacklevel=1 the warn() call
- go up the stack call and increment stacklevel by 1 every time you do
- once you hit the first public function / method (public user facing piblic i.e as in part of our doc), add 1 to stacklevel and stop.

Remi-Gau:
See also:

> finds the first place in the stack that is not inside [nilearn]

https://github.com/scikit-learn/scikit-learn/issues/26684

Remi-Gau:
there is a automatic stacklevel finding option from pandas: https://github.com/pandas-dev/pandas/blob/dc8401afea462c38cbe4ffd8749aa78330993cf0/pandas/util/_exceptions.py#L37-L63

Remi-Gau:
actually the automatic stacklevel finding would get around a common problem mentioned here:

> Is it possible that the call stack for a same warning is different so that stacklevel needs to be set dynamically? An example can be using fit versus fit_predict. The stack of the latter is commonly deeper (by 2 levels) so that stacklevel may need to be different as well.

https://github.com/scikit-learn/scikit-learn/issues/26684#issuecomment-1607669721

Remi-Gau:
> there is a automatic stacklevel finding option from pandas:
> 
> https://github.com/pandas-dev/pandas/blob/dc8401afea462c38cbe4ffd8749aa78330993cf0/pandas/util/_exceptions.py#L37-L63

actually I REALLY like this: this would make things a WHOLE lot easier

Remi-Gau:
will try it in this PR

Remi-Gau:
> will try it in this PR

ok maybe not in this PR but in another one

nilearn/connectome/group_sparse_cov.py
```diff
@@ -386,7 +386,7 @@ def _group_sparse_covariance(
 
                         if fder == 0:
                             msg = "derivative was zero."
-                            warnings.warn(msg, RuntimeWarning)
+                            warnings.warn(msg, RuntimeWarning, stacklevel=4)
```
bthirion, line 389: This is not covered ? Do you wnat to address this is another PR ?

Remi-Gau: sure but... one would first need to figure out how to even hit that line and this is the most (or one of the most) complex function in the codebase and between the abstract single letter variable and the nested debug statement, I kinda ran out of energy to refactor it to understand how to even get to that warning

so if anyone feels like trying to cover it: by all means be my guest

bthirion: Let's say it's for another PR ;-)

PR #5266, opened 2025-03-22
[MAINT] start deprecation of fetch_neurovault_motor_task

Remi-Gau:
- Closes none

- fetch_neurovault_motor_task is always used to fetch data from neurovault to load the same single image we ship with nilearn anyway
- should prevent some doc build failure when neurovault does not work (see doc build on main at the moment)

### DIFF
--- a/nilearn/datasets/neurovault.py
+++ b/nilearn/datasets/neurovault.py
@@ -2851,6 +2851,16 @@ def fetch_neurovault_motor_task(data_dir=None, verbose=1):
     """Fetch left vs right button press \
        group :term:`contrast` map from :term:`Neurovault`.
 
+    .. deprecated:: 0.11.2dev
+
+        This fetcher function will be removed in version>0.13.1
+        as it returns the same data
+        as :func:`nilearn.datasets.load_sample_motor_activation_image`.
+
+        Please use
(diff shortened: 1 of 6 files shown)


nilearn/datasets/neurovault.py
```diff
@@ -2851,6 +2851,16 @@ def fetch_neurovault_motor_task(data_dir=None, verbose=1):
     """Fetch left vs right button press \
        group :term:`contrast` map from :term:`Neurovault`.
 
+    .. deprecated:: 0.11.2dev
+
+        This fetcher function will be removed in version>0.13.1
```
man-shu, line 2856:
```suggestion
        This fetcher function will be removed in version>0.13.1
```

Shouldn't we stick to 0.13.0 for all deprecations?

man-shu (commented):
Needs a test for the warning

nilearn/datasets/neurovault.py
Remi-Gau: We could but we have the 2 releases rule, so I was trying to be of for 0.11.1 + 2 releases -> 0.13.1

Remi-Gau:
> Needs a test for the warning

done

man-shu (approved):

PR #5274, opened 2025-03-25
[MAINT] remove dead code

Remi-Gau:
- closes none
- remove dead code flagged by vulture
- move one function used only in testing

### DIFF
--- a/nilearn/image/tests/_testing.py
+++ b/nilearn/image/tests/_testing.py
@@ -26,3 +26,54 @@ def match_headers_keys(source, target, except_keys):
             )
         else:
             assert target.header[key] == source.header[key]
+
+
+def pad_array(array, pad_sizes):
+    """Pad an array with zeros.
+
+    Pads an array with zeros as specified in `pad_sizes`.
+
+    Parameters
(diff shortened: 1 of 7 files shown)


nilearn/image/tests/_testing.py
```diff
@@ -26,3 +26,54 @@ def match_headers_keys(source, target, except_keys):
...
         else:
             assert target.header[key] == source.header[key]
+
+
+def pad_array(array, pad_sizes):
+    """Pad an array with zeros.
+
+    Pads an array with zeros as specified in `pad_sizes`.
+
+    Parameters
```
hndgzkn, line 36: Could the functions in `_testing` module be moved to `conftest` in the same directory and be returned from a fixture?

Remi-Gau: In general (but there are exceptions) we have been using fixtures as shortcuts to setting up data for a test: so in general they return an instance of a nifti, surface image... rather than a function to call

also in general (once again there must be exceptions in the codebase) we stick things in a _testing module when the functions are not shared with another subpackage

PR #5278, opened 2025-03-25
[FIX] update docstrings

Remi-Gau:
- Closes none

- update some doc strings defaults

### DIFF
--- a/nilearn/plotting/img_plotting.py
+++ b/nilearn/plotting/img_plotting.py
@@ -658,8 +658,8 @@ def plot_anat(
     %(cmap)s
         Default=`gray`.
 
-    colorbar : :obj:`bool`, default=False
-        If True, display a colorbar on the right of the plots.
+    %(colorbar)s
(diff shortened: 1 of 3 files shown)


nilearn/plotting/img_plotting.py
```diff
@@ -954,13 +954,13 @@ def plot_roi(
         contours.
 
     %(cmap_lut)s
-        Default=`gist_ncar`.
+        Default = `gist_ncar`.
```
bthirion, line 957:
```suggestion
        Default=`gist_ncar`.
```(for consistencey, I'm not 100%sure that this is the right way to display it)

bthirion (commented):
LGTM otherwise

nilearn/plotting/img_plotting.py
Remi-Gau: I don't think we have brought our nitpicking to that level just yet 😉

PR #5284, opened 2025-03-26
[ENH] run tests with free threaded python

Remi-Gau:
- Closes none

- run tests with free threaded python


emdupre:
>  wondering if we want to run on every PR or less frequently (only on main)

AFAIU free threading is [still experimental](https://docs.python.org/3/howto/free-threading-python.html) so I'm not sure that we need to be running it on every PR ; I'd save that for a feature where we expect strong support / no bugs that would pollute our builds.

From the discussion @man-shu linked it joblib will likely be testing this well before us since they're advocating for its development. So, I'd say we could run these tests intermittently (not sure if you have a schedule in mind) or wait until it's at least made it into a joblib release (it was only added to their CI table after their latest release).

Remi-Gau:
OK I am reading a soft consensus to NOT have it on every PR but more on pushes to main or as part of scheduled workflow

emdupre:
I'd push towards a scheduled workflow, just if we want to anticipate issues that might be on the horizon. Once the next joblib release is cut (@jeromedockes might have more insight into that schedule ?), I think it would make sense to run this more regularly.

jeromedockes:
>  (@jeromedockes might have more insight into that schedule ?)

I don't know but I can ask

I agree there's no rush, it's probably too early to have it in every CI run

bthirion (commented):
LGTM so far.

emdupre (approved):
LGTM, thanks @Remi-Gau !

Remi-Gau:
ok let's merge and manually trigger on main to confirm everything is OK

Remi-Gau:
it works:

https://github.com/nilearn/nilearn/actions/runs/14238035765

Issue #5272, opened 2025-03-25
ensure volume / surface parity of behavior in nilearn.image


Remi-Gau:
In a follow up PR I'd like to go through a clean up of the doc of the whole nilearn.image module to make sure it is explicit that it can handle volume and surface.

I think we should also expand the example to make sure we showcase those functionalities with surface data.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/5224#issuecomment-2735913358_
       
For functions that support both surface / volume
- [x] check that documentation is clear about this
- [ ] update examples to show how to use those functions with surface
- [ ] ensure that if a function can do some operation on a volume, it can also do it on a surface (if this does not make sense for surface, make sure this properly documented)
  - [x] mean_img can take a list of 4D volume image and will compute the mean of each 4D before take the mean of means: currently I do not think an equivalent behavior is implemented for surfaces.
  - [ ] clean_img: https://nilearn.github.io/dev/modules/generated/nilearn.image.clean_img.html
  - [x] binarize_img: https://nilearn.github.io/dev/modules/generated/nilearn.image.binarize_img.html
  - [x] math_img: https://nilearn.github.io/dev/modules/generated/nilearn.image.math_img.html

PR #5290, opened 2025-03-27
[DOC] improve doc nilearn.image

Remi-Gau:
- relates to #5272

- fixes some mention of nilearn surface in the doc strings

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -1621,6 +1621,7 @@ def concat_imgs(
       if ``auto_resample=True``.
 
     - surface images of varying dimensions (i.e., 1D or 2D)
+      but with same number of vertices
 
     Parameters
     ----------
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -538,7 +538,7 @@ def mean_img(
 
     Returns
     -------
-    :class:`~nibabel.nifti1.Nifti1Image`
+    :obj:`~nibabel.nifti1.Nifti1Image` or :obj:`~nilearn.surface.SurfaceImage`
         Mean image.
 
     See Also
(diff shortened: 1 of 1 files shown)


Remi-Gau:
am sure I missed some but trying to fix that with small iterations

nilearn/image/image.py
```diff
@@ -1611,11 +1611,16 @@ def concat_imgs(
...
+
+    - Niimg-like objects of varying dimensions (i.e., 3D or 4D)
+      as well as different 3D shapes and affines,
+      as they will be matched to the first image in the list
+      if ``auto_resample=True``.
 
-    The niimgs list can contain niftis/paths to images of varying dimensions
-    (i.e., 3D or 4D) as well as different 3D shapes and affines, as they
-    will be matched to the first image in the list if auto_resample=True.
+    - surface images of varying dimensions (i.e., 1D or 2D)
```
bthirion, line 1623: Should we make clear that the images have to be associated with the same mesh ?

Remi-Gau: yeah good point because that's implicit at the moment (well it will be made very explicit when users get an error)

Remi-Gau: actually we should run 'check_same_n_vertices' more systematically on surface image operation in this module

also @man-shu, shouldn't we more often check that meshes are the same rather than just check they have same number of vertices?

we check that polymesh have the same parts and same number of vertices but 2 meshes can have same number of vertices and be wildly different: I think this would not be caught at the moment.

we have `assert_polymesh_equal` in nilearn.surface._testing

Remi-Gau: ok quick check and pretty sure that every call of check_same_n_vertices should be replaced by assert_polymesh_equal.

Remi-Gau: outside of the scope of this PR though

PR #5199, opened 2025-03-06
[TEST][MAINT] drop 3.9

Remi-Gau:
just checking if we should anticipate lots of dependencies upgrade in a few months from now

PR #5292, opened 2025-03-27
[TEST][MAINT] testing with python 3.14 (AKA py pi)

Remi-Gau:
- same as https://github.com/nilearn/nilearn/pull/5199
- not meant to be merged, meant to see what's coming on the horizon

### DIFF
--- a/maint_tools/check_gha_workflow.py
+++ b/maint_tools/check_gha_workflow.py
@@ -77,7 +77,15 @@ PAGES_TO_COLLECT = range(1, 30)
 UPDATE_TSV = True
 
 # used by set_python_version to filter jobs by their python version
-EXPECTED_PYTHON_VERSIONS = ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+EXPECTED_PYTHON_VERSIONS = [
+    "3.8",
+    "3.9",
+    "3.10",
+    "3.11",
+    "3.12",
+    "3.13",
+    "3.14",
+]
 
 
 def main(args=sys.argv) -> None:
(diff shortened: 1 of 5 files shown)


Remi-Gau:
seems pillow is has finished cooking for py pie

maint_tools/check_gha_workflow.py
```diff
@@ -77,7 +77,15 @@
 UPDATE_TSV = True
 
 # used by set_python_version to filter jobs by their python version
-EXPECTED_PYTHON_VERSIONS = ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+EXPECTED_PYTHON_VERSIONS = [
+    "3.8",
```
bthirion, line 81: Why do we have P3.8 here ?

Remi-Gau: good question

this maintenance script will be used to check past test runs to see how long they take sorted by OS and python version (see https://nilearn.github.io/dev/ci.html#check-gha-workflow-yml) to plotly graphs like this

test_runs_timing

the content of really old CI runs (those done with python 3.8) has been purged but their metadata regarding how long they took is still here, so that's why it's still here.

bthirion (approved):
LGTM so far.

Remi-Gau:
I suggest merging this one much later: just before the 0.13.0 release.

Reason: this way we do not run our CI one extra version of python until then --> faster CI.

Remi-Gau:
3.14 is officially out so it may be time to think about merging this one...

maint_tools/check_gha_workflow.py
```diff
@@ -82,13 +82,15 @@
     "3.11",
     "3.12",
     "3.13",
-    "3.13t",
+    "3.14",
+    "3.14t",
```
Remi-Gau, line 87:
```suggestion
    "3.13t",
    "3.14t",
```

PR #5293, opened 2025-03-27
[MAINT] check empty images

Remi-Gau:
- Closes none
- Relates to https://github.com/nilearn/nilearn/issues/4538

- check behavior of maskers when input data is empty

### DIFF
--- a/nilearn/decomposition/tests/test_base.py
+++ b/nilearn/decomposition/tests/test_base.py
@@ -109,6 +109,7 @@ def test_fast_svd(n_features, rng):
     )
 
 
+@pytest.mark.timeout(0)
 @pytest.mark.parametrize(
     "n_components,reduction_ratio,expected_shape_0",
     [
--- a/nilearn/plotting/tests/test_img_plotting/test_plot_prob_atlas.py
+++ b/nilearn/plotting/tests/test_img_plotting/test_plot_prob_atlas.py
@@ -8,6 +8,7 @@ from nibabel import Nifti1Image
 from nilearn.plotting import plot_prob_atlas
 
 
+@pytest.mark.timeout(0)
 @pytest.mark.parametrize(
     "params",
     [


nilearn/_utils/estimator_checks.py
```diff
@@ -589,6 +590,25 @@ def check_masker_refit(estimator):
...
+        imgs = Nifti1Image(data, _affine_eye())
+    else:
+        imgs = _make_surface_img()
+        data = {
+            part: np.empty(imgs.data.parts[part].shape)
+            for part in imgs.data.parts
+        }
+        imgs = SurfaceImage(imgs.mesh, data)
+
+    estimator.fit(imgs)
```
Remi-Gau, line 609: this does not seem to fail for our maskers but sklearn's ones fail when fitting empty arrays

what do we want to do ?

bthirion: Sorry, I need more explanations: why do we need to test on empty arrays ?

Remi-Gau: yeah sorry for the lack of context

sklearn are supposed to throw an ValueError when fitting empty array, trying to see how our estimators behave when we try to do the same

https://github.com/scikit-learn/scikit-learn/blob/0dbbac961ee251309aa7ec6deee658da83b0eaf0/sklearn/utils/estimator_checks.py#L2292-L2312

```
def check_estimators_empty_data_messages(name, estimator_orig):
    e = clone(estimator_orig)
    set_random_state(e, 1)


    X_zero_samples = np.empty(0).reshape(0, 3)
    # The precise message can change depending on whether X or y is
    # validated first. Let us test the type of exception only:
    err_msg = (
        f"The estimator {name} does not raise a ValueError when an "
        "empty data is used to train. Perhaps use check_array in train."
    )
    with raises(ValueError, err_msg=err_msg):
        e.fit(X_zero_samples, [])


    X_zero_features = np.empty(0).reshape(12, 0)
    # the following y should be accepted by both classifiers and regressors
    # and ignored by unsupervised models
    y = _enforce_estimator_tags_y(e, np.array([1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0]))
    msg = r"0 feature\(s\) \(shape=\(\d*, 0\)\) while a minimum of \d* is required."
    with raises(ValueError, match=msg):
        e.fit(X_zero_features, y)
```

Remi-Gau: my hunch is that we should do the same: throw errors on emtpy image for fit and trasnform

bthirion: Sounds right.

bthirion:
The failure seems unrelated, right ?

bthirion (approved):
LGTM, thx.

Remi-Gau:
- blocked by #5403 

will be better to also implement this on decoders so let's wait for  #5403  to be done

nilearn/_utils/niimg_conversions.py
```diff
@@ -284,6 +284,12 @@ def check_niimg(
     """
     from ..image import new_img_like  # avoid circular imports
 
+    if not isinstance(niimg, (str, Path, SpatialImage)):
+        raise TypeError(
+            "input should be a NiftiLike object. "
+            f"Got: {niimg.__class__.__name__}"
+        )
```
Remi-Gau: doc string said that type error should be raised if input is not niimg like but the code did not match this: turns out that one could pass numpy arrays to be fitted to some of our estimators and it would only lead to very obscure errors.

this should help

.github/workflows/testing_minimum.yml
```diff
@@ -16,7 +16,6 @@ on:
     push:
         branches:
         -   main
-    pull_request:
```
Remi-Gau: - should not have been merged in #5427

Remi-Gau:
@nilearn/core 

FYI I think I am seeing massive slow down in the test suite and examples build time since this PR was merged.
It'd be good to confirm and troubleshoot this before the next release.

I am not seeing any spikes in terms of time for that PR on the few benchmarks that we have: https://nilearn.github.io/benchmarks

It'd be good to create a "regression benchmark":
- create a minimum benchmark to demonstrate the slowing down
- fix the issue
- keep that benchmark in our benchmark suite to prevent this from happening  
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

bthirion:
oops, thx for the heads up.

Remi-Gau:
ok used one of our maint_tools script (https://github.com/nilearn/nilearn/blob/main/maint_tools/check_gha_workflow.py) to confirm that the run time of the test suite of the last PR is longer than an average hollywood movie and the doc build is getting close any lord the rings movie...

newplot

it seems to affect all OS and all python and also seems to affect runs with our oldest supported dependencies, so I am pretty sure it comes from this PR.

Remi-Gau:
also given that the doc build was affected as well, I doubt it's due to a changes in the test suite or how the test suite is run

nilearn/_utils/niimg_conversions.py
```diff
@@ -329,6 +352,9 @@ def check_niimg(
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
 
+    if safe_get_data(niimg).size == 0:
+        raise ValueError("The image is empty.")
```
Remi-Gau, line 356: commenting this part definitely sped things up

will open a PR to remove this so we can see if things get better

Remi-Gau: OK test suite time is back in its normal range when removing this.

image

The problem is that this line was the core of this PR to try to create a way for our estimators to replace sklearn behavior of failing when working with empty data.

For now I will say that we are not implementing it for performance reasons.

Remi-Gau:
side effect of this performance issue: it slowed down a lot of tests that would then fail because they would go past the timeout limit

PR #5299, opened 2025-03-31
[ENH][DOC] clean up doc strings low level functions

Remi-Gau:
- Closes none

- ensure consistency between doc strings, types and tests for several low level functions mostly for checking niimg

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -137,17 +137,18 @@ def load_niimg(
     img_data = _get_data(loaded_niimg)
     target_dtype = _get_target_dtype(img_data.dtype, dtype)
 
-    if target_dtype is not None:
-        copy_header = loaded_niimg.header is not None
-        new_niimg = new_img_like(
-            loaded_niimg,
-            img_data.astype(target_dtype),
-            loaded_niimg.affine,
-            copy_header=copy_header,
-        )
-        if copy_header:
-            loaded_niimg.header.set_data_dtype(target_dtype)
+    if target_dtype is None:
+        return loaded_niimg
 
+    copy_header = loaded_niimg.header is not None
+    new_niimg = new_img_like(
+        loaded_niimg,
+        img_data.astype(target_dtype),
+        loaded_niimg.affine,
+        copy_header=copy_header,
+    )
+    if copy_header:
+        new_niimg.header.set_data_dtype(target_dtype)
     return new_niimg
 
 


nilearn/_utils/niimg.py
```diff
@@ -108,40 +114,41 @@ def load_niimg(niimg, dtype=None):
...
-                niimg,
-                _get_data(niimg).astype(dtype),
-                niimg.affine,
-                copy_header=True,
-            )
-            niimg.header.set_data_dtype(dtype)
-        else:
-            niimg = new_img_like(
-                niimg, _get_data(niimg).astype(dtype), niimg.affine
-            )
```
Remi-Gau, line 142: I don't think that is ever used to load non spatial images so they always have a header

nilearn/_utils/helpers.py
```diff
@@ -218,19 +219,30 @@ def wrapper(*args, **kwargs):
     return _remove_params
 
 
-def stringify_path(path):
+T = TypeVar("T")  # Generic type variable for non-path-like inputs
+
+
+@overload
+def stringify_path(path: Union[str, os.PathLike[str]]) -> str: ...
```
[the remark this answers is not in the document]
Remi-Gau: I have added some doc strings above that section.
But pretty much this puts into type annotations what the return part of the doc string says.

nilearn/masking.py
```diff
@@ -52,6 +53,20 @@ class _MaskWarning(UserWarning):
 warnings.simplefilter("always", _MaskWarning)
 
 
+@overload
```
bthirion, line 56: I don't see that rendered in the doc. I don't yet understand how this work...

nilearn/_utils/niimg_conversions.py
```diff
@@ -218,13 +241,47 @@ def iter_check_niimg(
...
+        SpatialImageLike,
+        Iterable[SpatialImageLike],
+    ],
+    ensure_ndim: Optional[int] = ...,
+    atleast_4d: bool = ...,
+    dtype: Optional[DTypeLike] = ...,
+    return_iterator: Literal[True] = ...,
+    wildcards: bool = ...,
+) -> Iterable[SpatialImage]: ...
+
```
bthirion, line 276: We need to have this overloading ?

nilearn/_utils/niimg_conversions.py
```diff
@@ -240,9 +297,9 @@ def check_niimg(
         If it is an object, check if the affine attribute present and that
         nilearn.image.get_data returns a result, raise TypeError otherwise.
 
-    ensure_ndim : integer {3, 4}, optional
-        Indicate the dimensionality of the expected niimg. An
-        error is raised if the niimg is of another dimensionality.
+    ensure_ndim : integer {3, 4, None}, default=None
+        Indicate the dimensionality of the expected niimg.
+        An error is raised if the niimg is of another dimensionality.
```
bthirion, line 303: If I'm not mistaken, these docstrings are not rendered, right ? How do we check  them ?

nilearn/_utils/niimg.py
```diff
@@ -108,37 +114,42 @@ def load_niimg(niimg, dtype=None):
 
     Returns
     -------
-    img : image
+    img : spatialimages.SpatialImage
         A loaded image object.
     """
-    from ..image import new_img_like  # avoid circular imports
+    from nilearn.image import new_img_like  # avoid circular imports
```
hndgzkn, line 120: This could be imported a bit down in the code, when/if necessary

nilearn/_utils/niimg.py
```diff
@@ -108,37 +114,42 @@ def load_niimg(niimg, dtype=None):
...
         raise TypeError(
             "Data given cannot be loaded because it is"
             " not compatible with nibabel format:\n"
-            + repr_niimgs(niimg, shorten=True)
+            + repr_niimgs(tmp, shorten=True)
         )
 
-    img_data = _get_data(niimg)
-    target_dtype = _get_target_dtype(img_data.dtype, dtype)
+    assert isinstance(loaded_niimg, spatialimages.SpatialImage)
```
hndgzkn, line 135: nibabel already has [_stringify_path](https://github.com/nipy/nibabel/blob/b5ecf482648aac5c77630d0d65555f043fc5d757/nibabel/filename_parser.py#L26) and nibabel's `load` function calls that. Could this part be written differently? 

```
if isinstance(niimg, spatialimages.SpatialImage):
    loaded_niimg = tmp
elif isinstance(niimg, [path, os.PathLike]):
    load(niimg)
else:
     raise TypeError(
            "Data given cannot be loaded because it is"
            " not compatible with nibabel format:\n"
            + repr_niimgs(tmp, shorten=True)
        )
```

We have same number of conditions in if, we do not need the `stringify_path` and `assert` lines.

nilearn/_utils/niimg.py
hndgzkn, line 120: This could be imported a bit down in the code, when/if necessary. It if initially used after the first return statement.

nilearn/_utils/niimg.py
```diff
@@ -4,12 +4,15 @@
 import gc
 from copy import deepcopy
 from pathlib import Path
+from typing import Union
 from warnings import warn
 
 import numpy as np
 from nibabel import is_proxy, load, spatialimages
```
hndgzkn, line 11: Instead of importing whole module `spatialimages`, we can be specific on what is used?

nilearn/_utils/niimg_conversions.py
```diff
@@ -282,73 +339,81 @@ def check_niimg(
         iter_check_niimg, check_niimg_3d, check_niimg_4d
 
     """
-    from ..image import new_img_like  # avoid circular imports
+    # avoid circular imports
+    from nilearn.image import concat_imgs, new_img_like
```
hndgzkn, line 343: May be this function should also be in `nilearn.image` module?

Remi-Gau:
will rework on this from scratch later

Issue #5300, opened 2025-04-01
consider adding __eq__ methods to surface classes


Remi-Gau:
I think checking equality can be done through a class method. Otherwise seems ok to me.

_Originally posted by @hndgzkn in https://github.com/nilearn/nilearn/pull/5296#pullrequestreview-2731788034_

hndgzkn:
@Remi-Gau If we are going to do this, I think it would be better to do it before next release. Because https://github.com/nilearn/nilearn/pull/5296 adds 3 public functions, which should be deprecated if we implement equality through class methods.

Remi-Gau:
yeah but they are not public as 'in the doc public' (we need to find better terms for this) so it'd be fine.

Remi-Gau:
@hndgzkn 
trying this in #5310 but I don't think this will be sufficient to replace the utility functions in surface.utils that do more than check if meshes, polymehses, surface images are equal: they also raise errors explaining why they are not equal

PR #5304, opened 2025-04-01
[FIX] throw warning when meshes are unequal but have compatible number of vertices

### LINKED ISSUE #5303: use less stringent check on meshes
- We should partly revert https://github.com/nilearn/nilearn/pull/5296 as it leads to failure of the doc build

In many places, we have made it simple to work with 'related' meshes:
 
- for example plot a surface image that uses a 'pial' mesh but on background surface image with an 'inflated' mesh

For a similar reason, the following is crashing on main

```python
    fsaverage_meshes = load_fsaverage()
    surface_stat_image = SurfaceImage.from_volume(
        mesh=fsaverage_meshes["pial"],
        volume_img=img,
    )

    fsaverage = load_fsaverage("fsaverage5")
    destrieux = fetch_atlas_surf_destrieux()
    destrieux_atlas = SurfaceImage(
        mesh=fsaverage["inflated"],
        data={
            "left": destrieux["map_left"],
            "right": destrieux["map_right"],
        },
    )

    mask = destrieux_atlas
    for part in mask.data.parts:
        mask.data.parts[part] = mask.data.parts[part] == 34

    masker = SurfaceMasker(mask)
    masker.fit_transform(surface_stat_image)
```

Remi-Gau:
- Closes #5303

- check if meshes are equal in most cases and fall back to checking if they have the same number of vertices if that fails

### DIFF
--- a/nilearn/surface/utils.py
+++ b/nilearn/surface/utils.py
@@ -1,6 +1,11 @@
 """Utilities for surface mesh, data, images."""
 
+from warnings import warn
+
 import numpy as np
+
+from nilearn._utils.exceptions import MeshDimensionError
+from nilearn._utils.logger import find_stack_level
 
 
 def assert_polydata_equal(data_1, data_2):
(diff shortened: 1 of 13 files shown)


hndgzkn (commented):
I think we can define mesh comparison as a class method rather then putting it in utils. My suggestion is something similar to some of the previous reviews.

We have some objects like `SurfaceImage`, `SurfaceMesh`, `PoliMesh`, etc. We define those classes/objects, but we do not take advantage of the facilities that object oriented design provides us.

There are cases we either make some comparisons or an operation that require only the data that is contained already in that object. (`mean`, `get_data`, `equals`, etc.) I think for such cases it would be good to consider defining/implementing those operations in the class itself rather than in utils.

Otherwise we create many utility functions difficult to maintain and also this requires multiple imports where we actually have the object itself that can do the operation.

Otherwise, I think it is Ok.

nilearn/surface/utils.py
man-shu, line 1: Just one codecov complaint here.

man-shu (approved):
Otherwise LGTM!

bthirion (approved):
LGTM, thx.

Remi-Gau:
> I think we can define mesh comparison as a class method rather then putting it in utils. My suggestion is something similar to some of the previous reviews.
> 
> We have some objects like `SurfaceImage`, `SurfaceMesh`, `PoliMesh`, etc. We define those classes/objects, but we do not take advantage of the facilities that object oriented design provides us.
> 
> There are cases we either make some comparisons or an operation that require only the data that is contained already in that object. (`mean`, `get_data`, `equals`, etc.) I think for such cases it would be good to consider defining/implementing those operations in the class itself rather than in utils.
> 
> Otherwise we create many utility functions difficult to maintain and also this requires multiple imports where we actually have the object itself that can do the operation.
> 
> Otherwise, I think it is Ok.

- will tag https://github.com/nilearn/nilearn/issues/5300 for this comment, but I'd prefer to tackle this one in another PR

PR #5240, opened 2025-03-17
[ENH] Refactor nilearn.plotting - create plotting.matrix package

hndgzkn:
- Closes None

Changes proposed in this pull request:

- move `plotting.matrix_plotting` to `plotting.matrix.matrix_plotting`
- move `plotting.tests.test_matrix_plotting` to `plotting.matrix.tests.test_matrix_plotting`

PR #5305, opened 2025-04-01
[ENH] Refactor `nilearn.plotting.matrix` to separate utility and backend functions

hndgzkn:
- Closes None, but related to:
  - https://github.com/nilearn/nilearn/pull/5240

- move common utility functions in `nilearn.plotting.matrix.matrix_plotting` to `nilearn.plotting.matrix._utils`
- move common backend utility functions in `nilearn.plotting.matrix.matrix_plotting` to `nilearn.plotting.matrix._backend`
- move matplotlib backend utility functions in `nilearn.plotting.matrix.matrix_plotting` to `nilearn.plotting.matrix._matplotlib_backend`
- move/separate corresponding test functions in `nilearn.plotting.matrix.tests.test_matrix_plotting` to `nilearn.plotting.matrix.tests.test_backend` and `nilearn.plotting.matrix.tests.test_matplotlib_backend`

> This PR only moves code between modules, the actual code is not changed.


Remi-Gau (approved):

nilearn/plotting/matrix/_backend.py
man-shu, line 1: I am wondering if we really need this file.

IIUC we would like to separate things that need matplotlib vs. the ones that don't. 
So even though the functions here only indirectly need matplotlib, I think they still qualify to be under `_matplotlib_backend.py`?

bthirion: No strong opinion: could go to `_matplotlib_backend` indeed.

nilearn/plotting/matrix/_matplotlib_backend.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+    if axes.get_position().ymin < 1.1 * xlabel_height:
+        # we need to move it over
+        new_position = axes.get_position()
+        new_position.y0 = 1.1 * xlabel_height  # pad a little
+        axes.set_position(new_position)
+
+
+def _sanitize_figure_and_axes(figure, axes):
+    """Help for plot_matrix."""
```
bthirion, line 83: Could you make it a bit more specific ?

nilearn/plotting/matrix/tests/test_backend.py
```diff
@@ -0,0 +1,62 @@
...
+from nilearn.plotting.matrix.tests.test_matrix_plotting import VALID_TRI_VALUES
+
+VALID_REORDER_VALUES = (True, False, "single", "complete", "average")
+
+
+##############################################################################
+# Some smoke testing for graphics-related code
+
+
+def test_sanitize_labels():
```
bthirion, line 20: please add docstring

nilearn/plotting/matrix/tests/test_backend.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@pytest.mark.parametrize("tri", VALID_TRI_VALUES)
+def test_sanitize_tri(tri):
+    _sanitize_tri(tri)
+
+
+@pytest.mark.parametrize("tri", [None, "foo", 2])
+def test_sanitize_tri_error(tri):
+    with pytest.raises(
```
bthirion, line 40: docstring

nilearn/plotting/matrix/tests/test_backend.py
```diff
@@ -0,0 +1,62 @@
...
+def test_sanitize_reorder(reorder):
+    if reorder is not True:
+        assert _sanitize_reorder(reorder) == reorder
+    else:
+        assert _sanitize_reorder(reorder) == "average"
+
+
+@pytest.mark.parametrize("reorder", [None, "foo", 2])
+def test_sanitize_reorder_error(reorder):
+    with pytest.raises(
```
bthirion, line 59: docstring

nilearn/plotting/matrix/tests/test_matplotlib_backend.py
```diff
@@ -0,0 +1,33 @@
...
+from nilearn.plotting.matrix._matplotlib_backend import (
+    _sanitize_figure_and_axes,
+)
+
+
+@pytest.mark.parametrize(
+    "fig,axes", [("foo", "bar"), (1, 2), plt.subplots(1, 1, figsize=(7, 5))]
+)
+def test_sanitize_figure_and_axes_error(fig, axes):
+    with pytest.raises(
```
bthirion, line 13: docstring

bthirion (commented):
Can you increase coverage ?

nilearn/plotting/matrix/_backend.py
hndgzkn: We can decide its place depending on the answer for the question: 

If we implement a `plotly` or another backend one day, would these functions be used by the new backend, or these serve only for the case where the backend is `matplotlib`?

hndgzkn:
@bthirion In this PR, I just restructure the code, and try not to change it unless necessary for restructuring. But once restructured, I'll go over the code in a follow-up PR to fix the problems.

For coverage, I cannot understand the reason why the coverage decreases, but that is also something I would prefer to do in a follow-up PR. I'll take note of both  if that is Ok with you

nilearn/plotting/matrix/_backend.py
man-shu: I don't think we would need another backend for matrix plotting other than `matplotlib`.

I haven't looked at the code super deeply but my understanding is that the issue of two different backends arises only when we plot surface images...

hndgzkn: I had thought of creating a consistent structure for plotting subpackages in terms of backend separation. However if it is a strong possibility that no other backend will be implemented for matrix plotting functions, I agree, I can move these to `_matplotlib_backend.py`.

Remi-Gau:
> for coverage, I cannot understand the reason why the coverage decreases

pretty sure this is a fluke from codecov

I usually confirm this by getting an html coverage report locally and checking it

We should probably generate one for our CI, so we can cross check things manually in case of doubt.

But I agree that just moving functions should not affect coverage.

Remi-Gau:
same for the doc string: I would keep this PR, a move only PR and then we can fix doc strings and other things

man-shu (approved):
LGTM!

PR #5311, opened 2025-04-02
[FIX] fix filenames when saving GLM to BIDS

Remi-Gau:
- Closes but should fix doc build on main

- filenames for nifti / gifti was not properly generated (ended in .json)

### DIFF
--- a/examples/05_glm_second_level/plot_oasis.py
+++ b/examples/05_glm_second_level/plot_oasis.py
@@ -32,11 +32,16 @@ Note that more power would be obtained from using a larger sample of subjects.
 # %%
 # Load Oasis dataset
 # ------------------
-from nilearn import datasets, plotting
+from nilearn import plotting
+from nilearn.datasets import (
+    fetch_icbm152_2009,
+    fetch_icbm152_brain_gm_mask,
+    fetch_oasis_vbm,
+)
 
 n_subjects = 100  # more subjects requires more memory
 
-oasis_dataset = datasets.fetch_oasis_vbm(
+oasis_dataset = fetch_oasis_vbm(
     n_subjects=n_subjects,
 )
 gray_matter_map_filenames = oasis_dataset.gray_matter_maps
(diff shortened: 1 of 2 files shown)


examples/05_glm_second_level/plot_oasis.py
Remi-Gau, line 1: touching this example that is failing on main to make sure it builds in this PR

nilearn/glm/_base.py
```diff
@@ -546,6 +536,16 @@ def _generate_statistical_maps(
...
 
+        fields = {
+            "prefix": prefix,
+            "suffix": "statmap",
+            "extension": "nii.gz",
+            "entities": deepcopy(entities),
+        }
+
+        if generate_bids_name:
+            fields["prefix"] = None
```
Remi-Gau, line 547: values of fields would be updated later in the loop and lead to non sense on further iterations

man-shu: So IIUC, putting the dictionary in the loop resets all the fields?

nilearn/glm/_base.py
```diff
@@ -546,6 +536,16 @@ def _generate_statistical_maps(
         # Override automatic detection with explicit type if provided
         stat_type = contrast_types.get(contrast_name, stat_type)
 
+        fields = {
+            "prefix": prefix,
+            "suffix": "statmap",
+            "extension": "nii.gz",
+            "entities": deepcopy(entities),
```
hndgzkn, line 543: Do we really need deep copy of entities? As far as I see, `entities` is just  used to extract filename. May be I'm missing something.

hndgzkn (approved):
A part from the comment seems ok to me.

nilearn/glm/_base.py
Remi-Gau: yup exactly

nilearn/glm/_base.py
Remi-Gau: this is mostly to avoid side effects as entities needs to be modified in the loop in this function but we don't want it to be modified out of it: without the copy, entities would also the be changed in other places in the caller function

once again maybe I'm just to dumb to understand classes, OOP and references but this solved my issue

hndgzkn: Not really related to OOP, but yes, I missed the fact that entities change, sorry.

Remi-Gau: well in python everything is an object, so everything is OOP... 😜

PR #5316, opened 2025-04-03
[ENH] upload HTML report of code coverage

Remi-Gau:
- Closes none

- generate HTML report of code coverage and upload them after running tests

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -9,6 +9,16 @@
 #
 #    - ``[skip test]`` skip this workflow.
 #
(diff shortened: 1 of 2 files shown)


Remi-Gau:
in case some people have never used this:

when running the tests locally with tox this puts all the html files in ``htmlcov`` to have browsable reports of all the lines that were covered

image

Remi-Gau:
@bthirion @hndgzkn @man-shu 
I would suggest double checking those reports to determine if the reports from codecov may be 'wrong'

hndgzkn (approved):
Seems ok to me.

PR #5327, opened 2025-04-11
[FIX] update API

Remi-Gau:
- Closes none

- realized that some subpakcage were not listed in the public API
- also simplify one utility snippet that list all the functions, modules, classes that should be documented


man-shu (approved):
LGTM!

PR #5329, opened 2025-04-14
[FIX] Github action for running benchmarks

man-shu:
- Closes none but fixes the benchmarks action: https://github.com/nilearn/nilearn/actions/runs/14433876048/job/40471964529#step:7:10

- need to cd into asv_benchmarks and then run everything
- also allow running via github UI rather than a commit on the main

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -56,10 +56,11 @@ jobs:
         steps:
         -   uses: actions/checkout@v3
         -   uses: actions/setup-python@v3
-        -   name: Install asv
+        -   name: Install asv and switch to asv_benchmarks directory
             run: |
                 pip install --upgrade pip
                 pip install asv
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -43,21 +43,24 @@ on:
     # everday at 10pm UTC
     -   cron: 0 22 * * 0
 
+    workflow_dispatch:  # This enables manual triggering from GitHub UI
+
 jobs:
     benchmark:
         if: >-
-            (contains(github.event.head_commit.message, '[bm]') ||
+            (github.event_name == 'workflow_dispatch' ||
```
man-shu, line 51: Unrelated, but I think it's better to trigger via Github UI rather than commit to main.
Didn't know about this before...

bthirion (approved):
LGTM.

PR #5332, opened 2025-04-14
[FIX] allow generation of glm report with no contrasts

Remi-Gau:
- Closes none

- fixes minor bug that would prevent the generation of GLM report when no contrast was passed


man-shu (approved):
LGTM!

PR #5335, opened 2025-04-15
Revert "[FIX] Github action for running benchmarks"

man-shu:
Reverts nilearn/nilearn#5329.

This still fails, changing directory didnt work for some reason


man-shu:
Ok no this is not what I want to do. Will open another PR to fix this.

PR #5320, opened 2025-04-04
[ENH] use bids terminology in output of parse_bids_filename

Remi-Gau:
- Closes #5308 

Changes proposed in this pull request:

-
-

PR #5339, opened 2025-04-15
[FIX] use legacy=False to parse FLM filenames inputs as bids files

Remi-Gau:
- follow up to #5320

- use legacy=False to parse FLM filenames inputs as bids files

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -900,7 +900,7 @@ class FirstLevelModel(BaseGLM):
             self._reporting_data["run_imgs"][run_idx] = {}
             if isinstance(run_img, (str, Path)):
                 self._reporting_data["run_imgs"][run_idx] = (
-                    parse_bids_filename(run_img)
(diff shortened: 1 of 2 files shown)


nilearn/glm/first_level/first_level.py
```diff
@@ -900,7 +900,7 @@ def fit(
             self._reporting_data["run_imgs"][run_idx] = {}
             if isinstance(run_img, (str, Path)):
                 self._reporting_data["run_imgs"][run_idx] = (
-                    parse_bids_filename(run_img)
+                    parse_bids_filename(run_img, legacy=False)
```
man-shu, line 903: Just trying to understand, would this mean that if someone has older BIDS file structure our GLM object won't be able to detect those files?

Remi-Gau: nope 
the legacy in this case refers to the structure of the output (the name of the keys in the output dict), not the input

man-shu: ah ok

man-shu (approved):
LGTM!

PR #5340, opened 2025-04-15
[WIP] try running benchmark on self-hosted runner

man-shu:
- Closes none but tries to run the benchmarks workflow on self-hosted runner because very slight changes in running machine cause weird parameterization in the results:

For example on https://nilearn.github.io/benchmarks/ right now, previous results got wiped because of a small version change in the OS of github hosted runner
Screenshot 2025-04-15 at 4 54 58 PM

- currently initiated a runner on an HPC at INRIA: https://github.com/organizations/nilearn/settings/actions/runners
- added `runs-on: self-hosted`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -53,8 +53,7 @@ jobs:
         if: >-
             ((github.event_name == 'push' && contains(github.event.head_commit.message, '[bm]')) ||
             github.event_name == 'workflow_dispatch' ||
-            github.event_name == 'schedule') &&
-            github.repository == 'nilearn/nilearn'
+            github.event_name == 'schedule')
         # TODO bump to 24.04 when bumping minimum python to 3.10
         runs-on: self-hosted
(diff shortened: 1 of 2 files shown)


man-shu:
@Remi-Gau If this works we could even do this for building docs

man-shu:
We probably need to use an older version of the action-runner: https://github.com/man-shu/nilearn/actions/runs/14473810650/job/40594332328#step:2:17

man-shu:
Ok so this can work in theory: https://github.com/man-shu/nilearn/actions/runs/14491849971/job/40649877265

But not super convenient because:

1. it won't work on drago because the Ubuntu version is outdated (18.04 with glibc 2.27, [actions-runner](https://github.com/actions/runner) needs > 2.28)
2. it can work on margaret which is a slurm-based cluster. So one would always have to request a node manually and then start the actions-runner service which would then listen to the any requests sent via Github

Will try to find another solution. Closing for now.

PR #5322, opened 2025-04-08
[ENH] Refine `nilearn.plotting.surface`

hndgzkn:
- Closes None, but relates to:
  - #5234 
  - #5253 
  - #5276
  - #5282 

Changes proposed in this pull request:

- Add docstring for backend modules
- Set default value for parameters not implemented in one of the engines to None
- Add parameter check for the parameters not implemented for matplotlib
- Rename backend utility functions and remove engine name from function as they are in their own module and not used elsewhere
- Refactor tests accordingly
- Move `plotting.displays._figures.SurfaceFigure` tests to `plotting.displays.tests.test_figures`
- Move `plotting.displays._figures.PlotlySurfaceFigure` tests to `plotting.surface.tests.test_plotly_surface_figure`
- Remove leading `_` from functions defined in `plotting.surface._backend` but used in other modules.
- Add missing docstrings

PR #5282, opened 2025-03-26
[ENH] Refactor plotting.surface.plot_surf function to separate backend code for matplotlib and plotly

hndgzkn:
- Closes None, but relates to:
  - #5234 
  - #5253 
  - #5276

Changes proposed in this pull request:

- create `BaseSurfaceBackend`, `MatplotlibBackend` and `PlotlyBackend` classes
- refactor `nilearn.plotting.surface.plot_surf`:
   - move `_plot_surf_plotly` code to `PlotlyBackend._plot_surf`
   - move `_plot_surf_matplotlib` code to `MatplotlibBackend._plot_surf`
   - move common code in `nilearn.plotting.surface.plot_surf` to `BaseSurfaceBackend.plot_surf`
   - move matplotlib specific tests from `nilearn.plotting.surface.tests.test_surf_plotting` to `nilearn.plotting.surface.tests.test_matplotlib_backend`
   - move plotly specific tests from `nilearn.plotting.surface.tests.test_surf_plotting` to `nilearn.plotting.surface.tests.test_plotly_backend`
   
> This PR basically only moves code between modules, the actual code is not changed. The added parts are:
>  - `nilearn.plotting._utils.engine_warning` function
>  - `nilearn.plotting.surface._backend.get_surface_backend` function
> - `nilearn.plotting.surface.tests.test_backend` module (tests `get_surface_backend` and `BaseSurfaceBackend`)

PR #5276, opened 2025-03-25
[ENH] Refactor plotting.surface package to extract backend utility functions for plotly and matplotlib

hndgzkn:
- Closes None, but relates to:
  - #5234 
  - #5253 

Changes proposed in this pull request:

- move common backend utility functions in `nilearn.plotting.surface.surf_plotting` to `nilearn.plotting.surface._backend`
- move matplotlib backend utility functions in `nilearn.plotting.surface.surf_plotting` to `nilearn.plotting.surface._matplotlib_backend`
- move plotly backend utility functions in `nilearn.plotting.surface.surf_plotting` to `nilearn.plotting.surface._plotly_backend`
- move/separate corresponding test functions in `nilearn.plotting.surface.tests.test_surf_plotting` to `nilearn.plotting.surface.tests.test_backend`, `nilearn.plotting.surface.tests.test_matplotlib_backend` and to `nilearn.plotting.surface.tests.test_plotly_backend`

PR #5253, opened 2025-03-19
ENH Refactor plotting to separate surface related util functions

hndgzkn:
- Closes None, but relates to #5234 

Changes proposed in this pull request:

- move surface related utility functions in `nilearn.plotting._utils` to `nilearn.plotting.surface._utils`
- move tests for those functions from `nilearn.plotting.surface.tests.test_surf_plotting` to `nilearn.plotting.tests.test_utils`

PR #5349, opened 2025-04-23
Enh refactor refine plot surf contours

hndgzkn:
  - #5234 
  - #5253 
  - #5276
  - #5282 
  - #5322

-  Refactor `plotting.surface.surf_plotting.plot_surf_contours` to separate matplotlib code to `_matplotlib_backend`
- Move `_get_faces_on_edge` from `plotting.surface.surf_plotting` to `plotting.surface._utils` as `get_faces_on_edge`
- Use `get_faces_on_edge` from `plotting.surface._utils` in `plotting.displays._figures.PlotlySurfaceFigure` and remove `_get_faces_on_edge` method from `PlotlySurfaceFigure` and its corresponding test
- Add test for `get_faces_on_edge` in `plotly.surface.tests.test_utils`
- Update docstrings

### DIFF
--- a/nilearn/plotting/displays/_figures.py
+++ b/nilearn/plotting/displays/_figures.py
@@ -5,6 +5,7 @@ from scipy import linalg
 from scipy.spatial import distance_matrix
 
 from nilearn._utils.helpers import is_kaleido_installed, is_plotly_installed
+from nilearn.plotting.surface._utils import get_faces_on_edge
 from nilearn.surface import SurfaceImage
 from nilearn.surface.surface import get_data, load_surf_data
 
@@ -277,7 +278,7 @@ class PlotlySurfaceFigure(SurfaceFigure):
             3. Arrange the centroids such in a good order for plotting
         """
         # Mask indicating faces whose centroids will compose the boundary.
-        edge_faces = self._get_faces_on_edge(parc_idx=parc_idx)
+        edge_faces = get_faces_on_edge(faces=self._faces, parc_idx=parc_idx)
 
         # gather the centroids of each face
         centroids = []
@@ -424,29 +425,6 @@ class PlotlySurfaceFigure(SurfaceFigure):
         sorted_vertices.append(centroids[prev_first])
 
         return np.asarray(sorted_vertices)
-
-    def _get_faces_on_edge(self, parc_idx):
-        """Identify which faces lie on the outeredge of the parcellation \
-        defined by the indices in parc_idx.
-
-        Parameters
-        ----------
-        parc_idx : numpy.ndarray, indices of the vertices
-            of the region to be plotted
-        """
-        # count how many vertices belong to the given parcellation in each face
-        verts_per_face = np.isin(self._faces, parc_idx).sum(axis=1)
-
-        # test if parcellation forms regions
-        if np.all(verts_per_face < 2):
-            raise ValueError("Vertices in parcellation do not form region.")
-
-        vertices_on_edge = np.intersect1d(
-            np.unique(self._faces[verts_per_face == 2]), parc_idx
-        )
-        faces_outside_edge = np.isin(self._faces, vertices_on_edge).sum(axis=1)
-
-        return np.logical_and(faces_outside_edge > 0, verts_per_face < 3)
 
     @staticmethod
     def _project_above_face(point, t0, t1, t2, elevation=0.1):
--- a/nilearn/plotting/surface/_backend.py
+++ b/nilearn/plotting/surface/_backend.py
@@ -22,7 +22,10 @@ from warnings import warn
 import numpy as np
 
 from nilearn._utils.param_validation import check_params
-from nilearn.plotting.surface._utils import check_surface_plotting_inputs
+from nilearn.plotting.surface._utils import (
+    check_surface_plotting_inputs,
+    sanitize_hemi_for_surface_image,
+)
 from nilearn.surface import load_surf_data, load_surf_mesh
 from nilearn.surface.surface import (
     FREESURFER_DATA_EXTENSIONS,
@@ -141,6 +144,44 @@ class BaseSurfaceBackend:
             output_file=output_file,
             axes=axes,
             figure=figure,
+        )
+
+    def plot_surf_contours(
+        self,
+        surf_mesh=None,
+        roi_map=None,
+        hemi=None,
+        levels=None,
+        labels=None,
+        colors=None,
+        legend=False,
+        cmap="tab20",
+        title=None,
+        output_file=None,
+        axes=None,
+        figure=None,
+        **kwargs,
+    ):
+        hemi = sanitize_hemi_for_surface_image(hemi, roi_map, surf_mesh)
+        roi_map, surf_mesh, _ = check_surface_plotting_inputs(
+            roi_map, surf_mesh, hemi, map_var_name="roi_map"
+        )
+        check_extensions(roi_map, DATA_EXTENSIONS, FREESURFER_DATA_EXTENSIONS)
+
+        return self._plot_surf_contours(
+            surf_mesh=surf_mesh,
+            roi_map=roi_map,
+            hemi=hemi,
+            levels=levels,
+            labels=labels,
+            colors=colors,
+            legend=legend,
+            cmap=cmap,
+            title=title,
+            output_file=output_file,
+            axes=axes,
+            figure=figure,
+            **kwargs,
         )
 
 
--- a/nilearn/plotting/surface/_matplotlib_backend.py
+++ b/nilearn/plotting/surface/_matplotlib_backend.py
@@ -10,6 +10,7 @@ from warnings import warn
 import numpy as np
 
 from nilearn import DEFAULT_DIVERGING_CMAP
+from nilearn._utils import compare_version
 from nilearn.plotting._utils import (
     get_cbar_ticks,
     get_colorbar_and_data_ranges,
@@ -22,14 +23,20 @@ from nilearn.plotting.surface._backend import (
     check_surf_map,
     check_views,
 )
-from nilearn.surface import load_surf_data
+from nilearn.plotting.surface._utils import get_faces_on_edge
+from nilearn.surface import (
+    load_surf_data,
+    load_surf_mesh,
+)
 
 try:
     import matplotlib as mpl
     import matplotlib.pyplot as plt
     from matplotlib.cm import ScalarMappable
     from matplotlib.colorbar import make_axes
-    from matplotlib.colors import LinearSegmentedColormap, Normalize
+    from matplotlib.colors import LinearSegmentedColormap, Normalize, to_rgba
+    from matplotlib.patches import Patch
+    from mpl_toolkits.mplot3d.art3d import Poly3DCollection
 except ImportError:
     from nilearn.plotting._utils import engine_warning
 
@@ -333,6 +340,18 @@ class MatplotlibSurfaceBackend(BaseSurfaceBackend):
     def name(self):
         return "matplotlib"
 
+    def _check_figure_axes_inputs(self, figure, axes):
+        """Check if the specified figure and axes are matplotlib objects."""
+        if figure is not None and not isinstance(figure, plt.Figure):
+            raise ValueError(
+                "figure argument should be None or a "
+                "'matplotlib.pyplot.Figure'."
+            )
+        if axes is not None and not isinstance(axes, plt.Axes):
+            raise ValueError(
+                "axes argument should be None or a 'matplotlib.pyplot.Axes'."
+            )
+
     def _plot_surf(
         self,
         coords,
@@ -491,6 +510,105 @@ class MatplotlibSurfaceBackend(BaseSurfaceBackend):
             p3dcollec.set_edgecolors(face_colors)
 
         if title is not None:
+            axes.set_title(title)
+
+        return save_figure_if_needed(figure, output_file)
+
+    def _plot_surf_contours(
+        self,
+        surf_mesh=None,
+        roi_map=None,
+        hemi=None,
+        levels=None,
+        labels=None,
+        colors=None,
+        legend=False,
+        cmap="tab20",
+        title=None,
+        output_file=None,
+        axes=None,
+        figure=None,
+        **kwargs,
+    ):
+        self._check_figure_axes_inputs(figure, axes)
+
+        if figure is None and axes is None:
+            figure = self.plot_surf(surf_mesh, hemi=hemi, **kwargs)
+            axes = figure.axes[0]
+        elif figure is None:
+            figure = axes.get_figure()
+        elif axes is None:
+            axes = figure.axes[0]
+
+        if axes.name != "3d":
+            raise ValueError("Axes must be 3D.")
+
+        # test if axes contains Poly3DCollection, if not initialize surface
+        if not axes.collections or not isinstance(
+            axes.collections[0], Poly3DCollection
+        ):
+            _ = self.plot_surf(surf_mesh, hemi=hemi, axes=axes, **kwargs)
+
+        if levels is None:
+            levels = np.unique(roi_map)
+
+        if labels is None:
+            labels = [None] * len(levels)
+
+        if colors is None:
+            n_levels = len(levels)
+            vmax = n_levels
+            cmap = plt.get_cmap(cmap)
+            norm = Normalize(vmin=0, vmax=vmax)
+            colors = [cmap(norm(color_i)) for color_i in range(vmax)]
+        else:
+            try:
+                colors = [to_rgba(color, alpha=1.0) for color in colors]
+            except ValueError:
+                raise ValueError(
+                    "All elements of colors need to be either a"
+                    " matplotlib color string or RGBA values."
+                )
+
+        if not (len(levels) == len(labels) == len(colors)):
+            raise ValueError(
+                "Levels, labels, and colors "
+                "argument need to be either the same length or None."
+            )
+
+        _, faces = load_surf_mesh(surf_mesh)
+        roi = load_surf_data(roi_map)
+
+        patch_list = []
+        for level, color, label in zip(levels, colors, labels):
+            roi_indices = np.where(roi == level)[0]
+            faces_outside = get_faces_on_edge(faces, roi_indices)
+            # Fix: Matplotlib version 3.3.2 to 3.3.3
+            # Attribute _facecolors3d changed to _facecolor3d in
+            # matplotlib version 3.3.3
+            if compare_version(mpl.__version__, "<", "3.3.3"):
+                axes.collections[0]._facecolors3d[faces_outside] = color
+                if axes.collections[0]._edgecolors3d.size == 0:
+                    axes.collections[0].set_edgecolor(
+                        axes.collections[0]._facecolors3d
+                    )
+                axes.collections[0]._edgecolors3d[faces_outside] = color
+            else:
+                axes.collections[0]._facecolor3d[faces_outside] = color
+                if axes.collections[0]._edgecolor3d.size == 0:
+                    axes.collections[0].set_edgecolor(
+                        axes.collections[0]._facecolor3d
+                    )
+                axes.collections[0]._edgecolor3d[faces_outside] = color
+            if label and legend:
+                patch_list.append(Patch(color=color, label=label))
+        # plot legend only if indicated and labels provided
+        if legend and np.any([lbl is not None for lbl in labels]):
+            figure.legend(handles=patch_list)
+            # if legends, then move title to the left
+        if title is None and hasattr(figure._suptitle, "_text"):
+            title = figure._suptitle._text
+        if title:
             axes.set_title(title)
 
         return save_figure_if_needed(figure, output_file)
--- a/nilearn/plotting/surface/_plotly_backend.py
+++ b/nilearn/plotting/surface/_plotly_backend.py
@@ -353,3 +353,21 @@ class PlotlySurfaceBackend(BaseSurfaceBackend):
             plotly_figure.savefig()
 
         return plotly_figure
+
+    def _plot_surf_contours(
+        self,
+        surf_mesh=None,
+        roi_map=None,
+        hemi=None,
+        levels=None,
+        labels=None,
+        colors=None,
+        legend=False,
+        cmap="tab20",
+        title=None,
+        output_file=None,
+        axes=None,
+        figure=None,
+        **kwargs,
+    ):
+        raise NotImplementedError()
--- a/nilearn/plotting/surface/_utils.py
+++ b/nilearn/plotting/surface/_utils.py
@@ -2,6 +2,8 @@
 
 from warnings import warn
 
+import numpy as np
+
 from nilearn._utils.helpers import is_matplotlib_installed, is_plotly_installed
 from nilearn.surface import (
     PolyMesh,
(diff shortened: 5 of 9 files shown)


nilearn/plotting/displays/_figures.py
```diff
@@ -425,29 +426,6 @@ def _get_sorted_edge_centroids(self, parc_idx, elevation=0.1):
...
-        if np.all(verts_per_face < 2):
-            raise ValueError("Vertices in parcellation do not form region.")
-
-        vertices_on_edge = np.intersect1d(
-            np.unique(self._faces[verts_per_face == 2]), parc_idx
-        )
-        faces_outside_edge = np.isin(self._faces, vertices_on_edge).sum(axis=1)
-
-        return np.logical_and(faces_outside_edge > 0, verts_per_face < 3)
-
```
hndgzkn, line 450: This function is the same as `_get_faces_on_edge` in `surf_plotting`. I've moved that one to `nilearn.plotting.surface._utils` and I'm using that one in `PlotlySurfaceFigure`.

nilearn/plotting/surface/_matplotlib_backend.py
```diff
@@ -333,6 +340,18 @@ class MatplotlibSurfaceBackend(BaseSurfaceBackend):
...
+        if figure is not None and not isinstance(figure, plt.Figure):
+            raise ValueError(
+                "figure argument should be None or a "
+                "'matplotlib.pyplot.Figure'."
+            )
+        if axes is not None and not isinstance(axes, plt.Axes):
+            raise ValueError(
+                "axes argument should be None or a 'matplotlib.pyplot.Axes'."
+            )
+
```
hndgzkn, line 354: This function is moved from `surf_plotting` to here changing the contents. It was checking if the specified figure/axes were `PlotlySurfaceFigure` or not which actually does not cover valid values. I've changed comparison to None or matplotlib objects.

nilearn/plotting/surface/_matplotlib_backend.py
```diff
@@ -494,3 +513,102 @@ def _plot_surf(
...
+        title=None,
+        output_file=None,
+        axes=None,
+        figure=None,
+        **kwargs,
+    ):
+        self._check_figure_axes_inputs(figure, axes)
+
+        if figure is None and axes is None:
+            figure = self.plot_surf(surf_mesh, hemi=hemi, **kwargs)
```
hndgzkn, line 536: In the original code, `hemi` is not specified to `plot_surf` at line 536 and 550. The default value for `hemi` in `plot_surf` is `left`. The default value in `plot_surf_contours` is `None`. So by not specifying `hemi`, we are changing the actual specified value. 

I've updated to pass `hemi` to `plot_surf`. But then, the valid `None` value for `hemi` becomes invalid for `plot_surf`. What should be the correct behavior? Currently the tests are failing due to the change made, ie passing `hemi` in `plot_surf`.

nilearn/plotting/surface/_plotly_backend.py
```diff
@@ -353,3 +353,21 @@ def _plot_surf(
...
+        colors=None,
+        legend=False,
+        cmap="tab20",
+        title=None,
+        output_file=None,
+        axes=None,
+        figure=None,
+        **kwargs,
+    ):
+        raise NotImplementedError()
```
hndgzkn, line 373: Shall we implement this for `plotly` as well?

nilearn/plotting/surface/_backend.py
```diff
@@ -143,6 +146,44 @@ def plot_surf(
             figure=figure,
         )
 
+    def plot_surf_contours(
+        self,
+        surf_mesh=None,
+        roi_map=None,
+        hemi=None,
```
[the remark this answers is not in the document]
Remi-Gau: not a 100% sure
view_surf can also accept None for hemi

definitely something that should be cleaned up: also it'd be good to have a single default constant in plotting we can use to define the default hemisphere, to make it easier to keep things consistent across all functions.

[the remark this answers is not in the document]
Remi-Gau: see for example https://github.com/nilearn/nilearn/blob/34ff77ac402c09a6f49284359defa6aa4de9b07f/nilearn/__init__.py#L58-L59

for constant for default colormaps

nilearn/plotting/surface/_matplotlib_backend.py
Remi-Gau: but if it makes your like easier to solve things by implementing a 'global' default constant, go for it.

nilearn/plotting/surface/_plotly_backend.py
Remi-Gau: nope 
pretty sure that for plotly we rely on the figure.add_contours method to do this

so maybe making sure that the error message mention this would be better

Remi-Gau: See 

https://github.com/nilearn/nilearn/blob/34ff77ac402c09a6f49284359defa6aa4de9b07f/examples/01_plotting/plot_3d_map_to_surface_projection.py#L179-L207

hndgzkn: Wouldn't it be better to adopt the same behavior for both engines?

I would suggest to add `engine` parameter to `plot_surf_contours` and implement it for `plotly`. As the implementation already exists, it will be easy I think.

In the long term, I would suggest to create a `matplotlib` equivalent of `PlotlySurfaceFigure`, ie. `MatplotlibSurfaceFigure` that has the same functionality. What do you think? 

It seems a bit strange to me that depending on the engine we have different objects and different way of doing things. I think from users point of view, it should be enough to know the interface, and use interface methods/or functions for all engines. The implementation details should not be an issue for the user.

nilearn/plotting/surface/_matplotlib_backend.py
hndgzkn: > hemi = sanitize_hemi_for_surface_image(hemi, roi_map, surf_mesh)

This still returns `None` in the tests.

The code for `sanitize_hemi_for_surface_image` returns `None` when `hemi=None` and the specified `map` is not `SurfaceImage` and specified `mesh` is not `PolyMesh`.

What should it return? If we specify a value, I think that would fix the tests.

man-shu: > the specified map is not SurfaceImage and specified mesh is not PolyMesh

I think in this case we should not even reach the `sanitize_hemi_for_surface_image`?
We can only think about plotting if the map is SurfaceImage and mesh is PolyMesh.

So I think we should do `check_surface_plotting_inputs` before doing `sanitize_hemi_for_surface_image` in `plot_surf_contours`.

LMK if this makes sense.

hndgzkn: I think both `check_surface_plotting_inputs` and `sanitize_hemi_for_surface_image` requires a review. 

`sanitize_hemi_for_surface_image` does not seem to really sanitize hemi. 

In the tests, for example `plotting/surface/tests/test_matplotlib_backend.py::test_plot_surf_contours_fig_axes`, `in_memory_mesh` (type = nilearn.surface.surface.InMemoryMesh ) and `parcellation` (type=numpy.ndarray) are specified as `surf_mesh` and `roi_map`.  It checks for `PolyMesh` and `SurfaceImage`.

In docstring of `check_surface_plotting_inputs` it says that `surf_map` and `surf_mesh` are of type `numpy.ndarray`, but in the code it checks types for `SurfaceImage` and  `PolyMesh`.

Finally, I would suggest that we make default `hemi='left'`. Because actually as far as I see in the code `hemi=None` is not allowed in any case.

bthirion: My understanding is that the docstrings of `check_surface_plotting_inputs` just reflect what we were doing some time ago (actually a mesh cannot be an array !) so they're plain wrong.

bthirion: Do you have any suggestion for `sanitize_hemi_for_surface_image` ? I don't have any in mind.

hndgzkn: > My understanding is that the docstrings of `check_surface_plotting_inputs` just reflect what we were doing some time ago (actually a mesh cannot be an array !) so they're plain wrong.

I think I've missed the fact that the type definitions were for returned  values. So not that wrong. However, I think it needs clarification.

As far as I understand valid input values for parameters are:

surf_mesh : PolyMesh, numpy.ndarray, None 
surf_map: SurfaceImage, numpy.ndarray, None
bg_map: SurfaceImage, numpy.ndarray, None

And return values are:
surf_mesh: InMemoryMesh, numpy.ndarray
surf_map: numpy.ndarray
bg_map: numpy.ndarray

If this is the case, shall put a control to see if input values are of those types?

hndgzkn: > I'm a bit worried with default hemi = 'left'. We should review whether this default is used in other libraries such as Freesurfer. Otherwise, I'm worried of users plotting right-hemisphere features on left hemisphere without knowing, simply due to the default. If this makes sense, I would consider `both` as a less harmful prior.

I'll define a DEFAULT_HEMI as @Remi-Gau suggested. We can change its value. If both is OK for everyone, I can set default to both.

hndgzkn: > Do you have any suggestion for `sanitize_hemi_for_surface_image` ? I don't have any in mind.

We could define valid inputs here as well.

- Can map be only `SurfaceImage` or `numpy.ndarray` is also acceptable? 
- Can `mesh` be `InMemoryMesh` or it should only be `PolyMesh`?

Besides, `plot_surf_contours` calls `plot_surf`, and in `plot_surf` (for `matplotlib` backend), we have  axes.view_init(elev=elev, azim=azim)` which gets `elev` and `azim` from `_get_view_plot_surf(hemi, view)`.  hemi=None` is not a valid value for this function.

I think either we should define a valid value for `hemi=None` or not allow `hemi=None` in `plot_surf_contours`. In actual code, `plot_surf_contours` changes the `hemi=None` to `hemi='left'` implicitly by not passing `hemi` when calling `plot_surf`. 

`sanitize_hemi_for_surface_image` can return `DEFAULT_HEMI` or `"both"` but not `None`.

nilearn/plotting/surface/_utils.py
```diff
@@ -87,16 +91,16 @@ def _check_bg_map(bg_map, hemi):
...
-    corresponding mesh. If "both" is requested, combine the left and right
-    hemispheres.
+    """Check that a given hemisphere exists in a
+    :obj:`~nilearn.surface.PolyMesh` and return the corresponding mesh. If
+    "both" is requested, combine the left and right hemispheres.
     """
     if hemi == "both":
         return combine_hemispheres_meshes(mesh)
     elif hemi in mesh.parts:
         return mesh.parts[hemi]
```
hndgzkn, line 101:
```suggestion
    elif hemi in ["left", "right"]:
        if hemi in mesh.parts:
            return mesh.parts[hemi]
        else:
            raise ValueError(f"hemi = {=hemi} does not exist in mesh. Avaliable hemispheres are: {mesh.parts.keys()}")
```

hndgzkn: Does this seem reasonable?

nilearn/plotting/surface/_utils.py
```diff
@@ -109,17 +113,18 @@ def check_surface_plotting_inputs(
...
+    surf_map : :obj:`numpy.ndarray`
 
-    surf_mesh : numpy.ndarray
+    surf_mesh : :obj:`numpy.ndarray`
 
-    bg_map : str | pathlib.Path | numpy.ndarray | None
+    bg_map : :obj:`str` | :obj:`pathlib.Path` | :obj:`numpy.ndarray` | None
 
     """
     if surf_mesh is None and surf_map is None:
```
hndgzkn, line 130: How does it sound to call `sanitize_hemi...` here, and return the final `hemi`  with other parameters?

nilearn/plotting/surface/_utils.py
```diff
@@ -87,16 +91,16 @@ def _check_bg_map(bg_map, hemi):
...
     if hemi == "both":
         return combine_hemispheres_meshes(mesh)
     elif hemi in mesh.parts:
         return mesh.parts[hemi]
     else:
-        raise ValueError("hemi must be one of left, right or both.")
+        raise ValueError("hemi must be one of 'left', 'right' or 'both'.")
 
 
 def check_surface_plotting_inputs(
```
hndgzkn, line 106: Could we remove the parameters `map_var_name` and `mesh_var_name`? Is it really important to display to user if it is `roi_map` or `surf_map`? We can simply say `map` and `mesh`?

nilearn/plotting/surface/_matplotlib_backend.py
Remi-Gau: In this case `sanitize_hemi_for_surface_image` would only throw warnings for case like and should be renamed

```python
    if (
        not isinstance(map, SurfaceImage)
        and not isinstance(mesh, PolyMesh)
    ):
        warn(
            category=UserWarning,
            message=(
                f"{hemi=} was passed "
                f"with {type(map)=} and {type(mesh)=}.\n"
                "This value will be ignored as it is only used when "
                "'roi_map' is a SurfaceImage instance "
                "and  / or 'surf_mesh' is a PolyMesh instance."
            ),
            stacklevel=3,
        )
```

Remi-Gau: > I would suggest that we make default `hemi='left'`. Because actually as far as I see in the code `hemi=None` is not allowed in any case.

We should not allow hemi=none for view_img eithr then

nilearn/plotting/surface/_utils.py
Remi-Gau: yup but I think you will want:

` ValueError(f"{hemi=} does not exist in mesh. Avalaible hemispheres are: mesh.parts.keys()}")`

nilearn/plotting/surface/_utils.py
Remi-Gau: sure: but I was just trying to make it easier for users to figure out which parameters they have passed in public functions is "wrong"

nilearn/plotting/surface/_utils.py
Remi-Gau: yes that'd be fine but we do not hemi=None anywhere then sanitize_hemi should now just be function that warns users that hemi is irrelevant if they have passed objects where hemi is not relevant (when they pass directly a mesh and an array for the data, instead of a polymesh + polydata or surface image...)

nilearn/plotting/surface/_utils.py
hndgzkn: `mesh_var_name` only takes value `surf_mesh` in the code. Map can be either `roi`, `surf` or `stat`. If you think it is not clear enough for the user just to say `map`, I'll keep it.

nilearn/plotting/surface/_backend.py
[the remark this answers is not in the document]
hndgzkn: Finally I set `DEFAULT_HEMI='both'.`

nilearn/plotting/surface/_utils.py
```diff
@@ -86,42 +94,111 @@ def _check_bg_map(bg_map, hemi):
...
+            message=(
+                f"{hemi=} was passed with "
+                f"{type(surf_map)=} and {type(surf_mesh)=}.\n"
+                "This value will be ignored as it is only used when "
+                "'map' is a SurfaceImage instance and / or "
+                "'mesh' is a PolyMesh instance."
+            ),
+            stacklevel=3,
+        )
+
```
hndgzkn, line 201: I moved the warning for `sanitize_hemi_...` here in `check_surface_plotting_inputs`.

nilearn/plotting/surface/_utils.py
```diff
@@ -135,20 +212,17 @@ def check_surface_plotting_inputs(
             f"then {mesh_var_name} must be a SurfaceImage instance."
         )
 
-    if isinstance(surf_mesh, PolyMesh):
-        surf_mesh = _get_hemi(surf_mesh, hemi)
```
hndgzkn, line 139: Moved this down, just to make type checks first.

nilearn/plotting/surface/_utils.py
```diff
@@ -135,20 +212,17 @@ def check_surface_plotting_inputs(
...
     if isinstance(surf_mesh, SurfaceImage):
         raise TypeError(
             "'surf_mesh' cannot be a SurfaceImage instance. ",
             "Accepted types are: str, list of two numpy.ndarray, "
             "InMemoryMesh, PolyMesh, or None.",
         )
 
-    if isinstance(surf_map, SurfaceImage):
-        if surf_mesh is None:
-            surf_mesh = _get_hemi(surf_map.mesh, hemi)
```
hndgzkn, line 150: Moved this down, after shape check.

nilearn/plotting/surface/tests/test_matplotlib_backend.py
```diff
@@ -233,351 +227,3 @@ def test_compute_facecolors_deprecation():
             0.5,
             alpha,
         )
-
-
-def test_plot_surf_with_title(matplotlib_pyplot, in_memory_mesh, bg_map):
-    """Test if figure title is set correctly in
-    nilearn.plotting.surface.surf_plotting.plot_surf.
-    """
-    display = plot_surf(
```
hndgzkn, line 242: I've moved the tests below to `test_surf_plotting` thinking at the end that it is more convenient.

nilearn/plotting/surface/tests/test_plotly_backend.py
```diff
@@ -158,27 +149,77 @@
...
+    assert config["y"] == 0.96
+    assert config["xanchor"] == "center"
+    assert config["yanchor"] == "top"
+    assert config["font"]["size"] == 22
+    assert config["font"]["color"] == "green"
+
+
 @pytest.mark.parametrize("full_view", EXPECTED_CAMERAS_PLOTLY)
-def test_get_view_plot_surf(full_view):
-    """Test if nilearn.plotting.surface._plotly_backend._get_view_plot_surf
```
hndgzkn, line 163: I've split this test into 3 separate tests and added a case to increase coverage.

nilearn/plotting/surface/_utils.py
```diff
@@ -2,22 +2,29 @@
...
 from nilearn._utils.helpers import is_matplotlib_installed, is_plotly_installed
+from nilearn.plotting._utils import DEFAULT_ENGINE
 from nilearn.surface import (
     PolyMesh,
     SurfaceImage,
 )
 from nilearn.surface.surface import combine_hemispheres_meshes, get_data
 from nilearn.surface.utils import check_polymesh_equal
 
+DEFAULT_HEMI = "left"
```
hndgzkn, line 17: I've set the value to `'left'` for the moment to see if tests pass. 

When it's set to `'both'` as the previously saved images are not the same, the tests that create and compare images fail. 

I think we can keep this as `'left'` until refactoring `nilearn.plotting.surface` package is finished to make sure that we have not diverged from  previous images. When all is complete, we  can change value to `'both'`. How does it soound?

nilearn/_utils/docs.py
```diff
@@ -73,7 +73,22 @@ def custom_function(vertices):
     If `None`, the complete figure is used.
 """
 
-# bg_img
+# surf_mesh
+docdict["surf_mesh"] = """
+surf_mesh : :obj:`str` or :obj:`list` of two :class:`numpy.ndarray` \
```
Remi-Gau, line 78: FYI I think we try to keep the entries in this file in alphabetical order within each section, so this should probably be moved down

Remi-Gau (approved):
the diff is defo a bit big to review but I could not spot anything strange

there is a couple of codecov complaints but don't know if they are legit

hndgzkn:
> the diff is defo a bit big to review but I could not spot anything strange

You are right. If you prefer, I can divide that.
 
> there is a couple of codecov complaints but don't know if they are legit

Generally when I move code from one module to another, it complains. But for this one, may be I've added a few lines, minor code, that are not covered by the tests.

Remi-Gau:
> Let me know if you prefer that I divide this PR.

nope

went through it commit by commit: it helps that made them quite atomic, defo something I should do more often

nilearn/plotting/surface/_matplotlib_backend.py
```diff
@@ -494,3 +508,102 @@ def _plot_surf(
...
+                    )
+                axes.collections[0]._edgecolor3d[faces_outside] = color
+            if label and legend:
+                patch_list.append(Patch(color=color, label=label))
+        # plot legend only if indicated and labels provided
+        if legend and np.any([lbl is not None for lbl in labels]):
+            figure.legend(handles=patch_list)
+            # if legends, then move title to the left
+        if title is None and hasattr(figure._suptitle, "_text"):
+            title = figure._suptitle._text
```
Remi-Gau, line 605: that's a legit coverage complaint: see the updloaded HTML coverage report

https://github.com/nilearn/nilearn/actions/runs/14991516206/artifacts/3112733623

nilearn/plotting/surface/_utils.py
```diff
@@ -86,42 +94,111 @@ def _check_bg_map(bg_map, hemi):
...
+    -------
+    surf_mesh : :obj:`numpy.ndarray`,  :obj:`~nilearn.surface.InMemoryMesh`
+        Surface mesh corresponding to the specified ``hemi``.
+
+        - If ``hemi='left'`` or ``hemi='right'``, returns
+          :obj:`numpy.ndarray`.
+        - If ``hemi='both'``, returns :obj:`~nilearn.surface.InMemoryMesh`
     """
+    if not isinstance(surf_mesh, PolyMesh):
+        raise ValueError("mesh should be of type PolyMesh.")
```
Remi-Gau, line 120: that's a legit coverage complaint: see the updloaded HTML coverage report

https://github.com/nilearn/nilearn/actions/runs/14991516206/artifacts/3112733623

nilearn/plotting/surface/_utils.py
```diff
@@ -86,42 +94,111 @@ def _check_bg_map(bg_map, hemi):
...
     if hemi == "both":
-        return combine_hemispheres_meshes(mesh)
-    elif hemi in mesh.parts:
-        return mesh.parts[hemi]
+        return combine_hemispheres_meshes(surf_mesh)
+    elif hemi in ["left", "right"]:
+        if hemi in surf_mesh.parts:
+            return surf_mesh.parts[hemi]
+        else:
+            raise ValueError(
```
Remi-Gau, line 128: that's a legit coverage complaint: see the updloaded HTML coverage report

https://github.com/nilearn/nilearn/actions/runs/14991516206/artifacts/3112733623

Remi-Gau:
let's not worry about the codecov complaints for now given the size of the PR

I would suggest finishing the refactoring of the plotting module and then see where we got blind spot in coverage and fix those then: I would not be surprised if those lines are not hit on main either

nilearn/_utils/docs.py
```diff
@@ -438,7 +438,7 @@ def custom_function(vertices):
 
 # hemi
 docdict["hemi"] = """
-hemi : {"left", "right", "both"}, default="left"
+hemi : {"left", "right", "both"}, default="both"
```
man-shu, line 441: Sorry if I am missing something but I read somewhere that we are keeping the default "left" in this PR?

man-shu (approved):
Just that one minor comment, otherwise LGTM!

nilearn/_utils/docs.py
hndgzkn: Yes, I've changed in the code, but not doc thinking that finally it will be `both`. Do you think I should keep them consistent? If yes, I can change this to `left` as well.

hndgzkn: I've changed that finally.

bthirion:
Thx !

PR #5352, opened 2025-04-24
[ENH] Revisit example plot_mask_large_fmri

man-shu:
- Closes none but revisits example [plot_mask_large_fmri](https://nilearn.github.io/dev/auto_examples/07_advanced/plot_mask_large_fmri.html) after the findings in https://github.com/nilearn/nilearn/pull/5128#discussion_r2021726878

- I think the issues seen due to numpy mmap'ed arrays in [comment](https://github.com/nilearn/nilearn/pull/5128#discussion_r2021726878) do not show up here because we are doing (almost) everything inside functions
- the only exception here is the initial loading of images, which I have now moved under a function
- Additionally, I confirm that by profiling each case separately in a script via the command line `mprof`. This gives us timeseries of memory usage. These timeseries show two things:
  - each "child" process shows up separately, meaning memory_profiler is looking at each parallel process separately
  - the peak memory usage largely matches what we saw previously in the final plot in this example (even before moving the loading under a function)
   (but let's see if the results change after the changes here)

### DIFF
--- a/examples/07_advanced/plot_mask_large_fmri.py
+++ b/examples/07_advanced/plot_mask_large_fmri.py
@@ -60,17 +60,23 @@ from nilearn.image import concat_imgs
 N_SUBJECTS = 6
 N_REGIONS = 4
 
-fmri_data = fetch_adhd(n_subjects=N_SUBJECTS)
-fmri_img = concat_imgs(fmri_data.func)
-n_timepoints = fmri_img.shape[-1]
 
-output_dir = Path.cwd() / "results" / "plot_mask_large_fmri"
-output_dir.mkdir(parents=True, exist_ok=True)
-print(f"Large fmri file will be saved to:\n{output_dir}")
+def create_large_fmri(n_subjects):
+    fmri_data = fetch_adhd(n_subjects=n_subjects)
+    fmri_img = concat_imgs(fmri_data.func)
+    n_timepoints = fmri_img.shape[-1]
 
-fmri_path = output_dir / "large_fmri.nii.gz"
-fmri_img.to_filename(fmri_path)
+    output_dir = Path.cwd() / "results" / "plot_mask_large_fmri"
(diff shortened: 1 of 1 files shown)


man-shu:
## NiftiMasker, with path
mprofile_drago_nifti_masker_parellel_path

## NiftiMasker, with in-memory image
mprofile_drago_nifti_masker_parellel_inmem

## Numpy indexing
mprofile_drago_numpy_parallel

## Numpy indexing, with SharedMemory
mprofile_drago_shared_mem


For example, doing
```bash
mprof run -MC test_nifti_masker_parallel_path.py
```
where 

`test_nifti_masker_parallel_path.py` is as below:
```python
# file test_nifti_masker_parallel_path.py
from joblib import Parallel, delayed

N_REGIONS = 4


def nifti_masker_single(fmri, mask):
    return NiftiMasker(mask_img=mask).fit_transform(fmri)


def nifti_masker_parallel_path(fmri_path, mask_paths):
    Parallel(n_jobs=N_REGIONS)(
        delayed(nifti_masker_single)(fmri_path, mask_path)
        for mask_path in mask_paths
    )
    return


if __name__ == "__main__":
    fmri_path = "large_fmri.nii.gz"

    mask_paths = [
        "mask_1.nii.gz",
        "mask_2.nii.gz",
        "mask_3.nii.gz",
        "mask_4.nii.gz",
    ]
    nifti_masker_parallel_path(fmri_path, mask_paths)
```

And then you can do the this to plot the results:

```bash
mprof plot
```

man-shu:
Indeed, the peaks match with the [updated example
](https://output.circle-artifacts.com/output/job/16cf670b-cb4b-4cca-be97-2bec50593de4/artifacts/0/dev/auto_examples/07_advanced/plot_mask_large_fmri.html) (better than before the mods here)
sphx_glr_plot_mask_large_fmri_001 (1)

bthirion (approved):
LGTM, thx.

PR #5353, opened 2025-04-24
[REF] standardize how `n_elements` is handled across maskers

Remi-Gau:
- Closes none

- ensure consistent use of the n_elements_ attribute across maskers
- ensure all maskers send same error message if signal to inverse transfrom has the wrong shape
- better align behavior of SurfaceLabelMasker to that of NiftiLabelMasker: implement similar properties with same behavior, rely on fitted look up table to access name and indices of regions...


nilearn/maskers/surface_labels_masker.py
Remi-Gau, line 1: Summary of changes:
- apply mask to label_img at fit time: keep result in attribute labels_img_ 
- implement properties similar to those found in the nifti label masker

nilearn/maskers/surface_labels_masker.py
```diff
@@ -30,59 +32,18 @@
...
-            "(including background).",
-            stacklevel=find_stack_level(),
-        )
-    labels = np.unique(labels_data)
-    labels = labels[labels != background_label]
-
-    return labels_data, labels
-
-
 def signals_to_surf_img_labels(
```
bthirion, line 39: I this is a private or public function ? If public, it should have full docstrings.

bthirion (commented):
LGTM overall.

nilearn/maskers/surface_labels_masker.py
Remi-Gau: na it's private
used in the parcellation module, hence why it has no leading underscore

nilearn/maskers/base_masker.py
```diff
@@ -422,13 +422,23 @@ def inverse_transform(self, X):
...
         return img
 
+    def _check_signal_shape(self, signals: np.ndarray):
+        if signals.shape[-1] != self.n_elements_:
+            raise ValueError(
+                "Input to 'inverse_transform' has wrong shape.\n"
+                f"Last dimension should be {self.n_elements_}.\n"
+                f"Got {signals.shape[-1]}."
+            )
+
```
hndgzkn, line 441: If I'm not mistaken, this function is exactly the same as the one starting at line 627. Isn't it possible to define it once in a base class, or else where and call that function?

nilearn/maskers/surface_labels_masker.py
```diff
@@ -234,10 +195,55 @@ def __init__(
...
+    def n_elements_(self) -> int:
+        """Return number of regions.
+
+        This is equal to the number of unique values
+        in the fitted label image,
+        minus the background value.
+        """
+        check_is_fitted(self)
+        lut = self.lut_
+        return len(lut[lut["index"] != self.background_label])
```
hndgzkn, line 207: This is slightly different then the previous definition. Should they be the same, if yes is it possible to define this in a base class just once?

hndgzkn (approved):
Minor comments, otherwise OK.

nilearn/maskers/base_masker.py
Remi-Gau: yeah that's because we got a base class for surface masker and one for volume masker

I have been thinking of extracting commonalities in a base class but so far the amount of duplication is bearable in my opinion but definitely something to keep a eye on.

nilearn/maskers/surface_labels_masker.py
Remi-Gau: yeah I have had to change things to make sure that behavior matched that of nifti label masker

Remi-Gau: extracting in a parent class: I wanted to but n_elements is always a bit different from masker to masker (number of ROIs, number of voxels, number of maps...)

nilearn/maskers/base_masker.py
hndgzkn: May be putting the function in a common utility class and call that function from `_check_signal_shape` to avoid writing the same code twice?

Remi-Gau: yeah I will do that in the follow up PR: I need to refactor how some of the lut code is handled across label maskers

PR #5356, opened 2025-04-29
FIX] cast surface image to float before create a mask from them in surface masker

Remi-Gau:
- Closes none

- cast surface image to float before create a mask from them in surface masker
- fixes doc build


man-shu (approved):
LGTM!

PR #5357, opened 2025-04-29
[ENH] generate report for second level model when inputs are first level models or dataframe

### LINKED ISSUE #2406: GLM Reports: Cannot specify a list of first level contrasts when passing in a list of first level models.
I've also encountered an issue with the reports. If the inputs to your second-level analysis are a list of  `FirstLevelModel` instances, there is no way to define first-level contrasts of these models. In this case, defining first-level contrasts are necessary, otherwise `SecondLevelModel.compute_contrasts` will throw an error: 

```
ValueError: If second_level_input was a list of FirstLevelModel, then first_level_contrast is mandatory. It corresponds to the second_level_contrast argument of the compute_contrast method of FirstLevelModel
```

It seems to me that if you pass in a second-level model, it only accepts second-level contrasts. Not sure whether or not to make a separate issue with the impending nilearn merge.

_Originally posted by @danjgale in https://github.com/nistats/nistats/issues/409#issuecomment-569840429_

Remi-Gau:
- Closes #2406

- [x] update changelog
- [x] allow to generate report for second level model when inputs are first level models


bthirion:
Thx for taking care of this !

nilearn/glm/_base.py
```diff
@@ -219,6 +220,22 @@ def generate_report(
...
+                            or None, default=None
+            For :class:`~nilearn.glm.second_level.SecondLevelModel`,
+            in case a :obj:`list` of
+            :class:`~nilearn.glm.first_level.FirstLevelModel` was provided
+            as ``second_level_input``,
+            we have to provide a :term:`contrast`
+            to apply to the first level models
+            to get the corresponding list of images desired,
+            that would be tested at the second level.
+            This parameter is ignore for all other cases.
```
hndgzkn, line 235:
```suggestion
            This parameter is ignored for all other cases.
```

nilearn/interfaces/bids/glm.py
```diff
@@ -102,6 +108,22 @@ def save_glm_to_bids(
...
+                        or None, default=None
+        For :class:`~nilearn.glm.second_level.SecondLevelModel`,
+        in case a :obj:`list` of
+        :class:`~nilearn.glm.first_level.FirstLevelModel` was provided
+        as ``second_level_input``,
+        we have to provide a :term:`contrast`
+        to apply to the first level models
+        to get the corresponding list of images desired,
+        that would be tested at the second level.
+        This parameter is ignore for all other cases.
```
hndgzkn, line 123:
```suggestion
        This parameter is ignored for all other cases.
```

nilearn/reporting/glm_reporter.py
```diff
@@ -131,6 +132,22 @@ def make_glm_report(
...
+                        or None, default=None
+        For :class:`~nilearn.glm.second_level.SecondLevelModel`,
+        in case a :obj:`list` of
+        :class:`~nilearn.glm.first_level.FirstLevelModel` was provided
+        as ``second_level_input``,
+        we have to provide a :term:`contrast`
+        to apply to the first level models
+        to get the corresponding list of images desired,
+        that would be tested at the second level.
+        This parameter is ignore for all other cases.
```
hndgzkn, line 147:
```suggestion
        This parameter is ignored for all other cases.
```

nilearn/_utils/glm.py
```diff
@@ -154,6 +156,22 @@ def make_stat_maps(model, contrasts, output_type="z_score"):
...
+                        or None, default=None
+        For :class:`~nilearn.glm.second_level.SecondLevelModel`,
+        in case a :obj:`list` of
+        :class:`~nilearn.glm.first_level.FirstLevelModel` was provided
+        as ``second_level_input``,
+        we have to provide a :term:`contrast`
+        to apply to the first level models
+        to get the corresponding list of images desired,
+        that would be tested at the second level.
+        This parameter is ignore for all other cases.
```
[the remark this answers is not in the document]
hndgzkn: Could this docstring be moved to `nilearn._utils.docs`?

nilearn/glm/second_level/second_level.py
```diff
@@ -637,8 +637,8 @@ def compute_contrast(
 
         first_level_contrast : :obj:`str` or :class:`numpy.ndarray` of \
                             shape (n_col) with respect to \
-                            :class:`~nilearn.glm.first_level.FirstLevelModel`,
-                            default=None
+                            :class:`~nilearn.glm.first_level.FirstLevelModel` \
+                            or None, default=None
```
hndgzkn, line 641: Is the docstring here more elaborate than the previous ones? It considers the case `pandas.DataFrame` (a part from `FirstLevelModel`) being provided as `second_level_input`.

Remi-Gau: hmmm... I need to add a test to make all cases for second level are covered: then if that's the case we should probably reuse a larger chunk of the docstring and try to indeed rely on _utils.doc

nilearn/_utils/glm.py
```diff
@@ -134,7 +134,9 @@ def coerce_to_dict(input_arg):
     return input_arg
 
 
-def make_stat_maps(model, contrasts, output_type="z_score"):
+def make_stat_maps(
```
bthirion, line 137: The naming should make it explicit that this is computing second-level contrasts on the inputs: first-level models and corresponding contrasts.

Remi-Gau: I'd like that but this function is supposed to work with both first and second level models, so it may be better to keep generic.

nilearn/_utils/glm.py
```diff
@@ -156,19 +156,17 @@ def make_stat_maps(
...
         For :class:`~nilearn.glm.second_level.SecondLevelModel`,
         in case a :obj:`list` of
-        :class:`~nilearn.glm.first_level.FirstLevelModel` was provided
-        as ``second_level_input``,
+        :class:`~nilearn.glm.first_level.FirstLevelModel`
+        or a dataframe was provided as ``second_level_input``,
         we have to provide a :term:`contrast`
         to apply to the first level models
         to get the corresponding list of images desired,
         that would be tested at the second level.
```
hndgzkn, line 168: This part already exists in `nilearn/_utils/docs`, non? Or I am missing a detail.

nilearn/glm/_base.py
```diff
@@ -220,19 +220,17 @@ def generate_report(
...
             For :class:`~nilearn.glm.second_level.SecondLevelModel`,
             in case a :obj:`list` of
-            :class:`~nilearn.glm.first_level.FirstLevelModel` was provided
-            as ``second_level_input``,
+        :class:`~nilearn.glm.first_level.FirstLevelModel`
+        or a dataframe was provided as ``second_level_input``,
             we have to provide a :term:`contrast`
             to apply to the first level models
             to get the corresponding list of images desired,
             that would be tested at the second level.
```
hndgzkn, line 232: The same here.

Remi-Gau: not quite 

first_level_contrast in _utils.docs is the thing that is relevant only for SecondLevelModel that you will find in nilear/glm/second_level/second_level.py.

but for reports, the same API is used for both first and second level, so I feel that an extra explanation is needed to say why this parameter is necessary for second level model but not first level

nilearn/glm/second_level/second_level.py
Remi-Gau: added a test to make sure generate report also works when using a dataframe as input

nilearn/glm/_base.py
Remi-Gau: actually when rereading the rendered doc it was a bit weird to see some parameters in FirstLevelModel.generate_report() that do not explain they have to be ignored so I have put everything in _utils.doc

nilearn/glm/_base.py
```diff
@@ -157,164 +156,6 @@ def r_square(self):
             "r_square", result_as_time_series=False
         )
 
-    @fill_doc
-    def generate_report(
```
Remi-Gau, line 161: implemented differently in first and second level

bthirion (approved):
LGTM, thx.

PR #5362, opened 2025-04-30
[FIX] use property region_names_ in surface label examples

Remi-Gau:
- Closes none

- update one of the examples to use one of the new public property instead of one the old private attributes (that has been removed)


man-shu (approved):
LGTM!

PR #5367, opened 2025-05-05
[FIX] Failing benchmark workflow on schedule

man-shu:
- Closes None but fixes the failing benchmark workflow on schedule: https://github.com/nilearn/nilearn/actions/runs/14825724575/job/41618562617#step:9:1851

- The error code suggests it's a timeout error in setup_cache step
- We download a dataset in that step
- So my guess is that downloading the dataset jus tfails for some reason on Sunday 10pm UTC
- changing to every Monday at 9am UTC (should run at 11am Paris time)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -40,8 +40,8 @@ on:
         branches:
         -   main
     schedule:
-    # everday at 10pm UTC
-    -   cron: 0 22 * * 0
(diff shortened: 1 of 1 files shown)


man-shu:
Merging quickly to test it out at 11am!

man-shu:
Seems to be working! Let's see what happens next monday.

PR #5372, opened 2025-05-12
[MAINT] only test label maskers on real atlases on main

Remi-Gau:
- Closes none

- only test label maskers on real atlases on main to minimize CI runs (and also failures when atlases cannot be fetched)


Remi-Gau:
test failure on pre-release is unrelated: my hunch is that this is due to the last rc1 of sklearn 1.7.0

https://github.com/scikit-learn/scikit-learn/releases/tag/1.7.0rc1

Issue #3428, opened 2022-11-28
CanICA (and more generally decomposition._base) masker is applied on .fit but not on .transform


RaphaelMeudec:
I'm unsure if this is a bug, but:

- when you train CanICA initialized with specific masker parameters (e.g `low_pass=0.15`), the masker is applied on the inputs of the `.fit`
- whereas when you apply the `.transform`, the masker isn't applied to the inputs

This introduces a slight difference between the .fit where you don't have to preprocess the input and the .transform where you have to do the preprocessing on your own before transforming it. It confused me when applying CanICA to resting-state fmri.

Nilearn version:

### Expected behavior

Expected preprocessing of the inputs should be the same for .fit and .transform

### Actual behavior

CanICA preprocesses the inputs in .fit, but expects preprocessed inputs for .transform

bthirion:
Thx for reporting.
Note that the two maskers do conceptually different things. The first one prepares the data for ICA, the second one returns times series corresponding to ICs for a given fMRI dataset.
Apparently, the only way to filter these transformed times series is to call nilearn.clean manually.  I understand that in general, we may not want the two maskers to have the same parameters, and that it is cumbersome to specyf two sets of parameters. If we consider that this is a feature not a bug, we should document it better.

Remi-Gau:
Setting attributes on the masker in charge of transform after the fit should work for this, no?

```python
canica_estimator = CanICA(mask=MultiNiftiMasker(low_pass=0.15))
canica_estimator.fit(imgs)
canica_estimator.maps_masker_.low_pass=0.15
canica_estimator.transform(imgs)
```

If we do not want to always pass attributes across maskers we should at least, better document what the public attributes created at fit time are.

Remi-Gau:
> better document what the public attributes created at fit time are.

to be fair this statement is probably valid for most of our estimators

bthirion:
> Setting attributes on the masker in charge of transform after the fit should work for this, no?

probably... but I would check ;-)

PR #5377, opened 2025-05-14
[ENH] Make all decomposition methods work with surface data

### LINKED ISSUE #5093: [ENH] Ensure `DictLearning` works with surface data
I think this one is big enough, I will open another PR for DictLearning.

_Originally posted by @man-shu in https://github.com/nilearn/nilearn/issues/4756#issuecomment-2551484882_

I realized #4756 was focused on Parcellations and only extended DictLearning surface support partially. So opening this to track that.

man-shu:
- Closes #5093

- check file type (nifti or surface) from file extension in:
  - fit method of  _BaseDecomposition class
  - concat_imgs
  - seems like a systematic issue, so probably a new separate issue needed?

- check dict_learning, canica.py, _multi_pca.py
- check related examples
- add related tests

### DIFF
--- a/nilearn/decomposition/_base.py
+++ b/nilearn/decomposition/_base.py
@@ -31,6 +31,8 @@ from nilearn.maskers import NiftiMapsMasker, SurfaceMapsMasker, SurfaceMasker
 from nilearn.signal import row_sum_of_squares
 from nilearn.surface import SurfaceImage
 
+SURFACE_DATA_EXTENSIONS = ("gii", "gii.gz", "mgz")
+
 
 def _fast_svd(X, n_components, random_state=None):
     """Automatically switch between randomized and lapack SVD (heuristic \
@@ -478,11 +480,18 @@ class _BaseDecomposition(CacheMixin, TransformerMixin, BaseEstimator):
             )
 
         masker_type = "nii"
-        if isinstance(self.mask, (SurfaceMasker, SurfaceImage)) or any(
-            isinstance(x, SurfaceImage) for x in imgs
+        if (
+            isinstance(self.mask, (SurfaceMasker, SurfaceImage))
+            or any(isinstance(x, SurfaceImage) for x in imgs)
+            or any(
+                isinstance(x, str) and x.endswith(SURFACE_DATA_EXTENSIONS)
+                for x in imgs
+            )
         ):
             masker_type = "surface"
         self.masker_ = check_embedded_masker(self, masker_type=masker_type)
+
+        breakpoint()
 
         # Avoid warning with imgs != None
         # if masker_ has been provided a mask_img


nilearn/decomposition/_base.py
```diff
@@ -478,12 +480,19 @@ def fit(self, imgs, y=None, confounds=None):
...
         masker_type = "nii"
-        if isinstance(self.mask, (SurfaceMasker, SurfaceImage)) or any(
-            isinstance(x, SurfaceImage) for x in imgs
+        if (
+            isinstance(self.mask, (SurfaceMasker, SurfaceImage))
+            or any(isinstance(x, SurfaceImage) for x in imgs)
+            or any(
+                isinstance(x, str) and x.endswith(SURFACE_DATA_EXTENSIONS)
+                for x in imgs
+            )
```
man-shu, line 489: @Remi-Gau do we have a better way to check the file type via file extensions or paths.

I hit the same issue in concat_imgs.

man-shu: Ah ok it's because we never directly provide the paths to .gii images. 
I forgot that loading surface image needs to be done by putting together a mesh and data files (accessed via separate gii files):

https://nilearn.github.io/stable/auto_examples/00_tutorials/plot_surface_101.html#load-the-surface-image

Remi-Gau: yup for the moment we never instanciate SurfaceImage directly from disk

nilearn/decomposition/_base.py
```diff
@@ -274,10 +276,12 @@ class _BaseDecomposition(CacheMixin, TransformerMixin, BaseEstimator):
            :obj:`~nilearn.surface.SurfaceImage` or
            :obj:`~nilearn.maskers.SurfaceMasker` object, optional
         Mask to be used on data. If an instance of masker is passed,
-        then its mask will be used. If no mask is given, it will be computed
-        automatically by a MultiNiftiMasker for Niimg-like objects with default
-        parameters and no mask will be used for SurfaceImage objects.
+        then its mask will be used. If no mask is given, for Nifti images,
+        it will be computed automatically by a MultiNiftiMasker with default
+        parameters; for surface images, all the vertices will be used.
```
man-shu, line 281: This isn't exactly true as of yet.
- See: https://github.com/nilearn/nilearn/issues/5390. Will fix in a separate PR

nilearn/decomposition/dict_learning.py
```diff
@@ -45,11 +45,13 @@ class DictLearning(_BaseDecomposition):
 
     Parameters
     ----------
-    mask : Niimg-like object or MultiNiftiMasker instance, optional
+    mask : Niimg-like object or MultiNiftiMasker instance or
+           :obj:`~nilearn.surface.SurfaceImage` or
+           :obj:`~nilearn.maskers.SurfaceMasker` object, optional
```
Remi-Gau, line 50:
```suggestion
    mask : Niimg-like object or MultiNiftiMasker instance or \
           :obj:`~nilearn.surface.SurfaceImage` or \
           :obj:`~nilearn.maskers.SurfaceMasker` object, optional
```

nilearn/decomposition/dict_learning.py
```diff
@@ -63,7 +65,8 @@ class DictLearning(_BaseDecomposition):
     alpha : :obj:`float`, default=10
         Sparsity controlling parameter.
 
-    dict_init : Niimg-like object, optional
+    dict_init : Niimg-like object or
```
Remi-Gau, line 68:
```suggestion
    dict_init : Niimg-like object or \
```

nilearn/decomposition/dict_learning.py
```diff
@@ -146,29 +143,37 @@ class DictLearning(_BaseDecomposition):
...
-    components_img_ : 4D Nifti image
-        4D image giving the extracted components. Each 3D image is a component.
+    components_img_ : 4D Nifti image or 2D :obj:`~nilearn.surface.SurfaceImage`
+        The image giving the extracted components. Each 3D Nifti image or 1D
+        SurfaceImage is a component.
 
         .. versionadded:: 0.4.1
 
-    masker_ : instance of MultiNiftiMasker
+    masker_ : instance of MultiNiftiMasker or
```
Remi-Gau, line 160:
```suggestion
    masker_ : instance of MultiNiftiMasker or \
```

bthirion:
LMK when you want a review.

nilearn/decomposition/_base.py
```diff
@@ -32,6 +33,32 @@
 from nilearn.surface import SurfaceImage
 
 
+def _warn_ignored_surface_masker_params(estimator):
+    """Warn about parameters that are ignored by SurfaceMasker.
```
Remi-Gau, line 38: note that something similar should probably be generalized for other functions/methods that accept both volume and surface data, but use some parameters for volume only

nilearn/decomposition/_multi_pca.py
```diff
@@ -111,28 +109,33 @@ class _MultiPCA(_BaseDecomposition):
 
     Attributes
     ----------
-    masker_ : instance of MultiNiftiMasker
+    masker_ : instance of MultiNiftiMasker or \
```
Remi-Gau: doc strings for attributes seem to be VERY similar between classes: maybe let's try to refactor them using _utils/docs.py

nilearn/decomposition/_base.py
man-shu: Yeah I think we were looking at this and this function seemed like the most appropriate place for this: 

https://github.com/nilearn/nilearn/blob/ea4cd19f777ba8bd84018904154b092ff5cd3252/nilearn/_utils/masker_validation.py#L15

But we decided to keep that for another PR

nilearn/decomposition/_multi_pca.py
man-shu: Indeed I was just unsure how best to parameterize this because this `masker_` docstring would be different from that in other places like in maskers or decoders.

nilearn/decomposition/tests/test_base.py
```diff
@@ -60,38 +58,16 @@ def test_check_estimator_nilearn(estimator, check, name):  # noqa: ARG001
...
-        this_img[2:4, 2:4, 2:4, :] += 10
-
-        imgs.append(Nifti1Image(this_img, _affine_eye()))
-
-    return imgs
-
-
-@pytest.fixture
-def masker():
-    return MultiNiftiMasker(mask_img=_img_3d_ones()).fit()
```
Remi-Gau, line 82: moved and refactored all fixtures to nilearn/decomposition/tests/conftest.py

nilearn/decomposition/tests/test_canica.py
```diff
@@ -13,109 +13,10 @@
...
 from nilearn.decomposition.canica import CanICA
 from nilearn.image import get_data, iter_img
 from nilearn.maskers import MultiNiftiMasker
 
-SHAPE = (30, 30, 5)
-
-N_SUBJECTS = 2
-
-
-def _make_data_from_components(
```
Remi-Gau, line 26: moved and refactored to nilearn/decomposition/tests/conftest.py

nilearn/decomposition/tests/test_dict_learning.py
```diff
@@ -10,33 +10,10 @@
...
-from nilearn.conftest import _affine_eye
 from nilearn.decomposition.dict_learning import DictLearning
-from nilearn.decomposition.tests.test_canica import _make_canica_test_data
 from nilearn.image import get_data, iter_img
 from nilearn.maskers import NiftiMasker
 
-SHAPE = (30, 30, 5)
-
-
-@pytest.fixture(scope="module")
```
Remi-Gau, line 22: moved and refactored all fixtures to nilearn/decomposition/tests/conftest.py

nilearn/decomposition/tests/test_decomposition_estimators.py
Remi-Gau, line 1: A lot of tests were identical or nearly identical for _MultiPCA, DictLearning and CanICA.

I moved them all here and parametrize over the different estimators.

nilearn/decomposition/_base.py
```diff
@@ -477,15 +514,17 @@ def fit(self, imgs, y=None, confounds=None):
...
             # it early and raise a helpful message
             raise ValueError(
-                "Need one or more Niimg-like objects as input, "
+                "Need one or more Niimg-like or SurfaceImage "
+                "objects as input, "
                 "an empty list was given."
             )
 
-        masker_type = "nii"
+        masker_type = "multi_nii"
```
Remi-Gau, line 523: we are supposed to work with multiniftimaskers

Remi-Gau:
@hndgzkn @bthirion @man-shu 
this should be ready for review with all tests passing

sorry for the huge PR but the gist of is:
- refactor doc strings
- refactor tests to run the same tests across decomposition estimators

nilearn/_utils/docs.py
```diff
@@ -1260,6 +1265,53 @@ def custom_function(vertices):
         Type of atlas.
         See :term:`Probabilistic atlas` and :term:`Deterministic atlas`."""
 
+docdict["base_decomposition_attributes"] = """
```
man-shu, line 1268: Ok nice. Noted!

nilearn/decomposition/tests/test_canica.py
```diff
@@ -2,220 +2,66 @@
...
 )
-def test_check_estimator_nilearn(estimator, check, name):  # noqa: ARG001
-    """Check compliance with nilearn estimators rules."""
-    check(estimator)
+from nilearn.image import get_data, iter_img
+from nilearn.surface.surface import get_data as get_surface_data
 
 
-def test_threshold_bound_error(canica_data):
+@pytest.mark.parametrize("data_type", ["nifti"])
```
man-shu: Just to make sure I understand, where does the `data_type` go in as input here?
And should we not test this for surface as well?

nilearn/decomposition/tests/test_canica.py
```diff
@@ -2,220 +2,66 @@
...
-    # error is raised when no data is passed.
-    with pytest.raises(
-        TypeError,
-        match="missing 1 required positional argument: 'imgs'",
-    ):
-        canica.fit()
-
-
-def test_percentile_range(rng, canica_data):
+@pytest.mark.parametrize("data_type", ["nifti"])
```
man-shu: same thing here

nilearn/decomposition/tests/test_canica.py
```diff
@@ -2,220 +2,66 @@
...
 
     with pytest.warns(UserWarning, match="obtained a critical threshold"):
-        canica.fit(canica_data)
+        canica.fit(canica_data_single_img)
 
 
-def test_canica_square_img(mask_img):
-    data, components, rng = _make_canica_test_data(n_subjects=8)
-
+@pytest.mark.parametrize("data_type", ["nifti"])
```
man-shu, line 49: and here

nilearn/decomposition/tests/test_decomposition_estimators.py
```diff
@@ -0,0 +1,298 @@
...
+            est.fit(decomposition_data_single_img)
+    elif data_type == "nifti":
+        est.fit(decomposition_data_single_img)
+
+    check_decomposition_estimator(est, data_type)
+
+
+@pytest.mark.timeout(0)
+@pytest.mark.parametrize("estimator", [CanICA, _MultiPCA, DictLearning])
+@pytest.mark.parametrize("data_type", ["nifti"])
```
man-shu, line 175: Here only checking nifti makes sense because we are testing the deprecation of nifti_maps_masker_ attribute. 

But I don't understand where does `data_type` variable goes into the test?

nilearn/decomposition/tests/test_dict_learning.py
```diff
@@ -1,137 +1,78 @@
...
-    "estimator, check, name",
-    nilearn_check_estimator(estimators=ESTIMATORS_TO_CHECK),
-)
-def test_check_estimator_nilearn(estimator, check, name):  # noqa: ARG001
-    """Check compliance with nilearn estimators rules."""
-    check(estimator)
+from nilearn.surface.surface import get_data as get_surface_data
 
 
+@pytest.mark.parametrize("data_type", ["nifti"])
```
man-shu: we could technically run this for surface too no? Or is it too expensive as a smoke test to run twice?

nilearn/decomposition/tests/test_dict_learning.py
```diff
@@ -1,137 +1,78 @@
...
+    dict_learning.fit(canica_data)
 
+    check_decomposition_estimator(dict_learning, data_type)
 
-def test_dict_learning(mask_img):
-    data, components, _ = _make_canica_test_data()
 
-    masker = NiftiMasker(mask_img=mask_img).fit()
-    mask = get_data(mask_img) != 0
+@pytest.mark.parametrize("data_type", ["nifti"])
```
man-shu, line 44: same here, could we run this for surface?

nilearn/decomposition/tests/test_canica.py
man-shu: can `canica_data_single_img` fixture directly read from this parameterization?

Remi-Gau:
> Thanks a lot for taking over this

no worries
it allowed me to get a bit familiar with this part of the code base

nilearn/decomposition/tests/test_canica.py
Remi-Gau: yup: that's pytest magic for you

the thing I don't like for it is that, it is not clear from reading the code

do you think we should make it more explicit:


```python
import pytest

@pytest.fixture
def foo(implicit_value):
    if implicit_value == "nifti":
        return 1
    else:
        return 2

@pytest.mark.parametrize("implicit_value", ["nifti"])
def test_bla(foo):
    # is passed implicitly to foo
    assert foo == 1

@pytest.mark.parametrize("implicit_value", ["nifti"])
def test_bla_2(foo, implicit_value):
    # is passed a bit more explicitly to foo
    assert foo == 1

def test_bar(foo):
    #  will fail
    assert foo == 1

def test_baz(foo, implicit_value="nifti"):
    #  will fail
    assert foo == 1
```

nilearn/decomposition/tests/test_dict_learning.py
Remi-Gau: for this one it seems that the mock data has been tailored for this one to pass with volume data: make it pass for surface data is a headache

so I am not testing it with surface yet

Remi-Gau: would prefer to add a TODO and get it done in a later PR

nilearn/decomposition/tests/test_canica.py
Remi-Gau: This input validation check happens "very early" so it should be fine testing only on volume

nilearn/decomposition/tests/test_canica.py
Remi-Gau: for this one it seems that the mock data has been tailored for this one to pass with volume data: make it pass for surface data is a headache

so I am not testing it with surface yet

could be done in a later PR

nilearn/decomposition/tests/test_canica.py
man-shu: I feel like the `test_bla_2` option only delays the confusion from arising a bit.
Maybe we could just add a comment pointing to the pytest docs that mention this implicit parameterization?

Remi-Gau:
OK I have actually made sure almost all tests are on surface data.

Also tried to the number of warnings throws when running tests

nilearn/decomposition/tests/test_canica.py
man-shu: What about using this `indirect` parameter: https://docs.pytest.org/en/stable/example/parametrize.html#indirect-parametrization?

It does add a bit of code but at least signals that something is happening "indirectly"

Remi-Gau:
```python
import pytest

@pytest.fixture
def foo(request):
    if request.param == "nifti":
        return 1
    else:
        return 2
    
@pytest.fixture
def bar(request):
    if request.param == "nifti":
        return "bla"
    else:
        return "boop"

@pytest.mark.parametrize("foo", ["nifti"], indirect=True)
@pytest.mark.parametrize("bar", ["nifti"], indirect=True)
def test_bla_3(bar, foo):
    assert foo == 1
    assert bar == "bla"
```

not a big fan because there is no way to define your parametrization once for all the fixture that depend on it

data_type is passed to a bunch of fixtures in our case

if I do

```python
@pytest.mark.parametrize("foo", ["nifti", "surface"], indirect=True)
@pytest.mark.parametrize("bar", ["nifti", "surface"], indirect=True)
def test_bla_3(bar, foo):
    assert foo == 1
    assert bar == "bla"
```

it will pass inconsistent values across fixtures

```
FAILED test_foo.py::test_bla_3[surface-surface] 
FAILED test_foo.py::test_bla_3[nifti-surface]
FAILED test_foo.py::test_bla_3[surface-nifti]
```

nilearn/_utils/docs.py
```diff
@@ -1260,6 +1265,53 @@ def custom_function(vertices):
...
+
+            - for Nifti images, this contains automatically computed mask
+              via the selected ``mask_strategy``.
+
+            - for SurfaceImage objects, this mask encompasses all vertices of
+              the input images.
+        """
+
+docdict["multi_pca_attributes"] = """
+        masker_ : instance of MultiNiftiMasker or \
```
hndgzkn:
```suggestion
        masker_ :  :obj:`~nilearn.maskers.MultiNiftiMasker` or \
```

nilearn/_utils/docs.py
```diff
@@ -1260,6 +1265,53 @@ def custom_function(vertices):
...
+                :obj:`~nilearn.maskers.SurfaceMasker`
+            Masker used to filter and mask data as first step.
+            If an instance of MultiNiftiMasker
```
hndgzkn:
```suggestion
            If :obj:`~nilearn.maskers.MultiNiftiMasker`
```

nilearn/decomposition/_base.py
```diff
@@ -32,6 +34,45 @@
...
+    for param in params_to_ignore:
+        if param in tmp:
+            if (
+                tmp[param].default is None
+                and getattr(estimator, param) is not None
+            ):
+                # this should catch when user passes a numpy array
+                ignored_params.append(param)
+            elif getattr(estimator, param) != tmp[param].default:
+                ignored_params.append(param)
```
hndgzkn, line 62: Isn't this condition enough to cover the previous one, may be I'm missing something?

nilearn/decomposition/_base.py
```diff
@@ -112,12 +153,12 @@ def _mask_and_reduce(
 
     Parameters
     ----------
-    masker : NiftiMasker or MultiNiftiMasker or
-    :obj:`~nilearn.maskers.SurfaceMasker`
+    masker : NiftiMasker or MultiNiftiMasker or \
+        :obj:`~nilearn.maskers.SurfaceMasker`
```
hndgzkn:
```suggestion
    masker : :obj:`~nilearn.maskers.NiftiMasker` or \ 
        :obj:`~nilearn.maskers.MultiNiftiMasker` or \
        :obj:`~nilearn.maskers.SurfaceMasker`
```

nilearn/decomposition/_base.py
```diff
@@ -267,18 +310,21 @@ class _BaseDecomposition(CacheMixin, TransformerMixin, BaseEstimator):
     Parameters
     ----------
     n_components : int, default=20
-        Number of components to extract, for each 4D-Niimage
+        Number of components to extract,
+        for each 4D-Niimage or each 2D surface image
 
     %(random_state)s
 
     mask : Niimg-like object or MultiNiftiMasker instance or
```
hndgzkn:
```suggestion
    mask : Niimg-like object,  :obj:`~nilearn.maskers.MultiNiftiMasker` or
```

nilearn/decomposition/_multi_pca.py
```diff
@@ -32,22 +32,20 @@ class _MultiPCA(_BaseDecomposition):
 
     %(smoothing_fwhm)s
 
-    mask : Niimg-like object, instance of NiftiMasker
-        or MultiNiftiMasker, optional
+    mask : Niimg-like object, instance of NiftiMasker or MultiNiftiMasker or \
```
hndgzkn:
```suggestion
    mask : Niimg-like object, :obj:`~nilearn.maskers.NiftiMasker` or \ 
          :obj:`~nilearn.maskers.MultiNiftiMasker` or \
```

nilearn/decomposition/canica.py
```diff
@@ -22,14 +22,17 @@ class CanICA(_MultiPCA):
 
     Parameters
     ----------
-    mask : Niimg-like object or MultiNiftiMasker instance, optional
+    mask : Niimg-like object or MultiNiftiMasker instance or \
```
hndgzkn:
```suggestion
    mask : Niimg-like object, :obj:`~nilearn.maskers.MultiNiftiMasker` or \
```

nilearn/decomposition/dict_learning.py
```diff
@@ -41,15 +41,17 @@ class DictLearning(_BaseDecomposition):
 
     See :footcite:t:`Mensch2016`.
 
-     .. versionadded:: 0.2
+    .. versionadded:: 0.2
 
     Parameters
     ----------
-    mask : Niimg-like object or MultiNiftiMasker instance, optional
+    mask : Niimg-like object or MultiNiftiMasker instance or \
```
hndgzkn:
```suggestion
    mask : Niimg-like object, :obj:`~nilearn.maskers.MultiNiftiMasker` or \
```

nilearn/decomposition/tests/test_decomposition_estimators.py
hndgzkn: Yes, it seems like it is not used.

hndgzkn (approved):
A part from minor comments or suggestions, seems Ok to me.

nilearn/decomposition/tests/test_decomposition_estimators.py
Remi-Gau: see convo here: https://github.com/nilearn/nilearn/pull/5377#discussion_r2144582199

nilearn/decomposition/_base.py
Remi-Gau: without the previous one we can not evaluate the "truthyness" of numpy arrays when comparing to None:

```python
if np.asarray([1, 5]) == None:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

nilearn/decomposition/tests/test_decomposition_estimators.py
hndgzkn: Sorry and thank you, I've missed it.

Remi-Gau: if you know or find ways to make this use of fixture more explicit, let me know because I am not super happy with this TBH

Remi-Gau:
There is one test that fails only for python 3.9 and mac OS with the latest dependencies:

https://github.com/nilearn/nilearn/actions/runs/15678389371/job/44163941203?pr=5377#step:7:16636

No idea what is happening there. @man-shu ?

nilearn/decomposition/tests/test_decomposition_estimators.py
hndgzkn: I'd say `test_bla_2` would be Ok, I am familiar with that. I did not know `test_bla` also worked.

hndgzkn: But then may be ruff will complain?

Remi-Gau: yeah but we can silence it. :-p

nilearn/decomposition/tests/conftest.py
```diff
@@ -0,0 +1,319 @@
...
+            "right": np.ones(
+                (decomposition_mesh.parts["right"].coordinates.shape[0],)
+            ),
+        }
+        return SurfaceImage(mesh=decomposition_mesh, data=mask_data)
+
+    # TODO
+    # setting the shape of the mask to be a bit different
+    # shape_3d_large that is used for the data
+    # to force resampling
```
bthirion, line 59: You mean for images ?

nilearn/decomposition/tests/test_decomposition_estimators.py
```diff
@@ -0,0 +1,312 @@
...
+
+    est.fit(canica_data_single_img)
+
+    with pytest.deprecated_call(
+        match="The 'nifti_maps_masker_' attribute is deprecated"
+    ):
+        est.nifti_maps_masker_  # noqa: B018
+
+
+# TODO passing confounds does not affect output with CanICA, DictLearning
```
bthirion, line 188: I'm a bit lost here. Why is it the case ?

bthirion (approved):
LGTM overall. I just have a hard time understanding some details.

nilearn/decomposition/dict_learning.py
```diff
@@ -121,14 +124,12 @@ class DictLearning(_BaseDecomposition):
...
-             Depending on this value, the mask will be computed from
-             :func:`nilearn.masking.compute_background_mask`,
-             :func:`nilearn.masking.compute_epi_mask`, or
-             :func:`nilearn.masking.compute_brain_mask`.
-
         Default='epi'.
 
+        .. note::
+            These strategies are only relevant for Nifti images and the
+            parameterm is ignored for SurfaceImage objects.
```
hndgzkn:
```suggestion
            parameter is ignored for SurfaceImage objects.
```

nilearn/decomposition/tests/conftest.py
```diff
@@ -0,0 +1,319 @@
...
+
+    return nii_imgs
+
+
+@pytest.fixture
+def decomposition_data_single_img(
+    decomposition_data,
+) -> Union[SurfaceImage, Nifti1Image]:
+    """Return a single image for decomposition."""
+    return decomposition_data[0]
```
hndgzkn: Could this be done other way round? I mean this could generate single image and it can be used in image generation loop in `decomposition_data` to avoid generating 5 images when you only need 1.

nilearn/decomposition/tests/conftest.py
```diff
@@ -0,0 +1,319 @@
...
+    for mp in components:
+        assert mp.max() <= -mp.min()  # Goal met ?
+
+    return components
+
+
+@pytest.fixture
+def canica_data_single_img(canica_data) -> Nifti1Image:
+    """Create a canonical ICA data for testing purposes."""
+    return canica_data[0]
```
hndgzkn, line 346: The same here?

hndgzkn (approved):
A part from minor comments, LGTM.

nilearn/decomposition/tests/conftest.py
Remi-Gau: yeah all of those fixtures we use for testing decomposition on volume data, tend to generate data (image and mask) that have same the same shape and affine, so we never properly test that things go as expected if the we had a mask with a different field of view from the data (or the other way around)

I am pretty sure that we do not see it in terms of coverage because this is handled by the maskers and so this part of the code base is tested in other ways, but I'd prefer our tests for decomposition to not just test the "happy path".

nilearn/decomposition/tests/conftest.py
Remi-Gau: let me try

.github/workflows/test_with_tox.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```

apply before merging

nilearn/decomposition/tests/conftest.py
```diff
@@ -0,0 +1,332 @@
...
+
+    return surf_imgs
+
+
+@pytest.fixture
+def decomposition_images(
+    decomposition_img,
+) -> Union[list[SurfaceImage], list[Nifti1Image]]:
+    """Create "multi-subject" dataset with fake activation."""
+    return [decomposition_img for _ in range(N_SUBJECTS)]
```
hndgzkn: I might be mistaken, but if `rng` is not in signature of `decomposition_images`, the list of images, created here will all be the same.

hndgzkn: May be it should have all the parameters of `decomposition_img`?

Remi-Gau: yeah I have now reorganized things a bit

nilearn/decomposition/tests/test_decomposition_estimators.py
Remi-Gau: Not sure: I was just noting it down as something to investigate later (out of the scope of this PR), I was not trying to suggest that this is the expected behavior.

- I wonder if this relates to https://github.com/nilearn/nilearn/issues/3428

PR #5389, opened 2025-05-19
[ENH] Refactor to separate backends for plot_surf_stat_map

hndgzkn:
- Closes None, but relates to:
  - #5234 
  - #5253 
  - #5276
  - #5282 
  - #5322
  - #5349

Changes proposed in this pull request:

-  Refactor `plotting.surface.surf_plotting.plot_surf_stat_maps` to separate matplotlib code to `_matplotlib_backend` and plotly code to `_plotly_backend` and common parts to `backend`
- Add/update missing docstrings for tests in `test_surf_plotting` for `plot_surf_stat_maps`

PR #5391, opened 2025-05-19
[ENH] Refactor to separate backends for plot_surf_roi

hndgzkn:
  - #5234 
  - #5253 
  - #5276
  - #5282 
  - #5322
  - #5349
  - #5389

-  Refactor `plotting.surface.surf_plotting.plot_surf_roi` to separate matplotlib code to `_matplotlib_backend` and plotly code to `_plotly_backend` and common parts to `backend`
- Add/update missing docstrings for tests in `test_surf_plotting` for `plot_surf_roi`

### DIFF
--- a/nilearn/plotting/surface/_backend.py
+++ b/nilearn/plotting/surface/_backend.py
@@ -160,109 +160,6 @@ class BaseSurfaceBackend:
             figure=figure,
         )
 
-    def plot_surf_roi(
-        self,
-        surf_mesh=None,
-        roi_map=None,
-        bg_map=None,
-        hemi=DEFAULT_HEMI,
-        view=None,
-        cmap="gist_ncar",
-        colorbar=True,
-        avg_method=None,
-        threshold=1e-14,
-        alpha=None,
-        bg_on_data=False,
-        darkness=0.7,
-        vmin=None,
-        vmax=None,
-        cbar_tick_format="auto",
-        title=None,
-        title_font_size=None,
-        output_file=None,
-        axes=None,
-        figure=None,
-        **kwargs,
-    ):
-        check_params(locals())
-        roi_map, surf_mesh, bg_map = check_surface_plotting_inputs(
-            roi_map, surf_mesh, hemi, bg_map
-        )
-        check_extensions(roi_map, DATA_EXTENSIONS, FREESURFER_DATA_EXTENSIONS)
-
-        # preload roi and mesh to determine vmin, vmax and give more useful
-        # error messages in case of wrong inputs
-        roi = load_surf_data(roi_map)
-
-        if roi.ndim != 1:
-            raise ValueError(
-                "roi_map can only have one dimension but has "
-                f"{roi.ndim} dimensions"
-            )
-        if (roi < 0).any():
-            # TODO raise ValueError in release 0.13
-            warn(
-                (
-                    "Negative values in roi_map will no longer be allowed in"
-                    " Nilearn version 0.13"
-                ),
-                DeprecationWarning,
-                stacklevel=find_stack_level(),
-            )
-
-        mesh = load_surf_mesh(surf_mesh)
-        if roi.shape[0] != mesh.n_vertices:
-            raise ValueError(
-                "roi_map does not have the same number of vertices "
-                "as the mesh. If you have a list of indices for the "
-                "ROI you can convert them into a ROI map like this:\n"
-                "roi_map = np.zeros(n_vertices)\n"
-                "roi_map[roi_idx] = 1"
-            )
-
-        idx_not_na = ~np.isnan(roi)
-        if vmin is None:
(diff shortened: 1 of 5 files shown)


nilearn/plotting/surface/_backend.py
```diff
@@ -194,6 +196,109 @@ def plot_surf_contours(
...
+            # TODO raise ValueError in release 0.13
+            warn(
+                (
+                    "Negative values in roi_map will no longer be allowed in"
+                    " Nilearn version 0.13"
+                ),
+                DeprecationWarning,
+                stacklevel=find_stack_level(),
+            )
+
```
hndgzkn, line 248: I've moved these checks above.

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -1011,94 +1006,27 @@ def plot_surf_roi(
...
-    check_params(locals())
-    if view is None:
-        view = "dorsal" if hemi == "both" else "lateral"
-
-    roi_map, surf_mesh, bg_map = check_surface_plotting_inputs(
-        roi_map, surf_mesh, hemi, bg_map
-    )
-
-    if engine == "matplotlib" and avg_method is None:
-        avg_method = "median"
```
hndgzkn, line 1024: Is there a special reason that `avg_method=median` for `plot_surf_roi` or it was a random choice? 

In current implementation, I leave `plot_surf` to adjust this value, which sets `avg_method=mean`. If it is convenient, I'll leave the code as is, otherwise I'll update to set it as `median`.

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -1011,94 +1006,27 @@ def plot_surf_roi(
...
-            (
-                "Non-integer values in roi_map will no longer be allowed in"
-                " Nilearn version 0.13"
-            ),
-            DeprecationWarning,
-            stacklevel=find_stack_level(),
-        )
-
-    if cbar_tick_format == "auto":
-        cbar_tick_format = "." if engine == "plotly" else "%i"
```
hndgzkn, line 1075: Is there a special reason for the value of `cbar_tick_format` or it was a random choice?

In current implementation, I leave `plot_surf` to adjust this value, which sets 
for `matplotlib`:
```
cbar_tick_format = (
            "%.2g" if cbar_tick_format == "auto" else cbar_tick_format
        )
```
for `plotly`:
```
        cbar_tick_format = (
            ".1f" if cbar_tick_format == "auto" else cbar_tick_format
        )
```
 If it is convenient, I'll leave the code as is, otherwise I'll update.

nilearn/plotting/surface/tests/test_surf_plotting.py
```diff
@@ -877,6 +894,55 @@ def test_plot_surf_parcellation_plotly(
...
+        roi_map=roi_map,
+        colorbar=True,
+        engine="plotly",
+        **kwargs,
+    )
+    assert (
+        mpl_plot.axes[-1].get_ylim()[0] == plotly_plot.figure.data[1]["cmin"]
+    )
+
+
```
hndgzkn, line 945: I've just moved these 2 tests up in the file to keep all `plot_surf_roi` function tests together.

nilearn/plotting/surface/surf_plotting.py
Remi-Gau: I don't remember and I think the choice was made before I joined the team, so not sure.
I'd suggest leaving it as is.

hndgzkn: I just want to confirm that leaving as is, I meant `avg_method=mean` set by `plot_surf` which is different than `avg_method=median` in original code for `plot_surf_roi`.

Remi-Gau (changes requested):
some tests are failing when comparing to the baseline figures

may be worth checking what is causing the issue

hndgzkn:
> some tests are failing when comparing to the baseline figures
> 
> may be worth checking what is causing the issue

Yes, I've seen that. I think it is because of change of default values for `avg_method` and `cbar_tick_format`. 

I decided to set the previous default values.  Like `hemi=both` value, I think anything that changes the base images, we should not do it before all refactoring in the package is finished. What do you think?

Remi-Gau:
> I think anything that changes the base images, we should not do it before all refactoring in the package is finished, except for bugs. What do you think?

very strong agree

let's try to figure out, I would "revert" changes that were done to make sure those tests pass first.

Remi-Gau:
I think avg_method must be responsible for the biggest part of those failures.

hndgzkn:
> I think avg_method must be responsible for the biggest part of those failures.

At the end, both were responsible. It is fixed now.

Remi-Gau (approved):
LGTM

man-shu (approved):
LGTM!

PR #5359, opened 2025-04-30
[MAINT] ensure warning is thrown if estimators get non finite values in data to transform

Remi-Gau:
- Closes none but relates #4538

Changes proposed in this pull request:

- adds an equivalent to sklearn check_estimators_nan_inf

https://github.com/scikit-learn/scikit-learn/blob/19a6e61b6f8a7bb8a6d9dd5e8a5d40a741de28c3/sklearn/utils/estimator_checks.py#L2315

PR #5392, opened 2025-05-19
[FIX] avoid masking out vertices with zero-valued timepoints

### LINKED ISSUE #5390: [BUG] SurfaceMasker masks the vertices where even one timepoint is zero, when input mask is None

### Operating system

- [ ] Linux
- [x] Mac
- [ ] Windows

### Operating system version

- Mac OS Version 15.4.1 "Sequoia"

### Python version

- [ ] 3.13
- [x] 3.12
- [ ] 3.11
- [ ] 3.10
- [ ] 3.9

### nilearn version

0.11.2.dev306+gb9d6746b9

### Expected behavior

IIUC `SurfaceMasker.fit(img)` should ideally create a mask that keeps all the vertices in the input image if no mask is provided (input mask is None).

But current behavior masks the vertices that have at least one 0 value in the data.
I think we missed this because we do not have a surf image fixture that has zero values in it.

If we agree this is a bug, we could do something like the following:

```python
# Paste your code here
from nilearn.maskers import SurfaceMasker
from nilearn.conftest import _make_mesh
import numpy as np

mesh = _make_mesh()

# create first 2d surface image
data_left_1 = np.asarray(
    [
        [1, 0, 2, 4, 5],      # <---- this vertex will be masked just because one of the timepoints is 0
        [2, 2, 2, 4, 5],
        [3, 4, 2, 4, 5],
        [5, 6, 10, 4, 5],
    ]
)
data_right_1 = np.asarray(
    [
        [1, 4, 2, 4, 5],
        [2, 2, 2, 4, 5],
        [3, 4, 2, 4, 5],
        [5, 6, 10, 4, 5],
        [5, 6, 10, 4, 5],
    ]
)
data_1 = {
    "left": data_left_1,
    "right": data_right_1,
}

# %%
# create second 2d surface image
data_left_2 = np.asarray(
    [
        [1, 0, 2, 4, 5],     # <---- this vertex is already masked due to the first image
        [2, 2, 2, 4, 5],     
        [3, 0, 2, 4, 5],     # <---- same here
        [5, 0, 10, 4, 5],    # <---- and here
    ]
)
data_right_2 = np.asarray(
    [
        [2, 2, 2, 4, 5],
        [3, 4, 2, 4, 5],
        [5, 6, 10, 4, 5],
        [5, 6, 10, 4, 5],
        [5, 6, 10, 4, 5],
    ]
)
print(data_left_2.shape)
print(data_right_2.shape)
data_2 = {
    "left": data_left_2,
    "right": data_right_2,
}


# %%
surface_img_1 = SurfaceImage(mesh=mesh, data=data_1)
surface_img_2 = SurfaceImage(mesh=mesh, data=data_2)

masker = SurfaceMasker()
masker.fit([surface_img_1, surface_img_2])

n_true_vertices = np.sum(masker.mask_img_.data.parts["left"]) + np.sum(
    masker.mask_img_.data.parts["right"]
)
print(f"{masker.mask_img_.shape=}")
print(f"{n_true_vertices=}")

assert masker.mask_img_.shape == (n_true_vertices,)     
# we will only have 6 True vertices, when it should be 9

```

man-shu:
- Closes #5390

- [x] directly mask out vertices with NaN or infinite values
- [ ] add tests

### DIFF
--- a/nilearn/maskers/surface_masker.py
+++ b/nilearn/maskers/surface_masker.py
@@ -182,10 +182,8 @@ class SurfaceMasker(_BaseSurfaceMasker):
         img = at_least_2d(img)
         mask_data = {}
         for part, v in img.data.parts.items():
-            mask_data[part] = v.astype("float32")
-            non_finite_mask = np.logical_not(np.isfinite(mask_data[part]))
-            mask_data[part][non_finite_mask] = 0
-            mask_data[part] = mask_data[part].astype("bool").all(axis=1)
+            # mask out vertices with NaN or infinite values
+            mask_data[part] = np.isfinite(v.astype("float32")).all(axis=1)
         self.mask_img_ = SurfaceImage(mesh=img.mesh, data=mask_data)
 
     @rename_parameters(


nilearn/maskers/surface_masker.py
```diff
@@ -182,10 +182,8 @@ def _fit_mask_img(self, img):
         img = at_least_2d(img)
         mask_data = {}
         for part, v in img.data.parts.items():
-            mask_data[part] = v.astype("float32")
-            non_finite_mask = np.logical_not(np.isfinite(mask_data[part]))
-            mask_data[part][non_finite_mask] = 0
-            mask_data[part] = mask_data[part].astype("bool").all(axis=1)
+            # mask out vertices with NaN or infinite values
+            mask_data[part] = np.isfinite(v.astype("float32")).all(axis=1)
```
Remi-Gau, line 186: not sure what happens with nifti maskers but given that maskers should throw a warning if the data to transform contains non finite values (see https://github.com/nilearn/nilearn/pull/5359), it may be good to make sure they also warn if the image used to create the mask also contain non finite values

man-shu: Yeah I saw such a warning under _BaseSurfaceMasker too:

https://github.com/nilearn/nilearn/blob/6aa56e07b4bba646075066fe9e90df3b5fc1050d/nilearn/maskers/base_masker.py#L507

nilearn/maskers/tests/test_surface_masker.py
```diff
@@ -32,11 +32,14 @@ def test_check_estimator_invalid(estimator, check, name):  # noqa: ARG001
...
 
-    resulting mask should have a single 'timepoint'.
+    - resulting mask should have a single 'timepoint'
+    - all vertices should be included in the mask, because no mask is provided
+
     """
     masker = SurfaceMasker()
     masker.fit([surf_img_2d(3), surf_img_2d(5)])
     assert masker.mask_img_.shape == (surf_img_2d(1).shape[0],)
+    assert masker.mask_img_.shape == (masker.n_elements_,)
```
man-shu, line 42: tests all vertices are included in the mask

nilearn/maskers/surface_masker.py
man-shu: This PR is fairly small so I can add the warning here

man-shu: done!

man-shu:
Ok some unexpected failures, so the reviews can wait

nilearn/maskers/surface_masker.py
```diff
@@ -182,10 +182,13 @@ def _fit_mask_img(self, img):
...
+            warn(
+                "Non-finite values detected in the input image. "
+                "The computed mask will mask out these vertices.",
+                stacklevel=find_stack_level(),
+            )
```
hndgzkn, line 191: This line:

```
mask_data[part] = np.isfinite(v.astype("float32")).all(axis=1)
```

does not guarantee that there are infinite values. Shouldn't there be a condition that checks if infinite values exist and then give the warning? May be I am missing something.

hndgzkn (approved):
Seems ok to me a part from one comment.

nilearn/maskers/surface_masker.py
Remi-Gau: yeah I think that otherwise the warning is always thrown, no?

Remi-Gau: - BTW #5359 should include a single function to help dealing with non finite values in several places

man-shu: oops yes you're right

Remi-Gau (approved):
Let's wait for CI to finish but otherwise LGTM

PR #5396, opened 2025-05-21
[ENH] add surface support to math_img and binarize_img

### LINKED ISSUE #5289: add surface support for math_img

Remi-Gau:
- Closes #5289 
- Relates to #5272

- add surface support to math_img and binarize_img
- add tests
- fix some other doc strings in nilearn.image

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -48,7 +48,7 @@ from nilearn.surface.surface import (
     extract_data,
 )
 from nilearn.surface.surface import get_data as get_surface_data
-from nilearn.surface.utils import check_polymesh_equal
+from nilearn.surface.utils import assert_polymesh_equal, check_polymesh_equal
 from nilearn.typing import NiimgLike
 
 
@@ -1232,8 +1232,11 @@ def math_img(formula, copy_header_from=None, **imgs):
     imgs : images (:class:`~nibabel.nifti1.Nifti1Image` or file names \
            or :obj:`~nilearn.surface.SurfaceImage` object)
         Keyword arguments corresponding to the variables in the formula as
-        images. All input images should have the same 'geometry' (shape,
-        and affine for volume data, and mesh for surface data).
+        images.
+        All input images should have the same 'geometry':
+
+        - shape and affine for volume data
+        - mesh (coordinates and faces) for surface data
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


nilearn/image/image.py
```diff
@@ -1279,6 +1285,36 @@ def math_img(formula, copy_header_from=None, **imgs):
...
+            check_polymesh_equal(first_img.mesh, image.mesh)
+
+        # Computing input data as a dictionary of numpy arrays.
+        data_dict = {k: {} for k in first_img.data.parts}
+        for key, img in imgs.items():
+            for k, v in img.data.parts.items():
+                data_dict[k][key] = v
+
+        # Add a reference to numpy in the kwargs of eval
+        # so that numpy functionscan be called from there.
```
man-shu, line 1302:
```suggestion
        # so that numpy functions can be called from there.
```

nilearn/image/image.py
```diff
@@ -1279,6 +1285,36 @@ def math_img(formula, copy_header_from=None, **imgs):
     in FSL.
 
     """
+    is_surface = all(isinstance(x, SurfaceImage) for x in imgs.values())
+
+    if is_surface:
+        first_img = next(iter(imgs.values()))
+        for image in imgs.values():
+            check_polymesh_equal(first_img.mesh, image.mesh)
```
hndgzkn, line 1293: I think this might throw exception in case the meshes are not equal. It might be good to catch the exception and display a corresponding message to user.

nilearn/image/image.py
```diff
@@ -1360,13 +1397,16 @@ def binarize_img(
...
+
         .. versionadded:: 0.11.0
 
         This parameter will be set to True by default in 0.13.0.
 
     Returns
     -------
     :class:`~nibabel.nifti1.Nifti1Image`
+    or :obj:`~nilearn.surface.SurfaceImage`
         Binarized version of the given input image. Output dtype is int.
```
hndgzkn, line 1410:
```suggestion
        Binarized version of the given input image. Output dtype is int8.
```

hndgzkn (approved):
Otherwise ok.

man-shu (approved):
LGTM!

nilearn/image/image.py
hndgzkn: > isn't that what the function already does:

Yes, it does :) I had thought some message more specific to this function. But not really necessary may be.

hndgzkn: > here I have assumed that it is enough for the meshes to be compatible (same number of vertices), even if they do not have exactly the same geometry

I cannot comment on if we should be more strict or not, however in docstring it might be better to clearly indicate what is expected.

In that case, the exception message to user can be made more clear with the expected equality of meshes.

Remi-Gau: actually I will make it more strict: if users complain we can revisit it later

PR #5399, opened 2025-05-22
[MAINT] avoid extra warning when computing fixed effects

Remi-Gau:
- Closes none

- return z score image when computing fixed effects as this will be the new default in 0.13


bthirion (approved):
LGTM, thx.

PR #5400, opened 2025-05-22
[REF] extract nilearn check_estimators

Remi-Gau:
- Closes none

- refactor estimator checks to run the nilearn ones and the sklearn ones separately


bthirion:
LGTM. There are a few codecov complaints.

Remi-Gau:
> LGTM. There are a few codecov complaints.

yeah that's because some parts are the code are only tested with our oldest supported sklearn and the tests only run on main

PR #5409, opened 2025-05-30
[ENH] add testing for tedana load_confounds support

miltoncamacho:
This gets started with the integration of support for tedana confounds:

Closes #5386 

Changes proposed in this pull request:
Added dummy data to validate the functionality of the load_confounds function when using tedana strategy

Issue #5325, opened 2025-04-10
Check if the load_confounds_strategy still works with fmripost-aroma


man-shu:
> check if the load_confounds_strategy still works with fmripost-aroma

@htwangtw could you please look into this? I have no easy way to test with fmriprep outputs, and I don't think we have tests for that either.

From reading the [fmripost-aroma doc](https://fmripost-aroma.readthedocs.io/latest/outputs.html) superficially, I think the suffixes of the outputs have changed, and this function load_confounds_strategy may not work with the new ones.

_Originally posted by @man-shu in https://github.com/nilearn/nilearn/issues/5324#issuecomment-2792025051_

man-shu:
I see. Maybe @tsalo can help us with that.

Here's some context @tsalo:

We saw that since version 24.0.0 fMRIPrep does not perform ICA-AROMA anymore, and seems like the suffixes of files produced by fmripost-aroma are different from what we expect in `load_confounds_strategy`: https://nilearn.github.io/stable/modules/generated/nilearn.interfaces.fmriprep.load_confounds_strategy.html#nilearn.interfaces.fmriprep.load_confounds_strategy

Could you please provide some pointers to update this function to work with fmripost-aroma?

Thanks!

man-shu:
Thanks @tsalo ! I think @htwangtw wants to go with 2.

> If we can get the headers for the confound file that would be great

Do you have a sample confound file that you could share with us?

htwangtw:
@tsalo thank you!! This is fantastic! 

There are three action items to complete for us to adapt `fmripost-aroma`:

1. understand how to use these output to create ICA-AROMA denoised data that's aligned with the original paper's recommendation.
2. add new option (`orthaggrDenoised`) to the low level API `load_confound` and modify relevant files associated with the three denoised files.
3. Update the implementation in the high level API `load_confound_strategy`

We should get things ironed out for item 1 before going into the code!

I am trying to digest the conversation above and respond:

> I think @htwangtw wants to go with 2.

Not quite, it will be a combination of 1 and 2, depending on what has already been done! The aim of `load_confounds` is to implement denoising methods compliant to `fmriprep` (in this case `fmripost-aroma`). If `fmripost-aroma` generates a file that requires no further ICA-AROMA regressors, `load_confounds` will simply verify a) the nifti file is the correct file for the give strategy and b) provide confound variables for any addition requested denoising. We will not redo any steps `fmriprep` / `fmripost-aroma`

[...]

PR #5410, opened 2025-05-30
[ENH] add tedana support for load_confounds

### LINKED ISSUE #5386: [ENH] Multi Echo FMRI Processing. Add tedana confounds to fMRIprep load_confounds.py

For BIDS-flux and the [Canadian Pediatric Imaging Platform](https://www.cpipstudy.org/), @pbellec @bpinsard, we’re exploring the best approach to incorporate Tedana-derived rejected non-BOLD components (identified through a combination of automated and manual classification) into the regression pipeline alongside confounders from fMRIPrep.

Our objective is to minimize the number of transformations applied to the data when integrating these additional confounds and provide a tested tool that is able to do this to the research community pushing for standardization of processes.

Current pipeline steps:

BIDS conversion

fMRIPrep preprocessing

    using --dummy-scans <n of dummy scans>

    using output single echo processing data flag --me-output-echos

fMRIP outputs:

      pre-processed single echo runs to be used as inputs to tedana

     fMRIPrep confounders to be used as input to tedana

Tedana execution

    using a mask

    specifying the number of PCA components

    incorporating fMRIPrep confounders to inform the automatic ICA-based BOLD classification

Tedana outputs:

    Optimally combined image

    Denoised optimally combined image

    ICA time series (tsv)

    Component classification (accepted / rejected)

Next step:
We’re considering either extending or adapting [load_confounds.py](https://github.com/nilearn/nilearn/blob/7a8cd0ee7161c4ea5ce4c2df248f0113890e3070/nilearn/interfaces/fmriprep/load_confounds.py) from Nilearn to support loading the rejected component time series from Tedana as confound regressors for downstream analysis.

Since this could be a valuable addition for the community, we’d love to gather feedback early before proceeding with implementation.

### Benefits to the change

The goal is to minimize the number of transformation we perform in the data when loading the confounders as well as provide a tested tool that is able to do this to the research community pushing for standardization of processes.

```python
# insert your code below
nilearn.interfaces.fmriprep.load_confounds(img_files, denoise_strategy='simple', tedana_confound, tedana_component_status, **kwargs)
```

miltoncamacho:
Implemented support for loading confounds from tedana outputs.

- Closes #5386 

Adds support for another strategy "tedana" which allows you to load the rejected components computed using tedana.


nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -255,20 +283,26 @@ def get_confounds_file(image_file, flag_full_aroma):
...
+def get_json(confounds_raw_path, flag_tedana=False):
     """Return json data companion file to the confounds tsv file."""
-    # Load JSON file
-    return str(confounds_raw_path).replace("tsv", "json")
+    if flag_tedana:
+        # TEDANA does not have a json confound companion file
+        return None
+    else:
+        # Load JSON file
+        return str(confounds_raw_path).replace("tsv", "json")
```
Remi-Gau: we return early
else not needed

```suggestion
    # Load JSON file
    return str(confounds_raw_path).replace("tsv", "json")
```

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -237,14 +256,23 @@ def _get_file_name(nii_file):
...
+        raise ValueError(
+            f"Found more or less than two confound files:\n\t{found_str}"
+            "TEDANA should have two confound files mixing.tsv and table_status.tsv."
+        )
     else:
-        return found_files[0]
+        if flag_tedana:
+            return found_files
+        else:
+            return found_files[0]
```
Remi-Gau:
```python
    else:
        if flag_tedana:
            return found_files
        else:
            return found_files[0]
```

can be simplifed to

```python
     elif flag_tedana:
        return found_files

     return found_files[0]
```

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -403,19 +412,98 @@ def _load_confounds_for_single_image_file(
...
+            **kwargs,
+        )
+    else:
+        return _load_single_confounds_file(
+            confounds_file=confounds_file,
+            strategy=strategy,
+            demean=demean,
+            confounds_json_file=confounds_json_file,
+            **kwargs,
+        )
```
Remi-Gau: else not needed as we return early

```suggestion
    return _load_single_confounds_file(
            confounds_file=confounds_file,
            strategy=strategy,
            demean=demean,
            confounds_json_file=confounds_json_file,
            **kwargs,
        )
```

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -327,32 +365,44 @@ def load_confounds_file_as_dataframe(confounds_raw_path):
...
-    # check if the version of fMRIprep (>=1.2.0) is supported based on
-    # header format. 1.0.x and 1.1.x series uses camel case
-    if any(is_camel_case(col_name) for col_name in confounds_raw.columns):
-        raise ValueError(
-            "The confound file contains header in camel case. "
-            "This is likely the output from 1.0.x and 1.1.x series. "
-            "We only support fmriprep outputs >= 1.2.0."
-            f"{confounds_raw.columns}"
-        )
+    if not flag_tedana:
```
Remi-Gau: can you reverse the logic:
- deal with tedana first
- return early and do not use an else block

Remi-Gau (changes requested):
couple of changes to make the diff smaller

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -259,6 +264,9 @@ def load_confounds(
         - "full": use :term:`fMRIPrep` output
           `~desc-smoothAROMAnonaggr_bold.nii.gz`.
         - "basic": use noise independent components only.
+    
+    tedana : :obj:`str`, default="rejected"
+        - "rejected": use TEDANA rejected ICA components.
```
htwangtw: If there's only one option, do we really need to add a parameter here?

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -161,6 +164,8 @@ def _generate_confounds_file_candidates(nii_file):
     ----------
     nii_file : str
         Path to the functional image file.
+    include_ica : bool, optional
```
htwangtw: do you mean `flag_tedana`?

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -169,29 +174,41 @@ def _generate_confounds_file_candidates(nii_file):
...
-        for n_entities in range(1, len(entities.keys()) + 1)
-    ]
+    variants = []
 
-    # Flatten the list of lists
-    all_subsets = [list(item) for sublist in all_subsets for item in sublist]
-    # https://stackoverflow.com/a/3724558/2589328
-    unique_subsets = [list(x) for x in {tuple(x) for x in all_subsets}]
+    # Standard confounds
+    ent1 = deepcopy(entities)
```
htwangtw: avoid vague variable name such as `ent1`

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -169,29 +174,41 @@ def _generate_confounds_file_candidates(nii_file):
...
+    ent1["desc"] = "confounds"
+    variants.append(ent1)
 
-    # Require "desc"
-    unique_subsets = [subset for subset in unique_subsets if "desc" in subset]
+    if flag_tedana:
+        # ICA - mixing
+        ent2 = deepcopy(entities)
```
htwangtw: Same as above

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -169,29 +174,41 @@ def _generate_confounds_file_candidates(nii_file):
...
 
-    filenames = [
-        "_".join(["-".join([k, entities[k]]) for k in lst])
-        for lst in unique_subsets
-    ]
     return filenames
 
 
-def _get_file_name(nii_file):
+def _get_file_name(nii_file, flag_tedana=False):
```
htwangtw, line 234: docstring missing for this one

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -403,19 +412,98 @@ def _load_confounds_for_single_image_file(
...
+            **kwargs,
+        )
 
-    return _load_single_confounds_file(
-        confounds_file=confounds_file,
-        strategy=strategy,
-        demean=demean,
-        confounds_json_file=confounds_json_file,
-        **kwargs,
+def _load_tedana_confounds_file(
```
Remi-Gau: I am not sure we need another function _load_single_confound function specific to tedana

can't we just have all the logic under _load_single_confounds_file?

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -326,34 +362,44 @@ def load_confounds_file_as_dataframe(confounds_raw_path):
...
+        # TEDANA outputs are not camel case, but they have a different
+        # header format.
+        if any(col.startswith("ICA_") for col in confounds_raw.columns) or "Component" in confounds_raw.columns:
+            return confounds_raw
+        else:
+            raise ValueError(
+                "The confound file does not contain the expected columns for "
+                "TEDANA output. Expected 'ICA_xx' for mixing.tsv and 'Component' for the status_table.tsv columns."
+            )
+    else:
```
htwangtw: You don't really need the else here as the `if flag_tedana` block above has already be resolved, the rest can just stay the same. It will also help reduce the amount of differences between your changes and the original code and make things easier to read. There are a few instance of existing code block being moved and resulting differences in other places, I would suggest to revisit those and see if we can simplify the logic to avoid too many lines flagged as different

htwangtw (changes requested):
I had a pass on the code and we also need to make sure 
1. the pre-commit checks are resolved. 
2. Use the latest branch

Also, since the tests are now in this branch, shall we close #5409

Remi-Gau:
> shall we close #5409

yes

miltoncamacho:
The latest commit should pass the pytests, whenever you think it is ready I can clean up the history of the commits

Remi-Gau:
> clean up the history of the commits

no need: we do a squash-merge commit so only the name of the PR will appear in the final git history as a single commit

miltoncamacho:
This is an open dataset containing Tedana derivatives which can be used to test the new changes [doi:10.18112/openneuro.ds006190.v1.0.0](https://doi.org/10.18112/openneuro.ds006190.v1.0.0).

doc/changes/latest.rst
```diff
@@ -40,6 +40,8 @@ Fixes
 Enhancements
 ------------
 
+- :bdg-dark:`Code` Add tedana support for load_confounds and testing for tedana load_confounds support (:gh:`5410` by `Milton Camacho`_).
```
Remi-Gau, line 39: documentation build is failing because you need to add your details to the citation.cff file in the root of the repo

see more details in here https://nilearn.github.io/stable/development.html#changelog

Remi-Gau:
Some code needs to be tweaked to comply with out guidelines:

https://results.pre-commit.ci/run/github/1235740/1749171552.-f5czsJNQHqtHki4j44yog

Remi-Gau:
easiest way to check compliance with our guidelines

```bash
pip install tox
tox
```

htwangtw:
@miltoncamacho Looks like the test coverage can be improved.
I will have time to test it on the open neuro tedana derivative you shared next Monday!

doc/references.bib
```diff
@@ -1456,6 +1456,48 @@ @article{Varoquaux2010a
...
+	times (Kundu et al., 2013), and improving activation and connectivity mapping (Caballero-
+	Gaudes et al., 2019; Gonzalez-Castillo et al., 2016; Lynch et al., 2020) even in real time fMRI
+	(Heunis et al., 2020). In addition, artifactual non-T2* changes (known as S0 in this context)
+	may be identified and removed by leveraging the relationship between BOLD contrast and T2*
+	obtained with multi-echo fMRI (Kundu et al., 2012). Strategies to perform this efficiently and
+	robustly are in active development. Continuing these efforts, we present tedana (TE-Dependent ANAlysis) as an open-source
+	Python package for processing and denoising multi-echo fMRI data. tedana implements two
+	approaches to multi-echo preprocessing: (1) estimating a T2* map and using these values to
+	generate a weighted sum of individual echos, and (2) using echo-time dependent information
+	in analysis and denoising (Kundu et al., 2012).},
```
Remi-Gau:
```suggestion
```

I know that we have the abstract for quite a few other articles in our bib file, but I think we should not, so maybe let's not add more for now

Remi-Gau:
adding this one to our dev agenda: mostly to know whether we should try to include it in the 0.12 release

nilearn/interfaces/fmriprep/load_confounds.py
htwangtw: We need more documentation on what is the expected input here. See a few lines above for ICA AROMA as an example

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -18,6 +19,7 @@
     "aroma": "_desc-smoothAROMAnonaggr_bold",
     "nii.gz": "(_space-.*)?_desc-preproc_bold.nii.gz",
     "dtseries.nii": "(_space-.*)?_bold.dtseries.nii",
+    "tedana": "_desc-optcom_bold.nii.gz",
```
htwangtw, line 22: based on [`ds006190`](https://openneuro.org/datasets/ds006190/versions/1.0.0/download), there are three files that can be inputs to `load_confounds`:
- `*_desc-optcom_bold.nii.gz`
- `*_desc-optcomMIRDenoised_bold.nii.gz`
- `*_desc-denoised_bold.nii.gz`

I have a question: What do you think a reasonable behaviour would be if the user passes `_desc-optcomMIRDenoised_bold.nii.gz` or `_desc-denoised_bold.nii.gz`? (I genuinely don't know)

In the documentation, I see that `_desc-denoised_bold.nii.gz` was "Recommended dataset for analysis". It wouldn't make sense to throw an error if that is recommended by `tedana`.

The aim of `load_confounds` is to provide reasonable set of confound regressors for the given image file. If there's truly no further denoising to do with one file based on the given strategy, the expected behaviour would be to return an empty data frame. See the non-aggressive ICA-AROMA for example below:

```
from nilearn.interfaces.fmriprep.tests import test_load_confounds_utils as lcu
from pathlib import Path


img_path, cf_path =  lcu.create_tmp_filepath(Path('.'), image_type='ica_aroma', copy_confounds=True)

confound_df, sample_mask = load_confounds(img_path, strategy=['ica_aroma'])
assert confound_df.empty == True
```

@emdupre it would be great to get your guidance for this one

nilearn/interfaces/fmriprep/load_confounds_utils.py
```diff
@@ -209,14 +234,18 @@ def _get_file_name(nii_file):
...
+    )
 
     # fmriprep has changed the file suffix between v20.1.1 and v20.2.0 with
     # respect to BEP 012.
     # cf. https://neurostars.org/t/naming-change-confounds-regressors-to-confounds-timeseries/17637 # noqa: E501
     # Check file with new naming scheme exists or replace,
     # for backward compatibility.
     suffixes = ["_timeseries.tsv", "_regressors.tsv"]
+    if flag_tedana:  # tedana has different suffixes
+        suffixes = ["_mixing.tsv", "_status_table.tsv"]
```
htwangtw: This is leading to undesired behaviour.
I found out by testing the functionality in two different ways using the test data generated with `create_tmp_filepath` included in the tests:

1. Testing two strategies that can be mixed potentially, such as scrubbing + tedana denoising

Currently, the output of the following two use cases are identical:
```python
load_confounds(img_file_path, strategy=['tedana'])
load_confounds(img_file_path, strategy=['tedana', 'scrub'], fd_threshold=0.01)  # I am using an unreasonably small threashold here
```

We should include integration tests that would catch this kind of cases. 

2. Trying to use `load_confound to` recreate the first code example in [this tutorial](https://tedana.readthedocs.io/en/stable/denoising.html)

I was expecting to recreate it with 
```python
load_confounds(img_file_path, strategy=['tedana', 'motion', 'wm_csf'])
```
However I only get the ICA components here. 

I think the fix should be add `["_mixing.tsv", "_status_table.tsv"]` to the existing list of suffixes. There will be a lot of subsequent actions that will be triggered and changed.

nilearn/interfaces/fmriprep/tests/_testing.py
```diff
@@ -51,12 +57,23 @@
...
     path_data = Path(load_confounds_utils.__file__).parent / "data"
     suffix = "test-v21" if fmriprep_version == "21.x.x" else "test"
+
     if non_steady_state:
+        if tedana:
+            return [
+                path_data / filename
+                for filename in [
+                    "test_desc-ICA_mixing.tsv",
+                    "test_desc-ICA_status_table.tsv",
```
htwangtw:
```suggestion
                    f"{suffix}_desc-ICA_mixing.tsv",
                    f"{suffix}_desc-ICA_status_table.tsv",
```
Also, this might not be enough for testing if `~desc-confounds_regressors.tsv` is needed based on my understanding of [this tutorial](https://tedana.readthedocs.io/en/stable/denoising.html)

nilearn/interfaces/fmriprep/tests/_testing.py
```diff
@@ -97,6 +114,59 @@ def create_tmp_filepath(
     # create test files in temporary directory
     derivative = "regressors" if fmriprep_version == "1.2.x" else "timeseries"
 
+    if "tedana" in image_type:
+        # confound files
+        for suf in ["mixing", "status_table"]:
```
htwangtw:
```suggestion
        for suffix in ["mixing", "status_table"]:
```
Avoid ambiguous variable name

nilearn/interfaces/fmriprep/tests/test_load_confounds_utils.py
```diff
@@ -65,5 +67,44 @@ def test_get_file_name(tmp_path, flag, keyword, image_type):
...
+    """
+    func_file = tmp_path / "sub-01_task-rest_desc-optcom_bold.nii.gz"
+    func_file.write_text("dummy")
+
+    # TEDANA expects exactly these two suffixes, so we can test with one
+    for suffix in ["_mixing.tsv"]:
+        (tmp_path / f"sub-01_task-rest_desc-ICA{suffix}").write_text("dummy")
+
+    with pytest.raises(ValueError, match="expected 2 for TEDANA"):
+        _get_file_name(str(func_file), flag_tedana=True)
```
htwangtw: You should also add a case for tedana to `test_get_file_name` above for basic assessment of the filename keyword check. 
If there are additional tests needed for tedana, it's reasonable to have extra tests here, but you should use the existing utility `create_tmp_filepath` to generate files here for the test rather than create the files manually again.

nilearn/interfaces/fmriprep/tests/test_load_confounds_utils.py
```diff
@@ -65,5 +67,44 @@ def test_get_file_name(tmp_path, flag, keyword, image_type):
...
+
+
+def test_load_confounds_file_as_dataframe_tedana_invalid_columns(tmp_path):
```
htwangtw: See comments above for `test_get_file_name_raises_on_invalid_tedana_confounds`.

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
```diff
@@ -24,11 +24,28 @@
 from nilearn.tests.test_signal import generate_trends
 
 
-def _simu_img(tmp_path, trend, demean):
+def test_create_tmp_filepath_touch_confound_only(tmp_path):
```
htwangtw: This test is in the wrong place, it should be integrated to `test_get_file_name` in `test_load_confounds_utils.py`

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
```diff
@@ -24,11 +24,28 @@
...
+        tmp_path,
+        image_type="tedana",  # Any non-TEDANA image type
+        copy_confounds=False,  # This triggers tmp_conf.touch()
+    )
+
+    assert tmp_conf.exists()
+    assert tmp_conf.stat().st_size == 0  # File should be empty
+
+
+def _simu_img(tmp_path, trend, demean, image_type="regular"):
```
htwangtw: Remove any code changes to `_simu_img` and changes related to this change. The aim of the function is to generate a simulated image with motion injected as noise. We don't need to create specific simulated image for tedana.

This is very broad integration test for all denoising methods.

htwangtw (changes requested):
Thank you for putting the PR in! I know this is a lot of work! We have a lot to work with before the next round of the reivew. Please see each individual comment. 

The summary is the behaviour of the functions are not as I expected and the current added tests are not fit for purpose. There are no unit tests for the new tedana utility, which explains why the test coverage is dropping. 

I want to get back to the drawing board - I am basing off the [tutorial](https://tedana.readthedocs.io/en/stable/denoising.html) here for my review process, we can also use it to figure out the next steps.

We need to decide what are the expected behaviour of these three types of outputs from tedana (found in `ds006190`)

    *_desc-optcom_bold.nii.gz
    *_desc-denoised_bold.nii.gz
    *_desc-optcomMIRDenoised_bold.nii.gz

I cannot find documentation on the official docs for `_desc-optcomMIRDenoised_bold.nii.gz` so I need some help here to know what it is exactly.

I have a feeling that coming up with tests at this stage might not be a very concrete objective for @miltoncamacho, based on what I see here. I would encourage you to aim for implementing relevant changes in `load_confounds` so you can recreate the official tedana denoise tutorial with `ds006190`, and add example code for it for it in the comments so I can run it on my end for the next round of review. We can ignore if the tests are passing or not at this stage. 

If you need to come to the office hour to discuss this, give me a heads up in the comment here.

Here is a snippet of what I did to test the functionality of your code, given that I have downloaded `ds006190` locally at my home directory:

```python
from pathlib import Path


fun_dir_path = Path.home() / "ds006190/sub-24053/ses-1/func/" 

confounds, sample_masks = load_confounds(str(fun_dir_path / "sub-24053_ses-1_task-bao_dir-AP_run-01_desc-optcom_bold.nii.gz"), strategy=['tedana'])
assert 'ICA_0' in confounds.columns.tolist()

confounds, sample_masks = load_confounds(str(fun_dir_path / "sub-24053_ses-1_task-bao_dir-AP_run-01_desc-optcom_bold.nii.gz"), strategy=['tedana', 'motion', 'wm_csf'])

assert 'trans_x' in confounds.columns.tolist()
```
The above should break on the second assertion.

This one I imagined it should pass, but no:
```
# try a different file and it broke
confounds, sample_masks = load_confounds(str(fun_dir_path / "sub-24053_ses-1_task-rat_rec-nordic_dir-PA_run-01_desc-optcom_bold.nii.gz"), strategy=['tedana'])
```
output:
```
ValueError: Found 4 confound files (expected 2 for TEDANA):
	/home/haoting/ds006190/sub-24053/ses-1/func/sub-24053_ses-1_task-rat_rec-nordic_dir-PA_run-01_desc-ICA_status_table.tsv
	/home/haoting/ds006190/sub-24053/ses-1/func/sub-24053_ses-1_task-rat_rec-nordic_dir-PA_run-01_desc-ICA_mixing.tsv
	/home/haoting/ds006190/sub-24053/ses-1/func/sub-24053_ses-1_task-rat_dir-PA_run-01_desc-ICA_status_table.tsv
	/home/haoting/ds006190/sub-24053/ses-1/func/sub-24053_ses-1_task-rat_dir-PA_run-01_desc-ICA_mixing.tsv

TEDANA should produce exactly two confound files:
- mixing.tsv
- table_status.tsv
```

miltoncamacho:
> ValueError: Found 4 confound files (expected 2 for TEDANA):
> 	/home/haoting/ds006190/sub-24053/ses-1/func/sub-24053_ses-1_task-rat_rec-nordic_dir-PA_run-01_desc-ICA_status_table.tsv
> 	/home/haoting/ds006190/sub-24053/ses-1/func/sub-24053_ses-1_task-rat_rec-nordic_dir-PA_run-01_desc-ICA_mixing.tsv

`rec-nordic` file are not really outputs from tedana, we can just remove those for testing.

Not sure but these are likely produced by tsalo with additional extra-tedana processing.

miltoncamacho:
> If you need to come to the office hour to discuss this, give me a heads up in the comment here.

@htwangtw Yes lets meet

Remi-Gau:
@miltoncamacho 
We have a new release of nilearn coming soon.

I am going to turn to draft all PRs that are not expected to be finished for that release including yours. This is more for our own 'book keeping', so if your PR gets finished before the release then we can still merge it.

And no need to rush the PR to get it done in time: if it's not ready for this one, it will be for the following one. 😉

nilearn/interfaces/fmriprep/load_confounds_utils.py
emdupre: - `desc-optcom` is the optimal combination of the echos with no information removed
- `desc-denoised` is the timeseries after components that are classified as "noise" in the decision-tree are projected out of the optimally combined data
- `desc-optcomMIRDenoiseed` is the optimal combinations with Minimal intensity regression, which was designed to remove T1 noise but has limited empirical support (included in tedana for backwards compatibility with ME-ICA)

In general, we recommend using `desc-denoised` for analysis since this projects out noise while still retaining degrees of freedom. In the case of integration with `load_confounds`, though, it sounds like you'd like to do your own projection, which is why I imagine you're falling back to the optimal combination ? I haven't been closely tracking this PR, though, so I might be misunderstanding !

htwangtw: This is super useful to help me understand the context! I have a better idea now how this will align with the behaviour of existing similar use case (i.e. ICA-AROMA)

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -260,6 +265,11 @@ def load_confounds(
           `~desc-smoothAROMAnonaggr_bold.nii.gz`.
         - "basic": use noise independent components only.
 
+    tedana : :obj:`str`, default="rejected"
```
htwangtw: After having a better understanding of the `tedana` outputs, I want to point out that the current option suggested implemented doesn't fit the logic of `load_confounds`. 
I would use this [official tutorial of `tedana` on denoising](https://tedana.readthedocs.io/en/stable/denoising.html#remove-all-noise-correlated-fluctuations-aggressive-denoising) to implement a few options applicable to combined image `~desc-optcom_bold.nii.gz`:

- `aggressive`: return the rejected independent components. This is the current behaviour implemented :tada:
- `non_agressive`: return rejected and accepted independent components, and add prefix to the regressor column names with `reject` or `accept`. This will also create a warning to tell users that this should not be passed to masker objects directly. 
- `orthogonal_aggressive`: return orthogonalised noise components w.r.t. the accepted components prior to denoising.

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
```diff
@@ -39,7 +39,7 @@ def _simu_img(tmp_path, trend, demean):
     # demean set to False just for simulating signal based on the original
     # state
     confounds, _ = load_confounds(
-        file_nii, strategy=("motion",), motion="basic", demean=False
+        file_nii, strategy=("motion",), motion="basic", demean=demean
```
htwangtw: Please read the comment above why this was set to false. I have reverted this chance

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -455,6 +486,52 @@ def _load_single_confounds_file(
...
+            for col in mixing.columns
+        ]
+
+        # select matched columns
+        matched_components = [
+            c for c in rejected_components if c in mixing.columns
+        ]
+        confounds_all = mixing[sorted(matched_components)]
+
+        return prepare_output(confounds_all, demean)
```
htwangtw: This entire section should be moved to `load_confounds_components.py` as a function, and should be triggered by `_load_noise_component`.
I recommend reading how `_load_noise_component` gets confounds for fmriprep outputs to refactor this part of the code.

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -455,6 +486,52 @@ def _load_single_confounds_file(
...
+            )
+            .tolist()
+        )
+
+        # normalize column names in the mixing file
+        mixing = all_tedana_confounds["mixing"]
+        mixing.columns = [
+            re.sub(r"ICA_0*(\d+)$", lambda m: f"ICA_{int(m.group(1))}", col)
+            for col in mixing.columns
+        ]
```
htwangtw: I won't recommend changing any headers unless it's necessary. 

Also this change will create output that doesn't comply with [BIDS recommendation on index](https://bids-specification.readthedocs.io/en/stable/common-principles.html#participant-names-and-other-labels) - Indices should be zero padded.

htwangtw (changes requested):
I have added some tests based on the current behaviours. The most important part of the next practical step is to move the block of in `_load_single_confounds_file` to `load_confounds_components.py`. In the top of the file there is a description to show how this module is used by `_load_noise_component`.

I would highly recommend to read the code, say load motion regressor, and follow through the logic to implement it. 

Other than the code, now I have a better understanding of how tedana regressor is generated, I would like to discuss two things: API design and soft denoising support.

You can see my suggestion in the comment of documentation, but I am also copying this here:

> After having a better understanding of the `tedana` outputs, I want to point out that the current option suggested implemented doesn't fit the logic of `load_confounds`. 
> I would use this [official tutorial of `tedana` on denoising](https://tedana.readthedocs.io/en/stable/denoising.html#remove-all-noise-correlated-fluctuations-aggressive-denoising) to implement a few options applicable to combined image `~desc-optcom_bold.nii.gz`:
> 
> - `aggressive`: return the rejected independent components. This is the current behaviour implemented :tada:
> - `non_agressive`: return rejected and accepted independent components, and add prefix to the regressor column names with `reject` or `accept`. This will also create a warning to tell users that this should not be passed to masker objects directly. 
> - `orthogonal_aggressive`: return orthogonalised noise components w.r.t. the accepted components prior to denoising.

We can discuss this on Wednesday. This will also address #5325. If this is implemented we can potentially kill two birds with one stone.

The other topic I want to discuss is soft denoising, this is also relevant to #5325. My current idea of the outputs aims to make soft denoising easier to implement on user end. Combined with `_load_single_confounds_file`, we can refactor the tedana tutorial on denoisng with `nilearn`. Currently nilearn doesn't have direct implementation of soft denoising. I can think of reason for and against adding this other class of signal cleaning approach. I can write up something and move the discussion there

man-shu:
@handwerkerd, one of the maintainers on tedana, also offered to help/review in case it's needed.

@Remi-Gau maybe we can add him as one of the reviewers?

Remi-Gau:
> @Remi-Gau maybe we can add him as one of the reviewers?

yes of course

I think anyone can review, right?

bthirion:
Bzw, let us know when a review is needed. Best,

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -260,6 +264,20 @@ def load_confounds(
...
+
+        - "aggressive": use :term:`Tedana` tedana optimaly combined
+            image output.
+            `~desc-optcom_bold.nii.gz`.
+        - "non-aggressive": use :term:`Tedana` tedana optimaly combined
+            image output.
+            `~desc-optcom_bold.nii.gz`.
+        - "orthogonal": use :term:`Tedana` tedana optimaly combined
+            image output.
+            `~desc-optcom_bold.nii.gz`.
```
htwangtw: We might want to add some minimal documentation for each

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -24,9 +26,104 @@
...
+        a valid choice of strategy.
+    """
+    all_tedana_confounds = {}
+
+    if tedana not in ["aggressive", "non-aggressive", "orthogonal"]:
+        raise ValueError(
+            "Please select an option when using TEDANA strategy. "
+            f"Current input: {tedana}"
+        )
+    if tedana == "aggressive":
```
htwangtw, line 129: other strategies?

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -24,9 +26,104 @@
...
+        for tedana_conf in ["mixing", "metrics"]:
```
htwangtw: Is it possible to move this to `load_confounds_file_as_dataframe`?

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -455,6 +492,17 @@ def _load_single_confounds_file(
...
+    flag_tedana = ("tedana" in strategy) and (kwargs.get("tedana"))
+
+    missing = {"confounds": [], "keywords": []}
+    # tedana will load multiple files so we need to handle if
+    # before loading "regular" confounds from other sources
+    if flag_tedana:
+        confounds_all, missing = _load_noise_component(
+            confounds_file, "tedana", missing, **kwargs
+        )
+        return prepare_output(confounds_all, demean)
```
htwangtw: can we have this block integrated with line 518?

htwangtw (changes requested):
@miltoncamacho Thank you so much! I only have two comment with the structure - I wonder if it's possible to move the file loading out of `_load_tedana` to `load_confounds_file_as_dataframe` and have all the _load_noise_component in one place so the workflow in `_load_single_confounds_file` remains clean and consistent?

bthirion:
The documentation failure seems unrelated. Is this ready for a round of reviews ?

miltoncamacho:
Yes, it is ready for review

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -260,6 +264,21 @@ def load_confounds(
...
         - "basic": use noise independent components only.
 
+    tedana : :obj:`str`, default="aggressive"
+
+        - "aggressive": use :term:`Tedana` tedana optimaly combined
+            image output.
+            `~desc-optcom_bold.nii.gz`.
+            If aggressive is selected all the components classified as
+            rejected by tedana will be returned as a pandas DataFrame
+        - "non-aggressive": use :term:`Tedana` tedana optimaly combined
```
bthirion:
```suggestion
        - "non-aggressive": use :term:`Tedana` tedana optimally combined
```

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -260,6 +264,21 @@ def load_confounds(
           `~desc-smoothAROMAnonaggr_bold.nii.gz`.
         - "basic": use noise independent components only.
 
+    tedana : :obj:`str`, default="aggressive"
+
+        - "aggressive": use :term:`Tedana` tedana optimaly combined
```
bthirion:
```suggestion
        - "aggressive": use :term:`Tedana` tedana optimally combined
```

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -27,6 +29,102 @@
...
+    # 1. Get selected components from status table
+    selected_components = mixing[
+        metrics[metrics["classification"].isin(classification)]["Component"]
+    ]
+
+    #############################################################
+    # tedana versions like 24.0.2 do not have matching component
+    # names from the mixing and status tables. For backwards
+    # compatibility we will normalize the component names to
+    # match the mixing file normalize rejected component
```
bthirion:
```suggestion
    # match the mixing file  rejected component
``` (sorry if I misunderstood)

**Review by bthirion — COMMENTED** [review]
Just found a couple of typos, LGTM otherwise

File: nilearn/interfaces/fmriprep/load_confounds.py (hunk 31 above)
Comment (Remi-Gau) [inline nilearn/interfaces/fmriprep/load_confounds.py]: if you use :term:`Tedana` then you must add it to the glossary,
otherwise you get a failure in the doc build

/home/runner/work/nilearn/nilearn/.tox/doc/lib/python3.9/site-packages/nilearn/interfaces/fmriprep/load_confounds.py:docstring of nilearn.interfaces.fmriprep.load_confounds.load_confounds:147: WARNING: term not in glossary: 'Tedana'

https://github.com/nilearn/nilearn/actions/runs/16122467668/job/45504792675?pr=5410#step:13:1539

It should be added here: https://github.com/nilearn/nilearn/blob/347557ec24fe17ac9060ab3c28a48eff597ce2a6/doc/glossary.rst?plain=1#L302

File: nilearn/interfaces/fmriprep/load_confounds_components.py [hunk 33]
```diff
@@ -27,6 +29,102 @@
...
+            f"Current input: {tedana}"
+        )
+
+    if tedana == "aggressive":
+        return _tedana_strategy(
+            ["rejected"],
+            mixing=confounds_files["mixing"],
+            metrics=confounds_files["metrics"],
+        )
+    elif tedana == "non-aggressive":
```
htwangtw, line 135: Currently it just return all the components.
I suggest to prepend `noise_`/`signal_` (or `rejected_`/`accepted_` ) to the independent components, so we can simplify [this tutorial](https://tedana.readthedocs.io/en/stable/denoising.html).
This will be the only case where we want to modify the column names for easier downstream application.

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
```diff
@@ -652,6 +654,35 @@ def test_ica_aroma(tmp_path, fmriprep_version):
...
+    # check that the tedana nifti file loads correctly
+    conf, _ = load_confounds(tedana_nii, strategy=("tedana",))
+    assert conf.size > 0
+
+    # check that combining tedana with other strategies raises an warning
+    with pytest.warns(UserWarning, match="TEDANA strategy"):
+        conf, _ = load_confounds(
+            tedana_nii, strategy=("tedana", "motion"), motion="basic"
+        )
+
```
htwangtw, line 713: Add some test for different options.

htwangtw (commented):
I have a few last small requests!

doc/glossary.rst
```diff
@@ -129,6 +129,15 @@ If you wish to add a missing term, please
         or resting-state :term:`fMRI`, graph theory measures, surface or
         volume-based statistics, etc.
 
+    Tedana
+        `Tedana`_ is a multi-echo :term:`fMRI` data denoising pipeline.
+        :term:`Tedana` originally came about as a part of the ME-ICA pipeline,
+        although it has since diverged. An important distinction is that while
+        the ME-ICA pipeline originally performed both pre-processing and
```
handwerkerd:
```suggestion
        although it has since diverged. An important distinction is that
        the ME-ICA pipeline was written for python v2.7,
        and was not designed in a way that others could easily
        examine and improve the underlying algorithms.
```

The original me-ica didn't perform preprocessing. The input was the pre-processed echoes concatenated into a single file along the z axis.

nilearn/interfaces/fmriprep/load_confounds_utils.py
handwerkerd: I'm still trying to go through the code to understand what's happening here, but, if the goal is to remove the tedana `rejected` regressors at a later step then `desc-denoised` is `desc-optcom` after the confounding regressors are removed. Maybe you want to use `desc-optom` if you are subsequently regressing out the same time series.

nilearn/interfaces/fmriprep/load_confounds.py
```diff
@@ -260,6 +264,21 @@ def load_confounds(
...
+        - "aggressive": use :term:`Tedana` tedana optimally combined
+            image output.
+            `~desc-optcom_bold.nii.gz`.
+            If aggressive is selected all the components classified as
+            rejected by tedana will be returned as a pandas DataFrame
+        - "non-aggressive": use :term:`Tedana` tedana optimally combined
+            image output.
+            `~desc-optcom_bold.nii.gz`.
+            If non-aggresive is selected all the components classified as
+            either rejected or accepted will be returned as a pandas DataFrame
```
handwerkerd, line 278: I do not love this "aggressive" and "non-aggressive" terminology because we don't use those terms in `tedana` and, based on this docstring, I'm not sure what you mean.
To me it sounds like `aggressive` means you're regressing out the tedana-rejected time series as confounds. "non-aggressive" is including everything, but does that mean it's regressing out nothing and effectively passlng along the optimally combined volume. If so, that should probably be labeled something like "optimal-combination" or "no-ICA-denoising".

Also, a core design method of tedana is that there are multiple decision trees that can be used to defined what's rejected and some can be more agressive than others depending on use cases. Calling all of them "agressive" might add confusion.

To make this more complex, we also have a `--tedort` option which orthogonalizes the rejected component time series with respect to the accepted time series. The goal of this is to be conservative and retain any signal that is shared between accepted and rejected components. One could all this "non-agressive" or "conservative" but it's probably better to replicate the `tedort` term if you want to be consistent with tedana.

nilearn/interfaces/fmriprep/load_confounds_utils.py
handwerkerd, line 22: I noticed #5386 has comments expressing concerns for file naming stability. tedana has a file-name registry that lets users output files with (currently) 2 naming conventions. (Naming conventions are in : https://github.com/ME-ICA/tedana/blob/main/tedana/resources/config/outputs.json ).
You're using the BIDS naming convention here, which is fine if you assume everyone using nilearn with tedana will be using BIDS. If you want to make this more robust, `tedana` should output a file that ends `registry.json` and that file will contain the accurate file name value for the "combined img" key.

nilearn/interfaces/fmriprep/load_confounds.py
htwangtw: We had similar discussion while implementing ICA-AROMA. I certainly agree that non-aggressive vs aggressive are not the best term, and it would be great if you can provide some suggestions for alternatives, or @miltoncamacho do you have better ideas? Currently the terms are from the official tutorial [here](https://tedana.readthedocs.io/en/stable/denoising.html).

Just to make sure we are on the same page, I want to clarify that `load_confounds` only retrieve the confound regressors. I am here to provide review on how the code base is structured, but for naming I don't have the best understanding of tedana to provide ideas. There's The aim of this PR is to able to simplify the confounds loading in this tutorial. An example code snippet to replace the data loading in the first section of the denoise tutorial is as followed:

```python
from nilearn.interfaces.fmriprep.load_confounds import load_confounds, _load_single_confounds_file

# files from fMRIPrep
mask_file = "path/to/fmriprep/sub-01/sub-01_task-rest_desc-brain_mask.nii.gz"
fmriprep_confounds_file = "path/to/fmriprep/sub-01/sub-01_task-rest_desc-confounds_timeseries.tsv"
opt_combined_tendana = "path/to/tedana/sub-01/sub-01_task-rest_space-MNI152NLin2009cAsym_desc-optcom_bold.nii.gz"

# load the mixing matrix and confounds
tedana_mixing, _ = load_confounds(opt_combined_tendana, strategy=["tendana"], tedana="non_aggressive")
accepted_components = tedana_mixing[[col for col in tedana_mixing.columns if "accepted" in col]]
rejected_components = tedana_mixing[[col for col in tedana_mixing.columns if "rejected" in col]]
confounds, _ = _load_single_confounds_file(fmriprep_confounds_file, strategy=["motion", "wm_csf"])

# continue to follow the original tutorial
```

nilearn/interfaces/fmriprep/load_confounds_utils.py
htwangtw: Thanks for pointing this out! Currently we are only aiming to support BIDS-like conventions.

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -26,6 +28,114 @@
...
+
+    Returns
+    -------
+    pandas.DataFrame
+        DataFrame of TEDANA regressors.
+    """
+    if tedana not in ["aggressive", "non-aggressive"]:
+        raise ValueError(
+            "Please select an option when using TEDANA strategy. "
+            f"Current input: {tedana}"
```
htwangtw: This message can be improved by including the valid options

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -26,6 +28,114 @@
...
+    # names from the mixing and status tables. For backwards
+    # compatibility we will normalize the component names to
+    # match the mixing file rejected component
+    # names: ICA_04 -> ICA_4
+    #############################################################
+
+    # 3. Normalize mixing file column names and build a mapping
+    original_columns = mixing.columns.tolist()
+
+    # Map: normalized_name → original_name
```
htwangtw: This is a nick pick: can we avoid using special symbols that might not be rendered?

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -26,6 +28,114 @@
...
+    selected = mixing[sorted(matched_components)]
+
+    # 5. Rename columns using classification + original name
+    renamed_columns = {
+        col: f"{classification_lookup[col]}_{column_mapping[col]}"
+        for col in selected.columns
+    }
+    load_confounds = selected.rename(columns=renamed_columns)
+
+    return load_confounds
```
htwangtw: naming a variable with the same name of an existing function is confusing. 
```suggestion
    return selected.rename(columns=renamed_columns)
```

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -26,6 +28,114 @@
...
+    matched_components = [
+        c for c in classification_lookup if c in mixing.columns
+    ]
+    selected = mixing[sorted(matched_components)]
+
+    # 5. Rename columns using classification + original name
+    renamed_columns = {
+        col: f"{classification_lookup[col]}_{column_mapping[col]}"
+        for col in selected.columns
+    }
```
htwangtw: I suggest to make a note here that we are prefixing the column names with the classification regardless of the strategy selected. I personally found this confusing when only the rejected components are loaded, but have no strong opinion on this.

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
```diff
@@ -652,6 +654,51 @@ def test_ica_aroma(tmp_path, fmriprep_version):
...
+    # check that combining tedana with other strategies raises an warning
+    with pytest.warns(UserWarning, match="TEDANA strategy"):
+        conf, _ = load_confounds(
+            tedana_nii,
+            strategy=("tedana", "high_pass", "ica_aroma", "global_signal"),
+            motion="basic",
+        )
+
```
htwangtw, line 721: Add new tests for the different options here.

doc/glossary.rst
```diff
@@ -129,6 +129,16 @@ If you wish to add a missing term, please
...
+    Tedana
+        `Tedana`_ is a multi-echo :term:`fMRI` data denoising pipeline.
+        :term:`Tedana` originally came about as a part of the ME-ICA pipeline,
+        although it has since diverged. An important distinction is that
+        the ME-ICA pipeline was written for python v2.7,
+        and was not designed in a way that others could easily
+        examine and improve the underlying algorithms for
+        TE-dependent analysis of multi-echo :term:`fMRI` data, tedana now assumes
+        that you are working with data which has been previously preprocessed.
+
```
bthirion, line 148: I think that the description should say a few words on how Tedana works: my understanding is that it is based on a (spatial ?) ICA decomposition, but what makes it specific to ME is not obvious to me.

bthirion (commented):
As I'm not an expert of this method? I plainly as for a bit more pedagogy.

Also, I'm wondering whether we should demo that in examples or tutorials (for a future PR in any case).
Thx,

doc/glossary.rst
emdupre: A spatial ICA, yes ! It's performed on the "optimally combined" (weighted echo) time series, but the decision-tree to reject or include individual ICA components depends on the individually-acquired echos, so the whole workflow is ME specific.
I don't know that this need to be explained in depth here, but tedana has its own documentation (tedana.readthedocs.io) and there's a [course in development](https://me-ica.github.io/multi-echo-data-analysis/content/intro.html) to explain ME data and processing more deeply that would be worth linking out to. Maybe tedana could be added to the glossary, so we could have one place to expand on this information ?

nilearn/interfaces/fmriprep/load_confounds.py
handwerkerd: Sorry for the delayed response. I checked with @tsalo and others I realized that part of the confusion is the "aggressive" "non-aggressive" terminology in the documentation is about the various ways to regress the time series from rejected components when there is some collinearity with the time series of accepted components. You can remove all information in the rejected time series without regard to the accepted time series (aggressive) or fit the full model and only remove part of the shared fluctuations (non-aggressive). The confusion is that this is a general explanation and tedana doesn't actually implement the aggressive approach within our code. As such, the proposed terminology here is fine.

bthirion (commented):
I'm rather happy with the PR as is.
Disclaimer: I'm not a tedana/ME-ICA user.
I'm a bit concerned that this is not showcased in any example ?

htwangtw:
@bthirion - indeed. @miltoncamacho has pointed to some example datasets open neuro in the discussion; perhaps that's a good direction we can add to the PR?

miltoncamacho:
> @bthirion - indeed. @miltoncamacho has pointed to some example datasets open neuro in the discussion; perhaps that's a good direction we can add to the PR?

Where should the example be added?

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -28,6 +30,116 @@
...
+
+    metrics : :obj:`pd.DataFrame`
+        pandas dataframe with the metrics file.
+
+    Returns
+    -------
+    pandas.DataFrame
+        DataFrame of TEDANA regressors.
+    """
+    # 1. Get selected components from status table
```
Remi-Gau:
```suggestion
    # Get selected components from status table
```

Let's avoid numbering because it leads to extra maintenance (e.g if one step is added then you need to change the number of all the following step)

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -28,6 +30,116 @@
...
+
+    #############################################################
+    # tedana versions like 24.0.2 do not have matching component
+    # names from the mixing and status tables. For backwards
+    # compatibility we will normalize the component names to
+    # match the mixing file rejected component
+    # names: ICA_04 -> ICA_4
+    #############################################################
+
+    # 3. Normalize mixing file column names and build a mapping
```
Remi-Gau:
```suggestion
    # Normalize mixing file column names and build a mapping
```

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -28,6 +30,116 @@
...
+    classification_lookup = {
+        re.sub(
+            r"ICA_0*(\d+)$",
+            lambda m: f"ICA_{int(m.group(1))}",
+            row["Component"],
+        ): row["classification"]
+        for _, row in selected_metric_components.iterrows()
+    }
+
+    # 4. Select matched columns (normalized)
```
Remi-Gau:
```suggestion
    # Select matched columns (normalized)
```

nilearn/interfaces/fmriprep/load_confounds_components.py
```diff
@@ -28,6 +30,116 @@
...
+    matched_components = [
+        c for c in classification_lookup if c in mixing.columns
+    ]
+    selected = mixing[sorted(matched_components)]
+
+    # 5. Rename columns using classification + original name
```
Remi-Gau:
```suggestion
    # Rename columns using classification + original name
```

nilearn/interfaces/fmriprep/tests/test_load_confounds.py
```diff
@@ -650,6 +652,71 @@ def test_ica_aroma(tmp_path, fmriprep_version):
         )
 
 
+def test_tedana(tmp_path):
```
Remi-Gau:
```suggestion
def test_tedana(tmp_path):
```

could we split this into 2 tests:
- one for the errors and warnings
- one for the ['happy path'](https://en.wikipedia.org/wiki/Happy_path)

htwangtw:
> Where should the example be added?


Sorry for the late reply to this question.
An addition to the [existing load confounds example](https://nilearn.github.io/dev/auto_examples/03_connectivity/plot_signal_extraction.html#sphx-glr-auto-examples-03-connectivity-plot-signal-extraction-py) would be a good place to add the new example. I understand that we will need to either add new example data or use data from open neuro. @Remi-Gau @man-shu can either of you let @miltoncamacho know how to incorporate new data?

bthirion:
The PR LGTM. Can we merge with upstream and then merge this PR ? Thx !

Remi-Gau (approved):
good with me

Remi-Gau:
will fix the remaining CI errors

Remi-Gau:
> The PR LGTM. Can we merge with upstream and then merge this PR ? Thx !

@htwangtw seems you are requesting changes on this, is that still the case?

bthirion (approved):

htwangtw (approved):
Thank y'all for your patience! LGTM 
@miltoncamacho Congrats on your first PR. I am sure we will find some wrinkles down the line.

Remi-Gau:
OK the failures are not specific to this PR so this can be merged

Remi-Gau:
Ideally we would want an example to show how to use this but I am not sure we have a dataset for this.
Will open a separate issue for this.

Remi-Gau:
thanks @miltoncamacho for all the work and patience !!!! ❤️

PR #5088, opened 2025-01-23
Fix negative plotting threshold

hndgzkn:
- Closes None but adresses #4340

First attempt was to align the thresholding behavior in image plotting with image thresholding introducing the `two_sided` parameter as in #5074 . However a closer look at the documentation revealed that the same behavior can be obtained using `vmin` and `vmax` parameters of the plotting functions. 

Changes proposed in this pull request:

- Raise error when a negative threshold value is specified.
- Update documentation related to threshold
- Fix bug that set incorrect `symmetic_cbar` value when only one of `vmin`, or `vmax` is not None.
- Minor refactoring

Issue #4296, opened 2024-02-28
[BUG] `avg_method` and `symmetric_cbar` flags using `plot_surf_stat_map` working properly with plotly engine


albertofpena:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Debian testing (trixie)

### Python version

- [ ] 3.12
- [X] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

0.10.3

### Expected behavior

When using the plotly engine, the colorbar should not be symmetric if symmetric_cbar=False. The avg_method should be user settable. This seems related to #3941 bug.

### Current behavior & error messages

This is what I got:

The flag symmetric_cbar=False is ignored, always plotting symmetric colorbars. On the other hand, when avg_method flag is inserted, I get the following error:

```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[7], line 3
      1 from nilearn import plotting
----> 3 fig = plotting.plot_surf_stat_map(
      4     surf, texture, hemi='left',
      5     title='Surface left hemisphere', colorbar=True,
      6     engine='plotly', symmetric_cbar=False, avg_method='mean',
      7     darkness=0.5, cmap='black_red'
      8 )
     10 # change plotly lighting
     11 fig.figure.data[0].lighting.ambient = 0.5

File ~/.local/lib/python3.11/site-packages/nilearn/plotting/surf_plotting.py:1151, in plot_surf_stat_map(surf_mesh, stat_map, bg_map, hemi, view, engine, threshold, alpha, vmin, vmax, cmap, colorbar, symmetric_cbar, cbar_tick_format, bg_on_data, darkness, title, title_font_size, output_file, axes, figure, **kwargs)
   1142 # Call get_colorbar_and_data_ranges to derive symmetric vmin, vmax
   1143 # And colorbar limits depending on symmetric_cbar settings
   1144 cbar_vmin, cbar_vmax, vmin, vmax = get_colorbar_and_data_ranges(
   1145     loaded_stat_map,
   1146     vmin=vmin,
   1147     vmax=vmax,
   1148     symmetric_cbar=symmetric_cbar,
   1149 )
-> 1151 display = plot_surf(
   1152     surf_mesh, surf_map=loaded_stat_map,
   1153     bg_map=bg_map, hemi=hemi,
[...]

PR #5413, opened 2025-06-03
[FIX] prevent negative threshold for surface plots

hndgzkn:
- Closes None but relates to https://github.com/nilearn/nilearn/issues/4296.

Preventing negative threshold for image plotting is handled with https://github.com/nilearn/nilearn/pull/5088, however it does not address the same problem with surface plotting.

- Raise error when a negative threshold value is specified.
- Update documentation related to threshold
- Set default threshold value for `plot_surf_roi` to `None`.


nilearn/plotting/surface/surf_plotting.py
```diff
@@ -921,7 +918,7 @@ def plot_surf_roi(
         When using matplotlib as engine,
         `avg_method` will default to ``"median"`` if ``None`` is passed.
 
-    threshold : a number or None, default=1e-14
+    threshold : :obj:`int` or :obj:`float`, None, default=1e-14
```
hndgzkn: Could we set default value for `threshold` as `None` for this function?

Remi-Gau: and have the old default as a fall back if None is passed?

Remi-Gau:
test failure is unrelated: happening on main and several other places

nilearn/plotting/surface/surf_plotting.py
hndgzkn: Not really, in other functions in the module, threshold defaults to None, and the image is not thresholded. 

For `plot_surf_roi`, docstring says:
```
        Threshold regions that are labeled 0.
        If you want to use 0 as a label, set threshold to None.
```

So does that mean by default for roi the regions labeled 0 should be thresholded?

Remi-Gau: ha yeah especially plot_roi: would be better to have similar API across those 2

so that sounds like a 'bug' on our side

2 things to check:
- if setting it None breaks some of the 'baseline' test that do image comparison
- similarly if examples that use it produce visually different figures
- what values so surface atlas use to denote unknown regions, medial wall or in general things to ignore

hndgzkn: >     * what values so surface atlas use to denote unknown regions, medial wall or in general things to ignore

This is not clear to me :)

bthirion (commented):
Codecov seems a bit crasy but LGTM so far.

hndgzkn:
> test failure is unrelated: happening on main and several other places

Ok, I'll wait for them to be fixed to compare baseline images mentioned above.

nilearn/plotting/surface/surf_plotting.py
Remi-Gau: checked here the values used by the only surface atlas we 'ship':

https://nilearn.github.io/dev/modules/generated_reports/masker_reports_examples.html#surface-maskers

no unknown regions label and no vertex with a 0 value

BIDS does not have a label for unknown regions but does recommend using 0 for 'background'

So I would set the value to None to be consistent and let users change it to what they want.

Remi-Gau:
> compare baseline images mentioned above

the tests pass, so it seems fine

nilearn/plotting/surface/surf_plotting.py
Remi-Gau: @hndgzkn 
Will let you change the threshold default

hndgzkn:
> > compare baseline images mentioned above
> 
> the tests pass, so it seems fine

I haven't changed the value of the function yet. I'll try.

nilearn/plotting/surface/surf_plotting.py
hndgzkn: I've changed the default value to None. If the tests pass, I think we need to change the docstring as well. Is the following OK?

```
        By default the regions that are labeled 0 are not thresholded.
        If you want to threshold them, set threshold to a very small number, ex. 1e-14 .
```

Remi-Gau: the tests pass: the docstring suggestion would be good IMO

hndgzkn: I've updated the docstring with a slight difference.

Remi-Gau:
> LGTM but how do we see that it solves #4296 ?

#4296 hightlighted several problems: one of them was dealing with negative threshold

Remi-Gau (approved):

Issue #5261, opened 2025-03-21
[BUG] tkinter error prevents from building the doc with matplotlib 3.3


Remi-Gau:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

For example one of the following:
- Linux Ubuntu 24.04

### Python version

- [ ] 3.13
- [ ] 3.12
- [ ] 3.11
- [ ] 3.10
- [x] 3.9

### nilearn version

dev

### Expected behavior

for the doc to build

### Current behavior & error messages

This is what I got:

```bash
.pkg: _optional_hooks> python /home/remi-gau/miniconda3/lib/python3.12/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: get_requires_for_build_sdist> python /home/remi-gau/miniconda3/lib/python3.12/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: build_sdist> python /home/remi-gau/miniconda3/lib/python3.12/site-packages/pyproject_api/_backend.py True hatchling.build
doc: install_package> /home/remi-gau/miniconda3/bin/uv pip install --reinstall --no-deps nilearn@/home/remi-gau/github/nilearn/nilearn/.tox/.tmp/package/16/nilearn-0.11.2.dev244+g56a489514.d20250321.tar.gz
doc: commands[0]> make -C doc clean
make: Entering directory '/home/remi-gau/github/nilearn/nilearn/doc'
rm -rf _build/*
rm -rf auto_examples/
rm -rf generated/*
rm -rf modules/generated/*
rm -rf ../examples/**/nilearn_cache
rm -rf ../**/nilearn_cache
rm -rf modules/description/
make: Leaving directory '/home/remi-gau/github/nilearn/nilearn/doc'
doc: commands[1]> python maint_tools/citation_cff_maint.py
Traceback (most recent call last):
  File "/home/remi-gau/github/nilearn/nilearn/maint_tools/citation_cff_maint.py", line 10, in <module>
    from utils import root_dir
  File "/home/remi-gau/github/nilearn/nilearn/maint_tools/utils.py", line 109, in <module>
    mod = importlib.import_module(f"nilearn.{subpackage}")
  File "/home/remi-gau/.local/share/uv/python/cpython-3.9.21-linux-x86_64-gnu/lib/python3.9/importlib/__init__.py", line 127, in import_module
[...]

PR #5418, opened 2025-06-04
[FIX] fix tests failures on main

Remi-Gau:
- Closes none

- add baseline figures that were missing
- bump up time out time for test from 60 to 75 seconds as we otherwise have a lot of noise on main for some flaky tests that fail just because they took a bit longer than usual


Remi-Gau:
@hndgzkn @man-shu 
- I cannot run matplotlib 3.3.0 with python 3.9 on my machine at the moment because of https://github.com/nilearn/nilearn/issues/5261
so I tried generating the figures with a later matplotlib that does not have this bug, but this means that the figures generated will fail with 3.3.0
Could one of you generate the figures and update this PR with them?

To update the figures, run `tox run -e pytest_mpl` while on python 3.9

Remi-Gau:
actually wait maybe I can get it to work

Remi-Gau:
OK let's see if adding just the missing figures is enough

Remi-Gau:
> And one question: Is `pyarrow` used somewhere?

ah sorry forgot to explain: this is just to silence a warning we get in tests that tells us that in the future, it will be required to have pyarrow installed to use pandas

so just added to silence that warning

Remi-Gau:
> I cannot judge the content.

to be fair I am being pragmatic on this one: a test that was supposed to compare some figures to their baseline never did so because of a missing decorator and the baseline figure were missing

why didn't this fail earlier: no freaking idea

with infinite time I would very much like to know

PR #5421, opened 2025-06-04
Fix issue 5414

### LINKED ISSUE #5414: [BUG] extra warning / errors during plotting

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

ubuntu 24.04

### Python version

- [x] 3.13
- [ ] 3.12
- [ ] 3.11
- [ ] 3.10
- [ ] 3.9

### nilearn version

dev

### Expected behavior

some warnings when plotting cannot be silenced

This is what I got:

```bash
/home/remi-gau/github/nilearn/nilearn/tmp/bar.py:12: FutureWarning: From release 0.13.0 onwards, this function will, by default, copy the header of the input image to the output. Currently, the header is reset to the default Nifti1Header. To suppress this warning and use the new behavior, set `copy_header=True`.
  stat_img = threshold_img(

/home/remi-gau/github/nilearn/nilearn/tmp/bar.py:29: UserWarning: hemi='both' was passed with type(surf_map)=<class 'numpy.ndarray'> and type(surf_mesh)=<class 'str'>.
This value will be ignored as it is only used when 'map' is a SurfaceImage instance and / or 'mesh' is a PolyMesh instance.
  plotting.plot_surf_stat_map(

/home/remi-gau/github/nilearn/nilearn/tmp/bar.py:29: UserWarning: 'symmetric_cmap' is not implemented for the matplotlib engine.
Got 'symmetric_cmap = True'.
Use 'symmetric_cmap = None' to silence this warning.
  plotting.plot_surf_stat_map(
```

The last warning is there eventhough `symmetric_cmap = None`

Trying to set hemi=None to remove the second warning gives an error:

```
  File "/home/remi-gau/github/nilearn/nilearn/tmp/bar.py", line 29, in <module>
    plotting.plot_surf_stat_map(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        fsaverage.infl_right,
        ^^^^^^^^^^^^^^^^^^^^^
    ...<8 lines>...
        hemi=None
        ^^^^^^^^^
    )
    ^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/surface/surf_plotting.py", line 626, in plot_surf_stat_map
    fig = backend._plot_surf(
        surf_mesh,
    ...<22 lines>...
        **kwargs,
    )
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/surface/_matplotlib_backend.py", line 446, in _plot_surf
    elev, azim = _get_view_plot_surf(hemi, view)
                 ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/surface/_matplotlib_backend.py", line 385, in _get_view_plot_surf
    view = sanitize_hemi_view(hemi, view)
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/surface/_utils.py", line 366, in sanitize_hemi_view
    check_hemispheres([hemi])
    ~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/surface/_utils.py", line 73, in check_hemispheres
    raise ValueError(
    ...<3 lines>...
    )
ValueError: Invalid hemispheres definition!
Got: [None]
Supported values are: ('left', 'right', 'both')
```

```python
import numpy as np

from nilearn.image import threshold_img
import matplotlib


stat_img = threshold_img(
    stat_img,
    threshold=2,
    cluster_threshold=0,
    two_sided=False,
    copy=True,
)

curv_right = surface.load_surf_data(fsaverage.curv_right)
curv_right_sign = np.sign(curv_right)

engine = "matplotlib"

plotting.plot_surf_stat_map(
    fsaverage.infl_right,
    texture,
    title="Surface right hemisphere",
    colorbar=True,
    threshold=1.0,
    bg_map=curv_right_sign,
    engine=engine,
    symmetric_cbar=None,
    darkness=None,
    hemi="both"
)
plt.show()
```

hndgzkn:
- Closes #5414 

-
-


nilearn/plotting/surface/_utils.py
```diff
@@ -276,9 +276,7 @@ def check_surface_plotting_inputs(
             message=(
                 f"{hemi=} was passed with "
                 f"{type(surf_map)=} and {type(surf_mesh)=}.\n"
-                "This value will be ignored as it is only used when "
-                "'map' is a SurfaceImage instance and / or "
-                "'mesh' is a PolyMesh instance."
+                "Please make sure that the specified hemi corresponds to data."
```
hndgzkn: For the moment, I've changed this to give warning on all values of `hemi`. I can change it depending on the final decision.

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -442,6 +442,7 @@ def plot_surf_stat_map(
     view=None,
     engine=DEFAULT_ENGINE,
     cmap=DEFAULT_DIVERGING_CMAP,
+    symmetric_cmap=None,
```
hndgzkn, line 414: This did not exist in the original code. However we let the users already specify `symmetric_cmap` by kwargs. Then below it calls `plot_surf` with `symmetric_cmap=True`, then it doubles the param. 

So I've decided to introduce this if it does not cause any problems.

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -623,30 +636,31 @@ def plot_surf_stat_map(
...
         surf_mesh,
         surf_map=loaded_stat_map,
         bg_map=bg_map,
         hemi=hemi,
         view=view,
-        avg_method=avg_method,
-        threshold=threshold,
+        engine=engine,
         cmap=cmap,
-        symmetric_cmap=True,
```
hndgzkn, line 596: `symmetric_cmap` is set to `True` here. However now `matplotlib` backend also raises warnings on unimplemented parameters, I've first changed it to `None`. Then decided to take it from `plot_surf_stat_map` params.

There is a difference in the behavior at the end. I do not know if that was a random choice to set it to `True` here, because `plotly` backend sets it to `False` by default (if the value is None) in the backend.

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -623,30 +636,31 @@ def plot_surf_stat_map(
...
+        threshold=threshold,
         alpha=alpha,
         bg_on_data=bg_on_data,
         darkness=darkness,
-        vmax=vmax,
         vmin=vmin,
+        vmax=vmax,
+        cbar_vmin=cbar_vmin,
+        cbar_vmax=cbar_vmax,
+        cbar_tick_format=cbar_tick_format,
```
hndgzkn, line 631: Here just reordering of params, to make it easy to compare with `plot_surf` signature.

nilearn/plotting/surface/_utils.py
bthirion: Sorry if I missed the point, but I don't think that we should *always* spit a warning. this clutters the run interface.

nilearn/plotting/surface/surf_plotting.py
bthirion: If everything renders well, this should be fine I think.

bthirion (commented):
LGTM overall, thx !

nilearn/plotting/surface/_utils.py
hndgzkn: The discussions are in the #5414 , specifically starting from [here](https://github.com/nilearn/nilearn/issues/5414#issuecomment-2939167758).

Briefly, it won't raise warning when SurfaceImage is specified to `plot_surf_stat_map` or `plot_surf_roi`. It will be the case, when single hemisphere data is specified, and we do not know which hemi value would be correct for the data.

If it is not convenient, as I said above, I can change depending on the decided behavior.

bthirion: OK, I start to understand.  My impression is that you're doing things right. But it also calls for some explanations in the documentation !

Remi-Gau:
> Sorry if I missed the point, but I don't think that we should always spit a warning. this clutters the run interface.

Maybe worth chatting about at the next dev meeting?

nilearn/plotting/surface/surf_plotting.py
Remi-Gau: hm...
do you think we should leave a comment in the code to explain this?
because this kind of detail will otherwise only exist in the PR discussion.

hndgzkn: I've put in docstring below that it is added with version 0.12.0, is it enough or I should put a comment in the code?

nilearn/plotting/surface/_utils.py
hndgzkn: I've updated docstrings in `surf_plotting` module. Let me know if that's ok.

nilearn/plotting/surface/surf_plotting.py
bthirion: Regarding this, how homogeneous is this signature across plotting functions ?

nilearn/plotting/surface/_utils.py
```diff
@@ -309,21 +309,6 @@ def check_surface_plotting_inputs(
...
-            message=(
-                f"{hemi=} was passed with "
-                f"{type(surf_map)=} and {type(surf_mesh)=}.\n"
-                "This value will be ignored as it is only used when "
-                "'map' is a SurfaceImage instance and / or "
-                "'mesh' is a PolyMesh instance."
-            ),
-            stacklevel=find_stack_level(),
-        )
-
```
hndgzkn, line 326: Removed the warning.

nilearn/plotting/surface/tests/test_surf_plotting.py
```diff
@@ -65,6 +65,25 @@ def test_check_surface_plotting_inputs_error_negative_threshold(
...
+@pytest.mark.parametrize("hemi", ["left", "right", "both"])
+def test_check_surface_plotting_inputs_single_hemi_data(
+    in_memory_mesh, fn, hemi
+):
+    """Smoke test when single hemi data is passed."""
+    parcellation = np.zeros((in_memory_mesh.n_vertices,))
+    parcellation[in_memory_mesh.faces[3]] = 1
+    fn(in_memory_mesh, parcellation, hemi=hemi)
+
+
```
hndgzkn, line 86: I've added a test for functions that apply (Is this also valid for `plot_img_on_surf`, as far as I understand no, but to be sure). Is using `parcellation` here correct?

nilearn/plotting/surface/tests/test_surf_plotting.py
```diff
@@ -395,14 +414,6 @@ def test_surface_plotting_axes_error(matplotlib_pyplot, surf_img_1d):
...
 
 
-def test_plot_surf_contours_warning_hemi(in_memory_mesh):
-    """Test warning that hemi will be ignored."""
-    parcellation = np.zeros((in_memory_mesh.n_vertices,))
-    parcellation[in_memory_mesh.faces[3]] = 1
-    with pytest.warns(UserWarning, match="This value will be ignored"):
-        plot_surf_contours(in_memory_mesh, parcellation, hemi="left")
-
-
```
hndgzkn, line 405: Replaced this test with the above test.

nilearn/plotting/surface/surf_plotting.py
hndgzkn: There is a similar case for `plot_surf_stat_map`. It does not have `cbar_vmin` and `cbar_vmax` in its signature. However inside it calls `plot_surf` and it uses calculated `cbar_vmin` and `cbar_vmax`. If they are also specified in `kwargs`, that will raise an error.

hndgzkn: `symmetric_cbar` only exists for `plot_surf_stat_map` and `plot_img_on_surf`.

hndgzkn: `plot_img_on_surf` also calls `plot_surf` in backend and sets `symmetric_cmap=True`. So if it is passed with `kwargs`, this will also raise error. But there are other parameters as well, that might cause problems. I think, instead of kwargs, we can explicitly set the parameters.

hndgzkn: Checking consistency of parameters might be done in another PR. As I look into that, there are more things to consider.

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -619,18 +645,18 @@ def plot_img_on_surf(
...
+    inflate=False,
     vmin=None,
     vmax=None,
-    threshold=None,
     symmetric_cbar="auto",
-    cmap=DEFAULT_DIVERGING_CMAP,
     cbar_tick_format="%i",
+    title=None,
+    output_file=None,
     **kwargs,
```
hndgzkn, line 660: Here again only reordering of params.

Remi-Gau (approved):
good with me

PR #5407, opened 2025-05-28
[ENH] Refactor plotting surface - use modules as backend

hndgzkn:
- Closes None, but relates to:
  - #5234 
  - #5253 
  - #5276
  - #5282 
  - #5322
  - #5349
  - #5389
  - #5391
  - #5401 

Changes proposed in this pull request:

-  Refactor `plotting.surface` to use `plotting.surface._matplotlib_backend` and `plotting.surface._plotly_backend` modules as backend instead of `MatplotlibSurfaceBackend` and `PlotlySurfaceBackend` classes. 
- Remove `plotting.surface._backend` and `plotting.surface.tests.test_backend` modules

**There is actually no code change.**

- Utility functions in `plotting.surface._backend` are moved to `plotting.surface._utils` 
- Tests for utility functions in `plotting.surface.tests.test_backend` are moved to `plotting.surface.tests.test_utils`
- Contents of plotting methods of `plotting.surface._backend.BaseSurfaceBackend` class (which are common parts for both backends) are moved inside the functions in `plotting.surface.surf_plotting`.
- The class methods in `plotting.surface._matplotlib_backend.MatplotlibSurfaceBackend` and  `plotting.surface._plotly_backend.PlotlySurfaceBackend` became module functions.

PR #5401, opened 2025-05-22
[ENH] Refactor `nilearn.plotting.surface.plot_img_on_surf` to separate backends

hndgzkn:
- Closes None, but relates to:
  - #5234 
  - #5253 
  - #5276
  - #5282 
  - #5322
  - #5349
  - #5389
  - #5391 

Changes proposed in this pull request:

-  Refactor `plotting.surface.surf_plotting.plot_img_on_surf` to separate matplotlib code to `_matplotlib_backend` and plotly code to `_plotly_backend` and common parts to `backend`
- Add/update missing docstrings for tests in `test_surf_plotting` for `plot_img_on_surf`

PR #5426, opened 2025-06-05
[ENH] Refactor `nilearn.plotting.surface.html_surface` to separate backend

hndgzkn:
  - #5234 
  - #5253 
  - #5276
  - #5282 
  - #5322
  - #5349
  - #5389
  - #5391
  - #5401 
  - #5407 

-  Refactor `plotting.surface.html_surface` to extract `get_vertexcolor` function that uses matplotlib and move it to `_matplotlib_backend`
- Move tests accordingly

This is the final PR that refactors `nilearn.plotting.surface` package to separate `matplotlib` and `plotly` code to backend modules.

### DIFF
--- a/nilearn/_utils/tests/test_common.py
+++ b/nilearn/_utils/tests/test_common.py
@@ -43,7 +43,7 @@ def test_number_public_functions():
     If this is intentional, then the number should be updated in the test.
     Otherwise it means that the public API of nilearn has changed by mistake.
     """
-    assert len({_[0] for _ in all_functions()}) == 262
+    assert len({_[0] for _ in all_functions()}) == 261
 
 
 @pytest.mark.skipif(
--- a/nilearn/plotting/surface/_utils.py
+++ b/nilearn/plotting/surface/_utils.py
@@ -6,7 +6,9 @@ from warnings import warn
 import numpy as np
 
 from nilearn._utils import fill_doc
+from nilearn._utils.helpers import is_matplotlib_installed, is_plotly_installed
 from nilearn._utils.logger import find_stack_level
+from nilearn.plotting._utils import DEFAULT_ENGINE
 from nilearn.surface import (
     PolyMesh,
     SurfaceImage,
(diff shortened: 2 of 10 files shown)


nilearn/plotting/surface/_utils.py
```diff
@@ -31,6 +33,45 @@
 VALID_HEMISPHERES = "left", "right", "both"
 
 
+def get_surface_backend(engine=DEFAULT_ENGINE):
+    """Instantiate and return the required backend engine.
+
```
hndgzkn, line 38: Moved this to `_utils` as now it is imported from `html_surface` as well.

nilearn/_utils/tests/test_common.py
```diff
@@ -43,7 +43,7 @@ def test_number_public_functions():
     If this is intentional, then the number should be updated in the test.
     Otherwise it means that the public API of nilearn has changed by mistake.
     """
-    assert len({_[0] for _ in all_functions()}) == 262
+    assert len({_[0] for _ in all_functions()}) == 261
```
hndgzkn, line 46: As `get_vertexcolor` is moved from `html_surface` to `_matplotlib_backend`, the number of public functions descreases. 

Actually it is not used elsewhere, just in `plotting.surface` package. Would that be ok?

bthirion: Then shouldn't we call it `_get_vertexcolor` ?

bthirion (commented):
LGTM overall. The doc generation failure seems unrelated.
Just wondering about the function renaming.

nilearn/_utils/tests/test_common.py
hndgzkn: It is in `plotting.surface._matplotlib_backend` package which is private. However it is imported from `plotting.surface.html_surface` and `plotting.surface._plotly_backend` modules. As far as I understand, we name such functions without `_`, non?

Remi-Gau: hmm...
so you use the entire module as backend
so you never anywhere do:

```
from nilearn.plotting.surface._matplotlib_backend import get_vertexcolor
```

correct?

In that case I am tempted to say the function is still private to the _matplotlib_backend module and so should have a starting `_`.

what do you think?

hndgzkn: On one side I agree with you, on the other side I couldn't be sure:

`get_vertexcolor`, `adjust_colorbar_and_data_ranges` and `adjust_plot_roi_params` are called as functions of the module from outside of module. However, for example, ` _colorbar_from_array` and other functions starting with `_` in `_matplotlib_backend` module are only called from inside it. 

How shall I name it in that case?

hndgzkn: Although it is not explicitly imported, it is still imported. 

We might think of it as class methods which are only used inside the class and used as public instance methods?

Remi-Gau: > We might think of it as class methods which are only used inside the class and used as public instance methods?

yes that's the closest I could think too

I'd say let's add a `_` and only remove it if the function itself is imported and not the whole module: it is arbitrary but we gotta draw the line somewhere

hndgzkn: Ok, I've changed it together with `adjust_colorbar_and_data_ranges` and `adjust_plot_roi_params`. They are all used the same way.

PR #5427, opened 2025-06-06
[MAINT] identify slow tests

Remi-Gau:
- Closes none

- just a test
- reduce global timeout for all tests
- then mark individual slow test

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -66,7 +66,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
             matrix:
                 description: [latest dependencies]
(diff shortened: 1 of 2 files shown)


.github/workflows/test_with_tox.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```

Remi-Gau:
> The failure seems unrelated.

actually they are: I want to use this PR to detect our slow tests - I am tired of seeing them fail in main, so I am for now giving them 'infinite' amount of time to finish: once this is done we can start working on speeding them up

https://github.com/nilearn/nilearn/actions/runs/15559801681/job/43809010138?pr=5427#step:6:23039

PR #5434, opened 2025-06-11
[FIX] Invalid assignment to self in method

DimitriPapadopoulos:
- No need to assign, `self` is modified in-place.

### DIFF
--- a/nilearn/maskers/_utils.py
+++ b/nilearn/maskers/_utils.py
@@ -35,8 +35,6 @@ def sanitize_cleaning_parameters(masker):
     else:
         masker.clean_args_ = masker.clean_args
 
-    return masker
-
 
 def _check_dims(imgs):
     # check dims of one image if given a list
--- a/nilearn/maskers/multi_nifti_masker.py
+++ b/nilearn/maskers/multi_nifti_masker.py
@@ -284,7 +284,7 @@ class MultiNiftiMasker(NiftiMasker):
             "hover over the displayed image."
         )
 
-        self = sanitize_cleaning_parameters(self)
+        sanitize_cleaning_parameters(self)
 
         self.mask_img_ = self._load_mask(imgs)
 
--- a/nilearn/maskers/nifti_labels_masker.py
+++ b/nilearn/maskers/nifti_labels_masker.py
@@ -516,7 +516,7 @@ class NiftiLabelsMasker(BaseMasker):
                 f"parameter: {self.resampling_target}"
             )
 
-        self = sanitize_cleaning_parameters(self)
+        sanitize_cleaning_parameters(self)
 
         self._report_content = {
             "description": (
--- a/nilearn/maskers/nifti_maps_masker.py
+++ b/nilearn/maskers/nifti_maps_masker.py
@@ -404,7 +404,7 @@ class NiftiMapsMasker(BaseMasker):
                 "Set resampling_target to something else or provide a mask."
             )
 
-        self = sanitize_cleaning_parameters(self)
+        sanitize_cleaning_parameters(self)
 
         self._report_content = {
             "description": (
--- a/nilearn/maskers/nifti_masker.py
+++ b/nilearn/maskers/nifti_masker.py
@@ -444,7 +444,7 @@ class NiftiMasker(BaseMasker):
         if getattr(self, "_shelving", None) is None:
             self._shelving = False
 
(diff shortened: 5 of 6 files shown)


Remi-Gau:
hmm not sure about this: having the modification made explicit seems better to me

DimitriPapadopoulos:
Assigning to `self` or `cls` is bad style and should be avoided.

Remi-Gau:
> Assigning to `self` or `cls` is bad style and should be avoided.

can you explain a bit more so I don't just remember the rule but remember the reason for the rule?

Remi-Gau:
> > Assigning to `self` or `cls` is bad style and should be avoided.
> 
> can you explain a bit more so I don't just remember the rule but remember the reason for the rule?

from ruff ?

https://docs.astral.sh/ruff/rules/self-or-cls-assignment/

Given that sanitize_cleaning_parameters will be simplified (probably even removed) after 0.13 to only act on masker.clean_args_ I would suggest that we can live it till then.

DimitriPapadopoulos:
From [Is it safe to replace a self object by another object of the same type in a method?](https://stackoverflow.com/questions/1216356/is-it-safe-to-replace-a-self-object-by-another-object-of-the-same-type-in-a-meth):
> It is unlikely that replacing the `self` variable will accomplish whatever you're trying to do, that couldn't just be accomplished by storing the result of `func(self)` in a different variable. `self` is effectively a local variable only defined for the duration of the method call, used to pass in the instance of the class which is being operated upon. Replacing self will not actually replace references to the original instance of the class held by other objects, nor will it create a lasting reference to the new instance which was assigned to it.

So assigning to `self` is deceiving: it "explicitly" pretends that the class instance referred to by `self` is modified by the assignment - but it is not. What does modify the class instance is the `sanitize_cleaning_parameters()` call.

DimitriPapadopoulos:
See also [Why is `self` in Python objects immutable?](https://stackoverflow.com/questions/1015592/why-is-self-in-python-objects-immutable#1015602)

Remi-Gau:
ok then I would suggest inlining sanitize_cleaning_parameters so we can get rid of it

DimitriPapadopoulos:
What do you mean by inlining the function? Function [nilearn/maskers/_utils.sanitize_cleaning_parameters](https://github.com/nilearn/nilearn/blob/4e33f641a62c1634f7540fd100f8ebe8c3889dd7/nilearn/maskers/_utils.py#L7-L38) is quite complex.

Remi-Gau:
> What do you mean by inlining the function? Function [nilearn/maskers/_utils.sanitize_cleaning_parameters](https://github.com/nilearn/nilearn/blob/4e33f641a62c1634f7540fd100f8ebe8c3889dd7/nilearn/maskers/_utils.py#L7-L38) is quite complex.

https://refactoring.com/catalog/inlineFunction.html


instead of 

```python
self = sanitize_cleaning_parameters(self)
```

let's do:

```python
# TODO remove this if block when bumping to version >0.13
if hasattr(self, "clean_kwargs"):
    if self.clean_kwargs:
        tmp = [", ".join(list(self.clean_kwargs))]
        warnings.warn(
            f"You passed some kwargs to {self.__class__.__name__}: "
            f"{tmp}. "
            "This behavior is deprecated "
            "and will be removed in version 0.13.2.",
            DeprecationWarning,
            stacklevel=find_stack_level(),
        )
        if self.clean_args:
            raise ValueError(
                "Passing arguments via 'kwargs' "
                "is mutually exclusive with using 'clean_args'"
            )
    self.clean_kwargs_ = {
        k[7:]: v
        for k, v in masker.clean_kwargs.items()
        if k.startswith("clean__")
    }

if self.clean_args is None:
    masker.clean_args_ = {}
else:
    self.clean_args_ = masker.clean_args
```

DimitriPapadopoulos:
But the function is complex and used in 5 different files. I don't think it's a good idea to duplicate code.

I don't see a problem with a function that modifies a class instance in-place, like the [list.sort](https://docs.python.org/3/library/stdtypes.html#list.sort) method. The behaviour can be documented if needed.

Remi-Gau:
maybe let's turn that into a private method of the parent classes then and have 

self.sanitize_cleaning_parameters() ?

Remi-Gau:
will make a suggestion and you let me know what you think

PR #5437, opened 2025-06-12
[FIX] warn when some attributes passed at first level model init will be ignored at fit time

### LINKED ISSUE #5408: Misleading first level parameters logic
The current handling of the `slice_time_ref` parameter in `nilearn.glm.first_level.FirstLevelModel` can be misleading.

Although `slice_time_ref` is defined at the class initialization level, it is only used within the `fit` method **if** `design_matrices` is **not** provided. This creates a confusing situation where:

* Users may expect `slice_time_ref` to always be taken into account once set during instantiation.
* In practice, the parameter has **no effect** when precomputed design matrices are passed to `.fit()`.

I think it would be clearer if:

- the `slice_time_ref` parameter would move to the `fit` method signature (since it's only relevant there).
- Use `slice_time_ref` even when `design_matrices` are provided (e.g., by applying the same adjustment logic).

or 
-  `slice_time_ref` is an optional argument in `make_first_level_design_matrix`.
 - a **warning** in the documentation and/or during the execution of `.fit()` are displayed to inform users that `slice_time_ref` is ignored when `design_matrices` is manually supplied.

I understand that changing the function signature or behavior may introduce backward compatibility issues. However, given the potential for silent misuse, I believe it's worth addressing (even if through warnings ) to help users avoid incorrect assumptions in their analyses.

This issue is not linked to a specific platform nor a specific python/nilearn version.

Remi-Gau:
- Closes #5408 

- throw warning when some attributes passed at first level model init will be ignored at fit time

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -498,8 +498,15 @@ class FirstLevelModel(BaseGLM):
                 non_default_value.append("confounds")
             if events is not None:
                 non_default_value.append("events")
+            if non_default_value:
+                warn(
+                    "If design matrices are supplied, "
+                    f"{' and '.join(non_default_value)} will be ignored.",
+                    stacklevel=find_stack_level(),
+                )
 
             # check with the default of __init__
+            non_default_value = []
             attributes_used_in_des_mat_generation = [
                 "slice_time_ref",
                 "t_r",
@@ -521,9 +528,10 @@ class FirstLevelModel(BaseGLM):
             if non_default_value:
                 warn(
                     "If design matrices are supplied, "
-                    "confounds and events will be ignored.",
+                    f"[{', '.join(non_default_value)}] will be ignored.",
                     stacklevel=find_stack_level(),
                 )
+
             design_matrices = _check_run_tables(
                 run_imgs, design_matrices, "design_matrices"
             )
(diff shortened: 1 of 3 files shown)


nilearn/glm/first_level/first_level.py
```diff
@@ -483,42 +524,75 @@ def _check_fit_inputs(
...
         """Run input validation and ensure inputs are compatible."""
-        # Raise a warning if both design_matrices and confounds are provided
-        if design_matrices is not None and (
-            confounds is not None or events is not None
-        ):
-            warn(
-                "If design matrices are supplied, "
-                "confounds and events will be ignored.",
-                stacklevel=find_stack_level(),
+        if not isinstance(run_imgs, (list, tuple)):
```
hndgzkn, line 548: Would it be good to also check that `run_imgs != None`

nilearn/glm/first_level/first_level.py
```diff
@@ -483,42 +524,75 @@ def _check_fit_inputs(
...
+            run_imgs = [run_imgs]
+
+        if design_matrices is not None:
+            # If design_matrices is provided,
+            # throw warning for the attributes or parameters
+            # that were provided at init or fit time
+            # but that will be ignored
+            # because they will not be used to generate a design matrix.
+            non_default_value = []
```
hndgzkn: This array might be named differently, I think it creates confusion.

hndgzkn (approved):
Seems Ok to me. Just minor comments.

nilearn/glm/first_level/first_level.py
Remi-Gau: there is already a check for that around line 927

nilearn/glm/first_level/first_level.py
Remi-Gau: yeah indeed
I just gave them some more explicit names

nilearn/glm/first_level/first_level.py
hndgzkn: May be that check there might as well be moved to here. It is also an input check. What do you think?

hndgzkn: I mean this:

```

        if not isinstance(
            run_imgs, (str, Path, Nifti1Image, SurfaceImage, list, tuple)
        ) or (
            isinstance(run_imgs, (list, tuple))
            and not all(
                isinstance(x, (*NiimgLike, SurfaceImage)) for x in run_imgs
            )
        ):
            input_type = type(run_imgs)
            if isinstance(run_imgs, list):
                input_type = [type(x) for x in run_imgs]
            raise TypeError(
                "'run_imgs' must be a single instance / a list "
                "of any of the following:\n"
                "- string\n"
                "- pathlib.Path\n"
                "- NiftiImage\n"
                "- SurfaceImage\n"
                f"Got: {input_type}"
            )
```

Remi-Gau: indeed !!

done

PR #5438, opened 2025-06-12
[MAINT] Unassigned string statement

DimitriPapadopoulos:
- Remove duplicate docstring.
- Put docstring at top of function or module.
- Some docstrings are actual code comments.


nilearn/glm/tests/test_first_level.py
```diff
@@ -1,3 +1,5 @@
+"""Test the first level model on BIDS datasets."""
```
Remi-Gau:
```suggestion
"""Test related to first level model."""
```

not just bids 😉

PR #5309, opened 2025-04-02
[ENH] save surface GLM output to disk

Remi-Gau:
- Closes #5255

Changes proposed in this pull request:

- save surface GLM to disk
- ensure that figure saved to disk are referenced using full path when using mode.generate_report
-

PR #5448, opened 2025-06-16
[ENH] have GLM report "no supra-threshold found" in GLM report when relevant

### LINKED ISSUE #5269: GLM reports result figure when no voxel survives thresholding
I think it'd be less ambiguous if GLM reports had some sort of clear warning when no voxel survive thresholding:
at the moment users always to unfold the "cluster table" part to get confirmation of this:

Image

Remi-Gau:
- Closes #5269

- display "no supra-threshold found" in GLM report when relevant
- also add a message to explain the lack of cluster tables for surface GLM


pyproject.toml
```diff
@@ -167,7 +167,7 @@ module = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-ra --strict-config --strict-markers --doctest-modules --showlocals -s -vv --template=maint_tools/templates/index.html --timeout=45"
+addopts = "-ra --strict-config --strict-markers --doctest-modules --showlocals -s -v --template=maint_tools/templates/index.html --timeout=45"
```
Remi-Gau:
```suggestion
addopts = "-ra --strict-config --strict-markers --doctest-modules --showlocals -s -vv --template=maint_tools/templates/index.html --timeout=45"
```

Remi-Gau:
Before:

image

After:

image

Remi-Gau:
Before:

image


After:

image

nilearn/reporting/glm_reporter.py
```diff
@@ -640,32 +640,47 @@ def _make_stat_maps_contrast_clusters(
...
-
+            # FIXME
+            # The commented code below was there to reuse
+            # cluster tables generated by save_glm_to_bids
+            # to save time.
+            # However cluster tables may have been computed
+            # with different threshold, cluster_threshol...
+            # by save_glm_to_bids than those requested in
+            # generate_report.
+            # So we are skipping this for now.
```
Remi-Gau, line 651: Commenting code out for now to fix the issue spotted in this comment:

https://github.com/nilearn/nilearn/pull/5309#issuecomment-2776399837

(empty results table when figure clearly shows some activation)

The options I see for this are:
1. when generating reports we do not try to reuse results tables that were saved to disk previously by save_glm_to_bids and regenerate them from scratch (slower)
2.or we keep track of the parameters used to generate the tables in the filename where the results are saved (faster but lead to more complex names like ageEffectOnGM_contrast-age_control-fdr_thresh-05_k-10_alpha-1_clusters.tsv...)
3. check from the TSVs metadata, what parameters were used to generate the results table by save_glm_to_bids and only recompute it if they do not match the one requested when calling generate_repprt

For now I would say, let's take the slower option (1) and let's revisit maybe after the release.

examples/05_glm_second_level/plot_oasis.py
```diff
@@ -127,6 +127,8 @@
 )
 
 # %%
+# View results
```
Remi-Gau, line 130: just to trigger the build of this example to make sure the bug mentioned here (https://github.com/nilearn/nilearn/pull/5448/files#r2150256028) is fixed

nilearn/reporting/glm_reporter.py
bthirion: If the issue is to avoid recomputing when called twice with the same parameters, the solution is simply to use joblib's caching ?

bthirion (commented):
This PR calls for another one... Is this something you plan to do in a near future ?
Before the next release ?

Remi-Gau:
> Is this something you plan to do in a near future ?

yes

> Before the next release ?

no

I don't think it is that urgent

nilearn/reporting/glm_reporter.py
Remi-Gau: I will look into that but I'd rather do it after the next release

PR #5454, opened 2025-06-16
[MAINT] part revert of check empty image

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/5293#issuecomment-2977088126

- do not check nifti image to see if they are empty
- adapt home made estimator checks to only perform such checks with surface data
- should re-establish decent performance

### DIFF
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -1015,7 +1015,8 @@ def check_decoder_empty_data_messages(estimator):
         # because fit should fail early
         _, y = make_classification(
             n_samples=20,
-            n_features=5**3,
+            dim=5
+            n_features=dim**3,
             scale=3.0,
             n_informative=5,
             n_classes=2,


.github/workflows/test_with_tox.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
```
Remi-Gau: revert before merging

maint_tools/check_gha_workflow.py
```diff
@@ -61,17 +61,17 @@
...
 
 # Set to True if yu want to also include the runs of a CI workflow
 # that did not complete successfully.
-INCLUDE_FAILED_RUNS = False
+INCLUDE_FAILED_RUNS = True
 
 # Pages of runs to collect
 # 100 per page
-PAGES_TO_COLLECT = range(1, 30)
+PAGES_TO_COLLECT = range(1, 3)
```
Remi-Gau:
```suggestion
PAGES_TO_COLLECT = range(1, 30)
```

nilearn/_utils/niimg_conversions.py
```diff
@@ -352,8 +352,8 @@ def check_niimg(
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
 
-    if safe_get_data(niimg).size == 0:
-        raise ValueError("The image is empty.")
+    # if safe_get_data(niimg).size == 0:
+    #     raise ValueError("The image is empty.")
```
Remi-Gau:
```suggestion
```

Remi-Gau:
Not included in this PR, should be done in a follow up PR: adding a 'regression banchmark' that should detect this type of slowing down of performance.

Remi-Gau:
The run time of the test suite is now more acceptable.

image

.github/workflows/test_with_tox.yml
```diff
@@ -66,7 +66,7 @@ jobs:
...
+            fail-fast: false
```
```suggestion
            fail-fast: true
```

nilearn/_utils/estimator_checks.py
```diff
@@ -999,29 +1001,33 @@ def check_decoder_empty_data_messages(estimator):
...
+        # SearchLight do not support surface data directly
+        return None
 
     else:
         # we can use classification data even for regressors
         # because fit should fail early
         _, y = make_classification(
             n_samples=20,
-            n_features=dim**3,
+            n_features=5**3,
```
hndgzkn: Leaving `dim=5` might be good for clarity.

nilearn/_utils/niimg_conversions.py
```diff
@@ -352,9 +352,6 @@ def check_niimg(
     # Otherwise, it should be a filename or a SpatialImage, we load it
     niimg = load_niimg(niimg, dtype=dtype)
 
-    if safe_get_data(niimg).size == 0:
-        raise ValueError("The image is empty.")
-
```
hndgzkn, line 357: I was wondering if instead of putting this check in `check_niimg`, creating a function `is_img_empty` and doing the check only where it is necessary would help with performance? Because `check_niimg` is called everywhere. If I am not mistaken, we have to check this only when `fit` is called, where it already needs to get image data.

Remi-Gau: yeah that's a good idea.

Let's put a pin in it and get back to it after the release

nilearn/_utils/estimator_checks.py
```suggestion
            dim=5
            n_features=dim**3,
```

PR #5456, opened 2025-06-17
REL 0.12.0

Remi-Gau:
- Closes none

Changes proposed in this pull request:

- setting up things for the next release

PR #5457, opened 2025-06-17
[ENH] Refactor `nilearn.plotting.matrix`, partially reverting previous refactoring

hndgzkn:
- Closes None

- Remove `matplotlib_backend`
- Separate utility functions not dependent on `matplotlib` to `_utils` module
- Reorganize tests accordingly


Remi-Gau (approved):
good with me

nilearn/plotting/matrix/_utils.py
```diff
@@ -15,7 +23,95 @@ def _mask_matrix(mat, tri):
     return np.ma.masked_array(mat, mask)
 
 
-def _reorder_matrix(mat, labels, reorder):
+def pad_contrast_matrix(contrast_def, design_matrix):
```
hndgzkn, line 26: The test fail because I moved `pad_contrast_matrix` from `matrix_plotting` to `_utils`. So it is no more public. (I tried to move all functions that do not require `matplotlib` to _utils.)

Is there an inconvenience about it? It is not used elsewhere. If that is ok with you, I'll update the number of public functions.

Remi-Gau: yeah it should have been private already (it was only used in the module where it was defined), so go ahead.

Remi-Gau:
we forgot to clean the changelog that was changed in #5305

Remi-Gau:
fixing in  #5456

PR #5459, opened 2025-06-17
[DOC] measure and reduce number of warnings in doc build

Remi-Gau:
- Closes none

- list warnings thrown during doc build
- start silencing some warnings

### DIFF
--- a/examples/02_decoding/plot_haxby_glm_decoding.py
+++ b/examples/02_decoding/plot_haxby_glm_decoding.py
@@ -92,6 +92,7 @@ glm = FirstLevelModel(
     high_pass=0.008,
     smoothing_fwhm=4,
     memory="nilearn_cache",
+    memory_level=1,
 )
 
 # %%


bthirion (commented):
LGTM. Where can we see the output in terms of number of warnings ,

Remi-Gau:
They are uploaded as artefacts that can be accessed in the build_doc action (for PR or on main)

https://github.com/nilearn/nilearn/actions/runs/15734998690/job/44350685692?pr=5459#step:16:34

Will try to see if I can still remove a couple of warnings from our examples.

Remi-Gau:
ok let's stop there for now and then let's try to see how many warning we still have on main

nilearn/plotting/surface/_matplotlib_backend.py
```diff
@@ -745,6 +745,7 @@ def _plot_img_on_surf(
         symmetric_cbar = "auto"
     if cbar_tick_format is None:
         cbar_tick_format = "%i"
+    symmetric_cmap = kwargs.pop("symmetric_cmap", True)
```
hndgzkn, line 748: I'm a bit late to review this one, but this should set `None` instead of `True` because `symmetric_cmap` in not used in `plot_surf` for `matplotlib` backend.

Remi-Gau: so it was an error that is was set to True below (line 812-813)?

hndgzkn: It was set to `True` in original code, as `plotly` and `matplotlib` code was mixed, and it is used only in `plotly`, not implemented for `matplotlib`. 

`matplotlib` did not give warning for parameters not implemented for that. I had changed that so that `matplotlib` also raises warning. I should have changed this too, Apparently I've missed that.

hndgzkn: I can open a PR to fix that. 

Actually I had added `symmetric_cmap` to signature of `plot_surf_stat_map`. I think your solution is much nicer, and we can remove it. I can put that also in the PR so that it is fixed before the release.

Remi-Gau: > I can open a PR to fix that.

yup go for it

hndgzkn: I've rechecked the code: 

In the orginal code, `symmetric_cmap` is passed from `plot_img_on_surf` and `plot_surf_stat_map` to `plot_surf` as `symmetic_cmap=True` (If it was also set in `kwargs`, it could raise error).  It seems like, for `plotly` it was preferred to have `symmetric_cmap=True` for these two functions. 

However in `plot_surf` in `plotly`  engine, if `symmetric_cmap=None`, it is set to `False`. Is there a special reason for that? We could set it to `True` there. Then we won't need do any specific action to get `symmetric_cmap` from kwargs, it will be passed as it is.

Remi-Gau: > Is there a special reason for that? We could set it to `True` there.

Not that I know

> Then we won't need do any specific action to get `symmetric_cmap` from kwargs, it will be passed as it is.

Then let's maybe do that: as long as the examples run fine and don't throw warnings

Remi-Gau: By the way, the default for symmetric_cmap changed in plot_surf


None for the dev version of the doc

https://nilearn.github.io/dev/modules/generated/nilearn.plotting.plot_surf.html#nilearn.plotting.plot_surf

False for the stable version of the doc

https://nilearn.github.io/stable/modules/generated/nilearn.plotting.plot_surf.html#nilearn.plotting.plot_surf

It is OK because there is no change in actual default behavior, but this should be mentioned in the doc string of the function with a 

```rst
    .. versionchanged:: 0.12.0
        Default value changed to None.
```

The doc string should also explain what this value will take when using plotly.

hndgzkn: Yes, but in behavior it does not change actually. Because for `plotly` it is set to `False` when `None` and for  `matplotlib` it is not used.

I can add this in the `..versionchanged::` part.

Is it ok to change it to `True` instead of `False` for `plotly` in the backend and document it?

Remi-Gau: > Is it ok to change it to `True` instead of `False` for `plotly` in the backend and document it?

that would change the behavior in this case, no?

hndgzkn: Yes, the tests fail. So I do not change it.

PR #5464, opened 2025-06-18
[MAINT] Apply new ruff rules

DimitriPapadopoulos:

- Apply more rules, probably introduced by _ruff_ 0.12.
- Ignore rule introduced by _ruff_ 0.12.

### DIFF
--- a/nilearn/interfaces/bids/glm.py
+++ b/nilearn/interfaces/bids/glm.py
@@ -301,7 +301,7 @@ def save_glm_to_bids(
             if output_type in ["metadata", "results"]:
                 continue
 
-            img = statistical_maps[contrast_name][output_type]
+            img = contrast_maps[output_type]
             filename = filenames["statistical_maps"][contrast_name][
                 output_type
             ]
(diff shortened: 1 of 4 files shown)


Remi-Gau (changes requested):
Can you run `pre-commit autoupdate` and commit the changes ?

This should update the ruff version used by pre-commit.

Remi-Gau:
Nice! ❤️ 

Can you take care of that error?

https://results.pre-commit.ci/run/github/1235740/1750319259.ynBq1SsVQ3S_geoJC5watA

DimitriPapadopoulos:
Strange, this doesn't come up when running ruff directly from the command line. I guess the configuration of ruff and pre-commit are inconsistent, but cannot identify where.

Remi-Gau:
> Strange, this doesn't come up when running ruff directly from the command line. I guess the configuration of ruff and pre-commit are inconsistent, but cannot identify where.

yeah I have seen some things like that once in a while

usually I only run ruff via pre-commit to try to avoid this issue

and I have given on trying to figure out the root cause because life is too short

Remi-Gau (approved):
let's wait for CI to finish but this is good with me

thanks!

PR #5465, opened 2025-06-19
[MAINT] Consistency between ruff and pre-commit

DimitriPapadopoulos:
Make sure the same set of files is processed by ruff, whether run directly from the command line or through pre-commit.

- Change the set of files processed by ruff when run from the command line, to match the set of files processed by pre-commit.


Remi-Gau (approved):
good with me

PR #5466, opened 2025-06-19
[MAINT] Apply ruff rule RUF012

DimitriPapadopoulos:
RUF012 Mutable class attributes should be annotated with `typing.ClassVar`

- Use tuples instead of lists to avoid the above ruff error.


Remi-Gau (approved):
thanks

PR #5470, opened 2025-06-20
[FIX] `nilearn.plotting.surface` function params

hndgzkn:
- Closes None, but related to discussions [here](https://github.com/nilearn/nilearn/pull/5459#discussion_r2157083647).

- Instead of passing `symmetric_cmap=True` to `plot_surf` inside `plot_img_on_surf` and `plot_surf_stat_map` explicitly, pass it implicitly through `kwargs` without changing its value, which is by default `None`.


Remi-Gau:
can you ask for the full build of the doc: I want to see what warnings we are getting in the examples

Remi-Gau:
> can you ask for the full build of the doc: I want to see what warnings we are getting in the examples

requested

nilearn/plotting/surface/surf_plotting.py
```diff
@@ -411,7 +414,6 @@ def plot_surf_stat_map(
     view=None,
     engine=DEFAULT_ENGINE,
     cmap=DEFAULT_DIVERGING_CMAP,
-    symmetric_cmap=None,
```
hndgzkn, line 414: @Remi-Gau Is there a specific reason to introduce `symmetric_cbar` in `plot_surf_stat_map` but not in `plot_surf`?

Remi-Gau: nope

Remi-Gau: but modifying that would mean applying this codeblock to plot_surf

```
    cbar_vmin, cbar_vmax, vmin, vmax = (
        backend._adjust_colorbar_and_data_ranges(
            loaded_stat_map,
            vmin=vmin,
            vmax=vmax,
        )
    )
```

hndgzkn: Yes. But I am not sure if it is necessary. 

My questions are: 

- Why is it necessary in one function but not the other?

- Are there cases where one wants the cbar to be symmetric, but not cbar color, or vice versa?

Remi-Gau: not sure
I'd be tempted to leave this one for after the release rather than as part of this PR.

Remi-Gau:
Merging
let's see how many warnings we get in the doc build on main.

PR #5473, opened 2025-06-21
draft add anterior and posterior views

### LINKED ISSUE #5081: [ENH] Add more views to plots in surface masker reports
Exactly what the title describes. As of now we only show views = ["lateral", "medial"] in surface maskers, for example in SurfacLabelsMasker:

https://github.com/nilearn/nilearn/blob/1dc6345f5b05e679617c5dabe05fcb5c4e234e74/nilearn/maskers/surface_labels_masker.py#L547

Image

Might be good to show more views because a smaller region, say, on the dorsal side might not be visible from just these two angles.

chlohamp:
- Closes #5081 

-add anterior and posterior views
-

### DIFF
--- a/nilearn/maskers/surface_labels_masker.py
+++ b/nilearn/maskers/surface_labels_masker.py
@@ -251,9 +251,7 @@ class SurfaceLabelsMasker(_BaseSurfaceMasker):
         return lut["index"].to_dict()
 
     @fill_doc
-    @rename_parameters(
-        replacement_params={"img": "imgs"}, end_version="0.13.2"
-    )
+    @rename_parameters(replacement_params={"img": "imgs"}, end_version="0.13.2")
     def fit(self, imgs=None, y=None):
         """Prepare signal extraction from regions.
 
@@ -299,16 +297,12 @@ class SurfaceLabelsMasker(_BaseSurfaceMasker):
             # apply mask to label image
             for k in self.labels_img_.data.parts:
                 mask = self.mask_img_.data.parts[k]
-                self.labels_img_.data.parts[k][np.logical_not(mask)] = (
-                    self.background_label
-                )
+                self.labels_img_.data.parts[k][
+                    np.logical_not(mask)
+                ] = self.background_label
 
-            labels_before_mask = {
-                int(x) for x in np.unique(get_data(self.labels_img))
-            }
-            labels_after_mask = {
-                int(x) for x in np.unique(get_data(self.labels_img_))
-            }
+            labels_before_mask = {int(x) for x in np.unique(get_data(self.labels_img))}
+            labels_after_mask = {int(x) for x in np.unique(get_data(self.labels_img_))}
             labels_diff = labels_before_mask - labels_after_mask
             if labels_diff:
                 warnings.warn(
@@ -337,9 +331,7 @@ class SurfaceLabelsMasker(_BaseSurfaceMasker):
                 index=self.labels_img_,
             )
         else:
-            lut = generate_atlas_look_up_table(
-                function=None, index=self.labels_img_
-            )
+            lut = generate_atlas_look_up_table(function=None, index=self.labels_img_)
 
         self.lut_ = sanitize_look_up_table(lut, atlas=self.labels_img_)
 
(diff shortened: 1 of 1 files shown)


Remi-Gau:
@chlohamp can I ask you to check if you can maybe make similar changes in SurfaceMapsMasker and in SurfaceMasker.

If that's possible then maybe we can have that code once in the _BaseSurfaceMasker and all the Surface maskers would thus benefit from it.

If you don't have the time, that's OK too.

nilearn/maskers/surface_labels_masker.py
```diff
@@ -572,28 +572,29 @@ def _create_figure_for_report(self):
             img = mean_img(img)
             vmin, vmax = img.data._get_min_max()
 
-        # TODO: possibly allow to generate a report with other views
-        views = ["lateral", "medial"]
-        hemispheres = ["left", "right"]
+        hem_view = {
```
bthirion:
```suggestion
        hemi_view = {
``` (nitpick, but I find this more readable)

File: nilearn/maskers/surface_labels_masker.py [hunk 2]
```diff
@@ -572,28 +572,29 @@ def _create_figure_for_report(self):
...
+            "left": ["lateral", "medial"],
+            "right": ["lateral", "medial"],
+            "both": ["anterior", "posterior"],
+        }
```
bthirion: Not sure I understand your aim here: you want to plot left, right AND both ?

bthirion (commented):
Thx for tackling that !
I think that it will important to thinks of good defaults + a good example.

nilearn/maskers/surface_labels_masker.py
chlohamp: Screenshot 2025-06-23 at 3 07 53 PM

For the anterior and posterior views, if I plotted the hemispheres separately, there would be 4 plots (left/right) instead of the 2 with both hemispheres in the combined plots right now. 

I didn't screenshot the separate plots of the "left" and "right" hemispheres, but the plots looked redundant.

chlohamp: https://output.circle-artifacts.com/output/job/82385ad6-821b-4a66-8bb2-e7ad5dfe5c4f/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html

man-shu:
> If that's possible then maybe we can have that code once in the _BaseSurfaceMasker and all the Surface maskers would thus benefit from it.

@chlohamp and I were just looking at it, and I think it would need more work as the implementations across surface maskers are quite different. For now, we have the new views in SurfaceMasker and SurfaceLabelsMasker.

Remi-Gau:
> I think it would need more work as the implementations across surface maskers are quite different. For now, we have the new views in SurfaceMasker and SurfaceLabelsMasker.

okidoki: we may have to refactor things before this can be done then.

Let's roll with what you added here

Remi-Gau:
FYI: given this is an enhancement I'd prefer to keep this one for the 0.13 release rather than 0.12.1, but it's not a hill I will die on.

nilearn/maskers/surface_labels_masker.py
Remi-Gau: not related to this PR, but if we could find a way to have a single coloar bar for all plots, it'd be nice

Remi-Gau:
test failure is unrelated

Remi-Gau:
Will take over this one to add the feature to SurfaceMapsMasker

nilearn/plotting/surface/_utils.py
```diff
@@ -317,7 +317,8 @@ def check_surface_plotting_inputs(
     if surf_mesh is None and not isinstance(surf_map, SurfaceImage):
         raise TypeError(
             f"If you want to pass {mesh_var_name}=None, "
-            f"then {map_var_name} must be a SurfaceImage instance."
+            f"then {map_var_name} must be a SurfaceImage instance. "
+            f"Got surf_map={surf_map.__class__.__name__}"
```
Remi-Gau, line 321: unrelated but added it to help debugging some small issue

PR #5481, opened 2025-06-24
[ENH] Refactor plotting separate image

hndgzkn:
- Closes None

- Move `nilearn.plotting.img_plotting` to `nilearn.plotting.image.img_plotting`
- Move tests related to `img_plotting` under `nilearn.plotting.image.tests`
- Move utility functions in `img_plotting` to `nilearn.plotting.image.utils`

> No code change, just moved functions across modules.


hndgzkn:
I started a PR for that, but it might require more attention then `surface` and `matrix`.

- I am not sure if `MNI152TEMPLATE` is in public API?
- For `plot_img_comparison` function in `nilearn.plotting.img_plotting`, although it is imported from `nilearn.plotting` everywhere, there is a deprecation cycle. However, we haven't done this for other functions in `surf_plotting` and `matrix_plotting`. 
Is this deprecation really necessary?

If we change the location of `img_plotting`, this will again require a deprecation.

https://github.com/nilearn/nilearn/blob/a01c1ee4a8b1449b80adae6567fc22914c12e51b/nilearn/plotting/img_plotting.py#L2259-L2297

Remi-Gau:
> * I am not sure if `MNI152TEMPLATE` is in public API?

not in the doc and not showing up in any __init__ so not part of the user facing public API

Remi-Gau:
> * Is this deprecation really necessary?

no

I think we had added it in case some users would import from a module and not a subpackage

```python
from nilearn.foo.bar import bla
```

instead of 

```python
from nilearn.foo import bla
```

I would say that this deprecation is not needed but then this move of function should be mentioned in the changelog (like we did for all the refactoring of the plotting subpackage)

hndgzkn:
> > * I am not sure if `MNI152TEMPLATE` is in public API?
> 
> not in the doc and not showing up in any **init** so not part of the user facing public API

Then moving it will not necessitate a deprecation cycle.

hndgzkn:
@Remi-Gau Doc failure is due to moving `nilearn.plotting.img_comparison.py` under `nilearn.plotting.image`. 

Looking at [API Ref](https://nilearn.github.io/stable/modules/generated/nilearn.plotting.img_comparison.plot_img_comparison.html#nilearn.plotting.img_comparison.plot_img_comparison), I think we should not make this move, or at least do it with a deprecation. How shall I proceed?

Remi-Gau:
> we should not make this move

yeah maybe best to not do it then: sorry I had not spotted that

hndgzkn:
> > we should not make this move
> 
> yeah maybe best to not do it then: sorry I had not spotted that

Me neither, sorry, I'll revert this part.

nilearn/plotting/image/img_plotting.py
```diff
@@ -30,29 +30,30 @@
...
     create_colormap_from_lut,
-    get_cbar_ticks,
     get_colorbar_and_data_ranges,
     save_figure_if_needed,
 )
 from nilearn.plotting.displays import get_projector, get_slicer
+from nilearn.plotting.image._utils import (
+    MNI152TEMPLATE,
+    get_cropped_cbar_ticks,
+    load_anat,
```
hndgzkn, line 55: I moved these to `nilearn.plotting.image._utils`

nilearn/plotting/image/_utils.py
```diff
@@ -0,0 +1,164 @@
...
+                [abs(new_tick_locs - threshold) for _ in new_tick_locs]
+            )
+            new_tick_locs[idx_closest] = (
+                -threshold if threshold > cbar_vmax else threshold
+            )
+    return new_tick_locs
+
+
+def load_anat(anat_img=MNI152TEMPLATE, dim="auto", black_bg="auto"):
+    """Load anatomy, for optional diming."""
```
hndgzkn: I've put `load_anat` in `_utils` but I export in `__init__.py` as it is used in other packages. So it can be used by importing from `nilearn.plotting.image`. Is there an incovenience?

nilearn/plotting/glass_brain_files/plot_align_svg.py
Remi-Gau, line 1: random thought: should we move this script into our maint_tools folder?

Remi-Gau: maybe along with the script in that folder?

Remi-Gau: also given that these scripts are not tested, we should think of a way to run them on a regular basis as sanity checks

nilearn/reporting/glm_reporter.py
```diff
@@ -69,7 +69,7 @@
         plot_surf_stat_map,
     )
     from nilearn.plotting.cm import _cmap_d as nilearn_cmaps
-    from nilearn.plotting.img_plotting import (  # type: ignore[assignment]
+    from nilearn.plotting.image.img_plotting import (  # type: ignore[assignment]
```
Remi-Gau:
```suggestion
    from nilearn.plotting.image._utils import (  # type: ignore[assignment]
```

if we import directly from where it is defined, this make it clear that MNI152TEMPLATE should be in public module, so maybe we should just have a `nilearn.plotting.utils` and put it there?

nilearn/plotting/image/_utils.py
Remi-Gau: not an inconvenience but a change from what is done in the rest of the code base where we don't expose private functions in a subpackage

Remi-Gau: so maybe better not put it in the init

hndgzkn: Then both for `MNI152TEMPLATE` and `load_anat` we have the same problem. If I change `image._utils` to `image.utils`, that would solve the problem for both. How does it sound to you?

nilearn/plotting/image/img_plotting.py
```diff
@@ -30,29 +30,30 @@
...
     check_threshold_not_negative,
     create_colormap_from_lut,
-    get_cbar_ticks,
     get_colorbar_and_data_ranges,
     save_figure_if_needed,
 )
 from nilearn.plotting.displays import get_projector, get_slicer
+from nilearn.plotting.image.utils import (
+    MNI152TEMPLATE,
+    get_cropped_cbar_ticks,
```
hndgzkn, line 54: `get_cropped_cbar_ticks` need not necessarily be public. However as it is a utility function, I wanted to take it out. 
Is it preferable to put it to another module `_utils.py` ?

nilearn/plotting/image/_utils.py
hndgzkn: I've done it as I proposed above. But if you prefer differently, I can change.

nilearn/reporting/glm_reporter.py
hndgzkn: I've put in `nilearn.plotting.image.utils` thinking that it is relevant for images. However I doubt. Do you think `nilearn.plotting.utils` would be more relevant?

nilearn/plotting/glass_brain_files/plot_align_svg.py
hndgzkn: I think it is a good idea. Shall I do it in this PR, or a follow up one?

hndgzkn:
I need to update change log as well.

nilearn/plotting/glass_brain_files/plot_align_svg.py
Remi-Gau: in a follow up

nilearn/reporting/glm_reporter.py
Remi-Gau: > Do you think `nilearn.plotting.utils` would be more relevant?

you are right that this mlay be better: a single plotting.utils.py to expose plotting stuff to other nilearn subpackages

nilearn/plotting/image/img_plotting.py
Remi-Gau: maybe in nilearn/plotting/_utils.py ?

nilearn/reporting/glm_reporter.py
Remi-Gau: actually I think that the other strategy that we used in this case would to put things in nilearn._utils.plotting: this has the advantage of making it clear that the function is private (to the user) just by seeing where it is declared

Remi-Gau: we have a bit of both of those things happening:
- 'private' function foo declared in nilearn.X.utils
- 'private' function foo declared in nilearn._utils.X

according to our own doc we should probably go with the latter:

https://nilearn.github.io/dev/development.html#guidelines-for-private-functions

Remi-Gau: by our own conventions, the content of the following should be moved to nilearn._utils
- nilearn/reporting/utils.py
- nilearn/surface/utils.py

hndgzkn: > by our own conventions, the content of the following should be moved to nilearn._utils
> 

@Remi-Gau I have thought a bit on that, I think keeping them as they are (changing the ones which have _ to without _) might be a better alternative.

Actually don't we define public API with what is listed in `__all__` section of each package? 

A function `x` in a module `y` (not starting with an underscore-neither function nor module), not listed in the `__all__` section of `__init__.py`, does not actually become part of public API. If I do `import * from y`, `x` won't be there, non?

The inconvenience for me for `nilearn/_utils.py` is that:

- For coherence, each function that is part of a package should be in that package and be imported from there.
- Putting many utility functions from multiple subpackages in a single top package causes a lot of cyclic imports. 
   
I would suggest that we keep ../utils.py file for each module, and any function that can be used across the packages be put there, but not add to `__all__`. How does this sound?

nilearn/plotting/image/img_plotting.py
hndgzkn: As it is related to image processing, I wanted to keep it under `plotting.image`.

Remi-Gau: fair enough

nilearn/plotting/image/utils.py
```diff
@@ -0,0 +1,166 @@
...
+            self.vmax = data.max()
+            self._shape = anat_img.shape
+
+    @property
+    def _data_cache(self):
+        self.load()
+        return self.data
+
+    @property
+    def _dataobj(self):
```
bthirion, line 55: We need to redefine all these properties ... ?

bthirion (commented):
Besides the nesting question, the PR looks good.
(Maybe the nesting has been discussed previously, sorry if I'm a bit off)

hndgzkn:
> IIUC, you're correctly an additional level in the tree of the repo for plotting/Image.tests. I'm not a big fan of deep nesting, I'd rather like to have a flat structure as long as it makes sense.

It is the same as we did for `nilearn.plotting.surface` and `nilearn.plotting.matrix`. This makes more coherent subpackages. From the user point of view, the imports are still from `nilearn.plotting`.

nilearn/plotting/image/utils.py
hndgzkn: The class has just changed its module, it is not redefined. May be I misunderstood your question.

nilearn/reporting/glm_reporter.py
bthirion: This sounds reasonable. We can discuss that tomorrow.

hndgzkn: How shall I proceed in that?

doc/changes/latest.rst
```diff
@@ -18,3 +18,5 @@ Enhancements
 
 Changes
 -------
+
+:bdg-dark:`Code` Move ``nilearn.plotting.img_plotting`` under ``nilearn.plotting.image`` (:gh:`5481` by `Hande Gözükan`_).
```
hndgzkn: @Remi-Gau I remember you telling me sometime ago that we might name the package as `nilearn.plotting.volume` rather then `nilearn.plotting.image` as surface images are also images. Shall I change that to volume?

bthirion: No strong opinion on that. This is a bit more consistent maybe. We should look a bit more systematically at the naming used so far.

nilearn/plotting/image/utils.py
```diff
@@ -0,0 +1,166 @@
+"""Functions that help image plotting."""
```
Remi-Gau:
```suggestion
"""
Plotting utilities
that can be used outside of the plotting subpackage.
"""
```

nilearn/reporting/glm_reporter.py
Remi-Gau: > I would suggest that we keep ../utils.py file for each module, and any function that can be used across the packages be put there, but not add to `__all__`. How does this sound?

ok let's do that but that will require moving some things out of nilearn._utils to be consistent

hndgzkn: Yes, I think so. But shall we do it in a later PR?

hndgzkn:
Codecov complaint is due to moving functions/modules. There is no added code.

doc/changes/latest.rst
Remi-Gau: yes that makes sense to me as long as we have a clear division between the 2

nilearn/reporting/glm_reporter.py
Remi-Gau: yes in a later PR: easier to follow

Remi-Gau:
```suggestion
    from nilearn.plotting.image.utils import (  # type: ignore[assignment]
```

MNI152TEMPLATE is defined in nilearn.plotting.image.utils

nilearn/plotting/image/utils.py
```diff
@@ -0,0 +1,169 @@
+"""
+Plotting utilities
+that can be used outside of the plotting subpackage.
```
Remi-Gau:
```suggestion
that can be used outside of the plotting.image subpackage.
```

PR #5484, opened 2025-06-24
[MAINT] add non required target tag

Remi-Gau:
- Closes none

- specify in estimator tags if Y is required in fit or fit_transform
- https://scikit-learn.org/stable/modules/generated/sklearn.utils.TargetTags.html#sklearn.utils.TargetTags


Remi-Gau:
Only now remembered that inheriting from BaseEstimator means that this is already False.

https://github.com/scikit-learn/scikit-learn/blob/da08f3d99194565caaa2b6757a3816eef258cd70/sklearn/base.py#L455

Closing because not needed

Issue #5443, opened 2025-06-13
sklearn estimators checks to implement for ReNA


Remi-Gau:
Relates to #4538 

These are the sklearn checks for HierarchicalKMeans that are currenctly failing and need to be either ignored or replaced by a nilearn equivalent.

See https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html

```python
from nilearn.connectome import ReNA
from nilearn._utils.estimator_checks import return_expected_failed_checks

fc = return_expected_failed_checks(ReNA())

for k, v in fc.items():
    if v == "TODO":
        print(f'- [ ] {k}')
```

- [ ] check_clustering
- [ ] check_dict_unchanged
- [ ] check_dtype_object
- [ ] check_dont_overwrite_parameters
- [ ] check_estimators_dtypes
- [ ] check_estimators_fit_returns_self
- [ ] check_estimators_nan_inf
- [ ] check_estimators_overwrite_params
- [ ] check_estimators_pickle
- [ ] check_f_contiguous_array_estimator
- [ ] check_fit_idempotent
- [ ] check_fit_check_is_fitted
- [ ] check_fit_score_takes_y
- [ ] check_fit2d_predict1d
- [ ] check_methods_sample_order_invariance
- [ ] check_methods_subset_invariance
- [ ] check_n_features_in
- [ ] check_n_features_in_after_fitting
- [ ] check_pipeline_consistency
- [ ] check_positive_only_tag_during_fit
- [ ] check_readonly_memmap_input
- [ ] check_transformer_data_not_an_array
- [ ] check_transformer_general
- [ ] check_transformer_preserve_dtypes

bthirion:
Same question here: Do you have an idea regarding which one should be addressed ?

PR #5485, opened 2025-06-24
[MAINT] check compliance ReNa with sklearn guidelines

Remi-Gau:
- Relates to #5443

- make slight modification to ReNa to make it easier to test with sklearn checks


nilearn/regions/rena_clustering.py
```diff
@@ -703,30 +705,50 @@ def fit(self, X, y=None):
...
+            self.n_features_in_ = X.shape[1]
+
+        # If no mask images was passed we create a dummy nifti image
+        # with a single slice
+        self.mask_img_ = self.mask_img
+        if self.mask_img_ is None:
+            self.mask_img_ = Nifti1Image(
+                np.ones((self.n_features_in_, 1, 1), dtype=np.int8),
+                affine=np.eye(4),
+            )
```
Remi-Gau: @bthirion 
I am adding this to make it easier to ensure the compliance of the estimator with sklearn rules.

However this means that the fit will run even if user do not provide a meaningful mask that was used to extract the data.
So it means the clustering of the data will be unrelated to the original 'topology' of the data.

I am not really happy with this.

At minimum, I think we should throw a warning.

Otherwise I can try to hack something, so this dummy image is only created during tests.

what do you think?

nilearn/regions/rena_clustering.py
```diff
@@ -703,30 +705,50 @@ def fit(self, X, y=None):
         """
         del y
         check_params(self.__dict__)
-        X = check_array(
-            X, ensure_min_features=2, ensure_min_samples=2, estimator=self
-        )
-        n_features = X.shape[1]
 
+        if not SKLEARN_LT_1_6:
```
Remi-Gau: invert logic

nilearn/regions/rena_clustering.py
```diff
@@ -794,6 +816,19 @@ def transform(
         """
         check_is_fitted(self)
 
+        # TODO simplify when dropping sklearn 1.5
+        if not SKLEARN_LT_1_6:
```
Remi-Gau: invert logic

nilearn/regions/rena_clustering.py
bthirion: The behavior so far was that not providing a mask would rise an error ?
I agree that we should through a warning (at least)

Remi-Gau: > The behavior so far was that not providing a mask would rise an error ?

yup

I will try to find a way to create this dummy files only during tests: am not a fan of potentially retunrning non sense to the user if they forget to read the warning

Remi-Gau: added a _set_mask_img_for_tests method

it's ugly because we have a slightly different behavior in tests but it does make it easier to test sklearn compliance.

hndgzkn (approved):
LGTM.

PR #5491, opened 2025-06-25
[MAINT] ensure decomposition estimator return self

Remi-Gau:
- Closes none

- perform check that was skipped for decomposition estimators

### DIFF
--- a/nilearn/decomposition/tests/conftest.py
+++ b/nilearn/decomposition/tests/conftest.py
@@ -115,9 +115,9 @@ def _decomposition_images_surface(
 def _decomposition_img(
     data_type,
     rng,
-    mesh,
-    shape,
-    affine,
+    mesh=None,
+    shape=None,
+    affine=None,
     with_activation: bool = True,
 ) -> Union[SurfaceImage, Nifti1Image]:
     """Return a single image for decomposition."""
--- a/nilearn/regions/hierarchical_kmeans_clustering.py
+++ b/nilearn/regions/hierarchical_kmeans_clustering.py
@@ -247,7 +247,7 @@ class HierarchicalKMeans(ClusterMixin, TransformerMixin, BaseEstimator):
         if SKLEARN_LT_1_6:
             from nilearn._utils.tags import tags
 
-            return tags()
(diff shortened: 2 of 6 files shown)


nilearn/decomposition/tests/conftest.py
```diff
@@ -115,9 +115,9 @@ def _decomposition_images_surface(
 def _decomposition_img(
     data_type,
     rng,
-    mesh,
-    shape,
-    affine,
+    mesh=None,
```
bthirion, line 118: Why do these arguments default to `None` ?

nilearn/regions/hierarchical_kmeans_clustering.py
```diff
@@ -247,7 +247,7 @@ def __sklearn_tags__(self):
         if SKLEARN_LT_1_6:
             from nilearn._utils.tags import tags
 
-            return tags()
+            return tags(niimg_like=False)
```
bthirion, line 250: Can you explain this change ?

bthirion (commented):
Thx, this seems to work but a bit of pedagogy could help.

nilearn/decomposition/tests/conftest.py
Remi-Gau: Mostly added those to be able to use it, as a function outside of the pytests fixture context.

When use in tests with fixtures mesh, shape and affine are always defined even if we only need to generate some surface or some volume data.

But here I actually needed to only generate some surface data, so it felt strange to also pass some affine and shape just to 'comply' with the function signature.

Setting those parameters to None makes this a bit easier.

nilearn/regions/hierarchical_kmeans_clustering.py
Remi-Gau: yup

This was something that had been forgotten: if you look how tags are defined for more recent sklearn version (line 255), you see that this estimator should not accept niimg like data as input.

But the same had not been done for older version of sklearn and this led to some strange tests failures when testing with them: I eventually tracked down the issue to this.

hndgzkn (approved):
LGTM.

PR #5492, opened 2025-06-25
[FIX] Align `symmetric_cmap` behavior in plotting package for `plotly` backend with `matplotlib` backend

### LINKED ISSUE #4296: [BUG] `avg_method` and `symmetric_cbar` flags using `plot_surf_stat_map` working properly with plotly engine

hndgzkn:
- Closes #4296 

- Do not change value of `symmetric_cmap`  inside `nilearn.plotting.js_plotting_utils.colorscale` function depending on the values of the specified data.
- Let user set `vmin` even if `symmetric_cmap` is specified to be `True`.
- Remove `symmetric_cmap` from the dictionary returned from `nilearn.plotting.js_plotting_utils.colorscale`.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -11,6 +11,8 @@ NEW
 Fixes
 -----
 
+- :bdg-dark:`Code` Align ``symmetric_cmap`` behavior for ``plotly`` backend in :func:`~nilearn.plotting.surface.surf_plotting.plot_surf` function with ``matplotlib`` backend (:gh:`5492` by `Hande Gözükan`_).
+
 Enhancements
 ------------
 
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -11,7 +11,7 @@ NEW
 Fixes
 -----
 
-- :bdg-dark:`Code` Align ``symmetric_cmap`` behavior for ``plotly`` backend in :func:`~nilearn.plotting.surface.surf_plotting.plot_surf` function with ``matplotlib`` backend (:gh:`5492` by `Hande Gözükan`_).
+- :bdg-dark:`Code` Align ``symmetric_cmap`` behavior for ``plotly`` backend in :func:`~nilearn.plotting.plot_surf` function with ``matplotlib`` backend (:gh:`5492` by `Hande Gözükan`_).
 
 Enhancements
 ------------
--- a/nilearn/plotting/js_plotting_utils.py
+++ b/nilearn/plotting/js_plotting_utils.py
@@ -77,11 +77,11 @@ def colorscale(
     if not symmetric_cmap and (values.min() < 0):
         warnings.warn(
             "you have specified symmetric_cmap=False "
-            "but the map contains negative values; "
-            "setting symmetric_cmap to True",
+            "but the map contains negative values; ",
+            #           "setting symmetric_cmap to True",
             stacklevel=find_stack_level(),
         )
-        symmetric_cmap = True
+    #        symmetric_cmap = True
     if symmetric_cmap and vmin is not None:
         warnings.warn(
             "vmin cannot be chosen when cmap is symmetric",
--- a/nilearn/plotting/js_plotting_utils.py
+++ b/nilearn/plotting/js_plotting_utils.py
@@ -7,9 +7,13 @@ import warnings
 from pathlib import Path
 from string import Template
 
-import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
+from matplotlib.colors import (
+    LinearSegmentedColormap,
+    ListedColormap,
+    Normalize,
+)
 
 from nilearn._utils.extmath import fast_abs_percentile
 from nilearn._utils.html_document import (  # noqa: F401
@@ -97,7 +101,7 @@ def colorscale(
     if symmetric_cmap:
         vmax = max(abs(vmin), abs(vmax))
         vmin = -vmax
-    norm = mpl.colors.Normalize(vmin=vmin, vmax=vmax)
+    norm = Normalize(vmin=vmin, vmax=vmax)
     cmaplist = [cmap(i) for i in range(cmap.N)]
     abs_threshold = None
     if threshold is not None:
@@ -106,7 +110,7 @@ def colorscale(
         istop = int(norm(abs_threshold, clip=True) * (cmap.N - 1))
         for i in range(istart, istop):
             cmaplist[i] = (0.5, 0.5, 0.5, 1.0)  # just an average gray color
-    our_cmap = mpl.colors.LinearSegmentedColormap.from_list(
+    our_cmap = LinearSegmentedColormap.from_list(
         "Custom cmap", cmaplist, cmap.N
     )
     x = np.linspace(0, 1, 100)
@@ -159,7 +163,7 @@ def mesh_to_plotly(mesh):
 
 def to_color_strings(colors):
     """Return a list of colors as hex strings."""
-    cmap = mpl.colors.ListedColormap(colors)
+    cmap = ListedColormap(colors)
     colors = cmap(np.arange(cmap.N))[:, :3]
     colors = np.asarray(colors * 255, dtype="uint8")
     colors = [
--- a/nilearn/plotting/js_plotting_utils.py
+++ b/nilearn/plotting/js_plotting_utils.py
@@ -74,14 +74,7 @@ def colorscale(
     """Normalize a cmap, put it in plotly format, get threshold and range."""
     cmap = plt.get_cmap(cmap)
     abs_values = np.abs(values)
-    if not symmetric_cmap and (values.min() < 0):
-        warnings.warn(
-            "you have specified symmetric_cmap=False "
-            "but the map contains negative values; ",
-            #           "setting symmetric_cmap to True",
-            stacklevel=find_stack_level(),
-        )
-    #        symmetric_cmap = True
+
     if symmetric_cmap and vmin is not None:
         warnings.warn(
             "vmin cannot be chosen when cmap is symmetric",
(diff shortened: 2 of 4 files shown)


nilearn/plotting/js_plotting_utils.py
```diff
@@ -77,11 +77,11 @@ def colorscale(
...
         warnings.warn(
             "you have specified symmetric_cmap=False "
-            "but the map contains negative values; "
-            "setting symmetric_cmap to True",
+            "but the map contains negative values; ",
+            #           "setting symmetric_cmap to True",
             stacklevel=find_stack_level(),
         )
-        symmetric_cmap = True
+    #        symmetric_cmap = True
```
hndgzkn: For the moment, I've only commented out the part that fixes the problem. 

I think it would be good to discuss how to proceed:

- Do we need to give a warning when the user sets "symmetric_cmap=False" about the negative values? Because it does not exclude negative values. Can I completely remove this check?

- As far as I understand `symmetric_cbar` and `symmetric_cmap` is used for the same purpose, to determine the bounds of cbar. (That might be good if you confirm that.) In that case can we unify the parameter name?

- `plotly` and `matplotlib` backends calculate data bounds and thresholds in different functions. I think those could be unified. How does it sound to you?

- These can be done in separate PRs. This PR would only fix the issue.

Remi-Gau: > * `plotly` and `matplotlib` backends calculate data bounds and thresholds in different functions. I think those could be unified. How does it sound to you?

yes let's do the refactor in a follow up PR

Remi-Gau: > * Do we need to give a warning when the user sets "symmetric_cmap=False" about the negative values?

no I don't think we should throw a warning

Remi-Gau: > * As far as I understand `symmetric_cbar` and `symmetric_cmap` is used for the same purpose, to determine the bounds of cbar. (That might be good if you confirm that.) In that case can we unify the parameter name?

that would involve a change in the API, so I would keep for later in case this is something we want to do

nilearn/plotting/js_plotting_utils.py
```diff
@@ -74,14 +74,7 @@ def colorscale(
...
-            "you have specified symmetric_cmap=False "
-            "but the map contains negative values; "
-            "setting symmetric_cmap to True",
-            stacklevel=find_stack_level(),
-        )
-        symmetric_cmap = True
+
     if symmetric_cmap and vmin is not None:
         warnings.warn(
             "vmin cannot be chosen when cmap is symmetric",
```
hndgzkn: Shall we treat this case differently? Can we remove that and instead  do as below?

```
if vmin is not None and vmax is not None and vmin != -vmax:
    vmax = max(abs(vmin), abs(vmax))
    vmin = -vmax
    warnings.warn(f"Specified vmin and vmax values do not create a symmetic colorbar. Modifying the values as {vmin=} {vmax=}")
```

hndgzkn:
The failure is independent of the changes, I couldn't yet figure out the reason.

Remi-Gau:
> > The failure is independent of the changes, I couldn't yet figure out the reason.
> 
> I think it is because `typing.Callable` is deprecated and it is imported in [nilearn.typing](https://github.com/nilearn/nilearn/blob/7c88fbcf7758c4239e57dafb6abb7e29338023c1/nilearn/typing.py#L24).
> 
> However fixing that is not sufficient.

na it's because once in a while we have some slow test that fail:

in  nilearn/mass_univariate/tests/test_permuted_least_squares.py

for those tests

- test_permuted_ols_check_h0_noeffect_labelswap_uncentered
- test_permuted_ols_check_h0_noeffect_signswap

add the decorator:

@pytest.mark.timeout(0)

this should help us on the long term identify the slow tests and ideally find ways to speed them up

hndgzkn:
> na it's because once in a while we have some slow test that fail:


Yes, I've just realized that :)

hndgzkn:
> this should help us on the long term identify the slow tests and ideally find ways to speed them up

Thanks, done!

nilearn/plotting/js_plotting_utils.py
bthirion: Agreed this makes more sense.

bthirion (commented):
Thx. I think we can go a bit further and refactor the procedure following your suggestion.

nilearn/plotting/js_plotting_utils.py
```diff
@@ -122,7 +127,6 @@ def colorscale(
         "cmap": our_cmap,
         "norm": norm,
         "abs_threshold": abs_threshold,
-        "symmetric_cmap": symmetric_cmap,
```
hndgzkn, line 125: I removed `symmetric_cmap` from the returned dictionary. It is no more modified inside the function `colorscale`. And even when it was modified, it was not used anywhere else in the code as far as I could see.

hndgzkn: Although it is public, I think it is not part of the public API.

bthirion (commented):
LGTM, thx !

Remi-Gau:
@hndgzkn we need an entry in the changelog as this changes behavior to fix a bug.

also we probably want more explicit name for the PR to have a cleaner git log after merging

hndgzkn:
> @hndgzkn we need an entry in the changelog as this changes behavior to fix a bug.
> 
> also we probably want more explicit name for the PR to have a cleaner git log after merging

@Remi-Gau Done!

bthirion (approved):
LGTM. We should rebuild the documentation to make sure it works.
Thx !

doc/changes/latest.rst
```diff
@@ -11,6 +11,8 @@ NEW
 Fixes
 -----
 
+- :bdg-dark:`Code` Align ``symmetric_cmap`` behavior for ``plotly`` backend in :func:`~nilearn.plotting.surface.surf_plotting.plot_surf` function with ``matplotlib`` backend (:gh:`5492` by `Hande Gözükan`_).
```
```suggestion
- :bdg-dark:`Code` Align ``symmetric_cmap`` behavior for ``plotly`` backend in :func:`~nilearn.plotting.plot_surf` function with ``matplotlib`` backend (:gh:`5492` by `Hande Gözükan`_).
```

pretty sure that the doc only uses the public API references hence why you are getting warnings during the doc build that make the doc build fail

Remi-Gau:
> We should rebuild the documentation to make sure it works.

not sure we have an example in the doc build that uses plotly that way.

PR #5496, opened 2025-06-25
[MAINT] homogenize logger calls in maskers

### LINKED ISSUE #5494: homogenize logger calls in maskers
not urgent but it'd be nice to have a single function that sends the same log messages across maskers for:
- loading mask
- loading data
- resampling mask / data / maps / labels
- ...

Remi-Gau:
- Closes #5494

- create a single function to send same messages in all maskers

### DIFF
--- a/nilearn/maskers/base_masker.py
+++ b/nilearn/maskers/base_masker.py
@@ -215,52 +215,29 @@ def mask_logger(step, img=None, verbose=0):
         repr = img.__repr__()
         if verbose > 1:
             repr = repr_niimgs(img, shorten=True)
-        if verbose > 2:
+        elif verbose > 2:
             repr = repr_niimgs(img, shorten=False)
 
-    if step == "load_mask":
-        if repr is not None:
-            logger.log(
-                msg=f"Loading mask from {repr}",
-                verbose=verbose,
-            )
+    messages = {
+        "cleaning": "Cleaning extracted signals",
+        "compute_mask": "Computing mask",
+        "extracting": "Extracting region signals",
+        "fit_done": "Finished fit",
+        "inverse_transform": "Computing image from signals",
+        "load_data": f"Loading data from {repr}",
+        "load_mask": f"Loading mask from {repr}",
+        "load_regions": f"Loading regions from {repr}",
+        "resample_mask": "Resamping mask",
+        "resample_regions": "Resampling regions",
+    }
 
-    elif step == "load_data":
-        if repr is not None:
-            logger.log(
-                msg=f"Loading data from {repr}",
-                verbose=verbose,
-            )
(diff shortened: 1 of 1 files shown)


nilearn/maskers/base_masker.py
```diff
@@ -211,6 +208,61 @@ def prepare_confounds_multimaskers(masker, imgs_list, confounds):
...
 
+def mask_logger(step, img=None, verbose=0):
+    """Log similar messages for all maskers."""
+    repr = None
+    if img is not None:
+        repr = img.__repr__()
+        if verbose > 1:
+            repr = repr_niimgs(img, shorten=True)
+        if verbose > 2:
+            repr = repr_niimgs(img, shorten=False)
```
hndgzkn:
```suggestion
        shorten = True
        if verbose > 2:
            shorten = False
        repr = repr_niimgs(img, shorten=False)
```

hndgzkn: In current code, if `verbose=3`, it will execute both if conditions. Is it the desired behavior?

nilearn/maskers/base_masker.py
```diff
@@ -211,6 +208,61 @@ def prepare_confounds_multimaskers(masker, imgs_list, confounds):
...
+        logger.log("Extracting region signals", verbose=verbose)
+
+    elif step == "cleaning":
+        logger.log("Cleaning extracted signals", verbose=verbose)
+
+    elif step == "fit_done":
+        logger.log("Finished fit", verbose=verbose)
+
+    elif step == "inverse_transform":
+        logger.log("Computing image from signals", verbose=verbose)
```
hndgzkn: May be, instead of many if conditions, a dictionary with `step` as key and message as value can be used. Or enumeration is also a possibility.

hndgzkn: It is also possible to create a MaskLogger class with a static method for each step.

nilearn/maskers/base_masker.py
Remi-Gau: no it's not!
thanks for catching this one

nilearn/maskers/base_masker.py
Remi-Gau: will go with the dict option

let's create a class only when we 'really' need it

nilearn/maskers/base_masker.py
Remi-Gau: actually it would not have been an issue because, the "repr" value would just be overwritten but clearer to avoid having both be run

nilearn/maskers/base_masker.py
Remi-Gau: looks cleaner now

PR #5498, opened 2025-06-25
[MAINT] silence warnings in test and doc build

Remi-Gau:
- Closes none

- reduce number of warnings in doc build and test suite

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -1065,10 +1065,12 @@ class FirstLevelModel(BaseGLM):
             warn(
                 (
                     f"The same contrast will be used for all {n_runs} runs. "
-                    "If you want to apply a different contrast to each run, "
-                    "pass contrast as an expression using "
+                    "If the design matrices are not the same for all runs, "
+                    "(for example with different column names "
+                    "or column order across runs) "
+                    "you should pass contrast as an expression using "
                     "the name of the conditions "
-                    "as they appear in the design matrix."
+                    "as they appear in the design matrices."
                 ),
                 category=RuntimeWarning,
                 stacklevel=find_stack_level(),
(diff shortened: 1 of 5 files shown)


nilearn/glm/first_level/first_level.py
```diff
@@ -1063,8 +1063,14 @@ def compute_contrast(
...
         if n_contrasts == 1 and n_runs > 1:
             warn(
-                f"One contrast given, assuming it for all {n_runs} runs",
-                category=UserWarning,
+                (
+                    f"The same contrast will be used for all {n_runs} runs. "
+                    "If you want to apply a different contrast to each run, "
+                    "pass contrast as an expression using "
+                    "the name of the conditions "
+                    "as they appear in the design matrix."
```
Remi-Gau: @bthirion 
does this warning message make sense to you?

trying to make it more explicit to help user know what to do if this is not what they want to do

bthirion: Not sure  I understand the second sentence. To me providing an expression addresses the issue that design matrices could have different number of columns or column names across runs. It does not address the case where on would like to define different contrasts.

Remi-Gau: ha yeah you are right: my phrasing does not make that clear.

maybe something like 

```python
f"The same contrast will be used for all {n_runs} runs. "
"If the design matrices are not the same for all runs, "
"(for example different column names "
"or column order across runs) "
"you should pass contrast as an expression using "
"the name of the conditions "
"as they appear in the design matrices."
```

Remi-Gau:
This removes about 90 warnings from the test suite:

Before:

image

After:

image

nilearn/glm/first_level/first_level.py
bthirion: Yes !

PR #5499, opened 2025-06-25
[MAINT] ensure img estimators can be pickled

Remi-Gau:
- Closes none

- check that all nilearn image estimators can be pickled

### DIFF
--- a/nilearn/decomposition/tests/test_decomposition_estimators.py
+++ b/nilearn/decomposition/tests/test_decomposition_estimators.py
@@ -102,6 +102,7 @@ def test_fit_errors(data_type, decomposition_images, estimator):
         )
 
 
+@pytest.mark.timeout(0)
 @pytest.mark.parametrize("estimator", [CanICA, _MultiPCA, DictLearning])
 @pytest.mark.parametrize("data_type", ["nifti", "surface"])
 def test_masker_attributes_with_fit(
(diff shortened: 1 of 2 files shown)


nilearn/_utils/estimator_checks.py
```diff
@@ -678,7 +675,7 @@ def fit_estimator(estimator: BaseEstimator) -> BaseEstimator:
         data[2, 2, 2, y] = 2
         X = Nifti1Image(data, np.eye(4))
 
-        return estimator.fit(X, y)
+        return X, y
```
hndgzkn, line 678: Instead of each of these return statements, single statement outside of if conditions could avoid repetitions.

hndgzkn (approved):
LGTM.

nilearn/_utils/estimator_checks.py
```diff
@@ -882,6 +902,59 @@ def check_img_estimators_overwrite_params(estimator) -> None:
...
+            result[method] = getattr(estimator, method)(input)
+            result["input"] = input
+
+    for method, input in zip(check_methods, input_data):
+        if method not in result:
+            continue
+        unpickled_result = getattr(unpickled_estimator, method)(input)
+        if isinstance(unpickled_result, np.ndarray):
+            if isinstance(estimator, SearchLight):
+                # TODO check why Searchlight has lower absolute tolerance
```
bthirion, line 946: To be checked before merging ?

bthirion (approved):
LGTM besides one detail.

nilearn/_utils/estimator_checks.py
Remi-Gau: Not sure.
Would prefer to circle back it in another PR.

nilearn/_utils/estimator_checks.py
Remi-Gau: yeah but sometimes in several cases the 'nature / type' of the second variable will be very different: in some cases it is an actually "y" required for fitting, sometimes is it is a design matrix...

so I'd rather not.

in fact when I try to do it the mypy starts complaining because of a given variable is suddenly getting an unexpected type.

PR #5500, opened 2025-06-25
[MAINT] Unnecessary comprehension

DimitriPapadopoulos:
The built-in function being used does not require comprehension and can work directly with a generator expression.

Using a generator expression within these functions is faster than using a comprehension:
- all
- any
- enumerate
- iter
- itertools.cycle
- itertools.accumulate

### DIFF
--- a/nilearn/connectome/group_sparse_cov.py
+++ b/nilearn/connectome/group_sparse_cov.py
@@ -670,8 +670,6 @@ class GroupSparseCovariance(CacheMixin, BaseEstimator):
                 ensure_min_samples=2,
             )
 
-        self.n_features_in_ = next(iter(s.shape[1] for s in subjects))
-
         if self.memory is None:
             self.memory = Memory(location=None)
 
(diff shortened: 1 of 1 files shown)


bthirion (approved):
LGTM, thx.

nilearn/connectome/group_sparse_cov.py
```diff
@@ -670,7 +670,7 @@ def fit(self, subjects, y=None):
                 ensure_min_samples=2,
             )
 
-        self.n_features_in_ = next(iter({s.shape[1] for s in subjects}))
+        self.n_features_in_ = next(iter(s.shape[1] for s in subjects))
```
Remi-Gau: Can I be annoying and ask you to also move the line after the empirical covariance is computed?
Reason: all subjects should have the same number of features but this is only checked by `empirical_covariances` so `self.n_features_in_` should be set after that.

Don't hesitate to add a comment to explain it.

I added this line a couple of days ago and did not realize that this bit would "read strange".

nilearn/connectome/group_sparse_cov.py
```diff
@@ -1192,7 +1192,7 @@ def fit(self, subjects, y=None):
...
```
Remi-Gau: same thing for this one

Remi-Gau (changes requested):
minor change more for internal consistency

I can take care of it if you don't have the time

Remi-Gau (approved):
Merci

PR #5502, opened 2025-06-27
[MAINT] ignore warnings for nilearn check estimators

Remi-Gau:
- Closes none

- ignore warnings when running nilearn estimator checks


Remi-Gau:
> My understanding is that you just want to reduce the number of warnings, and that warning issued during estimators check (not fully sure why they are issued) are of no interest. Is it right ?

Yup that's exactly the reason.

Am using the same decorator that sklearn uses for their estimator checks.

Most of the warnings are thrown because the test data may not be ideal to avoid some warnings (brain mask too small, some things may not converge properly...)

bthirion (approved):
LGTM, thx.

PR #5504, opened 2025-06-30
[MAINT] reduce warnings and verbosity in doc

Remi-Gau:
- Closes none

- reduce versbosity in examples
- fix some spurious warnings thrown by caching: added a regression test for all our estimators which required a few extra fixes (actually use CacheMixin in surface maskers ...)


examples/03_connectivity/plot_data_driven_parcellations.py
```diff
@@ -74,7 +74,7 @@
     standardize=False,
     smoothing_fwhm=2.0,
     memory="nilearn_cache",
-    memory_level=1,
```
Remi-Gau: without this we are getting some warnings, that the memory_level is 0 and will be set to 1: I would consider this a bit of a buggy behavior that should fixed

Remi-Gau:
```python
from nilearn import datasets

from nilearn.regions import Parcellations

dataset = datasets.fetch_development_fmri(n_subjects=1, verbose=0)

ward = Parcellations(
    method="ward",
    n_parcels=1000,
    standardize=False,
    smoothing_fwhm=2.0,
    memory="nilearn_cache",
    memory_level=1,
    verbose=0,
)
ward.fit(dataset.func)
```

```
/home/remi-gau/github/nilearn/nilearn/tmp/foo.py:16: UserWarning: memory_level is currently set to 0 but a Memory object has been provided. Setting memory_level to 1.
  ward.fit(dataset.func)
```

example of the "bug": passing memory_level=1 throws a warning that "memory_level is currently set to 0" ???!!!

Remi-Gau: will try to add some tests to avoid have this come back

bthirion (commented):
LGTM so far.

nilearn/connectome/group_sparse_cov.py
```diff
@@ -1013,7 +1013,7 @@ def __call__(  # noqa: D102
 
 
 @fill_doc
-class GroupSparseCovarianceCV(CacheMixin, BaseEstimator):
```
Remi-Gau, line 1016: cache was never used

bthirion (approved):
LGTM, thx

PR #5507, opened 2025-07-02
[MAINT] Unnecessary comprehension

DimitriPapadopoulos:
I may be missing something, but I don't see why a set was added here.


Remi-Gau (approved):
indeed

Remi-Gau:
failing test is unrelated

Issue #5447, opened 2025-06-13
sklearn estimators checks to implement for maskers


Remi-Gau:
Relates to #4538 

These are the sklearn checks for NiftiMasker that are currently failing and need to be either ignored or replaced by a nilearn equivalent.

See https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html

```python
from nilearn.maskers import NiftiMasker
from nilearn._utils.estimator_checks import return_expected_failed_checks

fc = return_expected_failed_checks(NiftiMasker())

for k, v in fc.items():
    if v == "TODO":
        print(f'- [ ] {k}')
```

- [x] check_estimators_empty_data_messages
- [x] check_estimators_pickle
- [ ] check_estimators_nan_inf
- [x] check_estimators_overwrite_params
- [x] check_fit_idempotent
- [x] check_methods_sample_order_invariance
- [ ] check_methods_subset_invariance
- [ ] check_positive_only_tag_during_fit
- [x] check_pipeline_consistency
- [ ] check_readonly_memmap_input

Issue #5446, opened 2025-06-13
sklearn estimators checks to implement for GLM estimators


Remi-Gau:
Relates to #4538 

These are the sklearn checks for GLM estimators that are currently failing and need to be either ignored or replaced by a nilearn equivalent.

See https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html

```python
from nilearn.glm.first_level import FirstLevelModel
from nilearn._utils.estimator_checks import return_expected_failed_checks

fc = return_expected_failed_checks(FirstLevelModel())

for k, v in fc.items():
    if v == "TODO":
        print(f'- [ ] {k}')
```

- [x] check_dict_unchanged
- [x] check_fit_score_takes_y
- [ ] check_estimators_nan_inf
- [x] check_estimators_overwrite_params
- [x] check_fit_idempotent
- [x] check_methods_sample_order_invariance
- [ ] check_methods_subset_invariance
- [ ] check_positive_only_tag_during_fit
- [x] check_pipeline_consistency
- [ ] check_readonly_memmap_input

Issue #5445, opened 2025-06-13
sklearn estimators checks to implement for decomposition estimators


Remi-Gau:
Relates to #4538 

These are the sklearn checks for HierarchicalKMeans that are currenctly failing and need to be either ignored or replaced by a nilearn equivalent.

See https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html

```python
from nilearn.decomposition._base import _BaseDecomposition
from nilearn._utils.estimator_checks import return_expected_failed_checks

fc = return_expected_failed_checks(_BaseDecomposition())

for k, v in fc.items():
    if v == "TODO":
        print(f'- [ ] {k}')
```

- [ ] check_estimators_empty_data_messages
- [ ] check_estimators_pickle
- [ ] check_estimators_nan_inf
- [ ] check_estimators_overwrite_params
- [ ] check_fit_idempotent
- [ ] check_methods_sample_order_invariance
- [ ] check_methods_subset_invariance
- [ ] check_positive_only_tag_during_fit
- [ ] check_pipeline_consistency
- [ ] check_readonly_memmap_input

bthirion:
Split the list into "to ignore" and "to address"

Issue #5444, opened 2025-06-13
sklearn estimators checks to implement for decoders


Remi-Gau:
Relates to #4538 

These are the sklearn checks for HierarchicalKMeans that are currenctly failing and need to be either ignored or replaced by a nilearn equivalent.

See https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html

```python
from nilearn.decoding.decoder import _BaseDecoder
from nilearn._utils.estimator_checks import return_expected_failed_checks

fc = return_expected_failed_checks(_BaseDecoder())

for k, v in fc.items():
    if v == "TODO":
        print(f'- [ ] {k}')
```

- [ ] check_dict_unchanged
- [ ] check_estimators_dtypes
- [ ] check_estimators_empty_data_messages
- [ ] check_estimators_pickle
- [ ] check_estimators_nan_inf
- [ ] check_estimators_overwrite_params
- [ ] check_fit_check_is_fitted
- [ ] check_fit_idempotent
- [ ] check_fit_score_takes_y
- [ ] check_methods_sample_order_invariance
- [ ] check_methods_subset_invariance
- [ ] check_positive_only_tag_during_fit
- [ ] check_pipeline_consistency
- [ ] check_readonly_memmap_input
- [ ] check_supervised_y_2d

bthirion:
Maybe split the list into what should be ignored and what should be addressed.

Remi-Gau:
realizing that SpaceNet estimators are not checked for compatibility with sklearn...

PR #5509, opened 2025-07-03
[FIX] ensure better compliance of CacheMixin with sklearn guidelines

Remi-Gau:
- Relates to #5447, #5446, #5445, #5444, #5443

- make sure several estimators "memory" attributes passed at init are not modified at fit time and create a memory_ attributes instead


nilearn/_utils/cache_mixin.py
```diff
@@ -226,19 +231,9 @@ def _cache(self, func, func_memory_level=1, shelve=False, **kwargs):
...
         """
-        verbose = getattr(self, "verbose", 0)
-
-        # Creates attributes if they don't exist
-        # This is to make creating them in __init__() optional.
-        if not hasattr(self, "memory_level"):
-            self.memory_level = 0
-        if not hasattr(self, "memory"):
-            self.memory = Memory(location=None, verbose=verbose)
-        self.memory = check_memory(self.memory, verbose=verbose)
```
Remi-Gau, line 237: remove because this could modify an object at transform time

bthirion (commented):
The PR lGTM. Should be mergeable soon ?

nilearn/_utils/estimator_checks.py
```diff
@@ -559,18 +544,23 @@ def nilearn_check_generator(estimator: BaseEstimator):
...
         yield (clone(estimator), check_masker_smooth)
+        yield (clone(estimator), check_masker_transform_resampling)
         yield (clone(estimator), check_masker_transformer)
         yield (
             clone(estimator),
             check_masker_transformer_high_variance_confounds,
         )
 
+        if isinstance(estimator, NiftiMasker):
+            # TODO enforce for other maskers
```
Remi-Gau, line 567: implement shelving for other maskers in follow up PR?
- to be discussed

Remi-Gau:
> Should be mergeable soon ?

yup 
only one point to discuss but that is not a blocker for the merge

nilearn/_utils/estimator_checks.py
Remi-Gau: https://joblib.readthedocs.io/en/latest/memory.html#shelving-using-references-to-cached-values

nilearn/_utils/estimator_checks.py
```diff
@@ -2064,6 +2045,69 @@ def check_masker_fit_score_takes_y(estimator):
...
+    # imgs return by inverse_transform impossible to save
+    if accept_niimg_input(estimator):
+        cachedir = Path(mkdtemp())
+        estimator.memory = Memory(location=cachedir, mmap_mode="r")
+        X = estimator.transform(img)
+
+        # inverse_transform a first time, so that the result is cached
+        out_img = estimator.inverse_transform(X)
+
+        out_img = estimator.inverse_transform(X)
```
bthirion, line 2107: Can you make it clear why we need the second call ?

nilearn/_utils/cache_mixin.py
```diff
@@ -213,10 +228,6 @@ def _cache(self, func, func_memory_level=1, shelve=False, **kwargs):
             The memory_level from which caching must be enabled for the wrapped
             function.
 
-        shelve : bool, default=False
-            Whether to return a joblib MemorizedResult, callable by a .get()
-            method, instead of the return value of func.
-
```
hndgzkn, line 219: `shelve` is still a parameter in method signature. Is there a reason to remove it?

nilearn/_utils/cache_mixin.py
```diff
@@ -195,6 +195,21 @@ class CacheMixin:
...
     """
 
+    def _fit_cache(self):
+        """Set attributes during estimator fit.
+
+        _shelving : bool
+            Used during calls to the method _cache
+            to return a joblib MemorizedResult,
+            callable by a .get() method,
+            instead of the return value of func.
```
hndgzkn: `_shelving` is not a parameter of `_fit_cache`. Is there a reason to document it here?

nilearn/_utils/cache_mixin.py
```diff
@@ -226,19 +237,9 @@ def _cache(self, func, func_memory_level=1, shelve=False, **kwargs):
...
-        if not hasattr(self, "memory_level"):
-            self.memory_level = 0
-        if not hasattr(self, "memory"):
-            self.memory = Memory(location=None, verbose=verbose)
-        self.memory = check_memory(self.memory, verbose=verbose)
-
         # If cache level is 0 but a memory object has been provided, set
         # memory_level to 1 with a warning.
-        if self.memory_level == 0 and self.memory.location is not None:
+        if self.memory_level == 0 and self.memory_.location is not None:
```
hndgzkn, line 236: It might be good to also check `self.memory_` is not `None`.

nilearn/_utils/estimator_checks.py
```diff
@@ -2064,6 +2045,69 @@ def check_masker_fit_score_takes_y(estimator):
...
+    """Check behavior when shelving masker."""
+    img, _ = generate_data_to_fit(estimator)
+
+    estimator.verbose = 0
+
+    masker = clone(estimator)
+
+    epi = masker.fit_transform(img)
+
+    masker = clone(estimator)
```
hndgzkn: This `masker` seems to be unused.

hndgzkn (approved):
LGTM

nilearn/_utils/cache_mixin.py
Remi-Gau: nope

nilearn/_utils/cache_mixin.py
Remi-Gau: good point
self.memory_ will be set to something else than None when an estimator is fitted
so I will just use a check_fitted here

nilearn/_utils/estimator_checks.py
Remi-Gau: indeed

nilearn/_utils/estimator_checks.py
Remi-Gau: my bad excuse: it was in the original test I was inspired by

https://github.com/nilearn/nilearn/blob/5ec3e093e73e7dc3eaf8f91d465e93b4835c3bdf/nilearn/maskers/tests/test_nifti_masker.py#L293

Remi-Gau: was added in this commit: https://github.com/nilearn/nilearn/commit/7e792981f0063a47cab1c3be09287cfba4862a4f#diff-603e6b98ace002d93bbfaf6475072f3a4fd5f43bdd0426b3f415b657bb41221f

a little "while" back... (11 years)

my understanding is that because we are testing caching behavior then some operation need to be executed twice

nilearn/_utils/cache_mixin.py
Remi-Gau: nope

nilearn/_utils/cache_mixin.py
Remi-Gau: ok to be extra safe I am going to modify the way each estimator that uses cache_mixin checks it is fitted to include something to make sure memory_ has been set properly

Remi-Gau: > ok to be extra safe I am going to modify the way each estimator that uses cache_mixin checks it is fitted to include something to make sure memory_ has been set properly

this won't work as cache is sometimes called during the fit of some estimators

Issue #2293, opened 2020-01-23
it is not clear from the documention what  NiftiMapsMarker computes


chrplr:
I intended to use NiftiMapsMasker to extract the average 'activations' in regions of interests.

I noticed:
1. that the returned value with a binary mask is the sum rather than the average
2. that when the mask is not binary, the value is not the sum (probably the  dot product of voxel values in the mask and in the image)

The documentation of NiftiMapsMasker is lacking details explaining the returned values.
Also, shouldn't NiftiMapsMasker behave more similiartly uo NiftiLabelMAsker, that is, have a 'strategy' option to chose the operation to apply to the voxel values?

htwangtw:
I am looking into NiftiMasker performance and here's my understanding of the issue.
`NiftiLablesMasker` and `NiftiMapsMasker` uses different ways to extract signal per parcel and the method of extraction cannot be made into comparable behaviour as you described here. If there are comparable solutions, I assume the dev would not have made them into different classes 
`NiftiLablesMasker` uses functions under `scipy.ndimage`: https://github.com/nilearn/nilearn/blob/e1c2f8bd9cd22d9cce9028bb3a56ac7df73810b5/nilearn/regions/signal_extraction.py#L123
`NiftiMapsMasker` compute least-squares solution between the signal in the nifti file and the soft map: https://github.com/nilearn/nilearn/blob/e1c2f8bd9cd22d9cce9028bb3a56ac7df73810b5/nilearn/regions/signal_extraction.py#L291-L292
If you are using a binary mask you should be using `NiftiLablesMasker` for the behaviour you want.

PR #5511, opened 2025-07-04
[FIX] ensure image estimator can fit and preserve several dtype

Remi-Gau:
- Relates to #5447, #5446, #5445, #5444

- [x] fixes missing dtype attribute for surface maskers
- [x] ensure that estimators with a transform / inverse_transform method can output arrays of the requested dtype
- [x] update changelog


nilearn/_utils/estimator_checks.py
```diff
@@ -979,6 +983,78 @@ def check_img_estimator_pickle(estimator_orig):
...
+    np.int64 not tested: see no_int64_nifti in nilearn/conftest.py
+    """
+    for dtype in [np.float32, np.float64, np.int32]:
+        estimator = clone(estimator)
+
+        estimator.dtype = dtype
+
+        if isinstance(estimator, (NiftiLabelsMasker, SurfaceLabelsMasker)):
+            # use strategy that can conserve dtype
+            estimator.strategy = "median"
```
Remi-Gau: side note: why only label maskers have different strategies? could we not have them also for map maskers?

nilearn/maskers/tests/test_base_masker.py
```diff
@@ -47,3 +48,33 @@ def test_cropping_code_paths(rng):
     out_data_cropped = filter_and_mask(img, cropped_mask_img, parameters)
 
     assert_array_almost_equal(out_data_cropped, out_data_uncropped)
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64, np.int32, "auto"])
```
Remi-Gau: np.int32 fails but because the output data is never coerced to that type

but our doc does not mention this: I think we should probably throw a warning when trying to mask data to an integer type

nilearn/_utils/niimg.py
```diff
@@ -69,32 +70,6 @@ def safe_get_data(img, ensure_finite=False, copy_data=False) -> np.ndarray:
     return data
 
 
-def _get_target_dtype(dtype, target_dtype):
```
Remi-Gau, line 99: moved to numpy_conversions

nilearn/surface/surface.py
```diff
@@ -1492,6 +1492,13 @@ def to_filename(self, filename):
 
         _data_to_gifti(data, filename)
 
+    def _dtype(self):
+        """Return dtype of the first part.
+
+        Assume all parts have same dtype.
```
Remi-Gau, line 1508: We may want to ensure that SurfaceImages have similar type for both hemispheres, no?

Remi-Gau: Could enforce a certain dtype to Surface images at init

nilearn/_utils/estimator_checks.py
Remi-Gau: probably for another PR

Remi-Gau: nope: see https://github.com/nilearn/nilearn/issues/2293

Remi-Gau:
Discussed in the dev meeting: should there be restrictions / warnings when user require some dtype for some maskers?

nilearn/surface/surface.py
emdupre: Cross-linking that this is now to-be-discussed in #5524 !

nilearn/_utils/estimator_checks.py
```diff
@@ -132,7 +134,7 @@ def check_estimator(estimators: list[BaseEstimator], valid: bool = True):
     ----------
     estimators : list of estimator object
         Estimator instance to check.
-
+    dtype
```
bthirion: to be done ?

nilearn/_utils/estimator_checks.py
```diff
@@ -1120,6 +1133,137 @@ def check_img_estimator_pickle(estimator_orig):
...
 
 
+@ignore_warnings()
+def check_img_estimator_dtypes(estimator):
+    """Check estimator can fit/transform/inverse_transform \
+       with inputs of varying dtypes.
+
+    Replacement for sklearn check_estimators_dtypes.
+
+    For some estimators that only extract data (like niftimasker)
```
bthirion:
```suggestion
    For some estimators that only extract data (like NiftiMasker)
```

nilearn/_utils/estimator_checks.py
```diff
@@ -1120,6 +1133,137 @@ def check_img_estimator_pickle(estimator_orig):
...
+    and do not aggregate it (like maps masker)
```
bthirion:
```suggestion
    and do not aggregate it (like NiftiMapsMasker)
```

nilearn/_utils/numpy_conversions.py
```diff
@@ -130,3 +130,29 @@ def csv_to_array(csv_path, delimiters=" \t,;", **kwargs):
...
+
+    Returns
+    -------
+    dtype : dtype
+        The data type toward which the original data should be converted.
+    """
+    if target_dtype is None:
+        return None
+    if target_dtype == "auto":
+        target_dtype = np.int32 if dtype.kind == "i" else np.float32
```
bthirion, line 163:
```suggestion
        target_dtype = 'int32' if dtype.kind == "i" else 'float32'
``` (I thought we rather use core Python dtypes ?)

**Review by bthirion — COMMENTED** [review]
LGTM, with minor issues.

File: nilearn/surface/surface.py (hunk 4 above)
Comment (Remi-Gau) [reply nilearn/surface/surface.py:1508]: done in #5530

File: nilearn/_utils/estimator_checks.py (hunk 5 above)
Comment (Remi-Gau) [reply nilearn/_utils/estimator_checks.py]: nope: probably I went wild on the copy - paste

File: nilearn/_utils/numpy_conversions.py (hunk 8 above)
Comment (Remi-Gau) [reply nilearn/_utils/numpy_conversions.py:163]: fixing: should not make a big difference though

File: nilearn/maskers/nifti_masker.py [hunk 9]
```diff
@@ -615,7 +616,15 @@ def transform_single_imgs(
             confounds=confounds,
             sample_mask=sample_mask,
             copy=copy,
-            dtype=self.dtype,
         )
 
-        return data
+        if not isinstance(data, np.ndarray):
+            # in case data data is a cached MemorizedResult
```
Remi-Gau:
```suggestion
            # in case data is a cached MemorizedResult
            # only happens when _shelving is True
```

nilearn/maskers/nifti_masker.py
```diff
@@ -615,7 +616,16 @@ def transform_single_imgs(
...
-        return data
+        if not isinstance(data, np.ndarray):
+            # in case data is a cached MemorizedResult
+            # only happens when _shelving is True
+            return data
+
+        imgs = load_img(imgs)
+        target_dtype = get_target_dtype(img_data_dtype(imgs), self.dtype)
+        if target_dtype is None:
+            target_dtype = img_data_dtype(imgs)
```
Remi-Gau, line 707: for this masker and several others the new type is set after extracting / filtering the data rather than during this process (was trying to see if it was possible to do during, but 1) could not make it work, 2) casting to a different dtype may be better done as a last step because of rounding issues that may occur when going from float to int.

but only for this Masker and MultiMasker do we need extra check to deal with potentially MemorizedResult because only for those we use and test behavior with _shelving

https://github.com/nilearn/nilearn/issues/5523

nilearn/decomposition/_base.py
```diff
@@ -394,6 +398,7 @@ def __init__(
         low_pass=None,
         high_pass=None,
         t_r=None,
+        dtype=None,
```
bthirion, line 409: Adding a dtype here is means that a given masker will always transform data toward this unique dtype. Aren't there cases where this behavior is too rigid ?

nilearn/maskers/nifti_masker.py
bthirion: You mean, just add additional checks that would detect a potential caching issue that would result in wrong dtype ? If it's bot too much of a hassle, this makes sense.

nilearn/_utils/estimator_checks.py
```diff
@@ -1102,6 +1117,207 @@ def check_img_estimator_pickle(estimator_orig):
...
+        ]:
+            estimator = clone(estimator_orig)
+
+            if hasattr(estimator, "dtype"):
+                estimator.dtype = dtype
+
+            input_dtype = np.dtype(input_dtype)
+            if isinstance(estimator, NiftiMasker) and input_dtype == np.int32:
+                # FIXME
+                # This leads to some invalid mask for NiftiMasker
```
bthirion: Is it a bug or a feature ?

bthirion (commented):
Technically, the PR seems good. I'm still unsure about high-level questions on the behavior.

nilearn/maskers/nifti_masker.py
Remi-Gau: yeah I'd probably need another test for this

nilearn/_utils/estimator_checks.py
Remi-Gau: not sure yet: I think this is more an error in the data that I use to set up the test but I have not been able to figure out what I am doing wrong (yet)

nilearn/decomposition/_base.py
Remi-Gau: well not sure if it's too rigid: if the input image is in float, asking a label or maps masker to return some int will lead to some rounding of the results.

but I suspect that users then have good reasons to do this

note that passing int64 should results in some warning if trying to inverse_transform is called because nibabel will not create such image and will downcast it

bthirion: It means that in some cases, users will have to instantiate several maskers, e.g. one for signal and one for labels. But since the interpolation scheme is likely not the same depending on the data type, that sounds reasonable.

bthirion:
LMK when this is ready for the next round.

nilearn/maskers/nifti_masker.py
bthirion: LMK when you're done.

Remi-Gau: I have added some extra tests to make sure things are the same when caching

nilearn/_utils/estimator_checks.py
Remi-Gau: ok this was because the default mask creation strategy based on 'background' would return an empty mask: not sure if that's a bug with [the mask strategy](https://github.com/nilearn/nilearn/blob/015061b275b5f5349067f042e077dd792edfaf4f/nilearn/masking.py#L477): should I investigate further 

switching to an "epi" strategy works

Remi-Gau:
given the recent bug flagged and fixed by @tsalo on data type I feel, we should bring this PR up to date and merge it to ensure our estimator don't start changing data type in unexpected way

given the merge conflixt situation I am most likely going to only keep the estimator checks from this PR and we will be left with a bunch of failing tests to fix

.github/workflows/test_with_tox.yml
```diff
@@ -91,7 +91,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```
revert before merge

nilearn/_utils/estimator_checks.py
```diff
@@ -869,6 +889,13 @@ def check_set_output(estimator_orig) -> None:
         signal = estimator.transform(img)
 
 
+def check_transformer_set_output(estimator):
+    """Check that set_ouput throws a not implemented error."""
+    if hasattr(estimator, "transform"):
+        with pytest.raises(NotImplementedError):
+            estimator.set_output(transform="default")
+
+
```
```suggestion
```

nilearn/maskers/nifti_masker.py
```diff
@@ -200,6 +199,10 @@ def filter_and_mask(
     # we return 1D array
     if temp_imgs.ndim == 3 and sklearn_output_config is None:
         data = data.squeeze()
+
+    # TODO implement
```
bthirion: ?

bthirion (approved):
LGTM, thx.

nilearn/maskers/nifti_masker.py
Remi-Gau: that's left over from some old tweaking: let me clean up

PR #5517, opened 2025-07-08
[DOC] explain how map masker extract data

### LINKED ISSUE #2293: it is not clear from the documention what  NiftiMapsMarker computes

Remi-Gau:
- Closes #2293

- explain that maps masker extract data via least square

### DIFF
--- a/doc/manipulating_images/masker_objects.rst
+++ b/doc/manipulating_images/masker_objects.rst
@@ -422,13 +422,21 @@ This atlas defines its regions using maps. The path to the corresponding
 file is given in the ``maps_img`` argument.
 
 One important thing that happens transparently during the execution of
-:meth:`NiftiMasker.fit_transform` is resampling. Initially, the images
+:meth:`NiftiMapsMasker.fit_transform` is resampling. Initially, the images
 and the atlas do typically not have the same shape nor the same affine.
 Casting them into the same format is required for successful signal extraction
 The keyword argument ``resampling_target`` specifies which format
 (i.e., dimensions and affine) the data should be resampled to.
 See the reference documentation for :class:`NiftiMapsMasker` for every
 possible option.
+
+.. warning::
+
+    :class:`NiftiMapsMasker` does the data extraction
+    via least square regression, such that ``maps @ x = data``.
+    If you want to extract data from a map in another way (mean, max...),
+    you should use :class:`NiftiLabelsMasker`
+    on a thresholded map.
 
 .. topic:: **Examples**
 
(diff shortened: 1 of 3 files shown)


doc/manipulating_images/masker_objects.rst
```diff
@@ -422,14 +422,22 @@ This atlas defines its regions using maps. The path to the corresponding
...
 See the reference documentation for :class:`NiftiMapsMasker` for every
 possible option.
 
+.. warning::
+
+    :class:`NiftiMapsMasker` does the data extraction
+    via least square regression, such that ``maps @ x = data``.
+    If you want to extract data from a map in another way (mean, max...),
+    you should use :class:`NiftiLabelsMasker`
+    on a thresholded map.
```
Remi-Gau, line 439: I kinda feel this should be shown in an example.

bthirion: You want to show how to extract a median instead of a mean by binarizing and doing a call to NiftiLabelsMasker ?
Not super enthusiastic: I don't want to encourage that, since people will be struggling with the discretization of the labels, which is generally a bit cumbersome.

nilearn/maskers/nifti_maps_masker.py
```diff
@@ -57,6 +57,15 @@ class NiftiMapsMasker(BaseMasker):
         Inf or NaN present in the given input images are automatically
         put to zero rather than considered as missing data.
 
+    .. warning::
+
+        The data extraction is done via least square regression,
+        such that ``map @ x = data``.
+        If you want to extract data
```
bthirion, line 64: I'm not sure I want to add that comment.

bthirion (commented):
LGTM, but I 'm not 100% sure about the recommandation you make.

doc/manipulating_images/masker_objects.rst
htwangtw: I second this

man-shu: I think simply mentioning that NiftiMapsMasker does the extraction via least squares should be enough, no?

htwangtw: We can roll this out and see if users understand.

PR #5512, opened 2025-07-04
[FIX] issue 5493

hndgzkn:
- Closes #5493 

Changes proposed in this pull request:

-
-

Issue #5493, opened 2025-06-25
Calculating data bounds and thresholding data can be unified for  `plotly` and `matplotlib` backends in surface plotting


hndgzkn:
> * `plotly` and `matplotlib` backends calculate data bounds and thresholds in different functions. I think those could be unified. How does it sound to you?

yes let's do the refactor in a follow up PR

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5492#discussion_r2166268804_

Remi-Gau:
@hndgzkn 
do you know if this has been fixed?

hndgzkn:
Data bounds part is ok, thresholding not yet. We can discuss that.

PR #5521, opened 2025-07-09
[ENH] Refactor plotting to extract engine utils

hndgzkn:
- Closes None

- Move utility functions importing from `matplotlib`  and used by multiple modules to `plotting._engine_utils`.

> There is no change in code, just functions are moved across modules.

### DIFF
--- a/nilearn/plotting/_engine_utils.py
+++ b/nilearn/plotting/_engine_utils.py
@@ -1,0 +1,113 @@
+from warnings import warn
+
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.colors import (
+    LinearSegmentedColormap,
+    ListedColormap,
+    Normalize,
+)
+
+from nilearn._utils.extmath import fast_abs_percentile
+from nilearn._utils.logger import find_stack_level
+from nilearn._utils.param_validation import check_threshold
+
+
+def colorscale(
+    cmap, values, threshold=None, symmetric_cmap=True, vmax=None, vmin=None
+):
+    """Normalize a cmap, put it in plotly format, get threshold and range."""
+    cmap = plt.get_cmap(cmap)
+    abs_values = np.abs(values)
+
+    if (
+        symmetric_cmap
+        and vmin is not None
+        and vmax is not None
+        and vmin != -vmax
+    ):
+        warn(
+            f"Specified {vmin=} and {vmax=} values do not create a symmetric"
+            " colorbar. The values will be modified to be symmetric.",
+            stacklevel=find_stack_level(),
+        )
+    if vmax is None:
+        vmax = abs_values.max()
+    if vmin is None:
+        vmin = values.min()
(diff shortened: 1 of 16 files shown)


hndgzkn:
@Remi-Gau @bthirion We have not discussed [this](https://github.com/nilearn/nilearn/pull/5481#discussion_r2190908313) yesterday. If that is convenient, I can update `_engine_utils.py` to `engine_utils.py` but not put functions to `__init__.__all__` to make them private to package.

hndgzkn:
> @Remi-Gau @bthirion We have not discussed [this](https://github.com/nilearn/nilearn/pull/5481#discussion_r2190908313) yesterday. If that is convenient, I can update `_engine_utils.py` to `engine_utils.py` but not put functions to `__init__.__all__` to make them private to package.

Actually for this one this might not be necessary as these functions are not imported from other packages.

Remi-Gau:
is that a split from #5493?

hndgzkn:
> is that a split from #5493?

From #5512

Remi-Gau:
maybe I am missing something but why is this move needed ?

Remi-Gau:
> * Move utility functions for matplotlib backend used by multiple modules

the functions moved are not just for matplotlib, no?

nilearn/plotting/_engine_utils.py
Remi-Gau, line 1: I think we should probably start adding a doc string to explain what each module contains even if they are private modules.

bthirion:
> > @Remi-Gau @bthirion We have not discussed [this](https://github.com/nilearn/nilearn/pull/5481#discussion_r2190908313) yesterday. If that is convenient, I can update `_engine_utils.py` to `engine_utils.py` but not put functions to `__init__.__all__` to make them private to package.
> 
> Actually for this one this might not be necessary as these functions are not imported from other packages.

Yes, it looks like there is no need

hndgzkn:
> maybe I am missing something but why is this move needed ?

I try to gather functions importing from matplotlib in certain modules to avoid matplotlib imports everywhere and make the modules more coherent. 

This would also help finding the relevant functions for developers. There are many code pieces implemented the same way in the code and spread in different modules.

hndgzkn:
> > * Move utility functions for matplotlib backend used by multiple modules
> 
> the functions moved are not just for matplotlib, no?

You are right. They are also used by plotly backend, html modules, etc. The way I expressed it is not correct. 

These are the utility functions that import `matplotlib`. I have not put them all here yet. For example there is `save_figure_if_needed`. It requires more reflection I think.

The next step will be to extract common code spread in different modules, and extract new functions in this module. 

This is what I did in https://github.com/nilearn/nilearn/pull/5512 partially. But when functions are moved and then I make changes, it is not really possible to see the changes. So I wanted to do it in two steps so that we can discuss.

Remi-Gau:
Given this just moves function I feel this can be merged with a single review.

PR #5530, opened 2025-07-09
[ENH] enforce dtype for all parts of SurfaceImage and Polydata

### LINKED ISSUE #5524: ensure that SurfaceImages have similar dtype for both hemispheres
We may want to ensure that SurfaceImages have similar type for both hemispheres

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5511#discussion_r2189934499_

Remi-Gau:
- Closes #5524 

- add errors and test for instantiation of SurfaceImage and Polydata with specific dtype

### DIFF
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -1352,6 +1352,10 @@ class PolyData:
             the typical shape of the
             data for a hemisphere is ``(n_vertices, n_time_points)``.
 
+    dtype : DTypeLike object, default=None
+        dtype to enforce on the data.
+        If ``None`` the original dtype if used.
+
     Examples
     --------
     >>> import numpy as np
@@ -1372,7 +1376,7 @@ class PolyData:
     ValueError: Cannot create an empty PolyData. ...
     """
 
-    def __init__(self, left=None, right=None):
+    def __init__(self, left=None, right=None, dtype=None):
         if left is None and right is None:
             raise ValueError(
                 "Cannot create an empty PolyData. "
@@ -1386,6 +1390,7 @@ class PolyData:
                     param = load_surf_data(param)
                 parts[hemi] = param
         self.parts = parts
+        self._set_data_dtype(dtype)
 
         self._check_parts()
 
@@ -1404,6 +1409,13 @@ class PolyData:
                 f"Data arrays for keys 'left' and 'right' "
                 "have incompatible shapes: "
                 f"{parts['left'].shape} and {parts['right'].shape}"
+            )
+
+        if parts["left"].dtype != parts["right"].dtype:
+            raise TypeError(
+                "All parts should have same dtype. "
+                f"Got {parts['left'].dtype=} and {parts['right'].dtype=}. "
+                "You can fix this by passing a 'dtype' at instantiation."
             )
 
     @property
@@ -1491,6 +1503,11 @@ class PolyData:
             data = self.parts["right"]
 
         _data_to_gifti(data, filename)
+
+    def _set_data_dtype(self, dtype):
+        if dtype is not None:
+            for h, v in self.parts.items():
+                self.parts[h] = v.astype(dtype)
 
 
 def at_least_2d(input):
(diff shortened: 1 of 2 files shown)


Remi-Gau:
Should dtype be also enforced for the different hemispheres in a polymesh?

bthirion (approved):
LGTM, thx. 
Maybe to be put in the changelog.

nilearn/surface/surface.py
```diff
@@ -1386,6 +1390,7 @@ def __init__(self, left=None, right=None):
                     param = load_surf_data(param)
                 parts[hemi] = param
         self.parts = parts
+        self._set_data_dtype(dtype)
```
hndgzkn, line 1395: May be this should be go after `self._check_parts`. Otherwise, the check in line `1414` becomes unnecessary if I am not missing something.

nilearn/surface/surface.py
```diff
@@ -1893,8 +1907,9 @@ def __init__(self, mesh, data):
 
         if isinstance(data, PolyData):
             self.data = data
+            self.data._set_data_dtype(dtype)
         elif isinstance(data, dict):
-            self.data = PolyData(**data)
+            self.data = PolyData(**data, dtype=dtype)
```
hndgzkn: May be lines 1903-1906 could be moved to  `else` part of this `if` block.

nilearn/surface/surface.py
```diff
@@ -1893,8 +1907,9 @@ def __init__(self, mesh, data):
 
         if isinstance(data, PolyData):
             self.data = data
+            self.data._set_data_dtype(dtype)
```
hndgzkn, line 1915: If `_set_data_dtype` will be accessed from outside the class, shall we remove `_`?

nilearn/surface/surface.py
hndgzkn: Or may be you wanted to check difference in `dtype` only when `dtype` is not set.

hndgzkn (approved):

nilearn/surface/surface.py
```diff
@@ -1869,20 +1886,17 @@ class SurfaceImage:
...
            Data for the both hemispheres.
 
-    squeeze_on_save : :obj:`bool` or None, default=None
-            If ``True`` axes of length one from the data
-            will be removed before saving them to file.
-            If ``None`` is passed,
-            then the value will be set to ``True``
-            if any of the data parts is one dimensional.
+    dtype : DTypeLike object, default=None
+        dtype to enforce on the data.
```
Remi-Gau, line 1892: At the moment, this PR does not check if the mesh have consistent dtype across hemispheres but I feel this is less important.

nilearn/surface/surface.py
Remi-Gau: reluctant to add another thing to the public API for now: next release is supposed to only fix bugs and this PR already allows to set the the dtype at init

nilearn/surface/surface.py
Remi-Gau: you mean this?

```
        if not isinstance(data, (PolyData, dict)):
            raise TypeError(
                f"'data' must be one of[PolyData, dict].\nGot {type(data)}"
            )
```

nilearn/surface/surface.py
Remi-Gau: if the user does not pass a dtype, _set_data_dtype does nothing and then we want to make sure that all parts have the same dtype

nilearn/surface/surface.py
```diff
@@ -1869,32 +1888,33 @@ class SurfaceImage:
...
+        If ``None`` the original dtype if used.
```
hndgzkn:
```suggestion
        If ``None`` the original dtype is used.
```

nilearn/surface/surface.py
hndgzkn: Yes.

Remi-Gau: I prefer to have single guard clause up early in the code rather than at the very end of an if, ifelse, else block.

PR #5531, opened 2025-07-09
[REF] speed up some slow tests

Remi-Gau:
- Closes none

- hopefully this should speed up some tests


nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -164,11 +165,14 @@ def test_plot_carpet_default_params(img_4d_mni, img_3d_ones_mni):
...
 @pytest.mark.mpl_image_compare
-def test_plot_prob_atlas_default_params(img_3d_mni, img_4d_mni):
+def test_plot_prob_atlas_default_params(
+    img_3d_mni, shape_3d_default, affine_mni
+):
     """Smoke-test for plot_prob_atlas with default arguments."""
-    return plot_prob_atlas(img_4d_mni, bg_img=img_3d_mni)
+    # using only 2 regions to speed up the test
+    maps = generate_maps(shape_3d_default, n_regions=2, affine=affine_mni)
+    return plot_prob_atlas(maps, bg_img=img_3d_mni)
```
Remi-Gau: OK this will have to wait for 0.13 because I cannot generate a new figure locally because I cannot run the antiquated matplotlib 3.3 because of https://github.com/nilearn/nilearn/issues/5261

bthirion (approved):
LGTM, thx

nilearn/plotting/surface/tests/test_surf_plotting.py
```diff
@@ -131,7 +131,6 @@ def test_plot_surf(plt, engine, tmp_path, in_memory_mesh, bg_map):
 
     # Plot mesh with background
     plot_surf(in_memory_mesh, bg_map=bg_map, engine=engine)
-    plot_surf(in_memory_mesh, bg_map=bg_map, darkness=0.5, engine=engine)
```
emdupre, line 134: I guess this is removed since we're deprecating `darkness` ?

nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -168,6 +168,9 @@ def test_plot_carpet_default_params(img_4d_mni, img_3d_ones_mni):
 @pytest.mark.mpl_image_compare
 def test_plot_prob_atlas_default_params(img_3d_mni, img_4d_mni):
     """Smoke-test for plot_prob_atlas with default arguments."""
+    # TODO for 0.13.0
```
emdupre, line 171: Just to make sure I understand this to-do, why is this for 0.13.0 rather than 0.12.1 ?

emdupre (approved):
Small questions just to make sure I understand, but the changes LGTM ! Thanks, @Remi-Gau !

nilearn/plotting/surface/tests/test_surf_plotting.py
Remi-Gau: yes exactly

nilearn/plotting/tests/test_baseline_comparisons.py
Remi-Gau: We only compare the figure generated by the tests in that file to the baseline figure  when testing with python 3.9 and the oldest supported version of our dependencies 
So we must generate the baseline figures with the same version of the dependencies: but I cannot generate the figures because of https://github.com/nilearn/nilearn/pull/5531#discussion_r2195532885

This should not be an issue in version 0.13 when move to matplotlib 3.8 (ish).

PR #5534, opened 2025-07-10
added my name to citation list

gurraburra:
Never added my name to citation list when contributing to PR #4501, I have now done that.


bthirion (approved):
LGTM, thx.

PR #5535, opened 2025-07-10
[MAINT] add discovery module

Remi-Gau:
- Closes none

- add a discovery module (mostly taken fromm skearn.utils.discovery) to help find all the possibles estimators and functions that users can access. 
- Add a check during the doc build to make sure the number of objects in the user facing API does not change by mistake.
- [x] update changelog


Remi-Gau:
@hndgzkn 
this adds a small extra check that should make sure we cannot change the user facing API by mistake

maint_tools/utils.py
```diff
@@ -90,6 +90,14 @@ def list_nodes(
     return [c for c in node_definitions if not c.name.startswith("_")]
 
 
+# list objects that are part of the user facing API
+# the code below assumes
+# that we do not have more one 2 levels of "nesting" of subpacakges
```
hndgzkn:
```suggestion
# that we do not have more one 2 levels of "nesting" of subpackages
```

nilearn/datasets/__init__.py
```diff
@@ -71,9 +71,9 @@
 from .utils import get_data_dirs
 
 __all__ = [
-    "GM_MNI152_FILE_PATH",
-    "MNI152_FILE_PATH",
-    "WM_MNI152_FILE_PATH",
+    "GM_MNI152_FILE_PATH",  # not in the doc or used in the codebase
+    "MNI152_FILE_PATH",  # not in the doc, only used in one example
+    "WM_MNI152_FILE_PATH",  # not in the doc or used in the codebase
```
hndgzkn, line 75: Are these comments to think about removing them from `__all__`?

maint_tools/utils.py
```diff
@@ -102,3 +110,13 @@ def list_nodes(
             submod = importlib.import_module(f"nilearn.{subpackage}.{x}")
             if hasattr(submod, "__all__"):
                 public_api.extend(submod.__all__)
```
hndgzkn, line 102: Does `submod.__all__` return the function names with namespaces? If not, when we do `set(public_api)` below, it might decrease the number of functions. For example does it unify `get_data` function for image and surface image?

hndgzkn (approved):
Some comments, otherwise LGTM.

maint_tools/utils.py
Remi-Gau: ha good catch!!!

nilearn/datasets/__init__.py
Remi-Gau: yeah kind of, not for now because technically they are part of the 'public API' even if undocumented, so this may just removing them without deprecation may break some downstream code

maint_tools/utils.py
Remi-Gau: should be fixed now
had to create a separate list because some other maintenance scripts rely on the short form of the public api list

maint_tools/utils.py
```diff
@@ -90,15 +90,45 @@ def list_nodes(
...
+# - nilearn/glm/first_level/__init__.py
+# but not
+# - nilearn/glm/first_level/too_deep/__init__.py
 public_api = ["nilearn"]
+public_api_with_subpackage = ["nilearn"]
 for subpackage in nilearn.__all__:
     public_api.append(subpackage)
+    public_api_with_subpackage.append(subpackage)
     if subpackage.startswith("_"):
         continue
```
hndgzkn, line 95:
```suggestion
    if subpackage.startswith("_"):
        continue
    public_api.append(subpackage)
    public_api_with_subpackage.append(subpackage)
```

hndgzkn: Otherwise if the package starts with `_`, it is added to the public_api list.

maint_tools/utils.py
```diff
@@ -90,15 +90,45 @@ def list_nodes(
...
         if inspect.ismodule(mod.__dict__[x]):
             submod = importlib.import_module(f"nilearn.{subpackage}.{x}")
             if hasattr(submod, "__all__"):
                 public_api.extend(submod.__all__)
+                public_api_with_subpackage.extend(
+                    [
+                        f"{subpackage}.{submod.__name__}.{x}"
+                        for x in submod.__all__
+                    ]
+                )
```
hndgzkn:
```suggestion
                public_api_with_subpackage.extend(
                    [
                        f"{subpackage}.{x}.{y}"
                        for y in submod.__all__
                    ]
                )
```

hndgzkn: As for loop also uses `x`, to make the code more clear, I suggest to use another variable name, `y`, here.

hndgzkn: Or, instead of x and y, more self defining variable names could be used.

Remi-Gau: actually x was already used higher up in the loop

Remi-Gau:
actually, we could even have some like this (list all the functions that exist as part of nilearn) as part of the public API:
skllearn has things like this: https://scikit-learn.org/stable/modules/generated/sklearn.utils.discovery.all_estimators.html

Remi-Gau:
> actually, we could even have some like this (list all the functions that exist as part of nilearn) as part of the public API: skllearn has things like this: https://scikit-learn.org/stable/modules/generated/sklearn.utils.discovery.all_estimators.html

OK I think I will actually try to implement this 'discovery' feature (inspired from sklearn) in this PR

nilearn/utils/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .discovery import all_estimators, all_functions
```
hndgzkn: Should this be absolute import?

nilearn/utils/tests/test_discovery.py
```diff
@@ -0,0 +1,35 @@
...
+    ],
+)
+def test_all_estimators(
+    matplotlib_pyplot,  # noqa : ARG001
+    type_filter,
+    n_expected,
+):
+    """Check number of estimators in public API."""
+    estimators = all_estimators(type_filter=type_filter)
+    print(estimators)
```
hndgzkn, line 32: This is to remove I guess.

nilearn/utils/tests/test_discovery.py
```diff
@@ -0,0 +1,35 @@
...
+    assert len(estimators) == n_expected
+
+
+def test_all_functions(
+    matplotlib_pyplot,  # noqa : ARG001
+):
+    """Check number of functions in public API."""
+    fn = all_functions()
+    print(fn)
```
hndgzkn, line 41: This is to remove I guess.

nilearn/utils/discovery.py
```diff
@@ -0,0 +1,188 @@
...
+        Possible values are
+        "classifier", "cluster", "regressor", "masker",
+        "multi_masker", "transformer"
+        to get estimators only of these specific types,
+        or a list of these to get the estimators
+        that fit at least one of the types.
+
+    Returns
+    -------
+    estimators : functions : :obj:`list` of :obj:`tuple`
```
hndgzkn: I could not understand why we have `functions :` here.

nilearn/utils/discovery.py
```diff
@@ -0,0 +1,188 @@
...
+        List of (name, class),
+        where ``name`` is the class name as string
+        and ``class`` is the actual type of the class.
+
+    """
+    # lazy import to avoid circular imports
+    from nilearn.decoding._mixin import _ClassifierMixin, _RegressorMixin
+    from nilearn.maskers._mixin import _MultiMixin
+    from nilearn.maskers.base_masker import BaseMasker, _BaseSurfaceMasker
```
hndgzkn: Do we have dependency from these packages to `nilearn.utils` ? I think there should not be.
 
We are introducing `nilearn.utils` package now as a public package. I think that would be good to define what it serves and its boundaries in `__init__.py` of the package. What I understand is its aim is to provide users with some useful functionality.

In my opinion, this package should be able to import from other packages, but no other package should import from this package. (At least, in a way that might create cyclic dependency. However in any case, I think we should avoid this)

nilearn/utils/discovery.py
```diff
@@ -0,0 +1,188 @@
...
+        "masker": (BaseMasker, _BaseSurfaceMasker),
+        "multi_masker": _MultiMixin,
+        "regressor": _RegressorMixin,
+        "transformer": TransformerMixin,
+    }
+
+    if type_filter is not None and type_filter not in allowed_filters:
+        check_parameter_in_allowed(
+            type_filter, allowed_filters.keys(), "type_filter"
+        )
```
hndgzkn:
```suggestion
    if type_filter is not None:
        check_parameter_in_allowed(
            type_filter, allowed_filters.keys(), "type_filter"
        )
```

hndgzkn: `check_parameter_in_allowed` might be modified to accept a list as well?

nilearn/utils/__init__.py
hndgzkn: A docstring to define the module and its intensions might be good.

nilearn/utils/tests/test_discovery.py
```diff
@@ -0,0 +1,35 @@
+import pytest
```
hndgzkn, line 5: Docstring for module?

hndgzkn (approved):
LGTM with some comments.

nilearn/utils/tests/test_discovery.py
Remi-Gau: yeah not 100% sure actually: it is kind of useful to be able to get a list of what's in the API when running the test

nilearn/utils/__init__.py
Remi-Gau: done

nilearn/utils/discovery.py
Remi-Gau: agreed: adding that restriction into the pyproject.toml

nilearn/utils/discovery.py
Remi-Gau: good catch: check_parameter_in_allowed already accepts a list

nilearn/utils/__init__.py
Remi-Gau: done

nilearn/utils/tests/test_discovery.py
Remi-Gau: done

nilearn/utils/discovery.py
Remi-Gau: copy paste error

bthirion:
Need a green doc build...

nilearn/utils/discovery.py
```diff
@@ -0,0 +1,254 @@
...
+    )
+
+
+def _is_abstract(c):
+    if not (hasattr(c, "__abstractmethods__")):
+        return False
+    return len(c.__abstractmethods__)
+
+
+def _get_all_classes():
```
bthirion, line 45: Maybe a mini-docstring would be welcome here

nilearn/utils/tests/test_discovery.py
bthirion: But it's going to clutter outputs...

bthirion (commented):
LGTM  with very minor details.

nilearn/utils/tests/test_discovery.py
Remi-Gau: yeah but that's only in the test (and to be fair our test output is already cluttered)

Remi-Gau:
failure is unrelated

PR #5538, opened 2025-07-10
[MAINT] define general architecture

Remi-Gau:
- Closes none

- use [import-linter](https://import-linter.readthedocs.io/en/stable/index.html) to define /constrain the general architecture of nilearn

At the moment we more or less have 3 layers:

top layer: those packages can import from any of the packages below but not between each other
bottom layer : those packages can import from any of the packages below, between each other but not from any above

```mermaid
flowchart TD
    subgraph high
        direction LR
        glm <--X--> decoding 
    end
    subgraph low
        direction LR 
        maskers
    end
    high --> low
```


nilearn/glm/first_level/design_matrix.py
```diff
@@ -81,53 +82,6 @@ def _poly_drift(order, frame_times):
     return pol
 
 
-def create_cosine_drift(high_pass, frame_times):
```
Remi-Gau: moved to nilearn.signal

nilearn/_utils/tags.py
```diff
@@ -93,3 +94,33 @@ class InputTags(SkInputTags):
 
         # glm
         glm: bool = False
+
+
+def get_tag(estimator: BaseEstimator, tag: str) -> bool:
```
Remi-Gau: moved from nilearn._utils.check_estimator

nilearn/_utils/tables.py
```diff
@@ -0,0 +1,98 @@
+"""Utilities for tables."""
```
Remi-Gau: moved from nilearn._utils.glm

Remi-Gau:
@hndgzkn you may want to have a loot at this: I think I had mentioned this import-linter package to you and this could help enforce a certain structure and avoid circular imports

hndgzkn:
> @hndgzkn you may want to have a loot at this: I think I had mentioned this import-linter package to you and this could help enforce a certain structure and avoid circular imports

Yes, I had noted that. But without a general knowledge of the codebase and dependencies, I did not want to get into that. Thanks for opening this. I think this is a good starting point to discuss the architecture.

I have created an image of dependencies using [pydeps](https://github.com/thebjorn/pydeps). I do not know if there is a better tool. If anyone knows, we can try  with another tool as well.

With some modification in its code, I've obtained an image of package dependencies. It is not really readable, but it gives an idea where we can improve. 

nilearn

It also provides the dependencies in text format. I've cleaned [the file](https://github.com/user-attachments/files/21404356/deps_clean.json) to leave only nilearn packages.
 
I would suggest to discuss the above layered architecture by going over the packages one by one, removing the unnecessary dependencies, defining the package boundaries and its relations with other packages. How does it sound?

bthirion:
> > @hndgzkn you may want to have a loot at this: I think I had mentioned this import-linter package to you and this could help enforce a certain structure and avoid circular imports
> 
> Yes, I had noted that. But without a general knowledge of the codebase and dependencies, I did not want to get into that. Thanks for opening this. I think this is a good starting point to discuss the architecture.
> 
> I have created an image of dependencies using [pydeps](https://github.com/thebjorn/pydeps). I do not know if there is a better tool. If anyone knows, we can try with another tool as well.
> 
> With some modification in its code, I've obtained an image of package dependencies. It is not really readable, but it gives an idea where we can improve.
> 
> nilearn
> 
> It also provides the dependencies in text format. I've cleaned [the file](https://github.com/user-attachments/files/21404356/deps_clean.json) to leave only nilearn packages.
> 
> I would suggest to discuss the above layered architecture by going over the packages one by one, removing the unnecessary dependencies, defining the package boundaries and its relations with other packages. How does it sound?

Thx for having this discussion. still, I am wondering what problem we intent to solve here ? AFAIK we do not have major issues with dependencies ?

hndgzkn:
> Thx for having this discussion. still, I am wondering what problem we intent to solve here ? AFAIK we do not have major issues with dependencies ?

Currently there are many cyclic imports in the code. We solve it by doing cyclic imports from inside the functions. 
The boundaries and responsabilities of packages are either not clear or not respected. This increases the complexity of code.   

In practice it seems like we do not have a problem, as we find hacks to solve them, but actually we have an architectural problem. 

Usage of `is_matplotlib_installed` function is one example of that. `matplotlib` is not a direct dependency of `nilearn`. We tell that it is a dependency of `nilearn.plotting` package. In that case, we should check it only there. But we have these checks in many packages which contaminates code and increases complexity. Depending on the check result, we import functions from inside function. This is not a good practice. 

I think this PR is a good starting point to discuss the eventual architecture and use `import linter` to restrict imports, initially ignoring the existing undesired ones and cleaning them in time as we did for ruff errors. This probably won't require to change public API, or just in a few minor changes. The problems are mainly, as far as I see, in private functions.

Remi-Gau:
I agree with @hndgzkn, we often hack our way around cyclic import.
I think a bit like ruff helps us to format our code, import-linter can help tell us "where to put our code".

I will 'revert' some of the changes made here, to mostly only have the config file changes and put of the commits moving codes in a separate PR for easier reviews.

I agree that we should first see what changes we can most easily do that do not change the API or come with weird unexpected side effects.

hndgzkn:
> I will 'revert' some of the changes made here, to mostly only have the config file changes and put of the commits moving codes in a separate PR for easier reviews.

Yes, I think it is a good idea.

bthirion:
> * Closes none
> 
> 
> Changes proposed in this pull request:
> 
>     * use [import-linter](https://import-linter.readthedocs.io/en/stable/index.html) to define /constrain the general architecture of nilearn
> 
> 
> At the moment we more or less have 3 layers:
> 
> top layer: those packages can import from any of the packages below but not between each other bottom layer : those packages can import from any of the packages below, between each other but not from any above
> 
> ```mermaid
> flowchart TD
>     subgraph high
>         direction LR
>         glm <--X--> decoding 
>     end
>     subgraph low
>         direction LR 
>         maskers
>     end
>     high --> low
> ```

So, there are two layers ?

bthirion:
So, IIUC this PR creates a new automatic check of a branch, providing a sanity check of dependency structure ?

Remi-Gau:
> So, IIUC this PR creates a new automatic check of a branch, providing a sanity check of dependency structure ?

yup

Remi-Gau:
> So, there are two layers ?

So far yes: there should be more, but I'd rather start with the most simple architecture, fix all the exception that fail to match this architecture, then add more modules into a layer or more layers (with some exceptions) and fix those new exceptions...

Ideally we want to make small iterations to clean up the architecture a bit (within the constraints of not breaking the API)

bthirion:
> > So, there are two layers ?
> 
> So far yes: there should be more, but I'd rather start with the most simple architecture, fix all the exception that fail to match this architecture, then add more modules into a layer or more layers (with some exceptions) and fix those new exceptions...
> 
> Ideally we want to make small iterations to clean up the architecture a bit (within the constraints of not breaking the API)

Not super enthusiastic about seeing more layers: this could add a lot of complexity in the code structure...

bthirion (approved):
LGTM, thx.

Remi-Gau:
> Not super enthusiastic about seeing more layers: this could add a lot of complexity in the code structure...

well as it is the codebase probably HAS more than 2 layers, and they are probably not well isolated: when seeing the picture posted by @hndgzkn I think our current architecture verges on the [lasagna code](https://en.wikipedia.org/wiki/Spaghetti_code#Lasagna_code)

ideally after this we would at least have better separated layers

Remi-Gau:
update figure of our current "architecture" without external dependencies in the picture

```
pydeps nilearn -v --only nilearn --max-module-depth 2 --exclude-exact nilearn.conftest nilearn.tests nilearn.input_data
```

nilearn

Remi-Gau:
@hndgzkn 
do you think we can merge this one?

it's been reduced to just a config change now.

hndgzkn:
> @hndgzkn do you think we can merge this one?
> 
> it's been reduced to just a config change now.

Sorry, the notification did not appear. I just saw it. Are you planning to open another PR to continue discussing?

Remi-Gau:
there is a follow up PR: https://github.com/nilearn/nilearn/pull/5547

PR #5543, opened 2025-07-22
[MAINT] PEP 639 compliance

DimitriPapadopoulos:
- Declare licenses using only these two fields, as per PEP 639:
  * `license`: SPDX license expression consisting of one or more license identifiers
  * `license-files`: list of license file glob patterns
  
  Supported by hatchling ≥ 1.27.0:
  https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,6 @@ authors = [{name = "Nilearn developers"}]
 classifiers = [
     "Intended Audience :: Science/Research",
     "Intended Audience :: Developers",
-    "License :: OSI Approved",
     "Programming Language :: C",
     "Programming Language :: Python",
     "Topic :: Software Development",
@@ -36,7 +35,8 @@ dependencies = [
 description = "Statistical learning for neuroimaging in Python"
 # Version from setuptools_scm
 dynamic = ["version"]
-license = {text = "new BSD"}
+license = "BSD-3-Clause"
+license-files = ["LICENSE"]
 maintainers = [{name = "Bertrand Thirion", email = "bertrand.thirion@inria.fr"}]
 name = "nilearn"
 readme = "README.rst"
--- a/LICENSE
+++ b/LICENSE
@@ -1,34 +1,26 @@
-License for nilearn
-===================
-
-New BSD License
-
 Copyright (c) The nilearn developers.
 All rights reserved.
 
-
 Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
+modification, are permitted provided that the following conditions
+are met:
+1. Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+   notice, this list of conditions and the following disclaimer in the
+   documentation and/or other materials provided with the distribution.
+3. Neither the name of the University nor the names of its contributors
+   may be used to endorse or promote products derived from this software
+   without specific prior written permission.
 
-  a. Redistributions of source code must retain the above copyright notice,
-     this list of conditions and the following disclaimer.
-  b. Redistributions in binary form must reproduce the above copyright
-     notice, this list of conditions and the following disclaimer in the
-     documentation and/or other materials provided with the distribution.
-  c. Neither the name of the nilearn developers nor the names of
-     its contributors may be used to endorse or promote products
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -1,13 +1,12 @@
...
-requires = ["hatchling", "hatch-vcs"]
+requires = ["hatchling>=1.27", "hatch-vcs"]
 
 [project]
 authors = [{name = "Nilearn developers"}]
 classifiers = [
     "Intended Audience :: Science/Research",
     "Intended Audience :: Developers",
-    "License :: OSI Approved",
     "Programming Language :: C",
```
htwangtw, line 10: The updated license description actually doesn't match with our license

DimitriPapadopoulos: How doesn't it match?

DimitriPapadopoulos: According to the [SPDX License List](https://spdx.org/licenses/):
[BSD 3-Clause "New" or "Revised" License](https://spdx.org/licenses/BSD-3-Clause.html) ↔ `BSD-3-Clause`

htwangtw: I see! Thank you for digging this out.

Does this mean we need to update the title of our license file?

DimitriPapadopoulos: I don't think so, `LICENSE` is fine and pretty standard. There is no default licence file name that wouldn't require an explicit `license-files`.

DimitriPapadopoulos: Oh, you meant the title inside the file?
https://github.com/nilearn/nilearn/blob/8789531e96b88a9e89c84e10b853410218d452a0/LICENSE#L4

_New BSD License_ seems OK to me. Would you like to change to _BSD 3-Clause "New" or "Revised" License_? Actually, I would remove the title altogether. I don't think a title is required, there's no such title here:
https://spdx.org/licenses/BSD-3-Clause.html

DimitriPapadopoulos: I would copy the licence verbatim from a BSD source file, for example [`sys/net/radix.h`](https://cgit.freebsd.org/src/tree/sys/net/radix.h) from FreeBSD:
```
Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions
are met:
1. Redistributions of source code must retain the above copyright
   notice, this list of conditions and the following disclaimer.
2. Redistributions in binary form must reproduce the above copyright
   notice, this list of conditions and the following disclaimer in the
   documentation and/or other materials provided with the distribution.
3. Neither the name of the University nor the names of its contributors
   may be used to endorse or promote products derived from this software
   without specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE REGENTS AND CONTRIBUTORS ``AS IS'' AND
ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
ARE DISCLAIMED.  IN NO EVENT SHALL THE REGENTS OR CONTRIBUTORS BE LIABLE
FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
SUCH DAMAGE.
```

Remi-Gau: @DimitriPapadopoulos can I let you copy the license text as part of this PR?

DimitriPapadopoulos: Done.

I've removed the `License for nilearn` title - it is kind of obvious - and the `New BSD License` title as well.

Remi-Gau: thanks

Remi-Gau (approved):
good with me

PR #5547, opened 2025-08-04
[MAINT] fix some circular import issues

Remi-Gau:
- Closes none

- PR stacked on (and blocked by) #5538
- moves some function to avoid some architectural issues
  - replace `make_stat_maps` function with a `_make_stat_maps` method for FirstLevelModel and SecondLevelModel 
  - move tag related function from `nilearn._utils.estimator_checks` to `nilearn._utils.tags`
  - move `create_cosine_drift` to `nilearn.signal`
  - move save_figure_if_needed

### DIFF
--- a/nilearn/plotting/_utils.py
+++ b/nilearn/plotting/_utils.py
@@ -17,24 +17,46 @@ def engine_warning(engine):
     warn(message, stacklevel=find_stack_level())
 
 
-def get_cbar_ticks(vmin, vmax, offset, n_ticks=5):
-    """Help for BaseSlicer."""
+def get_cbar_ticks(vmin, vmax, threshold=None, n_ticks=5):
+    """Return an array of evenly spaced ``n_ticks`` tick values to be used for
+    the colorbar.
+
+    Parameters
+    ----------
+    vmin: :obj:`float`
+        minimum value for the colorbar
+    vmax: :obj:`float`
+        maximum value for the colorbar
+    threshold: :obj:`float`, :obj:`int` or None
+        if threshold is not None, ``-threshold`` and ``threshold`` values are
+    replaced with the closest tick values
+    n_ticks: :obj:`int`
+        number of tick values to return
+
+    Returns
+    -------
+    :class:`~numpy.ndarray`
+        an array with ``n_ticks`` elements if ``vmin`` != ``vmax``, else array
+        with one element.
+    """
     # edge case where the data has a single value yields
     # a cryptic matplotlib error message when trying to plot the color bar
     if vmin == vmax:
         return np.linspace(vmin, vmax, 1)
 
     # edge case where the data has all negative values but vmax is exactly 0
+    vmax_temp = vmax
     if vmax == 0:
-        vmax += np.finfo(np.float32).eps
+        vmax_temp = np.finfo(np.float32).eps
+
+    ticks = np.linspace(vmin, vmax, n_ticks)
 
     # If a threshold is specified, we want two of the tick
     # to correspond to -threshold and +threshold on the colorbar.
     # If the threshold is very small compared to vmax,
     # we use a simple linspace as the result would be very difficult to see.
-    ticks = np.linspace(vmin, vmax, n_ticks)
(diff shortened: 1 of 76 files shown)


Remi-Gau:
For :

```
pydeps nilearn -v --only nilearn --max-module-depth 2 --exclude-exact nilearn.conftest nilearn.tests nilearn.input_data
```

## Before

nilearn


## After

nilearn_after

after.jpg
Remi-Gau, line 1: remove the figures

nilearn/image/resampling.py
```diff
@@ -451,8 +451,6 @@ def resample_img(
     homogeneous.
 
     """
-    from .image import new_img_like  # avoid circular imports
```
Remi-Gau, line 454: can be safely hoisted

nilearn/plotting/_utils.py
```diff
@@ -19,42 +17,6 @@ def engine_warning(engine):
     warn(message, stacklevel=find_stack_level())
 
 
-def save_figure_if_needed(fig, output_file):
```
Remi-Gau, line 22: moved to _sliers.py

bthirion (commented):
LGTM.

Remi-Gau:
@hndgzkn 
this should fix a few circular import: can you give me a review when you have time?

pyproject.toml
```diff
@@ -127,18 +127,32 @@ root_package = "nilearn"
...
+    "nilearn.plotting.matrix.* -> nilearn.glm.**"
 ]
+# 1. glm, decoding, connectome, decomposition:
+# - cannot import from each other
+# - but can import from any lower layers
+# 2. mass_univariate:
+# - can import from any lower layers
+# 3. maskers, image, surface, plotting, masking:
+# - can import from each other
+# - can import from any liwer layers
```
hndgzkn:
```suggestion
# - can import from any lower layers
```

nilearn/plotting/_utils.py
hndgzkn: I guess you moved this to avoid importing `BaseSlicer`.

I would suggest to keep it here, as `image`, `surface`, `displays`, `matrix` subpackages of `nilearn.plotting` import this function. 

A solution might be to provide the convenient input where it is called.

nilearn/_utils/glm.py
```diff
@@ -134,59 +136,48 @@ def coerce_to_dict(input_arg):
     return input_arg
 
 
-def make_stat_maps(
-    model, contrasts, output_type="z_score", first_level_contrast=None
-):
-    """Given a model and contrasts, return the corresponding z-maps.
+def create_cosine_drift(high_pass, frame_times):
```
hndgzkn: Just as an idea: Could it be relevant to put this in `nilearn.signal`, and allow a dependency from `glm` to `signal` (For the defined architecture, it is already allowed)? As far as I understand, this is not something that would be used across many packages.

hndgzkn (approved):
Some suggestions, otherwise LGTM.

nilearn/_utils/glm.py
Remi-Gau: let me double check but I think that it makes sense indeed

nilearn/plotting/_utils.py
Remi-Gau: just to be clear: do you suggest I mostly revert commit de06f1abb1d56e41b19284b0c4faf03d4a61691f

hndgzkn: Yes. And instead call `save_figure_if_needed` with `fig.figure` where necessary. Then we won't need to import  `BaseSlicer` if I am not mistaken.

nilearn/_utils/glm.py
Remi-Gau: done

nilearn/plotting/_utils.py
Remi-Gau: there is no easy way for me to find where to do that without some type annotations so I think I will leave it here for now

PR #5551, opened 2025-08-04
[FIX] fix handling of regions names and indices by label maskers

### LINKED ISSUE #5542: [BUG] inconsistency between `region_ids_` and `region_names_` generated by `NiftiLabelsMasker` when not passing `labels`

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Linux Ubuntu 24.10

### Python version

- [ ] 3.13
- [ ] 3.12
- [ ] 3.11
- [x] 3.10
- [ ] 3.9

### nilearn version

`0.12.0`

### Expected behavior

When not providing labels, the masker will autogenerate `region_ids_` and `region_names_`, and I would expect:
```
>> masker.region_ids_
{0: 0.0, 1: 1.0, 2: 2.0}

>> masker.region_names_
{1: '1.0', 2: '2.0'}
```

But this is what I got:

```
>> masker.region_ids_
{0: 0.0, 1: 1.0, 2: 2.0}

>> masker.region_names_
{1: '0.0', 2: '1.0'}
```

```python

import numpy as np

# create a label image with two regions
atlas = np.zeros((8, 8, 8))
atlas[4, 4, 3:5] = 1
atlas[4, 4, 5:7] = 2
atlas = Nifti1Image(atlas, np.eye(4))

# fit the masker
masker = NiftiLabelsMasker(labels_img=atlas).fit()
masker_with_labels = NiftiLabelsMasker(labels_img=atlas, labels=['1', '2']).fit()
```

### LINKED ISSUE #5544: [BUG] NiftiLabelsMasker's region_ids_ and region_names_ keys do not match signals array

### Operating system

- [ ] Linux
- [x] Mac
- [ ] Windows

### Operating system version

- Mac OS Version 15.5 "Sequoia"

### Python version

- [ ] 3.13
- [ ] 3.12
- [ ] 3.11
- [ ] 3.10
- [x] 3.9

### nilearn version

0.12.0, but replicated on 0.12.1.dev31+g8789531e9

### Expected behavior

Based on the NiftiLabelsMasker docstring, I would expect `masker.region_ids_[0]` to equal the value of the first parcel in the atlas image (3).

`masker.region_ids_.keys()` and `masker.region_names_.keys()` are both equal to [2, 3, 4] instead of [0, 1, 2]. I think this is because the keys come from the index (not the 'index' column, just the index) of the `lut` after removing rows for parcels that don't exist in the atlas image. I think just resetting the index of the `lut` DataFrame after removing missing parcels would fix the issue, but I could be wrong.

```python
import nibabel as nb
import numpy as np
import pandas as pd

# Create an atlas that is missing the first two regions due to resampling
arr = np.zeros((3, 3, 3))
arr[0, 0, 0] = 3
arr[0, 0, 1] = 4
arr[0, 0, 2] = 5
atlas_img = nb.Nifti1Image(arr.astype(np.int32), np.eye(4))
lut = pd.DataFrame(
    columns=['index', 'name'],
    data=[[1, 'Region A'], [2, 'Region B'], [3, 'Region C'], [4, 'Region D'], [5, 'Region E']],
)

masker = NiftiLabelsMasker(atlas_img, lut=lut)
signals = masker.fit_transform(atlas_img)
# I would expect keys of these dicts to be 0, 1, 2 (index into signals array)
print(masker.region_ids_)
print(masker.region_names_)
print(signals.shape)
```
This outputs:

```
{2: 3, 3: 4, 4: 5}
{2: 'Region C', 3: 'Region D', 4: 'Region E'}
(3,)
```

Remi-Gau:
- Closes #5542
- Closes #5544

- [x] fix handling of regions names and indices
- [x] update changelog
- [x] fix similar issues for surface label masker
- [x] make sure any old behavior (before LUT was introduced) with too many / few labels are conserved


nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -695,6 +695,94 @@ def test_region_names(
...
 
 
+def test_regions_id_names_no_labels_no_lut(affine_eye):
+    """Check match between id and names.
+
+    Regression test for https://github.com/nilearn/nilearn/issues/5542.
+    """
+    atlas = np.zeros((8, 8, 8))
+    atlas[4, 4, 3:5] = 1
+    atlas[4, 4, 5:7] = 2
```
tsalo: One element that I think is important in both the tests and documentation is how non-consecutive values are treated/expected to be treated.

For example, with an atlas with values 3, 4, 5 and five labels, I expected the masker to use the third, fourth, and fifth labels (i.e., assuming that the values 1 and 2 were dropped due to resampling or something). @dangom expected the masker to use the first, second, and third labels instead.

@dangom's expectation makes more sense in a lot of ways, since many atlases don't have consecutive values and can go up to very high values, and you wouldn't want to have to fill out your list of labels with hundreds of Nones. On the other hand, that requires more work from users if they have an initial list of labels, but have to do resampling/warping prior to applying the mask that might eliminate some of the parcels.

Remi-Gau: if the user is expecting a certain mapping between values and names (or missing values) then they should pass an lut: way easier and less ambiguous

in theory the masker should warn you when some labels go missing because of resampling

bthirion (approved):

nilearn/maskers/tests/test_nifti_labels_masker.py
tsalo: I agree that using an lut is better, but I think the documentation and tests should still make it clear what will happen if you use labels.

nilearn/_utils/bids.py
```diff
@@ -12,7 +12,7 @@
 
 
 def generate_atlas_look_up_table(
-    function=None, name=None, index=None, strict=False
+    function=None, name=None, index=None, strict=False, background_label=None
```
hndgzkn, line 10: Doctring for `background_label` could be added.

hndgzkn (approved):
A part from a minor comment, seems Ok to me.

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -580,6 +582,304 @@ def test_nifti_labels_masker_with_mask(
     assert np.allclose(get_data(masker.region_atlas_), masked_labels_data)
 
 
+def generate_labels(n_regions, background=""):
+    """Create list of strings to use as labels."""
+    labels = []
+    if background:
+        labels.append(background)
+    labels.extend([f"region_{i + 1!s}" for i in range(n_regions)])
+    return labels
```
Remi-Gau: moved up

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -580,6 +582,304 @@ def test_nifti_labels_masker_with_mask(
...
+
+
+def generate_expected_lut(region_names):
+    """Generate a look up table based on a list of regions names."""
+    if "background" in region_names:
+        idx = region_names.index("background")
+        region_names[idx] = "Background"
+    return pd.DataFrame(
+        {"name": region_names, "index": list(range(len(region_names)))}
+    )
```
Remi-Gau: moved up

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -580,6 +582,304 @@ def test_nifti_labels_masker_with_mask(
...
+
+
+def check_region_names_after_fit(
```
Remi-Gau, line 605: moved up

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -580,6 +582,304 @@ def test_nifti_labels_masker_with_mask(
...
+        ]
+        region_names_after_fit.sort()
+        region_names.sort()
+        if background:
+            region_names = deepcopy(region_names)
+            region_names.pop(region_names.index(background))
+        assert region_names_after_fit == region_names
+
+
+def check_lut(masker, expected_lut):
```
Remi-Gau: moved up

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -580,6 +582,304 @@ def test_nifti_labels_masker_with_mask(
...
+    assert "Background" in masker.lut_["name"].to_list()
+    pd.testing.assert_series_equal(
+        masker.lut_["index"], expected_lut["index"], check_dtype=False
+    )
+    pd.testing.assert_series_equal(
+        masker.lut_["name"], expected_lut["name"], check_dtype=False
+    )
+
+
+def check_region_names_ids_match_after_fit(
```
Remi-Gau, line 674: moved up

nilearn/maskers/tests/test_nifti_labels_masker.py
Remi-Gau: there were already a few tests checking behavior of passing labels:
- too many labels
- non consecutive indices / labels
Added one for too few labels.

Made sure we have the same when passing LUT and that the behavior when passing LUT or labels was the same.

Tried to update the doc strings for labels and lut too.

Remi-Gau: the diff got much bigger in part because I moved some utility function in a test file

tsalo:
I checked out your branch locally and tested my code from #5544. The results are slightly different, but not what I expected.

The code mocks up an atlas with values of 0, 3, 4, and 5, along with an lut that covers values of 1 through 5 (Regions A through E).

The old results:
```
masker.region_ids_: {2: 3, 3: 4, 4: 5}
masker.region_names_: {2: 'Region C', 3: 'Region D', 4: 'Region E'}
signals.shape: (3,)
```

The new results:
```
masker.region_ids_: {0: 0, 3: 3, 4: 4, 5: 5}
masker.region_names_: {3: 'Region C', 4: 'Region D', 5: 'Region E'}
signals.shape: (3,)
```

What I expected:
```
masker.region_ids_: {0: 3, 1: 4, 2: 5}
masker.region_names_: {0: 'Region C', 1: 'Region D', 2: 'Region E'}
signals.shape: (3,)
```

The key thing here is that I would expect the keys from `masker.region_ids_` and `masker.region_names_` to be indices into the `signals` array, but they're instead the values of the parcels in the atlas. If my expectation is wrong then that's fine, but the descriptions of `region_ids_` and `region_names_` in the docstring seem to imply that it should be (assuming `region_signal` refers to the `signals` array returned by `masker.transform()`).

```
region_ids_
    Return dictionary containing the region ids corresponding
    to each column in the array returned by `transform`.

    The region id corresponding to ``region_signal[:,i]`` is ``region_ids_[i]``.
    ``region_ids_['background']`` is the background label.

region_names_
    Return a dictionary containing the region names corresponding
    to each column in the array returned by `transform`.

    The region names correspond to the labels provided in labels in input.
    The region name corresponding to ``region_signal[:,i]`` is ``region_names_[i]``.
```

Remi-Gau:
> The key thing here is that I would expect the keys from `masker.region_ids_` and `masker.region_names_` to be indices into the `signals` array, but they're instead the values of the parcels in the atlas. If my expectation is wrong then that's fine, but the descriptions of `region_ids_` and `region_names_` in the docstring seem to imply that it should be (assuming `region_signal` refers to the `signals` array returned by `masker.transform()`).
> 
> ```
> region_ids_
>     Return dictionary containing the region ids corresponding
>     to each column in the array returned by `transform`.
> 
>     The region id corresponding to ``region_signal[:,i]`` is ``region_ids_[i]``.
>     ``region_ids_['background']`` is the background label.
> 
> region_names_
>     Return a dictionary containing the region names corresponding
>     to each column in the array returned by `transform`.
> 
>     The region names correspond to the labels provided in labels in input.
>     The region name corresponding to ``region_signal[:,i]`` is ``region_names_[i]``.
> ```

oops indeed
let me fix that

Remi-Gau:
@tsalo it should be fixed now

tsalo:
The new results look good to me, thanks!

Remi-Gau:
I should probably check that the surface label masker behaves properly for this too

Remi-Gau:
ok similar issues in the surface label masker: fixing it there too.

nilearn/maskers/surface_labels_masker.py
```diff
@@ -340,13 +359,45 @@ def fit(self, imgs=None, y=None):
...
             )
         else:
             lut = generate_atlas_look_up_table(
-                function=None, index=self.labels_img_
+                function=None,
+                index=self.labels_img_,
+                background_label=add_background_to_lut,
+            )
+
+        # passed labels or lut may not include background label
```
Remi-Gau: a lot of the lut handling looks simmilar to what is done in the _generate_lut of the NiftiLabelMasker: see if possible to extract that into a separate function...

nilearn/maskers/surface_labels_masker.py
```diff
@@ -459,6 +510,7 @@ def transform_single_imgs(self, imgs, confounds=None, sample_mask=None):
         index = self.labels_
         if self.background_label in index:
             index.pop(index.index(self.background_label))
+        print(index)
```
Remi-Gau:
```suggestion
```

Remi-Gau:
@dangom can you check the changes I propose here and maybe checkout my branch, to make sure this will fix your issues?

dangom:
@Remi-Gau thanks so much!
It seems the example with the labels still works as before, but it is inconsistent with the lut:

```python
import nibabel as nb
import numpy as np
import pandas as pd

arr = np.zeros((3, 3, 3))
arr[0, 0, 0] = 3
arr[0, 0, 1] = 4
arr[0, 0, 2] = 5
atlas_img = nb.Nifti1Image(arr.astype(np.int32), np.eye(4))
labels = ['Region A', 'Region B', 'Region C', 'Region D', 'Region E']

masker = NiftiLabelsMasker(atlas_img, labels=labels)
signals = masker.fit_transform(atlas_img)

print(masker.region_ids_)
print(masker.region_names_)
print(signals.shape)
```

```
/sklearn/utils/_set_output.py:316: UserWarning: Too many names for the indices. Dropping excess names values.
  data_to_wrap = f(self, X, *args, **kwargs)
{'background': 0, 0: 3, 1: 4, 2: 5}
{0: 'Region A', 1: 'Region B', 2: 'Region C'}
(3,)
```

It does lead to a warning from sklearn, since it's arguibly undefined behavior. But perhaps it should err and ask the user to use the lut instead for clarity if these is a mistatch?  What to do with a list of labels if there are more entries than there are values?

Otherwise looks great! I hadn't used the lut behaviour before, but I'm happy to personally switch to that for robustness.

Remi-Gau:
> It does lead to a warning from sklearn, since it's arguibly undefined behavior. 

actually that warning comes from nilearn but it shows as coming from sklearn because our warnings show the first non-nilearn piece of code in the call stack that throws them: so if the stack is 

`nilearn function A  --calls--> sklearn function A  --calls--> nilearn function B with warning`

then the warning will report which sklearn function "threw the warning"

> But perhaps it should err and ask the user to use the lut instead for clarity if these is a mistatch? What to do with a list of labels if there are more entries than there are values?

currently the extra values are just dropped because there is no way to know how label names are supposed to map to values in the label image: that's why the LUT is better as it removes this ambiguity.

But that raises a good point: I should check what was the behavior of passing too many label names before the whole LUT thing was introduced and make sure that this behavior was not changed.

Remi-Gau:
OK when running this code on nilearn 0.11 

```python
from nilearn._utils.data_gen import generate_fake_fmri, generate_labeled_regions, generate_maps

img = generate_fake_fmri()[0]

label_img = generate_labeled_regions(shape=(10,11,12), n_regions=3)

# too many labels
masker = NiftiLabelsMasker(label_img, labels=["A", "B", "C", "E", "F"]).fit()

signals = masker.transform(img)
print(masker.region_ids_)
print(masker.region_names_)
print(signals.shape)
try:    
    masker.generate_report()
except ValueError as e:
    print(f"ERROR : {str(e)}")

# too few labels
masker = NiftiLabelsMasker(label_img, labels=["A", "B"]).fit()

signals = masker.transform(img)
print(masker.region_ids_)
print(masker.region_names_)
print(signals.shape)
try:    
    masker.generate_report()
except ValueError as e:
    print(f"ERROR : {str(e)}")

```

I get:

```
/home/remi-gau/github/nilearn/nilearn/tmp/masker_dtype.py:9: UserWarning: Mismatch between the number of provided labels (5) and the number of regions in provided label image (4).
  masker = NiftiLabelsMasker(label_img, labels=["A", "B", "C", "E", "F"]).fit()
/home/remi-gau/github/nilearn/nilearn/tmp/masker_dtype.py:9: UserWarning: Number of regions in the labels image does not match the number of labels provided.
  masker = NiftiLabelsMasker(label_img, labels=["A", "B", "C", "E", "F"]).fit()
/home/remi-gau/github/nilearn/nilearn/nilearn/maskers/base_masker.py:282: UserWarning: Mismatch between the number of provided labels (5) and the number of regions in provided label image (4).
Note that this may be due to some regions being dropped from the label image after resampling.
  return self.transform_single_imgs(
{'background': 0, 0: np.int32(1), 1: np.int32(2), 2: np.int32(3)}
{0: 'A', 1: 'B', 2: 'C'}
(17, 3)
ERROR : Mismatch between the number of provided labels (5) and the number of regions in provided label image (4).

/home/remi-gau/github/nilearn/nilearn/tmp/masker_dtype.py:21: UserWarning: Mismatch between the number of provided labels (2) and the number of regions in provided label image (4).
  masker = NiftiLabelsMasker(label_img, labels=["A", "B"]).fit()
/home/remi-gau/github/nilearn/nilearn/tmp/masker_dtype.py:21: UserWarning: Number of regions in the labels image does not match the number of labels provided.
  masker = NiftiLabelsMasker(label_img, labels=["A", "B"]).fit()
/home/remi-gau/github/nilearn/nilearn/nilearn/maskers/base_masker.py:282: UserWarning: Mismatch between the number of provided labels (2) and the number of regions in provided label image (4).
Note that this may be due to some regions being dropped from the label image after resampling.
  return self.transform_single_imgs(
{'background': 0, 0: np.int32(1), 1: np.int32(2), 2: np.int32(3)}
None
(17, 3)
ERROR : Mismatch between the number of provided labels (2) and the number of regions in provided label image (4).
```

So we have not thrown errors for mismatch between the number of labels and the number of regions in the image, we only throw warnings... except when building a report.

When there are too many labels, extra labels were dropped the same way we do it now.

When there are too few labels, `masker.region_names_` was just `None` (that's a bug I was not even aware this old version had).

OK so I think that we have parity of behavior: no error during fit or transform, only warnings.

nilearn/maskers/base_masker.py
```diff
@@ -763,3 +770,67 @@ def set_output(self, *, transform=None):
             This has not been implemented yet.
         """
         raise NotImplementedError()
+
+
+def generate_lut(labels_img, background_label, lut=None, labels=None):
```
Remi-Gau, line 773: extracted from niftilabelmasker to be used in both niftilabelmasker and surfacelabelmasker

Remi-Gau:
ok apparently this broke the use of LUT with some of our atlases: this was missed because those are tested only on main to avoid downloading all atlases in PR.

https://github.com/nilearn/nilearn/actions/runs/16943410127/job/48018208963

this lead to some examples failing

Issue #5318, opened 2025-04-04
consider "caching" compute_contrast results in GLM models


Remi-Gau:
I doubt I am the first one to think of it, so is there any reason why results from compute_contrast have to be recomputed every time the method is called rather than keeping track of contrasts that have already been computed and we re accessing them?

@bthirion @emdupre

emdupre:
Sorry, I'm less familiar with this part of the code base, so just to make sure :
Are you referring [to this issue](https://github.com/nilearn/nilearn/issues/2406#issuecomment-631632767), where `make_stat_maps` is repeatedly calling `compute_contrast` ? Or do you mean within `compute_contrast` itself things are being called repeatedly ?

bthirion:
IIUC in many practical cases, contrasts are computed several times, at least twice, for reporting and to write them on disk. Indeed caching would help here.

PR #5554, opened 2025-08-05
[FIX] cache some GLM results

Remi-Gau:
- Relates to #5318 
- Relates to #4663

- trying a few things to see if we can speed up some GLM related things (save GLM to BIDS and reporting) with a bit of caching

### DIFF
--- a/examples/04_glm_first_level/plot_adhd_dmn.py
+++ b/examples/04_glm_first_level/plot_adhd_dmn.py
@@ -92,7 +92,15 @@ contrasts = {"seed_based_glm": dmn_contrast}
 # Perform first level analysis
 # ----------------------------
 # Setup and fit GLM.
-first_level_model = FirstLevelModel()
+#
+# We cache some results on disk with 'memory= output_dir / "tmp"'
+# to avoid recomputing things when generating a GLM report at the end.
+from pathlib import Path
+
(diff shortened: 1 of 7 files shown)


nilearn/glm/io.py
```diff
@@ -228,6 +231,15 @@ def save_glm_to_bids(
...
             report_kwargs[key] = kwargs[key]
 
+    # force to do some caching as some computations
+    # are done twice: once during saving the GLM
+    # and then during report creation
+    # called at the end of this function
+    original_location = deepcopy(model.memory.location)
+    if original_location is None:
+        temp_dir = mkdtemp()
+        model.memory = check_memory(temp_dir)
```
Remi-Gau, line 350: Note: this is not in a context manager, so may lead to some extra things left in a temporary directory if the function crashes in the middle

nilearn/glm/io.py
Remi-Gau, line 1: the only indirect benchmark I have for this is the duration of the test which is just a poor proxy, could already help with some slow tests

Remi-Gau:
I am mostly trying things here to improve performance a bit.
Not sure it will make a great difference.

bthirion:
Have you been able to measure the performance difference ?

Remi-Gau:
> Have you been able to measure the performance difference ?

I have not take the time to try yet. 🙈

PR #5556, opened 2025-08-05
[DOC] improve doc and test around screening percentile adjustments

### LINKED ISSUE #5555: Better doc for screening_percentile
related to
https://neurostars.org/t/feature-selection-does-not-appear-to-work/33715

using screening percentile in combination with mask will often not make a difference because all the elements of the ROI (voxels or vertices) will be included anyway

I have not found this mentioned anywhere in the doc or the user guide and it should probably be 

plotted this get an idea of what to expect

from https://neurostars.org/t/feature-selection-does-not-appear-to-work/33715/7?u=remi-gau

Remi-Gau:
- Closes #5555

- improve doc for screening percentile adjustments

### DIFF
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -882,6 +882,24 @@ sample_mask : :obj:`list` of sample_mask, default=None
     This parameter is passed to :func:`nilearn.signal.clean`.
 """
 
+docdict["screening_percentile"] = """
+screening_percentile : int, float, \
+                       in the closed interval [0, 100], or None, \
+                       default=20
+        Percentile value for ANOVA univariate feature selection.
+        A value of ``100`` means "keep all features".
+        This percentile is expressed
+        with respect to the volume of either a standard (MNI152) brain
+        (if mask_img is a 3D volume)
+        or a the number of vertices in the standard brain mesh
+        (if mask_img is a SurfaceImage).
+        This means that the
+        ``screening_percentile`` is corrected at runtime by premultiplying it
+        with the ratio of volume of the
+        standard brain to the volume of the mask of the data.
+        If ``None`` is passed, it will be set to ``100``.
+"""
+
 # second_level_contrast
 docdict["second_level_contrast"] = """
 second_level_contrast : :obj:`str` or :class:`numpy.ndarray` of shape\
@@ -1325,8 +1343,13 @@ docdict["base_decoder_fit_attributes"] = """
             Classes to predict. For classification only.
 
         screening_percentile_ : :obj:`float`
-            Screening percentile corrected according to volume of mask,
-            relative to the volume of standard brain.
+            Percentile value for ANOVA univariate feature selection.
+            A value of 100 means 'keep all features'.
+            This percentile is expressed
+            with respect to the volume of either a standard (MNI152) brain
+            (if mask_img is a 3D volume)
+            or a the number of vertices in the standard brain mesh
+            (if mask_img is a SurfaceImage).
 
         coef_ : numpy.ndarray, shape=(n_classes, n_features)
             Contains the mean of the models weight vector across
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -891,7 +891,7 @@ screening_percentile : int, float, \
         This percentile is expressed
         with respect to the volume of either a standard (MNI152) brain
         (if mask_img is a 3D volume)
-        or a the number of vertices in the standard brain mesh
+        or a the number of vertices in the mask mesh
         (if mask_img is a SurfaceImage).
         This means that the
         ``screening_percentile`` is corrected at runtime by premultiplying it
@@ -1348,7 +1348,7 @@ docdict["base_decoder_fit_attributes"] = """
             This percentile is expressed
             with respect to the volume of either a standard (MNI152) brain
             (if mask_img is a 3D volume)
-            or a the number of vertices in the standard brain mesh
+            or a the number of vertices in the mask mesh
             (if mask_img is a SurfaceImage).
 
         coef_ : numpy.ndarray, shape=(n_classes, n_features)
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -887,17 +887,25 @@ screening_percentile : int, float, \
                        in the closed interval [0, 100], or None, \
                        default=20
         Percentile value for ANOVA univariate feature selection.
+        If ``None`` is passed, it will be set to ``100``.
         A value of ``100`` means "keep all features".
         This percentile is expressed
         with respect to the volume of either a standard (MNI152) brain
-        (if mask_img is a 3D volume)
+        (if mask_img_ is a 3D volume)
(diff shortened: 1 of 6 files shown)


nilearn/_utils/docs.py
```diff
@@ -882,6 +882,32 @@ def custom_function(vertices):
     This parameter is passed to :func:`nilearn.signal.clean`.
 """
 
+docdict["screening_percentile"] = """
+screening_percentile : int, float, \
+                       in the closed interval [0, 100], or None, \
+                       default=20
```
Remi-Gau, line 888: note that some function that use this doc string do not have this default but they are all private functions, so I feel the default=20 can stay there anyway.

nilearn/_utils/tests/test_param_validation.py
```diff
@@ -175,44 +173,50 @@ def test_check_threshold_for_error(matrix):
...
+                mask_img,
+                is_classif,
+            )
+    else:
+        select_percentile = check_feature_screening(
+            screening_percentile, mask_img, is_classif
+        )
+        assert isinstance(select_percentile, BaseEstimator)
+        if roi_size == "small":
+            assert select_percentile.percentile == 100
```
Remi-Gau: I wonder if a warning should be thrown in the case where the user passed a screening percentile < 100 and the effective screeining percentile is actually 100.

doc/decoding/decoding_intro.rst
```diff
@@ -368,6 +368,13 @@ argument. To keep the 10% most correlated voxels, just create us this parameter
    :start-after: # on nested cross-validation.
    :end-before: # Visualize the results
 
+.. note::
+
+    If ``screening_percentile`` is used in combination
+    with a mask that is too small
+    relative to the total brain volume,
+    then all the voxels of the mask
+    may be included even for very small ``screening_percentile``.
```
emdupre: "Too small" is hard to interpret here. Could we rephrase to something like :

```suggestion
    Providing a region-of-interest mask may interact with
    the ``screening_percentile`` parameter, particularly
    in cases where the mask extent is small relative to the
    total brain volume. In these cases, there may not be
    enough features in the mask to allow for further
    sub-selection with ``screening_percentile``.
```

nilearn/_utils/param_validation.py
```diff
@@ -9,11 +9,13 @@
...
 import nilearn.typing as nilearn_typing
 from nilearn._utils import logger
+from nilearn._utils.docs import fill_doc
 from nilearn._utils.logger import find_stack_level
 from nilearn._utils.niimg import _get_data
+from nilearn.surface import SurfaceImage
 
 # Volume of a standard (MNI152) brain mask in mm^3
-MNI152_BRAIN_VOLUME = 1827243.0
+MNI152_BRAIN_VOLUME = 1882989.0
```
emdupre, line 18: why this change ?

nilearn/_utils/tests/test_param_validation.py
emdupre: +1, I think passing a warning in those cases makes sense !

nilearn/_utils/tests/test_param_validation.py
```diff
@@ -175,44 +173,50 @@ def test_check_threshold_for_error(matrix):
...
 def test_get_mask_extent():
-    # Test that hard-coded standard mask volume can be corrected computed
-    if Path(mni152_brain_mask).is_file():
-        assert _get_mask_extent(load(mni152_brain_mask)) == MNI152_BRAIN_VOLUME
-    else:
-        warnings.warn(
-            f"Couldn't find {mni152_brain_mask} (for testing)", stacklevel=2
-        )
+    """Test that hard-coded standard mask volume can be corrected computed."""
+    assert _get_mask_extent(load_mni152_brain_mask()) == MNI152_BRAIN_VOLUME
```
Remi-Gau, line 173: this change is to make nilearn more internal consistent: we test get_extend on the mni brain mask we ship and not the one that may be on the machine only if FSL is around (see line 20)

but to pass this require to make a small update to the MNI152_BRAIN_VOLUME value as it slightly different than that of FSL

nilearn/_utils/tests/test_param_validation.py
```diff
@@ -18,6 +15,7 @@
     check_params,
     check_threshold,
 )
+from nilearn.datasets import load_mni152_brain_mask
 
 mni152_brain_mask = (
     "/usr/share/fsl/data/standard/MNI152_T1_1mm_brain_mask.nii.gz"
```
Remi-Gau: remove if the value is never used

nilearn/_utils/param_validation.py
Remi-Gau: ha yeah forgot to explain:
our reference volume was based on the FSL brain mask, to be internally consistent I changed that to the value of the brain mask that nilearn ships.

see also this changed test: https://github.com/nilearn/nilearn/pull/5556/files#r2260241625

I am pretty sure this value is only used to throw some warning relative to mask size, so it should not affect any computed results (will double check)

doc/decoding/decoding_intro.rst
Remi-Gau: that's better indeed

nilearn/_utils/tests/test_param_validation.py
```diff
@@ -175,44 +173,50 @@ def test_check_threshold_for_error(matrix):
...
+
+    if roi_size == "small":
+        mask_img_data[80:-80, 80:-80, 80:-80] = 1
+    else:
+        mask_img_data[40:-40, 40:-40, 40:-40] = 1
+
     mask_img = Nifti1Image(mask_img_data, affine=affine_eye)
 
-    for is_classif in [True, False]:
-        for screening_percentile in [100, None, 20, 101, -1, 10]:
```
Remi-Gau, line 194: loops were replaced by test parametrization

nilearn/_utils/tests/test_param_validation.py
bthirion: Agreed.

nilearn/_utils/param_validation.py
Remi-Gau: yeah it's only used in this file to throw warnings

nilearn/_utils/param_validation.py
```diff
@@ -302,22 +278,36 @@ def check_feature_screening(
...
+        if effective_screening_percentile == 100:
+            warnings.warn(
+                f"screening_percentile set to '100' despite "
+                f"requesting '{screening_percentile=}'. "
+                "\nAll elements in the mask will be included. "
+                "\nThis usually occurs when the mask image "
+                "is too small compared to full brain mask.",
+                category=UserWarning,
+                stacklevel=find_stack_level(),
+            )
```
Remi-Gau, line 306: @emdupre @bthirion 
any suggestion for a better wording for this warning is welcome

emdupre: I think this is reasonable, thank you !

emdupre (approved):
This LGTM, thank you @Remi-Gau  !

PR #5536, opened 2025-07-10
[DOC] document fitted attributes

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/issues/3428#issuecomment-3051938378

Changes proposed in this pull request:

- add a check for our image estimators to make sure their doc string details all the fitted attrributes (those that end in `_`)

PR #5557, opened 2025-08-07
[REF] turn "is_classification" and "clustering_percentile" into private properties

Remi-Gau:
- None but relates to https://github.com/nilearn/nilearn/pull/5536#issuecomment-3164167166

- change "is_classification" and "clustering_percentile" into properties
- is_classification can be inferred either from a private attribute that or from a tag
- "clustering_percentile" is only used for FREM decoders and should only be a public attribute for such classes


nilearn/decoding/decoder.py
```diff
@@ -352,7 +352,7 @@ def _parallel_fit(
     scorer,
     mask_img,
     class_index,
-    clustering_percentile,
+    clustering_percentile=100,
```
Remi-Gau:
```suggestion
    clustering_percentile,
```

nilearn/decoding/decoder.py
```diff
@@ -710,17 +720,28 @@ def fit(self, X, y, groups=None):
...
+            if (
+                hasattr(self, "clustering_percentile")
+                and self._clustering_percentile < 100
+            ):
+                clustering_percentile_msg = " and / or 'clustering_percentile'"
+            warning_msg = (
+                "The decoding model will be trained only "
+                f"on {n_final_features} features. "
+                f"{screening_percentile_msg}{clustering_percentile_msg}."
+            )
```
Remi-Gau, line 735: this mostly tries to refine the message:
- no point in mentioning things to raise above a 100%
- only mention clustering percentile for decoders that make use of it

nilearn/decoding/decoder.py
```diff
@@ -999,7 +1022,7 @@ def _fetch_parallel_fit_outputs(
     def _set_scorer(self):
         if self.scoring is not None:
             self.scorer_ = check_scoring(self.estimator_, self.scoring)
-        elif self.is_classification:
+        elif self.__sklearn_tags__().estimator_type == "classifier":
```
Remi-Gau:
```suggestion
        elif self._is_classification
```

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1221,7 +1223,7 @@ def test_decoder_adjust_screening_greaterthan_mask_surface(
 
 
 @pytest.mark.parametrize("mask", [None, SurfaceMasker()])
-@pytest.mark.parametrize("decoder", [_BaseDecoder, Decoder, DecoderRegressor])
```
Remi-Gau, line 1224: removing _BaseDecoder from some test parametrization: it should be enough to test it through its children classes

Remi-Gau:
something similar needs to be done for spacenet estimators

PR #5520, opened 2025-07-08
[MAINT] reduce warnings in doc build

Remi-Gau:
- Closes #

Changes proposed in this pull request:

-
-

PR #5561, opened 2025-08-08
[MAINT] use dataset metadata for examples instead of hard coded values

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/5520#discussion_r2262056507

- move t_r and slice timing reference to dataset metadata

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -95,13 +95,15 @@ def fetch_haxby(
             Paths to nifti with face-responsive brain regions.
 
         - 'mask_face_little': :obj:`list` of :obj:`str`.
-            Spatially more constrained version of the above.
+            Paths to nifti with face-responsive brain regions.
+            Spatially more constrained version of  ``mask_face``.
 
         - 'mask_house': :obj:`list` of :obj:`str`.
             Paths to nifti with house-responsive brain regions.
 
         - 'mask_house_little': :obj:`list` of :obj:`str`.
-            Spatially more constrained version of the above.
+            Paths to nifti with house-responsive brain regions.
+            Spatially more constrained version of ``mask_house``.
 
         - 't_r': :obj:`float`
             Repetition time of the functional data.
@@ -2866,13 +2868,13 @@ def fetch_spm_auditory(
         Dictionary-like object, the interest attributes are:
 
         - 'anat': :obj:`list` of :obj:`str`.
-            Path to anat image
+            Paths to anat images
 
         - 'func': :obj:`list` of :obj:`str`.
-            Path to functional image
+            Paths to functional images
 
         - 'events': :obj:`list` of :obj:`str`.
-            Path to events.tsv file
+            Paths to events.tsv files
 
         - 'description': :obj:`str`.
             Data description


Remi-Gau:
pretty sure that the ds000030 dataset we use also has a slice_time_ref of 0.5 as it was preprocessed with fmri prep: but I don't think we use it anywhere

emdupre:
I think this is a great addition, thank you @Remi-Gau !

My only question is on `slice_time_ref` ; we seem to include it in only `fetch_localizer_first_level` even though it's also known in the ADHD (and potentially the developmental) datasets. Does it make sense to include it as a dataset property only for this one dataset ? I know that we've tried to standardize the fetchers as possible, so I just wanted to double-check.

Remi-Gau:
> I think this is a great addition, thank you @Remi-Gau !
> 
> My only question is on `slice_time_ref` ; we seem to include it in only `fetch_localizer_first_level` even though it's also known in the ADHD (and potentially the developmental) datasets. Does it make sense to include it as a dataset property only for this one dataset ? I know that we've tried to standardize the fetchers as possible, so I just wanted to double-check.

yeah I need to double check for those datasets: could not easily find (or make decent assumption) about the slice timing ref

bthirion:
Actually, it is almost an arbitrary parameter. In most cases, it should be set to 0.5.

nilearn/datasets/func.py
```diff
@@ -77,22 +77,34 @@ def fetch_haxby(
...
-          Paths to nifti ventral temporal mask file.
+            Paths to nifti ventral temporal mask file.
+
         - 'mask_face': :obj:`list` of :obj:`str`.
-          Paths to nifti with face-responsive brain regions.
+            Paths to nifti with face-responsive brain regions.
+
         - 'mask_face_little': :obj:`list` of :obj:`str`.
-          Spatially more constrained version of the above.
+            Spatially more constrained version of the above.
```
hndgzkn:
```suggestion
            Paths to nifti with face-responsive brain regions.
            Spatially more constrained version of  ``mask_face``.
```

nilearn/datasets/func.py
```diff
@@ -77,22 +77,34 @@ def fetch_haxby(
...
+
         - 'mask_house': :obj:`list` of :obj:`str`.
-          Paths to nifti with house-responsive brain regions.
+            Paths to nifti with house-responsive brain regions.
+
         - 'mask_house_little': :obj:`list` of :obj:`str`.
-          Spatially more constrained version of the above.
+            Spatially more constrained version of the above.
```
hndgzkn:
```suggestion
            Paths to nifti with house-responsive brain regions.
            Spatially more constrained version of ``mask_house``.
```

nilearn/datasets/func.py
```diff
@@ -2828,10 +2864,21 @@ def fetch_spm_auditory(
...
-        - 'description': :obj:`str`. Data description
+
+        - 'anat': :obj:`list` of :obj:`str`.
+            Path to anat image
+
+        - 'func': :obj:`list` of :obj:`str`.
+            Path to functional image
+
+        - 'events': :obj:`list` of :obj:`str`.
+            Path to events.tsv file
```
hndgzkn:
```suggestion
        - 'anat': :obj:`list` of :obj:`str`.
            Paths to anat images

        - 'func': :obj:`list` of :obj:`str`.
            Paths to functional images

        - 'events': :obj:`list` of :obj:`str`.
            Paths to events.tsv files
```

hndgzkn (approved):
LGTM with minor suggestions.

PR #5562, opened 2025-08-08
[MAINT] add check that regressor do not have decision function

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/issues/5444

- add replacement for sklearn check_regressors_no_decision_function

### DIFF
--- a/nilearn/decoding/decoder.py
+++ b/nilearn/decoding/decoder.py
@@ -1290,9 +1290,6 @@ class Decoder(_ClassifierMixin, ClassifierMixin, _BaseDecoder):
         tags.estimator_type = "classifier"
         tags.classifier_tags = ClassifierTags()
 
-        # TODO
-        # check sklearn ClassifierMixin to adapt this
-
         return tags
 
 
@@ -1478,9 +1475,6 @@ class DecoderRegressor(MultiOutputMixin, RegressorMixin, _BaseDecoder):
 
         tags.estimator_type = "regressor"
         tags.regressor_tags = RegressorTags()
-
-        # TODO
-        # check sklearn RegressorMixin and MultiOutputMixin to adapt this
 
         return tags
 


Remi-Gau:
I believe our regressor and FremRegressor have some methods they should not have, no?

https://nilearn.github.io/dev/modules/generated/nilearn.decoding.DecoderRegressor.html#nilearn.decoding.DecoderRegressor.decision_function

Remi-Gau:
I suspect that fixing this should be done by extracting the methods that are only for classifier in a private mixin class:

see sklearn https://github.com/scikit-learn/scikit-learn/blob/6fd23fca53845b32b249f2b36051c081b65e2fab/sklearn/base.py#L478

Remi-Gau:
ok it seems that some of our regressors do need that decision function...

will fix this.

Remi-Gau:
realizing that some decoders methods do not accept surface data (decision function, predict, score...) when I think they should.

nilearn/decoding/decoder.py
```diff
@@ -1080,6 +1084,27 @@ def __sklearn_tags__(self):
         return tags
 
 
+@fill_doc
+class _ClassifierMixin:
+    def decision_function(self, X):
```
Remi-Gau, line 1089: this only exposes the private function as part of the public API

nilearn/decoding/decoder.py
```diff
@@ -1264,6 +1290,9 @@ def __sklearn_tags__(self):
         tags.estimator_type = "classifier"
         tags.classifier_tags = ClassifierTags()
 
+        # TODO
+        # check sklearn ClassifierMixin to adapt this
```
bthirion: Not sure I understand this

nilearn/decoding/decoder.py
```diff
@@ -1450,6 +1479,9 @@ def __sklearn_tags__(self):
         tags.estimator_type = "regressor"
         tags.regressor_tags = RegressorTags()
 
+        # TODO
+        # check sklearn RegressorMixin and MultiOutputMixin to adapt this
```
bthirion: and that

bthirion (approved):
LGTM overall.

nilearn/decoding/decoder.py
Remi-Gau: ha yeah sorry

this was a note to self

So the [ClassifierMixin](https://github.com/scikit-learn/scikit-learn/blob/c5497b7f7eacfaff061cf68e09bcd48aa93d4d6b/sklearn/base.py#L478) that we use as a one of the parent class for our classifiers does 2 things:
- sets the tags specific for classifiers
- adds a score method

Both of those things we need to implement ourselves because:
- Nilearn has a few extra tags (tags to mention that Surface or Nifti image are valid inputs...)
- We have a score function in the our BaseDecoder

So I wanted to make sure the tags we set match those used in the ClassifierMixin but I was also wondering whether we should not avoid using the sklearn ClassifierMixin because we need to override all the things it does.

nilearn/decoding/decoder.py
Remi-Gau: same reasoning for this

nilearn/decoding/decoder.py
Remi-Gau: Will remove the comments and open a separate issue to discuss and track and will suggest a PR to show what could be done.

Remi-Gau: See #5583

nilearn/decoding/decoder.py
Remi-Gau: See https://github.com/nilearn/nilearn/issues/5583

Issue #5558, opened 2025-08-07
allow to pass some parameters to the sklearn estimator that will be used by our decoders


Remi-Gau:
Is there a way to pass some parameters to the estimator that will be used to avoid having to silence warnings like this?

currently the estimator is initalised at fit time with max_iter=10000, but it could be easier for users to do:

```python
decoder = Decoder(
                estimator="svc",
                mask=mask_img,
                smoothing_fwhm=4,
                cv=3,
                standardize="zscore_sample",
                screening_percentile=sp,
                param_grid=param_grid,
                arg_estimator={"max_iter": 20000} # dict with args to estimator
            )
```

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5520#discussion_r2194804493_

Remi-Gau:
see also https://github.com/nilearn/nilearn/pull/5520#discussion_r2194810893

bthirion:
Agreed !

Remi-Gau:
- adapt the tests of some estimators to make sure they do not fail: https://github.com/nilearn/nilearn/pull/5566#discussion_r2276038829

PR #5566, opened 2025-08-08
[MAINT] check image estimator idempotence

Remi-Gau:
- Relates to #5444, #5447, #5446, #5445

-
-

### DIFF
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -268,6 +268,9 @@ def return_expected_failed_checks(
         "check_fit_check_is_fitted": (
             "replaced by check_img_estimator_fit_check_is_fitted"
         ),
+        "check_fit_idempotent": (
+            "replaced by check_img_estimator_fit_idempotent"
+        ),
         "check_fit_score_takes_y": (
             "replaced by check_masker_fit_score_takes_y"
         ),
@@ -280,7 +283,6 @@ def return_expected_failed_checks(
         # has not yet been created.
         "check_estimators_nan_inf": "TODO",
         "check_estimators_overwrite_params": "TODO",
-        "check_fit_idempotent": "TODO",
         "check_methods_sample_order_invariance": "TODO",
         "check_methods_subset_invariance": "TODO",
         "check_positive_only_tag_during_fit": "TODO",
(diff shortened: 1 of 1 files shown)


.github/workflows/test_with_tox.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
bthirion: this is unrelated ?

nilearn/_utils/estimator_checks.py
```diff
@@ -940,6 +945,73 @@ def check_img_estimator_cache_warning(estimator) -> None:
...
+                tol = 2 * np.finfo(np.float64).eps
+
+            if (
+                isinstance(estimator, FREMClassifier)
+                and method == "decision_function"
+            ):
+                # TODO
+                # Fails for FREMClassifier
+                # mostly on Mac and sometimes linux
+                continue
```
bthirion, line 1176: I think that this is because the method is stochastic.

nilearn/_utils/estimator_checks.py
```diff
@@ -940,6 +945,73 @@ def check_img_estimator_cache_warning(estimator) -> None:
...
+
+            # TODO
+            # some estimator can return some pretty different results
+            # investigate why
```
bthirion, line 1180: Open an issue for that ?

.github/workflows/test_with_tox.yml
Remi-Gau: yeah I just wanted to see if there was some specifc pattern in the tests failure across OS / python: it needs to be reverted before merging

nilearn/_utils/estimator_checks.py
Remi-Gau: yeah that's what I thought and was about to ask about this

any way we can make this more reproducible?

bthirion: if you increase the number of iterations, it converges to something... but this is costly. 
The best short-term option is to relax the parameters that measure convergence.

nilearn/_utils/estimator_checks.py
Remi-Gau: see #5584

.github/workflows/test_with_tox.yml
```suggestion
            fail-fast: true
```

nilearn/_utils/estimator_checks.py
Remi-Gau: - I'd be tempted to wait until we have implemented https://github.com/nilearn/nilearn/issues/5558 . It will make it a lot easier to modify on the fly the parameters of the sklearn estimators we are using under the hood.

unless I am missing something about how to do this

bthirion: Makes sense.

PR #5567, opened 2025-08-11
[TEST] check CI runs

Remi-Gau:
- Closes none

- seems that github CI is using a newer mac-OS for latest and that leads to some systematic failure
- also checking some failures of testing with our oldest dependencies on windows

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -66,7 +66,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
             matrix:
                 description: [latest dependencies]
                 py: ['3.13', '3.12', '3.11', '3.10', '3.9']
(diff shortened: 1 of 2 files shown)


Remi-Gau:
> * seems that github CI is using a newer mac-OS for latest and that leads to some systematic failure

see https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/

I suspect that the ancient version of matplotlib we support but that is no longer maintained will not get fixed for this: for now switching to testing on a older macos version

We should be able to go back to testing on macos latest when dropping python 3.9 as it should trigger a bump of matplotlib to something more recent.

Remi-Gau:
OK will merge this one because without it the CI of all other PR is stuck

.github/workflows/testing_minimum.yml
```diff
@@ -16,6 +16,9 @@ on:
     push:
         branches:
         -   main
+    pull_request:
+        branches:
+        -   '*'
```
```suggestion
```

.github/workflows/test_with_tox.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```

PR #5569, opened 2025-08-11
[FIX] check all decoders can work with surfaces

Remi-Gau:
- Closes none

- make sure all decoders are tested on surface images
- check that decoders that can work both with volume and surface data check compatibility of input images with their mask (for example to also avoid trying to get score from surface data when the model was fitted on volume)

### DIFF
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -1372,7 +1372,7 @@ def check_decoders_with_surface_data(estimator_orig):
             else:
                 getattr(estimator, method)(X)
 
-                
+
 @ignore_warnings
 def check_img_regressors_no_decision_function(regressor_orig):
     """Check that regressors don't have a decision_function.
@@ -1387,7 +1387,6 @@ def check_img_regressors_no_decision_function(regressor_orig):
     funcs = ["decision_function"]
     for func_name in funcs:
         assert not hasattr(regressor, func_name)
-
 
 
 # ------------------ MASKER CHECKS ------------------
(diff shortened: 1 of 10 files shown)


nilearn/decoding/searchlight.py
```diff
@@ -357,7 +357,7 @@ def __sklearn_tags__(self):
         from nilearn._utils.tags import InputTags
 
         tags = super().__sklearn_tags__()
-        tags.input_tags = InputTags(surf_img=True)
+        tags.input_tags = InputTags(surf_img=False)
```
Remi-Gau, line 362: searchlight does not fit surface image directly

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1220,73 +1220,6 @@ def test_decoder_adjust_screening_greaterthan_mask_surface(
     assert adjusted == 100
 
 
-@pytest.mark.parametrize("mask", [None, SurfaceMasker()])
```
Remi-Gau, line 1225: those tests were made more generic and moved to nilearn._utils.estimator_checks

nilearn/_utils/estimator_checks.py
```diff
@@ -1199,6 +1201,80 @@ def check_decoder_empty_data_messages(estimator):
...
+        with pytest.raises(
+            TypeError,
+            match=("Mask and input images must be of compatible types"),
+        ):
+            getattr(estimator, method)(*input)
+
+
+@ignore_warnings()
+def check_decoders_with_surface_data(estimator_orig):
+    """Test fit for surface image."""
```
```suggestion
    """Test fit and other methods with surface image."""
```

Remi-Gau:
ok it seems that the code for several of the methods of our decoders (predict, score, decision_function) can both accept image or arrays

those different behavior are not properly docummented or tested

will try to fix but in a follow up PR to keep changes small

PR #5573, opened 2025-08-11
[DOC] add example with parametric modulation

### LINKED ISSUE #3128: Create or enhance example with parametric and duration modulation
I think a good introduction to modulation would be to use reaction time, as in [Grinband et al. (2008)](https://doi.org/10.1016/j.neuroimage.2008.07.065). We could walk through how orthogonalization should be used with modulated regressors, as in [Mumford, Poline, & Poldrack (2015)](https://doi.org/10.1371/journal.pone.0126255).

I think a good dataset to use would be ds000030 on OpenNeuro. This dataset is used in [Example 9.5.9](https://nilearn.github.io/stable/auto_examples/04_glm_first_level/plot_bids_features.html#sphx-glr-auto-examples-04-glm-first-level-plot-bids-features-py), and files can be fetched with [`nilearn.datasets.fetch_openneuro_dataset_index ()`](https://nilearn.github.io/stable/modules/generated/nilearn.datasets.fetch_openneuro_dataset_index.html#nilearn.datasets.fetch_openneuro_dataset_index). Specifically, the stop-signal task has a constant duration across trials, but response time values that could be used for modulation.

This may be considered related to #2384, though I think the example would probably use some custom code to accomplish what #2384 would integrate somewhere into the API.

Remi-Gau:
- Closes #3128

-
-


bthirion:
Can you trigger a build ? so that we see how it looks. Best,

bthirion:
Still fails due to data fetching.

examples/04_glm_first_level/plot_parametric_modulation.py
```diff
@@ -0,0 +1,110 @@
...
+report
+
+
+# %%
+# First level model estimation with parametric modulation
+# ------------------------------------------
+events[0] = events[0].rename(columns={"ReactionTime": "modulation"})
+model.fit(imgs, events=events)
+
+report = model.generate_report(
```
bthirion, line 102: it seems that nothing survives the threshold.

examples/04_glm_first_level/plot_parametric_modulation.py
```diff
@@ -0,0 +1,110 @@
...
+    contrasts=["GO", "STOP"],
+    height_control=None,
+    threshold=3,
+    cluster_threshold=10,
+    plot_type="glass",
+    title="with parametric modulation",
```
bthirion, line 108:
```suggestion
    title="effect of reaction time on BOLD response",
```

PR #5575, opened 2025-08-12
[FIX] check n_features_in_ for estimators

Remi-Gau:
- Closes none

- implement replacement for sklearn checks on n_feature_in
- see https://scikit-learn.org/stable/developers/develop.html#universal-attributes
- [x] update changelog

### DIFF
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -386,8 +386,8 @@ def unapplicable_checks() -> dict[str, str]:
             "check_fit2d_1feature",
             "check_fit2d_1sample",
             "check_fit2d_predict1d",
-            "check_n_features_in",
-            "check_n_features_in_after_fitting",
+            "check_n_features_in",  # implementable for image
+            "check_n_features_in_after_fitting",  # implementable for image
         ],
         "not applicable for image input",
     )
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -1147,11 +1147,11 @@ def check_img_estimator_n_elements(estimator):
     ), f"Got {estimator.n_elements_=}"
 
     for method in [
-        "predict",
-        "transform",  # TODO
         "decision_function",
-        "score",
         "inverse_transform",
+        "predict",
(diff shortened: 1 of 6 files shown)


nilearn/glm/first_level/first_level.py
```diff
@@ -1266,6 +1266,8 @@ def _prepare_mask(self, run_img):
 
             self.masker_ = self.mask_img
 
+        self.n_elements_ = self.masker_.n_elements_
```
Remi-Gau, line 1269: our maskers have an `n_elements_` attribute that plays the same role as `n_feature_in_` attribute from sklearn, I decided to generalize and use the same name rather than having some of our estimators with n_elements_ and others with n_feature_in_

but we could also decide to use n_feature_in_ everywhere but that would require some deprecation cycle

Remi-Gau:
needs an update in the changelog and of doc strings

nilearn/glm/first_level/first_level.py
bthirion: I'm with you on that one.

nilearn/_utils/estimator_checks.py
```diff
@@ -350,6 +352,10 @@ def return_expected_failed_checks(
         expected_failed_checks.pop("check_estimator_sparse_tag")
 
     if is_masker(estimator):
+        expected_failed_checks |= {
+            "check_n_features_in": "TODO",
```
bthirion: what is the semantics of "TODO" here ?

bthirion (approved):
Mostly happy with the change, although I confess i get a bit lost with the formalism.

nilearn/_utils/estimator_checks.py
Remi-Gau: my bad those TODO should be replaced by "replaced by check_img_estimator_n_elements", this way when the sklearn estimators checks named "check_n_features_in" and "check_n_features_in_after_fitting" fail during our tests, they will show as a expected failed test and will display what's in the value of the dictionary for that check as reason why the test failed

Remi-Gau: TODOs have been replaced

PR #5577, opened 2025-08-13
[MAINt] clean conftest

Remi-Gau:
- Closes none

- minimize how much conftest import from nilearn so we avoid using our own code when generating fixtures

### DIFF
--- a/nilearn/conftest.py
+++ b/nilearn/conftest.py
@@ -7,13 +7,9 @@ import numpy as np
 import pandas as pd
 import pytest
 from nibabel import Nifti1Image
+from scipy.signal import get_window
 
 from nilearn import image
-from nilearn._utils.data_gen import (
-    generate_fake_fmri,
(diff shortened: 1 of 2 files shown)


nilearn/conftest.py
```diff
@@ -7,21 +7,16 @@
...
-from nilearn._utils.data_gen import (
-    generate_fake_fmri,
-    generate_labeled_regions,
-    generate_maps,
-)
 from nilearn._utils.helpers import is_matplotlib_installed
 
 # we need to import these fixtures even if not used in this module
 from nilearn.datasets.tests._testing import (
     request_mocker,  # noqa: F401
```
Remi-Gau, line 15: moving the request mocker should be moved too but it's a much bigger change so I want to do it in a separate PR

Remi-Gau:
closing for now
not sure of the benefit
may get back to it later

PR #5578, opened 2025-08-13
[FIX] make sure GLM reports plot according to two_sided and cluster_threshold parameters

### LINKED ISSUE #5572: [BUG] GLM report figures are always 2 sided even if one sided tests are requested

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 24.04

### Python version

- [x] 3.13
- [ ] 3.12
- [ ] 3.11
- [ ] 3.10
- [ ] 3.9

### nilearn version

dev

### Expected behavior

when requesting a GLM report with one sided results I would expect the figures to only show positive results

Image

```python
from nilearn.datasets import fetch_spm_auditory

subject_data = fetch_spm_auditory()
fmri_img = subject_data.func
mean_func_img = mean_img(subject_data.func[0], copy_header=True)

flm = FirstLevelModel(
    t_r=7,
    noise_model="ar1",
    standardize=False,
    hrf_model="spm",
    drift_model="cosine",
    high_pass=0.01,
    smoothing_fwhm=4
)

flm = flm.fit(subject_data.func, events=subject_data.events)

kwargs = {"cluster_threshold": 10,
          "contrasts"  : ["listening"], 
          "height_control"  : None, 
          "threshold": 5,
          "bg_img"  : mean_func_img}

report = flm.generate_report(two_sided=False, title="one sided", **kwargs)
report.open_in_browser()

report = flm.generate_report(two_sided=True, title="two sided", **kwargs)
report.open_in_browser()
```

Remi-Gau:
- Closes #5572

- [x] update changelog
- [x] add tests that check what the report figure should look like
- [x] check results on surface GLM
- [x] update get clusters table to be able to work with negative one sided case
- [x] update GLM report generation to take into account two_sided-ness and cluster_threshold in the plotted images
- [x] add a warning when trying to threshold an image with the non-default threshold when height_control is not None (threshold is not used in those cases)


Remi-Gau:
ok this affected the surface GLM:

before (this is buggy because this should only show the positive results)

image

after (this is now buggy because the background brain is not visible)

image

nilearn/reporting/tests/baseline/test_stat_map_to_png_volume_0-False--3.09-fpr-slice.png
Remi-Gau, line 1: for height_control != None, I think we cannot plot 'deactivation' because in those cases threshold_stats_img works with alpha values and those are positive.

nilearn/reporting/tests/baseline/test_stat_map_to_png_surface_None-False--0.5.png
Remi-Gau, line 1: subtle difference for the surface plots of negative tow-sided = False: the colormap is inverted (0 at the bottom and more negative values at the top) when compared to the volume counterpart (that has 0 at the top and more negative values at the bottom).

Remi-Gau: note this is not a bug

nilearn/reporting/glm_reporter.py
```diff
@@ -211,6 +211,28 @@ def make_glm_report(
...
+                UserWarning,
+                stacklevel=find_stack_level(),
+            )
+        # setting threshold to the value to its default in threshold_stats_img
+        # to avoid further warnings
+        # TODO
+        # homogenize the default threshold of both
+        # make_glm_report, threshold_stats_img,
+        # FirstLevelModel.generate_report, SecondLevelModel.generate_report
+        # to both be 3.0 or 3.09 and not a mix of both
```
Remi-Gau: Probably worth starting a deprecation cycle to standardize this.

nilearn/glm/tests/test_thresholding.py
```diff
@@ -317,3 +338,39 @@ def test_threshold_stats_img_surface_with_mask(surf_img_1d, surf_mask_1d):
...
+
+    result, _ = threshold_stats_img(
+        surf_img_1d, height_control=None, threshold=-0.5, two_sided=False
+    )
+
+    # one sided negative
+    assert_equal(result.data.parts["left"], np.asarray([0.0, -1.0, 0.0, 0.0]))
+    assert_equal(
+        result.data.parts["right"], np.asarray([0.0, -2.0, 0.0, 0.0, 0.0])
+    )
```
bthirion, line 411: bzw, I noticed that the behavior of `threshold_stats_img` is not documented in that case in particular. 
It's annoying because the behavior can be somewhat counterintuitive.

bthirion (commented):
LGTM overall, but may be some of the issues are still there in colorbar handling ?

nilearn/reporting/glm_reporter.py
bthirion: See #5601

Remi-Gau: done

Remi-Gau:
@emdupre @hndgzkn @man-shu 

this is one of the few PR that needs review before we can start the release process

nilearn/reporting/get_clusters_table.py
```diff
@@ -371,9 +375,14 @@ def get_clusters_table(
         # Now re-label and create table
         label_map = label(binarized, bin_struct)[0]
         clust_ids = sorted(np.unique(label_map)[1:])
-        peak_vals = np.array(
-            [np.max(temp_stat_map * (label_map == c)) for c in clust_ids]
-        )
+        if not two_sided and stat_threshold < 0:
+            peak_vals = np.array(
+                [np.min(temp_stat_map * (label_map == c)) for c in clust_ids]
```
bthirion, line 380: Don't we ned to re-flip the sign ?

bthirion (commented):
LGTM besides one detail. Thx,

nilearn/reporting/get_clusters_table.py
Remi-Gau: given that we are working on `temp_stat_map` and not on `stat_map` we are fine, no?

bthirion: Yes, because the sign has not been flipped in the first place. I guess that you have a test for that anyhow ?

Remi-Gau: let me double check

Remi-Gau: added a check in the test to make sure the content of the original image is unchanged

nilearn/glm/tests/test_thresholding.py
```diff
@@ -334,6 +365,62 @@ def test_threshold_stats_img_surface_with_mask(surf_img_1d, surf_mask_1d):
...
+
+    # one sided, with threshold = 0
+    result, _ = threshold_stats_img(
+        surf_img_1d, height_control=None, threshold=0
```
hndgzkn, line 415: Don't we need `two_sided=False` here?

nilearn/glm/thresholding.py
```diff
@@ -249,6 +249,15 @@ def threshold_stats_img(
        Desired threshold in z-scale.
        This is used only if height_control is None.
 
+       .. note::
```
hndgzkn, line 252: This could be documented in the same way as [nilearn.image.threshold_img](https://github.com/nilearn/nilearn/blob/0ba67f36ae8aa31f92a58ab114aab1876e10c299/nilearn/image/image.py#L933) for consistency.

nilearn/reporting/glm_reporter.py
```diff
@@ -160,6 +159,15 @@ def make_glm_report(
         Cluster forming threshold in same scale as `stat_img` (either a
         t-scale or z-scale value). Used only if height_control is None.
 
+        .. note::
```
hndgzkn, line 162: The same here.

nilearn/reporting/glm_reporter.py
```diff
@@ -821,44 +847,76 @@ def _stat_map_to_png(
...
+        plot_kwargs["threshold"] = threshold
+
+        if not two_sided and threshold < 0:
+            # we cannot use negative threshold in plot_surf_stat_map
+            # so we flip the sign of the image, the colormap, the threshold
+            # and we relabel the colorbar later
+            for k, v in stat_img.data.parts.items():
+                stat_img.data.parts[k] = -v
+            plot_kwargs["threshold"] = -threshold
+            (vmin, vmax) = (-vmax, -vmin)
```
hndgzkn, line 905: I couldn't be sure. Is this line necessary?

nilearn/reporting/glm_reporter.py
```diff
@@ -872,7 +930,10 @@ def _stat_map_to_png(
                 display_mode=display_mode,
                 cmap=cmap,
                 symmetric_cbar=symmetric_cbar,
-                threshold=threshold,
```
hndgzkn, line 875: Instead of removing threshold could we pass `abs(threshold)`. That would help make clear on the plot that those values are thresholded. In the below figures, it is written in the title, but not clear on the images. Also for two_sided thresholded images, the thresholded parts are all white, but not for one sided cases.

nilearn/reporting/glm_reporter.py
```diff
@@ -881,7 +942,9 @@ def _stat_map_to_png(
                 plot_abs=False,
                 symmetric_cbar=symmetric_cbar,
                 cmap=cmap,
-                threshold=threshold,
```
hndgzkn, line 884: Same here.

hndgzkn (commented):
@Remi-Gau Sorry, I was late for the review. I had started it, but could actually finish now. I am submitting my comments in any case.

nilearn/glm/tests/test_thresholding.py
Remi-Gau: I could add a test to make sure we get the same result with False or True given that the threshold is 0 (i.e: the image is not thresholded)

is that what you mean?

hndgzkn: This test is exactly the same with the one at line `384` if I am not mistaken. Since `two_sided=True` by default, no problem for line 384. But here, to test for one sided, I think there is a need for `two_sided=False`.

Remi-Gau: ha yeah
will update that

nilearn/reporting/glm_reporter.py
Remi-Gau: with this line:

with


without:
without

we are flipping everything so if we do not flip the min and max we get funky things

hndgzkn: Ok, I see. But do you know why colorbar is inverse for the first one?

nilearn/reporting/glm_reporter.py
Remi-Gau: done

nilearn/glm/thresholding.py
Remi-Gau: done

nilearn/reporting/glm_reporter.py
Remi-Gau: yeah because we are plotting this "as if" it were a positive threshold, so bigger absolute values are at the top, I then had to flip the signs of the labels in the colorbar later in the code to make everything fit.

Remi-Gau: see my comment: https://github.com/nilearn/nilearn/pull/5578#discussion_r2282497881

nilearn/reporting/glm_reporter.py
Remi-Gau: it was not removed but the 'correct' value (with or without sign flip) is now passed via `**plot_kwargs`

nilearn/reporting/glm_reporter.py
hndgzkn: May be for the lines `868--965` setting `symmetric_cbar` depending on the value of `two_sided` and setting the threshold as `abs(threshold)` would solve the problem. I think no other change is necessary. Plotting should handle it.

Because in plotting depending on `symmetric_cbar` value, as far as I remember, vmin, vmax, etc values are obtained from data (if not specified). So if we specify positive threshold, as the data is already thresholded, it will serve to view thresholded part for colorbar.

nilearn/reporting/glm_reporter.py
hndgzkn: Sorry, you are right.

nilearn/reporting/glm_reporter.py
Remi-Gau: implemented some of the changes you suggested in https://github.com/nilearn/nilearn/pull/5639

PR #5472, opened 2025-06-20
kaleido failure

Remi-Gau:
- Relates to #5471

Changes proposed in this pull request:

-
-

Issue #5471, opened 2025-06-20
Kaleido 1.0.0 requires Chrome to be installed


hndgzkn:
Kaleido no more bundles Chrome starting from its [1.0.0 release](https://github.com/plotly/Kaleido/releases/tag/v1.0.0). It requires Chrome to be installed.

Plotly starting from [version 6.1.0](https://github.com/plotly/plotly.py/releases/tag/v6.1.0) (needs verification) also forces to use Kaleido version >= 1.0.0.

This causes plotly image saving tests to fail.

Remi-Gau:
I think we failed to see the failure because of the version we install in our tests:

https://github.com/nilearn/nilearn/blob/8122befa503f1b91af57c1e51f8181d054c42ec6/tox.ini#L45-L53

Remi-Gau:
@hndgzkn 
I am not seeing the failures I would have expected in #5472 

But it seems that the bug that forces us to rely on older version on kaleido on windows is still there (tests hang forever)

Remi-Gau:
Also I would say that we should not delay the release because of this one.

hndgzkn:
> [@hndgzkn](https://github.com/hndgzkn) I am not seeing the failures I would have expected in [#5472](https://github.com/nilearn/nilearn/pull/5472)
> 

I think it is because the [Ubuntu 24.04 image](https://github.com/actions/runner-images/blob/ubuntu24/20250615.1/images/ubuntu/Ubuntu2404-Readme.md#browsers-and-drivers) has already Chrome in it.

PR #5579, opened 2025-08-13
Pixi support

dangom:

- add support for [pixi](https://pixi.sh/latest/) for environment management.

This makes is easier to contribute to nilearn, since all one needs to do to create an editable virtual environment is clone the repository and run:

```bash
pixi shell
# or to get a "conda" like experience
# eval $(pixi shell-hook -e dev)
```

PR should be fully backwards compatible.
(also removed the kaleido separate package management for Windows since that [bug has now been fixed.](https://github.com/plotly/Kaleido/issues/134)

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -656,7 +656,7 @@ Here are the key steps you need to go through to copy the repo before contributi
 
       git clone git@github.com:<your_username>/nilearn.git
 
-2. (optional but highly recommended) set up a virtual environment to wor
+2. (optional but highly recommended) set up a virtual environment to work
    in using whichever environment management tool you're used to and activate it.
    For example:
 
@@ -672,7 +672,15 @@ or:
       conda create -n nilearn pip
       conda activate nilearn
 
+or, if using `pixi <https://pixi.sh/latest/>`_ :
+
+.. code-block:: bash
+
+        eval $(pixi shell-hook -e dev)
+
 3. install the forked version of ``nilearn``
+
+If using pixi, this step is already taken care of by step (2).
 
 .. admonition:: Recommendation
 


Remi-Gau:
I am going to push back on this one.

The main reason is that I don't think we should cater to the preferences of all contributors: I for example loathe conda because it takes 15 business days to setup even the simplest environment and setups the environment outside of the repo I am in by default, so if it were up to me we would remove the conda instructions from our doc.

We could also argue that given that nilearn uses hatch for package build, we should mention how to set up environment with hatch.

Given the number of tools in python to manage environments and install packages:


https://alpopkes.com/posts/python/packaging_tools/

and how quickly they change (note that pixi is not in the venn diagramm that was last updated a year ago), the number of possible workflows is just too big.

I also would feel strange to have a lock file in a library. If nilearn was an app then it would make sense to lock all its dependencies but otherwise I think we should stick the looser contraints we have in the pyproject.toml

I am also afraid of the slippery slope: why accept pixi and not poetry?

So I would stick to what we have in terms of setup and runner (venv, conda, tox) as (I think) those are the most classic ones.

Curious to how others feel about this.

Remi-Gau:
- regarding kaleido there is this PR that was still showing the failures a while back https://github.com/nilearn/nilearn/pull/5472/

rerunning it now to see if more recent versions have been fixed

- see also: https://github.com/nilearn/nilearn/issues/5471

dangom:
That's fair. I'll close this PR.
If anyone ever comes here, you can set up pixi on nilearn by running `pixi init --format pyproject .` after cloning.

The lockfile need not be tracked. It is interesting in that it can be shared and diffed against in case of bug reports. Useful in cases where bugs don't seem reproducible.

Remi-Gau:
> The lockfile need not be tracked.

ah ok: that makes more sense

PR #5582, opened 2025-08-14
[FIX] give a dummy color to background when it needs to be added to LUT

Remi-Gau:
- Closes none but fixes atlas check on main

- when "background" is added to an atlas LUT give if a dummy color if needed

### DIFF
--- a/.github/workflows/test_atlases.yml
+++ b/.github/workflows/test_atlases.yml
@@ -8,9 +8,6 @@ on:
     push:
         branches:
         -   main
-    pull_request:
-        branches:
-        -   '*'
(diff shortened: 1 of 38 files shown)


.github/workflows/test_atlases.yml
```diff
@@ -8,6 +8,9 @@ on:
     push:
         branches:
         -   main
+    pull_request:
+        branches:
+        -   '*'
```
Remi-Gau:
```suggestion
```

Remi-Gau:
@man-shu @bthirion 
I would not mind a quick review of this one to fix our doc build on main

man-shu (approved):
LGTM!

PR #5585, opened 2025-08-18
[MAINT] simplify nilearn.utils.__init__.py 

Remi-Gau:
- Closes none

- remove most objects that are directly accessible from nilearn._utils to avoid importing several modules when any import is made from nilearn._utils

### DIFF
--- a/nilearn/_utils/__init__.py
+++ b/nilearn/_utils/__init__.py
@@ -6,8 +6,11 @@ import warnings
 from importlib import import_module
 from pathlib import Path
 
-from .docs import fill_doc
-from .niimg_conversions import check_niimg, check_niimg_3d, check_niimg_4d
+from nilearn._utils.niimg_conversions import (
+    check_niimg,
+    check_niimg_3d,
+    check_niimg_4d,
+)
 
 
 def all_modules(modules_to_ignore=None, modules_to_consider=None):
@@ -193,5 +196,4 @@ __all__ = [
     "check_niimg",
     "check_niimg_3d",
     "check_niimg_4d",
-    "fill_doc",
 ]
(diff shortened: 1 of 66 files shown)


Remi-Gau:
```python
import sys
from nilearn._utils import logger

print([x for x in sys.modules if 'nilearn' in x])
```

on main:

```
['nilearn._version', 'nilearn', 'nilearn._utils.logger', 'nilearn._utils.helpers', 'nilearn._utils.cache_mixin', 'nilearn._utils.docs', 'nilearn._utils.niimg', 'nilearn._utils.exceptions', 'nilearn._utils.path_finding', 'nilearn.typing', 'nilearn._utils.niimg_conversions', 'nilearn._utils.numpy_conversions', 'nilearn._utils']
```

in this PR:

```
['nilearn._version', 'nilearn', 'nilearn._utils', 'nilearn._utils.logger']
```

this should help avoid some circular imports

emdupre:
I guess this relates to #5538 ? Would be curious if it helps to simplify the plot generated there ! Otherwise, this will break anything relying on these helpers ; not innately a problem, since we don't support using private code, but would be nice to be clearer on what the benefits are :) 

Otherwise, the code changes LGTM ; there's a build failure, but it seems like it's unrelated.

Remi-Gau:
> this will break anything relying on these helpers

I guess you meant it will break if people do:

from nilearn._utils import check_niimg

instead of 

from nilearn._utils.niimg_conversion import check_niimg

Remi-Gau:
> Would be curious if it helps to simplify the plot generated there

will generate something but the change may be easier to read when seen at the sub-level package

Remi-Gau:
Example of dependency graph of nilearn.masking on main and on this PR

`pydeps nilearn/masking.py --only nilearn`

## on main

nilearn_masking_before


## on this PR

nilearn_masking_after

Remi-Gau:
what's notable is that more modules are purely imported and not imported AND importing.

emdupre:
> I guess you meant it will break if people do:
> 
> from nilearn._utils import check_niimg

Yes, as someone who does that 😅 

> what's notable is that more modules are purely imported and not imported AND importing.

Thanks for re-generating those plots and confirming ! Glad to see that

Remi-Gau:
> Yes, as someone who does that

you use private functions!!! 😱 #DramaQueen

PR #5594, opened 2025-08-21
[FIX] remove is_classif (and loss) parameter from SpaceNet classes

### LINKED ISSUE #5568: remove is_classif parameter from BaseSpaceNet class
something similar needs to be done for spacenet estimators

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/5557#issuecomment-3173476219_
            
rely on tags or private attributes / property

Remi-Gau:
- Closes #5568 

- remove is_classif parameter from SpaceNet classes and rely on tags to know the nature of the estimator

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,24 +13,21 @@ Fixes
 
 - :bdg-success:`API` Add a dummy ``y`` parameter to :meth:`~nilearn.decomposition.CanICA.score` and :meth:`~nilearn.decomposition.DictLearning.score` for compatibility with scikit-learn API (:gh:`5565` by `Rémi Gau`_).
 
-
 - :bdg-dark:`Code` Fix several issues in :class:`~nilearn.maskers.NiftiLabelsMasker` and :class:`~nilearn.maskers.SurfaceLabelsMasker` that lead to invalid ``region_names_``, ``region_ids_`` or look-up-table content (:gh:`5492` by `Rémi Gau`_).
 
 - :bdg-dark:`Code` Align ``symmetric_cmap`` behavior for ``plotly`` backend in :func:`~nilearn.plotting.plot_surf` function with ``matplotlib`` backend (:gh:`5492` by `Hande Gözükan`_).
+
 - :bdg-dark:`Code` Fix type of ``t_r`` to support numpy dtypes for python < 3.10 (:gh:`5550` by `Rémi Gau`_).
 
 - :bdg-dark:`Code` Enforce consistent ``dtype`` for all parts of :class:`~nilearn.surface.SurfaceImage` and :class:`~nilearn.surface.PolyData` (:gh:`5530` by `Rémi Gau`_).
 
-- :bdg-success:`API` The ``is_classif`` public attribute has been removed for :class:`~decoding.SpaceNetClassifier` and :class:`~decoding.SpaceNetRegressor`
-as it is a characteristic of the estimator that must not be changed. Accessing an equivalent characteristic can be done via the estimator's tags (``__sklearn_tags__()``) (:gh:`5594` by `Rémi Gau`_).
+- :bdg-success:`API` The ``is_classif`` public attribute has been removed for :class:`~decoding.SpaceNetClassifier` and :class:`~decoding.DecoderRegressor` as it is a characteristic of the estimator that must not be changed. Accessing an equivalent characteristic can be done via the estimator's tags (``__sklearn_tags__()``) (:gh:`5594` by `Rémi Gau`_).
 
-- :bdg-success:`API` The ``is_classification`` public attribute has been removed for :class:`~decoding.Decoder`, :class:`~decoding.Regressor`, :class:`~decoding.FREMClassifier` and  :class:`~decoding.FREMRegressor`
-as it is a characteristic of the estimator that must not be changed. Accessing an equivalent characteristic can be done via the estimator's tags (``__sklearn_tags__()``) (:gh:`5557` by `Rémi Gau`_).
+- :bdg-success:`API` The ``is_classification`` public attribute has been removed for :class:`~decoding.Decoder`, :class:`~decoding.DecoderRegressor`, :class:`~decoding.FREMClassifier` and  :class:`~decoding.FREMRegressor` as it is a characteristic of the estimator that must not be changed. Accessing an equivalent characteristic can be done via the estimator's tags (``__sklearn_tags__()``) (:gh:`5557` by `Rémi Gau`_).
 
-- :bdg-success:`API` The ``loss`` public attribute has been removed for :class:`~decoding.SpaceNetRegressor`
-it can only be ``'mse'`` and should not be changed (:gh:`5594` by `Rémi Gau`_).
+- :bdg-success:`API` The ``loss`` public attribute has been removed for :class:`~decoding.SpaceNetRegressor` it can only be ``'mse'`` and should not be changed (:gh:`5594` by `Rémi Gau`_).
 
-- :bdg-success:`API` The ``clustering_percentile`` public attribute has been removed for :class:`~decoding.Decoder` and :class:`~decoding.Regressor` as it is only relevant for :class:`~decoding.FREMClassifier` and  :class:`~decoding.FREMRegressor` (:gh:`5557` by `Rémi Gau`_).
+- :bdg-success:`API` The ``clustering_percentile`` public attribute has been removed for :class:`~decoding.Decoder` and :class:`~decoding.DecoderRegressor` as it is only relevant for :class:`~decoding.FREMClassifier` and  :class:`~decoding.FREMRegressor` (:gh:`5557` by `Rémi Gau`_).
 
 
 Enhancements


nilearn/decoding/space_net.py
```diff
@@ -1413,6 +1426,9 @@ class SpaceNetRegressor(BaseSpaceNet):
     penalty : :obj:`str`, default='graph-net'
         Penalty to used in the model. Can be 'graph-net' or 'tv-l1'.
 
+    loss : :obj:`str`, default="mse"
+        Loss to be used in the model. Must be "mse".
```
Remi-Gau: loss can only be 'mse' for SpaceNetRegressor, so it feels strange to have it as a public parameter that could be changed by users

so for this one too, I will only expose via a public property for SpaceNetRegressor

bthirion (approved):
The RP LGTM technically, but I'm not sure about one think: the only supported loss are still "mse" and "logistic"... where are we enforcing that (since  l.704 is removed) ?

Remi-Gau:
> The RP LGTM technically, but I'm not sure about one think: the only supported loss are still "mse" and "logistic"... where are we enforcing that (since l.704 is removed) ?

let me make it more explicit

nilearn/decoding/space_net.py
```diff
@@ -818,20 +818,8 @@ def _check_params(self):
...
-            self.loss is not None
-            and not self.is_classif
-            and (self.loss == "logistic")
-        ):
-            raise ValueError(
-                "'logistic' loss is only available for classification "
-                "problems."
-            )
+        if self._is_classification:
+            self._validate_loss(self.loss)
```
Remi-Gau, line 816: only for classification can the user choose the loss, so we validate here

nilearn/decoding/space_net.py
```diff
@@ -841,11 +829,23 @@ def _set_coef_and_intercept(self, w):
         if self.w_.ndim == 1:
             self.w_ = self.w_[np.newaxis, :]
         self.coef_ = self.w_[:, :-1]
-        if self.is_classif:
+        if self._is_classification:
             self.intercept_ = self.w_[:, -1]
         else:
             self._set_intercept(self.Xmean_, self.ymean_, self.Xstd_)
 
+    def _return_loss_value(self):
```
Remi-Gau, line 831: get loss for classification but automatically set it to "mse" for regressor

bthirion:
OK, makes sense, thx.

Remi-Gau:
will update the changelog as this modifies some public attributes

doc/changes/latest.rst
```diff
@@ -18,6 +18,17 @@ Fixes
...
+as it is a characteristic of the estimator that must not be changed. Accessing an equivalent characteristic can be done via the estimator's tags (``__sklearn_tags__()``) (:gh:`5594` by `Rémi Gau`_).
+
+- :bdg-success:`API` The ``is_classification`` public attribute has been removed for :class:`~decoding.Decoder`, :class:`~decoding.Regressor`, :class:`~decoding.FREMClassifier` and  :class:`~decoding.FREMRegressor`
+as it is a characteristic of the estimator that must not be changed. Accessing an equivalent characteristic can be done via the estimator's tags (``__sklearn_tags__()``) (:gh:`5557` by `Rémi Gau`_).
+
+- :bdg-success:`API` The ``loss`` public attribute has been removed for :class:`~decoding.SpaceNetRegressor`
+it can only be ``'mse'`` and should not be changed (:gh:`5594` by `Rémi Gau`_).
+
+- :bdg-success:`API` The ``clustering_percentile`` public attribute has been removed for :class:`~decoding.Decoder` and :class:`~decoding.Regressor` as it is only relevant for :class:`~decoding.FREMClassifier` and  :class:`~decoding.FREMRegressor` (:gh:`5557` by `Rémi Gau`_).
+
```
Remi-Gau: - also add entries for #5557

doc/changes/latest.rst
```diff
@@ -21,6 +21,17 @@ Fixes
...
 - :bdg-dark:`Code` Enforce consistent ``dtype`` for all parts of :class:`~nilearn.surface.SurfaceImage` and :class:`~nilearn.surface.PolyData` (:gh:`5530` by `Rémi Gau`_).
 
+- :bdg-success:`API` The ``is_classif`` public attribute has been removed for :class:`~decoding.SpaceNetClassifier` and :class:`~decoding.SpaceNetRegressor`
+as it is a characteristic of the estimator that must not be changed. Accessing an equivalent characteristic can be done via the estimator's tags (``__sklearn_tags__()``) (:gh:`5594` by `Rémi Gau`_).
+
+- :bdg-success:`API` The ``is_classification`` public attribute has been removed for :class:`~decoding.Decoder`, :class:`~decoding.Regressor`, :class:`~decoding.FREMClassifier` and  :class:`~decoding.FREMRegressor`
+as it is a characteristic of the estimator that must not be changed. Accessing an equivalent characteristic can be done via the estimator's tags (``__sklearn_tags__()``) (:gh:`5557` by `Rémi Gau`_).
+
+- :bdg-success:`API` The ``loss`` public attribute has been removed for :class:`~decoding.SpaceNetRegressor`
+it can only be ``'mse'`` and should not be changed (:gh:`5594` by `Rémi Gau`_).
```
```suggestion
- :bdg-success:`API` The ``loss`` public attribute has been removed for :class:`~decoding.SpaceNetRegressor` as
it can only be ``'mse'`` and should not be changed (:gh:`5594` by `Rémi Gau`_).
```

nilearn/decoding/space_net.py
```diff
@@ -782,6 +771,17 @@ def __sklearn_tags__(self):
...
+    def _is_classification(self) -> bool:
+        # TODO remove for sklearn>=1.6
+        # this private method can probably be removed
+        # when dropping sklearn>=1.5 and replaced by just:
+        #   self.__sklearn_tags__().estimator_type == "classifier"
+        if SKLEARN_LT_1_6:
+            # TODO remove for sklearn>=1.6
+            return self._estimator_type == "classifier"
+        return self.__sklearn_tags__().estimator_type == "classifier"
+
```
hndgzkn, line 784: I think we do not need this function in the base class. The classification class should make necessary checks itself.

nilearn/decoding/space_net.py
```diff
@@ -812,20 +812,8 @@ def _check_params(self):
...
-            self.loss is not None
-            and not self.is_classif
-            and (self.loss == "logistic")
-        ):
-            raise ValueError(
-                "'logistic' loss is only available for classification "
-                "problems."
-            )
+        if self._is_classification:
+            self._validate_loss(self.loss)
```
hndgzkn, line 816: The base class would not know which classes will extend it. So it should not make such checks depending on the type of the inheriting class. The inheriting class should overwrite `_check_params` function, call super class `_check_params` for common checks and continue for its own checks. For example in `SpaceNetClassifier` we should have :

```python
def _check_params(self):
    super()._check_params()
    self._validate_loss(self.loss)
```

nilearn/decoding/space_net.py
```diff
@@ -835,11 +823,23 @@ def _set_coef_and_intercept(self, w):
         if self.w_.ndim == 1:
             self.w_ = self.w_[np.newaxis, :]
         self.coef_ = self.w_[:, :-1]
-        if self.is_classif:
+        if self._is_classification:
             self.intercept_ = self.w_[:, -1]
         else:
             self._set_intercept(self.Xmean_, self.ymean_, self.Xstd_)
```
hndgzkn, line 829: The same here. The base class would have `_set_intercept` method that is common to all inheriting classes. And classifier should implement its own behavior. If there is no such common behavior, we can leave _set_intercept as abstract method and raise `NotImplemented` error and each class should implement its behavior.

nilearn/decoding/space_net.py
```diff
@@ -835,11 +823,23 @@ def _set_coef_and_intercept(self, w):
...
+    def _return_loss_value(self):
+        """Set loss value for instances where it is not defined.
+
+        For SpaceNetRegressor it is always "mse".
+        """
+        loss = getattr(self, "loss", None)
+        if loss is None:
+            loss = "logistic"
+            if not self._is_classification:
+                loss = "mse"
```
hndgzkn, line 840: The same here.

`SpaceNetRegressor` can have `_return_loss_value` function that returns `mse` and `SpaceNetRegressor` should have `_return_loss_value` function that implements desired behavior.

This will avoid having many if checks, simplify and clarify the code.

nilearn/decoding/space_net.py
```diff
@@ -883,10 +883,10 @@ def fit(self, X, y):
             ["csr", "csc", "coo"],
             dtype=float,
             multi_output=True,
-            y_numeric=not self.is_classif,
+            y_numeric=not self._is_classification,
```
hndgzkn, line 886: The same here

nilearn/decoding/space_net.py
```diff
@@ -883,10 +883,10 @@ def fit(self, X, y):
...
         )
 
-        if not self.is_classif and np.all(np.diff(y) == 0.0):
+        if not self._is_classification and np.all(np.diff(y) == 0.0):
```
hndgzkn, line 889: same here.

nilearn/decoding/space_net.py
```diff
@@ -932,18 +928,26 @@ def fit(self, X, y):
...
+            self._binarize_y(y)
+            if self._is_classification
+            else y[:, np.newaxis]
+        )
 
         n_problems = (
-            self.n_classes_ if self.is_classif and self.n_classes_ > 2 else 1
+            self.n_classes_
+            if self._is_classification and self.n_classes_ > 2
+            else 1
```
hndgzkn, line 950: same here.

nilearn/decoding/space_net.py
```diff
@@ -985,7 +989,7 @@ def fit(self, X, y):
                 solver_params,
                 n_alphas=self.n_alphas,
                 eps=self.eps,
-                is_classif=self.loss == "logistic",
+                is_classif=self._is_classification,
```
hndgzkn, line 992: we should not have `is_classif` here.

hndgzkn (commented):
@Remi-Gau  I think you address a very good point with this PR. However independent of this PR, there is a problem with design of these classes. 

The base class should have only common functionality, and inheriting classed should implement itself what is different from the base one. This PR can be a good opportunity to fix this design problem. The changes you proposed could be implemented in inheriting classes rather that the base class.

Remi-Gau:
@hndgzkn 

I agree with your suggestions but do you want me to try to implement those refactorings in this PR or in a follow up?

I'd rather do it in a follow up: because this PR is a bit of a blocker for tidying up the doc string of spacenet estimators and I'd rather keep PRs small

hndgzkn:
> @hndgzkn
> 
> I agree with your suggestions but do you want me to try to implement those refactorings in this PR or in a follow up?
> 
> I'd rather do it in a follow up: because this PR is a bit of a blocker for tidying up the doc string of spacenet estimators and I'd rather keep PRs small

No problem with doing it in another PR.

Remi-Gau:
> No problem with doing it in another PR.

ok so if you give me a 👍🏽 I will merge this and start the refactoring you suggested

hndgzkn:
> > No problem with doing it in another PR.
> 
> ok so if you give me a 👍🏽 I will merge this and start the refactoring you suggested

Ah yes, sorry!

hndgzkn (approved):

PR #5595, opened 2025-08-21
[REF] extract _ClassifierMixin and _RegressorMixin

### LINKED ISSUE #5583: use our own ClassifierMixin and RegressorMixin rather than using those from sklearn
the [ClassifierMixin](https://github.com/scikit-learn/scikit-learn/blob/c5497b7f7eacfaff061cf68e09bcd48aa93d4d6b/sklearn/base.py#L478) that we use as a one of the parent class for our classifiers does 2 things:
- sets the tags specific for classifiers
- adds a score method

Both of those things we need to implement ourselves because:
- Nilearn has a few extra tags (tags to mention that Surface or Nifti image are valid inputs...)
- We have a score function in the our BaseDecoder

So I wanted to make sure the tags we set match those used in the ClassifierMixin but I was also wondering whether we should not avoid using the sklearn ClassifierMixin because we need to override all the things it does.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5562#discussion_r2275923605_

Remi-Gau:
- Closes #5583

- [x] extract _ClassifierMixin and _RegressorMixin
- [x] update changelog

### DIFF
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -634,6 +634,7 @@ def nilearn_check_generator(estimator: BaseEstimator):
     else:
         requires_y = getattr(tags.target_tags, "required", False)
 
+    yield (clone(estimator), check_tags)
     yield (clone(estimator), check_transformer_set_output)
 
     if isinstance(estimator, CacheMixin):
@@ -875,6 +876,14 @@ def fit_estimator(estimator: BaseEstimator) -> BaseEstimator:
 
 
 # ------------------ GENERIC CHECKS ------------------
+
+
+def check_tags(estimator):
+    """Check tags are the same with old and new methods.
+
+    TODO (sklearn >= 1.6) remove this check when bumping sklearn above 1.5
+    """
+    assert estimator._more_tags() == estimator.__sklearn_tags__()
 
 
 def _check_mask_img_(estimator):
--- a/nilearn/decoding/_mixin.py
+++ b/nilearn/decoding/_mixin.py
@@ -4,8 +4,7 @@ from nilearn._utils.tags import SKLEARN_LT_1_6
 
 
 class _ClassifierMixin:
-    # TODO remove for sklearn>=1.6
-    _estimator_type = "classifier"
+    _estimator_type = "classifier"  # TODO (sklearn >= 1.6) remove
 
     def __sklearn_tags__(self):
         """Return estimator tags.
@@ -13,9 +12,8 @@ class _ClassifierMixin:
         See the sklearn documentation for more details on tags
         https://scikit-learn.org/1.6/developers/develop.html#estimator-tags
         """
-        # TODO
-        # get rid of if block
-        # bumping sklearn_version > 1.5
+        # TODO (sklearn >= 1.6) get rid of if block
+        # when bumping sklearn_version > 1.5
         # see https://github.com/scikit-learn/scikit-learn/pull/29677
         tags = super().__sklearn_tags__()
         if SKLEARN_LT_1_6:
@@ -30,8 +28,7 @@ class _ClassifierMixin:
 
 
 class _RegressorMixin:
-    # TODO remove for sklearn>=1.6
-    _estimator_type = "regressor"
+    _estimator_type = "regressor"  # TODO (sklearn >= 1.6) remove
 
     def __sklearn_tags__(self):
         """Return estimator tags.
@@ -39,9 +36,8 @@ class _RegressorMixin:
         See the sklearn documentation for more details on tags
         https://scikit-learn.org/1.6/developers/develop.html#estimator-tags
         """
-        # TODO
-        # get rid of if block
-        # bumping sklearn_version > 1.5
+        # TODO (sklearn >= 1.6) get rid of if block
+        # when bumping sklearn_version > 1.5
         # see https://github.com/scikit-learn/scikit-learn/pull/29677
         tags = super().__sklearn_tags__()
         if SKLEARN_LT_1_6:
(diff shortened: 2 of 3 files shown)


Remi-Gau:
> LGTM ! The PR looks ready ?

yup!

Remi-Gau:
> LGTM. Probably deserves a whatsnew entry.

done

doc/changes/latest.rst
```diff
@@ -33,3 +33,5 @@ Changes
 - :bdg-dark:`Code` Resampling of maps by :class:`~nilearn.maskers.NiftiMapsMasker` is now done with a linear instead of a continuous interpolation  (:gh:`5519` by `Rémi Gau`_).
 
 - :bdg-dark:`Code` Move ``nilearn.plotting.img_plotting`` under ``nilearn.plotting.image`` (:gh:`5481` by `Hande Gözükan`_).
+
+- :bdg-dark:`Code` Decoding estimators do not inherit from sklearn ``ClassifierMixin`` and ``RegressorMixing`` anymore. It is recommended to rely on an estimator tags (accessible via the ``'__sklearn_tags__()'`` special method) to know more about the characteristics of an instance  (:gh:`5595` by `Rémi Gau`_).
```
```suggestion
- :bdg-dark:`Code` Decoding estimators do not inherit from sklearn ``ClassifierMixin`` and ``RegressorMixing`` anymore. It is recommended to rely on estimator tags (accessible via the ``'__sklearn_tags__()'`` special method) to know more about the characteristics of an instance  (:gh:`5595` by `Rémi Gau`_).
```

nilearn/decoding/decoder.py
```diff
@@ -1290,24 +1258,30 @@ def __sklearn_tags__(self):
...
         https://scikit-learn.org/1.6/developers/develop.html#estimator-tags
         """
-        # TODO
-        # get rid of if block
-        # bumping sklearn_version > 1.5
-        # see https://github.com/scikit-learn/scikit-learn/pull/29677
-        tags = super().__sklearn_tags__()
-        if SKLEARN_LT_1_6:
-            return tags
+        return super().__sklearn_tags__()
```
hndgzkn: `_more_tags` can be defined in `_BaseDecoder` and deleted from all subclasses. If I am not mistaken it is the same in all inheriting classes.

hndgzkn: The definition of  `__sklearn_tags__` can be deleted.  If there is not a definition in current class, it will already call the super class method.

nilearn/decoding/decoder.py
```diff
@@ -1468,20 +1439,7 @@ def __sklearn_tags__(self):
...
-        if SKLEARN_LT_1_6:
-            tags["multioutput"] = True
-            return tags
-        from sklearn.utils import RegressorTags
-
-        tags.estimator_type = "regressor"
-        tags.regressor_tags = RegressorTags()
-
-        return tags
+        return super().__sklearn_tags__()
```
hndgzkn, line 1461: The same here.

nilearn/decoding/decoder.py
```diff
@@ -1671,21 +1626,7 @@ def __sklearn_tags__(self):
...
-            tags["multioutput"] = True
-            return tags
-
-        from sklearn.utils import RegressorTags
-
-        tags.estimator_type = "regressor"
-        tags.regressor_tags = RegressorTags()
-
-        return tags
+        return super().__sklearn_tags__()
```
hndgzkn, line 1648: The same here.

nilearn/decoding/decoder.py
```diff
@@ -1882,17 +1820,23 @@ def __sklearn_tags__(self):
...
         https://scikit-learn.org/1.6/developers/develop.html#estimator-tags
         """
-        # TODO
-        # get rid of if block
-        # bumping sklearn_version > 1.5
-        # see https://github.com/scikit-learn/scikit-learn/pull/29677
-        tags = super().__sklearn_tags__()
-        if SKLEARN_LT_1_6:
-            return tags
+        return super().__sklearn_tags__()
```
hndgzkn: The same here.

nilearn/decoding/space_net.py
```diff
@@ -1329,34 +1328,6 @@ def score(self, X, y):
         check_is_fitted(self)
         return accuracy_score(y, self.predict(X))
 
-    def _more_tags(self):
-        """Return estimator tags.
-
-        TODO remove when bumping sklearn_version > 1.5
-        """
-        return self.__sklearn_tags__()
```
hndgzkn: If  `_more_tags` is not necessary here, may be it is not necessary above as well?

hndgzkn (approved):
Some suggestions, otherwise LGTM.

nilearn/decoding/decoder.py
Remi-Gau: pretty sure I have tried in some tests and was getting some tests failure: can you double check to see if we reach the same conclusion?

I am not too fussed about having some extra _more_tags() lying around as they will be deleted anyway in the future.

nilearn/decoding/decoder.py
Remi-Gau: ok turns out that it is needed otherwise the DecoderRegressor will only get the tags from _RegressorMixin

nilearn/decoding/decoder.py
Remi-Gau: ok turns out that it is needed otherwise the DecoderRegressor will only get the tags from _RegressorMixin

Remi-Gau: with sklearn 1.5.2

```python
>>> from nilearn.decoding import FREMRegressor
>>> r = FREMRegressor()
>>> tags = r._more_tags()
>>> tags
{'multioutput': True}
>>> tags = r.__sklearn_tags__()
>>> tags
{'X_types': ['surf_img', 'niimg_like', '2darray'], 'require_y': True, 'multioutput': True}
```

PR #5597, opened 2025-08-21
[MAINT] remove duplicate tests space_net

Remi-Gau:
- Closes none

- remove tests that only test the object construction: there are better, more systematic equivalent from sklearn run via 'test_check_estimator_sklearn' 

for context: estimators that comply with sklearn rules have no logic in their and parameters of the constructor should be found in public attributes of the instance, so as long as our estimators comply with those rules, we do not need further checks on the constructor


bthirion (commented):
I like PRs that remove lines ! Is it final ?

Remi-Gau:
> I like PRs that remove lines ! Is it final ?

it is

PR #5601, opened 2025-08-22
[DEP] start deprecation cycle for default Z threshold

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/5578#discussion_r2282538028

- start deprecation cycle to standardize default Z threshold
- [x] add tests
- [x] update changelog


nilearn/_constants.py
Remi-Gau, line 1: did not want to put this in nilearn.__init__.py to avoid importing more things in nilearn

Remi-Gau: maybe we should make this module public

nilearn/glm/first_level/first_level.py
```diff
@@ -1310,6 +1311,17 @@ def generate_report(
         """
         from nilearn.reporting.glm_reporter import make_glm_report
 
+        if threshold == 3.0:
+            warnings.warn(
+                category=FutureWarning,
+                message=(
+                    "From nilearn version>=0.15, "
+                    "the default 'threshold' will be set to "
+                    f"{DEFAULT_Z_THRESHOLD}."
```
Remi-Gau: would be nice to tell users how to avoid getting this warning

nilearn/_constants.py
bthirion: Why do we need it to be public ?

bthirion (commented):
The PR LGTM, but some tests are needed it seems.

nilearn/_constants.py
Remi-Gau: hmm... let me try to move this in glm.thresholding

nilearn/glm/tests/test_thresholding.py
```diff
@@ -337,7 +337,11 @@ def test_threshold_stats_img_surface_with_mask(surf_img_1d, surf_mask_1d):
 @pytest.mark.parametrize("threshold", [3.0, 2.9, DEFAULT_Z_THRESHOLD])
 @pytest.mark.parametrize("height_control", [None, "bonferroni", "fdr", "fpr"])
 def test_deprecation_threshold(surf_img_1d, height_control, threshold):
-    """Check warning thrown when threshold==old threshold."""
+    """Check warning thrown when threshold==old threshold.
```
bthirion, line 340: Wondering whether we're not overdoing it. 
Shouldn't we simply let the use know that the default has changed when the function with changed default is called ?

Remi-Gau: Well we want to warn user that the default will change and when it will but also tell them what to do if they want to 'opt-in' early to avoid getting the warning.

bthirion: Yep, that's very thoughtful of you.

PR #5609, opened 2025-08-26
[FIX] handle cv is None for decoders

Remi-Gau:
- Closes none

- cv could be None for our decoders but would lead to behavior that was not documented and dfferent from passing the default value: this PR should fix this
- [ ] do the same for spacenet, searchlight (if possible)
- [x] refactor doc string for cv across all estimators

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -16,10 +16,9 @@ import re
 import sys
 from pathlib import Path
 
+from nilearn._version import __version__
 from sphinx.domains import changeset
 from sphinx.locale import _
-
-from nilearn._version import __version__
 
 # ----------------------------------------------------------------------------
 


nilearn/decoding/decoder.py
```diff
@@ -662,6 +662,11 @@ def fit(self, X, y, groups=None):
         # a CV splitter object do check_cv regardless of groups parameter.
         cv = self.cv
 
+        if cv is None:
+            cv = (
+                30 if isinstance(self, (FREMClassifier, FREMRegressor)) else 10
```
bthirion, line 653: why 30 ? Usually we use 5 or 10.

Remi-Gau: for the FREM we have a default of 30

https://github.com/nilearn/nilearn/blob/8e26f5c8c8c9fc50e706aaa94b8aa53d58fd8781/nilearn/decoding/decoder.py#L1646

Remi-Gau: We could change the default but that would require a deprecation cycle

nilearn/connectome/tests/test_group_sparse_cov.py
```diff
@@ -197,3 +200,20 @@ def test_group_sparse_covariance_errors(rng):
...
+        n_features=10,
+        min_n_samples=100,
+        max_n_samples=151,
+        random_state=rng,
+    )
+
+    gsc = GroupSparseCovarianceCV(verbose=0, cv=cv)
+    gsc.fit(signals)
+
+    assert gsc.cv_scores_.shape[1]
```
Remi-Gau: according to the doc string gsc.cv_scores_.shape[1] should be the number of fold but this fails

nilearn/decoding/decoder.py
bthirion: I see. It's because in that case, cv plays the role of a bootstrap iterator.

nilearn/connectome/tests/test_group_sparse_cov.py
```diff
@@ -198,3 +201,25 @@ def test_group_sparse_covariance_errors(rng):
...
+
+    gsc = GroupSparseCovarianceCV(verbose=0, cv=cv)
+    gsc.fit(signals)
+
+    cv_alphas_ = gsc.cv_alphas_
+    assert isinstance(cv_alphas_, list)
+    assert len(cv_alphas_) == 16
+
+    cv_scores_ = gsc.cv_scores_
+    assert cv_scores_.shape[0] == 16
```
Remi-Gau: ok maybe I am missing something but I would expect the size of this to change with different CV and it seems that the doc string is wrong


```
    cv_scores_ : numpy.ndarray, shape (n_alphas, n_folds)
        scores obtained on test set for each value of the penalization
        parameter explored.
```

nilearn/connectome/tests/test_group_sparse_cov.py
```diff
@@ -160,7 +163,7 @@ def test_group_sparse_covariance_check_consistency_between_classes(rng):
 
     # Check consistency between classes
     gsc1 = GroupSparseCovarianceCV(
-        alphas=4, tol=1e-1, max_iter=20, verbose=0, early_stopping=True
+        tol=1e-1, max_iter=20, verbose=0, early_stopping=True
```
bthirion: What's happening with alphas ? It is not used ?

Remi-Gau: na: 4 is the default value, so I removed it
same with a few other values in some tests

Remi-Gau:
> Mostly LGTM.

@bthirion 
I think this is still an issue though: https://github.com/nilearn/nilearn/pull/5609#discussion_r2428490926

nilearn/connectome/tests/test_group_sparse_cov.py
Remi-Gau: I get the feeling something is off in the way cv_scores_ are collected in this loop:
https://github.com/nilearn/nilearn/blob/ab49ef4892be52ede94dfc7a0d7ddb5606983754/nilearn/connectome/group_sparse_cov.py#L1348

nilearn/decoding/tests/test_space_net.py
```diff
@@ -18,7 +18,13 @@
 from sklearn.metrics import (
     accuracy_score,
 )
-from sklearn.utils.estimator_checks import parametrize_with_checks
+from sklearn.model_selection import (
```
bthirion:
```suggestion
from sklearn.model_selection import KFold
```

nilearn/connectome/tests/test_group_sparse_cov.py
Remi-Gau: will merge as is when I get a green CI and raise an issue about the point above so we can see if this is related to a doc issue or an actual bug

PR #5610, opened 2025-08-26
[REF] remove methods from parent class...

Remi-Gau:
- Closes #

-
-


Remi-Gau:
@hndgzkn starting to address some of your concerns here, but this hightlights common issues in nilearn/decoding/decoders.py and nilearn/decoding/spacenet.py 

- I think further refactoring should be possible when #5595 is merged (as it tries to extract common classes from both of those modules)

Remi-Gau:
further refactoring will be possible but they will be more easy to do once spacenet supports Surface data

nilearn/decoding/space_net.py
```diff
@@ -816,10 +798,10 @@ def fit(self, X, y):
             ["csr", "csc", "coo"],
             dtype=float,
             multi_output=True,
-            y_numeric=not self._is_classification,
+            y_numeric=is_regressor(self),
```
hndgzkn: In the original code `self._check_params` is called twice at line 778 and 785.

nilearn/decoding/space_net.py
```diff
@@ -816,10 +798,10 @@ def fit(self, X, y):
             ["csr", "csc", "coo"],
```
hndgzkn: Here I would suggest to create an abstract method `_check_X_y` for `BaseSpaceNet` :

```python

def _check_X_y(self, X, y):
    raise NotImplementedError()
```

In `SpaceNetClassifier`:

```python

def _check_X_y(self, X, y):
        X, y = check_X_y(
            X,
            y,
            ["csr", "csc", "coo"],
            dtype=float,
            multi_output=True,
            y_numeric=False,
        )
```

In `SpaceNetRegressor`:

```python

def _check_X_y(self, X, y):
        X, y = check_X_y(
            X,
            y,
            ["csr", "csc", "coo"],
            dtype=float,
            multi_output=True,
            y_numeric=True,
        )

        if np.all(np.diff(y) == 0.0):
            raise ValueError(
                "The given input y must have at least 2 targets"
                " to do regression analysis. You provided only"
                f" one target {np.unique(y)}"
            )

```

And in BaseSpaceNet.fit, just call `self._check_X_y(X, y)`.

nilearn/decoding/space_net.py
```diff
@@ -861,34 +843,26 @@ def fit(self, X, y):
         case2 = (alphas is not None) and min(len(l1_ratios), len(alphas)) > 1
         if case1 or case2:
             self.cv_ = list(
-                check_cv(
-                    self.cv, y=y, classifier=self._is_classification
-                ).split(X, y)
+                check_cv(self.cv, y=y, classifier=is_classifier(self)).split(
+                    X, y
+                )
```
hndgzkn, line 863: It seems better to me to avoid using `is_classifier` or `is_regressor` in the base class. The classifier class should know that it's a classifier and implement its own behavior. But such approach might require more refactoring. So I am just writing this as a suggestion:

The steps of fit can be clearly defined. Common functionality could be put in private methods, and each inheriting class can implement its fit using these common methods where necessary and implement differences in between.

nilearn/decoding/space_net.py
```diff
@@ -922,7 +896,7 @@ def fit(self, X, y):
                 solver_params,
                 n_alphas=self.n_alphas,
                 eps=self.eps,
-                is_classif=self._is_classification,
+                is_classif=is_classifier(self),
```
hndgzkn, line 916: The same here.

nilearn/decoding/space_net.py
```diff
@@ -944,8 +918,9 @@ def fit(self, X, y):
         self.cv_scores_ = np.array(self.cv_scores_)
         self.best_model_params_ = np.array(self.best_model_params_)
         self.alpha_grids_ = np.array(self.alpha_grids_)
+
         self.ymean_ /= n_folds
-        if not self._is_classification:
+        if is_regressor(self):
```
hndgzkn: The same here.

nilearn/decoding/space_net.py
```diff
@@ -1002,7 +977,7 @@ def predict(self, X):
             )
 
         # handle regression (least-squared loss)
-        if not self._is_classification:
+        if is_regressor(self):
```
hndgzkn: The same here.

hndgzkn (commented):
@Remi-Gau I've put comments on `space_net.py`. The same is valid for `decoder.py`.

I'll just write my thoughts but it is not necessary to really refactor in that direction. Especially, if it is not likely that new classes will extend these base classes, I am not sure if it is worth the effort.

For fit method of the estimators, I think that would be good to clarify the steps:
For ex:
1. check params
2. fit cache
3. fit masker
4. check x/y
etc.

Then for each step there can be a method in base class that has the common implementation, for changing ones, it can be abstract method to be implemented in concrete class. 
 
Base fit method can call those in order. In that way steps of fit can be clear.

Alternatively base fit method can be abstract and each concrete class can implement its own fit.

nilearn/decoding/space_net.py
Remi-Gau: will require to modify the tests: quite a few of them test BaseSpaceNet and not its children classes: so having some NotImplementedError will lead to test failures

hndgzkn: If we make these changes, we will need to update the tests to test concrete classes as well.

Remi-Gau: OK I have adapted the tests so they only work with concrete classes

nilearn/decoding/space_net.py
Remi-Gau: yeah calling it once is enough

bthirion:
> further refactoring will be possible but they will be more easy to do once spacenet supports Surface data

But I don't think that SpaceNet will ever support SurfaceData. To me the method is  here for legacy, and it's much easier to implement a surface version of FReM than of SpaceNet, for the same benefits.

nilearn/decoding/space_net.py
Remi-Gau: I think there are a few cases in there where it is simpler to leave the is_classifier in there.

Sometimes it is just used by an instance to tell another function (check_feature_screening, check_cv) what they are:
- in some cases we could just pass `self.__sklearn_tags__().estimator_type` to that function and then let that function run adapt its behavior, but in some cases (like check_cv) that function comes from sklearn so we can't change it
- the alternative is indeed to have one method per child class, but I think that beyond what's in this PR, it will lead to the code duplication and more worryingly being harded to read.

Remi-Gau: for example, I have replaced:
`classes = self.classes_ if self._is_classification else self._classes_`

to 
`classes = self._get_classes()`

now anyone who reads the code has to:
- locate all the places where _get_classes() is defined for all possible child classes: here it's not even in the same module because they are defined in the mixin
- figure out (and keep in mind) what this means for each type
- then resume reading the rest of the code

nilearn/decoding/decoder.py
```diff
@@ -1647,30 +1593,6 @@ def __sklearn_tags__(self):
         """
         return super().__sklearn_tags__()
 
-    @fill_doc
-    def fit(self, X, y, groups=None):
```
Remi-Gau, line 1660: moved to mixin class

nilearn/decoding/decoder.py
```diff
@@ -1460,29 +1429,6 @@ def __sklearn_tags__(self):
...
-        """Fit the decoder (learner).
```
Remi-Gau, line 1472: moved to mixin class

nilearn/decoding/_mixin.py
```diff
@@ -49,3 +78,36 @@ def __sklearn_tags__(self):
...
+            The dependent variable (age, sex, IQ, yes/no, etc.).
+            Target variable to predict.
+            Must have exactly as many elements as the input images.
+
+        %(groups)s
+
+        """
+        check_params(self.__dict__)
+        self._classes_ = ["beta"]
+        return super().fit(X, y, groups=groups)
```
hndgzkn, line 104: Moving here might create problems.

nilearn/decoding/decoder.py
```diff
@@ -677,29 +667,21 @@ def fit(self, X, y, groups=None):
             cv_object = LeaveOneGroupOut()
 
         else:
-            cv_object = check_cv(cv, y=y, classifier=self._is_classification)
+            cv_object = check_cv(cv, y=y, classifier=is_classifier(self))
```
hndgzkn, line 669: Starting from line 654 until 672, I thought an abstract method like `_setup_cv` could be used, and implemented in concrete classes.

nilearn/decoding/decoder.py
```diff
@@ -793,7 +775,7 @@ def fit(self, X, y, groups=None):
                 classes_, self.coef_, self.std_coef_
             )
 
-            if self._is_classification and (self.n_classes_ == 2):
+            if is_classifier(self) and (self.n_classes_ == 2):
```
hndgzkn, line 778: Could also the part between lines 754-794 be exracted to an abstract method and be implemented in the concrete class? 

At the same time, I did not understand how is it possible for estimator to be a DummyClassifier or a DummyRegressor. May be what I propose is not convenient.

nilearn/decoding/decoder.py
hndgzkn: Did this work as expected? The call to `super.fit` would run `_BaseDecoder.fit`. But `_BaseDecoder` is not a superclass of mixin.

nilearn/decoding/decoder.py
hndgzkn: The same here.

hndgzkn (commented):
I've added some comments. 

I think the method `more_tags` in decoder concrete classes can be moved to `_BaseDecoder`. It should be different from sklearn tags.

nilearn/decoding/space_net.py
hndgzkn: I understand your point. However I have a different point of view. 

First, looking from the point of view of object oriented design, a parent class should not know about inheriting classes. It should only define the common behavior to all possible inheriting classes, and may be define some abstract methods for the parts where implementation is likely to change depending on the behavior of the inheriting classes. 

In practice, if we assume to know the inheriting classes and implement every possibility in the parent class, with several if clauses, the complexity of code will increase, it will also increase the possibility of introducing bugs. 

If each subclass implements its difference in itself, when a new subclass comes, we can only add the new subclass, without changing the parent one or the other subclasses. However, if we put all code in parent class, then for each new coming class, me might need to add new if conditions in the parent class. 

Isn't it easier to check the concrete class definition to see implementation differences instead of trying to identify them checking several if conditions?

In your code, I thought `classes = self._get_classes()` easy to read. If I need to see the actual implementation for a certain class, I can go and see it in that class implementation, in that specific case, the mixin that it extends.

nilearn/decoding/decoder.py
hndgzkn: @Remi-Gau I had written this comment before seeing your comment above, so do not think that I am insisting :)

nilearn/decoding/space_net.py
bthirion: > I can go and see it in that class implementation, in that specific case, the mixin that it extends.

Yes, but indeed this is very cumbersome. This being said, I don't have any better suggestion than Remi's.

nilearn/decoding/_mixin.py
bthirion: Why ?

hndgzkn: As `RegressorMixin` does not have a parent class, so `super.fit` does not exist, and that would raise error. I understand that these mixins are implemented to be used together with estimators which already has fit method. In that case there will not be a problem. But it would be better to document that.

nilearn/decoding/space_net.py
hndgzkn: When we have if blocks that check type of the object, it is a sign that polymorphism might help. I think we are in that situation. Also there are RegressorMixin and ClassifierMixin classes to implement regressor/classifier specific logic.  Why not put all relevant code there?

Putting if conditions inside one function might seem easier to read (for me it is the opposite), however in long term it leads to complex code, harder to maintain, and open to bugs. Having modular parts for the code also facilitates testing.

bthirion:
some points could be discussed in an upcoming coredev meeting ?

Remi-Gau:
given that we are talking about mixin classes here, this was a timely reminder (at least for me) of the conventions in Python about Mixin classes

https://realpython.com/python-mixin/

.github/workflows/measure_mi.yml
Remi-Gau, line 1: wanted to add this to start measuring if a given PR improves code quality but will do that in a separate PR

Remi-Gau:
@hndgzkn 
will fix the failing tests and will add a TODO in places where further refactoring should be done in a follow up PR

nilearn/decoding/decoder.py
```diff
@@ -792,7 +774,7 @@ def fit(self, X, y, groups=None):
...
```
Remi-Gau, line 778:
```suggestion
            #TODO try to extract
            if is_classifier(self) and (self.n_classes_ == 2):
```

nilearn/decoding/space_net.py
```diff
@@ -854,39 +853,34 @@ def fit(self, X, y):
...
+
         # generate fold indices
         case1 = (None in [alphas, l1_ratios]) and self.n_alphas > 1
         case2 = (alphas is not None) and min(len(l1_ratios), len(alphas)) > 1
         if case1 or case2:
             self.cv_ = list(
-                check_cv(
-                    self.cv, y=y, classifier=self._is_classification
-                ).split(X, y)
+                check_cv(self.cv, y=y, classifier=is_classifier(self)).split(
```
Remi-Gau, line 861:
```suggestion
                #TODO try to extract
                check_cv(self.cv, y=y, classifier=is_classifier(self)).split(
```

nilearn/decoding/space_net.py
```diff
@@ -920,7 +914,7 @@ def fit(self, X, y):
                 solver_params,
                 n_alphas=self.n_alphas,
                 eps=self.eps,
-                is_classif=self._is_classification,
+                is_classif=is_classifier(self),
```
Remi-Gau, line 916:
```suggestion
                #TODO try to extract
                is_classif=is_classifier(self),
```

nilearn/decoding/decoder.py
```diff
@@ -631,13 +620,14 @@ def fit(self, X, y, groups=None):
...
+            Data on which model is to be fitted.
+            If this is a list,
             the affine is considered the same for all.
 
         y : numpy.ndarray of shape=(n_samples) or list of length n_samples
             The dependent variable (age, sex, IQ, yes/no, etc.).
-            Target variable to predict. Must have exactly as many elements as
-            3D images in niimg.
+            Target variable to predict.
+            Must have exactly as many elements as the input images.
```
bthirion:
```suggestion
            Must have exactly as many samples as the input images.
```

Remi-Gau:
> I'm curious about the questions raised by @hndgzkn

@bthirion 
those about the mixins?

nilearn/decoding/_mixin.py
```diff
@@ -26,6 +33,28 @@ def __sklearn_tags__(self):
...
+            return 1
+
+    def _binarize_y(self, y):
+        """Encode target classes as -1 and 1.
+
+        Helper function invoked just before fitting a classifier.
+        """
+        y = np.array(y)
+
+        self._enc = LabelBinarizer(pos_label=1, neg_label=-1)
```
hndgzkn, line 49: `_enc` seems to be not used elsewhere. If that is the case, it might not be necessary to define it as class variable.

nilearn/decoding/_mixin.py
```diff
@@ -26,6 +33,28 @@ def __sklearn_tags__(self):
...
+        y = self._enc.fit_transform(y)
+        self.classes_ = self._enc.classes_
```
hndgzkn, line 51: The classifier uses `self.classes_` and regressor uses `self._classes_`. (It comes from the original code. )  Could it be possible to use single variable?

nilearn/decoding/_mixin.py
```diff
@@ -26,6 +33,28 @@ def __sklearn_tags__(self):
...
+        self.n_classes_ = len(self.classes_)
+        return y
+
+    def _get_classes(self):
+        return self.classes_
+
```
hndgzkn, line 57: If it is possible to unify `self.classes_` and `self._classes_`, this method can be removed.

nilearn/decoding/_mixin.py
```diff
@@ -26,6 +33,28 @@ def __sklearn_tags__(self):
...
+        """Encode target classes as -1 and 1.
+
+        Helper function invoked just before fitting a classifier.
+        """
+        y = np.array(y)
+
+        self._enc = LabelBinarizer(pos_label=1, neg_label=-1)
+        y = self._enc.fit_transform(y)
+        self.classes_ = self._enc.classes_
+        self.n_classes_ = len(self.classes_)
```
hndgzkn, line 52: I do not know `self.classes_` could change elsewhere, but in any case I would suggest either to create a method `n_classes` that will return `len(self.classes_)` when called, or if it is not a variable that is used frequently, I would suggest to call `len(self.classes_)` wherever it is used.

nilearn/decoding/_mixin.py
```diff
@@ -49,3 +78,36 @@ def __sklearn_tags__(self):
...
+        y : numpy.ndarray of shape=(n_samples) or list of length n_samples
+            The dependent variable (age, sex, IQ, yes/no, etc.).
+            Target variable to predict.
+            Must have exactly as many elements as the input images.
+
+        %(groups)s
+
+        """
+        check_params(self.__dict__)
+        self._classes_ = ["beta"]
```
hndgzkn, line 103: `self._classes_ = ["beta"]` might go in `_binarize_y` to be consistent with classifier. 

Actually, `binarize_y` could be named as `set_classes` may be or another name.

If that is done, then `fit` can be removed from `_RegressorMixin`.

nilearn/decoding/decoder.py
```diff
@@ -676,29 +666,21 @@ def fit(self, X, y, groups=None):
...
-            self._binarize_y(y)
-            if self._is_classification
-            else y[:, np.newaxis]
-        )
-
-        if self._is_classification and self.n_classes_ > 2:
-            n_problems = self.n_classes_
-        else:
-            n_problems = 1
+        y = self._binarize_y(y)
```
hndgzkn, line 675: We are using `_binarize_y` in `_Base_Decoder` however, `_BaseDecoder` neither inherits the mixins, nor has a method `_binarize_y`. We can either define abstract methods as in `BaseSpaceNet`, or document that any class that inherits should also inherit from either RegressorMixin, or ClassifierMixin.

nilearn/decoding/space_net.py
```diff
@@ -1397,3 +1397,56 @@ def __init__(
...
+            The dependent variable (age, sex, QI, etc.).
+
+        Notes
+        -----
+        self : `SpaceNet` object
+            Model selection is via cross-validation with bagging.
+        """
+        # duplicated from BaseSpaceNet to not rely on the API / docstring
+        # from RegressorMixin
+        return BaseSpaceNet.fit(self, X, y)
```
hndgzkn, line 1443: If we remove `fit` from `_RegressorMixin`, this part might not be necessary.

bthirion (approved):
The PR LGTM.

tox.ini
```diff
@@ -317,3 +317,11 @@ commands =
...
     python maint_tools/citation_cff_maint.py
 	make -C doc linkcheck
+
+
+[testenv:wily_diff]
+deps =
+    wily
+commands =
+    wily build nilearn
+    wily diff --revision main nilearn
```
Remi-Gau:
```suggestion
```

Remi-Gau:
OK will merge this as is and try the refactoring suggestions suggested by @hndgzkn in follow up PRs

nilearn/decoding/_mixin.py
Remi-Gau: that's not possible: regressors don't have "classes", so we cannot have a public attribute with that name

nilearn/decoding/_mixin.py
Remi-Gau: that worked

nilearn/decoding/_mixin.py
Remi-Gau: nope: see https://github.com/nilearn/nilearn/pull/5610/files#r2369315482

nilearn/decoding/_mixin.py
Remi-Gau: done in a coming PR

nilearn/decoding/_mixin.py
Remi-Gau: removed in coming PR

nilearn/decoding/decoder.py
Remi-Gau: done in a coming PR

nilearn/decoding/decoder.py
Remi-Gau: sorted in a coming PR

PR #5614, opened 2025-08-28
[FIX] ensure consistent output of surface label masker for output signal, region_id and region_name

Remi-Gau:
- Closes #5596 

Changes proposed in this pull request:

- update check_atlas and test to ensure consistent region_name and region_id
- update region_name and region_id properties

PR #5616, opened 2025-08-29
[FIX] handle colorbar plotting for surfaces when vmin == vmax

Remi-Gau:
- Closes none

- fixes an bug encountered when working on https://github.com/nilearn/nilearn/pull/5614


Remi-Gau:
@hndgzkn can I get your thoughts on this?

what should be the behavior of our plotting tools when vmin and vmax are equal?

surface plotting currently fails with for maptplotlib when colorbar=True (with a fairly obscure error), all other cases are fine

I think volume plotting is fine in general.

it is an edge case but I wonder what's the best way to handle this?

only throw an error of our own for surface plotting with for maptplotlib when colorbar=True?
throw an error all the time when vmin and vmax are equal?

Remi-Gau:
ok so far I am going with the simplest solution, reset the min and max of the colorbar to not be equal.

hndgzkn:
> @hndgzkn can I get your thoughts on this?

I think your solution is convenient. I'd propose the same.

Remi-Gau:
ok cool let's merge this mini PR then
thanks @hndgzkn

Issue #5619, opened 2025-08-29
Roadmap 0.13.0


Remi-Gau:
- See milestone: https://github.com/nilearn/nilearn/milestone/30
- Expected release date: october-november 2025 

This release should be made:
- after the EOL of python 3.9
- after python 3.14 has been rolled out

Expected content:
- [x] drop out python 3.9 (#5199)
  - bump matplotlib and other relevant dependencies
 
- [x] support python 3.14 (#5292)
 
- deprecations:

  many : `grep -rni  '# TODO (nilearn >= 0.13.0)' nilearn`

  make sure to:
    - update example where relevant (remove extra parameters that were added to avoid warnings)
    - check that no new warnings is thrown in the tests

---

  list of deprecations:
  - [x] remove "darkness" from plotting function (#5625)
  - [x] copy_header=True (#5656)
  - [x] force_resample=True (#5635)
  - [x] one_mesh_info becomes a private function (#5627)
  - [x] update defaults load_confounds (#5633)
  - [x] rename img to imgs in several methods (#5624)
  - [x] change default output in permuted_ols (#5631)
  - [x] more return values for compute_fixed_effects (#5634)
  - [x] several atlases should return a single map (#5640)
  - [x] keep_masked_maps=False (#5632)
  - [x] stat_type instead of contrast_type (#5630 )
  - [x] t_r instead of tr (#5623)
  - [x] rm `nifti_maps_masker_` property from decomposition estimators (#5626)
  - [x] remove kwargs from the init of several maskers (#5628)
  - [x] #5660
  - [x] #5661
  - [x] #5662
  - [x] #5663
  - [x] #5674

Remi-Gau:
I am going to start working on some of the deprecations so we can start moving on swiftly once 0.12.1 is out.

Remi-Gau:
will close this: all other pending issues and PRs are in the milestone : https://github.com/nilearn/nilearn/milestone/30

PR #3927, opened 2023-08-25
[MAINT] Expose `standardize` parameter in function signatures to allow handling of deprecation

ymzayek:
- Closes #3825

Changes proposed in this pull request:

- Add standardize parameter to `plot_carpet` and `ConnectivityMeasure` so users can deal with "zscore" deprecation
- Update examples
- Update/add tests that warnings are not raised with new input

PR #5629, opened 2025-09-01
[DEP] change default standardize signal

Remi-Gau:
- Relates to #5619
- See also https://github.com/nilearn/nilearn/pull/3927 and https://github.com/nilearn/nilearn/pull/3474

- [ ] change default standardize signal
- [ ] changelog

### DIFF
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -1086,14 +1086,16 @@ docdict["standardize_maskers"] = """
 standardize : {'zscore_sample', 'zscore', 'psc', True, False}, default=False
     Strategy to standardize the signal:
 
-    - ``'zscore_sample'``: The signal is z-scored. Timeseries are shifted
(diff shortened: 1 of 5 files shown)


Remi-Gau:
need to double check but it seems that we need to change the default in a few other places too

Remi-Gau:
OK in general we should probably use FutureWarning instead of DeprecationWarning because some may not appear unless the user runs their code with `python -Wd` or `python -Wall`

Remi-Gau:
OK I think that this could change a LOT more than anticipated.

I think this should introduce new possibility to standarize signals in the decoders (right now it was in theory just True or False, but other values could be passed as long they were those also accepted by the maskers and clean)

So this PR should probably fix the doc as well.

Remi-Gau:
Closing this one.
Extending the depreciation cycle till 0.14.

PR #5637, opened 2025-09-02
[FIX] adapt mask_args depending on masking strategy for (Multi)NiftiMasker.

Remi-Gau:
- Closes none

- fixes a bug that would crash a (Multi)NiftiMasker.fit if the wrong extra arguments were passed to a masking strategy

### DIFF
--- a/nilearn/maskers/multi_nifti_masker.py
+++ b/nilearn/maskers/multi_nifti_masker.py
@@ -3,6 +3,7 @@ on multi subject MRI data.
 """
 
 import collections.abc
+import inspect
 import itertools
 import warnings
 
@@ -322,16 +323,36 @@ class MultiNiftiMasker(NiftiMasker):
             ):
                 imgs = [imgs]
 
-            mask_args = self.mask_args if self.mask_args is not None else {}
             compute_mask = _get_mask_strategy(self.mask_strategy)
-            self.mask_img_ = self._cache(
-                compute_mask,
-                ignore=["n_jobs", "verbose", "memory"],
-            )(
+
+            # add extra argument to pass
+            # to the mask computing function
+            # depending if they are supported.
+            sig = dict(**inspect.signature(compute_mask).parameters)
+            mask_args = {}
+            for arg in ["n_jobs", "target_shape", "target_affine"]:
+                if arg in sig and getattr(self, arg) is not None:
+                    mask_args[arg] = getattr(self, arg)
+            if self.mask_args:
+                skipped_args = []
+                for arg in self.mask_args:
+                    if arg in sig:
+                        mask_args[arg] = self.mask_args.get(arg)
+                    else:
+                        skipped_args.append(arg)
+                if skipped_args:
+                    warnings.warn(
+                        (
+                            "The following arguments are not supported by"
+                            f"the masking strategy '{self.mask_strategy}': "
+                            f"{skipped_args}"
+                        ),
+                        UserWarning,
+                        stacklevel=find_stack_level(),
+                    )
+
+            self.mask_img_ = self._cache(compute_mask, ignore=["verbose"])(
                 imgs,
-                target_affine=self.target_affine,
-                target_shape=self.target_shape,
-                n_jobs=self.n_jobs,
                 memory=self.memory_,
                 verbose=max(0, self.verbose - 1),
                 **mask_args,
--- a/nilearn/maskers/nifti_masker.py
+++ b/nilearn/maskers/nifti_masker.py
@@ -1,5 +1,6 @@
 """Transformer used to apply basic transformations on MRI data."""
 
+import inspect
 import warnings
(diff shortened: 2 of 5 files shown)


nilearn/masking.py
```diff
@@ -770,14 +769,6 @@ def compute_multi_brain_mask(
 
     %(verbose0)s
 
-    .. note::
-        Argument not used but kept to fit the API
-
-    **kwargs : optional arguments
-        Arguments such as 'target_affine' are used in the call of other
-        masking strategies, which then would raise an error for this function
-        which does not need such arguments.
```
Remi-Gau, line 779: removed in favor of a more explicit warnings that some parameters are ignored rather than than silently doing so

nilearn/maskers/multi_nifti_masker.py
```diff
@@ -322,16 +323,36 @@ def fit(
...
+                if skipped_args:
+                    warnings.warn(
+                        (
+                            "The following arguments are not supported by"
+                            f"the masking strategy '{self.mask_strategy}': "
+                            f"{skipped_args}"
+                        ),
+                        UserWarning,
+                        stacklevel=find_stack_level(),
+                    )
```
Remi-Gau: This could be refactored between NiftiMasker and its Multi counterpart but I will do it after the release.

nilearn/maskers/multi_nifti_masker.py
```diff
@@ -322,16 +323,36 @@ def fit(
...
+
+            self.mask_img_ = self._cache(compute_mask, ignore=["verbose"])(
                 imgs,
-                target_affine=self.target_affine,
-                target_shape=self.target_shape,
-                n_jobs=self.n_jobs,
```
Remi-Gau, line 334: Note
it seems that "n_jobs", "target_shape" and "target_affine" can be passed to the masking strategy in the multiniftimasker but not in niftimasker: this seems odd to me

nilearn/maskers/multi_nifti_masker.py
Remi-Gau: reason: see comment below https://github.com/nilearn/nilearn/pull/5637/files#r2315049226

nilearn/maskers/multi_nifti_masker.py
```diff
@@ -322,16 +323,36 @@ def fit(
...
             compute_mask = _get_mask_strategy(self.mask_strategy)
-            self.mask_img_ = self._cache(
-                compute_mask,
-                ignore=["n_jobs", "verbose", "memory"],
-            )(
+
+            # add extra argument to pass
+            # to the mask computing function
+            # depending if they are supported.
+            sig = dict(**inspect.signature(compute_mask).parameters)
```
bthirion:
```suggestion
            signature = dict(**inspect.signature(compute_mask).parameters)
```

nilearn/maskers/tests/test_nifti_masker.py
```diff
@@ -362,6 +362,23 @@ def test_compute_brain_mask(strategy, expected_mask, mask_args):
...
+    masker = NiftiMasker(
+        mask_strategy="background",
+        mask_args={"lower_cutoff": 0.1, "ensure_finite": False},
+    )
+    img, _ = data_gen.generate_random_img((9, 9, 5))
+
+    with pytest.warns(
+        UserWarning, match="The following arguments are not supported by"
+    ):
+        masker.fit(img)
```
Remi-Gau, line 379: given that this failure is present also for NiftiMasker I suspect this bug was not introduced just by a recent PR but was already in the codebase: should justify a new entry in the changelog

Remi-Gau: yup the bug was at least present in 0.12.0: updating changelog

nilearn/maskers/nifti_masker.py
```diff
@@ -508,11 +507,34 @@ def fit(self, imgs=None, y=None):
...
+                    warnings.warn(
+                        (
+                            "The following arguments are not supported by"
+                            f"the masking strategy '{self.mask_strategy}': "
+                            f"{skipped_args}"
+                        ),
+                        UserWarning,
+                        stacklevel=find_stack_level(),
+                    )
+
```
hndgzkn, line 537: This part seems to repeat in both modules. Could this be extracted to a common function?

Remi-Gau: yup the duplicate is intentional for now: 
see https://github.com/nilearn/nilearn/pull/5637#discussion_r2315047137

it seems that there is a slight difference in behavior between the 2 maskers in this PR: for exmaple multiniftimasker will make use of the self.target_affine when generating an implicit mask, but niftimasker does not

I am leaving the duplicate here until I have figured out why (will do that after the release) because this may affect how the refactoring is done

PR #5638, opened 2025-09-02
[ENH] Add regression tests for plotly generated figures

### LINKED ISSUE #5620: Add regression tests for plotly generated figures
We want to avoid unwanted changes in plotly based figures.

For matplotlib we have a series of tests that check that figure look similar to some expected baseline figure.

We need to implement something similar for figure generated by plotly.

Here is a suggestion on how to do it.

https://www.stefsmeets.nl/posts/plotly_pytest/

hndgzkn:
- Closes #5620

- Modify `tox.ini` to install `plotly` and use min dependencies for baseline comparisons
- Modify `nilearn.plotting.displays._figures.PlotlySurfaceFigure.savefig` method signature to be used by pytest-mpl
- Add tests for baseline comparisons for plotly generated figures


hndgzkn:
@Remi-Gau I've found a solution other than your proposed one. I've checked the code for `pytest-mpl` and it uses only `fig.savefig` and as `PlotlySurfaceFigure` has this method, I've made minor changes to take `kwargs` and make it work with `pytest-mpl`.

I've changed `tox.ini` to use `min-deps` for comparisons. Otherwise in my local I encounter problems running the tests due to `numpy v2` and incompatible packages.

tox.ini
```diff
@@ -35,6 +35,8 @@ description = environment with minimum matplotlib version
 skip_install = false
 deps =
     matplotlib==3.3.0
+    plotly
```
hndgzkn: We need to set the version for `plotly`.

tox.ini
```diff
@@ -35,6 +35,8 @@ description = environment with minimum matplotlib version
...
+    kaleido==0.2.1
```
hndgzkn, line 45: As later `kaleido` asks for `chrome`, I've put this version.

.github/workflows/test_with_tox.yml
```diff
@@ -84,7 +84,7 @@ jobs:
                     # TODO revert to latest when dropping py 3.9 and bumping matplotlib
                     os: macos-14
                     env: min
-                -   description: oldest dependencies - no plotly
```
hndgzkn, line 87: As `plotmin` installs `plotly`, I remove this.

tox.ini
Remi-Gau: I think it may worth trying 5.0.0 because of 

https://github.com/nilearn/nilearn/blob/e0b804401217bfa8be984b884242637e9ede09fd/pyproject.toml#L74

Remi-Gau:
FYI this will be only merged after 0.12.1 is out and we will not support python 3.9 after that so expect to have to regenerate the figures

.github/workflows/test_with_tox.yml
Remi-Gau: the problem if we do this is that we then have no workflows that runs the testsuite with matplotlib but not plotly which is needed to make sure we throw the right errors when plotly is missing

so we probably need to still have one tox environment to do this:

for speeding up testing though we could mark those specific tests (`pytest.mark.no_plotly`) and only run those tests only (with `pytest -m no_plotly`)

Remi-Gau: actually when re-reading our pyproject.toml plotly and matpltlib are always installed together 

https://github.com/nilearn/nilearn/blob/14a89340302aaefdad6bc41823ee60d64a13938a/pyproject.toml#L69-L75

so I wonder if we even need have a testing environement with matplotlib but not plotly

bthirion (commented):
LGTM so far...

.github/workflows/test_with_tox.yml
hndgzkn: I think it is good to keep the separate tests. I've added a new one, to only run baseline comparisons when plotly is installed. Could you check if that is OK?

tox.ini
```diff
@@ -36,6 +36,14 @@ skip_install = false
 deps =
     matplotlib==3.8.0
 
+[plot_full_min]
+description = environment with minimum matplotlib, plotly and kaleido versions
+skip_install = false
+deps =
+    matplotlib==3.8.0
```
Remi-Gau:
```suggestion
    {[plotmin]deps}
```

this should work, no?

.github/workflows/test_with_tox.yml
Remi-Gau: to recap, we have:
- environment 'plotting': test all OS, all supported python, latest version of all dependencies plotting included
- environment 'min': macOS, oldest supported python, oldest version of dependencies no plotting
- environment 'plot_min': macOS, oldest supported python, oldest version of dependencies with matplotlib
- environment 'plot_full_min': macOS, oldest supported python, oldest version of dependencies with matplotlib, plotly, kaleido

what's your reasoning for having 'plot_min' AND 'plot_full_min' rather than just having 'plot_full_min': in other words why should we test one engine and not both directly?

hndgzkn:
@Remi-Gau Do you understand why the tests fail? I create plotly images in my local and push them. The tests run without problems in my local. I could not figure out the problem.

Remi-Gau:
> @Remi-Gau Do you understand why the tests fail? I create plotly images in my local and push them. The tests run without problems in my local. I could not figure out the problem.

to be sure, did you generate them with python 3.10 using `tox run -e pytest_mpl_generate` ?

.github/workflows/test_with_tox.yml
hndgzkn: What you said above seemed reasonable to me. 

> the problem if we do this is that we then have no workflows that runs the testsuite with matplotlib but not plotly which is needed to make sure we throw the right errors when plotly is missing

 owever, running baseline comparisons for both `plot_min` and `plot_full_min` might not be necessary. We can run that only for `plot_full_min`. I do not have a strong preference though.

hndgzkn:
> tox run -e pytest_mpl_generate

Yes, I create an environment with `Python=3.10.11` and `tox=4.30.2`. I've recreated from scratch. The result is the same.

Remi-Gau:
OK thanks for trying

I think that we need to be able to see what the figures that are generated in CI look like so we need to upload the artefacts generated by pytest-mpl: otherwise we are "flying blind".

Do you want to try to get that done or do you want me to try to do it?

hndgzkn:
> Do you want to try to get that done or do you want me to try to do it?

Let me try, in case of problems, I'll come back ;)

hndgzkn:
@Remi-Gau The artifacts are [here](https://github.com/nilearn/nilearn/actions/runs/17734627810/artifacts/4012986982). In some images, text is shifted to the left. But I cannot understand the reason.

hndgzkn:
I think this is a problem related to `kaleido` and how it behaves on macos and linux.

I try to check different versions

Remi-Gau:
@hndgzkn 
If that makes things easier, we could try to restrict on what OS the figures generated by plotly are compared to their baseline.
It won't be perfect but it would still be better than nothing.

hndgzkn:
> @hndgzkn If that makes things easier, we could try to restrict on what OS the figures generated by plotly are compared to their baseline. It won't be perfect but it would still be better than nothing.

I am trying with latest version of kaleido (hoping that it is more reliable), and min supported plotly version for that is 6.1.1. If it does not work, or you don't think it is convenient, we can restrict the OS.

hndgzkn:
Finally trying with Ubuntu did not help either. 

I've fixed the font and font size to Arial 18 (fontsize can still be set by the user),  this helped on Ubuntu, but not MacOS. 

After these changes, tests pass with plotly v5.0.0 and kaleido v0.2.1 or plotly v6.1.1 and kaleido v1.0.0 on Ubuntu:latest image.

nilearn/plotting/displays/_figures.py
```diff
@@ -117,20 +117,24 @@ def show(self, renderer="browser"):
             self.figure.show(renderer=renderer)
             return self.figure
 
-    def savefig(self, output_file=None):
+    def savefig(self, output_file=None, **savefig_kwargs):  # noqa: ARG002
```
hndgzkn, line 120: This is changed so that `pytest-mpl` can call `figure.savefig` as for `matplotlib`.

tox.ini
```diff
@@ -170,7 +190,6 @@ passenv = USERNAME
 extras =
     test
 deps =
-    {[plotmin]deps}
```
hndgzkn, line 173: Is there an inconvenience to remove this dependency? As far as I see, `testenv:pytest_mpl` is not used alone, it is reused in other environments where dependencies are already specified.

nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -301,6 +302,43 @@ def test_plot_surf_surface(plot_func, view, hemi):
     )
 
 
+@pytest.mark.skipif(
+    not (is_plotly_installed() and is_kaleido_installed),
```
Remi-Gau:
```suggestion
    not (is_plotly_installed() and is_kaleido_installed()),
```

`is_kaleido_installed` is a function, right?

nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -323,6 +361,34 @@ def test_plot_surf_surface_colorbar(plot_func, colorbar, cbar_tick_format):
...
```
Remi-Gau:
```suggestion
    not (is_plotly_installed() and is_kaleido_installed()),
```

same here

.github/workflows/test_with_tox.yml
```diff
@@ -82,12 +82,15 @@ jobs:
...
                     os: macos-latest
-
                     env: min
                 -   description: oldest dependencies - no plotly
                     py: '3.10'
                     os: macos-latest
                     env: plot_min
+                -   description: oldest dependencies - with plotly
+                    py: '3.10'
+                    os: ubuntu-latest
```
Remi-Gau, line 95: Can you add a comment why we are changing the OS for this one?

tox.ini
```diff
@@ -138,6 +146,18 @@ commands =
     {[testenv:latest]commands}
     {[testenv:pytest_mpl]commands}
 
+[testenv:plot_full_min]
+description = Run tests on minimum version of all dependencies including plotly.
```
Remi-Gau:
```suggestion
description = Run tests to compare figure generated to their expected baseline using minimum version of all dependencies including plotly.
```

Remi-Gau: just to be clear we are not running ALL tests

tox.ini
Remi-Gau: yeah but if you wanted to run it alone locally, it would need those extra dependencies, no?

Remi-Gau: at least I am getting a whole bunch of failure when trying `tox run -e pytest_mpl` on this PR locally

.github/workflows/test_with_tox.yml
Remi-Gau: ok so now:
- plot_min: runs on macOS ALL tests with oldest plotting dependencies but not plotly (include tests that compare generated figures to their baselines)
- plot_full_min: runs on Ubuntu with oldest plotting dependencies including plotly / kaleido ONLY the tests that that compare generated figures to their baselines

is this correct?

Remi-Gau: This is getting hard to keep track of even for us, no?

Remi-Gau: note: I am ok merging this as is and cleaning things up in another PR

tox.ini
hndgzkn: Ok. I'll revert that.

hndgzkn: Done!

.github/workflows/test_with_tox.yml
hndgzkn: Done!

bthirion (approved):
I could not spot any issue. Thx !

Remi-Gau (approved):
Good with me and thanks a bunch for solving this headache

PR #5639, opened 2025-09-02
[DOC] minor fixes post #5578

Remi-Gau:
- Closes none

- addresses comments of #5578


nilearn/glm/tests/test_thresholding.py
```diff
@@ -410,9 +400,25 @@ def test_threshold_stats_img_surface_output(surf_img_1d):
...
+
+def test_threshold_stats_img_surface_output_threshold_0(surf_img_1d):
+    """Check output threshold_stats_img height_control=None, threshold=0."""
+    surf_img_1d.data.parts["left"] = np.asarray([1.0, -1.0, 3.0, 4.0])
+    surf_img_1d.data.parts["right"] = np.asarray([2.0, -2.0, 6.0, 8.0, 0.0])
+
+    # two sided, with threshold = 0
+    result, _ = threshold_stats_img(
+        surf_img_1d, height_control=None, threshold=0, two_sided=False
+    )
```
hndgzkn:
```suggestion
    # one sided, with threshold = 0
    result, _ = threshold_stats_img(
        surf_img_1d, height_control=None, threshold=0, two_sided=False
    )
```

nilearn/glm/tests/test_thresholding.py
```diff
@@ -410,9 +400,25 @@ def test_threshold_stats_img_surface_output(surf_img_1d):
...
+
+    assert_equal(result.data.parts["left"], np.asarray([1.0, 0, 3.0, 4.0]))
+    assert_equal(
+        result.data.parts["right"], np.asarray([2.0, 0, 6.0, 8.0, 0.0])
+    )
+
     # one sided, with threshold = 0
     result, _ = threshold_stats_img(
```
hndgzkn, line 420: The comment above should be two sided. I could not make suggestion. Line 419.

hndgzkn (commented):
It seems like that, the changes made does not fix the colorbar problem for negative values. However the code becomes simpler. 

I think there is a bug somewhere.

nilearn/reporting/glm_reporter.py
```diff
@@ -865,14 +885,10 @@ def _stat_map_to_png(
 
     cmap = DEFAULT_DIVERGING_CMAP
 
-    plot_kwargs = {}
+    plot_kwargs = {"threshold": abs(threshold)}
```
hndgzkn: We even don't need `plot_kwargs`. Isn't it possible to set in function call `threshold=abs(threshold)`?

Remi-Gau: updated

nilearn/glm/tests/test_thresholding.py
Remi-Gau: updated

nilearn/glm/tests/test_thresholding.py
Remi-Gau: updated

Remi-Gau:
> bug somewhere in plotting.

I think that this should be fixed in follow up PR.

PR #5640, opened 2025-09-02
[DEP] deprecate certain outputs of datasets

Remi-Gau:
- Relates to #5619

- [x] ensure datasets return a sklearn Bunch with a single maps under the key 'maps'
- [x] changelog


nilearn/datasets/atlas.py
```diff
@@ -1402,36 +1275,13 @@ def fetch_atlas_yeo_2011(
...
-                deprecation_message.format(version="0.13.0")
-                + (
-                    "To suppress this warning, "
-                    "Please use the parameters 'n_networks' and 'thickness' "
-                    "to specify the exact atlas image you want."
-                )
-            ),
-            stacklevel=find_stack_level(),
+    if n_networks not in (7, 17):
+        raise ValueError(f"'n_networks' must be 7 or 17. Got {n_networks=}")
```
Remi-Gau: there are >40 places in the code base where we have this similar pattern of
check if paramater is in allowed_set, if not raise ValueError

I think we could extract this into a standardized function

Remi-Gau: done

PR #5653, opened 2025-09-05
[MAINT] reduce warnings in doc

Remi-Gau:
- Closes none

- specify screening "percentile = 100" when using an explicit mask in decoding to avoid warnings being thrown


Remi-Gau:
> IIUC this does not change the numerical behavior since a small ROI is used. Am I right ?

exactly: we were getting warning of the type "using 100% of voxels even if screening percentile was set to 20%"

PR #5654, opened 2025-09-05
[DOC] add "nilearn" in version directive

Remi-Gau:
- Closes none 

- helps to distinguish version mentioned in method owned by nilearn and owned by dependencies (e. sklearn) - see https://github.com/scikit-learn/scikit-learn/issues/32045


Remi-Gau:
hmm... something better for this would be to replace the basic sphinx directive with our own:

https://stackoverflow.com/questions/39723122/is-it-possible-to-replace-one-directive-with-another-one

Remi-Gau:
from niftimasker doc that contains both version change information coming from nilearn and sklearn

image


image

equivalent page in the stable doc: https://nilearn.github.io/stable/modules/generated/nilearn.maskers.NiftiMasker.html#nilearn.maskers.NiftiMasker

Remi-Gau:
OK so I decided to go with a different strategy that should be more flexible in the long run: we now have our own directive for version change, deprecation...

bthirion:
It populates the zoo of Nilearn-specific stuff, but it's indeed easy for contributors to just copy and paste.

Remi-Gau:
> It populates the zoo of Nilearn-specific stuff, but it's indeed easy for contributors to just copy and paste.

yeah also because now we only have `nilearn_versionchanged` in the code base, it's easier to remember that's the one we need to use

doc/conf.py
```diff
@@ -456,7 +458,57 @@ def touch_example_backreferences(
...
+    def run(self):
+        """Run directive."""
+        return super().run()
+
+
 def setup(app):
+    app.add_directive("nilearn_versionadded", NilearnVersionAdded)
+    app.add_directive("nilearn_versionchanged", NilearnVersionChanged)
+    app.add_directive("nilearn_deprecated", NilearnVersionDeprecated)
+    app.add_directive("nilearn_versionremoved", NilearnVersionRemoved)
```
hndgzkn: Instead of defining these 4 classes, isn't it possible to use `changeset.VersionChange`? All classes have the same content except class name, and they call super class method. I did not see them being used elsewhere.

CONTRIBUTING.rst
```diff
@@ -239,8 +239,10 @@ with the tools we use for development and deployment.
 |   `Coding Style`_  |    Any        | - No new dependency                                 |
 |                    |               | - Backward compatibility                            |
 |                    |               | - All internal imports are absolute, not relative   |
-|                    |               | - Impacted docstrings have versionadded and/or      |
-|                    |               |   versionchanged directives as needed.              |
+|                    |               | - Impacted docstrings have                          |
+|                    |               |   ``.. nilearn_versionadded`` and/or                |
+|                    |               |   ``.. nilearn_versionchanged`` directives          |
+|                    |               |   as needed.                                        |
```
hndgzkn, line 245: Are `nilearn_deprecated` and `nilearn_versionremoved` missing here?

hndgzkn (approved):
LGTM apart from some comments.

doc/conf.py
Remi-Gau: let me check

CONTRIBUTING.rst
Remi-Gau: yeah but they are also a lot less frequent

doc/conf.py
Remi-Gau: it works: removing the classes

Remi-Gau:
hmmm... realizing that if any downstream package tries to inherit from our classes and tries to render their doc, they may get some errors in their doc build because they won't know those new directives

we should at least add this in the changelog

CONTRIBUTING.rst
hndgzkn: But it could be good to know those directives exist.

Remi-Gau:
> hmmm... realizing that if any downstream package tries to inherit from our classes and tries to render their doc, they may get some errors in their doc build because they won't know those new directives
> 
> we should at least add this in the changelog

done

PR #5656, opened 2025-09-08
[DEP] set copy_header default to True

Remi-Gau:

- [x] set copy_header default to True
- [x] set type of copy_header to bool to facilitate its validation
- [x] remove all places where `copy_header=True` as it is now the new default
- [x] update doc strings
- [x] adapt / simplify tests to make sure copy_header is properly tested with True or False
- [x] changelog

### DIFF
--- a/nilearn/typing.py
+++ b/nilearn/typing.py
@@ -73,7 +73,7 @@ TargetAffine: TypeAlias = ndarray | list | None
 # Note that this is usable as for static type checking,
 # as type checkers will complain
 # about using a generic and would prefer "list[int]" to "list".
-TargetShape: TypeAlias = tuple | list | None
+TargetShape: TypeAlias = tuple | list | ndarray | None
 
 Threshold: TypeAlias = float | int | np.floating | np.integer | str | None
 


asv_benchmarks/benchmarks/common.py
```diff
@@ -39,14 +39,12 @@ def setup_cache(self, n_subjects=10, n_masks=1):
             atlas_path,
             concat,
             interpolation="nearest",
-            force_resample=True,
```
Remi-Gau, line 42: There were a few force_resample=False left

nilearn/masking.py
```diff
@@ -701,10 +701,7 @@ def compute_brain_mask(
         )
 
     resampled_template = cache(resampling.resample_to_img, memory)(
-        template,
-        target_img,
-        copy_header=True,
-        force_resample=False,  # TODO (nilearn >= 0.13.0) update to True
```
man-shu, line 720: Ok so the test failures are because of changing force_resample = False to True.

@Remi-Gau, do we have a good reason why we want to change to True even when we explicitly set it to False? 
All I could find so far was that the implementation with False is wrong ([here](https://github.com/nilearn/nilearn/pull/4412)), but not the reasoning behind it.

man-shu: Ok, so I have just updated the test to not expect all zeros in the mask now. 
Since we have deprecated this properly, ideally users should be expecting changes in masking outputs...

Remi-Gau:
test failure comes from a flaky test
merging main to restart the tests

nilearn/image/image.py
```diff
@@ -1008,7 +988,8 @@ def threshold_img(
         maximum intensity of the data.
         If `two_sided` is True, threshold cannot be negative.
         If threshold is :obj:`str`,
-        the given string should be within the range of "0%" to "100%".
+        the given string should be
+        within the range of ``"0%%"`` to ``"100%%"``.
```
hndgzkn, line 992: I could not understand the reason for this change.

nilearn/image/image.py
```diff
@@ -1383,11 +1359,11 @@ def binarize_img(
         voxels which have intensities greater than this value will be kept.
         The given value should be within the range of minimum and
         maximum intensity of the input image.
-        If string, it should finish with percent sign e.g. "80%" and we
+        If string, it should finish with percent sign e.g. "80%%" and we
         threshold based on the score obtained using this percentile on
         the image data. The voxels which have intensities greater than
         this score will be kept. The given string should be
-        within the range of "0%" to "100%".
+        within the range of "0%%" to "100%%".
```
hndgzkn, line 1366: Why double percent sign here as well?

hndgzkn (approved):

nilearn/image/image.py
Remi-Gau: that's to escape the first percent sign is there to escape the 2nd percent sign during sphinx rendering... arcane stuff that has made me lose hours of my life and part of my sanity

nilearn/image/image.py
Remi-Gau: sorry
I have a tendency to sometimes add some semantic line breaks: https://sembr.org/

Remi-Gau:
I cannot approve my own PR but I approve the changes made by @man-shu

examples/06_manipulating_images/plot_affine_transformation.py
```diff
@@ -102,33 +102,29 @@
 from nilearn.image import resample_img
 
 img_in_mm_space = resample_img(
-    img, target_affine=np.eye(4), target_shape=(512, 512, 1), copy_header=True
+    img,
+    target_affine=np.eye(4),
+    target_shape=(512, 512, 1),
```
emdupre, line 107: Do we want this trailing comma to explode the line ? We don't in the subsequent `resample_img` calls, so double-checking.

emdupre:
> Great PR. I only have one small comment.

@bthirion it looks like we're missing your comment ? The ping showed me I hadn't completed my review, so I went ahead and approved. But I don't mean to take away from your feedback !

nilearn/maskers/tests/test_multi_nifti_masker.py
```diff
@@ -184,13 +184,11 @@ def test_mask_strategy_errors(list_random_imgs):
 )
 def test_compute_mask_strategy(strategy, shape_3d_default, list_random_imgs):
     """Check different strategies to compute masks."""
-    masker = MultiNiftiMasker(mask_strategy=strategy, mask_args={"opening": 1})
+    masker = MultiNiftiMasker(mask_strategy=strategy)
```
bthirion, line 187: I was wondering why you changed that?

bthirion (commented):
Sorry, here is my comment.

examples/06_manipulating_images/plot_affine_transformation.py
Remi-Gau: I think the multi-line is fine: if anything it may avoid some horizontal scrolling on smaller devices:

image

I think that if we wanted to be consistent in terms of mobile reading friendliness we'd enforce even shorter lines length in the doc but I think that'd be pushing things a bit far for no obvious gains (not sure how many people read the doc on their phone).

For now I can personally live with some inconsistencies #FML.

Remi-Gau: also horizontal scrolling is easier to do on mobile

nilearn/maskers/tests/test_multi_nifti_masker.py
Remi-Gau: cannot 100% remember

I think I was in mode "remove extra parameter from test"

The fact that it can be removed without anything crashing makes me think the extra parameter is not doing much: not sure if we added it for coverage only (doubtful as mask_args is tested in other places)

nilearn/image/resampling.py
```diff
@@ -688,6 +686,18 @@ def _check_resample_img_inputs(target_shape, target_affine, interpolation):
...
+            "'force_resample' must be a boolean."
+            f"Got: {force_resample.__class__.__name__}"
+        )
+
+    if not isinstance(copy_header, bool):
+        raise TypeError(
+            "'copy_header' must be a boolean."
+            f"Got: {copy_header.__class__.__name__}"
+        )
+
```
Remi-Gau: We do not have test for those.

Remi-Gau: will add them.

Remi-Gau: actually these lines were not needed because we can rely on check_params to check the type of force_resample and copy_header at run time

nilearn/typing.py
```diff
@@ -66,7 +68,7 @@
 SmoothingFwhm: TypeAlias = float | int | np.floating | np.integer | None
 Standardize: TypeAlias = Literal["zscore", "zscore_sample", "psc", True, False]
 StandardizeConfounds: TypeAlias = bool
-TargetAffine: TypeAlias = ndarray | None
+TargetAffine: TypeAlias = ndarray | list | None
```
Remi-Gau, line 71: turns out from some tests that affine can be a list of numbers and not just an array: not this is true for ALL the places where we use TargetAffine: to check in a follow up

nilearn/image/image.py
hndgzkn: Ok, I did not know about that. Seems reasonable. Thanks!

nilearn/maskers/tests/test_multi_nifti_masker.py
man-shu: I think I made this change. Basically, setting `force_resample` to True in the resample function has changed the expected "dummy" mask outputs.

We expected all zeros in the resulting mask, but having this parameter resulted in some 1s only with "wm" masking strategy.

To me, the test is more of a smoke test, and I don't see the reason to have the opening argument there. (The input image is just a bunch of random numbers, so I am not sure how one could predict the output mask)

Remi-Gau:
and we are good to go!

PR #5660, opened 2025-09-09
[DEP] disallow surface ROI with negative or non integer values

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/pull/4131

- [x] turn deprecation warnings into ValueErrors
- [x] changelog


bthirion (commented):
LGTM

nilearn/plotting/tests/baseline/test_plot_surf_surface_both-anterior-plot_surf_roi.png
bthirion, line 1: the new figures are terrible...

Remi-Gau: yeah that's because I am plotting some continous stuff 'cortical thickness' that I have binarized...
can try something else

Remi-Gau: ok using sulcal instead of thickness will look better

Remi-Gau: figures should look better now

Remi-Gau:
> I'm not completely sure what is the intent with the plot change.

just that plot_surface_roi only accepts positive integers now, so some of the previous test for plotting are now failing because we used surface maps with floats, so this needed changing.

bthirion:
> > I'm not completely sure what is the intent with the plot change.
> 
> just that plot_surface_roi only accepts positive integers now, so some of the previous test for plotting are now failing because we used surface maps with floats, so this needed changing.

OK, but why did we change from pial to inflated  surface ? Because it looks better ?

Remi-Gau:
> Because it looks better ?

kinda: the idea is , that more details is visible on the inflated surface, so if in the future something changes in the figure it will be easier to catch

bthirion (approved):
LGTM, thx.

PR #5663, opened 2025-09-09
[DEP] parse_bids_file return dict with bids keys

Remi-Gau:

- [x] parse_bids_file return dict with bids keys
- [x] update test to make sure that "+" are allowed in entity labels
- [x] changelog

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -45,3 +45,5 @@ Changes
 - :bdg-danger:`Deprecation` The parameter ``output_type`` of :func:`~mass_univariate.permuted_ols` was changed from ``legacy`` to ``dict``. The parameter ``output_type`` will be removed in Nilearn >= 0.15.0 (:gh:`5631` by `Rémi Gau`_).
 
 - :bdg-danger:`Deprecation` The parameter ``contrast_type`` of  :func:`~glm.compute_contrast` and :class:`~glm.Contrast` has been replaced by ``stat_type`` (:gh:`5630` by `Rémi Gau`_).
+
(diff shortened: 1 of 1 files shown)


nilearn/interfaces/tests/test_bids.py
```diff
@@ -379,25 +379,12 @@ def test_get_bids_files_no_space_entity(tmp_path):
 def test_parse_bids_filename():
     """Check that a typical BIDS file is properly parsed."""
     fields = ["sub", "ses", "task", "lolo"]
-    labels = ["01", "01", "langloc", "lala"]
-    file_name = "sub-01_ses-01_task-langloc_lolo-lala_bold.nii.gz"
+    labels = ["01", "01", "langloc+foo", "lala"]
+    file_name = "sub-01_ses-01_task-langloc+foo_lolo-lala_bold.nii.gz"
```
Remi-Gau, line 383: + are now allowed in entity label, making sure we are OK with this.

bthirion: Fine with me, as long as this is BIDS compliant.

bthirion (commented):
The change is OK for me.

nilearn/interfaces/tests/test_bids.py
Remi-Gau: that was very recently allowed: we are not in the business of doing BIDS validation, but once in a while it does not hurt to make sure we are not more strict that BIDS is. 😉

hndgzkn (approved):

bthirion (approved):

Issue #5646, opened 2025-09-03
create function to check if a parameter value is in allowed list


Remi-Gau:
there are >40 places in the code base where we have this similar pattern of
check if paramater is in allowed_set, if not raise ValueError

I think we could extract this into a standardized function

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5640#discussion_r2315925922_

PR #5665, opened 2025-09-09
[REF] extract function to check "value is of allowed type"

Remi-Gau:
- Closes none

- follow up on #5646 
- extract function to check value is of allowed type


bthirion (commented):
LGTM so far.

PR #5666, opened 2025-09-09
[ENH] refactor 3

hndgzkn:
- Closes #

-
-

### DIFF
--- a/nilearn/plotting/_utils.py
+++ b/nilearn/plotting/_utils.py
@@ -86,11 +86,12 @@ def get_cbar_ticks(vmin, vmax, threshold=None, n_ticks=5):
     if vmax == 0:
         vmax += np.finfo(np.float32).eps
 
+    ticks = np.linspace(vmin, vmax, n_ticks)
+
     # If a threshold is specified, we want two of the tick
     # to correspond to -threshold and +threshold on the colorbar.
     # If the threshold is very small compared to vmax,
     # we use a simple linspace as the result would be very difficult to see.
-    ticks = np.linspace(vmin, vmax, n_ticks)
     if threshold is not None and threshold / vmax > 0.12:
         diff = [abs(abs(tick) - threshold) for tick in ticks]
         # Edge case where the thresholds are exactly
--- a/nilearn/plotting/_utils.py
+++ b/nilearn/plotting/_utils.py
@@ -55,7 +55,7 @@ def save_figure_if_needed(fig, output_file):
     return None
 
 
-def get_cbar_ticks(vmin, vmax, offset, n_ticks=5):
+def get_cbar_ticks(vmin, vmax, threshold, n_ticks=5):
     """Help for BaseSlicer."""
     # edge case where the data has a single value yields
     # a cryptic matplotlib error message when trying to plot the color bar
@@ -71,8 +71,8 @@ def get_cbar_ticks(vmin, vmax, offset, n_ticks=5):
     # If the threshold is very small compared to vmax,
     # we use a simple linspace as the result would be very difficult to see.
     ticks = np.linspace(vmin, vmax, n_ticks)
-    if offset is not None and offset / vmax > 0.12:
-        diff = [abs(abs(tick) - offset) for tick in ticks]
+    if threshold is not None and threshold / vmax > 0.12:
+        diff = [abs(abs(tick) - threshold) for tick in ticks]
         # Edge case where the thresholds are exactly
         # at the same distance to 4 ticks
         if diff.count(min(diff)) == 4:
@@ -84,7 +84,7 @@ def get_cbar_ticks(vmin, vmax, offset, n_ticks=5):
             if 0 in ticks[idx_closest]:
                 idx_closest = np.sort(np.argpartition(diff, 3)[:3])
                 idx_closest = idx_closest[[0, 2]]
-        ticks[idx_closest] = [-offset, offset]
+        ticks[idx_closest] = [-threshold, threshold]
     if len(ticks) > 0 and ticks[0] < vmin:
         ticks[0] = vmin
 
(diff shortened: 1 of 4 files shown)


nilearn/plotting/_utils.py
```diff
@@ -66,13 +86,14 @@ def get_cbar_ticks(vmin, vmax, offset, n_ticks=5):
...
+
     # If a threshold is specified, we want two of the tick
     # to correspond to -threshold and +threshold on the colorbar.
     # If the threshold is very small compared to vmax,
     # we use a simple linspace as the result would be very difficult to see.
-    ticks = np.linspace(vmin, vmax, n_ticks)
-    if offset is not None and offset / vmax > 0.12:
-        diff = [abs(abs(tick) - offset) for tick in ticks]
+    if threshold is not None and threshold / vmax > 0.12:
+        diff = [abs(abs(tick) - threshold) for tick in ticks]
```
hndgzkn: Here just moved `ticks = np.linspace(vmin, vmax, n_ticks)` above the comment and renamed `offset` to `threshold`.

nilearn/plotting/displays/_slicers.py
```diff
@@ -739,26 +740,16 @@ def _show_colorbar(
...
-        n_ticks = 5 if cbar_vmin != cbar_vmax else 1
-        ticks = get_cbar_ticks(cbar_vmin, cbar_vmax, offset, n_ticks)
-        bounds = np.linspace(cbar_vmin, cbar_vmax, our_cmap.N)
-
-        # some colormap hacking
-        cmaplist = [our_cmap(i) for i in range(our_cmap.N)]
-        transparent_start = int(norm(-offset, clip=True) * (our_cmap.N - 1))
-        transparent_stop = int(norm(offset, clip=True) * (our_cmap.N - 1))
-        for i in range(transparent_start, transparent_stop):
-            cmaplist[i] = (*self._brain_color, 0.0)  # transparent
```
hndgzkn, line 754: This is actually `threshold_cmap` except lines 747-748 which are moved below.

hndgzkn:
@Remi-Gau @bthirion I would like to keep this PR small for an easy review and quick merge. The next step will be trying to unify `get_cbar_ticks` and `get_cropped_cbar_ticks` clarifying expected behavior.

nilearn/plotting/_utils.py
```diff
@@ -55,8 +55,28 @@ def save_figure_if_needed(fig, output_file):
...
+        if threshold is specified, -threshold and threshold is expected to
+    appear on the colorbar
+    n_ticks: :obj:`int`
+        number of tick values to return
+
+    Returns
+    -------
+    :class:`~numpy.ndarray`
+        an array with ``n_ticks`` elements if ``vmin`` != ``vmax``, else array
+    with one element.
```
Remi-Gau:
```suggestion
        an array with ``n_ticks`` elements if ``vmin`` != ``vmax``, else array
        with one element.
```

nilearn/plotting/_utils.py
```diff
@@ -55,8 +55,28 @@ def save_figure_if_needed(fig, output_file):
...
+    """Return an array of ``n_ticks`` tick values to be used for the
+    colorbar.
+
+    Parameters
+    ----------
+    vmin: :obj:`float`
+        minimum value for the colorbar
+    vmax: :obj:`float`
+        maximum value for the colorbar
+    threshold: :obj:`float` or :obj:`int`
```
Remi-Gau:
```suggestion
    threshold: :obj:`float`, :obj:`int` or None
```

probably worth mentioning what happens when it's None

Remi-Gau (approved):
minor comments on the doc string but GTM otherwise

nilearn/plotting/_utils.py
hndgzkn: Updated the docstring for `get_cbar_ticks` as:

```
Return an array of evenly spaced ``n_ticks`` tick values to be used for
    the colorbar.
```

and for threshold:

```
if threshold is not None, ``-threshold`` and ``threshold`` values are replaced with the closest tick values
```

Would that be ok?

Remi-Gau: yup

nilearn/plotting/_utils.py
```diff
@@ -55,24 +55,46 @@ def save_figure_if_needed(fig, output_file):
...
     # a cryptic matplotlib error message when trying to plot the color bar
     if vmin == vmax:
         return np.linspace(vmin, vmax, 1)
 
-    # edge case where the data has all negative values but vmax is exactly 0
+    # # edge case where the data has all negative values but vmax is exactly 0
+    vmax_temp = vmax
     if vmax == 0:
-        vmax += np.finfo(np.float32).eps
+        vmax_temp = np.finfo(np.float32).eps
```
hndgzkn, line 88: Here I do not change the actual value of `vmax`. I think it is only changed to avoid zero division in line 96. We do not need to change the value, using a temp value would do it. 

@Remi-Gau Do you see an inconvenience? The tests pass in local including baseline comparisons. (Better to see if they pass on CI as well.)

bthirion (approved):
LGTM, thx.

nilearn/plotting/_utils.py
Remi-Gau: yeah that makes sense

nilearn/plotting/_utils.py
```diff
@@ -55,24 +55,46 @@ def save_figure_if_needed(fig, output_file):
...
+        an array with ``n_ticks`` elements if ``vmin`` != ``vmax``, else array
+        with one element.
+    """
     # edge case where the data has a single value yields
     # a cryptic matplotlib error message when trying to plot the color bar
     if vmin == vmax:
         return np.linspace(vmin, vmax, 1)
 
-    # edge case where the data has all negative values but vmax is exactly 0
+    # # edge case where the data has all negative values but vmax is exactly 0
```
Remi-Gau:
```suggestion
    # edge case where the data has all negative values but vmax is exactly 0
```

PR #5668, opened 2025-09-10
[FIX] fix documentation about standardize

### LINKED ISSUE #5648: [DOC] standardize parameter for decoders (and some other classes) should just be a bool ?

- This stems from working on https://github.com/nilearn/nilearn/pull/5629.

Several classes have a standardize parameters that according to the doc only accepts a boolean.

For example the decoders

https://nilearn.github.io/stable/modules/generated/nilearn.decoding.Decoder.html#nilearn.decoding.Decoder

However, this parameter is then passed to either a masker that then passes it to signal.clean (the decoders do that) or this is directly passed to signal.clean (ConnectivityMeasure does that).

But signal.clean "standardize" can also accept things like "zscore" and "zscore_sample".

Actually there are several instances in the example where the decoders are initialised with "zscore_sample".

So either this is a documentation issue and "standardize" should accept the same values everywhere OR we only allow "standardize" to be a boolean but then we should enforce this.

```
grep -rni  '%(standardize)s' nilearn

nilearn/plotting/image/img_plotting.py:1923:    %(standardize)s
nilearn/decomposition/canica.py:57:    %(standardize)s
nilearn/decomposition/dict_learning.py:93:    %(standardize)s
nilearn/connectome/connectivity_matrices.py:403:    %(standardize)s
nilearn/decoding/decoder.py:523:    %(standardize)s
nilearn/decoding/decoder.py:1194:    %(standardize)s
nilearn/decoding/decoder.py:1370:    %(standardize)s
nilearn/decoding/decoder.py:1561:    %(standardize)s
nilearn/decoding/decoder.py:1754:    %(standardize)s
```

Remi-Gau:
- Closes #5648 

- most places where there is a standardize parameter, it will be passed to a masker and then to signal.clean
- so allowed values for standardize should be the same all the way through

### DIFF
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -1040,19 +1040,21 @@ standardize : :obj:`bool`, default={}
     in the time dimension.
 """
 docdict["standardize"] = standardize.format("True")
-docdict["standardize_false"] = standardize.format("False")
+
 
 # standardize as used within maskers module
-docdict["standardize_maskers"] = """
-standardize : {'zscore_sample', 'zscore', 'psc', True, False}, default=False
+standardize_maskers = """
+standardize : any of: 'zscore_sample', 'zscore', 'psc', True, False; default={}
     Strategy to standardize the signal:
 
-    - ``'zscore_sample'``: The signal is z-scored. Timeseries are shifted
-      to zero mean and scaled to unit variance. Uses sample std.
+    - ``'zscore_sample'``: The signal is z-scored.
+      Timeseries are shifted to zero mean and scaled to unit variance.
+      Uses sample std.
 
-    - ``'zscore'``: The signal is z-scored. Timeseries are shifted
-      to zero mean and scaled to unit variance. Uses population std
-      by calling default :obj:`numpy.std` with N - ``ddof=0``.
+    - ``'zscore'``: The signal is z-scored.
+      Timeseries are shifted to zero mean and scaled to unit variance.
+      Uses population std by calling default
+      :obj:`numpy.std` with N - ``ddof=0``.
 
     - ``'psc'``:  Timeseries are shifted to zero mean value and scaled
       to percent signal change (as compared to original mean signal).
@@ -1063,6 +1065,8 @@ standardize : {'zscore_sample', 'zscore', 'psc', True, False}, default=False
     - ``False``: Do not standardize the data.
 
 """
+docdict["standardize_maskers_false"] = standardize.format("False")
+docdict["standardize_maskers_zscore"] = standardize.format("zscore")
 
 # standardize_confounds
 docdict["standardize_confounds"] = """
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -1034,16 +1034,6 @@ smoothing_fwhm : :obj:`float` or :obj:`int` or None, optional.
 
 # standardize
 standardize = """
-standardize : :obj:`bool`, default={}
-    If `standardize` is `True`, the data are centered and normed:
-    their mean is put to 0 and their variance is put to 1
-    in the time dimension.
-"""
-docdict["standardize"] = standardize.format("True")
-
-
-# standardize as used within maskers module
-standardize_maskers = """
 standardize : any of: 'zscore_sample', 'zscore', 'psc', True, False; default={}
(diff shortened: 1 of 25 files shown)


Remi-Gau:
OK as expected the tests are now drowning in warnings:

image

nilearn/connectome/group_sparse_cov.py
```diff
@@ -750,6 +750,7 @@ def empirical_covariances(subjects, assume_centered=False, standardize=False):
     # single precision to double will be required or not.
     emp_covs = np.empty((n_features, n_features, n_subjects), order="F")
     for k, s in enumerate(subjects):
+        # TODO should we use sample std?
```
bthirion, line 753: the time courses should be standardized to have unit variance.

bthirion (commented):
LGTM overall.

Remi-Gau:
still too many warnings in the test: will try to make sure we get fewer.

nilearn/connectome/group_sparse_cov.py
Remi-Gau: This is somewhat unrelated to this PR, so will open a separate issue.

Remi-Gau:
OK I will leave this PR at this for now: I will add tests to increase "behavior" coverage in a follow up PR.

nilearn/_utils/docs.py
```diff
@@ -1034,25 +1034,17 @@ def custom_function(vertices):
...
-    their mean is put to 0 and their variance is put to 1
-    in the time dimension.
-"""
-docdict["standardize"] = standardize.format("True")
-docdict["standardize_false"] = standardize.format("False")
-
-# standardize as used within maskers module
-docdict["standardize_maskers"] = """
-standardize : {'zscore_sample', 'zscore', 'psc', True, False}, default=False
+standardize : any of: 'zscore_sample', 'zscore', 'psc', True, False; default={}
```
hndgzkn, line 1037: Instead of mixing boolean and string values, would it be possible to define only possible string values and `None` if we do not want to standardize the data?

nilearn/_utils/param_validation.py
```diff
@@ -261,11 +262,23 @@ def fit(X):
         type_to_check = TYPE_MAPS[k]
         value = fn_dict[k]
 
-        error_msg = (
-            f"'{k}' should be of type '{type_to_check}'.\nGot: '{type(value)}'"
-        )
-        if not isinstance(value, type_to_check):
-            raise TypeError(error_msg)
+        if get_origin(type_to_check) is Literal:
+            allowed_values = normalize_type(type_to_check)
```
hndgzkn: ` if get_origin(type_to_check) is Literal:` check is both done here and in `normalize_type`. Instead of creating  `normalize_type` function and calling here, you can call `get_arg(type_to_check)` here. 

Or if it will be a rather frequent check, lines 265-275 could be extracted as a function.

nilearn/_utils/docs.py
```diff
@@ -1034,25 +1034,17 @@ def custom_function(vertices):
...
+    - ``'zscore_sample'``: The signal is z-scored.
+      Timeseries are shifted to zero mean and scaled to unit variance.
+      Uses sample std.
 
-    - ``'zscore'``: The signal is z-scored. Timeseries are shifted
-      to zero mean and scaled to unit variance. Uses population std
-      by calling default :obj:`numpy.std` with N - ``ddof=0``.
+    - ``'zscore'``: The signal is z-scored.
+      Timeseries are shifted to zero mean and scaled to unit variance.
+      Uses population std by calling default
```
hndgzkn, line 1046: May be a deprecation notice can be put here as well as `z_score` will be deprecated.

nilearn/_utils/docs.py
Remi-Gau: We could start another deprecation cycle for this.
Note 'zscore' will be deprecated in the near future.
So in the end we would have:
- 'zscore_sample' 
- 'psc' 
- None

nilearn/_utils/param_validation.py
Remi-Gau: done

nilearn/_utils/docs.py
Remi-Gau: done

Issue #5539, opened 2025-07-11
introduce multi surface maskers


Remi-Gau:
Oening a discussion here because it may be useful to have surface counterparts to the multi nifti maskers.

Masker objects can transform both 3D and 4D image objects :
- transforming a 3D image produces a 1D (features,) array,
- transforming a list of 3D image produces a 2D (n images, features) array.
- transforming a 4D image produces a 2D (samples, features) array.

MultiMasker objects can transform both 3D, 4D, as well as list of 3D or 4D image objects :

- transforming a 3D image produces a 1D (features,) array, 
- transforming a 4D image produces a 2D (samples, features) array,
- transforming a list of 3D image produces a list of 1D (features,) array,
- transforming a list of 4D image produces a list of 2D (samples, features) array.

But surface maskers work a bit differently:

- transforming a 1D image produces a 1D (features,) array, 
- transforming a 2D image produces a 2D (samples, features) array,
- transforming a list of 1D image produces a 2D (number of images in list, features) array,
- transforming a list of 4D image produces a list of 2D (sum of samples across all images in list, features) array.

quick example:

```python
from nilearn.maskers import SurfaceMasker, NiftiMasker, MultiNiftiMasker
from nilearn.conftest import _make_surface_img
from nilearn._utils.data_gen import generate_fake_fmri
import numpy as np

# SurfaceMasker with list of surface image with a single sample
imgs = [_make_surface_img(1), _make_surface_img(1)]
masker = SurfaceMasker()
signals = masker.fit_transform(imgs)
assert isinstance(signals, np.ndarray)
assert signals.shape[0] == 2

# SurfaceMasker with list of surface image time series
imgs = [_make_surface_img(5), _make_surface_img(6)]
masker = SurfaceMasker()
signals = masker.fit_transform(imgs)
assert isinstance(signals, np.ndarray)
assert signals.shape[0] == 11

# MultiNiftiMasker with list of nifti image time series
# here multimasker returns a list of arrays
imgs = [generate_fake_fmri(length=5)[0], generate_fake_fmri(length=6)[0]]
masker = MultiNiftiMasker()
signals = masker.fit_transform(imgs)
assert isinstance(signals, list)
assert isinstance(signals[0], np.ndarray)
assert signals[0].shape[0] == 5

# NiftiMasker with list of nifti image with a single sample
# here masker returns an array as usual
imgs = [generate_fake_fmri(length=1)[0], generate_fake_fmri(length=1)[0]]
[...]

PR #5669, opened 2025-09-10
[REF] create a multi masker mixin

Remi-Gau:
- relates to #5539

- extract common behavior across multimaskers into a single mixin


Remi-Gau:
trying to see if we can do some refactoring before we start adding multimaskers for surfaces

PR #5671, opened 2025-09-10
[FIX] add warning about non interactive matplotlib backend

### LINKED ISSUE #5667: No warning thrown when the matplotlib backend does not allow plotting
No warning thrown when the matplotlib backend does not allow plotting.

> this is fixed but now the problem is different: will open a separate issue 

 _Originally posted by @Remi-Gau in [#5261](https://github.com/nilearn/nilearn/issues/5261#issuecomment-3270369492)_

after installing `nilearn[plotting]`

when running the following (adapted from examples/00_tutorials/plot_python_101.py)

```python
try:
except ImportError:
    raise RuntimeError("This script needs the matplotlib library")

# %%
# A simple example of basic Python numerics and how to plot it.

# import numpy: the module providing numerical arrays
import numpy as np

t = np.linspace(1, 10, 2000)

plt.plot(t, np.cos(t))

plt.show()
```

I get NO figure and this warning

```
/home/remi-gau/github/nilearn/nilearn/examples/00_tutorials/plot_python_101.py:21: UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown
  plt.show()
```

but when I run another example:

```bash
python examples/00_tutorials/plot_nilearn_101.py
```

I get NO figure and no warning.

```
Path to MNI152 template: PosixPath('/home/remi-gau/github/nilearn/nilearn/nilearn/datasets/data/mni_icbm152_t1_tal_nlin_sym_09a_converted.nii.gz')
[fetch_haxby] Dataset found in /home/remi-gau/nilearn_data/haxby2001
(124, 256, 256, 2)
Output will be saved to: /home/remi-gau/github/nilearn/nilearn/results/plot_nilearn_101
```

This is because we skip any type of plotting when "agg" is the backend used:

https://github.com/nilearn/nilearn/blob/2d1404665be3d74c30a2e3ea35edc66e3b98731c/nilearn/plotting/image/img_plotting.py#L62-L71

The problem is that this way there is no way to know how to fix the issue (where are my figures????!!!)

I understand this nilearn.plotting.show was introduced excessive amount of warnings in the doc build (https://github.com/nilearn/nilearn/pull/1795) but this came at the cost of explicit skipping useful warnings.

Remi-Gau:
- Closes #5667

- add warning about non interactive matplotlib backend
- should throw a single warning per script that calls some plotting function

### DIFF
--- a/examples/00_tutorials/plot_python_101.py
+++ b/examples/00_tutorials/plot_python_101.py
@@ -8,6 +8,7 @@ try:
 except ImportError:
     raise RuntimeError("This script needs the matplotlib library")
 
+
 # %%
 # A simple example of basic Python numerics and how to plot it.
 
@@ -17,3 +18,5 @@ import numpy as np
 t = np.linspace(1, 10, 2000)
 
 plt.plot(t, np.cos(t))
+
+plt.show()
--- a/nilearn/_utils/helpers.py
+++ b/nilearn/_utils/helpers.py
@@ -71,6 +71,19 @@ def set_mpl_backend(message=None):
                 stacklevel=find_stack_level(),
             )
 
+        if new_backend == "agg":
+            warnings.warn(
+                (
+                    f"\nYou are using the '{matplotlib.get_backend()}' "
+                    "matplotlib backend that is non-interactive."
+                    "\nNo figure will be plotted when calling "
+                    "matplotlib.pyplot.show() or nilearn.plotting.show()."
+                    "\nYou can fix this by installing a different backend: "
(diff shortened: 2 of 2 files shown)


examples/00_tutorials/plot_python_101.py
```diff
@@ -8,6 +8,7 @@
 except ImportError:
     raise RuntimeError("This script needs the matplotlib library")
 
+
```
Remi-Gau:
```suggestion
```

hndgzkn (approved):
LGTM

PR #5645, opened 2025-09-03
[ENH] Refactor `nilearn.plotting`: extract common code 2

hndgzkn:
- Closes None

Changes proposed in this pull request:

- Create `plotting._engine_utils.threshold_cmap` function, and reuse it from 
  - `plotting.surface._matplotlib_backend._plot_surf`, 
  - `plotting.surface._matplotlib_backend._colorbar_from_array`
  - `plotting._engine_utils.colorscale`
- Remove `plotting.surface._matplotlib_backend._get_cmap` and use `threshold_cmap` instead

> This PR is continuation of #5608.

PR #5608, opened 2025-08-26
[ENH] Refactor `nilearn.plotting`: extract common code

hndgzkn:
- Closes None

Changes proposed in this pull request:

- Refactors `nilearn.plotting.surface._matplotlib_backend` to extract common code to 3 functions:
    - get_bg_data
    - _normalize_bg_data
    - _apply_darkness

- Add docstring for `get_colorbar_and_data_ranges`

> This PR is part of changes in https://github.com/nilearn/nilearn/pull/5537/files to simplify code review.

PR #5680, opened 2025-09-12
[ENH] Refactor `nilearn.plotting`: extract common code 4

hndgzkn:
- Closes None

- Use single function `nilearn.plotting._utils.get_cbar_ticks` to calculate cbar ticks moving contents of `get_cropped_cbar_ticks` and `_get_ticks` inside it.
- Modify `nilearn.plotting._utils.get_cbar_ticks` 
  - to use tick values formatted with `cbar_tick_format` to avoid the same value to appear multiple times in the colorbar
  - to sort tick values to avoid incorrect order in the list
  - to add `threshold` value to linearly spaced ticks list, if it is not close enough to a tick value; this increases the number of ticks
  - always include `vmin`, `vmax` values and `threshold` (if not outside vmin, vmax) values in the tick list
  - Remove tick values between 0 and +-threshold if not equal to vmin/vmax.
- Modify tests for `get_cbar_tick`
  - update tests with explicit test values
  - remove repetitive test cases 
- Define `get_cbar_bounds` to set the `boundaries` for matplotlib colorbar. This returns bounds with formatted `vmin` and `vmax` to make sure that vmin/vmax values appear on the colorbar.
- Update `cbar_tick_format` from `%i` to `%.2g` for `nilearn.plotting.image.img_plotting.plot_roi` to avoid multiple cbar ticks to be rounded to 0.
- Fix colorbar negative tick values appearing on the positive side with `-0` for reporting plots when threshold is negative.
- Define `DEFAULT_TICK_FORMAT="%.2g"`
- Fix the bug to view negative threshold to appear in colorbar for images thresholded with negative threshold:
  - Fix bug in in `nilearn.plotting.displays._slicers.py`
  - Remove the code that tried to fix it in `nilearn.reporting.glm_reporter`

> This PR is continuation of:
>    - #5608 
>    - #5645
>    - #5666


hndgzkn:
@Remi-Gau @bthirion I've added more tests for `nilearn.plotting._utils.get_cbar_ticks` and tests for `nilearn.plotting.image.utils.get_cropped_cbar_ticks`. I've put comments where return values reveal bugs.

If I am not mistaken, these two functions actually should return the same values. My plan is to replace them with a single function as below if that is also ok with you.

```Python
def get_cbar_ticks_new(vmin, vmax, threshold=None, n_ticks=5):
    # Shall we do this independent of the value of threshold???
    if vmin == vmax:
        return np.linspace(vmin, vmax, 1)

    ticks = np.linspace(vmin, vmax, n_ticks)

    if threshold is not None:
        diff = abs(abs(ticks) - threshold)
        # if the values are either positive or negative
        if 0 <= vmin <= vmax or vmin <= vmax <= 0:
            idx_closest = np.argmin(diff)
            ticks[idx_closest] = threshold if vmin >= 0 else -threshold
        else:
            # Edge case where the thresholds are exactly
            # at the same distance to 4 ticks
            if np.count_nonzero(min(diff)) == 4:
                idx_closest = np.sort(np.argpartition(diff, 4)[:4])
                idx_closest = np.isin(ticks, np.sort(ticks[idx_closest])[1:3])
            else:
                # Find the closest 2 ticks
                idx_closest = np.sort(np.argpartition(diff, 2)[:2])
                if 0 in ticks[idx_closest]:
                    idx_closest = np.sort(np.argpartition(diff, 3)[:3])
                    idx_closest = idx_closest[[0, 2]]
            ticks[idx_closest] = [-threshold, threshold]
    return ticks
```

bthirion:
> @Remi-Gau @bthirion I've added more tests for `nilearn.plotting._utils.get_cbar_ticks` and tests for `nilearn.plotting.image.utils.get_cropped_cbar_ticks`. I've put comments where return values reveal bugs.
> 
> If I am not mistaken, these two functions actually should return the same values. My plan is to replace them with a single function as below if that is also ok with you.
> 
> ```python
> def get_cbar_ticks_new(vmin, vmax, threshold=None, n_ticks=5):
>     # Shall we do this independent of the value of threshold???
>     if vmin == vmax:
>         return np.linspace(vmin, vmax, 1)
> 
>     ticks = np.linspace(vmin, vmax, n_ticks)
> 
>     if threshold is not None:
>         diff = abs(abs(ticks) - threshold)
>         # if the values are either positive or negative
>         if 0 <= vmin <= vmax or vmin <= vmax <= 0:
>             idx_closest = np.argmin(diff)
>             ticks[idx_closest] = threshold if vmin >= 0 else -threshold
>         else:
>             # Edge case where the thresholds are exactly
>             # at the same distance to 4 ticks
>             if np.count_nonzero(min(diff)) == 4:
>                 idx_closest = np.sort(np.argpartition(diff, 4)[:4])
>                 idx_closest = np.isin(ticks, np.sort(ticks[idx_closest])[1:3])
>             else:
>                 # Find the closest 2 ticks
>                 idx_closest = np.sort(np.argpartition(diff, 2)[:2])
>                 if 0 in ticks[idx_closest]:
>                     idx_closest = np.sort(np.argpartition(diff, 3)[:3])
>                     idx_closest = idx_closest[[0, 2]]
>             ticks[idx_closest] = [-threshold, threshold]
>     return ticks
> ```

To make sure I understand: The original code seems to worry about whether the cbar is symmetric or not. In the above snippet you don't. Is there an obvious reason for that ?

hndgzkn:
> To make sure I understand: The original code seems to worry about whether the cbar is symmetric or not. In the above snippet you don't. Is there an obvious reason for that ?

Actually neither `get_cbar_ticks` nor `get_cropped_cbar_ticks` has `symmetric_cbar` parameter.  Only in `get_cropped_cbar_ticks` it checks if `-threshold` and `threshold` in between  `cbar_vmin` and `cbar_vmax`. As far as I can see, it does not make a difference. And it returns incorrect tick list in many cases.

In the code above, it returns threshold values within the bounds if +-threshold is within the bounds, if one of the threshold is out of bounds, the corresponding bound is replaced with the threshold.

hndgzkn:
Below are the same tests, and the return values with the proposed function:

```
    "vmin,vmax,threshold,num_ticks,expected",
    [
        (0, 0, None, 5, [0]),
        (3, 3, None, 5, [3]),
        (-3, -3, None, 5, [-3]),
        (-5, 5, None, 5, [-5, -2.5, 0, 2.5, 5]),
        (-3, 5, None, 5, [-3, -1, 1, 3, 5]),
        (-5, 3, None, 5, [-5, -3, -1, 1, 3]),
        (0, 3, None, 5, [0, 0.75, 1.5, 2.25, 3]),
        (-3, 0, None, 5, [-3, -2.25, -1.5, -0.75, 0]),
        (0, 0, 0.5, 5, [0]),
        (3, 3, 0.5, 5, [3]),
        (-3, -3, 0.5, 5, [-3]),
        (-3, 5, 0.5, 5, [-3, -0.5, 0.5, 3, 5]),  # fixes the bug
        (-3, 5, 0.5, 7, [-3, -1.66, -0.5, 0.5, 2.33, 3.66, 5]),
        (-3, 5, 1.5, 7, [-3, -1.5, -0.33, 1.5, 2.33, 3.66, 5]),
        (-3, 5, 0.5, 6, [-3, -0.5, 0.5, 1.8, 3.4, 5]),
        (-5, 3, 0.5, 5, [-5, -3, -0.5, 0.5, 3]),
        (-5, 3, 0.5, 7, [-5, -3.66, -2.33, -0.5, 0.5, 1.66, 3]),
        (-5, 3, 0.5, 6, [-5, -3.4, -1.8, -0.5, 0.5, 3]),
        (0, 3, 0.5, 5, [0, 0.5, 1.5, 2.25, 3]),  # fixes the bug
        (-3, 0, 0.5, 5, [-3, -2.25, -1.5, -0.5, 0]),  # fixes the bug
        (-3, 5, 0.7, 5, [-3, -0.7, 0.7, 3, 5]),
        (-3, 4, 5, 5, [-5, -1.25, 0.5, 2.25, 5]),
        (-5, 5, 0.5, 5, [-5, -0.5, 0, 0.5, 5]),
        (-5, 5, 0.5, 7, [-5, -3.33, -0.5, 0, 0.5, 3.33, 5]),
        (-5, 5, 0.5, 6, [-5, -3, -0.5, 0.5, 3, 5]),
        (-5, 5, 2.5, 6, [-5, -2.5, -1, 1, 2.5, 5]),
    ],

```

If it corresponds to the desired behavior, we can replace.

bthirion:
> (-3, 5, 0.5, 5, [-3, -0.5, 0.5, 3, 5]),  # fixes the bug

I'm surprised by the values here and in the following lines as the scale is neither linear nor piecewise linear. What is the logic here ?

bthirion:
To make it clear: the overall behavior is reasonable, but I think it would make sense to have a "formula" describing the tick locations.

hndgzkn:
> > (-3, 5, 0.5, 5, [-3, -0.5, 0.5, 3, 5]),  # fixes the bug
> 
> I'm surprised by the values here and in the following lines as the scale is neither linear nor piecewise linear. What is the logic here ?

What it does is, first it creates 5 equally spaced tick points between -3 and 5 which gives `[-3., -1.,  1.,  3.,  5.]`. Then it replaces the closest points to +-threshold with threshold values which gives `[-3., -0.5,  0.5,  3.,  5.]`. 

I did not change the algorithm in that sense. I just fixed the parts where threshold values did not appear, or appeared in wrong order.

bthirion:
> > > (-3, 5, 0.5, 5, [-3, -0.5, 0.5, 3, 5]),  # fixes the bug
> > 
> > 
> > I'm surprised by the values here and in the following lines as the scale is neither linear nor piecewise linear. What is the logic here ?
> 
> What it does is, first it creates 5 equally spaced tick points between -3 and 5 which gives `[-3., -1., 1., 3., 5.]`. Then it replaces the closest points to +-threshold with threshold values which gives `[-3., -0.5, 0.5, 3., 5.]`.
> 
> I did not change the algorithm in that sense. I just fixed the parts where threshold values did not appear, or appeared in wrong order.

OK, that makes sense, thx. I would just write this somewhere to let future contributors understand the logic.

bthirion:
LGTM overall !

Remi-Gau:
@hndgzkn 
do you want to merge as is and fix the bugs in a follow up?

hndgzkn:
> @hndgzkn do you want to merge as is and fix the bugs in a follow up?

I am thinking about the behavior. How important is the number of ticks or to see the threshold in the colorbar? 

I mean, for the case @bthirion mentioned for example, how would it be to return `[-3., -1, -0.5, 0.5, 1, 3., 5.]` instead of `[-3., -0.5, 0.5, 3., 5.]`  increasing the number of ticks from 5 to 7? We could adapt this behavior, when the difference of threshold and closest value is greater than a quarter of the distance between two ticks for example. Or instead of adding the threshold values, we should omit to have them in tick list. 

In the original code, there is `threshold / vmax_temp > 0.12` for such cases. Actually for some test cases that I've written, threshold does not appear because of this check. So they are not bugs as I mentioned.

What is the preferred behavior?

hndgzkn:
The below example is a better one.

For an interval `[-10, -5, 0, 5, 10]` with `threshold=0.5`, which is one of the below is preferable : 

- `[-10, -5, 0, 5, 10]`
- `[-10, -5, -0.5, 0, 0.5, 5, 10]`
- `[-10, -0.5, 0, 0.5, 10]`

Remi-Gau:
> The below example is a better one.
> 
> For an interval `[-10, -5, 0, 5, 10]` with `threshold=0.5`, which is one of the below is preferable :
> 
>     * `[-10, -5, 0, 5, 10]`
> 
>     * `[-10, -5, -0.5, 0, 0.5, 5, 10]`
> 
>     * `[-10, -0.5, 0, 0.5, 10]`

my gut feeling goes for the second one

hndgzkn:
> my gut feeling goes for the second one

Mine too

Remi-Gau:
> > my gut feeling goes for the second one
> 
> Mine too

great gut feel alike!

bthirion:
> > > my gut feeling goes for the second one
> > 
> > 
> > Mine too
> 
> great gut feel alike!

But if you're asking for 5 ticks, this is not OK...

Remi-Gau:
> But if you're asking for 5 ticks, this is not OK...

yeah but there is nowhere in the public API where the user can ask for the number of ticks (AFAIK), so I think we can adapt this a bit: we can add a comment explaining that n_ticks is more like a "general guideline" rather than a "strict rule". 😉

nilearn/plotting/_utils.py
```diff
@@ -17,10 +17,24 @@ def engine_warning(engine):
     warn(message, stacklevel=find_stack_level())
 
 
+def _add_to_ticks(ticks, threshold):
+    ticks = ticks[ticks != 0]
```
bthirion: A small docstring ?

bthirion (commented):
LGTM overall

bthirion:
> > But if you're asking for 5 ticks, this is not OK...
> 
> yeah but there is nowhere in the public API where the user can ask for the number of ticks (AFAIK), so I think we can adapt this a bit: we can add a comment explaining that n_ticks is more like a "general guideline" rather than a "strict rule". 😉

OK

nilearn/plotting/image/img_plotting.py
```diff
@@ -280,12 +279,6 @@ def _plot_img_with_bg(
         display.draw_cross()
     if title is not None and title != "":
         display.title(title)
-    if hasattr(display, "_cbar"):
-        cbar = display._cbar
-        new_tick_locs = get_cropped_cbar_ticks(
-            cbar.vmin, cbar.vmax, threshold, n_ticks=len(cbar.locator.locs)
-        )
-        cbar.set_ticks(new_tick_locs)
```
hndgzkn, line 288: I think this was repetitive. I removed it. Cbar ticks are already set above.

nilearn/plotting/image/img_plotting.py
```diff
@@ -751,7 +744,7 @@ def plot_roi(
     cmap="gist_ncar",
     dim="auto",
     colorbar=True,
-    cbar_tick_format="%i",
+    cbar_tick_format="%.2g",
```
hndgzkn: Integer format caused multiple values on the plot to be rounded to 0. 

We might define `DEFAULT_FORMAT = "%.2g"`.

nilearn/plotting/surface/tests/test_matplotlib_backend.py
```diff
@@ -109,35 +108,6 @@ def test_get_bounds(data, expected):
...
-        (0, 3, "%i", [0, 1, 2, 3]),
-        (0, 4, "%i", [0, 1, 2, 3, 4]),
-        (1, 5, "%i", [1, 2, 3, 4, 5]),
-        (0, 5, "%i", [0, 1.25, 2.5, 3.75, 5]),
-        (0, 10, "%i", [0, 2.5, 5, 7.5, 10]),
-        (0, 0, "%.1f", [0]),
-        (0, 1, "%.1f", [0, 0.25, 0.5, 0.75, 1]),
-        (1, 2, "%.1f", [1, 1.25, 1.5, 1.75, 2]),
-        (1.1, 1.2, "%.1f", [1.1, 1.125, 1.15, 1.175, 1.2]),
-        (0, np.nextafter(0, 1), "%.1f", [0.0e000, 5.0e-324]),
```
hndgzkn, line 125: These test cases are moved to `nilearn.plotting.tests.test_utils`

nilearn/plotting/_utils.py
```diff
@@ -17,60 +17,157 @@ def engine_warning(engine):
...
+
+    # check if threshold should be added to the tick list or replaced by a
+    # value in the list
+    return bool(
+        min_diff > abs(ticks[1] - ticks[0]) / 3
+        or min_diff >= threshold * 4 / 3
+    )
+
+
+def get_cbar_ticks(vmin, vmax, threshold=None, n_ticks=5, tick_format="%.2g"):
```
hndgzkn: I've introduced `tick_format` in the signature to make sure that the ticks returned from this function will all be different when plotted by matplotlib. For example When there are 9.96 and 10 in the tick list, they appear both as 10 in the colorbar.

nilearn/plotting/_utils.py
```diff
@@ -17,60 +17,157 @@ def engine_warning(engine):
...
+
+            ticks = np.append(ticks, [vmin, vmax])
+        # remove unnecessary ticks that would be between 0 and +-threshold
+        ticks = ticks[
+            np.where(
+                (ticks > threshold)
+                | (ticks < -threshold)
+                | (np.isin(ticks, [0, vmin, vmax, threshold, -threshold]))
+            )
+        ]
```
hndgzkn, line 179: This is to remove thresholded tick values from tick list unless it is vmin, vmax and 0.

nilearn/plotting/image/utils.py
```diff
@@ -89,34 +88,6 @@ def __repr__(self):
 MNI152TEMPLATE = _MNI152Template()
 
 
-def get_cropped_cbar_ticks(cbar_vmin, cbar_vmax, threshold=None, n_ticks=5):
-    """Return ticks for cropped colorbars."""
-    new_tick_locs = np.linspace(cbar_vmin, cbar_vmax, n_ticks)
```
hndgzkn, line 94: Moved this inside `get_cbar_ticks`.

nilearn/plotting/surface/_matplotlib_backend.py
```diff
@@ -314,21 +313,6 @@ def _get_bounds(data, vmin=None, vmax=None):
     return vmin, vmax
 
 
-def _get_ticks(vmin, vmax, cbar_tick_format, threshold):
-    """Help for plot_surf with matplotlib engine.
-
```
hndgzkn, line 319: Moved this inside `get_cbar_ticks`.

nilearn/plotting/surface/_matplotlib_backend.py
```diff
@@ -517,20 +501,17 @@ def _plot_surf(
...
+                tick_format=cbar_tick_format,
             )
-            if threshold is not None and (
-                cbar_tick_format == "%i" and int(threshold) != threshold
-            ):
-                warn(
-                    "You provided a non integer threshold "
-                    "but configured the colorbar to use integer formatting.",
-                    stacklevel=find_stack_level(),
-                )
```
hndgzkn, line 530: Moved this inside `get_cbar_ticks`.

hndgzkn:
I have some other improvements but the PR will become big to review. I'll continue in a follow up PR

bthirion (approved):
LGTM thx !

bthirion:
Is the buid failure somehow related ?

hndgzkn:
> Is the buid failure somehow related ?

I am not sure.

```
 nilearn/decomposition/tests/test_decomposition_estimators.py::test_check_estimator_nilearn[estimator30-check_img_estimator_standardization-check_img_estimator_standardization] - ValueError: array must not contain infs or NaNs
```
That should not be relevant. I'll check.

nilearn/reporting/glm_reporter.py
```diff
@@ -930,6 +932,15 @@ def _stat_map_to_png(
...
         check_parameter_in_allowed(plot_type, ["slice", "glass"], "plot_type")
+        if not two_sided and threshold < 0:
+            # we cannot use negative threshold in plot_stat_map
+            # so we flip the sign of the image, the colormap
+            # and we relabel the colorbar later
+            data = safe_get_data(stat_img)
+            affine = stat_img.affine
+            stat_img = new_img_like(stat_img, as_ndarray(-data), affine)
+            cmap = "Blues"
+
```
hndgzkn: This is to show one sided negative thresholded values in colorbar for images (for surface images it was already done).

nilearn/plotting/displays/_slicers.py
```diff
@@ -713,9 +715,6 @@ def _show_colorbar(
             Maximal value for the colorbar. If None, the maximal value
             is computed based on the data.
         """
-        offset = 0 if threshold is None else threshold
-        offset = min(offset, norm.vmax)
-
```
hndgzkn, line 718: This part was preventing thresholding the colorbar correctly for volume data. 

For example for data between [-7.9, 0] and threshold value 3, it takes offset as -7.9 and thresholds values below that in the colorbar which is neither visible nor correct.

nilearn/reporting/glm_reporter.py
```diff
@@ -907,14 +907,6 @@ def _stat_map_to_png(
...
 
     if isinstance(stat_img, SurfaceImage):
-        if not two_sided and threshold < 0:
-            # we cannot use negative threshold in plot_surf_stat_map
-            # so we flip the sign of the image, the colormap
-            # and we relabel the colorbar later
-            for k, v in stat_img.data.parts.items():
-                stat_img.data.parts[k] = -v
-            cmap = "Blues"
-
```
hndgzkn, line 918: This part was necessary because of the bug that I mentioned above. So it is no more necessary.

nilearn/reporting/glm_reporter.py
```diff
@@ -963,17 +956,6 @@ def _stat_map_to_png(
...
-            isinstance(stat_img, SurfaceImage)
-            and not two_sided
-            and threshold < 0
-        ):
-            # Because the image has been flipped
-            # replace labels with their negative
-            ticks = stat_map_plot._cbar.get_ticks()
-            stat_map_plot._cbar.set_ticks(ticks)
-            stat_map_plot._cbar.set_ticklabels([f"{-t:.2g}" for t in ticks])
-
```
hndgzkn, line 977: The same here.

nilearn/reporting/glm_reporter.py
```diff
@@ -907,14 +907,6 @@ def _stat_map_to_png(
             cmap = "Blues_r"
```
hndgzkn, line 908: This code is enough for plotting to view data thresholded with negative threshold value and `two_sided=False`

nilearn/plotting/displays/_slicers.py
bthirion: What is the new behavior, then ?
Do we have a test to enforce it ?

hndgzkn: > What is the new behavior, then ? 

The threshold will stay as 3, and for the above case the interval [-3, 0] will appear as thresholded on the colorbar.

I think this was implemented thinking that data will always have positive values. When there are positive values this works. The problem is when all data <= 0.

> Do we have a test to enforce it ?

I've added a test for this type of case. And also the following base image reveals the correction.

https://github.com/nilearn/nilearn/pull/5680/files#diff-29f084ccdb85a093ef5f980f5b780b69b80c740b4bb74af6110442759e1b3413

hndgzkn: There are such checks done in multiple places. There are differing defaults, not clear for which reason. 
I try to unify all, unify the behavior. 

I have some other improvements, but I'll add them in follow-up PRs.

hndgzkn: https://github.com/nilearn/nilearn/pull/5680/files#diff-32bc7fd16401910787a21d446acec7e4779eaf3eb2833984744025f18f9923a2 is another example which is fixed by this change and also removing the code in `nilearn.reporting.glm_reporter`.

nilearn/plotting/_utils.py
hndgzkn: An example is [here](https://github.com/nilearn/nilearn/pull/5680/files#diff-7ba6fa9da9565443e9ad37b532726a0d00f878657ba6b84e8d3ebc666c528fe4).

nilearn/plotting/displays/_slicers.py
hndgzkn: > https://github.com/nilearn/nilearn/pull/5680/files#diff-32bc7fd16401910787a21d446acec7e4779eaf3eb2833984744025f18f9923a2 is another example which is fixed by this change and also removing the code in `nilearn.reporting.glm_reporter`.

Correction: This is surface image, so this is fixed by removing code in `glm_reporter`.

nilearn/plotting/tests/baseline/test_plot_functions_vmin_-1-plot_roi.png
Remi-Gau, line 1: side comment: given that plot_roi does not allow image with negative values, it feels strange to allow negative vmin, no?

nilearn/plotting/tests/baseline/test_plot_functions_annotate_plot_roi.png
Remi-Gau, line 1: clear improvement of the color bar on this one !!!!

nilearn/plotting/tests/baseline/test_plot_functions_vmax_2-plot_anat.png
Remi-Gau, line 1: not sure it can or should be fixed in this PR, but shouldn't the colornbar include the 0 value here: we go from -1.7 to 2 and no idea when we are going from negative to positive.

will flag the other figure with "zero missing"

nilearn/plotting/tests/baseline/test_plot_functions_vmax_2-plot_epi.png
Remi-Gau, line 1: "zero missing"

nilearn/plotting/tests/baseline/test_plot_functions_vmax_2-plot_img.png
Remi-Gau, line 1: "zero missing"

nilearn/plotting/tests/baseline/test_plot_functions_vmax_3-plot_anat.png
Remi-Gau, line 1: "zero missing"

nilearn/plotting/tests/baseline/test_plot_functions_vmax_3-plot_epi.png
Remi-Gau, line 1: "zero missing"

nilearn/plotting/tests/baseline/test_plot_functions_vmax_3-plot_img.png
Remi-Gau, line 1: "zero missing"

nilearn/plotting/tests/baseline/test_plot_functions_vmax_3-plot_roi.png
Remi-Gau, line 1: the lack of zero on such color bar is less shocking as we have a symmetric -0.5 to 0.5 interval

nilearn/plotting/tests/baseline/test_plot_functions_vmin_-1-plot_anat.png
Remi-Gau, line 1: "zero missing"

nilearn/plotting/tests/baseline/test_plot_functions_vmin_-1-plot_epi.png
Remi-Gau, line 1: "zero missing"

nilearn/plotting/tests/baseline/test_plotting_functions_radiological_view_plot_anat.png
Remi-Gau, line 1: "zero missing"

nilearn/plotting/tests/baseline/test_plotting_functions_radiological_view_plot_roi.png
Remi-Gau, line 1: here we have a 4 on the positive side on the color bar but no -4 on the negative side: could this be fixed?

nilearn/reporting/tests/baseline/test_stat_map_to_png_surface_None-False--0.5.png
Remi-Gau, line 1: clear improvement here too

nilearn/reporting/tests/baseline/test_stat_map_to_png_volume_200-None-False--3-slice.png
Remi-Gau, line 1: defo better here too

nilearn/plotting/tests/baseline/test_plot_functions_vmax_2-plot_anat.png
Remi-Gau: @hndgzkn 
you might know better if this kind of tweak is out of scope for this PR

hndgzkn: I think this PR is the best place for fixing this :)

Normally 0 appears only if it comes from `np.linspace`, or it is in `[threshold, vmin, vmax]`.

If we decide 0 to appear for colorbars with positive and negative values, I can add it. I think it is a good idea.

nilearn/plotting/tests/baseline/test_plot_functions_vmin_-1-plot_roi.png
hndgzkn: Yes. Would you like me to do it in this PR, or a later one?

nilearn/plotting/tests/baseline/test_plot_functions_vmax_2-plot_anat.png
Remi-Gau: > If we decide 0 to appear for colorbars with positive and negative values

am not sure we systematically want it: in some cases this may lead to have 3 values in the color bar almost stacked on top of each other (when the threshold is really close to 0 compare the overall range of the colorbar)

but maybe it will hard to avoid such edge cases and should still prefer to always have a 0

nilearn/plotting/tests/baseline/test_plotting_functions_radiological_view_plot_roi.png
hndgzkn: Yes, I haven't paid attention. I'll check that.

hndgzkn:
@Remi-Gau [nilearn/plotting/image/tests/test_plot_stat_map.py::test_plot_stat_map_colorbar_variations](https://github.com/nilearn/nilearn/actions/runs/18176588870/job/51775822940?pr=5680#step:7:12683) started to fail today (I think, or yesterday) with 

```
 AssertionError: passed string "html5" as "writer" parameter; pass an instance, or use the 
"writer_name" parameter instead (in docutils.core.publish_* convenience functions).
```
Could you understand the actual reason? It is due to latest version of some dependency as far as I understand.

Remi-Gau:
> @Remi-Gau [nilearn/plotting/image/tests/test_plot_stat_map.py::test_plot_stat_map_colorbar_variations](https://github.com/nilearn/nilearn/actions/runs/18176588870/job/51775822940?pr=5680#step:7:12683) started to fail today (I think, or yesterday) with
> 
> ```
>  AssertionError: passed string "html5" as "writer" parameter; pass an instance, or use the 
> "writer_name" parameter instead (in docutils.core.publish_* convenience functions).
> ```
> 
> Could you understand the actual reason? It is due to latest version of some dependency as far as I understand.

have seen that in other places in our CI including on main, I suspect a testing dependency change probably linked to test report generation

hndgzkn:
> have seen that in other places in our CI including on main, I suspect a testing dependency change probably linked to test report generation

As far as I understand docutils has a recent [release](https://docutils.sourceforge.io/RELEASE-NOTES.html) that causes the problem. 

`pytest_reporter` makes this call `env.filters["rst"] = lambda s: publish_parts(source=s, writer="html5")["body"]` which does not comply with the latest version. A solution would be to limit `docutils<0.20`.

nilearn/plotting/tests/baseline/test_plot_functions_vmin_-1-plot_roi.png
Remi-Gau: no this is out of scope for this one

nilearn/plotting/tests/baseline/test_plotting_functions_radiological_view_plot_roi.png
hndgzkn: It is because the value rounded to 0 actually is not zero, so the threshold replaces that number, and -4. 

I propose to work on that in another PR.

Remi-Gau (approved):

Issue #5506, opened 2025-07-02
`plot_align_svg.py` script and similar can be moved to `maint_tools` folder?


hndgzkn:
random thought: should we move this script into our maint_tools folder?

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5481#discussion_r2179391883_

PR #5686, opened 2025-09-16
[MAINT] move maintenance script out of codebase

Remi-Gau:
- Closes https://github.com/nilearn/nilearn/issues/5506

- move maintenance script out of codebase


maint_tools/convert_templates.py
```diff
@@ -1,18 +1,25 @@
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#   "nilearn[plotting]>=0.12",
+#    "PyQt6"
+# ]
+# ///
```
Remi-Gau: FYI

this is a way to define the requirements for a script: https://peps.python.org/pep-0723/

maint_tools/plot_align_svg.py
```diff
@@ -1,6 +1,15 @@
 #!/usr/bin/env python
+
```
hndgzkn: This file might be renamed to something like `plot_align_glassbrain_svg.py`?

maint_tools/generate_json.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/sh
 
 # Quick and dirty script to regenerate the brain schematics JSONs from
-# the SVGs. This script has to be launched from the folder it lives
-# in, i.e. as ./generate_json.sh
+# the SVGs.
```
hndgzkn: This file might be renamed to something like `generate_json_from_svg.py`?

maint_tools/convert_templates.py
```diff
@@ -1,18 +1,25 @@
+# /// script
```
hndgzkn: This file could be renamed to something like `convert_brain_templates.py`?

hndgzkn (approved):
LGTM with minor suggestions.

maint_tools/generate_json.sh
Remi-Gau: oh actually I meant to remove this one: do you think we should keep it?
it's not like we are using updating the svg to glass brain that often so it feels like an overkill to have bash script for this

hndgzkn: I can't comment on that :) I do not really know where and when we use maintenance tools.

Remi-Gau: Will remove it

PR #5689, opened 2025-09-17
[FIX] revert one path change

Remi-Gau:
- Closes none but should fix doc build in main

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -55,7 +55,7 @@ clean:
 	-rm -rf modules/description/
 
 glm_reports:
-	python visual_testing/reporter_visual_inspection_suite.py full
+	python visual_testing/reporter_visual_inspection_suite.py partial
 
 sym_links_datasets:
 	# copy datasets RST from nilearn package to doc folder
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -55,7 +55,7 @@ clean:
 	-rm -rf modules/description/
 
 glm_reports:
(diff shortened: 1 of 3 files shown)


doc/Makefile
```diff
@@ -55,7 +55,7 @@ clean:
 	-rm -rf modules/description/
 
 glm_reports:
-	python visual_testing/reporter_visual_inspection_suite.py partial
+	python visual_testing/reporter_visual_inspection_suite.py full
```
Remi-Gau:
```suggestion
	python visual_testing/reporter_visual_inspection_suite.py partial
```


revert before merging

Remi-Gau:
merging as this should help us get the doc build back online on main

hndgzkn:
Ok. Do you know why it stopped working? I was trying to understand the reason of the failure.

nilearn/datasets/func.py
```diff
@@ -2490,7 +2490,7 @@ def patch_openneuro_dataset(file_list):
             if old_pattern in name:
                 new_name = name.replace(old_pattern, new_pattern)
                 if not Path(new_name).exists():
-                    Path(name).symlink_to(new_name)
```
hndgzkn, line 2493: Could this also solve the problem?

`Path(new_name).symlink_to(name)`

hndgzkn: https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

PR #5697, opened 2025-09-22
[MAINT] Use different dataset for benchmarks

### LINKED ISSUE #5657: Benchmarks workflow fails on schedule, only works when triggered manually
All previous scheduled runs of the benchmark workflow have been failing, eg: https://github.com/nilearn/nilearn/actions/runs/17373494278/job/49314526218#step:9:1898

All previous logs can be found here: https://github.com/nilearn/benchmarks/tree/main/logs

I think it only works when the workflow has been triggered manually. The reason is probably that the dataset being used is fetched from a host that restricts bot activity. We have noticed this sort of behaviour with the doc build sometimes.

So I am going to try a different dataset for this.

man-shu:
- Closes #5657 

- use `nilearn.datasets.fetch_abide_pcp` instead of `fetch_adhd`
- use python 3.10 instead of 3.9
- change cron schedule time to test in a few minutes

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -40,8 +40,8 @@ on:
         branches:
         -   main
     schedule:
-    # every Monday at 5pm UTC
-    -   cron: 0 17 * * MON
+    # every Tuesday at 8am UTC
+    -   cron: 0 8 * * TUE
 
     workflow_dispatch:  # This enables manual triggering from GitHub UI
 


man-shu:
@Remi-Gau just need a quick review if possible to test this out

emdupre:
In general this seems fine ; I'd want to wait until the checks pass to merge, though. It looks like it's running now, so should be done before the 3p UTC deadline !

emdupre:
It looks like it won't meet the 3p UTC deadline ; do you want to target slightly later, @man-shu ?

man-shu:
> do you want to target slightly later, @man-shu ?

Ok yes let's do 5pm.

.github/workflows/benchmark.yml
```diff
@@ -40,8 +40,8 @@ on:
         branches:
         -   main
     schedule:
-    # every Monday at 9am UTC
-    -   cron: 0 9 * * MON
+    # every Monday at 3pm UTC
```
emdupre:
```suggestion
    # every Monday at 5pm UTC
```

emdupre (approved):
Spurious failure on downloading from OSF ; I've retriggered, but given the balance of checks right now, going to go ahead and approve

man-shu:
Thanks! Just changed the time to 8am UTC and merged. Let's see if that works

man-shu:
The dataset had nifti images under a different parameter. Will revert and merge again...

PR #5700, opened 2025-09-22
[FIX] skip stack frame in warning

### LINKED ISSUE #5699: nilearn warnings reported as being sent by sklearn `_set_output`
the nilearn function `find_stack_level` used for warning should help users where in their code a nilearn function is throwing a warning so that hopefully they can do something about it

the way it works is that it will 'report' in the stack, the first non nilearn function it encounters because it assumes it is there that the user used a nilearn function

but sometimes the nilearn code does things like

`some_nilearn_methods` --calls--> `some_sklearn_methods` --calls--> `some_nilearn_methods` --> throw warning

for example the maskers calling sklearn _set_output.

this happens a lot in our tests:

```
pytest nilearn/maskers/tests/test_nifti_labels_masker.py::test_nifti_labels_masker_resampling
```

will show one of our warning shown at the stack level of sklearn/utils/_set_output.py

```
nilearn/maskers/tests/test_nifti_labels_masker.py::test_nifti_labels_masker_resampling[data]
  /home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.11/site-packages/sklearn/utils/_set_output.py:316: UserWarning: After resampling the label image to the data image, the following labels were removed: {np.int32(9), np.int32(2), np.int32(3), np.int32(4)}. Label image only contains 6 labels (including background).
    data_to_wrap = f(self, X, *args, **kwargs)
```

This is not very helpful for users to know where in their code, they can do something to deal with this warning

it may be good to change find_stack_level to skip some specific functions like this

Remi-Gau:
- Closes #5699 

- skip sklearn _set_output when reporting stacklevel for nilearn warnings

### DIFF
--- a/nilearn/_utils/logger.py
+++ b/nilearn/_utils/logger.py
@@ -152,7 +152,10 @@ def find_stack_level() -> int:
     pkg_dir = Path(nil.__file__).parent
 
     # list of stack frames to skip
-    skip_list = [Path("sklearn") / "utils" / "_set_output.py"]
+    skip_list = [
+        Path("sklearn") / "utils" / "_set_output.py",
+        Path("joblib") / "memory.py",
+    ]
 
     # https://stackoverflow.com/questions/17407119/python-inspect-stack-is-slow
     frame = inspect.currentframe()


Remi-Gau:
> This does the trick ?

yup!

PR #5701, opened 2025-09-22
[REF] refactor in decoding

Remi-Gau:
- Closes none but addresses some of the issues mentioned in https://github.com/nilearn/nilearn/pull/5610

- rename `_binarize_y` to `_set_classes`
- remove `fit` method from `_RegressorMixin`


Remi-Gau:
@hndgzkn 
there are other issues you pointed out but I want to make very small PRs that are easy to review

nilearn/decoding/_mixin.py
```diff
@@ -39,16 +35,16 @@ def _n_problems(self):
...
         """
         y = np.array(y)
 
-        self._enc = LabelBinarizer(pos_label=1, neg_label=-1)
-        y = self._enc.fit_transform(y)
-        self.classes_ = self._enc.classes_
+        enc = LabelBinarizer(pos_label=1, neg_label=-1)
+        y = enc.fit_transform(y)
+        self.classes_ = enc.classes_
         self.n_classes_ = len(self.classes_)
```
hndgzkn: Is `n_classes` a necessary attribute for classifiers? Instead of introducing that could we use `len(self._get_classes)` ? That would avoid introducing a new attribute and it would work both for regressors and classifiers. 

It would also help avoid defining two `n_problems` in two mixins. Below function should work for both.

```python
def _n_problems(self):
    if len(self._get_classes()) > 2:
        return len(self._get_classes())
    else:
        return 1
```

nilearn/decoding/decoder.py
```diff
@@ -672,7 +678,7 @@ def fit(self, X, y, groups=None):
 
         # Define the number problems to solve. In case of classification this
         # number corresponds to the number of binary problems to solve
-        y = self._binarize_y(y)
+        y = self._set_classes(y)
```
hndgzkn, line 693: This comment is not relevant here, but it's relevant to `decoder.py`.

At line `784`, the code:

```python
 if is_classifier(self) and (self.n_classes_ == 2):
```

I think `is_classifier` check is not necessary as `self.n_classes_` or `len(self._get_classes)` will never be 2 for a regressor.

The same is valid for line `799`.

hndgzkn (approved):
LGTM with suggestions.

nilearn/decoding/_mixin.py
Remi-Gau: as far as I can tell n_classes_ is pretty classic in the sklearn codebase, but I can at least make it a property rather a fixed attribute

nilearn/decoding/decoder.py
hndgzkn: However this requires to change the check from `self.n_classes_ == 2` to `len(self._get_classes)` as regressor won't have `n_classes` property.

nilearn/decoding/decoder.py
```diff
@@ -619,6 +619,11 @@ def _set_classes(self, y):
         raise NotImplementedError()
 
     def _n_problems(self) -> int:
+        """Define the number problems to solve.
+
+        In case of classification this
+        number corresponds to the number of binary problems to solve.
+        """
         if len(self._get_classes()) > 2:
```
hndgzkn, line 625: Should this be `>= 2` ? If not, in the below replacement, `self._n_problems` will never return 2.

nilearn/decoding/space_net.py
```diff
@@ -766,6 +766,11 @@ def _return_loss_value(self):
         return loss
 
     def _n_problems(self):
+        """Define the number problems to solve.
+
+        In case of classification this
+        number corresponds to the number of binary problems to solve.
+        """
         if len(self._get_classes()) > 2:
```
hndgzkn, line 770: The same here.

nilearn/decoding/decoder.py
Remi-Gau: indeed

Remi-Gau:
closing for now
will probably revisit later

PR #5702, opened 2025-09-22
[REF] extract label properties and functions to mixin

Remi-Gau:
- Closes none

- move methods and functions only used by label maskers into the LabelsMaskerMixin

### DIFF
--- a/nilearn/maskers/_mixin.py
+++ b/nilearn/maskers/_mixin.py
@@ -135,8 +135,6 @@ class _LabelMaskerMixin:
         The region id corresponding to ``region_signal[:,i]``
         is ``region_ids_[i]``.
         ``region_ids_['background']`` is the background label.
-
-        .. versionadded:: 0.10.3
         """
         check_is_fitted(self)
 
--- a/nilearn/maskers/_mixin.py
+++ b/nilearn/maskers/_mixin.py
@@ -166,16 +166,6 @@ class _LabelMaskerMixin:
         """Generate a look up table if one was not provided.
 
         Also sanitize its content if necessary.
-
-        Parameters
-        ----------
-        labels_img : Nifti1Image | SurfaceImage
-
-        background_label : int | float
-
-        lut : Optional[str, Path, pd.DataFrame]
-
-        labels : Optional[list[str]]
         """
         labels_img = self.labels_img_
         background_label = self.background_label
--- a/nilearn/maskers/_mixin.py
+++ b/nilearn/maskers/_mixin.py
@@ -1,9 +1,19 @@
 """Mixin classes for maskers."""
 
-import numpy as np
+from copy import deepcopy
+from pathlib import Path
 
+import numpy as np
+import pandas as pd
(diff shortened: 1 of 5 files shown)


nilearn/maskers/_mixin.py
```diff
@@ -67,6 +77,80 @@ def set_output(self, *, transform=None):
 
 
 class _LabelMaskerMixin:
+    lut_: pd.DataFrame
+    _lut_: pd.DataFrame
```
hndgzkn, line 81: Do we really have to have these two?

If that is the case, I would suggest to create a private function to get the `lut` where necessary. Something` like` :

```python

def _get_lut(self):
    if hasattr(self, "_lut_"):
        return self._lut_
    else:
        return self.lut_
```

nilearn/maskers/_mixin.py
```diff
@@ -77,3 +161,89 @@ def get_feature_names_out(self, input_features=None):
...
+
+        Parameters
+        ----------
+        labels_img : Nifti1Image | SurfaceImage
+
+        background_label : int | float
+
+        lut : Optional[str, Path, pd.DataFrame]
+
+        labels : Optional[list[str]]
```
hndgzkn: I think you have forgotten the parameters section in docstring. They are all taken from `self` now.

hndgzkn (approved):
LGTM with some suggestions.

nilearn/maskers/_mixin.py
Remi-Gau: > Do we really have to have these two?
yup: the private one is used for the NiftiLabelMasker when the label image is resampled at transform time, it may require some changes to the lut that are stored there.

we cannot touch the lut_ attribute in this case because transform operation are not supposed to change public attributes

nilearn/maskers/_mixin.py
Remi-Gau: yup indeed!

nilearn/maskers/_mixin.py
hndgzkn: One question : But then this does not create an inconsistent internal state? We do not expose the change to outside, but it is actually changed.

Remi-Gau: > But then this does not create an inconsistent internal state? 
you mean where lut_ and _lut_ have different content?

Yes it can if the images that are transformed have a different FOV (affine and/or shape) than those that were used at fit time (or those of the label images - if no image was passed at fit time).

But the idea is that most (if not all) of the properties moved in this PR should be used to inform on the output of transform, so in this case we should rely on _lut_.

I think we now throw a warning to try to nudge the user into avoiding resampling things at transform time, either by passing the images they want to transform at fit time or by using fit_transform, but in case they do not, it may lead to internal inconsistencies.

Remi-Gau: I think we all recognize this is not ideal but short of forbidding any resampling at transform time, I am not sure how to avoid this.

PR #5707, opened 2025-09-23
[MAINT] add wheel size check

Remi-Gau:
- Closes none

- test PR to add a check to make sure the package stays lean

### DIFF
--- a/.github/workflows/testing_install.yml
+++ b/.github/workflows/testing_install.yml
@@ -91,6 +91,8 @@ jobs:
                 include:
                 -   python: '3.10'
                     dependencies: min_plotting
+                -   python: '3.10'
+                    dependencies: min
 
         name: check install size
         defaults:


pyproject.toml
```diff
@@ -116,7 +116,10 @@ max-line-length = 120
 version-file = "nilearn/_version.py"
 
 [tool.hatch.build.targets.wheel]
-exclude = ["nilearn/plotting/tests/baseline"]
+exclude = [
+    "nilearn/plotting/tests/baseline",
+    "nilearn/reporting/tests/baseline"
```
Remi-Gau, line 136: realized while talking with @emdupre that we had added some figures to the package, so trying to add a check in CI to make sure the package size does not explode by mistake

Remi-Gau:
asked chatGPT for a bash script to plot the package size across versions
 
```bash
#!/usr/bin/env bash
set -euo pipefail

PKG="nilearn"
OUTDIR="nilearn_sizes"
CSV="sizes.csv"

mkdir -p "$OUTDIR"
echo "version,file,size_bytes" > "$CSV"

# 1. Get all available versions from PyPI
VERSIONS=$(curl -s "https://pypi.org/pypi/${PKG}/json" | jq -r '.releases | keys[]' | sort -V)

for v in $VERSIONS; do
    echo "Processing $PKG $v ..."
    # Get the URL of a wheel if available, otherwise fall back to sdist
    URL=$(curl -s "https://pypi.org/pypi/${PKG}/${v}/json" | jq -r '
        .urls[] | select(.packagetype == "bdist_wheel") | .url
    ' | head -n1)

    if [ -z "$URL" ]; then
        # fallback to sdist
        URL=$(curl -s "https://pypi.org/pypi/${PKG}/${v}/json" | jq -r '
            .urls[] | select(.packagetype == "sdist") | .url
        ' | head -n1)
    fi

    if [ -z "$URL" ]; then
        echo "⚠️ No artifact found for $v"
        continue
    fi

    FILE="$OUTDIR/$(basename "$URL")"
    if [ ! -f "$FILE" ]; then
        curl -sL "$URL" -o "$FILE"
    fi

    SIZE=$(stat -c%s "$FILE" 2>/dev/null || stat -f%z "$FILE")
    echo "$v,$(basename "$FILE"),$SIZE" >> "$CSV"
done

# 2. Generate plot using Python
python3 <<'EOF'
import pandas as pd

from packaging.version import Version

df = pd.read_csv("sizes.csv")
df["parsed_version"] = df["version"].apply(Version)
df = df.sort_values("parsed_version")

plt.figure(figsize=(10,5))
plt.plot(df["version"], df["size_bytes"]/1e6, marker="o")
plt.xticks(rotation=90)
plt.ylabel("Package size (MB)")
plt.xlabel("Nilearn version")
plt.title("Evolution of Nilearn package size over versions")
plt.tight_layout()
plt.grid(True)
plt.savefig("nilearn_size_evolution.png", dpi=150)
print("✅ Plot saved to nilearn_size_evolution.png")
EOF

```

nilearn_size_evolution

Remi-Gau:
git bisected with this to identify the size regression between 0.12.0 and 0.12.1

- package went above 11MB in this PR: https://github.com/nilearn/nilearn/pull/5578

because of all the figures used to test the figures in reporting: the current PR should fix his by ignoring those figures in the package build

```bash
#!/usr/bin/env bash
set -x -e

# --- Configuration ---
THRESHOLD=$((11 * 1024 * 1024))   # 15 MB, adjust to just above 0.12.0 size

# Clean old build artifacts
rm -rf dist build *.egg-info

# Build wheel
python -m build --wheel

# Get wheel file
WHEEL=$(ls dist/nilearn-*.whl || true)
if [ -z "$WHEEL" ]; then
    echo "❌ No wheel built"
    exit 125   # tell bisect this commit is "skip"
fi

# Measure size
SIZE=$(stat -c%s "$WHEEL" 2>/dev/null || stat -f%z "$WHEEL")
SIZE_MB=$(awk "BEGIN {print $SIZE/1024/1024}")

echo "Wheel size: $SIZE_MB MB (threshold: $(($THRESHOLD/1024/1024)) MB)"

# Decide
if [ "$SIZE" -gt "$THRESHOLD" ]; then
    echo "❌ Commit is BAD (too big)"
    exit 1
else
    echo "✅ Commit is GOOD"
    exit 0
fi

```

maint_tools/plot_package_size.sh
Remi-Gau, line 1: not needed: remove

pyproject.toml
```diff
@@ -61,6 +61,21 @@ doc = [
...
+    "requests==2.25.0",
+    "scikit-learn==1.4.0",
+    "scipy==1.8.0"
+]
+min_plotting = [
+    "nilearn[min]",
+    "matplotlib==3.8.0",
+    "plotly==5.0.0",
+    "kaleido"
+]
```
Remi-Gau, line 78: probably worth refactoring tox.ini thanks to this in a later PR

.github/workflows/testing_install.yml
```diff
@@ -53,3 +60,58 @@ jobs:
...
+            with:
+                python-version: '3.10'
+        -   name: Install dependencies for building Nilearn
+            run: |
+                set -e
+                python -m pip install --upgrade pip
+                pip install --prefer-binary build
+
+        -   name: Check wheel size
+            run: bash build_tools/github/wheel_size_check.sh
```
Remi-Gau:
```suggestion
            run: bash build_tools/github/whole_install_size_check.sh
```

Remi-Gau:
@emdupre 
installing nilearn with all its plotting dependencies (matplotlib, plotly, kaleido) gives an install at around 550 mb

https://github.com/nilearn/nilearn/actions/runs/18404323205/job/52440591960?pr=5707#step:4:15

Should we check the size of the install without plotting dependencies (to give us a lower bound to the bracket of install sizes we are dealing with)?

Remi-Gau:
> @emdupre
> installing nilearn with all its plotting dependencies (matplotlib, plotly, kaleido) gives an install at around 550 mb

not that installing nilearn WITHOUT the plotting dependencies gives 350 MB (so 200 less than wiht the plotting ones)

emdupre:
Sorry for being slow on this ; this is great, thank you ! Happy to see this has been merged 🚀

Issue #5599, opened 2025-08-22
consider updating tempita


Remi-Gau:
we vendor the templating engine we use in externals

there has been for a little while an official package for it: https://github.com/TurboGears/tempita available on pypi

how do people feel about?
- using tempita directly as a regular dependency
- updating the version we vendor

bthirion:
Rather in favor of "regular dependency", but no strong opinion.

Remi-Gau:
in case someones asks: the new tempita package comes with no 'transitive dependency' so it just relies on the standard library

Remi-Gau:
will make a draft PR to check the feasibility of it and see how the reports may be affected: the tests regarding our reports are very lightweight and may not catch some things

Remi-Gau:
other possibility would be to switching to using jinja2:
- pro:
  - it is MUCH MORE widely used / tested templating library and templating language (better support, more examples, better for maintainers and contributors)
  - it is more flexible (when one needs to iterate other things in a template)
  - supports partials native rather than the hack we have to do with tempita
- con:
  - some of our templates may need to be adapted
  - would add 1 extra transitive dependency: MarkupSafe (https://github.com/pallets/jinja/blob/5ef70112a1ff19c05324ff889dd30405b1002044/pyproject.toml#L20)

bthirion:
No  other opinion other than it's more work than what we discussed so far. So probably not high priority.

Remi-Gau:
not high priority indeed

emdupre:
Wow, great catch on the release -- and nice to see a new one after 11 years !

I'm hesitant to move away from vendoring it, only insomuch as I don't think the original reasoning for vendoring has been made obsolete. Namely, that it's not actively developed. Indeed, the [README](https://github.com/TurboGears/tempita) still says :
> It is not actually actively developed, and not an ambitious project. 

I imagine that their version 0.6.0 will work for a while yet, but I could see a future where we have to vendor it again.

Remi-Gau:
Good point.

Jinja would have the advantage of defo being maintained (and should be for a while)

Remi-Gau:
@emdupre in the dev meeting yesterday, there was a soft consensus on switching to jinja 'at some point in the future'. are you ok with this?

emdupre:
sorry for missing the meeting (and this notification, somehow !). 

[...]

PR #5710, opened 2025-09-23
[ENH] use jinja template for GLM reports

Remi-Gau:
- Relates to #5599

- [x] use jinja instead of tempita for GLM reports
- [x] improve tests of GLM report
- [x] leads to some template duplication for now that should be de-duplicated once we support jinja for masker report 
- [x] replace tempita.bunch by sklearn Bunch
- [x] adapt some templates from tempita to jinja
- [x] extract some other parts of the GLM template (navbar) as jinja can easily include other templates
- [x] update changelog
- [x] check package size
- [x] update contributing doc
- [x] change folder struct to

```
nilearn/reporting/data
├── css
│   ├── head.css
│   └── report.css
├── html
│   ├── body_base.jinja
│   ├── head.jinja
│   ├── partials
│   │   └── warnings.jinja
│   └── glm
│        ├── partials
│        │   ├── method_section.jinja
│        │   └── navbar.jinja
│        └── body_glm.jinja
├── js
│   └── carousel.js
└── README.md
```

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -100,7 +100,6 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            # reset to true before merging
             fail-fast: true
             matrix:
                 description: [latest dependencies]


nilearn/reporting/tests/test_glm_reporter.py
```diff
@@ -321,32 +490,23 @@ def test_flm_generate_report_surface_data_error(
...
-    # first level model with one run : no run carousel
-    report_one_run = flm.generate_report(contrasts=contrasts)
-
-    assert 'id="carousel-navbar"' not in report_one_run.__str__()
-
-    # first level model with 2 runs : run carousel
+    # first level model with 3 runs : run carousel
+    # TODO
+    # change name of design matrix columns and use contrast expression
+    # to have different plots for each run
```
Remi-Gau, line 512: in another PR

nilearn/reporting/tests/test_glm_reporter.py
```diff
@@ -251,9 +406,18 @@ def test_fir_delays_in_params(contrasts):
     model = FirstLevelModel(hrf_model="fir", fir_delays=[1, 2, 3])
     model.fit(fmri_data, design_matrices=design_matrices)
 
-    report = model.generate_report(contrasts=contrasts)
-
-    assert "fir_delays" in report.__str__()
+    # FIXME:
+    # matrices were passed at fit time
+    # so fir_delays should not appear in report
+    # as we do not know which HRF was used to build the matrix
```
Remi-Gau, line 416: in another PR

Remi-Gau:
- blocked by https://github.com/nilearn/nilearn/pull/5707

Remi-Gau:
- merged #5707 in this PR to check that using jinja won't blow up package size

pyproject.toml
```diff
@@ -61,6 +64,21 @@ doc = [
...
     "sphinxcontrib-mermaid>=1.0.0"
 ]
+min = [
+    "joblib==1.2.0",
+    "nibabel==5.2.0",
+    "numpy==1.22.4",
+    "pandas==2.2.0",
+    "requests==2.25.0",
+    "scikit-learn==1.4.0",
+    "scipy==1.8.0"
```
Remi-Gau:
```suggestion
    "scipy==1.8.0",
    "jinja2==3.1.0",
```

.github/workflows/testing_minimum.yml
```diff
@@ -13,6 +13,10 @@ name: test minimal version of each dependency
 
 
 on:
+
```
```suggestion
```

Remi-Gau: reverse before merge

nilearn/reporting/data/css/report.css
Remi-Gau, line 1: may consider extracting the navbar css into a separate file so the navbar is not included in reports that do not have a navbar

bthirion:
Hi, The PR definitely looks good buit for that one we need to see the generated doc before pressing the button.

.github/workflows/test_with_tox.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```
revert before merge

Remi-Gau:
> we need to see the generated doc before pressing the button

will ask for a full build to see what it looks like

.github/workflows/build-docs.yml
```diff
@@ -163,7 +163,7 @@ jobs:
     # Check the build type and files examples to generate
     # Upload as an artifact for other jobs.
     build_type:
-        needs: [check_skip_flags, validate_cff]
+        needs: [check_skip_flags]
```
Remi-Gau:
```suggestion
        needs: [check_skip_flags, validate_cff]
```

.github/workflows/testing_minimum.yml
```diff
@@ -13,6 +13,9 @@ name: test minimal version of each dependency
 
 
 on:
+    pull_request:
+        branches:
+        -   '*'
```
```suggestion
```

nilearn/glm/tests/test_second_level.py
```diff
@@ -886,6 +886,7 @@ def test_fmri_img_inputs_errors(confounds):
         SecondLevelModel().fit([*niimgs, []], confounds)
 
 
+@pytest.mark.timeout(0)
```
Remi-Gau:
```suggestion
@pytest.mark.slow
```

nilearn/glm/tests/test_second_level.py
```diff
@@ -1331,6 +1332,7 @@ def test_non_parametric_inference_contrast_computation(
     )
 
 
+@pytest.mark.timeout(0)
```
Remi-Gau:
```suggestion
@pytest.mark.slow
```

Remi-Gau:
@hndgzkn if you have some time, while I try to get a stable full build of the doc to confirm that the GLM reports are OK, maybe you can check that this PR is good code wise.

Remi-Gau:
> Hi, The PR definitely looks good buit for that one we need to see the generated doc before pressing the button.

@bthirion @hndgzkn 

Examples of GLM reports after this PR

https://output.circle-artifacts.com/output/job/f394a681-3ee3-4806-b31e-02f0aacaed35/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html


Can be compared to:

https://nilearn.github.io/dev/modules/generated_reports/glm_reports_examples.html

Remi-Gau:
tests failure are taken care of in https://github.com/nilearn/nilearn/pull/5802

hndgzkn:
> > Hi, The PR definitely looks good buit for that one we need to see the generated doc before pressing the button.
> 
> @bthirion @hndgzkn
> 
> Examples of GLM reports after this PR
> 
> https://output.circle-artifacts.com/output/job/f394a681-3ee3-4806-b31e-02f0aacaed35/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html
> 
> Can be compared to:
> 
> https://nilearn.github.io/dev/modules/generated_reports/glm_reports_examples.html

They seem to be exactly the same. 

I just realized that in the ex and new reports `Design Matrix` and `Contrast` links do not work and the link for `Correlation Matrix` is missing.

And may be in general, figures could be embedded a bit smaller so that the colorbar is not cut.

nilearn/reporting/data/html/glm/partials/method_section.jinja
```diff
@@ -3,52 +3,51 @@
   (version= {{ version }}; RRID:SCR_001362).
 </p>
 
-{{if model_type == "First Level Model"}}
+{% if model_type == "First Level Model" %}
   <p>
```
hndgzkn, line 7: What I write below is out of scope of this PR. Just to have some brainstorming on how to structure reporting package.

I am thinking whether we should keep reporting as a generic module, which provides functions and mixins for the classes/packages who want to add that functionality. Other packages can have dependency on reporting, but reporting should not have dependency on other packages (except packages like utils or image may be).

Beside it does not need to know which packages implement reporting. Thus it need not now about html template for each package implementing reporting.

It is an option to create a clear documentation how to add reporting to a class/package.

For example:

- Inherit from [nilearn.reporting._mixins.ReportingMixin](https://github.com/nilearn/nilearn/pull/5790/files#diff-15ed36357f57b0c94599c9e4c470b6ea21adcba0d955d8436100140895b57a62) and populate attributes specified and implement abstract functions.
- Define an html template file and locate it in `package/data/templates` directory.
- Add a class variable to indicate report template path.
- etc.


I think common templates that will be used in all reports can reside in reports package. For specific templates, it would be an option to put them in the specific package, however I am not sure. How does it sound to you @Remi-Gau ?

Remi-Gau:
> And may be in general, figures could be embedded a bit smaller so that the colorbar is not cut.

that gets tricky because most of the time those reports are viewed as a standalone page or in a notebook and there is more space then

stacking several reports on top of one another on 1 third of the full of dicumentation page is really an edge case! 😉

Remi-Gau:
> I just realized that in the ex and new reports `Design Matrix` and `Contrast` links do not work and the link for `Correlation Matrix` is missing.

let me double check that this is not just an issue of the navbar not working well in the doc page: maybe it works OK when the report is viewed as a single page

Remi-Gau:
> let me double check that this is not just an issue of the navbar not working well in the doc page: maybe it works OK when the report is viewed as a single page

ok this does not work even when viewing as a single page: will flag as an issue to fix in a follow up PR

nilearn/reporting/data/html/glm/partials/method_section.jinja
Remi-Gau: > I am thinking whether we should keep reporting as a generic module, which provides functions and mixins for the classes/packages who want to add that functionality. Other packages can have dependency on reporting, but reporting should not have dependency on other packages (except packages like utils or image may be).

In principle I agree but this morning I was trying to clean that up a bit and it won't be THAT simple.

I wanted to have something like this in terms of import structure because for now only glm and masker do some reporting:

```
nilearn.glm
nilearn.masker
nilearn.reporting
nilearn.plotting
```

Will show you in a demo PR the kind of errors that importlinter was reporting when doing this

Remi-Gau: > For specific templates, it would be an option to put them in the specific package

FYI jinja templating will be easier if our templates are centralize in one place and not in several folders that are not nested

Remi-Gau: > Will show you in a demo PR the kind of errors that importlinter was reporting when doing this

@hndgzkn 
see https://github.com/nilearn/nilearn/pull/5804
reported problems: https://results.pre-commit.ci/run/github/1235740/1761749734.uJ0sftLzSJu6vz6IaAmcqw

Some I think are solvable with some of your suggestions, some others will require more work.

hndgzkn: Yes, I agree. 
I think, for example in plotting, there is some logic which seems mis-located. However it is not straightforward how to fix that. Requires some discussion.

Remi-Gau:
@hndgzkn do you feel ok with the PR?

nilearn/utils/discovery.py
```diff
@@ -218,6 +218,10 @@ def all_displays(type_filter=None):
         List of (name, class), where ``name`` is the display class name as
         string and ``class`` is the actual type of the class.
     """
+    if not is_matplotlib_installed():
+        return []
+    from nilearn.plotting.displays import BaseAxes, BaseSlicer
+
```
Remi-Gau, line 224: fixes an error on import in doc build

hndgzkn:
It seems that doc build fails here as well due to importing `make_glm_report`.

In main branch in local as well, if I try to run `from nilearn.reporting import make_glm_report` in python shell initiated from nilearn environment, I get cyclic import problem.

Remi-Gau:
> It seems that doc build fails here as well due to importing `make_glm_report`.
> 
> In main branch in local as well, if I try to run `from nilearn.reporting import make_glm_report` in python shell initiated from nilearn environment, I get cyclic import problem.

it's fine in this PR somehow

Remi-Gau:
OK will merge as is and let's fix the other issues in follow up PRs

nilearn/utils/tests/test_discovery.py
```diff
@@ -39,7 +39,7 @@ def test_all_functions(
     """Check number of functions in public API."""
     fn = all_functions()
     print(fn)
-    assert len(fn) == 170
```
hndgzkn, line 42: @Remi-Gau Are you sure about this change. `test_discovery.test_all_function` is failing on main at this line.

Remi-Gau: There seems to be some flakyness in this function actually: I was trying to fix some failing tests in https://github.com/nilearn/nilearn/pull/5829 and it seems that sometimes not all expected functions are available through the API (make_glm_reporter is the one that goes missing).

This leads to SOMETIMES this test failing and also explains why we SOMETIMES get these fails in the doc build (because of a whole bunch of warnings related to make_glm_report not existing). 

- I'd like to try to fix things in #5757 by making sure that make_glm_reported is always importable even when no matplotlib engine is present.

PR #5718, opened 2025-09-26
[FIX] fix reporting

### LINKED ISSUE #5717: do not limit import from reporting when plotting dependencies are missing
The import from reporting varies depending on whether matplotlib is there or not.

https://github.com/nilearn/nilearn/blob/bd479cd5fc193fada516c31b2173a511e1d42ec0/nilearn/reporting/__init__.py#L10-L29

However with matplolib missing one can still do this:

```python

flm = FirstLevelModel()
report = flm.generate_report()

masker = NiftiMasker()
```

with some warnings (see below), but for some estimators we can even get a viewable report: so the reporting module is not essential to generating reports

```
/home/remi-gau/github/nilearn/nilearn/nilearn/glm/first_level/first_level.py:1370: UserWarning: nilearn.reporting.glm_reporter and nilearn.reporting.html_report requires nilearn.plotting.
Some dependencies of nilearn.plotting package seem to be missing.
They can be installed with:
/home/remi-gau/github/nilearn/nilearn/tmp/rep.py:4: UserWarning: No plotting back-end detected. Output will be missing figures.
  report = flm.generate_report()
/home/remi-gau/github/nilearn/nilearn/tmp/rep.py:9: ImportWarning: Matplotlib is not imported! No reports will be generated.
```

I think that: 
- we should change what is available in the reporting module depending on what dependencies is installed
- we should let lower level function / methods handle those warnings (pretty sure they are in place anyway)

Remi-Gau:
- Closes #5717

-
-

### DIFF
--- a/nilearn/maskers/nifti_masker.py
+++ b/nilearn/maskers/nifti_masker.py
@@ -472,17 +472,8 @@ class NiftiMasker(ClassNamePrefixFeaturesOutMixin, BaseMasker):
         del y
         check_params(self.__dict__)
 
-        self._report_content = {
-            "description": (
-                "This report shows the input Nifti image overlaid "
-                "with the outlines of the mask (in green). We "
-                "recommend to inspect the report for the overlap "
-                "between the mask and its input image. "
-            ),
-            "warning_message": None,
-            "n_elements": 0,
-            "coverage": 0,
-        }
+        self._init_report_content()
+
         self._overlay_text = (
             "\n To see the input Nifti image before resampling, "
             "hover over the displayed image."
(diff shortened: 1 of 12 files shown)


Remi-Gau:
this PR does too many things: will split it into simpler ones, probably one with just refactoring and another that tries to fix the issue

Remi-Gau:
given that this pr allows to have a method per masker to generate a figure we could create a baseline figure of what that figure should look like and we could test against it

nilearn/maskers/nifti_masker.py
```diff
@@ -404,21 +400,57 @@ def _reporting(self):
...
+
         if self._reporting_data["dim"] == 5:
             msg = (
                 "A list of 4D subject images were provided to fit. "
                 "Only first subject is shown in the report."
             )
             warnings.warn(msg, stacklevel=find_stack_level())
             self._report_content["warning_message"] = msg
+
+        resampl_img = None
```
bthirion:
```suggestion
        resampled_img = None
```

nilearn/maskers/nifti_spheres_masker.py
```diff
@@ -620,6 +635,20 @@ def fit(
 
         return self
 
+    def _init_report_content(self):
+        # content to inject in the HTML template
```
bthirion: mini-docstring ?

bthirion (commented):
LGTM technically.

nilearn/maskers/nifti_spheres_masker.py
Remi-Gau: done

nilearn/maskers/nifti_masker.py
Remi-Gau: done

Remi-Gau:
> given that this pr allows to have a method per masker to generate a figure we could create a baseline figure of what that figure should look like and we could test against it

will do that in a follow up

PR #5719, opened 2025-09-26
[MAINT] Use cache from doc build for benchmarks

### LINKED ISSUE #5657: Benchmarks workflow fails on schedule, only works when triggered manually

man-shu:
- Closes #5657 

- forgot to update Python, Ubuntu version in the workflow, so doing that now
- trying to use doc build cache
- also try running on PR

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -40,10 +40,13 @@ on:
         branches:
         -   main
     schedule:
-    # every Monday at 12noon UTC
-    -   cron: 0 12 * * TUE
+    # every Monday at 9 AM UTC
+    -   cron: 0 9 * * MON
 
     workflow_dispatch:  # This enables manual triggering from GitHub UI
+
+env:
+    NILEARN_DATA: /home/runner/work/nilearn/nilearn/nilearn_data
 
 jobs:
     benchmark:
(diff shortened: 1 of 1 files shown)


bthirion:
do you understand what's going on ?

man-shu:
Ok so just increasing the timeout seems to have done the job...
I will try a cron job now

bthirion:
what is the status of that PR ?

Remi-Gau:
> what is the status of that PR ?

we talked about it yesterday with @man-shu : this should be ready to merge and then we can add more benchmark and fix things as they come.

Remi-Gau:
actually @man-shu I am going to merge as is: we can try to see if there is a way to run benchmark as "strict" so that any failure is CI actually shows up

PR #5679, opened 2025-09-12
[ENH] Add A MultiSurfaceMasker class 

Remi-Gau:
- Relates to #5539 

Changes proposed in this pull request:

- moves several methods from the multi maskers into the MultiMixin
- create a MultiSurfaceMasker
- changelog
- adapt tests

Issue #5696, opened 2025-09-22
[BUG] Import regression from 0.12.0 to 0.12.1


ghisvail:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Guix

### Python version

- [ ] 3.13
- [ ] 3.12
- [x] 3.11
- [ ] 3.10

### nilearn version

0.12.1

### Expected behavior

[mapca](https://github.com/ME-ICA/mapca) uses utility functions from nilearn such as `check_niimg_3d`.

The package used to build and test fine with version 0.12.0 of nilearn, it fails with an import error on version 0.12.1.

Now I totally understand that functions under the private `_utils` are not meant for public use. That being said, nothing is really private in Python and client code can and will use these functions if they are found useful outside nilearn apparently.

I would propose that such changes in the private utils package are kept for minor version bumps (until 1.0 happens). No one would expect this kind of breakage for a patch release I think.

### Current behavior & error messages

This is what I got:

```bash
starting phase `check'
Using pytest
============================= test session starts ==============================
platform linux -- Python 3.11.11, pytest-8.3.3, pluggy-1.5.0 -- /gnu/store/isaa890hpwgd8s3jk44hv2fdqhqvasn0-python-wrapper-3.11.11/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/tmp/guix-build-python-mapca-0.0.5.drv-0/mapca-0.0.5/.hypothesis/examples')
rootdir: /tmp/guix-build-python-mapca-0.0.5.drv-0/mapca-0.0.5
configfile: pyproject.toml
plugins: hypothesis-6.54.5, cov-6.0.0
collecting ... collected 0 items / 1 error

==================================== ERRORS ====================================
_________________________ ERROR collecting mapca/tests _________________________
[...]
mapca/__init__.py:6: in <module>
    from mapca.mapca import MovingAveragePCA, ma_pca
mapca/mapca.py:25: in <module>
    from nilearn._utils import check_niimg_3d, check_niimg_4d
E   ImportError: cannot import name 'check_niimg_3d' from 'nilearn._utils' (/gnu/store/1yhs76dyf6ilmla6y00z61fgsxz2yvdk-python-nilearn-0.12.1/lib/python3.11/site-packages/nilearn/_utils/__init__.py)
=========================== short test summary info ============================
[...]

PR #5724, opened 2025-10-03
[REF] move `check_embedded_masker` to nilearn.masker

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/5679#discussion_r2368648758

- move one masker validation function and its tests out of nilearn._utils

### DIFF
--- a/nilearn/maskers/masker_validation.py
+++ b/nilearn/maskers/masker_validation.py
@@ -17,7 +17,7 @@ from nilearn.maskers import (
 )
 
 
-def check_embedded_masker(estimator, masker_type="multi_nii", ignore=None):
+def check_embedded_masker(estimator, masker_type, ignore=None):
     """Create a masker from instance parameters.
 
     Base function for using a masker within a BaseEstimator class
@@ -39,8 +39,7 @@ def check_embedded_masker(estimator, masker_type="multi_nii", ignore=None):
     instance : object, instance of BaseEstimator
         The object that gives us the values of the parameters
 
-    masker_type : {"multi_nii", "nii", "surface", "multi_surface"}, \
-                  default="mutli_nii"
+    masker_type : {"multi_nii", "nii", "surface", "multi_surface"}
         Indicates whether to return a MultiNiftiMasker, NiftiMasker,
         SurfaceMasker, or a MultiSurfaceMasker.
 
--- a/tox.ini
+++ b/tox.ini
@@ -317,3 +317,14 @@ commands =
     ; but did not run the maint_tools/citation_cff_maint.py script.
     python maint_tools/citation_cff_maint.py
 	make -C doc linkcheck
+
+
+[testenv:archi]
+description = build figure of the dependencies between Nilearn subpackages
+skip_install = true
+deps =
+    pydeps
+passenv =
+    {[global_var]passenv}
+commands =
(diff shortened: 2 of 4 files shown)


nilearn/_utils/masker_validation.py
```diff
@@ -1,163 +1,15 @@
...
-from nilearn._utils.class_inspect import get_params
-from nilearn._utils.logger import find_stack_level
 from nilearn._utils.param_validation import check_is_of_allowed_type
-from nilearn._utils.tags import is_glm
-from nilearn.surface import SurfaceImage
+from nilearn.surface.surface import SurfaceImage
 from nilearn.typing import NiimgLike
 
 
-def check_embedded_masker(estimator, masker_type="multi_nii", ignore=None):
```
Remi-Gau, line 16: moved to nilearn.maskers.masker_validation

nilearn/maskers/masker_validation.py
```diff
@@ -0,0 +1,155 @@
...
+    In all case, we forward system parameters of instance to new masker :
+    memory, memory_level, verbose, n_jobs
+
+    Parameters
+    ----------
+    instance : object, instance of BaseEstimator
+        The object that gives us the values of the parameters
+
+    masker_type : {"multi_nii", "nii", "surface", "multi_surface"}, \
+                  default="mutli_nii"
```
bthirion: Not a strong opinion, and I'm not sure that it matters, but isn't 'nii' the most standard use ?

Remi-Gau:
> The PR is technically good, but I still have a hard time understanding the motivation: is it to make the dependency graph cleaner ?

That was the original comment that motivated the refactoring

> Could `check_embedded_masker` function be moved to `nilearn.maskers`? It seems like it belongs there more. It would help avoid this import inside the function as well.
> 
> It is imported only by `glm`, `decoders` and `decomposition` packages which already have dependency on `nilearn.maskers`.

In general it's part of the general tendency to move things out of nilearn._utils and keep them in the subpackage where they belong

nilearn/maskers/masker_validation.py
Remi-Gau: let me check

Remi-Gau: actually in practice I think we never rely on the default and we always explicit pass a value so I am removing the default

tox.ini
```diff
@@ -317,3 +317,14 @@ commands =
     ; but did not run the maint_tools/citation_cff_maint.py script.
     python maint_tools/citation_cff_maint.py
 	make -C doc linkcheck
+
+
+[testenv:archi]
```
Remi-Gau, line 322: unrelated but I needed if this PR changed the architecture of the package and had to do a 10 minutes github dumpster dive to find the right command, so it should be easier to find now

emdupre:
I guess this also relates to https://github.com/nilearn/nilearn/issues/5696 ? Similarly, `check_embedded_masker` is relatively commonly used from `_utils`

Remi-Gau:
> I guess this also relates to #5696 ? Similarly, `check_embedded_masker` is relatively commonly used from `_utils`

are you sure? I think I added the function 'not long ago'.

does that meean you think we should mention this move in the changelog given https://github.com/nilearn/nilearn/issues/5696 ?

am afraid that if we keep adding to the changelog every time we move a private functiion, we'll be just adding noise to our changelog

PR #5729, opened 2025-10-03
[DEP] no bool for standardize

### LINKED ISSUE #5684: remove use of booleans for standardize
We could start another deprecation cycle for this.
Note 'zscore' will be deprecated in the near future.
So in the end we would have:
- 'zscore_sample' 
- 'psc' 
- None

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5668#discussion_r2343510040_

Remi-Gau:
- Closes #5684 

- start deprecation cycle to avoid using bool for standardize
- [x] check doc strings
- [x] reduce warnings in tests and in doc build

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -74,7 +74,7 @@ repos:
 
     # Lint and format Python code
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.0
+    rev: v0.14.1
     hooks:
     -   id: ruff-check
         # args: [--statistics]


nilearn/signal.py
```diff
@@ -59,15 +59,17 @@ def standardize_signal(
...
-    check_parameter_in_allowed(
-        standardize,
-        allowed=[True, False, "psc", "zscore", "zscore_sample"],
-        parameter_name="standardize",
-    )
 
     signals = _detrend(signals, inplace=False) if detrend else signals.copy()
 
-    if standardize:
+    assert not isinstance(standardize, bool)
```
Remi-Gau, line 67: add an assert here because at this level we should not have a boolean anymore

nilearn/_utils/docs.py
```diff
@@ -1080,21 +1081,44 @@ def custom_function(vertices):
...
+
+
 """
 # TODO (nilearn >= 0.14.0) update to ..versionchanged
 deprecation_notice = """
 
     .. nilearn_deprecated:: 0.10.1
 
-        The default will be changed to ``'zscore_sample'``
+        The default will be changed to ``'zscore_sample'`` (or ``True``)
```
bthirion: what does this mean ?

bthirion (commented):
LGTM overall.

nilearn/_utils/docs.py
Remi-Gau: yup that makes no sense, let me fix that

Remi-Gau: fixed

Issue #4326, opened 2024-03-19
[ENH] cluster threshold for surface GLM


mszinte:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe your proposed enhancement in detail.

As I use these days mostly (if not only) surface analysis of fMRI data, I use to the surface GLM analysis of nilearn, however I couldn't find the cluster_threshold method available for volume analysis. 
The idea of cluster analysis for surface would be to discard isolated vertices from the GLM output in order to generate thresholded maps with small cluster of vertices removed following a cluster_threshold argument.
Here is the volume based equivalent: 
https://github.com/nilearn/nilearn/blob/436cfb2196b69054d876e592a59c5a391b36f4c8/nilearn/glm/thresholding.py#L184
After early discussion I feel that for this improvement (@Remi-Gau @bthirion, @ulascombes) , we would need first:
* to associate each vertex to an area
* define connection component of each vertex
* set rules of connection of a vertices (maybe geodesic distances)

### Benefits to the change

This analysis is to my opinion an important missing step for the basic GLM surface analysis of nilearn, it will benefit to the community of users who moved toward fMRI surface analyses.

### Pseudocode for the new behavior, if applicable

```python
# insert your code below

```

bthirion:
Well, you don't want to have to compute geodesic distances to obtain connected clusters. You simply need to use the topology of the mesh (corresponding to the face definition).
But I agree that this would be useful.

Remi-Gau:
started working on this
- first step will be to identify clusters
- later on will work on actually doing stats on those clusters
- then will come getting report tables with just cluster label, size, stats. Never heard of a way to report surface cluster coordinates, but I am all ears.

bthirion:
For coordinates, the best option to me is not to provide any.
If we really want, take the 3D coordinates of the vertices ?
For fsaverage pial/white, they're supposed to be in Talairach space.

Remi-Gau:
> For coordinates, the best option to me is not to provide any.

yeah I would go for this for now, because most other options won't generalize well I think

Remi-Gau:
- reopening given that #5764 is not done yet

PR #5733, opened 2025-10-06
[ENH] add cluster level inference for surface data

Remi-Gau:
- Relates to #4326

- add cluster level inference for surface data


Remi-Gau:
maybe worth showing in an example?

bthirion:
> maybe worth showing in an example?

I think so.

Remi-Gau:
> > maybe worth showing in an example?
> 
> I think so.

we don't really have a group level GLM example with surface data, I could try use this in a subject level GLM example

doc/changes/latest.rst
```diff
@@ -38,7 +38,9 @@ Enhancements
 
 - :bdg-success:`API` Add a :class:`~maskers.MultiSurfaceMasker` (:gh:`5679` by `Rémi Gau`_).
 
-- :bdg-success:`API` Add support for ``cluster_threshold`` for :func:`~image.threshold_img`, :func:`~glm.threshold_stats_img`, :func:`~reporting.make_glm_report`, :meth:`~glm.first_level.FirstLevelModel.generate_report` and :meth:`~glm.second_level.SecondLevelModel.generate_report` (:gh:`5715` by `Rémi Gau`_).
+- :bdg-success:`API` Add support for ``cluster_threshold`` for :class:`~surface.SurfaceImage` in :func:`~image.threshold_img`, :func:`~glm.threshold_stats_img`, :func:`~reporting.make_glm_report`, :meth:`~glm.first_level.FirstLevelModel.generate_report` and :meth:`~glm.second_level.SecondLevelModel.generate_report` (:gh:`5715` by `Rémi Gau`_).
+
+- :bdg-success:`API` Add support for :class:`~surface.SurfaceImage` in :func:`~glm.cluster_level_inference` (:gh:`XXXX` by `Rémi Gau`_).
```
```suggestion
- :bdg-success:`API` Add support for :class:`~surface.SurfaceImage` in :func:`~glm.cluster_level_inference` (:gh:`5733` by `Rémi Gau`_).
```

nilearn/glm/thresholding.py
```diff
@@ -188,37 +189,109 @@ def cluster_level_inference(
...
+                        cluster_vals, hommel_value, alpha
+                    )
+                    proportion_true_discoveries[labels == label_] = proportion
+
+            tmp_img = masker.inverse_transform(proportion_true_discoveries)
+            data[hemi] = tmp_img.data.parts[hemi]
+
+        proportion_true_discoveries_img = new_img_like(stat_img, data)
+
+    else:
```
bthirion:
```suggestion
    else:  # data are volume-based
```

doc/changes/latest.rst
```diff
@@ -40,7 +40,11 @@ Enhancements
 
 - :bdg-success:`API` Add a :class:`~maskers.MultiSurfaceMasker` (:gh:`5679` by `Rémi Gau`_).
 
-- :bdg-success:`API` Add support for ``cluster_threshold`` for :func:`~image.threshold_img`, :func:`~glm.threshold_stats_img`, :func:`~reporting.make_glm_report`, :meth:`~glm.first_level.FirstLevelModel.generate_report` and :meth:`~glm.second_level.SecondLevelModel.generate_report` (:gh:`5715` by `Rémi Gau`_).
+- :bdg-success:`API` Add support for ``cluster_threshold`` for :class:`~surface.SurfaceImage` in :func:`~image.threshold_img`, :func:`~glm.threshold_stats_img`, :func:`~reporting.make_glm_report`, :meth:`~glm.first_level.FirstLevelModel.generate_report` and :meth:`~glm.second_level.SecondLevelModel.generate_report` (:gh:`5715` by `Rémi Gau`_).
+
+- :bdg-success:`API` Add support for :class:`~surface.SurfaceImage` in :func:`~glm.cluster_level_inference` (:gh:`5733` by `Rémi Gau`_).
+
+- :bdg-success:`API` Add support for non-parametric cluster-level inference on :class:`~surface.SurfaceImage` in :func:`~glm.second_level.non_parametric_inference` (:gh:`5733` by `Rémi Gau`_).
```
Remi-Gau: ok maybe I will address this in a separate PR to keep things small

Remi-Gau: - non parametric cluster level inference moved to https://github.com/nilearn/nilearn/pull/5764

bthirion:
Do you intend to add an example in this PR ?

Remi-Gau:
> Do you intend to add an example in this PR ?

just added one: will add link to it when doc build is done

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -170,38 +170,68 @@
...
     )
 
 show()
 
 # %%
-# Or we can save as an html file.
+# TODO: add explanation
+proportion_true_discoveries_img = cluster_level_inference(
+    results["audio - visual"], threshold=[3, 4, 5, 6], alpha=0.05
+)
```
Remi-Gau, line 244: was trying and we get "funky results" when trying to use this function with negative thresholds: is that something we want to support?

bthirion: In theory, there is no reason why the results should be weird.  
Maybe open a separate issue for that ?
I'll take  a look.

Remi-Gau:
Example usage:

https://output.circle-artifacts.com/output/job/4ea3699b-0ad4-47a8-80e1-2efc1372c09e/artifacts/0/dev/auto_examples/04_glm_first_level/plot_localizer_surface_analysis.html#sphx-glr-auto-examples-04-glm-first-level-plot-localizer-surface-analysis-py

Should we use it on all the contrasts of this example?

Note that the plots of the contrasts is two sided, but the proportion true positive is not.

examples/04_glm_first_level/plot_localizer_surface_analysis.py
Remi-Gau: posting it here first: this snippet will only work on this PR for now

a bit long but should show the issue

tries to get proprotion true postive with a threshold of 3 and -3

```python
from nilearn.datasets import fetch_localizer_first_level
from nilearn.datasets import load_fsaverage
import numpy as np
from scipy.stats import norm

from nilearn.datasets import load_fsaverage_data
from nilearn.glm import cluster_level_inference
from nilearn.plotting import plot_surf_stat_map, show


data = fetch_localizer_first_level()
t_r = data.t_r
slice_time_ref = data.slice_time_ref


fsaverage5 = load_fsaverage()
surface_image = SurfaceImage.from_volume(
    mesh=fsaverage5["pial"],
    volume_img=data.epi_img,
)

glm = FirstLevelModel(
    t_r=t_r,
    slice_time_ref=slice_time_ref,
    hrf_model="glover + derivative",
    minimize_memory=False,
).fit(run_imgs=surface_image, events=data.events)

design_matrix = glm.design_matrices_[0]
contrast_matrix = np.eye(design_matrix.shape[1])

basic_contrasts = {
    column: contrast_matrix[i]
    for i, column in enumerate(design_matrix.columns)
}

basic_contrasts["audio"] = (
    basic_contrasts["audio_left_hand_button_press"]
    + basic_contrasts["audio_right_hand_button_press"]
    + basic_contrasts["audio_computation"]
    + basic_contrasts["sentence_listening"]
)

basic_contrasts["visual"] = (
    basic_contrasts["visual_left_hand_button_press"]
    + basic_contrasts["visual_right_hand_button_press"]
    + basic_contrasts["visual_computation"]
    + basic_contrasts["sentence_reading"]
)

contrasts = {
    "audio - visual": basic_contrasts["audio"] - basic_contrasts["visual"],
}

p_val = 0.001
p001_uncorrected = norm.isf(p_val)

fsaverage_data = load_fsaverage_data(data_type="sulcal")

results = {}
for contrast_id, contrast_val in contrasts.items():
    results[contrast_id] = glm.compute_contrast(contrast_val, stat_type="t")

for contrast_id, z_score in results.items():
    hemi = "left"
    if contrast_id == "(left - right) button press":
        hemi = "both"

    plot_surf_stat_map(
        surf_mesh=fsaverage5["inflated"],
        stat_map=z_score,
        hemi=hemi,
        title=contrast_id,
        threshold=p001_uncorrected,
        bg_map=fsaverage_data,
    )

proportion_true_discoveries_img = cluster_level_inference(
    results["audio - visual"], threshold=3, alpha=0.05
)

plot_surf_stat_map(
    surf_mesh=fsaverage5["inflated"],
    stat_map=proportion_true_discoveries_img,
    hemi=hemi,
    threshold=0.0,
    cmap="inferno",
    title="audio - visual, proportion true positives, threshold= 3",
    bg_map=fsaverage_data,
)

proportion_true_discoveries_img = cluster_level_inference(
    results["audio - visual"], threshold=-3, alpha=0.05
)

plot_surf_stat_map(
    surf_mesh=fsaverage5["inflated"],
    stat_map=proportion_true_discoveries_img,
    hemi=hemi,
    threshold=0.0,
    cmap="inferno",
    title="audio - visual, proportion true positives, threshold= -3",
    bg_map=fsaverage_data,
)

show()
```

Figure_1g

Figure_fze2

Figurefs_3

Remi-Gau: will check if we have similar issues with negative threshold on volume data.

Remi-Gau: Modified one of the second level examples and tested it on main:

the result is not as funky but still feels wrong

```python
import pandas as pd
from scipy.stats import norm

from nilearn import plotting
from nilearn.glm import cluster_level_inference
from nilearn.glm.second_level import SecondLevelModel

n_subjects = 16
data = fetch_localizer_contrasts(
    ["left vs right button press"],
    n_subjects,
)


second_level_input = data["cmaps"]
design_matrix = pd.DataFrame(
    [1] * len(second_level_input), columns=["intercept"]
)


second_level_model = SecondLevelModel(smoothing_fwhm=8.0, n_jobs=2)
second_level_model = second_level_model.fit(
    second_level_input, design_matrix=design_matrix
)


z_map = second_level_model.compute_contrast(output_type="z_score")


p_val = 0.001
p001_uncorrected = norm.isf(p_val)


proportion_true_discoveries_img = cluster_level_inference(
    z_map, threshold=[3, 4, 5], alpha=0.05
)

proportion_true_discoveries_img_deactivated = cluster_level_inference(
    z_map, threshold=-3, alpha=0.05
)


plotting.plot_stat_map(
    proportion_true_discoveries_img,
    threshold=0.0,
    display_mode="z",
    vmax=1,
    cmap="inferno",
    title="group left-right button press, proportion true positives",
)

plotting.plot_stat_map(
    z_map,
    threshold=p001_uncorrected,
    display_mode="z",
    title="group left-right button press (uncorrected p < 0.001)",
)

plotting.plot_stat_map(
    proportion_true_discoveries_img_deactivated,
    threshold=p001_uncorrected,
    display_mode="z",
    title="group right-left button press, proportion true positives",
)

plotting.show()
```

Figure_1

Figure_2

Figure_3

Remi-Gau: In general I would fix this in a separate PR from the present one: though in this one I could simply add a check to make sure thresholds are positive and negative thresholds are not implemented yet

bthirion: Well , in the last case it returns a lower bound of the  TDP for the quasi-whole image,  which is indeed low.
What would you expect ?

bthirion:
I would consider only one contrasts. And yes, this is a one-sided thing. The problem is that any inference using some kind of cluster forming threshold becomes ambiguous in two-sided settings.
Would it be OK to warn that this is only defined for one-sided tests ?

examples/04_glm_first_level/plot_localizer_surface_analysis.py
Remi-Gau: ah yes indeed
For some reason I would have expected that if one does:

```python
cluster_level_inference(
    z_map,
    threshold = -3,
    alpha=0.05
)
```

it would in fact be doing something like

```python
cluster_level_inference(
    math_img("z_map * -1", zmap=zmap),
    threshold = 3,
    alpha=0.05
)
```

Remi-Gau:
> Would it be OK to warn that this is only defined for one-sided tests ?

not sure we can add a warning: under what circumstances would this warning be thrown?

or did you mean a warning in the doc string and in the example ?

though we could probably throw warnings if the threshold or list of threshold passed contains negative values (see my assuption in this comment https://github.com/nilearn/nilearn/pull/5733#discussion_r2445175997)

examples/04_glm_first_level/plot_localizer_surface_analysis.py
Remi-Gau: but we do agree that the surface results are unexpected?

bthirion: The behavior with negative thresholds where "everything gets inverted" has been introduced recently. I don't really like it, and indeed we are not following it here.

Remi-Gau: > where "everything gets inverted" has been introduced recently

I think it has been introduced only for things that support one-sided tests or thresholding: since this is not the case of this function, I think I just have the wrong expectations.

nilearn/glm/thresholding.py
```diff
@@ -185,13 +188,104 @@ def cluster_level_inference(
...
+    original_threshold = threshold
     if not isinstance(threshold, list):
         threshold = [threshold]
 
+    if any(x < 0 for x in threshold):
+        raise ValueError(
+            "'threshold' cannot be negative or "
+            "contain negative values. "
+            f"Got: 'threshold={original_threshold}'."
+        )
```
Remi-Gau, line 204: For now decided to go for a ValueError rather than a not implemented error

Remi-Gau:
> I would consider only one contrasts. And yes, this is a one-sided thing. The problem is that any inference using some kind of cluster forming threshold becomes ambiguous in two-sided settings. Would it be OK to warn that this is only defined for one-sided tests ?

see https://github.com/nilearn/nilearn/pull/5733/files#r2569474394

for now I went with actually throwing an error

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -201,11 +207,42 @@
...
+
+plot_surf_stat_map(
+    surf_mesh=fsaverage5["inflated"],
+    stat_map=proportion_true_discoveries_img,
+    hemi="left",
+    cmap="inferno",
+    title="audio - visual, proportion true positives",
+    bg_map=fsaverage_data,
+)
+
```
bthirion, line 259: Here there is a fixed value per cluster. We should find a  way to display the value without interpolation.

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -201,11 +207,42 @@
 show()
 
 # %%
-# Or we can save as an html file.
+# Cluster-level inference
+# -----------------------
+#
+# We can also perform cluster-level inference for a given contrast.
```
bthirion:
```suggestion
# We can also perform cluster-level inference (aka "All resolution Inference") for a given contrast.
# This gives a high-probability lower bound on the proportion of true discoveries in each cluster
```

bthirion (commented):
LGTM. A couple of improvements are needed for he example.

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -170,25 +170,31 @@
 
 # %%
 # Let's estimate the contrasts by iterating over them.
+#
+# We then plot each contrast on the inflated fsaverage mesh,
+# together with a suitable background to give an impression
+# of the cortex folding.
+#
+# We plot threshold the contrasts at uncorrected p < 0.001.
+from scipy.stats import norm
```
hndgzkn:
```suggestion
# We plot thresholded contrasts at uncorrected p < 0.001.
from scipy.stats import norm
```

nilearn/glm/thresholding.py
```diff
@@ -152,13 +152,16 @@ def cluster_level_inference(
...
 
-    mask_img : Niimg-like object, default=None
+    mask_img : Niimg-like object, or :obj:`~nilearn.surface.SurfaceImage` \
+        or None, default=None
         mask image
 
-    threshold : :obj:`list` of :obj:`float`, default=3.0
+    threshold : Positive :obj:`float`, :obj:`int`, \
+                 or :obj:`list` of \
+                 positive :obj:`float` or :obj:`int`, default=3.0
```
hndgzkn:
```suggestion
    threshold : Non-negative :obj:`float`, :obj:`int`, \
                 or :obj:`list` of \
                 non-negative :obj:`float` or :obj:`int`, default=3.0
```

nilearn/glm/thresholding.py
```diff
@@ -185,13 +189,104 @@ def cluster_level_inference(
...
+    if any(x < 0 for x in threshold):
+        raise ValueError(
+            "'threshold' cannot be negative or "
+            "contain negative values. "
+            f"Got: 'threshold={original_threshold}'."
+        )
+
+    is_surface = isinstance(stat_img, SurfaceImage) or isinstance(
+        mask_img, SurfaceImage
+    )
```
hndgzkn: It seems to be not used elsewhere. It could directly be used in if condition.

hndgzkn (approved):
LGTM with minor comments.

nilearn/glm/thresholding.py
Remi-Gau: done

examples/04_glm_first_level/plot_localizer_surface_analysis.py
Remi-Gau: Hmmm... Then there is something wrong:

because these are the unique values contained in the `proportion_true_discoveries_img`

```
[-0.          0.2         0.25        0.28571429  0.33333333  0.4
  0.5         0.55555556  0.57142857  0.625       0.63636364  0.66666667
  0.68        0.71428571  0.75        0.8         0.875       0.9
  0.90909091  1.        ]
```

Remi-Gau: checking on the volume example of this: we should be getting as many unique values in the image as we passed in the threshold value, no?

bthirion: There has to be one value per cluster.

bthirion:
Hm. There seems to be something fishy with the cluster definition...

Remi-Gau:
```python
import numpy as np

from scipy.stats import norm

from nilearn.glm import    cluster_level_inference

shape = (7, 8, 9)

affine_eye = np.eye(4)

p = np.prod(shape)
data_norm_isf = norm.isf(np.linspace(1.0 / p, 1.0 - 1.0 / p, p)).reshape(shape)
data = data_norm_isf
data[2:4, 5:7, 6:8] = 5.0
stat_img = Nifti1Image(data, affine_eye)

for threshold in [2.5, [2.5, 3.5], [2.5, 3.0, 3.5]]:
    th_map = cluster_level_inference(stat_img, threshold=threshold)
    vals = th_map.get_fdata()
    print("Thresholded map unique values:", np.unique(vals))
```

at least the code above gives me:

```
Thresholded map unique values: [-0.  1.]
Thresholded map unique values: [-0.  1.]
Thresholded map unique values: [-0.  1.]
```

Remi-Gau:
OK but when testing on more realistic data I get:

```python
import numpy as np

from nilearn.glm import cluster_level_inference

for threshold in [2.5, [2.5, 3.5], [2.5, 3.0, 3.5]]:
    th_map = cluster_level_inference(img, threshold=threshold)
    vals = th_map.get_fdata()
    print("Thresholded map unique values:", np.unique(vals))
```

```
Thresholded map unique values: [0.         0.52850878 0.67115903]
Thresholded map unique values: [0.         0.52850878 0.67115903 0.69791669 0.79207921 0.89490652]
Thresholded map unique values: [0.         0.52850878 0.63421053 0.67115903 0.69791669 0.77916855 0.79207921 0.89490652]
```

Remi-Gau:
> OK but when testing on more realistic data I get:

Just checked and this behavior has been there since at least version 0.8.0

bthirion:
> OK but when testing on more realistic data I get:
> 
> ```python
> import numpy as np
> 
> from nilearn.glm import cluster_level_inference
> 
> from nilearn.datasets import load_sample_motor_activation_image
> 
> 
> img = load_sample_motor_activation_image()
> for threshold in [2.5, [2.5, 3.5], [2.5, 3.0, 3.5]]:
>     th_map = cluster_level_inference(img, threshold=threshold)
>     vals = th_map.get_fdata()
>     print("Thresholded map unique values:", np.unique(vals))
> ```
> 
> ```
> Thresholded map unique values: [0.         0.52850878 0.67115903]
> Thresholded map unique values: [0.         0.52850878 0.67115903 0.69791669 0.79207921 0.89490652]
> Thresholded map unique values: [0.         0.52850878 0.63421053 0.67115903 0.69791669 0.77916855 0.79207921 0.89490652]
> ```

IIUC what happens is that we jointly consider the voxels obtained with different thresholds on the same map. So this makes sense. Sorry for the noise.

Remi-Gau:
> IIUC what happens is that we jointly consider the voxels obtained with different thresholds on the same map. So this makes sense. Sorry for the noise.

sounds good

will keep the tests that check the behavior using 'realistic' data because they seem to allow more complete behavioral coverage

bthirion:
> > IIUC what happens is that we jointly consider the voxels obtained with different thresholds on the same map. So this makes sense. Sorry for the noise.
> 
> sounds good
> 
> will keep the tests that check the behavior using 'realistic' data because they seem to allow more complete behavioral coverage

One further thought: we could take more margin between threshold to make the outcome look better, e.g. [2.5, 3.5, 4.5].

Remi-Gau:
OK that should be good to go

PR #5735, opened 2025-10-06
[MAINT] add test for using target_affine as list

### LINKED ISSUE #5691: turns out from some tests that affine can be a list of numbers and not just an array: not this is true for ALL the places where we use TargetAffine: to check in a follow up
turns out from some tests that affine can be a list of numbers and not just an array: not this is true for ALL the places where we use TargetAffine: to check in a follow up

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5656#discussion_r2357772942_

Remi-Gau:
- Closes #5691

- update doc to mention that target affine can be a list of numeric
- update tests to make sure  target affine as a list of numeric can be used and that the proper errors are thrown when the list has the wrong dimensions
-

### DIFF
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -1162,9 +1162,12 @@ t_r : :obj:`float` or :obj:`int` or None, default=None
 
 # target_affine
 docdict["target_affine"] = """
-target_affine : :class:`numpy.ndarray` or None, default=None
+target_affine : :class:`numpy.ndarray`, :obj:`list` of numeric, or None, \
+       default=None
     If specified, the image is resampled corresponding to this new affine.
-    `target_affine` can be a 3x3 or a 4x4 matrix.
+    `target_affine` can be a 3x3 or a 4x4 array.
+    If a list of numeric values is passed,
+    it must equivalent to a 3x3 or a 4x4 array.
 """
 
 # target_shape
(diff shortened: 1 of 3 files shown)


Remi-Gau:
probably worth using something similar as to what nibabel uses for the doc string (e.g: array-like)

nilearn/typing.py
```diff
@@ -67,7 +67,7 @@
 SmoothingFwhm: TypeAlias = float | int | np.floating | np.integer | None
 Standardize: TypeAlias = Literal["zscore", "zscore_sample", "psc", True, False]
 StandardizeConfounds: TypeAlias = bool
-TargetAffine: TypeAlias = ndarray | list | None
+TargetAffine: TypeAlias = ndarray | list | tuple | None
```
bthirion, line 70: In which case would it be a tuple ?

Remi-Gau: not sure but since tuples can be turned into arrays or used as affine by nibabel we probably don't want to disallow it

```python
import numpy as np
import nibabel as nib

affine = (
    (1, 2, 3, 0),
    (4, 5, 6, 0),
    (7, 8, 9, 0),
    (10, 11, 12, 0),
)

np.asarray(affine)

nib.Nifti1Image(np.ones((10, 11, 12)), affine=affine)

```

PR #5740, opened 2025-10-07
[STY] Apply ruff preview rules

DimitriPapadopoulos:

- Start applying preview rules.
- Some of them are questionable, such as 52ebec6d4e529cc36fea2eebad2e7fda0c9ffcaf.

### DIFF
--- a/nilearn/datasets/tests/test_neurovault.py
+++ b/nilearn/datasets/tests/test_neurovault.py
@@ -1142,7 +1142,7 @@ def test_timeout_error(capsys, request_mocker):
     assert len(data.images) == 0
 
     captured = capsys.readouterr()
-    match = re.search("Try increasing", captured.out)
+    match = re.search(r"Try increasing", captured.out)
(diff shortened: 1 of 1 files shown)


Remi-Gau:
> * Some of them are questionable, such as [52ebec6](https://github.com/nilearn/nilearn/commit/52ebec6d4e529cc36fea2eebad2e7fda0c9ffcaf).

agreed!!

especially in an example I'd rather stick to simpler python syntax: can you revert this one and update the pyproject.toml so this 'error' is not autofixed in the future?

DimitriPapadopoulos:
The thing is, `FURB113` is relevant in most cases. It just doesn't apply well to this specific case, because each new item depends on the previous item. It might not apply either to examples which requires simple Python syntax.

Remi-Gau:
test failure is unrelated

Remi-Gau:
thanks for this @DimitriPapadopoulos

PR #5742, opened 2025-10-08
reduce warning maps masker

Remi-Gau:
- Closes #

-
-


Remi-Gau:
hmm... this warning does not happen in the doc build so I suspect the warning happens because of the maps we used in tests that must be 'binary-like'

in any case, those are warnings the user cannot do anything about so they probably should be silenced

Remi-Gau:
> No strong opinon, but I'd rather keep linear as a default...

yeah me too: I was playing around to see if this change would remove all warnings from the tests

Remi-Gau:
- [x] should add a regression test to make sure the warning is not thrown

bthirion (approved):
The changes LGTM. The CI failure looks unrelated ?

Remi-Gau:
> The CI failure looks unrelated ?

yes there are unrelated

PR #5744, opened 2025-10-08
[MAINT] add tests to check figures surface masker reports

Remi-Gau:
- Closes none

- [x] ensure all maskers have a _create_figure_for_report method that returns a figure
- [x] add tests to check figures surface masker reports
- [x] do multi nifti maskers

### DIFF
--- a/nilearn/reporting/tests/test_baseline_comparisons.py
+++ b/nilearn/reporting/tests/test_baseline_comparisons.py
@@ -169,7 +169,6 @@ def test_nifti_masker_create_figure_for_report(src_masker, mask_img, img):
     masker.fit(img)
 
     displays = masker._create_figure_for_report()
-    assert isinstance(displays[0], str)
     return displays[0]
 
 


tox.ini
```diff
@@ -144,8 +144,8 @@ deps =
 allowlist_externals =
     rm
 commands =
-    rm -fr nilearn/plotting/tests/baseline/*.png
-    pytest nilearn/plotting/tests/test_baseline_comparisons.py -n auto --mpl --mpl-generate-path=nilearn/plotting/tests/baseline {posargs:}
+    ; rm -fr nilearn/plotting/tests/baseline/*.png
```
Remi-Gau: Revert before merging

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -378,6 +378,19 @@ def _get_displays(self):
         self._report_content["summary"] = table
         self._report_content["number_of_regions"] = self.n_elements_
 
+        return self._create_figure_for_report(labels_image)
+
+    def _create_figure_for_report(self, labels_image):
```
Remi-Gau, line 402: Extract in report mixin?

nilearn/reporting/tests/test_baseline_comparisons.py
```diff
@@ -127,3 +143,206 @@ def test_stat_map_to_png_surface(
...
+)
+@pytest.mark.parametrize("src_masker", [SurfaceMasker, MultiSurfaceMasker])
+def test_surface_masker_create_figure_for_report(src_masker, mask_img, img):
+    """Check figure generated in report of (Multi)SurfaceMasker."""
+    masker = src_masker(mask_img)
+    masker.fit(img)
+    return masker._create_figure_for_report()
+
+
+# TODO: add later as there seem to be some flaky tests failures
```
Remi-Gau, line 291: Tackle in a follow up pr

Remi-Gau:
@hndgzkn
I have seen a coue of things I need to change but if you 5 minutes I would take any comments on this.

nilearn/maskers/_mixin.py
```diff
@@ -462,3 +462,8 @@ def _reporting(self):
     @abc.abstractmethod
     def _get_displays(self):
         raise NotImplementedError()
+
+    @abc.abstractmethod
+    def _create_figure_for_report(self):
+        """Generate figure for report."""
+        raise NotImplementedError()
```
hndgzkn, line 469: I think if we are going to put the interface for this function, it should be in _ReportingMixin. 

However I am not sure if we should ask for an implementation. `_create_figure_for_report` is not called from outside of masker classes, not even from outside of `_get_displays`. `_get_displays` should be implemented, as it is called from html_report, but then we can leave the implementation of `get_displays` free and delete this definition from mixin class. 

Or if we want to force a certain code structure, that might be reasonable. Then in that case that would be good to document the reasoning and usage.

hndgzkn (approved):
LGTM a part from one comment.

nilearn/maskers/_mixin.py
Remi-Gau: let me double check

big picture: for this PR I needed to make sure that each masker has a function that returns a figure

Remi-Gau: actually it is used at least once outside of _get_display:

```
$ grep -rni  'masker_._create_figure' nilearn
nilearn/reporting/glm_reporter.py:539:        model.masker_._create_figure_for_report()
```

this is used when plotting the mask used in the GLM report

hndgzkn: > let me double check
> 
> big picture: for this PR I needed to make sure that each masker has a function that returns a figure

Isn't this `get_display` function?

Remi-Gau: this will also be used in the decoding reporter

hndgzkn: > actually it is used at least once outside of _get_display:
> 
> ```
> $ grep -rni  'masker_._create_figure' nilearn
> nilearn/reporting/glm_reporter.py:539:        model.masker_._create_figure_for_report()
> ```
> 
> this is used when plotting the mask used in the GLM report

Ok, I had just looked at the changes in the PR. 

However, should `glm_reporter` use `_get_display` or `_create_figure_for_report`? If the second, then that might be good to document the difference.

Remi-Gau: > > let me double check
> > big picture: for this PR I needed to make sure that each masker has a function that returns a figure
> 
> Isn't this `get_display` function?

several of the get_display do way more than just returning a figure or list of figures: makes me reconsider the renaming we recently did

hndgzkn: Ok, I've looked at the code of both. `_get_displays` in fact populates the report content and then also calls `create_figure_for_report` and returns. So it is not necessary for glm report I guess. Then it should stay.

hndgzkn: > makes me reconsider the renaming we recently did

I agree

hndgzkn: I would work on generalizing _ReportingMixin to other relevant estimators as we discussed [here](https://github.com/nilearn/nilearn/pull/5790#issuecomment-3469317687) if that is ok with you.

Remi-Gau: OK: suggestion.

- let's only flag potential issues and refactoring possiblities for now but not act upon them
- let's try to get through this PR and https://github.com/nilearn/nilearn/pull/5757 and https://github.com/nilearn/nilearn/pull/5809
- let's reassess and potentially refactor before we move on to expanding reports to other things like in https://github.com/nilearn/nilearn/pull/5815

I am afraid of us getting too often side tracked in PR by adding refactoring into them

Remi-Gau: > I would work on generalizing _ReportingMixin to other relevant estimators as we discussed [here](https://github.com/nilearn/nilearn/pull/5790#issuecomment-3469317687) if that is ok with you.

good with me but I think it will be easier to do once we have the 3 PR above merged first to avoid working on the same things

PR #5745, opened 2025-10-08
[MAINT] Avoid assigning result of a function...

DimitriPapadopoulos:
...where the function has no return.

https://github.com/nilearn/nilearn/blob/5083999d5b005050dd69f228823a4c1f6bd43dd3/nilearn/maskers/base_masker.py#L779


Remi-Gau:
failures are unrelated

DimitriPapadopoulos:
It ends up being an empty commit, after pre-commit reverted my change. It still feels wrong to assign the result of a function that doesn't return. Perhaps I should remove the assignment instead of explicitly returning `None`.

PR #5751, opened 2025-10-09
[ENH] add support for surface data to get_clusters_table

Remi-Gau:
- Closes https://github.com/nilearn/nilearn/issues/4326

- add support for generating cluster tables for surface data
- add cluster tables to report when GLM is run on surface data

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,7 +63,7 @@ repos:
     # Check formatting of CSS
     # prettier: https://prettier.io/
 -   repo: https://github.com/rbubley/mirrors-prettier
-    rev: v3.6.2
+    rev: v3.7.3
     hooks:
     -   id: prettier
         types_or: [css]
@@ -82,7 +82,7 @@ repos:
 
     # Lint and format Python code
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.6
+    rev: v0.14.7
     hooks:
     -   id: ruff-check
         # args: [--statistics]
@@ -105,7 +105,7 @@ repos:
         args: [--config=pyproject.toml, nilearn]
 
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.18.2
+    rev: v1.19.0
     hooks:
     -   id: mypy
         files: nilearn


nilearn/reporting/tests/test_get_clusters_table.py
```diff
@@ -100,26 +128,212 @@ def test_cluster_nearest_neighbor(shape):
...
+    #     math_img("img*-1", img=stat_img),
+    #     stat_threshold=stat_threshold,
+    #     cluster_threshold=cluster_threshold,
+    #     two_sided=False,
+    # )
+
+    # assert len(clusters_table_two_sided) == (
+    # len(clusters_table_positive) + len(
+    #     clusters_table_negative
+    # ))
```
Remi-Gau: not 100% this should always be the case

nilearn/reporting/get_clusters_table.py
```diff
@@ -424,7 +576,7 @@ def get_clusters_table(
             for subpeak in range(n_subpeaks):
                 if subpeak == 0:
                     row = [
-                        c_id + 1,
+                        c_id + offset + 1,
```
Remi-Gau, line 594: adding an offset to make sure clusters have all different  id

bthirion:
Do you need a review on that one ?

Remi-Gau:
something is not right: asking for two sided tables and only the cluster on the right is reported

image


image

Remi-Gau:
> something is not right: asking for two sided tables and only the cluster on the right is reported

should be fixed now

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -169,26 +173,33 @@
...
+# We also extract and print a table of clusters
+# that survive our threshold.
+
 from nilearn.datasets import load_fsaverage_data
 from nilearn.plotting import plot_surf_stat_map, show
+from nilearn.reporting import get_clusters_table
 
-#  let's make sure we use the same threshold
 threshold = 3.0
+cluster_trhreshold = 20
```
bthirion:
```suggestion
cluster_threshold = 20
```

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -198,10 +209,19 @@
         bg_map=fsaverage_data,
     )
 
+    table = get_clusters_table(
+        z_score,
+        stat_threshold=threshold,
+        cluster_threshold=cluster_trhreshold,
```
bthirion:
```suggestion
        cluster_threshold=cluster_threshold,
```

bthirion (commented):
M with only a small glitch. Thx !

examples/04_glm_first_level/plot_localizer_surface_analysis.py
```diff
@@ -217,15 +237,20 @@
     height_control=None,
     prefix="sub-01",
     out_dir=output_dir,
-    cluster_threshold=10,
+    cluster_threshold=cluster_trhreshold,
+    two_sided=True,
```
hndgzkn:
```suggestion
    cluster_threshold=cluster_threshold,
    two_sided=True,
```

examples/04_glm_first_level/plot_predictions_residuals.py
```diff
@@ -77,23 +77,31 @@
 # %%
 # Extract the largest clusters
 # ----------------------------
+# We can extract the 6 largest clusters surviving our threshold.
+# and get the x, y, and z coordinates of therir peaks.
+# We thennextract the time series from a sphere around each coordinate.
```
hndgzkn:
```suggestion
# and get the x, y, and z coordinates of their peaks.
# We then extract the time series from a sphere around each coordinate.
```

bthirion (approved):
LGTM.

nilearn/reporting/tests/test_get_clusters_table.py
```diff
@@ -138,110 +368,134 @@ def test_get_clusters_table_negative_threshold(shape, affine_eye):
...
     assert_array_equal(stat_img.get_fdata(), data_orig)
 
 
-@pytest.mark.slow
-def test_get_clusters_table_more(shape, affine_eye, tmp_path):
-    """Run more tests get_clusters_table.
+def test_get_clusters_table_negative_threshold_one_sided(
+    simple_stat_img, surf_img_1d
+):
+    """Check that one sided negative thresholds errs when two_sided=True."""
```
hndgzkn:
```suggestion
    """Check one sided negative thresholds errors when two_sided=True."""
```

nilearn/reporting/tests/test_get_clusters_table.py
```diff
@@ -100,26 +137,219 @@ def test_cluster_nearest_neighbor(shape):
...
+        (1.4, 10, 4),
+    ],
+)
+def test_get_clusters_table_surface_real_data(
+    stat_threshold, cluster_threshold, expected_n_cluster_two_sided
+):
+    """Test cluster table generation on real surface data."""
+    stat_img = load_fsaverage_data(mesh_type="inflated")
+
+    check_sum_negative_positive_clusters(
```
hndgzkn: Is this function used elsewhere. If not, the contents could be moved here.

nilearn/reporting/get_clusters_table.py
```diff
@@ -296,41 +303,198 @@ def get_clusters_table(
...
+
+            all_clusters.append(clusters)
+
+            data[hemi] = labels
+
+            offset += len(clusters)
+
+        if offset == 1:
+            warnings.warn(
+                f"No clusters found for than '{stat_threshold=}'.",
```
hndgzkn:
```suggestion
                f"No clusters found for '{stat_threshold=}'.",
```

nilearn/reporting/get_clusters_table.py
```diff
@@ -296,41 +303,198 @@ def get_clusters_table(
...
+
+    else:
+        signs = [1, -1]
+        for sign in signs:
+            temp_stat_map = threshold_img(
+                img=stat_img,
+                threshold=stat_threshold * sign,
+                cluster_threshold=cluster_threshold,
+                two_sided=False,
+            )
```
hndgzkn, line 450: If I do not miss something, thresholding is only done for `two_sided`, but not `one_sided`.

I would suggest to move thresholding under `if not two_sided` block.

Remi-Gau: actually when doing one sided the thresholding has already happened in `get_clusters_table` before the branching between volume and surface data

nilearn/reporting/tests/test_get_clusters_table.py
Remi-Gau: let me check: I think I had in mind to use this check more systematically but could in the end only use it once

if that's the case I will inline it

nilearn/reporting/get_clusters_table.py
hndgzkn: Ok, I see, but doesn't it cover `two_sided=True` as well in that case?

Remi-Gau: whether you do one or 2 sided, you threshold the image first
when one sided you just get the clusters
when 2 sided you then threshold twice (in a one sided manner, one for each side) and get the cluster for each side

you cannot 'just' call 'get_clusters_table' one-sided twice for the two-sided approach because we need to keep track of the ID of the cluster via the offset variable

nilearn/reporting/glm_reporter.py
Remi-Gau, line 1: some changes to apply to GLM reporting refactoring

nilearn/reporting/get_clusters_table.py
hndgzkn: Ok, thanks for the explanation. Not very clear to me now, I need to have deeper look at it.

PR #5757, opened 2025-10-13
[FIX] allow report generation even when plotting engine is missing

### LINKED ISSUE #5717: do not limit import from reporting when plotting dependencies are missing

Remi-Gau:
- Closes #5717
- Relates to #5828 

- [x] check that maskers generate valid html report when matplotlib or matplotlib is not installed
- [x] ensure `make_glm_report` is accessible when matplotlib is not installed
- [x] ensure that maskers and glm reports throw the same warning when generate report with not matplolib (also ensure the warning is included in the report)
- [x] changelog
- [x] ensure that some exaample reports embedded in the doc are generated on every PR
- [x] convert `doc/visual_testing/visualize_reports_in_notebook.ipynb` to a marimo notebook
- [x] extract common doc strings for `displayed_maps` for maps and sphere maskers
- [x] extract common validation for `displayed_maps` and `displayed_spheres`  for maps and sphere maskers
- [x] merge tests for errors and warnings for `displayed_maps` and `displayed_spheres`  for maps and sphere maskers

### DIFF
--- a/nilearn/plotting/tests/test_img_comparisons.py
+++ b/nilearn/plotting/tests/test_img_comparisons.py
@@ -229,6 +229,7 @@ def test_plot_bland_altman_surface(matplotlib_pyplot, surf_img_1d, masker):
     )
 
 
+@pytest.mark.slow
 def test_plot_bland_altman_errors(
     surf_img_1d, surf_mask_1d, img_3d_rand_eye, img_3d_ones_eye
 ):
--- a/nilearn/reporting/tests/test_get_clusters_table.py
+++ b/nilearn/reporting/tests/test_get_clusters_table.py
@@ -146,6 +146,7 @@ def test_get_clusters_table_negative_threshold(shape, affine_eye):
     assert_array_equal(stat_img.get_fdata(), data_orig)
 
 
+@pytest.mark.slow
 def test_get_clusters_table_more(shape, affine_eye, tmp_path):
     """Run more tests get_clusters_table.
 


Remi-Gau:
- ok will be easier to finish once https://github.com/nilearn/nilearn/pull/5744 has been merged

doc/visual_testing/visualize_reports_in_notebook.ipynb
Remi-Gau, line 1: convert jupyter notebook to marimo notebook

doc/visual_testing/reporter_visual_inspection_suite.py
Remi-Gau, line 1: modified this script so that even on partial doc build it will generate unfitted reports: should be even easier to check some reports of basic PR

doc/visual_testing/reporter_visual_inspection_suite.py
```diff
@@ -376,298 +367,166 @@ def report_surface_flm(build_type):
...
     )
 
-    masker.reports = True
-    masker.fit(data.func[0])
-    print(masker._reporting_data)
-    report = masker.generate_report(
-        title="Nifti Maps Masker - Fitted with Development FMRI",
-        displayed_maps=[2, 6, 7],
-    )
-    report.save_as_html(REPORTS_DIR / "nifti_maps_masker.html")
```
Remi-Gau, line 439: @hndgzkn I have sort of removed this check with report=False from this script (i.e fitting with reports=False, setting it to True and then asking for a report)

- I think it should have a test for it instead in https://github.com/nilearn/nilearn/pull/5836

.github/workflows/test_with_tox.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```

revert before merging

doc/visual_testing/reporter_visual_inspection_suite.py
hndgzkn: I've added the test to https://github.com/nilearn/nilearn/pull/5836.

Remi-Gau: awesome thanks

doc/modules/generated_reports/masker_reports_examples.rst
Remi-Gau, line 1: show unfitted and fitted maskers for all maskers

do not show multi maskers as it would get VERY redundant

nilearn/conftest.py
```diff
@@ -58,9 +56,6 @@
         [
             "_utils/plotting.py",
             "plotting",
-            "reporting/html_report.py",
-            "reporting/tests/test_html_report.py",
-            "reporting/tests/test_glm_report.py",
```
Remi-Gau, line 63: There is now a whole bunch of tests that can be run even when there is no plotting engine.

Tests that still must be skipped are done at the test level.

nilearn/conftest.py
```diff
@@ -25,9 +25,7 @@
     SurfaceImage,
 )
 
-collect_ignore = ["datasets/data/convert_templates.py"]
-collect_ignore_glob = ["reporting/_visual_testing/*"]
```
Remi-Gau, line 29: should have been removed a long time ago

nilearn/maskers/nifti_maps_masker.py
Remi-Gau, line 1: adds a series of checks that were present for the sphere masker but missing here

Remi-Gau: some of those checks are now common to nifti sphere masker, nifti maps masker and surface maps masker: could consider refactoring

nilearn/reporting/html_report.py
```diff
@@ -97,101 +104,116 @@ def __str__(self):
...
-        loader=FileSystemLoader(TEMPLATE_ROOT_PATH),
-        autoescape=select_autoescape(),
-        lstrip_blocks=True,
-        trim_blocks=True,
-    )
-
+def generate_report(estimator) -> list[None] | HTMLReport:
+    """Generate a report for Nilearn objects.
 
-def embed_img(display):
```
Remi-Gau, line 110: simply moved

nilearn/reporting/html_report.py
```diff
@@ -248,98 +283,26 @@ def assemble_report(body: str, title: str) -> HTMLReport:
     )
 
 
-def _define_overlay(estimator):
```
Remi-Gau: moved

nilearn/reporting/html_report.py
```diff
@@ -97,101 +104,116 @@ def __str__(self):
...
-        return display
-    return figure_to_svg_base64(display.frame_axes.figure)
+    data = {}
+    if hasattr(estimator, "_report_content"):
+        data = estimator._report_content
 
+    # Generate a unique ID for this report
+    data["unique_id"] = str(uuid.uuid4()).replace("-", "")
 
-def _update_template(
```
Remi-Gau, line 132: inlined into _create_report

nilearn/maskers/nifti_masker.py
```diff
@@ -422,6 +423,9 @@ def _create_figure_for_report(self):
         -------
         List of :class:`~nilearn.plotting.displays.OrthoSlicer`
         """
+        if not is_matplotlib_installed():
+            return [None]
+
```
hndgzkn: The same here.

nilearn/reporting/html_report.py
```diff
@@ -428,3 +338,23 @@ def is_notebook() -> bool:
...
+    if len(displays) == 1:  # set overlay to None
+        return None, displays[0]
+
+    elif isinstance(estimator, NiftiSpheresMasker):
+        return None, displays
+
+    elif len(displays) == 2:
+        return displays[0], displays[1]
+
+    return None, displays
```
hndgzkn: Do we really need this function?

If `_reporting ` function of each masker returns the expected output, there will be no need for `define_overlay` and `masker._reporting()` could be called instead of `_define_overlay` and this would remove dependency on `NiftiSpheresMasker`.

nilearn/reporting/html_report.py
```diff
@@ -97,101 +108,118 @@ def __str__(self):
...
-    """Set up the jinja Environment."""
-    return Environment(
-        loader=FileSystemLoader(TEMPLATE_ROOT_PATH),
-        autoescape=select_autoescape(),
-        lstrip_blocks=True,
-        trim_blocks=True,
-    )
-
+def generate_report(estimator) -> list[None] | HTMLReport:
+    """Generate a report for Nilearn objects.
```
hndgzkn: This function is not in public API, and only used by maskers. 

In that case, deleting this function and moving the contents of this function to `nilearn.maskers._mixin._ReportingMixin.generate_report` might be an option. And finally, when dependency issues are solved, `nilearn.maskers._mixin._ReportingMixin` could be moved to `nilearn.reporting` to be used by all estimators that want to implement reporting.

nilearn/reporting/tests/test_html_report.py
```diff
@@ -416,19 +378,6 @@ def test_overlaid_report(img_fmri):
     assert '<div class="overlay">' in str(html)
 
 
-def test_multi_nifti_masker_generate_report_imgs(img_fmri):
```
Remi-Gau, line 439: covered in estimator_checks

nilearn/reporting/tests/test_html_report.py
Remi-Gau, line 1: merged tests for errors and warnings for displayed_maps and displayed_spheres for maps and sphere maskers

nilearn/maskers/nifti_maps_masker.py
```diff
@@ -391,7 +340,8 @@ def _create_figure_for_report(self):
         if img is None:
             for component in maps_to_be_displayed:
                 display = plot_stat_map(
-                    index_img(maps_image, component), cmap=cm.black_blue
+                    index_img(maps_image, component),
+                    cmap=cm.black_blue,
```
Remi-Gau:
```suggestion
                    cmap=cm.black_blue
```

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -433,7 +428,7 @@ def _create_figure_for_report(self, labels_image):
         # Otherwise, simply plot the ROI of the label image
         # and give a warning to the user
         else:
-            display = plot_roi(labels_image)
+            display = plot_roi(labels_image, cut_coords=cut_coords)
```
Remi-Gau:
```suggestion
            display = plot_roi(labels_image)
```

nilearn/reporting/html_report.py
```diff
@@ -248,27 +162,7 @@ def assemble_report(body: str, title: str) -> HTMLReport:
...
-    if len(displays) == 1:  # set overlay to None
-        return None, displays[0]
-
-    elif isinstance(estimator, NiftiSpheresMasker):
-        return None, displays
-
-    elif len(displays) == 2:
-        return displays[0], displays[1]
-
-    return None, displays
```
Remi-Gau, line 268: removed in favor in making output of `_reporting` more consistent

Remi-Gau:
putting here all the reports generated with matplotlib missing

[reports_with_no_engine.zip](https://github.com/user-attachments/files/23551543/reports_with_no_engine.zip)

Remi-Gau:
@hndgzkn @bthirion 
this is a huge PR: let me know if you prefer that I split it into smaller ones (though this may prove tricky)

bthirion:
No, it's OK. We'll suffer a bit, but you did most of it !

doc/modules/generated_reports/masker_reports_examples.rst
bthirion: Sorry, where can we see the result of that ?

doc/visual_testing/reports_in_notebook.py
```diff
@@ -0,0 +1,159 @@
+"""Marimo notebook to visualize reports in a notebook."""
```
bthirion, line 1: How do we see the generated NB ?

nilearn/_utils/tags.py
```diff
@@ -79,7 +79,9 @@ class InputTags(SkInputTags):
         surf_img: bool = False
 
 
-def get_tag(estimator: BaseEstimator, tag: str) -> bool:
+def get_tag(estimator: Any, tag: str) -> bool:
```
bthirion, line 82: What is the motivation for this change ? `BaseEstimator` is not general enough ?

doc/visual_testing/reporter_visual_inspection_suite.py
bthirion: Sorry (again), where can one see the result ?

nilearn/_utils/tags.py
Remi-Gau: yeah I was getting an error when passing something that is not an estimator at all to this function.

our homemade function `is_masker` is used at some point in the GLM to see if the mask passed to the GLM object is a mask image or a masker: this would throw an error about because a mask image has no `__sklearn_tags__` method

so I had to modify the get_tag, is_masker and a few other functions to be more general and therefore be able to check any object: hence the more generic type annotation

Remi-Gau:
> The PR LGTM. Just I struggle to see the visual checks etc.

so by default with this PR, we will now always get the reports of unfitted estimators in the doc:
See
- https://output.circle-artifacts.com/output/job/43b5ba2b-811a-4b45-b203-8f9ef01fcea1/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html
- https://output.circle-artifacts.com/output/job/43b5ba2b-811a-4b45-b203-8f9ef01fcea1/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html

I will as for a build of the reports

doc/visual_testing/reports_in_notebook.py
Remi-Gau: at the moment you would have to run it locally to see that

note this is the same to what we currently have on main: https://github.com/nilearn/nilearn/blob/main/doc/visual_testing/visualize_reports_in_notebook.ipynb

What we could do as part of the CI is to ask for this notebook to be run in CI and we upload its 'html' output as an artifact of the CI so we can inspect it if we need.

Remi-Gau:
one thing that we cannot easily see in CI is how fiited notebook would look when there are no plotting engines but to give you an idea:

image

nilearn/reporting/html_report.py
Remi-Gau: I have removed the function.

nilearn/reporting/html_report.py
Remi-Gau: Am afraid that if we do that we are going to have move of the code of reporting into maskers, so I'd rather move the ReportingMixin out of maskers and then give it the generate_report method.

Though in any case I'd like to wait to see if we can use a single ReportingMixin for maskers AND GLM

doc/visual_testing/reporter_visual_inspection_suite.py
```diff
@@ -376,298 +368,169 @@ def report_surface_flm(build_type):
...
+    _generate_dummy_html(filenames=[f"{masker_class_name}_fitted.html"])
 
-    return report
+    return unfitted_report, None
```
hndgzkn: Is returning `None` necessary here?

It might be handled in the calling function. I see that this facilitates, but seems a bit strange to me that it always returns None (though I don't have a strong opinion).

nilearn/maskers/nifti_spheres_masker.py
```diff
@@ -358,172 +361,137 @@ def __init__(
...
-        else:
-            positions = [
-                np.round(
-                    coord_transform(*seed, np.linalg.inv(img.affine))
-                ).astype(int)
-                for seed in seeds
+        if not self._has_report_data():
+            return [None]
+
+        if self._has_report_data():
```
hndgzkn:
```suggestion
        else:
```

nilearn/maskers/surface_labels_masker.py
```diff
@@ -468,17 +473,23 @@ def inverse_transform(self, signals):
...
         -------
         displays : list
             A list of all displays to be rendered.
         """
+        # Handle the edge case where this function is called
+        # without matplolib or
+        # with a masker having report capabilities disabled
+        if not is_matplotlib_installed() or not self._has_report_data():
+            return [None]
+
```
hndgzkn, line 489: Could this be done for all masker `reporting` functions (in nifti maskers) and remove `not is_matplotlib_installed()` from `create_figure_for_report` function?

nilearn/reporting/tests/test_html_report.py
```diff
@@ -21,10 +20,9 @@
     SurfaceMapsMasker,
     SurfaceMasker,
```
hndgzkn, line 21: I think most of the tests in this module belongs to `nilearn.maskers` package. Actually this module does not import any functions from reporting and creates unnecessary dependency from `nilearn.reporting` to `nilearn.maskers` package.

nilearn/reporting/glm_reporter.py
```diff
@@ -369,6 +357,11 @@ def make_glm_report(
         contrasts_dict = output["contrasts_dict"]
 
     if is_matplotlib_installed():
+        from nilearn._utils.plotting import (
+            generate_contrast_matrices_figures,
+            generate_design_matrices_figures,
```
hndgzkn, line 362: Could these functions be moved to `nilearn.plotting`, for example `nilearn.plotting._reporting` or  `nilearn.plotting._glm` package? Seems to be used only for glm. Not in this PR, but could be discussed later.

nilearn/maskers/nifti_spheres_masker.py
Remi-Gau: actually given the previous branch has a return we can even get away without the else

doc/visual_testing/reporter_visual_inspection_suite.py
Remi-Gau: yeah good point 
fixing it

nilearn/maskers/surface_labels_masker.py
Remi-Gau: actually I would prefer the create_figure_for_report functions to not fail when called with no plotting engine is installed so the 

```
        if not is_matplotlib_installed():
            return [None]
```

should probably be found in `create_figure_for_report`

I agree that some refactoring can probably be still be done, but i'd rather do it in a follow up PR if that's OK with you

nilearn/reporting/glm_reporter.py
Remi-Gau: I agree that we need to find a better home for those in a follow up PR

nilearn/reporting/tests/test_html_report.py
Remi-Gau: That's a good idea.
I think the same logic should be applied to GLM reporting tests.

Unless we can have some reporting tests that can apply to a very wide range of estimators (in which case I think they should go to `nilearn/_utils/estimator_checks`) then we should try to keep those under the subpackage where those estimators are defined.

Will move things in a follow up PR.

Remi-Gau:
reports can be inspected here:

https://output.circle-artifacts.com/output/job/f452d8d8-326e-42b3-a0a3-28e6889b0b0e/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html

https://output.circle-artifacts.com/output/job/f452d8d8-326e-42b3-a0a3-28e6889b0b0e/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html

Remi-Gau:
something seems off with the maskers reports... checking to see if the issue also exists locally

Remi-Gau:
OK reports look fine now:

https://output.circle-artifacts.com/output/job/7822992b-c9a1-4b15-acca-6db78fd8297f/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html

https://output.circle-artifacts.com/output/job/7822992b-c9a1-4b15-acca-6db78fd8297f/artifacts/0/dev/modules/generated_reports/glm_reports_examples.html

Remi-Gau:
OK errors in the doc are unrelated: merging

PR #5759, opened 2025-10-13
[REF] rely on minimum dependencies mentioned in pyproject.toml

Remi-Gau:
- Closes none

- refactor tox.ini to rely on minimum dependencies mentioned in pyproject.toml

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -15,7 +15,8 @@ HIGHLIGHTS
  |
  | **Minimum supported versions of the following packages have been bumped up:**
  | - matplotlib -- 3.8.0
- | - SciPy -- v1.9.0
+ | - scipy -- 1.9.0
+ | - requests -- 2.30.0
 
 NEW
 ---


tox.ini
```diff
@@ -125,9 +114,9 @@ commands =
 [testenv:min]
 description = Run tests on minimum version of all dependencies (plotting not included).
 passenv = {[global_var]passenv}
-extras = test
-deps =
-    {[min]deps}
+extras =
+    min
```
bthirion, line 110: IIUC the min versions are  defined implicitly ?

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "nibabel>=5.2.0",
     "numpy>=1.22.4",
     "pandas>=2.2.0",
-    "requests>=2.25.0",
```
Remi-Gau, line 29: seems that the sphinx version we use for our doc build requires a more recent version of requests

bthirion (approved):
LGTM.

tox.ini
Remi-Gau: well we use to define them explicitly in the tox.ini but in this PR this is moved to the pyproject.toml, so in case anyone wants to check something with the oldest dependencies then we can just do:

```
pip install -e '.[min]'
```

PR #5727, opened 2025-10-03
[ENH] add multi surface maps masker

Remi-Gau:
- Relates to #5539

Changes proposed in this pull request:

- [x] add multi surface maps masker 
- [x] changelog

PR #5763, opened 2025-10-14
[FIX] init report multi surface maps masker

Remi-Gau:
- fixes https://github.com/nilearn/nilearn/pull/5727#issuecomment-3400319311

- init report for multi surface maps masker


Remi-Gau:
ok merging this one because without it we have mass failure everywhere in CI

Issue #5712, opened 2025-09-24
speeding up the test suite


Remi-Gau:
once again the test suite is getting too slow

see figure below for macos and linux (windows is not there because it's even worse)

<img width="1000" height="700" alt="Image" src="https://github.com/user-attachments/assets/a9682a60-0e21-4a22-94ca-b61d568d8dea" />

discussing with @man-shu some options

- speed up slow tests: I think we can gain a little on here, but not enough that it will be a viable way to do it
- pay to get some "proper" github action time to use more workers
- change testing strategy: 
  - mark test as slow / fast, run fast test first and only run slow test if fast test fail (con: the fail fast may hide some other tests failure that will only 'appear' once the fast tests are fixed)
- reduce how many tests we run per PR:
  - reduce the testing matrix:
    - run oldest supported python on all OS, run latest supported python on a single OS
    - only run the "no plotting" tests on main 
  - only run tests that are affected by a PR:
    - several pytest extensions 'claim' to be able to help with that
    - rely on our the architecture we are starting to enforce with import-linter to sub-select the tests to run, for example: 
      - nilearn.signal only import from nilearn.typing and nilearn.exceptions, so if nilearn.signal is modified we only need to modify the tests for nilearn.signal
      - nilearn.glm does not import from nilearn.decoding, so no need to run the tests for decoding when working on GLM.

Remi-Gau:
probably something to discuss at the next core dev meeting

bthirion:
But shouldn't we first diagnose the heavier tests ? We can probably make them lighter ?
I'm just a bit worried about the complexity of the strategies you outline above.

Remi-Gau:
> But shouldn't we first diagnose the heavier tests ? We can probably make them lighter ?

yeah there are things in place for this:
- look for the tests with `@pytest.mark.timeout(0)` or we can look in each test run the reported test timing (there is a step that uploads a CSV and some graphs of them all)

always worth trying but it won't get us far enough I think

some of the other options are not THAT difficult (like removing some of the combination in the OS X python version testing matrix

bthirion:
> some of the other options are not THAT difficult (like removing some of the combination in the OS X python version testing matrix

True

Remi-Gau:
closing for now

PR #5765, opened 2025-10-14
[MAINT] simplify tests matrix

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/issues/5712

- testenv:plot_full_min does the same as testenv:pytest_mpl so let's just use the later
- plot_min now runs 'all' tests with oldest dependencies: meaning those with matplotlib and plotly and not just matplotlib


Remi-Gau:
failure is unrelated so merging

PR #5773, opened 2025-10-17
[ENH] error surface dimension

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/issues/5539

- throw dimension error when surface image passed to maskers do not have the right type of dimension.


nilearn/maskers/base_masker.py
```diff
@@ -553,6 +553,7 @@ def _check_imgs(self, imgs) -> None:
                 "an iterable of SurfaceImage."
                 f"Got: {imgs.__class__.__name__}"
             )
+        check_surf_img(imgs)
```
Remi-Gau, line 613: make sure image is not empty

nilearn/maskers/base_masker.py
```diff
@@ -638,7 +639,6 @@ def transform(self, imgs, confounds=None, sample_mask=None):
         if not isinstance(imgs, list):
             imgs = [imgs]
         imgs = concat_imgs(imgs)
-        check_surf_img(imgs)
```
Remi-Gau, line 686: handled by _check_imgs above

nilearn/maskers/surface_labels_masker.py
```diff
@@ -222,18 +221,9 @@ def fit(self, imgs=None, y=None):
...
 
-        if imgs is not None:
-            check_surf_img(imgs)
-
-            if isinstance(imgs, SurfaceImage) and any(
-                hemi.ndim > 2 for hemi in imgs.data.parts.values()
-            ):
-                raise ValueError(
-                    "should only be SurfaceImage should 1D or 2D."
-                )
```
Remi-Gau, line 249: removing this in all maskers as we do not handled 3D surface image (with several 'layers') for now: added a check and a test for this in surface image creation

nilearn/surface/surface.py
```diff
@@ -1419,19 +1430,23 @@ def _get_min_max(self):
...
-                max_n_samples.append(1)
-        max_n_samples = np.max(max_n_samples)
-        if max_n_samples > samples:
-            raise ValueError(
-                f"Data for each part of {var_name} should be {samples}D. "
-                f"Found: {max_n_samples}."
-            )
+    def _check_n_samples(self, samples: int):
+        """Check that image has no more than n samples."""
+        # Ensure both parts have same number of samples
```
bthirion, line 1451: Do we agree that `samples` here are  vertices ?
I don't think that the two parts should have the same number of vertices in general (this is not the case of individual meshes with freesurfer).

bthirion (commented):
Note the codecov complaint.

nilearn/surface/surface.py
Remi-Gau: > Do we agree that `samples` here are vertices ?

nope, they are the number of 'timepoints' in a 2D surface image, so the equivalent of the 't' dimension on 4D volumes with x, y, z, t: for volumetric data, we may ask how many "volumes / timepoints" there are in a 4D image.

we still do not have a good official name for this for Surface internally in Nilearn (or in the doc or example for that matter): we mostly either call it samples or timepoints.

- sample is confusing because it has a clear meaning in stat and could also refer to vertices
- timepoints is inaccurate because for 2D surface maps, each 'timepoint' does not refer to time at all

> I don't think that the two parts should have the same number of vertices in general (this is not the case of individual meshes with freesurfer).

indeed we allow this

from the top of my head the 'rules' for surface images:
- meshes / data: can have different number of vertices for each hemisphere
- both hemisphere must have:
 - same datatype
 - same number of 'timepoints'

nilearn/surface/tests/test_surface.py
```diff
@@ -1295,3 +1301,41 @@ def test_find_surface_clusters_4_faces(mask, expected_n_clusters, surf_mesh):
...
+    data._check_n_samples(2)
+
+    if part.ndim > 1 and part.shape[1] > 1:
+        with pytest.raises(
+            DimensionError,
+            match="Expected 1 sample and you provided a 2 samples image",
+        ):
+            data._check_n_samples(1)
+    else:
+        data._check_n_samples(1)
```
Remi-Gau, line 1384: need to add the right test to make sure the proper error message is thrown when checking a list of surface images

nilearn/surface/surface.py
```diff
@@ -1364,9 +1365,34 @@ def __init__(self, left=None, right=None, dtype=None):
 
         self._check_parts()
 
+    @property
+    def shape(self):
```
Remi-Gau: was only moved

nilearn/surface/surface.py
```diff
@@ -1388,52 +1414,25 @@ def _check_parts(self):
...
-        return (
-            (sum_vertices, tmp.shape[1])
-            if len(tmp.shape) == 2
-            else (sum_vertices,)
-        )
-
-    def __repr__(self):
-        return f"<{self.__class__.__name__} {self.shape}>"
-
-    def _get_min_max(self):
```
Remi-Gau, line 1425: only moved

nilearn/surface/surface.py
bthirion: Since this 'samples' terminology is mostly internal, add a comment in the code so that people understand it easily.

bthirion:
A little bit more coverage and this should be good to go ?

nilearn/exceptions.py
```diff
@@ -48,18 +48,27 @@ class DimensionError(TypeError):
...
     ----------
-    file_dimension : integer
+    file_dimension : :obj:`int`
         Indicates the dimensionality of the bottom-level nifti file.
 
-    required_dimension : integer
+    required_dimension : :obj:`int`
         The dimension the nifti file should have.
 
+    msg_about_samples : :obj:`bool`
```
bthirion, line 57: aren't we overdesigning a little bit here ?

bthirion: Why not simply consider that this is True ?

bthirion (commented):
LGTM overall.

PR #5779, opened 2025-10-22
[FIX] issue 5732 - asymmetric labels in colorbar

### LINKED ISSUE #5732: assymetric labels in colorbar of some test figures
here we have a 4 on the positive side on the color bar but no -4 on the negative side: could this be fixed?

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5680#discussion_r2397887061_

hndgzkn:
- Closes #5732 

- Fix bug in `nilearn.plotting._utils.get_cbar_ticks` implementation
- Remove `0` from colorbar when `threshold` value is close to 0.


nilearn/plotting/_utils.py
```diff
@@ -65,13 +75,9 @@ def get_cbar_ticks(
...
-    The tick list will contain vmin, vmax and threshold values. If
-    necessary the number of ticks might increase by 2.
-
-    If the distance between threshold and the closest tick is smaller than or
-    equal to 1/3th of the tick spacing, or it is smaller than 4/3th of
-    threshold value the closest tick is replaced with threshold;
-    otherwise the threshold (and/or -threshold) is added to the tick list.
+    The final tick list will contain 0, vmin, vmax and threshold values. If
+    there are values very close to each other, it will favor threshold value
+    over 0.
```
hndgzkn, line 80: I've simplified the implementation. The previous implementation was checking the difference of tick values and threshold value trying to find where it can insert +-threshold. And then try to add 0, vmin, vmax values if they are replaced with threshold.

Now +-threshold and 0 are added to linearly spaced tick list between vmin and vmax, and then the difference of consecutive tick values are checked to see if they are very close. 
Very close means less then 1/3th of max distance between consecutive ticks. 

Then the tick which is not vmin, vmax, -threshold, threshold or 0 is removed. 

0 is deleted only if threshold and 0 are very close.
threshold is deleted only if threshold <= 1e-5

Remi-Gau (approved):
good with me

PR #5783, opened 2025-10-23
[MAINT] skip test intermediate python on PR and adapt schedule of workflows

Remi-Gau:
- Relates to #5712

- skip test intermediate python (3.11, 3.12) on PR
- do not run on main anymore:
  - benchmarks
  - linkchecks
  - nightly release of Nilearn dependencies
  - sourcery
  - test atlases (run on schedule instead)
  - testing minimum dependencies compatibilities
- run all scheduled workflows on mondays midnight: make sure that all of them (except check_gha_workflow, build_doc and benchmarks)

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -16,10 +16,9 @@ import re
 import sys
 from pathlib import Path
 
+from nilearn._version import __version__
 from sphinx.domains import changeset
 from sphinx.locale import _
-
-from nilearn._version import __version__
 
 # ----------------------------------------------------------------------------
 


bthirion:
Thx !  Do you have some measure of the impact on build time ?

Remi-Gau:
> Thx ! Do you have some measure of the impact on build time ?

well it was not working so I can't tell.

also we won't see the effect on a single PR because this mostly, reduces the number of jobs run per PR making things potentially faster overall when you have several PRs

.github/workflows/test_with_tox.yml
```diff
@@ -112,6 +111,11 @@ jobs:
             run: tox c
 
         -   name: Run test suite
+            # only run on main
+            # or PR for oldest and latest supported python versions
+            #
+            # NOTE: skipped versions will just show a green tick on PRs
```
Remi-Gau, line 123: let me know if this is confusing when seeing the PR results

I think I could adapt the PR matrix dynamically but this will make the workflow a bit more complicated

see https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#using-an-output-to-define-two-matrices

Remi-Gau: For example of something confusing, see the green tick and the very fast runtime on this pic:

image

hndgzkn: It seems Ok to me. But it might be good to update https://nilearn.github.io/stable/ci.html# with these changes.

Remi-Gau: OK that happens when you edit 'frontmatter' of the yaml file: will do that

.github/workflows/check_gha_workflow.yml
```diff
@@ -18,24 +18,12 @@ name: Benchmark test suite
...
-#
-#            ┌───────────── minute (0 - 59)
-#            │ ┌───────────── hour (0 - 23)
-#            │ │ ┌───────────── day of the month (1 - 31)
-#            │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
-#            │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
-#            │ │ │ │ │
-#            │ │ │ │ │
-#            │ │ │ │ │
-#            * * * * *
```
bthirion, line 34: This kind of schema is very useful for people like me. We could keep one where it makes sense.

Remi-Gau: OK I was not sure how useful it was:
will add it back in all places where we have a schedule

Remi-Gau:
OK a lot of things are gonna be run on schedule and that usually means that only a limited number of people (I think only the person who created or last modified the workflow file) will get notifications if the workflow fails.

So I will make sure those workflows open issues when they fail and notify the @nilearn/core

.sourcery.yaml
```diff
@@ -28,6 +28,7 @@ rule_settings:
     -   no-loop-in-tests
     -   no-conditionals-in-tests
     -   low-code-quality
+    -   inline-immediately-returned-variable
```
Remi-Gau, line 31: we are getting a number of 'errors' (about 80) for this that are I think not so important for now
https://docs.sourcery.ai/References/Sourcery-Rules/Python/Default-Rules/inline-immediately-returned-variable/

.github/workflows/linkcheck.yml
```diff
@@ -44,4 +52,23 @@ jobs:
...
+            uses: JasonEtco/create-an-issue@v2.9.2
+            env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                run_id: ${{ github.run_id }}
+                repository: ${{ github.repository }}
+                workflow_name: LINKCHECK
+            with:
+                filename: .github/workflow_failure.md
+                update_existing: true
+                search_existing: open
```
Remi-Gau, line 70: OK ideally I would prefer to have a way to append comment to an already existing issue but that will do for now

it could help track the different failures for the tests on different OS, python / dependencies version

Remi-Gau: maybe with https://github.com/peter-evans/create-or-update-comment

PR #5787, opened 2025-10-23
[MAINT] run slow tests last

### LINKED ISSUE #5712: speeding up the test suite

Remi-Gau:
- Closes #5712

- [x] run slow tests after fast test to make sure sure test suite run fail faster
- [x] ensure that coverage report is properly merged

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -62,13 +62,15 @@ deps =
 commands =
 	pre-commit run --all-files --show-diff-on-failure {posargs:}
 
+
 [testenv:latest]
 description = Run tests on latest version of all dependencies (plotting not included).
 passenv = {[global_var]passenv}
 extras = test
 commands =
-    pytest -m "not slow" --timeout=45 --cov=nilearn --cov-report=xml --cov-report=html --report=report.html -n auto --csv results/pytest_output/pytest_output.csv  {posargs:}
-    pytest -m slow --cov=nilearn --cov-report=xml --cov-report=html --report=report_slow.html -n auto --csv results/pytest_output/pytest_output_slow_tests.csv  {posargs:}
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -216,10 +216,13 @@ module = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-ra --strict-config --strict-markers --doctest-modules --showlocals -s -v --template=maint_tools/templates/index.html --timeout=45"
+addopts = "-ra --strict-config --strict-markers --doctest-modules --showlocals -s -v --template=maint_tools/templates/index.html"
```
Remi-Gau, line 219: move cutoff of 45 seconds for fast and slow tests to tox.ini

PR #5788, opened 2025-10-23
[ENH] make `check_niimg`, `check_niimg_3d`, `check_niimg_4d` functions public

### LINKED ISSUE #5696: [BUG] Import regression from 0.12.0 to 0.12.1

Remi-Gau:
- Closes #5696

- [x] make `check_niimg`, `check_niimg_3d`, `check_niimg_4d` public
- [x] move all the content of `nilearn._utils.niimg_conversion` and `nilearn.image.utils` to `nilearn.image.image`
- [x] changelog
- [x] add deprecation warnings in _utils for smoother transition for downstream users
- [x] update test and type annotations for check_niimg*

### DIFF
--- a/nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
+++ b/nilearn/regions/tests/test_hierarchical_kmeans_clustering.py
@@ -115,6 +115,7 @@ def test_hierarchical_k_means_clustering_transform():
     assert X_red.shape == (n_samples, n_clusters)
 
 
+@pytest.mark.single_process
 def test_hierarchical_k_means_clustering_inverse_transform():
     n_samples = 15
     n_clusters = 8


nilearn/image/image.py
```diff
@@ -40,6 +32,14 @@
 )
 from nilearn._utils.path_finding import resolve_globbing
 from nilearn.exceptions import DimensionError
+from nilearn.image.utils import (
+    _index_img,
+    check_niimg,
+    check_niimg_3d,
+    check_niimg_4d,
```
Remi-Gau: I figure that some of those could actually be moved into nilearn.image.image to avoid some circular import issues

nilearn/_utils/tests/test_niimg.py
```diff
@@ -87,3 +94,172 @@ def test_load_niimg(img1, tmp_path):
     filename = write_imgs_to_path(img1, file_path=tmp_path, create_files=True)
     filename = Path(filename)
     load_niimg(filename)
+
+
+def test_repr_niimgs():
```
Remi-Gau, line 99: those tests were in the wrong file

bthirion (commented):
Oops, it's not a small one.
But LGTM

Remi-Gau:
> Oops, it's not a small one. But LGTM

yeah I know but it looks big because:
- some things were moved from one file into another one
- I cleaned up the tests (split into smaller ones, added docstrings)

Remi-Gau:
OK working on this PR and trying to figure out some type annotations and their corresponding test actually shows that there is one behavior that we did not test and leads to an error:

I tested this on this PR (and on main to make sure this is is not due to some weird refactoring):

Asking check_niimg to return an iterator when giving it a 3D image leads to an error.

```python
def test_check_niimg_img_3d_zeros_eye(img_3d_zeros_eye):
    img = check_niimg(img_3d_zeros_eye, return_iterator=True)
```

gives

```
IndexError: tuple index out of range
```

Note that things are fine when asking it to coerce it to 4D first with atleast_4d=True.

```python 
    img = check_niimg(img_3d_zeros_eye, atleast_4d=True, return_iterator=True)
    assert isinstance(img, Iterable)
    assert len(next(img).shape) == 3
```

@nilearn/core 

what would be your expectation of the behavior of?

```python
img = check_niimg(img_3d_zeros_eye, return_iterator=True)
```

- returns an iterator (AKA return_iterator overrides the default atleast_4d=False)  - ideally with a warning
- returns a 3D image (AKA the default atleast_4d=False overrides return_iterator) - ideally with a warning
- an error with a better message indicating of why this is not working

My initial hunch was that it should return a 3D image but I am not so sure anymore...

Given that this edge case has not appeared yet, I suspect it is not very common: but if were are making this public I would rather cover all cases.

nilearn/image/image.py
```diff
@@ -1892,3 +1897,514 @@ def copy_img(img):
...
+                    *exc.args,
+                )
+                raise
+
+    # Raising an error if input generator is empty.
+    if i == -1:
+        raise ValueError("Input niimgs list is empty.")
+
+
+# ensure_ndim = 3 always returns a NiftiImage
```
Remi-Gau, line 2101: update those type annotation once the behavior has been clarified

nilearn/image/__init__.py
```diff
@@ -22,7 +25,7 @@
     swap_img_hemispheres,
     threshold_img,
 )
```
hndgzkn, line 27: Could these be divided into two modules. For example the functions that manipulate the image could be moved to `image.utils`.

nilearn/image/image.py
```diff
@@ -5,32 +5,37 @@
...
-    check_niimg,
-    check_niimg_3d,
-    check_niimg_4d,
-    check_same_fov,
-    iter_check_niimg,
+from nilearn._utils.niimg import (
+    _get_data,
+    load_niimg,
+    repr_niimgs,
+    safe_get_data,
```
hndgzkn, line 37: Could these also be moved to `nilearn.image`?

nilearn/image/image.py
```diff
@@ -1892,3 +1897,514 @@ def copy_img(img):
...
+    elif isinstance(image, SurfaceImage):
+        data = get_surface_data(image)
+    elif isinstance(image, np.ndarray):
+        data = image
+    else:
+        raise TypeError(
+            "Image to extract indices from must be one of: "
+            "Niimg-Like, SurfaceImage, numpy array. "
+            f"Got {image.__class__.__name__}"
+        )
```
hndgzkn: I am not sure how frequent we need to do these checks. Would it be beneficial if we defined a single `get_image_data` function, that would return the correct data depending on the image type? That would avoid repeating these checks in the code.

nilearn/image/image.py
```diff
@@ -1892,3 +1897,514 @@ def copy_img(img):
...
+    """Return True if img's field of view correspond to given \
+    shape and affine, False elsewhere.
+    """
+    return img.shape[:3] == shape and np.allclose(img.affine, affine)
+
+
+def check_same_fov(*args, **kwargs) -> bool:
+    """Return True if provided images have the same field of view (shape and \
+    affine) and return False or raise an error elsewhere, depending on the \
+    `raise_error` argument.
```
hndgzkn, line 1917: `raise_error` argument does not exist.

nilearn/image/image.py
```diff
@@ -1892,3 +1897,514 @@ def copy_img(img):
...
+    args : images
+        Images to be checked. Images passed without keywords will be labeled
+        as img_#1 in the error message (replace 1 with the appropriate index).
+
+    kwargs : images
+        Images to be checked. In case of error, images will be reference by
+        their keyword name in the error message.
+
+    raise_error : boolean, optional
+        If True, an error will be raised in case of error.
```
hndgzkn: This argument does not exist in the function signature. It is expected as kwargs, may be could be mentioned there.

nilearn/image/image.py
```diff
@@ -1892,3 +1897,514 @@ def copy_img(img):
...
+
+    Parameters
+    ----------
+    args : images
+        Images to be checked. Images passed without keywords will be labeled
+        as img_#1 in the error message (replace 1 with the appropriate index).
+
+    kwargs : images
+        Images to be checked. In case of error, images will be reference by
+        their keyword name in the error message.
```
hndgzkn:
```suggestion
        Images to be checked. In case of error, images will be referenced by
        their keyword name in the error message.
```

nilearn/image/image.py
```diff
@@ -1892,3 +1897,514 @@ def copy_img(img):
...
+    dtype=None,
+    memory=None,
+    memory_level=0,
+):
+    """Iterate over a list of niimgs and do sanity checks and resampling.
+
+    Parameters
+    ----------
+    niimgs : list of niimg or glob pattern or itertools.tee instance
+        Image to iterate over.
```
hndgzkn:
```suggestion
        Images to iterate over.
```

nilearn/image/image.py
```diff
@@ -1892,3 +1897,514 @@ def copy_img(img):
...
+    ensure_ndim: Literal[4] = ...,
+    atleast_4d=...,
+    dtype=...,
+    return_iterator: Literal[True] = ...,
+    wildcards=...,
+) -> Iterable[Nifti1Image]: ...
+
+
+@fill_doc
+def check_niimg(
```
hndgzkn, line 2139: This function not only checks the image but also loads it. As an idea, would it be possible to rename it so that the name is more clear about its functionality (before making it public)?

Remi-Gau: I kind of agree.
For me usually 'check' will return None or raise an error and things.
How do other people feel about this? @nilearn/core

hndgzkn:
>     * returns an iterator (AKA return_iterator overrides the default atleast_4d=False)  - ideally with a warning
> 
>     * returns a 3D image (AKA the default atleast_4d=False overrides return_iterator) - ideally with a warning
> 
>     * an error with a better message indicating of why this is not working
> 
> 
> My initial hunch was that it should return a 3D image but I am not so sure anymore...

I would go for the first option. If the user asks for an iterator, returning an iterator with a warning seems better to me. 

I look at the explanation of `return_iterator` which says:
` Returns an iterator on the content of the niimg file input.`

In that case would it be wrong to create a fix just changing the last line in `check_niimg` as follows:

```python
    if return_iterator:
        return (
            new_img_like(niimg, _get_data(niimg)[:, :, :, i], niimg.affine)
            for i in range(niimg.shape[-1])
        )
```
changing ` for i in range(niimg.shape[3])` to ` for i in range(niimg.shape[-1])`

nilearn/image/image.py
bthirion: So `check_and load` ?

bthirion:
Cowardly waiting for tests to pass before reviewing...

Remi-Gau:
> Cowardly waiting for tests to pass before reviewing...

You are very wise! Teach me how you do this! 😉

nilearn/_utils/tests/test_niimg.py
```diff
@@ -87,3 +94,172 @@ def test_load_niimg(img1, tmp_path):
...
+    # Tests with file path
+    assert repr_niimgs("test") == "test"
+    assert repr_niimgs("test", shorten=False) == "test"
+
+    # Shortening long names by default
+    long_name = "this-is-a-very-long-name-for-a-nifti-file.nii"
+    short_name = "this-is-a-very-lon..."
```
bthirion, line 114: not .nii?

nilearn/image/image.py
```diff
@@ -5,32 +5,37 @@
...
 
 import numpy as np
 from joblib import Memory, Parallel, delayed
 from nibabel import Nifti1Image, Nifti1Pair, load, spatialimages
+from nibabel.spatialimages import SpatialImage
+from numpy.testing import assert_array_equal
 from scipy.ndimage import gaussian_filter1d, generate_binary_structure, label
 from scipy.stats import scoreatpercentile
 
+import nilearn as ni
```
bthirion: why do we need that ?

bthirion (commented):
Looks good overall.

nilearn/_utils/tests/test_niimg.py
Remi-Gau: nope, when used by repr_img has shorten to True by default, so it will just cut long filenames

nilearn/image/image.py
Remi-Gau: needed for the gloabal EXPAND_PATH_WILDCARDS
refactored to make explicirt

Remi-Gau:
> If the user asks for an iterator, returning an iterator with a warning seems better to me.

I went with that option

nilearn/image/image.py
Remi-Gau: - pinging @ghisvail on you might have an opinion on this.
- context: this PR moves the check_img* private utils of nilearn to the public API (see your issue https://github.com/nilearn/nilearn/issues/5696
- any opinion on renaming those functions so their name match their behavior better?

nilearn/image/image.py
Remi-Gau: preferably in a later PR if it can be done at all

Remi-Gau:
> > If the user asks for an iterator, returning an iterator with a warning seems better to me.
> 
> I went with that option

ok actually now iter_img does not throw the expected error for `iter_img(img_3d)` so I think that in fact we should just thrown an error

nilearn/image/image.py
ghisvail: I am thinking of packagers like myself who will need to patch older scientific code, which are unlikely to be ported, to be compatible with the new API. In order to limit the complexity of those patches, the best would be to limit changes to the import path only, and avoid function renaming.

That being said Hande's comment is fair, so the old names could be kept for legacy usage, whilst new ones with a better naming convention could be introduced, like `check_and_load` as proposed by Bertrand. Sounds to me like the best of both worlds.

Remi-Gau: note that the "old import path" from nilearn/_utils/niimg_conversion.py will work until version 0.14 (with the old names) with a deprecation warning to give some time to adapt: the message there will be clear what the new function is called.

ghisvail: > with the old names) with a deprecation warning to give some time to adapt

which is very good!

But there won't be any adaptation from research code published today and no longer being actively maintained. So adaptation will come from downstream packagers eventually, and a patch like this is simpler to maintain:

```diff
-from nilearn._utils.niimg_conversion check_niimg
+from nilearn.image.image import check_niimg
```

than:

```diff
-from nilearn._utils.niimg_conversion check_niimg
+from nilearn.image.image import check_and_load

- check_niimg(...)
+ check_and_load(...)

- check_niimg(...)
+ check_and_load(...)

- check_niimg(...)
+ check_and_load(...)

...
```

My 2 cents.

bthirion: agreed.

Remi-Gau: OK reverted the name change

nilearn/_utils/tests/test_niimg.py
```diff
@@ -87,3 +94,172 @@ def test_load_niimg(img1, tmp_path):
...
+        + "this-is-again-another-super-very-long-name-for-a-nifti-file.nii]"
+    )
+
+    assert (
+        repr_niimgs(list_of_size_4, shorten=False)
+        == long_rep_long_list_long_names
+    )
+
+
+def test_repr_niimgs_with_niimg_pathlib():
```
bthirion, line 213: I like to have short docstrings for test functions

nilearn/_utils/tests/test_niimg.py
```diff
@@ -87,3 +94,172 @@ def test_load_niimg(img1, tmp_path):
...
+        "[this-is-a-very-lon...,\n         ...\n this-is-again-anot...]"
+    )
+
+    assert (
+        repr_niimgs(list_of_size_4, shorten=True)
+        == shortened_rep_long_list_long_names
+    )
+
+
+def test_repr_niimgs_force_long_names():
```
bthirion, line 161: I like to have short docstrings for test functions

nilearn/_utils/tests/test_niimg.py
bthirion, line 99: I like to have short docstrings for test functions

bthirion (approved):
Thx.

doc/changes/latest.rst
```diff
@@ -47,6 +47,8 @@ Fixes
 Enhancements
 ------------
 
+- :bdg-success:`API` The functions :func:`~image.check_niimg`, :func:`~image.check_niimg_3d` and :func:`~image.check_niimg_4d` are now part of our public API. The content of ``nilearn._utils.niimg_conversion`` and ``nilearn.image.utils`` was moved to ``nilearn.image.image``. For a smoother transition some functions from ``nilearn._utils.niimg_conversion`` will still be importable from this module till Nilearn 0.14.0 (:gh:`5788` by `Rémi Gau`_).
+
```
```suggestion
- :bdg-success:`API` The functions :func:`~image.check_niimg`, :func:`~image.check_niimg_3d` and :func:`~image.check_niimg_4d` are now part of our public API. The content of ``nilearn._utils.niimg_conversions`` and ``nilearn.image.utils`` was moved to ``nilearn.image.image``. For a smoother transition some functions from ``nilearn._utils.niimg_conversions`` will still be importable from this module till Nilearn 0.14.0 (:gh:`5788` by `Rémi Gau`_).

```

hndgzkn: If I don't miss something, I think there are a few problems here:

- The contents of  `nilearn.image.utils` seem to remain almost the same, not moved to `nilearn.image.image`.
- The name of the package is changed from `nilearn._utils.niimg_conversions` to`nilearn._utils.niimg_conversion`. Is there a reason for that? Because that will do the same as moving it to `nilearn.image`, it will change the import path.
- `For a smoother transition some functions` : Here **some** is ambiguous and gives the impression that there might be other functions still importable which is not the case.

nilearn/_utils/tests/test_niimg_conversions.py
```diff
@@ -1,632 +0,0 @@
-"""Test the niimg_conversions.
```
hndgzkn, line 1: Was there a reason to change the name of the file?

nilearn/_utils/niimg_conversions.py
```diff
@@ -0,0 +1,50 @@
+"""Module to be deprecated in nilearn 0.14.0.
```
hndgzkn, line 1: If there is not specific reason, I think the name of this module should remain the same, ie `niimg_conversions`.

Remi-Gau: good catch: defo a mistake on my part

doc/changes/latest.rst
Remi-Gau: > * The contents of  `nilearn.image.utils` seem to remain almost the same, not moved to `nilearn.image.image`.

I thought I had moved things: I suspect I messed up something on git operation. Will do again.

> * The name of the package is changed from `nilearn._utils.niimg_conversions` to`nilearn._utils.niimg_conversion`. Is there a reason for that? Because that will do the same as moving it to `nilearn.image`, it will change the import path.

fixed

> * `For a smoother transition some functions` : Here **some** is ambiguous and gives the impression that there might be other functions still importable which is not the case.

made it very explicit which functions are still there

nilearn/_utils/tests/test_niimg_conversions.py
Remi-Gau: nope: now reverted

Remi-Gau:
OK the failures are unrelated so I will merge and mark the slow tests in another PR

PR #5789, opened 2025-10-24
[ENH] Add mesh centering

pbarbarant:

This PR adds mesh centering to the matplotlib backend. The mesh is translated to have a barycenter at (0, 0, 0), which avoids annoying offsets, in particular when plotting both hemispheres.

Before:

After:
image


pbarbarant:
BTW the Plotly automatically centers the mesh, so I only modified the matplotlib backend.

bthirion (approved):
LGTM, thx.

nilearn/plotting/tests/baseline/test_plot_surf_surface_both-anterior-plot_surf.png
Remi-Gau, line 1: it seems that the anterior and lateral view has been strangely affected when using some type of surfaces (inflated is ok but pial looks... well... 'squished')

pbarbarant: That is weird and oddly specific 😅, I'm going to investigate

pbarbarant: Problem found, `pytest_mpl` uses the "classic" stylesheet by default which sets `plt.rcParams["axes.autolimit_mode"]` to `round_numbers` instead of `data`, thus squishing the plot.
We can either:

1. Set `plt.rcParams["axes.autolimit_mode"] = data` at the test level.
2. Set `plt.rcParams["axes.autolimit_mode"] = data` at the top of `test_baseline_comparisons.py` to affect all tests.
3. Blame it on `pytest_mpl` and keep the squished baselines since the figures are just for testing.

I'm hesitating between 1 and 3 but I'm curious about your opinion.

Remi-Gau: good catch!

that must have been a headache to trace

more in favor of 1 because this way, it's more explicit in the test why we do this and the figures look more what one sees usuall

pbarbarant: I added the override using a decorator to the test. It should be all good!

Remi-Gau:
thanks for taking care of this one @pbarbarant

PR #5790, opened 2025-10-24
[ENH] Add title for masker reports

### LINKED ISSUE #5516: [ENH] add titles to masker reports

GLM report allow users to specify a title, it'd be nice to be able to do the same for masker reports.

### Benefits to the change

mostly convenience, but can also help distinguish reports generated with maskers with different attributes

```python
from nilearn.datasets import fetch_atlas_msdl, fetch_development_fmri
from nilearn.maskers import NiftiMapsMasker

atlas = fetch_atlas_msdl()

atlas_filename = atlas["maps"]

data = fetch_development_fmri(n_subjects=1)

masker = NiftiMapsMasker(
    maps_img=atlas_filename,
    standardize="zscore_sample",
    standardize_confounds=True,
    memory="nilearn_cache",
    memory_level=1,
)
masker.fit()
report = masker.generate_report(displayed_maps=[2, 6, 7, 16, 21], title="OK")
report.open_in_browser()

masker.fit_transform(data.func[0], confounds=data.confounds)
report = masker.generate_report(displayed_maps=[2, 6, 7, 16, 21], title="Not OK")
report.open_in_browser()
```

hndgzkn:
- Closes #5516

- Create `nilearn.maskers._mixin._ReportingMixin` to add reporting to estimators
  - define `generate_report(title=None)` in _ReportingMixin 
- Inherit _ReportingMixin  in the following estimators
  - NiftiLabelsMasker
  - NiftiMapsMasker
  - Niftimasker
  - NiftiSpheresMasker
  - SurfaceLabelsMasker
  - SurfaceMapsMasker
  - SurfaceMasker
- Move dependencies to `nilearn.maskers` from `nilearn.reporting` inside the code

### DIFF
--- a/nilearn/maskers/nifti_labels_masker.py
+++ b/nilearn/maskers/nifti_labels_masker.py
@@ -26,7 +26,7 @@ from nilearn._utils.param_validation import (
     check_reduction_strategy,
 )
 from nilearn.image import get_data, load_img, resample_img
-from nilearn.maskers._mixin import _LabelMaskerMixin, _ReportingMixin
+from nilearn.maskers._mixin import _LabelMaskerMixin
 from nilearn.maskers._utils import compute_middle_image
 from nilearn.maskers.base_masker import (
     BaseMasker,
@@ -69,7 +69,7 @@ class _ExtractionFunctor:
 
 
 @fill_doc
-class NiftiLabelsMasker(_ReportingMixin, _LabelMaskerMixin, BaseMasker):
+class NiftiLabelsMasker(_LabelMaskerMixin, BaseMasker):
     """Class for extracting data from Niimg-like objects \
        using labels of non-overlapping brain regions.
 
--- a/nilearn/maskers/base_masker.py
+++ b/nilearn/maskers/base_masker.py
@@ -35,6 +35,7 @@ from nilearn.image import (
     resample_img,
     smooth_img,
 )
+from nilearn.maskers._mixin import _ReportingMixin
 from nilearn.masking import load_mask_img, unmask
 from nilearn.signal import clean
 from nilearn.surface.surface import SurfaceImage, at_least_2d, check_surf_img
@@ -215,6 +216,7 @@ def mask_logger(step, img=None, verbose=0):
 
 @fill_doc
 class BaseMasker(
+    _ReportingMixin,
     TransformerMixin,
     CacheMixin,
     BaseEstimator,
@@ -496,7 +498,9 @@ class BaseMasker(
         return signals
 
 
-class _BaseSurfaceMasker(TransformerMixin, CacheMixin, BaseEstimator):
+class _BaseSurfaceMasker(
+    _ReportingMixin, TransformerMixin, CacheMixin, BaseEstimator
+):
     """Class from which all surface maskers should inherit."""
 
     def _more_tags(self):
--- a/nilearn/maskers/nifti_maps_masker.py
+++ b/nilearn/maskers/nifti_maps_masker.py
@@ -17,7 +17,6 @@ from nilearn._utils.param_validation import (
     check_params,
 )
 from nilearn.image import clean_img, get_data, index_img, resample_img
-from nilearn.maskers._mixin import _ReportingMixin
 from nilearn.maskers._utils import compute_middle_image
 from nilearn.maskers.base_masker import (
     BaseMasker,
@@ -47,9 +46,7 @@ class _ExtractionFunctor:
 
 
 @fill_doc
-class NiftiMapsMasker(
-    _ReportingMixin, ClassNamePrefixFeaturesOutMixin, BaseMasker
-):
+class NiftiMapsMasker(ClassNamePrefixFeaturesOutMixin, BaseMasker):
     """Class for extracting data from Niimg-like objects \
        using maps of potentially overlapping brain regions.
 
--- a/nilearn/maskers/nifti_masker.py
+++ b/nilearn/maskers/nifti_masker.py
@@ -16,7 +16,6 @@ from nilearn._utils.niimg import img_data_dtype
 from nilearn._utils.niimg_conversions import check_niimg, check_same_fov
 from nilearn._utils.param_validation import check_params
 from nilearn.image import crop_img, resample_img
-from nilearn.maskers._mixin import _ReportingMixin
 from nilearn.maskers._utils import compute_middle_image
 from nilearn.maskers.base_masker import (
     BaseMasker,
@@ -204,9 +203,7 @@ def filter_and_mask(
 
 
 @fill_doc
-class NiftiMasker(
-    _ReportingMixin, ClassNamePrefixFeaturesOutMixin, BaseMasker
-):
+class NiftiMasker(ClassNamePrefixFeaturesOutMixin, BaseMasker):
     """Applying a mask to extract time-series from Niimg-like objects.
 
     NiftiMasker is useful when preprocessing (detrending, standardization,
--- a/nilearn/maskers/nifti_spheres_masker.py
+++ b/nilearn/maskers/nifti_spheres_masker.py
@@ -25,7 +25,6 @@ from nilearn._utils.niimg_conversions import (
 from nilearn.datasets import load_mni152_template
 from nilearn.image import resample_img
 from nilearn.image.resampling import coord_transform
-from nilearn.maskers._mixin import _ReportingMixin
 from nilearn.maskers._utils import compute_middle_image
 from nilearn.maskers.base_masker import (
     BaseMasker,
@@ -226,9 +225,7 @@ class _ExtractionFunctor:
 
 
 @fill_doc
-class NiftiSpheresMasker(
-    _ReportingMixin, ClassNamePrefixFeaturesOutMixin, BaseMasker
-):
+class NiftiSpheresMasker(ClassNamePrefixFeaturesOutMixin, BaseMasker):
     """Class for masking of Niimg-like objects using seeds.
 
     NiftiSpheresMasker is useful when data from given seeds should be
--- a/nilearn/maskers/surface_labels_masker.py
+++ b/nilearn/maskers/surface_labels_masker.py
@@ -19,7 +19,7 @@ from nilearn._utils.param_validation import (
     check_reduction_strategy,
 )
 from nilearn.image import mean_img
-from nilearn.maskers._mixin import _LabelMaskerMixin, _ReportingMixin
+from nilearn.maskers._mixin import _LabelMaskerMixin
 from nilearn.maskers.base_masker import _BaseSurfaceMasker, mask_logger
 from nilearn.surface.surface import (
     SurfaceImage,
@@ -51,9 +51,7 @@ def signals_to_surf_img_labels(
 
 
 @fill_doc
-class SurfaceLabelsMasker(
-    _ReportingMixin, _LabelMaskerMixin, _BaseSurfaceMasker
-):
+class SurfaceLabelsMasker(_LabelMaskerMixin, _BaseSurfaceMasker):
     """Extract data from a SurfaceImage, averaging over atlas regions.
 
     .. nilearn_versionadded:: 0.11.0
--- a/nilearn/maskers/surface_maps_masker.py
+++ b/nilearn/maskers/surface_maps_masker.py
@@ -25,7 +25,6 @@ from nilearn._utils.param_validation import (
     check_params,
 )
 from nilearn.image import index_img, mean_img
-from nilearn.maskers._mixin import _ReportingMixin
 from nilearn.maskers.base_masker import _BaseSurfaceMasker, mask_logger
 from nilearn.surface.surface import (
     SurfaceImage,
@@ -37,9 +36,7 @@ from nilearn.surface.utils import check_polymesh_equal
 
 
 @fill_doc
-class SurfaceMapsMasker(
-    _ReportingMixin, ClassNamePrefixFeaturesOutMixin, _BaseSurfaceMasker
-):
+class SurfaceMapsMasker(ClassNamePrefixFeaturesOutMixin, _BaseSurfaceMasker):
     """Extract data from a SurfaceImage, using maps of potentially overlapping
     brain regions.
 
--- a/nilearn/maskers/surface_masker.py
+++ b/nilearn/maskers/surface_masker.py
@@ -16,16 +16,13 @@ from nilearn._utils.masker_validation import (
 )
 from nilearn._utils.param_validation import check_params
 from nilearn.image import concat_imgs, mean_img
-from nilearn.maskers._mixin import _ReportingMixin
 from nilearn.maskers.base_masker import _BaseSurfaceMasker, mask_logger
 from nilearn.surface.surface import SurfaceImage, at_least_2d, check_surf_img
 from nilearn.surface.utils import check_polymesh_equal
 
 
 @fill_doc
-class SurfaceMasker(
-    _ReportingMixin, ClassNamePrefixFeaturesOutMixin, _BaseSurfaceMasker
-):
+class SurfaceMasker(ClassNamePrefixFeaturesOutMixin, _BaseSurfaceMasker):
     """Extract data from a :obj:`~nilearn.surface.SurfaceImage`.
 
     .. nilearn_versionadded:: 0.11.0


hndgzkn:
The below code (from #5516) displays the reports as follows:

```
from nilearn.datasets import fetch_atlas_msdl, fetch_development_fmri
from nilearn.maskers import NiftiMapsMasker

atlas = fetch_atlas_msdl()

atlas_filename = atlas["maps"]

data = fetch_development_fmri(n_subjects=1)

masker = NiftiMapsMasker(
    maps_img=atlas_filename,
    standardize="zscore_sample",
    standardize_confounds=True,
    memory="nilearn_cache",
    memory_level=1,
)
masker.fit()
report = masker.generate_report(displayed_maps=[2, 6, 7, 16, 21], title=None)
report.open_in_browser()
```






If `title=None`:

hndgzkn:
I've made a quick fix, but I think report generation could in general be simplified.

bthirion (commented):
The PR LGTM but it doesn't build...

Happy to see some simplifications in a future PR if you can propose some ;-)

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -307,11 +307,22 @@ def _post_masking_atlas(self, visualize=False):
...
+        title : str, default=None
+            title for the report
+
+        Returns
+        -------
+        list(None) or HTMLReport
+        """
         from nilearn.reporting.html_report import generate_report
 
         self._init_report_content()
```
Remi-Gau: I think all maskers should now have an `_init_report_content` that is called during fitr: I think we should set the title to None by default via this function.

nilearn/reporting/html_report.py
```diff
@@ -272,6 +268,9 @@ def generate_report(estimator):
     # Generate a unique ID for this report
     data["unique_id"] = str(uuid.uuid4()).replace("-", "")
 
+    if data.get("title") is None:
+        data["title"] = estimator.__class__.__name__
```
Remi-Gau, line 284: oh actually instead of setting title to none by default in _init_report_content we could set it to the name of the class of the masker

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -307,11 +307,22 @@ def _post_masking_atlas(self, visualize=False):
...
         return masked_atlas, removed_region_ids, removed_region_names, display
 
-    def generate_report(self):
-        """Generate a report."""
+    def generate_report(self, title=None):
+        """Generate a report.
+
+        Parameters
+        ----------
+        title : str, default=None
```
Remi-Gau:
```suggestion
        title : :obj:`str` or None, default=None
```

should be applied to all doc string

Remi-Gau:
> I think report generation could in general be simplified.

oh boy, don't even get me started on this: 100% agree.

I think that swithching to jinja may help as a starting point: am already on it

Remi-Gau:
FYI: you are getting some errors because it should be possible to use generate report on unfitted maskers: the user should be warned about it and the report should contain very clear warning about it

Remi-Gau:
there are some systematic tests for the report generation of the maskers in see at the end of nilearn/_utils/estimator_checks.py

probably worth adding some checks for that in there:
- when no title is passed the html of the webpage title and first heading contain the name of the maskers
- when a title is passed it should be the one passed

nilearn/maskers/nifti_labels_masker.py
hndgzkn: Shall we give a warning if not fitted and generate_report is called?

hndgzkn:
> > I think report generation could in general be simplified.
> 
> oh boy, don't even get me started on this: 100% agree.
> 
> I think that swithching to jinja may help as a starting point: am already on it

OK, I'll refactor a bit on the python side to simplify. Not to create merge conflicts later, I would like to ask you. I'm thinking of creating a reporting mixin and extracting the common implementation (generate_report, _init_report_content) there.

One more question: If unfitted model can also generate report, why don't we just create `_report_content` and `_report_data` if `report=True` in the constructor?

nilearn/maskers/nifti_labels_masker.py
hndgzkn: > Shall we give a warning if not fitted and generate_report is called?

This is already replied below, you can ignore this.

Remi-Gau:
> One more question: If unfitted model can also generate report, why don't we just create `_report_content` and `_report_data` if `report=True` in the constructor?

This may work: in general sklearn conventions about constructor are there should not be any logic in them and you just pass the parameters to similarly named public attributes.

I think that setting private attribute is ok though: please double check (I am pretty sure that some tests in estimator_checks will run such checks).

Other possiblity: when calling generator_report, run init_report_content if the masker is not fitted

hndgzkn:
One more question: If constructor is called with `reports=False` and then `generate_report` is called. What should be the behavior? Shall I return `None`?

Remi-Gau:
> One more question: If constructor is called with `reports=False` and then `generate_report` is called. What should be the behavior? Shall I return `None`?

https://github.com/nilearn/nilearn/blob/207e8bbbe5b354b351b125acc999220dadf031b0/nilearn/_utils/estimator_checks.py#L3699

pretty sure it still returns a report but an empty one: the idea being report=False is to skip all the stuff needed for reporting during fit() and transform()

Remi-Gau:
2 more things:
- possibly add this feature on some of the maskers reports generated by https://github.com/nilearn/nilearn/blob/main/doc/visual_testing/reporter_visual_inspection_suite.py, this way we can see the result in the reports on this page: https://nilearn.github.io/stable/modules/generated_reports/masker_reports_examples.html
- also may be worth using this feature in some of the examples

Remi-Gau:
@hndgzkn merging the main branch should fix some of the doc and tests failures

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -307,15 +317,7 @@ def _post_masking_atlas(self, visualize=False):
...
-    def generate_report(self):
-        """Generate a report."""
-        from nilearn.reporting.html_report import generate_report
-
-        self._init_report_content()
-
-        return generate_report(self)
-
-    def _reporting(self):
+    def _get_displays(self):
```
hndgzkn, line 319: Renamed `_reporting` in maskers as `_get_displays` and moved `_reporting` inside `ReportingMixin` to do common checks and then call `_get_displays`.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -570,25 +563,6 @@ def fit(self, imgs=None, y=None):
...
-        """
-        if not hasattr(self, "_report_content"):
-            self._report_content = {
-                "description": (
-                    "This reports shows the regions "
-                    "defined by the labels of the mask."
-                ),
-                "warning_message": None,
-                "number_of_regions": 0,
-            }
```
hndgzkn, line 587: Moved this part in all maskers to the constructor.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -570,25 +563,6 @@ def fit(self, imgs=None, y=None):
...
-
-        if not hasattr(self, "_reporting_data"):
-            self._reporting_data = None
-
```
hndgzkn, line 591: Unless the model is fitted, the estimator won't have `_reporting_data` and this can be checked using `has_report_data` function provided with `ReportingMixin`.

hndgzkn: With these two modifications no more need for such a function and to call it from fit.

nilearn/reporting/mixin.py
```diff
@@ -0,0 +1,67 @@
+"""Module that provides mixin class to add reporting functionality to
+estimators.
```
hndgzkn: Is it ok to name the module as `_mixin`? I am not sure about that.

Remi-Gau: yeah though I think that as it's being imported in another subpackage (maskers) I am not sure about the leading underscore

nilearn/reporting/mixin.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+
+import abc
+
+from nilearn.reporting.html_report import generate_report
+
+
+class ReportingMixin:
```
Remi-Gau: Before merging we should check how easily we can use this mixin for glm reports.

Probably doable but am afraid of having a mixin that is just tailored for the masker reports.

hndgzkn: I was planning to do some refactoring in a follow-up PR. It is also possible to that in this PR. Let me see if the PR still stays reviewable after using the mixin in glm as well.

hndgzkn:
The tests on Windows fail do to:

```
             for name in filenames:
                  fullname = os.path.join(dirpath, name)
                  try:
  >                   os.unlink(fullname)
  E                   PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpkfpr419w\\joblib\\nilearn\\image\\resampling\\resample_img\\a67ee14d360df70e05588fda17a89295\\output.pkl'
```

Do you have an idea on that @Remi-Gau ?

nilearn/reporting/mixin.py
Remi-Gau: yeah let's at least see how doable it would be in this PR: then we decide and maybe then maybe split into several PR for easier review.

FYI: given we want to guesstimate feasibility, don't worry too much about making tests pass for now: let's focus on what changes we should make at first.

Remi-Gau:
> The tests on Windows fail do to:
> 
> ```
>              for name in filenames:
>                   fullname = os.path.join(dirpath, name)
>                   try:
>   >                   os.unlink(fullname)
>   E                   PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpkfpr419w\\joblib\\nilearn\\image\\resampling\\resample_img\\a67ee14d360df70e05588fda17a89295\\output.pkl'
> ```
> 
> Do you have an idea on that @Remi-Gau ?

the error on 3.10 (https://github.com/nilearn/nilearn/actions/runs/18913626695/job/53991058947?pr=5790#step:8:13379) 

```
[gw1] win32 -- Python 3.10.11 D:\a\nilearn\nilearn\.tox\plotting\Scripts\python.exe
  worker 'gw1' crashed while running 'nilearn/glm/tests/test_first_level.py::test_glm_ar_estimates[ar_vals1]'
```

```
 FAILED nilearn/glm/tests/test_first_level.py::test_glm_ar_estimates[ar_vals1]
```

I have seen before and is a 'new thing' since we started only running tests on the code affected in a PR: have not yet had the time to investigate but it is usually due one of the parallel worker used by pytest-xdist failing

Remi-Gau:
OK the error on windows 3.14 is something new: I will push a commit to force a run on all OS and all python to check if this happens on all python versions

.github/workflows/test_with_tox.yml
```diff
@@ -84,9 +84,6 @@ jobs:
             id: py
             run: |
                 versions='py=["3.14", "3.13", "3.12", "3.11", "3.10"]'
-                if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
-                    versions='py=["3.14", "3.10"]'
-                fi
```
Remi-Gau, line 89: if we end up having to hack this often on PR, we could make it something that can be 'toggled' via a commit message: let's see how often we need to do it

Remi-Gau:
> OK the error on windows 3.14 is something new: I will push a commit to force a run on all OS and all python to check if this happens on all python versions

FYI I am curious to see if this is an error specific to windows and 3.14

nilearn/reporting/mixin.py
hndgzkn: Doc build gives errors. It is due to cyclic imports. Cannot import `nilearn.reporting.make_glm_report`. 

- `nilearn.maskers` has dependency on `nilearn.reporting`
- `nilearn.reporting` has dependency on `nilearn.glm`
- `nilearn.glm` has dependency on `nilearn.maskers` and `nilearn._utils` which also has dependency on `nilearn.maskers`.

I think we should remove this one by one. What I propose is:

- Move `nilearn._utils.estimator_checks` outside of `nilearn._utils`. It makes `nilearn._utils` dependent on all the estimators. If I am not mistaken, these are only for us, to make sure that estimators comply with sklearn estimators. Is it possible to move it to another private package, `nilearn._estimator_checks` for example ?
- `nilearn._utils.masker_validation` can be moved to `nilearn.maskers`. We can even define a method of masker class `is_compatible(imgs)` which checks the compatibility of the masker itself and the provided images.
- Using `ReportingMixin` we should remove the dependency from `nilearn.reporting` to `nilearn.glm`.
- Finally dependency from `nilearn.glm` to `nilearn.maskers` and `nilearn._utils` won't be a problem.

How does it sound to you @Remi-Gau ?

hndgzkn:
> FYI I am curious to see if this is an error specific to windows and 3.14

I think the tests were running on two different windows versions and both were failing. It was only on windows. But I did not verify for both of them if the reason was the same, I guess so.

nilearn/reporting/mixin.py
Remi-Gau: wait there is a small issue with some of the new utils.discovery function: should be fixed by https://github.com/nilearn/nilearn/pull/5710/files#r2477518255

hndgzkn: The above problem is independent. If I try to import make_glm_report in local, it is the same problem.

Remi-Gau: okidoki

hndgzkn:
@Remi-Gau I've moved `nilearn.reporting.mixin.ReportingMixin` to `nilearn.maskers._mixin._ReportingMixin`. 

This way we do not expose anything in public API and we fix the issue implementing `title` parameter for masker reports. even if we do not generalize _ReportingMixin, I think that is a clean solution for maskers that simplifies code, removes repeating code, etc. There is still room for improvement.

For generalizing this to GLM, or other reportable estimators, I think it would be better to 

- fix the issues I mentioned in [this](https://github.com/nilearn/nilearn/pull/5790#discussion_r2477246013) comment. 
- define a structure for reports depending on the estimator type 
  - which estimators will/would have reporting functionality
  - what is the structure of report for each estimator
- extract common parts of estimator reports, and think of how to implement a generic interface for diverging sections

How does it sound to you?

Remi-Gau:
> How does it sound to you?

I like the overall plan indeed 

> * which estimators will/would have reporting functionality

I can already tell you: for now all maskers (and the classes that inherit from them like some fo the nilearn.regions classes: but I do not think we properly implement / test those) and the FirstLevelModel and SecondLevelModel. In the near future several of the decoders should have reports too.

.github/workflows/test_with_tox.yml
```diff
@@ -84,9 +84,6 @@ jobs:
...
                 echo "Will run on the following python: ${versions}"
```
Remi-Gau, line 90:
```suggestion
                if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
                    versions='py=["3.14", "3.10"]'
                fi
                echo "Will run on the following python: ${versions}"
```

Remi-Gau: reverting

.github/workflows/test_with_tox.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```

Remi-Gau: reverting

nilearn/reporting/html_report.py
```diff
@@ -284,17 +282,19 @@ def generate_report(estimator):
...
             "\nReport generation not enabled!\nNo visual outputs created."
         )
 
-    if (
-        not hasattr(estimator, "_reporting_data")
-        or not estimator._reporting_data
-    ):
+    try:
+        check_is_fitted(estimator)
+    except NotFittedError:
```
Remi-Gau:
```suggestion
    if not estimator.__sklearn_is_fitted__()
```

simpler than a try catch and all estimators MUST have this method

nilearn/maskers/_mixin.py
```diff
@@ -393,3 +394,71 @@ def _generate_lut(self):
...
+        - summary : summary
+        - warning_message : possible warning message
+    _report_data : Contains data from model fit. If reporting is disabled, or
+    the model is not fit, this attribute does not exist.
+
+    Classes inheriting from ReportingMixin should implement ``_get_displays``
+    to return the displays to be embedded to the report.
+    """
+
+    def has_report_data(self):
```
Remi-Gau:
```suggestion
    def _has_report_data(self):
```

let's keep the method private to avoid exposing it in the public API

https://output.circle-artifacts.com/output/job/fc1e5af0-748e-4778-ade3-27f061419820/artifacts/0/dev/modules/generated/nilearn.maskers.NiftiMasker.html#nilearn.maskers.NiftiMasker.has_report_data


image

Remi-Gau (changes requested):
couple of things to change but the rest looks good to me AFAICT

Remi-Gau:
> oh actually I will ask for another change to start helping us standardize.
> 
> Can you add a check in estimator_checks:
> 
>     * if a given estimator has the report mixin it must have a non empty _report_content after initialisation

probably in `check_masker_generate_report`

nilearn/reporting/html_report.py
hndgzkn: I did not want to call a private method of estimator, outside of estimator class. 

I've seen [this page](https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html#sphx-glr-auto-examples-developing-estimators-sklearn-is-fitted-py) that recommends using `check_is_fitted`. 

I'd actually propose to change the other places where `__sklearn_is_fitted__()` is called outside of the estimator class, with `check_is_fitted`.

Remi-Gau: I don't thing that dunder methods (double underscore) are considered private.

Also check_is_fitted will call __sklearn_is_fitted__ and may return error whereas __sklearn_is_fitted__ always return a boolean making it easier to use.

Remi-Gau: for reference: https://www.datacamp.com/tutorial/python-private-methods-explained

hndgzkn: But they are not really meant to be called by the developer, no? My interpretation from [this link](https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html#sphx-glr-auto-examples-developing-estimators-sklearn-is-fitted-py) is that it is encouraged to use `check_is_fitted` to check if the estimator is fitted and only implement `__sklearn_is_fitted__` if necessary.

Remi-Gau: the __sklearn_is_fitted__ that are being called here have been implemented by us so I think we are fine calling them

Remi-Gau:
@hndgzkn
FYI am pretty sure that most of the failures on windows are not related to this PR.

nilearn/_utils/estimator_checks.py
```diff
@@ -3643,6 +3652,11 @@ def check_masker_generate_report(estimator):
     _check_html(report, is_fit=False)
     assert "Make sure to run `fit`" in str(report)
 
+    assert (
```
Remi-Gau: Am pretty sure those asserts can be moved almost at the top of this function, as it must receive an initialized masker as input.

Remi-Gau (changes requested):

Remi-Gau:
Just realized we don't have a changelog entry for this.

hndgzkn:
> Just realized we don't have a changelog entry for this.

Done!

I've also added tests. I'll add tests for the case `reports=False`

hndgzkn:
> I've also added tests. I'll add tests for the case `reports=False`

Done!

hndgzkn:
@Remi-Gau In this PR, each masker inherits from `_ReportingMixin`. If all maskers should support reporting, it is an option to only make the BaseMasker inherit _ReportingMixin.  The consequence could be that each new masker should implement necessary functions for reporting or it can raise an error that reporting is not implemented for that specific masker. I think this behavior would be more consistent.

Which one would you prefer?

Remi-Gau:
Yes !!! Move things to the 2 base maskers the nifti and the surface one.

We are not likely to add new maskers any time soon.

hndgzkn:
> Yes !!! Move things to the 2 base maskers the nifti and the surface one.

Done.

PR #5794, opened 2025-10-27
[MAINT] update atlas table image

Remi-Gau:
- Closes none

- update images used to show what the atlases look like
- given that those figures may still change before the next release, I would actually wait before merging and just update our release protocol to make updating those figures one of the classic steps to do before every release

### DIFF
--- a/doc/maintenance.rst
+++ b/doc/maintenance.rst
@@ -298,7 +298,7 @@ For example:
 Update atlas figures
 ....................
 
-The :ref:`datasets_ref <dataset API page>` has tables with figures listing all the atlases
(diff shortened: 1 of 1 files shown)


doc/images/deterministic_atlas_aal.png
Remi-Gau, line 1: I think that the big difference in colorscale for AAL is because the default version of the atlas was changed

bthirion (approved):
LGTM, thx.

emdupre:
Looks like the `build_docs` warning is valid ?

```
/home/runner/work/nilearn/nilearn/doc/maintenance.rst:301: WARNING: undefined label: 'dataset api page' [ref.ref]
```

PR #5795, opened 2025-10-27
[MAINT] use difumo instead of msdl for masker report embeddded in doc

Remi-Gau:
- Closes none

- getting a lot of doc build error trying to fetch the msdl atlas because it needed to build the masker reports (even when not running any examples: see https://nilearn.github.io/dev/modules/generated_reports/masker_reports_examples.html#nifti-maps-masker and https://nilearn.github.io/dev/modules/generated_reports/masker_reports_examples.html#surface-maps-masker), switching to difumo instead


Remi-Gau:
ok the get_data part of the doc build works so this is good to go: merging quickly because all the other PR are blocked because of this

PR #5800, opened 2025-10-28
[MAINT] explicitly define python versions to test in CI

### LINKED ISSUE #5798: 'hide' python versions that are not tested on PR
let me know if this is confusing when seeing the PR results

I think I could adapt the PR matrix dynamically but this will make the workflow a bit more complicated

see https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#using-an-output-to-define-two-matrices

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5783#discussion_r2460332077_

Remi-Gau:
- Closes #5798 

- explicitly define python versions to test in CI
- also adds 1 dummy slow test to make run that minimalist test runs at least always 1 slow test to run

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,13 +1,19 @@
 ---
 # similar to scikit-learn .codecov.yml
+# https://docs.codecov.com/docs/codecovyml-reference
 coverage:
     status:
+        # Commits should not make the overall
+        # project coverage decrease by more than 1%:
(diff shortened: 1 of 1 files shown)


Remi-Gau:
will merge as this should fix the tests on main

PR #5803, opened 2025-10-28
[FIX] circular import

Remi-Gau:
- Closes none but should fix warnings in the doc build

https://github.com/nilearn/nilearn/actions/runs/18875057040/job/53862963270#step:13:21588

Apparently some circular import was leading to nilearn.reporting not being properly imported during doc build leading to some function definition missing.

- avoid some circular import
- use minimum optional dependencies in tox


Remi-Gau:
affecting all other PR so I will merge

PR #5805, opened 2025-10-30
[FIX] remove estimator_type from input tags

Remi-Gau:
- Closes none

Changes proposed in this pull request:

- remove estimator_type from input tags
- should be more aligned with how sklearn does things

PR #5809, opened 2025-10-30
[REF] use jinja for masker reports and remove tempita 

Remi-Gau:
- Relates to #5599

- use jinja as template engine for masker reports
- remove tempita as it is now completely unused
- adapt masker report templates to be jinja templates
- allow to build reports embedded in doc by adding '[reports]' to commit message

### DIFF
--- a/nilearn/reporting/html_report.py
+++ b/nilearn/reporting/html_report.py
@@ -2,7 +2,6 @@
 
 import uuid
 import warnings
-from pathlib import Path
 from string import Template
 
 import pandas as pd
@@ -189,11 +188,11 @@ def _update_template(
     if template_name is None:
         template_name = "body_masker.jinja"
 
-    body_tpl_path = Path("html") / "maskers" / template_name
+    body_tpl_path = f"html/maskers/{template_name}"
 
     env = return_jinja_env()
 
-    body_tpl = env.get_template(str(body_tpl_path))
+    body_tpl = env.get_template(body_tpl_path)
 
     if "n_elements" not in data:
         data["n_elements"] = 0


nilearn/reporting/html_report.py
```diff
@@ -11,57 +12,91 @@
...
+    "SurfaceMapsMasker": "body_surface_maps_masker.jinja",
+    "MultiSurfaceMapsMasker": "body_surface_maps_masker.jinja",
 }
 
 
-def _get_estimator_template(estimator):
-    """Return the HTML template to use for a given estimator \
-    if a specific template was defined in ESTIMATOR_TEMPLATES, \
-    otherwise return the default template.
+class HTMLReport(HTMLDocument):
```
Remi-Gau, line 39: only moved to the top of the module

nilearn/reporting/data/html/maskers/body_nifti_labels_masker.jinja
```diff
@@ -1,61 +1,61 @@
...
   </div>
 </div>
+
 <div class="pure-u-1 pure-u-md-3-3">
   <p>
-    This report was generated based on information provided at instantiation and
-    fit time. Note that the masker can potentially perform resampling at
-    transform time.
+    This report was generated based on information provided at instantiation and  fit time.
+    Note that the masker can potentially perform resampling at transform time.
```
hndgzkn: This repeats in all reports. I do not remember if it was present for glm reports as well. If it is only for maskers, may be `base_masker.jinja` or alike might contain common parts of masker reports. If it would exist for all reports, could be moved up in a higher template.

hndgzkn: I've seen that it does not exist for some other maskers. Still might worth the effort to have a common base template for the ones that it works.

nilearn/reporting/html_report.py
```diff
@@ -156,22 +191,13 @@ def _update_template(
...
-        body_template_name = template_name
-    body_template_path = HTML_TEMPLATE_PATH / body_template_name
-    if not body_template_path.exists():
-        raise FileNotFoundError(f"No template {body_template_path}")
-    tpl = tempita.HTMLTemplate.from_filename(
-        str(body_template_path), encoding="utf-8"
-    )
+        template_name = "body_masker.jinja"
+
+    body_tpl_path = Path("html") / "maskers" / template_name
```
hndgzkn: Is there a way to get the type of the estimator? If we generalize the report for glm  and decoders as well, instead of generating the path with hard coded `maskers`, that would be good to use a parameter.

hndgzkn: An option might be to define `ESTIMATOR_TEMPLATES` with values as tuple (estimator_type, template_name).

hndgzkn (approved):
Some comments, otherwise LGTM.

nilearn/reporting/html_report.py
Remi-Gau: - each estimator should have a tag for this: have been reworking it in https://github.com/nilearn/nilearn/pull/5805

nilearn/reporting/data/html/maskers/body_nifti_labels_masker.jinja
Remi-Gau: extracted it in a separate partial

Remi-Gau:
> The PR LGTM. Is there a way to see the generated reports ?

will trigger a full doc build

nilearn/maskers/tests/test_mixin.py
```diff
@@ -69,10 +69,7 @@ def test_masker_reporting_true(masker, img_func):
     instances when ``reports=True``.
     """
     # check masker at initialization
-    assert masker._report_content is not None
```
Remi-Gau, line 72: these are already checked in estimator_checks

pretty more of the checks performed here can be removed as they are also done in estimator_checks but it's out of scope for this PR

hndgzkn: Keeping these tests here, we can observe which parts of the implementation are covered by the tests. I think estimator checks are not reflected in the coverage report. 

To avoid double run, do you see any inconvenience to remove them from estimator checks?

hndgzkn: I am asking this thinking that estimator checks are there more for checking compliance with sklearn's recommendations.

Remi-Gau: > I think estimator checks are not reflected in the coverage report.

I think they should be: if they are not we have a problem

Remi-Gau: > thinking that estimator checks are there more for checking compliance with sklearn's recommendations.

fair assumption but I think you have been mislead by the fact I did not update the doc in the estimators_check module: it should run the sklearn checks AND also our own homemade estiamtor checks: either those we create to replace some of the sklearn ones that do not make sense to run or the ones that have no equivalent in sklearn because our estimator have some very specific behavior.

Remi-Gau: will open a separate PR to update the doc there

.github/workflows/build-docs.yml
```diff
@@ -17,6 +17,7 @@
 #    - ``[skip doc]`` skip build of the documentation.
 #    - ``[example] name_of_example.py`` trigger build of some specific examples.
 #    - ``[force download]`` do not rely of cached datasets and force to redownload them.
+#    - ``[reports]`` build estimators reports that are embedded in the documentation.
```
Remi-Gau, line 20: @hndgzkn 
it should now be possible to trigger the build of the reports embedded in the doc in CI by commit message (without needing to build the entire doc)

PR #5822, opened 2025-11-04
[ENH] Refactor - simplify plotting surface functions

hndgzkn:
- Closes None

- Simplify `nilearn.plotting.surface` backend functions. Remove repetitive code.
- Reuse `nilearn.plotting._utils.get_colorbar_and_data_ranges` inside `nilearn.plotting._engine_utils.colorscale` to calculate colorbar ranges.
- In `nilearn.plotting._utils.get_colorbar_and_data_ranges` instead of raising error when `vmin != vmax` with `symmetric_cbar=True`, warn user and adjust values.

### DIFF
--- a/nilearn/glm/tests/test_first_level.py
+++ b/nilearn/glm/tests/test_first_level.py
@@ -313,6 +313,7 @@ def test_glm_target_shape_affine(shape_3d_default, affine_eye):
     assert z_image.shape == (10, 11, 12)
 
 
+@pytest.mark.slow
 def test_high_level_glm_with_data_with_mask(shape_3d_default):
     """Test GLM can be run with mask."""
     shapes, rk = [(*shape_3d_default, 5)], 3
@@ -374,6 +375,7 @@ def test_fmri_inputs_type_data_smoke(tmp_path, shape_4d_default):
     )
 
 
+@pytest.mark.slow
 def test_fmri_inputs_type_design_matrices_smoke(tmp_path, shape_4d_default):
     """Test processing of FMRI inputs with path, str for design matrix."""
     mask_file, fmri_files, design_files = write_fake_fmri_data_and_design(
@@ -740,6 +742,7 @@ def test_fmri_inputs_events_type(tmp_path):
     )
 
 
+@pytest.mark.slow
 def test_fmri_inputs_with_confounds(tmp_path):
     """Test with confounds and, events."""
     n_timepoints = 10
--- a/nilearn/plotting/tests/test_baseline_comparisons.py
+++ b/nilearn/plotting/tests/test_baseline_comparisons.py
@@ -94,6 +94,7 @@ def test_plot_functions_annotate(plot_func, img_3d_mni):
     return plot_func(img_3d_mni, annotate=False)
 
 
+@pytest.mark.slow
 @pytest.mark.mpl_image_compare
 @pytest.mark.parametrize(
     "display_mode", ["x", "y", "z", "yx", "xz", "yz", "ortho"]
--- a/nilearn/regions/tests/test_parcellations.py
+++ b/nilearn/regions/tests/test_parcellations.py
@@ -59,6 +59,7 @@ def test_errors_raised_in_check_parameters_fit(method, test_image):
         Parcellations(method=method, verbose=0).fit(test_image)
 
 
+@pytest.mark.slow
 @pytest.mark.flaky(
     reruns=5, reruns_delay=2, condition=sys.platform.startswith("win32")
 )


nilearn/plotting/tests/baseline/test_plot_surf_surface_colorbar_plotly_auto-True-plot_surf_roi.png
Remi-Gau, line 1: ok for this one and a few others the colorbar is different so it may justify committing it, but for many others we do not see any difference so maybe we do not need to update them.

Do you know if which tests fail if you do not generate new figures?

hndgzkn: The actual problem is, in main, if you run `tox run - pytest-mpl-generate` with Python 3.10, the images generated are not the same with the committed ones. We have lost compliance at some point. 

I think it is important to keep them aligned. When I started this PR, I've made some changes which would not affect the baseline images. However I've seen that the baseline images were modified, and I was surprised/confused. Then I've checked main, and have seen that the same problem is there. 

Later I've added some code in this PR that could change the baseline images. 

If we don't keep the baseline images up to date with code, it is not evident if my PR changed them or they were already different.

hndgzkn: I thought of opening a PR just to update the baseline images.

Remi-Gau: I think we should only update images if the tests that compare images with "tox run - pytest-ml" fail, no?

Remi-Gau: note that I OK having a separate PR to update those figures

Just to explain my concern about updating figures when tests do not fail.
If we update the figures every time git finds a difference even when no tests fail we are committing large(ish) change more often than needed and this my after a while inflate the size of the repo and make slow down cloning and other git operations of it in CI when getting the full history. 
On the long run that means that eventually have to purge the git history of some things an operation that I would like to avoid having to do too often.

hndgzkn: > I think we should only update images if the tests that compare images with "tox run - pytest-ml" fail, no?

Ok @Remi-Gau , I'm sorry to disturb you on this. It is my misunderstanding. I was thinking that, when baseline images changed, "tox run - pytest-ml" failed. Then I unnecessarily generated the images. No need to make a PR on that in that case. 

Then we only update the images which fail with "tox run - pytest-ml"?

Remi-Gau: no worries!!! 😉 

> Then we only update the images which fail with "tox run - pytest-ml"?

yeah that'd be my take: it's one of the thing that's run in CI, but as we saw in this PR some of the color bar labels have been changed, so I would expect some things to fail

but you highlight a problem: our tox recipe do not have a good way to only change some figures, they will always scrap the old ones and regenerate all of them

Remi-Gau: maybe let's remove the figures from this PR and see what tests fail?

nilearn/plotting/surface/_matplotlib_backend.py
```diff
@@ -663,8 +623,8 @@ def _plot_img_on_surf(
             colorbar=False,  # Colorbar created externally.
             threshold=threshold,
             bg_on_data=bg_on_data,
-            vmin=vmin,
-            vmax=vmax,
+            vmin=vmin_iter,
```
hndgzkn: This does not change the images, however vmin, vmax values in for loop wrote over vmin, vmax outside the loop. 

I am not sure if this is really necessary.

hndgzkn: I have not really done much change in this PR. Just simplified some parts, changed the place of some code, so that plotly and matplotlib backend functions can be comparable side by side. 

What I want to do is to reuse get_cbar_ticks in plotly as well to see the same ticks in plotly and matplotlib images. 
And also make colorscale reuse get_colorbar_and_data_ranges.

@Remi-Gau Would that be ok?

bthirion: > This does not change the images, however vmin, vmax values in for loop wrote over vmin, vmax outside the loop.

Does it have any impact ?

hndgzkn: For this image it does not have any impact. Actually here my question is not correct. 

The correct question is, while iterating with (hemi, view) values on image, 
- is it possible that the `vmin_iter`, `vmax_iter` values differ for `stat_map_iter = texture[hemi]`? 
  (I guess yes, but that is not the case for the example with `load_sample_motor_activation_image`)

- even if these values change, for all iter_image we display a single colorbar, with vmin - vmax obtained from whole image (or could have been specified by the user)

In that case, to comply with the colorbar, shall we not specify the `vmin`, `vmax` of the colorbar for plotting each iter_image?

hndgzkn: I've created two images one in main branch, the other is in this branch using the code below:

```python

fig, _ = plot_img_on_surf(
        load_sample_motor_activation_image(),
        title="multiple views of the 3D volume",
        hemispheres="both",
        views=["lateral", "medial", "dorsal", "ventral", "anterior", "posterior"]
    )
```

| On main branch: | On current branch: |
|-----------------------------|-----------------------------|
| test_plot_img_on_surf_main |test_plot_img_on_surf |


The test `tox run -e pytest_mpl` passes.

hndgzkn: @Remi-Gau There is no baseline image created for `plot_img_on_surf`. I can add the above code to baseline tests if that is convenient?

hndgzkn: Another question here, for each sub image, I can put subtitle titles as hemi - view.

nilearn/plotting/_engine_utils.py
```diff
@@ -72,30 +73,9 @@ def colorscale(
...
-        vmax = abs_values.max()
-    if vmin is None:
-        vmin = values.min()
-    # cast to float to avoid TypeError if vmax/vmin is a numpy boolean
-    vmax = float(vmax)
-    vmin = float(vmin)
-
-    if symmetric_cmap:
-        vmax = max(abs(vmin), abs(vmax))
-        vmin = -vmax
```
hndgzkn, line 98: Using `get_colorbar_and_data_ranges` instead.

nilearn/plotting/_utils.py
```diff
@@ -231,10 +231,16 @@ def get_colorbar_and_data_ranges(
             vmin = -vmax
         elif vmax is None:
             vmax = -vmin
-        elif not np.isclose(vmin, -vmax):
```
hndgzkn, line 234: Instead of raising error, user is warned that `vmin` and `vmax` values are going to be adjusted to form a symmetric colorbar. (The behavior in colorscale)

nilearn/plotting/surface/_matplotlib_backend.py
Remi-Gau: > @Remi-Gau There is no baseline image created for `plot_img_on_surf`. I can add the above code to baseline tests if that is convenient?

yes please !!! 🙏🏻

Remi-Gau: > Another question here, for each sub image, I can put subtitle titles as hemi - view.

let's try but I thought that the title parameter would only allow something for the main title figure, no?

hndgzkn: Yes, but inside the code it is possible to put title for each sub image.

Remi-Gau: ha ok but then that would force this extra label to always appear on all figures, no? Not sure that we want to have that without a way to toggle this off. And that would require an extra parameter, no?

Remi-Gau: that would be for another PR, but it could be an 'annotate' parameter  like we have for volume based plots: https://nilearn.github.io/dev/modules/generated/nilearn.plotting.plot_img.html

hndgzkn: Yes.

hndgzkn: > that would be for another PR, but it could be an 'annotate' parameter like we have for volume based plots: https://nilearn.github.io/dev/modules/generated/nilearn.plotting.plot_img.html

Ok, that seems reasonable.

PR #5827, opened 2025-11-06
[DOC] improve doc for estimator checks

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/5809#discussion_r2499758371

- improve doc for estimator_checks


Remi-Gau:
I wonder if some of this should be in the actual doc (in the maintenance page for example)

bthirion:
> I wonder if some of this should be in the actual doc (in the maintenance page for example)

This is more developer  than user stuff, right ?

Remi-Gau:
> > I wonder if some of this should be in the actual doc (in the maintenance page for example)
> 
> This is more developer than user stuff, right ?

yup

Remi-Gau:
internal doc changes so tests failures cannot be related: merging

PR #5829, opened 2025-11-06
[MAINT] fix some flaky tests on windows

Remi-Gau:
- Closes none

- testing directly on windows and reduce or improve comment on why tests are flaky
- some flaky tests are hard to make fail even on windows locally
- fix some line endings in some file by relying on pre-commit config

### DIFF
--- a/nilearn/utils/tests/test_discovery.py
+++ b/nilearn/utils/tests/test_discovery.py
@@ -35,8 +35,7 @@ def test_all_estimators(
 
 def test_all_functions():
     """Check number of functions in public API."""
-    fn = [x[0] for x in all_functions()]
-
+    fn = all_functions()
     print(fn)
     if is_matplotlib_installed():
         assert len(fn) == 170
--- a/nilearn/regions/tests/test_parcellations.py
+++ b/nilearn/regions/tests/test_parcellations.py
@@ -394,6 +394,7 @@ def test_parcellation_with_surface_and_confounds(method, rng):
     assert X_transformed.shape == (n_samples, 5)
 
 
+@pytest.mark.flaky(reruns=5, reruns_delay=2, condition=is_windows_platform())
 @pytest.mark.parametrize("method", METHODS)
 def test_parcellation_with_multi_surface(method, rng):
     """Test if parcellation works with surface data from multiple


.pre-commit-config.yaml
```diff
@@ -28,6 +28,7 @@ repos:
     -   id: check-yaml
     -   id: end-of-file-fixer
     -   id: mixed-line-ending
+        args: [--fix=lf]
```
Remi-Gau, line 30: this should help enforce consistent line endings across OS

could be done via git but slightly easier to do via pre-commit
https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings

nilearn/datasets/tests/test_struct.py
```diff
@@ -76,7 +80,12 @@ def test_fetch_oasis_vbm(tmp_path, request_mocker):
     assert isinstance(dataset.data_usage_agreement, str)
     assert request_mocker.url_count == 1
 
-    dataset = struct.fetch_oasis_vbm(
+
+def test_fetch_oasis_vbm_dartel_false(tmp_path, request_mocker):
```
Remi-Gau, line 84: main change in this file is to split this test in 2: should help with the data mocking

nilearn/interfaces/fmriprep/data/test_desc-ICA_mixing.tsv
Remi-Gau, line 1: change line endings to make consistent with other files in the repo

nilearn/interfaces/fmriprep/data/test_desc-tedana_metrics.tsv
Remi-Gau, line 1: change line endings to make consistent with other files in the repo

nilearn/utils/tests/test_discovery.py
```diff
@@ -33,13 +34,194 @@ def test_all_estimators(
...
+@pytest.mark.flaky(reruns=5, reruns_delay=2, condition=is_windows_platform())
+def test_all_functions():
     """Check number of functions in public API."""
-    fn = all_functions()
+    fn = [x[0] for x in all_functions()]
+
     print(fn)
-    assert len(fn) == 169
+
+    expected = (
```
Remi-Gau: adding this for now until the test flakyness is fixed

bthirion (approved):
LGTM, thx.

Remi-Gau:
> I feel a bit uncomfortable about flaky tests. Maybe we should discuss the politics about those.

@bthirion 
will add this iten to our next dev meeting, but will merge as this should help get more stable CI in the meantime

nilearn/decomposition/tests/test_canica.py
hndgzkn, line 88: I haven't checked the reason of failure. If that is due to this assertion, would it help to reduce precision? Default is 6.

hndgzkn: Another point, in my local I need to add `flaky` in `pyproject.toml` `markers` section otherwise the tests complain. It does not complain in CI?

Remi-Gau: just to check do you have `pytest-rerunfailures` installed?

if not, does intalling this fix it?

https://github.com/nilearn/nilearn/blob/fb65d63101b0bc640380af4c126f33781f8f0a23/pyproject.toml#L109

hndgzkn: Yes, thank you!

Remi-Gau: > I haven't checked the reason of failure. If that is due to this assertion, would it help to reduce precision? Default is 6.

testing locally and I am struggling to make this test fail so it is hard to debug.

Remi-Gau:
I will merge as is.
At least flaky tests are now easier to look for in the test suite as they have their own markers

Issue #5828, opened 2025-11-06
improve reporting


Remi-Gau:
mostly to keep track of some issues flagged in some PR:
- rename _get_display (maybe rename it back to _reporting)
- try to clarify the role of each method: _create_figure_for_report, _generate_figure, _get_displays...
- try to see if we can have similar methods and private attributes for all estimators with reporting functionalities: this would allow to have a single report mixin (see https://github.com/nilearn/nilearn/pull/5790#issuecomment-3469317687)
- see also TODOs in https://github.com/nilearn/nilearn/issues/5599

bthirion:
> rename _get_display (maybe rename it back to _reporting)

Why is it needed ?

hndgzkn:
>     * rename _get_display (maybe rename it back to _reporting)

Actually `_reporting` is still in `_ReportingMixin`. It contains common check in all `_reporting` functions, then calls `_get_displays`.

I think trying to unify a common `_ReportingMixin` for all estimators will clarify and fix necessary functions.

hndgzkn:
I have worked a bit on a generic `ReportMixin` that could possibly be inherited by estimators. Below is a draft to be used as a starting point of discussion if convenient:

```Python
from copy import deepcopy
from warnings import warn
from nilearn._utils.helpers import is_matplotlib_installed 

def _update_defaults(d: dict, u: dict):
   """Create and return a dictionary updating the values of d with values of u 
      recursively.
   """
    d = deepcopy(d)
    for k, v in u.items():
        if k in d:
            if isinstance(d[k], dict) and isinstance(v, dict):
                d[k] = _update_defaults(d[k], v)
            else:
                d[k] = v
    return d
        
        
class ReportMixin():
    """Mixin class to be inherited by the estimators with reporting functionality.

    Provides interfaces and implementations for common methods.

    Each inheriting estimator has below fields in its reporting content:

    title: Title of the report. 
           Can be set by the user at report generation; otherwise set to class name.
    description: Description of the report for that specific estimator. 
           Defined by the estimator class and cannot be changed by the user.
    summary: Summary of the report.
           Created depending on report data when generating report. Cannot be changed 
           by the user. 
    warnings: Warnings while generating the report. 
           If there are warnings an empty report displaying the warnings is generated.

    Inheriting estimators can define additional fields or update existing fields with 
    _REPORT_DEFAULTS class variable.

    A non-empty report has the following section:

    - head (title, description, estimator params, etc)
    - plots
    - summary
    
[...]

PR #5842, opened 2025-11-17
[REF] only include css navbar if navbar is used

Remi-Gau:
- Relates to #5828

- only include css navbar if navbar is used

### DIFF
--- a/nilearn/glm/tests/test_glm_reporter.py
+++ b/nilearn/glm/tests/test_glm_reporter.py
@@ -79,7 +79,7 @@ def check_glm_report(
 
     # check the navbar and its css is there
     includes.append('<nav class="navbar pure-g fw-bold" id="menu"')
-    includes.append('"Adapted from Pure CSS navbar" ')  # css
+    includes.append("Adapted from Pure CSS navbar")  # css
 
     if is_matplotlib_installed():
         excludes.extend(


nilearn/reporting/data/css/report.jinja
```diff
@@ -453,3 +376,5 @@ h6 {
 .nilearn-report .main-container {
   min-height: 700px !important;
 }
+
+{% block navbar %}{% endblock %}
```
Remi-Gau:
```suggestion
{% block navbar_css %}{% endblock %}
```

nilearn/reporting/data/css/report.jinja
```diff
@@ -453,3 +376,5 @@ h6 {
 .nilearn-report .main-container {
   min-height: 700px !important;
 }
+
+{% block navbar_css %}{% endblock %}
```
hndgzkn: This file resembles more of a css file. Isn't it possible to separate the css part as a css file?

nilearn/reporting/data/css/partials/navbar.css
```diff
@@ -0,0 +1,77 @@
+/* -----------------------  Navbar  ----------------------- */
```
hndgzkn, line 1: This seems like a css file, is the extension `jinja` a necessity?

Remi-Gau: yeah for this one we could keep the css extension

nilearn/reporting/data/css/report.jinja
Remi-Gau: let me check but I am afraid of css 'linters' to be a pain if we do this.

Remi-Gau: yeah this is flagged by pre-commit if I use the css extension

```
prettier.................................................................Failed
- hook id: prettier
- exit code: 2

nilearn/reporting/data/css/partials/navbar.css 63ms (unchanged)
[error] nilearn/reporting/data/css/report.css: SyntaxError: CssSyntaxError: Unknown word % (380:2)
[error]   378 | }
[error]   379 |
[error] > 380 | {% block navbar_css %}{% endblock %}
[error]       |  ^
[error]   381 |
```

let me try if I can pacify 'prettier'

hndgzkn: Isn't there a better place for `{% block navbar_css %}{% endblock %}`?

Or isn't it possible to keep this in `report.jinja` and separate the css part in `report.css` and include it.

nilearn/reporting/data/html/body_base.jinja
```diff
@@ -12,7 +12,7 @@
 />
 
 <style type="text/css">
-  {% include 'css/report.css' %}
+  {% include 'css/report.jinja' %}
 </style>
```
hndgzkn, line 17: Could it be possible to implement like:

```
<style type="text/css">

{% block report_css %}
  {% include 'css/report.css' %}
{% endblock %}

</style>
```

And if any report wants to add to this, it can add as :

```
{% block report_css %}
    {{ block.super }}
    {% include 'css/nav_bar.css' %}
{% endblock %}
```

I do not know jinja well but seems like calling super is possible https://www.djaodjin.com/blog/django-porting-templates-to-jinja2.blog

In that case, the jinja part in report.jinja can be removed and named as report.css may be.

nilearn/reporting/data/css/report.jinja
Remi-Gau: a previous comment got lost:

I tried the different 'ignore' methods listed in the prettier doc and they did not work.

- Seems djlint is better at handling jinja templates see https://github.com/nilearn/nilearn/pull/5853

Remi-Gau: > Isn't there a better place for `{% block navbar_css %}{% endblock %}`?

in any case the 'main css file' (report.css or report.jinja) will require to include some jinja syntax whhether it's  a 'block' or an 'if', which would upset prettier if we keep the css extension

nilearn/reporting/data/html/body_base.jinja
Remi-Gau: ha I see let me check
may be possible
even without calling super

hndgzkn: This is much better :)

Remi-Gau:
error is unrelated

PR #5850, opened 2025-11-19
[FIX] fix number of functions

Remi-Gau:
- Closes none but should fix failure on main

- fix number of functions


Remi-Gau:
merging to hopefully get green tests on main

PR #5851, opened 2025-11-19
[REF] do not use data_gen in conftest 

Remi-Gau:
- Closes none

- try to minimize relying on nilearn code to for the fixture defined globally in conftest

### DIFF
--- a/nilearn/maskers/tests/test_nifti_labels_masker.py
+++ b/nilearn/maskers/tests/test_nifti_labels_masker.py
@@ -72,6 +72,7 @@ def test_check_estimator_nilearn(estimator, check, name):  # noqa: ARG001
     check(estimator)
 
 
+@pytest.mark.slow
 def test_nifti_labels_masker(
     affine_eye, shape_3d_default, n_regions, length, img_labels
 ):
@@ -322,6 +323,7 @@ def test_nifti_labels_masker_resampling_to_data(affine_eye, n_regions, length):
     assert_array_equal(masker.labels_img_.affine, affine2)
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize("resampling_target", ["data", "labels"])
 def test_nifti_labels_masker_resampling(
     affine_eye,
@@ -493,6 +495,7 @@ def test_nifti_labels_masker_resampling_to_none(
         masker.fit_transform(fmri_img)
 
 
+@pytest.mark.slow
 def test_nifti_labels_masker_with_mask(
     shape_3d_default, affine_eye, length, img_labels
 ):
@@ -901,6 +904,7 @@ def test_check_labels_errors(shape_3d_default, affine_eye):
         masker.fit()
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize(
     "background",
     [
@@ -963,6 +967,7 @@ def test_region_names(
     )
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize(
     "background",
     [None, "background", "Background"],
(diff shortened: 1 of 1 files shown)


nilearn/image/utils.py
```diff
@@ -28,4 +28,7 @@ def get_indices_from_image(image) -> np.ndarray:
             "Niimg-Like, SurfaceImage, numpy array. "
             f"Got {image.__class__.__name__}"
         )
-    return np.unique(data)
+
+    labels_present = np.unique(data)
+
+    return labels_present[np.isfinite(labels_present)]
```
Remi-Gau, line 34: this PR sort of revealed that the look up tables of label masker would count nan and inf as region indices

nilearn/maskers/_mixin.py
```diff
@@ -312,6 +312,7 @@ def _generate_lut(self):
         labels = self.labels
 
         labels_present = get_indices_from_image(labels_img)
+
```
Remi-Gau:
```suggestion
```

bthirion:
> * Closes none
> 
> 
> Changes proposed in this pull request:
> 
>     * try to minimize relying on nilearn code to for the fixture defined globally in conftest

Hi, Can you explain why this is necesary ? Is it just for the sake of cleaning the dependency graph ?

Remi-Gau:
> > * Closes none
> > 
> > Changes proposed in this pull request:
> > ```
> > * try to minimize relying on nilearn code to for the fixture defined globally in conftest
> > ```
> 
> Hi, Can you explain why this is necesary ? Is it just for the sake of cleaning the dependency graph ?

In part, but I am also afraid of the circularity of using nilearn code to test nilearn code.

bthirion:
> > > * Closes none
> > > 
> > > Changes proposed in this pull request:
> > > ```
> > > * try to minimize relying on nilearn code to for the fixture defined globally in conftest
> > > ```
> > 
> > 
> > Hi, Can you explain why this is necesary ? Is it just for the sake of cleaning the dependency graph ?
> 
> In part, but I am also afraid of the circularity of using nilearn code to test nilearn code.

Sounds like philosophical a qualm ;-)  Anyway, the PR LGTM.

bthirion (approved):

nilearn/conftest.py
hndgzkn:
```suggestion
def img_maps(n_regions):
"""Generate fixture for default map image."""
return _img_maps(n_regions)
```

hndgzkn:
```suggestion
def img_labels(n_regions):
"""Generate fixture for default label image."""
return _img_labels(n_regions)
```

nilearn/conftest.py
```diff
@@ -484,13 +537,26 @@ def img_maps():
 def _img_labels():
```
hndgzkn:
```suggestion
def _img_labels(n_regions=None):
```

hndgzkn: It is mentioned that some tests require `n_regions` should not be changed. But it can be possible to add flexibility for some other tests. Just as a suggestion. 

If this is accepted, the value of `n_regions` should also be checked inside the code and set to default value.

hndgzkn: Another point is that all fixtures has a private twin. Why this is necessary? Could we think of a way to have only the public ones?

nilearn/conftest.py
hndgzkn: This is to provide to test the flexibility to call with different number of regions.

nilearn/conftest.py
Remi-Gau: In some places, we may want to import a function that does the same as the fixture but is not a fixture: that's the main reason for this setup

Remi-Gau:
failures are unrelated

PR #5856, opened 2025-11-21
[DOC] example on how to use the right template

Remi-Gau:
- Closes none

-
-


examples/07_advanced/plot_template_flow.py
```diff
@@ -0,0 +1,297 @@
...
+
+import numpy as np
+
+from nilearn.connectome import ConnectivityMeasure
+from nilearn.maskers import MultiNiftiLabelsMasker
+from nilearn.plotting import plot_matrix
+
+matrices = []
+
+for tpl, name in zip(
```
bthirion, line 99: Can you explain a bit more what you're going to do next ?

PR #5861, opened 2025-11-24
[FIX] More explicit warning and errors when cut coordinates are out of bounds

### LINKED ISSUE #5855: poor error handling when requesting to plot a slice that is out of bound
a bit of an edge case but the error message is not very helpful in this case

```python
from nilearn.plotting import plot_roi

dataset_ho = fetch_atlas_harvard_oxford("cort-maxprob-thr25-2mm")

# fine
plot_roi(
    dataset_ho.filename, display_mode="x", cut_coords=[-100, 100]
)

# fine
plot_roi(
    dataset_ho.filename, display_mode="x", cut_coords=[-100]
)

# fail
plot_roi(
    dataset_ho.filename, display_mode="x", cut_coords=[100]
)
```

```
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.10/site-packages/numpy/ma/core.py", line 7052, in min
    return obj.min(axis=axis, fill_value=fill_value, out=out, **kwargs)
AttributeError: 'list' object has no attribute 'min'

During handling of the above exception, another exception occurred:

  File "/home/remi-gau/github/nilearn/nilearn/tmp/tmp_er.py", line 15, in <module>
    plot_roi(
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/image/img_plotting.py", line 878, in plot_roi
    display = _plot_img_with_bg(
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/image/img_plotting.py", line 260, in _plot_img_with_bg
    display.add_overlay(
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/displays/_slicers.py", line 347, in add_overlay
    ims = self._map_show(
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/displays/_slicers.py", line 526, in _map_show
    kwargs["vmin"] = np.ma.min(
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.10/site-packages/numpy/ma/core.py", line 7056, in min
    return asanyarray(obj).min(axis=axis, fill_value=fill_value,
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.10/site-packages/numpy/ma/core.py", line 5978, in min
    result = self.filled(fill_value).min(
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.10/site-packages/numpy/_core/_methods.py", line 48, in _amin
    return umr_minimum(a, axis, None, out, keepdims, initial, where)
ValueError: zero-size array to reduction operation minimum which has no identity
```

isha822:
## Description

- closes #5855 

This PR fixes issue #5855 where requesting a cut coordinate outside the image bounds would fail silently (resulting in an empty plot with None data) or raise a confusing IndexError.

I have modified _slicers.py to catch the IndexError and raise a descriptive ValueError instead, informing the user that their coordinates are out of bounds.

## Changes
- Modified _map_show in nilearn/plotting/displays/_slicers.py.
- Replaced the silent except IndexError: pass block with a specific raise ValueError message.

## Testing
- Verified locally by attempting to plot an MNI image with cut_coords=[5000].
- Confirmed that the script now crashes with the correct ValueError message instead of failing silently.
-


isha822:
Thanks for the guidance! I have added the regression test in nilearn/plotting/tests/test_bug_fix.py as requested. I verified locally that it correctly catches the ValueError now.

nilearn/plotting/tests/test_bug_fix.py
```diff
@@ -0,0 +1,15 @@
...
+def test_out_of_bounds_error():
+    """Regression test: Ensure plotting out-of-bounds slice raises ValueError."""
+    # Create a small dummy image (10x10x10 voxels)
+    data = np.zeros((10, 10, 10))
+    img = nib.Nifti1Image(data, np.eye(4))
+
+    # This should crash with ValueError because 50 is > 10
+    # We use pytest.raises to "catch" the crash and mark the test as PASS if it crashes correctly
+    with pytest.raises(ValueError, match="out of bounds"):
+        plotting.plot_anat(img, display_mode="x", cut_coords=[50])
```
Remi-Gau: actually let's use plot_img instead of plot_anat: this pretty much the most generic plotting function we use, this should ensure that this fix can apply to all plotting functions

isha822: done!! Switched to plot_img

nilearn/plotting/tests/test_bug_fix.py
```diff
@@ -0,0 +1,20 @@
+import nibabel as nib
+import numpy as np
+import pytest
+
+from nilearn import plotting
+
+
+def test_out_of_bounds_error():
+    """Regression test: Ensure plotting out-of-bounds slice raises
+    ValueError.
```
Remi-Gau:
```suggestion
    """Ensure plotting out-of-bounds slice raises ValueError.
    
    Regression test for https://github.com/nilearn/nilearn/issues/5855

```

nilearn/plotting/tests/test_bug_fix.py
Remi-Gau, line 1: Let's move this test into: nilearn/plotting/image/tests/test_plot_img.py

instead of creating a new file

nilearn/plotting/tests/test_bug_fix.py
```diff
@@ -0,0 +1,20 @@
+import nibabel as nib
+import numpy as np
+import pytest
+
+from nilearn import plotting
+
+
+def test_out_of_bounds_error():
```
Remi-Gau:
```suggestion
def test_out_of_bounds_error(matplotlib_pyplot):
```

Remi-Gau: this should ensure that the test is only run when matplotlub is installed

nilearn/plotting/tests/test_bug_fix.py
Remi-Gau: this way we know where to find info about the bug that is being fixed

doc/changes/latest.rst
```diff
@@ -40,7 +40,7 @@ Fixes
 - :bdg-dark:`Code` Inform user correctly when reporting is enabled after the model is fit (:gh:`5836` by `Hande Gözükan`_).
 
 - :bdg-dark:`Code` Raise warning when :func:`~nilearn.image.crop_img` is called with an empty image and return the original image (:gh:`5837` by `Hande Gözükan`_).
-
+- :Fix :func:~nilearn.plotting.plot_img error handling when cut coordinates are out of bounds (:gh:5861) by :user:isha822.
```
```suggestion

- :Fix :func:~nilearn.plotting.plot_img error handling when cut coordinates are out of bounds (:gh:5861) by :user:isha822.
```

nilearn/plotting/image/tests/test_plot_img.py
```diff
@@ -146,3 +148,18 @@ def test_plot_img_transparency_binary_image(
...
+    """
+    # Create a small dummy image (10x10x10 voxels)
+    data = np.zeros((10, 10, 10))
+    img = nib.Nifti1Image(data, np.eye(4))
+
+    # This should crash with ValueError because 50 is > 10
+    # We use pytest.raises to "catch" the crash
+    #  and mark the test as PASS if it crashes correctly
+    with pytest.raises(ValueError, match="out of bounds"):
+        plotting.plot_img(img, display_mode="x", cut_coords=[50])
```
Remi-Gau:
```suggestion
def test_out_of_bounds_error(matplotlib_pyplot, img_3d_mni):
    """Ensure plotting out-of-bounds slice raises ValueError.
    
    This should crash with ValueError
    because the default 3D image fixture
    has less than 50 slices on the x axis.
    """
    with pytest.raises(ValueError, match="out of bounds"):
        plotting.plot_img(img_3d_mni, display_mode="x", cut_coords=[50])
```

Let's simplify this by using one of our image fixture.

bthirion (approved):

Remi-Gau:
@isha822 
this is causing some tests to fail when they should not.

See in the logs here:

https://github.com/nilearn/nilearn/actions/runs/19642736912/job/56279107981?pr=5861#step:8:21803


One example of test that should not fail:

```python
def test_plot_prob_atlas_radiological_view(matplotlib_pyplot, img_4d_rand_eye):
    """Smoke test for radiological view."""
    result = plot_prob_atlas(img_4d_rand_eye, radiological=True)
    assert result.axes.get("y").radiological is True
```

In this case the failure happens during a call to `BaseSlicer.add_overlay` that does not have cut coordinates as a parameter, so I suspect that the fix for this bug is misplaced.

Can you investigate?

isha822:
I've reverted the changes in _slicers.py to fix the regression. I moved the check up to plot_img in nilearn/plotting/img_plotting.py instead. This catches the error for users without breaking internal functions like add_overlay

nilearn/plotting/image/img_plotting.py
```diff
@@ -408,32 +408,38 @@ def plot_img(
...
+            transparency=transparency,
+            transparency_range=transparency_range,
+            **kwargs,
+        )
+    except IndexError:
+        raise ValueError(
+            "The requested cut coordinate is out of bounds for the image. "
+            "Please check that your cut_coords"
+            " are within the image field of view."
+        )
```
Remi-Gau: I am afraid this won't do: this is too generic, if '_plot_img_with_bg' may throw IndexError for other reasons than the one mentioned in the bug, so we can not have a blanket solution like this one

Remi-Gau:
if that helps working on this, I have updated the original issue with a traceback of how the bug shows when working with plot_img:

https://github.com/nilearn/nilearn/issues/5855#issuecomment-3579954884

isha822:
Thankyou for the feedback sir I completely removed the try/except block from _slicers.py to fix the regression.
I've implemented the validation logic in img_plotting.py as suggested (calculating the bounding box using the affine) to raise a ValueError before plotting starts. I also handled the imports locally to avoid circular dependency issues

Remi-Gau:
it seems to affect some of our high-level tests that check that the figures are unchanged: https://github.com/nilearn/nilearn/actions/runs/19706393170/job/56487042013?pr=5861#step:8:1767

Remi-Gau:
@hndgzkn 
can you have a look at this PR?
Where would be the best way to handle the problem, this PR is trying to fix.

nilearn/plotting/displays/_slicers.py
```diff
@@ -508,19 +508,14 @@ def _map_show(
...
                     )
-            except IndexError:
-                # We are cutting outside the indices of the data
-                data_2d = None
-                transparency_2d = None
 
-            data_2d_list.append(data_2d)
-            transparency_list.append(transparency_2d)
+                data_2d_list.append(data_2d)
+                transparency_list.append(transparency_2d)
```
hndgzkn: Removing `try-except`, this code block becomes part of the above if block which probably changes behavior.

hndgzkn: If the bug is fixed, is there a case that this try-except block might still be necessary?

nilearn/plotting/image/img_plotting.py
```diff
@@ -407,7 +407,59 @@ def plot_img(
     """
     check_params(locals())
     check_threshold_not_negative(threshold)
+    # --- Validation: Check if cut coords are in bounds ---
```
hndgzkn: Specified `cut_coords` could be checked in the corresponding slicer's `find_cut_coords` depending on slicer type.

isha822:
I've implemented the validation logic using check_niimg and coord_transform to calculate the bounding box, and wrapped it in a try/except block to avoid regressions.
However, I'm still hitting a persistent ValueError: zero-size array in the test_baseline_comparisons locally, even with the safety net. I suspect it might be an environment caching issue or something specific to my local setup.
I have upcoming university exams, so I won't be able to push further fixes for the next 5-6 days. Feel free to push to my branch if there's an obvious fix I missed! Otherwise, I will pick this back up next week.

bthirion:
> I've implemented the validation logic using check_niimg and coord_transform to calculate the bounding box, and wrapped it in a try/except block to avoid regressions. However, I'm still hitting a persistent ValueError: zero-size array in the test_baseline_comparisons locally, even with the safety net. I suspect it might be an environment caching issue or something specific to my local setup. I have upcoming university exams, so I won't be able to push further fixes for the next 5-6 days. Feel free to push to my branch if there's an obvious fix I missed! Otherwise, I will pick this back up next week.

Thx for your contribution, and good luck with your exams.

Remi-Gau:
> I've implemented the validation logic using check_niimg and coord_transform to calculate the bounding box, and wrapped it in a try/except block to avoid regressions. However, I'm still hitting a persistent ValueError: zero-size array in the test_baseline_comparisons locally, even with the safety net. I suspect it might be an environment caching issue or something specific to my local setup. I have upcoming university exams, so I won't be able to push further fixes for the next 5-6 days. Feel free to push to my branch if there's an obvious fix I missed! Otherwise, I will pick this back up next week.

thanks for pushing this and good luck with the exams

gonna have a look at this

Remi-Gau:
TODO:
- check test suite and doc for new warnings

Remi-Gau:
@hndgzkn I made a suggestion on where to raise errors and warnings: WDYT?

hndgzkn:
> @hndgzkn I made a suggestion on where to raise errors and warnings: WDYT?

While plotting image, the slicer is created using `plotting.displays._utils.get_create_display_fun` function which calls `Slicer.init_with_figure` class method on the requested slicer class. This calls `Slicer.find_cut_coords` class method. 

`find_cut_coords` is implemented for each slicer type. I think the best place would be there, for each type of slicer, check if the specified coordinates are compatible with slicer type, and raise error before doing any other operation. That would let us be precise on the problem with the specified `cut_coords` value as well.

Remi-Gau:
> > @hndgzkn I made a suggestion on where to raise errors and warnings: WDYT?
> 
> While plotting image, the slicer is created using `plotting.displays._utils.get_create_display_fun` function which calls `Slicer.init_with_figure` class method on the requested slicer class. This calls `Slicer.find_cut_coords` class method.
> 
> `find_cut_coords` is implemented for each slicer type. I think the best place would be there, for each type of slicer, check if the specified coordinates are compatible with slicer type, and raise error before doing any other operation. That would let us be precise on the problem with the specified `cut_coords` value as well.

yeah I agree this could provide slightly more information in message sent to the user but given how rarely we've hit this problem, I'd rather have a single check that happens a bit more downstream rather a different checks for 3 or 4 classes happening earlier: I don't find the trade-off convincing.

hndgzkn:
> I don't find the trade-off convincing.

I would say cleaner and simpler code.

Looking at the code, I think it's `find_cut_coords` duty to do this check. When it is not done there, the later checks add complexity to the function whose duty is to do something else.

Remi-Gau:
> > I don't find the trade-off convincing.
> 
> I would say cleaner and simpler code.
> 
> Looking at the code, I think it's `find_cut_coords` duty to do this check. When it is not done there, the later checks add complexity to the function whose duty is to do something else.

wanna give it a go: not sure it will be simpler but happy to proven wrong. 😉

hndgzkn:
> wanna give it a go: not sure it will be simpler but happy to proven wrong. 😉

Then, as this one fixes the problem, we can merge this as is (it does not introduce anything public). We can discuss the change in a follow-up PR. Would that be ok?

Remi-Gau:
> > wanna give it a go: not sure it will be simpler but happy to proven wrong. 😉
> 
> Then, as this one fixes the problem, we can merge this as is (it does not introduce anything public). We can discuss the change in a follow-up PR. Would that be ok?

OK let's do that

hndgzkn (approved):
LGTM.

Remi-Gau:
OK test failure is unrelated

Issue #5615, opened 2025-08-29
[ENH] allow feature selection in decoder to be independent of total number of voxels/vertices, and allow feature selection to be based on the number rather than percentage of voxels/vertices


hyang336:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe your proposed enhancement in detail.

Currently the screening percentile is normalized with respect to the whole brain even when an ROI mask is provided:

screening_percentile = screening_percentile * (
            reference_extent / mask_extent
        )
screening_percentile = min(screening_percentile, 100.0)

This behavior is counter-intuitive and limiting. The normalization with respect to the whole brain should be optional. In addition, it would be helpful to also allow user to specify the number of voxels/vertices to select, so that the same number of features can be selected across subjects when working with native space data.

### Benefits to the change

More transparent and intuitive behavior. 

### Pseudocode for the new behavior, if applicable

```python
# insert your code below
# I don't actually know how screening_percentile is used downstream so this is just my guess on how it would be implemented

if normalize:
    extent_ratio = (reference_extent / mask_extent)
else:
    extent_ratio = 1

if select_number:
    screening_percentile = screening_number/extent_ratio

screening_percentile = min(screening_percentile, 100.0)
```

Remi-Gau:
most of the logic is in here:

https://github.com/nilearn/nilearn/blob/0bea02cf0fc497bddfc6e6c96df1b09f26b352ba/nilearn/_utils/param_validation.py#L160

Remi-Gau:
> The normalization with respect to the whole brain should be optional.

I think that this would be too big a departure from what we have done, but we could add a flag to either: 
- normalize to the whole brain (current behavior, would stay the default)
- normalize to the mask
- normalize to select X features (

note that when n_features < 50 nilearn will start throwing warnings that you have too few features (https://github.com/nilearn/nilearn/blob/0bea02cf0fc497bddfc6e6c96df1b09f26b352ba/nilearn/decoding/decoder.py#L721)

hyang336:
> > The normalization with respect to the whole brain should be optional.
> 
> I think that this would be too big a departure from what we have done, but we could add a flag to either:
> 
> * normalize to the whole brain (current behavior, would stay the default)
> * normalize to the mask
> * normalize to select X features (
> 
> note that when n_features < 50 nilearn will start throwing warnings that you have too few features (
> 
[...]

PR #5863, opened 2025-11-24
[ENH] allow feature selection to be done by specifying a number of features to select

isha822:
## Description
- This PR retlates to issue #5615 by allowing screening_percentile to accept an absolute number of features (integer > 100) in addition to the existing percentile float (0-100).

Previously, users could only select features based on a percentage of the mask. This change enables selecting a fixed number of voxels (e.g., 500), which is useful when comparing models across masks of different sizes or ensuring a consistent feature count.

## Changes
- Modified check_feature_screening in nilearn/decoding/_utils.py.
- Added logic to return a SelectKBest selector when screening_percentile > 100.
- Updated imports to include SelectKBest.

## Testing
- Verified locally with a test script using a dummy Nifti image.
- Confirmed that passing 20 still creates SelectPercentile.
- Confirmed that passing 500 now successfully creates a SelectKBest(k=500) object.
-


nilearn/plotting/displays/_slicers.py
Remi-Gau, line 1: I think this comes from this PR: https://github.com/nilearn/nilearn/pull/5861

You probably want to remove this change in the current PR

isha822: Fixed! I have reverted the accidental changes to _slicers.py. This PR now only contains the changes for the decoder feature

isha822:
I have cleaned up the branch (removed the accidental file) and fixed the linting errors. All checks are green now. Ready for review!

bthirion:
Sorry that I missed the point in discussion #5615, but we cannot call "screening percentile" a variable that selects a given number of voxels. We need to find another name for that.

isha822:
Ah, I see your point! calling it 'percentile' is confusing since we are selecting an absolute number of voxels.
Would screening_k or screening_amount be better names? Let me know what you prefer and I will update it.

bthirion:
Maybe simply `n_voxels` or `screening_n_voxels`.

isha822:
@bthirion I have updated the code based on your feedback.

nilearn/decoding/_utils.py
```diff
@@ -140,7 +145,9 @@ def adjust_screening_percentile(screening_percentile, mask_img, verbose=0):
 
 @fill_doc
 def check_feature_screening(
-    screening_percentile, mask_img, is_classification, verbose=0
+    screening_percentile,
+    mask_img, is_classification,
```
bthirion:
```suggestion
    mask_img,
    is_classification,
```

nilearn/decoding/_utils.py
```diff
@@ -140,7 +145,9 @@ def adjust_screening_percentile(screening_percentile, mask_img, verbose=0):
...
+    verbose=0,screening_n_voxels=None,
```
bthirion:
```suggestion
    screening_n_voxels=None,
    verbose=0,
```

bthirion: or better, put `screening_n_voxels` just after  `screening_percentile`

nilearn/decoding/decoder.py
```diff
@@ -1143,6 +1146,9 @@ class Decoder(_ClassifierMixin, _BaseDecoder):
         predictions are estimated using default strategy.
 
     %(screening_percentile)s
+    screening_n_voxels : int, optional
+        Number of voxels to select. Default is None.
```
bthirion: We probably have to explain that screening_n_voxels and screening_percentile or mutually exclusive (or that we will always give the priority to screening_percentile if both are provided and raise a warning if screening_n_voxels is specified but ignored)

nilearn/plotting/displays/_slicers.py
```diff
@@ -515,10 +515,12 @@ def _map_show(
                         transparency, transparency_affine
                     )
             except IndexError:
-                # We are cutting outside the indices of the data
-                data_2d = None
-                transparency_2d = None
-
+                raise ValueError(
```
bthirion: The comes from another PR, right ?

nilearn/plotting/image/tests/test_plot_img.py
```diff
@@ -146,3 +147,14 @@ def test_plot_img_transparency_binary_image(
         img_3d_ones_mni,
         transparency=transparency_image,
     )
+
+
+def test_out_of_bounds_error(matplotlib_pyplot, img_3d_mni):
```
bthirion: same here

nilearn/decoding/decoder.py
```diff
@@ -1143,6 +1146,11 @@ class Decoder(_ClassifierMixin, _BaseDecoder):
         predictions are estimated using default strategy.
 
     %(screening_percentile)s
+    screening_n_voxels : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_voxels` takes priority.
```
bthirion: Sorry, I'd rather have screen_percentile to take priority, since this was the preferred method historically.

bthirion: Also we'll need a test to ensure that this behavior is enforced.

bthirion (commented):
Thx ! We're getting close

nilearn/decoding/decoder.py
isha822: Hi @bthirion, thanks for the guidance. I have addressed the feedback and updated the PR with the following changes:
Logic Priority: Modified _utils.py so that screening_percentile takes priority. The code now checks for screening_n_voxels only if screening_percentile is None.
Documentation: Updated the Decoder docstrings to explicitly state that screening_percentile takes priority when both are set.
Testing: Added a new unit test test_screening_priority_logic in test_decoder.py. This test directly verifies that SelectPercentile is initialized (instead of SelectKBest) when both parameters are provided.
I also ran pre-commit to resolve the formatting and linting errors

isha822:
pre-commit.ci autofix

isha822:
Hello! @Remi-Gau  I've reviewed the logs for the failing test check (Test / Test with 3.14 on windows-latest). The failure appears to be an environmental tox error: tox: FATAL: Run flaky tests and those that must be run on a single process. I suspect this is a CI configuration issue where a sequential test is being run in parallel. Could a maintainer please check the tox.ini setup for this environment, or advise which specific test should be excluded from the parallel run? Thanks

doc/conf.py
```diff
@@ -20,6 +20,7 @@
 from sphinx.domains import changeset
 from sphinx.locale import _
 
+# trigger ci
```
Remi-Gau:
```suggestion
```

CITATION.cff
```diff
@@ -811,3 +811,7 @@ authors:
     family-names: Nájera
     website: https://github.com/Titan-C
     affiliation: Checkmk
+  - given-names: Sanjana
+    family-names: Soni
+    website: https://github.com/isha822
+    affiliation: Government Engineering College, Bilaspur, India
```
Remi-Gau:
```suggestion
```

we already got your details from your previous PR 😉

man-shu (approved):
LGTM!

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1538,3 +1538,30 @@ def test_regressor_vs_sklearn(
...
+
+
+def test_screening_priority_logic():
+    # Test that check_feature_screening prefers percentile over n_voxels
+    import nibabel as nib
+    import numpy as np
+    from sklearn.feature_selection import SelectKBest, SelectPercentile
+
+    from nilearn.decoding._utils import check_feature_screening
+
```
hndgzkn: Imports could be placed at the top of the module and only necessary classes or functions could be imported.

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1538,3 +1538,30 @@ def test_regressor_vs_sklearn(
     )
     # also check individual scores are within 1% of each other
     assert np.allclose(scores_sklearn, scores_nilearn, atol=0.01)
+
+
+def test_screening_priority_logic():
+    # Test that check_feature_screening prefers percentile over n_voxels
```
hndgzkn:
```suggestion
    """Test that check_feature_screening prefers percentile over n_voxels."""
```

nilearn/decoding/_utils.py
```diff
@@ -140,7 +145,11 @@ def adjust_screening_percentile(screening_percentile, mask_img, verbose=0):
 
 @fill_doc
 def check_feature_screening(
-    screening_percentile, mask_img, is_classification, verbose=0
+    screening_percentile,
+    mask_img,
+    is_classification,
+    screening_n_voxels=None,
```
hndgzkn: `screening_n_voxels` is missing in docstring.

nilearn/decoding/decoder.py
```diff
@@ -570,6 +570,7 @@ def __init__(
         cv=10,
         param_grid=None,
         screening_percentile=20,
+        screening_n_voxels=None,
```
hndgzkn: `screening_n_voxels` is missing in docstring.

nilearn/decoding/decoder.py
```diff
@@ -1143,6 +1146,11 @@ class Decoder(_ClassifierMixin, _BaseDecoder):
         predictions are estimated using default strategy.
 
     %(screening_percentile)s
+    screening_n_voxels : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_percentile` takes priority.
+        Default is None.
+
```
hndgzkn:
```suggestion
    screening_n_voxels : int, optional, default=None
        Number of voxels to select. If `None, ``screening_percentile` is used.
        If both are set, `screening_percentile` takes priority.

```

nilearn/decoding/decoder.py
```diff
@@ -1312,6 +1322,10 @@ class DecoderRegressor(MultiOutputMixin, _RegressorMixin, _BaseDecoder):
         predictions are estimated using default strategy.
 
     %(screening_percentile)s
+    screening_n_voxels : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_voxels` takes priority.
+        Default is None.
```
hndgzkn:
```suggestion
    screening_n_voxels : int, optional, default=None
        Number of voxels to select. If `None, ``screening_percentile` is used.
        If both are set, `screening_n_voxels` takes priority.
```

nilearn/decoding/decoder.py
```diff
@@ -1476,6 +1493,10 @@ class FREMRegressor(MultiOutputMixin, _RegressorMixin, _BaseDecoder):
         equal to 10.
 
     %(screening_percentile)s
+    screening_n_voxels : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_voxels` takes priority.
+        Default is None.
```
hndgzkn:
```suggestion
    screening_n_voxels : int, optional, default=None
        Number of voxels to select. If `None, ``screening_percentile` is used.
        If both are set, `screening_n_voxels` takes priority.
```

nilearn/decoding/decoder.py
```diff
@@ -1643,6 +1667,10 @@ class FREMClassifier(_ClassifierMixin, _BaseDecoder):
         feature selection based on the Anova F-value for the input data will be
         performed. A float according to a percentile of the highest
         scores.
+    screening_n_voxels : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_voxels` takes priority.
+        Default is None.
```
hndgzkn:
```suggestion
    screening_n_voxels : int, optional, default=None
        Number of voxels to select. If `None, ``screening_percentile` is used.
        If both are set, `screening_n_voxels` takes priority.
```

nilearn/decoding/decoder.py
hndgzkn: Isn't it `screening_percentile` which takes priority when both are set?

nilearn/decoding/decoder.py
hndgzkn: Isn't it `screening_percentile` which takes priority when both are set?

nilearn/decoding/decoder.py
hndgzkn: Isn't it `screening_percentile` which takes priority when both are set?

hndgzkn (approved):
LGTM with some comments and suggestions.

hndgzkn:
That would be good to add an entry in change log as well.

isha822:
pre-commit.ci autofix

isha822:
pre-commit.ci autofix

isha822:
The macOS timeout issues and formatting are now fixed, and all 34 core tests are passing. The remaining failure on Windows (Python 3.14) appears to be from the flaky test suite and is unrelated to these changes. This is ready for a final look/merge whenever you have a moment! @Remi-Gau

bthirion:
Thank a lot @isha822 !
Apparently there one uncovered line:
https://app.codecov.io/gh/nilearn/nilearn/pull/5863?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=nilearn#05818715e6cf801a11fd9f1289e91fe6-R180
Would you mind doing an effort for it ?
Best,

isha822:
pre-commit.ci autofix

doc/changes/0.11.0.rst
```diff
@@ -95,6 +95,7 @@ Fixes
 
 Enhancements
 ------------
+- :bdg-dark:``Code`` Added ``screening_n_voxels`` parameter to :class:``~nilearn.decoding.Decoder`` and :class:``~nilearn.decoding.DecoderRegressor``
```
Remi-Gau: this should be in latest.rst

also I am pretty sure the changes should affect the FREMClassifier and FREMRegressor too

nilearn/glm/tests/test_second_level.py
```diff
@@ -1328,6 +1328,7 @@ def test_second_level_contrast_computation_errors(rng, n_subjects):
         model.compute_contrast(None)
 
 
+@pytest.mark.timeout(300)
```
Remi-Gau:
```suggestion
@pytest.mark.slow
```

nilearn/glm/tests/test_second_level.py
```diff
@@ -1345,6 +1346,7 @@ def test_second_level_t_contrast_length_errors(n_subjects):
         model.compute_contrast(second_level_contrast=[1, 2])
 
 
+@pytest.mark.timeout(300)
```
Remi-Gau:
```suggestion
@pytest.mark.slow
```

let's mark the tests as slow so they are run in separately with no timeout

nilearn/maskers/_mixin.py
```diff
@@ -264,7 +264,7 @@ def region_names_(self) -> dict[int, str]:
 
         sub_df = self.lut_[self.lut_["index"].isin(valid_ids)]
 
-        return sub_df["name"].reset_index(drop=True).to_dict()
+        return sub_df["name"].reset_index(drop=True).to_dict()  # type: ignore[return-value]
```
Remi-Gau: are you sure this silencing of mypy is required?

I think we would see a mypy error pop up on main or some other PR if that were the case.

nilearn/decoding/tests/test_decoder.py
Remi-Gau, line 1: I think we are going to need to add some more high-level tests using the various estimators affected by this : Decoder, DecoderRegresor, FREMClassifier, FREMRegressor.

nilearn/decoding/_utils.py
Remi-Gau: `screening_n_voxels` is too specific for a variable name as we can also work with surfaces that have vertices and not voxels.

I think sklearn estimators have the `n_features_in_` to denote the number of features seen by an estimators seen at fit time: a bit like the `n_elements_` that several nilearn estimators have.

so maybe `screening_n_features` ?

Remi-Gau:
General note for context.
- Original issues #5615 mentioned the possibility to:
  1. have the screening percentile to be relative to a mask passed by the user
  2. have (i) become the new default after the necessary deprecation
  3. select only n features in mask (explicit or implicit)

This PR addresses iii.

doc/changes/0.11.0.rst
Remi-Gau: actually from a brief look at the API I think this feature should also affect the SpaceNet estimators, no?

nilearn/decoding/decoder.py
Remi-Gau: I think we should add this as a template doc string in nilearn/_utils/docs.py and to make it easier to reuse and keep consistent across the different classes

nilearn/_utils/estimator_checks.py
```diff
@@ -2078,6 +2079,16 @@ def check_decoder_with_arrays(estimator_orig):
...
 
 
+def check_decoder_screening_n_voxels(estimator_orig):
+    """Check that only n features are selected."""
+    estimator = clone(estimator_orig)
+    estimator.screening_n_voxels = 100
+    estimator = fit_estimator(estimator)
+
+    n_elements_ = estimator.n_elements_
+    assert n_elements_ == 100
```
Remi-Gau: @isha822 
I added a top level test that should pass but currently does not:
n_elements represents the number of features kept by the estimator at fit time

Remi-Gau: adapted the check to skip searchlight estimators that should not have this behavior

nilearn/decoding/_utils.py
```diff
@@ -167,6 +176,9 @@ def check_feature_screening(
     """
     f_test = f_classif if is_classification else f_regression
 
+    if screening_percentile is None and screening_n_voxels is not None:
+        return SelectKBest(f_test, k=screening_n_voxels)
```
Remi-Gau: SelectKBest will do some input validation to make sure that screening_n_voxels is not < 0, but I wonder if we should not also do some validation to make sure it is not higher than the number of potential features in the mask_img

nilearn/decoding/decoder.py
Remi-Gau: that is my reading of the code as well: the tests actually show this

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1539,3 +1541,49 @@ def test_regressor_vs_sklearn(
...
+    assert not isinstance(selector, SelectKBest)
+
+
+def test_check_feature_screening_n_voxels_only():
+    """Test that screening_n_voxels works when percentile is None."""
+    import nibabel as nib
+    import numpy as np
+    from sklearn.feature_selection import SelectKBest
+
+    from nilearn.decoding.decoder import check_feature_screening
```
Remi-Gau: those imports should be at the top of the module

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1539,3 +1541,49 @@ def test_regressor_vs_sklearn(
...
+        mask_img=mask,
+        is_classification=True,
+        screening_n_voxels=50,
+    )
+
+    # 3. VERIFY: We should get a SelectPercentile object
+    # (meaning percentile won)
+    # If logic is wrong, this will be SelectKBest and the test will fail.
+    assert isinstance(selector, SelectPercentile)
+    assert not isinstance(selector, SelectKBest)
```
Remi-Gau:
```suggestion
def test_screening_priority_logic(img_3d_rand_eye):
    """Test that check_feature_screening prefers percentile over n_voxels."""
    # Call the function with BOTH options (Conflict!)
    # percentile=10, n_voxels=50
    selector = check_feature_screening(
        screening_percentile=10,
        mask_img=img_3d_rand_eye,
        is_classification=True,
        screening_n_voxels=50,
    )

    # VERIFY: We should get a SelectPercentile object
    # (meaning percentile won)
    # If logic is wrong, this will be SelectKBest and the test will fail.
    assert isinstance(selector, SelectPercentile)
    assert not isinstance(selector, SelectKBest)
```

let's use some of our fixtures

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1539,3 +1541,49 @@ def test_regressor_vs_sklearn(
...
+    selector = check_feature_screening(
+        screening_percentile=None,
+        mask_img=mask,
+        is_classification=True,
+        screening_n_voxels=15,
+    )
+
+    # 3. Verify it returned a SelectKBest object with the right 'k'
+    assert isinstance(selector, SelectKBest)
+    assert selector.k == 15
```
Remi-Gau:
```suggestion
def test_check_feature_screening_n_voxels_only(img_3d_rand_eye):
    """Test that screening_n_voxels works when percentile is None."""
    import nibabel as nib
    import numpy as np
    from sklearn.feature_selection import SelectKBest

    from nilearn.decoding.decoder import check_feature_screening

    # Trigger the red line: percentile is None, n_voxels is set
    selector = check_feature_screening(
        screening_percentile=None,
        mask_img=img_3d_rand_eye,
        is_classification=True,
        screening_n_voxels=15,
    )

    # Verify it returned a SelectKBest object with the right 'k'
    assert isinstance(selector, SelectKBest)
    assert selector.k == 15
```

use fixture

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1539,3 +1541,45 @@ def test_regressor_vs_sklearn(
...
+        mask_img=mask,
+        is_classification=True,
+        screening_n_voxels=50,
+    )
+
+    # 3. VERIFY: We should get a SelectPercentile object
+    # (meaning percentile won)
+    # If logic is wrong, this will be SelectKBest and the test will fail.
+    assert isinstance(selector, SelectPercentile)
+    assert not isinstance(selector, SelectKBest)
```
Remi-Gau:
```suggestion
def test_screening_priority_logic(img_3d_rand_eye):
    """Test that check_feature_screening prefers percentile over n_voxels."""
    # Call the function with BOTH options (Conflict!)
    # percentile=10, n_voxels=50
    selector = check_feature_screening(
        screening_percentile=10,
        mask_img=img_3d_rand_eye,
        is_classification=True,
        screening_n_voxels=50,
    )

    # VERIFY: We should get a SelectPercentile object
    # (meaning percentile won)
    # If logic is wrong, this will be SelectKBest and the test will fail.
    assert isinstance(selector, SelectPercentile)
    assert not isinstance(selector, SelectKBest)
```

isha822:
pre-commit.ci autofix

isha822:
pre-commit.ci autofix

isha822:
@Remi-Gau I have resolved the merge conflicts and fixed the rst documentation formatting. The remaining failures (macOS/pre-release) seem to be unrelated flaky tests. Could you please review? Lemme know I'll fix further if required... Thankyou

Remi-Gau:
> @Remi-Gau I have resolved the merge conflicts and fixed the rst documentation formatting. The remaining failures (macOS/pre-release) seem to be unrelated flaky tests. Could you please review? Lemme know I'll fix further if required... Thankyou

There still are some tests failing that are not about test flakyness: 
https://github.com/nilearn/nilearn/actions/runs/20487589108/job/58873138330?pr=5863#step:8:16472

These are related to the check I added: see here https://github.com/nilearn/nilearn/pull/5863#discussion_r2644057946

isha822:
pre-commit.ci autofix

nilearn/decoding/_utils.py
```diff
@@ -167,6 +176,19 @@ def check_feature_screening(
     """
     f_test = f_classif if is_classification else f_regression
 
+    if screening_percentile is None and screening_n_features is not None:
+        if mask_img is not None:
+            from nilearn.image import get_data
+
+            n_features_in_mask = np.sum(get_data(mask_img) != 0)
```
hndgzkn: Could it be sufficient to use `_get_data` instead of `get_data` ?

If not, is it possible to move import to the top of the file?

nilearn/decoding/decoder.py
```diff
@@ -570,6 +570,7 @@ def __init__(
         cv=10,
         param_grid=None,
         screening_percentile=20,
+        screening_n_features=None,
```
hndgzkn, line 598: `screening_n_features` is missing in `Parameters` section of base class docstring.

nilearn/decoding/decoder.py
```diff
@@ -1312,6 +1322,10 @@ class DecoderRegressor(MultiOutputMixin, _RegressorMixin, _BaseDecoder):
         predictions are estimated using default strategy.
 
     %(screening_percentile)s
+    screening_n_features : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_features` takes priority.
+        Default is None.
```
hndgzkn:
```suggestion
        If both are set, `screening_percentile` takes priority.
        Default is None.
```

nilearn/decoding/decoder.py
```diff
@@ -1476,6 +1493,10 @@ class FREMRegressor(MultiOutputMixin, _RegressorMixin, _BaseDecoder):
         equal to 10.
 
     %(screening_percentile)s
+    screening_n_features : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_features` takes priority.
+        Default is None.
```
hndgzkn:
```suggestion
        If both are set, `screening_percentile` takes priority.
        Default is None.
```

nilearn/decoding/decoder.py
```diff
@@ -1643,6 +1667,10 @@ class FREMClassifier(_ClassifierMixin, _BaseDecoder):
         feature selection based on the Anova F-value for the input data will be
         performed. A float according to a percentile of the highest
         scores.
+    screening_n_features : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_features` takes priority.
+        Default is None.
```
hndgzkn:
```suggestion
        If both are set, `screening_percentile` takes priority.
        Default is None.
```

isha822:
pre-commit.ci autofix

isha822:
pre-commit.ci autofix

isha822:
pre-commit.ci autofix

isha822:
pre-commit.ci autofix

isha822:
pre-commit.ci autofix

isha822:
@Remi-Gau  I investigated the check_decoder_screening_n_features failures.
It seems n_elements_ on the Decoder object reports the mask size (input features), not the number of features selected during the internal CV screening.
My dummy mask has 125 voxels.
I requested screening_n_features=10.
estimator.n_elements_ remains 125 after fit.
Since n_elements_ doesn't reflect the internal screening, I have updated the test to skip Decoder and FREM models for now to get the CI green. We might need a different way to verify screening (e.g. checking internal attributes) if we want this test to be active.

bthirion:
> @Remi-Gau I investigated the check_decoder_screening_n_features failures. It seems n_elements_ on the Decoder object reports the mask size (input features), not the number of features selected during the internal CV screening. My dummy mask has 125 voxels. I requested screening_n_features=10. estimator.n_elements_ remains 125 after fit. Since n_elements_ doesn't reflect the internal screening, I have updated the test to skip Decoder and FREM models for now to get the CI green. We might need a different way to verify screening (e.g. checking internal attributes) if we want this test to be active.

Indeed the docstrings says it's supposed to be the #voxels in the mask.

doc/changes/0.13.0.rst
```diff
@@ -74,6 +74,7 @@ Fixes
 
 Enhancements
 ------------
+- :bdg-dark:`Code` Added ``screening_n_features`` parameter to :class:`~nilearn.decoding.Decoder`,  :class:`~nilearn.decoding.DecoderRegressor`, :class:`~nilearn.decoding.FREMClassifier`,  and :class:`~nilearn.decoding.FREMRegressor`.
```
Remi-Gau: could you move this to doc/changes/latest.rst ?

Remi-Gau:
having a look at this PR, I realize we have an extra security to avoid selecting too few features: if the extracted data has less than 100 fetaures we will not do any feature selection.

```python
do_screening = (X_train.shape[1] > 100) and selector is not None
```

https://github.com/nilearn/nilearn/blob/70e1eb214f4aa58c5917d69a2f1f5a0e88849af1/nilearn/decoding/decoder.py#L383

Whatever we do, this magic number of 100 should stored as clear constant somewhere ideally with a comment justifying why this value was chosen.

I would also prefer to have things a bit more consistent: we throw a warning if we are decoding with less than 50 features before selection: I'd be tempted to use the same cutoff for both behavior.

https://github.com/nilearn/nilearn/blob/70e1eb214f4aa58c5917d69a2f1f5a0e88849af1/nilearn/decoding/decoder.py#L704

In any case within the context of this raises some issues:
- should we keep this lower limit? 
  - if yes we should:
    - check that the screening_n_features passed as input is not inferior to that number
    - if no screening will be performed we should probably tell the user with a warning
  - if not:
    - should it be removed or just lowered and if so to what value?

nilearn/_utils/estimator_checks.py
```diff
@@ -2078,6 +2079,58 @@ def check_decoder_with_arrays(estimator_orig):
...
+        return
+
+    # 2. Skip Decoder / DecoderRegressor
+    # The Decoder's 'n_elements_' attribute
+    #  tracks the Input Mask Size (125 voxels),
+    # NOT the features selected inside the
+    # internal cross-validation loop (10 features).
+    # Therefore, asserting n_elements_ == 10 will always fail.
+    if "Decoder" in class_name:
+        return
```
Remi-Gau: yes that's true but we should still have some trace somewhere of how many features were selected.

If I check the model weights in the end with the following code I get 28 instead of the 10 I would expect (see code below)

Note my expectation could be wrong, but as a user if I am provided with a way to select n features I would expect only 10 non-zero weights.


```

import numpy as np
import nibabel as nib

decoder = Decoder(standardize="zscore_sample", screening_percentile=None, screening_n_features=10)

X = rng.rand(5, 5, 5, 20)
X_img = nib.Nifti1Image(X, np.eye(4))
y = np.array([0] * 10 + [1] * 10)

decoder.fit(X_img, y)

print(np.sum(decoder.coef_ != 0))
```

nilearn/decoding/_utils.py
```diff
@@ -167,6 +177,17 @@ def check_feature_screening(
     """
     f_test = f_classif if is_classification else f_regression
 
+    if screening_percentile is None and screening_n_features is not None:
+        if mask_img is not None:
+            n_features_in_mask = np.sum(get_data(mask_img) != 0)
```
Remi-Gau: note that this won't work with surface data

Remi-Gau:
> We might need a different way to verify screening (e.g. checking internal attributes) if we want this test to be active.

agreed

sorry quickly written check was misleading: thanks for investigating

isha822:
@Remi-Gau Thanks for the detailed review!
1. Regarding the "28 vs 10" weights issue:
This behavior is expected because Decoder defaults to cv=10.
Since the dummy data is random noise, the "Top 10" features selected inside each fold will be different (unstable). The final coef_ attribute is the average of the models across all folds.
Therefore, the non-zero weights represent the union of features selected across all 10 folds.
The fact that you got 28 (instead of 125) confirms that screening is working!
2. Regarding Surface Data:
Good catch on get_data(mask_img). Is there a preferred Nilearn utility to count "features in a mask" that handles both Volume and Surface? Or should I simply wrap this in a try/except block?
3. Changelog & Constants:
I have extracted the magic number 100 into a constant _MIN_N_FEATURES_SCREENING.
Question: You mentioned moving the changelog to doc/changes/latest.rst, but that file does not exist in my tree (I only see 0.13.0.rst and whats_new.rst). Should I create latest.rst and include it in whats_new.rst, or should I just add the entry to the top of whats_new.rst?

Remi-Gau:
> 3\. Question: You mentioned moving the changelog to doc/changes/latest.rst, but that file does not exist in my tree (I only see 0.13.0.rst and whats_new.rst). Should I create latest.rst and include it in whats_new.rst, or should I just add the entry to the top of whats_new.rst?

fixed: the latest.rst was on main but not on your branch

Remi-Gau:
> 1. Regarding the "28 vs 10" weights issue:
>        This behavior is expected because Decoder defaults to cv=10.
>        Since the dummy data is random noise, the "Top 10" features selected inside each fold will be different (unstable). The final coef_ attribute is the average of the models across all folds.
>        Therefore, the non-zero weights represent the union of features selected across all 10 folds.
>        The fact that you got 28 (instead of 125) confirms that screening is working!

I agree with the diagnostic of why we are getting this result, but I would say **some** screening is happening but not exactly that the **expected** screening is happening.

If we decide to keep things this way then se should be talking about 'screening_min_n_features' rather than about 'screening_n_features' because that parameter will only set the lower bound on the number of features to select.

What is potentially annoying with this is that this may also result on having different number of features for each subject in group analyis.

@bthirion @emdupre @jeromedockes any opinion on this

bthirion:
> > 1. Regarding the "28 vs 10" weights issue:
> >    This behavior is expected because Decoder defaults to cv=10.
> >    Since the dummy data is random noise, the "Top 10" features selected inside each fold will be different (unstable). The final coef_ attribute is the average of the models across all folds.
> >    Therefore, the non-zero weights represent the union of features selected across all 10 folds.
> >    The fact that you got 28 (instead of 125) confirms that screening is working!
> 
> I agree with the diagnostic of why we are getting this result, but I would say **some** screening is happening but not exactly that the **expected** screening is happening.
> 
> If we decide to keep things this way then se should be talking about 'screening_min_n_features' rather than about 'screening_n_features' because that parameter will only set the lower bound on the number of features to select.
> 
> What is potentially annoying with this is that this may also result on having different number of features for each subject in group analyis.
> 
> @bthirion @emdupre @jeromedockes any opinion on this

Regarding the point on screening, to me it's a matter of documentation: we need to be explicit about what happens. Maybe the simplest way is to show that concretely in an example. I'm slightly against parameter renaming.

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1540,3 +1542,39 @@ def test_regressor_vs_sklearn(
     )
     # also check individual scores are within 1% of each other
     assert np.allclose(scores_sklearn, scores_nilearn, atol=0.01)
+
+
+def test_screening_priority_logic():
```
Remi-Gau: use the `img_3d_rand_eye` like in the test below

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1540,3 +1542,39 @@ def test_regressor_vs_sklearn(
...
+    """Test that check_feature_screening prefers percentile over n_voxels."""
+    # 1. Create a simple dummy mask
+    mask = nib.Nifti1Image(np.zeros((10, 10, 10)), np.eye(4))
+
+    # 2. Call the function with BOTH options (Conflict!)
+    # percentile=10, n_voxels=50
```
Remi-Gau: try to move all comments in the doc string rather than just before each step

nilearn/decoding/_utils.py
Remi-Gau: you can use use the `get_data` function from nilearn.surface.surface to do the same as 

```
n_features_in_mask = np.sum(get_data(mask_img) != 0)
```

for surface data

so something like

```python
# import at top of the module
from nilearn.surface.surface import get_data as get_sruface_data

if mask_img is not None:
  if isinstance(mask_img, SurfaceImage):
    data = get_surface_data(mask_img)
  else:
    data = get_data(mask_img)
  n_features_in_mask = np.sum(data != 0)
```

nilearn/decoding/decoder.py
```diff
@@ -514,6 +514,10 @@ class _BaseDecoder(CacheMixin, BaseEstimator):
         For classification, valid entries are: 'accuracy', 'f1', 'precision',
         'recall' or 'roc_auc'. Defaults to 'roc_auc'.
 
+    screening_n_features : int, optional
+        The number of features to keep. Default=None.
+        If both `screening_percentile` and `screening_n_features` are set,
+        `screening_percentile` takes priority.
```
Remi-Gau:
```suggestion
    %(screening_n_features)s
```

and add an entry in nilearn/_utils/docs.py

with something like

```python
docdict["screening_n_features"] = """
screening_n_features : :obj:`int`, default=None
        The number of features to keep.
        If both `screening_percentile` and `screening_n_features` are set,
        `screening_percentile` takes priority.
```

then you can reuse the same docstrings for all the other classes

Remi-Gau: also the doc string should be very clear that this is the number of feature selected PER CV and that therefore the the final number of feature weight can be superior to this value

nilearn/decoding/decoder.py
```diff
@@ -1143,6 +1150,11 @@ class Decoder(_ClassifierMixin, _BaseDecoder):
         predictions are estimated using default strategy.
 
     %(screening_percentile)s
+    screening_n_features : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_percentile` takes priority.
+        Default is None.
+
```
Remi-Gau:
```suggestion

%(screening_n_features)s

```

nilearn/decoding/decoder.py
```diff
@@ -1312,6 +1326,10 @@ class DecoderRegressor(MultiOutputMixin, _RegressorMixin, _BaseDecoder):
         predictions are estimated using default strategy.
 
     %(screening_percentile)s
+    screening_n_features : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_percentile` takes priority.
+        Default is None.
```
Remi-Gau:
```suggestion

%(screening_n_features)s

```

nilearn/decoding/decoder.py
```diff
@@ -1475,6 +1496,10 @@ class FREMRegressor(MultiOutputMixin, _RegressorMixin, _BaseDecoder):
         equal to 10.
 
     %(screening_percentile)s
+    screening_n_features : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_features` takes priority.
+        Default is None.
```
Remi-Gau:
```suggestion

%(screening_n_features)s

```

nilearn/decoding/decoder.py
```diff
@@ -1642,6 +1670,10 @@ class FREMClassifier(_ClassifierMixin, _BaseDecoder):
         feature selection based on the Anova F-value for the input data will be
         performed. A float according to a percentile of the highest
         scores.
+    screening_n_features : int, optional
+        Number of voxels to select. If `None, ``screening_percentile` is used.
+        If both are set, `screening_n_features` takes priority.
+        Default is None.
```
Remi-Gau:
```suggestion

%(screening_n_features)s

```

nilearn/_utils/estimator_checks.py
```diff
@@ -2078,6 +2079,58 @@ def check_decoder_with_arrays(estimator_orig):
...
+
+    if "clustering_percentile" in params:
+        params["clustering_percentile"] = 100
+    if "cv" in params:
+        params["cv"] = None
+
+    estimator = estimator_orig.__class__(**params)
+    estimator.fit(X_img, y)
+
+    assert estimator.n_elements_ == 10
```
Remi-Gau: I think that if update this test to run a single cross-validation then the following should be true, no?

```suggestion
    assert np.sum(estimator.coef_ != 0) == 10
```

Remi-Gau:
> 3\. I have extracted the magic number 100 into a constant _MIN_N_FEATURES_SCREENING.

maybe I missed that but where was that done?

nilearn/_utils/estimator_checks.py
```diff
@@ -2282,6 +2288,44 @@ def check_decoder_with_arrays(estimator_orig):
...
+        estimator.cv = 2
+
+    estimator = fit_estimator(estimator)
+
+    n_non_zero_weights = np.sum(estimator.coef_ != 0)
+
+    if isinstance(estimator, (Decoder, DecoderRegressor)):
+        assert n_non_zero_weights >= screening_n_features
+    else:
+        assert n_non_zero_weights == screening_n_features
```
Remi-Gau, line 2298: rewrote this check so it does not skip all our estimators and checks something kinda sensible

nilearn/decoding/decoder.py
```diff
@@ -1144,6 +1149,9 @@ class Decoder(_ClassifierMixin, _BaseDecoder):
 
     %(screening_percentile)s
 
+%(screening_n_features)s
```
Remi-Gau:
```suggestion
    %(screening_n_features)s
```

nilearn/decoding/decoder.py
```diff
@@ -1313,6 +1323,9 @@ class DecoderRegressor(MultiOutputMixin, _RegressorMixin, _BaseDecoder):
...
```
Remi-Gau:
```suggestion
    %(screening_n_features)s
```

nilearn/decoding/decoder.py
```diff
@@ -1476,6 +1492,9 @@ class FREMRegressor(MultiOutputMixin, _RegressorMixin, _BaseDecoder):
...
```
Remi-Gau:
```suggestion
    %(screening_n_features)s
```

nilearn/decoding/decoder.py
```diff
@@ -1643,6 +1665,9 @@ class FREMClassifier(_ClassifierMixin, _BaseDecoder):
         performed. A float according to a percentile of the highest
         scores.
 
+%(screening_n_features)s
```
Remi-Gau:
```suggestion
    %(screening_n_features)s
```

isha822:
Thanks for the fixes @Remi-Gau ! I see the latest run failed, but looking at the logs, it appears to be a Timeout in nilearn/plotting/tests/test_baseline_comparisons.py, which seems unrelated to the decoder changes.

Remi-Gau:
@isha822 
moving to draft until the 0.13.1 is done
after that we should be able to finalize this and merge it

isha822:
Thanks for the update, @Remi-Gau ! Understood completely regarding the 0.13.1 release. I'm happy to wait until then to finalize this.
Looking forward to getting this merged and contributing more to the project!

nilearn/decoding/tests/test_decoder.py
Remi-Gau: will fix that for you

nilearn/decoding/tests/test_decoder.py
Remi-Gau: done via estimator_checks

nilearn/decoding/_utils.py
Remi-Gau: will fix that

nilearn/_utils/docs.py
```diff
@@ -1038,6 +1038,13 @@ def custom_function(vertices):
             may be included even for very small ``screening_percentile``.
 
 """
+docdict[
+    "screening_n_features"
+] = """screening_n_features : :obj:int, default=None
+    The number of features to keep.
+    If both ``screening_percentile`` and ``screening_n_features`` are set,
+    ``screening_percentile`` takes priority.
+    """
```
Remi-Gau, line 1055: TODO
I think we should improve the doc string a bit to make it clear the screening_n_features applies to a single CV and that hence in total model can have more than screening_n_features non-zero weights

nilearn/decoding/tests/test_utils.py
```diff
@@ -27,7 +29,7 @@ def test_feature_screening(
 ):
     """Check that screening percentile is correctly adjusted.
 
-    For very small ROIs, all elements should be included.
+    For very small ROIs, all elements should be i_img_3d_randncluded.
```
Remi-Gau:
```suggestion
    For very small ROIs, all elements should be included.
```

examples/02_decoding/plot_haxby_anova_svm.py
```diff
@@ -62,7 +62,8 @@
     mask=mask_img,
     smoothing_fwhm=4,
     standardize="zscore_sample",
-    screening_percentile=5,
+    screening_percentile=None,
+    screening_n_features=500,
```
Remi-Gau: not sure if this is the best example where to add this but TBH I am not sure which example is the best to advertise this

bthirion: This sounds reasonable to me.

Remi-Gau:
those tests failures are unrelated: https://github.com/nilearn/nilearn/actions/runs/21875903276/job/63144656498?pr=5863#step:8:13313

bthirion:
> those tests failures are unrelated: https://github.com/nilearn/nilearn/actions/runs/21875903276/job/63144656498?pr=5863#step:8:13313

Should we merge then ?

Remi-Gau:
> Should we merge then ?

The doc string can still be improved I think to avoid users having wrong expectations: https://github.com/nilearn/nilearn/pull/5863#discussion_r2789363336

nilearn/_utils/docs.py
```diff
@@ -1038,6 +1038,21 @@ def custom_function(vertices):
...
+    The number of features to keep for a single cross-validation.
+    If both ``screening_percentile`` and ``screening_n_features`` are set,
+    ``screening_percentile`` takes priority.
+
+    .. admonition:: Important
+
+        Given ``screening_n_features``
+        is the number of features kept **for each fold** of a cross-validation,
+        the final modal can have
+        more than ``screening_n_features`` non-zero weights.
```
Remi-Gau, line 1053: hope it is obvious enough

Remi-Gau: @bthirion 
does that make sense?

nilearn/_utils/docs.py
```diff
@@ -1038,6 +1038,21 @@ def custom_function(vertices):
...
+] = """screening_n_features : :obj:`int`, default=None
+    The number of features to keep for a single cross-validation.
+    If both ``screening_percentile`` and ``screening_n_features`` are set,
+    ``screening_percentile`` takes priority.
+
+    .. admonition:: Important
+
+        Given ``screening_n_features``
+        is the number of features kept **for each fold** of a cross-validation,
+        the final modal can have
```
```suggestion
        the final model can have
```

Remi-Gau:
thanks @isha822 for your patience with this one

isha822:
Thank you so much for the merge and for guiding me through it, @Remi-Gau @bthirion ! It was a great learning experience.

PR #5865, opened 2025-11-25
[PERF] use image slicer before getting data

### LINKED ISSUE #5753: [PERF] _index_img underperforms for large images
`_index_img` is currently implemented as such:
https://github.com/nilearn/nilearn/blob/273d531daa81fb0e4f81e4599f65d89532355883/nilearn/_utils/niimg_conversions.py#L95-L99

Notice that we first load the entire dataobj in memory with `_get_data` before doing the slicing. This becomes quite suboptimal whenever the 4th dimension gets large (1000 and more).

A more efficient implementation would first slice the image object then do the loading with `_get_data`:
```{python}
    return new_img_like(img, _get_data(img.slicer[..., index]), img.affine)
```

Should I PR ?

Best,
PL

Remi-Gau:
- Closes #5753

- in index_img for nifti image, use image slicer before getting data

### DIFF
--- a/nilearn/image/tests/test_image.py
+++ b/nilearn/image/tests/test_image.py
@@ -696,7 +696,7 @@ def test_index_img_surface(surf_img_2d, length, index, expected_n_samples):
         assert_array_equal(value, expected_data_3d[hemi])
 
 
-def test_index_img_error_volumne_4d(affine_eye):
+def test_index_img_error_volume_4d(affine_eye):
     """Test impossible indices with volume data."""
     img_4d, _ = generate_fake_fmri(affine=affine_eye)
     fourth_dim_size = img_4d.shape[3]


man-shu:
Ok, finally got to this. I compared two commits as follows:

Run the benchmarks on each commit
```bash
asv run ff784094a302b30b23bbbe09eaadc563537b9086^! -b index_img
```

```bash
asv run cf69633ae29a56c422e44ee5c9674720e63b1d86^! -b index_img
```

Then, to compare the two in a tabular view:
```bash
asv compare ff784094a302b30b23bbbe09eaadc563537b9086 cf69633ae29a56c422e44ee5c9674720e63b1d86
```

```bash
| Change   | Before [ff784094] <main>   | After [cf69633a] <perf/index_img>   |   Ratio | Benchmark (Parameter)                           |
|----------|----------------------------|-------------------------------------|---------|-------------------------------------------------|
|          | 8.36±0.1s                  | 8.17±0.08s                          |    0.98 | tracking.image.IndexImgBenchmark.time_index_img |
```
The new implementation does seem a bit faster in this comparison, but I ran it multiple times, and that's not always the case. It could just be because the input 4D image is small (only 10 time points). I will try using a longer image and see if the comparisons reveal something more stable.

Remi-Gau:
> because the input 4D image is small (only 10 time points)

yeah I think we can defo use something much bigger there

man-shu:
With 50 images:

```bash
| Change   | Before [ff784094] <main>   | After [cf69633a] <perf/index_img~1>   |   Ratio | Benchmark (Parameter)                           |
|----------|----------------------------|---------------------------------------|---------|-------------------------------------------------|
|          | 1.15±0.01m                 | 1.18±0.04m                            |    1.03 | tracking.image.IndexImgBenchmark.time_index_img
```

Trying to see if I can do 500 locally...

asv_benchmarks/benchmarks/tracking/image.py
```diff
@@ -35,3 +35,15 @@ def peakmem_mean_img(self):
...
         img = load("nilearn")[1]
         mean_img(img)
+
+
+class IndexImgBenchmark(Benchmark):
+    """A benchmark that compares the performance of indexing a 4D image."""
+
+    def time_index_img(self):
+        """Time the loading the image followed by indexing a voxel."""
+        index_img("fmri_10.nii.gz", slice(1, 40, 2))
```
Remi-Gau: I think we need to index more images to be able to see a performance difference.
Also we need to make sure that the index is not fancy

https://nipy.org/nibabel/reference/nibabel.fileslice.html#nibabel.fileslice.is_fancy

Remi-Gau:
OK this PR makes me realize we have got 0 test to test index_img with surface data: will add some in a separate PR

asv_benchmarks/benchmarks/tracking/image.py
man-shu: Ok that's better indeed (with 50 images):

```bash
| Change   | Before [ff784094] <main>   | After [0edecb05] <perf/index_img>   | Ratio   | Benchmark (Parameter)                           |
|----------|----------------------------|-------------------------------------|---------|-------------------------------------------------|
|          | 1.47±0.06m                 | 1.20±0.02m                          | ~0.81   | tracking.image.IndexImgBenchmark.time_index_img |
```

man-shu: I also let it run for 100 images, not as big a difference, but it's there. Shall we merge it then, @Remi-Gau ?

```bash
| Change   | Before [ff784094] <main~11>   | After [24972e5b] <perf/index_img>   |   Ratio | Benchmark (Parameter)                           |
|----------|-------------------------------|-------------------------------------|---------|-------------------------------------------------|
|          | 2.56±0.01m                    | 2.49±0.1m                           |    0.97 | tracking.image.IndexImgBenchmark.time_index_img |
```

Remi-Gau: yup though there is one test to fix: https://github.com/nilearn/nilearn/actions/runs/20102027887/job/57675642610?pr=5865#step:8:12426

Remi-Gau:
ok the error is unrelated: merging

PR #5866, opened 2025-11-26
[FIX] bump versions used for kaleido and plotly

### LINKED ISSUE #5801: Cannot save plotly figures on windows
See https://github.com/nilearn/nilearn/actions/runs/18870740269/job/53848281405#step:8:14626

```
nilearn/plotting/surface/tests/test_plotly_surface_figure.py::test_plotly_savefig - ValueError: Transform failed with error code 1: Failed to serialize documen...
```

Related to https://github.com/plotly/Kaleido/issues/402

Remi-Gau:
- Closes #5801 

- [x] changelog
- [x] use same versions of plotly and kaleido used for all OS
- [x] try updating chrome if test fails (see https://github.com/plotly/Kaleido/issues/402#issuecomment-4064296132)
  - [x] would require either updating the doc or the code to let users update their chrome
- [x] reduce number of figures updated

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -76,6 +76,7 @@ jobs:
 
                 if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
                     versions='py=["3.14", "3.10"]'
+                    versions='py=["3.14", "3.13", "3.12", "3.11", "3.10"]'
                 fi
 
                 echo "Will run on the following python: ${versions}"
--- a/tox.ini
+++ b/tox.ini
@@ -50,8 +50,14 @@ description = Run tests on latest version of all dependencies (plotting not incl
 passenv = {[global_var]passenv}
 dependency_groups = test
 commands =
-    pytest -m "not slow and not flaky and not single_process" -W error::sklearn.exceptions.ConvergenceWarning --timeout=60 --cov=nilearn --cov-report=xml --cov-report=html --cov-append --report=report.html --numprocesses auto --csv results/pytest_output/pytest_output.csv  {posargs:}
-    pytest -m "slow and not flaky and not single_process" -W error::sklearn.exceptions.ConvergenceWarning --cov=nilearn --cov-report=xml --cov-report=html --cov-append --report=report_slow.html --numprocesses auto --csv results/pytest_output/pytest_output_slow_tests.csv  {posargs:}
+; run tests=
+; - in parallel
+; - with code coverage
+; - generating report and tsv of tests statistics
+; - sklearn.exceptions.ConvergenceWarning and FutureWarning raise an error: so that they are dealt with they appear in our dependencies
+    ; run tests with a timeout
+    pytest -m "not slow and not flaky and not single_process" -W error::sklearn.exceptions.ConvergenceWarning -W error::FutureWarning --timeout=60 --cov=nilearn --cov-report=xml --cov-report=html --cov-append --report=report.html --numprocesses auto --csv results/pytest_output/pytest_output.csv  {posargs:}
+    pytest -m "slow and not flaky and not single_process" -W error::sklearn.exceptions.ConvergenceWarning -W error::FutureWarning --cov=nilearn --cov-report=xml --cov-report=html --cov-append --report=report_slow.html --numprocesses auto --csv results/pytest_output/pytest_output_slow_tests.csv  {posargs:}
 
 
 [testenv:flaky]
--- a/doc/maintenance.rst
+++ b/doc/maintenance.rst
@@ -129,6 +129,26 @@ We welcome pull requests from all community members, if they follow the
 details on their process are available
 :sklearn:`here <developers/contributing.html#contributing-code-and-documentation>`).
 
+Deprecations
+============
+
+Use :class:`FutureWarning` for deprecations that users should know about
+on which they can have some control,
+for example by making sure that the warning message mentions
+what action to take to prevent the warning to be thrown.
+If there is no way for the user to prevent the warning to be thrown,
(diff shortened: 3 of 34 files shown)


.github/workflows/test_with_tox.yml
```diff
@@ -163,6 +163,7 @@ jobs:
             run: |
                 python build_tools/github/restrict_tests_to_run.py
                 tests_to_run=$(cat tests_to_run.txt)
+                tests_to_run='nilearn/plotting/surface/tests/test_plotly_surface_figure.py'
```
bthirion: Is it just for testing ?

Remi-Gau: yes just running one test

.github/workflows/test_with_tox.yml
```diff
@@ -150,6 +150,7 @@ jobs:
...
```
```suggestion
```

tox.ini
```diff
@@ -71,7 +71,10 @@ extras = plotting
 dependency_groups =
     test
     rich
+allowlist_externals =
+    kaleido_get_chrome
 commands =
+    kaleido_get_chrome
```
Remi-Gau: note that this would mean that when they run our tests with plotting figures, contributors install / update their version of chrome

if we do not do it in tox then we have to make it very explicit in our contributing that some tests may fail because of that

and also update CI to install chrome

hndgzkn: > note that this would mean that when they run our tests with plotting figures, contributors install / update their version of chrome

I would suggest to leave the decision to install chrome to users.  

> if we do not do it in tox then we have to make it very explicit in our contributing that some tests may fail because of that

I'm more in favor of this solution.

> and also update CI to install chrome

That might not be necessary as [Ubuntu](https://github.com/actions/runner-images/blob/ubuntu24/20250615.1/images/ubuntu/Ubuntu2404-Readme.md#browsers-and-drivers) already has Chrome installed.

hndgzkn: Also there were changes made in #6004 so that the correct error message is printed when chrome is missing. 
Now that incompatible version problem no more exists, some parts of added code in this PR can be removed.

hndgzkn: Explicitly:

- The below function can be removed:

https://github.com/nilearn/nilearn/blob/d135e0d7e2ba975aac7f731315b8fadb21b8801c/nilearn/_utils/versions.py#L63-L79

- below part can be removed:
https://github.com/nilearn/nilearn/blob/d135e0d7e2ba975aac7f731315b8fadb21b8801c/nilearn/plotting/displays/_figures.py#L170-L177

Remi-Gau: thanks @hndgzkn 
I would have completely overlooked those things

.github/workflows/test_with_tox.yml
```diff
@@ -91,7 +91,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```
revert before merging

.github/workflows/testing_minimum.yml
```diff
@@ -12,6 +12,8 @@
 name: test minimal version of each dependency
 
 on:
+    pull_request:
+
```
```suggestion
```

revert before merging

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -113,11 +113,29 @@
...
+# as we would do with a matplotlib figure.
+#
+# .. admonition:: Google Chrome needed
+#
+#     To be able to save images with plotly,
+#     make sure that Google Chrome is installed!
+#     You can install a compatible Chrome version using
+#     the ``kaleido_get_chrome`` command in command line or
+#     ``kaleido.get_chrome_sync()`` function
+#     in Python:
```
Remi-Gau, line 126: Should something like this also be added to the changelog?
@hndgzkn

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -113,11 +113,29 @@
...
+#
+#           import kaleido
+#           kaleido.get_chrome_sync()
+#
+from pathlib import Path
+
+output_dir = Path.cwd() / "results" / "plot_3d_map_to_surface_projection"
+output_dir.mkdir(exist_ok=True, parents=True)
+print(f"Output will be saved to: {output_dir}")
+fig.savefig(output_dir / "both_hemisphere.png")
```
Remi-Gau, line 138: Am feeling lucky so let's see if this works also when running the doc in CI

Remi-Gau:
OK I am not quite sure where else to put the kaleido install chrome warning to make it more obvious. Suggestions welcome @hndgzkn

examples/01_plotting/plot_3d_map_to_surface_projection.py
hndgzkn: I think it is a very good idea.

Remi-Gau:
@hndgzkn 
can you give this one a final look before I merge?

hndgzkn:
> OK I am not quite sure where else to put the kaleido install chrome warning to make it more obvious. Suggestions welcome @hndgzkn

It seems ok to me. In maskers reports as well, we have possibility to use plotly, but we do not need kaleido there, no?

Remi-Gau:
> > OK I am not quite sure where else to put the kaleido install chrome warning to make it more obvious. Suggestions welcome @hndgzkn
> 
> It seems ok to me. In maskers reports as well, we have possibility to use plotly, but we do not need kaleido there, no?

nope: we do not save images to disk in the masker reports

Remi-Gau:
> Your suggestion to add in changelog that new kaleido version requires Chrome to be installed separately would be nice.

will do

PR #5867, opened 2025-11-27
[MAINT] run flaky tests separately

Remi-Gau:
- Closes none

- run tests marked as flaky in separate job

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -199,6 +199,9 @@ jobs:
 
 
     test_flaky:
+        # Some tests are marked as flaky in order to reduce the noise in CI results.
+        # These tests are run separately here to monitor their status.
+        # We also run them on a single processor to reduce the chance of intermittent failures.
         needs: [check_skip_flags, define-matrix]
 
         if: github.repository == 'nilearn/nilearn'


Remi-Gau:
@nilearn/core 
I was getting tired of CI runs failing because of some flaky tests, so now all tests marked as flaky will be run in a separate CI job so that if they fail, we will quickly know if a PR is failing because of some flaky test or not.
Also not running those tests in parallel (with pytest xdist) should reduce the odds of them actually failing.

Remi-Gau:
> LGTM!
> 
> PS: Didn't know about the `pytest.mark.flaky` thing. I noticed that some of the tests are already marked with that.

the marker comes with 'pytest-rerunfailures': if you do not have it installed you may get some errors

Remi-Gau:
> The PR LGTM, but we should define a policy regarding when this flaky tests should be run. Otherwise, they will get forgotten.

@bthirion 
with this PR: having the flaky tests run separately not in parallel can also be a way to check if the flakyness does not come just from trying to run them in parallel

bthirion:
Yes, but you just want to run them before releases ? or when merging to main ?

Remi-Gau:
flaky tests are always run: on every PR and after the merge after main. But if any of them fail they will not stop all the other runs of the test suite (with different OS or python version) because they are run as part of a different job.

bthirion:
That's great, thx.

PR #5868, opened 2025-11-27
[ENH] Refactor reporting - move GLM reporting to `nilearn.glm`

hndgzkn:
- Closes None
- Related to #5828

This PR is a draft to demonstrate how GLM reporting code could be moved to `nilearn.glm` package, and removing dependency from `reporting` package to `glm` package.

Currently there is no implicit/explicit dependency from `reporting` package to `glm` and `maskers`.

- Create `nilearn.reporting.html_report.ReportMixin` that defines common functionality for the estimators that require reporting functionality.
- Rename `nilearn.maskers._mixin._ReportingMixin` as `MaskerReportMixin` and make it inherit from `nilearn.reporting.html_report.ReportMixin`.
- Move code in `nilearn.reporting.glm_reporter` to `nilearn.glm._base` and `nilearn.glm._report_utils` and make `nilearn.glm._base.BaseGLM` inherit from `nilearn.reporting.html_report.ReportMixin` and implement `generate_report`.
- All necessary fields for template rendering is set in `_report_content` and `_report_info` and while rendering template, these two dictionaries are merged.
-----

Questions:

1. `nilearn.reporting.glm_reporter.make_glm_report` is still in reporting package. It calls `generate_report` on the model specified. Shall we start a deprecation cycle informing the user that `generate_report` can now be called on the estimator?
2. As `nilearn.reporting.html_report.generate_report` was not in public API, I've removed it. Is that OK?
3. In the current code, I've introduced `_report_info` however instead `_report_content` can well be used. The inconvenience was that, after generating report, I clear `_report_info`. Some tests fail if `_report_content` is reset. If there is no inconvenience to do that, all report template data can be put in `_report_content` dictionary and this could save us to call rendering the template differently.
4. Is there an inconvenience to change the template directory name of maskers from `maskers` to `masker`? That would enable us to use `estimator_type` attribute of estimator while locating the estimator template.
5. Description or summary is not defined for GLM, do we need those?
-----

TODOs for follow-up PRs:

1. Instead of exposing internal data structure of `ReportMixin`, define functions that modify internal state.
2. Move some checks that exist in `generate_report` method of implementing estimators to `_run_report_checks` function of those estimators.
3. For GLM estimators, fit function creates `_reporting_data` without checking `reports` attribute. This data is later used in the code (may be in the tests, not verified yet). This should be changed and when `reporting=False`, `_reporting_data` should not exist.

### DIFF
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -568,7 +568,7 @@ class SecondLevelModel(BaseGLM):
         self.n_jobs = n_jobs
         self.minimize_memory = minimize_memory
 
-        self.reports = True
+        self.reports = reports
         self._reset_report()
 
     @fill_doc


Remi-Gau:
just read the top description of the PR and I think we agree on the big picture.

Remi-Gau:
> * `nilearn.reporting.glm_reporter.make_glm_report` is still in reporting package. It calls `generate_report` on the model specified. Shall we start a deprecation cycle informing the user that `generate_report` can now be called on the estimator?

I would also be in favor of deprecating make_glm_report and having users just call generate_report


> * As `nilearn.reporting.html_report.generate_report` was not in public API, I've removed it. Is that OK?

I would say yes


> * In the current code, I've introduced `_report_info` however instead `_report_content` can well be used. The inconvenience was that, after generating report, I clear `_report_info`. Some tests fail if `_report_content` is reset. If there is no inconvenience to do that, all report template data can be put in `_report_content` dictionary and this could save us to call rendering the template differently.

Would have to look in detail

> * Is there an inconvenience to change the template directory name of maskers from `maskers` to `masker`? That would enable us to use `estimator_type` attribute of estimator while locating the estimator template.

fine with me

> * Description or summary is not defined for GLM, do we need those?

not sure

nilearn/glm/first_level/first_level.py
```diff
@@ -1361,28 +1374,7 @@ def generate_report(
...
-            height_control=height_control,
-        )
-
-        if not hasattr(self, "_reporting_data"):
-            self._reporting_data = {
-                "trial_types": [],
-                "noise_model": getattr(self, "noise_model", None),
-                "hrf_model": getattr(self, "hrf_model", None),
-                "drift_model": None,
-            }
```
hndgzkn, line 1382: This is removed. If necessary, could be added if `self.reports=True`.

nilearn/glm/first_level/first_level.py
```diff
@@ -1361,28 +1374,7 @@ def generate_report(
...
-        parameters = inspect.signature(
-            FirstLevelModel.generate_report
-        ).parameters
-        warn_default_threshold(
-            threshold,
-            parameters["threshold"].default,
-            3.09,
-            height_control=height_control,
-        )
-
```
hndgzkn, line 1375: This is moved to `BaseGLM.generate_report`

nilearn/glm/second_level/second_level.py
```diff
@@ -890,81 +899,6 @@ def _get_element_wise_model_attribute(
             )
         return self.masker_.inverse_transform(voxelwise_attribute)
 
-    def generate_report(
-        self,
```
hndgzkn, line 894: No need to define it here, `BaseGLM.generate_report` satisfies.

nilearn/glm/_base.py
```diff
@@ -333,6 +349,231 @@ def _generate_filenames_output(
...
+
+        report_info["page_title"] = (
+            f"Statistical Report - {self.__str__()}{title}"
+        )
+        report_info["estimator_type"] = "glm"
+        report_info["model_type"] = self.__str__()
+        report_info["show_navbar"] = (
+            "style='display: none;'" if self._is_notebook() else ""
+        )
+        report_info["is_volume_glm"] = self._is_volume_glm()
```
hndgzkn, line 365: This is currently not used, might be removed.

nilearn/glm/_base.py
```diff
@@ -333,6 +349,231 @@ def _generate_filenames_output(
...
 
+    def _set_report_basics(self, title):
+        super()._set_report_basics(title)
+
+        report_info = self._report_info
+
+        report_info["page_title"] = (
+            f"Statistical Report - {self.__str__()}{title}"
+        )
+        report_info["estimator_type"] = "glm"
```
hndgzkn, line 360: If we change maskers template directory name to be `masker`, this can be removed as well.

nilearn/glm/_base.py
```diff
@@ -333,6 +349,231 @@ def _generate_filenames_output(
             "model_level_mapping": model_level_mapping,
         }
 
+    def _set_report_basics(self, title):
+        super()._set_report_basics(title)
+
+        report_info = self._report_info
```
hndgzkn, line 355: Instead of `_report_info` `_report_content` can also be used (once we agree that there is no inconvenience to resetting `_report_content` after report generation).

nilearn/glm/_base.py
```diff
@@ -333,6 +349,231 @@ def _generate_filenames_output(
...
+        )
+        report_info["is_volume_glm"] = self._is_volume_glm()
+        smoothing_fwhm = getattr(self, "smoothing_fwhm", 0)
+        if smoothing_fwhm == 0:
+            smoothing_fwhm = None
+        report_info["smoothing_fwhm"] = smoothing_fwhm
+
+    @fill_doc
+    def generate_report(
+        self,
```
hndgzkn, line 373: I've overwritten `nilearn.reporting.html_report.ReportMixin.generate_report` here with the contents of `make_glm_report`. There  are minor changes.

I've removed for example resizing the report with specified dimension. It is something to be fixed. Can it be generalized to all reports?

What I would prefer is to overwrite `ReportMixin._generate_report_data`. This requires a bit more reflection as for glm there are many other params in the signature.

nilearn/glm/_reporting_utils.py
```diff
@@ -0,0 +1,610 @@
+"""Utility functions to create an GLM HTML report."""
+
+import warnings
```
hndgzkn, line 3: All functions in this module are moved here from `nilearn.reporting.glm_reporter` as they are. Just `make_glm_report` is removed.

Remi-Gau:
FYI: we can start to decide a deprecation cycle of make_glm_report before the release (and add all the warnings and so on) but also before merging this PR: this way we can start warning users sooner.

nilearn/glm/_base.py
Remi-Gau: yes resizing should be something we can do for all report

nilearn/reporting/glm_reporter.py
```diff
@@ -222,754 +170,19 @@ def make_glm_report(
...
-    env = return_jinja_env()
-
-    body_tpl = env.get_template("html/glm/body_glm.jinja")
-
-    # TODO clean up docstring from RST formatting
-    docstring = model.__doc__.split("Parameters\n")[0]
-
-    body = body_tpl.render(
-        docstring=docstring,
+    return model.generate_report(
```
Remi-Gau, line 173: add deprecation warning

nilearn/reporting/html_report.py
```diff
@@ -93,6 +96,358 @@ def __str__(self):
...
+    def _embed_img(self, img):
+        """Embed the image."""
+        return embed_img(img)
+
+    def _is_notebook(self):
+        """Detect if we are running in a notebook.
+
+        From https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook
+        """
+        return is_notebook()
```
Remi-Gau, line 300: maybe is_notebook can directly be inlined here, no?

nilearn/reporting/html_report.py
```diff
@@ -93,6 +96,358 @@ def __str__(self):
...
+                parameters,
+                precision=2,
+                header=True,
+                sparsify=False,
+            )
+        return parameters
+
+    def _embed_img(self, img):
+        """Embed the image."""
+        return embed_img(img)
```
Remi-Gau, line 293: embed_img can be inlined here, no?

hndgzkn: I agree, I just couldn't be sure if that might be used elsewhere, thus required to be imported.

Remi-Gau: Double check bu I think it is not.

hndgzkn:
@Remi-Gau I would like to divide this PR into pieces if that's ok with you. 

I moved some code, but I also changed the contents. So I think this makes it difficult to review the changes. I would be more confident if we do this in small pieces. What I want to do is:

- First move contents of `make_glm_report` to `glm.generate_report` and create a `_GLMReportMixin` if necessary. 
- Then extract common interface with `_ReportingMixin` in maskers to `reporting` package.

How does it sound to you?

Remi-Gau:
> I moved some code, but I also changed the contents. So I think this makes it difficult to review the changes. I would be more confident if we do this in small pieces. What I want to do is:
> 
>     * First move contents of `make_glm_report` to `glm.generate_report` and create a `_GLMReportMixin` if necessary.
> 
>     * Then extract common interface with `_ReportingMixin` in maskers to `reporting` package.
> 
> 
> How does it sound to you?

yes that sounds like a good plan

actually the very first PR could be :
- move contents of `make_glm_report` to `glm.generate_report`
- start deprecation cycle of make_glm_report by adding a deprecation warning

Remi-Gau:
@hndgzkn
will close this one
I think further refactoring have either been done or documented in other issues.

PR #5869, opened 2025-11-27
[MAINT] test pre-release of dependencies separately

Remi-Gau:
- Closes none but should fix failure of CI in other PRs

- just tests with pre-release of our dependencies as it seems to have started failing

### DIFF
--- a/nilearn/decoding/space_net.py
+++ b/nilearn/decoding/space_net.py
@@ -398,6 +398,7 @@ def path_scores(
     # it is essential to center the data in regression
     # do not unpack tuple completely
     # as it returns more values starting from sklearn 1.8
+    # TODO: try to find a public function in sklearn to do this
     tmp = center_data(X_train, y_train, fit_intercept=True, copy=False)
     X_train = tmp[0]
     y_train = tmp[1]


Remi-Gau:
ok yeah the pre-release version of the tests if failing

will put it in a separate job, because its failure should not prevent other CI jobs from running (we have no control over the pre-release of our dependencies)

Remi-Gau:
ok looks like we have got a scikit-learn==1.8.0rc1 on the horizon

nilearn/decoding/space_net.py
```diff
@@ -396,9 +396,13 @@ def path_scores(
     X_test, y_test = X[test].copy(), y[test].copy()
 
     # it is essential to center the data in regression
-    X_train, y_train, _, y_train_mean, _ = center_data(
-        X_train, y_train, fit_intercept=True, copy=False
-    )
+    # do not unpack tuple completely
+    # as it returns more values starting from sklearn 1.8
+    tmp = center_data(X_train, y_train, fit_intercept=True, copy=False)
```
Remi-Gau, line 402: Center_data is actually a private function from sklearn.
Using a public equvalent may be better to avoid things breaking without warning.

bthirion:
LGTM so far.

nilearn/decoding/space_net.py
```diff
@@ -396,9 +396,13 @@ def path_scores(
     X_test, y_test = X[test].copy(), y[test].copy()
 
     # it is essential to center the data in regression
-    X_train, y_train, _, y_train_mean, _ = center_data(
-        X_train, y_train, fit_intercept=True, copy=False
-    )
+    # do not unpack tuple completely
+    # as it returns more values starting from sklearn 1.8
```
Remi-Gau, line 400:
```suggestion
    # as it returns more values starting from sklearn 1.8
    # TODO: try to find a public function in sklearn to do this
```

Remi-Gau:
OK will merge to free other PR from this problem

PR #5876, opened 2025-12-01
[ENH] Refactor GLM reporting

hndgzkn:
- Closes None, relates to #5828 

- Move report generation code for GLM from `nilearn.reporting.glm_reporter` to `nilearn.glm._base`
- Remove dependency from `nilearn.reporting` to `nilearn.glm`
- Start deprecation cycle for `make_glm_report`

### DIFF
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -48,7 +48,6 @@ from nilearn.glm.regression import (
     RegressionResults,
     SimpleRegressionResults,
 )
-from nilearn.glm.thresholding import warn_default_threshold
 from nilearn.image import check_niimg, get_data
 from nilearn.interfaces.bids import get_bids_files, parse_bids_filename
 from nilearn.interfaces.bids.query import (
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -37,7 +37,6 @@ from nilearn.glm.first_level.design_matrix import (
     make_second_level_design_matrix,
 )
 from nilearn.glm.regression import RegressionResults, SimpleRegressionResults
-from nilearn.glm.thresholding import warn_default_threshold
 from nilearn.image import check_niimg, concat_imgs, iter_img, mean_img
 from nilearn.maskers import NiftiMasker, SurfaceMasker
 from nilearn.maskers.masker_validation import check_embedded_masker


nilearn/glm/first_level/first_level.py
```diff
@@ -1361,28 +1360,7 @@ def generate_report(
...
-            height_control=height_control,
-        )
-
-        if not hasattr(self, "_reporting_data"):
-            self._reporting_data = {
-                "trial_types": [],
-                "noise_model": getattr(self, "noise_model", None),
-                "hrf_model": getattr(self, "hrf_model", None),
-                "drift_model": None,
-            }
```
hndgzkn, line 1381: This part is moved to `nilearn.glm._base.BaseGLM.generate_report`.

nilearn/glm/second_level/second_level.py
```diff
@@ -890,81 +888,6 @@ def _get_element_wise_model_attribute(
             )
         return self.masker_.inverse_transform(voxelwise_attribute)
 
-    def generate_report(
-        self,
```
hndgzkn, line 893: `generate_report` is not necessary in second_level. Base class `generate_report` is already doing it.

nilearn/glm/tests/test_glm_reporter.py
```diff
@@ -585,3 +585,17 @@ def test_report_threshold_deprecation_warning(
         assert n_warnings == 1
     else:
         assert n_warnings == 0
+
+
+def test_report_make_glm_deprecation_warning(flm, contrasts):
+    """Test deprecation warning for nilearn.reporting.make_glm_report.
```
hndgzkn, line 571: Added this function in `glm` package instead of `reporting` to avoid adding a new test file and also because fixtures necessary to run the test are defined in this module.

nilearn/glm/_base.py
```diff
@@ -333,6 +358,289 @@ def _generate_filenames_output(
             "model_level_mapping": model_level_mapping,
         }
 
+    def generate_report(
+        self,
```
hndgzkn, line 370: Moved `make_glm_report` code here. Made minor modifications to fix code complexity complaints.

nilearn/glm/_base.py
```diff
@@ -333,6 +358,282 @@ def _generate_filenames_output(
...
+        # we do not rely on filenames stored in the model.
+        output = None
+        if contrasts is None:
+            output = self._reporting_data.get("filenames", None)
+            if output is not None and output.get("use_absolute_path", True):
+                output = _turn_into_full_path(output, output["dir"])
+
+            warning_messages.append(
+                "No contrast passed during report generation."
+            )
```
hndgzkn, line 573: This was later done in another `if contrasts is None` block, I moved it here.

nilearn/glm/_base.py
```diff
@@ -333,6 +358,282 @@ def _generate_filenames_output(
...
+                if self.__str__() == "Second Level Model"
+                else self.design_matrices_
+            )
+
+            # TODO move matplotlib check to if condition below
+            MNI152TEMPLATE = None
+            if is_matplotlib_installed():
+                from nilearn.plotting.image.utils import (  # type: ignore[assignment]
+                    MNI152TEMPLATE,
+                )
```
hndgzkn: This was at the top of the module. I moved it here.

nilearn/glm/_reporting_utils.py
```diff
@@ -0,0 +1,567 @@
+import warnings
```
hndgzkn, line 1: The functions in this module are copied from `nilearn.reporting.glm_reporter` as they are.

nilearn/reporting/__init__.py
```diff
@@ -7,6 +7,8 @@
 from nilearn.reporting.glm_reporter import make_glm_report
 from nilearn.reporting.html_report import HTMLReport
 
+# TODO (nilearn >= 0.15.0) remove make_glm_report
```
Remi-Gau, line 10: Can we get an entry in the changelog for this deprecation? 😉

nilearn/glm/_reporting_utils.py
```diff
@@ -0,0 +1,562 @@
...
+                stat_threshold=threshold,
+                cluster_threshold=cluster_threshold,
+                min_distance=min_distance,
+                two_sided=two_sided,
+            )
+            cluster_table_html = dataframe_to_html(
+                cluster_table,
+                precision=2,
+                index=False,
+            )
```
Remi-Gau: See this comment to fix this:
https://github.com/nilearn/nilearn/pull/5751/files#r2580202696

hndgzkn: Ok, merged main branch and reflected the changes.

Remi-Gau: thanks

doc/conf.py
hndgzkn, line 33: I think this line makes pre-commit fail. However I don't have the failure in local and it does not fail in main either. 

I am not sure if the failure has anything to do with my moving nilearn import below. 

@Remi-Gau how do you think I should solve this one?

- Move this import above
- Or ignore by `# noqa: E402`

Remi-Gau: > Or ignore by # noqa: E402

if that fixes it let's do thit

nilearn/glm/tests/test_glm_reporter.py
hndgzkn: This test fails when `threshold=3.09` and `height_control=None`.

In main branch, the warning is raised from `FirstLevelGLM.generate_report` as below:

https://github.com/nilearn/nilearn/blob/ff784094a302b30b23bbbe09eaadc563537b9086/nilearn/glm/first_level/first_level.py#L1369-L1375

when old_default is specified as `3.09`.

However in make_glm_report, the same warning is raised when old_default is specified as `3.0`.

https://github.com/nilearn/nilearn/blob/a523ce130f20fabd74b7cb0388d14360fb82c68d/nilearn/reporting/glm_reporter.py#L244-L249

As far as I understand the correct behavior is the second one. So I kept value 3.0 in the code.

Is this correct and shall I change the threshold value `3.09` for the test to `3.0`?

Remi-Gau: need to have a look but it goes back to making sure we use the same default threshold and make_glm_report and generate_report had different threshold

- see https://github.com/nilearn/nilearn/pull/5601

doc/conf.py
hndgzkn: Finally I reverted the changes in file, and CI does not complain anymore. I could not make pass the above solution with pre-commit.

Remi-Gau: I get the same problem where running pre-commit on all files: need to check why this is happening locally but not in CI

hndgzkn: I can check it if you want. Not in this PR though.

nilearn/glm/tests/test_glm_reporter.py
hndgzkn: I think there is a problem with `warn_default_threshold` function and the warning from `glm.generate_report` and `make_glm_report`. The initial value for threshold in both first and second level `generate_report` methods and `make_glm_report` (when they were first introduced) seem to be `3.09`. 
Looking at the code, if the user does not touch threshold value, the warning will always be raised in `generate_report` because old_default is set as `3.09`. 

https://github.com/nilearn/nilearn/blob/a523ce130f20fabd74b7cb0388d14360fb82c68d/nilearn/glm/thresholding.py#L27-L49

However the default value is equal to `DEFAULT_Z_THRESHOLD ~= 3.09`

For `make_glm_report`, old_default was set as `3.0`, so no warning.

hndgzkn: I have removed the relevant part from code for now. If I am mistaken, I will revert it.

nilearn/reporting/__init__.py
hndgzkn: Done.

nilearn/glm/_reporting_utils.py
```diff
@@ -0,0 +1,552 @@
...
+
+    return model_attributes
+
+
+def _mask_to_plot(model, bg_img, cut_coords):
+    """Plot a mask image and creates PNG code of it.
+
+    Parameters
+    ----------
+    model
```
bthirion: to be completed

nilearn/glm/_reporting_utils.py
```diff
@@ -0,0 +1,552 @@
...
+
+    bg_img : Niimg-like object
+        See :ref:`extracting_data`.
+        The background image that the mask will be plotted on top of.
+        To turn off background image, just pass "bg_img=None".
+
+    cut_coords
```
bthirion: to be completed

nilearn/glm/_reporting_utils.py
```diff
@@ -0,0 +1,552 @@
...
+            precision=3,
+            header=False,
+        )
+
+        # FIXME
+        # The commented code below was there to reuse
+        # cluster tables generated by save_glm_to_bids
+        # to save time.
+        # However cluster tables may have been computed
+        # with different threshold, cluster_threshol...
```
bthirion, line 307: I guess that the idea would be default to parameters that have already been used --- if computations have already been performed.

bthirion (commented):
Thx. Just find a few minor comments;

nilearn/glm/_reporting_utils.py
hndgzkn: Done.

nilearn/glm/_reporting_utils.py
hndgzkn: Done.

nilearn/glm/_reporting_utils.py
hndgzkn: I copied the contents of this file from `nilearn.reporting.glm_reporter` without changing. This could be handled in another PR.

nilearn/glm/tests/test_glm_reporter.py
```diff
@@ -558,28 +557,15 @@ def test_carousel_several_runs(
...
     """
-    with warnings.catch_warnings(record=True) as warning_list:
-        flm.generate_report(
+    with pytest.warns(DeprecationWarning):
+        make_glm_report(
+            flm,
             contrasts=contrasts,
-            threshold=threshold,
-            height_control=height_control,
+            threshold=3.9,
```
bthirion: Why this threshold ? 3.09 would be a more standard choice (corresponding to  p< 0.001)

bthirion (approved):
LGTM overall wiyth one very minor comment --- thx !

nilearn/glm/tests/test_glm_reporter.py
hndgzkn: Actually I can even remove threshold from here. 3.09 is the default value for threshold. And the aim of this test is just to see if we get DeprecationWarning when `make_glm_report` is used.

hndgzkn:
```suggestion
```

bthirion:
LGTM, thx !

nilearn/reporting/glm_reporter.py
```diff
@@ -222,744 +167,45 @@ def make_glm_report(
...
-            )
-
-        logger.log(
-            "Generating contrast-level figures...", verbose=model.verbose
-        )
-        results = _make_stat_maps_contrast_clusters(
-            stat_img=statistical_maps,
-            threshold_orig=threshold,
+    warnings.warn(
+        category=DeprecationWarning,
```
Remi-Gau:
```suggestion
        category=FutureWarning,
```

We need to use a FutureWarning otheriwise it may not show for users unless they run in development mode: https://docs.python.org/3/library/devmode.html#devmode

Remi-Gau: I think the test for this will need to be adapted

nilearn/glm/tests/test_glm_reporter.py
```diff
@@ -558,28 +557,14 @@ def test_carousel_several_runs(
...
+def test_report_make_glm_deprecation_warning(flm, contrasts):
+    """Test deprecation warning for nilearn.reporting.make_glm_report.
 
     # TODO (nilearn >= 0.15)
     # remove
     """
-    with warnings.catch_warnings(record=True) as warning_list:
-        flm.generate_report(
+    with pytest.warns(DeprecationWarning):
+        make_glm_report(
```
hndgzkn:
```suggestion
    with pytest.warns(FutureWarning):
        make_glm_report(
```

nilearn/reporting/glm_reporter.py
hndgzkn: Done.

Remi-Gau:
We have a warning situation here:

Tried the following and I am getting more warnings than I would like and also not getting the warnings I would expect.

Can you double check? Do my comments make sense?

Note that some of that behavior is also not correct on main. So I'd say that we should merge this PR and fix this in a follow up PR (also to make the review easier).

```python
from nilearn.glm.thresholding import DEFAULT_Z_THRESHOLD
import numpy as np

    shapes=[(30,31,32,33)], rk=3
)

flm = FirstLevelModel(mask_img=mask).fit(
    fmri_data[0], design_matrices=design_matrices[0]
)

contrasts = [np.asarray([1, 0, 0]), np.asarray([1, 1, 0]), np.asarray([1, 1, 1])]


print("\ndefaults")
# Should get a SINGLE FutureWarning that the default threshold will change in 0.15.
# Currently getting one warning per contrasts,
# about the fact threshold is not used if 'height_control=None'
flm.generate_report(contrasts=contrasts)

print("\nthreshold=3")
# changing threshold: should get a warning threshold is not used unless 'height_control=None'
# current warning says "Set 'threshold' to '3.09' to avoid this warning."
# which is misleading: should probably says "Set 'height_control' to 'None' to actually use this threshold."
flm.generate_report(contrasts=contrasts, threshold=3)

print("\nheight_control=None")
# No warning but I would expect 
# to get a SINGLE FutureWarning that the default threshold will change in 0.15.
flm.generate_report(contrasts=contrasts, height_control=None)

print(f"\nthreshold={DEFAULT_Z_THRESHOLD}")
# Now I am getting contradictory warnings that say to set threshold to 3.0 or 3.09 to avoid warnings
# Using the future threshold that will be used:
# - should not get a deprecation warning (that's the case)  
# - should a single warning that the thresold will not be used unless 'height_control=None'
flm.generate_report(contrasts=contrasts, threshold=DEFAULT_Z_THRESHOLD)

print(f"\nheight_control=None, threshold={DEFAULT_Z_THRESHOLD}")
# no warning: as expected
flm.generate_report(contrasts=contrasts, height_control=None, threshold=DEFAULT_Z_THRESHOLD)

```

nilearn/glm/thresholding.py
```diff
@@ -33,7 +33,6 @@ def warn_default_threshold(
     Can be removed.
     """
     if height_control is None and threshold == current_default == old_default:
-        print(f"{threshold=} {current_default=} {old_default=}")
```
Remi-Gau: @hndgzkn 
pushed a commit to remove this: most likely left over from a print debugging session

nilearn/glm/_base.py
```diff
@@ -333,6 +357,275 @@ def _generate_filenames_output(
...
+        if height_control is not None and float(threshold) != float(
+            parameters["threshold"].default
+        ):
+            warning_messages.append(
+                f"\n'{threshold=}' is not used with '{height_control=}'."
+                "\n'threshold' is only used when 'height_control=None'. "
+                f"\nSet 'threshold' to '{parameters['threshold'].default}' "
+                "to avoid this warning."
+            )
+        model_attributes = _glm_model_attributes_to_dataframe(self)
```
Remi-Gau, line 552: related to my comment about warnings I think we should still use the warn_default_threshold function somewhere here and also prevent similar warnings to thrown downstream by thresholding functions

Remi-Gau: as I said: we can do that in a follow up PR

Remi-Gau:
> We have a warning situation here:
> 
> Tried the following and I am getting more warnings than I would like and also not getting the warnings I would expect.
> 
> Can you double check? Do my comments make sense?
> 
> Note that some of that behavior is also not correct on main. So I'd say that we should merge this PR and fix this in a follow up PR (also to make the review easier).

note to self: check if some of the cluster thresholding should be affected by the change of default threshold

Remi-Gau (changes requested):
TODO:
- [ ] removing use of `make_glm_report` from doc/visual_testing/reporter_visual_inspection_suite.py

nilearn/glm/_base.py
```diff
@@ -333,6 +357,275 @@ def _generate_filenames_output(
...
+        HTML file.
+
+        The :term:`GLM` must be fitted and have the computed design
+        matrix(ces).
+
+        .. note::
+
+            Refer to the documentation of
+            :func:`~nilearn.reporting.make_glm_report`
+            for details about the parameters
```
Remi-Gau: Can we instead also move the doc string from make_glm_report here, and instead shorten the one of make_glm_report?

nilearn/glm/second_level/second_level.py
```diff
@@ -890,81 +888,6 @@ def _get_element_wise_model_attribute(
...
-        HTML file.
-
-        The :term:`GLM` must be fitted and have the computed design
-        matrix(ces).
-
-        .. note::
-
-            Refer to the documentation of
-            :func:`~nilearn.reporting.make_glm_report`
-            for details about the parameters
```
Remi-Gau, line 923: I think the FirstLevelModel.generate_report must have a similar note that may not be ideal if we are going to deprecate make_glm_report

Remi-Gau:
@hndgzkn before we merge, can you at one point trigger a build of the reports in the doc by adding  `[reports]` to one of your commit message

I have checked them locally and they seem fine but having another look cannot hurt

Remi-Gau:
> note to self: check if some of the cluster thresholding should be affected by the change of default threshold

- see #5883

hndgzkn:
> We have a warning situation here:
> 
> Tried the following and I am getting more warnings than I would like and also not getting the warnings I would expect.
> 
> Can you double check? Do my comments make sense?
> 
> Note that some of that behavior is also not correct on main. So I'd say that we should merge this PR and fix this in a follow up PR (also to make the review easier).

I completely agree. Your comments make sense. I do not reply inline here for each of them. What I propose is to create an issue from your comment and discuss on that issue.

Remi-Gau:
> What I propose is to create an issue from your comment and discuss on that issue.

sounds good

hndgzkn:
> TODO:
> 
>     * [x]  removing use of `make_glm_report` from doc/visual_testing/reporter_visual_inspection_suite.py

I've converted `make_glm_report` to `model.generate_report`.

nilearn/glm/_base.py
hndgzkn: Done.

nilearn/glm/second_level/second_level.py
hndgzkn: Removed

hndgzkn:
> There is also a mention of make_glm_report in nilearn/glm/io.py that should probably be adapted nilearn/glm/io.py:156: See :func:`nilearn.reporting.make_glm_report` for more details.

I've removed the line. Is that ok?

Remi-Gau:
> I've removed the line. Is that ok?

yup

Remi-Gau:
Still need some docstring improvements:

- [first level generate_report](https://output.circle-artifacts.com/output/job/ff6b5899-533a-4ca4-82e0-373dd3ad49d7/artifacts/0/dev/modules/generated/nilearn.glm.first_level.FirstLevelModel.html#nilearn.glm.first_level.FirstLevelModel.generate_report): It'd be nice to be able to just rely on the docstring from the BaseGLM because otherwise we have to almost duplicate the entire docstring. I think the only difference in the signature between the first and second level generate_report is the `first_level_contrast` parameter, no?. I'd be ok just relying on the generate_report from BaseGLM for both first and second level but we have to make it clear in the docstring that `first_level_contrast` is only for SecondLevelModel (and also throw a warning if user are trying to pass some value to it when using it in FirstLevelModel)

- [second level generate_report](https://output.circle-artifacts.com/output/job/ff6b5899-533a-4ca4-82e0-373dd3ad49d7/artifacts/0/dev/modules/generated/nilearn.glm.second_level.SecondLevelModel.html#nilearn.glm.second_level.SecondLevelModel.generate_report): some template doc strings are not injected - I suspect we are missing a `@fill_doc` generator in the BaseGLM generate_report method.

hndgzkn:
> Still need some docstring improvements:
> 
>     * [first level generate_report](https://output.circle-artifacts.com/output/job/ff6b5899-533a-4ca4-82e0-373dd3ad49d7/artifacts/0/dev/modules/generated/nilearn.glm.first_level.FirstLevelModel.html#nilearn.glm.first_level.FirstLevelModel.generate_report): It'd be nice to be able to just rely on the docstring from the BaseGLM because otherwise we have to almost duplicate the entire docstring. I think the only difference in the signature between the first and second level generate_report is the `first_level_contrast` parameter, no?. I'd be ok just relying on the generate_report from BaseGLM for both first and second level but we have to make it clear in the docstring that `first_level_contrast` is only for SecondLevelModel (and also throw a warning if user are trying to pass some value to it when using it in FirstLevelModel)

Done

>     * [second level generate_report](https://output.circle-artifacts.com/output/job/ff6b5899-533a-4ca4-82e0-373dd3ad49d7/artifacts/0/dev/modules/generated/nilearn.glm.second_level.SecondLevelModel.html#nilearn.glm.second_level.SecondLevelModel.generate_report): some template doc strings are not injected - I suspect we are missing a `@fill_doc` generator in the BaseGLM generate_report method.

Done.

Thank you!

hndgzkn:
The failure should not be due to the PR.

Remi-Gau:
> The failure should not be due to the PR.

indeed

PR #5877, opened 2025-12-02
[MAINT] mark test as flaky

Remi-Gau:
- Closes none

- mark one test as flaky


Remi-Gau:
In a follow up PR, it will be good to check which of the 'flaky' tests are just tests that 'do not like' being run in parallel with pytest-xdist.

Similarly I believe we have a few tests that 'pretend' to test the joblib parallel processing but because of xdist they only have single core to run on: so it may be good to also run those without using xdist.

Remi-Gau:
will merge to avoid 'flaky' tests issues in other PRs

PR #5883, opened 2025-12-04
[FIX] check input image cluster inference

Remi-Gau:
- Closes none

- raise meaningful error when trying to perform cluster level inference on a 4D image (or its surface equivalent): explicit restriction of cluster level inference to 3D volume image (or their surface equivalent)
- check that default deprecation warning for threshold value is also thrown with surface data

### DIFF
--- a/nilearn/decomposition/tests/test_base.py
+++ b/nilearn/decomposition/tests/test_base.py
@@ -79,6 +79,7 @@ def test_mask_reducer_multiple_image(
     assert data.shape == expected_shape
 
 
+@pytest.mark.single_process
 @pytest.mark.parametrize("data_type", ["nifti", "surface"])
 def test_mask_reducer_single_image_same_with_multiple_jobs(
     data_type, decomposition_masker, decomposition_img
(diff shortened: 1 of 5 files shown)


bthirion (commented):
LGTM so far.

nilearn/glm/tests/test_thresholding.py
```diff
@@ -437,6 +438,21 @@ def test_all_resolution_inference_threshold_errors(
         cluster_level_inference(stat_img, threshold=threshold)
 
 
+def test_all_resolution_inference_shape_errors(img_4d_rand_eye, surf_img_2d):
+    """Raise error with 4D image or SurfaceImage with n_samples>1."""
+    with pytest.raises(
+        DimensionError,
+        match=("Input data has incompatible dimensionality"),
+    ):
+        cluster_level_inference(img_4d_rand_eye, threshold=0.5)
```
Remi-Gau, line 447: @bthirion we agree that cluster_level_inference should run on 3D images?
Note that it does fail currently on 4D images but the error is not very clear.

bthirion: Agreed.

PR #5886, opened 2025-12-05
[ENH] Refactor Slicers to check cut_coords depending on slicer type

hndgzkn:
- Closes #5880 
- Closes #5903

Changes proposed in this pull request:

- Check `cut_coord` value while initializing slicers depending on slicer type.
- Add tests for different slicer types.
- Update/unify docstrings

PR #3452, opened 2022-12-28
[ENH] add `cut_coords` parameter in `reporting.make_glm_report` method

AudreyDuran:
Closes #3442 .

Changes proposed in this pull request:

- add `cut_coords` parameter in `reporting.make_glm_report` method
- adapt pytest accordingly

I could not build the updated Sphinx documentation, so the docstrings changes should be carefully reviewed

PR #5892, opened 2025-12-11
[FIX] Fix warnings when generating reports

### LINKED ISSUE #5889: Fix warnings when generating GLM reports
We have a warning situation here:

Tried the following and I am getting more warnings than I would like and also not getting the warnings I would expect.

Can you double check? Do my comments make sense?

Note that some of that behavior is also not correct on main. So I'd say that we should merge this PR and fix this in a follow up PR (also to make the review easier).

```python
from nilearn.glm.thresholding import DEFAULT_Z_THRESHOLD
import numpy as np

    shapes=[(30,31,32,33)], rk=3
)

flm = FirstLevelModel(mask_img=mask).fit(
    fmri_data[0], design_matrices=design_matrices[0]
)

contrasts = [np.asarray([1, 0, 0]), np.asarray([1, 1, 0]), np.asarray([1, 1, 1])]


print("\ndefaults")
# Should get a SINGLE FutureWarning that the default threshold will change in 0.15.
# Currently getting one warning per contrasts,
# about the fact threshold is not used if 'height_control=None'
flm.generate_report(contrasts=contrasts)

print("\nthreshold=3")
# changing threshold: should get a warning threshold is not used unless 'height_control=None'
# current warning says "Set 'threshold' to '3.09' to avoid this warning."
# which is misleading: should probably says "Set 'height_control' to 'None' to actually use this threshold."
flm.generate_report(contrasts=contrasts, threshold=3)

print("\nheight_control=None")
# No warning but I would expect 
# to get a SINGLE FutureWarning that the default threshold will change in 0.15.
flm.generate_report(contrasts=contrasts, height_control=None)

print(f"\nthreshold={DEFAULT_Z_THRESHOLD}")
# Now I am getting contradictory warnings that say to set threshold to 3.0 or 3.09 to avoid warnings
# Using the future threshold that will be used:
# - should not get a deprecation warning (that's the case)  
# - should a single warning that the thresold will not be used unless 'height_control=None'
flm.generate_report(contrasts=contrasts, threshold=DEFAULT_Z_THRESHOLD)

print(f"\nheight_control=None, threshold={DEFAULT_Z_THRESHOLD}")
# no warning: as expected
flm.generate_report(contrasts=contrasts, height_control=None, threshold=DEFAULT_Z_THRESHOLD)

```

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/5876#issuecomment-3611344596_

hndgzkn:
- Closes #5889 

- when threshold value is changed and height_control is not None
  - Update warning message 
  - Raise single warning instead of raising for each contrast
- Raise warning for default threshold value change starting from nilearn >= 0.15.0 when threshold value is not set.
- Use default value of threshold only if height is None
- add some input validation to generate_report

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -85,6 +85,10 @@ jobs:
             run: |
                 versions='py=["3.14", "3.13", "3.12", "3.11", "3.10"]'
 
+                if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
+                    versions='py=["3.14", "3.10"]'
+                fi
+
                 echo "Will run on the following python: ${versions}"
                 echo ${versions} >> "$GITHUB_OUTPUT"
 


nilearn/glm/_base.py
```diff
@@ -529,25 +530,25 @@ def generate_report(
             )
             first_level_contrast = None
 
-        if not hasattr(self, "_reporting_data"):
-            self._reporting_data: dict[str, Any] = {
-                "trial_types": [],
-                "noise_model": getattr(self, "noise_model", None),
-                "hrf_model": getattr(self, "hrf_model", None),
-                "drift_model": None,
-            }
```
hndgzkn, line 538: This is just moved down.

nilearn/glm/_base.py
hndgzkn: I think there is a problem here: the user should not get warning with default values. 
we can either set `height_control` or `threshold` to `None`.

bthirion:
Thx for opening it ! LMK when you want a review.

nilearn/glm/_base.py
```diff
@@ -529,26 +529,28 @@ def generate_report(
...
         ):
             warning_messages.append(
-                f"\n'{threshold=}' is not used with '{height_control=}'."
+                f"'{threshold=}' is not used with '{height_control=}'."
                 "\n'threshold' is only used when 'height_control=None'. "
-                f"\nSet 'threshold' to '{parameters['threshold'].default}' "
-                "to avoid this warning."
+                "\nSetting 'height_control' to None. "
             )
+            height_control = None
```
hndgzkn: Here if the user specifies `threshold`, I set `height_control=None`.

nilearn/glm/_base.py
Remi-Gau: I don't think we should have no correction as default to avoid a cornucopia of false positive.

let me have a look

nilearn/glm/_base.py
```diff
@@ -529,26 +529,28 @@ def generate_report(
...
+        elif height_control is not None:
+            warning_messages.append(
+                f"'{threshold=}' is not used with '{height_control=}'."
+                "\n'threshold' is only used when 'height_control=None'. "
+                "\nSetting 'threshold' to None. "
+            )
+            threshold = None
```
hndgzkn: If `threshold` value is not changed, I assume that `height_control` has priority over threshold, and set `threshold=None`. Is this reasonable? 
I think if we fix the default values, we won't have this problem. We will warn the user, only depending on the parameters specified.

nilearn/glm/_reporting_utils.py
```diff
@@ -264,18 +263,6 @@ def _make_stat_maps_contrast_clusters(
...
-        #   and also because threshold_stats_img and make_glm_report
-        #   have different defaults.
-        # TODO (nilearn>=0.15)
-        # remove
-        warnings.filterwarnings(
-            "ignore",
-            category=FutureWarning,
-            message=".*default 'threshold' will be set.*",
-        )
-
```
hndgzkn, line 278: This filter causes some problems on notebook. If I run `frm.generate_code` if 5 different cells, only the first one raises warning, but not others.

nilearn/glm/thresholding.py
```diff
@@ -186,8 +167,14 @@ def cluster_level_inference(
...
-    parameters = dict(**inspect.signature(cluster_level_inference).parameters)
-    warn_default_threshold(threshold, parameters["threshold"].default, 3.0)
+    # TODO (nilearn >= 0.15.0) remove
+    warnings.warn(
+        "From nilearn version>=0.15, "
+        "the default 'threshold' will be set to "
+        f"{DEFAULT_Z_THRESHOLD}.",
+        FutureWarning,
+        stacklevel=find_stack_level(),
+    )
```
hndgzkn: Changed the code to raise the warning whatever value specified for `threshold`.

Comparing with default value causes conflicts.

nilearn/glm/thresholding.py
```diff
@@ -407,6 +394,23 @@ def threshold_stats_img(
...
 
     """
+    # TODO (nilearn >= 0.15.0) remove
+    warnings.warn(
+        "From nilearn version>=0.15, "
+        "the default 'threshold' will be set to "
+        f"{DEFAULT_Z_THRESHOLD}.",
+        FutureWarning,
+        stacklevel=find_stack_level(),
+    )
```
hndgzkn: Changed the code to raise the warning whatever value specified for `threshold`.

Comparing with default value causes conflicts. The function `threshold_stats_img` and `glm.generate_report` has different default thresholds. If one does not raise error, the other raises.

I've changed the code to raise this warning only where `threshold` it is used.

Remi-Gau:
For this test failure: https://github.com/nilearn/nilearn/actions/runs/20161807517/job/57875839552?pr=5892#step:8:1780

you can use `@pytest.mark.single_process` to the failing test to make sure it is not run on multi-process run with pytest xdist.

nilearn/glm/_reporting_utils.py
Remi-Gau: I would say that the priority of correct behavior should be in a script not a notebook.

Remi-Gau:
Can you add tests for the different cases described in #5889?

nilearn/glm/tests/test_thresholding.py
```diff
@@ -620,46 +620,25 @@ def test_threshold_stats_img_surface_output_threshold_0(surf_img_1d):
...
-            surf_img_1d, height_control=height_control, threshold=threshold
-        )
-
-    n_warnings = len(
-        [x for x in warning_list if issubclass(x.category, FutureWarning)]
-    )
-    if height_control is None and threshold == 3.0:
-        assert n_warnings == 1
-    else:
-        assert n_warnings == 0
```
Remi-Gau, line 642: not sure I agree with this: the warning that the default will be changed should only be thrown when the default threshold value is used when height is None, and in no other case.

nilearn/glm/tests/test_thresholding.py
```diff
@@ -620,46 +620,25 @@ def test_threshold_stats_img_surface_output_threshold_0(surf_img_1d):
...
-        with warnings.catch_warnings(record=True) as warning_list:
-            cluster_level_inference(stat_img, threshold=threshold)
-
-        n_warnings = len(
-            [x for x in warning_list if issubclass(x.category, FutureWarning)]
-        )
-        if threshold == 3.0:
-            assert n_warnings == 1
-        else:
-            assert n_warnings == 0
```
Remi-Gau, line 665: same here: the warning that the default will be changed should only be thrown when the default threshold value is used

nilearn/glm/_reporting_utils.py
hndgzkn: I agree.

nilearn/glm/tests/test_thresholding.py
hndgzkn: But then it raises warnings when it should not. 
Because `glm.generate_report` has default value `3.09`. The user specifies `3` which is not default for `generate_report`. So it should not raise warning.

However, `generate_report` later calls `threshold_stats_img` which has default `3`. When it receives default threshold value, it raises warning that default will be changed. Which is not the expected behavior from `generate_report`.

I thought it might in anycase be good to inform user about the change independent of the value set by the user.

I have no strong opinion, I can change as you prefer. It is just that, if we raise warning depending on default value change, we will have conflicting behavior.

nilearn/glm/thresholding.py
hndgzkn: Correction: I've changed `generate_report` to raise default threshold warning and filter if raised from deeper code.

nilearn/glm/tests/test_thresholding.py
Remi-Gau: yeah I know the problem and that's why I had tried to put in place something to ignore subsequent warnings of the same type / message.

nilearn/glm/_base.py
```diff
@@ -529,26 +530,47 @@ def generate_report(
...
+            "\nFrom nilearn version>=0.15, "
+            "the default 'threshold' will be set to "
+            f"{DEFAULT_Z_THRESHOLD}.",
+            FutureWarning,
+            stacklevel=find_stack_level(),
+        )
+        # We silence further warnings about threshold:
+        #   it would throw one per contrast and
+        #   and also because threshold_stats_img and make_glm_report
+        #   have different defaults.
```
hndgzkn:
```suggestion
        # We silence further warnings about threshold.
```

nilearn/glm/tests/test_thresholding.py
hndgzkn: Yes, but it does not prevent. That is partially the source of this issue. For the above example, the filter does not work.

hndgzkn:
> Can you add tests for the different cases described in #5889?

I'll add those tests, once we finalize the changes in the PR.

nilearn/glm/tests/test_thresholding.py
Remi-Gau: > I thought it might in anycase be good to inform user about the change independent of the value set by the user.

Except that now `threshold_stats_img` will ALWAYS throw a future warning about the change of the default threshold even when users do not use the default and with no indication as to how to silence the warning except by adding some filter warning in their own code.

This will be annoying to the user and even to us because our own documentation will have quite few of those warnings.

nilearn/glm/_base.py
Remi-Gau: yeah that makes sense

Remi-Gau:
If that helps to make things simpler, we can set the default of `threshold` of generate_report for the GLM to None, but we just have to handle the case of a user passing `height=None` and `threshold=None`

hndgzkn:
> Can you add tests for the different cases described in #5889?

I put current behavior here:

```python
print("\ndefaults")
# Should get a SINGLE FutureWarning that the default threshold will change in 0.15.
# Currently getting one warning per contrasts,
# about the fact threshold is not used if 'height_control=None'
flm.generate_report(contrasts=contrasts)
```
```
/tmp/ipykernel_140723/4239375303.py:5: FutureWarning: 
From nilearn version>=0.15, the default 'threshold' will be set to 3.090232306167813.
  flm.generate_report(contrasts=contrasts)
/tmp/ipykernel_140723/4239375303.py:5: UserWarning: 
'threshold=3.09' is not used with 'height_control='fpr''.
'threshold' is only used when 'height_control=None'. 
Setting 'threshold' to None. 
  flm.generate_report(contrasts=contrasts)
```

----
```python
print("\nthreshold=3")
# changing threshold: should get a warning threshold is not used unless 'height_control=None'
# current warning says "Set 'threshold' to '3.09' to avoid this warning."
# which is misleading: should probably says "Set 'height_control' to 'None' to actually use this threshold."
flm.generate_report(contrasts=contrasts, threshold=3)
```

```
/tmp/ipykernel_140850/4208632181.py:5: FutureWarning: 
From nilearn version>=0.15, the default 'threshold' will be set to 3.090232306167813.
  flm.generate_report(contrasts=contrasts, threshold=3)
/tmp/ipykernel_140850/4208632181.py:5: UserWarning: 
'threshold=3' is not used with 'height_control='fpr''.
'threshold' is only used when 'height_control=None'. 
Setting 'height_control' to None. 
  flm.generate_report(contrasts=contrasts, threshold=3)
```

----

```python
print("\nheight_control=None")
# No warning but I would expect 
# to get a SINGLE FutureWarning that the default threshold will change in 0.15.
flm.generate_report(contrasts=contrasts, height_control=None)
```

```
/tmp/ipykernel_141022/1611863345.py:4: FutureWarning: 
From nilearn version>=0.15, the default 'threshold' will be set to 3.090232306167813.
  flm.generate_report(contrasts=contrasts, height_control=None)
```

----

```python
print(f"\nthreshold={DEFAULT_Z_THRESHOLD}")
# Now I am getting contradictory warnings that say to set threshold to 3.0 or 3.09 to avoid warnings
# Using the future threshold that will be used:
# - should not get a deprecation warning (that's the case)  
# - should a single warning that the thresold will not be used unless 'height_control=None'
flm.generate_report(contrasts=contrasts, threshold=DEFAULT_Z_THRESHOLD)
```

```
/tmp/ipykernel_141126/131692033.py:6: FutureWarning: 
From nilearn version>=0.15, the default 'threshold' will be set to 3.090232306167813.
  flm.generate_report(contrasts=contrasts, threshold=DEFAULT_Z_THRESHOLD)
/tmp/ipykernel_141126/131692033.py:6: UserWarning: 
'threshold=3.090232306167813' is not used with 'height_control='fpr''.
'threshold' is only used when 'height_control=None'. 
Setting 'height_control' to None. 
  flm.generate_report(contrasts=contrasts, threshold=DEFAULT_Z_THRESHOLD)
```
------

```python
print(f"\nheight_control=None, threshold={DEFAULT_Z_THRESHOLD}")
# no warning: as expected
flm.generate_report(contrasts=contrasts, height_control=None, threshold=DEFAULT_Z_THRESHOLD)
```

```
/tmp/ipykernel_141225/3052850624.py:3: FutureWarning: 
From nilearn version>=0.15, the default 'threshold' will be set to 3.090232306167813.
  flm.generate_report(contrasts=contrasts, height_control=None, threshold=DEFAULT_Z_THRESHOLD)
```

hndgzkn:
> If that helps to make things simpler, we can set the default of `threshold` of generate_report for the GLM to None, but we just have to handle the case of a user passing `height=None` and `threshold=None`

I think that would be a good solution. 

I can update the documentation saying that `threshold=None` by default, however, the default becomes `3.09` when `height_control` is set to None by the user.

This way we can avoid unnecessary warning and preserve the default behavior of `generate_report` method.

Remi-Gau:
> `threshold=None` by default, however, the default becomes `3.09` when `height_control` is set to None by the user.

let's try that

Remi-Gau:
This is what I get with the script in the issue:


```
defaults
/home/remi-gau/github/nilearn/nilearn/tmp/glm_report.py:21: FutureWarning: 
From nilearn version>=0.15, the default 'threshold' will be set to 3.090232306167813.
  flm.generate_report(contrasts=contrasts)
/home/remi-gau/github/nilearn/nilearn/tmp/glm_report.py:21: UserWarning: 
'threshold=3.09' is not used with 'height_control='fpr''.
'threshold' is only used when 'height_control=None'. 
Setting 'threshold' to None. 
  flm.generate_report(contrasts=contrasts)

threshold=3
/home/remi-gau/github/nilearn/nilearn/tmp/glm_report.py:27: UserWarning: 
'threshold=3' is not used with 'height_control='fpr''.
'threshold' is only used when 'height_control=None'. 
Setting 'height_control' to None. 
  flm.generate_report(contrasts=contrasts, threshold=3)

height_control=None

threshold=3.090232306167813
/home/remi-gau/github/nilearn/nilearn/tmp/glm_report.py:39: UserWarning: 
'threshold=np.float64(3.090232306167813)' is not used with 'height_control='fpr''.
'threshold' is only used when 'height_control=None'. 
Setting 'height_control' to None. 
  flm.generate_report(contrasts=contrasts, threshold=DEFAULT_Z_THRESHOLD)

height_control=None, threshold=3.090232306167813
```

nilearn/glm/_base.py
hndgzkn: At the end, I could not be sure if changing the default values of the function was a good idea. 
I'm leaving it as it is.

hndgzkn: Just warning the user. 

If `threshold` is set with a value other than default, it is favored and `height_control` is set to `None`. Otherwise, `threshold` is set to `None`.

nilearn/glm/_reporting_utils.py
hndgzkn: This filter is moved to generate_report.

nilearn/glm/thresholding.py
```diff
@@ -186,8 +167,15 @@ def cluster_level_inference(
     .. footbibliography::
 
     """
-    parameters = dict(**inspect.signature(cluster_level_inference).parameters)
-    warn_default_threshold(threshold, parameters["threshold"].default, 3.0)
+    # TODO (nilearn >= 0.15.0) remove
+    if threshold == 3.0:
```
hndgzkn, line 172: Is there an inconvenience just to check with value of default?

Remi-Gau: not that I can think of

nilearn/glm/tests/test_first_level.py
```diff
@@ -2710,3 +2712,60 @@ def test_first_level_from_bids_surface(tmp_path):
...
+    with pytest.warns(UserWarning, match="'height_control' to None"):
+        flm.generate_report(contrasts=contrasts, threshold=DEFAULT_Z_THRESHOLD)
+
+    # no warning expected
+    with warnings.catch_warnings(record=True) as warning_list:
+        flm.generate_report(
+            contrasts=contrasts,
+            height_control=None,
+            threshold=DEFAULT_Z_THRESHOLD,
+        )
```
Remi-Gau: I guess you cannot just do ``assert len(wartning_list) == 0`` because some other warnings are thrown, right?

nilearn/glm/tests/test_first_level.py
```diff
@@ -2710,3 +2712,60 @@ def test_first_level_from_bids_surface(tmp_path):
...
+        fmri_data[0], design_matrices=design_matrices[0]
+    )
+
+    contrasts = [
+        np.asarray([1, 0, 0]),
+        np.asarray([1, 1, 0]),
+        np.asarray([1, 1, 1]),
+    ]
+
+    # should return a single warning, set threshold to None
```
Remi-Gau: actually for this one and the following could you check the number of warning thrown like you do below?

nilearn/glm/tests/test_first_level.py
hndgzkn: Yes, exactly.

nilearn/glm/_base.py
bthirion: But it's true that we should not have a warning with the default behavior.

hndgzkn: Yes, I agree. However we might discuss and fix it in another PR. I couldn't be sure how to fix it in this PR. Because we are already warning the user that the value of threshold is going to change in version 0.15.0. In that case wouldn't it be better to wait at least until 0.15.0 for changing the threshold default value/behavior?

Remi-Gau:
will take care of the merge conflict

nilearn/glm/_base.py
```diff
@@ -529,26 +529,44 @@ def generate_report(
...
+        if threshold is not None and height_control is not None:
+            # TODO (nilearn >= 0.15.0) update 3.09 to DEFAULT_Z_THRESHOLD
+            # if user specifies threshold
+            if float(threshold) != 3.09:
+                warning_messages.append(
+                    f"\n'{threshold=}' is not used with '{height_control=}'."
+                    "\n'threshold' is only used when 'height_control=None'. "
+                    "\nSetting 'height_control' to None. "
+                )
+                height_control = None
```
Remi-Gau: actually I don't think we should set `height_control=None`, I prefer to have more friction if users want to NOT do any multiple test correction.

nilearn/glm/thresholding.py
```diff
@@ -412,6 +395,38 @@ def threshold_stats_img(
...
+        if float(threshold) != 3.0:
+            warnings.warn(
+                f"\n'{threshold=}' is not used with '{height_control=}'."
+                "\n'threshold' is only used when 'height_control=None'. "
+                "\nEither set 'height_control=None' or 'threshold=None' to "
+                "avoid this warning.",
+                UserWarning,
+                stacklevel=find_stack_level(),
+            )
+            height_control = None
```
Remi-Gau: actually I don't think we should set `height_control=None`, I prefer to have more friction if users want to NOT do any multiple test correction.

nilearn/glm/_base.py
hndgzkn: Then is it ok to set `threshold=None` inside the code as it will not be used when `height_control` is not None?

nilearn/glm/io.py
```diff
@@ -210,6 +210,12 @@ def save_glm_to_bids(
         check_parameter_in_allowed(key, report_kwargs, "Extra key-word")
         report_kwargs[key] = kwargs[key]
 
+    if (
+        report_kwargs["height_control"] is None
+        and report_kwargs["threshold"] is None
+    ):
+        report_kwargs["threshold"] = 3.09
+
```
Remi-Gau: needed as save_glm_to_bids relies on the deault of generate_report

nilearn/glm/io.py
```diff
@@ -334,7 +341,7 @@ def save_glm_to_bids(
             height_control=report_kwargs["height_control"],
         )
         table_details = clustering_params_to_dataframe(
-            report_kwargs["threshold"],
+            threshold,
```
Remi-Gau, line 355: this should actually fix a bug that did not update the threshold before saving the cluster table to disk

nilearn/glm/tests/test_first_level.py
```diff
@@ -2710,3 +2712,96 @@ def test_first_level_from_bids_surface(tmp_path):
...
+@pytest.mark.parametrize(
+    "kwargs",
+    [
+        ({}),
+        ({"height_control": None, "threshold": DEFAULT_Z_THRESHOLD}),
+    ],
+)
+def test_generate_report_default(kwargs):
+    """Make sure generate_report throws no warning by default.
+    or with with height=None and the future default threshold.
```
hndgzkn:
```suggestion
    """Make sure generate_report throws no warning by default,
    or when height_control=None and the future default threshold.
```

nilearn/glm/tests/test_first_level.py
```diff
@@ -2710,3 +2712,96 @@ def test_first_level_from_bids_surface(tmp_path):
...
+
+    with warnings.catch_warnings(record=True) as warning_list:
+        flm.generate_report(contrasts=contrasts, **kwargs)
+        assert len(warning_list) == 0
+
+
+@pytest.mark.slow
+def test_generate_report_height_none_future_default():
+    """Make sure generate_report a single FutureWarning
+    about the deprecation of the default threshold.
```
hndgzkn:
```suggestion
    """Make sure generate_report raises a single FutureWarning
    about the deprecation of the default threshold.
```

nilearn/glm/tests/test_first_level.py
```diff
@@ -2710,3 +2712,96 @@ def test_first_level_from_bids_surface(tmp_path):
...
+            [x for x in warning_list if issubclass(x.category, FutureWarning)]
+        )
+        assert n_warnings == 1
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize("threshold", [4, DEFAULT_Z_THRESHOLD])
+def test_generate_report_threshold_unused(threshold):
+    """Make sure generate_report return a single warning,
+    about threshold not being used.
```
hndgzkn:
```suggestion
    """Make sure generate_report raises a single warning,
    about threshold not being used.
```

nilearn/glm/_base.py
```diff
@@ -435,9 +435,11 @@ def generate_report(
             The background image for mask and stat maps to be plotted on upon.
             To turn off background image, just pass "bg_img=None".
 
-        threshold : :obj:`float`, default=3.09
+        threshold : :obj:`float` or :obj:`int` or None, default=None
             Cluster forming threshold in same scale as `stat_img` (either a
             t-scale or z-scale value). Used only if height_control is None.
+            If ``threshold`` is set to None when ``height_control`` is None,
+            ``threshold`` will be set to 3.09.
```
hndgzkn, line 446: This prevents user to set `threshold=None` when `height_control=None`. Isn't there a case that user might want to do so?

nilearn/glm/io.py
hndgzkn: Shouldn't this already be satisfied with the code in `generate_report`?

hndgzkn: I've checked the above of the code, I see, it is only the defaults of method signature.

nilearn/glm/_base.py
bthirion: I don't think so.

nilearn/glm/_base.py
bthirion: Good with me.

nilearn/glm/_base.py
Remi-Gau: In that case I don't even think the code would run without crashing majestically.

nilearn/glm/io.py
Remi-Gau: > Shouldn't this already be satisfied with the code in `generate_report`?

yes but before we get to generate report at the end of `save_glm_to_bids` there is a call `threshold_stat_map` and given that `generate_report` and `threshold_stat_map` have different default threshold value, I felt it was better to set it to actual value early

Remi-Gau:
was trying to get fewer warnings in the test and realized that we have a regression on using cut_coords in glm report introduced by https://github.com/nilearn/nilearn/pull/3452

```python
import numpy as np

    shapes=[(30,31,32,33)], rk=3
)

flm = FirstLevelModel(mask_img=mask).fit(
    fmri_data[0], design_matrices=design_matrices[0]
)

contrasts = [np.asarray([1, 0, 0]), np.asarray([1, 1, 0]), np.asarray([1, 1, 1])]

report = flm.generate_report(contrasts=contrasts, title="defaults", alpha=0.1, cut_coords=[3.0, 3.0, 3.0])
report.open_in_browser()
```

```
  File "/home/remi-gau/github/nilearn/nilearn/tmp/glm_report.py", line 20, in <module>
    report = flm.generate_report(contrasts=contrasts, title="defaults", alpha=0.1, cut_coords=[3.0, 3.0, 3.0])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/glm/_base.py", line 605, in generate_report
    mask_plot = _mask_to_plot(self, bg_img, cut_coords)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/glm/_reporting_utils.py", line 157, in _mask_to_plot
    mask_plot = figure_to_png_base64(plt.gcf())
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/reporting/utils.py", line 52, in figure_to_png_base64
    return base64.b64encode(_figure_to_png_bytes(fig)).decode()
                            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/reporting/utils.py", line 42, in _figure_to_png_bytes
    return _figure_to_bytes(fig, format="png", bbox_inches="tight")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/reporting/utils.py", line 25, in _figure_to_bytes
    fig.savefig(
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.12/site-packages/matplotlib/figure.py", line 3490, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.12/site-packages/matplotlib/backends/backend_qtagg.py", line 75, in print_figure
    super().print_figure(*args, **kwargs)
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.12/site-packages/matplotlib/backend_bases.py", line 2157, in print_figure
    self.figure.draw(renderer)
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.12/site-packages/matplotlib/artist.py", line 94, in draw_wrapper
    result = draw(artist, renderer, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.12/site-packages/matplotlib/artist.py", line 71, in draw_wrapper
    return draw(artist, renderer)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.12/site-packages/matplotlib/figure.py", line 3246, in draw
    artists = self._get_draw_artists(renderer)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.12/site-packages/matplotlib/figure.py", line 166, in _get_draw_artists
    ax.apply_aspect(locator(ax, renderer) if locator else None)
                    ^^^^^^^^^^^^^^^^^^^^^
  File "/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/displays/_slicers.py", line 1758, in _locator
    [[left_dict[axes], y0], [left_dict[axes] + width_dict[axes], y1]]
      ~~~~~~~~~^^^^^^
KeyError: <Axes: >
```

hndgzkn:
> was trying to get fewer warnings in the test and realized that we have a regression on using cut_coords in glm report introduced by #3452

Actually it seems to be an already existing bug in slicers. The below code also raises the same error:

```python
from nilearn.datasets import load_mni152_template
from nilearn.plotting import plot_roi

img = load_mni152_template()
# display is an instance of the XSlicer class
display = plot_roi(img, display_mode="z", cut_coords=[3, 3, 3])
```

I can fix this in https://github.com/nilearn/nilearn/pull/5886.

hndgzkn:
However inside `_mask_to_plot`, `plot_roi` is called with `display_mode=z`. I do not know if adding `display_mode` to `generate_report` is relevant. If not, it could be good to indicate that `cut_coords` specified for generate_report should comply with `display_mode=z`.

Remi-Gau:
> However inside `_mask_to_plot`, `plot_roi` is called with `display_mode=z`. I do not know if adding `display_mode` to `generate_report` is relevant.

yeah I just saw that too: for now let's remove the `cut_coords` from `_mask_to_plot`

> If not, it could be good to indicate that `cut_coords` specified for generate_report should comply with `display_mode=z`.

I thought we had a check for this.

Remi-Gau:
> I can fix this in #5886.

can it be fixed out side of that PR?

nilearn/glm/io.py
```diff
@@ -378,7 +385,12 @@ def save_glm_to_bids(
     # as only the surface data but no mesh
     # was saved to disk.
     if model._is_volume_glm():
-        glm_report = model.generate_report(**kwargs)
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore",
+                message="No contrast passed during report generation.",
+            )
```
Remi-Gau, line 406: contrast are not passed on purpose, so let's silence warnings that users have no control over

hndgzkn:
> can it be fixed out side of that PR?

Yes, it can. But as that PR aims to check `cut_coords` compatibility with `display_mode` in the relevant slicer class, it seems to me to be a good place to fix that. We can open an issue for this bug, and reference to the issue in the PR.

Remi-Gau:
> > can it be fixed out side of that PR?
> 
> Yes, it can. But as that PR aims to check `cut_coords` compatibility with `display_mode` in the relevant slicer class, it seems to me to be a good place to fix that. We can open an issue for this bug, and reference to the issue in the PR.

OK add it in the same PR then

Remi-Gau:
> > If not, it could be good to indicate that `cut_coords` specified for generate_report should comply with `display_mode=z`.
> 
> I thought we had a check for this.

@hndgzkn 
I will add some checks of the inputs of generate_report in the current PR

hndgzkn:
> > However inside `_mask_to_plot`, `plot_roi` is called with `display_mode=z`. I do not know if adding `display_mode` to `generate_report` is relevant.
> 
> yeah I just saw that too: for now let's remove the `cut_coords` from `_mask_to_plot`

I just wrote quickly without checking `generate_report`'s signature. `display_mode` already exists there. Could setting `display_mode` to `_mask_to_plot` be an option, or is it irrelevant? In that case, I would remove it.

> > If not, it could be good to indicate that `cut_coords` specified for generate_report should comply with `display_mode=z`.
> 
> I thought we had a check for this.

My comment on this makes no sense after seeing that `display_mode` was already in `generate_report` signature.

hndgzkn:
> @hndgzkn I will add some checks of the inputs of generate_report in the current PR

Which ones are you thinking of?

Remi-Gau:
> > @hndgzkn I will add some checks of the inputs of generate_report in the current PR
> 
> Which ones are you thinking of?

height_control, cluster_threshold, plot_type

nilearn/glm/_reporting_utils.py
```diff
@@ -101,7 +170,7 @@ def _load_bg_img(bg_img, is_volume_glm):
         )
 
 
-def _mask_to_plot(model, bg_img, cut_coords):
```
Remi-Gau, line 104: removing cut_coords because the mask image is always plotted using display_mode='z', so the cut_coords passed to generate_report may not always be compatible when used in _mask_to_plot

this could be fixed but I think this is beyond the scope of the PR

nilearn/glm/_base.py
```diff
@@ -519,36 +525,22 @@ def generate_report(
             Contains the HTML code for the :term:`GLM` report.
 
         """
+        check_generate_report_input(
+            height_control, cluster_threshold, min_distance, plot_type
+        )
         check_params(locals())
```
hndgzkn, line 535: Not for this PR but in general, could we write a generic parameter validation function as a decorator that compares type annotations in method/function signature and the specified values and their types?

nilearn/glm/_reporting_utils.py
```diff
@@ -30,6 +31,89 @@
...
+            "\n'threshold' is only used when 'height_control=None'. "
+            "\n'threshold' was set to 'None'. "
+        )
+        threshold = None
+
+    if cut_coords is not None and plot_type == "glass":
+        warning_messages.append(
+            f"\n'{cut_coords=}' is not used with '{plot_type=}'."
+            "\n'cut_coords' is only used when 'plot_type='slice''. "
+            "\n'cut_coords' was set to None. "
```
hndgzkn, line 110: The docstring of `generate_report` says that, when `plot_type` is `glass`, `display_mode` is `ortho`. But does not tell that `cut_coords` should be set to `None`. May be this could also be indicated in the docstring.

nilearn/glm/io.py
hndgzkn: This import could be moved to module level, as the cyclic dependency is removed.

nilearn/glm/_base.py
Remi-Gau: check_params does this to an extend: it ensures that when a parameter has a certain name it is of a certain type

https://github.com/nilearn/nilearn/blob/06b804d39ac8993be5470bf1b2db1edf1c23d1c3/nilearn/_utils/param_validation.py#L216

Remi-Gau: I would not be opposed to have something more generic that makes use of type annotation to do input validation at run time.

Remi-Gau: note they are some private utilities in sklearn that do some input validation as well : https://github.com/scikit-learn/scikit-learn/blob/a4db9a9dc45bbd39e397ea4ebf9c5dcc6a658500/sklearn/utils/_param_validation.py#L150

But it does not just rely on the type annotation.

Remi-Gau: also I am pretty sure we cannot be the first to think of using type annotation at run time, so let's be careful about not reinventing the wheel

nilearn/glm/io.py
Remi-Gau: yeah I thought so as well but forgot to move it up

nilearn/glm/_base.py
hndgzkn: > note they are some private utilities in sklearn that do some input validation as well : https://github.com/scikit-learn/scikit-learn/blob/a4db9a9dc45bbd39e397ea4ebf9c5dcc6a658500/sklearn/utils/_param_validation.py#L150
> 
> But it does not just rely on the type annotation.

I think this can be a good starting point.

Remi-Gau: in any case what could be done before would be to apply the annotations from nilearn.typing to the functions / methods signature
would allow to at least to benefit from those during static analysis

PR #5893, opened 2025-12-11
[FIX] fix errors using pre-release version of pandas

Remi-Gau:
- Closes #

- fix string type error in pre-release (I suspect something with pandas==3.0.0rc0) - see https://pandas.pydata.org/docs/dev/whatsnew/v2.3.3.html#improvements-and-fixes-for-the-stringdtype

- https://github.com/nilearn/nilearn/actions/runs/20106610106/job/57692229636#step:8:13945

```
nilearn/datasets/tests/test_func.py::test_fetch_abide_pcp[True] - TypeError: Cannot interpret '<StringDtype(storage='python', na_value=nan)>' as a data type
```

- https://github.com/nilearn/nilearn/actions/runs/20106610106/job/57692229636#step:8:13941

```
https://github.com/nilearn/nilearn/actions/runs/20106610106/job/57692229636#step:8:13941
```

### DIFF
--- a/nilearn/_utils/tests/test_data_gen.py
+++ b/nilearn/_utils/tests/test_data_gen.py
@@ -7,7 +7,7 @@ import pandas as pd
 import pytest
 from nibabel import load
 from numpy.testing import assert_almost_equal
-from pandas.api.types import is_numeric_dtype, is_object_dtype
+from pandas.api.types import is_numeric_dtype, is_string_dtype
 from pandas.testing import assert_frame_equal
 
 from nilearn._utils.data_gen import (
@@ -68,7 +68,7 @@ def test_basic_paradigm(have_spaces):
     events = basic_paradigm(condition_names_have_spaces=have_spaces)
 
     assert events.columns.equals(pd.Index(["trial_type", "onset", "duration"]))
-    assert is_object_dtype(events["trial_type"])
+    assert is_string_dtype(events["trial_type"])
     assert is_numeric_dtype(events["onset"])
     assert is_numeric_dtype(events["duration"])
     assert events["trial_type"].str.contains(" ").any() == have_spaces


nilearn/datasets/_utils.py
hndgzkn, line 409: Is this correct? As far as I understand, this is not the only possible type.

nilearn/datasets/_utils.py
```diff
@@ -449,7 +450,10 @@ def _filter_column(array, col, criteria):
 
     # Handle strings with different encodings
     if isinstance(criteria, (str, bytes)):
-        criteria = np.array(criteria).astype(array[col].dtype)
+        dtype = array[col].dtype
+        if isinstance(dtype, pd.StringDtype):
+            dtype = "str"
+        criteria = np.array(criteria).astype(dtype)
```
hndgzkn, line 456: Is there a possibility that there are None values in this dataframe? 
If we use `dtype='str'`, None values become `'None'`. 
`dtype=object` might preserve data types.

hndgzkn (approved):

nilearn/datasets/_utils.py
Remi-Gau: If you have a single None in the series then it is not a StringDtype, so we won't cast to a string.

```python
import pandas as pd
from pandas.api.types import is_numeric_dtype, is_string_dtype

df = pd.DataFrame({'col1': [None, 1],
                   "col2": [None, None], 
                   'col3': ["str", None],
                   'col4': [1, 2],
                   'col5': ["foo", "bar"],
                   })

for x in df.columns:
    print()
    print(df[x].values)
    print(df[x].dtype)
    print(isinstance(df[x].dtype, pd.StringDtype))
    print(is_string_dtype(df[x]))
```

```
[nan  1.]
float64
False
False

[None None]
object
False
False

['str' None]
object
False
False

[1 2]
int64
False
False

['foo' 'bar']
object
False
True
```

nilearn/datasets/_utils.py
Remi-Gau: changed to array-like: the code tends to be built mostly around numpy arrays and pandas dataframe though

PR #5894, opened 2025-12-12
[MAINT] skip failing shelving test on windows

Remi-Gau:
- Closes none but skips a failing test on windows for now:
  - see on main

- skips a failing test on windows
  - this checks a non public way of using shelving with NiftiMaskers
  - the test could possibly pass if not run in parallel with pytest xdist: to be checked later (AKA after the release: easier for now to just skip it)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,7 +82,7 @@ repos:
 
     # Lint and format Python code
 -   repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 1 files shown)


.github/workflows/test_with_tox.yml
Remi-Gau, line 1: revert changes in this file before merge

Remi-Gau:
> The failure seems unrelated.

yup but still marked the test as slow

PR #5901, opened 2025-12-15
[ENH] save surface cluster tables when saving GLM

Remi-Gau:
- Closes none

- save cluster results table to disk for surface GLM


Remi-Gau:
> I guess this needs to be tested (codecov failure)? Otherwise LGTM!

I updated the test to make sure the new behavior is enforced

PR #5905, opened 2025-12-16
[FIX] allow user to override default in matrix plotting and `BaseSlicer.title` via `kwargs`

### LINKED ISSUE #5904: [ENH] Plotting should not enforce kwargs if they exist

In many nilearn plotting functions there's the option to specify keyword arguments that are utimately passed to the plotting backend (`imshow`, ...). As an example, `plot_matrix` does this:
```python
display = axes.imshow(
    mat, aspect="equal", interpolation="nearest", cmap=cmap, **kwargs
)
```
If an already specified arguments, like `interpolation`, is also specified in the kwargs a syntax error is thrown(`SyntaxError: keyword argument repeated: interpolation`).

I think arguments passed to `kwargs` should replace those by default.

### Benefits to the change

I took the example of `interpolation` because it can be desirable to set interpolation to `'none'` when displaying a matrix if matplotlib's backend is vectorial (e.g. pdf), it reduces file size. That's only an example though, by default matplotlib does interpolation.

```python
plotting.plot_matrix(mat, interpolation='none') # replaces interpolation='nearest' in axes.imshow

```

Remi-Gau:
- Closes #5904

- ensure that kwargs can override the defaults used internally by Nilearn for matrix plotting and BaseSlicer.title

### DIFF
--- a/nilearn/plotting/image/img_plotting.py
+++ b/nilearn/plotting/image/img_plotting.py
@@ -53,10 +53,7 @@ from nilearn.plotting._utils import (
 )
 from nilearn.plotting.displays import get_projector, get_slicer
 from nilearn.plotting.displays._slicers import save_figure_if_needed
-from nilearn.plotting.image.utils import (
-    MNI152TEMPLATE,
-    load_anat,
-)
+from nilearn.plotting.image.utils import MNI152TEMPLATE, load_anat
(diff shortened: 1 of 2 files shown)


hndgzkn (approved):
LGTM.

PR #5913, opened 2025-12-19
[DOC] update cut_coords in examples

Remi-Gau:
- Closes none but should fix doc build on main

- range is not accepted as cut_coord

### DIFF
--- a/examples/01_plotting/plot_3d_map_to_surface_projection.py
+++ b/examples/01_plotting/plot_3d_map_to_surface_projection.py
@@ -136,7 +136,7 @@ plot_stat_map(
     stat_map_img=stat_img,
     display_mode="x",
     threshold=1.0,
-    cut_coords=range(0, 51, 10),
+    cut_coords=list(range(0, 51, 10)),
     title="Slices",
 )
 


examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -136,7 +136,7 @@
     stat_map_img=stat_img,
     display_mode="x",
     threshold=1.0,
-    cut_coords=range(0, 51, 10),
+    cut_coords=list(range(0, 51, 10)),
```
Remi-Gau, line 139: - @hndgzkn this was introduced by https://github.com/nilearn/nilearn/pull/5886

I wonder if we should allow iterator and generator to avoid a regression like this one

hndgzkn: I think it depends. Iterator or generator was not mentioned in the doc as far as I remember. So I followed the specification in docstring to restrict the types. Is it likely/necessary to specify an iterator? 

I would actually be in favor of limiting types of parameters unless necessary. When we enlarge possible parameter types, that requires more checks.

Remi-Gau: agreed so let's leave things like this for now

PR #5915, opened 2025-12-22
[MAINT] rename private functions, fix doc strings

Remi-Gau:
- Closes none

- nilearn/glm/_reporting_utils.py : remove leading underscore from some functions imported in another module
- fix some doc strings flagged by maint_tools/check_docstrings.py

### DIFF
--- a/nilearn/glm/tests/test_first_level.py
+++ b/nilearn/glm/tests/test_first_level.py
@@ -1523,6 +1523,7 @@ def test_first_level_residuals_errors(shape_4d_default):
         model._get_element_wise_model_attribute("foo", True)
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize(
     "shapes",
     [


nilearn/datasets/neurovault.py
```diff
@@ -723,7 +723,7 @@ class ResultFilter:
 
     Parameters
     ----------
-    query_terms : dict, optional
+    query_terms : :obj:`dict`, optional
```
hndgzkn: It is a bit confusing to have `Parameters` section in class level docstring. There is also `Attributes` section just below. Is there a special reason for that?

hndgzkn: Another question is when do we use `optional` and when `default=None`?  Do we have a convention for that? While trying to update docstrings for slicers, I couldn't be sure which one to use when?

hndgzkn (approved):
LGTM.

nilearn/datasets/neurovault.py
Remi-Gau: > Another question is when do we use `optional` and when `default=None`? Do we have a convention for that? While trying to update docstrings for slicers, I couldn't be sure which one to use when?

in general I think we try to be explicit and say "default=..." rather than just "optional" because "optional" could also be used whatever the default value is.

Remi-Gau: > It is a bit confusing to have `Parameters` section in class level docstring. There is also `Attributes` section just below. Is there a special reason for that?

given we follow the numpydoc convention I am pretty sure this is where constructors parameters and attributes should be described: https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes

Remi-Gau: > > Another question is when do we use `optional` and when `default=None`? Do we have a convention for that? While trying to update docstrings for slicers, I couldn't be sure which one to use when?
> 
> in general I think we try to be explicit and say "default=..." rather than just "optional" because "optional" could also be used whatever the default value is.

There is a script to help detect some missing defaults: maint_tools/missing_default_in_docstring.py

but this is more systematically applied to the public API than the private one.

hndgzkn: > given we follow the numpydoc convention I am pretty sure this is where constructors parameters and attributes should be described: https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes

Yes, sorry, it's my ignorance. I was documenting in `__init__`.

PR #5920, opened 2025-12-24
[FIX] Improve and check docstrings of slicers and projectors

### LINKED ISSUE #5910: Improvement for slicers / projectors
the doc strings at the class level in displays are a bit of a mess...

2 things for later PRs:

- where possible we should enforce a certain type for attributes 

  to avoid the current behavior

  ```python
  from nilearn.plotting .displays import XSlicer
  import numpy as np
  
  for cut_coords in [[1,2,3], (1,2,3), np.asarray((1,2,3))]:
      x_slicer = XSlicer(cut_coords=cut_coords)
      print(x_slicer.cut_coords.__class__.__name__)
  ```

  ```
  list
  tuple
  ndarray
  ```

  knowing what we are dealing with should help reflect about the code in general

- we should have some checks to make sure all public attributes of public classes in displays are documented properly (we do that for our estimators so we could reuse one of the checks to do it)

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5886#discussion_r2626671207_

### LINKED ISSUE #5911: Add support for dictionnary for cut_coords in plotting functions using mosaic display
@Remi-Gau While adding tests in `test_display` module, I've realized the test below:

https://github.com/nilearn/nilearn/blob/73f30611318b3586ae155b086d075f8a80abfabd/nilearn/plotting/tests/test_displays.py#L283-L291

I think MosaicSlicer accepts a dictionary. However when we init the slicer with `init_with_figure`, it raises error. Seems to be a bug then?

_Originally posted by @hndgzkn in https://github.com/nilearn/nilearn/pull/5886#discussion_r2631079818_

hndgzkn:
- closes #5910 
- closes #5911
- Completes #5928

- Fix missing docstrings and docstring bugs for `nilearn.plotting.displays._slicers` and `nilearn.plotting.displays._projectors`
- Add test to make sure that:
  - public attributes are documented
  - `__init__` parameters are documented
  - each public function and parameters are documented
  - no duplicate parameter documentation
- Add dictionary support for `MosaicSlicer`
- Replace `optional` in parameter definitions with default value

### DIFF
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -1702,6 +1702,16 @@ rect : 4- :obj:`tuple`
     Position of axes within the figure.
 """
 
+docdict["slicer_init_parameters_partial"] = f"""
+{docdict["axes"]}
+{docdict["black_bg"]}
+{docdict["brain_color"]}
+kwargs : :obj:`dict`
+    Extra keyword arguments are passed to
+    :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
+    direction.
+"""
+
 docdict["projector_init_parameters"] = f"""
 Parameters
 ----------
--- a/nilearn/plotting/displays/_slicers.py
+++ b/nilearn/plotting/displays/_slicers.py
@@ -1255,18 +1255,7 @@ class OrthoSlicer(_MultiDSlicer):
         The world coordinates ``(x, y, z)`` of the point where the cut is
         performed.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------
@@ -1470,18 +1459,7 @@ class TiledSlicer(_MultiDSlicer):
         The world coordinates ``(x, y, z)`` of the point where the cut is
         performed.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------
@@ -2046,18 +2024,7 @@ class XSlicer(BaseStackedSlicer):
         direction 'x'.
         If ``None``, 7 cut positions are calculated automatically.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------
@@ -2105,18 +2072,7 @@ class YSlicer(BaseStackedSlicer):
         direction 'y'.
         If ``None``, 7 cut positions are calculated automatically.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------
@@ -2164,18 +2120,7 @@ class ZSlicer(BaseStackedSlicer):
         direction 'z'.
         If ``None``, 7 cut positions are calculated automatically.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------
@@ -2222,18 +2167,7 @@ class XZSlicer(OrthoSlicer):
         The world coordinates ``(x, z)`` of the point where the cut is
         performed.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------
@@ -2279,18 +2213,7 @@ class YXSlicer(OrthoSlicer):
         The world coordinates ``(x, y)`` of the point where the cut is
         performed.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------
@@ -2336,18 +2259,7 @@ class YZSlicer(OrthoSlicer):
         The world coordinates ``(y, z)`` of the point where the cut is
         performed.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------
@@ -2398,18 +2310,7 @@ class MosaicSlicer(BaseSlicer):
         ('x', 'y', 'z') and the values are sequences holding the cut
         coordinates.
 
-    %(axes)s
-        default=None
-
-    %(black_bg)s
-        default=False
-
-    %(brain_color)s
-
-    kwargs : :obj:`dict`
-        Extra keyword arguments are passed to
-        :class:`~nilearn.plotting.displays.CutAxes` used for plotting in each
-        direction.
+    %(slicer_init_parameters_partial)s
 
     Attributes
     ----------


nilearn/plotting/displays/_projectors.py
```diff
@@ -32,14 +34,43 @@ class OrthoProjector(OrthoSlicer):
         # display is an instance of the OrthoProjector class
         display = plot_glass_brain(img, display_mode="ortho")
 
+    Parameters
+    ----------
+    cut_coords : 3 sequence of :obj:`float` or :obj:`int` or None
+        The world coordinates ``(x, y, z)`` of the point where the cut is
+        performed.
```
hndgzkn: This seems to be not used for projectors neither in the constructor nor in `find_cut_coord`.

As far as I understand, cut_coords in each direction is `None` for all projectors.

nilearn/plotting/displays/_slicers.py
hndgzkn: Should `draw_cross` be implemented for all slicers?  It seems to be only implemented for `OrthoSlicer`. 

If we move method definition from `OrthoSlicer` to `BaseSlicer` and rename `_directions` for stacked slicers to `_cut_displayed`, it will run for all except `MosaicSlicer`, which could also be fixed, I guess, easily.  

Should it also be implemented for projectors?

bthirion:
LMK if you need a review. Do you understand the origin of the doc generation issue ?

hndgzkn:
> LMK if you need a review. 

Not yet, thank you. I'll let you know when ready.

> Do you understand the origin of the doc generation issue ?

It is a typo, it should be fixed now.

hndgzkn:
> LMK if you need a review. 

This is now ready for review. (I do not wait for the tests to pass, recent changes are all on docstrings. )

nilearn/plotting/displays/tests/test_displays.py
```diff
@@ -251,16 +259,25 @@ def test_mosaic_slicer_img_none_false(cut_coords, img):
...
+        {"x": [3, 5], "y": [3, 6]},
+        {"x": [3, 5], "y": [3, 6], "z": [1, 2], "k": [6, 8, 8]},
+        {"x": [3, 5], "y": [3, 6], "z": 7},
+    ],
+)
 def test_mosaic_slicer_wrong_inputs(cut_coords):
     """Tests that providing wrong inputs raises a ``ValueError``."""
     with pytest.raises(
-        ValueError, match="MosaicSlicer plotting expects a number, list"
+        ValueError, match="MosaicSlicer plotting expects a number, 3D list"
```
bthirion: Sorry, wdym with "3D list" ?

nilearn/plotting/displays/_projectors.py
```diff
@@ -32,14 +34,41 @@ class OrthoProjector(OrthoSlicer):
...
+        The world coordinates of the point where the cut is performed.
+
     axes : :obj:`dict` of :class:`~nilearn.plotting.displays.GlassBrainAxes`
-        The 3 axes used to plot each view ('x', 'y', and 'z').
+        The axes used for plotting in each direction ('x', 'y' and 'z' here).
 
     frame_axes : :class:`~matplotlib.axes.Axes`
         The axes framing the whole set of views.
 
+    rect : 4 :obj:`tuple`
```
bthirion: Is it possible to put a hyphen, i.e. 4-tuple ?
If possible complete the docstring.

nilearn/plotting/displays/_projectors.py
```diff
@@ -51,14 +80,53 @@ def find_cut_coords(
...
+
+        cut_coords : sequence of :obj:`float` or :obj:`int`, or None, \
+                    default=None
+            The world coordinates of the point where the cut is performed.
+
+        Returns
+        -------
+        cut_coords : :obj:`tuple` of :obj:`float` or :obj:`int`
+            tuple of `None`s of length corresponding to the number of
+        directions of this projector.
```
bthirion:
```suggestion
            directions of this projector.
```

nilearn/plotting/displays/_projectors.py
```diff
@@ -307,13 +379,40 @@ class XProjector(OrthoProjector):
...
+
     axes : :obj:`dict` of :class:`~nilearn.plotting.displays.GlassBrainAxes`
-           The axes used for plotting.
+        The axes used for plotting in each direction ('x' here).
 
     frame_axes : :class:`~matplotlib.axes.Axes`
-                 The axes framing the whole set of views.
+        The axes framing the whole set of views.
+
+    rect : 4 :obj:`tuple`
```
bthirion: same as above

nilearn/plotting/displays/tests/test_displays.py
hndgzkn: A list with 3 elements. Shall I say 3-list, or write more explicitly?

nilearn/plotting/displays/_projectors.py
hndgzkn: There are also parts that I use 3 sequence, 2 sequence, etc. Shall I change them as 3-sequence, etc?

hndgzkn:
Working on _slicers and _projectors, I think there are some points that are confusing and could be improved later.

Projectors inherit from slicers, hence requires `cut_coords` parameter but independent of the value specified, it is set to a tuple of None depending on projector type. 
`find_cut_coords`, `draw_cross` functions are not used for projectors. 

`draw_cross` is only used by two slicers, but it exists in other types of projectors and slicers with empty code.

nilearn/plotting/displays/_projectors.py
```diff
@@ -32,14 +34,42 @@ class OrthoProjector(OrthoSlicer):
         # display is an instance of the OrthoProjector class
         display = plot_glass_brain(img, display_mode="ortho")
 
+    Parameters
```
Remi-Gau: Can you double check the description above:

```
This visualization mode can be activated from
:func:`~nilearn.plotting.plot_glass_brain`, by setting
``display_mode='ortho'``:
```

ortho can be used by many other plotting functions than just the galss brain one, no?

nilearn/plotting/displays/_projectors.py
```diff
@@ -32,14 +34,42 @@ class OrthoProjector(OrthoSlicer):
...
+    %(brain_color)s
+
+    kwargs : :obj:`dict`
+        Extra keyword arguments are passed to
+        :class:`~nilearn.plotting.displays.GlassBrainAxes` used for plotting in
+        each direction.
+
     Attributes
     ----------
+    cut_coords : (None, None, None)
```
Remi-Gau, line 40: cut_coords is not `(None, None, None)` if some value was passed, no?

nilearn/plotting/displays/_projectors.py
```diff
@@ -32,14 +34,42 @@ class OrthoProjector(OrthoSlicer):
...
+        The axes used for plotting in each direction ('x', 'y' and 'z' here).
 
     frame_axes : :class:`~matplotlib.axes.Axes`
         The axes framing the whole set of views.
 
+    rect : 4- :obj:`tuple`
+        Position of axes within the figure.
+
+    _cut_displayed : "yxz"
+        The directions of cuts for ``OrthoProjector``.
```
Remi-Gau: do we want to document private attributes?

nilearn/plotting/displays/_projectors.py
```diff
@@ -307,13 +377,41 @@ class XProjector(OrthoProjector):
...
+
+    kwargs : :obj:`dict`
+        Extra keyword arguments are passed to
+        :class:`~nilearn.plotting.displays.GlassBrainAxes` used for plotting in
+        each direction.
+
     Attributes
     ----------
+    cut_coords : (None,)
+        The world coordinates of the point where the cut is performed.
```
Remi-Gau, line 359: smae here: cut_coords is not always (None,)

```python
from nilearn.plotting.displays import XProjector

proj = XProjector(cut_coords=tuple([5]))
print(proj.cut_coords)
```

nilearn/plotting/displays/_projectors.py
```diff
@@ -307,13 +377,41 @@ class XProjector(OrthoProjector):
...
 
     frame_axes : :class:`~matplotlib.axes.Axes`
-                 The axes framing the whole set of views.
+        The axes framing the whole set of views.
+
+    rect : 4- :obj:`tuple`
+        Position of axes within the figure.
+
+    _cut_displayed : "x"
+        The direction of cuts for ``XProjector``.
```
Remi-Gau: same about documenting private attr

nilearn/plotting/displays/_projectors.py
```diff
@@ -719,14 +1136,42 @@ class RProjector(OrthoProjector):
...
+
+    %(black_bg)s
+        default=False
+
+    %(brain_color)s
+
+    kwargs : :obj:`dict`
+        Extra keyword arguments are passed to
+        :class:`~nilearn.plotting.displays.GlassBrainAxes` used for plotting in
+        each direction.
```
Remi-Gau: if you got many repeats of the same parameters you can also combine them into a single entry in the _utils.doc: see https://github.com/nilearn/nilearn/blob/502bb579329dc0c683242bf4346dea1760d095b6/nilearn/_utils/docs.py#L1508

will let you decide if it makes sense to do it here

nilearn/plotting/displays/_projectors.py
hndgzkn: Yes, but this is for `OrthoProjector` in its own docstring. Or I misunderstood your point.

nilearn/plotting/displays/_slicers.py
```diff
@@ -41,23 +41,24 @@ class BaseSlicer:
...
-        transparent colorbars).
+    rect : 4- :obj:`tuple`
+        Position of axes within the figure.
 
     """
 
     # This actually encodes the figsize for only one axe
     _default_figsize: ClassVar[list[float]] = [2.2, 2.6]
-    _axes_class = CutAxes
+    _axes_class: ClassVar[CutAxes] = CutAxes  # type: ignore[assignment, misc]
```
Remi-Gau:
```suggestion
    _axes_class: type[CutAxes] = CutAxes
```

quickly check locally and that should keep mypy happy

nilearn/plotting/displays/_slicers.py
```diff
@@ -1102,29 +1128,36 @@ def savefig(self, filename, dpi=None, **kwargs):
...
 
-class _ThreeDSlicer(BaseSlicer):
-    _cut_displayed: ClassVar[str] = "yxz"
-    _axes_class = CutAxes
+class _MultiDSlicer(BaseSlicer):
+    # This should be set by each inheriting Slicer
+    _cut_displayed: ClassVar[str] = ""
 
-    @classmethod
     @fill_doc  # the fill_doc decorator must be last applied
```
Remi-Gau: I think the fill_doc decorator can now be removed, no?

nilearn/plotting/displays/tests/test_displays.py
```diff
@@ -280,7 +297,13 @@ def expected_cuts(cut_coords):
 
 
 @pytest.mark.parametrize(
-    "cut_coords", [(1, 1, 1), 5, {"x": [10, 20], "y": [30, 40], "z": [15, 16]}]
+    "cut_coords",
+    [
+        (1, 1, 1),
+        5,
+        {"x": [10, 20], "y": [30, 40], "z": [15, 16]},
```
Remi-Gau, line 304: can you update examples/01_plotting/plot_demo_more_plotting.py
to show that we can use dict with mosaic display?

it'd be nice to have a real world example of this

Remi-Gau:
Can you add an entry in the changelog at least regarding dict and mosaic slicer?

nilearn/plotting/displays/_projectors.py
hndgzkn: Originally yes, it used to set the value that is specified. But that value is not used as far as I see. 

The projectors are initialized only using `init_with_figure` which calls, `find_cut_coords`. And for projectors `find_cut_coords` always returns a tuple of `None`s.

The below code displays the following images:
```python
from nilearn.plotting.image.img_plotting import plot_glass_brain
from nilearn import datasets

plot_glass_brain(stat_img, threshold=3, cut_coords=(1,2,3))
plot_glass_brain(stat_img, threshold=3, cut_coords=(100,40,40))
```
glass_brains

The logic for sanitizing `cut_coords` is both inside `sanitize_cut_coords` and `find_cut_coords`. So I've replaced `sanitize_cut_coords` in `init_axes` with `find_cut_coords`. This way, the correct `cut_coords` is set upon initialization.

Remi-Gau: thanks for the explanation because this was confusing

I feel part of this should be somewhat explained in the doc string of the attributes, no?

nilearn/plotting/displays/_projectors.py
Remi-Gau: I mean that some of the other projectors just have things like wiht no mention of a specific function

```python
This visualization mode can be activated by setting display_mode='lyr'
```

Remi-Gau: also many other projector though have specific mention of

```rst
:func:`~nilearn.plotting.plot_glass_brain`
```

for example

```python
    """The ``YZProjector`` class enables to combine coronal and axial views \
    on the same figure through 2D projections with \
    :func:`~nilearn.plotting.plot_glass_brain`.

   ...
   """
```

But I think they should be more like:

```python
    """The ``YZProjector`` class enables to combine coronal and axial views \
    on the same figure through 2D projections.

   ...
   """
```

Remi-Gau: Also do you think we could get rid of the `?` in some of the doc strings

for example

```
The ``LYRProjector`` class enables ? visualization \
    on the same figure through 2D projections with \
    :func:`~nilearn.plotting.plot_glass_brain`.
```

hndgzkn: > ortho can be used by many other plotting functions than just the galss brain one, no?

Projections are only initialized from `plot_glass_brain` function. I think that's the reason it is mentioned. But I agree, no need to mention here.

nilearn/plotting/displays/_projectors.py
hndgzkn: I had put "Not used for projectors." for `cut_coords` in `Parameters` section on class definition thinking that it will be the value specified by user, so better to document it there. In `Attributes` section I've just put the value tuple to `None`s. Would that be ok, or would you prefer to be more explicit?

nilearn/plotting/displays/_projectors.py
hndgzkn: I do not remember why exactly I wanted to document this. It was in purpose, but now I think I can delete it :)

nilearn/plotting/displays/_projectors.py
Remi-Gau: this should be OK: thanks

nilearn/plotting/displays/_projectors.py
hndgzkn: > Also do you think we could get rid of the `?` in some of the doc strings
> 
> for example
> 
> ```
> The ``LYRProjector`` class enables ? visualization \
>     on the same figure through 2D projections with \
>     :func:`~nilearn.plotting.plot_glass_brain`.
> ```

Shall I only delete it, or is there an explanation that I can replace? I think the person writing the docstring was not sure what to write.

Remi-Gau: I'd say that when there is a way to definitely know which view (coronal, sagital, axial) will be used then it should be mentioned otherwise, let's delete it

hndgzkn: I've deleted, but might be good to have a second look at that.

nilearn/plotting/displays/tests/test_displays.py
hndgzkn: Added the below 

```python
# It is also possible to specify cut positions in each direction.
# In this situation, we still set ``display_mode='mosaic'``, but
# we specify the cut positions of slices as a dictionary with keys
# "x", "y", "z" and values as list of cut positions in that direction.

plotting.plot_stat_map(
    stat_img,
    display_mode="mosaic",
    cut_coords={"x": [42, 18, -45], "y": [-52, -28, 0], "z": [55, 67, 73]},
    title="display_mode='mosaic' with cut_coords as dict",
)
```

which gives 
example

nilearn/plotting/displays/_projectors.py
hndgzkn: If I do that, would 

```   
    %(axes)s
        default=None

    %(black_bg)s
        default=False

    %(brain_color)s

```

still work?

nilearn/plotting/displays/_projectors.py
bthirion: I'd do so indeed.

nilearn/plotting/displays/_projectors.py
Remi-Gau: not sure what you mean

nilearn/plotting/displays/tests/test_displays.py
hndgzkn: I've replaced with a` list, tuple or array of 3 numbers`.

nilearn/plotting/displays/_projectors.py
hndgzkn: Sorry, I was asking if `fill_doc` would still replace values if I used `%()s` notation in a definition in `_utils/doc`. I tested and saw that it does not work.

hndgzkn:
I realized that I haven't put boundary check for coordinates for MosaicSlicer when specified as dictionary and added that check and a test for that.

nilearn/plotting/displays/_slicers.py
```diff
@@ -156,14 +168,10 @@ def _check_cut_coords_in_bounds(cls, img, cut_coords) -> None:
...
         # bounds, warn user
         if any(coord_in) and not all(coord_in):
-            out_of_bounds = np.asarray(cut_coords)[
-                np.logical_not(np.asarray(coord_in))
-            ].tolist()
             warnings.warn(
                 (
-                    f"The following 'cut_coords':"
-                    f"\n{out_of_bounds} "
-                    "\nseem to be out of the image bounds:"
```
hndgzkn, line 166: I have removed the part that displays the coordinates that are out of bounds, as with mosaic slicer, cut_coords can be a dict, this does not work. 

I can find a solution but I did not want to complicate the code for that. 

This warning is only raised when cut_coords is specified by the user. So displaying the  boundaries of image could be sufficient to identify the ones that are not in bounds.

nilearn/_utils/docs.py
```diff
@@ -1683,6 +1694,35 @@ def custom_function(vertices):
...
+
+black_bg : :obj:`bool`, or "auto", default=False
+    If `True`, the background of the image is set to be black.
+    If you wish to save figures with a black background,
+    you will need to pass `facecolor="k", edgecolor="k"`
+    to :func:`matplotlib.pyplot.savefig`.
+
+brain_color : :obj:`tuple`, default=(0.5, 0.5, 0.5)
+    The brain color to use as the background color (e.g., for
+    transparent colorbars).
```
Remi-Gau: oh I see what you mean now with you previous comments about 'is %(brain_color)s gonna work'?

maybe you can use f strings to reuse previous dict entries.

```python
docdict["projector_init_parameters"] = f"""
Parameters
----------
cut_coords : sequence of :obj:`float` or :obj:`int`
    The world coordinates of the point where the cut is performed. Not used
    for projectors.

...

{docdict["brain_color"]}
"""
``

**Comment by Remi-Gau** [comment]
good with me

@hndgzkn let me know if you think it's worth trying the string substitution mentioned here: https://github.com/nilearn/nilearn/pull/5920#discussion_r2690225078

otherwise we can merge

File: nilearn/_utils/docs.py (hunk 16 above)
Comment (hndgzkn) [reply nilearn/_utils/docs.py]: Ay yes! @Remi-Gau I'll update.

File: doc/changes/latest.rst [hunk 17]
```diff
@@ -17,5 +17,7 @@ Fixes
 Enhancements
 ------------
 
+- :bdg-success:`API` Add support for dictionary as ``cut_coords`` for :class:`~plotting.displays.MosaicSlicer` and image plotting functions. (:gh:`5920` by `Hande Gözükan`_).
+
```
Remi-Gau, line 21: @hndgzkn 
missed this: should it not be in the fix category? From my understanding our old doc strings used to say that we could support dict but we in fact did not: so this was a bug in a way?

hndgzkn: I am not sure which would be better? As far as I remember, docs did not say. I am not sure, but may be mentioned only at one place. There was a test or a few tests where MosaicSlicer was initiated with a dict as in #5911.

But with your comment, I realize that, I did not update `cut_coords` in `_utils/docs` to tell that MosaicSlicer can accept dict. I would do it in another PR.

nilearn/plotting/displays/_slicers.py
```diff
@@ -2233,18 +2300,29 @@ class MosaicSlicer(BaseSlicer):
...
+        sequence of length 3 indicating the number of cuts in each direction
+        ``(x, y, z)``, or a dictionary where keys are the directions
+        ('x', 'y', 'z') and the values are sequences holding the cut
+        coordinates.
+
+    %(slicer_init_parameters_partial)s
+
     Attributes
     ----------
     cut_coords : :obj:`dict` <:obj:`str`: 1D :class:`~numpy.ndarray`>
```
Remi-Gau, line 2317: doc string used to say dict, so I think we could say this was a bug

bthirion: Agreed

Issue #3379, opened 2022-10-07
Reorganizing the User Guide


alexisthual:
This reorganization was discussed during a nilearn core devs meeting ([meeting notes](https://hackmd.io/-Oz3Dw_8QcebntSuVUKWug?edit)).

Mostly, we intend to:
1. Reorder sections in the User Guide (and potentially the Examples) so that they feel more natural to new comers
1. Make sure every subsection belongs to the right place
1. Go through the entire User Guide content to polish it for a targeted audience of grad students
1. Unify user-interface components throughout the User Guide

alexisthual:
Here are some of the points that were discussed during the last core dev meeting:

## New section ordering

1. introduction @alexisthual :heavy_check_mark: 
2. manipulation @htwangtw  (Reordering needed, add content on surface data, can move data fetching to different section)
3. plotting @jeromedockes 
4. glm @ymzayek  (I think content is good and flows nicely through the 3 sections, maybe 2nd level models can be expanded) :heavy_check_mark: 
5. connectivity @alexisthual  (nice content, could be slightly reworked) :heavy_check_mark: 
6. decoding @bthirion
7. advanced @emdupre 

- Within sub-sections, conserve advanced material until the end of the sub-section
- Read through sub-sections to determine order and relevant appropriateness of material for target audience

## Target audience

- Under-grad and grad students
    - with a CS/ML background
    - or with a computational-bio, neuroscience background

## Revisting structural elements

UI/UX guidelines
- think mobile-first
- 1 column only
- minimal use of indented content
    - try to keep nested indentations as low as possible because they render poorly
- select a limited number of UI components from Furo
    - admonitions (https://pradyunsg.me/furo/kitchen-sink/admonitions/)
        - I (@alexisthual) suggest we only keep
            - warning (content users should absolutely read)
            - note (important content that needs to be stressed, but that is not a warning)
    - code-blocks
        - use `.. code-block:: default`

Content / redaction guidelines
- simple vocabulary
- short sentences
- less text is better
- add images every time they ease understanding
- don't repeat examples, reference them instead

alexisthual:
What was suggested is that each person who was assigned a section of the user guide opens a PR (if need be) to do a first pass.
During this first pass, our goal is to:
- check that all subsections are in the right order (ie an order that's easy to follow for our targeted audience)
- put advanced content at the end of the section
- correct for big mistakes if there are any

After all sections have been reviewed, the goal is to do a second pass on the entire user guide so as to:
[...]

PR #5925, opened 2026-01-05
[FIX] use sklearn html representation of estimators

### LINKED ISSUE #5924: [BUG] HTML representation of estimator parameters do not show in notebooks

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

- Linux Ubuntu 22.04

### Python version

- [ ] 3.14
- [x] 3.13
- [ ] 3.12
- [ ] 3.11
- [ ] 3.10

### nilearn version

dev

### Expected behavior

since sklearn 1.8 estimators can have an improved HTML representation

https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_8_0.html#html-representation-of-estimators

I would expect that the Nilearn estimators should be able to take advantage of this but it does not see to be the case

This is what I got (not visible on the image but the parameters show up their documentation when hovering over them).

Image

tried with this marimo notebook

```python
import marimo

__generated_with = "0.18.4"
app = marimo.App(width="full")


@app.cell
def _():
    import numpy as np
    import nibabel as nib

    X = rng.rand(5, 5, 5, 20)
    X_img = nib.Nifti1Image(X, np.eye(4))
    y = np.array([0] * 10 + [1] * 10)

    decoder = Decoder()
    decoder.fit(X_img, y)
    return


@app.cell
def _():
    from sklearn import linear_model

    reg = linear_model.LinearRegression()

    reg.fit([[0, 0], [1, 1], [2, 2]], [0, 1, 2])
    return


if __name__ == "__main__":
    app.run()
```

Remi-Gau:
- Closes #5924

- [x] add a NilearnBaseEstimator
- [x] make sure that the URL to the doc of the each estimator doc can be generated
- [x] rely on the sklearn HTML representation of estimator in reports (for sklearn >= 1.7.0)
- [x] show pictures of before and after
- [x] changelog

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -13,6 +13,8 @@ Fixes
 
 - :bdg-dark:`Code` Use sklearn HTML representation of estimators and their parameters in notebooks and reports (:gh:`5925` by `Rémi Gau`_).
 
+- :bdg-dark:`Code` Only throw warning about non-interactive plotting backend when using :func:`nilearn.plotting.show` (:gh:`5929` by `Rémi Gau`_).
+
 - :bdg-info:`Plotting` drop background color when using look up table as colormap (:gh:`5936` by `Rémi Gau`_).
 
 - :bdg-dark:`Code` Reallow use non-multi maskers for :class:`~nilearn.regions.Parcellations` (:gh:`5930` by `Rémi Gau`_).
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -40,7 +40,7 @@ alphas : :obj:`float` or :obj:`list` of :obj:`float` or None, default=None
 # annotate
 docdict["annotate"] = """
 annotate : :obj:`bool`, default=True
-    If `annotate` is `True`, positions and left/right annotation
+    If `annotate` is `True` (like positions and / or  left/right annotation)
     are added to the plot.
 """
 
@@ -281,7 +281,7 @@ cmap : :class:`matplotlib.colors.Colormap`, or :obj:`str`, \
 # colorbar
 docdict["colorbar"] = """
 colorbar : :obj:`bool`, optional
-    If `True`, display a colorbar on the right of the plots.
+    If `True`, display a colorbar next to the plots.
 """
 
 # connected
--- a/nilearn/_utils/helpers.py
+++ b/nilearn/_utils/helpers.py
@@ -72,19 +72,6 @@ def set_mpl_backend(message=None):
                 stacklevel=find_stack_level(),
             )
 
-        if new_backend == "agg":
-            warnings.warn(
-                (
-                    f"You are using the '{matplotlib.get_backend()}' "
-                    "matplotlib backend that is non-interactive."
-                    "\nNo figure will be plotted when calling "
-                    "matplotlib.pyplot.show() or nilearn.plotting.show()."
-                    "\nYou can fix this by installing a different backend: "
-                    "for example via\n\tpip install PyQt6"
-                ),
-                stacklevel=find_stack_level(),
-            )
-
 
 def rename_parameters(
     replacement_params,
--- a/nilearn/plotting/displays/_projectors.py
+++ b/nilearn/plotting/displays/_projectors.py
@@ -166,6 +166,7 @@ class OrthoProjector(OrthoSlicer):
                 f"'node_coords' shape is {node_coords_shape}."
             )
 
+    @fill_doc
     def add_graph(
         self,
         adjacency_matrix,
@@ -214,7 +215,7 @@ class OrthoProjector(OrthoSlicer):
             - If it is a number only the edges with a value greater than
               ``edge_threshold`` will be shown.
             - If it is a string it must finish with a percent sign,
-              e.g. "25.3%", and only the edges with a abs(value) above
+              e.g. "25.3%%", and only the edges with a abs(value) above
               the given percentile will be shown.
 
         edge_kwargs : :obj:`dict`, default=None
@@ -226,8 +227,8 @@ class OrthoProjector(OrthoSlicer):
             :func:`~matplotlib.pyplot.scatter` which plots all the
             nodes at one.
 
-        colorbar : :obj:`bool`, default=False
-            If ``True``, display a colorbar on the right of the plots.
+        %(colorbar)s
+            Default=False.
         """
         # set defaults
         edge_kwargs = edge_kwargs or {}
--- a/nilearn/plotting/displays/_slicers.py
+++ b/nilearn/plotting/displays/_slicers.py
@@ -258,8 +258,8 @@ class BaseSlicer:
         leave_space : :obj:`bool`, default=False
             If ``True``, leave space between the plots.
 
-        colorbar : :obj:`bool`, default=False
-            If ``True``, display a colorbar on the right of the plots.
+        %(colorbar)s
+            Default=False.
 
         %(brain_color)s
 
@@ -425,8 +425,8 @@ class BaseSlicer:
                 value is used to threshold the image: values below the
                 threshold (in absolute value) are plotted as transparent.
 
-        colorbar : :obj:`bool`, default=False
-            If ``True``, display a colorbar on the right of the plots.
+        %(colorbar)s
+            Default=False.
 
         cbar_tick_format : str, default="%%.2g" (scientific notation)
             Controls how to format the tick labels of the colorbar.
@@ -499,10 +499,10 @@ class BaseSlicer:
         threshold : :obj:`int` or :obj:`float` or ``None``, default=1e-6
             Threshold to apply:
 
-                - If ``None`` is given, the maps are not thresholded.
-                - If number is given, it must be non-negative. The specified
-                  value is used to threshold the image: values below the
-                  threshold (in absolute value) are plotted as transparent.
+            - If ``None`` is given, the maps are not thresholded.
+            - If number is given, it must be non-negative. The specified
+                value is used to threshold the image: values below the
+                threshold (in absolute value) are plotted as transparent.
 
         filled : :obj:`bool`, default=False
             If ``filled=True``, contours are displayed with color fillings.
@@ -1136,10 +1136,10 @@ class _MultiDSlicer(BaseSlicer):
         threshold : :obj:`int` or :obj:`float` or None, default=None
             Threshold to apply:
 
-                - If ``None`` is given, the maps are not thresholded.
-                - If number is given, it must be non-negative. The specified
-                  value is used to threshold the image: values below the
-                  threshold (in absolute value) are plotted as transparent.
+            - If ``None`` is given, the maps are not thresholded.
+            - If number is given, it must be non-negative. The specified
+                value is used to threshold the image: values below the
+                threshold (in absolute value) are plotted as transparent.
 
         cut_coords : sequence of :obj:`float` or :obj:`int`, or None, \
                     default=None
--- a/nilearn/plotting/html_connectome.py
+++ b/nilearn/plotting/html_connectome.py
@@ -7,7 +7,9 @@ from matplotlib import cm as mpl_cm
 from scipy import sparse
 
 from nilearn import DEFAULT_DIVERGING_CMAP
+from nilearn._utils.docs import fill_doc
 from nilearn._utils.html_document import HTMLDocument
+from nilearn._utils.param_validation import check_params
 from nilearn.datasets import fetch_surf_fsaverage
 from nilearn.plotting._engine_utils import colorscale, to_color_strings
 from nilearn.plotting.js_plotting_utils import (
@@ -240,6 +242,7 @@ def _make_connectome_html(connectome_info, embed_js=True):
     return ConnectomeView(as_html)
 
 
+@fill_doc
 def view_connectome(
     adjacency_matrix,
     node_coords,
@@ -273,7 +276,7 @@ def view_connectome(
         If it is a number only connections of amplitude greater
         than threshold will be shown.
         If it is a string it must finish with a percent sign,
-        e.g. "25.3%", and only connections of amplitude above the
+        e.g. "25.3%%", and only connections of amplitude above the
         given percentile will be shown.
 
     edge_cmap : :obj:`str` or matplotlib colormap, default="RdBu_r"
@@ -288,8 +291,8 @@ def view_connectome(
     node_size : :obj:`float`, default=3.0
         Size of the markers showing the seeds in pixels.
 
-    colorbar : :obj:`bool`, default=True
-        Add a colorbar.
+    %(colorbar)s
+        default=True
 
     colorbar_height : :obj:`float`, default=0.5
         Height of the colorbar, relative to the figure height.
@@ -297,8 +300,7 @@ def view_connectome(
     colorbar_fontsize : :obj:`int`, default=25
         Fontsize of the colorbar tick labels.
 
-    title : :obj:`str` or None, default=None
-        Title for the plot.
+    %(title)s
 
     title_fontsize : :obj:`int`, default=25
         Fontsize of the title.
@@ -326,6 +328,7 @@ def view_connectome(
         surface.
 
     """
+    check_params(locals())
     node_coords = np.asarray(node_coords)
 
     connectome_info = _get_connectome(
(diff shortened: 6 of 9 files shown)


Remi-Gau:
There seems to be a small upstream issue: https://github.com/scikit-learn/scikit-learn/issues/33005

nilearn/reporting/html_report.py
```diff
@@ -264,14 +261,6 @@ def _create_report(
...
                 sparsify=False,
             )
-    parameters = model_attributes_to_dataframe(estimator)
-    with pd.option_context("display.max_colwidth", 100):
-        parameters = dataframe_to_html(
-            parameters,
-            precision=2,
-            header=True,
-            sparsify=False,
-        )
```
Remi-Gau, line 274: we are going to need this for older version of sklearn that do not have an html repr of the estimators parameter

nilearn/_utils/html_repr.py
```diff
@@ -0,0 +1,51 @@
...
+            version_url = (
+                f"{nil_version.major}.{nil_version.minor}.{nil_version.minor}"
+            )
+
+        return (
+            f"https://nilearn.github.io/{version_url}/modules/generated/"
+            "{estimator_module}.{estimator_name}.html"
+        )
+    
+    if SKLEARN_GTE_1_8:
```
Remi-Gau: this may work with some older version of sklearn: find cut off

nilearn/glm/_base.py
```diff
@@ -48,13 +49,34 @@
...
 
-class BaseGLM(CacheMixin, BaseEstimator):
+class BaseGLM(_NilearnHTMLDocumentationLinkMixin, CacheMixin, BaseEstimator):
     """Implement a base class \
     for the :term:`General Linear Model<GLM>`.
     """
 
     _estimator_type = "glm"  # TODO (sklearn >= 1.8) remove
 
+    if SKLEARN_GTE_1_8:
```
Remi-Gau: this may work with some older version of sklearn: find cut off

nilearn/glm/_base.py
```diff
@@ -537,15 +559,18 @@ def generate_report(
...
-
-        model_attributes = glm_model_attributes_to_dataframe(self)
-        with pd.option_context("display.max_colwidth", 100):
-            model_attributes_html = dataframe_to_html(
-                model_attributes,
-                precision=2,
-                header=True,
-                sparsify=False,
-            )
+        if SKLEARN_GTE_1_8:
```
Remi-Gau: this may work with some older version of sklearn: find cut off

nilearn/maskers/tests/test_html_report.py
```diff
@@ -74,7 +74,8 @@ def generate_and_check_masker_report(
         excludes.append('<div class="image">')
 
     else:
-        if masker.__sklearn_is_fitted__():
+        if not SKLEARN_GTE_1_8 and masker.__sklearn_is_fitted__():
```
Remi-Gau: this may work with some older version of sklearn: find cut off

also make sure there is a else block here to check content when relying on the sklearn html repr

Remi-Gau:
actually commit [db92f29](https://github.com/nilearn/nilearn/pull/5925/commits/db92f2937e1bf61fb12987ad3f6e470b980f37d3) may not be necessary given it is a bad hack to fix a 1.8.0 bug

Remi-Gau:
pictures to compare

before: 
image


after:

image

We are losing the readibility of some parameters, but we are gaining (on top of less code to maintain for us) the docstring of each parameter when hovering over each param and the hightlighting of parameters that do not have the default value.

Note that some of the funky looking things about t_r in the new table is a bug in sklearn 1.8.0 that should be fixed in 1.8.1.

nilearn/_utils/estimator_checks.py
```diff
@@ -989,6 +991,35 @@ def check_doc_attributes(estimator) -> None:
         )
 
 
+def check_doc_link(estimator_orig):
```
bthirion, line 1010: in that case, it could be useful to describe the type of estimator_orig in the docstring.

nilearn/_utils/html_repr.py
```diff
@@ -0,0 +1,52 @@
...
+
+    _doc_link_module = "nilearn"
+
+    @property
+    def _doc_link_template(self):
+        version_url = "dev"
+        nil_version = parse(__version__)
+        if nil_version.dev is None:
+            version_url = (
+                f"{nil_version.major}.{nil_version.minor}.{nil_version.minor}"
```
bthirion:
```suggestion
                f"{nil_version.major}.{nil_version.minor}.{nil_version.patch}"
```

Remi-Gau: note that the versiion object does not have patch attribute: https://packaging.pypa.io/en/stable/version.html#packaging.version.Version

Remi-Gau:
> I could not see the output in the built doc.

@bthirion 

see for example the reports for the maskers:

https://output.circle-artifacts.com/output/job/7b0be4b3-0b4e-4b69-a624-cd60df63fdc8/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html

or 

https://output.circle-artifacts.com/output/job/7b0be4b3-0b4e-4b69-a624-cd60df63fdc8/artifacts/0/dev/_static/notebook_reports_maskers.html

Remi-Gau:
> > I could not see the output in the built doc.
> 
> @bthirion
> 
> see for example the reports for the maskers:
> 
> https://output.circle-artifacts.com/output/job/7b0be4b3-0b4e-4b69-a624-cd60df63fdc8/artifacts/0/dev/modules/generated_reports/masker_reports_examples.html
> 
> or
> 
> https://output.circle-artifacts.com/output/job/7b0be4b3-0b4e-4b69-a624-cd60df63fdc8/artifacts/0/dev/_static/notebook_reports_maskers.html

@emdupre 
what do you think of the using the sklearn HTML representation of estimators in our reports (or for notebooks)?

Not visible in the doc here, but with sklearn >=1.8.0 users would also get the docstring of each attribute when hovering over the table.

emdupre:
> @emdupre
what do you think of the using the sklearn HTML representation of estimators in our reports (or for notebooks)?

> We are losing the readibility of some parameters, but we are gaining (on top of less code to maintain for us) the docstring of each parameter when hovering over each param and the hightlighting of parameters that do not have the default value.

I think this is a reasonable tradeoff, particularly given the decrease in maintenance cost ! Personally, it's a bit hard for me to intuit why some values are highlighted without the roll-over feature, but I don't think that's worth bumping our sklearn version for.

Should we note in the documentation somewhere that non-default values are highlighted, since we can't assume rollover works ?

Remi-Gau:
> Should we note in the documentation somewhere that non-default values are highlighted, since we can't assume rollover works ?

fine by me but where in the doc given how many estimators are affected

Remi-Gau:
> but I don't think that's worth bumping our sklearn version for.

agreed that this does not justify it: but the number of TODOs that are dependent on a sklearn is slowly growing (at least 50 on main already) : SPEC-0 suggests dropping support for sklearn in the first quarter of 2026 (https://scientific-python.org/specs/spec-0000/), so maybe we could drop it at the end of 2026... to slowly move forward...

bthirion:
> > but I don't think that's worth bumping our sklearn version for.
> 
> agreed that this does not justify it: but the number of TODOs that are dependent on a sklearn is slowly growing (at least 50 on main already) : SPEC-0 suggests dropping support for sklearn in the first quarter of 2026 (https://scientific-python.org/specs/spec-0000/), so maybe we could drop it at the end of 2026... to slowly move forward...

you probably mean sklearn-1.4

Remi-Gau:
> you probably mean sklearn-1.4

indeed !!!! 🤣

hndgzkn (approved):
LGTM. 

I am not sure if that's a good idea but might worth discussion. 

How does it sound to create a `BaseNilearnEstimator` and inherit from `_NilearnHTMLDocumentationLinkMixin`? That would avoid updating each estimator and also help us implement common code for all estimators.

Remi-Gau:
> How does it sound to create a `BaseNilearnEstimator` and inherit from `_NilearnHTMLDocumentationLinkMixin`? That would avoid updating each estimator and also help us implement common code for all estimators.

not sure if it would be very different: at the moment we can just update _NilearnHTMLDocumentationLinkMixin and all classes that use it will be affected, no?

hndgzkn:
> not sure if it would be very different: at the moment we can just update _NilearnHTMLDocumentationLinkMixin and all classes that use it will be affected, no?

If a change is required on _NilearnHTMLDocumentationLinkMixin functionality, yes, this fixes it. 

However in the future, if we need a new mixin to be extended by all estimators, we will need to put it in all estimators and create a new dependency for each.
 
Or, if there will be a behavior different than BaseEstimator of scikit-learn, we will need to put it in all estimator classes.  

For example, if we had a base class for nilearn estimators, we could define `_more_tags` and common parts of `__sklearn_tags__`there and we would not need to implement it in all estimators. 

In this PR, we are introducing a new dependency in all estimators. However, if we are going to make a change in all estimators, this PR would be an opportunity to remove direct dependency on sklearn `BaseEstimator` and make each estimator dependent only on `BaseNilearnEstimator` class. It would be a central place to make changes common to all estimators and avoid changing all estimators each time.

Remi-Gau:
> we could define `_more_tags` and common parts of `__sklearn_tags__`

hm... let me try with _more_tags first

less convinced about __sklearn_tags__

hndgzkn:
> less convinced about **sklearn_tags**

Yes, might not be possible, or simpler.

The main benefit would be to reduce dependencies for estimator classes and avoid touching them again when there is a change that would apply to all.

nilearn/_base.py
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+
+class NilearnBaseEstimator(_NilearnHTMLDocumentationLinkMixin, BaseEstimator):
+    def _more_tags(self):
+        """Return estimator tags.
+
+        TODO (sklearn >= 1.6.0) remove
+        """
+        return self.__sklearn_tags__()
```
Remi-Gau: added here and removed from all our estimators

Remi-Gau:
@emdupre @bthirion 
asking for another look since this PR now adds a `NilearnBaseEstimator` that all our estimator inherit from.

emdupre:
> fine by me but where in the doc given how many estimators are affected

Maybe we should have an example on the reports in our doc (thinking in [Viz](https://nilearn.github.io/stable/auto_examples/01_plotting/index.html)) similarly to the sklearn section on [Displaying Pipelines](https://scikit-learn.org/stable/auto_examples/miscellaneous/plot_pipeline_display.html). It would really just be giving visibility to the [nilearn.reporting examples](https://nilearn.github.io/stable/modules/reporting.html).

Not for this PR, but we can open a new issue if you agree !

emdupre:
> @emdupre @bthirion
asking for another look since this PR now adds a NilearnBaseEstimator that all our estimator inherit from.

I think the `NilearnBaseEstimator` makes sense ; agree that it will just be good to get the tests passing to confirm !

Remi-Gau:
> Maybe we should have an example on the reports in our doc (thinking in [Viz](https://nilearn.github.io/stable/auto_examples/01_plotting/index.html)) similarly to the sklearn section on [Displaying Pipelines](https://scikit-learn.org/stable/auto_examples/miscellaneous/plot_pipeline_display.html). It would really just be giving visibility to the [nilearn.reporting examples](https://nilearn.github.io/stable/modules/reporting.html).
> 
> Not for this PR, but we can open a new issue if you agree !

actually I remembered seeing the sklearn html repr of some our estimators in our examples:

https://nilearn.github.io/dev/auto_examples/04_glm_first_level/plot_bids_features.html#first-level-model-estimation-one-subject

image


so in a follow up PR we could make sure the very simplest (entry level) examples with maskers, decoders, glm... actually display this HTML representation of the estimator with some explanation of the color coding

Remi-Gau:
OK this is an example with decoder:

https://output.circle-artifacts.com/output/job/5eed32a0-d37c-4a96-b038-615be4704c1b/artifacts/0/dev/auto_examples/00_tutorials/plot_decoding_tutorial.html#decoding-with-support-vector-machine

image

Will do the same for one GLM and one masker

Remi-Gau:
@emdupre 
which example would you consider the best 'entry level' one that explains 'maskers' to users?
we have something for decoding and glm but nothing for maskers in `examples/00_tutorials`

emdupre:
Usually I point to this section of the User Guide : https://nilearn.github.io/stable/manipulating_images/masker_objects.html

There we point to these examples : 
1. https://nilearn.github.io/stable/auto_examples/06_manipulating_images/plot_mask_computation.html#sphx-glr-auto-examples-06-manipulating-images-plot-mask-computation-py
2. https://nilearn.github.io/stable/auto_examples/06_manipulating_images/plot_nifti_simple.html#sphx-glr-auto-examples-06-manipulating-images-plot-nifti-simple-py

So I'd suggest one of those, personally ! Given the way we structured the docs so far, I don't know that I would put anything in 00_tutorials

Remi-Gau:
> Usually I point to this section of the User Guide : https://nilearn.github.io/stable/manipulating_images/masker_objects.html
> 
> There we point to these examples :
> 
>     1. https://nilearn.github.io/stable/auto_examples/06_manipulating_images/plot_mask_computation.html#sphx-glr-auto-examples-06-manipulating-images-plot-mask-computation-py
> 
>     2. https://nilearn.github.io/stable/auto_examples/06_manipulating_images/plot_nifti_simple.html#sphx-glr-auto-examples-06-manipulating-images-plot-nifti-simple-py
> 
> 
> So I'd suggest one of those, personally ! Given the way we structured the docs so far, I don't know that I would put anything in 00_tutorials

Thanks @emdupre that's very useful.
I will rework those examples in this PR.

I actually feel that the [plot-nifti-simple-py](https://nilearn.github.io/stable/auto_examples/06_manipulating_images/plot_nifti_simple.html#sphx-glr-auto-examples-06-manipulating-images-plot-nifti-simple-py) could be moved to the tutorial section.

Remi-Gau:
> I actually feel that the [plot-nifti-simple-py](https://nilearn.github.io/stable/auto_examples/06_manipulating_images/plot_nifti_simple.html#sphx-glr-auto-examples-06-manipulating-images-plot-nifti-simple-py) could be moved to the tutorial section.

maybe in another PR

doc/maintenance.rst
```diff
@@ -279,6 +280,7 @@ but care should be taken that at least the following have not been affected:
 
 - ``.rst`` files containing the string ``image:: ../auto_examples``,
 - the tables in the :ref:`plotting` section,
+- ``literalinclude`` sphinx-directives with a ``:start-after:``
```
Remi-Gau, line 283: adding this for us to check before making a release

doc/manipulating_images/masker_objects.rst
```diff
@@ -148,10 +148,6 @@ method of the masker. The generated report can be viewed in a Jupyter notebook,
 opened in a new browser tab using ``report.open_in_browser()``,
 or saved as a portable HTML file ``report.save_as_html(output_filepath)``.
 
-.. literalinclude:: ../../examples/06_manipulating_images/plot_mask_computation.py
-    :start-after: # We need to specify an 'epi' mask_strategy, as this is raw :term:`EPI` data
-    :end-before: # Generate mask with strong opening
-
```
Remi-Gau, line 154: removing this literal include because in the end it just amounted to showing

```python
masker.generate_report()
```

doc/user_guide.rst
```diff
@@ -19,7 +19,7 @@ Table of contents
 
 .. toctree::
    :numbered:
-   :maxdepth: 3
+   :maxdepth: 2
```
Remi-Gau, line 22: just to avoid a TOC with some many entries that it requires doom scrolling through 25 pages

https://nilearn.github.io/stable/user_guide.html

Remi-Gau:
@emdupre I cleaned a bit the 2 examples you mentioned to also add some more headings

emdupre:
>> I actually feel that the [plot-nifti-simple-py](https://nilearn.github.io/stable/auto_examples/06_manipulating_images/plot_nifti_simple.html#sphx-glr-auto-examples-06-manipulating-images-plot-nifti-simple-py) could be moved to the tutorial section.

> maybe in another PR

@Remi-Gau yes let's discuss in another issue ; to me it could be related to https://github.com/nilearn/nilearn/issues/3379

PR #5928, opened 2026-01-06
[DOC] remove 'optional' from most doc strings

Remi-Gau:
- Closes none

- actually describe the default instead of just saying optional

### DIFF
--- a/nilearn/surface/surface.py
+++ b/nilearn/surface/surface.py
@@ -1453,7 +1453,7 @@ class PolyData:
         dim : int
             Dimensions the data should have.
 
-        var_name : str, default = "img"
+        var_name : str, default="img"
             Name of the variable to include in the error message.
 
         Returns


nilearn/surface/surface.py
```diff
@@ -1453,7 +1453,7 @@ def _check_ndims(self, dim: int, var_name="img"):
         dim : int
             Dimensions the data should have.
 
-        var_name : str, optional
+        var_name : str, default = "img"
```
bthirion:
```suggestion
        var_name : str, default="img"
```

Remi-Gau:
@hndgzkn I have not touched the doc strings in the slicer and projetor modules as you are working on those

nilearn/_utils/data_gen.py
```diff
@@ -1074,7 +1074,7 @@ def _check_entities_and_labels(entities):
 
     Parameters
     ----------
-    entities : :obj:`dict`, optional
```
man-shu, line 1077: I know this is an internal function but would it still be relevant to mention the default here? I think it would be an empty dict?

nilearn/_utils/data_gen.py
```diff
@@ -1135,7 +1135,7 @@ def _mock_bids_dataset(
         number of runs for the corresponding task.
         No run entity will be used if a value is equal to 1 or less.
 
-    entities : :obj:`dict`, optional
```
man-shu, line 1138: same here

nilearn/_utils/data_gen.py
```diff
@@ -1250,7 +1250,7 @@ def _mock_bids_derivatives(
         Use n_vertices == 10242 to match the number of vertices
         in fsaverage5.
 
-    spaces : :obj:`list` of :obj:`str`, optional.
```
man-shu, line 1253: here also

nilearn/_utils/data_gen.py
```diff
@@ -1461,7 +1461,7 @@ def _write_bids_derivative_func(
     rand_gen : :obj:`numpy.random.RandomState` instance
         Random number generator.
 
-    confounds_tag : :obj:`str`, optional.
```
man-shu, line 1464: and here

man-shu (approved):
Otherwise LGTM!

nilearn/_utils/data_gen.py
Remi-Gau: hm... I'd rather just rely on the signature of the function to determine what the default is: in this case entities is an arg not a kwarg so it does not have any default, no?

man-shu: Ok ok makes sense.

PR #5929, opened 2026-01-06
[FIX] only throw non-interactive plotting backend warning when using nilearn show()

### LINKED ISSUE #5927: [BUG] Undesirable matplotlib backend warning for headless users (and incorrect fix advice)

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Linux Ubuntu 22.04 (GitHub Actions Runner)

### Python version

- [ ] 3.14
- [ ] 3.13
- [ ] 3.12
- [x] 3.11
- [ ] 3.10

### nilearn version

0.13.0

### Expected behavior

Hello,

I use `nilearn` in my codebase to generate an image file (with `nilearn.plotting.plot_anat`). This code is ran on a server and in CI, (GitHub Actions Runner) and no GUI or interactive operation is expected from the library. Therefore I do not expect to see warnings about a `matplotlib` GUI interactive backend when I do not use it.

I got this warning in my CI:

```bash
UserWarning: You are using the 'agg' matplotlib backend that is non-interactive.
 No figure will be plotted when calling matplotlib.pyplot.show() or nilearn.plotting.show().
 You can fix this by installing a different backend: for example via
 pip install PyQt6
 from nilearn import image, plotting
```

Notice that the error suggests to install PyQT6, but this is likely not a good solution for users that do not need a GUI, and does not work in environments that do not have a GUI.

Import `nilearn.plotting` in an environment that uses the `agg` matplotlib backend.
```python
from nilearn import plotting
```

Relevant code:
https://github.com/nilearn/nilearn/blob/70e1eb214f4aa58c5917d69a2f1f5a0e88849af1/nilearn/_utils/helpers.py#L74-L85

Remi-Gau:
- Closes #5927 

- only throw warning about non-interactive backend when using show()


nilearn/plotting/image/img_plotting.py
```diff
@@ -66,6 +66,18 @@ def show():
     """
     if get_backend().lower() != "agg":  # avoid warnings
         plt.show()
+    else:
```
bthirion, line 69: Does this account for all possible occurrences ?

Remi-Gau: did you mean all occurences of plt.show or all possible backends?

Remi-Gau: if the former then yes

PR #5930, opened 2026-01-06
[FIX] reallow use non-multi maskers for Parcellation

### LINKED ISSUE #5926: [BUG] Decompositions don't work with NiftiMaskers
After the release of 0.13.0, decompositions only allow for `Multi{Surface;Nifti}Masker`. Passing in another masker results in a type error which breaks legacy code.

https://github.com/nilearn/nilearn/blob/70e1eb214f4aa58c5917d69a2f1f5a0e88849af1/nilearn/_utils/docs.py#L698-L703

Moreover such lines are never reached:

https://github.com/nilearn/nilearn/blob/70e1eb214f4aa58c5917d69a2f1f5a0e88849af1/nilearn/decomposition/_base.py#L567-L572

Remi-Gau:
- Closes #5926

- [x] add test for Parcellation with (multi)niftimasker and (multi)surfacemasker
- [x] check that other decomposition estimators only accept multimaskers
- [x] adapt doc strings
- [x] changelog

### DIFF
--- a/nilearn/decomposition/dict_learning.py
+++ b/nilearn/decomposition/dict_learning.py
@@ -14,11 +14,13 @@ from sklearn.linear_model import Ridge
 from nilearn._utils import logger
 from nilearn._utils.docs import fill_doc
 from nilearn._utils.helpers import transfer_deprecated_param_vals
-from nilearn._utils.param_validation import check_is_of_allowed_type
+from nilearn._utils.param_validation import (
+    check_is_of_allowed_type,
+    sanitize_verbose,
+)
 from nilearn.maskers import MultiNiftiMasker, MultiSurfaceMasker
 from nilearn.surface import SurfaceImage
 from nilearn.typing import NiimgLike
-from nilearn._utils.param_validation import sanitize_verbose
 
 from ._base import _BaseDecomposition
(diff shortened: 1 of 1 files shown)


Remi-Gau:
@bthirion @pbarbarant 
This should fix the bug introduced by 0.13.0.
But AFAICT other decomposition estimators were never allowed to use non multi maskers, so this PR tries to both:
- allow them non-multi maskers for Parcellation
- forbid them for DictLearning and CanICA

does this make sense?

pbarbarant:
> @bthirion @pbarbarant This should fix the bug introduced by 0.13.0. But AFAICT other decomposition estimators were never allowed to use non multi maskers, so this PR tries to both:
> 
> * allow them non-multi maskers for Parcellation
> * forbid them for DictLearning and CanICA
> 
> does this make sense?

I think this is the behavior we want to achieve.

bthirion (commented):
The PR LGTM. is there anything you want to add ?

Remi-Gau:
> is there anything you want to add ?

doc string update and changelog

Remi-Gau:
@pbarbarant if you have a bit of time, could you cross check this PR?

PR #5931, opened 2026-01-08
[FIX] improve compliance Parcellation estimator with sklearn guidelines

Remi-Gau:
- Closes none 

- run estimator checks on Parcellation estimator

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -143,7 +143,7 @@ jobs:
                 git config --global user.name "GitHub Actions"
                 git config --global user.email "actions@github.com"
                 git add .
-                git commit -m "Update benchmark results and logs (${{ github.event.repository.updated_at }}_${{ github.run_number }})"
+                git commit --message "Update benchmark results and logs (${{ github.event.repository.updated_at }}_${{ github.run_number }})"
                 git push origin main
         -   name: Push html to gh-pages branch
             if: github.event_name != 'pull_request'
@@ -154,5 +154,5 @@ jobs:
                 git checkout gh-pages
                 cp -r ../html/* .
                 git add .
-                git commit -m "Update benchmark HTML (${{ github.event.repository.updated_at }}_${{ github.run_number }})"
+                git commit --message "Update benchmark HTML (${{ github.event.repository.updated_at }}_${{ github.run_number }})"
                 git push origin gh-pages
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -32,7 +32,7 @@
 #
 # .. code-block:: bash
 #
-#   $ git commit -m "[full doc] request full build"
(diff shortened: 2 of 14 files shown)


nilearn/maskers/nifti_labels_masker.py
```diff
@@ -615,18 +615,7 @@ def fit_transform(self, imgs, y=None, confounds=None, sample_mask=None):
...
-            compatibility.
-
-        region_ids_: dict[str | int, int | float] = {}
-        if self.background_label in index:
-            index.pop(index.index(self.background_label))
-            region_ids_["background"] = self.background_label
-        for i, id in enumerate(index):
-            region_ids_[i] = id  # noqa : PERF403
-
-        return region_ids_
```
Remi-Gau, line 629: some copy paste error?

nilearn/regions/parcellations.py
```diff
@@ -362,8 +381,6 @@ def __init__(
         n_jobs=1,
         verbose=0,
     ):
-        if memory is None:
-            memory = Memory(location=None)
```
Remi-Gau, line 380: done during fit

nilearn/regions/parcellations.py
```diff
@@ -341,12 +359,13 @@ class Parcellations(_MultiPCA):
 
     def __init__(
         self,
-        method,
+        method=None,
```
Remi-Gau, line 378: required by some sklearn check

it's OK because we still get an error if the user does not set it to something else than None

bthirion (commented):
The PR LGTM but the CIs error are related to the PR.

pbarbarant:
@Remi-Gau can you ping me when you need the next round of reviews?

Remi-Gau:
> can you ping me when you need the next round of reviews?

will do but I think that if the CI comes out green this should be good for a review

bthirion (commented):
The PR LHGTM but the CI is still red.

Remi-Gau:
@pbarbarant ready for review

nilearn/regions/parcellations.py
```diff
@@ -646,6 +676,7 @@ def fit_transform(self, imgs, confounds=None):
             (number of scans, number of labels)
 
         """
+        del y
```
pbarbarant, line 679: Don't we want to warn the user about this side effect?

Remi-Gau: what side effect do you have in mind?

Tried to check on one of the many places where we do a `del y`:

```python
from nilearn.maskers import SurfaceMasker
from nilearn.conftest import _surf_img_1d

img = _surf_img_1d()
y = [1]

masker = SurfaceMasker()
masker.fit_transform(img, y=y)
print(y)
```

and it behaves as expected

pbarbarant: All good then!

PR #5937, opened 2026-01-12
[FIX] use reshape() to change shape of an array

Remi-Gau:
- Closes none but should prevent some issues related to a coming numpy deprecation: https://github.com/numpy/numpy/pull/29536/files created some failures when using the nightly build of numpy (see https://github.com/nilearn/nilearn/actions/runs/20918803064/job/60098863248?pr=5859#step:6:12852)

https://github.com/nilearn/nilearn/actions/runs/20918803064/job/60098863248?pr=5859#step:6:12852

- use the reshape method of numpy arrays rather than directly setting the array shape

### DIFF
--- a/nilearn/decoding/_proximal_operators.py
+++ b/nilearn/decoding/_proximal_operators.py
@@ -169,7 +169,7 @@ def prox_tvl1(
     """
     weight = float(weight)
     input_img_flat = input_img.view()
-    input_img_flat.shape = input_img.size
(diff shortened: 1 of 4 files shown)


Remi-Gau:
> The difficulty is to catch all occurrence I guess !

If we pay regular attention to the deprecation warnings coming from our dependencies we should be fine... Famous last words...

hndgzkn (approved):
LGTM

PR #5939, opened 2026-01-13
[FIX] warn when feature selection is not performed

Remi-Gau:
- Closes none but related to https://github.com/nilearn/nilearn/pull/5863#issuecomment-3710266667

- throw warning when requested screening is not done


nilearn/decoding/decoder.py
```diff
@@ -380,7 +382,20 @@ def _parallel_fit(
...
+        else:
+            warnings.warn(
+                (
+                    f"number of features ({X_train.shape[1]}) "
+                    f"less than or equal to {MIN_N_FEATURES_FOR_SCREENING}: "
+                    "no feature selection will be performed."
+                ),
+                category=UserWarning,
+                stacklevel=find_stack_level(),
+            )
```
Remi-Gau, line 494: something seems off with this:
AFAICT this is done separately for each cross validation:
- so **in theory** this could mean that some fold could have some feature selection and others not...
- this likely mean that the user will get one warning per fold

would it not be easier to decide before the parallel fit if we do the feature selection and throw a single warning?

nilearn/decoding/tests/test_decoder.py
```diff
@@ -1236,19 +1210,18 @@ def test_decoder_apply_mask_surface(_make_surface_class_data):
...
 def test_decoder_screening_percentile_surface_default(
     _make_surface_class_data,
 ):
     """Test default screening percentile with surface image."""
     X, y = _make_surface_class_data
 
     model = Decoder(mask=SurfaceMasker(), standardize="zscore_sample")
-    model.fit(X, y)
+    with pytest.warns(match="no feature selection will be performed"):
+        model.fit(X, y)
```
Remi-Gau, line 1299: with may want to have test surface data with more vertices otherwise we cannot properly test feature selection on surface data

PR #5940, opened 2026-01-13
[FIX] revert short option for pytest markers

Remi-Gau:
- Closes none

- reverts one change from https://github.com/nilearn/nilearn/pull/5935 that only lists the test markers instead of only selecting the tests with some markers


Remi-Gau:
@nilearn/core my bad on this: this means that a lot of the CI have not been running any test since yesterday

PR #5942, opened 2026-01-13
[MAINT] handle all FutureWarnings during tests

### LINKED ISSUE #3541: Suggestion to use warnings-as-errors with pytest in prerelease CI

Remi-Gau:
- closes #3541

- [x] raise error during tests if FutureWarning is thrown: in consequence handle all FutureWarning with `pytest.warns` in tests or adapt tests to not throw any FutureWarnings
- [x] use standardize=None to initialize a lot of maskers during tests (we will be able to remove a lot of that in nilearn >= 0.15 when this becomes the new default)
- [x] replace some catch_warnings with pytest.warns

### DIFF
--- a/.github/workflows/test_with_tox.yml
+++ b/.github/workflows/test_with_tox.yml
@@ -109,14 +109,6 @@ jobs:
                     py: '3.10'
                     os: macos-latest
                     env: plot_min
-                -   description: oldest dependencies - no plotting - ubuntu
-                    py: '3.10'
-                    os: ubuntu-latest
-                    env: plot_min
-                -   description: oldest dependencies - with plotting - windows
-                    py: '3.10'
-                    os: windows-latest
-                    env: plot_min
                 # plotly images saved on different OS can generate slightly
                 # different baseline images. As baseline images are generated
                 # on a linux machine, ubuntu-latest is used


nilearn/tests/test_signal.py
Remi-Gau: some of those tests are too long and do too many things: try to split them into smaller ones

nilearn/decoding/decoder.py
```diff
@@ -58,13 +60,27 @@
...
-    "logistic_l2": LogisticRegressionCV(penalty="l2", solver="liblinear"),
-    "logistic": LogisticRegressionCV(penalty="l2", solver="liblinear"),
+    "logistic_l1": LogisticRegressionCV(
+        l1_ratios=(1,), solver="liblinear", **kwarg_logistic_regression_cv
+    ),
+    "logistic_l2": LogisticRegressionCV(
+        l1_ratios=(0,), solver="liblinear", **kwarg_logistic_regression_cv
+    ),
+    "logistic": LogisticRegressionCV(
+        l1_ratios=(0,), solver="liblinear", **kwarg_logistic_regression_cv
```
Remi-Gau: applying sklearn recommendations

nilearn/_utils/docs.py
Remi-Gau: add some details about default that had not been added in the previous release

nilearn/regions/signal_extraction.py
```diff
@@ -221,7 +221,7 @@ def img_to_signals_labels(
     order="F",
     strategy="mean",
     keep_masked_labels=False,
-    return_masked_atlas=False,
```
Remi-Gau: this was supposed to change to True in the previous release

Remi-Gau:
OK this PR is touching on too many things: will split it into smaller ones because I am getting confused by it

doc/maintenance.rst
```diff
@@ -129,6 +129,26 @@ We welcome pull requests from all community members, if they follow the
...
+
+Use :class:`FutureWarning` for deprecations that users should know about
+on which they can have some control,
+for example by making sure that the warning message mentions
+what action to take to prevent the warning to be thrown.
+If there is no way for the user to prevent the warning to be thrown,
+consider using a :class:`DeprecationWarning`.
+
+Tests are run via tox using ``-W error::FutureWarning``
+so that Any FutureWarning thrown by a test will raise an error.
```
Remi-Gau:
```suggestion
so that any FutureWarning thrown by a test will raise an error.
```

Remi-Gau:
there is a test failing only on MacOS with oldest python / dependencies

needs further investigation

.github/workflows/test_with_tox.yml
```diff
@@ -91,7 +92,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
```suggestion
            fail-fast: true
```

revert before merging

.github/workflows/test_with_tox.yml
```diff
@@ -108,6 +109,14 @@ jobs:
...
                     os: macos-latest
                     env: plot_min
+                -   description: oldest dependencies - no plotting - ubuntu
+                    py: '3.10'
+                    os: ubuntu-latest
+                    env: plot_min
+                -   description: oldest dependencies - with plotting - windows
+                    py: '3.10'
+                    os: windows-latest
+                    env: plot_min
```
```suggestion
```

remove before merging

Remi-Gau:
ok not just macOS

PR #5947, opened 2026-01-15
[MAINT] mark some tests are thread unsafe

Remi-Gau:
- Closes none

- mark as thread_unsafe some tests that have failed on some PRs when running on several threads

### DIFF
--- a/nilearn/glm/tests/test_first_level.py
+++ b/nilearn/glm/tests/test_first_level.py
@@ -768,6 +768,7 @@ def test_fmri_inputs_design_matrices_csv(tmp_path, shape_4d_default):
     )
 
 
+@pytest.mark.slow
 def test_fmri_inputs_events_type(tmp_path):
     """Check events can be dataframe or pathlike to CSV / TSV."""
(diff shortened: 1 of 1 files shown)


Remi-Gau:
I think we should expect to once in a while have some more tests failing given the nondeterministic aspect of parallel testing using threads

nilearn/_utils/estimator_checks.py
```diff
@@ -1516,60 +1516,56 @@ def check_img_estimator_fit_idempotent(estimator_orig):
     """
     check_methods = ["predict", "transform", "decision_function"]
 
-    estimator = clone(estimator_orig)
```
Remi-Gau, line 1519: rewrote this test so the clone(estimator) is in the loop which should make less brittle when running it on several threads

Remi-Gau:
> I'm a bit frightened when I realize that failures are non-deterministic...

yeah I know: I have tried to increase the number of threads to increase the chance of tests failing if they can fail but from my understanding there may be a chance that some tests may only fail very rarely

Issue #4379, opened 2024-04-12
Doc builds are failing on main


Remi-Gau:
This seems to be a new thing:
https://github.com/nilearn/nilearn/actions/workflows/build-docs.yml?query=branch%3Amain

The build goes on for 6 hours and then is killed.

See here:
https://github.com/nilearn/nilearn/actions/runs/8642310331/job/23693214886#step:34:484

It seems to be related to plotting something with plotly: https://github.com/nilearn/nilearn/blob/71e9f41fbd5f95d99bd5be7f3113003d6ad52d4f/examples/01_plotting/plot_3d_map_to_surface_projection.py#L90

Possibly related to:
- https://github.com/nilearn/nilearn/issues/4376
- https://github.com/nilearn/nilearn/issues/3925

Issue #4376, opened 2024-04-10
[BUG] plotly in nilearn.surface.vol_to_surf opens two interactive windows


lilikapa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Operating system

- [ ] Linux
- [X] Mac
- [ ] Windows

### Operating system version

- Mac OS Version 14.4 "Sonoma"

### Python version

- [ ] 3.12
- [X] 3.11
- [ ] 3.10
- [ ] 3.9
- [ ] 3.8

### nilearn version

0.10.3

### Expected behavior

(see below)

### Current behavior & error messages

Hi nilearn community, 

Thank you very much for this very helpful tool. I am not really sure whether this is really a bug. However, since a warning and a request were stated in the documentation, I thought it might be useful to report the following behavior.

When selecting `engine='plotly'` as a parameter within the `plot_surf_roi` function in a JupyterNotebook (see snippet of my code below), an interactive window is open within my JupyterNotebook, but also another in my browser. I would have expected the interactive window to only open within the notebook. However, if this is the expected behavior, please ignore the issue. Other than that, the plotting seemed to work well.

best,
lya

### Steps and code to reproduce bug

```python
fig = plotting.plot_surf_roi(fsavg6['infl_left'], roi_map=roi_surf,
                             hemi='left', view='lateral', 
                             bg_map=fsavg6['sulc_left'], bg_on_data=True,  
                             cmap='Blues', 
                             colorbar=True, 
                             vmin=0, vmax=2,
                             engine='plotly',
                             title='title')
fig.show()
```

Remi-Gau:
I can reproduce this with the code below on Nilearn 0.10.4

```
from nilearn import datasets
from nilearn import plotting
import numpy as np
from nilearn import surface

stat_img = datasets.load_sample_motor_activation_image()

fsaverage = datasets.fetch_surf_fsaverage()

curv_right = surface.load_surf_data(fsaverage.curv_right)
curv_right_sign = np.sign(curv_right)

texture = surface.vol_to_surf(stat_img, fsaverage.pial_right)

engine="plotly"

fig = plotting.plot_surf_stat_map(
    fsaverage.infl_right, texture, hemi='right',
    title='Surface right hemisphere', colorbar=True,
    threshold=1., bg_map=curv_right_sign, bg_on_data=True,
    engine=engine
)
fig.show() 
```

Remi-Gau:
no idea if this is the expected behavior or if this is a new thing
[...]

PR #5020, opened 2025-01-06
[DOC] use matplolib by default in plot_3d_map_to_surface_projection.py

Remi-Gau:
- Closes #5018

Changes proposed in this pull request:

- use matplotlib instead of plotly for the figures that would not show on doc build

PR #5949, opened 2026-01-16
[FIX] View plotly surface figures only once in notebook

hndgzkn:
- Closes 
  - #4376 
  - #4379

- Update default renderer to `None` for plotly `figure.show` 
- Only return plotly figure from `nilearn.plotting.displays._figures.show` when building docs with sphinx-gallery. 
- Update `examples/01_plotting/plot_3d_map_to_surface_projection` to use plotly again.

### DIFF
--- a/nilearn/decoding/tests/test_space_net.py
+++ b/nilearn/decoding/tests/test_space_net.py
@@ -403,6 +403,7 @@ def test_log_reg_vs_graph_net_two_classes_iris(
     assert_array_equal(tvl1.predict(X_), sklogreg.predict(X))
 
 
+@pytest.mark.slow
 def test_lasso_vs_graph_net():
     """Test for one of the extreme cases of Graph-Net.
 
@@ -491,6 +492,7 @@ def test_crop_mask_empty_mask(mask_empty):
         _crop_mask(mask_empty)
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize("model", [SpaceNetRegressor, SpaceNetClassifier])
 def test_space_net_one_alpha_no_crash(model):
     """Regression test."""
--- a/nilearn/maskers/tests/test_nifti_masker.py
+++ b/nilearn/maskers/tests/test_nifti_masker.py
@@ -174,6 +174,7 @@ def test_nan(affine_eye):
     assert not mask[:, :, -1].any()
 
 
+@pytest.mark.slow
 def test_matrix_orientation():
     """Test if processing is performed along the correct axis."""
     # the "step" kind generate heavyside-like signals for each voxel.
@@ -248,6 +249,7 @@ def test_mask_4d(shape_3d_default, affine_eye):
     assert_array_equal(data_trans3, data_trans_direct_diff)
 
 
+@pytest.mark.slow
 def test_4d_single_scan(rng, shape_3d_default, affine_eye):
     """Test that list of 4D images with last dim=1 is treated as 3D."""
     shape_3d = (10, 10, 10)
--- a/nilearn/plotting/image/tests/test_plot_stat_map.py
+++ b/nilearn/plotting/image/tests/test_plot_stat_map.py
@@ -15,6 +15,7 @@ from nilearn.plotting import plot_stat_map
 from nilearn.plotting.find_cuts import find_cut_slices
 
 
+@pytest.mark.slow
 def test_plot_stat_map_bad_input(matplotlib_pyplot, img_3d_mni, tmp_path):
     """Test for bad input arguments (cf. #510)."""
     filename = tmp_path / "temp.png"
@@ -146,6 +147,7 @@ def test_plot_stat_map_colorbar_variations(
         plot_stat_map(img, cut_coords=(-90, -125, -70), **params)
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize(
     "shape,direction", [((1, 6, 7), "x"), ((5, 1, 7), "y"), ((5, 6, 1), "z")]
 )
@@ -158,6 +160,7 @@ def test_plot_stat_map_singleton_ax_dim(
     )
 
 
+@pytest.mark.slow
 def test_outlier_cut_coords(matplotlib_pyplot):
     """Test to plot a subset of a large set of cuts found for a small area."""
     bg_img = load_mni152_template(resolution=2)
--- a/nilearn/plotting/tests/test_html_stat_map.py
+++ b/nilearn/plotting/tests/test_html_stat_map.py
@@ -426,6 +426,7 @@ def test_view_img_3d_warnings_more():
     check_html_view_img(html_view, title="SOME_TITLE")
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize(
     "params",
     [
--- a/nilearn/regions/tests/test_parcellations.py
+++ b/nilearn/regions/tests/test_parcellations.py
@@ -100,6 +100,7 @@ def image_2():
     return Nifti1Image(data, affine=_affine_eye())
 
 
+@pytest.mark.slow
 def test_error_parcellation_method_none(image_1):
     with pytest.raises(
         ValueError, match=r"Parcellation method is specified as None. "
@@ -107,6 +108,7 @@ def test_error_parcellation_method_none(image_1):
         Parcellations(method=None).fit(image_1)
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize("method", ["kmens", "avg", "completed"])
 def test_errors_raised_in_check_parameters_fit(method, image_1):
     """Test whether an error is raised or not given a false method type."""
--- a/nilearn/regions/tests/test_signal_extraction.py
+++ b/nilearn/regions/tests/test_signal_extraction.py
@@ -519,6 +519,7 @@ def test_signal_extraction_with_maps(affine_eye, shape_3d_default, rng):
     assert_almost_equal(get_data(img_r), get_data(imgs))
 
 
+@pytest.mark.slow
 @pytest.mark.thread_unsafe
 def test_signal_extraction_with_maps_and_labels(
     labeled_regions, fmri_img, shape_3d_default
--- a/nilearn/tests/test_masking.py
+++ b/nilearn/tests/test_masking.py
@@ -89,6 +89,7 @@ def test_load_mask_img_surface(surf_mask_1d):
         assert hemi.dtype == "bool"
 
 
+@pytest.mark.slow
 def test_high_variance_confounds():
     """Test high_variance_confounds."""
     img, mask, conf = _simu_img()
@@ -684,6 +685,7 @@ def test_intersect_masks_f8(img_2d_mask_bottom_right, img_2d_mask_center):
     assert_array_equal(mask_ab, get_data(mask_ab_change_type))
 
 
+@pytest.mark.slow
 def test_intersect_masks(
     affine_eye, img_2d_mask_bottom_right, img_2d_mask_center
 ):
@@ -814,6 +816,7 @@ def test_compute_multi_epi_mask(affine_eye):
     assert_array_equal(mask_ab, get_data(mask_ab_))
 
 
+@pytest.mark.slow
 def test_compute_multi_brain_mask_error():
     """Check error raised if images with different shapes given as input."""
     imgs = [


nilearn/plotting/displays/_figures.py
```diff
@@ -104,18 +104,17 @@ def __init__(self, figure=None, output_file=None, hemi=DEFAULT_HEMI):
...
         Parameters
         ----------
-        renderer : :obj:`str`, default='browser'
+        renderer : :obj:`str`, default=None
             Plotly renderer to be used.
 
         """
         if self.figure is not None:
             self.figure.show(renderer=renderer)
-            return self.figure
```
hndgzkn, line 118: Was there a special reason to return the figure?

Remi-Gau: no idea

Remi-Gau:
by any chance, can you check if that helps with this issue: https://github.com/nilearn/nilearn/issues/4379
it may require digging a bit in the example to see where plotly could be used but is not (engine set to matplotlib for example).

We could merge this PR and then tackle 4379 in a follow up if you prefer.

nilearn/plotting/displays/_figures.py
hndgzkn: Can we say that there is no inconvenience for removing that?

hndgzkn:
> by any chance, can you check if that helps with this issue: #4379 it may require digging a bit in the example to see where plotly could be used but is not (engine set to matplotlib for example).
> 
> We could merge this PR and then tackle 4379 in a follow up if you prefer.

Does it mean trying reverting the changes in #5020 ?

nilearn/plotting/displays/_figures.py
Remi-Gau: I would say so

Remi-Gau:
> Does it mean trying reverting the changes in #5020 ?

yup let's try to see if reverting works or if the examples still hang

hndgzkn:
> > Does it mean trying reverting the changes in #5020 ?
> 
> yup let's try to see if reverting works or if the examples still hang

As far as I see, there are others, but can be a starting point to test.

Remi-Gau:
> As far as I see, there are others, but can be a starting point to test.

yup: if we see a failure in CI let's not try further

Remi-Gau:
hmmm... the expected figures do not seem to appear here: https://output.circle-artifacts.com/output/job/6f0d7458-38f3-416c-b7fb-42d3edf425ca/artifacts/0/dev/auto_examples/01_plotting/plot_3d_map_to_surface_projection.html

hndgzkn:
> > As far as I see, there are others, but can be a starting point to test.
> 
> yup: if we see a failure in CI let's not try further

It does not complain but it does not render either, we need to set the correct renderer.

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -100,10 +99,11 @@
     bg_on_data=True,
     engine=engine,  # Specify the plotting engine here
 )
+figure.show()
 
 # Uncomment the line below
 # to view the figure in browser.
-figure.show()
+# figure.show(renderer="browser")
```
Remi-Gau, line 106: I just run this example as a script and things open in a browser with the `figure.show()` from above.

So maybe we would not need this.

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -100,10 +99,11 @@
     bg_on_data=True,
     engine=engine,  # Specify the plotting engine here
 )
+figure.show(renderer="sphinx_gallery")
```
hndgzkn: I've tried `renderer="sphinx_gallery"` to test behavior. 

It does not work, because `sphinx_gallery` requires `IPython` to render the image. Raises

```
ValueError: Mime type rendering requires ipython but it is not installed
```

One possible solution is:

- install IPython during CI job
- set  `PLOTLY_RENDERER: sphinx_gallery` as environment variable in CI job.
- use `figure.show()` in the example and when run in CI, it will use renderer defined as environment variable, otherwise use the default renderer of where it is run. 

Other possible solution is:
- Instead of `figure.show()` write `figure.figure` in the example, then sphinx_gallery knows how to render without IPython.

@Remi-Gau Which would be preferable?

Remi-Gau: > * Instead of `figure.show()` write `figure.figure` in the example

out of curiosity will that preserve the expected behavior when running the example both as a script or as a notebook (also in a marimo notebook)? 

you can get a notebook version of an example from the bottom of the page in the doc:

image

Remi-Gau: >     * install IPython during CI job
> 
>     * set  `PLOTLY_RENDERER: sphinx_gallery` as environment variable in CI job.
> 
>     * use `figure.show()` in the example and when run in CI, it will use renderer defined as environment variable, otherwise use the default renderer of where it is run.

that would be for the CI job of the doc build only?

I have a small preference for this one as it would make us rely on a method called 'show" which would be consistent with the matplotlib case

hndgzkn: Yes, it is expected to preserve both when run as script, and as notebook.

hndgzkn: I don't have experience with marimo, I can check that.

hndgzkn: > that would be for the CI job of the doc build only?
> 

Actually, this will solve the problem to generate docs in CI job. And `PLOTLY_RENDERER` will only be set in CI. 

However, if someone tries to build the doc locally, if no notebook, etc is installed locally, I think the doc generated will not have these images.

We might add IPython to `doc` dependencies.

hndgzkn: Only returning the figure, sphinx-gallery can display it.

nilearn/plotting/displays/_figures.py
hndgzkn: Finally, I understand that the reason to return figure is to display it when docs are generated by sphinx-gallery. However, when we call `self.figure.show` and return the figure, in local the image is displayed twice with jupyter, in generated just docs once, and script can display it in browser.

However, if we do not call `figure.show`, when the example is run as a script, it is not displayed.

nilearn/plotting/displays/_figures.py
```diff
@@ -104,18 +108,20 @@ def __init__(self, figure=None, output_file=None, hemi=DEFAULT_HEMI):
...
             Plotly renderer to be used.
 
         """
         if self.figure is not None:
-            self.figure.show(renderer=renderer)
-            return self.figure
+            if is_sphinx_build():
+                return self.figure
+            else:
+                self.figure.show(renderer=renderer)
```
hndgzkn, line 128: @Remi-Gau Finally the problem is fixed with this solution that I am not really happy to introduce a `sphinx_build` check. See https://output.circle-artifacts.com/output/job/0a27d278-1f65-429d-812d-d8825b02ac3e/artifacts/0/dev/auto_examples/01_plotting/plot_3d_map_to_surface_projection.html

[Sphinx-gallery requires](https://sphinx-gallery.github.io/dev/auto_plotly_examples/plot_0_plotly.html#sphx-glr-auto-plotly-examples-plot-0-plotly-py) to have the figure object in the last line of the cell. So figure.show should return object for sphinx_gallery. 

Otherwise, notebook displays already the image even if we set `renderer=None` which is default. And if we return the figure as well, it displays twice. So I am not returning figure if it is not `sphinx_gallery` who is running the code.

This way, it works with doc build, notebook and script.

Do you think is OK to introduce sphinx check?

We might in the future create `MatplotlibSurfaceFigure` object may be which would have the same interface and behavior with `PlotlySurfaceFigure`.

hndgzkn: Using figure.figure does not display the plotly figure when run as a script. If that is not critical, we can remove sphinx check and use that solution as well.

Remi-Gau: > This way, it works with doc build, notebook and script.
> 
> Do you think is OK to introduce sphinx check?

yup that seems fine to me

Remi-Gau: > We might in the future create `MatplotlibSurfaceFigure` object may be which would have the same interface and behavior with `PlotlySurfaceFigure`.

possibly though I wonder if that's not over-engineering...

Remi-Gau: > Using figure.figure does not display the plotly figure when run as a script. If that is not critical, we can remove sphinx check and use that solution as well.

hum... I'd be tempted to have a fix for the current issues and maybe open another issue to remember to get back to this.

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -195,16 +195,19 @@
...
         levels=regions_indices,
         labels=labels,
         lines=[{"width": 5}],
     )
-    # view the contours in a browser
-    figure.show()
+
+    # Uncomment the line below
+    # to view the plotly figure in browser.
+    # figure.show(renderer="browser")
```
Remi-Gau:
```suggestion
```

examples/01_plotting/plot_3d_map_to_surface_projection.py
Remi-Gau: I guess we can now remove line 104-106

nilearn/plotting/displays/_figures.py
```diff
@@ -104,18 +108,20 @@ def __init__(self, figure=None, output_file=None, hemi=DEFAULT_HEMI):
...
-        renderer : :obj:`str`, default='browser'
+        renderer : :obj:`str`, default=None
             Plotly renderer to be used.
 
         """
         if self.figure is not None:
-            self.figure.show(renderer=renderer)
-            return self.figure
+            if is_sphinx_build():
+                return self.figure
```
Remi-Gau, line 124: let's add a comment why we do this because it's not obvious when just reading the code

nilearn/plotting/displays/_figures.py
hndgzkn: > possibly though I wonder if that's not over-engineering...

It would be reasonable to have a single interface both for users and maintainers.

Remi-Gau: yeah I know, just wondering if it is really needed given we've been able to get by till now (not a great argument I know)

hndgzkn: It is a not priority, but I think would be good to do it sometime.

hndgzkn:
@Remi-Gau Is there any other example that should be converted to plotly?

Remi-Gau:
> Is there any other example that should be converted to plotly?

can't think of one just now

hndgzkn:
> can't think of one just now

Then my changes to this PR have finished. Doc seems to be OK. https://output.circle-artifacts.com/output/job/d200fcc5-5485-4fe8-8bac-d24b32684871/artifacts/0/dev/auto_examples/01_plotting/plot_3d_map_to_surface_projection.html

Remi-Gau:
> > Is there any other example that should be converted to plotly?
> 
> can't think of one just now

also we can still fix / update them later if we find them

hndgzkn:
> Then my changes to this PR have finished. Doc seems to be OK. 

 Except doc failure :/

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -290,6 +286,7 @@
 view
 # view.open_in_browser()
```
Remi-Gau, line 297: maybe let's make sure that all examples using `open_in_browser` have an explanation why it's commented or when to uncomment it

hndgzkn: In that case, shall we add the comment for plotly part, as far as I understand they have the same purpose, to view interactively in the browser.

Remi-Gau: yes indeed

Remi-Gau:
> > Then my changes to this PR have finished. Doc seems to be OK.
> 
> Except doc failure :/

I think that's 500 error on fetching some data: a classic

examples/01_plotting/plot_3d_map_to_surface_projection.py
hndgzkn: Done!

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -288,15 +290,23 @@
...
 # We don't need to do the projection ourselves, we can use
 # :func:`~nilearn.plotting.view_img_on_surf`:
 from nilearn.plotting import view_img_on_surf
 
 view = view_img_on_surf(stat_img, threshold="90%")
 
 view
+# If plotly is not installed, it is still possible to have
+# interactive visualization in the browser by uncommenting
+# the below line.
```
Remi-Gau: is it not more because if we do not do open_in_browser when runnning as a script nothing shows up?

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -288,15 +290,23 @@
 # In a notebook, if ``view`` is the output of a cell,
 # it will be displayed below the cell
 view
+
+# If plotly is not installed, it is still possible to have
+# interactive visualization in the browser by uncommenting
+# the below line.
```
hndgzkn:
```suggestion

# If plotly is not installed or the code is run in script mode, 
# it is still possible to have interactive visualization in the 
# browser by uncommenting the below line.
```

examples/01_plotting/plot_3d_map_to_surface_projection.py
hndgzkn:
```suggestion

# If plotly is not installed or the code is run in script mode, 
# it is still possible to have interactive visualization in the 
# browser by uncommenting the below line.
```

examples/01_plotting/plot_3d_map_to_surface_projection.py
```diff
@@ -327,6 +337,10 @@
 )
 
 view
+
+# If plotly is not installed, it is still possible to have
+# interactive visualization in the browser by uncommenting
+# the below line.
```
hndgzkn:
```suggestion

# If plotly is not installed or the code is run in script mode, 
# it is still possible to have interactive visualization in the 
# browser by uncommenting the below line.
```

examples/01_plotting/plot_3d_map_to_surface_projection.py
hndgzkn: Updated, let me know if that's fine.

Remi-Gau:
got a bunch of tests failings because they are slow: @hndgzkn can you mark them with `@pytest.mark.slow`

https://github.com/nilearn/nilearn/actions/runs/21394178328/job/61588326301?pr=5949#step:8:16050

examples/01_plotting/plot_3d_map_to_surface_projection.py
bthirion: Should they be removed ?

PR #5950, opened 2026-01-16
[DOC] explain how contrasts names are related to saved GLM files

### LINKED ISSUE #5916: [ENH] Don't automatically rename contrasts with nilearn.interfaces.save_glm_to_bids

I am using 

`nilearn.interfaces.save_glm_to_bids` to compute contrasts for my first level analysis and save the relevant contrast images to my hard drive.  Right now this function will automatically rename my files which can get a bit tedious in the downstream analyses where I rely on deterministic outputs. Currently, it's not entirely clear to me:

1.) Why the contrasts are renamed? Is this simply for better readability? If yes, I don't find that the "improved" contrast names necessarily provide better readability than the names that I have set.
2.) Under which circumstances are the contrasts renamed? When is this behavior triggered?
2.) How the contrasts are renamed (i.e. "feedbackWinGtfeedbackNeutral" becomes "feedbacklossgtfeedbackneutral" - see 1.: I find my camel-case style more readable than the adapted name)

Right now, I'd rather rely on a deterministic behavior with simple breaking-behavior (i.e. simply fail when user provided invalid characters that can not be part of a filename on the hard drive)

### Benefits to the change

More deterministic behavior.

Remi-Gau:
- Closes #5916

- improve doc strings of save_glm_to_bids to explain how contrast name affects output filename

### DIFF
--- a/nilearn/_utils/tests/test_estimator_checks.py
+++ b/nilearn/_utils/tests/test_estimator_checks.py
@@ -2,6 +2,7 @@
 
 import pytest
 
+from nilearn._base import NilearnBaseEstimator
 from nilearn._utils.estimator_checks import (
     check_img_estimator_dict_unchanged,
     check_img_estimator_fit_check_is_fitted,
@@ -20,7 +21,7 @@ def test_check_estimator_has_sklearn_is_fitted():
     - running sklearn check_is_fitted on masker throws an error
     """
 
-    class DummyEstimator:
+    class DummyEstimator(NilearnBaseEstimator):
         def __init__(self):
             pass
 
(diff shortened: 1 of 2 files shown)


Remi-Gau:
@JohannesWiesner
can you have a look at the updated doc string in this PR to see if explains better how contrast names affect how output files are named?

nilearn/glm/io.py
```diff
@@ -129,6 +129,77 @@ def save_glm_to_bids(
         Arrays may be 1D or 2D, with 1D arrays typically being
         t-contrasts and 2D arrays typically being F-contrasts.
 
+        ..  admonition:: Contrasts names and output filenames
+            :class: important
+
+            It is import to note that the content of ``contrasts``
```
hndgzkn:
```suggestion
            It is important to note that the content of ``contrasts``
```

PR #5955, opened 2026-01-19
[REF] move cache mixin to nilearn._base

Remi-Gau:
- Closes none

- This PR is more a suggestion to discuss if the some classes used by many estimators should be moved to the nilearn._base module


Remi-Gau:
@bthirion @emdupre @hndgzkn 
this PR is more of a suggestion than anything strictly needed but given that we now a _base module with NilearnBaseEstimator, I was wondering if it would not also make sense to put there several other things that are used across the library (a bit the same way sklearn has several of its Mixin in sklearn.base)

Remi-Gau:
not convnved this makes a lot of sense anymore

closing for now

PR #5943, opened 2026-01-14
[MAINT] Run test on python free threaded: test maskers, glm, composition

Remi-Gau:
- Follow up to #5859

Changes proposed in this pull request:

- mark tests as thread_unsafe

PR #5956, opened 2026-01-19
[MAINT] check inputs of several function / methods

Remi-Gau:
- Closes none

- rely on the fact that most string substitution in doc strings have consistent type to do input validation of more function / methods


Remi-Gau:
> Just worried about the CI...

- I am trying to keep track of failing free threaded fails and flag the failing tests in a separate PR (#5943)

PR #5960, opened 2026-01-20
[MAINT] allow to skip free threaded tests

### LINKED ISSUE #5959: [BUG] Free threaded python tests do not check skip flags

### nilearn version

0.13.1.dev19+g3563c9085.d20260115

### Expected behavior

When I set `[skip test]` in commit message, the tests are expected to be skipped on CI. However `test_with_free_threaded_python` is run.

This is what I got:

`test_with_free_threaded_python` is run on CI.

hndgzkn:
- Closes #5959 

- Extract `check_skip_flags` to `‎.github/workflows/check_skip_flags.yml` and reuse it from `.github/workflows/test_with_tox.yml` and `.github/workflows/test_with_free_threaded_python.yml`

### DIFF
--- a/.github/workflows/check_skip_flags.yml
+++ b/.github/workflows/check_skip_flags.yml
@@ -3,7 +3,7 @@
 #
 # To reuse from a job, use:
 #
-#     uses: ./.github/workflows/job_skip_flags.yml
+#     uses: ./.github/workflows/check_skip_flags.yml
 #
 ###
 name: Check skip flags
--- a/.github/workflows/test_with_free_threaded_python.yml
+++ b/.github/workflows/test_with_free_threaded_python.yml
@@ -35,7 +35,7 @@ env:
 
 jobs:
     check_skip_flags:
-        uses: ./.github/workflows/job_skip_flags.yml
+        uses: ./.github/workflows/check_skip_flags.yml
 
     free_threaded:
         if: github.repository == 'nilearn/nilearn'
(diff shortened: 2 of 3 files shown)


.github/workflows/check_skip_flags.yml
```diff
@@ -0,0 +1,20 @@
+---
```
Remi-Gau, line 1: you need a "frontmatter" comment explaining what the workflow does otherwise the doc build will fail:

https://github.com/nilearn/nilearn/actions/runs/21167198699/job/60874988858?pr=5960#step:14:86

check the top of the other workflow for inspiration. 😜

hndgzkn: Ok, thanks, I've added.

Is it OK to name it as `job_skip_flags`?

Seems to work.

hndgzkn:
@Remi-Gau Can't we unify this with `skip doc` using a job variable?

.github/workflows/check_skip_flags.yml
Remi-Gau: yup that's a good name

Remi-Gau:
> @Remi-Gau Can't we unify this with `skip doc` using a job variable?

no idea

I have not yet played around with reusable workflow in github CI: feel free to try

Remi-Gau:
> Can't we unify this with `skip doc` using a job variable?

wanna try to do this in this PR?

hndgzkn:
> > Can't we unify this with `skip doc` using a job variable?
> 
> wanna try to do this in this PR?

I think it would be better to merge this one now. I found out this trying to avoid tests in #5949 .

But if you agree that it's a good idea, I can try it later.

Remi-Gau:
OK merging this then

PR #5961, opened 2026-01-20
[MAINT] deal with sklearn 1.8 deprecations

### LINKED ISSUE #5941: sklearn 1.8 deprecations
- [ ] sklearn/linear_model/_logistic.py:1811

```
'penalty' was deprecated in version 1.8 and will be removed in 1.10.
To avoid this warning, leave 'penalty' set to its default value
and use 'l1_ratios' instead.
Use l1_ratios=(0,) instead of penalty='l2' and l1_ratios=(1,) instead of penalty='l1'. 
```

- [ ] sklearn/linear_model/_logistic.py:1823
```
The fitted attributes of LogisticRegressionCV will be simplified in scikit-learn 1.10
to remove redundancy. Set`use_legacy_attributes=False` to enable the new behavior now,
or set it to `True` to silence this warning during the transition period
while keeping the deprecated behavior for the time being.
The default value of use_legacy_attributes will change from True to False in scikit-learn 1.10.
See the docstring of LogisticRegressionCV for more details. 
```

- [ ] sklearn/linear_model/_logistic.py:1780
```
The default value for l1_ratios will change from None to (0.0,) in version 1.10.
From version 1.10 onwards, only array-like with values in [0, 1] will be allowed,
None will be forbidden.
To avoid this warning, explicitly set a value, e.g. l1_ratios=(0,). 
```

Remi-Gau:
- Closes #5941

- deal with sklearn 1.8 deprecations
- extracted from #5942
- move most version comparing code to a separate module

### DIFF
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -138,7 +138,6 @@ from nilearn.surface import SurfaceImage
 from nilearn.surface.surface import get_data as get_surface_data
 from nilearn.surface.utils import assert_surface_image_equal
 
-SKLEARN_GTE_1_4 = compare_version(sklearn_version, ">=", "1.4.0")
 SKLEARN_GTE_1_5 = compare_version(sklearn_version, ">=", "1.5.0")
 SKLEARN_GTE_1_6 = compare_version(sklearn_version, ">=", "1.6.0")
 
@@ -249,8 +248,9 @@ def return_expected_failed_checks(
             "check_transformer_data_not_an_array": "TODO",
             "check_transformer_general": "TODO",
         }
-        if SKLEARN_GTE_1_4:
+        if SKLEARN_GTE_1_5:
             expected_failed_checks.pop("check_estimator_sparse_data")
+        if SKLEARN_GTE_1_6:
             expected_failed_checks |= {
                 "check_transformer_preserve_dtypes": "TODO",
             }
@@ -269,7 +269,7 @@ def return_expected_failed_checks(
             "check_estimator_sparse_matrix": "TODO",
             "check_estimator_sparse_tag": "TODO",
         }
-        if SKLEARN_GTE_1_4:
+        if SKLEARN_GTE_1_5:
             expected_failed_checks.pop("check_estimator_sparse_data")
         if isinstance(estimator, GroupSparseCovarianceCV):
             expected_failed_checks |= {


bthirion:
Loos like we're getting close.

nilearn/_utils/estimator_checks.py
```diff
@@ -4283,6 +4229,14 @@ def check_multimasker_generate_report(estimator_orig):
...
+            warnings_msg_to_check = [
+                "SurfaceMapsMasker has not been transformed",
+                "masker only has 6 map",
+            ]
         # TODO add a warning
         estimator.fit(input_img)
-        generate_and_check_masker_report(estimator)
+        generate_and_check_masker_report(
+            estimator, warnings_msg_to_check=warnings_msg_to_check
+        )
```
Remi-Gau, line 4242: removing the @ignore_warnings() forces us to deal with what warnings are thorwn

nilearn/_utils/estimator_checks.py
```diff
@@ -4283,6 +4229,14 @@ def check_multimasker_generate_report(estimator_orig):
         )
 
     else:
+        warnings_msg_to_check = []
+        if isinstance(estimator, SurfaceMapsMasker):
+            warnings_msg_to_check = [
+                "SurfaceMapsMasker has not been transformed",
+                "masker only has 6 map",
```
bthirion:
```suggestion
                "masker only has 6 maps",
```

nilearn/decoding/decoder.py
```diff
@@ -58,13 +60,27 @@
 from nilearn.regions.rena_clustering import ReNA
 from nilearn.surface import SurfaceImage
 
+SKLEARN_GET_1_8 = parse(sklearn_version).release[1] >= 8
```
hndgzkn: Could we extract this as a function in _utils which takes version number, lt, or gt and returns bool ?

Remi-Gau: actually I think we have a function for that: let me check

nilearn/_utils/versions.py
Remi-Gau, line 1: centralize most of the version related things and constants into a single place

nilearn/_utils/estimator_checks.py
```diff
@@ -140,7 +138,9 @@
 from nilearn.surface.surface import get_data as get_surface_data
 from nilearn.surface.utils import assert_surface_image_equal
 
-SKLEARN_MINOR = parse(sklearn_version).release[1]
+SKLEARN_GTE_1_4 = compare_version(sklearn_version, ">=", "1.4.0")
+SKLEARN_GTE_1_5 = compare_version(sklearn_version, ">=", "1.5.0")
+SKLEARN_GTE_1_6 = compare_version(sklearn_version, ">=", "1.5.0")
```
```suggestion
SKLEARN_GTE_1_6 = compare_version(sklearn_version, ">=", "1.6.0")
```

PR #5964, opened 2026-01-20
[REF] add _mask_img property to GLM

Remi-Gau:
- Closes none

- add an easy way to access mask image used during fitting of a GLM model

### DIFF
--- a/nilearn/glm/_base.py
+++ b/nilearn/glm/_base.py
@@ -51,6 +51,7 @@ from nilearn.typing import ClusterThreshold, HeightControl
 FIGURE_FORMAT = "png"
 
 
+@fill_doc
 class BaseGLM(CacheMixin, NilearnBaseEstimator):
     """Implement a base class \
     for the :term:`General Linear Model<GLM>`.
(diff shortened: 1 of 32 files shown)


nilearn/glm/_base.py
```diff
@@ -97,6 +99,20 @@ def _is_first_level_glm(self):
...
+            return self.masker_.mask_img_
+        else:
+            if self.mask_img is None:
+                return None
+            try:
+                # load mask_img if is a niiimg-like object
+                return check_niimg(self.mask_img)
+            except Exception:
+                return self.mask_img
+
```
hndgzkn, line 115: Could the property be public `mask_img` and internal attribute be private `self._mask_img`?

Remi-Gau: > Could the property be public `mask_img`

nope because we the GLM estimators have a parameter `mask_img` that (by sklearn convention) MUST be stored in an attribute `mask_img` that (also by sklearn convention) MUST NOT be changed after instantiation

Maybe for the 0.14.0 release, we can add a public `self.mask_img_` (that would be created during fit) but for 0.13.1 I just did not wanted to add anything public.

PR #5967, opened 2026-01-21
[MAINT] Extract common github workflow job

hndgzkn:
- Closes None

- Define `skip_type` variable in job definition of `check_skip_flags` in `.github/workflows/check_skip_flags.yml` 
- Define `test` and `doc` as variables for test and doc build workflows and reuse `check_skip_flags` from both.
- Do not run `validate_cff` and `quality_control` in doc build if `[skip doc]` is set.

### DIFF
--- a/doc/ci.jinja
+++ b/doc/ci.jinja
@@ -8,7 +8,7 @@ Please note that if one of the following markers appear in the latest commit mes
 ============================ ===================
 Commit message marker        Action taken by CI
 ============================ ===================
-[skip ci]                    Gtihub CI is skipped completely. Several other options are also possible, see `github documentation <https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs>`_.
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docs.yml
```diff
@@ -157,26 +164,10 @@ jobs:
...
-                if [[ $headCommitMsg == *"[skip doc]"* ]]; then
-                    echo "skipping doc build"
-                    exit 1
-                fi
-
     # Check the build type and files examples to generate
     # Upload as an artifact for other jobs.
     build_type:
-        needs: [check_skip_flags, validate_cff]
+        needs: [validate_cff]
```
hndgzkn, line 170: I've finally put this to needs section of `validate_cff` and `quality_control` thinking that if we want to skip doc build, we won't need them either. If there is an inconvenience, I can revert this change.

Remi-Gau: na that's fine

bthirion (approved):
LGTM, thx.

Issue #3568, opened 2023-03-06
Adding type annotation to nilearn


Remi-Gau:
This is a general issues to discuss the pros, cons of adding type annotations to the nilearn code base. And also to eventually discuss strategies to start adding type annotations.

### Pros

- improved internal consistency can help catch bug early
- more standardized type documentation rather than "just" relying on definitions in doc strings
- improved developer experience as many IDEs are type-aware

### Cons

- raises the bar for new contributions
- may lead "complex" types (for example: `Union[str, list[str], Path, list[Path], ...]`) for some APIs or uninformative ones (for example: `Any`)
- the usefulness of type annotation of nilearn is limited by how well nilearn's dependencies are themselves annotated
- may require substantial effort

### Strategies

Some suggestions that are not mutually exclusive

- use [monkeytype](https://monkeytype.readthedocs.io/en/stable/index.html) or similar tools to help automate the process
- add types gradually modules by module
- add types gradually when functions / methods are extracted

tsalo:
>may lead "complex" types (for example: `Union[str, list[str], Path, list[Path], ...]`) for some APIs or uninformative ones (for example: `Any`)

From previous dev meetings, I think this one was the biggest sticking point. A lot of nilearn tools accept many different datatypes for each input.

Remi-Gau:
> > may lead "complex" types (for example: `Union[str, list[str], Path, list[Path], ...]`) for some APIs or uninformative ones (for example: `Any`)
> 
> From previous dev meetings, I think this one was the biggest sticking point. A lot of nilearn tools accept many different datatypes for each input.

Put it there because @jeromedockes had mentioned it to me.

Though doesn't that mean that the doc strings documenting those API **already** have a very complex list of allowed types anyway?

ymzayek:
Let's make this a topic for the next core dev meeting. We should weigh the pros and cons of going in this direction and make some decisions together before implementing anything

Remi-Gau:
For the record.

Was decided to not add type annotation to the code base **for now** but that the issue could be revisited in a few months.

Was also mentioned that we don't have to aim for "100% coverage" in terms of annotation if we started adding them.

Remi-Gau:
regarding type annotation and documentation

Here is an example (from #3525) of how function definitions start looking in the doc when type annotation are added:

[...]

PR #5973, opened 2026-01-23
[TYPE] add type None when function / method returns None

Remi-Gau:
- Closes none
- Relates to #3568 

- add return type None to several methods / functions that do not return anything
- add return type book to `__sklearn_is_fitted__` method

### DIFF
--- a/maint_tools/check_docstrings.py
+++ b/maint_tools/check_docstrings.py
@@ -15,8 +15,8 @@
 """
 
 import ast
-import contextlib
 import re
+from contextlib import suppress
 from pathlib import Path
 
 from numpydoc.docscrape import NumpyDocString
@@ -61,7 +61,7 @@ def main() -> None:
         for func_def in list_functions(filename, include="all"):
             check_missing_return_annotation(func_def, filename)
 
-            docstring = _get_doctring(func_def, filename)
+            docstring = _get_docstring(func_def, filename)
             if docstring is None:
                 continue
 
@@ -70,7 +70,7 @@ def main() -> None:
             check_returns_yields_and_annotation(func_def, filename)
 
         for class_def in list_classes(filename, include="all"):
-            if _get_doctring(func_def, filename) is not None:
+            if _get_docstring(class_def, filename) is not None:
                 check_fill_doc_decorator(class_def, filename)
                 check_docstring(class_def, filename)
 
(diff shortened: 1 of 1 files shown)


maint_tools/check_docstrings.py
```diff
@@ -7,6 +7,7 @@
 """
 
 import ast
+import contextlib
```
hndgzkn: May be just suppress can be imported.

maint_tools/check_docstrings.py
```diff
@@ -148,20 +153,15 @@ def check_docstring(ast_node, filename: str, n_issues: int) -> int:
             f"- {ast_node.name} - [red] No docstring detected"
         )
     else:
-        try:
+        missing = None
+        with contextlib.suppress(Exception):
```
hndgzkn: Is there a specific exception type expected?

maint_tools/check_docstrings.py
```diff
@@ -187,5 +187,109 @@ def get_missing(docstring: str, values=None) -> list[tuple[str, str, str]]:
...
+
+
+def docstring_has_yields_section(docstring: str) -> bool:
+    """Return True if docstring contains a Yields section."""
+    try:
+        doc = NumpyDocString(docstring)
+    except Exception:
+        return False
+
+    return bool(doc["Yields"])
```
hndgzkn: These two might be implemented as one function, for ex. `docsting_has_field(docstring:str, field:str)` .

nilearn/_utils/estimator_checks.py
```diff
@@ -3922,7 +3922,9 @@ def check_multimasker_transformer_sample_mask(estimator_orig):
 
 
 @ignore_warnings()
-def check_multimasker_transformer_high_variance_confounds(estimator_orig):
+def check_multimasker_transformer_high_variance_confounds(
+    estimator_orig,
+):
```
hndgzkn, line 3927:
```suggestion
) -> None:
```

Remi-Gau: I think I had tried this one but it then triggered some mypy errors, because adding type annotation to the function signature forces mypy to then check type consistency in the function itself: so in the end I did not add a return None type annotation, to stick to function / methods that could be very easily annotated.

More involved annotation could come later.

Issue #5460, opened 2025-06-18
Only check if image is empty when necessary


Remi-Gau:
I was wondering if instead of putting this check in `check_niimg`, creating a function `is_img_empty` and doing the check only where it is necessary would help with performance? Because `check_niimg` is called everywhere. If I am not mistaken, we have to check this only when `fit` is called, where it already needs to get image data.

_Originally posted by @hndgzkn in https://github.com/nilearn/nilearn/pull/5454#discussion_r2153884893_

bthirion:
Sounds reasonable.

Remi-Gau:
More detail in : https://github.com/nilearn/nilearn/blob/9476b07a07b4a1a7b5a1821b837381d78e920cea/nilearn/_utils/estimator_checks.py#L2898

The intention is that all our estimators should have suitable replacement for sklearn: https://github.com/scikit-learn/scikit-learn/blob/b76422979fc4d215d8063dc27e4b8acdcd33c60d/sklearn/utils/estimator_checks.py#L2370

PR #5980, opened 2026-01-27
[ENH] Improve image manipulations

hndgzkn:
- Closes None, Relates to https://github.com/nilearn/nilearn/issues/5460

- Improvements in `nilearn._utils.niimg`
  - Create `ensure_finite_data` function and reuse it where necessary
  - Create `has_non_finite` function and reuse it where necessary
  - Improve `is_binary_niimg` for memory usage and performance
  - Add tests for `ensure_finite_data`, `has_non_finite`, `is_binary_niimg` 
  - Do not load array in `load_niimg` if dtype is not specified
- Replace `get_data` with `_get_data` in `nilearn.image.image` where `check_niimg` is already called before `get_data`
- Update `safe_get_data` to copy image if image data is not already in the cache, or `copy_data=True` to ensure that getting data does not change the state of the image.

### DIFF
--- a/nilearn/glm/tests/test_first_level.py
+++ b/nilearn/glm/tests/test_first_level.py
@@ -185,6 +185,7 @@ def test_flm_fit_verbose(shape_4d_default, capsys):
     # assert len(stdout_verbose_3) > len(stdout_verbose_2)
 
 
+@pytest.mark.slow
 def test_glm_fit_valid_mask_img(shape_4d_default):
     """Run fit on FLM with different valid masks."""
     rk = 3


hndgzkn:
@Remi-Gau I am planning to work on this issue in a series of PRs simplifying the code, removing inconsistencies or unnecessary checks. It seems like that we are doing the same checks multiple times in the code.

nilearn/_utils/niimg.py
```diff
@@ -47,25 +45,36 @@ def safe_get_data(img, ensure_finite=False, copy_data=False) -> np.ndarray:
...
         nilearn.image.get_data return from Nifti image.
     """
-    if copy_data:
-        img = deepcopy(img)
+    data = _get_data(img)
 
-    if is_gil_enabled():
-        # typically the line below can double memory usage
-        # that's why we invoke a forced call to the garbage collector
-        gc.collect()
```
Remi-Gau, line 56: @man-shu do we have a benchmark to check the memory performance of this?

nilearn/datasets/atlas.py
```diff
@@ -2047,14 +2046,14 @@ def _download_talairach(talairach_dir, verbose) -> None:
...
+            atlas_img, labels, talairach_dir, verbose=verbose
+        )
     finally:
         shutil.rmtree(temp_dir)
-    labels_text = atlas_img.header.extensions[0].get_content()
-    multi_labels = labels_text.strip().decode("utf-8").split("\n")
-    labels = [lab.split(".") for lab in multi_labels]
-    _separate_talairach_levels(
-        atlas_img, labels, talairach_dir, verbose=verbose
-    )
```
hndgzkn, line 2057: Moved this above, as after putting `if dtype is not None` check in `load_niimg`, the data is not loaded to memory from the file. The file is needed in `_separate_talairach_levels` as it loads data to memory. If we delete the tmp file earlier, it raises error.

nilearn/image/image.py
```diff
@@ -2299,7 +2300,7 @@ def check_niimg(
 
     if ensure_ndim == 3 and len(niimg.shape) == 4 and niimg.shape[3] == 1:
         # "squeeze" the image.
-        data = safe_get_data(niimg)
+        data = _get_data(niimg)
```
hndgzkn: I am replacing `safe_get_data` with `_get_data` where no copy or ensure finite is specified, as they do the same thing with default params of `safe_get_data`.

nilearn/_utils/niimg.py
```diff
@@ -124,21 +134,23 @@ def load_niimg(niimg, dtype=None):
             + repr_niimgs(niimg, shorten=True)
         )
 
-    img_data = _get_data(niimg)
-    target_dtype = _get_target_dtype(img_data.dtype, dtype)
+    # avoid loading data if dtype is None
```
hndgzkn, line 156: Would that have a side effect? 
I thought unless we demand a certain `dtype` that would require type conversion of data, it would be good to avoid loading proxy array to numpy array.

nilearn/image/tests/test_image.py
```diff
@@ -1923,6 +1924,7 @@ def test_copy_img():
 
 @pytest.mark.thread_unsafe
 def test_copy_img_side_effect(img_4d_ones_eye):
+    _get_data(img_4d_ones_eye)
```
hndgzkn, line 1929: Added this line so that the proxarray is loaded to ndarray. Otherwise the object hashs will be different, as copy_img loads the data.

hndgzkn: This is also a side-effect of `if dtype is not None` check in `load_niimg`.

nilearn/_utils/niimg.py
```diff
@@ -154,11 +166,24 @@ def is_binary_niimg(niimg):
...
+    for i in range(0, flat.size, block_size):
+        block = flat[i : i + block_size]
+
+        # Check if only 0, 1, inf, or nan
+        mask = (
+            (block == 0)
+            | (block == 1)
+            | np.isnan(block)
+            | (block == np.inf)
+            | (block == -np.inf)
```
bthirion: should we consider that an imlage with 0's, 1's and np.inf is binary ?

hndgzkn: This code is just equivalent of the previous one. The previous one replaces `inf` and `nan` with 0 at line 157 using `safe_get_data(niimg, ensure_finite=True)` and then checks at line 160 if unique values are 0 or 1. 

If the behavior should be different, then we can modify this function.

nilearn/glm/first_level/first_level.py
```diff
@@ -1261,10 +1260,7 @@ def _prepare_mask(self, run_img):
             Used for setting up the masker object.
         """
         masker_type = "nii"
-        # all elements of X should be of the similar type by now
-        # so we can only check the first one
-        to_check = run_img[0] if isinstance(run_img, Iterable) else run_img
-        if not self._is_volume_glm() or isinstance(to_check, SurfaceImage):
```
hndgzkn, line 1267: Removed this because this private function is only called from one place where already a single image is set as parameter, so no need to check.

nilearn/_utils/niimg.py
hndgzkn: I've updated with a new parameter to accept inf values or not.

nilearn/_utils/niimg.py
hndgzkn: I've used the below code in local:
```
import psutil
import time

def benchmark_nifti(process_func, func_var, n_runs=10000):
    times = []
    memory = []
    process = psutil.Process(os.getpid())
    
    for _ in range(n_runs):
        start_mem = process.memory_info().rss / (1024**2)
        start = time.time()
        process_func(func_var)
        end = time.time()
        end_mem = process.memory_info().rss / (1024**2)
        times.append(end - start)
        memory.append(end_mem-start_mem)

    results = {
        "mean_time": np.mean(times),
        "mean_mem": np.mean(memory)
    }

    print("\nSummary:")
    
    print(f"mean time {results['mean_time']:.8f}s")
    print(f"mean memory {results['mean_mem']:.8f}mb")
```

nilearn/image/image.py
```diff
@@ -256,7 +257,7 @@ def smooth_array(arr, affine, fwhm=None, ensure_finite=True, copy=True):
         arr = arr.copy()
     if ensure_finite:
         # SPM tends to put NaNs in the data outside the brain
-        arr[np.logical_not(np.isfinite(arr))] = 0
+        ensure_finite_data(arr, raise_warning=False)
```
hndgzkn, line 260: I've silenced warnings as it did not warn in original code. Would it be better to raise them?

nilearn/image/resampling.py
```diff
@@ -474,9 +482,8 @@ def resample_img(
             )
 
     # noop cases
-    input_img_is_string = isinstance(img, str)
```
hndgzkn, line 477: This line was not serving its purpose as `img` can no longer be `str` after `checkniimg(img)` is called. I'moved that above to change the value of `copy` (copy is not used elsewhere and introducing `input_img_is_string` was not really necessary).

nilearn/image/image.py
hndgzkn: @Remi-Gau `safe_get_data` docstring says:

```
Get the data in the image without having a side effect \
    on the Nifti1Image object.
```
It makes me think of may be the initial intention was to set `copy=True`. Otherwise docstring becomes misleading. 

I doubt if I am doing good by replacing `safe_get_data` with `_get_data`. May be setting default `copy=True` might be more correct.

nilearn/_utils/niimg.py
bthirion: And the result was ... ?

nilearn/_utils/niimg.py
```diff
@@ -47,25 +46,53 @@ def safe_get_data(img, ensure_finite=False, copy_data=False) -> np.ndarray:
...
-        non_finite_mask = np.logical_not(np.isfinite(data))
-        if non_finite_mask.sum() > 0:  # any non_finite_mask values?
+        ensure_finite_data(data)
+    return data
+
+
+@overload
+def has_non_finite(
+    data: np.ndarray, return_mask: Literal[False] = False
+) -> bool: ...
```
bthirion: Sorry for my ignorance, but why do we need that ?

nilearn/_utils/niimg.py
```diff
@@ -47,25 +46,53 @@ def safe_get_data(img, ensure_finite=False, copy_data=False) -> np.ndarray:
...
+
+
+@overload
+def has_non_finite(
+    data: np.ndarray, return_mask: Literal[True]
+) -> tuple[bool, np.ndarray]: ...
+
```
bthirion: Same question.

nilearn/_utils/niimg.py
```diff
@@ -154,11 +190,38 @@ def is_binary_niimg(niimg):
...
+    """Create a boolean mask for values equal to 0 or 1."""
+    return (block == 0) | (block == 1)
+
+
+def _mask_with_nonfinite(mask_base):
+    """Extend specified mask function to include non-finite values."""
+    return lambda block: mask_base(block) | ~np.isfinite(block)
+
+
+def is_binary_data(data, block_size=1_000_000, accept_non_finite=True) -> bool:
```
bthirion, line 204: Sorry, I'm not familiar with `block_size=1_000_000`
What does it mean ?

nilearn/image/image.py
bthirion: I don't think so.

bthirion (commented):
Thx ! 
I don't understand all the patterns you use, please explain them.

nilearn/_utils/niimg.py
hndgzkn: It is equivalent to `block_size=1000000` but easier to read the number: https://peps.python.org/pep-0515/
I can revert if it is confusing.

nilearn/_utils/niimg.py
hndgzkn: `has_non_finite` returns `bool`, or `(bool, ndarray)` depending on the value of `return_mask`. It is mypy that complains if I do not do that. We have the similar for check_niimg, see https://github.com/nilearn/nilearn/blob/d8e781d9453edb331932700b059e51b1049bf83c/nilearn/image/image.py#L2130-L2164

nilearn/_utils/niimg.py
hndgzkn: I've reverted to original code. 

Before posting the benchmark results, I wanted to re-check and found out that I had a very fundamental mistake while running benchmarks. As default value for copy=False in `safe_get_data`, the data was not copied in the functions and the implementation I proposed ran faster. 

I was thinking that `deepcopy` could not be more performant than `ndarray.copy` for copying data. However I've realized that `deepcopy` actually does not copy data in cache and thus it is much faster. So I reverted the changes.

nilearn/image/image.py
hndgzkn: I've looked at the changes in `safe_get_data` with `blame`. 

[In the original implementation](https://github.com/nilearn/nilearn/commit/25b3ec4b57b205ce1f7a09443c32d1ade05de23a), if data was not loaded to cache, it was creating deepcopy of the image and sending the data of the copy, and did not fill the cache of the image. 
However if the image had already data in the cache, it would send it without copying. So calling this function had no effect on the image's state. 

```
    if hasattr(img, '_data_cache') and img._data_cache is None:
        # By loading directly dataobj, we prevent caching if the data is
        # memmaped. Preventing this side-effect can save memory in some cases.
       img = copy.deepcopy(img)
```

Then copy is introduced with default=False. Now even if the image data is not loaded to cache, the data is loaded and the state of image is changed.

I think instead of removing `if hasattr(img, '_data_cache') and img._data_cache is None`,   `or copy_data` could be added to the if condition. That way, keeping `safe_get_data` where I replace with `_get_data` makes sense in my opinion.

nilearn/_utils/niimg.py
Remi-Gau: I do find 1_000_000 more readable

nilearn/_utils/niimg.py
bthirion: OK, but indeed having functions with different return arguments is a rather bad pattern. Can't it be avoided ?

hndgzkn: Normally I would not prefer to have functions  with different return arguments. 

However in this case, when we check infinite data, we always create the mask, but we do not always use the mask. 
- One option is to return both values each time, but then it will not be possible to directly use the function in if conditions.   
- The other option is to return only bool value, which I did initially. Then in resampling for example, it was not possible to reuse this function. 

To check the existence of finite data, there were two implementations in the code. `np.any(non_finite_mask)` and `non_finite_mask.sum() > 0`, the first being more efficient. 

So finally I decided that this would be a better option, as it is a helper function in private API and documented and it helps improve readability of the main code and provide a single implementation.

If it seems more confusing, I can revert it.

hndgzkn: One source of confusion is also specifying the return types. May be we do not need to specify return types for private functions. But I think it helps while developping.

Remi-Gau: suggestion: let's have `has_non_finite` always return `tuple[bool, np.ndarray]` and ignore the second value when we do not need it.

what do you think?

nilearn/_utils/niimg.py
Remi-Gau: I cannot think of an issue with this.
I am tempted to add a return type annotation on this `load_niimg` to maybe have some static type checking to help make sure we are not missing something

nilearn/_utils/niimg.py
```diff
@@ -154,11 +197,38 @@ def is_binary_niimg(niimg):
...
+    data = niimg.dataobj
+    return is_binary_data(data, block_size, accept_non_finite)
+
+
+def _base_mask(block: np.ndarray) -> np.ndarray:
+    """Create a boolean mask for values equal to 0 or 1."""
+    return (block == 0) | (block == 1)
+
+
+def _mask_with_nonfinite(mask_base):
```
Remi-Gau:
```suggestion
def _mask_with_nonfinite(mask_base: np.ndarray):
```

just to make sure I follow

nilearn/image/image.py
Remi-Gau: > I think instead of removing `if hasattr(img, '_data_cache') and img._data_cache is None`, `or copy_data` could be added to the if condition. That way, keeping `safe_get_data` where I replace with `_get_data` makes sense in my opinion.

yes I think that makes sense

Remi-Gau:
except for this point (https://github.com/nilearn/nilearn/pull/5980#discussion_r2757869135) the rest makes sense to me AFAICT

nilearn/image/image.py
hndgzkn: I've updated the condition as below as `ensure_data=True` also changes the data:

```python
if not img.in_memory or ensure_finite or copy_data:
        img = deepcopy(img)
```

Let me know if it is still ok. 
I am reverting `safe_get_data` to `get_data` changes.

nilearn/_utils/niimg.py
hndgzkn: I've simplified this. 
It is a callable rather than an np.ndarray. But I think that was unnecessarily complex.

nilearn/_utils/niimg.py
hndgzkn: I think that might be good.

Shall I add it? If I  put `SpatialImage`, would that be correct?

nilearn/_utils/niimg.py
hndgzkn: Ok, I've updated.

nilearn/image/image.py
hndgzkn: Finally I changed it as:

```
if not img.in_memory or copy_data:
        img = deepcopy(img)
```
And updated docstring of `safe_get_data`. Adding `ensure_finite` to condition changes behavior of `threshold_img`. It returns always copied data.

PR #5982, opened 2026-01-29
[ENH] Allow plot_<> functions to accept threshold in percentile

### LINKED ISSUE #2061: Allow plot_<> functions to accept threshold in percentile
Currently, `plotting.view_<>` family of functions can accept a threshold value in percentile as a string, but `plotting.plot_<>` family of functions do not.
So `plotting.view_img(img, threshold='97%')` works, but `plotting.plot_stat_map(img, threshold='97%')` does not.
They should both support string percentile values.

It makes the interface consistent and yes, the functions easier to use.

-->

saeed-neuro:
- Closes #2061

- We addes a threshold check call
- We tested locally and we got new behaviour

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -37,6 +37,21 @@ linkcode_resolve = make_linkcode_resolve
 # We also add the directory just above to enable local imports of nilearn
 sys.path.insert(0, str(Path("..").absolute()))
 
+# -- Plotly Configuration ----------------------------------------------------
+
+try:
+    import plotly.io as pio
+
+    pio.renderers.default = "sphinx_gallery"
+except ImportError:
+    import warnings
+
+    warnings.warn(
+        stacklevel=2,
+        message="Plotly is not installed. Plotly figures will not be shown.",
+        category=UserWarning,
+    )
+
 # -- General configuration ---------------------------------------------------
 
 # Add any Sphinx extension module names here, as strings. They can be
--- a/examples/01_plotting/plot_3d_map_to_surface_projection.py
+++ b/examples/01_plotting/plot_3d_map_to_surface_projection.py
@@ -1,3 +1,4 @@
+# %%
 """
 Making a surface plot of a 3D statistical map
 =============================================
@@ -81,12 +82,10 @@ fig.show()
 # you can easily configure :func:`~nilearn.plotting.plot_surf_stat_map`
 # to use ``plotly`` instead of ``matplotlib``:
 
-engine = "matplotlib"
+# If plotly is not installed, use matplotlib
+from nilearn._utils.helpers import is_plotly_installed
 
-# uncomment the following line if you use plotly
-# in the rest of this example
-
-# engine = "plotly"
+engine = "plotly" if is_plotly_installed() else "matplotlib"
 
 print(f"Using plotting engine {engine}.")
 
@@ -100,10 +99,11 @@ figure = plot_surf_stat_map(
     bg_on_data=True,
     engine=engine,  # Specify the plotting engine here
 )
-
-# Uncomment the line below
-# to view the figure in browser.
 figure.show()
+
+# Uncomment the line below to have interactive
+# visualization in the browser
+# figure.show(renderer="browser")
 
 # %%
 # When using ``matplolib`` as the plotting engine, a standard
@@ -186,7 +186,7 @@ figure = plot_surf_stat_map(
     engine=engine,
 )
 if engine == "matplotlib":
-    plot_surf_contours(
+    figure = plot_surf_contours(
         roi_map=destrieux_atlas,
         hemi=hemi,
         labels=labels,
@@ -195,7 +195,6 @@ if engine == "matplotlib":
         legend=True,
         colors=["g", "k"],
     )
-    show()
 elif engine == "plotly":
     figure.add_contours(
         roi_map=destrieux_atlas,
@@ -203,8 +202,11 @@ elif engine == "plotly":
         labels=labels,
         lines=[{"width": 5}],
     )
-    # view the contours in a browser
-    figure.show()
+    # Uncomment the line below to have interactive
+    # visualization in the browser
+    # figure.show(renderer="browser")
+
+figure.show()
 
 # %%
 # Plot with higher-resolution mesh
@@ -288,8 +290,13 @@ view = view_surf(
 # In a notebook, if ``view`` is the output of a cell,
 # it will be displayed below the cell
 view
+
+# If plotly is not installed or the code is run in script mode,
+# it is still possible to have interactive visualization in the
+# browser by uncommenting the below line.
 # view.open_in_browser()
 
+# %%
 # We don't need to do the projection ourselves, we can use
 # :func:`~nilearn.plotting.view_img_on_surf`:
 from nilearn.plotting import view_img_on_surf
@@ -297,6 +304,10 @@ from nilearn.plotting import view_img_on_surf
 view = view_img_on_surf(stat_img, threshold="90%")
 
 view
+
+# If plotly is not installed or the code is run in script mode,
+# it is still possible to have interactive visualization in the
+# browser by uncommenting the below line.
 # view.open_in_browser()
 
 # %%
@@ -327,6 +338,10 @@ view = view_img_on_surf(
 )
 
 view
+
+# If plotly is not installed or the code is run in script mode,
+# it is still possible to have interactive visualization in the
+# browser by uncommenting the below line.
 # view.open_in_browser()
 
 # sphinx_gallery_dummy_images=1
--- a/nilearn/_utils/helpers.py
+++ b/nilearn/_utils/helpers.py
@@ -292,3 +292,7 @@ def is_gil_enabled():
     except AttributeError:
         # sys._is_gil_enabled does not exist in standard Python builds
         return True
+
+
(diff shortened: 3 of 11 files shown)


Remi-Gau:
there is a bunch of tests failing on this:

[.github/workflows/test_with_tox.yml](https://github.com/nilearn/nilearn/actions/runs/21495890327/job/61970769890?pr=5982#step:8:13394)

emdupre:
> there is a bunch of tests failing on this:

> [.github/workflows/test_with_tox.yml](https://github.com/nilearn/nilearn/actions/runs/21495890327/job/61970769890?pr=5982#step:8:13394)

Thanks, @Remi-Gau ! We just walked through the whole PR workflow 👏 @saeed-neuro !
There was one flaky test in canICA it looked like (checking if it re-appears on re-run) and some actual failures in GLM reporting, though that's more to do with warning monitoring in those tests, so a little beyond a GFI. I told @saeed-neuro we might have to do a bit of maintainer edits on this branch to get it over the finish line and merged in 🚀

Remi-Gau:
> There was one flaky test in canICA it looked like (checking if it re-appears on re-run)

those seem to be gone on rerun

Remi-Gau:
> though that's more to do with warning monitoring in those tests, so a little beyond a GFI

let me take care of those for you

Remi-Gau:
@saeed-neuro 
as this is a a new feature, this will be merged for the 0.14.0 release: so I am going to set the PR as 'draft' for now, so we do not merge too early by mistake

emdupre:
Thank you, @Remi-Gau !!

Remi-Gau (approved):

nilearn/plotting/image/img_plotting.py
```diff
@@ -231,6 +231,13 @@ def _plot_img_with_bg(
             # voxels pass the threshold
             threshold = float(fast_abs_percentile(data)) - 1e-5
 
+        if threshold is not None:
+            threshold = check_threshold(
+                threshold,
+                data,
+                percentile_func=fast_abs_percentile,
+                name="threshold",
+            )
```
hndgzkn: This change is made in `_plot_img_with_bg` which is called directly or indirectly from almost all image plotting functions. 

One point is docstrings are feeded from `nilearn._utils.docs` which does not indicate percentage `str` as threshold value. That would be good to indicate that.

The second point is as `check_threshold_not_negative(threshold)` is already done in `_plot_img_with_bg`, it could be removed from other functions that call `_plot_img_with_bg`. (This is not changed in this PR, but I realized it looking at the code)

And actually after putting `check_threshold`, may be we need no more `check_threshold_not_negative` as `check_threshold` already checks this. (However it checks if `threshold < -1e-5`, where as `check_threshold` checks if `threhsold < 0`) I wonder if it would be good to replace `0` with `-1e-5` in `check_threshold` as well?

Finally, I do not know if it makes sense, but `plot_prob_atlas` first calls `plot_anat` without passing threshold value, and uses default threshold of `plot_anat` which is `threshold=None`. And then it sets its own threshold value later as follows:

```python
   if threshold is None:
        threshold = 1e-6
    elif threshold == "auto":
        # it will use default percentage,
        # strategy is to avoid maximum overlaps as possible
        if view_type == "contours":
            correction_factor = 1
        elif view_type == "filled_contours":
            correction_factor = 0.8
        else:
            correction_factor = 0.5
        threshold = f"{100 * (1 - 0.2 * correction_factor / n_maps):f}%"

    if isinstance(threshold, collections.abc.Iterable) and not isinstance(
        threshold, str
    ):
        threshold = list(threshold)
        if len(threshold) != n_maps:
            raise TypeError(
                "The list of values to threshold "
                "should be equal to number of maps"
            )
    else:
        threshold = [threshold] * n_maps
```
Would that difference in threhold values cause a problem?

Shall we alternatively call `check_threshold` in main plotting functions rather than backend function?

emdupre: > This change is made in _plot_img_with_bg which is called directly or indirectly from almost all image plotting functions.

Indeed, the ideal is to have the same behavior across all plotting functions ; the original impetus was that `view_img` can take a string as a threshold (as could `plot_prob_atlas`) but not other plotting functions.

> One point is docstrings are feeded from nilearn._utils.docs which does not indicate percentage str as threshold value. That would be good to indicate that.

Agreed, great catch !

> The second point is as check_threshold_not_negative(threshold) is already done in _plot_img_with_bg, it could be removed from other functions that call _plot_img_with_bg. (This is not changed in this PR, but I realized it looking at the code)

> And actually after putting check_threshold, may be we need no more check_threshold_not_negative as check_threshold already checks this. (However it checks if threshold < -1e-5, where as check_threshold checks if threhsold < 0) I wonder if it would be good to replace 0 with -1e-5 in check_threshold as well?

I think this is a slightly different check ; `check_threshold` allows for threshold to be a negative float [in certain cases](https://github.com/nilearn/nilearn/blob/7872a761b3b01a6c9631bfb48101fed6d3c432f0/nilearn/_utils/param_validation.py#L100-L118) while `check_threshold_non_negative` fails for values [less than -1e-5 ](https://github.com/nilearn/nilearn/blob/7872a761b3b01a6c9631bfb48101fed6d3c432f0/nilearn/plotting/_utils.py#L270-L277). So, we would include the more general case in the private function `_plot_img_with_bg` and the more specific case in individual plotting calls, where we may indeed not support negative thresholding.

> Finally, I do not know if it makes sense, but plot_prob_atlas first calls plot_anat without passing threshold value, and uses default threshold of plot_anat which is threshold=None. And then it sets its own threshold value later as follows:
> Would that difference in threhold values cause a problem?

I don't think the difference would cause a problem, as that data type is tricky to plot well, so I think the modified threshold was chosen deliberately. But, I would put any changes to that function as a new issue / PR cycle !

hndgzkn: Thanks for the answers!

> I think this is a slightly different check ; `check_threshold` allows for threshold to be a negative float [in certain cases](https://github.com/nilearn/nilearn/blob/7872a761b3b01a6c9631bfb48101fed6d3c432f0/nilearn/_utils/param_validation.py#L100-L118) 

Yes, actually keeping negative check might be good. Otherwise it has to be commented that `check_threshold` should never be called with `two_sided=False` from plotting functions.

hndgzkn: However, if `threshold=-1e-5` is given, I guess `check_negative_threshold` will let it pass, but `check_threshold` will complain.

saeed-neuro:
Thanks @emdupre, @Remi-Gau and @hndgzkn for your work and points!

nilearn/plotting/image/img_plotting.py
emdupre: Sorry, this is likely in my logic checks, but I don't see that case ? If we pass -1e-5, I thought that would only fail [if the data has no negative values lower than -1e-5](https://github.com/nilearn/nilearn/blob/7872a761b3b01a6c9631bfb48101fed6d3c432f0/nilearn/_utils/param_validation.py#L109-L118).

hndgzkn: It is a few lines above:

https://github.com/nilearn/nilearn/blob/7872a761b3b01a6c9631bfb48101fed6d3c432f0/nilearn/_utils/param_validation.py#L100-L104

It will raise error if `two_sided=True` and a negative threshold is specified. 

We had decided not to use `two_sided` parameter for plotting. Because what `two_sided` parameter provides us is achieved with `vmin`  and `vmax` for plotting.

hndgzkn: It is also in the docstring of `check_threshold`.

https://github.com/nilearn/nilearn/blob/7872a761b3b01a6c9631bfb48101fed6d3c432f0/nilearn/_utils/param_validation.py#L24-L30.

emdupre: Ah I see what you mean ; yes, it would trigger then. But is there a case touched here where two_sided is set as True ?

hndgzkn: two_sided is by default True for `check_threshold` and it should actually be True when called from plotting. 

If `threshold="auto"` is set, the threshold might have value -1e-5 as mentioned below:

https://github.com/nilearn/nilearn/blob/563c57572e94630bbb94100bb20597e0f8d533d7/nilearn/plotting/_utils.py#L270-L277

So the code would raise error, when it shouldn't.

And the lines that adjust threshold when set to "auto" is just a few lines above the line that call `check_threshold`.

hndgzkn: Replacing `if threshold is not None:` with `if isinstance(threshold, str)` would fix it, and avoid unnecessary check.

hndgzkn: Other option could be replacing the line 232

`threshold = float(fast_abs_percentile(data)) - 1e-5`

with 

`threshold="80%"` 

and then let `check_threshold` set the threshold.

In either case, it is not clear to me why we subtract `-1e-5` here, but not in `check_threshold`.

Isn't the reasoning below valid for all thresholded data?

https://github.com/nilearn/nilearn/blob/563c57572e94630bbb94100bb20597e0f8d533d7/nilearn/plotting/image/img_plotting.py#L229-L232

emdupre:
This closure was accidental ! So sorry, will work to fix.

PR #5986, opened 2026-02-02
[FIX] Check if images specified to nifti maskers transform are empty

### LINKED ISSUE #5460: Only check if image is empty when necessary

hndgzkn:
- Closes #5460

### DIFF
--- a/nilearn/_utils/estimator_checks.py
+++ b/nilearn/_utils/estimator_checks.py
@@ -2102,6 +2102,8 @@ def check_supervised_img_estimator_y_no_nan(estimator_orig) -> None:
 def check_decoder_empty_data_messages(estimator_orig):
     """Check that empty images are caught properly.
 
+    Run on both surfance and volume data.
+
     Replaces sklearn check_estimators_empty_data_messages.
     """
     estimator = clone(estimator_orig)
@@ -2123,18 +2125,22 @@ def check_decoder_empty_data_messages(estimator_orig):
         random_state=42,
     )
 
-    if accept_niimg_input(estimator):
-        data = np.zeros((64, 64, 0))
-        X = Nifti1Image(data, np.eye(4))
-    else:
-        imgs = _make_surface_img(n_samples)
-        data = {
-            part: np.empty(0).reshape((imgs.data.parts[part].shape[0], 0))
-            for part in imgs.data.parts
-        }
-        X = SurfaceImage(imgs.mesh, data)
+    data = np.zeros((64, 64, 0))
+    X = Nifti1Image(data, np.eye(4))
 
     y = _rng().random(y.shape)
+
+    with pytest.raises(ValueError, match="empty"):
+        estimator.fit(X, y)
+
+    estimator = clone(estimator_orig)
+
+    imgs = _make_surface_img(n_samples)
+    data = {
+        part: np.empty(0).reshape((imgs.data.parts[part].shape[0], 0))
+        for part in imgs.data.parts
+    }
+    X = SurfaceImage(imgs.mesh, data)
 
     with pytest.raises(ValueError, match="empty"):
         estimator.fit(X, y)


hndgzkn:
@Remi-Gau I couldn't be sure how to fix [this failure](https://github.com/nilearn/nilearn/actions/runs/21636117147/job/62361991176?pr=5986#step:8:12987). This is a sklearn check on sparse data as far as I understand. I cannot judge if it is relevant, or how my changes had an effect on this. Shall we put a tag for estimator? Do you have an idea?

```
  FAILED nilearn/decomposition/tests/test_decomposition_estimators.py::test_check_estimator_sklearn[CanICA(standardize='zscore_sample')-check_estimator_sparse_array] - AssertionError: Estimator CanICA doesn't seem to fail gracefully on sparse data: error message should state explicitly that sparse input is not supported if this is not the case, e.g. by using check_array(X, accept_sparse=False).
  FAILED nilearn/decomposition/tests/test_decomposition_estimators.py::test_check_estimator_sklearn[DictLearning(standardize='zscore_sample')-check_estimator_sparse_array] - AssertionError: Estimator DictLearning doesn't seem to fail gracefully on sparse data: error message should state explicitly that sparse input is not supported if this is not the case, e.g. by using check_array(X, accept_sparse=False).

```

hndgzkn:
> @Remi-Gau I couldn't be sure how to fix [this failure](https://github.com/nilearn/nilearn/actions/runs/21636117147/job/62361991176?pr=5986#step:8:12987). This is a sklearn check on sparse data as far as I understand. I cannot judge if it is relevant, or how my changes had an effect on this. Shall we put a tag for estimator? Do you have an idea?

I could locate a bit the when the tests fail. It starts to fail when `_check_imgs` is run is `MultiNiftiMasker` `fit` function. With other maskers, there is no problem.

Remi-Gau:
> @Remi-Gau I couldn't be sure how to fix [this failure](https://github.com/nilearn/nilearn/actions/runs/21636117147/job/62361991176?pr=5986#step:8:12987). This is a sklearn check on sparse data as far as I understand. I cannot judge if it is relevant, or how my changes had an effect on this. Shall we put a tag for estimator? Do you have an idea?

usually I would tweak what sklearn check is expected to fail via the return_expected_failed_checks function (look around line 390): it can turn into a bit of a headache at times because some fails will only happen for some very specific combination of dependencies...

Remi-Gau:
> > @Remi-Gau I couldn't be sure how to fix [this failure](https://github.com/nilearn/nilearn/actions/runs/21636117147/job/62361991176?pr=5986#step:8:12987). This is a sklearn check on sparse data as far as I understand. I cannot judge if it is relevant, or how my changes had an effect on this. Shall we put a tag for estimator? Do you have an idea?
> 
> usually I would tweak what sklearn check is expected to fail via the return_expected_failed_checks function (look around line 390): it can turn into a bit of a headache at times because some fails will only happen for some very specific combination of dependencies...

I know this is not the most satisfying approach but I am mostly trying to be pragmatic there: if we solve more problems than we create in a PR, we are moving forward.

nilearn/maskers/base_masker.py
```diff
@@ -308,8 +312,56 @@ class BaseMasker(
...
+        # in case where the masker was previously fitted
+        self._report_content["warning_messages"] = []
+
+        self.clean_args_ = {} if self.clean_args is None else self.clean_args
+
+        self._fit_cache()
+
+        self.mask_img_ = self._load_mask(imgs)
+
+        return self._fit(imgs)
```
Remi-Gau, line 363: should we add a _fit() method with a not implemented error in the base maskers to make it more clear from reading the code it must be implemented in each child class?

hndgzkn: Added.

nilearn/_utils/estimator_checks.py
```diff
@@ -395,15 +392,12 @@ def return_expected_failed_checks(
...
-            expected_failed_checks.pop("check_estimator_sparse_array")
-
-    if isinstance(estimator, (MultiNiftiMasker)) and SKLEARN_GTE_1_6:
-        expected_failed_checks.pop("check_estimator_sparse_tag")
+        # if (
+        #     not IS_SKLEARN_1_6_1_on_py_lt_3_13
+        #     and SKLEARN_GTE_1_5
+        #     and scipy_version != "1.9.0"
+        # ):
+        # expected_failed_checks.pop("check_estimator_sparse_array")
```
Remi-Gau: without this the checks are xfailed locally (as expected)

Remi-Gau: @hndgzkn

Remi-Gau: the problem is that there may be a combination of dependencies / python version where this is needed

will also run the minimum dependencies workflow on this PR to make sure we don't have some bad surprise after merging

.github/workflows/testing_minimum.yml
```diff
@@ -30,6 +30,7 @@ on:
     # This enables manual triggering from GitHub UI
     # https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch
     workflow_dispatch:
+    pull_request:
```
```suggestion
```
reverse before merging

nilearn/_utils/estimator_checks.py
hndgzkn: Thanks @Remi-Gau ! I thought I had tried that, but somehow it did not work. Probably I missed something.

hndgzkn:
@Remi-Gau I've made some more changes to add empty data check for decoders and glm as well, and added some docstring for new functions in `nilearn.image`. 

I would like to push them, but waiting CI jobs to complete.

I have a question. With this PR, we have three more functions in `nilearn.image` which are not private but none of them are listed in `__init__.all`. I think `is_volume_image` and `is_empty_volume` might be useful for users. Shall we expose them in the package?

I could not be sure about `check_volume_for_fit`. At least, if it becomes public, it would be better to rename it. What do you think?

Remi-Gau:
> I have a question. With this PR, we have three more functions in `nilearn.image` which are not private but none of them are listed in `__init__.all`. I think `is_volume_image` and `is_empty_volume` might be useful for users. Shall we expose them in the package?
> 
> I could not be sure about `check_volume_for_fit`. At least, if it becomes public, it would be better to rename it. What do you think?

Well in any case they would not be made public before version 0.14.0, because they would be considered new features.

I'd be tempted to wait a while before making those part of the user facing API: either wait for users to request it or if we see that the function is used in a lot of repos despite being private.

Remi-Gau:
those failing tests seem unrelated to me:

more of a flaky thing

https://github.com/nilearn/nilearn/actions/runs/21704064722/job/62590727253?pr=5986#step:8:20520

nilearn/image/image.py
```diff
@@ -57,6 +57,122 @@
 from nilearn.typing import ClusterThreshold, NiimgLike
 
 
+def is_volume_image(imgs):
```
```suggestion
def is_volume_image(imgs) -> bool:
```

nilearn/image/image.py
```diff
@@ -57,6 +57,122 @@
...
+        return False
+
+    if hasattr(imgs, "__iter__") and not isinstance(imgs, str):
+        for x in imgs:
+            if not is_volume_image(x):
+                return False
+    return True
+
+
+def is_empty_volume(img):
```
```suggestion
def is_empty_volume(img) -> bool:
```

nilearn/image/image.py
```diff
@@ -57,6 +57,122 @@
...
+            )
+            raise ValueError(message)
+        else:
+            raise ValueError(f"File not found: '{imgs}'")
+    elif not Path(imgs).exists():
+        raise ValueError(f"File not found: '{imgs}'")
+    return imgs
+
+
+def check_volume_for_fit(imgs):
```
```suggestion
def check_volume_for_fit(imgs) -> None:
```

hndgzkn:
@Remi-Gau Lets not put this in the release. There are still problems to solve on this.

There is [this complaint from mypy](https://results.pre-commit.ci/run/github/1235740/1770302507.bmkx3q1tTm2uyJbpvhxdiQ). Is there a way to overcome this?

Remi-Gau:
> @Remi-Gau Lets not put this in the release. There are still problems to solve on this.
> 
> There is [this complaint from mypy](https://results.pre-commit.ci/run/github/1235740/1770302507.bmkx3q1tTm2uyJbpvhxdiQ). Is there a way to overcome this?

fixed in https://github.com/nilearn/nilearn/pull/5986/commits/bc343ef4877a922ee1f992fcd2456b56f04486a8

required declaring the type of a variable and some rename to avoid some 'type clash'

hndgzkn:
> fixed in [bc343ef](https://github.com/nilearn/nilearn/commit/bc343ef4877a922ee1f992fcd2456b56f04486a8)
> 
> required declaring the type of a variable and some rename to avoid some 'type clash'

Thanks!

Remi-Gau:
if CI comes out green I will merge before the next release

Remi-Gau:
the errros are unrelated

Issue #5933, opened 2026-01-08
Making Nilearn threadsafe


Remi-Gau:
This is 'meta-issue' to centralize everything related to make Nilearn thread-safe

We should probably set some expectations on what part of nilearn can be made thread-sfe:

| subpackage       | scope | failing tests marked as unsafe | tests made safe | code made safe |
| ---------------- | ----- | ------------------------------ | --------------- | -------------- |
| connectome/      | [x]   | [x]                            | [ ]             | [ ]            |
| datasets/        | [ ]   | [ ]                            | [ ]             | [ ]            |
| decoding/        | [x]   | [x]                            | [ ]             | [ ]            |
| decomposition/   | [x]   | [x]                            | [ ]             | [ ]            |
| glm/             | [x]   | [x]                            | [ ]             | [ ]            |
| image/           | [x]   | [x]                            | [ ]             | [ ]            |
| interfaces/      | [x]   | [x]                            | [ ]             | [ ]            |
| maskers/         | [x]   | [x]                            | [ ]             | [ ]            |
| mass_univariate/ | [x]   | [x]                            | [ ]             | [ ]            |
| plotting/        | [x]   | [ ]                            | [ ]             | [ ]            |
| regions/         | [x]   | [x]                            | [ ]             | [ ]            |
| reporting/       | [x]   | [x]                            | [ ]             | [ ]            |
| surface/         | [x]   | [x]                            | [ ]             | [ ]            |
| tests/           | [x]   | [x]                            | [ ]             | [ ]            |
| _utils/          | [x]   | [x]                            | [ ]             | [ ]            |
| utils/           | [x]   | [x]                            | [ ]             | [ ]            |

---

- [x] properly test free-threading
- [ ] mark tests that are failing when multi-threaded as thread-unsafe
  - #5859
  - #5943 
- [ ] make tests threadsafe
  - see https://py-free-threading.github.io/testing/#pytest-is-not-thread-safe
- [ ] make code threadsafe

--- 

in general I would recommend trying to first tackle the lowest part of the architecture (for example nilearn.signal) before tackling higher level part of the API

PR #5987, opened 2026-02-03
[MAINT] run parallel thread tests on plotting

Remi-Gau:
- Relates to #5933 

- mark plotting tests as thread unsafe

### DIFF
--- a/nilearn/plotting/image/tests/test_plot_roi.py
+++ b/nilearn/plotting/image/tests/test_plot_roi.py
@@ -81,6 +81,7 @@ def test_plot_roi_view_type_error(matplotlib_pyplot):
         demo_plot_roi(view_type="flled")
 
 
+@pytest.mark.thread_unsafe
 @pytest.mark.slow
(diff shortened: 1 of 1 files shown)


Remi-Gau:
extracting things unrelated to the plotting module in a separate PR to pacify failing tests in other PRs

bthirion (approved):
Out of curiosity: what is the rule to mark tests as thread safe/unsafe ? Is this empirical ?

Remi-Gau:
> Out of curiosity: what is the rule to mark tests as thread safe/unsafe ? Is this empirical ?

at the moment it's very simple: if a test shows a "PARALLEL FAILED" then I mark it as thread unsafe (for now)

Remi-Gau:
OK this one is green for now.
Will merge but I still expect to see 'once in a while' some tests failing in nilearn.plotting when run multithreaded on other PRs after merging: we will have to mark those with `@pytest.mark.thread_unsafe` when that happens, but we will discover pore qiickly what tests are failing by running those tests over several PRs than just this one.

@hndgzkn pinging you so that you are aware of this.

PR #5990, opened 2026-02-06
[ENH] Make times and durations readable

rohan3412:
The displayed "Time Elapsed" isn't very readable

Current:
image
Proposed:
image

and also

image 
will become
image

- format displayed time

HR, MIN and SEC can be other formats as well (for example: Hour, Minute and Second)

### DIFF
--- a/nilearn/_utils/helpers.py
+++ b/nilearn/_utils/helpers.py
@@ -220,40 +220,6 @@ def remove_parameters(removed_params, reason, end_version="future"):
     return _remove_params
 
 
-def readable_time(seconds):
-    """Convert a duration in seconds to a human-readable string.
-
-    The output includes hours, minutes, and seconds as needed, using
-    abbreviated units (HR/HRS, MIN, SEC). Zero-value units are omitted.
-
-    Parameters
-    ----------
-    seconds : int or float
-        Time duration in seconds.
-
-    Returns
-    -------
-    str
-        Human-readable time string.
-    """
-    seconds = round(seconds)
-    if seconds <= 0:
-        return "0 SEC"
-
-    hour, remainder = divmod(seconds, 3600)
-    minute, seconds = divmod(remainder, 60)
-
-    parts = []
-    if hour:
-        parts.append(f"{hour} HR" if hour == 1 else f"{hour} HRS")
-    if minute:
-        parts.append(f"{minute} MIN")
-    if seconds:
-        parts.append(f"{seconds} SEC")
-
-    return " ".join(parts)
-
-
 def stringify_path(path):
     """Convert path-like objects to string.
 
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -18,8 +18,7 @@ from sklearn.utils.estimator_checks import check_is_fitted
 from nilearn._utils import logger
 from nilearn._utils.docs import fill_doc
 from nilearn._utils.glm import check_and_load_tables
-from nilearn._utils.helpers import readable_time
-from nilearn._utils.logger import find_stack_level
+from nilearn._utils.logger import find_stack_level, readable_time
 from nilearn._utils.masker_validation import (
     check_compatibility_mask_and_images,
 )
--- a/nilearn/_utils/logger.py
+++ b/nilearn/_utils/logger.py
@@ -1,6 +1,7 @@
 """Logging facility for nilearn."""
 
 import inspect
+import time
 import traceback
 from pathlib import Path
 
@@ -198,3 +199,24 @@ def one_level_deeper() -> int:
     Needs to be in a module that does not start with 'test'
     """
     return find_stack_level()
+
+
+def readable_time(seconds) -> str:
+    """Convert a duration in seconds to a human-readable string.
+
+    The output includes hours, minutes, and seconds as needed, using
+    abbreviated units (HR/HRS, MIN, SEC).
+
+    Parameters
+    ----------
+    seconds : int or float
+        Time duration in seconds.
+
+    Returns
+    -------
+    str
+        Human-readable time string.
+    """
+    seconds = round(seconds)
+
+    return time.strftime("%H HR %M MIN %S SEC", time.gmtime(seconds))
--- a/nilearn/_utils/tests/test_logger.py
+++ b/nilearn/_utils/tests/test_logger.py
@@ -12,6 +12,7 @@ from nilearn._utils.logger import (
     find_stack_level,
     log,
     one_level_deeper,
+    readable_time,
 )
 
 
@@ -124,3 +125,7 @@ with capture_output() as out:
             assert out[0].getvalue() == "[<module>] message from no function\n"
         else:
             assert out[0] == "[<module>] message from no function\n"
+
+
+def test_readable_time():
+    assert readable_time(56) == "00 HR 00 MIN 56 SEC"
--- a/nilearn/datasets/_utils.py
+++ b/nilearn/datasets/_utils.py
@@ -21,8 +21,7 @@ import requests
 
 from nilearn._utils import logger
 from nilearn._utils.docs import fill_doc
-from nilearn._utils.helpers import readable_time
-from nilearn._utils.logger import find_stack_level
+from nilearn._utils.logger import find_stack_level, readable_time
 from nilearn._utils.param_validation import (
     check_parameter_in_allowed,
     check_params,
--- a/nilearn/decoding/searchlight.py
+++ b/nilearn/decoding/searchlight.py
@@ -20,7 +20,7 @@ from sklearn.utils.estimator_checks import check_is_fitted
 from nilearn._base import NilearnBaseEstimator
 from nilearn._utils import logger
 from nilearn._utils.docs import fill_doc
-from nilearn._utils.helpers import readable_time
+from nilearn._utils.logger import readable_time
 from nilearn._utils.param_validation import check_params
 from nilearn._utils.versions import SKLEARN_LT_1_6
 from nilearn.image import check_niimg_3d, check_niimg_4d, new_img_like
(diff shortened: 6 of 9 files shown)


nilearn/glm/second_level/second_level.py
Remi-Gau, line 1: missing an import:

from nilearn._utils.helpers import readable_time

rohan3412:
pre-commit.ci autofix

Remi-Gau:
the test failure is unrelated: just a test timing out (will mark the test as slow for you)

I doubt the doc failure is related either: just not sure why we are now getting all these new warnings.

Remi-Gau:
@emdupre you gave a 👎🏻 on the top message: do you have a specific thing in mind for this PR?

Remi-Gau:
> I doubt the doc failure is related either: just not sure why we are now getting all these new warnings.

I think this is because of the error on the nipy webdomain

emdupre:
Thanks for following up @Remi-Gau ; sorry, I had written a message but clearly didn't send !

Thanks for the contribution, @rohan3412 ! I'm a little hesitant about incorporating this for a few reasons :
- The proposed functionality in `readable_time` is usually handled in the standard Python library `datetime` with [`timedelta`](https://docs.python.org/3/library/datetime.html#datetime.timedelta) ; is there a reason to add a new function rather than using a standard library ?
- I don't know that it makes sense to have all computations that you've touched with the same format ; indeed, `SpaceNetClassifier` is very slow, but a second level GLM is not (there I would not expect hours to be a useful unit). 

More generally, I'd prefer to have these kinds of conversations in issues, since I know both you and @Remi-Gau have now put time into this work ! Was there a discussion that I missed (very likely !) ? Apologies if so, but please feel free to point me there so I can catch up on the reasoning here.

Remi-Gau:
> Was there a discussion that I missed (very likely !) ?

pretty sure that this was a 'drive-by PR'

rohan3412:
> Thanks for following up @Remi-Gau ; sorry, I had written a message but clearly didn't send !
> 
> Thanks for the contribution, @rohan3412 ! I'm a little hesitant about incorporating this for a few reasons :
> 
> * The proposed functionality in `readable_time` is usually handled in the standard Python library `datetime` with [`timedelta`](https://docs.python.org/3/library/datetime.html#datetime.timedelta) ; is there a reason to add a new function rather than using a standard library ?
> * I don't know that it makes sense to have all computations that you've touched with the same format ; indeed, `SpaceNetClassifier` is very slow, but a second level GLM is not (there I would not expect hours to be a useful unit).
> 
> More generally, I'd prefer to have these kinds of conversations in issues, since I know both you and @Remi-Gau have now put time into this work ! Was there a discussion that I missed (very likely !) ? Apologies if so, but please feel free to point me there so I can catch up on the reasoning here.

@emdupre 
The timedelta function isn't formatted in a readable way,
the reason to use this, is that it is dynamic and can handle different orders of magnitude of time (as seen in the images)
[test_time.py](https://github.com/user-attachments/files/25204883/test_time.py)
image


output:
image

rohan3412:
also, the format can be changed too. 

for example, it could be
image

emdupre:
Thanks for confirming, @rohan3412 !

Yes, I would usually think to format `timedelta` ; I agree the default format isn't readable, but I'm not sure that we need to have a new function rather than format a standard library one. 

Again, too, I wonder if it makes sense to be using this outside of `SpaceNetClassifier`. Do you have real cases where e.g. a `SecondLevelGLM` is going to need to be formatted outside of seconds ?

emdupre:
> > Was there a discussion that I missed (very likely !) ?
> 
> pretty sure that this was a 'drive-by PR'

@rohan3412 just to explain here : our [contributing guide](https://nilearn.github.io/stable/development.html#contribute-code) asks for code contributions to come with a new issue first ; you can see the reasoning outlined a bit more [here](https://snarky.ca/setting-expectations-for-open-source-participation/#submittingacontribution), for example.

Broadly, though, having discussions about potential new features / enhancements before we start work on them (both on the coding and the reviewing side !) helps to make sure that everyone is aligned before there's real investment. 

I know you have already contributed to Nilearn before -- and thank you for continuing to contribute ! it's really appreciated -- but we try to follow this routine every time, even within the team. It helps to structure the process in a way that everyone is aligned ! Hopefully that makes sense.

rohan3412:
> > > Was there a discussion that I missed (very likely !) ?
> > 
> > 
> > pretty sure that this was a 'drive-by PR'
> 
> @rohan3412 just to explain here : our [contributing guide](https://nilearn.github.io/stable/development.html#contribute-code) asks for code contributions to come with a new issue first ; you can see the reasoning outlined a bit more [here](https://snarky.ca/setting-expectations-for-open-source-participation/#submittingacontribution), for example.
> 
> Broadly, though, having discussions about potential new features / enhancements before we start work on them (both on the coding and the reviewing side !) helps to make sure that everyone is aligned before there's real investment.
> 
> I know you have already contributed to Nilearn before -- and thank you for continuing to contribute ! it's really appreciated -- but we try to follow this routine every time, even within the team. It helps to structure the process in a way that everyone is aligned ! Hopefully that makes sense.

@emdupre Ah, my bad. I didn't know about that

rohan3412:
> Thanks for confirming, @rohan3412 !
> 
> Yes, I would usually think to format `timedelta` ; I agree the default format isn't readable, but I'm not sure that we need to have a new function rather than format a standard library one.
> 
> Again, too, I wonder if it makes sense to be using this outside of `SpaceNetClassifier`. Do you have real cases where e.g. a `SecondLevelGLM` is going to need to be formatted outside of seconds ?

You mention formatting timedelta, but timedelta has no built-in format function. Any formatting would just be writing the readable_time() function's logic from scratch. 

`td = timedelta(seconds=total_seconds)`
`minutes, seconds = divmod(td.seconds, 60)`
the function is simple and this is pretty much what it does

For a SecondLevelGLM, if there are thousands of subjects, it can be longer than seconds 
or
nilearn.glm.second_level.non_parametric_inference with a large number of permutations, it would be longer than seconds

the only thing thing to do, would be to add additional time measurement code for this

what i did so far, was just to apply the readable_time function to the already existing sections in the code, where time is measured and displayed

Remi-Gau:
@emdupre and I talked about this PR the other day.
My take was more: I am happy with anything that improves output readibility.
Curious to hear how the rest of the team feels about: @nilearn/core

bthirion:
No strong feeling about it. If people like it, let's take it. 
Best, B

hndgzkn:
I find it nice to have a single function that standardizes time formatting and increases output readability.

pbarbarant:
Sounds good IMHO, note that the built-in `time.gmtime` might help you avoid doing the conversion by hand:

```python

>>> import time
>>> seconds = 56
>>> print(time.strftime("%H HR %M MIN %S SEC", time.gmtime(seconds)))
00 HR 00 MIN 56 SEC
```

rohan3412:
> Sounds good IMHO, note that the built-in `time.gmtime` might help you avoid doing the conversion by hand:
> 
> ```python
> 
> >>> import time
> >>> seconds = 56
> >>> print(time.strftime("%H HR %M MIN %S SEC", time.gmtime(seconds)))
> 00 HR 00 MIN 56 SEC
> ```

One of the goals of the readable_time function is to dynamically drop zero-value time units to keep the output as concise as possible. As your example shows, time.strftime uses a static format string, which forces it to output 00 HR 00 MIN 56 SEC for a 56-second duration.

Remi-Gau:
> > Sounds good IMHO, note that the built-in `time.gmtime` might help you avoid doing the conversion by hand:
> > ```python
> > >>> import time
> > >>> seconds = 56
> > >>> print(time.strftime("%H HR %M MIN %S SEC", time.gmtime(seconds)))
> > 00 HR 00 MIN 56 SEC
> > ```
> 
> One of the goals of the readable_time function is to dynamically drop zero-value time units to keep the output as concise as possible. As your example shows, time.strftime uses a static format string, which forces it to output 00 HR 00 MIN 56 SEC for a 56-second duration.

We quickly discussed this in our core dev meeting yesterday.

We are still going to go with something that prints something like `00 HR 00 MIN 56 SEC` for the concern that some people may prefer to have consistent length of their log outputs.

Given that this is already a clear improvement of the readability of the output we determine it was in line with the original spirit of the PR.

Remi-Gau:
@rohan3412 
I will make the final change and add a small test (for easier debugging).

nilearn/_utils/helpers.py
```diff
@@ -220,6 +220,40 @@ def wrapper(*args, **kwargs):
     return _remove_params
 
 
+def readable_time(seconds):
```
Remi-Gau: will move to nilearn._utils.logging

rohan3412:
> > > Sounds good IMHO, note that the built-in `time.gmtime` might help you avoid doing the conversion by hand:
> > > ```python
> > > >>> import time
> > > >>> seconds = 56
> > > >>> print(time.strftime("%H HR %M MIN %S SEC", time.gmtime(seconds)))
> > > 00 HR 00 MIN 56 SEC
> > > ```
> > 
> > 
> > One of the goals of the readable_time function is to dynamically drop zero-value time units to keep the output as concise as possible. As your example shows, time.strftime uses a static format string, which forces it to output 00 HR 00 MIN 56 SEC for a 56-second duration.
> 
> We quickly discussed this in our core dev meeting yesterday.
> 
> We are still going to go with something that prints something like `00 HR 00 MIN 56 SEC` for the concern that some people may prefer to have consistent length of their log outputs.
> 
> Given that this is already a clear improvement of the readability of the output we determine it was in line with the original spirit of the PR.

I see

rohan3412:
> @rohan3412 I will make the final change and add a small test (for easier debugging).

Alright

Remi-Gau:
merging
thanks @rohan3412 for this one and very sorry it took so long with so much back and forth.

PR #5994, opened 2026-02-09
[MAINT] fix some ruff errors

Remi-Gau:
- Closes none

- fixes https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/
- fixes https://docs.astral.sh/ruff/rules/collection-literal-concatenation/

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -16,10 +16,9 @@ import re
 import sys
 from pathlib import Path
 
+from nilearn._version import __version__
 from sphinx.domains import changeset
 from sphinx.locale import _
-
-from nilearn._version import __version__
 
 # ----------------------------------------------------------------------------
 


hndgzkn (approved):
LGTM.  However I agree with Bertrand. 

I thought we could still ignore ruff errors for `from` clause as apparently it does not really change behavior of exceptions, and the code seems simpler without them.

For unpacking, I find `+` more clear, if unpacking does not have a performance advantage.

Remi-Gau:
> Good with me, though I'm not particularly enthusiastic about the changes. I did not know about this `from` pattern.

```python
try:
    a[0] = 1
except Exception as e:
    raise RuntimeError("foo")
```

gives

```
  File "/home/remi-gau/github/nilearn/nilearn/tmp/error.py", line 2, in <module>
    a[0] = 1
    ^
NameError: name 'a' is not defined

During handling of the above exception, another exception occurred:

  File "/home/remi-gau/github/nilearn/nilearn/tmp/error.py", line 4, in <module>
    raise RuntimeError("foo")
RuntimeError: foo
```

where as


```python
try:
    a[0] = 1
except Exception as e:
    raise RuntimeError("foo") from e
```

gives

```
  File "/home/remi-gau/github/nilearn/nilearn/tmp/error.py", line 2, in <module>
    a[0] = 1
    ^
NameError: name 'a' is not defined

The above exception was the direct cause of the following exception:

  File "/home/remi-gau/github/nilearn/nilearn/tmp/error.py", line 4, in <module>
    raise RuntimeError("foo") from e
RuntimeError: foo
```

So, instead of we get:

- During handling of the above exception, another exception occurred:

We get:

- The above exception was the direct cause of the following exception:

bthirion:
Thx for the explanations !

PR #6002, opened 2026-02-11
[REF] refactor get_params

Remi-Gau:
- Closes none

- the private function get_params is destined to help 'transfer' parameter value from one object instance to another but was mostly used to get the parameters of the maskers 
- this PR adds a _get_masker_params method to the BaseMasker that does only this
- extract a _clean method for all surface maskers


nilearn/maskers/masker_validation.py
```diff
@@ -19,6 +18,48 @@
 from nilearn.maskers._mixin import _MultiMixin
 
 
+def get_params(cls, instance, ignore: None | list[str] = None):
```
Remi-Gau: simply moved from nilearn._utils.class_inspect

nilearn/maskers/tests/test_masker_validation.py
```diff
@@ -4,9 +4,35 @@
...
 from sklearn.base import BaseEstimator
 
+from nilearn._base import NilearnBaseEstimator
 from nilearn._utils.versions import SKLEARN_LT_1_6
 from nilearn.maskers import MultiNiftiMasker, NiftiMasker, SurfaceMasker
-from nilearn.maskers.masker_validation import check_embedded_masker
+from nilearn.maskers.masker_validation import check_embedded_masker, get_params
+
+
+class A(NilearnBaseEstimator):
```
Remi-Gau, line 13: moved from nilearn/_utils/tests/test_class_inspect.py

nilearn/maskers/base_masker.py
```diff
@@ -324,6 +324,41 @@ def _check_imgs(self, imgs) -> None:
         """
         raise NotImplementedError()
 
+    def _get_masker_params(self, ignore: None | list[str] = None, deep=False):
```
bthirion: why is it useful ?

bthirion (commented):
LGTM overall.

nilearn/maskers/base_masker.py
Remi-Gau: this is mostly used to collect params in all the nifti maskers to collect the params to pass to resample, smooth, clean functions

it allows to also skip systematically returning some parameters that will not be used by those downstream functions (or would invalid the cache)

nilearn/maskers/base_masker.py
```diff
@@ -889,3 +924,22 @@ def _generate_figure(
     def _set_contour_colors(self, hemi):
         """Set the colors for the contours in the report."""
         del hemi
+
+    def _clean(self, region_signals: np.ndarray, confounds, sample_mask):
```
Remi-Gau: this was the same across all surface maskers so extracted into a method here

nilearn/maskers/base_masker.py
Remi-Gau: actually given this is only used by nifti masker I am going to move it to the base nifti masker

bthirion:
Is this finalized ? LGTM.

nilearn/maskers/base_masker.py
```diff
@@ -418,6 +418,39 @@ def __sklearn_tags__(self):
         tags.estimator_type = "masker"
         return tags
 
+    def _get_masker_params(self, ignore: None | list[str] = None, deep=False):
+        """Get parameters for this masker.
+
+        Very similar to the BaseEstimator.get_params() from sklearn
```
hndgzkn, line 413: If this will be used also by Surface Maskers, it might be better to put it in `_BaseMasker`.

Actually, I was planning to move `fit` to `_BaseMasker` in the PR that I worked for checking empty images. Because surface masker fit also has these common parts in fit, but needs to be tidied up a little bit. I abondoned it  for  `0.13.1` release to be on time. 
@Remi-Gau if that is ok with you, after this PR, I can open a PR to do that.

nilearn/maskers/masker_validation.py
hndgzkn: Doesn't this basically do the same thing with `_get_masker_params`? May be I am missing somethig, but do we really need this?

hndgzkn (approved):
A part from some comments, LGTM.

nilearn/maskers/base_masker.py
hndgzkn: I've just seen your answer to Bertrand. So it is only for Nifti maskers, ignore my comment on that please.

nilearn/maskers/masker_validation.py
Remi-Gau: Not quite.

_get_masker_params will get you the parameters and their values for a nifti masker instance

get_params takes 2 objects as input: get the parameter name from and the parameter values from the other (more or less), it's used for embedded maskers when we want to transfer the parameter values from the owning class to the embedded masker

PR #6004, opened 2026-02-12
[ENH] Improve error messages when Chrome is not installed with newer kaleido versions

### LINKED ISSUE #5471: Kaleido 1.0.0 requires Chrome to be installed

hndgzkn:
- Closes #5471

- Improves error handling due to:
  - `chrome` not being installed on the system depending on `kaleido` version (version starting from 1.0.0 do not bundle chrome)
  - incompatible `plotly` - `kaleido` installations : kaleido >= 1.0.0 requires plotly >= 6.1.1

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,6 +83,8 @@ plotly = ["nilearn[plotting]"]
 # See https://github.com/plotly/Kaleido/issues/134 for more detail
 plotting = [
     "matplotlib>=3.8.0",
+    # plotly 6.1.0 cannot save images
+    # https://github.com/plotly/plotly.py/issues/5187
     "plotly>=5.0.0,!=6.1.0 ; platform_system != 'Windows'",
     "kaleido ; platform_system != 'Windows'",
     "plotly>=5.0.0,<6.1.0 ; platform_system == 'Windows'",


pyproject.toml
```diff
@@ -83,7 +83,7 @@ plotly = ["nilearn[plotting]"]
 # See https://github.com/plotly/Kaleido/issues/134 for more detail
 plotting = [
     "matplotlib>=3.8.0",
-    "plotly>=5.0.0,!=6.1.0 ; platform_system != 'Windows'",
```
Remi-Gau, line 86: I vaguely remember there was an issue with that version of plotly but I cannot recall: have you tried running the plotly tests with it to make sure it's ok?

nilearn/_utils/helpers.py
```diff
@@ -280,6 +283,24 @@ def is_kaleido_installed():
     return True
 
 
+def is_kaleido_plotly_compatible():
```
Remi-Gau: maybe this could go in nilearn._utils.versions ?

not 100% convinced either way, so will let you choose

pyproject.toml
hndgzkn: Yes, you are right : https://github.com/plotly/plotly.py/issues/5187

I reverted it.

Shall I add a comment on that?

nilearn/_utils/helpers.py
hndgzkn: I moved, I agree it is better.

pyproject.toml
Remi-Gau: yes let's add a comment in the pyproject.toml with a link to the plotly issue: future us will be glad about it

hndgzkn: Added.

PR #5641, opened 2025-09-02
[ENH] pass extra parameters to embedded sklearn estimators

Remi-Gau:
- Closes #5558 

Changes proposed in this pull request:

- make it possible to pass extra parameters to embedded sklearn estimators

## TODO

- [x] changelog
- [x] improve verbosity checks of estimators
- [x] double check if the -1 is needed (https://github.com/nilearn/nilearn/pull/5641#discussion_r2947604488)

Issue #5584, opened 2025-08-14
Decoder, SearchLight, FREMClassifier are not idempotent


Remi-Gau:
see https://github.com/nilearn/nilearn/pull/5566/files#r2275685744

Decoder, SearchLight, FREMClassifier seem to give different results (some VERY different) 

when doing

``python
estimator.fit(X)
result = estimator.method(X, **kawrgs)
```

and

```python
estimator.fit(X).fit(X)
result = estimator.method(X, **kawrgs)
```

for some of the following method: "predict", "transform", "decision_function"

PR #3801, opened 2023-07-04
[MAINT] deal with new warnings related to sklearn 1.3

Remi-Gau:
Relates to #3800 

Changes proposed in this pull request:

silence some warnings in doc build:
- setting defaults for dual explicitly in LinearSVC
- settings default in  when fitting sklearn.linear_model.OrthogonalMatchingPursuit

PR #6020, opened 2026-02-16
[ENH] pass extra arguments for estimator to Searchlight

### LINKED ISSUE #5589: SearchLight estimator give different result when in a pipeline
yeah preferably: the nilearn/_utils/estimator_checks.py file is peppered with a whole bunch of TODOs and FIXMEs for a bunch of estimators that fail to meet some of the sklearn requirements for some 'edge' case.

I'd rather concentrate on the easy 80% for now, and then ~~lose what's left of may sanity on~~ work on the remaining 20%.

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/5560#discussion_r2269073038_

Remi-Gau:
- Closes #5589
- Relates to #5584
- Extracted from #5641
- Relates to #5558 

- Searchlight underlying estimator may give irreproducible calls if we do not set their random_state
- add a way to pass extra parameters to the underlying estimator

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,6 +86,8 @@ pattern.txt
 examples.txt
 gitlog.txt
 merge.txt
+__marimo__
+tmp.html
 
 # examples
 results/
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -34,14 +34,14 @@ NEW
 Fixes
 -----
 
-- :bdg-secondary:`Maint` Allow local installation with ```uv sync`` (:gh:`6024` by `Mathieu Dugré`_)
+- :bdg-secondary:`Maint` Allow local installation with ``uv sync`` (:gh:`6024` by `Mathieu Dugré`_)
 
 - :bdg-info:`Plotting` Fix ``nilearn.plotting.view_img`` resampling of non-isotropic images when no background image is used (:gh:`6031` by `Michelle Wang`_).
 
 Enhancements
 ------------
 
-- :bdg-success:`API` :class:`~decoding.Searchlight` now has a ``random_state`` and ``estimator_args`` to pass to the underlying estimator (:gh:`6020` by `Rémi Gau`_).
+- :bdg-success:`API` :class:`~nilearn.decoding.SearchLight` now has a ``random_state`` and ``estimator_args`` to pass to the underlying estimator (:gh:`6020` by `Rémi Gau`_).
 
 - :bdg-info:`Plotting` Allow string threshold in ``nilearn.plotting.plot_*`` functions (:gh:`5982` by `Saeed Babadi`_).
 


Remi-Gau:
wonder if this may explain some of the slightly different results we get on the searchlight example across our versions: https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_searchlight.html#f-scores-computation


## dev

 

## 0.13.1 (stable)

 

## 0.13.0 

 

## 0.12.0

nilearn/decoding/searchlight.py
```diff
@@ -437,11 +444,19 @@ def fit(self, imgs, y, groups=None):
...
 
+        self.estimator_args_ = (
+            {} if self.estimator_args is None else self.estimator_args
+        )
+
         estimator = self.estimator
         if estimator == "svc":
-            estimator = ESTIMATOR_CATALOG[estimator](dual=True)
+            estimator = ESTIMATOR_CATALOG[estimator](
+                dual=True,
```
Remi-Gau: given that `dual=True` can lead to irreproducible results I am tempted to somehow hard code a random_state for svc because I doubt users will otherwise set their random_state by passing extra arguments to do it.

@bthirion WDYT?

nilearn/_utils/docs.py
```diff
@@ -482,6 +482,14 @@ def custom_function(vertices):
     If None, data will not be converted to a new data type.
 """
 
+# estimator_args
+docdict["estimator_args"] = """
+estimator_args : dict[str, Any] or None, default=None
+    Extra parameters to pass to the sklearn learn estimators.
```
```suggestion
    Extra parameters to pass to the scikit-learn estimators.
```

nilearn/_utils/docs.py
```diff
@@ -1573,6 +1581,9 @@ def custom_function(vertices):
 
 estimator_ : Estimator object used during decoding.
 
+estimator_args_ : dict[str, Any]
+    Extra parameters passed to the sklearn learn estimators.
```
```suggestion
    Extra parameters passed to the scikit-learn estimators.
```

nilearn/_utils/docs.py
```diff
@@ -482,6 +482,14 @@ def custom_function(vertices):
     If None, data will not be converted to a new data type.
 """
 
+# estimator_args
+docdict["estimator_args"] = """
+estimator_args : dict[str, Any] or None, default=None
+    Extra parameters to pass to the scikit-learn estimators.
+
+    .. nilearn_versionadded:: 0.13.0dev
```
```suggestion
    .. nilearn_versionadded:: 0.13.1dev
```

nilearn/decoding/searchlight.py
bthirion: In the first place, do we really need dual=True ? Is it supposed to save CPU time ?
Providing a  random_state is fine imho.

Remi-Gau: dual was set explicitly to True or False in this old PR: https://github.com/nilearn/nilearn/pull/3801 because the new default was going to be 'auto'

for the doc on dual:

https://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVC.html#sklearn.svm.LinearSVC

> dual“auto” or bool, default=”auto”
>  Select the algorithm to either solve the dual or primal optimization problem. 
>  Prefer dual=False when n_samples > n_features.
>  dual="auto" will choose the value of the parameter automatically, based on the
>  values of n_samples, n_features, loss, multi_class and penalty.
>  If n_samples < n_features and optimizer supports chosen loss,
>  multi_class and penalty, then dual will be set to True, otherwise it will be set to False.

If we decide to switch that auto, we could still run into the issue of having dual=True sometimes and non-deterministic subsequent function calls, so I still think we should set the random_state somehow.

Remi-Gau: actually: dual="auto" is the default since sklearn 1.5 which is now the lowest sklearn version we support so maybe what we could do, is use dual="auto" everywhere and make sure we set a random_state in case dual gets set to True at fit time.

Remi-Gau: > actually: dual="auto" is the default since sklearn 1.5 which is now the lowest sklearn version we support so maybe what we could do, is use dual="auto" everywhere and make sure we set a random_state in case dual gets set to True at fit time.

but maybe we should do that first in a separate PR

nilearn/decoding/searchlight.py
```diff
@@ -268,6 +268,15 @@ class SearchLight(TransformerMixin, NilearnBaseEstimator):
...
 
+    %(random_state)s
+
+        .. note::
+
+            This parameter will be passed to the underlying
+            estimator.
+
+    %(estimator_args)s
+
```
Remi-Gau, line 279: - [x] change of API to document in the changelog

bthirion:
> wonder if this may explain some of the slightly different results we get on the searchlight example across our versions: https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_searchlight.html#f-scores-computation
> ## dev
> 
>  
> ## 0.13.1 (stable)
> 
>  
> ## 0.13.0
> 
>  
> ## 0.12.0
> 
>  

What is troubling is that the F maps , which is deterministic, also changes...

Remi-Gau:
Note that there is a small confound as the dependencies are not EXACTLY the same across these doc builds: only showing the ones that changed

## dev

```
    "numpy==1.26.0",
    "pandas==2.3.0",
    "scikit-learn==1.5.0",
    "scipy==1.11.1",
```

## 0.13.1 (stable)

```
    "numpy==1.22.4",
    "pandas==2.2.0",
    "scikit-learn==1.4.0",
    "scipy==1.9.0",
```

## 0.13.0 

```
    "numpy==1.22.4",
    "pandas==2.2.0",
    "scikit-learn==1.4.0",
    "scipy==1.9.0",
```

## 0.12.0

```
    numpy==1.22.4
    pandas==2.2.0
    scikit-learn==1.4.0
    scipy==1.8.0
```

Issue #1194, opened 2016-08-02
find_xyz_cut_coords default thresholding fails for binary inputs if integer type


AlexandreAbraham:
By default, a threshold is calculated on non-zero values (which is not exactly what is described in the doc).  If the image is binary, the threshold is 1, which masks everything.

`activation_threshold = fast_abs_percentile(my_map[my_map != 0].ravel(), 80)`

I will fix this in an upcoming PR.

GaelVaroquaux:
> By default, a threshold is calculated on non-zero values (which is not exactly
> what is described in the doc). If the image is binary, the threshold is 1,
> which masks everything.

> activation_threshold = fast_abs_percentile(my_map[my_map != 0].ravel(), 80)

Indeed. The standard solution is to remove 1e-7 to the threshold found.

AlexandreAbraham:
> Indeed. The standard solution is to remove 1e-7 to the threshold found.

I solve it by accepting values equal to the threshold. If failing on integer images is the intended behavior, we should raise an error.

GaelVaroquaux:
> I solve it by accepting values equal to the threshold.

Sounds good.

> If failing on integer images is the intended behavior, we should raise
> an error.

No, it's not.

PR #6025, opened 2026-02-17
[ENH] Allow plot_<> functions to accept threshold in percentile

### LINKED ISSUE #2061: Allow plot_<> functions to accept threshold in percentile

emdupre:
- Closes #2061
- Continues #5982

- We addes a threshold check call
- We tested locally and we got new behaviour

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -78,3 +78,7 @@ Changes
 - :bdg-dark:`Code` Remove aggressive garbage collection in safe_get_data for performance, mainly in CI. (:gh:`6039` by `Basile Pinsard`_).
 
 - :bdg-secondary:`Maint` Refactor and speed up tests for :func:`~plotting.view_img` (:gh:`6072` by `Michelle Wang`_)
+
+- :bdg-secondary:`Maint` Several optional dependencies (``dev``, ``test``, ``doc``...) have been swapped in favor of `dependency groups <https://packaging.python.org/en/latest/specifications/dependency-groups>`_, so they will no longer work when installing Nilearn from pypi. When installing locally, instead of ``pip install '.[dev]'`` do ``pip install . --group dev`` (:gh:`6108` by `Rémi Gau`_).
+
+- :bdg-secondary:`Maint` The optional dependency ``plotly`` has been removed in favor of ``plotting`` when so instead of ``pip install 'nilearn[plotly]'`` do ``pip install 'nilearn[plotting]`` (:gh:`6108` by `Rémi Gau`_).
--- a/.github/workflows/test_atlases.yml
+++ b/.github/workflows/test_atlases.yml
@@ -50,7 +50,7 @@ jobs:
             run: python -m pip install --upgrade pip setuptools rich templateflow
 
         -   name: Install nilearn
-            run: pip install .[plotting]
+            run: pip install '.[plotting]'
 
         -   name: Test atlases
             id: test_atlas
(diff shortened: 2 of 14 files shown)


nilearn/plotting/image/tests/test_plot_img.py
```diff
@@ -63,6 +63,11 @@ def test_display_methods_invalid_threshold(matplotlib_pyplot, img_3d_mni):
         )
 
 
+def test_check_string_threshold(matplotlib_pyplot, img_3d_mni):
+    """Checks threshold can be passed as string percentage value."""
+    plot_img(img_3d_mni, threshold="97%")
```
bthirion, line 68: Don't we try anything more than a smoke test ?

doc/changes/latest.rst
```diff
@@ -14,6 +14,9 @@ Fixes
 Enhancements
 ------------
 
+
+- :bdg-info:`Plotting` Allow string threshold in :func:`~plotting.plot_img` (:gh:`5982` by `Saeed Babadi`_).
```
hndgzkn: This affects some other -may be all- plotting functions in `plotting.image.img_plotting` module.

emdupre: Good point, updating !

emdupre:
Thank you, @hndgzkn !

Also, just to say, your comment on the previous PR was not lost :

> In either case, it is not clear to me why we subtract -1e-5 here, but not in check_threshold.

> Isn't the reasoning below valid for all thresholded data?

https://github.com/nilearn/nilearn/blob/563c57572e94630bbb94100bb20597e0f8d533d7/nilearn/plotting/image/img_plotting.py#L229-L232

I think, though, this is likely better discussed in a new issue !

nilearn/plotting/image/tests/test_plot_img.py
emdupre: Would you recommend updating one of the examples, or ?

bthirion: That would be great indeed.

hndgzkn:
> Thank you, @hndgzkn !
> 
> Also, just to say, your comment on the previous PR was not lost :
> 
> > In either case, it is not clear to me why we subtract -1e-5 here, but not in check_threshold.
> 

> I think, though, this is likely better discussed in a new issue !

I am opening an issue on this then. 

While checking code and other issues on thresholding, I've came accross #1194 which indicates the same problem. Then I realized that with #4951 I've changed thresholding behavior to exclude values equal to threshold. It might be good to revert it if we decide that actually the values equal to threshold should stay.

doc/changes/latest.rst
```diff
@@ -41,6 +41,9 @@ Fixes
 Enhancements
 ------------
 
+
+- :bdg-info:`Plotting` Allow string threshold in :func:`~plotting.plot_*` functions (:gh:`5982` by `Saeed Babadi`_).
```
```suggestion
- :bdg-info:`Plotting` Allow string threshold in ``nilearn.plotting.plot_*`` functions (:gh:`5982` by `Saeed Babadi`_).
```

Remi-Gau: this was causing the doc build to fail

emdupre: Thank you !

nilearn/plotting/image/tests/test_plot_img.py
emdupre: Thank you @Remi-Gau for opening #6124 on this ! I had been having trouble finding an example that made sense to adapt from the [plotting section ](https://nilearn.github.io/stable/auto_examples/01_plotting/index.html); very happy for recommendations.

Issue #5415, opened 2025-06-03
lowpass butterworth and cosine


bpinsard:
I have two observations regarding the lowpass filtering:
- With `filter_type='cosine'` silently skipping lowpass, some users might report low-pass filtering while it is not applied, and while it mentioned in the docstring, maybe a warning that shows in the logs?

- Due to border effects of the current low-pass filter config, I wonder if it is really 🧈worth keeping it as an option, as it could cause more harm than good. Here is an extreme example with 2 sine waves where it indeed removes the high frequency but has large border effects.  

```
raw = np.cos(np.arange(1000)) + np.sin(np.arange(1000)*.01)
filt = nilearn.signal.clean(raw[:, np.newaxis], low_pass=.1, t_r=1.0)
plot(raw)
plot(np.squeeze(filt))
```

![Image](https://github.com/user-attachments/assets/e797ea06-3521-4e29-b985-f6eb97aeff22). 

Are the default scipy butterworth parameters suited for fMRI? Would there be a signal-processing principled way to set defaults more adequates for the frequency of interest in fMRI? For example, assuming that BOLD signal "comes and goes", "even" padding could better reflect that, or "constant". 
Existing discussion here https://github.com/nilearn/nilearn/issues/3475#issuecomment-1408714387 from @tsalo .

bthirion:
> * With `filter_type='cosine'` silently skipping lowpass, some users might report low-pass filtering while it is not applied, and while it mentioned in the docstring, maybe a warning that shows in the logs?

But this sounds like a bug: we should actually also apply the lowpass filtering when  `filter_type='cosine'`. If not, what is the reason ?

bthirion:
> I have two observations regarding the lowpass filtering:
> 
>     * With `filter_type='cosine'` silently skipping lowpass, some users might report low-pass filtering while it is not applied, and while it mentioned in the docstring, maybe a warning that shows in the logs?
> 
>     * Due to border effects of the current low-pass filter config, I wonder if it is really 🧈worth keeping it as an option, as it could cause more harm than good. Here is an extreme example with 2 sine waves where it indeed removes the high frequency but has large border effects.
> 
> 
> ```
> raw = np.cos(np.arange(1000)) + np.sin(np.arange(1000)*.01)
> filt = nilearn.signal.clean(raw[:, np.newaxis], low_pass=.1, t_r=1.0)
> plot(raw)
> plot(np.squeeze(filt))
> ```
> 
[...]

PR #6027, opened 2026-02-17
[FIX] runtime warning of no low-pass with cosine

bpinsard:
- partially addresses #5415 

- add runtime warning for cosine lowpass
- change docstring from note to warning for consistency

### DIFF
--- a/nilearn/tests/test_signal.py
+++ b/nilearn/tests/test_signal.py
@@ -1017,20 +1017,6 @@ def test_clean_frequencies_using_power_spectrum_density():
         t_r=t_r,
     )
 
-    # check that user is warned about low-pass not applied with cosine
-    with pytest.warns(
-        UserWarning, match="low_pass is not implemented for filter='cosine'"
-    ):
-        res_cos = clean(
-            sx,
-            detrend=False,
-            standardize=None,
-            filter="cosine",
-            low_pass=low_pass,
-            high_pass=high_pass,
-            t_r=t_r,
-        )
-
     # Compute power spectrum density for both test
     f, Pxx_den_low = scipy.signal.welch(np.mean(res_low.T, axis=0), fs=t_r)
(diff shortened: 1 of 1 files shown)


nilearn/signal.py
```diff
@@ -718,6 +718,11 @@ def clean(
         confounds = _create_cosine_drift_terms(
             signals, confounds, high_pass, t_r
         )
+        if low_pass is not None:
+            warnings.warn(
+                "low_pass is not implemented for filter='cosine'",
+                stacklevel=2,
```
Remi-Gau:
```suggestion
                stacklevel=find_stack_level()
```

using find_stack_level (a function of ours) should help giving a more useful info to the user about what part of their code raised the warning:

https://github.com/nilearn/nilearn/blob/63c509d8a05b0c96a6abc5a6614bcaf0ede8ece4/nilearn/_utils/logger.py#L150

nilearn/tests/test_signal.py
```diff
@@ -1023,6 +1023,20 @@ def test_clean_frequencies_using_power_spectrum_density():
...
+    ):
+        res_cos = clean(
+            sx,
+            detrend=False,
+            standardize=None,
+            filter="cosine",
+            low_pass=low_pass,
+            high_pass=high_pass,
+            t_r=t_r,
+        )
```
Remi-Gau: could you extract that into a separate tests just for the warning, or maybe there is a test that checks all warnings of clean so it could be added in there

bthirion:
Shouldn't we also deprecate the use of 'cosine' ?

Remi-Gau:
the test failure is unrelated

nilearn/tests/test_signal.py
Remi-Gau: fixed

Issue #5588, opened 2025-08-19
Memory issue when fitting "large" image in GLM


Remi-Gau:
Adapted from https://neurostars.org/t/nilearn-firstlevelmodel-fit-ram-memory-explosion/34009/1

Images with "normal" resolution but a lot of time points become hard analyze.

The following code will crash because of memory error.

```python
from nilearn._utils.data_gen import generate_fake_fmri_data_and_design
from nilearn.datasets import load_mni152_brain_mask
from nilearn.glm.first_level import FirstLevelModel

mni_brain_mask = load_mni152_brain_mask()

shape = [(64, 76, 64, 6804)]
mask, fmri_data, design_matrices = generate_fake_fmri_data_and_design(shapes=shape, affine=mni_brain_mask.affine)

model = FirstLevelModel(t_r = 0.25, mask_img=mask)
model.fit(fmri_data, design_matrices=design_matrices)
```

Remi-Gau:
not sure that for something like this, the example we give on how to work with large images will be enough

emdupre:
Agreed that our current example doesn't explicitly cover this and the translation to a GLM isn't obvious. Sounds like a point of improvement 😸 

I'm wondering though : 6804 time points for a single run is a lot of time points ; I'd be curious to know if they really acquired it as a single run, or if it's been concatenated across multiple acquisitions. If the latter, a chunking strategy (as you suggested on NeuroStars, @Remi-Gau) seems quite reasonable.

Remi-Gau:
- 6800 with a fast TR (0.25 s) gives a bit less than half an hour. 
- that's a long run for task fmri
- less so for resting state but running a GLM on resting state is not super common in Nilearn - well except for https://github.com/nilearn/nilearn/issues/4165 - but I have seen it done in SPM as a strategy for denoising

note that I have not checked what's the number of TR that required to nuke my 32 Go of RAM, that could us a more sensible bound to work with (and maybe something to mention in our doc)

Remi-Gau:
things start to go BOINK when I tried with images 
- 1 run of (64, 64, 64, 6500)
- 3 runs of (64, 64, 64, 4000)
- 15 runs of (64, 64, 64, 1000)
- 30 runs of (64, 64, 64, 500)

So there also seems to be some issue with 'many' 'long' runs: need to check if that can maybe already be optimized.

bthirion:
The question is whether we create too many copies of the data in temporary arrays that blow up the memory.
[...]

Issue #2421, opened 2020-04-17
nilearn.stats AR is memory heavy?


kchawla-pi:
I discovered this snippet in this Issue (https://github.com/mne-tools/mne-nirs/issues/58#issue-601039480)

> The code is taking up too much RAM to run a full analysis in the GitHub runner. It seems to be due to the AR code in Nistats. So currently analysing just the first third of the measurement in the example. Either I move to a different CI or make the code more memory efficient.

Is this something that can/should be looked into?

Perhaps its time to redo #736

rob-luke:
Thank you @kchawla-pi  for considering my comment. I haven't had time to investigate this issue further yet, but have highlighted the issue in this PR (https://github.com/mne-tools/mne-nirs/pull/64). 

When I dig in to this further I will report back in this issue and in the PR above.

larsoner:
My guess is that it's unrelated to #736 since @rob-luke's code won't even use an MRI image. My guess is that it has to do with constructing a design matrix at the fNIRS sampling rate, which is probably higher than fMRI. Might need to look into using a `scipy.sparse` matrix instead.

But I could be way off, I just know we've needed to do this sort of thing in MNE when doing linear regressions...

PR #6030, opened 2026-02-17
FIX: Refactor GLM results for memory efficiency and core API stability (#2381)

### LINKED ISSUE #2381: Simplify the LikelihoodModelResults.
Stores un-necessary stuff (input data, residuals), hence memory-consuming.
Partially handled in nilearn/nistats#72, but full removal should be considered.

BabaSanfour:
Refactors GLM results to improve memory efficiency by eliminating redundant storage of large data arrays (input images, whitened data, and residuals) within the `RegressionResults` objects.

### Key Changes
- **Memory Optimization**: `RegressionResults` no longer stores `Y`, `whitened_Y`, or `residuals` by default. These are now computed or retrieved via methods only when explicitly requested.
- **Summary Statistics**: Common statistics like `SSE`, `r_square` and `MSE` are now computed once during `fit()` and stored as pre-calculated static attributes, maintaining fast performance for most use cases (like maps and summaries).

### API Evolution
Most voxel-wise attributes have transitioned from attributes/properties to methods.

#### Residuals and Predictions
- **Before**: 
  ```python
  res = model.residuals    # Attribute (High memory)
  pred = model.predicted   # Attribute
  ```
- **After**: 
  ```python
  res = model.residuals(Y)  # Method requiring input data 'Y'
  pred = model.predicted()  # Method call
  ```

#### Summary Statistics (R-squared)
- **Before**: 
  ```python
  r2 = model.r_square      # Property
  ```
- **After**: 
  ```python
  r2 = model.r_square()    # Method call (returns pre-calculated map)
  ```

#### Normalized Residuals
- **Before**: 
  ```python
  norm_res = model.normalized_residuals  # Attribute
  ```
- **After**: 
  ```python
  norm_res = model.normalized_residuals(Y)  # Method requiring input data 'Y'
  ```

### Verification Results
- **Code Quality**: Full cleanup with `ruff` and `pre-commit` validation across all modified files.
- **Test Suite**: 304 non-BIDS GLM tests passing (Regression, Model, First/Second Level, Reporter).

### Next Step Before Merging
- **Phase 2**: Merge and eventually deprecate `SimpleRegressionResults`. The new `RegressionResults` is now efficient enough to serve both standard and memory-minimized use cases.
- **Phase 3**: Audit and update `io.py` and BIDS saving functions. 

> [!NOTE]
> Currently, tests for `io.py` and some edge cases with `minimize_memory=True` (specifically for late-stage statistical map saving) are expected to fail. I will work on resolving these while addressing Phases 2 and 3.

### Linked Issues
- Fixes #2381 / Potentially also #5588
- Might help with #2421

### DIFF
--- a/nilearn/glm/second_level/second_level.py
+++ b/nilearn/glm/second_level/second_level.py
@@ -828,7 +828,7 @@ class SecondLevelModel(BaseGLM):
         }
 
     def _get_element_wise_model_attribute(
-        self, attribute, result_as_time_series
+        self, attribute, result_as_time_series, Y=None
     ):
         """Transform RegressionResults instances within a dictionary \
         (whose keys represent the autoregressive coefficient under the 'ar1' \
@@ -846,6 +846,10 @@ class SecondLevelModel(BaseGLM):
             whether the RegressionResult attribute has a value
             per timepoint of the input nifti image.
 
+        Y : Niimg-like object, default=None
+            The data from which to compute the attribute.
+            Required for 'residuals' and 'normalized_residuals'.
+
         Returns
         -------
         output : :obj:`list`
@@ -854,10 +858,10 @@ class SecondLevelModel(BaseGLM):
         """
         check_is_fitted(self)
         # check if valid attribute is being accessed.
-        all_attributes = dict(vars(RegressionResults)).keys()
         possible_attributes = [
-            prop for prop in all_attributes if "__" not in prop
+            prop for prop in dict(vars(RegressionResults)) if "__" not in prop
         ]
+        possible_attributes += RegressionResults.__static_attributes__
         if attribute not in possible_attributes:
             msg = f"attribute must be one of: {possible_attributes}"
             raise ValueError(msg)
@@ -889,11 +893,22 @@ class SecondLevelModel(BaseGLM):
         else:
             voxelwise_attribute = np.zeros((1, len(self.labels_)))
 
+        Y_matrix = self.masker_.transform(Y) if Y is not None else None
+
         for label_ in self.results_:
             label_mask = self.labels_ == label_
-            voxelwise_attribute[:, label_mask] = getattr(
-                self.results_[label_], attribute
-            )
+            attr = getattr(self.results_[label_], attribute)
+            if attribute in ["residuals", "normalized_residuals"]:
+                if Y_matrix is None:
+                    raise ValueError(f"Attribute {attribute} requires data Y.")
+                val = attr(Y_matrix[:, label_mask])
+            elif attribute == "predicted":
+                val = attr()
+            else:
+                val = attr
+
+            voxelwise_attribute[:, label_mask] = val
+
         return self.masker_.inverse_transform(voxelwise_attribute)
 
 
--- a/nilearn/glm/_base.py
+++ b/nilearn/glm/_base.py
@@ -10,7 +10,6 @@ from typing import Any, Literal
 import numpy as np
 import pandas as pd
 from nibabel import Nifti1Image
-from nibabel.onetime import auto_attr
 from sklearn.utils import Bunch
 from sklearn.utils.estimator_checks import check_is_fitted
 
@@ -184,10 +183,14 @@ class BaseGLM(CacheMixin, NilearnBaseEstimator):
 
     # @auto_attr store the value as an object attribute after initial call
     # better performance than @property
-    @auto_attr
-    def residuals(self):
+    def residuals(self, Y):
         """Transform voxelwise residuals to the same shape \
         as the input Nifti1Image(s).
+
+        Parameters
+        ----------
+        Y : Niimg-like object
+            The data from which to compute the residuals.
 
         Returns
         -------
@@ -196,10 +199,9 @@ class BaseGLM(CacheMixin, NilearnBaseEstimator):
 
         """
         return self._get_element_wise_model_attribute(
-            "residuals", result_as_time_series=True
+            "residuals", result_as_time_series=True, Y=Y
         )
 
-    @auto_attr
     def predicted(self):
         """Transform voxelwise predicted values to the same shape \
         as the input Nifti1Image(s).
@@ -214,7 +216,6 @@ class BaseGLM(CacheMixin, NilearnBaseEstimator):
             "predicted", result_as_time_series=True
         )
 
-    @auto_attr
     def r_square(self):
         """Transform voxelwise r-squared values to the same shape \
         as the input Nifti1Image(s).
--- a/nilearn/glm/first_level/first_level.py
+++ b/nilearn/glm/first_level/first_level.py
@@ -213,7 +213,7 @@ def run_glm(
             raise ValueError(err_msg) from e
 
         # compute the AR coefficients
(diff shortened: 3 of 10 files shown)


BabaSanfour:
- I was checking the rest of the open issues in the Kanban with no status ( #5588 ) and I found this other issue that I think is directly related to the proposed fixes above. I reanalyzed the code and thought of 3 potential suggestions. I will better write these tomorrow and then maybe we can have another discussion @emdupre @bthirion , maybe @Remi-Gau also as they opened the last issue.

bthirion:
Note that this is an API change, so we'll need to go through deprecation cycles etc.

BabaSanfour:
#### Why the memory explodes? 

@emdupre, @bthirion, as mentioned [here](https://neurostars.org/t/nilearn-firstlevelmodel-fit-ram-memory-explosion/34009/11) (#5588), processing large datasets (e.g., shape `(64, 76, 64, 6804)`) causes Nilearn to "eat through 32GB of RAM". I think this is primarily due to a "multi-copy" regression result structure. 

For an array of that shape at `float64`, a single copy is **17 GB**. The current implementation creates multiple redundant copies:

1.  **Duplicate Input**: `LikelihoodModelResults` saves the data (`Y`) in the object, besides the copy already in the workspace or on disk.
2.  **Redundant Result Copies** in `RegressionResults`:
    *   **Initialization**: Both `self.whitened_Y` and `self.whitened_residuals` are stored immediately.
    *   **Caching**: Usage of `@auto_attr` permanently caches `self.residuals` and `self.normalized_residuals` in RAM.

If the initial input $Y$ is **1 GB**, the current implementation ballooned to **5 GB** of persistent RAM usage. For the 17 GB dataset cited above, the overhead reaches **85 GB**.

---

### What could work?

1.  **Option 1: The Clean & Safe Refactor (This PR)**: 
    *   Move from attributes to methods for all data-dependent results.
    *   100% memory efficiency; aligns with `scikit-learn` API (estimators don't store $Y$ nor transformation of $Y$).
2.  **Option 2: The Minimum Change**: 
    *   Only drop `Y` from being saved in `LikelihoodModelResults`.
    *   Only addresses 20% of the total bloat. Maybe more if we find options to remove `self.whitened_Y` and `self.whitened_residuals`. But we are _always_ still left with `Y`, `self. residuals` and `self. normalized_residuals`
3.  **Option 3: A Hybrid Approach**: 
    *   Add a `return_residuals` flag to `fit()`. If `True`, it computes and stores residuals as attributes (`self. residuals_` and `self.normalized_residuals_` using `self. residuals(Y)` and `self.normalized_residuals(Y)` ; otherwise, they stay as methods and user could compute them using `model.residuals(Y)`. Eventually we can have a `return_normalized_residuals` flag too.   
    *   This keeps the option for users who want "everything in a box".

Although I more fan of **Option 1**, I think **Option 3** could eventually be the way to go. It also allows us to eventually **merge `RegressionResults` and `SimpleRegressionResults`** into a single, unified logic for handling GLM results and removes the need for manual memory workarounds.

For the API change, I can add deprecations into account and take care of that for the next 2/3 releases.

nilearn/glm/second_level/second_level.py
```diff
@@ -854,10 +858,10 @@ def _get_element_wise_model_attribute(
         """
         check_is_fitted(self)
         # check if valid attribute is being accessed.
-        all_attributes = dict(vars(RegressionResults)).keys()
         possible_attributes = [
-            prop for prop in all_attributes if "__" not in prop
+            prop for prop in dict(vars(RegressionResults)) if "__" not in prop
         ]
+        possible_attributes += RegressionResults.__static_attributes__
```
emdupre, line 934: I believe `__static_attributes__` was only added recently (maybe [as recently as 3.13](https://docs.python.org/fr/3/reference/datamodel.html#type.__static_attributes__)) ; we support back to 3.10, though, so we would need to use a more general method

emdupre: Could we revert to the previous formulation here ?

BabaSanfour: I changed this to avoid some tests that were crushing as we changed from attributes to methods. I could find a way around to go back to the prev formulation. This will be changed eventually depending on the Option we decide to go with.

bthirion:
> #### Why the memory explodes?
> 
> @emdupre, @bthirion, as mentioned [here](https://neurostars.org/t/nilearn-firstlevelmodel-fit-ram-memory-explosion/34009/11) (#5588), processing large datasets (e.g., shape `(64, 76, 64, 6804)`) causes Nilearn to "eat through 32GB of RAM". I think this is primarily due to a "multi-copy" regression result structure.
> 
> For an array of that shape at `float64`, a single copy is **17 GB**. The current implementation creates multiple redundant copies:
> 
>     1. **Duplicate Input**: `LikelihoodModelResults` saves the data (`Y`) in the object, besides the copy already in the workspace or on disk.
> 
>     2. **Redundant Result Copies** in `RegressionResults`:
>        
>        * **Initialization**: Both `self.whitened_Y` and `self.whitened_residuals` are stored immediately.
>        * **Caching**: Usage of `@auto_attr` permanently caches `self.residuals` and `self.normalized_residuals` in RAM.
> 
> 
> If the initial input Y is **1 GB**, the current implementation ballooned to **5 GB** of persistent RAM usage. For the 17 GB dataset cited above, the overhead reaches **85 GB**.
> ### What could work?
> 
>     1. **Option 1: The Clean & Safe Refactor (This PR)**:
>        
>        * Move from attributes to methods for all data-dependent results.
>        * 100% memory efficiency; aligns with `scikit-learn` API (estimators don't store 
>            Y
>           nor transformation of 
>            Y
>          ).
> 
>     2. **Option 2: The Minimum Change**:
>        
>        * Only drop `Y` from being saved in `LikelihoodModelResults`.
>        * Only addresses 20% of the total bloat. Maybe more if we find options to remove `self.whitened_Y` and `self.whitened_residuals`. But we are _always_ still left with `Y`, `self. residuals` and `self. normalized_residuals`
> 
>     3. **Option 3: A Hybrid Approach**:
>        
>        * Add a `return_residuals` flag to `fit()`. If `True`, it computes and stores residuals as attributes (`self. residuals_` and `self.normalized_residuals_` using `self. residuals(Y)` and `self.normalized_residuals(Y)` ; otherwise, they stay as methods and user could compute them using `model.residuals(Y)`. Eventually we can have a `return_normalized_residuals` flag too.
>        * This keeps the option for users who want "everything in a box".
> 
> 
> Although I more fan of **Option 1**, I think **Option 3** could eventually be the way to go. It also allows us to eventually **merge `RegressionResults` and `SimpleRegressionResults`** into a single, unified logic for handling GLM results and removes the need for manual memory workarounds.
> 
> For the API change, I can add deprecations into account and take care of that for the next 2/3 releases.

Thx for the detailed analysis !
* Storing `whitened residuals`, `residuals` and `normalized residuals` seems overly redundant. If we could keep only one among 3, this would be beneficial.
* option 3 sounds the right approach. Just, the sklearn API requires to provide these flags at instantiation time, not fit time, but this is a detail.
This being said, I'll be super happy to see this happen !

Issue #6036, opened 2026-02-18
use brainsprite to display images in nifti maskers and GLM on volume data


Remi-Gau:
for nifti masker and GLM reporting (on volume) it would be nice to leverage brainsprite to be able to interactively visualize the masks, labels, maps, or GLM results...

PR #6037, opened 2026-02-18
[ENH] use brainsprite for Nifti and NiftiLabel maskers reports

Remi-Gau:
- Relates #6036 

- [x] changelog
- [x] use brainsprite for Nifti and NiftiLabel maskers reports
- [x] add tests
- [x] modify at least one example

### DIFF
--- a/nilearn/maskers/nifti_masker.py
+++ b/nilearn/maskers/nifti_masker.py
@@ -450,7 +450,7 @@ class NiftiMasker(ClassNamePrefixFeaturesOutMixin, BaseMasker):
         -------
         list of :class:`~matplotlib.figure.Figure` or None
         """
-        if self._report_content["engine"] == "brainsprite":
+        if self._report_content.get("engine") == "brainsprite":
             bg_img = self._reporting_data["images"]
             stat_map_img = self._reporting_data["mask"]
             self._create_brainsprite(bg_img=bg_img, stat_map_img=stat_map_img)


nilearn/reporting/data/html/maskers/body_masker.jinja
```diff
@@ -22,6 +22,28 @@
                         </div>
                     {% endif %}
                 </div>
+                {% if engine == "brainsprite" %}
```
Remi-Gau: use a partial for this?

Remi-Gau:
while working on this I want to check whether we can have a slider to dynamically adapt opacity in view_img without any change to brainsprite.

if we can it would be nice to make it part of using brainsprite for the maskers as it makes viewing masks much easier when they are overlaid on an image.

@lunebellec

nilearn/maskers/tests/test_baseline_comparisons.py
```diff
@@ -92,7 +92,7 @@ def test_nifti_labels_masker_create_figure_for_report(
 
     labels_img = math_img("img1 + img2", img1=positive_img, img2=negative_img)
 
-    masker = src_masker(labels_img, mask_img=mask_img)
+    masker = src_masker(labels_img, mask_img=mask_img, cmap="RdBu_r")
```
Remi-Gau, line 95: testing that modifying the cmap alters the background image

nilearn/maskers/nifti_masker.py
```diff
@@ -372,8 +373,8 @@ def __init__(
         self._report_content = {
             "description": (
                 "This report shows the input Nifti image overlaid "
-                "with the outlines of the mask (in green). We "
```
Remi-Gau, line 375: removing this detail to make it more generic so it applies to report generating with either engine

nilearn/plotting/html_stat_map.py
Remi-Gau, line 1: mostly adding type annotation to know what we are dealing with

Remi-Gau:
examples :  

https://output.circle-artifacts.com/output/job/9f0bc312-189e-45da-81ad-21a9d8a02da4/artifacts/0/dev/auto_examples/06_manipulating_images/plot_nifti_simple.html#visualize-the-masker-report

https://output.circle-artifacts.com/output/job/9f0bc312-189e-45da-81ad-21a9d8a02da4/artifacts/0/dev/auto_examples/06_manipulating_images/plot_nifti_labels_simple.html#visualize-the-atlas

bthirion:
Nice !  For the second one, we don't see the MNI brain template,  but only the labels image.

lunebellec:
I think it is really useful to have the interactive version.
For the mask in particular.

As mentioned, I could easily add a feature in the brainsprite js viewer so you could either display a montage with specific slices (either sagital, axial or coronal) or the interactive viewer.
I could even add an option to skip a certain number of slices to make it compact. The default value could be passed to the report maybe?
so you could adjust it programmatically if you wanted.
All the assets are already present in memory, it's just a matter of choosing what to show in the canvas.

For the second one, I think it would be great to enable the brainsprite option to display values so you can tell which region you're looking at. It's already implemented in the js, it just needs to be added to the python interface. It would be straightforward to support labels as well, but it would require some minor dev on the js component.
________________________________

Remi-Gau:
> we don't see the MNI brain template, but only the labels image

yup I had noticed that too.

will try to fix

Remi-Gau:
> As mentioned, I could easily add a feature in the brainsprite js viewer so you could either display a montage with specific slices (either sagital, axial or coronal) or the interactive viewer. I could even add an option to skip a certain number of slices to make it compact. The default value could be passed to the report maybe? so you could adjust it programmatically if you wanted. All the assets are already present in memory, it's just a matter of choosing what to show in the canvas.

Just to help set priorities: I don't think this is the most urgent new feature

> For the second one, I think it would be great to enable the brainsprite option to display values so you can tell which region you're looking at. It's already implemented in the js, it just needs to be added to the python interface. It would be straightforward to support labels as well, but it would require some minor dev on the js component.

I can try to make use of what's available already.

Also I indeed would like to make it easier for user to know what color corresponds to which region in the labels maskers in general.

displaying the name of the regions next to the colorbar won't scale well when you have hundreds of region, but it would be nice to be able to use display the color of the region in the look up table that the report contains: 

image

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -365,6 +382,8 @@ def _reporting(self):
         if not self._has_report_data():
             return None
 
+        from nilearn.plotting.image.utils import load_anat
```
Remi-Gau: should be importable outside of this function

Remi-Gau:
> > we don't see the MNI brain template, but only the labels image
> 
> yup I had noticed that too.
> 
> will try to fix

fixed now: 

image

examples/06_manipulating_images/plot_nifti_simple.py
```diff
@@ -89,7 +89,11 @@
 # by generating a masker report.
 # This can be done using
 # the :meth:`~nilearn.maskers.NiftiMasker.generate_report` method.
-report = masker.generate_report()
+# Here we use the 'brainsprite' engine
+# that gives an interactive vizualtion
```
hndgzkn:
```suggestion
# that gives an interactive visualization
```

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -163,9 +163,18 @@ class NiftiLabelsMasker(_LabelMaskerMixin, BaseMasker):
...
         Only relevant for the report figures.
 
+        When using ``"matplotlib"`` as an engire to create reports,
+        ``cmap`` will be used for the background image:
+        in this case if ``None`` is passed then the ``"CMRmap_r"`` colormap
+        will be used for the background image.
+        When using ``"brainsprite"`` as an engire to create reports,
+        ``cmap`` will be used for the label image:
+        in this case if ``None`` is passed then the ``"tab20"`` colormap
+        will be used for the label image.
```
hndgzkn:
```suggestion
        When using ``"matplotlib"`` as an engine to create reports,
        ``cmap`` will be used for the background image:
        in this case, if ``None`` is passed, then the ``"CMRmap_r"`` colormap
        will be used for the background image.
        When using ``"brainsprite"`` as an engine to create reports,
        ``cmap`` will be used for the label image:
        in this case, if ``None`` is passed, then the ``"tab20"`` colormap
        will be used for the label image.
```

nilearn/maskers/base_masker.py
```diff
@@ -621,6 +623,84 @@ def _check_array(
 
         return signals
 
+    def _create_brainsprite(
+        self,
+        bg_img: Nifti1Image | None,
+        stat_map_img: Nifti1Image,
+        cmap=None,
+    ) -> None:
```
hndgzkn, line 631: Would it be possible to put this function in `nilearn.plotting` module? As it is a plotting function, it seems more coherent to me to put it there. It also makes this function available to other modules.

That would require to remove the parts that call `self`.

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -413,7 +432,27 @@ def _reporting(self):
...
+                bg_img, _, _, _ = load_anat()
+            stat_map_img = self._reporting_data["labels_image"]
+
+            cmap = self.cmap
+            if cmap is None:
+                cmap = "tab20"
+
+            self._create_brainsprite(
+                bg_img=bg_img, stat_map_img=stat_map_img, cmap=cmap
+            )
```
hndgzkn, line 460: These might go inside `create_figure_for_report`. 

If `create_brainsprite` is moved inside plotting module, `create_figure_for_report` can import relevant plotting function depending on engine type.

nilearn/maskers/nifti_masker.py
```diff
@@ -382,14 +382,21 @@ def __init__(
             "warning_messages": [],
         }
 
-    def generate_report(self, title: str | None = None):
+    def generate_report(
+        self,
+        title: str | None = None,
+        engine: str = "matplotlib",
```
hndgzkn, line 388:
```suggestion
        engine: Literal["matplotlib", "brainsprite"] = "matplotlib",
```

nilearn/maskers/nifti_labels_masker.py
```diff
@@ -318,14 +327,21 @@ def _post_masking_atlas(self, visualize=False):
 
         return masked_atlas, removed_region_ids, removed_region_names, display
 
-    def generate_report(self, title: str | None = None):
+    def generate_report(
+        self,
+        title: str | None = None,
+        engine: str = "matplotlib",
```
hndgzkn, line 333:
```suggestion
        engine: Literal["matplotlib", "brainsprite"] = "matplotlib",
```

nilearn/maskers/_mixin.py
```diff
@@ -414,7 +414,11 @@ def _has_report_data(self):
         """
         return hasattr(self, "_reporting_data")
 
-    def generate_report(self, title: str | None = None):
+    def generate_report(
+        self,
+        title: str | None = None,
+        engine: str = "matplotlib",
```
hndgzkn, line 420: Missing docstring for `engine`.

nilearn/maskers/nifti_maps_masker.py
```diff
@@ -249,6 +249,7 @@ def generate_report(
         | int
         | Literal["all"] = 10,
         title: str | None = None,
+        engine: Literal["matplotlib", "brainsprite"] = "matplotlib",
```
hndgzkn, line 252: Missing docstring for `engine`.

nilearn/maskers/nifti_spheres_masker.py
hndgzkn, line 375: missing `engine` parameter, also in docstring.

nilearn/maskers/surface_maps_masker.py
hndgzkn, line 412: Maybe replace `engine:str ...` with `engine: Literal["matplotlib", "brainsprite", "plotly"] = "matplotlib",`

and update docstring accordingly.

hndgzkn: If brainsprite does not apply surface maskers, it could be updated only with plotly.

nilearn/reporting/data/html/maskers/body_masker.jinja
```diff
@@ -7,21 +7,27 @@
...
+                             alt="No image provided." />
+                        {% if overlay %}
+                            <div class="overlay">
+                                <img class="pure-img"
+                                     width="100%"
+                                     src="data:image/svg+xml;base64,{{ overlay }}"
+                                     alt="No overlay found." />
+                            </div>
+                        {% endif %}
+                    </div>
```
hndgzkn: Could this be extracted as `matplotlib.jinja`?

nilearn/maskers/nifti_masker.py
```diff
@@ -433,7 +441,14 @@ def _reporting(self):
...
 
-        return self._create_figure_for_report()
+        if self._report_content["engine"] == "matplotlib":
+            return self._create_figure_for_report()
+
+        elif self._report_content["engine"] == "brainsprite":
+            bg_img = self._reporting_data["images"]
+            stat_map_img = self._reporting_data["mask"]
+            self._create_brainsprite(bg_img=bg_img, stat_map_img=stat_map_img)
+            return None
```
hndgzkn: Here as well, these could go inside `create_figure_for_report`.

nilearn/maskers/_mixin.py
hndgzkn: Shall we put `Literal["matplotlib", "brainsprite"] = "matplotlib",` here or did you prefer to keep it simple as for some reports, it is also possible to have `plotly`. 

May be writing a docstring to clarify this might be good.

nilearn/reporting/html_report.py
```diff
@@ -298,6 +302,23 @@ def _create_report(
...
     body_tpl = env.get_template(body_tpl_path)
 
+    js_query_code = None
+    brainsprite_code = None
+
+    if data.get("engine") == "brainsprite":
+        with (OTHER_JS / "jquery.min.js").open("r") as f:
+            js_query_code = f.read()
+        with (OTHER_JS / "brainsprite.min.js").open("r") as f:
+            brainsprite_code = f.read()
```
hndgzkn, line 312: Isn't it possible to do this in the template?

nilearn/reporting/html_report.py
```diff
@@ -298,6 +302,23 @@ def _create_report(
...
+
+        if estimator._has_report_data():
+            data["bg_base64"] = estimator._reporting_data["bg_base64"]
+            data["cm_base64"] = estimator._reporting_data["cm_base64"]
+            data["params"] = estimator._reporting_data["params"]
+            data["stat_map_base64"] = estimator._reporting_data[
+                "stat_map_base64"
+            ]
```
hndgzkn, line 320: If these data is directly set in `report_content`, there will be no need to copy it here. Is there an inconvenience to do that?

nilearn/maskers/base_masker.py
Remi-Gau: to me this method is too specific to the nifti maskers to move it into plotting, a bit like the BaseSurfaceMasker has a _generate_figure method or like several maskers have a _create_figure_for_report method

nilearn/maskers/nifti_masker.py
Remi-Gau: That won't work: I have tried.
If we do this then mypy complains that we have to say that `engine` in nilearn.maskers._mixin._ReportingMixin.generate_report should also `Literal["matplotlib", "brainsprite"]` but that won't work then for the surface maskers that do not support brainsprite.

So I stuck to str for now.

nilearn/maskers/nifti_maps_masker.py
Remi-Gau: done

nilearn/maskers/nifti_spheres_masker.py
Remi-Gau: maybe let's see if we add support for brainsprite for this masker in another PR

nilearn/maskers/base_masker.py
hndgzkn: But a major part seems almost the same as `plotting.html_stat_map.view_img`. Would it be possible to reuse `view_img` (it returns the html representation at the end, which is not the case here) or extract common parts to a new function? 

That would help avoid importing many functions (including private ones) from plotting as well.

nilearn/reporting/html_report.py
Remi-Gau: We could by using jinja partials BUT to use those JS files as partials they should be somewhere under the `nilearn/reporting/data/html` folder and not in the `nilearn/plotting/data/html` folder where they are now: in general I would be in favor of moving all HTML and JS files in one single place independently of if they are used for plotting or reporting.

I think this is beyond the scope of this PR though.

nilearn/maskers/base_masker.py
Remi-Gau: OK let me see if I can do some refactoring in plotting.html_stat_map first then!

nilearn/reporting/html_report.py
hndgzkn: > in general I would be in favor of moving all HTML and JS files in one single place independently of if they are used for plotting or reporting.
> 
> I think this is beyond the scope of this PR though.

I think it's a very good idea.

nilearn/maskers/base_masker.py
Remi-Gau: extracted most of view_img in a single create_brainsprite function

doc/visual_testing/notebook_reports_maskers.py
```diff
@@ -52,7 +52,7 @@ def _(cli_parser):
     BUILD_TYPE = args.build_type
     if isinstance(BUILD_TYPE, list):
         BUILD_TYPE = BUILD_TYPE[0]
-    # BUILD_TYPE = "full"
+    BUILD_TYPE = "full"
```
Remi-Gau:
```suggestion
    # BUILD_TYPE = "full"
```

nilearn/maskers/nifti_masker.py
Remi-Gau: done

nilearn/maskers/nifti_labels_masker.py
Remi-Gau: > might go inside create_figure_for_report
done

bthirion:
The example LGTM. I understand that you don(t want to include more refactoring in this PR ?

Remi-Gau:
> I understand that you don't want to include more refactoring in this PR ?

yup: I think we can have a PR just dedicated to moving HTML and JS

Issue #2209, opened 2019-11-12
get_data (via load_niimg) casts to float64 (high memory usage)


adelavega:
What version of Nilearn are you using? master branch
What were you trying to do? Load a functional image (stored as integer type)
What did you expect will happen? Load in float32
What actually happened? It cast to float64 prior to casting back to float32

---

Due to https://github.com/nipy/nibabel/issues/832, nilearn's implementation of `get_data`, will always cast image data to `float64` even when dtype requested is `float32`.

This leads to really high, but temporary memory spikes. For example, loading a `.nii.gz` 4d image that is `int16` on disk, and ~398M:

```
img = load_niimg(img_path, dtype='float32')
```

Has the following memory profile:
![Figure_1](https://user-images.githubusercontent.com/2774448/68696165-78c33800-054a-11ea-9ba8-f292a7991d35.png)

A fix for this is coming in nibabel (https://github.com/nipy/nibabel/pull/833), but it will be using `get_fdata`, and in nibabel 3.0. Using this fix, I get the following memory profile for the same file:
![Figure_2](https://user-images.githubusercontent.com/2774448/68697195-57634b80-054c-11ea-803d-0979ade97006.png)

I understand that nilearn wishes to continue using `get_data` that will be deprecated in `nibabel` (https://github.com/nipy/nibabel/issues/834)

One potential compromise would be to continue using `get_data` (now coped to here) when the user does not specify a float type, but when a dtype=float is specified, use `nibabel.get_fdata`. This would ensure some backwards compatability while still allowing more memory efficient data loading.

GaelVaroquaux:
> One potential compromise would be to continue using get_data (now coped to here) when the user does not specify a float type, but when a dtype=float is specified, use nibabel.get_fdata.

Sounds right!

Do you want to send a PR?

adelavega:
Sure

jeromedockes:
why not use `image.load_img(img).get_fdata()` ? or `nibabel.load_img(img).get_fdata()` ?

adelavega:
That's exactly what I'm suggesting, but there seems to be interesting in maintaining previous behavior in `load_niimg` for backwards compatibility. So I'm suggesting only using `get_fdata` when a `dtype` is specified. 

I'm suggesting this change so that `nilearn` and other programs that use `load_niimg` will benefit from the memory improvements.

effigies:
[...]

PR #6039, opened 2026-02-18
 [ENH] test without aggressive garbage collection

bpinsard:

- Simple test to check how fast tests run without agressive garbage collection

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -1,7 +1,6 @@
 """Neuroimaging file input and output."""
 
 import collections.abc
-import gc
 from copy import deepcopy
 from pathlib import Path
 from warnings import warn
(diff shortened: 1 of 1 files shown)


bthirion:
So, do we keep it ? Reducing time by a factor 3 is not minor...

bpinsard:
Running the build-docs gives comparable CPU/memory-usages, so maybe we can trust python automatic garbage collection.
That `gc.collect` was introduced at a time where RAM was more scarce and python was likely ~2.6 so many things have changed since then.

Remi-Gau:
@emdupre can we quickly cross-check this morning about this PR? I am very tempted to merge it so we can speed up tests during the sprint

emdupre:
> @emdupre can we quickly cross-check this morning about this PR? I am very tempted to merge it so we can speed up tests during the sprint

Happy to chat in person ; in general I think I'm good to see this merged !
The only thing I can think is we _might_ want to re-run [this check](https://github.com/nilearn/nilearn/issues/2209#issuecomment-3917261771) just to be extra sure that we don't see any change in the memory usage for `get_data` without the manual `gc.collect`.

mathdugre:
Looks the same

```python
from pathlib import Path

import nibabel as nib
from nilearn._utils.niimg import safe_get_data

img_path = Path(__file__).parent / "rand_img.nii.gz"

img = nib.load(img_path)
data = safe_get_data(img)
```

**Before:**
get_safe_data-before

**After:**
get_safe_data-after

PR #6041, opened 2026-02-18
[PERF] Parallel signal extraction for img_to_signals_labels

### LINKED ISSUE #6038: [PERF] Parallelize img_to_signals_labels extraction
#2693 Suggested parallelized label extraction across image samples to improve performance but the PR was never merged. Parallellizing would greatly improve performance on large images since scipy's ndimage reduction utilities don't leverage batching.

pbarbarant:
- Closes #6038, based on initial #2693 proposal.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -257,10 +257,12 @@ jobs:
             uses: actions/cache/restore@v5
             with:
                 path: nilearn_data
-                key: data_cache-${{ github.workflow }}_ref-${{ github.ref }}_run-${{ github.run_number }}_attempt-${{ github.run_attempt }}
+                key: data_cache-${{ github.workflow }}_month-${{ hashFiles('month_num') }}_ref-${{ github.ref }}_run-${{ github.run_number }}_attempt-${{
+                    github.run_attempt }}
                 restore-keys: |
-                    data_cache-${{ github.workflow }}_ref-${{ github.ref }}_run-${{ github.run_number }}
-                    data_cache-${{ github.workflow }}_ref-${{ github.ref }}
+                    data_cache-${{ github.workflow }}_month-${{ hashFiles('month_num') }}_ref-${{ github.ref }}_run-${{ github.run_number }}
+                    data_cache-${{ github.workflow }}_month-${{ hashFiles('month_num') }}_ref-${{ github.ref }}
+                    data_cache-${{ github.workflow }}_month-${{ hashFiles('month_num') }}
 
         -   name: Get data from a previous successful run for full builds
                   # only run it if we did not get the data from a previous attempt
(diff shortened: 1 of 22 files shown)


pbarbarant:
As discussed with @Remi-Gau, we might want to use the default loky backend, but this could lead to the (possibly large) labels_data array to be cloned.

bthirion (approved):
Maybe simply add a whatsnew entry

nilearn/regions/tests/test_signal_extraction.py
```diff
@@ -800,3 +800,30 @@ def test_img_to_signals_labels_non_float_type(target_dtype, rng):
...
+    length = 8
+
+    # Generate labels
+    labels = list(range(n_regions + 1))  # 0 is background
+    labels_img = generate_labeled_regions(shape, n_regions, labels=labels)
+
+    # Generate fake data
+    fmri_img, _ = generate_fake_fmri(
+        shape=shape, length=length, affine=labels_img.affine
+    )
```
Remi-Gau: We actually have fixtures in to help 'mock' those data more easily.
I think it may be worth finding a way to advertise them better for new contributors.

@pbarbarant let's quickly chat to see if we can improve their findability

Remi-Gau: May just be a matter of telling people to run

```
pytest nilearn/conftest.py --fixtures
```

ad make sure our doc strings are clear enough

Remi-Gau: @michellewang 
I'd be curious to have your take on this as your were asking what our mocks were?

nilearn/regions/signal_extraction.py
```diff
@@ -314,14 +319,14 @@ def img_to_signals_labels(
     data = safe_get_data(imgs, ensure_finite=True)
     target_datatype = np.float32 if data.dtype == np.float32 else np.float64
     # Nilearn issue: 2135, PR: 2195 for why this is necessary.
```
Remi-Gau, line 321: is this comment still relevant?
can you quickly check?

nilearn/regions/tests/test_signal_extraction.py
michellewang: @Remi-Gau I just ran `pytest nilearn/conftest.py --fixtures` and I think it would definitely be useful to mention that command in the docs. The only thing is that this will print out all available fixtures, which could be overwhelming since some of them are from plugins and don't have docstrings (e.g. those from `anyio`)

nilearn/regions/signal_extraction.py
pbarbarant: Looking into this

pbarbarant: Casting to `np.float32` fails `test_signal_extraction_with_maps_and_labels`. So it appears that this line is still relevant.

nilearn/regions/tests/test_signal_extraction.py
pbarbarant: I originally copy-pasted the tests from #2693. Now the fixtures are used.

nilearn/regions/signal_extraction.py
Remi-Gau: thanks for checking

PR #6052, opened 2026-02-18
[PERF] Add joblib support for parallel extraction in NiftiLabelsMasker

### LINKED ISSUE #6038: [PERF] Parallelize img_to_signals_labels extraction

pbarbarant:
- Closes #6038.
- Add an n_jobs parameter to the masker for parallelizing signal extraction across samples.

### DIFF
--- a/nilearn/maskers/tests/test_nifti_labels_masker.py
+++ b/nilearn/maskers/tests/test_nifti_labels_masker.py
@@ -1241,3 +1241,15 @@ def test_lut_shift(lut):
 
     assert masker.region_names_ == {0: "Frontal", 1: "Temporal"}
     assert masker.lut_["name"].to_list() == ["Frontal", "Temporal"]
+
+
+@pytest.mark.single_process
+def test_nifti_labels_masker_parallel(img_labels, img_fmri):
+    """Test NiftiLabelsMasker with n_jobs > 1."""
+    masker = NiftiLabelsMasker(labels_img=img_labels)
+    masker_parallel = NiftiLabelsMasker(labels_img=img_labels, n_jobs=2)
+
+    signals = masker.fit_transform(img_fmri)
+    signals_parallel = masker_parallel.fit_transform(img_fmri)
+
+    assert np.allclose(signals, signals_parallel)
--- a/nilearn/maskers/tests/test_nifti_labels_masker.py
+++ b/nilearn/maskers/tests/test_nifti_labels_masker.py
@@ -1244,7 +1244,7 @@ def test_lut_shift(lut):
 
 
 @pytest.mark.single_process
-def test_nifti_labels_masker_parallel(img_labels, img_fmri):
+def test_parallel(img_labels, img_fmri):
     """Test NiftiLabelsMasker with n_jobs > 1."""
     masker = NiftiLabelsMasker(labels_img=img_labels)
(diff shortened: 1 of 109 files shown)


pbarbarant:
This PR builds on #6041 which should be merged first

bthirion (approved):
Maybe add a whatsnew entry

Remi-Gau:
ok I think that's where we need to discuss how this affects the MultiNiftiLabelsMasker because in that case we will have nested parallelization with joblib

pbarbarant:
> ok I think that's where we need to discuss how this affects the MultiNiftiLabelsMasker because in that case we will have nested parallelization with joblib

As discussed with @tommoral, this should not be an issue. Though the users should be aware that they are spawning `n_subjects * n_jobs`.

bthirion:
> > ok I think that's where we need to discuss how this affects the MultiNiftiLabelsMasker because in that case we will have nested parallelization with joblib
> 
> As discussed with @tomMoral, this should not be an issue. Though the users should be aware that they are spawning `n_subjects * n_jobs`.

Do you mean to issue a warning if we detect that both n_subjects > 1 and n_jobs > 1 ?

tomMoral:
Sorry it was a bit unclear and in a rush.

What I said was that in principle, it is possible to use nested
parallelization with Processes and Threads.
By default, if you don't specify the backend and call nested joblib, the
first level is with processes, the second with threads and later on it is
sequential.

But for thread level parallelism, this only really parallelizes if the task
releases the GIL.
Otherwise, nested parallelization leads to slow down in many cases, so
might be safer to enforce `n_jobs=1`.

pbarbarant:
> Sorry it was a bit unclear and in a rush. What I said was that in principle, it is possible to use nested parallelization with Processes and Threads. By default, if you don't specify the backend and call nested joblib, the first level is with processes, the second with threads and later on it is sequential. But for thread level parallelism, this only really parallelizes if the task releases the GIL. Otherwise, nested parallelization leads to slow down in many cases, so might be safer to enforce `n_jobs=1`.

Thanks for your answer. In a nutshell, the inner loop calls `scipy.ndimage`  reductions function which, after checking, is actually implemented in plain python and not in cython (as I initially thought), thus the GIL is not released and we should stick to the `loky` backend.

nilearn/maskers/tests/test_nifti_labels_masker.py
```diff
@@ -1241,3 +1241,15 @@ def test_lut_shift(lut):
 
     assert masker.region_names_ == {0: "Frontal", 1: "Temporal"}
     assert masker.lut_["name"].to_list() == ["Frontal", "Temporal"]
+
+
+@pytest.mark.single_process
+def test_nifti_labels_masker_parallel(img_labels, img_fmri):
```
Remi-Gau:
```suggestion
def test_parallel(img_labels, img_fmri):
```

Remi-Gau:
> Otherwise, nested parallelization leads to slow down in many cases, so might be safer to enforce `n_jobs=1`.

what I understand from this, is that when using multi masker, parallelization will happen at the subject level and we should then set the n_job to be 1 when doing the signal extraction across labels

PR #6054, opened 2026-02-18
[MAINT] Refactoring of nilearn.signals::standardize_signal

### LINKED ISSUE #6051: [MAINT] Refactoring nilearn.signals.standardize_signal
- [ ] Convert asert statement to raise exception if condition is met.  https://github.com/nilearn/nilearn/blob/b029fb89b8d17596263a2018c3d70ff60682b8b0/nilearn/signal.py#L67-L68
- [ ] Simplify the logic of complexity by reducing nested code. https://github.com/nilearn/nilearn/blob/b029fb89b8d17596263a2018c3d70ff60682b8b0/nilearn/signal.py#L69-L109

mathdugre:
- Closes #6051 

- Convert the assert statement to conditional raise to ensure an exception is raised for Optimized python versions.
- Reduce the cognitive complexity of the function.

### DIFF
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -64,7 +64,6 @@ def standardize_signal(
 
     signals = _detrend(signals, inplace=False) if detrend else signals.copy()
 
-
     if standardize is None:
         return signals
 


bthirion (approved):
LGTM, thx.

nilearn/signal.py
```diff
@@ -64,46 +64,55 @@ def standardize_signal(
...
     signals = _detrend(signals, inplace=False) if detrend else signals.copy()
 
-    assert not isinstance(standardize, bool)
+    # Checks
+    if isinstance(standardize, bool):
+        raise ValueError(
+            "The 'standardize' parameter should be either None, 'psc', or "
+            "'zscore_sample'. The boolean value is not accepted anymore. "
+            "Please update your code accordingly."
+        )
```
Remi-Gau:
```suggestion
```

actually I realized that the check_parameter_in_allowed below should raise the proper error

nilearn/signal.py
Remi-Gau, line 44: something is off here

I think we should drop support for False and True later (like in 0.15).

The doc string says we can use them, but not the type annotation.

This is unrelated to this PR, but I need to check this: I think it's just that standardize_signal has never allowed boolean but some of other functions or methods have a standardize that allows boolean that then calls standardize_signal

PR #6056, opened 2026-02-18
[DOC] Add possibility of adding user-defined estimator without cv to searchlight

### LINKED ISSUE #2742: Add correlation-based tool for MVPA
I'd like to add a tool for multi-voxel pattern analysis (MVPA). It's adapted from the method described in _[Haxby et al., 2001](https://science.sciencemag.org/content/293/5539/2425)_, except searchlight spheres are analyzed instead of anatomical ROIs. I've already built the tool and it's available [here](https://github.com/shobrook/mvpa), but I would be happy to deprecate this and merge it into nilearn if you believe this will be a useful feature. 

<p align="center">

<br>
Fig. 3 from Haxby et al., 2001
</p>

### Benefits to the change

This tool allows users to identify areas of the brain containing multivariate representations, from which a pair of conditions can be decoded. This is similar to the set of [searchlight tools](https://nilearn.github.io/modules/reference.html#module-nilearn.decoding) provided by nilearn, but it is based on voxel-by-voxel correlations rather than SVM classification. Such correlations are notably much faster to compute, and results for a full brain can be generated within 10 minutes on a laptop, unlike the day(s) needed for the SVM searchlight. Additionally, the use of correlations as so is standard procedure for Representational Similarity Analysis, a popular analytic technique ([Kriegeskorte et al., 2008](https://www.frontiersin.org/articles/10.3389/neuro.06.004.2008/full); [Weaverdyck et al., 2020](https://academic.oup.com/scan/article/15/4/487/5824852)).

The code for this is available [here](https://github.com/shobrook/mvpa/blob/master/mvpa/correlation_searchlight.py), and I can adapt it to be part of the `nilearn.decoding` module.

mtorabi59:
- Closes #2742

- Adding the possibility of adding user-defined estimators without need for cross-validation (which is default behavior) to `SearchLight`.
- Creating a new tutorial showcasing definition of an MVPA-based estimator to be then applied through `SearchLight` on the Haxby data.


examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,195 @@
...
+
+    nilearn_searchlight_uses_cv = False
+
+    def __init__(
+        self, 
+        labels=("face", "house"), 
+        split="parity", 
+        fisher_z=True
+    ):
+        self.labels = labels
```
bthirion, line 77: I think we should document those arguments

examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,195 @@
...
+        self.split = split
+        self.fisher_z = fisher_z
+
+    def _fisher_z(self, r, eps=1e-12):
```
bthirion: these methods don't depend on self, hence could be defined as functions outside of the class.

examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,195 @@
...
+        else:
+            raise ValueError("Only split='parity' implemented here.")
+
+        def mean_pattern(lbl, mask):
+            sel = (y == lbl) & mask
+            if not np.any(sel):
+                return None
+            return X[sel].mean(axis=0)
+
+        A1 = mean_pattern(a, g1)
```
bthirion:
```suggestion
        a1 = mean_pattern(a, g1)
```

bthirion: I would not introduce capital letters here.

examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,195 @@
...
+    scores_img,
+    bg_img=mean_fmri,
+    title="Searchlight scores (face vs house)",
+    display_mode="z",
+    cut_coords=[-9],
+    vmin=0.0,
+    cmap="inferno",
+    black_bg=True,
+    colorbar=True,
+)
```
bthirion, line 210: the figure looks a bit weird...

bthirion (commented):
Thx ! This is a nice addition.

mtorabi59:
Thanks @bthirion , I will work on your comments. I agree, the figure looks a little bit weird, not sure why.

mtorabi59:
@bthirion the weird figure was because I changed the radius of spheres to 3. I changed it back to 5.6, same as the previous search light tutorial and now the figure looks fine.

examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,226 @@
...
+    This is a more stable and efficient way to compute correlation than
+    np.corrcoef for 1D patterns.
+    """
+    a = a - a.mean()
+    b = b - b.mean()
+    denom = (np.linalg.norm(a) * np.linalg.norm(b)) + 1e-12
+    return float(np.dot(a, b) / denom)
+
+
+class CorrelationMVPA(BaseEstimator):
```
bthirion, line 63: It's a bit annoying to have a class in an example, but I agree that here it's on purpose.

examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,226 @@
...
+        if any(v is None for v in (a1, a2, b1, b2)):
+            self.score_ = float("nan")
+            return self
+
+        r_aa = pattern_corr(a1, a2)
+        r_bb = pattern_corr(b1, b2)
+        r_ab = pattern_corr(a1, b2)
+        r_ba = pattern_corr(b1, a2)
+
+        if self.fisher_z_:
```
bthirion: Since it's one-shot thing, I would get rid of the `fisher_z` variable at apply that systematically.

examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,226 @@
...
+        computing the final score.
+    """
+
+    nilearn_searchlight_uses_cv = False
+
+    def __init__(
+        self, labels=("face", "house"), split="parity", fisher_z=True
+    ):
+        self.labels = labels
+        self.split = split
```
bthirion: for split, I think it's OK to hard-code the logic, and consider it's always `parity`

examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,226 @@
...
+    scores_img,
+    bg_img=mean_fmri,
+    title="Searchlight scores (face vs house)",
+    display_mode="z",
+    cut_coords=[-9],
+    vmin=0.0,
+    cmap="inferno",
+    black_bg=True,
+    colorbar=True,
+)
```
bthirion, line 210: We could improve the way it looks by thresholding the map, to be able to see the overlay.
Maybe also restrict the computation to a brain mask ?

bthirion (commented):
Looks mostly good. Just added a couple of suggestions.

nilearn/decoding/searchlight.py
```diff
@@ -30,6 +31,69 @@
...
+    # Must look like a sklearn estimator
+    if not hasattr(estimator, "fit"):
+        raise TypeError("SearchLight estimator must implement a 'fit' method.")
+    # Must be cloneable if uses_cv is True, because we clone it for each fold
+    # in cross_val_score
+    if uses_cv and not hasattr(estimator, "get_params"):
+        raise TypeError(
+            "SearchLight estimator must be cloneable (implement get_params). "
+            "Tip: inherit from sklearn.base.BaseEstimator."
+        )
```
Remi-Gau: I may be wrong but I think that those 3 can be replaced by 

```
if isinstance(estimator, BaseEstimator):
   raise TypeError("estimator must derive from sklearn.base.BaseEstimator.")
```

examples/07_advanced/plot_haxby_mvpa.py
mtorabi59: I added a threshold and it looks cleaner now. Not sure what you mean by brain mask and if that is something necessary.

examples/07_advanced/plot_haxby_mvpa.py
```diff
@@ -0,0 +1,212 @@
...
+labels = pd.read_csv(haxby_dataset.session_target[0], sep=" ")
+y = labels["labels"]
+run = labels["chunks"]
+
+# %%
+# Define our own MVPA estimator for use in SearchLight
+from sklearn.base import BaseEstimator
+
+
+def fisher_z(r, eps=1e-12):
```
bthirion, line 44: Note that a similar function  is used in `examples/03_connectivity/plot_seed_to_voxel_correlation.py`
Maybe we could add the function somewhere in `_utils`

bthirion (commented):
I'm happy with the PR. Just trying to factor code a liitle bit more.

examples/07_advanced/plot_haxby_mvpa.py
mtorabi59: @bthirion would it be okay to call a function from `_utils`, which is supposedly private, in an example?

mtorabi59: maybe we could add it to `nilearn.connectome` alongside `cov_to_corr`/`prec_to_partial` or in `nilearn.utils`?

Remi-Gau: yeah it would be better to put it somewhere public

Remi-Gau: it's so generic I would add a module in nilearn.utils

maybe nilearn.utils.stats ?

examples/07_advanced/plot_haxby_mvpa.py
bthirion: A few active voxels seem to be on the edge of the brain, they would be removed if we mask out the corresponding brain regions. Only interesting if there is a relevant mask in haxby or if the default epi mask computation works well on these data.

bthirion (commented):
LGTM overall, thx !

examples/07_advanced/plot_haxby_mvpa.py
Remi-Gau: I created a stricter mask to remove those.

bthirion (approved):
LGTM, thx

Remi-Gau (approved):

PR #6075, opened 2026-02-19
[FIX] set default butterworth padtype to constant 

### LINKED ISSUE #5415: lowpass butterworth and cosine

bpinsard:
- Closes #5415 

- set default butterworth padtype to constant  to avoid border effect with extrapolation

### DIFF
--- a/nilearn/signal.py
+++ b/nilearn/signal.py
@@ -310,7 +310,7 @@ def butterworth(
     low_pass=None,
     high_pass=None,
     order=5,
-    padtype="odd",
+    padtype="constant",
     padlen=None,
     copy=False,
 ):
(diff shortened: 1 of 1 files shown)


bthirion:
Is there a test to enforce better behavior of the filter ?

PR #6076, opened 2026-02-19
[DOC] Describe coef, intercept aggregation and predict in Decoder

### LINKED ISSUE #6049: [DOC] Update understanding decoder example
> Regarding https://nilearn.github.io/stable/auto_examples/02_decoding/plot_haxby_understand_decoder.html:
> 2. The ensemble model part is not really mentioned, maybe we should add some text at the end explaining the final `Decoder` model and which parameters are used in `Decoder.predict` 

 _Originally posted by @michellewang in [#2883](https://github.com/nilearn/nilearn/issues/2883#issuecomment-3923138639)_

man-shu:
- Closes #6049

- leave out a test set to later run Decoder.predict
- show how aggregation of coef and intercepts happens
- also show how the predictions are made

### DIFF
--- a/examples/02_decoding/plot_haxby_understand_decoder.py
+++ b/examples/02_decoding/plot_haxby_understand_decoder.py
@@ -414,12 +414,11 @@ show()
 
 # %%
 # .. note::
-#
-# The coefficients and intercepts from the Scikit-Learn pipeline and the
-# Nilearn decoder are actually similar but not identical. The differences
-# seem to depend on OS -- they are bigger on Linux than on Mac. However they
-# are not big enough to cause a difference in the predicted labels on the test
-# set.
+#   The coefficients and intercepts from the Scikit-Learn pipeline and the
+#   Nilearn decoder are not identical. We're unsure about the exact reason
+#   for this but the differences seem to depend on OS -- they are bigger on
+#   Linux than on Mac. However they are not big enough to cause a difference
+#   in the predicted labels on the test set.
 
 # %%
 # Compare the predicted labels on the left-out test set


man-shu:
Here's the rendered example: https://output.circle-artifacts.com/output/job/6e236717-2c9f-4739-9456-439a776cbef5/artifacts/0/dev/auto_examples/02_decoding/plot_haxby_understand_decoder.html

The coefs and intercepts turn out to be different when run on doc build, but locally they were close...

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -322,3 +347,80 @@
...
+# compare the values of the coefficients and intercepts from both approaches
+
+print(
+    "Coefficients are close:",
+    np.allclose(coefs_nilearn, av_sklearn_coef, atol=1e-2),
+)
+print(
+    "Intercepts are close:",
+    np.allclose(intercepts_nilearn, av_sklearn_intercept, atol=1e-2),
+)
```
bthirion, line 412: the output of his is `False`...

man-shu: Yes I noted that in my previous comment. 

Wondering whether we could just plot the coefs and say that "the numerical values are close but not identical". 

The next cell anyway shows that the predictions are identical. WDYT?

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -234,6 +234,25 @@
...
+# Leave out a test set for final evaluation
+# -----------------------------------------
+#
+# Before we train and cross-validate, let's leave out one run as a final test
+# set to evaluate the performance of the trained decoder on unseen data.
+
+test_run = 6
+test_mask = run == test_run
+fmri_img_test = index_img(fmri_img, test_mask)
+X_test = masker.fit_transform(fmri_img_test)
```
michellewang: Should the masker not be fit to the train data first?

```python
X = masker.fit_transform(fmri_img)
X_test = masker.transform(fmri_img_test)
```

with the `X_test` transformation done in the "Compare the predicted labels on the left-out test set" section

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -322,3 +347,80 @@
...
+# intercepts to predict the labels on new data via its ``predict``
+# method. So if we also compare the predicted labels from the decoder and
+# the Scikit-Learn pipeline on the left-out test set, we should see that they
+# are also identical.
+
+from sklearn.utils.extmath import safe_sparse_dot
+
+pred_nilearn = decoder.predict(X_test)
+decision_function_sklearn = (
+    safe_sparse_dot(X_test, av_sklearn_coef.T, dense_output=True)
```
michellewang, line 440: We should apply `feature_selector` to `X_test` here, no? I get that `screening_percentile` is 100 but I assume the reason we're showing it in the example in the first place is to show how one would do this with actual feature selection?

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -322,3 +347,80 @@
...
+#
+# So we can aggregate the coefficients and intercepts from the Scikit-Learn
+# pipeline by taking their mean across CV splits for each class vs. rest
+# problem to check if they are comparable to the coefficients and intercepts
+# from the Nilearn decoder.
+
+increment = len(np.unique(run))
+start_index_of_last_cv_split = (
+    increment * label_binarizer.classes_.shape[0] - increment
+) + 1
```
michellewang: I find this a bit cryptic. Would it be better to just use `len(coefs_sklearn)` as the `stop` arg for `range` below?

Maybe with some comments: `coefs_sklearn` has length `n_classes` * `n_runs`, with the the first `n_runs` being for class 0, etc.

examples/02_decoding/plot_haxby_understand_decoder.py
michellewang: I agree it's confusing if we show the print statements printing `False`. Not sure what is making the Nilearn vs Sklearn results different though.

Plotting is not a bad idea but since there are >400 coefs it's not that easy to tell visually.

man-shu: I tried a [`plot_img_comparison`](https://nilearn.github.io/stable/modules/generated/nilearn.plotting.plot_img_comparison.html#nilearn.plotting.plot_img_comparison) between the two. Let's see how that looks on the build.

examples/02_decoding/plot_haxby_understand_decoder.py
```diff
@@ -322,3 +345,99 @@
...
+)
+
+plot_img_comparison(
+    decoder.coef_img_["bottle"],
+    masker.inverse_transform(av_sklearn_coef[0]),
+    decoder.masker_,
+    ref_label="Nilearn",
+    src_label="Scikit-Learn",
+)
+
```
Remi-Gau, line 414:
```suggestion

show()

```

otherwise we see nothing when running as a script

note: may require an import statement

man-shu:
here's the doc: https://output.circle-artifacts.com/output/job/1c6c5e93-21a7-4f19-8911-54618d1c0f22/artifacts/0/dev/auto_examples/02_decoding/plot_haxby_understand_decoder.html#compare-the-coefficients-and-intercepts

examples/02_decoding/plot_haxby_understand_decoder.py
michellewang: Looks pretty convincing to me!

examples/02_decoding/plot_haxby_understand_decoder.py
man-shu: I agree that ideally feature selection should be done on the test set too. 

But in the example we want to show how the Decoder does things. Maybe I am missing something but it seems the `.predict()` method does not apply feature selection before predicting. See : 

https://github.com/nilearn/nilearn/blob/abbbe55a4b8fb540081488e9a142090817e8ed59/nilearn/decoding/decoder.py#L922

```python
    def predict(self, X):
        """Predict a label for all X vectors indexed by the first axis.

        Parameters
        ----------
        X : Niimg-like, :obj:`~nilearn.surface.SurfaceImage`, \
            :obj:`list` of Niimg-like objects \
            or :obj:`list` of :obj:`~nilearn.surface.SurfaceImage`, or \
            {array-like, sparse matrix}, shape = (n_samples, n_features)
            See :ref:`extracting_data`.
            Data on which prediction is to be made.

        Returns
        -------
        array, shape=(n_samples,) if n_classes == 2 else (n_samples, n_classes)
            Confidence scores per (sample, class) combination. In the binary
            case, confidence score for self.classes_[1] where >0 means this
            class would be predicted.
        """
        check_is_fitted(self)
        X = self._prep_input_post_fit(X)

        # Prediction for dummy estimator is different from others as there is
        # no fitted coefficient
        if isinstance(self.estimator_, (DummyClassifier, DummyRegressor)):
            n_samples = X.shape[0]
            scores = self._predict_dummy(n_samples)
        else:
            scores = self._decision_function(X)

        if is_classifier(self):
            if scores.ndim == 1:
                indices = (scores > 0).astype(int)
            else:
                indices = scores.argmax(axis=1)
            return self.classes_[indices]

        return scores

```

michellewang: I think the reason the Decoder doesn't explicitly do feature selection is that, when fitting the model, the coefficients get inverse transformed: https://github.com/nilearn/nilearn/blob/abbbe55a4b8fb540081488e9a142090817e8ed59/nilearn/decoding/decoder.py#L455-L457

So, even if feature selection initially reduced the number of features, the final coefficients are in the input feature space, and the feature selection is done implicitly at predict time through the coefficients. This also solves the issue of different numbers of features being selected in different CV folds.

I see that `inverse_transform` is also used in the example, but only at plotting time. Maybe instead we could apply it right after the model fit instead? In any case, since in this example all features are used, there should be no differences in the results.

man-shu: ah ok ok makes sense! Indeed I was missing something 😅

man-shu:
Here's the updated doc: https://output.circle-artifacts.com/output/job/818661a6-bb5c-4c83-ba06-2afb18be8b83/artifacts/0/dev/auto_examples/02_decoding/plot_haxby_understand_decoder.html

PR #6083, opened 2026-02-19
[MAINT] allow smoothing on surface for maskers and other 'estimators'

Remi-Gau:
- Closes none
- Removes warnings and code that would prevent from using surface smoothing for surface maskers and GLM

### DIFF
--- a/nilearn/tests/test_masking.py
+++ b/nilearn/tests/test_masking.py
@@ -375,16 +375,21 @@ def test_apply_mask(tmp_path, create_files, affine):
         assert_equal(proj.sum(), 9 / np.abs(affine[axis, axis]))
 
 
-@pytest.mark.parametrize("smoothing_fwhm", [None, 0, 1])
-def test_apply_mask_surface(surf_img_2d, surf_mask_1d, smoothing_fwhm):
-    """Test apply_mask on surface."""
-    length = 5
-    series = apply_mask(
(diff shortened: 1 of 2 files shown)


nilearn/tests/test_masking.py
```diff
@@ -375,15 +375,30 @@ def test_apply_mask(tmp_path, create_files, affine):
...
 
 
-def test_apply_mask_surface(surf_img_2d, surf_mask_1d):
+@pytest.mark.parametrize("smoothing_fwhm", [None, 0, 1])
+def test_apply_mask_surface(surf_img_2d, surf_mask_1d, smoothing_fwhm):
     """Test apply_mask on surface."""
     length = 5
-    series = apply_mask(surf_img_2d(length), surf_mask_1d)
+    series = apply_mask(
+        surf_img_2d(length), surf_mask_1d, smoothing_fwhm=smoothing_fwhm
```
bthirion: Souldn't we also test that setting smoothing to nonzero `does` something on the data ?

Remi-Gau: good point

Remi-Gau:
> LGTM but I would do a bit more than a smoke test ?

@bthirion 
AFAICT we do not have any test for GLMs that make sure that smoothing does actually something: I may be paranoid but maybe we should have those tests, no?

PR #6087, opened 2026-02-20
[MAINT] reduce number of FutureWarning thrown when checking estimators

Remi-Gau:
- Closes none but relates to #5942

- force to use the future new default for standardize when checking nilearn estimators to reduce the amount of warnings thrown


Remi-Gau:
should remove > 400 warnings from the tests logs

PR #6090, opened 2026-02-20
[DOC] document how to get a listing of the fixtures

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/6041#discussion_r2828449779

- explain how to get a listing of all our fixtures
- start improving doc string and type annotations of the fixtures

### DIFF
--- a/nilearn/_utils/data_gen.py
+++ b/nilearn/_utils/data_gen.py
@@ -387,8 +387,6 @@ def generate_fake_fmri(
     if n_blocks is None:
         return (Nifti1Image(fmri, affine), Nifti1Image(mask, affine))
 
-    assert n_blocks is not None
-
     flat_fmri = fmri[mask.astype(bool)]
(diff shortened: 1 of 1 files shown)


nilearn/_utils/data_gen.py
```diff
@@ -383,6 +387,8 @@ def generate_fake_fmri(
     if n_blocks is None:
         return (Nifti1Image(fmri, affine), Nifti1Image(mask, affine))
 
+    assert n_blocks is not None
```
pbarbarant: This assertions is always true due to the early return?

pbarbarant (approved):
LGTM, a minor comment below.

nilearn/_utils/data_gen.py
Remi-Gau: good catch: removing

PR #6096, opened 2026-02-22
[MAINT] apply sourcery refactoring suggestions

Remi-Gau:
- Closes none

- apply several fixes to avoid issues flagged on main by this workflow: https://github.com/nilearn/nilearn/actions/runs/22559202172/job/65342199667#step:4:98

### DIFF
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -450,12 +450,13 @@ def smooth_img(imgs, fwhm):
         imgs = [imgs]
 
     ret = []
-    for img in imgs:
-        if is_surface:
+    if is_surface:
+        for img in imgs:
             iterations = _mris_fwhm_to_niters(fwhm, img)
             ret.append(_smooth_surface_img(img, iterations))
 
-        else:
+    else:
+        for img in imgs:
             img = check_niimg(img)
             affine = img.affine
             filtered = smooth_array(


nilearn/decomposition/_base.py
```diff
@@ -540,7 +540,7 @@ def fit(self, imgs, y=None, confounds=None):
         elif any(isinstance(x, SurfaceImage) for x in imgs):
             masker_type = "multi_surface"
 
-        if masker_type in ["surface", "multi_surface"]:
+        if masker_type in {"surface", "multi_surface"}:
```
emdupre: A set versus a list feels a bit pedantic in this context ; I think it's still fine to adopt since the hit in readability is minimal, but I wonder if it's worth being less aggressive in flagging these kinds of checks.

nilearn/image/image.py
```diff
@@ -433,13 +433,12 @@ def smooth_img(imgs, fwhm):
         imgs = [imgs]
 
     ret = []
-    if is_surface:
-        for img in imgs:
+    for img in imgs:
+        if is_surface:
```
emdupre: I don't know that it makes sense (when tracing the logic) to invert these checks, since the boolean isn't defined per image.

nilearn/maskers/tests/test_masker_validation.py
```diff
@@ -170,23 +170,19 @@ def test_check_embedded_masker(mask, masker_type):
...
             "memory",
             "memory_level",
             "n_jobs",
             "verbose",
         ]:
-            # TODO (nilearn >= 0.15) if not needed anymore
-            # as this assertion should be true for all attributes
-            if param_key != "standardize":
-                assert getattr(masker, param_key) == getattr(mask, param_key)
-            # TODO (nilearn >= 0.15) remove elif
```
emdupre, line 183: We should carry over these deprecation notes

nilearn/reporting/get_clusters_table.py
```diff
@@ -385,18 +385,18 @@ def _get_clusters_table_surface(
     For other parameters, see `get_clusters_table`.
 
     """
-    cols = [
```
emdupre, line 388: I suppose the benefit here is that it's only defined in the clause where it's used ? In which case I might move it even closer to where it is used.

.sourcery.yaml
```diff
@@ -13,9 +13,7 @@
 version: '1'
 
 ignore: # A list of paths or files which Sourcery will ignore.
--   doc
 -   examples
--   maint_tools
```
emdupre, line 18: Hesitant to have it check the documentation, since the most pedagogical code isn't always the most concise. Is it running on the documentation already ?

nilearn/decomposition/_base.py
Remi-Gau: OK I had noticed that too but it did not irk me too much, but I agree with you: will try to revert this one and check if we can silence this kind of fix

.sourcery.yaml
Remi-Gau: I think that sourcery cannot fix python code in documentation file (as in .rst files), I think in this case it is more for python files in the doc folder like :

- doc/conf.py
- doc/conftest.py
- doc/get_ci_workflows_description.py
- doc/get_data_examples.py
- doc/make_atlas_table.py

But to your point, I think we did agree to avoid being too aggressive in the examples (we could also probably silence some of the ruff autofix from the examples).

nilearn/image/image.py
Remi-Gau: I need to double check that one.

nilearn/reporting/get_clusters_table.py
Remi-Gau: > the benefit here is that it's only defined in the clause where it's used

yup: it reduces cognitive load when reading the code

DimitriPapadopoulos:
These are mostly [Pylint](https://pylint.readthedocs.io/en/latest/) rules.

You could select ruff [Pylint (PL)](https://docs.astral.sh/ruff/rules/#pylint-pl) rules after this PR. I hadn't selected them because Pylint is very opinionated.

nilearn/image/image.py
Remi-Gau: will revert

nilearn/maskers/tests/test_masker_validation.py
Remi-Gau: pretty sure I forgot to do this before merging

# 🤦🏻

PR #6024, opened 2026-02-17
[FIX] allow local install with uv sync

mathdugre:
- Closes #6023

Changes proposed in this pull request:

We bump the version these packages:
- *numpy* `1.26.0`
- *pandas* `2.3.0`
- *scipy* `1.11.0`

note: Indicate the version bump of *scikit-learn* (from #5997) in the changelog.

PR #6108, opened 2026-03-03
[MAINT] use dependency groups to install dependencies only used in development

### LINKED ISSUE #6098: use dependency group in pyproject.toml to list dependencies that should not be installable from pypi
The suggestion comes scientific python: https://learn.scientific-python.org/development/guides/packaging-simple/#development-dependencies

The idea is that some dependencies are only used in development, but for now we still list them as optional dependencies like those used for plotting.

So currently people can run `pip install 'nilearn[dev]'` and install from pypi with all the dev dependencies the same way they would do `pip install 'nilearn[plotly]'`.

Scientific python recommends to make a distinction between dependencies that should only be installable when working on the local repo from those that are actually optional to use some functionalities of nilearn.

So one could then still do:

```
pip install 'nilearn[plotly]
```

or 

```
pip install -e `.[plotly]`
```

But would need to do the following to install all the doc and test dependencies:

```
pip install -e . --group dev
```

I am not 100% sure but I think some of the `uv sync` issues may be coming from this lack of distinction between optional and dev dependencies.

I can open a draft PR to see what this would look like

Remi-Gau:
- Closes #6098

- turn test, doc and dev into development dependencies

### DIFF
--- a/nilearn/glm/tests/test_first_level.py
+++ b/nilearn/glm/tests/test_first_level.py
@@ -1695,7 +1695,7 @@ def test_generate_report_default(kwargs):
         shapes=[(30, 31, 32, 33)], rk=3
     )
 
-    flm = FirstLevelModel(mask_img=mask).fit(
+    flm = FirstLevelModel(mask_img=mask, minimize_memory=False).fit(
         fmri_data[0], design_matrices=design_matrices[0]
     )
 
@@ -1723,7 +1723,7 @@ def test_generate_report_height_none_future_default():
         shapes=[(30, 31, 32, 33)], rk=3
     )
 
-    flm = FirstLevelModel(mask_img=mask).fit(
+    flm = FirstLevelModel(mask_img=mask, minimize_memory=False).fit(
         fmri_data[0], design_matrices=design_matrices[0]
     )
 
@@ -1752,7 +1752,7 @@ def test_generate_report_threshold_unused(threshold):
         shapes=[(30, 31, 32, 33)], rk=3
     )
 
-    flm = FirstLevelModel(mask_img=mask).fit(
+    flm = FirstLevelModel(mask_img=mask, minimize_memory=False).fit(
         fmri_data[0], design_matrices=design_matrices[0]
     )
 


Remi-Gau:
After this PR, the following will not work anymore:

```
➜  nilearn git:(maint/dependency_group) pip install -e '.[dev]'
Resolved 22 packages in 971ms
      Built nilearn @ file:///home/remi-gau/github/nilearn/nilearn
Prepared 1 package in 375ms
Uninstalled 1 package in 0.95ms
Installed 1 package in 1ms
 - nilearn==0.13.2.dev50+g78d79475f (from file:///home/remi-gau/github/nilearn/nilearn)
 + nilearn==0.13.2.dev51+g5f1eee449 (from file:///home/remi-gau/github/nilearn/nilearn)
warning: The package `nilearn @ file:///home/remi-gau/github/nilearn/nilearn` does not have an extra named `dev`
```

```
➜  nilearn git:(maint/dependency_group) pip install -e '.[test]'
Resolved 22 packages in 655ms
      Built nilearn @ file:///home/remi-gau/github/nilearn/nilearn
Prepared 1 package in 328ms
Uninstalled 1 package in 0.59ms
Installed 1 package in 1ms
 ~ nilearn==0.13.2.dev51+g5f1eee449 (from file:///home/remi-gau/github/nilearn/nilearn)
warning: The package `nilearn @ file:///home/remi-gau/github/nilearn/nilearn` does not have an extra named `test`
```

but instead the following will:

```
 pip install -e . --group dev
 pip install -e . --group test
```

note that once we have pushed a new release this will also applying when installing from pypi: currently one can do `pip install nilearn[test]` but this will not be possible after this PR and a new release

pyproject.toml
```diff
@@ -24,12 +81,12 @@ classifiers = [
...
-    "numpy>=1.22.4",
-    "pandas>=2.2.0",
+    "numpy>=1.26.0",
+    "pandas>=2.3.0",
     "requests>=2.30.0",
     "scikit-learn>=1.5.0",
     "jinja2>=3.1.2",
-    "scipy>=1.9.0",
+    "scipy>=1.11.0",
     "packaging"
```
Remi-Gau: - unrelated but we forgot to update those in: https://github.com/nilearn/nilearn/pull/6024

tox.ini
```diff
@@ -15,15 +16,6 @@ skip_install = false
 base_python = 3.10
 
 
-[plot_min]
-description = environment with minimum matplotlib, plotly and kaleido versions
-skip_install = false
-deps =
-    matplotlib==3.8.0
-    plotly==5.0.0
-    kaleido==0.2.1
```
Remi-Gau, line 24: removed and rely only on extra dependencies defined in pyproject.toml

pyproject.toml
```diff
@@ -78,49 +119,18 @@ min = [
     "scipy==1.11.0",
     "jinja2==3.1.2"
 ]
-min_plotting = ["nilearn[min]", "matplotlib==3.8.0", "plotly==5.0.0", "kaleido"]
-# the following is kept for "backward compatibility"
-plotly = ["nilearn[plotting]"]
```
Remi-Gau, line 83: now removed: maybe worth mentioning in the changelog

actually it is probably worth making an entry in the changelog anyway if we want to proceed with this

tox.ini
```diff
@@ -2,8 +2,9 @@
 [tox]
 requires =
     tox>=4
+    tox-uv
 envlist =
-  py3{10,11,12,13,14}-{plotting,flaky}
+    py3{10,11,12,13,14}-{plotting,flaky}
```
Remi-Gau, line 7: Note that just runnning `tox` does not run any test as I had expected but only creates virtual environments: to fix in a follow up PR

.github/workflows/test_with_tox.yml
```diff
@@ -91,7 +91,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: true
+            fail-fast: false
```
bthirion: Why that change ?

bthirion: Why that change ? Any comment ?

Remi-Gau: will revert: I was just checking something in CI

pyproject.toml
```diff
@@ -45,83 +118,22 @@ readme = "README.rst"
...
-    "pytest-csv",
-    "pytest-randomly",
-    "pytest-reporter-html1>=0.9.2,<0.9.4",
-    "pytest-mpl",
-    "pytest-rerunfailures",
-    "pytest-run-parallel",
-    "rich"
+    "plotly>=5.0.0,<6.1.0 ; platform_system == 'Windows'"
 ]
+rich = ["rich>=3.0.0"]
```
Remi-Gau: make it a dependency-group

pbarbarant (approved):

pyproject.toml
Remi-Gau: done

.github/workflows/testing_minimum.yml
```diff
@@ -30,6 +30,7 @@ on:
     # This enables manual triggering from GitHub UI
     # https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch
     workflow_dispatch:
+    pull_request:
```
```suggestion
```
remove before merging

Remi-Gau:
ok the test failure seems to be more windows specific and affects both python 3.11 and 3.14

for the sake of moving on I am tempted to mark as an xfail on windows for now...

pyproject.toml
```diff
@@ -2,6 +2,81 @@
...
+    "pytest-xdist[psutil]>=3.8"
+]
+# For running unit and docstring tests
+test_common = [
+    "numpydoc>=1.8.0",
+    "pytest>=6.0.0",
+    "pytest-csv>=3.0.0",
+    "pytest-randomly>=4.0.0",
+    "pytest-reporter-html1>=0.9.2,<0.9.4",
+    "pytest-mpl>=0.18.0",
```
Remi-Gau, line 69: pytest-mpl will install matplotlib so we should not have it when testing minimum dependencies

bthirion: You'll address that in this PR, right ?

Remi-Gau: - we already have this problem in main: see https://github.com/nilearn/nilearn/issues/6120

I played to around to try to fix it in this PR and it is more complicated than I thought so I would suggest doing it in a follow up PR dedicated to fix it

PR #6117, opened 2026-03-09
[ENH] ability to change crosshair color in brainsprite html viewer

### LINKED ISSUE #6123: add ability to change crosshair color in brainsprite html viewer
CURRENT:
`view = plotting.view_img(path, threshold=1.6, black_bg=False, radiological=True)`
image

DESIRED:
`view = plotting.view_img(path, threshold=1.6, black_bg=False, radiological=True, color_crosshair="#000000")`
image

- Relates to #6117

rohan3412:
- Closes #6123 

Changes: exposes the color_crosshair variable

CURRENT:
`view = plotting.view_img(path, threshold=1.6, black_bg=False, radiological=True)`
image

NEW:
`view = plotting.view_img(path, threshold=1.6, black_bg=False, radiological=True, color_crosshair="#000000")`
image

`view = plotting.view_img(path, threshold=1.6, black_bg=False, radiological=True, color_crosshair="#FF0000")`
image


bthirion:
Thx ! You probably want to add a test (at lest a smoke tets) for that feature. 
It could also be showcased in an example.  Otherwise, it will remain hidden.

Remi-Gau:
general thoughts regarding the general motivation for this:
- is it to improve visibility for both cases when we have a white and dark background?
- if so why is it an issue here and not for our other 'static' figures where we have crosshairs? am afraid that if we add it here then we are going to scope creep our way into doing it for all our plotting functions because of internal consistency?

though this suggestion highlights that we do hard code some of the ways we use brainsprite and that it may be harder for users to edit some of the things that brainsprite generates rather than a figure generated by matplotlib: so maybe we could think of passing a kwarg to view_img that may allow users to override some of our hard coded defaults...

emdupre:
+1 to @Remi-Gau's questions ; gentle reminder for the group to please open issues first to have these kinds of discussions outside of the PR.

bthirion:
> general thoughts regarding the general motivation for this:
> 
>     * is it to improve visibility for both cases when we have a white and dark background?
> 
>     * if so why is it an issue here and not for our other 'static' figures where we have crosshairs? am afraid that if we add it here then we are going to scope creep our way into doing it for all our plotting functions because of internal consistency?
> 
> 
> though this suggestion highlights that we do hard code some of the ways we use brainsprite and that it may be harder for users to edit some of the things that brainsprite generates rather than a figure generated by matplotlib: so maybe we could think of passing a kwarg to view_img that may allow users to override some of our hard coded defaults...

My view is that we should indeed have this feature in plot_*_img functions, but IIUC you consider that it's  a matplotlib-level feature and we should refrain from encapsulating too many features in Nilearn.

Remi-Gau:
> gentle reminder for the group to please open issues first to have these kinds of discussions outside of the PR.

I am going to [retcon](https://en.wikipedia.org/wiki/Retroactive_continuity) this and open an issue now to discuss things there: putting this PR on draft for now.

PR #6122, opened 2026-03-12
[FIX] Do not display colorbar if data is single valued for `plot_roi`

### LINKED ISSUE #6115: change of behavior in plot_roi
there seems to have been a change in the output of plot_roi between 0.12.1 and 0.13.0

## 0.12.1

https://nilearn.github.io/0.12.1/auto_examples/06_manipulating_images/plot_nifti_simple.html#simple-example-of-niftimasker-use

## 0.13.0

https://nilearn.github.io/0.13.0/auto_examples/06_manipulating_images/plot_nifti_simple.html#simple-example-of-niftimasker-use

I am not sure this was intentional.

Will try to bisect to see when that happened.

hndgzkn:
- Closes #6115

- Hide colorbar when data is single valued.

initial

---
TODO:
- [x]  Add baseline comparison test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,6 +86,8 @@ pattern.txt
 examples.txt
 gitlog.txt
 merge.txt
+__marimo__
+tmp.html
 
 # examples
 results/
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -34,13 +34,14 @@ NEW
 Fixes
 -----
 
-- :bdg-secondary:`Maint` Allow local installation with ```uv sync`` (:gh:`6024` by `Mathieu Dugré`_)
+- :bdg-secondary:`Maint` Allow local installation with ``uv sync`` (:gh:`6024` by `Mathieu Dugré`_)
 
 - :bdg-info:`Plotting` Fix ``nilearn.plotting.view_img`` resampling of non-isotropic images when no background image is used (:gh:`6031` by `Michelle Wang`_).
 
 Enhancements
 ------------
 
+- :bdg-success:`API` :class:`~nilearn.decoding.SearchLight` now has a ``random_state`` and ``estimator_args`` to pass to the underlying estimator (:gh:`6020` by `Rémi Gau`_).
 
 - :bdg-info:`Plotting` Allow string threshold in ``nilearn.plotting.plot_*`` functions (:gh:`5982` by `Saeed Babadi`_).
 
--- a/examples/02_decoding/plot_haxby_searchlight.py
+++ b/examples/02_decoding/plot_haxby_searchlight.py
@@ -75,7 +75,8 @@ n_jobs = 2
 # splitting the samples in 4 folds and make 4 runs using each fold as a test
 # set once and the others as learning sets
 #
-# The radius is the one of the Searchlight sphere that will scan the volume
+# The radius is the one of the Searchlight sphere that will scan the volume.
+#
 from sklearn.model_selection import KFold
 
 from nilearn.decoding import SearchLight
@@ -95,9 +96,8 @@ searchlight.fit(fmri_img, y)
 # %%
 # Visualization
 # -------------
-# %%
 # After fitting the searchlight, we can access the searchlight scores
-# as a NIfTI image using the `scores_img_` attribute.
+# as a NIfTI image using the ``scores_img_`` attribute.
 scores_img = searchlight.scores_img_
 
 # %%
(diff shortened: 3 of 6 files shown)


hndgzkn:
This change fixes the issue only for `plot_roi`. 

However `plot_stat_map` still displays the colorbar although data is single valued. 
Do we want the same for `plot_stat_map` as well?


stat

I am asking, because it would require some other changes, that will ignore the vmin, vmax values specified by the user.

nilearn/plotting/displays/_slicers.py
```diff
@@ -670,19 +676,23 @@ def _map_show(
...
                     data_bounds,
                     bounding_box,
                     type=type,
+                    norm=norm,
                     transparency=transparency_2d,
                     **kwargs,
                 )
                 ims.append(im)
+        kwargs["vmin"] = kwargs_vmin
+        kwargs["vmax"] = kwargs_vmax
```
hndgzkn: To be explicit, this part fixes the problem for `plot_stat_map` as well.

Remi-Gau:
@hndgzkn 
back merged main into this PR to hopefully avoid some tests failing in multi-threaded

bthirion:
> This change fixes the issue only for `plot_roi`.
> 
> However `plot_stat_map` still displays the colorbar although data is single valued. Do we want the same for `plot_stat_map` as well?
> stat
> 
> I am asking, because it would require some other changes, that will ignore the vmin, vmax values specified by the user.

I think that in 99% plot_stat_map will be used with images that contain multiple values. We should keep the colorbar.

nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -112,6 +112,14 @@ def test_plot_stat_map_display_mode(display_mode):
     )
 
 
+@pytest.mark.mpl_image_compare
+def test_plot_roi_single_value_data(img_mask_eye):
+    """Test `nilearn.plotting.image.img_plotting.plot_roi` to see that colorbar
+    does not appear in the plot when data displayed has single value.
+    """
+    return plot_roi(img_mask_eye, display_mode="y", cut_coords=3)
```
Remi-Gau: do you think we could get a bigger ROI plotted on the image because it is VERY easy to miss it on visual inspection ?

hndgzkn: I can change `img_mask_eye` to accept custom mask data. That way, I created some square regions more visible but probably not meaningful. Would that be ok?

Another option could be to fit a masker with `load_sample_motor_activation_image`. However I did not want to introduce dependency to maskers.

Remi-Gau: > That way, I created some square regions more visible but probably not meaningful.

yup though not sure that actually changing the img_mask_eye fixture is not over-engineering: will let you decide. :-)

nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -112,6 +113,19 @@ def test_plot_stat_map_display_mode(display_mode):
...
+def test_plot_roi_single_value_data(affine_eye):
+    """Test `nilearn.plotting.image.img_plotting.plot_roi` to see that colorbar
+    does not appear in the plot when data displayed has single value.
+    """
+    mask = np.zeros((53, 63, 42), dtype=np.uint8)
+    mask[20:35, 25:40, 10:25] = 1
+
+    return plot_roi(
+        Nifti1Image(mask, affine_eye), display_mode="y", cut_coords=3
+    )
```
hndgzkn, line 126: @Remi-Gau Is it ok this way?

bthirion (approved):

nilearn/plotting/tests/test_baseline_comparisons.py
Remi-Gau: yes that's great!!!!

Remi-Gau:
after the fact: wonder if this deserves a changelog entry given this is a bug fix.
@hndgzkn WDYT?

hndgzkn:
> after the fact: wonder if this deserves a changelog entry given this is a bug fix. @hndgzkn WDYT?

I think we can add. Is it ok if I add to this PR and remerge it or shall I open a new PR?

Remi-Gau:
maybe open a new PR: should be a small one

hndgzkn:
> maybe open a new PR: should be a small one

[Here](https://github.com/nilearn/nilearn/pull/6138)

PR #6129, opened 2026-03-12
[FIX] use pytest.warns instead of pytest.deprecated_call()

Remi-Gau:
- Closes none but fixes a couple of failing tests on main

see https://github.com/nilearn/nilearn/actions/runs/23011290182/job/66822277511#step:8:13467

- use pytest.warns instead of pytest.deprecated_call()


Remi-Gau:
Will merge to normalize the situation on main

PR #6142, opened 2026-03-19
[FIX] Update min scikit-learn version to 1.7.1 to use HTML representation from scikit-learn

### LINKED ISSUE #6141: [BUG] [BUG] Scikit-learn 1.7.0 has a bug for generating HTML representation

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Python version

- [ ] 3.14
- [x] 3.13
- [ ] 3.12
- [ ] 3.11
- [ ] 3.10

### nilearn version

0.13.2.dev134+g356bd6b23

### scikit-learn version

1.7.0

### Expected behavior

`nilearn.maskers/tests/test_html_report.py::test_multi_nifti_masker_generate_report_mask` should pass.

This is what I got:

```bash
...

>       if param_value != init_default_params[param_name] and not (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            is_scalar_nan(init_default_params[param_name])
            and is_scalar_nan(param_value)
        ):
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

init_default_params = {'clean_args': None, 'cmap': 'gray', 'detrend': False, 'dtype': None, ...}
param_name = 'target_affine'
param_value = array([[1., 0., 0., 0.],
       [0., 1., 0., 0.],
       [0., 0., 1., 0.],
       [0., 0., 0., 1.]])



```

Run test from command line
```bash
pytest nilearn/maskers/tests/test_html_report.py::test_multi_nifti_masker_generate_report_mask

```

### Solution

The bug is fixed in scikit-learn version [1.7.1](https://scikit-learn.org/stable/whats_new/v1.7.html#version-1-7-1).

hndgzkn:
- Closes #6141 

- Update min scikit-learn version to 1.7.1 to generate HTML representation with scikit-learn

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,10 @@ min = [
     "numpy==1.26.0",
     "pandas==2.3.0",
     "requests==2.30.0",
-    "scikit-learn==1.5.0",
+    # scikit-learn version 1.7.0 has a bug that prevents _repr_html_ to run
+    # correctly
+    # https://scikit-learn.org/stable/whats_new/v1.7.html#sklearn-base
+    "scikit-learn==1.5.0,!=1.7.0",
     "scipy==1.11.1",
     "jinja2==3.1.2"
 ]


pyproject.toml
```diff
@@ -103,7 +103,7 @@ dependencies = [
     "numpy>=1.26.0",
     "pandas>=2.3.0",
     "requests>=2.30.0",
-    "scikit-learn>=1.5.0",
+    "scikit-learn>=1.5.0,!=1.7.0",
```
Remi-Gau, line 112: do you think we should modify the `min` dependency gorup in the pyproject.toml to also exclude 1.7.0?

https://github.com/nilearn/nilearn/blob/d135e0d7e2ba975aac7f731315b8fadb21b8801c/pyproject.toml#L34

my reasoning is that in the future we may bump the minimum version and forget to avoid 1.7.0 at that point

also do you think we should add a comment / link to the issue as an explanation why we skip 1.7.0 to avoid having to git blame the change?

hndgzkn: I added a comment.

But actually doing the changes I thought, the code still runs with version 1.7.0. The only problem is that while generating reports, instead of using _repr_html starting from scikit version 1.7.0, we need to call it starting from 1.7.1. For versions smaller than 1.7.1, we already have our implementation to get html representation of parameters.

I couldn't be sure to exclude a version just for that. (On the otherhand there are some other bug fixes in version 1.7.1)

hndgzkn: > do you think we should modify the `min` dependency gorup in the pyproject.toml to also exclude 1.7.0?

if we are going to keep this solution, I think it is better to do that.

Remi-Gau: OK let's do it then! 👍🏻

hndgzkn: Done.

PR #6157, opened 2026-03-27
[FIX] Do not copy mask img's header in masking.unmask

tsalo:
- Closes #6150.

- Update `test_unmask_3d_with_files` to check that the datatype of the unmasked image created by `masking.unmask` is the same as the datatype of the array to be unmasked (not the datatype of the mask image). I have verified that this test fails with the code from before my fix.
- Include `copy_header=False` in the `new_img_like` call within `masking.unmask`.

### DIFF
--- a/nilearn/tests/test_masking.py
+++ b/nilearn/tests/test_masking.py
@@ -561,6 +561,10 @@ def test_unmask_3d_with_files(
     assert t[0].flags["F_CONTIGUOUS"]
     assert_array_equal(t[0], unmasked3D)
 
+    # Check that the unmasked image retains the datatype of the data array
+    t = unmask([masked3D], filename, order="F")
+    assert t[0].get_data_dtype() == data3D.dtype
+
 
 def test_unmask_errors(rng, affine_eye, shape_3d_default):
     """Test unmask errors."""
--- a/nilearn/tests/test_masking.py
+++ b/nilearn/tests/test_masking.py
@@ -561,11 +561,20 @@ def test_unmask_3d_with_files(
     assert t[0].flags["F_CONTIGUOUS"]
     assert_array_equal(t[0], unmasked3D)
 
-    # Check that the unmasked image retains the datatype of the data array
-    #
-    # see https://github.com/nilearn/nilearn/issues/6150
-    #
-    t = unmask([masked3D], filename, order="F")
+
+def test_unmask_retain_datatype(rng, affine_eye, shape_3d_default):
+    """Check that the unmasked image retains the datatype of the data array.
+
+    see https://github.com/nilearn/nilearn/issues/6150
+    """
+    data3D = rng.uniform(size=shape_3d_default)
+    mask = rng.integers(2, size=shape_3d_default, dtype="int32")
+    mask_img = Nifti1Image(mask, affine_eye)
+
+    mask = mask.astype(bool)
+    masked3D = data3D[mask]
+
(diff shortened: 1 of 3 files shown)


bthirion (approved):
Thx. Could deserve a whatsnew entry.

tsalo:
@bthirion sorry I had forgotten about that. I just added a whatsnew entry.

bthirion (approved):

nilearn/tests/test_masking.py
```diff
@@ -561,6 +561,10 @@ def test_unmask_3d_with_files(
     assert t[0].flags["F_CONTIGUOUS"]
     assert_array_equal(t[0], unmasked3D)
 
+    # Check that the unmasked image retains the datatype of the data array
```
Remi-Gau:
```suggestion
    # Check that the unmasked image retains the datatype of the data array
    #
    # see https://github.com/nilearn/nilearn/issues/6150
    #
```
Adding a link to the issue so that we avoid deleting that line in the future

Remi-Gau: actually I am going to create a separate 'pure' regression test for this

Remi-Gau:
thanks @tsalo for this fix

tsalo:
@Remi-Gau @bthirion is there any chance you could make a 0.13.2 release fairly soon? I made backwards-incompatible changes to downstream repos (esp. XCP-D) to support 0.13.x but this bug affects both releases in the series, so I'd have to switch back to supporting 0.12.1 until the next nilearn release to work around the bug.

Remi-Gau:
depends what do you call 'fairly soon'?

ideally I would like to get the next one out before OHBM (so early June): is that early enough?

tsalo:
My last three XCP-D versions are not going to be usable since they all pin 0.13.1, so I was hoping you'd make a patch release some time this week. If early June is when you plan to make the release, I can make XCP-D work with 0.12.1 instead.

I also just realized that NiftiMasker uses `masking.unmask` in `inverse_transform`, so should I open a PR to address that as well?

Remi-Gau:
@tsalo 
have a look at this old PR of mine: I think it tried to adress the problem you mention for all our estimators

- https://github.com/nilearn/nilearn/pull/5511

also suggested by @emdupre : would you be OK with a rc release ?

tsalo:
An RC release would be great, thank you.

I realized that fixing unmask fixed NiftiMasker, so you can ignore my last comment about a followup PR.

I took a look at your PR. AFAICT, the image that is returned by `new_img_like` will still have the mask's datatype, but that shouldn't impact the data until the image is written to a file, so changing the datatype after the image is created, but before the user writes it out to a file, should work.

Remi-Gau:
> An RC release would be great, thank you.
> 
> I realized that fixing unmask fixed NiftiMasker, so you can ignore my last comment about a followup PR.
> 
> I took a look at your PR. AFAICT, the image that is returned by `new_img_like` will still have the mask's datatype, but that shouldn't impact the data until the image is written to a file, so changing the datatype after the image is created, but before the user writes it out to a file, should work.

@tsalo 
I pushed 0.14.0rc0 to pypi:
- https://github.com/nilearn/nilearn/releases/tag/0.14.0rc0
- https://pypi.org/project/nilearn/0.14.0rc0/

You should be able to get it with `pip install --prerelease nilearn`.

Let me know if it does not work or fix your issue or if some new bugs were introduced.

tsalo:
Thank you so much!

PR #6172, opened 2026-04-09
[ENH] Refactor reporting glm to reuse `nilearn.reporting.mixin.ReportMixin`

hndgzkn:
- Closes None but relates to #5828

- Create `nilearn.glm.GLMReportMixin` that implements `nilearn.reporting.mixin.ReportMixin` and put common report generation functionality inside 
- Move `generate_report` from `nilearn.glm._base` to `nilearn.glm.mixin.GLMReportMixin`
- Fix some tests failing when matplotlib is not installed
- Move `summary` and `description` in `ReportMixin` to `MaskerReportMixin` as they seem to be processed only in masker reports


nilearn/glm/_reporting_utils.py
```diff
@@ -21,9 +23,364 @@
...
+
+                    Dict of :term:`contrast` names and coefficients,
+                    or list of :term:`contrast` names
+                    or list of :term:`contrast` coefficients
+                    or :term:`contrast` name
+                    or :term:`contrast` coefficient
+
+                    Each :term:`contrast` name must be a string.
+                    Each :term:`contrast` coefficient must be a list
+                    or numpy array of ints.
```
bthirion: I think a contrast can have non-int values

bthirion (commented):
LGTM overall.

nilearn/glm/_reporting_utils.py
hndgzkn: You mean float? 
This part is just moved, I haven't made any changes. If you confirm the correct type, I can change accordingly.

nilearn/glm/_reporting_utils.py
```diff
@@ -21,9 +23,364 @@
...
+        threshold: float | int | np.floating | np.integer | None = None,
+        alpha=0.001,
+        cluster_threshold: ClusterThreshold = 0,
+        height_control: HeightControl = "fpr",
+        two_sided: bool = False,
+        min_distance: float | int | np.floating | np.integer = 8.0,
+        plot_type: Literal["slice", "glass"] = "slice",
+        cut_coords=None,
+        display_mode=None,
+        report_dims=(1600, 800),
```
hndgzkn, line 142: I haven't changed these to kwargs to comply with the signature of the method defined in `nilearn.reporting.mixin.ReportMixin`. Below is how I did it for maskers in `nilearn.maskers._mixin.MaskersReportMixin`. May be it would be better to do it the same. 

```
    def generate_report(
        self, engine="matplotlib", title: str | None = None, **kwargs
    ) -> HTMLReport:
        """Generate an HTML report for this masker.


....

       engine : :obj:`str`, default="matplotlib"
            Choice of engine to display report figures.

            All maskers support ``"matplotlib"`` as engine.

            Other options are :

            - ``"brainsprite"`` :

              NiftiMasker, MultiNiftiMasker, NiftiLabelsMasker,
              MultiNiftiLabelsMasker:

            - ``"plotly"``:

              SurfaceMasker, MultiSurfaceMasker, SurfaceLabelsMasker,
              MultiSurfaceLabelsMasker, SurfaceMapsMasker,
              MultiSurfaceMapsMasker:

        title : :obj:`str` or None, default=None
            title for the report. If None, title will be the class name.

        kwargs : :obj:`dict` [ :obj:`str` , Any]
            Dictionary of key-word arguments necessary for report generation.

            Expected keys depending on masker type are:

            - For NiftiMapsMasker, MultiNiftiMapsMasker, SurfaceMapsMasker,
              MultiSurfaceMapsMasker :

                  %(displayed_maps)s

            - For NiftiSpheresMasker :

                  %(displayed_spheres)s
```

nilearn/glm/_reporting_utils.py
```diff
@@ -21,9 +23,364 @@
...
+
+    def _set_report_basics(self, engine, title):
+        super()._set_report_basics(engine, title)
+
+        report_content = self._report_content
+
+        title = f"<br>{title}" if title else ""
+        report_content["page_title"] = (
+            f"Statistical Report - {self.__str__()}{title}"
+        )
```
hndgzkn, line 57: Is this difference in title really necessary? Below is how it is done in base class:

```
        # Set title for report
        report_content["title"] = (
            title if title is not None else self.__class__.__name__
        )
        report_content["page_title"] = report_content["title"]
```

nilearn/glm/_reporting_utils.py
hndgzkn: Finally I decided not to do that way. Because it has implications on `nilearn.glm.io.save_glm_to_bids`.

nilearn/glm/_reporting_utils.py
bthirion: Yes, we should open the possibility to have floats

nilearn/glm/_reporting_utils.py
hndgzkn: I wanted to compare masker and GLM reports visually. GLM reports seem more structured. 

I thought may be we could define (in another PR) some common sections and structure for the reports, and each report can have diverging parts.

At least `title`, `description`, `model details` and `about` sections seem to be common but structured differently:

glm


maskers

nilearn/glm/_reporting_utils.py
```diff
@@ -21,9 +23,344 @@
...
+        check_generate_report_input(
+            height_control, cluster_threshold, min_distance, plot_type
+        )
+
+        if not hasattr(self, "_reporting_data"):
+            self._reporting_data: dict[str, Any] = {
+                "trial_types": [],
+                "noise_model": getattr(self, "noise_model", None),
+                "hrf_model": getattr(self, "hrf_model", None),
+                "drift_model": None,
```
Remi-Gau, line 84: it is slightly strange to have those fort both first and second level GLM: AFAICT these should only exist for first level GLM

nilearn/glm/_reporting_utils.py
Remi-Gau: > I thought may be we could define (in another PR) some common sections and structure for the reports, and each report can have diverging parts.

agreed

Remi-Gau:
@hndgzkn 
how do you feel about delaying the merge of this till we have visual test for HTML outputs - including reports (https://github.com/nilearn/nilearn/pull/6179)?

nilearn/glm/_reporting_utils.py
hndgzkn: It is the case in the original code. I can move this part to only run for first level. 

Alternatively, we can keep this PR as only refactoring and make changes in a follow-up PR. I think that would be better to identify the impact of changes.

hndgzkn:
> @hndgzkn how do you feel about delaying the merge of this till we have visual test for HTML outputs - including reports (#6179)?

I think it would be more safe, better to wait.

bthirion:
LMK if this is ready for a round of reviews.

hndgzkn:
> LMK if this is ready for a round of reviews.

I think it is ready. 

I do not want to make further changes. We are waiting the merge of #6179 so that we can see if this PR made any changes of the reports. This PR is just a refactoring, so we do not expect any change in reports.

Remi-Gau:
@hndgzkn can I let resolve those conflicts?

hndgzkn:
> @hndgzkn can I let resolve those conflicts?

Done. Let's see if tests pass without failures :)

Remi-Gau:
> Let's see if tests pass without failures

yeah, I think that if the 'visual comparison' tests become good enough, we may be able to rely more on them than on checking the HTML content

hndgzkn:
@Remi-Gau HTML tests fail however a part from report heading, there is no difference. It is mainly due to very minor shifts in text. Shall I increase tolerance or submit new images so that the difference that this PR introduces is more clear, not to doubt about it later?

Remi-Gau:
> submit new images so that the difference that this PR introduces is more clear

yeah let's update the images: this way you can also check it is easy enough to do or if more documentation is needed

nilearn/glm/_reporting_utils.py
```diff
@@ -21,9 +23,344 @@
...
+                    It should be greater than the minimum intensity
+                    of the input data.
+                    All intensities greater than the specified threshold
+                    will be set to zero.
+                    All other intensities keep their original values.
+
+                  - If the threshold is positive:
+
+                    It should be less than the maximum intensity
+                    of the input data.
```
bthirion, line 234: Why ? It's OK if it's higher -> everything gets set to 0.

bthirion (commented):
LGTM, thx !

nilearn/glm/_reporting_utils.py
hndgzkn: I have not changed this part in this PR.

However, I've added this part in a previous PR in accordance with the docstring in `check_threshold`:

https://github.com/nilearn/nilearn/blob/d3c35a140d32870fa42f9f05f4311a723d80fb29/nilearn/_utils/param_validation.py#L14-L35

If that is problematic, I suggest to make necessary changes in a follow-up PR.

Remi-Gau:
OK I will wait for CI to finish and then merge this

Remi-Gau (approved):

PR #6173, opened 2026-04-09
[ENH] add type annotation and further checks to math_img

### LINKED ISSUE #6170: Add type hints for formula and copy_header_from in math_img to clarify API
Adding type hints to nilearn.image.math_img would help clarify the API and improve the developer experience in editors and tools that make use of static type information.

In particular, the parameter copy_header_from may be ambiguous. From the name, it could be interpreted as accepting a Nifti image or a path, while the docstring specifies that it should be the name of one of the image variables used in the formula. Explicit type hints would make this clearer.

For example, the signature could specify:

formula: str
copy_header_from: str | None = None

This would make the intended usage more explicit and reduce potential confusion for users.

Additionally, when a value other than one of the image variable names is passed to copy_header_from, a KeyError is currently raised. Depending on the situation, a TypeError (if a non-string value is passed) or a ValueError (if a string that is not one of the variable names is passed) might be more appropriate and could provide clearer feedback to users.

If this sounds reasonable, I would be happy to open a PR adding these annotations.

Remi-Gau:
- Closes #6170 

- add type annotation to math_img


nilearn/image/image.py
```diff
@@ -1510,7 +1510,7 @@ def _apply_threshold(img_data, two_sided, cutoff_threshold):
     return img_data
 
 
-def math_img(formula, copy_header_from=None, **imgs):
+def math_img(formula: str, copy_header_from: str | None = None, **imgs):
```
Remi-Gau: maybe throw a warning if some value is passed by copy_header_from but we only work with surface images: it cannot be used in this case

nilearn/image/image.py
```diff
@@ -1522,13 +1522,26 @@ def math_img(formula, copy_header_from=None, **imgs):
...
         If None, the default
         :class:`~nibabel.nifti1.Nifti1Header` is used.
 
+        .. note:
+
+            It is technically possible to pass an image
+            to copy the header from,
+            but that is unused in the formula.
+
+            .. code-block:: python
```
Remi-Gau, line 1563: Is this intentional or is it a bug?

nilearn/image/image.py
```diff
@@ -1522,13 +1522,26 @@ def math_img(formula, copy_header_from=None, **imgs):
         numpy imported as 'np'.
 
     copy_header_from : :obj:`str` or None, default=None
-        Takes the variable name of one of the images in the formula.
```
Remi-Gau, line 1525: Rephrase rather then remove: takes the variable name of one of the images in imgs

nilearn/image/image.py
Remi-Gau: Add return type.
Also possibly add type for imgs

nilearn/image/tests/test_image.py
```diff
@@ -1492,11 +1492,24 @@ def test_math_img_exceptions(affine_eye, img_4d_ones_eye, surf_img_2d):
         math_img(formula, img1=img1, img3=img3, copy_header_from="img1")
 
     # Passing an 'img*' variable (to copy_header_from) that is not in the
-    # formula or an img* argument should raise a KeyError exception.
-    with pytest.raises(KeyError):
+    # formula or an img* argument should raise a ValueError exception.
+    with pytest.raises(
+        ValueError,
+        match=("copy_header_from='img2' but 'img2' is missing from 'imgs'"),
```
hndgzkn:
```suggestion
        match=("copy_header_from='img2' but 'img2' is missing in 'imgs'"),
```

nilearn/image/tests/test_image.py
```diff
@@ -1492,11 +1492,24 @@ def test_math_img_exceptions(affine_eye, img_4d_ones_eye, surf_img_2d):
...
         math_img(formula, img1=img1, img3=img3, copy_header_from="img2")
 
 
+def test_math_img_warnings(img_4d_ones_eye):
+    """Test warnings raised by math_img."""
+    img1 = img_4d_ones_eye
+    img3 = img_4d_ones_eye
+    with pytest.warns(
+        UserWarning, match=r"Some images .* are not mentioned in the formula"
+    ):
```
hndgzkn, line 1509: I haven't checked all tests. 

What is the expected behavior if the specified image is not in the formula, but is in `copy_header_from`. I would expect not to have warning as it is said to be possible in the docstring. In that case I wonder if there is a test case that guarantees that the warning is not raised.

nilearn/image/image.py
hndgzkn: Instead of the note below, can't it be said that the image name should either appear in formula or specified in imgs, and if it is None, takes one from formula.

nilearn/image/image.py
```diff
@@ -1594,8 +1607,18 @@ def math_img(formula, copy_header_from=None, **imgs):
     """
     is_surface = all(isinstance(x, SurfaceImage) for x in imgs.values())
 
+    img_missing_from_formula = [x for x in imgs if x not in formula]
+    if img_missing_from_formula:
```
hndgzkn, line 1635: If it is possible to set an image that does not appear in the formula as `copy_header_from`, I would add another condition here to check if that is not the case as well.

hndgzkn (approved):
LGTM apart form some comments and suggestions.

nilearn/image/tests/test_image.py
hndgzkn: `in` feels better to me, ignore it if you think `from` is ok.

nilearn/image/image.py
Remi-Gau: yeah I agree: we should raise the warning only when the image does not appear in imgs or copy_header_from

Remi-Gau: done

nilearn/image/tests/test_image.py
Remi-Gau: done

PR #6179, opened 2026-04-14
[MAINT] test html output

Remi-Gau:
- Closes #6163 

- add tests to make sure that there is no regression in HTML outputs (figure, reports...)


tox.ini
```diff
@@ -196,6 +196,20 @@ commands =
...
     pytest --numprocesses auto --report=report.html --csv results/pytest_output/pytest_output.csv  {posargs:}
 
+[testenv:test_html]
+description = Check HTML output.
+base_python = 3.10
+passenv = {[global_var]passenv}
+dependency_groups =
+    min_plotting
+allowlist_externals =
+    npm
```
Remi-Gau, line 207: update contributing doc to explain that node will be required to run those tests

tests/js/view_surf_reference.png
Remi-Gau, line 1: hmmm the plotly figures generated do not look convincing

.github/workflows/test_html.yml
Remi-Gau, line 1: tested on my main branch and it passes

Remi-Gau:
@hndgzkn 
I adapted the tests from brainsprite to check the output generated by the view_* functions and the reports

Only done for some function and one masker report for now but it could be extended.

Let me know what you think.

nilearn/glm/_base.py
```diff
@@ -747,6 +748,12 @@ def generate_report(
             else ""
         )
 
+        # We clean up the version number
+        # to make it more stable during tests.
+        version = str(
+            Version(__version__).__replace__(local=None, dev=None, pre=None)
+        )
+
```
hndgzkn: Instead of adding the same code in multiple places, I would suggest to add this part as `__short_version__` in `nilearn._version`.

tests/js/references/flm_bids_features_reference.png
hndgzkn, line 1: Would it be possible to scroll the file and save the following parts as well?

hndgzkn (commented):
I've put some minor comments, otherwise I think it seems to be very helpful.

Shall we do it for all types of maskers, fitted and not fitted, I am not sure?

nilearn/glm/_base.py
Remi-Gau: Yeah I was thinking of doing this.

tests/js/references/flm_bids_features_reference.png
Remi-Gau: Well I wanted to try something different which was to make the height of the report higher, but it seems we cannot resize reports.
It's a bug and I opened an issue about it.
Once it is fixed, we should indeed cover the whole height of the report.

Remi-Gau:
> Shall we do it for all types of maskers, fitted and not fitted, I am not sure?

I would start with the fitted ones. Once we have things fixed for the report height I would add some others.

tests/js/references/flm_bids_features_reference.png
hndgzkn: > It's a bug and I opened an issue about it. 

I can have a look at it if that's ok with you.

Remi-Gau: yes thanks: on that note probably worth making the `report_dims` of the generate_report of GLM a common parameters to all reports

I am not 100% sure but I suspect that making this resizing work will require some HTML and CSS tweaking: just a heads up

Remi-Gau: @hndgzkn I actually managed to change the size of the viewport in the generated figures so we should be getting the entire reports now

nilearn/_utils/testing.py
```diff
@@ -34,6 +32,8 @@ def with_memory_profiler(func):  # noqa: ARG001
         """Use as a decorator to skip tests requiring memory_profiler."""
 
         def dummy_func():
+            import pytest
```
Remi-Gau, line 35: moved the pytest import local to make it possible to import this module even when pytest is not installed

tests/js/masker_report.test.js
```diff
@@ -0,0 +1,16 @@
+const template = require('./template.js')
+
+const VIEWPORT = { x: 0, y: 0, width: 1200, height: 750 }
```
hndgzkn, line 3: May be height could be bigger for maskers as well, non? Some seem to be cut.

tests/js/references/flm_bids_features_reference.png
hndgzkn: Nice solution, the images are very good now.

tests/js/masker_report.test.js
Remi-Gau: yeah one of the report would then fails because if we display the whole report then there is file path displayed and obviously not the same when run locally and online so I as voluntarily keeping the viewport smaller.

Remi-Gau:
I am going to go ahead and merge this one as it is a bit of a blocker for many other things.

Issue #6184, opened 2026-04-15
Improve type hints for conditional return types (reduce IDE/type checker warnings)


mcsitter:
Hi all,

Since type annotations seem in principle welcome in nilearn, I’m opening this issue to highlight a recurring pattern in the codebase: function return types often depend on input arguments, but this is currently only documented in docstrings rather than expressed in type hints.

This leads to unnecessary type-checking warnings in downstream code and IDEs, for example when accessing attributes of figures returned by plotting functions.

## Case 1: `output_file` controls return type in plotting functions

In functions such as `plot_roi`, `plot_anat`, `plot_glass_brain`, etc.:

- when `output_file is None` → returns a display object (e.g. `OrthoSlicer`)
- when `output_file is str` → returns `None`

A typical overload would be:

```python
from typing import overload

@overload
def plot_roi(..., output_file: None = None) -> OrthoSlicer: ...

@overload
def plot_roi(..., output_file: str) -> None: ...
```

## Case 2: input type controls output type (SurfaceImage vs volumetric images)

In `new_img_like`:

- if `ref_niimg` is a `SurfaceImage` → returns `SurfaceImage`
- otherwise → returns a volumetric image type (typically a `FileBasedImage` subclass such as `Nifti1Image` or `Nifti2Image`)

This behavior is currently only expressed in runtime logic and docstrings.

## Question

Are there additional functions or groups of functions where this pattern is present?

I’m aware that typing image-related APIs is not always straightforward, especially since inputs can often be file paths as well as Python objects. That flexibility is of course very useful from a user perspective and should be preserved.

Still, making these conditional return types explicit would likely reduce a lot of unnecessary IDE and type-checker noise.

If helpful, I’d be happy to draft a PR implementing some of these changes.

Remi-Gau:
We are gradually adding type annotations in some places indeed.

I agree that priority should be given to clear returned types.

I would say let's start with the easy one you mentioned and deal with the returned types of plotting functions: feel free to open a draft PR

Remi-Gau:
@mcsitter do you want to try to add type annotation for one or more of the plotting functions?

mcsitter:
Should I keep this limited to adding overloads for the output_file-dependent return type (display object vs None), or would it also be in scope to introduce a few TypeAlias entries in nilearn.typing and start typing some of the other parameters as well?

Remi-Gau:
[...]

PR #6200, opened 2026-05-04
[MAINT] add some type annotation to plot_img

Remi-Gau:
- Relates to #6184

- add type annotation to plot_img


Remi-Gau:
@mcsitter is this what you had in mind?

mcsitter:
Yeah, that’s pretty much what I was going for. I tried adding types to more parameters and fixing a bunch of other mypy warnings at the same time, but it ended up getting out of hand. I couldn’t really split it into a clean PR or commits, and then I didn’t manage to get back to it afterward. Thank you!

nilearn/plotting/image/img_plotting.py
```diff
@@ -289,32 +299,86 @@ def _plot_img_with_bg(
     return save_figure_if_needed(display, output_file)
 
 
+@overload
+def plot_img(
```
hndgzkn, line 303: Is it really necessary to use overload when the function can return None instead of an OrthoSlicer instance?

To me, it seems reasonable to use overload when there are explicitly different return types. When it returns None as OrthoSlicer, that would be ok to just document it.

hndgzkn (approved):
LGTM with a minor comment.

Remi-Gau:
Note that adding type annotation does not make the function signature less readable in the doc: https://output.circle-artifacts.com/output/job/fcfeb541-3da7-47b5-8b4c-52d6f778196d/artifacts/0/dev/modules/generated/nilearn.plotting.plot_img.html#nilearn.plotting.plot_img

image

nilearn/plotting/image/img_plotting.py
Remi-Gau: if you do not use overload then IDEs will not be able to figure out what is the type of the output and you won't have the benefit that comes with that (tab completion, warning when using a non-sensical method).

see for example:

image

hndgzkn: Ok, I see, thanks.

nilearn/plotting/image/img_plotting.py
```diff
@@ -289,32 +299,86 @@ def _plot_img_with_bg(
...
+    resampling_interpolation=...,
+    bg_img=...,
+    vmin=...,
+    vmax=...,
+    radiological=...,
+    decimals=...,
+    cmap=...,
+    transparency=...,
+    transparency_range=...,
+) -> OrthoSlicer: ...
```
hndgzkn, line 326: @Remi-Gau I'm working on type annotations on plotting functions and I realize that OrthoSlicer might be misleading since the type of the slicer returned depends on the type of the display mode specified. I might replace OrthoSlicer with BaseSlicer.

Remi-Gau: otherwise you can maybe specify that using @overload

hndgzkn: But then it will be as many as number of slicers and it will be unmanageable, no?
BaseSlicer will cover all possible slicers.

Remi-Gau: ah yeah maybe easier to keep BaseSlicer then

hndgzkn: I'm moving this discussion to #6184. There are move that I would like to discuss.

PR #6202, opened 2026-05-05
[MAINT] add type to fill_doc

### LINKED ISSUE #6188: Typing fill_doc exposes existing signature conflicts in _MultiMixin and BaseMasker subclasses
I wanted to add type hints to docs.py for the fill_doc decorator.

fill_doc only mutates the decorated object's docstring in place and returns the same object unchanged, so it is a good candidate for a generic identity-preserving annotation:

```python
F = TypeVar("F")

def fill_doc(f: F) -> F:
```

This improves typing because decorated functions/classes keep their original inferred type signatures.

After adding the annotation above, mypy started failing in pre-commit with:
```bash
nilearn/maskers/multi_nifti_masker.py:65: error: Definition of "fit_transform"
in base class "_MultiMixin" is incompatible with definition in base class
"BaseMasker"  [misc]
    class MultiNiftiMasker(_MultiMixin, NiftiMasker):
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
nilearn/maskers/multi_nifti_maps_masker.py:9: error: Definition of
"fit_transform" in base class "_MultiMixin" is incompatible with definition in
base class "BaseMasker"  [misc]
    class MultiNiftiMapsMasker(_MultiMixin, NiftiMapsMasker):
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
nilearn/maskers/multi_nifti_labels_masker.py:9: error: Definition of
"fit_transform" in base class "_MultiMixin" is incompatible with definition in
base class "BaseMasker"  [misc]
    class MultiNiftiLabelsMasker(_MultiMixin, NiftiLabelsMasker):
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 3 errors in 3 files (checked 339 source files)
```
The classes that lead to the conflicting signatures by inheriting from _MultiMixin and via the other inheritance from BaseMasker are all in `multi_nifti_masker.py`:
- MultiNiftiMasker(_MultiMixin, NiftiMasker)
- MultiNiftiMapsMasker(_MultiMixin, NiftiMapsMasker)
- MultiNiftiLabelsMasker(_MultiMixin, NiftiLabelsMasker)

Before fill_doc was typed, mypy had less precise information for decorated methods.
Once fill_doc is typed as F -> F, mypy preserves the exact signature of decorated methods and can compare inherited methods more strictly.

That revealed an existing multiple-inheritance conflict between:
```python
class _MultiMixin:
    @fill_doc
    def fit_transform(self, imgs, y=None, confounds=None, sample_mask=None):
```
and
```python
class BaseMasker(_BaseMasker):
    @fill_doc
    def fit_transform( self, imgs, y=None, confounds=None, sample_mask=None, **fit_params ):
```
The relevant difference is that BaseMasker.fit_transform accepts **fit_params, while _MultiMixin.fit_transform does not.
### Summary

Typing fill_doc did not introduce a new bug. It exposed a pre-existing signature inconsistency that was previously hidden by the untyped decorator.

I would suggest aligning the signatures or simply removing the definition from _MultiMixin, since it appears to always be used in combination with some child of BaseMasker and is missing the **fit_params in the function signature.

I’d be happy to open a follow-up PR for whichever direction is preferred.

Remi-Gau:
- Closes #6188 

- add type to fill_doc

### DIFF
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -12,6 +12,7 @@ https://github.com/mne-tools/mne-python/blob/main/mne/utils/docs.py
 # sourcery skip: merge-dict-assign
 
 import sys
+from collections.abc import Callable
 
 ##############################################################################
 #
@@ -1937,7 +1938,7 @@ def _indentcount_lines(lines):
     return indentno
 
 
-def fill_doc(f):
+def fill_doc(f: Callable) -> Callable:
     """Fill a docstring with docdict entries.
 
     Parameters
@@ -1974,7 +1975,7 @@ def fill_doc(f):
         f.__doc__ = docstring % indented
     except (TypeError, ValueError, KeyError) as exp:
         funcname = f.__name__
-        funcname = docstring.split("\n")[0] if funcname is None else funcname
+        funcname = docstring.split("\n")[0]
         raise RuntimeError(
             f"Error documenting {funcname}:\n{exp!s}.\n"
             "Did you forget to escape a character with an extra '%'"
--- a/nilearn/_utils/docs.py
+++ b/nilearn/_utils/docs.py
@@ -1974,7 +1974,6 @@ def fill_doc(f: Callable) -> Callable:
     try:
         f.__doc__ = docstring % indented
     except (TypeError, ValueError, KeyError) as exp:
-        funcname = f.__name__
         funcname = docstring.split("\n")[0]
         raise RuntimeError(
             f"Error documenting {funcname}:\n{exp!s}.\n"
--- a/nilearn/maskers/_mixin.py
+++ b/nilearn/maskers/_mixin.py
@@ -54,6 +54,10 @@ class _MultiMixin:
         -------
         %(signals_transform_multi_nifti)s
         """
+        # although the implementation is
+        # the same as in the BaseMasker
+        # a specific method is required
+        # to allow for a slightly different doc string
         return self.fit(imgs, y=y, **fit_params).transform(
             imgs, confounds=confounds, sample_mask=sample_mask
         )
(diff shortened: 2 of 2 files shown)


nilearn/maskers/_mixin.py
```diff
@@ -28,7 +28,9 @@ class _MultiMixin:
     """Mixin class to add common MultiMasker functionalities."""
 
     @fill_doc
-    def fit_transform(self, imgs, y=None, confounds=None, sample_mask=None):
+    def fit_transform(
+        self, imgs, y=None, confounds=None, sample_mask=None, **fit_params
+    ):
```
Remi-Gau, line 33: @mcsitter 
turns out it was easier to do this

nilearn/_utils/docs.py
```diff
@@ -1974,7 +1975,7 @@ def fill_doc(f):
         f.__doc__ = docstring % indented
     except (TypeError, ValueError, KeyError) as exp:
         funcname = f.__name__
```
hndgzkn: If the below condition is removed, this line can also be removed as now it is overridden in any case.

hndgzkn (approved):
LGTM apart from a minor comment.

nilearn/_utils/docs.py
```suggestion
```

mcsitter:
Maybe you could make it explicit that types are preserved by adding the parameters to `Callable`?

```python
from typing import TypeVar, ParamSpec
from collections.abc import Callable

P = ParamSpec("P")
R = TypeVar("R")

def fill_doc(f: Callable[P, R]) -> Callable[P, R]:
    ...
```

bthirion (approved):
LGTM, thx.

PR #6204, opened 2026-05-05
[STY] use lower case 'default' in doc strings

Remi-Gau:
- Closes none

- make sure that "default=" is used instead of "Default="


Remi-Gau:
completely minor but tickled my OCD the wrong way

PR #3729, opened 2023-04-27
[ENH] Add niivue as rendering engine for `view_surf`

alexisthual:
- closes #3451

This PR is a follow-up on last weeks' experimentations regarding integrating `niivue` to `nilearn`.
It tries to add `niivue` as a rendering engine for `plotting.view_surf()`.

## TODO

- [x] check nilearn colormap are accepted
- [x] handle diverging colormap
- [x] handle vmin / vmax
- [x] symmetric_cmap
- [x] title
- [ ] add more tests that are more than smoke tests
   - [ ] possibly add tests to ensure that the output figure is 'stable' (maybe in a later PR)
- [x] refactor tests and also some of the code: maybe in a follow up PR
   - [x] _one_mesh_info and _one_mesh_info_niivue
   - [x] _fill_html_template and _fill_html_template_niivue
   - [x] probably worth using some jinja template across the various HTML figures

PR #6208, opened 2026-05-11
[MAINT] pin JS dependencies

Remi-Gau:
- Closes none

- forgotten to add a packages.json from #3729

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,6 @@
+{
+  "dependencies": {
+    "@niivue/niivue": "0.68.1",
+    "jquery": "3.6.0"
+  }
+}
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@niivue/niivue": "0.68.1",
+    "jquery": "3.6.0"
```
Remi-Gau, line 4: in theory we should also add plotly.js-gl3d-dist-min 1.46.1 as this is the one we ship with nilearn but it's old I could not find on npm:

https://www.npmjs.com/package/plotly.js-gl3d-dist-min?activeTab=versions

we may want to maybe bump it up

bthirion (approved):

PR #6211, opened 2026-05-11
[ENH] implement allow_overlap in SurfaceMapsMasker

### LINKED ISSUE #6092: SurfaceMapsMasker has `allow_overlap` but it is never used
AFAICT we never check if the maps do actually overlap or not

Remi-Gau:
- Closes #6092 

- add test to allow_overlap in SurfaceMapsMasker 
- implement check for overlap


Remi-Gau:
@hndgzkn 
this is the failing test I mentioned: let me know if if helps get you started

hndgzkn:
> @hndgzkn this is the failing test I mentioned: let me know if if helps get you started

Yes, thanks a lot!

nilearn/maskers/surface_maps_masker.py
```diff
@@ -304,6 +304,23 @@ def transform_single_imgs(self, imgs, confounds=None, sample_mask=None):
...
+                mask = maps_data != 0
+
+            # check overlap
+            if np.any(np.all(mask, axis=1)):
+                raise ValueError(
+                    "Overlap detected in the maps. The overlap may be "
+                    "due to the atlas itself or possibly introduced by "
+                    "resampling."
+                )
+
```
hndgzkn, line 331: I have started implementation looking at how it is implemented in `NiftiMapsMasker` and tried to improve on that. 

https://github.com/nilearn/nilearn/blob/104c462e5feb607117b23ed46e116e123103fc38/nilearn/maskers/nifti_maps_masker.py#L608-L623

I have some questions and remarks:

- In NiftiMapsMasker, it does in place replacement of values, but here `maps_data` is used afterwards, so I neither wanted to change the data, nor reuse `get_data`.
- As far as I understand, if two values in a row are greater than 0, there is overlap. I preferred to use a mask, rather then take sum on condition. I am creating the mask depending on dtype. Is it possible to have dtype other than float? If not, I can use a single mask.
- Another point is that, in NiftiMapsMasker the values less than eps are set to 0. Isn't it possible to have negative values? 
- Finally is the same error message reasonable for SurfaceMapsMasker?

Remi-Gau: > Finally is the same error message reasonable for SurfaceMapsMasker?

not quite because we don't have resampling of surface data (at least not within nilearn)

Remi-Gau: > Another point is that, in NiftiMapsMasker the values less than eps are set to 0. Isn't it possible to have negative values?

99% sure that values are going to be >=0 as they are supposed to be more or less akin to the probability of a vertex to belong a given brain region

Remi-Gau: > * Is it possible to have dtype other than float?

pretty sure it will almost always be some sort of float but I don't think we either enforce or check that anywhere

I guess in the rare case it is some sort of int, I would temporarily cast the maps to float just for check

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -142,3 +142,83 @@ def test_labels_img_none():
...
+            ]
+        ),
+    }
+    # multiply with random "probability" values
+    data = {part: data[part] * rng.random(data[part].shape) for part in data}
+    return SurfaceImage(surf_mesh, data)
+
+
+@pytest.mark.parametrize("allow_overlap", [True, False])
+def test_overlap(
```
Remi-Gau: I am realizing that we may want to check that the data extracted compared to what we would expect when we allow an overlap (i.e make sure this is more than just a smoke test)

nilearn/maskers/surface_maps_masker.py
hndgzkn: Deciding the mask depending on dtype would be less costly than a cast. I think we can keep it as it is now then.
My question was to avoid that if condition in case it was not necessary, ie no possibility of int dtype.

hndgzkn: > > 99% sure that values are going to be >=0 as they are supposed to be more or less akin to the probability of a vertex to belong a given brain region
> 
> If that is the case, that would be better to change the mask for float to `mask = maps_data > eps`

I'm sorry, I realized that I just changed your message thinking that I was responding :/

hndgzkn: I've reverted it, but some part might be missing.

hndgzkn: > 99% sure that values are going to be >=0 as they are supposed to be more or less akin to the probability of a vertex to belong a given brain region

If that is the case, that would be better to change the mask for float to `mask = maps_data > eps`

hndgzkn: > > Finally is the same error message reasonable for SurfaceMapsMasker?
> 
> not quite because we don't have resampling of surface data (at least not within nilearn)

In that case, would it be enough to keep only `Overlap detected in the map`?

Remi-Gau: > I've reverted it, but some part might be missing.

looks fine to me:

FYI github also tracks edits of messages:

image

Remi-Gau: > would it be enough to keep only `Overlap detected in the map`?

yes: we can try to improve the error message but that can be done in some later PR if we feel this is necessary

nilearn/maskers/tests/test_surface_maps_masker.py
hndgzkn: I've divided the tests into two, one for overlapping and one for non overlapping data and test for expected value. Is that reasonable?

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -142,3 +142,124 @@ def test_labels_img_none():
...
+        non_overlapping_maps, allow_overlap=allow_overlap, standardize=None
+    )
+    masker.fit()
+    surf_img = surf_img_2d(10)
+    region_signals = masker.fit_transform(surf_img)
+    assert np.allclose(
+        region_signals,
+        np.array(
+            [
+                [3.01710043e-01, 1.32242748e01],
```
Remi-Gau: hmm... maybe we should modify the tests to get numerical values that are more intuitive to check because at the moment I have no idea whether these values make sense or not (though I can see we get different value with overlapping maps and non overlapping ones)

maybe with a fixture with hard coded probability for only 1 vertex on each hemisphere

```
@pytest.fixture
def overlapping_maps(rng, surf_mesh):
    """Generate maps with overlapping regions.

    Some vertices have non null value for 2 different regions.
    """
    data = {
        "left": np.asarray(
            [
                [0, 0],
                [0, 0],
                [0, 0],
                [0.7, 0.1],  # overlap
            ]
        ),
        "right": np.asarray(
            [
                [0, 0],
                [0.3, 0.6],  # overlap
                [0, 0],
                [0, 0],  
                [0, 0],
            ]
        ),
    }
    return SurfaceImage(surf_mesh, data)
```

WDYT?

nilearn/maskers/tests/test_surface_maps_masker.py
Remi-Gau: > Is that reasonable?

yes though if we decide that checking exact numerical values is too annoying we could fall back on a single test that at least checks that overlapping and non overlapping maps give different results

nilearn/maskers/tests/test_surface_maps_masker.py
hndgzkn: I am not able to judge if the values makes sense, but I've added new tests with the fixture that you wrote here. Could you tell me if that is ok?

If the previous one is unnecessary, I can remove that.

One question: Should inverse transform give the same image used in fit_transform?

nilearn/maskers/tests/test_surface_maps_masker.py
hndgzkn: Checking exact numerical values might be good to detect unwanted changes, non?

Remi-Gau: yes but if we cannot easily assess if the numerical value 'make sense' then it is hard to know if the test is failing because the code is misbehaving or if because the test expected value should be updated

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -142,3 +142,180 @@ def test_labels_img_none():
...
+                [1.61248183e01, 5.21485653e01],
+                [3.19479265e01, 9.10728558e01],
+                [4.77710348e01, 1.29997146e02],
+                [6.35941430e01, 1.68921437e02],
+                [7.94172513e01, 2.07845727e02],
+                [9.52403595e01, 2.46770018e02],
+                [1.11063468e02, 2.85694308e02],
+                [1.26886576e02, 3.24618599e02],
+                [1.42709684e02, 3.63542889e02],
+            ]
```
bthirion: Do we need such a tight check ?

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -142,3 +142,180 @@ def test_labels_img_none():
...
+                    [5.71207998, 46.5370445],
+                    [8.65674774, 82.05929303],
+                    [11.60141551, 117.58154155],
+                    [14.54608327, 153.10379008],
+                    [17.49075104, 188.6260386],
+                    [20.4354188, 224.14828713],
+                    [23.38008657, 259.67053565],
+                    [26.32475433, 295.19278418],
+                    [29.2694221, 330.7150327],
+                ]
```
bthirion: Same here ?

nilearn/maskers/tests/test_surface_maps_masker.py
hndgzkn: Not really, I was just a bit lazy to make them smaller :) I will do that.

hndgzkn: Done

nilearn/maskers/tests/test_surface_maps_masker.py
hndgzkn: > One question: Should inverse transform give the same image used in fit_transform?

@emdupre Was your reaction to this question?

nilearn/maskers/tests/test_surface_maps_masker.py
```diff
@@ -142,3 +142,183 @@ def test_labels_img_none():
         match="provide a maps_img during initialization",
     ):
         SurfaceMapsMasker(maps_img=None).fit()
+
+
+@pytest.fixture
+def non_overlapping_maps(rng, surf_mesh):
+    """Generate maps with non-overlapping regions.
+
+    Each vertex belong to only 1 region.
```
bthirion:
```suggestion
    Each vertex belongs to only 1 region.
```

nilearn/maskers/tests/test_surface_maps_masker.py
bthirion: Sorry, I meant: do we need to check the exact numerical values ? testing that they're non-negative is not enough ?

hndgzkn: If there is not an inconvenience, I think keeping the exact value would be better.

bthirion (approved):

Remi-Gau:
@hndgzkn 
I made a suggestion via the commit https://github.com/nilearn/nilearn/pull/6211/commits/6b7e7dfba5b7abddf7216ab1fe10e3d592eea691
to try to simplify things: what do you think?

hndgzkn:
> @hndgzkn I made a suggestion via the commit [6b7e7df](https://github.com/nilearn/nilearn/commit/6b7e7dfba5b7abddf7216ab1fe10e3d592eea691) to try to simplify things: what do you think?

It seems much simpler and nice, thanks!

Remi-Gau:
OK let's wait for CI to finish and merge

PR #6237, opened 2026-05-22
[MAINT] fix plotting tests running in non-plotting tox environments

### LINKED ISSUE #6120: plotting are running even when they should not
see for example on this run : https://github.com/nilearn/nilearn/actions/runs/22897097137/job/66433902769#step:8:99

Image

to check `tox run -e min -- nilearn/plotting` should not run any tests in plotting but it does

this is in part due to the fact that pytest-mpl is installed with the testing dependencies and will install matplotlib

but when we do not installing pytest-mpl for the tests we get : 

```
tox run -e min -- nilearn/plotting

.pkg: _optional_hooks> python /home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.10/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: get_requires_for_build_sdist> python /home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.10/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: build_sdist> python /home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.10/site-packages/pyproject_api/_backend.py True hatchling.build
min: install_package> python -I -m pip install --force-reinstall --no-deps /home/remi-gau/github/nilearn/nilearn/.tox/.tmp/package/41/nilearn-0.13.2.dev62+gc761e43a3.d20260311.tar.gz
min: commands[0]> coverage erase
min: commands[1]> pytest -m 'not slow and not flaky and not single_process' -W error::sklearn.exceptions.ConvergenceWarning --timeout=60 --cov=nilearn --cov-report=xml --cov-report=html --cov-append --report=report.html --numprocesses auto --csv results/pytest_output/pytest_output.csv nilearn/plotting
ImportError while loading conftest '/home/remi-gau/github/nilearn/nilearn/nilearn/plotting/conftest.py'.
nilearn/plotting/__init__.py:5: in <module>
    set_mpl_backend()
nilearn/_utils/helpers.py:33: in set_mpl_backend
    import matplotlib
E   ModuleNotFoundError: No module named 'matplotlib'
min: exit 4 (1.12 seconds) /home/remi-gau/github/nilearn/nilearn> pytest -m 'not slow and not flaky and not single_process' -W error::sklearn.exceptions.ConvergenceWarning --timeout=60 --cov=nilearn --cov-report=xml --cov-report=html --cov-append --report=report.html --numprocesses auto --csv results/pytest_output/pytest_output.csv nilearn/plotting pid=241325
.pkg: _exit> python /home/remi-gau/github/nilearn/nilearn/.venv/lib/python3.10/site-packages/pyproject_api/_backend.py True hatchling.build
  min: FAIL code 4 (8.64=setup[7.44]+cmd[0.09,1.12] seconds)
  evaluation failed :( (8.75 seconds)
```

I would prefer to get no test ran than an error

Remi-Gau:
- Closes #6120

Use of AI tools:

- Move `pytest-mpl` out of the shared `test_common` dependency group into a new `test_mpl` group. `pytest-mpl` depends on `matplotlib`, so including it in every test environment caused `is_matplotlib_installed()` to return `True` in non-plotting envs (e.g. `min`, `latest`), bypassing the `collect_ignore` guard that is supposed to skip plotting tests.
- Add `test_mpl` to the `dependency_groups` of the two envs that actually pass `--mpl` (`pytest_mpl`, `pytest_mpl_generate`).
- Register `mpl_image_compare` as a known marker in `pyproject.toml` so `--strict-markers` does not error when `pytest-mpl` is not installed.
- Fix the crash that occurred when matplotlib is absent and pytest is pointed explicitly at `nilearn/plotting` (e.g. `tox run -e min -- nilearn/plotting`): `nilearn/plotting/__init__.py` now wraps `set_mpl_backend()` in a `try/except ImportError` so the package can be imported without matplotlib, allowing the conftest to load.
- Add a `pytest_ignore_collect` hook to `nilearn/plotting/conftest.py` that returns `True` for all paths inside the plotting directory when matplotlib is absent. This ensures zero tests are collected (rather than collection errors) when `nilearn/plotting` is an explicit command-line argument and matplotlib is not installed.
- Fix a pre-existing mypy error in `nilearn/_utils/ndimage.py` (wrapping `coordinates.transpose()` in `tuple()` to satisfy the `np.ravel_multi_index` overload).
- Removed some tests in nilearn/plotting that would require matplotlib to be absent to run


nilearn/_utils/ndimage.py
```diff
@@ -150,7 +150,7 @@ def peak_local_max(
 
     if coordinates.shape[0] > num_peaks:
         intensities = image.flat[
-            np.ravel_multi_index(coordinates.transpose(), image.shape)
+            np.ravel_multi_index(tuple(coordinates.transpose()), image.shape)
```
Remi-Gau, line 153: unrelated but fixes an issue that happens once in a while when running mypy locally

nilearn/plotting/__init__.py
Remi-Gau, line 1: @hndgzkn 
this was the suggestion of claude for this which makes sense to me

hndgzkn: I agree. I have two points though:

- actually `set_mpl_backend` is not used else where. May be its content can be moved here; however I am not sure if it is really a good idea.
- when there is ImportError, the functions in `__all__` are not imported. Doesn't that raise an error? Should `__all__` be defined both in `except` and `else` depending on matplotlib being installed?

hndgzkn: >     * actually `set_mpl_backend` is not used else where. May be its content can be moved here; however I am not sure if it is really a good idea.

Finally I think it would be reasonable. 

- try block in `__init__` can have only `import matplotlib`
- except block can print warning when `matplotlib` is not installed
- `set_mpl_backend` can be moved to `plotting._utils` removing the `import matplotlib` try-except block.
- `set_mpl_backend` can be called from `else` clause

Remi-Gau: will first make sure that all tests pass before I tweak things before I circle back to your comments

bthirion:
LMK if you want a pass on this.

nilearn/plotting/_utils.py
```diff
@@ -6,11 +6,77 @@
...
+    """
+    # We are doing local imports here to avoid polluting our namespace
+    try:
+        import matplotlib
+    except ImportError:
+        warning = (
+            "Some dependencies of nilearn.plotting package seem to be missing."
+            "\nThey can be installed with:\n"
+            " pip install 'nilearn[plotting]'"
+        )
```
hndgzkn: If I haven't missed, `set_mpl_backend` is not called from another place than `__init__` and only when matplotlib is installed. Then this try block becomes dead code. And warning will not be raised. 

I would suggest to delete try-except from here and move warning to except in `__init__`.

hndgzkn: Also tests could be adjusted accordingly.

hndgzkn (approved):
LGTM with a minor suggestion.

nilearn/plotting/_utils.py
bthirion: Thx !  This should be done before merging.

Remi-Gau: ah yes indeed

Remi-Gau:
realizing there is an issue with this approach: if we have tests in nilearn/plotting that need to run when matplotlib is NOT there, then they will never run because we are skipping ALL tests in nilearn/plotting when matplotlib is not there

from a quick check we only have a single test (test_get_surface_backend_matplotlib_not_installed) in this case

A more granular approach (but more tedious) to all this is to make more systematic use of our `matplotlib_pyplot` fixture that when used skips a test when matplotlib is not there

nilearn/_utils/tests/test_common.py
```diff
@@ -17,12 +16,8 @@ def test_all_modules_error(func):
         func(modules_to_ignore=["foo"], modules_to_consider=["bar"])
 
 
-@pytest.mark.skipif(
-    not is_matplotlib_installed(),
-    reason="This test requires matplotlib to be installed.",
-)
```
hndgzkn, line 23: This seems more clear to me unless matplotlib_pyplot is explicitly needed.

nilearn/plotting/_utils.py
```diff
@@ -6,11 +7,64 @@
...
+
+    If current backend is not usable, switch to default "Agg" backend.
+
+    Parameters
+    ----------
+    message: str, default=None
+        Message to be prepended to standard warning when matplotlib is not
+    installed.
+    """
+    with contextlib.suppress(ImportError):
```
hndgzkn, line 28: Do we need this line?

hndgzkn:
> realizing there is an issue with this approach: if we have tests in nilearn/plotting that need to run when matplotlib is NOT there, then they will never run because we are skipping ALL tests in nilearn/plotting when matplotlib is not there
> 
> from a quick check we only have a single test (test_get_surface_backend_matplotlib_not_installed) in this case

As it is not possible to import anything from plotting when matplotlib is not installed, we do not need to have tests for those cases. I think we can just delete that test.

nilearn/plotting/_utils.py
Remi-Gau: yup otherwise I get some tests failures:

```
pytest nilearn/plotting/tests/test_utils.py -k should
====================================================================== test session starts =======================================================================
platform linux -- Python 3.14.2, pytest-9.0.3, pluggy-1.6.0 -- /home/remi-gau/github/nilearn/nilearn/.venv/bin/python3
cachedir: .pytest_cache
Using --randomly-seed=3178076563
rootdir: /home/remi-gau/github/nilearn/nilearn
configfile: pyproject.toml
plugins: reporter-0.5.3, reporter-html1-0.9.3, cov-7.1.0, randomly-4.1.0, xdist-3.8.0, rerunfailures-16.3, datalad-1.4.1, csv-3.0.0, timeout-2.4.0
collected 105 items / 101 deselected / 4 selected                                                                                                                

nilearn/plotting/tests/test_utils.py::test_should_raise_import_error_for_version_check SKIPPED (could not import 'matplotlib.pyplot': No module named ...)
nilearn/plotting/tests/test_utils.py::test_should_not_raise_warning_if_backend_is_not_changed FAILED
nilearn/plotting/tests/test_utils.py::test_should_raise_warning_if_backend_changes FAILED
nilearn/plotting/tests/test_utils.py::test_should_switch_to_agg_backend_if_current_backend_fails SKIPPED (Test requires matplotlib to be installed.)
```

nilearn/_utils/tests/test_helpers.py
Remi-Gau, line 1: content was moved

nilearn/_utils/helpers.py
Remi-Gau, line 1: content was moved

nilearn/plotting/displays/tests/test_figures.py
```diff
@@ -27,15 +26,3 @@ def test_surface_figure_errors():
         figure.show()
     with pytest.raises(ValueError, match="You must provide an output file"):
         figure._check_output_file()
-
-
-@pytest.mark.skipif(
-    is_plotly_installed(),
-    reason="This test only runs if Plotly is not installed.",
-)
-def test_plotly_surface_figure_import_error():
```
Remi-Gau, line 36: test was never gonna be run

added a 'pragma no cover' in the code this was supposed to test

nilearn/plotting/surface/tests/test_utils.py
```diff
@@ -244,30 +240,6 @@ def test_get_faces_on_edge_matplotlib(in_memory_mesh):
         get_faces_on_edge(faces, [91])
 
 
-@pytest.mark.skipif(
```
Remi-Gau, line 247: those tests were never gonna be run

nilearn/plotting/displays/tests/test_figures.py
hndgzkn: Isn't it run when matplotlib is installed?

Remi-Gau: ah yes: let me revert

nilearn/plotting/surface/tests/test_utils.py
```diff
@@ -244,30 +240,6 @@ def test_get_faces_on_edge_matplotlib(in_memory_mesh):
...
-    is_plotly_installed(),
-    reason="This test is run only if plotly is not installed.",
-)
-def test_get_surface_backend_plotly_not_installed():
-    """Tests to see if get_surface_backend raises error when plotly is not
-    installed.
-    """
-    with pytest.raises(ImportError, match="Using engine"):
-        get_surface_backend("plotly")
-
```
hndgzkn, line 269: The above one is not run if matplotlib is installed. However this one could be run when matplotlib is installed but not plotly, no?

Remi-Gau: indeed

nilearn/plotting/surface/_utils.py
```diff
@@ -58,15 +58,15 @@ def get_surface_backend(engine=DEFAULT_ENGINE):
     if engine == "matplotlib":
         if is_matplotlib_installed():
             import nilearn.plotting.surface._matplotlib_backend as backend
-        else:
+        else:  # pragma: no cover
             raise ImportError(
                 "Using engine='matplotlib' requires that ``matplotlib`` is "
                 "installed."
             )
```
hndgzkn: I think here we can just do the import:

```
import nilearn.plotting.surface._matplotlib_backend as backend
```

if-else block is unnecessary.

nilearn/plotting/surface/_utils.py
```diff
@@ -58,15 +58,15 @@ def get_surface_backend(engine=DEFAULT_ENGINE):
...
     elif engine == "plotly":
         if is_plotly_installed():
             import nilearn.plotting.surface._plotly_backend as backend
-        else:
+        else:  # pragma: no cover
             raise ImportError(
                 "Using engine='plotly' requires that ``plotly`` is installed."
             )
```
hndgzkn, line 66: However for the case matplotlib is installed bu not plotly, this check is important.

nilearn/plotting/surface/_utils.py
```diff
@@ -58,15 +58,15 @@ def get_surface_backend(engine=DEFAULT_ENGINE):
...
+        else:  # pragma: no cover
             raise ImportError(
                 "Using engine='matplotlib' requires that ``matplotlib`` is "
                 "installed."
             )
     elif engine == "plotly":
         if is_plotly_installed():
             import nilearn.plotting.surface._plotly_backend as backend
-        else:
+        else:  # pragma: no cover
```
Remi-Gau:
```suggestion
        else:
```

nilearn/plotting/surface/_utils.py
Remi-Gau: I think it's fine to keep it this way we get the same error message for both backends

nilearn/plotting/_utils.py
hndgzkn: Does it mean that those tests are run when matplotlib is not installed?

nilearn/plotting/surface/_utils.py
hndgzkn: But `is_matplotlib_installed` will always evaluate to True and `else` part is dead code.

bthirion (approved):
LGTM thx.
It seems that the test failure is unrelated.

nilearn/plotting/surface/_utils.py
Remi-Gau: ah yes I see what you man now

PR #6240, opened 2026-05-26
[MAINT] raise error when maps or labels images contain no map or label after resampling

### LINKED ISSUE #6094: label and maps masker should tell if no ROI is left after masking / resampling
We actually have something like this in one our tests:

```python
from nilearn.conftest import _img_labels, _shape_3d_large, _affine_mni, _rng

estimator = NiftiLabelsMasker(labels_img=_img_labels())

input_img = Nifti1Image(
    _rng().random(_shape_3d_large()), _affine_mni()
)

signal = estimator.fit_transform(input_img)

print(estimator.n_elements_)
print(estimator.region_names_)
print(estimator.labels_)
print(estimator.region_ids_)
print(signal)
```

this gives:

```
/home/remi-gau/github/nilearn/nilearn/tmp/empty_img.py:11: UserWarning: After resampling the label image to the data image, the following labels were removed: {1, 2, 3, 4, 5, 6, 7, 8, 9}. Label image only contains 1 labels (including background).
  signal = estimator.fit_transform(input_img)
0
{}
[0]
{'background': 0}
[]
```

We get a warning that there is nothing left but the background (this is a generic warning about what labels were dropped), but I am thinking that this should maybe be an error when it means that your data extractor won't extract ANY data.

@nilearn/core how do people feel about this one?

Remi-Gau:
- Closes #6094 

Use of AI tools:

- raise error when maps or labels images contain no map or label after resampling


nilearn/maskers/nifti_labels_masker.py
```diff
@@ -805,6 +813,10 @@ def _resample_labels(self, imgs_):
             target_affine=imgs_.affine,
         )
         labels_after_resampling = set(np.unique(safe_get_data(labels_img_)))
+        if len(labels_after_resampling) <= 1:
```
hndgzkn, line 821: If `len(labels_after_resampling) == 1`, still no label?

hndgzkn (approved):
LGTM.

nilearn/maskers/nifti_labels_masker.py
Remi-Gau: yeah that's a bit confusing because labels_after_resampling also include the background value, so its minimum value should be 1 (am using <= for some weird edge cases where we would not even get some background voxel)

I could rename the variable to something like: labels_and_background_after_resampling

hndgzkn: Just putting a comment here might be sufficient.

bthirion (approved):
LGTM, thx.

PR #6160, opened 2026-03-27
[ENH] use brainsprite for masks in GLM

Remi-Gau:
- Relates #6036 

Changes proposed in this pull request:

- use brainsprite for masks in GLM

PR #6249, opened 2026-06-01
[FIX] fix order parameters glm generate_report

### LINKED ISSUE #6248: [BUG] glm report generation falsely claims no constrast was passed

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Linux Ubuntu 24.04

### Python version

- [ ] 3.14
- [ ] 3.13
- [ ] 3.12
- [ ] 3.11
- [x] 3.10

### nilearn version

dev

### Expected behavior

that when passing constrasts to a GLM object generate_report, contrasts are included in the report

see dev doc

https://nilearn.github.io/dev/auto_examples/04_glm_first_level/plot_localizer_surface_analysis.html#generate-a-report-for-the-glm

Image

compared to stable

https://nilearn.github.io/stable/auto_examples/04_glm_first_level/plot_localizer_surface_analysis.html#generate-a-report-for-the-glm

Image

```python
python examples/04_glm_first_level/plot_localizer_surface_analysis.py
```

Remi-Gau:
- Closes #6248 

Use of AI tools:

- change order of parameters in generate_report for GLM


tests/js/references/flm_bids_features_reference.png
Remi-Gau, line 1: not enough difference I think: let's reset it

tests/js/references/slm_oasis_reference.png
Remi-Gau, line 1: not enough difference I think: let's reset it

tests/js/references/view_connectome_reference.png
Remi-Gau, line 1: not enough difference I think: let's reset it

tests/js/references/flm_surf_reference.png
Remi-Gau, line 1: I figured it would be good to have one GLM report baseline with surface

examples/04_glm_first_level/plot_localizer_surface_analysis.py
Remi-Gau, line 1: dummy changes to force the example to run during doc build

nilearn/glm/_reporting_utils.py
```diff
@@ -36,6 +36,7 @@
 
 class GLMReportMixin(ReportMixin):
     _REPORT_DEFAULTS: ClassVar[dict[str, Any]] = {
+        "engine": "matplotlib",
```
hndgzkn: You do not need to put this, it already exists in `nilearn.reporting.mixin.ReportMixin`. 

Here we should have only fields specific to GLM. Those will be merged with the fields in `ReportMixin.`

Remi-Gau: ha yeah sorry that's a left over from me playing around with some other things

tests/js/references/flm_surf_reference.png
Remi-Gau: OK there a small failure here because of the order in which the conditions are printed: let's sort them out to make it consistent

nilearn/glm/_reporting_utils.py
Remi-Gau:
```suggestion
```

doc/visual_testing/reporter_visual_inspection_suite.py
```diff
@@ -110,7 +110,7 @@ def report_flm_adhd_dmn(build_type):
     )
 
     report = first_level_model.generate_report(
-        contrasts=contrasts,
+        contrasts,
```
hndgzkn: If we are going to try to unify the `generate_report` method for estimators (as discussed [here](https://github.com/nilearn/nilearn/issues/6248#issuecomment-4592807059)) inheriting from `ReportMixin`, I think it would be better to keep specifying parameter name, otherwise, we will have to revert this change and the similar following ones later.

nilearn/glm/_reporting_utils.py
```diff
@@ -140,6 +139,7 @@ def generate_report(
         cut_coords=None,
         display_mode=None,
         report_dims=(1600, 800),
+        engine="matplotlib",
```
hndgzkn: If we do not need to give a warning to user about this API change, and an entry in change log would do it, we could keep this and update the code where `generate_report` is called.

doc/visual_testing/reporter_visual_inspection_suite.py
Remi-Gau: yeah that's fair but we should have more tests at some point that should fail if we move the order of parameters

nilearn/glm/_reporting_utils.py
```diff
@@ -125,10 +125,9 @@ def _model_attributes_to_dataframe(self, model) -> pd.DataFrame:
     @fill_doc
     def generate_report(
         self,
-        engine="matplotlib",
```
Remi-Gau, line 128: @hndgzkn 
for I am removing engine from the method

I will probably will have to reintroduce in #6160 but I'd prefer to not have it in the API until it can have meaningfull effect

hndgzkn (approved):
LGTM

PR #6261, opened 2026-06-04
[MAINT] remove flaky marker

### LINKED ISSUE #4876: [BUG] test_canica_square_img fails on some x86_64 machines

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

Fedora Rawhide (future Fedora 42)

### Python version

- [x] 3.13
- [ ] 3.12
- [ ] 3.11
- [ ] 3.10
- [ ] 3.9

### nilearn version

0.11.0

### Expected behavior

All tests pass

On `x86_64` only – not on `aarch64`, `ppc64le`, or `s390x` – and seemingly consistently on real Fedora build machines, but not in what should be an identical chroot environment on my local workstation – the `test_canica_square_img` fails as follows:

```
FAILED nilearn/decomposition/tests/test_canica.py::test_canica_square_img - assert 2 == 4
 +  where 2 = <function sum at 0x7f87cfe8dab0>(array([[0.01609518, 0.63380754, 0.68215817, 0.00981432],\n       [       nan,        nan,        nan,        nan],\n       [0.90782467, 0.15731803, 0.18897407, 0.01181167],\n       [0.00775289, 0.00455801, 0.08132788, 0.94071187]]) > 0.9)
 +    where <function sum at 0x7f87cfe8dab0> = np.sum
```

It should suffice to just run the tests, in whatever way one finds practical, on an `x86_64` machine. One should be able to pass `-k test_canica_square_img` to save time.

The problem is that I don’t know why I can’t reproduce this locally, so I can’t give reliable instructions to make it actually fail in a particular environment.

I cannot reproduce this (still with 0.11.0) in Fedora 41 or Fedora 40, so this is specific to Rawhide. I can reproduce this with 0.10.4 in Fedora Rawhide, so this does not appear to be a 0.11.0 regression. It is likely related to an update of one of the dependencies, such as `scipy` or `scikit-learn`, but due to a combination of test-build logs being cleaned up and multiple regressions, I’m not able to easily pin down which update might have been associated with this specific regression.

Remi-Gau:
- Closes #4876

Use of AI tools:

- check if some tests have become less flaky

### DIFF
--- a/nilearn/decomposition/tests/test_canica.py
+++ b/nilearn/decomposition/tests/test_canica.py
@@ -47,7 +47,6 @@ def test_percentile_range(rng, canica_data_single_img):
         canica.fit(canica_data_single_img)
 
 
-@pytest.mark.flaky(reruns=5, reruns_delay=2, condition=is_windows_platform())
 @pytest.mark.parametrize("data_type", ["nifti"])
 def test_canica_square_img(
     decomposition_mask_img, canica_components, canica_data
--- a/nilearn/decomposition/tests/test_canica.py
+++ b/nilearn/decomposition/tests/test_canica.py
@@ -4,7 +4,6 @@ import numpy as np
 import pytest
 from numpy.testing import assert_array_almost_equal
 
-from nilearn._utils.helpers import is_windows_platform
 from nilearn.decomposition.canica import CanICA
 from nilearn.decomposition.tests.conftest import (
     RANDOM_STATE,
--- a/nilearn/decomposition/tests/test_decomposition_estimators.py
+++ b/nilearn/decomposition/tests/test_decomposition_estimators.py
@@ -57,7 +57,7 @@ else:
 
 
 @pytest.mark.slow
-@pytest.mark.flaky(reruns=5, reruns_delay=2, condition=is_windows_platform())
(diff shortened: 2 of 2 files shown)


nilearn/decomposition/tests/test_decomposition_estimators.py
```diff
@@ -57,7 +57,7 @@ def test_check_estimator_sklearn(estimator, check):
 
 
 @pytest.mark.slow
-@pytest.mark.flaky(reruns=5, reruns_delay=2, condition=is_windows_platform())
```
Remi-Gau, line 60: pretty sure this is not just a windows issue

nilearn/decomposition/tests/test_canica.py
```diff
@@ -47,7 +47,6 @@ def test_percentile_range(rng, canica_data_single_img):
         canica.fit(canica_data_single_img)
 
 
-@pytest.mark.flaky(reruns=5, reruns_delay=2, condition=is_windows_platform())
```
Remi-Gau, line 50: recent improvement in the fixture used to generate data for canica seem to have fixed this flaky test: let's see

Remi-Gau:
will merge and revert if the tests are still flaky

Issue #5525, opened 2025-07-09
should there be restrictions / warnings when user require some dtype for some maskers?


Remi-Gau:
Discussed in the dev meeting: should there be restrictions / warnings when user require some dtype for some maskers?

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/issues/5511#issuecomment-3051881218_

emdupre:
- I guess there's a loose relationship with #5524 ?

Remi-Gau:
> I guess there's a loose relationship with [#5524](https://github.com/nilearn/nilearn/issues/5524) ?

yes and no, but definitely having a way to initialize SurfaceImage with a specific dtype will help with some downstream things like the inverse_transform of the surface maskers

Remi-Gau:
OK the recent merge of https://github.com/nilearn/nilearn/pull/5511 has created a bug when trying to inverse transform to an array boolean 

See failure of doc build:

https://github.com/nilearn/nilearn/actions/runs/27087693450/job/79945450443#step:14:26006

I will open a PR to disallow initializing estimator with a dtype=bool and also making sure that if boolean are passed to inverse_transform they are first cast to int32

bthirion:
oops, thx.

PR #6271, opened 2026-06-08
[FIX] forbid  bool as dtype for estimators

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/issues/5525#issuecomment-4646130161

Use of AI tools:

- avoid dtype = bool for estimators and cast bool input to inverse_transform to int32 (with a warning)
- _gifti_img_to_data: try numeric array first; fall back to object
  only for genuinely ragged darrays (different shapes).
- PolyData.__init__ / _set_dtype: detect object dtype, emit a
  UserWarning, and cast to np.float32 automatically.

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -75,7 +75,7 @@ Fixes
 
 - :bdg-dark:`Code` Do not copy mask image's header in ``masking.unmask`` (:gh:`6157` by `Taylor Salo`_).
 
-- :bdg-dark:`Code` Make sure that area data of Freesurfer can loaded as :class:`~surface.SurfaceImage` by :func:`~nilearn.datasets.load_fsaverage_data` (:gh:`6230` by `Rémi Gau`_).
+- :bdg-dark:`Code` Make sure that area data of Freesurfer can be loaded as :class:`~surface.SurfaceImage` by :func:`~nilearn.datasets.load_fsaverage_data` (:gh:`6230` by `Rémi Gau`_).
 
 - :bdg-dark:`Code` Fix order of vertices and data for ``fsaverage3`` and ``fsaverage4`` datasets and warn user (:gh:`6127` by `Rémi Gau`_ and `Hande Gözükan`_).
 
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -63,6 +63,10 @@ Fixes
 
 - :bdg-info:`Plotting` Fix ``nilearn.plotting.view_img`` resampling of non-isotropic images when no background image is used (:gh:`6031` by `Michelle Wang`_).
 
+- :bdg-dark:`Code` Disallow boolean dtype for estimators. (:gh:`6271` by `Rémi Gau`_).
+
+- :bdg-dark:`Code` Better checks of dtype for :class:`~surface.SurfaceImage` and :class:`~nilearn.surface.PolyData`. (:gh:`6271` by `Rémi Gau`_).
+
 - :bdg-dark:`Code` Ensure that estimators that accept images can work will several datatypes as input and that their methods can output arrays or images of the requested datatype (:gh:`5511` by `Rémi Gau`_).
 
 - :bdg-info:`Plotting` Fix bug introduced in ``0.13.0`` while plotting single valued data with :func:`~nilearn.plotting.plot_roi`.  (:gh:`6122` by `Hande Gözükan`_).
@@ -71,7 +75,7 @@ Fixes
 
 - :bdg-dark:`Code` Do not copy mask image's header in ``masking.unmask`` (:gh:`6157` by `Taylor Salo`_).
 
-- :bdg-dark:`Code` Make sure that area data of Freesurfer can loaded as SurfaceImage by :func:`~nilearn.datasets.load_fsaverage_data` (:gh:`6230` by `Rémi Gau`_).
+- :bdg-dark:`Code` Make sure that area data of Freesurfer can loaded as :class:`~surface.SurfaceImage` by :func:`~nilearn.datasets.load_fsaverage_data` (:gh:`6230` by `Rémi Gau`_).
 
 - :bdg-dark:`Code` Fix order of vertices and data for ``fsaverage3`` and ``fsaverage4`` datasets and warn user (:gh:`6127` by `Rémi Gau`_ and `Hande Gözükan`_).
 
--- a/examples/02_decoding/plot_oasis_vbm.py
+++ b/examples/02_decoding/plot_oasis_vbm.py
@@ -83,20 +83,21 @@ print(
 # %%
 # Preprocess data
 # ---------------
+# The features with too low between-subject variance are removed using
+# :class:`sklearn.feature_selection.VarianceThreshold`.
+# Then we convert the data back to the mask image in order to use it for
+# decoding process
+#
 nifti_masker = NiftiMasker(
     standardize=False, smoothing_fwhm=2, memory="nilearn_cache", verbose=1
 )  # cache options
 gm_maps_masked = nifti_masker.fit_transform(gm_imgs_train)
 
-# The features with too low between-subject variance are removed using
-# :class:`sklearn.feature_selection.VarianceThreshold`.
 from sklearn.feature_selection import VarianceThreshold
 
 variance_threshold = VarianceThreshold(threshold=0.01)
 variance_threshold.fit_transform(gm_maps_masked)
 
-# Then we convert the data back to the mask image in order to use it for
-# decoding process
 mask = nifti_masker.inverse_transform(variance_threshold.get_support())
 
 # %%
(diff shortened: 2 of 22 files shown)


nilearn/_utils/estimator_checks.py
```diff
@@ -1780,23 +1815,32 @@ def check_img_estimator_dtypes(estimator_orig):
...
+        ]
+
+    memory_list = [None]
+    if hasattr(estimator_orig, "memory"):
+        memory = [None, Path(mkdtemp())]
+
+    for input_dtype in [np.float32, "float64", np.int32, "i4"]:
+        for dtype in dtype_list:
+            for memory in memory_list:
                 estimator = clone(estimator_orig)
```
Remi-Gau, line 1938: unrelated but should avoid re-running some unnecessary checks for estimators that do not have a dtype or memory attribute

nilearn/image/tests/test_image.py
```diff
@@ -2168,6 +2169,13 @@ def test_check_niimg(img_3d_zeros_eye, img_4d_zeros_eye):
     )
 
 
+@pytest.mark.thread_unsafe
```
Remi-Gau:
```suggestion
```

Remi-Gau:
Note that this kind of character `→` may cause to crash during tests: may be good to avoid having AI agents use them in comments or docstrings

maint_tools/check_docstrings.py
```diff
@@ -294,7 +294,7 @@ def check_returns_yields_and_annotation(
 
     np_docstring = NumpyDocString(docstring)
 
-    # function returns / yields a value → must have Returns / Yields section
+    # function returns / yields a value to must have Returns / Yields section
```
bthirion: This reads pretty bad. I don't think I understand

nilearn/surface/surface.py
```diff
@@ -1323,12 +1333,6 @@ class PolyData:
             the typical shape of the
             data for a hemisphere is ``(n_vertices, n_time_points)``.
 
-    dtype : DTypeLike object, default=None
-        dtype to enforce on the data.
-        If ``None`` the original dtype if used.
-
-        .. nilearn_versionadded:: 0.12.1
```
Remi-Gau, line 1330: this is parameter but not a public attribute

nilearn/surface/surface.py
Remi-Gau, line 1: This PR makes me realize that SurfaceImage can have boolean data but that NiftiImage cannot: I am wondering if we should not try to align the behavior of SurfaceImage and also disallow boolean data for SurfaceImage.

For the moment I am casting boolean SurfaceImage to int8 when saving them to disk to avoid an error, but we could maybe start throwing deprecation warnings that boolean data won't be allowed in the future.

Note: this is not a hill I will die on.

Will open a PR to also test this to see if any of our internal stuff would break if we disallowed boolean surface images.

examples/02_decoding/plot_oasis_vbm.py
Remi-Gau, line 1: dummy change to trigger build of that example that is failing on main

examples/07_advanced/plot_surface_image_and_maskers.py
Remi-Gau, line 1: dummy change to trigger build of that example that is failing on main

maint_tools/check_docstrings.py
Remi-Gau: hope that reads better now

doc/changes/latest.rst
```diff
@@ -71,7 +75,7 @@ Fixes
 
 - :bdg-dark:`Code` Do not copy mask image's header in ``masking.unmask`` (:gh:`6157` by `Taylor Salo`_).
 
-- :bdg-dark:`Code` Make sure that area data of Freesurfer can loaded as SurfaceImage by :func:`~nilearn.datasets.load_fsaverage_data` (:gh:`6230` by `Rémi Gau`_).
+- :bdg-dark:`Code` Make sure that area data of Freesurfer can loaded as :class:`~surface.SurfaceImage` by :func:`~nilearn.datasets.load_fsaverage_data` (:gh:`6230` by `Rémi Gau`_).
```
```suggestion
- :bdg-dark:`Code` Make sure that area data of Freesurfer can be loaded as :class:`~surface.SurfaceImage` by :func:`~nilearn.datasets.load_fsaverage_data` (:gh:`6230` by `Rémi Gau`_).
```

examples/02_decoding/plot_oasis_vbm.py
```diff
@@ -83,20 +83,21 @@
 # %%
 # Preprocess data
 # ---------------
+# The features with too low between-subject variance are removed using
+# :class:`sklearn.feature_selection.VarianceThreshold`.
+# Then we convert the data back to the mask image in order to use it for
+# decoding process
+#
```
hndgzkn, line 90: These comments seemed better in their previous place to me. Here there might be an introduction.

Remi-Gau:
will try to make some tests less flaky

nilearn/maskers/base_masker.py
```diff
@@ -644,6 +649,12 @@ def _check_array(
                 f"Got {signals.shape}."
             )
 
+        if signals.dtype == bool:
+            warnings.warn(
+                "Casting boolean input to int32", stacklevel=find_stack_level()
+            )
+            signals = signals.astype(np.int32)
```
hndgzkn: As far as I understand, `signal.dtype` will be converted to `self.dtype` (somewhere in the code, I haven't verified, but I deduce it from docstring of dtype) if it is different from `None`. Could it be better to convert it to `self.dtype` here, to avoid double conversion?

nilearn/surface/surface.py
```diff
@@ -1314,6 +1318,12 @@ class PolyData:
     right : 1/2D :obj:`numpy.ndarray` or :obj:`str` or :obj:`pathlib.Path` \
             or None, default = None
 
+    dtype : DTypeLike object, default=None
+        dtype to enforce on the data.
+        If ``None`` the original dtype if used.
```
hndgzkn:
```suggestion
        If ``None`` the original dtype is used.
```

pyproject.toml
```diff
@@ -502,8 +502,8 @@ convention = "numpy"
 # https://docs.astral.sh/ruff/settings/#lint_pylint_max-args
 max-args = 27
 # https://docs.astral.sh/ruff/settings/#lint_pylint_max-branches
-max-branches = 22
+max-branches = 25
 # https://docs.astral.sh/ruff/settings/#lint_pylint_max-returns
 max-returns = 8
 # https://docs.astral.sh/ruff/settings/#lint_pylint_max-statements
-max-statements = 96
+max-statements = 100
```
hndgzkn: Would it be possible to refactor code instead of increasing these?

hndgzkn (approved):
LGTM with some comments and suggestions.

pyproject.toml
Remi-Gau: let me check: that may require one of the estimator check in 2

examples/02_decoding/plot_oasis_vbm.py
Remi-Gau: OK I will make a not to revert them: still want to have some change in the PR while working on it to make sure the examples that fail on main now run without failing

examples/02_decoding/plot_oasis_vbm.py
Remi-Gau: - [ ] REVERT BEFORE MERGE

pyproject.toml
Remi-Gau: OK I could avoid bumping one of those of up but for the other I do not see an easy to change this.

Note that the splitting the estimator check in 2 leads to some code duplication

nilearn/maskers/base_masker.py
Remi-Gau: yup makes sense

nilearn/maskers/base_masker.py
```diff
@@ -684,7 +697,7 @@ def _create_brainsprite(
         self._reporting_data["stat_map_base64"] = json_view["stat_map_base64"]
         self._reporting_data["params"] = json.dumps(json_view["params"])
 
-    def _set_inverse_transform_output_dtype(
+    def _post_process_inverse_transform(
```
Remi-Gau, line 700: renamed so that the method has the same name for nifti and surface maskers

nilearn/_utils/estimator_checks.py
```diff
@@ -1767,36 +1769,179 @@ def check_img_estimator_pipeline_consistency(estimator_orig) -> None:
...
+    if isinstance(signal, list):
+        signal = [x.astype(bool) for x in signal]
+    else:
+        signal = signal.astype(bool)
+
+    with pytest.warns(UserWarning, match="Casting boolean input to int32"):
+        estimator.inverse_transform(signal)
+
+
+def check_img_estimator_dtypes_transform(estimator_orig):
```
Remi-Gau, line 1806: extracted from check_img_estimator_dtypes to have simpler checks

nilearn/surface/surface.py
bthirion: Yep, depecating bool for surfaces makes sense.

bthirion (approved):
LGTM, thx.

PR #6272, opened 2026-06-09
[MNT] Add type annotations to nilearn.plotting

### LINKED ISSUE #6184: Improve type hints for conditional return types (reduce IDE/type checker warnings)

hndgzkn:
- Closes #6184 

Use of AI tools:

- Update `nilearn.plotting.displays._slicers` `savefig` method to have its own save logic so that it can be called directly on slicer objects.
- Move `nilearn.plotting.displays._slicers.save_figure_if_needed` to `nilearn.plotting._engine_utils`
- Update `save_figure_if_needed` to only expect an object with `savefig` function : matplotlib figure, BaseSlicer or PlotlySurfaceFigure
- Ensure that all plotting functions return figure or axes, and not None in case output_file is specified to save the figure and update the docstrings and the tests accordingly
- Remove code depending on matpotlib version being greater than `3.8.0rc1` as min matplotlib version is `3.8.0` 
- Replace `OrthoSlicer` in function return types in `nilearn.plotting.image.img_plotting` with `BaseSlicer`
- Set function return types in `nilearn.plotting.matrix.matrix_plotting`
- Update change log

-------------------
Note : Does not set return types for `nilearn.plotting.surface.surf_plotting` functions. In my opinion it requires a bit more of a reflection:

Most functions have output depending on `engine` type. It can return a matplotlib figure, or a plotly surface figure. 
Might require `@overload`s or alternatively definition of a BaseSurfaceFigure.


Remi-Gau:
early feedback: that's gonna need a changelog entry

Remi-Gau:
I am seeing this doc build fail in other PRs:
https://github.com/nilearn/nilearn/actions/runs/27190405487/job/80271327757?pr=6272#step:14:1577

I suspect this is an upstream issue with the joblib doc

nilearn/plotting/image/img_plotting.py
```diff
@@ -2178,9 +2129,7 @@ def plot_carpet(
         else:
             ax0.set_yticks([])
 
-        # Carpet plot
-        if compare_version(mpl_version, ">=", "3.8.0rc1"):
-            axes.remove()  # remove axes for newer versions of mpl
+        axes.remove()  # remove axes for newer versions of mpl
         axes = plt.subplot(gs[1])  # overwrites axes with older versions of mpl
```
hndgzkn, line 2135: Do we still need this line? I removed the above if condition as min matplotlib version is now 3.8.0

nilearn/plotting/image/img_plotting.py
```diff
@@ -544,7 +493,7 @@ def plot_anat(
     vmin=None,
     vmax=None,
     **kwargs,
-) -> OrthoSlicer | None:
+) -> OrthoSlicer:
```
hndgzkn: The type of the object returned depends on the `display_mode` specified. Is it necessary to create overloads for each possibility or only writing `BaseSlicer` would be sufficient?

Remi-Gau: Let's simplify and go with BaseSlicer for now.

nilearn/plotting/image/img_plotting.py
Remi-Gau: I guess we can remove the comment as well.

hndgzkn: Done!

nilearn/plotting/image/img_plotting.py
hndgzkn: Done

mcsitter:
Thank you for the work! This is highly appreciated :)

nilearn/plotting/img_comparison.py
```diff
@@ -35,7 +36,7 @@ def plot_img_comparison(
     output_dir=None,
     axes=None,
     colorbar=True,
-):
+) -> list:
```
mcsitter: Maybe we could be more precise about the return type here; it should be list[float] or np.float, and if it is the latter we may want to cast the output to float.

nilearn/plotting/img_comparison.py
```diff
@@ -92,7 +93,7 @@ def plot_img_comparison(
 
     Returns
     -------
-    corrs : :class:`numpy.ndarray`
+    corrs : list
```
mcsitter: list[float]? see above

nilearn/plotting/_engine_utils.py
```diff
@@ -26,6 +27,30 @@
 )
 
 
+def save_figure_if_needed(fig, output_file):
```
mcsitter: Maybe we should align the type hints with the docstring, and also add an explicit None return type to the function signature.

nilearn/plotting/img_comparison.py
hndgzkn: I've put `list[float]` looking at the documentation of https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.pearsonr.html

nilearn/plotting/_engine_utils.py
hndgzkn: Done. Thank you for your comments @mcsitter .

bthirion (approved):
LGTM, thx !

doc/changes/latest.rst
```diff
@@ -177,6 +177,8 @@ Changes
 
 - :bdg-danger:`Deprecation` The ``"z_score"`` value for the ``standardize`` parameter is no longer supported. Use ``standardize="z_score_sample"`` instead (:gh:`5995` by `Rémi Gau`_).
 
+- :bdg-dark:`Code` Update plotting functions to return figure or axes instead of None  when an output file is specified to save the figure (:gh: `` by `Hande Gözükan`_).
```
```suggestion
- :bdg-dark:`Code` Update plotting functions to return figure or axes instead of None when an output file is specified to save the figure (:gh:`6272` by `Hande Gözükan`_).
```

nilearn/plotting/displays/_slicers.py
```diff
@@ -1118,6 +1118,16 @@ def savefig(self, filename, dpi=None, **kwargs) -> None:
...
         """
+        if filename is None:
+            log("Please specify a valid filename: {filename=}.", verbose=1)
+            return
+
+        output_file = Path(filename)
+        try:
+            output_file.parent.mkdir(exist_ok=True, parents=True)
+        except Exception:
+            log("Please specify a valid file path: {filename=}.", verbose=1)
```
Remi-Gau: maybe we should return early here too, no?

nilearn/plotting/_engine_utils.py
```diff
@@ -26,6 +29,30 @@
...
+
+    output_file: str, Path or None
+        file path to save the figure
+    """
+    if output_file is not None:
+        output_file = Path(output_file)
+        try:
+            output_file.parent.mkdir(exist_ok=True, parents=True)
+        except Exception:
+            log("Please specify a valid file path: {output_file=}.", verbose=1)
```
Remi-Gau: same here: maybe return early?

nilearn/plotting/displays/_slicers.py
```diff
@@ -1118,6 +1118,16 @@ def savefig(self, filename, dpi=None, **kwargs) -> None:
             :func:`matplotlib.pyplot.savefig`.
 
         """
+        if filename is None:
+            log("Please specify a valid filename: {filename=}.", verbose=1)
```
Remi-Gau: any reason why use log instead of a warning?

nilearn/plotting/displays/_slicers.py
mcsitter: It would be better for debugging and interactive use if the figure stays open on, if it can't be saved.

Remi-Gau: yeah that was my reading: the figure cannot be saved and then it's closed

annoying from a UX perspective

hndgzkn: > maybe we should return early here too, no?

Good catch!

hndgzkn: > It would be better for debugging and interactive use if the figure stays open on, if it can't be saved.

May be we shall not close figure at all.

Remi-Gau: yeah I think so too

nilearn/plotting/displays/_slicers.py
```diff
@@ -1118,14 +1118,30 @@ def savefig(self, filename, dpi=None, **kwargs) -> None:
...
+            return
+
+        output_file = Path(filename)
+        try:
+            output_file.parent.mkdir(exist_ok=True, parents=True)
+        except Exception:
+            warnings.warn(
+                "Please specify a valid file path: {filename=}.",
+                stacklevel=find_stack_level(),
+            )
```
Remi-Gau: I don't think we cover those warnings in tests, no?

nilearn/plotting/_engine_utils.py
```diff
@@ -26,6 +29,32 @@
...
+    """
+    if output_file is not None:
+        output_file = Path(output_file)
+        try:
+            output_file.parent.mkdir(exist_ok=True, parents=True)
+        except Exception:
+            warn(
+                "Please specify a valid file path: {output_file=}.",
+                stacklevel=find_stack_level(),
+            )
```
Remi-Gau: similarly here, I don't think we cover this warning in tests, no?

Remi-Gau:
@hndgzkn 
will add the tests and merge

nilearn/plotting/displays/_figures.py
```diff
@@ -39,27 +42,15 @@ def show(self):
...
-
     def add_contours(self):
         """Draw boundaries around roi."""
         raise NotImplementedError
 
+    def _check_output_file(
+        self, output_file: OutputFile = None
+    ) -> Path | None:
+        return validate_output_file(output_file)
+
```
hndgzkn: Actually this function is unnecessary. It can be deleted completely.

nilearn/plotting/_utils.py
```diff
@@ -325,3 +326,29 @@ def check_threshold_not_negative(threshold) -> None:
     """
     if isinstance(threshold, (int, float)) and threshold < -1e-5:
         raise ValueError("Threshold should be a non-negative number!")
+
+
+def validate_output_file(output_file: str | Path | None) -> None | Path:
+    """Ensure that we have a valid output filename.
+
+    Create the output dir if necessary.
+    """
```
hndgzkn: If this function can be used else where in the code base, it can be put under `nilearn._utils`. 

Otherwise the logic is simple and only reused twice in the code. In that case, I would favor repetition of code instead of creating a new function.

nilearn/plotting/_utils.py
```diff
@@ -325,3 +327,29 @@ def check_threshold_not_negative(threshold) -> None:
...
+            f"Please specify a valid filename: '{output_file=}'.",
+            stacklevel=find_stack_level(),
+        )
+        return None
+
+    output_file = Path(output_file)
+
+    try:
+        output_file.parent.mkdir(exist_ok=True, parents=True)
+        return output_file
```
hndgzkn:
```suggestion
    try:
        output_file = Path(output_file)
        output_file.parent.mkdir(exist_ok=True, parents=True)
        return output_file
```

nilearn/plotting/_engine_utils.py
```diff
@@ -12,20 +12,42 @@
...
+
+    Parameters
+    ----------
+    fig:  :class:`matplotlib.figure.Figure`
+        figure to save
+
+    output_file: str, Path or None
+        file path to save the figure
+    """
+    output_file = validate_output_file(output_file)
```
hndgzkn: Using `validate_output_file` here changes the initial logic.

If the `output_file` specified was None, originally there was no warning as this function is called from plotting functions and while we are using plotting functions we do not generally want to save the image and seeing a warning each time might not be desired from the user's point of view.

nilearn/plotting/_utils.py
Remi-Gau: can defo be used to validate the filename when saving reports to disk

nilearn/plotting/_engine_utils.py
Remi-Gau: ah yeah let me return early if we pass None

nilearn/plotting/_utils.py
hndgzkn: But seeing its simplicity I do not really see the necessity of making it a separate function.

nilearn/_utils/param_validation.py
```diff
@@ -343,3 +344,34 @@ def sanitize_verbose(verbose: int | bool) -> int:
...
+        warnings.warn(
+            f"Please specify a valid filename: '{output_file=}'.",
+            stacklevel=find_stack_level(),
+        )
+        return None
+
+    check_is_of_allowed_type(
+        output_file, (str, Path), parameter_name=parameter_name
+    )
+
```
hndgzkn:
```suggestion
```

hndgzkn: I think this is unnecessary as if the below code cannot convert `output_file` to `Path`, it will raise exception and hence warning and return None.

nilearn/plotting/_engine_utils.py
```diff
@@ -26,6 +31,27 @@
...
+
+    output_file: str, Path or None
+        file path to save the figure
+    """
+    if output_file is None:
+        return None
+
+    output_file = validate_output_file(output_file)
+    if output_file is not None:
+        fig.savefig(output_file)
```
hndgzkn: I find it more clear if we let each function decide its own logic with try-except blocks. 

This code seems cyclic to me. We need to check many times if output_file is None.

nilearn/_utils/html_document.py
```diff
@@ -227,17 +230,20 @@ def _repr_mimebundle_(self, include=None, exclude=None):
...
         ----------
-        file_name : :obj:`str`
+        file_name : :obj:`str`, :obj:`pathlib.Path` or None
             Path to the HTML file used for saving.
 
         """
-        with Path(file_name).open("wb") as f:
+        file_name = validate_output_file(file_name, parameter_name="file_name")
+        if file_name is None:
+            return None
```
hndgzkn: Here again, I think the function could have its own logic, that would be more clear than validating a value, and checking it again.

mcsitter:
If you appreciate my 2 cents: just drop the complexity and assume paths are valid. If a path is `None`, handle that separately (e.g. return `None`). Otherwise, don’t try to be clever—just let a `FileNotFoundError` be raised if something’s wrong and let the user deal with it.

hndgzkn:
> If you appreciate my 2 cents: just drop the complexity and assume paths are valid. If a path is `None`, handle that separately (e.g. return `None`). Otherwise, don’t try to be clever—just let a `FileNotFoundError` be raised if something’s wrong and let the user deal with it.

It depends how we would like to define the behavior of the function. 

My perception is that, nilearn code generally tries to handle the problems in inputs and informs the user about the problem which I find nice. 

If we put a try-except block in the code to handle different possibilities, in my opinion does not make it more complex. For the case here, I just have doubts on the value of extraction of a new function for this check.

nilearn/plotting/_engine_utils.py
Remi-Gau: @hndgzkn 
I reverted my changes.

Can I let you take care of adding the tests so we cover the warnings that have been added?

hndgzkn:
> If you appreciate my 2 cents: just drop the complexity and assume paths are valid. If a path is `None`, handle that separately (e.g. return `None`). Otherwise, don’t try to be clever—just let a `FileNotFoundError` be raised if something’s wrong and let the user deal with it.

Finally I decided to follow your suggestion @mcsitter . Path is rather permissive, I could not find a reasonable test case a part from `None` and `PermissionError` and may some extreme cases. I am handling `None` and I think user can deal with PermissionError.

nilearn/plotting/tests/test_baseline_comparisons.py
```diff
@@ -400,7 +400,9 @@ def test_plot_surf_surface(plot_func, view, hemi):
     not (is_plotly_installed() and is_kaleido_installed()),
     reason="This test requires plotly and kaleido to be installed",
 )
-@pytest.mark.mpl_image_compare(tolerance=5)
+@pytest.mark.mpl_image_compare(
+    tolerance=5, savefig_kwargs={}, deterministic=False
+)
 @pytest.mark.parametrize("plot_func", SURFACE_FUNCS)
```
hndgzkn, line 406: This addition is necessary, as `savefig_kwargs` was introduced [here](https://github.com/nilearn/nilearn/pull/5638#discussion_r2355318452) and was not passed to `figure.write_image`. In this PR, it is passed. However mpl_image_compare sets "metedata" key in savefig_kwargs and as this key does not exist in plotly, this raises error in baseline figure generation. 

Checking `mpl_image_compare` [code](https://github.com/matplotlib/pytest-mpl/blob/f58f8b6d8edc49457ce6682f1284f04c89e8561d/pytest_mpl/plugin.py#L727), "metadata" key is only set when `deterministic=True`. Thus I set it to False here.

nilearn/plotting/displays/_figures.py
```diff
@@ -127,23 +111,29 @@ def show(self, renderer=None):
...
 
         Parameters
         ----------
         output_file : :obj:`str` or None, default=None
             Path to output file.
 
-        savefig_kwargs:
+        kwargs: :obj:`dict`
+            Extra keyword arguments are passed to
+            `plotly.graph_objects.Figure.write_image`
```
hndgzkn: writing this as :func:`plotly.graph_objects.Figure.write_image` does not create link to plotly API reference correctly. I try to see if I can solve this.

hndgzkn: Fixed with final update.

nilearn/plotting/displays/_figures.py
```diff
@@ -127,23 +111,29 @@ def show(self, renderer=None):
             else:
                 self.figure.show(renderer=renderer)
 
-    def savefig(self, output_file=None, **savefig_kwargs) -> None:  # noqa: ARG002
+    def savefig(self, output_file: OutputFile = None, **kwargs) -> None:
         """Save the figure to file.
 
         Parameters
         ----------
         output_file : :obj:`str` or None, default=None
```
Remi-Gau:
```suggestion
        %(output_file)s
```

nilearn/plotting/_engine_utils.py
```diff
@@ -26,6 +29,25 @@
...
+
+    Create output path if required.
+
+    Parameters
+    ----------
+    fig:  :class:`matplotlib.figure.Figure`
+        figure to save
+
+    output_file: str, Path or None
+        file path to save the figure
```
Remi-Gau:
```suggestion
    %(output_file)s
```

Remi-Gau:
minor doc string fixes suggested

it seems that in the code base we sometimes use `output_file` and sometimes use `filename`

it may be good to use only one of those and make sure it always has the same type so we can validate it with nilearn._utils.param_validation.heck_params

but I would suggest that this is done in a follow up PR

Remi-Gau (approved):
Good to go!!
thanks @hndgzkn

Issue #5824, opened 2025-11-06
Add more example sections for each function in the user API


Remi-Gau:
It was pointed to us at the recent cogbase conference that having a mini-example for more of our functions / methods would be useful as sometimes readers do not want to read the whole story of an example to figure out how to use a function.

Ideally we would like to have something that resembles more the API documentation of numpy, pandas...

https://numpy.org/doc/stable/reference/generated/numpy.eye.html

<img width="748" height="311" alt="Image" src="https://github.com/user-attachments/assets/d23a41b5-3a5f-4de6-8617-fd038c7de935" />

https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html

<img width="804" height="210" alt="Image" src="https://github.com/user-attachments/assets/c034ee81-c071-482a-bbee-9c06558850a0" />

---

The below script should at least help us flag what functions have no example.
Something similar should be done for Classes and their methods.

```python
import inspect
from nilearn.utils import all_functions
from rich import print

all = all_functions()
wont_do = []

for fn in all:
    if fn[0].startswith("fetch") or fn[0].startswith("load"):
        wont_do.append(fn[0])
        continue

    try:
        params = inspect.signature(fn[1]).parameters
        param_count = len(params)
    except (ValueError, TypeError):
        param_count = "?"

    if "Examples" not in fn[1].__doc__ and ">>>" not in fn[1].__doc__:
        print(f"- [ ] {fn[0]} ({param_count} params)")
    else:
        print(rf"- \[x] {fn[0]} ({param_count} params)")

print(f"\n\n\n### Wont do (for now):\n\n")
for fn in wont_do:
    print(f"- {fn}")

```

- [x] all_displays (1 params)
- [x] all_estimators (1 params)
- [x] all_functions (0 params)
- [x] apply_mask (5 params)
- [x] binarize_img (5 params)
- [x] butterworth (8 params)
- [x] check_design_matrix (1 params)
- [x] check_niimg (6 params)
- [x] check_niimg_3d (2 params)
- [x] check_niimg_4d (3 params)
- [ ] clean (14 params)
- [x] clean_img (11 params)
- [ ] cluster_level_inference (5 params)
- [x] compute_background_mask (8 params)
- [ ] compute_brain_mask (7 params)
- [x] compute_contrast (4 params)
- [x] compute_epi_mask (11 params)
- [ ] compute_fixed_effects (5 params)
- [ ] compute_multi_background_mask (10 params)
[...]

PR #6302, opened 2026-06-12
[DOC] Add example to `smooth_img`

hndgzkn:
- Related to #5824 

Use of AI tools:

- Add example to `nilearn.image.smooth_img`


nilearn/image/image.py
```diff
@@ -440,6 +440,27 @@ def smooth_img(imgs, fwhm):
...
 
+    Examples
+    --------
+    Let's first create a Nifti1Image::
+
+    >>> import numpy as np
+    >>> from nibabel import Nifti1Image
+    >>> data = np.array([[[0., 0.2, 0.8],
+    ...                   [1.5, 3.0, 0.1],
+    ...                   [0.4, 2.2, 0.0]]])
```
```suggestion
    >>> data = np.array([[[0.0, 0.0, 0.0],
    ...                   [0.0, 3.0, 0.0],
    ...                   [0.0, 0.0, 0.0]]])
```

Remi-Gau: @hndgzkn I was thinking that maybe we could have something with a single voxel with a value so this way it is easy to see that some smoothing happened: the original voxel value will be lower and all other voxesl should have the same final value

Remi-Gau (approved):

nilearn/image/image.py
```diff
@@ -440,6 +440,27 @@ def smooth_img(imgs, fwhm):
     Niimg-like object, :obj:~nilearn.surface.SurfaceImage.
     Smoothed input image or surface.
 
+    Examples
+    --------
+    Let's first create a Nifti1Image::
```
hndgzkn:
```suggestion
    Let's first create a Nifti1Image:
```

nilearn/image/image.py
```diff
@@ -440,6 +440,27 @@ def smooth_img(imgs, fwhm):
...
+    Let's first create a Nifti1Image:
+
+    >>> import numpy as np
+    >>> from nibabel import Nifti1Image
+    >>> data = np.array([[[0.0, 0.0, 0.0],
+    ...                   [0.0, 3.0, 0.0],
+    ...                   [0.0, 0.0, 0.0]]])
+    >>> img = Nifti1Image(data, affine=np.eye(4))
+
+    Now we can smooth the image::
```
hndgzkn:
```suggestion
    Now we can smooth the image:
```

PR #6318, opened 2026-06-13
[DOC] added example for intersect_masks

mbedini:
- Relates to #5824 

Use of AI tools:

- Added example for `intersect_masks`


nilearn/masking.py
```diff
@@ -211,6 +211,24 @@ def intersect_masks(mask_imgs, threshold=0.5, connected=True):
...
     grp_mask : 3D :class:`nibabel.nifti1.Nifti1Image`
         Intersection of all masks.
+
+    Examples
+    --------
+    >>> import nibabel as nib
+    >>> from nilearn import masking
+
+    >>> # Load binary masks
+    >>> img_a = nib.load("region_a_bin.nii.gz")
```
bthirion: This example doesn't run, right ? Is there a way to use  actual binary images (or create them on the fly) ?

nilearn/masking.py
```diff
@@ -212,6 +212,24 @@ def intersect_masks(mask_imgs, threshold=0.5, connected=True):
...
+    >>> img_b = nib.load("region_b_bin.nii.gz")
```
emdupre: It would be great to update this snippet s.t. it's runnable ! we'd need to change these into numpy arrays and convert into Nifti1Image's. Here's a snippet you could use, for example :

```
import numpy as np

# create a Nifti1Image
data1 = np.array([[[0., 1, 1],
                  [0, 1, 1],
                  [0, 1, 1]]])
affine = np.diag((3, 3, 3, 1))
mg1 = Nifti1Image(data1, affine, dtype=np.int32)

Remi-Gau:
unrelated to this PR: ultra rare failure of a test that I thought was not flaky anymore

https://github.com/nilearn/nilearn/actions/runs/27946223073/job/82706750650?pr=6318#step:8:13793

remarking it as flaky in this PR

Remi-Gau (approved):

Issue #6265, opened 2026-06-05
Do a read through of all the examples


Remi-Gau:
Examples often get modified during a pull request with small changes here and there.

But after a while those small changes can lead to examples that do not read as well as they originally did.

We should do a regular read through of the examples to make sure that still read well or if they should be reworked.

- [ ] do a read through of all the examples
- [ ] update release protocol to do a read through of all the modified examples since last release

bthirion:
It's a task that should be distributed across coredevs before releases.

Remi-Gau:
Well it looks like we have quite a bit of reading to do then...

```
git diff 0.13.1 HEAD --name-only examples
```

```
examples/00_tutorials/plot_decoding_tutorial.py
examples/01_plotting/plot_3d_map_to_surface_projection.py
examples/01_plotting/plot_carpet.py
examples/01_plotting/plot_demo_plotting.py
examples/01_plotting/plot_haxby_masks.py
examples/01_plotting/plot_surf_atlas.py
examples/01_plotting/plot_templateflow.py
examples/02_decoding/plot_haxby_anova_svm.py
examples/02_decoding/plot_haxby_different_estimators.py
examples/02_decoding/plot_haxby_frem.py
examples/02_decoding/plot_haxby_full_analysis.py
examples/02_decoding/plot_haxby_grid_search.py
examples/02_decoding/plot_haxby_multiclass.py
examples/02_decoding/plot_haxby_searchlight.py
examples/02_decoding/plot_haxby_understand_decoder.py
examples/02_decoding/plot_mixed_gambles_frem.py
examples/02_decoding/plot_miyawaki_encoding.py
examples/02_decoding/plot_miyawaki_reconstruction.py
examples/02_decoding/plot_oasis_vbm.py
examples/02_decoding/plot_oasis_vbm_space_net.py
examples/03_connectivity/plot_atlas_comparison.py
examples/03_connectivity/plot_compare_decomposition.py
examples/03_connectivity/plot_data_driven_parcellations.py
examples/03_connectivity/plot_extract_regions_dictlearning_maps.py
examples/03_connectivity/plot_group_level_connectivity.py
examples/03_connectivity/plot_inverse_covariance_connectome.py
examples/03_connectivity/plot_multi_subject_connectome.py
examples/03_connectivity/plot_probabilistic_atlas_extraction.py
examples/03_connectivity/plot_seed_to_voxel_correlation.py
examples/03_connectivity/plot_signal_extraction.py
[...]

PR #6337, opened 2026-06-19
[DOC] update carpet plot example

Remi-Gau:
- Relates to #6265

Use of AI tools:

- update carpet plot example


emdupre (approved):
LGTM, thanks !

examples/01_plotting/plot_carpet.py
```diff
@@ -29,46 +31,76 @@
...
+
 discrete_atlas_img = image.new_img_like(
     atlas_img, discrete_version.astype(np.float32)
 )
 
+plot_roi(
+    discrete_atlas_img,
+    cmap="tab10",
+    title="gray matter / white matter / cerebrospinal fluid masks",
+)
```
bthirion, line 111: Please note that the colors don't match btw the two figures.

emdupre: Ah, good catch ! It could be confusing to have this difference, agreed.

bthirion (commented):
In the rendering of the carpet plot, there is just a slight issue with "cerebrospinal fluid" that gets cut.
LGTM otherwise

Remi-Gau:
> In the rendering of the carpet plot, there is just a slight issue with "cerebrospinal fluid" that gets cut.

fixed

PR #6338, opened 2026-06-19
[MAINT] Add return type annotations to public functions in nilearn.image

Remi-Gau:
- Related to #6184

Use of AI tools:

- Add return type annotations to several public functions in `nilearn/image/image.py` that were missing them (`check_niimg_4d`, `concat_imgs`, `copy_img`, `index_img`, `iter_img`, `largest_connected_component_img`, `mean_img`, `new_img_like`, `smooth_img`, `swap_img_hemispheres`).
- `binarize_img`, `check_niimg`, `check_niimg_3d`, `clean_img`, `get_data`, `high_variance_confounds`, `load_img`, `math_img` and `threshold_img` already had return type annotations.
- Annotating `concat_imgs` made mypy check its body and surfaced two pre-existing bugs that are fixed here: mutating a `tuple` in place when converting a list of `SurfaceImage` inputs, and a call to `extract_data()` missing a required argument when looking up the dtype.
- Annotating `iter_img` surfaced an mypy issue in `nilearn/plotting/image/img_plotting.py` (narrowed with `cast`, since the image is forced to `Nifti1Image` before the call) and in a decomposition test that reused a loop variable across incompatible types.
- Added `@overload` signatures for `new_img_like`, `index_img`, `iter_img`, `mean_img`, `concat_imgs`, `largest_connected_component_img` and `smooth_img` so their return type is narrowed precisely (e.g. `SurfaceImage` vs `Nifti1Image`, or single image vs `list`) based on the input type, instead of always exposing the full union.
- This PR generated with the help of an LLM (Claude).

### DIFF
--- a/nilearn/decomposition/tests/test_dict_learning.py
+++ b/nilearn/decomposition/tests/test_dict_learning.py
@@ -1,5 +1,8 @@
+from typing import cast
+
 import numpy as np
 import pytest
+from nibabel import Nifti1Image
 
 from nilearn.decomposition.dict_learning import DictLearning
 from nilearn.decomposition.tests.conftest import (
@@ -8,6 +11,7 @@ from nilearn.decomposition.tests.conftest import (
 )
 from nilearn.image import get_data, iter_img
 from nilearn.maskers import NiftiMasker, SurfaceMasker
+from nilearn.surface import SurfaceImage
 from nilearn.surface.surface import get_data as get_surface_data
 
 
@@ -105,6 +109,7 @@ def test_dict_learning(
         assert recovered_maps >= 2
 
 
+@pytest.mark.ai_generated
 @pytest.mark.slow
 @pytest.mark.parametrize("data_type", ["nifti", "surface"])
 def test_component_sign(
@@ -130,5 +135,9 @@ def test_component_sign(
     check_decomposition_estimator(dict_learning, data_type)
 
     for mp in iter_img(dict_learning.components_img_):
-        mp = get_data(mp) if data_type == "nifti" else get_surface_data(mp)
-        assert np.sum(mp[mp <= 0]) <= np.sum(mp[mp > 0])
+        data = (
+            get_data(cast(Nifti1Image, mp))
+            if data_type == "nifti"
+            else get_surface_data(cast(SurfaceImage, mp))
+        )
+        assert np.sum(data[data <= 0]) <= np.sum(data[data > 0])
--- a/nilearn/decomposition/tests/test_dict_learning.py
+++ b/nilearn/decomposition/tests/test_dict_learning.py
@@ -108,7 +108,6 @@ def test_dict_learning(
         assert recovered_maps >= 2
 
 
-@pytest.mark.ai_generated
 @pytest.mark.slow
 @pytest.mark.parametrize("data_type", ["nifti", "surface"])
 def test_component_sign(
--- a/nilearn/decomposition/tests/test_dict_learning.py
+++ b/nilearn/decomposition/tests/test_dict_learning.py
@@ -11,7 +11,6 @@ from nilearn.decomposition.tests.conftest import (
 )
 from nilearn.image import get_data, iter_img
 from nilearn.maskers import NiftiMasker, SurfaceMasker
-from nilearn.surface import SurfaceImage
 from nilearn.surface.surface import get_data as get_surface_data
 
 
@@ -138,6 +137,6 @@ def test_component_sign(
         data = (
             get_data(cast(Nifti1Image, mp))
             if data_type == "nifti"
-            else get_surface_data(cast(SurfaceImage, mp))
+            else get_surface_data(mp)
         )
         assert np.sum(data[data <= 0]) <= np.sum(data[data > 0])
--- a/nilearn/image/image.py
+++ b/nilearn/image/image.py
@@ -455,8 +455,11 @@ def smooth_img(
 
     Returns
     -------
-    Niimg-like object, :obj:~nilearn.surface.SurfaceImage.
-    Smoothed input image or surface.
+    :obj:`~nibabel.nifti1.Nifti1Image`, :obj:`~nilearn.surface.SurfaceImage`, \
+        :obj:`list` of :obj:`~nibabel.nifti1.Nifti1Image`, or :obj:`list` \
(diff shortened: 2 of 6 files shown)


bthirion (approved):
LGTM, thx

Remi-Gau:
Will also add the `@overload` for this.

I may have found some use for this coding agent thing...

nilearn/image/image.py
```diff
@@ -418,8 +418,30 @@ def smooth_array(arr, affine, fwhm=None, ensure_finite=True, copy=True):
...
+def smooth_img(
+    imgs: list[NiimgLike] | tuple[NiimgLike, ...], fwhm
+) -> list[Nifti1Image]: ...
+
+
 @fill_doc
-def smooth_img(imgs, fwhm):
+def smooth_img(
+    imgs, fwhm
+) -> Nifti1Image | SurfaceImage | list[Nifti1Image] | list[SurfaceImage]:
```
Remi-Gau: should support niimg like

Remi-Gau: also check if having an iterable of images is not just simpler than going for list or tuples

nilearn/image/image.py
```diff
@@ -802,6 +824,28 @@ def _compute_surface_mean(imgs: SurfaceImage) -> SurfaceImage:
     return new_img_like(imgs, data=data)
 
 
+@overload
+def mean_img(
+    imgs: SurfaceImage | Iterable[SurfaceImage],
```
Remi-Gau, line 828: double check that it can handle an iterable of 1D surface image

nilearn/image/image.py
```diff
@@ -1075,7 +1135,7 @@ def iter_img(imgs):
             index_img(imgs, i) for i in range(at_least_2d(imgs).shape[1])
         )
         return output
-    return check_niimg_4d(imgs, return_iterator=True)
+    return iter(check_niimg_4d(imgs, return_iterator=True))
```
Remi-Gau: surprised this is is needed: an issue in the type annotation of check_niimg_4d?

nilearn/image/image.py
```diff
@@ -2222,6 +2302,30 @@ def load_img(
     return check_niimg(img, wildcards=wildcards, dtype=dtype)
 
 
+@overload
+def concat_imgs(
+    niimgs: list[SurfaceImage] | tuple[SurfaceImage, ...],
```
Remi-Gau: same here: can we just not have an iterable of surface images?

nilearn/decomposition/tests/test_dict_learning.py
```diff
@@ -130,5 +135,9 @@ def test_component_sign(
     check_decomposition_estimator(dict_learning, data_type)
 
     for mp in iter_img(dict_learning.components_img_):
-        mp = get_data(mp) if data_type == "nifti" else get_surface_data(mp)
-        assert np.sum(mp[mp <= 0]) <= np.sum(mp[mp > 0])
+        data = (
+            get_data(cast(Nifti1Image, mp))
```
Remi-Gau, line 137: that seems like a very ugly use of cast: see if we can get rid of it?

nilearn/image/image.py
```diff
@@ -2311,12 +2415,13 @@ def concat_imgs(
         if len(niimgs) == 1:
             return niimgs[0]
 
+        niimgs = list(niimgs)
         for i, img in enumerate(niimgs):
             check_polymesh_equal(img.mesh, niimgs[0].mesh)
             niimgs[i] = at_least_2d(img)
 
         if dtype is None:
-            dtype = extract_data(niimgs[0]).dtype
```
Remi-Gau, line 2319: unrelated to this PR but extract_data seems to be only used by index_img, so it should probably be moved into this module and made private

also there is an issue with how index_img is tested for surface images: we use the implementation of index_img for surface images (meaning: extract_data) to determine what is the expected value of those tests: this sounds tautological and unlikely to catch any problem

nilearn/maskers/base_masker.py
```diff
@@ -477,7 +477,7 @@ def _load_mask(self, imgs) -> None | Nifti1Image:
         tmp = check_niimg(self.mask_img, atleast_4d=True)
         mask_data = safe_get_data(tmp, ensure_finite=True)
         mask = mask_data.astype(bool).all(axis=3)
-        mask_img_ = new_img_like(self.mask_img, mask)
+        mask_img_ = new_img_like(cast(Nifti1Image, self.mask_img), mask)
```
Remi-Gau: is this cast necessary?

nilearn/image/image.py
```diff
@@ -2343,7 +2448,10 @@ def concat_imgs(
 
     iterator, literator = itertools.tee(iter(niimgs))
     try:
-        first_niimg = check_niimg(next(literator), ensure_ndim=ndim)
+        first_niimg = check_niimg(
+            next(literator),
+            ensure_ndim=ndim,  # type: ignore[arg-type]
```
Remi-Gau, line 2452: are those ignore necessary?

nilearn/plotting/image/img_plotting.py
```diff
@@ -1210,7 +1211,7 @@ def plot_prob_atlas(
     for map_img, color, thr in zip(
         iter_img(maps_img), color_list, threshold, strict=False
     ):
-        data = get_data(map_img)
+        data = get_data(cast(Nifti1Image, map_img))
```
Remi-Gau: is this cast necessary?

nilearn/image/image.py
Remi-Gau: fixed: was due to a bad annotation of check_niimg

nilearn/image/image.py
Remi-Gau: for another PR

nilearn/decomposition/tests/test_dict_learning.py
```diff
@@ -105,6 +108,7 @@ def test_dict_learning(
         assert recovered_maps >= 2
 
 
+@pytest.mark.ai_generated
```
Remi-Gau:
```suggestion
```

nilearn/image/image.py
Remi-Gau: yes...
... for now

PR #6351, opened 2026-06-24
[REF] move extract_data

### LINKED ISSUE #6346: refactor extract_data for surface data
unrelated to this PR but extract_data seems to be only used by index_img, so it should probably be moved into this module and made private

also there is an issue with how index_img is tested for surface images: we use the implementation of index_img for surface images (meaning: extract_data) to determine what is the expected value of those tests: this sounds tautological and unlikely to catch any problem

_Originally posted by @Remi-Gau in https://github.com/nilearn/nilearn/pull/6338#discussion_r3452233015_

Remi-Gau:
- Closes #6346

Use of AI tools:

- move extract_data to nilearn.image
- avoid using it to test the function where it's used
- rename the function as it is now private


bthirion (approved):
LGTM, thx

PR #6354, opened 2026-06-26
[MAINT] hoist surface classes to top of module

Remi-Gau:
- Closes none

Use of AI tools:

- move all classes in nilearn.surface.surface to top of module to be able to use them as reference any where below in the rest of the module


nilearn/surface/surface.py
Remi-Gau, line 1: the diff is huge but the classes were only moved

Issue #3469, opened 2023-01-18
Improve documentation of nilearn support for nifti images


ymzayek:
After a discussion during the drop-in hours about #3462, it seems important to better document that in nilearn only nifti images are thoroughly tested and that non-nifti images are less supported. This could include throwing a warning when a non-nifti image is used when loaded with `image.load_img` and documenting that use of non-nifti can lead to unexpected behavior

### Affected pages
Possibly https://nilearn.github.io/stable/manipulating_images/input_output.html

@effigies I wonder if you have any input on this.

Remi-Gau:
While adding return-type annotations to `nilearn.image.image.check_niimg` (as part of #6184), I investigated how it (and related core functions) actually behave with non-Nifti `nibabel.spatialimages.SpatialImage` subtypes — sharing the findings here since they're directly relevant to this issue.

## What works today

`check_niimg` delegates to `nilearn._utils.niimg.load_niimg`, which only checks `isinstance(niimg, spatialimages.SpatialImage)` — it never coerces the result to `Nifti1Image`. I tested every code path (plain pass-through, `atleast_4d=True` 3D→4D conversion, `ensure_ndim=3` squeeze-from-4D, list/iterable input via `concat_imgs`, and `return_iterator=True`) with `Nifti2Image`, `MGHImage`, `AnalyzeImage`, `Spm99AnalyzeImage`, and `Spm2AnalyzeImage` — all of them work, and the original concrete class is always preserved (e.g. passing a `Nifti2Image` in gets a `Nifti2Image` back, not a `Nifti1Image`). Same story for `nilearn.image.get_data`, `_get_data`, and `safe_get_data` — all duck-type on `SpatialImage`'s shared interface (`.affine`, `._dataobj`, `.in_memory`, etc.) rather than requiring `Nifti1Image` specifically.

So at this low level, support for other `SpatialImage` formats is broader than documented.

## Where it breaks down

[...]

PR #6357, opened 2026-06-29
[ENH] Warn when check_niimg loads a non-Nifti1Image SpatialImage

Remi-Gau:
Related to #3469

Use of AI tools:

- While annotating return types in `nilearn.image.image.check_niimg` (#6184), I found that `check_niimg`/`check_niimg_3d`/`check_niimg_4d` (via `nilearn._utils.niimg.load_niimg`) already accept and preserve any `nibabel.spatialimages.SpatialImage` subtype at runtime, not only `Nifti1Image`. I posted a detailed investigation as a comment on #3469.
- `load_niimg` now emits a `UserWarning` when the loaded image is not a `Nifti1Image`, since most of Nilearn is only thoroughly tested with `Nifti1Image` objects and downstream behavior may be affected otherwise.
  - `Nifti2Image` is exempted: it is a genuine subclass of `Nifti1Image` in nibabel and shares the same tested code path, so it would be a false positive.
  - Internal nilearn sentinel classes (e.g. the lazy `_MNI152Template` used as the default background image in plotting functions) are exempted by leading-underscore class name, since warning about nilearn's own internal plumbing isn't useful to users. This was found via a real test failure (`test_nifti_label_masker_brainsprite`) before being excluded.
- Added a shared `%(non_nifti_image_admonition)s` docstring snippet in `nilearn._utils.docs` (same pattern as the existing `sk_compatible_admonition`), injected into `check_niimg`, `check_niimg_3d`, `check_niimg_4d` and `load_niimg` docstrings.
- Added the missing `@fill_doc` decorator to `load_niimg`, fixing a pre-existing bug where its `%(dtype)s` placeholder was never substituted.
- Added parametrized tests covering `MGHImage`/`AnalyzeImage` (should warn) and `Nifti1Image`/`Nifti2Image` (should not warn) across `check_niimg`, `check_niimg_3d` and `check_niimg_4d`.
- This PR generated with the help of an LLM (Claude).

### DIFF
--- a/nilearn/_utils/niimg.py
+++ b/nilearn/_utils/niimg.py
@@ -8,6 +8,7 @@ from warnings import warn
 import numpy as np
 from nibabel import Nifti1Image, is_proxy, load, spatialimages
 
+from nilearn._utils.docs import fill_doc
 from nilearn._utils.helpers import stringify_path
 from nilearn._utils.logger import find_stack_level
 from nilearn._utils.numpy_conversions import get_target_dtype
@@ -97,6 +98,7 @@ def ensure_finite_data(
     return data
 
 
+@fill_doc
 def load_niimg(niimg, dtype=None):
     """Load a niimg, check if it is a nibabel SpatialImage and cast if needed.
 
@@ -105,6 +107,8 @@ def load_niimg(niimg, dtype=None):
     niimg : Niimg-like object
         See :ref:`extracting_data`.
         Image to load.
+
+        %(non_nifti_image_admonition)s
 
     %(dtype)s
 
@@ -138,6 +142,19 @@ def load_niimg(niimg, dtype=None):
             )
             if copy_header:
                 niimg.header.set_data_dtype(target_dtype)
+
+    if not isinstance(
+        niimg, Nifti1Image
+    ) and not niimg.__class__.__name__.startswith("_"):
+        warn(
+            f"Loaded image is a {niimg.__class__.__name__}, "
+            "not a Nifti1Image. Most of Nilearn is only thoroughly "
+            "tested with Nifti1Image objects, "
+            "so downstream behavior may be affected. "
+            "See https://github.com/nilearn/nilearn/issues/3469 "
+            "for more information.",
+            stacklevel=find_stack_level(),
+        )
 
     return niimg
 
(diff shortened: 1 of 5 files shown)


nilearn/_utils/niimg.py
```diff
@@ -139,6 +143,19 @@ def load_niimg(niimg, dtype=None):
             if copy_header:
                 niimg.header.set_data_dtype(target_dtype)
 
+    if not isinstance(
+        niimg, Nifti1Image
+    ) and not niimg.__class__.__name__.startswith("_"):
+        warn(
```
Remi-Gau, line 149: I wonder if we should cast MGHImage and AnalyzeImage to NiftiImage make things easier in general.

@effigies 
This PR aims to better explain what nilearn 'means' by 'nifti like' image. Our loading and checking functions of images will actually return MGHImage and AnalyzeImage when those are passed (or read from disk).

Are there any 'side effects' we can expect, it we cast MGHImage and AnalyzeImage to NiftiImage?

bthirion: Coercing to NiftiImage would indeed be a great simplification, but I'm not aware of side effects.

effigies: You should be able to use `Nifti1Image.from_image(img)` and get nibabel's best-effort mapping of metadata between formats. This is not something that has been at the top of anybody's priority list to get a complete mapping, so there will probably be room for improvement.

For `AnalyzeImage` and its subclasses, the mapping should be pretty good, since NIfTI was intended to preserve the meaning of all Analyze fields anybody cared about.

For `MGHImage` and other images, I would probably raise a warning if you're going to do it implicitly. The affine (up to floating point rounding error) and data block should be correct, but there may be missing metadata.

nilearn/_utils/niimg.py
```diff
@@ -139,6 +143,19 @@ def load_niimg(niimg, dtype=None):
             if copy_header:
                 niimg.header.set_data_dtype(target_dtype)
 
+    if not isinstance(
+        niimg, Nifti1Image
+    ) and not niimg.__class__.__name__.startswith("_"):
```
effigies, line 148: Out of curiosity, what is the `_` case you're considering here? Just a way for someone to opt out of this with `class _MyMGHImage(nibabel.MGHImage): pass`?

Issue #6197, opened 2026-05-04
refactor html template for surface figures


Remi-Gau:
> further refactoring can be done to use the same head.jinja as for all other HTML documents but that can wait another PR 

 _Originally posted by @Remi-Gau in [#3729](https://github.com/nilearn/nilearn/pull/3729/changes#r3087438522)_

hndgzkn:
@Remi-Gau Can I work on this issue?

Remi-Gau:
of course

hndgzkn:
I am thinking of adding a baseline test for `view_surf` with `engine=niivue`.

Remi-Gau:
yes go ahead

PR #6359, opened 2026-06-30
[MAINT] add reference image for view_surf with niivue engine

hndgzkn:
- Relates to #6197 

Use of AI tools:

- Add reference image for `view_surf` function with `engine=niivue`


tests/js/references/view_surf_niivue_reference.png
hndgzkn, line 1: The color map is reversed compared to the image generated with `plotly_engine`:

Image

When I reverse the color map, I get:

Image

Remi-Gau: ok I think we can file that under 'bug to fix'

hndgzkn:
@Remi-Gau I think it might be better to : 

- merge this PR
- fix the bug of colormap in a new PR 
- then reuse head.jinja in a new PR so that we can identify the changes introduced

How does it sound to you?

hndgzkn:
The image generated in test html output for view_surf_niivue seems to be not complete.

@Remi-Gau Do you have any idea why this might be the case?

 
view_surf_niivue

Remi-Gau:
> * merge this PR
> 
>     * fix the bug of colormap in a new PR
> 
>     * then reuse head.jinja in a new PR so that we can identify the changes introduced
> 
> 
> How does it sound to you?

yup sounds good to me

Remi-Gau:
> @Remi-Gau Do you have any idea why this might be the case?

not sure 

let me check

tests/js/view.test.js
Remi-Gau: maybe increasing the timeout could give more time for the image to load in CI?

hndgzkn: Ok, I try that, thank you!

tests/js/template.js
```diff
@@ -30,6 +30,8 @@ module.exports.fullTest = (file, clip, tolerance = 1000, timeout = 5000) => {
       async () => {
         // take a screenshot of the page
         const fileCurrent = buildFilePNG(file, '', '')
+        setTimeout(() => {
+        }, 3000)
```
hndgzkn:
```suggestion
```

tests/js/view.test.js
Remi-Gau: apparently that worked

Remi-Gau (approved):

PR #6369, opened 2026-07-01
[MAINT] Add return type annotations to public functions in nilearn.regions

### LINKED ISSUE #6184: Improve type hints for conditional return types (reduce IDE/type checker warnings)

Remi-Gau:
- Closes #6184

Use of AI tools:

- Add `-> tuple[int, np.ndarray]` to `recursive_neighbor_agglomeration` in `nilearn/regions/rena_clustering.py`
- Add `-> tuple[Nifti1Image, list[int]] | tuple[None, None]` to `connected_regions` in `nilearn/regions/region_extractor.py`
- Add `-> Nifti1Image | tuple[Nifti1Image, list[str]]` to `connected_label_regions` in `nilearn/regions/region_extractor.py`
- Add `-> tuple[np.ndarray, list] | tuple[np.ndarray, list, Nifti1Image]` to `img_to_signals_labels` in `nilearn/regions/signal_extraction.py`
- Add `-> Nifti1Image` to `signals_to_img_labels` in `nilearn/regions/signal_extraction.py`
- Add `-> tuple[np.ndarray, list[int]]` to `img_to_signals_maps` in `nilearn/regions/signal_extraction.py`
- Add `-> Nifti1Image` to `signals_to_img_maps` in `nilearn/regions/signal_extraction.py`

### DIFF
--- a/nilearn/regions/signal_extraction.py
+++ b/nilearn/regions/signal_extraction.py
@@ -324,8 +324,9 @@ def img_to_signals_labels(
         Corresponding labels for each signal. signal[:, n] was extracted from
         the region with label labels[n].
 
-    masked_atlas : Niimg-like object
+    masked_atlas : :class:`nibabel.nifti1.Nifti1Image`
         Regions definition as labels after applying the mask.
+        Only returned when ``return_masked_atlas=True``.
 
     See Also
     --------
@@ -541,7 +542,7 @@ def img_to_signals_maps(
         Signals extracted from each region.
         Shape is: (scans number, number of regions intersecting mask)
 
-    labels : :obj:`list`
+    labels : :obj:`list` of :obj:`int`
         maps_img[..., labels[n]] is the region that has been used to extract
         signal region_signals[:, n].
 
--- a/nilearn/regions/signal_extraction.py
+++ b/nilearn/regions/signal_extraction.py
@@ -235,7 +235,7 @@ def img_to_signals_labels(
     keep_masked_labels=False,
     return_masked_atlas=True,
     n_jobs=1,
-):
+) -> tuple[np.ndarray, list] | tuple[np.ndarray, list, Nifti1Image]:
     """Extract region signals from image.
 
     This function is applicable to regions defined by labels.
@@ -364,7 +364,7 @@ def img_to_signals_labels(
 
 def signals_to_img_labels(
     signals, labels_img, mask_img=None, background_label=0, order="F"
-):
+) -> Nifti1Image:
     """Create image from region signals defined as labels.
 
     The same region signal is used for each :term:`voxel` of the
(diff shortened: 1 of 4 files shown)


Remi-Gau:
some overloads need to be clarified here

bthirion (commented):
LGTM so far.

nilearn/regions/signal_extraction.py
```diff
@@ -224,6 +225,34 @@ def _get_labels_data(
...
+    labels_img,
+    mask_img=...,
+    background_label=...,
+    order=...,
+    strategy=...,
+    keep_masked_labels=...,
+    return_masked_atlas: Literal[False] = ...,
+    n_jobs=...,
+) -> tuple[np.ndarray, list]: ...
+
```
bthirion, line 254: Are the two possible outputs really needed ?

bthirion (commented):
LGTM.

nilearn/regions/signal_extraction.py
Remi-Gau: we have a deprecation for this: in nilearn 0.15 it will always return: np.ndarray, list, Nifti1Image

PR #6379, opened 2026-07-03
[MAINT] add type annotations to nilearn.datasets

Remi-Gau:
- Closes none but relates to #3568

Use of AI tools:

- add type annotation to the public functions in struct, func, neurovault in nilearn.datasets

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -12,7 +12,7 @@ import re
 import warnings
 from io import BytesIO
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
 
 import numpy as np
 import pandas as pd
@@ -2126,7 +2126,7 @@ def fetch_development_fmri(
     data_dir=None,
     resume=True,
     verbose=1,
-    age_group="both",
+    age_group: Literal["adults", "child", "both"] = "both",
 ):
     """Fetch movie watching based brain development dataset (fMRI).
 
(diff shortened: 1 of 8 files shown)


nilearn/datasets/func.py
```diff
@@ -368,12 +384,14 @@ def fetch_adhd(n_subjects=30, data_dir=None, url=None, resume=True, verbose=1):
     )[0]
 
     # Load the csv file
-    phenotypic = pd.read_table(phenotypic, delimiter=",")
-    phenotypic["Subject"] = phenotypic["Subject"].apply(lambda x: f"{x:07d}")
+    phenotypic_df = pd.read_table(phenotypic, delimiter=",")
```
Remi-Gau, line 386: had to rename certain variables to avoid some static typing errors

bthirion (approved):
LGTM, AFAICT.
But the CI is a bit redish :-(

Remi-Gau:
> But the CI is a bit redish :-(

ha, good_
that shows that tests were likely more thorough than the docstring: let me check

nilearn/datasets/func.py
```diff
@@ -42,20 +44,25 @@
...
 def fetch_haxby(
-    data_dir=None,
-    subjects=(2,),
-    fetch_stimuli=False,
-    url=None,
-    resume=True,
-    verbose=1,
-):
+    data_dir: DataDir = None,
+    subjects: int | list[int] | tuple[int] = (2,),
```
Remi-Gau, line 60: no a big thing but we got either `n_subjects` (9 times) or `subjects` (once) used in this module with almost same type and repeated logic in several places to validate it: I think having a single function for all fetchers would simplify things

nilearn/datasets/func.py
hndgzkn, line 2817: Here it should return str(dataset_dir) instead of str(data_dir); causes failure in CI at the moment.

Remi-Gau: ha yes: thanks

PR #6381, opened 2026-07-05
[DOC] Update `plot_mask_computation.py` example

michellewang:
- Relates to #6265

Use of AI tools:

- Minor changes to text and formatting for better readability


examples/06_manipulating_images/plot_mask_computation.py
Remi-Gau:
```suggestion
```

I think that outputting twice the HTML representation of the estimator does not add much:

image

Remi-Gau:
```suggestion
# The default strategy to compute a mask
```

Remi-Gau:
```suggestion
# Here we fetch the movie watching based brain development fMRI dataset
```

Remi-Gau:
```suggestion
# and a different strategy is necessary.
```

Remi-Gau, line 227: I am getting a "no overlay found" error when hoverring: this may be a bug on our side

https://output.circle-artifacts.com/output/job/ec52ebf2-4730-48af-ab88-923ea205a693/artifacts/0/dev/auto_examples/06_manipulating_images/plot_mask_computation.html#compute-and-resample-a-mask

Remi-Gau:
```suggestion
report = masker.generate_report(engine="brainsprite")
```

maybe we could use the brainsprite engine on this one and explain how it can allow an interactive exploration of the mask in the report

examples/06_manipulating_images/plot_mask_computation.py
```diff
@@ -243,7 +243,7 @@
 # After mask computation: extracting time series
 # ----------------------------------------------
 #
-# We extract time series detrended and non-detrended.
+# We extract detrended and non-detrended time series.
 trended_data = NiftiMasker(mask_strategy="epi", verbose=1).fit_transform(
```
Remi-Gau:
```suggestion
trended_data = NiftiMasker(mask_strategy="epi", verbose=1, standardize="zscore_sample").fit_transform(
```

This would be to avoid any of the warnings:

boolean values for 'standardize' will be deprecated in nilearn 0.15.0.
Use 'zscore_sample' instead of 'True' or use 'None' instead of 'False'.

examples/06_manipulating_images/plot_mask_computation.py
Remi-Gau: In general I wonder if the extra verbosity of the output does not distract from the things we want to show: so let's just rely on the default verbosity=0 ?

Maybe we can have it for a single one in this example: maybe the last one where we actually extract data?

bthirion: agreed

examples/06_manipulating_images/plot_mask_computation.py
bthirion: Yes, but you simply need to reorganize this with one  line per argument.

bthirion (commented):
Looks good. I think you can implement your suggestions in this PR ?

Remi-Gau:
The brainsprite masks in the report look all white: 

https://output.circle-artifacts.com/output/job/582d2686-e3b5-4117-b92e-e0c004008b2d/artifacts/0/dev/auto_examples/06_manipulating_images/plot_mask_computation.html#simple-mask-extraction-from-epi-images

image

@michellewang can you confirm this also happens when you run the example locally?
if that's the case, let's drop the idea of having brainsprite for the example for now and let's open an issue to investigate.

michellewang:
@Remi-Gau I get this when I try to build the example locally

Screenshot 2026-07-09 at 13 05 35

which is not blank but definitely not right.

However, if I just run the example cell by cell in an interactive window in VS Code (+ `open_in_browser()`), I get something else:

Screenshot 2026-07-09 at 13 09 58

But neither of those have the green outline that is in the static image in the [docs](https://nilearn.github.io/dev/auto_examples/06_manipulating_images/plot_mask_computation.html) currently, so I'll revert to the default engine instead of brainsprite

Screenshot 2026-07-09 at 13 12 17

michellewang:
Actually: I do get the completely blank mask in the other reports, just not the first one

Screenshot 2026-07-09 at 13 15 37

So it does seem like there's some sort of bug with the brainsprite engine

Remi-Gau:
OK good
thanks for checking and reverting: let's merge and open issues to track these new bugs

PR #6311, opened 2026-06-13
[DOC] Add example to `butterworth`

hndgzkn:
- Related to #5824 

Use of AI tools:
- [ ] LLM tools were used to generate at least part of the content of this pull-request.
- [x] No LLM tools were used to generate any part of the content of this pull-request.

Changes proposed in this pull request:

- Add example to `nilearn.signal.butterworth`

PR #6383, opened 2026-07-06
[DOC] only run doctests when plotting dependencies are installed

Remi-Gau:
- Closes none but relates to https://github.com/nilearn/nilearn/pull/6311#discussion_r3519830010

Use of AI tools:

- doctests will not be collected by default on all runs of pytests
- they are only collected and run when running `tox run -e plotting` that ensures that the plotting dependencies are installed
- adapt doctests that required maplotlib to ensure that this works.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -347,7 +347,7 @@ module = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-ra --strict-config --strict-markers --doctest-modules --showlocals --capture=no -v --template=maint_tools/templates/index.html"
+addopts = "-ra --strict-config --strict-markers --showlocals --capture=no -v --template=maint_tools/templates/index.html"
 doctest_optionflags = "NORMALIZE_WHITESPACE ELLIPSIS"
 # see https://docs.pytest.org/en/stable/how-to/capture-warnings.html#controlling-warnings
 filterwarnings = [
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -347,7 +347,7 @@ module = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-ra --strict-config --strict-markers --doctest-modules --showlocals --capture=no -v --template=maint_tools/templates/index.html"
+addopts = "-ra --strict-config --strict-markers --showlocals --capture=no -v --template=maint_tools/templates/index.html"
```
Remi-Gau, line 353: should I add a comment explaining why we are not running doctests by default?

hndgzkn: That might be good.

tox.ini
```diff
@@ -77,8 +77,12 @@ commands =
 
 
 [testenv:plotting]
-description = Run tests on latest version of all dependencies.
+description = Run tests on latest version of all dependencies. Also run the doctests on the python modules.
```
hndgzkn:
```suggestion
description = Run tests on latest version of all dependencies. Also run the doctests in the python modules.
```

hndgzkn (approved):
LGTM with minor suggestion.

emdupre:
I see you're well-ahead of my suggestion in #6311 😸 
The changes look good to me ; I would just add a comment about why this is done.

emdupre (approved):

pyproject.toml
Remi-Gau: done!

PR #6389, opened 2026-07-06
[FIX] Fix return value for `fetch_openneuro_dataset`

### LINKED ISSUE #6388: Return variable not updated while renaming data_dir to dataset_dir
> Here it should return str(dataset_dir) instead of str(data_dir); causes failure in CI at the moment. 

 _Originally posted by @hndgzkn in [#6379](https://github.com/nilearn/nilearn/pull/6379/changes#r3529278409)_

hndgzkn:
- Closes #6388 

Use of AI tools:

- Fix return value for `nilearn.datasets.func.fetch_openneuro_dataset`

### DIFF
--- a/nilearn/datasets/func.py
+++ b/nilearn/datasets/func.py
@@ -2814,7 +2814,7 @@ def fetch_openneuro_dataset(
 
     patch_openneuro_dataset(downloaded)
 
-    return str(data_dir), sorted(downloaded)
+    return str(dataset_dir), sorted(downloaded)
 
 
 @fill_doc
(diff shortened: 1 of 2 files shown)


nilearn/datasets/func.py
```diff
@@ -2814,7 +2814,7 @@ def fetch_openneuro_dataset(
 
     patch_openneuro_dataset(downloaded)
 
-    return str(data_dir), sorted(downloaded)
+    return str(dataset_dir), sorted(downloaded)
```
Remi-Gau, line 2817: The fact that this was not caught in the PR, but only on main makes me think we should add a regression test for this, no?

Or did I miss a failing test in the PR?

hndgzkn: > Or did I miss a failing test in the PR?

I guess so. I see it in many PRs now, and in local as well. The failing test is `test html output`.

Adding a regression test might be good, I'll try to add one.

Remi-Gau: thanks!! you're the best!!

PR #6399, opened 2026-07-07
[MAINT] use absolute import

Remi-Gau:
- Closes none

Use of AI tools:

- use absolute import

Just read that this is what our doc say we should do and we did not do it everywhere.

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -444,6 +444,7 @@ select = [
     "PLR",
     "RUF",
     "SIM",
+    "TID",
     "UP",
     "W"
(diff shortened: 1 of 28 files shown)


pyproject.toml
```diff
@@ -444,6 +444,7 @@ select = [
     "PLR",
     "RUF",
     "SIM",
+    "TID",
```
Remi-Gau, line 447: should help catch some relative imports
https://docs.astral.sh/ruff/rules/relative-imports/

PR #6402, opened 2026-07-07
[FIX] Fix color map for view_surf when `engine=niivue`

### LINKED ISSUE #6375: The color map for view_surf with engine niivue is reversed compared to the image generated with plotly engine
The color map is reversed compared to the image generated with `plotly_engine`:

Image

When I reverse the color map, I get:

Image

_Originally posted by @hndgzkn in https://github.com/nilearn/nilearn/pull/6359#discussion_r3499214641_

hndgzkn:
- Closes #6375 

Use of AI tools:

- Remove code that reverses already reversed color map.


Remi-Gau:
@hndgzkn I will let you deal with the merge conflict

Remi-Gau:
actually it seems that the test HTML is still failing (at least on main) because of the image loading problem

maybe we should increase the timeout even further?

hndgzkn:
> actually it seems that the test HTML is still failing (at least on main) because of the image loading problem
> 
> maybe we should increase the timeout even further?

Yes, it fails here as well :(

hndgzkn:
> actually it seems that the test HTML is still failing (at least on main) because of the image loading problem
> 

Seems to be fixed now.

Remi-Gau:
> Seems to be fixed now.

indeed let's merge and see how stable that is

Remi-Gau (approved):

PR #6403, opened 2026-07-07
[MAINT] group dependabot PR and autoupdate JS dependencies

Remi-Gau:
- Closes none

Use of AI tools:

- update dependabot config to have single PR to update github action rather than one PR per action (see for example https://github.com/bids-standard/bids-website/pull/865)
- also add updating Javascript dependencies via dependabot: to be seen if that creates some issues once merged


Remi-Gau:
for this one as well the effects will not be seen until merged so I will go ahead and see

PR #6410, opened 2026-07-08
[MAINT] Add return type annotations to nilearn.surface classes

Remi-Gau:
- Related to #3568

Use of AI tools:

- Add return type annotations to the public methods of `FileMesh`, `InMemoryMesh`, `PolyData`, `PolyMesh`, `SurfaceImage`, and `SurfaceMesh` in `nilearn.surface.surface`.
- Annotating `SurfaceImage.__init__` and `PolyData`'s properties made mypy check their bodies for the first time (they were previously fully untyped), which surfaced a couple of latent type inconsistencies, fixed here as well:
  - `FileMesh.coordinates`/`faces` are read-only properties overriding a writable attribute declared on the `SurfaceMesh` base class; this is an intentional design choice so the override is silenced explicitly with `# type: ignore[override]`.
  - `get_data()` (in `nilearn.surface.surface`) and `_compute_surface_mean()` (in `nilearn.image.image`) each reused a single variable name across incompatible types; renamed/annotated to resolve.
- Add a changelog entry.

CI will run the full test suite; letting that catch anything a local subset run might miss.


nilearn/surface/surface.py
```diff
@@ -46,10 +46,22 @@ class SurfaceMesh(abc.ABC):
     # TODO those are properties are for compatibility with plot_surf_img.
```
Remi-Gau, line 46: see if we can deal with this TODO in this PR

bthirion (commented):
LGTM. The CI failures seem unrelated.

Remi-Gau:
> LGTM. The CI failures seem unrelated.

indeed

Issue #5966, opened 2026-01-21
[META][PERF] Performance issues, docs, benchmarking and improvements


man-shu:
### Issues
Several performance-related issues have been reported in Nilearn; these can be filtered using the "Performance" label here: https://github.com/nilearn/nilearn/issues?q=is:issue%20state:open%20label:Performance. But here is a list to keep track of things:
- #719 
- #1362 
- #1754
- #1883
- #2209
- #3177 
- #3592 
- #4811
- #5588
- https://neurostars.org/t/problem-with-smoothing-in-first-level-analysis/29553

### Documentation
There have also been efforts to understand/explain these issues. These are mostly the issues/PRs where we have used profilers to look at memory usage under-the-hood. So far [memory_profiler](https://github.com/pythonprofilers/memory_profiler) has been most useful because it can be used within ipython/within scripts/via command line and also allows tracking parallel processes:
- #3398
- #3399
- #3401 
- #4811
- We have a high-level example comparing different strategies to do masking in parallel: #5113 and #5185
- We noticed from the above example that memory usage in parallel depended on how the image is initially loaded. So we started to understand and explain the differences between nilearn vs. nibabel loading functions and consequently proxy vs. array images in #5128. At the time of writing this issue, in this PR, we are stuck at why the actual memory usage multiplier is higher than the number of parallel processes.

### Performance benchmarking
It also makes sense to track both computation time and memory usage over time, flag any regressions and have a way to reliably test proposed performance improvements. For now we are following sklearn (https://scikit-learn.org/dev/developers/contributing.html#monitoring-performance) in using [asv](https://asv.readthedocs.io/en/latest/) for this purpose
- #4573 
- Initial placeholder benchmarks were implemented in #5280 and the code lives in [asv_benchmarks](https://github.com/nilearn/nilearn/tree/main/asv_benchmarks) subdirectory.
- Reports from these benchmarks are deployed here: https://nilearn.github.io/benchmarks/
- Now we want to reorganize [asv_benchmarks](https://github.com/nilearn/nilearn/tree/main/asv_benchmarks) submodule-wise in #5590. 
- It could also be worthwhile to also have a distinction between benchmarks that are light on resources so that they can be run on regular intervals to flag any regressions vs. the ones that are resource-intensive and should/could only be run on specific PRs or locally.

### Improvements
[...]

PR #6422, opened 2026-07-10
[DOC] Document and verify asv benchmark workflow, add local-import benchmark example

### LINKED ISSUE #5348: [ENH] Find a way to "stabilize" minor OS version changes/ RAM size for benchmarking on GitHub CI
Not super urgent, but currently, when we run our benchmarks on the GitHub CI, small changes in the version of the OS or RAM size lead to irrelevant parameterization, making the plots hard to read. 

For example, we already have 3 minor OS versions x 2 RAM size changes as can be seen on the left-hand-side of the benchmarks dashboard: https://nilearn.github.io/benchmarks/

Image

I tried creating a self-hosted runner on INRIA clusters in #5340 , but to get them running involves manual intervention, which is not super convenient:

> Ok so this can work in theory: https://github.com/man-shu/nilearn/actions/runs/14491849971/job/40649877265
> 
> But not super convenient because:
> 
> 1. it won't work on drago because the Ubuntu version is outdated (18.04 with glibc 2.27, [actions-runner](https://github.com/actions/runner) needs >= 2.28)
> 2. it can work on margaret which is a slurm-based cluster. So one would always have to request a node manually and then start the actions-runner service which would then listen to the any requests sent via Github
> 
> Will try to find another solution. Closing for now.
> 
> _Originally posted by @man-shu in https://github.com/nilearn/nilearn/issues/5340#issuecomment-2809371743_

The point of this issue is just to bring this to everyone's attention, open a discussion, and see if we are still interested in that. Or to see if there are other solutions that I might be missing.

Remi-Gau:
- Related to #5966
- Closes #5348 

Use of AI tools:

This PR was produced with the help of an AI coding agent (Claude Code), verifying the "Performance monitoring" section of `CONTRIBUTING.rst` is sufficient for a contributor to actually run the benchmarks, and improving it based on what was found.

- Extend `asv_benchmarks/hashestobenchmark.txt` with all major (non-bugfix) nilearn releases since 0.8: `0.8.0`, `0.9.0`, `0.10.0`, `0.11.0`, `0.12.0`, `0.13.0`, `0.14.0`.
- Document in `CONTRIBUTING.rst` that versions older than 0.11.0 cannot currently be benchmarked this way: their `nilearn/__init__.py` imports `pkg_resources`, which was removed from `setuptools>=81` (the version pip installs by default in the virtualenv asv builds).
- Add a benchmark for `nilearn.plotting.plot_design_matrix_correlation` and one for `nilearn.utils.all_estimators` — both were added recently to nilearn. Their imports are done locally inside `setup()` rather than at module level, since a single top-level import of a function missing from an older nilearn version otherwise breaks the import of the *whole* benchmark file, marking every other (unrelated) benchmark in that file as failed too.
- Document this local-import pattern in `CONTRIBUTING.rst`'s "Adding new benchmarks" section, pointing to `BenchMarkAllEstimators` in `asv_benchmarks/benchmarks/discovery.py` as a concrete example.


.github/workflows/benchmark.yml
```diff
@@ -119,9 +155,63 @@ jobs:
...
+            # to running the full benchmark suite.
+            run: |
+                asv run --show-stderr --verbose --append-samples --parallel 2 --machine fv-az1113-357 -b "NiftiMaskerBenchmark" 2>&1 | tee log_${{ github.event.repository.updated_at }}_${{ github.run_number }}
+        -   name: Run a single benchmark on the PR HEAD and upstream main (PR only)
+            if: github.event_name == 'pull_request'
+            # TODO: restrict to "-b NiftiMaskerBenchmark" temporarily while
+            # debugging this workflow in CI. Remove this filter to go back
+            # to running the full benchmark suite.
+            run: |
+                asv run --show-stderr --verbose --append-samples --parallel 2 --machine fv-az1113-357 -b "NiftiMaskerBenchmark" HASHFILE:pr_hashes.txt 2>&1 | tee log_${{ github.event.repository.updated_at }}_${{ github.run_number }}
```
bthirion: what is the reason to focus on the NiftiMaskerBenchmark in particular ?

Remi-Gau: that's just to get things to run faster in CI on this PR: only runs one benchmark (takes a few seconds rather than 30 minutes)

faster iterations to test things in CI

Remi-Gau:
OK I will merge this so we can get a clean reset on the benchmarks on main.

Issue #6048, opened 2026-02-18
[BUG] fix docs to use whitelist instead of blacklist in datasets prefetching


bpinsard:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Operating system

- [x] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

For example one of the following:
- Linux Ubuntu 22.04
- Mac OS Version 12 "monterey"
- Windows 11

### Python version

- [ ] 3.14
- [ ] 3.13
- [ ] 3.12
- [ ] 3.11
- [x] 3.10

### nilearn version

- dev

### Expected behavior

Should fetch less data, currently docs builds or rerunning examples at tutorials are slow due to large prefetching. 

### Current behavior & error messages

https://github.com/nilearn/nilearn/blob/b029fb89b8d17596263a2018c3d70ff60682b8b0/doc/get_data_examples.py#L25-L42 seems to maybe fetch more files than necessary, with unecessarily slow build/execution times and bloating of CI caches.

### Steps and code to reproduce bug

```python
# Paste your code here


```

Remi-Gau:
select_from_index actually has an inclusion_filters

will try to only select the stuff we need

PR #6425, opened 2026-07-13
[FIX] Only save the monthly doc-build dataset cache once per month

Remi-Gau:
Use of AI tools:

This pull request was generated with the help of an LLM (Claude Code): investigation, diagnosis, and the workflow fix itself were produced by the assistant and reviewed by me.

Relates to #6048

## Why

`test_html.yml` and `build-docs.yml` fail intermittently with 403/503 errors from OSF while fetching datasets such as `development_fmri` and `difumo_atlases`.

The monthly dataset cache (`data_cache-DocumentationBuilder_month-<hash>`) is meant to be shared by every workflow run in a given month. But every run of `build-docs.yml`'s `get_data` job was *also* saving its own ~4GB copy of `nilearn_data` under a unique per-run/per-attempt key, none of which are ever reused. Doing this on every run used to evict the shared monthly entry as soon as the repository-wide 10GB GitHub Actions cache quota was hit — well before its natural 7-day unused-cache TTL — which is why `development_fmri`/`difumo_atlases` would intermittently be missing and have to be re-downloaded live from OSF (the source of the flaky 403/503 errors).

I confirmed this via `gh cache list -R nilearn/nilearn`: only 2 of these ~4GB blobs already consumed ~8GB of the 10GB budget, and via workflow logs, where the exact same monthly cache key that was a hit on one day was a miss a few days later, despite no month rollover.

**Second issue found while fixing the first one**: `doc/get_data_examples.py` fetches a much larger dataset superset for full builds (`build_type in ["full", "html", "html-strict"]`) than for partial builds. `build_type.py` marks a build as full (`html-strict`) on every push to `main`, tags, and `[full doc]`, but PR runs (partial builds) vastly outnumber those. With a single cache saved by whichever run reaches it first each month, it would almost always end up being a partial build's run, freezing the shared cache with only the reports-only subset for the rest of the month — so every subsequent full build (e.g. every push to `main`) would have to live-fetch the full-only datasets (haxby, neurovault, msdl, spm_multimodal, etc.) from OSF on every run, reintroducing the same flakiness for full builds instead. The save step is now gated to only run for full builds, so the shared cache always holds the complete superset once it exists.

## Changes proposed in this pull request

- Collapse the three-tier restore/save logic in `get_data` down to a single monthly cache key that is only *saved* when it doesn't already exist for the month (`cache-hit != 'true'`), so PR/push runs never write a redundant copy.
- Only let a full build (`push` to `main`, tag, `[full doc]`, or scheduled run) perform that save, so the shared cache is always a superset of what partial builds need too.
- Replace the `get_data` -> `build_docs` hand-off within a single workflow run (previously a second, duplicate cache entry) with a short-lived (1 day) `upload-artifact`/`download-artifact` pair, so it no longer competes with the monthly cache for the shared quota.


Remi-Gau:
This should simplify our cache policy and avoid several workflows to fail because we are 'DDoSing' OSF or some other place where we get our data from.

Remi-Gau:
OK let's clear the github cache and merge this to trigger the creations of full doc build data cache

PR #6426, opened 2026-07-13
[MAINT] Fail CI when a benchmark reports as failed

Remi-Gau:
- Related to #5966

Use of AI tools:

This PR was produced with the help of an AI coding agent (Claude Code).

- Drop nilearn versions older than 0.11.0 from `asv_benchmarks/hashestobenchmark.txt` (they cannot currently be benchmarked, see `CONTRIBUTING.rst`).
- Minor wording tweaks to `CONTRIBUTING.rst`.
- **Fix**: the benchmark CI workflow never failed even when a benchmark reported `asv: benchmark failed` in the logs (visible in several past runs pushed to the `benchmarks` repo). Root cause: `asv run` itself exits 0 and keeps going when individual benchmarks fail, and piping its output through `tee` (needed to write the log file) already discards whatever exit code it did have, since a bash pipeline's exit status defaults to that of the last command. Added a step that greps the log for `asv: benchmark failed` after the run and fails the workflow if found.

## Test plan

- [x] Verified locally that `asv run` exits with a non-zero code when a benchmark fails, while the existing `asv run | tee log` pipeline in CI silently exits 0 (confirmed the root cause).
- [x] Verified locally that the new log-grep step correctly detects `asv: benchmark failed` and would exit 1.
- [x] Confirm this PR's own CI run for the benchmark workflow actually fails (due to `tmp_ci_failure_check.py`), then remove the temporary file and the `-b` debug filter in a follow-up commit.


Remi-Gau:
ok seems that we've had some benchmarks that always failed...

emdupre:
Do you think it makes sense to roll those back for now and get them passing in new, separate PRs ?

Remi-Gau:
> Do you think it makes sense to roll those back for now and get them passing in new, separate PRs ?

na: the repo where we store our benchmark is a bit of a mess because the "history" of the benchmarks was "lost" when benchmarks were renamed or moved to a new module.

so I don't think we are losing much at the moment

I will clean the benchmark repo and try to see how we can have some benchmarks across all the stable versions

asv_benchmarks/benchmarks/glm/first_level/first_level.py
Remi-Gau, line 1: move in a module to match the structure of nilearn

asv_benchmarks/benchmarks/glm/io.py
Remi-Gau, line 1: move in a module to match the structure of nilearn

asv_benchmarks/benchmarks/glm/io.py
```diff
@@ -0,0 +1,46 @@
...
+
+        # save_glm_to_bids is permanently moving to nilearn.glm.io in
+        # nilearn 0.15.0: nilearn.interfaces.bids.save_glm_to_bids (its
+        # current, deprecated, location) will be removed then. Import
+        # locally so that benchmarking versions on either side of that
+        # move only fails this benchmark instead of the whole module.
+        try:
+            from nilearn.interfaces.bids import save_glm_to_bids
+        except ImportError:
+            from nilearn.glm.io import save_glm_to_bids
```
Remi-Gau, line 29: added this because otherwise the benchmark will start failing in the next release

CONTRIBUTING.rst
```diff
@@ -1243,6 +1239,19 @@ Adding new benchmarks
...
+not private helpers (leading underscore) or implementation details.
+
+The structure of ``asv_benchmarks/benchmarks`` must mirror
+the structure of the ``nilearn`` package itself:
+a benchmark for a function or class living in ``nilearn/glm/first_level/first_level.py``
+belongs in ``asv_benchmarks/benchmarks/glm/first_level/first_level.py``,
+and a benchmark for ``nilearn/glm/io.py`` belongs in ``asv_benchmarks/benchmarks/glm/io.py``.
+This makes it easy to find the benchmark(s) for a given piece of the public API,
+and keeps benchmark files from growing unrelated content
+as nilearn's own modules get reorganized over time.
```
Remi-Gau, line 1253: given recent discussions I felt it was better to add those to the doc

PR #6430, opened 2026-07-14
[MAINT] Benchmark previous stable versions of Nilearn

Remi-Gau:
- Related to #5966

Use of AI tools:

This PR was produced with the help of an AI coding agent (Claude Code).

Running the full benchmark suite for several commits one after another in a single job is slow, and gets worse as more commits are added to `asv_benchmarks/hashestobenchmark.txt`.

- `list-history-commits`: reads `asv_benchmarks/hashestobenchmark.txt` into a JSON array for use as a matrix.
- `benchmark-history`: a matrix job that benchmarks each commit from that list in parallel, on its own runner (asv's own `--parallel` flag only parallelizes building environments, not running the timed benchmarks themselves, so real parallelism needs separate CI jobs — confirmed by checking `asv run --help`).
- `combine-history`: downloads every matrix entry's results, merges the per-commit/per-machine result files into a single `results` directory, and publishes one combined `asv publish` HTML view as a workflow artifact.

For now this only uploads the combined view as an artifact. The plan (see #5966) is to eventually have it overwrite a previous baseline on the `benchmarks` repo, similar to the existing push-based main-vs-latest-release comparison, so a PR or main HEAD can easily be compared against the last few stable releases.

## Test plan

- [x] Verified locally that the commit-listing script produces valid JSON from `hashestobenchmark.txt`.
- [x] Verified locally that the merge script correctly combines several fake per-commit artifact directories into one `results` directory.
- [x] Verified locally that `asv run "<commit>^!"` correctly benchmarks a single historical commit.
- [x] Verified locally that `asv publish` succeeds without a local `asv-machine.json` (only needed when running new benchmarks, not when publishing existing results).
- [x] Confirm this PR's own CI run actually exercises the matrix + merge mechanism end to end, then remove `pull_request` from the three jobs' triggers and drop the `-b` filter in a follow-up commit.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -53,8 +53,11 @@
 # - Then we push the new results back to the `benchmarks <https://github.com/nilearn/benchmarks>`_ repo.
 #   This will automatically deploy the new results to `nilearn.github.io/benchmarks <https://nilearn.github.io/benchmarks/>`_.
 #
-# A second, independent set of jobs benchmarks the fixed list of release
-# commits in ``asv_benchmarks/hashestobenchmark.txt`` (see ``CONTRIBUTING.rst``).
+# A second, independent set of jobs, run only on schedule or manual
+# trigger (running the full suite for several commits is too expensive
+# to also run on every pull request), benchmarks the fixed list of
+# release commits in ``asv_benchmarks/hashestobenchmark.txt`` (see
+# ``CONTRIBUTING.rst``).
 # Running the full suite for several commits one after another in a single
 # job is slow, so this is split into a matrix job (one commit per matrix
 # entry, run in parallel on separate runners) followed by a job that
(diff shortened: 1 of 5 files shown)


Remi-Gau:
just playing around to see if we can have a way to get an idea of how the performance evolved over the different stable versions

Remi-Gau:
ok increasing the number of repeats per benchmarks makes it hard to scale when we will have more benchmarks, so reverting the last commit

Remi-Gau:
For the record:

Running benchmarks on the last few stable 'major' releases shows improvements of some benchmarks with the last release 

image

Remi-Gau:
OK so this works but the only problem is that the run of benchmark may be run on slightly different architecture for every stable version of nilearn we test, so difference in performance caught here would have to be confirmed manually by running the problematic benchmark locally (on the same architecture for all versions).

Remi-Gau:
OK the benchmark workflow takes too long on PR: will apply the same 'only run the required ones' that we use to run the tests that relates to a certain part of the codebase.

emdupre:
LMK if this is ready for review

Remi-Gau:
> OK the benchmark workflow takes too long on PR: will apply the same 'only run the required ones' that we use to run the tests that relates to a certain part of the codebase.

will do that in a separate PR

bthirion (approved):
The proof is in the pudding...

Remi-Gau:
Ok let's eat pudding then

PR #6416, opened 2026-07-09
[MAINT] further annotate plotting functions

Remi-Gau:
- Closes #3568 

Use of AI tools:
- [ ] LLM tools were used to generate at least part of the content of this pull-request.
- [x] No LLM tools were used to generate any part of the content of this pull-request.

Changes proposed in this pull request:

- add more type annotations to nilearn.plotting functions

PR #6431, opened 2026-07-14
[FIX] find_parcellation_cut_coords fails on labels_img with a trailing singleton 4th dim

Remi-Gau:
- Relates to https://github.com/nilearn/nilearn/pull/6416#issuecomment-4968895530

Use of AI tools:

- PR #6416 renamed the local variable holding the reordered/checked `labels_img` (`loaded_labels_img`) but left `labels_data = get_data(labels_img)` using the original, unchecked image. `check_niimg_3d` squeezes a 4D image with a trailing singleton dimension down to 3D, so `labels_data` (still 4D) and `labels_affine`/hemisphere splits (3D, from the reordered image) became inconsistent, causing `center_of_mass` to raise `ValueError: too many values to unpack`.
- Fix `find_parcellation_cut_coords` in [nilearn/plotting/find_cuts.py](nilearn/plotting/find_cuts.py) to use `get_data(loaded_labels_img)` consistently.
- Add a regression test in [nilearn/plotting/tests/test_find_cuts.py](nilearn/plotting/tests/test_find_cuts.py) reproducing this with a synthetic `(10, 10, 10, 1)` labels image; confirmed it fails on the pre-fix code and passes after the fix.

### DIFF
--- a/nilearn/plotting/tests/test_find_cuts.py
+++ b/nilearn/plotting/tests/test_find_cuts.py
@@ -502,6 +502,29 @@ def test_find_parcellation_cut_coords_hemispheres(affine_mni):
     assert labels == [1]
 
 
+@pytest.mark.ai_generated
+def test_find_parcellation_cut_coords_4d_input_with_trailing_dim(
+    affine_eye,
+):
+    """Regression test for labels_img with a trailing singleton 4th dim.
+
+    ``check_niimg_3d`` (used to compute ``labels_affine``) squeezes such an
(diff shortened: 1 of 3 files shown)


nilearn/plotting/tests/test_find_cuts.py
```diff
@@ -502,6 +502,29 @@ def test_find_parcellation_cut_coords_hemispheres(affine_mni):
     assert labels == [1]
 
 
+@pytest.mark.ai_generated
```
Remi-Gau:
```suggestion
```

PR #6432, opened 2026-07-14
[FIX] Use inclusion_filters instead of exclusion_filters for ds000030 fetch

### LINKED ISSUE #6048: [BUG] fix docs to use whitelist instead of blacklist in datasets prefetching

Remi-Gau:
- Closes #6048

Use of AI tools:

- The `exclusion_filters` used in `examples/04_glm_first_level/plot_bids_features.py`, `doc/get_data_examples.py`, and `doc/visual_testing/reporter_visual_inspection_suite.py` to restrict the `ds000030` download to a single subject/task did not work as intended: `derivatives/task/sub-*/{bart,taskswitch}.feat` do not contain a `task-task` substring, so the `*task-task*` pattern never excluded them, and the full FSL derivatives for those two unrelated tasks were downloaded alongside the `stopsignal` ones.
- Switch all three scripts to `inclusion_filters=["*sub-*stopsignal*"]` (plus two small `exclusion_filters` to drop unused fMRIPrep output-space variants: `space-T1w` and `space-fsaverage`), which only pulls the files actually needed for the `stopsignal` analysis.

**Extra data being downloaded before this fix**, measured against the actual local cache produced by the old `exclusion_filters`:

| | files | size |
|---|---|---|
| Old (`exclusion_filters`) | 719 | 776.2 MB |
| New (`inclusion_filters`) | 168 | 253.1 MB |
| **Excess removed** | **553** | **523.1 MB (67% of the total)** |

Breakdown of the 553 excess files:
- `bart.feat` derivatives (wrong task): 204 files, 149.7 MB
- `taskswitch.feat` derivatives (wrong task): 321 files, 151.1 MB
- raw + fMRIPrep anat derivatives (never read by `first_level_from_bids`, which only globs `func/` derivatives): 19 files, 207.2 MB
- misc top-level metadata not required by `first_level_from_bids` (`README`, `dataset_description.json`, `participants.tsv`, other tasks' sidecar jsons, fMRIPrep HTML reports): 9 files, 15.1 MB

Verified end-to-end against a clean cache directory: `first_level_from_bids` + contrast computation + FSL comparison still run successfully with the reduced file set.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/doc/changes/latest.rst
+++ b/doc/changes/latest.rst
@@ -22,6 +22,8 @@ NEW
 Fixes
 -----
 
+- :bdg-dark:`Code` Fix ``examples/04_glm_first_level/plot_bids_features.py``, ``doc/get_data_examples.py``, and ``doc/visual_testing/reporter_visual_inspection_suite.py`` downloading the full FSL derivatives for the ``bart`` and ``taskswitch`` tasks of the ``ds000030`` dataset in addition to ``stopsignal``, because the ``*task-task*`` exclusion filter did not match the ``derivatives/task/sub-*/taskswitch.feat`` folder name; switch to ``inclusion_filters`` to only fetch the files needed for the ``stopsignal`` analysis (:gh:`6432` by `Rémi Gau`_).
+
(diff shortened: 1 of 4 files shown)


Remi-Gau:
Checked manually and some more files could be excluded compared to what is described in the top message.

PR #6433, opened 2026-07-15
[FIX] Fix subject-level anatomical BIDS lookup

### LINKED ISSUE #6291: get_bids_file cannot find anat files in fmriprep datasets with several sessions
This is because in this case anat data is averaged across sessions and we get a single anat folder under the subject folder at the same level as some of the session folder

```
inputs/ds000114-fmriprep/sub-01
├── anat <--------- contains all the relevant images
├── figures
├── log
├── ses-retest
│   ├── anat
│   └── func
└── ses-test
    ├── anat
    └── func
```

While I am afraid of scope creep I believe this is still a common use case that we may want to cover.

mohammedsadeghi:
- Closes #6291

Use of AI tools:

- Add an anatomical-only subject-level fallback when a multi-session BIDS dataset has no matching session-level anatomical derivative for a subject.
- Preserve session-specific functional lookup and prefer matching session-level anatomical derivatives.
- Add regression coverage for session precedence, entity filters, multiple subjects, no-match behavior, and session isolation.

Root cause:

- When a dataset contained session directories, `get_bids_files` searched only session-level modality folders. This omitted shared fMRIPrep anatomical derivatives stored in `sub-*/anat`.

Validation:

- `pytest nilearn/interfaces/tests/test_query.py nilearn/interfaces/fmriprep/tests/test_load_confounds.py -q` — 136 passed, 2 pre-existing xfails
- Targeted pre-commit checks, including ruff, import-linter, doc8, and mypy.


nilearn/interfaces/bids/query.py
```diff
@@ -281,26 +285,71 @@ def get_bids_files(
...
+                parsed_files = [
+                    file_
+                    for file_ in parsed_files
+                    if (entity not in file_["entities"] and label == "")
+                    or (
+                        entity in file_["entities"]
+                        and file_["entities"][entity] == label
+                    )
+                ]
+            return parsed_files
```
Remi-Gau: extract the function: do not keep it as local function

should be easier to test it

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +333,201 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+        "sub-01/anat/sub-01_space-MNI152NLin6Asym_desc-preproc_T1w.nii.gz",
+    )
+    _add_bids_file(
+        tmp_path,
+        "sub-01/ses-test/func/sub-01_ses-test_task-rest_bold.nii.gz",
+    )
+    _add_bids_file(
+        tmp_path,
+        "sub-01/ses-retest/func/sub-01_ses-retest_task-rest_bold.nii.gz",
+    )
```
Remi-Gau:
```suggestion
```

I don't think those are needed

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +333,201 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+    _add_bids_file(
+        tmp_path,
+        "sub-01/anat/"
+        "sub-01_ses-retest_space-MNI152NLin6Asym_desc-preproc_T1w.nii.gz",
+    )
```
Remi-Gau:
```suggestion
```
not a valid BIDS filename

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +333,201 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+        file_type="nii.gz",
+        filters=[("ses", "test")],
+    )
+
+    assert selection == [str(session_anat)]
+
+
+@pytest.mark.ai_generated
+def test_get_bids_files_fmriprep_session_level_anat_takes_precedence(tmp_path):
+    """Prefer a matching session-level anatomical derivative."""
```
Remi-Gau: OK I don't think that should be the behavior: if the filter includes session, then we should only return the session level anat.

If no session filter is passed we should return both

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +333,201 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+
+    selection = get_bids_files(
+        tmp_path,
+        modality_folder="anat",
+        file_tag="T1w",
+        file_type="nii.gz",
+        filters=[("ses", "test"), ("desc", "preproc")],
+    )
+
+    assert selection == [str(session_anat), str(subject_anat)]
```
Remi-Gau: here again we are filtering by session, so the subject anat should not be there

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +333,201 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+
+    selection = get_bids_files(
+        tmp_path,
+        modality_folder="func",
+        file_tag="bold",
+        file_type="nii.gz",
+        filters=[("ses", "test")],
+    )
+
+    assert selection == [str(test_file)]
```
Remi-Gau:
```suggestion
```

that feels unnecessary

doc/changes/latest.rst
```diff
@@ -22,6 +22,11 @@ NEW
 Fixes
 -----
 
+- :bdg-dark:`Code` Fix :func:`~interfaces.bids.get_bids_files` missing
+  subject-level anatomical derivatives in multi-session datasets. Matching
+  session-level anatomical derivatives are preferred, while functional file
+  lookups remain session-specific (:gh:`6291` by `Mohammad Sadeghi Hardengi`_).
```
```suggestion
- :bdg-dark:`Code` Fix :func:`~interfaces.bids.get_bids_files` missing subject-level anatomical derivatives in multi-session datasets (:gh:`6291` by `Mohammad Sadeghi Hardengi`_).
```

nilearn/interfaces/bids/query.py
```diff
@@ -224,6 +224,9 @@ def get_bids_files(
         of the dataset provider. For example the 'func' and 'anat' standard
         folders. If given as the empty string '', files will be searched
         inside the sub-label/ses-label directories.
+        For ``'anat'``, subject-level anatomical derivatives are considered
+        as a fallback in datasets containing session directories. Matching
+        session-level anatomical derivatives take precedence per subject.
```
Remi-Gau:
```suggestion
```

Remi-Gau:
I am also realizing that the original issue was too specific for anat files.

If you have a derivative dataset with:

- sub-01/func/sub-01_task-foo_space-MNI_desc-mean_bold.nii
- sub-01/ses-01/func/sub-01_ses-01_task-foo_space-MNI_desc-preproc_bold.nii
- sub-01/ses-02/func/sub-01_ses-02_task-foo_space-MNI_desc-preproc_bold.nii

and you use get_bids_file to get func files, then you should get 3 files.

mohammedsadeghi:
> I am also realizing that the original issue was too specific for anat files.
> 
> If you have a derivative dataset with:
> 
> * sub-01/func/sub-01_task-foo_space-MNI_desc-mean_bold.nii
> * sub-01/ses-01/func/sub-01_ses-01_task-foo_space-MNI_desc-preproc_bold.nii
> * sub-01/ses-02/func/sub-01_ses-02_task-foo_space-MNI_desc-preproc_bold.nii
> 
> and you use get_bids_file to get func files, then you should get 3 files.

Thank you for your guidance and patience.

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +334,159 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+        "sub-01_ses-test_space-MNI152NLin6Asym_desc-preproc_T1w.nii.gz",
+        "sub-01_ses-retest_space-T1w_desc-brain_T1w.nii.gz",
+    ]
+
+    selection = _filter_bids_files(
+        files,
+        [("ses", "test"), ("space", "MNI152NLin6Asym")],
+    )
+
+    assert selection[0]["file_path"] == files[0]
```
Remi-Gau, line 358:
```suggestion
    assert len(selection) == 1
    assert selection[0]["file_path"] == files[0]
```

I would expect a single file in the output

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +334,159 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+
+
+@pytest.mark.ai_generated
+def test_get_bids_files_fmriprep_subject_level_anat(tmp_path):
```
Remi-Gau: One suggestion given that those tests are very similar:

let's have a single test with:
- same "dummy dataset" gets set up once using your _add_bids_file
- test parametrization with different "filters, n_expected_files"

for an example see the 'test_get_bids_files' in the same module

nilearn/interfaces/bids/query.py
```diff
@@ -281,30 +282,55 @@ def get_bids_files(
...
         )
+        if modality_folder == "anat" and session_folder_exists:
+            subject_level_anat_files = glob.glob(
+                str(
+                    main_path
+                    / f"sub-{sub_label}"
+                    / "anat"
+                    / f"sub-{sub_label}*_{file_tag}.{file_type}"
+                )
+            )
```
Remi-Gau:
```suggestion
        if modality_folder in ["anat", "func"] and session_folder_exists:
            subject_level_files = glob.glob(
                str(
                    main_path
                    / f"sub-{sub_label}"
                    / modality_folder
                    / f"sub-{sub_label}*_{file_tag}.{file_type}"
                )
            )
```

We may sometimes have 'func' at the subject level (see previous comment).

nilearn/interfaces/bids/query.py
```diff
@@ -281,30 +282,55 @@ def get_bids_files(
...
-        return [ref_file["file_path"] for ref_file in files]
+    if filters or subject_level_anat_files:
+        filtered_files = _filter_bids_files(files, filters)
+        if subject_level_anat_files and not any(
+            entity == "ses" for entity, _ in filters
+        ):
+            filtered_files.extend(
+                _filter_bids_files(subject_level_anat_files, filters)
+            )
+            filtered_files.sort(key=lambda file_: str(file_["file_path"]))
```
Remi-Gau:
```suggestion
    if filters or subject_level_files:
        filtered_files = _filter_bids_files(files, filters)
        if subject_level_files and not any(
            entity == "ses" for entity, _ in filters
        ):
            filtered_files.extend(
                _filter_bids_files(subject_level_files, filters)
            )
            filtered_files.sort(key=lambda file_: str(file_["file_path"]))
```

nilearn/interfaces/bids/query.py
Remi-Gau: probably worth adding a test for the `func` case too

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +334,82 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+            [("space", "MNI152NLin6Asym"), ("desc", "preproc")],
+            3,
+        ),
+        ("anat", "T1w", [("desc", "preproc")], 4),
+        (
+            "anat",
+            "T1w",
+            [("ses", "test"), ("space", "MNI152NLin6Asym")],
+            1,
+        ),
```
Remi-Gau, line 386:
```suggestion
        ),
        ("anat", "T1w", [("ses", "")], 1),
```

I am adding one test case: when an entity has an empty string it means that files containing this entity should NOT be returned: so in this case the function should only return anat files in sub-01/anat

See https://github.com/nilearn/nilearn/pull/4802 and the test test_get_bids_files_no_space_entity

I suspect the code will have to be adapted to make this pass.

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +334,82 @@ def test_get_bids_files_fmriprep(tmp_path):
...
+    selection = _filter_bids_files(
+        files,
+        [("ses", "test"), ("space", "MNI152NLin6Asym")],
+    )
+
+    assert len(selection) == 1
+    assert selection[0]["file_path"] == files[0]
+
+
+@pytest.mark.ai_generated
```
Remi-Gau:
```suggestion
```

Remove before merge

nilearn/interfaces/tests/test_query.py
```diff
@@ -333,6 +334,82 @@ def test_get_bids_files_fmriprep(tmp_path):
...
 
 
+def _add_bids_file(bids_path, relative_path):
+    file_path = bids_path / relative_path
+    file_path.parent.mkdir(parents=True, exist_ok=True)
+    file_path.touch()
+    return file_path
+
+
+@pytest.mark.ai_generated
```
Remi-Gau:
```suggestion
```

mohammedsadeghi: Hi,
I hope you’re doing well.
Are the latest changes okay with you?

nilearn/interfaces/tests/test_query.py
Remi-Gau: @mohammedsadeghi 
I have applied this suggestion and this will probably require some updates to your changes.

Remi-Gau: OK the test pass and I would not have expected that: let me run it locally to make sure everything is fine

I may have the wrong expectation.

mohammedsadeghi: Sorry I just saw this comment.
Tell me if any change needs to be done by me.

Remi-Gau: I think you are all good! 
I just needed to check something manually locally to better understand it: I detect a small duplication in code that I removed.

Let's wait for CI to finish and then we can probably merge this

